Issue #4063, opened 2019-07-15
Remove Send Tab build flags from Push (and maybe Send Tab)


jonalmeida:
We're currently using the same build flag for Send Tab and Push (Firebase) since those two features were strongly coupled when landing this feature. When we find Send Tab and Push are working in a decent state, we should remove these flags from there.

We need to also remove the connection between the AccountManager and the push service that observes the manager and starts/stops itself on that state.

Also, if Send Tab is functioning well, we can remove the flags from the send tab feature as well. 

See: [`BackgroundServices#accountObserver`](https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/java/org/mozilla/fenix/components/BackgroundServices.kt#L138-L150).

PR #4932, opened 2019-08-26
For #4063 - Always enable the push service

boek:


app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -55,6 +55,7 @@ open class FenixApplication : Application() {
 
         setupMegazord()
         setupLogging()
+        setupPush()
```
pocmo: I assume this needs to move lower and should only happen in the main process and not all processes.

Issue #4783, opened 2019-08-16
Enhanced Tracking Protection needs Page Load Testing


colintheshots:
## User Story
- As a member of the Fenix team, I would like performance measurements on the relative page load speeds of Tracking Protection versus the new Enhanced Tracking Protection feature.

## Dependencies
- This will require some assistance from the Performance Team to learn how to run comparative Raptor testing using Try with live loads and custom parameters and/or APKs.

### Acceptance Criteria
- We have accurate page load performance measurements of ETP vs. TP on the latest Fenix with enough runs to have reasonable certainty in the results.

vesta0:
The goal of this experiment is to determine performance impact of switching from TP (current) to ETP, and based on the results make a decision on whether Fenix default/recommended setting should be Basic or Strict (see [here](https://mozilla.invisionapp.com/share/R8T6GDUDW3B#/screens))

Cohort a (ETP):
- Block Cross-site Tracking Cookies from Disconnect's Advertising, Analytics, and Social list

Cohort B (Current TP implementation):
- Blocking tracking scripts (not just the tracking cookies) from Disconnect's Advertising, Analytics, and Social list

@Amejia481 and @ekager would you please add the name of lists/prefs to be used for each cohort? Thank you!

Amejia481:
**Cohort a (ETP):**

- `TrackingProtectionPolicy.recommended()`

**Cohort B (Current TP implementation):**

- `TrackingProtectionPolicy.strict()`

liuche:
Remaining work E2.

vesta0:
> **Cohort a (ETP):**
> 
>     * `TrackingProtectionPolicy.strict()`

I don't believe this is correct. ETP (#1749) is a new category and refers to blocking Cross-site Tracking Cookies from Disconnect's Advertising, Analytics, and Social list. It is less strict than what we currently block in Fenix, and corresponds to the new `standard` option in the [new UI](https://mozilla.invisionapp.com/share/R8T6GDUDW3B#/screens).
> 
> 
> **Cohort B (Current TP implementation):**
> 
>     * `TrackingProtectionPolicy.recommended()`

The current TP implementation (blocking tracking scripts/sub-resources, not just cookies) is more strict and corresponds to the `strict` option in the new UI.

Amejia481:
The default policies on ac have been updated on `12.0.0` to be aligned with the mockups and what Desktop is using https://github.com/mozilla-mobile/android-components/issues/4349.

[...]

PR #4935, opened 2019-08-27
Perf test strict mode for #4783

colintheshots:
Testing ETP. Pay no attention to this draft PR. I'm merely creating a TaskCluster index using the PR's artifacts.


colintheshots:
This was a nice thought, but the resulting APK isn't signed for testing. I'll come up with a better option.

colintheshots:
I got what I needed here.

PR #4936, opened 2019-08-27
No issue: fix Nightly Raptor performance runs

colintheshots:
The location of all artifacts changed, so all nightly Raptor runs are breaking.

### DIFF
--- a/automation/taskcluster/lib/tasks.py
+++ b/automation/taskcluster/lib/tasks.py
@@ -12,7 +12,6 @@ import taskcluster
 from lib.util import upper_case_first_letter, convert_camel_case_into_kebab_case, lower_case_first_letter
 
 DEFAULT_EXPIRES_IN = '1 year'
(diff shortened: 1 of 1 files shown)


automation/taskcluster/lib/tasks.py
```diff
@@ -627,8 +626,10 @@ def _craft_raptor_task(
             name_prefix, '(on 64-bit-device)' if force_run_on_64_bit_device else ''
         )
 
+        default_apk_artifact_location = 'public/build/{}/geckoNightly/target.apk'.format('armeabi-v7a' if abi == 'arm' else 'arm64-v8a')
+
         apk_url = '{}/{}/artifacts/{}'.format(_DEFAULT_TASK_URL, signing_task_id,
-                                                        DEFAULT_APK_ARTIFACT_LOCATION)
+                                              default_apk_artifact_location)
```
jhugman: Nit: naming. Is this overridable now. Is it just the `apk_artifact_location` (i.e. not just the default one).

PR #4966, opened 2019-08-27
Remove SendTab feature flags and workarounds

### LINKED ISSUE #4218: [Push] Remove nightly workaround for force registration
When we have majority of our users on the latest update (which forces re-registration of push), we can remove our [hacky change](https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/java/org/mozilla/fenix/components/BackgroundServices.kt#L120-L131) that enables it.

jonalmeida:
We've had Send Tab enabled for all builds for a while now. We're also using Firebase push for more things so it's safe to remove the flag.

This also fixes #4218
### Pull Request checklist


jonalmeida:
1.4 branch is created so this is safe to land now.

PR #4986, opened 2019-08-28
Use Mozilla Android Components 10.0.1.

pocmo:
Time to use a stable Android Components version again in order to get ready to cut the release branch.


app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
```diff
@@ -116,7 +116,7 @@ class DefaultToolbarMenu(
                     ThemeManager.resolveAttribute(R.attr.primaryText, context),
                 highlight = if (hasAccountProblem) {
                     BrowserMenuHighlightableItem.Highlight(
-                        imageResource = R.drawable.ic_alert,
+                        startImageResource = R.drawable.ic_alert,
```
pocmo: @sblatz Is that the right change after the API change?

pocmo: We talked on Slack. `endImageResource` is the right one here.

Issue #4889, opened 2019-08-22
Enable UI tests on github-pull-request


rpappalax:
We would like to enable UI tests to run on github-pull-request. Currently they are only activated for github-push:
https://github.com/mozilla-mobile/fenix/blob/master/automation/taskcluster/decision_task.py#L82

We can easily enable this in the decision_task, but will need some assistance from taskcluster team to ensure we have proper permissions/scopes.
@JohanLorenzo can you assist with this?

This is currently enabled on mozilla-mobile/firefox-tv and we'd like to also enable it for fenix.

Workflow would be:

1. I work on a patch
2. I submit a PR
3. Taskcluster runs unit tests and UI tests
4. Taskcluster must pass ALL of the tests for me to get a green check
5. I merge

If a contributor submits a patch, we will manually trigger step 3 as allowPullRequests will be restrictec to collaborators as in firefox-tv project:
https://github.com/mozilla-mobile/firefox-tv/blob/master/.taskcluster.yml#L6

rpappalax:
@sblatz @npark-mozilla

JohanLorenzo:
Some patches are also referenced under https://bugzilla.mozilla.org/show_bug.cgi?id=1577479

rpappalax:
This has already been merged here:

@rpappalax rpappalax referenced this issue on 30 Aug
Bug 1577479 - Run UI tests on PRs #5011

PR #4996, opened 2019-08-28
Restrict Taskcluster to collaborators on PR

JohanLorenzo:
Pre-requisite to solve #4889 

I understand we need UI tests to run on PRs because many regressions happened over the past weeks. @rpappalax told me we get a couple of PRs opened by external contributors each week. So an okay-solution is to not run TC at all on PRs for external contributors.

This way, we can run privileged tests like the UI ones on PRs.  

### Pull Request checklist

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -3,7 +3,8 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 version: 1
 policy:
-  pullRequests: public
+  # XXX We restrict taskcluster to collaborators so priviledged tests (like UI tests) can run on PRs
+  pullRequests: collaborators
 tasks:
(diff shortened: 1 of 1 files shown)


JohanLorenzo:
Interesting! Do you have a link to Taskcluster + Github hooks? I don't think I've worked with them before.

pocmo:
This is one of the reasons why AC uses `bors-ng`: It runs (among other things) UI tests on a repository branch before merging. It also has the advantage that it runs those tasks against the master state (= the state master is in after merge) and not against the random state the (unrebased) PR is in.

pocmo:
> I think bors-ng sounds cool, but anyone can type a comment. Does it enforce repo permissions?

Of course.

PR #5023, opened 2019-08-29
Use new async API to load and access search engines

hawkinsw:
Use the new asynchronous features of the search engine
manager from a-c to avoid the latency required to load
the default search engine.

@csadilek and I worked on this PR together (I mostly typed what he told me to type). 

### Pull Request checklist

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/Search.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/Search.kt
@@ -31,12 +31,11 @@ class Search(private val context: Context) {
         ).apply {
             registerForLocaleUpdates(context)
             GlobalScope.launch {
-                loadAsync(context).await()
+                defaultSearchEngine = getDefaultSearchEngineAsync(
+                    context,
+                    Settings.getInstance(context).defaultSearchEngineName
+                )
             }
-            defaultSearchEngine = getDefaultSearchEngine(
-                context,
-                Settings.getInstance(context).defaultSearchEngineName
-            )
         }
     }
 }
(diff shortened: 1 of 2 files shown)


csadilek:
> @csadilek and I worked on this PR together (I mostly typed what he told me to type).

@hawkinsw deserves more credit :). He found these two cases where we (unnecessarily) block the main thread waiting for the default search engine to be loaded. We added new API to AC to make this easier.

With this we can create the `SearchSuggestionProvider` without the default search engine by providing the `SearchEngineManager`, which will load the default engine asynchronously. We can start rendering the UI and load the engine(s) in parallel. Fenix already has logic in place to update the UI when the search engines are loaded (e.g. in case it hasn't completed by the time the HomeFragment is created).

There's some follow-up work here to switch all calls to `getDefaultSearchEngine` over to the new `getDefaultSearchEngineAsync`, but this small change here should address the performance problem Will identified.

The "worst case" here is that the user resumes directly to the BrowserFragment and immediately starts typing a search, which could cause the main thread to block if the search engines haven't been loaded. We weren't able to reproduce this case and it seems like the right trade-off. I will do another round of testing tomorrow.

boek:
@csadilek sounds good. Can you file issues for any follow up that is needed for this? 👍

csadilek:
@boek will do, I am also just testing this change and actually running into a problem where the `HomeFragment` doesn't update once the engines are loaded. I'll investigate and file follow-up issues or put up a fix.

csadilek:
Follow-up PR to fix the race I found when testing this earlier: https://github.com/mozilla-mobile/fenix/pull/5049

PR #5035, opened 2019-08-30
Rename *Store/State/Action to *FragmentStore/State/Action

pocmo:
This series of patches renames our `*Store/State/Action` to `*FragmentStore/State/Action`(For example `BrowserStore/BrowserState/BrowserAction` to `BrowserFragmentStore/BrowserFragmentState/BrowserFragmentAction`.

The reason for this is that the AC team is close to starting to migrate components away from `browser-session` to `browser-state` (from which we derived the more generic `lib-state`). This means we will introduce `browser-state` to Fenix soon. `brower-state` already has classes like `BrowserStore`, `BrowserAction` and `BrowserState` keeping the overall global state of the browser (what used to be in `SessionManager` and `Session`). Avoiding name conflicts and confusion lead us to proposing to rename the classes in Fenix (I talked to Christian about this a while ago).

At first I wasn't happy about the naming. But now I think it's actually better to add the "scope" to the name. It will make it clear that `BrowserFragmentState` is the state of `BrowserFragment` (as opposed to the state of the whole browser) and dispatching a `BrowserFragmentAction` on the `BrowserFragmentStore` will change the state of the `BrowserFragment` (and not the whole browser). The same is true for other classes like "bookmarks": `BookmarksFragmentStore` deals with the state of the `BookmarksFragment` and is not responsible for bookmarks in general (which may have been a fair assumption from the name).


csadilek:
@boek can you take a look at this rename as well?

PR #5042, opened 2019-08-30
Add browser-state component.

pocmo:
Related to PR #5035.

AC is soon landing patches that will migrate components from `browser-session` to `browser-state`. This will mean that those components will want a `BrowserStore` instead of a `SessionManager`. With this patch we will already have it.

Passing the `BrowserStore` instance to `SessionManager` will take care of keeping both in sync until eventually `SessionManager` is no longer needed anymore (🍻).


pocmo:
Now part of #5100.

PR #5043, opened 2019-08-30
Reduce require calls in BrowserFragment 🧐

NotWoods:
No behavioural changes, just reducing how much `!!` and `require` need to be typed out.

### Pull Request checklist

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
@@ -148,12 +148,13 @@ abstract class BaseBrowserFragment : Fragment(), BackHandler, SessionManager.Obs
     @Suppress("ComplexMethod", "LongMethod")
     @CallSuper
     protected open fun initializeUI(view: View): Session? {
-        val sessionManager = requireComponents.core.sessionManager
+        val context = requireContext()
+        val sessionManager = context.components.core.sessionManager
 
         return getSessionById()?.also { session ->
 
             val browserToolbarController = DefaultBrowserToolbarController(
-                context!!,
+                context,
                 findNavController(),
                 (activity as HomeActivity).browsingModeManager,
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
```diff
@@ -193,7 +194,7 @@ abstract class BaseBrowserFragment : Fragment(), BackHandler, SessionManager.Obs
 
             findInPageIntegration.set(
                 feature = FindInPageIntegration(
-                    sessionManager = requireComponents.core.sessionManager,
+                    sessionManager = context.components.core.sessionManager,
```
csadilek: nit: This could just be `sessionManager = sessionManager` as we have val declared on line 152?

app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
```diff
@@ -227,19 +228,19 @@ abstract class BaseBrowserFragment : Fragment(), BackHandler, SessionManager.Obs
             )
 
             windowFeature.set(
-                feature = WindowFeature(requireComponents.core.sessionManager),
+                feature = WindowFeature(context.components.core.sessionManager),
```
csadilek: nit: just `sessionManager`?

app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
```diff
@@ -88,29 +89,30 @@ class BrowserFragment : BaseBrowserFragment(), BackHandler {
...
 
         return super.initializeUI(view)?.also {
 
             thumbnailsFeature.set(
                 feature = ThumbnailsFeature(
-                    requireContext(),
+                    context,
                     view.engineView,
-                    requireComponents.core.sessionManager
+                    context.components.core.sessionManager
```
csadilek: nit: same here, we have a session manager val we can use.

Issue #3440, opened 2019-06-13
Decide to use requireComponents or safe unwrap context


sblatz:
Currently we use the two interchangeably throughout the code base and it doesn't seem to have reasoning behind it (or at least reasoning that is transparent to other engineers). I'm of the mind that we should just be using `requireComponents` so we don't have silent failures.

ekager:
there are still a few `context!!`s in our codebase that can be converted to `requireContext()`s so we can close out this ticket

PR #5044, opened 2019-08-30
For #3440 - Reduce HomeFragment require calls

NotWoods:
By using `val context = requireContext()` we can reduce a few require checks. Also removed some dead code.

### Pull Request checklist

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
@@ -38,6 +38,8 @@ import com.google.android.material.snackbar.Snackbar
 import kotlinx.android.synthetic.main.fragment_home.*
 import kotlinx.android.synthetic.main.fragment_home.view.*
 import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.Dispatchers.IO
+import kotlinx.coroutines.Dispatchers.Main
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -270,17 +273,19 @@ class HomeFragment : Fragment(), AccountObserver {
...
-                collections = requireComponents.core.tabCollectionStorage.cachedTabCollections
+                mode = currentMode(context),
+                collections = context.components.core.tabCollectionStorage.cachedTabCollections
             )
         )
 
         (activity as AppCompatActivity).supportActionBar?.hide()
 
-        requireComponents.backgroundServices.accountManager.register(this, owner = this)
+        context.components.backgroundServices.accountManager.register(this, owner = this)
```
jhugman: Please can we go one step further and use

```kotlin
val components = requireContext().components
```

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -270,17 +273,19 @@ class HomeFragment : Fragment(), AccountObserver {
                 }
             }
 
+        val context = requireContext()
+
         getManagedEmitter<SessionControlChange>().onNext(
             SessionControlChange.Change(
                 tabs = getListOfSessions().toTabs(),
-                mode = currentMode(context!!),
```
jhugman: 😻

csadilek:
@NotWoods this one needs another rebase. Let's land it :).

Issue #4892, opened 2019-08-22
String needs update to apostrophe


Delphine:
I didn't block on this in quarantine, as it can be fixed in a follow-up commit:
preferences_delete_browsing_data_cookies_subtitle : string needs a curly apostrophe.

Note: no need to update string ID for this one :)
thanks!

mcarare:
@Delphine how about sign_out_confirmation_message ?

Delphine:
@mcarare Good catch, that seems to have fallen through the cracks 🤔
Absolutely!

mcarare:
Fixed by #5062

PR #5062, opened 2019-09-02
For #4892: Replaced straight single quote with curly apostrophe in contractions

mcarare:
### Pull Request checklist

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -643,7 +643,7 @@
     <!-- Title for the cookies item in Delete browsing data -->
     <string name="preferences_delete_browsing_data_cookies">Cookies</string>
     <!-- Subtitle for the cookies item in Delete browsing data -->
-    <string name="preferences_delete_browsing_data_cookies_subtitle">You\'ll be logged out of most sites</string>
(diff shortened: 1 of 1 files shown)


ekager:
This may have the same issue as https://github.com/mozilla-mobile/fenix/pull/5081 ? I'm not sure we can change strings in our files because they changes may be overwritten with the next l10n import

mcarare:
I haven't seen the process documented somewhere, but I thought only new strings are being imported.

Delphine:
This should be good because it's dealing with the original en-US strings (vs actual localizations like in https://github.com/mozilla-mobile/fenix/pull/5081).
However, the backslash should be taken out as well in these strings (so, just leave the curly apostrophe).
Localizers have probably decided to use their correct apostrophe already, so we don't need a string ID here to notify them all of this change. Just udpating these strings should be enough, and we'll pick them up once they land on master on your side
(CC @Pike just in case I'm missing something here)

mcarare:
@Delphine Thank you for clarifying these string issues. I've also removed the preceding backslash not needed for escaping anymore.

Pike:
This fix is right, en-US strings are maintained here, l10n strings are maintained in l10n.

Also, typographic fixes shouldn't get ID changes.

Issue #4517, opened 2019-08-02
[Bug] inconsistent back button behaviour in library


cadeyrn:
### Steps to reproduce

1a) open your bookmarks
1b) open a bookmark
1c) press the back button

2a) open your history
2b) open a page from your history
2c) press the back button

### Expected behavior

Either return to the library view or to the start screen, but do the same in both cases. 

I clearly prefer the behaviour of the history (return to the library view).

### Actual behavior

after 1c) Fenix opens the start screen
after 2c) Fenix returns to the history

### Device information

* Android device: HTC U11 / Android 8.0
* Fenix version: master branch revision 46b2d6f937f267137f7744256d829eb5d5d40550

boek:
@cadeyrn Good catch!

ValentinTimisica:
Hi @boek , Can you please tell us how it should work?
1) Both return to the library.
2) Both return to the home screen.

I saw in the [nav_graph.xml](https://github.com/mozilla-mobile/fenix/blob/8194e0778e38ed95b769c13d91b89919fa5b4b38/app/src/main/res/navigation/nav_graph.xml#L201) that currently this is the expected behaviour for bookmark fragment (to go back to the home screen in this case).
```
<action
      android:id="@+id/action_bookmarkFragment_to_browserFragment"
      app:destination="@id/browserFragment"
      app:popUpTo="@id/libraryFragment"
      app:popUpToInclusive="true" />
```

eggsgent:
The current behaviour is terrible, not only does the Android back button not return the user to the Bookmark tree view or History view which the user had (painstakingly for bookmark power users with folder structures) navigated to before opening an entry (at least make it an option!), but currently it also closes the browser and returns the user to the Android homescreen.

1.3.1 build 12341950

ekager:
AFAIK no UX decision has been made about this behavior, so I'm going to ask them for feedback here. 
UX - When going to browser from **home** via a library page: do we want pressing back from browser to always get you to home (current behavior from bookmarks), or do you want to maintain the stack and keep bookmarks/history so you can press back from browser to get back to them (current behavior from history)?

lime124:
@topotropic is out at a conference but can respond tomorrow when she returns. thanks!

topotropic:
Let's maintain the stack and get people back to history or bookmarks when tapping on back button.

sv-ohorvath:
@ekager 
STR: 
1. Open a webpage.
[...]

PR #5064, opened 2019-09-02
For #4517: Keep Bookmarks fragment when navigating to browser

mcarare:
Removed popUp when navigating from bookmarks to browser

### Pull Request checklist

### DIFF
--- a/app/src/main/res/navigation/nav_graph.xml
+++ b/app/src/main/res/navigation/nav_graph.xml
@@ -242,9 +242,7 @@
             app:argType="string" />
         <action
             android:id="@+id/action_bookmarkFragment_to_browserFragment"
-            app:destination="@id/browserFragment"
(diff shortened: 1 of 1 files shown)


mcarare:
@ekager, requested review from you because behaviour was introduced in https://github.com/mozilla-mobile/fenix/pull/2665 and https://github.com/mozilla-mobile/fenix/pull/3043
Also tested the issues, not reproducing with PR changes. TY!

ekager:
See my comment in #4517 - I don't think this behavior has been decided on by UX so I asked them for a decision

colintheshots:
UX: We need feedback on the intended behavior here as mentioned in bug #4517.

topotropic:
https://github.com/mozilla-mobile/fenix/issues/4517#issuecomment-528755522

Issue #5053, opened 2019-08-31
[Bug] share menu, copy url crashes fenix on Android Q


raghur:
## Steps to reproduce
Og pixel xl on Android Q beta 6
Open any site. Click share menu. Click copy url

### Expected behavior
Fenix does not crash

### Actual behavior
Fenix crashes

### Device information

* Android device: yes
* Fenix version: 2.0.0 rc1

mcarare:
Issue is reproducible on Android P, and I suspect any Android device with recent updates to Google Maps.
@raghur Please check if you have in your share list multiple "copy to clipboard" items. The one causing the crash is actually an intent handler from Maps. The other one should work just fine.
I will submit a PR to handle the issue.

raghur:
I have only one 'Copy to clipboard' Item in the share menu

![image](https://user-images.githubusercontent.com/516267/64251635-98c8f080-cf36-11e9-9103-f8471919e8f5.png)

matfantinel:
I'm on Android P, Galaxy S10e.

I can reproduce the issue, but in my case it has two items in the share menu: "Copy to clipboard" and "Copy". The latter works just fine, while the first crashes the browser.

ekager:
Closing for #4911

Issue #3371, opened 2019-06-11
FNX3-16972 ⁃ [Bug] Crash when selecting copy item from Share Menu


ekager:
From the trace it looks like it's my adobe CC app that's giving this second "copy to clipboard" option here. Not sure if there's anything we can do because Pocket's share sheet has the same crash. Maybe we could catch it though? 

> 2019-06-11 15:19:04.776 8014-8014/org.mozilla.fenix.debug E/AndroidRuntime: FATAL EXCEPTION: main
>     Process: org.mozilla.fenix.debug, PID: 8014
>     java.lang.SecurityException: Permission Denial: starting Intent { act=android.intent.action.SEND typ=text/plain flg=0x10000001 cmp=com.adobe.cc/.share.CopyToClipboardActivity clip={text/plain T:https://blog.mozilla.org/firefox/firefox-data-privacy-promise/} (has extras) } from ProcessRecord{e0f3ff1 8014:org.mozilla.fenix.debug/u0a400} (pid=8014, uid=10400) not exported from uid 10312
>         at android.os.Parcel.createException(Parcel.java:2069)
>         at android.os.Parcel.readException(Parcel.java:2037)
>         at android.os.Parcel.readException(Parcel.java:1986)
>         at android.app.IActivityTaskManager$Stub$Proxy.startActivity(IActivityTaskManager.java:3821)
>         at android.app.Instrumentation.execStartActivity(Instrumentation.java:1705)
>         at android.app.Activity.startActivityForResult(Activity.java:5094)
>         at androidx.fragment.app.FragmentActivity.startActivityForResult(FragmentActivity.java:676)
>         at androidx.core.app.ActivityCompat.startActivityForResult(ActivityCompat.java:234)
>         at androidx.fragment.app.FragmentActivity.startActivityFromFragment(FragmentActivity.java:791)
>         at androidx.fragment.app.FragmentActivity$HostCallbacks.onStartActivityFromFragment(FragmentActivity.java:933)
>         at androidx.fragment.app.Fragment.startActivity(Fragment.java:1176)
>         at androidx.fragment.app.Fragment.startActivity(Fragment.java:1164)
>         at org.mozilla.fenix.share.ShareFragment$subscribeToActions$1.accept(ShareFragment.kt:131)
>         at org.mozilla.fenix.share.ShareFragment$subscribeToActions$1.accept(ShareFragment.kt:34)
[...]

PR #5083, opened 2019-09-03
For #5053: Removed google maps as intent handler for clipboard copy

mcarare:
Filtered provider list removing intent handler by Google Maps that causes crash.

`SendTextToClipboardActivity` was added as intent handler in Google Maps and it looks exactly like the system's _copy to clipboard_ (same title and icon) but it raises a `SecurityException`. Probably a bug that will be fixed on Google's side in the future.

### Pull Request checklist

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/share/ShareFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/share/ShareFragment.kt
@@ -14,6 +14,7 @@ import android.os.Parcelable
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
+import androidx.annotation.VisibleForTesting
 import androidx.appcompat.app.AppCompatDialogFragment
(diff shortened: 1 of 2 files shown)


ekager:
If we look at https://github.com/mozilla-mobile/fenix/issues/3371 this is also happening with other apps, so we may just want to catch the exception and show a toast that the item couldn't be shared. We probably shouldn't filter by the exact apps because it may include a lot of different apps.

mcarare:
You have a point, but, as a user, I would prefer not to have options on the share sheet that will never work. Especially one that masquerades as something else that always works. The average user will think the problem is with Fenix, not something else.
Maybe we could do both? Filtering known issues and having a try-catch as a backup?

mcarare:
As a side-note, the system share filters that option, it does not show two copy to clipboard items.

mcarare:
Closing, as the issue was resolved through PR #5210

Issue #4877, opened 2019-08-22
[Bug] Saving a page to a Collection, doesn't remove the page from opened tabs


softvision-miralobontiu:
## Steps to reproduce
1. Open a few tabs.
2. Save one of them into a collection.

### Expected behavior
The page saved to a collection is automatically closed from open tabs.

### Actual behavior
The saved page is not closed.

Is it expected or not for the saved to a collection page to be dismissed.

### Device information
* Android device: Samsung Galaxy Note 8 (Android 9), Nexus 9 (Android 7.1.1), and HTC Desire 820 (Android 6.0.1)
* Fenix version: Nightly 190821 (#12330640), and debug build 1.0.1934 (a-c 10.0.0, e94e6cc8e, 70.0a1-20190821095414 gecko)

thewhitetulip:
> The page saved to a collection is automatically closed from open tabs.

Why should the page be closed? When I bookmark a tab I don't expect it to close the tab do I? Also in yhe latest stable version my issue is that the tab is closed when I add it to collections.

softvision-miralobontiu:
I filed this issue because I don't really know the expected result: in an earlier build, the page saved to a collection closed, and in the 8/21 build didn't close anymore.
What is expected?

thewhitetulip:
I hope that the default is not close.

AndiAJ:
Hi, reproducible on **Fenix 2.0.0-rc.1** & **Nightly Build #12460618 from 3/9** using a google Pixel 3a (Android 9)

thewhitetulip:
@mcarare Is this the expected result? A tab closed when it is added to collections? We don't have the same logic for bookmarks right?

mcarare:
Expected result:
- Close tab when in home screen
- Keep tab when in browser screen

sv-ohorvath:
Verified as fixed on Nightly 9/16.

PR #5086, opened 2019-09-03
For #4877: Restored logic for closing tabs when not on browserFragment 

mcarare:
Added return in function when saving collections coming from browser fragment

### Pull Request checklist

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -447,6 +447,8 @@ dependencies {
 
     androidTestImplementation Deps.androidx_test_core
     androidTestImplementation Deps.espresso_idling_resources
+    debugImplementation Deps.androidx_fragment_testing
+
 
     androidTestImplementation Deps.tools_test_runner
     androidTestImplementation Deps.tools_test_rules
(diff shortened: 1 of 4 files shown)


buildSrc/src/main/java/Dependencies.kt
```diff
@@ -64,6 +64,7 @@ object Versions {
     const val robolectric = "4.2"
 
     const val google_ads_id_version = "16.0.0"
+    const val androidx_fragment_testing_version = "1.2.0-alpha02"
```
jonalmeida: Shouldn't we be using `androidx_testing` from above instead?

mcarare: That refers to a different artifact

app/build.gradle
```diff
@@ -447,6 +447,8 @@ dependencies {
 
     androidTestImplementation Deps.androidx_test_core
     androidTestImplementation Deps.espresso_idling_resources
+    debugImplementation Deps.androidx_fragment_testing
```
ekager: Why are we using `debugImplementation` here? Does `testImplementation` work?

mcarare: No,  `testImplementation` does not work, currently.`FragmentScenario` runs in an empty activity that houses the tested fragment. If we declare it test APK the Activity will run in a different process than the one the tested fragment needs. Even in the sample offered on developer.android.com `debugImplementation` is used. IMO this limitation will be fixed in future versions of `fragment-testing`.

ekager: Okay thanks for the explanation!

PR #5087, opened 2019-09-03
Fix #4856 - Rebind FindInPageFeature in onStart()

### LINKED ISSUE #4856: [Bug]After switching back to Fenix the "Find in page" bar can't be dismissed
## Steps to reproduce
1. Navigate to any website
2. From the lower right 3dot button select "Find in page"
3. Switch to the Gmail app
4. Switch back to Fenix
5. Tap the 'x' (close) button from the "Find in page"

### Expected behavior
After tapping the  'x' (close) the "Find in page" bar should be dismissed  

### Actual behavior
After tapping the  'x' (close) the "Find in page" bar isn't dismissed

### Device information

* Android device: 
• Google Pixel 3a (Android 9)
• Sony Xperia Z2 tablet (Android 6.0.1)

* Fenix version: Nightly Build #12330640 from 21/8

### Notes
► Video from Google Pixel 3a

Mugurell:
FindInPageFeature is used inside the app as a LifecycleAwareFeature and as such
it receives the onStart / onStop lifecycle calls.
The onStart() lifecycle call would not get passed to the feature but in
onStop() FindInPageFeature will detach it's Presenter and Interactor so when
the user comes back to the screen she could not interact anymore with the
feature.
To fix this we'll rebind it's Presenter and Interactor in onStart().

When working on this I saw the FindInPageBarBehavior is unused so I made
a second commit for code cleanup.

### Pull Request checklist

- [x] **Quality**: This PR builds and passes detekt/ktlint checks.
- [x] **Tests**: This PR does not includes tests as it comes with only small modifications for not yet tested components.
- [x] **Accessibility**: The code in this PR or does not include any user facing features.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/InflationAwareFeature.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/InflationAwareFeature.kt
@@ -49,7 +49,7 @@ abstract class InflationAwareFeature(
      * when the view is inflated.
      */
     override fun start() {
-        // We don't do anything because we only want to start the feature when it's being used.
+        feature?.start()
     }
 
     override fun stop() {
--- a/app/src/main/java/org/mozilla/fenix/components/FindInPageIntegration.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/FindInPageIntegration.kt
@@ -4,17 +4,13 @@
 
 package org.mozilla.fenix.components
 
-import android.content.Context
-import android.util.AttributeSet
 import android.view.View
 import android.view.ViewStub
-import androidx.coordinatorlayout.widget.CoordinatorLayout
 import mozilla.components.browser.session.SessionManager
 import mozilla.components.browser.session.runWithSessionIdOrSelected
 import mozilla.components.browser.toolbar.BrowserToolbar
 import mozilla.components.concept.engine.EngineView
 import mozilla.components.feature.findinpage.FindInPageFeature
-import mozilla.components.feature.findinpage.view.FindInPageBar
 import mozilla.components.feature.findinpage.view.FindInPageView
 import mozilla.components.support.base.feature.LifecycleAwareFeature
 import org.mozilla.fenix.test.Mockable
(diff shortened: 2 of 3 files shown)


app/src/main/java/org/mozilla/fenix/components/InflationAwareFeature.kt
```diff
@@ -49,7 +50,11 @@ abstract class InflationAwareFeature(
      * when the view is inflated.
      */
     override fun start() {
-        // We don't do anything because we only want to start the feature when it's being used.
+        // We usually don't do anything because we only want to start the feature when it's being used.
+
+        // onStop() would basically leak FindInPageFeature which would remain displayed on screen
+        // but without registering any user interaction anymore.
+        (feature as? FindInPageFeature)?.start()
```
jonalmeida: Any reason we can't just do `feature?.start()`?

The `InflationAwareFeature` is abstract of any specific feature, so this would have issues with other features that will extend `InflationAwareFeature` as well.

app/src/main/java/org/mozilla/fenix/components/InflationAwareFeature.kt
```diff
@@ -49,7 +50,11 @@ abstract class InflationAwareFeature(
      * when the view is inflated.
      */
     override fun start() {
-        // We don't do anything because we only want to start the feature when it's being used.
+        // We usually don't do anything because we only want to start the feature when it's being used.
```
jonalmeida: We can remove this comment since it's no longer relevant.

jonalmeida:
Reminds me that I should probably look into moving the `InflationAwareFeature` to components! :)

app/src/main/java/org/mozilla/fenix/components/InflationAwareFeature.kt
pocmo: 👍 I agree. Looking at this code `feature?.start()` seems to be the right thing to do here. Is this used by other features?

jonalmeida: > Is this used by other features?

Not yet, I think @colintheshots is looking into more UI features that can use more lazy view inflations.

pocmo: For some (like find in page) we probably should just make the feature work with a `ViewStub` instead of a `View`. That may be less complicated (code-wise) than wrapping the feature.

jonalmeida:
@Mugurell looks like TaskCluster is failing on a unit test that needs to be fixed: https://tools.taskcluster.net/groups/bz3PN_oRQuO-GePzKkjraA/tasks/WuaSaFxITTqEI-nZi7Avhg/runs/0/logs/public%2Flogs%2Flive.log

jonalmeida:
>  0% of diff hit (target 8.92%)

Codecov seems wrong, since you fixed the test that adds the check, albeit not the nullable part, but it shouldn't say 0%.

Issue #964, opened 2019-03-12
Add ability to disable all suggestion providers


sblatz:
### Why/User Benefit/User Problem
I may not want to expose my clipboard, nor search my history (which tends to be rather crowded). It'd be great if we had a toggle to just disable all (or on a per provider basis) the suggestion providers of AwesomeBar.

### Acceptance Criteria (how do I know when I’m done?)
Users have the ability to toggle on/off which suggestion providers they want to use.

**InVision:** https://mozilla.invisionapp.com/share/M4SBYCLJPQG#/365634921_Settings_-_Search_Alt

sblatz:
_Note: this would be pretty small engineering effort due to the way these are implemented_

kbrosnan:
We have largely done this with the options in the search suggestion settings. Is there any other part that is blocking this @sblatz?

sblatz:
@kbrosnan this is mainly a UX/product decision on if or how we want to do this. It'd be easy to add toggles for each type of suggestion in the search settings menu and allow users to enable as they'd prefer.

vesta0:
I suggest creating a separate toggle for each type of suggestion (history, bookmarks)

We already have a toggle to turn off search provider suggestions.

lime124:
removing the ux label - this is on our radar already as it's part of Q3 plans.

apbitner:
@vesta0 @sblatz InVision link added to AC above, please let me know if you have any feedback.

vesta0:
Thanks @apbitner :)

@jpfaller would you please review/approve the copy?

sblatz:
@apbitner Are we able to add a toggle for hiding clipboard suggestions as well? This was a big reason for the creation of this ticket and was frequently requested by users.

We'll want copy for this too. @jpfaller.

apbitner:
Ah, I was going off of @vesta0 's comment on Jul 2 that mentioned history and bookmarks. Let's add a third toggle then, pending copy.

jpfaller:
Show clipboard suggestions

sv-ohorvath:
Verified on Nightly 9/9. Devices: Google Pixel 3 (Android 9), HTC Desire 820 (Android 6.0.1)

PR #5096, opened 2019-09-03
For #964: Adds ability to disable each suggestion provider

sblatz:
### Pull Request checklist

image

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/search/SearchFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/search/SearchFragment.kt
@@ -91,7 +91,7 @@ class SearchFragment : Fragment(), BackHandler {
                     showShortcutEnginePicker = displayShortcutEnginePicker,
                     searchEngineSource = currentSearchEngine,
                     defaultEngineSource = currentSearchEngine,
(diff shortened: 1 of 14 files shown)


app/src/main/java/org/mozilla/fenix/search/SearchFragment.kt
```diff
@@ -91,8 +91,10 @@ class SearchFragment : Fragment(), BackHandler {
                     showShortcutEnginePicker = displayShortcutEnginePicker,
                     searchEngineSource = currentSearchEngine,
                     defaultEngineSource = currentSearchEngine,
-                    showSuggestions = Settings.getInstance(requireContext()).showSearchSuggestions,
-                    showVisitedSitesBookmarks = Settings.getInstance(requireContext()).shouldShowVisitedSitesBookmarks,
+                    showSuggestions = Settings.getInstance(requireContext()).shouldShowSearchSuggestions,
```
jyeontaek: Looks good! Just a minor nit, I think we should take this opportunity to rename `showSuggestions` to `showSearchSuggestions` just to make it a bit clearer.

sblatz: Great point--in fact the kdoc comment above it even specifies it's a `search suggestion` not just a `suggestion`. Updated 😄

PR #5101, opened 2019-09-04
Automatically generate metrics docs using the Glean SDK

Dexterp37:
Create the docs from the Glean SDK metrics and ping files.

The current problem with this is that the file is generated in `app/docs/metrics.md`, but the existing file lives in `docs/metrics.md`.

Potential solutions:

- we change the glean parser to allow products to specify a custom target output directory
- we keep the file in the current location

### Pull Request checklist


Dexterp37:
This is now ready for review.

Dexterp37:
As a follow-up to this PR it might be useful for the Fenix team to change the current `telemetry.md` file to list the dependencies collecting metrics and link to their autogenerated docs.

sblatz:
@liuche who do we need to reach out to to ensure that this is sufficient telemetry documentation?

liuche:
Good question! I will follow up on this.

This looks like quite a large diff, so I'm going to go through it and spot-check it, and also get confirmation that this an acceptable format. Mostly looking for details on the extras, making sure those are all there.

Dexterp37:
> 
> 
> Good question! I will follow up on this. I believe the Glean team has already checked, but I will make sure the final output fits the requirements.

There is an email thread (you should be CC'd on it @liuche ) in which we asked Michael Feldman about the requirements. We should be fine as long as the format is similar to what Fenix is currently using. For additional safety, we can ask Michael to review the generated file for Fenix. @liuche , are you taking care of that?

> This looks like quite a large diff, so I'm going to go through it and heavily spot-check it, and also get confirmation that this an acceptable format. Mostly looking for details on the extras, making sure those are all there.

I can already tell you that static labels for `labeled_` metric types are not there yet ([bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1578345)). Once you merge this, changes will be automatically picked up next time somebody does a PR on this repo (after we fix the bug mentioned above). However, Fenix does not use [labeled stuff](https://github.com/mozilla-mobile/fenix/blob/master/app/metrics.yaml) with static labels (search stuff uses dynamic labels).

> @Dexterp37 this doesn't handle Leanplum, right? and just ignores those.

No, this is Glean only. Any non-Glean documentation should be moved to a separate file. Alternatively, you could add some notice in the metrics description in `metrics.yaml` saying that "this is also sent via Leanplum". That way it will be automatically present in the generated docs. Given that there's only an handful of Leanplum metrics, I think it's about the same.

**Note**: this PR is aiming at a moving target. Every time a metric is added, this PR needs to be rebased in order to generate up-to-date docs :)

PR #1440, opened 2019-04-06
For #469: remove "Your" from "Your library" string

cadeyrn:
(see https://github.com/mozilla-mobile/fenix/issues/469#issuecomment-480400794)

Issue #469, opened 2019-02-13
"Your Library" menu item:- move to bottom of list & change name to "My Library"


Cheap-Skate:
IIUC Fenix is targeting one-hand use with the toolbar at the bottom
The "Your Library" menu is likely to be used very often.
It could be near the bottom of the screen not the top to be easier to reach one handed
It's also kind of about navigation so it maybe makes sense for it to be near the navigation forward/back/reload buttons

Also, who is "You" in the menu title, should it be "**My** Library" not "**Your** Library"?   ;)

Galaxy S7
Android 8.0.0

topotropic:
@Cheap-Skate makes sense, we'll explore thumb-friendlier options for the library – and good catch with the menu title. Thanks!

topotropic:
sorry, closed it accidentely

cadeyrn:
> Also, who is "You" in the menu title, should it be "My Library" not "Your Library"? ;)

Changing that would result in inconsistencies across the product.

In "About Fenix" you can read "Fenix puts _you_ in control", not "Fenix puts _me_ in control". In the home screen of a private session you can read "_You're_ in a private session" and "Fenix clears _your_ search and browsing history", not "_I'm_ in a private session" and "Fenix clears _my_ search and browsing history" and so on. It doesn't make sense to change the string "Your Library" to "My Library" and not changing all the other strings. To be honest if would be more confusing if we mix "I" and "You" across the product.

topotropic:
Ah, sorry, I was more thinking of removing the pronoun altogether. So just going with "Library" – but don't have strong opinions on it. @mheubusch maybe you can chime in here. Thanks!

Cheap-Skate:
Yes I see that there tends to be "you" not "me" in Fennec & Desktop. Fine. 

But I would suggest that **Your** Library is unnecessary, Fennec & Desktop just have History and Top Sites not **Your** History etc. And if Fenix is going down the Your route I would expect **Your** Favorites, **Your** Sessions **Your** Downloads etc which we don't have ?

So my suggestion would be get rid of **Your** as @topotropic suggested.

cadeyrn:
Please note that there is a new mockup for Firefox for iOS with a "Your Library" label: https://mozilla.invisionapp.com/share/NVQNKFF2C7Z#/screens/348322247

Whatever you decide ("Your Library" or "Library"), maybe you could synchronisize the decision with the iOS team so that both products use the same wording. :)

[...]

PR #5102, opened 2019-09-04
Import strings from android-l10n

Pike:
State: mozilla-l10n/android-l10n@6d2fe8331a703e7f46050b93ee4db4812af36095

### DIFF
--- a/app/src/main/res/values-de/strings.xml
+++ b/app/src/main/res/values-de/strings.xml
@@ -15,7 +15,7 @@
     <!-- No Open Tabs Message Header -->
     <string name="no_open_tabs_header">Keine Tabs geöffnet</string>
     <!-- No Open Tabs Message Description -->
-    <string name="no_open_tabs_description">Ihre offenen Tabs werden hier angezeigt.</string>
+    <string name="no_open_tabs_description">Offene Tabs werden hier angezeigt.</string>
 
     <!-- Private Browsing -->
     <!-- Title for private session option -->
@@ -42,7 +42,7 @@
     <!-- Browser menu button that opens the settings menu -->
     <string name="browser_menu_settings">Einstellungen</string>
     <!-- Browser menu button that opens a user's library -->
-    <string name="browser_menu_your_library">Ihre Bibliothek</string>
+    <string name="browser_menu_your_library">Bibliothek</string>
     <!-- Browser menu toggle that requests a desktop site -->
     <string name="browser_menu_desktop_site">Desktop-Website</string>
     <!-- Browser menu button that opens the find in page menu -->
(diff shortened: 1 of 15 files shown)


app/src/main/res/values-de/strings.xml
```diff
@@ -42,7 +42,7 @@
     <!-- Browser menu button that opens the settings menu -->
     <string name="browser_menu_settings">Einstellungen</string>
     <!-- Browser menu button that opens a user's library -->
-    <string name="browser_menu_your_library">Ihre Bibliothek</string>
+    <string name="browser_menu_your_library">Bibliothek</string>
```
cadeyrn: Why was the "Your" from" Your Library" removed in the German translation? It was an explicit decision a few months ago that this menu item should be "Your Library" and not only "Library". And it's still "Your Library" in the original string.

pocmo: CC @Pike @Delphine

Pike: You can see on https://pontoon.mozilla.org/de/android-l10n/mozilla-mobile/fenix/app/src/main/res/values/strings.xml/?search=browser_menu_your_library&string=198474 that this change was made by @Archaeopteryx

Archaeopteryx: Whose else library shall it be? It's also not addressed as "Your library" but as 'Library' in the view it opens and similar concepts like 'Collections' don't use 'Your'. Avoiding 'Your' and similar words caters to both audiences preferring to be addressed formally and casually.

cadeyrn: Instead of just changing a string only for one language should be discussed with UX so that every language does the same.

Please also note that it was already discussed in the past. In fact I wrote a PR to change "Your Library" to "Library" a few months ago [1] but it was reverted because UX wanted the "Your Library" [2] string, so it was an explicit choice *not* to use "Library". There is also an explanation why UX made this decision.

Also note that Firefox for iOS also uses "Your Library" and both products should use the same wording. Before my string change was reverted I also asked the Firefox for iOS team to change the string for the sake of consistency but they don't want to change "Your Library" to "Library" either [3].

[1] https://github.com/mozilla-mobile/fenix/pull/1440
[2] https://github.com/mozilla-mobile/fenix/issues/469#issuecomment-485878998
[3] https://github.com/mozilla-mobile/firefox-ios/issues/4814

cadeyrn: @Pike Can you change it back? I can't suggest the change on Pontoon because it's identical to an older translation.

Delphine: I've talked with Michelle from UX, and it is acceptable for a localizer to drop a pronoun and adapt tone depending on product and country. However it's important to remain consistent throughout products.
@mheubusch can elaborate more on this if needed.

Issue #4908, opened 2019-08-23
[Telemetry] Send Tab


vesta0:
#### Description & Product Manager / Data Scientist User Story
- As a product owner, I want to know how users interact with Send Tab and how useful it is. 

#### What questions will you answer with this data?
- How often users send a tab to their other connected devices
- How often users sign up for an account or sign into their account in order to send a tab.

#### Why does Mozilla need to answer these questions? Are there benefits for users? Do we need this information to address product or business requirements?
-To assess usefulness of this feature and identify improvements. 

#### What probes (suggested, if applicable)
- Tap on any FxA connected device
- Tap on "sign in to send a tab"

### Acceptance Criteria (Added by PM)
- Event pings can be queried via re:dash
- We are sending telemetry events for the actions listed in the requirements
- We have documented the telemetry
- We have asked a data steward to [review](https://github.com/mozilla/data-review/blob/master/request.md) the telemetry
- NOT an AC: Data science to create dashboard or further graphs (this will be a separate issue, this issue is only about hooking up the events described and that they can be queried in re-dash)

AndiAJ:
Hi, **verified as fixed** on Fenix 2.0.0-rc.2 using a Samsung Galaxy Tab A (Android 9).
Sent tabs using the following cases:

1. From the QAB  - Ping **eba96ce1-b0c4-49af-82e0-3ff522955058** ✔️
2. From the Home screen open tabs view (multiple tabs)- Ping **2480f8c1-0320-4151-b589-c302441169e2** ✔️
3. From the page menu - Ping **c7b2cc6f-cf92-45d3-a49d-c714dc7bb718** ✔️
4. From History - Ping **cbfe1b15-c8d1-49b9-a43f-75145e4c7cb8** ✔️
5. From History multi-select - Ping **56394b9a-c5fd-4af0-8dca-9214230c25c0** ✔️
6. From Bookmarks - Ping **be1623c-3292-43a4-a8ca-814faf16cdeb** ✔️
7. From Collections - Ping **2108e414-8fcc-4a2f-8fe2-fde371f7776f** ✔️

[Logcat](https://github.com/mozilla-mobile/fenix/files/3609153/send.txt)
[Glean dashboard](https://debug-ping-preview.firebaseapp.com/pings/aj-130902)

@sblatz I would kindly ask you to check and in case I've missed something re-open the issue. :)

PR #5108, opened 2019-09-04
For #4908: Adds telemetry for send tab

sblatz:
### Pull Request checklist

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/share/ShareFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/share/ShareFragment.kt
@@ -86,6 +86,7 @@ class ShareFragment : AppCompatDialogFragment() {
 
         shareInteractor = ShareInteractor(
             DefaultShareController(
+                context = requireContext(),
                 fragment = this,
                 sharedTabs = tabs,
                 navController = findNavController(),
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/mozilla/fenix/share/ShareFragment.kt
```diff
@@ -86,6 +86,7 @@ class ShareFragment : AppCompatDialogFragment() {
 
         shareInteractor = ShareInteractor(
             DefaultShareController(
+                context = context!!,
```
jonalmeida: nit: `requireContext()` so we get better logs if this fails.

sblatz: @jonalmeida what better logs does `requireContext()` give vs. `context!!`? Just curious because I didn't know they were different!

jonalmeida: @sblatz requireContext, explicitly [throws an exception](http://androidxref.com/9.0.0_r3/xref/frameworks/support/fragment/src/main/java/androidx/fragment/app/Fragment.java#680) which gives better stack traces than `KotlinNullPointException` (imo).

Issue #419, opened 2019-02-08
Launch icon for private browsing


vesta0:
### Why/User Benefit/User Problem
As a user, I want to be able to open the browser in private mode for my quick private searches where no history is retained. 

### What / Requirements
Build a solution for a 1-click private mode launch

### Acceptance Criteria (how do I know when I’m done?)
User can launch a search in private mode without opening the regular browser first.

brampitoyo:
I just tested Facebook version 1.9.0 ([Download APK](https://appchive.net/apk/com.facebook.katana/facebook/14776/download)). It was originally built around the time of Android version 4 (Ice Cream Sandwich and Jelly Beans), but can still be installed in Android Pie. It has lots of warnings, but many functionalities were surprisingly intact.

This version of Facebook was special, because it adds three launcher icons for three separate activities: Facebook, Camera and Messenger.

<img width="360" src="https://user-images.githubusercontent.com/51007/52980372-9ee5fc80-343e-11e9-986c-15830468144a.png">

## Pre sign in

Tapping all icons would bring you to the same sign in screen:

<img width="360" src="https://user-images.githubusercontent.com/51007/52980781-baea9d80-3440-11e9-97d3-2169a0d575a5.png">

## Post sign in

* Tapping Facebook would bring you to your timeline on Facebook.app
* Tapping Messenger would bring you to the “Messages” tab on Facebook.app
* Tapping Camera would bring you to a camera view (?) that behaves as if it was a separate app

### Facebook and Messenger

Both launcher icons share the same behaviour:

* They are shortcuts to two views within the same Facebook.app
* If you go to the app switcher, Android System sees one app

Facebook itself promotes these icons not as separate apps, but as “shortcuts for sharing”.

<img width="360" src="https://user-images.githubusercontent.com/51007/52980644-00f33180-3440-11e9-90af-e258ef5dc917.png">

### Camera

This launcher icon behaves differently:

* Distinct, unique UI from Facebook/Messenger
* There’s no entry point to access Facebook/Messenger from Camera, and vice versa
* If I go to the app switcher, Android System thinks that it’s totally different from Facebook/Messenger

In other words, Camera behaves as if it’s a separate app.

![screenshot_20190219-115744](https://user-images.githubusercontent.com/51007/52980611-e4ef9000-343f-11e9-86b5-47599e6a7659.png)

brampitoyo:
### Possible next steps

It seems as if we can design private browsing to behave in two ways:

[...]

PR #5109, opened 2019-09-04
For #419: Add strings for "Add private browsing shortcut" feature

jyeontaek:
### Pull Request checklist

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -2,6 +2,9 @@
    - License, v. 2.0. If a copy of the MPL was not distributed with this
    - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
 <resources>
+    <!-- App name for private browsing mode, only the "(Private)" portion should be localized. -->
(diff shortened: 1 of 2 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -2,6 +2,9 @@
    - License, v. 2.0. If a copy of the MPL was not distributed with this
    - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
 <resources>
+    <!-- App name for private browsing mode, only the "(Private)" portion should be localized. -->
+    <string name="app_name_private">Firefox Preview (Private)</string>
```
boek: We should use placeholder text for the app_name like we do here: https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/res/values/strings.xml#L24

jyeontaek: This string is only used in the app manifest file, and I'm not sure how we can make placeholder text work with `AndroidManifest.xml`.

boek: 👍

Issue #3106, opened 2019-06-03
Granular options for clearing user data in Fenix


vesta0:
## User Story
* As a user, I want to have the option of clearing different types of personal data, so I can have more control over what I choose to keep or delete.

Types of personal data:

-Open Tabs
-Browsing history
-Search history
-Cookies and site data
-Cached images and files
-Site settings

### Requirements
* Behind a Feature Flag so QA can spend extra time testing

### Acceptance Criteria
* I can choose what type of user data to delete at any time

UX Spec (strings are done).
https://mozilla.invisionapp.com/share/63TIO0DPXTS

lime124:
removing the ux label - this is on our radar already as it's part of Q3 plans.

liuche:
Also look at #3700 because they seem like they'll be similar UI.

liuche:
QA has said they'd like a little extra time to test this, so this should be ideally behind a feature flag.

liuche:
Still some outstanding work, but should be resolved by 8/19.

colintheshots:
Wait. The UX says "Site Permissions", but the user story says "Site Settings". Those are two, different things. We already have a screen that lets the user delete all site permissions. The new deletion options is Site Settings. I think we need to separate out that part of this story and get new strings to complete it properly. What should the new title and subtitle be? @mheubusch

colintheshots:
I'll complete the story as written and we can update it later.

jpfaller:
Verified the string as written (Site Permissions) is ok with @colintheshots

colintheshots:
This is behind a feature flag in Nightly for now.

andreicristianpetcu:
@colintheshots which flag?

cadeyrn:
> @colintheshots which flag?

https://github.com/mozilla-mobile/fenix/blob/2b9efccfca4ef23a48e19c7934955a0bff32a409/app/src/main/java/org/mozilla/fenix/FeatureFlags.kt#L44

neekless:
Will we eventually get even finer granularity i.e. Delete per site/domain?

AndiAJ:
Hi, all **options work properly** from a functional point of view.

On the other hand, I've found a couple of minor issues on the latest Nightly Build #12590611 from 16/9 using a Google Pixel 3a (Android 9) as it follows:
• The delete browsing data is always enabled
• "Fenix" string placeholder %s is displayed in the confirmation pop-up message.

►Video
[...]

PR #5121, opened 2019-09-05
For #3106: Granular options for clearing user data

colintheshots:
device-2019-09-05-134353

### Pull Request checklist

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/settings/DeleteBrowsingDataController.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/DeleteBrowsingDataController.kt
@@ -10,6 +10,7 @@ import kotlinx.coroutines.delay
 import kotlinx.coroutines.withContext
 import mozilla.components.concept.engine.Engine
 import mozilla.components.feature.tab.collections.TabCollection
+import org.mozilla.fenix.FeatureFlags
 import org.mozilla.fenix.ext.components
 import kotlin.coroutines.CoroutineContext
 
@@ -34,7 +35,11 @@ class DefaultDeleteBrowsingDataController(
 
     override suspend fun deleteBrowsingData() {
         withContext(coroutineContext) {
-            context.components.core.engine.clearData(Engine.BrowsingData.all())
+            if (FeatureFlags.granularDataDeletion) {
+                context.components.core.engine.clearData(Engine.BrowsingData.select(Engine.BrowsingData.DOM_STORAGES))
+            } else {
+                context.components.core.engine.clearData(Engine.BrowsingData.all())
+            }
         }
         context.components.core.historyStorage.deleteEverything()
     }
--- a/app/src/main/java/org/mozilla/fenix/settings/DeleteBrowsingDataController.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/DeleteBrowsingDataController.kt
@@ -1,0 +1,76 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.settings
+
+import android.content.Context
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.withContext
+import mozilla.components.concept.engine.Engine
+import mozilla.components.feature.tab.collections.TabCollection
+import org.mozilla.fenix.ext.components
+import kotlin.coroutines.CoroutineContext
+
+interface DeleteBrowsingDataController {
+    suspend fun deleteTabs()
+    suspend fun deleteBrowsingData()
+    suspend fun deleteCollections(collections: List<TabCollection>)
+    suspend fun deleteCookies()
+    suspend fun deleteCachedFiles()
+    suspend fun deleteSitePermissions()
+}
+
+class DefaultDeleteBrowsingDataController(
+    val context: Context,
+    val coroutineContext: CoroutineContext = Dispatchers.Main
+) : DeleteBrowsingDataController {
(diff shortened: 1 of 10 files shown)


colintheshots:
For UX, is it okay that cookies and cached files are always selectable and not disabled? We don't have any means to count them currently.

ekager:
Tagging @AmyYLee for confirmation, but I think we decided in design reviews that not having the counts and always selectable is fine for now

app/src/main/java/org/mozilla/fenix/settings/DeleteBrowsingDataController.kt
```diff
@@ -0,0 +1,76 @@
...
+    val context: Context,
+    val coroutineContext: CoroutineContext = Dispatchers.Main
+) : DeleteBrowsingDataController {
+    override suspend fun deleteTabs() {
+        withContext(coroutineContext) {
+            context.components.useCases.tabsUseCases.removeAllTabs.invoke()
+        }
+    }
+
+    override suspend fun deleteBrowsingData() {
```
ekager: My understanding was that we wanted to update this to only delete history here and separate out the other categories. If we have this one clear `Engine.BrowsingData.all()` then the categories are kind of obsolete

ekager: But we should probably maintain the old behavior in an if/else via the feature flag

colintheshots:
@ekager I have confirmation from @AmyYLee about the expected behavior. We should stop disabling checkboxes and also select all by default if the feature flag is enabled. The PR is updated.

app/src/main/java/org/mozilla/fenix/settings/DeleteBrowsingDataController.kt
```diff
@@ -0,0 +1,81 @@
...
+        withContext(coroutineContext) {
+            if (FeatureFlags.granularDataDeletion) {
+                context.components.core.engine.clearData(Engine.BrowsingData.select(Engine.BrowsingData.DOM_STORAGES))
```
ekager: Where did you get that DOM storages should be part of this one? I'm not sure I saw this in my mocks for delete on quit. We should just make sure what we're clearing is consistent

colintheshots: It seems like if we want everything selected to delete everything, we don't want to miss something. Where do DOM storages belong on this UI? With cached files and images for recovering disk space?

app/src/main/java/org/mozilla/fenix/settings/DeleteBrowsingDataController.kt
```diff
@@ -0,0 +1,81 @@
...
+            context.components.core.engine.clearData(
+                Engine.BrowsingData.select(Engine.BrowsingData.ALL_CACHES)
+            )
+        }
+    }
+
+    override suspend fun deleteSitePermissions() {
+        withContext(coroutineContext) {
+            context.components.core.engine.clearData(
+                Engine.BrowsingData.select(Engine.BrowsingData.ALL_SITE_SETTINGS)
```
ekager: I think we should remove this from this category because `ALL_SITE_SETTINGS` includes `PERMISSIONS` (no-op IIUC from the GV team) and then caches which the user can clear separately?

colintheshots: All site settings does not include any caches, as far as I can tell. I'm not sure if it's entirely a no-op either though.

```
            const val COOKIES: Int = 1 shl 0
            const val NETWORK_CACHE: Int = 1 shl 1
            const val IMAGE_CACHE: Int = 1 shl 2
            const val DOM_STORAGES: Int = 1 shl 4
            const val AUTH_SESSIONS: Int = 1 shl 5
            const val PERMISSIONS: Int = 1 shl 6
            const val ALL_CACHES: Int = NETWORK_CACHE + IMAGE_CACHE
            const val ALL_SITE_SETTINGS: Int = (1 shl 7) + PERMISSIONS
            const val ALL_SITE_DATA: Int = (1 shl 8) + COOKIES + DOM_STORAGES + ALL_CACHES + ALL_SITE_SETTINGS
            const val ALL: Int = 1 shl 9
```

app/src/main/java/org/mozilla/fenix/settings/DeleteBrowsingDataController.kt
ekager: I agree with you that it looks like `DOM_STORAGES` and `AUTH_SESSIONS` now belong to none of these categories. We should probably verify with product and UX where these should be cleared and then keep it consistent between hte two features?

app/src/main/java/org/mozilla/fenix/settings/DeleteBrowsingDataController.kt
ekager: Ah okay I missed that `ALL_SITE_SETTINGS` didn't include `ALL_CACHES`. I guess if it's just clearing `PERMISSIONS` that may or may not do anything in addition to using `permissionStorage` this could be fine

app/src/main/java/org/mozilla/fenix/settings/DeleteBrowsingDataController.kt
colintheshots: I believe AUTH_SESSIONS really belongs in its own category, but DOM_STORAGES are a type of history or cache, depending upon how you look at it. The reason I included it here is because the delete cache checkbox mentions that its intent is to reduce storage usage. Deleting history and DOM storage offers more of a privacy improvement than storage savings.

colintheshots: I move AUTH_SESSIONS to the cookies category based on feedback from the GV team. I think DOM_STORAGES belongs here based upon the descriptions we're offering to users.

PR #5127, opened 2019-09-05
Remove ThumbnailsFeature.

pocmo:
We currently do not use thumbnails anywhere in the app. Not using the feature means we are
not taking thumbnails on every page load which means we are saving memory and CPU cycles.


pocmo:
CC @fluffyemily As you said this was really unused and can be removed for now. :)

Issue #2897, opened 2019-05-27
FNX3-16660 ⁃ Add action states to share sheet


jonalmeida:
As a user, I want to be able to initiate the following actions from the share sheet:

* "Verify your email address" - if a user has signed up for an account but hasn't verified their email address during that process before they can make use of sync.
* "Connect another device" - user has signed into their account but doesn't have other devices syncing to send a tab to.
* "Reconnect to Sync" - error occurred within user account and user needs to re-login.
* "Offline" - No network connection.

Mocks: https://mozilla.invisionapp.com/share/QSR2F8UD4PN#/screens/374762782

Acceptance criteria:
-If I have signed up for an account but haven't yet verified my email address, I will be see a "Verify your email address" icon when I open the share sheet. 
-If I have signed into my account but don't have other devices syncing, I will see a "Connect another device" icon when I open the share sheet. 
~~If there is an error within my account sync, I will be notified that I need to re~~login by seeing a "Reconnect to Sync" icon when I open the share sheet. 
-If I don't have network connection and try to share something, I will see the "Offline" icon when I open the share sheet.

jonalmeida:
Related to https://github.com/mozilla-mobile/fenix/issues/2897 and depends on the same fixes.

jonalmeida:
Adding a-c label since this may involve some a-c changes as well before we can consume the a-s bits.

jonalmeida:
@vrudesai I've changed the feature label since this this isn't related to the QAB directly.

@vesta0 if we want to do this for Post-MVP, there is still a small amount of work to hide the send-tab section if we encounter an error. Shall I file another issue for that?

liuche:
We're looking at this during Epic sizing, and @vesta0 could you add some Acceptance Criteria - it's not clear to me what needs to be added.

mheubusch:
@jonalmeida can you describe the errors to me so I can write strings? Do you just need the label under the two states above?

jonalmeida:
@mheubusch I've updated the title to say 'action states' since they aren't all errors.

For these action states there aren't any messages we want to display other than the error title. You can see the user flows here: https://mozilla.invisionapp.com/share/QSR2F8UD4PN#/screens/364885746

The ones for this ticket are:
- "Verify your email address" - if a user has signed up for an account but hasn't verified their email address during that process before they can make use of sync.
- "Connect another device" - we have an account, my no other devices to send to.
[...]

PR #5146, opened 2019-09-06
For #2897 - Add Action States To Share Sheet

ekager:
image
image

Waiting on one UX clarification in #2897 
### Pull Request checklist

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/share/ShareFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/share/ShareFragment.kt
@@ -9,6 +9,9 @@ import android.content.Intent
 import android.content.Intent.ACTION_SEND
 import android.content.Intent.FLAG_ACTIVITY_NEW_TASK
 import android.content.pm.ResolveInfo
+import android.net.ConnectivityManager
+import android.net.Network
+import android.net.NetworkRequest
 import android.os.Bundle
 import android.os.Parcelable
 import android.view.LayoutInflater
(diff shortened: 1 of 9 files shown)


app/src/main/java/org/mozilla/fenix/share/ShareFragment.kt
```diff
@@ -70,6 +80,37 @@ class ShareFragment : AppCompatDialogFragment() {
...
+            lifecycleScope.launch {
+                delay(NETWORK_RESTORED_DELAY)
+                reloadDevices()
+            }
+        }
+    }
+
+    private fun reloadDevices() {
+        context?.let {
+            val list = lifecycleScope.async(Dispatchers.IO) {
```
jonalmeida: I'm not sure I follow why this is done async so it might be helpful to add a comment around it.

Is it for adding the delay to updating the UI, or is it to give the account manager enough time to load new devices? If it's the latter then we may have to do a device refresh instead to make sure we have the most recent devices (i.e. when you come online, the device list may have changed).

ekager: It's the latter. How do I do a device refresh? I thought I was by building a new device list :/

app/src/main/java/org/mozilla/fenix/share/ShareFragment.kt
```diff
@@ -70,6 +79,38 @@ class ShareFragment : AppCompatDialogFragment() {
...
+
+    private fun reloadDevices() {
+        context?.let {
+            val fxaAccountManager = it.components.backgroundServices.accountManager
+            val list = lifecycleScope.async(Dispatchers.IO) {
+                val refreshDevicesAsync =
+                    fxaAccountManager.authenticatedAccount()?.deviceConstellation()
+                        ?.refreshDevicesAsync()
+                refreshDevicesAsync?.await()
+                buildDeviceList(fxaAccountManager)
```
jonalmeida: `buildDeviceList` just constructs a device list for the UI to consume so it doesn't need to be done on the IO dispatcher, but I guess that's fine as well.

Please leave some comments on why we're doing this here though for when someone else comes in to read this.

ekager: Okay took it off the IO and seems to work and be simpler :) thanks!

PR #5158, opened 2019-09-06
For 5092 - Show a Snackbar with retry option if sharing to devices fails

Mugurell:
This is just WIP. Tests are failing.

To allow the user to retry a failed send I've used a GlobalScope'd coroutine but now I don't really know how to test `shareToDevicesWithRetry()`.
Suggestions are very much welcomed.

----
(Edit by @sblatz) 
error snack 2019-09-25 09_43_54

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/share/ShareController.kt
+++ b/app/src/main/java/org/mozilla/fenix/share/ShareController.kt
@@ -10,14 +10,18 @@ import android.content.Intent.ACTION_SEND
 import android.content.Intent.EXTRA_TEXT
 import android.content.Intent.FLAG_ACTIVITY_NEW_TASK
 import androidx.annotation.VisibleForTesting
-import androidx.fragment.app.Fragment
 import androidx.navigation.NavController
 import com.google.android.material.snackbar.Snackbar
+import kotlinx.coroutines.Deferred
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.launch
 import mozilla.components.concept.sync.Device
 import mozilla.components.concept.sync.TabData
 import mozilla.components.feature.sendtab.SendTabUseCases
 import org.mozilla.fenix.R
 import org.mozilla.fenix.components.FenixSnackbar
+import org.mozilla.fenix.components.FenixSnackbarPresenter
 import org.mozilla.fenix.components.metrics.Event
 import org.mozilla.fenix.ext.getRootView
 import org.mozilla.fenix.ext.metrics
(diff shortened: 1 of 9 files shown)


app/src/main/java/org/mozilla/fenix/share/ShareController.kt
```diff
@@ -73,15 +83,15 @@ class DefaultShareController(
     }
 
     override fun handleShareToDevice(device: Device) {
-        sendTabUseCases.sendToDeviceAsync(device.id, sharedTabs.toTabData())
-        (fragment.activity as ShareFragment.TabsSharedCallback).onTabsShared(sharedTabs.size)
-        dismiss()
+        with(fragment.activity!!.applicationContext) {
```
jonalmeida: Maybe it's possible to pass in the context of the activity to the controller and use that?

The `!!` assertion has a tendency to bite back with Android lifetimes. I don't believe you'd need the application lifetime to send the tab either.

jonalmeida: Having just reviewed https://github.com/mozilla-mobile/fenix/pull/5108/files, you should have access to the context from there when it merges since it adds in that change.

sblatz:
@Mugurell Are you still working on this? :)

app/src/main/java/org/mozilla/fenix/share/ShareController.kt
Mugurell: Thanks!

Mugurell:
> @Mugurell Are you still working on this? :)

Thanks for pinging me on this!
It should've been ready for review but I forgot to resolve the above discussion with jonalmeida.
Although the `GlobalScope.launch` block is not tested it's parent method is and so are the rest of the methods in `DefaultShareController`.
Just rebased everything on master, should be ready for review and an eventual merge.

sblatz:
Just fixing the linting errors, then will review :)

sblatz:
I'm struggling to test this. I cannot get anything besides an "offline" button when i throw my phone in airplane mode in the share menu. Even when I do it quickly and try to press before the devices disappear it doesn't work. 

@Mugurell / @jonalmeida any ideas on how to test? If it's this hard to reproduce maybe we don't need this? I guess internet could cut out mid-sharing but I don't know how to replicate that environment. 

image

Mugurell:
@sblatz Initially getting this triggered was indeed easy with just turning off internet access after the list of devices was already built.
With the changes from #2897 which now automatically reloads the list of devices whenever there's a connectivity change this scenario does not help in triggering that retry snackbar anymore.

I feel this patch is still useful for other possible scenarios and has value as just a sanity check, a failsafe for maybe other things that could go wrong.

When I recently rebased this, to be able to easily test I just commented in code the newly added [`reload devices method`](https://github.com/mozilla-mobile/fenix/blob/464122fda1ab7255a562472e79fda859788a2c5b/app/src/main/java/org/mozilla/fenix/share/ShareFragment.kt#L84) which allowed me to test again after loosing Internet connection.

sblatz:
@Mugurell Ah, good idea! I will test this once FxA is back up (having some downtime issues today)

sblatz:
I have made a small change to this PR to allow coloring of the fenix snackbar. I'm going to have @ekager review my commit :)

The rest of this work looks great @Mugurell and we will land right after we cut our 2.1 release since this has a new string.

Issue #1114, opened 2019-03-21
FNX-4807 ⁃ Show which tab is playing audio


pocmo:
### User Story
As a user I want to know which tab is playing audio and be able to pause each tab, so I know where audio is coming from specially when I have multiple tabs open.

### Acceptance Criteria:
-I can see which tab is playing audio
-I can easily pause audio for each tab

vesta0:
@topotropic just a note to think about accessibility of the playback controls (e.g. too close to close tab x)

topotropic:
**inactive (play)** 
- light theme: disabled_light_theme  #6620123A
- dark theme: disabled_dark_theme  #66FBFBFE
- private theme: disabled_private_theme  #66FBFBFE

**active (pause)** 
- light theme: accent_bright_light_theme  (#592ACB)
- dark theme: accent_bright_dark_theme  #592ACB
- private theme: accent_bright_private_theme  #592ACB

[1114-home-tile-with-play-button.pdf](https://github.com/mozilla-mobile/fenix/files/3519303/1114-home-tile-with-play-button.pdf)
[icons.zip](https://github.com/mozilla-mobile/fenix/files/3519237/icons.zip)

liuche:
Questions:
- is this too busy?
- Verify that we have the API to determine which tab is playing.

pocmo:
AC exposed that state already. I filed the linked issue above to add some more convenient helpers to examine that state.

topotropic:
Happy to discuss the design, please ping me!

sblatz:
@vesta0 will we want telemetry for engaging with this feature?

sblatz:
@topotropic are you able to link me to the invision or design file for this? I'd like to see the margins between some of the other views (close button and play button, etc.)

vesta0:
@sblatz I am guessing we can add telemetry for pause/play interaction? Anything else you recommend? I can file a telemetry issue for this + pause/play interaction on the notification.

sblatz:
@vesta0 that sounds sufficient to me!

topotropic:
@vesta0 can we figure out how many open tabs have audio on pause/play simultaneously with telemetry on pause/play interactions?

vesta0:
@topotropic that's interesting, I will look into that.

sblatz:
[...]

PR #5165, opened 2019-09-06
For #1114: Show playing tab

sblatz:
Known issues: 
* it can take 250ms to update the icon of the play/pause button: https://github.com/mozilla-mobile/android-components/issues/4353
* selectableItemBackground does not play nicely with the play/pause button since it is above another `imageView`. I believe this may be an Android bug, unfortunately.

### Pull Request checklist

Tests: no tests included as these are all things added to the HomeFragment. It should likely get some UI tests, though, and I will file a ticket for those.

Screenshots: 

image

image

Accessibility: The button itself is only 24x24dp, that being said I increase the tap size to 48x48 so make it in line with the accessibility guidelines. Ive also added a dynamic ContentDescription that updates to play/pause based on the state of the button.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
@@ -48,6 +48,11 @@ import mozilla.components.concept.sync.AccountObserver
 import mozilla.components.concept.sync.AuthType
 import mozilla.components.concept.sync.OAuthAccount
 import mozilla.components.concept.sync.Profile
+import mozilla.components.feature.media.ext.getSession
+import mozilla.components.feature.media.ext.pauseIfPlaying
+import mozilla.components.feature.media.ext.playIfPaused
+import mozilla.components.feature.media.state.MediaState
+import mozilla.components.feature.media.state.MediaStateMachine
 import mozilla.components.feature.tab.collections.TabCollection
 import org.jetbrains.anko.constraint.layout.ConstraintSetBuilder.Side.BOTTOM
 import org.jetbrains.anko.constraint.layout.ConstraintSetBuilder.Side.END
@@ -144,6 +149,7 @@ class HomeFragment : Fragment(), AccountObserver {
         val sessionObserver = BrowserSessionsObserver(sessionManager, singleSessionObserver) {
             emitSessionChanges()
         }
+
         lifecycle.addObserver(sessionObserver)
 
         if (!onboarding.userHasBeenOnboarded()) {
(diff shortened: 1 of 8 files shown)


app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/TabViewHolder.kt
```diff
@@ -76,13 +109,49 @@ class TabViewHolder(
         this.tab = tab
         updateTabUI(tab)
         item_tab.transitionName = "$TAB_ITEM_TRANSITION_NAME${tab.sessionId}"
+
+        containerView?.context!!.components.core.sessionManager.findSessionById(tab.sessionId)?.register(mediaObserver)
```
jonalmeida: In general, `!!` should be avoided like the plague. If an instance is non-null then we know we can safely use it all the time, otherwise use safe calls on it, if you really need it non-null and have reasonable certainty it's safe, try to find a `require` version that throws exceptions that can be better traced out.

Here, the context can be null if the `containerView` is destroyed far sooner than we can get the context from it.

Think of an Android context as 'the lifetime of an android thing', and you can't control when the lifetime of that thing will be destroyed.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -353,6 +357,29 @@ class HomeFragment : Fragment(), AccountObserver {
...
+                        if (media.state == Media.State.PAUSED) {
+                            media.controller.play()
+                        } else {
+                            media.controller.pause()
+                        }
+                    }
+                }
+            }
+            is TabAction.PlayMedia -> {
+                // TODO: This will be replaced with a simple call to `MediaStateMachine.play` once AC exposes
```
sblatz: Will be exposed here: https://github.com/mozilla-mobile/android-components/pull/4351

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -884,13 +905,19 @@ class HomeFragment : Fragment(), AccountObserver {
...
     }
 }
 
+fun MediaState.getSession(): Session? {
+    return when (this) {
+        is MediaState.Playing -> session
+        is MediaState.Paused -> session
+        else -> null
+    }
+}
```
pocmo: This helper should be available through AC now too :)

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlAdapter.kt
```diff
@@ -37,7 +40,28 @@ sealed class AdapterItem(@LayoutRes val viewType: Int) {
...
+        // Tell the adapter exactly what values have changed so it only has to draw those
+        override fun getChangePayload(newItem: AdapterItem): Any? {
+            (newItem as TabItem).let {
+                val shouldUpdateUrl = newItem.tab.url != this.tab.url
+                val shouldUpdateHostname = newItem.tab.hostname != this.tab.hostname
+                val shouldUpdateTitle = newItem.tab.title != this.tab.title
+                val shouldUpdateSelected = newItem.tab.selected != this.tab.selected
+                val shouldUpdateMediaState = newItem.tab.mediaState != this.tab.mediaState
+
+                return AdapterItemDiffCallback.TabChangePayload(
```
ekager: Should we file a followup issue to include the favicon in this payload and update from session listener?

sblatz:
@ekager filed a followup! https://github.com/mozilla-mobile/fenix/issues/5241

Issue #4658, opened 2019-08-09
[Telemetry] Private Mode Features


vrudesai:
Owner: Product Manager Intern

Overall, this issue is for the new private mode features.

All under this meta: [Meta] Open In Private Browsing #4390:

Launcher Icon for PBM: #419
Allow user to always open 3rd party links in PBM + TP: #514 
Persistent notification to close private browsing tabs: #2053 
Delete all Private Tabs Redesign: #2834 

#### Description & Product Manager / Data Scientist User Story
As a product owner intern, I want to know if users find the above features usable and discoverable. The data will also help us prepare for the sunset of Focus and the best way to onboard users onto Fenix.

#### Hypothesis
We believe this feature is useful to users, and successful when:

- % of Fenix users adding Private Mode as Default Increases
- MAU increases as a result of Focus users hopping on Fenix
- % of Fenix users tapping on garbage icon increases 

#### What questions will you answer with this data?

- Whether or not users add Private Browsing Shortcut and/or use the Always Open in Private Browsing Feature
- Which system notification options users are interacting with the most
- Whether or not the garbage icon is discoverable and useful to users
- If users are tapping on the garbage icon accidentally with the use of the "undo" option on the snack bar

#### Why does Mozilla need to answer these questions? Are there benefits for users? Do we need this information to address product or business requirements?

Since Focus will be sunset, the initial data retrieved from Fenix users will help us understand if the private mode features are discoverable/useful. Going forward, if we detect discoverability issues, changes can be made for the onboarding on Focus users.

The benefits mostly revolve around our product being fast/private by design. Will promote safety when browsing the web and opening 3rd party links.

#### What probes (suggested, if applicable)

- Under Settings-> Privacy -> Private Browsing:
    1. Tap on the Add Private Browsing Shortcut -> Touch and Hold/Cancel/Add Automatically
    2. Toggle the "Always Open in Private Browsing" setting 
- Interact with the System Notification which includes:
   1. Tap on notification to delete private browsing history
   2. Tap Open button on notification
   3. Tap Delete and Open button
- Tap on the garbage can icon on the Private Browsing home page
- Tap on "undo" on the snackbar (after garbage icon is tapped)
- User added Firefox Preview (Private) as default browser (if feasible)

### Dependencies (Added by PM and Eng)

- Glean SDK (Event ping)

### Acceptance Criteria (Added by PM)
- Event pings can be queried via Amplitude
- We are sending telemetry events for the actions listed in the requirements
[...]

PR #5194, opened 2019-09-10
For #4658: Add private browsing mode shortcut telemetry

jyeontaek:
### Pull Request checklist

### DIFF
--- a/app/metrics.yaml
+++ b/app/metrics.yaml
@@ -1237,6 +1237,74 @@ private_browsing_mode:
       - fenix-core@mozilla.com
     expires: "2020-03-01"
 
+private_browsing_shortcut:
+  create_shortcut:
+    type: event
+    description: >
+      A user pressed the "Add private browsing shortcut" button in settings.
+    bugs:
+      - 4658
+    data_reviews:
+      - https://github.com/mozilla-mobile/fenix/pull/5194
+    notification_emails:
+      - fenix-core@mozilla.com
+    expires: "2020-03-01"
+  cfr_add_shortcut:
+    type: event
+    description: >
+      A user pressed the "Add shortcut" button when the contextual feature recommender appeared.
+    bugs:
+      - 4658
+    data_reviews:
+      - https://github.com/mozilla-mobile/fenix/pull/5194
+    notification_emails:
+      - fenix-core@mozilla.com
+    expires: "2020-03-01"
+  cfr_cancel:
+    type: event
+    description: >
+      A user pressed the "No thanks" button when the contextual feature recommender appeared.
+    bugs:
+      - 4658
+    data_reviews:
+      - https://github.com/mozilla-mobile/fenix/pull/5194
+    notification_emails:
+      - fenix-core@mozilla.com
+    expires: "2020-03-01"
+  pinned_shortcut_priv:
+    type: event
+    description: >
+      A user pressed the pinned private shortcut in Android home screen, opening up a new private search.
+    bugs:
+      - 4658
+    data_reviews:
+      - https://github.com/mozilla-mobile/fenix/pull/5194
+    notification_emails:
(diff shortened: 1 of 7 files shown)


app/metrics.yaml
```diff
@@ -1236,6 +1236,72 @@ private_browsing_mode:
...
+    description: >
+      A user pressed the "Add private browsing shortcut" button in settings.
+    bugs:
+      - 4658
+    data_reviews:
+      - https://github.com/mozilla-mobile/fenix/pull/5194
+    notification_emails:
+      - fenix-core@mozilla.com
+    expires: "2020-03-01"
+  cfr_add_shortcut:
```
NotWoods: I'd prefer not using an acronym here since its meaning isn't clear without looking at the description.

jyeontaek: I agree, but there's a character limit and I couldn't think of any other good names.

NotWoods: Ah, I wasn't aware there is a limit. In that case I can't think of a better name either :/

jyeontaek:

- Whether or not users add Private Browsing Shortcut
- Which shortcuts (long-press and pinned) are users interacting with the most

* Since Focus will be sunset, the initial data retrieved from Fenix users will help us understand if the private mode features are discoverable/useful. Going forward, if we detect discoverability issues, changes can be made for the onboarding on Focus users.

The benefits mostly revolve around our product being fast/private by design. Will promote safety when browsing the web and opening 3rd party links.

* Currently there are no events to track this.

* All data is Category 2.

* Until 03/01/2020

* Glean / Amplitude

app/metrics.yaml
```diff
@@ -1236,6 +1236,72 @@ private_browsing_mode:
...
+    description: >
+      A user pressed the "No thanks" button when the contextual feature recommender appeared.
+    bugs:
+      - 4658
+    data_reviews:
+      - https://github.com/mozilla-mobile/fenix/pull/5194
+    notification_emails:
+      - fenix-core@mozilla.com
+    expires: "2020-03-01"
+  pinned_shortcut_priv:
```
boek: would `opened_from_shortcut` make more sense here?

Also by home screen in the description do we mean the Android home screen or the Fenix home screen? If it's the former maybe we should call that out?

jyeontaek: There's different kinds of shortcuts, so I think `opened_from_shortcut` is also kind of ambiguous. 

That's a good point about the home screen; I'll clarify it in the description.

app/metrics.yaml
```diff
@@ -1236,6 +1236,72 @@ private_browsing_mode:
...
+      - 4658
+    data_reviews:
+      - https://github.com/mozilla-mobile/fenix/pull/5194
+    notification_emails:
+      - fenix-core@mozilla.com
+    expires: "2020-03-01"
+  static_shortcut_tab:
+    type: event
+    description: >
+      A user pressed the long-press shortcut "Open new tab", opening up a new search.
```
boek: Would it make sense to group these telemetry items together under maybe `private_browsing_shortcut`?

app/metrics.yaml
```diff
@@ -1236,6 +1236,72 @@ private_browsing_mode:
     notification_emails:
       - fenix-core@mozilla.com
     expires: "2020-03-01"
+  settings_shortcut:
```
boek: nit: name this `create_shortcut`

app/metrics.yaml
boek: @jyeontaek Are there multiple types of shortcuts in private_browsing_mode? Because this is nested it will get it's context from its parent

jyeontaek: Yes there are.

pocmo:
Should be fixed with the next build.

PR #5297, opened 2019-09-13
Refactored Tests - Closes #4556

### LINKED ISSUE #4556: Add @Before to reduce code duplication in tests
Lots of our test would benefit greatly from having an `@Before` function. It would make our testing code a lot easier to read. 

See [DefaultBrowserControllerTest](https://github.com/mozilla-mobile/fenix/pull/4493/files#diff-1144e67f32c3d338008db3149cbb1e60R57) for a good example.

usmanrafi:
Refactored tests to increase readability and reduce code duplication. Closes #4556 

### Pull Request checklist

### After merge
- [x] **Milestone**: Make sure issues finished by this pull request are added to the [milestone](https://github.com/mozilla-mobile/fenix/milestones) of the version currently in development.


app/src/test/java/org/mozilla/fenix/components/InflationAwareFeatureTest.kt
```diff
@@ -13,6 +13,11 @@ import org.mockito.Mockito.verify
 import java.lang.ref.WeakReference
 
 class InflationAwareFeatureTest {
+
+    private val stub: ViewStub = mock()
+    private val inflationFeature: InflationAwareFeature = spy(TestableInflationAwareFeature(stub))
+    private val innerFeature: LifecycleAwareFeature = mock()
```
pocmo: Reusing spys and mocks across tests can lead to a bunch of very hard to debug issues and wrong assertions.

We could use `lateinit var` and set them up from a `@Before` method for every test. But tbh I'm not sure whether its worth is. Explicit test isolation by duplication is not necessarily bad.

usmanrafi: Yes, I understand your point. I actually ran the tests and all of them were passing. Since the spies were created the same way again I decided to extract them out of the functions. If you'd like me to change it back to the way it was before (each function creates a spy), let me know. However, using `lateinit var` and `@Before` would result in the same thing I guess. And I opted for `val` to keep the code as consistent as possible.

Also, any other suggestions?

sblatz: @usmanrafi I think using `@Before` and `private lateinit var` is the pattern I'd suggest here. It's what we're using in most of our new tests and what I was suggesting in #4556.

usmanrafi: @sblatz Okay, I'll refactor the tests again. What do you think about the comment by @pocmo ? I think reusing mocks is the better option (only in the case that same mocks are being used and created in the same way)

sblatz: Yep, agreed.

boek:
Hey @usmanrafi, is this still something you're wanting to work on?

usmanrafi:
> Hey @usmanrafi, is this still something you're wanting to work on?

@boek Yes, I still want to work on this. I've been a little busy lately, but I do plan to do this in the next couple of days.

boek:
Hey @usmanrafi I'm going to close this for now. Feel free to open a new one if you ever pick it back up :)

app/src/test/java/org/mozilla/fenix/components/InflationAwareFeatureTest.kt
starkej2: > Reusing spys and mocks across tests can lead to a bunch of very hard to debug issues and wrong assertions.
> We could use lateinit var and set them up from a @Before method for every test.

I actually don't think there's a functional difference between initializing the test member variables at the declaration site vs initializing in `@Before` [since a fresh instance of the class is created for each test method](http://junit.sourceforge.net/javadoc/org/junit/Test.html).

PR #5311, opened 2019-09-13
Use new feature-contextmenu API.

pocmo:
In preparation for upcoming API breakage once the following PR lands in AC and the following snapshots roll out:
https://github.com/mozilla-mobile/android-components/pull/4407


pocmo:
Snapshots are getting release right now. This needs to land very soonish.

Issue #1744, opened 2019-04-17
FNX-5223 ⁃ Show "fill in from clipboard" on search engine screen


Verdi:
### Why/User Benefit/User Problem
If we open search shortcuts automatically when you:

1. Tap into the search field from the front page
1. You open a new tab
1. You clear the search field

We make it easy for people to select an alternate search engine.

### What / Requirements
* The Search Shortcuts should be displayed below the "Fill link from clipboard" function if applicable.
* The Search Shortcuts button should be selected
* Once you start typing, the Search Shortcuts should close

![Search Shortcuts](https://user-images.githubusercontent.com/107992/56324865-c2bf7900-6135-11e9-9244-cefa40491bf5.png)

sblatz:
I'm not sure I agree with this. I think it's a great idea for the search widget, but if it feels a bit weird to launch with this button pressed.

Verdi:
I wouldn't do something different for the search widget. We want the search experience to be consistent. As far as starting with the shortcuts expanded, I talked this over with both shorlander and mconnor and we agreed we'd like to try this.

sblatz:
@Verdi If the user doesn't want to use search shortcuts, they would have to press their default engine before getting search suggestions. Typing alone does not dismiss this menu. That seems confusing to me. What's the point in having a default engine if I have to pick each time 🤔

Verdi:
@sblatz My interaction requirements above say, "once you start typing, the Search Shortcuts should close."

sblatz:
@Verdi Ah, missed that! Sorry for the confusion :)

softvision-miralobontiu:
Verified as fixed on the latest Nightly build 190806 (#12180610) with LG G7 Fit (Android 8.1), and Nexus 5 (Android 6).

Verdi:
The shortcuts appear but the "Fill in from clipboard" command doesn't show. 

> ### What / Requirements
> 
>    * The Search Shortcuts should be displayed below the "Fill link from clipboard" function if applicable.
> 
> 
> ![Search Shortcuts](https://user-images.githubusercontent.com/107992/56324865-c2bf7900-6135-11e9-9244-cefa40491bf5.png)

sblatz:
@Verdi should that clipboard suggestion appear any time a user is in the shortcuts screen (from "shortcuts" button press AND from just opening search screen)? Should it disappear when the user starts typing?

Verdi:
[...]

PR #5389, opened 2019-09-17
 For #1744: Adds clipboard provider to search engine screen 

sblatz:
### Pull Request checklist

This functionality should be covered by UI tests. I've filed a follow-up to [add tests to the util file](https://github.com/mozilla-mobile/fenix/issues/5390)

demo

### After merge

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/Utilities.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/Utilities.kt
@@ -11,6 +11,7 @@ import mozilla.components.feature.intent.processing.TabIntentProcessor
 import mozilla.components.feature.search.SearchUseCases
 import mozilla.components.feature.session.SessionUseCases
 import org.mozilla.fenix.test.Mockable
+import org.mozilla.fenix.utils.ClipboardHandler
 
 /**
  * Component group for miscellaneous components.
@@ -38,5 +39,9 @@ class Utilities(
 
     val customTabIntentProcessor by lazy {
         CustomTabIntentProcessor(sessionManager, sessionUseCases.loadUrl, context.resources)
+    }
+
+    val clipboardHandler by lazy {
+        ClipboardHandler(context)
     }
 }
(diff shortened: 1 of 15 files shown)


app/src/main/java/org/mozilla/fenix/components/Utilities.kt
```diff
@@ -39,4 +40,8 @@ class Utilities(
     val customTabIntentProcessor by lazy {
         CustomTabIntentProcessor(sessionManager, sessionUseCases.loadUrl, context.resources)
     }
+
```
csadilek: I'd recommend moving this directly the root Component class: https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/java/org/mozilla/fenix/components/Components.kt

See also: https://github.com/mozilla-mobile/fenix/commit/4922e5d09df650a1bf952dbd2a6652fdaa4669c3#diff-6a1dd8505f3ec4f786426dccbb4a5ca6 (where adding to utils caused a perf. problem).

I think it's also cleaner as otherwise more and more stuff will be added to "utils" and it will eventually become messy. We could also rename `Utilities` to `IntentProcessors` to prevent that.

app/src/main/java/org/mozilla/fenix/utils/ClipboardHandler.kt
```diff
@@ -0,0 +1,44 @@
...
+
+package org.mozilla.fenix.utils
+
+import android.content.ClipData
+import android.content.ClipboardManager
+import android.content.Context
+import mozilla.components.support.utils.WebURLFinder
+
+private const val MIME_TYPE_TEXT_PLAIN = "text/plain"
+
```
csadilek: nit: kdocs.

app/src/main/java/org/mozilla/fenix/utils/ClipboardHandler.kt
```diff
@@ -0,0 +1,44 @@
...
+class ClipboardHandler(context: Context) {
```
csadilek: Can we add a unit test for this class? :)

app/src/main/java/org/mozilla/fenix/components/Utilities.kt
sblatz: I moved this out and renamed Utilities to IntentProcessors so it's more clear utils shouldn't be added there :)

sblatz:
@csadilek this is ready for review again!

PR #5424, opened 2019-09-19
 Add basic test coverage for Glean metrics in Fenix 

Dexterp37:
This introduces test coverage, using the Glean SDK testing API, for the metrics that are set at startup
by Fenix in the GleanMetricsService.

This additional adds a basic test for the translation of the `app_opened` event.

**Note** this only adds basic things to catch any regression due to Glean SDK migration.

cc @liuche 

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt
@@ -401,28 +401,36 @@ class GleanMetricsService(private val context: Context) : MetricsService {
             Glean.registerPings(Pings)
             Glean.initialize(context, Configuration(channel = BuildConfig.BUILD_TYPE))
 
-            Metrics.apply {
-                defaultBrowser.set(Browsers.all(context).isDefaultBrowser)
-                MozillaProductDetector.getMozillaBrowserDefault(context)?.also {
-                    defaultMozBrowser.set(it)
-                }
-                mozillaProducts.set(MozillaProductDetector.getInstalledMozillaProducts(context))
-            }
-
-            SearchDefaultEngine.apply {
-                val defaultEngine = context
-                    .components
-                    .search
-                    .searchEngineManager
-                    .defaultSearchEngine ?: return@apply
-
-                code.set(defaultEngine.identifier)
-                name.set(defaultEngine.name)
-                submissionUrl.set(defaultEngine.buildSearchUrl(""))
-            }
-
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt
```diff
@@ -401,28 +401,36 @@ class GleanMetricsService(private val context: Context) : MetricsService {
...
+            setStartupMetrics()
+        }
+    }
 
-                code.set(defaultEngine.identifier)
-                name.set(defaultEngine.name)
-                submissionUrl.set(defaultEngine.buildSearchUrl(""))
+    /**
+     * Sets some basic Glean metrics required by Fenix.
+     * This is a separate function to simplify testing.
```
sblatz: nit: I don't think this comment is necessary. Having a helper function for this to make it more testable seems perfectly reasonable without justification :)

app/src/test/java/org/mozilla/fenix/components/metrics/GleanMetricsServiceTest.kt
```diff
@@ -0,0 +1,74 @@
...
+@ObsoleteCoroutinesApi
+@RunWith(RobolectricTestRunner::class)
+@Config(application = TestApplication::class)
+class GleanMetricsServiceTest {
+    @get:Rule
+    val gleanRule = GleanTestRule(testContext)
+
+    @Test
+    fun `setStartupMetrics sets some base metrics`() {
+        val gleanService = GleanMetricsService(testContext)
```
sblatz: Perhaps we should swap this for a `private lateinit var gleanService: GleanMetricsService`
and add an `@Before` that inits it? I'm assuming we'll want to use it in most tests here and we're already using it in 100% of the tests in this file :)

This would also better pattern match the tests we've been writing recently.

app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt
Dexterp37: Ah, good point :)

app/src/test/java/org/mozilla/fenix/components/metrics/GleanMetricsServiceTest.kt
Dexterp37: Sounds good to me! As long as we don't track any state between tests, that's fine by me!

PR #5430, opened 2019-09-19
Bug 1580778 - Migrate lint and ui-test tasks to taskgraph

JohanLorenzo:
Similar to https://github.com/mozilla-mobile/reference-browser/pull/892

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/taskcluster/ci/lint/kind.yml
+++ b/taskcluster/ci/lint/kind.yml
@@ -1,0 +1,44 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+---
+
+loader: taskgraph.loader.transform:loader
+
+transforms:
+    - taskgraph.transforms.job:transforms
+    - taskgraph.transforms.task:transforms
+
+job-defaults:
+    attributes:
+        code-review: true
+    worker-type: b-android
+    worker:
+        docker-image: {in-tree: base}
+        max-run-time: 7200
+    run:
+        use-caches: false
+    run-on-tasks-for: [github-pull-request, github-push]
+
+jobs:
+    detekt:
+        description: 'Running detekt over all modules'
+        run:
+            using: gradlew
+            gradlew: [detekt]
+    ktlint:
+        description: 'Running ktlint over all modules'
+        run:
+            using: gradlew
+            gradlew: [ktlint]
+    compare-locales:
+        description: 'Validate strings.xml with compare-locales'
+        run:
+            using: run-task
+            cwd: '{checkout}'
+            command: 'pip install --user "compare-locales>=5.0.2,<6.0" && compare-locales --validate l10n.toml .'
+    lint:
+        description: 'Running tlint over all modules'
+        run:
+            using: gradlew
+            gradlew: [lintDebug]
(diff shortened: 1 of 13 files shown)


sblatz:
@JohanLorenzo looks like `test-geckoNightlyDebug` is failing:

```
taskcluster.exceptions.TaskclusterRestFailure: This request requires Taskcluster credentials that satisfy the following scope expression:
secrets:get:project/mobile/fenix/public-tokens
```

taskcluster/ci/lint/kind.yml
```diff
@@ -0,0 +1,44 @@
...
+
+job-defaults:
+    attributes:
+        code-review: true
+    worker-type: b-android
+    worker:
+        docker-image: {in-tree: base}
+        max-run-time: 7200
+    run:
+        use-caches: false
```
mitchhentges: Why don't you want to use caches for this task?

taskcluster/ci/test/kind.yml
```diff
@@ -0,0 +1,50 @@
...
+    ui:
+        attributes:
+            build-type: debug
+            code-review: true
+        treeherder:
+            symbol: ui
+            platform: 'ui-test/opt'
+        run-on-tasks-for: [github-pull-request, github-push]
+        run:
+            # TODO Generate APKs in a build task instead
```
mitchhentges: I don't quite follow - why do we want to generate APKs in a separate build task?

taskcluster/ci/lint/kind.yml
JohanLorenzo: That's a great question. I primarily copied what @tomprince did in reference-browser: https://github.com/mozilla-mobile/reference-browser/blob/4748d9e766d62c5156829cb1ed72c73ba996582c/taskcluster/ci/lint/kind.yml#L20

After thinking about it more thoroughly, we run `gradlew clean` before each of these tasks, so mounting the cache doesn't provide any value. Do I get that right, @tomprince ? 

Anyway, I'm happy to make the change in a followup.

taskcluster/ci/test/kind.yml
JohanLorenzo: We have some duplication of builds here. `assembleDebug` is done in the PR build task. That's not entirely true, because we just `assembleNightlyDebug` there, but the UI tests are just interested in the nightly build anymore.

While this debug build is assembled, we can also parallelize `assembleAndroidTest`. This way, the UI tests just run the tests, and should take less time to execute overall. I heard the duration of the UI tests is becoming a pain point in PRs.

JohanLorenzo:
> @JohanLorenzo looks like `test-geckoNightlyDebug` is failing: [...]

Thanks! I found the root cause. It turns out `get-secrets.py` is called under 2 different environments:

* the most recent one, where taskgraph builds a docker images which contains the most recent `taskcluster` python library.
* the legacy one, where the docker image is still `mozillamobile/fenix:1.4` which contains an old `taskcluster` python lib.

The latter will go away in a couple of weeks. So I added support for both libraries in `get-secrets.py`. It now works everywhere!

taskcluster/ci/lint/kind.yml
tomprince: It has been awhile since I did this, but I suspect the reason is I didn't want to think too hard about scopes for docker-worker caches at the time, and then never revisited it.

Issue #2765, opened 2019-05-23
FNX3-16565 ⁃ Fenix should run places maintenance periodically


thomcc:
@grigoryk indicated that it does not, which will result in worse perf, more disk usage, etc.

sblatz:
@thomcc @grigoryk can you please clarify what this means? This ticket is not actionable for the fenix engineering team as it stands.

thomcc:
There's a function call on writable places connections called runMaintenance: https://github.com/mozilla/application-services/blob/ce7652ddc1a8724d617bd187c6a08e996f329b96/components/places/android/src/main/java/mozilla/appservices/places/PlacesConnection.kt#L663-L677

This should be called once per day. I don't know how it's exposed through A-C's history adapters. I also don't have the ability to remove the waiting label.

PR #5466, opened 2019-09-20
For #2765 - Run places maintenance daily

ekager:
### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
+++ b/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
@@ -16,6 +16,7 @@ import kotlinx.coroutines.Deferred
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.GlobalScope
 import kotlinx.coroutines.async
+import kotlinx.coroutines.launch
 import kotlinx.coroutines.runBlocking
 import mozilla.appservices.Megazord
 import mozilla.components.concept.push.PushProcessor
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -110,6 +113,18 @@ open class FenixApplication : Application() {
...
         components.core.sessionManager.register(NotificationSessionObserver(this))
+
+        if (!DateUtils.isToday(this.settings.lastPlacesStorageMaintenance)) {
+            runStorageMaintenance()
+        }
+    }
+
+    private fun runStorageMaintenance() {
+        GlobalScope.launch {
+            bookmarkStorage()?.runMaintenance()
```
grigoryk: `bookmarks` and `history` are both handled by the same `writer` instance internally - they both sit on top of the same sqlite db file. Underneath, both of these calls will end up as the same call to Rust: `places_run_maintenance`.
So, just pick one - doesn't matter which - and add a comment to explain why it's just one and not both.

Once we add logins, it'll need its own maintenance call, since it's backed by a different db file.

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -110,6 +113,18 @@ open class FenixApplication : Application() {
         registerActivityLifecycleCallbacks(visibilityLifecycleCallback)
 
         components.core.sessionManager.register(NotificationSessionObserver(this))
+
+        if (!DateUtils.isToday(this.settings.lastPlacesStorageMaintenance)) {
```
grigoryk: I'd write this as "if (last maintenance happened more than 24hrs ago) {}". I _think_ `isToday` does that..?

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -110,6 +113,18 @@ open class FenixApplication : Application() {
...
+            runStorageMaintenance()
+        }
+    }
+
+    private fun runStorageMaintenance() {
+        GlobalScope.launch {
```
grigoryk: You probably want this to run on the IO dispatcher?

Issue #5468, opened 2019-09-21
[Bug] Long Press on Toolbar in Custom Tab Only Copies Host 


ekager:
## Steps to reproduce
Open GitHub link from slack
Long press on custom tab toolbar
Select copy 

### Expected behavior
Have the GitHub Issue in my clipboard 

### Actual behavior
Have "GitHub.com" in my clipboard 

### Device information

* Android device: ?
* Fenix version: ?

AndiAJ:
Hi, this is **still reproducible** on Nightly Build #12671258 from 24/9 using the following devices:
• Google Pixel 3a (Android 9)
• Huawei Mate 20 Lite (Android 8.1.0)
• Samsung Galaxy S7 (Android 7)
• OnePlus A3 (Android 6.0.1)

#### Note
I've used Gmail to check the above described issue.
► Video
![20190925_110824](https://user-images.githubusercontent.com/51314259/65582542-4bb0cb00-df86-11e9-8f4e-45dab879ca87.gif)

AndiAJ:
I've just had a chat with @mcarare and will retest on the next nightly build.

AndiAJ:
Hi, **verified as fixed** on Nightly Build #12681804 from 25/9 using the following devices:
• Google Pixel 3a (Android 9)
• Huawei Mate 20 Lite (Android 8.1.0)
• Samsung Galaxy S7 (Android 7)
• OnePlus A3 (Android 6.0.1)
• LG Nexus 4 (Android 5.1.1)

► Video
![20190926_121612](https://user-images.githubusercontent.com/51314259/65676260-3dcb7a80-e058-11e9-8c94-563ffeae4f60.gif)

PR #5514, opened 2019-09-24
For #5468 Set correct info source for url on clipboard copy in customTab

mcarare:
### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarView.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarView.kt
@@ -62,6 +62,8 @@ class BrowserToolbarView(
                 true
             )
 
+            val selectedSession = container.context.components.core.sessionManager.selectedSession
+
             popupWindow.elevation = view.context.dimen(R.dimen.mozac_browser_menu_elevation).toFloat()
 
             customView.paste.isVisible = !clipboard.text.isNullOrEmpty() && !isCustomTabSession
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarView.kt
```diff
@@ -69,7 +69,11 @@ class BrowserToolbarView(
 
             customView.copy.setOnClickListener {
                 popupWindow.dismiss()
-                clipboard.text = view.url.toString()
+                if (isCustomTabSession) {
+                    clipboard.text = customTabSession?.url
+                } else {
+                    clipboard.text = view.url.toString()
```
Amejia481: There are some conversions around not showing the protocol (HTTP or HTTPS) for saving space on the toolbar, this may break in the future. Is it possible to use the `url` from the session?

mcarare: As far as I have tested, If you have Fenix already opened, using `selectedSession` will return the url from that tab.

mcarare: But the conversion you mention is just for the customTab `view.url,` `customTabSession` still retains the full url.

Amejia481: Yes, but I'm referring to the normal tabs case, where we are using the `url` directly from the view `clipboard.text = view.url.toString()`. If this view doesn't include the protocol it will happen something similar to what we are fixing with custom tabs now

mcarare: I understand now. I was focused only on my own changes. :)
I will change that too, to get url from the session. Thanks for the heads=up! 👍

Amejia481: 👍

Issue #5552, opened 2019-09-25
[Bug] Fenix crashes on es locales


boek:
https://github.com/mozilla-mobile/fenix/pull/5482/commits/2886b0d31e44066f5c2ddaa94c201b03513512d6#diff-30eefb34d3cd95f1e83b7ca91d80dcb0R213

is causing: https://sentry.prod.mozaws.net/operations/fenix-nightly/issues/6326356/

@Pike @Delphine What is the process to get this fixed in the l10n repo? :)

Pike:
That's part of https://github.com/mozilla-l10n/android-l10n/pull/86, which is coming back here in #5562. Sorry for the trouble.

The general way to fix this is for you to request Delphine to go in and fix or unapprove (remove) the string on Pontoon, and then also get it here by the way of the import step.

ekager:
@boek did we ever get this resolved on the Pontoon side or did we just reintroduce the same crash with latest string exports 😆

boek:
I think we fixed it on both. We’ll want to take out the temporary fix though!

-j

sv-ohorvath:
This was verified on Nightly 1/20.
devices: Nexus 5 (android 6.0.1), Galaxy Note 8 (android 9)

PR #5553, opened 2019-09-25
For #5552 - Temporary fix for startup crash on es locales

boek:


app/src/main/java/org/mozilla/fenix/components/BackgroundServices.kt
```diff
@@ -61,12 +62,20 @@ class BackgroundServices(
...
+                R.string.default_device_name,
+                context.getString(R.string.app_name),
+                Build.MANUFACTURER,
+                Build.MODEL
+            )
+        } catch (ex: FormatFlagsConversionMismatchException) {
+            "%s on %s %s".format(
+                context.getString(R.string.app_name),
+                Build.MANUFACTURER,
+                Build.MODELq
```
Baron-Severin: Nit: `MODELq` looks like a typo

app/src/main/java/org/mozilla/fenix/components/BackgroundServices.kt
```diff
@@ -61,12 +62,20 @@ class BackgroundServices(
...
-        Build.MODEL
-    )
+    fun defaultDeviceName(context: Context): String = try {
+            context.getString(
+                R.string.default_device_name,
+                context.getString(R.string.app_name),
+                Build.MANUFACTURER,
+                Build.MODEL
+            )
+        } catch (ex: FormatFlagsConversionMismatchException) {
```
Baron-Severin: Nit: since this is a temporary fix, would you mind adding a TODO pointing to the issue and with a quick explanation?  Something small, like `// A malformed string is causing crashes. This will be removed when the string is fixed. See #5552`

app/src/main/java/org/mozilla/fenix/components/BackgroundServices.kt
```diff
@@ -40,6 +40,7 @@ import org.mozilla.fenix.ext.components
 import org.mozilla.fenix.ext.settings
 import org.mozilla.fenix.isInExperiment
 import org.mozilla.fenix.test.Mockable
+import java.util.*
```
csadilek: nit: wildcard import / will fail ktlint.

Issue #3584, opened 2019-06-20
FNX3-17135 ⁃ [Bug][a11y] Three-dot menu is reopened when double tapped anywhere outside the opened menu


softvision-miralobontiu:
## Steps to reproduce
1. Turn Talkback on.
1. Tap on three-dot menu.
1. Exit the menu double-tapping outside it.

### Expected behavior
Double tapping outside the menu should dismiss the menu.

### Actual behavior
The three-dot menu is opened again and again when tapping anywhere outside the menu.

### Device information
* Android device: Samsung Galaxy Tab S3 (Android 8)
* Fenix version: 1.0.0-rc.2 (#11700147)

![menu1](https://user-images.githubusercontent.com/40694158/59847590-9a3d4080-936b-11e9-8e4e-3184caf47861.gif)

pdjstone:
I get a similar issue on my phone - if I double-tap the three-dot menu button, it opens twice (you have to complete the double-tap before the menu opens, a slower device probably helps). Two separate taps outside the menu are needed to close both.

mihai-adrian:
I think the first issue mentioned is actually the correct way to respond to a double tap, when a button is selected. The official documentation mentions:"When the focus reaches an item you'd like to select, double-tap **anywhere** on the screen to select the focused item" (https://support.google.com/accessibility/android/answer/6006598?hl=en).

The second issue mentioned is that due to the above functionality, after a double tap, the button still has focus and has the same onClickListener, and if you double tap it another settings pop-up is shown.

There are three approaches to fix this : 
- change the listener to dismiss the settings pop-up when it is already opened (chrome, skype and others do that)
- check if popup is already shown and prevent showing a second one
- remove focus from button after pop-up is shown.

Which one would be the preferred approach?

lime124:
I think this is more of an eng ask of how they want to fix the bug? /cc @boek

DiegoARL38:
As a regular screen reader user, dismissing the menu if already open is what I expect, since most applications do this.

sv-ohorvath:
Verified as fixed. The settings option is selected by default now when you open the 3 dot menu.

Mugurell:
Closing as per https://github.com/mozilla-mobile/fenix/issues/3584#issuecomment-541715484

PR #5674, opened 2019-09-30
For #3584 - Check if menu is open before showing a new one

Mugurell:
Add a simple check for if the menu is showing depending on which we will know not to construct and show a new menu.
This check is to be done by both the "Open tabs menu" and the "Home menu"

### Pull Request checklist

- [x] **Quality**: This PR builds and passes detekt/ktlint checks
- [x] **Tests**: This PR does not includes tests as it only contains small changes in not yet tested components
- [x] **Screenshots**: This PR does not include screenshots because there are no UI changes, just small behavior ones which are easy to test.
- [x] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md)

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
@@ -239,11 +239,19 @@ class HomeFragment : Fragment() {
             }
         }
 
-        view.menuButton.setOnClickListener {
-            homeMenu?.menuBuilder?.build(requireContext())?.show(
-                anchor = it,
-                orientation = BrowserMenu.Orientation.DOWN
-            )
+        with(view.menuButton) {
+            var menu: PopupWindow? = null
+            setOnClickListener {
+                if (menu == null) {
+                    menu = homeMenu?.menuBuilder?.build(requireContext())?.show(
+                        anchor = it,
+                        orientation = BrowserMenu.Orientation.DOWN,
+                        onDismiss = { menu = null }
+                    )
+                } else {
+                    menu?.dismiss()
+                }
+            }
         }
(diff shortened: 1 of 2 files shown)


sblatz:
Rebasing now that master builds 😬

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -239,11 +239,19 @@ class HomeFragment : Fragment() {
...
+                if (menu == null) {
+                    menu = homeMenu?.menuBuilder?.build(requireContext())?.show(
+                        anchor = it,
+                        orientation = BrowserMenu.Orientation.DOWN,
+                        onDismiss = { menu = null }
+                    )
+                } else {
+                    menu?.dismiss()
+                }
+            }
```
sblatz: Is this else branch necessary? Can't we just choose to only inflate if menu is null?

Mugurell: Thanks Sawyer!
Indeed, the commit title and description should be improved to better describe the issue and the solution.
The `else` allows us to close an already opened menu if the 3-dot menu button is pressed again. This is the same behavior as for Chrome and basically the same recent implementation for the browser toolbar - https://github.com/mozilla-mobile/android-components/pull/4594

sblatz: Ah, great. I guess it makes sense for TalkBack enabled devices too :)

Thanks!

sblatz:
bors r=sblatz

Issue #3563, opened 2019-06-19
Avoid calling `notifyDataSetChanged` in RecyclerView adapters


NotWoods:
`notifyDataSetChanged` tells the recycler view that every single item should be updated. However, its usually called when only 1 item has changed. Our `RecyclerView.Adapter`s should be changed to only notify that the changed items should be updated.

The easiest way to do this is to just use [`ListAdapter`](https://developer.android.com/reference/kotlin/androidx/recyclerview/widget/ListAdapter) which does diffing automatically.

sblatz:
This would be really nice to have but is probably not worth the ROI for us, would be great for a contributor :)

dector:
I can work on this issue. Suppose that [DiffUtils](https://developer.android.com/reference/android/support/v7/util/DiffUtil) is what we actually need here.

NotWoods:
Yeah, either using DiffUtils or ListAdapter (which uses DiffUtils automatically) will work 😄

J6ey:
Keep in mind ListAdapter is deprecated. To notify a particular item has changed use  `notifyItemChanged` or `notifyItemRangeChanged`

NotWoods:
I was referring to the (confusingly named) AndroidX [`ListAdapter`](https://developer.android.com/reference/kotlin/androidx/recyclerview/widget/ListAdapter), which builds ontop of the `RecyclerView.Adapter` and isn't deprecated.

NotWoods:
Closing as we've addressed most of the cases, and the remaining ones aren't called frequently.

PR #5698, opened 2019-10-01
For #3563. Use `ListAdapter` for App Share list.

dector:
Breaking change: not using external state for `applications` list (it's not used elsewhere anyway and is kinda error-prone practice).

### Pull Request checklist

- [x] ~~**Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not~~
- [x] ~~**Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features~~

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/share/listadapters/AppShareAdapter.kt
+++ b/app/src/main/java/org/mozilla/fenix/share/listadapters/AppShareAdapter.kt
@@ -7,14 +7,14 @@ package org.mozilla.fenix.share.listadapters
 import android.graphics.drawable.Drawable
 import android.view.LayoutInflater
 import android.view.ViewGroup
-import androidx.recyclerview.widget.RecyclerView
+import androidx.recyclerview.widget.DiffUtil
+import androidx.recyclerview.widget.ListAdapter
 import org.mozilla.fenix.share.ShareToAppsInteractor
 import org.mozilla.fenix.share.viewholders.AppViewHolder
 
 class AppShareAdapter(
-    private val interactor: ShareToAppsInteractor,
-    private val applications: MutableList<AppShareOption> = mutableListOf()
-) : RecyclerView.Adapter<AppViewHolder>() {
+    private val interactor: ShareToAppsInteractor
+) : ListAdapter<AppShareOption, AppViewHolder>(DiffCallback) {
 
     override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): AppViewHolder {
         val view = LayoutInflater.from(parent.context)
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/mozilla/fenix/share/listadapters/AppShareAdapter.kt
```diff
@@ -7,14 +7,19 @@ package org.mozilla.fenix.share.listadapters
...
 
 class AppShareAdapter(
     private val interactor: ShareToAppsInteractor,
-    private val applications: MutableList<AppShareOption> = mutableListOf()
-) : RecyclerView.Adapter<AppViewHolder>() {
+    applications: List<AppShareOption> = emptyList()
+) : ListAdapter<AppShareOption, AppViewHolder>(DiffCallback()) {
+
+    init {
+        submitList(applications)
```
NotWoods: I think we should just replace uses of the second constructor parameter with a call to `submitList`

app/src/main/java/org/mozilla/fenix/share/listadapters/AppShareAdapter.kt
```diff
@@ -23,19 +28,24 @@ class AppShareAdapter(
...
     }
 
-    override fun getItemCount(): Int = applications.size
-
     override fun onBindViewHolder(holder: AppViewHolder, position: Int) {
-        holder.bind(applications[position])
+        holder.bind(getItem(position))
     }
 
     fun updateData(applications: List<AppShareOption>) {
```
NotWoods: We should just use `submitList` instead, no need to keep the `updateData` alias

app/src/main/java/org/mozilla/fenix/share/listadapters/AppShareAdapter.kt
```diff
@@ -23,19 +28,24 @@ class AppShareAdapter(
...
-        this.applications.clear()
-        this.applications.addAll(applications)
-        notifyDataSetChanged()
+        submitList(applications)
     }
 }
 
+private class DiffCallback : DiffUtil.ItemCallback<AppShareOption>() {
```
NotWoods: nit: use `object`
```suggestion
private object DiffCallback : DiffUtil.ItemCallback<AppShareOption>() {
```

app/src/main/java/org/mozilla/fenix/share/listadapters/AppShareAdapter.kt
dector: I guess it's a good idea.

app/src/main/java/org/mozilla/fenix/share/listadapters/AppShareAdapter.kt
dector: :+1:

PR #5456, opened 2019-09-20
1581556: Initialize Glean on the main thread

mdboom:
`Glean.initialize` registers a couple of `LifecycleObservers` and this registration must happen on the main thread.

See [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1581556) for all of the gory details and the problems this causes.

This will be enforced in a-c in https://github.com/mozilla-mobile/android-components/pull/4477, but let's merge this PR first so an upgrade to a-c won't break us here.

### Pull Request checklist

- [ ] **Quality**: This PR builds and passes detekt/ktlint checks (A pre-push hook is recommended)
- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features

### After merge
- [ ] **Milestone**: Make sure issues finished by this pull request are added to the [milestone](https://github.com/mozilla-mobile/fenix/milestones) of the version currently in development.

### To download an APK when reviewing a PR:
1. click on Show All Checks,
2. click Details next to "Taskcluster (pull_request)" after it appears and then finishes with a green checkmark,
3. click on the "Fenix - assemble" task, then click "Run Artifacts".
4. the APK links should be on the left side of the screen, named for each CPU architecture

PR #5706, opened 2019-10-01
1585250: Initialize metrics off of the main thread

mdboom:
This fixes an intermittent failure introduced by https://github.com/mozilla-mobile/fenix/pull/5456

That fix calls Glean on the main thread (which is required because Glean registers lifecycle observers).  However, it also inadvertently captured another API call that must be called off of the main thread [getAdvertisingIdInfo](https://developers.google.com/android/reference/com/google/android/gms/ads/identifier/AdvertisingIdClient#public-static-advertisingidclient.info-getadvertisingidinfo-context-context).  This simply moves as much as possible back to the non-main coroutine thread.

The failure this resolves, as reported by @sblatz on Slack is:

image

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt
@@ -430,12 +430,17 @@ class GleanMetricsService(private val context: Context) : MetricsService {
         if (initialized) return
         initialized = true
 
+        // We have to initialize Glean *on* the main thread, because it registers lifecycle
+        // observers. However, the activation ping must be sent *off* of the main thread,
+        // because it calls Google ad APIs that must be called *off* of the main thread.
+        // These two things actually happen in parallel, but that should be ok because Glean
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt
```diff
@@ -430,12 +430,17 @@ class GleanMetricsService(private val context: Context) : MetricsService {
         if (initialized) return
         initialized = true
 
+        // We have to initialize Glean *on* the main thread, because it registers lifecycle
+        // observers. However, the activation ping must be sent *off* of the main thread,
+        // because it calls Google ad APIs that must be called *off* of the main thread.
+        // These two things actually happen in parallel, but that should be ok because Glean
```
Dexterp37: Yes, I think this should be covered, especially because we call `setUploadEnabled` before. I think we should manually check that the activation ping gets sent, if possible.

PR #5801, opened 2019-10-04
Closes #1022: Auto-publication workflow for android-component 

### LINKED ISSUE #1022: FNX-4726 ⁃ Dependency substitution for android-components
This will allow us to easily develop against a local checkout of `android-components`.

Currently blocked on a-c moving to AndroidX (https://github.com/mozilla-mobile/android-components/issues/842). Also, see similar r-b issue: https://github.com/mozilla-mobile/reference-browser/issues/365

grigoryk:
This requires an android-components counterpart: https://github.com/mozilla-mobile/android-components/pull/4658

Initially, I went down the "make composite builds work well with a-c" rabbithole, and while they generally worked, experience in Android Studio was horrible. It would get _very_ confused and broken. So instead, I've settled for automating a publication workflow.

This patch enabled support for an auto-publication workflow for android-components.
It automates a common pattern seen in local development:

Old way:
- after every change in a-c, publish it locally with a unique version (bumping it manually)
- manually modify Fenix to consume a custom version of a-c from a mavenLocal repository
- this involves modifying Dependencies.kt for each new a-c version, fixing a few things in `app/build.gradle`, and specifying a new repository in `build.gradle`. 

New way:
- set a flag in fenix's local.properties to enable auto-publication
- run Fenix builds after making changes to a-c. Changes in a-c will be automatically picked up.
- no other changes are necessary to any Fenix files other than a single flag in local.properties!
- manually bumping android-components is no longer needed!

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,29 @@ To push without running the pre-push hook (e.g. doc updates):
 git push <remote> --no-verify
 ```
 
+## local.properties helpers
+There are multiple helper flags available via `local.properties` that will help speed up local development workflow
+when working across multiple layers of the dependency stack - specifically, with android-components, geckoview or application-services.
+
+### android-components auto-publication workflow
+Specify a relative path to your local `android-components` checkout via `autoPublish.android-components.dir`.
+
+If enabled, during a Fenix build android-components will be compiled and locally published if it has been modified,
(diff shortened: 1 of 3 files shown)


grigoryk:
Some tasks failed for server-side issues (500 while fetching dependencies). Trying again.

PR #5878, opened 2019-10-09
Fennec data migration basics: open tabs, history, bookmarks

grigoryk:
This will be a rolling PR for a little bit, until things settle down.
Companion A-C PR: https://github.com/mozilla-mobile/android-components/pull/4693

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -415,6 +415,9 @@ dependencies {
     implementation Deps.mozilla_support_rustlog
     implementation Deps.mozilla_support_utils
 
+    // We only care about support-migration in builds that will be overwriting Fennec.
+    fennecProductionImplementation Deps.mozilla_support_migration
+
     implementation Deps.mozilla_ui_colors
     implementation Deps.mozilla_ui_icons
     implementation Deps.mozilla_ui_publicsuffixlist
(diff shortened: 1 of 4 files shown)


app/build.gradle
```diff
@@ -413,6 +413,7 @@ dependencies {
     implementation Deps.mozilla_support_ktx
     implementation Deps.mozilla_support_rustlog
     implementation Deps.mozilla_support_utils
+    implementation Deps.mozilla_support_migration
```
pocmo: nit: Lets only add this for the `fennecProduction` build type. This will also make sure that we are not landing anything in "regular fenix" by mistake. :)

app/src/fennecProduction/java/org/mozilla/fenix/MigratingFenixApplication.kt
```diff
@@ -0,0 +1,20 @@
+package org.mozilla.fenix
```
pocmo: nit: license header.

app/build.gradle
grigoryk: Good call!

grigoryk:
@pocmo this should good to land, just pending mozilla-mobile/android-components#4693.

pocmo:
Re-added the label to wait for the AC part to land first.

grigoryk:
bors r=pocmo

pocmo:
Force merge to resolve a situation where we need two PR to land in order to fix the builds.

Issue #5534, opened 2019-09-24
[Bug] Find in page UI Test broken on API 25 for both fenix and r-b


rpappalax:
## Steps to reproduce
Run findInPage automated test.
Pulls up a static html asset with known content.
searches for a specific combination of chars

https://github.com/mozilla-mobile/fenix/blob/master/app/src/androidTest/java/org/mozilla/fenix/ui/NavigationToolbarTest.kt#L145

### Expected behavior
Fenix should display 1/3 for value of find_in_page_result_text

### Actual behavior
Fenix displays 1/5 for value of find_in_page_result_text and thus fails on this line:https://github.com/mozilla-mobile/fenix/blob/master/app/src/androidTest/java/org/mozilla/fenix/ui/NavigationToolbarTest.kt#L145

### Device information

* Android device: Nexus9 (x86 devicer)
* Fenix version: master

rpappalax:
<img width="460" alt="image" src="https://user-images.githubusercontent.com/3699494/65790587-cc9dcd00-e114-11e9-92f8-e67f83036e6b.png">

rpappalax:
@sblatz ^ this is what we're seeing intermittently.  The find in page for an asset that has consistent static content should be 1/7 and is instead 11/42

ekager:
It looks like there are maybe 42 "i"s on the page and it's updating incorrectly? This could be a bug with find in page that would have to be fixed in AC

rpappalax:
@ekager yes, I'm seeing the same issue in r-b except in that case, the expect text is missing altogether

```
failed: findInPageTest(org.mozilla.reference.browser.ui.ThreeDotMenuTest)
09-26 17:42:56.222: E/TestRunner(11929): ----- begin exception -----
09-26 17:42:56.223: E/TestRunner(11929): androidx.test.espresso.base.DefaultFailureHandler$AssertionFailedWithCauseError: 'with text: is "1/3"' doesn't match the selected view.
09-26 17:42:56.223: E/TestRunner(11929): Expected: with text: is "1/3"
[...]

PR #4914, opened 2019-08-23
Closes #778 - Progressive Web Apps!

NotWoods:
Closes #778
Closes #772
Closes #770 
Closes #769
Closes #779

### Pull Request checklist

- [x] **Quality**: This PR builds and passes detekt/ktlint checks (A pre-push hook is recommended)
- [x] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [x] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features

To download an APK when reviewing a PR:
1. click on Show All Checks,
2. click Details next to "Taskcluster (pull_request)" after it appears and then finishes with a green checkmark,
3. click on the "Fenix - assemble" task, then click "Run Artifacts".

Issue #778, opened 2018-11-29
Install a PWA to my home screen


pocmo:
### User Story
As a user, I want to be able to install a PWA to my home screen, so I can launch it later directly from my homescreen similar to a standalone app.

#Acceptance Criteria
-I can install a PWA to my homescreen
-The launcher icon uses the PWA's icon (web app manifest)

bennypowers:
I'm looking forward to WebAPK support in a Mozilla browser

reinhart1010:
Duplicate: https://groups.google.com/forum/m/#!topic/fenix-nightly/y0s-eiLwA0o

PR #5901, opened 2019-10-09
For #778: Follow-up to fix find in page

csadilek:
Find in page just broke on master after merging https://github.com/mozilla-mobile/fenix/pull/4914. 

This expression was incorrectly inverted:
https://github.com/mozilla-mobile/fenix/pull/4914/files#diffa6785c308c16739108b147e7862f3997R36

We want to hide the toolbar only if we're not in a custom tab to show the find in page query input. For custom tabs, the toolbar is on top and doesn't need to be hidden.

Ran into this when investigating https://github.com/mozilla-mobile/fenix/issues/5534 which now failed because of this. So, the UI tests would have caught this, but it's ignored right now because of intermittent failures I was trying to reproduce :).

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/FindInPageIntegration.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/FindInPageIntegration.kt
@@ -33,7 +33,9 @@ class FindInPageIntegration(
 
     override fun onLaunch(view: View, feature: LifecycleAwareFeature) {
         store.state.findCustomTabOrSelectedTab(sessionId)?.let { tab ->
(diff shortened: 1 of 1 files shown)


Amejia481:
@csadilek  Should we write a test for it? To avoid future regressions ?

csadilek:
@Amejia481 :). See description. There is a test that caught this, but it's `@Ignored` right now. I was trying to find out why it's failing intermittently on Bitbar on some device so we can activate it again, but that's a different issue: https://github.com/mozilla-mobile/fenix/issues/5534

Amejia481:
Great :)

PR #5914, opened 2019-10-09
Closes #5913: Add new sessionToken scope

### LINKED ISSUE #5913: Manage Account: declare sessionToken scope during FxA initialization
We need a new sessionToken scope in order to enable "Manage Account" functionality in Fenix's Account Settings.

For existing FxA users, introduction of this scope will be a non-issue, until they will need to interact with upcoming "Manage Account" functionality (a button, which opens up FxA web-content and passes it the sessionToken via webchannel). Once that happens, we will need to trigger a "need to re-authenticate" state. This is unfortunate, but the alternative is worse - triggering that state for everyone upon introduction of this scope. See a-c where this work is tracked: https://github.com/mozilla-mobile/android-components/issues/4703

grigoryk:


grigoryk:
Tested by installing Fenix w/o this scope, signing in to FxA, and then installing over Fenix that requests this new scope. Everything continues to behave as expected.

The rest will be done in 'Manage Account' issues (https://github.com/mozilla-mobile/fenix/issues/5915)

PR #5974, opened 2019-10-11
Ext Test For ImageButton

kglazko:
---

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/test/java/org/mozilla/fenix/ext/ImageButtonTest.kt
+++ b/app/src/test/java/org/mozilla/fenix/ext/ImageButtonTest.kt
@@ -1,0 +1,34 @@
+package org.mozilla.fenix.ext
+
+import kotlinx.coroutines.ObsoleteCoroutinesApi
+import mozilla.components.support.test.robolectric.testContext
+import org.mozilla.fenix.TestApplication
+import org.junit.Test
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertTrue
+import org.junit.Assert.assertEquals
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.annotation.Config
+import android.widget.ImageButton
+
+@ObsoleteCoroutinesApi
+@RunWith(RobolectricTestRunner::class)
+@Config(application = TestApplication::class)
+class ImageButtonTest {
+    val imageButton = ImageButton(testContext)
+
+    @Test
+    fun `Hide and disable`() {
+        imageButton.hideAndDisable()
+        assertFalse(imageButton.isEnabled)
+        assertEquals(4, imageButton.visibility)
+    }
+
+    @Test
+    fun `Show and enable`() {
+        imageButton.showAndEnable()
+        assertTrue(imageButton.isEnabled)
+        assertEquals(0, imageButton.visibility)
+    }
+}


app/src/test/java/org/mozilla/fenix/ext/ImageButtonTest.kt
```diff
@@ -0,0 +1,33 @@
+package org.mozilla.fenix.ext
```
mcomella: nit: license

app/src/test/java/org/mozilla/fenix/ext/ImageButtonTest.kt
```diff
@@ -0,0 +1,33 @@
...
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.annotation.Config
+import io.mockk.mockk
+import io.mockk.verify
+import android.widget.ImageButton
+
+@ObsoleteCoroutinesApi
+@RunWith(RobolectricTestRunner::class)
+@Config(application = TestApplication::class)
```
mcomella: fwiw, this is outside the scope of this PR but you can reduce duplication in the tests by merging the config setting into your test runner so each test looks like this:
```kotlin
@RunWith(FirefoxRobolectricTestRunner::class)
class SomeTest {
```

We do that in Fire TV: https://searchfox.org/mozilla-mobile/rev/2a57cc10d768832199bfd755d1e08fd9630baccc/firefox-tv/app/src/test/java/org/mozilla/tv/firefox/helpers/FirefoxRobolectricTestRunner.kt#26

app/src/test/java/org/mozilla/fenix/ext/ImageButtonTest.kt
```diff
@@ -0,0 +1,33 @@
...
+
+class ImageButtonTest {
```
mcomella: nit: there shouldn't be newline of whitespace between the class and the annotations

app/src/test/java/org/mozilla/fenix/ext/ImageButtonTest.kt
```diff
@@ -0,0 +1,33 @@
...
+    val mockIb: ImageButton = mockk(relaxed = true)
```
mcomella: nit: in Java/Kotlin convention, you generally avoid acronyms. I'd write `mockImageButton`. To be honest, since this is an image button test, `imageButton` is also appropriate. I haven't found the `mock` or `spy` prefixes to be all that useful and they're redundant to the type system, which already prevents you from making the wrong calls.

app/src/test/java/org/mozilla/fenix/ext/ImageButtonTest.kt
```diff
@@ -0,0 +1,33 @@
...
+
+    @Test
+    fun `Hide and disable`() {
+        mockIb.hideAndDisable()
+        verify { mockIb.setVisibility(4) }
```
mcomella: You should use the View constant here and in the other function.

app/src/test/java/org/mozilla/fenix/ext/ImageButtonTest.kt
mcomella: This imageButton is created once for the entire class and re-used for every test. This is a problem because this button is stateful and the order the tests are run in can affect whether they pass or not (e.g. in this case, whether or not a certain method was called will persist between tests). You should re-init it for every test using a pattern like this:
```kotlin
class ... {
    private lateinit var imageButton: ImageButton

    @Before
    fun setUp() {
        imageButton = mockk(relaxed =true)
    }
...
```

THat being said, mockk has a short-hand for this:
```kotlin
class ... {
    @MockK(relaxed = true) private lateinit var imageButton: ImageButton

    @Before
    fun setUp() {
        MockKAnnotations.init(this) // this lets the @Mockk annotations work
    }
...
```

This approach isn't simpler when there's only a single mock but it helps when you add more.

Note also that, while it doesn't _really_ matter, it's good to make these things private if they're not intended to be exposed outside of this class.

app/src/test/java/org/mozilla/fenix/ext/ImageButtonTest.kt
```diff
@@ -0,0 +1,33 @@
...
+        verify { mockIb.setEnabled(false) }
```
mcomella: Call verification usually indicates the test is coupled to the implementation and is a red flag (but sometimes necessary). In this case, this test would pass if this was the method implementation despite there being a bug:
```kotlin
fun hideAndDisable() {
    this.visibility = View.INVISIBLE
    this.isEnabled = false
    this.visibility = View.VISIBLE
}
```

A better practice is to test return values or the final state of an object, for which the test will stay the same even if the implementation changes. In this case, you should be able to inflate an ImageButton via robolectric (though I don't know how hard that is in practice) and verify the final state after the test. Here and in the other test.

ekager:
bors r=mcomella

Issue #4456, opened 2019-08-01
total_uri_count should be added to the metrics ping


boek:

AndiAJ:
Hi, I've just checked  on the latest Nightly Build #13040608 from 31/10 using a Google Pixel 3a XL (Android 9) and noticed the following:

• Performed 3 searches
The baseline ping 294b7f7a-c381-483f-b7b8-9cdcc0a3a1c0 shows "events.total_uri_count": 4
The metrics ping shows bb9612e8-a4c9-4e85-9cee-caebc871798a shows "metrics.total_uri_count": "0"
[Logcat](https://github.com/mozilla-mobile/fenix/files/3794365/3searches.txt)
[Glean dashboard](https://debug-ping-preview.firebaseapp.com/pings/aj-test311002)

• Performed 6 searches
The baseline 6b6cf012-661e-4fd3-b505-452b8565a0cc ping shows "events.total_uri_count": 7
The metrics ping shows 5f94bcc1-9239-4ed9-bb99-57c3750d1330 shows "metrics.total_uri_count": "0"
[Logcat](https://github.com/mozilla-mobile/fenix/files/3794366/6searches.txt)
[Glean Dashboard](https://debug-ping-preview.firebaseapp.com/pings/aj-test311003)

@sblatz please help me out with the following:
Why is the total_uri_count = to the numer of searches + 1? in both cases
Why the "metrics.total_uri_count" isn't displayed in the logs from Android Studio?

To generate the metrics ping I've force stopped Fenix, set the date to T+1 (01/11) and set the time to 03:59AM.  At around 4AM the metrics ping is generated and sent.
Based on the "client_id" & the "first_run_date" in both cases the metrics is properly generated.

Please review the above. :blush:
I'll remove the QA needed label until further clarifications.

sblatz:
>  Why is the total_uri_count = to the numer of searches + 1? in both cases

I think @Baron-Severin may be able to answer this better?

> Why the "metrics.total_uri_count" isn't displayed in the logs from Android Studio?

Investigating this one.

Baron-Severin:
> > Why is the total_uri_count = to the numer of searches + 1? in both cases
> 
> I think @Baron-Severin may be able to answer this better?

I don't actually have any context on this one.  I noticed [a related bug](https://github.com/mozilla-mobile/fenix/issues/6126), but don't know anything else about this ping.

sblatz:
@boek maybe? I think he was the one who added this :)

[...]

PR #6003, opened 2019-10-14
For #4456: Adds total_uri_count to metrics core ping

sblatz:
---

### Pull Request checklist

### After merge

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -307,6 +307,11 @@ class Settings private constructor(
         default = 0
     )
 
+    var totalUriCount by longPreference(
+        appContext.getPreferenceKey(R.string.pref_key_total_uri),
+        default = 0
```
sblatz: Is there a security risk here? We don't normally store telemetry data in prefs. 😬

sblatz:

-How many URI's does a user navigate to?

- We want to know how many sites a user visits in order to better understand engagement with the product.

* Currently we track this event in Glean every time a user enters a URI, however we don't get this in the core ping so we cannot query it in as useful of a way.

* All data is Category 2.

Until 03/01/2020

* Glean / Amplitude

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
liuche: I don't think there's a security risk in this information here - number of urls opened is not PII or anything else.

liuche: Do we ever reset this number? It's not clear to me that we do, and if we don't, this number will only be useful if we normalize it (e.g. 100 urls / days opened).

sblatz:
@liuche:

> Do we ever reset this number? It's not clear to me that we do, and if we don't, this number will only be useful if we normalize it (e.g. 100 urls / days opened).

When **should** we reset this? Do we reset it currently in any way for the other metric? My understanding is that it's a counter that always grows. @boek 

Looks like Chenxia's suggestion is to do this: 
> Shouldn't we set this count to 0 in preferences, when sending the ping?

Are we wanting total_uri_count to be reset each session?

liuche:
Quick glance shows me 3 increments in that file: two of them are counters so we stop showing a hint after a certain number of times, which I agree with - that one won't grow unbounded.

The other counter is "number of time PB has been opened". I really do think that this one is also suspect - just having a huge number doesn't really make sense, and it's hard to understand what we could actually do with that number! Like would we do something different if we knew a user EVER opened PB 100x vs 10000x, vs opened PB **in a session** 2x vs 10x?

I'd argue that "total number of urls opened even" is much less useful of a metric than "number of urls opened in a session".

liuche:
@Baron-Severin also a related total_uri_count

liuche:
@fbertsch from a data perspective, is there an easy way to look at total_uri_count for the lifetime of the app vs. per session? Do you have any opinions on how much information each of those provides? (Would we need to normalize lifetime uri count for instance, and if so is that easy to do?)

Baron-Severin:
> @Baron-Severin also a related total_uri_count

It looks like this is sent from the same place as the code I was looking at, so unless I'm mistaken it will suffer from https://github.com/mozilla-mobile/fenix/issues/6126 as well.

fbertsch:
@liuche almost certainly we want `total_uri_count` for the session. We can always aggregate the data on our end across a user's history if we want to get it over their lifetime, but the specific use-case is to determine if the user was active on a given day (`total_uri_count > 5`).

liuche:
@sblatz I'd go with Frank's suggestion here! It sounds like it will both give us the information that is useful for data analysis, as well as be aggregated to give the total lifetime count.

sblatz:
@liuche I am now zeroing this value out inside of `setupApplication` which should happen on each new "session": https://github.com/mozilla-mobile/fenix/pull/6003/files#diff-3c4353a6227c0da83801f46757491524R67

PR #6038, opened 2019-10-15
ExtTest for BookmarkNode

kglazko:
---

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/test/java/org/mozilla/fenix/ext/BookmarkNodeTest.kt
+++ b/app/src/test/java/org/mozilla/fenix/ext/BookmarkNodeTest.kt
@@ -1,0 +1,110 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.ext
+
+import org.junit.Assert.assertEquals
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.mozilla.fenix.TestApplication
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.annotation.Config
+import mozilla.components.concept.storage.BookmarkNode
+import mozilla.components.concept.storage.BookmarkNodeType
+
+@RunWith(RobolectricTestRunner::class)
+@Config(application = TestApplication::class)
+class BookmarkNodeTest {
+
+    @Test
+    fun `GIVEN a bookmark node with children WHEN minusing a sub set of children THEN the children subset is removed and rest remains`() {
+        val bookMarkChild1 = newBookmarkNode("Child 1", 1, null)
+        val bookMarkChild2 = newBookmarkNode("Child 2", 2, null)
+        val allChildren = listOf(bookMarkChild1, bookMarkChild2)
+        val bookmarkNode = newBookmarkNode("Parent 1", 0, allChildren)
+        val subsetToSubtract = setOf(bookMarkChild1)
+        val expectedRemainingSubset = listOf(bookMarkChild2)
+        val bookmarkNodeSubsetRemoved = bookmarkNode.minus(subsetToSubtract)
+        assertEquals(expectedRemainingSubset, bookmarkNodeSubsetRemoved.children)
+    }
+
+    @Test
+    fun `GIVEN a bookmark node with children WHEN minusing a set of all children THEN all children are removed and empty list remains`() {
+        val bookMarkChild1 = newBookmarkNode("Child 1", 1, null)
+        val bookMarkChild2 = newBookmarkNode("Child 2", 2, null)
+        val allChildren = listOf(bookMarkChild1, bookMarkChild2)
+        val bookmarkNode = newBookmarkNode("Parent 1", 0, allChildren)
+        val setofAllChildren = setOf(bookMarkChild1, bookMarkChild2)
+        val bookmarkNodeAllChildrenRemoved = bookmarkNode.minus(setofAllChildren)
+        assertEquals(emptyList<BookmarkNode>(), bookmarkNodeAllChildrenRemoved.children)
+    }
+
+    @Test
+    fun `GIVEN a bookmark node with children WHEN minusing a set of non-children THEN no children are removed`() {
+        val bookMarkChild1 = newBookmarkNode("Child 1", 1, null)
+        val bookMarkChild2 = newBookmarkNode("Child 2", 2, null)
+        val allChildren = listOf(bookMarkChild1, bookMarkChild2)
+        val bookMarkChild3 = newBookmarkNode("Child 3", 3, null)
+        val setofNonChildren = setOf(bookMarkChild3)
+        val bookmarkNode = newBookmarkNode("Parent 1", 0, allChildren)
+        val bookmarkNodeNonChildrenRemoved = bookmarkNode.minus(setofNonChildren)
+        assertEquals(allChildren, bookmarkNodeNonChildrenRemoved.children)
+    }
+
+    @Test
+    fun `GIVEN a bookmark node with children WHEN minusing an empty set THEN no children are removed`() {
+        val bookMarkChild1 = newBookmarkNode("Child 1", 1, null)
+        val bookMarkChild2 = newBookmarkNode("Child 2", 2, null)
+        val allChildren = listOf(bookMarkChild1, bookMarkChild2)
+        val bookmarkNode = newBookmarkNode("Parent 1", 0, allChildren)
+        val bookmarkNodeEmptySetRemoved = bookmarkNode.minus(emptySet())
+        assertEquals(allChildren, bookmarkNodeEmptySetRemoved.children)
+    }
+
+    @Test
+    fun `GIVEN a bookmark node with an empty list as children WHEN minusing a set of non-children from an empty parent THEN an empty list remains`() {
+        val parentWithEmptyList = newBookmarkNode("Parent 1", 0, emptyList<BookmarkNode>())
+        val bookMarkChild1 = newBookmarkNode("Child 1", 1, null)
+        val bookMarkChild2 = newBookmarkNode("Child 2", 2, null)
+        val setofAllChildren = setOf(bookMarkChild1, bookMarkChild2)
+        val parentWithEmptyListNonChildRemoved = parentWithEmptyList.minus(setofAllChildren)
+        assertEquals(emptyList<BookmarkNode>(), parentWithEmptyListNonChildRemoved.children)
+    }
+
+    @Test
+    fun `GIVEN a bookmark node with null as children WHEN minusing a set of non-children from a parent with null children THEN null remains`() {
+        val parentWithNullList = newBookmarkNode("Parent 1", 0, null)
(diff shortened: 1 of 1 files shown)


app/src/test/java/org/mozilla/fenix/ext/BookmarkNodeTest.kt
```diff
@@ -0,0 +1,41 @@
...
+import org.robolectric.annotation.Config
+import mozilla.components.concept.storage.BookmarkNode
+import mozilla.components.concept.storage.BookmarkNodeType
+
+@RunWith(RobolectricTestRunner::class)
+@Config(application = TestApplication::class)
+class BookmarkNodeTest {
+
+    @Test
+    fun `minus`() {
```
mcomella: This test tests only one case: we're subtracting a subset of the children. However, there are a few other classes of problems:
- The minus children equals the parent's children so they're all removed
- The minus children is a has some parent's children and some not parent's children: ensure only the appropriate ones are removed.
- Minus children is independent of parent's children: nothing is removed
- Parents children is an empty list (expected: no crash & an empty list at the end?)
- Parents children is null (expected: no crash & null at the end?)
- Minus children is an empty set: nothing is removed.

These are potentially all worth testing too. tbh, I'm not sure it's worth the time. In fact, there are even more cases to consider which may or may not be worth the time:
- Does minus ensure the other attributes are unmodified?
- Do the other values in the object (E.g. a null `parentGuid`) modify the results of `minus`?

A big benefit of testing (especially as a dev writing your own tests) is to be forced to think through all of these different "classes of problems" and ensure the implementation is correct for these edge cases. As for what's actually worth writing, I usually start with:
- What you have (i.e. a happy path)
- The edge cases: null state, empty collections, etc.

I'd like to see this here as well :) but it's not strictly necessary: I don't know what else you have to test and so what your priorities are. Note that if you do do this, make sure to leverage helper functions to avoid duplication as much as possible. e.g. because you don't care about all the other values in `BookmarkNode`, you could conceivably have a `newBookmarkNode(name: ..., children...): BookmarkNode` function that greatly simplifies node creation.

mcomella: nit: I know this is common on the team but I don't like this test naming scheme – i.e. `minus`. It should describe what it's doing. I really like the [GIVEN WHEN THEN](https://martinfowler.com/bliki/GivenWhenThen.html) that we used on Fire TV: GIVEN <some initial app state> WHEN <some event happens> THEN <the expected result>. It encourages the tester to think through what _this specific test_ is testing and thus what this specific test is _not_ testing. Then the tester might cover a few more classes of problem (see my above comment).

app/src/test/java/org/mozilla/fenix/ext/BookmarkNodeTest.kt
```diff
@@ -0,0 +1,41 @@
...
+        val childrenNodes = listOf(bookMarkChild1, bookMarkChild2)
+
+        // Create a set of children nodes to pass to minus function
+        val childrenSubsetNodes = setOf(bookMarkChild1)
+
+        // Create a list of the remainders
+        val minusSubsetNodes = listOf(bookMarkChild2)
+
+        // Create the node to test
+        var bookmarkNode = BookmarkNode(nodeType, "11", "00", 0, "Testing", "testing.url", childrenNodes)
```
mcomella: nit: there's no need to use var here. You could just create a new variable instead of overriding and you're (usually) less likely to run into problems that way. Avoid mutability unless it's really the easiest way to express something (and think thrice about that :D).

app/src/test/java/org/mozilla/fenix/ext/BookmarkNodeTest.kt
```diff
@@ -0,0 +1,74 @@
...
+    val subsetToSubtract = setOf(bookMarkChild1)
+    val expectedRemainingSubset = listOf(bookMarkChild2)
+    val setofAllChildren = setOf(bookMarkChild1, bookMarkChild2)
+    val setofNonChildren = setOf(bookMarkChild3)
+    val emptyChildrenList: List<BookmarkNode> = emptyList()
+    val emptyChildrenSet: Set<BookmarkNode> = emptySet()
+    val bookmarkNode = BookmarkNode(nodeType, "11", "00", 0, "Testing", "testing.url", allChildren)
+
+    @Test
+    fun `PASS WHEN minusing a sub set of children THEN the children subset is removed`() {
```
mcomella: nit: PASS is unnecessary: sorry, that was an example of what's printed when running the tests, e.g. via a command line interface. Just `GIVEN ... WHEN ... THEN`

app/src/test/java/org/mozilla/fenix/ext/BookmarkNodeTest.kt
```diff
@@ -0,0 +1,74 @@
...
+@RunWith(RobolectricTestRunner::class)
+@Config(application = TestApplication::class)
+class BookmarkNodeTest {
+    val nodeType = BookmarkNodeType.ITEM
+    val bookMarkChild1 = newBookMarkNode("Child 1", 1, null)
+    val bookMarkChild2 = newBookMarkNode("Child 2", 2, null)
+    val bookMarkChild3 = newBookMarkNode("Child 3", 3, null)
+    val allChildren = listOf(bookMarkChild1, bookMarkChild2)
+    val subsetToSubtract = setOf(bookMarkChild1)
+    val expectedRemainingSubset = listOf(bookMarkChild2)
```
mcomella: nit: this is only used in a single test so it's definition should be moved into the test.

In general, you should try to colocate data creation to the test that uses it because, if you need to update the data in a way that will change depending on the test, it's much more complicated to do so if the data is defined globally. For example:
```
// Used by two tests
val bmNode = BookmarkNode(Type.WHATEVER, children = null)

// But the BookmarkNode type changes such that we need two values:
val bmNode1 = BookmarkNode(Type.NO_CHILDREN, children = null)
val bmNode2 = BookmarkNode(Type.NO_CHILDREN_WITH_PARENT, children = null)
```

This is a weird contrived example but the general idea is that the dependencies between the data and the tests that uses it is unclear so it's hard to change the data, if it needs to be updated for whatever reason. It's better to colocate the data. /ramble 😁

app/src/test/java/org/mozilla/fenix/ext/BookmarkNodeTest.kt
```diff
@@ -0,0 +1,74 @@
...
+    val setofAllChildren = setOf(bookMarkChild1, bookMarkChild2)
+    val setofNonChildren = setOf(bookMarkChild3)
+    val emptyChildrenList: List<BookmarkNode> = emptyList()
+    val emptyChildrenSet: Set<BookmarkNode> = emptySet()
```
mcomella: nit: I think `children = emptySet()` is just as readable, and cuts down on unnecessary lines, than `emptyChildrenSet`

app/src/test/java/org/mozilla/fenix/ext/BookmarkNodeTest.kt
```diff
@@ -0,0 +1,74 @@
...
+
+    @Test
+    fun `PASS WHEN minusing a set of non-children from a parent with null children THEN null remains`() {
+        val parentWithNullList = newBookMarkNode("Parent 1", 0, null)
+        val parentWithNullListNonChildRemoved = parentWithNullList.minus(setofAllChildren)
+        assertEquals(null, parentWithNullListNonChildRemoved.children)
+    }
+
+    private fun newBookMarkNode(name: String, id: Int, myChildren: List<BookmarkNode>?): BookmarkNode {
+        val createdBookmarkNode = BookmarkNode(nodeType, "11", "11", id, name, "mockurl.com", myChildren)
```
mcomella: nit: you could just `return BookmarkNode` or even:
```kotlin
private fun ... = BookmarkNode(
    nodeType,
    "11", ...
)
```

app/src/test/java/org/mozilla/fenix/ext/BookmarkNodeTest.kt
```diff
@@ -0,0 +1,84 @@
...
+    fun `GIVEN a bookmark node with children WHEN minusing a sub-set of children THEN the rest of the parents object should remain the same`() {
+        val resultBookmarkNode = bookmarkNode.minus(subsetToSubtract)
+        val restOfResult = resultBookmarkNode.copy(children = expectedRemainingSubset)
+        val restofOriginal = bookmarkNode.copy(children = expectedRemainingSubset)
+        assertEquals(restOfResult, restofOriginal)
+    }
+
+    private fun newBookmarkNode(name: String, id: Int, myChildren: List<BookmarkNode>?) = BookmarkNode(
+            BookmarkNodeType.ITEM,
+            "11",
```
mcomella: nit: when reading the code without an IDE (such as for these reviews), it's helpful to include the argument names: `value = ...`

Also, when you have multiple args in a row with the same type (e.g. strings), it's pretty easy to mess them up, which is another reason to add explicit names.

app/src/test/java/org/mozilla/fenix/ext/BookmarkNodeTest.kt
mcomella: Just a note that this was not addressed.

app/src/test/java/org/mozilla/fenix/ext/BookmarkNodeTest.kt
```diff
@@ -0,0 +1,84 @@
...
+
+    @Test
+    fun `GIVEN a bookmark node with children WHEN minusing a sub-set of children THEN the rest of the parents object should remain the same`() {
+        val resultBookmarkNode = bookmarkNode.minus(subsetToSubtract)
+        val restOfResult = resultBookmarkNode.copy(children = expectedRemainingSubset)
+        val restofOriginal = bookmarkNode.copy(children = expectedRemainingSubset)
+        assertEquals(restOfResult, restofOriginal)
+    }
+
+    private fun newBookmarkNode(name: String, id: Int, myChildren: List<BookmarkNode>?) = BookmarkNode(
```
mcomella: nit: -> `children`. I think `my` is redundant here and is very unconventional.

app/src/test/java/org/mozilla/fenix/ext/BookmarkNodeTest.kt
```diff
@@ -0,0 +1,84 @@
...
+    fun `GIVEN a bookmark node with null as children WHEN minusing a set of non-children from a parent with null children THEN null remains`() {
+        val parentWithNullList = newBookmarkNode("Parent 1", 0, null)
+        val parentWithNullListNonChildRemoved = parentWithNullList.minus(setofAllChildren)
+        assertEquals(null, parentWithNullListNonChildRemoved.children)
+    }
+
+    @Test
+    fun `GIVEN a bookmark node with children WHEN minusing a sub-set of children THEN the rest of the parents object should remain the same`() {
+        val resultBookmarkNode = bookmarkNode.minus(subsetToSubtract)
+        val restOfResult = resultBookmarkNode.copy(children = expectedRemainingSubset)
```
mcomella: nit: I don't think it'd be immediately obvious what's going on here so I think this could be clearer with a comment that explains we're pinning `children` to the same value so we can compare the rest.

kglazko:
bors r=mcomella

PR #6043, opened 2019-10-16
Update to AC 19.0.0-SNAPSHOT and use new toolbar API.

pocmo:
The change hasn't landed in AC yet. However in 18.0.0 we are going to land a toolbar refactoring (which is now using ConstraintLayout internally, yay!) and with that we are changing the API. This patch migrates Fenix to use that new API.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarView.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarView.kt
@@ -17,8 +17,8 @@ import kotlinx.android.synthetic.main.browser_toolbar_popup_window.view.*
 import mozilla.components.browser.domains.autocomplete.ShippedDomainsProvider
 import mozilla.components.browser.session.Session
 import mozilla.components.browser.toolbar.BrowserToolbar
+import mozilla.components.browser.toolbar.display.DisplayToolbar
 import mozilla.components.support.ktx.android.util.dpToFloat
-import mozilla.components.support.ktx.android.util.dpToPx
 import org.jetbrains.anko.dimen
 import org.mozilla.fenix.R
 import org.mozilla.fenix.customtabs.CustomTabToolbarMenu
@@ -54,7 +54,7 @@ class BrowserToolbarView(
     init {
         val isCustomTabSession = customTabSession != null
 
-        view.setOnUrlLongClickListener {
+        view.display.setOnUrlLongClickListener {
             val clipboard = view.context.components.clipboardHandler
             val customView = LayoutInflater.from(view.context)
                 .inflate(R.layout.browser_toolbar_popup_window, null)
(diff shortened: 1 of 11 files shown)


app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarView.kt
```diff
@@ -109,34 +109,40 @@ class BrowserToolbarView(
...
-
-                suggestionBackgroundColor = ContextCompat.getColor(
-                    container.context,
-                    R.color.suggestion_highlight_color
-                )
+                display.progressGravity = if (isCustomTabSession) {
+                    DisplayToolbar.Gravity.TOP
+                } else {
+                    DisplayToolbar.Gravity.TOP
+                }
```
cadeyrn: This looks wrong because there is identical code in the _if_ and in the _else_ branch. I guess it should be `DisplayToolbar.Gravity.BOTTOM` for custom tabs?

app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarView.kt
```diff
@@ -109,34 +109,40 @@ class BrowserToolbarView(
...
+
+                display.onUrlClicked = {
                     interactor.onBrowserToolbarClicked()
                     false
                 }
 
-                browserActionMargin = browserActionMarginDp.dpToPx(resources.displayMetrics)
-
-                urlBoxView = if (isCustomTabSession) null else urlBackground
-                progressBarGravity = if (isCustomTabSession) PROGRESS_BOTTOM else PROGRESS_TOP
```
cadeyrn: PROGRESS_BOTTOM and PROGRESS_TOP are still in the companion object but seems to be unused now.

app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarView.kt
pocmo: Thank you! Updated.

app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarView.kt
pocmo: Removed. :)

app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarView.kt
```diff
@@ -109,34 +109,40 @@ class BrowserToolbarView(
...
                     ThemeManager.resolveAttribute(R.attr.secondaryText, container.context)
                 )
+
+                display.colors = display.colors.copy(
+                    text = primaryTextColor,
+                    securityIconSecure = primaryTextColor,
+                    securityIconInsecure = primaryTextColor,
+                    menu = primaryTextColor,
+                    hint = secondaryTextColor,
+                    trackingProtection = primaryTextColor
```
Amejia481: We also have to indicate the `separator` color  it should be `?toolbarDivider`

pocmo: Thanks! Updated. :)

app/src/main/java/org/mozilla/fenix/components/toolbar/ToolbarIntegration.kt
```diff
@@ -60,18 +61,33 @@ class ToolbarIntegration(
...
+                    toolbar.display.icons = toolbar.display.icons.copy(
+                        emptyIcon = AppCompatResources.getDrawable(
+                            context,
+                            R.drawable.ic_bookmark_filled
+                        )!!,
+                        trackingProtectionTrackersBlocked = lottieDrawable,
+                        trackingProtectionNothingBlocked = AppCompatResources.getDrawable(
                             context,
                             R.drawable.ic_tracking_protection_enabled
+
```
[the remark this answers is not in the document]
pocmo: Thanks. Fixed.

PR #6046, opened 2019-10-16
Update to Mozilla Android Components 18.0.0-SNAPSHOT.

pocmo:
We haven't updated in a while and therefore some fixes didn't make it into Fenix (Nightly) yet.


pocmo:
Looks like a test needs to be updated.

pocmo:
Test updated and taskcluster is ✅

PR #6084, opened 2019-10-17
Closes #6031 - Wrap waits with assert check; adjust timer check

### LINKED ISSUE #6031: [Bug] UI tests continue with false positive on mDevice.wait(Until
Currently for UI tests, we are using `mDevice.wait(Until` to wait until a UI element is present.  However, this operation by itself will only wait until the object is found or until the end of the wait period specified.  Without an assert to determine if the element was found or not, the test will proceed with a false positive result.

per @npark-mozilla, mDevice.wait will not fail the test, it’ll only time out. you need to wrap it with Assert to get something to fail, see: [UIDevice](https://developer.android.com/reference/android/support/test/uiautomator/UiDevice.html#wait(android.support.test.uiautomator.SearchCondition%3CR%3E,%20long))

We should use the return value, returns: The final result returned by the condition, or null if the condition was not met before the timeout.

AaronMT:
See #6031

- Adjust waits with assert check
- Adjusts timer to speed things up
- Test fix and cleanup

### DIFF
--- a/app/src/androidTest/java/org/mozilla/fenix/syncintegration/SyncIntegrationTest.kt
+++ b/app/src/androidTest/java/org/mozilla/fenix/syncintegration/SyncIntegrationTest.kt
@@ -27,6 +27,7 @@ import androidx.test.uiautomator.Until
 import org.hamcrest.Matchers.allOf
 import org.mozilla.fenix.R
 import org.mozilla.fenix.helpers.TestAssetHelper
+import org.junit.Assert.assertNotNull
 
 @Suppress("RECEIVER_NULLABILITY_MISMATCH_BASED_ON_JAVA_ANNOTATIONS")
 class SyncIntegrationTest {
@@ -96,7 +97,7 @@ class SyncIntegrationTest {
     }
 
     fun tapOnContinueButton() {
-        val continueButton = mDevice.findObject(By.res("submit-btn"))
+        val continueButton = mDevice.findObject(By.res("org.mozilla.fenix.debug:id/submit-btn"))
         continueButton.clickAndWait(Until.newWindow(), TestAssetHelper.waitingTime)
     }
 
@@ -110,7 +111,7 @@ class SyncIntegrationTest {
     }
 
     fun tapOnSignIn() {
-        mDevice.wait(Until.findObjects(By.text("Sign in")), TestAssetHelper.waitingTimeShort)
+        assertNotNull(mDevice.wait(Until.findObjects(By.text("Sign in")), TestAssetHelper.waitingTimeShort))
         // Let's tap on enter, sometimes depending on the device the sign in button is
         // hidden by the keyboard
         mDevice.pressEnter()
@@ -132,7 +133,7 @@ class SyncIntegrationTest {
     }
 
     fun seeBookmark() {
-        mDevice.wait(Until.findObjects(By.text("Bookmark")), TestAssetHelper.waitingTime)
+        assertNotNull(mDevice.wait(Until.findObjects(By.text("Bookmark")), TestAssetHelper.waitingTime))
         val bookmarkButton = mDevice.findObject(By.text("Bookmark"))
         bookmarkButton.click()
     }
--- a/app/src/androidTest/java/org/mozilla/fenix/ui/TabbedBrowsingTest.kt
+++ b/app/src/androidTest/java/org/mozilla/fenix/ui/TabbedBrowsingTest.kt
@@ -9,6 +9,7 @@ import androidx.test.uiautomator.By
 import androidx.test.uiautomator.UiDevice
 import androidx.test.uiautomator.Until
 import okhttp3.mockwebserver.MockWebServer
+import org.junit.Assert.assertNotNull
 import org.junit.After
 import org.junit.Before
 import org.junit.Rule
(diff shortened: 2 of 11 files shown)


app/src/androidTest/java/org/mozilla/fenix/ui/TabbedBrowsingTest.kt
```diff
@@ -74,7 +72,6 @@ class TabbedBrowsingTest {
         }.openHomeScreen { }
 
         homeScreen {
-            mDevice.wait(Until.findObject(By.res("R.id.item_tab")), TestAssetHelper.waitingTime)
```
npark-mozilla: I think most of the deleted mDevice.waits might be all necessary.  The failure I see in Firebase seems to be because it's trying to search for the test result too fast

app/src/androidTest/java/org/mozilla/fenix/ui/robots/NavigationToolbarRobot.kt
```diff
@@ -28,7 +29,6 @@ class NavigationToolbarRobot {
         val mDevice = UiDevice.getInstance(InstrumentationRegistry.getInstrumentation())
 
         fun enterURLAndEnterToBrowser(url: Uri, interact: BrowserRobot.() -> Unit): BrowserRobot.Transition {
-            mDevice.wait(Until.findObject(By.text("Search or enter address")), waitingTime)
```
npark-mozilla: for example without this wait the test might fail looking for urlBar?

AaronMT: For this one in particular we can't use "Search or enter address." as that text is only initially visible and this function is reused.

app/src/androidTest/java/org/mozilla/fenix/ui/TabbedBrowsingTest.kt
rpappalax: If I recall, in general, we only added mDevice waits when there was originally a timing issue of some kind, so I think No-Jun may be right.

AaronMT:
> @AaronMT is there a reason why the mDevice.waits were removed? In general, I don't think any of these were added randomly. Each one of them was originally added afaik to address timing issues

I didn't think they were necessary (one in particular above was failing a test) (aside it doesn't make sense to have them in the test but rather should be in the robot)

npark-mozilla:
@AaronMT So let's work on fixing intermittents on this PR until we make a one clean pass.  In this run, only the `shortcutButtonTest` failed, and when I looked at it, it has a good reason why it failed.
after `clickDuckDuckGoResult()`, we don't wait for anything before we call `verifyDuckDuckGoURL()`.  The DOM tree at the time of the failure shows that it was still in the search view, it didn't even open the web view at the time.  I think within `verifyDuckDuckGoURL()`, you have to put a `wait.Until()` to wait for the url text.

app/src/androidTest/java/org/mozilla/fenix/syncintegration/SyncIntegrationTest.kt
```diff
@@ -109,7 +110,7 @@ class SyncIntegrationTest {
     }
 
     fun tapOnSignIn() {
-        mDevice.wait(Until.findObjects(By.text("Sign in")), TestAssetHelper.waitingTime)
+        assertNotNull(mDevice.wait(Until.findObjects(By.text("Sign in")), TestAssetHelper.waitingTime))
```
Baron-Severin: It occurs to me that we could make this easier to reuse with an extension function.  Something like the following:

```kotlin
    /**
     * Blocks the test for [waitTime] miliseconds before continuing.
     *
     * Will cause the test to fail is the condition is not met before the timeout.
     */
    fun UiDevice.waitNotNull(
        searchCondition: SearchCondition<*>,
        waitTime: Long = TestAssetHelper.waitingTimeShort
    ) = assertNotNull(wait(searchCondition, waitTime))
```

Which would update L113 to look like `mDevice.waitNotNull(Until.findObjects(By.text("Sign in")))`, or (if we wanted to use a different timeout) `mDevice.waitNotNull(Until.findObjects(By.text("Sign in")), TestAssetHelper.waitingTimeShort)`

npark-mozilla: ^ @rpappalax fyi

app/src/androidTest/java/org/mozilla/fenix/ui/TabbedBrowsingTest.kt
AaronMT: Restored. Changed the selector to a different one because FindObject is unable to find the CardView resource name `R.id.item_tab` for whatever reason.

npark-mozilla: @AaronMT oh, the resource name is not `R.id.item_tab`for sure.  it's what that refers to, you can see it in uiautomatorviewer.  it's probably something like `org.mozilla.fenix.debug:id/item_tab` .  R.id.item_tab is just a reference to that string.  We can't use it like that.

app/src/androidTest/java/org/mozilla/fenix/syncintegration/SyncIntegrationTest.kt
AaronMT: We can open a new issue.

AaronMT:
Green @npark-mozilla (quick re-review) :)

npark-mozilla:
@AaronMT let's see how this one goes. :) and also let's create an issue based on Severin's suggestion and see how that goes as well, this might refactor our tests better.

PR #6155, opened 2019-10-21
Closes #5913: Add new sessionToken scope

### LINKED ISSUE #5913: Manage Account: declare sessionToken scope during FxA initialization

vladikoff:
Relanding https://github.com/mozilla-mobile/fenix/pull/5914

@grigoryk r?

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/BackgroundServices.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/BackgroundServices.kt
@@ -32,6 +32,7 @@ import mozilla.components.service.fxa.ServerConfig
 import mozilla.components.service.fxa.SyncConfig
 import mozilla.components.service.fxa.SyncEngine
 import mozilla.components.service.fxa.manager.FxaAccountManager
+import mozilla.components.service.fxa.manager.SCOPE_SESSION
 import mozilla.components.service.fxa.manager.SCOPE_SYNC
 import mozilla.components.service.fxa.sync.GlobalSyncableStoreProvider
 import mozilla.components.service.sync.logins.SyncableLoginsStore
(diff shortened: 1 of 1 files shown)


grigoryk:
Still seems to crash in @vladikoff 's local testing.

grigoryk:
@vladikoff any updates on this?

vladikoff:
> @vladikoff any updates on this?

We should probably wait for Firefox 71. 

@rfk do you have any other ideas? I am wondering if we can do something to "handle" crashes that happen if the OAuth `code` cannot be exchanged.

grigoryk:
I'm fine with waiting for 71 (a month from now).

rfk:
> I'm fine with waiting for 71 (a month from now).

Depends how we feel about the experience on ESR.

The reason this is a problem on 70 but not 71, is that 70 uses BrowserID assertions for authorizing OAuth codes, while 71 directly uses its sessionToken. Since BrowserID assertions don't tell you what sessionToken was used to generate them, we can't track the sessionToken info through the OAuth code grant and use it to generate a cloned session on the other side.

ow that danny has landed phase 2 of the auth/oauth server merger, it's possible that we could do something more clever on the server-side here. For example, I believe BrowserID assertions embed the originating deviceId, we might be able to use that to look up the sessionToken that should be associated with a particular OAuth code.

Alternately, when claiming the code, the server could detect the case where we can't generate a sessionToken and could remove it from the list of granted scopes, rather than failing the whole request. The client would need to be prepared to handle the case where it doesn't get granted all the scopes it requested, but OAuth clients should in theory be prepared for that at any time.

vladikoff:
@grigoryk Rebased!

Issue #5793, opened 2019-10-04
[Bug] Collections do not get saved intermittently in low memory x86 VM 


nojunpark:
## Steps to reproduce
This is seen only on Firebase UI Test, where the tests are running on a x86 emulator VM with (possibly) a lower spec, since the tests seem to run slower than local android emulator.

https://console.firebase.google.com/project/moz-fenix/testlab/histories/bh.66b7091e15d53d45/matrices/8583168991722205749 is where the test passes on one setup, but fails on another setup.  The UI Test scenario is:
1. Open a webpage, save to collection
2. Go homeview, open another webpage, save to a different collection
3. Go homeview, check both collections are created
### Expected behavior
Test should pass

### Actual behavior
It intermittently fails to create the second collection, and I'm pretty sure it's not a timing issue, because logcat shows the collection is requested to be saved.
[...]

PR #6160, opened 2019-10-21
Collection UI Test (disabled currently due to #5793) added 

npark-mozilla:
Add Collections UI test that tests:
- Create Collections
- Delete Collections
- Rename Collections
- Adding a tab to existing Collections

But due to #5793, it is currently disabled.  Once Collections refactoring is done by a-c, it should be enabled.

### DIFF
--- a/app/src/androidTest/java/org/mozilla/fenix/ui/CollectionTest.kt
+++ b/app/src/androidTest/java/org/mozilla/fenix/ui/CollectionTest.kt
@@ -4,25 +4,37 @@
 
 package org.mozilla.fenix.ui
 
+import android.net.Uri
+import androidx.test.espresso.Espresso.onView
+import androidx.test.espresso.NoMatchingViewException
+import androidx.test.espresso.assertion.ViewAssertions
+import androidx.test.espresso.matcher.ViewMatchers
 import androidx.test.platform.app.InstrumentationRegistry
+import androidx.test.uiautomator.By
 import androidx.test.uiautomator.UiDevice
+import androidx.test.uiautomator.Until
 import okhttp3.mockwebserver.MockWebServer
-import org.junit.Rule
+import org.junit.Ignore
 import org.junit.Before
 import org.junit.After
-import org.junit.Ignore
 import org.junit.Test
+import org.junit.Rule
+import org.mozilla.fenix.R
 import org.mozilla.fenix.helpers.AndroidAssetDispatcher
 import org.mozilla.fenix.helpers.HomeActivityTestRule
+import org.mozilla.fenix.helpers.TestAssetHelper
+import org.mozilla.fenix.helpers.click
+import org.mozilla.fenix.ui.robots.browserScreen
+import org.mozilla.fenix.ui.robots.homeScreen
+import org.mozilla.fenix.ui.robots.navigationToolbar
 
 /**
- *  Tests for verifying basic functionality of history
+ *  Tests for verifying basic functionality of tab collection
  *
  */
 
 class CollectionTest {
     /* ktlint-disable no-blank-line-before-rbrace */ // This imposes unreadable grouping.
-
     private val mDevice = UiDevice.getInstance(InstrumentationRegistry.getInstrumentation())
     private lateinit var mockWebServer: MockWebServer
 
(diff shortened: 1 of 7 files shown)


app/src/androidTest/java/org/mozilla/fenix/ui/CollectionTest.kt
```diff
@@ -42,31 +54,164 @@ class CollectionTest {
...
+        // Save the current page to the testcollection_1
+        navigationToolbar {
+        }.openThreeDotMenu {
+            // click save to collection menu item, type collection name
+            clickBrowserViewSaveCollectionButton()
+            org.mozilla.fenix.ui.robots.mDevice.wait(
+                Until.findObject(By.text("testcollection_1")),
+                TestAssetHelper.waitingTime)
+            Espresso.onView(ViewMatchers.withText("testcollection_1")).click()
+            Thread.sleep(5000)
```
rpappalax: Having this sleep looks brittle, but I think you may have explained why this was necessary? If you must use a sleep here, perhaps could we leave a comment why it's necessary?  Also, instead of leaving time integer here, could we instead define that with the other time constants?  https://github.com/mozilla-mobile/fenix/blob/master/app/src/androidTest/java/org/mozilla/fenix/helpers/TestAssetHelper.kt#L18-L19   Note: Eventually, we should probably move all of them to the more generic TestHelper since they're being used for things other than test assets.

app/src/androidTest/java/org/mozilla/fenix/ui/CollectionTest.kt
```diff
@@ -42,31 +54,164 @@ class CollectionTest {
...
+
+        // Save the current page to the testcollection_1
+        navigationToolbar {
+        }.openThreeDotMenu {
+            // click save to collection menu item, type collection name
+            clickBrowserViewSaveCollectionButton()
+            org.mozilla.fenix.ui.robots.mDevice.wait(
+                Until.findObject(By.text("testcollection_1")),
+                TestAssetHelper.waitingTime)
+            Espresso.onView(ViewMatchers.withText("testcollection_1")).click()
```
rpappalax: We don't have to use `Espresso.onView`, `onView` is enough.

app/src/androidTest/java/org/mozilla/fenix/ui/CollectionTest.kt
```diff
@@ -42,31 +54,164 @@ class CollectionTest {
...
+            Thread.sleep(5000)
+            mDevice.pressBack() // go to main page
+        }
+
+        // close currently opened tab
+        homeScreen {
+            verifyHomeScreen()
+            Espresso.onView(ViewMatchers.withId(R.id.close_tab_button)).click()
```
rpappalax: same for all these....

app/src/androidTest/java/org/mozilla/fenix/ui/CollectionTest.kt
```diff
@@ -42,31 +54,164 @@ class CollectionTest {
...
+        browserScreen {
+            verifyPageContent("Page content: 2")
+        }
     }
 
-    @Ignore
     @Test
-    fun OpenTabFromCollectionTest() {
-        // Open one tab from Collection in the Homescreen view
+    @Ignore("Temp disable test - see: https://github.com/mozilla-mobile/fenix/issues/5793")
```
rpappalax: I guess all these disabled tests work on physical device, just not on x86 (for now) at least til #5793 is resolved, true?

app/src/androidTest/java/org/mozilla/fenix/ui/CollectionTest.kt
```diff
@@ -42,31 +54,164 @@ class CollectionTest {
...
+
+            // Check for No collections caption
+            Espresso.onView(ViewMatchers.withText("No collections"))
+                .check(ViewAssertions
+                    .matches(ViewMatchers.withEffectiveVisibility(ViewMatchers.Visibility.VISIBLE)))
+        }
     }
 
-    @Ignore
+    // Open 2 webpages, and save each of them to a single collection
```
rpappalax: I like these explanatory / summary notes in places where it's not 100% clear what the test objective is.  We should do this more often so that it's easier for someone else to maintain the test in the future.

app/src/androidTest/java/org/mozilla/fenix/ui/CollectionTest.kt
npark-mozilla: Ah, this sleep was for investigation purpose, good catch.

app/src/androidTest/java/org/mozilla/fenix/ui/CollectionTest.kt
npark-mozilla: Yes they work on physical devices when ran locally.

PR #6220, opened 2019-10-23
Closes #6172 - Convert waitNotNull into extension function for UI tests

### LINKED ISSUE #6172: Convert waitNotNull into extension function for UI tests
https://github.com/mozilla-mobile/fenix/pull/6084#discussion_r337173560

AaronMT:
#6172 

From previous re-factor

> It occurs to me that we could make this easier to reuse with an extension function. Something like the following:
> 
> ```kotlin
>     /**
>      * Blocks the test for [waitTime] miliseconds before continuing.
>      *
>      * Will cause the test to fail is the condition is not met before the timeout.
>      */
>     fun UiDevice.waitNotNull(
>         searchCondition: SearchCondition<*>,
>         waitTime: Long = TestAssetHelper.waitingTimeShort
>     ) = assertNotNull(wait(searchCondition, waitTime))
> ```
> 
> Which would update L113 to look like `mDevice.waitNotNull(Until.findObjects(By.text("Sign in")))`, or (if we wanted to use a different timeout) `mDevice.waitNotNull(Until.findObjects(By.text("Sign in")), TestAssetHelper.waitingTimeShort)`

### DIFF
--- a/app/src/androidTest/java/org/mozilla/fenix/syncintegration/SyncIntegrationTest.kt
+++ b/app/src/androidTest/java/org/mozilla/fenix/syncintegration/SyncIntegrationTest.kt
@@ -27,7 +27,7 @@ import androidx.test.uiautomator.Until
 import org.hamcrest.Matchers.allOf
 import org.mozilla.fenix.R
 import org.mozilla.fenix.helpers.TestAssetHelper
-import org.junit.Assert.assertNotNull
+import org.mozilla.fenix.helpers.ext.waitNotNull
 
 @Suppress("RECEIVER_NULLABILITY_MISMATCH_BASED_ON_JAVA_ANNOTATIONS")
 class SyncIntegrationTest {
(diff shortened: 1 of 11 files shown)


app/src/androidTest/java/org/mozilla/fenix/syncintegration/SyncIntegrationTest.kt
```diff
@@ -111,7 +111,7 @@ class SyncIntegrationTest {
     }
 
     fun tapOnSignIn() {
-        assertNotNull(mDevice.wait(Until.findObjects(By.text("Sign in")), TestAssetHelper.waitingTimeShort))
+        mDevice.waitNotNull(Until.findObjects(By.text("Sign in")), TestAssetHelper.waitingTimeShort)
```
Baron-Severin: Optional nit: since `waitingTimeShort` is the default for this method, we could remove it from this call.  afaict, there are 6 other places where we could make that change as well.

Issue #5041, opened 2019-08-30
[Bug] Custom Tab does not restore itself when recreated


jonalmeida:
## Steps to reproduce
1. (Optional for easy reproducibility) Set Developer options > Background process limit to 'No background processes'.
2. Open a custom tab (and don't close it at the next step).
3. Do other things, use other memory-intensive apps.
4. Return to the custom tab.

### Expected behavior
1. The custom tab is loaded (or it shows https://github.com/mozilla-mobile/fenix/issues/2383)

### Actual behavior
1. The custom tab shows a completely blank page with a greyed out section on the bottom where the BrowserToolbar was.

### Device information

* Android device: Android 9 (Pixel 2)
* Fenix version: 9b633f7f (master)

> 1.0.1935 (Build #1)
> 📦: 10.0.1, a573b4116
> 🦎: 69.0-20190824140027

jonalmeida:
We have a fix for #2383 but I'm unable to verify if it works because of this other bug. The fix I'm trying to test: https://github.com/mozilla-mobile/reference-browser/pull/894

See this a-c issue for technical history of it: https://github.com/mozilla-mobile/android-components/issues/3773

This is also no related to #2383 since the Custom Tabs code never gets invoked when returning to the Custom Tab, so it seems to be something else unrelated to the lifecycle of the app.

softvision-miralobontiu:
Verified on the latest Nightly build from 2/24, and, with:
- Samsung Galaxy Note 8 (Android 9), Nokia 6 (Android 7.1.1), Sony Xperia Z5 Premium (Android 7.1.1) - I could not reproduce the issue,
- however, with Google Pixel (Android 10), after step 4, the custom tab showed a completely blank page. I've opened a custom tab in gmail, minimized it, opened Reddit, went back to gmail.

Please see the attached logcat from Google Pixel:

[logcat.txt](https://github.com/mozilla-mobile/fenix/files/4245238/logcat.txt)

I'll remove the qa-needed label for now.

PR #6282, opened 2019-10-25
Closes #5678: Selected session may render in ExternalAppBrowserActivity

### LINKED ISSUE #5678: [Bug] [Crash] Selected session may be rendered in ExternalAppBrowserActivity
We've fixed different bugs causing this crash in the past, but it is still happening (although at a lower rate than before). Definitely needs investigation and STRs:

https://sentry.prod.mozaws.net/operations/fenix-nightly/issues/6170288/?environment=fenixNightly

https://sentry.prod.mozaws.net/operations/fenix-nightly/issues/6191212/?environment=fenixNightly

It seems to happen for both regular and custom tabs (looking at the breadcrumbs). The root cause is always that we're trying to set the same session on multiple views. See:

https://github.com/mozilla-mobile/fenix/issues/360
https://github.com/mozilla-mobile/fenix/issues/575
https://github.com/mozilla-mobile/fenix/issues/897
https://github.com/mozilla-mobile/fenix/issues/4438
https://github.com/mozilla-mobile/fenix/issues/4486

csadilek:
This fixes the two variants of the "display already acquired" crash described in https://github.com/mozilla-mobile/fenix/issues/5678#issuecomment-546464188

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
+++ b/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
@@ -17,6 +17,7 @@ import androidx.appcompat.app.AppCompatActivity
 import androidx.appcompat.widget.Toolbar
 import androidx.lifecycle.lifecycleScope
 import androidx.navigation.NavDestination
+import androidx.navigation.NavDirections
 import androidx.navigation.fragment.NavHostFragment
 import androidx.navigation.ui.AppBarConfiguration
 import androidx.navigation.ui.NavigationUI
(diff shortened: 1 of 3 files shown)


csadilek:
Add work in progress, as I want to improve the test a bit, but ran out of time :).

csadilek:
OK, test updated so we can land this.

> I think that's the best quick fix to the issue we can land. Let's chat next week about some bigger changes that may prevent such a thing in the future.

Yes, let's do that. This will just fix the two crashes, but we also still have https://github.com/mozilla-mobile/fenix/issues/5041 to fix.

PR #6283, opened 2019-10-26
Add support for profiling with Nimbledroid from Gradle.

hawkinsw:
Modify the build.gradle file to support ND profiling from within gradle.

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -23,6 +23,50 @@ buildscript {
 
 plugins {
     id("io.gitlab.arturbosch.detekt").version("1.0.0-RC16")
+    id("com.nimbledroid.profiler").version("1.2.0")
+}
+
+// Placeholder for ND plugin configuration.
+//
+// The ND plugin expects this to be where gradle
+// stores configuration information. Variables
+// are set by the setNimbledroidConfiguration task.
+nimbledroid {
+}
+
+// Set the appropriate variables to perform the ND upload.
+//
+// Use the API key from the environment. Always upload a
+// nightly build.
+task setNimbledroidConfiguration {
+    doLast {
+        if (System.env.ND_APIKEY == null) {
+                throw new Exception("API key must be set in the environment (ND_APIKEY) for nimbledroidUpload")
+        }
+        nimbledroid.apiKey = "${System.env.ND_APIKEY}";
+        nimbledroid.apkFilename = "geckoNightly/fenixNightly/app-geckoNightly-armeabi-v7a-fenixNightly-unsigned.apk"
+    }
+}
+
+// Upload a nightly build to ND for profiling.
+//
+// Require that gradle do a nightly build before
+// attempting to upload it, obviously! Make sure that
+// the ND plugin is initialized properly
+// before uploading happens.
+task nimbledroidUpload {
+    ndUpload.mustRunAfter 'setNimbledroidConfiguration'
+    setNimbledroidConfiguration.mustRunAfter 'app:assembleFenixNightly'
+    dependsOn ndUpload, setNimbledroidConfiguration, 'app:assembleFenixNightly'
+}
+
+// Download the profile of the most recent build
+//
+// Make sure that the ND plugin is initialized
+// properly before profile downloading happens.
+task nimbledroidGetProfile {
+    ndGetProfile.mustRunAfter setNimbledroidConfiguration
+    dependsOn ndGetProfile, setNimbledroidConfiguration
 }
 
 allprojects {
--- a/build.gradle
+++ b/build.gradle
@@ -23,7 +23,7 @@ buildscript {
 
 plugins {
     id("io.gitlab.arturbosch.detekt").version("1.0.0-RC16")
-    id("com.nimbledroid.profiler").version("1.2.0")
+    id("com.nimbledroid.profiler").version("1.2.1")
 }
 
 // Placeholder for ND plugin configuration.


build.gradle
```diff
@@ -23,6 +23,50 @@ buildscript {
...
     id("io.gitlab.arturbosch.detekt").version("1.0.0-RC16")
+    id("com.nimbledroid.profiler").version("1.2.1")
+}
+
+// Placeholder for ND plugin configuration.
+//
+// The ND plugin expects this to be where gradle
+// stores configuration information. Variables
+// are set by the setNimbledroidConfiguration task.
+nimbledroid {
```
mcomella: I think this config should move to the `app/build.gradle` file because I don't think it makes sense to use nimbledroid on anything but the final app APKs.

If this config stays in this file, it should probably happen below `allProjects`, I think (that structure makes more sense to me).

build.gradle
```diff
@@ -23,6 +23,50 @@ buildscript {
...
+}
+
+// Set the appropriate variables to perform the ND upload.
+//
+// Use the API key from the environment. Always upload a
+// nightly build.
+task setNimbledroidConfiguration {
+    doLast {
+        if (System.env.ND_APIKEY == null) {
```
mcomella: fwiw, this doesn't follow the convention we already have for adding API keys: https://searchfox.org/mozilla-mobile/rev/574cb270fae8924647990be2be04ca223774438e/fenix/app/build.gradle#267

I'm not sure it's the best practice (e.g. it should be built into a re-usable function too!) but it's good to be consistent so that it's easier to change later (i.e. refactor them all to one function and then change the implementation of that function).

build.gradle
```diff
@@ -23,6 +23,50 @@ buildscript {
...
+// Upload a nightly build to ND for profiling.
+//
+// Require that gradle do a nightly build before
+// attempting to upload it, obviously! Make sure that
+// the ND plugin is initialized properly
+// before uploading happens.
+task nimbledroidUpload {
+    ndUpload.mustRunAfter 'setNimbledroidConfiguration'
+    setNimbledroidConfiguration.mustRunAfter 'app:assembleFenixNightly'
+    dependsOn ndUpload, setNimbledroidConfiguration, 'app:assembleFenixNightly'
```
mcomella: `app:assembleFenixNightly` will build every build variant with `FenixNightly` which is Beta and Nightly. I notice we only upload `geckoNightly` so should we just run `app:assembleGeckoNightlyFenixNightly` instead?

build.gradle
```diff
@@ -23,6 +23,50 @@ buildscript {
...
+    doLast {
+        if (System.env.ND_APIKEY == null) {
+                throw new Exception("API key must be set in the environment (ND_APIKEY) for nimbledroidUpload")
+        }
+        nimbledroid.apiKey = "${System.env.ND_APIKEY}";
+        nimbledroid.apkFilename = "geckoNightly/fenixNightly/app-geckoNightly-armeabi-v7a-fenixNightly-unsigned.apk"
+    }
+}
+
+// Upload a nightly build to ND for profiling.
```
mcomella: nit: I think info like this should go in the task description field. iirc:
```groovy
task nimbledroidUpload {
    description = "Upload a nightly build to Nimbledroid for profiling"
    ...
}
```

build.gradle
```diff
@@ -23,6 +23,50 @@ buildscript {
...
+//
+// Require that gradle do a nightly build before
```
mcomella: nit: This first sentence seems redundant to the definition of the task.

build.gradle
```diff
@@ -23,6 +23,50 @@ buildscript {
...
+// attempting to upload it, obviously! Make sure that
+// the ND plugin is initialized properly
+// before uploading happens.
+task nimbledroidUpload {
+    ndUpload.mustRunAfter 'setNimbledroidConfiguration'
+    setNimbledroidConfiguration.mustRunAfter 'app:assembleFenixNightly'
+    dependsOn ndUpload, setNimbledroidConfiguration, 'app:assembleFenixNightly'
+}
+
+// Download the profile of the most recent build
```
mcomella: nit: same, move to task `description`

build.gradle
```diff
@@ -23,6 +23,50 @@ buildscript {
...
+        }
+        nimbledroid.apiKey = "${System.env.ND_APIKEY}";
+        nimbledroid.apkFilename = "geckoNightly/fenixNightly/app-geckoNightly-armeabi-v7a-fenixNightly-unsigned.apk"
+    }
+}
+
+// Upload a nightly build to ND for profiling.
+//
+// Require that gradle do a nightly build before
+// attempting to upload it, obviously! Make sure that
```
mcomella: How does one make sure the ND plugin is initialized properly?

build.gradle
mcomella: nit: we should probably make the string a constant

mcomella:
I should also note that I'm not a core Fenix team member so I don't know how comfortable they are with me reviewing code but if all goes according to plan it shouldn't affect the app so maybe it'll be fine.

build.gradle
```diff
@@ -23,6 +23,50 @@ buildscript {
...
+//
+// Use the API key from the environment. Always upload a
+// nightly build.
+task setNimbledroidConfiguration {
+    doLast {
+        if (System.env.ND_APIKEY == null) {
+                throw new Exception("API key must be set in the environment (ND_APIKEY) for nimbledroidUpload")
+        }
+        nimbledroid.apiKey = "${System.env.ND_APIKEY}";
+        nimbledroid.apkFilename = "geckoNightly/fenixNightly/app-geckoNightly-armeabi-v7a-fenixNightly-unsigned.apk"
```
mcomella: I'm surprised this doesn't need the prefix, `app/build/outputs/apk/`

mcomella: Does uploading unsigned APKs work? You can't install unsigned builds on production Android devices so I just want to double check.

hawkinsw: Yup. They have to resign them at ND so that they can run on their test devices.

hawkinsw: The plugin prepends some part of the location, per the documentation.

build.gradle
hawkinsw: Definitely see your point here, but I'm wondering if there is a difference that is important: those things stored in files seem to be described as tokens and not keys. Are tokens supposed to be secret? It seems to me that we want to keep keys in environment variables (a best practice) rather than putting them anywhere on disk. 

What do you think?

build.gradle
hawkinsw: Good call! I will definitely move it!

build.gradle
mcomella: Those files are supposed to be kept secret: they're added in the CI environment and for local developers who need access to them. Maybe files are more convenient for local developers, especially those who may not know the best practices for setting up environment variables.

In any case, I'm not too opinionated here but I don't consider this key much different than the other ones so I'd lean towards consistency but I haven't thought through the implications.

build.gradle
```diff
@@ -23,6 +23,50 @@ buildscript {
 
 plugins {
     id("io.gitlab.arturbosch.detekt").version("1.0.0-RC16")
+    id("com.nimbledroid.profiler").version("1.2.1")
```
mcomella: There's an option to add a custom label to the uploads: I wonder if this is useful or if it'd just spam our "My Apps" screen.

build.gradle
```diff
@@ -23,6 +23,50 @@ buildscript {
...
+    }
+}
+
+// Upload a nightly build to ND for profiling.
+//
+// Require that gradle do a nightly build before
+// attempting to upload it, obviously! Make sure that
+// the ND plugin is initialized properly
+// before uploading happens.
+task nimbledroidUpload {
```
mcomella: I'm pretty sure we can modify the built-in tasks:
```groovy
task setNimbledroidConfiguration {
    ...

    // Maybe this isn't the best place for it: we can also set it at the top level.
    //
    // setNimbledroidConfiguration might be able to be replaced by `this`
    ndUpload.dependsOn setNimbledroidConfiguration, 'app:geckoNightlyFenixNightly'
    ndGetProfile.depends setNimbledroidConfiguration, 'app:geckoNightlyFenixNightly'
}
```

I think declaring it this way will mean we don't need to use `mustRunAfter` anymore too and it won't leave around two seemingly redundant tasks: `ndUpload` and `nimbledroidUpload`.

build.gradle
mcomella: nit: we might be able to set this one in the `nimbledroid` plugin block too though not sure that really gains us anything.

hawkinsw:
Abandoning this to pursue a different approach.

PR #6364, opened 2019-10-31
Use new WindowFeature API after browser-state migration 

csadilek:
This is preparing for incoming API breakage in A-C 20.0.0-SNAPSHOT once this A-C PR lands: https://github.com/mozilla-mobile/android-components/pull/4901


csadilek:
This is ready but we need to wait for snapshots to go out for this to be green.

csadilek:
Snapshots just went out, so we should be able to get this green now.

csadilek:
There's a problem with the latest snaphots. releng is investigating.

csadilek:
Force merging to fix master which wouldn't compile with latest snapshots otherwise. There's a UI test failure I can't repro locally and is entirely unrelated `deleteMultipleSelectionTest`

PR #6449, opened 2019-11-05
Fixes #4528 - Prevent share menu from jumping 

### LINKED ISSUE #4528: Feature request: don't push 'send to device' icons up when populating share menu
When the share menu is first opened after  a restart, the icons under  'send to device'  are above one empty row with a 'share a link' label. After the share menu is completely populated, the icons under 'send to device' are pushed up an extra row. This makes it annoying to click on one of those icons, because it moves before you can click on it. A good way to fix this would be to keep the icons under 'send to device' in place, and always have the 'share a link' section have two rows.

NotWoods:
Plus a bunch of docs and refactoring. 

- Converted `AppShareOption` into a sealed class, so that one option is an invisible app in order to display the correct height on the share sheet. In the future there could be "DirectShare" and "PWAShareTarget" subclasses too.
- Changed `AccountDevicesShareAdapter` to use the async `ListAdapter`.
- Replaced `SyncShareOption.Mobile` and `.Desktop` with `SyncShareOption.SingleDevice`, since all the properties are shared.

---

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,11 +10,12 @@ apply from: "$project.rootDir/automation/gradle/versionCode.gradle"
 apply plugin: 'androidx.navigation.safeargs.kotlin'
 apply plugin: 'com.google.android.gms.oss-licenses-plugin'
 
-import com.android.build.gradle.internal.tasks.AppPreBuildTask
+
+import com.android.build.OutputFile
 import org.gradle.internal.logging.text.StyledTextOutput.Style
 import org.gradle.internal.logging.text.StyledTextOutputFactory
+
 import static org.gradle.api.tasks.testing.TestResult.ResultType
-import com.android.build.OutputFile
 
 android {
     compileSdkVersion 28
@@ -412,6 +413,7 @@ dependencies {
     implementation Deps.androidx_navigation_fragment
     implementation Deps.androidx_navigation_ui
     implementation Deps.androidx_recyclerview
+    implementation Deps.androidx_lifecycle_livedata
     implementation Deps.androidx_lifecycle_runtime
     implementation Deps.androidx_lifecycle_viewmodel
     implementation Deps.androidx_core
--- a/app/src/main/java/org/mozilla/fenix/share/ShareViewModel.kt
+++ b/app/src/main/java/org/mozilla/fenix/share/ShareViewModel.kt
@@ -1,0 +1,162 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.share
+
+import android.app.Application
+import android.content.Context
+import android.content.Intent
+import android.content.pm.ResolveInfo
+import android.net.ConnectivityManager
+import android.net.Network
+import android.net.NetworkRequest
+import androidx.annotation.VisibleForTesting
+import androidx.annotation.WorkerThread
+import androidx.core.content.getSystemService
+import androidx.lifecycle.AndroidViewModel
+import androidx.lifecycle.LiveData
+import androidx.lifecycle.MutableLiveData
+import androidx.lifecycle.viewModelScope
+import kotlinx.coroutines.Dispatchers.IO
+import kotlinx.coroutines.launch
+import mozilla.components.concept.sync.DeviceCapability
+import mozilla.components.service.fxa.manager.FxaAccountManager
+import org.mozilla.fenix.ext.components
+import org.mozilla.fenix.share.listadapters.AppShareOption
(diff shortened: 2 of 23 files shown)


app/src/main/java/org/mozilla/fenix/share/listadapters/AccountDevicesAdapter.kt
```diff
@@ -23,25 +27,33 @@ class AccountDevicesShareAdapter(
...
+ */
 sealed class SyncShareOption {
     object Reconnect : SyncShareOption()
     object Offline : SyncShareOption()
     object SignIn : SyncShareOption()
     object AddNewDevice : SyncShareOption()
     data class SendAll(val devices: List<Device>) : SyncShareOption()
-    data class Mobile(val name: String, val device: Device) : SyncShareOption()
-    data class Desktop(val name: String, val device: Device) : SyncShareOption()
+    data class SingleDevice(val device: Device) : SyncShareOption()
```
jonalmeida: Consider if this still works well when we introduce other devices like [VR, TV, etc](https://searchfox.org/mozilla-mobile/source/android-components/components/concept/sync/src/main/java/mozilla/components/concept/sync/Devices.kt#113).

app/src/main/java/org/mozilla/fenix/share/listadapters/AppShareAdapter.kt
```diff
@@ -26,17 +29,24 @@ class AppShareAdapter(
...
 }
 
+/**
+ * Represents an app that can be shared to.
+ *
+ * @property name Name of the app.
+ * @property icon Icon representing the share target.
+ * @property packageName Package of the app.
+ * @property activityName Activity that will be shared to.
+ */
```
jonalmeida: Nice!

app/src/main/java/org/mozilla/fenix/share/ShareViewModel.kt
```diff
@@ -0,0 +1,162 @@
...
+import mozilla.components.concept.sync.DeviceCapability
+import mozilla.components.service.fxa.manager.FxaAccountManager
+import org.mozilla.fenix.ext.components
+import org.mozilla.fenix.share.listadapters.AppShareOption
+import org.mozilla.fenix.share.listadapters.SyncShareOption
+
+class ShareViewModel(application: Application) : AndroidViewModel(application) {
+
+    private val connectivityManager by lazy { application.getSystemService<ConnectivityManager>() }
+    private val fxaAccountManager by lazy { application.components.backgroundServices.accountManager }
```
jonalmeida: A bit unnecessary since the accountManager is lazy itself in the component.

app/build.gradle
```diff
@@ -411,6 +412,7 @@ dependencies {
     implementation Deps.androidx_navigation_fragment
     implementation Deps.androidx_navigation_ui
     implementation Deps.androidx_recyclerview
+    implementation Deps.androidx_lifecycle_livedata
```
jonalmeida: Why do we need to introduce LiveData for the share sheet? Is it just for caching the app list and separating the connectivity checks? The AccountManager already has a cache for the shareable devices, and the app list is a short-lived in-memory cache while you're in the ShareFragment. We probably move the connectivity checks to separate class as well.

app/src/main/java/org/mozilla/fenix/share/listadapters/AccountDevicesAdapter.kt
NotWoods: The intent here was to make it easier to introduce other devices

app/build.gradle
NotWoods: We already have LiveData, this just switches the KTX version instead.

I wanted to break out the app and device list code into a separate class for testing, and realized I was just recreating a view model. `LiveData` is a better representation instead of `Deferred` for the data, since the device list can be updated.

I previously checked with the Fenix team about using LiveData and it's fine to use it where appropriate.

app/src/main/java/org/mozilla/fenix/share/ShareViewModel.kt
```diff
@@ -0,0 +1,162 @@
...
+
+    private val devicesListLiveData = MutableLiveData<List<SyncShareOption>>(emptyList())
+    private val appsListLiveData = MutableLiveData<List<AppShareOption>>(emptyList())
+
+    @VisibleForTesting
+    internal val networkCallback = object : ConnectivityManager.NetworkCallback() {
```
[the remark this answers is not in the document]
NotWoods: In this case I'm not sure if that makes sense, as everything but `refreshDevicesAsync` depends on something from the outer class.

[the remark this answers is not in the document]
jonalmeida: Let's land this as-is, we can always refactor later. 🙂

Issue #5295, opened 2019-09-13
[Bug] [a11y] Unable to tap the Undo button in snackbars


sv-ohorvath:
## Steps to reproduce
1. Enable TalkBack.
2. Close a tab, or delete a bookmark, or other action that can be undone.
3. Try to tap undo.
### Expected behavior
The undo button is focused by TalkBack, users can double tap it.
### Actual behavior
The undo button is not focused by TalkBack, users can't tap it quickly enough.
### Device information

* Android device: Sony Xperia Z5 Premium (Android 7.1.1), Pixel (Android 9)
* Fenix version: 2.0-rc2

ekager:
Maybe we should implement a confirmation dialog instead of the undo snackbar for TalkBack users.

brampitoyo:
@ekager What do you think if we show the snackbar permanently for TalkBack users? The snackbar can only be dismissed if they tap/scroll anywhere outside of its area.

This way, TalkBack users have time to undo any decision, but the snackbar is still really easy to dismiss.

But does it make it _too_ easy to dismiss?

ekager:
You _can_ make a snackbar have length indefinite but I'm not sure how easy they are to dismiss in that way. We'd probably have to set some sort of touch listener/interceptor only for TalkBack users. It would definitely get a bit messy, but I suppose also creating dialogs instead of Snackbars at each of these locations would also.

ekager:
I think if I had to choose one I'd prefer to implement a confirmation dialog and take away undo snackbars for TalkBack users since it makes the action much more deliberate

brampitoyo:
@ekager I’ve just checked Material Design’s snackbar guidelines, and it says:

> Defaults to 4 seconds and can be set up to the maximum duration […]
> Maximum duration allowed […] Set to 10 seconds. If longer durations are necessary than this value consider creating a custom UI for your use case.

Sources:
* https://material.io/develop/ios/components/snackbars/api-docs/Classes/MDCSnackbarMessage.html#/c:objc(cs)MDCSnackbarMessage(py)duration
* https://material.io/develop/ios/components/snackbars/api-docs/Constants.html#/c:@MDCSnackbarMessageDurationMax

I agree with your idea of implementing a persistent-by-default confirmation dialogue. The question is, would it be annoying to be shown a blocking confirmation dialogue after every undo-able actions, or is it an expected behaviour for TalkBack users?

brampitoyo:
[...]

PR #6513, opened 2019-11-08
For #5295 [A11y] Unable to tap snackbar

kglazko:
---

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/utils/Undo.kt
+++ b/app/src/main/java/org/mozilla/fenix/utils/Undo.kt
@@ -10,6 +10,10 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.launch
 import org.mozilla.fenix.components.FenixSnackbar
+import android.app.AlertDialog
+import org.mozilla.fenix.R
+import android.content.Context
+import android.view.accessibility.AccessibilityManager
 import java.util.concurrent.atomic.AtomicBoolean
 
 internal const val UNDO_DELAY = 3000L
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/fenix/utils/Undo.kt
```diff
@@ -42,43 +47,55 @@ fun CoroutineScope.allowUndo(
...
-
-    // Launch an indefinite snackbar.
-    val snackbar = FenixSnackbar
-        .make(view, FenixSnackbar.LENGTH_INDEFINITE)
-        .setText(message)
-        .setAnchorView(anchorView)
-        .setAction(undoActionTitle) {
-            requestedUndo.set(true)
+    if (isAccessibilityEnabled) {
+        // CONFIRMATION DIALOG: TO-D0 CHECK IF THERE IS TALK BACK VIA ACCESSIBILITY THEN DO THIS
```
Baron-Severin: The code looks good, but these blocks have gotten big enough that they're getting a little tough to follow.  I think this would be easier if they were broken into helper functions, so we could get:
```kotlin
if (isAccessibilityEnabled) {
  showUndoDialog()
else {
  showUndoSnackbar()
}
```

app/src/main/java/org/mozilla/fenix/utils/Undo.kt
```diff
@@ -40,29 +45,60 @@ fun CoroutineScope.allowUndo(
...
+                operation.invoke()
+            }
         }
     }
+
+    // Check if accessibility is enabled
+    val isAccessibilityEnabled = accessibilityEnabled(view)
+
+    // Show either Snackbar or dialog based on result
+    if (isAccessibilityEnabled) showUndoDialog() else showUndoSnackbar()
```
Baron-Severin: This is a small project style thing, but would you remind reformatting this into:
```kotlin
if (isAccessibilityEnabled) {
  showUndoDialog()
} else {
  showUndoSnackbar()
}
```

app/src/main/java/org/mozilla/fenix/utils/Undo.kt
```diff
@@ -40,29 +45,60 @@ fun CoroutineScope.allowUndo(
...
-            operation.invoke()
+            if (!requestedUndo.get()) {
+                snackbar.dismiss()
+                operation.invoke()
+            }
         }
     }
+
+    // Check if accessibility is enabled
+    val isAccessibilityEnabled = accessibilityEnabled(view)
```
Baron-Severin: This line doesn't seem to help readabbility too much.  Maybe just call `accessibilityEnabled(view)` inside the if statement on 96?

app/src/main/java/org/mozilla/fenix/utils/Undo.kt
```diff
@@ -40,29 +45,60 @@ fun CoroutineScope.allowUndo(
...
+
+    // Show either Snackbar or dialog based on result
```
Baron-Severin: There's a maxim on code comments that says they should explain how, not what.  I don't think we need this comment, but it probably would be worth explaining why we're doing this in the first place.  Would you mind replacing this with something like `// It is difficult to use our Snackbars quickly enough with accessibility settings enabled, so in that case we show a dialog instead`?

app/src/main/java/org/mozilla/fenix/utils/Undo.kt
```diff
@@ -40,29 +45,60 @@ fun CoroutineScope.allowUndo(
...
     val requestedUndo = AtomicBoolean(false)
 
-    // Launch an indefinite snackbar.
-    val snackbar = FenixSnackbar
-        .make(view, FenixSnackbar.LENGTH_INDEFINITE)
-        .setText(message)
-        .setAnchorView(anchorView)
-        .setAction(undoActionTitle) {
-            requestedUndo.set(true)
+    // Launch an alert dialog for undo- for accessibility users
```
Baron-Severin: This is the kind of comment that could age poorly.  If we explain the accessibility problem down where we're conditioning on accessibility mode, hopefully it will be removed if that ever goes out of date.  However, it's not unreasonable to think that some other chunk of code could someday call `showUndoDialog` for a use case that has nothing to do with accessibility.  At that point, this comment will have rotted unintentionally.  I would remove this one entirely.

Commenting is _really hard._

app/src/main/java/org/mozilla/fenix/utils/Undo.kt
```diff
@@ -40,29 +45,60 @@ fun CoroutineScope.allowUndo(
...
                 onCancel.invoke()
             }
         }
+        val alert = dialogBuilder.create()
+        alert.show()
+    }
 
-    // If user engages with the snackbar, it'll get automatically dismissed.
-    snackbar.show()
+    // Launch an indefinite snackbar
```
Baron-Severin: I think we could probably get rid of this comment

app/src/main/java/org/mozilla/fenix/utils/Undo.kt
```diff
@@ -40,29 +45,60 @@ fun CoroutineScope.allowUndo(
...
+
+    // Show either Snackbar or dialog based on result
+    if (isAccessibilityEnabled) showUndoDialog() else showUndoSnackbar()
+}
+
+
+
+fun accessibilityEnabled(view:View): Boolean {
+    val am = view.context.getSystemService(Context.ACCESSIBILITY_SERVICE) as AccessibilityManager
+    return am.isEnabled
```
Baron-Severin: Good helper function, this really improves the clarity.

Normally I think it's good to narrow parameters as much as possible (passing `Context` is better than `View`, `AccessibilityManager` is better than `Context`).  There are a lot of good reasons for doing that, but in this instance the lookup is all the function is really doing.  So this is probably good as is!  No changes needed.

app/src/main/java/org/mozilla/fenix/utils/Undo.kt
```diff
@@ -40,29 +45,60 @@ fun CoroutineScope.allowUndo(
...
+            .setText(message)
+            .setAnchorView(anchorView)
+            .setAction(undoActionTitle) {
+                requestedUndo.set(true)
+                launch {
+                    onCancel.invoke()
+                }
+            }
+
+        // If user engages with the snackbar, it'll get automatically dismissed.
```
Baron-Severin: I think we could probably get rid of this comment

kglazko:
VideoToGif_GIF
Screenshot GIF

kglazko:
@lime124 See attached GIF above

lime124:
@AmyYLee can you check this out tomorrow? thanks!

AmyYLee:
> VideoToGif_GIF
> Screenshot GIF

This looks good to me @kglazko Thanks!

Issue #6196, opened 2019-10-22
Initialization of metrics unexpectedly takes 70ms


mcomella:
### User problem
Colin noticed the initialization of our metrics – Glean & Leanplum – takes about 70ms (on which device?) and that there may be wins to be had here.

### Impact
If we don't do this, we may be leaving behind <= 70ms of perf improvements on startup (cold, warm?).

### Acceptance criteria
TODO

hawkinsw:
I am resolving this as a result of @csadilek merge above. However, I am going to open a related issue for the performance problem of starting the Leanplum metrics.

PR #6575, opened 2019-11-13
Issue #6196: Make initialization of Glean metrics faster

hawkinsw:
setStartupMetrics is very expensive and does not need to be
done synchronously at the time the Glean Metrics Service is
initialized.

---

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt
@@ -6,9 +6,8 @@ package org.mozilla.fenix.components.metrics
 
 import android.content.Context
 import kotlinx.coroutines.Job
-import kotlinx.coroutines.MainScope
 import kotlinx.coroutines.launch
-import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.MainScope
 import mozilla.components.service.glean.BuildConfig
 import mozilla.components.service.glean.Glean
 import mozilla.components.service.glean.config.Configuration
@@ -416,7 +415,8 @@ class GleanMetricsService(private val context: Context) : MetricsService {
      * We need to keep an eye on when we are done starting so that we don't
      * accidentally stop ourselves before we've ever started.
      */
-    private lateinit var starter: Job
+    private lateinit var gleanInitializer: Job
+    private lateinit var gleanSetStartupMetrics: Job
 
     private val activationPing = ActivationPing(context)
 
@@ -431,7 +431,7 @@ class GleanMetricsService(private val context: Context) : MetricsService {
         // because it calls Google ad APIs that must be called *off* of the main thread.
         // These two things actually happen in parallel, but that should be ok because Glean
         // can handle events being recorded before it's initialized.
-        starter = MainScope().launch {
+        gleanInitializer = MainScope().launch {
             Glean.registerPings(Pings)
             Glean.initialize(context,
                 Configuration(channel = BuildConfig.BUILD_TYPE,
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt
```diff
@@ -475,7 +480,7 @@ class GleanMetricsService(private val context: Context) : MetricsService {
         /*
          * We cannot stop until we're done starting.
          */
-        runBlocking { starter.join(); }
+        runBlocking { gleanInitializer.join(); gleanSetStartupMetrics.join() }
```
csadilek: @mdboom so, I wasn't entirely sure and had to test, but looks like this can cause a dead lock (very unlikely though). We're blocking the main thread until other work on the main thread (`Glean.initialize`) is completed. Tested this by delaying initialization. This code was there before though so we don't need to block this PR on it, but should schedule a fix?

mdboom: Good catch, @csadilek.  Indeed, I don't think `getStartupMetrics` needs to run on the main thread.  Only `Glean.initialize` does.  So maybe put it on some other thread?

csadilek: @mdboom to be more precise: It's very unlikely because stop would have to happen *before* initialization is launched for this to deadlock. It can not happen *while* initialization is running, because both stop and initialization run on the main thread. Still wondering if we really need to do this though (wait for it to be started)?

csadilek: Looks like `setStartupMetrics` does mutate quite a bit of state that is not thread-safe. So, we can't just move it to a different thread. I think the change is fine as is, as the deadlocking case (stopping before starting) was there before. I am just wondering if we really need to join on the `gleanInitializer` and `gleanSetStartupMetric` when stopping? Isn't calling ` Glean.setUploadEnabled(false)` good enough?

mdboom: Ah, I see.  `Glean.setUploadEnabled()` is completely independent of initialization.  If Glean isn't initialized, it's just a flag.  If it were to be called in parallel with initialization, it would just take effect immediately after initialization (i.e. still resulting in no data being sent).  So it's probably completely safe just to remove this `runBlocking` altogether.

csadilek: @mdboom ah ok. In this case let's remove the `runBlocking` here and simply cancel the jobs if they are still running when we call `stop`.

hawkinsw:
profile

csadilek:
profile

app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt
hawkinsw: @mdboom Thanks for reviewing and working with @csadilek ! Can you take a look at this new version of the PR that we just pushed? Thanks!

hawkinsw:
Timeout while waiting for ND to complete profiling the application.

hawkinsw:
> Timeout while waiting for ND to complete profiling the application.

@csadilek It looks like it took longer than our default wait time to profile the application :-( I wait for 20 minutes, but I guess that I should wait longer. I will see how long it takes before this ND profile completes and then set that time as baseline going forward.

hawkinsw:
I just wanted to include a link to the related bugzilla issue that @mdboom filed where his great team is working on doing a complete refactor of the way that glean will handle initialization:

https://bugzilla.mozilla.org/show_bug.cgi?id=1595952

hawkinsw:
profile

hawkinsw:
Scenario | Status | Time (ms)
---------|--------|----------
Cold Startup | success | 1520
Enable private browsing | success | 500
Turn On Sync | fail | 0
Search or enter address | fail | 0
Add tab | fail | 0
More options | success | 380
Disable private browsing | success | 420
Enable private browsing | success | 430

PR #6588, opened 2019-11-14
4844 fix url elision

Baron-Severin:
---

### Pull Request checklist

### After merge

1. click on Show All Checks,

Screen Shot 2019-11-13 at 7 27 28 PM

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/collections/CollectionCreationView.kt
+++ b/app/src/main/java/org/mozilla/fenix/collections/CollectionCreationView.kt
@@ -27,7 +27,7 @@ import org.mozilla.fenix.R
 import org.mozilla.fenix.components.metrics.Event
 import org.mozilla.fenix.ext.components
 import org.mozilla.fenix.ext.increaseTapArea
-import org.mozilla.fenix.ext.urlToTrimmedHost
+import org.mozilla.fenix.ext.toShortUrl
 import org.mozilla.fenix.home.sessioncontrol.Tab
 import org.mozilla.fenix.home.sessioncontrol.TabCollection
 
@@ -268,7 +268,7 @@ class CollectionCreationView(
                 Tab(
                     tab.id.toString(),
                     tab.url,
-                    tab.url.urlToTrimmedHost(view.context),
+                    tab.url.toShortUrl(view.context.components.publicSuffixList),
                     tab.title
                 )
             }.let { tabs ->
--- a/app/src/test/java/org/mozilla/fenix/ext/StringTest.kt
+++ b/app/src/test/java/org/mozilla/fenix/ext/StringTest.kt
@@ -15,8 +15,8 @@ import org.mozilla.fenix.TestApplication
 import org.robolectric.RobolectricTestRunner
 import org.robolectric.annotation.Config
 
-const val puny = "xn--kpry57d"
-const val idn = "台灣"
+const val PUNYCODE = "xn--kpry57d"
+const val IDN = "台灣"
 
 @RunWith(RobolectricTestRunner::class)
 @Config(application = TestApplication::class)
@@ -133,7 +133,7 @@ class StringTest {
 
     @Test
     fun `should convert host to idn when calling shortURL`() {
-        "http://$puny.blah.com" shortenedShouldBecome "$idn.blah.com"
+        "http://$PUNYCODE.blah.com" shortenedShouldBecome "$IDN.blah.com"
     }
 
     @Test
@@ -174,7 +174,7 @@ class StringTest {
 
     @Test
     fun `should return idn etld for www-only non-etld`() {
-        "https://www.$puny/foo" shortenedShouldBecome idn
+        "https://www.$PUNYCODE/foo" shortenedShouldBecome IDN
     }
 
     @Test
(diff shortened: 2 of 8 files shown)


app/src/test/java/org/mozilla/fenix/ext/StringTest.kt
```diff
@@ -41,4 +54,188 @@ class StringTest {
...
+    }
+
+    @Test
+    fun testIsIPv4WithIPv6() {
+        assertThat("2001:db8::1 ".isIpv4()).isFalse()
+        assertThat("2001:db8:0:1:1:1:1:1".isIpv4()).isFalse()
+        assertThat("[2001:db8:a0b:12f0::1]".isIpv4()).isFalse()
+    }
+    // END test cases borrowed from FFTV
+
```
colintheshots: I know you went down the rabbit hole on this one, but I'm not sure if these tests are nearly evil and malicious enough. Maybe some of these are extra, but the URL encoding one could be a very real attack.

- What about URL encoding?
"https://user%40really.evil.domain%2F@mail.google.com" shortenedShouldBecome "really.evil.domain" (not mail.google.com or a crash, both of which are plenty bad)

- What about DWORD IP addresses?
"https://16843009" shortenedShouldBecome "1.1.1.1"

- What about octal IP addresses?
"https://00001.00001.00001.00001" shortenedShouldBecome "1.1.1.1"

- What about hex IP addresses?
"http://0x01010101" shortenedShouldBecome "1.1.1.1"
"http://0x100200401" shortenedShouldBecome "1.1.1.1"

Baron-Severin: Appreciate the extra cases!  IMO, I think we can safely push the last three into a follow up for prioritization (I'll add them as ignored cases, if that's alright with you), but that first one seems _extremely_ important.  I'll start working on fixing it now.

colintheshots: I agree with your assessment. The last three are a big can of worms and aren't nearly as important.

jonalmeida:
Any reason to not use [WebURLFinder](https://searchfox.org/mozilla-mobile/source/android-components/components/support/utils/src/main/java/mozilla/components/support/utils/WebURLFinder.kt)?

app/src/main/java/org/mozilla/fenix/collections/CollectionCreationView.kt
```diff
@@ -27,7 +27,7 @@ import org.mozilla.fenix.R
 import org.mozilla.fenix.components.metrics.Event
```
Baron-Severin: (moving this here to allow a comment chain)

> Any reason to not use [WebURLFinder](https://searchfox.org/mozilla-mobile/source/android-components/components/support/utils/src/main/java/mozilla/components/support/utils/WebURLFinder.kt) or modify it work for your use case?

@jonalmeida I wasn't aware of it.  I should have asked you guys if AC had anything that might help.

I was also trying to match existing behavior in desktop and FFTV as best I could, which meant copying some existing code (and rewriting JS).

I think in the long run here, it might be worth investing in a Rust component that handles these questions in a canonical way, and can be shared between all Moz products.  This is my _I'm-way-behind-schedule-but-we-need-something-now_ attempt.  I'm not particularly proud of it, but I'm trying.

jonalmeida: I peeked at the PR, after the title caught my interest, so I was curious if we had considered if there was something that didn't work with the WebURLFinder. Refactoring is always possible in the future (and not a one time thing!).

Really sorry, I didn't mean to sound like the effort wasn't worth it, you're doing great!

Baron-Severin: I don't think you did!  I think the correct approach here would be to write a rock-solid component that can be shared between desktop and mobile, because otherwise we're duplicating _very_ error-prone code.  I'm frustrated with my code, not with your comment.  Very sorry to have implied otherwise!

app/src/test/java/org/mozilla/fenix/ext/StringTest.kt
```diff
@@ -41,4 +52,209 @@ class StringTest {
...
+        "something" shortenedShouldBecome "something"
+        "http:" shortenedShouldBecome "http:"
+        "http::double-colon" shortenedShouldBecome "http::double-colon"
+        // The largest allowed port is 65,535
+        "http://badport:65536/" shortenedShouldBecome "http://badport:65536/"
+    }
+
+    @Test
+    fun `should convert host to idn when calling shortURL`() {
+        "http://$PUNYCODE.blah.com" shortenedShouldBecome "$IDN.blah.com"
```
colintheshots: The Punycode to IDN rules seem quite complicated. I think we're supposed to check that all of the IDN uses the same codepage or else show Punycode instead.

Baron-Severin:
> Thanks for fixing the crash!

Thanks for finding the crash!

app/src/test/java/org/mozilla/fenix/ext/StringTest.kt
Baron-Severin: They do seem pretty complex.  Do you mind if I leave a comment describing that change, but push implemention to a TODO?  IMO, it should go through prioritization and sizing.

Baron-Severin: Spoke with colintheshots about this and opened [a follow up issue](https://github.com/mozilla-mobile/fenix/issues/6642) for this change.

PR #6632, opened 2019-11-15
No Issue - Adds proguard rules for Lifecycle component ↩️

boek:
This is a potential fix for https://bugzilla.mozilla.org/show_bug.cgi?id=1596302

### DIFF
--- a/app/proguard-rules.pro
+++ b/app/proguard-rules.pro
@@ -108,3 +108,7 @@
     boolean isLimitAdTrackingEnabled();
 }
 -keep public class com.android.installreferrer.** { *; }
+
+# Keep Android Lifecycle methods
+# https://bugzilla.mozilla.org/show_bug.cgi?id=1596302
+-keep class androidx.lifecycle.** { *; }


app/proguard-rules.pro
```diff
@@ -107,4 +107,15 @@
...
     boolean isLimitAdTrackingEnabled();
 }
--keep public class com.android.installreferrer.** { *; }
\ No newline at end of file
+-keep public class com.android.installreferrer.** { *; }
+
+# Keep Android Lifecycle methods
+# https://bugzilla.mozilla.org/show_bug.cgi?id=1596302
+# https://twitter.com/rakshakhegde/status/875107392198856706
+-keep public class android.arch.lifecycle.** {
```
mdboom: Sanity check: Is this correct even though Glean imports it as `androidx.lifecycle`?

https://github.com/mozilla/glean/blob/master/glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/GleanLifecycleObserver.kt#L7

boek: Ah yes, good catch

Issue #2477, opened 2019-05-14
FNX3-16352 ⁃ Force enable zoom


vesta0:
## User Story
* As a user with low vision, I want to be able to override a website's request to prevent zooming in, so I can always read the content on a page. 

## Dependencies
* Needs GV API: https://bugzilla.mozilla.org/show_bug.cgi?id=1540615

### Acceptance Criteria
-I can zoom in on every page
-I can enable the "Always enable zoom" setting in my browser (for accessibility settings, because some sites can be problematic)

Lululalu:
It looks the same as #1372

vesta0:
Thank you @Lululalu ! I had totally forgotten that I had filed this before!

sblatz:
Let's make sure this is properly prioritized with GV

@bifleming, @cpeterso 

UX to decide how this toggle will look in the accessibility settings.

cpeterso:
> Let's make sure this is properly prioritized with GV

The GV zoom bug is a P2, meaning we will try to work on it (some time) in Q4. If you would like this zoom API delivered in early Q4 (like October), please let me know.

liuche:
@cpeterso if you could get ready for us earlier that would be really helpful!

cpeterso:
> @cpeterso if you could get ready for us earlier that would be really helpful!

OK. I increased the GV bug's priority rank so we will consider it for GV's October sprint.

cpeterso:
The GV bug fix will be in tomorrow's GV 71 Nightly, so I will remove this issue's `needs:gv` and `waiting` labels.

https://bugzilla.mozilla.org/show_bug.cgi?id=1540615#c10

liuche:
Adding this to the UX backlog since we wanted to do this earlier, and were waiting on this showing up in GV.

We need UX to decide how this toggle will look in the accessibility settings (because this could break some sites, so we don't want to turn it on by default).

boek:
This feature is behind a feature flag enabled on Nightly

softvision-miralobontiu:
@boek, this feature is not displayed in Settings - Accessibility so far.

I set the browser.ui.zoom.force-user-scalable pref from about:config to true, and it works. Pages like instagram.com, facebook.com can be zoomed after this pref was set to "true". When the pref is set to "false", those pages cannot be zoomed.

[...]

PR #6646, opened 2019-11-15
For #2477 - Removes the feature flag for Force Enable Zoom

boek:
Removes the feature flag for Force Enable Zoom


app/src/main/java/org/mozilla/fenix/components/Core.kt
```diff
@@ -1,4 +1,4 @@
-/* This Source Code Form is subject to the terms of the Mozilla Public
+e/* This Source Code Form is subject to the terms of the Mozilla Public
```
NotWoods: Looks like something `e`xtra slipped in here

Issue #1019, opened 2019-03-14
FNX-443 ⁃ Make search suggestions a user opt-in option in private mode


Lululalu:
I believe it will make the private mode a little more private. Even Chrome disables search suggestions on private mode. Fenix should follow, probably.

### User Story
As a user, I want the option to disable search suggestions in private mode only.

### Acceptance Criteria
* [ ] Move search suggestions setting to bottom of list
* [ ] Add setting to disable search suggestions in private mode while still getting search suggestions in regular mode. 
* [ ] Add contextual hint (see mock) for private mode search suggestions

Mocks: https://app.goabstract.com/share/e4322cfc-9036-4fbc-b417-5ac7df581b19?mode=design&sha=9fe9a001eabfa8ac0898d9f298791ead964230a5

- - -
EDIT (@sblatz):

* Search suggestions are disabled by default in private mode
* I get the option to toggle them on in the search screen (as an onboarding type message as seen in Focus)

Let's make sure we also consider how this conflicts with the current toggles in the search screen that are global settings.

vesta0:
@bbinto what do you think?

yoasif:
I think it'd be better to have a preference for private mode search engine, a la Brave with  the same toggle we have already for showing search suggestions working independently depending on the browsing mode.

bbinto:
How about we do this similar to what we've done in Focus?

First time, users enter a search term in private browsing, show a prompt to enable/disable search suggestions, and this choice can be toggled under settings > search.

![Screenshot_20190320-123840](https://user-images.githubusercontent.com/122897/54702534-715ba400-4b0d-11e9-8dba-67ec9d5bc802.png)

/cc @topotropic and @brampitoyo

joebloggs7657:
Definitely retain search suggestions or at most give an option to turn off. Private browsing is more so about keeping data private locally on the device rather than (ineffectively) ensurelibg private data handling by websites.

lime124:
removing the ux label - this is on our radar already as it's part of Q3 plans.

julienw:
As a +1, I was very surprised to get search suggestions in private mode, which means that what I type goes to the search engine. This isn't the case in Firefox/Android, and as far as I know there's even no way to turn it on in that browser.
I'd be super happy to be able to disable this in Fenix!

vesta0:
@lime124  can we re-use the Focus designs for this one? Attached above.

vesta0:
[...]

PR #6664, opened 2019-11-18
For #1019 - Implement show search suggestions in private option and info panel

mcarare:
---

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/search/SearchController.kt
+++ b/app/src/main/java/org/mozilla/fenix/search/SearchController.kt
@@ -14,11 +14,11 @@ import org.mozilla.fenix.BrowserDirection
 import org.mozilla.fenix.HomeActivity
 import org.mozilla.fenix.R
 import org.mozilla.fenix.components.metrics.Event
-import org.mozilla.fenix.ext.components
 import org.mozilla.fenix.ext.metrics
 import org.mozilla.fenix.ext.nav
-import org.mozilla.fenix.ext.searchEngineManager
 import org.mozilla.fenix.ext.settings
+import org.mozilla.fenix.ext.components
+import org.mozilla.fenix.ext.searchEngineManager
 
 /**
  * An interface that handles the view manipulation of the Search, triggered by the Interactor
@@ -68,6 +68,12 @@ class DefaultSearchController(
         store.dispatch(SearchFragmentAction.UpdateQuery(text))
         store.dispatch(SearchFragmentAction.ShowSearchShortcutEnginePicker(
             text.isEmpty() && context.settings().shouldShowSearchShortcuts
+        ))
+        store.dispatch(SearchFragmentAction.ShowSearchSuggestionsHint(
+            text.isNotEmpty() &&
+                    (context as HomeActivity).browsingModeManager.mode.isPrivate &&
+                    !context.settings().shouldShowSearchSuggestionsInPrivate &&
+                    !context.settings().showSearchSuggestionsInPrivateOnboardingFinished
         ))
     }
 
--- a/app/src/main/java/org/mozilla/fenix/search/SearchFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/search/SearchFragment.kt
@@ -81,13 +81,20 @@ class SearchFragment : Fragment(), BackHandler {
             requireComponents.search.provider.getDefaultEngine(requireContext())
         )
 
+        val showSearchSuggestions = if ((activity as HomeActivity).browsingModeManager.mode.isPrivate) {
+            requireContext().settings().shouldShowSearchSuggestions &&
+                    requireContext().settings().shouldShowSearchSuggestionsInPrivate
+        } else {
+            requireContext().settings().shouldShowSearchSuggestions
+        }
+
         searchStore = StoreProvider.get(this) {
             SearchFragmentStore(
                 SearchFragmentState(
                     query = url,
                     searchEngineSource = currentSearchEngine,
                     defaultEngineSource = currentSearchEngine,
-                    showSearchSuggestions = requireContext().settings().shouldShowSearchSuggestions,
+                    showSearchSuggestions = showSearchSuggestions,
                     showSearchShortcuts = requireContext().settings().shouldShowSearchShortcuts && url.isEmpty(),
                     showClipboardSuggestions = requireContext().settings().shouldShowClipboardSuggestions,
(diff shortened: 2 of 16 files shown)


app/src/main/java/org/mozilla/fenix/search/SearchController.kt
```diff
@@ -69,6 +70,12 @@ class DefaultSearchController(
         store.dispatch(SearchFragmentAction.ShowSearchShortcutEnginePicker(
             text.isEmpty() && context.settings().shouldShowSearchShortcuts
         ))
+        store.dispatch(SearchFragmentAction.ShowSearchSuggestionsHint(
+            text.isNotEmpty() &&
+                    !context.settings().shouldShowSearchSuggestionsInPrivate &&
+                    (context.asActivity() as HomeActivity).browsingModeManager.mode.isPrivate &&
```
NotWoods: Instead of casting context, it would be better to have `browsingModeManager` be passed in the constructor. We should also have this be the second condition instead of the third, since it just checks the memory value instead of possibly touching the disk.

app/src/main/java/org/mozilla/fenix/search/SearchFragment.kt
```diff
@@ -177,6 +189,55 @@ class SearchFragment : Fragment(), BackHandler {
             qrFeature.get()?.scan(R.id.container)
         }
 
+        val stubListener = ViewStub.OnInflateListener { _, inflated ->
+            val sharedPrefEdit = requireActivity().getSharedPreferences(
```
NotWoods: It would be better to expose a setter using `var` on `Settings` instead of getting a new reference to shared preferences.

app/src/main/java/org/mozilla/fenix/search/SearchFragment.kt
```diff
@@ -288,6 +350,27 @@ class SearchFragment : Fragment(), BackHandler {
         } else null
     }
 
+    private fun updateSearchSuggestionsHintVisibility(state: SearchFragmentState) {
+        if (state.showSearchSuggestionsHint) {
+            showSearchSuggestionsHint()
```
NotWoods: Since the helper functions are very similar, they should be collapsed into a single function. Kotlin view binding and `isVisible` can make this even shorter.

```kotlin
private fun updateSearchSuggestionsHintVisibility(state: SearchFragmentState) {
  view?.search_suggestions_onboarding?.isVisible = state.showSearchSuggestionsHint
}
```

app/src/main/java/org/mozilla/fenix/search/SearchFragmentStore.kt
```diff
@@ -48,6 +48,7 @@ data class SearchFragmentState(
     val searchEngineSource: SearchEngineSource,
     val defaultEngineSource: SearchEngineSource.Default,
     val showSearchSuggestions: Boolean,
+    val showSearchSuggestionsHint: Boolean,
```
NotWoods: nit: Update the KDocs for SearchFragmentState

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -253,6 +253,15 @@ class Settings private constructor(
...
     )
 
+    val shouldShowSearchSuggestionsInPrivate by booleanPreference(
+        appContext.getPreferenceKey(R.string.pref_key_show_search_suggestions_in_private),
+        default = false
+    )
+
+    val showSearchSuggestionsInPrivateOnboardingFinished by booleanPreference(
+            appContext.getPreferenceKey(R.string.pref_key_show_search_suggestions_in_private_onboarding),
+            false)
```
NotWoods: nit: Use named arguments for the boolean
```suggestion
        default = false
    )
```

app/src/main/res/xml/search_preferences.xml
```diff
@@ -37,5 +38,19 @@
             android:key="@string/pref_key_search_bookmarks"
             android:title='@string/preferences_search_bookmarks'
             app:iconSpaceReserved="false" />
+        <SwitchPreference
+        android:defaultValue="true"
```
NotWoods: nit: indentation

app/src/main/java/org/mozilla/fenix/search/SearchController.kt
mcarare: Actually, I removed `.asActivity() ` because it wasn't needed. The passed context is always HomeActivity, so IMO it is better to keep this single reference to it in the constructor. I have moved the condition up, thank you for the suggestion, I overlooked the issue.

app/src/main/java/org/mozilla/fenix/search/SearchFragment.kt
mcarare: I was using them separately to make it obvious that when showing the hint, the target view could be a `ViewStub` (setting visible on a stub also inflates it). I've collapsed them into a single function, but I kept the explicit `findViewById`, because synthetic import does not play well with viewstub. (it keeps the type of `ViewStub`, when actually the `ViewStub` becomes a `View`.)

app/src/main/java/org/mozilla/fenix/search/SearchController.kt
NotWoods: In that case, could you change the `context` parameter from `Context` to `HomeActivity` to make the dependency more explicit?

```kotlin
class DefaultSearchController(
    private val activity: HomeActivity,
    private val store: SearchFragmentStore,
    private val navController: NavController
) : SearchController {
  ...
```

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -253,6 +253,15 @@ class Settings private constructor(
...
     )
 
+    var shouldShowSearchSuggestionsInPrivate by booleanPreference(
+        appContext.getPreferenceKey(R.string.pref_key_show_search_suggestions_in_private),
+        default = false
+    )
+
+    var showSearchSuggestionsInPrivateOnboardingFinished by booleanPreference(
+        appContext.getPreferenceKey(R.string.pref_key_show_search_suggestions_in_private_onboarding),
+        default = false)
```
NotWoods: nit: Change the parenthesis location to match with other preferences
```suggestion
        default = false
    )
```

Issue #2754, opened 2019-05-22
FNX3-16557 ⁃ Add tab cards to share sheet


jonalmeida:
Parent: #331

The card on the top of the share sheet that shows the tab(s) being sent. On the right is what the share sheet currently looks like.

![](https://user-images.githubusercontent.com/1370580/58425209-3fb00d80-8067-11e9-8061-fa22e0217597.png)

jonalmeida:
Spoke with Nicole and Vesta offline, we're going to going to use the same styling of collections for the tab list.

colintheshots:
@topotropic 
1. How should this appear if I'm sharing a collection of series of tabs instead of a single tab?
2. What if I'm sharing a site from disk (bookmarks, history, collections, etc.) that I don't have a thumbnail handy for? It seems really inefficient to load the entire page only to present UI for sharing. Should we use some kind of placeholder?

liuche:
Will try to get UX feedback by first day of sprint.

liuche:
@topotropic @colintheshots how did this go? If it's ready to go just a reminder to move it into the Sprint Backlog.

topotropic:
Yes, this got resolved – here's the updated spec https://share.goabstract.com/33797d88-bf55-41fb-9b7e-9e90e38ef645?mode=design&sha=0086540706ce6705ac5fe791cdf6fef98c9baf9f

- For sharing Open tabs, a collection or anything within the library we use a list element with favicon; the header on the share view should be updated from "Send and Share" to "Share"
- For sharing a page from the page 3-dot menu, we decrease opacity of the scrim so that the page is still visible

sblatz:
@colintheshots are you planning on picking this up? :)

colintheshots:
I'm still on this. I have most of it implemented. I'm adding the finishing touches.

abodea:
Verified as fixed on the latest **`Nightly 190927 (#12700627)`** using Samsung Galaxy s10+(A9), Google Pixel 3a XL(A9).
Note that the behavior is the one presented in [this](https://github.com/mozilla-mobile/fenix/issues/2754#issuecomment-532317276) comment.

PR #6698, opened 2019-11-19
Use ShareData with ShareFragment 

NotWoods:
Preparing to integrate ShareFragment with web share

---

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/share/ShareFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/share/ShareFragment.kt
@@ -6,7 +6,6 @@ package org.mozilla.fenix.share
 
 import android.content.Context
 import android.os.Bundle
-import android.os.Parcelable
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
@@ -16,7 +15,6 @@ import androidx.lifecycle.ViewModelProvider.AndroidViewModelFactory
 import androidx.lifecycle.observe
 import androidx.navigation.fragment.findNavController
 import androidx.navigation.fragment.navArgs
-import kotlinx.android.parcel.Parcelize
 import kotlinx.android.synthetic.main.fragment_share.view.*
 import mozilla.components.feature.sendtab.SendTabUseCases
 import org.mozilla.fenix.R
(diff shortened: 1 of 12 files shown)


app/src/main/java/org/mozilla/fenix/share/ShareFragment.kt
```diff
@@ -71,7 +68,7 @@ class ShareFragment : AppCompatDialogFragment() {
         shareToAccountDevicesView =
             ShareToAccountDevicesView(view.devicesShareLayout, shareInteractor)
 
-        if (args.url != null && args.tabs == null) {
+        if (args.translucent) {
```
jonalmeida: Since you introduced a single concept `shareData`, couldn't this check just be `shareData.size == 1` instead of the caller needing to tell us that there is only one item?

jonalmeida:
Adding `do not land` until the a-c changes needed are in tomorrow's snapshot.

app/src/main/java/org/mozilla/fenix/share/ShareFragment.kt
NotWoods: I wasn't sure if it was always the case that 1 item == translucent, which is why I used the explicit flag. I can take out the flag and just check the list size.

jonalmeida: According to the comments there, that's what I understand from it. If you chose to keep the flag, let's update those comments and use another name instead of `translucent`. I was confused what it was referring to.

NotWoods: https://github.com/mozilla-mobile/fenix/issues/2754#issuecomment-532317276 indicates that the translucent scrim is for sharing from the browser fragment. Possible names: `overlay`, `fromBrowser`, `translucentScrim`?

jonalmeida: How about `showBrowser` or `showShareTabs` (depending on which way you want to flip the if-else)? Describing the action seems clearer to me than the thing it does internal, but that's just my opinion. 🤷‍♂

jonalmeida: Ah, I see now that the comment is in-accurate and based off an older implementation. So it confused the feature with the old impl? Sorry for making you do the comment clean up since it isn't related to your review.

Issue #5972, opened 2019-10-11
[Bug] Can't set as default browser on Android 5


sblatz:
### Acceptance Criteria
- [ ] Set as default always should work on Android 5
- [ ] Contact SUMO with the results of what we decided so they can update the Private Browsing Mode article to explain if this doesn't work on Android 5
- [ ] If this goes past the E3 timebox, then highlight it

### Details

On Android 5, when I press "always" to launch links in Fenix or Private Fenix, the preference is not remembered.

This is the root cause of #5870. 

I did a bit of investigation, and we **could** set the default on Preview v2.0. I believe adding the activity-alias for "always open links in private mode" silently breaks things on Android 5.

The issue can be seen below.

![set as default 2019-10-11 14_38_05](https://user-images.githubusercontent.com/4400286/66686581-c0496200-ec34-11e9-8355-e349afd86457.gif)

liuche:
We first want to test taking out the `activity-alias`.

Questions:
- How many Fennec users are on Android 5?
- Is this broken on all versions of 5?

Some options:
- Have a separate manifest that disables the `activity-alias` for Android 5 (and disabling the Private Mode launcher for those users)

boek:
This feature doesn't work on Android 5. The fix was to remove it until we set the minimum Android version to 6 or above.

AndiAJ:
Hi @boek , I've just checked this issue on the latest **Nightly Build #13310607 from 11/27** using the following devices:


► Have only Fenix installed (Chrome disabled) ✔️
Fenix is set as the default browser without any other actions, and work properly 
![20191127_152904](https://user-images.githubusercontent.com/51314259/69727688-b66bb780-112b-11ea-98f7-42f5f50292d0.gif)


► Having Chrome installed and afterwards install Fenix ❌
Seems that Chrome is set as default browser.
Can't actually disable it (Settings>Apps>Chrome>Clear defaults) the "Clear defaults" button seems to be disabled.
The only workaround is to disable Chrome and to have only Fenix installed.
![20191127_154628](https://user-images.githubusercontent.com/51314259/69729630-54ad4c80-112f-11ea-8d08-f8ddec76e98c.gif)

► Having only Firefox and Fenix installed (Chrome disabled) ❌
Firefox is set as default browser.
Can't disable it (Settings>Apps>Chrome>Clear defaults) the "Clear defaults" button seems to be disabled.
The only workaround is to uninstall Firefox and to have only Fenix installed.
[...]

PR #6702, opened 2019-11-20
For #5972 - Fixes bugs when launching in private mode on Android 5

boek:
There are some issues on Android 5 using activity-aliases so we took that out. In the process of testing opening new links in private mode we noticed the HomeFragment wasn't getting an onResume call when navigating from the browserFragment back to the homeFragment after opening an intent link. Moving that logic to onStart/onStop fixed it.


ekager:
Can you clean up ktlint pls :)

Issue #6679, opened 2019-11-18
[Performance] Leanplum metric initialization is very, very slow. 


hawkinsw:
This will track a solution to the problem that initialization of Leanplum is very, very slow and blocks startup. After the resolution of #6196, initialization of this component accounts for a very large part of the app's startup time.

hawkinsw:
See #6703

mcomella:
@hawkinsw I noticed this was in the "Waiting" column. Is this still blocked? If not, what are the remaining action items for this issue?

hawkinsw:
Blocked on Jonathan's work

mcomella:
@hawkinsw What is Jonathan's work? The delayed push initialization https://github.com/mozilla-mobile/fenix/issues/6730?

hawkinsw:
Yes

mcomella:
@hawkinsw Can you elaborate on how that impacts Leanplum?

Are there additional action items we can take here? For example, I wonder if we should file investigations into:
- Calling the Leanplum SDK ourselves (as you suggested)
- Forking the Leanplum SDK to remove or postpone certain unnecessary calls

hawkinsw:
I commented in #7448. I hope that answers your questions?

mcomella:
Unassigning Will: not actively being worked on so I want to minimize confusion. This is still blocked on delaying push init.

mcomella:
I think we're addressing Leanplum in its entirety in https://github.com/mozilla-mobile/fenix/issues/7448: closing.

PR #6703, opened 2019-11-20
Issue #6679: Move preference-fetching operations in `LeanplumMetricsService.start()` off the main thread

hawkinsw:
As a first step toward making it faster to start the release
metrics controller, move the Mozilla-product-related calls off
the main thread in `LeanplumMetrics`' `start` method.

---

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/metrics/LeanplumMetricsService.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/metrics/LeanplumMetricsService.kt
@@ -10,6 +10,12 @@ import com.leanplum.Leanplum
 import com.leanplum.LeanplumActivityHelper
 import com.leanplum.annotations.Parser
 import com.leanplum.internal.LeanplumInternal
+import kotlinx.coroutines.cancel
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.runBlocking
 import org.mozilla.fenix.BuildConfig
 import org.mozilla.fenix.ext.settings
 import java.util.UUID.randomUUID
@@ -52,6 +58,8 @@ class LeanplumMetricsService(private val application: Application) : MetricsServ
     }
 
     private val token = Token(LeanplumId, LeanplumToken)
+    lateinit var lpmsSetUserAttributeJob: Job
+    private val lpmsSetUserAttributeCoroutineScope = CoroutineScope(Dispatchers.IO)
 
     override fun start() {
         when (token.type) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/fenix/components/metrics/LeanplumMetricsService.kt
```diff
@@ -69,21 +74,33 @@ class LeanplumMetricsService(private val application: Application) : MetricsServ
...
-            "fennec_installed" to installedApps.contains(MozillaProductDetector.MozillaProducts.FIREFOX.productName),
-            "focus_installed" to installedApps.contains(MozillaProductDetector.MozillaProducts.FOCUS.productName),
-            "klar_installed" to installedApps.contains(MozillaProductDetector.MozillaProducts.KLAR.productName),
-            "fxa_signed_in" to application.settings().fxaSignedIn,
-            "fxa_has_synced_items" to application.settings().fxaHasSyncedItems,
-            "search_widget_installed" to application.settings().searchWidgetInstalled
-        ))
+        Leanplum.start(application)
+
+        lpmsSetUserAttributeJob = MainScope().launch {
```
NotWoods: I'm not sure if this runs off the main thread since you're using MainScope. I think here it would be better to have a scope property on the service and then launch with the IO Dispatcher.

```
class LeanplumMetricsService {
  private val scope = CoroutineScope()

  start() {
    scope.launch(IO) { ... }
  }

  stop() {
    runBlocking { job.join() }
    scope.cancel()
  }

**Comment by hawkinsw** [comment]
You are right! It's a bad commit message! Sorry. I did mean to leave it on
the main thread, just do it later!

**Comment by hawkinsw** [comment]
Oh geez. I think I confused myself. Sorry. I will look at it when I am back
at my desk! Thank you for the review!

Will

**Comment by hawkinsw** [comment]
@NotWoods Thanks again for the review. Give this one a look, if you'd like!

File: app/src/main/java/org/mozilla/fenix/components/metrics/LeanplumMetricsService.kt [hunk 2]
```diff
@@ -69,21 +77,34 @@ class LeanplumMetricsService(private val application: Application) : MetricsServ
...
-            "klar_installed" to installedApps.contains(MozillaProductDetector.MozillaProducts.KLAR.productName),
-            "fxa_signed_in" to application.settings().fxaSignedIn,
-            "fxa_has_synced_items" to application.settings().fxaHasSyncedItems,
-            "search_widget_installed" to application.settings().searchWidgetInstalled
-        ))
+        Leanplum.start(application)
+
+        lpmsSetUserAttributeJob = lpmsSetUserAttributeCoroutineScope.launch {
+            val installedApps = MozillaProductDetector.getInstalledMozillaProducts(application)
+            Leanplum.setUserAttributes(
```
csadilek: Looking at `Leanplum`, I don't see anything providing thread-safety or visibility guarantees. Why are we confident that it is safe to set the user attributes later on a different thread? To me it looks like we need to switch back to main before calling `setUserAttributes`.

If we fixed that we would still make the assumption that it's OK to start without those attributes and set them later. I don't know enough about Leanplum to be confident this works as expected.

hawkinsw: > Looking at `Leanplum`, I don't see anything providing thread-safety or visibility guarantees. Why are we confident that it is safe to set the user attributes later on a different thread? To me it looks like we need to switch back to main before calling `setUserAttributes`.

`setUserAttributesInternal`, what `setUserAttributes` ultimately uses to do its work, calls `RequestOld.post()` which gets an instance of a `RequestFactory` and uses that factory's `createRequest` which calls the constructor of `RequestOld` which uses code that seems to rely on the fact that it is on the main thread.

However, the code that relies on the fact that it is on the main thread is instantiating that must be instantiated on the main thread. That singleton, however, is *always* instantiated before this call because `Leanplum.start()` instantiates it as its first order of business. 

> 
> If we fixed that we would still make the assumption that it's OK to start without those attributes and set them later. I don't know enough about Leanplum to be confident this works as expected.

After a very detailed discussion with @csadilek IRL, we decided that that risk of setting user attributes asynchronously is *too* great when compared with any potential performance improvement. @csadilek and I came up with an alternate approach and we are going to consider that. 

I am closing this PR as a result.

PR #6712, opened 2019-11-20
WIP: Refactor BackgroundServices to remove push.

hawkinsw:
Make push its own component so that it can be accessed without having to instantiate every background service. This will result in a significant savings at startup. 

---

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
+++ b/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
@@ -103,7 +103,7 @@ open class FenixApplication : Application() {
             components.analytics.metrics.start()
         }
 
-        setupPush()
+        setupPushFeature()
 
         visibilityLifecycleCallback = VisibilityLifecycleCallback(getSystemService())
         registerActivityLifecycleCallbacks(visibilityLifecycleCallback)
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -166,11 +166,11 @@ open class FenixApplication : Application() {
         }
     }
 
-    private fun setupPush() {
-        // Sets the PushFeature as the singleton instance for push messages to go to.
-        // We need the push feature setup here to deliver messages in the case where the service
+    private fun setupPushFeature() {
+        // Sets the PushFeature as the singleton instance for pushFeature messages to go to.
```
grigoryk: nit: this edit seems unnecessary

grigoryk:
"requested changes" is mostly just me saying - I'd like to see a detailed response to https://github.com/mozilla-mobile/fenix/pull/6712#pullrequestreview-320418769 first.

I would approach this by by thinking through (and/or measuring) where we're encountering largest startup penalties, and what parts of background services we can load later than in onCreate.

My bet is that bulk of the cost here is in loading native a-s rust libraries. But, since we're in a megazord, instantiating push means you load everything anyway! So, this refactoring doesn't get us those cost savings.

And the rest of this class _should_ be quick to init eagerly - any real work that FxaAccountManager does will happen on a background thread. That's not entirely free and still costs us some startup of course, but is probably not a huge impact either (especially considering product benefits this enables).

Could you provide some before/after number? Thinking through this, my expectation is that it won't be a huge difference. I'd love to know if that's wrong, and what exactly is giving you large gains.

mcomella:
I'm looking to clean up the Fenix PR list by closing inactive PRs. @hawkinsw what's the status of this PR and what do you think we should we do with this PR? Assuming this is blocked longer than the short term, my suggestion would be to have a prioritized issue in our board (e.g. in the Waiting column) that links to the PR in case we want to reopen it.

hawkinsw:
Closing this. I think that this was left open from before @jonalmeida took over wondering on extracting the push service from LP from ...

If we need to reopen later, then we always can!

jonalmeida:
Yes, I haven't forgotten this, it's just been moved a little lower on my list at the moment. Looking to spend some time on getting this fixed. 🤞

PR #6738, opened 2019-11-21
Closes #4982: Migrate Gecko files before initializing engine.

### LINKED ISSUE #4982: Fennec migration: Clarify if Gecko profile will be re-used
It is unclear whether the Gecko profile will be re-used after paving over Fennec. This may affect whether we continue to use data like:

- Cookies
- Website Cache
- about:config flags (We were consider nuking them since there's an endless combination of configurations that we may not be able to support correctly).

pocmo:
This is the Fenix part of #4982 - for migrating Gecko internal files when updating from Fennec.

As part of this patch I restructured `onCreate()` in `FenixApplication` to call `setupInAllProcesses()` and `setupInMainProcessOnly()`.

I also moved more things to be only initialized in the main process. I do not think they are needed in other processes (which currently are only spawned by Gecko afaik).
- megazord setup and rust log initialization is only done in the main process now (@grigoryk can you confirm?)
- day/night theme is only setup in the main process now (others do not have any UI)
- Rx exception handler is only setup in the main process (Rx is only used in app code and that's only in the main process)
- strict mode is only setup in the main process (since we care about our own code here primarily).

### DIFF
--- a/app/src/fennecProduction/java/org/mozilla/fenix/MigratingFenixApplication.kt
+++ b/app/src/fennecProduction/java/org/mozilla/fenix/MigratingFenixApplication.kt
@@ -4,25 +4,32 @@
 
 package org.mozilla.fenix
 
-import mozilla.components.support.ktx.android.content.isMainProcess
+import kotlinx.coroutines.runBlocking
 import mozilla.components.support.migration.FennecMigrator
 
 /**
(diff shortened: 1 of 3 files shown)


pocmo:
Okay, all green. :)

Issue #5183, opened 2019-09-10
Initializing megazord in application start is slow


hawkinsw:
`setupMegazord()` being called from `FenixApplication.setupApplication` blocks startup and is slow. `FenixApplication.setupApplication` takes 290ms and `setupMegazord` itself takes 43ms (14.8%)

Within `setupMegazord`, `RustHttpConfig.setClient` is _the_ slow function. I have worked with a-c in the past on this issue and I _think_ that they already have a solution -- we will just need to incorporate it. See PR https://github.com/mozilla-mobile/fenix/pull/4391

mcomella:
@hawkinsw It looks like this merged: should this issue be closed?

hawkinsw:
Code merged.

PR #6770, opened 2019-11-23
Issue #5183: Initialize megazord asynchronously

hawkinsw:
Thanks to the work of the team at application-services, it is possible
to asynchronously initialize the Http client and logger used by megazord
Use that power to hide the time necessary to complete this initialization
within the time that it takes to warmup the gecko runtime.

---

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
+++ b/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
@@ -13,11 +13,12 @@ import androidx.annotation.CallSuper
 import androidx.appcompat.app.AppCompatDelegate
 import androidx.core.content.getSystemService
 import io.reactivex.plugins.RxJavaPlugins
+import kotlinx.coroutines.async
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.runBlocking
 import kotlinx.coroutines.Deferred
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.GlobalScope
-import kotlinx.coroutines.async
-import kotlinx.coroutines.launch
 import mozilla.appservices.Megazord
 import mozilla.components.concept.push.PushProcessor
 import mozilla.components.service.experiments.Experiments
@@ -75,19 +76,24 @@ open class FenixApplication : Application() {
 
     @CallSuper
     open fun setupInMainProcessOnly() {
-        setupMegazord()
+        run {
+            // Attention: Do not invoke any code from a-s in this scope.
+            val megazordSetup = setupMegazord()
 
-        // We want rust logging to go through the log sinks.
-        // This has to happen after initializing the megazord.
-        RustLog.enable()
+            setDayNightTheme()
+            registerRxExceptionHandling()
+            enableStrictMode()
 
-        setDayNightTheme()
+            // Make sure the engine is initialized and ready to use.
+            components.core.engine.warmUp()
 
-        registerRxExceptionHandling()
-        enableStrictMode()
-
-        // Make sure the engine is initialized and ready to use.
-        components.core.engine.warmUp()
+            // Just to make sure it is impossible for any application-services pieces
+            // to invoke parts of itself that require complete megazord initialization
+            // before that process completes, we wait here, if necessary.
+            if (!megazordSetup.isCompleted) {
+                runBlocking { megazordSetup.await(); }
+            }
+        }
 
         // We want to call this function as early as possible, but only once and
         // on the main process, as it uses Gecko to fetch experiments from the server.
@@ -205,11 +211,15 @@ open class FenixApplication : Application() {
      * - https://github.com/mozilla/application-services/blob/master/docs/design/megazords.md
      * - https://mozilla.github.io/application-services/docs/applications/consuming-megazord-libraries.html
      */
-    private fun setupMegazord() {
-        // Note: This must be called as soon as possible
+    private fun setupMegazord(): Deferred<Unit> {
+        // Note: Megazord.init() must be called as soon as possible ...
         Megazord.init()
-        // This (and enabling RustLog) may be delayed if needed for performance reasons
-        RustHttpConfig.setClient(lazy { components.core.client })
+
+        return GlobalScope.async(Dispatchers.IO) {
+            // ... but RustHttpConfig.setClient() and RustLog.enable() can be called later.
+            RustHttpConfig.setClient(lazy { components.core.client })
+            RustLog.enable()
+        }
     }
 
     override fun onTrimMemory(level: Int) {


app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -75,20 +76,21 @@ open class FenixApplication : Application() {
...
 
         // Make sure the engine is initialized and ready to use.
         components.core.engine.warmUp()
 
+        // Just to make sure it is impossible for any application-services pieces
+        // to invoke parts of itself that require complete megazord initialization
+        // before that process completes, we wait here, if necessary.
+        if (!megazordSetup.isCompleted) {
+            runBlocking { megazordSetup.await(); }
+        }
```
pocmo: So if we ever add any code, that may call into Rust AS code, between `setupMegazord()` and the `await` here, then this could potentially cause issues (if the setup has not been completed by that time), is that right? Unfortunately that is hard to even know and the compiler can't protect us against this mistake. But I assume we are seeing a performance gain that is worth the risk?

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -205,11 +207,15 @@ open class FenixApplication : Application() {
...
+    private fun setupMegazord(): Deferred<Unit> {
+        // Note: Megazord.init() must be called as soon as possible ...
         Megazord.init()
-        // This (and enabling RustLog) may be delayed if needed for performance reasons
-        RustHttpConfig.setClient(lazy { components.core.client })
+
+        return GlobalScope.async {
+            // ... but RustHttpConfig.setClient() and RustLog.enable() can be called later.
+            RustHttpConfig.setClient(lazy { components.core.client })
+            RustLog.enable()
```
pocmo: I wonder if you want to specify a different dispatcher here (e.g. `GlobalScope.async(Dispatchers.IO)`). Otherwise I think this will still happen on the same thread - just not sequentially?

hawkinsw: I am definitely not an expert on this, *but* according to the documentation, GlobalScope uses dispatchers that are [unconfined](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-global-scope/). That said, it does seem like a good idea to specify a launcher here that is explicitly off the the main thread. Do you agree?

NotWoods: Colin has also asked that we explicitly specify the dispatcher. I agree with @pocmo that IO should be used here.

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
hawkinsw: > So if we ever add any code, that may call into Rust AS code, between `setupMegazord()` and the `await` here, then this could potentially cause issues (if the setup has not been completed by that time), is that right? Unfortunately that is hard to even know and the compiler can't protect us against this mistake. 

That is absolutely correct. I think that the best we can do is put in some very, very explicit comments. See below.

> But I assume we are seeing a performance gain that is worth the risk?

We do see a performance gain, but *I* am not the person who can make the decision about risk/reward ratio! :-)

thomcc: It shouldn't be that bad. If the call has to do networking, that will fail until the networking part of this runs, and it won't perform any logging (well, megazord setup itself will probably log, but after that, not a peep), until the logging part. But in general failing to set up these things shouldn't be that bad.

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -205,11 +207,15 @@ open class FenixApplication : Application() {
      * - https://github.com/mozilla/application-services/blob/master/docs/design/megazords.md
      * - https://mozilla.github.io/application-services/docs/applications/consuming-megazord-libraries.html
      */
-    private fun setupMegazord() {
-        // Note: This must be called as soon as possible
+    private fun setupMegazord(): Deferred<Unit> {
+        // Note: Megazord.init() must be called as soon as possible ...
         Megazord.init()
```
thomcc: FWIW call should be [very cheap](https://github.com/mozilla/application-services/blob/master/megazords/fenix/android/src/main/java/mozilla/appservices/Megazord.kt#L12-L15) -- and becomes completely optional if we do https://github.com/mozilla/application-services/issues/1466, which we'd like to.

hawkinsw: It is indeed! Thanks :-)

hawkinsw:
@pocmo Let me know if you think that using the `run {}` with the comment will allay your fears about developers not being aware of the restriction. I wish that you could declare an anonymous scope in Kotlin, but it looks like this style is the closest that I can get. 

Also, the new version of the PR uses the `Dispatchers.IO` dispatcher!

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
csadilek: > We do see a performance gain, but I am not the person who can make the decision about risk/reward ratio! 

I am just curious, but can you share what the performance gains are we see here or why `RustHttpConfig.setClient` and `RustLog.enable()` are "expensive" and worth moving to a different thread?

csadilek: To be clear, all for it if it makes startup faster, but would be good to have an idea how much :).

hawkinsw: Of the 193ms it takes to execute setupInMainProcessOnly under ND, setClient() takes 22ms. 

For a single iteration of this on ND, the cumulative gains are

`setupInMainProcessOnly`: 193ms vs 174ms
`FenixApplication.onCreate`: 203ms vs 190ms

We do not regain the entire 22ms because it takes a certain amount of time to create the IO dispatchers for the first time. I am thinking about how we could potentially hide that cost as well. 

I will run additional tests on the Pixel hardware I have at my desk and update this accordingly.

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
hawkinsw: Let me know what you think about the new version!

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
hawkinsw: I did 100 runs of Fenix before/after the change here. In the course of those invocations on Fenix, there are anomalies (startup taking "negative" time) and I threw out those for both the before and after. The result was 71 and 75 "valid" Fenix startups. I averaged those startup times and calculated the percent difference. The result: 4.5%.

Issue #4776, opened 2019-08-16
Redesign About page in Settings


lime124:
pulling up Amy's comment from below:
> 
> Here is the updated mock with additional items we would like to include. Please note that this is a long list so please allow for page scrolling when adding additional items. Thanks!
> 
> What’s new in Firefox Preview (in current build) - https://support.mozilla.org/en-US/kb/whats-new-firefox-preview
> 
> Support - https://support.mozilla.org/en-US/products/firefox-preview
> 
> Privacy notice - https://www.mozilla.org/en-US/privacy/firefox/
> 
> Know your rights - about:rights
> 
> Licensing Information - about:license
> 
> GeckoView libraries we use (in current build)
> 
> Other open source libraries we use (in current build)
> 
> TBD **Release Notes** - need to confirm if we have these, @Tif will reach out to Joni Chan
> 
> ![Settings - About Firefox Preview@1x](https://user-images.githubusercontent.com/5992746/67527680-90af4680-f685-11e9-8023-f52606b875da.png)

lime124:
@vesta0 this should be pretty small. it would be nice to get in if we have a bit of time.

AmyYLee:
Let me know if there's anything else that needs to be added here:

![Settings - About Firefox Preview](https://user-images.githubusercontent.com/5992746/66350154-92a8a400-e928-11e9-891d-d447d798d2b1.png)

NotWoods:
@AmyYLee Is this ready for eng?

vesta0:
@lime124 please review Amy's design and let us know if this is ENG ready so we can size it during backlog grooming on Wednesday.

lime124:
@AmyYLee for a few technical reasons, we need to call out gv libraries separately. Can you add a third row a the bottom under "Open source..." for this. Same layout and styling. 

Text suggestion: "GeckoView libraries we use"

let me ping michelle to see who jeff's back up is. once Amy updates and we have final string we can label ready for eng. thanks!

lime124:
@mheubusch can confirm the copy suggested in my prior comment. thanks!

mheubusch:
I recommend this order/text: 
What's new in Firefox Preview
GeckoView libraries we use  
Other open source libraries we use

AmyYLee:
@lime124 @mheubusch I've updated the mock with the updated copy. This is ready for eng.

[...]

PR #6781, opened 2019-11-25
For #4776: Redesign and refactor the About Page 

ValentinTimisica:
---
Added new items to be displayed.
Due to complexity increase I decided to refactor the component to LibState.
Also, I added Unit tests for this component.

### Pull Request checklist

- [x] **Quality**: This PR builds and passes detekt/ktlint checks.
- [x] **Screenshots**: This PR includes a screenshot of the about page.

### After merge

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/settings/about/AboutFragment.kt
```diff
@@ -0,0 +1,184 @@
...
+
+        consumeFrom(store) {
+            aboutView.update(it)
+        }
+    }
+
+    private fun initializeState(): AboutPageState {
+        val items: MutableList<AboutPageItem> = mutableListOf()
+
+        items.add(populateAboutHeader())
```
NotWoods: This can be simplified to `listOf(populateAboutHeader()) + populateAboutList()`

app/src/main/java/org/mozilla/fenix/settings/about/AboutFragment.kt
```diff
@@ -0,0 +1,184 @@
...
+        val content = getString(R.string.about_content, appName)
+        val buildDate = BuildConfig.BUILD_DATE
+
+        return AboutPageItem.Header(aboutText, content, buildDate)
+    }
+
+    private fun populateAboutList(): List<AboutPageItem> {
+        val myList: MutableList<AboutPageItem> = mutableListOf()
+
+        context?.let {
```
NotWoods: We probably want to crash here if context isn't found, rather that displaying an empty list. `getString` in the previous method will crash anyways.

ValentinTimisica:
I have chosen a RecyclerView because the code is cleaner, simpler and it's easier to modify it in the future (adding or removing an element from the list vs manually adding or removing views in the xml) and also because @AmyYLee  said - https://github.com/mozilla-mobile/fenix/issues/4776#issuecomment-546115459 that the list is long and should be scrollable. 

For this exact reason I didn't separate the header from the rest of the items because it is expected they will fill the entire screen and also after checking the previous design I saw that the clickable items are just below the header and not on the bottom of the screen.

@AmyYLee how I should proceed? 
- Right now it looks like [this](https://drive.google.com/open?id=1BXN3BzkID_b6VU6nVkvvO6wV8VlYNg1H). We have all the items in one scrollable list making it easy for adding or removing items later.
- @NotWoods proposed implementation would have the header in fixed position at the top of the screen and the rest of the items scrollable and aligned to bottom. This would require an important refactoring of my patch.

app/src/main/java/org/mozilla/fenix/settings/about/AboutFragment.kt
ValentinTimisica: Thanks! This looks nice.

AmyYLee:
> I have chosen a RecyclerView because the code is cleaner, simpler and it's easier to modify it in the future (adding or removing an element from the list vs manually adding or removing views in the xml) and also because @AmyYLee said - [#4776 (comment)](https://github.com/mozilla-mobile/fenix/issues/4776#issuecomment-546115459) that the list is long and should be scrollable.
> 
> For this exact reason I didn't separate the header from the rest of the items because it is expected they will fill the entire screen and also after checking the previous design I saw that the clickable items are just below the header and not on the bottom of the screen.
> 
> @AmyYLee how I should proceed?
> 
>     * Right now it looks like [this](https://drive.google.com/open?id=1BXN3BzkID_b6VU6nVkvvO6wV8VlYNg1H). We have all the items in one scrollable list making it easy for adding or removing items later.

@ValentinTimisica requested access to view above link. 
> 
>     * @NotWoods proposed implementation would have the header in fixed position at the top of the screen and the rest of the items scrollable and aligned to bottom. This would require an important refactoring of my patch.

AmyYLee:
> I have chosen a RecyclerView because the code is cleaner, simpler and it's easier to modify it in the future (adding or removing an element from the list vs manually adding or removing views in the xml) and also because @AmyYLee said - [#4776 (comment)](https://github.com/mozilla-mobile/fenix/issues/4776#issuecomment-546115459) that the list is long and should be scrollable.
> 
> For this exact reason I didn't separate the header from the rest of the items because it is expected they will fill the entire screen and also after checking the previous design I saw that the clickable items are just below the header and not on the bottom of the screen.
> 
> @AmyYLee how I should proceed?
> 
>     * Right now it looks like [this](https://drive.google.com/open?id=1BXN3BzkID_b6VU6nVkvvO6wV8VlYNg1H). We have all the items in one scrollable list making it easy for adding or removing items later.
> 
>     * @NotWoods proposed implementation would have the header in fixed position at the top of the screen and the rest of the items scrollable and aligned to bottom. This would require an important refactoring of my patch.

The scroll portion in the video is correct (header fixed in place and page scrolls - this aligns with android settings).

There are some spacing and text styling that I would like to see adjusted, the leading looks tight. See mock. Also can we remove the emojis? 

Screen Shot 2019-11-26 at 11 12 09 AM

NotWoods:
Note that I'm not asking for the header to be fixed in place, I just want the different items to be aligned to the bottom. 

If the content height is less than the screen height, then the items should be aligned to the bottom and the header should have a large enough margin to fill the remaining space.

If the content height is greater than the screen height, then the entire page should scroll. The current implementation is fine for this case.

app/src/main/java/org/mozilla/fenix/settings/about/AboutPageStore.kt
```diff
@@ -0,0 +1,54 @@
...
+ * the view's state.
+ */
+sealed class AboutPageAction : Action
+
+@Suppress("UNUSED_PARAMETER") // the action parameter is unused
+fun aboutPageReducer(
+    currentState: AboutPageState,
+    action: AboutPageAction
+): AboutPageState {
+    // There is no possible action that can change this View's state while it is displayed to the user.
```
boek: Hey @ValentinTimisica, I appreciate the work that has been put into this! But I think I'm leaning towards @NotWoods comment as well. It feels like we're adding a lot of complexity pulling in `lib-state` here for what is ultimately a static list that will probably never change. I like the forward thinking but after reading this comment this solution does start to feel a little heavy. We could probably accomplish all of this in just the AboutFragment/xml. I'm happy to talk about it more if you still feel strongly about it 😊

ValentinTimisica:
I wanted to discuss with you @boek @NotWoods a little bit more about why I’ve chosen to use ```LibState``` and RecyclerView, a solution in which I greatly believe.

Using LibState, although atm there is little reason for the view’s state to change, means respecting the application’s architecture and not setting a precedent for other contributors to be tempted to follow just because it would be easier.

More than this, my solution already works, the code is clean, it’s easier to modify it in the future, has unit tests and downsizing to a “simpler” solution would mean spending more time to achieve the same result.

If we only modify AboutFragment.xml, we will have 6 different views instead of the RecyclerView, and 6 click listeners, all inside ```AboutFragment.kt```.

If you still feel strongly about needing me to change the current implementation, I will do so.

Thanks for the review!

NotWoods:
I'm OK with using a RecyclerView here. There's good reason to avoid duplicating the same view code throughout the list. However, it should be changed to only handle the list items and not the header. The header and list should be connected using a constraint layout so that the header can be aligned with the top of the page and the list can be aligned with the bottom, as seen in the mocks.

I still think we should avoid using LibState when the content is static. If and when the about page has dynamic functionality, we can pull it in. Until then I think a static list suffices.

PR #6874, opened 2019-11-29
Update Application Services to 0.44.0..

eoger:
A-C is bumping the dependency version in https://github.com/mozilla-mobile/android-components/pull/5131.


jonalmeida:
This needs the a-c changes to avoid crashing on startup. Please don't merge yet.

jonalmeida:
A-C snapshots should be available now after an a-c retrigger.

jonalmeida:
It seems like Firebase may have some caching issues with snapshots since @csadilek noticed there are crashes with nightly. We should force merge this then if we can't find a way to fix the UI tests. 🤷‍♂

PR #6938, opened 2019-12-04
Closes #6937 - Explicit ordering for migration and db operations in onCreate

### LINKED ISSUE #6937: [Bug] SQLITE_BUSY during fennec migration - overlapping runMaintenance call
In application's `onCreate`, we sometimes call `runMaintenance`. Currently, we'll definitely call it on first launch. Once that code runs, depending on timing of the migration logic, we may try to lock the underlying database concurrently.

grigoryk:
This patch adds an explicit order for migration logic and runMaintanence call.
Ordering these operations makes sure they won't run concurrently, thus avoiding
potential for SQLITE_BUSY errors.

---

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
+++ b/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
@@ -74,8 +74,12 @@ open class FenixApplication : Application() {
         Log.addSink(AndroidLogSink())
     }
 
-    @CallSuper
     open fun setupInMainProcessOnly() {
+        setupOperationsNotInvolvingStorageLayers()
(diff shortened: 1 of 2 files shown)


grigoryk:
This is somewhat similar to what you have to do for Gecko stuff, @pocmo.

grigoryk:
Yeah, that's tricky and I don't think we strictly need to worry about this - we can just disable that maintenance call for the time being instead. This approach made sense since it was simple, but I don't think coordinating this sort of thing with a background service will be simple...

grigoryk:
I'll close this for now, disable the offending db call, and file an issue to re-enable it.

Issue #6997, opened 2019-12-06
[Bug]Cut off "Search string to use" error message


AndiAJ:
## Steps to reproduce
1. Navigate to Settings
2. Tap in Search
3. Tap Add search engine
4. Tap custom
5. Tap the upper right check mark icon
6. Observe the "Search string to use" error message

### Expected behavior
The error message is properly displayed.

### Actual behavior
The error is cut off.

### Device information

* Android device: 
• Google Pixel 3a (Android 9)
• Nexus 5 (Android 6.0.1)

* Fenix version: 
• Fenix Beta 3.0.0-beta.3 from 12/5
• Fenix Nightly Build #13391806 from 12/5

### Notes
► Screenshot
![rsz_screenshot_20191206-151715](https://user-images.githubusercontent.com/51314259/70326821-27068880-183e-11ea-88c2-b28427aebefe.png)

► Video
![20191206_151627](https://user-images.githubusercontent.com/51314259/70326834-2ec62d00-183e-11ea-93fb-b3a8b0980f15.gif)

AndiAJ:
Hi, **verified as fixed** on the latest **Nightly Build #13450608 from 12/11** using the following devices:
• Google Pixel 3a  (Android 9)
• Huawei Mate 20 Lite (Android 8.1.0)
• Samsung Galaxy S7 (Android 7)
• Motorola G4 (Android 6.0.1)
• LG Nexus 4 (Android 5.1.1)

► Screenshot
![rsz_screenshot_20191212-110311](https://user-images.githubusercontent.com/51314259/70607150-98f81c80-1c06-11ea-8990-9b105996e2d0.png)

► Video
![20191212_110214](https://user-images.githubusercontent.com/51314259/70607198-a90ffc00-1c06-11ea-8e21-39330766b841.gif)

PR #6998, opened 2019-12-06
For #6997 Avoid overwriting empty_search_string error

mcarare:
Setting error twice also clips error message

---

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/settings/search/AddSearchEngineFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/search/AddSearchEngineFragment.kt
@@ -141,6 +141,7 @@ class AddSearchEngineFragment : Fragment(), CompoundButton.OnCheckedChangeListen
         }
     }
 
+    @Suppress("ComplexMethod")
     private fun createCustomEngine() {
         custom_search_engine_name_field.error = ""
         custom_search_engine_search_string_field.error = ""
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/fenix/settings/search/AddSearchEngineFragment.kt
```diff
@@ -170,7 +170,7 @@ class AddSearchEngineFragment : Fragment(), CompoundButton.OnCheckedChangeListen
         if (searchString.isEmpty()) {
             custom_search_engine_search_string_field
                 .error = resources.getString(R.string.search_add_custom_engine_error_empty_search_string)
-            hasError = true
+            return
```
Baron-Severin: This definitely fixes the overwriting problem, but I feel like the early return also adds to the complexity of the code.  I would prefer if we could link these two error checks (`isEmpty` and `contains`) to show that we only expect to handle one at a time.  For example, maybe something along these lines.

```kotlin
val searchFieldError = when {
     searchString.isEmpty() -> R.string.search_add_custom_engine_error_empty_search_string
    !searchString.contains("%s") -> R.string.search_add_custom_engine_error_missing_template
    else -> null
}
if (searchFieldError != null) {
    custom_search_engine_search_string_field.error = resources.getString(searchFieldError)
    hasError = true
}
```

mcarare: @Baron-Severin This fixes the visual problem. The cutoff is due to a known bug in Android that cuts the text when error is set multiple times, like in this case. The fix in this PR is intended to fix both the visual bug and the logic of error handling for the field.

Baron-Severin: Huh, TIL.  Thanks @mcarare!  Approving with this readability change as an optional nit, then.

PR #7004, opened 2019-12-06
For perf-frontend-issues#33: Add HotStartPerformanceMonitor!

mcomella:
---

Sending the review to Will first to make sure it meets our requirements and then I'll send to a core Fenix reviewer.

### Pull Request checklist

  - No UI changes

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
+++ b/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
@@ -52,6 +52,7 @@ import org.mozilla.fenix.home.intent.SpeechProcessingIntentProcessor
 import org.mozilla.fenix.home.intent.StartSearchIntentProcessor
 import org.mozilla.fenix.library.bookmarks.BookmarkFragmentDirections
 import org.mozilla.fenix.library.history.HistoryFragmentDirections
+import org.mozilla.fenix.perf.HotStartPerformanceMonitor
 import org.mozilla.fenix.search.SearchFragmentDirections
 import org.mozilla.fenix.settings.AboutFragmentDirections
 import org.mozilla.fenix.settings.DefaultBrowserSettingsFragmentDirections
@@ -68,6 +69,8 @@ open class HomeActivity : AppCompatActivity() {
     lateinit var browsingModeManager: BrowsingModeManager
 
     private var sessionObserver: SessionManager.Observer? = null
+
+    private val hotStartMonitor = HotStartPerformanceMonitor()
 
     private val navHost by lazy {
         supportFragmentManager.findFragmentById(R.id.container) as NavHostFragment
@@ -123,6 +126,16 @@ open class HomeActivity : AppCompatActivity() {
                 }
             }
         }
+    }
+
+    final override fun onRestart() {
+        hotStartMonitor.onRestartFirstMethodCall()
+        super.onRestart()
+    }
+
+    final override fun onPostResume() {
+        super.onPostResume()
+        hotStartMonitor.onPostResumeFinalMethodCall()
     }
 
     private fun unsetOpenLinksInAPrivateTabIfNecessary() {
--- a/app/src/main/java/org/mozilla/fenix/perf/HotStartPerformanceMonitor.kt
+++ b/app/src/main/java/org/mozilla/fenix/perf/HotStartPerformanceMonitor.kt
@@ -22,19 +22,22 @@ import android.util.Log
  * To see logs from this class, the user must enable VERBOSE logging for the appropriate tag:
  *   adb shell setprop log.tag.FenixPerf VERBOSE
  */
-class HotStartPerformanceMonitor {
+class HotStartPerformanceMonitor(
+    private val log: (String) -> Unit = { Log.v(Performance.TAG, it) }, // android log to minimize overhead.
+    private val getElapsedRealtime: () -> Long = { SystemClock.elapsedRealtime() }
+) {
 
     private var onRestartMillis: Long = -1
 
     fun onRestartFirstMethodCall() {
-        onRestartMillis = SystemClock.elapsedRealtime()
+        onRestartMillis = getElapsedRealtime()
     }
 
     fun onPostResumeFinalMethodCall() {
         // If onRestart was never called, this is not a hot start: ignore it.
         if (onRestartMillis >= 0) {
-            val elapsedMillis = SystemClock.elapsedRealtime() - onRestartMillis
-            Log.v(Performance.TAG, "hot start: $elapsedMillis") // android log to minimize overhead.
+            val elapsedMillis = getElapsedRealtime() - onRestartMillis
+            log("hot start: $elapsedMillis")
         }
     }
 }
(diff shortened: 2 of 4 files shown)


app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -69,6 +70,8 @@ open class HomeActivity : AppCompatActivity() {
 
     private var sessionObserver: SessionManager.Observer? = null
 
+    private val hotStartMonitor = HotStartPerformanceMonitor()
```
hawkinsw: Do you think that it's a good idea to include this in every version of the application? If not, we could do something like Metrics.kt does and have a "production version" that is essentially a noop and instantiate that here when the build variant is not performancetest.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -125,6 +128,16 @@ open class HomeActivity : AppCompatActivity() {
...
+    final override fun onRestart() {
+        hotStartMonitor.onRestartFirstMethodCall()
+        super.onRestart()
+    }
+
+    final override fun onPostResume() {
+        super.onPostResume()
+        hotStartMonitor.onPostResumeFinalMethodCall()
+    }
+
```
hawkinsw: I am convinced that this captures what we want. That said, I *hate* the lifecycle and reserve the right to later say, "Yes, of course," when we discover that this somehow misses a case.

app/src/main/java/org/mozilla/fenix/perf/HotStartPerformanceMonitor.kt
```diff
@@ -0,0 +1,43 @@
...
+ * Unfortunately, this monitor does not capture the entire duration of hot start because the
+ * framework calls several Android framework methods in the application's process before we are able
+ * to add monitoring code (i.e. the first time our application code is called in onRestart). An
+ * alternative implementation could measure performance from outside the application.
+ *
+ * To see logs from this class, the user must enable VERBOSE logging for the appropriate tag:
+ *   adb shell setprop log.tag.FenixPerf VERBOSE
+ */
+class HotStartPerformanceMonitor(
+    private val log: (String) -> Unit = { Log.v(Performance.TAG, it) }, // android log to minimize overhead.
```
hawkinsw: You do *not* have to do this, but I would have called this something like "report" because it indicates that we are reporting the hot start time.

app/src/test/java/org/mozilla/fenix/perf/HotStartPerformanceMonitorTest.kt
```diff
@@ -0,0 +1,43 @@
...
+    @Before
+    fun setUp() {
+        MockKAnnotations.init(this)
+        monitor = HotStartPerformanceMonitor(log, getElapsedRealtime = { elapsedRealtime })
+    }
+
+    @Test
+    fun `WHEN onRestart is not called but onPostResume is called THEN we do not log`() {
+        monitor.onPostResumeFinalMethodCall()
+        verify { log.invoke(any()) wasNot Called }
```
hawkinsw: Cool syntax. Testing, go figure.

mcomella:
> My only meta concern is that this framework could be used for us to build a complete system that reports startup times for all the different types (cold, hot, warm ...). By naming it HotStartPerformanceMonitor we are limiting ourselves. If we called it StartupPerformanceMonitor (or something like that) we could make space for future expansion if that's what we wanted.

This growth pattern is very likely so it's good to think about. There are two reasons I didn't go in that direction:

1) I try to avoid making the code generic too early because, by the time the generic code is needed, the requirements might have changed. Instead I try to do the simplest thing needed at the time because the simplest thing is usually the easiest to change. For example, it's easy to change the name of this class and add more measurements to this class if we wanted to but it'd be harder to refactor a more generic version if the initial requirements don't match the latest requirements.

2) If a class does one and only one thing, it's much easier to reason about. As such, I think ColdStart*, WarmStart*, and HotStart* should be separated into their own objects. There's a greater performance overhead but I think it's trivial – maybe I'm wrong about that.

mcomella:
> Of course, I am totally bikeshedding and don't want to upset you.

I think it's good to have (and timebox :P) these discussions: making readable & maintainable code is one of the most important things we can focus on and these discussions help us learn from each other. :)

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
mcomella: What concerns do you have about it being in production versions?

Taking some guesses:
- Logging in production could be a problem but VERBOSE logging is not visible by default nor do I think this is particularly sensitive information (it can be mostly inferred from other Fenix logs anyway)
- Performance penalty could be a problem but we'll want to record this data for telemetry eventually anyway so I don't think it's avoidable

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
mcomella: That's fair – I totally recognize measurement is making a best effort and iterating as necessary (e.g. to reduce noise) and Android doesn't make that any easier. :)

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
mcomella: > VERBOSE logging is not visible by default

I'll clean up the comment to clarify this.

app/src/main/java/org/mozilla/fenix/perf/HotStartPerformanceMonitor.kt
mcomella: Despite the way we're using it (to report), this method still functionally acts as a generic logging method so I'm inclined to leave as it is.

fwiw, if I did, `{ Log.v(Performance.TAG, "hot start: $it") }`, then a name like `reportHotStart` would make sense to me. :)

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
hawkinsw: I was concerned mostly about performance. If we are planning to eventually record it with Glean (etc), then this is moot!

hawkinsw:
> > My only meta concern is that this framework could be used for us to build a complete system that reports startup times for all the different types (cold, hot, warm ...). By naming it HotStartPerformanceMonitor we are limiting ourselves. If we called it StartupPerformanceMonitor (or something like that) we could make space for future expansion if that's what we wanted.
> 
> This growth pattern is very likely so it's good to think about. There are two reasons I didn't go in that direction:
> 
>     1. I try to avoid making the code generic too early because, by the time the generic code is needed, the requirements might have changed. Instead I try to do the simplest thing needed at the time because the simplest thing is usually the easiest to change. For example, it's easy to change the name of this class and add more measurements to this class if we wanted to but it'd be harder to refactor a more generic version if the initial requirements don't match the latest requirements.
> 

I think that this is a very good point. We'll go with it. Plus, you already have the higher-level Performance object that we could always stuff them all into if/when we need to expand. 

>     2. If a class does one and only one thing, it's much easier to reason about. As such, I think ColdStart*, WarmStart*, and HotStart* should be separated into their own objects. There's a greater performance overhead but I think it's trivial – maybe I'm wrong about that.

hawkinsw:
Sorry about closing this earlier. I think that I got confused between closing the conversation and closing the PR.

mcomella:
A unit test inside settings failed for permissions issues – it sounds unrelated. I'm going to rebase, force push, and see if CI passes.

mcomella:
I can't reproduce these issues locally: re-running.

mcomella:
I got approval from Jeff that we don't necessarily need core Fenix review on our patches. Since this one is not too complicated, I'm going to land.

PR #7034, opened 2019-12-09
Run migration in background service.

pocmo:
In order to avoid a half done migration we are moving the migration to a background service (that is running in the "foreground").

This is the Fenix part of:
https://github.com/mozilla-mobile/android-components/issues/4879

### DIFF
--- a/app/src/migration/java/org/mozilla/fenix/MigratingFenixApplication.kt
+++ b/app/src/migration/java/org/mozilla/fenix/MigratingFenixApplication.kt
@@ -4,6 +4,7 @@
 
 package org.mozilla.fenix
 
+import android.content.Context
 import kotlinx.coroutines.runBlocking
 import mozilla.components.support.migration.FennecMigrator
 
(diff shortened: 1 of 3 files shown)


app/src/migration/java/org/mozilla/fenix/MigratingFenixApplication.kt
```diff
@@ -4,19 +4,29 @@
...
 /**
  * An application class which knows how to migrate Fennec data.
  */
 class MigratingFenixApplication : FenixApplication() {
+    val migrator by lazy {
+        FennecMigrator.Builder(this, this.components.analytics.crashReporter)
+            .migrateOpenTabs(this.components.core.sessionManager)
+            .migrateHistory(this.components.core.historyStorage)
+            .migrateBookmarks(this.components.core.bookmarksStorage)
+            .migrateFxa(this.components.backgroundServices.accountManager)
```
grigoryk: missing .migrateLogins, which I bet is what you're conflicting on.

grigoryk:
Tested this out with my own profile, works beautifully!

grigoryk:
@pocmo pushed a conflict fix since I had it locally.

bors r+

pocmo:
Oh, damn, misunderstood you and pushed the same fix :)

PR #7070, opened 2019-12-10
Feature/#220 language menu

BranescuMihai:
---

### Pull Request checklist

- [X] **Screenshots**: This PR includes link to video at the bottom of this comment

### After merge

1. click on Show All Checks,

[**Link with feature**](https://drive.google.com/open?id=1suU1pUNDwIuaSdoub2UpBVCzksyFstRs)

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -429,6 +429,7 @@ dependencies {
     implementation Deps.mozilla_support_ktx
     implementation Deps.mozilla_support_rustlog
     implementation Deps.mozilla_support_utils
+    implementation Deps.mozilla_support_locale
 
     // We only care about support-migration in builds that will be overwriting Fennec.
     fennecProductionImplementation Deps.mozilla_support_migration
@@ -600,6 +601,23 @@ task printVariants {
         ])
         println 'variants: ' + groovy.json.JsonOutput.toJson(variants)
     }
+}
+
+task buildTranslationArray {
+    def foundLocales = new StringBuilder()
+    foundLocales.append("new String[]{")
+
+    fileTree("src/main/res").visit { FileVisitDetails details ->
+        if(details.file.path.endsWith("/strings.xml")){
+            def languageCode = details.file.parent.tokenize('/').last().replaceAll('values-','').replaceAll('-r','-')
+            languageCode = (languageCode == "values") ? "en-US" : languageCode
+            foundLocales.append("\"").append(languageCode).append("\"").append(",")
+        }
+    }
+
+    foundLocales.append("}")
+    def foundLocalesString = foundLocales.toString().replaceAll(',}','}')
+    android.defaultConfig.buildConfigField "String[]", "SUPPORTED_LOCALE_ARRAY", foundLocalesString
 }
 
 def glean_android_components_tag = (
--- a/app/src/main/java/org/mozilla/fenix/settings/advanced/LocaleAdapter.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/advanced/LocaleAdapter.kt
@@ -1,0 +1,150 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.settings.advanced
+
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import androidx.core.view.isVisible
+import androidx.recyclerview.widget.DiffUtil
+import androidx.recyclerview.widget.RecyclerView
+import kotlinx.android.synthetic.main.locale_settings_item.view.locale_selected_icon
+import kotlinx.android.synthetic.main.locale_settings_item.view.locale_subtitle_text
+import kotlinx.android.synthetic.main.locale_settings_item.view.locale_title_text
+import org.mozilla.fenix.R
+import java.util.Locale
+
+class LocaleAdapter(private val interactor: LocaleSettingsViewInteractor) :
+    RecyclerView.Adapter<BaseLocaleViewHolder>() {
+
+    private var localeList: List<Locale> = listOf()
+    private var selectedLocale: Locale = Locale.getDefault()
+
+    override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): BaseLocaleViewHolder {
+        val view =
+            LayoutInflater.from(parent.context)
+                .inflate(R.layout.locale_settings_item, parent, false)
+
+        return when (viewType) {
+            ItemType.DEFAULT.ordinal -> SystemLocaleViewHolder(
+                view,
+                interactor,
+                selectedLocale
+            )
+            ItemType.LOCALE.ordinal -> LocaleViewHolder(
+                view,
+                interactor,
+                selectedLocale
+            )
+            else -> throw IllegalStateException("ViewType $viewType does not match to a ViewHolder")
+        }
+    }
+
+    override fun getItemCount(): Int {
+        return localeList.size
+    }
+
+    override fun onBindViewHolder(holder: BaseLocaleViewHolder, position: Int) {
+        holder.bind(localeList[position])
+    }
+
+    override fun getItemViewType(position: Int): Int {
+        return when (position) {
+            0 -> ItemType.DEFAULT
+            else -> ItemType.LOCALE
+        }.ordinal
+    }
+
+    fun updateData(localeList: List<Locale>, selectedLocale: Locale) {
+        val diffUtil = DiffUtil.calculateDiff(
+            LocaleDiffUtil(
+                this.localeList,
+                localeList,
+                this.selectedLocale,
+                selectedLocale
+            )
+        )
+        this.localeList = localeList
+        this.selectedLocale = selectedLocale
+
+        diffUtil.dispatchUpdatesTo(this)
+    }
+
+    inner class LocaleDiffUtil(
+        private val old: List<Locale>,
+        private val new: List<Locale>,
+        private val oldSelectedLocale: Locale,
+        private val newSelectedLocale: Locale
+    ) : DiffUtil.Callback() {
+
+        override fun areItemsTheSame(oldItemPosition: Int, newItemPosition: Int): Boolean {
+            val selectionChanged =
+                old[oldItemPosition] == oldSelectedLocale && oldSelectedLocale != newSelectedLocale
+            return old[oldItemPosition] == new[newItemPosition] && !selectionChanged
+        }
+
+        override fun areContentsTheSame(oldItemPosition: Int, newItemPosition: Int): Boolean =
+            old[oldItemPosition].toLanguageTag() == new[newItemPosition].toLanguageTag()
+
+        override fun getOldListSize(): Int = old.size
+        override fun getNewListSize(): Int = new.size
+    }
+
+    enum class ItemType {
+        DEFAULT, LOCALE;
+    }
+}
+
+class LocaleViewHolder(
+    view: View,
+    private val interactor: LocaleSettingsViewInteractor,
+    private val selectedLocale: Locale
+) : BaseLocaleViewHolder(view) {
+    private val icon = view.locale_selected_icon
+    private val title = view.locale_title_text
+    private val subtitle = view.locale_subtitle_text
+
+    override fun bind(locale: Locale) {
+        // capitalisation is done using the rules of the appropriate locale (endonym and exonym)
+        title.text = locale.getDisplayName(locale).capitalize(locale)
+        subtitle.text = locale.displayName.capitalize(Locale.getDefault())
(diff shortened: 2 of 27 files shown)


app/src/main/java/org/mozilla/fenix/settings/advanced/LocaleManagerExtension.kt
```diff
@@ -0,0 +1,72 @@
...
+    localeList.add(Locale("ta"))
+    localeList.add(Locale("tr"))
+    return localeList
+}
+
+/**
+ * Returns the locale that corresponds to the language stored locally by us. If no suitable one is found,
+ * return default.
+ */
+fun LocaleManager.getSelectedLocale(context: Context, localeList: List<Locale>): Locale {
```
Baron-Severin: `localeList` definitely makes this easier to test, but it doesn't seem to me that consumers should have to pass that in order to get the selected locale.  What about making it an optional param?  `localeList: List<Locale> = getSupportedLocales()`

app/src/main/java/org/mozilla/fenix/settings/advanced/LocaleManagerExtension.kt
```diff
@@ -0,0 +1,72 @@
...
+    val selectedLocale = getCurrentLocale(context)
+    if (selectedLocale == null) {
+        return localeList[0]
```
Baron-Severin: Based on the kdoc, it sounds like this is assuming that the first entry will always be the default value.  Could we just return `getSystemDefault()` here instead?  Seems easier to understand to me.

app/src/main/java/org/mozilla/fenix/settings/advanced/LocaleManagerExtension.kt
```diff
@@ -0,0 +1,72 @@
...
+    } else {
+        for (locale in localeList) {
+            if (locale.toLanguageTag() == selectedLocale.toLanguageTag()) {
+                return locale
+            }
+        }
+    }
+    return localeList[0]
```
Baron-Severin: Optional nit: this one's up to personal preference, but since `toLanguageTag` is guaranteed to be nonnull this method could be rewritten as

```kotlin
fun LocaleManager.getSelectedLocale(context: Context, localeList: List<Locale>): Locale {
    val selectedLocale = getCurrentLocale(context)?.toLanguageTag()
    val supportedMatch = localeList
        .firstOrNull { it.toLanguageTag() == selectedLocale }

    return supportedMatch ?: getSystemDefault()
}
```

app/src/main/java/org/mozilla/fenix/settings/advanced/LocaleSettingsFragment.kt
```diff
@@ -0,0 +1,80 @@
...
+        super.onViewCreated(view, savedInstanceState)
+        consumeFrom(store) {
+            localeView.update(it)
+        }
+    }
+
+    private fun getStore(): LocaleSettingsStore {
+        val supportedLocales = LocaleManager.getSupportedLocales()
+        val selectedLocale: Locale
+        selectedLocale = LocaleManager.getSelectedLocale(requireContext(), supportedLocales)
```
Baron-Severin: Nit: `selectedLocale` assignment could be inlined with its declaration.

app/src/main/java/org/mozilla/fenix/settings/advanced/LocaleManagerExtension.kt
```diff
@@ -0,0 +1,72 @@
...
+    localeList.add(Locale("ru"))
+    localeList.add(Locale("sk"))
+    localeList.add(Locale("es"))
+    localeList.add(Locale("es", "AR"))
+    localeList.add(Locale("es", "ES"))
+    localeList.add(Locale("sv", "SE"))
+    localeList.add(Locale("ta"))
+    localeList.add(Locale("tr"))
+    return localeList
+}
```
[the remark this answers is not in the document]
Pike: Is there a way to do this w/out hard-coding the list?

For one, this list is already out of date, we just added more locales.

Also, we might skip locales as part of the release builds, and then the UI would claim support for locales we don't ship, right?

app/src/main/java/org/mozilla/fenix/settings/advanced/LocaleAdapter.kt
```diff
@@ -0,0 +1,158 @@
...
+    private val interactor: LocaleSettingsViewInteractor,
+    private val selectedLocale: Locale
+) : BaseLocaleViewHolder(view) {
+    private val icon = view.locale_selected_icon
+    private val title = view.locale_title_text
+    private val subtitle = view.locale_subtitle_text
+
+    override fun bind(locale: Locale) {
+        title.text = locale.getDisplayName(locale).capitalize(locale)
+        subtitle.text = locale.displayName.capitalize(Locale.getDefault())
```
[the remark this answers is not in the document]
BranescuMihai: Yes, the reason is it is capitalising the text using the appropriate locale (for example, in Turkish, capital 'i' is not I). Will leave a comment:)

app/src/main/java/org/mozilla/fenix/settings/advanced/LocaleSettingsFragment.kt
BranescuMihai: true, inlined

app/src/main/java/org/mozilla/fenix/settings/advanced/LocaleAdapter.kt
```diff
@@ -0,0 +1,158 @@
...
+            icon.visibility = View.INVISIBLE
+        }
+
+        itemView.setOnClickListener {
+            interactor.onLocaleSelected(locale)
+        }
+    }
+}
+
+class DefaultLocaleViewHolder(
```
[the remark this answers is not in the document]
BranescuMihai: Changed to `SystemLocaleViewHolder`

app/src/main/java/org/mozilla/fenix/settings/advanced/LocaleManagerExtension.kt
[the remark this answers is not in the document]
Amejia481: In Focus and FFTV, we have a Gradle task that auto-generates this list based on the the different variants of `/res/values-{android_locale}/strings.xml` directories. We could do something similar here but ideally we were looking to create a Gradle plugin on Android Components to share this functionality between multiple apps, you can find more info here https://github.com/mozilla-mobile/android-components/issues/3586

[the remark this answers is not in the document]
Amejia481: This also another related https://github.com/mozilla-mobile/android-components/issues/3585 issue, hope it helps!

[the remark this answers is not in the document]
BranescuMihai: @Pike totally agree with removing the hardcoded list, it's just a shame Android has a method in `AssetManager.java` called `getNonSystemLocales()` that is exactly what we need, but it's in their internal API.
I had no idea we had a similar implementation on Focus and FFTV, so I created my own Gradle task, but I feel it's actually a bit simpler than those (for example, using build config instead of generating a java class). 
@Amejia481 please have a look and tell me your opinion :)

app/src/main/java/org/mozilla/fenix/settings/advanced/LocaleAdapter.kt
```diff
@@ -0,0 +1,158 @@
...
+    private val subtitle = view.locale_subtitle_text
+
+    override fun bind(locale: Locale) {
+        title.text = locale.getDisplayName(locale).capitalize(locale)
+        subtitle.text = locale.displayName.capitalize(Locale.getDefault())
+
+        if (locale === selectedLocale) {
+            icon.visibility = View.VISIBLE
+        } else {
+            icon.visibility = View.INVISIBLE
```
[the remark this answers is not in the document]
BranescuMihai: Changed, had to modify the goneMargin, but you're right, it's a bit cleaner this way, thanks!

app/src/main/java/org/mozilla/fenix/settings/advanced/LocaleManagerExtension.kt
BranescuMihai: Yep, sounds right, they shouldn't. Added as optional

app/src/main/java/org/mozilla/fenix/settings/advanced/LocaleManagerExtension.kt
BranescuMihai: Fair enough, changed to getSystemDefault instead of the first entry

app/src/main/java/org/mozilla/fenix/settings/advanced/LocaleManagerExtension.kt
BranescuMihai: Changed it, thanks!

app/src/main/java/org/mozilla/fenix/settings/advanced/LocaleManagerExtension.kt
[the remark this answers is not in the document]
Pike: I left a note on the a-c issue, I came across https://developer.android.com/reference/android/content/res/Configuration.html#getLocales(). But you'd need to practically test it to see if it does the trick. Also in the case of us restricting resources we ship.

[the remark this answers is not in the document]
BranescuMihai: But regardless of which Gradle task we will use in the end, I agree, we should move it to AC so all 3 products can use it

app/build.gradle
```diff
@@ -614,6 +615,23 @@ task printVariants {
     }
 }
 
+task buildTranslationArray {
```
Amejia481: Would you mind to move this to AC? :) We can have it as a Gradle plugin on the  `mozilla.components.support.locale`package, we have some examples of Gradle plugins in [buildSrc](https://github.com/psymoon/android-components/tree/08661a2af1d0afa09902daeec2423e40014006ce/buildSrc) and [here](https://github.com/psymoon/android-components/blob/c58d6a32d3c3815ced44fa65a811a42a51f0daee/build.gradle#L278). It will be pretty nice if we could generate an extension function on `LocaleManager` for people that use the plugging. Maybe [Kotlin Poet ](https://github.com/square/kotlinpoet#extension-functions)could be very helpful. What do you think?

Something like:

```kotlin
fun LocaleManager.getSupportedLocales(): List<Locale>{
    return listOf<Locale>(
            Locale("eu"),
            Locale("ca"),
            Locale("zh", "CN"),
            Locale("zh", "TW"),
            Locale("cs"),
            Locale("nl"),
            Locale("en"),
            Locale("fi"),
            Locale("fr"),
            Locale("de"),
            Locale("hu"),
            Locale("in"),
            Locale("it"),
            Locale("ja"),
            Locale("ko"),
            Locale("pl"),
            Locale("pa", "IN"),
            Locale("ru"),
            Locale("sk"),
            Locale("es"),
            Locale("es", "AR"),
            Locale("es", "ES"),
            Locale("sv", "SE"),
            Locale("ta"),
            Locale("tr")
    )
} 
```

BranescuMihai: Hmm, turns out that it's not as straight-forward as I hoped at all. Looked into Kotlin Poet and the class generation is the doable part, but the following issues appeared:
- From what I see, declaring the task as a plugin in buildSrc makes it harder for me to share across projects
- Fenix needs to have it's own task, that simply calls the locale task in AC - currently I hardcode an absolute path to the locale task in AC, but I don't know exactly how to do that with the actual AC dependency
- AC task needs to access the res folder in fenix(and eventually other projects) in order to get the locales, and for that we need a path - currently I can't find how to send the path as a param (maybe how we do with ktlint?) - [here is their implementation](https://github.com/pinterest/ktlint/blob/master/ktlint/src/main/kotlin/com/pinterest/ktlint/Main.kt)

**TLDR**: I need to call an AC task from Fenix with params. If you have experience with this, I would appreciate the help! Otherwise, I would suggest going forward with this PR as is, since it will be working for Fenix and we're kinda on a short schedule for the planned release

Amejia481: Let's me try to take a look at what you have done so far (Could you share your repo?). If it gets too complicated we can proceed as it is and improve it in a follow issue.

Amejia481: Please, proceed with what you have so far here. I will try to find an alternative on AC.

Baron-Severin:
@BranescuMihai when you have a moment, would you mind resolving the conflicts and then landing this?

Baron-Severin:
I think @BranescuMihai may be on PTO, but these conflicts are very minor.  I'll fix them up and merge this in.

BranescuMihai:
> I think @BranescuMihai may be on PTO, but these conflicts are very minor. I'll fix them up and merge this in.

Yep, I was on PTO, thanks for helping with the conflicts @Baron-Severin !

PR #7081, opened 2019-12-11
Add version.txt for use in release automation via ship-it.

Callek:
To support doing Fenix releases via Ship-It, we need a version.txt here. @JohanLorenzo what do you think of this patch?


taskcluster/fenix_taskgraph/__init__.py
```diff
@@ -40,7 +40,9 @@ def get_decision_parameters(graph_config, parameters):
     head_tag = os.environ.get("MOBILE_HEAD_TAG", "").decode('utf-8')
     parameters["head_tag"] = head_tag
     parameters["release_type"] = _resolve_release_type(head_tag)
-    parameters["release_version"] = head_tag[1:] if head_tag else ""
+    parameters["release_version"] = read_version_file()
+    if head_tag:
```
JohanLorenzo: Good thinking on this override! 👍

taskcluster/fenix_taskgraph/__init__.py
```diff
@@ -56,6 +58,11 @@ def get_decision_parameters(graph_config, parameters):
         parameters["target_tasks_method"] = "release"
 
 
+def read_version_file():
+    with open(os.path.join(os.path.dirname(__file__), '..', '..', 'version.txt')) as f:
```
JohanLorenzo: 👍

I don't remember taskgraph importing any other file in the repo. We may eventually want to extract this logic in a dedicated function (like taskgraph in Gecko does). That said, let's keep it this way for now 😃

taskcluster/fenix_taskgraph/__init__.py
```diff
@@ -56,6 +58,11 @@ def get_decision_parameters(graph_config, parameters):
...
+        return f.read().strip()
```
JohanLorenzo: Nit: You likely need to add `.decode('utf-8')`

PR #7127, opened 2019-12-12
Set Onboarding Status from CI through an Intent

hawkinsw:
---

If the intent that starts the HomeActivity includes an extra
boolean flag named finishonboarding and it is set to true,
Fenix will "finish onboarding".

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
+++ b/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
@@ -52,6 +52,7 @@ import org.mozilla.fenix.home.intent.SpeechProcessingIntentProcessor
 import org.mozilla.fenix.home.intent.StartSearchIntentProcessor
 import org.mozilla.fenix.library.bookmarks.BookmarkFragmentDirections
 import org.mozilla.fenix.library.history.HistoryFragmentDirections
+import org.mozilla.fenix.onboarding.FenixOnboarding
 import org.mozilla.fenix.perf.HotStartPerformanceMonitor
 import org.mozilla.fenix.search.SearchFragmentDirections
 import org.mozilla.fenix.settings.about.AboutFragmentDirections
@@ -96,6 +97,10 @@ open class HomeActivity : AppCompatActivity() {
         setContentView(R.layout.activity_home)
 
         setupToolbarAndNavigation()
+
+        if (intent.getBooleanExtra(EXTRA_FINISH_ONBOARDING, false)) {
+            FenixOnboarding(this).finish()
+        }
 
         if (settings().isTelemetryEnabled) {
             lifecycle.addObserver(BreadcrumbsRecorder(components.analytics.crashReporter,
@@ -355,5 +360,6 @@ open class HomeActivity : AppCompatActivity() {
         const val PRIVATE_BROWSING_MODE = "private_browsing_mode"
         const val EXTRA_DELETE_PRIVATE_TABS = "notification_delete_and_open"
         const val EXTRA_OPENED_FROM_NOTIFICATION = "notification_open"
+        const val EXTRA_FINISH_ONBOARDING = "finishonboarding"
     }
 }


app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -147,6 +149,9 @@ class HomeFragment : Fragment() {
 
         lifecycle.addObserver(sessionObserver)
 
+        if (finishOnboarding)
```
mcomella: nit: generally we always brace if/else statements, unless it's used as a one-line expression.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -147,6 +149,9 @@ class HomeFragment : Fragment() {
...
+            onboarding.finish()
```
mcomella: I wonder if calling `hideOnboarding()` is more appropriate. Functionally, I think it's the same, however.

I suspect this is moot if we do my suggestion above.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
mcomella: We could make this A) more close to what users experience and B) more robust by adding the override directly to the `FenixOnboarding` class in the function used by the application code, e.g.:

```kotlin
fun userHasBeenOnboarded() = isOverriddenThroughIntent ||
        onboardingPrefs.onboardedVersion == CURRENT_ONBOARDING_VERSION
```

It's more robust such that when someone refactors this code, it's more likely that they won't accidentally break our code.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -97,6 +99,8 @@ open class HomeActivity : AppCompatActivity() {
 
         setupToolbarAndNavigation()
 
+        finishOnboarding = intent.getBooleanExtra("finishonboarding", false)
```
mcomella: nit: the key should be a constant: see the HomeActivity companion object for examples.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
hawkinsw: Good call!

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
hawkinsw: Good call!

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -97,6 +98,10 @@ open class HomeActivity : AppCompatActivity() {
 
         setupToolbarAndNavigation()
 
+        if (intent.getBooleanExtra(EXTRA_FINISH_ONBOARDING, false)) {
```
mcomella: nit: Not your fault but I don't think `FenixOnboarding` is an intuitive name: it should be something like, `UserOnboardedState`. Similarly, I don't think `finish` is an intuitive name for that method either. It's more of a, `setUserOnboarded`. Relatedly, I find `EXTRA_FINISH_ONBOARDING` to be misleading too. I don't think this is worth the time to fix now.

On that note, if you happen upon code that you think could be better and relatively easy to change, I think it's okay for us to make those sorts of changes too. :)

mcomella: nit: it might be good to add a comment mentioning this flag is for automation, or performance work, because if someone finds this flag in isolation and doesn't see it used anywhere else in the code, they may think it's unused and remove it.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -97,6 +98,10 @@ open class HomeActivity : AppCompatActivity() {
...
+            FenixOnboarding(this).finish()
```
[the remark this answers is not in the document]
hawkinsw: You are right, but the good news is that all of `FenixOnboarding`'s state is recreated from `SharedPreferences` every time it is instantiated. There is no runtime state passed between instances. So, this is basically a convenience method for setting the onboarding status in the app's shared preferences. 

I agree 100% that it would be good to move to `Components` so that we do not have to instantiate this multiple times (here [potentially], HomeFragment and other places that might be added in the future).  Each instantiation requires reading from `SharedPreferences`, an expensive operation. Great idea and great catch!

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
hawkinsw: Yup, that would have been a great idea! I can't believe I didn't think of doing that. Great catch!

Issue #4978, opened 2019-08-28
Fennec migration: Keep widget on home screen


pocmo:
> We have added a search widget to Fennec and would like users to keep the widget after upgrade to Fenix. We would need to make sure the Android system can find and access the widget the same way as before in Fenix.

pocmo:
```
       <receiver
            android:name="org.mozilla.gecko.search.SearchWidgetProvider">

            <intent-filter>

                <action
                    android:name="android.appwidget.action.APPWIDGET_UPDATE" />
            </intent-filter>

            <meta-data
                android:name="android.appwidget.provider"
                android:resource="@ref/0x7f130012" />
        </receiver>
```

```
        <activity
            android:theme="@ref/0x7f1100e6"
            android:name="org.mozilla.gecko.search.SearchWidgetConfigurationActivity">

            <intent-filter>

                <action
                    android:name="android.appwidget.action.APPWIDGET_CONFIGURE" />
            </intent-filter>
        </activity>
```

pocmo:
```
<?xml version="1.0" encoding="utf-8"?>

<!-- This Source Code Form is subject to the terms of the Mozilla Public
   - License, v. 2.0. If a copy of the MPL was not distributed with this
   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->

<appwidget-provider xmlns:android="http://schemas.android.com/apk/res/android"
    android:initialLayout="@layout/widget_search_default_col_layout"
    android:minWidth="360dp"
    android:minHeight="40dp"
    android:minResizeWidth="64dp"
    android:minResizeHeight="40dp"
    android:previewImage="@drawable/search_widget_preview"
    android:resizeMode="horizontal">
</appwidget-provider>
```

ebalazs-sv:
@pocmo I had tested this on Relase, Beta, Nightly migration builds from 01/08 with Google Pixel 3 (Android Q), Samsung Galaxy S8 (Android 9), Huawei Honor 8 (Android 7.0):
- For Release the search widget is kept and functional;
- For Beta and Nightly the search widget bar is kept, but the Fenix icon is not displayed, also this can't be tested because of the start-up crash https://github.com/mozilla-mobile/android-components/issues/5276.

[Screenshot](https://drive.google.com/file/d/1tTU-sSQ4mnOw1zbFUKBb4MQaieQ9Jtl0/view?usp=sharing)

ebalazs-sv:
[...]

PR #7202, opened 2019-12-16
Issue #4978: Reuse widget package name of Fennec..

pocmo:
This patch moves `SearchWidgetProvider` from `org.mozilla.fenix.widget` to `org.mozilla.gecko.search`. Using the same package name as Fennec will prevent the widget from disappearing (or before that becoming unresponsive) when migrating to Fenix. The downside of that is that existing Firefox Preview users will have to add the widget to their home screen again (The reverse is happening). But since that is a "preview" version with much less users that seems to be a good trade off.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -176,7 +176,7 @@
 
         <service android:name=".downloads.DownloadService" />
         <receiver
-            android:name=".widget.SearchWidgetProvider">
+            android:name="org.mozilla.gecko.search.SearchWidgetProvider">
             <intent-filter>
                 <action android:name="android.appwidget.action.APPWIDGET_UPDATE" />
             </intent-filter>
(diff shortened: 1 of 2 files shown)


pocmo:
Adding do not land label: Doing some more testing. We recently opened the "home screen launcher" issue and this looks like it could have the same issue.

pocmo:
Okay, I verified this is working for all our Fennec flavors when upgrading. This is now ready to land.

pocmo:
@Mugurell Can you double check this too? :)

Mugurell:
> @Mugurell Can you double check this too? :)

Tested this with the Official and Beta variants and saw that
Immediately after the update, before opening Fenix
- the widget is kept on the screen although the ff logo and search hint are not visible
- tapping on the widget opens fenix but does not start searching

After having Fenix opened once the widget looks right and user taps do start the intended action.
https://drive.google.com/file/d/1jbq4NDn9adKp2BrTz_QiQ6D2O-i94IXi/view?usp=sharing
(at the top of the screen is the "release" app widget, at the bottom is the beta one. I only interact with the "release" one)

To note though that in some cases the widget did properly update and still just opened Fenix but not in search mode even after having it opened multiple times / cold / hot start.

pocmo:
Interesting. It may need to be rebased with 16b4680c06c936bdda22ef9eead71469e5a771c7 (That's how I tested it locally). I think we should go ahead and merge and then let QA look at this for different device/android configurations.

PR #7221, opened 2019-12-16
Fix 5366 - Use VectorDrawableCompat for all imagery

darkwing:
---

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/settings/Extensions.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/Extensions.kt
@@ -26,7 +26,7 @@ fun SitePermissions.toggle(featurePhone: PhoneFeature): SitePermissions {
 }
 
 /**
- * In devices with Android 6, when we use android:button="@null" app:drawableStartCompat doesn't work via xml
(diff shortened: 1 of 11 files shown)


app/src/main/java/org/mozilla/fenix/settings/Extensions.kt
```diff
@@ -25,12 +26,12 @@ fun SitePermissions.toggle(featurePhone: PhoneFeature): SitePermissions {
 }
 
 /**
- * In devices with Android 6, when we use android:button="@null" android:drawableStart doesn't work via xml
+ * In devices with Android 6, when we use android:button="@null" app:drawableStartCompat doesn't work via xml
```
Baron-Severin: Nit: did you check that this is still true with `drawableStartCompat`?  If not, I'd rather leave this comment as it was.  Ideally we would test the new version (and set this in the XML if it worked), but it's probably not worth the time to go digging for an appropriate device unless you have one handy.

darkwing: Ooops, you're right, I meant to go back and fix that.  Thank you for spotting it!

PR #7239, opened 2019-12-17
Closes #4976: Use dynamic launcher entry in manifest.

### LINKED ISSUE #4976: Fennec migration: Keep existing legacy launcher icons working
> If the user has added Fennec to the home screen then those icons may disappear from the home screen unless we take care that the same launcher entry points exist in Fenix.

pocmo:
Fixing reopened #6872 

`org.mozilla.gecko.App` worked fine for local builds of Fennec. However Fennec's arcane build process turns that into a different value for our production builds (App ID + `.App`). This follows the same approach by using the manifest placeholder `${applicationId}`. Initially I wanted to just use `.App` like Fennec does in its manifest. But it turns out that Gradle and Fennec's build will turn that into different things in the final APK.

I tested this with re-signed Fennec production builds and it is now working for all fennec flavors of Fenix:

migrating

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -30,12 +30,12 @@
         tools:ignore="UnusedAttribute">
 
         <!--
-          We inherited this entry (org.mozilla.gecko.App) from Fennec. We need to keep this as our
+          We inherited this entry (${applicationId}.App) from Fennec. We need to keep this as our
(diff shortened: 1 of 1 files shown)


pocmo:
> Interesting that org.mozilla is enough.

Every flavor has an `applicationIdSuffix` defined. So the final ID will be `org.mozilla.firefox` etc.

PR #7246, opened 2019-12-17
Issue 7192 move unset open link in a private tab if necessary off main thread

emalysz:
---

### Pull Request checklist

### After merge

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -147,7 +151,7 @@ open class HomeActivity : AppCompatActivity() {
         // Toggle off the open_link_in_private_tab pref if we are no longer set as the default browser
         // We do this on a separate thread to alleviate performance issues
         val weakReferenceContext = WeakReference(this)
-        lifecycleScope.launch {
+        scope.launch {
```
NotWoods: To change which thread is used, we need to change the dispatcher rather than the coroutine scope. The scope defines when the work will be cancelled (when the activity is destroyed in the case of `lifecycleScope`), and the dispatcher defines the thread. `Dispatchers.Main` will keep running code on the main thread.

```suggestion
        lifecycleScope.launch(Dispatchers.IO) {
```

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -147,7 +148,7 @@ open class HomeActivity : AppCompatActivity() {
         // Toggle off the open_link_in_private_tab pref if we are no longer set as the default browser
         // We do this on a separate thread to alleviate performance issues
         val weakReferenceContext = WeakReference(this)
-        lifecycleScope.launch {
+        lifecycleScope.launch(Dispatchers.IO) {
```
mcomella: Every time I see something become asynchronous, I ask: is there a simpler way to do this? There are two things I can think of in this case:
1. Does this need to happen here, or at all?
1. If yes, can we make Browsers faster?

I looked into 1 and I don't think it has to happen here: can you look into whether or not an alternative implementation is worthwhile? Given this is 20ms (on my Pixel 2), I'm wondering if it's worth our time to come up with a new, simpler implementation.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -147,7 +148,7 @@ open class HomeActivity : AppCompatActivity() {
...
             val context = weakReferenceContext.get() ?: return@launch
             if (!Browsers.all(context).isDefaultBrowser) {
```
mcomella: This code is [roughly duplicated](https://github.com/mozilla-mobile/fenix/blob/562eb89356e291f1037cde6f87cbfe108488f58e/app/src/main/java/org/mozilla/fenix/IntentReceiverActivity.kt#L39-L41). However, the other time it also records a metric – due to the duplication, I believe there's a bug and our metric is undercounting. Would you mind addressing this too or, if it's too much trouble, filing a follow-up?

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
mcomella: Note to self to follow-up on which Dispatcher is appropriate here. afaik, IO is intended for blocking tasks so it may do more expensive things like create new threads to support those operations.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -147,7 +148,7 @@ open class HomeActivity : AppCompatActivity() {
...
                 context.settings().preferences
```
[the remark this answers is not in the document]
NotWoods: Does SharedPrefs actually run on the IO thread? I have no idea, but I've seen SharedPrefs-related strict mode problems.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
NotWoods: There are 3 primary dispatchers:
- `Main` (The UI thread)
- `IO`
- `Default`

@colintheshots did some research and found that we should avoid `Default` as much as possible. IIRC you only get 2 threads and search uses one of them already, so using `Default` causes a lot of perf issues.

For pretty much every case we use `IO`, unless we need to show something on the UI.

mcomella: > For pretty much every case we use IO, unless we need to show something on the UI.

I'm concerned about this from a performance perspective: in my understanding of the original implementation (which has since changed and I have not had time to fully review the new one yet), the `IO` dispatcher may create additional threads if none are available. Creating threads is an expensive operation but, since this dispatcher is intended for blocking tasks, the thread creation time may be insignificant compared to the blocking task and there is a need for immediate execution (e.g. start network calls ASAP).

However, for CPU-bound tasks, the time of the task may not be worth the creation of an entire thread. Furthermore, context switching between threads adds unnecessary CPU burden so, in some sense, it's inefficient to have too many threads for active CPU work (as opposed to the sleeping threads for blocking calls) and it's kind of a feature to have the same number of threads available as the number of CPU cores.

There is, however, a distinction the coroutine dispatchers don't make: background work intended to return to the main thread and background work that completes silently. Background work that returns to the main thread should happen ASAP while background work that completes silently can often wait.

I looked through Colin's reports – [Audit](https://docs.google.com/document/d/17WNYoS-ucL6FR8bAM64NzH5CAT_v7wIYgdHN0mQL_r0/edit) & [ANR](https://docs.google.com/document/d/1THmvymMO5HpDb34v9neECU1453DsiWCyHqgrAq3LEdc/edit#) – and I notice they're focused on startup. For startup, many background operations are intended to return to the main thread so the distinction is important: we may want to consider alternative solutions. However, I'm concerned a blanket-policy for using the IO dispatcher has additional, unnecessary inefficiencies.

Outside of startup, it's unlikely we'll run into many CPU-bound tasks simultaneously so I think the default dispatcher is still acceptable in this case (though we may want to consider the distinctions of CPU-bound tasks I mention above).

@colintheshots @hawkinsw Any thoughts here?

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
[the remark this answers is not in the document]
mcomella: afaik, SharedPrefs executes on the thread that you call it on. afaik, it's thread safe – so you can use it from multiple threads – but they use different wording. From SharedPrefs reference docs:

> Note: This class provides strong consistency guarantees. It is using expensive operations which might slow down an app. Frequently changing properties or properties where loss can be tolerated should use other mechanisms. For more details read the comments on SharedPreferences.Editor.commit() and SharedPreferences.Editor.apply().

My concern is that the Settings class was not designated as thread safe by the developers who wrote it so, in the future, it may not enforce the thread safety SharedPreferences has.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
mcomella: CC @MarcLeclair so all the FE perf folks have context on this discussion. :) It seems like an important one

app/src/main/java/org/mozilla/fenix/settings/DefaultBrowserPreference.kt
```diff
@@ -34,4 +37,18 @@ class DefaultBrowserPreference @JvmOverloads constructor(
         val browsers = Browsers.all(context)
         switchView?.isChecked = browsers.isDefaultBrowser
     }
+
+    fun unsetOpenLinksInAPrivateTabIfNecessary() {
```
emalysz: @mcomella I ended up moving the shared method to DefaultBrowserPreference, but let me know if you think there is a better alternative

MarcLeclair: Although I don't know Fenix that much, to me it makes sense to call this only when we unset ourselves as the default browser. However,  @mcomella  or @emalysz correct me if I'm wrong, what happens with the following use case: User kills browser -> goes in Settings/apps/default apps/ -> removes Fenix as default browser -> re-opens Fenix. Then we don't go through this loop as we start I think, right? I'm not too sure, which is why I'm asking 😛

emalysz: @MarcLeclair  I believe that use case is picked up by calling unsetOpenLinksInAPrivateTabIfNecessary (at least it did when I tested it), but it would be great if @mcomella could confirm :)

mcomella: @MarcLeclair afaict, the preference is used in two places:
- In the settings, to display the preference & provide the option to change it
- in IntentReceiverActivity, to use the preference, i.e. determine behavior based on it

This means we can update it:
- (current impl) Every time we open the app (so the most recent value is always available)
  - Caveat: Android doesn't have a single entry point so it's challenging to do this correctly
- (proposed impl) "just-in-time", right before we need to use it

mcomella: > I ended up moving the shared method to DefaultBrowserPreference, but let me know if you think there is a better alternative

I think we should move it somewhere else because `DefaultBrowserPreference` is a view: that means 1) we're initializing irrelevant code to call this method and 2) mixing our view and model layers (since the new code is at the model layer).

I think we could move it to `Settings`: there's some precedent there.

app/src/main/java/org/mozilla/fenix/settings/DefaultBrowserPreference.kt
```diff
@@ -34,4 +37,18 @@ class DefaultBrowserPreference @JvmOverloads constructor(
...
+        if (!Browsers.all(this.context).isDefaultBrowser) {
+            /* If the user has unset us as the default browser, unset openLinksInAPrivateTab */
```
mcomella: nit: this comment is redundant to the code

app/src/main/java/org/mozilla/fenix/settings/DefaultBrowserPreference.kt
```diff
@@ -34,4 +37,18 @@ class DefaultBrowserPreference @JvmOverloads constructor(
         val browsers = Browsers.all(context)
         switchView?.isChecked = browsers.isDefaultBrowser
     }
+
+    fun unsetOpenLinksInAPrivateTabIfNecessary() {
+        if (!Browsers.all(this.context).isDefaultBrowser) {
```
mcomella: nit: I prefer to exit early than heavily nest:
```kotlin
if (Browser.all(this.context).isDefaultBrowser) {
    return
} 

// do thing
```

app/src/main/java/org/mozilla/fenix/settings/DefaultBrowserSettingsFragment.kt
```diff
@@ -89,6 +89,8 @@ class DefaultBrowserSettingsFragment : PreferenceFragmentCompat() {
             isChecked = context.settings().openLinksInAPrivateTab
             onPreferenceChangeListener = SharedPreferenceUpdater()
         }
+
+        DefaultBrowserPreference(requireContext()).unsetOpenLinksInAPrivateTabIfNecessary()
```
mcomella: I think this needs to be called before `findPreference<CheckBox...` so that `openLinksInAPrivateTab` has the latest value.

app/src/main/java/org/mozilla/fenix/settings/DefaultBrowserSettingsFragment.kt
```diff
@@ -89,6 +89,8 @@ class DefaultBrowserSettingsFragment : PreferenceFragmentCompat() {
             isChecked = context.settings().openLinksInAPrivateTab
```
mcomella: No action needed: I noticed the `Browsers` initializer is pretty slow (20ms on my Pixel 2) and we're fetching it three times in this method: `updateSwitch`, the explicit usage, and in `unsetOpenLinks...`. Optimizing `DefaultBrowserSettingsFragment` is not our priority at the moment but I wanted to raise it.

The `Browsers` impl may do some internal caching that might mitigate the repeated access though.

app/src/main/java/org/mozilla/fenix/settings/DefaultBrowserPreference.kt
```diff
@@ -34,4 +37,18 @@ class DefaultBrowserPreference @JvmOverloads constructor(
...
+            /* If the user has unset us as the default browser, unset openLinksInAPrivateTab */
+            this.context.settings().openLinksInAPrivateTab = false
+            context?.components?.analytics?.metrics?.track(
```
mcomella: Not your fault but I think this metrics tracking is wrong: it says we toggled a preference every time we get an intent and we're not the default browser. If you agree, can you file a bug?

Furthermore, the docs say, "A user toggled a boolean preference in settings". However, this is 1) not being toggled by a user and 2) is not toggled in Settings.

We'll keep the functionality the same in this PR so the Fenix team can address it. Please mention we are changing the code but intend to keep the functionality the same so they don't blame us. :P

app/src/main/java/org/mozilla/fenix/IntentReceiverActivity.kt
```diff
@@ -36,15 +36,7 @@ class IntentReceiverActivity : Activity() {
...
-        if (!Browsers.all(this).isDefaultBrowser) {
-            /* If the user has unset us as the default browser, unset openLinksInAPrivateTab */
-            this.settings().openLinksInAPrivateTab = false
-            components.analytics.metrics.track(Event.PreferenceToggled(
-                preferenceKey = getString(R.string.pref_key_open_links_in_a_private_tab),
-                enabled = false,
-                context = applicationContext
-            ))
-        }
+        DefaultBrowserPreference(applicationContext).unsetOpenLinksInAPrivateTabIfNecessary()
```
mcomella: For later: I realized that we improved the performance of `HomeActivity.onResume` but `IntentReceiverActivity`, used for opening links from external apps and homescreen icons, still has this same performance hit. Can you file a follow-up bug?

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -30,7 +30,6 @@ import mozilla.components.service.fxa.sync.SyncReason
 import mozilla.components.support.base.feature.UserInteractionHandler
```
mcomella: nit: a few notes on commit best practices:
- Commit messages should be manually wrapped around 80 characters; this is so they display correctly in a terminal. You can use the extended commit summary to provide more detailed information.
- When you make changes during a review as an additional commit, unless it's sufficiently complicated that the reviewer wants to see the diff (and sometimes they'll request this) or it'll take you too long to do it, it's often a good idea to squash them so there are no "fix nits" commits in the history (we don't usually squash commits when merging).
  - If the commit to change is on top, you can do `git commit --amend` to modify it
  - If the commit is somewhere in the middle, you have to do an interactive rebase: `git rebase -i <some-commit-above-the-commits-to-change>`; I usually do `git rebase -i HEAD~10` where 10 is some large number :)
- All of your commits should include the issue number (the way we merge is completely linear so you can't distinguish a PR without it

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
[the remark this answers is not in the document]
hawkinsw: I just wanted to add to this discussion:

@MarcLeclair and I were hoping that we could move execution of GleanMetrics.setStartupMetrics() off of the main thread and that would also involve using the Settings class off the main thread. I think that means we will need to come to a conclusion about whether or not we can use Settings off MT. 

I think that there are two related issues:
1. Whether Settings can be used off MT (SharedPrefs itself is usable off MT, iirc)
2. Whether Setting is thread safe (SharedPrefs itself is thread safe, iirc)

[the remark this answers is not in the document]
mcomella: If we identify that Settings is thread safe, imo it's within our authority to add a comment explaining that it's thread safe and therefore making others enforce those guarantees. :)

app/src/main/java/org/mozilla/fenix/settings/DefaultBrowserPreference.kt
emalysz: https://github.com/mozilla-mobile/fenix/issues/7384

app/src/main/java/org/mozilla/fenix/IntentReceiverActivity.kt
```diff
@@ -36,17 +35,10 @@ class IntentReceiverActivity : Activity() {
...
-            /* If the user has unset us as the default browser, unset openLinksInAPrivateTab */
-            this.settings().openLinksInAPrivateTab = false
-            components.analytics.metrics.track(Event.PreferenceToggled(
-                preferenceKey = getString(R.string.pref_key_open_links_in_a_private_tab),
-                enabled = false,
-                context = applicationContext
-            ))
-        }
+        val settings = settings()
+        settings.unsetOpenLinksInAPrivateTabIfNecessary()
```
mcomella: The perf issue still affects app link, custom homescreen shortcuts, etc.: can you file a follow-up to see if we can improve this for these use cases?

PR #7249, opened 2019-12-17
Use localized strings in the quick settings panel

darkwing:
---

Noticed that some of these are hardcoded -- we should localize instead.


darkwing:
Oh weird; feel free to close if not desired.

NotWoods:
I'll merge since the localized equivalents of the strings already exist, so it's not a big deal

Issue #6836, opened 2019-11-27
Limit Leanplum usage by language (Mirror Fennec)


pocmo:
Fennec does not enable Leanplum for all users. Instead it uses a Switchboard experiment. The switchboard experiment enables Leanplum only for specific languages, see `leanplum-start`:
https://firefox.settings.services.mozilla.com/v1/buckets/fennec/collections/experiments/records

We should mirror this behavior before migrating Fennec users. Otherwise we add a lot more users to Leanplum and afaik that will cost us a lot of money.

st3fan:
@lwa-moz Loren I know you are investigating removing the locale limitation in Fennec, to reach more people _before_ the migration. Can you let us know if that would be temporary or permanent?  Right now this limitation does not exist in Fenix, which would mean that after the migration, the whole user base would be on Leanplum.

@pocmo If we do this then I would suggest to not depend on Switchboard and instead use a hard-coded list. That is a lot simpler, removes some tricky timing issues and may work just as well since we don't really update that list very often.

lwa-moz:
It's my understanding that we are limiting our reach in Leanplum by language (regardless of geo). We are enabling Leanplum in Fennec for users who have set their device language to one of the following: EN, ZH, DE, FR, IT, ID, PT, ES, PL, RU

After detailed discussion with the Marketing team, we would like to add the following languages to our list: HI, FA, AR, JA

Adding these countries will give us the best balance of cost, marketing capability (from a localization and campaign implementation standpoint) and reach.

Mugurell:
For QA: 
Although this ticket appears on the migration board it is not actually about migrating anything.
We just need Fenix to be available only to a small subset of locales, the same as Fennec. But nothing is migrated from Fennec to Fenix.
So to test this would be useful to just know that indeed Leanplum gets enabled for the locales from https://github.com/mozilla-mobile/fenix/issues/6836#issuecomment-564601017 and not for any others.

abodea:
### Hello, note that I tested this issue on the latest Nightly build from **1/10** for more details please check the information from below:

| Locale |  Events displayed?| Total Occurrences modified? |
| :---         |     :---: |  :---: |
|  **"eng", // English**   |  ✅   |      Yes      |
|  **"zho", // Chinese**    |  ✅     |   Yes      |
|  **"deu", // German**    | ✅      |    Yes     |
| **"fra", // French**     |  ✅      |    Yes    |
[...]

PR #7273, opened 2019-12-18
For #6836 - Limit Leanplum depending on current locale.

Mugurell:
### Pull Request checklist

- [x] **Tests**: Not yet tested component.
- [x] **Screenshots**: No visible user change.

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/metrics/LeanplumMetricsService.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/metrics/LeanplumMetricsService.kt
@@ -10,8 +10,10 @@ import com.leanplum.Leanplum
 import com.leanplum.LeanplumActivityHelper
 import com.leanplum.annotations.Parser
 import com.leanplum.internal.LeanplumInternal
+import mozilla.components.support.locale.LocaleManager
 import org.mozilla.fenix.BuildConfig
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/fenix/components/metrics/LeanplumMetricsService.kt
```diff
@@ -54,6 +55,12 @@ class LeanplumMetricsService(private val application: Application) : MetricsServ
     private val token = Token(LeanplumId, LeanplumToken)
 
     override fun start() {
+        val currentLocale = Locale.getDefault().isO3Language
```
Baron-Severin: I have definitely used `Locale.getDefault`, many times in our codebase, but I _believe_ we should be using `LocaleManager.getCurrentLocale` here.  I know that the `LocaleManager` will be updated when a user changes language in-app, and I'm not sure whether or not `Locale.getDefault` will.

Mugurell: Testing `Locale.getDefault().isO3Language` I saw it returning the expected language (system default/app set language) but I agree that being more explicit and also having more certainty about the result is better.
Implemented this and rebased.
Thanks!

PR #7301, opened 2019-12-19
Closes #7085 - Remove anko version in Dependencies.kt

### LINKED ISSUE #7085: Consider moving away from anko
Anko has been deprecated and will not be continued:
https://github.com/Kotlin/anko/blob/master/GOODBYE.md

There are some areas where we are using anko:
https://github.com/mozilla-mobile/fenix/search?q=anko&unscoped_q=anko

I can see at least:
- Some helpers like `px2dip`. There should be an equivalent helper in AC or *-ktx from Androidx. If not then maybe we can land one in `support-ktx` or `support-utils`.
- Layout DSL: We use it to layout components dynamically. In this case I wonder if we should revisit whether we actually need that: The initial idea was to allow dynamic layouts in cases of A/B tests. So I wonder: Is that the kind of A/B test that we are actually doing? Is it worth the complexity by default? Maybe we can get back to a simpler way of layouting more statically. There are alternatives to anko in this area, but I'd really consider not piling up other framework stacks.

NotWoods:
Removes anko now that we don't use it in the codebase.

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/buildSrc/src/main/java/Dependencies.kt
+++ b/buildSrc/src/main/java/Dependencies.kt
@@ -9,7 +9,6 @@ object Versions {
     const val rxAndroid = "2.1.0"
     const val rxKotlin = "2.3.0"
     const val rxBindings = "3.0.0-alpha2"
-    const val anko = "0.10.8"
     const val sentry = "1.7.10"
     const val leakcanary = "2.0"
     const val leanplum = "5.2.3"


jonalmeida:
@NotWoods we ended up merging mcomella's PR, sorry about that!

There is still this https://github.com/mozilla-mobile/fenix/pull/7433#issuecomment-569964668 to address if you'd like. 🙂

PR #7306, opened 2019-12-20
Add mark-as-shipped task

Callek:
For ship-it based Fenix releases, we'll need to have a task to say, that yes, we did finish shipping.

The dependency on pushapk makes sense in this regard. Though I have not yet fully tested this code locally, I suspect it should be good as-is, I still want @JohanLorenzo  to :+1: first

### DIFF
--- a/taskcluster/ci/mark-as-shipped/kind.yml
+++ b/taskcluster/ci/mark-as-shipped/kind.yml
@@ -1,0 +1,32 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+---
+loader: fenix_taskgraph.loader.multi_dep:loader
+
+transforms:
+    - fenix_taskgraph.transforms.multi_dep:transforms
+    - fenix_taskgraph.transforms.mark_as_shipped:transforms
+    - taskgraph.transforms.task:transforms
+
+kind-dependencies:
+    - push-apk
+
+primary-dependency: push-apk
+
+group-by: build-type
+
+only-for-build-types:
+    - beta
+    - production
+
+job-template:
+    description: Mark Fenix as shipped in ship-it
+    worker-type: ship-it
+    scopes:
+        - project:releng:ship-it:action:mark-as-shipped
+    worker: {}
+
+    treeherder:
+        job-symbol: mark-shipped
+        kind: build
--- a/taskcluster/fenix_taskgraph/transforms/mark_as_shipped.py
+++ b/taskcluster/fenix_taskgraph/transforms/mark_as_shipped.py
@@ -1,0 +1,28 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+"""
+Apply some defaults and minor modifications to the jobs defined in the build
+kind.
+"""
+
+from __future__ import absolute_import, print_function, unicode_literals
+
+from taskgraph.transforms.base import TransformSequence
+from taskgraph.util.schema import resolve_keyed_by
+
+
+transforms = TransformSequence()
+
+
+@transforms.add
+def make_task_description(config, jobs):
+    for job in jobs:
+        product = "Fenix"
+        version = config.params['release_version'] or "{ver}"
+        job['worker']['release-name'] = '{product}-{version}-build{build_number}'.format(
+            product=product,
+            version=version,
+            build_number=config.params.get('build_number', 1)
+        )
+        yield job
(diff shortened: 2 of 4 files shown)


taskcluster/fenix_taskgraph/worker_types.py
```diff
@@ -94,3 +94,26 @@ def build_push_apk_payload(config, task, task_def):
             scope_prefix, worker["product"], ":dep" if worker["dep"] else ""
         )
     )
+
+
+@payload_builder(
+    "scriptworker-shipit",
+    schema={
+        Required("upstream-artifacts"): [
```
JohanLorenzo: Nit: Do we need `upstream-artifacts` here? I don't see it defined in gecko, for instance https://searchfox.org/mozilla-central/rev/b243debf6235b050b42fd2eb615fdc729636ca6b/taskcluster/taskgraph/transforms/task.py#1212-1220

taskcluster/fenix_taskgraph/transforms/mark_as_shipped.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+transforms = TransformSequence()
+
+
+@transforms.add
+def make_task_description(config, jobs):
+    for job in jobs:
+        product = "Fenix"
+        version = config.params['release_version'] or "{ver}"
```
JohanLorenzo: Question: What is `"{ver}"` in this context? I don't see it defined in gecko https://searchfox.org/mozilla-central/rev/b243debf6235b050b42fd2eb615fdc729636ca6b/taskcluster/taskgraph/transforms/release_mark_as_shipped.py

taskcluster/fenix_taskgraph/transforms/mark_as_shipped.py
```diff
@@ -0,0 +1,28 @@
...
+        job['worker']['release-name'] = '{product}-{version}-build{build_number}'.format(
+            product=product,
+            version=version,
+            build_number=config.params.get('build_number', 1)
```
JohanLorenzo: Good call on defaulting build_bumber while relpro lands 👍

taskcluster/ci/mark-as-shipped/kind.yml
```diff
@@ -0,0 +1,32 @@
...
+group-by: build-type
+
+only-for-build-types:
+    - beta
+    - production
+
+job-template:
+    description: Mark Fenix as shipped in ship-it
+    worker-type: ship-it
+    scopes:
```
JohanLorenzo: Question: Do we want to use a staging/production scope like gecko does? https://searchfox.org/mozilla-central/source/taskcluster/ci/release-mark-as-shipped/kind.yml#26-33

taskcluster/fenix_taskgraph/worker_types.py
Callek: I added upstream-artifacts merely because the multi_dep transforms add it, and I didn't think it was worth removing it.... running the full graph I didn't see them used. Am I wrong in that assumption for here?

taskcluster/fenix_taskgraph/transforms/mark_as_shipped.py
Callek: I used it as a clear "this isn't right" designation, because default `release_version` is `""` so it was generating a `Fenix--build1` thing by default. -- I can remove it if this is less clear.

taskcluster/fenix_taskgraph/transforms/mark_as_shipped.py
Callek: It was also because the full graph will still use this bit of code, even when not part of ship-it taskgraph. So defaulting ensures we get "something" here thats not busted.

taskcluster/ci/mark-as-shipped/kind.yml
Callek: Probably - I just wasn't sure of any `by-*` that would work here to achieve that.

sblatz:
@Callek is this good for me to land? 😄

Callek:
@sblatz yea, lets land it, if there are followups that @JohanLorenzo wants we can do them in a different PR - Thanks!

taskcluster/fenix_taskgraph/worker_types.py
JohanLorenzo: Your approach makes sense to me 🙂 I think we should leave a comment that explains why it's here while not used. This way, we won't confuse future readers.

taskcluster/fenix_taskgraph/transforms/mark_as_shipped.py
JohanLorenzo: Ah, I see! I agree the magic string method is a good way to go. Maybe we should set it to something that shows `release_version` was missing and it's obviously wrong. How about something like `"MISSING_VERSION"`?

taskcluster/ci/mark-as-shipped/kind.yml
JohanLorenzo: Got it! I agree `by-release-level` won't work since it's gecko-specific. `by-level` would, I think. That said, we can add this scope later.

JohanLorenzo:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1608050 for followup fix

Issue #7041, opened 2019-12-09
Opening history items and bookmarks doesn't respect current mode


sblatz:
If you're in private mode and open links, they will open in normal mode. 

Reported in https://github.com/mozilla-mobile/fenix/issues/6472#issuecomment-563400287

shahsurajk:
Can i try this?

NotWoods:
Go for it!

AndiAJ:
Hi, **verified as fixed** on the latest **Nightly Build #13580607 from 12/24** using the following devices:
• Google Pixel 3a (Android 9)
• Huawei Mate 20 Lite (Android 8.1.0)
• Samsung Galaxy S7 (Android 7)
• OnePlus A3 (Android 6.0.1)
• LG Nexus 4 (Android 5.1.1)

► Video
![20191224_103442](https://user-images.githubusercontent.com/51314259/71404134-67397980-263a-11ea-8b7c-5ecf33b4f88f.gif)

PR #7311, opened 2019-12-20
For #7041 - Respect current mode when opening library items 

shahsurajk:
### Pull Request checklist

### After merge

1. click on Show All Checks,

BrowsingMode is respected in private mode, written test case around the same as well to be confident. 

Added fix for opening links via bookmarks

### DIFF
--- a/app/src/test/java/org/mozilla/fenix/library/bookmarks/BookmarkControllerTest.kt
+++ b/app/src/test/java/org/mozilla/fenix/library/bookmarks/BookmarkControllerTest.kt
@@ -107,7 +107,7 @@ class BookmarkControllerTest {
     }
 
     @Test
-    fun `handleBookmarkTapped should respect browsing mode`(){
+    fun `handleBookmarkTapped should respect browsing mode`() {
         // if in normal mode, should be in normal mode
         every { homeActivity.browsingModeManager.mode } returns BrowsingMode.Normal
 
--- a/app/src/test/java/org/mozilla/fenix/library/bookmarks/BookmarkControllerTest.kt
+++ b/app/src/test/java/org/mozilla/fenix/library/bookmarks/BookmarkControllerTest.kt
@@ -22,6 +22,7 @@ import io.mockk.verifyOrder
 import mozilla.appservices.places.BookmarkRoot
 import mozilla.components.concept.storage.BookmarkNode
 import mozilla.components.concept.storage.BookmarkNodeType
+import org.junit.Assert
 import org.junit.Before
 import org.junit.Test
 import org.mozilla.fenix.BrowserDirection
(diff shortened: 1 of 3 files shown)


app/src/test/java/org/mozilla/fenix/library/bookmarks/BookmarkControllerTest.kt
```diff
@@ -101,11 +102,25 @@ class BookmarkControllerTest {
...
         }
     }
 
+    @Test
+    fun `handleBookmarkTapped should respect browsing mode`() {
+        // if in normal mode, should be in normal mode
+        every { homeActivity.browsingModeManager.mode } returns BrowsingMode.Normal
+
+        controller.handleBookmarkTapped(item)
+        Assert.assertEquals(BrowsingMode.Normal.isPrivate, homeActivity.browsingModeManager.mode.isPrivate)
```
NotWoods: nit:
```suggestion
        assertEquals(BrowsingMode.Normal, homeActivity.browsingModeManager.mode)
```

app/src/test/java/org/mozilla/fenix/library/history/HistoryControllerTest.kt
```diff
@@ -143,4 +146,25 @@ class HistoryControllerTest {
...
     }
+
+    @Test
+    fun `openHistoryItem respects browsing mode`() {
+        val homeActivity = mockk<HomeActivity>()
+
+        every { homeActivity.browsingModeManager.mode } returns BrowsingMode.Private
+        DefaultHistoryController(
+            mockk(),
+            {
```
NotWoods: nit: it's easier to read the test if named arguments are used

app/src/test/java/org/mozilla/fenix/library/history/HistoryControllerTest.kt
```diff
@@ -143,4 +146,25 @@ class HistoryControllerTest {
...
+                homeActivity.openToBrowserAndLoad(
```
NotWoods: This test seems to just check that the lambda is called, so I think `onPressHistoryItemInNormalMode` already covers this. 

We might want to move `HistoryFragment#openItem` into the controller so you can test it, but that can be left for the future.

shahsurajk: I agree, the mechanism used for Bookmarks actually fits this. 

Regarding the redundant check, do you think it would be good to remove the `openToBrowserAndLoad` method?  Cause the original intent of the test is to prove that if a new tab is opened from history, then the browsing mode is not changed.

app/src/test/java/org/mozilla/fenix/library/bookmarks/BookmarkControllerTest.kt
shahsurajk: yeah, this just seems so obvious, no clue why i consumed the `isPrivate` thing

app/src/test/java/org/mozilla/fenix/library/history/HistoryControllerTest.kt
NotWoods: We should remove the method, unless you feel like moving `openItem` to the controller (like Bookmarks). 

If you decide to move `openItem` then this is a good test to keep, otherwise you're really only testing the mock instead of the controller.

shahsurajk: Yes. Make sense. It does actually look like I'm testing the mock! Will remove this.

shahsurajk: Fixed. Sorry for resolving the conversation before actually pushing a commit. I realized later.

Issue #6464, opened 2019-11-05
Upgrade forPerformanceTest to "first class variant"


liuche:
### Why/User Benefit/User Problem
Performance tests must be run on builds that are as close as possible to release with the debuggable flag set.

The performance build variant was added because the debug variant does not have telemetry enabled (#6447 for enabling it). However, e.g. if we complete that issue, we may be able to remove the variant: let's investigate!

If we can't remove it, we should make sure it's treated as a proper build variant, e.g. by making code using it exhaustive.

### Impact
When making changes, the performance build variant is a special case and can be confusing, especially for non-performance engineers. For example, this may have caused a crash:

> it's because when we changed the build types we missed updating the build names somewhere else.
>
> (because the string was using old build variants)

That's `Config.channel`: https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/java/org/mozilla/fenix/Config.kt#L40

### Acceptance Criteria (how do I know when I’m done?)
- Decide if the performance variant should be removed. Consider:
  - Should perf builds be minified/proguarded? If not, we may be able to use the debug variant
  - That non-Android perf eng should be able to easily make a build without modifying source code
- If we decide to remove it:
  - Ensure replacement variant is capable of doing perf runs
  - Remove perf variant
  - Provide instructions for perf eng to make a build without modifying source code (can be as simple as specifying a build variant in the readme)
- If we decide to keep it:
  - Ensure all uses of BuildConfig are exhaustive. See below for an example
  - Ensure [`Config.channel`](https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/java/org/mozilla/fenix/Config.kt#L40) is exhaustive (i.e. perf is included and the `else` case is a crash).

---

Note: [this was the patch](https://github.com/mozilla-mobile/fenix/pull/6462/files) that fixed the issue.

mcomella:
I did an analysis into our build variants (https://github.com/mozilla-mobile/fenix/issues/6931) and we decided to keep `forPerformanceTest`: repurposing this bug to make it a first class variant. This includes:
- Disable StrictMode logging in forPerformanceTest builds
- Fix captured Exception for “invalid” variant forPerformanceTest (i.e. the issue described in comment 0)
- Investigate if further work is needed

mcomella:
Raising priority as a follow-up to my recommendations on build variants: https://github.com/mozilla-mobile/fenix/issues/6931

mcomella:
[...]

PR #7319, opened 2019-12-20
For #6464: make "forPerformanceTest" a "first class" variant

mcomella:
### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/Config.kt
+++ b/app/src/main/java/org/mozilla/fenix/Config.kt
@@ -4,8 +4,6 @@
 
 package org.mozilla.fenix
 
-import io.sentry.Sentry
-
 enum class ReleaseChannel {
     FenixDebug, FenixNightly, FenixBeta, FenixProduction, FennecProduction;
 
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/mozilla/fenix/Config.kt
```diff
@@ -47,10 +45,15 @@ object Config {
         "fenixNightly" -> ReleaseChannel.FenixNightly
         "debug" -> ReleaseChannel.FenixDebug
         "fennecProduction" -> ReleaseChannel.FennecProduction
+
+        // Builds for local performance analysis, recording benchmarks, automation, etc.
+        // This should be treated like a released channel because we want to test
+        // what users experience and there are performance-impacting changes in debug
+        // release channels (e.g. logging) that are never intended to be shipped.
+        "forPerformanceTest" -> ReleaseChannel.FenixProduction
+
```
hawkinsw: I actually am not sure what to think about this. I am debating whether this should be a separate enum value entirely. 

Obviously the performance test variant should mimic a released variant, but I am not sure that they should be equal. 

What do you think about making the performance variant a separate enum and then modifying the enum class's methods to take that into account when determining the values of isRelease() and isDebug() etc?

app/src/main/java/org/mozilla/fenix/Config.kt
```diff
@@ -47,10 +45,15 @@ object Config {
...
         else -> {
-            Sentry.capture("BuildConfig.BUILD_TYPE ${BuildConfig.BUILD_TYPE} did not match expected channels")
-            // Performance-test builds should test production behaviour
-            ReleaseChannel.FenixProduction
+            throw IllegalStateException("Unknown build type: ${BuildConfig.BUILD_TYPE}")
```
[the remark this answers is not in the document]
mcomella: Yes – I almost always prefer exhaustive matching. :)

app/src/main/java/org/mozilla/fenix/Config.kt
mcomella: I originally wrote it that way and decided not to because, if additional methods are added or someone checks `Config.channel` directly, they'd almost certainly not take into account the performance variant.

If we were to make it a separate enum value, I'd want to refactor the code such that `Config.channel` is not exposed so the only entry points are those `isReleaseOrDebug` wrapping functions and the `when` statements are exhaustive so devs will, by following convention, have to consider the `forPerformanceTest` builds.

I think I'd prefer that, tbh, but felt some resistance (from myself) towards doing the refactor – e.g. something I didn't understand or might take too long to do – but I can go back and do it if you'd like.

hawkinsw: I think that we should make it a separate enum value and do the refactor. However, I guess the expediency/easiness argument will win.

mcomella: I filed https://github.com/mozilla-mobile/fenix/issues/7416 for the follow-up, thanks.

Issue #7352, opened 2019-12-24
Integrate highlightable menu button changes from A-C


Baron-Severin:
Make browser menu items highlightable as designed in https://github.com/mozilla-mobile/fenix/issues/4281.  This AC was originally pushed out due to complications with the menu API which have since been resolved.

#### Acceptance Criteria
- [ ] Reader Mode, Reader Mode Appearance, PWA, and Open in App buttons should all be highlighted in the menu bar whenever they are visible
- [ ] The three dot button that opens the menu bar should be highlighted any time any of the above buttons are available

#### Engineering notes
- We should use BrowserMenuHighlightableSwitch ([added by this AC PR](https://github.com/mozilla-mobile/android-components/issues/5022)) to highlight Reader Mode

softvision-miralobontiu:
Hi all!

I've verified that "Reader Mode", "PWA", and "Open in App" buttons are highlighted in the menu bar whenever they are visible. The three-dot button that opens the menu bar is highlighted any time any of the above buttons are available.
However, the "Reader Mode Appearance" is not highlighted in the Settings menu (as seen in the screenshot below). @Baron-Severin Is this expected? 
Tested on the latest Nightly build from 01/06 (#20060606) with Google Pixel 3 (Android 10), and HTC Desire 820 (Android 6.0.1).

![Screenshot_20200106-105714](https://user-images.githubusercontent.com/40694158/71807394-e8016800-3073-11ea-8679-e14576037b4e.png)

I'll remove the qa-needed flag for now.

topotropic:
@Baron-Severin can we also have a blue dot on the "new tab / home" menu whenever we have a release and a blue dot next to "What's new"? Thanks!

Baron-Severin:
My apologies @softvision-miralobontiu, after demo UX requested that the dot be removed from `Appearance`, but I neglected to update the ACs.  Updating them now.  I'm going to close the story, please feel free to reopen if you have any other issues with it.

@topotropic I opened https://github.com/mozilla-mobile/fenix/issues/7506 to track that change, would you mind taking a look at it and making sure I understood correctly?

PR #7353, opened 2019-12-24
For #7352: integrate highlightable browser menu changes

Baron-Severin:
### Pull Request checklist

### After merge

1. click on Show All Checks,

highlights1
highlights2

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
@@ -14,6 +14,7 @@ import mozilla.components.browser.menu.BrowserMenuBuilder
 import mozilla.components.browser.menu.BrowserMenuHighlight
 import mozilla.components.browser.menu.item.BrowserMenuDivider
 import mozilla.components.browser.menu.item.BrowserMenuHighlightableItem
+import mozilla.components.browser.menu.item.BrowserMenuHighlightableSwitch
 import mozilla.components.browser.menu.item.BrowserMenuImageSwitch
 import mozilla.components.browser.menu.item.BrowserMenuImageText
 import mozilla.components.browser.menu.item.BrowserMenuItemToolbar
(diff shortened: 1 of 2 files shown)


Baron-Severin:
Need to look into a problem with this before it can be merged.  The three-dot menu is supposed to be highlighted any time any of its items can be, but this isn't holding true.

STR
- Open new tab
- Navigate to https://onezero.medium.com/irl-ads-are-taking-scary-inspiration-from-social-media-7088e8241beb
- Wait for load

Expected
- In the browser menu, Reader View is highlighted
- The three dot menu is highlighted

Actual
- In the browser menu, Reader View is highlighted
- The three dot menu is **_not_** highlighted

NotWoods:
You need to call `invalidate` on the toolbar menu so it checks the list again (not a great API right now...sorry!). There's a helper class that just calls invalidate in certain conditions.

app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
```diff
@@ -277,26 +278,38 @@ class DefaultToolbarMenu(
...
+        highlight = BrowserMenuHighlight.LowPriority(
+            label = context.getString(R.string.browser_menu_read),
+            notificationTint = getColor(context, R.color.whats_new_notification_color)
+        )
     ) { checked ->
         onItemTapped.invoke(ToolbarMenu.Item.ReaderMode(checked))
     }
 
-    private val readerAppearance = BrowserMenuImageText(
+    private val readerAppearance = BrowserMenuHighlightableItem(
```
NotWoods: Does appearance also need a highlight? It feels redundant.

Baron-Severin: @apbitner what do you think?  I have no strong feelings either way on this one.

app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
```diff
@@ -14,6 +14,7 @@ import mozilla.components.browser.menu.BrowserMenuBuilder
 import mozilla.components.browser.menu.BrowserMenuHighlight
```
Baron-Severin: (unrelated to this LoC, just putting it here so we can chain comments)

> You need to call `invalidate` on the toolbar menu so it checks the list again (not a great API right now...sorry!). There's a helper class that just calls invalidate in certain conditions.

@NotWoods hmm, I don't know if we can.  We don't have a reference to the view here, and it's a required parameter for invalidate.  I remember going through this problem with Christian a few months ago and being more or less stumped.  Do you have any suggestions?

NotWoods: You could override `onReaderableStateUpdated` in `MenuPresenter`

Baron-Severin: Works great, thanks for the help!

NotWoods:
Code looks good, I'll approve once we get a UX decision on the extra appearance dot.

app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
apbitner: I don't think it's necessary. 'Appearance' may not live in the toolbar menu permanently anyways.

Baron-Severin: Thanks Adam, I'll pull it out.

app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
```diff
@@ -277,10 +278,14 @@ class DefaultToolbarMenu(
...
+    private val readerMode = BrowserMenuHighlightableSwitch(
         label = context.getString(R.string.browser_menu_read),
-        imageResource = R.drawable.ic_readermode,
-        initialState = readerModeStateProvider
+        startImageResource = R.drawable.ic_readermode,
+        initialState = readerModeStateProvider,
+        highlight = BrowserMenuHighlight.LowPriority(
+            label = context.getString(R.string.browser_menu_read),
+            notificationTint = getColor(context, R.color.whats_new_notification_color)
+        )
```
NotWoods: `isHighlighted` should also be set. I don't think the menu currently checks visibility.

Baron-Severin: `isHighlighted` defaults to true, which is what we always want here.  But it's probably not a bad idea to put it in here to be explicit.

NotWoods: If you include these options in the toolbar list when reader mode is unavailable, then the highlight may always appear.

Baron-Severin: Oh, I see what you mean.  Thanks for clarifying.

Baron-Severin: We looked into this and it seems that it does filter is visibility when adding the highlight to the three dot menu, so I'm resolving this comment.

Issue #6413, opened 2019-11-03
[Bug] Snackbar messages cover address bar


bxtom:
Message placement is inconsistent. On the left screenshot is ok, on the right message covers address bar. I think message placement should be corrected (moved up).

![MessagePlacement](https://user-images.githubusercontent.com/29851015/68085018-fba60d80-fe3b-11e9-990f-c5609aced67e.png)

* Android device: android 9
* Fenix version: 2.2.0 (Build #12901947)

sblatz:
This one is a bit complicated since the ShareFragment is a separate fragment from BrowserFragment and cannot access the browserToolbar as an anchor. I'm digging into this.

sblatz:
Note: this also happens with "Syncing"

sblatz:
Have a fix up for review now; I believe I fixed this is all locations 😄

AndiAJ:
Hi, I've just checked this matter on the latest **Nightly Build #13650606 from 12/31** using a Google Pixel 3a (Android 9)

| Status  | Description|
| ------------- | ------------- |
| ✔️ | Close a tab from the Home screen - Open tabs section  |
| ✔️ | Close all tabs from the Home screen - Open tabs section  |
| ✔️ | Create a new Collection  |
| ✔️ | Rename a collection |
| ✔️ | Save a page to a collection from the Navigation bar (3dot button overlay)  |
| ✔️ | Add a bookmark |
| ✔️ | Delete a bookmark  |
| ✔️ | Send tab |
| ✔️ | Delete browsing data on quit (from the home screen)  |
| ❌ | Delete browsing data on quit while having a website opened (from the Navigation bar- 3dot button overlay) |
| ❌ | Delete browsing data  |
| ❓ | Syncing (After the user successfully logs in and the auto sync kicks in)|

#### ►Full run video
![20191231_100910](https://user-images.githubusercontent.com/51314259/71615369-9e1f0a80-2bb9-11ea-8a34-bfd4302457ca.gif)


#### ❌ Delete browsing data on quit while having a website opened (from the Navigation bar- 3dot button overlay) 

![20191231_101357](https://user-images.githubusercontent.com/51314259/71615734-5d27f580-2bbb-11ea-9fcb-880bd98fb467.gif)

#### ❌ Delete browsing data
![20191231_101524](https://user-images.githubusercontent.com/51314259/71615847-f3f4b200-2bbb-11ea-9c18-565983e2d448.gif)

[...]

PR #7415, opened 2019-12-30
For #6413: Fixes snackbar positionining for bottom toolbar

sblatz:
New behavior:

demo snackbar 2019-12-30 10_16_48

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/FenixSnackbar.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/FenixSnackbar.kt
@@ -148,7 +148,11 @@ private class FenixSnackbarCallback(
     }
 }
 
-class FenixSnackbarPresenter(
+/**
+ * This snackbar presenter should be used when displaying a snackbar that will appear in
+ * the BrowserFragment as it takes into account the position of the BrowserToolbar
+ */
+class BrowserSnackbarPresenter(
     private val view: View
 ) {
     fun present(
@@ -158,20 +162,20 @@ class FenixSnackbarPresenter(
         actionName: String? = null,
         isError: Boolean = false
     ) {
-        val snackbar = FenixSnackbar.make(view, length, isError).setText(text).let {
-            if (action != null && actionName != null) it.setAction(actionName, action) else it
-        }
+        val shouldUseBottomToolbar = view.context.settings().shouldUseBottomToolbar
+        val toolbarHeight = view.context.resources
+            .getDimensionPixelSize(R.dimen.browser_toolbar_height)
 
-        if (view.context.settings().shouldUseBottomToolbar) {
-            snackbar.view.setPadding(
+        FenixSnackbar.make(view, length, isError).apply {
+            if (action != null && actionName != null) setAction(actionName, action)
+            setText(text)
+            view.setPadding(
                 0,
                 0,
                 0,
-                view.context.resources.getDimensionPixelSize(R.dimen.browser_toolbar_height
-                )
+                if (shouldUseBottomToolbar) toolbarHeight else 0
             )
+            show()
         }
-
-        snackbar.show()
     }
 }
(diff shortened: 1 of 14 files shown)


gabrielluong:
I also encountered this while working on top sites. I thin we can also clean up the `getSnackbarAnchor` getter since we can just rely on this hardcoded bottom padding for bottom toolbar. https://github.com/mozilla-mobile/fenix/search?q=getSnackbarAnchor&unscoped_q=getSnackbarAnchor

sblatz:
@gabrielluong Good call, cleaning that up now.

app/src/main/java/org/mozilla/fenix/components/FenixSnackbar.kt
```diff
@@ -157,8 +162,20 @@ class FenixSnackbarPresenter(
         actionName: String? = null,
         isError: Boolean = false
     ) {
-        FenixSnackbar.make(view, length, isError).setText(text).let {
+        val snackbar = FenixSnackbar.make(view, length, isError).setText(text).let {
             if (action != null && actionName != null) it.setAction(actionName, action) else it
```
Baron-Severin: Nit: I find this usage of `let` a little confusing, I think it would work better without it.  We're creating a view and then configuring a bunch of properties on it, so `apply` seems like it could work well here.

app/src/main/java/org/mozilla/fenix/components/FenixSnackbar.kt
```diff
@@ -157,8 +162,20 @@ class FenixSnackbarPresenter(
...
-        }.show()
+        }
+
+        if (view.context.settings().shouldUseBottomToolbar) {
+            snackbar.view.setPadding(
+                0,
```
Baron-Severin: Nit: Is the default padding (used for top toolbar) just 0 for all sides?  Could we make that explicit?

app/src/main/java/org/mozilla/fenix/components/FenixSnackbar.kt
```diff
@@ -157,8 +162,20 @@ class FenixSnackbarPresenter(
...
+                0,
+                0,
+                view.context.resources.getDimensionPixelSize(R.dimen.browser_toolbar_height
+                )
+            )
+        }
+
+        snackbar.show()
```
Baron-Severin: If you agree with my above two comments, it might wind up looking something like this.

```kotlin
) {
    val shouldUseBottomToolbar = view.context.settings().shouldUseBottomToolbar
    val toolbarHeight = view.context.resources
        .getDimensionPixelSize(R.dimen.browser_toolbar_height)

    FenixSnackbar.make(view, length, isError).apply {
        setText(text)
        if (action != null && actionName != null) setAction(actionName, action)
        view.setPadding(
            0,
            0,
            0,
            if (shouldUseBottomToolbar) toolbarHeight else 0
        )
        show()
    }
}
```

app/src/main/java/org/mozilla/fenix/library/bookmarks/edit/EditBookmarkFragment.kt
```diff
@@ -185,17 +185,14 @@ class EditBookmarkFragment : Fragment(R.layout.fragment_edit_bookmark) {
...
-                                                R.string.bookmark_deletion_snackbar_message,
-                                                it.url?.toShortUrl(context.components.publicSuffixList) ?: it.title
-                                            )
-                                        )
-                                        .show()
-                                }
+
+                            bookmarkNode?.let {
+                                BrowserSnackbarPresenter(activity.getRootView()!!).present(
+                                    getString(
```
Baron-Severin: Nit: there are enough layers here that I think it would help clarity if we named `it` as `node`.

app/src/main/java/org/mozilla/fenix/components/FenixSnackbar.kt
sblatz: This feels a bit cleaner, thanks!

app/src/main/java/org/mozilla/fenix/library/bookmarks/edit/EditBookmarkFragment.kt
sblatz: Changed 😄

PR #7468, opened 2020-01-02
No issue: remove lint baseline file

mcomella:
More code for you to remove! 😁 

### Pull Request checklist

### After merge

1. click on Show All Checks,


jonalmeida:
With warningsAsErrors, would this cause the builds to fail now or just clog up the build log? There's an unused import so I'm not sure if that would happen or not. Maybe we should incrementally remove those baseline warnings instead if that's the case?

mcomella:
> With warningsAsErrors, would this cause the builds to fail now or just clog up the build log? 

This is a change for Android Lint, which does not have `warningsAsErrors` enabled (that's kotlin). It also runs as a separate task so it will not clog up the build log (but in any case, I believe it writes it to an HTML/XML report instead of printing straight to the build log anyway).

> There's an unused import so I'm not sure if that would happen or not.

This is for ktlint, which is unrelated.

> Maybe we should incrementally remove those baseline warnings instead if that's the case?

I think I've addressed your concerns in a good way but let me know if you disagree. :)

mcomella:
Addressed the ktlint error.

mcomella:
> # Configuration problem
> 
> bors.toml: not found

# 🙀

hakkikaancaliskan:
Isn't bors removed from Fenix?

jonalmeida:
Oh right, I forget. :)

PR #7470, opened 2020-01-02
Issue 7281 move set startup metrics to worker thread

MarcLeclair:
### Pull Request checklist

### After merge

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -36,6 +37,7 @@ import java.security.InvalidParameterException
 /**
  * A simple wrapper for SharedPreferences that makes reading preference a little bit easier.
  */
+@AnyThread
```
mcomella: nit: I don't think this clearly means, "Thread Safe" to most people (I've gotten questions about my use of `@UiThread` annotations and I think those are clearly). I think we should comment "thread safe" in the kdoc.

app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt
```diff
@@ -495,7 +496,7 @@ class GleanMetricsService(private val context: Context) : MetricsService {
         }
         // setStartupMetrics is not a fast function. It does not need to be done before we can consider
         // ourselves initialized. So, let's do it, well, later.
-        gleanSetStartupMetrics = MainScope().launch {
+        gleanSetStartupMetrics = MainScope().launch(IO) {
             setStartupMetrics()
```
mcomella: This doesn't remove the redundant computation in `setStartupMetrics`, wasting user's battery and clogging up our worker threads.

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
MarcLeclair: Could we possibly refer to the Android Doc? Where I agree we can put in the kdoc, I think this at least makes it clear to anyone with the source code too since you can command+click into the code 😄

mcomella:
After Will's `BrowsersCache` PR, I noticed this takes ~17.15ms on my P2, which is ~0.70% of startup. This doesn't seem like a huge win but if it's safe and free (which I'm not sure it is), we could take it.

mcomella:
I didn't do extensive research but `SearchEngineManager.defaultSearchEngine`, which is read by `setStartupMetrics`, does not appear to be thread safe. We'd have to validate this and potentially other methods. I'd rather focus on other larger low-hanging fruit in the short term.

@MarcLeclair What do you think? Do you agree or see something differently?

MarcLeclair:
@mcomella  talking to the Glean team (more specifically @Dexterp37 ), I was told that any metrics set up was okay to do asynchronously. I'm taking a look as we speak to verify myself 😄

MarcLeclair:
Also, looking at `SearchEngineManager`, I see that they use it asynchronously by initializing it with a `scope` object. I'll edit my comment as I dig in the code 👍

Dexterp37:
> 
> 
> @mcomella talking to the Glean team (more specifically @Dexterp37 ), I was told that any metrics set up was okay to do asynchronously. I'm taking a look as we speak to verify myself 😄

More precisely: the Glean metrics APIs do not care which thread they are called from. Collecting data before it's pushed to the Glean SDK might be different, though.

mcomella:
Next steps:
- @MarcLeclair will sanity check another mid-range device to see if it also has perf problems in `setStartupMetrics()`
- @MarcLeclair will put the associated issue back into triage. We'll focus on issues that affect both devices first and, if we get P2 over the finish line but the G5 is lagging behind, take this bug out to improve performance on the G5.

mcomella:
Rationale: ^ we want to make changes that improve all devices so we shouldn't focus right on the G5 but we should sanity check a similar device to see if it also has problems.

mcomella:
This isn't actively being worked on so I'm going to close the PR to keep the Fenix PRs list clean. That being said, we may still wish to fix the associated issue so this may get reopened.

Issue #7425, opened 2019-12-31
Redundant expensive calls to `Browsers.all` in `LeanplumMetricsService.start`


mcomella:
`LeanplumMetricsService.start` calls two of our methods that take up :
- `getInstalledMozillaProducts`
- `getMozillaBrowsersDefault`

Both of these methods call `Browser.all` redundantely for a total of ~40ms on my Pixel 2. Eyeballing it, these calls together appear to take 35-40% of `LeanplumMetricsService.start()` which is already a huge part of startup.

Can we wait to call these until after startup or at least only initialize `Browsers` once?

hawkinsw:
FYI: This will not be a startup issue after @MarcLeclair lands the patch to move `setStartupMetrics` off the main thread. 

That does not mean, however, that the underlying issue will be fixed.

mcomella:
> FYI: This will not be a startup issue after @MarcLeclair lands the patch to move `setStartupMetrics` off the main thread.

@hawkinsw `setStartupMetrics` does not seem related to Leanplum: I only see `GleanMetricsService.setStartupMetrics`. Am I missing something?

mcomella:
Oh, I see a similar pattern in GleanMetricsService, which must be what Will means: https://github.com/mozilla-mobile/fenix/blob/c3389dcdb58d0853e97ed4fb9bc9e13dd1936ee8/app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt#L507

hawkinsw:
> Oh, I see a similar pattern in GleanMetricsService, which must be what Will means:
> 
> https://github.com/mozilla-mobile/fenix/blob/c3389dcdb58d0853e97ed4fb9bc9e13dd1936ee8/app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt#L507

Sorry, these are really two different things. `GleanMetricsService.setStartupMetrics` does use these methods and @MarcLeclair 's patch will move those off the MT. 

I will go back and look into potential mitigations w.r.t this and Leanplum. Sorry for the confusion. I will assign this to myself.

hawkinsw:
See also https://github.com/mozilla-mobile/fenix/pull/7477

hawkinsw:
QA: To verify, I think that there are two options:

1. Check that metrics (Leanplum, Glean, etc) gathered about a user's device should contain an accurate list of all installed browsers. 
[...]

PR #7475, opened 2020-01-02
Issue #7425: Set Leanplum user attributes asynchronously

hawkinsw:
It is possible to delay setting the user attributes until after initializing Leanplum. Because getting the values needed to fill out the user attributes can be slow, this will save us time on startup.

There is a risk, however. 

When calling `Leanplum.start()` with user attributes, the Leanplum session (on the server) is always has associated user attributes. When calling `Leanplum.start()` without user attributes and setting the user attributes later (using `Leanplum.setUserAttributes()`), the session on the server starts without any associated attributes. When a session's user attributes are set in this two-step dance, there is a time where sessions stored on the server will not have any associated attributes. As a result, it is possible that Leanplum analytics on the server for a particular session may not have valid attributes if the user's device loses network connectivity after the call to `Leanplum.start()` and before the call to `Leanplum.setUserAttributes()`. 

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/metrics/LeanplumMetricsService.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/metrics/LeanplumMetricsService.kt
@@ -10,6 +10,9 @@ import com.leanplum.Leanplum
 import com.leanplum.LeanplumActivityHelper
 import com.leanplum.annotations.Parser
 import com.leanplum.internal.LeanplumInternal
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.MainScope
+import kotlinx.coroutines.launch
 import org.mozilla.fenix.BuildConfig
 import org.mozilla.fenix.ext.settings
 import java.util.UUID.randomUUID
@@ -70,17 +73,30 @@ class LeanplumMetricsService(private val application: Application) : MetricsServ
 
         LeanplumActivityHelper.enableLifecycleCallbacks(application)
 
-        val installedApps = MozillaProductDetector.getInstalledMozillaProducts(application)
+        Leanplum.start(application)
 
-        Leanplum.start(application, hashMapOf(
-            "default_browser" to MozillaProductDetector.getMozillaBrowserDefault(application).orEmpty(),
-            "fennec_installed" to installedApps.contains(MozillaProductDetector.MozillaProducts.FIREFOX.productName),
-            "focus_installed" to installedApps.contains(MozillaProductDetector.MozillaProducts.FOCUS.productName),
-            "klar_installed" to installedApps.contains(MozillaProductDetector.MozillaProducts.KLAR.productName),
-            "fxa_signed_in" to application.settings().fxaSignedIn,
-            "fxa_has_synced_items" to application.settings().fxaHasSyncedItems,
-            "search_widget_installed" to application.settings().searchWidgetInstalled
-        ))
+        MainScope().launch(Dispatchers.IO) {
+            val installedApps = MozillaProductDetector.getInstalledMozillaProducts(application)
+
+            Leanplum.setUserAttributes(
+                hashMapOf(
+                    "default_browser" to
+                            MozillaProductDetector.getMozillaBrowserDefault(application).orEmpty(),
+                    "fennec_installed" to
+                            installedApps.contains(MozillaProductDetector.MozillaProducts.FIREFOX.productName),
+                    "focus_installed" to
+                            installedApps.contains(MozillaProductDetector.MozillaProducts.FOCUS.productName),
+                    "klar_installed" to
+                            installedApps.contains(MozillaProductDetector.MozillaProducts.KLAR.productName),
+                    "fxa_signed_in" to
+                            application.settings().fxaSignedIn,
+                    "fxa_has_synced_items" to
+                            application.settings().fxaHasSyncedItems,
+                    "search_widget_installed" to
+                            application.settings().searchWidgetInstalled
+                )
+            )
+        }
     }
 
     override fun stop() {


app/src/main/java/org/mozilla/fenix/components/metrics/LeanplumMetricsService.kt
```diff
@@ -70,17 +73,30 @@ class LeanplumMetricsService(private val application: Application) : MetricsServ
...
-        ))
+        Leanplum.start(application)
+
+        MainScope().launch(Dispatchers.IO) {
+            val installedApps = MozillaProductDetector.getInstalledMozillaProducts(application)
+
+            Leanplum.setUserAttributes(
+                hashMapOf(
+                    "default_browser" to
+                            MozillaProductDetector.getMozillaBrowserDefault(application).orEmpty(),
```
mcomella: This PR doesn't remove the redundant computation in each of the `MozillaProductDetector` method calls, which are wasting the users battery & clogging up our background threads.

hawkinsw:
> Curious to ask since I can't find an answer after extensively googling... Looking at the Coroutine Docs, and seeing how the `MainScope()` [is used](https://github.com/Kotlin/kotlinx.coroutines/blob/master/docs/coroutine-context-and-dispatchers.md#coroutine-scope), would we have to explicitly cancel it in this case?

Very good point. We discussed this asynchronously on slack. Thanks for bringing it up.

mcomella:
I have a few thoughts.

Are we sure this is safe to call when not on the UI thread? I noticed that `Leanplum.start` is not documented as thread safe. Looking at the code, I see it assigning to a few variables that are not synchronized, nor `@Volatile`.

Are we seeing worthwhile gains? In a local profile before this patch, inside `LeanplumMetricsService.start`, I see:
* This is the first method to prime the `BrowsersCache` (20.75ms on Pixel 2)
* `TuplesKt.to` is visible but takes a minimal amount of time (1.23 ms on Pixel 2)

I don't think this is significant enough to put on a background thread but maybe I missed something. However, if we wanted to improve this, I think it might be safer to "prime" `BrowsersCache` on a background thread in anticipation of this code, due to the reasons above and below.

---

Thanks for the explanation of the possible side effects, hawkinsw. I noticed an additional one [in the docs](https://docs.leanplum.com/reference#user-attributes):

> // Passing attributes at session start allows us to target content based on the attributes.
> Leanplum.start(userAttributes: ["gender":"Female", "age": 29])
>
> // You can also pass them later on in the session, but you won't be able to
> // target variables or messages at these for that session.
> Leanplum.setUserAttributes(["gender":"Female", "age": 29])

Earlier on in the docs, they mention:

> Each session has its own user attributes, but they get copied from one session to the next.

And [elsewhere](https://docs.leanplum.com/reference#sessions) they mention that sessions can be paused/resumed and how new sessions are created:

> When the app is backgrounded, Leanplum will pause the session. When the app is re-opened, the session will resume. If the pause (background) lasts over 30 minutes, resuming the session will count as a new session. 

It sounds like our first session would not be targetable on these attributes and a new session will not create until 1) (maybe, based on when `Leanplum.start` is called) the process is restarted or 2) the app hasn't been used in 30 minutes.

I'm concerned this might not meet our product requirements. That being said, if we can resolve the other concerns, we can bring this to our Leanplum consumers and discuss the trade-offs.

---

@hawkinsw What are your thoughts?

hawkinsw:
> I have a few thoughts.
> 
> Are we sure this is safe to call when not on the UI thread? I noticed that `Leanplum.start` is not documented as thread safe. Looking at the code, I see it assigning to a few variables that are not synchronized, nor `@Volatile`.

I don't see the problem. We are not calling `.start()` off MT with this patch. I must not be understanding correctly what you are saying. 

Calling `.start()` off MT is *definitely* not safe. We have documented the issues with doing *that* in other issues. However, doing `.setUserAttributes()` off MT is okay, I think.

> 
> Are we seeing worthwhile gains? In a local profile before this patch, inside `LeanplumMetricsService.start`, I see:
> 
>     * This is the first method to prime the `BrowsersCache` (20.75ms on Pixel 2)
> 
>     * `TuplesKt.to` is visible but takes a minimal amount of time (1.23 ms on Pixel 2)
> 
> 
> I don't think this is significant enough to put on a background thread but maybe I missed something. However, if we wanted to improve this, I think it might be safer to "prime" `BrowsersCache` on a background thread in anticipation of this code, due to the reasons above and below.
> 
> Thanks for the explanation of the possible side effects, hawkinsw. I noticed an additional one [in the docs](https://docs.leanplum.com/reference#user-attributes):
> 
> > // Passing attributes at session start allows us to target content based on the attributes.
> > Leanplum.start(userAttributes: ["gender":"Female", "age": 29])
> > // You can also pass them later on in the session, but you won't be able to
> > // target variables or messages at these for that session.
> > Leanplum.setUserAttributes(["gender":"Female", "age": 29])
> 
> Earlier on in the docs, they mention:
> 
> > Each session has its own user attributes, but they get copied from one session to the next.
> 
> And [elsewhere](https://docs.leanplum.com/reference#sessions) they mention that sessions can be paused/resumed and how new sessions are created:
> 
> > When the app is backgrounded, Leanplum will pause the session. When the app is re-opened, the session will resume. If the pause (background) lasts over 30 minutes, resuming the session will count as a new session.
> 
> It sounds like our first session would not be targetable on these attributes and a new session will not create until 1) (maybe, based on when `Leanplum.start` is called) the process is restarted or 2) the app hasn't been used in 30 minutes.
> 
> I'm concerned this might not meet our product requirements. That being said, if we can resolve the other concerns, we can bring this to our Leanplum consumers and discuss the trade-offs.
> 
> @hawkinsw What are your thoughts?

This, more operational issue, seems like the real question. Great catch! Should we bring in someone from the team that actually uses Leanplum to decide whether or not this is necessary?

mcomella:
Next steps: 
- @mcomella will check with Leanplum consumers if this trade-off is acceptable (yes/no answer, they shouldn't need to think about it)
- If so, we should investigate if the change is worth making
- If it is, we can investigate the viability of making the change further

mcomella:
@hawkinsw Our Leanplum consumers mentioned we need _some_ user attributes for the first session. In particular:

> important ones: default browser, fennec installed, search widget installed

Given the increasing scope and my earlier Pixel 2 measurement where I didn't notice any significant wins, I'm inclined to say we defer this in favor of higher impact work but I'm happy to hear your thoughts.

hawkinsw:
Agree 100 percent. Thank you for confirming with the LP consumers.

mcomella:
I'm closing to close this and move the issue back to "Needs triage". We can reopen if we revisit the issue. Thanks for putting this together, hawkinsw.

mcomella:
This PR pointed at #7425, which was fixed by the `BrowsersCache` change, so I filed a new issue: https://github.com/mozilla-mobile/fenix/issues/7704

Issue #7281, opened 2019-12-18
Move setStartupMetrics() to worker thread to reduce start up time of Fenix 


MarcLeclair:
### Why/User Benefit/User Problem
Currently, every time the application starts up , we see a heavy load from `setStartupMetrics()` . I.e: on a 5 seconds start up time, the method can take up to 500ms (on a Moto G5+). This can mean up to 10% of load time only on setting up Metrics.

* Talking with the Glean team, it seems that we can initialize metrics on a separate threads. I'm not sure if there was a specific reason in Fenix for us to just initialize it on the UI thread.

### Impact
Longer start up, means less appealing to users. 

### Acceptance Criteria (how do I know when I’m done?)
The method doesn't run on initialization ( or is done in a separate thread)

boek:
I know @hawkinsw has spent some time looking at this. You might want to connect with him to share some context!

hawkinsw:
@mcomella Based on your comment from a different issue, I know that you had some hesitations with accessing Settings off MT. We will need to address your fear before this can go forward, I think.

hawkinsw:
Though caching the results of `Browsers.all()` (see PR: XX) will make the performance of `setStartupMetrics` better, I do not think that it is a bad idea to also do it asynchronously. @mcomella and @MarcLeclair, what do you think?

mcomella:
We spoke on Zoom: iirc, after hawkinsw's change, we noticed this affected the G5 pretty heavily but didn't affect the P2. We decided that, for now, we should focus on changes that only affect both reference devices. As such, I'm sending this back to triage.

mcomella:
See the WIP in https://github.com/mozilla-mobile/fenix/pull/7470.

mcomella:
Dupe https://github.com/mozilla-mobile/fenix/issues/7448

PR #7477, opened 2020-01-03
No Issue: Cache the results of getMozillaBrowserDefault and getInstalledMozillaProducts

hawkinsw:
Calling `getMozillaBrowserDefault` and `getInstalledMozillaProducts` can return cached results. Doing so will increase performance of Fenix on startup.

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/metrics/MozillaProductDetector.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/metrics/MozillaProductDetector.kt
@@ -27,8 +27,29 @@ object MozillaProductDetector {
         LOCKWISE("mozilla.lockbox")
     }
 
+    // The results of getInstalledMozillaProducts or getMozillaBrowserDefault can be cached. When
+    // either of those functions is called with the same context as the previous invocation, return
+    // the cached results from cachedMozillaProducts or cachedDefaultBrowserPackageName, respectively.
+    private var cachedContext: Context? = null
+    private var cachedMozillaProducts: MutableList<String>? = null
+    private var cachedDefaultBrowserPackageName: String? = null
+    private var cachedDefaultBrowserPackageNameSet: Boolean = false
+
+    /**
+     * Returns a list of the Mozilla products installed on the user's device.
+     */
+    @Synchronized
     fun getInstalledMozillaProducts(context: Context): List<String> {
-        val mozillaProducts = mutableListOf<String>()
+        // If there is a cached context and there is a cached list of products, return the cached
+        // list of products if _context_ and _cachedContext_ are equal.
+        if (cachedContext != null && cachedMozillaProducts != null && context == cachedContext) {
+            return cachedMozillaProducts!!
+        }
+
+        // Otherwise, set the cached context and build the list of installed products.
+        cachedContext = context
+
+        var mozillaProducts = mutableListOf<String>()
 
         for (product in MozillaProducts.values()) {
             if (packageIsInstalled(context, product.productName)) { mozillaProducts.add(product.productName) }
@@ -40,7 +61,8 @@ object MozillaProductDetector {
             }
         }
 
-        return mozillaProducts
+        cachedMozillaProducts = mozillaProducts
+        return cachedMozillaProducts!!
     }
 
     private fun packageIsInstalled(context: Context, packageName: String): Boolean {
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/fenix/components/metrics/MozillaProductDetector.kt
```diff
@@ -27,8 +27,29 @@ object MozillaProductDetector {
         LOCKWISE("mozilla.lockbox")
     }
 
+    // The results of getInstalledMozillaProducts or getMozillaBrowserDefault can be cached. When
+    // either of those functions is called with the same context as the previous invocation, return
+    // the cached results from cachedMozillaProducts or cachedDefaultBrowserPackageName, respectively.
+    private var cachedContext: Context? = null
+    private var cachedMozillaProducts: MutableList<String>? = null
+    private var cachedDefaultBrowserPackageName: String? = null
+    private var cachedDefaultBrowserPackageNameSet: Boolean = false
```
mcomella: How does this improve startup time? afaik, at most call sites we create multiple instances but instance variable caching would not improve this behavior.

Also, I think the callers can do instance caching more simply:
```kotlin
val installedMozillaProducts = Browsers.all().getInstalledMozillaProducts(context)
```

---

fwiw, before we invest too much time into caching, which is generally complex, I'd rather ensure:
- Browsers is defined efficiently: it seems to initialize more functionality than it needs to return the results for a single method call
- We only call Browsers when necessary & we re-use the calculations from Browsers if it's used multiple times in the same place
- When possible, we defer Browsers calls to other threads

I'm not sure if we've done this already.

mcomella: Oh, this is in `MozillaProductDetector`: let me look again.

hawkinsw:
Yes.

app/src/main/java/org/mozilla/fenix/components/metrics/MozillaProductDetector.kt
mcomella: To take a step back and look at the root cause, on startup we initialize Browsers, a slow operation, 6 times. 5 of these calls are wrapped in the MozillaProductDetector methods. They are in:
- LeanplumMetricsService 2x
- GleanMetricsService 3x
- HomeFragment.onStart calls `Settings.showPrivateModeContextualFeatureRecommender`

Furthermore, `MozillaProductDetector.getInstalledMozillaProducts` calls `packageIsInstalled`, which is slow but takes only ~15% of the time as `Browsers.all`, 3 times during startup.

To be thorough, outside of startup, `DefaultBrowserSettingsFragment` calls Browsers 3 times when shown & `IntentReceiverActivity` calls it once per processed Intent (I think that's most intents besides the homescreen intent).

---

I think the root cause is that the existing code doesn't treat `Browsers` (and its wrappers) as an expensive operation. This root cause has three PRs opened against it: this, https://github.com/mozilla-mobile/fenix/pull/7475, and https://github.com/mozilla-mobile/fenix/pull/7470. I think we should try to unify our solution.

Your patch partially addresses the root cause by adding caching in the wrapper but it can do better by also addressing the re-use of Browsers. Another issue that I see with the current implementation is that it caches the content indefinitely which may lead to incorrectness: if the user has backgrounded the app and returns, they may have uninstalled valid Mozilla products or changed their default browser. Even if the app remains open, app installations can change in the background, e.g. if the Play Store just finished installing an app or via adb.

Furthermore, after we add caching, we should question whether it's still valuable to initialize Leanplum & Glean metrics on a worker thread.

To try to sum up my thinking, **I think we should get these values once during startup and keep things on the main thread to keep it simple.** If LeanplumMetrics and GleanMetrics are still slow, we can file a follow-up issue to address it because we'll still have big wins by reducing `Browsers` calls from 6 to 1.

MarcLeclair:
I think I would agree with unifying a solution in this case since it seems as if the issue that @hawkinsw  is trying to address here and the one I was in #7281 have a deeper root cause than what we originally saw ( at least for my ticket).  So, to resume my thought, I think that addressing the root problem is ideal instead of just throwing it on a worker thread and forgetting about it.

mcomella:
Closed in favor of https://github.com/mozilla-mobile/fenix/pull/7490.

PR #7490, opened 2020-01-04
Issue #7425 (et al): Cache the list of installed browsers

hawkinsw:
Cache the list of installed browsers. Calling `Browsers.all`
the application directly redundantly recalculates the list.
Accessing the list of installed browsers through this cache
will reduce that overhead.

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
+++ b/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
@@ -60,6 +60,7 @@ import org.mozilla.fenix.settings.SettingsFragmentDirections
 import org.mozilla.fenix.settings.TrackingProtectionFragmentDirections
 import org.mozilla.fenix.theme.DefaultThemeManager
 import org.mozilla.fenix.theme.ThemeManager
+import org.mozilla.fenix.utils.BrowsersCache
 
 @SuppressWarnings("TooManyFunctions", "LargeClass")
 open class HomeActivity : LocaleAwareAppCompatActivity() {
@@ -139,6 +140,16 @@ open class HomeActivity : LocaleAwareAppCompatActivity() {
         hotStartMonitor.onPostResumeFinalMethodCall()
     }
 
+    final override fun onPause() {
+        super.onPause()
+
+        // Every time the application goes into the background, it is possible that the user
+        // is about to change the browsers installed on their system. Therefore, we reset the cache of
+        // all the installed browsers.
+        //
+        // NB: There are ways for the user to install new products without leaving the browser.
+        BrowsersCache.resetAll()
+    }
     /**
      * Handles intents received when the activity is open.
      */
--- a/app/src/main/java/org/mozilla/fenix/utils/BrowsersCache.kt
+++ b/app/src/main/java/org/mozilla/fenix/utils/BrowsersCache.kt
@@ -1,0 +1,46 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.utils
+
+import android.content.Context
+import androidx.annotation.VisibleForTesting
+import mozilla.components.support.utils.Browsers
+
+/**
+ * Caches the list of browsers installed on a user's device.
+ *
+ * BrowsersCache caches the list of installed browsers is gathered lazily when it is first accessed
+ * after initial creation or invalidation. For that reason, a context is required every time
+ * the cache is accessed.
+ *
+ * Users are responsible for invalidating the cache at the appropriate time. It is left up to the
+ * user to determine appropriate policies for maintaining the validity of the cache. If, when the
+ * cache is accessed, it is filled, the contents will be returned. As mentioned above, the cache
+ * will be lazily refilled after invalidation. In other words, invalidation is O(1).
+ *
+ * This cache is threadsafe.
+ */
+object BrowsersCache {
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal var cachedBrowsers: Browsers? = null
+
+    @Synchronized
+    fun all(context: Context): Browsers {
+        run {
+            val cachedBrowsers = cachedBrowsers
+            if (cachedBrowsers != null) {
+                return cachedBrowsers
+            }
+        }
+        return Browsers.all(context).also {
+            this.cachedBrowsers = it
+        }
+    }
+
+    @Synchronized
+    fun resetAll() {
+        cachedBrowsers = null
+    }
+}
--- a/app/src/test/java/org/mozilla/fenix/utils/BrowsersCacheTest.kt
+++ b/app/src/test/java/org/mozilla/fenix/utils/BrowsersCacheTest.kt
@@ -1,0 +1,155 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+@file:Suppress("DEPRECATION")
+
+package org.mozilla.fenix.utils
+
+import android.content.Intent
+import android.content.pm.ActivityInfo
+import android.content.pm.PackageInfo
+import android.content.pm.ResolveInfo
+import android.net.Uri
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import mozilla.components.support.test.robolectric.testContext
+import mozilla.components.support.utils.Browsers
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertNotNull
+import org.junit.Assert.assertNull
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.mozilla.fenix.TestApplication
+import org.robolectric.Shadows.shadowOf
+import org.robolectric.annotation.Config
+
+@RunWith(AndroidJUnit4::class)
+@Config(application = TestApplication::class)
+class BrowsersCacheTest {
+
+    // NB: There is always one more browser than pretendBrowsersAreInstalled installs because
+    // the application we are testing is recognized as a browser itself!
+
+    @Test
+    fun `cached list of browsers match before-after installation when cache is not invalidated`() {
+        BrowsersCache.resetAll()
+        pretendBrowsersAreInstalled(
+            browsers = listOf(
+                Browsers.KnownBrowser.FIREFOX_NIGHTLY.packageName,
+                Browsers.KnownBrowser.REFERENCE_BROWSER.packageName
+            )
+        )
+
+        val initialBrowserList = BrowsersCache.all(testContext)
+        assertEquals(3, initialBrowserList.installedBrowsers.size)
+
+        pretendBrowsersAreInstalled(
+            browsers = listOf(
+                Browsers.KnownBrowser.FIREFOX_NIGHTLY.packageName,
+                Browsers.KnownBrowser.FIREFOX.packageName,
+                Browsers.KnownBrowser.CHROME.packageName,
+                Browsers.KnownBrowser.SAMSUNG_INTERNET.packageName,
+                Browsers.KnownBrowser.DUCKDUCKGO.packageName,
+                Browsers.KnownBrowser.REFERENCE_BROWSER.packageName
+            )
+        )
+        val updatedBrowserList = BrowsersCache.all(testContext)
+        assertEquals(3, updatedBrowserList.installedBrowsers.size)
+    }
+
+    @Test
+    fun `cached list of browsers change before-after installation when cache is invalidated`() {
+        BrowsersCache.resetAll()
+        pretendBrowsersAreInstalled(
+            browsers = listOf(
+                Browsers.KnownBrowser.FIREFOX_NIGHTLY.packageName,
+                Browsers.KnownBrowser.REFERENCE_BROWSER.packageName
+            )
+        )
+
+        val initialBrowserList = BrowsersCache.all(testContext)
+        assertEquals(3, initialBrowserList.installedBrowsers.size)
+
+        pretendBrowsersAreInstalled(
+            browsers = listOf(
+                Browsers.KnownBrowser.FIREFOX_NIGHTLY.packageName,
+                Browsers.KnownBrowser.FIREFOX.packageName,
+                Browsers.KnownBrowser.CHROME.packageName,
+                Browsers.KnownBrowser.SAMSUNG_INTERNET.packageName,
+                Browsers.KnownBrowser.DUCKDUCKGO.packageName,
+                Browsers.KnownBrowser.REFERENCE_BROWSER.packageName
+            )
+        )
+
+        BrowsersCache.resetAll()
+
+        val updatedBrowserList = BrowsersCache.all(testContext)
+        assertEquals(7, updatedBrowserList.installedBrowsers.size)
+    }
+
+    @Test
+    fun `resetting the cache should empty it`() {
+        BrowsersCache.resetAll()
+
+        BrowsersCache.all(testContext)
+
+        assertNotNull(BrowsersCache.cachedBrowsers)
+
+        BrowsersCache.resetAll()
+
+        assertNull(BrowsersCache.cachedBrowsers)
+    }
+
+    // pretendBrowsersAreInstalled was taken, verbatim, from a-c.
+    // See support/utils/src/test/java/mozilla/components/support/utils/BrowsersTest.kt
+    private fun pretendBrowsersAreInstalled(
+        browsers: List<String> = listOf(),
+        defaultBrowser: String? = null,
+        url: String = "http://www.mozilla.org",
+        browsersExported: Boolean = true,
+        defaultBrowserExported: Boolean = true
+    ) {
+        val packageManager = testContext.packageManager
+        val shadow = shadowOf(packageManager)
+
+        browsers.forEach { packageName ->
+            val intent = Intent(Intent.ACTION_VIEW)
+            intent.`package` = packageName
+            intent.data = Uri.parse(url)
+
+            val packageInfo = PackageInfo().apply {
+                this.packageName = packageName
+            }
+
+            shadow.installPackage(packageInfo)
+
+            val activityInfo = ActivityInfo().apply {
+                exported = browsersExported
+                this.packageName = packageName
+            }
+
+            val resolveInfo = ResolveInfo().apply {
+                resolvePackageName = packageName
+                this.activityInfo = activityInfo
+            }
+
+            shadow.addResolveInfoForIntent(intent, resolveInfo)
+        }
+
+        if (defaultBrowser != null) {
+            val intent = Intent(Intent.ACTION_VIEW)
+            intent.data = Uri.parse(url)
+
+            val activityInfo = ActivityInfo().apply {
+                exported = defaultBrowserExported
+                packageName = defaultBrowser
+            }
+
+            val resolveInfo = ResolveInfo().apply {
+                resolvePackageName = defaultBrowser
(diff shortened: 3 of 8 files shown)


app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -139,6 +140,14 @@ open class HomeActivity : LocaleAwareAppCompatActivity() {
         hotStartMonitor.onPostResumeFinalMethodCall()
     }
 
+    final override fun onPause() {
+        super.onPause()
+
+        // Every time the application goes into the background, it is possible that the user
+        // is about to change the browsers installed on their system. Therefore, we reset the cache of
+        // all the installed browsers.
+        BrowsersCache.resetAll()
```
NotWoods: It's also possible for the installed browsers to change while the app is still running. You can install an app to your phone from the Google Play website, and it will get installed in the background.

mcarare: @hawkinsw , @NotWoods Maybe we could use a `BroadcastReceiver` for `ACTION_PACKAGE_ADDED` and `ACTION_PACKAGE_REMOVED`   ?

hawkinsw:
We absolutely *could* use such a tactic but I was explicitly trying *not*
to do that. That's lots of overhead for a case that seems exceedingly rare.
I would really like someone on product to weigh in on whether or not this
is a case that we need to consider?

NotWoods:
I'm happy with just changing the comment to clarify that it's a good enough signal rather than a definite signal.

hawkinsw:
> I'm happy with just changing the comment to clarify that it's a good enough signal rather than a definite signal.

Sgtm. Will do! Thanks @NotWoods

app/src/test/java/org/mozilla/fenix/utils/BrowsersCacheTest.kt
```diff
@@ -0,0 +1,146 @@
...
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertNotNull
+import org.junit.Assert.assertNull
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.mozilla.fenix.TestApplication
+import org.robolectric.Shadows.shadowOf
+import org.robolectric.annotation.Config
+
+@RunWith(AndroidJUnit4::class)
```
mcomella: nit: you can remove this – it's the default so it's redundant.

Then again, you're using robolectric in this class: does this need to be a robolectric annotation?

app/src/test/java/org/mozilla/fenix/utils/BrowsersCacheTest.kt
```diff
@@ -0,0 +1,146 @@
...
+    fun testInstallNewBrowsersWithoutReset() {
+        pretendBrowsersAreInstalled(
+            browsers = listOf(
+                Browsers.KnownBrowser.FIREFOX_NIGHTLY.packageName,
+                Browsers.KnownBrowser.REFERENCE_BROWSER.packageName
+            )
+        )
+
+        var browsers = BrowsersCache.all(testContext)
+        assertEquals(browsers.installedBrowsers.size, 3)
```
mcomella: nit: argument order should be reversed throughout the file. It's `assertEquals(expected, actual)`

app/src/test/java/org/mozilla/fenix/utils/BrowsersCacheTest.kt
```diff
@@ -0,0 +1,146 @@
...
+        )
+
+        BrowsersCache.resetAll()
+
+        browsers = BrowsersCache.all(testContext)
+        assertEquals(browsers.installedBrowsers.size, 7)
+    }
+
+    @Test
+    fun testResetAll() {
```
mcomella: nit: not strictly necessary but I like to name my tests with the [GIVEN WHEN THEN pattern](https://martinfowler.com/bliki/GivenWhenThen.html): `GIVEN <some test set up> WHEN <event happens> THEN <expected result>`.

It's immediately clear what the test should be testing, which is useful when quickly identifying which tests already exist or when finding the problem when the test fails. It also encourages tests to focus on one bit of functionality. For example, this one might be, `GIVEN the internal cache has a value WHEN resetAll is called THEN the internal cache should be null`.

app/src/test/java/org/mozilla/fenix/utils/BrowsersCacheTest.kt
```diff
@@ -0,0 +1,146 @@
...
+    @Test
+    fun testInstallNewBrowsersWithoutReset() {
+        pretendBrowsersAreInstalled(
+            browsers = listOf(
+                Browsers.KnownBrowser.FIREFOX_NIGHTLY.packageName,
+                Browsers.KnownBrowser.REFERENCE_BROWSER.packageName
+            )
+        )
+
+        var browsers = BrowsersCache.all(testContext)
```
mcomella: super-nit: it's safer to use value with two separate names because if the code refactored poorly, the reassignment may not happen as expected.

app/src/test/java/org/mozilla/fenix/utils/BrowsersCacheTest.kt
```diff
@@ -0,0 +1,146 @@
...
+
+        BrowsersCache.resetAll()
+
+        assertNull(BrowsersCache.cachedBrowsers)
+        assertNull(BrowsersCache.cachedContext)
+    }
+
+    private fun pretendBrowsersAreInstalled(
+        browsers: List<String> = listOf(),
+        defaultBrowser: String? = null,
```
mcomella: nit: afaict, none of these arguments are used so we should remove them to simplify the code.

app/src/test/java/org/mozilla/fenix/utils/BrowsersCacheTest.kt
```diff
@@ -0,0 +1,146 @@
...
+        url: String = "http://www.mozilla.org",
+        browsersExported: Boolean = true,
+        defaultBrowserExported: Boolean = true
+    ) {
+        val packageManager = testContext.packageManager
+        val shadow = shadowOf(packageManager)
+
+        browsers.forEach { packageName ->
+            val intent = Intent(Intent.ACTION_VIEW)
```
mcomella: nit: idiomatic kotlin is:
```kotlin
val intent = Intent(Intent.ACTION_VIEW).apply {
    package = packageName
    data = Uri.parse(url)
}
```

Here and below.

app/src/main/java/org/mozilla/fenix/utils/BrowsersCache.kt
```diff
@@ -0,0 +1,26 @@
+package org.mozilla.fenix.utils
```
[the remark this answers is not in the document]
hawkinsw: Helpful! Thank you!

app/src/main/java/org/mozilla/fenix/utils/BrowsersCache.kt
```diff
@@ -0,0 +1,26 @@
...
+
+import android.content.Context
+import mozilla.components.support.utils.Browsers
+
+object BrowsersCache {
```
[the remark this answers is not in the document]
hawkinsw: Helpful. Thank you.

app/src/main/java/org/mozilla/fenix/utils/BrowsersCache.kt
```diff
@@ -0,0 +1,26 @@
...
+    internal var cachedBrowsers: Browsers? = null
+    internal var cachedContext: Context? = null
+
+    @Synchronized
+    fun all(context: Context): Browsers {
+        if (cachedContext != null && cachedBrowsers != null && cachedContext!! == context) {
```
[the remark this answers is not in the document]
hawkinsw: Belt and suspenders. Should there come a time when more than one context is used (not that I can imagine one), we want to make sure that we don't return the wrong result.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -139,6 +140,16 @@ open class HomeActivity : LocaleAwareAppCompatActivity() {
...
 
+    final override fun onPause() {
+        super.onPause()
+
+        // Every time the application goes into the background, it is possible that the user
+        // is about to change the browsers installed on their system. Therefore, we reset the cache of
+        // all the installed browsers.
+        //
+        // NB: There are ways for the user to install new products without leaving the browser.
+        BrowsersCache.resetAll()
```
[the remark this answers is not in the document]
hawkinsw: I see your point, but I disagree. Adding it to the HomeActivity makes it explicit that we are going to be invalidating the cache at this point. I believe that hiding that behavior as a side-effect of the implementation could be dangerous down the road. There may also be other places where invalidation of this cache would happen and don't want to imply that invalidation is taken care of entirely internally.

app/src/test/java/org/mozilla/fenix/utils/BrowsersCacheTest.kt
hawkinsw: I honestly don't know. This is "cargo cult"ed from a different test.

app/src/test/java/org/mozilla/fenix/utils/BrowsersCacheTest.kt
hawkinsw: Very helpful! I completely meant to do that. Thank you!

app/src/test/java/org/mozilla/fenix/utils/BrowsersCacheTest.kt
hawkinsw: I don't understand, sorry.

app/src/test/java/org/mozilla/fenix/utils/BrowsersCacheTest.kt
hawkinsw: That is true, but they are useful to have *if* we do need them. I took this from a different test and they are important there. I think that we should leave them because there may come a time where it is necessary to create contexts using them for the purposes of testing this cache.

app/src/test/java/org/mozilla/fenix/utils/BrowsersCacheTest.kt
hawkinsw: Again, this was not my code but I am happy to make that change.

app/src/main/java/org/mozilla/fenix/utils/BrowsersCache.kt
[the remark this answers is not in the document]
hawkinsw: Good point. I didn't realize that the compiler was smart enough to make those guarantees at compile time. Definitely learned something here! Thank you.

app/src/test/java/org/mozilla/fenix/utils/BrowsersCacheTest.kt
hawkinsw: I honestly don't know. I "cargo cult"ed.

app/src/test/java/org/mozilla/fenix/utils/BrowsersCacheTest.kt
mcomella: `browsers` is being re-assigned to later in the function to be used for a different, but similar, purpose: I think it's safer to use two `val`s (with different names) instead.

app/src/test/java/org/mozilla/fenix/utils/BrowsersCacheTest.kt
mcomella: Can we move this to a shared helper function then? I don't see any precident for this in the Fenix codebase but I usually add shared test code files to a `org.mozilla.fenix.helpers` package.

app/src/test/java/org/mozilla/fenix/utils/BrowsersCacheTest.kt
mcomella: Since it's not your code, no need to clean it up but see my thoughts above.

app/src/test/java/org/mozilla/fenix/utils/BrowsersCacheTest.kt
hawkinsw: Got it! Changing now!

app/src/test/java/org/mozilla/fenix/utils/BrowsersCacheTest.kt
hawkinsw: switching to descriptive function names!

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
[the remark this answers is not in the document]
mcomella: I don't feel too strongly this needs to change but I see this wasn't resolved so I'll make a last comment:

> There may also be other places where invalidation of this cache would happen and don't want to imply that invalidation is taken care of entirely internally.

This is a fair criticism. I wonder if we made it `resetAll` private if it'd make more sense to use an observer. I think one benefit of this is that we can explain in the class kdoc when the cache is valid, so users don't need to read the code to understand the invalidation strategy to know if it's able to solve their problem or not.

> I believe that hiding that behavior as a side-effect

But I don't see it as hiding – more like delegating lifecycle events so `HomeActivity` is more focused – so perhaps that's a difference in perspective.

app/src/main/java/org/mozilla/fenix/utils/BrowsersCache.kt
```diff
@@ -0,0 +1,54 @@
...
+    internal var cachedBrowsers: Browsers? = null
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal var cachedContext: Context? = null
+
+    @Synchronized
+    fun all(context: Context): Browsers {
+        run {
+            val cachedContext = cachedContext
+            val cachedBrowsers = cachedBrowsers
+            if (cachedContext != null && cachedBrowsers != null && cachedContext == context) {
```
mcomella: I see two issues with this:
1) We're storing an Activity cache statically so this actually creates a memory leak. If we pass in an Application Context instead, the check isn't useful
2) I think this might unnecessarily reset the cache if:
* `IntentReceiverActivity` fetches Browsers
* We fetch Browsers in `MainActivity`

iirc, the problem we were trying to solve was:

> Should there come a time when more than one context is used (not that I can imagine one), we want to make sure that we don't return the wrong result.

Can you describe from a user perspective what this might look like? I'm struggling to come up with one: it might dictate a better solution.

That being said, it sounds like we're trying to solve a problem we don't have yet so I wonder if there's a better way to prevent us from having bugs when that problem comes. For example, we could assume the relationship between `MainActivity` and `IntentReceiverActivity` won't change, for which this class is written, and do something like:
```kotlin
    fun all(context: Context): Browsers {
        require(!(context is MainActivity::class) && !(context is IntentReceiverActivity::class)) {
            "This class is written with assumptions about the lifecycles of MainActivity and IntentReceiverActivity and how that affects cache invalidation. This class needs to be adapted to this new Context."
        }
       // ...
```

app/src/main/java/org/mozilla/fenix/utils/BrowsersCache.kt
```diff
@@ -0,0 +1,54 @@
...
+ */
+object BrowsersCache {
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal var cachedBrowsers: Browsers? = null
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal var cachedContext: Context? = null
+
+    @Synchronized
+    fun all(context: Context): Browsers {
+        run {
```
mcomella: nit: why is this in a run block? It seems unnecessary and adds a potential source of confusion since `run` returns the last expression so `return cachedBrowsers` is actually escaping that scoping to the parent function, which may be unclear.

app/src/main/java/org/mozilla/fenix/utils/BrowsersCache.kt
mcomella: > We're storing an Activity cache statically so this actually creates a memory leak. If we pass in an Application Context instead, the check isn't useful

I guess we actually empty the cache in `onPause` so the current implementation is fine (assuming `IntentReceiverActivity` has no shenanigans) but it's fragile for future scenarios (e.g. `resetAll` is moved or a new Context is introduced where `resetAll` is used differently).

app/src/test/java/org/mozilla/fenix/utils/BrowsersCacheTest.kt
```diff
@@ -0,0 +1,153 @@
...
+    @Test
+    fun `cached list of browsers change before-after installation when cache is invalidated`() {
+        pretendBrowsersAreInstalled(
+            browsers = listOf(
+                Browsers.KnownBrowser.FIREFOX_NIGHTLY.packageName,
+                Browsers.KnownBrowser.REFERENCE_BROWSER.packageName
+            )
+        )
+
+        var initialBrowserList = BrowsersCache.all(testContext)
```
mcomella: nit: this can be val

app/src/main/java/org/mozilla/fenix/utils/BrowsersCache.kt
```diff
@@ -0,0 +1,54 @@
...
+            val cachedContext = cachedContext
+            val cachedBrowsers = cachedBrowsers
+            if (cachedContext != null && cachedBrowsers != null && cachedContext == context) {
+                return cachedBrowsers
+            }
+        }
+        this.cachedBrowsers = Browsers.all(context)
+        this.cachedContext = context
+
+        return this.cachedBrowsers!!
```
mcomella: nit: we could do:
```kotlin
this.cachedContext = context
return Browsers.all(context).also {
    this.cachedBrowsers = it
}
```

To avoid the `!!`

app/src/main/java/org/mozilla/fenix/utils/BrowsersCache.kt
hawkinsw: Not unclear. It's exactly what I want to happen. I am basically using it as a scoping operator.

mcomella: I don't think it's idiomatic in Kotlin – which can make it non-obvious to devs that it's a scoping operator – but wfm.

I wonder if a set of braces (also non-idiomatic but maybe more obvious) can actually be used as a scoping operator too... Or is that an anonymous function? lol

mcomella: The IDE yells at my for braces: it's an anonymous function.

app/src/main/java/org/mozilla/fenix/utils/BrowsersCache.kt
```diff
@@ -0,0 +1,46 @@
...
+object BrowsersCache {
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal var cachedBrowsers: Browsers? = null
+
+    @Synchronized
+    fun all(context: Context): Browsers {
+        run {
+            val cachedBrowsers = cachedBrowsers
+            if (cachedBrowsers != null) {
+                return cachedBrowsers
```
mcomella: nit: you're gonna lol but a more idiomatic kotlin way way to do this might be:
```kotlin
cachedBrowsers?.let { return it }
```

Since let is an assignment-shorthand and (idiomatically) the anonymous function will only execute if it's non-null.

mcomella: Though maybe it's weird to return from a let statement. 🤔

mcomella:
I'm going to merge so we're more likely to get this in tomorrow's build.

mcomella:
@hawkinsw Can you move the corresponding issue to "Done", add a QA needed label, add a comment to QA for how they can test this change (probably to verify this works as expected and not to verify the perf improvement), and add this to the current milestone (3.1)?

I need to email about Fenix's QA process. x_x It's annoying manually too. :(

Issue #6832, opened 2019-11-27
IllegalStateException: Fragment HomeFragment not attached to a context.


pocmo:
https://sentry.prod.mozaws.net/operations/fenix-nightly/issues/6162466/
https://sentry.prod.mozaws.net/operations/fenix-nightly/issues/6162466/
Socorro [link](https://crash-stats.mozilla.org/signature/?product=Fenix&signature=java.lang.IllegalStateException%3A%20at%20androidx.fragment.app.Fragment.requireContext%28Fragment.java%29&date=%3E%3D2019-07-21T07%3A06%3A00.000Z&date=%3C2020-04-22T07%3A06%3A00.000Z#graphs)

```
IllegalStateException: Fragment HomeFragment{1a41004} (84c6a2c6-6203-448c-b564-fee8f55208fc)} not attached to a context.
    at androidx.fragment.app.Fragment.requireContext(Fragment.java:2)
    at androidx.fragment.app.Fragment.getResources(Fragment.java:1)
    at org.mozilla.fenix.home.HomeFragment$onViewCreated$2.invokeSuspend(HomeFragment.kt:9)
    at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:2)
    at kotlinx.coroutines.DispatchedTask.run(Dispatched.kt:15)
...
(2 additional frame(s) were not displayed)

Fragment HomeFragment{1a41004} (84c6a2c6-6203-448c-b564-fee8f55208fc)} not attached to a context.
```

pocmo:
Because of the UUID in the stack trace this is split over multiple crash reports in Sentry.

Crash stats seems to be a bit smarter:
https://crash-stats.mozilla.com/signature/?product=Fenix&version=72.0a1&signature=java.lang.IllegalStateException%3A%20at%20androidx.fragment.app.Fragment.requireContext%28Fragment.java%29&date=%3E%3D2019-11-20T10%3A20%3A00.000Z&date=%3C2019-11-27T10%3A20%3A00.000Z&_columns=date&_columns=product&_columns=version&_columns=build_id&_columns=platform&_columns=reason&_columns=address&_columns=install_time&_columns=startup_crash&_sort=-date&page=1#reports

According to crash stats the crash volume is increasing and it's now our top 6 crash.

vesta0:
Need STR for this.

Baron-Severin:
[...]

PR #7509, opened 2020-01-06
For #6832: fix unattached context crash in HomeFragment

Baron-Severin:
### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
@@ -230,9 +230,12 @@ class HomeFragment : Fragment() {
         setupHomeMenu()
 
         viewLifecycleOwner.lifecycleScope.launch(IO) {
+            // This should be unnecessary, but we've seen crashes around it. See #6832
+            val context = context ?: return@launch
+
             val iconSize = resources.getDimensionPixelSize(R.dimen.preference_icon_drawable_size)
 
-            val searchEngine = requireComponents.search.provider.getDefaultEngine(requireContext())
+            val searchEngine = requireComponents.search.provider.getDefaultEngine(context)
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -230,9 +230,12 @@ class HomeFragment : Fragment() {
         setupHomeMenu()
 
         viewLifecycleOwner.lifecycleScope.launch(IO) {
+            // This should be unnecessary, but we've seen crashes around it. See #6832
+            val context = context ?: return@launch
```
Baron-Severin: If I understand correctly, `viewLifecycleOwner.lifecycleScope` should cancel after `onDestroy`, which should be guaranteed to occur prior to `onDetatch`, so it should be impossible for context to ever be null here.  That said, the stack traces from the Sentry events seem pretty clear.  Am I missing something here?  I'd love to get your opinion @pocmo @csadilek @colintheshots

pocmo: We have been seeing issues with `lifecycleScope` and `Activity.recreate()` (which Fenix does for theme and private mode changes etc.). In that situation the scope from the activity that got destroyed never gets cancelled. We found this after forever debugging such a "this cannot happen" situation too. I never got around trying to isolate this in an empty app to file an androidx issue. :(
So, we got rid of `lifecycleScope` in that situation and started using a `view bound` scope, e.g. see here:
https://github.com/mozilla-mobile/android-components/blob/master/components/lib/state/src/main/java/mozilla/components/lib/state/ext/Fragment.kt#L32-L56

Baron-Severin: 😱

PR #7517, opened 2020-01-07
No Issue: Add a coroutine scope to MetricController

hawkinsw:
The scope is managed by the lifetime of the controller and
passed to each `ServiceController`. `ServiceController`s can use
that scope to launch asynchronous tasks with the reassurance
that they will be GC'd and canceled appropriately.

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/metrics/Metrics.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/metrics/Metrics.kt
@@ -5,6 +5,9 @@
 package org.mozilla.fenix.components.metrics
 
 import android.content.Context
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.cancel
 import mozilla.components.browser.errorpages.ErrorType
 import mozilla.components.browser.search.SearchEngine
 import mozilla.components.browser.toolbar.facts.ToolbarFacts
@@ -389,7 +392,7 @@ private fun Fact.toEvent(): Event? = when (Pair(component, item)) {
 }
 
 interface MetricsService {
-    fun start()
+    fun start(scope: CoroutineScope)
     fun stop()
     fun track(event: Event)
     fun shouldTrack(event: Event): Boolean
@@ -400,6 +403,8 @@ interface MetricController {
     fun stop()
     fun track(event: Event)
 
+    var managedScope: CoroutineScope?
+
     companion object {
         fun create(services: List<MetricsService>, isTelemetryEnabled: () -> Boolean): MetricController {
             return if (BuildConfig.TELEMETRY) return ReleaseMetricController(services, isTelemetryEnabled)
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/mozilla/fenix/components/metrics/Metrics.kt
```diff
@@ -441,14 +450,21 @@ private class ReleaseMetricController(
...
+        managedScope = CoroutineScope(Dispatchers.IO)
+        services.forEach { it.start(managedScope!!) }
         initialized = true
     }
 
     override fun stop() {
         if (!initialized) { return }
 
         services.forEach { it.stop() }
+        run {
```
mcomella: nit: can this just be `managedScope?.cancel()`? fwiw, if I understand the intent, I haven't seen `run` used to group code before: usually whitespace is sufficient. Then again, I almost never see `run` used :P

app/src/main/java/org/mozilla/fenix/components/metrics/Metrics.kt
```diff
@@ -441,14 +450,21 @@ private class ReleaseMetricController(
     override fun start() {
         if (!isTelemetryEnabled.invoke() || initialized) { return }
 
-        services.forEach { it.start() }
+        managedScope = CoroutineScope(Dispatchers.IO)
```
mcomella: nit: as per the last PR, I'd prefer to store this in a temporary reference to avoid `!!`

hawkinsw:
> This seems like a good code clean-up change: don't let async coroutines keep going when `start` is called. However, if possible, I'd like to use the `scope` reference before we land this patch (e.g. so it doesn't go unused).

I completely agree with that. I was just trying to keep the two issues separate since they are really different "issues". What do you think?

MarcLeclair:
> This seems like a good code clean-up change: don't let async coroutines keep going when `start` is called. However, if possible, I'd like to use the `scope` reference before we land this patch (e.g. so it doesn't go unused).

@mcomella , I would use it too in  #7281, but as @hawkinsw  said, I think it's just a matter of keeping code separate so we don't get conflicts and to be able to track our changes 😄

sblatz:
Can this be merged, @MarcLeclair? 😄

hawkinsw:
I'd like to wait for @mcomella , if that's okay with you! Thanks everyone!

sblatz:
@mcomella Please remove the do not land label when you're ready to merge 😄

mcomella:
I have a few concerns regarding the viability of:
* async `LeanplumMetricsService.start` https://github.com/mozilla-mobile/fenix/pull/7475#issuecomment-572812615
* async `GleanMetricsService.setStartupMetrics` https://github.com/mozilla-mobile/fenix/pull/7470#issuecomment-572814302

Since this PR is made in preparation for those PRs, I want to unblock those before we move forward on this one: let's discuss in the individual issues.

Sorry this took so long to unblock.

mcomella:
Like https://github.com/mozilla-mobile/fenix/pull/7475#issuecomment-574479318, we deprioritized the associated issues so I'm closing the open PRs to clean up the Fenix PR list. If we want to revisit these issues, we can reopen the PRs.

PR #7541, opened 2020-01-08
Use Android Components 24.0.1 with Fenix 3.0.2

pocmo:
Updating the release branch for Fenix 3.0.2 to use Android Components 24.0.1.


mcarare:
This change https://github.com/mcarare/fenix/commit/c96af458f48e3d431aa96610c4ba7afffd59dc5b should be included in this release to fix the lint issue.

pocmo:
@mcarare Thank you. Added this patch on top.

pocmo:
Looks like some UI test failure happened. Investigating now..

pocmo:
It is only `testDownloadNotification` failing and only on one device. So I think we are still fine here.

liuche:
I'll kick off a new release.

PR #7542, opened 2020-01-08
Use Mozilla Android Components 27.0.2 with Fenix 3.1.

pocmo:
Updating the 3.1 release branch to use Mozilla Android Components 27.0.2.


pocmo:
It looks like we need another patch here and ship AC 27.0.3.

Issue #768, opened 2018-11-29
Know when a site is available as a PWA


pocmo:
### User Story
As a user, I want to be notified when a website I visit is available as a PWA, so I can install it as a short-cut on my homescreen. 

### Requirements
This is about notifying the user when a site has a web app manifest.

### Acceptance Criteria
- [ ] When I navigate to a site that is available as a PWA, my browser lets me know that I can install the PWA.
- [ ] The way my browser notifies me is not intrusive to my browsing experience.
- [ ] Telemetry Ticket TBD
- [ ] This does NOT include the PWA-app splash screen in the mock below

Abstract: https://share.goabstract.com/3815ff0e-5b89-4a41-a1de-7414f2e57964?mode=design&sha=63bb397f7bc0a0c98f7d225f90c092026dd264e5

**Note to ENG from Vesta:** 

-Most ENG work related to basic implementation will be completed by Tiger in Q4. This should be the only remaining PWA ticket from the original META #204 and it is focused on the UI aspect of the user experience
-Please ignore the splash screen in the mocks for now as it is out of scope for Q1 :)
-Feature brief for reference: https://docs.google.com/document/d/1Dd0OnivaUMe6rvEP5yi6Gt7Hz785ZstDLytaxnXz7Dk/edit

reinhart1010:
Duplicate: https://groups.google.com/forum/m/?utm_medium=email&utm_source=footer#!topic/fenix-nightly/y0s-eiLwA0o

apbitner:
Description updated with Abstract link showing First Use and Toolbar Notification flows

madb1lly:
This is a nice mockup, but what is the reason that the icon cannot be put in the toolbar like in Fennec? Is there less space in Fenix?

Also, "install" may be confusing as these are not really installed in the traditional Android sense - they will not appear in the list of installed apps. This could be mitigated by Fenix having a record of installed PWAs, e.g. "about:pwas" or "about webapps" where they can be managed. Firefox Accounts could also keep a record so that they can easily be reinstalled.

apbitner:
@madb1lly we have a handful of features that are only available depending on the website you're on such as reader mode, open in native app, install PWA, etc. We don't want to put all of those icons in the address bar so we're working on a way to make these contextual features easily noticeable without cluttering up the address bar.

liuche:
@NotWoods my understanding here is that once the notification dot is done, this bug will be done? If so I'm removing it from this board.

[...]

PR #7579, opened 2020-01-09
For #768: Add first time PWA dialog 

NotWoods:
### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
@@ -10,6 +10,7 @@ import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
 import androidx.coordinatorlayout.widget.CoordinatorLayout
+import androidx.navigation.fragment.findNavController
 import androidx.transition.TransitionInflater
 import com.google.android.material.snackbar.Snackbar
 import kotlinx.android.synthetic.main.fragment_browser.*
@@ -34,6 +35,7 @@ import org.mozilla.fenix.ext.components
 import org.mozilla.fenix.ext.nav
 import org.mozilla.fenix.ext.requireComponents
 import org.mozilla.fenix.ext.settings
+import org.mozilla.fenix.shortcut.FirstTimePwaObserver
 import org.mozilla.fenix.trackingprotection.TrackingProtectionOverlay
 
 /**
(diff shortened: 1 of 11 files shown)


thecodingdude:
Is this a modal that pops-up automatically? Can it be inline with the browser UI itself?

I detest the thought of continuing to browse the web with hideous pop-up UI's (the cookie banners are just awful). We got pop-up blockers for a reason ;-)

NotWoods:
This would only show up the very first time a PWA is encountered, similar to a popup in Fennec.

FluorescentHallucinogen:
@NotWoods What about something like this? :wink:

firefox-pwa-install-icon-1

firefox-pwa-install-icon-2

thecodingdude:
Much prefer the option above, seems slightly less intrusive.

app/src/main/java/org/mozilla/fenix/shortcut/FirstTimePwaObserver.kt
```diff
@@ -0,0 +1,39 @@
...
+    session: Session?,
+    owner: LifecycleOwner,
+    private val navController: NavController,
+    private val settings: Settings,
+    private val webAppUseCases: WebAppUseCases
+) : Session.Observer {
+
+    init {
+        if (session != null && settings.shouldShowFirstTimePwaFragment) {
+            session.register(this, owner, autoPause = true)
```
jonalmeida: Any reason we can't just use the LifecycleAwareFeature (rename to `FirstTimePwaFeature` as well)? Seems like you're trying to do the same for it already.

app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
```diff
@@ -114,14 +116,23 @@ class BrowserFragment : BaseBrowserFragment(), UserInteractionHandler {
...
-            context = requireContext(),
-            settings = requireContext().settings()
+            context = context,
+            settings = context.settings()
         ) {
             browserToolbarView.view
         }
         getSessionById()?.register(toolbarSessionObserver, this, autoPause = true)
         updateEngineBottomMargin()
+        FirstTimePwaObserver(
```
jonalmeida: You could also add a check for `shouldShowFirstTimePwaFragment` so you don't need to add the observer after it succeeded the first time.

app/src/main/java/org/mozilla/fenix/shortcut/FirstTimePwaObserver.kt
NotWoods: I register it here so that I can do the settimgs check you mentioned in the other comment, without having to repeat the check if we use this class elsewhere.

jonalmeida: I think we want to move away from this style of "setup observers in the init block". I'm quite guilty of doing this myself, but @csadilek had some good points to avoid it.

See: https://github.com/mozilla-mobile/android-components/issues/5449

liuche:
Any concerns w/ landing this?

also tagging @apbitner to check it out!

NotWoods:
@liuche Should be safe to land

apbitner:
@NotWoods can we make sure we're matching [the spec](https://share.goabstract.com/0a9600f9-2a52-47fd-be4a-0ae54a881aea?)? Some of the fonts look off along with the line spacing.

NotWoods:
UI issues have been fixed

ekager:
Because of conflicts and contributor CI issues closing for https://github.com/mozilla-mobile/fenix/pull/9308 (and fixed conflicts)

PR #7589, opened 2020-01-10
Closes #7587: Add shortcuts.xml for fennecNightly variant.

### LINKED ISSUE #7587: Create shortcuts.xml override for fennecNightly build type
We are overriding shortcuts.xml in different build types because it contains the package name. We need to do the same for the fennecNightly flavor.

pocmo:
We have override versions for all our build variants since those files contain the package name. So we needed one for the fennecNightly variant too. Tested that now shortcuts are working when using a fennecNightly build.


grigoryk:
@pocmo we'd need `fennecProduction` and `fennecBeta` overrides as well, right?

grigoryk:
Added notes to our board so we don't forget.

pocmo:
> @pocmo we'd need fennecProduction and fennecBeta overrides as well, right?

Yep!

PR #7610, opened 2020-01-10
Issue #4893: Enable addon migration

csadilek:
Just bringing in the new `migrateAddons` call.


csadilek:
Have to block this from landing until https://bugzilla.mozilla.org/show_bug.cgi?id=1604222 lands in GV. This is so we can disable and indicate that the addon is not supported.

csadilek:
@grigoryk and I did a manual test run to verify. This is ready to land now.

Issue #6274, opened 2019-10-25
SearchFragment is noticibly slow on first run


Baron-Severin:
## Steps to reproduce
- Open a fresh Fenix instance
- Click on the URL bar

### Expected behavior
- Search fragment should appear quickly

### Actual behavior
- Search fragment is noticeably slow to appear
- This only happens the first time the SearchFragment is launched

### Device information

* Android device: physical Pixel 2, emulator
* Fenix version: 9d91108f4e6576860bc1da734768b4422d08cba8 (current head of master)

### Notes
I didn't spend much time on this, but on my emulator the first launch (from URL bar click to onViewCreated) took about 160 MS.  Follow up launches varied, but hovered around 30-60 MS.  I believe the animated transition makes these numbers look worse than they are (follow up launches feel quite snappy).

emalysz:
After capturing this in the profiler, SearchFragment's onCreateView() indeed takes longer after first run. First run: 1,048,032 μs. Second run: 322,882 μs.
Breaking this down, it looks like the increase is mostly due to the time spend in <init> for both ToolbarView and AwesomeBarView.

hakkikaancaliskan:
This problem looks like fixed for me after
Nightly 191128 11:47 (Build #13321153)
📦: 24.0.0, 3183401f0
🦎: 72.0a1-20191126093448
XZ1, Android 10
Can someone confirm? Or initializing megazord asynchronously meet the required time of searchfragment?

Baron-Severin:
@boek and I were still able to reproduce this, although it's less noticeable on newer devices.  We did some investigation and it looks like there's also a very substantial amount of time spent in the `SearchFragment` `consumeBy(searchStore)` block.  This might be easier for us to fix since view init is handled in AC.  Once we have a better idea of what's going on, we can open an AC issue as well.

Baron-Severin:
It may be possible to fix the perceived perf issue here without modifying toolbar internals by [inflating the toolbar on a background thread](https://developer.android.com/reference/android/support/v4/view/AsyncLayoutInflater.html) on app init.

hakkikaancaliskan:
> 
> 
> It may be possible to fix the perceived perf issue here without modifying toolbar internals by [inflating the toolbar on a background thread](https://developer.android.com/reference/android/support/v4/view/AsyncLayoutInflater.html) on app init.

[...]

PR #7622, opened 2020-01-11
Issue 6274: Asynchronously load the search fragment and its components

hawkinsw:
WIP as it stands. Undocumented, for now. 

NB: There are three *long* inflations that happen when the Search Bar is first displayed. This WIP only removes the first of those three. However, I do see a noticeable performance improvement and this framework seems to be applicable to the other two long inflations. If you think what is proposed here is a sane idea, we can move forward on applying this tool to the other two. Until then, I want to ask for your feedback!

Here is the idea:
1. Create a thing called a `WaitableAsyncInflater`.
2. Use that to asynchronously pre-inflate the search fragment (and its components) when the Home Fragment is created.
3. Get the pre-created view when the Search Bar is displayed.

A `WaitiableAsyncInflater` is a wrapper around an asynchronous layout inflater upon which a caller can invoke `get()` to obtain the inflated view. If the inflation already happened and the view is created by the time the call to `get()` happens, the view is returned immediately. If the inflation is in progress asynchronously at the time of the `get()` call, the function waits for inflation to complete and then returns the view. If the inflation has not yet started by the time of the `get()` call, the inflation happens synchronously and the view is returned. 

Users invoke `inflate()` when they simply want to start the layout asynchronously and are not interested in the view that is actually inflated. 

I wonder if this is a sane approach? I am sharing this with you early to get your feedback on both 

1. the method and
2. the perceived decrease in latency for loading the search fragment the first time.
### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/utils/WaitableAsyncInflater.kt
+++ b/app/src/main/java/org/mozilla/fenix/utils/WaitableAsyncInflater.kt
@@ -1,0 +1,83 @@
+package org.mozilla.fenix.utils
+
+import android.content.Context
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+
+/**
+ * Wrapper around an asynchronous layout inflater that gives callers the ability to
+ * synchronously access the inflated view, whether the asynchronous inflation is
+ * complete or in progress.
+ *
+ * A WaitiableAsyncInflater is a wrapper around an asynchronous layout inflater upon which
+ * a caller can invoke get() to obtain the inflated view. If the inflation already happened
+ * and the view is created by the time the call to get() happens, the view is returned
+ * immediately. If the inflation is in progress asynchronously at the time of the get()
+ * call, the function waits for inflation to complete and then returns the view. If the
+ * inflation has not yet started by the time of the get() call, the inflation happens
+ * synchronously and the view is returned.
+ *
+ * Users invoke inflate() when they simply want to start the inflation asynchronously and
+ * are not interested in the view itself.
+ *
+ * This class is thread safe.
+ */
+class WaitableAsyncInflater(val resId: Int) {
+
+    var asyncInflationStarted = false
+    var condition = Object()
+    var view: View? = null
+    val scope = CoroutineScope(Dispatchers.IO)
+
+    private fun startAsynchronousInflation(context: Context, group: ViewGroup, attachToRoot: Boolean = false) {
+        scope.launch {
+            view = LayoutInflater.from(context).inflate(resId, group, attachToRoot)
+            synchronized(condition) {
+                condition.notifyAll()
+            }
+        }
+        asyncInflationStarted = true
+    }
+
+    private fun doSynchronousInflation(context: Context, group: ViewGroup, attachToRoot: Boolean = false) {
+        val inflater = LayoutInflater.from(context)
+        view = inflater.inflate(resId, group, attachToRoot)
+    }
+
+    /**
+     * TODO
+     *
+     * In this documentation, be sure to mention that *it is assumed* that the view can be inflated
+     * off the MT.
+     */
+    @Synchronized
+    fun inflate(context: Context, group: ViewGroup, attachToRoot: Boolean = false) {
+        if (view == null && !asyncInflationStarted) {
+            startAsynchronousInflation(context, group, attachToRoot)
+        }
+    }
+
+    /**
+     * TODO
+     */
+    @Synchronized
+    fun get(context: Context, group: ViewGroup, attachToRoot: Boolean = false): View {
+        if (view == null && !asyncInflationStarted) {
+            // When there is neither a view already created nor an asynchronous inflation in progress
+            // do the inflation on demand.
+            doSynchronousInflation(context, group, attachToRoot)
+        } else {
+            // Otherwise, potentially wait for the view to be created.
+            while (view == null) {
+                synchronized(condition) {
+                    condition.wait()
+                }
+            }
+        }
+        return view!!
+    }
+}
--- a/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
@@ -41,12 +41,16 @@ import androidx.transition.TransitionInflater
 import com.google.android.material.snackbar.Snackbar
 import kotlinx.android.synthetic.main.fragment_home.*
 import kotlinx.android.synthetic.main.fragment_home.view.*
+import kotlinx.android.synthetic.main.fragment_home.view.toolbar_wrapper
+import kotlinx.android.synthetic.main.fragment_search.view.*
+import kotlinx.coroutines.launch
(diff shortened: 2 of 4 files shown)


app/src/main/java/org/mozilla/fenix/utils/WaitableAsyncInflater.kt
```diff
@@ -0,0 +1,65 @@
...
+                                                            _: ViewGroup? ->
+                view = newView
+                synchronized(condition) {
+                    condition.notifyAll()
+                }
+            })
+        asyncInflater = newInflater
+    }
+
+    private fun doSynchronousInflation(context: Context, group: ViewGroup) {
```
MarcLeclair: I'm not sure you need to the `doSynchronousInflation` since `AsyncInflater` will default to do it on the MT if it can't do it async.

Ref: 

> [For a layout to be inflated asynchronously it needs to have a parent whose ViewGroup.generateLayoutParams(AttributeSet) is thread-safe and all the Views being constructed as part of inflation must not create any Handlers or otherwise call Looper.myLooper(). If the layout that is trying to be inflated cannot be constructed asynchronously for whatever reason, AsyncLayoutInflater will automatically fall back to inflating on the UI thread. ](https://developer.android.com/reference/androidx/asynclayoutinflater/view/AsyncLayoutInflater)

MarcLeclair: Unless I misunderstood the use of the `doSynchronousInflation` 😛

hawkinsw: Thanks for the review! I _think_ we are talking about two different things. `WaitiableAsyncInflater` does this synchronously because it needs to be done immediately not because it needs to be done on the UI thread. Please tell me where I am wrong, but the way I read the document you quoted (which was really helpful to read, thank you!), `AsyncLayoutInflater` will properly handle whether or not it can do the actual inflation in a background thread or only on the UI thread but it will do the inflation asynchronously no matter what. 

`WaitiableAsyncInflater` has to make the distinction between doing it synchronously and asynchronously. In the case where `get()` invokes `doSynchronousInflation()`, the caller of `get()` needs the view but there is a) no view already inflated and b) no ongoing asynchronous action to create that view. Because of the way that the `AsyncLayoutInflater` notifies the callback that inflation is complete, `WaitiableAsyncInflater` cannot simply launch an `AsyncLayoutInflater` job in this case and wait for the callback because it will deadlock. 

This is the case: 
1. From the UI thread, a user of the `WaitiableAsyncInflater` calls `get()`.
2. There is no view and no ongoing asynchronous action to create it, so `WaitiableAsyncInflater` starts an asynchronous action and then blocks the UI thread until it is ready.
3. When `AsyncLayoutInflater` completes the asynchronous inflation, it posts a task to the UI thread to invoke the associated callback. 

Because `WaitiableAsyncInflater` is blocking the UI thread on the completion of the asynchronous inflation, the UI thread will not continue processing incoming tasks posted to it and, therefore, will never execute the callback. Boom.

Please, please scrutinize my logic to make sure that I am not saying anything you disagree with. Thanks again for a) the review, b) including the documentation and c) checking my logic.

jonalmeida:
Drive-by comment 🚗 : This is super cool! It seems in-line with some other view inflation stuff we wanted to put into a-c after adding in individual patches for lazy inflation in ReaderView and FindInPage.

This is definitely something we should add into components. 🙂 

The a-c ticket for reference: https://github.com/mozilla-mobile/android-components/issues/5491

mcomella:
I haven't looked at the code yet but the general premise seems like a good idea: it makes doing a background inflation readable and fairly safe.

Running this locally, I noticed this is still an issue on my P2 but perhaps that's because it's only hooked up to one inflation (iiuc).

wrt fixing #6274 (SearchFragment is slow on first run), I would be curious to see what the root problems are and if there's a simpler way to solve this before jumping to async inflation. I remember this bug being a sudden change, e.g. maybe the GeckoSession initialization is reading the disk and we can undo that to solve this. That being said, you pointed out that there are 3 slow inflations so perhaps it's worth doing anyway and, given our time constraints, if we have this solution, it already works, and it's fairly safe, maybe we should just go for it. Perhaps we can chat tomorrow after stand-up about the root causes and our options.

---

Another place the WaitableAsyncInflater could be handy is if Gecko must initialize on the main thread, could we be inflating HomeFragment's view hierarchy on a background thread? Or if the root of HomeFragment is inflating on the main thread, could we inflate a sub-hierarchy (e.g. the RecycleView) on a background thread?

hawkinsw:
> I haven't looked at the code yet but the general premise seems like a good idea: it makes doing a background inflation readable and fairly safe.

Great!

> 
> Running this locally, I noticed this is still an issue on my P2 but perhaps that's because it's only hooked up to one inflation (iiuc).
> 

Would you be willing to send me the profile you have where you still see the issue? I just wanted to see if you are seeing what I see. When I run it on my P2 the problem is gone. I am curious about what I missed! Thank you!

> wrt fixing #6274 (SearchFragment is slow on first run), I would be curious to see what the root problems are and if there's a simpler way to solve this before jumping to async inflation. I remember this bug being a sudden change, e.g. maybe the GeckoSession initialization is reading the disk and we can undo that to solve this. That being said, you pointed out that there are 3 slow inflations so perhaps it's worth doing anyway and, given our time constraints, if we have this solution, it already works, and it's fairly safe, maybe we should just go for it. Perhaps we can chat tomorrow after stand-up about the root causes and our options.

I believe (and definitely could be wrong!) that the context of this delay is a slow transition from when the user first taps the closed URL bar to showing the Search Fragment. *If* I am understanding correctly, there are no session creations that happen in this delay. Again, *if* I am thinking clearly, the session creation issue is solely related to *the other Search Bar* issue (#7707 ). So many issues here!

I think that we should definitely talk about this in person and figure out the right way to proceed!

> 
> Another place the WaitableAsyncInflater could be handy is if Gecko must initialize on the main thread, could we be inflating HomeFragment's view hierarchy on a background thread? Or if the root of HomeFragment is inflating on the main thread, could we inflate a sub-hierarchy (e.g. the RecycleView) on a background thread?

I think that you and @MarcLeclair had the same idea and I think it is an excellent one!

Baron-Severin:
> wrt fixing #6274 (SearchFragment is slow on first run), I would be curious to see what the root problems are and if there's a simpler way to solve this before jumping to async inflation. 

This seems prudent to me.  I feel like we should keep this option open if we aren't able to satisfactorily resolve the issue otherwise, though.  This delay is pretty bad, and particularly if we intend to land this component for use elsewhere, the change doesn't add all _that_ much complexity.

> I remember this bug being a sudden change

FWIW, I don't think this is true.  We noticed this suddenly, when we had an unrelated reason to look closely at the `SearchFragment`, but I don't think we really know when this was introduced.  I remember checking the release build after finding it on master and being able to reproduce it there, at the very least.  I have no way of knowing, but I wouldn't be surprised if this had been around for weeks, or even months.

liuche:
@mcomella just wondering, is this something that we want to keep open, or can we close/fix+merge it?

mcomella:
> @mcomella just wondering, is this something that we want to keep open, or can we close/fix+merge it?

We're not focused on this area and I don't know when we will be again so I'm happy to close to avoid polluting the open PR list.

PR #7639, opened 2020-01-13
Test that 'baseline' pings are sent by Fenix

Dexterp37:
This is an initial instrumented test for Fenix that checks if a 'baseline' ping is generated when
going to background.

~~**Note**: this won't work until Fenix is a new A-C snapshot which contains the latest updates from mozilla/glean#634 and mozilla-mobile/android-components#5549~~

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -494,6 +494,8 @@ dependencies {
         exclude group: 'com.android.support', module: 'support-annotations'
     }
 
+    androidTestImplementation Deps.androidx_junit
+    androidTestImplementation Deps.androidx_work_testing
     androidTestImplementation Deps.mockwebserver
     testImplementation Deps.mozilla_support_test
     testImplementation Deps.androidx_junit
--- a/app/src/androidTest/java/org/mozilla/fenix/glean/BaselinePingTest.kt
+++ b/app/src/androidTest/java/org/mozilla/fenix/glean/BaselinePingTest.kt
@@ -1,0 +1,107 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.glean
+
+import androidx.test.core.app.ApplicationProvider
+import androidx.test.platform.app.InstrumentationRegistry
+import androidx.test.rule.ActivityTestRule
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import org.junit.Assert.assertEquals
+
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+import androidx.test.uiautomator.UiDevice
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.launch
+import mozilla.components.service.glean.Glean
+import mozilla.components.service.glean.config.Configuration
+import mozilla.components.service.glean.testing.GleanTestLocalServer
+import org.json.JSONObject
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertTrue
+import org.junit.BeforeClass
+import org.mozilla.fenix.HomeActivity
+import org.mozilla.fenix.helpers.HomeActivityTestRule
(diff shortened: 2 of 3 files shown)


app/build.gradle
```diff
@@ -628,7 +630,7 @@ def glean_android_components_tag = (
 // Generate markdown docs for the collected metrics.
 ext.gleanGenerateMarkdownDocs = true
 ext.gleanDocsDirectory = "$rootDir/docs"
-apply from: 'https://github.com/mozilla-mobile/android-components/raw/' + glean_android_components_tag + '/components/service/glean/scripts/sdk_generator.gradle'
+apply from: 'https://github.com/mozilla-mobile/android-components/raw/master/components/service/glean/scripts/sdk_generator.gradle'
```
Dexterp37: TODO: this needs to be reverted before merging.

app/src/androidTest/java/org/mozilla/fenix/FirefoxTestRunner.kt
```diff
@@ -4,13 +4,81 @@
...
+ *
+ * @return an `Int` containing the server port.
+ */
+@VisibleForTesting(otherwise = VisibleForTesting.NONE)
+internal fun getPingServerPort(): Int {
+    val testRunner: FirefoxTestRunner = InstrumentationRegistry.getInstrumentation() as FirefoxTestRunner
+    return testRunner.pingServerPort!!
+}
 
 class FirefoxTestRunner : AndroidJUnitRunner() {
```
boek: @rpappalax Do you have any concerns with adding this code here?

rpappalax: You can add this here but then the lazy ping server will be running in the background for every test (which it prob doesn't need to be). It seems like it might make more sense to extend the existing Mock webserver with anything you need to add and then init it in your test itself.  That way you could also shut it down right after the test was done.

rpappalax: Something like this: https://github.com/mozilla-mobile/fenix/blob/master/app/src/androidTest/java/org/mozilla/fenix/ui/NavigationToolbarTest.kt#L38-L49

Baron-Severin: +1 on everything @rpappalax said, with extra emphasis that this approach could leak state between tests if we're not super careful.  I'd sleep better if we were reinstantiating this between tests.

Dexterp37: The reason why I didn't do that, is that the server address needs to be know when using [the test rule](https://github.com/mozilla-mobile/fenix/pull/7639/files#diff-cc48215f38c48b0b4bfa4f9c1a9833acR39). That test rule ensures that the Glean singleton points to the local server.

Maybe there's another way to do that, but I'm not super familiar with JUnit and rules. Any suggestion on how to achieve that, @rpappalax or @Baron-Severin ?

Baron-Severin: Apologies in advance if I'm misunderstanding your problem, but these changes _seem_ to work for me.

I'm creating a `MockWebServer` at the top of `BaselinePingTest.kt` and passing its port into the rule.  LMK if I'm missing something here and I'll try to help.

https://github.com/Baron-Severin/fenix/compare/a799d1a493d6ef79a491ad879e1e7580354e43b0...Baron-Severin:integration_test

liuche:
Sounds like there are some changes that could be tried out here, so I'll hold off on landing for now.

app/src/androidTest/java/org/mozilla/fenix/FirefoxTestRunner.kt
Dexterp37: Woha, @Baron-Severin , you were indeed right. This works like a charm, I've updated my PR with your changes.

Dexterp37:
> 
> 
> Sounds like there are some changes that could be tried out here by @Dexterp37, so I'll hold off on landing for now.

@liuche I just fixed the PR as suggested by @Baron-Severin , so this is now ready for another look!

Issue #5545, opened 2019-09-24
In settings, I can choose whether I want my browser to remember my logins


ekager:
### Why/User Benefit/User Problem
As a user I want to choose whether Fenix saves my logins for all sites.

### What/Requirements
- Depends on #5542 
- Settings page to give users the choice to save or not save logins on all sites.

### Acceptance Criteria (how do I know when I’m done?)
- A user can choose to save or not save logins globally within Fenix. See #1284 for mocks.

sv-sdeiac:
Verified as fixed using the latest version of Firefox Preview Nightly 1/16 build #20160806.

Devices: 
- Nokia 6 (Android 7.1.1);
- Motorola Moto G6 (Android 8.0);
- Samsung Galaxy Note 8 (Android 9);
- Google Pixel (Android Q).

Issue #5542, opened 2019-09-24
Allow saving login when I log into a website


liuche:
The GV API that triggers doorhangers will be done as part of #6696 

### Why/User Benefit/User Problem
When a user logs in on a website, they can save the login.

### What/Requirements

This is only partially waiting, but since it's one of our MUSTs for the quarter, we can start working on the UI parts.

- [x] [Mock this out so it unblocks subsequent work - will receive API 11/8] Fenix can detect that a login has been submitted on a site - GV bug for autofill API listening for [PasswordManager:onFormSubmit ](https://searchfox.org/mozilla-central/rev/325c1a707819602feff736f129cb36055ba6d94f/toolkit/components/passwordmgr/LoginManagerContent.jsm#1519-1528): https://bugzilla.mozilla.org/show_bug.cgi?id=1577005
- [x] Show a doorhanger w/ save/don't-save options
- [ ] ~~Add Exceptions for sites users don't want to save logins for~~ #7094
- [x] Fenix will write login credentials to Sync Login Storage using existing AS/AC APIs

### Acceptance Criteria (how do I know when I’m done?)
Logins can be saved when submitting on a site
Does not include viewing logins, that is #5544 

#### Not in this bug
Choosing secure local storage, see #5550

liuche:
I chatted w/ GV about this, and at the very highest level, Fenix will be doing login saving (as in, GV will pass the login blob to be saved to Logins storage).

This means that we can start the in-app notification work for saving a login, and just write dummy content to the Logins storage. #5074 should be in Fenix with the next AC version 20.0 on 11/4.

We'll receive a GV API for capture by 11/8, to unblock the other part of this work (triggering the login).

liuche:
Moving this into the sprint because we're moving out #5543 due to new GV work being done on it.

ekager:
Resized to an E3 to build up the UI and saving a "login" to the storage while waiting on the GV APIs.

cpeterso:
The GV bug has been added to GV's November sprint.

https://bugzilla.mozilla.org/show_bug.cgi?id=1577005#c13

liuche:
@ekager could we show this dialog when GV detects that there is a login? Also, can we add the exceptions list? Basically everything except writing the login to the Sync storage.

Would be down to land this preffed off entirely, or Nightly-only just disabling the "Save" button in the dialog.

liuche:
[...]

PR #7667, opened 2020-01-14
For #5545 For #5542 Closes #6696 Integrate logins API, adds Settings for Autofilling/Saving Logins

### LINKED ISSUE #6696: Integrate Autofill AC API in Fenix for GV Login Autofill Delegate (Land Autofill in Fenix)
#5543 is tracking the AC work of creating the AC wrapper around the new GV API. After the GV APIs land, Fenix will need to do work in this ticket to integrate the AC API and actually autofill logins.

### Acceptance criteria
- [ ] Logins autofill (by initializing GV RT)

### Dependencies
- [x] AC work for GV autofill API #5543 
- [x] Login Storage Gecko Toolkit Implementation for heuristics of logins
[Bug 1596660 - Create a GeckoView nsILoginManagerStorage backend](https://bugzilla.mozilla.org/show_bug.cgi?id=1596660)
- [x] Login Storage GV API Implementation to autofill in Fenix (depends on Gecko Toolkit)
[Bug 1593069 - Implement the Login Storage API](https://bugzilla.mozilla.org/show_bug.cgi?id=1593069)

ekager:
### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/settings/LoginsFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/LoginsFragment.kt
@@ -20,6 +20,7 @@ import androidx.lifecycle.lifecycleScope
 import androidx.navigation.fragment.findNavController
 import androidx.preference.Preference
 import androidx.preference.PreferenceFragmentCompat
+import androidx.preference.SwitchPreference
 import kotlinx.coroutines.Dispatchers.Main
 import kotlinx.coroutines.launch
 import mozilla.components.concept.sync.AccountObserver
@@ -36,7 +37,7 @@ import org.mozilla.fenix.ext.settings
 import org.mozilla.fenix.ext.showToolbar
 import java.util.concurrent.Executors
 
-@Suppress("TooManyFunctions")
+@Suppress("TooManyFunctions", "LargeClass")
 class LoginsFragment : PreferenceFragmentCompat(), AccountObserver {
 
     @TargetApi(M)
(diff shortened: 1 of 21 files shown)


ekager:
To explain the autofill logic: according to this mock comment autofill should always be disabled / off when "save logins" is not enabled.
Screen Shot 2020-01-14 at 5 01 47 PM

ekager:
Screenshot_20200114-170309
Screenshot_20200114-170316
Screenshot_20200114-170345
Screenshot_20200114-170354

app/src/main/java/org/mozilla/fenix/settings/LoginsFragment.kt
```diff
@@ -264,6 +287,14 @@ class LoginsFragment : PreferenceFragmentCompat(), AccountObserver {
         }
     }
 
+    private fun navigateToSaveLoginSettingFragment() {
+        if (findNavController().currentDestination?.id == R.id.loginsFragment) {
```
Baron-Severin: Why does this condition on the current destination?  What happens if it doesn't?

app/src/main/java/org/mozilla/fenix/settings/logins/SaveLoginSettingFragment.kt
```diff
@@ -0,0 +1,43 @@
...
+    override fun onCreatePreferences(savedInstanceState: Bundle?, rootKey: String?) {
+        setPreferencesFromResource(R.xml.save_logins_preferences, rootKey)
+    }
+
+    override fun onResume() {
+        super.onResume()
+        showToolbar(getString(R.string.preferences_passwords_save_logins))
+        bindSave()
+        bindNeverSave()
+        setupRadioGroups()
```
Baron-Severin: This is a simple enough example that it's not scary here, but there's an important concept here that can make code a lot more difficult to work with as it grows.  

This method will fail if these methods are called out of order (they're ["temporally coupled"](https://www.pluralsight.com/tech-blog/forms-of-temporal-coupling/#sequencing), because everything in programming needs an intimidating name), but there's nothing here telling the compiler (or the reader) that that will happen.  So we need to either read through the method implementations, or just poke around at runtime to find that out.

The easiest way to clean up this problem is, wherever possible, to avoid having methods reach outside of their own scope.  In this case, that's really easy to do.  `bindSave` and `bindNeverSave` can return their buttons, and `setupRadioGroups` can accept them as parameters.  Then the compiler will enforce that they're called in order, and whoever is reading the code isn't forced to keep class state in mind while reading a method.

That could look something like this (names and structure could be played with a bit, but you get the point):

```kotlin
override fun onResume() {
    super.onResume()
    showToolbar(getString(R.string.preferences_passwords_save_logins))
    val save = bindSave()
    val neverSave = bindNeverSave()
    setupRadioGroups(save, neverSave)
}

private fun bindSave(): RadioButtonPreference {
    val keyStrict = getString(R.string.pref_key_save_logins)
    return requireNotNull(findPreference(keyStrict))
}

private fun bindNeverSave(): RadioButtonPreference {
    val keyStandard = getString(R.string.pref_key_never_save_logins)
    return requireNotNull(findPreference(keyStandard))
}

private fun setupRadioGroups(radioNeverSave: RadioButtonPreference, radioSave: RadioButtonPreference) {
    radioNeverSave.addToRadioGroup(radioSave)
    radioSave.addToRadioGroup(radioNeverSave)
}
```

hakkikaancaliskan:
So sorry but just a little question about or.
Is this covering autofilling via Android's autofill service like filling via bitwarden? Or just internal Fenix autofill?

app/src/main/java/org/mozilla/fenix/settings/LoginsFragment.kt
ekager: It used to be a race condition that if you pressed on two settings items at once time it would crash but we've fixed that in https://github.com/mozilla-mobile/fenix/pull/6999 so removing this!

PR #7686, opened 2020-01-14
Update branding of fennecNightly build type..

### LINKED ISSUE #7203: Update branding of fennecNightly build type
Currently a build of `fennecNightly` will look just like any other "Firefox Preview" build. That's probably not our intention and can be quite confusing.

My assumption (!) is that we want to:
- Name it "Firefox Nightly" like the existing app
- Use the same (updated) icon as the existing app
- Update the big "Firefox Preview" banner on the home screen with something else (?)
- Update any other "Firefox Preview" references - are there any?

pocmo:
Closes #7203: Reusing the Fennec Nightly icon and fixing the location of the `static_strings.xml` file.


pocmo:
The UI test failure looks unrelated (the fennec build type is not tested anyways; and there's not a specific test failing, the whole device looks like it is in a bad state.)

Issue #7092, opened 2019-12-11
Migration screen


liuche:
### Why/User Benefit/User Problem
As a user I'd like to see information as my Firefox upgrades to Fenix/new browser.

### Context
From Firefox, we'll receive a store that state about the migration, that Fenix can observe.

### Dependencies
- [x] Final copy
- [x] AC API for migration store state, to observe when migration is complete https://github.com/mozilla-mobile/android-components/issues/5509
- Minor UX changes, if at all

### Acceptance Criteria (how do I know when I’m done?)
- [x] Conditionally display screen if Migrating vs new user
- [x] Handle all different entry points into Firefox (e.g. View Intents, Search widget?, PB, homscreen shortcuts, etc)
- [x] Lock user during migration, do not expose browsing capability (database lock) - otherwise, Firefox might hang with multiple writes to the db
- [x] ~~(Mocks TBD) Handle error states (e.g. failed bookmarks) and display to user~~
- [x] Link to "Learn more/FAQ" on "Start browsing" screen - Clicking this link will **NOT** skip onboarding
- [x] Make sure this check ("did the user migrate?") doesn't check affect cold startup (cc @mcomella can talk to @pocmo about measuring this effect) - we might take this out after X months if it's a big impact?

### Mocks
Relevant is the **Data Migration** section of https://miro.com/app/board/o9J_kw1CGow=/

vesta0:
@liuche is this similar to existing #6511

liuche:
Checking on the dependencies:
@apbitner are there any additional changes to the UI, that came out of the user study?
@mheubusch are there final strings for this? String freeze for migration strings in 1/15, so we need to have those string so @sblatz can pre-land them.

liuche:
Sorry @sblatz there was some miscommunication here - @jonalmeida will actually be doing this piece, just needed a Fenix-by-AC label.

pocmo:
@jonalmeida will help us push that into a functional, "good enough" state for Nightly. There will likely still be follow-up for Beta/Production migration (which may need a Fenix owner). :)

apbitner:
@liuche The UI and strings in the Data Migration section are final

liuche:
@jonalmeida could you go through the checklist in comment 0, and check off all the parts that you finished with that PR? if they're mostly done and there's no value in keeping this bug open bc the only remaining work is polish or something, could you file another bug listing off what still needs to be done?

jonalmeida:
[...]

PR #7706, opened 2020-01-15
For #7092: Add initial Migration UI.

jonalmeida:
This is PR does multiple things:
 1. Moving the `MigrationStore` to the Components class (and replacing the support-migration dependency on fennec variants with all).
 1. Integrates the `MigrationIntentProcessor` into the `IntentReceiverActivity`.
 2. Create a `MigrationProgressActivity` that extends `AbstractMigrationProgressActivity` based on https://github.com/mozilla-mobile/android-components/pull/5570 changes.

In the migration UI, we only show the progress for the following items (per designs): Bookmarks, History, Logins, Settings.

out

### Pull Request checklist

### After merge

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/IntentProcessors.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/IntentProcessors.kt
@@ -15,6 +15,8 @@ import mozilla.components.feature.pwa.intent.WebAppIntentProcessor
 import mozilla.components.feature.pwa.intent.TrustedWebActivityIntentProcessor
 import mozilla.components.feature.search.SearchUseCases
 import mozilla.components.feature.session.SessionUseCases
+import mozilla.components.support.migration.MigrationIntentProcessor
+import mozilla.components.support.migration.state.MigrationStore
 import org.mozilla.fenix.BuildConfig
 import org.mozilla.fenix.home.intent.FennecBookmarkShortcutsIntentProcessor
 import org.mozilla.fenix.test.Mockable
@@ -29,7 +31,8 @@ class IntentProcessors(
     private val sessionUseCases: SessionUseCases,
     private val searchUseCases: SearchUseCases,
     private val httpClient: Client,
-    private val customTabsStore: CustomTabsServiceStore
+    private val customTabsStore: CustomTabsServiceStore,
+    private val migrationStore: MigrationStore
 ) {
     /**
      * Provides intent processing functionality for ACTION_VIEW and ACTION_SEND intents.
@@ -66,5 +69,9 @@ class IntentProcessors(
             WebAppIntentProcessor(sessionManager, sessionUseCases.loadUrl, ManifestStorage(context)),
             FennecBookmarkShortcutsIntentProcessor(sessionManager, sessionUseCases.loadUrl)
         )
+    }
+
+    val migrationIntentProcessor by lazy {
+        MigrationIntentProcessor(migrationStore)
     }
 }
--- a/app/src/main/java/org/mozilla/fenix/migration/MigrationProgressActivity.kt
+++ b/app/src/main/java/org/mozilla/fenix/migration/MigrationProgressActivity.kt
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-package org.mozilla.fenix
+package org.mozilla.fenix.migration
 
 import android.os.Bundle
 import android.view.LayoutInflater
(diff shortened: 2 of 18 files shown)


app/src/migration/java/org/mozilla/fenix/MigratingFenixApplication.kt
```diff
@@ -38,6 +39,10 @@ class MigratingFenixApplication : FenixApplication() {
 
         // The rest of the migrations can happen now.
         migrator.startMigrationIfNeeded(migrationStore, MigrationService::class.java)
+
+        // Start migration UI
+        val intent = Intent(this, MigrationProgressActivity::class.java)
+        startActivity(intent)
```
jonalmeida: This part is untested, but I believe this is the place to start the UI.

jonalmeida:
```
[task 2020-01-15T07:08:42.182Z] > Task :app:processGeckoNightlyDebugResources
[task 2020-01-15T07:09:12.182Z] > Task :app:compileGeckoNightlyDebugKotlin
[task 2020-01-15T07:09:16.182Z] e: /builds/worker/checkouts/src/app/src/main/java/org/mozilla/fenix/components/IntentProcessorType.kt: (10, 26): Unresolved reference: MigrationProgressActivity
[task 2020-01-15T07:09:16.182Z] e: /builds/worker/checkouts/src/app/src/main/java/org/mozilla/fenix/components/IntentProcessorType.kt: (23, 26): Unresolved reference: MigrationProgressActivity
[task 2020-01-15T07:09:16.182Z] 
```

I guess I need to put the activity in the `main` variant as well. 🤔

app/src/migration/java/org/mozilla/fenix/MigratingFenixApplication.kt
pocmo: * This does always start the migration UI? Wouldn't we need to check the state first?
 * As-is this doesn't work, I think:
   * From an application context you can't launch an activity without at least adding the FLAG_ACTIVITY_NEW_TASK flag.
   * For application starts without UI (e.g. some random background task) we wouldn't want to suddenly pop up the migration UI.

pocmo: I wonder if we have to launch it (and shortcut) `HomeActivity`? I assume that's where regular app launches as well as intents (after the migration processor ignored them) end up?

app/src/main/java/org/mozilla/fenix/components/IntentProcessors.kt
```diff
@@ -67,4 +70,8 @@ class IntentProcessors(
             FennecBookmarkShortcutsIntentProcessor(sessionManager, sessionUseCases.loadUrl)
         )
     }
+
+    val migrationIntentProcessor by lazy {
```
grigoryk: What about going the other direction - overwrite `IntentProcessor` in `src/migration`? And, following that pattern for any other changes you need.

That will allow you to avoid any `support-migration` from trickling into `main`, I think.

jonalmeida: We tried this during the original test fit, but anything in main can't reference code in a variant. It can only work the opposite way.

See: https://github.com/mozilla-mobile/android-components/pull/5548#issuecomment-573510696

app/src/migration/java/org/mozilla/fenix/MigratingFenixApplication.kt
jonalmeida: I think you're right. I just nicked this bit in on the end, and wanted to get feedback if this was the right location. I'll look into this more now.

app/src/main/java/org/mozilla/fenix/migration/MigrationProgressActivity.kt
```diff
@@ -0,0 +1,151 @@
...
+import org.mozilla.fenix.ext.components
+
+class MigrationProgressActivity : AbstractMigrationProgressActivity() {
+    private val statusAdapter = MigrationStatusAdapter()
+    override val store: MigrationStore by lazy { components.migrationStore }
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        setContentView(R.layout.activity_migration)
+        init()
+        super.onCreate(savedInstanceState)
```
pocmo: nit: Is calling `super.onCreate()` at the end on purpose? I have no problem with that, it just looks different than usual. :)

jonalmeida: It's funny, I was reading [this](https://stackoverflow.com/a/18874519) to see if it mattered for onStart and onStop, but forgot about onCreate. :)

I'll put it on the top!

jonalmeida:
The build failure is from Glean, so I guess I need to rebase master:
```
[task 2020-01-15T19:27:57.035Z] e: /builds/worker/checkouts/src/app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt: (501, 17): Type mismatch: inferred type is Configuration but Boolean was expected
```

app/src/main/java/org/mozilla/fenix/components/IntentProcessors.kt
grigoryk: 👍

jonalmeida:
We're correctly failing a bunch of tests for the IntentReceiver and other intent unit tests.

PR #7717, opened 2020-01-15
Always initialize glean. Set upload flag after migration.

pocmo:
* We need to always initialize glean.
* At first it is disabled until we a potential migration is completed to determine whether we can enable it. This is to prevent us from collecting telemetry and then later knowing that we should not have.
* Glean then later gets enabling during the migration (so that the migration can upload telemetry too) or once `GleanMetricsService` gets started.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
+++ b/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
@@ -20,6 +20,9 @@ import kotlinx.coroutines.runBlocking
 import mozilla.appservices.Megazord
 import mozilla.components.concept.push.PushProcessor
 import mozilla.components.service.experiments.Experiments
+import mozilla.components.service.glean.Glean
+import mozilla.components.service.glean.config.Configuration
+import mozilla.components.service.glean.net.ConceptFetchHttpUploader
 import mozilla.components.support.base.log.Log
 import mozilla.components.support.base.log.logger.Logger
 import mozilla.components.support.base.log.sink.AndroidLogSink
@@ -37,6 +40,7 @@ import org.mozilla.fenix.session.VisibilityLifecycleCallback
 @SuppressLint("Registered")
 @Suppress("TooManyFunctions")
 open class FenixApplication : LocaleAwareApplication() {
+    private val logger = Logger("FenixApplication")
 
     open val components by lazy { Components(this) }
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -56,6 +60,27 @@ open class FenixApplication : LocaleAwareApplication() {
...
+        // We need to always initialize Glean and do it early here. Note that we are disabling it
+        // here too (uploadEnabled = false). If needed Glean will be enabled later by the migration
+        // code (if this user used to be a fennec user with the right flags enabled) or by
+        // GleanMetricsService if telemetry is enabled for this user.
+        // It is important that this initialization happens *here* before calling into
+        // setupInMainProcessOnly() which behaves differently for fenix and fennec builds.
+        // Glean needs to be disabled initially because otherwise we may already collect telemetry
+        // before we know whether we want that (which is *after* the migration).
+        // As a side effect this means pings submitted between the initialization here and until we
+        // potentially enable Glean would be lost. However such pings do not exist at this moment.
```
grigoryk: GV may emit something, but that's done after our blocking/telemetry migration, so it should be fine.

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -56,6 +60,27 @@ open class FenixApplication : LocaleAwareApplication() {
...
+        logger.debug("Initializing Glean (uploadEnabled=false)")
+        Glean.initialize(
```
boek: I know the performance team has done a bit of investigation around our cold startup time and deferred this as much as possible. We may want to give them a heads up with this

cc: @mcomella

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -56,6 +60,27 @@ open class FenixApplication : LocaleAwareApplication() {
...
+            applicationContext = this,
+            configuration = Configuration(
+                channel = BuildConfig.BUILD_TYPE,
+                httpClient = ConceptFetchHttpUploader(
+                    lazy(LazyThreadSafetyMode.NONE) { components.core.client }
+                )),
+            uploadEnabled = false
```
mdboom: I am concerned that this will cause the dropping of any pre-Glean-initialized events, of which there are a non-zero number currently in Fenix.  (I'm building now to get a current number of what they actually are and how many).  But given that playing them back was the source of the recent performance issue, I don't expect that number to be low.

Of course, given this move to initializing here, perhaps that will no longer be the case, but it would be good to confirm.  Passing in `false` here unconditionally sort of removes one of Glean's completeness guarantees.

Is there any way to get the migrated value for this from Fennec prior to this?

app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt
```diff
@@ -497,13 +498,6 @@ class GleanMetricsService(private val context: Context) : MetricsService {
         // can handle events being recorded before it's initialized.
         gleanInitializer = MainScope().launch {
             Glean.registerPings(Pings)
```
mdboom: Ideally, you would move this to before the new call site of `Glean.initialize` as well.  In practice, it currently only affects custom pings containing events, though, and I don't think Fenix has any custom pings containing events.

grigoryk:
Thank you for an excellent summary, @Dexterp37 . Let's go with the current solution for nightly, and revisit it later this week/next week as we continue getting this ready.

liuche:
Need to merge this so we can cut a Nightly - this is failing `multipleBookmarkDeletions` on Pixel 6 only, so I'm filing a bug to fix this.

PR #7763, opened 2020-01-16
Closes #5305: Update list of tabs after onboarding screen was dismissed.

### LINKED ISSUE #5305: Pin down flank.jar to stable version in Dockerfile
Per the recommendation of the flank team we always download the latest flank.jar to get bug fixes, etc.  The most recent flank version v7.0.2 has a breaking change so, going forward, we'd like pin down the version instead.

@pocmo is there a place we can specify a version number outside the Dockerfile (and pass that in as an env var) that way we don't have to push a new Docker image, should we wish to upgrade?

pocmo:
Fix for https://github.com/mozilla-mobile/android-components/issues/5305.

Whenever we dismiss the onboarding screen (by clicking on "Start browsing") we end up on the home screen with an empty list of tabs. Usually that is no problem. However if we migrated from Fennec then we may already have some tabs that we inherited from Fennec. Those tabs didn't show up until you did something that caused the home screen to reload (e.g. going to another screen and back).

This is a quick fix and updates the list of tabs when we dismiss the onboarding screen. Eventually we may just want to listen to `BrowserStore` and always display the tabs in it without keeping a separate copy that we only update sometimes.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
@@ -487,7 +487,9 @@ class HomeFragment : Fragment() {
     private fun hideOnboarding() {
         onboarding.finish()
         homeFragmentStore.dispatch(
-            HomeFragmentAction.ModeChange(currentMode.getCurrentMode()))
+            HomeFragmentAction.ModeChange(
+                mode = currentMode.getCurrentMode(),
+                tabs = getListOfSessions().toTabs()))
     }
 
     private fun setupHomeMenu() {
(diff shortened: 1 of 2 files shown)


boek:
@pocmo We only use then mode change in a couple of places to switch between private and normal. This should be good, but I think a safer fix that won’t require as much testing would be to default to a nullable list and only replace the parameter in the reducer if it exists.

I can also pull the branch down in a bit and see if I notice anything weird with it defaulting to an empty list :)

pocmo:
Well, I didn't want to change the behavior for all other cases. It was an empty list before.. it is an empty list now. :)

boek:
@pocmo true! In the onboarding case.

But https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt#L300 uses the same action, which with this change would replace the current tab list with an empty list. I still think the behavior is good. I’m mostly concerned with how it will look when you tap the private browsing mode icon and how that transition looks with this change.

pocmo:
@boek But the reducer always sets an empty list for this action?
https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/java/org/mozilla/fenix/home/HomeFragmentStore.kt#L86

boek:
🤦🏻‍♂️ ignore me. Someone must have fixed the bug I mentioned already!

:shipit:

PR #7778, opened 2020-01-16
Issue 7473 switch toolbar to viewstub!!

emalysz:
### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
+++ b/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
@@ -13,6 +13,7 @@ import androidx.annotation.CallSuper
 import androidx.annotation.IdRes
 import androidx.annotation.VisibleForTesting
 import androidx.annotation.VisibleForTesting.PROTECTED
+import androidx.appcompat.app.ActionBar
 import androidx.appcompat.widget.Toolbar
 import androidx.lifecycle.lifecycleScope
 import androidx.navigation.NavDestination
@@ -20,6 +21,7 @@ import androidx.navigation.NavDirections
 import androidx.navigation.fragment.NavHostFragment
 import androidx.navigation.ui.AppBarConfiguration
 import androidx.navigation.ui.NavigationUI
+import kotlinx.android.synthetic.main.activity_home.*
 import kotlinx.coroutines.launch
 import mozilla.components.browser.search.SearchEngine
 import mozilla.components.browser.session.Session
@@ -73,6 +75,8 @@ open class HomeActivity : LocaleAwareAppCompatActivity() {
 
     private val hotStartMonitor = HotStartPerformanceMonitor()
 
+    private var isToolbarInflated = false
+
     private val navHost by lazy {
         supportFragmentManager.findFragmentById(R.id.container) as NavHostFragment
     }
@@ -94,7 +98,8 @@ open class HomeActivity : LocaleAwareAppCompatActivity() {
         setupThemeAndBrowsingMode(getModeFromIntentOrLastKnown(intent))
         setContentView(R.layout.activity_home)
         Performance.instrumentColdStartupToHomescreenTime(this)
-        setupToolbarAndNavigation()
+
+        externalSourceIntentProcessors.any { it.process(intent, navHost.navController, this.intent) }
 
         if (intent.getBooleanExtra(EXTRA_FINISH_ONBOARDING, false)) {
             FenixOnboarding(this).finish()
@@ -220,17 +225,29 @@ open class HomeActivity : LocaleAwareAppCompatActivity() {
         themeManager.applyStatusBarTheme(this)
     }
 
-    private fun setupToolbarAndNavigation() {
+    /**
+     * Returns the [supportActionBar], inflating it if necessary.
+     * Everyone should call this instead of supportActionBar.
+     */
+    fun getSupportActionBarAndInflateIfNecessary(): ActionBar {
         // Add ids to this that we don't want to have a toolbar back button
-        val appBarConfiguration = AppBarConfiguration.Builder().build()
-        val navigationToolbar = findViewById<Toolbar>(R.id.navigationToolbar)
-        setSupportActionBar(navigationToolbar)
-        NavigationUI.setupWithNavController(navigationToolbar, navHost.navController, appBarConfiguration)
-        navigationToolbar.setNavigationOnClickListener {
-            onBackPressed()
-        }
+        if (!isToolbarInflated) {
+            val navigationToolbar = navigationToolbarStub.inflate() as Toolbar
 
-        externalSourceIntentProcessors.any { it.process(intent, navHost.navController, this.intent) }
+            setSupportActionBar(navigationToolbar)
+
+            NavigationUI.setupWithNavController(
+                navigationToolbar,
+                navHost.navController,
+                AppBarConfiguration.Builder().build()
+            )
+            navigationToolbar.setNavigationOnClickListener {
+                onBackPressed()
+            }
+
+            isToolbarInflated = true
+        }
+        return supportActionBar!!
     }
 
     protected open fun getIntentSessionId(intent: SafeIntent): String? = null
--- a/app/src/main/java/org/mozilla/fenix/ext/Fragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/ext/Fragment.kt
@@ -12,6 +12,8 @@ import androidx.navigation.NavDirections
 import androidx.navigation.NavOptions
 import androidx.navigation.Navigator
 import androidx.navigation.fragment.NavHostFragment.findNavController
+import kotlinx.android.synthetic.main.fragment_home.*
+import org.mozilla.fenix.HomeActivity
 import org.mozilla.fenix.components.Components
 
 /**
@@ -40,7 +42,7 @@ fun Fragment.getPreferenceKey(@StringRes resourceId: Int): String = getString(re
  */
 fun Fragment.showToolbar(title: String) {
     (requireActivity() as AppCompatActivity).title = title
-    (requireActivity() as AppCompatActivity).supportActionBar?.show()
+    (activity as HomeActivity).getSupportActionBarAndInflateIfNecessary().show()
 }
 
 /**
--- a/app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkFragment.kt
@@ -11,7 +11,6 @@ import android.view.MenuInflater
 import android.view.MenuItem
 import android.view.View
 import android.view.ViewGroup
-import androidx.appcompat.app.AppCompatActivity
 import androidx.fragment.app.activityViewModels
 import androidx.lifecycle.ViewModelProvider
 import androidx.lifecycle.lifecycleScope
@@ -37,6 +36,7 @@ import mozilla.components.concept.sync.AuthType
 import mozilla.components.concept.sync.OAuthAccount
 import mozilla.components.lib.state.ext.consumeFrom
 import mozilla.components.support.base.feature.UserInteractionHandler
+import org.mozilla.fenix.HomeActivity
 import org.mozilla.fenix.R
 import org.mozilla.fenix.components.FenixSnackbar
 import org.mozilla.fenix.components.StoreProvider
@@ -126,7 +126,7 @@ class BookmarkFragment : LibraryPageFragment<BookmarkNode>(), UserInteractionHan
     override fun onResume() {
         super.onResume()
 
-        (activity as? AppCompatActivity)?.supportActionBar?.show()
+        (activity as HomeActivity).getSupportActionBarAndInflateIfNecessary().show()
         context?.components?.backgroundServices?.accountManager?.let { accountManager ->
             sharedViewModel.observeAccountManager(accountManager, owner = this)
             accountManager.register(refreshOnSignInListener, owner = this)
--- a/app/src/main/java/org/mozilla/fenix/library/bookmarks/edit/EditBookmarkFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/library/bookmarks/edit/EditBookmarkFragment.kt
@@ -34,6 +34,7 @@ import mozilla.components.concept.storage.BookmarkNode
 import mozilla.components.concept.storage.BookmarkNodeType
 import mozilla.components.support.ktx.android.content.getColorFromAttr
 import mozilla.components.support.ktx.android.view.hideKeyboard
+import org.mozilla.fenix.HomeActivity
 import org.mozilla.fenix.R
 import org.mozilla.fenix.components.FenixSnackbar
 import org.mozilla.fenix.components.metrics.Event
@@ -113,14 +114,15 @@ class EditBookmarkFragment : Fragment(R.layout.fragment_edit_bookmark) {
 
     private fun initToolbar() {
         val activity = activity as? AppCompatActivity
-        val toolbar = activity?.findViewById<Toolbar>(R.id.navigationToolbar)
+        val actionBar = (activity as HomeActivity).getSupportActionBarAndInflateIfNecessary()
+        val toolbar = activity.findViewById<Toolbar>(R.id.navigationToolbar)
         context?.let {
             toolbar?.setToolbarColors(
                 foreground = it.getColorFromAttr(R.attr.primaryText),
                 background = it.getColorFromAttr(R.attr.foundation)
             )
         }
-        activity?.supportActionBar?.show()
+        actionBar.show()
     }
 
     override fun onPause() {
--- a/app/src/main/java/org/mozilla/fenix/settings/search/AddSearchEngineFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/search/AddSearchEngineFragment.kt
@@ -28,6 +28,7 @@ import kotlinx.coroutines.launch
 import kotlinx.coroutines.runBlocking
 import kotlinx.coroutines.withContext
 import mozilla.components.browser.search.SearchEngine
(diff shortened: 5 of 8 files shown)


app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -72,6 +73,10 @@ open class HomeActivity : LocaleAwareAppCompatActivity() {
 
     private val hotStartMonitor = HotStartPerformanceMonitor()
 
+    private var navigationToolbar: Toolbar? = null
+
```
mcomella: nit: rm whitespace: these should be grouped together.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -72,6 +73,10 @@ open class HomeActivity : LocaleAwareAppCompatActivity() {
...
+    private var toolbarHasInflated = false
```
mcomella: nit: java convention would be `isToolbarInflated`

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -218,17 +222,36 @@ open class HomeActivity : LocaleAwareAppCompatActivity() {
...
-    private fun setupToolbarAndNavigation() {
+    fun getSupportActionBarAndInflateIfNecessary(): Toolbar? {
         // Add ids to this that we don't want to have a toolbar back button
-        val appBarConfiguration = AppBarConfiguration.Builder().build()
-        val navigationToolbar = findViewById<Toolbar>(R.id.navigationToolbar)
-        setSupportActionBar(navigationToolbar)
-        NavigationUI.setupWithNavController(navigationToolbar, navHost.navController, appBarConfiguration)
-        navigationToolbar.setNavigationOnClickListener {
-            onBackPressed()
+        if (!this.toolbarHasInflated) {
```
mcomella: nit: we generally don't use `this` unless it's necessary or helps disambiguate.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -218,17 +222,36 @@ open class HomeActivity : LocaleAwareAppCompatActivity() {
...
+            }
+
+            externalSourceIntentProcessors.any {
+                it.process(
+                    intent,
+                    navHost.navController,
+                    this.intent
+                )
+            }
+            this.toolbarHasInflated = true
```
mcomella: nit: unnecessary `this`

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -218,17 +222,36 @@ open class HomeActivity : LocaleAwareAppCompatActivity() {
         themeManager.applyStatusBarTheme(this)
     }
 
-    private fun setupToolbarAndNavigation() {
+    fun getSupportActionBarAndInflateIfNecessary(): Toolbar? {
```
mcomella: nit: we should add some kdoc because why you might want to use this method is not clear. The first sentence in kdoc is always what the thing does (even if it's obvious) because it's used as a summary in the generated kdoc. The remaining sentences, in this case, can be used to say something like, "Callers trying to get the [supportActionBar] from the Activity should call this instead." and maybe a brief explanation why (b/c ViewStub).

Note kdoc uses 2 starts to start:
```kotlin
/**
 * kdoc block
 */
```

mcomella: This should always return a non-null Toolbar, right? Can we update the return type?

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -218,17 +222,36 @@ open class HomeActivity : LocaleAwareAppCompatActivity() {
...
         // Add ids to this that we don't want to have a toolbar back button
-        val appBarConfiguration = AppBarConfiguration.Builder().build()
-        val navigationToolbar = findViewById<Toolbar>(R.id.navigationToolbar)
-        setSupportActionBar(navigationToolbar)
-        NavigationUI.setupWithNavController(navigationToolbar, navHost.navController, appBarConfiguration)
-        navigationToolbar.setNavigationOnClickListener {
-            onBackPressed()
+        if (!this.toolbarHasInflated) {
+            val appBarConfiguration = AppBarConfiguration.Builder().build()
```
mcomella: super-nit: I think we can inline this function call: I don't see any value in leaving it out. It just makes this block more verbose.

app/src/main/java/org/mozilla/fenix/ext/Fragment.kt
```diff
@@ -39,6 +40,7 @@ fun Fragment.getPreferenceKey(@StringRes resourceId: Int): String = getString(re
  * Throws if the fragment is not attached to an [AppCompatActivity].
  */
 fun Fragment.showToolbar(title: String) {
+    (activity as? HomeActivity)?.getSupportActionBarAndInflateIfNecessary()
```
mcomella: Why not replace the `supportActionBar` in `supportActionBar.show()` line with this?

mcomella: nit: we shouldn't make a nullable cast: if HomeActivity is not available, this behavior will break and the app will enter an undefined state. We should just crash if the cast fails.

app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkFragment.kt
```diff
@@ -126,6 +127,7 @@ class BookmarkFragment : LibraryPageFragment<BookmarkNode>(), UserInteractionHan
     override fun onResume() {
         super.onResume()
 
+        (activity as? HomeActivity)?.getSupportActionBarAndInflateIfNecessary()
```
mcomella: Same notes as previous fragment.

app/src/main/java/org/mozilla/fenix/library/bookmarks/edit/EditBookmarkFragment.kt
```diff
@@ -135,6 +136,7 @@ class EditBookmarkFragment : Fragment(R.layout.fragment_edit_bookmark) {
 
     private fun initToolbar() {
         val activity = activity as? AppCompatActivity
+        (activity as? HomeActivity)?.getSupportActionBarAndInflateIfNecessary()
         val toolbar = activity?.findViewById<Toolbar>(R.id.navigationToolbar)
```
mcomella: We can just get `toolbar` from `getSupportActionBar...`, right?

I'm tempted to say we should rename `getSupportActionBar...` as `getToolbar` but then it won't be as discoverable. Maybe we should comment as much in the method kdoc. :)

app/src/main/java/org/mozilla/fenix/settings/search/AddSearchEngineFragment.kt
```diff
@@ -70,6 +71,7 @@ class AddSearchEngineFragment : Fragment(), CompoundButton.OnCheckedChangeListen
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
+        (activity as? HomeActivity)?.getSupportActionBarAndInflateIfNecessary()
```
mcomella: If we're doing a line-by-line replacement, should we put this in onResume when `show()` is called?

Then again, this is where "inflation" happens so maybe it's better to leave here. I'm not sure but I'm leaning towards the latter.

app/src/main/java/org/mozilla/fenix/settings/sitepermissions/SitePermissionsExceptionsFragment.kt
```diff
@@ -44,6 +45,7 @@ class SitePermissionsExceptionsFragment :
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
+        (activity as? HomeActivity)?.getSupportActionBarAndInflateIfNecessary()
```
mcomella: Same comments as others – we really don't want anyone calling for `supportActionBar`.

app/src/main/res/layout/activity_home.xml
```diff
@@ -9,16 +9,12 @@
     android:orientation="vertical"
     tools:context=".HomeActivity">
 
-    <androidx.appcompat.widget.Toolbar
-        android:id="@+id/navigationToolbar"
+    <ViewStub
+        android:id="@+id/navigationToolbar_stub"
+        android:inflatedId="@id/navigationToolbar_stub"
```
mcomella: nit: -> navigationToolbarStub

Side note: I don't love that the inflated ID is still called "stub" but I think it prevents people from accidentally using it, thinking it's the full toolbar.

app/src/main/res/layout/example.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
```
mcomella: Is this unused?

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -20,6 +20,7 @@ import androidx.navigation.NavDirections
 import androidx.navigation.fragment.NavHostFragment
```
mcomella: Did you file a bug for the Fenix team for the behavior where the BookmarksFragment shows "Firefox Preview" briefly before the proper title?

mcomella:
I got you access to the UI tests. One of the failures (`threeDotMenuItemsTest`) is:

>  androidx.test.espresso.NoMatchingViewException: No views in hierarchy found matching: (with text: is "Bookmarks" and has parent matching: with id: org.mozilla.fenix.debug:id/navigationToolbar) 

It's possible our async loading throws off the test because it doesn't wait long enough for the view to arrive. I see it's also using the old view ID.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
emalysz: https://github.com/mozilla-mobile/fenix/issues/7789

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -94,12 +99,19 @@ open class HomeActivity : LocaleAwareAppCompatActivity() {
         setupThemeAndBrowsingMode(getModeFromIntentOrLastKnown(intent))
         setContentView(R.layout.activity_home)
         Performance.instrumentColdStartupToHomescreenTime(this)
-        setupToolbarAndNavigation()
 
         if (intent.getBooleanExtra(EXTRA_FINISH_ONBOARDING, false)) {
             FenixOnboarding(this).finish()
         }
 
+        externalSourceIntentProcessors.any {
```
mcomella: nit: can this be put back on one-line? It's especially important if it's in a busy method like `onCreate`.

mcomella: This line used to occur above the `if` block above. I haven't looked into what this line does but I think it should still go above the `if` block.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -219,18 +231,34 @@ open class HomeActivity : LocaleAwareAppCompatActivity() {
         themeManager.setActivityTheme(this)
         themeManager.applyStatusBarTheme(this)
     }
-
-    private fun setupToolbarAndNavigation() {
+    /**
```
mcomella: nit: whitespace above here, between methods

app/src/main/res/layout/navigation_toolbar.xml
```diff
@@ -0,0 +1,16 @@
...
+<androidx.appcompat.widget.Toolbar
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    android:id="@+id/navigationToolbar"
+    android:layout_width="0dp"
+    android:layout_height="0dp"
+    app:titleMarginStart="16dp"
+    app:titleMarginEnd="16dp"
+    app:titleTextAppearance="@style/ToolbarTitleTextStyle"
+    android:background="?foundation"
```
mcomella: This background used to be `?above`: why is it `?foundation` now?

app/src/main/java/org/mozilla/fenix/settings/search/AddSearchEngineFragment.kt
```diff
@@ -70,6 +71,7 @@ class AddSearchEngineFragment : Fragment(), CompoundButton.OnCheckedChangeListen
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
+        (activity as HomeActivity).getSupportActionBarAndInflateIfNecessary()
```
mcomella: Why can't we put this on `supportActionBar.show()` like the others? Is it because it's called in `onResume` after which the view might already be laid out?

app/src/main/java/org/mozilla/fenix/library/bookmarks/edit/EditBookmarkFragment.kt
```diff
@@ -113,14 +114,15 @@ class EditBookmarkFragment : Fragment(R.layout.fragment_edit_bookmark) {
...
+        (activity as HomeActivity).getSupportActionBarAndInflateIfNecessary()
+        val toolbar = activity.findViewById<Toolbar>(R.id.navigationToolbar)
         context?.let {
             toolbar?.setToolbarColors(
                 foreground = it.getColorFromAttr(R.attr.primaryText),
                 background = it.getColorFromAttr(R.attr.foundation)
             )
         }
-        activity?.supportActionBar?.show()
+        activity.supportActionBar!!.show()
```
mcomella: It's my opinion that we should redirect all callers to use `getSupportActionBar...` instead of calling `activity.supportActionBar`. We should follow that behavior ourselves here too! We can store the actionBar in a local variable above and re-use it here.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -219,18 +231,34 @@ open class HomeActivity : LocaleAwareAppCompatActivity() {
...
-        val appBarConfiguration = AppBarConfiguration.Builder().build()
-        val navigationToolbar = findViewById<Toolbar>(R.id.navigationToolbar)
-        setSupportActionBar(navigationToolbar)
-        NavigationUI.setupWithNavController(navigationToolbar, navHost.navController, appBarConfiguration)
-        navigationToolbar.setNavigationOnClickListener {
-            onBackPressed()
-        }
+        if (!isToolbarInflated) {
+            val navigationToolbar = (navigationToolbarStub.inflate() as Toolbar).also {
+                this.navigationToolbar = it
```
mcomella: nit: we don't actually need this reference if we return `supportActionBar!!` instead.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -219,18 +231,34 @@ open class HomeActivity : LocaleAwareAppCompatActivity() {
...
-    private fun setupToolbarAndNavigation() {
+    /**
+     * Returns the inflated [supportActionBar].
+     * Callers trying to get the [supportActionBar] from the Activity should call this.
+     * Name chosen to be more discoverable and avoid confusion with references to getToolbar
+     * We initially use a ViewStub for startup performance benefits, and
+     * after inflation, it becomes a Toolbar.
+     */
+    fun getSupportActionBarAndInflateIfNecessary(): ActionBar {
         // Add ids to this that we don't want to have a toolbar back button
```
mcomella: nit: this comment should go closer to where it originally was (i.e. inside the if block). tbh, I don't know what this comment means (and it sounds outdated) so... yeah. 😬

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -219,18 +231,34 @@ open class HomeActivity : LocaleAwareAppCompatActivity() {
         themeManager.setActivityTheme(this)
         themeManager.applyStatusBarTheme(this)
     }
-
-    private fun setupToolbarAndNavigation() {
+    /**
+     * Returns the inflated [supportActionBar].
```
mcomella: nit: as we spoke about, more flowing. :) And probably don't need the perf note.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -73,6 +75,9 @@ open class HomeActivity : LocaleAwareAppCompatActivity() {
 
     private val hotStartMonitor = HotStartPerformanceMonitor()
 
+    private var navigationToolbar: Toolbar? = null
```
mcomella: nit: this is unused now

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -220,17 +226,29 @@ open class HomeActivity : LocaleAwareAppCompatActivity() {
         themeManager.applyStatusBarTheme(this)
     }
 
-    private fun setupToolbarAndNavigation() {
+    /**
+     * Returns the [supportActionBar], inflating it if necessary.
+     * Name chosen to be more discoverable and avoid confusion with references to getToolbar
```
mcomella: nit: I'd include a comment about, "Everyone should call this instead of `supportActionBar`."

For the line I highlighted, maybe something like, "We chose this name because it was more discoverable than `getToolbar`... (though I'm not sure we need this line, it actually makes sense that it's an action bar and it's confusing to explain why we wanted to call it toolbar anyway)

mcomella:
Some CI failures due to lint & unit test failures: reran it.

PR #7780, opened 2020-01-17
For perf-frontend-issues#50: instrument imperfect reportFullyDrawn.

mcomella:
We don't capture the second RV draw yet. Unfortunately, I don't know if
it's the best use of my time to capture that as well.

Random thoughts on `reportFullyDrawn`:
* It is expected to be called when content is drawn; i.e. it will not automatically attach it’s own onDrawListener for the next drawn frame.
  * Determined because calling it in onCreate has a big difference in results than calling it in a onPreDrawListener set in onCreate
* When instrumented correctly, it produces different results when launching via the homescreen icon and launching via the command line.
  * When launching with the homescreen icon, it will take longer than “Displayed” time and sometimes “Displayed” is not shown at all
  * When launching via the command line, “Displayed”
* Appears it can be used to measure warm start (i.e. swipe away app in recent app switcher); not sure about hot start

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/perf/Performance.kt
+++ b/app/src/main/java/org/mozilla/fenix/perf/Performance.kt
@@ -4,9 +4,37 @@
 
 package org.mozilla.fenix.perf
 
+import androidx.core.view.doOnPreDraw
+import kotlinx.android.synthetic.main.activity_home.*
+import org.mozilla.fenix.HomeActivity
+
 /**
  * A collection of objects related to app performance.
  */
 object Performance {
     const val TAG = "FenixPerf"
+
+    /**
+     * Instruments cold startup time for use with our internal measuring system, FNPRMS. This may
+     * also appear in Google Play Vitals dashboards.
+     *
+     * This will need to be rewritten if any parts of the UI are changed to be displayed
+     * asynchronously.
+     *
+     * In the current implementation, we only intend to instrument cold startup to the homescreen.
+     * To save implementation time, we ignore the fact that the RecyclerView draws twice if the user
+     * has tabs, collections, etc. open: the "No tabs" placeholder and a tab list. This
+     * instrumentation will only capture the "No tabs" draw.
+     */
+    fun instrumentColdStartupToHomescreenTime(activity: HomeActivity) {
+        // For greater accuracy, we could add an onDrawListener instead of a preDrawListener but:
+        // - single use onDrawListeners are not built-in and it's non-trivial to write one
+        // - the difference is timing is minimal (< 7ms on Pixel 2)
+        // - if we compare against another app using a preDrawListener, it should be comparable
+        //
+        // Unfortunately, this is tightly coupled to the root view of HomeActivity's view hierarchy
+        activity.rootContainer.doOnPreDraw {
+            activity.reportFullyDrawn()
+        }
+    }
 }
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mozilla/fenix/perf/Performance.kt
```diff
@@ -4,9 +4,37 @@
...
+     *
+     * In the current implementation, we only intend to instrument cold startup to the homescreen.
+     * To save implementation time, we ignore the fact that the RecyclerView draws twice if the user
+     * has tabs, collections, etc. open: the "No tabs" placeholder and a tab list. This
+     * instrumentation will only capture the "No tabs" draw.
+     */
+    fun instrumentColdStartupToHomescreenTime(activity: HomeActivity) {
+        // For greater accuracy, we could add an onDrawListener instead of a preDrawListener but:
+        // - single use onDrawListeners are not built-in and it's non-trivial to write one
+        // - the difference is timing is minimal (< 7ms on Pixel 2)
```
hawkinsw: uber nit: typo -- in not is.

app/src/main/java/org/mozilla/fenix/perf/Performance.kt
```diff
@@ -4,9 +4,37 @@
...
+        // - if we compare against another app using a preDrawListener, it should be comparable
+        //
+        // Unfortunately, this is tightly coupled to the root view of HomeActivity's view hierarchy
+        activity.rootContainer.doOnPreDraw {
+            activity.reportFullyDrawn()
```
hawkinsw: From my investigation, as best I can tell, the `Displayed` value is calculated based on the time to /complete/ the first draw. So, that might be the source of some of the difference that you mentioned in the notes to this PR.

hawkinsw: Can you tell me why we are adding the listener to the activity container here and not the recycler view that holds the session controller? I know that we are only concerned about the time that it takes when there are no sessions/tabs/etc open, but to capture even the drawing of those placeholder elements in that controller seems like part of the time that we want to capture in this metric.

I am sure that you know what you are doing which is why I want you to help me understand!

app/src/main/java/org/mozilla/fenix/perf/Performance.kt
mcomella: I'll address this in my next PR so we can just land this one without re-running CI.

app/src/main/java/org/mozilla/fenix/perf/Performance.kt
mcomella: > So, that might be the source of some of the difference that you mentioned in the notes to this PR.

Yep, I agree.

> Can you tell me why we are adding the listener to the activity container here and not the recycler view that holds the session controller?

Implementation simplicity. If all of the views are drawn at the same time, we can add a preDraw listener to any of the views and get the same result. However, this is fragile because, if the draw time of the view we add a listener to diverges from the views we care about, those listeners will no get called at the same time.

In this case, I chose to go with the simple implementation instead of the correct one because I figured if we needed `reportFullyDrawn` for real, we'd implement #7781 anyway.

mcomella: Let me know if that doesn't make sense!

app/src/main/java/org/mozilla/fenix/perf/Performance.kt
mcomella: There was no second PR so I made it: https://github.com/mozilla-mobile/fenix/pull/7848

PR #7592, opened 2020-01-10
Open "migration FAQ" when selecting "What's new" in a Fennec build..

pocmo:
This is the Fenix part of https://github.com/mozilla-mobile/android-components/issues/5481.

For Fennec builds we want to send users that click on "What's new?" to the migration FAQ page on SUMO.

PR #7787, opened 2020-01-17
AboutFragment: Use SupportUtils.getWhatsNewUrl().

pocmo:
In PR #7592 I missed that there is a second entry point to "What's new?" from the "about" screen.


pocmo:
@NotWoods We could. But I think once the migration is rolled out we will move back to only one what's new page anyways.

Issue #7636, opened 2020-01-13
[Bug]The page is no longer loaded when switching from ETP Strict to Standard and viceversa


softvision-sergiulogigan:
**Browser / Version**
Firefox Preview Nightly 200112 (🦎 74.0a1-20200110095023)

**Operating System**
Samsung Galaxy S8 (Android 9) - 1440 x 2960 pixels, 18.5:9 ratio (~570 ppi density)

**Steps to reproduce:**
1. Navigate to: https://mobile.reuters.com/
2. Tap the Shield button.
3. Tap the Protection Settings option.
4. Tap the Standard radio button.
5. Tap Back.

**Expected Behavior:**
The page is refreshed and loaded.

**Actual Behavior:**
The page is refreshed, but never fully loads.

**Watchers:**
@softvision-oana-arbuzov
@softvision-sergiulogigan
@cipriansv

Amejia481:
The same happens with other sites like cnn.com, it looks like the issue happens with sites that have trackers blocked

Amejia481:
This is `GeckoView` issue I opened a bug on Bugzilla for it. 
https://bugzilla.mozilla.org/show_bug.cgi?id=1609701

Amejia481:
I tested on the latest Nightly, it looks is fixed. Let's wait for QA verification to be sure.

![image](https://user-images.githubusercontent.com/773158/73760838-ecd13e80-473b-11ea-85b2-a417398c594a.png)

AndiAJ:
Hi, **verified as fixed** on the latest **Nightly Build #20360606 from 2/5** using the following devices:

• Google Pixel 3a XL (Android 10)
• Google Pixel 3a  (Android 9)
• Huawei Mate 20 Lite (Android 8.1.0)
• Samsung Galaxy S7 (Android 7.0.0)
• OnePlus A3 (Android 6.0.1)
• LG Nexus 4 (Android 5.1.1)

► Video
![20200205_091930](https://user-images.githubusercontent.com/51314259/73820059-86aaf100-47f9-11ea-90a0-32d3973629c5.gif)

PR #7815, opened 2020-01-20
For issue #7636: After changing a tracking protection setting make sure to reload the selected session only when the EngineView is visible.

Amejia481:
### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/settings/TrackingProtectionFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/TrackingProtectionFragment.kt
@@ -143,7 +143,7 @@ class TrackingProtectionFragment : PreferenceFragmentCompat() {
         context?.components?.let {
             val policy = it.core.createTrackingProtectionPolicy()
             it.useCases.settingsUseCases.updateTrackingProtection.invoke(policy)
-            it.useCases.sessionUseCases.reload.invoke()
+            context?.settings()?.shouldReload = true
         }
     }
 
--- a/app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
@@ -144,6 +144,7 @@ class BrowserFragment : BaseBrowserFragment(), UserInteractionHandler {
             (activity as HomeActivity).updateThemeForSession(it)
         }
         requireComponents.core.tabCollectionStorage.register(collectionStorageObserver, this)
+        reloadIfATrackingProtectionSettingHasChanged()
     }
 
     override fun onBackPressed(): Boolean {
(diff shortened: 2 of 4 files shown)


Amejia481:
You can find more context around this change here https://bugzilla.mozilla.org/show_bug.cgi?id=1609701#cr-7

Amejia481:
The issues are fixed now :)

ekager:
Two questions:
1. So the other sessions will just not reload/react to a TP change? The user will have to manually reload now?
2. If the user changes the TP setting, then closes their tab(s), the first time they reopen the app next time their first session will start loading and then immediately reload because of the flag, right? Are we okay with that?

Amejia481:
> 1.So the other sessions will just not reload/react to a TP change? The user will have to manually reload now?

At the moment, we can't not reload sessions that are not visible in the screen as will cause #7636. This patch will stop #7636 as a permanent [fix came out](https://bugzilla.mozilla.org/show_bug.cgi?id=1609701#cr-7).

> 2.If the user changes the TP setting, then closes their tab(s), the first time they reopen the app next time their first session will start loading and then immediately reload because of the flag, right? Are we okay with that?

I will try the approach that you suggested on slack **"Is it possible to check the current session's TP setting and compare it to what we want? We could put the check in the same place, I'm wondering if that would be cleaner than setting bool?"**

app/src/main/java/org/mozilla/fenix/settings/TrackingProtectionFragment.kt
```diff
@@ -143,7 +143,7 @@ class TrackingProtectionFragment : PreferenceFragmentCompat() {
         context?.components?.let {
             val policy = it.core.createTrackingProtectionPolicy()
             it.useCases.settingsUseCases.updateTrackingProtection.invoke(policy)
-            it.useCases.sessionUseCases.reload.invoke()
+            context?.settings()?.shouldReload = true
```
pocmo: I wonder why this is a setting (do we really need to persist it?) and why we could not hide this flag inside `GeckoEngineSession` - can we not handle this internally (maybe hiding some of it behind a usecase if needed)?

Amejia481: > I wonder why this is a setting (do we really need to persist it?)

Good catch, I think we could do it without persisting the setting, as if the app get closed the next time the user opens a session it will reload.

> Why we could not hide this flag inside GeckoEngineSession - can we not handle this internally (maybe hiding some of it behind a usecase if needed)?

I brought up the patch to momentarily stop the freezing until a more permanent solution is created from GV. They submitted a patch but I'm not quite sure if it will permanently fix the issue https://bugzilla.mozilla.org/show_bug.cgi?id=1609701#c11

Amejia481:
We are not going to need this patch as the issue is getting fixed directly on GeckoView https://bugzilla.mozilla.org/show_bug.cgi?id=1609701#c12

Issue #7820, opened 2020-01-21
Prefetch BrowsersCache for improved startup performance (26.36ms perfTest)


mcomella:
BrowsersCache.all takes 26.36ms to run (perfTest): we can pre-warm it on a background thread before it's called in `App.onCreate` to improve overall startup time.

Difficulty: 1

Impact:
- 26.36ms perfTest
- ?

mcomella:
I'm waiting to see if this goes away for free once we move things after visual completeness. I don't like the existing solution because it decouples the preWarm from the code so if the code is ever removed, we might be wasting time preWarming for nothing.

mcomella:
This issue isn't urgent or important and it's distracting on our board so I'm going to close it. We don't want to remove the code because it may have a performance improvement but we don't want to validate that because it'll take time to do so. I don't love leaving around complex code but it's fairly well isolated so it'll probably be fine... 😅

PR #7821, opened 2020-01-21
For #7820: warm BrowsersCache on background thread.

mcomella:
In an early iteration of his patch on a beta build, I saw no improvement over
6 runs (Pixel 2):
- Before: 401.5ms
- After: 402.17ms

This may be attributed to noise in startup performance. However,
BrowsersCache disappears from profiles completely and results in
theoretical performance gains.

When using the StartupTimeline class (not landed yet), I see a 27.75ms
runtime improvement on beta builds after this patch.

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
+++ b/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
@@ -35,6 +35,7 @@ import org.mozilla.fenix.components.Components
 import org.mozilla.fenix.ext.settings
 import org.mozilla.fenix.session.NotificationSessionObserver
 import org.mozilla.fenix.session.VisibilityLifecycleCallback
+import org.mozilla.fenix.utils.BrowsersCache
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -253,6 +255,14 @@ open class FenixApplication : LocaleAwareApplication() {
         }
     }
 
+    private fun warmBrowsersCache() {
+        // We avoid blocking the main thread for BrowsersCache on startup by loading it on
+        // background thread.
```
hawkinsw: Only real nit: We should say that "We try to avoid blocking" because this warmup method could still be running by the time that the app gets to first time that the `BrowsersCache` is needed. Not sure it's necessary to point that out, but perhaps it is? Up to you!

Issue #7822, opened 2020-01-21
Add StartupTimeline class to better divide & instrument startup


mcomella:
This will log to the console with startup durations to help us better understand startup.

mcomella:
This is not important for right now: moving to triage.

mcomella:
WIP https://github.com/mozilla-mobile/fenix/pull/7823

mcomella:
We're already doing this as part of the telemetry efforts: https://github.com/mozilla-mobile/fenix/issues/8803

PR #7823, opened 2020-01-21
For #7822: add StartupTimeline

mcomella:
### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/perf/StartupTimeline.kt
+++ b/app/src/main/java/org/mozilla/fenix/perf/StartupTimeline.kt
@@ -1,0 +1,108 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.perf
+
+import android.util.Log
+import androidx.annotation.UiThread
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.launch
+import java.util.EnumMap
+import java.util.concurrent.TimeUnit
+
+private const val TAG = Performance.TAG
+
+/**
+ * Instruments points along the startup code path and logs these metrics.
+ *
+ * Currently, logged values are only guaranteed to be valid during cold startup to homescreen.
+ */
+@UiThread // to minimize overhead and simplify call patterns.
+object StartupTimeline {
+
+    enum class Method {
+        APP_ON_CREATE,
+        HOME_ACTIVITY_ON_CREATE,
+        HOME_ACTIVITY_ON_RESUME,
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/mozilla/fenix/perf/StartupTimeline.kt
```diff
@@ -0,0 +1,108 @@
...
+        HOME_FRAGMENT_ON_CREATE_VIEW,
+        HOME_FRAGMENT_ON_VIEW_CREATED,
+        HOME_FRAGMENT_ON_START,
+        HOME_FRAGMENT_ON_RESUME
+    }
+
+    private val methodStart = EnumMap<Method, Long>(Method::class.java)
+    private val methodEnd = EnumMap<Method, Long>(Method::class.java)
+
+    private var isColdStartDumpComplete = false
```
hawkinsw: We could use this flag that other way around:

1. Call it something like `isColdStart` and default to `true`
1. Only execute the main operations of `dumpInstrumentation()` when `isColdStart` is `true`.
1. Before completion of `dumpInstrumentation`, set `isColdStart` to `false`. 

In that case, when we are executing the dump, we know that the logged values are all non-null and that could simplify things? I think that this approach will work given that the lifetime of `object`s is tied to the lifetime of the application. Therefore, this flag will not get reset unless the entire application is restarted, exactly what we want. 

That's just an idea I had while thinking to myself.

app/src/main/java/org/mozilla/fenix/perf/StartupTimeline.kt
```diff
@@ -0,0 +1,108 @@
...
+            // class is intended to be used for specific tests, it's fine.
+            //
+            // Here and below: methodStart/End values may be null in startup scenarios that aren't
+            // cold startup to homescreen so we don't assert them to be non-null.
+            val totalElapsedMillis = methodStart[Method.APP_ON_CREATE]?.let { first ->
+                methodEnd[Method.HOME_FRAGMENT_ON_RESUME]?.minus(first)
+            }?.let { TimeUnit.NANOSECONDS.toMillis(it) }
+
+            val totalRecordedMillis = methods.fold(0L) { acc, method ->
+                val durationMillis = getDurationNanos(method)?.let { TimeUnit.NANOSECONDS.toMillis(it) }
```
hawkinsw: What value does `it` get when there are no valid metrics for `method`? I am sure that you are handling this case, I am just curious.

app/src/main/java/org/mozilla/fenix/perf/StartupTimeline.kt
```diff
@@ -0,0 +1,108 @@
...
+                acc + (durationMillis ?: 0L)
+            }
+
+            val methodLines = enumValues<Method>().joinToString(
```
hawkinsw: We used an `enum class` so that we could easily get the "stringified" version of the constants for printing?

mcomella:
Not a focus right now so closing to avoid polluting open PR list.

Issue #7698, opened 2020-01-14
[Bug] Back button should animate to the left before search screen disappears


sblatz:
## Steps to reproduce
1. Open the search screen
2. Press back

### Expected behavior
1. The back button animates to the left (off of the screen) while the rest of the screen transitions

### Actual behavior
1. The back button stays in place while the rest of the screen transitions and then flickers out.

### Device information
* Android device: Galaxy S9 Android 9
* Fenix version: 1/14/20 nightly

sblatz:
Updated animation:

![back_button_animation](https://user-images.githubusercontent.com/4400286/72846363-496d2d80-3c55-11ea-9ebb-b62d6232f8a6.gif)

sv-sdeiac:
Verified as fixed using the latest version of Firefox Preview Nightly 1/28 #20280606.
Devices:
- Nokia 6 (Android 7.1.1);
- OnePlus 5T (Android 9).

PR #7840, opened 2020-01-21
For #7698: Adds search back button animation

sblatz:
### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/search/SearchController.kt
+++ b/app/src/main/java/org/mozilla/fenix/search/SearchController.kt
@@ -76,6 +76,7 @@ class DefaultSearchController(
     }
 
     override fun handleEditingCancelled() {
+        store.dispatch(SearchFragmentAction.UpdateEditingCanceled)
         navController.navigateUp()
     }
 
--- a/app/src/main/java/org/mozilla/fenix/search/SearchFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/search/SearchFragment.kt
@@ -17,6 +17,7 @@ import android.view.ViewGroup
 import android.view.ViewStub
 import androidx.appcompat.app.AlertDialog
 import androidx.core.view.isVisible
+import androidx.core.view.marginStart
 import androidx.fragment.app.Fragment
 import androidx.interpolator.view.animation.FastOutSlowInInterpolator
 import androidx.navigation.fragment.findNavController
@@ -112,7 +113,8 @@ class SearchFragment : Fragment(), UserInteractionHandler {
                     showBookmarkSuggestions = requireContext().settings().shouldShowBookmarkSuggestions,
                     session = session,
                     pastedText = pastedText,
-                    searchAccessPoint = searchAccessPoint
+                    searchAccessPoint = searchAccessPoint,
+                    isAnimatingOut = false
                 )
             )
         }
@@ -133,8 +135,7 @@ class SearchFragment : Fragment(), UserInteractionHandler {
             view.toolbar_component_wrapper,
             searchInteractor,
             historyStorageProvider(),
-            (activity as HomeActivity).browsingModeManager.mode.isPrivate,
-            ::animateBackButtonAway
+            (activity as HomeActivity).browsingModeManager.mode.isPrivate
         )
 
         val urlView = toolbarView.view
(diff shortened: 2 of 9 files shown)


app/src/main/java/org/mozilla/fenix/search/SearchFragment.kt
```diff
@@ -352,7 +357,17 @@ class SearchFragment : Fragment(), UserInteractionHandler {
         }
     }
 
+    private fun animateBackButtonAway() {
+        val xTranslation =
+            requireView().back_button.width.toFloat() * BACK_BUTTON_ANIMATION_WIDTH_MULTIPLIER
+
+        requireView().back_button
+            .animate()
+            .translationX(xTranslation).interpolator = FastOutSlowInInterpolator()
```
Baron-Severin: Nit: would you mind line breaking before `.interpolator...`?

app/src/main/java/org/mozilla/fenix/search/SearchFragment.kt
```diff
@@ -352,7 +357,17 @@ class SearchFragment : Fragment(), UserInteractionHandler {
         }
     }
 
+    private fun animateBackButtonAway() {
+        val xTranslation =
+            requireView().back_button.width.toFloat() * BACK_BUTTON_ANIMATION_WIDTH_MULTIPLIER
```
Baron-Severin: It looks like this constant is to make up for padding + margin?  That works, but it's a little bit messy, opens us up to weird bugs if we ever adjust either of those values, and requires a reader to figure out what it's handling.  What about something like this:

```kotlin
val backButton = requireView().back_button
val xTranslation = with(backButton) {
    -(width + marginStart + paddingStart).toFloat()
}
```

app/src/main/java/org/mozilla/fenix/search/SearchFragment.kt
```diff
@@ -198,8 +200,11 @@ class SearchFragment : Fragment(), UserInteractionHandler {
             qrFeature.get()?.scan(R.id.container)
         }
 
-        view.back_button.setOnClickListener {
-            findNavController().navigateUp()
+        view.back_button.apply {
+            setOnClickListener {
+                animateBackButtonAway()
+                findNavController().navigateUp()
```
Baron-Severin: I think it's a bad sign that this animates then navigates up, while the ToolbarView calls the interactor, which calls the controller, which... animates then navigates up.  This is one of the things our architecture is meant to unify, so let's clean it up while we're here.

If we:
1) move `animateBackButtonAway` off of `ToolbarView` and on to `SearchController`
2) replace 205 & 206 in `SearchFragment` with `searchInteractor.onEditingCanceled()`

then we have a more reasonable and consistent data flow.

It seems to me that we want back press and back button clicks to behave the same here, so I'm happy to have them call the same interactor method.  If that ever changes in the future, we can break them out as needed.

app/src/main/java/org/mozilla/fenix/search/SearchFragment.kt
sblatz: Seems reasonable to me!

app/src/main/java/org/mozilla/fenix/search/SearchController.kt
```diff
@@ -76,6 +76,7 @@ class DefaultSearchController(
     }
 
     override fun handleEditingCancelled() {
+        store.dispatch(SearchFragmentAction.EditingCancelled)
```
Baron-Severin: Nit: there's nothing wrong with this name, but just to align with AC I'd prefer to name it `UpdateEditingCanceled`.

app/src/main/java/org/mozilla/fenix/search/SearchFragment.kt
```diff
@@ -352,6 +363,18 @@ class SearchFragment : Fragment(), UserInteractionHandler {
         }
     }
 
+    private fun animateBackButtonAway() {
+        val backButton = requireView().back_button
+        val xTranslation = with(backButton) {
+            -(width + marginStart + paddingStart).toFloat()
+        }
+
+        requireView().back_button
```
Baron-Severin: Nit: you can replace this line with `backButton`

Issue #7762, opened 2020-01-16
Glean initialization follow-up


pocmo:
Follow-up from #7717.

See: https://github.com/mozilla-mobile/fenix/pull/7717#pullrequestreview-343354760

We may want to get back to a state where Glean can buffer events pre-initialization (but also not over buffer like before) and work nicely with the migration that determines the state of telemetry "a bit later".

Dexterp37:
> We may want to get back to a state where Glean can buffer events pre-initialization (but also not over buffer like before) and work nicely with the migration that determines the state of telemetry "a bit later".

The ideal state would be the following:

- Load the correct telemetry setting
- Initialize Glean (which means calling `Glean.initialize` not `GleanMetricsService.start()`!) as soon as possible with that value.

If that's not possible, we should still try to call `Glean.initialize` ASAP, and not too late during startup. We risk, otherwise, of overflowing the pre-init buffer due to components (e.g. GV) who send lots of metrics.

liuche:
Feel free to close if there is nothing that QA can verify.

AndiAJ:
Hi, **verified as fixed** on **Nightly Migration Build # 5** using a Google Pixel 3a XL (Android 10). 
The Fenix "client_id" values are generated and displayed properly. (No new values are assigned after force stopping and re-opening Fenix).

[Logcat](https://github.com/mozilla-mobile/fenix/files/4117529/B5.txt)
[Glean dashboard](https://debug-ping-preview.firebaseapp.com/pings/aj-test2701000)

Dexterp37:
> 
> 
> Hi, **verified as fixed** on **Nightly Migration Build # 5** using a Google Pixel 3a XL (Android 10).
> The Fenix "client_id" values are generated and displayed properly. (No new values are assigned after force stopping and re-opening Fenix).
> 
> [Logcat](https://github.com/mozilla-mobile/fenix/files/4117529/B5.txt)
> [Glean dashboard](https://debug-ping-preview.firebaseapp.com/pings/aj-test2701000)

Does this result in a fresh client id in older builds?

AndiAJ:
> Does this result in a fresh client id in older builds?

Yes, for instance, on Migration Build # 4 a new "client_id" was assigned after force stopping and resuming Fenix.

Dexterp37:
> 
> 
> > Does this result in a fresh client id in older builds?
> 
[...]

PR #7857, opened 2020-01-22
Issue #7762: Enable glean based on setting for non-fennec builds..

pocmo:
This is not the final patch for #7762, but a quick-fix that we can ship now to unblock some of the Gecko metrics collection.

For Fenix builds we will initialize Glean immediately based on the setting. This will unblock Gecko metrics collection happening for our Fenix builds. For Fennec builds we keep the behavior of disabling Glean first and then potentially re-enabling after migrating the setting. This still leaves us with the one case to be fixed: Non first run Fennec with no migration needed. In this situation we still may enable Glean too late. I am going to look into fixing this too - but there are currently more important migration issues to tackle and those builds basically do not exist in the wild for now anyways.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
+++ b/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
@@ -59,17 +59,21 @@ open class FenixApplication : LocaleAwareApplication() {
             return
         }
 
-        // We need to always initialize Glean and do it early here. Note that we are disabling it
-        // here too (uploadEnabled = false). If needed Glean will be enabled later by the migration
-        // code (if this user used to be a fennec user with the right flags enabled) or by
-        // GleanMetricsService if telemetry is enabled for this user.
+        // We need to always initialize Glean and do it early here.
         // It is important that this initialization happens *here* before calling into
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -59,25 +59,27 @@ open class FenixApplication : LocaleAwareApplication() {
...
-        // potentially enable Glean would be lost. However such pings do not exist at this moment.
-        logger.debug("Initializing Glean (uploadEnabled=false)")
+        val enableGlean = if (Config.channel.isFennec) {
+            // We are disabling Glean here because for Fennec builds we may not know yet whether
+            // we can enable telemetry yet. We first need to migrate the setting from Fennec to
+            // know the user's choice. The blocking migration in `MigratingFenixApplication` will
+            // notify glean once the value has been migrated.
+            false
+        } else {
+            settings().isTelemetryEnabled
```
Dexterp37: nit: can we add a comment stating that it's important to know this at this time otherwise we might end up loosing data for components that collect telemetry very early?

Issue #6704, opened 2019-11-20
Enable FxA state encryption for release


grigoryk:
https://github.com/mozilla-mobile/fenix/issues/5938 enabled account state encryption for nightly users.

Things to watch out for in Sentry:
- instances of `AbnormalFxaEvent`, specifically `MissingExpectedAccountAfterStartup`
- instances of `AbnormalAccountStorageEvent@UnexpectedlyMissingAccountState`

If we don't see a spike in these after a few weeks, let's consider flipping encrypted storage for the release channels.

### Acceptance Criteria
- [x] Look on Sentry and verify that there are not a lot of the highlighted issues. If there are, file and prioritize them.
- [ ] Enable FxA state encryption on release

grigoryk:
Note that there's currently _some_ volume of `MissingExpectedAccountAfterStartup`. I expect those to drop to zero once https://github.com/mozilla-mobile/android-components/pull/5103 makes its way to nightly users.

grigoryk:
There was a non-zero volume of Android KeyStore related crashes in Sentry. This needs more investigation.

liuche:
Still seeing [a lot of crashes](https://crash-stats.mozilla.org/signature/?product=Fenix&signature=%5BINFO%5D%20org.mozilla.fenix.components.AbnormalFxaEvent%24MissingExpectedAccountAfterStartup%3A%20at%20org.mozilla.fenix.components.AccountAbnormalities%24accountManagerInitializedAsync%241.invokeSuspend%28AccountAbnormalities.kt%3A18%29&version=74.0a1&date=%3C2020-01-18T03%3A59%3A41%2B00%3A00&date=%3E%3D2020-01-11T03%3A59%3A41%2B00%3A00) for `AbnormalFxaEvent$MissingExpectedAccountAfterStartup`

ValentinTimisica:
Based on this [comment](https://github.com/mozilla-mobile/fenix/blob/e25ee98d7de743fe4ac86eaf5ef3d5284383462c/app/src/main/java/org/mozilla/fenix/components/BackgroundServices.kt#L76) I understand that FxA state encryption needs to be enabled for all channels. So I made a [PR](https://github.com/mozilla-mobile/fenix/pull/7858) where I set ```secureStateAtRest = true```. Is this correct?
If it needs to be enabled just for Nightly and Release I will change it accordingly.

ekager:
[...]

PR #7858, opened 2020-01-22
For #6704: Enable FxA state encryption for all channels 

ValentinTimisica:
### Pull Request checklist

- [x] **Tests**: This PR does not include thorough tests.
- [x] **Screenshots**: This PR does not include screenshots or GIFs.

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/BackgroundServices.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/BackgroundServices.kt
@@ -31,7 +31,6 @@ import mozilla.components.service.fxa.manager.SCOPE_SYNC
 import mozilla.components.service.fxa.sync.GlobalSyncableStoreProvider
 import mozilla.components.service.sync.logins.SyncableLoginsStore
 import mozilla.components.support.base.log.logger.Logger
-import org.mozilla.fenix.Config
 import org.mozilla.fenix.FeatureFlags
 import org.mozilla.fenix.R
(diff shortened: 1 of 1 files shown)


liuche:
@grigoryk ^ so we can get this into 3.3 and we have a chance to see this in the wild before Beta Migration.

ekager:
@grigoryk so can we land this and I'll monitor Sentry for the next few days for any weirdness?

grigoryk:
To complete the loop: we've talked about this on Slack; there are a number of low-volume, device-specific issues associated with keystore access, but we think they will be mostly addressed by https://github.com/mozilla-mobile/android-components/pull/5796

The only remaining bit of due diligence that I think we should do is to make sure that our worst-case scenario is graceful. Loosing data may be inevitable due to the nature of these issues, but at least let's make sure we won't crash in case of keystore-related exceptions, and submit them to Sentry as `info` level exceptions instead to keep track of their volume.

Depending on what key material was affected, a corrupt keystore may mean either:
- loosing our FxA state - in which case, we need to make sure that account manager is transferred into a "Logged out" state
- loosing access to data in the login storage, in which case we need to make sure that logins-related functionality continues to function (can create new logins, autofill, etc).

ekager:
Closing this because we've decided to wait a bit longer to enable for all channels see #8324

ekager:
Ugh meant to close the PR not the issue, sorry

PR #7894, opened 2020-01-23
Closes #7762: Initialize Glean later in Fennec builds....

### LINKED ISSUE #7762: Glean initialization follow-up

pocmo:
PR #7857 was a quick fix that unblocked telemetry collection from `fenix` builds and fixed the telemetry ID reset. This is now fixing the same issues for `fennec` builds.

We are now avoiding setting `uploadEnabled` to `false` in Glean by splitting the initialization: For `fenix` builds we do it immediately and set the flag based on the user setting. For `fennec` builds we do it a bit later once we have migrated the setting from Fennec to honor the user's choice.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/Config.kt
+++ b/app/src/main/java/org/mozilla/fenix/Config.kt
@@ -57,6 +57,9 @@ enum class ReleaseChannel {
 
     val isFennec: Boolean
         get() = this in fennecChannels
+
+    val isFenix: Boolean
+        get() = !isFennec
 }
 
 object Config {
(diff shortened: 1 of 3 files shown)


pocmo:
Tested this with the migration build and telemetry on/off in Fennec before migration.

PR #7901, opened 2020-01-23
No Issue: Create a post-visual completeness executor

hawkinsw:
Create an object that will execute its enqueued, paused tasks when Fenix is visually complete.

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt
@@ -5,6 +5,7 @@
 package org.mozilla.fenix.components.metrics
 
 import android.content.Context
+import kotlinx.coroutines.CoroutineStart
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.MainScope
@@ -44,6 +45,7 @@ import org.mozilla.fenix.GleanMetrics.UserSpecifiedSearchEngines
 import org.mozilla.fenix.ext.components
 import org.mozilla.fenix.ext.settings
 import org.mozilla.fenix.utils.BrowsersCache
+import org.mozilla.fenix.utils.OnVisualCompletenessExecutionQueue
 
 private class EventWrapper<T : Enum<T>>(
     private val recorder: ((Map<T, String>?) -> Unit),
@@ -490,16 +492,20 @@ class GleanMetricsService(private val context: Context) : MetricsService {
         // because it calls Google ad APIs that must be called *off* of the main thread.
         // These two things actually happen in parallel, but that should be ok because Glean
         // can handle events being recorded before it's initialized.
-        gleanInitializer = MainScope().launch {
+        gleanInitializer = MainScope().launch(start = CoroutineStart.LAZY) {
             Glean.registerPings(Pings)
         }
+
         // setStartupMetrics is not a fast function. It does not need to be done before we can consider
         // ourselves initialized. So, let's do it, well, later.
-        gleanSetStartupMetrics = MainScope().launch {
+        gleanSetStartupMetrics = MainScope().launch(start = CoroutineStart.LAZY) {
             setStartupMetrics()
         }
 
         context.settings().totalUriCount = 0
+
+        OnVisualCompletenessExecutionQueue.add(gleanInitializer)
+        OnVisualCompletenessExecutionQueue.add(gleanSetStartupMetrics)
     }
 
     internal fun setStartupMetrics() {
--- a/app/src/main/java/org/mozilla/fenix/utils/OnVisualCompletenessExecutionQueue.kt
+++ b/app/src/main/java/org/mozilla/fenix/utils/OnVisualCompletenessExecutionQueue.kt
@@ -1,0 +1,76 @@
+package org.mozilla.fenix.utils
+
+import kotlinx.coroutines.Job
+import mozilla.components.support.base.log.Log
+
+/**
+ * A queue of paused `Job`s whose execution is resumed when Fenix is visually complete.
+ *
+ * This queue contains paused `Job`s. Each Job in the queue will be restarted once Fenix
+ * is visually complete (as long as it has not be cancelled since it was added to the queue).
+ * Each Job will execute on the thread on which it was originally scheduled.
+ *
+ * This is threadsafe.
+ */
+object OnVisualCompletenessExecutionQueue {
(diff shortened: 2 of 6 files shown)


app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt
```diff
@@ -490,16 +492,20 @@ class GleanMetricsService(private val context: Context) : MetricsService {
...
         // ourselves initialized. So, let's do it, well, later.
-        gleanSetStartupMetrics = MainScope().launch {
+        gleanSetStartupMetrics = MainScope().launch(start = CoroutineStart.LAZY) {
             setStartupMetrics()
         }
 
         context.settings().totalUriCount = 0
+
+        OnVisualCompletenessExecutionQueue.add(gleanInitializer)
+        OnVisualCompletenessExecutionQueue.add(gleanSetStartupMetrics)
```
pocmo: There is a hidden assumption here that we will always show UI and run through `HomeActivity`. But this is not true. A `FenixApplication` object will be created for every process that is launched by the app. We may launch the main process without UI (e.g. some background job runs) and we may launch other processes that have other UI (e.g. the crash reporter) and we may launch into the app and not go into the HomeActivity (not sure if those exist at the moment though). In those cases we would never run those methods. Is that a problem?

app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt
```diff
@@ -490,16 +492,20 @@ class GleanMetricsService(private val context: Context) : MetricsService {
         // because it calls Google ad APIs that must be called *off* of the main thread.
         // These two things actually happen in parallel, but that should be ok because Glean
         // can handle events being recorded before it's initialized.
-        gleanInitializer = MainScope().launch {
+        gleanInitializer = MainScope().launch(start = CoroutineStart.LAZY) {
```
Dexterp37: How much would this delay the call to `Glean.registerPings(Pings)`? Would this still be guaranteed to be called before [this](https://github.com/mozilla-mobile/fenix/blob/0199363577dc654562d9f6c7d7d087ed8fde0014/app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt#L541)? If not, then the `activation` ping will not be sent.

app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt
Dexterp37: > In those cases we would never run those methods. Is that a problem?

Yes - if the `registerPings` is not called, then Glean won't know about Fenix custom pings. If the metrics are not set, then they won't end up in any ping.

hawkinsw: > There is a hidden assumption here that we will always show UI and run through `HomeActivity`. But this is not true. A `FenixApplication` object will be created for every process that is launched by the app. We may launch the main process without UI (e.g. some background job runs) and we may launch other processes that have other UI (e.g. the crash reporter) and we may launch into the app and not go into the HomeActivity (not sure if those exist at the moment though). In those cases we would never run those methods. Is that a problem?

This is a very good point. I traced through the intent-handling code and it looks like any way that a user can launch the app (that's key to this statement) will go through the HomeActivity in some way. E.g., even when the app is cold started with an intent to view a URL it still goes through the HomeActivity.

That said, it never occurred to me that we would really want Glean metrics in the other cases you mentioned, @pocmo , so I appreciate you bringing them to my attention. I think that it would be great to have a way to distinguish the cases where there will be a UI loaded and where there will not be a UI loaded. In the former case we would delay these operations until visual completeness and the others we would do them immediately. I suppose it would be possible to draw this distinction by examining (the presence or lack of) intents during FenixApplication creation. What do you think, @pocmo ?

app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt
hawkinsw: Yes, that method would be called later it would still be called in the same order as they are now. In other words, the code in the Job of the `gleanInitializer()` will execute before the code in the Job of the `setStartupMetrics()`. Does that make sense?

hawkinsw: Of course, if what you are saying is a requirement, then doing what we are currently doing in Fenix is not safe. It is possible that the scheduler swaps these two Jobs and executes `setStartupMetrics` before `gleanInitializer` because we do not specifically order them.

app/src/main/java/org/mozilla/fenix/utils/OnVisualCompletenessExecutionQueue.kt
```diff
@@ -0,0 +1,76 @@
+package org.mozilla.fenix.utils
```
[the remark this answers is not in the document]
hawkinsw: Dang! Completely forgot. Thanks!

app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt
Dexterp37: Nope

> 
> 
> Of course, if what you are saying is a requirement, then doing what we are currently doing in Fenix is not safe. It is possible that the scheduler swaps these two Jobs and executes `setStartupMetrics` before `gleanInitializer` because we do not specifically order them.

This part should be fine. I'm concerned about Glean.registerPings which should be called before [this](https://github.com/mozilla-mobile/fenix/blob/0199363577dc654562d9f6c7d7d087ed8fde0014/app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt#L541)

mcomella:
I haven't read through the recent comments but I'm concerned that the init code will not be executed if HomeActivity is not shown (e.g. if we restore from OOM to a different Activity) because we hard-code to run this in `HomeActivity`: https://github.com/mozilla-mobile/fenix/pull/7901/files#diff-44fc84568dc7e79b2cb8921c2135183aR120

mcomella:
If it's not already written here, I would suggest we separate the visual completeness queue patch from making changes to Glean for simplicity.

mcomella:
Close in favor of https://github.com/mozilla-mobile/fenix/pull/8610

PR #7904, opened 2020-01-23
Bump a-s dependency to 0.48.3.

vladikoff:
Changelog: https://github.com/mozilla/application-services/releases/tag/v0.48.3

Ref: mozilla-mobile/android-components#5703

@grigoryk @ekager r?

### DIFF
--- a/buildSrc/src/main/java/Dependencies.kt
+++ b/buildSrc/src/main/java/Dependencies.kt
@@ -39,7 +39,7 @@ object Versions {
     // that we depend on directly for the fenix-megazord (and for it's
     // forUnitTest variant), and it's important that it be kept in
     // sync with the version used by android-components above.
-    const val mozilla_appservices = "0.48.2"
+    const val mozilla_appservices = "0.48.3"
 
     const val mozilla_glean = "23.0.0"
 


liuche:
Quick question @vladikoff, how should the timing of AS version bumps work, to avoid "megazord version mismatch" problems, and Nightly startup crashes? By the time this AS version bump PR is opened against, is the bump already propagating through the AC snapshot, so we should try to land it ASAP? Or do we need to wait (for something?) and then trigger a Fenix Nightly as soon as it lands?

Trying to avoid Nightly startup crashes from this again.

grigoryk:
@liuche once the corresponding A-C PR lands, we can trigger an A-C snapshot rebuild, and once the snapshots are updated the Fenix PR can land. Assumption I'm making here is that most people who work on Fenix use snapshot builds of a-c - as opposed to building against local a-c checkouts. If it's mostly the latter though, we can land Fenix PR as soon as the a-c PR hits master, without waiting for the snapshots to be rebuild.

In either case, let's address the need to worry about any of this soon, maybe while we're all together next week.

grigoryk:
@vladikoff it's not clear to me if we need a data review for this or not. A-S release docs mention that this release is collecting new metrics (`places` perf/quality), so I'm assuming that we need a product level data review?

grigoryk:
I am going to assume that we don't need a separate data review here, since one was already done for the library itself.

grigoryk:
Yeah, don't a data review, as per https://mail.mozilla.org/pipermail/fx-data-dev/2020-January/000313.html:
```

Things that do not require Data Collection Review:

1. A Product upgrading an integrated Component to a new version that has
new data collections. (This is covered by (1) above. The Product could be
included in the review by name, or as a product that embeds $MyComponent.
If clarification is desired, we can amend the data collection review to
specifically include the Product by name. No biggie.)
```

pocmo:
AC snapshot ready.

Issue #7524, opened 2020-01-07
[Investigate] Abnormally high device-commands traffic


grigoryk:
There are concerns from @rfk that Fenix is generating an abnormally high level of device-commands traffic (e.g. checking for send tabs). This issue is to investigate what's going on.

#### Acceptance Criteria
- [ ] Communicate with AS to determine how frequently we should be syncing (update these ACs with result)
- [ ] Ensure that we are only syncing as frequently as expected

jonalmeida:
@grigoryk Where my knowledge is lacking: are the "device-commands traffic" the same as if someone is pressing the "sync now" button?

If so, it might just be this one call in [`onResume`](https://github.com/mozilla-mobile/fenix/blob/9c28cb632c3fc2e980315ca6ee9cb01460f7b0f8/app/src/main/java/org/mozilla/fenix/HomeActivity.kt#L126). We had this here originally because send tab landed before push, and for usability we sync'd every time. We could remove it now.

rfk:
>  are the "device-commands traffic" the same as if someone is pressing the "sync now" button?

I believe it's part of what happens when you press "sync now", although other stuff also happens that is not related to device-commands.

+1 to trying to remove the `onResume` trigger and seeing how much of a difference that makes.

rfk:
The fix in #7908 definitely seems to have helped with the traffic load here. It's still significantly higher than what we see on other clients, but about half what it used to be.

I don't have any other theories for why the rate should be higher on Fenix than on other clients. AFAICT there are three ways that Fenix could trigger a request to `GET /v1/account/device/commands`:

* By receiving and handling a "command received" push message.
* By the user clicking the "sync now" button, of which we [do not have very many](https://sql.telemetry.mozilla.org/queries/68007). 
* By invoking the `postAuthenticated` handler on the `FxaAccountManager`.

IIUC the `postAuthenticated` handler may be invoked on application startup thanks to the `Event.AccountRestored` event, though I'm not 100% sure I follow the logic there completely. If we're polling for commands on every startup that might explain the higher volume?

rfk:
> By invoking the postAuthenticated handler on the FxaAccountManager.

[...]

PR #7908, opened 2020-01-24
For #7524: Remove DeviceConstellation polling from on Resume!! 

jonalmeida:
@grigoryk @rfk let's start with just removing the device constellation polling, and if we see the numbers fall, we can consider continuing with removing/moving the `accountManager.syncNowAsync` as well.


rfk:
Thanks! How can I tell when this is live in Fenix nightly, to check whether it had the desired effect of reducing polling traffic.

jonalmeida:
@rfk this looks like it was merged 3 days ago and there was a build today, so the code change should be in. :)

rfk:
It's hard to get a precise read on the change in behaviour here, but this as definitely helped! On newer versions of Fenix I am seeing about half the number of device-commands requests per active user as we were seeing with older versions.

Issue #7818, opened 2020-01-21
Replace IntentReceiverActivity with MigrationDecisionActivity on cold startup from homescreen to improve perf


mcomella:
When clicking the homescreen icon, `IntentReceiverActivity` is opened and redirects the user to `HomeActivity`. This is a recent behavior change for migration and may be indicative of a regression.

`IntentReceiverActivity` seems like it also opens applink URLs.

`IntentReceiverActivity` seems expensive for several reasons:
- `IntentReceiverActivity.processIntent` is expensive (81.69ms)
- Running through the Activity lifecycle is expensive (creating, starting, resuming, pausing, and stopping one Activity only to start another one)
  - 32.37ms: performLaunchActivity
  - 5.24ms: handleStart
  - 17.15ms: performResumeActivity
  - 54.63ms: performTraversals (for a frame never drawn)
  - 14.53ms: executeLifecycleState (for pause)
  - Total: 123.92ms
  - Excludes starting HomeActivity (which, to be fair, some of the lifecycle work done above may not happen here again)

### Priority
**Difficulty** (1-5): 3? 4? Requires coordination with other teams, complex flow tightly coupled with app launch from many configurations

**Impact** (Pixel 2):
  - `forPerformanceTest`: ~150ms+ ?
  - production: ?

### Questions
- Why is IntentReceiverActivity in the startup path at all? HomeActivity is most commonly used so can it redirect to other activities as necessary?

Work with Sebastian and Jonathan, who originally landed these changes, on if it can change.

jonalmeida:
A reference issue for evaluating the multiple entry points into the app.

mcomella:
To summarize the less prescriptive problem statement in comment 0, for this issue my understanding is that the primary problem is that we’re routing all of our Intents through a separate, headless activity and then delegating to the Activity that actually needs to handle them. This additional Activity adds non-trivial overhead and maybe perceived performance issues when clicking on the homescreen icon. My naive solution is to route everything through HomeActivity - which handles most of our requests anyway - and delegate from there as necessary.

pocmo:
@mcomella What does the question mark behind 150ms mean? Are there more measurements required? I wonder how the full startup compares to the previous state, not looking at this activity in isolation.

Some background on why this change was made:

The migration has two requirements:
- The data migration needs an exclusive lock on the database. We can't let other code access the database since it will likely deadlock / ANR.
- We need to show a modal migration UI for all app entry points while a migration is ongoing.

[...]

PR #7917, opened 2020-01-24
Issue #7818: Use different target activities for launcher activity alias (fenix vs. fennec).

pocmo:
(Proposal for regression described in #7818. See issue for more details).

* With this patch fenix build variants will launch HomeActivity directly from the launcher.
  This will get rid of the performance regression outlined in #7818.

* Fennec build variants will launch into MigrationDecisionActivity which will show the
  migration UI or launch normally into the app. This is faster than going through
  IntentReceiverActivity (measured and perceived).

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -31,16 +31,14 @@
 
         <!--
           We inherited this entry (${applicationId}.App) from Fennec. We need to keep this as our
-          main launcher to avoid launcher icons on the home screen disappearing for all our users
-          on a Fennec build.
+          main launcher to avoid launcher icons on the home screen disappearing for all our users.
 
-          !!HERE BE DRAGONS!!
-          We need the targetActivity to be IntentReceiverActivity so that we can check if we have
-          a migration in progress. If not, the HomeActivity is invoked.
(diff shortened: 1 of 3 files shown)


jonalmeida:
From the issue:
> However going to the white screen directly feels faster. So then I tried to have my router activity have a white background and switch to the home/migration activity immediately without an animation. 

Should we also remove the transparent theming from the IntentReceiverActivity as well them?

pocmo:
> Should we also remove the transparent theming from the IntentReceiverActivity as well them?

Possibly. But I think `IntentReceiverActivity` itself deserves its own issue. There are multiple things we could try.

mcomella:
@pocmo One question: does this draw the right background color for dark mode users? iirc, under some configurations, they may get a black screen. The two variables to check are system theming and Fenix's preference.

pocmo:
> @pocmo One question: does this draw the right background color for dark mode users? iirc, under some configurations, they may get a black screen. The two variables to check are system theming and Fenix's preference.

I did check that it behaves like `HomeActivity` - which unfortunately means that both are showing a brief white screen at the moment. This is something to address too - but since it affects `HomeActivity` too - maybe in another bug.

PR #8037, opened 2020-02-02
Base for Venetan language

conky77:
I wrote to mozilla-dev-l10n-new-locales@lists.mozilla.org to propose a new Venetan localization for Fenix, code ISO 639-3: VEC ( https://en.wikipedia.org/wiki/Venetan_language ). 
I propose this file as base for the Venetan translation.

### DIFF
--- a/app/src/main/res/values-vec/strings.xml
+++ b/app/src/main/res/values-vec/strings.xml
@@ -1,0 +1,1158 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+
+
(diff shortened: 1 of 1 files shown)


flodolo:
Sorry, this is not going to work. Localization happens through Pontoon and lands in a different repository, synced with this one. 

You should be able to upload your file in Pontoon once the project has been enabled.

cc @Delphine

Pike:
We're all set up on Pontoon, closing this.

Issue #5695, opened 2019-10-01
Add "Search" to the text selection context menu


vesta0:
### User Story

As a user, I want to be able to search a selected text through the context menu, so I can get to my search results in as few steps as possible.

### Deps
- [x] GV bug https://bugzilla.mozilla.org/show_bug.cgi?id=1581161

### Requirements

-Add "search" to the context menu (@jpfaller to provide exact string)
-Add telemetry probe to measure taps on "search"

### Acceptance Criteria

- [ ] When I select a text on a webpage, I see a "search" option in the context menu (this is called "Google Search" in Fennec (see screen shot) and "Web Search" in Chrome
- [ ] When I search for the selected text, a new tab opens with my search results
- [ ] When I search for the selected text in private mode, a new private tab opens with my search results

@jpfaller please note that this issue is different than adding a "search in Fenix" entry to the 3rd party browser context menu, similar to Focus, Opera, DuckDuckGo

![1028964090](https://user-images.githubusercontent.com/42752644/65935867-98fac580-e3e8-11e9-9e0f-a2240d38116a.png)

vesta0:
@lime124 I propose that we skip UX for this issue and refer to the Fennec UI if needed.

lime124:
totally fine to reuse Fennec IxD but we should revisit the "Google Search" string. UR has been showing some conflation in people's mental models around the browser and the search engine. I think this is one spot we can help by reinforcing the  app they are in vs the search engine. @jpfaller for help with this. thanks!

liuche:
This context menu is in GeckoView (because it's in web content), and we'd be able to use the [BasicSelectionActionDelegate](https://mozilla.github.io/geckoview/javadoc/mozilla-central/org/mozilla/geckoview/BasicSelectionActionDelegate.html) to override the actions created.

We might need to do this in AC.

liuche:
Passing the action to SearchFragment to handle the Search action makes this a little trickier.

ekager:
FYI to whomever picks this up. Some changes may be happening to the SelectionActionDelegate https://bugzilla.mozilla.org/show_bug.cgi?id=1581161

jpfaller:
Let's make the strings:

Firefox Search
Firefox Private Search

These options should be available when a user selects text OS wide.

We should monitor complaints to see if users are frustrated by this option being available without a way to turn it off.

vesta0:
[...]

PR #8156, opened 2020-02-06
Fix for #5695 - Migration steps' status displayed before the migration completed 

Sparky93:
The migration steps' status was already being shown as succeeded instead of showing an "unchecked" list of steps.
These changes will make the steps being "checked" only after the migration completed by starting
with an "unchecked" list.

- [x] **Tests**: This is already covered by tests.
- [x] **Screenshots**: This PR isn't adding any new visual changes just behavioral.
- [x] **Accessibility**: The code in this PR does not include any **new** user facing features.

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/migration/MigrationProgressActivity.kt
+++ b/app/src/main/java/org/mozilla/fenix/migration/MigrationProgressActivity.kt
@@ -102,7 +102,13 @@ internal fun MigrationResults.toItemList() = filterKeys {
         type,
         status.success
     )
-}
+}.toMutableList()
+    .plus(
+        whiteList
+            .filterKeys { !this.containsKey(it) }
+            .keys
+            .map { MigrationItem(it, false) }
+    ).sortedBy { it.migration.javaClass.simpleName }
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/fenix/migration/MigrationProgressActivity.kt
```diff
@@ -95,12 +95,12 @@ internal val whiteList = mapOf(
...
 )
 
-internal fun MigrationResults.toItemList() = filterKeys {
+internal fun MigrationResults.toItemList(migrationCompleted: Boolean = false) = filterKeys {
     whiteList.keys.contains(it)
 }.map { (type, status) ->
     MigrationItem(
         type,
-        status.success
+        if (migrationCompleted) status.success else false
```
pocmo: I don't think this is the intention here. We DO want to show the state of every item as it is completed. However right now we only show any items at all if they are completed. Because the state from the store does only include completed items. Since we already have a whitelist here we could just use this to "fill up" the list for all items that are not completed. Or we add the state for incompleted migration items from AC.

PR #8226, opened 2020-02-07
Remove Tracking protection popup for visual testing

MarcLeclair:
### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/perf/Performance.kt
+++ b/app/src/main/java/org/mozilla/fenix/perf/Performance.kt
@@ -20,7 +20,7 @@ import org.mozilla.fenix.utils.Settings
  */
 object Performance {
     const val TAG = "FenixPerf"
-    const val EXTRA_IS_PERFORMANCE_TEST = "performancetest"
+    private const val EXTRA_IS_PERFORMANCE_TEST = "performancetest"
 
     /**
      * Instruments cold startup time for use with our internal measuring system, FNPRMS. This may
@@ -51,19 +51,48 @@ object Performance {
      * on) and removes the onboarding screen.
      */
     fun processIntentIfPerformanceTest(intent: Intent, context: Context) {
-        if (!intent.getBooleanExtra(EXTRA_IS_PERFORMANCE_TEST, true)) {
+        if (!isPerformanceTest(intent, context)) {
             return
         }
 
-        val batteryStatus = context.registerReceiver(null, IntentFilter(Intent.ACTION_BATTERY_CHANGED))
-        val isPhonePlugged = batteryStatus!!.getIntExtra(BatteryManager.EXTRA_PLUGGED, -1) ==
-                BatteryManager.BATTERY_PLUGGED_USB
-        val isAdbEnabled = AndroidSettings.Global.getInt(context.contentResolver,
-            AndroidSettings.Global.ADB_ENABLED, 0)
+        disableOnboarding(context)
+        disableTrackingProtectionPopups(context)
+    }
 
-        if (isPhonePlugged && (isAdbEnabled == 1)) {
-            FenixOnboarding(context).finish()
-            Settings.getInstance(context).isOverrideTPPopupsForPerformanceTest = true
+    /**
+     * The checks for the USB connections and ADB debugging are checks in case another application
+     * tries to leverage this intent to trigger a code path for Firefox that shouldn't be used unless
+     * it is for testing visual metrics. These checks aren't full proof but most of our users won't have
+     * ADB on and USB connected at the same time when running Firefox.
+     */
+    fun isPerformanceTest(intent: Intent, context: Context): Boolean {
+        if (!intent.getBooleanExtra(EXTRA_IS_PERFORMANCE_TEST, false)) {
+            return false
         }
+        val batteryStatus = context.registerReceiver(null, IntentFilter(Intent.ACTION_BATTERY_CHANGED))
+        batteryStatus?.let {
+            val isPhonePlugged = it.getIntExtra(BatteryManager.EXTRA_PLUGGED, -1) ==
+                    BatteryManager.BATTERY_PLUGGED_USB
+            val isAdbEnabled = AndroidSettings.Global.getInt(
+                context.contentResolver,
+                AndroidSettings.Global.ADB_ENABLED, 0
+            ) == 1
+            return isPhonePlugged && isAdbEnabled
+        }
+        return false
+    }
+
+    /**
+     * Bypasses the onboarding screen on launch
+     */
+    fun disableOnboarding(context: Context) {
+        FenixOnboarding(context).finish()
+    }
+
+    /**
+     * Disables the tracking protection popup. However, TP is still on.
+     */
+    fun disableTrackingProtectionPopups(context: Context) {
+        Settings.getInstance(context).isOverrideTPPopupsForPerformanceTest = true
     }
 }
--- a/app/src/main/java/org/mozilla/fenix/perf/Performance.kt
+++ b/app/src/main/java/org/mozilla/fenix/perf/Performance.kt
@@ -4,15 +4,23 @@
 
 package org.mozilla.fenix.perf
 
+import android.content.Context
+import android.content.Intent
+import android.content.IntentFilter
+import android.os.BatteryManager
 import androidx.core.view.doOnPreDraw
 import kotlinx.android.synthetic.main.activity_home.*
 import org.mozilla.fenix.HomeActivity
+import org.mozilla.fenix.onboarding.FenixOnboarding
+import android.provider.Settings as AndroidSettings
+import org.mozilla.fenix.utils.Settings
 
 /**
  * A collection of objects related to app performance.
  */
 object Performance {
     const val TAG = "FenixPerf"
+    const val EXTRA_IS_PERFORMANCE_TEST = "performancetest"
 
     /**
      * Instruments cold startup time for use with our internal measuring system, FNPRMS. This may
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mozilla/fenix/perf/Performance.kt
```diff
@@ -37,4 +44,22 @@ object Performance {
             activity.reportFullyDrawn()
         }
     }
+
+    fun processIntentIfPerformanceTest(intent: Intent, context: Context) {
+
+        if (intent.getBooleanExtra(HomeActivity.EXTRA_IS_PERFORMANCE_TEST, false)) {
+            val batteryStatus = context.registerReceiver(null, IntentFilter(Intent.ACTION_BATTERY_CHANGED))
+            val isPhonePlugged = batteryStatus!!.getIntExtra(BatteryManager.EXTRA_PLUGGED, -1)
```
mcomella: It's entirely possible batteryStatus returns null: can we handle the null case? Normally, I'd say wave it because this is test code but it's possible another app can trigger this maliciously so I think we should handle it.

mcomella: super-nit: I think it's more readable if we do the boolean evaluation up here instead of in the `if`:
```kotlin
val isPhonePluggedViaUSB = batteryStatus!!.getIntExtra(BatteryManager.EXTRA_PLUGGED, -1)
    == BatteryManager.BATTERY_PLUGGED_USB
```

(or whatever the recommended styling is). Here and below.

app/src/main/java/org/mozilla/fenix/perf/Performance.kt
```diff
@@ -37,4 +44,22 @@ object Performance {
...
+            val isAdbenabled = android.provider.Settings.Global.getInt(context.getContentResolver(),
```
mcomella: nit: `isAdbEnabled`

mcomella: nit: you can import this like:
```kotlin
import android.provider.Settings as AndroidSettings
```

To avoid using the fully qualified name.

app/src/main/java/org/mozilla/fenix/perf/Performance.kt
```diff
@@ -37,4 +44,22 @@ object Performance {
...
+                                                android.provider.Settings.Global.ADB_ENABLED, 0)
+
+            if ((isPhonePlugged == BatteryManager.BATTERY_PLUGGED_USB) && (isAdbenabled == 1)) {
+                FenixOnboarding(context).finish()
+
+                for (i in 0..Settings.trackingProtectionOnboardingMaximumCount) {
```
mcomella: nit: it'd be clearer to encapsulate this in a function called something like `disableTrackingProtectionPopups` and we should add a note on why we're doing this.

It's not very self explanatory to understand what is happening and why when seeing just a for loop with an imperactive operation inside.

mcomella: Rather than incrementing here, can we just add an override directly into Settings? It'd reduce the impact on startup time and be clearer.

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -48,6 +48,7 @@ class Settings private constructor(
         const val trackingProtectionOnboardingMaximumCount = 2
         const val FENIX_PREFERENCES = "fenix_preferences"
 
+        var overrideTPPopupsForPerformanceTest = false
```
mcomella: afaict, this probably shouldn't be in the companion object: static state is generally hard to reason about and this object has no precedence for managing such state.

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -135,7 +136,7 @@ class Settings private constructor(
 
     val shouldShowTrackingProtectionOnboarding: Boolean
         get() = trackingProtectionOnboardingCount < trackingProtectionOnboardingMaximumCount &&
-                !trackingProtectionOnboardingShownThisSession
+                !trackingProtectionOnboardingShownThisSession && !overrideTPPopupsForPerformanceTest
```
mcomella: If feel like this would be clearer if we separated these statements with OR: at a high-level, either we're overriding (so don't show onboarding) OR we're checking the conditions for showing this session.

app/src/main/java/org/mozilla/fenix/perf/Performance.kt
```diff
@@ -4,15 +4,23 @@
...
+import org.mozilla.fenix.onboarding.FenixOnboarding
+import android.provider.Settings as AndroidSettings
+import org.mozilla.fenix.utils.Settings
 
 /**
  * A collection of objects related to app performance.
  */
 object Performance {
     const val TAG = "FenixPerf"
+    const val EXTRA_IS_PERFORMANCE_TEST = "performancetest"
```
mcomella: nit: private

app/src/main/java/org/mozilla/fenix/perf/Performance.kt
```diff
@@ -37,4 +45,25 @@ object Performance {
             activity.reportFullyDrawn()
         }
     }
+
+    /**
+     * Processes intent for Performance testing to remove protection pop up ( but keeps the TP
```
mcomella: nit: in hindsight, explaining this comment at a higher level in code instead would make this is clearer and less likely to get out of date:
```kotlin
fun processIntentIfPerformanceTest(intent: Intent, context: Context) {
    if (!isPerformanceTest()) { return }

    disableOnboarding()
    disableTrackingProtectionPopups()
}

File: app/src/main/java/org/mozilla/fenix/perf/Performance.kt [hunk 8]
```diff
@@ -37,4 +45,25 @@ object Performance {
...
+     * on) and removes the onboarding screen.
+     */
+    fun processIntentIfPerformanceTest(intent: Intent, context: Context) {
+        if (!intent.getBooleanExtra(EXTRA_IS_PERFORMANCE_TEST, true)) {
+            return
+        }
+
+        val batteryStatus = context.registerReceiver(null, IntentFilter(Intent.ACTION_BATTERY_CHANGED))
+        val isPhonePlugged = batteryStatus!!.getIntExtra(BatteryManager.EXTRA_PLUGGED, -1) ==
```
mcomella: Can we handle the case if this is null? Other apps could add this extra to trigger this intent so we should make it handled gracefully.

app/src/main/java/org/mozilla/fenix/perf/Performance.kt
```diff
@@ -37,4 +45,25 @@ object Performance {
...
+                BatteryManager.BATTERY_PLUGGED_USB
+        val isAdbEnabled = AndroidSettings.Global.getInt(context.contentResolver,
```
mcomella: Can we do the `==` here too, making this property a boolean?

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -133,9 +133,12 @@ class Settings private constructor(
 
     private var trackingProtectionOnboardingShownThisSession = false
```
mcomella: nit: remove this newline

app/src/main/java/org/mozilla/fenix/perf/Performance.kt
```diff
@@ -37,4 +45,25 @@ object Performance {
             activity.reportFullyDrawn()
         }
     }
+
+    /**
+     * Processes intent for Performance testing to remove protection pop up ( but keeps the TP
+     * on) and removes the onboarding screen.
+     */
+    fun processIntentIfPerformanceTest(intent: Intent, context: Context) {
+        if (!intent.getBooleanExtra(EXTRA_IS_PERFORMANCE_TEST, true)) {
```
mcomella: I think this got inverted in the last PR: this should be false, right? Because it's inverted by `!`

app/src/main/java/org/mozilla/fenix/perf/Performance.kt
```diff
@@ -37,4 +45,59 @@ object Performance {
             activity.reportFullyDrawn()
         }
     }
+
+    /**
+     * Processes intent for Performance testing to remove protection pop up ( but keeps the TP
+     * on) and removes the onboarding screen.
+     */
+    fun processIntentIfPerformanceTest(intent: Intent, context: Context) {
+        if (isPerformanceTest(intent)) {
```
mcomella: nit: !isPerformanceTest

app/src/main/java/org/mozilla/fenix/perf/Performance.kt
```diff
@@ -37,4 +45,59 @@ object Performance {
...
+            return
+        }
+
+        if (evaluatePhoneStatus(context)) {
```
mcomella: Why not include this as part of `isPerformanceTest`? e.g. `if (isValidPerformanceTest)...`

app/src/main/java/org/mozilla/fenix/perf/Performance.kt
```diff
@@ -37,4 +45,59 @@ object Performance {
...
+            disableOnboarding(context)
+            disableTrackingProtectionPopups(context)
+        }
+    }
+
+    /**
+     * Checks if the performance Intent was received for performance testing.
```
mcomella: nit: kind of redundant to the code in this case – I'd remove it

app/src/main/java/org/mozilla/fenix/perf/Performance.kt
```diff
@@ -37,4 +45,59 @@ object Performance {
...
+     */
+    fun isPerformanceTest(intent: Intent): Boolean {
+        return !intent.getBooleanExtra(EXTRA_IS_PERFORMANCE_TEST, false)
```
mcomella: This shouldn't be !, right?

app/src/main/java/org/mozilla/fenix/perf/Performance.kt
```diff
@@ -37,4 +45,59 @@ object Performance {
...
+    }
+
+    /**
+     * The test will be allowed to run if the phone is connected through usb and if it has
```
mcomella: This doesn't really describe what this method does – it describes how it's being used, which you should generally try to avoid because it's not colocated with the code that calls it and is very likely to get outdated. In general, I'm not sure this comment is needed.

app/src/main/java/org/mozilla/fenix/perf/Performance.kt
MarcLeclair: I kept the intent with the !, so I didn't add it here, which in hindsight isn't quite obvious

MarcLeclair:
> lgtm – thanks Marc!

Woohoo I got through it! 🎉

PR #8261, opened 2020-02-10
Update Mozilla Android Components to 32.0.0-SNAPSHOT.

pocmo:


pocmo:
🤔

I'll try to reproduce locally.

pocmo:
Updated the two parts in the code to use `viewLifecycleOwner`. Not sure why this comes up now though.

pocmo:
Alright, passes now. Using the view lifecycle is definitely the right call for the livedata observers. Landing.

PR #8265, opened 2020-02-10
Import l10n.

Pike:
superceding #8185 to trigger automation.

CC @flod, @JohanLorenzo, @Callek, @tomprince  .

We should find an automated way to run tests here :-/


JohanLorenzo:
Weird, I see you're part of a subteam of the `mozilla-mobile` org. Let me try on a PR I create.

PR #8294, opened 2020-02-11
No Issue: Create a post-visual completeness executor

emalysz:
Create an object that will execute its enqueued, paused tasks
when Fenix is visually complete.

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
+++ b/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
@@ -14,12 +14,14 @@ import androidx.annotation.IdRes
 import androidx.annotation.VisibleForTesting
 import androidx.annotation.VisibleForTesting.PROTECTED
 import androidx.appcompat.widget.Toolbar
+import androidx.core.view.doOnPreDraw
 import androidx.lifecycle.lifecycleScope
 import androidx.navigation.NavDestination
 import androidx.navigation.NavDirections
 import androidx.navigation.fragment.NavHostFragment
 import androidx.navigation.ui.AppBarConfiguration
 import androidx.navigation.ui.NavigationUI
+import kotlinx.android.synthetic.main.activity_home.*
 import kotlinx.coroutines.launch
 import mozilla.components.browser.search.SearchEngine
 import mozilla.components.browser.session.Session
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/mozilla/fenix/utils/OnVisualCompletenessExecutionQueue.kt
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * A queue of paused `Job`s whose execution is resumed when Fenix is visually complete.
+ *
+ * This queue contains paused `Job`s. Each Job in the queue will be restarted once Fenix
+ * is visually complete (as long as it has not be cancelled since it was added to the queue).
+ * Each Job will execute on the thread on which it was originally scheduled.
+ *
+ * This is threadsafe.
+ */
+object OnVisualCompletenessExecutionQueue {
```
mcomella: I spoke to you on Slack but I think this is more complex than we need: I think we really only need to hard-code methods to call like:
```kotlin
fun onVisualCompleteness() {
    GleanMetricsService.start()
    LeanplumMetricsService.start()
    // ...
}
```

I'm concerned about adding something more complicated than we need because it's harder to maintain, harder to make changes to, and more likely to be hiding bugs.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -110,6 +114,12 @@ open class HomeActivity : LocaleAwareAppCompatActivity() {
                 ?.also { components.analytics.metrics.track(Event.OpenedApp(it)) }
         }
         supportActionBar?.hide()
+
+        // Just before the root container is drawn for the first time, consider Fenix to be visually
+        // complete.
+        rootContainer.doOnPreDraw {
+            this.asContext.application.onApplicationVisuallyComplete()
```
mcomella: There are two issues with the current implementation:
- this is not visual completeness: we're attaching a pre-draw listener in `HomeActivity.onCreate`, before any other frames are drawn, so the first onPreDraw call will be for the first frame drawn
- If we don't launch `HomeActivity` (e.g. the framework has removed us from memory and we reopen the app on the Settings screen), our visually complete jobs will never run. We need to be resilient to the app being opened from any entry point

fwiw, solving these problems is the real challenge of this issue.

PR #8401, opened 2020-02-13
Dismiss "migration notification" after leaving migration activity..

pocmo:
For https://github.com/mozilla-mobile/android-components/issues/5886.

With this patch we will dismiss the "migration completed" notification once the users
clicks on the button in the migration activity.

Needs https://github.com/mozilla-mobile/android-components/pull/5927 in AC before it can land.


pocmo:
> Are we going to dismiss the notification when tapping on it, as well?

Ah, let me check that.

pocmo:
No, currently we do that. I'll prepare a patch, but that will have to land in AC.

pocmo:
Follow-up in AC: https://github.com/mozilla-mobile/android-components/pull/5935

PR #8402, opened 2020-02-13
For #8143: Allow add-on popups to display prompts

### LINKED ISSUE #8143: [Bug]uBlock settings freeze after tapping the "Reset to default settings" button
## Prerequisites
Have uBlock Origin installed.

## Steps to reproduce
1. From the Add-ons manager tap on uBlock Origins
2. Tap and open "Settings"
3. Swipe down Settings section tap the  "Reset to default settings" button

### Expected behavior
A confirmation prompt asking the user to confirm the reset of Ublock to factory settings should be displayed.
After the user confirms the reset, he should be able to access the other Settings sections (e.g. "File lists, "My filters" etc.)  

### Actual behavior
After tapping the button you can't access the other Settings sections (e.g. "File lists, "My filters" etc.)  

### Device information

* Android device: 
• Google Pixel 3a XL (Android 10)
• Google Pixel 3a (Android 9)
• Huawei Mate 20 Lite (Android 8.1.0)
• Samsung Galaxy S7 (Android 7.0.0)
• OnePlus A3 (Android 6.0.1)
• LG Nexus 4 (Android 5.1.1)

* Fenix version: Nightly #20370604 from 2/6

### Notes
<details>
    <summary>Video comparison between Fennec VS Fenix (dropdown)</summary>

► Fenix

►Fennec

</details>

csadilek:
This is a refactoring to introduce a base class for our add-on popup fragments. The base class now also brings in the `PromptFeature` to fix #8143.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/addons/AddonPopupBaseFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/addons/AddonPopupBaseFragment.kt
@@ -1,0 +1,106 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.addons
+
+import android.os.Bundle
+import android.view.View
+import androidx.fragment.app.Fragment
+import androidx.navigation.fragment.NavHostFragment.findNavController
+import androidx.navigation.fragment.findNavController
+import mozilla.components.browser.state.action.ContentAction
+import mozilla.components.browser.state.action.CustomTabListAction
+import mozilla.components.browser.state.state.CustomTabSessionState
+import mozilla.components.browser.state.state.EngineState
+import mozilla.components.browser.state.state.SessionState
+import mozilla.components.browser.state.state.createCustomTab
+import mozilla.components.concept.engine.EngineSession
+import mozilla.components.concept.engine.prompt.PromptRequest
+import mozilla.components.concept.engine.window.WindowRequest
+import mozilla.components.feature.prompts.PromptFeature
+import mozilla.components.support.base.feature.ViewBoundFeatureWrapper
+import org.mozilla.fenix.ext.requireComponents
+
+/**
+ * Provides shared functionality to our fragments for add-on settings and
+ * browser/page action popups.
+ */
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mozilla/fenix/addons/AddonPopupBaseFragment.kt
```diff
@@ -0,0 +1,105 @@
...
+                    session.id,
+                    promptRequest
+                )
+            )
+        }
+    }
+
+    override fun onWindowRequest(windowRequest: WindowRequest) {
+        if (windowRequest.type == WindowRequest.Type.CLOSE) {
+            activity?.onBackPressed()
```
psymoon: I think `findNavController().popBackStack()` is better here to keep it consistent.

app/src/main/java/org/mozilla/fenix/addons/WebExtensionActionPopupFragment.kt
```diff
@@ -44,7 +41,7 @@ class WebExtensionActionPopupFragment : Fragment(), EngineSession.Observer {
     ): View? {
         // Grab the [EngineSession] from the store when the view is created if it is available.
         if (engineSession == null) {
-            engineSession = coreComponents.store.state.extensions[args.webExtensionId]?.popupSession
+            createSession(coreComponents.store.state.extensions[args.webExtensionId]?.popupSession)
```
psymoon: What about when we consume the `popupSession` from the store below?

app/src/main/java/org/mozilla/fenix/addons/AddonPopupBaseFragment.kt
csadilek: OK, this isn't a change I made, but let me try.

app/src/main/java/org/mozilla/fenix/addons/WebExtensionActionPopupFragment.kt
csadilek: What about it :) ?

psymoon: Shouldn't we make sure to dispatch `AddCustomTabAction` for that as well?

csadilek: @psymoon no, we don't need to do that, as this is handled by the base class in createView/destroyView.

Issue #8277, opened 2020-02-10
Update branding for fennecBeta build type


pocmo:
- [x] App name
- [x] Launcher icons
- [x] Homescreen header
- [x] Widget logo
- [x] Widget preview image
- [x] About screen

pocmo:
For some reason the widget is using a different icon with different proportions - just using the app logo instead does not work. I tried to look at the original issue and PR and could not figure out why, but I had no luck. Since this was implemented by a contributor the reason for that is a bit lost. I'll try if I can get the widget to use the app logo instead. This will avoid duplicating this over multiple flavors.

pocmo:
Done!

ebalazs-sv:
I had tested this on Beta Migration Build 2/19 with LG G7 fit (Android 8.1) and the results are:
- The app name is Firefox Beta;
- The launcher icon, the homescreen shortcuts icon, the pwa's icon, the widget icon are updated to the icon from https://github.com/mozilla-mobile/fenix/issues/6610#issuecomment-576825543;
- The wordmark from the onboarding screen, homescreen and about Firefox is updated to the wordmark available https://github.com/mozilla-mobile/fenix/issues/6610#issuecomment-576825543;
I will leave this on Ready for QA until it will be verified on the official Beta Migration build. Thanks!

ebalazs-sv:
Verified as fixed on Fennec Beta migration Build 1, 02/24, with LG G7 fit (Android 8.1) , Google Pixel 2 (Android 9) and the Firefox Beta name, icon and wordmark are updated as presented in the previous comment.
I will move this to Done. Thanks!

PR #8434, opened 2020-02-14
Issue #8277: Update widget preview for fennecBeta build type.

pocmo:
This should be the last branding piece for #8277. :)


kglazko:
Just updating an icon, checked that it's the correct icon, looks good to me :)

PR #8439, opened 2020-02-14
Issue #5194: Reuse Fennec taskAffinity flag.

pocmo:
Reusing the same taskAffinity makes sure that we are launching into a potentially already
existing task from Fennec. Without that we may spawn a new task and will show up twice
in "Recent apps" - with one entry still having a fennec screenshot.


pocmo:
> What will happen to Fenix channel updates? Will we see a similar bug there to what we see with the migration builds? I guess that doesn't really matter, just curious.

Yes, I think that is in the realm of possibilities. I think it will be less confusing though since the other task will at least like the app they expect to see there. :)

PR #8477, opened 2020-02-18
Fix #8469 - Properly align radio buttons in two-pane onboarding layouts.

### LINKED ISSUE #8469: [Bug][RTL] Radio buttons for the Theme selection on the welcome screen are overlapping the images
Originally filed in [bug 1616101](https://bugzilla.mozilla.org/show_bug.cgi?id=1608218):
תמונה

The radio buttons should be placed in the punch holes.

Mugurell:
### Pull Request checklist

- [x] **Tests**: This PR does not include tests as it only adds minor cosmetic changes
- [x] **Screenshots**: This PR includes screenshots for the before and after changes
- [x] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md)

### After merge

1. click on Show All Checks,

ProperlyAlignedRadioButtonsWithOnboardingImages


ekager:
Hey @Mugurell ! This looks great, did you try using `android:autoMirrored="true"` for these resources to be flipped as well? I think it makes more sense for RTL to have the radio button and cutout be on the right edge for RTL, but if it's not working, then this is definitely an improvement

Mugurell:
@ekager `autoMirrored` doesn't work but we can mirror the view the old fashioned way - just rotate it 180 degrees.
There are some inaccuracies in the images now (most visible being the virtual navigation back button) for which I think we'd need a new ticket for UX to provide updated resources to use in RTL though.

ekager:
@Mugurell could we do something like this? I think creating an `ic_onboarding_dark_theme` and `ic_onboarding_light_theme` that just wraps the pngs in a bitmap with automirroring true could work and be a bit simpler? Let me know! 
https://stackoverflow.com/a/44637677

Mugurell:
Thanks! Indeed this works great!

Issue #7448, opened 2019-12-31
Leanplum has a large impact on cold startup (5.68% to first frame drawn)


mcomella:
We're trying to move everything non-essential until after visual completeness in order to improve startup time. We should see if we can move Leanplum too. If not, consult the team: if the impact on startup is non-trivial, we may wish to remove it for GA until we can come up with a better solution.

---

_old comment 0_

Note: #6679 already exists but I'm not sure what the deliverables are so I filed a new issue: maybe we can dupe.

`Leanplum.start()` is pretty slow: it is 11.21% of `FenixApplication.onCreate` and has various intermittent startup problems that take up even more time: #7442. It may take up additional time to startup in other methods too.

We should consider creative solutions in order to improve leanplum performance on cold startup:
- Open PRs against the open source SDK
- Fork the SDK
- Write our own wrapper around the Leanplum API (Will's idea)
- Defer init until after startup
- Init off the main thread (this was tried already and may not be possible)
- Remove it entirely

**Priority recommendation:** this issue may take a long time to implement so it's probably mid-range priority.

---

Jan 19 update: when commenting out `LeanplumMetricsService`'s initializer, I see a 5.68% improvement in Displayed time, Pixel 2, local geckoBetaFenixBeta build: 395.83ms to 373.33ms.

This is with a debug key: Leanplum may have a smaller impact in release variants.

### Acceptance criteria
- Identify best course of action wrt improving Leanplum performance
- Attempt to implement or close this issue if unactionable

hawkinsw:
1. Writing our own wrapper around the API would be far more difficult than @boek and I originally imagined. When we discussed this option, we thought that we did not use Leanplum's Push mechanics. After talking to @jonalmeida, we realized that this is not the case. Reimplementing their Push mechanics would be relatively expensive (in terms of time and testing) and probably not something that we want to pursue.
2. With respect to deferring initialization until after startup, this is the option that @jonalmeida is pursuing in #6730.
[...]

PR #8558, opened 2020-02-19
For #7448: Move LP to IO dispatcher

MarcLeclair:
### Pull Request checklist 

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/metrics/LeanplumMetricsService.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/metrics/LeanplumMetricsService.kt
@@ -10,7 +10,14 @@ import com.leanplum.Leanplum
 import com.leanplum.LeanplumActivityHelper
 import com.leanplum.annotations.Parser
 import com.leanplum.internal.LeanplumInternal
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.Dispatchers.Main
 import mozilla.components.support.locale.LocaleManager
+import org.mozilla.fenix.components.metrics.MozillaProductDetector.MozillaProducts
 import org.mozilla.fenix.BuildConfig
 import org.mozilla.fenix.ext.settings
 import java.util.Locale
@@ -37,6 +44,9 @@ private val Event.name: String?
     }
 
 class LeanplumMetricsService(private val application: Application) : MetricsService {
+    val scope = CoroutineScope(Dispatchers.IO)
+    var leanplumJob: Job? = null
+
     data class Token(val id: String, val token: String) {
         enum class Type { Development, Production, Invalid }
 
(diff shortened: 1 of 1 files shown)


MarcLeclair:
So quick summary, we saw gains of ~40-50ms on a G5+ and Px2, so it could be safe to assume a higher gain on the G5. Also, @mcomella  found out that we use Leanplum on a background thread in Fennec. Finally, looking inside Leanplum, this background operation doesn't seem to affect anything that it does. It also shouldn't affect our push services.

app/src/main/java/org/mozilla/fenix/components/metrics/LeanplumMetricsService.kt
```diff
@@ -58,46 +61,52 @@ class LeanplumMetricsService(private val application: Application) : MetricsServ
...
-            "fxa_has_synced_items" to application.settings().fxaHasSyncedItems,
-            "search_widget_installed" to application.settings().searchWidgetInstalled,
-            "fenix" to true
-        ))
+            Leanplum.setIsTestModeEnabled(false)
+            Leanplum.setApplicationContext(application)
+            Leanplum.setDeviceId(randomUUID().toString())
+            Parser.parseVariables(application)
+
+            LeanplumActivityHelper.enableLifecycleCallbacks(application)
```
csadilek: This one is more concerning as it sets a `ActivityLifecycleCallback`, which receives callbacks on the main thread. It in turn invokes `onPause`/`onResume`/`onStop` on `Leanplum`, which reads/writes state that isn't thread-safe either. For some of these states it doesn't matter as they are only written/read on the main thread e.g. `isActivityPaused` and `currentActivity`. However, `pause` also checks `hasCalledStart` which is now set on an IO thread when starting `Leanplum`. Since, there's no visibility guarantee for that field, it could be that we fail to pause properly. As far as I can tell, this could cause us to continue to send heartbeats while paused. I think we need to discuss this more and do some manual testing.

(Sorry if you get this twice, Github somehow dropped my earlier comment)

mcomella: > Since, there's no visibility guarantee for that field

For a short term quick fix, could we fork the SDK and add `@Volatile`?

csadilek: I think if `calledStart` is our only concern we could also switch to the MainScope() after we successfully started and set it ourselves (on the right thread). What we have to do, I think, is go over all states set by the lifecylce callback and make sure we understand the consequences. It seems doable to review those states.

app/src/main/java/org/mozilla/fenix/components/metrics/LeanplumMetricsService.kt
```diff
@@ -58,46 +67,61 @@ class LeanplumMetricsService(private val application: Application) : MetricsServ
...
+                "fxa_signed_in" to application.settings().fxaSignedIn,
+                "fxa_has_synced_items" to application.settings().fxaHasSyncedItems,
+                "search_widget_installed" to application.settings().searchWidgetInstalled,
+                "fenix" to true
+            ))
+
+            withContext(Main) {
+                LeanplumInternal.setCalledStart(true)
+                LeanplumInternal.setHasStarted(true)
+                // Verify that this is currently set to true also
```
csadilek: Ah, this was meant as a TODO before landing :). We should verify that we do in fact always set `startedInBackround` to true during start. Then we can set it here as well to make sure.

app/src/main/java/org/mozilla/fenix/components/metrics/LeanplumMetricsService.kt
```diff
@@ -113,6 +137,7 @@ class LeanplumMetricsService(private val application: Application) : MetricsServ
         // as LP stores it's state internally and check against it
         LeanplumInternal.setCalledStart(false)
         LeanplumInternal.setHasStarted(false)
+        leanplumJob!!.cancel()
```
csadilek: This needs to be `leanplumJob?.cancel()`. If there's ever overlap of starting/stopping we don't want crash with an NPE, but simply do nothing as there is no job to cancel.

app/src/main/java/org/mozilla/fenix/components/metrics/LeanplumMetricsService.kt
csadilek: @MarcLeclair I've verified that this is the case now. So, you can remove the comment. As we're passing the application context here `LeanplumActivityHelper.currentActivity` will always be null and therefore `startedInBackground` will be true.

PR #8579, opened 2020-02-20
Bug 1616641 - let fennec-nightly tasks ride along vanilla fenix nightlies

MihaiTabara:
We no longer need standalone `fennec-nightly` graphs. Hence we remove their graph definitions and let those tasks ride along the vanilla fenix nightlies. 

Moreover, we want the fennec-nightly push-apk job to push to `beta` track directly, rather than `alpha` track so that Fenix folks have one less thing to take care of, now that we're 100% migrated on Fenix nightly. That change is tracked in https://github.com/mozilla-releng/scriptworker-scripts/pull/148

### DIFF
--- a/.cron.yml
+++ b/.cron.yml
@@ -12,12 +12,6 @@ jobs:
       when:
           - {hour: 6, minute: 0}
           - {hour: 18, minute: 0}
-    - name: fennec-nightly
-      job:
-          type: decision-task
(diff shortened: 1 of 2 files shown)


MihaiTabara:
Tested via `taskgraph` locally and the three additional tasks show up successfully in the fenix graph: build, signing and pushapk.

PR #8585, opened 2020-02-20
Bug 1613416 - Hide "Report site issue" button in Fenix Production.

denschub:
This is for [Bugzilla Bug 1613416](https://bugzilla.mozilla.org/show_bug.cgi?id=1613416). We are a bit worried that we'll receive too many issues from the Fenix release population, so we'll go the Desktop-route and have the reporter only enabled in prerelease builds of Fenix. For now, at least.

I'm honestly not 100% sure how to test this change. There is a test for the button's handler in `app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarController.kt`, which should be working either way, but the ReleaseChannel-based switch is untested. I'm happy to add tests if someone nudges me into the right direction.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
@@ -22,7 +22,9 @@ import mozilla.components.browser.menu.item.BrowserMenuItemToolbar
 import mozilla.components.browser.session.Session
 import mozilla.components.browser.session.SessionManager
 import mozilla.components.concept.storage.BookmarksStorage
+import org.mozilla.fenix.Config
 import org.mozilla.fenix.R
+import org.mozilla.fenix.ReleaseChannel
 import org.mozilla.fenix.browser.browsingmode.BrowsingMode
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
```diff
@@ -132,6 +134,8 @@ class DefaultToolbarMenu(
             context.components.browsingModeManager.mode == BrowsingMode.Normal
         val shouldDeleteDataOnQuit = Settings.getInstance(context)
             .shouldDeleteBrowsingDataOnQuit
+        val shouldShowWebcompatReporter =
+            Config.channel != ReleaseChannel.FenixProduction
```
pocmo: You want to add `FennecProduction` here too.

denschub: Would it be a good idea for me to add an `isRelease` check into `app/src/main/java/org/mozilla/fenix/Config.kt` in addition to `isReleaseOrBeta`, or is this use-case too niche? :)

boek:
Fixed with #8613

Issue #8089, opened 2020-02-05
Use Inter font for body text


brampitoyo:
In #3687, we proposed using the Metropolis font family for headers, buttons and snackbars. That issue is not yet complete, because we haven’t changed the body text font.

We propose setting all body text font using Inter: https://github.com/rsms/inter/

jonalmeida:
Keeping this offline discussion logged here: 

We want this to land with Fenix UI styled as well as the BrowserToolbar in AC at the same time. For this, we need to make the BrowserToolbar styleable (ac work) and use the font in the Fenix UI (fenix work).

amedyne:
@topotropic @brampitoyo Do we still think that we need to pursue this change? Asking in connection to @topotropic's topic on the last Work Week re carbon footprint.

gabrielluong:
Closing this since I don't believe there's a plan to use Inter.

PR #6986, opened 2019-12-06
For #3687 Use Metropolis Font for Header

kglazko:
---

### Pull Request checklist

- [ ] **Quality**: This PR builds and passes detekt/ktlint checks (A pre-push hook is recommended)
- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features

### After merge
- [ ] **Milestone**: Make sure issues finished by this pull request are added to the [milestone](https://github.com/mozilla-mobile/fenix/milestones) of the version currently in development.

### To download an APK when reviewing a PR:
1. click on Show All Checks,
2. click Details next to "Taskcluster (pull_request)" after it appears and then finishes with a green checkmark,
3. click on the "Fenix - assemble" task, then click "Run Artifacts".
4. the APK links should be on the left side of the screen, named for each CPU architecture

PR #8602, opened 2020-02-20
For #8089 Use InterFont for body

kglazko:
### Pull Request checklist

### After merge

1. click on Show All Checks,


liuche:
Does this need a rebase, or a repush to get the UI tests running again? If this is related to #6986 should we flag @ekager for review too?

kglazko:
Yes this definitely needs a rebase and then a review.

app/src/main/res/values/styles.xml
```diff
@@ -67,6 +68,7 @@
         <item name="bottomBarBackground">@drawable/home_bottom_bar_background</item>
         <item name="bottomBarBackgroundTop">@drawable/home_bottom_bar_background_top</item>
         <item name="privateBrowsingButtonBackground">@android:color/transparent</item>
+        <item name="privateBrowsingButtonBackground">@android:color/transparent</item>
```
ekager, line 71: nit - duplicate line  the above line already has this :)

kglazko:
Thanks @ekager ! I am having an issue in this PR I need to fix before I can merge. Some of the Metropolis headers are getting overriden by declaring InterFont in the normal theme. I'm trying to figure out why, let me know if you have any ideas :D

kglazko:
Some notes about the implementation- there are textviews that unfortunately don't use BodyText or Header styles, and those require manually adding InterFont into their style. In the future style refactor, I anticipate that every view will use a style that is correct for it (i.e. body text will use text). This PR is longer because it accounts for those elements where they are not using an existing text style but represent simple, non-header text.

The ideal practice would be- Metropolis is defined in Header/Button/Snackbar style, InterFont in the body text style (and all the elements that use a text not header would use the text style) @sblatz @ekager

app/src/main/res/layout/component_awesomebar.xml
```diff
@@ -14,6 +14,7 @@
     android:requiresFadingEdge="vertical"
     app:layout_constraintTop_toTopOf="@id/awesomeBar_barrier"
     app:layout_constraintEnd_toEndOf="parent"
+    app:fontFamily="@font/inter_regular"
```
sblatz, line 17: Is there a reason we're not just applying a style to this 🤔 does AC not support that?

sblatz:
Empty collections text looks a bit off, though I noticed this is also in master. Should this be changed?

image

It also doesn't seem to match up with the "no open tabs" text or are my eyes playing tricks on me?

image

sblatz:
@kglazko and I chatted about this, let's follow up with UX before landing this 😄

brampitoyo:
@sblatz @kglazko I concur with your findings. The body text:

- “Collect the things that matter to you…” is set in Metropolis.
- “Your open tabs is shown here” is set in Roboto – but on this image, all body text and caption seems to also be set in Roboto

Is there any other screen that we should review? Would you like us to go through each screen one by one and perform a visual check? We can help!

kglazko:
@brampitoyo Upon discussing with Sawyer, the way that the elements are currently distributed across AC and Fenix would requiring manually setting the attributes for some of the elements in Android Components. This would have to happen for some new things added to AC as well. 

This makes maintaining a consistent overall theme with 2 different fonts very difficult, practically speaking. However, adding the ability to set an overall global custom font through AC would ensure that all the elements adhere to one. 

Would UX consider for the future perhaps choosing a global font such as Metropolis for everything or InterFont for everything? This would make maintainable and consistency much easier in the future as the browser continues to grow. If not, we can certainly work around it for now- but it could become difficult to keep everything matching in the future.

brampitoyo:
Our next step is to meet with @csadilek and discuss setting custom font in AC.

liuche:
hey @brampitoyo, is this something that we still want? did you meet with @csadilek about this or is this something for eng to do?

kglazko:
After talking to @csadilek today, I'm going to file tickets to make 'Library' and 'Url bar' able to be customizable style-wise in Android Components. More of these will likely crop up and need tickets as well, but this seems to be the best approach since adding fonts to AC doesn't seem to be a good approach at this time.

brampitoyo:
@kglazko Thanks for unblocking me on this issue. I know that we want not only one custom font, but two – and this is a big ask for the A-C team. This will have to be an ongoing discussion. I also didn’t expect it to be such a big issue!

My preference is for Inter to be landed in all parts of the app all at once, instead of piecemeal (e.g. URL bar or main menu only). This way, users would only ever see one visual transition and won’t get the uncanny effect of “is this Roboto, or is this Inter?” because the two fonts can look similar.

In my view, it would be ok to keep using Roboto until the day that we can transition to Inter 100%. In the meantime, Metropolis has landed in all but one spot in the app – and Roboto can be paired with Metropolis just fine.

What do you think?

liuche:
Hi @kglazko since we're waiting on getting all the fonts over at once, should we close this for now?

kglazko:
Sure we can do that!

Issue #8570, opened 2020-02-20
[Bug]Disabling uBlock and quickly opening the Settings section crashes Fenix


AndiAJ:
## Prerequisites
Have uBlock Origins installed.

## Steps to reproduce
1. Disable uBlock and quickly tap the Settings section.

### Expected behavior
The Settings section should be properly loaded.

### Actual behavior
Crash

### Device information

* Android device: 
•  LG G7 fit (Android 8.1)
•  Huawei P9 Lite (Android 7)
• OnePlus A3 (Android 6.0.1)

* Fenix version: 
• Beta 4.0.0-beta.1 from 2/19
• Nightly #20510607 from 2/20
• Beta Migration build 2/19
• Nightly migration build 9, 02/14

### Notes
► Log
```kotlin
    --------- beginning of crash
02-20 12:26:43.460 28685-28685/? E/AndroidRuntime: FATAL EXCEPTION: main
    Process: org.mozilla.fenix.nightly, PID: 28685
    org.mozilla.geckoview.GeckoResult$UncaughtException: java.lang.IllegalStateException: remove_add_on must not be null
        at org.mozilla.geckoview.GeckoResult.dispatchLocked(GeckoResult.java:4)
        at org.mozilla.geckoview.GeckoResult.completeExceptionally(GeckoResult.java:4)
        at org.mozilla.geckoview.GeckoResult.lambda$thenInternal$2$GeckoResult(GeckoResult.java:10)
        at org.mozilla.geckoview.-$$Lambda$GeckoResult$gwCgOUK_EYQn2g6GolfZvo6A_WE.run(lambda)
        at org.mozilla.geckoview.GeckoResult.lambda$dispatchLocked$3(GeckoResult.java:2)
        at org.mozilla.geckoview.-$$Lambda$GeckoResult$PehjG2jgyDa_p37vZrKrSi2I94s.run(lambda)
        at android.os.Handler.handleCallback(Handler.java:739)
        at android.os.Handler.dispatchMessage(Handler.java:95)
        at android.os.Looper.loop(Looper.java:152)
        at android.app.ActivityThread.main(ActivityThread.java:5507)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:726)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:616)
[...]

PR #8604, opened 2020-02-20
For issue #8570: Disabling uBlock and quickly opening the Settings section cause a crash

Amejia481:
### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/addons/InstalledAddonDetailsFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/addons/InstalledAddonDetailsFragment.kt
@@ -13,7 +13,6 @@ import androidx.core.view.isVisible
 import androidx.fragment.app.Fragment
 import androidx.navigation.Navigation
 import androidx.navigation.findNavController
-import kotlinx.android.synthetic.main.activity_addons.view.*
 import kotlinx.android.synthetic.main.fragment_installed_add_on_details.*
 import kotlinx.android.synthetic.main.fragment_installed_add_on_details.view.*
 import kotlinx.coroutines.CoroutineScope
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/fenix/addons/InstalledAddonDetailsFragment.kt
```diff
@@ -94,7 +94,7 @@ class InstalledAddonDetailsFragment : Fragment() {
         switch.setOnCheckedChangeListener { v, isChecked ->
             val addonManager = v.context.components.addonManager
             switch.isClickable = false
-            remove_add_on.isEnabled = false
+            view.remove_add_on.isEnabled = false
```
psymoon: Interesting. Does kotlin view binding not recognize/handle the differences here?

Amejia481: It looks like the Kotlin synthetic view binding variable get null at some point when we are switching between screen and as this code gets called async the null pointer exception happens

app/src/main/java/org/mozilla/fenix/addons/InstalledAddonDetailsFragment.kt
```diff
@@ -129,14 +128,14 @@ class InstalledAddonDetailsFragment : Fragment() {
                     }
                 )
             } else {
+                view.settings.isVisible = false
```
Amejia481: I had to add this extra step here, as we are hiding the settings section after  `onSuccess` if the user clicks fast enough can enter to the setting page, but when the user arrives there the add-on is already disabled and the page can't be shown.

image

gabrielluong:
Can you also check if we need to make this change in r-b and sample-browser

Amejia481:
Sure!

Issue #8034, opened 2020-02-01
Allow jobs to run after visual completeness


mcomella:
This will allow us to prevent initialization of some non-essential components from blocking startup.

Will has a WIP: https://github.com/mozilla-mobile/fenix/pull/7901

mcomella:
We discussed that we may not want to spend too much time on this in the short term so we should consider **doing something simple**. For example, we can create an abstraction for "visual completeness" that has a dumb implementation: e.g. scheduling a delay after the first frame is drawn. Some notes:
* How does Google determine visual completeness in their apps (e.g. to register `reportFullyDrawn`)?
* We could identify the last view we expect to be drawn (assuming there is only one async draw event) and call post on it.
* To address the problem that other Activities can open without HomeActivity, other Activities can opportunistically try to start these things on the main thread in their `onCreate` methods (if they're not already started) and HomeActivity can use the delayed approach. This is okay because HomeActivity (or similar), according to telemetry, is the most common thing to open (according to csadilek).
* I've used `getViewTreeObserver` + pre-draw listener and Marc mentioned the `Choreographer` API as options to find when frames are drawn.

emalysz:
I found this very useful blog post: https://developers.google.com/web/fundamentals/performance/user-centric-performance-metrics

Google these identified 3 important topics in measuring load performance:
1. **Is it happening?** 
       -First paint: first point at which we render anything that is different than previous screen
       -First contentful paint: first DOM content is rendered 
2. **Is it useful?**
       -First meaningful paint: time when hero elements (most important elements) are rendered
3. **Is it usable?** 
       -Visually rendered and usable. JS loaded and main thread is idle

They also use APIs like the `Performance observer` where you can subscribe to performance events and respond asynchronously. A polyfill also has a `getConsistentlyInteractive` method that tracks the time to interactive.

emalysz:
QA testing:
Ensure that startup is possible by navigating through various activities. 
Launches through IntentReceiverActivity, IntentReceiverActivity, and OSSLicenses (when passed a proper intent) should all be verified. 

Additionally, navigate to activities again to ensure the second time the activity is rendered we don't crash.

sv-ohorvath:
Verified this on Nightly 3/3, with 2 devices: Motorola Moto G6 (Android 8.0), Mi 4i (Android 5.0.2)
Areas covered:
- launch app from PlayStore
- homescreen launcher  
- private browsing launcher 
[...]

PR #8610, opened 2020-02-20
For #8034: Create a post-visual completeness executor

emalysz:
Create an object that will execute its enqueued, paused tasks
when Fenix is visually complete.

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
+++ b/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
@@ -37,6 +37,7 @@ import org.mozilla.fenix.components.Components
 import org.mozilla.fenix.components.metrics.MetricServiceType
 import org.mozilla.fenix.ext.settings
 import org.mozilla.fenix.session.NotificationSessionObserver
+import org.mozilla.fenix.session.PerformanceActivityLifecycleCallbacks
 import org.mozilla.fenix.session.VisibilityLifecycleCallback
 import org.mozilla.fenix.utils.BrowsersCache
 import org.mozilla.fenix.utils.Settings
@@ -143,6 +144,10 @@ open class FenixApplication : LocaleAwareApplication() {
         // if ((System.currentTimeMillis() - settings().lastPlacesStorageMaintenance) > ONE_DAY_MILLIS) {
         //    runStorageMaintenance()
         // }
+
+        registerActivityLifecycleCallbacks(
+            PerformanceActivityLifecycleCallbacks(components.performance.visualCompletenessTaskManager)
+        )
     }
 
     // See https://github.com/mozilla-mobile/fenix/issues/7227 for context.
--- a/app/src/main/java/org/mozilla/fenix/session/PerformanceActivityLifecycleCallbacks.kt
+++ b/app/src/main/java/org/mozilla/fenix/session/PerformanceActivityLifecycleCallbacks.kt
@@ -1,0 +1,79 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.session
+
+import android.app.Activity
+import android.app.Application
+import android.os.Bundle
+import org.mozilla.fenix.HomeActivity
+import org.mozilla.fenix.IntentReceiverActivity
+import org.mozilla.fenix.browser.BrowserPerformanceTestActivity
+import org.mozilla.fenix.settings.account.AuthIntentReceiverActivity
+import org.mozilla.fenix.utils.StartupTaskManager
+import org.mozilla.fenix.widget.VoiceSearchActivity
+
+/**
+ * These callbacks handle binding performance code to the activity lifecycle.
+ */
+@SuppressWarnings("EmptyFunctionBlock")
+class PerformanceActivityLifecycleCallbacks(
+    private val visualCompletenessTaskManager: StartupTaskManager
+) : Application.ActivityLifecycleCallbacks {
+
+    override fun onActivityCreated(activity: Activity, bundle: Bundle?) {
+        ifNecessaryPostVisualCompleteness(activity)
+    }
+
+    /**
+     * Returns true if it is a terminal activity, or false if it
+     * an activity used in transition.
+     */
+    private fun isTransientActivity(activity: Activity): Boolean {
+        // These are the current list of non terminal activites.
+        // They have been whitelisted in case new activities are added to the application
+        // to ensure these new activities would not crash the application.
+        return isTransientActivityInMigrationVariant(activity) ||
+                (activity is IntentReceiverActivity) ||
+                (activity is VoiceSearchActivity) ||
+                (activity is AuthIntentReceiverActivity) ||
+                (activity is BrowserPerformanceTestActivity)
+    }
+
+    /**
+     * This starts the StartupTaskManager, eitther delayed or right away, if the activity is a
+     * terminal activity. If not, do nothing.
+     */
+    private fun ifNecessaryPostVisualCompleteness(activity: Activity) {
+        fun shouldStartVisualCompletenessQueueImmediately(): Boolean {
+            return !isTransientActivity(activity)
+        }
+
+        if (activity is HomeActivity) {
+            // We should delay the visualCompletenessTaskManager when reaching the HomeActivity
+            // to ensure all tasks are delayed until after visual completeness
+            activity.postVisualCompletenessQueue(visualCompletenessTaskManager)
+        } else if (shouldStartVisualCompletenessQueueImmediately()) {
+            // If we do not go through the home activity, we have to start the tasks
+            // immediately to avoid spending time implementing it.
+            visualCompletenessTaskManager.start()
+        }
+    }
+
+    override fun onActivityStarted(activity: Activity?) {}
+    override fun onActivityStopped(activity: Activity?) {}
+    override fun onActivityResumed(activity: Activity?) {}
+    override fun onActivityPaused(activity: Activity?) {}
+    override fun onActivitySaveInstanceState(activity: Activity?, bundle: Bundle?) {}
+    override fun onActivityDestroyed(activity: Activity?) {}
+
+    companion object {
+        /**
+         * The source files are different from migration and non migration variants.
+         * We use this property to extend the [isTransientActivity] implementation for the
+         * migration variants.
+         */
+        var isTransientActivityInMigrationVariant: (Activity) -> Boolean = { false }
+    }
+}
--- a/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
+++ b/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
@@ -15,13 +15,14 @@ import androidx.annotation.VisibleForTesting
 import androidx.annotation.VisibleForTesting.PROTECTED
 import androidx.appcompat.app.ActionBar
 import androidx.appcompat.widget.Toolbar
+import androidx.core.view.doOnPreDraw
 import androidx.lifecycle.lifecycleScope
 import androidx.navigation.NavDestination
 import androidx.navigation.NavDirections
 import androidx.navigation.fragment.NavHostFragment
 import androidx.navigation.ui.AppBarConfiguration
 import androidx.navigation.ui.NavigationUI
-import kotlinx.android.synthetic.main.activity_home.navigationToolbarStub
+import kotlinx.android.synthetic.main.activity_home.*
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.launch
 import mozilla.components.browser.search.SearchEngine
@@ -69,6 +70,7 @@ import org.mozilla.fenix.settings.logins.SavedLoginsFragmentDirections
 import org.mozilla.fenix.theme.DefaultThemeManager
 import org.mozilla.fenix.theme.ThemeManager
 import org.mozilla.fenix.utils.BrowsersCache
+import org.mozilla.fenix.utils.StartupTaskManager
 
 @SuppressWarnings("TooManyFunctions", "LargeClass")
 open class HomeActivity : LocaleAwareAppCompatActivity() {
@@ -76,6 +78,10 @@ open class HomeActivity : LocaleAwareAppCompatActivity() {
     private var webExtScope: CoroutineScope? = null
     lateinit var themeManager: ThemeManager
     lateinit var browsingModeManager: BrowsingModeManager
+
+    private var isVisuallyComplete = false
+
+    private var visualCompletenessTaskManager: StartupTaskManager? = null
 
     private var sessionObserver: SessionManager.Observer? = null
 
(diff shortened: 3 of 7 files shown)


app/src/main/java/org/mozilla/fenix/utils/OnVisualCompletenessExecutionQueue.kt
```diff
@@ -0,0 +1,81 @@
...
+ * A queue of paused `Job`s whose execution is resumed when Fenix is visually complete.
+ *
+ * This queue contains paused `Job`s. Each Job in the queue will be restarted once Fenix
+ * is visually complete (as long as it has not be cancelled since it was added to the queue).
+ * Each Job will execute on the thread on which it was originally scheduled.
+ *
+ * This is threadsafe.
+ */
+object OnVisualCompletenessExecutionQueue {
+    internal var jobs = mutableListOf<Job>()
```
mcomella: At a high level, what we need is to execute tasks at a given point in time. We're currently using `Job` to do that but `Job` is more complex than we need, i.e. "a task executing at a given point in time": Jobs force you to use coroutines, support cancellation, etc. and that's prevalent by all the extra if statement checks below and the need for unit tests. Can we use something simpler?

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -479,6 +480,7 @@ class HomeFragment : Fragment() {
     override fun onResume() {
         super.onResume()
         hideToolbar()
+        OnVisualCompletenessExecutionQueue.post(sessionControlView!!.view)
```
mcomella: This doesn't take into account what happens if we restore in another Activity without going through `MainActivity`: I wrote up a potential solution here https://github.com/mozilla-mobile/fenix/issues/8034#issuecomment-587966952

tbh, I'm not sure if we can actually test that use case on device: afaik, Don't Keep Activities doesn't kill the app process like the system would.

app/src/main/java/org/mozilla/fenix/utils/OnVisualCompletenessExecutionQueue.kt
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    /**
+     * Posts a runnable action to the view and triggers the queue to start executing its jobs.
+     *
+     *@param v: A view that maintains its message queue.
+     */
+    @Synchronized
+    fun post(v: View) {
+        v.post {
```
mcomella: Does this actually run after visual completeness? It seems like it will run after the first frame is drawn and the framework yields the UI thread to our posted Runnable, i.e. before the tabs are restored.

app/src/main/java/org/mozilla/fenix/utils/OnVisualCompletenessExecutionQueue.kt
```diff
@@ -0,0 +1,81 @@
...
+        jobs.forEach { if (!it.isCancelled) { it.start() } }
+        jobs.clear()
+    }
+
+    /**
+     * Reset the state of the queue and clear its jobs
+     *
+     * This is for testing only.
+     */
+    @Synchronized
```
mcomella: @VisibleForTesting

app/src/main/java/org/mozilla/fenix/utils/OnVisualCompletenessExecutionQueue.kt
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * A queue of paused `Job`s whose execution is resumed when Fenix is visually complete.
+ *
+ * This queue contains paused `Job`s. Each Job in the queue will be restarted once Fenix
+ * is visually complete (as long as it has not be cancelled since it was added to the queue).
+ * Each Job will execute on the thread on which it was originally scheduled.
+ *
+ * This is threadsafe.
+ */
+object OnVisualCompletenessExecutionQueue {
```
mcomella: nit: I would prefer `class`

app/src/main/java/org/mozilla/fenix/utils/StartupTaskManager.kt
```diff
@@ -0,0 +1,40 @@
...
+    @Synchronized
+    private fun add(task: StartupTask) {
+        tasks.add(task)
+    }
+
+    /**
+     * Start all tasks in the queue. When all the tasks have been started,
+     * clear the queue.
+     */
+    @Synchronized
```
mcomella: This concurrency could result in `add` being blocked for a long time: e.g. if `start` takes a non-trivial amount of time, `add` would be blocked waiting for `start` to finish.

In general, handling concurrency is a non-trivial task: what if we made no thread-safety guarantees here? It's not really a requirement for this bit of code, right?

app/src/main/java/org/mozilla/fenix/utils/StartupTaskManager.kt
```diff
@@ -0,0 +1,40 @@
...
+    private var tasks = mutableListOf<StartupTask>()
+
+    /**
+     * Add a task to the queue.
+     *
+     * @param task: The task to add to the queue.
+     */
+    @Synchronized
+    private fun add(task: StartupTask) {
+        tasks.add(task)
```
mcomella: What happens if we add to the queue after it has been started? Should we enforce some guarantees?

app/src/main/java/org/mozilla/fenix/utils/StartupTaskManager.kt
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    /**
+     * Start all tasks in the queue. When all the tasks have been started,
+     * clear the queue.
+     */
+    @Synchronized
+    private fun start() {
+        tasks.forEach { it.run {} }
+        tasks.clear()
```
mcomella: Why do we need to clear the queue?

One thought: anything captured by the lambda will remain captured if we hold on to these tasks, taking up more memory than we need to. If this is the reason for this to exist, we should comment so.

app/src/main/java/org/mozilla/fenix/utils/StartupTaskManager.kt
```diff
@@ -0,0 +1,40 @@
...
+
+package org.mozilla.fenix.utils
+
+typealias StartupTask = () -> Unit
+
+/**
+ * A queue of tasks that are performed at specific points during Fenix startup.
+ *
+ * This queue contains a list of startup tasks. Each task in the queue will be started once Fenix
+ * is visually complete (as long as it has not be cancelled since it was added to the queue).
```
mcomella: nit: we no longer handle cancellation so we should remove this.

app/src/main/java/org/mozilla/fenix/utils/StartupTaskManager.kt
```diff
@@ -0,0 +1,40 @@
...
+ * Each task will execute on the main thread.
```
mcomella: nit: this comment should probably go where `add` is because:
1) devs are more likely to see method kdoc than the class kdoc
2) if we change the behavior, it's very likely `add` will let devs specify what thread it should run on

app/src/main/java/org/mozilla/fenix/components/Performance.kt
```diff
@@ -0,0 +1,11 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.components
+
+import org.mozilla.fenix.utils.StartupTaskManager
+
+class Performance {
```
mcomella: nit: please include kdoc for every class!

app/src/main/java/org/mozilla/fenix/components/Performance.kt
```diff
@@ -0,0 +1,11 @@
...
+    val startupTaskManager = StartupTaskManager()
```
mcomella: nit: lazy

app/src/main/java/org/mozilla/fenix/components/Performance.kt
mcomella: nit: there's a namespace collision between this and the `object Performance`: should be call this `PerformanceComponent` instead? I don't think anyone will ever need to refer to it by name.

app/src/main/java/org/mozilla/fenix/session/OnRegisterActLifeCycleCallback.kt
```diff
@@ -0,0 +1,35 @@
...
+
+    override fun onActivityResumed(activity: Activity?) {}
+
+    override fun onActivityPaused(activity: Activity?) {}
+
+    override fun onActivitySaveInstanceState(activity: Activity?, bundle: Bundle?) {}
+
+    override fun onActivityDestroyed(activity: Activity?) {}
+
+    override fun onActivityCreated(activity: Activity?, bundle: Bundle?) {
```
mcomella: nit: can we put this at the top of the class so we don't have to scroll to access it?

app/src/main/java/org/mozilla/fenix/session/OnRegisterActLifeCycleCallback.kt
```diff
@@ -0,0 +1,35 @@
...
+import android.app.Application
+import android.os.Bundle
+import org.mozilla.fenix.components.Performance
+
+/**
+ * This ActivityLifecycleCallbacks implementations tracks if a frame is visible and
+ * allows us to start a queue of tasks that have been delayed until after visual completeness.
+ */
+@SuppressWarnings("EmptyFunctionBlock")
+class OnRegisterActLifeCycleCallback(private val performance: Performance) :
```
mcomella: nit: typically in a dependency injection pattern, we'd pass in the smallest component needed by the class: in this case, the `startupTaskManager` rather than `Performance`.

mcomella: What is `OnRegsiterActLifeCycleCallback` mean? I don't really understand this classes' single responsibility. Perhaps it should be `PerformanceActivityLifecycleCallbacks` and it should handle binding all performance code to the lifecycle callbacks?

app/src/main/java/org/mozilla/fenix/utils/StartupTaskManager.kt
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * A queue of tasks that are performed at specific points during Fenix startup.
+ *
+ * This queue contains a list of startup tasks. Each task in the queue will be started once Fenix
+ * is visually complete.
+ */
+class StartupTaskManager {
+    private var tasks = mutableListOf<StartupTask>()
+    internal var hasExecuted = false
```
mcomella: nit: private. `internal` doesn't provide any guarantees when compared to public in the Fenix codebase.

app/src/main/java/org/mozilla/fenix/utils/StartupTaskManager.kt
```diff
@@ -0,0 +1,57 @@
...
+    /**
+     * Add a task to the queue.
+     * Each task will execute on the main thread.
+     *
+     * @param task: The task to add to the queue.
+     */
+    @Synchronized
+    fun add(task: StartupTask) {
+        if (hasExecuted) {
+            Log.log(
```
mcomella: nit: remove this: the exception is logged to logcat so we don't need to log in addition.

app/src/main/java/org/mozilla/fenix/utils/StartupTaskManager.kt
mcomella: nit: `hasStarted`? It makes the naming consistent with the `start()` method.

app/src/main/java/org/mozilla/fenix/utils/StartupTaskManager.kt
```diff
@@ -0,0 +1,57 @@
...
+                Log.Priority.WARN,
+                startupTaskManagerQueueLogTag,
+                null,
+                "Queue already executed. Not adding the Job."
+            )
+            throw StartupTaskManagerException("Queue already executed.")
```
mcomella: nit: can we use `IllegalStateException`? I don't know that defining our own exception type really grants us anything here (but feel free to disagree!). It might make our calls easier to find in Sentry but I'm not sure that's worth defining – and remembering to use – a new type over.

app/src/main/java/org/mozilla/fenix/utils/StartupTaskManager.kt
```diff
@@ -0,0 +1,57 @@
...
+        tasks.add(task)
+    }
+
+    /**
+     * Start all tasks in the queue. When all the tasks have been started,
+     * clear the queue.
+     */
+    fun start() {
+        tasks.forEach { it.run {} }
+        hasExecuted = true
```
mcomella: If we change this is `hasStarted`, we can move this before `tasks.forEach` so, just in case this is called from a separate thread, we have a chance of crashing.

app/src/main/java/org/mozilla/fenix/utils/StartupTaskManager.kt
```diff
@@ -0,0 +1,57 @@
...
+
+typealias StartupTask = () -> Unit
+
+/**
+ * A queue of tasks that are performed at specific points during Fenix startup.
+ *
+ * This queue contains a list of startup tasks. Each task in the queue will be started once Fenix
+ * is visually complete.
+ */
+class StartupTaskManager {
```
mcomella: nit: add a comment that this class isn't thread safe and we should call from the main thread?

app/src/main/java/org/mozilla/fenix/session/OnRegisterActLifeCycleCallback.kt
```diff
@@ -0,0 +1,35 @@
...
+    override fun onActivityResumed(activity: Activity?) {}
+
+    override fun onActivityPaused(activity: Activity?) {}
+
+    override fun onActivitySaveInstanceState(activity: Activity?, bundle: Bundle?) {}
+
+    override fun onActivityDestroyed(activity: Activity?) {}
+
+    override fun onActivityCreated(activity: Activity?, bundle: Bundle?) {
+        performance.startupTaskManager.start()
```
mcomella: I don't think this is going to run at visual completeness or even before the first frame: should we adjust it or follow-up?

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -142,6 +146,9 @@ open class FenixApplication : LocaleAwareApplication() {
         // if ((System.currentTimeMillis() - settings().lastPlacesStorageMaintenance) > ONE_DAY_MILLIS) {
         //    runStorageMaintenance()
         // }
+
+        onRegisterLifecycleCallback = OnRegisterActLifeCycleCallback(components.performance)
```
mcomella: nit: we don't use this reference for anything else: why do we need to hold on to it?

app/src/main/java/org/mozilla/fenix/utils/StartupTaskManager.kt
```diff
@@ -0,0 +1,55 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.utils
```
mcomella: nit: we should put in the omf.perf package

app/src/main/java/org/mozilla/fenix/session/PerformanceActivityLifecycleCallbacks.kt
```diff
@@ -0,0 +1,79 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.session
```
mcomella: nit: we should put in the omf.perf package, maybe?

PR #8613, opened 2020-02-20
🎲 Bug 1613416 - Hide "Report site issue" button in Fenix Production.

boek:
@denschub 
> This is for [Bugzilla Bug 1613416](https://bugzilla.mozilla.org/show_bug.cgi?id=1613416). We are a bit worried that we'll receive too many issues from the Fenix release population, so we'll go the Desktop-route and have the reporter only enabled in prerelease builds of Fenix. For now, at least.
>
> I'm honestly not 100% sure how to test this change. There is a test for the button's handler in `app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarController.kt`, which should be working either way, but the ReleaseChannel-based switch is untested. I'm happy to add tests if someone nudges me into the right direction.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
@@ -22,7 +22,9 @@ import mozilla.components.browser.menu.item.BrowserMenuItemToolbar
 import mozilla.components.browser.session.Session
 import mozilla.components.browser.session.SessionManager
 import mozilla.components.concept.storage.BookmarksStorage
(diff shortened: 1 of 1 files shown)


gabrielluong:
There is a related PR at https://github.com/mozilla-mobile/fenix/pull/8585.

boek:
@gabrielluong CI doesn't run on non-members/forks. So I had to push to my fork for CI to run :|

gabrielluong:
> @gabrielluong CI doesn't run on non-members/forks. So I had to push to my fork for CI to run :|

Ah, that's right! I originally made a note since I was unsure whether or not to merge and to also remind whoever did merge to close that other PR.

Issue #7781, opened 2020-01-17
FNPRMS fully drawn time does not take into account top sites (M4)


mcomella:
When the user has tabs or collections, the RecyclerView draws once for "No tabs" and draws again when the tabs arrive. The current `reportFullyDrawn` implementation does not take this second draw into account and thus is not entirely accurate.

hawkinsw:
After investigation today with @MarcLeclair , I believe that this needs to be a very high priority. After a recent change to the application (during the nightly migration), there is a new intent dispatcher that has been inserted into startup. This impacts when the application's first layout traversal happens -- it moves it from after the home fragment is inflated to before the home fragment is created. This means that the Displayed time (upon which we rely for our FNPRMS numbers) will now no longer reflect the time that it takes to inflate the home fragment. 

@MarcLeclair and I confirmed this from a theoretical and an empirical perspective. In order to test it empirically, we added a `Thread.sleep(5000)` in the `HomeActivity.onCreate()` method and watched that our Displayed time did not skyrocket. We compared that to adding a `Thread.sleep(5000)` to the `GleanMetricsService.setStartupMetrics()` method (which executes before the first layout traversal) where that addition did affect the Displayed time by an appropriate amount. 

Not only does this reinforce the fact that we need to add reportFullyDrawn and use that for FNPRMs, but it also confirms the investigation into when the Displayed flag.

mcomella:
I landed https://github.com/mozilla-mobile/fenix/pull/7780 and I believe it already addresses the case hawkinsw is concerned about above https://github.com/mozilla-mobile/fenix/issues/7781#issuecomment-576451112. Addressing this ticket would fix the case where we have open tabs (and thus we redraw the RV again) but that isn't one of our current requirements. Sending this back to triage...

mcomella:
Visual completeness is strange here: we display an animation to show the async-loaded tabs but we probably don't want to include the animation in our time to visual completeness. Theoretically, we could instrument the first frame of that animation though.

mcomella:
The PR https://github.com/mozilla-mobile/fenix/pull/8615 landed. Unfortunately, I don't see the consistent regression we expect in the FNPRMS results: it landed on Feb 28 and Feb 29 is 8ms regression while Mar 1 is additional 9ms regression. I wonder if another improvement landed.

mcomella:
This does not look like it will impact stability nor can it be tested by QA so not requesting QA and can close.

PR #8615, opened 2020-02-20
Issue #7781: instrument top sites with report fully drawn

mcomella:
### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/perf/Performance.kt
+++ b/app/src/main/java/org/mozilla/fenix/perf/Performance.kt
@@ -65,7 +65,7 @@ object Performance {
      * it is for testing visual metrics. These checks aren't full proof but most of our users won't have
      * ADB on and USB connected at the same time when running Firefox.
      */
-    fun isPerformanceTest(intent: Intent, context: Context): Boolean {
+    private fun isPerformanceTest(intent: Intent, context: Context): Boolean {
         if (!intent.getBooleanExtra(EXTRA_IS_PERFORMANCE_TEST, false)) {
             return false
         }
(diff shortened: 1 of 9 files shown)


app/src/main/java/org/mozilla/fenix/perf/Performance.kt
```diff
@@ -23,27 +23,29 @@ object Performance {
...
-        activity.rootContainer.doOnPreDraw {
-            activity.reportFullyDrawn()
-        }
+        // We keep the code simple by calling this more than once, which we don't mind because:
+        // - reportFullyDrawn is a no-op if it's called more than once
+        // - top sites won't be bound often so redundant pre draw listeners aren't expected to have
+        // a significant perf impact
+        //
+        // Ideally we wouldn't cast to HomeActivity but we want to save implementation time.
+        holder.itemView.doOnPreDraw { (it.context as HomeActivity).reportFullyDrawn() }
```
MarcLeclair: I think that's a good start, but looking at frames callback ( not the one I wrote but `systrace` version, I notice this would be called ~100-200 (it hovers between those two). So we actually miss that ending time. I have a few ideas on how to get that to that extra 100-200ms. Also, this changes with more top sites we have since we get the onPreDraw before any viewholders are rendered

mcomella: > I notice this would be called ~100-200

What does this mean? systrace's frame drawn time is called 100ms-200ms after our "fully drawn" is reported? If so, I think there are a few possible attributions:
- I instrumented the first frame drawn before the animation rather than at the end of the animation because I don't think we should include an animation duration within our measurement. I'm checking with Eric if this is what is desired & will add comments if so. If you were looking at the last frame drawn, this could be where the discrepancy came from.
- using `onPreDraw` rather than `onDraw`
- possible measurement overhead when using systrace or other profiling tools – not sure if it's real
- possible timestamps that don't line up between profiling tools & runtime

Keep in mind our current goal is to compare Fennec to Fenix so we only need for them to be consistent with one another, not necessarily 100% accurate (which is more complex and thus maybe not worth it right now).

MarcLeclair: So if our goal is only to be close to Fennec, I say let's go ahead with this. Only caveat to the comment is that we're missing the onBind (so layout pass) of the view holders + the animation.  If our goal is to be close to fennec, we would need to maybe get that after the onbind pass but before the animation pass

mcomella:
After this PR, when disabling RV animations (with `view.itemAnimator = null`) & posting from the `onPreDraw` call instrumented here, I consistently get this frame drawn (on both P2 & G5):

device-2020-02-24-141248

This makes me think we've got the right spot.

mcomella:
Rebasing so I can merge...

PR #8632, opened 2020-02-21
Fix #8631 - Persist the same activity for http/s schemes after migration

### LINKED ISSUE #8631: Fennec migration: Default browser state is not saved after migration on Huawei device
STR:
1. Make sure to have another browser installed, e.g. Chrome.
2. Set Fennec as default browser.
3. Update Fennec to Fenix from Gplay, or through adb.
4. Check the default browser setting.

Expected results:
The Firefox nightly app is still the default browser.

Actual result:
Default browser is reset and changed to the other installed app (Chrome).

Device: Huawei P20 lite (Android 9), Huawei Mate 20 Lite (Android 8.1.0)
*not reproducing on other types of OS's.
Build: Fennec Nightly migration build 5

Mugurell:
Most of the times setting an application as default does not restrict it's
activities so that only that initial one can handle a specific Intent.
If that activity is renamed/deleted or it's intent filter is now declared for
another, normally, the same application will remain set as default.

IRL we saw that Huawei imposes strict restrictions regarding the default apps
settings, such that any changes in preferred activities would result in
dropping the current default app to let the user pick again a default one.

To protect against such edgecases and keep the default app status previously set
for Fennec we'll keep the same Fennec component declared in Fenix's manifest
as the activity set to handle http/s schemes.

### Pull Request checklist

- [x] **Tests**: This PR does not includes tests. Can only be manually tested.
- [x] **Screenshots**: This PR includes a video showing the issue being fixed.
https://drive.google.com/file/d/1ZaTT2uS_XITCFqJfE9KsUq5fsVTi2bQR/view?usp=sharing
(the Fennec -> Fenix update starts at around 00:45)
- [x] **Accessibility**: The code in this PR follows does not include any user facing features.

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -35,6 +35,10 @@
 
           Note that `fennec*` build types override the targetActivity property in the Manifest
           inside their source set.
+
+          Fennec's activity set to handle http/s schemes.
+          Need to also use it in Fenix to keep the previously default app status after migration.
+          Also the Fennec declared entry for homescreen pinned shortcuts.
         -->
         <activity-alias
             android:name="${applicationId}.App"
@@ -43,6 +47,14 @@
                 <action android:name="android.intent.action.MAIN" />
 
                 <category android:name="android.intent.category.LAUNCHER" />
+            </intent-filter>
+
+            <intent-filter>
+                <action android:name="android.intent.action.VIEW" />
+                <category android:name="android.intent.category.DEFAULT" />
+                <category android:name="android.intent.category.BROWSABLE" />
+                <data android:scheme="http" />
+                <data android:scheme="https" />
             </intent-filter>
 
             <meta-data
@@ -122,12 +134,7 @@
 
             <intent-filter>
                 <action android:name="android.intent.action.VIEW" />
-
-                <category android:name="android.intent.category.DEFAULT" />
-                <category android:name="android.intent.category.BROWSABLE" />
                 <category android:name="mozilla.components.pwa.category.SHORTCUT" />
-                <data android:scheme="http" />
-                <data android:scheme="https" />
             </intent-filter>
 
             <intent-filter>


app/src/main/AndroidManifest.xml
```diff
@@ -109,15 +121,6 @@
...
         <activity android:name=".IntentReceiverActivity" android:theme="@style/Theme.Transparent">
-            <intent-filter>
-                <action android:name="android.intent.action.VIEW" />
-
-                <category android:name="android.intent.category.DEFAULT" />
-                <category android:name="android.intent.category.BROWSABLE" />
-                <category android:name="mozilla.components.pwa.category.SHORTCUT" />
-                <data android:scheme="http" />
-                <data android:scheme="https" />
-            </intent-filter>
```
pocmo: I am not fully understanding what the fix is. **Not** having this filter here is the fix?

app/src/main/AndroidManifest.xml
```diff
@@ -51,12 +51,24 @@
...
         <activity-alias
             android:name="org.mozilla.gecko.BrowserApp"
             android:targetActivity=".IntentReceiverActivity"
             android:exported="true">
+            <intent-filter>
+                <action android:name="android.intent.action.VIEW" />
+
+                <category android:name="android.intent.category.DEFAULT" />
+                <category android:name="android.intent.category.BROWSABLE" />
+                <category android:name="mozilla.components.pwa.category.SHORTCUT" />
```
pocmo: Doesn't that break PWAs that need to go through `IntentReceiverActivity`? (Haven't tested the patch yet)

app/src/main/AndroidManifest.xml
Mugurell: The fix would be to keep the `action.VIEW` `category.DEFAULT` `http\https` intent filter for `org.mozilla.gecko.BrowserApp` since this is the preferred activity set for the default browser on Fennec.
Seems like on Huawei the default app setting is tied to the activity, not to just the package name.
I removed the same intent-filter from IntentReceiverActivity since when testing I saw that keeping it there also (after initially duplicating it for ..BrowserApp) means it having a higher precedence? (same intent filter for two activities..?) and Android would see it as the one who should handle the http schemes -> this would mean again a change from BrowserApp so the default app status would again be lost.

app/src/main/AndroidManifest.xml
Mugurell: Right. Thanks. Just tested and indeed the PWAs are lost.

Mugurell:
Ohh, sorry, I added that intent-filter to ...BrowserApp erroneously as I based my testing on a modified Fennec manifest.
I tested this more and saw that the Fennec's preferred activity for when setting a default browser is actually `.org.mozilla.gecko.App`.

The same "fix" logic still holds though so after adding a new

         <activity-alias
            android:name="org.mozilla.gecko.App"
            android:targetActivity=".IntentReceiverActivity"
            android:exported="true">
            <intent-filter>
                <action android:name="android.intent.action.VIEW" />

                <category android:name="android.intent.category.DEFAULT" />
                <category android:name="android.intent.category.BROWSABLE" />

                <data android:scheme="http" />
                <data android:scheme="https" />
            </intent-filter>
        </activity-alias>

I see that after migration the default app is still Firefox.

Will check that pwas and shortcuts still work and will post apks for QA to test before review and merge.

liuche:
> Will check that pwas and shortcuts still work and will post apks for QA to test before review and merge.

Can you re-flag sebastian for review when these fixes are in? Or if this is no longer necessary, close it.

ekager:
cc @Mugurell could you rebase and make sure CI passes and reflag pocmo for review?

Mugurell:
Sorry for not updating this earlier.
Had discussed with Sebastian and Stefan.
Since this would not fully resolve the issue and would bring some uncertainty regarding future behavior it should not be merged.

PR #8634, opened 2020-02-21
Use same configuration for "forPerformanceTest" builds like we do for "fennec" flavor builds.

pocmo:
@mcomella If I remember correctly this is what we decided to do in Berlin.

With this patch a `forPerformancetest` build will use the same code/configuration as a Fennec / migration build of Fenix.

This allows us to:
* Test the performance of a configuration that ships to our large user base.
* Include the "Do we need to migrate?" code path into our performance testing (which needs to do some checks on app startup).

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -54,6 +54,13 @@ android {
             manifestPlaceholders.isRaptorEnabled = "true"
             applicationIdSuffix ".fenix.performancetest"
             debuggable true
+            manifestPlaceholders = [
+                // Since we configure this build to behave like a "fennec" flavored build, we need
+                // to set a shared user id for the manifest. The actual values does not matter.
+                // However we pick a unique value to not "clash" with other Fenix/Fennec builds
+                // installed on the device.
+                "sharedUserId": "org.mozilla.fenix.performancetest.sharedID"
+            ]
         }
         fenixNightly releaseTemplate >> {
             applicationIdSuffix ".fenix.nightly"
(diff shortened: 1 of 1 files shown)


app/build.gradle
```diff
@@ -54,6 +54,13 @@ android {
             manifestPlaceholders.isRaptorEnabled = "true"
             applicationIdSuffix ".fenix.performancetest"
             debuggable true
+            manifestPlaceholders = [
+                // Since we configure this build to behave like a "fennec" flavored build, we need
+                // to set a shared user id for the manifest. The actual values does not matter.
+                // However we pick a unique value to not "clash" with other Fenix/Fennec builds
+                // installed on the device.
+                "sharedUserId": "org.mozilla.fenix.performancetest.sharedID"
```
mcomella: Will this force devs to uninstall existing installs? This may impact our automated testing so we should send out an FYI if so.

liuche:
@mcomella can we land this? or close it if we don't want it anymore.

mcomella:
We should be able to land this: rerunning CI...

mcomella:
> We should be able to land this

To be explicit, I checked with related parties and everyone seems okay with the change.

mcomella:
Compilation fails because of a SNAPSHOTs related error.

@pocmo Can you rebase on master and merge if it's green?

mcomella:
I tried to rebase myself but I'm unable to test due to: https://github.com/mozilla-mobile/fenix/issues/9598

mcomella:
Dupe https://github.com/mozilla-mobile/fenix/pull/8634 where I rebased onto master.

mcomella:
Correction: dupe https://github.com/mozilla-mobile/fenix/pull/9609.

Issue #7559, opened 2020-01-09
[Bug] changing language does not work if Fenix theme !== device theme


cadeyrn:
### Steps to reproduce

1. open menu > Settings > Language
2. change the language

### Expected behavior

The language of the app changes.

### Actual behavior

The language of the app does not change. Even after removing Fenix from the Android recent apps and starting Fenix again Fenix still has the original language.

I also noted that after going to the language settings again no language has a checkmark.

### Device information

* Android device: OnePlus 7T Pro McLaren Edition / Android 10
* Fenix version: master branch revision 95d4de416dbf00b860e60ccb30fa5e68b6e63793

Amejia481:
On my device, I'm able to change the language but the selected language doesn't have a check-mark,after you go leave the language screen and re-enter. I'm using a pixel 3 with Android 10 

```
Nightly 200109 06:02 (Build #20090616)
28.0.0, 11219297f
GV: 74.0a1-20200107215758
Thursday 1/9 @ 6:15 AM
```

Amejia481:
cc @BranescuMihai

cadeyrn:
I tested again and it's not true that it's not working _at all_ - a few strings are changed but most strings are not.

Some observations:

- The label of the search widget on the Android home screen changes the language.
- In the settings overview screen the the active value of the "Theme" and the "Toolbar" category change the language.
- But the active value of "Enhanced Tracking Protection" and "Delete browsing data on quit" on the same screen does not change the language.
- If you open the "Theme" or the "Toolbar" category there is still the device language used - even if there is another language used in the settings overview screen.
- All other strings seems to be unchanged.

I tested:

a) with German as device language and English as setting in Fenix
b) with English as device language and German as setting in Fenix

cadeyrn:
Some new observations:

- using dark system theme and light Fenix theme: it does not work
- using dark system theme and dark Fenix theme: it does work
- using light system theme and light Fenix theme: it does work
- using light system theme and dark Fenix theme: it does not work

=> So it only works when the system and Fenix theme both use the light or dark theme. It does not work if system and Fenix theme are different.

cadeyrn:
This works now on my device.

AndiAJ:
Hi, **verified as fixed** on the latest **`Nightly Build #20350605 from 2/4`** using the following devices:

• Google Pixel 3a XL (Android 10)
• Google Pixel 3 XL (Android 9)
[...]

PR #8677, opened 2020-02-23
Fixes theme change crashes

boek:


app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -293,7 +294,12 @@ open class FenixApplication : LocaleAwareApplication() {
                 components.core.store,
                 onNewTabOverride = {
                     _, engineSession, url ->
-                        val session = Session(url, components.browsingModeManager.mode.isPrivate)
+                        val shouldCreatePrivateSession =
```
boek: This is the alternative fix to messing with making the `browsingModeManager` a singleton on `components`. Ultimately we need to move towards that solution but there is a bit more we need to think about that we can do post code freeze

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -99,21 +99,6 @@ open class HomeActivity : LocaleAwareAppCompatActivity() {
         )
     }
 
-    private val browsingModeListener = object : BrowsingModeListener {
-        override fun onBrowsingModeChange(newMode: BrowsingMode) {
-            themeManager.currentTheme = newMode
-        }
-    }
-
-    override fun applyOverrideConfiguration(overrideConfiguration: Configuration?) {
```
boek: This is reverted from: https://github.com/mozilla-mobile/fenix/commit/1b7cb117cade9500e0130d7f3177b115ae7bb626#diff-44fc84568dc7e79b2cb8921c2135183a

cadeyrn:
@boek: Since one commit mentions "Reverts For #7559" - Does this mean that the language selection will be broken again or do your changes include another fix for this bug?

boek:
@cadeyrn I'm still working on a patch for that. But I think I might land this to bring the crash rate back down in Nightly

PR #8683, opened 2020-02-24
Implement new feature-tabs API..

pocmo:
We migrated `feature-tabs`, `concept-tabstray` and `browser-tabstraty` to no longer depend on `browser-session`. This also needs a small change in Fenix (Introducing an additional method in `TabsUseCases.SelectTabUseCase`).


pocmo:
New snapshots are currently being built. The first run of this PR may fail.

pocmo:
Okay, new snapshot is available now.

Issue #8540, opened 2020-02-19
[Bug]Disabling all options from Custom ETP still track all your browser


sv-sdeiac:
### Prerequisites
Make sure that ETP from the Firefox settings is set to "Custom" with all options unchecked
![Screenshot_20200219-180736_Firefox Preview](https://user-images.githubusercontent.com/40264165/74850549-ccf86800-5342-11ea-9a90-93cb72e18915.jpg)

## Steps to reproduce
1. Go to any test page for tracking protection(e.g. https://mozilla.github.io/tracking-test/social.html.);

### Expected behavior
A black cat is displayed informing the user that the Tracking Protection is disabled.

### Actual behavior
The Tracking Protection is enabled even if all options from the Custom submenu are disabled.

### Device information

* Android device: Samsung Galaxy S8 (Android 9)
* Fenix version: Firefox Preview Beta 4.0.0-beta.1
* Notes:
- Also, you can't browse if you want to login to a page(e.g. Twitter.com) with all options disabled from the Custom mode of ETP which is a second case that makes this an important issue.
![20200219_181730](https://user-images.githubusercontent.com/40264165/74852006-b2bf8980-5344-11ea-9ef4-da454cbf3b45.gif)

mcarare:
@sv-sdeiac Please retest this in the next Nightly (it should be fixed after fixing https://github.com/mozilla-mobile/fenix/issues/8529)

sv-sdeiac:
Hello, I am still able to reproduce the issue using the latest version of Firefox Preview Nightly 2/20 #20510607.
Now the only ETP who is disabled as in the settings is "Tracking content" and the other 3(Cross-site tracking cookies, fingerprinters, cryptominers) are still blocked by de browser even if in the settings
are disabled.
Device: Samsung Galaxy S8 (Android 9).

liuche:
If we can't land this by EOD, then we should turn this off so we can start Beta Migration.

liuche:
@Baron-Severin 👀

liuche:
Was this a problem due to testing on an outdated site (rather than https://senglehardt.com/test/trackingprotection/test_pages/ )? Could you retest please, because we learned yesterday that the github site does not have the latest lists?

Baron-Severin:
> Was this a problem due to testing on an outdated site (rather than https://senglehardt.com/test/trackingprotection/test_pages/ )? Could you retest please, because we learned yesterday that the github site does not have the latest lists?

[...]

Issue #8594, opened 2020-02-20
[Bug] ETP settings information is incorrect


liuche:
In Settings > Enhanced Tracking Protection menu:

- Standard (Recommended) is not correct (we recommend Strict)
- The "info" about what trackers are blocked in Standard and Strict are the same (but should be different) - this is a regression from Fenix 3.2 release

* Android device: P3
* Fenix version: 2/20 Nightly

![Screenshot_20200220-100348](https://user-images.githubusercontent.com/1082253/74965175-987bbf00-53c9-11ea-9afb-00a152c7b565.png)

softvision-miralobontiu:
Verified as fixed on the latest Nightly build from 2/26 on HTC 10 (Android 8), and Google Pixel (Android 10).

PR #8573, opened 2020-02-20
For #8565: Fix updating dropdown preferences in ETP settings!

mcarare:
### Pull Request checklist

- [x] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [x] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not
- [x] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features. In addition, it includes a screenshot of a successful [accessibility scan](https://play.google.com/store/apps/details?id=com.google.android.apps.accessibility.auditor&hl=en_US) to ensure no new defects are added to the product.

### After merge
- [ ] **Milestone**: Make sure issues finished by this pull request are added to the [milestone](https://github.com/mozilla-mobile/fenix/milestones) of the version currently in development.

### To download an APK when reviewing a PR:
1. click on Show All Checks,
2. click Details next to "Taskcluster (pull_request)" after it appears and then finishes with a green checkmark,
3. click on the "Fenix - assemble" task, then click "Run Artifacts".
4. the APK links should be on the left side of the screen, named for each CPU architecture

Issue #8565, opened 2020-02-20
[Bug]Cookie option from Custom is reset every time you close the ETP panel or enable/disable option


sv-sdeiac:
### Prerequisites
The ETP from the settings is set to "Custom" mode;

## Steps to reproduce
1. Select an option for Cookies in the "Custom" mode;
2 Go to any page(e.g. youtube.com);
3. Tap on the ETP toolbar icon;

### Expected behavior
The Blocked cookie is the same as the selected one in the ETP setting.

### Actual behavior
The Blocked cookies are reset to the: "Cross-Site Tracking Cookie" anytime you close the ETP menu.

### Device information

* Android device: Samsung Galaxy S8 (Android 9).
* Fenix version: 4.0.0-beta.1.
* Also, this issue is reproducible even when you just close the ETP panel and reopen it.
![20200220_090640](https://user-images.githubusercontent.com/40264165/74909669-7b42f280-53c1-11ea-9290-d4a2e1d419f1.gif)

liuche:
If we can't land this by EOD, then we should turn this off so we can start Beta Migration.

liuche:
@Baron-Severin 👀

mcarare:
> If we can't land this by EOD, then we should turn this off so we can start Beta Migration.

https://github.com/mozilla-mobile/fenix/pull/8688 & https://github.com/mozilla-mobile/fenix/pull/8687 should have been merged too.

sv-sdeiac:
Verified that the option made in the ETP settings is saved and anytime you return to the ETP panel, you will see the last saved option for the Custom submenu.

Device: Samsung Galaxy S8 (Android 9).
Build: Firefox Preview Nightly 2/25 @20560606. 

Anyway, I see a difference between the ETP settings and the ETP menu from the toolbar since the cookies blocked it doesn't match, you can find the bug with this issue [here](https://github.com/mozilla-mobile/fenix/issues/8717).

Baron-Severin:
> #8688 & #8687 should have been merged too.

🤦‍♂

PR #8688, opened 2020-02-24
For #8540: Update tracking protection policy on options selection!

mcarare:
### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/Core.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/Core.kt
@@ -25,9 +25,6 @@ import mozilla.components.browser.storage.sync.PlacesBookmarksStorage
 import mozilla.components.browser.storage.sync.PlacesHistoryStorage
 import mozilla.components.concept.engine.DefaultSettings
 import mozilla.components.concept.engine.Engine
-import mozilla.components.concept.engine.EngineSession.TrackingProtectionPolicy
-import mozilla.components.concept.engine.EngineSession.TrackingProtectionPolicy.CookiePolicy
-import mozilla.components.concept.engine.EngineSession.TrackingProtectionPolicy.TrackingCategory
 import mozilla.components.concept.engine.mediaquery.PreferredColorScheme
 import mozilla.components.concept.fetch.Client
 import mozilla.components.feature.customtabs.store.CustomTabsServiceStore
@@ -67,7 +64,7 @@ class Core(private val context: Context) {
             requestInterceptor = AppRequestInterceptor(context),
             remoteDebuggingEnabled = context.settings().isRemoteDebuggingEnabled,
             testingModeEnabled = false,
-            trackingProtectionPolicy = createTrackingProtectionPolicy(),
+            trackingProtectionPolicy = trackingProtectionPolicyFactory.createTrackingProtectionPolicy(),
             historyTrackingDelegate = HistoryDelegate(historyStorage),
             preferredColorScheme = getPreferredColorScheme(),
             automaticFontSizeAdjustment = context.settings().shouldUseAutoSize,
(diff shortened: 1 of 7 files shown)


mcarare:
@Amejia481 The first issue is solved in https://github.com/mozilla-mobile/fenix/pull/8687
The second in: https://github.com/mozilla-mobile/fenix/pull/8573

app/src/main/java/org/mozilla/fenix/components/Core.kt
```diff
@@ -278,13 +278,17 @@ class Core(private val context: Context) {
     }
 
     private fun geCustomCookiePolicy(): CookiePolicy {
-            return when (context.settings().blockCookiesSelectionInCustomTrackingProtection) {
+        return if (context.settings().blockCookiesInCustomTrackingProtection) {
```
Amejia481: I think we need to improve the logic here as we only want to set `CookiePolicy.ACCEPT_ALL` when the **Custom -> Cookies checkbox** is unchecked this is going to return true every time the checkbox is checked, and by default `blockTrackingContentInCustomTrackingProtection` [is set to true](https://github.com/mozilla-mobile/fenix/blob/388c7017d8378500eb986e5473465572055d804a/app/src/main/java/org/mozilla/fenix/utils/Settings.kt#L254)

Amejia481: I think we need to improve the logic here as we only want to set `CookiePolicy.ACCEPT_ALL` when in the UI ETP the **Custom Radio Button** is marked and the **Cookies checkbox** is unchecked this is going to return true every time the checkbox is checked, and by default `blockCookiesInCustomTrackingProtection` [is set to true](https://github.com/mozilla-mobile/fenix/blob/388c7017d8378500eb986e5473465572055d804a/app/src/main/java/org/mozilla/fenix/utils/Settings.kt#L245)

Amejia481: I think we need to improve the logic here as we only want to set `CookiePolicy.ACCEPT_ALL` when in the  ETP UI the **Custom Radio Button** is marked and the **Cookies checkbox** is unchecked this is going to return true every time the checkbox is checked, and by default `blockCookiesInCustomTrackingProtection` [is set to true](https://github.com/mozilla-mobile/fenix/blob/388c7017d8378500eb986e5473465572055d804a/app/src/main/java/org/mozilla/fenix/utils/Settings.kt#L245)

app/src/main/java/org/mozilla/fenix/components/TrackingProtectionPolicyFactory.kt
```diff
@@ -0,0 +1,89 @@
...
+
+package org.mozilla.fenix.components
+
+import mozilla.components.concept.engine.EngineSession.TrackingProtectionPolicy
+import org.mozilla.fenix.utils.Settings
+
+/**
+ * Handles the logic behind creating new [TrackingProtectionPolicy]s.
+ */
+class TrackingProtectionPolicyFactory(private val settings: Settings) {
```
Baron-Severin: Moved the logic into this class to make it easier to test.

app/src/main/java/org/mozilla/fenix/settings/TrackingProtectionFragment.kt
```diff
@@ -213,34 +214,66 @@ class TrackingProtectionFragment : PreferenceFragmentCompat() {
 
         customCookies.onPreferenceChangeListener = object : SharedPreferenceUpdater() {
             override fun onPreferenceChange(preference: Preference, newValue: Any?): Boolean {
-                    updateTrackingProtectionPolicy()
                 customCookiesSelect.isVisible = !customCookies.isChecked
-                return super.onPreferenceChange(preference, newValue)
+                return super.onPreferenceChange(preference, newValue).also {
+                    updateTrackingProtectionPolicy()
```
Baron-Severin: Since `updateTrackingProtectionPolicy` uses information from shared prefs, it's important that we call super (and write to SP) before calling it.

app/src/main/java/org/mozilla/fenix/settings/TrackingProtectionFragment.kt
```diff
@@ -213,34 +214,66 @@ class TrackingProtectionFragment : PreferenceFragmentCompat() {
...
+                }
             }
         }
 
         customCookiesSelect.onPreferenceChangeListener = object : StringSharedPreferenceUpdater() {
             override fun onPreferenceChange(preference: Preference, newValue: Any?): Boolean {
-                updateTrackingProtectionPolicy()
-                val newValueEntry = (preference as DropDownListPreference).findEntry(key = newValue)
-                return super.onPreferenceChange(preference, newValueEntry)
```
Baron-Severin: This whole ETP thing is really complex and hard to follow.  I think we were mistaken before @mcarare, this is working fine without any of the preference key/value mapping.

Baron-Severin:
@mcarare would you mind looking this over?  I've tested and https://github.com/mozilla-mobile/fenix/issues/8594, https://github.com/mozilla-mobile/fenix/issues/8540, and https://github.com/mozilla-mobile/fenix/issues/8565 all seem to be working properly, but it's been a long day and I could have missed something.

Be aware that social trackers won't be blocked in custom mode.  There's already a patch to fix it here https://github.com/mozilla-mobile/android-components/pull/6080/.

app/src/main/java/org/mozilla/fenix/settings/TrackingProtectionFragment.kt
mcarare: Yes, I think it handles that internally. The initial problem was using `SharedPreferenceUpdater` that only handles booleans, not string. With the new `StringSharedPreferenceUpdater` it works without the mapping.

app/src/main/java/org/mozilla/fenix/settings/TrackingProtectionFragment.kt
```diff
@@ -103,16 +103,18 @@ class TrackingProtectionFragment : PreferenceFragmentCompat() {
             getString(R.string.preference_enhanced_tracking_protection_strict_info_button)
         radioStrict.onPreferenceChangeListener = object : SharedPreferenceUpdater() {
             override fun onPreferenceChange(preference: Preference, newValue: Any?): Boolean {
-                if (newValue == true) {
-                    updateTrackingProtectionPolicy()
-                    context?.metrics?.track(
-                        Event.TrackingProtectionSettingChanged(
-                            Event.TrackingProtectionSettingChanged.Setting.STRICT
+                return super.onPreferenceChange(preference, newValue).also {
```
mcarare: @Baron-Severin @Amejia481  I moved  ` updateTrackingProtectionPolicy()` after the preference change, like Severin recommended for the checkboxes. Also, I've debugged this and with previous setting all the other preference keys remained true even if visually the radioButton was unchecked. That is why I added the `setCheckedWithoutClickListener` call. Any thoughts?

app/src/main/java/org/mozilla/fenix/components/TrackingProtectionPolicyFactory.kt
```diff
@@ -0,0 +1,89 @@
...
+     */
+    @Suppress("ComplexMethod")
+    fun createTrackingProtectionPolicy(
+        normalMode: Boolean = settings.shouldUseTrackingProtection,
+        privateMode: Boolean = true
+    ): TrackingProtectionPolicy {
+        val trackingProtectionPolicy =
+            when {
+                settings.useStrictTrackingProtection -> TrackingProtectionPolicy.strict()
+                settings.useCustomTrackingProtection -> return TrackingProtectionPolicy.select(
```
Amejia481: nit: Could we  extract the custom tracking protection policy into a new function

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -247,9 +247,10 @@ class Settings private constructor(
         appContext.getPreferenceKey(R.string.pref_key_tracking_protection_custom_cookies),
         true
     )
+
     val blockCookiesSelectionInCustomTrackingProtection by stringPreference(
         appContext.getPreferenceKey(R.string.pref_key_tracking_protection_custom_cookies_select),
-        ""
+        "all"
```
Amejia481: "all" is to aggressive to be the default we should have "social" -> "Cross-site and Social media trackers" as default following the same order as the spinner

Amejia481: Also I noticed in desktop if we switch policies like from **Custom** to **Strict** or **Standard** the custom values return to the default values, I think we need to include this functionality, as the user could get a lot of site breakage if we keep the previous values as the user could think that is our default values

mcarare: > Also I noticed in desktop if we switch policies like from **Custom** to **Strict** or **Standard** the custom values return to the default values, I think we need to include this functionality, as the user could get a lot of site breakage if we keep the previous values as the user could think that is our default values

We've discussed this and this behaviour remains TBD. Thanks @Amejia481 😄

Amejia481: 👍

ekager:
Looks like there are some extraneous commits here. Could we rebase and push before merging?

Baron-Severin:
> Looks like there are some extraneous commits here. Could we rebase and push before merging?

Removed it then squashed.

PR #8704, opened 2020-02-24
 No issue: Fixes swipeRefresh UI test crash

sblatz:
### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
@@ -475,7 +475,10 @@ abstract class BaseBrowserFragment : Fragment(), UserInteractionHandler, Session
     private fun adjustBackgroundAndNavigate(directions: NavDirections) {
         context?.let {
             engineView.captureThumbnail { bitmap ->
-                lifecycleScope.launch {
+                viewLifecycleOwner.lifecycleScope.launch {
+                    // isAdded check is necessary because of a bug in viewLifecycleOwner. See AC#3828
+                    if (!this@BaseBrowserFragment.isAdded) return@launch
+
                     // If the bitmap is null, the best we can do to reduce the flash is set transparent
                     swipeRefresh.background = bitmap?.toDrawable(it.resources)
                         ?: ColorDrawable(Color.TRANSPARENT)


app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
```diff
@@ -477,8 +477,10 @@ abstract class BaseBrowserFragment : Fragment(), UserInteractionHandler, Session
             engineView.captureThumbnail { bitmap ->
                 lifecycleScope.launch {
                     // If the bitmap is null, the best we can do to reduce the flash is set transparent
-                    swipeRefresh.background = bitmap?.toDrawable(it.resources)
-                        ?: ColorDrawable(Color.TRANSPARENT)
+                    if (swipeRefresh != null) {
+                        swipeRefresh.background = bitmap?.toDrawable(it.resources)
+                            ?: ColorDrawable(Color.TRANSPARENT)
```
Baron-Severin: I think the fact that we're crashing here is important, and this null check fixes a symptom and not the real cause.  If we're accessing the `swipeRefresh` when it's null it means something about our scope is wrong.  In this case I'm guessing it's that we're using the fragment lifecycle instead of the fragment view lifecycle, allowing us to access the views before `onCreateView` and after `onDestroyView`.  Line 478 should be `viewLifecycleOwner.lifecycleScope.launch {`.

However!  There's currently a bug in `viewLifecycleOwner` (see https://github.com/mozilla-mobile/android-components/issues/3828), so we need to add an additional check here.  The final code should look something like:

```kotlin
viewLifecycleOwner.lifecycleScope.launch {
    // isAdded check is necessary because of a bug in viewLifecycleOwner. See AC#3828
    if (!this@BaseBrowserFragment.isAdded) return@launch
    // If the bitmap is null, the best we can do to reduce the flash is set transparent
    swipeRefresh.background = bitmap?.toDrawable(it.resources)
        ?: ColorDrawable(Color.TRANSPARENT)

    ...
```

sblatz: Ah, wasn't aware of this bug. Thanks :) let's try this.

Issue #8637, opened 2020-02-21
[Bug] Crash when opening link from custom tab in Fenix (open links in private previously set)


sv-ohorvath:
## Steps to reproduce
1. Install the Nightly build from 2/19.
2. Set Nightly as default browser.
3. Set open links in private On.
4. Update to the latest version from PlayStore (2/21).
5. Leave Fenix open in the background.
6. Open Gmail, open a custom tab, tap Open link in Firefox Nightly.
### Expected behavior
The link is opened in Fenix.
### Actual behavior
App crashes. Logcat: 
[open link crash log.txt](https://github.com/mozilla-mobile/fenix/files/4236659/open.link.crash.log.txt)

* If you clear app data or uninstall the previous build before installing the 2/21 build, the crash doesn't happen.
### Device information

* Android device: Pixel 3 (Android 10), Samsung Galaxy S8 (Android 9)
* Fenix version: Nightly 2/21

sv-ohorvath:
Actually, it also happens on Beta 4.0-build1 by setting Fenix as default browser and Open link in private on.

ekager:
> 2020-02-21 16:14:40.573 25454-25454/? E/AndroidRuntime: FATAL EXCEPTION: main
    Process: org.mozilla.fenix.nightly, PID: 25454
    java.lang.IllegalStateException: FragmentManager has been destroyed
        at androidx.fragment.app.FragmentManager.ensureExecReady(FragmentManager.java:4)
        at androidx.fragment.app.FragmentManager.execPendingActions(FragmentManager.java:1)
        at androidx.fragment.app.FragmentManager$4.run(FragmentManager.java:1)
        at android.os.Handler.handleCallback(Handler.java:883)
        at android.os.Handler.dispatchMessage(Handler.java:100)
        at android.os.Looper.loop(Looper.java:214)
        at android.app.ActivityThread.main(ActivityThread.java:7356)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:492)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:930)

ekager:
Probably related to https://github.com/mozilla-mobile/fenix/issues/8580

sv-ohorvath:
@ekager Still happening on Nightly 2/24 after #8580 was fixed.

pocmo:
[...]

PR #8709, opened 2020-02-24
🎲 For #8637: Remove setActivityTheme outside activity lifecycle scope

psymoon:
Context: having activity instance referenced in outside object is generally fragile because activities get recreated all the time so the instance referenced may no longer be relevant.

Not only does `setActivityTheme(activity)` whenever the mode changes is dangerous for single activity framework, but also for deep link cases (i.e. custom tab) where there could be multiple activities. This call is also completely redundant and trivial because we recreate the activity shortly thereafter and would be correctly handled in onCreate (https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/java/org/mozilla/fenix/HomeActivity.kt#L239)


boek:
4th time is the charm??

PR #8733, opened 2020-02-25
Re-enable ETP UI tests and update local asset

### LINKED ISSUE #8732: Re-enable and fix ETP UI tests
ETP UI tests were disabled in https://github.com/mozilla-mobile/fenix/issues/7907 due to first-launch load Gecko not loading Tracker listings quick enough. Swapped the test to use `trackertest.org` (special test trackers that are loaded earlier). Let's see if these help these tests stay green.

AaronMT:
Closes #8732

### DIFF
--- a/app/src/androidTest/assets/pages/etp.html
+++ b/app/src/androidTest/assets/pages/etp.html
@@ -1,41 +1,18 @@
-<!-- https://github.com/mozilla/dummytracker -->
-<!-- https://github.com/englehardt/englehardt.github.io/tree/master/test/trackingprotection/test_pages -->
-<html>
+<!DOCTYPE HTML>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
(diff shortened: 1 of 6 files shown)


AaronMT:
Status didn't bubble up but a re-enabled test hit the crash in https://github.com/mozilla-mobile/fenix/pull/8704

AaronMT:
For whatever reason, on a Nexus 6 (of which is not reproducible in an equivalent AVD), in `testStrictVisitDisable`, the `click` to open Settings in the menu is not happening. I'm trying multiple approaches.

AaronMT:
> 02-26 17:32:11.738: I/ViewInteraction(8647): Checking 'MatchesViewAssertion{viewMatcher=at least 100 percent of the view's area is displayed to the user.}' assertion on view (with res-name that is "text" and with string from resource id: <2131886186>[browser_menu_settings] value: Settings)

>02-26 17:32:11.741: I/ViewInteraction(8647): Performing 'single click' action on view (with res-name that is "text" and with string from resource id: <2131886186>[browser_menu_settings] value: Settings)

> 02-26 17:32:11.794: D/App(8647): DebugMetricController: track event: BrowserMenuItemTapped(item=SETTINGS)

The video shows no click on settings. I'll have to investigate further. Or just keep this disabled on this device.

Issue #8699, opened 2020-02-24
[Bug] Crash on rapid selection in addons menu


s-ankur:
## Steps to reproduce
open addons menu, ublock origin
click on the disable addon toggle
quickly click on any other option

### Expected behavior
No crash
### Actual behavior
Crash
### Device information

* Android device: Asus Zenfone
* Fenix version: Fenix Nightly

AndiAJ:
Hi @s-ankur , I wasn't able to reproduce this issue on the latest **`Nightly Build #20560606`** from 02/25 using the following devices:
• Google Pixel 3a XL (Android 10)
• Google Pixel 3a (Android 9)
• Huawei Mate 20 Lite (Android 8.1.0)
• Samsung Galaxy S7 (Android 7)
• OnePlus A3 (Android 6.0.1)

In case it you are still encountering this problem could you please help us out with a crashlog? ☺️

s-ankur:
Yes , its not reproducible in Nightly anymore

s-ankur:
Ok wait, its still there. You have to remove addon instead of disabling it

AndiAJ:
Thx @s-ankur ! ☺️ 

I've managed to reproduce the issue on the latest **`Nightly Build #20570604 from 02/26`** using the following devices:
• Google Pixel 3a (Android 9)
• OnePlus A3 (Android 6.0.1)

The STR that worked for me:
1. Right after installation, tap and open uBlock
2. Remove and quickly press "Details"

► Video
![20200226-170030](https://user-images.githubusercontent.com/51314259/75357859-b6aa5900-58ba-11ea-9abc-3a5b74484f43.gif)

```kotlin
 --------- beginning of crash
2020-02-26 17:00:41.047 4197-4197/? E/AndroidRuntime: FATAL EXCEPTION: main
    Process: org.mozilla.fenix.nightly, PID: 4197
    org.mozilla.geckoview.GeckoResult$UncaughtException: java.lang.IllegalStateException: enable_switch must not be null
        at org.mozilla.geckoview.GeckoResult.dispatchLocked(GeckoResult.java:4)
        at org.mozilla.geckoview.GeckoResult.completeExceptionally(GeckoResult.java:4)
        at org.mozilla.geckoview.GeckoResult.lambda$thenInternal$2$GeckoResult(GeckoResult.java:10)
[...]

PR #8776, opened 2020-02-26
For #8699: Add setAllInteractiveViewsClickable for add-on pending removal

psymoon:


app/src/main/java/org/mozilla/fenix/addons/InstalledAddonDetailsFragment.kt
```diff
@@ -217,7 +218,7 @@ class InstalledAddonDetailsFragment : Fragment() {
                 },
                 onError = { _, _ ->
                     runIfFragmentIsAttached {
-                        enable_switch.isClickable = true
+                        setAllInteractiveViewsClickable(true)
```
Amejia481:
```suggestion
                        setAllInteractiveViewsClickable(view,true)
```

app/src/main/java/org/mozilla/fenix/addons/InstalledAddonDetailsFragment.kt
```diff
@@ -231,6 +232,13 @@ class InstalledAddonDetailsFragment : Fragment() {
         }
     }
 
+    private fun setAllInteractiveViewsClickable(clickable: Boolean) {
+        enable_switch.isClickable = clickable
+        details.isClickable = clickable
+        permissions.isClickable = clickable
+        remove_add_on.isClickable = clickable
+    }
```
Amejia481: Could we use view to call this references, as some of them get execute inside `runIfFragmentIsAttached` we could get a null-pointer exception in some cases

```suggestion
    private fun setAllInteractiveViewsClickable(view: View,clickable: Boolean) {
        view.enable_switch.isClickable = clickable
        view.details.isClickable = clickable
        view.permissions.isClickable = clickable
        view.remove_add_on.isClickable = clickable
    }
```

PR #8823, opened 2020-02-27
Simplify intent handling and handle manifest file not existing in FennecWebAppIntentProcessor

pocmo:
This series of patches addresses https://github.com/mozilla-mobile/android-components/issues/5974.

The first patch simplifies the Intent handling in IntentReceiverActivity and removes the need to process() and then match() on every intent processor. A successful process() means that we already matched. Doing this in two steps caused issues when the process() failed but later the simple check in match() returned true.

The second patch makes FennecWebAppIntentProcessor handle a situation where we can't read a manifest file. In that situation we open the URL just as a custom tab.


boek:
@pocmo does this need to be uplifted?

pocmo:
@boek Yeah, I think that would be a good candidate if we end up shipping a new build. Would you like me to create a PR for merging into the release branch?

PR #8835, opened 2020-02-27
 For #167: Improves home to browser animation 

sblatz:
### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
@@ -6,8 +6,6 @@ package org.mozilla.fenix.browser
 
 import android.content.Context
 import android.content.Intent
-import android.graphics.Color
-import android.graphics.drawable.ColorDrawable
 import android.os.Bundle
 import android.view.Gravity
 import android.view.LayoutInflater
@@ -15,11 +13,9 @@ import android.view.View
 import android.view.ViewGroup
 import androidx.annotation.CallSuper
 import androidx.coordinatorlayout.widget.CoordinatorLayout
-import androidx.core.graphics.drawable.toDrawable
 import androidx.core.net.toUri
 import androidx.fragment.app.Fragment
 import androidx.lifecycle.lifecycleScope
-import androidx.navigation.NavDirections
 import androidx.navigation.fragment.findNavController
 import com.google.android.material.appbar.AppBarLayout
 import com.google.android.material.snackbar.Snackbar
@@ -89,6 +85,7 @@ import org.mozilla.fenix.ext.sessionsOfType
 import org.mozilla.fenix.ext.settings
 import org.mozilla.fenix.settings.SupportUtils
 import org.mozilla.fenix.theme.ThemeManager
+import java.lang.ref.WeakReference
 
 /**
  * Base fragment extended by [BrowserFragment].
@@ -98,6 +95,7 @@ import org.mozilla.fenix.theme.ThemeManager
 @Suppress("TooManyFunctions", "LargeClass")
 abstract class BaseBrowserFragment : Fragment(), UserInteractionHandler, SessionManager.Observer {
     protected lateinit var browserFragmentStore: BrowserFragmentStore
+    private lateinit var browserAnimator: BrowserAnimator
 
     private var _browserInteractor: BrowserToolbarViewInteractor? = null
     protected val browserInteractor: BrowserToolbarViewInteractor
@@ -164,6 +162,15 @@ abstract class BaseBrowserFragment : Fragment(), UserInteractionHandler, Session
 
         initializeEngineView(toolbarHeight)
 
+        browserAnimator = BrowserAnimator(
+            fragment = WeakReference(this),
+            engineView = WeakReference(engineView),
+            swipeRefresh = WeakReference(swipeRefresh),
+            arguments = arguments!!
+        ).apply {
+            beginAnimationIfNecessary()
+        }
+
         return getSessionById()?.also { session ->
             val browserToolbarController = DefaultBrowserToolbarController(
                 store = browserFragmentStore,
@@ -177,10 +184,9 @@ abstract class BaseBrowserFragment : Fragment(), UserInteractionHandler, Session
                 browsingModeManager = (activity as HomeActivity).browsingModeManager,
                 sessionManager = requireComponents.core.sessionManager,
                 findInPageLauncher = { findInPageIntegration.withFeature { it.launch() } },
-                browserLayout = view.browserLayout,
                 engineView = engineView,
                 swipeRefresh = swipeRefresh,
-                adjustBackgroundAndNavigate = ::adjustBackgroundAndNavigate,
+                browserAnimator = browserAnimator,
                 customTabSession = customTabSessionId?.let { sessionManager.findSessionById(it) },
                 getSupportUrl = {
                     SupportUtils.getSumoURLForTopic(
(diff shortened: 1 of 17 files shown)


app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
```diff
@@ -147,6 +150,25 @@ abstract class BaseBrowserFragment : Fragment(), UserInteractionHandler, Session
         browserInitialized = initializeUI(view) != null
     }
 
+    private fun animateBrowserEngine(browserEngine: View) {
```
Baron-Severin: I think this might be easier to read if `animateBrowserEngine` and `captureEngineViewAndDrawStatically` were moved to a new class, and they both seem relatively easy to move.  It would let us avoid bloating the fragment, group the constants a little more logically, etc.

app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarController.kt
```diff
@@ -89,12 +82,14 @@ class DefaultBrowserToolbarController(
     internal var ioScope: CoroutineScope = CoroutineScope(Dispatchers.IO)
 
     override fun handleToolbarPaste(text: String) {
-        adjustBackgroundAndNavigate.invoke(
-            BrowserFragmentDirections.actionBrowserFragmentToSearchFragment(
+        captureEngineViewAndDrawStatically.invoke {
+            val directions = BrowserFragmentDirections.actionBrowserFragmentToSearchFragment(
```
Baron-Severin: This is conceptually fine, but I find it a little difficult to read as is.  It would be nice if this could look more like

`captureEngineViewAndDrawStatically.andThen { nav }`, or `captureEngineViewAndDrawStatically.doOnEnd { nav }`

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -269,6 +265,10 @@ class HomeFragment : Fragment() {
                 sessionControlView!!.view.layoutManager?.onRestoreInstanceState(parcelable)
             }
             homeViewModel.layoutManagerState = null
+
+            // We have to delay so that the keyboard collapses and the view is resized before the
+            // animation from SearchFragment happens
+            delay(ANIMATION_DELAY)
```
Baron-Severin: Good comment, that would be very confusing without it.

app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
```diff
@@ -64,7 +64,7 @@ class BrowserFragment : BaseBrowserFragment(), UserInteractionHandler {
         savedInstanceState: Bundle?
     ): View {
         val view = super.onCreateView(inflater, container, savedInstanceState)
-        view.browserLayout.transitionName = "$TAB_ITEM_TRANSITION_NAME${getSessionById()?.id}"
+
         startPostponedEnterTransition()
```
ekager: Do we still need the postponeEnterTransition and startPostponedEnterTransition if we're removing the shared element transition from browser?

app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
ekager: I agree these should be moved to a new class out of the fragment. 

Also in theory these shouldn't be needed in `BaseBrowserFragment` only `BrowserFragment` because we will only animate in normal browsing not custom tabs, right? Not sure if the timing works out the same, but if nothing else we should remove any delays or things that would impact custom tab or action view intent performance and are unneeded 🤷‍♀️

sblatz: I have refactored this logic all into `BrowserAnimator.kt`. 

@ekager I actually do need to have the animation in BaseBrowserFragment because I pass the "captureEngineView" logic to the browserToolbar so it can play it as needed (to prevent "flickering" of GV). That being said, I am **only delaying** if the newly added `shouldAnimate` flag is set (which defaults to false) and is only set on Home -> Browser transition. This will avoid there being any delays in custom tabs or on resuming the fragment 😄

app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
sblatz: @ekager Potentially. I don't need it for the animations I'm doing right now, but I'm going to have a follow up PR that moves the browser toolbar up with Browser -> Search transition. So I'd like to keep it for now.

app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
```diff
@@ -164,6 +163,16 @@ abstract class BaseBrowserFragment : Fragment(), UserInteractionHandler, Session
 
         initializeEngineView(toolbarHeight)
 
+        browserAnimator = BrowserAnimator(
+            fragment = WeakReference(this),
+            viewLifeCycleScope = viewLifecycleOwner.lifecycleScope,
```
Baron-Severin: Nit: since you're passing the fragment, don't need to pass the scope as well.

app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
```diff
@@ -164,6 +163,16 @@ abstract class BaseBrowserFragment : Fragment(), UserInteractionHandler, Session
...
+            engineView = engineView,
+            swipeRefresh = swipeRefresh,
```
Baron-Severin: If we're taking the `WeakReference` route here, we'll need to wrap these two views as well.

app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
```diff
@@ -177,10 +186,9 @@ abstract class BaseBrowserFragment : Fragment(), UserInteractionHandler, Session
                 browsingModeManager = (activity as HomeActivity).browsingModeManager,
                 sessionManager = requireComponents.core.sessionManager,
                 findInPageLauncher = { findInPageIntegration.withFeature { it.launch() } },
-                browserLayout = view.browserLayout,
                 engineView = engineView,
                 swipeRefresh = swipeRefresh,
-                adjustBackgroundAndNavigate = ::adjustBackgroundAndNavigate,
+                captureEngineViewAndDrawStatically = browserAnimator::captureEngineViewAndDrawStatically,
```
Baron-Severin: Optional nit: I think I would personally just pass the animator here, but it's not a strong preference.  Up to you.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlController.kt
```diff
@@ -341,27 +340,19 @@ class DefaultSessionControlController(
...
-        val directions = HomeFragmentDirections.actionHomeFragmentToBrowserFragment(null)
-        val extras =
-            FragmentNavigator.Extras.Builder()
-                .addSharedElement(
-                    tabView,
-                    "$TAB_ITEM_TRANSITION_NAME$sessionId"
-                )
-                .build()
-        navController.nav(R.id.homeFragment, directions, extras)
+        (activity).openToBrowser(BrowserDirection.FromHome)
```
ekager: nit - can we just write `activity`?

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlController.kt
```diff
@@ -341,27 +340,19 @@ class DefaultSessionControlController(
...
-        navController.nav(
-            R.id.homeFragment,
-            HomeFragmentDirections.actionHomeFragmentToBrowserFragment(null)
+        if (url == SupportUtils.POCKET_TRENDING_URL) { metrics.track(Event.PocketTopSiteClicked) }
+        activity.components.useCases.tabsUseCases.addTab.invoke(
+            url = url,
+            selectTab = true,
+            startLoading = true
         )
+        (activity).openToBrowser(BrowserDirection.FromHome)
```
ekager: same nit here

ekager:
Just did one more run through and found 2 nits, but everything else looks good to me!

PR #8872, opened 2020-02-28
Closes #6730: Lazily initialize account manager on new push message

### LINKED ISSUE #6730: Delay initialization of Push
Perf folks: before working on this issue, check in with Jonathan on the status: this issue may be worked on by Jonathan and it may also be blocked on other changes.

Priority: 5-10% of startup spent here. Last big chunk of perf wins before jumping into view inflation.

---

Some thoughts from a conversation with @hawkinsw from reviewing https://github.com/mozilla-mobile/fenix/pull/6712/.

Since the consumers for Push are FxaAccountManager and Engine, we were initializing push immediately since that was required by the Firebase SDK.

What Will rightfully noticed, is that maybe we can delay this initialization to be done much later since the consumers are not lazily initialized through UI interactions or activity creation.

This needs investigations (mostly from me) before we do this:
 - Where can be move to be initialized to that isn't immediately on startup, but soon enough to receive push messages on "startup".
 - Can we safely do this since the black box is the Firebase SDK which assumes initialization is always done in `Application#onCreate`.
 - Will we cause any race conditions from our initialization, i.e. Push init'd and receives a message before our consumers have init'd.

jonalmeida:
Weakly depends on https://github.com/mozilla-mobile/android-components/issues/6125, but we can be merged this even though it will have no effect until it's linked PR is fixed.

### Pull Request checklist

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
+++ b/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
@@ -36,6 +36,7 @@ import mozilla.components.support.webextensions.WebExtensionSupport
 import org.mozilla.fenix.components.Components
 import org.mozilla.fenix.components.metrics.MetricServiceType
 import org.mozilla.fenix.ext.settings
+import org.mozilla.fenix.push.PushFxaIntegration
 import org.mozilla.fenix.session.NotificationSessionObserver
 import org.mozilla.fenix.session.PerformanceActivityLifecycleCallbacks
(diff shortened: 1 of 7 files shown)


jonalmeida:
Tests can probably be written for this..

jonalmeida:
Will write tests for this in a follow-up.

jonalmeida:
Rebased since it seems there were breaking APIs and conflicts from where I was at last.

Issue #8412, opened 2020-02-13
[Crash] ActivityNotFound


boek:
```
android.content.ActivityNotFoundException: 
 
  at android.app.Instrumentation.checkStartActivityResult (Instrumentation.java:2018)
 
  at android.app.Instrumentation.execStartActivity (Instrumentation.java:1673)
 
  at android.app.Activity.startActivityForResult (Activity.java:4688)
 
  at androidx.fragment.app.FragmentActivity.startActivityForResult (FragmentActivity.java:6)
 
  at android.app.Activity.startActivityForResult (Activity.java:4646)
 
  at androidx.fragment.app.FragmentActivity.startActivityForResult (FragmentActivity.java:3)
 
  at android.app.Activity.startActivity (Activity.java:5007)
 
  at androidx.core.app.ActivityCompat.startActivity (ActivityCompat.java:2)
 
  at mozilla.components.feature.customtabs.CustomTabWindowFeature$start$1$invokeSuspend$$inlined$collect$1.emit (Collect.kt:14)
 
  at mozilla.components.support.ktx.kotlinx.coroutines.flow.FlowKt$ifChanged$$inlined$filter$1$2.emit (Collect.kt:7)
 
  at mozilla.components.feature.customtabs.CustomTabWindowFeature$start$1$invokeSuspend$$inlined$mapNotNull$1$2.emit (Collect.kt:6)
 
  at kotlinx.coroutines.EventLoopKt.emitAll (EventLoop.kt:15)
 
  at kotlinx.coroutines.flow.FlowKt__ChannelsKt$emitAll$1.invokeSuspend (Unknown Source:10)
 
  at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith (ContinuationImpl.kt:2)
 
  at kotlinx.coroutines.DispatchedTask.run (DispatchedTask.kt:15)
 
  at android.os.Handler.handleCallback (Handler.java:873)
 
  at android.os.Handler.dispatchMessage (Handler.java:99)
 
  at android.os.Looper.loop (Looper.java:214)
 
  at android.app.ActivityThread.main (ActivityThread.java:6986)
 
  at java.lang.reflect.Method.invoke (Native Method)
 
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:494)
 
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:1445)
```
[...]

PR #8903, opened 2020-03-02
For #8412: Passes error handling function to 'CustomTabWindowFeature'

ValentinTimisica:
Change required for showing error message when the app can't handle a specific
scheme. Implemented in AC:
https://github.com/mozilla-mobile/android-components/pull/6122

### Pull Request checklist

- [x] **Tests**: This PR does not include thorough tests.
- [x] **Screenshots**: This PR includes a screenshot of the changes made.

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/customtabs/ExternalAppBrowserFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/customtabs/ExternalAppBrowserFragment.kt
@@ -33,6 +33,7 @@ import org.mozilla.fenix.R
 import org.mozilla.fenix.browser.BaseBrowserFragment
 import org.mozilla.fenix.browser.CustomTabContextMenuCandidate
 import org.mozilla.fenix.browser.FenixSnackbarDelegate
+import org.mozilla.fenix.components.FenixSnackbar
 import org.mozilla.fenix.ext.components
 import org.mozilla.fenix.ext.nav
 import org.mozilla.fenix.ext.requireComponents
@@ -50,7 +51,7 @@ class ExternalAppBrowserFragment : BaseBrowserFragment(), UserInteractionHandler
     private val windowFeature = ViewBoundFeatureWrapper<CustomTabWindowFeature>()
     private val hideToolbarFeature = ViewBoundFeatureWrapper<WebAppHideToolbarFeature>()
 
-    @Suppress("LongMethod")
+    @Suppress("LongMethod", "ComplexMethod")
     override fun initializeUI(view: View): Session? {
         return super.initializeUI(view)?.also {
             val activity = requireActivity()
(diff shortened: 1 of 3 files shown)


ValentinTimisica:
Should be merged after the [AC issue](https://github.com/mozilla-mobile/android-components/issues/6094) change will be merged and available in Fenix. Also I need feedback from UX for the string used.

betsymi:
I can help with this string. Have a few clarifying questions that will help:

- What's a _specific scheme?_ I'm not familiar with that term. 
- Do I understand the action that triggered this correctly: User tried to open a link they saw in Firefox Preview in a different app?
- Anything the user can do to resolve this or not really? Try again?

ValentinTimisica:
1) By "when the app can't handle a specific scheme" I mean that the app can't handle any scheme, but just `http` and `https`. So when a user will access an URL with let's say `tel` scheme, the app will crash. The most common form of URI is the URL (web address). Each URI begins with a scheme name:
URI = scheme:[//authority]path[?query][#fragment].
Examples of popular schemes include http, https, ftp, mailto, file, data etc.
For more informations about this you can visit [a].
2) The crash happens in customTabs when a user opens a link (from the customTab) that has a scheme which cannot be handled by the app (we are currently only registered to handle `http` and `https` schemes). You can find the video with the crash in [b].
3) No, the user can't resolve this. The only way to prevent such issues is to handle other more common URI schemes in the future (but for now we have to fix the crash).

[a] https://en.wikipedia.org/wiki/Uniform_Resource_Identifier
[b] https://drive.google.com/file/d/1DzMHFijnITOKUL5TSoIWWpmEdmMLyHxE/view

betsymi:
There is some debate in UX if a snackbar is the appropriate UI element to display this message. Snackbars are typically to confirm that a user action has been taken, and they quickly disappear. 

If we were to propose a different element for this message and new string, when would you need that by?

ValentinTimisica:
As far as I know today (03/06) is the last day before code freeze. So I need that as soon as you can. Thanks!

betsymi:
Here's the string you can use `Unable to connect. Unrecognizable URL scheme.`

Longer term, we may like to explore a different UI element to surface this message that the user can dismiss. Snackbars are intended to be used as confirmation based on a user action (delete tab, save boomark, create collection, etc.) and disappear a few seconds later.

ValentinTimisica:
Thanks @betsymi!

liuche:
Hey @ValentinTimisica, this looks like it needs to be rebased, or is failing the tests for some other reason. Can you get this PR passing tests so it can be reviewed?

ValentinTimisica:
Hello @liuche, the build is failing because I modified the public API in the [AC PR](https://github.com/mozilla-mobile/android-components/pull/6122) and it is not merged yet. After merge it will also take a while until it will be available in Fenix at which point I will rebase this.
In the meantime someone can review this patch but shouldn't merge it yet.

liuche:
Looks like this just merged, and should be in the next AC Nightly release! https://github.com/mozilla-mobile/android-components/issues/6094

Issue #7863, opened 2020-01-22
Ensure parent process triggers speculative connection when possible.


fluffyemily:
[Originally raised as bug against GeckoView. ](https://bugzilla.mozilla.org/show_bug.cgi?id=1603853)

Perhaps Fenix has already some speculative connection enabled. (and if so, just close this. I couldn't find such code immediately.)
But if not, it probably should have.

https://bugzilla.mozilla.org/show_bug.cgi?id=1383299 is where the stuff was added to desktop.

`GeckoWebExecutor` in GeckoView provides an API to create speculative connections that Fenix can implement.

mcomella:
@fluffyemily In practice, what would an implementation of this look like? e.g. do we speculatively connect while the user types URL in the URL bar? 

I'm not sure how to prioritize this yet.

mcomella:
Triage from Christian: we already do speculative connections for applink. However, we can introduce this when typing in the URL bar – Fennec already does this and we could copy their implementation for a page load win.

mcomella:
Adding group-triage: not sure if we have the cycles to address this short term but it could be a good page load win so consulting Fenix triage.

csadilek:
See: 
https://github.com/mozilla-mobile/android-components/issues/1912
https://searchfox.org/mozilla-esr68/source/mobile/android/base/java/org/mozilla/gecko/home/BrowserSearch.java#522-525
https://searchfox.org/mozilla-esr68/source/mobile/android/base/java/org/mozilla/gecko/home/BrowserSearch.java#623-627
https://searchfox.org/mozilla-esr68/source/mobile/android/chrome/content/browser.js#8029-8050

We will look into this in an upcoming A-C sprint. Marked for triage in A-C.

acreskeyMoz:
FYI, I have data showing that the applink speculative connect that Fenix makes (via `RemoteWebNavigation.jsm) is very effective.

Compare `fenix_intent`, to `fenix_intent_no_spec`
https://docs.google.com/spreadsheets/d/133a2JzAw4nD6gWplm1X5m0ulbEX24ySgHPzh2mCEtc4/edit#gid=433931785&range=466:492

More discussion in this bug:
https://bugzilla.mozilla.org/show_bug.cgi?id=1612575

csadilek:
Fenix now (since Nightly 200303 18:01) triggers speculative connects when:
- An intent is received (this was working before)
[...]

PR #8909, opened 2020-03-02
Issue #7863: Ensure parent process triggers speculative connect

csadilek:
Small change to the views to make sure we pass along the engine for speculative connects in the toolbar and awesomebar. We only want to issue speculative connect calls for regular (!private) tabs.

We already have speculative connect calls for custom tabs and when opening links through intents.

This needs https://github.com/mozilla-mobile/android-components/pull/6111 to land.


pocmo:
Landed the AC patch and retriggered snapshots. There will be another breaking API change in the snapshot. I'll update this PR with the required change and then will take care of landing everything together.

pocmo:
Updated PR.

Issue #8859, opened 2020-02-28
[Bug]The tab drawer number is not centered when open Top Sites pages


sv-sdeiac:
## Steps to reproduce
1. Open the Homepage of the app;
2. Tap on a Top Site page;
3. Observe the tab drawer number.

### Expected behavior
The number is displayed without any issue.

### Actual behavior
The number is not centered.

### Device information

* Android device: Samsung Galaxy Note 8 (Android 9).
* Fenix version: Firefox Preview Nightly 2/28 #20590606.
![20200228_165903](https://user-images.githubusercontent.com/40264165/75559223-c0af9180-5a4b-11ea-91d1-4a3e7028820e.gif)

darkwing:
We should revisit this when @gabrielluong has https://github.com/mozilla-mobile/fenix/issues/8791 ready.

sblatz:
This is a pretty annoying visual bug and is not isolated to top sites.

cadeyrn:
related: #8859.

sv-sdeiac:
Verified as fixed using the latest version of Firefox Preview Nightly 3/3 @20630607).

Devices:
- Nokia 6 (Android 7.1.1);
- Samsung Galaxy Note 8 (Android 9);
- Google Pixel (Android Q).

AndiAJ:
Reproducible on **`Beta 4.0.0-beta.3`** using a OnePlus A3 (Android 6.0.1)
► Screenshot
![rsz_screenshot_20200304-122217](https://user-images.githubusercontent.com/51314259/75870790-32088f00-5e14-11ea-85a0-063fdbdf2f0b.png)

Issue #5346, opened 2019-09-15
[Bug] Sometimes the number of tabs on the tab pile button is not centered


klint:
## Steps to reproduce
I haven't find a pattern, but it happens quite often

### Expected behavior
The number of tabs is centered in the button

### Actual behavior
The number is on the top right corner of the button.
By the way, maybe it's another bug, but is the space between the shield icon and the lock icon not too wide?

![image](https://user-images.githubusercontent.com/218971/64927171-7755e780-d807-11e9-8393-a2fd7c78ee7b.png)

### Device information
* Android device: Redmi Note 7
* Fenix version: 12580614

ekager:
@mcomella mentioned in a dupe (#5868) that he also sees this and could be related to device font size set to "Large"

cadeyrn:
I don't see this very often but two times in the last few weeks. My device font is not set to "Large".

darkwing:
I haven't seen this in my time working on Fenix.  I did check this out with a hundred tabs open and get this, which looks nice:

![Screenshot_20200211-162618-1](https://user-images.githubusercontent.com/46655/74257680-ca4ea080-4cba-11ea-8b99-a861fce286b7.png)

I'm seeing the textSize set to `12sp` as well as `android:layout_gravity="center"`, so I'm not sure what could be causing this.

abodea:
Verified as fixed with latest Nightly 2/14 with Google Pixel XL(Android 10), Samsung Galaxy S10+(Android 9), Samsung Galaxy S8+(Android 8), Samsung Galaxy S7(Android 7), OnePlus (Android 6), LG K4 (Android 5).

Now...depending on the device screen size this would adjust.
I also tried with 1k tabs opened and the tab number position was correct.

@darkwing @ekager I will keep looking after this while testing and add new info if it will be the case.
Please try to review this and let me know if this verification looks good.

 - Android 10, 9, 8
![Screenshot_20200216-001935](https://user-images.githubusercontent.com/42831109/74544844-7c52ca80-4f50-11ea-9609-0ca67a10bd9e.png)

- Android 7 
![Screenshot_20200214-163354](https://user-images.githubusercontent.com/42831109/74544877-8d9bd700-4f50-11ea-91a6-85244906980a.png)

- Android 6 
![Screenshot_20200214-162813](https://user-images.githubusercontent.com/42831109/74544886-92608b00-4f50-11ea-803c-f9900746a0cf.png)

- Android 5 
[...]

PR #8918, opened 2020-03-03
For #5346 & For #8859: Fixes textSize & animation on TabCounter.

sblatz:
### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/toolbar/TabCounter.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/toolbar/TabCounter.kt
@@ -11,10 +11,9 @@ import android.graphics.Typeface
 import android.util.AttributeSet
 import android.util.TypedValue
 import android.view.LayoutInflater
-import android.view.ViewTreeObserver
 import android.widget.RelativeLayout
 import kotlinx.android.synthetic.main.mozac_ui_tabcounter_layout.view.*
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mozilla/fenix/components/toolbar/TabCounter.kt
```diff
@@ -199,34 +197,21 @@ open class TabCounter @JvmOverloads constructor(
             ONE_DIGIT_SIZE_RATIO
         }
 
-        if (newRatio != currentTextRatio) {
```
ekager: Should we keep this check for doing nothing if the ratios match?

gabrielluong:
Gonna add a link to https://github.com/mozilla-mobile/android-components/issues/2743 here so I remind myself to add these changes to the AC TabCounter.

app/src/main/java/org/mozilla/fenix/components/toolbar/TabCounter.kt
sblatz: I don't think this is necessary since it's not running a complicated calculation and Android is going to eat the draw/resize request if we're setting the same size. 

It was good before when it was updating after a long draw request & was doing complicated calculations.

We can add back if we think it'll speed up perf tho

Issue #7847, opened 2020-01-22
Improve startup performance of the Fenix wrapper around Glean (~130ms perfTest)


mcomella:
We're trying to move everything non-essential until after visual completeness in order to improve startup time. GleanMetricsService runs code that may be fine to run after visual completeness: find out and if so, move it!

We should ask the Glean team to review the PR to ensure we're using it correctly.

---

_Old comment 0:_

The Glean SDK has performance issues on startup to be addressed in https://github.com/mozilla-mobile/fenix/issues/7746. However, the way Fenix wraps Glean also has performance issues to be addressed by this ticket. In particular, the following methods run (times on Pixel 2, forPerformanceTest build):
* 14.35ms `GleanMetricsService.start`
* Coroutines launched in `GleanMetricsService`
  * 3.7ms `Pings.<clint>`
  * 78.28ms `setStartupMetrics`
    * ? `ActivationPing.checkAndSend`
* 3.01ms x3 GleanMetricsSerivce.shouldTrack
* 9.65ms x3 GleanMetricsService.track (called in other methods)

For a total of 134.31ms.

### Questions
- Can all of these operations wait until after visual completeness? In particular...
  - Each `track` call
  - `setStartupMetrics`?
  - the `ActivationPing`?
- If not, can these operations be moved to a background thread during init? Glean has mentioned all of their SDK methods should be thread safe.

---

Reach out to the Fenix team (boek?) to determine which methods need to be called when. We can also consider delegating some product requirements (e.g. `ActivityPing`, `StartupMetrics` needed right away?) to ecsmyth.

Dexterp37:
I wonder if this abstraction is still necessary: given that Leanplum only has a few metrics and most of the metrics use Glean, what about using the Glean APIs at the call sites directly? Then there would not be any need for touching multiple files when adding a metric and, more importantly, there will be no performance bottleneck.

jimporter:
@ekager: @mcomella suggested you might have thoughts about how to refactor our usage of `GleanMetricsService` (possibly even removing it entirely and just using Glean directly as @Dexterp37 suggested). Any thoughts here? I think our main goal should probably be deferring as much of the Glean code as possible to after startup... though maybe just removing `GleanMetricsService` would cut back on some overhead here?

jimporter:
[...]

PR #8923, opened 2020-03-03
For #7847: Improve startup performance of the Fenix wrapper around Glean

jimporter:
This should address #7847. I'm not quite sure what I should do about cancelling the callbacks we're deferring until visual completeness, so I removed that for now. I think we probably want to add it back in, so let me know if you have any ideas. Based on my tests, this shaves about 38ms off our app-link startup time (for some reason, I can't get numbers for regular startup...).

@emalysz Am I using the visual completeness manager correctly here? I didn't see any uses of it yet in the tree, so I just made an educated guess as to what the right way here is (Kotlin isn't my strongest language).

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt
@@ -5,9 +5,6 @@
 package org.mozilla.fenix.components.metrics
 
 import android.content.Context
-import kotlinx.coroutines.Job
-import kotlinx.coroutines.MainScope
-import kotlinx.coroutines.launch
 import mozilla.components.service.glean.Glean
 import mozilla.components.service.glean.private.NoExtraKeys
 import mozilla.components.support.base.log.logger.Logger
@@ -509,12 +506,6 @@ class GleanMetricsService(private val context: Context) : MetricsService {
 
     private val logger = Logger("GleanMetricsService")
     private var initialized = false
-    /*
-     * We need to keep an eye on when we are done starting so that we don't
-     * accidentally stop ourselves before we've ever started.
-     */
-    private lateinit var gleanInitializer: Job
-    private lateinit var gleanSetStartupMetrics: Job
 
     private val activationPing = ActivationPing(context)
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt
```diff
@@ -531,12 +527,13 @@ class GleanMetricsService(private val context: Context) : MetricsService {
         // because it calls Google ad APIs that must be called *off* of the main thread.
         // These two things actually happen in parallel, but that should be ok because Glean
         // can handle events being recorded before it's initialized.
-        gleanInitializer = MainScope().launch {
+        context.components.performance.visualCompletenessTaskManager.add {
```
emalysz: nit: Add a comment about why we are delaying these functions & how they will be run after visual completeness

app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt
```diff
@@ -531,12 +527,13 @@ class GleanMetricsService(private val context: Context) : MetricsService {
...
             Glean.registerPings(Pings)
         }
+
         // setStartupMetrics is not a fast function. It does not need to be done before we can consider
         // ourselves initialized. So, let's do it, well, later.
-        gleanSetStartupMetrics = MainScope().launch {
+        context.components.performance.visualCompletenessTaskManager.add {
```
emalysz: nit: we could consider creating a variable to hold `context.components.performance.visualCompletenessTaskManager` to avoid duplication and make it easier to follow

app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt
```diff
@@ -575,8 +572,6 @@ class GleanMetricsService(private val context: Context) : MetricsService {
     }
 
     override fun stop() {
-        gleanInitializer.cancel()
```
emalysz: Since we're now avoiding the job/coroutine scope, deleting this portion is fine. The queue doesn't support cancelling them, so we will end up running them even if app is backgrounded

Mike had brought up an interesting point:  "If we never draw the frame that these get launched from, will they launch when the app is reopened?" I verified that tasks run in the background and are not triggered again when we go navigate back into the app unless we force a cold start (in which case we should be entering this block again)

cc: @mcomella

jimporter:
@emalysz I got my local perf tests working again, and so far it seems like the improvements I was seeing earlier are no longer there. I'm going to look into this a bit more, since my latest run shows app-link startup being even worse with the patch!

jimporter:
Ok, I retested everything and I think I was doing something wrong with my last round of tests, since I'm once again seeing a consistent ~20ms improvement to startup times. I've also updated the patch to address all the review comments, so once CI passes, I'll merge this in.

Normal

| name     |  time | stdev |   diff |
| ---------|-------|-------|------- |
| baseline | 1.090 | 0.020 | 100.0% |
| defer    | 1.067 | 0.026 |  97.9% |

Applink

| name     |  time | stdev |   diff |
| ---------|-------|-------|------- |
| baseline | 1.052 | 0.017 | 100.0% |
| defer    | 1.033 | 0.017 |  98.2% |

Issue #8899, opened 2020-03-02
[Bug] Snackbar not showing in customTab with bottom toolbar


ValentinTimisica:
### Steps to reproduce
- Set toolbar to bottom
- Go to `Settings` -> `About Firefox Preview` -> `What's new in Firefox Preview`
- Long tap on URL bar
- Press on `Copy`

### Expected behavior
- Snackbar with the message `URL copied to clipboard` should show
### Actual behavior
- Nothing happens

### Additional info
- [Video](https://drive.google.com/open?id=1qzGRUo6LTX_ulyM2kl3GIkD8YA9x8dm2) with the bug
- Note that if the toolbar is set to Top, the Snackbar is showing properly, as you can see in the video above.

### Device information
- Android device: Samsung Galaxy Note 10 (Android 10)
- Fenix version: Latest Nightly from store (Version: Nightly 200302 06:00)

ekager:
We'll probably have to change the view that we're passing into the `FenixSnackbar`.

sv-sdeiac:
Hi, I can confirm that this issue is fixed using the latest Firefox Preview Nightly 3/23 #20830607.

Devices:
- Google Pixel 3 (Android 10);
- Samsung Galaxy Tab S6 (Android 9);
- LG g7 fit (Android 8.1).

PR #8933, opened 2020-03-03
For #8899: Set viable anchor for snackbars in both custom and normal tab

mcarare:
### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarView.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarView.kt
@@ -99,7 +99,10 @@ class BrowserToolbarView(
                     clipboard.text = selectedSession?.url
                 }
 
-                FenixSnackbar.make(view, Snackbar.LENGTH_SHORT)
+                FenixSnackbar.makeWithToolbarPadding(
+                    view.rootView.findViewById(android.R.id.content),
+                    Snackbar.LENGTH_SHORT
+                )
                     .setText(view.context.getString(R.string.browser_toolbar_url_copied_to_clipboard_snackbar))
                     .show()
             }


mcarare:
@ekager Aside from this, I could not provide a viable anchor for the snackbar to show up in custom tabs with bottom toolbar. And given this, shouldn't it be safe to use `android.R.id.content` as an anchor for all snackbars and maybe default to this implementation? Just a thought...

app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarView.kt
```diff
@@ -99,7 +99,10 @@ class BrowserToolbarView(
                     clipboard.text = selectedSession?.url
                 }
 
-                FenixSnackbar.make(view, Snackbar.LENGTH_SHORT)
+                FenixSnackbar.makeWithToolbarPadding(
+                    view.rootView.findViewById(android.R.id.content),
```
sblatz: are we able to just do `view.rootView` here? My understanding of how findViewById(android.R.id.content) works is that it just finds the root element of the view, but that should just be `view.rootView` no?

mcarare: No. `view.rootview` returns the DecorView. This is the topmost view in the current view hierarchy (it does not have an ID).
And think as `android.R.id.content` is just the view that contains whatever elements we put, not the system.
All these can be visually differentiated using layout inspector.

BranescuMihai: Yep, you're right @mcarare, view.rootview returns the DecorView. 

I'm thinking, since android.R.id.content gives us the ContentFrameLayout (which gets changed with each activity), why not go one step further, and use the FrameLayout which sits on top (grandparent of content). This way, we avoid overlapping the navigation bar, but we preserve the snackbar between activities, so it can be used in a general way.

PR #8948, opened 2020-03-03
WIP re-enable ContextMenusTest.kt

AaronMT:
We'll see which fails still on this PR and work from there.

### DIFF
--- a/app/src/androidTest/assets/pages/generic4.html
+++ b/app/src/androidTest/assets/pages/generic4.html
@@ -9,11 +9,11 @@
     <a href="generic3.html">Link 3</a>
     <p>
     <a href="../resources/rabbit.jpg">
-        <img src="../resources/rabbit.jpg" title="test_link_image">
+        <img src="../resources/rabbit.jpg" alt="test_link_image">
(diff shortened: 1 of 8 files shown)


AaronMT:
So yet another race. I've simulated and reproduced the intermittent failures happening here happening with 
`setBodyDelay(3, TimeUnit.SECONDS)` on MockResponse probably waiting for the images.

We need to use something like https://github.com/JakeWharton/okhttp-idling-resource (although no AndroidX support https://github.com/JakeWharton/okhttp-idling-resource/issues/19)

AaronMT:
Been using this PR as a playground for implementing sessionloadedidlingresource. However I think I'm actually hitting what FirefoxLite it hitting: https://github.com/mozilla-tw/FirefoxLite/issues/3110

We're currently using 3.11.0 

@cnevinc did upgrading MockWebServer fix this for you? Is sessionloadedidlingresource working out for you?

ekager:
Looks like this is still waiting on #9238

AaronMT:
Going to re-open different PR.

Issue #8017, opened 2020-01-31
Advanced block autoplay settings


vesta0:
## User Story
As a user, I want to be able to block Audio only or choose how autoplay is blocked over wifi, so I can control my experience to match my needs. 

## Dependencies

GV: https://bugzilla.mozilla.org/show_bug.cgi?id=1593843#c27

See #255 for original issue

See [mocks](https://app.abstract.com/share/cdba1651-670e-4412-8721-4d70ca025ffd?collectionLayerId=35603d5c-263c-4c70-b3fe-1ae23bd449c8&mode=design&sha=49143c0fafdf6c2d01f7c0b665c92a9fb7c0ebec)

### Acceptance Criteria
-I can choose to only block audio. When I block audio only, only audio is blocked
-I can choose to allow audio and video over wifi only. When I am not on wifi, audio and video are blocked, and when I am on wifi audio and video auto play.

<img width="350" alt="image" src="https://user-images.githubusercontent.com/4400286/72918787-1f6e4680-3cfb-11ea-9520-bcdafe562618.png">

Baron-Severin:
It looks like newer versions of Android encourage us to check "is this an unmetered connection?" instead of "are we on WIFI?"  @vesta0 should we stick with "allow on WIFI" or change to "allow on unmetered"?

vesta0:
Good point @Baron-Severin Let's see what @betsymi thinks!

sv-ohorvath:
@vesta0 @Baron-Severin @liuche included this in the 4.1 release notes, but I don't see any changes on Beta 4.1 or Nightly and the PR looks like it's still open. Should there be anything to verify on this issue for now?

betsymi:
> @vesta0 @Baron-Severin @liuche included this in the 4.1 release notes, but I don't see any changes on Beta 4.1 or Nightly and the PR looks like is still open. Should there be anything to verify on this issue for now?

Am still working on unpacking this. For a future state, I don't think this is a simple matter of changing the strings we have. May need to make adjustments in the design to provide more clarity for users, particularly regarding `Allow audio and video on Wi-Fi only`. Does this means that audio and video will be blocked on cellular data?

liuche:
@sv-ohorvath that was my mistake! I've removed it from the release notes.

Baron-Severin:
> Does this means that audio and video will be blocked on cellular data?

@betsymi correct.

[...]

PR #8911, opened 2020-03-02
For 8017: add strings for advanced video autoplay

Baron-Severin:
See https://github.com/mozilla-mobile/fenix/issues/8503#issuecomment-592658604

### Pull Request checklist

- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features. In addition, it includes a screenshot of a successful [accessibility scan](https://play.google.com/store/apps/details?id=com.google.android.apps.accessibility.auditor&hl=en_US) to ensure no new defects are added to the product.

### After merge
- [ ] **Milestone**: Make sure issues finished by this pull request are added to the [milestone](https://github.com/mozilla-mobile/fenix/milestones) of the version currently in development.

### To download an APK when reviewing a PR:
1. click on Show All Checks,
2. click Details next to "Taskcluster (pull_request)" after it appears and then finishes with a green checkmark,
3. click on the "Fenix - assemble" task, then click "Run Artifacts".
4. the APK links should be on the left side of the screen, named for each CPU architecture

PR #8953, opened 2020-03-03
For #8017: fix wifi string

Baron-Severin:
### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -596,14 +596,16 @@
     <string name="tracking_protection_on">On</string>
     <!-- Summary of tracking protection preference if tracking protection is set to off -->
     <string name="tracking_protection_off">Off</string>
-    <!-- Label that indicates video and audio autoplay is blocked -->
+    <!-- DEPRECATED: Label that indicates video and audio autoplay is blocked -->
     <string name="preference_option_autoplay_blocked">Video and audio blocked</string>
-    <!-- Label that indicates video and audio autoplay is allowed -->
+    <!-- DEPRECATED: Label that indicates video and audio autoplay is allowed -->
     <string name="preference_option_autoplay_allowed">Video and audio allowed</string>
     <!-- Label that indicates that all video and audio autoplay is allowed -->
     <string name="preference_option_autoplay_allowed2">Allow audio and video</string>
-    <!-- Label that indicates that video and audio autoplay is only allowed over WIFI -->
+    <!-- DEPRECATED: Label that indicates that video and audio autoplay is only allowed over WIFI -->
     <string name="preference_option_autoplay_allowed_wifi_only">Allow audio and video on WIFI only</string>
+    <!-- Label that indicates that video and audio autoplay is only allowed over WIFI -->
+    <string name="preference_option_autoplay_allowed_wifi_only2">Allow audio and video on Wi-Fi only</string>
     <!-- Label that indicates that video autoplay is allowed, but audio autoplay is blocked -->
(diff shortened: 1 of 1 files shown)


Baron-Severin:
@betsymi would you mind verifying these strings before we merge?  Please lmk if you would like any additional changes.  [These mocks](https://user-images.githubusercontent.com/4400286/72918787-1f6e4680-3cfb-11ea-9520-bcdafe562618.png) provide some context of where they will be used in app (#8017).

flodolo:
@Baron-Severin 
I'm having a hard time understanding this patch.

1) Why do you need to keep around the old strings? A "DEPRECATED" comment is not going to prevent 100 languages from translating them, and will waste time and resources. 
2) Why are we deprecating more than just the Wi-Fi string? Does it mean that we need new strings to replace them? If that's the case, we should remove all these strings, and land them when they're ready. As far as I can tell, these strings landed without code behind in https://github.com/mozilla-mobile/fenix/pull/8911

Baron-Severin:
@flodolo 

> Why do you need to keep around the old strings?

Our process has been to mark as deprecated then periodically go through and remove old strings.  I don't remember what problem it was solving, tbh.  I'll bring it up with my team since it sounds like it doesn't work well for you.

> Why are we deprecating more than just the Wi-Fi string? 

I forgot to deprecate a few while adding new strings in #8911.

> Does it mean that we need new strings to replace them? 

No, we should have everything we need here unless requirements change.

> we should remove all these strings, and land them when they're ready.

Normally we would.  This close to deadlines we've been avoiding adding any strings at all.  Where they're absolutely necessary, we're landing them early (before code is ready) to give localizers as much time as possible.

flodolo:
> > Why do you need to keep around the old strings?
> 
> Our process has been to mark as deprecated then periodically go through and remove old strings. I don't remember what problem it was solving, tbh. I'll bring it up with my team since it sounds like it doesn't work well for you.

That makes sense if you're already shipping a feature, and want to be able to ship more updates while the new strings are not used yet. At that point you definitely need the old (deprecated) strings in the repository. 

As I said, these strings seem to have landed without code behind, so we're asking to localize twice the content for no good reason. Is that assumption correct?

Note that both deprecated and new strings haven't been exposed to localization yet, they're stuck in quarantine until I understand if we can avoid this.

Baron-Severin:
`preference_option_autoplay_allowed` and `preference_option_autoplay_blocked` have been in the app since September of last year, they need to be deprecated following the normal approach.  `preference_option_autoplay_allowed_wifi_only` has not yet been released, I duplicated it because I have been told to never change strings that have been committed to master.  I don't know how the localization implementation details work, I can remove `preference_option_autoplay_allowed_wifi_only` if this is an exception to that rule.

If that looks good to you @flodolo, please take a look at https://github.com/mozilla-mobile/fenix/pull/9068 and verify that we're set to move forward.

flodolo:
> `preference_option_autoplay_allowed` and `preference_option_autoplay_blocked` have been in the app since September of last year, they need to be deprecated following the normal approach.

Thanks, I missed that.

> `preference_option_autoplay_allowed_wifi_only` has not yet been released, I duplicated it because I have been told to never change strings that have been committed to master. I don't know how the localization implementation details work, I can remove `preference_option_autoplay_allowed_wifi_only` if this is an exception to that rule.

"Do not change strings in master" is the correct approach, because you don't know if a string is only in quarantine, or exposed for localization. In this case, it's stuck in quarantine and can be removed, or changed without a new ID.

Baron-Severin:
👍  thanks for helping clear that up @flodolo

Issue #8291, opened 2020-02-10
Content process is created on demand, introducing a 180ms+ stall


acreskeyMoz:
## Steps to reproduce
Launch Firefox Preview
Enter a search term or a URL in the URL bar, followed by the enter key

### Expected behavior
The Content process is created at some point during the url entry so that it is available when the url is finalized.

### Actual behavior
The Content process is created after the user hits enter.

In this profile from a Pixel 3, there was at least a 180ms delay after the key event:
https://perfht.ml/2H8gSkh

Bug originally logged against Geckoview:
https://bugzilla.mozilla.org/show_bug.cgi?id=1614486

### Device information
Profile is from  Pixel 3

* Android device: Seeing this on all devices (Pixel 3, Moto G5)
* Fenix version: 73.0b13 (PlayStore, Feb 10)

mstange:
Snorp added info on the bug:
> The easiest solution here is for Fenix to create a `GeckoSession` as soon as it knows it needs one (e.g., entering the urlbar from the home screen). This will cause a content process to be created if there isn't one available.

acreskeyMoz:
It looks like the content process used to be preloaded: https://github.com/mozilla-mobile/fenix/issues/3588

snorp:
Right, preloading could also help, but it doesn't fully start the content process. It just starts the `Service` and loads the libraries, which is probably doesn't help *that* much.

mcomella:
Triage: assign to Christian to hand-off to a-c & GV teams to optimize the implementation.

csadilek:
@pocmo and I chatted about this more: We could create a `GeckoSession` right away when the engine is created and hold it there until it's first needed. This seems quite hackish though, as it makes assumptions on implementation details. Is doing this once enough i.e. is there ever only one content process and should we have to know about this?

@snorp is doing this once guaranteed to be enough? Any API call we could add/introduce to make this easier? When we introduce the speculative connect call (in https://github.com/mozilla-mobile/fenix/issues/7863) would that help here as well?

snorp:
There will be more content processes, but you don't really need to care about that. It doesn't seem like as much of a hack if you think of it more as speculatively creating the `GeckoSession`.

Speculative connect doesn't help because that doesn't spin up all of the necessary bits to create a `GeckoSession`. I think we could probably add speculative `GeckoSession` creation in all of the places we do a speculative connect (so here in urlbar and also intents).

csadilek:
[...]

PR #8955, opened 2020-03-03
Issue #8291: Speculatively create content process in search fragment

csadilek:
Requires: https://github.com/mozilla-mobile/android-components/pull/6165 

Summary: This makes sure we have an engine (gecko) session ready when the user enters a search term or a URL. 

NB: The reason this is needed in Fenix (and not R-B) is due to the different UI implemenations. R-B opens the tab (and creates an engine session) before the toolbar and awesomebar are displayed. In Fenix, the new tab is opened after the user "hits enter" in the toolbar of the search fragment.


csadilek:
Added the upgrade to AC 36 to get this green.

PR #8959, opened 2020-03-03
Bug 1619795: Make it easier to specify paths of toolchain artifacts

tomprince:
This updates taskgraph to allow specifying `toolchain-artifact` on toolchain tasks, to support the mozilla-central ffmpeg task which has an artifact at `public/` instead of `public/build/`.


tomprince:
This depends on https://phabricator.services.mozilla.com/D65250

PR #8971, opened 2020-03-04
Update IntentProcessor implementation to comply with new interface

pocmo:
We changed the Fenix implementation in https://github.com/mozilla-mobile/fenix/pull/8823. And removed `matches()` from the interface in AC in https://github.com/mozilla-mobile/android-components/issues/6097. Now we can remove the `override` keyword from the `IntentProcessor` implementations living in Fenix.


pocmo:
Oh, missed the tests. Updating!

PR #8978, opened 2020-03-04
8017 advanced autoplay controls.

Baron-Severin:
### Pull Request checklist

### After merge

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/utils/WifiConnectionMonitor.kt
```diff
@@ -0,0 +1,77 @@
...
+ * Example:
+ * ```kotlin
+ *  app.components.wifiConnectionListener.addOnWifiConnectedChangedListener { isConnected ->
+ *    if (isConnected) {
+ *      downloadThing()
+ *    }
+ *  }
+ * ```
+ */
+class WifiConnectionMonitor(app: Application) {
```
Amejia481: This is a good candidate to be completely ported  to AC :)

app/src/main/java/org/mozilla/fenix/utils/WifiConnectionMonitor.kt
```diff
@@ -0,0 +1,77 @@
...
+        object : ConnectivityManager.NetworkCallback() {
+            init {
+                val request = NetworkRequest.Builder()
+                    .addTransportType(NetworkCapabilities.TRANSPORT_WIFI)
+                    .build()
+
+                val cm = app.getSystemService(Context.CONNECTIVITY_SERVICE) as
+                        ConnectivityManager
+
+                cm.registerNetworkCallback(request, this)
```
Amejia481: Should this class have `start` and `stop` function this way consumers can control when to start/stop listing i.e. ([registerNetworkCallback](https://developer.android.com/reference/android/net/ConnectivityManager#unregisterNetworkCallback(android.net.ConnectivityManager.NetworkCallback)))?

app/src/main/java/org/mozilla/fenix/settings/sitepermissions/SitePermissionsManagePhoneFeatureFragment.kt
```diff
@@ -73,66 +128,120 @@ class SitePermissionsManagePhoneFeatureFragment : Fragment() {
...
+     */
+    private fun saveActionInSettings(context: Context, autoplaySetting: Int) {
+        settings.setAutoplayUserSetting(autoplaySetting)
+        val (audible, inaudible) = when (autoplaySetting) {
+            AUTOPLAY_ALLOW_ALL -> ALLOWED to ALLOWED
+            AUTOPLAY_ALLOW_ON_WIFI -> {
+                addWifiConnectedListener(context.application)
+                return
+            }
+            AUTOPLAY_BLOCK_AUDIBLE -> BLOCKED to ALLOWED
```
Amejia481: This setting confuse me a bit.

 In the UI this is "Block audio", when the user selects this setting, I think he/she will be expected **just audio to be blocked** not video ,right? But What this setting is stating is `audible = BLOCKED` and `inaudible = ALLOWED`, this will cause that **videos that are not muted will be blocked** as they contain audible content, can you help me to clarify?

Baron-Severin: Yeah, we're matching desktop behavior, but it is a little bit confusing.  @betsymi do you have any thoughts on this?  If we do want to make any changes, we'll have to reach out to desktop about their copy too.

app/src/main/java/org/mozilla/fenix/utils/WifiConnectionMonitor.kt
Baron-Severin: I could go either way on this one.  We could always add it when we encountered a use case.  I like that setting the wifi listener is a simple as just saying `monitor.addOnWifiConnectedChangedListener { }`.

Amejia481: My only concern is performance, we are only interested on calls when the app is in the foreground, otherwise will be getting notified without any gain. As we just want to consume what is strictly needed Wdy? :)

app/src/main/java/org/mozilla/fenix/settings/sitepermissions/SitePermissionsManagePhoneFeatureFragment.kt
betsymi: Is this the setting we're talking about, or someplace else? Could you provide a screenshot so I can see the full context of the string? 

Screenshot_20200309-152229

Baron-Severin: @betsymi yep, that's the screen.  After this PR lands, it will look like this though:

https://user-images.githubusercontent.com/4400286/72918787-1f6e4680-3cfb-11ea-9520-bcdafe562618.png

app/src/main/java/org/mozilla/fenix/utils/WifiConnectionMonitor.kt
Baron-Severin: I don't expect this to block anything important, so I'm not massively worried about it, but it's still a reasonable concern.  Sure, I'll add it in.

app/src/main/java/org/mozilla/fenix/utils/WifiConnectionMonitor.kt
Baron-Severin: Didn't occur to me until now, and I think it's out of scope for this story, but we should change how this class exposes its callbacks if/when we move it to AC.  We would get two big benefits if, instead of just adding/removing callbacks, we exposed this as a `WifiStore` (with `WifiState(val isConnected: Boolean)`).  1 if we ever needed to listen to other framework wifi callbacks, adding them would be simple and easy (as is it would get messy fast), and 2 it would be much more in line with the rest of our architecture.

Amejia481: 👍

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -182,6 +182,20 @@ open class HomeActivity : LocaleAwareAppCompatActivity() {
         BrowsersCache.resetAll()
     }
 
+    override fun onStart() {
+        super.onStart()
+
+        with(application.components.wifiIntegration) {
```
Amejia481: What do you think about to put it on [BaseBrowserFragment](https://github.com/Baron-Severin/fenix/blob/2db2d9ceae0c43e933050555d60d67c5667b4d00/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt#L86)?  [As it behave the same way as other features that are related to the browser mode](https://github.com/Baron-Severin/fenix/blob/2db2d9ceae0c43e933050555d60d67c5667b4d00/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt#L91), it should be only active while the user is on the browser mode.

You could create something similar to [ToolbarIntegration](https://github.com/Baron-Severin/fenix/blob/master/app/src/main/java/org/mozilla/fenix/components/toolbar/ToolbarIntegration.kt) as you are decorating an existing component with a new behaviour:

You could rename `WifiIntegration` to `SitePermissionsIntegration`

```kotlin
class SitePermissionsIntegration(
    private val wifiConnectionMonitor: WifiConnectionMonitor
): LifecycleAwareFeature {

    /**
     * Adds listener for autplay setting [AUTOPLAY_ALLOW_ON_WIFI]. Sets all autoplay to allowed when
     * WIFI is connected, blocked otherwise.
     */
    private val wifiConnectedListener: ((Boolean) -> Unit) by lazy {
        { connected: Boolean ->
            val setting =
                if (connected) SitePermissionsRules.Action.ALLOWED else SitePermissionsRules.Action.BLOCKED
            settings.setSitePermissionsPhoneFeatureAction(PhoneFeature.AUTOPLAY_AUDIBLE, setting)
            settings.setSitePermissionsPhoneFeatureAction(PhoneFeature.AUTOPLAY_INAUDIBLE, setting)
        }
    }

    /**
     * If autoplay is only enabled on WIFI, sets a WIFI listener to set them accordingly. Otherwise
     * noop.
     */
   internal fun maybeAddWifiConnectedListener() {
        if (settings.getAutoplayUserSetting(default = AUTOPLAY_BLOCK_ALL) == AUTOPLAY_ALLOW_ON_WIFI) {
            addWifiConnectedListener()
        }
    }

    fun addWifiConnectedListener() {
        wifiConnectionMonitor.addOnWifiConnectedChangedListener(wifiConnectedListener)
    }

    internal fun removeWifiConnectedListener() {
        wifiConnectionMonitor.removeOnWifiConnectedChangedListener(wifiConnectedListener)
    }

    override fun start() {
        wifiConnectionMonitor.start()
        maybeAddWifiConnectedListener()
    }

    override fun stop() {
        removeWifiConnectedListener()
        wifiConnectionMonitor.stop()
    }
}
```
 You could init it in [BaseBrowserFragment](https://github.com/Baron-Severin/fenix/blob/2db2d9ceae0c43e933050555d60d67c5667b4d00/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt#L86) like similar integrations 

```kotlin

    private val sitePermissionsIntegration = ViewBoundFeatureWrapper<SitePermissionsIntegration>()

    protected open fun initializeUI(view: View): Session? {
            // Other initializations
 
            sitePermissionsIntegration.set(
                feature= SitePermissionsIntegration(
                    requireContext().settings(),
                    requireContext().wifiConnectionMonitor
                ),
                owner = this,
                view = view
            )

}
```

Baron-Severin: Right now only browser stuff cares about the WIFI state, but it doesn't seem to me like there's a natural connection there.  We could easily want other things to listen to WIFI state across the app, which is why I put it in the activity.  I think there's an argument to be made for not overgeneralizing, but this still feels more appropriate to me.  What do you think?

Baron-Severin:
> There are some issues trying with autoplay when using it with a mobile network [see](https://drive.google.com/file/d/1CcRWQEY5rokiyJ3qypG4scDiXnBOmyce/view?usp=sharing), it's only working with WIFI network even if we don't have that setting active.

🤦‍♂  the rules are being saved as expected, but that doesn't forward them to the feature.  I'll put a fix up soon.

Baron-Severin:
> There are some issues trying with autoplay when using it with a mobile network [see](https://drive.google.com/file/d/1CcRWQEY5rokiyJ3qypG4scDiXnBOmyce/view?usp=sharing), it's only working with WIFI network even if we don't have that setting active.

Huh, good catch @Amejia481.  This is reproducible on release Fenix and GeckoView Example as well, I've opened https://bugzilla.mozilla.org/show_bug.cgi?id=1621825.

app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
```diff
@@ -391,6 +391,11 @@ abstract class BaseBrowserFragment : Fragment(), UserInteractionHandler, Session
                 view = view
             )
 
+            context.settings().setSitePermissionSettingListener(viewLifecycleOwner) {
+                this.context?.let { assignSitePermissionsRules(it) }
+            }
+            assignSitePermissionsRules(context)
```
Amejia481: nit: I think we are calling `assignSitePermissionsRules` twice.

Baron-Severin: We need to assign when first entering the fragment, and need to again whenever site permissions rules change (atm, this can happen when the user flips their wifi on/off from the browser fragment).  HOWEVER, it is calling twice every time the WIFI changes (because that updates AUTOPLAY_AUDIBLE and AUTOPLAY_INAUDIBLE).

I measured, and this assignment is taking ~.15 - .2MS on my Pixel 2.  Not a ton, but not nothing.  I spent some time trying to avoid the double call but haven't come up with a better solution.  Open to suggestions, but I don't think it's worth spending much more time on it.

Baron-Severin:
Do not land until https://github.com/mozilla-mobile/fenix/issues/8017?email_token=ADEPNARBLZ337GF637ZM7QTRGXMWLA5CNFSM4KOFD4QKYY3PNVWWK3TUL52HS4DFVREXG43VMVBW63LNMVXHJKTDN5WW2ZLOORPWSZGOEOKCBSA#issuecomment-600878861 is resolved.

Baron-Severin:
CC @flodolo

flodolo:
> @flodolo see [2c43e4b](https://github.com/mozilla-mobile/fenix/commit/2c43e4b3ccebc88133b6dcf3627d6321215b5f0d)

LGTM. But what about the subtext in [Betsy's comment](https://github.com/mozilla-mobile/fenix/issues/8017#issuecomment-599802120)?

>  Audio and video will play on Wi-Fi

Baron-Severin:
> LGTM. But what about the subtext in [Betsy's comment](https://github.com/mozilla-mobile/fenix/issues/8017#issuecomment-599802120)?

Thank you, I missed that.  I'll fix it before merging.

PR #8989, opened 2020-03-05
Use new MemoryConsumer.onTrimMemory() API.

pocmo:
The latest snapshot (currently building) comes with an updated API. Instead of `onLowMemory()` we now support `onTrimMemory(level)` from Android's `ComponentCallbacks2`. This allows us to make better decisions based on the `level`. In addition to that I also added the call to `BrowserIcons` which also supports trimming its memory usage if needed.


Amejia481:
It looks like the TC hasn't pick the new snapshot :(

pocmo:
Needed to push it to get it over the finish line, now it should be available.

Issue #8826, opened 2020-02-27
Asynchronously inflate toolbar from within BrowserToolbarView


emalysz:
Let's see if we can asynchronously inflate the toolbar to improve the performance BaseBrowserFragment.initializeUI. 

This seems like it could be a win for app link, and a breakdown of the time spent in this function can be found at https://github.com/mozilla-mobile/fenix/issues/8061#issuecomment-590584913

mcomella:
@emalysz What's this waiting on again?

emalysz:
@mcomella we were waiting to see if the perceived performance is worth the change cc @ecsmyth

ecsmyth:
Still in a holding pattern. I'll hopefully have another update this week.

ecsmyth:
[UR request filed](https://bugzilla.mozilla.org/show_bug.cgi?id=1633404). Waiting on triage.

ecsmyth:
UR will not be able to study before the end of June. Assigning to UX for guidance.

brampitoyo:
Have just caught up with @ecsmyth, and focusing particularly on this observation:

> The change, when measured locally with a simple test site, improves Fenix startup by roughly 270 milliseconds (i.e. a 25% improvement) […] However, to me, the original experience feels better, faster, possibly as a result of the animations that give a sense of progress.

Combining the fact that:

- Objective performance is measurably higher, but isn’t guaranteed to translate to subjective perceived performance (ie. time until users perceive Fenix as “finished loading” and take their first action within the app)
- Asynchronously inflated toolbar would’ve excised Android’s default app opening animation, which would’ve made Fenix go against system convention.

I think that we shouldn’t go ahead with this change.

We may still be able to do some things that could increase perceive performance around loading time. For example:

- Use transition time to pre-load UI elements and page content, so everything on the page can be shown all at once
- Shorten animation and duration scale (maybe scale to 0.5x?) so transitions look and feel faster

mcomella:
> I think that we shouldn’t go ahead with this change.

@brampitoyo @ecsmyth Shall we close this issue then? Thanks for looking into it.

mcomella:
Triage: we have no time to pass this through UR now so we'll move to low impact.

mcomella:
I'm going to close this issue and link it from our "perceived performance" ticket https://jira.mozilla.com/browse/FXP-1264 on JIRA, given it seems unactionable and will likely continue to be so for a while.

PR #9003, opened 2020-03-05
For #8826: use async layout inflater for browser toolbar view

emalysz:
By delaying the UI initialization, we can improve performance of applink.

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
@@ -218,31 +218,87 @@ abstract class BaseBrowserFragment : Fragment(), UserInteractionHandler, Session
                 lifecycleOwner = this.viewLifecycleOwner
             )
 
-            toolbarIntegration.set(
-                feature = browserToolbarView.toolbarIntegration,
-                owner = this,
-                view = view
-            )
+            browserToolbarView.loadAsync { toolbarView ->
+                toolbarIntegration.set(
+                    feature = browserToolbarView.toolbarIntegration,
+                    owner = this@BaseBrowserFragment,
+                    view = view
+                )
 
-            findInPageIntegration.set(
-                feature = FindInPageIntegration(
-                    store = store,
-                    sessionId = customTabSessionId,
-                    stub = view.stubFindInPage,
-                    engineView = view.engineView,
-                    toolbar = browserToolbarView.view
-                ),
-                owner = this,
-                view = view
-            )
+                findInPageIntegration.set(
+                    feature = FindInPageIntegration(
+                        store = store,
+                        sessionId = customTabSessionId,
+                        stub = view.stubFindInPage,
+                        engineView = view.engineView,
+                        toolbar = toolbarView
+                    ),
+                    owner = this@BaseBrowserFragment,
+                    view = view
+                )
 
-            browserToolbarView.view.display.setOnSiteSecurityClickedListener {
-                showQuickSettingsDialog()
-            }
+                toolbarView.display.setOnSiteSecurityClickedListener {
+                    showQuickSettingsDialog()
+                }
 
-            browserToolbarView.view.display.setOnTrackingProtectionClickedListener {
-                context.metrics.track(Event.TrackingProtectionIconPressed)
-                showTrackingProtectionPanel()
+                toolbarView.display.setOnTrackingProtectionClickedListener {
+                    context.metrics.track(Event.TrackingProtectionIconPressed)
+                    showTrackingProtectionPanel()
+                }
+
+                fullScreenFeature.set(
+                    feature = FullScreenFeature(
+                        sessionManager,
+                        SessionUseCases(sessionManager),
+                        customTabSessionId
+                    ) { inFullScreen ->
+                        if (inFullScreen) {
+                            FenixSnackbar.make(view.rootView, Snackbar.LENGTH_SHORT)
+                                .setText(getString(R.string.full_screen_notification))
+                                .show()
+                            activity?.enterToImmersiveMode()
+                            toolbarView.visibility = View.GONE
+
+                            if (FeatureFlags.dynamicBottomToolbar) {
+                                engineView.setDynamicToolbarMaxHeight(0)
+                                // TODO We need to call force expand here to update verticalClipping #8697
+                                // Without this, fullscreen has a margin at the top.
+                                engineView.setVerticalClipping(0)
+                            }
+                        } else {
+                            activity?.exitImmersiveModeIfNeeded()
+                            (activity as? HomeActivity)?.let { activity ->
+                                activity.themeManager.applyStatusBarTheme(activity)
+                            }
+                            toolbarView.visibility = View.VISIBLE
+                            if (FeatureFlags.dynamicBottomToolbar) {
+                                engineView.setDynamicToolbarMaxHeight(toolbarHeight)
+                            }
+                        }
+                        if (!FeatureFlags.dynamicBottomToolbar) {
+                            updateLayoutMargins(inFullScreen)
+                        }
+                    },
+                    owner = this@BaseBrowserFragment,
+                    view = view
+                )
+
+                session.register(observer = object : Session.Observer {
+                    override fun onLoadRequest(
+                        session: Session,
+                        url: String,
+                        triggeredByRedirect: Boolean,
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarView.kt
```diff
@@ -227,7 +227,23 @@ class BrowserToolbarView(
...
+    }
+
+    /**
+     * Performs all actions that refer to the BrowserToolbarView that don't require user action.
+     * They are set in the callback because we need to ensure phones execute the code before
+     * we are done inflating.
+     */
+    fun loadAsync(action: View.() -> Unit) {
+        AsyncLayoutInflater(container.context).inflate(toolbarLayout, container) { finalView, _, parent ->
+            view = finalView.findViewById((R.id.toolbar))
```
mcomella: nit: redundant parens

app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarView.kt
```diff
@@ -227,7 +227,23 @@ class BrowserToolbarView(
...
+            view.app_bar?.setExpanded(true)
+        }
+    }
+
+    /**
+     * Performs all actions that refer to the BrowserToolbarView that don't require user action.
+     * They are set in the callback because we need to ensure phones execute the code before
+     * we are done inflating.
+     */
+    fun loadAsync(action: View.() -> Unit) {
```
mcomella: My overall suggestion is to change this to `(View) -> Unit` and, instead of accessing `browserToolbarView.view` in the caller, access the argument passed in instead. There are several things going on here:
- `lateinit var browserToolbarView.view` is not type-safe: if we pass the view in instead, we can guarantee its type safety. In general, passing arguments directly into functions, instead of accessing them through members, is safer (e.g. can't be rebound, types are as up-to-date as possible, etc.) and is one of the primary reasons functional programming is gaining popularity
- Adding a receiver (i.e. `View.`) should be used carefully since it binds `this` in the caller. This is an issue because 1) it's often not clear which `this` is being referred to and 2) `this` can be used without explicitly writing `this` which becomes a common source of bugs, especially when refactoring. In general, receivers should only used for very small blocks of code or in very specific contexts where the meaning of `this` is well known. How `this` is used in JavaScript is a good example of why this is an unsafe pattern.

app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
```diff
@@ -215,31 +215,86 @@ abstract class BaseBrowserFragment : Fragment(), UserInteractionHandler, Session
                 lifecycleOwner = this.viewLifecycleOwner
             )
 
-            toolbarIntegration.set(
-                feature = browserToolbarView.toolbarIntegration,
-                owner = this,
-                view = view
-            )
+            var owner = this
+            browserToolbarView.loadAsync {
```
mcomella: nit: as I mention in my large comment below, I think we should do something like:
```kotlin
browserToolbarView.loadAsync { toolbarView ->
```

And replace references to `browserToolbarView.view` with `toolbarView` for safer binding.

app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarView.kt
```diff
@@ -54,21 +55,20 @@ class BrowserToolbarView(
 
     private val settings = container.context.settings()
 
+    lateinit var toolbarIntegration: ToolbarIntegration
+
+    lateinit var view: BrowserToolbar
+
+    lateinit var layout: View
```
mcomella: nit: I think this can be private (or if my suggested change below is incorrect, is unused)

app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarView.kt
```diff
@@ -227,7 +227,23 @@ class BrowserToolbarView(
 
     fun expand() {
         if (!settings.shouldUseBottomToolbar) {
-            layout.app_bar?.setExpanded(true)
+            view.app_bar?.setExpanded(true)
```
mcomella: I don't think this will work because `app_bar` contains `view`: I think you need `layout.app_bar` or maybe even just `app_bar`.

emalysz: @mcomella Hm the receiver type it is expecting is View.app_bar. The previous private val layout that they used here was also a view (since it was the return value of LayoutInflater.inflate). Does that mean it was incorrect in the first place?

app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarView.kt
```diff
@@ -227,7 +227,23 @@ class BrowserToolbarView(
...
+     * They are set in the callback because we need to ensure phones execute the code before
+     * we are done inflating.
+     */
+    fun loadAsync(action: View.() -> Unit) {
+        AsyncLayoutInflater(container.context).inflate(toolbarLayout, container) { finalView, _, parent ->
+            view = finalView.findViewById((R.id.toolbar))
+            with(parent!!) {
+                addView(finalView)
+                setupView()
+                action()
```
csadilek: I think we should pass the `finalView` as a parameter to the action lambda. Basically, the same as Mike suggested here: https://github.com/mozilla-mobile/fenix/pull/9003/files#r391224957

app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
```diff
@@ -215,31 +215,86 @@ abstract class BaseBrowserFragment : Fragment(), UserInteractionHandler, Session
                 lifecycleOwner = this.viewLifecycleOwner
             )
 
-            toolbarIntegration.set(
-                feature = browserToolbarView.toolbarIntegration,
-                owner = this,
-                view = view
-            )
+            var owner = this
```
csadilek: Let's remove this `var`. (see below)

app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
```diff
@@ -215,31 +215,86 @@ abstract class BaseBrowserFragment : Fragment(), UserInteractionHandler, Session
...
+            browserToolbarView.loadAsync {
+                toolbarIntegration.set(
+                    feature = browserToolbarView.toolbarIntegration,
+                    owner = owner,
```
csadilek: This can just be `owner = this@BaseBrowserFragment`

app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarView.kt
```diff
@@ -227,7 +227,23 @@ class BrowserToolbarView(
 
     fun expand() {
         if (!settings.shouldUseBottomToolbar) {
-            layout.app_bar?.setExpanded(true)
+            view.app_bar?.setExpanded(true)
+        }
+    }
+
+    /**
+     * Performs all actions that refer to the BrowserToolbarView that don't require user action.
```
csadilek: nit: Hm can we rephrase this? I didn't really understand the "that don't require user action" part, but maybe I am missing something...

Maybe something along these lines:

Triggers async inflation of this [BrowserToolbarView] and invokes the provided [action] once inflation is completed so dependent functionality (views, features) can be initialized.

mcomella:
I ran this through FNPRMS on a P2, `geckoNightlyFennecNightly` and didn't see much change. On 10 runs:
- Before: 1.334s
- After: 1.329s

(re-running for sanity) On 5 runs:
- Before: 1.3228s
- After: 1.3272s

We've previously noticed that UI doesn't seem to have a big impact on the P2 so we should re-run on the G5.

liuche:
@mcomella have you re-run this on the G5? If you don't think there are gains please close, to help us clean up our PR list, thanks!

mcomella:
We're currently blocked on a decision from ecsmyth: https://github.com/mozilla-mobile/fenix/issues/8826#issuecomment-603595201

mcomella:
We'll reopen or create a new one if we decide we want to move forward with this approach.

Issue #8908, opened 2020-03-02
Fenix integration: Notify users when previously unsupported addons become available


Amejia481:
Related issue https://github.com/mozilla-mobile/android-components/issues/5434

Amejia481:
Closed by https://github.com/mozilla-mobile/fenix/pull/9060/commits

sv-ohorvath:
@Amejia481 should this notification be visible on the Fenix Beta 4.1 update, or is it meant only for the migration build? Asking for the new beta 4.1 build validation.

sv-ohorvath:
Arturo's response for future reference: "I belive you have to be migrated from Fennec to Fenix, then some of your add-ons are going to be disabled as they are not supported, then when they become supported , the notification will show up"

PR #9060, opened 2020-03-09
For issue #8908 Notify users when previously unsupported add-ons become available.

Amejia481:


Amejia481:
This ✅  is good to go as mozilla-mobile/android-components#6123 landed

Issue #8803, opened 2019-12-17
Add telemetry probe for dex load time


ecsmyth:
_edit:_ this issue was repurposed to only instrument dex load time instead of startup time but the description for startup probes remain below.

### Why/User Benefit/User Problem
Startup time is an oft cited reason for friction and churn. We need to understand how long it takes Fenix to startup for our users so that we can prioritize work to address regressions and outliers that result in decreased engagement or churn. Local and CI-based testing provides an incomplete picture of how Fenix performs at startup.

### Impact
Improved understanding of how Fenix performs on startup for our users.

### Acceptance Criteria (how do I know when I’m done?)
We should make a best-effort to _incrementally_ instrument as much of startup as possible - until we hit diminishing returns - because it's impossible to do it all. Unfortunately, it will be difficult to define what this looks like in advance so take this as a guiding principle. Msg mcomella if this does not make sense.

Some incremental steps we should try to implement for GA:
* Add a metric that tracks the start time of GV
* Add a metric that tracks the duration of `FenixApplication.onCreate` (this may or may not be helpful but we want to add it to see if it is in practice)
* Add a metric that tracks `FenixApplication.onCreate` until first frame drawn
* Distinguish between first run and other cold starts (don't let this impact performance!)
* (is this reasonable?) Distinguish between which page was shown, i.e. which startup path was used
* Time from first frame drawn to reportFullyDrawn? Essentially, capture loading top sites & open tabs because this is dependent on content quantity

Some incremental steps we probably can't implement by GA but we should verify:
* Add a probe that tracks `FenixApplication.onCreate` until visual completeness
* Add warm/hot start times, distinguish from cold start
* Distinguish between cold starts after upgrading and other cold starts

Here is ecsmyth's ideal startup time probe that inspired these incremental steps:
> Glean telemetry probe landed in Fenix release that measures time, as closely as is practical, from when user initiates startup until the app is started and the initial screen is visually complete. The probe should differentiate between hot, cold, and warm startup scenarios and account for differences in the first page rendered (e.g., how desktop measures first paint and first paint of about:home)

mcomella:
Waiting on ecsmyth to determine if time for `GeckoRuntime.init` (i.e. the time Gecko starts on the main thread before continuing work on a background thread) is a valuable probe.

mcomella:
[...]

PR #9153, opened 2020-03-13
For #8803: add StartupTimelineMeasurements + PoC metric

mcomella:
**PLEASE DO NOT REVIEW: I WILL TAG REVIEWERS AS NEEDED**

This PR adds a class to make measuring parts of startup easy – `StartupTimelineMeasurements` – and uses it to implement a proof-of-concept metric – `PerfMeasurements.geckoRuntimeCreate`. Note: we deprioritized adding many probes on startup so we probably won't add additional probes for a bit.

I want to get a lot of feedback on this. Here's what I want to have before landing:
- [x] Feedback from glean team on how I'm recording these metrics
- [ ] Feedback from ecsmyth: do we actually want to record GeckoRuntime.create?
- [x] Feedback from Fenix: okay to call Glean APIs directly?
- [ ] Feedback from bdekoz on how I record telemetry – e.g. is this similar to Gecko?
- [ ] Code review
- [ ] Data request & review

### Pull Request checklist

  - No UI

### After merge

1. click on Show All Checks,


mcomella:
@bdekoz Could you give me feedback on my approach to taking startup time measurements? In particular, I'm curious about [the API provided by `DurationContainer` in `DurationContainer.kt`](https://github.com/mozilla-mobile/fenix/pull/9153/files#diff-1e4d6b7872857d3f4f6bc5b05836fc80). Some specific questions I'd have:
* How does this compare to the desktop model?
* Is `DurationContainer` too strict? I'm concerned that the caller's implementation will change and we'll be left with bad data so I wanted to put as many guarantees into the code as possible

Given your performance telemetry background, I figured this could be a good area for you but let me know if there's someone else I should ask!

mcomella:
@Dexterp37 Could you give me feedback on my approach to taking telemetry measurements? In particular, I'm curious if I set my metrics up correctly and if the `DurationContainer` model makes sense with Glean. Note that I implemented my own capturing of timestamps instead of using `timespan` because I wanted to be able to 1) make multiple sets of calculations using the same set of timestamps and 2) log the durations I've recorded. Some specific questions I had:
* Given what I'm trying to do, does the `DurationContainer` model make sense here?
* Is String the best metric type for what I'm trying to do? It seems expensive (on the network)
* Is it okay to call Glean directly or do you recommend wrapping it (e.g. in `GleanMetricsService`, like the other Fenix code)?

app/metrics.yaml
```diff
@@ -1856,3 +1856,16 @@ installation:
     notification_emails:
       - fenix-core@mozilla.com
     expires: "2020-09-01"
+
+perf_measurements:
+  gecko_runtime_create:
+    type: string
```
Dexterp37: Using a string metric type it's almost always a bad idea if the datum is not really a string. Coercing a number to a string makes it impossible for the SDK to validate data and to report errors. Moreover, it makes it impossible for the tooling (i.e. GLAM) to aggregate the data for you.

We tried to capture the spirit of this more in [our docs](https://mozilla.github.io/glean/book/user/adding-new-metrics.html).

app/metrics.yaml
```diff
@@ -1856,3 +1856,16 @@ installation:
     notification_emails:
       - fenix-core@mozilla.com
     expires: "2020-09-01"
+
+perf_measurements:
```
Dexterp37: nit: since this is really startup timelime, what about making the category `startup.timeline` and the metric `gecko_runtime_create`?

`perf_measurements` feels a bit generic given that there's other perf measurements being sent..

app/src/main/java/org/mozilla/fenix/perf/DurationContainer.kt
```diff
@@ -0,0 +1,101 @@
...
+ * // OR
+ *
+ * val functionReturnValue = container.measure {
+ *     functionDurationToMeasure()
+ * }
+ * valueForTelemetry = container.duration
+ * ```
+ */
+class DurationContainer(
+    private inline val getTimeNanos: () -> Long = SystemClock::elapsedRealtimeNanos
```
Dexterp37: I would encourage you to use the Glean primitives to measure time rather than implementing your own time measurement framework. Measuring is hard, especially if you want to compare against other components measuring things (e.g. Gecko).

We can work together on any shortcoming of our measurement APIs.

mcomella:
> > Note that I implemented my own capturing of timestamps instead of using timespan because I wanted to be able to 1) make multiple sets of calculations using the same set of timestamps
> 
> @mcomella what do you mean by this? In this code it looks like you're measuring only once, but maybe I misunderstood your requirement. Would you kindly expand on this?

@Dexterp37 This hasn't been implemented yet so we don't know for sure how useful it will be but here's an example: we're measuring cold startup. We measure the total time and break it into multiple small sections to more clearly identify where a regression (or improvement) might occur. Two such sections are time from process start to when we first can run code (`FenixApplication.onCreate`) and the duration of `FenixApplication.onCreate`. We might record these, and calculate their durations, as:
```kotlin
class StartupTimeMeasurements {
    var processStartTimeNanos: Long
    var fenixApplicationOnCreateStartNanos: Long
    var fenixApplicationOnCreateEndNanos: Long

    val frameworkStartupDurationNanos: Long get() = fenixApplicationOnCreateStartNanos - processStartTimeNanos
    val fenixApplicationOnCreateDurationNanos: Long get() = fenixApplicationOnCreateEndNanos - fenixApplicationOnCreateStartNanos
}

// Dump numbers to logcat...
fun dumpInstrumentationForPerfMeasurementInCI() {
    Log.d(TAG, "startup: $frameworkStartupDurationNanos\nonCreate: $fenixApplicationOnCreateDurationNanos)
}

// Upload to glean...
fun uploadPerfTelemetry() {
    StartupTimeline.frameworkStartupDuration.set(frameworkStartupDurationNanos)
    StartupTimeline.fenixApplicationOnCreateDurationNanos.set(fenixApplicationOnCreateDurationNanos)
}
```

For flexibility, I think it's ideal to capture these as timestamps – e.g. it's trivial to add a new calculation for duration here from ApplicationOnCreateEnd to ActivityOnCreateStart – instead of using glean's `timespan` metrics, as they're currently implemented. In some places, it's actually impossible to use the glean `timespan` metric (e.g. we can't execute code at process start time but we can query for the process start time).

> What is your use-case for requiring logging here?

It's mentioned above but we'd like to log to capture data for our test harnesses, in local testing and CI. Ideally we can use the same values as we do in telemetry.

> Timespan is the most appropriate for measuring how much time it takes to, for example, execute a function.

I see the note in the docs now about:

> It's possible to explicitly set the timespan value, in nanoseconds. This API should only be used if your library or application requires recording times in a way that can not make use of start/stop/cancel.

After explaining this, my assumption is that I should keep the current implementation but upload to Glean using this explicit set `timespan` methodology. @Dexterp37 What do you think?

Dexterp37:
> For flexibility, I think it's ideal to capture these as timestamps – e.g. it's trivial to add a new calculation for duration here from ApplicationOnCreateEnd to ActivityOnCreateStart – instead of using glean's `timespan` metrics, as they're currently implemented. In some places, it's actually impossible to use the glean `timespan` metric (e.g. we can't execute code at process start time but we can query for the process start time).

Got it! I now better understand your usecase :) In this case yes, setting the values on timestamps is a good option.
 
> > What is your use-case for requiring logging here?
> 
> It's mentioned above but we'd like to log to capture data for our test harnesses, in local testing and CI. Ideally we can use the same values as we do in telemetry.

We discussed this same use-case with @bdekoz a while back. The solution we found, to prevent building a new measurement infrastructure, was to define a new ping in Glean, the "performance" ping, and send that. In local testing, pings can be sent logged to logcat or tagged to use the debug view. On CI, pings could potentially be sent to the Glean ingestion pipeline to enable tools such as GLAM to conveniently aggregate these for you. See [this comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1618747#c1) for more details.

@mcomella would this be of help?

> I see the note in the docs now about:
> 
> > It's possible to explicitly set the timespan value, in nanoseconds. This API should only be used if your library or application requires recording times in a way that can not make use of start/stop/cancel.
> 
> After explaining this, my assumption is that I should keep the current implementation but upload to Glean using this explicit set `timespan` methodology. @Dexterp37 What do you think?

Yes, with the new details about your future usecase it makes sense to use a timestamp and set its values explicitly. Please do flag me for review after you make the appropriate code change, as setting the value has implications depending on the `lifetime` of the metric (i.e. I think these should be `lifetime: application` metrics).

mcomella:
> The solution we found, to prevent building a new measurement infrastructure, was to define a new ping in Glean, the "performance" ping, and send that. In local testing, pings can be sent logged to logcat or tagged to use the debug view.

@Dexterp37 What do you mean by "prevent building a new measurement infrastructure"? Is this so we don't have to maintain the code 1) to log duration values and 2) for our test harness to parse those values?

One note I have is that there might be some values we want to record locally but not transmit to the server to reduce bandwidth and simplify analysis. For example, we might want to measure the length of `FenixApplication.onCreate` in automation to look for regressions (b/c it has less noise than full startup) but it may be less useful for telemetry purposes so it seems wasteful to collect it. In this case, we'd have to create our own parallel logging anyway (though we could model it after the Glean logging to make it easier for the test harness to parse).

As I understand it, from your perspective it sounds like the changes I should make to this PR are:
- Switching to the `timespan` metric type
- (tentative on resolving conversation above) Switch my logging implementation to use a new `performance` ping type

Is that right? Or are there other things you feel I should change?

mcomella:
> As I understand it, from your perspective it sounds like the changes I should make to this PR are:

@Dexterp37 I added a commit with this change: let me know if it looks right to you!

app/src/main/java/org/mozilla/fenix/perf/StartupTimeline.kt
```diff
@@ -44,4 +54,6 @@ object StartupTimeline {
     private fun advanceState(startingActivity: StartupActivity) {
         state = StartupTimelineStateMachine.getNextState(state, startingActivity)
     }
+
+    fun <R> measureGeckoRuntimeCreate(block: () -> R): R = measurements.geckoRuntimeCreate.measure { block() }
```
Dexterp37: Not really a problem, but more of a curiosity: is there any advantage in introducing this function over calling something like `measure(gleanMetric, block)`? It would save you the trouble of defining a function for each segment you want to measure.

app/metrics.yaml
```diff
@@ -1856,3 +1856,17 @@ installation:
     notification_emails:
       - fenix-core@mozilla.com
     expires: "2020-09-01"
+
+startup_timeline_telemetry:
+  gecko_runtime_create:
+    type: timespan
+    description: >
```
Dexterp37: As currently defined, this would be sent in the `metrics` ping only _once_. The metric will be cleared after the ping is collected and won't be available in other `metrics` pings that may be generated within this session.

If that's the intent, then nothing to do. If you want this metric to be available in most of the `metrics` pings being generated, then you need to add `lifetime: application` here. [See our docs](https://mozilla.github.io/glean/book/user/metric-parameters.html#optional-metric-parameters).

If you send this within a custom ping, then leaving the default lifetime may be fine.

app/metrics.yaml
```diff
@@ -1856,3 +1856,17 @@ installation:
     notification_emails:
       - fenix-core@mozilla.com
     expires: "2020-09-01"
+
+startup_timeline_telemetry:
```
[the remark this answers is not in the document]
mcomella: My concern is there is a namespace collision: we already have a class called `StartupTimeline` so anyone who imports or reads code with either class needs to disambiguate: I'd rather they were named uniquely.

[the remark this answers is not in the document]
mcomella: @Dexterp37 Do you have suggestions on what else I can call this (or the other `StartupTimeline` class) to prevent a namespace collision?

app/src/main/java/org/mozilla/fenix/perf/DurationContainer.kt
```diff
@@ -0,0 +1,101 @@
...
+ * This class helps ensure data integrity by making guarantees about how it's supposed
+ * to be used since usage outside of these expectations is often indicative of faulty
+ * assumptions or programming errors. We crash in these cases, rather than sending an
+ * error to the server, to reduce the feedback time between change in implementation and
+ * new errors, to increase the likelihood the errors are noticed and fixed rather than
+ * being used for bad data analyses, and to reduce the amount of bad data hitting the server
+ * since it would complicate data analysis and make it more error-prone. It's unfortunate to
+ * crash crash over data collection concerns but the trade-offs seem worth it.
+ *
+ * One large downside to these guarantees is that this class will become harder to use.
```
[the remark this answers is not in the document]
mcomella: It's purely theoretical: with complex classes sometimes devs write their own simple version rather than figuring out how to use the complex one. Then again, those classes are usually much larger than this one.

mcomella:
> Measuring is hard, the more we use a shared set of tools that were tested the less likely it is for things to be broken :-) It's a general advice, as we've jumped to quite some hoops in order to get timing right.

This is a good point: I've felt this a lot with performance. 🙄 

> Is this case very common? I'd still encourage you to go through the custom ping path and add an extra measure (or more), rather than going custom.

Good point, it's probably not common. I agree I'd rather use the existing tools than maintain our own measurement classes and tools.

I don't think going through Glean is the most intuitive – i.e. adding measurements to a ping and logging the ping to get the measurements locally – but it sounds like it fits our problem pretty well. I'll create a PoC with a custom ping and see how it goes.

app/metrics.yaml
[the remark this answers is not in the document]
mcomella: Note: names like `startup.timeline` don't seem to work with the Glean parser in fenix:
```
      'performance.ping' is not valid under any of the given schemas
          'performance.ping' does not match '^[a-z][a-z0-9-]{0,29}$|^deletio
          n_request$|^bookmarks_sync$|^history_sync$|^session_end$|^all_ping
          s$|^glean_.*$'
          'performance.ping' is not one of ['$schema']
```

app/src/main/java/org/mozilla/fenix/perf/DurationContainer.kt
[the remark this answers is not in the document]
mcomella: Removed this class so nevermind. :)

app/src/main/java/org/mozilla/fenix/perf/StartupTimeline.kt
mcomella: I love this solution! Great idea!

mcomella:
@Dexterp37 What do you think of the new solution? This feels a lot better to me and removed a lot of unnecessary code – thanks for the suggestion and understanding my use case. :)

app/metrics.yaml
```diff
@@ -1856,3 +1856,21 @@ installation:
     notification_emails:
       - fenix-core@mozilla.com
     expires: "2020-09-01"
+
+startup.timeline.telemetry:
```
Dexterp37:
```suggestion
startup.perf.timeline:
```

I see your point about naming confusion. The metrics are generated within the `org.mozilla.fenix.GleanMetrics` package, so they would end up in `org.mozilla.fenix.GleanMetrics.StartupTimeline`. In the future, having `telemetry` as a suffix will produce a build time error :-)

app/src/main/java/org/mozilla/fenix/perf/StartupTimeline.kt
```diff
@@ -44,4 +51,66 @@ object StartupTimeline {
...
+ * A [LifecycleObserver] for [HomeActivity] focused on startup performance measurement.
+ */
+class StartupHomeActivityLifecycleObserver : LifecycleObserver {
+    @OnLifecycleEvent(Lifecycle.Event.ON_STOP)
+    fun onStop() {
+        // Startup metrics placed in the Activity should be re-recorded each time the Activity
+        // is started so we need to clear the ping lifetime by submitting once per each startup.
+        // It's less complex to add it here rather than the visual completeness task manager.
+        //
+        // N.B.: this submission location may need to be changed if we add metrics outside of the
```
Dexterp37: I'm a bit confused by this: metrics are cleared whenever you send a ping (`ping` lifetime). You should only be concerned in sending the ping after the metrics are set: startup only happens once per session anyway, right?

app/pings.yaml
```diff
@@ -29,3 +29,21 @@ installation:
     - https://github.com/mozilla-mobile/fenix/pull/8074#issuecomment-586512202
   notification_emails:
     - fenix-core@mozilla.com
+
+startup-timeline:
+  description: >
+    This ping is intended to provide an understanding of startup performance. It is
+    currently not submitted to the server but is intended to be provided as a proof-
```
Dexterp37: How are you making sure this doesn't get submitted? Because you're calling `Pings.startupTimeline.submit()` so, if Glean is initialized and upload is enabled, it will be sent. Were you planning on making this ping only available in the "performancetest" flavour? (maybe it already is, I'm just not seeing this in this PR).

If that's the case, would you kindly be more specific in this comment?

app/metrics.yaml
mcomella: How about `startup.timeline.metrics` or `startup.timeline.measurements`? I think the existing `StartupTimeline` could easily be confused with `StartupPerfTimeline`, e.g. when looking through an autocomplete menu.

fwiw, in Java, I don't think it's uncommon for the package name to have redundancies with the class name because the class is often referenced without the package. For example, `androidx.fragment.app.Fragment`.

app/pings.yaml
mcomella: That was a mistake, oops! I wrote this before I added ping submission. I intend to make it available everywhere: I'll update the comment.

app/metrics.yaml
Dexterp37: `metrics` and `measurements` would still be problematic, [see our docs](https://mozilla.github.io/glean/book/user/adding-new-metrics.html#be-as-specific-as-possible).

Telemetry, metrics and measurements are all a bit redundant :-)

mcomella: I see: it's redundant in the end-to-end system:

> It is not necessary to put the type of the metric in the category or name, since this information is retained in other ways through the entire end-to-end system.

But from a developer perspective it's not redundant when reading only the Java class name. One way to resolve this problem is that Glean's generated methods could add one of these endings, e.g. in YAML: `startup.timeline` = in Java `StartupTimelineMetrics`.

That being said, this is what we have now so maybe I'll stick with `StartupTimeline` (as the most accurate value and shortest for upload through the pipeline) and educate the small number of developers using the code that there's a namespace collision.

mcomella: Made the changed but keeping unresolved so we can continue the conversation.

app/src/main/java/org/mozilla/fenix/perf/StartupTimeline.kt
mcomella: This is currently placed in the `HomeActivity.onStop` method but if the user enters the app through a separate Activity and closes the app without seeing HomeActivity, this will never be called. At some point, we'll probably want to measure these alternative use cases.

After we're more invested in startup telemetry measurement, we should probably move it to work for startup of every activity. I didn't want to do the obvious thing right now (call it in onStop of all activities because it's unclear if other activities can start before the previous one stops, which would break the lifecycle).

I added a brief comment to elaborate. Please resolve if this makes sense.

mcomella:
> Ideally, you'd really want to not log and just send pings to a local http endpoint.

@Dexterp37 To view the data...
- For automation, setting up a local http endpoint makes sense.
- For local debug builds, what's the best way? The Glean Debug View, maybe?

mcomella:
@Dexterp37 A few more questions for you:
- When I use the GleanDebugActivity in Debug builds, the pings send instantly. However, they don't seem to send immediately in fennecBeta builds (in fact, I couldn't get them to send at all): do you know why that is?
- If we want to measure some values in automation that are not uploaded and have some others be uploaded, is having two pings – one that is uploaded (`StartupTimeline`) and one that is not (`StartupTimelineNoUpload`) – be a reasonable solution?

mcomella:
@Dexterp37 More questions!
- Do we care to shorten probe names or lower granularity of returned values in order to minimize data usage? I notice my naming schemes are pretty long
- Do we have plans in place to prevent a user from sending too much data? e.g. do we know how much data we're sending per user?

app/metrics.yaml
Dexterp37: > > It is not necessary to put the type of the metric in the category or name, since this information is retained in other ways through the entire end-to-end system.
> 
> But from a developer perspective it's not redundant when reading only the Java class name. One way to resolve this problem is that Glean's generated methods could add one of these endings, e.g. in YAML: `startup.timeline` = in Java `StartupTimelineMetrics`.

That's true. If we were to allow this I'm afraid we'd need to consider all the other platforms/bindings the Glean SDK is currently available on. Some bindings might not support aliasing at all, I'm afraid :-( On a quick look, it seems that complexity outweights advantages on this one (but I might be wrong).

Please kindly [file a bug](https://bugzilla.mozilla.org/enter_bug.cgi?product=Data+Platform+and+Tools&component=Glean%3A+SDK) about this if you feel strongly about it :-) 

> That being said, this is what we have now so maybe I'll stick with `StartupTimeline` (as the most accurate value and shortest for upload through the pipeline) and educate the small number of developers using the code that there's a namespace collision.

Another option that we did not consider was to rename `StartupTimeline.kt` to somethine like `StartupTimelineMonitor.kt` (or similar). Would this make it clearer?

mcomella:
>  Do you have a specific concern for which you might not want to add the value to the ping?

I imagine there are some values we'll want to log forever locally and in automation to make it convenient for developers to identify regressions in portions of the code but that might not provide value via telemetry. For example, we might find that `GeckoRuntime.create()` takes a trivial amount of time on all devices and having a probe for it provides little value. However, we might still want to keep it in automation so that if there's a regression, the perf sheriffs can alert us and we fix it.

In theory, we could have a lot of these probes (e.g. if we broken up startup into multiple milestones, they may not all provide value via user telemetry).

That being said, is it a big deal if we're collecting say, two extra duration values? Six? A dozen? At what point is it worth our time, server costs, and burden to the users' data plans to design an alternative code path?

> Any metric that you define in metrics.yaml does not get automatically sent in your pings unless their value is set.
With that said, I think you can stick to one custom ping and just not set the value? (unless I'm misuderstanding the problem)

This _could_ work though I'm concerned it doesn't quite fit the use case. The use case I see is that these durations are logged in _all_ builds, including release, but are not uploaded to the server. We should be able to access their results in automation and locally (either they're always available, e.g. via logcat, or available on-demand such as the DebugActivity intent).

My concern with "just not setting the value" is that it sounds complex to implement or doesn't fit the requirements well, e.g.:
- A new runtime intent won't be available before startup so either we need to always set the values outside of glean and, when the intent is received, only then set the values in glean OR set a shared pref when the intent is received and set the values on the subsequent run but reading from shared prefs will impact performance
- A build-time flag doesn't let us easily get the data from _all_ builds

---

I'm figuring out fennecBeta & will get back to you

mcomella:
> Are you, by any chance, explicitly using sendPing in the first case? For your specific case, you should not do that, but rather wait for your activity to trigger the submission.

I must have been doing that. It appears:
```
adb shell am start -n "org.mozilla.firefox_beta/mozilla.telemetry.glean.debug.GleanDebugActivity" --ez logPings true --es tagPings performance
```

Works correctly. Thanks for your help!

---

@Dexterp37 One more question: the values we're logging are in nanoseconds, which isn't familiar to us (i.e. not human readable). We'd prefer to read milliseconds or seconds. Do you have a suggestion on the best way to convert the data when doing local development (e.g. looking at the Glean Debug View?). Perhaps we could write a browser extension to convert the raw JSON links from the Glean Debug Viewer or a script to download and parse them?

app/metrics.yaml
mcomella: I filed a bug https://bugzilla.mozilla.org/show_bug.cgi?id=1624066 though I'm feeling less sure about it because the name of the metric gets longer, making it more tedious to use. We should get more feedback.

> Another option that we did not consider was to rename StartupTimeline.kt to somethine like StartupTimelineMonitor.kt (or similar). Would this make it clearer?

We _could_ do this to avoid the collision but:
- `StartupTimeline` is a name from desktop so I wanted to keep them consistent
- `StartupTimelineMonitor` gets long and annoying to call statically (i.e. the same problem I realized above)
- If someone sees `StartupTimelineMonitor`, they might be confused what the class `StartupTimeline` does – does the monitor monitor this other class in some way?

That being said, thinking about it now, `StartupTimeline` is a confusing name on it's own and it should really be `StartupTimelineMonitor` 🤔. This is confusing for a Friday evening: I'll just leave everything where it is now. :)

Dexterp37: > > Another option that we did not consider was to rename StartupTimeline.kt to somethine like StartupTimelineMonitor.kt (or similar). Would this make it clearer?
>
> We _could_ do this to avoid the collision but:
> 
>     * `StartupTimeline` is a name from desktop so I wanted to keep them consistent

I would not stick to that too much. That might change as well, once glean comes to desktop.

>     * `StartupTimelineMonitor` gets long and annoying to call statically (i.e. the same problem I realized above)

That could be a problem, but do you expect to use this more than 10/15 times throughout the code?

>     * If someone sees `StartupTimelineMonitor`, they might be confused what the class `StartupTimeline` does – does the monitor monitor this other class in some way?

This could be eased by comments, no?

Dexterp37:
> @Dexterp37 One more question: the values we're logging are in nanoseconds, which isn't familiar to us (i.e. not human readable). We'd prefer to read milliseconds or seconds. Do you have a suggestion on the best way to convert the data when doing local development (e.g. looking at the Glean Debug View?). Perhaps we could write a browser extension to convert the raw JSON links from the Glean Debug Viewer or a script to download and parse them?

I'm afraid I don't have a good suggestion on this. Writing an extension for converting the things you care about might be a good option.

app/metrics.yaml
mcomella: > I would not stick to that too much. That might change as well, once glean comes to desktop.

That's fair.

> That could be a problem, but do you expect to use this more than 10/15 times throughout the code?

10/15 seems like the minimum: e.g. we'd likely instrument each of the major lifecycle methods in each major activity and fragment at the start and end of the method (that's ~16 already).

> This could be eased by comments, no?

True but it's better to avoid this if possible. :)

Dexterp37: > > That could be a problem, but do you expect to use this more than 10/15 times throughout the code?
> 
> 10/15 seems like the minimum: e.g. we'd likely instrument each of the major lifecycle methods in each major activity and fragment at the start and end of the method (that's ~16 already).

This could further be eased by import aliasing, e.g. `import org.mozilla.fenix.perf.StartupTimelineMonitor as stm` or something like that if we come to that, right?

mcomella:
Closing in favor of https://github.com/mozilla-mobile/fenix/pull/9788, which adds a probe we're actually interested in.

PR #9168, opened 2020-03-15
WIP: (do not land) re-enable ignored UI tests for recent Firebase runs

AaronMT:
Re-enable all currently commented `@Ignore` tests (aside from stubs and those which can not run on virtual device) to see which are still failing on a recent re-base of master. Tracking will be in a [spreadsheet](https://docs.google.com/spreadsheets/d/14khbZJmgDnqm1hKfWsHnE_0a83hzRmFAax2W8tE5HyQ/edit?usp=sharing). Runs will fire off currently just on manual re-base.

(CC @rpappalax @isabelrios)

### DIFF
--- a/app/src/androidTest/java/org/mozilla/fenix/ui/BookmarksTest.kt
+++ b/app/src/androidTest/java/org/mozilla/fenix/ui/BookmarksTest.kt
@@ -11,7 +11,6 @@ import mozilla.appservices.places.BookmarkRoot
 import okhttp3.mockwebserver.MockWebServer
 import org.junit.After
 import org.junit.Before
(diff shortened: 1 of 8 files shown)


AaronMT:
Going to start with a fresh PR/branch for a re-run.

liuche:
I think we should re-enable some of these tests if we can! @AaronMT could we break this into pieces if that makes it easier to re-enable?

AaronMT:
I forgot about this one. We're in the habit of enabling and investigating these separately now with custom soak-testing configurations for Firebase. I'll close this one as it was a collective, now out of date.

Issue #7726, opened 2020-01-15
[Bug] Crash when tapping on "Retry" when the tab wasn't sent


sflorean:
## Prerequisites
Sign in with the same account on Desktop and on Firefox Preview.

## Steps to reproduce
1. On Android, open a webpage and tap Share from the main menu (don't tap on the account displayed).
2. On Desktop - your account - Manage account and change the password.
3. Immediately after the password is changed, go to Android and tap on the desktop account (displayed at step 1).
4. Tap on "Retry" from the snackbar. 

### Expected behavior
"Unable to send RETRY" toast is presented. Same as for step 3. 

### Actual behavior
Firefox Preview crashes.

### Note
On Nightly, after performing step 4, no error is displayed again. 

### Device information
* Android device: Xiaomi MiPad 2 (Android 5.1), OnePlus 5T (Android 9).
* Fenix version: 3.1.0 - beta 3

[retrycrash.txt](https://github.com/mozilla-mobile/fenix/files/4065692/retrycrash.txt)

Mugurell:
Could not reproduce the crash but I see another issue on various devices: 
After pressing retry the current snackbar is dismissed but another new one isn't shown.
Trying to track this issue down was a rabbit hole but with enough logs in our code and in the `com.google.android.material.snackbar` package I got to see that
- when user presses retry [we dismiss the current snackbar](https://github.com/mozilla-mobile/fenix/blob/b0e4453fb7ba558a6308158e192c5ebfa9dfafb4/app/src/main/java/org/mozilla/fenix/components/FenixSnackbar.kt#L70) and the dismiss animations starts
- if the same action fails we again [showFailureWithRetryOption()](https://github.com/mozilla-mobile/fenix/blob/b0e4453fb7ba558a6308158e192c5ebfa9dfafb4/app/src/main/java/org/mozilla/fenix/share/ShareController.kt#L162)
[...]

PR #9171, opened 2020-03-16
For #7726 - Support displaying consecutive Snackbars

Mugurell:
This flow does not seem supported by the underlying "material.snackbar"
library. When we try to show a new Snackbar the library will update the
properties of the allready on-screen one. While still dismissing it because we
are trying to show a new one.

To properly support our usecase we need to queue the call to show a new
Snackbar and effectively wait until the previous Snackbar is dismissed and
removed from the screen to only then show a new one.

### Pull Request checklist

- [x] **Tests**: This PR includes thorough tests as it adds a small modification to a not yet tested component.
- [x] **Screenshots**: This PR includes does not includes screenshots. The end result is clear.
- [x] **Accessibility**: The code in this PR follows does not include any user facing accessibility feature.

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/FenixSnackbar.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/FenixSnackbar.kt
@@ -28,6 +28,7 @@ class FenixSnackbar private constructor(
     contentViewCallback: FenixSnackbarCallback,
     isError: Boolean
 ) : BaseTransientBottomBar<FenixSnackbar>(parent, content, contentViewCallback) {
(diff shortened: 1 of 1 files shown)


liuche:
Looking at the linked issue, this is a bit of an edge case. However, this workaround seems to add a lot of complexity.
I appreciate all the investigation! The documentation in the bug is helpful. If we see this in another snackbar that is a major problem, we should revisit this solution.

For now, I'm going to close this PR, since it's not directly fixing the crash in the issue, and adds complexity.

Issue #220, opened 2019-01-21
FNX-511 ⁃ Add ability to change Firefox Preview (browser UI) language


bbinto:
### Why/User Benefit/User Problem
As a user, I want to experience the app in a language other than the system / phone language

### Deps
* [x] AC API for switching languages

### Acceptance Criteria (how do I know when I’m done?)
* [ ] I have the ability to change browser UI language separate from the system language
* [ ] Browser language will follow the system language by default

**Mocks**: https://miro.com/app/board/o9J_kwN9u-I=/
SVG asset for Language icon: https://dzwonsemrish7.cloudfront.net/items/1j0I1L1E282t24241N0l/language.svg

#### NOT in this issue
* #6502 Allow webpage content language selection
* expanding the locales we currently support #5496 

**This is just scoped to making the picker inside of Fenix, NOT inside AC**

cpeterso:
AFAICT, Chrome's Language settings only configure the user's preferred language for web content (i.e. the `Accept-Language` HTTP header sent to web servers), not Chrome's app UI language.

GV bug https://bugzilla.mozilla.org/show_bug.cgi?id=1543823 is about letting the browser customize Gecko's `Accept-Language` HTTP header and/or making Gecko use the Android system language.

UX question to consider:
1. Assume Android system language is en-US.
2. User changes Fenix's language setting to en-CA.
3. User then changes the Android system language to en-GB.
4. Should Fenix's language setting preserve its non-default setting (en-CA) or switch to new system language (en-GB)?

yoasif:
> 1.     Assume Android system language is en-US.
> 2.     User changes Fenix's language setting to en-CA.
> 3.     User then changes the Android system language to en-GB.
> 4.     Should Fenix's language setting preserve its non-default setting (en-CA) or switch to new system language (en-GB)?

It feels to me like Fenix should be en-GB and and websites should be en-CA *unless* Fenix implements separate website and browser UI preferences for language. 

Also, FWIW, Chrome also allows users to set up several languages based on priority, while Fennec only allows users to select a single one (likely because the preference applies to both UI and websites).

sblatz:
@vesta0 are we able to get a UX/product decision on this:

(From @cpeterso):
> This issue is about "browser language", but does that just affect the app UI's language? Or should it also change the user's web content language preference (i.e. Gecko's `Accept-Language` HTTP header sent to servers)?
[...]

PR #9191, opened 2020-03-17
For #220 - Removes feature flag for language picker

boek:


ekager:
Is this waiting on looking into #7559?

liuche:
We should make sure this other bug is fixed #9413

Issue #9252, opened 2020-03-19
[Bug] "The logins you save or sync to Firefox Nightly will show up here" text is displayed right before all the accounts are displayed


abodea:
## Steps to reproduce
1. Launch Fenix.
2. Login with valid credentials
3. Access logins and passwords section.
4. Tap on the saved logins.
### Expected behavior
The saved logins are displayed.
### Actual behavior
"The logins you save or sync to Firefox Nightly will show up here" text is displayed right before all the accounts are displayed.
### Device information

* Android device: **Samsung Galaxy S7 (Android 7).**
* Fenix version: **Nightly 19/3.**

![VID-20200319-140141](https://user-images.githubusercontent.com/42831109/77077391-d5a19400-69fd-11ea-8af9-1c2163952c63.gif)

abodea:
Verified as fixed on the latest Nightly **3/30** with **Google Pixel 4 XL (10), Samsung Galaxy S7 (7).**
Note that everything works as expected and the text is not displayed anymore while logged in.

PR #9277, opened 2020-03-20
For #9252: Update views visibility ony when list has finished loading

mcarare:
### Pull Request checklist

- [x] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features. 

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/settings/logins/SavedLoginsView.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/logins/SavedLoginsView.kt
@@ -69,8 +69,10 @@ class SavedLoginsView(
     }
 
     fun update(state: SavedLoginsFragmentState) {
-        view.saved_logins_list.isVisible = state.items.isNotEmpty()
-        view.saved_passwords_empty_view.isVisible = state.items.isEmpty()
+        if (!state.isLoading) {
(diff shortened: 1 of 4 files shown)


ekager:
This is a bit out of the scope of the bug... But could you possibly add a ProgressBar for the loading state like we have in history if we are going to add this `loading` state? Thank you!

app/src/main/java/org/mozilla/fenix/settings/logins/SavedLoginsView.kt
```diff
@@ -13,7 +13,9 @@ import android.widget.FrameLayout
 import androidx.core.view.isVisible
 import androidx.recyclerview.widget.LinearLayoutManager
 import kotlinx.android.extensions.LayoutContainer
+import kotlinx.android.synthetic.main.component_history.view.*
```
ekager: I don't think we want this import from history?

mcarare: Not sure how this escaped ktlint 😯 . Thanks for the review! 😄

Issue #8759, opened 2020-02-26
[Bug] "Recently Used" section in Share menu occupies too much space


SL-Lee:
# Steps to reproduce

1. Open the share menu and tap on an icon. Repeat this step for several times.

# Expected behavior

The "Recently Used" section occupies only a small portion of the available space in the Share menu.

# Actual behavior

The "Recently Used" section occupies majority of the space, resulting in the "All Apps" section having a ridiculously little amount of space.

![](https://user-images.githubusercontent.com/45960387/75329800-43edad80-58bb-11ea-934b-0addb48c0924.png)

# Device information

* Android device: Oppo Reno 2, Android 9
* Fenix version: Nightly 200226 06:01 (Build #20570605)

mcarare:
Question for UX: Should we limit the space for "Recently used "(while keeping it scrollable), or the current behaviour is the one expected? TY!

dep-dt:
I'd love a simple long-press on a pinned share item to remove it. that'd help.

ekager:
If I understand the problem correctly, I think the recently used app section should "scroll with" the other list? Does that make sense/is possible?

brampitoyo:
Would it help if the amount of recent apps/actions are reduced to 6? That way, you’ll have space for at least 2 apps in the “All” list.

dep-dt:
> Would it help if the amount of recent apps/actions are reduced to 6? That way, you’ll have space for at least 2 apps in the “All” list.

I think 6 is a good max. maybe even 4.  Max of 6 with longpress to remove would solve the problem for me.

dep-dt:
> If I understand the problem correctly, I think the recently used app section should "scroll with" the other list? Does that make sense/is possible?

I like @ekager's idea as well, in addition to a sensible max.  👍

brampitoyo:
Great. That behaviour works for me. Thanks @ekager @dep-dt for the ideas and inputs!

Let’s decide on:

- Having 6 recently used actions
- The “Recently used” section should be able to scroll horizontally, so “All apps” can be accessed

DrCesar:
I like to give this a shot. Currently limited to 6 items and trying to make the section horizontally scrollable. Any tip is welcome.

brampitoyo:
@mcarare @ekager Hmm. Apparently the Share sheet isn’t yet implemented according to spec. See #9332.

The spec doesn’t have “Recently used” or “All actions”. Instead, it calls for a section called “Share a link”.

[...]

Issue #9332, opened 2020-03-23
[Share] Share sheet implemented not according to spec


AmyYLee:
Hi, 

The share sheet has has horizontal scroll instead of a vertical scroll layout. Please reference original spec for guidance (@topotropic):

https://mozilla.invisionapp.com/share/PSQFWZ1Y4W8#/screens/346059768

1. "Recently Used" should be it's own row with no grey background
2. "All Actions" should be the next row under it separated by a hairline above with vertical scroll (like native android share sheet). 

<img width="417" alt="Screen Shot 2020-03-23 at 16 52 12" src="https://user-images.githubusercontent.com/5992746/77362181-e8320b00-6d26-11ea-810d-c6c8e571d2c7.png">

brampitoyo:
Clarification question for @AmyYLee and @topotropic: does the “Share a link” section have a certain ordering of items?

For example, “Copy link” seems to be the first item. Is it always the first item in all situations?

The app icons that come after “Copy link”, do they get sorted by recently used, alphabetical, etc.?

cadeyrn:
https://mozilla.invisionapp.com/share/PSQFWZ1Y4W8#/screens/346059768 is an outdated spec, no? It still shows the "sessions" concept which no longer exists in Fenix, so it's a very old mockup. This mockup does not have a section for "recently used" at all. This feature was added in #4231 and there is a much newer mockup: https://app.abstract.com/share/e1a39a6b-2c3e-4fc4-b46c-0c78ec491112?mode=design&sha=f3bcad9f449216555fbf5cc31c3b9933b6538a96

AmyYLee:
> https://mozilla.invisionapp.com/share/PSQFWZ1Y4W8#/screens/346059768 is an outdated spec, no? It still shows the "sessions" concept which no longer exists in Fenix, so it's a very old mockup. And this mockup does not have a section for "recently used" at all. This feature was added in #4231 and there is a much newer mockup: https://app.abstract.com/share/e1a39a6b-2c3e-4fc4-b46c-0c78ec491112?mode=design&sha=f3bcad9f449216555fbf5cc31c3b9933b6538a96

@topotropic - Can you provide feedback/guidance? It looks like this was a temporary fix.

mcarare:
@topotropic I think we should also consider having the option to remove apps from recent list.

ekager:
Yeah and the "recently used apps" section was added very recently but maybe was based on an older spec?
[...]

PR #9344, opened 2020-03-24
Issue 8759 recently used.

DrCesar:
For #8759
#### Tests
I run the pre-push hooks and fixed all the linting errors, but i received an error in ShareControllerTest.kt
tests
error-tests-share-controller

#### Screenshots
The changes for the share app view, i made the view horizontally scrollable and added the feature to remove a recently share app option.

share-apps-scroll
share-apps-before-remove
share-apps-after-remove


mcarare:
@dragneelfps Thank you for putting up this PR! There are some things we must consider before:

- there is already a change made in android components to use activityName instead of packageName, and after https://github.com/mozilla-mobile/fenix/pull/9317/files will be merged the change will be also in AC.
 - regarding the UI changes you made, we should hold of on implementing them, until a decision has been made here: https://github.com/mozilla-mobile/fenix/issues/9332.

We'll keep in touch, and thank you again for your involvement!

liuche:
This is definitely broken and we should fix it, so let's get the fix in and get the final spec changes later.

ekager:
Closing for https://github.com/mozilla-mobile/fenix/pull/11077

PR #9363, opened 2020-03-24
Closes #9333: Prevent compression of omni.ja to improve startup time

### LINKED ISSUE #9333: [Bug] Fenix should specify `noCompress 'ja'`
The `omni.ja` file contains all the javascript code that GeckoView uses internally. Right now the `omni.ja` file is compressed in the APK. This causes GeckoView to uncompress the omni.ja file in memory before it can start doing anything else. This takes a long time, on some profiles it delays startup by about 400ms.

We should specify `noCompress 'ja'` in the `aaptOptions` to avoid that.

See also: https://bugzilla.mozilla.org/show_bug.cgi?id=1623758

csadilek:


app/build.gradle
```diff
@@ -169,6 +169,14 @@ android {
         }
     }
 
+    aaptOptions {
+        // All JavaScript code used internally by GeckoView is packaged in a
+        // file called omni.ja. If this file is compressed in the APK,
+        // GeckoView must uncompress it before it can do anything else which
+        // causes a significant delay on startup.
+        noCompress 'ja'
```
Amejia481: Should we document this on AC?

csadilek: Yes, GV and A-C should document this. We should put this in our README of browser-engine-gecko (all 3).

Amejia481: 👍

Issue #9391, opened 2020-03-25
[Bug] Accessing Settings without network connection causes Fenix to crash


SL-Lee:
## Steps to reproduce

**Prerequisite: Firefox sync must be turned on.**

1. Make sure Fenix is closed (using the quit button in the menu), then disconnect the device from any network.
2. Open Fenix and access settings.

### Expected behavior

Fenix opens settings.

### Actual behavior

Fenix crashes.

**Other Notes**

* After performing the STR, restart Fenix and turn on the network connection. Go to settings > account. The device name is reset to the default value (might be related to #7199).

  ![](https://user-images.githubusercontent.com/45960387/77521657-0120f480-6ebe-11ea-8fe8-ddbf71d124fb.gif)

  Tapping on it will bring up a dialog box with the changed name, and performing a sync will change it back. But this doesn't work reliably -- I've seen occurrences where the device name resets to the default value.

### Device information

* Android device: Oppo Reno 2, Android 9
* Fenix version: Nightly 200325 06:01 (Build #20850609)

grigoryk:
This is fixed by https://github.com/mozilla-mobile/fenix/pull/9532

PR #9397, opened 2020-03-25
For #9391: Consider network connectivity on updating account UI

mcarare:
### Pull Request checklist

- [x] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features. 
pass

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/settings/SettingsFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/SettingsFragment.kt
@@ -7,9 +7,13 @@ package org.mozilla.fenix.settings
 import android.content.ActivityNotFoundException
 import android.content.Context
 import android.content.Intent
+import android.net.ConnectivityManager
+import android.net.Network
+import android.net.NetworkRequest
 import android.net.Uri
 import android.os.Bundle
 import androidx.appcompat.content.res.AppCompatResources
+import androidx.core.content.getSystemService
 import androidx.lifecycle.lifecycleScope
 import androidx.navigation.NavDirections
 import androidx.navigation.findNavController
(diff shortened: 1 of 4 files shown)


mcarare:
@topotropic I think we should show the user the info in the attached screenshot when there is no network connectivity. 
The icon and text remains to be approved. Can you help with this, or redirect to someone else? TY!

topotropic:
@betsymi can you check the copy here?

I think for an icon we should align with what we use here https://share.goabstract.com/66058508-b4ec-4b44-a82e-cec3de7e5446?mode=design&sha=c56b553b26dfeefc55e704ece2ff2d6f0673d592 

Thanks for reaching out!

mcarare:
@topotropic I have used the same icon we use in the share sheet when we are offline and cannot show list of devices. But I will change it to the other one, if you think it's more suggestive! Thank you for your input!

grigoryk:
I don't think we should do this.

Issue that causing the crash (see https://github.com/mozilla-mobile/fenix/issues/9391, https://github.com/mozilla-mobile/fenix/issues/9530) was resolved is resolved (by https://github.com/mozilla-mobile/fenix/issues/9530). So, this change doesn't resolve an immediate problem anymore.

In my experience, Android's network availability notifications could be quite flaky, and their behaviour differs quite a bit based on the device, API version, etc. I have a feeling we could be chasing bugs around this conditional for years if we land this :) Maybe it got better now, but we support API21+, so...

Besides, devices could go offline for short periods of time, forcing an unlucky user to go in-and-out of the Settings UI if they're trying to simply open account settings while on a flaky connection.

ekager:
cc @topotropic knowing that the crash has been fixed do you still think we should do this?

> 
> Issue that was causing the crash (see #9391, #9530) was resolved (by #9532). So, this change doesn't help with an immediate problem anymore. Just from this alone, I'd vote for less code/complexity.
> 
> In my experience, Android's network status/availability notifications could be quite flaky, and their behaviour differs quite a bit based on the device, API version, etc. I have a feeling we could be chasing bugs around this conditional for years if we land this :) Maybe it got better now, but we support API21+, so I wouldn't bet on it.
> 
> Besides, devices could go offline for short periods of time, forcing an unlucky user to go in-and-out of the Settings UI if they're trying to simply open account settings while on a flaky connection.

mcarare:
Alternative solution was implemented.

Issue #5208, opened 2019-09-11
[Bug] Opening a tab that was saved to a collection should restore the history as well


sv-ohorvath:
## Steps to reproduce
This is a regression, the initial functionality was added in #3215.
Same STR.
### Expected behavior
When restoring a tab with history, you can navigate back/forward to previously visited pages.
### Actual behavior
Can't navigate back/forward in the restored tab from a collection.
### Device information

* Android device: Pixel (Android 10)
* Fenix version: RC 1.4.1, Nightly 9/11

obarzpaw:
I would like to do some investigation.

darkwing:
Can confirm this is still a problem.

ekager:
Filed https://github.com/mozilla-mobile/android-components/issues/5529

topotropic:
@ekager is the AC part done here?

Diana-Rus:
Hi, issue fixed verified on Firefox Preview Nightly 200327 (Build #20870627) GV: 76.0a1-20200325093457 with:
- Google Pixel 3 XL (Android 9) ✔️
GIF:
![20200327-111620](https://user-images.githubusercontent.com/49859863/77741108-03e62b80-701d-11ea-9c60-9652436597de.gif)
- Samsung Galaxy S9 (Android 8) ✔️
- Sony Xperia Z5 (Android 7) ✔️

PR #9399, opened 2020-03-25
Issue #5208: Use new tab/collection restore API.

pocmo:
This will fix restoring the "engine state" of tabs from tab collections (scroll state, history etc).

This first requires the following AC patch to land and be available in an AC nightly release:
https://github.com/mozilla-mobile/android-components/pull/6381


pocmo:
Okay, this should be ready now. :)

PR #9436, opened 2020-03-26
Update to new feature-media API.

pocmo:
We are about to land a refactoring of `feature-media` that will make it use `browser-state` instead of `browser-session` (https://github.com/mozilla-mobile/android-components/pull/6362). This is the Fenix counterpart. I'll update this PR with the matching AC Nightly version once the AC PR landed.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/collections/CollectionCreationView.kt
+++ b/app/src/main/java/org/mozilla/fenix/collections/CollectionCreationView.kt
@@ -20,8 +20,20 @@ import androidx.transition.AutoTransition
 import androidx.transition.Transition
 import androidx.transition.TransitionManager
 import kotlinx.android.extensions.LayoutContainer
-import kotlinx.android.synthetic.main.component_collection_creation.*
-import kotlinx.android.synthetic.main.component_collection_creation.view.*
+import kotlinx.android.synthetic.main.component_collection_creation.back_button
+import kotlinx.android.synthetic.main.component_collection_creation.collection_constraint_layout
+import kotlinx.android.synthetic.main.component_collection_creation.name_collection_edittext
+import kotlinx.android.synthetic.main.component_collection_creation.save_button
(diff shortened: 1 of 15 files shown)


pocmo:
This is now ready too. :)

buildSrc/src/main/java/AndroidComponents.kt
```diff
@@ -3,5 +3,5 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 object AndroidComponents {
-    const val VERSION = "38.0.20200326160141"
+    const val VERSION = "38.0.20200327101347"
```
Amejia481: @pocmo  could you use the latest ac version? `20200327130125` :)

pocmo: At the time of opening the PR this was the latest version. Can we leave the bumping to MickeyMoz? :)

Amejia481: Sure, let me re-open it :)

pocmo: It will open a new PR. No need to open it now as it will fail without this here landing. :)

app/src/main/java/org/mozilla/fenix/collections/CollectionCreationView.kt
```diff
@@ -20,8 +20,20 @@ import androidx.transition.AutoTransition
 import androidx.transition.Transition
 import androidx.transition.TransitionManager
 import kotlinx.android.extensions.LayoutContainer
-import kotlinx.android.synthetic.main.component_collection_creation.*
-import kotlinx.android.synthetic.main.component_collection_creation.view.*
+import kotlinx.android.synthetic.main.component_collection_creation.back_button
```
ekager: do you usually not use the star import for kotlin synthetics?

ekager:
Okay I understand now we land this and then AC bot updates will pass 👍

PR #9459, opened 2020-03-27
Update NotYetSupported add-ons copy

psymoon:
Do not land until after the release.


ekager:
After which release @psymoon ?

Amejia481:
> After which release @psymoon ?

The string is already in AC `38.0.20200329190103`

Amejia481:
I think is safe to land :)

PR #9462, opened 2020-03-27
Use relative package name for 'App' activity alias

jonalmeida:
The `${applicationID}` does ~n't~ include the suffix of variants which causes launching from Android Studio to fail. Using the relative package name solves this and is also in alignment with Fennec (see [here](https://searchfox.org/mozilla-esr68/rev/0ef234279fd0608a8fd6bb1f792d080087d638d8/mobile/android/base/AndroidManifest.xml.in#72-74)).

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -37,7 +37,7 @@
           inside their source set.
         -->
         <activity-alias
-            android:name="${applicationId}.App"
+            android:name=".App"
             android:targetActivity=".HomeActivity">
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />


jonalmeida:
This shouldn't affect migration/shortcuts, but I'll file an issue for QA to verify.

pocmo:
> Using the relative package name solves this and is also in alignment with Fennec 

It is the same as in the raw manifest. But the manifest that ends up in the APK is still different. In Fenix the relative package name gets translated to an absolute name afaik. But Fennec itself kept the relative one if I remember correctly.

jonalmeida:
In the merged manifest, I was seeing it resolve to `${applicationId}.debug.App`, which confused Android Studio at least.

It would try and launch: `${applicationId}.debug/${applicationId}.debug.App`
It really wanted: `${applicationId}.debug/${applicationId}.App`

I'll do some fennec testing to double-check. Thanks for flagging!

jonalmeida:
Maybe the alias should only exist on fennec variants.. 🤔

ekager:
@jonalmeida do you think we still want to do this? Do we think the variant cleanup would help?

jonalmeida:
I'll close this PR. I don't think it's valuable right now.

PR #9555, opened 2020-03-31
Closes #9553: Only update account UI state once when creating SettingsFragment 

### LINKED ISSUE #9553: Don't update account state UI twice when creating SettingsFragment
We update this piece of UI at three points:
- in `onCreate`
- in `onResume`
- whenever `AccountObserver` callbacks are triggered

When SettingsFragment is being created (e.g. when user opens the settings ui), we will end up initializing this bit of UI twice, while we only need to do it once. For a logged in user, cost here is doing work on the background thread to re-fetch an avatar (it's cached as of https://github.com/mozilla-mobile/fenix/pull/9532, but we will still go pretty deep into GeckoFetchClient to actually hit the on-disk cache), process it into a drawable, etc.

grigoryk:
To track the states correctly, I've added a `creatingFragment` flag which is reset at the end of `onResume`. This doesn't feel great, but it's simple and easy to understand.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/settings/SettingsFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/SettingsFragment.kt
@@ -68,6 +68,11 @@ class SettingsFragment : PreferenceFragmentCompat() {
         override fun onAuthenticationProblems() = updateAccountUi()
     }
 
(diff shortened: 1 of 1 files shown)


grigoryk:
`ui-test-x86-debug` failure is strange, logs don't have a link to the test matrix. Maybe something went wrong in the CI? Let's re-run...

PR #9592, opened 2020-04-01
No issue: Annotate scanButtonTest (UI) with RequiresDevice

AaronMT:
https://developer.android.com/reference/androidx/test/filters/RequiresDevice

This test requires a physical device.


AaronMT:
Trying once more. FTL might not like this. If so, I'll close this and make a note.

PR #9606, opened 2020-04-01
Fix #9461 - Re-trigger taskcluster jobs on dep download fail

### LINKED ISSUE #9461: Provide solution/handler for intermittent taskcluster job dependency download failures
In an effort to better handle intermittent taskcluster job failures, we'd like to address the issue where dependencies temporarily fail to download by retriggering the job. This will handle the most common case where there was a network interruption or momentary service outage (but probably not the case where there was a longer service outage).

@JohanLorenzo mentions we can do this in the following way:
[1] capture exit code returned by gradle 
[2] update the task payload to rerun the task whenever that same exit code is returned 

@JohanLorenzo is there a good example of updating task payload we could follow to implement this?

rpappalax:
Fixes #9461 
### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/taskcluster/ci/build/kind.yml
+++ b/taskcluster/ci/build/kind.yml
@@ -40,6 +40,7 @@ job-defaults:
         docker-image: {in-tree: base}
         max-run-time: 7200
         chain-of-trust: true
+        retry-exit-status: [1]
 
 
 jobs:
--- a/taskcluster/ci/test/kind.yml
+++ b/taskcluster/ci/test/kind.yml
@@ -31,6 +31,7 @@ job-defaults:
     worker:
         docker-image: {in-tree: base}
         max-run-time: 7200
+        retry-exit-status: [1]
 
 jobs:
     debug:


rpappalax:
> From a taskgraph point of view, this patch is the right one! I don't know if we need to catch any other exit code. `1` seems quite generic to me and we may end up rerunning genuine failures.
> 
> That said, let's give it a try and see how it goes!

Will do, thanks, @JohanLorenzo!  would there be an `easy` of distinguishing re-running genuine failures vs. a retry (so we wouldn't do that)?  I assume this option quits after x number of attempts?

JohanLorenzo:
When `retry-exit-status` is set, tasks will stop retrying after 5 attempts. I remember the Taskcluster UI shows this counter: click on "See more" on the task view, then the second item shows "Retries left: X of 5". Treeherder doesn't but maybe displays these jobs with a different color (I need to check). 

From a programmatic point of view, I think there's no way to catch a genuine failure from a regular retry.

AaronMT:
I've seen deps fail to fetch on signed builds too.

JohanLorenzo:
Interesting! Do you have a link to one of these failures? I don't think the root causes are the same and I'd like to know what this is about.

AaronMT:
> Interesting! Do you have a link to one of these failures? I don't think the root causes are the same and I'd like to know what this is about.

Not right no, I think when I did (last week) I just rebased to get a new run triggered and things were good from there.

Issue #9605, opened 2020-04-01
Reduce unit test duplication by including custom test runner


mcomella:
Each unit test using Robolectric starts with:
```kotlin
@RunWith(RobolectricTestRunner::class)
@Config(application = TestApplication::class)
```

or 

```kotlin
@RunWith(AndroidJUnit4::class)
@Config(application = TestApplication::class)
```
(the latter delegates to the former). This can be replaced with:
```kotlin
@RunWith(FenixRobolectricTestRunner::class)
```

In the bad code samples above, I'm concerned folks might not understand what the code does and why it is necessary, which, in previous projects, I've seen lead to copy-pasting to attempt to solve problems. This is problematic because:
- robolectric non-trivially increases test runtime so we should only use it when it's really necessary
- copy-pasting to solve problems leads to tests with a lot of things that aren't strictly necessary and increase test complexity

Unifying this in one clear annotation may alleviate these issues, in addition to reducing duplication (a general coding principle).

mcomella:
Closed with https://github.com/mozilla-mobile/fenix/pull/9632

PR #9607, opened 2020-04-01
For #9605: remove duplication in @RunWith annotation; add lint check

mcomella:
The primary focus of this PR is to remove redundancy in how we specify `@RunWith` and `@Config` in our unit tests by replacing them with a single invocation. It adds a gradle task lint check to prevent it from regressing again.

Secondarily, the PR removes any trivially unnecessary `@RunWith` annotations, removing Robolectric from those tests. In aggregate, this increases the unit test suite runtime by ~10s (note: I didn't do any thorough measurement so this value is subject to be inaccurate but lines up with theoretical Robolectric start times of ~1s).

Example failure output of the task:
```
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':app:lintUnitTestRunner'.
> Lint failure: saw unexpected unit test runners. The following code blocks:

      @RunWith(AndroidJUnit4::class)
      @Config(application = TestApplication::class)
  OR
      @RunWith(RobolectricTestRunner::class)
      @Config(application = TestApplication::class)

  should be replaced with:

      @RunWith(FenixRobolectricTestRunner::class)

  To reduce redundancy of setting @Config. No @Config specification is necessary because
  the FenixRobolectricTestRunner sets it automatically.

  Relatedly, adding robolectric to a test increases its runtime non-trivially so please
  ensure Robolectric is necessary before adding it.

  The following files were found to be in violation:
      app/src/test/java/org/mozilla/fenix/library/bookmarks/BookmarkAdapterTest.kt
```

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/test/java/org/mozilla/fenix/ext/LogTest.kt
+++ b/app/src/test/java/org/mozilla/fenix/ext/LogTest.kt
@@ -10,13 +10,6 @@ import io.mockk.mockkStatic
 import io.mockk.verify
 import org.junit.Before
 import org.junit.Test
-import org.junit.runner.RunWith
-import org.mozilla.fenix.TestApplication
-import org.robolectric.RobolectricTestRunner
-import org.robolectric.annotation.Config
-
-@RunWith(RobolectricTestRunner::class)
-@Config(application = TestApplication::class)
 
 class LogTest {
 
--- a/app/src/test/java/org/mozilla/fenix/helpers/FenixRobolectricTestRunner.kt
+++ b/app/src/test/java/org/mozilla/fenix/helpers/FenixRobolectricTestRunner.kt
@@ -1,0 +1,32 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.helpers
+
+import org.mozilla.fenix.TestApplication
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.annotation.Config
+
+/**
+ * A test runner that starts Robolectric with our custom configuration for use in unit tests.
+ *
+ * usage:
+ * ```
+ * @RunWith(FenixRobolectricTestRunner::class)
+ * class ExampleUnitTest {
+ * ```
+ *
+ * IMPORTANT NOTES:
+ * - This should ALWAYS be used instead of RobolectricTestRunner and AndroidJUnit4 (note: the latter
+ * just delegates to the former)
+ * - You should only use Robolectric when necessary because it non-trivially increases test duration.
+ */
+class FenixRobolectricTestRunner(testClass: Class<*>) : RobolectricTestRunner(testClass) {
+
+    override fun buildGlobalConfig(): Config {
+        return Config.Builder()
+            .setApplication(TestApplication::class.java)
+            .build()
+    }
+}
--- a/app/src/test/java/org/mozilla/fenix/helpers/FenixRobolectricTestRunner.kt
+++ b/app/src/test/java/org/mozilla/fenix/helpers/FenixRobolectricTestRunner.kt
@@ -4,7 +4,6 @@
 
 package org.mozilla.fenix.helpers
 
-import org.mozilla.fenix.TestApplication
 import org.robolectric.RobolectricTestRunner
 import org.robolectric.annotation.Config
 
--- a/app/src/test/java/org/mozilla/fenix/helpers/FenixRobolectricTestRunner.kt
+++ b/app/src/test/java/org/mozilla/fenix/helpers/FenixRobolectricTestRunner.kt
@@ -25,7 +25,7 @@ class FenixRobolectricTestRunner(testClass: Class<*>) : RobolectricTestRunner(te
 
     override fun buildGlobalConfig(): Config {
         return Config.Builder()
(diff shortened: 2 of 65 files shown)


app/src/test/java/org/mozilla/fenix/helpers/FenixRobolectricTestRunner.kt
```diff
@@ -0,0 +1,31 @@
...
+ * @RunWith(FenixRobolectricTestRunner::class)
+ * class ExampleUnitTest {
+ * ```
+ *
+ * IMPORTANT NOTES:
+ * - This should ALWAYS be used instead of RobolectricTestRunner and AndroidJUnit4 (note: the latter
+ * just delegates to the former)
+ * - You should only use Robolectric when necessary because it non-trivially increases test duration.
+ */
+class FenixRobolectricTestRunner(testClass: Class<*>) : RobolectricTestRunner(testClass) {
```
kglazko, line 24: I find the name of this to be a little bit confusing because this test runner replace both the @RunWith(RobolectricTestRunner::class) and @RunWith(AndroidJUnit4::class). Someone may not realize that AndroidJUnit4 is also replaced by this when writing a future test and may opt to use AndroidJUnit4 instead of your test runner.

app/src/test/java/org/mozilla/fenix/ext/LogTest.kt
```diff
@@ -10,13 +10,6 @@ import io.mockk.mockkStatic
 import io.mockk.verify
 import org.junit.Before
 import org.junit.Test
-import org.junit.runner.RunWith
-import org.mozilla.fenix.TestApplication
-import org.robolectric.RobolectricTestRunner
-import org.robolectric.annotation.Config
-
-@RunWith(RobolectricTestRunner::class)
-@Config(application = TestApplication::class)
```
kglazko, line 19: I like that you did some cleanup in this PR and removed the TestRunner where it is unnecessary as well 👍

app/src/test/java/org/mozilla/fenix/helpers/FenixRobolectricTestRunner.kt
mcomella: > Someone may not realize that AndroidJUnit4 is also replaced by this when writing a future test and may opt to use AndroidJUnit4 instead of your test runner.

My hope is that the linter would catch that. Unfortunately, I don't think I can name it in a way that captures the intent of both `AndroidJUnit4` and `FenixRobolectricTestRunner`. :\

In any case, as we discussed, I can add a comment to explain why `AndroidJUnit4` delegates to `FenixRobolectricTestRunner`

kglazko: That would be great!

kglazko:
I would recommend testing in Android Studio and adding @RunWith(RobolectricTestRunner::class) to random tests that don't have it like LogTest and making sure everything still works. On mine, doing this gave me some test failures so it may be good to double check.

mcomella:
> I would recommend testing in Android Studio and adding @RunWith(RobolectricTestRunner::class) to random tests that don't have it like LogTest and making sure everything still works. On mine, doing this gave me some test failures so it may be good to double check.

Thanks for checking this! Android Studio's compilation command called `compileGeckoBetaDebugUnitTestSources` directly, rather than going through `assemble` so I modified the code to hook into the `compile` command rather than assemble which works on both the CL and AS.

app/src/test/java/org/mozilla/fenix/helpers/FenixRobolectricTestRunner.kt
mcomella: New class comment:
```kotlin
/**
 * A test runner that starts Robolectric with our custom configuration for use in unit tests. This
 * should ALWAYS be used instead of RobolectricTestRunner and AndroidJUnit4. You should only use
 * Robolectric when necessary because it non-trivially increases test duration.
 *
 * usage:
 * ```
 * @RunWith(FenixRobolectricTestRunner::class)
 * class ExampleUnitTest {
 * ```
 *
 * There were three common test runners before this patch:
 * 1. The default (@RunWith not specified) = JUnit4
 * 2. @RunWith(RobolectricTestRunner::class) = JUnit4 with support for the Android framework via Robolectric
 * 3. @RunWith(AndroidJUnit4::class) = JUnit4 with support for the Android framework. This currently
 * delegates to Robolectric but is presumably generically named so that it can support different
 * implementations in the future. The name creates confusion on over the difference between this and
 * JUnit without any Android support (1).
 *
 * We chose the name RobolectricTestRunner because we want folks to know they're starting Robolectric
 * because it increases test runtime. Furthermore, the naming of 3) is unclear so we didn't want to
 * use that name.
 */
```

mcomella:
Addressed the comments and rebasing on master to ensure any newly landed changes won't break the build.

mcomella:
UI test failure on `verifyAboutFirefoxPreview` but I didn't touch the UI tests.

But from Aaron on Matrix:
> A patch last night landed to remove re-runs on intermittent  which would decrease the time of the ui-test-x86 job (up to 2 re-runs per failure), but ultimately I think we should restore it as its revealing more intermittents

So I assume this is an intermittent. This PR is very likely to cause merge conflicts so I'd like to take it sooner rather than later – I'm going to take a risk and force merge.

mcomella:
I also double-checked that I can't reproduce the `SettingsAboutTest` failure locally.

Landing... 💥

PR #9622, opened 2020-04-02
Bug 1625971 - Add google-search-restaurants to pageload tests in browsertime.

gmierz:
This patch adds the `google-search-restaurants` to the pageload tests being run on browsertime. A temporary workaround for the activity issue is also added. This will be removed before landing this change.


gmierz:
Actually I'm going to leave the activity change because we're using this in m-c.

gmierz:
Thanks @JohanLorenzo! Can you merge this when you get a chance?

PR #9629, opened 2020-04-02
Bug 1627027 - Use fennec-nightly Fenix build on Raptor/Browsertime tests.

gmierz:
This patch changes the build that we run performance tests on from the `performance-test` variant to the `fennec-nightly` variant.


gmierz:
@JohanLorenzo I have to work through some issues with this - I'll let you know when this is ready for review.

Issue #9626, opened 2020-04-02
Optimize CollectionCreationView layout


emalysz:
When we try adding a tab to a collection, we initialize the CollectionCreationView, which takes 83% of CollectionCreationFragment.onCreateView. 

I notice that we clone three times during our `init` call. This includes inflating the `component_collection_creation_name_collection,` which can take ~8-10ms on a Pixel2 (GeckoBetaFennecBeta variant). I suspect this may be even larger for G5. 

However, if the user has any collections, we do not immediately need this view. Let's see if we can clone and inflate the first time this view is actually needed, not just creating it everytime in `init`.

emalysz:
The three times this is used is if the user is creating a collection, renaming a collection, or has no collections and attempts to "Save to collection." We can not navigate between these use cases, so I don't see a benefit to storing this as a variable. 

We should instead do this when it is needed, not every time we initialize this view.

mcomella:
We haven't see any movement on this in a while so I'm unassigning and sending it back to triage. Please feel free to change it back if this is in error.

mcomella:
Triage: we think this may be related to startup performance so we prioritized with other startup issues.

mcomella:
Thank you @mcarare !

PR #9634, opened 2020-04-02
For #9626: improve Save To Collections Performance

emalysz:
We should avoid cloning the naming collections layout unless we need it to improve performance.

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/collections/CollectionCreationView.kt
+++ b/app/src/main/java/org/mozilla/fenix/collections/CollectionCreationView.kt
@@ -73,10 +73,6 @@ class CollectionCreationView(
             view.context,
             R.layout.component_collection_creation_select_collection
         )
-        nameCollectionConstraints.clone(
-            view.context,
-            R.layout.component_collection_creation_name_collection
-        )
 
         view.bottom_bar_icon_button.apply {
             increaseTapArea(increaseButtonByDps)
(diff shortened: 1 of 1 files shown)


emalysz:
This doesn't change any functionality of the code. It just makes sure we are only cloning when it's actually needed, instead of always doing it within `init`. This should save 8-10ms on "Save To Collections" when the user has >1 collection.

app/src/main/java/org/mozilla/fenix/collections/CollectionCreationView.kt
```diff
@@ -73,10 +73,6 @@ class CollectionCreationView(
             view.context,
             R.layout.component_collection_creation_select_collection
         )
-        nameCollectionConstraints.clone(
-            view.context,
-            R.layout.component_collection_creation_name_collection
```
mcomella, line 78: I notice this file contains two ConstraintLayouts and is being added to a ConstraintLayout: I wonder if the performance problems would go away with less complexity if we simplified this layout and the other nested layouts added rather than deferring initialization. That being said, to do that, we'd have to change the top-level layout to use `<include>` and each included layout to have a top-level `<merge>`.

app/src/main/java/org/mozilla/fenix/collections/CollectionCreationView.kt
```diff
@@ -243,6 +239,10 @@ class CollectionCreationView(
         }
 
         view.name_collection_edittext.showKeyboard()
+        nameCollectionConstraints.clone(
```
mcomella, line 242: Not your fault and nothing you need to do but there are a few things I don't like about this file:
- It nests layouts dynamically using `.clone` and `ConstraintSet.applyTo`: the conventional way to do this is using `<include>` tags.
- Each layout it includes has an additional root `ConstraintLayout`, afaict by necessity of ^ method; this slows down inflation
- There is duplication across methods in the code (e.g. `updateForNameCollection` and `updateForRenameCollection`)

mcomella:
We want to look at the alternative methods so closing in favor of trying those.

PR #9635, opened 2020-04-02
No issue: remove num-flaky-test-attempts from all Flank templates

AaronMT:
With re-runs enabled Flank is adding test executions. This experimental flag might be also masking intermittent tests. Let's see what happens with the flag removed.

### DIFF
--- a/automation/taskcluster/androidTest/flank-x86.yml
+++ b/automation/taskcluster/androidTest/flank-x86.yml
@@ -21,10 +21,6 @@ gcloud:
   # see: https://github.com/TestArmada/flank/issues/341
   #results-history-name: tmp_parallel 
 
-  # The number of times a TestExecution should be re-attempted if one or more\nof its test cases fail for any reason.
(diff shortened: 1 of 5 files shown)


automation/taskcluster/androidTest/flank-x86.yml
```diff
@@ -23,7 +23,7 @@ gcloud:
 
   # The number of times a TestExecution should be re-attempted if one or more\nof its test cases fail for any reason.
   # The maximum number of reruns allowed is 10. Default is 0, which implies no reruns.
-  num-flaky-test-attempts: 2
+  num-flaky-test-attempts: 0
```
rpappalax, line 26: Let's just remove the parameter altogether.  This is an optional param that would be 0 anyway if left off.  I'll approve for now since just simple nit.  Let's get this landed as soon as green 👍

AaronMT:
I'm thinking we should revert this now as more intermittent are showing up as a result.

Issue #2768, opened 2019-05-23
FNX3-16568 ⁃ Option to disable screenshots in private mode


vesta0:
## User Story
* As a user, I want the option to disable taking screenshots in private browsing mode, so I can further protect my privacy.

### Acceptance Criteria
* I can disable taking screenshots in private mode (off by default)
* When disabled in private mode, I am not able to take a screenshot and instead I see a system notification indicating that taking screenshots isn't allowed in private mode.
* I don't see the contents of my private tab in Android multitasking view

topotropic:
We should use a system notification, similar to what Focus does.

vesta0:
Sounds good @topotropic

buttercookie42:
Should this be configurable - not least for bug reports and suchlike?

joebloggs7657:
Could this not be an option? Being able to take screenshots in private mode doesnt exactly compromise the premise of private browsing. I will say from experience in Bromite that not being able to screenshot is more of a hassle than anything related to private browsing.

yoasif:
This should be an option - perhaps even default behavior, but an option nonetheless, imo.

ekager:
If we do this, I think we should make it off by default. From Focus experience, not all users understand or want this feature, and the ones that want it are probably more likely to understand how to seek it out.

lime124:
removing the ux label - this is on our radar already as it's part of Q3 plans.

Djfe:
this is the same bug as
#2248
#3422
#3707

I think 3 of the 4 bugs should be closed as duplicates :)

NotWoods:
I've started working on the blocking functionality in Android Components (mozilla-mobile/android-components#4551).

NotWoods:
The feature has been added and should be incorporated into `BaseBrowserFragment`, in a similar manner to our list of features inside that fragment.

KaairaGupta:
@NotWoods is the feature of disabling screenshots on private mode up? Or is there any work still left to do? I asked because this issue was still showing under the good first bugs. If there is any work, i would love to contribute.

NotWoods:
You can integrate SecureWindowFeature from the feature-privatemode component in our Android Components library. You'll need to edit BaseBrowserFragment and possibly the build.gradle & Dependencies.kt files.

KaairaGupta:
@NotWoods i can't seem to find this library in the code. Please help me with it.

KaairaGupta:
[...]

PR #9659, opened 2020-04-03
For #2768 - Prevent screenshots in private mode

BMEduard:
Added a new option in Private browsing menu to allow or prevent screenshots from being taken while in private mode by adding or removing the FLAG_SECURE flag from the home activity's window.

 This method is called whenever the activity is initialized to account for the browsing mode being changed and whenever the setting from the Private browsing menu is changed.

 The setting is by default set to true (screenshots are allowed to be taken). The system notification 
 that is displayed while screenshots are blocked is : "Can't take screenshot due to security policy"

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
+++ b/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
@@ -48,6 +48,7 @@ import org.mozilla.fenix.components.metrics.BreadcrumbsRecorder
 import org.mozilla.fenix.components.metrics.Event
 import org.mozilla.fenix.exceptions.ExceptionsFragmentDirections
 import org.mozilla.fenix.ext.alreadyOnDestination
+import org.mozilla.fenix.ext.checkAndUpdateScreenshotPermission
 import org.mozilla.fenix.ext.components
 import org.mozilla.fenix.ext.nav
 import org.mozilla.fenix.ext.settings
@@ -116,6 +117,7 @@ open class HomeActivity : LocaleAwareAppCompatActivity() {
         components.publicSuffixList.prefetch()
 
         setupThemeAndBrowsingMode(getModeFromIntentOrLastKnown(intent))
+        checkAndUpdateScreenshotPermission(settings())
         setContentView(R.layout.activity_home)
 
         // Must be after we set the content view
--- a/app/src/main/java/org/mozilla/fenix/settings/PrivateBrowsingFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/PrivateBrowsingFragment.kt
@@ -11,8 +11,10 @@ import androidx.preference.SwitchPreference
 import org.mozilla.fenix.R
 import org.mozilla.fenix.components.PrivateShortcutCreateManager
 import org.mozilla.fenix.components.metrics.Event
+import org.mozilla.fenix.ext.checkAndUpdateScreenshotPermission
 import org.mozilla.fenix.ext.getPreferenceKey
 import org.mozilla.fenix.ext.metrics
+import org.mozilla.fenix.ext.settings
 import org.mozilla.fenix.ext.showToolbar
 
 /**
@@ -40,6 +42,16 @@ class PrivateBrowsingFragment : PreferenceFragmentCompat() {
 
         findPreference<SwitchPreference>(getPreferenceKey(R.string.pref_key_open_links_in_a_private_tab))?.apply {
             onPreferenceChangeListener = SharedPreferenceUpdater()
+        }
+
+        findPreference<SwitchPreference>(getPreferenceKey
+            (R.string.pref_key_allow_screenshots_in_private_mode))?.apply {
+            onPreferenceChangeListener = object : SharedPreferenceUpdater() {
+                override fun onPreferenceChange(preference: Preference, newValue: Any?): Boolean {
+                    return super.onPreferenceChange(preference, newValue).also {
+                        requireActivity().checkAndUpdateScreenshotPermission(requireActivity().settings()) }
+                }
+            }
         }
     }
 }
--- a/app/src/main/java/org/mozilla/fenix/ext/Activity.kt
+++ b/app/src/main/java/org/mozilla/fenix/ext/Activity.kt
@@ -7,6 +7,7 @@ package org.mozilla.fenix.ext
 import android.app.Activity
 import android.view.View
 import android.view.WindowManager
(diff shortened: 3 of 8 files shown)


app/src/main/java/org/mozilla/fenix/settings/PrivateBrowsingFragment.kt
```diff
@@ -41,5 +42,17 @@ class PrivateBrowsingFragment : PreferenceFragmentCompat() {
         findPreference<SwitchPreference>(getPreferenceKey(R.string.pref_key_open_links_in_a_private_tab))?.apply {
             onPreferenceChangeListener = SharedPreferenceUpdater()
         }
+
+        findPreference<Preference>(getPreferenceKey(R.string.pref_key_allow_screenshots_in_private_mode))?.apply {
+            setOnPreferenceClickListener {
+                (requireActivity() as HomeActivity).checkAndUpdateScreenshotBlockFlag()
```
Mugurell, line 48: Can this cast be avoided? It adds tight coupling to an implementation we should not know about.
Could we use `SharedViewModel by activityViewModels()`?

Mugurell: Actually I don't think we even need to call something in Activity.
How about using a separate object  which would know to handle enabling/disabling screenshots, one that can be used from both the activity and the fragment?

app/src/main/java/org/mozilla/fenix/settings/PrivateBrowsingFragment.kt
```diff
@@ -41,5 +43,17 @@ class PrivateBrowsingFragment : PreferenceFragmentCompat() {
...
         }
+
+        findPreference<Preference>(getPreferenceKey(R.string.pref_key_allow_screenshots_in_private_mode))?.apply {
+            setOnPreferenceClickListener {
+                ScreenshotsGuardian().invoke(requireActivity() as AppCompatActivity)
+                true
+            }
+        }
+
+        findPreference<SwitchPreference>(getPreferenceKey
```
Mugurell, line 54: Could we merge the two listeners together?
Maybe for something like:
```
onPreferenceChangeListener = object : SharedPreferenceUpdater() {
    override fun onPreferenceChange(preference: Preference, newValue: Any?): Boolean {
        ScreenshotsGuardian().invoke(requireActivity() as AppCompatActivity)
        return super.onPreferenceChange(preference, newValue)
    }
}
```

app/src/main/java/org/mozilla/fenix/utils/ScreenshotsGuardian.kt
```diff
@@ -0,0 +1,21 @@
...
+
+package org.mozilla.fenix.utils
+
+import android.view.WindowManager
+import androidx.appcompat.app.AppCompatActivity
+import org.mozilla.fenix.ext.settings
+
+class ScreenshotsGuardian {
+    operator fun invoke(activity: AppCompatActivity) {
+        if (!activity.baseContext.settings().allowScreenshotsInPrivateMode &&
```
Mugurell, line 13: I don't think we need to explicitly use `baseContext`.

app/src/main/java/org/mozilla/fenix/utils/ScreenshotsGuardian.kt
```diff
@@ -0,0 +1,21 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.utils
+
+import android.view.WindowManager
+import androidx.appcompat.app.AppCompatActivity
+import org.mozilla.fenix.ext.settings
+
+class ScreenshotsGuardian {
+    operator fun invoke(activity: AppCompatActivity) {
```
Mugurell, line 12: Great choice to extract this responsibility in a separate object that can be freely used from anywhere in the app!
I'm thinking just that `ScreenshotsGuardian.invoke()` is not really telling about what will happen.
Maybe `ScreenshotsGuardian.updateScreenshotsPermission()` would be more indicative of the end result?

An alternative would be to write this method as an extension function to Activity, similar to `Activity.enterToImmersiveMode()`

app/src/main/java/org/mozilla/fenix/settings/PrivateBrowsingFragment.kt
```diff
@@ -41,5 +43,17 @@ class PrivateBrowsingFragment : PreferenceFragmentCompat() {
         findPreference<SwitchPreference>(getPreferenceKey(R.string.pref_key_open_links_in_a_private_tab))?.apply {
             onPreferenceChangeListener = SharedPreferenceUpdater()
         }
+
+        findPreference<Preference>(getPreferenceKey(R.string.pref_key_allow_screenshots_in_private_mode))?.apply {
+            setOnPreferenceClickListener {
+                ScreenshotsGuardian().invoke(requireActivity() as AppCompatActivity)
```
Mugurell, line 49: We'd not need this cast if ScreenshotsGuardian's methods would take `FragmentActivity` as a parameter.

app/src/main/java/org/mozilla/fenix/utils/ScreenshotsGuardian.kt
```diff
@@ -0,0 +1,21 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.utils
+
+import android.view.WindowManager
+import androidx.appcompat.app.AppCompatActivity
+import org.mozilla.fenix.ext.settings
+
+class ScreenshotsGuardian {
```
Mugurell, line 11: Given it's importance and responsibility maybe some comments to document the behavior and some small tests would help.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -377,6 +393,14 @@ open class HomeActivity : LocaleAwareAppCompatActivity() {
         browsingModeManager.mode = sessionMode
     }
 
+    fun checkAndUpdateScreenshotBlockFlag() {
+        if (!this.settings().allowScreenshotInAPrivateTab && browsingModeManager.mode.isPrivate) {
```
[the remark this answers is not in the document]
BMEduard: Thanks!

PR #9675, opened 2020-04-03
No issue: Restore num-flaky-test-attempts in flank-x86.yml

AaronMT:
Removing the Flank template option is revealing lots of intermittent failures. I think we should revert our decision to remove this in #9635 and limit the re-runs to 2.

### DIFF
--- a/automation/taskcluster/androidTest/flank-x86.yml
+++ b/automation/taskcluster/androidTest/flank-x86.yml
@@ -21,6 +21,10 @@ gcloud:
   # see: https://github.com/TestArmada/flank/issues/341
   #results-history-name: tmp_parallel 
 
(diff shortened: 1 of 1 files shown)


automation/taskcluster/androidTest/flank-x86.yml
```diff
@@ -21,6 +21,10 @@ gcloud:
   # see: https://github.com/TestArmada/flank/issues/341
   #results-history-name: tmp_parallel 
 
+  # The number of times a test execution should be re-attempted if one or more failures occur.
+  # The maximum number of reruns allowed is 10. Default is 0, which implies no reruns.
+  num-flaky-test-attempts: 2
```
rpappalax, line 26: let's do just 1 per our conversation.  My understanding is that 0 will run once and 1 will repeat once

PR #9677, opened 2020-04-03
Update not yet unavailable add-on string.

psymoon:


pocmo:
Can you also uplift this patch to the `fennec/beta` branch? :)

PR #9680, opened 2020-04-03
Update not yet unavailable add-on string for fennec/beta branch

psymoon:


pocmo:
Actually, let's wait with merging this. I think I need to uplift the string first since that may not be in the AC release we are using on this branch. I can look at this on Monday.

psymoon:
@pocmo we should also wait for the translations to be in place, right?

pocmo:
For the previous string change we decided to ship the new one even if it is not translated to all languages yet.

ekager:
@psymoon can we do a rebase and land assuming we still need this?

pocmo:
This is no longer needed! :)

Issue #9685, opened 2020-04-03
[Bug] Possible intermittent UI test failure - SettingsAboutTest.verifyAboutFirefoxPreview


mcomella:
In my PR https://github.com/mozilla-mobile/fenix/pull/9607, I received a UI test failure ([firebase](https://console.firebase.google.com/project/moz-fenix/testlab/histories/bh.66b7091e15d53d45/matrices/4945431146187137688/executions/bs.7d3456f0d9f01a64)) in `SettingsAboutTest.verifyAboutFirefoxPreview`. However, I only touched unit test code and build code – no production app or UI test code – so the UI test results should remain unchanged.

aaronmt briefly looked into this and noted:
> Open bug with the intermittent label for now mcomella -- that one is weird, "Test instrumentation process crashed. Check org.mozilla.fenix.ui.SettingsAboutTest#verifyAboutFirefoxPreview.txt for details"

Also from the video tab:
> Video is not available for test cases shorter than 0.5 seconds or where Firebase Test Lab records at a low framerate, such as for virtual devices. 

---

aaronmt continued investigation:

> Oh the log-cat shows it got the crash reporter dialog and quit
I wonder if we can do some kind of graceful shutdown or better crash identification 04-03 11:54:46.113: I/ActivityManager(1663):   Force finishing activity ActivityRecord{fb957a6 u0 org.mozilla.fenix.debug/mozilla.components.lib.crash.prompt.CrashReporterActivity t5}
Are there many crash reports from Firebase AVD devices (how do you even identify them?)

AaronMT:
https://console.firebase.google.com/u/0/project/moz-fenix/testlab/histories/bh.66b7091e15d53d45/matrices/5253904521782985921

Crash happening, is there any way to associate a report here with full details?

> 04-05 20:29:14.754: I/ActivityManager(1662): Start proc 9736:org.mozilla.fenix.debug:mozilla.components.lib.crash.CrashHandler/u0a95 for service org.mozilla.fenix.debug/mozilla.components.lib.crash.handler.CrashHandlerService


[...]

PR #9788, opened 2020-04-07
For #8803: add frameworkStart telemetry

mcomella:
This is intended to measure the "dex" time before we can execute code. This builds upon my learning of how we want to capture telemetry metrics in https://github.com/mozilla-mobile/fenix/pull/9153.

@Dexterp37 Please review this from a Glean perspective.

@boek or @liuche Please review as a data steward.

@MarcLeclair Please review as the primary code reviewer!

### Reviews requests:
- [ ] Data review
- [ ] Review from the Glean team
- [ ] Code review

---

### Pull Request checklist

### After merge

1. click on Show All Checks,


mcomella:

1) What questions will you answer with this data?

- How long does the Android framework block for on various devices before giving us the ability to execute code?

2) Why does Mozilla need to answer these questions?  Are there benefits for users? Do we need this information to address product or business requirements? Some example responses:

- Understand how significant the time before we can execute code is so that we can choose whether or not we want to invest in optimizing the behavior.

3) What alternative methods did you consider to answer these questions? Why were they not sufficient?

- We could measure locally on our reference devices but we wouldn't be able to determine 1) the variance across a wide range of devices or 2) how frequent and impactful outliers to local testing may be.

4) Can current instrumentation answer these questions?

No.

5) List all proposed measurements and indicate the category of data collection for each measurement, using the [Firefox data collection categories](https://wiki.mozilla.org/Firefox/Data_Collection) found on the Mozilla wiki.   

**Note that the data steward reviewing your request will characterize your data collection based on the highest (and most sensitive) category.**

<table>
  <tr>
    <td>Measurement Description</td>
    <td>Data Collection Category</td>
    <td>Tracking Bug #</td>
  </tr>
  <tr>
    <td>framework_start: timespan</td>
    <td>Category 1</td>
    <td>#8803</td>
  </tr>
  <tr>
    <td>framework_start_error: event</td>
    <td>Category 1</td>
    <td>#8803</td>
  </tr>
  <tr>
    <td>clock_ticks_per_second: counter</td>
    <td>Category 1</td>
    <td>#8803</td>
  </tr>
</table>


6) How long will this data be collected?  Choose one of the following:

* I want this data to be collected for 6 months initially (potentially renewable).

7) What populations will you measure?

All.

8) If this data collection is default on, what is the opt-out mechanism for users?

Standard telemetry opt-out.

9) Please provide a general description of how you will analyze this data.

- Look at `frameworkStart` to understand how long it takes for the framework to start across all devices
- Watch `frameworkStartError` to determine if there are any implementation errors.
- Look at `clockTicksPerSecond` to see if this value changes across devices and, if so, potentially segment `frameworkStart` data to group data to the same significant figures

10) Where do you intend to share the results of your analysis?

Internally.

11) Is there a third-party tool (i.e. not Telemetry) that you are proposing to use for this data collection? If so:

No.

Dexterp37:
> 
> 
> This is intended to measure the "dex" time before we can execute code. 

I can tell you that time! (sorry, had to make the joke)

app/metrics.yaml
```diff
@@ -1984,3 +1984,57 @@ addons:
...
+      `processStartTimestamp` is derived from the clock tick time unit, which is expected to be
+      less granular than nanoseconds. Therefore, we convert and round our timestamps to clock ticks
+      before computing the difference and convert back to nanoseconds to report.
+
+      For debugging purposes, `clock_ticks_per_second`, which may vary between devices, is also
+      reported as a metric.
+    bugs:
+      - https://github.com/mozilla-mobile/fenix/issues/8803
+    data_reviews:
+      - https://github.com/mozilla-mobile/fenix/pull/9788#issuecomment-610648980
```
Dexterp37, line 2006: Note that you need to link to the **answers** of the data-review by the data stewards.

app/metrics.yaml
```diff
@@ -1984,3 +1984,57 @@ addons:
...
+    send_in_pings:
+      - startup-timeline
+    type: event
+    description: >
+      An error when attempting to record `framework_start` - the application init timestamp returned
+      a negative value - which is likely indicative of a bug in the implementation.
+    bugs:
+      - https://github.com/mozilla-mobile/fenix/issues/8803
+    data_reviews:
+      - https://github.com/mozilla-mobile/fenix/pull/9788#issuecomment-610648980
```
Dexterp37, line 2021: Same here

app/metrics.yaml
```diff
@@ -1984,3 +1984,57 @@ addons:
...
+    send_in_pings:
+      - startup-timeline
+    type: counter
+    description: >
+      The number of clock tick time units that occur in one second on this particular device. This
+      value is expected to be used in conjunction with the `framework_start` metric.
+    bugs:
+      - https://github.com/mozilla-mobile/fenix/issues/8803
+    data_reviews:
+      - https://github.com/mozilla-mobile/fenix/pull/9788#issuecomment-610648980
```
Dexterp37, line 2036: And here.

app/pings.yaml
```diff
@@ -29,3 +29,19 @@ installation:
...
+    This ping is intended to provide an understanding of startup performance.
+
+    The ping is intended to be captured by performance testing automation to report results
+    there, in addition to user telemetry. We place these metrics into their own ping in order
+    to isolate them and make this process easier.
+  include_client_id: false
+  bugs:
+    - https://github.com/mozilla-mobile/fenix/issues/8803
+  data_reviews:
+    - https://github.com/mozilla-mobile/fenix/pull/9788#issuecomment-610648980
```
Dexterp37, line 44: And here :)

app/src/main/java/org/mozilla/fenix/perf/StartupFrameworkStartMeasurement.kt
```diff
@@ -0,0 +1,86 @@
...
+     * We defer these metrics, rather than setting them as soon as the values are available,
+     * because they are slow to fetch and we don't want to impact startup.
+     */
+    fun setExpensiveMetric() {
+        // The application is only init once per process lifetime so we only set this value once.
+        if (isMetricSet) return
+        isMetricSet = true
+
+        if (applicationInitNanos < 0) {
+            telemetry.frameworkStartError.record()
```
Dexterp37, line 54: Do you really need an event for this? You only expect this to happen _once_ per application run, right? I'd go with a `boolean` metric type instead, and just flip it to true here. See [our docs](https://mozilla.github.io/glean/book/user/adding-new-metrics.html) for a discussion on why `event` is not the right fit in this case.

app/metrics.yaml
mcomella: I'll apply these changes after data review. :)

app/src/main/java/org/mozilla/fenix/perf/StartupFrameworkStartMeasurement.kt
mcomella: I didn't realize that page existed – I just clicked directly into the different metric types when looking between them, which is why overlooked the boolean option.

I take it:

> Importantly, events are the most expensive metric type to record, transmit, store and analyze, so they should be used sparingly, and only when none of the other metric types are sufficient for answering your question.

Is why I don't want `event`. `boolean` wfm.

mcomella: @Dexterp37 I wasn't sure which you wanted:
- frameworkStartError to be false on all runs and switched to true when there's an error
- unset on all runs and set true when there's an error

I flagged you for re-review to double-check my approach.

app/metrics.yaml
```diff
@@ -1984,3 +1984,57 @@ addons:
...
+      For debugging purposes, `clock_ticks_per_second`, which may vary between devices, is also
+      reported as a metric.
+    bugs:
+      - https://github.com/mozilla-mobile/fenix/issues/8803
+    data_reviews:
+      - https://github.com/mozilla-mobile/fenix/pull/9788#issuecomment-610648980
+    notification_emails:
+      - perf-android-fe@mozilla.com
+      - mcomella@mozilla.com
+    expires: "2020-10-01"
```
pocmo, line 2010: I wonder if we can be more aggressive here - assuming this is primarily to better understand how fast code loading is and whether there is an issue we need to address, as opposed to a long-term measurement.

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -49,6 +50,10 @@ import org.mozilla.fenix.utils.Settings
 @SuppressLint("Registered")
 @Suppress("TooManyFunctions", "LargeClass")
 open class FenixApplication : LocaleAwareApplication() {
+    init {
+        recordOnInit() // DO NOT MOVE ANYTHING ABOVE HERE: the timing of this measurement is critical.
+    }
```
pocmo, line 55: It's somewhat scary to run all of that code without any crash reporting. There's probably not much we can do. But it may be another reason to shorten the lifetime of this measurement.

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -365,4 +370,10 @@ open class FenixApplication : LocaleAwareApplication() {
             Logger.error("Failed to initialize web extension support", e)
         }
     }
+
+    protected fun recordOnInit() {
+        // This gets called by more than one process. Ideally we'd only run this in the main process
+        // but the code to check which process we're in crashes because the Context isn't valid yet.
+        StartupTimeline.onApplicationInit()
```
pocmo, line 377: nit: The comment in `init` is warning that nothing should happen before this. However `recordOnInit()` looks fairly harmless and I wonder if it could happen that someone adds something in front of it inside this method. Is there any reason we can't inline the call to `StartupTimeline`?

app/src/migration/java/org/mozilla/fenix/MigratingFenixApplication.kt
```diff
@@ -13,8 +13,9 @@ import org.mozilla.fenix.session.PerformanceActivityLifecycleCallbacks
  * An application class which knows how to migrate Fennec data.
  */
 class MigratingFenixApplication : FenixApplication() {
-
     init {
+        recordOnInit() // DO NOT MOVE ANYTHING ABOVE HERE: the timing of this measurement is critical.
```
pocmo, line 17: Since `MigratingFenixApplication` extends `FenixApplication` both initializer blocks should get invoked resulting in `recordOnInit()` getting called twice.

app/src/main/java/org/mozilla/fenix/perf/StartupTimeline.kt
```diff
@@ -45,3 +67,29 @@ object StartupTimeline {
...
+ */
+internal class StartupHomeActivityLifecycleObserver(
+    private val frameworkStartMeasurement: StartupFrameworkStartMeasurement,
+    private val startupTimeline: PingType<NoReasonCodes> = Pings.startupTimeline
+) : LifecycleObserver {
+
+    @OnLifecycleEvent(Lifecycle.Event.ON_STOP)
+    fun onStop() {
+        GlobalScope.launch { // use background thread due to expensive metrics.
+            // Ensure any last metrics are set before submission.
```
MarcLeclair, line 82: Is there a reason why we use GlobalScope? is this the onStop for the Activity, which is only called when the app gets killed?

app/src/main/java/org/mozilla/fenix/perf/StartupFrameworkStartMeasurement.kt
```diff
@@ -0,0 +1,86 @@
...
+
+    fun onApplicationInit() {
+        // This gets called from multiple processes: don't do anything expensive. See call site for details.
+        //
+        // In the main process, there are multiple Application impl so we ensure it's only set by
+        // the first one.
+        if (!isApplicationInitCalled) {
+            isApplicationInitCalled = true
+            applicationInitNanos = getElapsedRealtimeNanos()
+        }
```
MarcLeclair, line 39: Question, the `getElapsedRealtimeNanos` is the time since boot, so that means x from when the power was turned on, right?

app/src/main/java/org/mozilla/fenix/perf/StartupTimeline.kt
mcomella: > Is there a reason why we use GlobalScope?

My understanding is that `GlobalScope` has no built in cancellation. Assuming you don't need things to return to the UI thread and don't mind running while the app is backgrounded (e.g. you might want to to conserve battery/CPU), it's fine to use. I think this represents our case. Ideally, I think we'd want to avoid the latter case but, to keep this simple and because it's unlikely to continue running long anyway, I think it's fine.

> is this the onStop for the Activity,

Yes, this observer gets registered on HomeActivity.

> which is only called when the app gets killed?

It should get called in that case and when the app is backgrounded by a user and not killed.

MarcLeclair: So yes, from what I read `GlobalScope` is lost in the void. Is there any event where we would want to cancel this job?

app/src/main/java/org/mozilla/fenix/perf/StartupFrameworkStartMeasurement.kt
mcomella: > the getElapsedRealtimeNanos is the time since boot, so that means x from when the power was turned on, right?

Yes.

> If that's the case, what exactly are we timing here?

We capture this as a timestamp to later compare it with the "process start time". When combining these (`appInitTime - processStartTime`), we determine the duration that the framework takes before letting us execute code. Notably, the major thing in there is that dex file time that appears quite large in profiles but may not be as large in non-debuggable apps. This probe is primarily to determine if this is worth worrying about.

Since `processStartTime` is determined from clock ticks rather than `getElapsedRealtimeNanos`, I made sure they're comparable – see the comments for details.

app/src/migration/java/org/mozilla/fenix/MigratingFenixApplication.kt
mcomella: That's by design: `StartupFrameworkStartMeasurement` handles multiple callers. I'll add a comment to `StartupTimeline.onApplicationInit` to make that assumption clearer in that wrapper method though.

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
mcomella: Good point. I wanted to:
1. avoid duplicating the comment, "this gets called by more than one process..."
1. (I care about this less) make it easy for additional `Application` classes to call this method, if necessary

@pocmo Is `MigratingFenixApplication` going away soon (merged into `FenixApplication`)? If so, I'll inline with duplicated comment. If not, I'll add a comment to this method not to add anything above that line.

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
mcomella: That's a very good point I did not consider. Luckily this is a few assignments and an if statement and I don't expect the code to change as much. I agree with shortening the lifetime though.

app/metrics.yaml
mcomella: That's accurate and I agree we should minimize this.

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
mcomella: This is covered by Google Play crash services though, right? Perhaps I should monitor that when this lands.

app/src/main/java/org/mozilla/fenix/perf/StartupTimeline.kt
mcomella: Not that I'm aware of – in fact, to keep things simple, my goal is to not to have to deal with the possibility. :)

mcomella:
- Unit tests failed with failure I can't reproduce but is in the test that I added
- UI tests failed with failure in `SettingsAboutTest` which seems like the intermittent #9685. That failure apparently can happen if the app crashes

This is the failure:
```
[task 2020-04-16T00:50:17.726Z] SUITE: org.mozilla.fenix.perf.StartupHomeActivityLifecycleObserverTest
[task 2020-04-16T00:50:17.726Z]   TEST: WHEN onStop is called THEN the metrics are set and the ping is submitted
[task 2020-04-16T00:50:17.726Z]   FAILURE
[task 2020-04-16T00:50:17.726Z] 
[task 2020-04-16T00:50:17.726Z] java.lang.AssertionError: Verification failed: number of calls happened not matching exact number of verification sequence
[task 2020-04-16T00:50:17.726Z] 
[task 2020-04-16T00:50:17.726Z] Matchers: 
[task 2020-04-16T00:50:17.726Z] StartupFrameworkStartMeasurement(frameworkStartMeasurement#1678).setExpensiveMetric())
[task 2020-04-16T00:50:17.726Z] PingType(startupTimeline#1679).submit(null()))
[task 2020-04-16T00:50:17.726Z] 
[task 2020-04-16T00:50:17.726Z] Calls:
[task 2020-04-16T00:50:17.726Z] 1) StartupFrameworkStartMeasurement(frameworkStartMeasurement#1678).setExpensiveMetric()
```

I'll rebase and see if it goes away.

PR #9829, opened 2020-04-09
Cleans up nav_graph.xml

boek:
For phase 1 we wanted to simplify the nav_graph as much as possible. This will allow us to make some bigger changes in the navigation architecture without touching as much code in the future.

I broke these into smaller commits to make it easier to review. I will probably squash them before merging :)

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/crashes/CrashReporterController.kt
+++ b/app/src/main/java/org/mozilla/fenix/crashes/CrashReporterController.kt
@@ -11,10 +11,8 @@ import kotlinx.coroutines.Job
 import kotlinx.coroutines.launch
 import mozilla.components.browser.session.Session
 import mozilla.components.lib.crash.Crash
-import org.mozilla.fenix.R
 import org.mozilla.fenix.components.Components
 import org.mozilla.fenix.components.metrics.Event
-import org.mozilla.fenix.ext.nav
 import org.mozilla.fenix.utils.Settings
 
 class CrashReporterController(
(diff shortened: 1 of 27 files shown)


app/src/main/java/org/mozilla/fenix/settings/deletebrowsingdata/DeleteBrowsingDataFragment.kt
```diff
@@ -147,7 +147,7 @@ class DeleteBrowsingDataFragment : Fragment(R.layout.fragment_delete_browsing_da
         if (popAfter) viewLifecycleOwner.lifecycleScope.launch(
             Dispatchers.Main
         ) {
-            returnToDeletionOrigin()
+            findNavController().popBackStack()
```
ekager, line 150: A little worried about this interaction, I think maybe the `returnToDeletionOrigin` also pops off the browser fragment in case the user closes all tabs? What do you think?

app/src/main/java/org/mozilla/fenix/crashes/CrashReporterController.kt
```diff
@@ -55,10 +53,7 @@ class CrashReporterController(
 
         components.useCases.tabsUseCases.removeTab(session)
         components.useCases.sessionUseCases.crashRecovery.invoke()
-        navController.nav(
```
ekager, line 58: It looks like this extension fun was created to "safely" navigate. Are we sure we don't need this here anymore?

app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkFragment.kt
```diff
@@ -218,7 +217,7 @@ class BookmarkFragment : LibraryPageFragment<BookmarkNode>(), UserInteractionHan
 
     private fun navigate(directions: NavDirections) {
         invokePendingDeletion()
-        nav(R.id.bookmarkFragment, directions)
+        findNavController().navigate(directions)
```
ekager, line 220: same nit here about the nav function?

Issue #9208, opened 2020-03-17
In-product prompt


vesta0:
## User Story
- As a user, I want to know that my app is changing, so I can continue having a smooth user experience. 

### Acceptance Criteria
- When I open the app, I see a persistent prompt directing me to the new app.

### UX
Mocks w/ copy - https://share.goabstract.com/917b9c45-0031-4042-8fed-789a6611e50e?mode=design

betsymi:
Clarification question: Does this mean when user opens Fennec, we direct them to Fenix?

liuche:
Related: #6610 for Nightly

vesta0:
@betsymi @apbitner here is the sumo link for the message: https://support.mozilla.org/1/mobile/%VERSION%/%OS%/%LOCALE%/sync-delist

Once we have added the mocks and strings to this issue, please mark it as `eng:ready` Thank you.

apbitner:
Added in comment 0; mocks have current copy.

sblatz:
QA, You can verify this by changing the `build.gradle` `applicationSuffix` https://github.com/mozilla-mobile/fenix/blob/master/app/build.gradle#L55 to the variations `fennec_aurora_`, `fenix_nightly`, `fennec`. Please message me if you have questions 😄

abodea:
@sblatz I created 2 builds:
```kotlin
fennec_aurora -> The Firefox Nightly has moved prompt was correctly displayed after onboarding
```
```Kotlin
fenix_nightly -> No prompt displayed after onboarding.
```
Is this the intended behavior? Is the `fennec_aurora` the only variation that needs to be verified here?
Note that the `Get the new Nightly` redirected me correctly to the Play Store, and as I had it installed it opened the newly Nightly.

![Screenshot_20200427-132639](https://user-images.githubusercontent.com/42831109/80374250-12836500-889f-11ea-8a68-38916417aab4.png)

sblatz:
@abodea can you also please check for `fennec` and the fenix nightly one is actually supposed to be `fenix.nightly` sorry, had a typo in my original comment! Those two versions should also have prompts :)

Thanks!

abodea:
I retested this issue for more details please check the following:

- `.fenix.nightly` I was able to install the build and the prompt was correctly displayed when I tapped on the `Get Mozilla Firefox Browser` I was redirected to the play store and `Firefox Nightly for Developers (Early access) was displayed.

[...]

PR #9836, opened 2020-04-09
Issue #9208: Adds in-product prompt to homescreen

sblatz:
### Pull Request checklist

### After merge

1. click on Show All Checks,


sblatz:

- How often do users receive a product prompt
- How often do users engage with the product prompt by taking its action
- How often do users close the product prompt

- We need to know if users are getting the prompt they should be receiving and whether or not they're engaging with it so we can make the prompt more effective in the future.

* No, this is a new feature and there is no telemetry surrounding it.

* All data is Category 2.

* Until 09/01/2020

* Glean / Amplitude

app/src/main/java/org/mozilla/fenix/home/tips/TipManager.kt
```diff
@@ -0,0 +1,49 @@
...
+import android.content.Context
+import androidx.annotation.DrawableRes
+import androidx.annotation.StringRes
+import org.mozilla.fenix.FeatureFlags
+
+enum class Type {
+    BUTTON
+}
+
+open class Tip(
```
boek, line 16: Nit: Since we're adding the abstraction of a ButtonType. Could we move some of the data that is associated with that type to the type itself?

```
sealed class Type {
    data class Button(val title: String, action: () -> Unit)
}
```

app/src/main/java/org/mozilla/fenix/home/tips/TipManager.kt
```diff
@@ -0,0 +1,49 @@
...
+    @StringRes val identifier: Int,
+    @DrawableRes val icon: Int = -1,
+    @StringRes val title: Int,
+    @StringRes val description: Int,
```
boek, line 20: Could we have "learn more" be a link inside of the description versus having a separate section? This would allow us to have it inline like the mocks

boek, line 20: what are the thoughts behind making this resource values versus the actual string / drawables?

sblatz: The reason I didn't do this is for accessibility purposes. I remember a while back that we had to move our "learn more" links into separate views so they would be separately selectable for screen readers. See the private browsing "myths" and the private browsing "allow search suggestions" hint.

sblatz: Really happy with doing either here, do you think doing actual strings/drawables is cleaner? :)

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -137,6 +137,12 @@ class Settings private constructor(
         default = true
     )
 
+    // If any of the prefs have been modified, quit displaying the fenix moved tip
+    fun shouldDisplayFenixMovingTip(): Boolean =
+        preferences.getBoolean(appContext.getString(R.string.pref_key_migrating_from_fenix_nightly_tip), true) &&
```
boek, line 142: should this be || instead of &&?

sblatz: I don't think so. Each of these default to "true" and each user will get served *one* of these fenix moving tips. If *any* of them become "false" we no longer want to display the tip.

sblatz: If it was an OR here, then they'd always get the tip and have no way to set the other pref keys to false :)

Issue #9857, opened 2020-04-10
Create UI Style for Text Buttons


sblatz:
### Acceptance Criteria (how do I know when I’m done?)
* There is a style engineers can import for buttons
* There is a living (and public) style guide that both engineers & UX can point to when discussing designs
* Every text button in the app will adhere to a common pattern so we have consistency 

Stretch goal: This style lives in a library we can import separately from Fenix (like in AC)

### General Process
- [x] UX identifies components we’re using
- [x] Engineering finds usages
- [x] UX & Engineering collaborate on “rules”
- [x] Engineer implements the style
- [x] UX & Engineering meet to do “final approval” of the component
	1. Walk through all affected screens
	2. Chat about remaining questions
- [ ] UX & Engineering place component in the library

sblatz:
This has been implemented in Android Components with the ui-widgets project. It also has a temporary style guide representation from UX in this [slide deck](https://docs.google.com/presentation/d/1b7mAuUwH5e-rLOKd8dJg2lD08KSl-3hiVEesm2y1t1k/edit#slide=id.g8147bd26ac_0_5). The last bit of work left for this one is to put it in a long-standing open source style guide document!

PR #9858, opened 2020-04-10
For #9857: Create button style

sblatz:
**Note: this is the first step for implementing this. Ideally this will live in AC, but I want to get something merged in for now so we don't make the problem worse!**

### Pull Request checklist

### After merge

1. click on Show All Checks,


app/src/main/res/layout/save_tab_group_button.xml
```diff
@@ -2,24 +2,11 @@
...
-    android:background="@drawable/button_background"
-    android:letterSpacing="0"
     android:text="@string/save_to_collection"
-    android:textAllCaps="false"
-    android:textColor="?contrastText"
-    app:backgroundTint="?accent"
-    app:icon="@drawable/ic_tab_collection"
-    app:iconGravity="textStart"
-    app:iconPadding="8dp"
-    app:fontFamily="@font/metropolis_medium"
```
kglazko, line 24: You may want to verify with UX cases like this here in that this button was one of the few and only cases in which semi-bold was too bold and they requested medium instead. We have iterated on this a couple of times though, but may want to check with them to see whether they're happy with it in semi-bold or medium. In your implementation, you inherit the styling for the positive button from the neutral one, there may be a couple more cases like this as you go through headers etc. later where it may require a different font than from the default button.

app/src/main/res/values/styles.xml
```diff
@@ -180,18 +180,32 @@
         <item name="android:windowExitAnimation">@anim/fade_out</item>
     </style>
 
-    <style name="ThemeIndependentMaterialGreyButton" parent="Widget.MaterialComponents.Button.TextButton">
+    <!-- UI button styling -->
+    <style name="NeutralButton" parent="Widget.MaterialComponents.Button.TextButton">
+        <item name="iconTint">@color/button_text_color</item>
```
kglazko, line 185: Are you adding new styling for icons in buttons here?

app/src/main/res/values/styles.xml
```diff
@@ -180,18 +180,32 @@
...
+        <item name="iconGravity">textStart</item>
+        <item name="android:textAlignment">center</item>
+        <item name="android:layout_width">match_parent</item>
+        <item name="android:layout_height">48dp</item>
         <item name="android:textStyle">bold</item>
         <item name="android:textAllCaps">false</item>
         <item name="backgroundTint">@color/grey_button_color</item>
         <item name="android:textColor">@color/button_text_color</item>
         <item name="rippleColor">?secondaryText</item>
+        <item name="android:letterSpacing">0</item>
```
kglazko, line 196: Why did the letter spacing need to be specified here and/or is different from the other styles?

mozilla-lint-rules/src/main/java/org/mozilla/fenix/lintrules/ButtonStyleXmlDetector.kt
```diff
@@ -0,0 +1,59 @@
...
+import com.android.resources.ResourceFolderType
+import com.android.tools.lint.detector.api.Category
+import com.android.tools.lint.detector.api.Implementation
+import com.android.tools.lint.detector.api.Issue
+import com.android.tools.lint.detector.api.ResourceXmlDetector
+import com.android.tools.lint.detector.api.Scope
+import com.android.tools.lint.detector.api.Severity
+import com.android.tools.lint.detector.api.XmlContext
+import org.w3c.dom.Element
+
```
kglazko, line 13: A KDOC here would be nice

app/src/main/res/values/styles.xml
sblatz: The letters looked super spread out when I didn't change this. It's defined as "0" in several places thorughout the app, and I think it should be a consistent thing. I will confirm with UX and file a follow up for addressing this, but I think 0 is fine for now?

app/src/main/res/values/styles.xml
sblatz: My understanding from UX is that the icons should match the text tint :) another thing I will confirm in the style guide before it's formalized.

app/src/main/res/layout/save_tab_group_button.xml
sblatz: I will check with them before I get UX sign off! 😄

mozilla-lint-rules/src/main/java/org/mozilla/fenix/lintrules/ButtonStyleXmlDetector.kt
sblatz: Added 😄

ekager:
Looks like there are some parts of the app that use a `TextView` embedded in a `FrameLayout` with background to get the look of a `MaterialButton`. Not really sure why... But could we try fix these to adopt the same styles as well?
For example, this is in `onboarding_private_browsing` and `onboarding_privacy_notice`. Those are the only two I saw with a quick search but there could be more :/ 
Screenshot_1586905198

PR #9964, opened 2020-04-15
Close #9960: Launch URI in Fenix if failed to launch in custom tab

### LINKED ISSUE #9960: Launch URI in Fenix if failed to launch in custom tab
### Why/User Benefit/User Problem
Currently if a site opened in custom tab launched a new window request with URL that is not http/https, this will fail.

### What/Requirements
When new window request fails in custom tab, Fenix should open it in non custom tab.

### Acceptance Criteria (how do I know when I’m done?)
Failed custom tab new window request is opened in non custom tab.

rocketsroger:
### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -33,9 +33,11 @@ android {
         resValue "bool", "IS_DEBUG", "false"
         buildConfigField "boolean", "USE_RELEASE_VERSIONING", "false"
         buildConfigField "String", "AMO_COLLECTION", "\"7e8d6dc651b54ab385fb8791bf9dac\""
+        def deepLinkSchemeValue = "fenix-dev"
+        buildConfigField "String", "DEEP_LINK_SCHEME", "\"$deepLinkSchemeValue\""
         manifestPlaceholders = [
                 "isRaptorEnabled": "false",
-                "deepLinkScheme": "fenix-dev"
+                "deepLinkScheme": deepLinkSchemeValue
         ]
     }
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/fenix/Config.kt
```diff
@@ -60,6 +60,16 @@ enum class ReleaseChannel {
 
     val isFenix: Boolean
         get() = !isFennec
+
+    val deepLinkScheme: String
+        get() = when (this) {
+            FenixNightly -> "fenix-nightly"
```
jonalmeida, line 66: At the time, when we added the deepLinkScheme, the only use case for it was [in the manifest](https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/AndroidManifest.xml#L82).

I don't think we can avoid the duplication to set the values in the manifest an in the Kotlin layer, but what do you think about using a `buildConfigField` so we set the values in the same `build.gradle` place? We wouldn't need the `Config.deepLinkScheme` parsing and it would be less likely for errors if we decide to change our variants tomorrow (it's quite likely actually with LeanPlum things 😄 ).

rocketsroger: Sure, I'll look into creating a `buildConfigField`. Thanks,

app/build.gradle
```diff
@@ -33,9 +33,11 @@ android {
         resValue "bool", "IS_DEBUG", "false"
         buildConfigField "boolean", "USE_RELEASE_VERSIONING", "false"
         buildConfigField "String", "AMO_COLLECTION", "\"7e8d6dc651b54ab385fb8791bf9dac\""
+        def deepLinkSchemeValue = "fenix-dev"
```
jonalmeida, line 36: Oh nice! I didn't think this would work for some reason.

rocketsroger: Me too.  Luckily found this: https://developer.android.com/studio/build/gradle-tips#share-properties-with-the-manifest

jonalmeida: 👏

Issue #9987, opened 2020-04-16
[Bug] "Secure your logins and passwords" prompt is displayed while the recording in PB is disabled


abodea:
### Preconditions
Fresh install, as this prompt is displayed only once.
## Steps to reproduce
1. Launch Fenix.
2. Reach settings and disable PB recording option.
3. Access `Saved Logins`
### Expected behavior
Nothing is displayed in PB while the recording in PB is off.
### Actual behavior
"Secure your logins and passwords" prompt is displayed while the recording in PB is disabled
### Device information

* Android device: **Google Pixel 4 XL (10).**
* Fenix version: **Nightly 4/16.**

![20200416-132150](https://user-images.githubusercontent.com/42831109/79450807-1b964b80-7fee-11ea-89f0-f3f14e54736c.gif)

mcarare:
The dialog also shows up in Account Settings, and it was secured in https://github.com/mozilla-mobile/fenix/pull/9998. For the dialog in Saved Logins it is secured here: https://github.com/mozilla-mobile/fenix/pull/10134
Please recheck after the second PR is merged. TY!

Diana-Rus:
Hi verified as fixed with Google Pixel 3 XL (Android 9), Samsung Galaxy S9 (Android 8) and Sony Xperia Z5 (Android 7) on Firefox Preview Nightly 20200424
**- Settings/Logins and passwords - "Secure your logins and passwords" prompt**
**Pixel 3 XL - GIF**
![20200424-102701](https://user-images.githubusercontent.com/49859863/80187146-cb377300-8617-11ea-98e0-bf864c456a96.gif)

**Samsung Galaxy S9 - GIF**
![20200424-102234](https://user-images.githubusercontent.com/49859863/80187055-acd17780-8617-11ea-9fb3-c0dfa2b36b61.gif)

**Sony Xperia Z5 - GIF**
![20200424-103110](https://user-images.githubusercontent.com/49859863/80187165-d38fae00-8617-11ea-8756-f32387fc7a29.gif)


**- Log in with a valid account/Account Settings- Check and uncheck Logins checkbox in order to trigger the "Secure your logins and passwords" prompt
**Pixel 3 XL - GIF**
![20200424-115734](https://user-images.githubusercontent.com/49859863/80196298-dbeee580-8625-11ea-9620-a5453a35f960.gif)

**Samsung Galaxy S9 - GIF**
[...]

PR #9998, opened 2020-04-16
 For #9987: Set FLAG_SECURE to dialog when flag set in activity.

mcarare:
### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/ext/Dialog.kt
+++ b/app/src/main/java/org/mozilla/fenix/ext/Dialog.kt
@@ -1,0 +1,24 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.ext
+
+import android.app.Activity
+import android.app.Dialog
+import android.view.WindowManager
+
+/**
+ * Checks if activity's window has a FLAG_SECURE set and sets it to dialog
+ */
+fun Dialog.secure(activity: Activity?) {
+    this.window.apply {
+        val flags = activity?.window?.attributes?.flags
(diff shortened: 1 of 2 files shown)


ekager:
Do we know why this dialog is behaving differently from any other dialog? Seems weird that this would be the only one?

app/src/main/java/org/mozilla/fenix/ext/Dialog.kt
```diff
@@ -0,0 +1,24 @@
...
+package org.mozilla.fenix.ext
+
+import android.app.Activity
+import android.app.Dialog
+import android.view.WindowManager
+
+/**
+ * Checks if activity's window has a FLAG_SECURE set and sets it to dialog
+ */
+fun Dialog.secure(activity: Activity?) {
```
jonalmeida, line 14: This would be cool to have something like this in [`feature-privatemode`](https://github.com/mozilla-mobile/android-components/tree/master/components/feature/privatemode). 🙂

ekager:
Also we also use `showPinDialogWarning` in `LoginsFragment` do we need to do the same thing there?

app/src/main/java/org/mozilla/fenix/ext/Dialog.kt
boek: 👍 added it to the upstreaming board! 

https://github.com/orgs/mozilla-mobile/projects/38

boek:
@mcarare Didn't see @ekager before I landed. Feel free to answer / follow up with a new PR if needed

mcarare:
@ekager IMO we should use on all dialogs even if they do not present sensitive data, but it has to do with the confidence the user has that the setting (do not allow screenshots) is respected. I think the menus also appear on recordings. 

> Do we know why this dialog is behaving differently from any other dialog? Seems weird that this would be the only one?

This is a behaviour in android when the system creates a window for dialog, pop-up, spinner, etc but does not pass all flags, including FLAG_SECURE.

Issue #8428, opened 2020-02-14
About Firefox Nightly  page is not completely translated in some languages


lobontiumira:
## Steps to reproduce
1. Change the language of the device to German / Deutsch, Spanish / Espanol, French / Francais, etc.
2. Open Fenix and go to Settings - About Firefox Nightly and observe the page.

### Expected behavior
The Firefox Nightly page is entirely translated.

### Actual behavior
The "Firefox Nightly is produced by Mozilla." text is not translated.

### Device information
* Android device: all
* Fenix version: 2/14 Nightly build

![Screenshot_20200214-115816](https://user-images.githubusercontent.com/40694158/74521512-2cf5a580-4f22-11ea-8d59-d0d8fe15015e.png)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-15738)

cadeyrn:
The beta migration is almost finished. Could this be resolved before the final migration, please? There should really no untranslated strings in the stable release and it doesn't sound like a difficult problem. Thanks!

Diana-Rus:
Hi, issue is reproducible with Google Pixel 3 XL (Android 9) and Samsung Galaxy S9 (Android 8) on Beta 4.3.0-beta.1.

softvision-miralobontiu:
This is still reproducible on 12/31 Nightly build with Azerbaycan language:

<img src="https://user-images.githubusercontent.com/40694158/103400552-3560ca00-4b4e-11eb-9b37-4ae525d29dda.jpg" width="300" height="550" />

lobontiumira:
The issue still reproduces with Azerbaycan language, on the 6/30 Nightly build, using a Google Pixel (Android 10).

lobontiumira:
This is still reproducible on 11/2 Nightly build with Azerbaycan language, with Google Pixel (Android 10).

sflorean:
This is still reproducible on Fenix versions.

SoftVision-LorandJanos:
This issue is still reproducible on the latest Beta 108.0b3 with the Azerbaycan language:
![Screenshot_2022-11-24-13-55-03-90_ffb2f5e1b976ff98cfc94f359fbce8de](https://user-images.githubusercontent.com/48829350/203778950-73946ffb-6051-4ad7-a575-da034681ab65.jpg)

csadilek:
[...]

PR #10005, opened 2020-04-16
For #8428 - move about text from static_strings.xml to strings.xml

cadeyrn:
I moved the string from static_strings.xml to strings.xml and fixed the comment. From the user's POV nothing changed.

Is something else needed to get this translated?


ekager:
Nope this looks good! Just pulled it into https://github.com/mozilla-mobile/fenix/pull/10009

Issue #9782, opened 2020-04-07
[Bug] Clean up unused FeatureFlags


sblatz:
We currently have quite a few feature flags going unused. We should go through each of these and remove any there are no longer necessary.

grigoryk:
#10017 removes one of the flags, `asFeatureFxAPairingDisabled`.

hakkikaancaliskan:
Exactly which ones needed to remove? Looks like just one of them unused.
https://github.com/mozilla-mobile/fenix/blob/7e6116512c354eb1d8524f8922f7970fe155c65c/app/src/main/java/org/mozilla/fenix/FeatureFlags.kt#L30

NotWoods:
Fixed by 87a65d2

PR #10017, opened 2020-04-17
For #4046: Always display 'Desktop' bookmarks folder

### LINKED ISSUE #4046: [Bug] Desktop Bookmarks are only visible when signed-in
Currently we only display `desktop bookmarks` in the bookmarks library view if the user is signed-in.

If I sync my desktop bookmarks and then sign-out, I am no longer able to access them unless I sign-in again. At least from that point of view, this clearly reads as a bug to me.

I think display of the `Desktop bookmarks` folder in the library and edit views should be predicated on the presence of the desktop roots in the tree instead.

This should work fine for editing as well. If the user wants to move a bookmark into one of the `desktop` folders, there's no problem there as long as we actually display desktop bookmarks in the library. If they re-connect sync, our bookmark merging logic should sensibly merge their local edits with the remote desktop tree. Merging two complete bookmarks trees is one of the benchmark use cases for https://github.com/mozilla/dogear, which is what we use to do the heavy lifting here.

grigoryk:
See https://github.com/mozilla-mobile/fenix/issues/4046 for a detailed discussion of this.

In short, this patch removes code that would conditionally hide desktop bookmarks depending
on the signed-in state of the browser.

### Pull Request checklist

### After merge

1. click on Show All Checks,


grigoryk:
The only place where we would now display a "sign-in" prompt in bookmarks is while inside of the "Desktop Bookmarks" virtual folder, as discussed in https://github.com/mozilla-mobile/fenix/issues/4046#issuecomment-540260829.

In place of a "sign-in banner", I've re-used the existing button. We can improve this in a follow-up. The copy of this button isn't quite right, but that too can be a follow-up.
Screen Shot 2020-04-17 at 12 11 21 AM

app/src/main/res/layout/component_sign_in.xml
```diff
@@ -8,7 +8,7 @@
     android:id="@+id/bookmark_folders_sign_in"
     style="@style/NeutralButton"
     android:text="@string/bookmark_sign_in_button"
-    android:visibility="gone"
```
ekager, line 11: nit - we can remove this line entirely I think? AFAICT we're not messing with the visibility of this view dynamically anymore

app/src/main/java/org/mozilla/fenix/components/Services.kt
```diff
@@ -56,25 +44,4 @@ class Services(
...
     }
-
-    /**
-     * Launches the sign in and pairing custom tab from any screen in the app.
-     * @param context the current Context
-     * @param navController the navController to use for navigation
-     */
-    fun launchPairingSignIn(context: Context, navController: NavController) {
-        // Do not navigate to pairing UI if camera not available or pairing is disabled
-        if (context.hasCamera() && !FeatureFlags.asFeatureFxAPairingDisabled) {
```
ekager, line 67: Ha we were just wondering if we still needed this feature flag the other day. Can you also remove `asFeatureFxAPairingDisabled` from `FeatureFlags`? It's the only place this is used
https://github.com/mozilla-mobile/fenix/issues/9782

ekager:
Thanks for your contribution!

PR #10033, opened 2020-04-18
No issue: Fix leaks in Home and Bookmarks

ekager:
Screenshot_20200417-182847

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
@@ -222,7 +222,7 @@ class HomeFragment : Fragment() {
                 showDeleteCollectionPrompt = ::showDeleteCollectionPrompt,
                 openSettingsScreen = ::openSettingsScreen,
                 openSearchScreen = ::navigateToSearch,
-                openWhatsNewLink = { openCustomTab(SupportUtils.getWhatsNewUrl(view.context)) },
+                openWhatsNewLink = { openCustomTab(SupportUtils.getWhatsNewUrl(activity)) },
                 openPrivacyNotice = { openCustomTab(SupportUtils.getPrivacyNoticeUrl()) }
             )
         )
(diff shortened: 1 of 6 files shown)


ekager:
Screenshot_20200417-192110

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -356,6 +359,7 @@ class HomeFragment : Fragment() {
 
     override fun onDestroyView() {
         super.onDestroyView()
+        _sessionControlInteractor = null
```
grigoryk, line 362: Interesting, I guess this helps to GC this earlier - I wonder if it's fair to call this a leak in a true sense. Once HomeFragment goes through `onCreateView` again, the old instance of sessionControlInteractor will no longer be referenced, and it will get GC'd. But, that won't happen _until_ user goes back to HomeFragment, so we are retaining this stuff in memory for longer than necessary.

You _could_ similarly null more things here, `currentMode`, `homeFragmentStore`, etc - same logic applies.

app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkFragment.kt
```diff
@@ -298,6 +300,11 @@ class BookmarkFragment : LibraryPageFragment<BookmarkNode>(), UserInteractionHan
         )
     }
 
+    override fun onDestroyView() {
+        super.onDestroyView()
+        _bookmarkInteractor = null
```
grigoryk, line 305: Why is this needed?

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
ekager: The main problem with these two objects specifically is that it they retains references to the fragment views that are being destroyed so they are leaking and this can't be GC I believe. Here specifically we are passing in a lambda that references the fragment view and in bookmarks, we pass in a view reference for the snackbar.

ekager: https://stackoverflow.com/a/59504797

app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkFragment.kt
ekager: Here's a second possible fix to the same leak: https://github.com/mozilla-mobile/fenix/compare/master...ekager:bookmark-snackbar-fix2

Maybe slightly preferable since I really don't think the interactor should have references to any views to begin with... But also always leaves the possibility that someone else adds a reference later and regresses that fix. The definitive nulling out of the interactor protects against that. 🤷‍♀️

Issue #9856, opened 2020-04-10
[Bug] "Successfully installed uBlock Origin" is not aligned with the device navigation bar


Diana-Rus:
## Steps to reproduce
1. Open Fenix and go to "Settings"/"Add-ons".
2. Install uBlock Origin addon by tapping on the "+" button .
3. From the "Add uBlock Origin" pop up tap on "Add".
### Expected behavior
The "Successfully installed uBlock Origin" snackbar is displayed correctly.
### Actual behavior
The snackbar is not aligned with the navigation bar.
### Device information

* Android device: Samsung Galaxy S9 (Android 8)
* Fenix version: debug build based on master with(39.0.20200409130114, 52ca085fb) -  GV: 77.0a1-20200408033650

GIF:
![20200410-191305](https://user-images.githubusercontent.com/49859863/79006766-8a912180-7b62-11ea-8259-0edc36671bbd.gif)

Note: It may impact https://github.com/mozilla-mobile/fenix/issues/9196

csadilek:
Closing as we no longer have the "successfully installed" snackbar. It was replaced with this dialog: https://github.com/mozilla-mobile/fenix/issues/10398#issuecomment-624313139

PR #10061, opened 2020-04-20
For #9856 - AddonsManagementFragment: Use FenixSnackbar for install msg

JLRoberts:
Commit 0def456 uses activity rootview if available to show snackbars,
if using this view we need to be able to set to not use toolbar padding.

Before:
Screenshot_20200420-103725

After change:
Screenshot_20200420-112615

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/addons/AddonsManagementFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/addons/AddonsManagementFragment.kt
@@ -31,6 +31,7 @@ import mozilla.components.feature.addons.ui.translatedName
 import mozilla.components.lib.state.ext.flowScoped
 import mozilla.components.support.ktx.kotlinx.coroutines.flow.ifChanged
 import org.mozilla.fenix.R
+import org.mozilla.fenix.components.FenixSnackbar
 import org.mozilla.fenix.ext.components
 import org.mozilla.fenix.ext.getRootView
 import org.mozilla.fenix.ext.showToolbar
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/fenix/addons/AddonsManagementFragment.kt
```diff
@@ -203,13 +204,16 @@ class AddonsManagementFragment : Fragment(R.layout.fragment_add_ons_management),
...
-                    showSnackBar(
-                        rootView,
-                        getString(
-                            R.string.mozac_feature_addons_successfully_installed,
-                            it.translatedName
-                        )
+                    FenixSnackbar.make(
+                        view = rootView,
+                        duration = FenixSnackbar.LENGTH_SHORT,
+                        isDisplayedWithBrowserToolbar = false
```
psymoon, line 210: Hi @JLRoberts, thanks for this! So your _suggested_ patch could as simple as setting `isDisplayedOnBrowserFragment=false` in [here](https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/java/org/mozilla/fenix/addons/Extensions.kt#L32).

The reasoning behind attaching `rootView` was so that the snackbar would stay visible when you press back to `BrowserFragment`. So by setting `isDisplayedOnBrowserFragment=false`, would it mess up the alignment for when you go back to `BrowserFragment` after the confirmation Snackbar is shown? I think this is a good start to the underlying padding issue. What do you think?

JLRoberts: I'll check. I thought that previous patch was for fixing some snackbars not being shown depending on the behavior of the add-on. As it seems some are different. I changed it specifically here because it looks like there are alot of snackbars that use the extension method and some need the padding. Some don't. So may need a separate one for showing without padding or just using the actual methods if more crop up to need no padding.

csadilek:
Thanks for the fix @JLRoberts. Based on other UX feedback we have landed a new install confirmation UI in the meantime that replaces the Snackbar: https://github.com/mozilla-mobile/fenix/pull/10608

Issue #9975, opened 2020-04-16
Add telemetry for duration to enter/exit fullscreen


mcomella:
As part of my fullscreen performance investigation https://github.com/mozilla-mobile/fenix/issues/9341, we should add telemetry to:
1) determine how enter/exit fullscreen performs on all devices (by adding telemetry locally, we can test consistently on our own devices too)
2) potentially watch for regressions to this behavior, depending on how noisy the measurement is

mcomella:
I need help from the Glean team: as far as I can tell, with the current Glean API, it's difficult to take this measurement. I'll reach out to them on Matrix.

mcomella:
I got help from the glean team but I want to double-check with ecsmyth why we want telemetry before we invest more time in making a proper telemetry solution. My WIP with mostly functional telemetry is here: https://github.com/mcomella/fenix/tree/9975-fullscreen-telemetry

mcomella:
> I want to double-check with ecsmyth why we want telemetry before we invest more time in making a proper telemetry solution. 

i.e. there are two thoughts:
1) Can we test locally on a low-end device to rule this out as a perf problem area? (if so, no telemetry needed)
2) Are we intending to use this telemetry to monitor for regressions? (if not, no telemetry needed)

mcomella:
Copying my question from Matrix so it's less likely to get lost. @ecsmyth, please respond!

>  Relatedly, for our non-startup use cases, why do we want to record telemetry vs. just measuring locally? Is that to prevent regressions (rather than identifying if there's a problem to begin with)?

ecsmyth:
My recommendation is that we close this issue as `won't fix`. IIUC the widespread use of MSE means this aspect of the UX is controlled by the JS player used by a particular site to play the video (i.e. largely out of our ability to control/impact).

mcomella:
wfm. My WIP PR is here: https://github.com/mozilla-mobile/fenix/pull/10068

Issue #9341, opened 2020-03-23
Investigate performance of video use case


emalysz:
_edit (mcomella)_: I expanded the scope of the bug to be all video use cases.

One of our non-start up user flows (detailed in #7380) is the performance of video, and one component of that would be making videos full screen.

I have filed the following issues that impact the perceived performance:

- https://github.com/mozilla-mobile/fenix/issues/9339: The bottom nav bar has an obvious flicker when moving to full screen.
- https://github.com/mozilla-mobile/fenix/issues/9340: The entire UI will flicker if the link automatically directs to a full screen video

mcomella:
On my Pixel 2, I aggressively seeked, rotated the device, and play/paused and **was unable to detect any performance issues in fullscreen mode and non-fullscreen mode.** This is a big improvement over my past experiences where seeking would take a very long time.

However, I did find two additional bugs that are out of scope for the perf group:
- https://github.com/mozilla-mobile/fenix/issues/9939 video controls are tiny on non-mobile pages
- https://github.com/mozilla-mobile/fenix/issues/9940 Calling HTML5 APIs rapidly may permanently pause video

mcomella:
I re-ran my test using 720p on YouTube (the default for me was 360p) and I was still unable to identify issues. I checked the NFR and the only RC there are dropped frames, which is covered by automated testing. I think we've covered this issue comprehensively.

Ideally I could have tested on a lower end device but I do not have one available to me.

@ecsmyth Can you confirm this is what you want to see out of investigations into non-startup use cases?

mcomella:
I filed a separate issue for picture-in-picture performance analysis because it's a less used case than general video playback and it's not released yet: https://github.com/mozilla-mobile/fenix/issues/9943

ecsmyth:
I'm checking with the Media team to confirm, but I think that we've hit the key points for Fenix responsiveness (as opposed to issues with the players themselves) related to video playback.

mcomella:
As a sanity check, I took profiles during these fullscreen operations to see if we were doing anything silly. I found https://github.com/mozilla-mobile/fenix/issues/9972 but nothing else.

mcomella:
Last action from me: in https://github.com/mozilla-mobile/fenix/pull/10068, I'm asking MarcLeclair to measure the duration it takes to enter and exit fullscreen mode on the Moto G5. Otherwise, still waiting on ecsmyth & the Media team.

ecsmyth:
[...]

PR #10068, opened 2020-04-20
DO NOT LAND: for #9975: measure fullscreen performance

mcomella:
@MarcLeclair can you 1) review this code for correctness and 2) use this code to measure how long it takes to enter and exit fullscreen mode? The code won't be landing so just review for correctness, not other things like how easy is it to maintain. The steps to measure are:
- Install the build
- Go to a page, open and close fullscreen (to close, use the in-page button and back button)
- Execute `adb shell am start -n "org.mozilla.<app_id>/mozilla.telemetry.glean.debug.GleanDebugActivity" --ez logPings true --es sendPing metrics --es tagPings performance`, switching out the app ID, to send the measurements to the glean debug view
- Check the Glean Debug View for the result; feel free to send it to me if you want me to interpret
  - Note that due to the metric type we capture here (Timing Distribution), the values are bucketed rather than actual measurements

### Pull Request checklist

### After merge

1. click on Show All Checks,


mcomella:
I tested on my GS5 on youtube.com:
- 319ms, 380ms, and 638ms to enter fullscreen, rounded to Glean bucketing
- 123ms, 615ms 2x to exit fullscreen, rounded to Glean bucketing

This doesn't match our responsiveness goals but as Eric mentioned https://github.com/mozilla-mobile/fenix/issues/9341#issuecomment-616951023:
> Existing investigation found a number of functional issues, however, the responsiveness of this portion of he UX is predominately controlled by the MSE JS players.

So I'm not sure how actionable this is.

mcomella:
I'll wait for Marc to complete his analysis so we can compare the two devices and double-check with ecsmyth that there is nothing to do here.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -211,6 +213,8 @@ open class HomeActivity : LocaleAwareAppCompatActivity() {
     }
 
     final override fun onBackPressed() {
+        FullscreenTransitionMeasurement.onBackPressed() // DO NOT MOVE ANYTHING ABOVE HERE.
```
MarcLeclair, line 216: is there a reason why we can't move anything above here? Just wondering.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -219,6 +223,11 @@ open class HomeActivity : LocaleAwareAppCompatActivity() {
         super.onBackPressed()
     }
 
+    override fun dispatchTouchEvent(ev: MotionEvent): Boolean {
+        FullscreenTransitionMeasurement.onHomeActivityDispatchTouchEvent(ev) // DO NOT MOVE ANYTHING ABOVE HERE.
```
MarcLeclair, line 227: same question as above, is there a reason why?

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
mcomella: If code is moved above here and it takes a long time, it won't be included in the measurement: we want the measurement to happen first.

I wonder if there's a more foolproof way to do this (e.g. with annotations?).

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
mcomella: Same answer. :)

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
MarcLeclair: So, I think with anything like this, we shouldn't couple performance code with our app code. I was thinking more of an AOP way ( I'm thinking the way Spring does things) but looking around quickly I found this article: https://medium.com/dont-code-me-on-that/aspect-oriented-programming-in-kotlin-95ff8598913 which highlights the way we can do it in kotlin :) THoughts?

mcomella: Oh interesting: that's pretty neat and avoids the footgun of someone moving code where they're not supposed to. Sure, let's aim for something like that in the future (though I'm going to do a little more reading on AOP to understand it more holistically).

MarcLeclair: Yeah for sure, if we don't want to do this here, I think we can go ahead with this PR. It helps with less coupling. 😄

mcomella: We're not intending to land this: sorry if that wasn't clear. The code is just here for you to take measurements.

MarcLeclair:
So for the G5+ I'm getting  approximately 355ms to enter Fullscreen and ~1.4 seconds to exit full screen.

mcomella:
That's all we need for this PR: I will complete the analysis and discussion in the comments following https://github.com/mozilla-mobile/fenix/issues/9341#issuecomment-616951023.

PR #10077, opened 2020-04-21
Close #9338: Enable WebPush engine integration in all builds

### LINKED ISSUE #9338: Remove nightly-only feature flag for WebPush
Once [bug 1619778][0] lands in GeckoView beta (76+), we can remove [the feature flag][1] from Fenix and let it ride the trains.

The next GV beta is scheduled for April 7th.

[0]: https://bugzilla.mozilla.org/show_bug.cgi?id=1619778
[1]: https://github.com/mozilla-mobile/fenix/blob/9feb44852b3970d94de9acaccfb797c20e65ae95/app/src/main/java/org/mozilla/fenix/FeatureFlags.kt#L62-L69

jonalmeida:


ekager:
Are we keeping the feature flag around just in case we want to toggle it off again?

jonalmeida:
> again

For now, yes. :)

jonalmeida:
Forgot to amend my commit before pushing.

PR #10084, opened 2020-04-21
Fixes #10064: Open help articles in normal tab instead of custom tab

### LINKED ISSUE #10064: FNX-3266 ⁃ [Bug] Open settings links as regular links, not custom tabs
Currently our settings links (like in the about page) open up as custom tab links. According to Paul on the SUMO team this makes it difficult to follow advice from the help articles since they talk about doing things in a normal tab. We should update these to just use `openToBrowserAndLoad` instead.

ValentinTimisica:
### Pull Request checklist

- [x] **Screenshots**: This PR does not include screenshots or GIFs.

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/androidTest/java/org/mozilla/fenix/helpers/Matchers.kt
+++ b/app/src/androidTest/java/org/mozilla/fenix/helpers/Matchers.kt
@@ -7,6 +7,10 @@ package org.mozilla.fenix.helpers
 import android.graphics.Bitmap
 import android.view.View
 import android.view.ViewGroup
+import androidx.test.espresso.ViewInteraction
+import androidx.test.espresso.assertion.ViewAssertions.matches
+import androidx.test.espresso.matcher.ViewMatchers
+import junit.framework.AssertionFailedError
 import org.hamcrest.CoreMatchers.not
 import org.hamcrest.Description
 import org.hamcrest.Matcher
@@ -55,4 +59,14 @@ fun nthChildOf(
             return parentMatcher.matches(view.parent) && group.getChildAt(childPosition) == view
         }
     }
+}
+
+fun ViewInteraction.isVisibleForUser(): Boolean {
+    try {
+        check(matches(ViewMatchers.isCompletelyDisplayed()))
+    } catch (assertionError: AssertionFailedError) {
+        return false
+    }
+
+    return true
 }
--- a/app/src/androidTest/java/org/mozilla/fenix/helpers/TestHelper.kt
+++ b/app/src/androidTest/java/org/mozilla/fenix/helpers/TestHelper.kt
@@ -17,6 +17,7 @@ import androidx.test.espresso.action.GeneralClickAction
 import androidx.test.espresso.action.Press
 import androidx.test.espresso.action.Tap
 import androidx.test.espresso.action.ViewActions.longClick
+import androidx.test.espresso.assertion.ViewAssertions
 import androidx.test.espresso.matcher.ViewMatchers.withId
 import androidx.test.espresso.matcher.ViewMatchers.withText
 import androidx.test.platform.app.InstrumentationRegistry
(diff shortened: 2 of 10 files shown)


ekager:
@sblatz do you still have time to review this?

app/src/androidTest/java/org/mozilla/fenix/ui/robots/SettingsSubMenuAboutRobot.kt
```diff
@@ -96,106 +112,102 @@ private fun assertCurrentTimestamp() {
...
 
 private fun assertKnowYourRights() {
-    TestHelper.scrollToElementByText("Know your rights")
+    if (!onView(withText("Know your rights")).isVisibleForUser()) {
+        onView(withId(R.id.about_layout)).perform(ViewActions.swipeUp())
+    }
+
     onView(withText("Know your rights"))
-        .check(matches(withEffectiveVisibility(Visibility.VISIBLE)))
+        .check(matches(withEffectiveVisibility(ViewMatchers.Visibility.VISIBLE)))
```
sblatz, line 174: Is there a reason we're using `isVisibleForUser` in some cases an not others like this?

app/src/androidTest/java/org/mozilla/fenix/helpers/TestHelper.kt
```diff
@@ -79,6 +81,19 @@ object TestHelper {
         )
     }
 
+    fun verifyUrl(urlSubstring: String, resourceName: String, resId: Int) {
+        waitUntilObjectIsFound(resourceName)
+        onView(withId(resId)).check(
+            ViewAssertions.matches(
+                withText(
+                    CoreMatchers.containsString(
+                        urlSubstring
```
sblatz, line 90: Can we clean this up a bit? I don't think we need this many newlines for this method. The linter doesn't seem to complain if I just make it one line so I'd recommend doing that:
`onView(withId(resId)).check(ViewAssertions.matches(withText(CoreMatchers.containsString(urlSubstring)))`

app/src/androidTest/java/org/mozilla/fenix/helpers/Matchers.kt
```diff
@@ -56,3 +60,13 @@ fun nthChildOf(
         }
     }
 }
+
+fun ViewInteraction.isVisibleForUser(): Boolean {
+    try {
+        check(matches(ViewMatchers.isCompletelyDisplayed()))
```
sblatz, line 66: @kglazko / @rpappalax are you aware of any issues with using this over something like `scrollToElementByText`? 🤔

ValentinTimisica: The problem that I encountered here was that on all test methods from `SettingsSubMenuAboutRobot` scroll was performed (example: TestHelper.scrollToElementByText("What’s new in Firefox Preview")). This test class was flaky because on larger devices it could not perform this scroll and it would throw an error (this approach worked only for smaller devices).

That being said I had to find a solution to scroll only when needed. Another problem was that `perform(click())` will work only if the view is at least 90% visible, so I had to make sure that the view is completely visible. This is why I used `isVisibleForUser` extension where I check if `ViewMatchers.isCompletelyDisplayed()`.

app/src/androidTest/java/org/mozilla/fenix/ui/robots/SettingsSubMenuAboutRobot.kt
ValentinTimisica: It is not necessarily to use it here because it is clearly that the view is 100% visible since we already checked above that it is completely visible (check the comment below).

ValentinTimisica:
I saw that `validateBaselinePing` test is failing after rebasing but it's not related to this task. This test also fail on other PRs like https://github.com/mozilla-mobile/fenix/pull/10688.

Issue #6123, opened 2019-10-18
WebExtensions in Private Browsing


vesta0:
### User Story
As a user, I want to choose whether I want to allow a WebExtensions in private browsing, so I have control over the data and permissions shard in private browsing.

### Dependencies
[A-C project board](https://github.com/orgs/mozilla-mobile/projects/44)

### Acceptance Criteria

- I can choose to allow a WebExtension to run in private browsing
- This option is at the individual WebExtension level

brampitoyo:
Progress report:
* In Q4 2019, we’ve decided that WebExtensions will run in both normal and private browsing modes
* In the future, we may provide a choice to disable individual WebExtension in private browsing

csadilek:
Blocked on: https://bugzilla.mozilla.org/show_bug.cgi?id=1599139

csadilek:
@brampitoyo As the default for this is "false" (add-ons do not run in private browsing until users turn it off) do we need some form of hint / notification? This is re: https://github.com/mozilla-mobile/fenix/issues/10120#issuecomment-618999870.

Defaulting to false seems correct as this is what Desktop does, just wondering if we need to make this more obvious to users?

brampitoyo:
@csadilek If these are true:

- Many popular add-ons – and most add-ons that Fenix supports – are ad blockers and/or focused around security
- Private browsing typically demands more stringent security, stricter trackers blocking, etc.

Then does it make sense to set our default policy to TRUE: allow add-on to run in private browsing?

I immediately see huge problems around consent. Add-ons may “phone home” and leak private data (e.g. URLs opened in private browsing). But does our Recommended Extension programme not protect against this?

I think that many users’ expectation is for add-ons to work everywhere by default, but also for those add-ons to not leak data while browsing privately.

-----

Either way, I do agree that WebExtensions that run in private browsing should be clearly labelled.

I propose doing something similar to desktop: have a small (16×16dp) private browsing icon to the side of the add-on name. Thoughts?

<img width="360" src="https://user-images.githubusercontent.com/51007/80325454-4c497100-8889-11ea-9687-d85db86d5e70.png">

Icon asset: https://dzwonsemrish7.cloudfront.net/items/2D2h3z2m2P3G0Q2a3I0y/private-browsing.svg

AndiAJ:
Hi, **verified as fixed** on the latest **`Nightly Build #21180609 from 4/27`** using the following devices:
• Google Pixel 3a (Android 10)
[...]

PR #10093, opened 2020-04-21
For #6123: Expose WebExtensions in private browsing control

psymoon:
### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/addons/InstalledAddonDetailsFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/addons/InstalledAddonDetailsFragment.kt
@@ -81,6 +81,7 @@ class InstalledAddonDetailsFragment : Fragment() {
         bindSettings(view)
         bindDetails(view)
         bindPermissions(view)
+        bindAllowInPrivateBrowsingSwitch(view)
         bindRemoveButton(view)
     }
 
@@ -211,6 +212,32 @@ class InstalledAddonDetailsFragment : Fragment() {
         }
     }
 
+    private fun bindAllowInPrivateBrowsingSwitch(view: View) {
+        val switch = view.allow_in_private_browsing_switch
+        switch.isChecked = addon.isAllowedInPrivateBrowsing()
+        switch.setOnCheckedChangeListener { v, isChecked ->
+            val addonManager = v.context.components.addonManager
+            switch.isClickable = false
+            view.remove_add_on.isEnabled = false
+            addonManager.setAddonAllowedInPrivateBrowsing(
+                addon,
+                isChecked,
+                onSuccess = {
+                    runIfFragmentIsAttached {
+                        this.addon = it
+                        switch.isClickable = true
+                        view.remove_add_on.isEnabled = true
+                    }
+                },
+                onError = {
+                    runIfFragmentIsAttached {
+                        switch.isClickable = true
+                        view.remove_add_on.isEnabled = true
+                    }
+                }
+            )
+        }
+    }
     private fun bindRemoveButton(view: View) {
         view.remove_add_on.setOnClickListener {
             setAllInteractiveViewsClickable(view, false)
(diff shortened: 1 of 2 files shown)


psymoon:
@brampitoyo here are the screenshots:

Screenshot_20200421-154237
Screenshot_20200421-154244

app/src/main/java/org/mozilla/fenix/addons/InstalledAddonDetailsFragment.kt
```diff
@@ -211,6 +212,28 @@ class InstalledAddonDetailsFragment : Fragment() {
...
+            addonManager.setAddonAllowedInPrivateBrowsing(
+                addon,
+                isChecked,
+                onSuccess = {
+                    runIfFragmentIsAttached {
+                        this.addon = it
+                        switch.isClickable = true
+                        switch.isChecked = isChecked
+                        view.settings.isVisible = shouldSettingsBeVisible()
+                        view.remove_add_on.isEnabled = true
```
Amejia481, line 231: I'm not sure if we need these lines? Is there any specific reason? :)

app/src/main/java/org/mozilla/fenix/addons/InstalledAddonDetailsFragment.kt
```diff
@@ -211,6 +212,28 @@ class InstalledAddonDetailsFragment : Fragment() {
...
+            switch.isClickable = false
+            view.remove_add_on.isEnabled = false
+            addonManager.setAddonAllowedInPrivateBrowsing(
+                addon,
+                isChecked,
+                onSuccess = {
+                    runIfFragmentIsAttached {
+                        this.addon = it
+                        switch.isClickable = true
+                        switch.isChecked = isChecked
```
Amejia481, line 229: Is this going to re-triger `setOnCheckedChangeListener`? Why do we need to change its value again?

app/src/main/java/org/mozilla/fenix/addons/InstalledAddonDetailsFragment.kt
```diff
@@ -211,6 +212,28 @@ class InstalledAddonDetailsFragment : Fragment() {
...
+        val switch = view.allow_in_private_browsing_switch
+        switch.isChecked = addon.isAllowedInPrivateBrowsing()
+        switch.setOnCheckedChangeListener { v, isChecked ->
+            val addonManager = v.context.components.addonManager
+            switch.isClickable = false
+            view.remove_add_on.isEnabled = false
+            addonManager.setAddonAllowedInPrivateBrowsing(
+                addon,
+                isChecked,
+                onSuccess = {
```
Amejia481, line 225: Should we also handle `onError` ? As we are disabling the the clicks `switch.isClickable = false`

app/src/main/java/org/mozilla/fenix/addons/InstalledAddonDetailsFragment.kt
Amejia481: Ok, we are changing  ` view.remove_add_on.isEnabled = false` when the switch is clicked and checking it back in `onSuccess` that means the we need to handle `onError` because if I'm error happens 
` view.remove_add_on.isEnabled = false` will persisted

psymoon: We are getting new `addon` instance reference so these are safety UI "locks" similarly to enable/disable.

app/src/main/java/org/mozilla/fenix/addons/InstalledAddonDetailsFragment.kt
psymoon: We do the same thing `Switch.setState()` for enable/disable.

app/src/main/java/org/mozilla/fenix/addons/InstalledAddonDetailsFragment.kt
psymoon: Nice catch, we should reset the UI elements again on any callbacks :).

brampitoyo:
Hi @psymoon, just a couple of minor feedback:

- Match desktop strings: “Run in private browsing”
- Reorder so it appears as the second item from the top, below “Enabled”
- Show the mask icon to the left of the list item

Issue #10162, opened 2020-04-23
FNX-3326 ⁃ [Bug] When enable/disable or settting allowing private browsing fails, should go back to previous state


psymoon:
## Steps to reproduce
* Go to Addon detail page.
* Enable/disable or set "Run in PrivateBrowsing"  switch on/off

### Expected behavior
* When enable/disable or set "Run in PrivateBrowsing"  switch on/off, UI is expected to go back to previous state.

### Actual behavior
* UI switches on/off even though the action failed.

### Device information
* Android device: ?
* Fenix version: ?

PR #10164, opened 2020-04-23
For #10162: Set switch back to prev state when failed to enable/disable addon

psymoon:
Debugged and verified that setting `switch.isChecked`  inside `switch.setOnCheckedChangeListener` doesn't fire up the callback again.

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/addons/InstalledAddonDetailsFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/addons/InstalledAddonDetailsFragment.kt
@@ -116,6 +116,7 @@ class InstalledAddonDetailsFragment : Fragment() {
                         runIfFragmentIsAttached {
                             switch.isClickable = true
                             view.remove_add_on.isEnabled = true
+                            switch.setState(addon.isEnabled())
                             showSnackBar(
                                 view,
                                 getString(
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/fenix/addons/InstalledAddonDetailsFragment.kt
```diff
@@ -149,6 +151,7 @@ class InstalledAddonDetailsFragment : Fragment() {
                         runIfFragmentIsAttached {
                             switch.isClickable = true
                             view.remove_add_on.isEnabled = true
+                            switch.setState(currState)
```
Amejia481, line 154: To be safe should we fallback to `switch.setState(addon.isEnabled())` instead? As it will hold the latest successful `isEnabled` value, because `currState` capture the value of `addon.isEnabled()` when `bindEnableSwitch` was executed, and it could contain the wrong value as multiple successful to `enableAddon/disableAddon `can happen in between

Amejia481: The same applies to `bindAllowInPrivateBrowsingSwitch`.
Another alternative is to move `val currState = addon.isEnabled()` inside of `switch.setOnCheckedChangeListener { v, isChecked ->`

psymoon: Sounds good to me.

Issue #8648, opened 2020-02-21
[Library] Deleting is inconsistent between history/bookmarks


AmyYLee:
1. Deleting one bookmark has different strings than group deleting bookmarks in the snackbar
2. Deleting all of history has confirmation pop-up and no snackbar
3. Deleting individual history items has no  confirmation pop-up or snackbar
4. Deleting individual bookmark from edit menu has confirmation pop-up and snackbar (location is too high on the screen)
5. Deleting a folder doesn't show confirmation pop-up, shows snackbar

@betsymi for strings @topotropic to confirm expected behaviour

betsymi:
@AmyYLee Can you help me track down these strings? Any screenshots available?

AmyYLee:
> @AmyYLee Can you help me track down these strings? Any screenshots available?

Spoke to @betsymi about strings and it's been resolved.

betsymi:
1. **Deleting bookmarks snackbars**
1 bookmark deleted: `Deleted [bookmark url] UNDO`
 2+ bookmarks deleted: `Bookmarks deleted UNDO`

1. **Deleting all history snackbar**
`Browsing data deleted`

3. **Deleting individual history item**
`Deleted [url]`

4. N/A for strings, keep as-is

5. **Deleting a folder**
`Are you sure you want to delete this folder?
CANCEL DELETE`

BMEduard:
Hello @betsymi, @AmyYLee. I have two questions regarding this issue:

1. Adding a confirmation pop-up for deleting individual history items will create a discrepancy since deleting individual browsing items (except folders) does not require a confirmation from the user. 
Should I still go ahead and add the confirmation pop-up or should this additional step be kept for more “impactful” actions such as deleting a bookmark folder or for deleting all the history?
2. What should the string message be for the snackbar when the user deletes multiple history items?

I also want to mention that while working on this issue I have also come across some other possible discrepancies which are listed in this new ticket: #9949 .

AmyYLee:
> Hello @betsymi, @AmyYLee. I have two questions regarding this issue:
> 
>     1. Adding a confirmation pop-up for deleting individual history items will create a discrepancy since deleting individual browsing items (except folders) does not require a confirmation from the user.
>        Should I still go ahead and add the confirmation pop-up or should this additional step be kept for more “impactful” actions such as deleting a bookmark folder or for deleting all the history?
> 
>     2. What should the string message be for the snackbar when the user deletes multiple history items?
> 
> 
[...]

PR #10181, opened 2020-04-24
For issue #8648 Delete inconsistencies between history and bookmark items

BMEduard:
1. Updated the strings to account for the deletion of one or more bookmark items
2. Added the string and the snackbar to confirm the deletion of all history
3. Added the string and snackbar to confirm the deletion of one or more history items
4. Set the flag for isDisplayedWithBrowserToolbar to false since there is no browser toolbar shown in that screen
5. Added the dialog for deleting a folder by adding another method to the interactor and controller that is called when the user deletes a single bookmark folder item

Also, I removed an " character from a test method inside CrashReporterControllerTest that was causing an error like this: https://youtrack.jetbrains.com/issue/KT-17438

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/library/bookmarks/edit/EditBookmarkFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/library/bookmarks/edit/EditBookmarkFragment.kt
@@ -166,7 +166,7 @@ class EditBookmarkFragment : Fragment(R.layout.fragment_edit_bookmark) {
                             bookmarkNode?.let { bookmark ->
                                 FenixSnackbar.make(
                                     view = activity.getRootView()!!,
-                                    isDisplayedWithBrowserToolbar = true
+                                    isDisplayedWithBrowserToolbar = args.requiresSnackbarPaddingForToolbar
                                 )
                                     .setText(
                                         getString(
--- a/app/src/main/java/org/mozilla/fenix/library/history/HistoryFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/library/history/HistoryFragment.kt
@@ -28,6 +28,7 @@ import mozilla.components.support.base.feature.UserInteractionHandler
 import org.mozilla.fenix.BrowserDirection
 import org.mozilla.fenix.HomeActivity
 import org.mozilla.fenix.R
+import org.mozilla.fenix.addons.showSnackBar
 import org.mozilla.fenix.browser.browsingmode.BrowsingMode
 import org.mozilla.fenix.components.Components
 import org.mozilla.fenix.components.FenixSnackbar
@@ -38,6 +39,7 @@ import org.mozilla.fenix.ext.components
 import org.mozilla.fenix.ext.nav
 import org.mozilla.fenix.ext.requireComponents
 import org.mozilla.fenix.ext.showToolbar
+import org.mozilla.fenix.ext.toShortUrl
 import org.mozilla.fenix.library.LibraryPageFragment
 
 @SuppressWarnings("TooManyFunctions", "LargeClass")
@@ -106,6 +108,7 @@ class HistoryFragment : LibraryPageFragment<HistoryItem>(), UserInteractionHandl
     }
 
     private fun deleteHistoryItems(items: Set<HistoryItem>) {
+        val message = getMultiSelectSnackBarMessage(items)
         viewLifecycleOwner.lifecycleScope.launch {
             context?.components?.run {
                 for (item in items) {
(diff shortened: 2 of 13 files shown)


app/src/main/java/org/mozilla/fenix/library/history/HistoryFragment.kt
```diff
@@ -155,11 +159,13 @@ class HistoryFragment : LibraryPageFragment<HistoryItem>(), UserInteractionHandl
             true
         }
         R.id.delete_history_multi_select -> {
+            val message = getMultiSelectSnackBarMessage(selectedItems)
             viewLifecycleOwner.lifecycleScope.launch(Main) {
                 deleteSelectedHistory(historyStore.state.mode.selectedItems, requireComponents)
                 viewModel.invalidate()
                 historyStore.dispatch(HistoryFragmentAction.ExitDeletionMode)
             }
+            showSnackBar(view!!, message)
```
ekager, line 168: nit - can you move this inside the viewLifecycleOwner.lifecycleScope launch? It _should_ ensure there's a view that exists when we hit this

app/src/main/java/org/mozilla/fenix/library/history/HistoryFragment.kt
```diff
@@ -193,6 +199,17 @@ class HistoryFragment : LibraryPageFragment<HistoryItem>(), UserInteractionHandl
         else -> super.onOptionsItemSelected(item)
     }
 
+    private fun getMultiSelectSnackBarMessage(historyItems: Set<HistoryItem>): String {
+        return if (historyItems.size > 1) {
+            getString(R.string.history_delete_multiple_items_snackbar)
+        } else {
+            getString(
+                R.string.history_delete_single_item_snackbar,
+                historyItems.first().url.toShortUrl(context!!.components.publicSuffixList)
```
ekager, line 208: Instead of a double bang can we use `requireComponents()` here?

app/src/main/java/org/mozilla/fenix/library/history/HistoryFragment.kt
```diff
@@ -226,6 +243,7 @@ class HistoryFragment : LibraryPageFragment<HistoryItem>(), UserInteractionHandl
                     }
 
                     dialog.dismiss()
+                    showSnackBar(view!!, getString(R.string.preferences_delete_browsing_data_snackbar))
```
ekager, line 246: Same nit as above - can we move this into the `launch(Main)` above to tie this to the view lifecycle?

app/src/main/java/org/mozilla/fenix/library/bookmarks/edit/EditBookmarkFragment.kt
```diff
@@ -166,7 +166,7 @@ class EditBookmarkFragment : Fragment(R.layout.fragment_edit_bookmark) {
                             bookmarkNode?.let { bookmark ->
                                 FenixSnackbar.make(
                                     view = activity.getRootView()!!,
-                                    isDisplayedWithBrowserToolbar = true
+                                    isDisplayedWithBrowserToolbar = false
```
ekager, line 169: We added this because if you add a bookmark from the browser and then click the "edit" button in the snackbar, you get to this view, and then if you save the changes, you get back to the browser. Without adding a flag this will be wrong somewhere :/

app/src/main/java/org/mozilla/fenix/library/history/HistoryFragment.kt
BMEduard: Agreed, moved it inside the scope.

app/src/main/java/org/mozilla/fenix/library/history/HistoryFragment.kt
BMEduard: Agreed, moved it inside the scope.

app/src/main/java/org/mozilla/fenix/library/history/HistoryFragment.kt
BMEduard: Yes, modified it.

app/src/main/java/org/mozilla/fenix/library/bookmarks/edit/EditBookmarkFragment.kt
BMEduard: I added a safe argument flag to this fragment so we know if the snackbar should account for the toolbar being visible or not.
It's default value is set to false, so we only need to add it when we navigate to the EditBookmarkFragment from the "edit" button of the snackbar or from the "star" icon that navigates to the edit fragment if the page is already added to the bookmarks.

PR #10183, opened 2020-04-24
Avoid allowaccessmodification in proguard

badboy:
I'm currently working on changing some internal details of Glean, that uses JNA internally for native function calls.
While testing this new code in Fenix I was facing a crash like this:

```
E  Caused by: java.lang.IllegalArgumentException: Invalid Structure field in class mozilla.telemetry.glean.net.FfiPingUploadTask$ByValue, field name 'activeField' (class com.sun.jna.Structure$StructField): The type "com.sun.jna.Structure$StructField" is not supported: Native size for type "com.sun.jna.Structure$StructField" is unknown
```

Fenix is build with `./gradlew app:assemblegeckonightlyFenixNightly` and tested in the simulator (x86).
Setting `minifyEnabled` and `shrinkResources` to `false` in the `app/build.gradle` mitigated the crash.

This led me down to the Proguard/r8 configuration and with some trial and error I was able to trim it down to the `-allowaccessmodification` line introduced by the `proguard-android-optimize.txt` file (included in [`app/build.gradle`](https://github.com/mozilla-mobile/fenix/blob/6f5f48c30efe8b010e6047b8a200a042dca37544/app/build.gradle#L47).

I found a similar issue in the ProGuard bug tracker:

* [#711 VerifyError: Bad access to protected data in getfield
](https://sourceforge.net/p/proguard/bugs/711/)

I tried upgrading ProGuard to 6.2 (using an overwrite in build.gradle), to no effect.
I also know Fenix uses r8, which is a "newer proguard" or extends it? I actually have no idea how these two interact with each other. 

Anyway, I couldn't find a way to solve this differently:

* I dumped the content of `app/build/intermediates/proguard-files/proguard-android-optimize.txt-3.5.0` (which I _hope_ is actually the proguard file in use) into the local proguard file
* In a second commit I removed the `-allowaccessmodification` line

My build now works fine and Glean is happily uploading pings again.

cc @Dexterp37, @mozilla/glean 

 
### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -44,7 +44,7 @@ android {
     def releaseTemplate = {
         shrinkResources true
         minifyEnabled true
-        proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
+        proguardFiles 'proguard-android-optimize-3.5.0-modified.txt', 'proguard-rules.pro'
         matchingFallbacks = ['release'] // Use on the "release" build type in dependencies (AARs)
     }
 
--- a/app/proguard-android-optimize-3.5.0-modified.txt
+++ b/app/proguard-android-optimize-3.5.0-modified.txt
@@ -20,7 +20,6 @@
 # thoroughly if you go this route.
 -optimizations !code/simplification/arithmetic,!code/simplification/cast,!field/*,!class/merging/*
 -optimizationpasses 5
--allowaccessmodification
 
 -dontusemixedcaseclassnames
 -dontskipnonpubliclibraryclasses
--- a/app/proguard-android-optimize-3.5.0-modified.txt
+++ b/app/proguard-android-optimize-3.5.0-modified.txt
@@ -1,0 +1,121 @@
+####################################################################################################
+# Copy of `proguard-android-optimize.txt`.
+# We need to _remove_ some of the configuration, so the only way to achieve that is to dump it here
+# and modify it.
+####################################################################################################
+
+# This is a configuration file for ProGuard.
+# http://proguard.sourceforge.net/index.html#manual/usage.html
+#
+# Starting with version 2.2 of the Android plugin for Gradle, this file is distributed together with
+# the plugin and unpacked at build-time. The files in $ANDROID_HOME are no longer maintained and
+# will be ignored by new version of the Android plugin for Gradle.
+
+# Optimizations: If you don't want to optimize, use the proguard-android.txt configuration file
+# instead of this one, which turns off the optimization flags.
+# Adding optimization introduces certain risks, since for example not all optimizations performed by
+# ProGuard works on all versions of Dalvik.  The following flags turn off various optimizations
+# known to have issues, but the list may not be complete or up to date. (The "arithmetic"
+# optimization can be used if you are only targeting Android 2.0 or later.)  Make sure you test
+# thoroughly if you go this route.
+-optimizations !code/simplification/arithmetic,!code/simplification/cast,!field/*,!class/merging/*
+-optimizationpasses 5
+-allowaccessmodification
+
+-dontusemixedcaseclassnames
+-dontskipnonpubliclibraryclasses
+-verbose
+
+# Preserve some attributes that may be required for reflection.
+-keepattributes *Annotation*,Signature,InnerClasses,EnclosingMethod
+
+-keep public class com.google.vending.licensing.ILicensingService
+-keep public class com.android.vending.licensing.ILicensingService
+-keep public class com.google.android.vending.licensing.ILicensingService
+-dontnote com.android.vending.licensing.ILicensingService
+-dontnote com.google.vending.licensing.ILicensingService
+-dontnote com.google.android.vending.licensing.ILicensingService
+
+# For native methods, see http://proguard.sourceforge.net/manual/examples.html#native
+-keepclasseswithmembernames class * {
+    native <methods>;
+}
+
+# Keep setters in Views so that animations can still work.
+-keepclassmembers public class * extends android.view.View {
+    void set*(***);
+    *** get*();
+}
+
+# We want to keep methods in Activity that could be used in the XML attribute onClick.
+-keepclassmembers class * extends android.app.Activity {
+    public void *(android.view.View);
+}
+
+# For enumeration classes, see http://proguard.sourceforge.net/manual/examples.html#enumerations
+-keepclassmembers enum * {
+    public static **[] values();
+    public static ** valueOf(java.lang.String);
+}
+
+-keepclassmembers class * implements android.os.Parcelable {
+    public static final ** CREATOR;
+}
+
+-keepclassmembers class **.R$* {
+    public static <fields>;
+}
+
+# Preserve annotated Javascript interface methods.
+-keepclassmembers class * {
+    @android.webkit.JavascriptInterface <methods>;
+}
+
+# The support libraries contains references to newer platform versions.
+# Don't warn about those in case this app is linking against an older
+# platform version. We know about them, and they are safe.
+-dontnote android.support.**
+-dontnote androidx.**
+-dontwarn android.support.**
+-dontwarn androidx.**
+
+# This class is deprecated, but remains for backward compatibility.
+-dontwarn android.util.FloatMath
+
+# Understand the @Keep support annotation.
+-keep class android.support.annotation.Keep
+-keep class androidx.annotation.Keep
+
+-keep @android.support.annotation.Keep class * {*;}
+-keep @androidx.annotation.Keep class * {*;}
+
+-keepclasseswithmembers class * {
+    @android.support.annotation.Keep <methods>;
+}
+
+-keepclasseswithmembers class * {
+    @androidx.annotation.Keep <methods>;
+}
+
+-keepclasseswithmembers class * {
+    @android.support.annotation.Keep <fields>;
+}
+
+-keepclasseswithmembers class * {
+    @androidx.annotation.Keep <fields>;
+}
+
+-keepclasseswithmembers class * {
+    @android.support.annotation.Keep <init>(...);
+}
+
+-keepclasseswithmembers class * {
+    @androidx.annotation.Keep <init>(...);
+}
+
+# These classes are duplicated between android.jar and org.apache.http.legacy.jar.
+-dontnote org.apache.http.**
+-dontnote android.net.http.**
+
+# These classes are duplicated between android.jar and core-lambda-stubs.jar.
+-dontnote java.lang.invoke.**


badboy:
> Hm, duplicating `proguard-android-optimize.txt` is not super nice, but I am not a proguard (or R8) expert either. I wonder if there's a way to only disable this rule in an override or limit the exclusion of the rule to a specific part of the code?

Obviously I'm not a fan either, especially because that way we're also loosing any new configurations that might be applied in the `proguard-android-optimize.txt` file.

I did not find any way to overwrite/disable rules. (The only part where this seems possible seems `-optimizations`, where you can disable them by prefixing `!`)

mcomella:
> I also know Fenix uses r8, which is a "newer proguard" or extends it? I actually have no idea how these two interact with each other. 

My understanding is Google's R8 is a drop-in replacement for ProGuard (which is a non-Google tool) and is enabled by default in current versions of Android.

As for possible solutions to this problem, I can think of:
- we're running R8 in `fullMode` ([docs](https://developer.android.com/studio/build/shrink-code#full-mode)) which I think is experimental. Since it's experimental, I'm a little skeptical of using it. Are you able to reproduce this issue if you set `fullMode` to false? ([code](https://searchfox.org/mozilla-mobile/rev/ccba1b7379e0cd0d0e581775b50ee47664a90b8a/fenix/gradle.properties#19))
- I think the version of R8 is tied to the Android Gradle Plugin version: perhaps there's a newer version of the plugin & R8 that addresses this issue?

Otherwise, I don't think I have enough knowledge of proguard/R8 to come up with a solution off the top of my head.

@badboy Do you happen to know what `-allowaccessmodification` does to the code?

I didn't dig in too deeply – let me know if you want me to do that to better understand the performance implications.

> It would be nice if we could measure whether disabling this has an impact on anything.

The best we can probably do is measure cold startup – which runs through a large variety of code types – but it's pretty noisy so it might be difficult to spot a change here.

badboy:
The docs about `-allowaccessmodification` say:

> Specifies that the access modifiers of classes and class members may be broadened during processing. This can improve the results of the optimization step. For instance, when inlining a public getter, it may be necessary to make the accessed field public too.

Which in theory shouldn't become a problem (access is broadened, not narrowed), still it fixes the issue (I suspect it's a bug in how it plays with other optimizations).

> we're running R8 in `fullMode` ([docs](https://developer.android.com/studio/build/shrink-code#full-mode)) which I think is experimental. Since it's experimental, I'm a little skeptical of using it. Are you able to reproduce this issue if you set `fullMode` to false? ([code](https://searchfox.org/mozilla-mobile/rev/ccba1b7379e0cd0d0e581775b50ee47664a90b8a/fenix/gradle.properties#19))

I _think_ I tried that (we had issues with `fullMode` and Glean in the past so it was the first knob I tried to tune). For good measure I'm going to run another test just to be sure.

> I think the version of R8 is tied to the Android Gradle Plugin version: perhaps there's a newer version of the plugin & R8 that addresses this issue?

I did try to upgrade to no success.

> I didn't dig in too deeply – let me know if you want me to do that to better understand the performance implications.

That's for Fenix devs to figure out. I have no clue about performance measurements on Fenix and how much of an impact this one setting has.

mcomella:
@badboy In the ProGuard bug you linked in comment 0, a comment mentions:

> If we protect (-keep) the entire class the issues goes away, but we'd like to understand the root cause here.

Rather than disable this optimization, could we similarly add a rule to prevent proguard from running on this class? I think the rule would look something like:
```
-keep class mozilla.telemetry.glean.net.FfiPingUploadTask { *; }
```

Based on the pattern I see [here](https://searchfox.org/mozilla-mobile/rev/fd32c1b030795f42f68da92c36e9bdef97b52d29/fenix/app/proguard-rules.pro#18). There may be a way to be more specific (e.g. to only keep the field that has a problem) but this is a good first step and it may not be worth the effort to figure out the syntax to be more specific.

If an approach like this doesn't work, I can look more into disabling this optimization and the performance implications.

badboy:
I tried running with `r8.fullMode = false` to no success. Same error.

I added the additional proguard rule with no success either. Same error.
(And I would have been surprised, we already have a more general rule in the Glean consumer file: `-keep class mozilla.telemetry.** { *; }`).

mcomella:
Sorry that this has been taking so long. Unfortunately I ran out of time today and I'm OOO tomorrow so I won't get to this until Monday.

@csadilek Do you happen to have any time to investigate this?

fwiw, this is the code being added by this PR: https://github.com/mozilla/glean/blob/m16/ping-uploading/glean-core/android/src/main/java/mozilla/telemetry/glean/net/Upload.kt

mcomella:
I'm trying to build Glean locally so I can see the error and fiddle around with it but I've run into a build issue: https://bugzilla.mozilla.org/show_bug.cgi?id=1637429 I was told the associated PR merged to Glean master so I should just be able to build glean locally with a local a-c into a local fenix https://github.com/mozilla/glean/pull/872

Note to self: I was suggested to look at these docs to build glean locally with a local a-c:
- https://mozilla.github.io/glean/book/dev/howtos/development-with-android-components.html
- https://mozilla.github.io/glean/book/dev/howtos/locally-published-components-in-fenix.html

badboy:
I just released Glean v30.0.0 that contains the new code.
It requires some changes in a-c to build, which I pushed here: https://github.com/mozilla-mobile/android-components/compare/master...badboy:with-glean-upload

(My build's still running locally, but that worked before, I'll update if needed)

If that's enough it should allow you to build a-c locally and use that in Fenix.
In the meantime I will also write some notes/fix our docs on how to build Glean in a-c locally.

badboy:
spoke to soon, my local build currently fails with above code. I'll have that fixed in about an hour hopefully

badboy:
Ok, the code in the `with-glean-upload` branch now compiles.

mcomella:
@badboy I checked out your a-c branch and built fenix (372bb421) with the local a-c: I'm unable to reproduce on device (geckoNightlyFennecNightly) and on my API 28 x86 emulator (geckoNightlyFenixNightly). Is there anything you can help me do to reproduce?

I'm also curious: can you reproduce the issue if you do a clean fenix build? I wonder if there is something incorrect cached in the build.

I'll try to start my day at 8am PT tomorrow and focus on this first so we get some overlap.

badboy:
A clean build on my other machine, that definitely didn't build Fenix before, leads to the same error in the emulator (x86, API 28)

mcomella:
I didn't realize this was a log statement and not an app crash: I am able to reproduce.

I also am able to reproduce in the glean sample app, which should make iteration faster, with the following local changes:
```
diff --git a/samples/android/app/build.gradle b/samples/android/app/build.gradle
index 8d5d8653..686de95f 100644
--- a/samples/android/app/build.gradle
+++ b/samples/android/app/build.gradle
@@ -25,8 +25,9 @@ android {

     buildTypes {
         release {
-            minifyEnabled false
-            proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
+            minifyEnabled true
+            proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
+            signingConfig signingConfigs.debug
         }
     }
 }
```

mcomella:
> I also am able to reproduce in the glean sample app

It's worth mentioning that if this is reproducible in all apps, we should either update the glean documentation to mention this as a requirement or figure out a way to define it in the glean library so consumers don't have to.

If we're going to disable a global optimization, I'd like to file a bug upstream somewhere as well so perhaps one day we can disable it.

---

I'm just going to document some of the debugging steps I've been taking.

Looking at r8 debug usage output ([add `-printusage <path>` to proguard-rules.pro](https://developer.android.com/studio/build/shrink-code#usage)), I can confirm that none of the classes involved seem to be in R8's optimization list such as:
- `FfiPingUploadTask*`
- `com.sun.jna.Structure*`

The rules to block out most jna & glean telemetry classes seem to be working.

---

In the sample app I used `proguard-android` instead of `proguard-android-optimize` and it worked fine. I then manually added `-allowaccessmodification` to my proguard rules and it broke. As such, I think we can be fairly confident it's this optimization, in isolation from the others, that causes the issue.

mcomella:
My next thoughts on debugging are:
- See if we can disable `-allowaccessmodification` in a certain area instead of globally
- Try proguard instead of R8: is this a proguard issue or an R8 issue? If the latter, we could consider using proguard instead but there may be a performance penalty
- Is there any change to the code we can make that would trick this optimization into working correctly?

Assuming we can't fix it with a local disable, we should measure the performance implications of this change to determine if it's a go or no go.

mcomella:
I switched to using ProGuard:
```
+++ b/gradle.properties
@@ -21,3 +21,6 @@ libVcsUrl=https://github.com/mozilla/glean.git

 libLicense=MPL-2.0
 libLicenseUrl=https://www.mozilla.org/en-US/MPL/2.0/
+
+android.enableR8=false

+++ b/samples/android/app/proguard-rules.pro
@@ -19,3 +19,8 @@
 # If you keep the line number information, uncomment this to
 # hide the original source file name.
 #-renamesourcefileattribute SourceFile
+
+
+-ignorewarnings
```

(in addition to upgrading the android gradle plugin and gradle but I think these are unrelated) and I'm unable to reproduce the error: **I think this is a bug in R8 with the `-allowaccessmodification` optimization.** We should file this upstream.

---

This leads us with:
- Can we disable the optimization locally?
- If not, can we change the code to work around the bug?
- If not, do we want to use proguard instead of disabling the optimization globally?

And performance analyses.

mcomella:
I tried a bunch of things that didn't work:
- Add `@Keep` to related classes
- Adding unrelated functionality to `ByValue` (so it wasn't just a one-liner)
- Adding various flags to proguard (dontoptimize, keepattributes, ...)
- Didn't see anything related on the troubleshooting page https://www.guardsquare.com/en/products/proguard/manual/troubleshooting
- I moved ByValue to not be an inner class
- Skimmed through all proguard flags ([official cheat sheet](https://www.guardsquare.com/en/products/proguard/manual/refcard)) looking for alternatives

I'm forced to lightly conclude there is no way to disable the optimization locally nor do I have any insights into how we can change the code to work around the bug (@badboy do you have any ideas? I assume you've thought of this already).

Relatedly, I found [the R8 issue tracker](https://issuetracker.google.com/issues?q=componentid:326788) but couldn't find any related issues on there (maybe they've already been closed?).

One additional thing I'd like to try is using the pre-release versions of the Android Gradle Plugin, which may contain newer versions of R8 that have addressed this bug. After that, we're down to:
- Decide whether to use R8 or ProGuard
- Take performance measurements

badboy:
Thanks for tackling this.

> I'm forced to lightly conclude there is no way to disable the optimization locally nor do I have any insights into how we can change the code to work around the bug (@badboy do you have any ideas? I assume you've thought of this already).

Yup, that's the same conclusion I came to.

> If not, can we change the code to work around the bug?

We have some changes planned and I see if we can speed development up to at least test them this week.

mcomella:
> One additional thing I'd like to try is using the pre-release versions of the Android Gradle Plugin, which may contain newer versions of R8 that have addressed this bug.

I upgraded the Android Gradle Plugin to `4.1.0-alpha09` and the problem still exists (note: this is two versions ahead of the current version so it'll probably be a while before any fix comes back to us). Here's my PR (which required some hard-coded fixes to get around the breaking changes in the Gradle version upgrade) if someone wants to play around with it: https://github.com/mcomella/glean/tree/host-10183-allowaccess

That leaves us with:
- Seeing if badboy can change the code to make the error go away
- If not, decide whether to use R8 (w/ optimization disabled) or ProGuard (w/ normal optimizations; PG is deprecated)
- Take performance measurements for changes

@badboy Does it make sense for me to wait on your changes before investigating the latter two options?

---

@badboy Would you mind filing a bug against R8 upstream for our problem? I wonder if the Glean Sample App would be a good reproducible test case for them. I believe this is their issue tracker: https://issuetracker.google.com/issues?q=componentid:326788

badboy:
We have the code changed, I'll get A-C/Fenix builds running today, so I can report back.

I can also take care of filing the bug with Google then

badboy:
Tested with the code from this PR: https://github.com/mozilla/glean/pull/891
Essentially moving from "pass by value" to a "pass by reference" (actually just an "out" pointer we write to).

Same error though, upload is failing as the upload worker task is throwing an exception

<details>

```
       WM-WorkerWrapper  E  Work [ id=c8d361d0-5b95-47ec-96cb-ed42d4d9cc7a, tags={ mozilla.telemetry.glean.scheduler.PingUploadWorker, mozac_service_glean_ping_upload_wo
                            rker } ] failed because it threw an exception/error
                         E  java.util.concurrent.ExecutionException: java.lang.IllegalArgumentException: Invalid Structure field in class mozilla.telemetry.glean.net.Ffi
                            PingUploadTask$ByReference, field name 'activeField' (class com.sun.jna.Structure$StructField): The type "com.sun.jna.Structure$StructField"
                            is not supported: Native size for type "com.sun.jna.Structure$StructField" is unknown
                         E      at androidx.work.impl.utils.futures.AbstractFuture.getDoneValue(AbstractFuture.java:4)
                         E      at androidx.work.impl.utils.futures.AbstractFuture.get(AbstractFuture.java:48)
                         E      at androidx.work.impl.WorkerWrapper$2.run(WorkerWrapper.java:1)
                         E      at androidx.work.impl.utils.SerialExecutor$Task.run(SerialExecutor.java:1)
                         E      at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
                         E      at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
                         E      at java.lang.Thread.run(Thread.java:764)
                         E  Caused by: java.lang.IllegalArgumentException: Invalid Structure field in class mozilla.telemetry.glean.net.FfiPingUploadTask$ByReference, fi
                            eld name 'activeField' (class com.sun.jna.Structure$StructField): The type "com.sun.jna.Structure$StructField" is not supported: Native size
                            for type "com.sun.jna.Structure$StructField" is unknown
                         E      at com.sun.jna.Structure.validateField(Structure.java:8)
                         E      at com.sun.jna.Structure.validateFields(Structure.java:3)
                         E      at com.sun.jna.Structure.<init>(Structure.java:15)
                         E      at com.sun.jna.Structure.<init>(Structure.java:6)
                         E      at com.sun.jna.Structure.<init>(Structure.java:3)
                         E      at com.sun.jna.Structure.<init>(Structure.java:1)
                         E      at com.sun.jna.Union.<init>(Union.java:1)
                         E      at mozilla.telemetry.glean.net.FfiPingUploadTask.<init>(Upload.kt:3)
                         E      at mozilla.telemetry.glean.net.FfiPingUploadTask.<init>(Upload.kt:2)
                         E      at mozilla.telemetry.glean.net.FfiPingUploadTask$ByReference.<init>(Upload.kt:1)
                         E      at mozilla.telemetry.glean.scheduler.PingUploadWorker.doWork(PingUploadWorker.kt:1)
                         E      at androidx.work.Worker$1.run(Worker.java:1)
                         E      ... 3 more
                         E  Caused by: java.lang.IllegalArgumentException: The type "com.sun.jna.Structure$StructField" is not supported: Native size for type "com.sun.j
                            na.Structure$StructField" is unknown
                         E      at com.sun.jna.Native.getNativeSize(Native.java:11)
                         E      at com.sun.jna.Structure.getNativeSize(Structure.java:2)
                         E      at com.sun.jna.Structure.getNativeSize(Structure.java:1)
                         E      at com.sun.jna.Structure.validateField(Structure.java:6)
                         E      ... 14 more
```

</details>

mcomella:
I ran FNPRMS HANOOB startup duration benchmarks locally on my GS5, fennec-nightly 25 runs each (which I've never done before so I don't know how noisy the results are) and I got the following:

R8 without `-allowaccessmodification` optimization:
- 1.460ms
- 1.430ms
- 1.422ms

R8 with `-allowaccessmodification` optimization:
- 1.399ms
- 1.405ms
- 1.379ms

So this change appears to create a ~43ms regression in startup on a fairly low-end device that may be similar to the Moto G5. It may create similar slow downs elsewhere in the app. Considering the wins we find in startup are generally in the 10ms - 20ms range on the Moto G5, this isn't trivial. I'll check with the perf team about this.

mcomella:
I discovered my measurements were flawed – in my most recent run, my device only appears to have kept the last 8 runs rather than 25 – these benchmarks probably have around that many runs in them which is subject to a lot of noise.

That being said, I switched to ProGuard and took some measurements:
- 1.303499s
- 1.314625s
- 1.403249s

That's ~54ms better than we got with R8. We might be better off switching to ProGuard because it seems like R8 may actually be slower than ProGuard.

I'll try to retake my measurements without the issues I previously mentioned.

---

I did speak with my team and they mentioned that a ~43ms regression is a lot so we should try to avoid it if possible.

mcomella:
I retook the numbers with the full run count (25). In all but the last run, I experienced an unexpected 10s outlier for the first run so I removed them (i.e. 24 runs accounted for). Between each run, I cleared data and manually cleared the onboarding.

ProGuard
- 1.3422916666666669s
- 1.321833333333333s
- average: 1.3320625s

R8 w/ optim
- 1.39975s
- 1.4179583333333337s
- average: 1.408854166666667s

R8 w/ optim disabled
- 1.4241666666666666s
- 1.41756s (no outlier = 25 runs)
- average: 1.420863333333333s

That means:
- Switching to ProGuard improves hanoob startup performance by ~77ms
- Disabling the optimization regresses hanoob startup performance by ~12ms

Caveats & notes:
- This is on a Galaxy S5 which we don't fully understand the perf characteristics of
- We may see different behavior on other devices, especially high-end ones
- I had to modify the FNPRMS code to take logcat after each run rather than dumping it all at the end because my device didn't have a large enough logcat buffer
- We normally do 100 runs to avoid noise but to save time I did 25 in 2 iterations
- I wanted to ensure the results I got were not biased by heat throttling (because the slowest results came last) so I re-ran R8 w/ optim disabled after 20-30 minutes of idle device time: it was 1.4472 so this results seem to hold

---

I want to talk to the team about switching to ProGuard as the solution to this problem. We may want to consider talking with R8 folks to understand why our performance is worse with that solution.

mcomella:
I got numbers on the P2 (25 iterations, 1 run):
- ProGuard 0.55628
- R8 0.5448

And the Moto G5 (25 iterations, 1 run):
- Original (so R8):701 ms
- ProGuard: 712ms

**ProGuard is actually slower**. Given that these devices are more recent, this probably means R8 is at least better on newer devices.

---

My gut says we shouldn't revert to ProGuard and I think the team would agree. That leaves us with deciding:
- Can we make changes to the code to prevent the bug? OR
- Remove the optimization (~10ms regression on GS5 on startup, which is the slowest of the devices tested)

@badboy Have you tried making changes to the code that, while maybe not the best long term, might prevent the regression in the product? We would leave it in temporarily. For example, it sounds like making something public that shouldn't be public might prevent the issue (given what the optimization is supposed to do). I would prefer that but I'm looking at this from a performance perspective 😅 

I think we're more okay with taking a 10ms regression though we should do thorough measurements (i.e. 100 runs) on our reference devices to ensure that's accurate: the G5 and the P2.

I think **the next steps are**:
- Seeing if @badboy has an wiggle room to prevent the performance regression by making changes to the code
- If not, taking further benchmarks to make sure that it's okay to make this change

---

Sorry for the short notice 😓 but I'll be out of office until Tuesday. I'll ask @MarcLeclair to look after this bug for me while I'm gone. I uploaded [builds to this folder](https://drive.google.com/drive/folders/1Rwgrb0br7VRFDm3KBZTvvnfK7vsxLz9T) if MarcLeclair needs to take benchmarks.

badboy:
> @badboy Have you tried making changes to the code that, while maybe not the best long term, might prevent the regression in the product?

We had 2 bigger changes landed, none of which prevent this issue.
I will try again with making stuff public temporarily, though as it fails on some JNA internals I don't have super high hopes.

badboy:
I spent all morning on trying to change our code around, including making classes and attributes public (or not), explictly triggering certain code paths in JNA, trying to read the offending field and also reading the JNA code itself, upgrading JNA, ...

Nothing worked, I still get the same exception that the native size for an internal field is unknown. 
This btw is the code raising the exception: https://github.com/java-native-access/jna/blob/5.2.0/src/com/sun/jna/Native.java#L1394-L1395

I'm not even sure why it tries to read the size of that field, it has nothing to do with the memory layout of the data being read.

mcomella:
I think our only option is to land with the optimization disabled then and take the performance hit. I just need to review this to ensure the change was made correctly (e.g. we're pulling the right optimization file from android) and when it lands we'll monitor our startup tests for regressions.

Unfortunately, I didn't leave enough time today to review this – I will do so tomorrow.

mcomella:
@badboy Can you link to the R8 bug filed with Google?

badboy:
Sure. Filed it now here: https://issuetracker.google.com/issues/157535162 (couldn't file it directly in the r8 component, I guess that's up to them to move it)

app/proguard-rules.pro
```diff
@@ -1,5 +1,128 @@
 -dontobfuscate
 
+####################################################################################################
+# Copy of `proguard-android-optimize.txt`.
+# We need to _remove_ some of the configuration, so the only way to achieve that is to dump it here
+# and modify it.
```
mcomella, line 6: Let's be explicit and add some info about:
- the modifications we made, what issue number they were made in, and (in this case) link to the R8 bug so we know when this can be removed
- the version of this file we're using (e.g. file name for me locally is proguard-android-optimize.txt-3.5.0)

app/build.gradle
```diff
@@ -44,7 +44,7 @@ android {
     def releaseTemplate = {
         shrinkResources true
         minifyEnabled true
-        proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
+        proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
```
mcomella, line 47: I would assume `proguard-android.txt` and `proguard-android-optimize.txt` don't both need to be used – is there any specific reason you chose to do that? Also, I'd prefer not to merge this file with our proguard config. What if we did:
```
proguardFiles 'proguard-android-optimize-3.5.0-modified.txt', 'proguard-rules.pro'
```

Instead?

badboy: That makes much more sense indeed.

badboy:
I simplified the approach as requested (and properly documented it in the commit message).

Additionally I got feedback on the r8 bug report already.
The bug is known (I somehow missed the earlier bug report) and a fix landed recently (like 3 weeks ago). 
An r8 release including this fix is expected next week (I assume, given that this week is nearly over).

I'm about to give it a try run with the pre-release to verify.

Depending on that we have 2 options:
1. pull this in now (maybe after verifying above numbers?) and land the Glean upgrade as well
	* we planned to time this to get the longest amount of test time before it rolls out fully, but I don't know the current Fenix schedule.
2. Wait for the release of r8, upgrade that (it would be a bigger jump, probably requiring additional performance tests to verify it doesn't cause regressions)

badboy:
Ran against r8 2.1.28-dev locally (after some other build failure first) and it seems it fixed the issue indeed.

It also looks like the release just got tagged: https://r8-review.googlesource.com/c/r8/+/51755

mcomella:
> Ran against r8 2.1.28-dev locally (after some other build failure first) and it seems it fixed the issue indeed.
> 
> It also looks like the release just got tagged: https://r8-review.googlesource.com/c/r8/+/51755

Good to know – as we discussed, the timing isn't great to update the shrinker so let's disable the optimization and file a follow-up to remove that & update the shrinker.

mcomella:
I filed https://github.com/mozilla-mobile/fenix/issues/11047 as a follow-up to use the new R8 version.

mcomella:
I got a timing of 1.3535833333333336s with this change – looks to me.

mcomella:
Thank you!

PR #10209, opened 2020-04-25
Fixes #4840 - Adds dark search widget based on system theme

### LINKED ISSUE #4840: Search widget is not dark in dark mode
How to reproduce:
- Set Fenix to dark mode.
- Add the Fenix search widget to the home screen

Actual result: a search bar with a white background is added to the home screen 

Expected result: a search bar with a dark background should be added (to be consistent with the dark home screen)

shahsurajk:
### Pull Request checklist

### After merge

1. click on Show All Checks,

## Screenshots: 
* System theme dark, fenix theme light
image

* System theme light, fenix theme light
image

### DIFF
--- a/app/src/main/java/org/mozilla/gecko/search/SearchWidgetProvider.kt
+++ b/app/src/main/java/org/mozilla/gecko/search/SearchWidgetProvider.kt
@@ -10,6 +10,7 @@ import android.appwidget.AppWidgetManager.OPTION_APPWIDGET_MIN_WIDTH
 import android.appwidget.AppWidgetProvider
 import android.content.Context
 import android.content.Intent
+import android.content.res.Configuration
 import android.os.Build
 import android.os.Bundle
 import android.speech.RecognizerIntent
@@ -52,7 +53,7 @@ class SearchWidgetProvider : AppWidgetProvider() {
             // It's not enough to just hide the microphone on the "small" sized widget due to its design.
             // The "small" widget needs a complete redesign, meaning it needs a new layout file.
             val showMic = (voiceSearchIntent != null)
-            val layout = getLayout(layoutSize, showMic)
+            val layout = getLayout(context, layoutSize, showMic)
             val text = getText(layoutSize, context)
 
             val views = createRemoteViews(context, layout, textSearchIntent, voiceSearchIntent, text)
@@ -72,7 +73,7 @@ class SearchWidgetProvider : AppWidgetProvider() {
         val currentWidth = appWidgetManager.getAppWidgetOptions(appWidgetId).getInt(OPTION_APPWIDGET_MIN_WIDTH)
         val layoutSize = getLayoutSize(currentWidth)
         val showMic = (voiceSearchIntent != null)
-        val layout = getLayout(layoutSize, showMic)
+        val layout = getLayout(context, layoutSize, showMic)
         val text = getText(layoutSize, context)
 
         val views = createRemoteViews(context, layout, textSearchIntent, voiceSearchIntent, text)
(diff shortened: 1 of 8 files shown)


shahsurajk:
Auto changing the widget theme based on configuration changes (system wide theme change) is not done. Have little clue on how it would be done. 

Checked the `CONFIGURATION_CHANGED` broadcast receiver, based on the dark theme docs found here, but not sure how it would be done. 

ref: https://developer.android.com/guide/topics/ui/look-and-feel/darktheme#configuration_changes

ekager:
@sblatz could you take a look at this and decide if this is too risky to land before release?

liuche:
hey @sblatz, could you take a look at this PR? we're past code freeze now, so this should be less risky to land.

Mugurell:
This looks like a great addition to the app!
I can help reviewing.

Mugurell:
@shahsurajk Thank you for working on this and sorry for not reviewing sooner.
Since there are some changes needed and also 2 PRs open for the same ticket I just want to make sure you are still interested into updating this.
The changes I'm thinking of would simply involve using the `night` qualifier for the new dark themes layouts and so avoid modifying the widget provider with if/else and then focusing on the best approach for automatic refresh when the system theme changes.

shahsurajk:
I would be interested in doing the same. :)

Mugurell:
Awesome, thank you @shahsurajk !

Well, as I've said I think we'd really benefit from just having the new resources in a new `layout-night` resource directory.
Also in https://github.com/mozilla-mobile/fenix/pull/11575 I saw many more files touched, do we have all the needed resources here?

Regarding actually changing the layout, we refresh the widget at each 60 minutes.
For theme changes in-between this interval we should be able to use something like
>         sendBroadcast(Intent(this, SearchWidgetProvider::class.java).apply {
>             action = AppWidgetManager.ACTION_APPWIDGET_UPDATE
>             putExtra(
>                 AppWidgetManager.EXTRA_APPWIDGET_IDS,
>                 AppWidgetManager.getInstance(applicationContext).getAppWidgetIds(
>                     ComponentName(applicationContext, SearchWidgetProvider::class.java)
>                 )
>             )
>         })

from [application's onConfigurationChanged()](https://github.com/mozilla-mobile/fenix/blob/c9f8986d2b9d0452fd654ba9abbf1979fdee3403/app/src/main/java/org/mozilla/fenix/FenixApplication.kt#L424) after we verify that the uiMode has changed.
^^ This guarantees that while the process is alive we'll be able to update the layout. If the app is not running, I think there is very little we can do, especially with Oreo's background limitations, will checkup with Bram.

Mugurell:
To not have duplicated code and also simplify things maybe that `sendBroadcast...` block of code could be put in a method inside `SearchWidgetProvider`'s companion object?

shahsurajk:
@Mugurell there will still be changes to the SearchWidgetProvider as we resolve the attributes not via night mode directly, but using 
`ThemeManager.resolveAttribute`. Thus we'll have to have changes in the SearchWidgetProvider class.

Mugurell:
> To not have duplicated code and also simplify things maybe that `sendBroadcast...` block of code could be put in a method inside `SearchWidgetProvider`'s companion object?

Regarding this...
Based on https://github.com/mozilla-mobile/fenix/issues/4840#issuecomment-527452868 we don't need to update the widget when the user set's a different theme inside the app.
We should always follow the system theme.

So we probably need to broadcast to update the widget only in the `onConfigurationChanged` callback in `FenixApplication`.
Then, with the resources having the same name but in a different by qualifier resource directory the widget should automagically update to the needed ui.

shahsurajk:
@Mugurell the widget updates even if you resize it or at multiple other instances. Thus, it will re-draw itself at multiple points and thus pick up the night resources  if the app is in night mode and not the system. 

if you wish it to only follow system's behaviour them i think we'll have to do it programmatically, the way we've done in this PR.

shahsurajk:
@Mugurell, could you give me an update on the problem i  mentioned, which is that if we chose doing it via `uiMode`, it will even refresh for app theme changes.

Mugurell:
> @Mugurell the widget updates even if you resize it or at multiple other instances. Thus, it will re-draw itself at multiple points and thus pick up the night resources if the app is in night mode and not the system.
> 
> if you wish it to only follow system's behaviour them i think we'll have to do it programmatically, the way we've done in this PR.

I tested on a Pixel3 with Android 10 and on a Samsung S7 with Android 8 and saw that using qualified resources do get us the intended result: widget following the system theme, not the app theme.
(On the Samsung I did not find the Battery Saver system setting to test the dark layout but the widget was always using the light theme, even when the app was set to use the dark theme).

Have you seen a different behavior on other devices?

shahsurajk:
@Mugurell  my bad, it works. was looking at the wrong place since the start :)

shahsurajk:
Closes in favor of #13975

PR #10230, opened 2020-04-27
Open in Firefox Preview" from Custom Tab should open tab, not home page.

cnevinc:
This PR is not ready to land. I'd like to seek some early advice before I refine it.

I can't find a good place to place the logic for Intents from `BrowserToolbarController` to 
 `HomeActivity` [here](https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarController.kt#L259)

If I follow the contract in [TabIntentProcessor](https://github.com/mozilla-mobile/android-components/blob/945d1a4727c78efd0a9be2143e6bcc1c72b2226c/components/feature/intent/src/main/java/mozilla/components/feature/intent/processing/TabIntentProcessor.kt#L50) and  set data Uri in `openInFenixIntent` intent,  it will create an additional session.
```
    private fun processViewIntent(intent: SafeIntent): Boolean {
        val url = intent.dataString

        return if (url.isNullOrEmpty()) {
            false
        } else {
            val session = createSession(url, private = isPrivate, source = Source.ACTION_VIEW)
             // ----- a new session. will be created.-------
            loadUrlUseCase(url, session, LoadUrlFlags.external())
            true
        }
    }
```
If I want to follow the logic in [OpenBrowserIntentProcessor](https://github.com/mozilla-mobile/fenix/blob/9f154dc3a0584c8f5753c729a015374536c0c9fd/app/src/main/java/org/mozilla/fenix/home/intent/OpenBrowserIntentProcessor.kt#L24), and put `OPEN_TO_BROWSER` in `openInFenixIntent` intent,
```
override fun process(intent: Intent, navController: NavController, out: Intent): Boolean {
       // ----- I can't follow the contract here...continue reading for the reason.... -----
        return if (intent.extras?.getBoolean(HomeActivity.OPEN_TO_BROWSER) == true) {
            out.putExtra(HomeActivity.OPEN_TO_BROWSER, false)

            activity.openToBrowser(BrowserDirection.FromGlobal, getIntentSessionId(intent.toSafeIntent()))
            true
        } else {
            false
        }
    }
```
It's not possible because this flag will be clear by [IntentRecieverActivity](https://github.com/mozilla-mobile/fenix/blob/c38cfae69d01c7b2d757f9f752d4abbf359d0142/app/src/main/java/org/mozilla/fenix/IntentReceiverActivity.kt#L45)  due to there isn't a right `IntentProcessor` for it so it was treated as "OTHERS" `IntentProcessorType`
```
    suspend fun processIntent(intent: Intent) {
        val processor = getIntentProcessors().firstOrNull { it.process(intent) }
        // -----above will return null, thus below will be `OTHERS`-----
        val intentProcessorType = components.intentProcessors.getType(processor)

        launch(intent, intentProcessorType)
    }

    private fun launch(intent: Intent, intentProcessorType: IntentProcessorType) {
        intent.setClassName(applicationContext, intentProcessorType.activityClassName)

        // ----- and later `OPEN_TO_BROWSER` will be set to false again because shouldOpenToBrowser will return false for "OTHERS" IntentProcessorType -----
        intent.putExtra(HomeActivity.OPEN_TO_BROWSER, intentProcessorType.shouldOpenToBrowser(intent))

        startActivity(intent)
        finish() 
    }
```

So when the time `HomeActivity` [wants to process it](https://github.com/mozilla-mobile/fenix/blob/c38cfae69d01c7b2d757f9f752d4abbf359d0142/app/src/main/java/org/mozilla/fenix/HomeActivity.kt#L140) this flag will still be false
```
    override fun process(intent: Intent, navController: NavController, out: Intent): Boolean {
       // ----- `OPEN_TO_BROWSER` is false so this will return false.-----
        return if (intent.extras?.getBoolean(HomeActivity.OPEN_TO_BROWSER) == true) {
            out.putExtra(HomeActivity.OPEN_TO_BROWSER, false)

            activity.openToBrowser(BrowserDirection.FromGlobal, getIntentSessionId(intent.toSafeIntent()))
            true
        } else {
            false
        }
    }
```

In my two cents, it seems to make sense to have both `HomeIntentProcessor`(in Fenix only) and `IntentProcessor`(from feature-intent) for different purposes of communication.
But the interaction will make it hard. Avoid using the same Intent Action, or Intent Data as logic for `process` seems to be unavoidable.  Since there's already a contract for intent processing(`IntentProcessor`from feature-intent), we should extend it and have the centralized control there. That's why I create another `BrowserToolbarIntentProcessor` and implement `IntentProcessor` instead of `HomeIntentProcessor`

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
+++ b/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
@@ -435,6 +435,7 @@ open class HomeActivity : LocaleAwareAppCompatActivity() {
         const val OPEN_TO_BROWSER_AND_LOAD = "open_to_browser_and_load"
         const val OPEN_TO_SEARCH = "open_to_search"
         const val PRIVATE_BROWSING_MODE = "private_browsing_mode"
+        const val BROWSER_TOOLBAR_MODE = "browser_toolbar_mode"
         const val EXTRA_DELETE_PRIVATE_TABS = "notification_delete_and_open"
         const val EXTRA_OPENED_FROM_NOTIFICATION = "notification_open"
         const val delay = 5000L
--- a/app/src/main/java/org/mozilla/fenix/IntentReceiverActivity.kt
+++ b/app/src/main/java/org/mozilla/fenix/IntentReceiverActivity.kt
@@ -69,7 +69,8 @@ class IntentReceiverActivity : Activity() {
             intent.putExtra(HomeActivity.PRIVATE_BROWSING_MODE, false)
             listOf(
                 components.intentProcessors.customTabIntentProcessor,
-                components.intentProcessors.intentProcessor
+                components.intentProcessors.intentProcessor,
+                components.intentProcessors.browserToolbarIntentProcessor
             )
         }
 
--- a/app/src/main/java/org/mozilla/fenix/components/IntentProcessorType.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/IntentProcessorType.kt
@@ -11,7 +11,7 @@ import org.mozilla.fenix.customtabs.ExternalAppBrowserActivity
 import org.mozilla.fenix.migration.MigrationProgressActivity
 
 enum class IntentProcessorType {
-    EXTERNAL_APP, NEW_TAB, MIGRATION, OTHER;
+    EXTERNAL_APP, NEW_TAB, MIGRATION, BROWSER_TOOLBAR, OTHER;
 
     /**
      * The destination activity based on this intent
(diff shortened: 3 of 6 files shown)


ekager:
CC @pocmo I know you changed some of this in #8823 do you have any thoughts here?

cnevinc:
> Oh interesting. I have to dig a bit into the code to understand what is going on and why that may broke with the intent processor refactoring (I do not see why it should right now).
> 
Here's my finding:
Previously  at line 71, if `it.process(intent)` return false, `intentProcessorType` at line 73 will not be null cause `it.matches(intent)` at line 74 will still return true.

image
now with the new approach introduced in #8823, `matches(intent)` in A-C is merged to `process(intent)`  and only `process(intent)` at line 42 will be called. So `intentProcessorType` at line 43 will be null. 
image

https://github.com/mozilla-mobile/fenix/commit/a477f14042150af806dd4a766571cbc0161f3020#diff-842c3d4ac654e221b378bcb5855d8cc1L74

pocmo:
Thank you for all the hints and figuring out why this broke, Nevin. That was super helpful in understanding what went wrong. I created PR #10917 as an alternative solution: I think we should allow intents to define that they want to switch to the browser and not leave that only to the intent processors. That may be helpful in other situations too.

Issue #9880, opened 2020-04-13
[Bug] sync settings: checkboxes are on the wrong side


cadeyrn:
### Steps to reproduce

0. make sure you're connected to Sync
1. open the sync settings

### Expected behavior

The checkboxes are on the left side, as on every other screen.

### Actual behavior

The checkboxes are on the right side.

![Screenshot_20200413-142438](https://user-images.githubusercontent.com/1100614/79120234-bb579d80-7d92-11ea-9f42-4d0c97240341.jpg)

While you're on the screen you could fix the large gap on the left side. No other setting screen has such a large gap on the left side.

### Device information

* Android device: OnePlus 7T Pro McLaren Edition / Oxygen OS 10.0.8 (Android 10)
* Fenix version: master branch revision 52857b09d70c398fc1a5af81d0e271ec1b736601

cadeyrn:
I can fix this. As you can see with my fix there is more spacing between the checkboxes and I don't really like this large spacing but it's consistent because it's exactly the same spacing as on other screens like "Delete browsing data" and "Delete browsing data on quit". If you think it's okay I can create a PR.

![before-after](https://user-images.githubusercontent.com/1100614/79502784-68027b00-8030-11ea-9ebb-880939f5e7f8.png)

brampitoyo:
@cadeyrn Making the spacing consistent, and making every checkbox sits on the left side, is absolutely the right call to make here. Thanks!

sv-ohorvath:
VErified as fixed on Beta 5.0.

PR #10251, opened 2020-04-27
For #9880 - fixed sync settings style

cadeyrn:
I also had to change checkbox_left_preference.xml, otherwise the checkboxes were truncated in RTL.

Before:

before

After:

after


ekager:
Thanks! Opened as https://github.com/mozilla-mobile/fenix/pull/10259 because of the CI issues

Issue #8725, opened 2020-02-25
Intermittently(?) failing test: org.mozilla.fenix.share.ShareControllerTest


pocmo:
```
[task 2020-02-25T12:36:33.813Z] SUITE: org.mozilla.fenix.share.ShareControllerTest
[task 2020-02-25T12:36:33.813Z]   TEST: handleShareToApp should start a new sharing activity and close this
[task 2020-02-25T12:36:33.813Z]     [Robolectric] org.mozilla.fenix.share.ShareControllerTest.handleShareToApp should start a new sharing activity and close this: sdk=28; resources=LEGACY
[task 2020-02-25T12:36:33.813Z]     [Robolectric] NOTICE: legacy resources mode is deprecated; see http://robolectric.org/migrating/#migrating-to-40
[task 2020-02-25T12:36:33.913Z]     I/AssetManager: no overlays.list file found
[task 2020-02-25T12:36:33.913Z]     I/MonitoringInstr: Instrumentation started!
[task 2020-02-25T12:36:33.913Z]     I/MonitoringInstr: Setting context classloader to 'org.robolectric.internal.AndroidSandbox$SdkSandboxClassLoader@54673886', Original: 'org.robolectric.internal.AndroidSandbox$SdkSandboxClassLoader@54673886'
[task 2020-02-25T12:36:33.913Z]     D [FenixApplication] Initializing Glean (uploadEnabled=true, isFennec=false)
[task 2020-02-25T12:36:33.913Z]     E/glean/Glean: Glean should not be initialized multiple times
[task 2020-02-25T12:36:37.013Z]   FAILURE
[task 2020-02-25T12:36:37.013Z] 
[task 2020-02-25T12:36:37.014Z] java.lang.AssertionError: Verification failed: calls are not in verification order
[task 2020-02-25T12:36:37.014Z] 
[task 2020-02-25T12:36:37.014Z] Matchers: 
[task 2020-02-25T12:36:37.014Z] RecentAppsStorage(#65).updateRecentApp(eq(package)))
[task 2020-02-25T12:36:37.014Z] Activity(#71).startActivity(eq(Intent { act=android.intent.action.SEND typ=text/plain flg=0x10000000 cmp=package/activity (has extras) })))
[task 2020-02-25T12:36:37.014Z] Function1(#64).invoke(eq(SUCCESS)))
[task 2020-02-25T12:36:37.014Z] 
[task 2020-02-25T12:36:37.014Z] Calls:
[...]

PR #10269, opened 2020-04-28
Update Android Components version to 40.0.20200428130031.

jonalmeida:
1. click on Show All Checks,


jonalmeida:
test-debug failure is an intermittent that is filed as https://github.com/mozilla-mobile/fenix/issues/8725.

Issue #10196, opened 2020-04-24
FNX-3348 ⁃ [Bug]"Run in private browsing" should remain enabled even if the user disables the Add-on


AndiAJ:
## Steps to reproduce
1. Install any of the featured Add-ons and open it
1. Enable "Run in private browsing"
1. Disable the Add-on
1. Exit to the Add-ons manager
1. Re-open the Add-on and check the state of the "Run in private browsing" option

### Expected behavior
"Run in private browsing" should remain enabled (*As per Firefox on Desktop*)

### Actual behavior
"Run in private browsing" is also disabled

### Device information
* Android device: Google Pixel 3a (Android 10)
* Fenix version: Nightly Build #21150608 from 4/24

### Notes
► Video
![20200424-180656](https://user-images.githubusercontent.com/51314259/80228278-88958b00-8657-11ea-9bea-6ebe24449736.gif)

Related to #6123

Amejia481:
I think we should hide the **"Run in private browsing"** when the add-on is disabled, similar to what we are doing with the **"Settings"** section because when an add-on is disabled we can't apply any changes to **"Run in private browsing"**

Amejia481:
@AndiAJ  @psymoon landed a related patch, would mind retesting? :)

AndiAJ:
Hi @Amejia481 & @psymoon , I've just checked this matter on the latest **`Nightly 200504 from 5/4`** using the following devices: 

• Google Pixel 3a (Android 10)
• Huawei Mate 20 Lite (Android 9)
• OnePlus A3 (Android 6.0.1)

✔️ The "Run in private browsing" is hidden when the add-on is disabled
✔️ The status of the "Run in private browsing" is properly displayed after re-enable 
❌ The "Run in private browsing" can't be used after the user disables and re-enables the Add-on
❓ Noticed sometimes a delay in hiding the "Run in private browsing" after disabling the Add-on

► Video delay in hiding "Run in private browsing"
![20200504-103421-online-video-cut](https://user-images.githubusercontent.com/51314259/80945222-03973800-8df4-11ea-834c-345709aa63a0.gif)


► Video "Run in private browsing" can't be used after the user disables and re-enables the Add-on
![20200504-102325](https://user-images.githubusercontent.com/51314259/80944637-b4043c80-8df2-11ea-9562-75d07e735e3d.gif)


I'll **remove the QA needed label** until further notice.

psymoon:
[...]

Issue #10195, opened 2020-04-24
FNX-3347 ⁃ [Bug] Private browsing setting can re-enable add-on


AndiAJ:
## Prerequisites
Ublock Origin should be installed and enabled
"Run in private browsing" should not be disabled
Switch to Private Browsing 

## Steps to reproduce
1. Navigate to the Add-ons manager and open uBlock
1. Enable "Run in private browsing" and quickly disable uBlock

### Expected behavior
Ublock should be disabled 

### Actual behavior
Ublock is re-enabled automatically

### Device information
* Android device: Google Pixel 3a (Android 10)
* Fenix version: Nightly Build #21150608 from 4/24

### Notes
► Video
![20200424-170116](https://user-images.githubusercontent.com/51314259/80225418-b7a9fd80-8653-11ea-94fe-c5fc15e76f66.gif)

❗ Bumped into an isolated case where it **gets disabled two times**.

## Steps to reproduce
1. Disable "Run in private browsing" 
1. Quickly disable uBlock and exit to the Add-ons manager
1. Open again uBlock

►Video
![ezgif com-video-to-gif](https://user-images.githubusercontent.com/51314259/80225903-62222080-8654-11ea-8e5f-8cc2e6510b9a.gif)

Related to #6123

Amejia481:
@AndiAJ  @psymoon landed a related patch, would mind retesting? :)

AndiAJ:
Hi, **verified as fixed** on the latest **`Nightly 200504 from 5/4`** using the following devices:
• Google Pixel 3a (Android 10)
• Huawei Mate 20 Lite (Android 9)
• OnePlus A3 (Android 6.0.1)

► Video
![20200504-112516](https://user-images.githubusercontent.com/51314259/80948326-8622f600-8dfa-11ea-90b0-d54b63e8ebd7.gif)

PR #10273, opened 2020-04-28
For #10195, #10196: Allow private browsing mode switch to sync with enable state

psymoon:
### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/addons/InstalledAddonDetailsFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/addons/InstalledAddonDetailsFragment.kt
@@ -88,10 +88,12 @@ class InstalledAddonDetailsFragment : Fragment() {
     @SuppressWarnings("LongMethod")
     private fun bindEnableSwitch(view: View) {
         val switch = view.enable_switch
+        val privateBrowsingSwitch = view.allow_in_private_browsing_switch
         switch.setState(addon.isEnabled())
         switch.setOnCheckedChangeListener { v, isChecked ->
             val addonManager = v.context.components.addonManager
             switch.isClickable = false
+            privateBrowsingSwitch.isClickable = false
             view.remove_add_on.isEnabled = false
             if (isChecked) {
                 addonManager.enableAddon(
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/fenix/addons/InstalledAddonDetailsFragment.kt
```diff
@@ -228,13 +239,15 @@ class InstalledAddonDetailsFragment : Fragment() {
                     runIfFragmentIsAttached {
                         this.addon = it
                         switch.isClickable = true
+                        enableSwitch.isClickable = true
                         view.remove_add_on.isEnabled = true
                     }
                 },
                 onError = {
                     switch.isChecked = addon.isAllowedInPrivateBrowsing()
```
csadilek, line 247: This line (not change here) should move into the `runIfFragmentAttached` block just in case.

Amejia481:
> If I `enable PB` -> disable the add-on -> go back to add-ons admin -> go to the details of the add-on -> enable the add-on -> **then PB is disable but it should be enable** see the link below:
> https://drive.google.com/file/d/1PkFYqU9GCrAmz_4rcTEIObZz__3FYW2U/view?usp=sharing

I think the problem is that when we enter to the add-on details page with an disabled add-on, we don't have access to certain properties like (PB and optionsPageUrl), when we enable the add-on, I think we have to update the PB UI with the latest add-on

app/src/main/java/org/mozilla/fenix/addons/InstalledAddonDetailsFragment.kt
```diff
@@ -100,6 +102,8 @@ class InstalledAddonDetailsFragment : Fragment() {
                         runIfFragmentIsAttached {
                             this.addon = it
                             switch.isClickable = true
+                            privateBrowsingSwitch.isClickable = true
```
Amejia481, line 105: We need to update the PB UI we the new add-on `privateBrowsingSwitch.isChecked = it.isAllowedInPrivateBrowsing()`. I will be better to add it before we make it visible this way is not going to blink
See https://github.com/mozilla-mobile/fenix/pull/10273#issuecomment-622408827

psymoon: > > If I `enable PB` -> disable the add-on -> go back to add-ons admin -> go to the details of the add-on -> enable the add-on -> **then PB is disable but it should be enable** see the link below:
> > https://drive.google.com/file/d/1PkFYqU9GCrAmz_4rcTEIObZz__3FYW2U/view?usp=sharing
> 
> I think the problem is that when we enter to the add-on details page with an disabled add-on, we don't have access to certain properties like (PB and optionsPageUrl), when we enable the add-on , get this properties back, I think we have to update the PB UI with the latest add-on. For this reason if you go back to the add-on admin page and re-enter it has the right state.

@Amejia481 I think this is related to https://bugzilla.mozilla.org/show_bug.cgi?id=1634493.
 When we enable/disable, the new web-extension/add-on instance we get back from GV does not persist previously set `allowedInPrivateBrowsing`.

Amejia481: @psymoon when I added `privateBrowsingSwitch.isChecked = it.isAllowedInPrivateBrowsing()` the issue goes away, did you try it?

Issue #10285, opened 2020-04-29
FNX-3406 ⁃ [Bug]Common myth hyperlink and other links in app provide confusing description and action info for screen readers


mcarare:
Follow-up for #7887.

We have several links in app that are announced as a button would be.
This prevents screen readers users to have a complete and correct info regarding the element type and available action for element.

Proposed solution: 
<strike>Announce action as "_Follow link_" ("Double tap to follow link") and Textview type as "_Link_".</strike>
Edit : Should be announced as "_link_", without double tap and long press action.

mcarare:
@asadotzler any thoughts on this issue or the proposed solution? TY!

eeejay:
Hi! I don't know what gets posted where. But I added an implementation suggestion here:
https://github.com/mcarare/fenix/commit/eb0828c2e765804f0b7324bec99f3cc252463533#r38963440

sflorean:
Verified as fixed on Nightly 7/24 and Beta 79.0.0 with Samsung Galaxy Note 10 (Android 10).

Issue #182, opened 2019-01-16
FNX-5265 ⁃ [UX] Quick Action Bar


lime124:
Update 3/21 - Need to remove Screenshots from QAB and Library and notify engineering

**User Story:**
As a user I want easy access to common functionality for using the information I find on the web.

**Current Design Status:**

* In Good Shape
* Foundational Design Done
* Needs more design detail work

**Design Priority:**
P1

**Mockups:**
https://mozilla.invisionapp.com/share/Y5Q1NM7VMN9#/341653120_Quck_Action_Bar

shorlander:
**User Story:**
As a user I want easy access to common functionality for using the information I find on the web.

**Current Design Status:**
- In Good Shape
- Foundational Design Done
- Needs more design detail work

**Design Priority:**
P1

**Mockups:**
https://mozilla.invisionapp.com/share/Y5Q1NM7VMN9#/341653120_Quck_Action_Bar

bbinto:
This is only the animation and UI, no functionality for screenshots etc needed.

@colintheshots can this be pulled into milestone 1?

sblatz:
Eitan mentioned an accessibility note (for when we implement this--just don't have an engineering ticket right now):

The custom gesture we use to activate the quick action menu will require some extra work to support TalkBack since all gestures are redirected when it is enabled.

boek:
@colintheshots Do we still want to talk about a single actionsheet?

colintheshots:
Yes, I like the idea of a single actionsheet. I have mixed feelings about leaving half of the buttons revealed with a tiny drawer handle. If it were me, I'd make the handle bigger, use a touch delegate to make it bigger still, and bounce the sheet briefly so the user knows the buttons are there. Affordances aren't usually easy or obvious and I won't pretend to have the answers.

shorlander:
> Yes, I like the idea of a single actionsheet. I have mixed feelings about leaving half of the buttons revealed with a tiny drawer handle. If it were me, I'd make the handle bigger, use a touch delegate to make it bigger still, and bounce the sheet briefly so the user knows the buttons are there. Affordances aren't usually easy or obvious and I won't pretend to have the answers.

@boek @colintheshots do you mean having the browser toolbar and the quick actions on the same sheet?

boek:
@shorlander Yep! I I think the main concern with how many touch targets there would be at the bottom of the phone. I have no problem problem implementing the current design so we can test it out and iterate based on feedback :)

[...]

PR #10288, opened 2020-04-29
For #10285: Improve a11y for links in app.

mcarare:
### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/utils/LinkTextView.kt
+++ b/app/src/main/java/org/mozilla/fenix/utils/LinkTextView.kt
@@ -1,0 +1,34 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.utils
+
+import android.content.Context
+import android.util.AttributeSet
+import android.view.accessibility.AccessibilityNodeInfo
+import android.view.accessibility.AccessibilityNodeInfo.AccessibilityAction.ACTION_LONG_CLICK
+import androidx.appcompat.widget.AppCompatTextView
+import org.mozilla.fenix.R
+
+/**
+ * An [AppCompatTextView] that announces as link in screen readers for a11y purposes
+ */
+class LinkTextView @JvmOverloads constructor(
+    context: Context,
+    attrs: AttributeSet? = null,
+    defStyleAttr: Int = 0
+) : AppCompatTextView(context, attrs, defStyleAttr) {
+
+    override fun onInitializeAccessibilityNodeInfo(info: AccessibilityNodeInfo?) {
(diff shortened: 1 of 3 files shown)


sblatz:
@MReschenberg please take a look and make sure this is a reasonable thing to announce with Screen Reader :)

MReschenberg:
Solution looks good to me, checking with eitan to make sure this is a reasonable string.

mcarare:
@eeejay I have previously investigated the possibility of implementing something similar to the code you've linked. Unfortunately, we cannot implement the `roleDescription` in the same manner, because AccessibilityNodeInfo for Views does not have the same info as in your example. (`getExtra` return an empty Bundle, we do not have access to GeckoBundle). So, I can only add the "Link" as a `roleDescription` in the extra bundle for AccessibilityNodeInfo (already implemented the change), but it still has do be localized.

As to the action description, I felt that " double tap to activate" is not relevant info for a link, as it is the same announcement for any other view that has a click listener (buttons, images, etc). That is why I implemented the "to follow link" part.
Also, I have removed the long click announcement, as there is in fact no action for long click.

eeejay:
> @eeejay I have previously investigated the possibility of implementing something similar to the code you've linked. Unfortunately, we cannot implement the `roleDescription` in the same manner, because AccessibilityNodeInfo for Views does not have the same info as in your example. (`getExtra` returns an empty Bundle, we do not have access to GeckoBundle). So, I can only add the "Link" as a `roleDescription` in the extra bundle for AccessibilityNodeInfo (already implemented the change), but it still has to be localized.

The GeckoBundle type you see there is unrelated. It is just how we pass things via JNI from native gecko. The string value of key `AccessibilityNodeInfo.roleDescription` needs to simply be a lowercase `link`. You might be right about the localization. But either way, that is less to translate and less to customize. The click action can be left alone.

ekager:
@sblatz looks like this has been updated - could you rereview?

sblatz:
@mcarare are there more places we can apply this while doing this update? :)

mcarare:
@sblatz Yes, there are more "learn more" links in the app. I did not work on changing them also until I had the go ahead with this solution.

sblatz:
This looks good to me. @MReschenberg or @eeejay can you confirm from the accessibility side?

liuche:
@mcarare looks like this has approval from both a11y and a reviewer, so feel free to rebase and then land it!

mcarare:
Re-running for unrelated test-debug failure: 

`[task 2020-06-26T05:29:37.446Z]   TEST: GIVEN collection WHEN renameCollection is called THEN collection should be renamed

[task 2020-06-26T05:29:37.546Z]   FAILURE
[task 2020-06-26T05:29:37.547Z] 

[task 2020-06-26T05:29:37.548Z] java.lang.AssertionError: Verification failed: call 2 of 3: TabCollectionStorage(tabCollectionStorage#180).renameCollection(eq(TabCollection(#182)), eq(name))) was not called`

PR #10293, opened 2020-04-29

MickeyMoz:


ekager:
Closing because we are going to wait for the next dot release because a schema was updated and this version will cause crashes.

PR #10330, opened 2020-04-30
Revert "Update Android Components version"

gabrielluong:
Reverts mozilla-mobile/fenix#10320


gabrielluong:
This PR won't help with the db migration error with top sites. We also don't want to revert all the way back to AC v39. We should ideally just wait for the fix in https://github.com/mozilla-mobile/android-components/pull/6839 to land and get a new cut of AC.

PR #10361, opened 2020-05-01
DO NOT LAND: measure performance of edit bookmark menu

mcomella:
This code is never intended to be landed. This PR is opened solely for code review of methodology for measuring:

> GIVEN Fenix on 2018 (low-end|high-end) Android reference
AND the Context Menu is open
AND the current website is bookmarked
WHEN the bookmark button is actioned
THEN the edit bookmark UI renders in less than 250ms

This PR also needs https://github.com/mozilla-mobile/android-components/pull/6845 to work.

N.B.: intermittently I find the `onDraw` method is called twice before the listener is removed. I believe it's because our UI thread code doesn't get scheduled before drawing two frames. It seems inconsequential (only a few ms in between) so I just take the later measurement.

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/library/bookmarks/edit/EditBookmarkFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/library/bookmarks/edit/EditBookmarkFragment.kt
@@ -6,10 +6,14 @@ package org.mozilla.fenix.library.bookmarks.edit
 
 import android.content.DialogInterface
 import android.os.Bundle
+import android.os.Handler
+import android.os.Looper
+import android.os.SystemClock
 import android.view.Menu
 import android.view.MenuInflater
 import android.view.MenuItem
 import android.view.View
+import android.view.ViewTreeObserver
 import androidx.appcompat.app.AlertDialog
 import androidx.appcompat.app.AppCompatActivity
 import androidx.appcompat.widget.Toolbar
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/fenix/library/bookmarks/edit/EditBookmarkFragment.kt
```diff
@@ -108,6 +114,27 @@ class EditBookmarkFragment : Fragment(R.layout.fragment_edit_bookmark) {
...
+
+        override fun onDraw() {
+            // I verified that this frame is the one that draws the EditBookmark by adding thread.sleep
+            // into uiHandler.post.
+            uiHandler.post {
+                view.viewTreeObserver.removeOnDrawListener(this)
+            }
+
+            val now = SystemClock.elapsedRealtime()
+            android.util.Log.e("lol", "displayed ${now - MenuButtonClick.initialMillis}")
```
MarcLeclair, line 128: NIT: Not sure if we should leave the log there 😄

mcomella: We're not landing this code so I think it's fine.

liuche:
@mcomella @MarcLeclair is this something we can just close and you can reference the closed PR? If not, please reopen.

PR #10397, opened 2020-05-04

MickeyMoz:


Amejia481:
Tested locally and the app still load websites!

Issue #10253, opened 2020-04-22
FNX-3389 ⁃ Progressive Web App pop-up interferes with collection of visual metrics -- please provide a method to disable


acreskeyMoz:
When loading a site that supports PWA, e.g., https://www.instagram.com, the "Add to Home screen" pop-up is displayed. (attached below).

Unfortunately this interferes with various performance metrics that measure when content is presented to the user. (e.g. SpeedIndex, ContentfulSpeedIndex, etc).

Please provide a method so that this behaviour may be disabled for performance work.
A user preference, if possible, would be ideal.

![](https://user-images.githubusercontent.com/44072237/80024010-6b708900-84ac-11ea-9980-038f97def9a7.png)

jonalmeida:
Moved this back to Fenix based on Tiger's comment above.

st3fan:
I added it to the A-C sprint planning because I think this is a blocker that we can probably quickly resolve.

psymoon:
@acreskeyMoz feel free to re-open this issue if this is still an issue!

acreskeyMoz:
@bdekoz -- no, it's built into Fenix

https://github.com/mozilla-mobile/fenix/commit/5ba9cc6ec9ab729b75f37c79ca5c568013892b9b

Is it not working?
The performancetest mode of fenix has some requirements -- device attached by usb, adb connection:

https://github.com/mozilla-mobile/fenix/blob/82facbfa00b4a701b6919d618df3ee6459bc74b7/app/src/main/java/org/mozilla/fenix/perf/Performance.kt#L42-L56

PR #10442, opened 2020-05-06
For #10253: Add disableFirstTimePWAPopup for performance testing

psymoon:


MarcLeclair:
@acreskeyMoz  do we want to runt he intent separately from the other intent? I.e `-ez performancetest true -ez disableFirstTimePWA true` or all-in-one ?

acreskeyMoz:
@MarcLeclair I think an "all-in-one" solution, `-ez performancetest true` for performance tests would be the best

Issue #10434, opened 2020-05-06
FNX-4283 ⁃ [Bug] getProcessStartTimeTicks crash


liuche:
Crash in `fenix.perf` with `getProcessStartTimeTicks` in 5.0.0 release build.

Sentry log: https://sentry.prod.mozaws.net/operations/fenix/issues/8293980/

liuche:
@mcomella is this something from perf team monitoring?

mcarare:
@liuche This seems to be a problem specific to using a privacy manager that relies on a hooking mechanism (SandHook) and it blocks reading the file in [Stat.kt](https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/java/org/mozilla/fenix/perf/Stat.kt#L28). The most we can do is catch the exception or maybe @mcomella has another idea.
 I do not think this crash will have a high recurrence.

mcomella:
>  The most we can do is catch the exception or maybe @mcomella has another idea.

Yeah, I think we'll have to catch the exception.

This doesn't seem to have a high crash count and does seem to be related to OS customization so it doesn't urgent: I'll wait for triage later today to figure out the priority. This is probably a permacrash though and the fix is straightforward so I think we should address it, despite the low crash count.

liuche:
This is with 1% rollout, but reasons seem good, I'll remove the RB label.

PR #10481, opened 2020-05-07
For #10434: Handle cases when proc/$pid/stat is not accessible.

mcarare:
### Pull Request checklist

### After merge

1. click on Show All Checks,


mcarare:
@mcomella I wasn't sure how we would want to handle exceptions. I thought that a value of 0 will be ignored when assessing this metric. Let me know if this should be handled differently. TY!

app/src/main/java/org/mozilla/fenix/perf/StartupFrameworkStartMeasurement.kt
```diff
@@ -68,7 +68,13 @@ internal class StartupFrameworkStartMeasurement(
         //
         // Similarly, we round app init to a whole integer tick value because process start only
         // comes in integer ticks values.
-        val processStartTicks = stat.getProcessStartTimeTicks(Process.myPid())
+        val processStartTicks =
+            try {
+                stat.getProcessStartTimeTicks(Process.myPid())
+            } catch (e: NegativeArraySizeException) {
```
mcomella, line 74: Why are we catching `NegativeArraySizeException`? iiuc, the error is `FileNotFoundException`: https://sentry.prod.mozaws.net/operations/fenix/issues/8293980/

app/src/main/java/org/mozilla/fenix/perf/Stat.kt
```diff
@@ -25,6 +26,7 @@ private const val FIELD_POS_STARTTIME = 21 // starttime naming matches field in
 open class Stat {
 
     @VisibleForTesting(otherwise = PRIVATE)
+    @Throws(IOException::class)
```
mcomella, line 29: nit: `@Throws` appears to be designed for Java interop only which doesn't really apply to us ([source](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.jvm/-throws/)):

> This annotation indicates what exceptions should be declared by a function when compiled to a JVM method.

I'd change this to a kdoc comment instead, i.e.:
```kotlin
/**
 * @throws [java.io.FileNotFoundException]
 */
```

I'd also use the most specific exception possible. We should also document this on the public facing method, `getProcessStartTimeTicks`

app/src/main/java/org/mozilla/fenix/perf/StartupFrameworkStartMeasurement.kt
```diff
@@ -68,7 +68,13 @@ internal class StartupFrameworkStartMeasurement(
...
+                return 0
```
mcomella, line 75: Rather than returning 0, I think it's preferable to report an error through a new telemetry probe.

That being said, if that's too much work, it's probably fine to catch the exception in `setExpensiveMetric` and record nothing if it's thrown. Just bump the metric version and change the docs to make it explicit this is happening.

mcomella:
Sorry it took so long to get to this 😞

mcomella:
Ideally, we could extend the tests to cover this case as well (I think `Stat` & friends have high code coverage).

app/src/main/java/org/mozilla/fenix/perf/StartupFrameworkStartMeasurement.kt
mcarare: Oh, sorry, I think I merged the wrong branch, I was testing throwing random exceptions on another.

app/src/main/java/org/mozilla/fenix/perf/Stat.kt
mcarare: I was looking in the code that  explicitly throws this exception, and I thought maybe another privacy manager would choose to throw another `IOException`. I will use the most specific we have for now, but if in the future it shows another, we should change it to more general _IOException_.

mcarare:
> Sorry it took so long to get to this 😞

No worries. I know it happens. Thank you for the review!

app/src/main/java/org/mozilla/fenix/perf/StartupFrameworkStartMeasurement.kt
mcarare: @mcomella I added a telemetry probe, going to also post data review form and mark this PR with needs:data-review. Let me know if any more changes are needed. TY!

mcarare:

1) What questions will you answer with this data?

- How often a custom OS will block us from reading stats needed for startup time evaluation?

2) Why does Mozilla need to answer these questions?  Are there benefits for users? Do we need this information to address product or business requirements? Some example responses:

- Measure how often our startup time implementation is not suitable for OS customizations.

3) What alternative methods did you consider to answer these questions? Why were they not sufficient?

- We could not gather data locally on a wide range of devices and OS customizations that are not suitable for our startup measurement implementation.

4) Can current instrumentation answer these questions?

No.

5) List all proposed measurements and indicate the category of data collection for each measurement, using the [Firefox data collection categories](https://wiki.mozilla.org/Firefox/Data_Collection) found on the Mozilla wiki.   

**Note that the data steward reviewing your request will characterize your data collection based on the highest (and most sensitive) category.**

<table>
  <tr>
    <td>Measurement Description</td>
    <td>Data Collection Category</td>
    <td>Tracking Bug #</td>
  </tr>
    <td>framework_start_error: event</td>
    <td>Category 1</td>
    <td>#10434</td>
  </tr>
</table>


6) How long will this data be collected?  Choose one of the following:

* I want this data to be collected until 2020-07-15 (like the rest of startup telemetry data)

7) What populations will you measure?

All.

8) If this data collection is default on, what is the opt-out mechanism for users?

Standard telemetry opt-out.

9) Please provide a general description of how you will analyze this data.

- Watch `frameworkStartReadError` to determine if there are recurrent read errors.

10) Where do you intend to share the results of your analysis?

Internally.

11) Is there a third-party tool (i.e. not Telemetry) that you are proposing to use for this data collection? If so:

No.

app/src/main/java/org/mozilla/fenix/perf/Stat.kt
```diff
@@ -25,6 +25,9 @@ private const val FIELD_POS_STARTTIME = 21 // starttime naming matches field in
 open class Stat {
 
     @VisibleForTesting(otherwise = PRIVATE)
+    /**
+     * @throws [java.io.FileNotFoundException]
+     */
```
mcomella, line 30: nit: I _think_ kdoc is conventionally above the annotation

mcomella, line 30: nit: I'd add a kdoc to `getFrameworkStartNanos` too. It's more important than this method because it's the non-private method of this class

mcomella:
@boek Please data review.

mcomella:
@liuche Please data review.

Issue #9044, opened 2020-03-08
[Bug] Download Completed Notification is too intrusive


s-ankur:
## Steps to reproduce
Download any file
### Expected behavior
Small nonintrusive notification appears. It does not affect browsing in any way and user can continue to do what they were doing before .Notification automatically disabled after a while or after interaction with the webpage. 
### Actual behavior
Large notification and dark screen intruding on usability especially if you plan to download many images continuosly
### Device information

* Android device: Asus Zenfone
* Fenix version: Fenix Nightly

test2a:
Yes. Came here to say this. The popup should fade away say after 2 seconds like it does on ff proper

liuche:
cc-ing @brampitoyo here from UX, if you want to address these. Our snackbars tend to be standardized, so may also just be something we won't plan to address.

brampitoyo:
@s-ankur @test2a I agree with part of your proposal.

Agree. Our dialog is currently blocking you from doing any other action until you dismiss it (tap “x”). We should make this dialog non-blocking. In other words, let’s not show any scrim. This way, you’d be able to select, scroll, long-press, etc. on the site. You can proceed downloading the next file, or perform any other action.

Disagree. We shouldn’t change the form of our dialog. It shouldn’t be a snackbar that fades away after 2 seconds. What if you want to view the image later on, but not right after it finishes downloading? The only way to do this would be to swipe down to the Notification Centre, then select the downloaded file – not impossible, but inconvenient. By keeping the notification as a non-blocking dialogue, you can easily open the file, dismiss it, or continue doing your current task.

Thoughts?

s-ankur:
Yeah I agree. Non blocking is whats really important to me, since it really affects the ability to download things in quick succession. But also I would definitely prefer if the dialog takes up less screen-space. 
Secondly, the dialog disappearing by itself would really be counterproductive, as you point out. What I can suggest is that instead of it being on a timeout, you could have it be dynamic, like the dynamic addressbar that we have now.  At least that is the way chrome does it, and it does feel a lot less intrusive that way.

brampitoyo:
@s-ankur Sounds good. Here’s an updated “Download completed” notification that’s non-interruptive.

<img width="360" src="https://user-images.githubusercontent.com/51007/78515569-7075dd00-780a-11ea-803d-2228144f9e8a.png">

codrut-topliceanu:
[...]

PR #10491, opened 2020-05-07
For #9044 - Replace DownloadNotificationBottomSheetDialog with unobtrusive view

codrut-topliceanu:
For https://github.com/mozilla-mobile/fenix/issues/9044

**Note:** Needs merging **after** https://github.com/mozilla-mobile/android-components/pull/6899
### Pull Request checklist

### After merge

1. click on Show All Checks,

new download notification


Amejia481:
Hi 👋  @codrut-topliceanu 
I was taking a look at the feature it looks great. A minor thing, sometimes when I scroll the toolbar is displayed but the dialog is still hidden [video](https://drive.google.com/file/d/1mtTRAcO8royBGHFLRTzpAULNyvdxaIOe/view?usp=sharing)

codrut-topliceanu:
Hey @Amejia481 , thanks for taking a look at this.
Yeah, i know what you mean. Basically when you stop scrolling the view does a snap either UP or DOWN. And I had to make the snap DOWN just as sensitive as the one for the URL bar otherwise you could end up with the download notification view just floating above a hidden bottom URL bar (if you were on a page and scrolled just a bit so that the bottom URL bar would hide). 

```
          val minHeight: Float =
                if (bottomToolbarHeight > 0) bottomToolbarHeight else child.height.toFloat()

            if (child.translationY >= minHeight / 2) {
                animateSnap(child, SnapDirection.DOWN)
            } else {
                animateSnap(child, SnapDirection.UP)
            }
```

As a consequence the amount of "Scroll" needed to snap DOWN is half the height of the URL toolbar, so very little, but that means that to snap UP needs A LOT more scrolling to trigger. I'm not entirely sure how to fix this to make it more smooth. On the bright side, at least it is very difficult to accidentally snap UP the notification to block your view :P

app/src/main/java/org/mozilla/fenix/downloads/DownloadNotification.kt
```diff
@@ -0,0 +1,143 @@
...
+import org.mozilla.fenix.ext.settings
+
+/**
+ * [DownloadNotification] is used to show a view in the current tab to the user, triggered when
+ *  downloadFeature.onDownloadStopped gets invoked. It uses [DownloadNotificationBehavior] to
+ *  hide when the users scrolls through a website as to not impede his activities. Uses
+ *  [DownloadState.dismissed] to track whether the user has closed the view.
+ * */
+
+class DownloadNotification(
```
Amejia481, line 29: As mentioned on https://github.com/mozilla-mobile/android-components/pull/6899#discussion_r425428064, I think it will be better to rename this class to include the word dialog as `DownloadNotification` can be a source of confusion with the download notifications in the status bar.

codrut-topliceanu:
This patch needs more work (following [the discussion here](https://github.com/mozilla-mobile/android-components/pull/6899)). Converting to draft.

app/src/main/res/layout/download_dialog_layout.xml
```diff
@@ -1,5 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
-<!-- This Source Code Form is subject to the terms of the Mozilla Public
+<?xml version="1.0" encoding="utf-8"?><!-- This Source Code Form is subject to the terms of the Mozilla Public
```
Amejia481, line 1: nit: This should be in separate lines

app/src/main/java/org/mozilla/fenix/downloads/DynamicDownloadDialog.kt
```diff
@@ -0,0 +1,140 @@
...
+        }
+
+        view.download_dialog_close_button.setOnClickListener {
+            dismiss(view)
+        }
+
+        view.download_dialog_filename.text = downloadState.fileName
+    }
+
+    fun expand() {
```
Amejia481, line 126: Should we rename it to `show` instead of `expand`? Show  feels more familiar as it is normally use for dialog to display its content [see](https://developer.android.com/reference/androidx/fragment/app/DialogFragment#show(androidx.fragment.app.FragmentManager,%20java.lang.String))  What do you think?

codrut-topliceanu: Yep, makes sense. I'll rename it.

codrut-topliceanu:
Hey, thanks :D 
I'll take a look at the UI tests.

Amejia481:
@codrut-topliceanu  see https://github.com/mozilla-mobile/fenix/issues/10798#issuecomment-631531412 it's possible that the test `testDownloadPrompt` is failing Intermittently

Amejia481:
Thanks @codrut-topliceanu for all the great work!

PR #10519, opened 2020-05-08

MickeyMoz:


AaronMT:
FTL:

> Hello everyone, we're experiencing issues preparing virtual devices. We're looking into and and we'll keep you updated. In the meantime, Android virtual tests have been paused. Physical devices and iOS seem unaffected

PR #10526, opened 2020-05-08
No issue: Update Flank to v20.05.0 (stable)

AaronMT:
snapshot -> to -> https://github.com/Flank/flank/releases/tag/v20.05.0


AaronMT:
> Do we want to move to specific versions vs. the snapshot now?

I think we can use the new release as a floor for now and continue to check-in often with the Flank team to see if any recent changes are worth taking. We would have to swap back to snapshot and touch the Dockerfile of course. I think for now v20.05.0 is a safe release.

PR #10530, opened 2020-05-08
No issue: Exclude perftests in nightly cron tasks.

gmierz:
The nightly cron tasks are currently failing because the browsertime/raptor/visual-metrics tests are being included in them when they shouldn't be. This patch prevents those tests from making it through.


gmierz:
We're [reverting the offending patch](https://github.com/mozilla-mobile/fenix/pull/10533), these changes will be in the patch next time we attempt to land it.

Issue #10473, opened 2020-05-07
FNX-4310 ⁃ [Bug](Regression) The Add-ons are re-enabled after enabling "Run in private browsing"


AndiAJ:
## Prerequisites
Ublock Origin should be installed and enabled
"Run in private browsing" should not be disabled

## Steps to reproduce
1. Navigate to the Add-ons manager and open uBlock
1. Enable "Run in private browsing" and quickly disable uBlock

### Expected behavior
Ublock should be disabled

### Actual behavior
Ublock is re-enabled automatically

### Device information
* Android device: 
• Google Pixel 3a (Android 10)
• Huawei Mate 20 Lite (Android 9)
• OnePlus A3 (Android 6.0.1)
* Fenix version: Nightly build 200507 from 5/7

### Notes
❗ Regression of #10195 
❗ Other Add-ons are also impacted

► Video
![20200507-104420](https://user-images.githubusercontent.com/51314259/81272386-18730600-9056-11ea-836b-7a56f78862b3.gif)

Noticed while re-checking #10196

AndiAJ:
Hi, **verified as fixed** on the latest **`Nightly Build 200512 from 5/12`** using the following devices:
• Google Pixel 3a (Android 10)
• Huawei Mate 20 Lite (Android 9)
• OnePlus A3 (Android 6.0.1)

► Video
![20200512-102028](https://user-images.githubusercontent.com/51314259/81651970-c1957400-943b-11ea-95e8-915347ae35e6.gif)

PR #10531, opened 2020-05-08
For #10473: Fetch the addon from addon manager instead of the store

psymoon:
This makes the add-on instance reference synchronized between different screen transitions a lot simpler and consistent by utilizing `AddonManager`'s `pendingAddonActions` logic.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/addons/InstalledAddonDetailsFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/addons/InstalledAddonDetailsFragment.kt
@@ -11,17 +11,15 @@ import android.view.ViewGroup
 import android.widget.Switch
 import androidx.core.view.isVisible
 import androidx.fragment.app.Fragment
+import androidx.lifecycle.lifecycleScope
 import androidx.navigation.Navigation
 import androidx.navigation.findNavController
 import kotlinx.android.synthetic.main.fragment_installed_add_on_details.view.*
-import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.cancel
-import kotlinx.coroutines.flow.collect
-import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
 import mozilla.components.feature.addons.Addon
+import mozilla.components.feature.addons.AddonManagerException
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/fenix/addons/InstalledAddonDetailsFragment.kt
```diff
@@ -44,36 +48,42 @@ class InstalledAddonDetailsFragment : Fragment() {
...
-                            val newState = addon.installedState?.copy(enabled = addonState.enabled)
-                            this.addon = addon.copy(installedState = newState)
-                            view.enable_switch.setState(addon.isEnabled())
+    private fun bindAddon(view: View) {
+        lifecycleScope.launch(Dispatchers.IO) {
+            try {
+                val addons = requireContext().components.addonManager.getAddons()
+                lifecycleScope.launch(Dispatchers.Main) {
+                    addons.find { addon.id == it.id }.let {
+                        if (it == null) {
```
psymoon, line 65: May be safer to wrap this in `runIfFragmentIsAttached`? 🤔

app/src/main/java/org/mozilla/fenix/addons/InstalledAddonDetailsFragment.kt
```diff
@@ -44,36 +41,46 @@ class InstalledAddonDetailsFragment : Fragment() {
...
-                            view.enable_switch.setState(addon.isEnabled())
+    private fun bindAddon(view: View) {
+        lifecycleScope.launch(Dispatchers.IO) {
+            try {
+                val addons = requireContext().components.addonManager.getAddons()
+                lifecycleScope.launch(Dispatchers.Main) {
+                    runIfFragmentIsAttached {
+                        addons.find { addon.id == it.id }.let {
+                            if (it == null) {
+                                throw AddonManagerException(NullPointerException())
```
Amejia481, line 60: nit: What about instead of the `NullPointerException`, we could create a `AddonManagerException("Addon ${it.id} not found")` indicating that we couldn't find the addon.

psymoon: @Amejia481 `class AddonManagerException(throwable: Throwable) : Exception(throwable)`, is there a way to pass string input instead?

Issue #3552, opened 2019-06-19
FNX3-17101 ⁃ [Bug] Content crash reports unsubmitted in about:crashes


sv-ohorvath:
## Steps to reproduce
1. Open about:crashcontent.
1. Select the "send crash" checkbox. Restart tab.
    1. Expected behavior
The report is sent to https://crash-stats.mozilla.com
    1. Actual behavior
The report is unsubmitted in about:crashes. Tap submit there and then it will appear in https://crash-stats.mozilla.com

*about:crashparent crashes are submitted automatically.

### Device information
* Android device: Galaxy S8 (Android 9)
* Fenix version: 1.0.0. RC 2

vesta0:
@sarentz FYI

sblatz:
I thought about:crashes was not working properly and we were planning on removing the about page🤔 I think we removed them in a-c.

sv-ohorvath:
The reports from about:crashcontent don't go to https://crash-stats.mozilla.com automatically. That's how I ended up checking about:crashes, to verify if they appear somewhere at least.

sblatz:
This seems critical if users are actually unable to report crashes.

sarentz:
We do get plenty content crashes reported so I think it is just `about:crashcontent` not working.

dbolter:
Note this issue appears to be pertaining to about:crashes, which is useful for triaging dogfooder crashes. @snorp is this a GV bug?

snorp:
`about:crashes` is not accurate. We should probably just remove it.

snorp:
I should add: crash reporting is done at the app level, so we can't easily make `about:crashes` work. Some other `about:crashes`-like system in a-c would be great.

st3fan:
Why don't we just turn this into something in the Settings? That is also much more discoverable. Does mean we need to write some UI though, so this is not trivial and should be planned as a new feature.

pocmo:
A replacement for `about:crashes` landed in Fenix and is available via "Settings > About > Crashes". For now it also opens if you type `about:crashes` into the toolbar (but not for other ways of navigating to that URL).

Issue #2295, opened 2019-05-06
FNX-89 ⁃ Support about:crashes for Fenix


agi90:
### Why/User Benefit/User Problem
It's really helpful to have a crash report attached to a bug report and not all users know how or want to get stacktraces from logcat (or even if they do they might not be near a laptop when the crash happens).

This also helps engineers make sure that their crashes are being reported correctly to the crash reporter.

### What/Requirements
Have a page (preferably `about:crashes`) that displays the list of crashes reported for the current profile with a link to the crash report.

vesta0:
Thanks @agi90 ! This would be great, does it require GV work? cc @cpeterso

cpeterso:
GV does have an about:crashes page, but it is incomplete (https://github.com/mozilla-mobile/fenix/issues/1221) because crash uploading (GV native crashes sent to Socorro and Fenix exceptions sent to Sentry) is delegated to the app (Fenix + AC's lib-crash component). We have a GV bug suggesting we disable GV's about:crashes in https://bugzilla.mozilla.org/show_bug.cgi?id=1541152

Fenix can implement its own about:crashes (that includes both GV native crashes and Fenix exceptions) because AC's lib-crash component knows about all crash reports that have been uploaded.

djc:
I hope this will get some priority soon. As a longtime Nightly user, the current experience where I can not easily check if crashes are known and/or track progress of fixing them makes me feel rather powerless and frustrated when crashes occur (and over the past few days Fenix has been supercrashy for me in seemingly very common scenarios).

agi90:
AC Issue: https://github.com/mozilla-mobile/android-components/issues/1705

pocmo:
A replacement for `about:crashes` landed in Fenix and is available via "Settings > About > Crashes". For now it also opens if you type `about:crashes` into the toolbar (but not for other ways of navigating to that URL).

PR #10561, opened 2020-05-11
Integrate activity for showing past crashes.

### LINKED ISSUE #8769: [Bug] "about:crashes" page not doing its job
## Steps to reproduce

I am going to assume that Firefox Preview 3 or Nightly has crashed at least once on your system. Preferably, it has automatically sent a crash report. Open the about:crashes page in the address bar.

### Expected behavior

Judging by the name of the page, you should see a list crashes and/or of (un-) submitted crash reports.

### Actual behavior

You will only see the text "No crash reports have been submitted".

### Device information

* Android device: Nokia 5 (model TA-1024), Android 9.
* Fenix version: all of them, it seems; both Nightly and Preview 3.

pocmo:
The `lib-crash` component now saves information about past crashes to disk and comes with a bit of UI to show this list of crashes. It can directly link to crash reports on Socorro and Sentry which hopefully will make it easier to figure out why the app has crashed and to report crashes.

There are two ways to access the crash list, via "settings > about" and by just typing "about:crashes" into the URL bar (like desktop/Fennec).

Closes #8769, #3552, #2295

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/search/SearchController.kt
+++ b/app/src/main/java/org/mozilla/fenix/search/SearchController.kt
@@ -6,6 +6,7 @@
 package org.mozilla.fenix.search
 
 import android.content.Context
+import android.content.Intent
 import androidx.navigation.NavController
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.delay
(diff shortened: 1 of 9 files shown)


app/src/main/java/org/mozilla/fenix/search/SearchController.kt
```diff
@@ -52,6 +54,14 @@ class DefaultSearchController(
 ) : SearchController {
 
     override fun handleUrlCommitted(url: String) {
+        if (url == "about:crashes") {
+            // The list of past crashes can be accessed via "settings > about", but desktop and
```
mcarare, line 58: There is also the possibility of pasting "about:crashes" with "fill link from clipboard". Do you think we should intercept that also?

pocmo: Good point!

I don't think it is worth the effort at this point. Even this here is something I didn't really intented to do and was just more for convenience.

PR #10612, opened 2020-05-12
Closes #10611 - Fix exit status of Flank

### LINKED ISSUE #10611: FNX-4767 ⁃ [Bug] Fenix UI test reports logging success with failures
I believe starting with the changes we added in https://github.com/mozilla-mobile/fenix/pull/10089 I am seeing Matrix Results report as failure but logged as success for the past couple days since this landed.

I think this is happening due to the following

```Shell
$JAVA_BIN -jar $FLANK_BIN android run \
	--config=$flank_template \
	--max-test-shards=$num_shards \
	--app=$APK_APP --test=$APK_TEST \
	--local-result-dir="${RESULTS_DIR}" \
	--project=$GOOGLE_PROJECT \
	| tee flank.log

exitcode=$?
failure_check
```

I believe the tee call is overwriting the exit check we want from calling Flank just prior :upside_down_face:

AaronMT:
Closes #10611

### DIFF
--- a/automation/taskcluster/androidTest/ui-test.sh
+++ b/automation/taskcluster/androidTest/ui-test.sh
@@ -128,7 +128,7 @@ echo "EXECUTE TEST(S)"
 echo
 # Note that if --local-results-dir is "results", timestamped sub-directory will
 # contain the results. For any other value, the directory itself will have the results.
-$JAVA_BIN -jar $FLANK_BIN android run \
+set -o pipefail && $JAVA_BIN -jar $FLANK_BIN android run \
 	--config=$flank_template \
 	--max-test-shards=$num_shards \
 	--app=$APK_APP --test=$APK_TEST \


AaronMT:
> I'm not sure this will work. The reason why we use a function to set the exit code is that we still want some artifacts to be written regardless of pass or fail outcome. Most of the terminate on fail bash functions will end the script abruptly. I think we still need to capture the return code from the Java execution in a variable and hold onto it til the end. Not sure if pipefail is doing that

Setting pipetail will get us the the exit code of the process that's piped to another. 

E.g, what's happening now
```Bash
$ false | true; echo $?
0
```
and after (see the ui-test-x86 debug failed above showing the right result now)
```Bash
$ set -o pipefail
$ false | true; echo $?
1
```

rpappalax:
> > I'm not sure this will work. The reason why we use a function to set the exit code is that we still want some artifacts to be written regardless of pass or fail outcome. Most of the terminate on fail bash functions will end the script abruptly. I think we still need to capture the return code from the Java execution in a variable and hold onto it til the end. Not sure if pipefail is doing that
> 
> Setting pipefail will get us the the exit code of the process that's piped to another (sets the exit status $? to the exit code of the last program to exit non-zero (or zero if all exited successfully)).
> 
> E.g, what's happening now
> 
> ```shell
> $ false | true; echo $?
> 0
> ```
> 
> and after (see the ui-test-x86 debug failed above showing the right result now)
> 
> ```shell
> $ set -o pipefail
> $ false | true; echo $?
> 1
> ```



> > I'm not sure this will work. The reason why we use a function to set the exit code is that we still want some artifacts to be written regardless of pass or fail outcome. Most of the terminate on fail bash functions will end the script abruptly. I think we still need to capture the return code from the Java execution in a variable and hold onto it til the end. Not sure if pipefail is doing that
> 
> Setting pipefail will get us the the exit code of the process that's piped to another (sets the exit status $? to the exit code of the last program to exit non-zero (or zero if all exited successfully)).
> 
> E.g, what's happening now
> 
> ```shell
> $ false | true; echo $?
> 0
> ```
> 
> and after (see the ui-test-x86 debug failed above showing the right result now)
> 
> ```shell
> $ set -o pipefail
> $ false | true; echo $?
> 1
> ```

per our zoom sesion LGTM 👍

PR #10648, opened 2020-05-14
Bug 1627027 - Use nightly Fenix variant instead of performancetest.

gmierz:
This patch switches the build variant that is used in performance-tests (browsertime, raptor) from the performancetest variant to the nightly variant. It also removes the existing cron jobs for those two frameworks and runs these tests in the nightly cron.


gmierz:
Sounds good, thanks! I've reverted the test changes.

PR #10634, opened 2020-05-13
for #7032 scroll to dismiss keyboard in search fragment

sraturi:
### Pull Request checklist

- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features. In addition, it includes a screenshot of a successful [accessibility scan](https://play.google.com/store/apps/details?id=com.google.android.apps.accessibility.auditor&hl=en_US) to ensure no new defects are added to the product.

### After merge
- [ ] **Milestone**: Make sure issues finished by this pull request are added to the [milestone](https://github.com/mozilla-mobile/fenix/milestones) of the version currently in development.

### To download an APK when reviewing a PR:
1. click on Show All Checks,
2. click Details next to "Taskcluster (pull_request)" after it appears and then finishes with a green checkmark,
3. click on the "Fenix - assemble" task, then click "Run Artifacts".
4. the APK links should be on the left side of the screen, named for each CPU architecture

Issue #7032, opened 2019-12-09
[Bug] Don't allow and Allow buttons cannot be accessed in landscape mode


abodea:
## Steps to reproduce
1. Launch Fenix.
2. Access Private browsing(make sure search suggestions are on).
3. Type anything in the URL bar.
4. The `Allow search suggestions in private session?` prompt is displayed.
5. Rotate the device from portrait to landscape mode.
### Expected behavior
The screen is scrollable and the `Don't allow and Allow` buttons can be accessed.
### Actual behavior
The bottom part of the `Allow search suggestions in private session?` prompt is not displayed and `Don't allow and Allow` buttons cannot be accessed.
### Device information

* Android device: **Google Pixel 3a XL(Android 10), OnePlus 6T(Android 9)**
* Fenix version: **Latest Nightly 12/9.**

![ezgif-6-733c1fa8e9ce](https://user-images.githubusercontent.com/42831109/70442794-ac3ca800-1a9f-11ea-847e-ebd7ac4200c5.gif)

### Design Work

- [x] Request for Comments from UX team
- [ ] ~Usability testing if a new pattern or potentially confusing UI~ Not required – it’s a bugfix
- [x] Request for Comments from Fenix team and/or A/C, GV, Sync, etc as needed
- [x] Interaction details included
- [x] Accessibility ~- is this a review by someone,~ just keeping it in mind?
- [ ] ~For existing UI changes, deltas are highlighted~ No UI – this is a behaviour change
- the[ ] ~Strings/content reviewed and approved~ No string – this is a behaviour change
    - [ ] ~Are they too long for the component in other locales~

### Deliverable

- [ ] ~Assets uploaded to Google Drive~
- [ ] ~Links to deliverables (spec, assets) posted in comment 0~

### Deliverable Links

- ~User flow: (link to e.g. Abstract artboard, LucidChart, Whimsical, Miro, etc.)~
- ~Spec: (link to e.g. Abstract artboards)~
- ~Assets: (link to Google Drive folder)~

### Behaviours

We’ve tried designing new landscape-specific layouts for this dialogue. In the end, the solution that worked without introducing any layout exceptions is to **make the search results scrollable**. This is the behaviour that we should have.

mcarare:
How should we handle this? Hide keyboard when prompt is touched or somehow scroll the prompt ?

brampitoyo:
@mcarare Ideally, we’d align the buttons right, make them the same width, and stack them vertically:

[...]

PR #10698, opened 2020-05-16
Revert "for #7032 scroll to dismiss keyboard in search fragment" 🎲

ekager:
Reverts mozilla-mobile/fenix#10634

I'm seeing the homescreen on current master completely unresponsive with the following STR
1. Type something in Search Fragment
2. Scroll to dismiss keyboard
3. Press back 
4. Try to press toolbar again to do another search

(sometimes it just pops you entirely out of Fenix when you press back too)


darkwing:
Reproduced the error on master, then verified the fix with this PR.  Good spot @ekager !

Issue #9949, opened 2020-04-15
[Library] Possible inconsistencies between deleting history/bookmark items


BMEduard:
After investigating #8648 I have come across some possible discrepancies between deleting history and bookmark items. These are:

1. Should we allow users to be able to undo the deletion of individual and/or multiple history items, as it is the case for bookmarks?
2. Should we also require the user to confirm the deletion of a folder/folders by adding a confirmation pop-up if he uses multiselect to delete multiple bookmark items and one (or more) of them happens to be a folder? 
This would make it consistent with the requested change (in #8648) to add a confirmation pop-up for deleting a folder. As for the confirmation message, it could be something along the lines of “You have selected one or more folders for deletion. Are you sure you want to delete these items?”. This way the user will be unable to delete a folder without having a confirmation pop up first.

@AmyYLee What are your thoughts regarding these two possible unintended inconsistencies?

AmyYLee:
> After investigating #8648 I have come across some possible discrepancies between deleting history and bookmark items. These are:
> 
>     1. Should we allow users to be able to undo the deletion of individual and/or multiple history items, as it is the case for bookmarks?
> 
>     2. Should we also require the user to confirm the deletion of a folder/folders by adding a confirmation pop-up if he uses multiselect to delete multiple bookmark items and one (or more) of them happens to be a folder?
>        This would make it consistent with the requested change (in #8648) to add a confirmation pop-up for deleting a folder. As for the confirmation message, it could be something along the lines of “You have selected one or more folders for deletion. Are you sure you want to delete these items?”. This way the user will be unable to delete a folder without having a confirmation pop up first.
> 
> 
> @AmyYLee What are your thoughts regarding these two possible unintended inconsistencies?

@betsymi @topotropic What are your thoughts for the above?

betsymi:
1. For deleting an individual history item/several history items, I see no issue with a snackbar where you could undo. For both: `History deleted UNDO`
2. Sure, the confirmation dialog can read the following. `[Browser name] will delete the selected items. CANCEL DELETE`
2a. I did this action myself and if possible would like to update the string on the snackbar that appears after to `Deleting selected folders UNDO`

liuche:
We hit string freeze last Friday, so we can't land any more strings (per our agreement with the L10N team, so that they don't have strings coming in at the last minute).

It's unfortunate because this is such a small change, but we need to back this out and land it after July 20, which is our cut for the Release build (and when we can land strings again).

[...]

PR #10738, opened 2020-05-18
For issue #9949 - Inconsistencies between deleting bookmark and history items

BMEduard:
- Added a new check inside the "deleteMulti" method from BookmarkFragment that calls the showRemoveFoldersDialog to prevent the user from being able to delete one or more bookmark folders without being asked for confirmation, as in #8648.

- Added the undo action for deleting individual history items by creating a new field to the history state containing the id's of the history items that are pending for deletion; This field is used inside the update function from the view to show/hide the items. 

- In order to show the correct headers while items are being hidden/deleted, a map was added in the adapter that keeps the position of the headers and updates it as items are pending for deletion/deleted.

- In addition, the state's modes have been reduced to "Normal" and "Editing", since the "Deleting" mode's UI updates (showing and hiding the deletion progress bar) is doesn't overlap the other states.

### Pull Request checklist

- [ X ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ X ] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not
- [ X ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features. In addition, it includes a screenshot of a successful [accessibility scan](https://play.google.com/store/apps/details?id=com.google.android.apps.accessibility.auditor&hl=en_US) to ensure no new defects are added to the product.

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkFragment.kt
@@ -274,13 +274,17 @@ class BookmarkFragment : LibraryPageFragment<BookmarkNode>(), UserInteractionHan
     }
 
     private fun deleteMulti(selected: Set<BookmarkNode>, eventType: Event = Event.RemoveBookmarks) {
+        selected.forEach { if (it.type == BookmarkNodeType.FOLDER) {
+            showRemoveFolderDialog(selected)
+            return
+        } }
         updatePendingBookmarksToDelete(selected)
 
         pendingBookmarkDeletionJob = getDeleteOperation(eventType)
 
         val message = when (eventType) {
             is Event.RemoveBookmarks -> {
-                getRemoveBookmarksSnackBarMessage(selected)
+                getRemoveBookmarksSnackBarMessage(selected, containsFolders = false)
             }
             is Event.RemoveBookmarkFolder,
             is Event.RemoveBookmark -> {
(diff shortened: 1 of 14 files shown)


app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkFragment.kt
```diff
@@ -269,13 +269,17 @@ class BookmarkFragment : LibraryPageFragment<BookmarkNode>(), UserInteractionHan
...
+            return
+        } }
         updatePendingBookmarksToDelete(selected)
 
         pendingBookmarkDeletionJob = getDeleteOperation(eventType)
 
         val message = when (eventType) {
             is Event.RemoveBookmarks -> {
-                getRemoveBookmarksSnackBarMessage(selected)
+                getRemoveBookmarksSnackBarMessage(selected, false)
```
kglazko, line 282: Instead of just setting to 'false', it may be a good idea to create a variable for this boolean so people know what exactly is being set false.

app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkFragment.kt
```diff
@@ -309,29 +320,38 @@ class BookmarkFragment : LibraryPageFragment<BookmarkNode>(), UserInteractionHan
...
                 setNegativeButton(R.string.delete_browsing_data_prompt_cancel) { dialog: DialogInterface, _ ->
                     dialog.cancel()
                 }
                 setPositiveButton(R.string.delete_browsing_data_prompt_allow) { dialog: DialogInterface, _ ->
-                    updatePendingBookmarksToDelete(setOf(selected))
+                    updatePendingBookmarksToDelete(selected)
                     pendingBookmarkDeletionJob = getDeleteOperation(Event.RemoveBookmarkFolder)
                     dialog.dismiss()
-                    val message = getDeleteDialogString(selected)
+                    val snackbarMessage = getRemoveBookmarksSnackBarMessage(selected, true)
```
kglazko, line 348: Same here, except now this same variable would be set to true.

app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkFragment.kt
ekager: I think also just showing the named params in the function call serve the same purpose!

BMEduard: Thank you for the suggestions, I added the named params to the function call, so it should be clear now what the flags are used for.

PR #10789, opened 2020-05-19
Migrate some SessionManager usage to BrowserStore

NotWoods:
### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
+++ b/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
@@ -13,6 +13,7 @@ import androidx.appcompat.app.AppCompatDelegate
 import androidx.core.content.getSystemService
 import kotlinx.coroutines.Deferred
 import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.GlobalScope
 import kotlinx.coroutines.async
 import kotlinx.coroutines.launch
@@ -21,6 +22,7 @@ import mozilla.appservices.Megazord
 import mozilla.components.browser.session.Session
 import mozilla.components.concept.push.PushProcessor
 import mozilla.components.feature.addons.update.GlobalAddonDependencyProvider
+import mozilla.components.lib.crash.CrashReporter
 import mozilla.components.service.glean.Glean
 import mozilla.components.service.glean.config.Configuration
 import mozilla.components.service.glean.net.ConceptFetchHttpUploader
@@ -46,14 +48,12 @@ import org.mozilla.fenix.session.PerformanceActivityLifecycleCallbacks
 import org.mozilla.fenix.session.VisibilityLifecycleCallback
 import org.mozilla.fenix.utils.BrowsersCache
 import org.mozilla.fenix.utils.Settings
-import mozilla.components.lib.crash.CrashReporter
 
 /**
  *The main application class for Fenix. Records data to measure initialization performance.
  *  Installs [CrashReporter], initializes [Glean]  in fenix builds and setup Megazord in the main process.
  */
-@SuppressLint("Registered")
-@Suppress("TooManyFunctions", "LargeClass")
+@Suppress("Registered", "TooManyFunctions", "LargeClass")
 open class FenixApplication : LocaleAwareApplication() {
     init {
         recordOnInit() // DO NOT MOVE ANYTHING ABOVE HERE: the timing of this measurement is critical.
@@ -66,6 +66,7 @@ open class FenixApplication : LocaleAwareApplication() {
     var visibilityLifecycleCallback: VisibilityLifecycleCallback? = null
         private set
 
+    @ExperimentalCoroutinesApi
     override fun onCreate() {
         super.onCreate()
 
@@ -114,6 +115,7 @@ open class FenixApplication : LocaleAwareApplication() {
         Log.addSink(AndroidLogSink())
     }
 
+    @ExperimentalCoroutinesApi
     @CallSuper
     open fun setupInMainProcessOnly() {
         run {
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/mozilla/fenix/session/NotificationSessionObserver.kt
```diff
@@ -5,38 +5,39 @@
...
+import mozilla.components.support.ktx.kotlinx.coroutines.flow.ifChanged
 import org.mozilla.fenix.ext.components
-import org.mozilla.fenix.ext.sessionsOfType
 
 /**
  * This observer starts and stops the service to show a notification
  * indicating that a private tab is open.
  */
-
+@ExperimentalCoroutinesApi
```
jonalmeida, line 22: Not sure if you've considered this already: we have components that use flows and to avoid the annotation from being littered through the codebase, we add [this suppressor](https://github.com/mozilla-mobile/android-components/blob/master/components/feature/syncedtabs/build.gradle#L30-L34) to the `build.gradle`.

NotWoods: fenix's gradle file is a lot more intimidating :P I'll try to find a spot to add the suppressor tomorrow

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -151,7 +150,7 @@ open class FenixApplication : LocaleAwareApplication() {
         visibilityLifecycleCallback = VisibilityLifecycleCallback(getSystemService())
         registerActivityLifecycleCallbacks(visibilityLifecycleCallback)
 
-        components.core.sessionManager.register(NotificationSessionObserver(this))
+        NotificationSessionObserver(this)
```
csadilek, line 153: Hm we're not doing anything with the instance created here, but rely on a side-effect of creating it, which looks strange :). We also don't ever call the `stop` method, as this is scoped to the lifecycle of the app. Should we just create a private method instead with the logic to observe the store? Or if we want to keep the class, add a start method so we get:

```kotlin
val notificationObserver = NotificationSessionObserver(this)
notificationObserver.start()
```

or similar?

app/src/main/java/org/mozilla/fenix/settings/deletebrowsingdata/DeleteBrowsingDataFragment.kt
```diff
@@ -25,23 +28,21 @@ import org.mozilla.fenix.ext.showToolbar
...
 
         controller = DefaultDeleteBrowsingDataController(requireContext())
 
-        sessionObserver = object : SessionManager.Observer {
-            override fun onSessionAdded(session: Session) = updateTabCount()
-            override fun onSessionRemoved(session: Session) = updateTabCount()
-            override fun onSessionSelected(session: Session) = updateTabCount()
-            override fun onSessionsRestored() = updateTabCount()
-            override fun onAllSessionsRemoved() = updateTabCount()
+        scope = requireComponents.core.store.flowScoped(viewLifecycleOwner) { flow ->
```
csadilek, line 40: We usually create the scope in `onStart` and cancel it `onStop`. Mostly curious if you did this differently for a reason? We never cancel the returned scope here though. Should we `scope?.cancel` in `onDestroyView` or switch to `onStart/onStop`?

NotWoods: I'll switch to `onStart/onStop`

app/src/main/java/org/mozilla/fenix/session/NotificationSessionObserver.kt
NotWoods: Do you know what might be causing this error? 
```
Class kotlinx.coroutines.ExperimentalCoroutinesApi is not an experimental API marker annotation
```

csadilek: @NotWoods hm yeah, it seems this is not needed anymore?
```
tasks.withType(org.jetbrains.kotlin.gradle.tasks.AbstractKotlinCompile).all {
    kotlinOptions.freeCompilerArgs += ["-Xuse-experimental=kotlinx.coroutines.ExperimentalCoroutinesApi"]
}
```

NotWoods: For now I'll switch back to adding the annotations manually.

csadilek: 👍

app/src/main/java/org/mozilla/fenix/session/NotificationSessionObserver.kt
```diff
@@ -5,38 +5,46 @@
...
-    }
-
-    override fun onSessionAdded(session: Session) {
-        // Custom tabs are meant to feel like part of the app that opened them, not Fenix, so we
-        // don't need to show a 'close tab' notification for them
-        if (session.private && !session.isCustomTabSession()) {
-            notificationService.start(context)
-        }
+    fun stop() {
+        scope?.cancel()
```
ekager, line 48: Quick Q: Is it a possible problem that this scope cancel is never called AFAICT? It's not lifecycle aware and we are manually starting it in `FenixApplication` cc @NotWoods @csadilek

NotWoods: Should be fine since we want the notification to stick around

csadilek: We have a few store observers / scopes tied to the whole application lifecycle e.g. `MigrationTelemetryListener`, `MigrationPushRenewer`. Those never get stopped and don't need to. I guess we could've not assigned the scope and removed the `stop` method, but yes, it should not be a problem.

csadilek: @ekager @NotWoods Actually, we do have a problem here after all! The `NotificationSessionObserver` is leaking the `HomeActivity` as it's initialized in `HomeActivity.onCreate` and uses it as a context. When the `HomeActivitiy` is re-created, we never cancel the old scope and therefore leak the old `HomeActivity`. You can reproduce this by enabling "Don't keep activities"...then leak canary gets pretty mad at us :).

@NotWoods I saw you were working on an uplift of this class so we can fix it there? Let's chat in our PR...

PR #10806, opened 2020-05-20
Update Kotlin dependency

NotWoods:
After updating I needed to suppress some experimental API warnings with `OptIn`.


app/build.gradle
```diff
@@ -600,6 +600,12 @@ dependencies {
     lintChecks project(":mozilla-lint-rules")
 }
 
+tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
+    kotlinOptions.freeCompilerArgs += [
+        "-Xuse-experimental=kotlinx.coroutines.ExperimentalCoroutinesApi"
```
dector, line 605: Was renamed to `-Xopt-in`.

NotWoods:
Removing the coroutine version bump as it causes Glean issues right now. Thankfully its only Kotlin that's blocking other PRs.

PR #10821, opened 2020-05-21
Bug 1642657 - Add a per-commit fenix build that simulates the nightly.

gmierz:
This patch adds a new build that is a copy of the nightly build but it's built on each commit instead of once daily so that we can easily run performance tests on each commit (for app-link tests).

### DIFF
--- a/taskcluster/ci/browsertime/kind.yml
+++ b/taskcluster/ci/browsertime/kind.yml
@@ -105,7 +105,6 @@ jobs:
 
     tp6m-2-cold:
         test-name: google
-        run-on-tasks-for: [github-pull-request]
         treeherder:
             symbol: 'Btime(tp6m-2-c)'
 
--- a/taskcluster/ci/browsertime/kind.yml
+++ b/taskcluster/ci/browsertime/kind.yml
@@ -13,7 +13,7 @@ kind-dependencies:
 primary-dependency: signing
 
 only-for-build-types:
-    - android-test
+    - nightly-simulation
 
 only-for-abis:
     - armeabi-v7a
--- a/taskcluster/ci/browsertime/kind.yml
+++ b/taskcluster/ci/browsertime/kind.yml
@@ -13,7 +13,7 @@ kind-dependencies:
 primary-dependency: signing
 
 only-for-build-types:
-    - nightly
+    - android-test
 
 only-for-abis:
     - armeabi-v7a
@@ -105,6 +105,7 @@ jobs:
 
     tp6m-2-cold:
         test-name: google
+        run-on-tasks-for: [github-pull-request]
         treeherder:
             symbol: 'Btime(tp6m-2-c)'
 
--- a/taskcluster/ci/build/kind.yml
+++ b/taskcluster/ci/build/kind.yml
@@ -104,7 +104,7 @@ jobs:
         treeherder:
             symbol: nightly(B)
 
-    android-test:
+    nightly-simulation:
         attributes:
             nightly: false
         run-on-tasks-for: [github-push]
@@ -112,9 +112,9 @@ jobs:
         include-shippable-secrets: true
         run:
             geckoview-engine: geckoNightly
-            gradle-build-type: fenixNightly
+            gradle-build-type: fennecNightly
         treeherder:
-            symbol: nightly(Bo)
+            symbol: nightlySim(B)
 
     beta:
         attributes:
--- a/taskcluster/ci/build/kind.yml
+++ b/taskcluster/ci/build/kind.yml
@@ -104,6 +104,18 @@ jobs:
         treeherder:
             symbol: nightly(B)
 
+    android-test:
+        attributes:
+            nightly: false
+        run-on-tasks-for: [github-push]
+        include-nightly-version: true
+        include-shippable-secrets: true
+        run:
+            geckoview-engine: geckoNightly
+            gradle-build-type: fenixNightly
+        treeherder:
+            symbol: nightly(Bo)
+
     beta:
         attributes:
             release-type: beta
--- a/taskcluster/ci/signing/kind.yml
+++ b/taskcluster/ci/signing/kind.yml
@@ -66,7 +66,7 @@ job-template:
           default: autograph_apk
     index:
         by-build-type:
-            (fennec-.+|nightly|performance-test|beta|production|debug):
+            (fennec-.+|nightly|performance-test|beta|production|debug|nightly-simulation):
                 type: signing
             default: {}
     run-on-tasks-for: []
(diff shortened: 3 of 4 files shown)


gmierz:
This is in a draft state but I've requested review because I'm not entirely sure if this will do what we need from it:
1. Make a build each commit.
2. Is a copy of the nightly build with a different name.
3. Will have a taskcluster indexed route or an easy way to get the build artifacts.

Thanks!

taskcluster/ci/build/kind.yml
```diff
@@ -104,6 +104,18 @@ jobs:
...
 
+    android-test:
+        attributes:
+            nightly: false
+        run-on-tasks-for: [github-push]
+        include-nightly-version: true
+        include-shippable-secrets: true
+        run:
+            geckoview-engine: geckoNightly
+            gradle-build-type: fenixNightly
```
mcomella, line 115: If we keep the new build type, we'd actually want to use `fennecNightly` (though I expect `fenixNightly` to go away soon anyway).

taskcluster/ci/build/kind.yml
```diff
@@ -104,6 +104,18 @@ jobs:
         treeherder:
             symbol: nightly(B)
 
+    android-test:
+        attributes:
+            nightly: false
+        run-on-tasks-for: [github-push]
```
mcomella, line 110: Could we just make the `fennec-nightly` job run on `github-push` instead of duplicating it?

taskcluster/ci/build/kind.yml
```diff
@@ -104,6 +104,18 @@ jobs:
         treeherder:
             symbol: nightly(B)
 
+    android-test:
```
mcomella, line 107: If we keep the new build type, `android-test` as a name conflicts with the name `androidTest`, which is used for unit/integration tests on device (it's an Android default name). I'm not sure what's a better name – I kind of just want it to say, "nightly" :)

taskcluster/ci/build/kind.yml
gmierz: The issue with this is that it somewhat conflates the term "nightly" which should only be built once per day - not on each commit.

taskcluster/ci/build/kind.yml
gmierz: This is confusing - is `fennecNightly` the nightly fenix build or is it `fenixNightly`?

taskcluster/ci/build/kind.yml
gmierz: I agree, the name isn't good - I was thinking of calling it `opt` or `shippable`. I thought of calling it nightly as well but as I mentioned above, the term nightly should be reserved for things that are build once per day.

taskcluster/ci/build/kind.yml
mcomella: `fennecNightly` is the one we want.

For context, they're both nightly builds of Fenix – one has Firefox Preview branding and one has Firefox for Android branding + the migration code. We've been shipping two different nightly channels. This extra flavor should be removed soon though.

taskcluster/ci/build/kind.yml
mcomella: Technically, that's exactly we want – we want a nightly build to test against for each commit.

Maybe we can call it "nightly simulation" to disambiguate?

taskcluster/ci/build/kind.yml
gmierz: Ah I see, thanks!

taskcluster/ci/build/kind.yml
gmierz: Nightly simulation sounds good! I'll make the change.

gmierz:
I can't test this build with browsertime/raptor because we don't have conditioned profiles for the fennecNightly build.

taskcluster/ci/build/kind.yml
JohanLorenzo: I'm on the fence on this one too. If I'm not mistaken, we had the same kind of issues on the gecko repo. I don't know the details of it. @Callek, do you think we should implement something like a `shippable` flavor?

taskcluster/ci/build/kind.yml
JohanLorenzo: For more information, [this document](https://docs.google.com/document/d/1VeAut1_QL_Jvr3xaNj0PADtzmVJq5evbssp8YlYtYVk/edit#) explains what we're going to do with fennec/fenix flavors.

jeanygong:
Based on @JohanLorenzo recent comment above.. @Callek can you please provide your input ASAP?  This patch is critical and a blocker for work on a regression detection system to support the upcoming Fenix launch.

taskcluster/ci/browsertime/kind.yml
```diff
@@ -105,6 +105,7 @@ jobs:
 
     tp6m-2-cold:
         test-name: google
+        run-on-tasks-for: [github-pull-request]
```
JohanLorenzo, line 108: Nit: remove this line, I think.

taskcluster/ci/signing/kind.yml
```diff
@@ -62,7 +66,7 @@ job-template:
           default: autograph_apk
     index:
         by-build-type:
-            (fennec-.+|nightly|performance-test|beta|production|debug):
+            (fennec-.+|nightly|performance-test|beta|production|debug|nightly-simulation):
```
JohanLorenzo, line 69: Based on Callek's input, I think `nightly-simulation` can be a temporary solution while we implement what Gecko has. Do you think we need to expose the `nightly-simulation` builds to https://firefox-ci-tc.services.mozilla.com/tasks/index/mobile.v2.fenix ? If so, please keep this change. If not, you revert this line.

I don't have any strong opinion. I leave the final call to you @gmierz 🙂

gmierz: I checked with some other people in CI about this and it seems like there's no easy way of doing that - active data will be the simplest solution for this. So, I'll remove this change :)

mcomella:
I believe `test-debug` fails because, In the commit you based your patch off of, the unit tests don't compile – I think if you rebase it should pass.

gmierz:
@JohanLorenzo I spoke with taskcluster and we do have a method for this! It's just a matter of changing `task` to `tasks` in the api call: https://firefox-ci-tc.services.mozilla.com/api/index/v1/tasks/project.mobile.fenix.v2.fennec-nightly.2020.06.01

So I've re-added the indexing in the last commit.

taskcluster/ci/signing/kind.yml
JohanLorenzo: Chatted on Matrix. We want to keep this index for the purpose of being able to query builds made on a given day.

PR #10822, opened 2020-05-21
Issue:116, Issue:173 Verify the nav toolbar buttons[UI Tests]

TejaswiKarasani:
### Pull Request checklist

### After merge

1. click on Show All Checks,


app/src/androidTest/java/org/mozilla/fenix/ui/SmokeTest.kt
```diff
@@ -0,0 +1,70 @@
...
+                // Verify 'x' button actions
+                verifyXButtonActions()
+                pressBack()
+
+                // Verify various items after returning back to the initial WebPage
+                verifyPageContent(defaultWebPage.content)
+                verifyNavURLBar()
+                clickTabCounter()
+
+                // Verify items on HomeScreen
```
isabelrios, line 64: From Browser robot you should use the [openHomescreen](https://github.com/mozilla-mobile/fenix/blob/master/app/src/androidTest/java/org/mozilla/fenix/ui/robots/BrowserRobot.kt#L338) transition go to to Homescreen and then once there use its robots methods for checking that you are  there, going to other menus from there...otherwise you  are adding code from that robot to the BrowserRobot

TejaswiKarasani: yeah got that. So, that might be the test failure in emulator may be?

TejaswiKarasani: image
Do you mean in this way?

isabelrios: yes, something like that 👍

TejaswiKarasani: But when function `clickTabCounter()` is used it will take us to homescreen automatically. Hence, again using transition` openHomeScreen` causing error as it is already taken to the homescreen on clicking the number of tabs button.

isabelrios: :eyes: There has  to be a  way, I would  need to check, but can't you go using a transition directly?

TejaswiKarasani: Yeah I couldn't find any :/

isabelrios: So,  I tried that, and locally is working:
 ```
               verifyNavURLBar()
            }.openHomeScreen {
                verifyHomeScreen()
                verifyExistingTabList()
            }
```

I think the issue is in after the `pressBack()` that the test does not wait  enough until the website is loaded to continue.
To be sure, try adding `TestAssetHelper.waitingTime` after that pressBack() if test passes, we will try to  find a way to be sure the website is loaded.

Comment (TejaswiKarasani) [reply app/src/androidTest/java/org/mozilla/fenix/ui/SmokeTest.kt:64]: It's seems you missed out adding 'clickTabCounter' function🤔

Comment (isabelrios) [reply app/src/androidTest/java/org/mozilla/fenix/ui/SmokeTest.kt:64]: No, I didn't, it can go from navigation to homescreen if there is a transition defined...

Comment (TejaswiKarasani) [reply app/src/androidTest/java/org/mozilla/fenix/ui/SmokeTest.kt:64]: Oops,I should have gone through it carefully. Thanks for pointing out :)

Comment (TejaswiKarasani) [reply app/src/androidTest/java/org/mozilla/fenix/ui/SmokeTest.kt:65]: After pushing the code, test is failing again!

File: app/src/androidTest/java/org/mozilla/fenix/ui/SmokeTest.kt [hunk 2]
```diff
@@ -0,0 +1,71 @@
...
+                // Verify the items of Nav URL Bar
+                verifyNavURLBarItems()
+
+                // Verify 'x' button actions
+                verifyXButtonActions()
+                pressBack()
+                TestAssetHelper.waitingTime
+
+                // Verify various items after returning back to the initial WebPage
+                verifyPageContent(defaultWebPage.content)
```
isabelrios, line 61: @TejaswiKarasani I think the error you are seeing comes from this line...can you try to wait longer just to check that? if it still fails, try commenting out  this line...thinking about more options here in the meantime...

TejaswiKarasani: Ok

isabelrios: Other option may be to check the content in the url bar...and see if that works...

TejaswiKarasani: > Other option may be to check the content in the url bar...and see if that works...

When I saw in UI Automator, I couldn't find any specific id for the content in the url bar

isabelrios: there is text (although I'm not sure you can say starts with...the port number will change each time) 
Screenshot 2020-05-25 at 17 56 50

app/src/androidTest/java/org/mozilla/fenix/ui/SmokeTest.kt
```diff
@@ -0,0 +1,71 @@
...
+                verifyNavURLBar()
+                }.openHomeScreen {
+
+                // Verify items on HomeScreen
+                verifyHomeScreen()
+                verifyExistingTabList()
```
isabelrios, line 67: Oh and if I'm not mistaken, we are missing here step 6:

> Back on the page, tap the menu button.
> Expected Result
> The main menu is displayed.
>

app/src/androidTest/java/org/mozilla/fenix/ui/SmokeTest.kt
TejaswiKarasani: Yeah, I thoght it was the whole red highlight box and didn't struck my mind that it is for text inside the red box and by the way there is some particular function for testhelper.longWaitingTime or testhelper.longWaitingTime...and set time for minutes and seconds? or should I create a separate function for the long waiting time

app/src/androidTest/java/org/mozilla/fenix/ui/SmokeTest.kt
TejaswiKarasani: Yeah I missed that since that text was long, I forgot to expand the excel sheet :/. Thanks for the check :)

app/src/androidTest/java/org/mozilla/fenix/ui/SmokeTest.kt
isabelrios: you may be able to use one off those

TejaswiKarasani: Yeah cool.!

app/src/androidTest/java/org/mozilla/fenix/ui/SmokeTest.kt
```diff
@@ -0,0 +1,76 @@
...
+                verifyNavURLBarItems()
+
+                // Verify 'x' button actions
+                verifyXButtonActions()
+                pressBack()
+                TestAssetHelper.waitingTime
+
+                // Verify various items after returning back to the initial WebPage
+                verifyNavURLBar()
+                verifyUrl((defaultWebPage.url).toString())
```
isabelrios, line 62: oh, now test is failing here, it does not find the  urlBar element...it is like nothing in that page is seen when running the test...what about tryiing  to add aftter line 58 this:
` browserScreen {
                    verifyNavURLBar()
                    verifyUrl((defaultWebPage.url).toString())
                }`

Kind of forcing to say where you are.. and then continue with `.openHomeScreen{....}` at least locally that works...worth trying...

TejaswiKarasani: image

Is the the one the approach you were suggesting or should I need to add any other?

isabelrios: that's right but change the name, is not openBrowserScreen but just browserScreen

TejaswiKarasani: > that's right but change the name, is not openBrowserScreen but just browserScreen

There isn't any transition named browserScreen , I think I need to create one should I go ahead with that one? Sorry for so many follow ups :/

isabelrios: it is not a transition, it is a function in BrowserRobot

Issue #7225, opened 2019-12-16
Enable more aggressive StrictMode penalties on startup in debug builds


mcomella:
See #7226 for aggressive StrictMode being enabled everywhere.

### What
StrictMode allows us to identify performance issues where users do IO on the main thread. Since we're making improvements to startup, we don't want it to regress: we should enable more advanced penalties such as a blocking dialog or crashing the app to ensure we don't regress the behavior. Note that devs can override the behavior for a single operation if the main thread IO is necessary.

### Impact
Without this, we may suffer from unintentional regressions as users do main thread IO, esp. from SharedPreferences calls, on startup. With this, devs will be more aware and may think of alternative solutions.

### Acceptance criteria
- StrictMode with more aggressive penalties is enabled for startup and disabled after startup
  - Is realistic? detekt rule instead?
- Violations are either fixed or ignored. If they are ignored but fixable, we should file an issue on our project board to address them

mcomella:
Here's some notes on accessing `SharedPreferences` on the main thread, from 9 years ago: https://stackoverflow.com/a/4371883

hawkinsw:
I do not believe that we can do this detection offline. I don't think that detekt has enough context when it executes to determine whether a) the method cannot be called on MT and/or b) the method is invoked in a context that actually moves it off the MT (e.g., a `CoroutineScope`).

PR #10831, opened 2020-05-21
for #7225 enable more aggressive strictmode penalty for startup

sraturi:
### Pull Request checklist

### After merge

1. click on Show All Checks,


sraturi:
Logs from Strictmode on startup 
[strictModeLogs.txt](https://github.com/mozilla-mobile/fenix/files/4663971/strictModeLogs.txt)

AaronMT:
ui-test-x86-debug 

```Kotlin
java.lang.RuntimeException: Could not launch intent Intent { act=android.intent.action.MAIN flg=0x14000000 cmp=org.mozilla.fenix.debug/org.mozilla.fenix.HomeActivity } within 45000 milliseconds.
```

mcomella:
Sorry for blocking the review and then punting 😓 but I asked MarcLeclair to take the first stab at your reviews going forward though let me know if there's ever anything I can help with!

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -314,23 +315,24 @@ open class FenixApplication : LocaleAwareApplication() {
 
     private fun enableStrictMode() {
         if (Config.channel.isDebug) {
-            StrictMode.setThreadPolicy(
-                StrictMode.ThreadPolicy.Builder()
-                    .detectAll()
+            Handler().postAtFrontOfQueue(Runnable {
+                StrictMode.setThreadPolicy(
+                    StrictMode.ThreadPolicy.Builder()
+                        .detectAll().penaltyLog()
```
MarcLeclair, line 321: I feel as if we should set `penaltyDialog()` instead of `penaltyLog()` and (maybe, not sure about this one, what do you think about this @mcomella) using `penaltyDeathOnNetwork()` since it would allow us to kill the app on network operations on the MT on start up.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -167,11 +169,28 @@ class HomeFragment : Fragment() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         postponeEnterTransition()
+        changeStrictModePenalty()
```
MarcLeclair, line 172: This will only change policy when the browser is started from home screen. However, if we start the browser any other way (app link, i.e: click a link in gmail and open it in Fenix. Another example would be a shortcut on the phone home screen)

sraturi: So I checked different ways to open the app such as opening the lite browser in Gmail, starting Fenix from a Gmail link. the `IntentReceiverActivity` led to the home fragment in both cases. For the shortcut, did you mean just dragging the app to the home screen to create a shortcut and launch the application from there? 
is there another start case I am missing?

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
sraturi: Yup done! I am still waiting on `penaltyDeathOnNetwork()` suggestion from @mcomella

mcomella: MarcLeclair connected with me async and I said something like,

> Hmm, that's curious – I thought Android automatically killed you, even in production, on main thread network access (but I guess MLS bug proves that wrong)
>
> I'm down but maybe land it in a separate patch so it's easy to see if it causes problems?

mcomella: So yes, penaltyDeathOnNetwork sounds good to me.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
MarcLeclair: @sraturi after looking at it for a bit, I think we could go ahead and move the strictMode changes back to `HomeActivity` (right before transitioning to `homefragment`) and see how well it works. in catching violations. How does that sound?

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -174,6 +176,10 @@ class HomeFragment : Fragment() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         postponeEnterTransition()
+        context?.runOnlyInMainProcess {
```
MarcLeclair, line 179: More of a question than a request for change, why the `runOnlyInMainProcess` call here, is there a specific reason?

sraturi: Originally, the FenixApplication class was enabling the strict mode inside the `setupInMainProcessOnly` function. since we are also calling the `enableStrictMode` from HomeFragment, I figured they should be called under same condition?

MarcLeclair: So I don't think this is to be worried since `HomeFragment.onCreate` should always be on the MT 😄  I'd go with removing `runOnlyInMainProces` in this case. iirc the reason why we called `setupInMainProcessOnly` is that there's option to offload to background threads in our application start and we can't do that with some of our stuff.

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -391,4 +369,35 @@ open class FenixApplication : LocaleAwareApplication() {
         applicationContext.resources.configuration.uiMode = config.uiMode
         super.onConfigurationChanged(config)
     }
+
+    companion object {
```
MarcLeclair, line 373: As for a companion object, I think that `FenixApplication` should not own `strictMode`. A personal file might help us be clearer in what our object does.  So maybe creating a `StrictModeManager` (or any other name that conveys what we're doing !) would help clear any confusion and allow other devs to know what we're doing with it 😄

PR #10833, opened 2020-05-21
Bug 1639955 - Fix geckodriver path for performance tests.

gmierz:
This patch fixes an issue with geckodriver toolchain paths (from public/build/ to public/).

### DIFF
--- a/taskcluster/ci/browsertime/kind.yml
+++ b/taskcluster/ci/browsertime/kind.yml
@@ -105,7 +105,6 @@ jobs:
 
     tp6m-2-cold:
         test-name: google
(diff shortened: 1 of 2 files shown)


whimboo:
Please note that I consider this a regression, and that I would like to see the geckodriver binaries still under `public/build`.  Please see https://bugzilla.mozilla.org/show_bug.cgi?id=1639955 for updates.

whimboo:
This PR is not necessary because it will be fixed with the next merge to mozilla-central via the before-mentioned bug.

gmierz:
Closed since this was addressed in mozilla-central.

Issue #9563, opened 2020-03-31
[Bug] Performance: Leanplum delays app link (non-startup) for 10-100ms by doing SQL on the UI thread


mstange:
Profile: https://bit.ly/33YhjIi

In this profile, `android.app.Activity.onResume` takes over 100ms because leanplum's `com.leanplum.internal.RequestOld.saveRequestForLater` inserts something into a database on the UI thread and the insertion is taking a long time.

This profile is a recording of opening mozilla.org via an app link intent in a Fenix that's already running. In this particular run, the first network request starts 310ms after Fenix receives the intent. 100ms of that delay are due to the Leanplum issue.

I've collected other profiles where this SQL operation takes less time.

### Device information

* Android device: Moto G5
* Fenix version: Nightly 200331 06:01 (Build #20910606) 38.0.20200329190103, 1341d92fa GV: 76.0a1-202003027094805 AS: 0.55.0

_edit:_ see https://github.com/mozilla-mobile/fenix/issues/7442 for an alternative report.

csadilek:
Unfortunately, this isn't fixable in Fenix, as it happens as part of lifecycle callbacks in Leanplum internals.

From the docs:
_If your application class does not extend LeanplumApplication, you must call enableLifecycleCallbacks (like the example below). Failure to do so will prevent the SDK from properly tracking states and displaying in-app messages on app resume._

In both cases `LeanplumActivityHelper.enableLifecycleCallbacks(application)` will be called resulting in lifecycle events being observed on the main thread with subsequent calls to `LeanplumEventDataManager.insertEvent`.

Looks like we need to file this with Leanplum.

csadilek:
Leanplum got back to us confirming that there's currently no way to prevent these DB writes on the main thread. This may be addressed in a future version of their SDK.

csadilek:
Latest Nightly (200525 06:01) was upgraded to Leanplum 5.4.0 which contains a fix for this. I can no longer reproduce this problem.

The `insertEvent` calls that caused this are now all added to a queue which is processed on a background thread: https://github.com/Leanplum/Leanplum-Android-SDK/blob/master/AndroidSDKCore/src/main/java/com/leanplum/internal/RequestOld.java#L227

@mstange Do you want to / have time to verify too before closing this?

csadilek:
[...]

PR #10854, opened 2020-05-22
Closes #10827: Upgrade to Leanplum 5.4.0

### LINKED ISSUE #10827: FNX-4605 ⁃ Upgrade to Leanplum 5.4.0
There's a new version of Leanplum addressing the performance concerns we've raised in: https://github.com/mozilla-mobile/fenix/issues/9563

Let's upgrade and verify.

csadilek:
This upgrade should fix the performance issue described in: https://github.com/mozilla-mobile/fenix/issues/9563


csadilek:
Smoke tested metrics/events...all good. Let's verify that push is working before landing. Also verified that the original code causing the performance issues is no longer there :), but let's verify by creating profiles once we have this in Nightly.

Issue #10494, opened 2020-05-07
FNX-4326 ⁃ [Bug] Fenix is stuck after a few second


ValentinTimisica:
## Steps to reproduce
* Open Fenix
* Open fast the 3 dot menu
* Wait a few second (I waited around 10 second)
* Press outside the menu

### Expected behavior
* The app works without issues.

### Actual behavior
* The app is stuck. You can only select the items from the menu (check the video below).

### Device information
* Android device: Samsung Galaxy Note 10 (Android 10)
* Fenix version: latest master

### Additional info
* Very similar to #10110 and #10488. It may be the same problem.
* After I tested a few times I noticed that is 100% reproducible if you tap fast on the 3 dot menu after you opened the app. In case you wait a few seconds (let's say 10s) before opening the menu the bug will not reproduce.
* I don't know if it's related, but I was already logged in.
* [Video](https://drive.google.com/file/d/1anaoCIPtZmSobGZPvIOnIF6nIRHCwIjg/view?usp=sharing)

ekager:
I just saw this and it's quite disruptive.

ValentinTimisica:
After further investigation I found out that you need to be logged in to reproduce the bug.
I discovered that commenting the code inside [onAuthenticated](https://github.com/mozilla-mobile/fenix/blob/1e83f54a942efe8bc866da1ad67be6794f324701/app/src/main/java/org/mozilla/fenix/home/HomeMenu.kt#L183) will no longer reproduce the bug.

Another interesting thing is that if you open the app and don't press the 3 dot menu 3 callbacks related to FxaAccount will be triggered:
1) in `HomeMenu.kt` - [line](https://github.com/mozilla-mobile/fenix/blob/1e83f54a942efe8bc866da1ad67be6794f324701/app/src/main/java/org/mozilla/fenix/home/HomeMenu.kt#L183)
2) in `Mode.kt` - [line](https://github.com/mozilla-mobile/fenix/blob/6c38b773151e1c17a8016d851407bb3cd7de4ef9/app/src/main/java/org/mozilla/fenix/home/Mode.kt#L75)
3) in `HomeFragment.kt` - [line](https://github.com/mozilla-mobile/fenix/blob/f7b4f1c9594bc6e942941252d7539e4649f76aef/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt#L417)

[...]

PR #10863, opened 2020-05-22
For #10494: Process HomeMenu builder changes on the main thread

grigoryk:
Problem was that we were trying to process menu changes (in response to account manager events) on some background thread as that's what account manager emits them on, so some code internally in PopupWindow's dismiss handling (i think, didn't dig very deeply here) was silently giving up and we'd get into a bad state.

The reason this seemingly only happened if you quickly opened a menu on startup is because account manager isn't initialized until sometime after the startup finished. So the trick was to open the menu (and register account manager state callbacks) before it got initialized, so that the callbacks are invoked.

This should also reproduce in other, much more obscure ways, e.g. if you open the menu right before sync is scheduled to run in the background, change FxA password on another connected client, and then eventually receive a onAuthenticationProblem callback.

One "problem" this introduces is that if you open the menu right after the startup, it will be dismissed in a few seconds (as account manager is initialized). Fixing that will require changes to a-c, so I'll leave that to a follow-up.

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/home/HomeMenu.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/HomeMenu.kt
@@ -8,6 +8,9 @@ import android.content.Context
 import androidx.core.content.ContextCompat.getColor
 import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.LifecycleOwner
+import androidx.lifecycle.lifecycleScope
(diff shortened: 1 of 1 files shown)


grigoryk:
I think you reviewed the original refactor, @ekager? r?

Issue #10865, opened 2020-05-22
FNX-4640 ⁃ Tab Tray - Implement Three-Dot Menu


darkwing:
We need to restore the three-dot functionality to the tab tray.  Menu items should be:

* Save to collection
* Share all tabs
* Close all tabs

Functionality was removed with: https://github.com/mozilla-mobile/fenix/pull/10835

AndiAJ:
Hi, **verified as fixed** on the latest **`Nightly Build 200529 from 5/29`** using the following devices:
• Google Pixel 3a (Android 10)
• Huawei Mate 20 Lite (Android 9)
• OnePlus A3 (Android 6.0.1)

► Video
![20200529-133556](https://user-images.githubusercontent.com/51314259/83253551-d3a24100-a1b5-11ea-97bf-0177bb6d49c5.gif)

PR #10869, opened 2020-05-22
For #10865 - Implement 3 dot menu for tab tray

darkwing:
## To Do:

- [x] Wire up "close all tabs"
- [x] Wire up "save to collection"
- [x] Share - The data is being fetched correctly but the navigation to share fragment isn't happening.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
@@ -95,7 +95,6 @@ import org.mozilla.fenix.ext.nav
 import org.mozilla.fenix.ext.requireComponents
 import org.mozilla.fenix.ext.sessionsOfType
 import org.mozilla.fenix.ext.settings
-import org.mozilla.fenix.home.HomeFragmentDirections
 import org.mozilla.fenix.home.SharedViewModel
 import org.mozilla.fenix.tabtray.TabTrayDialogFragment
 import org.mozilla.fenix.theme.ThemeManager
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
```diff
@@ -22,6 +22,7 @@ import androidx.navigation.fragment.findNavController
 import com.google.android.material.snackbar.Snackbar
 import kotlinx.android.synthetic.main.fragment_browser.*
 import kotlinx.android.synthetic.main.fragment_browser.view.*
+import kotlinx.android.synthetic.main.tab_header.view.*
```
boek, line 25: What is this import for?

app/src/main/java/org/mozilla/fenix/tabtray/TabTrayDialogFragment.kt
```diff
@@ -14,13 +14,19 @@ import kotlinx.android.synthetic.main.component_tabstray.view.*
...
 import org.mozilla.fenix.R
+import org.mozilla.fenix.ext.requireComponents
 
 class TabTrayDialogFragment : AppCompatDialogFragment(), TabTrayInteractor {
+
     interface Interactor {
         fun onTabSelected(tab: Tab)
         fun onNewTabTapped(private: Boolean)
+        fun onShareTabsClicked(private: Boolean)
+        fun onSaveToCollectionClicked(private: Boolean)
```
boek, line 28: I would remove the private flag here as we don't allow to save private tabs to collections

app/src/main/java/org/mozilla/fenix/tabtray/TabTrayView.kt
```diff
@@ -88,6 +100,26 @@ class TabTrayView(
             TabsTouchHelper(tray.tabsAdapter).attachToRecyclerView(tray)
         }
 
+        tabTrayItemMenu = TabTrayItemMenu(view.context, { view.tab_layout.selectedTabPosition == 0 }) {
+            when (it) {
+                is TabTrayItemMenu.Item.ShareAllTabs -> interactor.onShareTabsClicked(
+                    view.tab_layout.selectedTabPosition == 1
```
boek, line 106: let's add a computed property to this view now that we're using it everywhere

`val isPrivateModeSelected: Boolean get() = view.tab_layout.selectedTabPosition == PRIVATE_TAB_ID`

Issue #10864, opened 2020-05-22
FNX-4639 ⁃ Tab Tray - Implement Snack bar & Undo for closing Tabs


darkwing:
With the new tab drawer, we no long show the snackbar after closing a tab.  We should restore this functionality.

liuche:
Filed #10986 for restoring tabs in-place, to match current tabs tray behavior and make "Undo" actually feel like "undoing"

AndiAJ:
Hi, **verified as fixed** on the latest **`Nightly Build 200529 from 5/29`** using the following devices:
• Google Pixel 3a (Android 10)
• Huawei Mate 20 Lite (Android 9)
• OnePlus A3 (Android 6.0.1)

► Video Normal Browsing
![20200529-120759](https://user-images.githubusercontent.com/51314259/83242776-9a61d500-a1a5-11ea-9fd4-95b79dbf28dd.gif)

► Video Private Browsing
![20200529-120859](https://user-images.githubusercontent.com/51314259/83242839-ba919400-a1a5-11ea-809a-1ca856c7a57d.gif)

PR #10876, opened 2020-05-23
For #10864 - Display snack bar to undo closing tabs

darkwing:
This should be the correct setup but the snackbar item doesn't display; I'm worried that it's not displaying on top of our new tab tray dialog.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/tabtray/TabTrayDialogFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/tabtray/TabTrayDialogFragment.kt
@@ -68,7 +68,6 @@ class TabTrayDialogFragment : AppCompatDialogFragment(), TabTrayInteractor {
     }
 
     override fun onTabClosed(tab: Tab) {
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/mozilla/fenix/tabtray/TabTrayDialogFragment.kt
```diff
@@ -10,17 +10,22 @@ import android.view.View
...
 import org.mozilla.fenix.R
+import org.mozilla.fenix.ext.components
+import org.mozilla.fenix.ext.requireComponents
+import org.mozilla.fenix.utils.allowUndo
 
 class TabTrayDialogFragment : AppCompatDialogFragment(), TabTrayInteractor {
     interface Interactor {
         fun onTabSelected(tab: Tab)
         fun onNewTabTapped(private: Boolean)
+        fun onTabClosed(tab: Tab)
```
boek, line 28: Do we still need this if we moved the snackbar to the `TabTrayDialogFragment`?

Issue #9815, opened 2020-04-09
[Bug] "Open in Firefox Preview" from Custom Tab opens home page not tab


ekager:
## Steps to reproduce
1. Open any website in custom tab
2. Open menu and click "Open in Firefox Preview"

### Expected behavior
Webpage opens in app

### Actual behavior
App opens to home screen

### Device information

* Android device: ?
* Fenix version: ?

mcarare:
For me the tab is opened, but without navigating to it.

ekager:
Right, it should navigate to it.

ekager:
CC @pocmo I'm not positive, just started digging around, but could this be a regression from https://github.com/mozilla-mobile/fenix/pull/8823 ? It was the only recent change to intent processing I found (and nothing with the `OpenInFenixIntent` was changed).

Looks like it's being processed as an `OTHER` intent (with a `null` processor). Should we be adding another custom extra here for HomeActivity to process and open to the browser correctly?

apoorv9990:
Hi, I would like to work on this. I tried to repro it but not sure how to open a custom tab on the Firefox app  :sweat_smile:. Any pointers would be helpful.

linuxandria:
> Hi, I would like to work on this. I tried to repro it but not sure how to open a custom tab on the Firefox app 😅. Any pointers would be helpful.

Open a link in an app that uses custom tabs, eg Telegram

liuche:
We'll add it to our sprint, and the eng work here is to shepherd it along and make sure it's making progress.

pocmo:
Just landed a patch for this and it should get into the next Nightly build of Fenix.

softvision-miralobontiu:
Verified as fixed on the latest Nightly build from 5/29 with:
* OnePlus 5T (Android 9),
* Google Pixel (Android 10), and
* Sony Xperia Z5 Premium (Android 7.1.1).

PR #10917, opened 2020-05-26
Issue #9815: Allow Intents to set HomeActivity.OPEN_TO_BROWSER.

pocmo:
This is a fix for issue #9815. We are allowing `Intent`s to set the `HomeActivity.OPEN_TO_BROWSER` flag and only override it with `intentProcessorType.shouldOpenToBrowser(intent)` if no value was provided. We then use this to always set it to `true` in `DefaultBrowserToolbarController.openInFenixIntent` when switching from a custom tab to the browser.


pocmo:
The ui test failures look unrelated (download prompt). So I'll try rerunning.

pocmo:
Okay, green, landing! :)

Issue #8319, opened 2020-02-11
Nest extension toolbar icons in a submenu entry


yoasif:
Been seeing feedback in various places that it the main menu is overly long when uBlock Origin is installed, and this is just going to get worse as more extensions are supported with toolbar icons.

In addition, it seems to me a little weird that an extension is allowed to place icons in the main menu, which feels like primary UI to me that doesn't really feel like a model that feels clean or thought through.

### Why/User Benefit/User Problem

As a user, I would like to not clutter my main menu with extension icons, but still have them easily accessible in order to cleanly separate between Fenix functionality vs. extension functionality, and to make the menu a manageable size on smaller screens. 

### What/Requirements

1. submenu entry labeled "Add-ons"
2. Put "Add-ons manager" inside here
3. Put extension icons here, like uBlock Origin's toolbar icon

### Acceptance Criteria (how do I know when I’m done?)

There is a submenu containing both the add-on manager and add-on toolbar icons.

There is a similar idea in experiments on Chrome desktop, that may be interesting to look at from a UX perspective: https://lifehacker.com/how-to-enable-chromes-new-extension-menu-1837559402

csadilek:
Thanks for filing @yoasif. I think there's an updated design in the works /cc @brampitoyo.

brampitoyo:
@yoasif It’s exactly as you’ve described.

Our problem today:

1. Fenix’s main menu is too full
    - In addition to other items, we have 2 more: “Add-ons Manager” and “uBlock Origin”
2. We’re planning to support more add-ons in the future
    - To add to our 2 items, we might have anywhere between one and many

The solution:

1. Either take add-ons out of the menu
2. Or keep it in the menu, but make it scaleable

Solution 2 seems the most realistic in the short term. Note: we should still try other solutions e.g. putting add-ons as an icon in the toolbar, but remember that the toolbar is also full with URL at the moment.

So here’s what I’m proposing – something that we will do post-release:

1. There’s only one item in the main menu: “Add-ons”

<img width="360" src="https://user-images.githubusercontent.com/51007/74397248-12b3bd00-4e79-11ea-8cbf-1010f67bed0f.png">

2. When tapped, it will show a submenu
    - To dismiss, tap anywhere outside the menu
    - To go back to the main menu, tap the back arrow – is this necessary?
    - For the moment, the first add-on is located up top and Add-ons Manager is on bottom.
    - We could think of a more thumb-friendly design, where perhaps the order is reversed.

[...]

PR #10923, opened 2020-05-26
For #8319: Integrate nest WebExtensionBrowserMenu support

psymoon:
Also update Android Components version to 43.0.20200526130035.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
@@ -68,6 +68,10 @@ class DefaultToolbarMenu(
             menuItems,
             endOfMenuAlwaysVisible = !shouldReverseItems,
             store = store,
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
```diff
@@ -68,6 +68,9 @@ class DefaultToolbarMenu(
             menuItems,
             endOfMenuAlwaysVisible = !shouldReverseItems,
             store = store,
+            onAddonsManagerTapped = {
```
Amejia481, line 71: nit: I think we need to provide a value for  `webExtIconTintColorResource`  without it, the icons don't show properly  
image

Amejia481: We could use `primaryTextColor()` :)

psymoon: Oops, added.

PR #10928, opened 2020-05-26
Closes #10927: Improve wait on session loaded for UI tests

### LINKED ISSUE #10927: FNX-4695 ⁃ [UI Tests] Improve wait on session loaded
A number of our tests are flaky because Espresso continues during page load which causes them to break on assertions afterwards. We should use a an IdlingResource implementation that waits until the current session is not loading anymore.

AaronMT:
Closes #10927

> A number of our tests are flaky because Espresso continues during page load which causes them to break on assertions afterwards. We should use a an IdlingResource implementation that waits until the current session is not loading anymore.

Opening as draft to run a few times.


AaronMT:
ETP tests with the onboarding notification blocking my matcher check 🙄

sv-ohorvath:
> ETP tests with the onboarding notification blocking my matcher check 🙄

Oooh that might be a good reason why the tests fail, but weird it doesn't appear in videos or happen on emulator.

AaronMT:
Thanks. It's late my time, I'll merge tomorrow to keep an eye.

Issue #11004, opened 2020-05-28
FNX-11861 ⁃ [Bug] The migration screen's body text is missing


softvision-miralobontiu:
## Steps to reproduce
1. Have Fennec Beta installed.
1. In a tab, open https://firefox-ci-tc.services.mozilla.com/tasks/index/project.mobile.fenix.v2.fennec-beta.2020.05.27.revision/34e156ac1409855a9c10971b2ff079cc5da54a04 and select to install the right build.
1. Allow installation, and install the migration build.
1. Observe the migration screen.

### Expected behavior
Loading screen contains: 

* Firefox Preview logo
* "Welcome to an all-new Firefox" title
* "A completely redesigned browser awaits, with improved performance and features to help you do more online." text
* "Please wait while we update Firefox with your"
* "Settings"
* "History"
* "Bookmarks"
* "Passwords"

### Actual behavior
The migration screen doesn't contain all the elements - please see the screenshot below.

### Device information
* Android devices: OnePlus 5T (Android 9), Google Pixel 3a (Android 10)
* Fenix version: Beta 5.1.0-beta.2 migration build / Fennec 68.7b1

![picturemessage](https://user-images.githubusercontent.com/40694158/83143245-506fe500-a0fa-11ea-84df-a4028dd08a64.png)UNITO-UNDERSCORE!t4ak4ylp fnj!

ekager:
Ah makes sense, we are forcing the background to be white, but setting the text to the theme attr of primaryText in `Body16TextStyle`, so any user who is in their phone in dark mode will see a broken view (we follow device theme by default).

ebalazs-sv:
Verified as fixed on Beta Migration build 5.1.0-beta.3 with OnePlus 5T (Android 9), Google Pixel 3a (Android 10).
I will remove the qa:needed label. Thanks!

PR #11006, opened 2020-05-28
For #11004: Fix invisible text on migration UI

jonalmeida:
For some text and colors, we were using the default styling where
possible. These styles contain references that react to theme changes
like dark mode. Since the migration UI does not respect these changes,
we should not use them.

1. click on Show All Checks,


jonalmeida:
I'm going to make more one change that removes another color reference.

PR #11014, opened 2020-05-28
Fixes #9832 - Change targetSdkVersion to Android 10

### LINKED ISSUE #9832: Update compile and target SDK to 29
With the Android 11 (SDK 30) soon to be released, we would be two versions below that.

We should update our `compileSdkVersion` and `targetSdkVersion` in `Config.kt` to 29 sooner than later.

NotWoods:
No major changes to upgrade now!

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/settings/logins/SavedLoginsAuthFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/logins/SavedLoginsAuthFragment.kt
@@ -18,6 +18,7 @@ import androidx.appcompat.app.AlertDialog
 import androidx.biometric.BiometricManager
 import androidx.biometric.BiometricPrompt
 import androidx.core.content.ContextCompat
+import androidx.core.content.getSystemService
 import androidx.lifecycle.lifecycleScope
 import androidx.navigation.fragment.findNavController
 import androidx.preference.Preference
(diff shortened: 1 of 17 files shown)


app/src/main/java/org/mozilla/fenix/settings/logins/SavedLoginsAuthFragment.kt
```diff
@@ -256,8 +257,9 @@ class SavedLoginsAuthFragment : PreferenceFragmentCompat(), AccountObserver {
         }
     }
 
+    @Suppress("Deprecation")
     private fun showPinVerification() {
-        val manager = activity?.getSystemService(KEYGUARD_SERVICE) as KeyguardManager
+        val manager = activity?.getSystemService<KeyguardManager>()!!
```
jonalmeida, line 262: I don't think we should use un-safe calls if the newer target SDK asserts them as nullable. Some devices maybe not support some system services and we shouldn't crash because of that.

app/src/main/java/org/mozilla/fenix/utils/ClipboardHandler.kt
```diff
@@ -16,7 +17,7 @@ private const val MIME_TYPE_TEXT_HTML = "text/html"
  * A clipboard utility class that allows copying and pasting links/text to & from the clipboard
  */
 class ClipboardHandler(context: Context) {
-    private val clipboard = context.getSystemService(Context.CLIPBOARD_SERVICE) as ClipboardManager
+    private val clipboard = context.getSystemService<ClipboardManager>()!!
```
jonalmeida, line 20: Same as the previous comment.

app/src/main/java/org/mozilla/fenix/settings/logins/SavedLoginsAuthFragment.kt
NotWoods: We're already using unsafe calls as we case nullable to not-null. The SDK doesn't assert if a system service is null or not, the extension method simply always returns a nullable type.

PR #11108, opened 2020-06-01

MickeyMoz:


AaronMT:
Just to double check on ui-test-x86, is this rebased on a recent master?

mcarare:
This branch is only 7 commits behind, so I think it includes those changes. Closing and reopening to see if the issue persists.

PR #11167, opened 2020-06-03
Fixes mozilla-mobile#10989: Link tabs feature to view lifecycle

ebarrenechea:
The `TabsTrayFeature` was not previously connected to the view lifecycle owner, which resulted in the memory leaks affecting both the `TabsTrayPresenter` and `TabsTrayInteractor`.
 
### Pull Request checklist

- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features. In addition, it includes a screenshot of a successful [accessibility scan](https://play.google.com/store/apps/details?id=com.google.android.apps.accessibility.auditor&hl=en_US) to ensure no new defects are added to the product.

### After merge
- [ ] **Milestone**: Make sure issues finished by this pull request are added to the [milestone](https://github.com/mozilla-mobile/fenix/milestones) of the version currently in development.

### To download an APK when reviewing a PR:
1. click on Show All Checks,
2. click Details next to "Taskcluster (pull_request)" after it appears and then finishes with a green checkmark,
3. click on the "Fenix - assemble" task, then click "Run Artifacts".
4. the APK links should be on the left side of the screen, named for each CPU architecture

Issue #10989, opened 2020-05-28
FNX-11846 ⁃ [Bug] TabTrayView memory leak


ekager:
## Steps to reproduce
1. Have Tabs Tray enabled
1. Click in and out of tabs tray while browsing

![Screenshot](https://user-images.githubusercontent.com/25442310/83164957-bb50fa00-a0c1-11ea-9a6e-9cd2970ee43e.png)UNITO-UNDERSCORE!20200528-115822!

### Device information
* Android device: ?
* Fenix version: ?

ebarrenechea:
@boek I was playing around with the new tab tray feature and it seems this is still an issue (latest from master). I had a look at it and I think it's because the coroutine to collect the flow in `TabsTrayPresenter` is not being reset after the `TabTrayDialogFragment` is destroyed. 

I had a look at the code and the `TabsFeature` already extends `LifecycleAwareFeature` which should take care of stopping the coroutine based on lifecycle events. The missing piece seems to be registering the feature with the dialog fragment's lifecycle. Passing the lifecycle to the `TabsTrayView` constructor and having the view register on `init` works but I don't know if that's the correct pattern to apply in this codebase (maybe there's a component that handles that registration for the view).

I can put up a PR with my changes which seem to address the leak if you'd like but I don't know how useful it would be if there is a different approach to observing lifecycle events from features.

AndiAJ:
Hi, **verified as fixed** on the latest **`Fenix debug build from Master`** using the following devices:
• Google Pixel 3a (Android 10)
• Huawei Mate 20 Lite (Android 9)

► Screenshot
![rsz_screenshot_20200609-144307](https://user-images.githubusercontent.com/51314259/84144265-a7fa4300-aa60-11ea-8c62-94c1a6e513d8.png)

PR #11170, opened 2020-06-03
For #10989 - Fixes memory leak caused by retaining the TabTrayView and TabsFeature

boek:


ekager:
I'm assuming after this lands we can close https://github.com/mozilla-mobile/fenix/pull/11167 ?

Issue #11201, opened 2020-06-03
FNX-12086 ⁃ Copy updates for synced tabs


topotropic:
* Update "Synced Tabs" to "Synced tabs" throughout the app
* Update "Tabs" to "Open tabs" in account settings

Diana-Rus:
Hi verified with Google Pixel 3 XL (Android 9) and Samsung Galaxy S9 (Android 8) on Firefox Preview Nightly 6/15.
- "Open tabs" in account settings ✔️
![Screenshot_20200615-160911](https://user-images.githubusercontent.com/49859863/84664759-ad0b3680-af27-11ea-81bc-8cbf536145e2.png)

- "Synced tabs" in menu ✔️
![Screenshot_20200615-160945](https://user-images.githubusercontent.com/49859863/84664702-9a90fd00-af27-11ea-97f7-1230e5b5efd1.png)

- "Synced tabs" page title ✔️
![Screenshot_20200615-161006](https://user-images.githubusercontent.com/49859863/84664741-a8468280-af27-11ea-9261-50de81250d51.png)

Note: The "Synced tabs" string is not translated in all the available in browser app languages. I'll create a new ticket for that.

PR #11220, opened 2020-06-04
For #11201: Updates 'synced tabs' and account settings 'tabs' strings

ValentinTimisica:
### Pull Request checklist

- [x] **Tests**: This PR does not include thorough tests. Strings changes.
- [x] **Screenshots**: This PR does not include screenshots or GIFs of the changes made.

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -96,7 +96,7 @@
     <!-- Browser menu toggle that installs a Progressive Web App shortcut to the site on the device home screen. -->
     <string name="browser_menu_install_on_homescreen">Install</string>
     <!-- Menu option on the toolbar that takes you to synced tabs page-->
-    <string name="synced_tabs">Synced Tabs</string>
+    <string name="synced_tabs">Synced tabs</string>
     <!-- Browser menu button that opens the find in page menu -->
     <string name="browser_menu_find_in_page">Find in page</string>
     <!-- Browser menu button that creates a private tab -->
@@ -289,7 +289,7 @@
     <!-- Preference for syncing logins -->
     <string name="preferences_sync_logins">Logins</string>
     <!-- Preference for syncing tabs -->
-    <string name="preferences_sync_tabs">Tabs</string>
+    <string name="preferences_sync_tabs_2">Open tabs</string>
     <!-- Preference for signing out -->
     <string name="preferences_sign_out">Sign out</string>
     <!-- Preference displays and allows changing current FxA device name -->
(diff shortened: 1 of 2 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -96,7 +96,7 @@
     <!-- Browser menu toggle that installs a Progressive Web App shortcut to the site on the device home screen. -->
     <string name="browser_menu_install_on_homescreen">Install</string>
     <!-- Menu option on the toolbar that takes you to synced tabs page-->
-    <string name="synced_tabs">Synced Tabs</string>
+    <string name="synced_tabs_2">Synced tabs</string>
     <!-- Browser menu button that opens the find in page menu -->
```
cadeyrn, line 100: Please see https://github.com/mozilla-mobile/fenix/commit/44d9dc9237e054349da832487a3d94384a2663e8#r39455402

> changing capitalization to existing strings doesn't require to use new IDs. The reason is that, by changing the ID, we require all languages to retranslate. But each language has its own capitalization rules, so they should already be using them consistently, independently from what English does.

BranescuMihai: @cadeyrn thanks for the reminder! 
@flodolo if we don't change the ID, so we don't retranslate the strings, there will be languages (for example English CA) that will still keep the wrong capitalisation. So is there a way to say: hey, there are multiple languages that don't respect the correct capitalisation rule, please look over them, without changing the ID?

flodolo: There's no way, but you can ping me and I can fix the English variants directly.

BranescuMihai: got it, thanks!

ValentinTimisica:
@flodolo I changed the strings only for one English variant. From what I understand, after this is merged you will change the other english variants?

flodolo:
> @flodolo I changed the strings only for one English variant. From what I understand, after this is merged you will change the other english variants?

I'll change them right away, so I won't risk forgetting about it.

app/src/main/res/values/strings.xml
```diff
@@ -289,7 +289,7 @@
     <!-- Preference for syncing logins -->
     <string name="preferences_sync_logins">Logins</string>
     <!-- Preference for syncing tabs -->
-    <string name="preferences_sync_tabs">Tabs</string>
+    <string name="preferences_sync_tabs">Open tabs</string>
```
flodolo, line 292: Wait, this one is not OK. You're not just changing capitalization, you need a new ID for this.

ValentinTimisica: Oh right, my bad. I'll change it now.

ValentinTimisica: Done

PR #11231, opened 2020-06-04
Update built-in extensions for new AC/GV API

csadilek:
This will be needed once https://github.com/mozilla-mobile/android-components/pull/7245/files lands in Nightly. We need these new IDs (in this format), and the new permission for the new GV API to install built-in extensions.

I've also noticed that the extensions use the mozac (Mozilla Android Components) namespace, but no need to change that now. We should consider uplifting these extensions some time in the future: https://github.com/mozilla-mobile/android-components/issues/7249

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -100,3 +100,7 @@ test_artifacts/
 /build/test-tools/google-cloud-sdk/
 /build/test-tools/*.jar
 /build/test-tools/*.gz
+
+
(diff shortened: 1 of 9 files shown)


csadilek:
We decided to postpone the API refactoring in A-C for reasons described here: https://github.com/mozilla-mobile/android-components/issues/6356#issuecomment-639822748

Closing this in the meantime.

Issue #11267, opened 2020-06-05
FNX-12167 ⁃ [Bug] Tab Tray A11y: "No Tabs" Message Not Read


darkwing:
From @asadotzler:

> One quick thing I just found. When you have no tabs open in one of the views with TalkBack enabled, the text "Your open tabs will be shown here" isn't read to the user when focusing the empty list area. Instead, "Labels available" is read to the user.
>
> When I swipe past the add tab button to the next focusable element it seems then I get the Your open tabs will be shown here" read.
>
> so that's a bit awkward. I'd expect focus to move from the open tabs button to the private tabs button, to the empty list (where the message of your tabs will be shown here" should be announced) to the plus button and back to the open tabs button.

Additional detail:

> Each swipe moves focus to the next focusable element and reads its label/content. When I open the tabs menu and have no open tabs (in this case I have no open regular tabs but several private tabs) and I have the open tabs button selected, when I swipe to the section where the tabs would be there's a message "your open tabs will be shown here" but that message isn't read. Instead I get a message about using the local context menu to set labels. Swiping two more times, once to the + button and once more puts the focus on the whole tab tray and then the message about open tabs shown here is read.

https://mozilla.slack.com/files/U4CBSAZ5E/F0158F8M4N5/screen_recording_20200604-152928.mp4

sflorean:
Verified as fixed on Nightly 6/9 with Google Pixel 3 (Android 10):  "Your open tabs will be shown here" is read.

PR #11272, opened 2020-06-05
For #11267 - Announce "Your open tabs will be shown here

darkwing:
1.  Enable Talkback
2.  Open the tab tray with no tabs
3.  Swipe right
4.  "Your private tabs will be shown here" should be announced.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/tabtray/TabTrayView.kt
+++ b/app/src/main/java/org/mozilla/fenix/tabtray/TabTrayView.kt
@@ -192,6 +192,7 @@ class TabTrayView(
                 }
             }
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/fenix/tabtray/TabTrayView.kt
```diff
@@ -190,6 +190,9 @@ class TabTrayView(
                 } else {
                     view.context?.getString(R.string.no_open_tabs_description)
                 }
+                view.tabsTray.asView().isVisible = false
```
boek, line 193: might be cleaner to just do this line with `isVisible = hasNoTabs` next to the `tab_tray_empty_view`. Up to you!

PR #11276, opened 2020-06-05
Bug 1642657 - Remove nightly-simulation signing-type specification.

gmierz:
This patch removes the nightly signing-type specification for nightly-simulation.


taskcluster/ci/signing/kind.yml
```diff
@@ -36,10 +36,6 @@ job-template:
                     by-level:
                         '3': fennec-nightly-signing
                         default: dep-signing
-                nightly-simulation:
```
JohanLorenzo, line 39: nightly-sim is gonna be signed with the dummy key, aka the `dep` key so we're good to go!

Issue #11045, opened 2020-05-29
FNX-11904 ⁃ Add Reader Mode only (not Open in App) to address bar


apbitner:
## User Story
As a user, I want to easily know if a site supports reader mode so I have the option to easily switch to reader mode and back, without being distracted by the blue dot every time. 

This features are currently not as easily discoverable in the toolbar menu and there are issues with the current implementation of the blue dot notification.

### Acceptance Criteria
* I am notified if a site supports reader mode without having to open the toolbar menu.
* Reader mode button appears in URL bar when available.
* I don't see the reader mode option in the toolbar menu.
* Telemetry ticket: #11269 11269

## UX
**Specs**: [Abstract](https://share.goabstract.com/ea3f49e5-5ff9-45b4-a44b-3aad13846467)

*Note in the mocks that reader mode appearance settings is being moved to the app settings

Dunexus:
Looking at the abstract, it will reduce the amount of text shown in the address bar.
We already had issues about the address bar not showing enough text, see https://github.com/mozilla-mobile/fenix/issues/3727.

SS1113:
@Dunexus Having the Reader Mode button in the adress bar is intuitive, I can confirm that I have rarely used it or even forgot it's even there. Having it in the address bar will definitely make me use it more instead of having it in the toolbar menu. But I also agree that it'll make text show in the address bar even shorter.

yoasif:
See also https://github.com/mozilla-mobile/fenix/issues/6639

@apbitner Have you given any thought to https://github.com/mozilla-mobile/fenix/issues/7912 ? I think it makes sense to make reader mode appearance options more discoverable - desktop just did a revamp of this as well in https://bugzilla.mozilla.org/show_bug.cgi?id=1550836

cadeyrn:
> Looking at the abstract, it will reduce the amount of text shown in the address bar.
We already had issues about the address bar not showing enough text, see #3727.
Won't this be an issue ?

[...]

PR #11283, opened 2020-06-05
For #11045: Add reader mode to urlView

sblatz:
## Demo

reader mode demo 2020-06-05 11_45_28

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
@@ -10,6 +10,7 @@ import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
 import androidx.coordinatorlayout.widget.CoordinatorLayout
+import androidx.core.content.ContextCompat
 import androidx.lifecycle.Observer
 import androidx.navigation.fragment.findNavController
 import com.google.android.material.snackbar.Snackbar
@@ -17,6 +18,8 @@ import kotlinx.android.synthetic.main.fragment_browser.*
 import kotlinx.android.synthetic.main.fragment_browser.view.*
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import mozilla.components.browser.session.Session
+import mozilla.components.browser.state.selector.findTab
+import mozilla.components.browser.toolbar.BrowserToolbar
 import mozilla.components.feature.app.links.AppLinksUseCases
 import mozilla.components.feature.contextmenu.ContextMenuCandidate
 import mozilla.components.feature.readerview.ReaderViewFeature
@@ -52,6 +55,8 @@ class BrowserFragment : BaseBrowserFragment(), UserInteractionHandler {
     private val windowFeature = ViewBoundFeatureWrapper<WindowFeature>()
     private val searchFeature = ViewBoundFeatureWrapper<SearchFeature>()
 
+    private var readerModeAvailable = false
+
     override fun onCreateView(
         inflater: LayoutInflater,
         container: ViewGroup?,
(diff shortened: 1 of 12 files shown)


gabrielluong:
I can see this being useful on AC and Open In App by having the gradient change on AC. Would you be able to file an issue and upstream this gradient change onto AC as well?

app/src/main/res/values/attrs.xml
```diff
@@ -49,6 +49,9 @@
     <attr name="onboardingDeselected" format="reference"/>
     <attr name="addOnPrivateBrowsingExteriorCircleBackground" format="reference"/>
     <attr name="addOnPrivateBrowsingInteriorIconBackground" format="reference"/>
+    <attr name="readerModeStartGradient" format="reference"/>
+    <attr name="readerModeEndGradient" format="reference"/>
+
```
gabrielluong, line 54: Nit: extra new line here.

app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
```diff
@@ -68,6 +71,25 @@ class BrowserFragment : BaseBrowserFragment(), UserInteractionHandler {
...
         return super.initializeUI(view)?.also {
+            val readerModeAction =
+                BrowserToolbar.ToggleButton(
+                    image = ContextCompat.getDrawable(requireContext(), R.drawable.ic_readermode)!!,
+                    imageSelected = ContextCompat.getDrawable(requireContext(), R.drawable.ic_readermode_selected)!!,
+                    contentDescription = requireContext().getString(R.string.browser_menu_read),
+                    contentDescriptionSelected = requireContext().getString(R.string.browser_menu_read_close),
+                    visible = {
+                        getSessionById()?.let {
+                            activity?.components?.core?.store?.state?.findTab(it.id)?.readerState?.readerable
```
csadilek, line 82: Unfortunately, this isn't sufficient here. Whether or not a page is readerable is determined once the page is loaded and once our built-in web extension (using readability.js) determined that. This may or may not have happened at this point.

In general, I much prefer this solution using the toolbar, as the user will be notified that a page is readerable instead of having to check the menu. Fenix invalidates the menu for that exact reason to make sure it's up-to-date once we know if the page is "readerable": https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt#L82

In R-B and Sample Browser, which also uses a page action in the toolbar we do this: https://github.com/mozilla-mobile/android-components/blob/master/samples/browser/src/main/java/org/mozilla/samples/browser/integration/ReaderViewIntegration.kt#L58

This is important, otherwise we may miss to display the button, although the page is readerable.

csadilek: So it should be sufficient to make sure the button visibility is updated when this callback is invoked: https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt#L78

csadilek: You can see the problem in action on cnbc.com, or another slower site. Click on an article that's readerable (right now "NFL condemns racism..."), the button doesn't appear. Refresh and it does.

sblatz: I've updated this and it should be fixed now 😄

app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
```diff
@@ -68,6 +73,21 @@ class BrowserFragment : BaseBrowserFragment(), UserInteractionHandler {
         val components = context.components
 
         return super.initializeUI(view)?.also {
+            val readerModeAction =
+                BrowserToolbar.ToggleButton(
+                    image = ContextCompat.getDrawable(requireContext(), R.drawable.ic_readermode)!!,
+                    imageSelected = ContextCompat.getDrawable(requireContext(), R.drawable.ic_readermode_selected)!!,
+                    contentDescription = requireContext().getString(R.string.browser_menu_read),
+                    contentDescriptionSelected = requireContext().getString(R.string.browser_menu_read_close),
+                    visible = { readerModeAvailable },
```
csadilek, line 82: Yes, the visibility lambda will work. However, we still need to make sure we invalidate actions when a change happens so they get re-drawn. See, comment below.

app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
```diff
@@ -78,6 +98,8 @@ class BrowserFragment : BaseBrowserFragment(), UserInteractionHandler {
                     if (available) {
                         components.analytics.metrics.track(Event.ReaderModeAvailable)
                     }
+
+                    readerModeAvailable = available
                     runIfFragmentIsAttached {
```
csadilek, line 103: Here's we'd need to call `browserToolbarView.view.invalidateActions()` to make sure the actions update. See also https://github.com/mozilla-mobile/reference-browser/blob/master/app/src/main/java/org/mozilla/reference/browser/browser/ReaderViewIntegration.kt#L67

Issue #11260, opened 2020-06-05
FNX-12158 ⁃ [Bug]Tabs Tray: Crash: java.util.ConcurrentModificationException: at java.util.LinkedHashMap$LinkedHashIterator.nextNode(LinkedHashMap.java)


AndiAJ:
## Prerequisites
Have more then 10 open tabs

## Steps to reproduce
1. Open one of the listed tabs
1. Tap the tab counter
1. Open another tab
1. Put Fenix to background
1. ResumeFenix
1. Open the Tabs Tray
1. Open a tab

### Expected behavior
Properly load the open tabs

### Actual behavior
Crash

### Device information
* Android device: 
• Google Pixel 3a (Android 10)
• OnePlus A3 (Android 6.0.1)
* Fenix version: Nightly Build 200605 from 6/5

### Notes
```
java.util.ConcurrentModificationException
* Sentry: https://sentry.prod.mozaws.net/operations/fenix-nightly/?query=66b16d667e32454eaae509857dee0a11
----
java.util.ConcurrentModificationException
at java.util.LinkedHashMap$LinkedHashIterator.nextEntry(LinkedHashMap.java:346)
at java.util.LinkedHashMap$KeyIterator.next(LinkedHashMap.java:366)
at mozilla.components.support.base.observer.ObserverRegistry.notifyObservers(ObserverRegistry.kt:2)
at mozilla.components.browser.tabstray.TabsAdapter.notifyObservers(TabsAdapter.kt)
at org.mozilla.fenix.tabtray.TabTrayViewHolder$bind$3.onClick(TabTrayViewHolder.kt:1)
at android.view.View.performClick(View.java:5233)
at android.view.View$PerformClick.run(View.java:21209)
at android.os.Handler.handleCallback(Handler.java:739)
at android.os.Handler.dispatchMessage(Handler.java:95)
at android.os.Looper.loop(Looper.java:152)
at android.app.ActivityThread.main(ActivityThread.java:5507)
at java.lang.reflect.Method.invoke(Native Method)
at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:726)
at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:616)

```

Crash report from Google Pixel 3a : https://crash-stats.mozilla.org/report/index/991f1e5b-87c3-4181-88c5-072bb0200605

► Video
[...]

PR #11317, opened 2020-06-07
For #11260 - Fixes ConcurrentModificationException on tabs tray

ekager:
### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/tabtray/TabTrayDialogFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/tabtray/TabTrayDialogFragment.kt
@@ -146,7 +146,9 @@ class TabTrayDialogFragment : AppCompatDialogFragment(), TabTrayInteractor {
     }
 
     override fun onTabSelected(tab: Tab) {
-        dismissAllowingStateLoss()
+        // dismiss the dialog on the next pass as it may cause a ConcurrentModificationException
+        // when TabsTrayInteractor.stop() unregisters with ObserverRegistry
+        view?.post { dismissAllowingStateLoss() }
         if (findNavController().currentDestination?.id == R.id.browserFragment) return
         if (!findNavController().popBackStack(R.id.browserFragment, false)) {
             findNavController().navigate(R.id.browserFragment)


ebarrenechea:
@ekager I think there's an issue with this change. This morning I got a crash when switching to a private browsing tab when opening a link in private mode. I was able to reproduce the issue and the stack trace points to the changes I made yesterday. IMO it's a worse bug than the one being fixed here and we should address it before getting this merged.

Steps to reproduce:
1. open link in new private tab
2. do _not_ switch to tab using the snackbar action
3. open tab tray and switch to the private tab

I haven't had the time to look into it ion detail but I think it's related to the dialog being detached by a different route (e.g. navigation component). The crash can be avoided by checking `isAdded` before dismissing the dialog, but that results in the tab tray opening after switching between private and standard browsing tabs.

I'll have a better look at whenever I have some free time today. Here's a copy of the stack trace.
```
java.lang.IllegalStateException: Fragment TabTrayDialogFragment{f63ff68} (76f94413-055e-4eeb-a136-63ca0308e02d)} not associated with a fragment manager.
    at androidx.fragment.app.Fragment.getParentFragmentManager(Fragment.java:945)
    at androidx.fragment.app.DialogFragment.dismissInternal(DialogFragment.java:271)
    at androidx.fragment.app.DialogFragment.dismissAllowingStateLoss(DialogFragment.java:238)
    at org.mozilla.fenix.tabtray.TabTrayDialogFragment$onTabSelected$1.run(TabTrayDialogFragment.kt:151)
    at android.os.Handler.handleCallback(Handler.java:883)
    at android.os.Handler.dispatchMessage(Handler.java:100)
    at android.os.Looper.loop(Looper.java:214)
    at android.app.ActivityThread.main(ActivityThread.java:7356)
    at java.lang.reflect.Method.invoke(Native Method)
    at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:492)
    at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:930)
```

ebarrenechea:
@ekager I investigated the issue in more detail and I didn't know we're recreating `HomeActivity` when we switch between normal and private browsing modes. It happens before we have the chance to dismiss the tabs tray dialog. Dismissing the dialog before we restart the activity looks like a dead end.

This morning I saw that you and @boek have a branch with changes where you manage the view and the feature lifecycle manually in the dialog. This will probably work but I'd also suggest a solution where we change how `ObserverRegistry` handles `notifyObservers`. I think it would be better for it to operate on a copy of the `observers` set instead of directly on the set. This will allow modifications to the original set while the registry can safely continue to notify the observers.

Something below will avoid the `ConcurrentModificationException` that is the original issue we're trying to solve and shouldn't (hopefully) cause problems elsewhere.
``` 
observers.toList().forEach {
    if (!pausedObservers.contains(it)) {
        it.block()
    }
}
```

In any case, go with whatever you think is the right approach here as you have a lot more context and understanding on this codebase. Thanks for you efforts in fixing this issue. :+1:

PR #11334, opened 2020-06-08
Add tests for ext package

NotWoods:
Add in tests for the View extensions, NavController extensions, Log methods

### Pull Request checklist


app/src/main/java/org/mozilla/fenix/ext/View.kt
```diff
@@ -15,10 +15,7 @@ fun View.increaseTapArea(extraDps: Int) {
     parent.post {
         val touchRect = Rect()
         getHitRect(touchRect)
-        touchRect.top -= dips
-        touchRect.left -= dips
-        touchRect.right += dips
-        touchRect.bottom += dips
+        touchRect.inset(-dips, -dips)
```
ekager, line 18: it looks like some of them are adding? am i misunderstanding what the inset fun is doing?

PR #11344, opened 2020-06-08
For #11342 - Fixes missing space for the FAB

### LINKED ISSUE #11342: FNX-12255 ⁃ [Bug] FAB covers last tab in tab tray

boek:
Fixes #11342

image

Also fixes top offset in landscape

image


app/src/main/java/org/mozilla/fenix/tabtray/TabTrayDialogFragment.kt
```diff
@@ -57,6 +57,9 @@ class TabTrayDialogFragment : AppCompatDialogFragment(), TabTrayInteractor {
     override fun onConfigurationChanged(newConfig: Configuration) {
         super.onConfigurationChanged(newConfig)
 
+        val isLandscape = newConfig.orientation == Configuration.ORIENTATION_LANDSCAPE
```
ekager, line 60: could we combine this with the next check?

boek: doh, I meant to do that! Good catch!

PR #11351, opened 2020-06-09
Add tests for Metrics

NotWoods:
Additionally breaks up the giant Metrics file, but keeps refactoring minimal beyond moving code into different files.

### Pull Request checklist

### DIFF
--- a/app/src/test/java/org/mozilla/fenix/components/metrics/MetricControllerTest.kt
+++ b/app/src/test/java/org/mozilla/fenix/components/metrics/MetricControllerTest.kt
@@ -1,0 +1,170 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.components.metrics
+
+import io.mockk.MockKAnnotations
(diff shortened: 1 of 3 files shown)


app/src/test/java/org/mozilla/fenix/components/metrics/MetricControllerTest.kt
```diff
@@ -0,0 +1,113 @@
...
+        every { dataService.shouldTrack(Event.TabMediaPause) } returns false
+        every { marketingService.shouldTrack(Event.TabMediaPause) } returns true
+
+        controller.start(MetricServiceType.Marketing)
+        controller.track(Event.TabMediaPause)
+        verify { marketingService.track(Event.TabMediaPause) }
+    }
+
+    @Test
+    fun `tracking terminates if other service is not enabled`() {
```
NotWoods, line 100: Is this intended behaviour? Tracking is stopped if there are any disabled services in the chain, instead of going to the first available service. (See also `MetricController.track`)

ekager: cc @boek do you know about this?

boek: Ah, I think we wanted a `return@forEach` there.

Issue #2165, opened 2019-04-30
FNX3-16111 ⁃ Pull-down bookmarks/history lists to trigger sync


sv-ohorvath:
### Why/User Benefit/User Problem
It would be a simple, fast way to bring new synced items, without the need to navigate back & forth between account settings and library.

### What/Requirements
Pull-down to refresh and sync bookmarks & history.
Or, add a button in Bookmarks & history menus for that.

ayushi1998:
Hey, I am an outreachy applicant. Can I work on this? How to go about the same? New to open source

person808:
Gonna write down a couple of potential edge cases that I think I need UX feedback on.

Given this bookmark structure:

```
Mobile Bookmarks
|__A
   |__B
```

Edge cases:
1. If I'm in folder B, delete folder B on another device, sync that device, then trigger sync while in folder B, we end up viewing a folder that no longer exists, causing a crash.
2. If I'm in folder A, delete folder B on another device, sync that device, trigger a sync while in folder A, and navigate to folder B, then we are viewing a folder that no longer exists. I haven't tested yet, but this probably causes a crash.

Some potential solutions for 1:
1. Only allow the pull-to-refresh gesture when viewing folders that we know can't be deleted (ie Desktop Bookmarks, Booksmarks Toolbar, etc.) This is easy to implement but is a bit inconsistent.
2. If we sync and end up viewing an invalid folder, return to the top level bookmarks folder (Mobile Bookmarks). Harder to implement than 1, but I think still doable. Probably would need to show some kind of message indicating that the folder was removed?
3. Always go back to the top level bookmarks folder after syncing. Similar difficulty to implement as 2. Consistent behavior, but what if a user just wants to see a new bookmark from another device in a folder? Seems like having to navigate back to the folder would be annoying.

Some potential solutions for 2:
1. Cancel sync when opening a folder.
2. Implement one of solutions 2 or 3 from above.

Currently I have solution 1 implement for both problems as those are the easiest to implement but I believe the other solutions wouldn't be too much trouble. The only problem with the other solutions is the changes needed break both the back and up buttons so they would immediately go back to the previous screen instead of navigating up the bookmark tree. It looks like the back button is easy to fix, but I'm not sure about the up button.

person808:
[...]

PR #11386, opened 2020-06-09
For #2165 - Implement pull-to-refresh gesture to sync history.

person808:
Note: This is only 1/2 gestures described in #2165 

### Pull Request checklist

output

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/test/java/org/mozilla/fenix/library/history/HistoryControllerTest.kt
+++ b/app/src/test/java/org/mozilla/fenix/library/history/HistoryControllerTest.kt
@@ -9,10 +9,14 @@ import android.content.ClipboardManager
 import android.content.res.Resources
 import androidx.navigation.NavController
 import androidx.navigation.NavDirections
+import io.mockk.coVerifyOrder
 import io.mockk.every
 import io.mockk.mockk
 import io.mockk.slot
 import io.mockk.verify
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.test.TestCoroutineScope
 import mozilla.components.concept.engine.prompt.ShareData
 import org.junit.Assert.assertEquals
 import org.junit.Assert.assertFalse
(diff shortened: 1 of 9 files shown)


app/src/test/java/org/mozilla/fenix/library/history/HistoryControllerTest.kt
```diff
@@ -189,4 +202,13 @@ class HistoryControllerTest {
         assertEquals(historyItem.title, (directions.captured.arguments["data"] as Array<ShareData>)[0].title)
         assertEquals(historyItem.url, (directions.captured.arguments["data"] as Array<ShareData>)[0].url)
     }
+
+    @Test
+    fun onRequestSync() {
+        controller.handleRequestSync()
+
+        verify {
+            syncHistory.invoke(any())
```
Mugurell, line 211: I think we should also verify here that the right actions are emitted.
`StartSync` is easy to be verified, but verifying that syncHistory gets passed the `FinishSync` callback is a bit trickier.. maybe we could return that store.dispatch call from a method and then in the test verify if 
- `store.dispatch(HistoryFragmentAction.StartSync)` was called
- `store.dispatch(any())` was called 2 times
- the method returning `store.dispatch(HistoryFragmentAction.FinishSync)` was called ?

person808:
@NotWoods Converted `syncHistory` to a suspend function and updated the test to verify actions are dispatched in the right order, but the test fails with [this](https://firefox-ci-tc.services.mozilla.com/tasks/QZTbjKXcSb2jkgTZMjWyTQ/runs/0/logs/https%3A%2F%2Ffirefox-ci-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FQZTbjKXcSb2jkgTZMjWyTQ%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive.log#L4703) exception. Could you take a look and see what I'm missing to make the test run properly?

NotWoods:
You need to use a real coroutine scope instead of a mock, like this:
```kt
private val scope = TestCoroutineScope()
```
You'll also need to add an `@ExperimentalCoroutinesApi` annotation above the `@RunWith` annotation.

Issue #9029, opened 2020-03-06
[Telemetry] Additional Collections telemetry


ecsmyth:
Owner: Product Manager

#### Description & Product Manager / Data Scientist User Story
- As the performance product manager, I want to know how many collections users have and the numbers of tabs saved to the collections, so that we can test raw and perceived performance under representative conditions.
- As a product manager, ..., so that we can improve the UX based on how users use the feature.

#### Hypothesis
- The current UI may result in poor performance and a frustrating experience with a large number of collections or tabs saved per collection.

#### What questions will you answer with this data?
- What variations of our responsiveness tests do we need to ensure we're testing realistic user flows?

#### Why does Mozilla need to answer these questions? Are there benefits for users? Do we need this information to address product or business requirements?
- User research and exit surveys regularly point to startup time as a point of friction leading to churn. Improving the realism of our test automation improves our ability to detect user-meaningful regressions and make the improvements that matter most to our users.

#### What probes (suggested, if applicable)
- Count of collections
- Histogram of tabs per collection

### Acceptance Criteria (Added by PM)
- Event pings can be queried via re:dash
- We are sending telemetry events for the actions listed in the requirements
- We have documented the telemetry
- We have asked a data steward to [review](https://github.com/mozilla/data-review/blob/master/request.md) the telemetry

sraturi:
Blocked by: https://bugzilla.mozilla.org/show_bug.cgi?id=1648258

mcomella:
Upstream issue was closed as wontfix: sending back to triage.

PR #11396, opened 2020-06-09
#9029 Add telemetry to collect total number of collections

sraturi:
### TLDR

Added `counter` metrics to record the total number of collections on startup. The metric `lifecycle` is the `application`, meaning every time the app restarts, the counter value is reset. This ensures that we are sending only the latest total number of collections. The counter will not report the metric if total number of collections is 0. Therefore, we assume whenever this metric is missing that collection count is 0. The ping used in this metric is the `metrics` ping type which only gets sent once a day.

### Pull Request checklist

### After merge

1. click on Show All Checks,


sraturi:
I switched the metric type from `counter` to `quantity`. Since user may create or delete multiple collections before a particular ping is sent, we don't want to keep adding to the counter.  For example, the `event` ping is sent when app is in the background or the total number of events has reached a certain number (500?)

app/metrics.yaml
```diff
@@ -1562,6 +1562,21 @@ collections:
     notification_emails:
       - fenix-core@mozilla.com
     expires: "2020-09-01"
+  total_collections:
+    type: quantity
```
mcomella, line 1566: The docs say,

> Note: Quantities are currently only allowed for GeckoView metrics (the gecko_datapoint parameter is present) and thus have only a Kotlin API.

Did you verify with the glean team that this was okay?

app/src/main/java/org/mozilla/fenix/components/TabCollectionStorage.kt
```diff
@@ -59,6 +60,7 @@ class TabCollectionStorage(
     fun createCollection(title: String, sessions: List<Session>) {
         collectionStorage.createCollection(title, sessions)
         notifyObservers { onCollectionCreated(title, sessions) }
+        Collections.totalCollections.set(getTabCollectionsCount().toLong())
```
mcomella, line 63: Did we want a histogram with number of total collections and pages per collection?

mcomella, line 63: Why do we do this each time a collection is created or removed rather than on startup? It could be valuable to comment because other people may have the same questions I have.

app/src/main/java/org/mozilla/fenix/components/TabCollectionStorage.kt
```diff
@@ -80,6 +82,7 @@ class TabCollectionStorage(
 
     fun removeCollection(tabCollection: TabCollection) {
         collectionStorage.removeCollection(tabCollection)
+        Collections.totalCollections.set(getTabCollectionsCount().toLong())
```
mcomella, line 85: `getTabCollectionsCount()` seems to touch the DB (but I don't know how Room works) and I don't like that performance hit if so – is Room actually pulling it from in memory or is there an in-memory representation we can use instead?

app/src/main/java/org/mozilla/fenix/components/metrics/Metrics.kt
```diff
@@ -547,7 +547,15 @@ interface MetricController {
                     isDataTelemetryEnabled,
                     isMarketingDataTelemetryEnabled
                 )
-            } else DebugMetricController()
+            } else {
+                // todo switch it back to normal
+                ReleaseMetricController(
```
mcomella, line 552: No change needed: for future reference, I usually change this one line to enable telemetry for debug builds: https://searchfox.org/mozilla-mobile/rev/58d77b139e5e992a204a0c6b1f40120344f4820d/fenix/app/build.gradle#342

app/src/main/java/org/mozilla/fenix/components/TabCollectionStorage.kt
sraturi: I think we are omitting the histogram for now? @ecsmyth might be able to confirm this?

app/src/main/java/org/mozilla/fenix/components/TabCollectionStorage.kt
sraturi: Seems like its querying the DB each time. I will look for an alternate

app/src/main/java/org/mozilla/fenix/components/metrics/Metrics.kt
sraturi: Thanks!

app/src/main/java/org/mozilla/fenix/components/TabCollectionStorage.kt
sraturi: Initially, I had it at startup but changed it  so that we collect more accurate information. But I am thinking about changing it back to startup again given its a startup telemetry ticket?

app/metrics.yaml
sraturi: Ah I figured that was a note as to why api is not available in other languages.  I will double check it the Glean team

Dexterp37: Yup, @mcomella instinct is correct :-D This should not be used!

app/metrics.yaml
```diff
@@ -1562,6 +1562,21 @@ collections:
...
+    unit: unit
+    gecko_datapoint: NA
```
Dexterp37, line 1568: This is the reason why the build is not failing: you're providing a bogus `gecko_datapoint` but we have no way to tell that this is not correct :)

sraturi: Thanks for clearing up the confusion! 😄

app/metrics.yaml
sraturi: I will change it up.

liuche:
When this is ready, please also include a data review request, and add a "data-review" label so it can be addressed!
https://github.com/mozilla/data-review/blob/master/request.md

app/metrics.yaml
```diff
@@ -1562,6 +1562,21 @@ collections:
...
+    send_in_pings:
+      - events
```
Dexterp37, line 1570: @mcomella after talking to @sraturi , he made me notice that this would be sent in the `events` ping. Why is that?

We also discussed a bit about using the `counter` metric type for this. I believe that should work, even with the restriction related to not being able to set "0": you would only set the value if there were collections. If no collections were available, you would not set the counter. This means that, for versions of Fenix that have to have the metric there, you can assume the "metric being missing" = 0.

mcomella: > this would be sent in the events ping. Why is that?

@Dexterp37 It's a general use metric so I wanted to include it in the most general ping. I thought that ping was events. My understanding is that Metrics is used for data that doesn't change much (e.g. OS version) and nothing else seemed general use.

> you can assume the "metric being missing" = 0.

If this allows us to determine the number of users with collection 0 then that seems okay to me. I'm concerned that:
- this doesn't seem semantically meaningful
- it forces logic onto the server and the analysis phase while other glean implementations seem to have logic on the client (e.g. timing distribution). I guess the major downside is we can't analyze this with GLAM (when it's available)?

I'm not up-to-date on your discuss with sraturi so I'm not sure if the ideal solution, long term is to create a new metric type. If not, I'm concerned that we won't be the only folks who run into this sort of problem (want to send a non-counting integer where 0 is meaningful) so I feel like we should create a new metric type but, disclaimer, I don't have much context on the existing discussions. :)

Dexterp37: > > this would be sent in the events ping. Why is that?
> 
> @Dexterp37 It's a general use metric so I wanted to include it in the most general ping.

That's really the [metrics ping](https://mozilla.github.io/glean/book/user/pings/metrics.html).

> I thought that ping was events. My understanding is that Metrics is used for data that doesn't change much (e.g. OS version) and nothing else seemed general use.

Oh, that's interesting. Was it something wrong in our docs that led to this misunderstanding? The metrics ping is meant for all kinds of data, regardless of their frequency of change. In fact, if you don't specify which pings to send stuff on, all the metrics (except the events) go in that ping!

> > you can assume the "metric being missing" = 0.
> 
> If this allows us to determine the number of users with collection 0 then that seems okay to me. I'm concerned that:
> 
>     * this doesn't seem semantically meaningful

@sraturi filed a bug for using `quantity` outside of Geckoview. In addition to that, @grigoryk filed [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1626372) for a similar issue.

This will be looked into, as the process for adding/changing metric type moves on.
Unfortunately, since we only recently introduced the process, we're still going through the backlog, but we're planning on getting to this issue in the upcoming weeks.

>     * it forces logic onto the server and the analysis phase while other glean implementations seem to have logic on the client (e.g. timing distribution). I guess the major downside is we can't analyze this with GLAM (when it's available)?

This is a good point. This is definitely one of the aspects that will be cleared when going through the new metric type process. Depending on how soon you want the data, you may still stick to counter. 

> I'm not up-to-date on your discuss with sraturi so I'm not sure if the ideal solution, long term is to create a new metric type. If not, I'm concerned that we won't be the only folks who run into this sort of problem (want to send a non-counting integer where 0 is meaningful) so I feel like we should create a new metric type but, disclaimer, I don't have much context on the existing discussions. :)

Sure, filing a bug was a good call! And making sure folks don't stumble into the same problems is the reason why we're building Glean in the first place :)

mcomella: > Oh, that's interesting. Was it something wrong in our docs that led to this misunderstanding?

I showed up to a presentation late that talked about the differences in the pings and that's the impression that I got but I will admit that I wasn't paying full attention. 😞

app/src/main/java/org/mozilla/fenix/components/TabCollectionStorage.kt
sraturi: Added  an observer to report only once the data is in memory

app/src/main/java/org/mozilla/fenix/components/metrics/Metrics.kt
```diff
@@ -557,7 +557,15 @@ interface MetricController {
...
```
mcomella, line 562: We should revert this part before landing.

fwiw, you can do `git add <path-to-file>` before committing to prevent adding changes like this or, if you want more granularity than per file, `git add -p <path-to-file>` to be able to select individual lines to add.

I can also help you undo this change using git commands if you'd like!

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -732,6 +735,11 @@ class HomeFragment : Fragment() {
             requireComponents.core.tabCollectionStorage.getCollections().observe(this, observer)
         }
     }
+    private fun singleUseCollectionsTelemetryObserver() {
```
mcomella, line 738: nit: whitespace above this function

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -161,6 +163,7 @@ class HomeFragment : Fragment() {
                 requireComponents.analytics.metrics.track(Event.OpenedAppFirstRun)
             }
         }
+        singleUseCollectionsTelemetryObserver()
```
mcomella, line 166: I'd prefer this to be in HomeActivity: the lifecycle of fragments is complicated and hard to predict so moving it into HomeActivity gives more me more confidence of correctness.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -732,6 +735,11 @@ class HomeFragment : Fragment() {
             requireComponents.core.tabCollectionStorage.getCollections().observe(this, observer)
         }
     }
+    private fun singleUseCollectionsTelemetryObserver() {
+        requireComponents.core.tabCollectionStorage.getCollections().observeOnce {
+            Collections.totalCollections.add(it.size)
```
mcomella, line 740: Since the metrics ping is submitted once a day and this can be called more than once a day, I'm concerned that we'll add together the current collections count multiple times a day, e.g. if the app is started at 5pm and 6pm and the user has 5 collections, we'd get 5 + 5 = 10 instead of 5.

I think we either need to:
- submit the ping each time this metric is collected (in which case, maybe it should be a part of `startup-timeline` because it's transmitted once per start/stop cycle; though I'm not sure what happens if the application crashes - maybe this strategy won't work then)
- add code to ensure we only call this once per ping lifecycle (e.g. once a day for metrics) though this seems a bit tricky too...

@Dexterp37 Is that right or am I misunderstanding how counters work?

sraturi: I set the life cycle of the metric to be `application`. With this, I believe the metric gets reset every time the app restarts. For example, if we start the app at 5 pm with 6 collections and then again at 6 pm with 5 collections. The metric will only report 5 (which is the latest). Unless I missunderstood the usecase for lifecycle @Dexterp37 ?

app/src/main/java/org/mozilla/fenix/components/metrics/Metrics.kt
sraturi: Ah, thanks! I think I figured it out now. Didn't know about the -p option that's very neat!

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
sraturi: fixed it.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
sraturi: done!

app/metrics.yaml
```diff
@@ -1608,6 +1608,21 @@ collections:
...
+    send_in_pings:
+      - metrics
+    lifetime: application
+    description: |
+      reports the total number of collections user has on startup.
+      If no value is set, we assume total number of collections is 0.
+    bugs:
+      - https://github.com/mozilla-mobile/fenix/issues/9029
+    data_reviews:
+      - N/A
```
Dexterp37, line 1622: Don't forget to ask for data review before merging this. This is a blocker :)

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
Dexterp37: > 
> 
> Since the metrics ping is submitted once a day and this can be called more than once a day, 

Does this get called multiple times during a single Fenix run?

> I'm concerned that we'll add together the current collections count multiple times a day, e.g. if the app is started at 5pm and 6pm and the user has 5 collections, we'd get 5 + 5 = 10 instead of 5.

@sraturi set the correct `lifetime` for this, meaning that next time Fenix starts the metric will be cleared. So, if this function gets called multiple time during a Fenix run, then your concern is valid. Otherwise, if it only runs once, you should not be worried :)

> I think we either need to:
> 
>     * submit the ping each time this metric is collected (in which case, maybe it should be a part of `startup-timeline` because it's transmitted once per start/stop cycle; though I'm not sure what happens if the application crashes - maybe this strategy won't work then)
> 
>     * add code to ensure we only call this once per ping lifecycle (e.g. once a day for metrics) though this seems a bit tricky too...
> 
> 
> @Dexterp37 Is that right or am I misunderstanding how counters work?

In general, you should not really care about _when_ built-in pings are sent. This is opaque for this exact reason: the ping scheduler dictates when the "snapshot" of the current values is taken. Governing when things are cleared is done through [lifetimes](https://mozilla.github.io/glean/book/user/metric-parameters.html#optional-metric-parameters).

> 
> 
> I set the `lifetime` of the metric to be `application`. With this, I believe the counter gets reset every time the app restarts. For example, if we start the app at 5 pm with 6 collections and then again at 6 pm with 5 collections. The metric will only report 5 (which is the latest). Unless I misunderstood the use case for lifecycle @Dexterp37 ?

Yes, that's true @sraturi :)

sraturi:
Closing this PR since we decided that we do not have the correct metric type.  we will wait until the correct metric type is available to implement this telemetry.

PR #11441, opened 2020-06-10
For #11291 - Provide New Tab button when accessibility enabled

### LINKED ISSUE #11291: FNX-12196 ⁃ [Bug] Tab Tray A11y:  FAB isn't focusable by TalkBack when Tab Tray is Extended
## Steps to reproduce
1. Turn on talkback
1. Open the tab tray to extended mode
1. Try to tap the FAB -- doesn't get selected
1. Gesture right until you get to the close button of the last tab -- FAB doesn't get selected

### Expected behavior
FAB should get selected when tray is in extended mode

### Actual behavior
FAB DOES NOT get selected when tray is in extended mode

### Device information
* Android device: Pixel 4
* Fenix version: Current master

darkwing:
Fixes #11291. 

This provides an "Add tab" button at the header of the tab tray so that users with accessibility features enabled can more easily add a tab instead of the FAB.  Even better is that the user wouldn't need to scroll through all tabs to get to the "New tab" FAB -- it would be the third control available!

Screenshot_20200610-124510

To test:
1.  Open the tab tray with talkback off
2. View FAB, don't view "add tab" button in header
3.  Turn on Talkback
4.  Open the tab tray
5.  View "add tab" button, don't view FAB

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/tabtray/TabTrayView.kt
+++ b/app/src/main/java/org/mozilla/fenix/tabtray/TabTrayView.kt
@@ -25,6 +25,7 @@ import mozilla.components.browser.state.state.TabSessionState
 import mozilla.components.browser.tabstray.BrowserTabsTray
 import org.mozilla.fenix.R
 import org.mozilla.fenix.ext.components
+import org.mozilla.fenix.ext.settings
 
 interface TabTrayInteractor {
     fun onNewTabTapped(private: Boolean)
@@ -61,14 +62,18 @@ class TabTrayView(
         get() = container
 
     init {
+        val hasAccessibilityEnabled = view.context.settings().accessibilityServicesEnabled
+
         toggleFabText(isPrivate)
 
         behavior.addBottomSheetCallback(object : BottomSheetBehavior.BottomSheetCallback() {
             override fun onSlide(bottomSheet: View, slideOffset: Float) {
-                if (slideOffset >= SLIDE_OFFSET) {
-                    fabView.new_tab_button.show()
-                } else {
-                    fabView.new_tab_button.hide()
+                if (!hasAccessibilityEnabled) {
+                    if (slideOffset >= SLIDE_OFFSET) {
+                        fabView.new_tab_button.show()
+                    } else {
+                        fabView.new_tab_button.hide()
+                    }
                 }
             }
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/fenix/tabtray/TabTrayView.kt
```diff
@@ -142,8 +143,30 @@ class TabTrayView(
                 }
         }
 
-        fabView.new_tab_button.setOnClickListener {
-            interactor.onNewTabTapped(isPrivateModeSelected)
+        val accessibilityManager =
+            view.context.applicationContext.getSystemService(Context.ACCESSIBILITY_SERVICE) as? AccessibilityManager
+        val hasAccessibilityEnabled = accessibilityManager?.isEnabled ?: false
```
darkwing, line 148: @sblatz Is this block sufficient to know if the user is using talkback or another a11y service?  Anything more I should do or check?

darkwing:
> I don't have TalkBack enabled but see the +- icon next to the 3-dot menu

Ohhh, the FAB still shows when in expanded mode.  OK, I'll fix that!

app/src/main/java/org/mozilla/fenix/tabtray/TabTrayView.kt
sblatz: You should check against the settings param: `accessibilityServicesEnabled` it checks for both switch services and touch exploration services :)

app/src/main/java/org/mozilla/fenix/tabtray/TabTrayView.kt
```diff
@@ -142,8 +147,26 @@ class TabTrayView(
                 }
         }
 
-        fabView.new_tab_button.setOnClickListener {
-            interactor.onNewTabTapped(isPrivateModeSelected)
+        view.tab_tray_new_tab.apply {
+            visibility = if (hasAccessibilityEnabled) {
```
boek, line 151: Nit: use isVisivible

app/src/main/java/org/mozilla/fenix/tabtray/TabTrayView.kt
```diff
@@ -142,8 +147,26 @@ class TabTrayView(
...
+            } else {
+                View.GONE
+            }
+            setOnClickListener {
+                interactor.onNewTabTapped(isPrivateModeSelected)
+            }
+        }
+
+        fabView.new_tab_button.apply {
+            visibility = if (hasAccessibilityEnabled) {
```
boek, line 162: Same as above

PR #11487, opened 2020-06-11
For 9640 - Focus on last selected tab in tab tray with a11y turned on

darkwing:
To test:

1. Turn on talkback
2. Open a few tabs in Fenix
3. Open tab tray

Each time you open the tab tray the current tab should be selected

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/tabtray/TabTrayView.kt
+++ b/app/src/main/java/org/mozilla/fenix/tabtray/TabTrayView.kt
@@ -8,6 +8,7 @@ import android.content.Context
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
+import android.view.accessibility.AccessibilityEvent
 import androidx.cardview.widget.CardView
 import androidx.core.content.ContextCompat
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mozilla/fenix/tabtray/TabTrayView.kt
```diff
@@ -113,6 +119,19 @@ class TabTrayView(
                     if (!hasLoaded) {
                         hasLoaded = true
                         tray.layoutManager?.scrollToPosition(selectedBrowserTabIndex)
+                        if (view.context.settings().accessibilityServicesEnabled) {
+                            GlobalScope.launch {
+                                delay(1000)
```
darkwing, line 124: This is obviously not ideal but I'm at a loss.  This works with over 100 tabs on my Pixel 4, but I don't have an older device to test, and arbitrary delays aren't ideal.

boek: We should probably use the view lifecycle scope to make sure this gets properly cancelled if you quickly back out

Issue #5772, opened 2019-10-03
Automatically switch to PWA as long as web app is installed


NotWoods:
### Why/User Benefit/User Problem
When we navigate to the Twitter website, if the Twitter Android app is installed we can be prompted to open in app instead. It would be nice to open in PWA if installed.

### What/Requirements
As long as a PWA is installed, automatically open the session in a PWA activity - similar to opening a custom tab in the full browser.

### Acceptance Criteria (how do I know when I’m done?)
PWAs automatically open corresponding links after installation.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-13747)

TitanNano:
It would be great to be able to choose, though. I sometimes find myself wanting to open multiple instances if the same PWA.

TitanNano:
Copy of the duplicate I filled recently:

> ## Why/User Benefit/User Problem
> 
> Similar to native apps, users want to view links which point to content in the scope of a standalone PWA inside standalone mode and not the browser.
> ## What/Requirements
> 
> Launch the PWA activity when the user navigates to a URL which falls into the scope of a PWA. The best solution would be if this would behave like it does with native apps. (same settings entries and main menu options).
> ## Acceptance Criteria (how do I know when I’m done?)
> There is a way for users to automatically transition a PWA into standalone mode.

TitanNano:
@NotWoods would it be possible for me to work on this?

NotWoods:
We're not certain if we want to implement this feature yet. 

However, it shares dependencies with some other features we do want. We want to be able to open a PWA for some scope if its installed. For example, notifications should open in the PWA. You're welcome to work on that component if you would like 😄

TitanNano:
@NotWoods yes sure, can you point me to the right issue?

NotWoods:
https://github.com/mozilla-mobile/android-components/issues/5667

vesta0:
@NotWoods FYI

liuche:
@NotWoods can you give an estimate on how much work this is? Or point to where someone might look to implement this? We should include this in release, and having some more steps would make it easier for people to pick up.

TitanNano:
@liuche @NotWoods I'd love to work on this, if it's available.

NotWoods:
[...]

PR #11543, opened 2020-06-12
Automatically switch to PWA when web app is installed

TitanNano:
For #5772 and requires mozilla-mobile/android-components#7367 and mozilla-mobile/android-components#7381

- android components has a new web app interceptor which needs to be integrated by Fenix.
- `ExternalAppBrowserActivity` should not navigate to `externalAppBrowserFragment` if it's already there.
- `ExternalAppBrowserActivity` has to release its session`onPause`, because it's possible that it will be replaced by a new activity. This happens when the activity is launched both from the android launcher and from inside Fenix.

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
+++ b/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
@@ -19,6 +19,7 @@ import androidx.appcompat.app.ActionBar
 import androidx.appcompat.widget.Toolbar
 import androidx.core.view.doOnPreDraw
 import androidx.lifecycle.lifecycleScope
+import androidx.navigation.NavController
 import androidx.navigation.NavDestination
 import androidx.navigation.NavDirections
 import androidx.navigation.fragment.NavHostFragment
@@ -114,6 +115,9 @@ open class HomeActivity : LocaleAwareAppCompatActivity() {
         supportFragmentManager.findFragmentById(R.id.container) as NavHostFragment
     }
 
+    protected val navController: NavController
+        get() = navHost.navController
+
     private val externalSourceIntentProcessors by lazy {
         listOf(
             SpeechProcessingIntentProcessor(this, components.analytics.metrics),
@@ -186,7 +190,7 @@ open class HomeActivity : LocaleAwareAppCompatActivity() {
         }
     }
 
-    final override fun onPause() {
+    override fun onPause() {
         if (settings().lastKnownMode.isPrivate) {
             window.addFlags(WindowManager.LayoutParams.FLAG_SECURE)
         } else {
--- a/app/src/main/java/org/mozilla/fenix/customtabs/ExternalAppBrowserActivity.kt
+++ b/app/src/main/java/org/mozilla/fenix/customtabs/ExternalAppBrowserActivity.kt
@@ -14,7 +14,9 @@ import mozilla.components.support.utils.SafeIntent
 import org.mozilla.fenix.BrowserDirection
 import org.mozilla.fenix.HomeActivity
 import org.mozilla.fenix.NavGraphDirections
+import org.mozilla.fenix.R
 import org.mozilla.fenix.components.metrics.Event
+import org.mozilla.fenix.ext.alreadyOnDestination
 import org.mozilla.fenix.ext.components
 import java.security.InvalidParameterException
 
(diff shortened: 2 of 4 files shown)


app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -110,7 +110,7 @@ open class HomeActivity : LocaleAwareAppCompatActivity() {
         WebExtensionPopupFeature(components.core.store, ::openPopup)
     }
 
-    private val navHost by lazy {
+    protected val navHost by lazy {
```
NotWoods, line 113: Let's expose the nav controller instead of the nav host.

app/src/main/java/org/mozilla/fenix/customtabs/ExternalAppBrowserActivity.kt
```diff
@@ -57,6 +65,12 @@ open class ExternalAppBrowserActivity : HomeActivity() {
         }
     }
 
+    override fun onPause() {
+        super.onPause()
+
+        engineView.release()
```
NotWoods, line 71: Does this need to be in `onPause` instead of `onStop` or `onDestroy`? I'm concerned about cases like opening the notification tray, or focusing on another app in multi-window mode. If this activity is resumed, do we recapture the engine view?

TitanNano: In my tests `onStop` was not called before `onStart` was called in the new activity, this is the root of the issue, because the session would be released `onStop`. But engine view will automatically be recovered `onStart` and `onResume`. Notification drawer works without issues. Split screen also mostly works, only if I go to the homecreen while in split screen and then return to the split screen view, the WebApp is blank. I will look at this again.

TitanNano: I tested this again and `onStop` does really fit better here. Only when I put a PWA in split screen, go to the launcher and then launch the same PWA again, Fenix crashes. It appears android is not calling `onStop` in that particular situation.

TitanNano: @NotWoods I think this is best solved by releasing the session through it's observer during the activity creation. I implemented this here https://github.com/mozilla-mobile/android-components/commit/b6cd8418a2742a9fa0788348dc3b0731624c0f96. What do you think? Should I create a new AC issue + PR for this?

NotWoods: Yeah, please file a new issue and PR :)

TitanNano:
@NotWoods everything from the Fenix side should be addressed now.

app/src/main/java/org/mozilla/fenix/customtabs/ExternalAppBrowserActivity.kt
```diff
@@ -74,4 +82,12 @@ open class ExternalAppBrowserActivity : HomeActivity() {
             }
         }
     }
+
+    private fun ensureSessionIsReleased() {
+        val sessionManager = components.core.sessionManager
+        val sessionId = intent.getSessionId() ?: return
+        val session = sessionManager.findSessionById(sessionId) ?: return
+
+        sessionManager.getOrCreateEngineSession(session).releaseFromView()
```
NotWoods, line 91: Ideally this can be done in `BrowserStore` instead of `SessionManager`.

app/src/main/java/org/mozilla/fenix/customtabs/ExternalAppBrowserActivity.kt
TitanNano: @NotWoods PR is here: https://github.com/mozilla-mobile/android-components/pull/7381

app/src/main/java/org/mozilla/fenix/customtabs/ExternalAppBrowserActivity.kt
TitanNano: @NotWoods can you give me some pointer for what you have in mind here? I can't find much `WebApp` or `Session` related that goes through the `BrowserStore`.

liuche:
@NotWoods how do the dependencies look now, is it ready to start moving again? (generally the waiting label is better if we're waiting on deps, rather than choosing not to land for some other reason)

NotWoods:
We're waiting on designing a UX flow and testing in Sample Browser, so we can close this PR for now and re-open it later on when more is set in stone.

PR #11590, opened 2020-06-15
readme: Note that to build from the CLI ANDROID_SDK_ROOT must be set.

emilio:
Otherwise locally this fails with:

> Task 'assembleGeckoBetaDebug' not found in project ':app'.


ekager:
Thanks! pulled into https://github.com/mozilla-mobile/fenix/pull/11934 so we can run CI and land

emilio:
Thanks @ekager!

PR #11605, opened 2020-06-15
For #9712 - show "AC: " before AC version on "About" screen

cadeyrn:
Screenshot_20200615-200630

Issue #9712, opened 2020-04-04
Show "AC: " before AC version number on About screen


cadeyrn:
It's only a small detail but I propose to add "AC: " before the AC Version number on the "About" screen to be consistent with the other versions numbers. There is already a "GV: " before the GV version number and a "AS: " before the AS version number.

![Screenshot_20200404-094717](https://user-images.githubusercontent.com/1100614/78421668-bd668180-7659-11ea-8c78-2de48f46ac78.jpg)

cadeyrn:
I understand that this one is a very low priority issue and shouldn't bind developer resources but it's also very easy to implement - so my question is: Is this something you want or not?

If yes - I can create a PR.
If no - this issue can be closed.

Thanks! 🙂

csadilek:
@cadeyrn +1 from me. This caused confusion before. Definitely nice to have!

cadeyrn:
This is done and either needs a QA label or can be closed if you don't want a QA verification.

AndiAJ:
Hi, **verified as fixed** on the latest **`Nightly Build 200624 from 6/24`** using the following devices:
• Google Pixel 3a (Android 10)
• Huawei Mate 20 Lite (Android 9)
• OnePlus A3 (Android 6.0.1)

► Screenshot
![rsz_screenshot_20200624-111135](https://user-images.githubusercontent.com/51314259/85520653-e28cdf80-b60b-11ea-8f22-e556b291d9d9.png)

PR #11644, opened 2020-06-16
For #9712 - show "AC: " before AC version on "About" screen

csadilek:
Opening in place for https://github.com/mozilla-mobile/fenix/pull/11605.


csadilek:
Looked like real UI failures there, but I can't reproduce locally. So...running again,

Issue #2874, opened 2019-05-27
FNX3-16642 ⁃ [Bug] Tabs briefly visible in PB and gone from normal browsing homescreen (and vice-versa)


sv-ohorvath:
## Steps to reproduce
1. Open a tab in normal browsing.
1. Switch to private browsing and back to normal. Repeat.
*same steps are valid switching from private browsing to normal.

### Expected behavior
N.b. tabs are not displayed in P.b. N.b tabs are displayed on the homescreen all the time.

### Actual behavior
Tabs are displayed in P.b. 
Returning to normal browsing, the tabs are not displayed until you scroll the homescreen up.
*Same thing if you open tabs in p.b and switch to normal.

![20190527](https://user-images.githubusercontent.com/4037709/58420894-3474d600-8097-11e9-8ac4-0771cb8d4933.gif)UNITO-UNDERSCORE!154921!

### Device information
* Android device: On Nexus 5 (Android 6) & Mi4 (Android 5)
*Not reproducing on Galaxy Note 8 (Android 9), Sony Xperia Z5 (Android 7.1.1)
* Fenix version: 5/27

vesta0:
I couldn't reproduce. @sv-ohorvath could you please verify that this is still happening?

sv-ohorvath:
Still reproducing, checked on P9 Lite (Android 6.0). I think it's an issue on Android 6 and lower.

sblatz:
@sv-ohorvath is this still happening on the latest nightly?

sv-ohorvath:
I can't reproduce on the latest nightly 6/28 build.

sv-ohorvath:
Reproducing again on RC 2.0 and Nightly 9/3.
Devices:
• HTC Desire 820 (Android 6.0.1)
• Samsung Galaxy S6 (Android 6.0.1)
• OnePlus A3 (Android 6.0.1)
• LG Nexus 4 (Android 5.1.1)

softvision-miralobontiu:
Verified as fixed on the latest Nightly build 190927 (#12700627) with Samsung Galaxy Tab A6 (Android 5.1.1), Huawei Media Pad M2 (Android 5.1.1), and Huawei P9 Lite (Android 6).

Issue #11660, opened 2020-06-16
Allow prefetching of topsite with observeForever and allow to update homefragment in onCreateView


MarcLeclair:
Allowing to do so would save us ~100-150ms (time taken on a g5+). Note: the time on higher end device can be between 20-40ms.

I used nanoscope to profile a non debugabble build on start up (simpleperf is still struggling a bit right now so I used this), the UI isn't very friendly and works better on chrome at the moment. At ~1220656473ns in the profile, we can see the `LiveData.setValue()` callback after our first traversal. By prefetching our topsites, we'd be able to avoid this callback and show our topsites on the first layout pass. 
[noprefetch.html.zip](https://github.com/mozilla-mobile/fenix/files/4789257/noprefetch.html.zip)

On top of that, we have an issue where we update our UI with the `TopSites` and `collectionStorage` through the call `consumeFrom` which triggers a coroutine. The hiccup here is that the `launch` isn't triggered right away on our UI thread which means that on our first view traversal, even though our data should be available, we don't update it there which means a slower start up since we have to wait for the `consumeFrom` callback.

Here's an image that depics this better. The first call stack on the right is our first traversal (so the firefox logo showing, the url bar on the bottom, etc...). Topsites will then get drawn after the `livedata` callback which occurs where the two lines intersect ( the gray box in the bottom left indicates that this triggers our topsite change with `HomeFragment$onTopSiteChange`

![image](https://user-images.githubusercontent.com/16614353/84831751-e6ac7200-aff9-11ea-8b5b-48b295f25c96.png)

Issue #5375, opened 2019-09-17
[Bug] [Android 5&6] Opening a tab in normal browsing takes the user to private browsing and vice-versa


softvision-miralobontiu:
## Prerequisites
Make sure to have a clean profile.

## Steps to reproduce
1. Open a page in normal browsing.
2. From the page opened at step 1, long-tap on a hyper-link text and select to open in a new private tab.
3. Select "Switch". The user is in private browsing now, in the newly opened tab.
4. Exit private browsing and open the tab from step 1.

### Expected behavior
The user should be able to open the page opened in normal browsing in normal browsing.

### Actual behavior
After step 4, the user is redirected to private browsing's opened page.

### Note
If the user is in private browsing and opens the page from step 2, is redirected to the page opened in normal browsing.

### Device information
* Android device: Xiaomi Mi Pad 2 (Android 5.1), Huawei P9 Lite (Android 6), Huawei MediaPad M2 (Android 5.1.1)
* Fenix version: RC 2.0.0-rc.3

[logcat.txt](https://github.com/mozilla-mobile/fenix/files/3620571/logcat.txt)

![20190917_Trim](https://user-images.githubusercontent.com/40694158/65031955-e556f780-d94a-11e9-8591-be61f81bc200.gif)

ekager:
Mm I can reproduce this and it is related to #5131 . Looks like when you switch the mode the homescreen tabs don't update until you scroll the page, meaning the tabs are for the opposite mode

softvision-miralobontiu:
Verified as fixed on the latest Nightly build 190927 (#12700627) with Samsung Galaxy Tab A6 (Android 5.1.1), Huawei Media Pad M2 (Android 5.1.1), and Huawei P9 Lite (Android 6).

Issue #5131, opened 2019-09-05
[Bug] Android 5&6 - Tabs not visible when you open history/bookmarks in new tabs


sv-ohorvath:
## Steps to reproduce
1. Go to history/bookmarks.
2. Select some items.
3. Tap the 3 dot menu and select "open in new tab"/ "open in private tab".
4. Observe the homescreen. 
5. Scroll the homescreen.
### Expected behavior
The tabs created are visible on the normal/private homescreen.
### Actual behavior
The tabs are not visible at first, only after yyou scroll the homescreen a little.
Similar to https://github.com/mozilla-mobile/fenix/issues/2874
### Device information

* Android device: Huawei P9 Lite (Android 6.0), Xiaomi Mi Pad 2 (Android 5.1)
* Fenix version: Nightly 9/5 & RC 1.4.0

softvision-miralobontiu:
Verified as fixed on the latest Nightly 190927 (#12700627) with Huawei P9 Lite (Android 6), Samsung Galaxy Tab A6 (Android 5.1.1), Huawei Media Pad M2 (Android 5.1.1).

PR #11668, opened 2020-06-17
For #11660: added prefetch for topsites and update in onCreateView()

MarcLeclair:
Took the timing for this fix on a g5+ 

With changes:
```
Fully drawn org.mozilla.fennec_aurora/org.mozilla.fenix.HomeActivity: +614ms (total +1s293ms)
Fully drawn org.mozilla.fennec_aurora/org.mozilla.fenix.HomeActivity: +622ms (total +1s304ms)
Fully drawn org.mozilla.fennec_aurora/org.mozilla.fenix.HomeActivity: +610ms (total +1s294ms)
Fully drawn org.mozilla.fennec_aurora/org.mozilla.fenix.HomeActivity: +597ms (total +1s257ms)
Fully drawn org.mozilla.fennec_aurora/org.mozilla.fenix.HomeActivity: +609ms (total +1s270ms)
```
and without the changes:
```
Fully drawn org.mozilla.fennec_aurora/org.mozilla.fenix.HomeActivity: +911ms (total +1s572ms)
Fully drawn org.mozilla.fennec_aurora/org.mozilla.fenix.HomeActivity: +930ms (total +1s558ms)
Fully drawn org.mozilla.fennec_aurora/org.mozilla.fenix.HomeActivity: +946ms (total +1s558ms)
Fully drawn org.mozilla.fennec_aurora/org.mozilla.fenix.HomeActivity: +793ms (total +1s428ms)
Fully drawn org.mozilla.fennec_aurora/org.mozilla.fenix.HomeActivity: +828ms (total +1s461ms)
```
I do see a net gain with these changes. It might be worthwhile to implement the same `prefetch` for collections since it uses the same `livedata` mechanism 😄 

FYI I also did a VIEW (the new app-link name) test with FNPRMS and my results are the same as they are currently (so ~2.9 seconds). This change shouldn't affect it.

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
+++ b/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
@@ -142,6 +142,7 @@ open class FenixApplication : LocaleAwareApplication() {
             }
         }
 
+        prefetchForHomeFragment()
         setupLeakCanary()
         if (settings().isTelemetryEnabled) {
             components.analytics.metrics.start(MetricServiceType.Data)
@@ -228,6 +229,13 @@ open class FenixApplication : LocaleAwareApplication() {
         // no-op, LeakCanary is disabled by default
     }
 
+    // This is for issue https://github.com/mozilla-mobile/fenix/issues/11660. We prefetch our info for startup
+    // so that we're sure that we have all the data available as our fragment is launched.
+    private fun prefetchForHomeFragment() {
+        StrictMode.allowThreadDiskReads().resetPoliciesAfter {
+            components.core.topSiteStorage.prefetch()
+        }
+    }
     private fun setupPush() {
         // Sets the PushFeature as the singleton instance for push messages to go to.
         // We need the push feature setup here to deliver messages in the case where the service
--- a/app/src/main/java/org/mozilla/fenix/components/TopSiteStorage.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/TopSiteStorage.kt
@@ -14,6 +14,7 @@ import mozilla.components.feature.top.sites.TopSite
 import mozilla.components.feature.top.sites.TopSiteStorage
 import mozilla.components.support.locale.LocaleManager
 import org.mozilla.fenix.R
+import org.mozilla.fenix.ext.observeOnce
 import org.mozilla.fenix.ext.settings
 import org.mozilla.fenix.settings.SupportUtils
 import org.mozilla.fenix.settings.advanced.getSelectedLocale
@@ -84,6 +85,12 @@ class TopSiteStorage(private val context: Context) {
                 }
             }
             context.settings().defaultTopSitesAdded = true
+        }
+    }
+
+    fun prefetch() {
+        getTopSites().observeOnce {
+            cachedTopSites = it
         }
     }
 }
--- a/app/src/main/java/org/mozilla/fenix/ext/LiveData.kt
+++ b/app/src/main/java/org/mozilla/fenix/ext/LiveData.kt
@@ -1,0 +1,20 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.ext
+
+import androidx.lifecycle.LiveData
+import androidx.lifecycle.Observer
+
+/**
+ * Observe a LiveData once and unregister from it as soon as the live data returns a value
+ */
+fun <T> LiveData<T>.observeOnce(observer: (T) -> Unit) {
+    observeForever(object : Observer<T> {
+        override fun onChanged(value: T) {
+            removeObserver(this)
+            observer(value)
+        }
+    })
+}
--- a/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
@@ -215,6 +215,12 @@ class HomeFragment : Fragment() {
             sessionControlInteractor,
             homeViewModel
         )
+
+        // This has to be called separately from the consumeFrom block since the coroutine doesn't
+        // allow our UI to be updated right away and delays our start up. The init block allows us to
+        // force our UI to render with the data available in our store at fragment creation time.
+        sessionControlView?.update(homeFragmentStore.state)
+
         activity.themeManager.applyStatusBarTheme(activity)
 
         view.consumeFrom(homeFragmentStore, viewLifecycleOwner) {
--- a/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
@@ -216,16 +216,9 @@ class HomeFragment : Fragment() {
             homeViewModel
         )
 
-        // This has to be called separately from the consumeFrom block since the coroutine doesn't
-        // allow our UI to be updated right away and delays our start up. The init block allows us to
-        // force our UI to render with the data available in our store at fragment creation time.
-        sessionControlView?.update(homeFragmentStore.state)
+        updateSessionControlView(view)
 
         activity.themeManager.applyStatusBarTheme(activity)
-
-        view.consumeFrom(homeFragmentStore, viewLifecycleOwner) {
-            sessionControlView?.update(it)
-        }
 
         view.consumeFrom(requireComponents.core.store, viewLifecycleOwner) {
             val tabCount = if (currentMode.getCurrentMode() == Mode.Normal) {
@@ -238,6 +231,20 @@ class HomeFragment : Fragment() {
         }
 
         return view
+    }
+
+    /**
+     * The [SessionControlView] is forced to update with our current state when we call
+     * [HomeFragment.onCreateView] in order to be able to draw everything at once with the current
+     * data in our store. The [View.consumeFrom] coroutine dispatch
+     * doesn't get run right away which means that we won't draw on the first layout pass.
+     */
+    fun updateSessionControlView(view: View) {
+        sessionControlView?.update(homeFragmentStore.state)
+
+        view.consumeFrom(homeFragmentStore, viewLifecycleOwner) {
+            sessionControlView?.update(it)
+        }
     }
 
     private fun updateLayout(view: View) {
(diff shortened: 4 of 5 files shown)


app/src/main/java/org/mozilla/fenix/ext/LiveData.kt
```diff
@@ -0,0 +1,20 @@
...
+
+package org.mozilla.fenix.ext
+
+import androidx.lifecycle.LiveData
+import androidx.lifecycle.Observer
+
+/**
+ * Observe a LiveData once and unregister from it as soon as the live data returns a value
+ */
+fun <T> LiveData<T>.observeOnce(observer: (T) -> Unit) {
```
mcomella, line 13: nit: `observeForeverOnce`? It makes it clear it isn't bound to any lifecycle though it's a bit confusing, lol.

mcomella, line 13: nit: this `observer` name is redundant to the anonymous observer we create. Maybe `callback` or simple `function`?

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -215,6 +215,12 @@ class HomeFragment : Fragment() {
             sessionControlInteractor,
             homeViewModel
         )
+
+        // This has to be called separately from the consumeFrom block since the coroutine doesn't
+        // allow our UI to be updated right away and delays our start up. The init block allows us to
+        // force our UI to render with the data available in our store at fragment creation time.
+        sessionControlView?.update(homeFragmentStore.state)
```
mcomella, line 222: nit: move next to `consumeFrom` block you're referecing.

Also, it might be helpful to encapsulate them both in a helper function so the grouping is clear:
```kotlin
fun updateSessionControlView() { // maybe a better name
    fun update(state) {
        sessionControlView?.update(state)
    }

    // explain update now
    update(homeFragmentStore.state)
    
    // explain update later
    addObserver(::update)
}
```

Two best practices in play here:
- when I'm adding code with comments, I always think if it'd be more appropriate to put the code in a function to encapsulate the complexity
- when I'm adding code to lifecycle methods, I always think about using a helper to keep the lifecycle methods short and discussing things at a high level (this doesn't seem to be common on Fenix but I think it greatly helps clarity)

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -215,6 +215,12 @@ class HomeFragment : Fragment() {
             sessionControlInteractor,
             homeViewModel
         )
+
+        // This has to be called separately from the consumeFrom block since the coroutine doesn't
```
mcomella, line 219: nit: this is unclear. Which consumeFrom block (see my comment below)? What coroutine? What do you mean by "delays our start up"? How? What init block? How does it force our UI to render?

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
mcomella, line 222: **Big issue**: what happens if the data isn't available when this is called? We might want to consider blocking on it to make this more deterministic (e.g. if our fnprms starts reporting two drastically different numbers randomly, we'll be confused).

I confirmed this isn't deterministic as is: when running debug builds on Pixel 2, I see the top sites animation.

app/src/main/java/org/mozilla/fenix/components/TopSiteStorage.kt
```diff
@@ -86,4 +87,10 @@ class TopSiteStorage(private val context: Context) {
             context.settings().defaultTopSitesAdded = true
         }
     }
+
+    fun prefetch() {
```
mcomella, line 91: nit: I think we need to explain why we do this. We should probably only do so once in the code so maybe say to reference some other method (or the other explanation should reference here?)

We should probably also explain why this is innocuous.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
mcomella, line 222: **Big issue**: I don't love that we're "setting top sites twice" but only dispatching the change once but _it seems_ innocuous on P2. However, on my GS5, the whole homescreen flashes, presumably because it relays out the RV after this line: https://searchfox.org/mozilla-mobile/rev/d81a3c3fce58d3a6c3ebd3257ad1566c9bd462ef/fenix/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlView.kt#147 The current behavior does this too but it doesn't look as bad because top sites gets added instead of just replaced.

I'm not sure what we should do about that... Let's discuss

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -229,6 +230,13 @@ open class FenixApplication : LocaleAwareApplication() {
         // no-op, LeakCanary is disabled by default
     }
 
+    // This is for issue https://github.com/mozilla-mobile/fenix/issues/11660. We prefetch our info for startup
```
mcomella, line 233: I think it's confusing that we say, "our info" but really mean "top sites" so I'd rather say "We prefetch our top sites for startup" and it can be changed if the code becomes more general later.

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -229,6 +230,13 @@ open class FenixApplication : LocaleAwareApplication() {
...
+    // so that we're sure that we have all the data available as our fragment is launched.
+    private fun prefetchForHomeFragment() {
+        StrictMode.allowThreadDiskReads().resetPoliciesAfter {
+            components.core.topSiteStorage.prefetch()
+        }
+    }
     private fun setupPush() {
```
mcomella, line 240: nit: whitespace above

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -229,6 +230,13 @@ open class FenixApplication : LocaleAwareApplication() {
         // no-op, LeakCanary is disabled by default
     }
 
+    // This is for issue https://github.com/mozilla-mobile/fenix/issues/11660. We prefetch our info for startup
+    // so that we're sure that we have all the data available as our fragment is launched.
+    private fun prefetchForHomeFragment() {
+        StrictMode.allowThreadDiskReads().resetPoliciesAfter {
```
mcomella, line 236: Is this necessary? I don't see anywhere that would cause us to read on the main thread but perhaps I misunderstand.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
MarcLeclair: Interesting. I don't see it on my G5. We could do a quick "if this is the first call from the consumeFrom, ignore"

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
MarcLeclair: Not sure why, but our strictmodes complained when I didn't add it and the CI tests failed ( some of them). I could try to remove it

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
MarcLeclair: The list should be empty and will just update itself as its doing now. we could block on it too if needed

app/src/main/java/org/mozilla/fenix/ext/LiveData.kt
MarcLeclair: Yeah, the name isn't very  indicative. I could put `observeOnceWithoutLifecycleOwner` longer name ,but more self explanatory?

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
mcomella: The problem is fixed if we remove that line. The line was added in https://github.com/mozilla-mobile/fenix/pull/5591 and was trying to fix https://github.com/mozilla-mobile/fenix/issues/2874, https://github.com/mozilla-mobile/fenix/issues/5131, and https://github.com/mozilla-mobile/fenix/issues/5375. The first two relate to problems with tabs updating on the homescreen – no longer a problem – and I'm unable to reproduce the 3rd issue (maybe also tabs related)? I tried reproducing similar issues with collections:
- open homescreen, open tabs tray, click 3-dot to add to collection, see homescreen
- open history/bookmarks, add collection, open homescreen

But I can't reproduce. I think this is safe to remove but we should check with ekager.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -240,6 +234,19 @@ class HomeFragment : Fragment() {
         return view
     }
 
+    /**
+     * The [SessionControlView] is forced to update with our current state when we call [HomeFragment.onCreateView]
+     * in order to be able to draw everything at once with the current data in our store. The [View.consumeFrom] coroutine dispatch
+     * doesn't get run right away which means that we won't draw on the first layout pass.
+     * */
```
mcomella, line 241: nit: `*/`, not `* */`

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -240,6 +234,19 @@ class HomeFragment : Fragment() {
         return view
     }
 
+    /**
+     * The [SessionControlView] is forced to update with our current state when we call [HomeFragment.onCreateView]
```
mcomella, line 238: nit: why do we need to draw everything at once? Make it clear this is a perf issue

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -240,6 +234,19 @@ class HomeFragment : Fragment() {
...
+     * in order to be able to draw everything at once with the current data in our store. The [View.consumeFrom] coroutine dispatch
+     * doesn't get run right away which means that we won't draw on the first layout pass.
+     * */
+    fun updateSessionControlView(view: View) {
+        sessionControlView?.update(homeFragmentStore.state)
```
mcomella, line 243: nit: I like the inner function version I suggested because it makes it clear that the code is repeated and prevents someone from missing that if they decide to refactor this. But I can go either way

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -216,16 +216,10 @@ class HomeFragment : Fragment() {
             homeViewModel
```
mcomella, line 216: nit: In the extended commit summary, I would have included information on why it's okay to remove this code (though I guess it's fine if it's just in the PR text? maybe you can link to what's in the PR).

Issue #8847, opened 2020-02-28
Make dynamic nav bar optional 


tom-james-watson:
### Why/User Benefit/User Problem

I, and many other users, find dynamic nav bars to be a frustrating UX. I don't like UI elements to move around underneath me and just find that it adds extra steps to achieve the same things when compared to a fixed nav bar.

As this was previously controlled by a flag, I assume adding a setting for this should be trivial, and would make a lot of users happy. 

## What/Requirements

Add a setting (probably to Customize->Toolbars) for whether nav bar should be dynamic or not. 

### Acceptance Criteria (how do I know when I’m done?)

I can toggle this setting off and have the nav bar be static.

mattcoz:
Yes, this used to be controlled by browser.chrome.dynamictoolbar, but I've tried setting that to false and it does not do anything. Respecting that flag would seem to be the best thing to do here.

AlbertoFabbri93:
According to me the best option would be to add this setting below the one that let you choose if you want the toolbar at the top or at the bottom. There is plenty of space in that screen.

Mugurell:
Thank you for the patches @fredboy!
At a quick glance they look good! 👍

I know this is a highly requested feature but the UX specs for this are still in flux.
We have some initial feedback here - https://github.com/mozilla-mobile/fenix/issues/10240#issuecomment-641715313 but still no clear specs.
Also at this time we are in a stabilization period so big changes like this are unlikely.

But because of the work you already put in what could maybe help expedite this feature is documenting the changes (like the title for this new preference, removing the nightly flag, etc) and also adding a quick video so that UX can easily review and decide if and when we want this change.

willfs84:
Also, this is probably an accessibility issue too. IIRC that is one of the reasons the option was added to Fennec. So not only should this be a preference under "Customize", but the dynamic nav bar should be disabled for users that have Android's "Remove animations" accessibility option enabled.

hwinnemoe:
This has been implemented by #10240.

Mugurell:
Closing as per the above comment.

Issue #10240, opened 2020-04-27
Allow users disabling gesture based features


Mugurell:
### Why/User Benefit/User Problem
Based on user requests for an option to disable gesture based features we should allow an easy way of toggling:
- pull to refresh in browser
- dynamic toolbar (top / bottom)
- swipe toolbar horizontally to change tabs
- swipe toolbar up to show open tabs (to be implemented after https://github.com/mozilla-mobile/fenix/issues/11862)

### What/Requirements
New toggles for the above features.
Strawman mockup - https://github.com/mozilla-mobile/fenix/issues/10240#issuecomment-647241235:
<img width="300" src="https://user-images.githubusercontent.com/51007/85244431-c0268500-b498-11ea-87d0-c53f08957bf7.png">
String resources -  https://github.com/mozilla-mobile/fenix/issues/13442#issuecomment-682618440
> Here are strings for all four gestures. Can you confirm that the first two will be what will be implemented? The ones from the mock are outdated.
> 
> * Pull to refresh
> * Scroll to hide toolbar
> * Swipe toolbar sideways to switch tabs
> * Swipe toolbar up to open tabs

### Acceptance Criteria (how do I know when I’m done?)
Users can easily find the settings to enable / disable gesture based features.

Mugurell:
@UX how should this be organized?
- In new "Gestures" settings screen
  - In a new "Gestures" settings screen?
  - In a new "Gestures" section?
  - Individual toggles?
- Should the above be inside Settings -> Accessibility or in a new page inside Settings?

brampitoyo:
@Mugurell Would you be able to describe some of our user issues with pull to refresh and dynamic toolbar?

I assume that pull to refresh might interfere with certain site header, and dynamic toolbar would sometimes not minimise?

Mugurell:
Basically:
- user gestures trigger pull down / dynamic toolbar when they shouldn't (horizontal scrolls, zoom motions)
- there still edgecases with websites for which pull down/dynamic toolbar are triggered when they shouldn't / are not triggered when they should.
- for the dynamic toolbar there are indeed cases in which it interferes with the page layout, like https://github.com/mozilla-mobile/fenix/issues/8768

And there are users that requested an option to disable these: https://github.com/mozilla-mobile/fenix/issues/8847, https://github.com/mozilla-mobile/fenix/issues/9719

AmyYLee:
@brampitoyo to follow-up

[...]

PR #11678, opened 2020-06-17
For #8847: Add preference for dynamic toolbar to Customization menu

fredboy:
Added new preference key and string value for dynamic toolbar. The new preference is now used instead of feature flag for dynamic bottom toolbar. This preference can be changed in settings, customization menu with CheckBox, which I've also added. 

When toolbar was on top, it checked if user has screen reader enabled to determine if it should not be dynamic. I left it as it was, because I'm not sure, how it should be. So if user has a dynamic top toolbar enabled, it still won't be dynamic if the screen reader is also enabled. If dynamic toolbar is enabled at the bottom, it will be dynamic regardless of screen reader.

Here's the GIF of how toolbar is behaving with different settings:
Screen-Recording-20200615-231412

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/FeatureFlags.kt
+++ b/app/src/main/java/org/mozilla/fenix/FeatureFlags.kt
@@ -21,11 +21,6 @@ object FeatureFlags {
     const val asFeatureSyncDisabled = false
 
     /**
-     * Enables dynamic bottom toolbar
-     */
-    val dynamicBottomToolbar = Config.channel.isNightlyOrDebug
-
-    /**
      * Enables deleting individual tracking protection exceptions.
      */
     val deleteIndividualTrackingProtectionExceptions = Config.channel.isNightlyOrDebug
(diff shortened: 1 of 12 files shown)


Mugurell:
Thank you @fredboy !
This looks great!

But I see in https://github.com/mozilla-mobile/fenix/issues/10240#issuecomment-647241235 some proposed specs different that what is used in this patch.

Sorry for the work invested already but thank you for pushing through with this!

Maybe after we get the final specs you would like to update the patch accordingly?

fredboy:
@Mugurell yes, sure. I'll be glad to

AmyYLee:
My feedback is that if we want to add a new "dynamic" setting that we should have strings to describe what this does. I'll let @brampitoyo make the final call.

brampitoyo:
@fredboy The dynamic toolbar idea makes a lot of sense, but as the only difference between dynamic and non-dynamic concerns the use of gesture (ie. does it minimise when you scroll up/down the page, or not?), we should put it under a “Gestures” section.

Otherwise, I agree that “Customise” is the right sub-page to put this setting menu item, so thanks for proposing it!

brampitoyo:
Closing this issue in favour of implementing gesture options on https://github.com/mozilla-mobile/fenix/issues/10240#issuecomment-647241235.

Issue #11616, opened 2020-06-15
[Bug] Flash when opening the app on low API levels


mcomella:
This may be related to https://github.com/mozilla-mobile/fenix/issues/9904 or maybe not.

## Steps to reproduce
* Open the app (either through homescreen icon or clicking a link)

### Expected behavior
App opens smoothly

### Actual behavior
App flashes during opening animation:

* On the Galaxy S5 (API 23), the device homescreen will flash in the middle of the animation
* On the Moto G5 (API 26 or 27), the device will flash black in the middle of the animation

Videos reproducing the issue:

* [Galaxy S5](https://drive.google.com/file/d/14sU9RydD0vX8UvgPTmbpMN_3EAQ_tU6m/view?usp=sharing) (fennec is loaded at the end for comparison)
* [Moto G5+](https://streamable.com/c45y3x)

### Device information
* Android device: See above; assumed any device at a low API level
* Fenix version: Nightly 200615

- - -
I believe the root cause is that we're starting an Activity (either `MigrationDecisionActivity` or `IntentReceiverActivity`), `finish`ing it, and starting a new Activity. This does not reproduce in non-migration builds on the homescreen (e.g. compare latest fennecNightly to fenixNightly), which backs up this belief.

In addition to this flash, these redirection activities also have a significant impact on performance for low-end devices: somewhere between 100ms and 200ms on the G5 (most perf wins we find are 10-20ms). I filed https://github.com/mozilla-mobile/fenix/issues/7927 to generically remove these and #11608 + https://github.com/mozilla-mobile/fenix/issues/11072 were later filed to investigate removing each of these.

This does not appear to affect newer devices: I assume at some point, Google changed the Android OS behavior to not flash when switching activities. Therefore, **I expect this to affect all devices of a certain API level and below** (though perf affects all devices). _edit:_ the G5+ is Android 8.0 or so – perhaps this assumption is false.

mcomella:
CC @MarcLeclair

PR #11717, opened 2020-06-18
DO NOT LAND: Draft for removing black screen flicker on start

MarcLeclair:
@pocmo  not sure if I'm missing much but doing this avoid a heavy black flickr on start up and seems to not change too much of our migration code. Would this change be okay with you?

cc @mcomella 

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/migration/java/org/mozilla/fenix/MigrationDecisionActivity.kt
+++ b/app/src/migration/java/org/mozilla/fenix/MigrationDecisionActivity.kt
@@ -4,9 +4,8 @@
 
 package org.mozilla.fenix
 
-import android.app.Activity
+import android.content.Context
 import android.content.Intent
-import android.os.Bundle
 import mozilla.components.support.migration.state.MigrationProgress
 import org.mozilla.fenix.ext.components
 import org.mozilla.fenix.migration.MigrationProgressActivity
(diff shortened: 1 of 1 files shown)


MarcLeclair:
@pocmo That makes sense. Yeah, the black flicker happens mostly on lower end devices (not *just* low ends  as in the G5 but I also notice it on a nexus 6p huawei p10 and a few other "recent" phones that are pre 2017). I got a link here: https://streamable.com/c45y3x. FYI the black screen is worse on the `geckonightlyfennecnightly` build for some odd reason.

I tried multiple different ways to remove that by playing around with `MigrationDecisionActivity` and it seems to be the fact that we're killing the activity and starting a new one since it doesn't happen with the patch here. The issue is that this black screen happens everytime I open the app. Not just once when the user is migrating but every time (plus going through the `startactivity` seems to take 100-150ms on medium/lower end device, go wonder why its so slow there)

mcomella:
> However I would like to learn more about this black screen flicker. Is there an issue about this?

https://github.com/mozilla-mobile/fenix/issues/11616

MarcLeclair:
I'll close this since it seems I've found a possible alternative solution that seems a lot safer than jeopardizing the whole migration workflow. For reference, here's the draft PR : #11796

pocmo:
Awesome! That's great. :)

PR #11718, opened 2020-06-18
For #11185: Keep a RecycledViewPool for TabsTray views

### LINKED ISSUE #11185: Improve performance of async tab load when opening tabs tray
**Release criteria** for opening tabs tray: Target 250ms from input to UI rendered. (note: 2x: entry points from browser and home screen are different)

In [our performance analysis](https://github.com/mozilla-mobile/fenix/issues/7512#issuecomment-641621240), we determined the tabs tray appears quickly enough but **the tabs take a long time to appear.** The duration they take to appear seems to at least somewhat tied to the number of tabs opened. **On a GS5 with 8 open tabs, we're at 462ms,** about 212ms longer than we want to be.

We should improve the performance of the async tab loading.

Speculation: perhaps the tabs take so long to appear because they're blocking on getting the tab previews before loading the tabs. If not, I'd expect loading tabs to be fast because it's just a bunch of strings.

If so, this is a problem: it'll get very unperformant with a large number of open tabs. We should load the tabs and asynchronously fetch the tab when they're needed.

- - -
_The original text of this issue is described here._

## Prerequisites
Open more than 10 websites in different tabs

## Steps to reproduce
1. After opening twitter.com (12th tab)
1. Open the Tabs Tary
1. Tap any other open tab
1. Open the Tabs Tary
1. Tap and open another tab
1. Open the Tabs Tary
1. Swipe delete the first 3 open tabs 

### Expected behavior
Steps 4,6: Responsive Tabs Tray, displays properly the open tabs
Step 7: The tabs should be deleted

### Actual behavior
Steps 4,6: Laggy display of the open tabs 
Step 7: Empty space left in the Tabs Tray, 
After a couple of seconds the tabs are displayed again as open and only after a couple of more seconds are properly deleted

### Device information
* Android device: Google Pixel 3a (Android 10)
* Fenix version: Nightly Build 200603 from 6/3

### Notes
<details>
    <summary>Web site I've used (dropdown)</summary>

Home - YouTube
https://m.youtube.com/

Vimeo | The high-quality home for video hosting and watching
https://vimeo.com/

Dailymotion - the home for videos that matter
https://www.dailymotion.com/us

SoundCloud - Hear the world’s sounds
https://m.soundcloud.com/

Wikipedia
https://www.wikipedia.org/

Pocket: Explore
https://getpocket.com/explore?src=ff_android&cdn=0

Office 365 Login | Microsoft Office
https://www.office.com/

Microsoft – Pagina de pornire oficială
https://www.microsoft.com/ro-ro/

Internet for people, not profit — Mozilla
https://www.mozilla.org/en-US/exp/

Facebook - Log In or Sign Up
https://m.facebook.com/

Instagram
https://www.instagram.com/

Explore / Twitter
https://mobile.twitter.com/explore

</details>

► Video

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-1606)

jonalmeida:
Fixes #11185

We use a `RecycledViewPool` so that we can re-use existing views from a previous tabs tray adapter.

On subsequent opens of the tabs tray, we're always creating a new view holder each time to inflate and draw a view ([profile](https://share.firefox.dev/2USU27k)):

Screen Shot 2020-06-18 at 12 12 58 AM

With a `RecycledViewPool` and the LayoutManager set to `recycleChildrenOnDetach`, we see a 30ms - 100ms saving depending on how many open tabs you had loaded. In the screenshot below that contains this change, we're only creating two new viewholders instead of the original 7 and saving 56ms to layout the `RecyclerView` ([profile](https://share.firefox.dev/2Y8IGhE)):

Screen Shot 2020-06-18 at 12 16 24 AM

---

Some other things I investigated was to re-use the same adapter and layout manager instead of recreating it all the time as well, but this proves to be difficult:
 - A layout manager can only be attached to one RecyclerView ever.
 - The adapter throws invalid adapter positions that crash the app.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
+++ b/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
@@ -25,6 +25,7 @@ import androidx.navigation.fragment.NavHostFragment
 import androidx.navigation.ui.AppBarConfiguration
 import androidx.navigation.ui.NavigationUI
 import androidx.recyclerview.widget.LinearLayoutManager
+import androidx.recyclerview.widget.RecyclerView
 import kotlinx.android.synthetic.main.activity_home.*
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.launch
@@ -125,6 +126,19 @@ open class HomeActivity : LocaleAwareAppCompatActivity() {
             OpenBrowserIntentProcessor(this, ::getIntentSessionId),
             OpenSpecificTabIntentProcessor(this)
         )
+    }
+
+    /**
+     * A single view pool to be re-used when the tabs tray is subsequently opened.
+     *
+     * This lives on the activity level since the tabs tray is frequently used and we can save our
+     * recycled views until the activity is done.
+     *
+     * This is especially useful along with [LinearLayoutManager.setRecycleChildrenOnDetach] as true
+     * to make maximum use of recycled views.
+     */
+    private val tabsTrayViewPool by lazy {
+        RecyclerView.RecycledViewPool()
     }
 
     final override fun onCreate(savedInstanceState: Bundle?) {
@@ -239,15 +253,16 @@ open class HomeActivity : LocaleAwareAppCompatActivity() {
             }
         }.asView()
         TabsTray::class.java.name -> {
+            val thumbnailLoader = ThumbnailLoader(components.core.thumbnailStorage)
+            val adapter = FenixTabsAdapter(context, thumbnailLoader)
             val layout = LinearLayoutManager(context).apply {
                 reverseLayout = true
                 stackFromEnd = true
+                recycleChildrenOnDetach = true // See [tabsTrayViewPool] docs.
             }
-
-            val thumbnailLoader = ThumbnailLoader(components.core.thumbnailStorage)
-            val adapter = FenixTabsAdapter(context, thumbnailLoader)
-
-            BrowserTabsTray(context, attrs, 0, adapter, layout)
+            BrowserTabsTray(context, attrs, 0, adapter, layout).apply {
+                setRecycledViewPool(tabsTrayViewPool)
+            }
         }
         else -> super.onCreateView(parent, name, context, attrs)
     }


mcomella:
Measurements with the patch on gs5:
- First click: 5 runs, average: 427.4; [485, 435, 412, 518, 287]
- Subsequent clicks: 10 runs, average: 249.8; [226, 237, 244, 253, 252, 268, 284, 219, 266, 249]

Measurements without the patch on the gs5:
- First click: 5 runs, average: 348.6; [407, 318, 317, 286, 415]
- 10 runs, average: 306.8; [282, 321, 300, 398, 275, 254, 309, 280, 368, 281]

Caveats:
- there's a high variance on the first click so 5 runs is definitely not enough (even 10 is questionable to get a stable number)

Conclusions:
- the patch gains 57ms on subsequent clicks with greatly reduced variance
- the patch appears to lose 78.8ms on the first click (putting it into the 427ms range; note our RC is 250ms and 300ms if we're stretching it); however, there's high variance and a possible outlier in the sample set for the first click. I'll try running a few more runs because the extremely decreased variance seems valuable

mcomella:
First click with this patch: 10 runs, average: 403.7; [400, 378, 369, 559, 352, 431, 410, 408, 333, 397]

First click without the patch: 10 runs, average: 367.2; [361, 378, 391, 379, 345, 362, 325, 288, 529, 314]

So that's 36.5ms worse with the patch on first click.

I noticed the google logo in the toolbar takes a long time to appear so I waited until it was displayed to click the tabs tray button just in case that pending operation affected the results.

---

I'm curious how asynchronously inflating the RV views in advance of the first click might improve performance.

@MarcLeclair What do you think about these results and how we should approach fixing this issue? fwiw, I provided some other possible improvements in https://github.com/mozilla-mobile/fenix/issues/11185#issuecomment-644923798.

jonalmeida:
> I'm curious how asynchronously inflating the RV views in advance of the first click might improve performance.

This could work as well. The [`AsyncLayoutInflater`](https://developer.android.com/reference/androidx/asynclayoutinflater/view/AsyncLayoutInflater) can work on the background thread and we can create `ViewHolder`s and put them in into the view pool with [`RecycledViewPool#putRecycledView`](https://developer.android.com/reference/androidx/recyclerview/widget/RecyclerView.RecycledViewPool#putRecycledView(androidx.recyclerview.widget.RecyclerView.ViewHolder)).

> fwiw, I provided some other possible improvements in #11185 (comment).

The other options are something we should definitely investigate as well, although it would require larger changes to how the tabs tray is created that I'm not familiar with. This seemed like low hanging fruit 🥭  so it was worth spending some time on it sooner.

If we can async create the tabs tray before it is needed, that would help a lot and undo the need for this change.

MarcLeclair:
My thoughts after reading this is the following: If we get a perf gain from using the `RecyclerViewPool`, I think we should do it. However, the first click being affected does seem unoptimal. Could we initialized it after the first draw of the tabs tray? I.e: `RecyclerView.postDelayedAnimation  { //init pool here } ``  -- Although, I'm not sure if you can set it after it was rendered.

Another thought for the Async stuff with RVs is that,  you need to attach it to its parent ( or at least pass it a ViewGroup if I remember correctly). I like the idea, but I'm not sure how we'd approach it. I do remember trying to use AsyncLayoutInflater + FrameLayouts to try to improve the perf of start up a while back (when tabs were still on the homescreen). I ran into the issue that the recycling of views didn't go well and would mix and match with each other. I didn't dig more into it, but not sure if this would happen again.

mcomella:
Adding `do not land` to prevent us landing without further perf understanding. Maybe next step is to retake first click results to see if the first click regressions are accurate.

jonalmeida:
I have some more next step ideas:
 - Determine what is causing the first-click perf loss here. A view pool is created internally so there shouldn't be a loss. Maybe the order matters?
 - Create the view pool when the visually complete callback is fired, even if we don't use the pool immediately. This might help to reduce the first-click losses.
    - Try to create some views to dump in there as well.
 - Investigate (again) if we can create the entire BrowserTabsTray lazily. Leaving the view attachment for when it's needed.

jonalmeida:
For https://github.com/mozilla-mobile/android-components/issues/7079, I'm working combining multiple adapters for one RecyclerView. This would alter our current ideas of how to cache the view pool, so I'm going to recommend halting on this patch until then.

I'm going to close it to avoid PR clutter, but we can re-open it or move the discussion to the original issue as well.

PR #11748, opened 2020-06-18
For #10642: Comment out verifyPageContent calls

AaronMT:
Issue #10642 (verifyPageContent) is very flaky and causing test timeouts in test-failures. I wouldn't be surprised if the MockWebServer version we're using from 2018 is part of the problem, for now comment out each ``verifyPageContent`` call and the function. Retaining its definition so we can still investigate on the side.


rpappalax:
> Issue #10642 (verifyPageContent) is very flaky and causing test timeouts in test-failures. I wouldn't be surprised if the MockWebServer version we're using from 2018 is part of the problem, for now comment out each `verifyPageContent` call and the function. Retaining its definition so we can still investigate on the side.

@AaronMT thanks, Aaron. Is the follow-up work being tracked in a separate issue?  I want to make sure this has a clear owner before we move on

PR #11751, opened 2020-06-19
For #2165 - Add swipe to refresh gesture to the bookmarks view.

person808:
Requires https://github.com/mozilla-mobile/android-components/pull/7437 which updates A-S to v61.0.4

Will leave as draft until we pick up those changes but this should otherwise be ready for review.

### Pull Request checklist

output

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkController.kt
+++ b/app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkController.kt
@@ -11,13 +11,19 @@ import android.content.res.Resources
 import androidx.core.content.getSystemService
 import androidx.navigation.NavController
 import androidx.navigation.NavDirections
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.launch
+import mozilla.appservices.places.BookmarkRoot
 import mozilla.components.concept.engine.prompt.ShareData
 import mozilla.components.concept.storage.BookmarkNode
+import mozilla.components.service.fxa.sync.SyncReason
 import org.mozilla.fenix.BrowserDirection
 import org.mozilla.fenix.HomeActivity
 import org.mozilla.fenix.R
 import org.mozilla.fenix.browser.browsingmode.BrowsingMode
 import org.mozilla.fenix.components.metrics.Event
+import org.mozilla.fenix.ext.bookmarkStorage
+import org.mozilla.fenix.ext.components
 import org.mozilla.fenix.ext.nav
 
 /**
@@ -26,16 +32,20 @@ import org.mozilla.fenix.ext.nav
  */
 @SuppressWarnings("TooManyFunctions")
 interface BookmarkController {
+    fun handleBookmarkChanged(item: BookmarkNode)
     fun handleBookmarkTapped(item: BookmarkNode)
     fun handleBookmarkExpand(folder: BookmarkNode)
     fun handleSelectionModeSwitch()
     fun handleBookmarkEdit(node: BookmarkNode)
     fun handleBookmarkSelected(node: BookmarkNode)
+    fun handleBookmarkDeselected(node: BookmarkNode)
+    fun handleAllBookmarksDeselected()
     fun handleCopyUrl(item: BookmarkNode)
     fun handleBookmarkSharing(item: BookmarkNode)
     fun handleOpeningBookmark(item: BookmarkNode, mode: BrowsingMode)
     fun handleBookmarkDeletion(nodes: Set<BookmarkNode>, eventType: Event)
     fun handleBookmarkFolderDeletion(node: BookmarkNode)
+    fun handleRequestSync()
     fun handleBackPressed()
 }
 
@@ -43,6 +53,10 @@ interface BookmarkController {
 class DefaultBookmarkController(
     private val context: Context,
     private val navController: NavController,
+    private val scope: CoroutineScope,
+    private val store: BookmarkFragmentStore,
+    private val sharedViewModel: BookmarksSharedViewModel,
+    private val loadBookmarkNode: suspend (String) -> BookmarkNode?,
     private val showSnackbar: (String) -> Unit,
     private val deleteBookmarkNodes: (Set<BookmarkNode>, Event) -> Unit,
     private val deleteBookmarkFolder: (BookmarkNode) -> Unit,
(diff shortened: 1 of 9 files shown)


app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkController.kt
```diff
@@ -69,7 +93,22 @@ class DefaultBookmarkController(
     }
 
     override fun handleBookmarkSelected(node: BookmarkNode) {
-        showSnackbar(resources.getString(R.string.bookmark_cannot_edit_root))
+        if (store.state.mode is BookmarkFragmentState.Mode.Syncing) {
+            return
+        }
+
+        when (node.inRoots()) {
```
NotWoods, line 100: nit: use an `if` statement instead of `when (boolean)`

app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkFragment.kt
```diff
@@ -126,6 +127,18 @@ class BookmarkFragment : LibraryPageFragment<BookmarkNode>(), UserInteractionHan
...
+            // Don't want to pester user too much with it, and if there are lots of bookmarks present,
+            // it'll just get visually lost. Inside of the "Desktop Bookmarks" node, it'll nicely stand-out,
+            // since there are always only three other items in there. It's also the right place contextually.
+            if (it.tree?.guid == BookmarkRoot.Root.id &&
+                requireComponents.backgroundServices.accountManager.authenticatedAccount() == null
+            ) {
+                signInView.view.visibility = View.VISIBLE
+            } else {
+                signInView.view.visibility = View.GONE
+            }
```
NotWoods, line 143: nit: is `isVisible` here.

```suggestion
            signInView.view.isVisible = it.tree?.guid == BookmarkRoot.Root.id &&
                requireComponents.backgroundServices.accountManager.authenticatedAccount() == null
```

This would probably be easier to read if we get `accountManager` outside of `consumeFrom`:
```kt
val accountManager = requireComponents.backgroundServices.accountManager
consumeFrom(bookmarkStore) {
	...accountManager..authenticatedAccount() == null
}
```

app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkFragment.kt
```diff
@@ -138,36 +151,23 @@ class BookmarkFragment : LibraryPageFragment<BookmarkNode>(), UserInteractionHan
...
-            signInView.view.visibility = View.VISIBLE
-        } else {
-            signInView.view.visibility = View.GONE
-        }
 
-        initialJob = loadInitialBookmarkFolder(currentGuid)
+        // Reload bookmarks when returning to this fragment in case they have been edited
+        val currentGuid = bookmarkStore.state.tree?.guid
+            ?: if (BookmarkFragmentArgs.fromBundle(requireArguments()).currentRoot.isNotEmpty()) {
+                BookmarkFragmentArgs.fromBundle(requireArguments()).currentRoot
```
NotWoods, line 160: We should save the arguments to a local variable and remove a redundant second parsing step. The `navArgs` delegate can also make the code a bit shorter.

```suggestion
        val args by navArgs<BookmarkFragmentArgs>()
        val currentGuid = bookmarkStore.state.tree?.guid
            ?: if (args.currentRoot.isNotEmpty()) {
                args.currentRoot
```

app/src/test/java/org/mozilla/fenix/library/bookmarks/BookmarkControllerTest.kt
```diff
@@ -36,13 +47,18 @@ import org.mozilla.fenix.components.metrics.Event
 import org.mozilla.fenix.ext.components
 
 @SuppressWarnings("TooManyFunctions", "LargeClass")
+@ExperimentalCoroutinesApi
 class BookmarkControllerTest {
 
     private lateinit var controller: BookmarkController
+    private val mainThreadSurrogate = newSingleThreadContext("UI thread")
```
NotWoods, line 55: I think `MainCoroutineRule` would be sufficient here. Android Components uses it a lot to mock the main dispatcher, you can see example usage there.

app/src/test/java/org/mozilla/fenix/library/bookmarks/BookmarkControllerTest.kt
```diff
@@ -36,13 +47,18 @@ import org.mozilla.fenix.components.metrics.Event
...
 
     private val bookmarkStore = spyk(BookmarkFragmentStore(BookmarkFragmentState(null)))
     private val context: Context = mockk(relaxed = true)
+    private val scope: TestCoroutineScope = TestCoroutineScope()
```
NotWoods, line 59:
```suggestion
    private val scope = TestCoroutineScope()
```

app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkFragment.kt
```diff
@@ -389,4 +391,17 @@ class BookmarkFragment : LibraryPageFragment<BookmarkNode>(), UserInteractionHan
             }
         }
     }
+
+    private suspend fun syncBookmarks() {
```
NotWoods, line 395: As far as I can tell, this logic doesn't seem coupled to the fragment. Could it be moved into the bookmarks controller?

app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkFragmentStore.kt
```diff
@@ -20,17 +20,22 @@ class BookmarkFragmentStore(
...
  * @property tree The current tree of bookmarks, if one is loaded
  * @property mode The current bookmark multi-selection mode
+ * @property guidBackstack A set of guids for bookmark nodes we have visited. Used to traverse back
+ *                  up the tree after a sync.
+ * @property isLoading true if bookmarks are still being loaded from disk
  */
 data class BookmarkFragmentState(
     val tree: BookmarkNode?,
     val mode: Mode = Mode.Normal(),
+    val guidBackstack: Set<String> = linkedSetOf(),
```
NotWoods, line 30: It doesn't seem like we use any properties of a Set here. Maybe a list should be used instead?

app/src/test/java/org/mozilla/fenix/library/bookmarks/BookmarkControllerTest.kt
person808: It seems like `MainCoroutineRule` isn't needed either. Tests seem to pass without it anyway

app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkFragment.kt
person808: Keeping it in the fragment makes testing a bit easier since it allows passing the function to the controller's constructor.

NotWoods: If you just need to check if this function is called in the tests, I think it would be better to make it a `@VisibleForTesting` function in the controller and use a spy (and optionally mock it so it doesn't run any code in other tests).

person808: Ended up moving this logic to `handleRequestSync` in the controller. It was giving me a lot of trouble trying to test it so I just tested the order of sync related dispatches which is what I was more concerned with anyway.

Issue #9144, opened 2020-03-13
[Bug] White screen when using back swipe to close Firefox Preview and reopening from 'Recents' 


adityamurugan:
## Steps to reproduce
1. Close Firefox preview
2. Open external link in Firefox Preview
3. Use back button/back swipe to close Firefox Preview
4. Open Firefox Preview again from 'Recents'

### Expected behavior
Firefox Preview should open the home page without any white screens

### Actual behavior
Firefox preview shows a blank white screen and need to be closed and opened to fix

### Device information

* Android device: Pixel 3a
* Fenix version: Nightly 200312 18.01

sblatz:
Video of the behavior: https://www.reddit.com/r/firefox/comments/fhsemk/bugfenix_white_screen_when_using_back_swipe_to/

adityamurugan:
There has been a small change in behavior but the bug is still there.

-The first back swipe closes the external link tab and the home page is shown
-The second back swipe minimizes the app to recents
-Now, trying to reopen the app from the recents drawer shows a white screen. 

Version: Nightly 200416 08:30 (Build #21070840)

sflorean:
Tested on Nightly 6/12 and I was able to reproduce the issue on the following devices, **only** if the page from step 2 is opened in the browser:
- Google Pixel 3 (Android 11-beta)
- LG G7 FIT (Android 8) 
If I have more than 1 open tabs, the previous open tab is displayed after I open Nightly from Recents.

abodea:
Verified as fixed on the latest **Nightly 6/23** with **Google Pixel 4 XL (10).**
Note that the home page was displayed in both cases when I had 1 or more tabs open.

![20200623-115708](https://user-images.githubusercontent.com/42831109/85384734-1bfd1680-b54a-11ea-87b0-3528048eb252.gif)

PR #11772, opened 2020-06-19
For #9144 - Ignore previous Intent if activity is started from Recents

Mugurell:
We'll now clearly differentiate between cold / hot starts of HomeActivity.kt.
This is needed because Android will resend the original Intent which initially
started the Activity whenever it is restarted from the Recents Screen if the
activity is already destroyed at that time. So in the event that the activity
was before started with an Intent to open a webpage for example whenever the
activity is restarted from Recents it will receive the same Intent to open a
webpage even though that Intent has already been consumed.

Activity's onCreate() will only use the intent processors when the activity is
cold started so that we'll only initially act upon Intents configured for
different behaviors inside the app.
If the activity is destroyed while in background and opened from Recents it
will not act upon the original Intent which is now resent by Android.

Activity's onNewIntent() will be called to act upon a new Intent if the
activity is hot started since we are declared as singleTask and it now has the
responsibility to delegate various intent processors to consume that Intent.

### Pull Request checklist

- [x] **Tests**: This PR includes tests for the newly added method.
- [x] **Screenshots**: This PR includes GIFs of the changes made.

Gif showing the blank screen not appearing anymore if app is restarted from Recents after pressing the back button or after being put to background while using the "Don't keep activities".
NoBlankScreenAfterStartingTheApp

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
+++ b/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
@@ -154,7 +154,9 @@ open class HomeActivity : LocaleAwareAppCompatActivity() {
 
         sessionObserver = UriOpenedObserver(this)
 
-        externalSourceIntentProcessors.any { it.process(intent, navHost.navController, this.intent) }
+        if (isActivityColdStarted(intent, savedInstanceState)) {
+            externalSourceIntentProcessors.any { it.process(intent, navHost.navController, this.intent) }
+        }
 
         Performance.processIntentIfPerformanceTest(intent, this)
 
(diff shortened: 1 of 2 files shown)


ekager:
Thanks for looking into this! One more thing we could add is pulling this piece from Focus into the action view intent removal in `BrowserFragment#removeSessionIfNeeded` to make sure to remove the task on a back press when there are no more sessions? May be redundant with your changes... I'm not sure. Thoughts? 
https://github.com/mozilla-mobile/focus-android/blob/899540bba6c8a70898e30e3e30ebf1593179ab02/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt#L951-L957

Mugurell:
> Thanks for looking into this! One more thing we could add is pulling this piece from Focus into the action view session removal in `BrowserFragment#removeSessionIfNeeded` to make sure to remove the task on a back press when there are no more sessions? May be redundant with your changes... I'm not sure. Thoughts?
> https://github.com/mozilla-mobile/focus-android/blob/899540bba6c8a70898e30e3e30ebf1593179ab02/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt#L951-L957

This would be an interesting approach.. 
It could replace the changes from this PR but it would only work for resolving one STR from the two that we found:
- user presses back to close the tab and the activity. We could now remove the task so at the next start from launcher we would get a new clean Intent.
- user closes the only tab but stays on HomeFragment and puts the app in background, where after a time HomeActivity is destroyed. When opening it again from Recents this issue would reproduce.
We could catch when the user puts the app to background or maybe we could try to get when the activity is destroyed and remove the task then but I think it's reasonable for the user to expect to find it's backgrounded app in Reecnts and be able to start it from there.

Giving that this approach would fix only half of the issue and the fact that https://github.com/mozilla-mobile/focus-android/issues/888 had a lot of discussions going before deciding to remove the task when exiting the app I think we should go with just the solution initially proposed.

If it's ok with you also we could file a new issue for UX and product to decide if we should remove the task from Recents when closing the app with no open tabs. (I would prefer to keep it. Most of the apps do so, it's good to have it available in more places for the user, I think Focus had a different usecase than ours).

PR #11796, opened 2020-06-22
For 11616: removed black flicker with placeholder animation

MarcLeclair:
Video with the changes: https://streamable.com/zk47js (no flicker)
Video without the changes (with black screen flicker): https://streamable.com/1qqee7

I think this is a better option than changing the `MigrationDecisionActivity` to inherit `HomeActivity`, especially right before release.  I'm not sure how adding the animation would affect newer devices though. It seemed fine on my Pixel 4 though.

Thoughts? 

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/res/anim/placeholder_animation.xml
+++ b/app/src/main/res/anim/placeholder_animation.xml
@@ -1,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<translate
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    android:fromXDelta="0"
+    android:toXDelta="0"
+    android:duration="900" />
--- a/app/src/main/res/values/styles.xml
+++ b/app/src/main/res/values/styles.xml
@@ -20,6 +20,7 @@
         <item name="alertDialogStyle">@style/DialogStyleNormal</item>
         <item name="alertDialogTheme">@style/DialogStyleNormal</item>
         <item name="android:windowEnableSplitTouch">false</item>
+        <item name="android:navigationBarColor">@android:color/transparent</item>
         <item name="android:splitMotionEvents">false</item>
 
         <item name="mozacInputLayoutErrorTextColor"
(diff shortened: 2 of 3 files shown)


app/src/main/res/values/styles.xml
```diff
@@ -20,6 +20,7 @@
         <item name="alertDialogStyle">@style/DialogStyleNormal</item>
         <item name="alertDialogTheme">@style/DialogStyleNormal</item>
         <item name="android:windowEnableSplitTouch">false</item>
+        <item name="android:navigationBarColor">@android:color/transparent</item>
```
mcomella, line 23: Why is this necessary? (add a comment)

app/src/main/res/anim/placeholder_animation.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<translate
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    android:fromXDelta="0"
+    android:toXDelta="0"
+    android:duration="100" />
```
mcomella, line 9: Does having a duration slow us down by 100ms?

app/src/migration/java/org/mozilla/fenix/MigrationDecisionActivity.kt
```diff
@@ -39,6 +39,6 @@ class MigrationDecisionActivity : Activity() {
         // and then we switch to the actual activity without an animation. This visually looks like
         // a faster start than launching this activity invisibly and switching to the actual
         // activity after that.
-        overridePendingTransition(0, 0)
+        overridePendingTransition(0, R.anim.placeholder_animation)
```
mcomella, line 42: Can you elaborate on the comment on why having a placeholder is helpful?

app/src/main/res/values/styles.xml
MarcLeclair: It lessens the "blow" when transitioning since the bottom bar isn't switching from grey to white.

app/src/main/res/anim/placeholder_animation.xml
MarcLeclair: No it doesn't , it gives it a max amount of time to make a frame iirc.

app/src/migration/java/org/mozilla/fenix/MigrationDecisionActivity.kt
MarcLeclair: Having the placeholder animation seems to force the app to create an animation where the `0` just shuts it down before the other activity starts. This was more of me playing around with the animation and finding a place holder. This is more a conclusion I came to from playing around with the different animations

boek:
This looks good. Have we tested the flow going to the migration activity? I'm assuming it's exactly the same but just want to double check :)

MarcLeclair:
@boek  I have tested locally but I don't have any explicit migration accounts ( I just put `if(true)` and forced the migration activity). It worked fine, not sure if that's enough testing 😛

PR #11821, opened 2020-06-22
For 11660: fixing nits for previous #11668

MarcLeclair:
Fixing nits for #11668 

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
+++ b/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
@@ -257,8 +257,9 @@ open class FenixApplication : LocaleAwareApplication(), Provider {
         // no-op, LeakCanary is disabled by default
     }
 
-    // This is for issue https://github.com/mozilla-mobile/fenix/issues/11660. We prefetch our info for startup
-    // so that we're sure that we have all the data available as our fragment is launched.
+    /**
+     * See [TopsiteStore.prefetch] for details on pre fetching.
+     */
     private fun prefetchForHomeFragment() {
         StrictMode.allowThreadDiskReads().resetPoliciesAfter {
             components.core.topSiteStorage.prefetch()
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mozilla/fenix/components/TopSiteStorage.kt
```diff
@@ -88,8 +88,15 @@ class TopSiteStorage(private val context: Context) {
         }
     }
 
+    /**
+     * This is for issue https://github.com/mozilla-mobile/fenix/issues/11660. We prefetch the top
+     * sites for startup so that we're sure that we have all the data available as our fragment is
+     * launched. This method doesn't affect performance since the [getTopSites] runs on the a
```
mcomella, line 94: "This method doesn't affect performance" – this seems to contradict the point of this method which is to affect performance (positively).

We probably shouldn't say, "This method doesn't negatively affect performance" because we don't know that for sure (all code that runs negatively affects performance, technically) but we can just mention the facts (this runs on a background thread) and let others draw conclusions from them.

app/src/main/java/org/mozilla/fenix/components/TopSiteStorage.kt
```diff
@@ -88,8 +88,15 @@ class TopSiteStorage(private val context: Context) {
...
+     * background thread. ObserveOnce also removes our observer once the value of the
```
mcomella, line 95: nit: I don't think the comment about ObserverOnce is necessary – it's kind of redundant to the code.

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -229,13 +229,15 @@ open class FenixApplication : LocaleAwareApplication() {
         // no-op, LeakCanary is disabled by default
     }
 
-    // This is for issue https://github.com/mozilla-mobile/fenix/issues/11660. We prefetch our info for startup
-    // so that we're sure that we have all the data available as our fragment is launched.
+    /**
+     * [TopsiteStore.prefetch] top sites for home fragment startup.
```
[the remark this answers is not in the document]
mcomella: Actually, this should probably say somethingl ike, `See [method] for details on prefetching` or something like that. This comment should be explicit that the details are to be seen elsewhere

liuche:
@MarcLeclair when you get the chance could you land this? Thanks!

liuche:
This PR actually got backed out and will need to be re-landed

jonalmeida:
We can re-land this PR without the changes to top sites storage with live data since that doesn't exist anymore.

PR #11897, opened 2020-06-24
No issue: Update Flank to v20.06.2

AaronMT:
Update Flank to [v20.06.2](https://github.com/Flank/flank/releases/tag/v20.06.2)

This release has a fix which stores `@Ignore` tests in the Unit XML without sending ignored tests to FTL.

On performance:

> It’s definitely faster, the run is still only as fast as the slowest shard. I see this as more of a correctness win than performance, although both are improved.


AaronMT:
ui-test-x86-debug: https://github.com/mozilla-mobile/fenix/issues/11901

mcarare:
> ui-test-x86-debug: #11901

Put up https://github.com/mozilla-mobile/fenix/pull/11904 for this.
But AFAIK the `collectionItemInfo` should never be null for an item that belongs to a collection.

PR #11952, opened 2020-06-25

MickeyMoz:


pocmo:
@ekager What means `needs-landing` in this case? Was CI still running at that time?

The next time @MickeyMoz runs it will force-push the update to this PR and may override your patch :)

ekager:
Yep! `needs-landing` just means to me `land this as soon as CI passes` :)
We made it before the next push though!

jonalmeida:
Thanks for fixing this @ekager ! You got to it before I could. :)

Issue #12007, opened 2020-06-26
FNX2-13037 ⁃ [Design] Follow ux mock on Settings --> Search screen


hakkikaancaliskan:
Breakdown from issue #7065
![resim](https://user-images.githubusercontent.com/17825767/85984332-b8636500-b9f1-11ea-8cd6-b76f73e51bce.png)

hakkikaancaliskan:
I'm working on this

hakkikaancaliskan:
Looks like add search engine radio groups using wrap content for height.
![resim](https://user-images.githubusercontent.com/17825767/86454685-490aaf80-bd28-11ea-8bc9-7f1b8ab3ff26.png)

abodea:
Verified as fixed on the latest Nightly build **7/7** with **Google Pixel 4 XL (10).**
![Screenshot_20200707-173522](https://user-images.githubusercontent.com/42831109/86802055-4fc55980-c07d-11ea-9a05-85bd8c687b99.png)
![Screenshot_20200707-173527](https://user-images.githubusercontent.com/42831109/86802062-505df000-c07d-11ea-80cc-85e91f39b7ce.png)

PR #12009, opened 2020-06-26
For #12007: Follow ux mock on Settings --> Search screen

hakkikaancaliskan:
resimresim

### Pull Request checklist

### After merge
- [x] **Milestone**: Make sure issues finished by this pull request are added to the [milestone](https://github.com/mozilla-mobile/fenix/milestones) of the version currently in development.

1. click on Show All Checks,


app/src/main/res/layout/preference_cat_style.xml
```diff
@@ -7,7 +7,8 @@
     style="?android:attr/listViewStyle"
     android:layout_marginTop="16dp"
     android:textAlignment="viewStart"
-    android:layout_marginStart="16dp"
+    android:layout_marginStart="72dp"
```
hakkikaancaliskan, line 10: This cat_style used in this layouts but no worry because this change will align all preference category titles.
Maybe we can remove margin_bottom.

resim

hakkikaancaliskan: Annotation 2020-06-26 201334
Annotation 2020-06-26 201420

Yes, we need to remove margin_bottom.

hakkikaancaliskan:
cc @brampitoyo ready for review.

app/src/main/res/xml/search_preferences.xml
```diff
@@ -29,41 +29,47 @@
...
             android:key="@string/pref_key_search_browsing_history"
             android:title='@string/preferences_search_browsing_history'
+            android:layout="@layout/switch_preference"
             app:iconSpaceReserved="false" />
         <SwitchPreference
             android:defaultValue="true"
             android:key="@string/pref_key_search_bookmarks"
             android:title='@string/preferences_search_bookmarks'
+            android:layout="@layout/switch_preference"
             app:iconSpaceReserved="false" />
```
NotWoods, line 51: Does the iconSpace trick work here too?

hakkikaancaliskan: Yes, it worked.

hakkikaancaliskan:
@NotWoods 
resim
This one is Android's default checkbox.
resim
This one is from ux.

Android's one has more padding, so it looks like misaligned. How can i achieve this?

hakkikaancaliskan:
resim
Other views have this problem too. So i will not care with it.

app/src/main/res/xml/search_preferences.xml
```diff
@@ -28,42 +27,36 @@
...
-            app:iconSpaceReserved="false"
-            app:allowDividerAbove="false"/>
+            app:allowDividerAbove="false" />
         <CheckBoxPreference
             android:defaultValue="false"
             android:dependency="@string/pref_key_show_search_suggestions"
             android:key="@string/pref_key_show_search_suggestions_in_private"
             android:title="@string/preferences_show_search_suggestions_in_private"
-            android:layout="@layout/checkbox_left_preference"
+            android:layout="@layout/checkbox_left_sub_preference"
```
NotWoods, line 58: Let's leave precise checkbox alignment for another PR as you said, and just use icon space here

hakkikaancaliskan: @NotWoods using icon space takes icon to the right. It's on the left in ux mock. Is there a way to do this?

hakkikaancaliskan: If i use checkbox_left_preference icon is aligned to back arrow, not top bar. So looks like i must use checkbox_left_sub_preference. What do you think?

NotWoods: My bad, I didn't catch that the icon space was already reserved.

From what I can tell, the checkbox is already in the correct position and doesn't need to be moved. The design changes focus on the search engine icons and headers.

brampitoyo:
@hakkikaancaliskan Looks good to me!

hakkikaancaliskan:
@NotWoods solved conflict. Ready to land.

NotWoods:
Hi @hakkikaancaliskan, do you mind removing the custom checkbox layout? As I mentioned in a comment, I think the checkbox is fine currently and doesn't need to be focused on.

hakkikaancaliskan:
@NotWoods You mean checkbox_left_sub_preference.xml? It's needed because without it checkbox will be on left and aligned to top bar, we don't want this. In ux mock its aligned to upper preference.

NotWoods:
Sorry, I was looking at a different set of mocks that had the checkbox line up with the radio buttons. I'll check with @brampitoyo to see which version we should use.

hakkikaancaliskan:
> 
> 
> Sorry, I was looking at a different set of mocks that had the checkbox line up with the radio buttons. I'll check with @brampitoyo to see which version we should use.

Look to the a little upper :)
@brampitoyo confirmed this ux

brampitoyo:
@NotWoods @hakkikaancaliskan I can confirm that the checkbox “Show in private browsing” is indeed meant to look nested. It doesn’t align with other checkboxes on the page.

This is because, if the toggle above is turned off, the checkbox should be disabled (ie. can’t be selected).

@hakkikaancaliskan Last question: what’s the text margin we use for this special checkbox?

- Is the text margin identical to the one we use for other, non-special checkbox? 
- Or, is the text margin identical to the margin between the search provider icon and the name (it’s a little narrower)?
- Or, is the text margin a custom value that’s totally different from everything else?

Ideally, we don’t want any custom value. Let’s use margins that are already in use, somewhere else in the app.

hakkikaancaliskan:
> @hakkikaancaliskan Last question: what’s the text margin we use for this special checkbox?
> 
>     * Is the text margin identical to the one we use for other, non-special checkbox?
> 
>     * Or, is the text margin identical to the margin between the search provider icon and the name (it’s a little narrower)?
> 
>     * Or, is the text margin a custom value that’s totally different from everything else?
> 
> 
> Ideally, we don’t want any custom value. Let’s use margins that are already in use, somewhere else in the app.

There is 8dp margin between checkbox and title currently but while checking ux again its 12 in ux. I'm gonna fix it. This isn't about to checkbox misalignment.

Let me explain why that checkbox looks misaligned.

https://github.com/mozilla-mobile/fenix/pull/12009#issuecomment-651319324 as you can see in this comment Android's default checkbox has so much padding (we edit anything). There is same problem on delete browsing data screen and delete browsing data on quit.
If we move the custom checkbox view to left like 4dp, we could hide that padding or we should find a way to remove that padding.

hakkikaancaliskan:
@brampitoyo When i properly apply ux mocks to "+ add search engine" here is the result.
resim

+ is aligned with back arrow but radio buttons have problem too.

hakkikaancaliskan:
@NotWoods looks like we have some real problems with radiobutton and checkboxes. Could you look at it please?

hakkikaancaliskan:
@brampitoyo This is the exact copy of ux mock but there is padding problem with radiobutton too :(
Screenshot_20200701-000726_Firefox_Preview

hakkikaancaliskan:
Radio button and checkboxes have padding start more than what we have in ux mocks.

hakkikaancaliskan:
@brampitoyo @NotWoods 
**Found the root of the problem**: Android's default radio button and checkbox is 32x32, not 24x24 like ux mock.
Thus if we force 24x24 it gets a cut from right.

hakkikaancaliskan:
If we decrease start margin 4dp, it fits perfectly.
Screenshot_20200701-015621_Firefox_Preview

brampitoyo:
@hakkikaancaliskan This looks almost right to me!

Then I have just three minor fixes:

1. The checkbox should vertically align with the search provider icon
2. The checkbox text label should vertically align with the search provider text label
3. The search provider text label (Google, Amazon, Bing, DuckDuckGo, Yandex, Twitter, Wikipedia, etc.) and “Add new search engine” should be the same size as the rest of the text on this page – not smaller. They should have the same 48dp height, too – just like other menu items.



With these fixed, we’re all good to ship the changes.

hakkikaancaliskan:
@brampitoyo as i said before Android's default checkbox and radiobutton 32x32 sooo i needed to change margins to match ux mock.
Is this okey?

@NotWoods ready to review

resim

brampitoyo:
Looks great, @hakkikaancaliskan!

liuche:
thank you! opening here: https://github.com/mozilla-mobile/fenix/pull/12177

PR #12023, opened 2020-06-26
Test settings about/advanced

NotWoods:
Additionally:
- Since `AboutPageItem` is a sealed class with a single subclass, its been simplified it to a single class.
- `SecretSettingsPreference` -> `SecretSettingsFragment`


hakkikaancaliskan:
Test God 😛

PR #12055, opened 2020-06-27
Configure M3 AMO collection for all builds

csadilek:
This can come in with the next A-C upgrade (48.0.0 and GV 79 beta). Marking as "Do Not Land" until A-C 48.0.0.

M3 collection is currently on Fenix/Fennec Nightly and debug, with this we make it the default for all builds (same we did with M2 before).


liuche:
cc @pocmo can you make sure this is part of the AC 48.0.0 release, that will go out along with the GV 79 update tomorrow?

cadeyrn:
@liuche This did not land on the 79 branch and the new add-ons are not available in Firefox 79 Beta 5. Is this expected? Your last comment sounds like you would like to have the add-ons in Firefox 79 Beta.

Issue #11991, opened 2020-06-26
FNX-13001 ⁃ [Bug] 0 tabs counter after the migration from Fennec to Fenix


sv-ohorvath:
## Steps to reproduce
1. Open a few tabs in Fennec Beta.
1. Install the migration build.
1. Check the tabs counter after the migration, when you are still on the Fenix onboarding screen
    1. Expected behavior
The correct number of tabs is shown.
    1. Actual behavior
The counter shows 0 tabs.
The number of tabs is shown after you dismiss the onboarding screen.
    1. Device information

* Android device: ?Huawei P20 lite (Android 9)
* Fenix version: Beta 78.0.0

ebalazs-sv:
This issue is reproducible on Beta Migration 78.0.1-beta.1 GV 78 from 6/27 with Motorola Moto G6 (Android 8).

ebalazs-sv:
Verified as fixed on Beta Migration 79.0.0-beta.1 GV 79 from 6/30 with Motorola Moto G6 (Android 8), Huawei P20 lite (Android 9).
I will remove the qa:needed label. Thanks!

PR #12097, opened 2020-06-29
For #11991 - Show tab count during onboarding

ekager:
### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
@@ -30,6 +30,7 @@ import androidx.constraintlayout.widget.ConstraintSet.TOP
 import androidx.coordinatorlayout.widget.CoordinatorLayout
 import androidx.core.content.ContextCompat
 import androidx.core.view.doOnLayout
+import androidx.core.view.isVisible
 import androidx.core.view.updateLayoutParams
 import androidx.fragment.app.Fragment
 import androidx.fragment.app.activityViewModels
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -225,10 +225,10 @@ class HomeFragment : Fragment() {
         activity.themeManager.applyStatusBarTheme(activity)
 
         view.consumeFrom(requireComponents.core.store, viewLifecycleOwner) {
-            val tabCount = if (currentMode.getCurrentMode() == Mode.Normal) {
-                it.normalTabs.size
-            } else {
+            val tabCount = if (browsingModeManager.mode.isPrivate) {
```
grigoryk, line 228: An exhaustive `when` would have prevented this bug, but i guess we can't do that w/ sealed classes?

grigoryk: Actually, you can switch to an exhaustive `when` here, so let's do that instead.

ekager: We were just using the wrong way to get the mode. `getCurrentMode()` returns `Mode` -> private, normal, onboarding https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/java/org/mozilla/fenix/home/Mode.kt#L21-L24

Now this is just a boolean (which is how we use it everywhere else that doesn't care about onboarding status): https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/java/org/mozilla/fenix/browser/browsingmode/BrowsingModeManager.kt#L18

grigoryk: Ah, cool -  i've misread that `if` statemen. That's what i get for writing code, reviewing code and listening to a meeting all at once.

PR #11940, opened 2020-06-25
for #11830 added new metric for collecting startup method from all startup phases

sraturi:
### Pull Request checklist

- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features. In addition, it includes a screenshot of a successful [accessibility scan](https://play.google.com/store/apps/details?id=com.google.android.apps.accessibility.auditor&hl=en_US) to ensure no new defects are added to the product.

### After merge
- [ ] **Milestone**: Make sure issues finished by this pull request are added to the [milestone](https://github.com/mozilla-mobile/fenix/milestones) of the version currently in development.

### To download an APK when reviewing a PR:
1. click on Show All Checks,
2. click Details next to "Taskcluster (pull_request)" after it appears and then finishes with a green checkmark,
3. click on the "Fenix - assemble" task, then click "Run Artifacts".
4. the APK links should be on the left side of the screen, named for each CPU architecture

Issue #11830, opened 2020-06-22
[Telemetry] Add telemetry to record how the app was opened on cold/warm/hot startup(home screen, applink, custom tab)


sraturi:
## Description & Product Manager / Data Scientist User Story
We'd like to have this telemetry so we're able to create performance tests that are representative of user experiences.

## What questions will you answer with this data?
How do users open the app most frequently?

## Acceptance Criteria
* [ ] ENG files a [DS JIRA](https://jira.mozilla.com/projects/DO/issues/DO-228?filter=allopenissues) request outlining their methodology.
* [ ] DS sign off on instrumentation methodology addressing product questions.
* [ ] Event pings can be queried via re:dash
* [ ] Event pings can be queried via amplitude
* [ ] We are sending telemetry events for the actions listed in the requirements
* [ ] We have documented the telemetry
* [ ] We have asked a data steward to [review](https://github.com/mozilla/data-review/blob/master/request.md) the telemetry

sraturi:
Reopening issue again to find a proper solution

PR #12114, opened 2020-06-29
Telemetry 11830, track how app was opened on all startup

sraturi:
### Pull Request checklist

### After merge

1. click on Show All Checks,


sraturi:
# Test Flows
**_(startup type, app open tracked value) example: (warm startup, app_icon)_**

- Start app from app icon (cold startup, app_icon) -> press home button ->press the app icon again(warm startup, app_icon)
- Start app from app icon (cold startup, app_icon) -> press home button ->open Gmail link [open to browser option](warm startup, link)
- open Gmail link[in app browser](cold startup, custom_tab)-> press home button -> open twitter link [in app browser] (cold startup, custom_tab)
    - This always begins with a cold startup... I'm assuming because Each time, we create a new ExternalAppBrowserActivity instance in the respective in-app processes?
- open Gmail link [open to browser option](cold startup, link) -> press home button ->open twitter link [open to browser option](warm startup, link)
- Start app from app icon (cold startup, app_icon) -> adb intent to open link (nothing tracked since app is already in foregound)
- Start app from app icon (cold startup, app_icon) -> press home button -> adb intent to open link (warm startup, startup recorded as link). 

**adb intent command**: `adb shell am start -t 'text/html' -d 'https://apple.com' -a android.intent.action.VIEW org.mozilla.fenix.debug/org.mozilla.fenix.IntentReceiverActivity`

sraturi:

All questions are mandatory. You must receive review from a data steward peer on your responses to these questions before shipping new data collection.

**1.  What questions will you answer with this data?**
		How do users open the app (custom tab, app_icon, link, unknown) most frequently on all startups (cold, warm, hot)

**2.  Why does Mozilla need to answer these questions? Are there benefits for users? Do we need this information to address product or business requirements? Some example responses:**
	Creating accurate understanding of real world usage to prioritize performance improvements  
  

**3.  What alternative methods did you consider to answer these questions? Why were they not sufficient?**
    There is an existing app_opened metrics, However it is only used for cold startup.  
  

**4.  Can current instrumentation answer these questions?**
    no  
      
    

5.  **List all proposed measurements and indicate the category of data collection for each measurement, using the [Firefox data collection categories](https://wiki.mozilla.org/Firefox/Data_Collection) found on the Mozilla wiki.**
   data is Interaction data.How the user opens app , options -> custom tab, app_icon, link  
  

**6.  Please provide a link to the documentation for this data collection which describes the ultimate data set in a public, complete, and accurate way.**
   This is documented in the metrics.md file

**7.  How long will this data be collected?**
    Until 2020-12-01  
  

**8.  What populations will you measure?**
All release, beta, and nightly users with telemetry enabled.  
  

**9.  If this data collection is default on, what is the opt-out mechanism for users?**
default on, opt out by Settings ->Data collection

**10.  Please provide a general description of how you will analyze this data.**
analyze so we can set up better tests and prioritize analyses to improve performance in the most impactful areas.

**11.  Where do you intend to share the results of your analysis?**
Glean, Amplitude and with mobile teams.  
  

**12.  Is there a third-party tool (i.e. not Telemetry) that you are proposing to use for this data collection? If so:**
N/A

@liuche  This is the same data review form from https://github.com/mozilla-mobile/fenix/pull/11940.  Made the PR to more accurately capture the data for the ticket #11830 
 I will add this PR to the data review part of the metric

app/src/main/java/org/mozilla/fenix/components/metrics/AppAllSourceStartTelemetry.kt
```diff
@@ -0,0 +1,59 @@
...
+    private val isExternalAppBrowserActivity: Boolean
+) {
+
+    /**
+     * Track the event on a cold startup
+     */
+    init {
+        intent?.toSafeIntent()?.let(::getIntentAllStartSource)
+            ?.also {
+                metrics.track(Event.AppOpenedAllSourceStartup(it))
```
mcomella, line 30: This code is inconsistent: why do we track one event as part of object initialization and another as part of a function call?

I think it'd be simpler to track both events in functions – code run in object initialization and passed into the initializer can impact the entire class but a function on a class with no mutable variables will generally only impact itself.

Additionally, in general I don't like creating side effects in object initialization but to be honest, I'm struggling to put my finger on why. Conceptually, I think of object initialization as a function to allocate memory and configure an object: I guess it makes the method have more than one responsibility to do more than that. It also makes the object harder to mock (if applicable) because the initializer will run in some mocking frameworks before other methods. It's also less flexible – e.g. if you block the thread to init an object, you can't use the object concurrently until it's done. Unfortunately, I guess it's just a few hand-wavy reasons for this so maybe I'm wrong but I still think we're handling these inconsistently and that handling the functionality through functions is simpler.

---

If we handle this as a function, we can further simplify this by initializing this object as part of `Components`. This is simpler because the lifecycle of the object becomes the same as the Application, i.e. for practical purposes, forever, instead of the lifecycle of the Activity.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -197,6 +195,12 @@ open class HomeActivity : LocaleAwareAppCompatActivity() {
         }
 
         captureSnapshotTelemetryMetrics()
+
+        setupAppAllStartTelemetry()
+    }
+
+    protected open fun setupAppAllStartTelemetry() {
+        appAllSourceStartTelemetry = AppAllSourceStartTelemetry(intent, components.analytics.metrics, false)
```
mcomella, line 203: What if the app is backgrounded and HomeActivity needs to be recreated? We need to ensure we're checking backgrounded here too, right? This is essentially the same case as `onNewIntent`, right? Can we just use the same function for both?

Also this doesn't cover the case that HomeActivity isn't launched at all. Better yet, could we use `ProcessLifecycleOwner.onStart` for this?

sraturi: If the app is backgrounded, and HomeActivity is recreated, `setupAppAllStartTelemetry` would get created again since I'm calling `setupAppAllStartTelemetry()` on HomeActivity's onCreate method. which would ensure that we are still tracking the event?

Im not sure what you mean by 
> This is essentially the same case as onNewIntent, right? Can we just use the same function for both?

The `onNewIntent` is not called when activity is first launched. 

What would be a case where `HomeActivity` is not launched?

app/src/main/java/org/mozilla/fenix/components/metrics/AppAllSourceStartTelemetry.kt
sraturi: That makes sense. I will update this to make it more consistent so that we dont have side effects on init.

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -55,7 +60,7 @@ import org.mozilla.fenix.utils.Settings
  *  Installs [CrashReporter], initializes [Glean]  in fenix builds and setup Megazord in the main process.
  */
 @Suppress("Registered", "TooManyFunctions", "LargeClass")
-open class FenixApplication : LocaleAwareApplication() {
+open class FenixApplication : LocaleAwareApplication(), LifecycleObserver {
```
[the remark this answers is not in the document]
sraturi: Gotcha! I moved the lifecycle observer to `AppAllSourceStartTelemetry`

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -407,4 +413,10 @@ open class FenixApplication : LocaleAwareApplication() {
     companion object {
         private const val KINTO_ENDPOINT_PROD = "https://firefox.settings.services.mozilla.com/v1"
     }
+
+    @OnLifecycleEvent(Lifecycle.Event.ON_STOP)
+    fun applicationInBackground() {
+        // let AppAllSourceStartTelemetry know application is in background
```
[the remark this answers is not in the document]
sraturi: oops

app/src/main/java/org/mozilla/fenix/components/metrics/AppAllSourceStartTelemetry.kt
sraturi: So I moved the `LifecycleObserver` inside `AppAllSourceStartTelemetry`. The only thing inside init now is attaching our class to `ProcessLifecycleOwner` 's observers list.

app/metrics.yaml
```diff
@@ -8,18 +8,12 @@ no_lint:
...
-      changes such as loading a url. This can be used loosely as a
-      heuristic for when the user requested to open the app. The
-      HomeActivity encompasses the home screen and browser screen but
-      may include other screens. This differs from the app_opened
+  app_opened_all_startup:
+    type: event
+    description: |
+      A user opened the app to the HomeActivity. The HomeActivity
+      encompasses the home screen and browser screen but may
+      include other screens. This differs from the app_opened
```
liuche, line 16: What other screens could this include, could you provide more information? I want to make sure that someone using this probe understands the cases where this could happen. Update this documentation if there's more information.

Do the `extra_keys` values remain the same?

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
mcomella: > Better yet, could we use ProcessLifecycleOwner.onStart for this?

I think my other comments are irrelevant if we use `ProcessLifecycleOwner` instead of attaching to `HomeActivity` directly.

> What if the app is backgrounded and HomeActivity needs to be recreated?

Sorry, I meant if the app is not backgrounded and HomeActivity needs to be recreated. I'm not sure this is possible without Don't Keep Activities in the current code base but writing code to protect against this (assuming it's not too time consuming) is valuable.

app/metrics.yaml
sraturi: Other screens could include settings, collections? basically any fragment in the nav_graph. I will update the doc with some more info. 

`extra_keys` value are the same as for `app_received_intent`

app/metrics.yaml
```diff
@@ -8,18 +8,12 @@ no_lint:
...
-      The system received an Intent for the HomeActivity. An intent
-      is received an external entity wants to the app to display
-      content. Intents can be received when the app is closed – at
-      which point the app will be opened – or when the app is
-      already opened – at which point the already open app will make
-      changes such as loading a url. This can be used loosely as a
-      heuristic for when the user requested to open the app. The
-      HomeActivity encompasses the home screen and browser screen but
-      may include other screens. This differs from the app_opened
+  app_opened_all_startup:
```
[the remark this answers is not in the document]
sraturi: I put back the original `on_intent_received `metric. I fixed the spelling typo for `received` so its showing up in the changes.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
sraturi: ok I think i added all the changes we discussed.

app/src/main/java/org/mozilla/fenix/components/metrics/AppAllSourceStartTelemetry.kt
```diff
@@ -0,0 +1,58 @@
...
+ * we only track the event if app is launching first time or was in background state because
+ * we don't want random intents with [Intent.ACTION_VIEW] to be tracked as app opened
+ * For example, Start app through APP_ICON -> run adb intent command to open a link.
+ * We want to ignore these adb commands since app is already on foreground.
+ */
+class AppAllSourceStartTelemetry(
+    private val metrics: MetricController
+) : LifecycleObserver {
+
+    private var wasApplicationInBackground = true
```
mcomella, line 25: When I see files with logic like this, where it's easy to flip a flag and mess up the behavior, I like to write tests to ensure I got it all right. Can we write a test for it?

app/src/main/java/org/mozilla/fenix/components/metrics/AppAllSourceStartTelemetry.kt
```diff
@@ -0,0 +1,58 @@
...
+    private fun setAppOpenedAllSourceFromIntent(intent: SafeIntent, isExternalAppBrowserActivity: Boolean) {
+        if (!wasApplicationInBackground)
+            return
+
+        when {
+            isExternalAppBrowserActivity -> Event.AppOpenedAllSourceStartup.Source.CUSTOM_TAB
+            intent.isLauncherIntent -> Event.AppOpenedAllSourceStartup.Source.APP_ICON
+            intent.action == Intent.ACTION_VIEW -> Event.AppOpenedAllSourceStartup.Source.LINK
+            else -> Event.AppOpenedAllSourceStartup.Source.UNKNOWN
+        }.also {
```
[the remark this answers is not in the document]
sraturi: Ah Gotcha. Thanks!

app/src/test/java/org/mozilla/fenix/components/metrics/AppAllSourceStartTelemetryTest.kt
```diff
@@ -0,0 +1,73 @@
...
+import android.content.Intent
+import io.mockk.MockKAnnotations
+import io.mockk.every
+import io.mockk.impl.annotations.RelaxedMockK
+import io.mockk.verify
+import mozilla.components.support.utils.toSafeIntent
+import org.junit.Before
+import org.junit.Test
+
+internal class AppAllSourceStartTelemetryTest {
```
mcomella, line 12: The comment you're looking for about any missing tests. :) Do you know how to run code coverage tools? They're really helpful in identifying lines that weren't run during the tests. After running it, I see:
image

The `onApplicationOnStop` method was never called (note the red bars to the left!). That means we never tested that functionality.

From a theory perspective, I'm also concerned that this isn't tested: the boolean logic is the most tricky thing in here so we should **write a test to ensure resetting works as expected.**

**Additionally, there's one more test I'd want to write from a theory perspective: `UNKNOWN`.** This shows a limitation in code coverage tools: since they only show lines you've executed during the test, not what's actually tested, one test passes in UNKNOWN so that line is executed even if we never assert for it.

So please write those two tests and try out the code coverage tools if you haven't done so already! :) It's the shield button in AS, next to the debug button.

If we wanted to be really tested, we could also test what happens if `onApplicationOnStop` is called before an intent is received... but that's probably overkill. :)

sraturi: Thanks! @mcomella  Ive updated the test cases based on this comment

app/metrics.yaml
```diff
@@ -8,6 +8,36 @@ no_lint:
   - CATEGORY_GENERIC
 
 events:
+  app_opened_all_startup:
+    type: event
+    description: |
+      A user opened the app to the HomeActivity. The HomeActivity
+      encompasses the home screen, browser screren, settings screen,
```
mcomella, line 15: nit: typo -> screen

app/src/main/java/org/mozilla/fenix/components/metrics/AppAllSourceStartTelemetry.kt
```diff
@@ -0,0 +1,59 @@
...
+ * Tracks how the application was opened through [Event.AppOpenedAllSourceStartup].
+ * We only considered to be "opened" if it received an intent and the app was in the background.
+ */
+class AppAllSourceStartTelemetry(private val metrics: MetricController) : LifecycleObserver {
+
+    // default value is true to capture the first launch of the application
+    private var wasApplicationInBackground = true
+
+    init {
+        ProcessLifecycleOwner.get().lifecycle.addObserver(this)
```
mcomella, line 26: No action needed: since this is lazily loaded and we only attach the `ProcessLifecycleOwner` in the initializer, it's possible we may miss `onStop` if the code is refactored such that we don't get called in `*Activity.onCreate`. Ideally, we'd separate the initialization from the first call to help avoid this but let's skip it because I don't think it's worth the time or additional churn.

mcomella:
Note: @liuche previously approved the data review https://github.com/mozilla-mobile/fenix/pull/12114#pullrequestreview-445245341 and sraturi mentioned re-requesting review was an accident so I don't think we need additional data review: merging.

Issue #3481, opened 2019-06-14
FNX3-17048 ⁃ Switch Tabs via Swiping Address Bar


vrudesai:
### Why/User Benefit/User Problem
As a user I want to be able to quickly switch back and forth between tabs by swiping right and left on the address bar.

For left-to-right languages the oldest tab is on the left and a left swiping brings you to your most recent tabs.

For right-to-left languages the oldest tab is on the right and right swiping brings you to your most recent tabs.

Neikon:
Gives us this feature pls pls pls

Sergiu-Ghita:
I'd like to work on it

KazaHesto:
This should be tested with Android Q to see how it interacts with the new navigation gestures.

kbrosnan:
Dupe of #1283 / #176

ChristianCiach:
@kbrosnan
It may be a dupe, but please compare the upvotes of this issue to the issues this is a dupe of. The discoverability of this issue is way higher, so please don't ignore this issue when assessing the importance of this feature.

vesta0:
@mozilla-mobile/ux please consider this for the tab management re-design

lime124:
removing the ux label - this is on our radar already as it's part of Q3 plans.

vesta0:
Related GV bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1509249

sr0n:
Just had an idea that I would like to mention - I would also like this feature implemented but quite a few people I assume also like being able to use this gesture to see the full URL as is currently the case in Fennec. Could there be an option to switch between the two if they are ever both added?

yoasif:
> @kbrosnan
> It may be a dupe, but please compare the upvotes of this issue to the issues this is a dupe of. The discoverability of this issue is way higher, so please don't ignore this issue when assessing the importance of this feature.

Pretty sure it is a dupe of https://github.com/mozilla-mobile/fenix/issues/396

scratchmex:
Following #10548 , the gesture to swipe down (or up) the address bars to trigger the tabs overview could be useful as well.

person808:
Starting to look into this now. One question for UX: do we want this gesture to work on the home page and if so, how would it work (what tabs would swiping go to)? Its a bit weird since the home page isn't a "real" tab but we do show a toolbar there.

yoasif:
[...]

PR #12174, opened 2020-07-01
For #3481 - Implement swipe on toolbar to switch tabs.

person808:
Things to do
- [x] Swap swipe direction for rtl layout as described in the issue
- [x] Requires https://github.com/mozilla-mobile/android-components/pull/7534 for full-size thumbnail loading
- [x] Check if the tab preview toolbar needs to be a ConstraintLayout (double check rtl layout)
- [x] Fix thumbnail size when the thumbnails are taken from a different device orientation (ex. portrait thumbnails in landscape mode) (edit: reused the TabThumbnailView to scale the thumbnails)

Video - https://streamable.com/tk27vq

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/browser/TabPreview.kt
+++ b/app/src/main/java/org/mozilla/fenix/browser/TabPreview.kt
@@ -8,10 +8,11 @@ import android.content.Context
 import android.util.AttributeSet
 import android.view.Gravity
 import android.view.LayoutInflater
+import android.view.View
 import android.widget.FrameLayout
-import androidx.core.content.ContextCompat
 import androidx.core.content.res.ResourcesCompat
 import androidx.core.view.updateLayoutParams
+import kotlinx.android.synthetic.main.mozac_ui_tabcounter_layout.view.*
 import kotlinx.android.synthetic.main.tab_preview.view.*
 import mozilla.components.browser.thumbnails.loader.ThumbnailLoader
 import mozilla.components.support.images.ext.loadIntoView
@@ -44,12 +45,9 @@ class TabPreview @JvmOverloads constructor(
             )
         }
 
-        menuButton.setColorFilter(
-            ContextCompat.getColor(
-                context,
-                ThemeManager.resolveAttribute(R.attr.primaryText, context)
-            )
-        )
+        // Change view properties to avoid confusing the UI tests
+        tab_button.counter_box.id = View.NO_ID
+        tab_button.counter_text.id = View.NO_ID
     }
 
     override fun onLayout(changed: Boolean, left: Int, top: Int, right: Int, bottom: Int) {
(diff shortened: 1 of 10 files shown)


person808:
Updated with (hopefully) the final revision to my A-C PR. Tests will fail because the A-C changes break the tab tray. Will fix when we pick up the A-C changes.

app/src/main/java/org/mozilla/fenix/browser/TabPreview.kt
```diff
@@ -0,0 +1,67 @@
...
+            )
+        )
+    }
+
+    override fun onLayout(changed: Boolean, left: Int, top: Int, right: Int, bottom: Int) {
+        super.onLayout(changed, left, top, right, bottom)
+        previewThumbnail.translationY = if (!context.settings().shouldUseBottomToolbar) {
+            fakeToolbar.height.toFloat()
+        } else {
+            0f
```
boek, line 60: linting might complain about this magic number

app/src/main/java/org/mozilla/fenix/browser/ToolbarGestureHandler.kt
```diff
@@ -0,0 +1,374 @@
...
+    private val defaultVelocity = TypedValue.applyDimension(
+        TypedValue.COMPLEX_UNIT_DIP,
+        DP_PER_SECOND,
+        activity.resources.displayMetrics
+    )
+
+    private var gestureDirection = GestureDirection.LEFT_TO_RIGHT
+
+    init {
+        CoroutineScope(Main).launch {
```
boek, line 85: Are we launching here for a race condition? Is there another way we can fix it? (If not, we should probably at least attach it to a lifecycleScope)

app/src/main/java/org/mozilla/fenix/browser/TabPreview.kt
person808: Lint doesn't care it the magic number is 0 😄

app/src/main/java/org/mozilla/fenix/browser/ToolbarGestureHandler.kt
person808: Went with using a lifecycleScope. In theory the rootview should already be attached so we can get the insets right away, but I don't want to risk that not always being the case and crashing.

person808:
Just an FYI for anyone on the perf team testing this. This requires building against a local AC that has https://github.com/mozilla-mobile/android-components/commit/800b465772f2faa0a48ecafd2e93c05a24e9d4ea until automation gets fixed and we pick up the next AC release

app/src/main/java/org/mozilla/fenix/browser/ToolbarGestureHandler.kt
person808: I found one scenario where the insets can change (switching from gesture -> three-button nav or vice-versa) so I decided to get the insets everytime I need to check them. To do that I changed `getInsets` so it'll just return null if the `decorView` isn't attached. In theory, we shouldn't run into this scenario and we'll be able to get the insets.

andreicristianpetcu:
Can somebody please trigger a nightly build? Can't wait to test this!

Issue #10401, opened 2020-05-04
FNX-3515 ⁃ [Bug] If you have only one search engine it still shows this time search with option during search


s-ankur:
## Steps to reproduce
Delete all search engines except one
Open search

### Expected behavior
No prompt

### Actual behavior
This time seach with[the engine im already seaching with] prompt

### Device information
* Android device: Asus Zenfone M1
* Fenix version: fenix Nightly

AmyYLee:
Assigned to @brampitoyo

brampitoyo:
@s-ankur Agreed. If users only have one search engine, it doesn’t make sense to show “This time, search with” on the search screen. In fact, it doesn’t make sense to have the “🔍 Shortcuts” button.

So my recommendation is as follows:

IF only one search engine is installed
THEN don’t show or select “🔍 Shortcuts” button on the search screen

By not displaying the search shortcuts button, the heading “This time, search with” will also not show up.

s-ankur:
@brampitoyo one more suggestion, the search engine that is default shouldnt appear in this time search with, as you are already searching with it

ebalazs-sv:
Verified as fixed on Nightly 200707 GV 80, from 7/7 with Google Pixel 2 (Android 9), Huawei P9 Lite (Android 7).
I filed a new issue for the missing "Search engine settings" when there is only one search engine: https://github.com/mozilla-mobile/fenix/issues/12335.

PR #12188, opened 2020-07-02
For #10401 - Hide Shortcuts if just one search engine is installed

Mugurell:
Otherwise, the Shortcuts option which allows to choose with what search engines
to search would be redundant.

### Pull Request checklist

- [x] **Tests**: This PR includes a new test for the change.

NoShortcutsForJustOneSearchEngin

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/search/SearchFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/search/SearchFragment.kt
@@ -99,6 +99,7 @@ class SearchFragment : Fragment(), UserInteractionHandler {
 
         requireComponents.analytics.metrics.track(Event.InteractWithSearchURLArea)
 
+        val areShortcutsAvailable = areShortcutsAvailable()
         searchStore = StoreProvider.get(this) {
             SearchFragmentStore(
                 SearchFragmentState(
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mozilla/fenix/search/SearchFragment.kt
```diff
@@ -330,6 +334,13 @@ class SearchFragment : Fragment(), UserInteractionHandler {
             )
         }
 
+        // Users can from this fragment go to install/uninstall search engines and then return.
+        with(areShortcutsAvailable()) {
```
NotWoods, line 338: I don't think `with` is the right tool here, since there are no properties to work with on a boolean. We're better off just setting a `val` like above.
```suggestion
        val areShortcutsAvailable = areShortcutsAvailable()
        if (searchStore.state.areShortcutsAvailable != areShortcutsAvailable) {
            searchStore.dispatch(SearchFragmentAction.UpdateShortcutsAvailability(areShortcutsAvailable))
        }
```

app/src/main/java/org/mozilla/fenix/search/SearchFragment.kt
```diff
@@ -330,6 +334,13 @@ class SearchFragment : Fragment(), UserInteractionHandler {
...
+            if (searchStore.state.areShortcutsAvailable != this) {
+                searchStore.dispatch(SearchFragmentAction.UpdateShortcutsAvailability(this))
```
NotWoods, line 340: Perhaps this logic should be handled in the reducer? We could always dispatch, but only return a new state if the `areShortcutsAvailable` value changes.

app/src/main/java/org/mozilla/fenix/search/SearchFragment.kt
Mugurell: Thanks!
Maybe a personal preference of mine to use `with()` instead of declaring a val.
Indeed, not in the spirit of what `with()` should be used for. Will change.

app/src/main/java/org/mozilla/fenix/search/SearchFragment.kt
Mugurell: Thanks!
Thinking about it, the reducer has to return a state. It could do that check and not modify anything in the state, but it has to return a state.
State which would trigger an update for the view which could be unnecessary.
By only dispatching when we want to modify the state we can control and limit view updates.

NotWoods: You're right, I thought `consumeFrom` checked if the state had changed first.

PR #12190, opened 2020-07-02
SearchFragmentStore: Remove Session reference from state and read values from BrowserStore.

pocmo:
In AC I am preparing to slowly remove properties from `Session` in favor of using `BrowserStore`. One of the things I want to start with is `searchTerms`. Before removing anything in AC I want to migrate code in Fenix over. This patch primarily addresses `SearchFragmentStore`. There are more things that access `searchTerms` that I want to look at in future patches.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/search/SearchFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/search/SearchFragment.kt
@@ -30,6 +30,7 @@ import kotlinx.android.synthetic.main.fragment_search.*
 import kotlinx.android.synthetic.main.fragment_search.view.*
 import kotlinx.android.synthetic.main.search_suggestions_onboarding.view.*
 import kotlinx.coroutines.ExperimentalCoroutinesApi
+import mozilla.components.browser.state.selector.findTab
 import mozilla.components.browser.toolbar.BrowserToolbar
 import mozilla.components.concept.storage.HistoryStorage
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/mozilla/fenix/search/SearchFragment.kt
```diff
@@ -86,11 +87,13 @@ class SearchFragment : Fragment(), UserInteractionHandler {
...
         val args by navArgs<SearchFragmentArgs>()
-        val session = args.sessionId
-            ?.let(requireComponents.core.sessionManager::findSessionById)
+
+        val tabId = args.sessionId
+        val tab = tabId?.let { requireComponents.core.store.state.findTab(it) }
 
         val view = inflater.inflate(R.layout.fragment_search, container, false)
-        val url = session?.url.orEmpty()
+        val url = tab?.content?.url ?: ""
```
NotWoods, line 95: nit: 
```suggestion
        val url = tab?.content?.url.orEmpty()
```

As we only use searchTerms to create the state, we could just move it into the `SearchFragmentState` constructor.

```kt
SearchFragmentState(
  searchTerms = tab?.content?.searchTerms.orEmpty(),
  ...
)
```

pocmo:
> Two tests need a fix: DefaultSearchControllerTest.kt and SearchFragmentStoreTest.kt.

Weird, I remember working on them. I'll try again and update the PR.

PR #12254, opened 2020-07-03
Closes #12171: (fenixProduction) Add vector image for full Nightly logo.

### LINKED ISSUE #12171: FNX2-15060 ⁃ Update Firefox Nightly branding to include one-line logo
We're cleaning up the Firefox Nightly channels, and we need to make our branding consistent there.
ScreenshotUNITO-UNDERSCORE!20200707-135351~2!

We have the 2-line branding from before for Nightly, but need to have the 1-line version, similar to https://github.com/mozilla-mobile/fenix/issues/8278#issuecomment-637618003

cc @shorlander could you get us this updated branding? I think this should be last one, because we've already bothered you for the Production and the Beta at this point :joy:

pocmo:
All PNG icons were already the correct ones, only the vector image needed updating.


pocmo:
ui test failed: `verifyAboutFirefoxPreview`. That shouldn't be caused by the icon update. Rerunning...

Issue #11886, opened 2020-06-24
[Bug][codecov] permanent debug(T) automation timeout after codecov.io/bash update


Archaeopteryx:
Push with failures: https://treeherder.mozilla.org/#/jobs?repo=fenix&group_state=expanded&selectedTaskRun=NUL-XqGFSiqa_UZgn3YwzQ.1&resultStatus=testfailed%2Cbusted%2Cexception%2Cretry%2Cusercancel%2Crunnable&revision=673507d9b391dfe01aa028b94b482ff37bd2bd87

Log: https://treeherder.mozilla.org/logviewer.html#?job_id=307339961&repo=fenix

[task 2020-06-24T07:25:36.810Z] +++ curl -s https://codecov.io/bash

```
[taskcluster:error] Task timeout after 7200 seconds. Force killing container.
[taskcluster 2020-06-24 09:17:09.914Z] === Task Finished ===
[taskcluster 2020-06-24 09:17:09.914Z] Unsuccessful task run with exit code: -1 completed in 7226.577 seconds
```

The downloaded script has been updated from `20200602-f809a24` to `20200622-40cebbd` when it started to fail.

liuche:
Thanks for filing, and finding the cause @Archaeopteryx !

As a temporary workaround, we've disabled codecov for now in #11928 because we need to land PRs for code freeze (6/26).

From talking to Aki from the releng team, here are a couple of next steps:
- if the script is broken by running it locally, contact codecov
- potentially vendoring the script (landing it in-tree)
- or using a TC fetch task so we can compare versions of the script if it changes in the future

NotWoods:
It seems like the script is broken, [landing it in-tree](https://github.com/mozilla-mobile/fenix/pull/12258) still causes a timeout.

PR #12258, opened 2020-07-03
For #11886 - Re-enable codecov

NotWoods:
Copies the script locally

### DIFF
--- a/automation/taskcluster/.codecov
+++ b/automation/taskcluster/.codecov
@@ -1,0 +1,1750 @@
+#!/usr/bin/env bash
+
+# Apache License Version 2.0, January 2004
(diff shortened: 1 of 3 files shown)


liuche:
Talk to @escapewindow about this (Aki from release engineering)

escapewindow:
I think this is a good stopgap, or ongoing solution if a) we have something that regularly checks the script, and b) we can't get codecov to help debug and address the hanging issue.

jonalmeida:
Thanks for doing this work @NotWoods ! It saved us from [the massive breach](https://www.theverge.com/2021/4/18/22390379/federal-investigators-breach-software-codecov-solarwinds) that took place. 🙃

PR #12329, opened 2020-07-06
Switch to groovy in buildSrc

NotWoods:
In [AC](https://github.com/mozilla-mobile/android-components/pull/7630) this change saved us 13 seconds of build time.

### DIFF
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -3,7 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 plugins {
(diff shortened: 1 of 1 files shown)


liuche:
cc @mcomella from talking to him about buildSrc and kotlin

boek:
@mcomella Is this ok to land?

mcomella:
I don't think there are enough details on the methodology in the a-c bug to draw conclusions that this change has a positive impact. However, the change has only a minor impact on developers so I'm not too concerned about it and I don't have time to investigate further so it seems fine to land to me.

Issue #12353, opened 2020-07-07
FNX2-15185 ⁃ [Bug] URL copied from reader mode is useless: hard-coded as `moz-extension://...`


mcomella:
## Steps to reproduce
* Load a page
* Enter reader mode
* After page load, long press on URL
* Select "copy"
* (go see what's in your clipboard by clicking URL bar and seeing clipboard suggestion or pasting it somewhere)

### Expected behavior
URL is original page URL

### Actual behavior
URL is reader mode URL: `moz-extension://4cb11be8-83cb-41af-81c3-58e30a4717ab/readerview.html?url=https%3A%2F%2Fwww.nytimes.com%2F2020%2F07%2F07%2Fworld%2Fcoronavirus-updates.html&id=137438953892`

This behavior does not appear to apply to sharing the URL.

### Device information
* Android device: P2
* Fenix version: Nightly 200706

csadilek:
See https://github.com/mozilla-mobile/android-components/issues/2879#issuecomment-649101920 for a way to fix this with protocol headers.

abodea:
Verified as fixed on the latest **Nightly 7/9** with **Google Pixel 4 XL (10).**

PR #12397, opened 2020-07-08
For #12353: URL copied to clipboard should not be extension/reader URL

csadilek:
I don't like that bug :), plus this gets rid of a `SessionManager` call.


app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarView.kt
```diff
@@ -300,5 +296,15 @@ class BrowserToolbarView(
 
     companion object {
         private const val TOOLBAR_ELEVATION = 16
+
+        @VisibleForTesting
+        internal fun getUrlFromClipboard(store: BrowserStore, customTabSession: Session? = null): String? {
```
ekager, line 301: nit - should this be `getUrlForClipboard`?

csadilek: Oh, yes, definitely. Fixed.

Issue #12311, opened 2020-07-06
FNX2-15165 ⁃ [Bug] Crash when custom tab is opened in browser


vesta0:
## Steps to reproduce
-Set the new nightly (formerly Preview) as default browser
-Open a link from Gmail in a custom tab
-Open the toolbar menu and tap on "open in Firefox nightly"

### Expected behavior
Custom tab opens in Nightly

### Actual behavior
Browser crashes (although it does open in the browser in the background before it crashes - you'll see it in the tab list if opening the browser after the crash)

### Device information
* Android device:  pixel 4
* Fenix version: Nightly 200706 06:01 (Build #21880609)

s-ankur:
Dupe #12277

ekager:
26b0f567-d616-4c5c-a768-57568c9215b9
java.lang.IllegalStateException: SET SESSION: Current activity: HomeActivity Other activity: ExternalAppBrowserActivity
----
java.lang.IllegalStateException: SET SESSION: Current activity: HomeActivity Other activity: ExternalAppBrowserActivity
	at mozilla.components.browser.engine.gecko.GeckoEngineView.render(GeckoEngineView.kt:17)
	at mozilla.components.feature.session.engine.EngineViewPresenter.access$onEngineSession(EngineViewPresenter.kt:2)
	at mozilla.components.feature.session.engine.EngineViewPresenter$start$1$invokeSuspend$$inlined$collect$1.emit(Collect.kt:2)
	at mozilla.components.support.ktx.kotlinx.coroutines.flow.FlowKt$ifChanged$$inlined$filter$1$2.emit(Collect.kt:7)
	at mozilla.components.feature.session.engine.EngineViewPresenter$start$1$invokeSuspend$$inlined$map$2$2.emit(Collect.kt:3)
	at mozilla.components.feature.session.engine.EngineViewPresenter$start$1$invokeSuspend$$inlined$map$1$2.emit(Collect.kt:3)
	at kotlinx.coroutines.flow.FlowKt.emitAllImpl$FlowKt__ChannelsKt(Unknown Source:11)
	at kotlinx.coroutines.flow.FlowKt__ChannelsKt$emitAllImpl$1.invokeSuspend(Unknown Source:11)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:3)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:21)
[...]

PR #12428, opened 2020-07-09
Issue #12311: BrowserToolbarController: Stop SessionFeature and release session from EngineView.

pocmo:
Now with BrowserStore we need to call `release()` on the feature, so that the feature stops observing the store. Otherwise it is possible that the feature will reattach the session and then we render it in the custom tab again before we switch to the browser.


st3fan:
I removed the labels because this is not a release blocker and does not need an uplift to the 79 branch.

Issue #11333, opened 2020-05-19
FNX2-13431 ⁃ Fenix always creates a GeckoRuntime with no ContentBlocking.Settings


droeh:
This is in relation to the GV bug https://bugzilla.mozilla.org/show_bug.cgi?id=1621340 and (closed) AC issue https://github.com/mozilla-mobile/android-components/issues/6193

It appears that Fenix always creates a `GeckoRuntime` with no `ContentBlocking.Settings` supplied by the `GeckoRuntimeSettings`: https://github.com/mozilla-mobile/fenix/blob/master/app/src/geckoBeta/java/org/mozilla/fenix/engine/GeckoProvider.kt#L45 -- this means that we effectively set the content blocking settings twice; once to the GV default values, and once to whatever a given Fenix install has chosen as its values. This is a pretty expensive startup perf issue according to the profiling Markus did on the (bugzilla) bug.

I don't think there's anything we can do in GV to fix this; my recommendation would be to update `GeckoProvider.getOrCreateRuntime` to optionally take `ContentBlocking.Settings` (or even `GeckoRuntimeSettings`) and pass them to `GeckoRuntime.create()`.

mcomella:
@Amejia481 What's the status of this bug? I see it was removed from the a-c board but is still assigned.

Amejia481:
I was about to start working on it, but I got sidetracked to https://github.com/mozilla-mobile/fenix/issues/11387 issues as it has a higher priority. As soon we get https://github.com/mozilla-mobile/fenix/issues/11387 done, I will return to this issue :)

sraturi:
According to the [profile](https://perfht.ml/3aaEIsB) by @mstange, Seems like we can avoid unnecessary calls to `readPref` inside `SafeBrowsing`. Which would save us approximately ~130 ms?

MarcLeclair:
@mcomella  (or me) need to review the patch

mcomella:
On my GS5, I took a profile:
- before the change b26e39906db511850e71c9ffe62cbf693d27aca5: https://share.firefox.dev/2PqDikQ
- after the change c143f95819aa34e65017ded6bc519d87c2c8a0e0: https://share.firefox.dev/3fyeSjL

[...]

PR #12432, opened 2020-07-09
Issue #11333: Set ContentBlocking settings directly on GeckoRuntime

jonalmeida:
Requires https://github.com/mozilla-mobile/android-components/pull/7666 and https://bugzilla.mozilla.org/show_bug.cgi?id=1651454.

We set the ContentBlockingSettings directly on the GeckoRuntime now to
improve the startup of the engine.

This change has requirements from Android Components and GeckoView, so
we would only see the full perf benefits in Nightly as the changes ride
the train, although we might start to see some of them as we're updating
the GeckoProvider for the `geckoBeta` variant as well.

Co-authored-by: Arturo Mejia <arturomejiamarmol@gmail.com>

### DIFF
--- a/app/src/geckoBeta/java/org/mozilla/fenix/engine/GeckoProvider.kt
+++ b/app/src/geckoBeta/java/org/mozilla/fenix/engine/GeckoProvider.kt
@@ -5,7 +5,9 @@
 import android.content.Context
 import android.os.Bundle
 import mozilla.components.browser.engine.gecko.autofill.GeckoLoginDelegateWrapper
+import mozilla.components.browser.engine.gecko.ext.toContentBlockingSetting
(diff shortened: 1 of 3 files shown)


Amejia481:
Awesome work 👍

Amejia481:
@mstange would you mind taking a look to see if the issue still persist?

jonalmeida:
> @mstange would you mind taking a look to see if the issue still persist?

@Amejia481 should we land the fix first so it can be tested in the next nightly?

Amejia481:
No problem, lets do that!

mstange:
> @mstange would you mind taking a look to see if the issue still persist?

Sorry for the delay - I just took another profile with a recent Nightly on my phone, and the issue is indeed fixed! Nice work!

Issue #12345, opened 2020-07-07
FNX2-15180 ⁃ [Crash] org.mozilla.fenix.perf.StartupReportFullyDrawn$attachReportFullyDrawn


st3fan:
> https://crash-stats.mozilla.org/signature/?product=Fenix&signature=java.lang.SecurityException%3A%20at%20android.os.Parcel.readException%28Parcel.java%29&version=79.0.0-beta.5&date=%3C2020-07-07T14%3A58%3A43%2B00%3A00&date=%3E%3D2020-06-30T14%3A58%3A43%2B00%3A00

Mike, this is the number two top (startup) crash in Fenix 79b4 and 79b5 currently. Can you please take a look?

```
java.lang.SecurityException
	at android.os.Parcel.readException(Parcel.java:1967)
	at android.os.Parcel.readException(Parcel.java:1913)
	at android.app.IActivityManager$Stub$Proxy.reportActivityFullyDrawn(IActivityManager.java:8579)
	at android.app.Activity.reportFullyDrawn(Activity.java:2015)
	at org.mozilla.fenix.perf.StartupReportFullyDrawn$attachReportFullyDrawn$$inlined$doOnPreDraw$1.run(View.kt:1)
	at androidx.core.view.OneShotPreDrawListener.onPreDraw(OneShotPreDrawListener.java:2)
	at android.view.ViewTreeObserver.dispatchOnPreDraw(ViewTreeObserver.java:1045)
	at android.view.ViewRootImpl.performTraversals(ViewRootImpl.java:2813)
	at android.view.ViewRootImpl.doTraversal(ViewRootImpl.java:1785)
	at android.view.ViewRootImpl$TraversalRunnable.run(ViewRootImpl.java:7832)
	at android.view.Choreographer$CallbackRecord.run(Choreographer.java:911)
	at android.view.Choreographer.doCallbacks(Choreographer.java:723)
	at android.view.Choreographer.doFrame(Choreographer.java:658)
	at android.view.Choreographer$FrameDisplayEventReceiver.run(Choreographer.java:897)
	at android.os.Handler.handleCallback(Handler.java:789)
	at android.os.Handler.dispatchMessage(Handler.java:98)
	at android.os.Looper.loop(Looper.java:164)
	at android.app.ActivityThread.main(ActivityThread.java:6944)
	at java.lang.reflect.Method.invoke(Native Method)
[...]

PR #12441, opened 2020-07-09
For #12345: avoid throwing regular exceptions for reportFullyDrawn. 

mcomella:
This should minorly improve performance – maybe it's not worth it.

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/ext/Activity.kt
+++ b/app/src/main/java/org/mozilla/fenix/ext/Activity.kt
@@ -5,10 +5,14 @@
 package org.mozilla.fenix.ext
 
 import android.app.Activity
+import android.os.Build
 import android.view.View
 import android.view.WindowManager
 import mozilla.components.support.base.log.Log
+import mozilla.components.support.base.log.Log.Priority
 import org.mozilla.fenix.perf.Performance
+
+private const val API_LEVEL_ANDROID_9 = 28
 
 /**
  * Attempts to call immersive mode using the View to hide the status bar and navigation buttons.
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/fenix/ext/Activity.kt
```diff
@@ -29,12 +33,35 @@ fun Activity.enterToImmersiveMode() {
...
 fun Activity.reportFullyDrawnSafe() {
+    // An exception is thrown when calling reportFullyDrawn on some Samsung devices. We could catch
+    // the exception but catching exceptions is slow so we disable the call on all Samsung devices
+    // except the ones we're using for performance testing.
+    //
+    // We were unable to identify the root cause of the Exception but suspect it's related to Samsung
+    // security features. See
+    // https://github.com/mozilla-mobile/fenix/issues/12345#issuecomment-655058864 for details.
+    //
+    // We briefly measured how slow the Exception code path is here: on a Galaxy S5, it's ~1ms, ~6x
```
BranescuMihai, line 44: @mcomella this is the first time I've seen code trying to reduce start-up time with 1ms, I'm impressed!

mcomella: 😓

mcomella:
I opened another Sentry report with a similar stack trace and found it was the same cause. Apparently, there are several of them and they're different stack traces because they're running different versions of Android. Therefore, my patch won't work as written :\ I'll rewrite it.

mcomella:
@MarcLeclair `reportFullyDrawn` will throw exceptions on Samsung devices sometimes. My original code just caught the exception but, since exceptions are slow, this means a perf hit on these devices (on my GS5, it was ~1ms, ~6x slower than a check to avoid calling `reportFullyDrawn` on these devices). However, my check doesn't work as written so we'd need a new check. I was thinking of adding a blocklist of all Samsung devices except for a small set we allowlist (i.e. the ones we use in perf testing). However, this also won't work because the S7 shows up in crash reports.

To solve this problem, I can only think of using a separate app launch to set a preference we'd read on startup to run `reportFullyDrawn` but then that also means performance hits and much more complexity. At this point, I don't think it's worth implementing a fix for the perf hit of throwing exceptions.

What do you think? Perhaps there's another solution that comes to mind for you?

mcomella:
> What do you think? Perhaps there's another solution that comes to mind for you?

iirc, I spoke to Marc on Zoom and we decided that we didn't want to add the complexity.

PR #12463, opened 2020-07-10
Show coverage using jacoco artifacts

NotWoods:
Workaround for Codecov problems


NotWoods:
Codecov is back so this is unneeded

PR #12499, opened 2020-07-13
Replaces verifyPageContent in UI tests

sv-ohorvath:
For: https://github.com/mozilla-mobile/mobile-test-eng/issues/51
Replaced verifyPageContent with verifyUrl where it was fit.
Replaced with waitForIdle where it was only used to wait for the page load.
Changed the assertion method in verifyPageContent for verifying the page refresh. This assertion is also used now in MediaNotificationTest (the prior method returned a false positive) - I'll keep an eye on these for a while.
note:
Tests passed on running ~50 times on Firebase: https://github.com/mozilla-mobile/fenix/runs/857803644, https://github.com/mozilla-mobile/fenix/runs/858196969

### DIFF
--- a/app/src/androidTest/java/org/mozilla/fenix/ui/BookmarksTest.kt
+++ b/app/src/androidTest/java/org/mozilla/fenix/ui/BookmarksTest.kt
@@ -234,7 +234,7 @@ class BookmarksTest {
             IdlingRegistry.getInstance().register(bookmarksListIdlingResource!!)
         }.openThreeDotMenu(defaultWebPage.url) {
         }.clickOpenInNewTab {
-            // verifyPageContent(defaultWebPage.content)
+            verifyUrl(defaultWebPage.url.toString())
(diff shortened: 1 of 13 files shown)


AaronMT:
Im having trouble understanding how it's any different as both were waiting for an object. Can you explain a little bit? Also, I would run this a bunch of times to check.

sv-ohorvath:
> Im having trouble understanding how it's any different as both were waiting for an object. Can you explain a little bit? Also, I would run this a bunch of times to check.



> Im having trouble understanding how it's any different as both were waiting for an object. Can you explain a little bit? Also, I would run this a bunch of times to check.

So after talking a bit more with mcarare, the difference between waitNotNull and waitForExists would be that the first is trying to find a view in the hierarchy at regular intervals before reaching the timeout, and the later is searching for an accessibility node info and waits for the view to become visible. I chose waitForExists because it's a method of UiObject and not an extension.

PR #12507, opened 2020-07-13
Ensure logins deletion

Mugurell:
Currently when the user goes back to the list of saved logins from the detail view she will briefly see the previous list of logins, possible containing even the one just edited/deleted.
And in some situations, until we get the updated list of saved logins, users would be able to tap the same just deleted login to open in a detail view, this leading to a crash.

To resolve the above a solution would have been to use a nested nav-graph for all logins fragment and a ViewModel scoped to it. But by working with the same set of logins (which could have been updated from another device if synced) for possibly a longer time we would expose ourselves to strange concurrent modification issues.

The solution I went with is simply to clear the list of saved logins when the user selects a login to open in details view so that upon returning and until receiving an updated list from `components.core.passwordsStorage.list()` he would not see a potential obsolete list, but would see (possibly, very briefly) the "loading" ux.

### Pull Request checklist

- [x] **Screenshots**: No screenshots. Tried to make a video but there is no apparent change, everything happening too fast on my device.

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/settings/logins/LoginsFragmentStore.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/logins/LoginsFragmentStore.kt
@@ -54,6 +54,12 @@ sealed class LoginsAction : Action {
     data class UpdateLoginsList(val list: List<SavedLogin>) : LoginsAction()
     data class UpdateCurrentLogin(val item: SavedLogin) : LoginsAction()
     data class SortLogins(val sortingStrategy: SortingStrategy) : LoginsAction()
+
+    /**
+     * Cleanup the list of logins currently shown and enter in a "loading" state
+     * until [UpdateLoginsList] is emitted with a new list of logins to be displayed.
+     */
+    data class LoginSelected(val item: SavedLogin) : LoginsAction()
 }
 
 /**
@@ -108,6 +114,13 @@ private fun savedLoginsStateReducer(
                 state.searchedForText,
                 action.sortingStrategy,
                 state
+            )
+        }
+        is LoginsAction.LoginSelected -> {
+            state.copy(
+                    isLoading = true,
+                    loginList = emptyList(),
+                    filteredItems = emptyList()
             )
         }
     }
--- a/app/src/main/java/org/mozilla/fenix/settings/logins/LoginsFragmentStore.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/logins/LoginsFragmentStore.kt
@@ -54,11 +54,6 @@ sealed class LoginsAction : Action {
     data class UpdateLoginsList(val list: List<SavedLogin>) : LoginsAction()
     data class UpdateCurrentLogin(val item: SavedLogin) : LoginsAction()
     data class SortLogins(val sortingStrategy: SortingStrategy) : LoginsAction()
-
-    /**
-     * Cleanup the list of logins currently shown and enter in a "loading" state
-     * until [UpdateLoginsList] is emitted with a new list of logins to be displayed.
-     */
     data class LoginSelected(val item: SavedLogin) : LoginsAction()
 }
 
--- a/app/src/main/java/org/mozilla/fenix/settings/logins/SavedLoginsFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/logins/SavedLoginsFragment.kt
@@ -37,12 +37,10 @@ import org.mozilla.fenix.BrowserDirection
 import org.mozilla.fenix.HomeActivity
 import org.mozilla.fenix.R
 import org.mozilla.fenix.components.StoreProvider
-import org.mozilla.fenix.components.metrics.Event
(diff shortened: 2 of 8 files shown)


app/src/main/java/org/mozilla/fenix/settings/logins/SavedLoginsFragment.kt
```diff
@@ -88,9 +86,14 @@ class SavedLoginsFragment : Fragment() {
...
         val savedLoginsController: SavedLoginsController =
-            SavedLoginsController(savedLoginsStore, requireContext().settings())
-        savedLoginsInteractor =
-            SavedLoginsInteractor(savedLoginsController, ::itemClicked, ::openLearnMore)
+            SavedLoginsController(
+                    store = savedLoginsStore,
+                    navController = findNavController(),
+                    browserNavigator = ::openToBrowserAndLoad,
+                    settings = requireContext().settings(),
+                    metrics = requireContext().components.analytics.metrics
```
eliserichards, line 94: I'm not sure that I like sending the metrics component into the controller since it is not necessary for the functionality 🤔

app/src/main/java/org/mozilla/fenix/settings/logins/SavedLoginsView.kt
```diff
@@ -60,6 +65,7 @@ class SavedLoginsView(
     }
 
     fun update(state: LoginsListState) {
+        // todo MVI views should not have logic. Needs refactoring.
```
eliserichards, line 68: Let's file an eng health followup for this! 😄

app/src/main/java/org/mozilla/fenix/settings/logins/SavedLoginsView.kt
```diff
@@ -73,29 +79,61 @@ class SavedLoginsView(
...
 }
 
 /**
  * Controller for the saved logins screen
+ *
+ * @param store Store used to hold in-memory collection state.
+ * @param navController NavController manages app navigation within a NavHost.
+ * @param browserNavigator Controller allowing browser navigation to any Uri.
+ * @param settings SharedPreferences wrapper for easier usage.
+ * @param metrics Controller that handles telemetry events.
```
eliserichards, line 106: Same thing here, I do not want the metrics component passed since it doesn't have anything to do with the logic of the controller.

app/src/main/java/org/mozilla/fenix/settings/logins/LoginsFragmentStore.kt
```diff
@@ -54,6 +54,12 @@ sealed class LoginsAction : Action {
     data class UpdateLoginsList(val list: List<SavedLogin>) : LoginsAction()
     data class UpdateCurrentLogin(val item: SavedLogin) : LoginsAction()
     data class SortLogins(val sortingStrategy: SortingStrategy) : LoginsAction()
+
+    /**
+     * Cleanup the list of logins currently shown and enter in a "loading" state
+     * until [UpdateLoginsList] is emitted with a new list of logins to be displayed.
+     */
```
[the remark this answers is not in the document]
Mugurell: Indeed.. the comment talks about what will happen down the road and not about when it should be used.
And when it should be used is pretty self-explanatory from the name.
I'll remove the comment.

app/src/main/java/org/mozilla/fenix/settings/logins/SavedLoginsFragment.kt
Mugurell: The way I see it the Controller handles the business logic.
Views render State and catch user actions. They should have no logic.
Interactors just delegate user actions to Controllers/other Interactors. They should have no logic.
Fragments are the puppeteers deciding what MVI View goes where and constructs them. They should have no other logic. (But because of architectural limitations in Android Controllers do get to delegate Fragments for different actions, like asking for runtime permissions).
Otherwise, all logic stays in Controllers (ideally decoupled from anything Android, like the old Presenters, easy to test, easy to port/refactor), they are the ones deciding what happens in the app.
And in this case the MetricController is needed because the Controller decides to log some telemetry and it is the only component that can do this.

Happy to discuss if you see it differently.

app/src/main/java/org/mozilla/fenix/settings/logins/SavedLoginsView.kt
Mugurell: Sure!
Filed https://github.com/mozilla-mobile/fenix/issues/12588.

app/src/main/java/org/mozilla/fenix/settings/logins/SavedLoginsView.kt
Mugurell: As above, the way I see it, the controller is the only one that knows that a certain event needs to be logged or not. It handles the business logic.
So everything needed to do it's job needs to preferably be injected in the constructor.

eliserichards:
Whoops ^ didn't mean to do that

PR #12517, opened 2020-07-13
Change enabled lint rules

NotWoods:
We have lots of translation lints that we don't want, let's turn them off.

Experimenting with enabling a couple other lints.


eliserichards:
Can we separate out the cleanup from the lint rules? I'm happy to approve both parts of this 👍

PR #12520, opened 2020-07-13
Add helper for underlining text

NotWoods:
Adding an underline to a TextView seems to be a fairly common pattern, so lets put it in a helper method. Also adds more tests to some of the changed components.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/ext/TextView.kt
+++ b/app/src/main/java/org/mozilla/fenix/ext/TextView.kt
@@ -1,0 +1,19 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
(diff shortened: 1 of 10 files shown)


app/src/main/java/org/mozilla/fenix/ext/TextView.kt
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * Adds an underline effect to the text displayed in the TextView.
+ */
+fun TextView.addUnderline() {
+    val currentText = text
+    text = currentText.toSpannable().apply {
+        setSpan(UnderlineSpan(), 0, currentText.length, 0)
+    }
+}
```
jhugman, line 19: I think I'd prefer an extension property here, or at least make the method take an optional `start, end` or `start, len` arguments.

NotWoods: What sort of property are you thinking of? I'd prefer not to add the arguments as all the uses of UnderlineSpan just underline the entire string.

Issue #12400, opened 2020-07-08
FNX2-15220 ⁃ [Bug] swipe to delete bug in collection


sheikh-azharuddin:
## Steps to reproduce
Add one tab to collection
Swipe right to delete. It will show a confirmation prompt. 
Click cancel 
It is showing a persistent delete action with blank row. 

### Expected behavior
Should show the tab under collections

### Actual behavior
Is showing persistent delete icon with blank row

### Device information
* Android device: ? A10
* Fenix version: ? Latest nightly 

![Screenshot](https://user-images.githubusercontent.com/58127841/86972018-bc198900-c161-11ea-9db3-8038eca4d3c5.jpg)UNITO-UNDERSCORE!20200709-025411!

PR #12538, opened 2020-07-14
For issue #12400 - Refresh swiped collection tab view

BMEduard:
Item is now refreshed by calling notifyDataSetChanged on the adapter when the last tab from the collection has been swiped away and the user cancels the deletion by pressing the cancel button from the dialog.
Also added a "wasSwiped" flag to onCollectionRemoveTab in order to check if the tab was deleted from a swipe action and not by pressing the "X" button.

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
@@ -211,7 +211,8 @@ class HomeFragment : Fragment() {
                 hideOnboarding = ::hideOnboardingAndOpenSearch,
                 registerCollectionStorageObserver = ::registerCollectionStorageObserver,
                 showDeleteCollectionPrompt = ::showDeleteCollectionPrompt,
-                showTabTray = ::openTabTray
+                showTabTray = ::openTabTray,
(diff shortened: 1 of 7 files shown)


ekager:
Whew okay bugs galore in collections homescreen 😬 Pulled this PR down and seems to work well in the case of 1 item in a collection being swiped away (which was the original bug, so great!!)

2 other bugs I found: 
1. If a collection has 2 items and you swipe them away one by one, it won't prompt you or delete the collection on the last item's removal (so then you end up with an empty collection). 
2. Swiping away on the bottom tab of a collection doesn't update the remaining item to be the next bottom item visually. 

These two bugs are probably related. @BMEduard do you want to look into these as well since it's related, or would you prefer I break these into 2 new issues?

BMEduard:
> 
> These two bugs are probably related. @BMEduard do you want to look into these as well since it's related, or would you prefer I break these into 2 new issues?

@ekager I think it would be for the best to break those bugs into new issues.

ekager:
Sounds good opened https://github.com/mozilla-mobile/fenix/issues/13177

PR #12568, opened 2020-07-14
Update coroutines dependency

NotWoods:
Breaking this out from #10806 to see if it causes Glean issues


NotWoods:
Blocked by BZ-1653107

boek:
Looks like this is on the glean backlog. We can re-open this once it lands.

boek:
Created a new issue to track: #13097

PR #12549, opened 2020-07-14
Update Android Components version

MickeyMoz:

PR #12572, opened 2020-07-14
Test A-C thumbnail changes.

person808:
Testing breaking thumbnail change in https://github.com/mozilla-mobile/android-components/pull/7534

cc @jonalmeida

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/tabtray/TabTrayViewHolder.kt
+++ b/app/src/main/java/org/mozilla/fenix/tabtray/TabTrayViewHolder.kt
@@ -11,6 +11,7 @@ import android.widget.TextView
 import androidx.annotation.VisibleForTesting
 import androidx.appcompat.widget.AppCompatImageButton
 import androidx.core.content.ContextCompat
+import androidx.core.view.doOnNextLayout
 import mozilla.components.browser.state.state.MediaState
 import mozilla.components.browser.tabstray.TabViewHolder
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/fenix/tabtray/TabTrayViewHolder.kt
```diff
@@ -72,7 +73,10 @@ class TabTrayViewHolder(
         if (tab.thumbnail != null) {
             thumbnailView.setImageBitmap(tab.thumbnail)
         } else {
-            imageLoader.loadIntoView(thumbnailView, ImageRequest(tab.id))
+            // Make sure we have the view's dimensions so we can load the image at the correct size
+            thumbnailView.doOnNextLayout {
+                imageLoader.loadIntoView(thumbnailView, tab.id)
+            }
```
person808, line 79: Alternatively we could hardcode the image size to 100dp instead of using the view dimensions through the extension function. In my testing loading with the extension function requests a slightly smaller image (253px vs 275px on Pixel 3 emulator)

person808:
I think the problems I noticed in my A-C pr were related to earlier revisions of it loading the full size thumbnail for each tab tray item. I was probably triggering some out of memory errors, not filling up the disk cache any faster. So I don't think they'll crop up now 🤞

jonalmeida:
Closing in place of https://github.com/mozilla-mobile/fenix/pull/12549.

Issue #12565, opened 2020-07-14
Clean up use of Context and Activity within Controllers


eliserichards:
Clean up uses of `Context` and `Activity` in our code, specifically within `Controllers`, based on conversation with @Mugurell in https://github.com/mozilla-mobile/fenix/pull/11208:

> I'm normally against passing it around and then expect all receivers to know how to build/extract their end dependencies from it. (Demeter's law).
To keep the code clean, have a better separation of responsibilities, loose coupling and ease testing I would propose to have end dependencies sent as parameters to the controller.
Here I see that anyway the context, cast to a specific activity!? is only used for obtaining a reference to components.core.passwordsStorage. That can easily be injected in the constructor without even worrying about it become bloated.

gabrielluong:
@mcarare curious if we should be closing this as in we are completely done with all the clean up of these usages?

PR #12562, opened 2020-07-14
Use MainCoroutineRule in tests

NotWoods:

PR #12579, opened 2020-07-15
For #12565: Add Locale settings tests & use activity

NotWoods:
- Changes `DefaultLocaleSettingsController` to take activity instead of context, since it casts context anyways.
- Moves the 3 viewholder classes out of LocaleAdapter to its own file.
- Cleans up tests to make them more readable


app/src/main/java/org/mozilla/fenix/settings/advanced/DefaultLocaleSettingsController.kt
```diff
@@ -16,32 +16,44 @@ interface LocaleSettingsController {
...
 
     override fun handleSearchQueryTyped(query: String) {
         localeSettingsStore.dispatch(LocaleSettingsAction.Search(query))
     }
+
+    /**
+     * Update the locale for the configuration of the app context's resources
+     */
+    @Suppress("Deprecation")
+    fun LocaleManager.updateBaseConfiguration(context: Context, locale: Locale) {
```
eliserichards, line 52: is there an alternative to this deprecated method?

NotWoods: I'd prefer to leave that as follow up work. I filed #12562

PR #12581, opened 2020-07-15
Add tests for sync, wifi, and push

NotWoods:
In addition to lots of tests, there has been some cleanup:
- SyncedTabsIntegration takes in syncedTabsStorage, instead of using context to get syncedTabsStorage from components.
- WifiConnectionMonitor now uses the Observable code from AC instead of re-implementing the observable pattern.
- SitePermissionsWifiIntegration has been updated to call AC observation methods

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/Components.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/Components.kt
@@ -4,9 +4,9 @@
 
 package org.mozilla.fenix.components
 
-import android.app.Application
 import android.content.Context
 import android.content.Intent
+import androidx.core.content.getSystemService
 import androidx.core.net.toUri
 import mozilla.components.feature.addons.AddonManager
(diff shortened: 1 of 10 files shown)


app/src/main/java/org/mozilla/fenix/components/BackgroundServices.kt
```diff
@@ -159,7 +159,10 @@ class BackgroundServices(
             notificationManager.showReceivedTabs(context, device, tabs)
         }
 
-        SyncedTabsIntegration(context, accountManager).launch()
+        SyncedTabsIntegration(
+            syncedTabsStorage = syncedTabsStorage,
```
jonalmeida, line 163: Please don't land this without testing the feature carefully. If I recall correctly this causes a cyclic dependency, which is why we accessed the storage via the context instead of passing it in through he constructor.

Issue #11851, opened 2020-06-23
FNX2-13646 ⁃ [Bug] Close last tab doesn't close tab tray


B0pol:
## Steps to reproduce
1. Open a page
1. Close it, you now have 0 tab
    1. Expected behavior
The tab drawer is closed so that I can access collections / top sites / settings / …
    1. Actual behavior
Tab tray is still shown
    1. Device information

* Android device: 9
* Fenix version: 5.2

hakkikaancaliskan:
@gabrielluong If ux team decide how it should be, i can take this issue.

gabrielluong:
> @gabrielluong If ux team decide how it should be, i can take this issue. Needs ux feedback.

Requested. @topotropic is still away for another week I think. Thanks!

topotropic:
Yes, please close the drawer when people don't have any open tabs and show new tab/home with top sites, collections.... Thanks!

hakkikaancaliskan:
@gabrielluong i don't have time to work on this currently, maybe later. Other guys could work on this.

ekager:
@topotropic when you close the last tab - we should still show the undo option though? This piece could be tricky

person808:
The way I handled that for the tab counter menu was to provide a the sessionId as an argument to the nav action that returns to `HomeFragment` and have `HomeFragment` handle the tab closing and snackbar.

ekager:
Nice, so I'm assuming that's what we'd want to do here as well.

topotropic:
> The way I handled that for the tab counter menu was to provide a the sessionId as an argument to the nav action that returns to `HomeFragment` and have `HomeFragment` handle the tab closing and snackbar.

does this mean we close the tab tray and show the undo snackbar on home when people close their last tab?

person808:
@topotropic Yup. If we take the same approach as the tab counter menu then thats what it'll look like.

sblatz:
What should happen in the case where a user has:

1 normal tab
2 private tabs

Closes last normal tab.... should it dismiss (same goes for the inverse)? @topotropic

sblatz:
The behavior that happens by "default" with my change is:

### user state
1 normal tab
2 private tabs

Then...
Close my normal tab

Then...
Now Fenix shows private home screen (with tray dismissed). 

Which actually seems pretty reasonable to me?

sflorean:
[...]

Issue #12383, opened 2020-07-08
[Bug]Deletion confirmation snackbar missing when closing all PB Open Tabs


AndiAJ:
## Prerequisites
Have a couple of Normal and Private tabs open.

## Steps to reproduce
1. Open one of the Private Tabs
1. Tap the tab counter
1. Tap the ⋮ "More options button"
1. Select "Close all tabs"

### Expected behavior
The Deletion confirmation snackbar is properly displayed, I can undo the deletion

### Actual behavior
The Deletion confirmation snackbar is missing, can't undo the deletion

### Device information
* Android device: all
* Fenix version: Nightly Build 200708 from 7/8

### Notes
►Video
![20200708-130753](https://user-images.githubusercontent.com/51314259/86907138-42f84280-c11d-11ea-80f1-97762be4050e.gif)

ekager:
Similar to #11851

Diana-Rus:
Reproducible on Firefox Preview Beta 81.1.0 with Google Pixel 3XL (Android 9)

AndiAJ:
Hi, **verified as fixed** on the latest **`Nightly Build 200928`** using the following devices:
• Google Pixel 3a (Android 11)
• Huawei Mate 20 Lite (Android 10)
• OnePlus A3 (Android 6.0.1)

Filled #15490 for an inconsistency I've encountered.

PR #12615, opened 2020-07-15
For #11851: Close tab tray when last tab closed

sblatz:
### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/tabtray/TabTrayController.kt
+++ b/app/src/main/java/org/mozilla/fenix/tabtray/TabTrayController.kt
@@ -117,6 +117,7 @@ class DefaultTabTrayController(
         }
 
         showUndoSnackbar(snackbarMessage, snapshot)
+        dismissTabTray()
     }
 
     @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
--- a/app/src/main/java/org/mozilla/fenix/tabtray/TabTrayDialogFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/tabtray/TabTrayDialogFragment.kt
@@ -38,6 +38,7 @@ import org.mozilla.fenix.components.FenixSnackbar
 import org.mozilla.fenix.components.TabCollectionStorage
 import org.mozilla.fenix.components.metrics.Event
 import org.mozilla.fenix.ext.components
+import org.mozilla.fenix.ext.getRootView
 import org.mozilla.fenix.ext.requireComponents
 import org.mozilla.fenix.ext.settings
 import org.mozilla.fenix.utils.allowUndo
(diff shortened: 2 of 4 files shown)


app/src/main/java/org/mozilla/fenix/utils/Undo.kt
```diff
@@ -63,6 +66,31 @@ fun CoroutineScope.allowUndo(
...
+        val shouldUseBottomToolbar = view.context.settings().shouldUseBottomToolbar
+        val toolbarHeight = view.context.resources
+            .getDimensionPixelSize(R.dimen.browser_toolbar_height)
+
+        snackbar.view.setPadding(
+            0,
+            0,
+            0,
+            if (
+                paddedForBottomToolbar &&
```
ekager, line 78: Won't this break other places the Undo is used? from a quick search - bookmarks, search engine list? I suppose having unnecessary padding is better than covering the toolbar? But still weird.

app/src/main/java/org/mozilla/fenix/tabtray/TabTrayController.kt
```diff
@@ -117,6 +117,7 @@ class DefaultTabTrayController(
         }
 
         showUndoSnackbar(snackbarMessage, snapshot)
+        dismissTabTray()
```
ekager, line 120: Do we need this? Aren't we already calling `dismissTabTrayIfNecessary` in the `showUndoSnackbar` fun?

app/src/main/java/org/mozilla/fenix/utils/Undo.kt
ekager: Ah I see it's a param we're passing in not just the bottom toolbar setting. Okey that should be fine. We should also just generally retest the Undo in the places it's used to make sure they all still work 🆗

app/src/main/java/org/mozilla/fenix/tabtray/TabTrayDialogFragment.kt
```diff
@@ -185,19 +186,31 @@ class TabTrayDialogFragment : AppCompatDialogFragment() {
...
-                getString(R.string.snackbar_deleted_undo),
-                {
-                    sessionManager.add(snapshot.session, isSelected, engineSessionState = state)
-                    tabTrayView.scrollToTab(snapshot.session.id)
-                },
-                operation = { },
-                elevation = ELEVATION,
-                anchorView = snackbarAnchor
-            )
+        val isLastOpenTab = sessionManager.sessions.size == 1
```
ekager, line 189: So we only want to close the tabs tray if private and normal tabs are gone ?

ekager:
Can we also verify the remove all tabs works in the same way for both normal and private (ie the followup from https://github.com/mozilla-mobile/fenix/issues/12383)

app/src/main/java/org/mozilla/fenix/tabtray/TabTrayController.kt
sblatz: `showUndoSnackbarForTab` is the only one that does `dismissTabTrayIfNecessary`. The one for `showUndoSnackbar` does not. I should probably rename the `forTab` one to explicitly call out that it dismisses the tab tray.

app/src/main/java/org/mozilla/fenix/tabtray/TabTrayDialogFragment.kt
sblatz: Still waiting on a UX answer for this, but I'm probably going to change it to be based on the active session type (private or normal).

Issue #12551, opened 2020-07-14
FNX-13634 ⁃ Swipe to delete bookmarks


person808:
Broken out from #176 

### What is the user problem or growth opportunity you want to see solved?
Users should be able to swipe to delete bookmarks.

### How do you know that this problem exists today? Why is this important?
This is not possible currently. Deletion is done by tapping the three dot menu and selecting delete.

### Who will benefit from it?
Everyone who deletes items from bookmarks.

person808:
Going to break swipe to delete history out into a separate issue because there are some problems with implementing it that need more thought.

ShatteredPixel:
@person808 

There's a bug with this behaviour. See the GIFs. You may need to swipe a bit diagonally to reproduce it. Even if I try to swipe exactly horizontally in a straight line it reproduces most of the time and of course I can do the swipe to delete gestures action if I try harder but it's not intuitive and the swipe action seems to be triggering the 3-dot menu which I think it should not.

See the GIFs: 



Thank you. :)

person808:
Hm that definitely doesn't seem right. Could you open a new issue with what you posted and tag me? I'll add it to my backlog. Thanks!

sflorean:
Verified as fixed on Nightly 7/24 with Samsung Galaxy Note 10(Android 10) and Huawei MediaPad M2(Android 5.1.1). 

@ShatteredPixel I was able to reproduce your issue also, I will leave a comment on the bug.

PR #12639, opened 2020-07-16
For #12551 - Implement swipe-to-delete for bookmarks

person808:
### Pull Request checklist

### After merge

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkTouchHelper.kt
```diff
@@ -0,0 +1,111 @@
...
+        val itemView = viewHolder.itemView
+        val iconLeft: Int
+        val iconRight: Int
+        val margin =
+            SwipeToDeleteCallback.MARGIN.dpToPx(recyclerView.context.resources.displayMetrics)
+        val iconWidth = icon.intrinsicWidth
+        val iconHeight = icon.intrinsicHeight
+        val cellHeight = itemView.bottom - itemView.top
+        val iconTop = itemView.top + (cellHeight - iconHeight) / 2
+        val iconBottom = iconTop + iconHeight
```
eliserichards, line 70: I think we can separate some things out here to make it more readable.
* I like that you've made `val`s for these so that the `when` statement is more readable
* Some things that don't depend on the _immediate_ updated state of the draw function, such as `margin`, can be moved to private `val`s outside of the function
* Keep the more complicated `val`s like `iconTop`
* The simpler onces, like `itemView`, can just be done inline

app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkTouchHelper.kt
```diff
@@ -0,0 +1,111 @@
...
+        when {
+            dX > 0 -> { // Swiping to the right
+                iconLeft = itemView.left + margin
+                iconRight = itemView.left + margin + iconWidth
+                background.setBounds(
+                    itemView.left, itemView.top,
+                    (itemView.left + dX).toInt() + SwipeToDeleteCallback.BACKGROUND_CORNER_OFFSET,
+                    itemView.bottom
+                )
+                icon.setBounds(iconLeft, iconTop, iconRight, iconBottom)
```
eliserichards, line 81: Since these two calls are made in every when scenario, let's pull them out to a little helper method

eliserichards: "these two calls" = `background.setBounds()` and `icon.setBounds()`

app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkAdapter.kt
```diff
@@ -68,9 +68,8 @@ class BookmarkAdapter(val emptyView: View, val interactor: BookmarkViewInteracto
     }
 
     override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): BookmarkNodeViewHolder {
-        val view = LibrarySiteItemView(parent.context).apply {
-            layoutParams = ViewGroup.LayoutParams(MATCH_PARENT, WRAP_CONTENT)
-        }
+        val view = LayoutInflater.from(parent.context)
+            .inflate(R.layout.bookmark_list_item, parent, false) as LibrarySiteItemView
```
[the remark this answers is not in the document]
person808: Unfortunately we can't 😞. It has to be casted.

app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkTouchHelper.kt
person808: Tried to keep variables that do common calculations like `iconTop` and `iconBottom` here and move the rest inside the `when`. Can't remove `margin` because it needs a Context.

PR #12725, opened 2020-07-20
[79.0.0] Uplift translations.

pocmo:


boek:
Looks like we deleted strings :(

liuche:
Thanks all!

PR #12747, opened 2020-07-20
Create ToolbarPosition enum

NotWoods:
Using an enum makes code much easier to read because the top/bottom association is more obvious. This just adds a new getter to `Settings` without changing the backing field.

```kt
        // Before
        val isBottomToolbar = Settings.getInstance(context).shouldUseBottomToolbar

        layout.drop_down_triangle.isGone = isBottomToolbar
        layout.pop_up_triangle.isVisible = isBottomToolbar
```

```kt
        // After
        val toolbarPosition = context.settings().toolbarPosition

        layout.drop_down_triangle.isVisible = toolbarPosition == ToolbarPosition.TOP
        layout.pop_up_triangle.isVisible = toolbarPosition == ToolbarPosition.BOTTOM
```

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarView.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarView.kt
@@ -17,6 +17,7 @@ import androidx.annotation.VisibleForTesting
 import androidx.coordinatorlayout.widget.CoordinatorLayout
 import androidx.core.content.ContextCompat
 import androidx.core.view.isVisible
+import androidx.core.view.updateLayoutParams
 import androidx.lifecycle.LifecycleOwner
 import com.google.android.material.appbar.AppBarLayout
 import com.google.android.material.appbar.AppBarLayout.LayoutParams.SCROLL_FLAG_ENTER_ALWAYS
@@ -59,7 +60,7 @@ interface BrowserToolbarViewInteractor {
 @SuppressWarnings("LargeClass")
 class BrowserToolbarView(
     private val container: ViewGroup,
-    private val shouldUseBottomToolbar: Boolean,
+    private val toolbarPosition: ToolbarPosition,
     private val interactor: BrowserToolbarViewInteractor,
     private val customTabSession: Session?,
     private val lifecycleOwner: LifecycleOwner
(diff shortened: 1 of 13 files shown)


app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarView.kt
```diff
@@ -167,10 +168,9 @@ class BrowserToolbarView(
                     false
                 }
 
-                display.progressGravity = if (shouldUseBottomToolbar) {
-                    DisplayToolbar.Gravity.TOP
-                } else {
-                    DisplayToolbar.Gravity.BOTTOM
+                display.progressGravity = when (toolbarPosition) {
```
BranescuMihai, line 171: This one should be the other way around, opposite to the toolbar position

app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarView.kt
```diff
@@ -268,30 +271,27 @@ class BrowserToolbarView(
      * Note that the bottom toolbar has a feature flag for being dynamic, so it may not get flags set.
      */
     fun setScrollFlags(shouldDisableScroll: Boolean = false) {
-        if (view.context.settings().shouldUseBottomToolbar) {
-            if (view.layoutParams is CoordinatorLayout.LayoutParams) {
-                (view.layoutParams as CoordinatorLayout.LayoutParams).apply {
+        when (settings.toolbarPosition) {
```
BranescuMihai, line 274: Nice, this method is much more readable like this!

app/src/main/java/org/mozilla/fenix/trackingprotection/TrackingProtectionOverlay.kt
```diff
@@ -62,10 +63,10 @@ class TrackingProtectionOverlay(
 
         val layout = LayoutInflater.from(context)
             .inflate(R.layout.tracking_protection_onboarding_popup, null)
-        val isBottomToolbar = Settings.getInstance(context).shouldUseBottomToolbar
+        val toolbarPosition = context.settings().toolbarPosition
```
BranescuMihai, line 66: nit: `settings` instead of `context.settings()`?

Issue #12759, opened 2020-07-20
FNX-13876 ⁃ Separate release locale list from master locale list.


person808:
We use `l10n.toml` to define what locales to include in our builds. For release we want to use a separate list of locales than on the master branch. To allow us to have the same `l10n.toml` file on `master` and release branches we should create a separate list that is used to filter out locales on release builds. We can't modify [this](https://github.com/mozilla-mobile/fenix/blob/09d0688e6483d0e2ca33009c7a77855f29b9556d/l10n.toml#L4) list on `master` because Pontoon uses it to expose locales to translators.

person808:
Note for QA: we want to check that the apk from [here](https://firefox-ci-tc.services.mozilla.com/tasks/index/mobile.v2.fenix.fennec-production) should have only the locales in [`l10n-release.toml`](https://github.com/mozilla-mobile/fenix/blob/master/l10n-release.toml) in future builds. If we can verify the locale filter is running on one build in TC then it should be working for all build variants its needed for.

softvision-miralobontiu:
Verified all the languages from the locales in [l10n-release.toml](https://github.com/mozilla-mobile/fenix/blob/master/l10n-release.toml), on the build from [here](https://firefox-ci-tc.services.mozilla.com/tasks/index/mobile.v2.fenix.fennec-production.latest/arm64-v8a) with OnePlus 5T (Android 9).
Three languages are missing on the build from above, but **they are available on the 7/27 Nightly build**:
1. Hebrew "he".
2. Bahasa Indonesia "id".
3. Triqui "trs".

I'll remove the qa:needed label for now.
@person808 please let us know if there is anything else we can do.
Thanks!

person808:
Likely just a delay in pulling in the translations thanks!

Issue #3088, opened 2019-06-03
FNX3-16780 ⁃ Allow different locales for Nightly and Release


pocmo:
### Why?
Fenix gets translated on Pontoon and translated strings get imported into this repository "automatically" (kind of - from this team's point of view; @Pike opens PRs for that).

Most of them time it's okay to include all locales into the builds (local builds, Nightly builds), e.g. to allow developers/QA to test the app in other languages and for translators to verify their translations in the app. For other builds (release) we want to explicitly ship a list of "approved" locales.

### How?
For release builds we want to run `filter-release-translations.py` (before invoking Gradle) which will remove all strings of locales that are not in the release list. This will allow engineers and the L10N team to maintain a list of "release locales" in this repository.

### Acceptance Criteria
* [ ] Add Release locales list, like in [Focus](https://github.com/mozilla-mobile/focus-android/blob/master/tools/l10n/locales.py#L22)
* [ ] Port [filter-release-translations.py](https://github.com/mozilla-mobile/focus-android/blob/master/tools/l10n/filter-release-translations.py) from Focus to Fenix and add them to release build run (so only the Release locales get shipped)
* [ ] Verify that locales are filtered out correctly

Next step is actually adding new Nightly locales and making sure they run in #5496

pocmo:
Dupe of #1723.

Pike:
Wondering, should this also cover Beta? @Delphine ? I'd suggest so, but not my call.

Also, I found https://developer.android.com/studio/build/shrink-code.html#unused-alt-resources, which might do what we need w/out removing code for release.

I also found posts on stackoverflow indicating that the exposed locale list of the APK might be affected by what we have in dependent libraries like a-c, so we might not get what we expect by just reducing the fenix locales in-tree.

pocmo:
@Pike I think that is a great idea. :+1: Maintaining the locale list as `resConfigs` is way less painful - and I assume we can do that for release (and maybe beta) builds separately, so that local builds and Nightlies still have all strings.

Baron-Severin:
Looking into using [built in Gradle code shrinking](https://developer.android.com/studio/build/shrink-code.html#unused-alt-resources) as an alternative to our existing pruning script.

Delphine:
Yes, I agree that we should be covering Beta as well. Thank you!

Baron-Severin:
[...]

PR #12765, opened 2020-07-20
For #12759 - Keep a separate list of release locales.

person808:
This also updates the `release_locale` list to match the locales on the`releases/v79.0.0` branch. Tested the filtering script locally and it worked for me.

cc @liuche Do we want to uplift this to the release branch?


liuche:
@person808 this doesn't need to be uplifted - pontoon only localizes from master, so changing the list on the release branch doesn't have any effect, beyond removing those locale dirs that we don't want to include.

automation/taskcluster/l10n/locales.py
```diff
@@ -42,7 +42,7 @@ def get_release_locales():
     for line in file:
         if line == OPEN_LOCALES:
             locales_opened = True
-        elif line == CLOSE_LOCALES:
+        elif line == CLOSE_LOCALES and locales_opened == True:
```
liuche, line 45: Does this catch an edge case?

person808: It's so we don't exit the loop when we see the closing bracket for the regular `locales` list.

person808:
Hm. Just realized that the locale list I tested the TC apk with had all the locales in `l10n.toml` so there would have been no difference in locales even with the script running. It should still be working (https://github.com/mozilla-mobile/fenix/issues/3088#issuecomment-592170944) but I'll merge this tonight so I can double check a TC apk tomorrow.

Pike:
This changes an l10n project  configuration file w/out any feedback or review from the l10n team.

This is not OK, and I expect additional entries in l10n.toml files to be errors in linters at some point.

Please find a different spot to put this data.

CC @Delphine

PR #12776, opened 2020-07-21
[79.0.0] Start using release builds of GeckoView.

pocmo:


pocmo:
Tested the build from taskcluster and it is looking good, using the release version of GeckoView 79.

Issue #10026, opened 2020-04-17
FNX2-13050 ⁃ [Bug] Changing TalkBack Setting While Browsing Doesn't Change Toolbar Scroll Flags


ekager:
## Steps to reproduce
1. Go to any website without TalkBack on
1. Turn TalkBack on via a shortcut without leaving browsing

### Expected behavior
Toolbar readjusts scroll flags to new a11y setting

### Actual behavior
Toolbar does not readjust. 

I think we just have to add an Accessibility Service Listener in the browser that changes the settings and flags responsively to fix this. We used to have this working in Focus IIRC

### Device information
* Android device: ?
* Fenix version: ?

abodea:
Hi, @ekager note that I re-checked this issue with the latest **Nightly 7/27** with **Google Pixel 4 XL (Android 11).**
Please note the following: Following the steps from the description after enabling the TalkBack nothing major happened I was able to normally scroll and the Toolbar disappeared when I scrolled down and appeared when I scrolled up.
Is this the intended behavior? I don't know exactly how I can check or see the Toolbar Scroll Flags.
[Video](https://drive.google.com/file/d/10GFO-r909E-lXTqwDzDZZAJX-PSQrzYy/view).

abodea:
**Verified** as **fixed** on the latest **Nightly 7/30** with **Google Pixel 4 XL (Android 11).**
I checked the following scenarios:
 - When the talkback is ON before launching Fenix, the toolbar doesn't disappear when scrolling, but if deactivating without leaving the app, the toolbar is scrolled again.
- When the talkback is OFF and reaching any website the toolbar disappears, after activating the talkback toolbar doesn't disappear anymore.

PR #12826, opened 2020-07-22
For #10026: Update scroll flags on a11y state change.

mcarare:
### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
@@ -1026,6 +1026,6 @@ abstract class BaseBrowserFragment : Fragment(), UserInteractionHandler, Session
     }
 
     override fun onAccessibilityStateChanged(enabled: Boolean) {
-        TODO("Not yet implemented")
+        browserToolbarView.setScrollFlags(enabled)
     }
 }
--- a/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
@@ -12,6 +12,7 @@ import android.view.Gravity
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
+import android.view.accessibility.AccessibilityManager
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
```diff
@@ -1012,4 +1017,8 @@ abstract class BaseBrowserFragment : Fragment(), UserInteractionHandler, Session
         private const val REQUEST_CODE_PROMPT_PERMISSIONS = 2
         private const val REQUEST_CODE_APP_PERMISSIONS = 3
     }
+
+    override fun onAccessibilityStateChanged(enabled: Boolean) {
+        browserToolbarView.setScrollFlags(enabled)
```
ekager, line 1022: could you verify on-device that this works for bottom toolbar? Looking at `setScrollFlags` it look like it might early return if `shouldUseBottomToolbar` before disabling scroll? Didn't test this myself yet though 🤔

mcarare: Yes, but if the toolbar is at bottom the flags weren't set in the first place, so there is no need to update? 🤔

ekager: So is it not possible to disable the toolbar hiding on the bottom toolbar? Sorry I don't know much about these toolbar behaviors 😓

mcarare: AFAIK bottom toolbar uses a custom `BrowserToolbarBottomBehavior` to scroll away, while top toolbar uses these flags  to scroll, and they are updated and the toolbar is expanded when accessibility services are enabled. So these flags are set and updated only for top toolbar.

ekager: So ideally, the bottom toolbar would also be able to be static for talkback users/dynamic otherwise in the same way as the top toolbar, but we can file a followup for this (there might be a very good reason why we can't do that right now)

PR #12225, opened 2020-07-03
Bug 1652979 - Stop producing fennecNightly, fenixNightly and fenixBeta

JohanLorenzo:
### Pull Request checklist

- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features. In addition, it includes a screenshot of a successful [accessibility scan](https://play.google.com/store/apps/details?id=com.google.android.apps.accessibility.auditor&hl=en_US) to ensure no new defects are added to the product.

### After merge
- [ ] **Milestone**: Make sure issues finished by this pull request are added to the [milestone](https://github.com/mozilla-mobile/fenix/milestones) of the version currently in development.

### To download an APK when reviewing a PR:
1. click on Show All Checks,
2. click Details next to "Taskcluster (pull_request)" after it appears and then finishes with a green checkmark,
3. click on the "Fenix - assemble" task, then click "Run Artifacts".
4. the APK links should be on the left side of the screen, named for each CPU architecture

PR #12827, opened 2020-07-22
Simplify build variants to just: debug, nightly, beta and release.

pocmo:


pocmo:
@JohanLorenzo That's the direction I would like to go in. After this change we will only have *debug*, *nightly*, *beta* and *release*.

* This removes the engine flavor dimension since we will have a 1:1 mapping from build type to geckoview release channel
* This uses the name "release" again instead of "production". This is more in line with our wording in other places and also in the Android world.
* Instead of the `forPerformance` build type we now have flags that can be set in *local.properties* (e.g. `debuggable`) which lets us create builds for performance tests that resemble our actual release configurations.
* To simplify things we will now use the fennec-way of generating a version code across all release channels, no need to have the fenix way around just for Nightly.

pocmo:
Thank you for all the hints. I'll try to update those files and will report back. :)

> By the way, let's sync up whenever you want to land it, I'll have to swap secrets right after it.

Will this cause problems with the two active release branches? Or will we be able to have both systems, old and new, working? :)

JohanLorenzo:
Good question, let's see in details. 

* `debug` has no secrets so nothing to do here. 
* `nightly` was done in #12225 so nothing to here either.
* `beta` is currently mapped to fenixBeta, but #12225 landed on both master and the release branches, so it's safe to handle now.
* `release` is a brand new name which mean I can even fill it in now. 

Therefore, I don't any secrets to swap right after the landing, after all 😁

taskcluster/fenix_taskgraph/transforms/push_apk.py
```diff
@@ -40,11 +40,7 @@ def build_worker_definition(config, tasks):
         )
 
         build_type = task["attributes"]["build-type"]
-        # Fenix production doesn't follow the rule {product}-{channel}
-        if build_type == "production":
-            worker_definition["certificate-alias"] = "fenix"
-        # Neither do Fennec flavored builds
-        elif build_type.startswith("fennec-"):
+        if build_type == "beta" or build_type == "nightly":
             worker_definition["certificate-alias"] = build_type
```
pocmo, line 44: This is now a different value for `build-type` and I do not know what `certificate-alias` is used for.

JohanLorenzo: Let's keep the same values for certificate alias, for now. We have to coordinate the renaming between here and there https://github.com/mozilla-releng/scriptworker-scripts/blob/d9f5f854fd3aff0db498ad6e502e00f63d79f6fe/pushapkscript/docker.d/worker.yml#L79. I'll take care of it in a followup.

In order words, we could either change the logic here to keep the output the same, or we could put all values in kind.yml. The latter is probably simpler and would look like this: 

```yaml
 job-template: 
    worker:
        certificate-alias:
            # TODO rename these values to use non-legacy terms.
            by-build-type:
                nightly: fenix-production
                beta: fennec-beta
                release: fennec-production
```

You'll have to update https://github.com/mozilla-mobile/fenix/blob/e2e684b4d452378cf07c70b00f921afe9b5b687a/taskcluster/fenix_taskgraph/transforms/push_apk.py#L21 to include `worker.certificate-alias`.

taskcluster/ci/signing/kind.yml
```diff
@@ -20,15 +20,15 @@ job-template:
...
-            (fennec-.+|nightly|android-test-nightly):
+            (nightly|beta|release|android-test-nightly):
                 by-level:
                     '3': signing
                     default: dep-signing
             default: dep-signing
     worker:
         signing-type:
             by-build-type:
-                fennec-(beta|production):
```
JohanLorenzo, line 31: Last nit: Update `index.by-build-type` down below, with the same kind of changes. My local testing shows the beta signing task isn't indexed anymore.

taskcluster/fenix_taskgraph/transforms/push_apk.py
```diff
@@ -29,24 +29,3 @@ def resolve_keys(config, tasks):
                 }
             )
         yield task
-
-
-@transforms.add
-def build_worker_definition(config, tasks):
```
JohanLorenzo, line 35: 👍 to this removal

JohanLorenzo:
FYI: This patch caused a regression in mozilla-central. It's tracked at https://bugzilla.mozilla.org/show_bug.cgi?id=1657898

PR #12869, opened 2020-07-23
No issue: Fix recent UI test breakage

AaronMT:
Fixes recent UI test breakage. This workaround just checks for the a view containing a string of either plural or not. Also added a `waitForIdle` which was missing to better stabilize opening the nav toolbar.

```Kotlin
    <!-- Message announced to the user when tab tray is selected with 1 tab -->
    <string name="open_tab_tray_single">1 open tab. Tap to switch tabs.</string>
    <!-- Message announced to the user when tab tray is selected with 0 or 2+ tabs -->
    <string name="open_tab_tray_plural">%1$s open tabs. Tap to switch tabs.</string>

### DIFF
--- a/app/src/androidTest/java/org/mozilla/fenix/ui/robots/BrowserRobot.kt
+++ b/app/src/androidTest/java/org/mozilla/fenix/ui/robots/BrowserRobot.kt
@@ -366,7 +366,7 @@ class BrowserRobot {
         }
 
         fun openNavigationToolbar(interact: NavigationToolbarRobot.() -> Unit): NavigationToolbarRobot.Transition {
-
+            mDevice.waitForIdle(waitingTime)
             navURLBar().click()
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: app/src/androidTest/java/org/mozilla/fenix/ui/robots/BrowserRobot.kt [hunk 1]
```diff
@@ -366,7 +366,7 @@ class BrowserRobot {
         }
 
         fun openNavigationToolbar(interact: NavigationToolbarRobot.() -> Unit): NavigationToolbarRobot.Transition {
-
+            mDevice.waitForIdle(waitingTime)
```
rpappalax, line 369: is there a reason why you're not using waitNotNull instead?  I'm going to merge this because of all the failures happening, but we should change this later if this isn't what's desired here

Issue #12904, opened 2020-07-24
FNX2-15541 ⁃ Swipe-to-switch tabs feature flag doesn't remove wrapper SwipeGestureLayout


person808:
To implement swipe-to-switch tabs, I wrapped the `fragment_browser` layout in a `SwipeGestureLayout`. This has caused some issues with snackbars that expect to be attached to a `CoordinatorLayout` but now receive a glorified `FrameLayout`. This wrapping may also have performance implications. While we currently have a feature flag for the gesture, the `SwipeGestureLayout` is always present in the layout. To properly gate this feature until it is ready for release we should conditionally inflate the `SwipeGestureLayout` if the feature flag is on.

PR #12909, opened 2020-07-24
For #12904 - Conditionally inflate SwipeGestureLayout in BaseBrowserFragment.

person808:
Changing the root view of BaseBrowserFragment from a CoordinatorLayout
to a SwipeGestureLayout has caused some regressions, particularly in
snackbar behavior. Lets prevent those regressions from occuring in
builds where the feature flag for gestures is off by only adding the
SwipeGestureLayout when the feature flag is on.

Layout with feature flag on:
Screen Shot 2020-07-24 at 8 50 21 AM

Feature flag off:
Screen Shot 2020-07-24 at 8 46 17 AM

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/FeatureFlags.kt
+++ b/app/src/main/java/org/mozilla/fenix/FeatureFlags.kt
@@ -30,9 +30,9 @@ object FeatureFlags {
     val tabTray = Config.channel.isNightlyOrDebug
 
     /**
-     * Enables swipe on toolbar to switch tabs
+     * Enables gestures on the browser chrome that depend on a [SwipeGestureLayout]
      */
-    val swipeToSwitchTabs = Config.channel.isNightlyOrDebug
+    val browserChromeGestures = Config.channel.isNightlyOrDebug
 
     /**
      * Enables viewing tab history
--- a/app/src/main/res/layout/browser_gesture_wrapper.xml
+++ b/app/src/main/res/layout/browser_gesture_wrapper.xml
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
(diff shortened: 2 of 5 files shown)


app/src/main/java/org/mozilla/fenix/FeatureFlags.kt
```diff
@@ -26,7 +26,7 @@ object FeatureFlags {
     val tabTray = Config.channel.isNightlyOrDebug
 
     /**
-     * Enables swipe on toolbar to switch tabs
+     * Enables gestures on the browser chrome that depend on a [SwipeGestureLayout]
      */
-    val swipeToSwitchTabs = Config.channel.isNightlyOrDebug
+    val browserChromeGestures = Config.channel.isNightlyOrDebug
```
person808, line 31: Renamed since future gestures might also rely on `SwipeGestureLayout`

person808:
@boek Wonder if you think this is necessary? I think I've caught most of the snackbar issues but I'm not sure if there are other places/ways we assume the root layout of BrowserFragment is a CoordinatorLayout

boek:
@person808 Hmm.. I don't think it would hurt!

app/src/main/res/layout/browser_gesture_wrapper.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<org.mozilla.fenix.browser.SwipeGestureLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    android:id="@+id/gestureLayout"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent">
+
+    <org.mozilla.fenix.browser.TabPreview
```
boek, line 10: Does the ordering of these matter? Before the TabPreview was underneath the coordinatorlayout and now I think it will be above it?

person808: I haven't noticed any difference in behavior.

PR #12918, opened 2020-07-24
Revert "Bug 1652979 - Stop producing fennecNightly, fenixNightly and …

mcomella:
…fenixBeta (#12225)"

This reverts commit 7e7d69cb8e4514c86c4677b3b18193c5cb66f884
PR https://github.com/mozilla-mobile/fenix/pull/12225

This caused the nightly FNPRMS tests to fail because the
`fennec-nightly` variant no longer exists. We could change the variant
but then the migration code is not covered. The least time consuming
solution for the FE perf team is to revert the change and atomically
implement the long term solution rather than making an intermediate
change now and another change after the long term solution arrives,
which would skew our dashboards (that we don't have the ability to
annotate) with multiple incomparable builds.

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -72,6 +72,20 @@ android {
                 "sharedUserId": "org.mozilla.fenix.performancetest.sharedID"
             ]
         }
+        fenixNightly releaseTemplate >> {
+            applicationIdSuffix ".fenix.nightly"
+            buildConfigField "boolean", "USE_RELEASE_VERSIONING", "true"
+            def deepLinkSchemeValue = "fenix-nightly"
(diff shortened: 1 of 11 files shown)


mcomella:
@JohanLorenzo @pocmo Due to the issues in https://github.com/mozilla-mobile/perf-frontend-issues/issues/137, I'm reverting this change. I'll send an email for us to discuss some solutions. :)

mcomella:
csadilek raised the concern that this change could cause us to upload the wrong build to the Play Store so, despite the performance tests being broken, it's too risky to land without @JohanLorenzo seeing. I'm going to close to prevent this from accidentally being merged but this needs to be resolved soon because the performance tests won't be running without it.

Issue #12642, opened 2020-07-16
[Bug] delay when loading homescreen content


eliserichards:
This has been reported in https://github.com/mozilla-mobile/fenix/issues/12629 and https://github.com/mozilla-mobile/fenix/issues/12284

There is a major visual delay while we load our homescreen icons, collections, etc. We should think about improving the loading experience (either UX or programatically) to reduce the confusion here.

eliserichards:
CC: @brampitoyo @apbitner

brampitoyo:
@eliserichards Thanks for pinging us on this. Technically, performance is my area of responsibility, and homescreen is @apbitner’s.

I have some more questions that will be quicker to answer and resolve in person.

Talk to you at tomorrow’s stand up, or shortly afterwards!

brampitoyo:
@eliserichards Taking a second look at this issue, it seems as if the problem doesn’t just apply to the homescreen content. It applies to webpages, as well.

**While viewing a webpage**, here’s what happens if I minimise Firefox and reopen it again:

![Untitled](https://user-images.githubusercontent.com/51007/88014618-7f8f5980-cb73-11ea-95d6-70542b7964c9.gif)

**While viewing the new tab page**, notice how the collection box get “redrawn” and have a slight grey flicker:

![Untitled](https://user-images.githubusercontent.com/51007/88014999-5de2a200-cb74-11ea-8224-74769e460959.gif)

**While the tab tray is open**, notice how we replay the open transition/animation, even though the tray was already open:

![Untitled](https://user-images.githubusercontent.com/51007/88015236-d8132680-cb74-11ea-9832-a8337853418c.gif)

**While switching tabs**, notice how each webpage briefly flashes or fades in before they appear, giving a sense that they reload all the time:

![Untitled](https://user-images.githubusercontent.com/51007/88015805-31c82080-cb76-11ea-8b52-4a5b5e5bf155.gif)

-----

The behaviour we want is pretty straightforward, but possibly hard to achieve:

**An instantaneous transition with zero flickering** whenever Firefox changes state: whether it be from minimised to maximised, from one tab to another, etc.

This is different from transitions when you open a modal bottom sheet (Site Information Panel, tab tray, etc.), long press an item to open a context menu, go back and forth within Settings or menu/submenu. In those scenarios, transitions are necessary to give a sense of continuity between pages.

We already have this behaviour on our in-app pages: Settings, History, Bookmarks, etc.

[...]

PR #13009, opened 2020-07-27
For #12642 - Cache top sites and collections to avoid flash

ekager:
This is a first try at removing some of the flash on fresh start. 

ezgif com-video-to-gif(3)

### Pull Request checklist

### After merge

1. click on Show All Checks,


mcomella:
@ekager Can you double-check with the a-c team (maybe @csadilek) on the approach? iirc, at one point, @MarcLeclair wrote an independent cache for top sites to improve startup but a-c wanted to avoid duplication of state and have a more comprehensive solution.

We ended up shipping a hacky solution – prefetch in Application.onCreate and read from the LiveData observer in the Activity https://github.com/mozilla-mobile/fenix/pull/11668/files – but we're in the process of adding a better long term solution – querying the DB on demand instead of using a hacky prefetch https://github.com/mozilla-mobile/android-components/issues/7459

Also, I've heard we're considering changing the start screen so I wonder if this will be necessary if that's changed?

mcomella:
Also, thank you for looking into this – I'm really excited to improve our performance and avoid unnecessary redraws and flashes! :)

mcomella:
> Also, I've heard we're considering changing the start screen so I wonder if this will be necessary if that's changed?

Ignore this ^ comment: I expect users will still hit this when having zero open tabs or in other circumstances

ekager:
Did #11668 solve the top sites issue (even hackily)? I still see them flicker/slide in quickly on fresh start?

MarcLeclair:
@ekager  we've noticed that too. We've requested a different API from AC. Although, I'm not sure caching them in shared prefs being preferable (as @mcomella  said , I did try to do that). I think a better approach would be for us to be able to block the UI until we get what we need, which we can't do with LiveData right now.  SharedPrefs could work too since we don't have to rework the current APIs and we could block on them if ever they took too long.  Thoughts?

ekager:
I haven't actually profiled the difference, but I'm assuming (yes, I know the warning about assuming 😉) that shared pref reads are faster than DB storage reads? Might not be significant enough to care though and if AC is already looking into getting the new APIs then I am happy to abandon this hack! I honestly don't have strong thoughts except I am anti-flashy/jittery homescreen. 🙅

mcomella:
> Did #11668 solve the top sites issue (even hackily)?

The hack solution doesn't block so it causes a race: if the top sites are available before the initial draw, they're added. If not, they flicker. fwiw, debug builds are always too slow to win the race. Unfortunately, something changed in nightly that causes this to not work as consistently anymore but it's pretty effective in the migration build for the default top sites.

> but I'm assuming (yes, I know the warning about assuming 😉) that shared pref reads are faster than DB storage reads?

It's dangerous! 😜 In theory, DBs are generally hyper optimized to be performant while SharedPrefs' focus isn't performance (I think it makes some compromises for data consistency against multiple readers to the same file). In any case, one thing we've learned about performance is that you really don't know for sure until you measure! For example, maybe SharedPrefs is faster for 3 top sites but significantly slower for 100! Or SharedPrefs is slower than DBs for the initial read but faster for later reads. 🤷 

To avoid redundant work, perhaps the best path forward here is to wait for the changes to the API I heard gabrielluong was making, ensure we have a blocking API for use on startup and then measure. If it's not performant enough, we can try adding a cache (SharedPrefs or otherwise). Are those changes expected to happen soon enough?

> I honestly don't have strong thoughts except I am anti-flashy/jittery homescreen. 🙅

➕ 1️⃣

Issue #12856, opened 2020-07-23
FNX2-15504 ⁃ Save a collection button in tabs tray


vesta0:
## User Story
* As a user, I want to be able to quickly save collection directly from the tabs tray, so that it is convenient and easy to find.

## Dependencies
* AC work in progress to allow adding buttons to tabs tray. 

### Acceptance Criteria
* [ ] I can quickly save a collection directly from the tabs tray
* [ ] UX to re-evaluate whether we should remove "save collection" from homescreen" once we add it to the tabs tray 

Abstract link: https://app.abstract.com/projects/8afc4b50-7667-11e9-94ba-3558f8596822/branches/master/commits/562779285de5fd8e7000e10116956a1693981e42/files/af6f466c-1a56-494d-ae53-20dd23f8a685/layers/99FC9CA6-50D5-4A85-835E-F566185D8075

<img width=350 src="https://user-images.githubusercontent.com/1370580/88430527-97aae700-cdc6-11ea-9bba-03b59f8abd1f.png"/>

vesta0:
I believe @topotropic has mocks for this already.

ebalazs-sv:
Hi! I had tested this issue on Nightly 200729 (Build #22110610) GV 80.0a1 from 7/29 with Pixel 2 (Android 9) and the "Save to collection" button is available in tabs tray as described in the screenshot description. 
I do not have access to the abstract, so I requested it and I will keep the qa:needed label to verify this further after receiving the access.

s-ankur:
Hi, coming from #13065 .  this functionality is already represented in the three dot menu and long press to open multi select. So I feel that the button is redundant. It also makes the tab tray seem slower(due to the newer animation). Please consider reverting

cadeyrn:
> So I feel that the button is redundant. It also makes the tab tray seem slower(due to the newer animation).

I don't have a problem with the redudancy but I agree that the tab tray _feels_ noticeably slower.

ebalazs-sv:
Verified as fixed on Nightly 200729 (Build #22110610) GV 80.0a1 from 7/29 with Pixel 2 (Android 9)  and Beta 80.0.0-beta.1 GV 79 from 7/30 with Samsung Galaxt Tab S3 (Android 8).
I will remove the qa:needed label for now.

ebalazs-sv:
This issue is verified as fixed on Nightly 200818 Build #2015758643 from 8/18 with Pixel 2 (Android 9). 

[...]

PR #13012, opened 2020-07-28
For #12856: Add save to collections button to Tabs Tray

jonalmeida:
Using the ConcatAdapter, we're now able to insert multiple data sources of information into one RecyclerView and preserve layout/scrolling in addition to adding the 'Save to Collection' button.

It was not reasonably feasible to add tests to the TabTrayView so I omitted that for now until we have refactored this class into smaller parts.

out

@topotropic on clicking the 'Save to collection' button, it starts the multi-select state that was recently added instead of the overlay. If we prefer the latter, that should be a small change.

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/migration/MigrationStatusAdapter.kt
+++ b/app/src/main/java/org/mozilla/fenix/migration/MigrationStatusAdapter.kt
@@ -96,7 +96,7 @@ internal class MigrationStatusItemDecoration(
         parent: RecyclerView,
         state: RecyclerView.State
     ) {
-        val position = parent.getChildViewHolder(view).adapterPosition
+        val position = parent.getChildViewHolder(view).bindingAdapterPosition
         val itemCount = state.itemCount
 
         outRect.left = spacing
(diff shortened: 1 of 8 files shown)


ekager:
Pulled this down and have 2 observations: 
- We probably want to hide the button while in multiselect mode since it doesn't really make sense to keep it around
- This might be out of scope of this ticket - but it's pretty disorienting the first time you open the tab tray on fresh start now with the collection button flying to the bottom of the list as the tabs appear. Maybe we should file a followup about improving the first load behavior here? 
ezgif com-video-to-gif(4)

jonalmeida:
> * We probably want to hide the button while in multiselect mode since it doesn't really make sense to keep it around

Thanks, I felt similarly, but I wanted to hear from @topotropic if multi-select is the way forward first - the original designs followed to the Collections overlay which would make this a no-op.

> * This might be out of scope of this ticket - but it's pretty disorienting the first time you open the tab tray on fresh start now with the collection button flying to the bottom of the list as the tabs appear. Maybe we should file a followup about improving the first load behavior here?

Yeah, we push back [this perf ticket](https://github.com/mozilla-mobile/fenix/issues/11185) till after we had the opportunity to land multiple adapters for fallout like this. I'm going to evaluate the effects of Sync Tabs when that lands before we work on perf improvements there.

jonalmeida:
Since we've merged this already, we can iterate after I guess.

@ekager I let some more ideas for perf gains based on your comments: https://github.com/mozilla-mobile/fenix/issues/11185#issuecomment-665206619

Issue #12990, opened 2020-07-27
FNX2-15602 ⁃ [Bug] keyboard stays up on tab switching using swipe


s-ankur:
## Steps to reproduce
Open keyboard e.g for writing a comment here
Drag out the dynamic toolbar and swipe to switch tab

### Expected behavior
Keyboard dismissed

### Actual behavior
Not dismissed

### Device information
* Android device: Asus Zenfone
* Fenix version: Nightly 200724 06:05 (Build #22060613)
AC: 52.0.20200723130248, 69562438f
GV: 80.0a1-20200722094100
AS: 61.0.10

quaintdev:
Theirs more to this. When using keyboard to write a comment one can basically swipe on screen anywhere to switch the tab not just the address bar.

s-ankur:
@quaintdev yes that is a seperate issue #12990

person808:
Going to disable the gesture when the keyboard is open as part of the fix for #12865 so this should become a non-issue.

AndiAJ:
Hi, **verified as fixed** on the latest **`Nightly Build 200730`** using the following devices:
• Google Pixel 3a (Android 10)
• Huawei Mate 20 Lite (Android 9)
• OnePlus A3 (Android 6.0.1)

► Video
![20200730-125055](https://user-images.githubusercontent.com/51314259/88909870-543bf700-d264-11ea-8a00-2d5a8505fad2.gif)

Issue #12865, opened 2020-07-23
FNX2-15511 ⁃ when keypad is out, swiping anywhere switches the tabs


s-ankur:
## Steps to reproduce
(Related to #3481)
Open the keypad e.g. write a comment for this issue
Swipe sideways anywhere

### Expected behavior
Scroll the page maybe

### Actual behavior
Changes tabs

### Device information
* Android device: Asus Zenfone
* Fenix version: Nightly 200723 06:01 (Build #22050611)
AC: 52.0.20200722023149, 5c2458cb8
GV: 80.0a1-20200721094241

person808:
Hi! I'm unable to reproduce this on my Pixel 2 XL running Android 10. Where on the screen are you swiping (over the text box/address bar/web page/etc?). If you are able to provide a screen recording with [show taps](https://www.skyneel.com/show-touches-in-android-phone) enabled that would be very helpful.

person808:
As a side note, we should disable the swipe to switch tabs when the keyboard is visible. Chrome does the same thing.

s-ankur:
@person808 Take a look

![GifMaker_lowquality](https://user-images.githubusercontent.com/33252312/88326454-aef4c080-cd43-11ea-9af9-f6e1690acb5b.gif)

person808:
Thanks! Weird that I can't reproduce. What Android version do you have?

s-ankur:
It's LineageOS with Android 9

Augures01:
Can reproduce on Samsung S6 Edge, Android 7.0

hakkikaancaliskan:
I'm able to reproduce on XZ1 with Android 10
Los 17.1

madb1lly:
Hi all,
I also have this bug with latest nightly on Nokia 8 with android 9.
Cheers 🙂

person808:
@hakkikaancaliskan Would you be able to post a recording as well? I tracked down the issue on older android versions but I still can't reproduce it on Android 10. If its an issue on Android 10 then we definitely need to disable the gesture when the keyboard is open which is unfortunately a harder task.

hakkikaancaliskan:
@person808 here is the video :)
http://imgur.com/a/B1L9BDJ

s-ankur:
It would be very helpful if this issue is prioritised. At the moment it is very difficult to type something and make corrections

Neikon:
this bug doesn't affect me.
[...]

PR #13035, opened 2020-07-28
For #12865, #12990 - Disable swipe to switch tabs gesture when the ke…

person808:
…yboard is visible.

I cannot get my screen recording to work right now, but I've tested on API 21, 24, and 29 (gesture nav and three button nav)

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/browser/ToolbarGestureHandler.kt
+++ b/app/src/main/java/org/mozilla/fenix/browser/ToolbarGestureHandler.kt
@@ -9,7 +9,6 @@ import android.animation.AnimatorListenerAdapter
 import android.app.Activity
 import android.graphics.PointF
 import android.graphics.Rect
-import android.os.Build
 import android.util.TypedValue
 import android.view.View
 import android.view.ViewConfiguration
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mozilla/fenix/browser/ToolbarGestureHandler.kt
```diff
@@ -309,6 +314,30 @@ class ToolbarGestureHandler(
         }.start()
     }
 
+    private fun isKeyboardVisible(rootView: View): Boolean {
```
ekager, line 317: Maybe we could make this as an extension method off of View?

person808:
Was going to try and write some tests but it seems like the only way to mock the sdk version is by creating an object to wrap `Build.Version.SDK_INT` calls or using reflection to set it. I'm not sure if either of those are things we want to do.

See https://github.com/mockk/mockk/issues/107

ekager:
Do you just want to verify certain things happen on certain API versions? You can use `@Config(sdk = [21])` to annotate your tests with the desired SDK versions to run on with `@RunWith(AndroidJUnit4::class)`. (we do this in AC see if that helps :) )

eliserichards:
Let's get these tests written and then this will be ready for merge 😄

Issue #12906, opened 2020-07-24
FNX2-15543 ⁃ [Bug] Visual issue when swiping to delete folder + cancel


sflorean:
## Steps to reproduce
1. Add 1-2 folders.
1. Go to Bookmarks.
1. Swipe **left** for one folder in order to delete the folder.
1. Tap "Cancel" from the confirmation pop-up.

### Expected behavior
No visual issue is displayed.

### Actual behavior
![20200724](https://user-images.githubusercontent.com/58466444/88410033-79e97c00-cdde-11ea-85c9-7e5ec2a4777e.gif)UNITO-UNDERSCORE!183238!

### Device information
* Android device: Samsung Galaxy Note 10 (Android 10)
* Fenix version: Nightly 7/24

cc @person808

sflorean:
Verified as fixed on Nightly 8/10. Closing as verified.

PR #13046, opened 2020-07-28
For #12906 - Fix swipe-to-delete bookmark folder.

person808:
This changes the animation slightly (the folder animates back to its original position while the dialog is displayed, but its the simplest way to update the correct viewholder.

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkTouchHelper.kt
+++ b/app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkTouchHelper.kt
@@ -10,6 +10,7 @@ import android.graphics.drawable.Drawable
 import androidx.appcompat.content.res.AppCompatResources
 import androidx.recyclerview.widget.ItemTouchHelper
 import androidx.recyclerview.widget.RecyclerView
+import mozilla.components.concept.storage.BookmarkNodeType
 import mozilla.components.support.ktx.android.content.getColorFromAttr
 import mozilla.components.support.ktx.android.content.getDrawableWithTint
 import mozilla.components.support.ktx.android.util.dpToPx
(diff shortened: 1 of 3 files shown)


person808:
> (Basically we want to modify the State that should be displayed by the View)

I think its a bit weird to modify the state for this since its a one-time side effect, not a persistent state. For now, I went with creating a new method in the interactor that calls a method in the controller to rebind the viewholder when folders are swiped.

app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkFragment.kt
```diff
@@ -412,4 +413,8 @@ class BookmarkFragment : LibraryPageFragment<BookmarkNode>(), UserInteractionHan
             }
         }
     }
+
+    private fun rebindListItem(position: Int) {
+        bookmarkView.bookmarkAdapter.notifyItemChanged(position)
```
Mugurell, line 418: We should avoid this as it breaks the encapsulation. Now the View exposes the adapter which exposes the Interactor (that should be private in the first place)..
All in all according to our architecture Views should only have one method - `update()` that consumes a State - what the user sees, exposing more methods will lead to technical debt.

app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkTouchHelper.kt
```diff
@@ -41,7 +42,9 @@ class BookmarkTouchCallback(private val interactor: BookmarkViewInteractor) :
      */
     override fun onSwiped(viewHolder: RecyclerView.ViewHolder, direction: Int) {
         val item = (viewHolder as BookmarkNodeViewHolder).item
-        item?.let { interactor.onDelete(setOf(it)) }
+        item?.let {
```
Mugurell, line 45: Based on the previous comments which show that we cannot really obtain an MVI loop for this change I propose going with your first implementation (https://github.com/person808/fenix/commit/55665f13fdd979aae5d8f7a12abe88d556f4f144) here, just that instead of having the adapter being passed to BookmarkTouchHelper we could simply use `viewHolder.bindingAdapter?.notifyItemChanged(viewHolder.bindingAdapterPosition)`.
What do you say?

Mugurell: Maybe not the best solution.. but if the touch helper knows to swipe the item.. it could also know to revert that..?

person808:
Added a couple of tests!

PR #13047, opened 2020-07-29
Closes #12522: Reuse exceptions code

### LINKED ISSUE #12522: FNX2-15305 ⁃ Reuse common elements of tracking protection and login exceptions
In theory these UIs are very similar. Should be able to use the same general layouts, adapters, viewholders, and interactors with different copy, storage layers, and data types. 

Some difficulties could lie the differences in the TP exceptions storage and the Login Exceptions storage, ie reloading, fetching, deleting and working around the AC-provided data types.

NotWoods:
These classes are pretty much copies, so this code moves most logic into shared base classes. We're planning to introduce a shared list component later so this work helps with that too.

- Everything is under `org.mozilla.fenix.exceptions` now
- ViewHolders are identical for both fragments
- Logic is moved from fragments to interactors, making it testable


app/src/main/java/org/mozilla/fenix/exceptions/login/LoginExceptionsInteractor.kt
```diff
@@ -0,0 +1,35 @@
...
+import kotlinx.coroutines.launch
+import mozilla.components.feature.logins.exceptions.LoginException
+import mozilla.components.feature.logins.exceptions.LoginExceptionStorage
+import org.mozilla.fenix.exceptions.ExceptionsInteractor
+
+/**
+ * Interface for the ExceptionsViewInteractor. This interface is implemented by objects that want
+ * to respond to user interaction on the ExceptionsView
+ */
+interface LoginExceptionsInteractor : ExceptionsInteractor<LoginException>
```
ekager, line 17: nit should this kdoc be on the ExceptionsInteractor interface? Seems more general 🤔

app/src/main/java/org/mozilla/fenix/exceptions/login/LoginExceptionsView.kt
```diff
@@ -0,0 +1,32 @@
...
+
+import android.view.ViewGroup
+import androidx.core.view.isVisible
+import kotlinx.android.synthetic.main.component_exceptions.*
+import mozilla.components.feature.logins.exceptions.LoginException
+import org.mozilla.fenix.R
+import org.mozilla.fenix.exceptions.ExceptionsView
+
+/**
+ * View that contains and configures the Exceptions List
```
ekager, line 15: Same here, should this be on `ExceptionsView`?

app/src/main/java/org/mozilla/fenix/exceptions/trackingprotection/TrackingProtectionExceptionsInteractor.kt
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    override fun onDeleteAll() {
+        trackingProtectionUseCases.removeAllExceptions()
+        reloadExceptions()
+    }
+
+    override fun onDeleteOne(item: TrackingProtectionException) {
+        trackingProtectionUseCases.removeException(item)
+        Log.e("Remove one exception", "$item")
```
ekager, line 49: nit - we can remove this log now (I think this was left over from when removing one exception wasn't supported) :)

Issue #13066, opened 2020-07-29
FNX2-15651 ⁃ [Bug] translate release locales to android locale codes


Pike:
## Steps to reproduce
Create a release build

### Expected behavior
Hebrew and Indonesian files should be kept, in `values-iw` and `values-in`.

### Actual behavior
The files are removed.

person808:
Note for QA: Similar to #7104 we need to check if Hebrew and Indonesian files are present in `fennec-production` builds. I updated the `releases/v79.0.0`, `releases/v79.1.0` and `master` branch so I'm not sure if you can check apks built from each branch?

Pike:
Looking at the v80 branch build, https://firefox-ci-tc.services.mozilla.com/tasks/VAmXWrKwRIaahs5KBC_GOQ/runs/0/logs/https%3A%2F%2Ffirefox-ci-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FVAmXWrKwRIaahs5KBC_GOQ%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive.log, I see

```
REMOVE: bg, uz
```

which matches @Delphine 's list, given that `ms`, `ne-NP`, and `nv` don't have files in Fenix at all yet, so they don't show up in the available locales (it's listing files, not l10n.toml).

I wasn't able to find v79.* builds.

sflorean:
@Delphine @Pike 
Tested on 80.0.0 build and Hebrew and Indonesian languages are displayed in the app, currently, 86 languages are being listed. 

> Note for QA: Similar to #7104 we need to check if Hebrew and Indonesian files are present in `fennec-production` builds. I updated the `releases/v79.0.0`, `releases/v79.1.0` and `master` branch so I'm not sure if you can check apks built from each branch?

@person808 I think we will need signed apks/other builds in order to check. on 79.0.0 We checked the existing builds but the languages aren't updated.
We will check to see if we can build local.

sflorean:
@person808 We tried to make a local build for 79.0.0 with the recent changes but we don't have languages displayed in the app (maybe because it built en) and we don't know how to add them. If there is another way to test this, or someone could provide us a build, we are happy to check.

person808:
[...]

PR #13082, opened 2020-07-29
[Release v79.0.0] For #13066 - Translate locale codes to Android specific variants befo…

person808:
…re filtering

Tested locally and verified that `values-iw` and `values-in` resource folders were present after running `python2 ./automation/taskcluster/l10n/filter-release-translations.py`

### DIFF
--- a/automation/taskcluster/l10n/locales.py
+++ b/automation/taskcluster/l10n/locales.py
@@ -11,11 +11,21 @@ import re
 OPEN_LOCALES = "locales = ["
 CLOSE_LOCALES = "]"
 
+# Android uses non-standard locale codes, these are the mappings back and forth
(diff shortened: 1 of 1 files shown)


automation/taskcluster/l10n/locales.py
```diff
@@ -11,11 +11,23 @@
...
     match = re.search('\s*"([a-z]+-?[A-Z]*)",\s*', str)
     if not match:
         raise Exception("Failed parsing locale found in l10n.toml: " + str)
-    return match.group(1)
+
+    locale = match.group(1)
+    if locale in ANDROID_LEGACY_MAP:
+        return ANDROID_LEGACY_MAP[locale]
+    else:
+        return locale
```
Pike, line 30: This is easier done with 

```
return ANDROID_LEGACY_MAP.get(locale, locale)
```

PR #13083, opened 2020-07-29
For #13066 - Translate locale codes to Android specific variants befo…

person808:
…re filtering

### DIFF
--- a/automation/taskcluster/l10n/locales.py
+++ b/automation/taskcluster/l10n/locales.py
@@ -11,11 +11,21 @@ import re
 OPEN_LOCALES = "locales = ["
 CLOSE_LOCALES = "]"
 
+# Android uses non-standard locale codes, these are the mappings back and forth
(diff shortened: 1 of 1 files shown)


automation/taskcluster/l10n/locales.py
```diff
@@ -11,11 +11,23 @@
 OPEN_LOCALES = "locales = ["
 CLOSE_LOCALES = "]"
 
+# Android uses non-standard locale codes, these are the mappings back and forth
+ANDROID_LEGACY_MAP = {
```
liuche, line 15: Do we know if this is all of them? (or just going through the list of missing ones Delphine mentioned)

liuche: If there's android documentation for these non-standard mappings, include a link to it in the comment - consider what someone might need to look through, once we start adding more locales (that could also be non-standard).

person808: It seems like its only these three. I'll add a link to the docs

PR #13100, opened 2020-07-29
Don't init glean in unit tests

NotWoods:
Our unit tests rely on sending data to `MetricController` rather than Glean itself. Glean adds overhead and clutters our logs, it should be fine to turn it off in unit tests.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
+++ b/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
@@ -12,8 +12,8 @@ import android.util.Log.INFO
 import androidx.annotation.CallSuper
 import androidx.appcompat.app.AppCompatDelegate
 import androidx.core.content.getSystemService
-import androidx.work.Configuration.Provider
 import androidx.work.Configuration.Builder
+import androidx.work.Configuration.Provider
 import kotlinx.coroutines.Deferred
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.GlobalScope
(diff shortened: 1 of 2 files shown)


boek:
I think we have a couple of unit tests for Glean and plans to use Glean more directly in the future 🤔

mcomella:
> Our unit tests rely on sending data to `MetricController` rather than Glean itself. Glean adds overhead

@NotWoods Do we know how much overhead? I wonder what challenges this adds to testing vs. the benefits.

Do you know if this makes testing any more challenging?

> clutters our logs,

We should file a bug against glean so they can fix this behavior. If the logging is excessive, I wonder if that's what causes the slowdown in our tests (rather than the Glean.init).

---

@Dexterp37 I think I know what you're going to say 😁 but what do you think about this change?

NotWoods:
It seems to add a couple seconds to tests, but that could be caused by logs. Since we don't test Glean itself it most of our tests it feels unnessecary to initialize it. Tests that do use Glean aren't affected by this change.

mcomella:
> It seems to add a couple seconds to tests,

@NotWoods How many times did you run the tests to validate the difference in run time? i.e. could the couple seconds of perceived gain be attributed to noise?

> Since we don't test Glean itself it most of our tests it feels unnessecary to initialize it. Tests that do use Glean aren't affected by this change.

That's fair – let's see what dexterp37's opinion is and react accordingly.

Dexterp37:
> > clutters our logs,
> 
> We should file a bug against glean so they can fix this behavior. If the logging is excessive, I wonder if that's what causes the slowdown in our tests (rather than the Glean.init).
> 
> @Dexterp37 I think I know what you're going to say 😁 but what do you think about this change?

🤣 Thanks for flagging me on this :-) And let me surprise you!

I believe that, given the small amount of tests in the Fenix repo using the Glean SDK Testing APIs, changing this sounds reasonable! However, I do agree with @mcomella : @NotWoods would you kindly file a bug in our repo? We can consider adding APIs to suppress our logging. 

Are the ~2s added to EACH unit tests? Or to the whole test suite?

NotWoods:
Filed https://github.com/mozilla/glean/issues/1127. I measured the overhead as a total change (whole test suite).

PR #13124, opened 2020-07-30
For #11664 — Fixup MissingResourceExceptions being thrown in exotic Locales

### LINKED ISSUE #11664: FNX2-13560 ⁃ LeanPlum MissingResourceException  No 3-letter language code for locale: cak, lij
Sentry report on Nightly: https://sentry.prod.mozaws.net/operations/fenix-nightly/issues/7513010/?query=is%3Aunresolved%20level%3Afatal

Looks like Leanplum is trying to send messages to the locale `cak` but it's not working, and causing a crash.

Supported languages are [here](https://github.com/mozilla-mobile/fenix/blob/master/l10n.toml) and it does look like `cak` is supported. Maybe it's not supported by the Leanplum SDK/dashboard/localization? cc @lwa-moz

jhugman:
Fixes #11664.

Our kotlin code is not catching the `MissingResourceException` in the `LeanplumMetricsService` which results in the app crashing when the locale isn't known by the device.

Catches the exception, and falls back to the ISO 639 language code. This isn't a great solution, because ISO 639 isn't especially stable.

In practice however this is almost certainly never going to be a problem because Leanplum isn't going to be supported in such exotic locales.

In this case, using the ISO 639 language code allows the error message to be more informative.

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/metrics/LeanplumMetricsService.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/metrics/LeanplumMetricsService.kt
@@ -21,6 +21,7 @@ import org.mozilla.fenix.BuildConfig
 import org.mozilla.fenix.components.metrics.MozillaProductDetector.MozillaProducts
 import org.mozilla.fenix.ext.settings
 import java.util.Locale
+import java.util.MissingResourceException
 import java.util.UUID.randomUUID
 
 private val Event.name: String?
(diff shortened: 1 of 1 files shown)


st3fan:
@liuche this is a patch we would like to ship in 79.0.1

app/src/main/java/org/mozilla/fenix/components/metrics/LeanplumMetricsService.kt
```diff
@@ -170,6 +178,12 @@ class LeanplumMetricsService(private val application: Application) : MetricsServ
         return LEANPLUM_ENABLED_LOCALES.contains(locale)
     }
 
+    private val Locale.iso3LanguageOrNull: String?
```
csadilek, line 181: Later, this looks like a good candidate for re-use. We could uplift to A-C and add to the support-ktx module.

app/src/main/java/org/mozilla/fenix/components/metrics/LeanplumMetricsService.kt
```diff
@@ -83,12 +84,19 @@ class LeanplumMetricsService(private val application: Application) : MetricsServ
...
-            val currentLocale = when (applicationSetLocale != null) {
-                true -> applicationSetLocale.isO3Language
-                false -> Locale.getDefault().isO3Language
-            }
-            if (!isLeanplumEnabled(currentLocale)) {
-                Log.i(LOGTAG, "Leanplum is not available for this locale: $currentLocale")
+            val currentLocale = applicationSetLocale ?: Locale.getDefault()
+            val languageCode =
+                currentLocale.iso3LanguageOrNull
+                ?: currentLocale.language.let {
```
csadilek, line 90: I double-checked this as `language` is annotated with `@RecentlyNotNull` and I wasn't sure what that meant. Seems safe though!

app/src/main/java/org/mozilla/fenix/components/metrics/LeanplumMetricsService.kt
jhugman: 👍

app/src/main/java/org/mozilla/fenix/components/metrics/LeanplumMetricsService.kt
NotWoods: `RecentlyNotNull` is equivalent to `NotNull` but lets you incrementally update rather than breaking builds (aka no warning). You can just treat it as not null.

jhugman: 👍

Issue #12887, opened 2020-07-24
FNX2-15527 ⁃ [Bug] Constantly flickering icons in tab history for a loading page


s-ankur:
## Steps to reproduce
Open a webpage that takes a few secs to load
Long press back button

### Expected behavior
The tab history looks ok

### Actual behavior
It flickers constantly

### Device information
* Android device: Asus Zenfoen
* Fenix version: Nightly 200723 06:01 (Build #22050611)
AC: 52.0.20200722023149, 5c2458cb8
GV: 80.0a1-20200721094241
AS: 61.0.10

ekager:
We'll need to add a change payload here (in `TabHistoryAdapter`) and more efficiently notify/bind the specific changes we need to update in the list (instead of updating the entire list on every change).

eliserichards:
#12973 also has good STR and build info if needed

ebalazs-sv:
Verified as fixed on Nightly 200803 (Build #22161809) GV 81.0a1 from 8/3 with Motorola Moto G6 (Android 8) and Pixel 2 (Android 9).

sv-ohorvath:
Verified as fixed on 80.0.0-beta6 on several devices

PR #13132, opened 2020-07-30
 For #12887: Use ListAdapter to diff history 

NotWoods:
### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/tabhistory/TabHistoryAdapter.kt
+++ b/app/src/main/java/org/mozilla/fenix/tabhistory/TabHistoryAdapter.kt
@@ -23,7 +23,7 @@ class TabHistoryAdapter(
 
     override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): TabHistoryViewHolder {
         val view = LayoutInflater.from(parent.context)
-            .inflate(R.layout.history_list_item, parent, false)
+            .inflate(R.layout.tab_history_list_item, parent, false)
         return TabHistoryViewHolder(view, interactor)
     }
 
--- a/app/src/main/java/org/mozilla/fenix/tabhistory/TabHistoryAdapter.kt
+++ b/app/src/main/java/org/mozilla/fenix/tabhistory/TabHistoryAdapter.kt
@@ -6,7 +6,8 @@ package org.mozilla.fenix.tabhistory
 
 import android.view.LayoutInflater
 import android.view.ViewGroup
-import androidx.recyclerview.widget.RecyclerView
+import androidx.recyclerview.widget.DiffUtil
+import androidx.recyclerview.widget.ListAdapter
 import org.mozilla.fenix.R
 
 data class TabHistoryItem(
@@ -18,23 +19,23 @@ data class TabHistoryItem(
 
 class TabHistoryAdapter(
     private val interactor: TabHistoryViewInteractor
-) : RecyclerView.Adapter<TabHistoryViewHolder>() {
-
-    var historyList: List<TabHistoryItem> = emptyList()
-        set(value) {
-            field = value
-            notifyDataSetChanged()
-        }
+) : ListAdapter<TabHistoryItem, TabHistoryViewHolder>(DiffCallback) {
 
     override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): TabHistoryViewHolder {
-        val view =
-            LayoutInflater.from(parent.context).inflate(R.layout.history_list_item, parent, false)
+        val view = LayoutInflater.from(parent.context)
+            .inflate(R.layout.history_list_item, parent, false)
         return TabHistoryViewHolder(view, interactor)
     }
 
     override fun onBindViewHolder(holder: TabHistoryViewHolder, position: Int) {
-        holder.bind(historyList[position])
+        holder.bind(getItem(position))
     }
 
-    override fun getItemCount(): Int = historyList.size
+    internal object DiffCallback : DiffUtil.ItemCallback<TabHistoryItem>() {
+        override fun areItemsTheSame(oldItem: TabHistoryItem, newItem: TabHistoryItem) =
+            oldItem.url == newItem.url
+
+        override fun areContentsTheSame(oldItem: TabHistoryItem, newItem: TabHistoryItem) =
+            oldItem == newItem
+    }
 }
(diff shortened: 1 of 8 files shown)


app/src/main/java/org/mozilla/fenix/tabhistory/TabHistoryViewHolder.kt
```diff
@@ -7,28 +7,34 @@ package org.mozilla.fenix.tabhistory
...
+    }
 
     fun bind(item: TabHistoryItem) {
-        view.history_layout.overflowView.isVisible = false
-        view.history_layout.urlView.text = item.url
-        view.history_layout.loadFavicon(item.url)
+        this.item = item
 
-        view.history_layout.titleView.text = if (item.isSelected) {
+        history_layout.displayAs(LibrarySiteItemView.ItemType.SITE)
```
person808, line 28: Oh cool I didn't know about this. But it shows the non-functional overflow menu. Maybe...
```suggestion
        history_layout.displayAs(LibrarySiteItemView.ItemType.SITE)
        history_layout.overflowView.isVisible = false
```

We might also want to update the padding for the title and url TextViews to account for the hidden overflow button though that could also be done in a future PR

app/src/main/java/org/mozilla/fenix/tabhistory/TabHistoryAdapter.kt
```diff
@@ -18,23 +19,23 @@ data class TabHistoryItem(
...
-            field = value
-            notifyDataSetChanged()
-        }
+) : ListAdapter<TabHistoryItem, TabHistoryViewHolder>(DiffCallback) {
 
     override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): TabHistoryViewHolder {
-        val view =
-            LayoutInflater.from(parent.context).inflate(R.layout.history_list_item, parent, false)
+        val view = LayoutInflater.from(parent.context)
+            .inflate(R.layout.history_list_item, parent, false)
```
person808, line 26: I shouldn't have used this layout because it contains the header and button views we use in the history fragment. It also wraps the whole `LibrarySiteItemView` in another LinearLayout. I think it would be best to create a new layout file that just contains the `LibrarySiteItemView` with the same attributes that `history_layout` has. And this becomes something like

```suggestion
            .inflate(R.layout.tab_history_list_item, parent, false)
```

app/src/main/java/org/mozilla/fenix/tabhistory/TabHistoryView.kt
```diff
@@ -60,6 +57,7 @@ class TabHistoryView(
     init {
         tabHistoryRecyclerView.adapter = adapter
         tabHistoryRecyclerView.layoutManager = layoutManager
+        tabHistoryRecyclerView.itemAnimator = null
```
person808, line 60: Is this just to disable change animations? We could do something like this:

```suggestion
        (tabHistoryRecyclerView.itemAnimator as SimpleItemAnimator).supportsChangeAnimations = false
```

We also do this in `HistoryView` since the alpha animation in the change animations causes flickering.

app/src/main/java/org/mozilla/fenix/tabhistory/TabHistoryViewHolder.kt
```diff
@@ -7,28 +7,34 @@ package org.mozilla.fenix.tabhistory
...
-    private val view: View,
+    view: View,
     private val interactor: TabHistoryViewInteractor
-) : RecyclerView.ViewHolder(view) {
+) : ViewHolder(view) {
+
+    private lateinit var item: TabHistoryItem
+
+    init {
+        itemView.setOnClickListener { interactor.goToHistoryItem(item) }
```
person808, line 22: This needs to be moved back to the bind method. Otherwise we're attempting to go to an item that isn't initialized yet

person808: Also it seems like we may need to set the listener on `history_layout` because this isn't working at all for me when its set on `itemView` for some reason. So it becomes `history_layout.setOnClickListener { interactor.goToHistoryItem(item) }` in `bind`

app/src/main/java/org/mozilla/fenix/tabhistory/TabHistoryView.kt
person808: Did a little bit of testing and I'm not sure we need to disable any animations at all since `ListAdapter` won't rebind anything if it didn't change.

app/src/main/java/org/mozilla/fenix/tabhistory/TabHistoryViewHolder.kt
NotWoods: `displayAs` handles hiding the overflow button already, and the layout handles it

app/src/main/java/org/mozilla/fenix/tabhistory/TabHistoryViewHolder.kt
NotWoods: This pattern is a bit better for perf because each listener is a new object allocation. If you're worried about `bind` being called after clicking, we can change `item` to be nullable instead of lateinit.

NotWoods: Also agree on moving to history_layout, that's a typo :)

person808: > This pattern is a bit better for perf because each listener is a new object allocation. If you're worried about bind being called after clicking, we can change item to be nullable instead of lateinit.

Double check myself and you're right. :) We can leave it where it is.

app/src/main/java/org/mozilla/fenix/tabhistory/TabHistoryViewHolder.kt
person808: I still see the overflow menu when I test this. Looking at the method it looks like the overflow menu is only hidden for separators

NotWoods: Oops, sorry! I was mixing up components

Issue #13179, opened 2020-07-31
[Bug] RuntimeException Cannot confirm/dismiss a Prompt twice.


liuche:
Not a ton of crashes, but easily repros: https://sentry.prod.mozaws.net/operations/fenix-fennec-beta/issues/8978923/?query=is%3Aunresolved%20baseprompt

## Steps to reproduce
Go to a youtube video
Click the forward button to open the share sheet
Back out of sharesheet

### Expected behavior
Doesn't crash

### Actual behavior
Crashes

### Device information

* Android device: Pixel 3
* Fenix version: Nightly 7/29

softvision-miralobontiu:
Verified as fixed on the latest Firefox Nightly build from 8/5 with:
* Samsung Galaxy Tab S6 (Android 9),
* Google Pixel (Android 10),
* Sony Xperia Z5 Premium (Android 7.1.1),
* OnePlus 5T (Android 9).

PR #9671, opened 2020-04-03
Closes #9495: Dismiss the ShareFragment in onPause()

Mugurell:
Doing this will ensure no corruption of the initial share to apps list while the user went away.

### Pull Request checklist

- [x] **Tests**: No tests. Small change.
- [x] **Screenshots**: No screenshots. Simple flow.
- [x] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md).

### After merge
- [ ] **Milestone**: Make sure issues finished by this pull request are added to the [milestone](https://github.com/mozilla-mobile/fenix/milestones) of the version currently in development.

### To download an APK when reviewing a PR:
1. click on Show All Checks,
2. click Details next to "Taskcluster (pull_request)" after it appears and then finishes with a green checkmark,
3. click on the "Fenix - assemble" task, then click "Run Artifacts".
4. the APK links should be on the left side of the screen, named for each CPU architecture

PR #13238, opened 2020-08-03
For #13179 - Ensure we only dismiss share prompt once

ekager:
### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/share/ShareFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/share/ShareFragment.kt
@@ -43,11 +43,6 @@ class ShareFragment : AppCompatDialogFragment() {
         viewModel.loadDevicesAndApps()
     }
 
-    override fun dismiss() {
-        consumePrompt { onDismiss() }
-        super.dismiss()
-    }
-
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setStyle(STYLE_NO_TITLE, R.style.ShareDialogStyle)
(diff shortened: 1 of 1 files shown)


ekager:
@jonalmeida from what I can tell via debugging - we are calling the dialog `dismiss` in `onPause` as well (so we close the share sheet when we leave Fenix) and end up triggering this twice
```
consumePrompt {
            onDismiss()
        }
```

ekager:
Are you saying we should be able to consume the prompt twice without crashing?

jonalmeida:
Yes, it shouldn't crash with the current behaviour. I can see from a quick git-blame lookup that the original behaviour that was added worked as expected when it landed: https://github.com/mozilla-mobile/fenix/commit/fe034226a335a2e48b4fa94e94aa80d7fe278f57#diff-950ceadcf9bd0ff8d1b61fae91f56240R45

I imagine something has either changed in AC or the share fragment lifecycle has changed where we are holding on to a reference of the hold prompt request (my bet is on the former).

ekager:
Later we added the `onPause` callback that calls `dismiss` as well -> https://github.com/mozilla-mobile/fenix/pull/9671/files

jonalmeida:
Ah, great find! I missed that.

In order to support that behaviour and not make the duplicate calls to consume the same prompt request twice, we could move the `consumePrompt` call to the `onPause` and remove the overridden `dismiss`. This would avoid adding extra state to the fragment.

I tested on that change for a bit and it seems to support the early dismissal that we're looking for without crashing on the consume.

PR #13263, opened 2020-08-04
For #13262: Update gradle 

### LINKED ISSUE #13262: Update Android Gradle plugin
Release notes: Release notes: developer.android.com/studio/releases/gradle-plugin

NotWoods:


NotWoods:
Android Gradle Plugin 3.3 and up is compatible with Gradle 6.5. I'll work on AGP updates too but that requires additional work as the breaking changes aren't well documented.

PR #13268, opened 2020-08-04
FNX-14449 ⁃ Gradle configuration avoidance

NotWoods:


liuche:
@mcomella you were recently looking at task parallelization, is this something you could review?
https://docs.gradle.org/current/userguide/task_configuration_avoidance.html

mcomella:
@NotWoods Do you happen to know what the savings are here or how a similar change has impacted android-components?

NotWoods:
We implemented similar changes in [AC](https://github.com/mozilla-mobile/android-components/pull/7036). 
> These few commits get us to 3749 tasks created during config phase, vs current baseline of 23519.

app/build.gradle
```diff
@@ -735,5 +738,7 @@ tasks.register("updateCookiesExtensionVersion", Copy) { task ->
     updateExtensionVersion(task, 'src/main/assets/extensions/cookies')
 }
 
-preBuild.dependsOn updateAdsExtensionVersion
-preBuild.dependsOn updateCookiesExtensionVersion
+preBuild.configure {
```
mcomella, line 741: What's `preBuild`? Does this run every time we execute a gradle command? Or is it smart and only executes before Android compilation jobs?

app/build.gradle
```diff
@@ -621,11 +624,11 @@ task printVariants {
             engine: 'geckoNightly',
             name: 'androidTest',
         ])
-        println 'variants: ' + groovy.json.JsonOutput.toJson(variants)
+        println 'variants: ' + JsonOutput.toJson(variants)
     }
 }
 
-task buildTranslationArray {
+tasks.register('buildTranslationArray') {
```
mcomella, line 631: I don't see this called from anywhere and it doesn't set dependencies but it seems to correctly fill the `SUPPORTED_LOCALE_ARRAY`: do you know why this works and why it will continue to work after lazy configuration?

I'm wondering if this has been running in the configuration phase this whole time (e.g. the code is not in `doLast`) and lazily registering it will prevent it from working.

NotWoods: I think you're correct and the array is being populated as a side effect. I'll revert this line and file an issue.

app/build.gradle
NotWoods: I'm not sure where preBuild is set up. I tried to keep the logic consistent by hooking into `configure`, but we can also keep this line as is for now.

PR #13309, opened 2020-08-05
FNX-14490 ⁃ Default to email sign in if no camera

Mugurell:
App can be installed on devices with no camera modules. Like Android TV boxes.
Will skip presenting the option to sign in by scanning a qr code in this case
and default to login with email and password.

### Pull Request checklist

- [x] **Tests**: No tests. Small change.

DefaultToSignInWithEmail
[video](https://drive.google.com/file/d/1xnyJppfeRIK1-ufxyXjDI6kxQudGHJxG/view?usp=sharing)

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/settings/SettingsFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/SettingsFragment.kt
@@ -25,7 +25,6 @@ import mozilla.components.concept.sync.AccountObserver
 import mozilla.components.concept.sync.AuthType
 import mozilla.components.concept.sync.OAuthAccount
 import mozilla.components.concept.sync.Profile
-import mozilla.components.support.ktx.android.content.hasCamera
 import org.mozilla.fenix.BrowserDirection
 import org.mozilla.fenix.Config
 import org.mozilla.fenix.HomeActivity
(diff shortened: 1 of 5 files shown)


Mugurell:
The code change is small but not very elegant.
Also there is the issue that the first time sign-in is started users might catch a clear glimpse of an empty fragment.
This was something I tried to avoid with the previous solution but if we're going with this I'd say it's better than to have them see and maybe even interact with the default UX for `TurnOnSyncFragment`.

What other solutions I thought of / tried:
- just hiding the button to start the qr code scanning UX
This was swiftly disregarded as `TurnOnSyncFragment` would have an UI that promoted "Sign in with your camera" and then there would be no option to do so.
@ UX: Maybe we could have a different fragment UI for promoting "Sign in with email" in these cases of devices with no camera?
- having each component wanting to start sign-in checking for device camera and then easily deciding what UX should be used. Many places with duplicated code. Not a solution.
- a nested graph with `startDestination` being a new headless Fragment that would just be responsible for routing to `TurnOnSyncFragment` or to pair with just email.
Doesn't work because this nav-graph/TurnOnSyncFragment can be started with/without animations. Didn't see a way to get the same animations the nav-graph is started with to start the TurnOnSyncFragment.
- an object encapsulating the routing logic to be used by all components wanting to start sign-in. Change needed in many places, sometimes not in line with the code flow, would bring technical debt and would offer not guarantees future components wanting to start sign-in would know to use this instead of navigating directly to TurnOnSyncFragment.

mcarare (approved):
LGTM!

Issue #13218, opened 2020-08-03
Intermittent UI test failure - settingsAddonsItemsTest


sv-ohorvath:
### Firebase Test Run:
Flaky test, failed 1/2 tries.
https://console.firebase.google.com/project/moz-fenix/testlab/histories/bh.66b7091e15d53d45/matrices/8312248502354333158/executions/bs.97b043d607c2c373/testcases/1/test-cases
### Stacktrace:
`androidx.test.espresso.NoMatchingViewException: No views in hierarchy found matching: (with id: org.mozilla.fenix.debug:id/title and with text: is "Recommended")`
### Build:

rpappalax:
More failures:

https://console.firebase.google.com/u/0/project/moz-fenix/testlab/histories/bh.66b7091e15d53d45/matrices/4832838801233999744/executions/bs.5a80c8ca6f4c7e76/testcases/1/test-cases

https://console.firebase.google.com/u/0/project/moz-fenix/testlab/histories/bh.66b7091e15d53d45/matrices/7008303131039026487/executions/bs.48e4df333592e5f/testcases/1/test-cases

sv-ohorvath:
Failed again 2/2: https://console.firebase.google.com/project/moz-fenix/testlab/histories/bh.66b7091e15d53d45/matrices/7730566038549231460/executions/bs.bedea22c9ce042c6/testcases/1/test-cases

sv-ohorvath:
Failed on 8/29, 2/2: https://console.firebase.google.com/project/moz-fenix/testlab/histories/bh.66b7091e15d53d45/matrices/7061633375658984827/executions/bs.a335edcff4ed8e62/testcases/1/test-cases

PR #13311, opened 2020-08-05
For #13218: Fix settingsAddonsItemsTest

AaronMT:
For #13218 – trying the other style of `wait` that @sv-ohorvath has used recently to maybe fix this flaky test.

```androidx.test.espresso.NoMatchingViewException: No views in hierarchy found matching: (with id: org.mozilla.fenix.debug:id/title and with text: is "Recommended")```

Passes locally. Let's try Firebase.


sv-ohorvath:
Passed on firebase, only some flaky tests un-related to this.

Issue #13229, opened 2020-08-03
[Bug] Leanplum deviceId not beign saved


boek:
## Steps to reproduce
Launch app
### Expected behavior
generated device id is cached and reused
### Actual behavior
new deviceId generated every cold boot

### Device information

* Android device: ?
* Fenix version: ?

AndiAJ:
Hi, I've just checked this matter on both the latest **`Nightly 200807`** and **`80.0.0-beta.4`** using the following devices:
• Google Pixel 3a (Android 10)
• OnePlus A3 (Android 6.0.1)

The Device ID displayed after the first launch of Fenix is being used each time the user resumes Fenix after force closing it.

#### ✔️ Nightly 
• Google Pixel 3a (Android 10) - Device ID `25fc7588-0cab-416a-8bda-0eebc8b5d567`

<details>
    <summary>Screenshots (▼ dropdown )</summary>

![image](https://user-images.githubusercontent.com/51314259/89639512-5b8f8000-d8b6-11ea-8034-e8372debe222.png)
![image](https://user-images.githubusercontent.com/51314259/89639633-9c879480-d8b6-11ea-82e1-0e83da017d87.png)
![image](https://user-images.githubusercontent.com/51314259/89640128-98a84200-d8b7-11ea-9273-13e4e6758015.png)

   
</details>


 •OnePlus A3 (Android 6.0.1) - Device ID `bfc47577-4bb4-4819-a243-3cb289fd3b44`

<details>
    <summary>Screenshots (▼ dropdown)</summary>

![image](https://user-images.githubusercontent.com/51314259/89639488-4ca8cd80-d8b6-11ea-92e7-263a100569cc.png)
![image](https://user-images.githubusercontent.com/51314259/89639606-8da0e200-d8b6-11ea-80bc-712d0efa7f27.png)
![image](https://user-images.githubusercontent.com/51314259/89640144-a3fb6d80-d8b7-11ea-9326-ffb7beef875e.png)

   
</details>



#### ✔️ Beta 
• Google Pixel 3a (Android 10)  - Device ID `d315ad72-cb08-4c0d-a3d1-ac0d2dc155c2`

<details>
    <summary>Screenshots (▼ dropdown )</summary>

![image](https://user-images.githubusercontent.com/51314259/89641349-27b65980-d8ba-11ea-937a-9dde9c4ebfe9.png)
[...]

PR #13312, opened 2020-08-05
v79.1.0 uplift - For #13229 - Cache deviceId for reuse

boek:


pocmo:
Change looks good. But I assume it needs to go to `releases/v80.0.0`? 🤔 

79.1 is on the beta branch, but will it also ship to release users?

liuche:
Since we're rolling 80.0.0 out very soon, we can close this.

Issue #13276, opened 2020-08-04
Wire up toolbar and awesomebar together in the new search experience


boek:
For #13052

Issue #13273, opened 2020-08-04
Add the awesomebar component to the new search experience


boek:
parent #13052

Issue #13272, opened 2020-08-04
Add the toolbar to the new search experience


boek:
Parent #13052

PR #13314, opened 2020-08-05
Progress on new search experience with Toolbar and Awesomebar placeholders

boek:
For #13272, #13273, #13276

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/search/awesomebar/AwesomeBarView.kt
+++ b/app/src/main/java/org/mozilla/fenix/search/awesomebar/AwesomeBarView.kt
@@ -5,7 +5,6 @@
 package org.mozilla.fenix.search.awesomebar
 
 import android.content.Context
-import android.view.ViewGroup
 import androidx.appcompat.content.res.AppCompatResources.getDrawable
 import androidx.core.graphics.BlendModeColorFilterCompat.createBlendModeColorFilterCompat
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/mozilla/fenix/search/awesomebar/AwesomeBarView.kt
```diff
@@ -91,7 +92,7 @@ class AwesomeBarView(
     init {
         view.itemAnimator = null
 
-        val context = container.context
+        val context = context
```
eliserichards, line 95: This can be removed

app/src/main/java/org/mozilla/fenix/search/awesomebar/AwesomeBarView.kt
```diff
@@ -255,7 +256,7 @@ class AwesomeBarView(
     }
 
     private fun isBrowsingModePrivate(): Boolean {
-        return (container.context.asActivity() as? HomeActivity)?.browsingModeManager?.mode?.isPrivate
+        return (context.asActivity() as? HomeActivity)?.browsingModeManager?.mode?.isPrivate
```
eliserichards, line 259: would `context.asActivity() as? HomeActivity)?` ever be null?

boek: I'm not sure :)

PR #13324, opened 2020-08-05
Wires up controller, store and interactor. 

boek:


app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarController.kt
```diff
@@ -81,6 +81,9 @@ class DefaultBrowserToolbarController(
     private val onCloseTab: (Session) -> Unit
 ) : BrowserToolbarController {
 
+    private val useNewSearchExperience
+        get() = activity.settings().useNewSearchExperience
```
NotWoods, line 85: We're moving to injecting dependencies in the constructor instead, `Settings` should be passed in via the constructor.

eliserichards: Let's put this into the next piece of the work @boek

Issue #11903, opened 2020-06-24
FNX-12904 ⁃ [Bug] White almost invisible "close" and "share" icons in custom tabs when Fenix is always set to open tabs in a Private Browsing mode


ghost:
## Steps to reproduce
Have the setting to open all links in Private Browsing enabled by going into 3-dot main menu > settings > Private Browsing > Open links in a private tab > enable 

I tested this with [Sync for Reddit](https://play.google.com/store/apps/details?id=com.laurencedawson.reddit_sync) and [Boost for Reddit](https://play.google.com/store/apps/details?id=com.rubenmayayo.reddit), when opening any link in custom tabs from them the icons for "close" (to close the tab) and "share" completely white such that they are almost invisible.  Please see the screenshots for clarity.

The issue doesn't reproduce if "Open links in a private tab" is not enabled

### Expected behavior
Icons shouldn't be white and invisible

### Actual behavior
Icons are almost invisible

### Device information
* Android device: **Motorola One Power, Android 10**.
* Fenix version: **`Nightly 200624 06:01`**.

![Screenshot](https://user-images.githubusercontent.com/42088907/85582961-62777180-b62d-11ea-942f-1452313992ad.jpg)UNITO-UNDERSCORE!20200624!UNITO-UNDERSCORE!204239!
![Screenshot](https://user-images.githubusercontent.com/42088907/85582988-66a38f00-b62d-11ea-96c5-d00a8859ca37.jpg)UNITO-UNDERSCORE!20200624!UNITO-UNDERSCORE!204157!

AndiAJ:
Hi, just checked this on the latest **Nightly Build 200625 from 6/25** using a Google Pixel 3a (Android 10)

✔️ The "Close" and Action (Share) button are fixed
❌ On the other hand, The TP, Site certificate, URL, and ⋮ Main menu are impropely displayed

►Screenshots
![rsz_screenshot_20200625-123347](https://user-images.githubusercontent.com/51314259/85696172-5cdb6380-b6e1-11ea-985d-e2653ae82da1.png)

@NotWoods - Pleasse review. ☺️

sflorean:
Tested on Nightly 6/30 and 78.0.1-beta.1 and here are my findings:

1. **With Fenix opened in the background in normal browsing**: I am still able to **reproduce** the issue, meaning that "close" and "share" buttons are almost invisible. 

[...]

PR #13327, opened 2020-08-05
For #11903: Fix private custom tabs

NotWoods:
### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/IntentReceiverActivity.kt
+++ b/app/src/main/java/org/mozilla/fenix/IntentReceiverActivity.kt
@@ -10,6 +10,7 @@ import android.os.Bundle
 import android.os.StrictMode
 import androidx.annotation.VisibleForTesting
 import mozilla.components.feature.intent.processing.IntentProcessor
+import org.mozilla.fenix.HomeActivity.Companion.PRIVATE_BROWSING_MODE
 import org.mozilla.fenix.components.IntentProcessorType
 import org.mozilla.fenix.components.getType
 import org.mozilla.fenix.components.metrics.Event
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/fenix/IntentReceiverActivity.kt
```diff
@@ -43,7 +44,15 @@ class IntentReceiverActivity : Activity() {
...
-        val processor = getIntentProcessors().firstOrNull { it.process(intent) }
+        val private = settings().openLinksInAPrivateTab
+        intent.putExtra(PRIVATE_BROWSING_MODE, private)
+        if (private) {
+            components.analytics.metrics.track(Event.OpenedLink(Event.OpenedLink.Mode.PRIVATE))
+        } else {
+            components.analytics.metrics.track(Event.OpenedLink(Event.OpenedLink.Mode.NORMAL))
+        }
+
+        val processor = getIntentProcessors(private).firstOrNull { it.process(intent) }
```
ekager, line 55: So AFAICT this `getIntentProcessors` will also process `externalAppIntentProcessors` and others. Before even though it wasn't working (lol) it was attempting to just add this extra on the custom tab and intent processors. Will there be any bad side effects from adding this extra to other intent entry points?

NotWoods: There shouldn't be side effects when we add extra intents. If we want to be super safe we can set the boolean on both `this.intent` (before) and the home activity intent (now).

ekager: Sorry that was a confusing comment. I meant as an example - now the TWA/PWA externalAppIntentProcessors will also be attaching `PRIVATE_BROWSING_MODE`, right? Does that mean PWAs will also be attempting to present / be in PBM?

NotWoods: Honestly we wanted it to work that way in the first place, so if they do present in PBM that fixes another bug :)

We can do another pass in the future to re-evaulate what data is getting passed in intents like this.

Issue #12862, opened 2020-07-23
FNX2-15510 ⁃ Migrate library menus to concept-menu


NotWoods:
Classes still using browser-menu:

* [x] BookmarkItemMenu
* [x] HistoryItemMenu

After converting to the new library, the menus should still work in the same way.

sflorean:
Verified as fixed on Nightly 8/19 with Samsung Galaxy Note 10 (Android 10) and Google Pixel 3 (Android 11).

PR #13332, opened 2020-08-06
FNX-14513 ⁃ For #12862: Use concept-menu in library

NotWoods:
### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkItemMenu.kt
+++ b/app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkItemMenu.kt
@@ -5,65 +5,89 @@
 package org.mozilla.fenix.library.bookmarks
 
 import android.content.Context
-import mozilla.components.browser.menu.BrowserMenuBuilder
-import mozilla.components.browser.menu.item.SimpleBrowserMenuItem
-import mozilla.components.concept.storage.BookmarkNode
+import androidx.annotation.VisibleForTesting
+import mozilla.components.browser.menu2.BrowserMenuController
+import mozilla.components.concept.menu.MenuController
+import mozilla.components.concept.menu.candidate.TextMenuCandidate
+import mozilla.components.concept.menu.candidate.TextStyle
 import mozilla.components.concept.storage.BookmarkNodeType
+import mozilla.components.support.ktx.android.content.getColorFromAttr
 import org.mozilla.fenix.R
-import org.mozilla.fenix.library.LibraryItemMenu
-import org.mozilla.fenix.theme.ThemeManager
 
 class BookmarkItemMenu(
     private val context: Context,
-    private val item: BookmarkNode,
-    private val onItemTapped: (BookmarkItemMenu.Item) -> Unit = {}
-) : LibraryItemMenu {
+    private val onItemTapped: (Item) -> Unit
+) {
 
-    sealed class Item {
-        object Edit : Item()
-        object Select : Item()
-        object Copy : Item()
-        object Share : Item()
-        object OpenInNewTab : Item()
-        object OpenInPrivateTab : Item()
-        object Delete : Item()
+    enum class Item {
+        Edit,
+        Copy,
+        Share,
+        OpenInNewTab,
+        OpenInPrivateTab,
+        Delete;
     }
 
-    override val menuBuilder by lazy { BrowserMenuBuilder(menuItems) }
+    val menuController: MenuController by lazy { BrowserMenuController() }
 
-    private val menuItems by lazy {
-        listOfNotNull(
-            if (item.type in listOf(BookmarkNodeType.ITEM, BookmarkNodeType.FOLDER)) {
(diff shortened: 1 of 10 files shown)


app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkItemMenu.kt
```diff
@@ -5,65 +5,64 @@
...
-                    onItemTapped.invoke(BookmarkItemMenu.Item.Copy)
-                }
+    @VisibleForTesting
+    internal fun menuItems(itemType: BookmarkNodeType): List<TextMenuCandidate> {
+        return listOfNotNull(
+            if (itemType != BookmarkNodeType.SEPARATOR) TextMenuCandidate(
+                text = context.getString(R.string.bookmark_menu_copy_button)
+            ) {
+                onItemTapped.invoke(Item.Copy)
             } else null,
```
person808, line 40: Kind of nitpicky but removing the brackets from the if expression here + returning a multi-line object and using trailing lambdas makes this a little hard to read. I'd prefer something like this.

```suggestion
            if (itemType != BookmarkNodeType.SEPARATOR) {
                TextMenuCandidate(
                    text = context.getString(R.string.bookmark_menu_copy_button)
                ) { onItemTapped.invoke(Item.Copy) }
            } else {
                null
            },
```

app/src/main/java/org/mozilla/fenix/library/bookmarks/viewholders/BookmarkFolderViewHolder.kt
```diff
@@ -39,7 +39,7 @@ class BookmarkFolderViewHolder(
         setSelectionListeners(item, selectionHolder)
 
         if (!item.inRoots()) {
-            setupMenu(item)
+            super.bind(item, mode)
```
person808, line 42: It feels a bit weird to me to conditionally call the super method here. I don't think I've seen that very often. It might be better to have an `updateMenu` method in the base viewholder class for this

person808: Especially if we wanted to start adding stuff to the super method that applies to all folders for whatever reason, this would have to be changed anyway. Basically I'd rather anyone making future changes to the super method not have to be aware that it may not always be called.

app/src/main/java/org/mozilla/fenix/library/bookmarks/viewholders/BookmarkSeparatorViewHolder.kt
```diff
@@ -25,6 +25,6 @@ class BookmarkSeparatorViewHolder(
     ) {
         this.item = item
         containerView.displayAs(LibrarySiteItemView.ItemType.SEPARATOR)
-        setupMenu(item)
+        super.bind(item, mode)
```
person808, line 28: If you want to keep the super method, I think this should be the first line of the function for consistency.

app/src/main/java/org/mozilla/fenix/library/bookmarks/viewholders/BookmarkItemViewHolder.kt
```diff
@@ -36,7 +36,7 @@ class BookmarkItemViewHolder(
         } else {
             containerView.overflowView.showAndEnable()
         }
-        setupMenu(item)
+        super.bind(item, mode)
```
person808, line 39: Same here

app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkItemMenu.kt
NotWoods: Agreed, I was just keeping the changes down by sticking with the existing style. I'll update it.

app/src/main/java/org/mozilla/fenix/library/bookmarks/viewholders/BookmarkNodeViewHolder.kt
```diff
@@ -5,38 +5,41 @@
...
+import org.mozilla.fenix.utils.Do
 
 /**
  * Base class for bookmark node view holders.
  */
 abstract class BookmarkNodeViewHolder(
-    override val containerView: LibrarySiteItemView,
+    protected val containerView: LibrarySiteItemView,
     private val interactor: BookmarkViewInteractor
-) : RecyclerView.ViewHolder(containerView), LayoutContainer {
```
person808, line 22: I think we do need to implement LayoutContainer here so the views are cached properly. We also seem to have our own ViewHolder class that does that for us.

NotWoods: In this case all the views are on `LibrarySiteItemView`, so `LayoutContainer` is redundant.

PR #13355, opened 2020-08-06
For #8043 Hook up AwesomeBarView.setOnEditSuggestionListener

jonalmeida:
Re-push of https://github.com/mozilla-mobile/fenix/pull/13346 with required AC version bump.

### Pull Request checklist

- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features. In addition, it includes a screenshot of a successful [accessibility scan](https://play.google.com/store/apps/details?id=com.google.android.apps.accessibility.auditor&hl=en_US) to ensure no new defects are added to the product.

### After merge
- [ ] **Milestone**: Make sure issues finished by this pull request are added to the [milestone](https://github.com/mozilla-mobile/fenix/milestones) of the version currently in development.

### To download an APK when reviewing a PR:
1. click on Show All Checks,
2. click Details next to "Taskcluster (pull_request)" after it appears and then finishes with a green checkmark,
3. click on the "Fenix - assemble" task, then click "Run Artifacts".
4. the APK links should be on the left side of the screen, named for each CPU architecture

PR #13394, opened 2020-08-07

MickeyMoz:


jonalmeida:
The tests here are failing consistently, I don't think restarting will help until we figure out what changed to cause the failure and how we should modify the test to be accurate.

jonalmeida:
Closing in favour of https://github.com/mozilla-mobile/fenix/pull/13355 where we update the AC version there along with the breaking tests that the update brings in.

PR #13417, opened 2020-08-08

MickeyMoz:


pocmo:
Unit test failures:

```
BrowsersCacheTest. cached list of browsers change before-after installation when cache is invalidated
BrowsersCacheTest. cached list of browsers match before-after installation when cache is not invalidated
```

PR #13453, opened 2020-08-10
FNX-14633 ⁃ For #162 - Add OSS build flavor.

### LINKED ISSUE #162: F-Droid
Would be nice to have this browser available in the F-Droid repository, when it'll be released (I hope soon). That's why I should ask, do you support this idea?

gilbsgilbs:
This PR adds an OSS build flavor allowing to build fenix without proprietary components, namely:

- play-services-ads-identifier
- play-services-oss-licenses (the plugin is FOSS, but not the library, so it was replaced with [AboutLibraries](https://github.com/mikepenz/AboutLibraries))
- lib-push-firebase (which isn't closed-source by itself yet bundles proprietary libraries such as firebase-core and play-services)
- leanplum (which isn't closed-source by itself but includes dependencies to firebase and gms)

I didn't do the work for other SDKs that are open-source but promote closed-source tracking services (such as Adjust), as it is not strictly necessary to release fenix on F-Droid as far as I know.

Build the oss / non-oss flavor:
```
./gradlew assembleOssDebug assembleNonOssDebug
```

Simple check for proprietary dependencies:
```console
$ dexdump app/build/outputs/apk/nonOss/debug/app-nonOss-arm64-v8a-debug.apk  | grep -E '(firebase|gms)' | wc -l
21121
$ dexdump app/build/outputs/apk/oss/debug/app-oss-arm64-v8a-debug.apk  | grep -E '(firebase|gms)' | wc -l      
0
```

So I guess it works.

I think this PR even closes #162 as the work of making fenix available on F-Droid will mostly be political after that.

### Pull Request checklist

- [x] **Tests**: This PR includes thorough tests or an explanation of why it does not - mostly consists in stubbing existing implementations
- [x] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not - no UI changes
- [x] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features. In addition, it includes a screenshot of a successful [accessibility scan](https://play.google.com/store/apps/details?id=com.google.android.apps.accessibility.auditor&hl=en_US) to ensure no new defects are added to the product. - no UI changes

### After merge

1. click on Show All Checks,

ping @pocmo


app/src/free/java/org/mozilla/fenix/components/metrics/LeanplumMetricsServiceImpl.kt
```diff
@@ -0,0 +1,21 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.components.metrics
+
+import android.app.Application
+import android.net.Uri
+import java.util.UUID
+
+class LeanplumMetricsServiceImpl(
```
NotWoods, line 11: LeanplumMetricsService is used for two cases, both of which already have interfaces. Rather than making a dummy metrics service, let's configure those two locations.

First one is easy, the service is used in MetricController.create and discarded if the Telemetry flag is set to false. No change needed.

Second location is for the DeepLinkIntentProcessor. It wants a DeepLinkVerifier instance. You could pass in a dummy version that always returns false based on build flags.

app/src/free/java/org/mozilla/fenix/settings/about/OssLicensesMenuImpl.kt
```diff
@@ -0,0 +1,15 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.settings.about
+
+import android.content.Context
+import android.content.Intent
+
+class OssLicensesMenuImpl : OssLicensesMenu {
+    override fun getIntent(context: Context?, title: String): Intent {
+        // This menu shouldn't displayed on the free variant anyways.
```
NotWoods, line 12: Why not? AFAIK open source libraries should be fine to leave in, and we'll need to credit them to use them in the app.

app/build.gradle
```diff
@@ -17,6 +17,8 @@ import org.mozilla.fenix.gradle.tasks.LintUnitTestRunner
 
 import static org.gradle.api.tasks.testing.TestResult.ResultType
 
+def useFreeImplementation = project.properties['org.mozilla.fenix.useFreeImplementation'].toBoolean()
```
NotWoods, line 20: This might be better off as another build variant, like debug.

app/src/free/java/org/mozilla/fenix/components/metrics/AdvertisingIDImpl.kt
```diff
@@ -0,0 +1,13 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.components.metrics
+
+import android.content.Context
+
+interface AdvertisingIDImpl : AdvertisingID {
```
NotWoods, line 9: Should this be a class instead of an interface?

app/src/main/java/org/mozilla/fenix/components/Push.kt
```diff
@@ -42,5 +42,5 @@ class Push(context: Context, crashReporter: CrashReporter) {
         PushConfig(projectId)
     }
 
-    private val pushService by lazy { FirebasePushService() }
+    private val pushService by lazy { PushServiceImpl() }
```
NotWoods, line 45: Can we leave the class name as-is for readability?

gilbsgilbs:
I've moved to using build flavors instead of a build flag and fixed tests for both flavors. As expected, taskcluster will have to be updated to run the OSS tests suite and produce the OSS apks. I'd appreciate some help as I don't know how to test this locally.

app/build.gradle
gilbsgilbs: So this is done. I've used build flavors which definitely makes more sense. Needing help to update taskcluster.

gilbsgilbs:
I've tried to blindly update the taskcluster config, but it'll be hard to go further without being able to test. Please tell me there's anything I can do to get this PR to the next step. Thanks!

(and thoughts for all the Mozilla employees who were laid off :cry:)

pocmo:
Yeah, challenging times. I'll get back to that PR. Thank you so far for all that work! 👍

app/src/free/java/org/mozilla/fenix/settings/about/OssLicensesMenuImpl.kt
NotWoods: I'm still a bit confused here. Why can't we use the Google library? As far as I can tell its open source and just figures out the play services libraries we use, rather than depending on them. 

If we need to change the license implementation I think it would be nice to have it split into a separate PR so its easier to review.

app/src/free/java/org/mozilla/fenix/components/metrics/LeanplumMetricsServiceImpl.kt
NotWoods: Sorry, I think I was getting confused with BuildConfig flags and build variants. I thought this could be accomplished with an if statement. 

Are we allowed to use any form of telemetry in the F-Droid version? If not, we may want to replace the entire MetricController. We have a dummy `DebugMetricController` already that could be used.

app/src/free/java/org/mozilla/fenix/settings/about/OssLicensesMenuImpl.kt
gilbsgilbs: Actually it **is** confusing. The gradle plugin that lists the dependencies is open-source and under Apache 2 license, but the library that is shipped with the APK is proprietary and, as far as I can tell, closed-source. Morever, the library has dependencies to gms (which you can easily see with dexdump) which is unsuitable for OSS anyways.

Agreed that it would make sense to split this into a separate PR though.

NotWoods: Got it, thanks for breaking it down. Using the most popular FLOSS version instead makes sense to me.

app/src/free/java/org/mozilla/fenix/components/metrics/LeanplumMetricsServiceImpl.kt
gilbsgilbs: > Are we allowed to use any form of telemetry in the F-Droid version?

I'm pretty sure that yes, telemetry is allowed as long as the frameworks being used are completely FLOSS (so Glean and Adjust should be OK, not firebase because of the dependency to gms). You can even display ads under the same conditions.

However, telemetry, especially when enabled by default like it is in Fenix, is considered an "anti-feature". This means there will be a warning prior downloading the app on the store. Such warning is already displayed for Fennec, so that's not too bad I guess.

app/src/free/java/org/mozilla/fenix/settings/about/OssLicensesMenuImpl.kt
gilbsgilbs: #13767

app/src/free/java/org/mozilla/fenix/components/metrics/AdvertisingIDImpl.kt
gilbsgilbs: Rolledback to an interface because otherwise detekt was complaining of an abstract class containing only abstract method (see [UnncessaryAbstractClass](https://arturbosch.github.io/detekt/style.html#unnecessaryabstractclass)). Tell me if you think it would be better off keeping a class and ignoring the warning using an annotation.

gilbsgilbs:
@pocmo thanks for your response. Although build flavors are generally the intended way to do this, I can totally understand your reasoning. I'll bring back build flags then.

pocmo:
Thank you! I am on PTO next week, so it may take a bit until can take another look.

gilbsgilbs:
Enjoy yourself :smile: .

gilbsgilbs:
I'll close this PR and make tinier, per-build flag PRs to make reviews easier.

Issue #13404, opened 2020-08-07
[Bug] Access denied when trying to view unit test failure results


jonalmeida:
When clicking on "Unit Test Results" in the Github UI to see the results AND in the Taskcluster artifacts, we see access denied for  many resources aside from the `index.html`.

```
<Error>
<Code>AccessDenied</Code>
<Message>Access Denied</Message>
<RequestId>97ED96C9E4A014DE</RequestId>
<HostId>
X1BVn71kINqfRMTF7aiiw7TmhCBbfjyNIvJvJOC2VbDLNLBlGAjfyUFNjVYfuSLumo/kZtc+99M=
</HostId>
</Error>
```

Example link: https://firefox-ci-tc.services.mozilla.com/tasks/dk6jXf6ST5OjXtr0XKkGzg#artifacts

jonalmeida:
cc @NotWoods for reference since I believe you had originally worked on this.

NotWoods:
It seems that the URL changed, I'll send a PR soon to update it

PR #13475, opened 2020-08-10
FNX-14655 ⁃ For #13404: Update Android lint URL

NotWoods:
Missed a spot


liuche:
These need to be rebased, master was broken for most of a day when a broken test was landed.

PR #13481, opened 2020-08-10
FNX-14661 ⁃ Delete drawables

NotWoods:
Mostly quick action bar assets


app/src/main/res/drawable/ic_link.xml
```diff
@@ -8,6 +8,6 @@
     android:viewportWidth="24"
     android:viewportHeight="24">
     <path
-        android:fillColor="?attr/accent"
+        android:fillColor="?primaryText"
```
ekager, line 11: Hm good find it looks like we are retinting this to primary everywhere we use it 🤔 we could probably also remove those tints now (might be slightlllyyy more efficient?)

Issue #13030, opened 2020-07-28
FNX2-15625 ⁃ Investigate swipe to switch tabs animation improvements


person808:
Was reading [this](https://blog.discord.com/how-discord-made-android-in-app-navigation-easier-fbd904d12521) recently which mentions the material design guidelines for animations. Currently the swipe to switch tabs animation is using physics based animations. Discord open sourced the components in the post so it would be good to take a look and see if we can improve the animations using ideas from their implementation.

PR #13489, opened 2020-08-10
For #13030 - Use material design animation values for swipe to switch…

person808:
… tabs.

Hopefully this feels a little more natural compared to the physics-based animations (with values I made up).
Values taken from https://material.io/design/motion/speed.html

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -429,7 +429,6 @@ dependencies {
     implementation Deps.androidx_lifecycle_viewmodel
     implementation Deps.androidx_core
     implementation Deps.androidx_core_ktx
-    implementation Deps.androidx_dynamic_animation
     implementation Deps.androidx_transition
     implementation Deps.androidx_work_ktx
     implementation Deps.google_material
--- a/app/src/main/java/org/mozilla/fenix/browser/ToolbarGestureHandler.kt
+++ b/app/src/main/java/org/mozilla/fenix/browser/ToolbarGestureHandler.kt
@@ -6,19 +6,19 @@ package org.mozilla.fenix.browser
 
 import android.animation.Animator
 import android.animation.AnimatorListenerAdapter
+import android.animation.ValueAnimator
 import android.app.Activity
 import android.graphics.PointF
 import android.graphics.Rect
-import android.util.TypedValue
 import android.view.View
 import android.view.ViewConfiguration
 import androidx.annotation.Dimension
 import androidx.annotation.Dimension.DP
+import androidx.core.animation.doOnEnd
 import androidx.core.graphics.contains
 import androidx.core.graphics.toPoint
 import androidx.core.view.isVisible
-import androidx.dynamicanimation.animation.DynamicAnimation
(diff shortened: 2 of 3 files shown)


person808:
@brampitoyo @Mugurell https://streamable.com/mu9a6g

Scenarios shown are:
1. Overscroll and release when there's no tab to swipe to.
2. Fling to next tab.
3. Slow scroll and release to switch to next tab.
4. Slow scroll and fling in opposite direction to stay on the same tab.

person808:
As for values, I'm using 250ms when we animate to a new tab and 200ms when staying on the same tab (a "canceled" gesture). These are the values recommended for medium sized animations. I chose medium because thats what Discord uses for their gestures which do similar things (they animate the whole screen to show a new panel).

brampitoyo:
@Mugurell Thanks for tagging me. I agree @person808, this looks and feels good!

I have one question around point 4:

> 4. Slow scroll and fling in opposite direction to stay on the same tab.

Compare these two animations:

- 0:18-0:20 – fling back from tab 2 (right) to tab 1 (left)
- 0:23-0:25 - fling back from tab 1 to tab 2

The first animation seems to be faster and ‘snappier’ than the second. Is this because of how you release the hold gesture?

Paging @topotropic just as an FYI.

person808:
> The first animation seems to be faster and ‘snappier’ than the second. Is this because of how you release the hold gesture?

I think it is because of where I release the hold gesture. Since the animation uses a fixed duration now, the first animation is faster because it has to animate a farther distance in the same amount of time. Maybe if we animate without switching tabs it might make sense to use an even shorter animation duration (like 150ms?). I did notice in my testing that flinging to "cancel" the gesture was the weirdest feeling one because you are likely to lift your finger closer to the edge we are animating towards, resulting in a shorter distance to animate.

brampitoyo:
@person808 wrote:

> Since the animation uses a fixed duration now, the first animation is faster because it has to animate a farther distance in the same amount of time.

Understood. Thanks heaps for the explanation! This makes sense to me.

> Maybe if we animate without switching tabs it might make sense to use an even shorter animation duration (like 150ms?).

Yes. Flinging is generally done in one go. People tend to swipe fast, not slowly or studiously. So if this fling is cancelled (ie. animating without switching tabs), we might want to shorten the duration.

if a 150ms value is easy enough to change, would you be willing to give it a quick test and report back as to how it feels?

person808:
@brampitoyo I tried out a couple of values. Both are an improvement in my opinion.

150ms - https://streamable.com/y1oooy

100ms - https://streamable.com/1xprl0

brampitoyo:
@person808 I thought that the 100ms value felt snappier – but it was also very snappy, at times it felt jarring.

What do you think about using the 150ms value as a compromise?

Great work on this. Whatever your decision end up being, it gets an r+ from me. Thanks heaps!

person808:
Update with 150ms value for flinging to cancel gesture.

Issue #12573, opened 2020-07-14
Add telemetry probe: was startup cold, warm, or hot?


mcomella:
We want to know how often the app opens in a cold, warm, or hot state. via https://jira.mozilla.com/browse/FXP-1071

sraturi:
QA: can you ensure the probe app_startup_all accurately represents what it says it should represent and edge cases are handled correctly.

Some test cases:

    Launch app, FenixApplication is created = COLD startup
    Launch app on external activity = COLD
    Launch app through icon ->press home button -> open external activity (creating a new activity) = WARM
    Launch app through icon -> press home button -> open a link = HOT

AndiAJ:
Hi I've just checked this matter on the latest Nightly Build 200908 using a Google Pixel 3a (Android 10)


• Open first time Fenix using: **`adb shell am start -n org.mozilla.fenix/mozilla.telemetry.glean.debug.GleanDebugActivity --ez logPings true --es tagPings aj-t08090`**

Ping **b1731157-230b-42b2-9735-7677b0cf6b84**
```
{
          "category": "events",
          "extra": {
            "has_saved_instance_state": "false",
            "source": "UNKNOWN",
            "type": "COLD"
          },
          "name": "app_opened_all_startup"

```

• While Fenix is in background - > Open a custom tab 
Ping **c5875daa-9c5d-42cb-94fa-dc28f5e97b84**
```
{
          "category": "events",
          "extra": {
            "has_saved_instance_state": "false",
            "source": "CUSTOM_TAB",
            "type": "WARM"
          },
          "name": "app_opened_all_startup"
```


• While Fenix is in background -> Open a custom tab -> Open the custom tab in Fenix (using the "Open in Firefox Nightly" from the custom tabs overlay )
Ping **67e65f06-0bb1-4e02-bc72-2af786f225aa**
```
"events": [
        {
          "category": "events",
          "extra": {
            "source": "UNKNOWN",
            "type": "HOT"
          },
          "name": "app_opened_all_startup"
```



• Open a custom tab -> Force close Fenix -> Open Fenix from the App icon
Ping **71db0d4b-506e-4fa0-ab7d-f57934efbb64**
```
{
          "category": "events",
          "extra": {
            "has_saved_instance_state": "false",
            "source": "APP_ICON",
            "type": "WARM"
          },
          "name": "app_opened_all_startup"

```

• Force close the custom tab -> While Fenix is in background tap the App icon
[...]

PR #13494, opened 2020-08-11
for #12573, added telemetry for startup type

sraturi:
### Pull Request checklist

### After merge

1. click on Show All Checks,


sraturi:
#### Tested scenarios
##### App starting fresh
1. Launch app, FenixApplication is created = COLD startup
2. Launch app on external activity = COLD
3. Launch app through icon ->press home button -> open external activity (creating a new activity) = WARM 
4.  Launch app through icon -> press home button -> open a link = HOT
##### App is already launched 
1. press home button (Homeactivity stopped)-> launch app = HOT 
2. press the back button (Homeactivity destroyed)-> launch app = WARM
3. press home button-> show all recent tasks -> pick fenix = HOT

app/src/main/java/org/mozilla/fenix/components/metrics/AppStartupTypeTelemetry.kt
```diff
@@ -0,0 +1,83 @@
...
+            recordAppStartupType(COLD_SAVED_INSTANCE)
+        } else if (applicationOnCreateCalled) {
+            recordAppStartupType(COLD)
+        } else if (hasSavedInstanceState) {
+            recordAppStartupType(WARM_SAVED_INSTANCE)
+        } else {
+            recordAppStartupType(WARM)
+        }
+
+        applicationOnCreateCalled = false
```
mcomella, line 56: This line of code feels misleading because of the variable name – if the application onCreate was called, how can it be uncalled? I'd suggest `wasApplicationOnCreateJustCalled` or even more specific, `wasApplicationOnCreateCalledBeforeHomeActivityOnCreate` (though it's getting a little lengthy - maybe you can shorten with `wasAppCreateCalledBeforeActivityCreate`?). With a name like that, it's very easy to understanding what a value represents and doesn't, validate that it's implemented correctly, and to read through the class quickly without second guessing your understanding.

app/src/main/java/org/mozilla/fenix/components/metrics/AppStartupTypeTelemetry.kt
```diff
@@ -0,0 +1,83 @@
...
+    fun onHomeActivityOnRestart() {
+        recordAppStartupType(HOT)
+    }
+
+    /**
+     * Records metric type if it hasn't been recorded already
+     * @param type the type of startup
+     */
+    private fun recordAppStartupType(type: Type) {
+        if (isMetricRecorded) {
```
mcomella, line 68: nit: To me, this reads, "this metric can only be recorded once."

I think a more specific name would help: `isMetricRecordedSinceAppWasForegrounded` or similar

app/src/main/java/org/mozilla/fenix/components/metrics/AppStartupTypeTelemetry.kt
```diff
@@ -0,0 +1,83 @@
...
+        metrics.track(Event.AppStartupType(type))
+        isMetricRecorded = true
+    }
+
+    @OnLifecycleEvent(Lifecycle.Event.ON_STOP)
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    fun onApplicationOnStop() {
+        // application was backgrounded, we need to record the new metric type if
+        // application was to come to foreground again.
+        // Therefore we set the isMetricRecorded flag to false.
```
mcomella, line 80: Good comment

app/src/main/java/org/mozilla/fenix/components/metrics/AppStartupTypeTelemetry.kt
```diff
@@ -0,0 +1,83 @@
...
+
+    init {
+        ProcessLifecycleOwner.get().lifecycle.addObserver(this)
+    }
+
+    // keeps track of whether we already recorded the metric
+    private var isMetricRecorded = false
+
+    // keeps track of whether the onCreate() was called for FenixApplication.
+    // helps choose which startup type to report during onHomeActivityOnCreate()
```
mcomella, line 34: nit: This comment could easily go out-of-date: what if it is no longer used to help choose a startup type or it's used for something else? For variables and methods, you can generally let the code do the talking unless it's difficult to come up with a name that accurately represents what's happening at a high-level. I'd remove this comment

Also note that names should generally be focused on what a value represents, not how it's intended to be used: someone can always add code to use your value in a new way and naming it after it's intended use both limits its re-use and is more likely to go out of date and become misleading. (this is a form of duplication: the name duplicates the code) You're doing the right thing with these variables but I just wanted to clarify because I was on the subject. :)

app/metrics.yaml
```diff
@@ -8,6 +8,26 @@ no_lint:
   - CATEGORY_GENERIC
```
mcomella, line 8: nit: can we squash the commits?

Also, the extended commit summary isn't super helpful:

>remove logs
>
>fix formatting
>
>fix formatting

Can we remove that too?

app/metrics.yaml
```diff
@@ -12,13 +12,21 @@ events:
     type: event
```
mcomella, line 12: nit: can we get issue numbers for the commits?

Also, git best practices is that a commit summary is 80 characters: that's why github does this wrapping:
image

That's pretty short though so I go over a bit if necessary – we could shorten the commit summary and use the extended summary if we wanted to add more detail.

app/src/test/java/org/mozilla/fenix/components/metrics/AppStartupTypeTelemetryTest.kt
```diff
@@ -0,0 +1,118 @@
...
+    private lateinit var appStartupTypeTelemetry: AppStartupTypeTelemetry
+
+    @Before
+    fun setup() {
+        MockKAnnotations.init(this)
+        appStartupTypeTelemetry = AppStartupTypeTelemetry(metricController)
+    }
+
+    @Test
+    fun `when FenixApplication and HomeActivity is created without savedInstance, then metric reports COLD startup type`() {
```
mcomella, line 32: nit: with `GIVEN WHEN THEN`, you should capitalize them to make it clear that you're following a convention, not just writing something coincidentally. The comma before then is also unnecessary with the convention.

app/src/test/java/org/mozilla/fenix/components/metrics/AppStartupTypeTelemetryTest.kt
```diff
@@ -0,0 +1,118 @@
...
+        appStartupTypeTelemetry.onHomeActivityOnCreate(true)
+
+        val validType = COLD_SAVED_INSTANCE
+
+        verify(exactly = 1) { metricController.track(Event.AppStartupType(validType)) }
+    }
+
+    @Test
+    fun `when application exists, and HomeActivity is created without savedInstanceState, then metric reports WARM startup type`() {
+        appStartupTypeTelemetry.onHomeActivityOnCreate(false)
```
mcomella, line 53: This code does not represent `WHEN application exists` – it represents "when this class is created and HomeActivity.onCreate` is called, which is never expected to happen. Let's fix the test to look at application being created (i.e. the name is fine)

When I'm writing test names (and thus designing the test at a high level), I'm careful to make sure my test name represents the code. When I am, I might happen to notice, in this case, 1) that my test isn't testing what I thought it was and needs to change and 2) there's a code path I didn't think of.

wrt 2, ideally we would throw if this unexpected code path occurs but it's probably not worth complicating the code to make that happen (because then we'd need to handle `onRestart`s impossible states too...). fwiw, if we wanted to solve this in a generic way, a state machine is a good way to do it because it doesn't require special-cased code to validate current state. But again, I don't think it's worth the complexity in this case.

app/src/test/java/org/mozilla/fenix/components/metrics/AppStartupTypeTelemetryTest.kt
```diff
@@ -0,0 +1,118 @@
...
+
+        val validType = WARM
+
+        verify(exactly = 1) { metricController.track(Event.AppStartupType(validType)) }
+    }
+
+    @Test
+    fun `when application exists, and HomeActivity is created with savedInstanceState, then metric reports WARM_SAVED_INSTANCE startup type`() {
```
mcomella, line 61: Same as above: let's fix the test

app/src/test/java/org/mozilla/fenix/components/metrics/AppStartupTypeTelemetryTest.kt
```diff
@@ -0,0 +1,118 @@
...
+        appStartupTypeTelemetry.onHomeActivityOnCreate(true)
+
+        val validType = WARM_SAVED_INSTANCE
+
+        verify(exactly = 1) { metricController.track(Event.AppStartupType(validType)) }
+    }
+
+    @Test
+    fun `when application and HomeActivity exists, then metric reports HOT startup type`() {
+        appStartupTypeTelemetry.onHomeActivityOnRestart()
```
mcomella, line 71: Same as above

app/src/test/java/org/mozilla/fenix/components/metrics/AppStartupTypeTelemetryTest.kt
```diff
@@ -0,0 +1,118 @@
...
+
+        val validType = HOT
+
+        verify(exactly = 1) { metricController.track(Event.AppStartupType(validType)) }
+    }
+
+    @Test
+    fun `when application comes to foreground and HomeActivity already exists, then metrics reports HOT startup type`() {
+        // application goes to background
+        appStartupTypeTelemetry.onApplicationOnStop()
```
mcomella, line 81: We're already in an impossible state: how can we call onStop without onCreate? Let's initialize the state of this object correctly (onCreate, onCreate then onStop). Here and below

Note also that this becomes a use-case for GIVEN (GIVEN the state WHEN onRestart...)

app/metrics.yaml
```diff
@@ -8,6 +8,34 @@ no_lint:
   - CATEGORY_GENERIC
 
 events:
+  app_startup_type:
+    type: event
+    description:
+      A user opens the application. we record what type of start
+      up it was. If HomeActivity has a savedInstanceState (a way for
```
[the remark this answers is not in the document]
sraturi: This makes sense. was wondering  

> app started AND HomeActivity created = impossible

is an arbitrary case or an actual case? although I also see:

> app started AND HomeActivity created = warm

sraturi:
@mcomella  I made all the changes to merge together the two metrics (app startup source and type). Let me know if this is what you were looking for? If not I can revert back to the previous commit.

app/metrics.yaml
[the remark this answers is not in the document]
mcomella: > app started AND HomeActivity created = impossible

I meant `app created AND HomeActivity started = impossible` – sorry about that

app/src/main/java/org/mozilla/fenix/components/metrics/AppStartupTelemetry.kt
```diff
@@ -0,0 +1,109 @@
...
+        setStartupSourceFromIntent(safeIntent, false)
+    }
+
+    fun onExternalAppBrowserOnCreate(safeIntent: SafeIntent, hasSavedInstanceState: Boolean) {
+        appOpenedAllSourceStartup.hasSavedInstance = hasSavedInstanceState
+        setAppStartupType()
+        setStartupSourceFromIntent(safeIntent, true)
+    }
+
+    fun onHomeActivityOnRestart() {
```
mcomella, line 54: ^ Related to the comment above – it's confusing how the source gets set from this method.

app/src/main/java/org/mozilla/fenix/components/metrics/AppStartupTelemetry.kt
```diff
@@ -0,0 +1,109 @@
...
+        appOpenedAllSourceStartup.type = Type.HOT
+        appOpenedAllSourceStartup.hasSavedInstance = false
+    }
+
+    fun receivedIntentInHomeActivity(safeIntent: SafeIntent) {
```
mcomella, line 59: ^ Related to the comment above above – how do the other vars get set? I need to run the code myself and relearn which methods are called when to validate the data is set correctly. Is it possible to write your code so that it's obvious to me what your assumptions are? That I can validate the code works without even running it, based on your assumptions (perhaps they need to be documented in comments)?

app/src/main/java/org/mozilla/fenix/components/metrics/AppStartupTelemetry.kt
```diff
@@ -0,0 +1,109 @@
...
+        setStartupSourceFromIntent(safeIntent, false)
+    }
+
+    private fun setAppStartupType() {
+        appOpenedAllSourceStartup.type = if (wasAppCreateCalledBeforeActivityCreate) COLD else WARM
+        wasAppCreateCalledBeforeActivityCreate = false
```
mcomella, line 65: nit; Ideally this is called directly in `onCreate` or similar; otherwise someone else could call this because it seems convenient to call and it might break our boolean flag (note: boolean flags are also fragile (like this) and thus a red flag; one way around them are state machines but they add a lot of complexity and usually aren't worth it for simple cases).

app/src/main/java/org/mozilla/fenix/components/metrics/AppStartupTelemetry.kt
```diff
@@ -0,0 +1,109 @@
...
+    fun onHomeActivityOnResume() {
+        recordMetric()
+    }
+
+    private fun recordMetric() {
+        if (!isMetricRecordedSinceAppWasForegrounded) {
+            metrics.track(appOpenedAllSourceStartup)
+            isMetricRecordedSinceAppWasForegrounded = true
+        }
+        // we don't want any weird previous states to persist on our next metric record.
```
mcomella, line 96: Good idea – you can invalidate some of what I said above but I'm not going to rewrite it because I already typed too much 😅

note: If we switch to an immutable data container, we'd probably assign to the "no data" state here instead.

app/src/main/java/org/mozilla/fenix/components/metrics/AppStartupTelemetry.kt
```diff
@@ -0,0 +1,109 @@
...
+        appOpenedAllSourceStartup = AppStartupSourceAndType()
+    }
+
+    @OnLifecycleEvent(Lifecycle.Event.ON_STOP)
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    fun onApplicationOnStop() {
+
```
mcomella, line 103: nit: no need for whitespace

app/src/test/java/org/mozilla/fenix/components/metrics/AppStartupTelemetryTest.kt
```diff
@@ -0,0 +1,163 @@
...
+    @Before
+    fun setup() {
+        MockKAnnotations.init(this)
+        appStartupTelemetry = AppStartupTelemetry(metricController)
+    }
+
+    @Test
+    fun `WHEN application is launch for the first time through application icon THEN startup metric records {APP_ICON, COLD, FALSE}`() {
+        every { intent.action } returns Intent.ACTION_MAIN
+        every { intent.categories.contains(Intent.CATEGORY_LAUNCHER) } returns true
```
mcomella, line 38: it'd be less fragile to have categories return a set with `CATEGORY_LAUNCHER` rather than mock `categories`

app/src/test/java/org/mozilla/fenix/components/metrics/AppStartupTelemetryTest.kt
```diff
@@ -0,0 +1,163 @@
...
+import org.mozilla.fenix.components.metrics.Event.AppStartupSourceAndType.Source.UNKNOWN
+import org.mozilla.fenix.components.metrics.Event.AppStartupSourceAndType.Type.COLD
+import org.mozilla.fenix.components.metrics.Event.AppStartupSourceAndType.Type.WARM
+import org.mozilla.fenix.components.metrics.Event.AppStartupSourceAndType.Type.HOT
+
+class AppStartupTelemetryTest {
+
+    @RelaxedMockK
+    private lateinit var metricController: MetricController
+    @RelaxedMockK
```
mcomella, line 24: nit: if you can easily avoid relaxing a mock, you should try. Non-relaxed mocks help you ensure you're mocking all the methods that are called and 1) are testing all the possible code paths and 2) aren't writing in states that would never happen in the code (where unit is returned instead of a real value).

Relaxing is best used e.g. when an object is passed in as an argument and it might be queried but it's values do not impact the final result (which does assume some knowledge about the implementation, to be fair, and that's an anti-pattern).

app/src/test/java/org/mozilla/fenix/components/metrics/AppStartupTelemetryTest.kt
```diff
@@ -0,0 +1,163 @@
...
+    private lateinit var appStartupTelemetry: AppStartupTelemetry
+
+    @Before
+    fun setup() {
+        MockKAnnotations.init(this)
+        appStartupTelemetry = AppStartupTelemetry(metricController)
+    }
+
+    @Test
+    fun `WHEN application is launch for the first time through application icon THEN startup metric records {APP_ICON, COLD, FALSE}`() {
```
mcomella, line 36: Rather than saying `{APP_ICON, COLD, FALSE}`, which could go out of date if the implementation changes, you should probably use prose: `cold homescreen launch with no instance state`. That way if we refactor to send two probes, you could still potentially use the test without changing the name.

Honestly, you could also just say, "THEN records the correct values" because the front half of the test name is redundant to the back half. I would probably do this one.

mcomella, line 36: nit: you're missing the instance state on the left side, here and below

app/src/test/java/org/mozilla/fenix/components/metrics/AppStartupTelemetryTest.kt
```diff
@@ -0,0 +1,163 @@
...
+        appStartupTelemetry.onFenixApplicationOnCreate()
+        appStartupTelemetry.onHomeActivityOnCreate(intent.toSafeIntent(), false)
+        appStartupTelemetry.onHomeActivityOnResume()
+
+        val validMetric = AppStartupSourceAndType(LINK, COLD, false)
+        verify(exactly = 1) { metricController.track(validMetric) }
+    }
+
+    @Test
+    fun `WHEN application is launch for the first time through an external app THEN startup metric records {CUSTOM_TAB, COLD, FALSE}`() {
```
mcomella, line 61: nit: I think "an external app" is ambiguous – maybe custom tab?

app/src/test/java/org/mozilla/fenix/components/metrics/AppStartupTelemetryTest.kt
```diff
@@ -0,0 +1,163 @@
...
+
+        val validMetric = AppStartupSourceAndType(CUSTOM_TAB, COLD, false)
+        verify(exactly = 1) { metricController.track(validMetric) }
+    }
+
+    @Test
+    fun `GIVEN, application  exists and is backgrounded, WHEN application is launched again through app icon and HomeActivity is recreated THEN startup metric records {APP_ICON, WARM, FALSE}`() {
+        every { intent.action } returns Intent.ACTION_MAIN
+        every { intent.categories.contains(Intent.CATEGORY_LAUNCHER) } returns true
+        appStartupTelemetry.onApplicationOnStop()
```
mcomella, line 74: I think you need to do `onCreate; onCreate; onResume; onStop` (or at least `onCreate; onStop` if HomeActivity isn't launched) to get the test set up correctly, right? Otherwise, can we be sure about the state of the object assuming we don't know about the implementation?

This "application is backgrounded" is also a good candidate for a helper function.

app/src/test/java/org/mozilla/fenix/components/metrics/AppStartupTelemetryTest.kt
sraturi: done!

app/src/test/java/org/mozilla/fenix/components/metrics/AppStartupTelemetryTest.kt
sraturi: done.

app/metrics.yaml
```diff
@@ -18,26 +18,36 @@ events:
...
       between the time application goes into background and the time
       android reports the application going into the background.
-      Note: This metric does not cover the following cases:
-      Case # 1 -> a). open a link(for example, gmail) with in-app
-      Browser (metric report custom_tab startup) b). press home button
-      c). open gmail again (which brings us back to in app browser).
-      Step c will not report startup metric. Case # 2 -> a). open fenix
-      b). press home button c). launch fenix through app switcher/recent
-      apps. step c will not report startup type.
+      Note: This metric has an unknown case:
```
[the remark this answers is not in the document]
sraturi: done!

app/src/main/java/org/mozilla/fenix/components/metrics/AppStartupTelemetry.kt
```diff
@@ -0,0 +1,109 @@
...
+ * [AppStartupSourceAndType]and finally report them on Activity's onResume() function.
+ * If application wasn't backgrounded before onResume() is called, we don't report the metric.
+ */
+class AppStartupTelemetry(private val metrics: MetricController) : LifecycleObserver {
+
+    init {
+        ProcessLifecycleOwner.get().lifecycle.addObserver(this)
+    }
+
+    private var appOpenedAllSourceStartup = AppStartupSourceAndType()
```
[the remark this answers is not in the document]
sraturi: done!

app/src/main/java/org/mozilla/fenix/components/metrics/AppStartupTelemetry.kt
```diff
@@ -0,0 +1,109 @@
...
+
+    private var isMetricRecordedSinceAppWasForegrounded = false
+    private var wasAppCreateCalledBeforeActivityCreate = false
+
+    fun onFenixApplicationOnCreate() {
+        wasAppCreateCalledBeforeActivityCreate = true
+    }
+
+    fun onHomeActivityOnCreate(safeIntent: SafeIntent, hasSavedInstanceState: Boolean) {
+        appOpenedAllSourceStartup.hasSavedInstance = hasSavedInstanceState
```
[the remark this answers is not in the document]
sraturi: > send a probe saying "error happened" rather than transmit the bad data because we have no way of knowing that we forgot to set one of the vars.

For this reason, I had set up the initial state for all metric fields to be UNKNOWN, so to avoid sending wrong data (random states) if we forget to set a field and set the metric state to UNKNOWN after every time we record the metric.

I changed all the metric fields to be null by default(when init the metric object) for now and added a boolean check to make sure we are never sending a null state. I guess my question for now is, do we crash if we enter the impossible(theoretically) state where a metric is being recorded but some fields are null?

For example, 


    private fun recordMetric() {
        if (!isMetricRecordedSinceAppWasForegrounded && metricDataContainer.isAllDataCollected()) {
            metrics.track(metricDataContainer)
            isMetricRecordedSinceAppWasForegrounded = true
        } else if (!isMetricRecordedSinceAppWasForegrounded && !metricDataContainer.isAllDataCollected()) {
            // we should never be in this state
            //todo do we log it or crash?
            Log.e(Companion.LOG_TAG, "App startup telemetry not sent. Data: $metricDataContainer")
        }
        // we don't want any weird previous states to persist on our next metric record.
        metricDataContainer = AppStartupSourceAndType()
    }

PS: I definitely appreciate all the feedbacks (nits or otherwise) so keep em coming! 😄

app/src/main/java/org/mozilla/fenix/components/metrics/AppStartupTelemetry.kt
sraturi: added a  comment

app/src/main/java/org/mozilla/fenix/components/metrics/AppStartupTelemetry.kt
sraturi: Done! initially, I was trying to remove code duplicate but this makes sense.

app/src/test/java/org/mozilla/fenix/components/metrics/AppStartupTelemetryTest.kt
sraturi: I tried something like that but was getting weird mock error for some reason... ill give it another shot

app/src/test/java/org/mozilla/fenix/components/metrics/AppStartupTelemetryTest.kt
sraturi: sorry not sure what you mean by instance on the left side?

app/src/main/java/org/mozilla/fenix/components/metrics/AppAllSourceStartTelemetry.kt
```diff
@@ -1,59 +0,0 @@
-/* This Source Code Form is subject to the terms of the Mozilla Public
- * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-
-package org.mozilla.fenix.components.metrics
```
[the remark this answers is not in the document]
sraturi: Ah sweet, I didn't realize git move existed.. thanks.

In this case, I didn't really rename anything. I started with a fresh file and wrote everything new, it just looks similar because I used the same concept I guess 😅

app/metrics.yaml
sraturi: sure!

app/metrics.yaml
```diff
@@ -18,18 +18,29 @@ events:
       startups, not just cold startup. Note: There is a short gap
```
mcomella, line 18: nit: I think you need to build and then app `metrics.md` to this PR.

app/src/main/java/org/mozilla/fenix/components/metrics/Event.kt
```diff
@@ -319,11 +319,20 @@ sealed class Event {
...
+    ) : Event() {
+        enum class Source { APP_ICON, LINK, CUSTOM_TAB, ERROR }
+        enum class Type { COLD, WARM, HOT, ERROR }
 
         override val extras: Map<Events.appOpenedAllStartupKeys, String>?
-            get() = hashMapOf(Events.appOpenedAllStartupKeys.source to source.name)
+            get() = hashMapOf(
+                Events.appOpenedAllStartupKeys.source to source.toString(),
+                Events.appOpenedAllStartupKeys.startupType to type.toString(),
+                Events.appOpenedAllStartupKeys.startupType to hasSavedInstance.toString()
```
mcomella, line 334: This is the wrong key

Side notes:
- if glean had extras type safety, it would have caught this
- a test for the conversion function would have caught this
- you may not have done this but there's a (tbh, hard to believe) phenomenon that happens where if you copy, paste, and edit the paste, you're more likely to make an error in the last replacement of your copy paste: it's an "off-by-one" problem :P It's a reason to 1) write code by hand instead of copy-pasting when feasible or 2) ensuring code where copy-and-pasting is unnecessary. However, I don't see a clear way to do that here.

app/src/main/java/org/mozilla/fenix/components/metrics/Event.kt
```diff
@@ -319,11 +319,20 @@ sealed class Event {
             get() = hashMapOf(Events.appOpenedAllStartupKeys.source to source.name)
     }
 
-    data class AppOpenedAllSourceStartup(val source: Source) : Event() {
-        enum class Source { APP_ICON, LINK, CUSTOM_TAB, UNKNOWN }
+    data class AppStartupSourceAndType(
```
mcomella, line 322: nit: I think we can just call it `AppStartup` or similar (whatever name that doesn't clash with the cold startup probe – maybe we should rename that one). This convention is not really sustainable – what happens when we add timing values to it too? Why isn't instanceState represented?

app/metrics.yaml
```diff
@@ -18,18 +18,29 @@ events:
...
-      Case # 1 -> a). open a link(for example, gmail) with in-app
-      Browser (metric report custom_tab startup) b). press home button
-      c). open gmail again (which brings us back to in app browser).
-      Step c will not report startup metric. Case # 2 -> a). open fenix
-      b). press home button c). launch fenix through app switcher/recent
-      apps. step c will not report startup type.
+      Note: This metric does not record souce when app opened from
+      task switcher: open application -> press home button -> open
+      recent tasks -> choose fenix. In this case will report
+      [source = unknown, type = hot, has_saved_instance_state = false].
```
mcomella, line 24: Do we use `unknown` and not `app_switcher` because we don't know if there are other ways this startup can happen?

app/src/main/java/org/mozilla/fenix/components/metrics/AppStartupTelemetry.kt
```diff
@@ -0,0 +1,148 @@
...
+
+import org.mozilla.fenix.components.metrics.Event.AppStartupSourceAndType.Type
+import org.mozilla.fenix.components.metrics.Event.AppStartupSourceAndType.Type.COLD
+import org.mozilla.fenix.components.metrics.Event.AppStartupSourceAndType.Type.WARM
+
+/**
+ * Tracks application startup source, type, and whether or not activity has savedInstance to restore
+ * the activity from. Sample metric = [source = COLD, type = APP_ICON, hasSavedInstance = false].
+ * The basic idea is to collect these metrics from different phases of startup through
+ * [AppStartupSourceAndType]and finally report them on Activity's onResume() function.
```
mcomella, line 29: nit: space before and

app/src/main/java/org/mozilla/fenix/components/metrics/AppStartupTelemetry.kt
```diff
@@ -0,0 +1,148 @@
...
+ * If application wasn't backgrounded before onResume() is called, we don't report the metric.
```
mcomella, line 30: The comment duplicates the code a bit (e.g. does someone using this class care that it's reported from onResume? What happens if we change that, e.g. for duration measurement?). We also have docs to explain the metrics extras & options.

app/src/main/java/org/mozilla/fenix/components/metrics/AppStartupTelemetry.kt
```diff
@@ -0,0 +1,148 @@
...
+     * https://developer.android.com/reference/android/app/Activity#onNewIntent(android.content.Intent)
+     *
+     */
+    fun onHomeActivityOnResume() {
+        recordMetric()
+    }
+
+    private fun recordMetric() {
+        if (!isMetricRecordedSinceAppWasForegrounded) {
+            metrics.track(metricDataContainer.getAppStartupSourceAndType())
```
mcomella, line 109: This implementation has similar problems to the previous implementation:
- it's not explicit how the data from `onRestart` and `onNewIntent` are joined
- we don't handle, and document, the app_switcher case
- we don't handle, and document, the error case too explicitly

For this method, I'm essentially thinking:
```kotlin
if (!isMetricRecordedSinceAppWasForegrounded) {
    if (onResumeData != null && onNewIntentData != null) {
        // create immutable & track
    } else if ( // handle other cases such as app switcher
```

It explicitly documents our assumption (that `onNewIntent` and `onResume` should be called together) and explicitly documents each of the possible cases in a central location.

app/src/main/java/org/mozilla/fenix/components/metrics/Event.kt
sraturi: Ah Idk if I can blame this mistake on that specific phenomenon 😓  but I have definitely come across it while working with xml files

app/src/main/java/org/mozilla/fenix/components/metrics/Event.kt
sraturi: done

app/src/main/java/org/mozilla/fenix/components/metrics/AppStartupTelemetry.kt
sraturi: fun getAppStartupSourceAndType(): AppAllStartup {
            setNullFieldsToError()
            return AppAllStartup(source!!, type!!, hasSavedInstance!!)
        }

is essentially where I'm doing that. I think we decided to send `ERROR` if a field is not set? 

`setNullFieldsToError()` does seem to be hidden away, and maybe we need to call this inside `recordMetric()`
so something like 

```
if (!isMetricRecordedSinceAppWasForegrounded) {
            setNullFieldsToError()
           track(event....)
}

```
Doing callback specific `if`  statement seems a bit weird since the same data is recorded in multiple states (eg `onCreate` records all the data)

app/metrics.yaml
sraturi: Yup, the problem is we don't have a new intent to get the information from.

app/src/main/java/org/mozilla/fenix/components/metrics/AppStartupTelemetry.kt
sraturi: done removed the comment

sraturi:

All questions are mandatory. You must receive a review from a data steward peer on your responses to these questions before shipping new data collection.

**1.  What questions will you answer with this data?**
    

What type of startup user uses most? For example, cold, warm, hot.

What type of startup source users uses most? For example, app_icon, link, custom_tab

**2.  Why does Mozilla need to answer these questions? Are there benefits for users? Do we need this information to address product or business requirements? Some example responses:**
    

Creating an accurate understanding of real-world usage to prioritize performance improvements  
  

**3.  What alternative methods did you consider to answer these questions? Why were they not sufficient?**
    

There is no existing metrics that answer this question.  
  

**4.  Can current instrumentation answer these questions?**
    

No, we are modifying the existing probe to add additional startup data.  
  

**5.  List all proposed measurements and indicate the category of data collection for each measurement, using the [Firefox data collection categories](https://wiki.mozilla.org/Firefox/Data_Collection) found on the Mozilla wiki.**
    

data is Interaction data. What type of startup user use most.

Options

-   cold
    
-   Cold (with saved_instance state)
    
-   warm
    
-   warm, (with saved_instance state)
    
-   hot.  
      
    

**6.  Please provide a link to the documentation for this data collection which describes the ultimate data set in a public, complete, and accurate way.**
    

This is documented in the metrics.md file

**7.  How long will this data be collected?**
    

Until 2021-02-01  
  

**8.  What populations will you measure?**
    

All release, beta, and nightly users with telemetry enabled.  
  

**9.  If this data collection is default on, what is the opt-out mechanism for users?**
    

default on, opt out by Settings ->Data collection

**10.  Please provide a general description of how you will analyze this data.**
    

we can set up better tests and prioritize analyses to improve performance in the most impactful areas.

**11.  Where do you intend to share the results of your analysis?**
    

Glean, Amplitude, and mobile teams.  
  

**12.  Is there a third-party tool (i.e. not Telemetry) that you are proposing to use for this data collection? If so:**
    

N/A

app/metrics.yaml
```diff
@@ -18,28 +18,42 @@ events:
...
+      unknown is for tasks switcher or other unanticipated startup source.
```
mcomella, line 25: nit: -> "unknown is for startup sources where we can't pinpoint the cause. One UNKNOWN case is the app switcher where we don't know what variables to check to ensure this startup wasn't caused by something else"

Or something similar but shorter.

We should move this to the "source" description field rather than explain it in the top level.

app/src/main/java/org/mozilla/fenix/components/metrics/AppStartupTelemetry.kt
```diff
@@ -0,0 +1,159 @@
...
+            getStartupSourceFromIntent(safeIntent, true),
+            getAppStartupType(),
+            hasSavedInstanceState
+        )
+        wasAppCreateCalledBeforeActivityCreate = false
+    }
+
+    /**
+     * we are not setting [Source] in this method since source is derived from a [SafeIntent].
+     * therefore source either gets set in oncreate() or onNewIntent().
```
mcomella, line 71: This is false: we're only looking for a source from `onNewIntent`, right?

Also, I'd put this comment in the function body  for two reasons:
- it's closer to the code it affects so if the implementation changes, the comment is more likely to be updated
- kdoc is generally reserved for information relevant to callers, not implementation details

app/src/main/java/org/mozilla/fenix/components/metrics/AppStartupTelemetry.kt
```diff
@@ -0,0 +1,159 @@
...
+     */
+    fun onHomeActivityOnRestart() {
+        onRestartData = Pair(HOT, false)
+    }
+
+    /**
+     * we are only setting [Source] in this method since source is derived from a [SafeIntent].
+     * other metric fields are set in onCreate() and onRestart()
```
mcomella, line 79: This is also false and we should move it to the function body.

tbh, I'm also fine not having this comment because:
- the metrics tracking code makes that assumption somewhat clear
- it adds complexity to this function: now the dev starts thinking about a broader context than, "lifecycle function is called and we cache some data about the call" when I feel they can that complexity when they get to the usage of that data.

but I could go either way

app/src/main/java/org/mozilla/fenix/components/metrics/AppStartupTelemetry.kt
```diff
@@ -0,0 +1,159 @@
...
+        )
+        wasAppCreateCalledBeforeActivityCreate = false
+    }
+
+    /**
+     * we are not setting [Source] in this method since source is derived from a [SafeIntent].
+     * therefore source either gets set in oncreate() or onNewIntent().
+     */
+    fun onHomeActivityOnRestart() {
+        onRestartData = Pair(HOT, false)
```
mcomella, line 74: I just noticed: since it's always `false`, can't we just omit this extra in the telemetry? (e.g. a nullable field in the metric class that, when null, does not add the extra)

app/src/main/java/org/mozilla/fenix/components/metrics/AppStartupTelemetry.kt
```diff
@@ -0,0 +1,159 @@
...
+
+    private fun getStartupSourceFromIntent(
+        intent: SafeIntent,
+        isExternalAppBrowserActivity: Boolean
+    ): Source {
+        return when {
+            isExternalAppBrowserActivity -> CUSTOM_TAB
+            intent.isLauncherIntent -> APP_ICON
+            intent.action == Intent.ACTION_VIEW -> LINK
+            else -> UNKNOWN
```
mcomella, line 97: nit; Can we add a comment above here that this includes the app switcher case?

app/src/main/java/org/mozilla/fenix/components/metrics/AppStartupTelemetry.kt
```diff
@@ -0,0 +1,159 @@
...
+        }
+    }
+
+    /**
+     * The reason we record metric on resume is because we need to wait for onNewIntent(), and
+     * we are not guaranteed that onNewIntent() will be called before or after onStart() / onRestart().
+     * However we are guaranteed onResume() will be called after onNewIntent() and  onStart(). Source:
+     * https://developer.android.com/reference/android/app/Activity#onNewIntent(android.content.Intent)
+     *
```
mcomella, line 106: nit: unnecessary blank line

app/src/main/java/org/mozilla/fenix/components/metrics/AppStartupTelemetry.kt
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    private fun mergeOnRestartAndOnNewIntentIntoStartup(): AppAllStartup {
+        return AppAllStartup(
+            onNewIntentData!!,
+            onRestartData!!.first,
+            onRestartData!!.second
+        )
+    }
+    private fun getErrorStartup(): AppAllStartup {
```
mcomella, line 137: nit: newline above

app/src/main/java/org/mozilla/fenix/components/metrics/AppStartupTelemetry.kt
```diff
@@ -0,0 +1,159 @@
...
+     *
+     */
+    fun onHomeActivityOnResume() {
+        recordMetric()
+    }
+
+    private fun recordMetric() {
+        if (!isMetricRecordedSinceAppWasForegrounded) {
+            val appAllStartup: AppAllStartup = if (onCreateData != null) {
+                    onCreateData!!
```
mcomella, line 115: nit: too much indentation and inconsistent

app/metrics.yaml
```diff
@@ -18,28 +18,42 @@ events:
...
+      Error is for impossible cases.
```
mcomella, line 26: nit: we should add, "Please file a bug if you see ERROR."

app/src/main/java/org/mozilla/fenix/components/metrics/AppStartupTelemetry.kt
```diff
@@ -0,0 +1,159 @@
...
+
+    private fun mergeOnRestartAndOnNewIntentIntoStartup(): AppAllStartup {
+        return AppAllStartup(
+            onNewIntentData!!,
+            onRestartData!!.first,
+            onRestartData!!.second
+        )
+    }
+    private fun getErrorStartup(): AppAllStartup {
+        return if (onNewIntentData == null && onRestartData == null) {
```
mcomella, line 138: Based on our understanding of the lifecycle, what are the impossible cases?

| onNewIntent | onRestart | Type | Source |
| - | - | - | - |
| data | data | HOT | from Intent |
| null | data | HOT | UNKNOWN (possibly app switcher) |
| data | null | ERROR | from Intent |
| null | null | ERROR | UNKNOWN |

The error cases occur because we never the lifecycle methods to get called in that order for it to happen, right?
- In the 3rd row, this case is we didn't create the activity (onCreate) and we were backgrounded but `onRestart` wasn't called: we don't expect that to ever happen, right?
- In the 4th row, this case is we didn't create the activity and we were backgrounded but neither `onRestart` or `onNewIntent` was called: we don't expect that to happen either, right?

Generally, the assumption is `onRestart` should always be called if we're backgrounded and `onCreate` wasn't called.

Do you agree?

**I believe the code is wrong: can you update it to fit this model?** Note how I didn't duplicate error/unknown through type/source too.

Generating tables like these is helpful for ensuring you think through all possible states.

app/src/main/java/org/mozilla/fenix/components/metrics/AppStartupTelemetry.kt
```diff
@@ -0,0 +1,153 @@
...
+        wasAppCreateCalledBeforeActivityCreate = false
+    }
+
+    fun onExternalAppBrowserOnCreate(safeIntent: SafeIntent, hasSavedInstanceState: Boolean) {
+        onCreateData = AppAllStartup(
+            getStartupSourceFromIntent(safeIntent, true),
+            getAppStartupType(),
+            hasSavedInstanceState
+        )
+        wasAppCreateCalledBeforeActivityCreate = false
```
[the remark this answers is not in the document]
mcomella: Please consider doing this.

app/metrics.yaml
```diff
@@ -18,28 +18,42 @@ events:
...
     extra_keys:
       source:
         description: |
           The method used to open Fenix. Possible values are `app_icon`,
-          `custom_tab`, `link` or `unknown`
+          `custom_tab`, `link`, `unknown` or `error`.
+      startup_type:
```
mcomella, line 32: nit: why not call this `type`?

mcomella, line 32: nit: can we move this above `source`? I'd consider `type` to be the primary key using my intuition, particularly if we're storing the error state there. It's important for folks to see that ERROR state as early as possible.

app/src/main/java/org/mozilla/fenix/components/metrics/AppStartupTelemetry.kt
sraturi: When its a cold or warm startup , we are getting the source from onCreate (using the launching intent). Whenever onCreate is called, onNewIntent() is not called.

app/src/main/java/org/mozilla/fenix/components/metrics/AppStartupTelemetry.kt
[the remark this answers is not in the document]
sraturi: done!

app/src/main/java/org/mozilla/fenix/components/metrics/AppStartupTelemetry.kt
sraturi: not sure what you mean by omit extra in telemetry?

mcomella: Instead of recording [HOT, APP_ICON, false], record [HOT, APP_ICON]

app/src/main/java/org/mozilla/fenix/components/metrics/AppStartupTelemetry.kt
```diff
@@ -0,0 +1,147 @@
...
+            isMetricRecordedSinceAppWasForegrounded = true
+        }
+        // we don't want any weird previous states to persist on our next metric record.
+        onCreateData = null
+        onNewIntentData = null
+        onRestartData = null
+    }
+
+    private fun mergeOnRestartAndOnNewIntentIntoStartup(): AppAllStartup {
+        return if (onRestartData != null && onNewIntentData != null) {
```
mcomella, line 128: I think you can replace this if/else block with elvis operators:
```kotlin
return AppAllStartup(onNewIntentData ?: UNKNOWN, onRestartData.first ?: ERROR, onRestartData.second ?: false)
```

That being said, you'll need to adjust for having `hasSavedInstance` being an optional value if you follow my suggestion above.

app/src/main/java/org/mozilla/fenix/components/metrics/AppStartupTelemetry.kt
mcomella: Be sure to document the fact that it's optional

app/src/main/java/org/mozilla/fenix/components/metrics/AppStartupTelemetry.kt
sraturi: ah keep forgetting about the elvis operator! thanks!

app/src/test/java/org/mozilla/fenix/components/metrics/AppStartupTelemetryTest.kt
```diff
@@ -0,0 +1,241 @@
...
+        appStartupTelemetry.onHomeActivityOnResume()
+
+        val validMetric = AppAllStartup(CUSTOM_TAB, COLD, false)
+        verify(exactly = 1) { metricController.track(validMetric) }
+    }
+
+    @Test
+    fun `GIVEN, application  exists and is backgrounded, WHEN application is launched again through app icon and HomeActivity is recreated THEN records the correct values`() {
+        setupIntentMock(APP_ICON)
+        launchApplicationAndPutApplicationInBackground()
```
mcomella, line 83: Yes, exactly!

app/src/test/java/org/mozilla/fenix/components/metrics/AppStartupTelemetryTest.kt
```diff
@@ -0,0 +1,241 @@
...
+        verify(exactly = 1) { metricController.track(validMetric) }
+    }
+
+    private fun launchApplicationAndPutApplicationInBackground() {
+        appStartupTelemetry.onFenixApplicationOnCreate()
+        appStartupTelemetry.onHomeActivityOnCreate(intent.toSafeIntent(), false)
+        appStartupTelemetry.onHomeActivityOnResume()
+
+        // have to clear the mock function calls so it doesnt interfere with tests
+        clearMocks(metricController, answers = false)
```
mcomella, line 183: What does answers = false do?

app/src/test/java/org/mozilla/fenix/components/metrics/AppStartupTelemetryTest.kt
```diff
@@ -0,0 +1,241 @@
...
+
+        appStartupTelemetry.onHomeActivityOnCreate(intent.toSafeIntent(), false)
+        appStartupTelemetry.onHomeActivityOnResume()
+
+        val validMetric = AppAllStartup(APP_ICON, WARM, false)
+        verify(exactly = 1) { metricController.track(validMetric) }
+    }
+
+    @Test
+    fun `GIVEN, application  exists and is backgrounded, WHEN application is launched again through url link and HomeActivity is recreated THEN records the correct values`() {
```
mcomella, line 93: No action needed

Note: I notice you're writing a test for many combinations: 3 startup types foreground, 3 startup types background, etc.

I probably would have written all these tests 😅 but it probably is a perfect is the enemy of the good situation where it's probably not necessary: doing all the combinations in one type of test and hitting all the other types of tests is probably good enough.

That being said, junit supports a feature called parameterized tests which would make writing all these tests combinations much less code but I haven't ever been able to set it up.

mcomella:
@sraturi btw, there is also a UI test failure now – I assume it's intermittent given what we're changing.

app/src/test/java/org/mozilla/fenix/components/metrics/AppStartupTelemetryTest.kt
sraturi: I couldn't really find something official and clear but it's consistent so I went on to use it. basically, it is clearing up calls to metricController.
For example, 

```
metricController.track(something)
clearMocks(metricController, answers = false) 
metricController.track(anotherThing)

verify(exactly = 1) { metricController.track(any()) } // this will be true
```

mcomella: Do you need `answers = false` or does `clearMocks(metricController)` work? I would prefer we do the latter if it does and are not sure what the arguments do.

sraturi: we need the `answers` part. otherwise, we need to re-init the mock every time.

app/src/test/java/org/mozilla/fenix/components/metrics/AppStartupTelemetryTest.kt
```diff
@@ -0,0 +1,266 @@
...
+
+        appStartupTelemetry.onHomeActivityOnNewIntent(intent.toSafeIntent())
+        appStartupTelemetry.onHomeActivityOnResume()
+
+        val validMetric = AppAllStartup(APP_ICON, ERROR)
+        verify(exactly = 1) { metricController.track(validMetric) }
+    }
+
+    @Test
+    fun `WHEN AppAllStartup does not have savedInstanceState THEN do not return savedInstanceState`() {
```
mcomella, line 228: These `AppAllStartup` tests should technically be in the corresponding test file for where `AppAllStartup` is defined.

However, let's avoid another back and forth: let's land it as is and if you have time and you're feeling inspired afterwards, you can move it. :) Thanks for writing tests for these cases.

app/metrics.yaml
```diff
@@ -18,23 +18,44 @@ events:
...
-      Case # 1 -> a). open a link(for example, gmail) with in-app
-      Browser (metric report custom_tab startup) b). press home button
-      c). open gmail again (which brings us back to in app browser).
-      Step c will not report startup metric. Case # 2 -> a). open fenix
-      b). press home button c). launch fenix through app switcher/recent
-      apps. step c will not report startup type.
+      Note: This metric does not record souce when app opened from
+      task switcher: open application -> press home button -> open
+      recent tasks -> choose fenix. In this case will report
+      [source = unknown, type = hot, has_saved_instance_state = false].
```
mcomella, line 24: In a follow-up PR (so we can get this landed), can we document two additional things:
1. that custom tab never sees COLD start and briefly explain why?
1. that if you click "Open in Firefox Preview" from a custom tab, a second event is not recorded

sraturi: I think there are places where custom tab might have a cold start. One case would be me: launch gmail (fenix in background), close fenix from recent tasks, and launch a link from gmail.

app/src/test/java/org/mozilla/fenix/components/metrics/AppStartupTelemetryTest.kt
sraturi: The events have tons of data classes so would it make sense to create a single Events test file and put all the tests for all the classes there? or make a folder called event, and put all the individual files there? 

I think with a single file, `before/after `might get confusing? unless we use Junit5? which has nested support.

PR #13502, opened 2020-08-11
Closes #13219: Fixing long-tap delay for UI tests

### LINKED ISSUE #13219: FNX2-18858 ⁃ Intermittent UI test failure - verifyAboutFirefoxPreview
### Firebase Test Run:
Flaky test:
https://console.firebase.google.com/project/moz-fenix/testlab/histories/bh.66b7091e15d53d45/matrices/6112134536254409084/executions/bs.f75165a672162639/videos

### Stacktrace:
`java.lang.AssertionError
at org.junit.Assert.fail(Assert.java:86)
at org.junit.Assert.assertTrue(Assert.java:41)
at org.junit.Assert.assertNotNull(Assert.java:712)
at org.junit.Assert.assertNotNull(Assert.java:722)
at org.mozilla.fenix.helpers.ext.WaitNotNullKt.waitNotNull(WaitNotNull.kt:20)
at org.mozilla.fenix.ui.robots.BrowserRobot$Transition.openTabDrawer(BrowserRobot.kt:380)
at org.mozilla.fenix.ui.robots.SettingsSubMenuAboutRobotKt.navigateBackToAboutPage(SettingsSubMenuAboutRobot.kt:68)`

Probably we should add the touch & hold delay increase to all tests that use the tab drawer:
https://github.com/mozilla-mobile/fenix/blob/master/app/src/androidTest/java/org/mozilla/fenix/ui/ThreeDotMenuMainTest.kt#L41

sv-ohorvath:
Changes the long-tap delay before starting the activity so it applies to all tests.
Fixed the Libraries menu toolbar resource ID because it has changed.

### DIFF
--- a/app/src/androidTest/java/org/mozilla/fenix/helpers/HomeActivityTestRule.kt
+++ b/app/src/androidTest/java/org/mozilla/fenix/helpers/HomeActivityTestRule.kt
@@ -5,7 +5,9 @@
 package org.mozilla.fenix.helpers
 
 import androidx.test.espresso.intent.rule.IntentsTestRule
+import androidx.test.platform.app.InstrumentationRegistry
 import androidx.test.rule.ActivityTestRule
+import androidx.test.uiautomator.UiDevice
 import org.mozilla.fenix.HomeActivity
 
 /**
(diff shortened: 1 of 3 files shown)


sv-ohorvath:
@AaronMT @isabelrios Do you think I should add this to all test classes to be safe?

sv-ohorvath:
> We should move this into a spot where this setting can be inherited into tests like into the ActivityTestRule. E.g, can use the apply method and write an override in https://github.com/mozilla-mobile/fenix/blob/master/app/src/androidTest/java/org/mozilla/fenix/helpers/HomeActivityTestRule.kt e.g, https://github.com/timrijckaert/BetterActivityTestRule/blob/61ad0802859218d67cbb4907df35f78bd05b3220/app/src/androidTest/java/be/rijckaert/tim/disableanimations/BetterActivityTestRule.kt#L36 – doesn't have to be as convoluted but see where they use executeShellCommand and where the functions are called there.

Done, please review: Adding this to the apply method would stop the automatic launch of the activity for reasons I couldn't explain. So, I used a different method from ActivityTestRule, beforeActivityLaunched() which according to the documentation does what we want: "Override this method to execute any code that should run before your {@link Activity} is created and launched. This method is called before each test method, including any method annotated with Before".

Issue #13424, opened 2020-08-09
FNX-14605 ⁃ Detekt errors on releases/v80.0.0 and master


sysrqb:
Running `./gradlew detekt` there are two separate issues.

1) `The original exception message was: Value "trued" set for config parameter "complexity > LongParameterList > active" is not of required type Boolean.`

This error exists on `master`, too, and it is fixed by
```
diff --git a/config/detekt.yml b/config/detekt.yml
index 56b6041c2..2568a929e 100644
--- a/config/detekt.yml
+++ b/config/detekt.yml
@@ -76,7 +76,7 @@ complexity:
     # https://github.com/mozilla-mobile/fenix/issues/4861
     threshold: 75
   LongParameterList:
-    active: trued
+    active: true
     excludes: "**/*Controller.kt, **/*Integration.kt"
     functionThreshold: 6
     constructorThreshold: 7
```

2) After correcting (1), detekt complains about:
```
complexity - 2h debt
        ComplexMethod - 15/15 - [start] at /home/androidstudio/fenix/app/src/main/java/org/mozilla/fenix/components/metrics/LeanplumMetricsService.kt:96:18
        LongMethod - 78/75 - [toEvent] at /home/androidstudio/fenix/app/src/main/java/org/mozilla/fenix/components/metrics/Metrics.kt:535:18
        LargeClass - 200/200 - [AwesomeBarView] at /home/androidstudio/fenix/app/src/main/java/org/mozilla/fenix/search/awesomebar/AwesomeBarView.kt:36:7
        LargeClass - 226/200 - [ToolbarGestureHandler] at /home/androidstudio/fenix/app/src/main/java/org/mozilla/fenix/browser/ToolbarGestureHandler.kt:38:7
        TooManyFunctions - 11/11 - [ToolbarGestureHandler] at /home/androidstudio/fenix/app/src/main/java/org/mozilla/fenix/browser/ToolbarGestureHandler.kt:38:7
        ComplexCondition - 4/4 - [onSwipeStarted] at /home/androidstudio/fenix/app/src/main/java/org/mozilla/fenix/browser/ToolbarGestureHandler.kt:80:13
style - 15min debt
        MaxLineLength - [onLoadRequest] at /home/androidstudio/fenix/app/src/main/java/org/mozilla/fenix/AppRequestInterceptor.kt:31:17
[...]

PR #13510, opened 2020-08-11
For #13424 - Fix detekt config and warnings.

person808:
### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/geckoRelease/java/org/mozilla/fenix/engine/GeckoProvider.kt
+++ b/app/src/geckoRelease/java/org/mozilla/fenix/engine/GeckoProvider.kt
@@ -5,6 +5,7 @@
 import android.content.Context
 import android.os.Bundle
 import mozilla.components.browser.engine.gecko.autofill.GeckoLoginDelegateWrapper
+import mozilla.components.browser.engine.gecko.ext.toContentBlockingSetting
 import mozilla.components.browser.engine.gecko.glean.GeckoAdapter
 import mozilla.components.concept.engine.EngineSession.TrackingProtectionPolicy
 import mozilla.components.concept.storage.LoginsStorage
@@ -50,8 +51,7 @@ object GeckoProvider {
         val runtimeSettings = builder
             .crashHandler(CrashHandlerService::class.java)
             .telemetryDelegate(GeckoAdapter())
-            // TODO: Fix me!
-            // .contentBlocking(policy.toContentBlockingSetting())
+            .contentBlocking(policy.toContentBlockingSetting())
             .aboutConfigEnabled(Config.channel.isBeta)
             .debugLogging(Config.channel.isDebug)
             .build()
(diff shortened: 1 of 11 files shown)


app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarView.kt
```diff
@@ -250,6 +250,7 @@ class BrowserToolbarView(
         private const val TOOLBAR_ELEVATION = 16
     }
 
+    @Suppress("ComplexCondition")
     private fun ToolbarMenu.Item.performHapticIfNeeded(view: View) {
         if (this is ToolbarMenu.Item.Reload && this.bypassCache ||
```
person808, line 255: This warning goes away if this is turned into a when instead of an if, but imo its not really anymore readable

app/src/geckoRelease/java/org/mozilla/fenix/engine/GeckoProvider.kt
```diff
@@ -47,6 +47,7 @@ object GeckoProvider {
         // Use meeee.
         policy.hashCode()
 
+        @Suppress("ForbiddenComment")
```
person808, line 50: The TODO comment this is for is removed in the beta and nightly versions. Maybe it should just be removed here?

jonalmeida: Hmm.. the git history seems to have been lost for this file when we simplified our build variants. 

@liuche this API was made so that we didn't have to comment out the line and when the AC version is upgraded the underlying changes would be applied. Something seems to have changed here that I can't see just yet.

jonalmeida: I understand what happened now. During the variant simplification, this PR went up before merge day was complete, so this API didn't work before that.

My guess is that we committed this but forgot to uncomment this code line out.

@person808 I think this would be the right fix: remove the TODO and uncomment the `contentBlocking` line.

cc: @Amejia481

person808: Done. Was this supposed to be updated too?

https://github.com/mozilla-mobile/android-components/blob/c3d7281db2a6f227679b1f907f629f16a1ca7aaf/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/ext/TrackingProtectionPolicy.kt#L24

jonalmeida: That's the underlying API which will be fixed during the next merge day. I believe we're using geckoview beta in Fenix releases so it should work fine after the last merge day.

Issue #13479, opened 2020-08-10
FNX-14659 ⁃ [Bug] HomeActivity doOnPreDraw Runnable memory leak


cadeyrn:
/cc @mcomella

### Steps to reproduce

Unfortunately no STR.

### Expected behavior

No leak.

### Actual behavior

Leak pattern: instance field android.os.Message#next

┬───
│ GC Root: System class
│
├─ android.app.ActivityThread class
│    Leaking: NO (MessageQueue↓ is not leaking and a class is never leaking)
│    ↓ static ActivityThread.sMainThreadHandler
├─ android.app.ActivityThread$H instance
│    Leaking: NO (MessageQueue↓ is not leaking)
│    ↓ ActivityThread$H.mQueue
├─ android.os.MessageQueue instance
│    Leaking: NO (MessageQueue#mQuitting is false)
│    ↓ MessageQueue.mMessages
│                   ~~~~~~~~~
├─ android.os.Message instance
│    Leaking: UNKNOWN
│    ↓ Message.next
│              ~~~~
├─ android.os.Message instance
│    Leaking: UNKNOWN
│    ↓ Message.next
│              ~~~~
├─ android.os.Message instance
│    Leaking: UNKNOWN
│    ↓ Message.next
│              ~~~~
├─ android.os.Message instance
│    Leaking: UNKNOWN
│    ↓ Message.next
│              ~~~~
├─ android.os.Message instance
│    Leaking: UNKNOWN
│    ↓ Message.next
│              ~~~~
├─ android.os.Message instance
│    Leaking: UNKNOWN
│    ↓ Message.next
│              ~~~~
├─ android.os.Message instance
│    Leaking: UNKNOWN
│    ↓ Message.next
│              ~~~~
├─ android.os.Message instance
│    Leaking: UNKNOWN
│    ↓ Message.next
│              ~~~~
├─ android.os.Message instance
│    Leaking: UNKNOWN
│    ↓ Message.next
│              ~~~~
├─ android.os.Message instance
│    Leaking: UNKNOWN
│    ↓ Message.next
│              ~~~~
├─ android.os.Message instance
│    Leaking: UNKNOWN
│    ↓ Message.next
│              ~~~~
├─ android.os.Message instance
│    Leaking: UNKNOWN
│    ↓ Message.next
│              ~~~~
├─ android.os.Message instance
│    Leaking: UNKNOWN
│    ↓ Message.next
│              ~~~~
├─ android.os.Message instance
[...]

PR #13887, opened 2020-08-17
 for #13479, added a VisualCompletenessQueue.kt class to handle all the functionality related to visual completeness

sraturi:


app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -164,10 +165,12 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
         // Must be after we set the content view
         if (isVisuallyComplete) {
             rootContainer.doOnPreDraw {
+                val homeActivityWeakRef = WeakReference<HomeActivity>(this)
+
                 // This delay is temporary. We are delaying 5 seconds until the performance
                 // team can locate the real point of visual completeness.
                 it.postDelayed({
```
mcomella, line 172: I'm concerned that:
- lambdas (or any anonymous class, for that matter) always capture `this`
- someone will just add something to this call that creates a new memory leak (e.g. logging a member variable)

Can we encapsulate this code such that this memory leak cannot happen easily (e.g. make it non-anonymous)?

---

fwiw, I have felt it weird that we're doing this low-level stuff in a lifecycle method (one of those red flags I mentioned) – it feels like this whole bit should be in at least a helper function, if not an entirely separately file.

sraturi: has to add a static function inside `HomeActivity` since `RunWhenReadyQueue` class is read only.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -164,13 +165,7 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
...
         // Must be after we set the content view
         if (isVisuallyComplete) {
-            rootContainer.doOnPreDraw {
-                // This delay is temporary. We are delaying 5 seconds until the performance
-                // team can locate the real point of visual completeness.
-                it.postDelayed({
-                    visualCompletenessQueue!!.ready()
-                }, delay)
-            }
+            setupRootContainerPreDrawDelay(rootContainer, WeakReference(this))
```
mcomella, line 168: nit: this name describes the implementation detail rather than what we're trying to achieve. That makes it redundant to the code and generally not helpful when scanning through `onCreate()`. "Huh, what's the root container? What pre draw delay? Does it delay drawing the root container? Why?"

Rather, if this was called `attachVisualCompletenessQueue`, its intention is clearer and the way we attach the queue (the implementation) can change without needing to rename the method.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -723,5 +718,23 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
         // PWA must have been used within last 30 days to be considered "recently used" for the
         // telemetry purposes.
         const val PWA_RECENTLY_USED_THRESHOLD = DateUtils.DAY_IN_MILLIS * 30L
+
+        /**
+         * set a delay until we locate visual completeness queue.
```
mcomella, line 723: nit: locate -> run

We're just running the queue after the delay. That being said, when this method is renamed, this is explaining implementation details that are subject to change: I think this line can be changed in favor of the high-level goal we're trying to accomplish.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -723,5 +718,23 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
...
+         * static function to avoid memory leak in case we are trying to access visualCompletenessQueue
+         * or other activity members when activity is already destroy.
+         */
+        private fun setupRootContainerPreDrawDelay(
```
mcomella, line 727: Not that it needs to change but why did you decide to put this in `HomeActivity` rather than a new file or a file related to the visual completeness queue?

If it makes a difference, I think the `RunWhenReadyQueue` is named in a way that makes it prematurely generic (e.g. what else do we use it for?) so I would support renaming it to the `VisualCompletenessQueue` if that helps (and it'd be easier to find in code search too).

Keeping things simple makes them easier to change and prematurely making things generic, meaning you have to support use cases that you're not actively using them for yet, makes things harder to change. Most times the prematurely generic thing is never repurposed or the requirements for the generic object change so it needs to be refactored anyway.

---

Also note: I don't usually like putting static functions in `companion` blocks of a class that has member functions. It's a usually code smell that some bit of functionality unrelated to the current class is getting piggy-backed on the current class and can be better named somewhere else (e.g. in this case, it's doing visualCompleteness related work – it's attaching the visual completness queue ready up to respond to the UI state).

This is a problem because:
- It can make the functionality harder to find
- It increases the size of the class, making it harder to skim to identify what the class is about
- It often violates the single-responsibility principle for the class
- It makes the `companion object` (unnecessarily) longer

I think what you wrote is probably the only time where I I feel less strongly about it – it could live here but it needs isolation from the capturing.

But in general, Activities are bloated classes and it's good to get as much code out of them, and into better encapsulations, as possible. e.g. HomeActivity is 740 lines, many of which are low-level things, when its primary responsibility is passing high-level lifecycle events (onResume, onStop, etc.). It seems better to include it in a 70 line low-level visual completeness queue file than here. (Note: fenix doesn't often do this but I think they should)

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -723,5 +718,23 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
...
+            rootContainer: View,
+            homeActivityWeakRef: WeakReference<HomeActivity>
+        ) {
+            rootContainer.doOnPreDraw {
+                // This delay is temporary. We are delaying 5 seconds until the performance
+                // team can locate the real point of visual completeness.
+                it.postDelayed({
+                    homeActivityWeakRef.get()?.visualCompletenessQueue!!.ready()
```
mcomella, line 735: Why do we need a WeakReference to HomeActivity? Could we just pass in the visualCompletenessQUeue instead? I noticed it's already part of `Components`: `components.performance.visualCompletenessQueue`. So it shouldn't be dependent on the Activity.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
mcomella: Sorry, I missed this comment before my review. It's defined in android-components – maybe we could wrap it or just define it in a new file?

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
mcomella: > Keeping things simple makes them easier to change and prematurely making things generic, meaning you have to support use cases that you're not actively using them for yet, makes things harder to change. Most times the prematurely generic thing is never repurposed or the requirements for the generic object change so it needs to be refactored anyway.

I looked through the history and it looks like this was refactored so it could be re-used (i.e. it's not prematurely generic). I wonder what the best way to merge our needs with the class defined in a-c is.

sraturi: @mcomella  Would adding a class inside  `PerformanceComponent` file to deal with this make sense? I was a bit hesitant to just randomly create a file in the components section. I was thinking maybe the classes can co-exist in the same place. If not, I can add another file.

app/src/main/java/org/mozilla/fenix/perf/VisualCompletenessQueue.kt
```diff
@@ -0,0 +1,26 @@
...
+ */
+class VisualCompletenessQueue(var queue: RunWhenReadyQueue) {
+
+    fun runVisualCompletenessQueue(container: View, delay: Long) {
+
+        container.doOnPreDraw {
+            // This delay is temporary. We are delaying 5 seconds until the performance
+            // team can locate the real point of visual completeness.
+            it.postDelayed({
+                queue.ready()
```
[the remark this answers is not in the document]
mcomella: yes

Issue #12287, opened 2020-07-05
FNX2-15143 ⁃ Show Synced Tabs after Normal And Private Tabs


s-ankur:
### What is the user problem or growth opportunity you want to see solved?
Syced tabs are such a useful feature but right now it is difficult to use. Adding synced tabs to the tab tray as a third option will make it more comfortable to use and also switch between synced and non synced tabs 

### How do you know that this problem exists today? Why is this important?
Synced tabs are nice but difficult to use feature.

### Who will benefit from it?
Everyone who uses synced tabs

SS1113:
Yes, such feature is already in consideration according one of the developers https://github.com/mozilla-mobile/fenix/issues/11995#issuecomment-650444399

s-ankur:
Nice! Would love to see it implemented

cadeyrn:
@jonalmeida The screenshots in #13893 shows that the synced tabs are below the regular tabs. Could these be moved to a third tab (regular tabs, private tabs, synced tabs) instead? I have hundreds of open tabs on other devices and most times I don't want to access these from my smartphone. But sometimes I want to access these so it's not an option to disable the synchronisation of tabs. Moving the synced tabs to a own tab would be great because my tabs view would keep clean and the only sometimes needed synced tabs would still be easy accessible.

jonalmeida:
@cadeyrn that's great feedback! I wonder what @topotropic thinks about it.

madb1lly:
Hi all,

Nice to to see synced tabs in the tab tray, it makes up for the lack of the Fennec awesomescreen/page whatever it was called. _However_ please can you make it so that individual devices can be expanded and collapsed like in #13449?

I second @cadeyrn's suggestion too, this will be much more pleasing. Was it Opera that someone showed this from a few months ago? Much nicer.

Cheers 🙂

yoasif:
Mentioned previously in https://github.com/mozilla-mobile/fenix/issues/7913 - would definitely like to see this - the current UI leaves much to be desired. I actually missed it after turning it on initially in secret settings.

yoasif:
> _However_ please can you make it so that individual devices can be expanded and collapsed like in #13449?

I think that is https://github.com/mozilla-mobile/fenix/issues/11402

amkcpu:
I did a quick mock-up to try to visualize what this could look like (using the screenshot generously provided in #13893):
- A sync can be forced by tapping the "sync" floating action button 
[...]

PR #13893, opened 2020-08-18
For #12287: Add Synced Tabs to Tabs Tray 

jonalmeida:
The feature is put behind a feature flag from the Secret Settings. Might be easier to review per commit.

Abstract [design link](https://share.goabstract.com/0bae23a7-b55b-43f9-a57a-1fdb7b3596fc?mode=build&sha=a813facef071a6c43e3d40d20afa4aaf4f75e269) to verify against.

out

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/res/layout/view_synced_tabs_title.xml
+++ b/app/src/main/res/layout/view_synced_tabs_title.xml
@@ -1,0 +1,29 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    android:orientation="horizontal">
+
+    <TextView
+        style="@style/Header16TextStyle"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_marginStart="16dp"
+        android:layout_marginTop="60dp"
+        android:layout_weight="1"
+        android:text="@string/synced_tabs" />
+
+    <ImageView
+        android:id="@+id/refresh_icon"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_marginTop="60dp"
+        android:layout_marginEnd="16dp"
+        app:srcCompat="@drawable/mozac_ic_refresh"
+        app:tint="?primaryText" />
+
+</LinearLayout>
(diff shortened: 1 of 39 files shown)


jonalmeida:
The parts I'm most unsure of are the navigation graph additions.

app/src/main/res/layout/view_synced_tabs_title.xml
```diff
@@ -0,0 +1,29 @@
...
+        style="@style/Header16TextStyle"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_marginStart="16dp"
+        android:layout_marginTop="60dp"
+        android:layout_weight="1"
+        android:text="@string/synced_tabs" />
+
+    <ImageView
+        android:id="@+id/refresh_icon"
```
ekager, line 21: nit - let's give this a `selectableItemBackground` so it visually clicks

ekager: Same with the synced tabs items - they should have a `selectableItemBackground` 
`

jonalmeida: > Same with the synced tabs items - they should have a `selectableItemBackground`

I already see that the synced tab items have the ripple animation when long pressed. Are you referring to something else?

ekager: Okay I see the ripple it's just verrrryyy faint on dark mode. Also the padding/margin is uneven when you click it - let's fix that then 😅

jonalmeida: Yeah! I actually noticed and fixed that because of this comment and then wasn't sure if _that_ was what you were referring to. 😁

ekager: Brain waves 🧠 〰️

jonalmeida:
Thanks for the review; super helpful!

>  1. Do we want this to show in the private tab in the tabs tray?

We don't see it in private tabs, I'll leave it up to @topotropic to comment (also consider the separate tab idea that Sorin mentioned).

>  2. We probably want to hide this in multiselect mode like we do for the save to collection button because otherwise it looks a bit strange?

Fixed! Check out the last commit that addresses the review comments.

>  3. When opening the tabs tray with synced tabs enabled and 1 open tab, it seems to jump to the synced tabs section and opens at an awkward point in the list.
>  4. Sometimes (can't repro 100%) it even opens further down on the synced tabs list and I would have to scroll _way_ up to see open tabs. When I swipe up to try to scroll, it closes. Got a video from after it opened.

These two are related and are a side-effect to how the tabs tray scrolls to the selected tab. The hacky approach that we currently have (calculate the index based on the adapter size), does not consider that the adapter is loaded async and we need to scroll when all the adapters are complete.

I've filed https://github.com/mozilla-mobile/fenix/issues/13905 to address this and added it as a blocker for removing the feature flag.

> Moving this into it's own Tabs Tray Tab (phew mouthful) would likely solve all these problems, so if we decide to move that way (via this comment [#12287 (comment)](https://github.com/mozilla-mobile/fenix/issues/12287#issuecomment-675285951) ) we can probably disregard these issues.

Yes! I'd really like us to look into this but I don't think it'll block us from landing this change.

PR #13932, opened 2020-08-19
 For #12565: Clean up controllers

NotWoods:
Additionally fix up tests

### DIFF
--- a/app/src/test/java/org/mozilla/fenix/share/ShareControllerTest.kt
+++ b/app/src/test/java/org/mozilla/fenix/share/ShareControllerTest.kt
@@ -9,12 +9,13 @@ import android.content.Context
 import android.content.Intent
 import androidx.navigation.NavController
 import com.google.android.material.snackbar.Snackbar
+import io.mockk.MockKAnnotations
 import io.mockk.Runs
 import io.mockk.every
+import io.mockk.impl.annotations.MockK
 import io.mockk.just
 import io.mockk.mockk
 import io.mockk.slot
-import io.mockk.spyk
 import io.mockk.verify
 import io.mockk.verifyOrder
 import kotlinx.coroutines.ExperimentalCoroutinesApi
@@ -37,7 +38,6 @@ import org.mozilla.fenix.R
 import org.mozilla.fenix.components.FenixSnackbar
 import org.mozilla.fenix.components.metrics.Event
 import org.mozilla.fenix.components.metrics.MetricController
-import org.mozilla.fenix.ext.metrics
 import org.mozilla.fenix.ext.nav
 import org.mozilla.fenix.helpers.FenixRobolectricTestRunner
 import org.mozilla.fenix.share.listadapters.AppShareOption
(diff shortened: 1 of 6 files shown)


app/src/test/java/org/mozilla/fenix/share/ShareControllerTest.kt
```diff
@@ -60,41 +66,36 @@ class ShareControllerTest {
...
-    private val recentAppStorage = mockk<RecentAppsStorage>(relaxed = true)
-    private val controller = DefaultShareController(
-        context, shareSubject, shareData, sendTabUseCases, snackbar, navController,
-        recentAppStorage, testCoroutineScope, dismiss
-    )
 
     @Before
     fun setUp() {
-        every { context.metrics } returns metrics
+        MockKAnnotations.init(this)
```
eliserichards, line 72: 👏

app/src/test/java/org/mozilla/fenix/share/ShareControllerTest.kt
```diff
@@ -60,41 +66,36 @@ class ShareControllerTest {
...
     }
 
     @Test
     fun `handleShareToApp should start a new sharing activity and close this`() = runBlocking {
-        val appPackageName = "package"
-        val appClassName = "activity"
-        val appShareOption = AppShareOption("app", mockk(), appPackageName, appClassName)
+        val appShareOption = AppShareOption(
+            name = "app",
+            icon = mockk(),
```
eliserichards, line 87: This is cool! didn't realize you could do this `(... icon = mockk(), ...)`

app/src/test/java/org/mozilla/fenix/share/ShareControllerTest.kt
```diff
@@ -106,8 +107,8 @@ class ShareControllerTest {
         assertEquals(textToShare, shareIntent.captured.extras!![Intent.EXTRA_TEXT])
         assertEquals("text/plain", shareIntent.captured.type)
         assertEquals(Intent.FLAG_ACTIVITY_NEW_TASK, shareIntent.captured.flags)
-        assertEquals(appPackageName, shareIntent.captured.component!!.packageName)
-        assertEquals(appClassName, shareIntent.captured.component!!.className)
+        assertEquals("package", shareIntent.captured.component!!.packageName)
```
eliserichards, line 110: I'd prefer to keep these `val`s

app/src/test/java/org/mozilla/fenix/share/ShareControllerTest.kt
```diff
@@ -149,6 +153,7 @@ class ShareControllerTest {
         val deviceId = slot<String>()
         val tabsShared = slot<List<TabData>>()
 
+        val controller = buildController()
```
eliserichards, line 156: Can we `buildController()` in setup instead of calling it in every test?

app/src/test/java/org/mozilla/fenix/share/ShareControllerTest.kt
```diff
@@ -272,7 +275,7 @@ class ShareControllerTest {
 
     @Test
     fun `getShareText should respect concatenate shared tabs urls`() {
-        assertEquals(textToShare, controller.getShareText())
+        assertEquals(textToShare, buildController(data = shareData).getShareText())
```
eliserichards, line 278: I'm guessing that the answer to above q is "no" based on this 😛

eliserichards (dismissed):

app/src/test/java/org/mozilla/fenix/share/ShareControllerTest.kt
NotWoods: Yeah, the point of `buildController` is to make it easier to override in some tests. We could always build a plain controller and sometimes make a second one, but I felt that it wasn't a lot of code to just call the plain function in some locations.

ekager (approved):

Issue #13478, opened 2020-08-10
FNX-14658 ⁃ [Bug][LeakCanary] CoordinatorLayout leaked


cadeyrn:
/cc @mcomella 

### Steps to reproduce

uninstall an add-on.

### Expected behavior

No leak.

### Actual behavior

┬───
│ GC Root: System class
│
├─ android.view.inputmethod.InputMethodManager class
│    Leaking: NO (InputMethodManager↓ is not leaking and a class is never leaking)
│    ↓ static InputMethodManager.sInstance
├─ android.view.inputmethod.InputMethodManager instance
│    Leaking: NO (DecorView↓ is not leaking and InputMethodManager is a singleton)
│    ↓ InputMethodManager.mCurRootView
├─ com.android.internal.policy.DecorView instance
│    Leaking: NO (RecyclerView↓ is not leaking and View attached)
│    mContext instance of com.android.internal.policy.DecorContext, wrapping activity org.mozilla.fenix.HomeActivity with mDestroyed = false
│    Parent android.view.ViewRootImpl not a android.view.View
│    View#mParent is set
│    View#mAttachInfo is not null (view attached)
│    View.mWindowAttachCount = 1
│    ↓ DecorView.mAttachInfo
├─ android.view.View$AttachInfo instance
│    Leaking: NO (RecyclerView↓ is not leaking)
│    ↓ View$AttachInfo.mScrollContainers
├─ java.util.ArrayList instance
│    Leaking: NO (RecyclerView↓ is not leaking)
│    ↓ ArrayList.elementData
├─ java.lang.Object[] array
│    Leaking: NO (RecyclerView↓ is not leaking)
│    ↓ Object[].[0]
├─ androidx.recyclerview.widget.RecyclerView instance
│    Leaking: NO (View attached)
│    mContext instance of org.mozilla.fenix.HomeActivity with mDestroyed = false
│    View.parent androidx.coordinatorlayout.widget.CoordinatorLayout attached as well
│    View#mParent is set
│    View#mAttachInfo is not null (view attached)
│    View.mID = R.id.add_ons_list
│    View.mWindowAttachCount = 1
│    ↓ RecyclerView.mAdapter
│                   ~~~~~~~~
├─ mozilla.components.feature.addons.ui.AddonsManagerAdapter instance
│    Leaking: UNKNOWN
│    ↓ AddonsManagerAdapter.mObservable
│                           ~~~~~~~~~~~
[...]

PR #13944, opened 2020-08-19
for #13478, release fragment resource inside onDestroyView of AddonsManagementFragment to avoid memory leak

sraturi:


sraturi:
@MarcLeclair Looks like the leak is not happing after removing the add-on but rather when we are leaving the `AddonsManagementFragment` (background) and onDestroyView is called but resources are not released. 
More in-depth detail in https://stackoverflow.com/questions/58080268/keeping-a-reference-to-a-view-in-a-fragment-causes-memory-leaks

liuche:
Might need a rebase, there was some lint breakage on master earlier.

Issue #13368, opened 2020-08-06
FNX-14548 ⁃ [Bug] Strictmode violation on cold startup for SharedPrefs in LeanplumMetricsService.init


sraturi:
There is a strict mode violation (on disk read) caused by sharedpref on startup inside `LeanplumMetricsService`

Tested on pixel 2 and samsung

*** Here are the logs from the violation

```
2020-08-06 13:40:54.156 10765-10765/org.mozilla.fenix.debug D/StrictMode: StrictMode policy violation; ~duration=20 ms: android.os.StrictMode$StrictModeDiskReadViolation: policy=196671 violation=2
        at android.os.StrictMode$AndroidBlockGuardPolicy.onReadFromDisk(StrictMode.java:1438)
        at java.io.UnixFileSystem.checkAccess(UnixFileSystem.java:251)
        at java.io.File.exists(File.java:807)
        at android.app.ContextImpl.ensurePrivateDirExists(ContextImpl.java:565)
        at android.app.ContextImpl.ensurePrivateDirExists(ContextImpl.java:556)
        at android.app.ContextImpl.getPreferencesDir(ContextImpl.java:512)
        at android.app.ContextImpl.getSharedPreferencesPath(ContextImpl.java:704)
        at android.app.ContextImpl.getSharedPreferences(ContextImpl.java:372)
        at android.content.ContextWrapper.getSharedPreferences(ContextWrapper.java:178)
        at org.mozilla.fenix.components.metrics.LeanplumMetricsService.<init>(LeanplumMetricsService.kt:82)
        at org.mozilla.fenix.components.metrics.LeanplumMetricsService.<init>(LeanplumMetricsService.kt:57)
        at org.mozilla.fenix.components.Analytics$metrics$2.invoke(Analytics.kt:91)
        at org.mozilla.fenix.components.Analytics$metrics$2.invoke(Analytics.kt:36)
        at kotlin.SynchronizedLazyImpl.getValue(LazyJVM.kt:74)
        at org.mozilla.fenix.components.Analytics.getMetrics(Unknown Source:2)
        at org.mozilla.fenix.FenixApplication.startMetricsIfEnabled(FenixApplication.kt:231)
        at org.mozilla.fenix.FenixApplication.setupInMainProcessOnly(FenixApplication.kt:146)
        at org.mozilla.fenix.FenixApplication.onCreate(FenixApplication.kt:93)
[...]

PR #13971, opened 2020-08-20
For #13368  - Fix StrictMode violation in LeanPlum

codrut-topliceanu:
For https://github.com/mozilla-mobile/fenix/issues/13368
### Pull Request checklist

- [x] **Screenshots**: The code in this PR does not include any user facing features.

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/metrics/LeanplumMetricsService.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/metrics/LeanplumMetricsService.kt
@@ -6,8 +6,8 @@ package org.mozilla.fenix.components.metrics
 
 import android.app.Application
 import android.content.Context.MODE_PRIVATE
+import android.content.SharedPreferences
 import android.net.Uri
-import android.os.StrictMode
 import android.util.Log
 import androidx.annotation.VisibleForTesting
 import com.leanplum.Leanplum
@@ -23,7 +23,6 @@ import kotlinx.coroutines.withContext
 import mozilla.components.support.locale.LocaleManager
 import org.mozilla.fenix.BuildConfig
 import org.mozilla.fenix.components.metrics.MozillaProductDetector.MozillaProducts
-import org.mozilla.fenix.ext.resetPoliciesAfter
 import org.mozilla.fenix.ext.settings
 import org.mozilla.fenix.home.intent.DeepLinkIntentProcessor
 import java.util.Locale
@@ -83,9 +82,7 @@ class LeanplumMetricsService(
     override val type = MetricServiceType.Marketing
     private val token = Token(LeanplumId, LeanplumToken)
 
-    private val preferences = StrictMode.allowThreadDiskReads().resetPoliciesAfter {
-        application.getSharedPreferences(PREFERENCE_NAME, MODE_PRIVATE)
-    }
+    private lateinit var preferences: SharedPreferences
 
     @VisibleForTesting
     internal val deviceId by lazy {
@@ -110,7 +107,9 @@ class LeanplumMetricsService(
         Parser.parseVariables(application)
 
         leanplumJob = scope.launch {
-
+            withContext(Dispatchers.IO) {
+                preferences = application.getSharedPreferences(PREFERENCE_NAME, MODE_PRIVATE)
+            }
             val applicationSetLocale = LocaleManager.getCurrentLocale(application)
             val currentLocale = applicationSetLocale ?: Locale.getDefault()
             val languageCode =
--- a/app/src/main/java/org/mozilla/fenix/components/metrics/LeanplumMetricsService.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/metrics/LeanplumMetricsService.kt
@@ -99,21 +99,19 @@ class LeanplumMetricsService(
 
     @Suppress("ComplexMethod")
     override fun start() {
-        val preferencesLoaded = scope.async(Dispatchers.IO) {
+        val preferencesLoaded = scope.async {
                 preferences = application.getSharedPreferences(PREFERENCE_NAME, MODE_PRIVATE)
         }
 
         if (!application.settings().isMarketingTelemetryEnabled) return
 
-        scope.launch {
+        leanplumJob = scope.launch {
             preferencesLoaded.await()
             Leanplum.setIsTestModeEnabled(false)
             Leanplum.setApplicationContext(application)
             Leanplum.setDeviceId(deviceId)
             Parser.parseVariables(application)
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/fenix/components/metrics/LeanplumMetricsService.kt
```diff
@@ -106,7 +107,9 @@ class LeanplumMetricsService(
         Parser.parseVariables(application)
 
         leanplumJob = scope.launch {
-
+            withContext(Dispatchers.IO) {
+                preferences = application.getSharedPreferences(PREFERENCE_NAME, MODE_PRIVATE)
```
mcomella, line 111: This gets initialized too late – it's needed for `Leanplum.setDeviceId` – and results in a unit test failure.

I wonder if we couldn't just move `Leanplum.setDeviceId` into the IO dispatcher – we'd have to make sure it can be set off the main thread and possibly used concurrently first though.

mcomella: We should also ensure earlier code doesn't need it (e.g. `Parser.parseVariables`)

app/src/main/java/org/mozilla/fenix/components/metrics/LeanplumMetricsService.kt
```diff
@@ -101,16 +99,21 @@ class LeanplumMetricsService(
 
     @Suppress("ComplexMethod")
     override fun start() {
+        val preferencesLoaded = scope.async(Dispatchers.IO) {
```
mcomella, line 102: Scope is already defined as `Dispatchers.IO` – could we make this a part of the leanplumJob too? (see below)

codrut-topliceanu:
Hey, thank you for taking a look at this. I think that maybe this issue is beyond me.

app/src/main/java/org/mozilla/fenix/components/metrics/LeanplumMetricsService.kt
```diff
@@ -101,15 +99,18 @@ class LeanplumMetricsService(
 
     @Suppress("ComplexMethod")
     override fun start() {
+        val preferencesLoaded = scope.async {
+                preferences = application.getSharedPreferences(PREFERENCE_NAME, MODE_PRIVATE)
+        }
 
         if (!application.settings().isMarketingTelemetryEnabled) return
```
mcomella, line 106: This will read from SharedPreferences under the hood: I'm surprised it doesn't set off StrictMode. Maybe we should move it into the launch statement too?

app/src/main/java/org/mozilla/fenix/components/metrics/LeanplumMetricsService.kt
```diff
@@ -101,15 +99,18 @@ class LeanplumMetricsService(
 
     @Suppress("ComplexMethod")
     override fun start() {
+        val preferencesLoaded = scope.async {
+                preferences = application.getSharedPreferences(PREFERENCE_NAME, MODE_PRIVATE)
```
mcomella, line 103: optional nit: in the current state, the `preferences` reference is fragile: e.g. in a future refactor, someone can access it before its initialized. This is particularly a problem if someone tries to access `deviceId` because it isn't clear it relies on `preferences` being initialized before its initialized.

However, I can think of one way to make it at least a little better: `preferences` is only used to set the deviceId – why don't we only create it in the scope where it's needed to init `deviceId`? Something like...
```kotlin
lateinit var deviceId: String

override fun start() {
    leanplumJob = scope.launch {
        // ...
        val preferences = application.getSharedPreference(PREFERENCE_NAME, MODE_PRIVATE)
        this.deviceId = fetchOrGenerateDeviceId(preferences)
        Leanplum.setDeviceId(this.deviceId)
        // ...
```

And `fetchOrGenerateDeviceId` is the contents of the current deviceId lazy getter.

`deviceId` is still fragile but its more clear why and we avoid exposing preferences where it can be used easily to violate StrictMode.

app/src/test/java/org/mozilla/fenix/components/metrics/LeanplumMetricsServiceTest.kt
```diff
@@ -31,9 +31,11 @@ class LeanplumMetricsServiceTest {
         assertNull(sharedPreferences.getString("LP_DEVICE_ID", null))
 
         val leanplumMetricService = LeanplumMetricsService(testContext.application, idGenerator)
+        leanplumMetricService.start()
```
mcomella, line 34: nit: maybe add a comment here and below that this is needed to init deviceId?

It's curious that in the original implementation, a deviceId could be fetched even if the user had telemetry disabled, huh? StrictMode seems to be making things better in other ways too. :)

app/src/main/java/org/mozilla/fenix/components/metrics/LeanplumMetricsService.kt
```diff
@@ -101,15 +99,18 @@ class LeanplumMetricsService(
...
-        Leanplum.setApplicationContext(application)
-        Leanplum.setDeviceId(deviceId)
-        Parser.parseVariables(application)
-
         leanplumJob = scope.launch {
+            preferencesLoaded.await()
+            Leanplum.setIsTestModeEnabled(false)
+            Leanplum.setApplicationContext(application)
+            Leanplum.setDeviceId(deviceId)
+            Parser.parseVariables(application)
```
mcomella, line 113: I can't spend enough time on this to know if this is thread safe. It's not documented as thread safe but many of the components it uses are synchronized. However, like a lot of Leanplum, none of the fields are documented as `volatile`, which makes me think it's technically not thread safe but we're treating it as such anyway.

However, from a theory perspective, I think this is safe (minus the volatile issue): leanplum hasn't started yet so no other code should be running to cause concurrency issues. It could be a problem if `LeanplumMetricsSerivce.start` is called twice without a `stop` in between, however. But this shouldn't happen because it's started in the Activity.

Also, we don't add any Leanplum annotations (e.g. `@Variable`) so in the current code, this method will exit early. That's good for now but it might create issues if we later do use other annotations.

@boek What do you think? Is it safe to put `Parser.parseVariables` on a background thread?

mcomella (commented):
Let's wait to see what @boek has to say before landing this. With boek's approval, I think this can land as is but I think we can make this less fragile and perhaps more performant by following the nits.

mcomella:
By the way, thanks for working on this! :)

boek:
We're no longer using shared prefs in the leanplum metric service

Issue #11465, opened 2020-06-11
[Bug] Searching through voice assistant from the search widget is done through a different search engine than the one set as default


softvision-miralobontiu:
## Steps to reproduce
1. Go to Settings - Search, add a custom search engine, like imdb.com.
1. Set imdb as the default search engine.
1. Browse a page using the new search engine.
1. Delete all the search engines - I remained with DuckDuckGo as default.
1. Add Fenix widget.
1. Tap on the microphone from the search widget.
1. Say a word and observe which search engine performs the search.

### Expected behavior
The search is performed by DuckDuckGo, in this scenario.

### Actual behavior
The search is performed by imdb.

### Device information
* Android devices: Samsung Galaxy Tab S6 (Android 9), and Sony Xperia Z5 Premium (Android 7.1.1)
* Fenix version: 6/11 Nightly build, Beta 5.2.0-beta.1

![](https://user-images.githubusercontent.com/40694158/84369121-106f2e80-abdf-11ea-811e-e9974c5a388c.gif)

codrut-topliceanu:
@kbrosnan I'm still working on the PR for this issue. :D

softvision-miralobontiu:
I was not able to reproduce this issue anymore on the latest Nightly build from 9/28, nor on RC 81.1.1, using Samsung Galaxy Tab S9 (Android 10).
I will close this issue.

PR #14015, opened 2020-08-21
For #11465  - Fix default engine for widget voice search

codrut-topliceanu:
For https://github.com/mozilla-mobile/fenix/issues/11465

### Pull Request checklist

- [x] **Screenshots**: This PR includes a GIF.
### After merge

1. click on Show All Checks,

voice widget

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/settings/search/SearchEngineListPreference.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/search/SearchEngineListPreference.kt
@@ -102,6 +102,10 @@ abstract class SearchEngineListPreference @JvmOverloads constructor(
             engineItem.tag = engineId
             if (engineId == selectedEngine) {
                 updateDefaultItem(engineItem.radio_button)
+                /* #11465 -> radio_button.isChecked = true does not trigger
+                * onSearchEngineSelected because searchEngineGroup has null views at that point.
+                * So we trigger it here.*/
+                onSearchEngineSelected(engine)
             }
             searchEngineGroup!!.addView(engineItem, layoutParams)
         }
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/fenix/settings/search/SearchEngineListPreference.kt
```diff
@@ -76,6 +76,11 @@ abstract class SearchEngineListPreference @JvmOverloads constructor(
             it.identifier == defaultEngine
         } ?: searchEngineList.list.first()).identifier
 
+        context.components.search.searchEngineManager.defaultSearchEngine =
+            searchEngineList.list.find {
+                it.identifier == selectedEngine
```
Mugurell, line 81: Setting the default involves two calls:
>        context.components.search.searchEngineManager.defaultSearchEngine = searchEngine
>        context.settings().defaultSearchEngineName = searchEngine.name

app/src/main/java/org/mozilla/fenix/settings/search/SearchEngineListPreference.kt
```diff
@@ -76,6 +76,11 @@ abstract class SearchEngineListPreference @JvmOverloads constructor(
             it.identifier == defaultEngine
         } ?: searchEngineList.list.first()).identifier
 
+        context.components.search.searchEngineManager.defaultSearchEngine =
```
Mugurell, line 79: This would work but I don't think it's the right place to set the default.
This adds another responsibility to a method that should just "refreshSearchEngineViews".
If the issue is in `deleteSearchEngine()` maybe we can fix it there?

Mugurell (changes requested):

codrut-topliceanu:
The issue seems to be located in `SearchEngineListPreference.onCheckedChanged()`. 
When a search engine is deleted `onCheckedChanged()` should be called, and then `when...true -> onSearchEngineSelected(engine)`, however that never happens as `wrapper` is always null at that point.

Mugurell (approved):
Thanks for looking into this!

Issue #13935, opened 2020-08-19
UX Polish and Custom Icons For #349 View Downloads in Browser


kglazko:
## User Story
- As a user, I want View Downloads to change icons in the list depending on the type of file the download is
- As a user, I would like to see clearer copy when viewing the empty state of downloads, bookmarks and history

## Dependencies
- List dependencies on other issues/teams etc.

### Acceptance Criteria
- I can do …  (e.g. add a bookmark)
- Empty state copy for the three sections should be as follows:
  - **Downloads:** `No downloaded files`
  - **Bookmarks:** `No pages bookmarked`
  - **History:** `No browsing history`

brampitoyo:
> As a user, I want View Downloads to change icons in the list depending on the type of file the download is

@kglazko Hi! @apbitner and I are on track to deliver the file type icons by the end of the week.

I’ve got the assets ready for you to try in product:
https://drive.google.com/drive/folders/1Kgbotds9orO1ORB6ikfZY0kxZLYQWol7?usp=sharing

The file names are all prefixed with `file-type`
- apk
- audio (use `note`)
- default (used for generic or unknown file type)
- document
- image
- video
- zip

And we should be using the same outer frame as the one used elsewhere in the product – rounded rect with border.

<img width="360" src="https://user-images.githubusercontent.com/51007/90711033-cea3e800-e2f3-11ea-91ba-c654189f6edc.png">

brampitoyo:
@kglazko wrote:

> …our default currently in History/Library is rounded icons. Is it ok to keep them consistent for this iteration?

Yes. In this iteration, let’s keep them rounded – consistent with History, Library, Bookmarks, etc.

In the future, will it be easy to re-change file types to be consistent with all other frames, when they all go from rounded → rounded rectangle? Our intention is to make sure that total consistency will happen at some point.

softvision-miralobontiu:
Hi all,
Verified on the latest Nightly build from 8/26 with Samsung Galaxy Tab S6 (Android 9), and Sony Xperia Z5 Premium (Android 7.1.1), that:
- the icons from the Downloads page are rounded, 
- the icons represent the type of downloaded files (image, audio, video, zip, etc),
- and that when there are no downloaded files, the empty state has "No downloads here" text displayed.

![downloads](https://user-images.githubusercontent.com/40694158/91295293-86793f80-e7a3-11ea-858c-7d1e48c4047f.png)

[...]

PR #14036, opened 2020-08-21
For #13935: Enhanced File Type List Icons

kglazko:
### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/ext/DownloadItem.kt
+++ b/app/src/main/java/org/mozilla/fenix/ext/DownloadItem.kt
@@ -1,0 +1,48 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.ext
+
+import org.mozilla.fenix.R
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mozilla/fenix/ext/DownloadItem.kt
```diff
@@ -0,0 +1,48 @@
...
+
+import org.mozilla.fenix.R
+import org.mozilla.fenix.library.downloads.DownloadItem
+
+// While this looks complex, it's actually pretty simple.
+@SuppressWarnings("ComplexMethod")
+fun DownloadItem.getIcon(): Int {
+    fun getIconCornerCases(fileName: String?): Int {
+        return when {
+            fileName?.endsWith("apk") == true -> R.drawable.ic_file_type_apk
```
st3fan, line 15: @kglazko these checks are not very strong and fail on common filenames. for example `blahzip` is also recognized as a zip file and below `application/tartsaretasty` is recognized as a `ic_file_type_zip` because  it `contains` the substring `tar`.

It is better to be more precise for example by splitting the content type on `/` and looking for exact matches instead of `contains`. Similar with filenames which you can split on the final `.` and then look at the extension.

Amejia481: That's a good catch! We would use `File("filename.ext").extension` for validating.

Issue #9709, opened 2020-04-04
[Bug] "phone" in translations


cadeyrn:
Some translations contains the word "phone", for example:

`<string name="add_to_homescreen_description">You can easily add this website to your phone’s Home screen to have instant access and browse faster with an app-like experience.</string>`

`<string name="onboarding_firefox_account_auto_signin_header_2">You are signed in as %s on another Firefox browser on this phone. Would you like to sign in with this account?</string>`

(edit: the second one is already filed as #8343)

Maybe replace this with "device"? Because a tablet is not a phone and you can use Fenix on phones and tablets.

sflorean:
Verified as fixed on Nightly 9/11 and 81.1.1-beta.3 with LG G7 Fit (Android 8.0).

Issue #8343, opened 2020-02-12
[Bug] Consider changing the sign-in message from "on this phone" with "on this device"


softvision-miralobontiu:
### Prerequisites
1. Use a tablet.
2. Sign-in on Fennec with a valid account.

## Steps to reproduce
1. Open Fenix Nightly with a clean profile.
2. Observe the "Sign-in" card.

### Expected behavior
"You are signed in as ... on another Firefox browser on this device." should be displayed.

### Actual behavior
"You are signed in as ... on another Firefox browser on this phone." is displayed.

### Device information
* Android device: Sony Xperia Z2 (Android 6.0.1)
* Fenix version: 2/12 Nightly build

![Untitled](https://user-images.githubusercontent.com/40694158/74321617-3c8eb600-4d8b-11ea-97b1-c4080b92b976.png)

ekager:
CC @betsymi for feedback!

betsymi:
Good catch! Let's update. Quick Q, will this sign them out from the other browser? Or, will they stay signed in, then ultimately be signed in in two places?

cadeyrn:
@betsymi Could you also have a look at #9709, it's a very similar issue. Thanks!

AmyYLee:
Sounds like UX is resolved. Removing label.

sflorean:
Verified as fixed on Nightly 9/11 with Google Pixel 3 (Android 11).
On Beta 81.1.1-beta.3 the issue can't be verified due to #6521.

PR #14083, opened 2020-08-22
For #8343, #9709 - replace "phone" with "device" in translations

cadeyrn:
The change got approval in https://github.com/mozilla-mobile/fenix/issues/8343#issuecomment-620259024 for the one string, and #9709 is exactly the same for another string so it shouldn't need explicit approval.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -1034,7 +1034,7 @@
     <!-- text for the firefox account onboarding card header when we detect you're already signed in to
         another Firefox browser. (The word `Firefox` should not be translated)
         The first parameter is the email of the detected user's account -->
-    <string name="onboarding_firefox_account_auto_signin_header_2">You are signed in as %s on another Firefox browser on this phone. Would you like to sign in with this account?</string>
+    <string name="onboarding_firefox_account_auto_signin_header_3">You are signed in as %s on another Firefox browser on this device. Would you like to sign in with this account?</string>
     <!-- text for the button to confirm automatic sign-in -->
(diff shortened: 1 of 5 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -981,7 +981,7 @@
     <!-- text for the firefox account onboarding card header when we detect you're already signed in to
         another Firefox browser. (The word `Firefox` should not be translated)
         The first parameter is the email of the detected user's account -->
-    <string name="onboarding_firefox_account_auto_signin_header_2">You are signed in as %s on another Firefox browser on this phone. Would you like to sign in with this account?</string>
```
liuche, line 984: Thanks for adding this as a new string id so it will get translated!

One thing that we're doing is quarantining string removals, since our localization will be happening during the Beta period, so new strings will need to stick around bc the new translations land in Nightly and need to be uplifted. See https://github.com/mozilla-mobile/fenix/issues/14176

cadeyrn:
@liuche I restored the old strings with a deprecation note.

liuche:
There's an onboarding test that this fails, when I run [the pre-push hook](https://github.com/mozilla-mobile/fenix/#pre-push-hooks): `OnboardingAutomaticSignInViewHolderTest`

In the future, the pre-push hook also catches a lot of lint and unit test failures!

Could you look at it and update?

cadeyrn:
@liuche 

> There's an onboarding test that this fails, when I run the pre-push hook: OnboardingAutomaticSignInViewHolderTest

Sorry about that, I should have noticed this. I fixed the test!

Issue #5764, opened 2019-10-03
Refactor Downloads for Android 11 Storage + Change download path


vesta0:
**Part of Meta: #4417**

### User Story
As a user, I want to be able to choose where my downloads are saved in my phone's storage, so I can have control and flexibility over my downloaded content. 

### Acceptance Criteria
[ ] I can see the default download location/file path
[ ] I can change the default download location

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-13743)

Poopooracoocoo:
this is also great on devices with not much storage and SD cards

eliserichards:
Chrome only shows this setting if you alter the flag (`chrome://flags/#enable-downloads-location-change`). This may need to live in our Advanced Settings, with some regex to guarantee that any change is valid & reversible

Poopooracoocoo:
Google says that they want to get rid of flags after some time (but Google says a lot of things lol). I think they'll eventually turn it on by default and remove it.

eliserichards:
There will be [major changes in Android 11](https://github.com/mozilla-mobile/fenix/pull/14121#issuecomment-683151556) to the way that they are allowing us to access storage. If we continue with this card, we will have to do a pretty big refactor on the AC side to use the new Android 11 storage, and this card would block additional work on downloads bugs. 

Confirmed with @vesta0 that this should go back into the backlog for now.

Imerion:
This would be very useful indeed. Does it include an option to always choose where to save a file? That is something I'm really missing! Could open another issue for that otherwise.

Poopooracoocoo:
and something Chrome's been testing out is _when_. Like waiting until there is a WiFi connection, for example.

Amejia481:
> This would be very useful indeed. Does it include an option to always choose where to save a file? That is something I'm really missing! Could open another issue for that otherwise.

Yes, this includes the default path where files will be saved.

Imerion:
@Amejia481 Right, but I mean choosing the desired location for every initiated download, not just having one folder where everything gets dumped.

Amejia481:
Got it, that's a different feature!

cadeyrn:
This is still a valid feature request (comment for bot).

Chaz6:
This bug is almost 2 years old but it represents a major usability problem. Is there any ETA for when this will be fixed?

Chaz6:
I will commit to a £25 bounty when this is fixed.

[...]

Issue #13378, opened 2020-08-07
FNX-14558 ⁃ Duplicate shared preferences


Amejia481:
In [Setting.kt](https://github.com/mozilla-mobile/fenix/blob/06ba8a020062073281f19c846c6ec6135bc16b32/app/src/main/java/org/mozilla/fenix/utils/Settings.kt), we are writing  to [fenix_preferences.xml](https://github.com/mozilla-mobile/fenix/blob/06ba8a020062073281f19c846c6ec6135bc16b32/app/src/main/java/org/mozilla/fenix/utils/Settings.kt#L57) while  [SettingsFragment.kt](https://github.com/mozilla-mobile/fenix/blob/06ba8a020062073281f19c846c6ec6135bc16b32/app/src/main/java/org/mozilla/fenix/settings/SettingsFragment.kt) we are writing to [org.mozilla.fenix.debug_preferences.xml](https://github.com/mozilla-mobile/fenix/blob/06ba8a020062073281f19c846c6ec6135bc16b32/app/src/main/java/org/mozilla/fenix/settings/SettingsFragment.kt#L122). Is that intentional? 🤔 

We have multiple keys duplicated in both files:

![image](https://user-images.githubusercontent.com/773158/89608163-5e508d80-d842-11ea-8c73-edb38ee785f7.png)
![image](https://user-images.githubusercontent.com/773158/89608919-690c2200-d844-11ea-81a3-bf193fcff64d.png)

____________________________________

I'm not sure if this is intentional, but it caused me a lot of confusion while I was adding a new settings on `SettingsFragment.kt`.

<img width="472" alt="image" src="https://user-images.githubusercontent.com/773158/89609519-1b90b480-d846-11ea-8998-518ee0518ccf.png">

To access it programmatically, I added a new property to `Settings.kt`, as I thought it was the convention for accessing share preference from `SettingsFragment.kt`. 

```koltin
    var allowExternalDownloadManager by booleanPreference(
        appContext.getPreferenceKey(R.string.pref_key_external_download_manager),
        default = false
    )
```

[...]

PR #14121, opened 2020-08-24
For #5764: change download path

eliserichards:
For #5764 
Depends on AC pr: https://github.com/mozilla-mobile/android-components/pull/8194

todo:
 - [ ] land AC pr - in progress
 - [x] what is the correct way of pulling the settings from `core.engine` when initializing Fenix settings and when saving a new value?
 - [x] allow user to choose between local storage `Environment.DIRECTORY_DOWNLOADS` and an SD card if present
   - we don't want to let them save anywhere because 1. we'd have to validate the path, and 2. Android 11 will introduce [scoped storage](https://developer.android.com/preview/privacy/storage)

### Pull Request checklist

### After merge

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
```diff
@@ -291,8 +291,13 @@ class BrowserFragment : BaseBrowserFragment(), UserInteractionHandler {
...
             view,
-            FenixSnackbarDelegate(view)
+            FenixSnackbarDelegate(view),
+            getDefaultDownloadPath()
         ) + ContextMenuCandidate.createOpenInExternalAppCandidate(requireContext(),
             contextMenuCandidateAppLinksUseCases)
     }
+
+    private fun getDefaultDownloadPath(): String {
+        return requireActivity().settings().downloadPath
```
Amejia481, line 301: It this getting the latest updated value? I was experiencing issues using the `setting` class see https://github.com/mozilla-mobile/fenix/issues/13378. I was forced to use : `PreferenceManager.getDefaultSharedPreferences(context)` to get the latest value

Amejia481: I was thinking that maybe `getDefaultDownloadDirectoy` must clear, I think we should do the same on ac part :)

eliserichards:
Android 11 is going to introduce some issues in the way we are handling downloads. Currently--in Android 10 and below--an app can have access to any & all directories given the read/write permission. This means that there's no differentiation between public directories (like `/Download`) or private ones. Starting in Android 11, storage will be restricted; apps will only have access to public directories and their own directories, and permissions will be required for access to any others. 

We are seeing this functionality in Opera, where

eliserichards:
Closing for now. See comment: https://github.com/mozilla-mobile/fenix/issues/5764#issuecomment-683879718

Issue #13507, opened 2020-08-11
In-app review prompt


vesta0:
## User Story
- As a user, I want to have the option to easily rate the app without it disruptive to my browsing flow. 

## Dependencies
Using the [Google In-app review API](https://developer.android.com/guide/playcore/in-app-review) since it will allow user to rate the app without having the leave the browser. 

### Acceptance Criteria
- I can rate the app without being taken out of the app (to the play store)
- I am not prompted excessively for a review (max 3 times a year)

liuche:
I think we decided that this would be something that we want to do through LeanPlum.

Do we need to talk to anyone from marketing for what [specific probes](https://docs.google.com/document/d/1-RWxA9WsMr1FvYW6ap2ZVDMl235vmm2a4xvn0rFZDZI/edit?disco=AAAAKE7G-Q0&ts=5f32f2ff&usp=comment_email_document&usp_dm=false) to add @vesta0 ? (I think comment 0 may need to be updated)

liuche:
Moving this back to sourpatch bc the search work is almost done!

liuche:
Just to be explicit, this does **not require any strings**, per the Android guidelines. We can only control when the prompt is shown, and we must not show it too often.

```
Your app should not ask the user any questions before or while presenting the rating button or card, including questions about their opinion (such as “Do you like the app?”) or predictive questions (such as “Would you rate this app 5 stars”).
```

liuche:
This **should** be triggered on Nightly (but depends on Google Play Services):
- fenix is set to default browser
- restart browser 5 times

We're not entirely clear on when Google Play chooses to show this, but those steps will set it up to be able to be shown (so if it doesn't happen exactly with those steps, please note it in the comments, but we don't actually have full control over this)

If you want to trigger it again, it needs to be a fresh install because one of the restrictions is that GPS will not show the prompt too close together (by a few weeks).
This may not work if you have already rated Firefox Nightly.

Link to Google API can provide more info on how it should look.

LaurentiuApahideanSV:
I tested the issue on Firefox Preview Nightly 200831 (Build #2015761139) but was unable to trigger the In-App Review prompt. I tried the following scenario:
1. Log in the Play Store with a google account that has not rated Firefox Preview Nightly;
2. Remove all other google accounts found on the device;
3. Restart Firefox Preview more than 5 times;
4. Browse through web pages;
5. Navigate through the setting menu.

[...]

Issue #13289, opened 2020-08-05
FNX2-18892 ⁃ SPIKE: Add LeanPlum instrumentation for in-product user feedback


vesta0:
We would like to build a non-intrusive way of asking for user feedback (rating/reviews) within the browser (preferably without taking the user out of the app)

### Acceptance Criteria
- Asking for feedback once every 4 months (if dismissed)
- Shown to active users (e.g. users opened the app at least 5 times and completed a meaningful action (creating a collection, switching to reader mode, installing an add-on, or looking at ETP or Private Browsing settings.
- Prompt should include deeplink to play store review page.
- If a user says no to the prompt 3 times, we should stop asking them.

Also see: https://developer.android.com/guide/playcore/in-app-review

bifleming:
Timebox to 5 points

liuche:
The first step would be to meet w/ the Mobile engagement team to determine the requirements for this.

The goals we're aiming for would be:
- have an idea of what the Leanplum work would be from talking to the team
- recommend the Google API vs the Leanplum as an option

person808:
From a brief reading of the API guide:

1. The actual code to launch the review popup looks fairly simple. Something like this (basically copied from the docs):

```kotlin
val request = manager.requestReviewFlow()
request.addOnCompleteListener { request ->
    if (request.isSuccessful) {
        // We got the ReviewInfo object
        val reviewInfo = request.result
        val flow = manager.launchReviewFlow(activity, reviewInfo)
        flow.addOnCompleteListener { _ ->
            // The flow has finished. The API does not indicate whether the user
            // reviewed or not, or even whether the review dialog was shown. Thus, no
            // matter the result, we continue our app flow.
        }
    } else {
        // There was some problem, continue regardless of the result.
    }
}
```

The docs say we should request the `reviewInfo` ahead of time, but only when we know we want to show the review dialog which is what most of the work on our end would be.

2. The API comes with some strict requirements on its usage listed in the overview linked in comment 0. No prompting the user to trigger the review dialog and theres a quota to how often this dialog can be triggered. If we exceed the quota, this API will silently fail, which is why we should not trigger the review dialog through explicit user action (such as a dialog asking them to rate our app).

3. There is a page on testing the API (https://developer.android.com/guide/playcore/in-app-review/test). This is getting into an area I'm not familiar with so just read the page I guess? I might whip up a small test app to try out the API if I have the time.

person808:
[...]

PR #14126, opened 2020-08-24
For #13507: Allow play store in app review to be shown to users

eliserichards:
For #13507 based on spike #13289

todo
- [x] add play store core [dependencies](https://developer.android.com/guide/playcore)
- [x] track num of times app has been opened since install
- [x] should show review prompt calculation: 
   > Review prompt should be triggered only for users that have Firefox set as their default browser AND after the 5th time the user opened the app.
- [x] should show max 3x a year aka every 4 mo aka every 120 days-ish
- [ ] _what's a good way to test this??_
   - manually, will be QA'd in Nightly
- [x] ensure it's shown correctly
   > The card and the card’s background should be on the topmost layer
- [x] ensure it's dismissed based on user action (click/swipe away or button click)

Should look like this: 

image

### Pull Request checklist

### After merge

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
@@ -176,6 +176,8 @@ class HomeFragment : Fragment() {
             if (!onboarding.userHasBeenOnboarded()) {
                 requireComponents.analytics.metrics.track(Event.OpenedAppFirstRun)
             }
+
+            requireComponents.services.reviewPromptController.promptReview(requireActivity())
         }
     }
 
@@ -570,24 +572,8 @@ class HomeFragment : Fragment() {
             recommendPrivateBrowsingShortcut()
         }
 
-        // In-app review prompt
-        requireContext().settings().incrementNumTimesOpenedAfterInstall()
-        handleInAppReviewPrompt()
-
         // We only want this observer live just before we navigate away to the collection creation screen
         requireComponents.core.tabCollectionStorage.unregister(collectionStorageObserver)
-    }
-
-    private fun handleInAppReviewPrompt() {
-        if (requireContext().settings().shouldShowUserFeedbackPrompt) {
-            lifecycleScope.launch {
-                val manager = ReviewManagerFactory.create(requireContext())
-                val reviewInfo = manager.requestReview()
-                manager.launchReview(requireActivity(), reviewInfo)
-                
-                requireContext().settings().incrementNumTimesFeedbackPromptShown()
-            }
-        }
     }
 
     private fun dispatchModeChanges(mode: Mode) {
--- a/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
@@ -173,8 +173,6 @@ class HomeFragment : Fragment() {
             if (!onboarding.userHasBeenOnboarded()) {
                 requireComponents.analytics.metrics.track(Event.OpenedAppFirstRun)
             }
-
-            requireComponents.services.reviewPromptController.promptReview(requireActivity())
         }
     }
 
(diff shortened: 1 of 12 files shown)


app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -554,10 +557,26 @@ class HomeFragment : Fragment() {
             recommendPrivateBrowsingShortcut()
         }
 
+        // In-app review prompt
+        requireContext().settings().incrementNumTimesOpenedAfterInstall()
```
boek, line 561: We'll probably want to move this into something that waits till after first paint (We may want to move this to FenixApplication `initVisualCompletenessQueueAndQueueTasks`. @mcomella is this the correct place?)

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -786,6 +789,62 @@ class Settings(private val appContext: Context) : PreferencesHolder {
...
+        ).apply()
+    }
+
+    private val timeWhenPromptWasLastShown: Int
+        get() = preferences.getInt(
+            appContext.getPreferenceKey(R.string.pref_key_time_prompt_shown),
+            0
+        )
+
+    @RequiresApi(Build.VERSION_CODES.O)
```
boek, line 824: Will this work on API level 21?

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -786,6 +789,62 @@ class Settings(private val appContext: Context) : PreferencesHolder {
             0
         )
 
+    private val numTimesFeedbackPromptShown: Int
```
boek, line 792: a-c has preference helpers to help with these!

`private val numTimesFeedbackPromptShown by intPreference(key, defaultValue)`

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -786,6 +789,62 @@ class Settings(private val appContext: Context) : PreferencesHolder {
...
+     */
+    @RequiresApi(Build.VERSION_CODES.O)
+    val shouldShowUserFeedbackPrompt: Boolean =
+        numTimesOpenedAfterInstall >= 5
+                && isDefaultBrowser()
+                && (hasBeenFourMonthsSince(timeWhenPromptWasLastShown))
+
+    @RequiresApi(Build.VERSION_CODES.O)
+    private fun hasBeenFourMonthsSince(timeWhenPromptWasLastShown: Int): Boolean {
+        val numDays = LocalDate.now().dayOfYear - timeWhenPromptWasLastShown
```
boek, line 844: We might want to account the year into this as well otherwise we might get stuck if gets set late into the year

liuche:
Breaking lint: https://firefoxci.taskcluster-artifacts.net/EsqYaefoRoaFRVFyuBTPmg/0/public/reports/lint-results-debug.html

eliserichards:
Upgrading from v1.3.3 to the newest, v1.3.9, [would fix](https://github.com/Kotlin/kotlinx.coroutines/issues/2004) the lint-lint error: 
```
../../../../../fetches/android-gradle-dependencies/jcenter/org/jetbrains/kotlinx/kotlinx-coroutines-core/1.3.6/kotlinx-coroutines-core-1.3.6.jar: Invalid package reference in library; not included in Android: java.lang.instrument. Referenced from kotlinx.coroutines.debug.AgentPremain.
```

Updating the coroutines version will break Glean, so we are using the workaround provided to disable 'InvalidPackage' in the `lintOptions`. This is a known issue for the Glean team but not priority at this time (pending GH/bugzilla ticket).

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
mcomella: Yes.

Note: You can also add tasks from anywhere in the codebase by accessing the queue through Components but this doesn't seem like a use case for it.

PR #14137, opened 2020-08-24
Replace setPadding with updatePadding

NotWoods:
This makes code easier to read when its just setting padding on one side.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/FenixSnackbar.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/FenixSnackbar.kt
@@ -12,6 +12,7 @@ import android.widget.FrameLayout
 import androidx.appcompat.content.res.AppCompatResources
 import androidx.appcompat.widget.ContentFrameLayout
 import androidx.coordinatorlayout.widget.CoordinatorLayout
+import androidx.core.view.updatePadding
 import androidx.core.widget.TextViewCompat
 import com.google.android.material.snackbar.BaseTransientBottomBar
 import com.google.android.material.snackbar.ContentViewCallback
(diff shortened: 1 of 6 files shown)


person808:
core-ktx has updatePadding and updatePaddingRelative. https://developer.android.com/reference/kotlin/androidx/core/view/package-summary#(android.view.View).updatePadding(kotlin.Int,%20kotlin.Int,%20kotlin.Int,%20kotlin.Int)

NotWoods:
I didn't know about updatePadding, that makes more sense to use here. The 0 values are really just used as defaults.

buildSrc/src/main/java/Dependencies.kt
```diff
@@ -15,7 +15,7 @@ object Versions {
     const val androidx_appcompat = "1.2.0-rc01"
     const val androidx_biometric = "1.1.0-alpha01"
     const val androidx_coordinator_layout = "1.1.0-rc01"
-    const val androidx_constraint_layout = "2.0.0-beta6"
```
eliserichards, line 18: I'm guessing there was a reason that we were using the beta version here. Can you look for the blame of that and double check that it's ok to downgrade to the stable version?

settings.gradle
```diff
@@ -64,7 +64,7 @@ if (localProperties != null) {
         // As above, hacks to execute .py files on Windows.
         def publishAcCmd = [];
         if (System.properties['os.name'].toLowerCase().contains('windows')) {
-            publishAcCmd << "py";
+            publishAcCmd << "python3";
```
eliserichards, line 67: What's this for? Can we move to a separate PR if it's not relevant to the padding changes?

buildSrc/src/main/java/Dependencies.kt
NotWoods: Oops, I didn't mean to put this change in this PR

settings.gradle
NotWoods: Same here 😅

Issue #14168, opened 2020-08-25
[Bug] Cannot easily click on migration button


jonalmeida:
## Steps to reproduce
1. Start a migration from Fennec to Fenix.
2. When migration is complete, click on the edge of the migration button.

### Expected behavior
- The button is clicked.

### Actual behavior
- The button is not clicked - you have to click on the 'Start Firefox Preview' text.

<img width="317" alt="Screen Shot 2020-08-25 at 10 32 21 AM" src="https://user-images.githubusercontent.com/1370580/91187756-83f0ea00-e6be-11ea-9100-cfab597c9a15.png">

### Device information

* Android device: Android 10, Pixel 4
* Fenix version: Nightly 25/08/2020

ebalazs-sv:
This issue is reproducible on RC 80.1.0 GV 80, from 8/25 with Pixel 2 (Android 9).
I will keep the qa:needed label to verify it on the next RC migration build.

ebalazs-sv:
This issue is still reproducible on RC 80.1.2 GV 80.0.1 from 9/1 with Motorola Moto G6 (Android 8).

jonalmeida:
Adding the uplift label to the issue instead of the PR in case there is another minor for 80.

jonalmeida:
cc: @liuche

jonalmeida:
This will be in the 81 release instead.

ebalazs-sv:
Verified as fixed on Beta 81.1.0-beta.2 (Build #2015761657), GV 81.0, from 9/2 with Pixel 2 (Android 9) and Samsung Galaxy Tab S3 (Android 8).

PR #14169, opened 2020-08-25
For #14168: Fix click listener on migration button

jonalmeida:
Move the click listener back onto the button instead of the text.


pocmo (approved):

boek:
Does this have to be uplifted @pocmo @jonalmeida?

jonalmeida:
@boek yeah, for the 80 release, it would be good to get this fixed.

Issue #13228, opened 2020-08-03
[Bug] Increased memory usage when using private mode (unknown MeasurementBrokerService)


Scripterr:
## Steps to reproduce

1. Open tab in private mode.
2. Background the browser.
3. Go to Developer Options -> Running services, and click on the Fenix app.
4. Observe the list of processes.
5. Connect to a Sync account.
6. Send a tab from a desktop to this device.
7. Observe the device notifications.

### Expected
 - At 4, there should be no `MessurementBrokerService` listed.
 - At 7, a send tab notification should be received immediately.

### Actual
 - There is a `MeasurementBrokerService` that is not part of Fenix.

### Device information

* Android device: 8.0.0
* Fenix version: latest nightly

### Screenshot
![Screenshotgsksj_1](https://user-images.githubusercontent.com/42364681/89198840-6783f500-d5cb-11ea-953a-c29e1cc692c3.png)

### Additional info
Also when accessing running services from developer settings there is some weird service running(see screenshot above) within fenix when using private browsing and the app also apears to be running in background whereas other browsers don't seem to be running when backgrounded when using their private browsing.

ekager:
From a quick search this seems related to the private session notification service.

mcomella:
I'm unfamiliar with GPS using so much memory and this could be a bug.

However, I would also expect private browsing mode to use more memory than normal browsing mode because in private browsing nothing is saved to disk but rather kept in memory. Other browsers might continue to write files to disk (and later clear them, hopefully) which could create a discrepancy in memory usage between us and them.

@snorp Do you have familiarity with this behavior?

ekager:
To me it looks like something like this is more likely because of the very specific process reported https://stackoverflow.com/a/42162217

@jonalmeida are we correctly disabling the firebase analytics? In the docs it looks like we are using the temporary disabling vs the permanent one? - https://firebase.google.com/docs/analytics/configure-data-collection?platform=android#permanently_deactivate_collection

jonalmeida:
@ekager I'm not sure what the actual differences are between permanent and temporary, but we could try the permanent version to see if it solves this issue.

ekager:
Not an expert but based on the SO post looks like it's related to the interaction with the persistent notification we show in PBM 🤷🏼

opusforlife2:
[...]

PR #14186, opened 2020-08-25
For #13228: Removes Firebase service dependencies from all builds

jonalmeida:
We no longer have the MeasurementBrokerService running when the private mode service is running. I've done some simple testing to ensure that WebPush (Firebase) still functions, but it would be better if we could get QA to test this out.

We should also not land this until after the beta release is cut so that it can ride the trains.

Screen Shot 2020-08-25 at 3 41 59 PM

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -172,6 +172,12 @@ android {
     }
 }
 
+configurations {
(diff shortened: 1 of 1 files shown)


jonalmeida:
We don't need this - I'll comment in the issue on the reasoning.

PR #14219, opened 2020-08-26
Add diagnostic breadcrumbs for debugging "Display already aquired" crashes

pocmo:
For: https://github.com/mozilla-mobile/android-components/issues/7960

The additional error logging we added in AC shows that we have multiple `HomeActivitiy` that try to render the same `Session`. With the additional breadcrumbs added here, we hope to get a better idea of where the duplicated activities are coming from and what their lifecycle is (what intent action was used to launch them? are they getting recreated? ...)

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/ext/Activity.kt
+++ b/app/src/main/java/org/mozilla/fenix/ext/Activity.kt
@@ -7,6 +7,7 @@ package org.mozilla.fenix.ext
 import android.app.Activity
 import android.view.View
 import android.view.WindowManager
+import mozilla.components.support.base.crash.Breadcrumb
 
 /**
  * Attempts to call immersive mode using the View to hide the status bar and navigation buttons.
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/mozilla/fenix/ext/Activity.kt
```diff
@@ -22,3 +23,20 @@ fun Activity.enterToImmersiveMode() {
             or View.SYSTEM_UI_FLAG_FULLSCREEN
             or View.SYSTEM_UI_FLAG_IMMERSIVE_STICKY)
 }
+
+fun Activity.breadcrumb(
+    message: String,
+    data: Map<String, String> = emptyMap()
+) {
+    val activityName = this::class.simpleName ?: return
```
csadilek, line 31: Does `this:class.simpleName` work in a production build without kotlin reflection? We had to do this to get to the actual class name in migration: https://github.com/mozilla-mobile/android-components/pull/7897

`this::class.java` or `this.javaclass` work fine not needin Kotlin reflection.

app/src/main/java/org/mozilla/fenix/ext/Fragment.kt
```diff
@@ -57,3 +58,28 @@ fun Fragment.redirectToReAuth(destinations: List<Int>, currentDestination: Int?)
         findNavController().popBackStack(R.id.savedLoginsAuthFragment, false)
     }
 }
+
+fun Fragment.breadcrumb(
+    message: String,
+    data: Map<String, String> = emptyMap()
+) {
+    val fragmentName = this::class.simpleName ?: return
```
csadilek, line 66: See q above.

app/src/main/java/org/mozilla/fenix/ext/Fragment.kt
```diff
@@ -57,3 +58,28 @@ fun Fragment.redirectToReAuth(destinations: List<Int>, currentDestination: Int?)
...
+    val activityName = if (activity != null) {
+        activity!!::class.simpleName.toString()
```
csadilek, line 68: See q above.

PR #14278, opened 2020-08-27
For #13765 - Fix TopSites item sizes

### LINKED ISSUE #13765: Top sites are more prominent on the homescreen
Based on a large number of user feedback, a lot of our users are finding it read the title of top sites as they are too small. Especially when a user have 2 links saved from the same domain (similar favicon) they find it difficult to distinguish between the two. Users have also expressed that the top site icons are too small for something that is supposed to help them get to where they need faster, whereas the "collections" CTA/placeholder is much larger in proportion. 

## User Story
- As a user, I want my top sites to look more prominent on my homescreen so they are more useful to me.

### Acceptance Criteria
- I can easily read the title of a top sites and distinguish between two top sites from the same domain.
- The size of the top site icons are larger and more prominent than the collections CTA/placeholder. 

### UX
- Specs for the top sites section, and for the updated favicon below
- The Top Sites section should have 4 top sites in a row. This should also allow the full 'Top Articles' text to be visible for the default Pocket top site.
- The favicon size should be increased to 32x32px, leaving 4px of padding inside the favicon container

**Top Sites**
Home - Spec

**Top Site Favicon**
Favicon

codrut-topliceanu:
Fixes #13765 

### Pull Request checklist

1. click on Show All Checks,

I followed the specs in #13765 as best as I could, however the distance between columns needs to be variable(X value in the screenshot) otherwise wider devices will only have top sites on the left side of their screens. I have also added 12dp spacing between rows as it felt appropriate. You can find further screenshots from a Pixel 3XL (for both orientations) below, please let me know if any more changes are required. 
Screenshot 2020-08-27 at 12 14 15

device-2020-08-27-123731

### DIFF
--- a/app/src/main/res/values/styles.xml
+++ b/app/src/main/res/values/styles.xml
@@ -614,6 +614,14 @@
         <item name="behavior_halfExpandedRatio">0.4</item>
     </style>
 
+    <style name="Mozac.Widgets.TopSite.Favicon" parent="Mozac.Widgets.Favicon">
+        <item name="android:layout_width">@dimen/top_sites_favicon_size</item>
+        <item name="android:layout_height">@dimen/top_sites_favicon_size</item>
+        <item name="android:scaleType">fitCenter</item>
+        <item name="android:padding">@dimen/top_sites_favicon_padding</item>
+        <item name="android:background">@drawable/mozac_widget_favicon_background</item>
+    </style>
+
     <style name="TabTrayFab" parent="Widget.MaterialComponents.ExtendedFloatingActionButton">
         <item name="elevation">90dp</item>
         <item name="android:stateListAnimator">@null</item>
--- a/app/src/main/res/values/styles.xml
+++ b/app/src/main/res/values/styles.xml
@@ -614,7 +614,7 @@
         <item name="behavior_halfExpandedRatio">0.4</item>
     </style>
 
-    <style name="Mozac.Widgets.TopSite.Favicon" parent="Mozac.Widgets.Favicon">
+    <style name="TopSite.Favicon" parent="Mozac.Widgets.Favicon">
         <item name="android:layout_width">@dimen/top_sites_favicon_size</item>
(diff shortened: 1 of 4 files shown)


liuche:
Hi @codrut-topliceanu do we think this could be done in time to make the "late" cut on Tuesday? thank you!

codrut-topliceanu:
I've made the requested changes. @apbitner could you please take a look and tell me if they are ok? (I've also tested this on ldpi and xhdpi emulators/devices and everything seems in order)

landscape top sites

app/src/main/res/values/styles.xml
```diff
@@ -614,6 +614,14 @@
         <item name="behavior_halfExpandedRatio">0.4</item>
     </style>
 
+    <style name="Mozac.Widgets.TopSite.Favicon" parent="Mozac.Widgets.Favicon">
```
Mugurell, line 617: IMO mozac is still kept in resources names only when we intend to overwrite the value from AC.
Here we extend an AC style, but the above need does not exist.
Could we make it our own style with a simpler name, without "mozac"? Thoughts?

Mugurell (approved):
Looks good, thanks!

app/src/main/res/values/styles.xml
gabrielluong: Agreed with the above. I think it probably makes sense to drop the Mozac to something simpler in Fenix, and only keep to Mozac naming within AC.

gabrielluong: I think this is resolved.

softvision-miralobontiu:
Hi all,
I've noticed on the latest Beta 81.1.0-beta.1, and on 9/2 Nightly, on the Samsung Galaxy Tab S6 (Android 9), and on Samsung Galaxy Tab S3 (Android 8) that the default 3 top sites are not centralized. Is this expected?
Thanks,

Untitled

Mugurell:
They were already aligned to the left.
This patch changed nothing in this regard.

PR #14286, opened 2020-08-27
Bug 1654346 Add YTP 2020 tests

alexandru-io:


taskcluster/ci/browsertime/kind.yml
```diff
@@ -244,3 +244,44 @@ jobs:
                 armeabi-v7a:
                     - '--test-url-params=exclude=1,2,9,10,17,18,21,22,26,28,30,32,39,40,47,48,55,56,63,64,71,72,79,80,83,84,89,90,95,96'
                 default: []
+
+    youtube-playback-av1-sfr:
+        test-name: youtube-playback-av1-sfr
+        run-visual-metrics: False
```
gmierz, line 250: You need to add this to all the other tasks too.

taskcluster/ci/raptor/kind.yml
```diff
@@ -96,3 +96,45 @@ jobs:
                 armeabi-v7a:
                     - '--test-url-params=exclude=1,2,9,10,17,18,21,22,26,28,30,32,39,40,47,48,55,56,63,64,71,72,79,80,83,84,89,90,95,96'
                 default: []
+
+    youtube-playback-av1-sfr:
```
gmierz, line 100: These don't need to be added to the raptor tests.

taskcluster/ci/browsertime/kind.yml
```diff
@@ -244,3 +244,44 @@ jobs:
                 armeabi-v7a:
                     - '--test-url-params=exclude=1,2,9,10,17,18,21,22,26,28,30,32,39,40,47,48,55,56,63,64,71,72,79,80,83,84,89,90,95,96'
                 default: []
+
+    youtube-playback-av1-sfr:
```
gmierz, line 248: We should make some changes in the transform to only run this AV1 test on pixel 2.

taskcluster/ci/browsertime/kind.yml
```diff
@@ -244,3 +244,44 @@ jobs:
...
+        treeherder:
+            symbol: Btime(ytp-h264-sfr)
+
+    youtube-playback-h264:
+        description: "Raptor YouTube Playback H264 on Fenix"
+        test-name: youtube-playback-h264
+        treeherder:
+            symbol: Btime(ytp-h264)
+
+    youtube-playback-vp9-sfr:
```
gmierz, line 265: This should only run on the P2 (like above).

taskcluster/ci/browsertime/kind.yml
```diff
@@ -244,3 +244,44 @@ jobs:
...
+        treeherder:
+            symbol: Btime(ytp-widevine-hfr)
+
+    youtube-playback-widevine-h264-sfr:
+        description: "Raptor YouTube Playback Widevine H264 SFR on Fenix"
+        test-name: youtube-playback-widevine-h264-sfr
+        treeherder:
+            symbol: Btime(ytp-widevine-h264-sfr)
+
+    youtube-playback-widevine-vp9-sfr:
```
gmierz, line 283: This should only run on P2.

taskcluster/fenix_taskgraph/transforms/browsertime.py
```diff
@@ -123,3 +123,22 @@ def fill_email_data(config, tasks):
...
         yield task
+
+
+@transforms.add
+def filter_ytp_tests(config, tasks):
+    """Youtube-playback AV1, H264 and VP9 tests should run on P2 only"""
+    for task in tasks:
+        test_platform = task['test-platform']
+        test_name = task["test-name"]
+        if test_platform.startswith('android-hw-p2'):
```
gmierz, line 134: With this transform we will only schedule AV1, H264, and VP9 tests on P2. All the other tests will be skipped.

You will need to let all tests for P2 through, but restrict the tests that can run on the G5 (all but VP9 and AV1 tests).

gmierz (changes requested):

taskcluster/fenix_taskgraph/transforms/browsertime.py
alexandru-io: In the comments above you said that AV1, H264, and VP9 tests should only run on P2. Did something changed?
If the device is not P2, it won't get into the `if` but it will be yielded by the last line of the function.

gmierz: Nothing changed - you misunderstood the comment. Those test variants should only run on P2 and not the G5. What you implemented here is "P2 should only run these X tests" rather than "these X tests should only run on P2"

taskcluster/fenix_taskgraph/transforms/browsertime.py
```diff
@@ -124,3 +124,18 @@ def fill_email_data(config, tasks):
...
         yield task
+
+
+@transforms.add
+def filter_ytp_tests(config, tasks):
+    """Youtube-playback AV1, H264 and VP9 tests should run on P2 only"""
+    for task in tasks:
+        test_platform = task['test-platform']
+        test_name = task["test-name"]
+        if "AV1" in test_name or "H264" in test_name or "VP9" in test_name:
```
gmierz, line 135: H264 tests should be fine to run on G5, we have them running in raptor atm. You can remove that restriction: https://treeherder.mozilla.org/#/jobs?repo=mozilla-central&tier=1%2C2%2C3&searchStr=youtube&revision=2cca9cb89b46d67e1e25051495cb89f71b856c32

gmierz (changes requested):

gmierz:
Running tests here: https://github.com/mozilla-mobile/fenix/pull/15157

gmierz:
Tests failed: failed: https://github.com/mozilla-mobile/fenix/pull/15157/checks?check_run_id=1129102894

alexandru-io:
@gmierz The tests seems like failing but they look similar to the ones for YTP 2019 so I guess we're good.
https://github.com/mozilla-mobile/fenix/pull/15256

gmierz:
We're certainly not good to go. Please look at the errors at least: https://firefox-ci-tc.services.mozilla.com/tasks/UNI_Y0S3TMaT5xg6tEfKYg/runs/0/logs/https%3A%2F%2Ffirefox-ci-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FUNI_Y0S3TMaT5xg6tEfKYg%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive.log#L1285

alexandru-io:
That's because in `raptor.ini` is only `benchmarks/youtube-playback.ini`

taskcluster/fenix_taskgraph/transforms/browsertime.py
```diff
@@ -124,3 +124,32 @@ def fill_email_data(config, tasks):
...
         yield task
+
+
+@transforms.add
+def filter_ytp_tests(config, tasks):
+    """Youtube-playback AV1 and VP9 tests should run on P2 only"""
+    for task in tasks:
+        test_platform = task['treeherder']['platform']
+        test_name = task['name']
+        if "AV1" in test_name or "VP9" in test_name:
```
gmierz, line 135: Looking at the tests, these AV1 and VP9 tests are passing on both devices, we can remove this restriction now.

Also, note that this transform didn't work because after a `yield` call, the code returns to the same point when we are generating the next task. You need to call `continue` after the `yield` if you don't want it to continue.

taskcluster/fenix_taskgraph/transforms/browsertime.py
```diff
@@ -124,3 +124,32 @@ def fill_email_data(config, tasks):
...
+            if 'android-hw-p2' in test_platform:
+                yield task
+            else:
+                continue
+
+        # Bug 1668537 The following tests are failing intermittently
+        # and are disabled until the error is fixed
+        if "WIDEVINE-H264" in test_name or "WIDEVINE-H264-HFR" in test_name:
```
gmierz, line 143: Why is this capitalized? If there are issues with capitalization you should compare a lower case string to `test_name.lower()`. Also, I don't think we need a split below because it doesn't work on any of the devices, so we should change all that to a continue.

taskcluster/fenix_taskgraph/transforms/browsertime.py
```diff
@@ -124,3 +124,32 @@ def fill_email_data(config, tasks):
...
+            if 'android-hw-g5' in test_platform:
+                yield task
+            else:
+                continue
+
+        if "WIDEVINE-HFR" in test_name or "WIDEVINE-VP9-SFR" in test_name:
```
gmierz, line 149: Same as above.

gmierz (changes requested):
Some minor things to fix up below.

taskcluster/fenix_taskgraph/transforms/browsertime.py
gmierz: Actually, you can combine this condition with the one above by checking if `widevine` is in the test name.

taskcluster/fenix_taskgraph/transforms/browsertime.py
alexandru-io: I guess i can remove the transform method at all.

gmierz: Yes, let's remove it completely.

gmierz (approved):

gmierz:
@alexandru-io please remove the github-pull-requests entries before we land.

@JohanLorenzo can you review this and merge if it's good to you?

JohanLorenzo (approved):
Changes look good to me from a taskgraph perspective!

PR #14308, opened 2020-08-27
Extract checkbox selected mask

NotWoods:
Making this available to bookmarks later.

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/tabtray/FenixTabsAdapter.kt
+++ b/app/src/main/java/org/mozilla/fenix/tabtray/FenixTabsAdapter.kt
@@ -8,6 +8,7 @@ import android.content.Context
 import android.view.LayoutInflater
 import android.view.View
 import androidx.core.view.isVisible
+import kotlinx.android.synthetic.main.checkbox_item.view.*
 import kotlinx.android.synthetic.main.tab_tray_item.view.*
 import mozilla.components.browser.tabstray.TabViewHolder
 import mozilla.components.browser.tabstray.TabsAdapter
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mozilla/fenix/tabtray/FenixTabsAdapter.kt
```diff
@@ -104,7 +105,6 @@ class FenixTabsAdapter(
     private fun showCheckedIfSelected(tab: Tab, view: View) {
         val shouldBeChecked =
             mode is TabTrayDialogFragmentState.Mode.MultiSelect && selectedItems.contains(tab)
-        view.checkmark.isVisible = shouldBeChecked
```
liuche, line 107: does the checkmark visibility not need to be set anymore to match `shouldBeChecked` with this change? I don't see anything else updating its checked state. Was it unnecessary in the first place?

liuche (commented):
Makes sense, literally pulling a layout into

app/src/main/java/org/mozilla/fenix/tabtray/FenixTabsAdapter.kt
NotWoods: Previously the layout tree looked like:
```
- parent
  |- image
  |- mask
  |- checkmark
```

Now it looks like:
```
- parent
  |- image
  |- mask
     |- checkmark
```

Since the checkmark's parent is hidden we don't need to hide the checkmark itself.

liuche (approved):

Issue #10069, opened 2020-04-20
Add telemetry probe for startup time


mcomella:
We should ensure this probe matches how we measure startup time in CI testing, if possible.

- - -
We repurposed https://github.com/mozilla-mobile/fenix/issues/8803 so this probe is to revisit the telemetry probe for startup. Now that we better know how to write telemetry, we probably want to do the minimal thing here rather than the large number of steps mentioned below.

### Why/User Benefit/User Problem
Startup time is an oft cited reason for friction and churn. We need to understand how long it takes Fenix to startup for our users so that we can prioritize work to address regressions and outliers that result in decreased engagement or churn. Local and CI-based testing provides an incomplete picture of how Fenix performs at startup.

### Impact
Improved understanding of how Fenix performs on startup for our users.

### Acceptance Criteria (how do I know when I’m done?)
We should make a best-effort to _incrementally_ instrument as much of startup as possible - until we hit diminishing returns - because it's impossible to do it all. Unfortunately, it will be difficult to define what this looks like in advance so take this as a guiding principle. Msg mcomella if this does not make sense.

Some incremental steps we should try to implement for GA:

* ~~Add a metric that tracks the start time of GV~~ I believe this is already available in Gecko telemetry
* Add a metric that tracks the duration of `FenixApplication.onCreate` (this may or may not be helpful but we want to add it to see if it is in practice)
* Add a metric that tracks `FenixApplication.onCreate` until first frame drawn
* Distinguish between first run and other cold starts (don't let this impact performance!)
* (is this reasonable?) Distinguish between which page was shown, i.e. which startup path was used
* Time from first frame drawn to reportFullyDrawn? Essentially, capture loading top sites & open tabs because this is dependent on content quantity

Some incremental steps we probably can't implement by GA but we should verify:

* Add a probe that tracks `FenixApplication.onCreate` until visual completeness
* Add warm/hot start times, distinguish from cold start
* Distinguish between cold starts after upgrading and other cold starts

Here is ecsmyth's ideal startup time probe that inspired these incremental steps:
> Glean telemetry probe landed in Fenix release that measures time, as closely as is practical, from when user initiates startup until the app is started and the initial screen is visually complete. The probe should differentiate between hot, cold, and warm startup scenarios and account for differences in the first page rendered (e.g., how desktop measures first paint and first paint of about:home)

sraturi:
TLDR from a conversation with Glean:

[...]

PR #14375, opened 2020-08-28
for #10069 added AppLaunchTime.kt to handle logic of startup time for cold, warm, and hot startup types.

sraturi:


sraturi:
**TLDR from a conversation with Glean:** 

It is okay for now to report the launch time as an event extra key (type safety from Fenix side). There is some work done on glean to have t[ype safety for extra_keys](https://bugzilla.mozilla.org/show_bug.cgi?id=1613944). Once that is finished, we can implement type safety for this metric (given we don't change metric or keys name) without disrupting data flow. 
There is also a [struct metric type](https://bugzilla.mozilla.org/show_bug.cgi?id=1629950) (not implemented yet) that could be a good alternative for this. However, there will be data disruption if we change the metric type in the future

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -305,6 +318,21 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
...
+        // remove predraw listener for 2 reasons
+        // 1. the application is shutting down and we want to avoid memory leak.
+        // 2. in case the application is also going in the background (instead of just the activity),
+        // we need to stop listening to the preDraw listener because if we have onPreDraw listener
+        // attached and the application comes foreground again, the activity will invoke onPreDraw()
+        // before onRestart and we want to avoid that.
+        rootContainer.viewTreeObserver.removeOnPreDrawListener(components.appStartupTelemetry)
+
+        // reset the startup time to avoid any weird state.
+        components.appStartupTelemetry.resetStartupTime()
```
MarcLeclair, line 335: I think the comment is kinda redundant the name is self explanatory 😄

app/src/main/java/org/mozilla/fenix/components/metrics/AppLaunchTimeMeasurement.kt
```diff
@@ -0,0 +1,59 @@
...
+class AppLaunchTimeMeasurement(private val stats: Stat = Stat()) {
+
+    // when user launched the time. i.e press app_icon, launched a link.
+    private var processLaunchTimeStampNanoTime: Long? = null
+    // we are considering screen to be visible after the first pre draw call.
+    private var elapsedNanoTimestampOnHomeActivityPreDraw: Long? = null
+
+    fun onApplicationOnCreate() {
+        // There is disk read involve and we dont want to block the main thread.
+        GlobalScope.launch(Dispatchers.IO) {
```
MarcLeclair, line 26: I think avoiding to use `GlobalScope` is ideal. The coroutine could hang if ever something happens with `stats` or reading from the file. Maybe using `MainScope()` ?

app/src/main/java/org/mozilla/fenix/components/metrics/AppLaunchTimeMeasurement.kt
```diff
@@ -0,0 +1,59 @@
...
+            processLaunchTimeStampNanoTime = stats.getProcessStartTimeStampNano(Process.myPid())
+        }
+    }
+
+    fun onHomeActivityOnCreate() {
+        if (processLaunchTimeStampNanoTime == null) {
+            // if timestamp was not set, this is likely a warm
+            processLaunchTimeStampNanoTime = homeActivityInitTimeStampNano
```
MarcLeclair, line 34: I think the comment is a bit vague (could do without I think). Imo,  since this function only serves one purpose, maybe name it to be more self-explanatory such as `onHomeActivityOnCreateIsColdStart` (or something a big easier to understand).

app/src/main/java/org/mozilla/fenix/components/metrics/AppLaunchTimeMeasurement.kt
```diff
@@ -0,0 +1,59 @@
...
+        }
+    }
+
+    fun onHomeActivityOnRestart() {
+        // for hot startup, we know this to be the earliest function.
+        processLaunchTimeStampNanoTime = SystemClock.elapsedRealtimeNanos()
```
MarcLeclair, line 40: Same thing for the comment. I think the `OnRestart` explains that this is for the restart of the activity

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -114,6 +116,13 @@ import java.lang.ref.WeakReference
 @SuppressWarnings("TooManyFunctions", "LargeClass")
 open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
 
+    init {
+        // we need to store startup timestamp for warm startup. we cant directly store
+        // inside AppStartupTelemetry since that class lives inside components and
+        // components requires context to access. we store it as a static member
+        AppLaunchTimeMeasurement.homeActivityInitTimeStampNano = SystemClock.elapsedRealtimeNanos()
```
MarcLeclair, line 123: So I'm not a fan  of static variables, not sure how feasible this is for you, but how about getting the `init` time in here and passing it when we have access to `components` ?

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
sraturi: I guess I was trying to explain as to why im reseting it.

app/src/main/java/org/mozilla/fenix/components/metrics/AppLaunchTimeMeasurement.kt
sraturi: Hmm, good point. I guess I always associated the main scope with dealing with stuff related to UI.  Also, I thought Mainscope had same lifetime as GlobalScope? Maybe keep track of the job and if its not done when needed, cancel it?

MarcLeclair: Ah yes it is. but you can still launch form a different dispatchers IIRC so `MainScope().launch(Dispatchers.IO)` but preferably would be to have our own scope 😄

app/src/main/java/org/mozilla/fenix/components/metrics/AppLaunchTimeMeasurement.kt
sraturi: Hmm good point. I can make it more clear? This function will be called for both cold and warm startup so I was having a hard time to find a good name.

sraturi:
@MarcLeclair ready for round 2 review

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -236,13 +244,28 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
     }
 
     protected open fun startupTelemetryOnCreateCalled(safeIntent: SafeIntent, hasSavedInstanceState: Boolean) {
-        components.appStartupTelemetry.onHomeActivityOnCreate(safeIntent, hasSavedInstanceState)
+        components.appStartupTelemetry.onHomeActivityOnCreate(
+            safeIntent,
+            hasSavedInstanceState, homeActivityInitTimeStampNanoSeconds
+        )
+
+        OneShotPreDrawListener.add(rootContainer) {
```
mcomella, line 252: Wouldn't it be better to call this in `onHomeActivityOnCreate`? That way we're not splitting up the logic and minimizing low-level code in an Activity class.

Also, `rootContainer.doOnPreDraw { ... }` is more natural than this invocation style.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -236,13 +244,28 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
...
     override fun onRestart() {
         super.onRestart()
 
+        startupTelemetryOnRestartCalled()
+    }
+
+    private fun startupTelemetryOnRestartCalled() {
         components.appStartupTelemetry.onHomeActivityOnRestart()
+
+        OneShotPreDrawListener.add(rootContainer) {
```
mcomella, line 266: Same: can we call this in `onRestart`?

app/src/main/java/org/mozilla/fenix/perf/Stat.kt
```diff
@@ -64,6 +64,11 @@ open class Stat {
         return getStatText(pid).split(' ')[FIELD_POS_STARTTIME].toLong()
     }
 
+    fun getProcessStartTimeStampNano(pid: Int): Long {
+        val timeInTicks = getStatText(pid).split(' ')[FIELD_POS_STARTTIME].toLong()
```
mcomella, line 68: nit: This is duplication: can we call `getProcessStartTimeTicks()` and convert the result?

app/src/main/java/org/mozilla/fenix/customtabs/ExternalAppBrowserActivity.kt
```diff
@@ -36,7 +38,13 @@ open class ExternalAppBrowserActivity : HomeActivity() {
     final override fun getIntentSessionId(intent: SafeIntent) = intent.getSessionId()
 
     override fun startupTelemetryOnCreateCalled(safeIntent: SafeIntent, hasSavedInstanceState: Boolean) {
-        components.appStartupTelemetry.onExternalAppBrowserOnCreate(safeIntent, hasSavedInstanceState)
+        components.appStartupTelemetry.onExternalAppBrowserOnCreate(
+            safeIntent,
+            hasSavedInstanceState, homeActivityInitTimeStampNanoSeconds
+        )
+        OneShotPreDrawListener.add(rootContainer) {
```
mcomella, line 45: Same, can we embed this in the other method?

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -236,13 +244,28 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
     }
 
     protected open fun startupTelemetryOnCreateCalled(safeIntent: SafeIntent, hasSavedInstanceState: Boolean) {
-        components.appStartupTelemetry.onHomeActivityOnCreate(safeIntent, hasSavedInstanceState)
+        components.appStartupTelemetry.onHomeActivityOnCreate(
+            safeIntent,
+            hasSavedInstanceState, homeActivityInitTimeStampNanoSeconds
```
mcomella, line 249: nit: kotlin convention is either have the args all on one line or all the args on different lines

app/src/main/java/org/mozilla/fenix/components/metrics/AppStartupTelemetry.kt
```diff
@@ -143,4 +154,16 @@ class AppStartupTelemetry(private val metrics: MetricController) : LifecycleObse
...
+    /**
+     * We record metric on onPreDraw() since we need to wait for the application to draw the
+     * first frame. which will always happen after onResume. The hot startup time is a bit complicated
+     * because if we have onPreDraw listener attached, Activity will invoke onPreDraw() before onRestart.
+     * For this reason, we need to make sure that we remove the preDraw listener when activity is going
+     * in the background and reattach it during onRestart().
+     */
+    fun onPreDraw() {
+        appLaunchTimeMeasurement.onActivityOnPreDraw()
+        recordMetric()
```
mcomella, line 167: We shouldn't be doing such expensive computation in a pre-draw listener as this may cause us to drop frames: can we post it to run later or could this create invalid data? (I guess if we did it functionally and passed all the current values (assuming they're immutable) into the function as args, this wouldn't be a problem though – see why it's pre-emptively good to make things immutable 😄?).

app/src/main/java/org/mozilla/fenix/components/metrics/AppLaunchTimeMeasurement.kt
```diff
@@ -0,0 +1,51 @@
...
+class AppLaunchTimeMeasurement(private val stats: Stat = Stat()) {
+
+    // when user launched the time. i.e press app_icon, launched a link.
+    private var processLaunchTimeStampNanoSeconds: Long? = null
+    // we are considering screen to be visible after the first pre draw call.
+    private var elapsedNanoTimestampOnHomeActivityPreDraw: Long? = null
+
+    fun onApplicationOnCreate() =
+        MainScope().launch(Dispatchers.IO) {
+        processLaunchTimeStampNanoSeconds = stats.getProcessStartTimeStampNano(Process.myPid())
```
mcomella, line 26: There is a race condition: this may not be set by the time we record the metric (and confusingly `ActivityonCreate` though the end result should be the same).

app/src/main/java/org/mozilla/fenix/components/metrics/AppLaunchTimeMeasurement.kt
```diff
@@ -0,0 +1,51 @@
...
+import kotlinx.coroutines.launch
+import org.mozilla.fenix.perf.Stat
+
+/**
+ * Handles the logic of figuring out launch time for cold, warm and hot startup.
+ */
+class AppLaunchTimeMeasurement(private val stats: Stat = Stat()) {
+
+    // when user launched the time. i.e press app_icon, launched a link.
+    private var processLaunchTimeStampNanoSeconds: Long? = null
```
mcomella, line 20: It's kind of confusing that we're re-using this variable to mean many things:
- processLaunch time for COLD
- activity init time for WARM
- onRestart time for HOT

Why don't we just set them as different variables? Before reporting a value, we then have the ability to validate that only one value is set (as we'd expect) and report an error if not. This would make the code a lot more explicit: right now it's very unclear which lifecycle method refers to which startup case and why we set this value so many times.

app/src/main/java/org/mozilla/fenix/components/metrics/AppLaunchTimeMeasurement.kt
```diff
@@ -0,0 +1,51 @@
...
+
+    fun onActivityOnPreDraw(currentTimeNanoSeconds: Long = SystemClock.elapsedRealtimeNanos()) {
+        elapsedNanoTimestampOnHomeActivityPreDraw = currentTimeNanoSeconds
+    }
+
+    /**
+     * if we have both start and finish time for launch, return it otherwise we return null
+     */
+    fun getApplicationLaunchTime(): Long? {
+        return (elapsedNanoTimestampOnHomeActivityPreDraw?.minus(processLaunchTimeStampNanoSeconds!!))
```
mcomella, line 49: Why would elapsedNano every be null? Comment

app/src/main/java/org/mozilla/fenix/components/metrics/AppStartupTelemetry.kt
```diff
@@ -143,4 +154,16 @@ class AppStartupTelemetry(private val metrics: MetricController) : LifecycleObse
         // Therefore we set the isMetricRecorded flag to false.
         isMetricRecordedSinceAppWasForegrounded = false
     }
+
+    /**
+     * We record metric on onPreDraw() since we need to wait for the application to draw the
+     * first frame. which will always happen after onResume. The hot startup time is a bit complicated
+     * because if we have onPreDraw listener attached, Activity will invoke onPreDraw() before onRestart.
+     * For this reason, we need to make sure that we remove the preDraw listener when activity is going
```
mcomella, line 162: afaict, we never do this: is that a problem?

mcomella (commented):
Thanks sraturi! Did a skim of the production code: this looks like it works but I'm concerned about:
- readability of AppLaunchTimeMeasurement.kt 
- why is elapsedNano null?
- the race condition

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
sraturi: that would require us passing in the rootContainer inside the `onHomeActivityOnCreate`, which I was trying to avoid since I wanted to avoid passing in UI elements to the data class. maybe I misunderstood the suggestion?

Also `OneShotPreDrawListener` removes the listener after the first call, which is what we need. but if we were to use `rootContainer.doOnPreDraw { ... } `then we have the overhead of removing the listener ourself.

app/src/main/java/org/mozilla/fenix/perf/Stat.kt
sraturi: thanks! fixed it

app/src/main/java/org/mozilla/fenix/components/metrics/AppLaunchTimeMeasurement.kt
sraturi: hmm would it make sense to keep track of whether or not the application oncreate was called. and than during onPreDraw() launch a coroutine to collect this data?

app/src/main/java/org/mozilla/fenix/components/metrics/AppStartupTelemetry.kt
sraturi: see https://github.com/mozilla-mobile/fenix/pull/14375#discussion_r486831613

sraturi: id say it pretty important to set data at this point since user can quit or background application anytime . once app is backgrounded, we have to start recording a fresh metric

app/src/main/java/org/mozilla/fenix/components/metrics/AppStartupTelemetry.kt
sraturi: ahh another good reason to avoid comments :sweat_smile: I updated the listener and forgot about the comment. my apologies haha

app/src/main/java/org/mozilla/fenix/components/metrics/AppLaunchTimeMeasurement.kt
sraturi: fixed

app/src/main/java/org/mozilla/fenix/components/metrics/AppLaunchTimeMeasurement.kt
sraturi: I changed the code to align more with this

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
sraturi: fixed

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
sraturi: see https://github.com/mozilla-mobile/fenix/pull/14375#discussion_r486829128

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
mcomella: > which I was trying to avoid since I wanted to avoid passing in UI elements to the data class.

By data class, do you mean you wanted to avoid passing UI elements into the model? If so, that's an interesting point – I haven't been conceptualizing consider the split between model and UI in the app startup telemetry class. I've been thinking of it as an extension of the Activity – its responsibility is to record startup data, primarily by managing arguments and lifecycles to do it, i.e. no separation between model and UI. In my model, it makes sense to put ALL of the logic in App Startup Telemetry (or files only app startup telemetry references), to encapsulate it.

Plus, as you know, I hate adding more than a line or so per PR to Activity and Fragment methods. :)

But I'll let you decide – I'm not sure how you're designing the class.

> but if we were to use rootContainer.doOnPreDraw { ... } then we have the overhead of removing the listener ourself.

`doOnPreDraw` uses OneShot... under the hood so I don't think that's true. Note: a design philosophy of KTX functions (this is one) is that they never add functionality over what's in the framework/androidx: they just make it more natural to access in Kotlin.

mcomella (commented):
GH bug making it impossible to respond to outdated comments 🙄 so responding here:

> hmm would it make sense to keep track of whether or not the application oncreate was called. and than during onPreDraw() launch a coroutine to collect this data?

Yeah, I think so – fetch it just prior to submitting.

> id say it pretty important to set data at this point since user can quit or background application anytime . once app is backgrounded, we have to start recording a fresh metric

If we copy the data (which should be relatively simple with immutable objects – you just pass the reference rather than copying), we can send the metric slightly after this even if new data starts to get recorded.

app/src/main/java/org/mozilla/fenix/components/metrics/AppLaunchTimeMeasurement.kt
mcomella: Yeah, I think so – fetch it just prior to submitting.

app/src/main/java/org/mozilla/fenix/components/metrics/AppStartupTelemetry.kt
mcomella: If we copy the data (which should be relatively simple with immutable objects – you just pass the reference rather than copying), we can send the metric slightly after this even if new data starts to get recorded.

app/src/main/java/org/mozilla/fenix/components/metrics/AppLaunchTimeMeasurement.kt
```diff
@@ -0,0 +1,78 @@
...
+import kotlinx.coroutines.withContext
+import org.mozilla.fenix.perf.Stat
+
+/**
+ * Handles the logic of figuring out launch time for cold, warm and hot startup.
+ */
+class AppLaunchTimeMeasurement(private val stats: Stat = Stat()) {
+    private var isColdStartup = false
+    private var isWarmStartup = false
+    private var isHotStartup = false
```
mcomella, line 19: This is acceptable – no action needed unless you want to – but consider a red flag here: these states are mutually exclusive. We're using 3 booleans to represent one state value. A safer formation is:
```kotlin
// can we re-use this enum from somewhere else without breaking
// encapsulation?
enum class StartupMode { COLD, WARM, HOT, UNKNOWN }

private var mode = StartupMode.UNKNOWN

fun ...onCreate() {
    mode = StartupMode.COLD
}
```

This makes it more intuitive to the programmer when invalid states can happen, e.g. we can't get `isColdStartup == true && isWarmStartup == true`. And we can do fairly straightforward validation on the `mode` property – i.e. only UNKNOWN -> COLD/WARM/HOT or COLD/WARM/HOT -> UNKNOWN – rather than checking 3 booleans.

app/src/main/java/org/mozilla/fenix/components/metrics/AppLaunchTimeMeasurement.kt
```diff
@@ -0,0 +1,78 @@
...
+    private var isOnPreDrawCalled = false
+
+    private var applicationOnCreateTimeStampNanoSeconds: Long? = null
```
mcomella, line 22: No need to change now but for future reference, `applicationOnCreateNanos` would be acceptable and clear by convention.

app/src/main/java/org/mozilla/fenix/components/metrics/AppLaunchTimeMeasurement.kt
```diff
@@ -0,0 +1,78 @@
...
+    private var homeActivityInitTimeStampNanoSeconds: Long? = null
+    private var homeActivityOnRestartTimeStampNanoSeconds: Long? = null
+    // we are considering screen to be visible after the first pre draw call.
+    private var homeActivityOnPreDrawTimeStampNanoSeconds: Long = INVALID_TIMESTAMP
+
+    fun onApplicationOnCreate() {
+        isColdStartup = true
+    }
+
+    fun onHomeActivityOnCreate(currentTimeStampNanoSeconds: Long) {
```
mcomella, line 32: nit: -> activityInitNanos

app/src/main/java/org/mozilla/fenix/components/metrics/AppLaunchTimeMeasurement.kt
```diff
@@ -0,0 +1,78 @@
...
+        if (!isColdStartup) {
+            isWarmStartup = true
+        }
+        this.homeActivityInitTimeStampNanoSeconds = currentTimeStampNanoSeconds
+    }
+
+    fun onHomeActivityOnRestart(currentTimeStampNanoSeconds: Long = SystemClock.elapsedRealtimeNanos()) {
```
mcomella, line 39: We should capture this time before `super.onRestart`, right? It might take a non-trivial amount of time to run that and other folks might add code in between that and this.

Either call this first or capture the time first and pass it to this method.

app/src/main/java/org/mozilla/fenix/components/metrics/AppLaunchTimeMeasurement.kt
```diff
@@ -0,0 +1,78 @@
...
+        if (!isColdStartup && !isWarmStartup) {
```
mcomella, line 40: Why do we check this? Neither should ever be set, right? Or is this related to navigating between activities when the process is open (have we tested this case? what lifecycle methods are called?)? If it's an illegal state, should we throw or set an error if we hit this illegal state?

app/src/main/java/org/mozilla/fenix/components/metrics/AppLaunchTimeMeasurement.kt
```diff
@@ -0,0 +1,78 @@
...
+    fun onActivityOnPreDraw(currentTimeNanoSeconds: Long = SystemClock.elapsedRealtimeNanos()) {
+        isOnPreDrawCalled = true
+        homeActivityOnPreDrawTimeStampNanoSeconds = currentTimeNanoSeconds
+    }
+
+    /**
+     * if we have both start and finish time for launch, return it otherwise we return null
+     */
+    suspend fun getApplicationLaunchTime(): Long? = withContext(Dispatchers.IO) {
+        when {
```
mcomella, line 55: If we don't move to the enum model, I'm tempted to say we should do validation and ensure only one of the states is set.

app/src/main/java/org/mozilla/fenix/components/metrics/AppStartupTelemetry.kt
```diff
@@ -45,17 +54,28 @@ class AppStartupTelemetry(private val metrics: MetricController) : LifecycleObse
...
+    fun onExternalAppBrowserOnCreate(
+        safeIntent: SafeIntent,
+        hasSavedInstanceState: Boolean,
+        homeActivityInitTimeStampNanoSeconds: Long
+    ) {
+        setOnCreateData(safeIntent, hasSavedInstanceState, homeActivityInitTimeStampNanoSeconds, true)
     }
 
     fun onHomeActivityOnRestart() {
+        appLaunchTimeMeasurement.onHomeActivityOnRestart()
```
mcomella, line 77: nit: mention nothing above here because it measures

app/src/main/java/org/mozilla/fenix/components/metrics/AppStartupTelemetry.kt
```diff
@@ -143,4 +157,17 @@ class AppStartupTelemetry(private val metrics: MetricController) : LifecycleObse
...
     }
+
+    /**
+     * We record metric on onPreDraw() since we need to wait for the application to draw the
+     * first frame. which will always happen after onResume. The hot startup time is a bit complicated
+     * because if we have onPreDraw listener attached, Activity will invoke onPreDraw() before onRestart.
+     * Therefore, we use OneShotPreDrawListener, which removes the listener after first call.
+     */
+    fun onPreDraw() {
+        appLaunchTimeMeasurement.onActivityOnPreDraw()
```
mcomella, line 168: nit: comment that there should be nothing above here because it measures

app/src/main/java/org/mozilla/fenix/components/metrics/AppStartupTelemetry.kt
```diff
@@ -143,4 +157,17 @@ class AppStartupTelemetry(private val metrics: MetricController) : LifecycleObse
...
+        MainScope().launch(Dispatchers.IO) {
```
mcomella, line 169: There's still a race – what if `onStop` clears this data before we're done in `recordMetric`?

I think the simplest solution is to transmit in `onStop` and block, assuming we don't expect it to take too long. A few other solutions:
- We can record here but we copy the data into the `recordMetric` function so that it can be cleared later without affecting the state. Note that we'll likely end up moving this later to minimize what runs before visual completeness
- We lock the data from here until we're done recording such that `onStop` can't clear it until we're done clearing (this is hard because we have to lock it before we even dispatch and even when we suspend to guarantee correctness... there might be a coroutines lock to help here)

mcomella (commented):
I don't have time to do more today but I'm still concerned about the race. It generally looks like it works though but I'll have to spend more time on it.

app/src/main/java/org/mozilla/fenix/components/metrics/AppStartupTelemetry.kt
mcomella: I commented in my review about this

mcomella (changes requested):
Meant to request changes. Note: after the next iteration, I still need to:
- Review more thoroughly (including build & testing, especially edge cases)
- Read through the tests

app/src/main/java/org/mozilla/fenix/components/metrics/AppStartupTelemetry.kt
sraturi: done

app/src/main/java/org/mozilla/fenix/components/metrics/AppLaunchTimeMeasurement.kt
sraturi: moved it to the enum convention!

app/src/main/java/org/mozilla/fenix/components/metrics/AppLaunchTimeMeasurement.kt
sraturi: move t the enum model

app/src/main/java/org/mozilla/fenix/components/metrics/AppStartupTelemetry.kt
sraturi: > transmit in onStop and block

guess simplest is to use `runBlocking`? or a coroutine on main thread. though I guess they both gets us same result at the end for this case?

app/src/main/java/org/mozilla/fenix/components/metrics/AppLaunchTimeMeasurement.kt
sraturi: fixed it.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
sraturi: Ah my bad! I mistook the `doOnPreDraw` listener for `ViewTreeObserver.OnPreDrawListener `.

I changed `oneShotListener` to `doOnPreDraw`. 

As for thinking of `AppStartupTelemetry` as an extension, thats a pretty good way of thinking about it and I agree with it. 
However, in this case, since we are adding a listener on, I wonder if the pros (4 less line in `homeactivty`) outweighs the cons (unit testing a listener inside a function). I haven't looked at how complicated it is but I'm also more inclined to keep testing simple?

app/src/main/java/org/mozilla/fenix/components/metrics/AppStartupTelemetry.kt
sraturi: in the case where, `onstop` is called before `onPreDraw`, Im thinking of omiting the startup time but still report the other metrics?

mcomella: > are you thinking of using runBlockingor a coroutine on main thread? though I guess they both gets us same result at the end for this case?

I think we should get rid of the `suspend` keywords and just run it on the main thread. If you do `runBlocking` to a different dispatcher, you're potentially utilizing two threads (one will be sleeping though) and spending some resources dispatching. We might as well just run it on the main thread.

Since we're reading from the pseudo-filesystem and not the file system, I'm not too concerned about the performance of this (though we could measure locally or add a probe to measure the length just in case 🤷).

> in the case where, onstop is called before onPreDraw, Im thinking of omiting the startup time but still report the other metrics?

I'm not sure – in what cases can this happen? Does it change how our startup mental model should work?

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
mcomella: > However, in this case, since we are adding a listener on, I wonder if the pros (4 less line in homeactivty) outweighs the cons (unit testing a listener inside a function). I haven't looked at how complicated it is but I'm also more inclined to keep testing simple?

What do you mean by unit testing a listener inside a function?

fwiw, I think we should consider the incidental complexity that occurs when we're leaving the code here:
- we're making HomeActivity longer and more complex.
- we're making AppStartupTelemetry more complex by splitting its functionality into multiple files

And at a more nuanced, low-level:
- in the future, it's more likely that if this method is modified, HomeActivity will be coupled to this listener. 
- we're duplicating the `onPreDraw` call here (e.g. I noticed you didn't use `doOnPreDraw` in `ExternalAppBrowserActivity`). if we fix it with a helper function, we're making HomeActivity more complicated even though we already have a deduplication mechanism inside the AppStartupTelemetry (so there would be duplication of the deduplication mechanism)

app/src/main/java/org/mozilla/fenix/components/metrics/AppLaunchTimeMeasurement.kt
```diff
@@ -0,0 +1,92 @@
...
+                    }
+                    else -> {
+                        null
+                    }
+                }
+            }
+        }
+    }
+
+    enum class StartupMode { COLD, WARM, HOT, UNKNOWN }
```
mcomella, line 87: nit: private

app/src/main/java/org/mozilla/fenix/components/metrics/AppLaunchTimeMeasurement.kt
```diff
@@ -0,0 +1,92 @@
...
+        isOnPreDrawCalled = true
+        homeActivityOnPreDrawTimeStampNanoSeconds = activityOnPreDrawNanos
+    }
+
+    /**
+     * if we have both start and finish time for launch, return the difference otherwise return null.
+     */
+    suspend fun getApplicationLaunchTime(): Long? = withContext(Dispatchers.IO) {
+        when {
+            !isOnPreDrawCalled -> {
```
mcomella, line 57: When will `onPreDraw` not be called? Should it be better represented in our startup states somehow? Add a comment.

app/src/main/java/org/mozilla/fenix/components/metrics/AppLaunchTimeMeasurement.kt
```diff
@@ -0,0 +1,92 @@
...
+                        homeActivityOnPreDrawTimeStampNanoSeconds.minus(
+                            homeActivityInitTimeStampNanoSeconds!!
+                        )
+                    }
+                    HOT -> {
+                        homeActivityOnPreDrawTimeStampNanoSeconds.minus(
+                            homeActivityOnRestartTimeStampNanoSeconds!!
+                        )
+                    }
+                    else -> {
```
mcomella, line 79: When can this happen? Should we be reporting an error instead?

mcomella, line 79: -> UNKNOWN

If you have an exhaustive type such as an enum, you should specify all of them instead of using `else`: using non-else values, the compiler will warn you if you don't handle all the cases, which is helpful if you later add a new case and forget to handle it everywhere.

app/src/main/java/org/mozilla/fenix/components/metrics/AppLaunchTimeMeasurement.kt
```diff
@@ -0,0 +1,92 @@
...
+        }
+        this.homeActivityInitTimeStampNanoSeconds = activityInitNanos
+    }
+
+    fun onHomeActivityOnRestart(activityOnRestartNanos: Long = SystemClock.elapsedRealtimeNanos()) {
+        startupMode = HOT
+        homeActivityOnRestartTimeStampNanoSeconds = activityOnRestartNanos
+    }
+
+    fun onActivityOnPreDraw(activityOnPreDrawNanos: Long = SystemClock.elapsedRealtimeNanos()) {
```
mcomella, line 47: nit: I wonder if we should rename this and the variables in this function to be `onFirstFrameDrawn` – that's what we're trying to represent with these values and it's not clear anywhere in this file right now.

app/src/main/java/org/mozilla/fenix/components/metrics/AppLaunchTimeMeasurement.kt
```diff
@@ -0,0 +1,92 @@
...
+                    }
+                }
+            }
+        }
+    }
+
+    enum class StartupMode { COLD, WARM, HOT, UNKNOWN }
+
+    companion object {
+        private const val INVALID_TIMESTAMP: Long = -1
```
mcomella, line 90: This isn't used anywhere – why not remove this and use null for the initial value like all the other values?

(Technically, I think it'd be better if everything was non-nullable and started as -1 because otherwise there's autoboxing but it should be negligble; probably no need to name -1 in that case)

mcomella (changes requested):
I don't understand the edge cases (iirc, if startupMode is not set and onPreDraw not being called) so let's figure those out before I manually test, check the tests, etc.

If you have the time, it might be easier to resolve these synchronously on Matrix or Zoom.

app/src/main/java/org/mozilla/fenix/components/metrics/AppLaunchTimeMeasurement.kt
sraturi: since we are sending data onStop(), an example would be launching the app and pressing the home button before onDraw is called.

app/src/main/java/org/mozilla/fenix/components/metrics/AppLaunchTimeMeasurement.kt
sraturi: Do you mean report error as in throw an exception or report error to glean?

app/src/main/java/org/mozilla/fenix/components/metrics/AppLaunchTimeMeasurement.kt
sraturi: I guess its ok to do that as long as it is common to assume onPreDraw() to be same as firstFrameDrawn?. I always assumed first frame drawn = first call to onDraw. not sure how many MS diff it would be.

app/src/main/java/org/mozilla/fenix/components/metrics/AppLaunchTimeMeasurement.kt
mcomella: Is that possible? Won't we always draw a frame before `onStop()` is called? e.g.
```
onStart()
onResume()
onDraw() // first frame at least
onPause()
onStop()
```

Or, is it possible that if onStart/onResume take too long, the user can click home and `onDraw` is not called?

app/src/main/java/org/mozilla/fenix/components/metrics/AppLaunchTimeMeasurement.kt
mcomella: Probably report an error to glean – since this gets called every startup and startup is complicated, there is a high risk that this could case could happen. If we crashed, we might disrupt a lot of users' experiences.

app/src/main/java/org/mozilla/fenix/components/metrics/AppLaunchTimeMeasurement.kt
mcomella: That's true – maybe it's preferred to say `onFirstFramePreDraw` :). I think they're close enough to that it's not worth implementing a more complex solution, given onPreDraw is already implemented for us.

app/src/main/java/org/mozilla/fenix/components/metrics/AppLaunchTimeMeasurement.kt
mcomella: Spoke on Matrix – `onDraw` isn't called. Can we add a comment to explain this behavior?

>  Should it be better represented in our startup states somehow? 

Not sure how important this is.

app/src/main/java/org/mozilla/fenix/components/metrics/AppLaunchTimeMeasurement.kt
mcomella, line 87: This state is duplicated from `AppStartupTelemetry` – is there a way to pass it in from there?

app/src/main/java/org/mozilla/fenix/components/metrics/AppLaunchTimeMeasurement.kt
sraturi: done!

app/src/main/java/org/mozilla/fenix/components/metrics/AppLaunchTimeMeasurement.kt
sraturi: switched it to null for now

app/src/main/java/org/mozilla/fenix/components/metrics/AppLaunchTimeMeasurement.kt
sraturi: fixed it.

app/src/main/java/org/mozilla/fenix/components/metrics/AppLaunchTimeMeasurement.kt
sraturi: done

mcomella (commented):
I see two small changes that I think we should make. I'll need to repush this in order to get CI to run so I'll just take care of them when I do that.

I still need to test locally before I can r+.

mcomella (approved):
I tested the primary workflows and confirmed the start time was there.

I took two profiles on my GS5 Nightly build to see if `runBlocking` in `onStop` slowed down the app too much and the method doesn't even appear in the profiles (I confirmed the telemetry was being sent):
- https://share.firefox.dev/2SkEWpm
- https://share.firefox.dev/2HKHnzD

I also confirmed StrictMode does kill us for calling `Stat` but maybe that's because it doesn't know how fast (apparently) the pseudo-filesystem seems to be. Perhaps we can later simplify this code by making it synchronous, when we test the speed of the pseudo-filesystem.

mcomella:
Thanks @sraturi !

app/src/main/java/org/mozilla/fenix/customtabs/ExternalAppBrowserActivity.kt
```diff
@@ -36,7 +37,12 @@ open class ExternalAppBrowserActivity : HomeActivity() {
     final override fun getIntentSessionId(intent: SafeIntent) = intent.getSessionId()
 
     override fun startupTelemetryOnCreateCalled(safeIntent: SafeIntent, hasSavedInstanceState: Boolean) {
-        components.appStartupTelemetry.onExternalAppBrowserOnCreate(safeIntent, hasSavedInstanceState)
+        components.appStartupTelemetry.onExternalAppBrowserOnCreate(
+            safeIntent,
+            hasSavedInstanceState,
+            homeActivityInitTimeStampNanoSeconds,
```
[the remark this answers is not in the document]
mcomella: Surprisingly to me, `HomeActivity.init` is called first. Here's the timings I pulled:
```
External app 619720736139198
HomeActivity 619720735597114
```

The difference is ~0.5ms on the Pixel 2. Sounds like we should keep it as is. I'll comment about this to avoid confusion.

app/metrics.yaml
```diff
@@ -48,6 +52,13 @@ events:
           for example, when we are doing a hot start up, we cant have a
           savedInstanceState therefore we report only [APP_ICON, HOT] instead
           of [APP_ICON, HOT, false].
+      launch_time_nano_seconds:
```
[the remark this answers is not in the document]
mcomella: Done in the new PR.

app/src/main/java/org/mozilla/fenix/customtabs/ExternalAppBrowserActivity.kt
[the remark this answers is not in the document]
mcomella: Done in the new PR.

mcomella:
Closing in favor of https://github.com/mozilla-mobile/fenix/pull/15605 because we can't run CI on contributor PRs.

Issue #11561, opened 2020-06-14
[Bug] (a11y) tabs tray: bad contrast for inactive mode icon in dark theme


cadeyrn:
### Steps to reproduce
1. enable the dark theme
1. open the tabs tray

### Expected / Actual behavior
The mode icons (regular or private tabs) should have a sufficient color contrast ratio. While I think the icon of the active mode have a good contrast in the light and in the dark theme I can't say the same for the icon of the inactive mode. It's hard to see in the dark mode. In fact there is a big difference in the color contrast ratio when I compare the light and the dark theme:

Light theme: 2.85:1
Dark theme: 1.91:1

![Screenshot](https://user-images.githubusercontent.com/1100614/84588667-d2197f80-ae29-11ea-8a8b-1ac417ee4199.jpg)UNITO-UNDERSCORE!20200614-101951!
(In this screenshot the icon for private tabs is the inactive icon)

### Device information
* Android device: OnePlus 7T Pro McLaren Edition with Oxygen OS 10.0.9 (Android 10)
* Fenix version: master branch revision 7e6116512c354eb1d8524f8922f7970fe155c65c

topotropic:
@AmyYLee here's a a11y contrast issue with the inactive PBM icon. Can you have a look? Thanks!

AmyYLee:
@gabrielluong @topotropic - For deselected states can we keep it to Light Grey 5 (#FBFBFE) - This would also apply to the 3-dot menu as well. Thanks

ebalazs-sv:
Verified as fixed on Nightly 200929 05:08 (Build #2015766697) GV 83.0a1 from 9/29 with Motorola Moto G6 (Android 8) and Samsung Galaxy Tab S3 (Android 8).
I will close this issue and remove the qa:needed label.

![rsz_screenshot_20200929-145100_firefox_nightly](https://user-images.githubusercontent.com/48799384/94555036-a64cc900-0263-11eb-970c-39f8a836ba3e.jpg)

PR #14439, opened 2020-08-29
For #11561 - changed inactive heading and menu icon color in dark theme tabs tray

cadeyrn:
For #11561.

As requested by UX I changed the deselected state for the inactive heading icons and for the menu icon as well to Light Grey 5 in dark theme. Nothing was changed for the light theme.

Before:

before

After:

after

### DIFF
--- a/app/src/main/res/layout/component_tabstray.xml
+++ b/app/src/main/res/layout/component_tabstray.xml
@@ -150,7 +150,7 @@
             android:background="?android:attr/selectableItemBackgroundBorderless"
             android:contentDescription="@string/open_tabs_menu"
             android:visibility="visible"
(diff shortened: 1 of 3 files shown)


brampitoyo (commented):
Tagging @topotropic specifically for design review.

ekager:
pinging @topotropic 🏓

cadeyrn:
Since @topotropic it not reacting maybe someone else from the UX team could give the _okay_? I would like to get this change included because it's an a11y issue. Please also note that the color was provided by @AmyYLee from the UX team in #11561 so it should be okay anyway.

topotropic:
Apologies for my super late reply. Thanks for taking care! LGTM

topotropic (approved):

PR #14587, opened 2020-09-01
Prevent GeckoView from resizing while searching

boek:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/searchdialog/SearchDialogFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/searchdialog/SearchDialogFragment.kt
@@ -17,6 +17,7 @@ import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
 import android.view.ViewStub
+import android.view.WindowManager
 import androidx.appcompat.app.AlertDialog
 import androidx.appcompat.app.AppCompatDialogFragment
 import androidx.constraintlayout.widget.ConstraintProperties.BOTTOM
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/fenix/searchdialog/SearchDialogFragment.kt
```diff
@@ -80,6 +81,21 @@ class SearchDialogFragment : AppCompatDialogFragment(), UserInteractionHandler {
     private val qrFeature = ViewBoundFeatureWrapper<QrFeature>()
     private val speechIntent = Intent(RecognizerIntent.ACTION_RECOGNIZE_SPEECH)
 
+    override fun onStart() {
+        super.onStart()
+        // https://github.com/mozilla-mobile/fenix/issues/14279
+        // To prevent GeckoView from resizing we're going to change the softInputMode to not adjust
+        // the size of the window.
+        requireActivity().window!!.setSoftInputMode(WindowManager.LayoutParams.SOFT_INPUT_ADJUST_NOTHING)
```
csadilek, line 89: Don't think you need the `!!` as window is not nullable here?

app/src/main/java/org/mozilla/fenix/searchdialog/SearchDialogFragment.kt
```diff
@@ -80,6 +81,21 @@ class SearchDialogFragment : AppCompatDialogFragment(), UserInteractionHandler {
...
+    }
+
+    override fun onStop() {
+        super.onStop()
+        // https://github.com/mozilla-mobile/fenix/issues/14279
+        // Let's reset back to the default behavior after we're done searching
+        requireActivity().window!!.setSoftInputMode(WindowManager.LayoutParams.SOFT_INPUT_ADJUST_RESIZE)
```
csadilek, line 96: Don't think you need the `!!` as window is not nullable here?

person808:
Swipe to switch tabs is disabled when the keyboard is visible due to some hitbox detection issues (I don't know the issue numbers off the top of my head). The keyboard visibility detection might only work with ADJUST_RESIZE (https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/java/org/mozilla/fenix/ext/View.kt#L69). I'm not sure if the new visibility apis in androidx-core 1.5 alpha would handle this correctly.

PR #14951, opened 2020-09-10
Fix #14950 - Make the "Add new tab" button visible in private mode on light theme

### LINKED ISSUE #14950: [Bug] Add new tab plus button is impossible to see in private mode when using light theme and accessibility settings (ex: TalkBack) is on
## Steps to reproduce
1. Enable light theme in Firefox. 
1. Turn on TalkBack in the system settings on your Android device.
1. Now, in Firefox, enable "private mode" either by clicking the eye mask icon on the top right of the home page, OR by opening the tabs tray and clicking the private tabs icon and clicking the new tab button. 
    * Note: please test with the latest master of Firefox or any build that includes the fix for #14540, otherwise the tabs tray will simply crash the app if you have an open tab!
1. Now, in private tabs mode, open the tabs tray again. Note that the + sign in the top right of the tray is nearly invisible. This is because it uses white color, which is the primary text color in private mode.

### Expected behavior

I expect that the new tab plus button is visible in the tabs tray in all scenarios.

### Actual behavior

For the specific scenario of using a light theme and in private tabs mode, with accessibility settings enabled (ex: TalkBack), the button is impossible to see. See the screenshot below. If you download the image and inspect closely, you can see a white button on a very light background. The button exists and talkback can highlight it in green as seen. Also, note that the issue does not exist in a dark theme or in regular tabs mode.

#### Cause for the issue
The culprit is: https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/res/drawable/ic_new.xml#L11
```
        android:fillColor="?primaryText"
```
should instead be:
```
        android:fillColor="@color/primary_text_normal_theme"
```

### Device information

* Android device: Pixel 2, running Android 10
* Fenix version: Issue still exists on the latest master 2e3863578187d7fce4ca8e90b26ace165c1ea15c

abhijitvalluri:
This PR fixes the issue I reported in #14950. For completeness I shall repeat the issue here as well.

When using the light theme and in private tabs mode, with accessibility settings enabled (ex: TalkBack), the "add new tab" button is impossible to see. See the screenshot below (before the fix version). If you download the image and inspect closely, you can see a white button on a very light background. The button exists and talkback can highlight it in green as seen. Note that the issue does not exist in a dark theme or in regular tabs mode, but only in light theme mode and in private tabs with accessibility setting such as TalkBack enabled.

The issue is because `android:fillColor="?primaryText"` is used in `ic_new.xml`. Instead, it should be `android:fillColor="@color/primary_text_normal_theme"`. This is because, in private tabs mode primary text color is white, and so it is conflicting with the light theme. So, we should use the primary text color that correctly changes for both light and dark theme.

### Pull Request checklist

- [x] **Tests**: This PR is very trivial, so it does not include any test code. However, there is a visual test in terms of screeshots.
- [x] **Screenshots**: This PR includes screenshots of the fix for the hard to see add new tab button
- [x] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features. In addition, it includes a screenshot of a successful [accessibility scan](https://play.google.com/store/apps/details?id=com.google.android.apps.accessibility.auditor&hl=en_US) to ensure no new defects are added to the product. This code change improves user experience when using accessibility features (specifically TalkBack).

1. click on Show All Checks,

### DIFF
--- a/app/src/main/res/drawable/ic_new.xml
+++ b/app/src/main/res/drawable/ic_new.xml
@@ -8,6 +8,6 @@
     android:viewportWidth="24"
     android:viewportHeight="24">
     <path
-        android:fillColor="?primaryText"
+        android:fillColor="@color/primary_text_normal_theme"
         android:pathData="M13,4a1,1 0,1 0,-2 0v7H4a1,1 0,1 0,0 2h7v7a1,1 0,1 0,2 0v-7h7a1,1 0,1 0,0 -2h-7V4z" />
 </vector>
--- a/app/src/main/res/drawable/ic_new.xml
+++ b/app/src/main/res/drawable/ic_new.xml
@@ -8,6 +8,6 @@
     android:viewportWidth="24"
     android:viewportHeight="24">
     <path
-        android:fillColor="@color/primary_text_normal_theme"
+        android:fillColor="?primaryText"
         android:pathData="M13,4a1,1 0,1 0,-2 0v7H4a1,1 0,1 0,0 2h7v7a1,1 0,1 0,2 0v-7h7a1,1 0,1 0,0 -2h-7V4z" />
 </vector>
--- a/app/src/main/res/layout/component_tabstray.xml
+++ b/app/src/main/res/layout/component_tabstray.xml
@@ -139,6 +139,7 @@
             app:layout_constraintBottom_toBottomOf="@id/tab_layout"
             app:layout_constraintEnd_toStartOf="@id/tab_tray_overflow"
             app:layout_constraintTop_toTopOf="@id/tab_layout"
+            android:tint="@color/primary_text_normal_theme"
             app:srcCompat="@drawable/ic_new" />
 
         <ImageButton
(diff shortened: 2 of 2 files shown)


ekager (changes requested):
Thanks for taking this on!
So this _will_ fix it in this case, but we also use `ic_new` elsewhere we we don't want to make this change (ie from a quick search, bookmark or collection creation I'm assuming this will look strange from this change if you want to verify). 
The right fix I think we want is to tint this in place (in `components_tabstray.xml`) with `primary_text_normal_theme` where we use it in the Tabs Tray.

ekager (changes requested):
Nice almost there! Let's use `app:tint` so that it works properly on all API versions

abhijitvalluri:
Thanks for the review! That's a fair point. I noticed that in pretty much all the places where `ic_new` is used, the tint was either set programmatically in the code, or in the xml. So, most likely my original change would not have effected anything visually. Nevertheless, setting the tint in `component_tabstray.xml` rather than relying on the tint present in the drawable file is indeed the better approach and is consistent with other usages.

ekager (approved):

PR #14957, opened 2020-09-10
Fixes #12099 - Use bookmark keywords in browser searches

### LINKED ISSUE #12099: Bookmark keyword support - awesomebar navigation
On desktop and iOS browsers we have support for bookmark keywords. It allows setting a "keyword" for a bookmark, and then navigating to that bookmark by typing that keyword into the url/awesomebar.

We have support for syncing the keywords already via a-s, so we can implement the awesomebar part of this already. (depends on https://github.com/mozilla-mobile/android-components/issues/7558).

Another part to this is to allow editing keywords from Fenix. Currently, we don't have storage APIs for this at all. I've split that into a [separate issue](https://github.com/mozilla-mobile/fenix/issues/12100).

Mugurell:
For the moment users are not able to view/add/modify bookmark keywords in Fenix
but they can in Firefox running on other platforms.
After syncing the bookmarks Fenix users will be able to use the same already
set keywords in Fenix and have the same search experience with included support
for query placeholders.

### Pull Request checklist

- [x] **Tests**: No tests. Simple wiring of new dependencies.
- [x] **Accessibility**: The code in this PR or does not include any user facing features.

### This patch comes to integrate https://github.com/mozilla-mobile/android-components/pull/8375 which is a breaking change.

BookmarkKeywordsInFenix
[video demoing bookmark keywords in Fenix](https://drive.google.com/file/d/14Z9KEp8IgtkhQbs-dNjEgN_OKZNDmr83/view?usp=sharing)

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/Components.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/Components.kt
@@ -54,7 +54,8 @@ class Components(private val context: Context) {
             core.store,
             search.searchEngineManager,
             core.webAppShortcutManager,
-            core.topSitesStorage
+            core.topSitesStorage,
+            core.bookmarksStorage
         )
     }
     val intentProcessors by lazy {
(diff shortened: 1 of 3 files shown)


Mugurell:
We are transitioning from bookmark keywords to custom search engines.

Issue #12829, opened 2020-07-22
[Bug] Add Create Account on the the Sign In page on Fenix


AnaMedinac:
@davismtl noticed that we are seeing a steady decrease in FxA registrations on iOS since we implemented the QR pairing flow. We figured that Fenix might be seeing a similar problem since the current UX has no clear call to action for registering as it's very sign in focused: 

![](https://user-images.githubusercontent.com/22782298/88195941-9c2c9f80-cc0e-11ea-81ba-2a1264def06b.png)

@j19s84 On the FxA side has mocked up a proposal for the UX to include a third button - `Create account`: 

![](https://user-images.githubusercontent.com/22782298/88197390-6d172d80-cc10-11ea-9bca-b785ce21dbcd.png)

The `Create account` button would redirect the user to the same web page as the `Use email instead flow`. The addition of this button should help users easily register for a Firefox Account on mobile too.

We also increased the font size for the URL as it went since it sometimes went un-noticed by users.

davismtl:
This may be redundant but hoping it adds some additional clarity:
We expect the pairing flow to help drive logins when Fenix lands in the release channel but since iOS had before/after data in regards to the roll-out of the pairing experience (unlike Fenix), we observed that it hurt their registration numbers. Fenix and Fx iOS have the same pairing UX so we believe that the Android release channel will suffer the same fate when Fenix replaces Fennec.

We also tried to stick to the most minimal UX change to resolve this problem to avoid the Fenix team more work in what is likely a very busy time.

For reference, here is the mirror bug on the iOS side:
https://github.com/mozilla-mobile/firefox-ios/issues/7016

AnaMedinac:
After talking to @betsymi and @j19s84, this is the proposed design: 

<img width="377" alt="image" src="https://user-images.githubusercontent.com/22782298/88957581-0ecff780-d26d-11ea-851b-65b6fee051d4.png">

Where the `Create one` button leads to the same webpage as `Use Email instead` button.

mcarare:
@mozilla-mobile/ux  Can UX specify the color to be used for the "create account" text, in dark/light theme? Thank you!

apbitner:
@mcarare I don't believe we change the color of inline text links, we just add an underline.

Also just to clarify, the last screenshot above is for iOS, so the only change needed for Fenix is the addition of the "No account? .... " sentence.

mcarare:
[...]

PR #15063, opened 2020-09-14
For #12829:  Add create account option on the the sign in screen.

mcarare:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/settings/account/TurnOnSyncFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/account/TurnOnSyncFragment.kt
@@ -6,6 +6,7 @@ package org.mozilla.fenix.settings.account
 
 import android.Manifest
 import android.os.Bundle
+import android.text.Spanned
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
@@ -25,6 +26,7 @@ import org.mozilla.fenix.HomeActivity
 import org.mozilla.fenix.R
 import org.mozilla.fenix.components.FenixSnackbar
 import org.mozilla.fenix.components.metrics.Event
+import org.mozilla.fenix.ext.addUnderline
 import org.mozilla.fenix.ext.requireComponents
 import org.mozilla.fenix.ext.settings
 import org.mozilla.fenix.ext.showToolbar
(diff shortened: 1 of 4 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -1138,6 +1138,10 @@
     <string name="sign_in_with_camera">Sign in with your camera</string>
     <!-- Text shown for settings option for sign with email -->
     <string name="sign_in_with_email">Use email instead</string>
+    <!-- Text shown for settings option for sign with email -->
```
ekager, line 1141: nit - let's add to the string description that Firefox is intentionally hardcoded?

ekager: Also I don't think these strings descriptions are right? 🤔 Let's update those too!

app/src/main/java/org/mozilla/fenix/settings/account/TurnOnSyncFragment.kt
```diff
@@ -89,6 +100,21 @@ class TurnOnSyncFragment : Fragment(), AccountObserver {
...
+        val spanEnd = spanStart + createAccountActionText.length
+
+        view.createAccount.apply {
+            text = fullText
+            addUnderline(
+                spanStart,
+                spanEnd,
+                Spanned.SPAN_EXCLUSIVE_EXCLUSIVE
+            )
+            setOnClickListener(createAccountClickListener)
```
ekager, line 116: nit : could we just do this here?
```suggestion
            setOnClickListener { navigateToPairWithEmail() }
```

ekager (approved):
R+ with a few nits! Thanks!

app/src/main/res/values/strings.xml
ekager: Oh and we probably also want to add what the substitution in the string is about generally

app/src/main/java/org/mozilla/fenix/settings/account/TurnOnSyncFragment.kt
mcarare: We could do that, but I intentionally made this like that, so any future change would be made in `createAccountClickListener` not inside the click listener setter. 
Also,  it is more easily readable that it is not a mistake, setting the same action on 2 different views.

ekager (approved):

Issue #13943, opened 2020-08-19
Change strictMode penalty to death so it's harder to ignore


mcomella:
Blocked by https://github.com/mozilla-mobile/fenix/issues/13368

mcarare:
@mcomella Are you referring only to `ThreadPolicy`? Or also `VmPolicy`. `VmPolicy` has multiple violations that could block changing policy to death.

mcomella:
> @mcomella Are you referring only to `ThreadPolicy`? Or also `VmPolicy`. `VmPolicy` has multiple violations that could block changing policy to death.

@mcarare Just the main thread policy right now – specifically, we're trying to prevent main thread IO like network access and disk reads.

mcomella:
Addressed with https://github.com/mozilla-mobile/fenix/pull/15064.

PR #15064, opened 2020-09-14
For #13943: Change strict mode penalty to penaltyDeath.

mcarare:
Also, keep dialogPenalty as replacement option for penaltyDeath.

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/StrictModeManager.kt
+++ b/app/src/main/java/org/mozilla/fenix/StrictModeManager.kt
@@ -16,16 +16,24 @@ object StrictModeManager {
 
     /***
      * Enables strict mode for debug purposes. meant to be run only in the main process.
-     * @param setPenaltyDialog boolean value to decide setting the dialog box as  a penalty.
+     * @param setPenaltyDeath boolean value to decide setting the penaltyDeath as a penalty.
+     * @param setPenaltyDialog boolean value to decide setting the dialog box as a penalty.
+     * Note: dialog penalty cannot be set with penaltyDeath
      */
-    fun enableStrictMode(setPenaltyDialog: Boolean) {
+    fun enableStrictMode(setPenaltyDeath: Boolean, setPenaltyDialog: Boolean = false) {
         if (Config.channel.isDebug) {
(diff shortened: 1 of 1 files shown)


MarcLeclair:
@mcarare  thank you so much for doing this! Would you mind just removing `penaltyDialog` and just keeping `penaltyDeath`? I think this might be better since we *really* want engineers to know what they're doing shouldn't be happening and the `penaltyDialog` can be quite easy to ignore 😄

app/src/main/java/org/mozilla/fenix/StrictModeManager.kt
```diff
@@ -16,16 +16,24 @@ object StrictModeManager {
 
     /***
      * Enables strict mode for debug purposes. meant to be run only in the main process.
-     * @param setPenaltyDialog boolean value to decide setting the dialog box as  a penalty.
+     * @param setPenaltyDeath boolean value to decide setting the penaltyDeath as a penalty.
+     * @param setPenaltyDialog boolean value to decide setting the dialog box as a penalty.
+     * Note: dialog penalty cannot be set with penaltyDeath
      */
-    fun enableStrictMode(setPenaltyDialog: Boolean) {
+    fun enableStrictMode(setPenaltyDeath: Boolean, setPenaltyDialog: Boolean = false) {
```
MarcLeclair, line 23: I think just leaving `setPenaltyDeath` would be preferable

MarcLeclair (changes requested):

app/src/main/java/org/mozilla/fenix/StrictModeManager.kt
mcarare: This just keeps dialogPenalty as an option, so it will be easier to switch between them in the future. Right now this code enforces death penalty.

MarcLeclair: That's true. I guess we can keep it for now and remove it if we see no one uses it 😄  sorry for the late reply!

MarcLeclair (approved):

Issue #14230, opened 2020-08-26
[Bug] Adding/removing top sites (frecent too) causes a flash


jonalmeida:
## Steps to reproduce
1. Add/remove a top site.
2. Observe the top sites list.
3. Enable frecent top sites.
3. Visit a new site.
4. Go back to the home screen.
5. Observe the top sites list.

### Expected behavior
 - At 2, a top site item should animate in or out (depending on add/remove).
 - At 5, the new frecent top site items should animate in.

### Actual behavior
 - At 2, all the top sites disappear and reappear with the new change.
 - At 2, all the top sites disappear and reappear with the new change.

### Device information

* Android device: Android 9, Pixel 4
* Fenix version:  Nightly 25/08/2020 - this happens before the new top sites feature landed, but is more prominently visible with frecent top sites turned on.

Removing a pinned top site:
<img src="https://user-images.githubusercontent.com/1370580/91328400-e4525b00-e794-11ea-9fc0-90ae5f502584.gif" width=300/>

Adding a frecent top site:
<img src="https://user-images.githubusercontent.com/1370580/91328392-ddc3e380-e794-11ea-9e89-8f519212ee4d.gif" width=300 />

jonalmeida:
Updating the ticket from my last investigation: when adding break points in the DiffUtils calls. 

When you start with three top sites, remove one, then remove a second, the first one that was removed shows up before being removed with the second.

Somewhere (I haven't figured that out yet), there is an old cache which is not coming from the storage layer.

jonalmeida:
I think I have an idea what the bug is. We have two adapters that have caused confusion: `TopSitesAdapter` and `TopSitesPagerAdapter` - the latter is the inner adapter that holds the actual top sites but doesn't using `DiffUtils` to distinguish which top sites to update, so therefore the call the `notifyDataSetChanged` in it will update the entire list of top sites.

jonalmeida:
Adding an adapter diff check with `DiffUtils` seems to solve a different bug. We're still seeing the flash happening. 

I'm going to try next to see if I can modify the DiffUtils check to be more specific with which top site screen.

Ideally, we reduce the nested RecyclerViews and write our own custom LayoutManager to draw a the layout as we need it, but this is a much larger change.

jonalmeida:
[...]

PR #15126, opened 2020-09-15
For #14230: Stop redrawing all top sites when one is removed

jonalmeida:
In the `SessionControlAdapter.TopSitePager`, we need to be checking if the lists have changed (`sameAs`) or not before we look at the inner diff of the contents (`contentsSameAs`). Since the default implementation of equality on a list is to check the elements inside it as well, we were always returning `false` so they either top site list would refresh. What we really want is to have the adapter go into each top site and diff them separately.

With this fix though there are some other bugs that might have been hidden until now:
 - Updating one page of top sites, sometimes breaks the grid layout for the second page.
   - If we stick with a horizontal multi-row layout, we should strongly consider reducing the nested layouts we have here so we can reduce complexity and draw the top sites in one RecyclerView alone.
 - You can see frecent top sites move around the screen (usually on the second page) every time you revisit the homescreen.
   - I'm not sure how big of a deal it is, but in my testing where I was trying to trigger more frecent top site changes, the moving top sites with a visual pain for me.
 - There is still a smaller fade in animation as you can see the gifs below.
   - This is a larger improvement from the previous behaviour. We can better reduce this further with a combination of the changes above and also having a stable ID for our TopSites which now only come from pinned sites I believe.

@ekager I believe you worked a bit on the session adapter as well, so I wonder if you have any thoughts over here too.

| Before | After |
|--------|------|
| cMONEp | evXXpE |

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/TopSitePagerViewHolder.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/TopSitePagerViewHolder.kt
@@ -36,7 +36,8 @@ class TopSitePagerViewHolder(
     }
 
     fun bind(topSites: List<TopSite>) {
-        topSitesPagerAdapter.updateData(topSites)
+        val chunkedTopSites = topSites.chunked(TOP_SITES_PER_PAGE)
+        topSitesPagerAdapter.submitList(chunkedTopSites)
 
         // Don't show any page indicator if there is only 1 page.
         val numPages = if (topSites.size > TOP_SITES_PER_PAGE) {
--- a/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlAdapter.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlAdapter.kt
@@ -45,7 +45,7 @@ sealed class AdapterItem(@LayoutRes val viewType: Int) {
     data class TopSitePager(val topSites: List<TopSite>) : AdapterItem(TopSitePagerViewHolder.LAYOUT_ID) {
         override fun sameAs(other: AdapterItem): Boolean {
             val newTopSites = (other as? TopSitePager) ?: return false
-            return newTopSites.topSites == this.topSites
+            return newTopSites.topSites.size == this.topSites.size
         }
 
         override fun contentsSameAs(other: AdapterItem): Boolean {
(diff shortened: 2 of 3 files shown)


jonalmeida:
The failed UI test `openHistoryItemInNewPrivateTabTest` doesn't seem related to any of my changes on the home screen.

```
 java.lang.IllegalStateException: activityTestRule.activit…ewById(R.id.history_list) must not be null 
```
I most definitely didn't remove/add/alter any res IDs.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/TopSitePagerViewHolder.kt
```diff
@@ -36,7 +36,8 @@ class TopSitePagerViewHolder(
     }
 
     fun bind(topSites: List<TopSite>) {
-        topSitesPagerAdapter.updateData(topSites)
+        val chunkyTopSites = topSites.chunked(TOP_SITES_PER_PAGE)
```
ekager, line 39: im lolling at this naming

ekager: they chonk

gabrielluong:
I am not quite sure we can say the flickering is totally resolve since it's still there, except that now we have the list updating correctly and removing top sites work as intended so it is a net improvement. I still look at how Firefox Lite performs with their ViewPager2.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/TopSitePagerViewHolder.kt
jonalmeida: Oops, that wasn't supposed to go in!

jonalmeida:
> I am not quite sure we can say the flickering is totally resolve since it's still there, except that now we have the list updating correctly and removing top sites work as intended so it is a net improvement. I still look at how Firefox Lite performs with their ViewPager2.

I agree, I mentioned this in my opening comment. This is a best effort approach only.

Our complexity seems to be coming from the _three_ layers of RecyclerViews instead of the two that Lite probably has.

ekager (approved):
Happy w this as a first step of getting rid of some of the blinks and agree that the nested recyclerviews are getting a bit confusing and probably not performant. What is the next step here? Should we file a new ticket?

jonalmeida:
> Happy w this as a first step of getting rid of some of the blinks and agree that the nested recyclerviews are getting a bit confusing and probably not performant. What is the next step here? Should we file a new ticket?

Since we're in agreement, yeah we should do some follow-up work here to investigate if having stable IDs will fix the blink before moving on to making our own ViewPager or LayoutManager so solve the blinking. I've filed this one here: https://github.com/mozilla-mobile/fenix/issues/15180

Issue #15116, opened 2020-09-15
UninitializedPropertyAccessException: lateinit property browserToolbar has not been initialized


agi:
https://sentry.prod.mozaws.net/share/issue/30c5c45e3fd642d4920cbda02c0ca1d4/

AC issue from a little while ago: https://github.com/mozilla-mobile/android-components/issues/8305

```
UninitializedPropertyAccessException: lateinit property browserToolbar has not been initialized
    at kotlin.jvm.internal.ArrayIteratorKt.throwUninitializedPropertyAccessException(ArrayIterator.kt:2)
    at mozilla.components.browser.toolbar.behavior.BrowserToolbarBottomBehavior.tryToScrollVertically$browser_toolbar_release(BrowserToolbarBottomBehavior.kt:7)
    at mozilla.components.browser.toolbar.behavior.BrowserToolbarBottomBehavior$createGestureDetector$1.invoke(BrowserToolbarBottomBehavior.kt:3)
    at mozilla.components.browser.toolbar.behavior.BrowserGestureDetector$gestureDetector$1.invoke(BrowserGestureDetector.kt:5)
    at mozilla.components.browser.toolbar.behavior.BrowserGestureDetector$CustomScrollDetectorListener.onScroll(BrowserGestureDetector.kt:1)
...
(64 additional frame(s) were not displayed)

lateinit property browserToolbar has not been initialized
```

agi:
My Fenix nightly crashes non stop with this signature.

STR

Open Twitter PWA
Scroll down (like you were trying to pull to refresh)
crash

agi:
Samsung S10
Android 10
Latest Nightly from playstore

SeburoSUMO:
I am seeing the same thing.

Moto E5
Android 8.0
Latest Nightly from Google Play store.

kbrosnan:
See also 

* https://github.com/mozilla-mobile/android-components/issues/8405
* https://github.com/mozilla-mobile/android-components/issues/8305

jonalmeida:
I had filled this one for beta too: https://github.com/mozilla-mobile/fenix/issues/15086

kbrosnan:
Going to close this a dupe then.

liuche:
Okay, it looks like we've been duping this issue to some older issues - I'm going to re-open this bc it has a clear STR, and is for a recent Nightly.

liuche:
[...]

PR #15132, opened 2020-09-16
For #15116: Do not use bottom toolbar custom behavior in PWA

mcarare:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarView.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarView.kt
@@ -23,6 +23,7 @@ import kotlinx.android.synthetic.main.component_browser_top_toolbar.*
 import kotlinx.android.synthetic.main.component_browser_top_toolbar.view.*
 import mozilla.components.browser.domains.autocomplete.ShippedDomainsProvider
 import mozilla.components.browser.session.Session
+import mozilla.components.browser.state.state.ExternalAppType
 import mozilla.components.browser.toolbar.BrowserToolbar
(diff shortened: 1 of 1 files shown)


ekager (changes requested):
Could we just add a few comments about the behavior here? It's a bit confusing that we *aren't* setting the behavior if `isUrlbarHidingEnabled` is NOT enabled. I would assume from a point of not knowing about this part of the codebase that we would set the behavior if we want to do something with URL hiding. Thanks!

ekager:
Just tested this locally and it fixes the crashes I saw! Nice!

I did notice that when you click away from the main PWA scope (ie in Twitter go to Settings -> About Twitter -> About) the toolbar will no longer scroll. Ideally we would maintain the same behavior we had before - that might mean changing where/when we add or check for this behavior?

If that doesn't sound like an easy fix to find then we can merge these changes for now and file a followup! Just tag me for review again whenever

mcarare:
@ekager Thank you for pointing out, I was also confused by that Boolean. My first instinct is that it is set incorrectly in AC for PWA's. Because custom tabs should have `isUrlbarHidingEnabled` set to true, and custom tabs for PWA should have it set to false in config. I will investigate and come back with the changes. :)

mcarare:
@ekager I have changed to a more reliable indicator for PWA - `externalAppType,` also added this checks for TWA, as I tested and they also crashed.

ekager:
So I think `isUrlbarHidingEnabled` actually refers to when there is NO url shown at all (like when using the Twitter PWA). Let me pull this down and try it!

ekager (approved):
Tested locally on PWA and changing scope and looks good!

Issue #15112, opened 2020-09-15
[Bug] Recently Closed dissapears when you delete the last history item


s-ankur:
## Steps to reproduce
On a fresh install, open one webiste and then close it
go to history and delete the history entry

### Expected behavior
Recently closed stays because you do have recently closed tabs

### Actual behavior
Recently closed dissapears. The tabs are still saved there but you can't see them
### Device information

* Android device: Asus Zenfone
* Fenix version: Nightly

AndiAJ:
Hi, **verified as fixed** on the latest **`Nightly Build 200922`** using the following devices:
• Google Pixel 3a (Android 11)
• Huawei Mate 20 Lite (Android 10)
• OnePlus A3 (Android 6.0.1)

Issue #15088, opened 2020-09-14
[UX] Closed tabs look just like history entries in recently closed tabs list


yoasif:
There is no indication that it is anything *other* than a history entry - tabs contain more than just the last page visited in the tab, but the current UI does not show that in any way.

Some ideas (not fleshed out at all) might be:

* show a "bundle" of favicons when different sites exist in the history stack
* show number of pages in the stack 
* show when tab was closed and opened (lifetime of tab)
* show a list of sites present in the stack
* think of a tab like a mini collection? 

_Originally posted by @yoasif in https://github.com/mozilla-mobile/fenix/issues/15084#issuecomment-692244371_

Poopooracoocoo:
perhaps it could include a preview of the tab to make it look more like the list of tabs in the tab tray

apbitner:
Pinging @topotropic for feedback

topotropic:
> perhaps it could include a preview of the tab to make it look more like the list of tabs in the tab tray

That could help to make it more obvious that the closed tab has a history stack. That said, I'm not sure if the value warrants the effort here.

jawz101:
I think there are like 7 ways to open old things (maybe more if I think harder):

- visited sites on home page
- top sites on home page
- collections on home page
- tab manager
- history menu
- recently closed tabs in history menu
- bookmarks

PR #15140, opened 2020-09-16
For #15112 - Still show recently closed when history empty

ekager:
### Pull Request checklist

1. click on Show All Checks,


gabrielluong:
I am not sure if this should be the expected behavior tbh. I would've expected that if we deleted the history, the recently closed tab would be empty.

ekager:
So the STR to get into a weird state is to delete the history entries 1 by 1. If you "delete all" it will clear recently tabs. We can't hook these things together much closer because history entries are just a URL and recently closed tabs are actually entire sessions. 

I agree it's a bit weird and we probably need to think about the UX a bit more here and we have https://github.com/mozilla-mobile/fenix/issues/15088 and others open to address these further.
I think our current behavior is pretty bad here though because by not showing recently closed tabs in history when there are tabs, it implies you don't have anything there.

gabrielluong (approved):

Issue #13157, opened 2020-07-31
[Bug] you can place a bookmarks folder inside itself; deleting it after this does not work


s-ankur:
## Steps to reproduce
Make a bookmarks folder
Edit it using 3 dot menu
Set the folder to the folder itself
Save it and try to delete the folder
### Expected behavior
It shouldn't be possible to save inside itself
### Actual behavior
Deleting does not work; no bookmarks operations work at this point
### Device information

* Android device: Asus Zenfone
* Fenix version: Nightly

s-ankur:
Related #6977

grigoryk:
Trying this out, the move isn't actually executed (bookmark structure within places.sqlite is in-tact, folder remains under the mobile root). However, editing bookmarks does stop working. Restarting the app gets everything back to normal.

grigoryk:
Should be fixed by https://github.com/mozilla-mobile/fenix/pull/15203

AndiAJ:
Hi, **verified as fixed** on the latest **`Nightly Build 200921`** using the following devices:
• Google Pixel 3a (Android 11)
• Huawei Mate 20 Lite (Android 10)
• OnePlus A3 (Android 6.0.1)

PR #15203, opened 2020-09-18
For #13157: Don't allow bookmark folder to be moved into itself

grigoryk:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/res/navigation/nav_graph.xml
+++ b/app/src/main/res/navigation/nav_graph.xml
@@ -305,7 +305,7 @@
             android:id="@+id/action_bookmarkSelectFolderFragment_to_bookmarkAddFolderFragment"
             app:destination="@id/bookmarkAddFolderFragment" />
         <argument
-            android:name="visitedAddBookmark"
+            android:name="allowCreatingNewFolder"
             android:defaultValue="false"
             app:argType="boolean" />
     </fragment>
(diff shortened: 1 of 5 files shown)


app/src/main/res/navigation/nav_graph.xml
```diff
@@ -298,20 +298,21 @@
...
             android:id="@+id/action_bookmarkSelectFolderFragment_to_bookmarkAddFolderFragment"
             app:destination="@id/bookmarkAddFolderFragment" />
         <argument
-            android:name="visitedAddBookmark"
+            android:name="allowCreatingNewFolder"
             android:defaultValue="false"
             app:argType="boolean" />
+        <argument
+            android:name="hideFolderGuid"
+            android:defaultValue="null"
```
ekager, line 313: I don't think this is actually `null` I think it's setting the default value to "null". Try `@null`.

ekager (approved):
Other than nav arg, 👍

app/src/main/res/navigation/nav_graph.xml
grigoryk: Ah, thanks! It's right in the docs, too 👍 

So I've tried it both ways, and in both cases (`null` and `@null`) I get a real `null` value, not a "null" string. But let's go with what the docs say.

Issue #2792, opened 2019-05-23
Consider displaying bookmark folders at the top of the list


grigoryk:
Currently they're being displayed according to their order, but that doesn't work all that well on mobile. Navigating out of a folder takes you to the top of the list, which makes browsing multiple folders quite hard.

In Fennec, folders are pinned to the top, which makes things much more straightforward.

ekager:
I agree that Fennec's pinning is better. I was able to get all folders at the top by doing some sorting, but I wasn't sure how to handle the bookmark separators between a bookmark and a folder... We may need some UX feedback here.

mheubusch:
@AmyYLee Can you review this and point emily to any existing specs that address this if they exist? If not pls reassign to nicole so she can address with any other bookmark considerations.

AmyYLee:
Assigning this to @topotropic a since she had worked on bookmarks.

topotropic:
@ekager sounds good – can we sort the folders so that the latest created is at the top? 

Can you explain the issue with the separators – not sure if I understand it. Thanks!

ekager:
On Desktop we allow users to create/move/edit separators that they can place between bookmarks in a specific place. We represent those in Fenix but don't allow users to add, move, or edit them. If we start sorting bookmark folders to the top, the separators could end up in strange places. 

like if the order was 
- bookmark
- folder
- separator
- folder
- separator

and we moved all the folders to the top you would end up with 
- folder
- folder
- bookmark
- separator
- separator

From quick testing I'm not sure Fennec shows separators. Maybe that's the direction we should take as well.

lime124:
removed ux label as this is part of the q3 plans.

topotropic:
@vesta0 this is related to https://github.com/mozilla-mobile/fenix/issues/2961 so we should either close it as well or give it a lower priority

vesta0:
@topotropic this issue is about grouping the folders at the top before displaying the bookmarks (right now they are just added based on the order they are last created) so it's not the same issue but both issues are impacted by bookmark separators. @ekager says we can choose to not show that view type that include the separators. Would you look into the bookmarks separators and let us know if you think they add value from a user experience perspective, over the ability to sort folders and bookmarks?

lime124:
@topotropic and @ekager to collaborate on it.

grigoryk:
[...]

PR #15213, opened 2020-09-19
For #2792: Display bookmark folders on top of the bookmarks UI

grigoryk:
This is a naive (read: maybe slowish? a single full pass + array smushing) way to do this, but it's very simple. I think getting this to go faster would require storage API level changes, so it'll do for now.

### Pull Request checklist

1. click on Show All Checks,


ekager:
Don't we have to deal with getting rid of separators in this same issue/before these changes? Otherwise we'll have ugliness like this: https://github.com/mozilla-mobile/fenix/issues/2792#issuecomment-511971100

grigoryk:
@ekager sure, but we can do that as a follow-up. That's a fairly narrow use case.

I _think_ we can just hide separators from the UI, but I want to think about that a bit in terms of how it interacts with Sync, etc. Filed https://github.com/mozilla-mobile/fenix/issues/15214, I'll get to that on Monday.

ekager (approved):
okey

Issue #11800, opened 2020-06-22
[Bug]Saved logins error state icon misaligned


AndiAJ:
## Prerequisites
Have at least a saved login

## Steps to reproduce
1. Open "saved logins"
1. Open any login
1. Tap the upper right ⋮ "More options button"
1. Select Edit
1. Tap the ⓧ from the "Password" section
1. Check the red warning error state icon

### Expected behavior
Properly displayed, in place of the ⓧ icon

### Actual behavior
Misplaced error state icon

### Device information
* Android device: all
* Fenix version: Nightly Build 200622 from 6/22

### Notes
► Screenshot
![rsz](https://user-images.githubusercontent.com/51314259/85263776-55655180-b478-11ea-8bbb-c6c3e836569f.jpg)UNITO-UNDERSCORE!1rsz!UNITO-UNDERSCORE!img!UNITO-UNDERSCORE!20200622!UNITO-UNDERSCORE!104507!

Noticed while checking #10172

mcarare:
While investigating this, I've noticed that are a few more differences between implementation and mocks: [UX Mocks](https://share.goabstract.com/33afd8de-f90e-4504-90e3-ee2a3bcb1925?mode=design&sha=2b7a1bf44fa2c02a3b7e470075e94c587a491d3b
)

Are those mocks up to date, or the screen needs an implementation review and changes? cc @AmyYLee

ekager:
<img width="808" alt="Screen Shot 2020-09-14 at 11 19 52 AM" src="https://user-images.githubusercontent.com/25442310/93122967-3bed3380-f67c-11ea-902e-2b8e38bc9ab2.png">
For reference linked mocks above

joc-a:
Hello! I would like to work on this.

joc-a:
#15244

abodea:
Verified as fixed on the latest **Nightly 9/25** with **Google Pixel 4 XL (11).**

PR #15244, opened 2020-09-20
For #11800: Hide reveal and clear password icons if the password is empty when editing a saved login

joc-a:
### Pull Request checklist

1. click on Show All Checks,

Before fix - Light Mode
Before fix - Light Mode

Before fix - Dark Mode
Before fix - Dark Mode

After fix - Light Mode
After fix - Light Mode

After fix - Dark Mode
After fix - Dark Mode

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/settings/logins/fragment/EditLoginFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/logins/fragment/EditLoginFragment.kt
@@ -237,6 +237,7 @@ class EditLoginFragment : Fragment(R.layout.fragment_edit_login) {
                 usernameChanged = true
                 validUsername = false
                 it.setErrorIconDrawable(R.drawable.mozac_ic_warning)
+                it.setErrorIconTintList(ColorStateList.valueOf(ContextCompat.getColor(requireContext(), R.color.design_error)))
                 it.error = context?.getString(R.string.saved_login_duplicate)
             }
         } else {
--- a/app/src/main/java/org/mozilla/fenix/settings/logins/fragment/EditLoginFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/logins/fragment/EditLoginFragment.kt
@@ -236,9 +236,9 @@ class EditLoginFragment : Fragment(R.layout.fragment_edit_login) {
             inputLayoutUsername?.let {
                 usernameChanged = true
                 validUsername = false
+                it.error = context?.getString(R.string.saved_login_duplicate)
                 it.setErrorIconDrawable(R.drawable.mozac_ic_warning)
                 it.setErrorIconTintList(ColorStateList.valueOf(ContextCompat.getColor(requireContext(), R.color.design_error)))
-                it.error = context?.getString(R.string.saved_login_duplicate)
             }
         } else {
             usernameChanged = true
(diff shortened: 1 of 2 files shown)


ekager:
Hi @joc-a thanks for picking this up!

So we also set errors on the username state if you have 2 logins for a site and edit one login to match the other username (ie username dupe). Let's also verify that looks correct. 

Secondly, I think we also want to get the error icon vertically aligned with the clear icon/reveal icon to match the mocks. We might be able to do this through some drawable padding (either in the layout or we can create a new layer-list drawable to pad the error icon as we need).

Let me know if this makes sense and if you have any other questions!

joc-a:
Hi @ekager !

I aligned the error icon for the password. It now looks like this:
After fix - Light Mode
After fix - Dark Mode


I will do the same thing for the username but first I'd like to know what we want it to look like when the username is a duplicate. It currently looks like this:
Screenshot_1600721223


Also, do we want to allow saving an empty username?
Screenshot_1600721227

ekager:
Thanks @joc-a this is looking great!
That's a very good question! 🤔 cc @apbitner from our UX team and @eliserichards as the eng who originally worked on this to answer these 2 questions:
1. What should the error state of the username field look like with the error icon and clear icon displayed?
2. Do we want to allow empty usernames to be saved? @eliserichards do you remember?

apbitner:
> What should the error state of the username field look like with the error icon and clear icon displayed?
It seems to me that for the password error state we are not showing the clean icon and only showing the error icon, is that correct? If so, we might as well use the same pattern for username.

eliserichards:
> Thanks @joc-a this is looking great!
> That's a very good question! 🤔 cc @apbitner from our UX team and @eliserichards as the eng who originally worked on this to answer these 2 questions:
> 
>     1. What should the error state of the username field look like with the error icon and clear icon displayed?

The only error state for username is if the username entered would make the login a duplicate of another login that the user has saved. 

>     2. Do we want to allow empty usernames to be saved? @eliserichards do you remember?

Empty usernames are valid (as an empty string, it should never be`null`)

eliserichards:
I agree with @apbitner that we shouldn't show the username `clear` icon if there's a username error 👍

app/src/main/res/drawable/mozac_ic_warning_with_bottom_padding.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
```
ekager, line 1: nit - this needs license text at the top, look at any other xml file and copy it :)

app/src/main/java/org/mozilla/fenix/settings/logins/fragment/EditLoginFragment.kt
```diff
@@ -231,21 +239,26 @@ class EditLoginFragment : Fragment(R.layout.fragment_edit_login) {
...
         }
     }
 
     private fun setPasswordError() {
         inputLayoutPassword?.let { layout ->
             validPassword = false
             layout.error = context?.getString(R.string.saved_login_password_required)
-            layout.setErrorIconDrawable(R.drawable.mozac_ic_warning)
+            layout.setErrorIconDrawable(R.drawable.mozac_ic_warning_with_bottom_padding)
+            layout.setErrorIconTintList(ColorStateList.valueOf(ContextCompat.getColor(requireContext(), R.color.design_error)))
```
ekager, line 261: L244 and L261 are hitting the max line length error, you can autoformat with Android Studio or just manually break it into multiple lines!

ekager (changes requested):
Looking good! Just pulled it down and left 2 comments of things we need to change.

ekager (approved):
Looks good! Thanks for the contribution!

joc-a:
This was my first time contributing to Fenix and my second time contributing on GitHub. Thanks for being friendly and making it a pleasant experience! @ekager 
I do have a question: can external contributors rebase on master when they're done, or is that something you'd rather do internally?

ekager:
I can squash and merge when the PR passes CI!

Issue #15111, opened 2020-09-15
[Bug] Renaming Collections does not work


s-ankur:
## Steps to reproduce
Try to rename a collection
### Expected behavior
The change shows up
### Actual behavior
It does, but only after clicking on the collection
### Device information

* Android device: Asus Zenfone
* Fenix version: Nightly

AndiAJ:
Dupe of #13836

AndiAJ:
Hi, **verified as fixed** on the latest **`Nightly Build 201020`** using the following devices:
• Google Pixel 3a (Android 11)
• Huawei Mate 20 Lite (Android 10)
• OnePlus A3 (Android 6.0.1)

► Video
![20201020-085643](https://user-images.githubusercontent.com/51314259/96546377-d14e9980-12b2-11eb-8adb-b5d474dac43f.gif)

PR #15253, opened 2020-09-21
For #15111 - Update the UI when a collection is renamed

kleintom:
### Pull Request checklist

- [X] **Tests**: I think this would require a UI test and I think the Collections UI tests are currently disabled (https://github.com/mozilla-mobile/fenix/blob/master/app/src/androidTest/java/org/mozilla/fenix/ui/CollectionTest.kt.ignore), but I'm not sure about that - if there's something you'd like me to do please let me know, thanks.
- [X] **Screenshots**:
rename_lots

- [X] **Accessibility**: Only updating existing text.

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
@@ -140,6 +140,9 @@ class HomeFragment : Fragment() {
         }
 
         override fun onCollectionRenamed(tabCollection: TabCollection, title: String) {
+            lifecycleScope.launch(Main) {
+                view?.sessionControlRecyclerView?.adapter?.notifyDataSetChanged()
+            }
             showRenamedSnackbar()
         }
     }


app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -140,6 +140,9 @@ class HomeFragment : Fragment() {
         }
 
         override fun onCollectionRenamed(tabCollection: TabCollection, title: String) {
+            lifecycleScope.launch(Main) {
+                view?.sessionControlRecyclerView?.adapter?.notifyDataSetChanged()
```
ekager, line 144: Instead of calling `notifyDataSetChanged` directly on the adapter can we dispatch a collections change? `homeFragmentStore.dispatch(HomeFragmentAction.CollectionsChange(it))`

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -140,6 +140,9 @@ class HomeFragment : Fragment() {
         }
 
         override fun onCollectionRenamed(tabCollection: TabCollection, title: String) {
+            lifecycleScope.launch(Main) {
```
ekager, line 143: nit - let's use `viewLifecycleOwner.lifecycleScope.launch` to match the lifecycle of the fragment view

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
kleintom: Yes, dispatching a collections change is how the adapter is currently notified of the change, but via the observer installed on the tab collection storage here:
https://github.com/mozilla-mobile/fenix/blob/7d5c199e5188f1d8d553a4e732fbf9b649fd38d3/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt#L847-L850
rather than in onCollectionRenamed - both of those methods get called (on different threads) from the TabCollectionStorage code when a rename occurs.
The problem is, once we're in either of those methods, the TabCollection object has already been renamed. And since the collection that was renamed in TabCollectionStorage is the same (same object) as the collection that is currently sitting on the adapter list, when it got renamed in TabCollectionStorage it also got renamed on the adapter. So now calling the dispatch which ultimately pushes the new list of TabCollections to the adapter doesn't make the RecyclerView think it needs to update the UI, because the list currently sitting on its adapter already matches the updated collections list (data-wise (but no longer object-wise (but the new-objects list that was dispatched to the adapter is now setup for the same thing to happen again on the next rename, because it's the same - object-wise - as the list that will be sent to the rename fragment and from there to TabCollectionStorage))).

kleintom: I hadn't really thought of it explicitly before, but it's possible the notifyDataSetChanged() call I added and the existing dispatch linked above are racy.  Luckily it doesn't matter because, as I mentioned, the new dispatched list already matches the list currently on the adapter datawise, so either way, the changed collection name will appear whenever the notifyDataSetChanged() arrives.

gabrielluong:
Requesting a re-review to see how we should proceed.

liuche:
Let's also change the commit message to match #15480 which is the open bug for this issue. That way, no one will pick it up and duplicate work.

kleintom:
> Let's also change the commit message to match #15480 which is the open bug for this issue. That way, no one will pick it up and duplicate work.

Updated the commit message, thanks for the nudge.

ekager (approved):

Issue #13959, opened 2020-08-20
 StrictMode.resetPoliciesAfter – strict number of checks, and add performance team as code owners for the file


sraturi:
There are currently two versions

- resetPoliciesAfter(in fenix)
- resetAfter (outside of fenix)
- usages in GV (maybe we shouldn't worry about these right now & file a follow-up)

We want to restrict the number of calls to these methods to prevent regressions.

Archaeopteryx:
@mcomella Can you reland this? Despite what the retriggers of the task showed, the failure for which this got reverted was caused by the push which landed before.

mcomella:
@Archaeopteryx As far as I can tell, my changes weren't backed out – perhaps you have the wrong issue?

mcomella:
Addressed in https://github.com/mozilla-mobile/fenix/pull/15570

Mugurell:
I think this causes the app to crash when changing orientation with the following log:

> org.mozilla.fenix.debug E/ExceptionHandler: Uncaught exception handled: 
    java.lang.IllegalStateException: Failed to initialize GeckoRuntime
        at org.mozilla.geckoview.GeckoRuntime.create(GeckoRuntime.java:490)
        at GeckoProvider.createRuntime(GeckoProvider.kt:56)
        at GeckoProvider.getOrCreateRuntime(GeckoProvider.kt:28)
        at org.mozilla.fenix.components.Core$engine$2.invoke(Core.kt:108)
        at org.mozilla.fenix.components.Core$engine$2.invoke(Core.kt:81)
        at kotlin.SynchronizedLazyImpl.getValue(LazyJVM.kt:74)
        at org.mozilla.fenix.components.Core.getEngine(Unknown Source:2)
        at org.mozilla.fenix.FenixApplication.onConfigurationChanged(FenixApplication.kt:484)

the problem seemingly stemming from [components.core.engine.profiler?.addMarker(..)](https://github.com/mozilla-mobile/fenix/blob/4302cd20b225790feffacd3e6277a748d3895cbf/app/src/main/java/org/mozilla/fenix/StrictModeManager.kt#L130)
@mcomella can you please check?

mcomella:
Thanks. I confirmed that I see the same behavior so I opened https://github.com/mozilla-mobile/fenix/issues/15751

PR #15329, opened 2020-09-22
For #13959: clean up StrictMode code

mcomella:
Pre-work for heavy lifting in #13959.

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/StrictModeManager.kt
+++ b/app/src/main/java/org/mozilla/fenix/StrictModeManager.kt
@@ -46,11 +46,11 @@ object StrictModeManager {
     }
 
     /**
-     * Revert strict mode to disable penalty. Tied to fragment lifecycle since strict mode
+     * Revert strict mode to disable penalty based on fragment lifecycle since strict mode
      * needs to switch to penalty logs. Using the fragment life cycle allows decoupling from any
      * specific fragment.
      */
-    fun changeStrictModePolicies(fragmentManager: FragmentManager) {
+    fun attachListenerToDisablePenaltyDeath(fragmentManager: FragmentManager) {
         fragmentManager.registerFragmentLifecycleCallbacks(object :
             FragmentManager.FragmentLifecycleCallbacks() {
             override fun onFragmentResumed(fm: FragmentManager, f: Fragment) {
--- a/app/src/main/java/org/mozilla/fenix/StrictModeManager.kt
+++ b/app/src/main/java/org/mozilla/fenix/StrictModeManager.kt
@@ -11,6 +11,10 @@ import androidx.fragment.app.FragmentManager
 
 /**
  * Manages strict mode settings for the application.
+ *
+ * Due to the static dependencies in this class, it's getting hard to test: if this class takes any
+ * many more static dependencies, consider switching it, and other code that uses [StrictMode] like
+ * [StrictMode.ThreadPolicy].resetPoliciesAfter, to a class with dependency injection.
  */
 object StrictModeManager {
 
@@ -19,6 +23,9 @@ object StrictModeManager {
      * @param setPenaltyDeath boolean value to decide setting the penaltyDeath as a penalty.
      */
     fun enableStrictMode(setPenaltyDeath: Boolean) {
+        // The expression in this if is duplicated in StrictMode.ThreadPolicy.resetPoliciesAfter
+        // because the tests break in unexpected ways if the value is shared as a constant in this
+        // class. It wasn't worth the time to address it.
         if (Config.channel.isDebug) {
             val threadPolicy = StrictMode.ThreadPolicy.Builder()
                 .detectAll()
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/mozilla/fenix/StrictModeManager.kt
```diff
@@ -47,26 +44,22 @@ object StrictModeManager {
                 builder.detectContentUriWithoutPermission()
             }
             if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
-                if (setPenaltyDeath || setPenaltyDialog) {
-                    builder.permitNonSdkApiUsage()
-                } else {
-                    builder.detectNonSdkApiUsage()
-                }
+                builder.detectNonSdkApiUsage()
```
MarcLeclair, line 47: So, I'm not sure why we added this exactly. But I remember Sachin added this check because Firebase kept screwing up with random StrictMode violations with the `detectNonSdkApiUsage`. Do you think we should let this one be and if Firebase has issues we'll change it then?

mcomella: I mention this in the extended commit summary for the commit that made this change:

> We don't use penalty death for the VM policy so we theoretically don't
need to disable this check if penalty death is enabled.

Since there is no comment explaining why this theory doesn't make sense, we can't know whether this is true or not without looking at the issue https://github.com/mozilla-mobile/fenix/issues/7225 and PR https://github.com/mozilla-mobile/fenix/pull/10831 – but I don't see any info there – or trying to change it. Given we have no information, I'd rather try to clean up the code than leaving around "if we touch it, it may break!" type issues, gradually increasing the complexity of the code (e.g. remember that bug we fixed where we unnecessarily `setAdapter(null); setAdapter(content)` on every data refresh? Imagine if someone cleaned that up for us already because they didn't know what it was and figured it was unnecessary).

That being said, it's about cleaning up gradually – we're making a handful of changes here so we could make one fewer and clean this up later. It's up to you.

mcomella:
After this patch, when I run a subset of the tests locally, they fail:
```
    StrictModeManagerTest. test enableStrictMode when can be enabled
    StrictModeManagerTest. test enableStrictMode when cannot be enabled
```

We should avoid landing this until this is resolved.

mcomella:
git bisect indicates the error occurred in f77ec9e651e1d1938d74ff00370f5ba4d716f5f6.

mcomella:
It looks like Robolectric might be dying when trying to initialize StrictMode:
```
Caused by: java.lang.NullPointerException: classInstanceLimit == null
	at android.os.StrictMode$VmPolicy.__constructor__(StrictMode.java:726)
	at android.os.StrictMode$VmPolicy.<init>(StrictMode.java)
	at android.os.StrictMode$VmPolicy.__staticInitializer__(StrictMode.java:711)
	at org.robolectric.internal.bytecode.RobolectricInternals.performStaticInitialization(RobolectricInternals.java:61)
```

I don't really like relying on robolectric or static mocking due to the black magic and errors like these – perhaps we should move to a dependency injection model.

mcomella:
It's weird though – my changes in that commit didn't really change this functionality. Maybe it's an intermittent failure or works only when you run the whole test suite.

mcomella:
If I drop that commit, the tests pass – I guess there's something broken with robolectric + the static mocking for that particular combination of code (maybe it's because the `Config` value is set in the initializer instead of in a function call?).

mcomella:
I switched `canBeEnabledByBuildConfig` to a property with a getter function and it worked correctly: I guess something just breaks when it's assigning to a property/field directly? The error occurs even when the assignment is a constant rather than `Config.channel.isDebug`

mcomella:
I ended up dropping the commits – it's too hard to fix the tests.

app/src/main/java/org/mozilla/fenix/StrictModeManager.kt
MarcLeclair: I agree with that reasoning. The `setAdapter` is a good example. Since we have context now, it might be better to get it out since it seems to not affect our code. I vote to get it removed since it did work locally on an emulated pixel 2 with Android P, so this works for me.

MarcLeclair:
@mcomella  a thought about the mocking, maybe it might be because of the kotlin to Java and Robo and static Mocking doesn't digest it well? Not sure though, cause we do it in other tests

MarcLeclair (approved):
Looks good to me. Tested on a P2  emulator and it worked as expected. Introduced a Strict Mode violation and it crashed both on P2 emulator and G5 as expected

mcomella:
> @mcomella a thought about the mocking, maybe it might be because of the kotlin to Java and Robo and static Mocking doesn't digest it well? Not sure though, cause we do it in other tests

I think it's probably a clash of robolectric + static mocking, since they're trying to interact with the same instances (if they're even instances). It seems less crazy with instance mocking because at least you have a separate instance (i.e. whatever Robolectric overrides, mocking is likely to override that; with static, it probably depends more on call order). I have no idea – mocking is nuts.

PR #15465, opened 2020-09-26
Closes #15410: Ripple Animations on Home Screen Toolbar actions

### LINKED ISSUE #15410: [Bug] Home Toolbar actions have no ripple animations on click
## Steps to reproduce
1. On home, click the tab counter or menu icon on toolbar

### Expected behavior
See ripple that matches ripple while browsing

### Actual behavior
No ripple or indication they were clicked. 

### Device information

* Android device: ?
* Fenix version: ?

philippbara:
Closes #15410: Ripple Animations on Home Screen Toolbar actions:
Added on-click ripple effect for both buttons in the home screen toolbar (tab selection + 3-dot menu). See screenshots below.
Also works during private browsing + Pre-push hook tests all succeeded

Screenshot from 2020-09-26 23-03-54

Screenshot from 2020-09-26 23-04-11

### Pull Request checklist

1. click on Show All Checks,


ekager (changes requested):
Thanks! Let's use `selectableItemBackgroundBorderless` so the ripple will extend past the square of the icon. :)

philippbara:
When using `selectableItemBackgroundBorderless` the ripple is "hidden" by the background of the View element (see screenshot). The only way I found to make the ripple visible is by removing the `android:background="?bottomBarBackground"`line from the View element which I guess is not wanted. I found the issue that might be causing this:  

https://github.com/googlecodelabs/constraint-layout/issues/50

Maybe the View's background (`@drawable/home_bottom_bar_background`) could be changed somehow?
If not I would reconsider the "square ripple" :-)

Thanks for any advice

Screenshot from 2020-09-29 20-08-23

gabrielluong:
> When using `selectableItemBackgroundBorderless` the ripple is "hidden" by the background of the View element (see screenshot). The only way I found to make the ripple visible is by removing the `android:background="?bottomBarBackground"`line from the View element which I guess is not wanted. I found the issue that might be causing this:
> 
> [googlecodelabs/constraint-layout#50](https://github.com/googlecodelabs/constraint-layout/issues/50)
> 
> Maybe the View's background (`@drawable/home_bottom_bar_background`) could be changed somehow?
> If not I would reconsider the "square ripple" :-)
> 
> Thanks for any advice
> 
> Screenshot from 2020-09-29 20-08-23

@ekager any advice here?

pocmo:
Could we consider using the actual toolbar here? I think some time ago we added all needed functionality in AC for that. And those are exactly the issues we fixed in the toolbar - we kind of relive all of that again here. :)

mcarare:
This PR is obsolete, as the issue is fixed in AC.

Issue #13270, opened 2020-08-04
[Bug] Two Firefox tasks in recent app switcher (permanently on Android 11 Beta; briefly on cold startup in Android 10)


mcomella:
Originally reported by our community in https://github.com/mozilla-mobile/fenix/issues/12577#issuecomment-667085187

## Steps to reproduce
- Cold start Firefox and very quickly...
- Hit the recent app switcher button

### Expected behavior
One Firefox task is present in the recent app switcher

### Actual behavior
Two Firefox tasks are present in the recent app switcher. The second task (it's inactive) will quickly dismiss itself.

![image](https://user-images.githubusercontent.com/68282543/87494267-2e133600-c66c-11ea-8fa2-5541324da577.jpg)

Our community uploaded recordings of the issue:
- https://github.com/mozilla-mobile/fenix/files/5006834/Screen.Record.2.apps.zip
- https://github.com/mozilla-mobile/fenix/files/5006840/App.isn.t.available.error.zip
  - In this case, a toast says, "App is not available". I suspect it's because the recent app switcher button was hit twice – i.e. we were trying to switch to the second Firefox task that is now dismissed

### Device information
* Android device: P2
* Fenix version: migration RC1 or 2 (can't remember); I'm unable to reproduce on Nightly 200804

---

I would want someone with more task experience to chime in but it _seems_ harmless so I'd be tempted to set a low priority. @pocmo Do you agree that this is harmless?

Given I can't reproduce on the Nightly but can reproduce on the migration build, I suspect this may be related to the migration code. I noticed a discrepancy in taskAffinity: all the fenix activities have `taskAffinity=""` except for the activities/alias in `.../migration/AndroidManifest.xml` which have `taskAffinity="${applicationId}.BROWSER"` ([search](https://searchfox.org/mozilla-mobile/search?q=taskaffinity&path=)) so perhaps that's related.

pocmo:
I agree that this should be harmless.

The task affinity may be the cause here. Looking at the commits I noticed that I introduced this here: https://github.com/mozilla-mobile/fenix/commit/f1a3dffba27df72b34836a2fb7b0c4a668586faa - This was a workaround to launch into the Fennec task that still could be around during the migration (It's weird). Maybe at some point after the migration we can just get rid of that again.

jawz101:
This isn't brief for me.  The second instance won't go away in its own

mcomella:
> This isn't brief for me. The second instance won't go away in its own.

[...]

PR #15496, opened 2020-09-28
[Uplift] Issue #13270: Do not launch into Fennec's task anymore.

pocmo:
Uplifting #15345 to 82.0 (Beta).


st3fan (approved):
Previously approved for landing on master by @ekager - same patch but now for Beta.

PR #15500, opened 2020-09-28
Uplifting mls revert

boek:


st3fan (approved):
Looks good - same patch as https://github.com/mozilla-mobile/fenix/pull/15161

PR #15511, opened 2020-09-28
For #13959: change StrictModeManager into class; move resetAfter into it

mcomella:
Pre-work for adding suppression counts, which adds state to StrictModeManager, which is easier to manage and test in a class.

### Pull Request checklist

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/components/Components.kt
```diff
@@ -105,13 +108,14 @@ class Components(private val context: Context) {
...
 
-    val analytics by lazy { Analytics(context) }
+    val analytics by lazy { Analytics(context, strictMode) }
     val publicSuffixList by lazy { PublicSuffixList(context) }
     val clipboardHandler by lazy { ClipboardHandler(context) }
     val migrationStore by lazy { MigrationStore() }
     val performance by lazy { PerformanceComponent() }
     val push by lazy { Push(context, analytics.crashReporter) }
     val wifiConnectionMonitor by lazy { WifiConnectionMonitor(context as Application) }
+    val strictMode by lazy { StrictModeManager(Config) }
```
MarcLeclair, line 118: I'm wondering if having strictMode in `Components` makes sense? I was wondering,  could we keep it elsewhere since it isn't really a component? Maybe I'm just being pedantic, I'm not sure. It just feels odd (at least to me) to stack it in here.

mcomella: Why do you think it's not a component? I think a `StrictModeManager` is a component.

Also, the pattern `Components` is modeled after is the Service Locator pattern (I think it should be named as such but it's not...) and in that pattern, all "static" objects generally live on the Service Locator.

MarcLeclair: First, the fact that it is modeled after the Service Locator pattern makes much more sense, I took it by literal meaning to represent everything from Android Component (hence the name).

However, and I'm not sure if this will make sense, couldn't we have one Performance component and keep everything there, especially since we're using StrictMode mostly for performance reasons?

mcomella: Trying to use the terminology I think fenix would follow, a component is just a class that has ideally one responsibility (i.e. in `components.strictMode`, the component is `strictMode` not `components`). Following best practices, for easiest development, each component should be small and the components should be composable to achieve more complex purposes. This component does one thing – manage StrictMode – and combining it with a larger performance component works against that.

We do have a component sub-tree (not sure what to call this) for Performance – https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/java/org/mozilla/fenix/components/PerformanceComponent.kt (I think this is misnamed) – and we could conceivably put it there but StrictMode is used for things that aren't performance related – e.g.
```kotlin
                .detectLeakedSqlLiteObjects()
                .detectLeakedClosableObjects()
                .detectLeakedRegistrationObjects()
                .detectActivityLeaks()
                .detectFileUriExposure()
```

so I opted to put it in a general place instead.

mcomella:
Turns out I forget to test by starting the app: we use suppression too early in the lifecycle, before the Context is ready, so this crashes. I'll have to come up with another solution.

MarcLeclair (approved):
Changes look good. Works on my g5 and p2

mcomella:
re test failure: tests passed locally when I pushed and that specific test runs locally so I think it must be an intermittent. I'm going to re-run and merge.

PR #15564, opened 2020-09-30
Add youtube-playback 2020 tests

alexandru-io:


gmierz:
r- because some tests are gone now with no explanation: `youtube-playback-h264`, `youtube-playback-widevine-h264-sfr`, `youtube-playback-widevine-vp9-sfr`

Issue #15555, opened 2020-09-29
[Bug] Intermittent failure debug(T) org.mozilla.fenix.settings.SettingsFragmentTest Add-on collection override pref is visible if debug menu active


Archaeopteryx:
Regression from #14034 cc @csadilek 

Log: https://firefoxci.taskcluster-artifacts.net/TU1hZI6jRDe1oRhTXHY9nA/0/public/logs/live_backing.log

```
[task 2020-09-29T23:06:26.913Z] SUITE: org.mozilla.fenix.settings.SettingsFragmentTest
[task 2020-09-29T23:06:26.913Z]   TEST: Add-on collection override pref is visible if debug menu active
[task 2020-09-29T23:06:26.913Z]     [Robolectric] org.mozilla.fenix.settings.SettingsFragmentTest.Add-on collection override pref is visible if debug menu active: sdk=28; resources=LEGACY
[task 2020-09-29T23:06:26.913Z]     [Robolectric] NOTICE: legacy resources mode is deprecated; see http://robolectric.org/migrating/#migrating-to-40
[task 2020-09-29T23:06:26.913Z]     I/AssetManager: no overlays.list file found
[task 2020-09-29T23:06:26.913Z]     I/MonitoringInstr: Instrumentation started!
[task 2020-09-29T23:06:27.013Z]     D/LifecycleMonitor: Lifecycle status change: androidx.fragment.app.FragmentActivity@46072b10 in: PRE_ON_CREATE
[task 2020-09-29T23:06:27.013Z]     D/LifecycleMonitor: Lifecycle status change: androidx.fragment.app.FragmentActivity@46072b10 in: CREATED
[task 2020-09-29T23:06:27.013Z]     V/FragmentManager: Run: BackStackEntry{295e30ed}
[task 2020-09-29T23:06:27.013Z]     V/FragmentManager: add: SettingsFragment{647d69f7} (136fd6a8-f26c-4da4-a27d-dbf5317a2307) test}
[task 2020-09-29T23:06:27.013Z]     V/FragmentManager: Added fragment to active set SettingsFragment{647d69f7} (136fd6a8-f26c-4da4-a27d-dbf5317a2307) test}
[task 2020-09-29T23:06:27.013Z]     D/FragmentManager: moveto ATTACHED: SettingsFragment{647d69f7} (136fd6a8-f26c-4da4-a27d-dbf5317a2307) test}
[...]

PR #15570, opened 2020-09-30
For #13959: add test to count StrictMode suppressions, lint check

mcomella:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/StrictModeManager.kt
+++ b/app/src/main/java/org/mozilla/fenix/StrictModeManager.kt
@@ -2,6 +2,11 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+// This class implements the alternative ways to suppress StrictMode with performance
+// monitoring by wrapping the raw methods. This lint check tells us not to use the raw
+// methods so we suppress the check.
+@file:Suppress("MozillaStrictModeSuppression")
+
 package org.mozilla.fenix
 
 import android.os.Build
@@ -26,6 +31,14 @@ class StrictModeManager(config: Config) {
     // This is public so it can be used by inline functions.
     val isEnabledByBuildConfig = config.channel.isDebug
 
+    /**
+     * The number of times StrictMode has been suppressed. StrictMode can be used to prevent main
+     * thread IO but it's easy to suppress. We use this value, in combination with:
+     * - a test: that fails if the suppression count increases
+     * - a lint check: to ensure this value gets used instead of functions that work around it
+     * - code owners: to prevent modifications to these above items without perf knowing
+     * to make suppressions a more deliberate act.
+     */
     @VisibleForTesting(otherwise = PRIVATE)
     var suppressionCount: Long = 0
 
@@ -78,12 +91,14 @@ class StrictModeManager(config: Config) {
     /**
      * Runs the given [functionBlock] and sets the given [StrictMode.ThreadPolicy] after its
      * completion when in a build configuration that has StrictMode enabled. If StrictMode is
-     * not enabled, simply runs the [functionBlock].
+     * not enabled, simply runs the [functionBlock]. This function is written in the style of
+     * [AutoCloseable.use].
      *
-     * This function is written in the style of [AutoCloseable.use].
-     *
-     * This is significantly less convenient to run than when it was written as an extension function
-     * on [StrictMode.ThreadPolicy] but I think this is okay: it shouldn't be easy to ignore StrictMode.
+     * This function contains perf improvements so it should be
+     * called instead of [mozilla.components.support.ktx.android.os.resetAfter] (using the wrong
+     * method should be prevented by a lint check). This is significantly less convenient to run than
+     * when it was written as an extension function on [StrictMode.ThreadPolicy] but I think this is
+     * okay: it shouldn't be easy to ignore StrictMode.
      *
      * @return the value returned by [functionBlock].
      */
--- a/app/src/main/java/org/mozilla/fenix/StrictModeManager.kt
+++ b/app/src/main/java/org/mozilla/fenix/StrictModeManager.kt
@@ -6,9 +6,12 @@ package org.mozilla.fenix
 
 import android.os.Build
 import android.os.StrictMode
+import androidx.annotation.VisibleForTesting
+import androidx.annotation.VisibleForTesting.PRIVATE
 import androidx.fragment.app.Fragment
 import androidx.fragment.app.FragmentManager
 import mozilla.components.support.ktx.android.os.resetAfter
+import org.mozilla.fenix.perf.Performance
 
 private const val MANUFACTURE_HUAWEI: String = "HUAWEI"
 private const val MANUFACTURE_ONE_PLUS: String = "OnePlus"
@@ -18,8 +21,13 @@ private const val MANUFACTURE_ONE_PLUS: String = "OnePlus"
  */
 class StrictModeManager(config: Config) {
 
+    val logger = Performance.logger // public to be accessible by inline functions.
+
     // This is public so it can be used by inline functions.
     val isEnabledByBuildConfig = config.channel.isDebug
+
+    @VisibleForTesting(otherwise = PRIVATE)
+    var suppressionCount: Long = 0
 
     /***
      * Enables strict mode for debug purposes. meant to be run only in the main process.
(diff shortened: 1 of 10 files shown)


csadilek:
> Test failure is intermittent test failure: #15555

This was fixed last night, but this branch needs to be rebased to bring in the fix. 🤞

app/src/main/java/org/mozilla/fenix/StrictModeManager.kt
```diff
@@ -70,19 +103,30 @@ class StrictModeManager(config: Config) {
...
         // Calling resetAfter takes 1-2ms (unknown device) so we only execute it if StrictMode can
         // actually be enabled. https://github.com/mozilla-mobile/fenix/issues/11617
         return if (isEnabledByBuildConfig) {
+            // This can overflow and crash. However, it's unlikely we'll suppress StrictMode 9
+            // quintillion times in a build config where StrictMode is enabled so we don't handle it
+            // because it'd increase complexity.
+            suppressionCount += 1
+            logger.warn("StrictMode violation suppressed: #$suppressionCount")
+            if (Thread.currentThread() == mainLooper.thread) { // markers only supported on main thread.
+                components.core.engine.profiler?.addMarker("StrictMode.suppression", "Count: $suppressionCount")
```
MarcLeclair, line 127: Just to make sure, are we adding a marker to be able to see the count # on the profiler? 

If so, I think we should document that somewhere to highlight Logcat vs marker logging. I don't think adding a comment here would be worthwhile since it isn't the goal of the method. However, I could see a someone coming across this code and wondering why we're using `profiler?.addMarker` when we use `logger.warn` right before. 

I'm not sure if my thought made sense, I can try to re-explain it if it didn't!

app/src/main/java/org/mozilla/fenix/StrictModeManager.kt
```diff
@@ -2,25 +2,58 @@
...
-class StrictModeManager(config: Config) {
+class StrictModeManager(
+    config: Config,
+
+    // Ideally, we'd pass in a more specific value but there is a circular dependency: StrictMode
+    // is passed into Core but we'd need to pass in Core here. Instead, we take components and later
+    // fetch the value we need from it.
+    //
+    // Public to be accessible by inline functions.
+    val components: Components
```
MarcLeclair, line 37: I was looking at `Components.kt` and what the other components seem to do is just pass `context`. Should we do that here and access `Profiler` from it?  I can see how that would be redundant though since `context.components....` would be similar to just passing `this` from components.

mcomella: I think that's an anti-pattern. With dependency injection style patterns, you should pass in the most specific object you can. The problem with passing in objects with larger scope, is that:
1) you give the class access to a large number of classes. By having access to a wider range of classes, your class can (in theory) change much more about the program. By limiting the objects that are passed in, it's easier for devs to make analyses when looking at the code.
2) your test code is more fragile because it relies on accessing objects several levels down and any of the upper layers can be re-factored
3) ^ for similar reasons, your code is less explicit
4) passing in context more closely ties your code to the Android framework and the Android framework is difficult to test

app/src/main/java/org/mozilla/fenix/StrictModeManager.kt
mcomella: > Just to make sure, are we adding a marker to be able to see the count # on the profiler?

I thought it would be useful, when we're analyzing start up performance, to be able to see all of the places that we access the disk or the network on the main thread.

> I'm not sure if my thought made sense, I can try to re-explain it if it didn't!

@MarcLeclair could you elaborate further? Maybe it'd be good to talk about in stand up. I honestly don't know how I feel about both adding a marker and logging yet. The decision to do both was confusing for me.

mozilla-detekt-rules/src/main/java/org/mozilla/fenix/detektrules/MozillaStrictModeSuppression.kt
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    private fun reportIfImportAcResetAfter(importDirective: KtImportDirective) {
+        if (importDirective.importPath?.toString() == "mozilla.components.support.ktx.android.os.resetAfter") {
+            report(CodeSmell(issue, Entity.from(importDirective), VIOLATION_MSG))
+        }
+    }
+
+    private fun reportIfCallStrictModeSetPolicy(expression: KtCallExpression) {
+        // There is probably a more correct way of doing this but this API is not well documented
```
MarcLeclair, line 50: @mcomella  what do you think about using ```override fun visitExpression(expression: KtExpression)```? I think this just goes through all the expressions in a file. So we could do something like : 

``` 
     if(expression.containingKtFile.name.contains("StrictModeManager")){
            if(expression.children.filter{ it.text == "setVmPolicy" 
                        || it.text ==   "setThreadPolicy"}.isNotEmpty()){ // make error here }
```

I'm not sure if its better, I just wasn't sure about looking for the receiver and caller method with the ```?``` .  Though, I only came up with this looking through the methods and playing with detekt.

app/src/main/java/org/mozilla/fenix/StrictModeManager.kt
MarcLeclair: @mcomella  I think that having both statement ```logger.warn``` and ```profiler?.addMarker``` can be confusing for someone looking at how we did things. I was wondering if we should document somewhere why we used the marker API here along with the Logger.

MarcLeclair (approved):
These changes worked locally. I think all I have are mostly questions about the detekt API (or rather, an idea but it's just from playing around with detekt trial and error style)

mozilla-detekt-rules/src/main/java/org/mozilla/fenix/detektrules/MozillaStrictModeSuppression.kt
mcomella: To address specific concerns I see with the example above,
1. With the visitor pattern, I think it's preferred to use the most specific node type that you can. If you use a more generic node type then you'll have to do that same filtering yourself and you can mess it up. For example, I believe your code would throw a violation on an assignment expression – `setVmPolicy = true`.
1. If you operate on all children rather than a specific child, then you could end up violating on the wrong expression. For example, in a hypothetical method, `strictModeManager.setVmPolicy.resetSuppressionCount()` – I believe your code would also violate here.

> I'm not sure if its better, I just wasn't sure about looking for the receiver and caller method with the ? . 

I don't think the goal should be to remove every optional type. Optional types are intended to express the absence of something and it's logical to use them in this capacity. The problem with optional types is that many programmers use them to express default values, edge case states, etc.

In this case, I think the optional types are used to express the absence of something. For example,
```
val receiver = expression.parent?
```

If I understand the API correctly, the callExpression means a function being invoked, while the parent is the receiver of that function. In some cases the receiver doesn't exist. For example, you can call `runBlocking { }`. Therefore, I think the comparison between `receiver == "StrictMode"` is fine given if there's no receiver, we're unlikely to be calling `setVmPolicy`.

I don't really understand why `calleeExpression` can be nullable though and that's a little concerning (maybe infix functions?). I'm not sure what to do about that though.

---

Your comment pointed out a bug in my code though – if setVmPolicy is imported directly, we won't catch it because I look for the StrictMode receiver. I'll add a comment about this and perhaps when we're adding a test we can go back and address this. I don't trust myself to write this correctly without tests (though I guess it could be an import statement block...).

app/src/main/java/org/mozilla/fenix/StrictModeManager.kt
mcomella: Sure, I'll add a comment.

mcomella:
The test failure is:
>  java.lang.AssertionError: StrictMode startup suppression count does not match expected count. 
>  expected:<12> but was:<11> 

On Pixel 2 virtual, API 28. It's strange b/c I have a P2 locally – maybe the test doesn't wait long enough after startup to catch the full count?

mcomella:
Or some state is carried over from a previous test?

MarcLeclair:
@mcomella  could you try to hard code a reset count and see how much we get from that test? I think we can provide the output in the test right?

mozilla-detekt-rules/src/main/java/org/mozilla/fenix/detektrules/MozillaStrictModeSuppression.kt
MarcLeclair: That makes sense to me. Its a bit difficult to know what's happening with the rule, that was another thing that was concerning. But I think that mostly has to do with Detekt APIs being so mysterious and lacking document, so I guess we can just add this part to our Detekt guide? I think the two points you mentioned are very valid and they aren't something I would've thought of. My first thought was "how can I get access to the ```StrictMode``` call?" which seems hard to do if you just look at the possible ```visit()``` methods and not all of them have a super clear name on what they do.

gabrielluong (approved):

Issue #15471, opened 2020-09-27
[Bug] Delete button in multi select three dot menu is not red 


s-ankur:
## Steps to reproduce
Open multi select in history or bookmarks
Open 3 dot menu
### Expected behavior
The delete button is red, which is the case for the delete button in every other 3dot menu
### Actual behavior
It is normal
### Device information

* Android device: Asus Zenfone
* Fenix version: Nightly

joc-a:
I'd like to work on this.

joc-a:
#15576

sflorean:
Verified as fixed on Nightly 10/7 with Google Pixel 3 (Android 11) and Xiaomi Redmi Note 8T (Android 9).

PR #15576, opened 2020-09-30
For #15471: Show Delete button in red in multi-select overflow menu

joc-a:
### Pull Request checklist

1. click on Show All Checks,

Bookmarks:
Bookmarks

History:
History

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkFragment.kt
@@ -6,6 +6,7 @@ package org.mozilla.fenix.library.bookmarks
 
 import android.content.DialogInterface
 import android.os.Bundle
+import android.text.SpannableString
 import android.view.LayoutInflater
 import android.view.Menu
 import android.view.MenuInflater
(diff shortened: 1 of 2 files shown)


ekager (changes requested):
I think we could use an already-created extension method we have here: see `SpannableString.kt`, we have a method `SpannableString.setTextColor` - would that work?

app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkFragment.kt
```diff
@@ -44,12 +45,7 @@ import org.mozilla.fenix.R
 import org.mozilla.fenix.components.FenixSnackbar
 import org.mozilla.fenix.components.StoreProvider
 import org.mozilla.fenix.components.metrics.Event
-import org.mozilla.fenix.ext.bookmarkStorage
```
ekager, line 47: One more nit and then we're there! Our lint checks don't allow wildcard imports so these all need to be declared separately. The IDE probably auto changed this to try to be smart :/ but you can change the Android Studio settings for this in Preferences -> Editor -> Code style -> Kotlin and then select "Use single name import" 👍

ekager (changes requested):
One lint thing to fix and then we're good!

app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkFragment.kt
joc-a: Single name imports it is! 😊

ekager:
bors try

ekager (approved):
Looks great!

ekager:
bors r=ekager

PR #15582, opened 2020-09-30
Issue #15555: Set channel to Nightly explicitly for SettingsFragmentTest

csadilek:
This is to fix the intermittent test failure.


Amejia481 (approved):

app/src/test/java/org/mozilla/fenix/settings/SettingsFragmentTest.kt
```diff
@@ -41,6 +44,9 @@ class SettingsFragmentTest {
         // Mock client for fetching account avatar
         val client = testContext.components.core.client
         every { client.fetch(any()) } throws IOException("test")
+
+        mockkObject(Config)
+        every { Config.channel } returns ReleaseChannel.Nightly
```
Amejia481, line 49: Tests from other classes can't run in parallel , right? All of the tests in this class have to run first before any other test from another class can run?

Issue #15586, opened 2020-10-01
[Bug] Leanplum deviceId is not migrated


st3fan:
The full story can be read at https://github.com/mozilla-mobile/fenix/issues/15392#issuecomment-701740271

The short version is that Leanplum does not allow the `deviceId` to be changed once it has been set. This is a problem for migrated users because for those users the `deviceId` is actually properly migrated over through the Leanplum shared preferences. But since we did not know we cannot actually change the `deviceId`, the new one that we generate is simply ignored.

HOWEVER, we do use that `deviceId` that _we_ generated to check if the `uid` parameter in deep links matches.

I think a simple fix here can be to check if the `org.mozilla.gecko.BrowserApp` preferences exist and then transfer the `android.not_a_preference.leanplum.device_id` entry over to our new `LEANPLUM_PREFERENCES` / `LP_DEVICE_ID` pref.

grigoryk:
We could do this as a follow-up migration, so that this code only runs for affected users (well, that's probably _most_ users...), and only runs once. Not sure how "once" is important, since as you say, this ID isn't allowed to change, and so we don't expect leanplum to change it after we'll set it back to what it was in Fennec.

st3fan:
Closing this issue as we understand this problem better now. I will file a new one soon.

Issue #15392, opened 2020-09-24
Verify that the Leanplum Device Identifier was properly migrated or generated


st3fan:
I am debugging a case where Leanplum is:

- properly initialized
- has a push token
- can receive remote notification messages and in-app messages

But what fails is opening deeplinks, specifically the following two:

 - `fenix://open?url=https%3A%2F%2Fsupport.mozilla.org%2Fen-US%2Fkb%2Fwhats-new-firefox-android%2526utm_source%3Dleanplum-push-notification%2526utm_campaign%3D09.28.20-EN-fenix81-qa&uid={{User`
 - `fenix://settings?uid={{User ID}}`

Note that these work properly for me (@st3fan) but not for @cbonacuse 

In the Leanplum console I see that both our test devices are up to date (Firefox 81.1.1) and both have a push token assigned.

Since this works well for me, end-to-end, that gives me confidence that the `LeanplumMetricsService` and `fenix://` intent handling works as expected.  However, we did make two changes relatively recently related to the `uid={{User ID}}` part of those URLs that I am trying to rule out:

 - We used to (incorrectly) randomize the device identifier at every app cold start. This was fixed by caching the device identifier. See https://github.com/mozilla-mobile/fenix/pull/13230

- We used to not check the `uid` parameter in the deeplink URL, which was fixed with https://github.com/mozilla-mobile/fenix/commit/77263aad704844782a461b49a92b51a8bed31c16

It is important to rule these out because if push notification messages appear but the links are rejected that will most likely mean that the check for the device identifier fails.

This issue is to re-read / re-review those two patches and think about edge cases that could cause the caching or the check to fail. Keeping in mind the following scenarios:

 - Migration
 - A user upgrading from a Fenix version that did not have these changes

cc to @boek and @pocmo

st3fan:
Looking at the device identifier caching code:

```
    internal val deviceId by lazy {
        var deviceId = preferences.getString(DEVICE_ID_KEY, null)

        if (deviceId == null) {
            deviceId = deviceIdGenerator.invoke()
            preferences.edit().putString(DEVICE_ID_KEY, deviceId).apply()
        }

        deviceId
    }
```

I think there is very little that can go wrong here. Before this change, the `DEVICE_ID_KEY` preference did not exist. So the first thing that happens when a user installs this update is that we generate a new device id that then gets stored in the preferences.

[...]

PR #15589, opened 2020-10-01
For #15586: Migrate Fennec's device ID

grigoryk:
### Pull Request checklist

1. click on Show All Checks,


pocmo (commented):
Please see comment in #15392.

I think we have more situations where the ID we keep track of is just wrong. I think we should stop double caching and only use Leanplum as the source of truth for the ID and not save anything on our side at all.

grigoryk:
@pocmo so IIUC, we could:
- not call setDeviceId at all (remove all of this & existing id caching code)
- use Leanplum.getDeviceId() for uid verification

This will get the migrated builds to verify against what Leanplum things is the canonical deviceId (which should be the same as what we'd get from the Fennec's pref). And fresh builds will get a new device ID generated by Leanplum itself.

@pocmo do you recall why we cache deviceId at all on our side?

pocmo:
>  -  not call setDeviceId at all (remove all of this & existing id caching code)
>  -  use Leanplum.getDeviceId() for uid verification

Yeah, definitely using the getter - even if we continue to set it for some reason.

> @pocmo do you recall why we cache deviceId at all on our side?

I am not sure.

We did it in Fennec, maybe we can find something tracing this back?
https://searchfox.org/mozilla-esr68/source/mobile/android/base/java/org/mozilla/gecko/mma/MmaDelegate.java#110-115

Maybe we just didn't look as closely at the internals as Stefan and assumed we will always have to set it ourselves? Just guessing.. :)

pocmo:
Integration bug:
https://bugzilla.mozilla.org/show_bug.cgi?id=1351585

Looks like we didn't do it in the first iteration? 🤔 
https://hg.mozilla.org/mozilla-central/rev/a1e493554517

pocmo:
Here it is:
https://bugzilla.mozilla.org/show_bug.cgi?id=1400150

pocmo:
https://hg.mozilla.org/mozilla-central/rev/3045288dcb2e

pocmo:
So if I read that correctly back then we only did this to have a way to (temporarily) replace the ID with a test ID? Which is no longer something we do.

st3fan:
I'm going to close this because I think we are going to need a different approach. Will file a new issue soon.

Issue #15515, opened 2020-09-28
Validate JVM profiler marker calls are fast


mcomella:

mcomella:
If yes, we can move our StrictMode suppression marker to run everytime rather than just in debuggable builds.

mcomella:
## Data
Using this PR https://github.com/mozilla-mobile/fenix/pull/15613 on local Nightly builds, I get the following values:
```
| device | profiler active? | Results (ns) |
|-|-|-|
| P2 | no | 80979.2 |
| P2 | yes | 103958.25 |
| GS5 | no | 241708.3 |
| GS5 | yes | 226367.25 |
```

Notes:
- In no profiler active, I ran 10 runs and ran tests on cold startup: i.e. I looped `force-stop && sleep 3 && start`
- In profiler active, I took 4 runs and ran tests on cold startup

## Conclusion
It's 200k nanoseconds (on the GS5), which is 0.2ms. By comparison, logging (`android.util.Log.e("lol", "test logging")`) is ~83k nanoseconds (with high variance & removing the outliers) so if we log three times, we're slower than adding a profiler marker. So it seems _okay_ to sprinkle these about, though slightly more carefully than we do log statements. If we're concerned about performance critical areas (e.g. start-up), we could record timestamps and call these methods at a later point.

The marker calls aren't significantly slower when the profiler is running either.

## Raw results
These are in the same order as the table:
- 10 runs, average: 80979.2
[74010, 82917, 79167, 77656, 90989, 76927, 83125, 85157, 74375, 85469]
- 4 runs, average: 103958.25
[90208, 90365, 141562, 93698]
- 10 runs, average: 241708.3
[291771, 221041, 217083, 234844, 210417, 306042, 238385, 226302, 231719, 239479]
- 4 runs, average: 226367.25
[224896, 244479, 226667, 209427]

MarcLeclair:
@mcomella  I see the GS5 without profiler active being longer than with profiler active: 

```| GS5 | no | 241708.3 |```   vs  ```| GS5 | yes | 226367.25 |```

is that a formatting issue since you mentioned it was 200k slower with the profiler on?

mcomella:
@MarcLeclair No, it was slower with the profiler on. Did you force stop the process between your tests?

mcomella:
[...]

PR #15613, opened 2020-10-01
DO NOT LAND: measure duration of addMarker.

mcomella:
@MarcLeclair Please validate this measurement and the results https://github.com/mozilla-mobile/fenix/issues/15515#issuecomment-702433101 seem sound. **This PR is not intended to land.**

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
+++ b/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
@@ -10,6 +10,7 @@ import android.content.res.Configuration
 import android.os.Build
 import android.os.Bundle
 import android.os.StrictMode
+import android.os.SystemClock
 import android.text.format.DateUtils
 import android.util.AttributeSet
 import android.view.KeyEvent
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -292,6 +293,11 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
                 }
             }
         }
+
+        val startTime = SystemClock.elapsedRealtimeNanos()
+        components.core.engine.profiler?.addMarker("TestMarkerTime")
```
MarcLeclair, line 298: While this is probably a non-issue, I'm wondering, if we used the function that uses a time value if it would change the time measurement. I'm looking at this: https://searchfox.org/mozilla-central/source/mobile/android/geckoview/src/main/java/org/mozilla/gecko/GeckoJavaSampler.java#144

When there's a `startTime` provided but no `endTime` we'd end up calling into the profiler code and i wonder how much overhead that would add? or maybe that doesn't matter if we're not thinking of adding time readings to our markers.

mcomella: I'm not too concerned about the specific time in this instance but that's good to keep in mind for future calls (e.g. ones being used for CI start up measurements).

mcomella:
I confirmed on matrix that this review is approved and good to close.

PR #15619, opened 2020-10-01
15276 Moving HomeActivity.onCreate to StartupManager.activityOnCreate for performance code ownership.

MarcLeclair:
The performance team has had issues tracking regression for our start up path. It's quite hard to keep track of everything being added. Therefore, we thought of allowing us to be code owners for logic that happens during the application startup. However, some of that code resides in `HomeActivity.onCreate` and becoming code owners of the whole `HomeActivity` file is quite excessive and would tag us as reviewers automatically on certain changes that wouldn't affect start up.

There are some changes in this PR that are more than just "moving" the code from `HomeActivity` to `StartupManager` such as `BrowsingModeManager`. I put up this PR right now to get some early reviews in case anyone had concerns about our idea.

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
+++ b/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
@@ -30,7 +30,6 @@ import androidx.navigation.ui.AppBarConfiguration
 import androidx.navigation.ui.NavigationUI
 import kotlinx.android.synthetic.main.activity_home.*
 import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Dispatchers.IO
 import kotlinx.coroutines.Dispatchers.Main
 import kotlinx.coroutines.ExperimentalCoroutinesApi
@@ -48,7 +47,6 @@ import mozilla.components.feature.privatemode.notification.PrivateNotificationFe
 import mozilla.components.feature.search.BrowserStoreSearchAdapter
 import mozilla.components.service.fxa.sync.SyncReason
 import mozilla.components.support.base.feature.UserInteractionHandler
-import mozilla.components.support.ktx.android.arch.lifecycle.addObservers
 import mozilla.components.support.ktx.android.content.call
 import mozilla.components.support.ktx.android.content.email
 import mozilla.components.support.ktx.android.content.share
@@ -58,14 +56,10 @@ import mozilla.components.support.locale.LocaleAwareAppCompatActivity
 import mozilla.components.support.utils.SafeIntent
 import mozilla.components.support.utils.toSafeIntent
 import mozilla.components.support.webextensions.WebExtensionPopupFeature
-import org.mozilla.fenix.GleanMetrics.Metrics
 import org.mozilla.fenix.addons.AddonDetailsFragmentDirections
 import org.mozilla.fenix.addons.AddonPermissionsDetailsFragmentDirections
-import org.mozilla.fenix.browser.UriOpenedObserver
 import org.mozilla.fenix.browser.browsingmode.BrowsingMode
 import org.mozilla.fenix.browser.browsingmode.BrowsingModeManager
-import org.mozilla.fenix.browser.browsingmode.DefaultBrowsingModeManager
-import org.mozilla.fenix.components.metrics.BreadcrumbsRecorder
 import org.mozilla.fenix.components.metrics.Event
 import org.mozilla.fenix.exceptions.trackingprotection.TrackingProtectionExceptionsFragmentDirections
 import org.mozilla.fenix.ext.alreadyOnDestination
@@ -84,8 +78,7 @@ import org.mozilla.fenix.home.intent.StartSearchIntentProcessor
 import org.mozilla.fenix.library.bookmarks.BookmarkFragmentDirections
 import org.mozilla.fenix.library.history.HistoryFragmentDirections
 import org.mozilla.fenix.library.recentlyclosed.RecentlyClosedFragmentDirections
-import org.mozilla.fenix.perf.Performance
-import org.mozilla.fenix.perf.StartupTimeline
+import org.mozilla.fenix.perf.StartupManager
 import org.mozilla.fenix.search.SearchDialogFragmentDirections
 import org.mozilla.fenix.session.PrivateNotificationService
 import org.mozilla.fenix.settings.SettingsFragmentDirections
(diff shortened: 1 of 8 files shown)


st3fan:
This PR will need a descriptive title and a short description.

MarcLeclair:
@st3fan  sorry about that! Added a short description. Happy to expand on my thoughts in there for anyone 😄

app/src/main/java/org/mozilla/fenix/perf/StartupManager.kt
```diff
@@ -0,0 +1,280 @@
...
+            activity: HomeActivity,
+            savedInstanceState: Bundle?,
+            isVisuallyComplete: Boolean,
+            webExtensionPopupFeature: WebExtensionPopupFeature,
+            getBreadcrumbMessage: (NavDestination) -> String,
+            externalSourceIntentProcessors: KProperty0<List<HomeIntentProcessor>>,
+            action: () -> Unit
+        ): StartupData {
+            var themeManager: ThemeManager?
+            var browsingModeManager: BrowsingModeManager?
```
mcomella, line 62: nit: don't hoist variables: declare them close to where you need them. In that case, these can probably be `val` too

app/src/main/java/org/mozilla/fenix/perf/StartupManager.kt
```diff
@@ -0,0 +1,280 @@
...
+import kotlin.reflect.KProperty0
+
+class StartupManager {
+
+    @Suppress("TooManyFunctions")
+    companion object {
+        @OptIn(ExperimentalCoroutinesApi::class)
+        @Suppress("LongParameterList", "LongMethod")
+        fun activityOnCreate(
+            activity: HomeActivity,
```
mcomella, line 53: I wonder if we can get away with not passing `activity`? If we pass in a context or Activity, this method can basically do anything and it's hard to know what's _really_ happening when it's called. However, if we pass in a subset of components, we can make more well-reasoned guesses about what it's able to change. Maybe it's not worth it though – I think we'd have to get everyone on-board about separating code from Activities (and other android framework methods) and it's hard to do that from a perf team perspective.

app/src/main/java/org/mozilla/fenix/perf/StartupManager.kt
```diff
@@ -0,0 +1,280 @@
...
+            ).also {
+                it.start()
+            }
+
+            if (isActivityColdStarted(activity.intent, savedInstanceState)) {
+                externalSourceIntentProcessors.get().any {
+                    it.process(
+                        activity.intent,
+                        navHost.navController,
+                        activity.intent
```
mcomella, line 116: Why is intent passed twice? 🤔

mcomella (commented):
The general structure of the change looks fine to me:
- activity.onCreate is readable
- StartupManager.onCreate is readable (and the size of this file makes it really clear just how much we're doing on Activity startup)

I think some counter-arguments to making this change are:
- the file is tricky to change because of the added abstraction. My counter-points:
  - we want to do less in the critical startup path anyway so ideally this file wouldn't have much code added (and number of lines could almost be an indicator of our progress in improving startup)
  - the abstraction layer will help insulate us from the Activity and android framework, making it easier to understand what this method is doing (and write tests! e.g. we can actually mock the activity now)
- we're losing blame history. Can't really argue with that

Let's see what @csadilek thinks about the format of the change before I review further. In the meantime, I focused my comments correctness and things that would make this more readable.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -149,94 +140,23 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
...
-        }
-
-        if (isActivityColdStarted(intent, savedInstanceState)) {
-            externalSourceIntentProcessors.any {
-                it.process(
-                    intent,
-                    navHost.navController,
-                    this.intent
-                )
+        val startupData = StartupManager.activityOnCreate(this,
```
csadilek, line 144: TBH, I don't like `StartupManager` as a name here. It really doesn't manage startup. It's not responsible for coordinating or executing startup steps/tasks. Maybe just `Startup`, or `StartupStep`/`StartupTask`, as it's really a static utility containing various startup steps.

How is `StartupData` going to look if it contains multiple steps? I assume we wouldn't want to have multiple `nullable` fields in there. So, we will need to partition this somehow into `StartupData.HomeActivity`, `StartupData.Application`? 

I also think `activityOnCreate` sounds very generic given that it will only work for `HomeActivity`.

csadilek (commented):
I understand the intention of this, but this is partitioning / structuring code based on code ownership alone, which is an interesting case for Conway's law :)

Left a few first comments below. I think our goal here must be to have an API for startup coordination so that startup feels like a "feature." I am happy to help iterate on this and see if we can get it there, but I think we need to start with at least two tasks to see how the API will "feel."

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
MarcLeclair: I agree on the naming convention(s) used in here. I think the reason I chose `StartupManager` was because I was hoping to use this for more than just activity.onCreate and eventually allow it to be more than what this is. However, I think that's thinking too far ahead for this. Would you have any other name to suggest?  I didn't have much other candidate beside :  `HomeActivity` such as `HomeActivityStartupPath`

app/src/main/java/org/mozilla/fenix/perf/StartupManager.kt
MarcLeclair: I would like to avoid that too.  Multiple places in the code I moved uses `HomeActivity` as a variable such as `onActivityCreateEndHome`. I'm not sure how we would manage that one. Also, what do you mean by a "subset of components"?

app/src/main/java/org/mozilla/fenix/perf/StartupManager.kt
MarcLeclair: Good question. Looking at the method from `HomeIntentProcessor.process` it seems as if the first param is: "The intent to process" and the third param ( Or second intent) is " intent to mutate"

app/src/main/java/org/mozilla/fenix/perf/StartupManager.kt
mcomella: > Also, what do you mean by a "subset of components"?

For example, instead of `components`, pass `components.strictMode` and `components.intentProcessor` (that's not what we need in this case but this is just an example).

app/src/main/java/org/mozilla/fenix/perf/StartupManager.kt
mcomella: Weird. I wonder if we should have someone from android components look this over just to make sure it's correct because it seems strange to me.

mcomella (commented):
.

mcomella:
Closing as inactive

Issue #15596, opened 2020-10-01
[Bug] What's new in toolbar menu opens the wrong link


vesta0:
## Steps to reproduce
- Tap on What's new in toolbar menu

### Expected behavior
- Opens the migration FAQ

### Actual behavior
- It should open What's new in 82

### Device information

* Android device: ?
* Fenix version: ?

LaurentiuApahideanSV:
The issue also occurs on Firefox Nightly and has been reported here: https://github.com/mozilla-mobile/fenix/issues/15463.

AndiAJ:
Hi, I've just checked this matter on both the latest **`Nightly 201007`** and **`Beta 82.0.0-beta.4`** and it's still reproducible.

❌ [This link](https://support.mozilla.org/en-US/kb/whats-new-firefox-android-79) is displayed on both builds
✔️ As mentioned [here](https://github.com/mozilla-mobile/fenix/issues/15463#issue-709608110), [this link](https://support.mozilla.org/en-US/kb/whats-new-firefox-android) should be displayed. 

@boek - Please review. ☺️

liuche:
We asked SUMO to fix the redirect on their side - we will flag for retesting when that is ready.

AndiAJ:
Hi, **verified as fixed** on the latest **`Nightly 201009`** using the following devices:
• Google Pixel 3a (Android 10)
• Huawei Mate 20 Lite (Android 9)
• OnePlus A3 (Android 6.0.1)

The users are now re-directed to this [link](https://support.mozilla.org/en-US/kb/whats-new-firefox-android?as=u&utm_source=inproduct&redirectslug=whats-new-firefox-android-79&redirectlocale=en-US)

► Video
![20201009-091022](https://user-images.githubusercontent.com/51314259/95549412-e8b69880-0a0f-11eb-923d-6c8801301709.gif)

PR #15643, opened 2020-10-02
For #15596 - Removes upgrade faq link from fennec builds

boek:


pocmo (approved):

gabrielluong:
Passing this over to @liuche for uplift(?).

liuche:
It looks like the uplifts already happened, bc this commit is in 82. Maybe we never landed this for Nightly?

Issue #15644, opened 2020-10-02
[Bug] Custom Android lint rules are not running when run via command line


mcomella:
## Steps to reproduce
- Change this code:
```
+++ b/app/src/main/res/layout/tab_list_row.xml
@@ -48,7 +48,7 @@
             app:layout_constraintCircleRadius="28dp"
             app:layout_constraintEnd_toEndOf="@id/favicon_image"
             app:layout_constraintTop_toTopOf="@id/favicon_image"
-            app:srcCompat="@drawable/media_state_pause" />
+            android:src="@drawable/media_state_pause" />
```
- Run `./gradlew lintDebug`

### Expected behavior
Lint on the command line fails with an error, IDE shows the same error.
![image](https://user-images.githubusercontent.com/759372/94973846-9f22f680-04c1-11eb-8641-57b88f33db09.png)

### Actual behavior
Lint on the command line passes, while the IDE shows the error.

Lint via the command line doesn't appear to be picking up our custom lint checks at all.

### Device information

* Android device: ?
* Fenix version: master-ish a173749ad8ae65a705ed091ea6931d09890fa330

mcomella:
I believe this issue is this: https://issuetracker.google.com/issues/143455360 One of our dependencies declares a lint.jar via ServiceLoader dependencies, which causes our lint.jar to not run (or maybe only run sometimes). It looks like it's fixed in android gradle plugin 3.6, according to the post.

mcomella:
Posted a WIP in https://github.com/mozilla-mobile/fenix/pull/15646 but I can't fix all the errors myself.

mcomella:
Merged with https://github.com/mozilla-mobile/fenix/pull/15646

Issue #15641, opened 2020-10-02
[Bug] Possible accessibility issues found by Android Lint warnings; severity change could prevent future accessibility issues


mcomella:
## Steps to reproduce
- Run Android Lint: `./gradlew lintDebug`
- See "Accessibility" checks category on results page

### Expected behavior
No WARNINGS

### Actual behavior
Several WARNINGS:
> 1 | warning ClickableViewAccessibility: Accessibility in Custom Views
9 | warning ContentDescription: Image without contentDescription
4 | warning LabelFor: Missing accessibility label

In order, this could be indicative of:
- a view that cannot be accessed with accessibility tools turned on
- a view that will not speak when highlighted
- an editable field that doesn't speak clearly for accessibility issues

We should consider fixing these. Furthermore, if we change the severity of these checks (and other accessibility checks) to ERROR, **we can prevent these issues from occurring in the future before the PR merges.** Even if we don't fix these issues now, we could suppress them and change the severity so future issues are not introduced.

### Device information

* Android device: ?
* Fenix version: master-ish a173749ad8ae65a705ed091ea6931d09890fa330

Issue #20, opened 2018-04-18
FNX-5388 ⁃ We have a functioning URL bar with autocomplete based on your history


pocmo:

sblatz:
I believe this is a mockup of what this ticket is referencing

<img width="366" alt="screen shot 2019-01-28 at 11 39 22 am" src="https://user-images.githubusercontent.com/4400286/51861600-8415e180-22f1-11e9-8ea5-a53bc57b3623.png">

bbinto:
@sblatz - let me know if you want to use this as a breakdown issue of the History Meta as it's one of the acceptance criteria. If not, please close.

PR #15646, opened 2020-10-02
For #15644: update Android Gradle Plugin, restarting custom lint checks

mcomella:
However, I can't fix all these lint errors by myself! Posting as a WIP.

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/lint.xml
+++ b/app/lint.xml
@@ -7,8 +7,6 @@
         <!-- The Sentry SDK is compiled against parts of the Java SDK that are not available in the Android SDK.
              Let's just ignore issues in the Sentry code since that is a third-party dependency anyways. -->
         <ignore path="**/sentry*.jar" />
-        <!-- Temporary until https://github.com/Kotlin/kotlinx.coroutines/issues/2004 is resolved. -->
-        <ignore path="**/kotlinx-coroutines-core-*.jar"/>
     </issue>
 
     <!-- Lints that don't apply to our translation process -->
--- a/app/lint.xml
+++ b/app/lint.xml
@@ -10,13 +10,19 @@
         <!-- Temporary until https://github.com/Kotlin/kotlinx.coroutines/issues/2004 is resolved. -->
         <ignore path="**/kotlinx-coroutines-core-*.jar"/>
     </issue>
+
     <!-- Lints that don't apply to our translation process -->
     <issue id="MissingTranslation" severity="ignore" />
     <issue id="PluralsCandidate" severity="ignore" />
     <issue id="StringFormatCount" severity="ignore" />
     <issue id="TypographyEllipsis" severity="ignore" />
     <issue id="ExtraTranslation" severity="warning" />
+
     <!-- Lints that are disabled by default -->
     <issue id="ConvertToWebp" severity="warning" />
+
+    <!-- Performance: we haven't validated that addressing these checks have a significant impact
+         on performance but they're very quick to fix so we escalate them to error. -->
+    <issue id="InefficientWeight" severity="error" />
 </lint>
 
--- a/app/lint.xml
+++ b/app/lint.xml
@@ -21,6 +21,44 @@
 
     <!-- Performance: we haven't validated that addressing these checks have a significant impact
          on performance but they're very quick to fix so we escalate them to error. -->
+    <!-- Performance: big wins from a theoretical perspective so we escalate to error. -->
+    <issue id="DrawAllocation" severity="error" />
+    <issue id="Wakelock" severity="error" />
+    <issue id="WakelockTimeout" severity="error" />
+    <issue id="Recycle" severity="error" />
+    <issue id="StaticFieldLeak" severity="error" />
+    <issue id="ViewTag" severity="error" />
+    <issue id="ViewHolder" severity="error" />
+    <issue id="HandlerLeak" severity="error" />
+    <issue id="NestedWeights" severity="error" />
+
+    <!-- Performance: quick-to-fix violations so we escalate to error.
+         We haven't validated that they have a significant impact though. -->
+    <issue id="ObsoleteLayoutParam" severity="error" />
+    <issue id="ObsoleteSdkInt" severity="error" />
+    <issue id="AnimatorKeep" severity="error" />
+    <issue id="DuplicateDivider" severity="error" />
+    <issue id="MergeRootFrame" severity="error" />
+    <issue id="UseOfBundledGooglePlayServices" severity="error" />
+    <issue id="UseValueOf" severity="error" />
     <issue id="InefficientWeight" severity="error" />
+    <issue id="DisableBaselineAlignment" severity="error" />
+    <issue id="UselessLeaf" severity="error" />
+    <issue id="UselessParent" severity="error" />
+    <issue id="UnusedNamespace" severity="error" />
+
+    <!-- Performance: checks we'd like to eventually set to error. -->
+    <issue id="UseCompoundDrawables" severity="warning" />
+    <issue id="Overdraw" severity="warning" />
+    <issue id="UnusedResources" severity="warning" />
+
+    <!-- Performance: checks that we're unsure of the value of that we might want to investigate. -->
+    <issue id="UnpackedNativeCode" severity="informational" />
+    <issue id="LogConditional" severity="informational" />
+    <issue id="VectorPath" severity="informational" />
+    <issue id="UseSparseArrays" severity="informational" /> <!-- hurts developer convenience of kotlin Map... -->
+    <issue id="TooDeepLayout" severity="warning" /> <!-- depth can be customized -->
+    <issue id="TooManyViews" severity="warning" /> <!-- view count can be customized -->
+
 </lint>
 
--- a/app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt
@@ -758,7 +758,7 @@ class GleanMetricsService(
 
     internal fun setStartupMetrics() {
         setPreferenceMetrics()
-        Metrics.apply {
+        with(Metrics) {
             defaultBrowser.set(browsersCache.all(context).isDefaultBrowser)
             mozillaProductDetector.getMozillaBrowserDefault(context)?.also {
                 defaultMozBrowser.set(it)
@@ -812,7 +812,7 @@ class GleanMetricsService(
         // We purposefully make all of our preferences the string_list format to make data analysis
         // simpler. While it makes things like booleans a bit more complicated, it means all our
         // preferences can be analyzed with the same dashboard and compared.
-        Preferences.apply {
+        with(Preferences) {
             showSearchSuggestions.set(context.settings().shouldShowSearchSuggestions.toStringList())
             remoteDebugging.set(context.settings().isRemoteDebuggingEnabled.toStringList())
             telemetry.set(context.settings().isTelemetryEnabled.toStringList())
--- a/app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt
@@ -719,7 +719,11 @@ private val Event.wrapper: EventWrapper<*>?
         is Event.ChangedToDefaultBrowser -> null
     }
 
-class GleanMetricsService(private val context: Context) : MetricsService {
+class GleanMetricsService(
+    private val context: Context,
+    private val browsersCache: BrowsersCache = BrowsersCache,
+    private val mozillaProductDetector: MozillaProductDetector = MozillaProductDetector
+) : MetricsService {
     override val type = MetricServiceType.Data
 
     private val logger = Logger("GleanMetricsService")
@@ -755,11 +759,11 @@ class GleanMetricsService(private val context: Context) : MetricsService {
     internal fun setStartupMetrics() {
         setPreferenceMetrics()
         Metrics.apply {
-            defaultBrowser.set(BrowsersCache.all(context).isDefaultBrowser)
-            MozillaProductDetector.getMozillaBrowserDefault(context)?.also {
+            defaultBrowser.set(browsersCache.all(context).isDefaultBrowser)
+            mozillaProductDetector.getMozillaBrowserDefault(context)?.also {
                 defaultMozBrowser.set(it)
             }
-            mozillaProducts.set(MozillaProductDetector.getInstalledMozillaProducts(context))
+            mozillaProducts.set(mozillaProductDetector.getInstalledMozillaProducts(context))
 
             adjustCampaign.set(context.settings().adjustCampaignId)
             adjustAdGroup.set(context.settings().adjustAdGroup)
--- a/app/src/main/java/org/mozilla/fenix/components/toolbar/TabCounter.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/toolbar/TabCounter.kt
@@ -39,7 +39,7 @@ class TabCounter @JvmOverloads constructor(
         counter_root.contentDescription = if (count == 1) {
             context?.getString(R.string.open_tab_tray_single)
         } else {
-            context?.getString(R.string.open_tab_tray_plural, count.toString())
+            String.format(context.getString(R.string.open_tab_tray_plural), count.toString())
         }
     }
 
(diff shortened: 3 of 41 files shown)


ekager:
That should fix all of the immediate "errors" but we should definitely file issues to look into the warnings as well. I'm concerned about some of the a11y ones

ekager:
Oh nice you already did :) https://github.com/mozilla-mobile/fenix/issues/15641

mcomella:
Let's re-run CI.

mcomella:
A bunch of random test failures:
>android.view.InflateException: Binary XML file line #20: Binary XML file line #20: Error inflating class <unknown>
Caused by: android.view.InflateException: Binary XML file line #20: Error inflating class <unknown>
Caused by: java.lang.reflect.InvocationTargetException
...
Caused by: java.lang.UnsupportedOperationException: Failed to resolve attribute at index 5: TypedValue{t=0x2/d=0x7f0402eb a=-1}
	at android.content.res.TypedArray.getColorStateList(TypedArray.java:546)

ekager:
Yeah also mentioned here: https://github.com/mozilla-mobile/fenix/pull/15687#issuecomment-703679828

pocmo:
Now I am definitely in favor of landing yours first. :)

pocmo:
What I tried so far is bumping Robolectric (and mockk) since this often was needed for a new plugin or SDK version... but that didn't help.

mcomella:
I added some generally low risk configuration changes to lint.xml. Due to the low risk, I added them here instead of taking extra time to open a new pull request and coordinate the two landing.

Amejia481 (approved):

pocmo:
I experimented with different runners and configurations with not success so far.

After debugging this a bit I noticed that this is caused by values like `?accent` and `?contastText` in the theme. Somehow at test runtime those cannot be resolved.

pocmo:
Another observation: If I take an isolated test like `OnboardingFinishViewHolderTest` and I wrap `testContext` like:

```Kotlin
val context = ContextThemeWrapper(
    testContext,
    R.style.NormalTheme
)
```

.. and if I use this `context` instead of `testContext` then the test passes. So for some reason `testContext` doesn't seem to have the theme applied.

pocmo:
I did experiment with replacing `testContext` with a `fenixTestContext` that applies the theme and that brought it down to 26 failures (from ~84?). Commit: https://github.com/mozilla-mobile/fenix/pull/15687/commits/826ffca2e9996f47540a4091c21a7c991e38d5aa - Still not sure why/if this is needed.

pocmo:
@mcomella @ekager Is anyone of you still looking into getting this landed?

mcomella:
I found an issue on the Robolectric forums with the same exception as us: https://github.com/robolectric/robolectric/issues/5348#issuecomment-575890396

One commenter said:
> The problem was I had defined different themes for my app, which I didn't need to set for my Splash screen because it was always the same.
> 
>The solution for me was to add
>
>        val settings = applicationContext.getSharedPreferences(Constants.THEME, Context.MODE_PRIVATE)
>        setTheme(settings.getInt(Constants.THEME, R.style.AppTheme_Blue))
>
> to my onCreate and my tests ran as expected

(btw, I tried bumping us to Robolectric 4.4 but it didn't resolve the issue)

mcomella:
> I did experiment with replacing `testContext` with a `fenixTestContext` that applies the theme

I noticed that adding `testContext.setTheme(R.style.NormalTheme)` before the line that throws also seems to work too.

Applying that globally:
```kotlin
class FenixRobolectricTestApplication : FenixApplication() {

    override fun onCreate() {
        super.onCreate()
        setTheme(R.style.NormalTheme)
    }
```

I get one test failure over the whole suite in `GleanMetricsServiceTest. setStartupMetrics sets some base metrics`.

### Cause?
After my change, I noticed the test log output has messages like:

>     [Robolectric] org.mozilla.fenix.session.PrivateNotificationServiceTest.service opens home activity with PBM flag set to false: sdk=28; resources=BINARY

When I run before this PR, I see:

>     [Robolectric] org.mozilla.fenix.components.TrackingProtectionPolicyFactoryTest.GIVEN
 custom policy WHEN some tracking policies THEN tracking policies should match passed po
licies: sdk=28; resources=LEGACY

It looks like we got upgraded to running in BINARY resources mode (legacy has been deprecated for a long time). This comment https://github.com/robolectric/robolectric/issues/4961#issuecomment-488517645 from a robo maintainer mentions:

> Previously with legacy xml resource implementation, if you attempt to look
up an attribute that doesn't exist in your context's theme, Robolectric
would silently ignore it. However, with binary resources, just like real
Android, that results in a runtime error.

In the production app, I'm guessing we're not setting our theme completely at compile time (in XML) but setting it at runtime in the Activity. When Robolectric tries to inflate a view before the Activity is created, it cannot look up the appropriate attributes and crashes (I'm assuming Application is created before `setUp()` due to my fix).

 ### Questions
- Is this really the root cause?
- If so, should we do something differently in the production app? Given that we seem to set the theme in HomeActivity, I'm concerned about the possibility of these:
  - If a different activity starts the app, we won't set a theme and we will crash when certain attributes are accessed (are there crash reports for this?)
  - If we don't provide a theme, we may default to an android framework default theme. This could mean different app appearance between devices because it's not provided by the support library but by specific API levels
  - others...? Themes always seemed complicated to me
- If we make no changes to production, is this the right solution for robolectric? Seems okay? 🤷‍♂️ 
- nit: I guess we should set `android.enableUnitTestBinaryResources` to true?

I don't have the expertise to delve into these theme issues (and should really focus on perf stuff) – @ekager, are these remaining questions something you could look into?

mcomella:
Note: I didn't provide a commit for my test fixes because it might be misleading because it might not be the right solution and the diff (see comment above) is small enough to quickly make locally.

pocmo:
Looks like we have been going through the same debugging paths. We set `NormalTheme` on the `application` node in the manifest, so it should be available everywhere: https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/AndroidManifest.xml#L29 - It doesn't make sense to me that we would have to set it again in tests.

mcomella:
> Looks like we have been going through the same debugging paths. We set `NormalTheme` on the `application` node in the manifest, so it should be available everywhere:

I wonder if 1) our multiple manifest files configurations somehow overrides this value, 2) if having a `FenixRobolectricTestApplication` somehow overrides this value, 3) our activities somehow choose to ignore this style, 4) we're accessing the code before it's ready in robolectric (e.g. maybe it can only be used in `@Test` and not `@Before` ?), or if it's just a dumb bug, perhaps related to things before. :)

ekager:
Hi ya'll. So I looked into this today. 

This looks like a fairly common issue and the recommendation from the team is to [use one of these 3 options](https://github.com/robolectric/robolectric/issues/2641#issuecomment-435213957)

```
Right, that is the problem, your application context will not have your applications theme set on it as declared in it will have the platforms default theme.

I know, I know this is very counter intuitive, but that is how it works on a real device :-)

You can do three things.

Application app = ApplicationProvider.getApplicationContext();
app.setTheme(R.style.myTheme);

Context context = new ContextThemeWrapper(ApplicationProvider.getApplicationContext(), R.style.myTheme)

Activity activity = Robolectric.setupActivity(Activity.class);
```
Mike's [solution from this comment ](https://github.com/mozilla-mobile/fenix/pull/15646#issuecomment-705133690) seems to be using answer number 1. 

Applying [Sebastian's patch from this comment](https://github.com/mozilla-mobile/fenix/pull/15646#issuecomment-704257326) that was option number 2 from the maintainers, but then we hit another issue where we are getting StackOverflowErrors where this is used with spy. What we could do: [we could possibly try to use Mockito.spy instead of Mockk as suggested in this issue.](https://github.com/mockk/mockk/issues/41#issuecomment-547749813) Or we could use the original `testContext` for those cases where we use spy. (I tested this and it works)

With either of these solutions, all of the tests pass except `setStartupMetrics sets some base metrics`.

I think we should go with Mike's since it is less overall change?

pocmo:
> I think we should go with Mike's since it is less overall change?

Agree 👍

> all of the tests pass except `setStartupMetrics sets some base metrics.`

What is the failure you are seeing? Do you know if it's glean related? My PR contained a Glean version bump (via AC) that fixed an issue with version 4 of the plugin. Just wondering if that may be a problem with 3.6 already? Just guessing though. :)

ekager:
Failure is 
```
expected:<true> but was:<false>
Expected :true
Actual   :false
<Click to see difference>

java.lang.AssertionError: expected:<true> but was:<false>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:118)
	at org.junit.Assert.assertEquals(Assert.java:144)
	at org.mozilla.fenix.components.metrics.GleanMetricsServiceTest.setStartupMetrics sets some base metrics(GleanMetricsServiceTest.kt:40)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.rules.TestWatcher$1.evaluate(TestWatcher.java:55)
	at org.junit.rules.RunRules.evaluate(RunRules.java:20)
	at org.robolectric.RobolectricTestRunner$HelperTestRunner$1.evaluate(RobolectricTestRunner.java:546)
	at org.robolectric.internal.SandboxTestRunner$2.lambda$evaluate$0(SandboxTestRunner.java:252)
	at org.robolectric.internal.bytecode.Sandbox.lambda$runOnMainThread$0(Sandbox.java:89)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:834)
```

MarcLeclair (approved):

mcomella:
Still need to address that last unit test failure:
```
java.lang.AssertionError: expected:<true> but was:<false>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:118)
	at org.junit.Assert.assertEquals(Assert.java:144)
	at org.mozilla.fenix.components.metrics.GleanMetricsServiceTest.setStartupMetrics sets some base metrics(GleanMetricsServiceTest.kt:40)
```

mcomella:
> I know, I know this is very counter intuitive, but that is how it works on a real device :-)

@ekager Could this mean that we're currently using the platform default theme in the production application? Would we want to change that behavior rather than changing the test code?

mcomella:
The test failure occurs because the test package name is `org.mozilla.fenix.debug`, `MozillaProductDetector.getInstalledMozillaProducts` returns false when it doesn't recognize an application, and the debug app ID is not recognized. Looking for a fix...

mcomella:
I'm guessing the app ID behavior changed between versions.

mcomella:
Given that the robolectric app ID is set on the build variant and this test is currently coupled to a specific application ID, I think our solutions are:
- Most robust: pass in mocks of the dependencies to return the values we desire
- Least pain, if possible: override application ID in robolectric

mcomella:
I pushed a fix for the test failure and a few other small clean-ups. I added a comment for why we replace the context but I need the answer to my previous comment https://github.com/mozilla-mobile/fenix/pull/15646#issuecomment-708646969 to finish it – @ekager, let me know if you have thoughts.

ekager:
@mcomella I am good with just changing the test code (since this was all working before the upgrades) if you are!

mcomella:
Failure is a timeout:

```
[task 2020-10-15T05:01:33.998Z]   TEST: GIVEN sharedprefs contains installed engines WHEN installedSearchEngineIdentifiers THEN defaultEngines + customEngines ids are returned
[task 2020-10-15T05:01:33.998Z]     [Robolectric] org.mozilla.fenix.components.searchengine.FenixSearchEngineProviderTest.GIVEN sharedprefs contains installed engines WHEN installedSearchEngineIdentifiers THEN defaultEngines + customEngines ids are returned: sdk=28; resources=BINARY
[task 2020-10-15T05:01:36.898Z]     I/MonitoringInstr: Instrumentation started!
[task 2020-10-15T05:01:59.598Z]   SUCCESS
[task 2020-10-15T05:01:59.598Z] 
[task 2020-10-15T05:01:59.598Z] 
[task 2020-10-15T05:01:59.598Z] SUITE: org.mozilla.fenix.library.history.HistoryItemMenuTest
[task 2020-10-15T05:01:59.698Z]   TEST: builds menu items
[task 2020-10-15T05:02:00.298Z]     [Robolectric] org.mozilla.fenix.library.history.HistoryItemMenuTest.builds menu items: sdk=28; resources=BINARY
[task 2020-10-15T05:02:04.898Z]     I/MonitoringInstr: Instrumentation started!
[task 2020-10-15T05:02:08.298Z]   SUCCESS
[task 2020-10-15T05:02:08.298Z] 
[task 2020-10-15T05:02:08.298Z]   TEST: delete item has special styling
[task 2020-10-15T05:02:08.898Z]     [Robolectric] org.mozilla.fenix.library.history.HistoryItemMenuTest.delete item has special styling: sdk=28; resources=BINARY
[task 2020-10-15T05:02:17.998Z]     I/MonitoringInstr: Instrumentation started!
[task 2020-10-15T05:02:22.498Z]   SUCCESS
[task 2020-10-15T05:02:22.498Z] 
[task 2020-10-15T05:02:22.499Z] 
[task 2020-10-15T05:02:22.499Z] SUITE: org.mozilla.fenix.library.history.HistoryFragmentStoreTest
[task 2020-10-15T05:02:23.698Z]   TEST: startSync
[taskcluster:error] Task timeout after 7200 seconds. Force killing container.
```

TaskCluster was going strangely slow for this ^ look how long it took between tests. It could be:
- TC running slowly for some reason
- Spawning too many threads in roboletric: apparently they're not automatically cleaned up between tests:
> DO limit the number of the threads that are running during each test. Rogue threads often cause test pollution because they are not automatically cleaned up between tests. Oftentimes threads are inadvertently spawned when using third-party libraries (e.g for networking) or background processing components. One of the main sources of additional threads during tests are ExecutorServices that maintain thread pools. If possible, mock dependent components that spawn threads, or use a DirectExecutor. If it’s necessary to run multiple threads during a test, make sure to explicitly stop all threads and ExecutorServices to avoid test pollution.

http://robolectric.org/best-practices/

mcomella:
> @mcomella I am good with just changing the test code (since this was all working before the upgrades) if you are!

@ekager I was concerned that, in production, if we are using the platform default theme and not our theme, who provides that platform default theme? e.g. is it provided by the OEM? is it provided by the API level? In those two cases, the theme may be different between devices and cause inconsistent visuals (visuals changing on API level, OEM was a common problem a few years ago but less common today so perhaps it's not a problem). If it's by jetpack, it's should be fine.

I briefly looked into this though and I think we're using our theme as expected: `HomeActivity.onApplyThemeResource` is called before `onCreate` and seems to be called with the `NormalTheme` as `resources.getResourceEntryName(resid)` for the resid passed in returns NormalTheme. `applicationInfo.theme` returned the same resid so I think the Application is fine too.

I wonder why robolectric doesn't work if "that is how it works on a real device". I feel like it must be a robolectric bug...

🤷‍♂️ I guess I'm fine to ship it too.

mcomella:
I removed the TODO from the comment with an amended commit and force pushed.

mcomella:
`toolchain-linux64-android-gradle-dependencies` failed:
```
[task 2020-10-15T15:49:33.353Z] FAILURE: Build failed with an exception.
[task 2020-10-15T15:49:33.353Z] 
[task 2020-10-15T15:49:33.353Z] * What went wrong:
[task 2020-10-15T15:49:33.353Z] Could not determine the dependencies of task ':app:compileBetaJavaWithJavac'.
[task 2020-10-15T15:49:33.353Z] > Could not resolve all task dependencies for configuration ':app:betaCompileClasspath'.
[task 2020-10-15T15:49:33.353Z]    > Could not find com.leanplum:leanplum-core:5.4.0.
[task 2020-10-15T15:49:33.353Z]      Required by:
[task 2020-10-15T15:49:33.353Z]          project :app
[task 2020-10-15T15:49:33.353Z]          project :app > com.leanplum:leanplum-fcm:5.4.0
[task 2020-10-15T15:49:33.353Z]          project :app > com.leanplum:leanplum-push:5.4.0
```

I wonder if these were unpublished? 😱

mcomella:
I rebased on master locally and confirmed lintDebug still passes so I think we're good to merge! 🥳

Issue #12062, opened 2020-06-28
[UX] Re-evaluate user’s starting point on return visits


vesta0:
## User Story
As a user, I want to to have a choice in where I land when I re-launch the browser (e.g. Firefox homescreen, last open tab, custom URL, etc)

Product note: this is a recommendation from user research feedback. There is also performance considerations.

### Acceptance Criteria
* I can choose where my starting point is when re-launching the browser.

BranescuMihai:
One slight disadvantage that I see is that with the tab tray feature and now this one, I feel like the collections feature will become even less visible to the users

yoasif:
This is a desired feature (as you know) - see https://www.reddit.com/r/firefox/comments/i617id/any_way_to_make_firefox_on_android_open_on_my/

@BranescuMihai Collections has enough issues that I don't use them - we shouldn't keep user experience issues open just to promote a feature that isn't great yet.

data-sync-user:
➤ Betty Fleming commented:

🚩 Flag added 

null

yoasif:
In addition to UR, I am seeing a *lot* of feedback on this on social media and Play Store reviews. Even if the default is not changed (I would argue that it ought to be, because users expect state restoration as a platform convention), it really ought to be an option.

yoasif:
Couple of points:

* it feels to me that the current behavior is intended to make it easy for people to choose to start a new session or to "continue" an existing session using collections
* this seems to be the reason that the top sites, address bar and collections are immediately available 

This feels like the wrong solution - we are interrupting a person's workflow in order to present our value, instead of making it easy to choose:

* we don't currently have a home button to allow people to easily get back to the home screen where collections/top sites are present
* search and tab tray is present on every screen, so that choice is already present
* collections don't need to only be present on the home screen (or at all!) - see https://github.com/mozilla-mobile/fenix/issues/10372 for my prior suggestion here, so it is not necessary to force a home page to get to "core" functionality or to hold off on https://github.com/mozilla-mobile/fenix/issues/7551

[...]

PR #15647, opened 2020-10-02
For #12062 - Launch back to browser if we previously had tabs

ekager:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
+++ b/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
@@ -39,6 +39,7 @@ import kotlinx.coroutines.delay
 import kotlinx.coroutines.launch
 import mozilla.components.browser.search.SearchEngine
 import mozilla.components.browser.session.SessionManager
+import mozilla.components.browser.state.selector.getNormalOrPrivateTabs
 import mozilla.components.browser.state.state.SessionState
 import mozilla.components.browser.state.state.WebExtensionState
 import mozilla.components.concept.engine.EngineSession
@@ -126,7 +127,8 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
 
     private var isVisuallyComplete = false
 
-    private var privateNotificationObserver: PrivateNotificationFeature<PrivateNotificationService>? = null
+    private var privateNotificationObserver: PrivateNotificationFeature<PrivateNotificationService>? =
+        null
 
     private var isToolbarInflated = false
 
@@ -192,14 +194,18 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
             it.start()
         }
 
-        if (isActivityColdStarted(intent, savedInstanceState)) {
-            externalSourceIntentProcessors.any {
+        if (isActivityColdStarted(
+                intent,
+                savedInstanceState
+            ) && !externalSourceIntentProcessors.any {
                 it.process(
                     intent,
                     navHost.navController,
                     this.intent
                 )
             }
+        ) {
+            navigateToBrowserOnColdStart()
         }
 
         Performance.processIntentIfPerformanceTest(intent, this)
@@ -240,7 +246,10 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
         StartupTimeline.onActivityCreateEndHome(this) // DO NOT MOVE ANYTHING BELOW HERE.
     }
 
-    protected open fun startupTelemetryOnCreateCalled(safeIntent: SafeIntent, hasSavedInstanceState: Boolean) {
+    protected open fun startupTelemetryOnCreateCalled(
+        safeIntent: SafeIntent,
+        hasSavedInstanceState: Boolean
+    ) {
         components.appStartupTelemetry.onHomeActivityOnCreate(
             safeIntent,
             hasSavedInstanceState,
@@ -322,6 +331,10 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
     }
 
     final override fun onPause() {
+        // We should return to the browser if there were normal tabs when we left the app
+        settings().shouldReturnToBrowser =
+            components.core.store.state.getNormalOrPrivateTabs(private = false).isNotEmpty()
+
         if (settings().lastKnownMode.isPrivate) {
             window.addFlags(WindowManager.LayoutParams.FLAG_SECURE)
         }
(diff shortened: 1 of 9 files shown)


app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -307,6 +312,10 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
     }
 
     final override fun onPause() {
+        // We should return to the browser if there were normal tabs when we left the app
+        settings().shouldReturnToBrowser =
+            components.core.store.state.getNormalOrPrivateTabs(private = false).isNotEmpty()
```
csadilek, line 317: There's one case we're not handling here: When the user exits the app from home, we would still return them to the a tab (if one exists), but we should probably return to home in this case, e.g. User has a few tabs, opens home via "+" in the tabs tray, exits app. What do you think? Should we cover that?

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -744,6 +753,14 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
         }
     }
 
+    open fun navigateToBrowserOnColdStart() {
```
csadilek, line 756: Let's cover this one with a unit test. Gave it a try to see how painful it is. I think we might want to refactor `HomeActivityTest` later, but for now something like this will do: 

```kotlin
@Test
    fun `navigateToBrowserOnColdStart`() {
        clearMocks(activity)
        
        val browsingModeManager: BrowsingModeManager = mockk()
        every { browsingModeManager.mode } returns BrowsingMode.Normal

        val settings: Settings = mockk()
        every { settings.shouldReturnToBrowser } returns true
        every { activity.components.settings.shouldReturnToBrowser } returns true
        every { activity.openToBrowser(any(), any()) } returns Unit

        activity.browsingModeManager = browsingModeManager
        activity.navigateToBrowserOnColdStart()

        verify(exactly = 1) { activity.openToBrowser(BrowserDirection.FromGlobal, null) }
    }
```

(We can also over the case where `browserModeManager.mode.isPrivate` is `true`)

app/src/main/java/org/mozilla/fenix/customtabs/ExternalAppBrowserActivity.kt
```diff
@@ -39,6 +39,10 @@ open class ExternalAppBrowserActivity : HomeActivity() {
         components.appStartupTelemetry.onExternalAppBrowserOnCreate(safeIntent, hasSavedInstanceState)
     }
 
+    override fun navigateToBrowserOnColdStart() {
+        // No-op for external app
```
csadilek, line 43: 👍 good catch. Let's also cover this one in `ExternalAppBorwserActivityTest`. We can simple take our test from above and make sure we never call `openToBrowser`, just in case someone removes this empty override.

csadilek (commented):
Works great! Tested all cases manually, and didn't find anything. I have some questions/nits below, and we can add some quick unit tests too :)

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
ekager: I think this could be up to UX but I personally think it makes sense to treat "home" as the ephemeral new tab state that I wouldn't want to be returned to. Like if you went to settings, you wouldn't expect to be returned to settings on a cold start.

I think a larger discussion could happen around this interaction, and I also have brought up to UX that we could make this a Setting - ie launch to home/launch to browser with tabs.

ekager:
Tests added and nits addressed!

app/src/test/java/org/mozilla/fenix/customtabs/ExternalAppBrowserActivityTest.kt
```diff
@@ -37,6 +42,25 @@ class ExternalAppBrowserActivityTest {
         assertEquals(Event.OpenedApp.Source.CUSTOM_TAB, activity.getIntentSource(otherIntent))
     }
 
+    @Test
+    fun `navigateToBrowserOnColdStart does nothing for external app browser activity`() {
+        val activity = spyk(ExternalAppBrowserActivity())
+        clearMocks(activity)
```
csadilek, line 48: nit: We don't need the `clearMocks` call here, see comment above.

app/src/test/java/org/mozilla/fenix/HomeActivityTest.kt
```diff
@@ -87,6 +93,42 @@ class HomeActivityTest {
         assertFalse(activity.isActivityColdStarted(startingIntent, null))
     }
 
+    @Test
+    fun `navigateToBrowserOnColdStart in normal mode navigates to browser`() {
+        clearMocks(activity)
```
csadilek, line 98: nit: The `clearMocks` call is only needed because of `every { activity.applicationContext } returns testContext` in `@Before`. 

Let's move that one line from `@Before` to the one test that needs it: `getModeFromIntentOrLastKnown returns mode from settings when intent does not set`, then we can remove the `clearMocks` calls in those other tests.

app/src/test/java/org/mozilla/fenix/HomeActivityTest.kt
```diff
@@ -87,6 +93,42 @@ class HomeActivityTest {
...
+
+        activity.browsingModeManager = browsingModeManager
+        activity.navigateToBrowserOnColdStart()
+
+        verify(exactly = 1) { activity.openToBrowser(BrowserDirection.FromGlobal, null) }
+    }
+
+    @Test
+    fun `navigateToBrowserOnColdStart in private mode does not navigate to browser`() {
+        clearMocks(activity)
```
csadilek, line 116: nit: see above :)

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
csadilek: > I personally think it makes sense to treat "home" as the ephemeral new tab state that I wouldn't want to be returned to.

Interesting, yes. I think we can land and refine later. The reason I brought it up is because it's not necessarily clear which tab to restore, for example:

- open a few normal tabs and select one
- open a private tab
- restart the app
- now user is in private mode with no open tab (Good)
- exit private mode
- now user is on home screen (Good)
- restart the app
- now user is on newest normal tab as no tab was selected anymore (this seems fine, but launching back to home seemed better to me?)

csadilek (approved):
✨ 🚀 Did another round of testing, just two nits (in tests) and a comment.

Issue #15497, opened 2020-09-28
[Bug] Three dot menu for desktop bookmarks is hidden but still functions


s-ankur:
## Steps to reproduce
This is with sync turned off
go to bookmarks/desktop bookmarks
click on where the 3 dot menu should be 
### Expected behavior
nothing?
or maybe you want the 3 dot menu to be visble

### Actual behavior
its not visible but it works!

### Device information

* Android device: Asus Zenfone
* Fenix version: Nightly

s-ankur:
I can trigger crashes by renaming the folders and also this happens with sync on as well

s-ankur:
![Screenshot_2020-09-29-00-38-19-929_org mozilla fenix](https://user-images.githubusercontent.com/33252312/94474996-17e93080-01ec-11eb-89f0-480529b9b807.jpg)

ekager:
Let's make sure we're using `removeAndDisable()` for the menu button here https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/java/org/mozilla/fenix/library/bookmarks/viewholders/BookmarkNodeViewHolder.kt#L67

mrwhoknows55:
@s-ankur   I would like to work on this issue

ekager:
Anyone can work on this!

mrwhoknows55:
I'll submit PR after hacktoberfest kicks off!

ekager:
We can't keep issues reserved, so it's just the first PR that we get in that will be reviewed.

Anutrix:
It says "Can't edit default folders" for me.
![Screenshot_20201001_123839_org mozilla firefox](https://user-images.githubusercontent.com/11459028/94778953-bbe8fc80-03e3-11eb-90ee-96b211e56f7f.jpg)
Is this expected?

Mugurell:
@sijanr also raised an important detail: the bookmarks folders' titles can occupy the entire row while the individual bookmarks are always clipped by the size of the menu button, even when that is not showing.
For this I filed #15776

AndiAJ:
Hi, **verified as fixed** on the latest **`Nightly Build 201012`** using the following devices:
• Google Pixel 3a (Android 10)
• Huawei Mate 20 Lite (Android 9)
• OnePlus A3 (Android 6.0.1)

PR #15649, opened 2020-10-03
For #15497: Hide and remove the three dot menu from the Desktop Bookmarks folder

sijanr:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/library/bookmarks/viewholders/BookmarkNodeViewHolder.kt
+++ b/app/src/main/java/org/mozilla/fenix/library/bookmarks/viewholders/BookmarkNodeViewHolder.kt
@@ -64,7 +64,7 @@ class BookmarkNodeViewHolder(
 
         // Hide menu button if this item is a root folder or is selected
         if (item.type == BookmarkNodeType.FOLDER && item.inRoots()) {
-            containerView.overflowView.visibility = View.GONE
+            containerView.overflowView.hideAndDisable()
         } else if (payload.modeChanged) {
             if (mode is BookmarkFragmentState.Mode.Selecting) {
                 containerView.overflowView.hideAndDisable()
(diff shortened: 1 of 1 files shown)


Mugurell (changes requested):
Thanks for looking into this!

This indeed fixes the issue (because of the `enabled = false` part) but using `hideAndDisable` would still have the button in the layout, just invisible. 

Should we use `removeAndDisable` so that the unactionable button won't be part of the layout?

sijanr (commented):
This should remove the button as well, right?

Mugurell:
> This should remove the button as well, right?

I think you had some specific code in mind that you wanted to point at...
I'm seeing that indeed, below the change needed here we use `hideAndDisable` which would still have that button occupy space.
And then if we are to use `removeAndDisable` for the "desktop bookmarks" node we would get a different behavior.
I'm not sure having the button still occupying space was intended or not. In the original ticket that implemented this I don't see exact specifications for this. 
I'll ask in the ticket for UX feedback. Thank you for pointing this out!

Mugurell:
Seeing that previously we were setting `View.GONE` for folders and `View.INVISIBLE` for other nodes and Emily also recommending using `removeAndDisable` here let's merge this and file another ticket to discuss with UX if this is the expected behavior or not.

Mugurell:
bors try

app/src/main/java/org/mozilla/fenix/library/bookmarks/viewholders/BookmarkNodeViewHolder.kt
```diff
@@ -12,10 +12,7 @@ import mozilla.components.concept.storage.BookmarkNode
 import mozilla.components.concept.storage.BookmarkNodeType
 import mozilla.components.support.ktx.android.content.getDrawableWithTint
 import org.mozilla.fenix.R
-import org.mozilla.fenix.ext.components
-import org.mozilla.fenix.ext.hideAndDisable
-import org.mozilla.fenix.ext.loadIntoView
-import org.mozilla.fenix.ext.showAndEnable
+import org.mozilla.fenix.ext.*
```
Mugurell, line 15: Wildcard imports will fail our detekt/ktlint.
Please use individual imports.

Mugurell:
bors try

Mugurell (approved):
LGTM
Thank you for pushing through with this!

Mugurell:
bors r=Mugurell

Issue #15402, opened 2020-09-24
[Bug] ETP pop-up can appear when toolbar is scrolled off-screen


mcomella:
## Steps to reproduce
- Clear data
- Click URL bar
- Load a site with trackers (e.g. kotaku.com)
- Scroll down

### Expected behavior
ETP pop-up points to ETP shield in toolbar.

### Actual behavior
Toolbar is offscreen so ETP pop-up points at nothing. We seem to wait for the page to completely load before showing the pop-up so I'd guess this happens to users pretty often who are probably scrolling before the page finishes loading.

![image](https://user-images.githubusercontent.com/759372/94182299-a577f980-fe55-11ea-83bc-2b344b394564.png)

### Device information

* Android device: P2
* Fenix version: Nightly 200924

s-ankur:
Duplicate of #14012, probably that can be closed

mcomella:
Thanks – I closed the other issue.

ebalazs-sv:
Verified as fixed on Nightly 201015 (Build #2015769771) GV 83.0a1 from 10/15 with Pixel 2 (Android 9).
I will remove the qa:needed label and close this issue.

PR #15667, opened 2020-10-04
For #15402: Hide ETP pop-up if the toolbar is not visible

sijanr:
ETP

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/trackingprotection/TrackingProtectionOverlay.kt
+++ b/app/src/main/java/org/mozilla/fenix/trackingprotection/TrackingProtectionOverlay.kt
@@ -63,7 +63,7 @@ class TrackingProtectionOverlay(
                 }
             }
             ToolbarPosition.TOP -> {
-                val appBarLayout = getToolbar().parent as AppBarLayout?
+                val appBarLayout = getToolbar().parent as? AppBarLayout
                 appBarLayout?.let { appBar ->
                     if(appBar.y != 0.toFloat()) {
                         return
--- a/app/src/main/java/org/mozilla/fenix/trackingprotection/TrackingProtectionOverlay.kt
+++ b/app/src/main/java/org/mozilla/fenix/trackingprotection/TrackingProtectionOverlay.kt
@@ -49,7 +49,7 @@ class TrackingProtectionOverlay(
 
     @Suppress("MagicNumber", "InflateParams")
     private fun showTrackingProtectionOnboarding() {
-        if (!getToolbar().hasWindowFocus()) return
+        if (getToolbar().translationY > 0) return
 
         val trackingOnboardingDialog = object : Dialog(context) {
             override fun onTouchEvent(event: MotionEvent): Boolean {
--- a/app/src/main/java/org/mozilla/fenix/trackingprotection/TrackingProtectionOverlay.kt
+++ b/app/src/main/java/org/mozilla/fenix/trackingprotection/TrackingProtectionOverlay.kt
@@ -54,13 +54,15 @@ class TrackingProtectionOverlay(
 
         if (!getToolbar().hasWindowFocus()) return
 
-        when (getToolbar().parent) {
-            is CoordinatorLayout -> {
+        val toolbarPosition = settings.toolbarPosition
+
+        when (toolbarPosition) {
+            ToolbarPosition.BOTTOM -> {
                 if(getToolbar().translationY > 0) {
                     return
                 }
             }
-            is AppBarLayout -> {
+            ToolbarPosition.TOP -> {
                 val appBarLayout = getToolbar().parent as AppBarLayout?
                 appBarLayout?.let { appBar ->
                     if(appBar.y != 0.toFloat()) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/fenix/trackingprotection/TrackingProtectionOverlay.kt
```diff
@@ -49,7 +49,7 @@ class TrackingProtectionOverlay(
 
     @Suppress("MagicNumber", "InflateParams")
     private fun showTrackingProtectionOnboarding() {
-        if (!getToolbar().hasWindowFocus()) return
+        if (getToolbar().translationY > 0) return
```
Mugurell, line 52: Should we also keep the `getToolbar().hasWindowFocus()` check?

Mugurell (changes requested):
Thank you for looking into this!

For a complete fix we need the same issue also resolved for when the toolbar is set at top (to know about this I think you can use a [AppBarLayout.OnOffsetChangedListener](https://developer.android.com/reference/com/google/android/material/appbar/AppBarLayout.OnOffsetChangedListener))

Changes after which [TrackingProtectionOverlayTest](https://github.com/mozilla-mobile/fenix/blob/24983af94e8a916113b81c86ad3a84510ed0ddbb/app/src/test/java/org/mozilla/fenix/trackingprotection/TrackingProtectionOverlayTest.kt) will probably need to be updated.

app/src/main/java/org/mozilla/fenix/trackingprotection/TrackingProtectionOverlay.kt
sijanr: I didn't notice any change with or without the check, but I'll include it in the next commit.

sijanr (commented):
The issue with adding the listener would be there's already one added for the app bar in BrowserToolbarView.kt.

https://github.com/mozilla-mobile/fenix/blob/ad2b99f9f99f7458fd353ae72fbf68b4e23b9f12/app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarView.kt#L107

So, I checked the parent of the toolbar instead.

app/src/main/java/org/mozilla/fenix/trackingprotection/TrackingProtectionOverlay.kt
```diff
@@ -49,8 +51,25 @@ class TrackingProtectionOverlay(
 
     @Suppress("MagicNumber", "InflateParams")
     private fun showTrackingProtectionOnboarding() {
+
         if (!getToolbar().hasWindowFocus()) return
 
+        when (getToolbar().parent) {
```
Mugurell, line 57: One slight improvement regarding knowing if the toolbar is at the top or at the bottom since the current check is pretty frail: What we usually use and something that would also increase readability is 
> when (settings.shouldUseBottomToolbar && settings.isDynamicToolbarEnabled)

Mugurell:
Thank you!
It looks and functions nicely!
Just one improvement to be made regarding knowing if we have a top or bottom toolbar.
Some small tests just for the return would be nice but I see the `showTrackingProtectionOnboarding` does quite a lot so fully testing this could be done in a later ticket.

sijanr (commented):
Yeah I agree with you that the code couldn't be easily read. I found a property to check the toolbar's position in Settings class and included it in the check because it appeared more readable. Thanks for your help.

app/src/main/java/org/mozilla/fenix/trackingprotection/TrackingProtectionOverlay.kt
```diff
@@ -49,8 +51,27 @@ class TrackingProtectionOverlay(
...
+        val toolbarPosition = settings.toolbarPosition
+
+        when (toolbarPosition) {
+            ToolbarPosition.BOTTOM -> {
+                if(getToolbar().translationY > 0) {
+                    return
+                }
+            }
+            ToolbarPosition.TOP -> {
+                val appBarLayout = getToolbar().parent as AppBarLayout?
```
Mugurell, line 66: One small thing but an important one:
`getToolbar().parent as AppBarLayout?` could still throw a `ClassCastException` if the parent is to be something different than `AppBarLayout?`.

Should we use safe cast here in the form
`getToolbar().parent as? AppBarLayout`
?

Mugurell:
> Yeah I agree with you that the code couldn't be easily read. I found a property to check the toolbar's position in Settings class and included it in the check because it appeared more readable. Thanks for your help.

Thank you!
It looks indeed better!

One small thing related to a safe cast to be resolved and then I'll merge this.

app/src/main/java/org/mozilla/fenix/trackingprotection/TrackingProtectionOverlay.kt
sijanr: Thanks for pointing it out. I'll push another commit.

Mugurell:
bors try

Mugurell:
@sijanr 
Looks like ktlint fails at
> TrackingProtectionOverlay.kt:16:1: Unused import
TrackingProtectionOverlay.kt:61:19: Missing spacing after "if"
TrackingProtectionOverlay.kt:68:23: Missing spacing after "if"

Can you please check and update the patch?

sijanr (commented):
I think this should fix the ktlint errors

Mugurell:
bors try

Mugurell (approved):
LGTM
Thanks for pushing through with this!

Mugurell:
bors r=Mugurell

Mugurell:
bors r=Mugurell

Issue #15508, opened 2020-09-28
[Bug] Edit Bookmark not showing error state when URL is empty


joc-a:
## Steps to reproduce

1. Have a bookmark saved
2. Edit bookmark -> clear name and URL -> click save -> you will see that nothing changes in the name and URL
3. Edit bookmark -> clear the name -> click save -> you will see that the name now becomes the same as the URL (I think this behavior is fine, but please correct me if I'm wrong)
4. Edit bookmark -> clear the url -> click save

The above steps are displayed in the video attachment.

### Expected behavior
After steps 2 and 4, we should not be able to save an empty URL in a bookmark. The Save icon on the top right should be disabled when the URL is in the error state.

### Actual behavior
We can save an empty URL in a bookmark. The Save icon remains enabled and after pressing it, the password field briefly shows a red error icon in its error state before leaving the Edit Bookmark screen.

### Device information

* Android device: Samsung A50, Android 10
* Fenix version: 81.1.1

![Edit Bookmark Bug](https://user-images.githubusercontent.com/38375996/94475624-05fd9280-01d8-11eb-9a8d-8f21686bfb00.gif)

I'd be happy work on this when it gets UX feedback 🙂

s-ankur:
I think it is a duplicate of https://github.com/mozilla-mobile/fenix/issues/14450, not sure

joc-a:
@s-ankur oh it's not crashing with me, it's just not showing the error state, but these two issues are related

ekager:
Looking at the code it looks like we currently do allow the title to be empty, but not the URL. We should definitely be showing an error on empty URL at least and looks like we have an error string: `bookmark_invalid_url_error`. Let's figure out why that isn't working and make sure you can't submit a blank or invalid URL!

ebalazs-sv:
Verified as fixed on Beta 84.0.0-beta.1 from 11/18 and Nightly 201119 from 11/19 with Samsung Galaxy Tab S3 (Android 8).
When trying to save an empty or invalid bookmark "Invalid URL" message is displayed.

I will close this issue.

PR #15674, opened 2020-10-04
For #15508: Show error when trying to save empty or invalid bookmark URL

joc-a:
### Pull Request checklist

1. click on Show All Checks,
4. the APK links should be on the left side of the screen, named for each CPU  #architecture

To be consistent with the UI in the Saved Logins screen, I wrapped the ClearableEditText of the bookmark URL in a TextInputLayout. But to be able to clear the error in the input layout when the text is changed, I needed to add a TextChangedListener in EditBookmarkFragment. The problem with that is that when I override onTextChanged in the listener, the code in onTextChanged in ClearableEditText will not execute. To work around this problem, I moved the logic in onTextChanged in ClearableEditText to a separate function that I can call in onTextChanged in the listener, and then proceed to clear the error in the TextInputLayout after. Please let me know if there's a better way to approach this issue and if I overlooked anything.

Edit Bookmark URL

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/library/bookmarks/edit/EditBookmarkFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/library/bookmarks/edit/EditBookmarkFragment.kt
@@ -5,7 +5,10 @@
 package org.mozilla.fenix.library.bookmarks.edit
 
 import android.content.DialogInterface
+import android.content.res.ColorStateList
 import android.os.Bundle
+import android.text.Editable
+import android.text.TextWatcher
 import android.view.Menu
 import android.view.MenuInflater
 import android.view.MenuItem
@@ -13,6 +16,7 @@ import android.view.View
 import androidx.appcompat.app.AlertDialog
 import androidx.appcompat.app.AppCompatActivity
 import androidx.appcompat.widget.Toolbar
+import androidx.core.content.ContextCompat
 import androidx.fragment.app.Fragment
 import androidx.fragment.app.activityViewModels
 import androidx.lifecycle.ViewModelProvider
@@ -134,6 +138,23 @@ class EditBookmarkFragment : Fragment(R.layout.fragment_edit_bookmark) {
                 placeCursorAtEnd()
                 showKeyboard()
             }
+
+            view.bookmarkUrlEdit.addTextChangedListener(object : TextWatcher {
+                override fun beforeTextChanged(s: CharSequence?, start: Int, count: Int, after: Int) {
+                    // NOOP
+                }
+
+                override fun onTextChanged(s: CharSequence?, start: Int, before: Int, count: Int) {
+                    bookmarkUrlEdit.onTextChanged(s)
+
+                    inputLayoutBookmarkUrl.error = null
+                    inputLayoutBookmarkUrl.errorIconDrawable = null
+                }
+
+                override fun afterTextChanged(s: Editable?) {
+                    // NOOP
+                }
+            })
         }
     }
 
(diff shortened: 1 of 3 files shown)


gabrielluong:
@joc-a Thank you for the contribution! Throwing this over to get some UX feedback.

apbitner:
@gabrielluong do we have a standardized way of handling input errors like this? I think this looks fine, but ideally we should be handling all such errors the same.

For example, this is the error state for a similar situation when adding a custom search engine:
image

joc-a:
> @gabrielluong do we have a standardized way of handling input errors like this? I think this looks fine, but ideally we should be handling all such errors the same.
> 
> For example, this is the error state for a similar situation when adding a custom search engine:
> image

@apbitner I followed the error state for editing Saved Logins.
Screenshot_1602783332

apbitner (approved):

gabrielluong:
> @gabrielluong do we have a standardized way of handling input errors like this? I think this looks fine, but ideally we should be handling all such errors the same.

I have no idea about any input error handling.

apbitner:
> > @gabrielluong do we have a standardized way of handling input errors like this? I think this looks fine, but ideally we should be handling all such errors the same.
> 
> I have no idea about any input error handling.

@joc-a used the pattern from editing logins so that's good. Just wanted to make sure we weren't starting from scratch when we have error handling elsewhere in the app.

liuche:
@joc-a when pushing PRs, can you also run the pre-push hook? it looks like this is failing our linter.
Instructions for setting it up are [here](https://github.com/mozilla-mobile/fenix/#pre-push-hooks) (or you can also look through [the file itself](https://github.com/mozilla-mobile/fenix/blob/master/config/pre-push-recommended.sh) and see what's in it!)

joc-a:
> @joc-a when pushing PRs, can you also run the pre-push hook? it looks like this is failing our linter.
> Instructions for setting it up are [here](https://github.com/mozilla-mobile/fenix/#pre-push-hooks) (or you can also look through [the file itself](https://github.com/mozilla-mobile/fenix/blob/master/config/pre-push-recommended.sh) and see what's in it!)

@liuche Done, thanks!

app/src/main/java/org/mozilla/fenix/library/bookmarks/edit/EditBookmarkFragment.kt
```diff
@@ -245,13 +266,28 @@ class EditBookmarkFragment : Fragment(R.layout.fragment_edit_bookmark) {
...
+                    inputLayoutBookmarkUrl.setErrorIconDrawable(R.drawable.mozac_ic_warning_with_bottom_padding)
+                    inputLayoutBookmarkUrl.setErrorIconTintList(
+                        ColorStateList.valueOf(
+                            ContextCompat.getColor(requireContext(), R.color.design_error)
+                        )
+                    )
                 }
             }
         }
-        progress_bar_bookmark.visibility = View.INVISIBLE
```
gabrielluong, line 254: I think we should consider keeping this here since we call it in both 270 and 279 now.

app/src/main/java/org/mozilla/fenix/library/bookmarks/edit/EditBookmarkFragment.kt
```diff
@@ -245,13 +266,28 @@ class EditBookmarkFragment : Fragment(R.layout.fragment_edit_bookmark) {
...
+
+                    findNavController().popBackStack()
+                }
             } catch (e: UrlParseFailed) {
                 withContext(Main) {
-                    bookmarkUrlEdit.error = getString(R.string.bookmark_invalid_url_error)
+                    progress_bar_bookmark.visibility = View.INVISIBLE
+
+                    inputLayoutBookmarkUrl.error = getString(R.string.bookmark_invalid_url_error)
+                    inputLayoutBookmarkUrl.error
```
gabrielluong, line 282: nit: I think this was left here unintentionally?

gabrielluong (changes requested):

gabrielluong (approved):

PR #15700, opened 2020-10-05
For #15596 - Removes upgrade faq link from fennec builds

boek:


st3fan (approved):
Approving this change as it is the same as was just reviewed by @pocmo - this is just for the uplift to v82.

PR #15727, opened 2020-10-06
Update Android Components version

MickeyMoz:

Issue #14593, opened 2020-09-01
[Bug] Intermittent test failure audioPlaybackSystemNotificationTest 


Archaeopteryx:
Log: https://console.firebase.google.com/project/moz-fenix/testlab/histories/bh.66b7091e15d53d45/matrices/5904482774751161600/executions/bs.b49c68b675e0d8e2/testcases/2/test-cases

java.lang.AssertionError
at org.junit.Assert.fail(Assert.java:86)
at org.junit.Assert.assertTrue(Assert.java:41)
at org.junit.Assert.assertTrue(Assert.java:52)
at org.mozilla.fenix.ui.robots.BrowserRobot.waitForPlaybackToStart(BrowserRobot.kt:344)
at org.mozilla.fenix.ui.MediaNotificationTest$audioPlaybackSystemNotificationTest$2.invoke(MediaNotificationTest.kt:96)
at org.mozilla.fenix.ui.MediaNotificationTest$audioPlaybackSystemNotificationTest$2.invoke(MediaNotificationTest.kt:29)
at org.mozilla.fenix.ui.robots.NavigationToolbarRobot$Transition.enterURLAndEnterToBrowser(NavigationToolbarRobot.kt:107)
at org.mozilla.fenix.ui.MediaNotificationTest.audioPlaybackSystemNotificationTest(MediaNotificationTest.kt:93)
at java.lang.reflect.Method.invoke(Native Method)
at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
at androidx.test.internal.runner.junit4.statement.RunBefores.evaluate(RunBefores.java:80)
at androidx.test.internal.runner.junit4.statement.RunAfters.evaluate(RunAfters.java:61)
at androidx.test.rule.ActivityTestRule$ActivityStatement.evaluate(ActivityTestRule.java:531)
at org.junit.rules.RunRules.evaluate(RunRules.java:20)
at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
[...]

PR #15734, opened 2020-10-06
Upgrade to Android Component 62.0.20201006190820 with breaking changes

jonalmeida:
AC upgrade that fixes the breaking changes with `ImageLoader` and `ImageLoadRequest`.

1. click on Show All Checks,

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -336,6 +336,7 @@ dependencies {
     implementation Deps.leanplum_core
     implementation Deps.leanplum_fcm
 
+    implementation Deps.mozilla_concept_base
     implementation Deps.mozilla_concept_engine
     implementation Deps.mozilla_concept_menu
     implementation Deps.mozilla_concept_push
--- a/app/src/androidTest/java/org/mozilla/fenix/ui/MediaNotificationTest.kt
+++ b/app/src/androidTest/java/org/mozilla/fenix/ui/MediaNotificationTest.kt
@@ -8,6 +8,7 @@ import androidx.test.uiautomator.UiSelector
 import okhttp3.mockwebserver.MockWebServer
 import org.junit.After
 import org.junit.Before
+import org.junit.Ignore
 import org.junit.Rule
 import org.junit.Test
 import org.mozilla.fenix.helpers.AndroidAssetDispatcher
@@ -53,6 +54,7 @@ class MediaNotificationTest {
     }
 
     @Test
+    @Ignore("https://github.com/mozilla-mobile/fenix/issues/15754")
     fun videoPlaybackSystemNotificationTest() {
         val videoTestPage = TestAssetHelper.getVideoPageAsset(mockWebServer)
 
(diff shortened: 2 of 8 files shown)


pocmo (approved):

jonalmeida:
This is now failing on the same media tests as https://github.com/mozilla-mobile/fenix/pull/15727.

jonalmeida:
I'm beginning to suspect the tests.

When debugging `audioPlaybackSystemNotificationTest` (which fails consistently), adding a breakpoint at `waitForPlaybackToStart` returns false the few times when it looks for the text "Media file is playing". However, I can see that text on the screen.

Evaluating that line a few times, and then the value returns true all the time. I don't see any direct changes to the tests so far or the helper classes.

I'm going to re-test this with an older GV version to see if they start to pass.

liuche:
@jonalmeida there's also an intermittent for that media test: https://github.com/mozilla-mobile/fenix/issues/14593

If it's too flaky, then you can disable the test too.

st3fan:
@jonalmeida +1 to disabling the tests and filing an issue to get them going - this will unblock people sooner.

jonalmeida:
There are a few tests that are failing, so I ended up testing with `tabMediaControlButtonTest` from the following list:
- `audioPlaybackSystemNotificationTest` - apparent intermittent; consistent failing now
- `mediaSystemNotificationInPrivateModeTest` - hard to reproduce; need older API level device that I do not have.
- `tabMediaControlButtonTest` - should be consistent.

I tried a bunch of GV builds to see what the media failure is but I've come across some confusing results. Using the latest AC (`3e68528074`) with these GV Nightly versions tested:
- `83.0.20201005093214` - ❌ (monday's GV after media regression was fixed)
- `83.0.20201001094020` - ❌ (GV media breakage)
- `83.0.20200930092918` - ❌ (before GV media breakage; expected to fail)
- `83.0.20200929094055` - ❌ (second before media breakage; expected to fail)

GV media regression was fixed (feature was disabled) in 20201001094020 so the `83.0.20201005093214` _should_ have worked.
  - https://hg.mozilla.org/mozilla-central/rev/e4e35e1aae77

---

I've disabled UI test failures for now so we can investigate without blocking anyone and I've filed the follow issue: https://github.com/mozilla-mobile/fenix/issues/15754

liuche:
Thanks for following through with these - @jonalmeida i also gave the testing team a heads up in https://chat.mozilla.org/#/room/#mobile-test-eng:mozilla.org which is where @rpappalax and team look into intermittents.

Do you have any insight into whether these tests are failing as intended (e.g. catching an actual problem) or that we need to go through and redo how we're using the testing framework?

jonalmeida:
> Thanks for following through with these - @jonalmeida i also gave the testing team a heads up in https://chat.mozilla.org/#/room/#mobile-test-eng:mozilla.org which is where @rpappalax and team look into intermittents.
> 
> Do you have any insight into whether these tests are failing as intended (e.g. catching an actual problem) or that we need to go through and redo how we're using the testing framework?

The tests were originally catching legitimate regressions in GeckoView, but it seems those same tests are still failing with newer GeckoView versions. For example, I've verified that the media tests are able to press play on the media elements and the verification text on the screen _should_ have been correctly asserted, but those tests are still failing.

Issue #14735, opened 2020-09-04
[Bug] Don't dump URLs in the logs for play store builds


agi:
We shouldn't log full URLs for play store builds, e.g.

```
09-03 23:34:25.979 27227 27227 D GeckoSession: handleMessage GeckoView:LocationChange uri=https://www.instagram.com/<redacted>
```

This can be achieved easily by removing `Log.d` and `Log.i` entirely from the build, FxR does this: https://searchfox.org/mozilla-mobile/rev/d7c5ad84a23415853b1da7e36be37bff9694aa71/FirefoxReality/app/proguard-rules.pro#23-31

mcarare:
@agi Should we also remove other verbosity logging levels? (`v`, `w`, `e `?)

agi:
Yes for `v`  (verbose). We should keep `w` (warnings) and `e` (errors) as they're often very helpful when dealing with user reports.

AndiAJ:
Hi, **verified as fixed** on both **`Nightly 201014`** and **`82.0.0-beta.5`** using a OnePlus A3 (Android 6.0.1)

PR #15747, opened 2020-10-07
For #14735: Remove debug logs from release build.

mcarare:
### Pull Request checklist

1. click on Show All Checks,


mcomella (commented):
This change seems to be in response to agi's concerns about printing urls to the console. I can think of two problems with this:
1. If user share logs, they will share their URLs
1. In older versions of android, other apps can read logs and steel URLs

Well I like the idea of being able to solve this problem fairly comprehensively, removing all debug logs seems like a big change – I'm concerned that this would impact our ability to help users debug or for them to file bugs. @mcarare **Have you consulted the broader Fenix team that this is a desirable change to make?**

On fennec, we solve the problem of outputting URLs by frequently checking the logs and ensuring that urls are not printed. Theoretically, we could also write an automated test to look through the logs for us. I wonder if we should consider one of these alternative approaches?

In any case, I don't think I am the best person to review this change so I'll try to pass it on to the fenix team.

app/proguard-rules.pro
```diff
@@ -45,6 +45,18 @@
...
+####################################################################################################
+# Remove logs from
+####################################################################################################
+-assumenosideeffects class android.util.Log {
+    public static boolean isLoggable(java.lang.String, int);
+    public static int v(...);
+    public static int i(...);
+    public static int w(...);
+    public static int d(...);
+    public static int e(...);
```
mcomella, line 57: agi mentions in the issue that we shouldn't remove error or warning logs.

mcomella:
I also realize that if we don't get rid of warning and error logs, we may still be logging urls.

@agi , do you have any thoughts about my comments above?

I also realized that our old performance test system relies on log statements like these in order to take measurements. If we decide to implement this, we should identify a solution for that system before landing this. Unfortunately, our replacement test system is not ready yet. We run our performance tests on the `nightly-simulation` build so perhaps we can remove these new proguard statements just from that build.

liuche:
cc @st3fan too - I hadn't thought much about removing info-logging on Release, but maybe you have some perspectives?

app/proguard-rules.pro
mcarare: This is just a draft, put up before any answers to that thread.

csadilek: +1 on landing this. Looks like we also need to capture `android.util.Log.println` which is used by A-C's LogSink:
https://github.com/mozilla-mobile/fenix/blob/d767cd199ede1364d5f69dc7e7ddc7d4fbb9b24a/app/src/main/java/org/mozilla/fenix/FenixApplication.kt#L117

https://github.com/mozilla-mobile/android-components/blob/master/components/support/base/src/main/java/mozilla/components/support/base/log/sink/AndroidLogSink.kt#L31

pocmo:
I share some of @mcomella's concerns and feel like this could benefit from a discussion (if this would be in AC I'd recommend an RFC) more than landing a patch quickly.

st3fan:
I think it could be interesting to have a discussion or RFC about logging in general. For example we could decide that the info logging is actually useful but that we want to anonymize log entries by removing things that look like URLs.

csadilek:
@pocmo @st3fan I think removing debug logs from release builds is a good first step (which I thought is the main motivation here), and can be independent of a wider discussion in an RFC about all log levels and additional safe guards.

st3fan:
@csadilek i agree. lets land this and I think we should also uplift it to beta so that it will ship in v82. Low risk right?

agi:
From a GeckoView perspective, debug logs are and were never intended to run on production builds, we add `.d` logs liberally all over the codebase and they could even have impact performance. I'm not sure however if you can remove GV log statements and not your own debug logs.

As it stands right now, we leak every single page load to the logs which is pretty bad for privacy.

st3fan:
@Mergifyio  backport releases/v82.0.0

st3fan:
@Mergifyio  backport releases/v82.0.0

app/proguard-rules.pro
csadilek: OK, based on the discussion, we do want to strip just debug logs as a first step. Just need to find way to exclude A-C's debug logs as well, which is trickier as A-C is using `println` directly.

We could hook up a custom LogSink in Fenix release builds and filter out debug logs there?

st3fan:
@mcarare We think this is important enough to ship in 82, so if you can wrap this up on no later than Monday that would be great. Can you also open an uplilft PR to land this on the `releases/v82.0.0` branch and cc me on that one?

mcomella:
Stripping the debug logs will break our performance tests. However, we can land the PR mostly as is to get it in to the upcoming release and fast follow to fix the tests because we only run on Nightly and can miss a day or two.

wrt whether or not we should do this, I'm still concerned it'll hurt our ability to debug but otherwise don't have a strong opinion and my role doesn't have a strong stake in this so I'm OK to land.

### Fixing perf tests
We rely on these debug level logs for our performance test harness:
- `D/GeckoSession: handleMessage GeckoView:PageStart uri=...`
- `D/GeckoSession: handleMessage GeckoView:PageStop uri=null`

Can we find a solution that keeps these logs for our tests? The simplest change I can think of is to make a new, temporary (as this perf test system is being replaced) build variant that has a different proguard configuration and gets built when nightly does. I think that means:
- Creating a new TC task type ([like this](https://github.com/mozilla-mobile/fenix/blob/5153a3f2c5011a1f77c303832f3cd22692a68207/taskcluster/ci/build/kind.yml#L98))
- Add a new gradle build type in TC ([like this](https://github.com/mozilla-mobile/fenix/blob/5153a3f2c5011a1f77c303832f3cd22692a68207/taskcluster/ci/build/kind.yml#L104) - maybe this String is used verbatim into gradle)?
- Add a new build variant in gradle patterned after nightly ([like this](https://github.com/mozilla-mobile/fenix/blob/5153a3f2c5011a1f77c303832f3cd22692a68207/app/build.gradle#L65))
- Change the proguard configuration in the nightly build to not remove these lines (i.e. maybe append a new file [to this](https://github.com/mozilla-mobile/fenix/blob/5153a3f2c5011a1f77c303832f3cd22692a68207/app/build.gradle#L44) that undoes the log removal)

Can anyone think of a simpler solution?

CC @bdekoz that we'll have to change FNPRMS a bit to compensate.

agi:
> Can anyone think of a simpler solution?

@mcomella You could add similar logs on the AC side in response to `onPageStart`, `onPageStop` from GV. Without the URI obviously.

mcomella:
> > Can anyone think of a simpler solution?
> 
> @mcomella You could add similar logs on the AC side in response to `onPageStart`, `onPageStop` from GV. Without the URI obviously.

@agi Do we receive those events in ac at the same time they're logged in GV? Given we have no interest in maintaining this perftest system long term, I would like to keep the changes comparable to the current test.

agi:
@mcomella yes it's a synchronous call: https://searchfox.org/mozilla-central/rev/919607a3610222099fbfb0113c98b77888ebcbfb/mobile/android/geckoview/src/main/java/org/mozilla/geckoview/GeckoSession.java#619-625

csadilek (approved):
Thanks @mcarare. I think this is good to ship as an initial solution and we can refine later. However, let's add a test for the `LogSink` before landing, e.g:

```kotlin
class FenixLogSinkTest {

    @Before
    fun setup() {
        mockkStatic(Log::class)
    }

    @After
    fun teardown() {
        unmockkStatic(Log::class)
    }

    @Test
    fun `GIVEN we're in a release build WHEN we log debug statements THEN logs should not be forwarded`() {
        val logSink = FenixLogSink(false)
        logSink.log(mozilla.components.support.base.log.Log.Priority.DEBUG, "test", message = "test")
        verify(exactly = 0) { Log.println(Log.DEBUG, "test", "test") }
    }

    @Test
    fun `GIVEN we're in a debug build WHEN we log debug statements THEN logs should be forwarded`() {
        val logSink = FenixLogSink(true)
        logSink.log(mozilla.components.support.base.log.Log.Priority.DEBUG, "test", message = "test")
        verify(exactly = 1) { Log.println(Log.DEBUG, "test", "test") }
    }
}
```

csadilek:
Opened a PR in A-C to introduce replacement logs for performance tests / FNPRMS: https://github.com/mozilla-mobile/android-components/pull/8663

mcomella:
> Opened a PR in A-C to introduce replacement logs for performance tests / FNPRMS: [mozilla-mobile/android-components#8663](https://github.com/mozilla-mobile/android-components/pull/8663)

I've confirmed that a variation on this PR will be sufficient for the performance tests so I'm good for this to land whenever.

mcomella:
FYI @bdekoz with the variation on the PR I'm suggesting, we won't need any changes to FNPRMS

PR #15769, opened 2020-10-07
No issue: remove inline from resetAfter

mcomella:
I believe this was a premature optimization: it's unclear that it would
actually improve performance and yet  we had to add several weird
workarounds to make it work that broke encapsulation. It's also
possible it would be worse for the APK size because of excessive
inlining.

The difficulty in mocking StrictMode.resetAfter is concerning.
I'm starting to second-guess whether or not making strict mode manager a class
was a good idea.

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/StrictModeManager.kt
+++ b/app/src/main/java/org/mozilla/fenix/StrictModeManager.kt
@@ -23,6 +23,9 @@ import org.mozilla.fenix.perf.Performance
 private const val MANUFACTURE_HUAWEI: String = "HUAWEI"
 private const val MANUFACTURE_ONE_PLUS: String = "OnePlus"
 
+private val logger = Performance.logger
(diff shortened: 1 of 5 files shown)


csadilek (approved):
👍

app/src/main/java/org/mozilla/fenix/StrictModeManager.kt
```diff
@@ -114,7 +113,7 @@ class StrictModeManager(
      *
      * @return the value returned by [functionBlock].
      */
-    inline fun <R> resetAfter(policy: StrictMode.ThreadPolicy, functionBlock: () -> R): R {
+    fun <R> resetAfter(policy: StrictMode.ThreadPolicy, functionBlock: () -> R): R {
```
csadilek, line 116: FWIW, I agree that inlining this isn't worth breaking encapsulation.

Issue #14923, opened 2020-09-10
[UX] Users don't know that they can adjust their browser font size


vesta0:
## User Story
- As a user, I want to adjust my browser font size based on my needs so I can use the browser. 

## User problem:
Based on user feedback, many users don't know that they are able to adjust the browser font. This may be due to the fact that the Font Size scale feature is not visible when "Automatic font sizing" is toggled on. Perhaps we can still show a grayed out version of the font size option when "Automatic font sizing" is toggled on.

### Acceptance Criteria
- I know how to adjust my font size.

black7375:
### Simple Solution
A very easy solution exists.

1. Remove option `Automatic font size`
2. Replace `Font Size` with `Font Size Control`
3. When the option is off, disable the slider bar. (darkly)
![Font-Size](https://user-images.githubusercontent.com/25581533/94321373-e28bdb00-ff7e-11ea-8e52-79b25ec8d26a.jpg)

### New menu
Creating a page setup menu.
Similar to Safari's approach.

Pressing the current `Tracking Protection` menu position makes `Text Size`, `Tracking Protection`, `Site Setting`, ..ETC appear.
- Left Menu:  Specify site settings
- Right Menu: Browser-wide Features
![Safari-FontSize](https://user-images.githubusercontent.com/25581533/94321712-ee2bd180-ff7f-11ea-836a-5510c458d866.png)
[Refer Image](https://mcmw.abilitynet.org.uk/making-text-larger-safari-ios-13-iphone-ipad-and-ipod-touch)

You can both apply it. :smile:

apbitner:
@ekager To @black7375's point above, I'm wondering do we really need the `Automatic font sizing` toggle? Could we just have the slider and it would be implied that 100% on the slider is the equivalent of automatic font sizing? Chromium browsers just have the scale, no toggle.

ekager:
Yep @apbitner that would work - so do you want me to move forward with those changes?

apbitner:
@ekager Let's confirm first.

@vesta0 it doesn't seem the `Automatic font sizing` toggle is completely necessary so we propose removing it and just having the slider, which would make this option more discoverable. Sound good?

vesta0:
Let's do it!

apbitner:
@ekager I think we're good to move forward. Let me know if you have any design questions but it should just be removing the toggle and moving up the content beneath.

AndiAJ:
[...]

PR #15775, opened 2020-10-08
For #14923 - Removes auto size preference toggle

ekager:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/androidTest/java/org/mozilla/fenix/ui/robots/SettingsSubMenuAccessibilityRobot.kt
+++ b/app/src/androidTest/java/org/mozilla/fenix/ui/robots/SettingsSubMenuAccessibilityRobot.kt
@@ -21,7 +21,6 @@ import android.widget.TextView
 import androidx.test.espresso.UiController
 import androidx.test.espresso.ViewAction
 import androidx.test.espresso.ViewAssertion
-import androidx.test.espresso.assertion.ViewAssertions.doesNotExist
 import org.hamcrest.CoreMatchers.allOf
 import androidx.test.espresso.matcher.ViewMatchers.withText
 import androidx.test.espresso.matcher.ViewMatchers.withId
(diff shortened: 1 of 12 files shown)


app/src/androidTest/java/org/mozilla/fenix/ui/robots/SettingsSubMenuAccessibilityRobot.kt
```diff
@@ -49,13 +48,7 @@ class SettingsSubMenuAccessibilityRobot {
...
     }
 
-    fun verifyAutomaticFontSizingMenuItems() = assertAutomaticFontSizingMenuItems()
-
-    fun clickFontSizingSwitch() = toggleFontSizingSwitch()
-
-    fun verifyNewMenuItems() = assertNewMenuItems()
-
-    fun verifyNewMenuItemsAreGone() = assertNewMenuItemsAreGone()
+    fun verifyMenuItems() = assertNewMenuItems()
```
gabrielluong, line 51: [optional][nit] Should we rename assertNewMenuItems() to assertMenuItems() as well?

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -310,11 +310,6 @@ class Settings(private val appContext: Context) : PreferencesHolder {
         default = false
     )
 
-    var shouldUseAutoSize by booleanPreference(
```
gabrielluong, line 313: I am wondering if we can turn this into a getter (might not have to be in Settings or use the same name) and return the boolean value of `fontSizeFactor == 1F`. The way I think about this is that `fontSizeFactor == 1F` could be abstracted into a getter and with the right naming could help others reason what the setting means when looking at the other classes and not having the context of font size resizing.

gabrielluong (approved):
Changes look good. I would consider abstracting away `fontSizeFactor == 1F` into a getter to help provide a bit more context.

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
ekager: Good call out! Changed to reflect that and make it more usable

Issue #15265, opened 2020-09-21
[Bug]"metrics.has_top_sites": remains set to false


AndiAJ:
## Prerequisites
Have a clean profile or a fresh install.

## Steps to reproduce
1. Generate the **`metrics`** ping having only the default top sites displayed on the home screen
2. Navigate to a coupe of pages and pin the as top sites
3. Navigate to a couple of websites displaying them as frequently visited top sites
4. Generate the **`metrics`** ping

### Expected behavior
Step 1 & 4 "metrics.has_top_sites": true

### Actual behavior
Step 1 & 4  "metrics.has_top_sites": false

### Device information

* Android device: all
* Fenix version: Nightly Build 200921

### Notes
From the attached Logcat:
Step 1 `metrics` Ping **`5e4582d1-f96c-4897-a65b-22574b4a648f`**
Step 4 `metrics` Ping **`3228e355-b7bf-4d9a-b784-03571462ec0d`**

[Logcat](https://github.com/mozilla-mobile/fenix/files/5255460/MTS.txt)
[Glean dashboard](https://debug-ping-preview.firebaseapp.com/pings/aj-t21091)

ekager:
originally added these in https://github.com/mozilla-mobile/fenix/issues/8125 but looks like when we changed the way we show top sites, we never are updating `Settings.topSitesSize` anymore

AndiAJ:
Hi, **verified as fixed** on the latest **`Nightly 201015`** build using a Google Pixel 3a (Android 11)

✔️ Have **only the default top sites** - Ping 5ad0accb-a370-4067-8a28-7f31a9afee15
`"metrics.has_top_sites": true`
✔️ **Delete all the default top sites** - Ping 327c8838-b664-4880-9184-45ff6f0f598e
`"metrics.has_top_sites": false`
✔️ Have **only pinned top sites** - Ping e799c348-0298-4b61-abc5-e27ca1501315
`"metrics.has_top_sites": true`
✔️ **Delete all the pinned top sites** - Ping 79ad935e-0cd4-4f47-9f63-536b2b69f075
`"metrics.has_top_sites": false`
✔️ Have **only frequently visited top sites** - Ping 832a22ef-c832-4d9e-bf27-8d4f679308a1
`"metrics.has_top_sites": true`
✔️ **Delete all the frequently visited top sites** - Ping 07677185-c9ea-475d-9d56-e37b6c9e7da8
`"metrics.has_top_sites": false`

[...]

PR #15861, opened 2020-10-13
For #15265 - Save cached top sites for metrics ping

ekager:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
+++ b/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
@@ -326,6 +326,10 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
             window.addFlags(WindowManager.LayoutParams.FLAG_SECURE)
         }
 
+        // We will remove this when AC code lands to emit a fact on getTopSites in DefaultTopSitesStorage
+        // https://github.com/mozilla-mobile/android-components/issues/8679
+        settings().topSitesSize = components.core.topSitesStorage.cachedTopSites.size
+
         super.onPause()
 
         // Diagnostic breadcrumb for "Display already aquired" crash:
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -326,6 +326,8 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
             window.addFlags(WindowManager.LayoutParams.FLAG_SECURE)
         }
 
+        settings().topSitesSize = components.core.topSitesStorage.cachedTopSites.size
```
ekager, line 329: Since the `TopSiteFeature` handles fetching the list of top sites and adding them to `cachedTopSites` we don't have a callback AFAIK and I wasn't sure how to handle this otherwise without making changes in AC. Any ideas?

gabrielluong: This reminds me of how we emitted facts to tell how many web extensions were installed for the metric pings. See https://github.com/mozilla-mobile/android-components/pull/5998. However, there was a callback for initializing the web extensions compare to the TopSiteFeature. 

I think one way of doing this is to emit a fact with the size when we call `getTopSites` in [DefaultTopSitesStorage](https://github.com/mozilla-mobile/android-components/blob/be232773113848eb2ebd270b17271448423aee46/components/feature/top-sites/src/main/java/mozilla/components/feature/top/sites/DefaultTopSitesStorage.kt#L89), which we can follow up with setting the `topSitesSize` in Fenix. 

I would recommend making the change in AC because I think it is following a better practice of isolating the logic to their component for reusability despite the extra work required, and it also helps potential consumers of AC beyond Fenix (which I think would be really good for mobile).

ekager:
Added a comment about the AC bug so let's land this now so the metrics aren't totally broken in the mean time :)

PR #15895, opened 2020-10-14
Use new SearchEngineProvider interface.

pocmo:
This is for a change that will come with the next AC update. So for now that will fail to build.


gabrielluong:
Assigning @csadilek for this review since you did the sample-browser review too.

gabrielluong:
Throwing a pr:needs-changes since we got some try errors.

ekager (approved):

pocmo:
@ekager Thank you for fixing and landing this.

PR #15940, opened 2020-10-15
Closes #11286: Add TelemetryMiddleware to remove Session[Manager] observers

### LINKED ISSUE #11286: HomeActivity: Refactor UriOpenedObserver to use browser store
`UriOpenedObserver` is currently a `SessionManager` observer.

We want to get rid of SessionManager observers so we can finish the A-C refactoring to browser state.

csadilek:
Next step in migrating from A-C's `browser-session` (`SessionManager`) to `browser-state` (`BrowserStore`). This removes a `SessionManager` and a `Session` observer that can now be implemented in a cleaner way using a middleware on the browser store. We can simply "intercept" the needed actions and apply our side effects (recording telemetry in this case). With this change `HomeActivity` no longer has any references to `SessionManager`.

This does not change any behaviour and I did some manual testing to make sure the same events are being recorded. The only thing I removed is a temporary fix that is no longer needed (see deleted `TelemetrySessionObserver`).


app/src/main/java/org/mozilla/fenix/TelemetryMiddleware.kt
```diff
@@ -0,0 +1,106 @@
...
+        fun add(url: String) {
+            chain.add(url)
+        }
+    }
+
+    @Suppress("TooGenericExceptionCaught")
+    override fun invoke(
+        context: MiddlewareContext<BrowserState, BrowserAction>,
+        next: (BrowserAction) -> Unit,
+        action: BrowserAction
```
eliserichards, line 55: So our `action` will always be a `ContentAction`, and `next(action)` will always give us a `TabListAction`? Can we use the more specific types?

app/src/main/java/org/mozilla/fenix/components/Core.kt
```diff
@@ -251,11 +257,15 @@ class Core(
...
-        AdsTelemetry(context.components.analytics.metrics)
+        AdsTelemetry(metrics)
     }
 
     val searchTelemetry by lazy {
-        InContentTelemetry(context.components.analytics.metrics)
+        InContentTelemetry(metrics)
+    }
+
+    val metrics by lazy {
```
eliserichards, line 267: Nit: move this above `adsTelemetry` and `searchTelemetry`

buildSrc/src/main/java/AndroidComponents.kt
```diff
@@ -3,5 +3,5 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 object AndroidComponents {
-    const val VERSION = "63.0.20201014190200"
+    const val VERSION = "63.0.20201015143124"
```
eliserichards, line 6: Can we leave this update out of the PR and wait for the normal AC update?

eliserichards (commented):
I have a few questions, but it's looking great!

buildSrc/src/main/java/AndroidComponents.kt
csadilek: @eliserichards yes, this is just here so the PR compiles, as I need a new action introduced in this version. The A-C update will land before this PR anyway.

app/src/main/java/org/mozilla/fenix/TelemetryMiddleware.kt
csadilek: Ah no the action is always a `BrowserAction` here. The type of action is specified by the store. For example the type signature of the `BrowserStore` (which this middleware is for) is `Store<BrowserState, BrowerAction>`, basically defining that this store manages `BrowserState` which gets updated by `BrowserAction`s. 

`ContentAction`, `TabListAction` etc. are all subtypes of `BrowserAction` so in order to intercept actions on the store we need the top-level `BrowserAction`.

See: https://github.com/mozilla-mobile/android-components/blob/master/components/browser/state/src/main/java/mozilla/components/browser/state/store/BrowserStore.kt#L25

This is the same pattern for all stores in Fenix, say the `LoginsFragmentStore` which is a `Store<LoginsListState, LoginsAction>`. If we wanted to intercept actions there we'd use a `Middleware<LoginsListState, LoginsAction>`.

See: https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/java/org/mozilla/fenix/settings/logins/LoginsFragmentStore.kt#L45

app/src/main/java/org/mozilla/fenix/components/Core.kt
csadilek: OK, moved :).

app/src/main/java/org/mozilla/fenix/TelemetryMiddleware.kt
csadilek: > and next(action) will always give us a TabListAction?

Using a middleware it's possible to intercept actions before and after these actions are processed by store (and the corresponding state is updated). `next(action)` is what propagates actions to the next middleware in line and then ultimately to the store. 

So the actions we process before the call to `next(action)` haven't yet reached the store, which is useful if we want to read the state or apply side effects before the state has changed. So, this is useful for preprocessing i.e. "before this action is applied do this." 

The actions we process after the call to `next(action)` have already reached the store and the state has already been changed. So, this is useful for postprocessing i.e. "after this action has been applied do this."

csadilek: We modelled this after https://redux.js.org/advanced/middleware which has a good introduction as well, but happy to chat!

ekager (approved):

pocmo (approved):
✨ 👍

PR #15942, opened 2020-10-15
15278 detekt rule runblocking

MarcLeclair:
Added a `detekt` rule for `runBlocking` since we noticed that it can affect performance at times. So I switched `runBlocking` to `runBlockingCounter` (which just wraps the `runBlocking` call) so that we can monitor the # of times it is being called (through tests and other means) 

### Pull Request checklist

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/RunBlockingCounter.kt
```diff
@@ -0,0 +1,40 @@
...
+ */
+@VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+class RunBlockingCounter {
+    companion object {
+        var runBlockingCount = 0L
+    }
+}
+
+/**
+ * Wrapper around `runBlocking`
```
mcomella, line 29: Why do we do that? If someone wants to know what this method does, it's the first place they're going to look so it should be the place with the most robust description of what this method does and why we do it.

app/src/main/java/org/mozilla/fenix/RunBlockingCounter.kt
```diff
@@ -0,0 +1,40 @@
...
+ */
+fun <T> runBlockingIncrement(
+    context: CoroutineContext = EmptyCoroutineContext,
```
mcomella, line 32: I think you should use null to signify the absence of a value passed in. `EmptyCoroutineContext` may be a reasonable parameter to pass in now or in the future so we shouldn't use it as the optional argument state.

app/src/test/java/org/mozilla/fenix/RunBlockingCounterTest.kt
```diff
@@ -0,0 +1,18 @@
...
+
+import org.junit.Assert.assertEquals
+import org.junit.Test
+
+class RunBlockingCounterTest {
+
+    @Test
+    fun `GIVEN we call runBlockingCounter THEN the count should increase`() {
+        assertEquals(0L, RunBlockingCounter.runBlockingCount)
+        runBlockingIncrement {}
```
mcomella, line 15: Should we also test that the correct value is returned from runBlockingIncrement?

Note: I don't think tests are guaranteed to run in a specific order and static members don't get reset between tests so you'll need to zero the counter between tests, probably.

app/src/test/java/org/mozilla/fenix/RunBlockingCounterTest.kt
```diff
@@ -0,0 +1,18 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix
```
mcomella, line 5: If you change the package, don't forget to change this one too.

mozilla-detekt-rules/src/main/java/org/mozilla/fenix/detektrules/MozillaRunblockingCheck.kt
```diff
@@ -0,0 +1,38 @@
...
+        "us to monitor the code for performance regressions."
+
+/**
+ * A check to prevent us from working around mechanisms we implemented to prevent suppressing StrictMode.
+ *
+ * IF YOU UPDATE THIS FILE NAME, UPDATE CODE OWNERS.
+ */
+
+
+class MozillaRunblockingCheck(config: Config) : Rule(config) {
```
mcomella, line 24: We need to add this class to CODEOWNERS.

app/src/androidTest/java/org/mozilla/fenix/ui/PerformanceStartupTest.kt
```diff
@@ -28,6 +30,19 @@ private const val FAILURE_MSG = """StrictMode startup suppression count does not
...
+
+    If so, please do your best to implement a solution without blocking the main thread.
+    Please consult the perf team if you have questions or believe that using runBlocking
+    is the optimal solution.
+
+"""
+
 /**
  * A performance test to limit the number of StrictMode suppressions on startup.
  * This test was written by the perf team.
```
mcomella, line 48: We changed the file name: we need to change the class name and test name too.

I think `PerformanceStartupTest` is too general: how about `StartupExcessiveResourceUseTest` or similar?

mcomella, line 48: We changed the file name: we need to update CODEOWNERS.

config/detekt.yml
```diff
@@ -117,6 +117,8 @@ mozilla-detekt-rules:
     excludes: "**/*Test.kt, **/*Spec.kt, **/test/**, **/androidTest/**"
   MozillaCorrectUnitTestRunner:
     active: true
+  MozillaRunblockingCheck:
```
mcomella, line 120: Don't forget to update this name if you rename this

app/src/main/java/org/mozilla/fenix/RunBlockingCounter.kt
```diff
@@ -0,0 +1,40 @@
...
+import kotlin.coroutines.CoroutineContext
+import kotlin.coroutines.EmptyCoroutineContext
+
+/**
+ * Counts the number of runBlocking calls made
+ */
+@VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+class RunBlockingCounter {
+    companion object {
+        var runBlockingCount = 0L
```
mcomella, line 24: Given this runs in production code, I'm starting to think we should prevent overflow. The app shouldn't have a hard limit on how long it can run. Maybe in the setter? We could probably change it back to an Int in that case.

app/src/main/java/org/mozilla/fenix/RunBlockingCounter.kt
```diff
@@ -0,0 +1,40 @@
...
+    }
+}
+
+/**
+ * Wrapper around `runBlocking`
+ */
+fun <T> runBlockingIncrement(
```
mcomella, line 31: I don't love this name. While it's telling us what it does, it's also kind of vague – increment what?

Naming this seems really hard. The intentions I can think of are:
- tell them what it does (*increment, *perf, *monitor, ...)
- tell them it's preferred (*safe, *preferred, ...)

Maybe `runBlockingMonitored`? It sounds intimidating so maybe people will think twice before using it. 🤣 Or maybe they'll think we're the perf 👮‍♀️ :(

app/src/main/java/org/mozilla/fenix/RunBlockingCounter.kt
mcomella, line 24: This needs to be synchronized (if it's synchronized, does it automatically get volatile too? I'm not sure)

mcomella (changes requested):
This looks pretty good – I have fair number of comments but they're mostly nits. Requesting to see the changes to make sure the naming seems sound.

app/src/test/java/org/mozilla/fenix/RunBlockingCounterTest.kt
mcomella: > Note: I don't think tests are guaranteed to run in a specific order and static members don't get reset between tests so you'll need to zero the counter between tests, probably.

I think this is why the test failed on CI

app/src/main/java/org/mozilla/fenix/RunBlockingCounter.kt
```diff
@@ -0,0 +1,40 @@
...
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.runBlocking
+import kotlin.coroutines.CoroutineContext
+import kotlin.coroutines.EmptyCoroutineContext
+
+/**
+ * Counts the number of runBlocking calls made
+ */
+@VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+class RunBlockingCounter {
```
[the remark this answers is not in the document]
MarcLeclair: I was thinking it might move with our other Coroutines ideas but for now, I think leaving it as `object...{` is enough

mozilla-detekt-rules/src/main/java/org/mozilla/fenix/detektrules/MozillaRunblockingCheck.kt
```diff
@@ -0,0 +1,38 @@
...
+
+    override val issue = Issue(
+        "MozillaRunblockingCheck",
+        Severity.Performance,
+        "Prevents us from working around mechanisms we implemented to prevent suppressing StrictMode",
+        Debt.TEN_MINS
```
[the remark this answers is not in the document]
MarcLeclair: I changed it to 20min. I think I put it that way since I thought "people will probably not remove the `runBlocking` call but rather just call the `runBlockingCounter` "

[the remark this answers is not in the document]
mcomella: Oh, you're right – maybe this should be 5 min then. :P (is this detekt feature even useful?)

app/src/main/java/org/mozilla/fenix/RunBlockingCounter.kt
MarcLeclair: That's a fair point. I just tried to keep our function signature as closed as possible to the original `runBlocking` which takes an `EmptyCoroutineContext`

app/src/androidTest/java/org/mozilla/fenix/ui/PerformanceStartupTest.kt
MarcLeclair: I think that could work, I just changed it to that. I have some qualms with `StartupExcessiveResourceUseTest` since it specifies that we're checking for resources which to me sounds like we're checking for memory and other type of hardware related resources. However, I can't think of a better name that isn't longer or more concise with similar number of letters.

app/src/main/java/org/mozilla/fenix/RunBlockingCounter.kt
MarcLeclair: I can always add it to the `runBlockingIncrement` function. I just felt uneasy about adding more functionality (albeit its probably minimal) to a function that's just suppose to wrap one provided by Kotlin

app/src/main/java/org/mozilla/fenix/RunBlockingCounter.kt
MarcLeclair: I guess that can work. I was thinking  that 1) if the kdoc explains it further in details, then a looser name could be fine? Or 2) we could go longer, but more precise such as `runBlockingPerformanceTracker`? It seems a bit less intimidating and its purpose is conveyed a lot more from the name.

app/src/main/java/org/mozilla/fenix/RunBlockingCounter.kt
MarcLeclair: If I add `@Synchronize set` for this variable, isn't there an issue where Thread 1 --> Increases counter, then runs its method, returns and Thread 2 --> waits for the method to return before increasing?   Although I don't think that would happen, since the lock would be released as soon as we're done incrementing. I guess the only issue would be if a billion threads queue'd up at the same time to increment the counter, which I think is unlikely.

Maybe I'm overthinking it though, I added the `@Synchronized` with my latest fix. I saw nothing locally but it was just a thought.

app/src/main/java/org/mozilla/fenix/RunBlockingCounter.kt
mcomella: I think loose names are only okay if the function is frequently used, e.g. `map` or `add`. Given this is a function people aren't supposed to call a lot anyway, I think it's fine to have a longer name. However, I wanted to avoid making it too tied to performance so people don't think, "perf police!"

app/src/main/java/org/mozilla/fenix/RunBlockingCounter.kt
mcomella: Why do you want to avoid adding too much functionality to a wrapping method?

mcomella: Yes, that's possible. I'm concerned that the value will not be accessed correctly in concurrent situations, however.

Actually, I don't know how the java memory model works. Is assignment atomic? In that case, I don't think we need synchronized, just volatile. We could also change it to an `AtomicInteger` (or whatever) to to avoid having to figure it out.

app/src/androidTest/java/org/mozilla/fenix/ui/StartupExcessiveResourceUseTest.kt
```diff
@@ -39,20 +54,23 @@ private const val FAILURE_MSG = """StrictMode startup suppression count does not
  *
  * IF YOU UPDATE THE TEST NAME, UPDATE CODE OWNERS.
```
mcomella, line 55: Did you read the comment? 😝 Can you add this to code owners?

app/src/androidTest/java/org/mozilla/fenix/ui/StartupExcessiveResourceUseTest.kt
```diff
@@ -28,6 +30,19 @@ private const val FAILURE_MSG = """StrictMode startup suppression count does not
...
+    Did you know that using runBlocking can have negative perfomance implications?
+
+    If so, please do your best to implement a solution without blocking the main thread.
+    Please consult the perf team if you have questions or believe that using runBlocking
+    is the optimal solution.
+
+"""
+
 /**
  * A performance test to limit the number of StrictMode suppressions on startup.
```
mcomella, line 47: nit: update the comment

app/src/main/java/org/mozilla/fenix/components/searchengine/FenixSearchEngineProvider.kt
```diff
@@ -116,40 +116,47 @@ open class FenixSearchEngineProvider(
...
-                installedIdentifiers.contains(it.identifier)
-            }.sortedBy { it.name.toLowerCase(Locale.getDefault()) },
-            default = engineList.default?.let {
-                if (installedIdentifiers.contains(it.identifier)) {
-                    it
-                } else {
-                    null
+    fun installedSearchEngines(context: Context): SearchEngineList =
+        runBlockingIncrement {
+            val installedIdentifiers = installedSearchEngineIdentifiers(context)
```
mcomella, line 121: nit: can we avoid breaking blame? Or is it impossible because the line is too long?

app/src/main/java/org/mozilla/fenix/components/searchengine/FenixSearchEngineProvider.kt
```diff
@@ -116,40 +116,47 @@ open class FenixSearchEngineProvider(
...
-        )
-    }
+            )
+        }
 
-    fun allSearchEngineIdentifiers() = runBlocking {
-        loadedSearchEngines.await().list.map { it.identifier }
-    }
+    fun allSearchEngineIdentifiers() =
+        runBlockingIncrement {
```
mcomella, line 139: nit: avoid blame

app/src/main/java/org/mozilla/fenix/components/searchengine/FenixSearchEngineProvider.kt
```diff
@@ -162,14 +169,19 @@ open class FenixSearchEngineProvider(
...
+        isCustom: Boolean = false
+    ) = runBlockingIncrement {
         if (isCustom) {
             CustomSearchEngineStore.removeSearchEngine(context, searchEngine.identifier)
             reload()
         } else {
             val installedIdentifiers = installedSearchEngineIdentifiers(context).toMutableSet()
             installedIdentifiers.remove(searchEngine.identifier)
-            prefs(context).edit().putStringSet(localeAwareInstalledEnginesKey(), installedIdentifiers).apply()
+            prefs(context).edit()
```
mcomella, line 183: nit: you didn't change this line so don't break blame

app/src/main/java/org/mozilla/fenix/customtabs/FennecWebAppIntentProcessor.kt
```diff
@@ -58,7 +58,12 @@ class FennecWebAppIntentProcessor(
         val url = safeIntent.dataString
 
         return if (!url.isNullOrEmpty() && matches(intent)) {
-            val webAppManifest = runBlocking { loadManifest(safeIntent, url) }
+            val webAppManifest = runBlockingIncrement {
+                loadManifest(
```
mcomella, line 62: nit: one line for `loadManifest` (if not the whole block)

app/src/migration/java/org/mozilla/fenix/MigratingFenixApplication.kt
```diff
@@ -81,7 +80,7 @@ class MigratingFenixApplication : FenixApplication() {
             .migrateSettings()
             .build()
 
-        runBlocking {
+        runBlockingCounter {
```
mcomella, line 83: This name is inconsistent with others: `runBlockingIncrement`. Something is wrong.

config/detekt.yml
```diff
@@ -117,6 +117,9 @@ mozilla-detekt-rules:
     excludes: "**/*Test.kt, **/*Spec.kt, **/test/**, **/androidTest/**"
   MozillaCorrectUnitTestRunner:
     active: true
+  MozillaRunBlockingCheck:
+    active: true
+    excludes: "**/*Test.kt"
```
mcomella, line 122: nit: we should copy the same exclusion blocks as all the others. e.g. we don't want to enforce this in `androidTest` either.

app/src/main/java/org/mozilla/fenix/perf/RunBlockingCounter.kt
```diff
@@ -0,0 +1,45 @@
...
+import kotlinx.coroutines.runBlocking
+import kotlin.coroutines.CoroutineContext
+
+/**
+ * Counts the number of runBlocking calls made
+ */
+@VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+object RunBlockingCounter {
+    var count = 0L
+        @Synchronized set
```
mcomella, line 23: As per my earlier comment, this might just need to be volatile – can you check how java assignment works? Is `+=` atomic? Maybe we need `AtomicInteger` to increment.

app/src/main/java/org/mozilla/fenix/perf/RunBlockingCounter.kt
```diff
@@ -0,0 +1,45 @@
...
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.runBlocking
+import kotlin.coroutines.CoroutineContext
+
+/**
+ * Counts the number of runBlocking calls made
+ */
+@VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+object RunBlockingCounter {
+    var count = 0L
```
mcomella, line 22: Handle overflow? fwiw, I like handling it here, in the setter, rather than in runBlockingIncrement because it keeps the other method simple. That being said, it might not be possible with `AtomicInteger`.

app/src/main/java/org/mozilla/fenix/perf/RunBlockingCounter.kt
```diff
@@ -0,0 +1,45 @@
...
+ * count the number of runBlocking call to try to limit them as much as possible to encourage
+ * alternatives solutions whenever this function might be needed.
+ */
+fun <T> runBlockingIncrement(
+    context: CoroutineContext? = null,
+    action: suspend CoroutineScope.() -> T
+): T {
+    RunBlockingCounter.count += 1
+
+    context?.let {
```
mcomella, line 40: This doesn't do what you think it does. It does:
```kotlin
if (context != null) {
    val returnValue = runBlocking(context) { action() }
    if (returnValue != null) {
        runBlocking(action())
    }
}
```

I guess unless return return from the containing function and not the lambda. Even if the code is correct, the fact that it can be misinterpreted isn't great. In general, it's good to be careful when using symbols (`?:`) or unintuitively short functions (`let`, `run`, and friends) to express something.

I would prefer
```kotlin
return if (context != null) {
    runBlocking(it, ::action)
} else {
    runBlocking(::action)
}
```

Note that the function reference is an optimization that avoids us having to create a new lambda (if the compiler isn't smart enough to do it for us). Note also the deduplication of the `return` statement.

mozilla-detekt-rules/src/main/java/org/mozilla/fenix/detektrules/MozillaRunBlockingCheck.kt
```diff
@@ -0,0 +1,39 @@
...
+
+import io.gitlab.arturbosch.detekt.api.*
+import org.jetbrains.kotlin.psi.*
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.calls.callUtil.getCall
+import org.jetbrains.kotlin.resolve.calls.callUtil.getCalleeExpressionIfAny
+import kotlin.math.exp
+
+private const val VIOLATION_MSG = "Please use `org.mozilla.fenix.perf.RunBlockingCounter" +
+        ".fenixRunBlocking` instead because it allows us to monitor the code for performance " +
```
mcomella, line 15: this needs to be updated. Remember, the file name won't be used.

mcomella (changes requested):
Pretty close but we still have a few things to iron out such as:
- concurrency
- clarity of logic

app/src/main/java/org/mozilla/fenix/RunBlockingCounter.kt
MarcLeclair: From what I remember, if the interger is `volatile` it is kept in the main memory and access to that integer is the same throughout different threads. I think, however, that `AtomicInteger` can solve this issue

MarcLeclair: @mcomella  So I added `AtomicInteger` but I also think we need to `synchronized(...)` on the check for overflow since two threads my overlap. what do you think?

app/src/main/java/org/mozilla/fenix/perf/RunBlockingCounter.kt
```diff
@@ -0,0 +1,48 @@
...
+ * blocks the current thread interruptibly until its completion`. This can have negative
+ * side-effects on the our main thread which could lead to significant jank. This wrapper aims to
+ * count the number of runBlocking call to try to limit them as much as possible to encourage
+ * alternatives solutions whenever this function might be needed.
+ */
+fun <T> runBlockingIncrement(
+    context: CoroutineContext? = null,
+    action: suspend CoroutineScope.() -> T
+): T {
+    synchronized(lock = Any()) {
```
mcomella, line 38: This won't synchronized properly: each time we go through this, we lock a new instance so there is never any contention for this code block. You need to use a shared instance.

That being said, do we know that this class throws an exception on overflow? Maybe it just does something safe.

If not, I think the synchronization block could be replaced with something like `getAndUpdate`: https://developer.android.com/reference/java/util/concurrent/atomic/AtomicInteger?hl=en#getAndUpdate(java.util.function.IntUnaryOperator)

MarcLeclair: @mcomella  so `getAndUpdate` is min API 24. Also, the `AtomicInteger` seems to wrap around like a regular integer.  We could just have a function in the `object` and let it handle it itself?

app/src/main/java/org/mozilla/fenix/perf/RunBlockingCounter.kt
```diff
@@ -0,0 +1,55 @@
...
+import kotlin.coroutines.CoroutineContext
+
+/**
+ * Counts the number of runBlocking calls made
+ */
+@VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+object RunBlockingCounter {
+    var count = AtomicInteger(0)
+
+    fun incrementCounter() {
```
mcomella, line 25: nit: I think it's simpler to move this to be an extension function, e.g. `fun AtomicInteger.getAndIncrementNoOverflow()`. It lends itself to better testing (why isn't this method tested? :P), re-use, and more greatly discourages touching any code outside of the AtomicInteger, unlike a function in a companion object.

app/src/main/java/org/mozilla/fenix/perf/RunBlockingCounter.kt
```diff
@@ -0,0 +1,55 @@
...
+        var prev: Int
+        var next: Int
+        do {
+            prev = count.get()
+            next = prev + 1
+            if (next == Integer.MAX_VALUE) next = 0
```
mcomella, line 31: nit: while we're not going to hit this case, I think it's more honest to cap it at `Integer.MAX_VALUE` rather than reset the counter. e.g. if you look at the counter after running the program after a few days and its 100, you might be a little confused.

app/src/main/java/org/mozilla/fenix/perf/RunBlockingCounter.kt
```diff
@@ -0,0 +1,55 @@
...
+import kotlinx.coroutines.runBlocking
+import java.util.concurrent.atomic.AtomicInteger
+import kotlin.coroutines.CoroutineContext
+
+/**
+ * Counts the number of runBlocking calls made
+ */
+@VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+object RunBlockingCounter {
+    var count = AtomicInteger(0)
```
mcomella, line 23: nit: val

app/src/main/java/org/mozilla/fenix/perf/RunBlockingCounter.kt
```diff
@@ -0,0 +1,55 @@
...
+            next = prev + 1
+            if (next == Integer.MAX_VALUE) next = 0
+        } while (!count.compareAndSet(prev, next))
+    }
+}
+
+/**
+ * Wrapper around `runBlocking`. RunBlocking seems to be a "fix-all" to return values to the thread
+ * where the coroutine is called. The official doc explains runBlocking:  "Runs a new coroutine and
+ * blocks the current thread interruptibly until its completion`. This can have negative
```
mcomella, line 39: nit: "negative side-effects on our main thread"? Maybe we should just say, "block the main thread"

mcomella (approved):
The code looks good to me but I'd like some of the nits addressed. If we change `incrementCounter` to an extension function as I suggested, it might be good to flag me for re-review on that one specific section.

app/src/main/java/org/mozilla/fenix/ext/AtomicInteger.kt
```diff
@@ -0,0 +1,19 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.ext
+
+import java.util.concurrent.atomic.AtomicInteger
+
+/**
+ *  Increases an AtomicInteger safely.
+ */
+fun AtomicInteger.getAndIncrementNoOverflow(counter: AtomicInteger) {
```
mcomella, line 12: You shouldn't pass it as an argument – you should use `this` instead.

mcomella, line 12: Can we write tests for this?

mozilla-detekt-rules/src/main/java/org/mozilla/fenix/detektrules/MozillaRunBlockingCheck.kt
```diff
@@ -0,0 +1,39 @@
...
+package org.mozilla.fenix.detektrules
+
+import io.gitlab.arturbosch.detekt.api.*
+import org.jetbrains.kotlin.psi.*
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.calls.callUtil.getCall
+import org.jetbrains.kotlin.resolve.calls.callUtil.getCalleeExpressionIfAny
+import kotlin.math.exp
+
+private const val VIOLATION_MSG = "Please use `org.mozilla.fenix.perf.RunBlockingCounter.runBlockingIncrement`" +
```
mcomella, line 14: It's `org.mozilla.fenix.perf.runBlockingImplement` – top level functions don't include the file name in their fully-qualified class name.

mcomella (commented):
Some brief clean up changes – these could land as a follow-up so we can get this landed.

Note that you'll also have to rebase.

Issue #14619, opened 2020-09-02
Only show a tiny offset to indicate that the selected tab is not at the top of list in the tabs tray


gabrielluong:
In #13264, we took a shortcut to show that we are not at the top of the list in the tabs tray by scroll to the tab before the selected tab which is more than a little offset as desired by UX.

This issue is to implement the proper solution of only scrolling a small offset above the selected tab if it is not at the top of the list.

Mugurell:
@UX what should that offset be?
Maybe 1/3 or 1/4 of the size of a tab tray item? ([item which has 88dp in height](https://github.com/mozilla-mobile/fenix/blob/5153a3f2c5011a1f77c303832f3cd22692a68207/app/src/main/res/layout/tab_tray_item.xml#L10))

Maybe we should think about doing similar to show that that there are tabs at the bottom?
I think this needs a deeper investigation.

liuche:
I think 1/2 the size of the previous tab item would be fine, which would be better than what we have now. I don't think we necessarily need to do this for tabs on the bottom (if the user is on one of the last couple of tabs, they can see that there are a stack of tabs above, and therefore they can scrool).

cc @topotropic, if you have any additional thoughts (maybe due to the newest/oldest top ordering that you've been thinking baout).

topotropic:
Yes, 1/2 sounds good.

mcarare:
@gabrielluong I do not think it's in the scope of this issue, but AFAIU a different scroll should be set for grid view.

amkcpu:
The problem with using 1/2 of the tab is visible in the PR (#15955): Hiding half of the close button `x` makes it look like there is an arrow pointing up in the upper right corner.
For this reason and to save space, I would prefer showing only 1/3 or 1/4 of the tab.
What do you think?

AndiAJ:
Hi, **verified as fixed** on both latest **`Nightly Build 201022`** and **`Firefox RC 83.0.0-beta.1`** using the following devices:
• Google Pixel 3a (Android 11)
• Huawei Mate 20 Lite (Android 10)
• OnePlus A3 (Android 6.0.1)

► Video
![20201022-121955](https://user-images.githubusercontent.com/51314259/96851974-4790fe80-1461-11eb-8104-dfcb6b0d1abf.gif)

PR #15955, opened 2020-10-16
For #14619: Show half a tab top offset when scrolling to selected tab.

mcarare:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/tabtray/TabTrayView.kt
+++ b/app/src/main/java/org/mozilla/fenix/tabtray/TabTrayView.kt
@@ -652,13 +652,15 @@ class TabTrayView(
 
             // We offset the tab index by the number of items in the other adapters.
             // We add the offset, because the layoutManager is initialized with `reverseLayout`.
-            // We also add 1 to display the tab item above the selected browser tab.
             val recyclerViewIndex = selectedBrowserTabIndex +
                 collectionsButtonAdapter.itemCount +
-                syncedTabsController.adapter.itemCount +
-                1
+                syncedTabsController.adapter.itemCount
 
             layoutManager?.scrollToPosition(recyclerViewIndex)
+            smoothScrollBy(
+                0,
+                - resources.getDimensionPixelSize(R.dimen.tab_tray_tab_item_height) / 2
+            )
         }
     }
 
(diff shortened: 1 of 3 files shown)


mcarare:
This seems stuck. Closing and reopening.

app/src/main/java/org/mozilla/fenix/tabtray/TabTrayView.kt
```diff
@@ -659,6 +659,10 @@ class TabTrayView(
                 1
```
gabrielluong, line 659: I am wondering if we can remove this "1" that I added for displaying the tab item above the selected browser tab.

app/src/main/java/org/mozilla/fenix/tabtray/TabTrayView.kt
```diff
@@ -659,6 +659,10 @@ class TabTrayView(
...
 
             layoutManager?.scrollToPosition(recyclerViewIndex)
+            smoothScrollBy(
```
gabrielluong, line 662: We should avoid scrolling if the first item is selected.

gabrielluong (changes requested):

app/src/main/java/org/mozilla/fenix/tabtray/TabTrayView.kt
mcarare: I have reversed the scrolling and removed the `+1`.
So, we scroll to the selected item, then scroll "down" so half of the item above will be visible. If there is no item above, the scroll would not happen (there is not scroll space left).

gabrielluong (approved):
Changes look good. Just waiting for a ux review.

mcarare:
@topotropic can you review this from the UX side? Thank you!

topotropic (commented):
Thanks! Apologies for not getting back to the comment in the original issue that the x looks like and arrow when you cut it in half (https://github.com/mozilla-mobile/fenix/issues/14619#issuecomment-710645863). I just checked out the implementation and it looks like that when you are on tab 1, 2 or 3 we cut in half but once you are on the 4th tab we show 3/4 or so of the tab. Is that correct?

mcarare:
@topotropic No, we always show 1/2. When the selected tab is further down the list, it depends on the total number of items, as the selected tab could not be shown at the top if there are not sufficient items below it to scroll.
So, what is the proposed solution for the arrow issue?

topotropic (approved):
Looks good –  with the 2lines we either have the choice to roughly show 1/3 of the tab which would hide the URL when you have a 1line title or showing 2/3 of the tab which might be too much. The implemented solution is fine. Thank you!

Issue #14846, opened 2020-09-08
[Bug] Crash when starting Fenix whose all search engines deleted


hokonch:
## Steps to reproduce
1. Add a custom search engine (for example, example.com/?q=%s)
2. Delete search engines except for the custom engine.
3. Switch system language (for example, en -> ja)
4. Restart fenix and you can see local search engines and the custom engine enabled
5. Delete the custom engine
6. Switch back system language (ja -> en)
7. Crash and can't start fenix

## Crash report

7647615f-5c59-4224-884a-06b2ae3c75a1
java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
 * Sentry: https://sentry.prod.mozaws.net/operations/firefox-nightly/?query=798bbc0fa2984dba9d6944046d58f444
 * Socorro: https://crash-stats.mozilla.org/report/index/bp-5dbf983c-f3e9-40f5-a68f-7908d0200908
----

> java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
	at java.util.ArrayList.get(ArrayList.java:437)
	at mozilla.components.browser.search.SearchEngineManager.getProvidedDefaultSearchEngineAsync(SearchEngineManager.kt:7)
	at mozilla.components.browser.search.SearchEngineManager.getDefaultSearchEngineAsync(SearchEngineManager.kt:7)
	at mozilla.components.browser.search.SearchEngineManager$getDefaultSearchEngineAsync$1.invokeSuspend(Unknown Source:12)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:3)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:21)
	at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:1)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:11)
`

Diana-Rus:
Reproducing the crash with Samsung Galaxy S9 (Android 8) on Nightly 9/8 this issue is similar to the crash mentioned here #8406

Mugurell:
Well.. this is interesting..
I don't think the user should ever get into the situation in which she can delete all search engines.
Or maybe if that's the case we should just re-add a default one because otherwise how would search work anymore?
But what if there is not a default one for the locale?

hokonch:
[...]

PR #15958, opened 2020-10-16
For #14846 - Reinstall a search engine when none are available

Mugurell:
Fixes an edgecase where the user could "uninstall" all search engines for a
locale - by changing locales and delete all the ones from a previous locale.
If no engines are available we'll "reinstall" the default one or the first
available.

### Pull Request checklist

- [x] **Screenshots**: No screenshots. Not a UI change.
- [x] **Accessibility**: The code in this PR follows does not include any user facing features.

1. click on Show All Checks,


Mugurell:
Seems like the recent changes around this code also fixed the issue from #14846 .
Closing this PR since it's not needed anymore.

PR #15970, opened 2020-10-16
Refactored tabs tray grid item layout

Mugurell:
Refactored layout for the grid item of the tabs tray.
Since it needed new views (play button / border) placed at the exterior of the current items I've wrapped the previous in bigger FrameLayouts that should not affect performance but serve as a buffer space in which those items can extend.

### Pull Request checklist

- [x] **Tests**: No tests. UI refactoring.

NewGridLayout
[video](https://drive.google.com/file/d/11TSRErh_mQ1LWhGQfANfuiCkrVKbUpGu/view?usp=sharing)

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/tabtray/TabTrayView.kt
+++ b/app/src/main/java/org/mozilla/fenix/tabtray/TabTrayView.kt
@@ -43,6 +43,7 @@ import mozilla.components.browser.state.state.BrowserState
 import mozilla.components.browser.tabstray.TabViewHolder
 import mozilla.components.feature.syncedtabs.SyncedTabsFeature
 import mozilla.components.support.base.feature.ViewBoundFeatureWrapper
+import mozilla.components.support.ktx.android.util.dpToPx
 import org.mozilla.fenix.R
 import org.mozilla.fenix.browser.InfoBanner
 import org.mozilla.fenix.components.metrics.Event
@@ -373,6 +374,15 @@ class TabTrayView(
             }
 
             layoutManager = gridLayoutManager
+
+            // Ensure items have the same all around padding - 16 dp. Avoid the double spacing issue.
+            // A 8dp padding is already set in xml, pad the parent with the remaining needed 8dp.
+            updateLayoutParams<ConstraintLayout.LayoutParams> {
+                val padding = GRID_ITEM_PARENT_PADDING.dpToPx(resources.displayMetrics)
+                // Account for the already set bottom padding needed to accommodate the fab.
+                val bottomPadding = paddingBottom + padding
+                setPadding(padding, padding, padding, bottomPadding)
+            }
         }
     }
 
@@ -671,6 +681,8 @@ class TabTrayView(
         private const val SELECTION_DELAY = 500
         private const val NORMAL_HANDLE_PERCENT_WIDTH = 0.1F
         private const val COLUMN_WIDTH_DP = 180
+        // The remaining padding offset needed to provide a 16dp column spacing between the grid items.
+        const val GRID_ITEM_PARENT_PADDING = 8
     }
 }
 
--- a/app/src/main/res/layout/tab_tray_grid_item.xml
+++ b/app/src/main/res/layout/tab_tray_grid_item.xml
@@ -13,6 +13,7 @@ A FrameLayout here is an efficient way of having a views stack while allowing:
     xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:app="http://schemas.android.com/apk/res-auto"
     xmlns:tools="http://schemas.android.com/tools"
+    android:id="@+id/tab_tray_grid_item"
     android:layout_width="match_parent"
     android:layout_height="202dp"
     android:clipChildren="false"
--- a/app/src/main/res/layout/tab_tray_grid_item.xml
+++ b/app/src/main/res/layout/tab_tray_grid_item.xml
@@ -70,8 +70,8 @@ A FrameLayout here is an efficient way of having a views stack while allowing:
 
             <androidx.appcompat.widget.AppCompatImageButton
                 android:id="@+id/mozac_browser_tabstray_close"
-                android:layout_width="wrap_content"
-                android:layout_height="wrap_content"
+                android:layout_width="24dp"
+                android:layout_height="24dp"
                 android:background="?android:attr/selectableItemBackgroundBorderless"
                 android:contentDescription="@string/close_tab"
                 app:layout_constraintBottom_toTopOf="@+id/horizonatal_divider"
--- a/app/src/main/res/layout/tab_tray_grid_item.xml
+++ b/app/src/main/res/layout/tab_tray_grid_item.xml
@@ -2,114 +2,135 @@
 <!-- This Source Code Form is subject to the terms of the Mozilla Public
    - License, v. 2.0. If a copy of the MPL was not distributed with this
    - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
-<com.google.android.material.card.MaterialCardView xmlns:android="http://schemas.android.com/apk/res/android"
+
+<!--
+A FrameLayout here is an efficient way of having a views stack while allowing:
+    - a sufficient enough canvas for the play button to be outside translated
+    without it getting clipped by the GridLayoutManager.
+    - enough padding for the primary view that needs to be displayed with a border around it.
+-->
+<FrameLayout
+    xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:app="http://schemas.android.com/apk/res-auto"
     xmlns:tools="http://schemas.android.com/tools"
     android:layout_width="match_parent"
-    android:layout_height="wrap_content"
-    android:layout_marginStart="8dp"
-    android:layout_marginTop="16dp"
-    android:layout_marginEnd="8dp"
-    app:cardBackgroundColor="@color/photonWhite"
-    app:cardCornerRadius="@dimen/tab_tray_grid_item_border_radius"
-    app:cardElevation="0dp"
-    app:strokeColor="@color/photonLightGrey30"
-    app:strokeWidth="1dp">
+    android:layout_height="202dp"
+    android:clipChildren="false"
+    android:clipToPadding="false"
+    android:importantForAccessibility="no"
+    android:padding="8dp">
 
-    <androidx.constraintlayout.widget.ConstraintLayout
-        android:id="@+id/tab_item"
+    <com.google.android.material.card.MaterialCardView
         android:layout_width="match_parent"
-        android:layout_height="match_parent">
+        android:layout_height="wrap_content"
+        android:layout_gravity="center"
+        android:importantForAccessibility="yes"
+        app:cardBackgroundColor="@color/photonWhite"
+        app:cardCornerRadius="@dimen/tab_tray_grid_item_border_radius"
+        app:cardElevation="0dp"
+        app:strokeColor="@color/photonLightGrey30"
+        app:strokeWidth="1dp">
 
-        <ImageButton
-            android:id="@+id/play_pause_button"
-            android:layout_width="24dp"
-            android:layout_height="24dp"
-            android:layout_marginTop="23dp"
-            android:background="?attr/selectableItemBackgroundBorderless"
-            android:contentDescription="@string/mozac_feature_media_notification_action_pause"
-            android:elevation="10dp"
-            android:visibility="gone"
-            app:layout_constraintEnd_toEndOf="parent"
-            app:layout_constraintTop_toTopOf="parent"
-            app:srcCompat="@drawable/media_state_play" />
-
-        <ImageView
-            android:id="@+id/mozac_browser_tabstray_favicon_icon"
-            android:layout_width="16dp"
-            android:layout_height="16dp"
-            android:layout_marginStart="8dp"
-            android:importantForAccessibility="no"
-            app:layout_constraintBottom_toTopOf="@id/horizonatal_divider"
-            app:layout_constraintStart_toStartOf="parent"
-            app:layout_constraintTop_toTopOf="parent" />
-
-        <TextView
-            android:id="@+id/mozac_browser_tabstray_title"
-            android:layout_width="0dp"
-            android:layout_height="30dp"
-            android:layout_alignParentTop="true"
-            android:ellipsize="none"
-            android:fadingEdgeLength="25dp"
-            android:paddingHorizontal="7dp"
-            android:paddingVertical="5dp"
-            android:requiresFadingEdge="horizontal"
-            android:singleLine="true"
-            android:textColor="@color/photonInk80"
-            android:textSize="14sp"
-            android:visibility="visible"
-            app:layout_constraintEnd_toStartOf="@id/mozac_browser_tabstray_close"
-            app:layout_constraintStart_toEndOf="@id/mozac_browser_tabstray_favicon_icon"
-            app:layout_constraintTop_toTopOf="parent"
-            tools:text="Example Domain" />
-
-        <androidx.appcompat.widget.AppCompatImageButton
-            android:id="@+id/mozac_browser_tabstray_close"
-            android:layout_width="wrap_content"
-            android:layout_height="wrap_content"
-            android:background="?android:attr/selectableItemBackgroundBorderless"
-            android:contentDescription="@string/close_tab"
-            app:layout_constraintBottom_toTopOf="@+id/horizonatal_divider"
-            app:layout_constraintEnd_toEndOf="parent"
-            app:layout_constraintTop_toTopOf="parent"
-            app:srcCompat="@drawable/mozac_ic_close"
-            app:tint="@color/photonInk80" />
-
-        <View
-            android:id="@+id/horizonatal_divider"
+        <androidx.constraintlayout.widget.ConstraintLayout
+            android:id="@+id/tab_item"
             android:layout_width="match_parent"
-            android:layout_height="1dp"
-            android:background="@color/photonLightGrey30"
-            app:layout_constraintTop_toBottomOf="@+id/mozac_browser_tabstray_title" />
-
-        <androidx.cardview.widget.CardView
-            android:id="@+id/mozac_browser_tabstray_card"
-            android:layout_width="match_parent"
-            android:layout_height="@dimen/tab_tray_grid_item_thumbnail_height"
-            android:backgroundTint="?tabTrayThumbnailItemBackground"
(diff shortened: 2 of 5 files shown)


Mugurell:
Regarding a11y, there is one issue with the play button which is too small.
.. though [playPauseButtonView.increaseTapArea](https://github.com/mozilla-mobile/fenix/blob/5bd829f389ad439719642baaf6a70b909df10e6f/app/src/main/java/org/mozilla/fenix/tabtray/TabTrayViewHolder.kt#L97) is already being used.

Here is a complete report:


_Image contrast
org.mozilla.fenix.debug:id/mozac_browser_tabstray_thumbnail
The image's contrast ratio is 1.91. This ratio is based on an estimated foreground color of #5B5B66 and an estimated background color of #32313C. Consider increasing this ratio to 3.00 or greater. This item may be obscured by other on-screen content. Consider manually testing this item's contrast._

_Image contrast
org.mozilla.fenix.debug:id/mozac_browser_tabstray_thumbnail
The image's contrast ratio is 1.31. This ratio is based on an estimated foreground color of #E0E0E6 and an estimated background color of #FFFFFF. Consider increasing this ratio to 3.00 or greater. This item may be obscured by other on-screen content. Consider manually testing this item's contrast._

_Text contrast
org.mozilla.fenix.debug:id/counter_text
The item's text contrast ratio is 4.12. This ratio is based on an estimated foreground color of #9059FF and an estimated background color of #1C1B22. Consider using colors that result in a contrast ratio greater than 4.50 for small text, or 3.00 for large text._

_Text contrast
org.mozilla.fenix.debug:id/new_tab_button
The item's text contrast ratio is 4.12. This ratio is based on an estimated foreground color of #1C1B22 and an estimated background color of #9059FF. Consider using colors that result in a contrast ratio greater than 4.50 for small text, or 3.00 for large text._

Touch target
org.mozilla.fenix.debug:id/play_pause_button
This item's size is 24dp x 24dp. Consider making this touch target 48dp wide and 48dp high or larger. A parent container may be handling touch events for this item. If selecting the larger container performs the same action as selecting this item, consider defining this item as not clickable. If a different action is performed, consider increasing the size of this item.

app/src/main/res/drawable/tab_tray_grid_item_selected_background.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
```
gabrielluong, line 1: I think this should be named tab_tray_grid_item_selected_border.xml

app/src/main/res/drawable/tab_tray_grid_item_selected_background.xml
```diff
@@ -0,0 +1,15 @@
...
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<shape
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    android:shape="rectangle">
+
+    <stroke
+        android:width="4dp"
+        android:color="#592ACB" />
```
gabrielluong, line 11: Can we use the @color/photonViolet70?

app/src/main/res/layout/tab_tray_grid_item.xml
```diff
@@ -1,115 +1,134 @@
-<?xml version="1.0" encoding="utf-8"?>
-<!-- This Source Code Form is subject to the terms of the Mozilla Public
+<?xml version="1.0" encoding="utf-8"?><!-- This Source Code Form is subject to the terms of the Mozilla Public
```
gabrielluong, line 1: nit: license formatting

gabrielluong (changes requested):
I am noticing some issues with the margin/padding. See screenshot. I am also getting crashes probably from the layout recomputing when we switch orientation between horizontal and vertical

Screen Shot 2020-10-17 at 10 37 14 AM

app/src/main/res/drawable/tab_tray_grid_item_selected_background.xml
Mugurell: Didn't know that was already defined. Thanks!

Mugurell:
Thanks!
- The overscroll effect issue was because I've used margins. Changed commit3 to use padding. Coupled with the already set `clipToPadding="false"` the effect should now be as intended.
- Does the crash you're seeing stems from `FenixApplication.onConfigurationChanged`? That would be a known issue not related to the changes here - https://github.com/mozilla-mobile/android-components/issues/8639.

gabrielluong:
> * Does the crash you're seeing stems from `FenixApplication.onConfigurationChanged`? That would be a known issue not related to the changes here - [mozilla-mobile/android-components#8639](https://github.com/mozilla-mobile/android-components/issues/8639).

I didn't see the same logs. I could not reproduce the crash with the old grid view so I would like to make sure we kind of go through the changes with a bit of scrutiny to rule out that it is caused by any of our changes.

Mugurell:
> > * Does the crash you're seeing stems from `FenixApplication.onConfigurationChanged`? That would be a known issue not related to the changes here - [mozilla-mobile/android-components#8639](https://github.com/mozilla-mobile/android-components/issues/8639).
> 
> I didn't see the same logs. I could not reproduce the crash with the old grid view so I would like to make sure we kind of go through the changes with a bit of scrutiny to rule out that it is caused by any of our changes.

Can you please post the logcat for the crash you're seeing?
I'm encountering a crash when I rotate the screen, but that is because of mozilla-mobile/android-components#8639, crash which can be avoided by commenting out https://github.com/mozilla-mobile/fenix/blob/0e2657e2620457c502e6b79b40f67ced6f8f0117/app/src/main/java/org/mozilla/fenix/FenixApplication.kt#L435-L437)

gabrielluong:
> > > * Does the crash you're seeing stems from `FenixApplication.onConfigurationChanged`? That would be a known issue not related to the changes here - [mozilla-mobile/android-components#8639](https://github.com/mozilla-mobile/android-components/issues/8639).
> > 
> > 
> > I didn't see the same logs. I could not reproduce the crash with the old grid view so I would like to make sure we kind of go through the changes with a bit of scrutiny to rule out that it is caused by any of our changes.
> 
> Can you please post the logcat for the crash you're seeing?
> I'm encountering a crash when I rotate the screen, but that is because of [mozilla-mobile/android-components#8639](https://github.com/mozilla-mobile/android-components/issues/8639), crash which can be avoided by commenting out
> 
> https://github.com/mozilla-mobile/fenix/blob/0e2657e2620457c502e6b79b40f67ced6f8f0117/app/src/main/java/org/mozilla/fenix/FenixApplication.kt#L435-L437
> 
> )

```
2020-10-19 21:55:03.838 23982-24050/org.mozilla.fenix.debug E/GeckoConsole: [JavaScript Error: "[Exception... "Component returned failure code: 0x80520006 (NS_ERROR_FILE_TARGET_DOES_NOT_EXIST) [nsIProcess.init]"  nsresult: "0x80520006 (NS_ERROR_FILE_TARGET_DOES_NOT_EXIST)"  location: "JS frame :: resource://gre/modules/CrashService.jsm :: runMinidumpAnalyzer/< :: line 67"  data: no]"]
    runMinidumpAnalyzer/<@resource://gre/modules/CrashService.jsm:67:15
    runMinidumpAnalyzer@resource://gre/modules/CrashService.jsm:60:10
    maybeRunMinidumpAnalyzer@resource://gre/modules/CrashService.jsm:33:9
    addCrash@resource://gre/modules/CrashService.jsm:223:11
2020-10-19 21:55:04.126 23982-24050/org.mozilla.fenix.debug E/GeckoConsole: [JavaScript Error: "TypeError: can't access property "sendAsyncMessage", this._manager.messageManager is null" {file: "chrome://geckoview/content/geckoview.js" line: 510}]
    _updateContentModuleState@chrome://geckoview/content/geckoview.js:510:5
    set enabled@chrome://geckoview/content/geckoview.js:490:10
    init/</<@chrome://geckoview/content/geckoview.js:106:9
    forEach@chrome://geckoview/content/geckoview.js:137:19
    init/<@chrome://geckoview/content/geckoview.js:105:12
2020-10-19 21:55:04.157 23982-24050/org.mozilla.fenix.debug E/GeckoConsole: [JavaScript Error: "[Exception... "Component returned failure code: 0x80520001 (NS_ERROR_FILE_UNRECOGNIZED_PATH) [nsIXPCComponents_Utils.readUTF8URI]"  nsresult: "0x80520001 (NS_ERROR_FILE_UNRECOGNIZED_PATH)"  location: "JS frame :: resource://gre/modules/L10nRegistry.jsm :: L10nRegistry.loadSync :: line 684"  data: no]"]
    L10nRegistry.loadSync@resource://gre/modules/L10nRegistry.jsm:684:19
    fetchFile@resource://gre/modules/L10nRegistry.jsm:599:31
    generateResourceSetSync/<@resource://gre/modules/L10nRegistry.jsm:504:19
    generateResourceSetSync@resource://gre/modules/L10nRegistry.jsm:499:22
    generateResourceSetsForLocaleSync@resource://gre/modules/L10nRegistry.jsm:441:44
    generateResourceSetsForLocaleSync@resource://gre/modules/L10nRegistry.jsm:450:15
    generateBundlesSync@resource://gre/modules/L10nRegistry.jsm:178:7
    next@resource://gre/modules/Localization.jsm:147:49
    formatWithFallbackSync@resource://gre/modules/Localization.jsm:287:5
    formatValuesSync@resource://gre/modules/Localization.jsm:396:17
    formatValueSync@
2020-10-19 21:55:04.891 23982-23982/org.mozilla.fenix.debug E/lla.fenix.debu: Invalid ID 0x00000000.
```
I am getting the following when I change the orientation. Specifically, I am seeing the tab crash upon orientation change (showing "Sorry, Firefox Preview can't load that page.") and then I think the crash manager crashes.

gabrielluong:
I am still seeing the scrollbar has some right margin from the edge of the screen.

app/src/main/java/org/mozilla/fenix/tabtray/TabTrayView.kt
```diff
@@ -373,6 +374,15 @@ class TabTrayView(
             }
 
             layoutManager = gridLayoutManager
+
+            // Ensure items have the same all around border - 16 dp. Avoid the double spacing issue.
```
gabrielluong, line 378: Did you mean padding instead of border?

app/src/main/res/layout/tab_tray_grid_item.xml
```diff
@@ -2,114 +2,146 @@
...
-    app:cardElevation="0dp"
-    app:strokeColor="@color/photonLightGrey30"
-    app:strokeWidth="1dp">
-
-    <androidx.constraintlayout.widget.ConstraintLayout
-        android:id="@+id/tab_item"
+    android:clipChildren="false"
+    android:clipToPadding="false"
+    android:importantForAccessibility="no"
+    android:padding="8dp">
```
gabrielluong, line 22: I think this should actually be 4dp now from comparing the before and after with this PR. I think the external border does actually add a 4dp padding even though there is a -4dp margin. Can you verify? It might be best to also present a before and after screenshot of the difference. I know prior the gap between the tabs were 16dp since they had a marginStart and marginEnd of 8dp.

app/src/main/java/org/mozilla/fenix/tabtray/TabTrayView.kt
gabrielluong, line 378: I am gonna try rephrasing this comment so that there is more context around it, but I think we will need to touch up the explanation. I know you pointed me to https://stackoverflow.com/questions/28531996/android-recyclerview-gridlayoutmanager-column-spacing/28533234#28533234 regarding the issue:

> Set the correct column spacing in the grid layout to ensure a column spacing of 16dp between each grid item and avoid double spacing issue [TODO: I think the double spacing issue should be explained more clearly]. A 8dp padding is already set in tab_tray_grid_item.xml, so the parent needs to be padded with a remaining 8dp.

The point is to be pretty clear what this bit of code is doing and the problem it is solving. I know there's also the item decoration approach to solving the column spacing so I am wondering what was the difference between the 2 approaches.

app/src/main/res/layout/tab_tray_grid_item.xml
```diff
@@ -2,114 +2,146 @@
...
-<com.google.android.material.card.MaterialCardView xmlns:android="http://schemas.android.com/apk/res/android"
+
+<!--
+A FrameLayout here is an efficient way of having a views stack while allowing:
+    - a sufficient enough canvas for the play button to be outside translated
+    without it getting clipped by the GridLayoutManager
+    - enough padding for the primary view that needs to be displayed
+    where a border around it can be placed
+-->
+<FrameLayout
```
gabrielluong, line 13: Open question - I am curious if this needs to be a FrameLayout compare to a Linear or RelativeLayout? I don't really know the implications of having a FrameLayout nested inside of a FrameLayout compare to other Layouts.

app/src/main/java/org/mozilla/fenix/tabtray/TabTrayView.kt
```diff
@@ -671,6 +681,12 @@ class TabTrayView(
         private const val SELECTION_DELAY = 500
         private const val NORMAL_HANDLE_PERCENT_WIDTH = 0.1F
         private const val COLUMN_WIDTH_DP = 180
+        // UI specs say grid items need to have 16dp padding.
+        // The layout already has a 8dp padding.
+        // (also used as buffer space for the external border and outside translation of the play button)
+        // We leverage the double spacing issue to have 8dp * 2 as in-between items padding.
+        // Need to add another 8dp of padding to the parent.
```
gabrielluong, line 688: I think we can simplify this text a bit since we already have a pretty thorough explanation above. 

Something like:

> The remaining padding offset needed to provide a 16dp column spacing between the grid items.

gabrielluong (commented):
Gonna add @ekager for a super review.

app/src/main/res/layout/tab_tray_grid_item.xml
Mugurell: To me FrameLayout made more sense from a mental model, it basically allowing for a stack of:
- a big enough space for the thumbnail and where the play pause button to be translated
- a big enough space where an external border can be place
- then the item we want to show.
A ConstraintLayout or a even RelativeLayout (even with it's two measure passes) in place of the two FrameLayouts would probably get the same results with the same performance. I think the layout is pretty simple to see any performance hits by going either way.

Mugurell:
> I am still seeing the scrollbar has some right margin from the edge of the screen.

Thanks!
Moved it to the outer edge of te screen.

app/src/main/java/org/mozilla/fenix/tabtray/TabTrayView.kt
Mugurell: This to me seems simpler, easier to maintain and more performant since it's a one-time operation instead of setting new dimensions for each new item view.

Mugurell: Indeed, thanks!

app/src/main/java/org/mozilla/fenix/tabtray/TabTrayView.kt
Mugurell: Sure. Thanks!

app/src/main/res/layout/tab_tray_grid_item.xml
Mugurell: This `8dp` padding is needed here because the play/pause button will be `-8dp` externally translated.

Mugurell: For the `tab_tray_grid_item_border` I've used `layout_margin="-4dp"` to ensure a 4dp space for the border to the outside of the item and then `android:padding="4dp"` to ensure the item with `layout_width="match_parent"` doesn't extend to occupy that -4dp space.
Will try checking the spacing between items. The spacing does look bigger than before.

Mugurell:
One thing that would help this would be to eliminate those negative margins.
I was using them because I built the layout from the center out - because we knew the item should have 186dp height.
But instead of using negative margins for the parent layouts we can just add that size to their height, with the 4dp (times 2) border and 4dp (times 2) space for the play button to be translated giving us a layout height of 202dp.

app/src/main/res/layout/tab_tray_grid_item.xml
Mugurell: Indeed, the padding should have been 4dp there. Thanks!

app/src/main/res/layout/tab_tray_grid_item.xml
Mugurell: Refactored the original patch to use just one new layout as parent for the item, this being needed because of the exterior border and to allow for the play/pause button being outside translated.
So this should be as performant as possible.

Mugurell:
Refactored the original patch to use just one new layout as parent for the item, this being needed because of the exterior border and to allow for the play/pause button being outside translated.
So this should be as performant as possible.
Also got rid of the negative margins.
Used [Yet another android layout inspector](https://github.com/Grigory-Rylov/android-layout-inspector) to easily confirm the 16dp spacing between items and between items and the RecyclerView and also the 186dp height of the item, 202dp height with border and spacing needed for translating the play/pause button.

Before/after showing the same item height:


&nbsp;
I'd appreciate if in the next review the reviewer would also comment out https://github.com/mozilla-mobile/fenix/blob/0e2657e2620457c502e6b79b40f67ced6f8f0117/app/src/main/java/org/mozilla/fenix/FenixApplication.kt#L435-L437 to be sure the crash talked above is happening because of this changes or others.
(For me the latest master also crashes when changing orientation but does not anymore after commenting out ^^. This would be an issue only on debug.)

&nbsp;
We remain with a question for @topotropic about the order of announcements for a11y:
Currently when swiping to a new item, and advancing in the same direction talkback reads
- the tab layout - tab title is read
- the close button - "close tab < tab title > button. double tap to activate"
- play button (if there is one) - "play button. double tap to activate"
Is the order and the spoken text okay?

&nbsp;
Another aspect where we'd need UX feedback is regarding the touch target for the play/pause button I think it cannot be increased while also following the current design since there is no space where to extend it's touch receiver. The play/pause button already extends to the very corner of the layout and occupies another 8dp. To make it bigger we could either increase the spacing between items - to allow for more space for where it could extend or move it more to the center of the item.


&nbsp;
To easily test these changes one could install a CI build from https://firefox-ci-tc.services.mozilla.com/tasks/YuVY6uzUQ06uJViR7MghQQ#artifacts.

gabrielluong (approved):

PR #16021, opened 2020-10-19
 Update Android Components, Kotlin, Gradle plugin

pocmo:
Okay, here we go again. We landed the updates in AC again and this is the matching Fenix PR with the AC version bump and other required changes.


pocmo:
Failed ui test passes locally, so I assume it's just an intermittent. Let's retry..

pocmo:
It's green. It's greeeeen. 💚

rocketsroger (approved):

PR #16051, opened 2020-10-20
FennecWebAppIntentProcessor: Catch JSONException.

pocmo:
Randomly noticed that this can crash if the provided JSON could not be parsed.


app/src/main/java/org/mozilla/fenix/customtabs/FennecWebAppIntentProcessor.kt
```diff
@@ -113,6 +116,10 @@ class FennecWebAppIntentProcessor(
 
             WebAppManifestParser().parse(manifestField).getOrNull()
         } catch (e: IOException) {
+            logger.error("Failed to parse web app manifest due to IOException", e)
+            null
+        } catch (e: JSONException) {
+            logger.error("Failed to parse web app manifest due to JSONException", e)
```
csadilek, line 122: We're still missing multi-catch in Kotlin! ☕

csadilek (approved):
👍

app/src/main/java/org/mozilla/fenix/customtabs/FennecWebAppIntentProcessor.kt
pocmo: Yeah :)

Issue #16054, opened 2020-10-20
Enable pull to refresh in Nightly and Debug builds


Mugurell:
With all the big issues from https://github.com/mozilla-mobile/fenix/issues/9766 being solved pull to refresh should be enabled again in Nightly.

### What is the user problem or growth opportunity you want to see solved?
An easier, more enjoyable UX.

### How do you know that this problem exists today? Why is this important?
User requested feature usually found in many Android apps.

### Who will benefit from it?
All users.
We tentatively added a setting to enable/disable this feature but as per https://github.com/mozilla-mobile/fenix/issues/15629 it should not be used.

liuche:
We discussed this and there is still one outstanding issue on Google Maps (see meta #9766, and #16055) but this should be good try again on Nightly!

abodea:
**Verified** as fixed on the latest **Nightly build 10/21** with **Google Pixel 4 XL (11), Samsung Galaxy S7 (7), Google Pixel 4 (10), Nexus 9 (7), Huawei P20 (9), OnePlus A3 (6.0.1).**

Please note that many websites were checked and here are few examples: **bing.com/covidtwitter.com, uBlock settings page, phonearena.com, pinterest.com, ifehacker.com.**

Note that this is working as expecting but because of the large number of websites there can still be issues and we will keep testing/investigating this but we will add all the information to the **[Meta bug.
](https://github.com/mozilla-mobile/fenix/issues/9766)** 
As mentioned above there are still issues that are present in the **[Meta bug](https://github.com/mozilla-mobile/fenix/issues/9766)** and the **[Google Issue.](https://github.com/mozilla-mobile/fenix/issues/16055)**

Issue #15629, opened 2020-10-02
Optimize gesture settings


Mugurell:
As a followup up to #10240 and #13442, based on UX feedback the `Gestures` settings group should be phased out.

https://github.com/mozilla-mobile/fenix/issues/13442#issuecomment-693027043:
> An update from UX on this from @mheubusch:
> 1. **Pull to refresh**: when it is ready it can go to release, NO setting to control it
Pull to refresh is a default for many Android browsers. Here, we are shipping something to catch up to their feature set. Because it’s shipped with so many browsers, we think that people would not only expect pull to refresh to be a default behaviour. We also think that they don’t want to turn it off. So we should ship this gesture to Release, and not create a setting for it.
>
> 2. **Dynamic/Pinned toolbar**: when this is ready it can go to Release, with a setting under "Toolbar" section
Hide address bar on scroll is a default for us and many other Android browsers. Some contributors wanted the address bar to always show, independent of scrolling. That’s fine. We can ship it on release, but we won’t make this persistent toolbar behaviour a default. It will be controllable under Toolbar Settings.
> 
> 3. **Swipe toolbar up for Tabs Tray** #11862: this feature should stay on Nightly, but we can include a setting
Other gestures control more advanced features. For example: swipe left/right on address bar to switch tabs. So far, we only received a signal that people want this feature through our contributors. But they’re not our general audience. So let’s hold off on shipping these other gestures, and keep them only in Nightly.
> 
> 4. **Swipe toolbar left/right to switch tabs**: TBD (since it is already in release)

Mugurell:
I'm not quite sure about the timeline for this changes regarding release channels .
Should the changes be implemented right away and then depending on feature availability be uplifted to Release? 

Currently Release doesn't have the new `Gestures` settings group. I understand we should hold it off of Release.

Mugurell:
Regarding
> 3. **Swipe toolbar up for Tabs Tray** #11862: this feature should stay on Nightly, but we can include a setting
Other gestures control more advanced features. For example: swipe left/right on address bar to switch tabs.

Can we get a mockup & strings?
@betsymi @mheubusch

apbitner:
@betsymi Let me know if any mocks are needed, happy to help.

betsymi:
For Nightly only, this can go under **Gestures** on the Customize screen and the string is: _Swipe toolbar up to view tabs_ 

Attached the screenshot of our current Customize screen for reference. 
[...]

PR #16060, opened 2020-10-20
For #16054 - Enable pull to refresh webpage in Nightly and Debug

Mugurell:
### Pull Request checklist

- [x] **Tests**: No tests. Small change.
- [x] **Accessibility**: The code in this PR does not include a11y any user facing features.

PullToRefresh

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/FeatureFlags.kt
+++ b/app/src/main/java/org/mozilla/fenix/FeatureFlags.kt
@@ -12,7 +12,7 @@ object FeatureFlags {
      * Pull-to-refresh allows you to pull the web content down far enough to have the page to
      * reload.
      */
-    const val pullToRefreshEnabled = false
+    val pullToRefreshEnabled = Config.channel.isNightlyOrDebug
 
     /**
      * Shows Synced Tabs in the tabs tray.


app/src/main/java/org/mozilla/fenix/settings/CustomizationFragment.kt
```diff
@@ -142,11 +142,6 @@ class CustomizationFragment : PreferenceFragmentCompat() {
     }
 
     private fun setupGesturesCategory() {
-        requirePreference<SwitchPreference>(R.string.pref_key_website_pull_to_refresh).apply {
```
gabrielluong, line 145: I am wondering if we should remove this setting from the Customization Settings with this PR. Maybe we should file a separate issue for removing the setting to get UX to weigh in.

gabrielluong (changes requested):

app/src/main/java/org/mozilla/fenix/settings/CustomizationFragment.kt
Mugurell: Removing that setting follows UX recommendations from https://github.com/mozilla-mobile/fenix/issues/15629

Mugurell: Would've been better to resolve remove this setting as part of that ticket but that refers to other changes also and this will probably be merged before the others, Just wanted to have it right from the start.

gabrielluong: Gotcha, I think it makes sense if we were enabling Pull to Refresh beyond Nightly, but since we are only enabling for Nightly, maybe we should keep it for the time being and have it serve as an easy debugging tool to see if problems with pull to refresh is due to the feature itself. We can keep the settings removal in a separate PR until we have a bit more QA time baked in.

gabrielluong (approved):

PR #16094, opened 2020-10-21

MickeyMoz:


csadilek:
UI test failure was `testStrictVisitSheetDetails`. Let's see if it's intermittent, looks like it should be fixed: https://github.com/mozilla-mobile/fenix/issues/14524 ?

csadilek:
There's also a bunch of other UI test failures in `HistoryTest`.

csadilek:
Failures where caused by regression in GV, we're publishing a new A-C with latest GV that should fix this.

PR #16137, opened 2020-10-22
Closes #11285: Replace Session[Manager] observers in BaseBrowserFragment

### LINKED ISSUE #11285: BaseBrowserFragment: Refactor session manager observers to use store
`BaseBrowserFragment` is currently using both `SessionManager` and the browser store. Let's refactor it to use the store only.

We want to get rid of `SessionManager` observers so we can finish the A-C refactoring to browser state.

csadilek:
This replaces two `SessionManager` observes and one `Session` observer in `BaseBrowserFragment` and removes a workaround that is no longer needed (we launched a coroutine to prevent a `ConcurrentModificationException` which is no longer necessary).

The `Session` observer was used to make sure the toolbar is expanded on page navigation which is now handled by `expandToolbarOnNavigation`.

We had one `SessionManager` observer to initialize the UI, if needed, and update the theme. A second one to reset fullscreen mode, expand the toolbar, and resume downloads in case the UI was initialized and switched to a different session. This is now all captured in one place / one observer.

Additionally this adds unit tests for all of the above. 

*Let's also do some manual smoke testing (pause / resume / scroll and navigate to a different page) before landing this. I'd also vote for landing on Monday so we're around in case we see crashes.*

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
@@ -14,6 +14,7 @@ import android.view.View
 import android.view.ViewGroup
 import android.view.accessibility.AccessibilityManager
 import androidx.annotation.CallSuper
+import androidx.annotation.VisibleForTesting
 import androidx.coordinatorlayout.widget.CoordinatorLayout
 import androidx.core.net.toUri
 import androidx.core.view.isVisible
@@ -30,7 +31,6 @@ import kotlinx.coroutines.Dispatchers.IO
 import kotlinx.coroutines.Dispatchers.Main
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.Job
-import kotlinx.coroutines.MainScope
 import kotlinx.coroutines.flow.collect
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.mapNotNull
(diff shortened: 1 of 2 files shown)


pocmo (approved):
👍 

✨  tests ✨

pocmo:
> [task 2020-10-22T21:52:10.931Z]   TEST: GIVEN we're in a release build WHEN resetAfter is called THEN the old policy is not set
[task 2020-10-22T21:52:11.530Z]     [Robolectric] org.mozilla.fenix.StrictModeManagerTest.GIVEN we're in a release build WHEN resetAfter is called THEN the old policy is not set: sdk=28; resources=BINARY
[task 2020-10-22T21:53:03.430Z]     I/MonitoringInstr: Instrumentation started!
[taskcluster:error] Task timeout after 7200 seconds. Force killing container.

Looks like we have a problem with a test that can time out? Anyhow, looks like it is not related to this change here, so I'll retry.

csadilek:
> Looks like we have a problem with a test that can time out? Anyhow, looks like it is not related to this change here, so I'll retry.

Yeah, timeouts are unrelated. I saw them a few times yesterday. It seems we're running OOM too occasionally running all tests on CI.

eliserichards:
Tested on Pixel 3, API 29; and Pixel 3, API 30. No problems that I've found 👍

pocmo (approved):
👍

csadilek:
(Unit tests stuck downloading from sonatype :()

eliserichards:
Once the CI is green this should be ready to land :) my manual tests have all been successful

Issue #15379, opened 2020-09-24
[Bug] Invalid URL shared from page in reader mode


sv-ohorvath:
## Steps to reproduce
1. In Fenix, open a Wikipedia article, turn on Reader mode.
2. Send the tab to your desktop and another mobile device.
3. Open the tab sent.
### Expected behavior
Shares the page URL.
### Actual behavior
The URL is sent in the form of "moz-extension://..." and the desktop browser opens it in a tab but the page can't be loaded.
Fenix doesn't even open it in a tab when you tap the notification.
### Device information

* Android device: Nexus 9 (Android 7.1.1), Pixel 4
* Fenix version: Beta 82.0-beta1, Nightly 9/24

532910:
can confirm

AndiAJ:
Hi, **verified as fixed** on the latest **`Nightly Build 201029`** using the following devices:
• Google Pixel 3a (Android 11)
• Huawei Mate 20 Lite (Android 10)
• OnePlus A3 (Android 6.0.1)

Issue #15413, opened 2020-09-25
[Bug] Invalid URL stored in bookmarks for reader mode


s-ankur:
## Steps to reproduce
Open reader mode for a page
Add it to bookmark

### Expected behavior
The proper page link should be saved. Bookmarks can be synced so it is a very unwelcome surprise that the bookmark does not work on the synced device and instead show some gibberish url
### Actual behavior
The moz extension link is saved
### Device information

* Android device: Asus Zenfone
* Fenix version: Nightly

s-ankur:
Related https://github.com/mozilla-mobile/fenix/issues/15379 https://github.com/mozilla-mobile/fenix/issues/13036
https://github.com/mozilla-mobile/fenix/issues/11699
https://github.com/mozilla-mobile/fenix/issues/5190

AndiAJ:
Hi, **verified as fixed** on the latest **`Nightly Build 201029`** using the following devices:
• Google Pixel 3a (Android 11)
• Huawei Mate 20 Lite (Android 10)
• OnePlus A3 (Android 6.0.1)

PR #16181, opened 2020-10-26
For #15379: Use proper url when sharing.

mcarare:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarMenuController.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarMenuController.kt
@@ -17,6 +17,8 @@ import kotlinx.coroutines.launch
 import mozilla.appservices.places.BookmarkRoot
 import mozilla.components.browser.session.Session
 import mozilla.components.browser.session.SessionManager
+import mozilla.components.browser.state.selector.findTab
+import mozilla.components.browser.state.store.BrowserStore
 import mozilla.components.concept.engine.EngineSession.LoadUrlFlags
 import mozilla.components.concept.engine.prompt.ShareData
 import mozilla.components.feature.session.SessionFeature
@@ -66,7 +68,8 @@ class DefaultBrowserToolbarMenuController(
     private val bookmarkTapped: (Session) -> Unit,
     private val scope: CoroutineScope,
     private val tabCollectionStorage: TabCollectionStorage,
-    private val topSitesStorage: DefaultTopSitesStorage
+    private val topSitesStorage: DefaultTopSitesStorage,
+    private val browserStore: BrowserStore
 ) : BrowserToolbarMenuController {
 
     private val currentSession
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarMenuController.kt
```diff
@@ -295,6 +297,16 @@ class DefaultBrowserToolbarMenuController(
         }
     }
 
+    private fun getProperUrl(currentSession: Session?): String? {
```
csadilek, line 300: Ah we already addressed this in a few places. Best thing to do is get the original / article URL from the store e.g.:
https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/java/org/mozilla/fenix/collections/CollectionCreationStore.kt#L98

This way you don't have to parse the URL which may break in the future.

Here you can get to it via `currentSession?.toTabSessionState()?.readerState` for now until we refactor this class to use the `BrowserStore` instead of `SessionManager`.

csadilek (changes requested):
Thanks for fixing this! There's an easier / safer way to get to the original URL. See comment below.

Let's also add a test for this to `DefaultBrowserToolbarMenuControllerTest` :).

app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarMenuController.kt
```diff
@@ -184,7 +187,7 @@ class DefaultBrowserToolbarMenuController(
                 val directions = NavGraphDirections.actionGlobalShareFragment(
                     data = arrayOf(
                         ShareData(
-                            url = currentSession?.url,
+                            url = getProperUrl(currentSession),
```
csadilek, line 190: nit: I would call this `getShareableUrl` or `ensureUrlShareable`.

app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarMenuController.kt
```diff
@@ -295,6 +298,16 @@ class DefaultBrowserToolbarMenuController(
         }
     }
 
+    private fun getProperUrl(currentSession: Session?): String? {
+        return currentSession?.id?.let {
+            if (browserStore.state.findTab(it)?.readerState?.active == true) {
+                browserStore.state.findTab(it)?.readerState?.activeUrl
```
csadilek, line 304: nit: this isn't super expensive, but I would still only call `findTab` once, by introducing a local `val tab = store.state.findTab(it)`

csadilek (approved):
Nice! Thanks for adding the tests! Just two nits below before landing.

app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarMenuController.kt
mcarare: I was thinking maybe using this method also for another issue, with the bookmarks (https://github.com/mozilla-mobile/fenix/issues/15413). That is why I wanted to leave it more generic.

csadilek: ok :). We you tackle that we could also uplift this function to A-C. It seems it could be useful it many places and we could add an extension function `TabSessionState` for example.

csadilek (approved):

Issue #8169, opened 2020-02-06
[Bug] Upgrade mockwebserver dependency


AaronMT:
mockwebserver ([3.11.0](https://github.com/mozilla-mobile/fenix/blob/9d993cae925faed0ba843d5287df2f47934c330d/buildSrc/src/main/java/Dependencies.kt#L56))(2018) -> [4.3](https://mvnrepository.com/artifact/com.squareup.okhttp3/mockwebserver)?

kbrosnan:
So quick look at this is non-trivial as OkHttp 4.x relies on Okio 2 which creates a circular dependency because we use [Lottie from AirBnB which requires Okio 1](https://github.com/airbnb/lottie-android/blob/9298ee2a1c165761346fce10650eafea50f2dc8d/lottie/build.gradle#L36)

mcomella:
Have we opened a bug against the Lottie library? Maybe if they had external users asking for the upgrade, they'd consider it.

That being said, I looked at the source [and noticed they're intentionally not upgrading](https://github.com/airbnb/lottie-android/blob/708b54b2e3c605cc429906e47015f5a075f54b72/lottie/build.gradle#L37):
```
// Do not ugprade to 2.0 because it will bring in Kotlin as a transitive dependency.
```

kbrosnan:
I never filed one because they explicitly don't want to use Kotlin according to that comment.

AaronMT:
https://github.com/mozilla-mobile/fenix/issues/7435#issuecomment-698503060

AaronMT:
Upgrading to 4.9.0 requires a number of API changes.

https://github.com/mozilla-mobile/fenix/blob/master/app/src/androidTest/java/org/mozilla/fenix/helpers/MockWebServer.kt#L42

```Unresolved reference: setDispatcher```

Every call is affected

```Kotlin
fun setUp() {
        mockWebServer = MockWebServer().apply {
            setDispatcher(AndroidAssetDispatcher())
            start()
        }
    }

PR #16186, opened 2020-10-26
For #8169 - Upgrade Mockwebserver

AaronMT:
Upgrades Mockwebserver to 4.9.0 from 3.11.0 (!)

```setDispatcher``` was removed at some point in the API so all calls need to be updated. There was also an unsafe call warning in a Glean test after upgrading as well as in the dispatcher code itself.


isabelrios (approved):
LGTM, although I'm not very familiar with that mockwebserver code...but if tests are green after being run for a few times, I will go aheas with these changes and monitor that everything keeps working fine.

PR #16217, opened 2020-10-27
No issue: Fix onConfiguationChange crashes in Debug build

csadilek:
In debug builds `resetAfter` adds performance markers on the engine via `components.core.engine`. On child processes this causes additional `GeckoRuntime`s to be created which in turn causes crashes (processes are getting killed).


app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -431,8 +431,17 @@ open class FenixApplication : LocaleAwareApplication(), Provider {
         // https://issuetracker.google.com/issues/143570309#comment3
         applicationContext.resources.configuration.uiMode = config.uiMode
 
-        // random StrictMode onDiskRead violation even when Fenix is not running in the background.
-        components.strictMode.resetAfter(StrictMode.allowThreadDiskReads()) {
+        if (isMainProcess()) {
```
mcomella, line 434: Not necessarily: I reopened https://github.com/mozilla-mobile/android-components/issues/8639#issuecomment-717553059 if you wanted to call this commit "Issue #whatever: ..."

mcomella (approved):
Makes sense to me given my debugging of this issue and what I spoke to Christian about. Thanks for looking into this!

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
csadilek: OK, thanks, I think I will go easy on CI and land this, close the ticket manually :).

Issue #11376, opened 2020-06-09
Reverse the direction that Fenix adds new tabs from


ghost:
• A user on Reddit is asking this feature;

Currently Fenix adds new tabs to the top but chance it so that it adds the new tabs one after the other in a downwards direction like how Chrome does it.

Now if you ask me which one I prefer I don't know. I have been using Fenix since it launched from early last year and I haven't had any problems navigating through tabs. Perhaps a switch to change the direction could be added to satisfy the both types of users and since tab navigation is one of the central features of a browser this should be a customisable behaviour.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-1853)

tukusejssirs:
I wish the new tabs would be added the same way as on the desktop FF:
1. when opening a new tab (empty one, `about:blank`), it would append it to the bottom of the list;
2. when opening a tab with _Open link in new tab_, it should be open right after (below) the tab I am currently on; if multiple links are open this way (without changing the currently open tab), they should be appended to each other (i.e. `current_tab`, `first_new_tab`, `second_new_tab`, `third_new_tab`, …, `nth_new_tab`, following by the other, already open tabs).

The current behaviours drives me nuts as it changes the order I am used to and that makes sense (to me).

liuche:
Copying over a comment from another user issue:

**What is the growth opportunity you want to see solved?**
Avoid long thumb travel across the screen by reversing the tab list when using bottom navigation bar.

**Why?**
The new navigation bar and buttons are at the bottom of the screen. This is to easy navigation with phone getting bigger and avoid long thumb travel across the screen. But when opening the tab menu in Fenix the latest opened tabs are at the top of the screen. Those latest opened tabs have an high probability to be the destination of the user. Thus, reversing the tag lost order will easy access to more recents tabs as the user will already has touch the bottom of the screen to open the tab menu.

Dunexus:
See https://github.com/mozilla-mobile/fenix/issues/12861 and https://github.com/mozilla-mobile/fenix/issues/12108#issuecomment-669048249

apbitner:
@topotropic is currently looking into the way we sort various items within the app

yoasif:
Like the rest of you, I have been using the new direction since it was changed.

[...]

PR #16245, opened 2020-10-29
For #11376  - Reverse tabs list order

codrut-topliceanu:
### Pull Request checklist

- [x] **Screenshots**: This PR includes a screenshot of the changes.

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/tabtray/TabTrayView.kt
+++ b/app/src/main/java/org/mozilla/fenix/tabtray/TabTrayView.kt
@@ -193,13 +193,11 @@ class TabTrayView(
             tabsAdapter.tabTrayInteractor = interactor
             tabsAdapter.onTabsUpdated = {
                 if (view.context.settings().gridTabView) {
+                    concatAdapter.addAdapter(syncedTabsController.adapter)
                     concatAdapter.addAdapter(collectionsButtonAdapter)
-                    concatAdapter.addAdapter(syncedTabsController.adapter)
                 } else {
-                    // Put the 'Add to collections' button after the tabs have loaded.
-                    concatAdapter.addAdapter(0, collectionsButtonAdapter)
-                    // Put the Synced Tabs adapter at the end.
                     concatAdapter.addAdapter(syncedTabsController.adapter)
+                    concatAdapter.addAdapter(collectionsButtonAdapter)
                 }
 
                 if (hasAccessibilityEnabled) {
@@ -442,10 +440,7 @@ class TabTrayView(
 
     private fun setupListTabView() {
         view.tabsTray.apply {
-            layoutManager = LinearLayoutManager(container.context).apply {
-                reverseLayout = true
-                stackFromEnd = true
-            }
+            layoutManager = LinearLayoutManager(container.context)
         }
     }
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/fenix/tabtray/TabTrayView.kt
```diff
@@ -186,13 +186,13 @@ class TabTrayView(
...
                     concatAdapter.addAdapter(syncedTabsController.adapter)
+                    concatAdapter.addAdapter(collectionsButtonAdapter)
                 } else {
-                    // Put the 'Add to collections' button after the tabs have loaded.
-                    concatAdapter.addAdapter(0, collectionsButtonAdapter)
-                    // Put the Synced Tabs adapter at the end.
+                    // Synced Tabs adapter
                     concatAdapter.addAdapter(syncedTabsController.adapter)
+                    // Put the 'Add to collections' button after the tabs have loaded.
+                    concatAdapter.addAdapter(collectionsButtonAdapter)
```
gabrielluong, line 195:
```suggestion
                    concatAdapter.addAdapter(syncedTabsController.adapter)
                    concatAdapter.addAdapter(collectionsButtonAdapter)
```
I think we might as well remove the comments here since it is self explanatory.

gabrielluong (approved):

gabrielluong:
Pinging @topotropic to inform her of the changes, but since she's away this week. I think we can land it for now and wait for a retroactive ux feedback.

topotropic (changes requested):
Thanks for the PR! 

I noticed that when you open a bunch of tabs and go back to the tab tray the tab you're currently on is not always in view. It should follow the same logic as described here https://github.com/mozilla-mobile/fenix/issues/14619 and be in view.

gabrielluong:
> Thanks for the PR!
> 
> I noticed that when you open a bunch of tabs and go back to the tab tray the tab you're currently on is not always in view. It should follow the same logic as described here #14619 and be in view.

@codrut-topliceanu I removed "pr:needs-landing" so we can investigate this.

app/src/main/java/org/mozilla/fenix/tabtray/TabTrayView.kt
```diff
@@ -432,8 +430,8 @@ class TabTrayView(
     private fun setupListTabView() {
         view.tabsTray.apply {
             layoutManager = LinearLayoutManager(container.context).apply {
-                reverseLayout = true
-                stackFromEnd = true
+                reverseLayout = false
```
Mugurell, line 433: I think `false` is the default value for these properties so we don't really need these?
For gridView we would have the same behavior and for it we are not explicit about this behavior.
If there aren't any other specific changes for listView maybe we don't need this method at all?

Mugurell:
I suspect the issue Nicole talked about is because of the offset were adding when scrolling?
https://github.com/mozilla-mobile/fenix/blob/396065816fff8846d45cac9e486fd059436c1942/app/src/main/java/org/mozilla/fenix/tabtray/TabTrayView.kt#L694
This was needed only when the items were arranged in reversed order but if with these changes the order will be the default top->down we won't need the offset anymore.

app/src/main/java/org/mozilla/fenix/tabtray/TabTrayView.kt
```diff
@@ -702,7 +700,7 @@ class TabTrayView(
             layoutManager?.scrollToPosition(recyclerViewIndex)
             smoothScrollBy(
                 0,
-                - resources.getDimensionPixelSize(R.dimen.tab_tray_tab_item_height) / 2
```
gabrielluong, line 705: I don't think we want to remove this / 2 either because that's removing the offset of showing a bit of the tab item above.

gabrielluong (changes requested):

topotropic (approved):
LGTM! Thanks

app/src/main/java/org/mozilla/fenix/tabtray/TabTrayView.kt
codrut-topliceanu: You're right. 👍

codrut-topliceanu:
> I suspect the issue Nicole talked about is because of the offset we're adding when programatically scrolling to the selecting item?
> https://github.com/mozilla-mobile/fenix/blob/396065816fff8846d45cac9e486fd059436c1942/app/src/main/java/org/mozilla/fenix/tabtray/TabTrayView.kt#L694
> 
> 
> This was needed only when the items were arranged in reversed order but if with these changes the order will be the default top->down we won't need the offset anymore.

You were right @Mugurell , thanks! 👍

Mugurell (approved):
LGTM. Thanks!
Let's wait also for Gabriel's review.

gabrielluong (approved):

PR #16101, opened 2020-10-21
For #15707: enable lint test CI; write a few lint tests

mcomella:
@JohanLorenzo Can you help review the changes to the taskcluster files? I want the tests to run from the custom lint rules Gradle modules.

### Pull Request checklist

- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features. In addition, it includes a screenshot of a successful [accessibility scan](https://play.google.com/store/apps/details?id=com.google.android.apps.accessibility.auditor&hl=en_US) to ensure no new defects are added to the product.

### To download an APK when reviewing a PR:
1. click on Show All Checks,
2. click Details next to "Taskcluster (pull_request)" after it appears and then finishes with a green checkmark,
3. click on the "Fenix - assemble" task, then click "Run Artifacts".
4. the APK links should be on the left side of the screen, named for each CPU architecture

Issue #15279, opened 2020-09-21
Avoid unnecessary initialization in nested lazy initializers


mcomella:
MarcLeclair found a pattern that prevents kotlin's lazy initialization pattern from working as expected in nested lazy initializers when combined with our serviceLocator system:
```kotlin
    val useCases by lazy {
        UseCases(
            search.searchEngineManager,
        )
    }
```

via [code sample](https://searchfox.org/mozilla-mobile/rev/7626be04d759ad7f3e3c050f8b61c524fc1d4fda/fenix/app/src/main/java/org/mozilla/fenix/components/Components.kt#55)

If `useCases` is referenced and thus initialized, `search.searchEngineManager` will unexpectedly be initialized because it's referenced to pass it into the `useCases` constructor, even if it's not actually used in the initializer. In practice, we've identified improvements of up to 60-80ms in just one case where this occurred (in `useCases`).

We should try to identify a generic solution to prevent this anti-pattern for good.

### Initial solution ideas
- we can wrap `Lazy`, or implement our own, to intelligently pass getters rather than the actual object
- maybe we can adapt this one-off solution that ac has used in the past ([definition](https://searchfox.org/mozilla-mobile/rev/7626be04d759ad7f3e3c050f8b61c524fc1d4fda/fenix/app/src/main/java/org/mozilla/fenix/components/Core.kt#250), [usage](https://searchfox.org/mozilla-mobile/rev/7626be04d759ad7f3e3c050f8b61c524fc1d4fda/fenix/app/src/main/java/org/mozilla/fenix/components/Components.kt#40))
- as a last ditch effort, we could implement static analysis to try to prevent this from occurring but it'll add a layer of complexity possibly unlike other solutions

mcomella:
Baron-Severin and I had an interesting discussion: with nested lazy initialization, we can later get a performance issue where, say we're moving through a fragment transaction and trigger a cascading initialization of many components, causing us to drop frames. Given this can be triggered by random code paths, we may not notice this in local development but our users may notice it.

This is no reason not to fix this bug – being in an indeliberate half lazy state is just as bad if not worse – but we should consider a solution for this in the future – I filed https://jira.mozilla.com/browse/FXP-1358 to consider implementing this in the future.

mcomella:
[...]

PR #16292, opened 2020-10-31
For #15279: add lazyMonitored to count # of components init

mcomella:
This PR relies on the changes from https://github.com/mozilla-mobile/fenix/pull/16101 to make the detekt tests run.

See https://github.com/mozilla-mobile/fenix/issues/15279#issuecomment-719853480 for motivations. This PR will implement lazyMonitored to count the number of components init and adds a lint rule to ensure new additions use it. ~~I would have liked to also add the automated test to count the number of usages but that would be best to land after https://github.com/mozilla-mobile/fenix/pull/16101.~~ Added to the PR

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/perf/LazyMonitored.kt
+++ b/app/src/main/java/org/mozilla/fenix/perf/LazyMonitored.kt
@@ -1,0 +1,44 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.perf
+
+import mozilla.components.support.base.log.logger.Logger
+import java.util.concurrent.atomic.AtomicInteger
+
+private val logger = Logger("LazyMonitored")
+
+/**
+ * A container for the number of components initialized.
+ */
+object ComponentInitCount {
+    val count = AtomicInteger(0)
+}
+
+/**
+ * A convenience function for setting the [LazyMonitored] property delegate, which wraps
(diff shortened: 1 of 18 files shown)


app/src/main/java/org/mozilla/fenix/perf/LazyMonitored.kt
```diff
@@ -0,0 +1,44 @@
...
+ * [lazy] to add performance monitoring.
+ */
+fun <T> lazyMonitored(initializer: () -> T) = LazyMonitored(initializer)
+
+/**
+ * A wrapper around the [lazy] property delegate to monitor for performance related issues.
+ * For example, we can count the number of components initialized to see how the number of
+ * components initialized on start up impacts start up time.
+ */
+class LazyMonitored<T>(initializer: () -> T) {
```
MarcLeclair: Could we put this private, and just say `lazyMonitored` returns a `Lazy<T>` type? or is this not feasible?

MarcLeclair (approved):
Only had one question, which would be a nit more than anything. Otherwise, it looks good and detected my lazy when I tried it locally. I also checked to see if there was any overhead and perf didn't change.

app/src/main/java/org/mozilla/fenix/perf/LazyMonitored.kt
mcomella: Done. It wasn't possible before because you can't expose a private class through a non-private function but now I can hide it behind `Lazy` instead.

mcomella:
Rebasing...

mcomella:
Got the memory leak error: [task 2020-11-04T23:38:27.025Z] java.lang.OutOfMemoryError: GC overhead limit exceeded I'll restart

Issue #6313, opened 2019-10-29
Set initial loading content page color to non-white when using dark mode


cpeterso:
### Why/User Benefit/User Problem

Currently opening a new page temporarily shows a pure white page before loading actual content regardless of the browser theme.

### What/Requirements

Fenix does not give a way to control this while on Windows this can be controlled by `browser.in-content.dark-mode` pref (which is on by default on Nightly 71 and makes the page gray).

The white flash is the page clear color, which Fenix can already set clear color using GV API `CompositorController.setClearColor`.

Fenix clear color should probably change the clear color depending on Fenix's or Android's dark mode settings.

This bug was originally filed in Bugzilla:
https://bugzilla.mozilla.org/show_bug.cgi?id=1589576

### Acceptance Criteria (how do I know when I’m done?)

Opening a new page when dark mode is enabled does not flash a pure white page before loading actual content.

nt1m:
@cpeterso `browser.in-content.dark-mode` is enabled by default even on Release since Firefox 70. See: https://bugzilla.mozilla.org/show_bug.cgi?id=1565051

I think Fenix should just follow the theme that's chosen by the user though.

ekager:
From Bugzilla:
> The white flash is the page clear color, which Fenix can already set clear color using GV API CompositorController.setClearColor.

ekager:
Chatted with the GV team on Riot and unfortunately the above solution likely does not work anymore with WR enabled. FxR uses a dummy surface with the correct color until they get a `onFirstContentfulPaint` and that's probably the direction we'd have to go as well to fix this.

dholbert:
Here's a screencast of what this looks like, FWIW:
https://www.dropbox.com/s/zne4d8tdfwybtnj/flash-of-white.mp4?dl=0

There's a flash of white (due to this bug) at around 0:04, while the keyboard is disappearing.

532910:
There was a gif in #10158

ekager:
Opened AC issue https://github.com/mozilla-mobile/android-components/issues/6757

Kreuger:
@ekager Hey is therr any update on this? I see you submitted 2 commits for this. Does that mean it's being improved?

ekager:
@Kreuger yeah I have a work in progress that will hide the web content until the content appears so it should improve the behavior a bit. Unfortunately on a site that doesn't support dark mode and is very white (ie CNN) it'll still be jarring when the content appears

[...]

PR #16310, opened 2020-11-01
For #6313 - Remove unneeded delay in BrowserAnimator

ekager:
With https://github.com/mozilla-mobile/android-components/pull/8797 landed, the short white flash should now be gone, so we can remove the remaining animation delay (that was a hack to avoid seeing this)

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/browser/BrowserAnimator.kt
+++ b/app/src/main/java/org/mozilla/fenix/browser/BrowserAnimator.kt
@@ -12,7 +12,6 @@ import androidx.core.graphics.drawable.toDrawable
 import androidx.fragment.app.Fragment
 import androidx.lifecycle.LifecycleCoroutineScope
 import androidx.navigation.NavOptions
-import kotlinx.coroutines.delay
 import kotlinx.coroutines.launch
 import mozilla.components.concept.engine.EngineView
 import org.mozilla.fenix.R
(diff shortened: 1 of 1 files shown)


gabrielluong (approved):

app/src/main/java/org/mozilla/fenix/browser/BrowserAnimator.kt
```diff
@@ -38,11 +37,8 @@ class BrowserAnimator(
         get() = swipeRefresh.get()
 
     fun beginAnimateInIfNecessary() {
-        viewLifecycleScope.get()?.launch {
-            delay(ANIMATION_DELAY)
-            unwrappedEngineView?.asView()?.visibility = View.VISIBLE
-            unwrappedSwipeRefresh?.background = null
-        }
+        unwrappedEngineView?.asView()?.visibility = View.VISIBLE
```
pocmo, line 40: When `coverUntilFirstPaint` was introduced I asked Agi about those visibility changes and I think he mentioned that we can get rid of them now too. Maybe we could test this and remove this in a follow-up in Fenix and AC? :)

ekager: I will investigate! In Fenix I know we also toggle the engine view visiblity when we `captureEngineViewAndDrawStatically` when navigating to Settings for example and then return to the browser.

PR #16314, opened 2020-11-02
Closes #16256: Introduce build flag for "Mozilla Online" variant.

### LINKED ISSUE #16256: Introduce a build flag for Mozilla Online
Introduce a build flag for Mozilla Online. This can be used in code to conditionally make decisions specifically for Mozilla Online builds.

If a flag is not adequate for the Mozilla Online needs then later on we can add variants. 

Let's name the flag 'Mozilla Online'. Or if it a key/value pair, something like Distribution: Mozilla Online? (I don't know how that translates to gradle & code . But I am thinking in the future we may have more distribution specific changes.)

pocmo:
This patch introduces a build flag that can be used on demand on the command line (`./gradlew -PmozillaOnline [..}`) or
permanently when added to local.properties (`mozillaOnline`).

At runtime `Config.channel.isMozillaOnline` will return `true` if the flag was set at build time.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -36,6 +36,13 @@ android {
         manifestPlaceholders = [
                 "deepLinkScheme": deepLinkSchemeValue
         ]
+
+        // Build flag for "Mozilla Online" variants. See `Config.isMozillaOnline`.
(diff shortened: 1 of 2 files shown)


pocmo:
CC @st3fan

Amejia481 (approved):

PR #16316, opened 2020-11-02
For #13446 UI tests: wait for historyList to exist

sv-ohorvath:
Tests in #13446 are failing because there might be some UI covering the app and is not visible in the Firebase recordings. Added an extra wait to see if this would permit the recordings to capture that when the tests fail.


mcarare (approved):
LGTM!

PR #16325, opened 2020-11-02
Bug 1670284 - Enable webrender for all tests and run a subset without webrender.

gmierz:
This patch flips our default settings for tests to use webrender. It also adds a subset of tests that were identified as "high-value" to run without webrender until we are fully switched to use webrender.

### DIFF
--- a/taskcluster/fenix_taskgraph/transforms/browsertime.py
+++ b/taskcluster/fenix_taskgraph/transforms/browsertime.py
@@ -119,6 +119,7 @@ def enable_webrender(config, tasks):
         task["run"]["command"].append("--enable-webrender")
         task["name"] += "-wr"
         task["description"] += "-wr"
+        task['extra']['treeherder']["groupSymbol"] += "-wr"
         yield task
 
 
--- a/taskcluster/fenix_taskgraph/transforms/browsertime.py
+++ b/taskcluster/fenix_taskgraph/transforms/browsertime.py
@@ -119,8 +119,15 @@ def enable_webrender(config, tasks):
         task["run"]["command"].append("--enable-webrender")
         task["name"] += "-wr"
         task["description"] += "-wr"
+
+        # Setup group symbol
         group, sym = task["treeherder"]["symbol"].split("(")
         task["treeherder"]["symbol"] = "{}-wr({})".format(group, sym[:-1])
+
+        # Setup the platform name
+        platform, plttype = task["treeherder"]["platform"].split("/")
+        task["treeherder"]["platform"] = "{}-qr/{}".format(platform, plttype)
+
         yield task
 
 
--- a/taskcluster/fenix_taskgraph/transforms/browsertime.py
+++ b/taskcluster/fenix_taskgraph/transforms/browsertime.py
@@ -119,7 +119,9 @@ def enable_webrender(config, tasks):
         task["run"]["command"].append("--enable-webrender")
         task["name"] += "-wr"
         task["description"] += "-wr"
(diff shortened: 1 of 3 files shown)


gmierz:
@JohanLorenzo could you review this patch when you get a chance? You can find the browsertime tests here: https://github.com/mozilla-mobile/fenix/runs/1347747374

The failures are intermittents and not related to this patch.

taskcluster/fenix_taskgraph/transforms/browsertime.py
```diff
@@ -110,6 +110,23 @@ def build_browsertime_task(config, tasks):
         yield task
 
 
+@transforms.add
+def enable_webrender(config, tasks):
+    for task in tasks:
+        if not task.pop("web-render-only", False):
+            newtask = copy.deepcopy(task)
+            yield newtask
+        task["run"]["command"].append("--enable-webrender")
```
MihaiTabara, line 119: Hm, quick question: I wonder if we don't need an `else` statement here? I might be wrong, but wouldn't this mean that we're going to `yield` both the task (which has the `web-render-only` config set to False`) but also its equivalent with `-wr` version on it? 

If I get this right, we want to `webrender-ify` all the tasks by default, except the ones that are specifically turned off for that, no?

gmierz: In this code, if we have the field `web-render-only` set to True, then only webrender tasks will be created (it won't run the code in the if), otherwise if `web-render-only` is False, then we'll create non-webrender and webrender tasks for that test.

gmierz: You can check how it works by looking at the tasks that were scheduled in this test run: https://github.com/mozilla-mobile/fenix/runs/1347747374

MihaiTabara: Hm, makes sense. My question lay ground in noticing the tasks that change upon your PR. I ran taskgraph diff against all possible Fenix graphs and releases to see what changes. Results are [here](https://gist.github.com/MihaiTabara/bb52bda5eb3f93e98c2398b9f549be31). I was particularly worried about the tasks that seem to be deleted and amended into `-wr` in [here](https://gist.github.com/MihaiTabara/bb52bda5eb3f93e98c2398b9f549be31#file-taskgraph-diff-L445) but it also might be the `git diff` diff misunderstanding.

gmierz: I don't see anything concerning in those results. The task you pointed at is here now: https://gist.github.com/MihaiTabara/bb52bda5eb3f93e98c2398b9f549be31#file-taskgraph-diff-L708

git diff isn't good for looking at task definition changes. I have a better tool here that we use for complex cases: https://github.com/gmierz/moz-current-tests/blob/master/generate_test_report.py

MihaiTabara: Sounds good, thanks for pointing at that. Patch LGTM, happy to r+ this or defer to @JohanLorenzo for another set of eyes, up to you.

JohanLorenzo (approved):
These changes look good to me from a taskgraph point of view! Thank you @MihaiTabara for looking at it first 🙂

taskcluster/fenix_taskgraph/transforms/browsertime.py
gmierz: Either you or @JohanLorenzo would be fine, thanks for taking a look! :)

gmierz:
Thanks for the review @JohanLorenzo! All ready for merging now.

PR #16267, opened 2020-10-29
No issue: Make sure we are cleaning up test coroutines

ekager:
### Pull Request checklist

- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features. In addition, it includes a screenshot of a successful [accessibility scan](https://play.google.com/store/apps/details?id=com.google.android.apps.accessibility.auditor&hl=en_US) to ensure no new defects are added to the product.

### To download an APK when reviewing a PR:
1. click on Show All Checks,
2. click Details next to "Taskcluster (pull_request)" after it appears and then finishes with a green checkmark,
3. click on the "Fenix - assemble" task, then click "Run Artifacts".
4. the APK links should be on the left side of the screen, named for each CPU architecture

PR #16349, opened 2020-11-03
Fixes #16348 - Increase Gradle daemon memory size

### LINKED ISSUE #16348: Increase Gradle daemon memory size
Recently we have been running out of memory during test tasks.

Let's try to increase the Gradle memory size in `gradle.properties` a bit to see if these problems go away. We run on 16GB instances so we have quite a bit of room. To keep this laptop friendly, the suggestion is to increase to 6GB for now.

> `org.gradle.jvmargs=-Xmx6144m`

We may be running out of memory because we have more tests or more complicated tests, but it is also entirely possible that we have a legit memory leak in our tests or the surrounding test infrastructure.

See also

* https://github.com/mozilla-mobile/fenix/issues/15981
* https://github.com/mozilla-mobile/fenix/issues/16237

Also uplift to the Release and Beta branches.

st3fan:
This patch increases Gradle's memory to 6GB (from 4GB) to hopefully avoid OutOfMemory exceptions during tests.

If this PR goes through properly then also uplift this to the active release and beta branch.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -6,7 +6,7 @@
 # http://www.gradle.org/docs/current/userguide/build_environment.html
 # Specifies the JVM arguments used for the daemon process.
 # The setting is particularly useful for tweaking memory settings.
-org.gradle.jvmargs=-Xmx4096m
+org.gradle.jvmargs=-Xmx6144m
 # When configured, Gradle will run in incubating parallel mode.
 # This option should only be used with decoupled projects. More details, visit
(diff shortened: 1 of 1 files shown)


st3fan:
@ekager just found out that you also have https://github.com/mozilla-mobile/fenix/pull/16267 - I'll uplift both? (I think 4GB is a bit low anyway.)

Amejia481 (approved):

Issue #13427, opened 2020-08-09
FNX-14608 ⁃ [Bug ]Collections: UI shows selected tab but isn't selected


B0pol:
## Steps to reproduce
1. Open few tabs
2. go to home (+ button)
3. Click on save as collection (screenshot 1)
### Expected behavior
No tab is selected, or all tabs or selected?
### Actual behavior
One tab is shown as selected but isn't really selected as you can't save it (screenshot 2)

You need to click again to select it (screenshot 3)
### Device information

* Android device: Android 9, Pocophone F1
* Fenix version: Nightly 200809 06:01

### Screenshots
1. 
<img src="https://user-images.githubusercontent.com/58657617/89733361-4c930400-da55-11ea-9a62-edcd40d5f3f7.png" width="300">
2. 
<img src="https://user-images.githubusercontent.com/58657617/89733368-5583d580-da55-11ea-80e6-de3cc22c3dfc.png" width="300">

3. 
<img src="https://user-images.githubusercontent.com/58657617/89733378-69c7d280-da55-11ea-873e-0ff6f971ca9d.png" width="300">

yoasif:
The STR could be simplified as:

1. load 1 or more pages in tabs
2. open a new tab (hold on tab switcher button and select new tab)
3. tab tab switcher icon

What happens:

A tab shows a "selected" highlight (which seems to be what Fenix shows in the tab tray when *within* a tab to show which tab will be shown if a user taps back or into the content area behind the tab tray)

Expected result:

Nothing should show as highlighted because no tab is open (or make it so tabs are created that appear as separate tabs in the tab tray [this is preferable and matches other browsers, including Fennec and Chrome).

liuche:
Since we just added this feature, @topotropic do you have any thoughts on the initial state when opening tab collections multi-select?

yoasif:
@liuche @topotropic this doesn't just apply with multi-select - the highlight doesn't seem to be related to a selection.

topotropic:
The highlight indicates which tab was last visited;

As for multi-select, I agree that it can be confusing to have a highlighted tab in there. If it's feasible we should not highlight anything in multi-select mode. Thanks!

Mugurell:
Waiting on https://github.com/mozilla-mobile/fenix/issues/15968 to have a common solution to both the list based and the grid based tab trays layouts.

softvision-miralobontiu:
[...]

PR #16370, opened 2020-11-04
For #13427 - Disable the "selected tab" decoration when in Multiselect

Mugurell:
Selecting a tab while in Multiselect would add a different decoration to the one
already set for the last tab opened and this would confuse users.
Let's avoid this.

After the changes here which are based on https://github.com/mozilla-mobile/android-components/pull/8894:
selectAndCheck
[video](https://drive.google.com/file/d/1HAAG-Bn1qpbOlp0vHzH2QjZvnLzN2g4n/view?usp=sharing)

### Pull Request checklist

- [x] **Accessibility**: The code in this PR does not include any a11y user facing features.

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/tabtray/FenixTabsAdapter.kt
+++ b/app/src/main/java/org/mozilla/fenix/tabtray/FenixTabsAdapter.kt
@@ -59,11 +59,20 @@ class FenixTabsAdapter(
     override fun onBindViewHolder(
         holder: TabViewHolder,
         position: Int,
-        payloads: MutableList<Any>
+        payloads: List<Any>
     ) {
         if (payloads.isNullOrEmpty()) {
             onBindViewHolder(holder, position)
             return
+        }
+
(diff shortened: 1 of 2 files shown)


codrut-topliceanu (changes requested):
I think we need to make the same check regarding the decoration when the view is created. At the moment if the "current-tab" is not in view when we switch to multiselect then bringing the "current-tab" into view will apply the original decoration:

multi select

codrut-topliceanu (changes requested):
Gridview works great. 👍 
But there is still a bit of an issue with the listview mode.

listview_decoration

codrut-topliceanu (approved):
Scratch that, there was an issue with AC on my end.

Work great. 👍

csadilek:
We will land as part of https://github.com/mozilla-mobile/fenix/pull/16921 as we need an A-C upgrade with multiple changes.

Issue #15934, opened 2020-10-15
[Crash] ConcurrentModificationException org.mozilla.fenix.wifi.WifiConnectionMonitor


eliserichards:
Followup to #14680 

The main `ObserverRegistry` crash was resolved (see [Sebastian's comment](https://github.com/mozilla-mobile/fenix/issues/14680#issuecomment-688686650)), but there is another related crash that we are seeing:

```
java.util.ConcurrentModificationException: null
    at java.util.LinkedHashMap$LinkedHashIterator.nextNode(LinkedHashMap.java:757)
    at java.util.LinkedHashMap$LinkedKeyIterator.next(LinkedHashMap.java:780)
    at org.mozilla.fenix.wifi.WifiConnectionMonitor$frameworkListener$1.onAvailable(WifiConnectionMonitor.kt:2)
    at android.net.ConnectivityManager$NetworkCallback.onAvailable(ConnectivityManager.java:3322)
    at android.net.ConnectivityManager$CallbackHandler.handleMessage(ConnectivityManager.java:3607)
    at android.os.Handler.dispatchMessage(Handler.java:106)
    at android.os.Looper.loop(Looper.java:223)
    at android.os.HandlerThread.run(HandlerThread.java:67)
```

Sentry: https://sentry.prod.mozaws.net/share/issue/4234e7770c9f4d05abd2c60f63aafb3c/

AndiAJ:
Managed to reproduce it on **`82.0.0-beta.6`** using a OnePlus A3 (Android 6.0.1) on https://www.walkinto.in/tour/WyKhAFf0iXZJgF2CKzAj7

8346bcbb-5448-409b-97bd-f275bb46894e
java.util.ConcurrentModificationException
* Sentry: https://sentry.prod.mozaws.net/operations/fenix-fennec-beta/?query=bba951eef99b4260a26a1e7cce002b7a
* Sentry: https://sentry.prod.mozaws.net/operations/fenix-fennec-beta/?query=04288cd1909148ef9b94c76b3da4043d
* Socorro: https://crash-stats.mozilla.org/report/index/bp-cddcf565-5d1b-4795-b9b0-9bc800201016
* Socorro: https://crash-stats.mozilla.org/report/index/bp-0b955d98-a872-4cfa-bd48-2e1900201016
----
java.util.ConcurrentModificationException
at java.util.LinkedHashMap$LinkedHashIterator.nextEntry(LinkedHashMap.java:346)
[...]

PR #16372, opened 2020-11-04
For issue #15934 fix ConcurrentModificationException on WifiConnectionMonitor

Amejia481:
It looks like we were [iterating through](https://docs.oracle.com/javase/7/docs/api/java/util/ConcurrentModificationException.html) the `callbacks` list while it was modified by another thread, creating defensive copy while looping should fix the issue, I tested but I don't have the exact STR, but while testing I wasn't able to reproduce. Creating test for this edge case it's a bit tricky, for this reason I didn't include it.

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/wifi/WifiConnectionMonitor.kt
+++ b/app/src/main/java/org/mozilla/fenix/wifi/WifiConnectionMonitor.kt
@@ -26,7 +26,7 @@ import android.net.NetworkRequest
  * ```
  */
 class WifiConnectionMonitor(app: Application) {
-    private val callbacks = mutableSetOf<(Boolean) -> Unit>()
+    private val callbacks = mutableListOf<(Boolean) -> Unit>()
     private val connectivityManager = app.getSystemService(Context.CONNECTIVITY_SERVICE) as
             ConnectivityManager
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/fenix/wifi/WifiConnectionMonitor.kt
```diff
@@ -35,16 +35,21 @@ class WifiConnectionMonitor(app: Application) {
...
 
         override fun onAvailable(network: Network?) {
-            callbacks.forEach { it(true) }
+            notifyListeners(true)
             lastKnownStateWasAvailable = true
         }
     }
 
+    internal fun notifyListeners(value: Boolean) {
+        val items = ArrayList(callbacks)
```
eliserichards, line 49: Can we instantiate this (line 29) as an ArrayList? Or are you doing this here to make it non-mutable?

Amejia481: [MutableList](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-mutable-list/) is a more abstract interface, and could be swapped by other implementations, under the hood is backed by a `ArrayList` when we call `mutableListOf` :)

image

Here we want to create a defensive copy of the initial list, to iterate safely while the original list can be modified.

eliserichards: Excellent! Thank you :D

eliserichards (approved):
Once the CI passes this is good to go! 🚀

PR #16404, opened 2020-11-05
Closes #16327: Refactor MenuPresenter to use browser store

### LINKED ISSUE #16327: Refactor MenuPresenter to observe session/selection via store
`MenuPresenter` is currently using a `SessionManager.Observer` to detect when a session was selected and changes. This can be done using the store instead.

We want to get rid of SessionManager observers so we can finish the A-C refactoring to browser state.

csadilek:
Replacing another `SessionManager` observer to use the store instead, + tests.


app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
```diff
@@ -1207,8 +1207,6 @@ abstract class BaseBrowserFragment : Fragment(), UserInteractionHandler,
         private const val REQUEST_CODE_DOWNLOAD_PERMISSIONS = 1
         private const val REQUEST_CODE_PROMPT_PERMISSIONS = 2
         private const val REQUEST_CODE_APP_PERMISSIONS = 3
-
-        private const val LOADING_PROGRESS_COMPLETE = 100
```
csadilek, line 1211: This wasn't used anymore.

pocmo (approved):
👍 

Nice. That looks pretty straight-forward.

PR #16419, opened 2020-11-06
For #15847: use Espresso's RecyclerView scrollTo action

sv-ohorvath:
Fixes and re-enables #15847.
Ran the solution 30 times on Firebase, all passed: https://github.com/mozilla-mobile/fenix/runs/1362989784, compared with the runs before the change: https://github.com/mozilla-mobile/fenix/runs/1347485985


isabelrios (approved):
Changes LGTM! looks like test passed and I'm sure it has run several times to be sure it is not flaky anymore. Thanks!

PR #16431, opened 2020-11-06
Closes #16430: ExternalAppBrowserActivity should not handle incoming intents

### LINKED ISSUE #16430: ExternalAppBrowserActivity should not handle incoming intents in onNewIntent().
In `HomeActivity.onNewIntent()` we handle various `Intents` (coming from the widget, notifications etc.). `ExternalAppBrowserActivity` inherits from `HomeActivity` and will therefore handle the same `Intent`s. This seems like we do not want. For example launching a PWA from the home screen, when it is already running, will bring it to the front and invoke `onNewIntent()` which will do all kinds of things that do not make sense in `ExternalAppBrowserActivity` (like launching into "the browser").

We probably never want to handle `onNewIntent()` in `ExternalAppBrowserActivity` and for now should just do nothing and let the activity come to the front.

csadilek:
Just introducing `handleNewIntent` and making sure it's a NOOP for `ExternalAppBrowserActivity`.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
+++ b/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
@@ -397,8 +397,12 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
      */
     final override fun onNewIntent(intent: Intent?) {
         super.onNewIntent(intent)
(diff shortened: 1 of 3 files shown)


pocmo (approved):
👍

PR #16492, opened 2020-11-11
[l10n] Uplift strings

st3fan:
This patch imports strings. I was not entirely sure about the `l10n-uplift.py` script, so I ran

```
$ ./l10n-uplift.py releases/v83.0.0 --verbose
...
Uplifting (dry-run)...
Cherry picking 5739140ff31b9513fa9072e96d77f41e41a35287 from 'master' to 'releases/v83.0.0'
Cherry picking bf69519dc96dfadd05940a0518d56c3c7e835243 from 'master' to 'releases/v83.0.0'
Cherry picking e61e5b80dd6fb76d06e8e106803ca4a62d4c06be from 'master' to 'releases/v83.0.0'
Cherry picking 1ad5ea58a2931f47c463c4635621ca7c37b061e9 from 'master' to 'releases/v83.0.0'
```

And then manually `git cherry-pick` those commits into this PR branch.

The resulting changes look syntactically good to me, but I don't know enough about the strings process to understand if these are the _correct_ changes. cc @Delphine


st3fan:
Adding @Mugurell as a reviewer - I think you did this last time right? Can you confirm this looks good?

Delphine:
I feel like this looks good, but tbh I don't do reviews on this usually. So would rather someone else confirm as well :)

pocmo:
> And then manually git cherry-pick those commits into this PR branch.

I think you want to let the script do that (`--uplift`) since it leaves a hint about the cherry-pick in the commit message (`cherry-pick -x`), so that the next time it runs it doesn't import those again.

Mugurell:
> > And then manually git cherry-pick those commits into this PR branch.
> 
> I think you want to let the script do that (`--uplift`) since it leaves a hint about the cherry-pick in the commit message (`cherry-pick -x`), so that the next time it runs it doesn't import those again.

Indeed, that why I've did also.
The script should do the cherry-picking. We should just run it.
If there are no conflicts, nothing else to do.

Mugurell:
The commits seem ok, other than they missing the cherry-pick comments.
Some UI tests are failing seemingly not related to these changes.

st3fan:
Closing this PR - will re-run with the proper `cherry-pick` commands. (The script fails on a trivial conflict but it will not complete the import on any kind of failure - so we can't use that until we have improved it.)

PR #16504, opened 2020-11-11
String import to releases/v83 (2020-11-11)

st3fan:
This is the final string import for `releases/v83`. Because the script failed, I ran the following:

```
git cherry-pick 5739140ff31b9513fa9072e96d77f41e41a35287 -x
git cherry-pick bf69519dc96dfadd05940a0518d56c3c7e835243 -x
git cherry-pick e61e5b80dd6fb76d06e8e106803ca4a62d4c06be -x
git cherry-pick 1ad5ea58a2931f47c463c4635621ca7c37b061e9 -x
```


Mugurell (approved):
Manually cherry-picking but with using `cherry-pick -x` got the right commit message based on which on future runs of the uplift script these commits would be recognized as already uplifted.

Mugurell:
`test-debug` has timed out.
Confirmed with Stefan this is safe to be landed as no english strings were changed so there should be no changes in code.

Mugurell:
Seeing here a change for the `sat` locale I asked @Delphine if this should be enabled in Release. Case in which we'll also need to modify https://github.com/mozilla-mobile/fenix/blob/master/l10n-release.toml and enable it there.

st3fan:
@Mugurell I just merged https://github.com/mozilla-mobile/fenix/pull/16490 to include `gl` and `sat` on `master`.

Issue #16477, opened 2020-11-10
[Bug][a11y][Grid View]Tab index of total open tabs not read 


AndiAJ:
## Prerequisites
Talkback should be enabled
Have a couple couple of open tabs

## Steps to reproduce
1. Open the Tabs Tray
2. Tap on the first listed tab
3. Tap on the second listed tab

### Expected behavior
Step 2: 1 of x (number of open tabs)
Step 3: 2 of x (number of open tabs)

### Actual behavior
The tab index of total open tabs is not read

### Device information

* Android device: Google Pixel 3a (Android 11)
* Fenix version: Nightly Build 201110

mcarare:
As a note, also to be taken into consideration for QA when checking this: The expected behavior for a grid view item is that Talkback announces the column and row the item occupies in the grid. 
For example: " Row 1, Column 1" accompanied by "In grid 3 rows, 2 columns" only when selecting a tab after the previously selected element is outside the grid. Also, the column or row number is announced only when the row of the column is changed. For example, after " Row 1, Column 1" if an item in the same row is selected only "Column 2" will be announced.

sflorean:
Verified as fixed on Nightly 11/17 with Samsung Galaxy Note10 (Android 10).

PR #16510, opened 2020-11-11
For #16477: Update a11y info to collection item for both grid and list.

mcarare:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/tabtray/FenixTabsAdapter.kt
+++ b/app/src/main/java/org/mozilla/fenix/tabtray/FenixTabsAdapter.kt
@@ -8,13 +8,15 @@ import android.content.Context
 import android.view.LayoutInflater
 import android.view.View
 import androidx.core.view.isVisible
+import androidx.recyclerview.widget.GridLayoutManager
+import androidx.recyclerview.widget.RecyclerView
 import kotlinx.android.synthetic.main.checkbox_item.view.*
 import kotlinx.android.synthetic.main.tab_tray_item.view.*
 import mozilla.components.browser.tabstray.TabViewHolder
 import mozilla.components.browser.tabstray.TabsAdapter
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/mozilla/fenix/tabtray/FenixTabsAdapter.kt
```diff
@@ -67,10 +67,30 @@ class FenixTabsAdapter(
...
+        val itemIndex = if (isListTabView) {
+            tabCount - position - 1
+        } else {
+            position
+        }
+
+        val rowIndex = if (isListTabView) {
+            itemIndex
+        } else {
+            itemIndex / 2
```
ekager, line 82: Does this have to be 2? Looking below it sounds like it could be more but I might be misunderstanding?

app/src/main/java/org/mozilla/fenix/tabtray/TabTrayView.kt
```diff
@@ -656,21 +656,11 @@ class TabTrayView(
...
-                info?.let {
-                    info.collectionInfo = CollectionInfo.obtain(
-                        tabsAdapter.tabCount,
-                        1,
-                        false
-                    )
-                }
-            }
-        }
+        val isListTabView = view.context.settings().listTabView
```
ekager, line 659: Do you think we could rename this function from `updateTabCounterContentDescription` to `updateViewAccessibility` or something? 
Also - could we possibly only call this function if `hasAccessibilityEnabled`?

ekager (approved):

app/src/main/java/org/mozilla/fenix/tabtray/TabTrayView.kt
mcarare: I think this should also be called with accessibility disabled, in case someone turns it on after it has opened the tabs tray.

Issue #16376, opened 2020-11-04
Test the view hierarchy depth on startup and count the number of ConstraintLayout in RecyclerView children


MarcLeclair:
View hierarchy depth has an impact on performance and we should try to limit the number of levels we have. For startup, this would include traversing the startup views and counting its number of levels to make sure these don't vary to much. 

Time: 3 Days

mcomella:
I opened a PR for both of these https://github.com/mozilla-mobile/fenix/pull/16514

mcomella:
PR merged in https://github.com/mozilla-mobile/fenix/commit/72ab058551244bd0451d56c8c9bbcaa72849758c

PR #16514, opened 2020-11-11
For #16376: assert view hierarchy depth & ConstraintLayout children on start up

mcomella:
For ConstraintLayout count assertion:

    I'm not convinced this is a useful test because the performance
    characteristics seem unfortunately nuanced: if the test fails, it implies
    you shouldn't add more but that isn't necessarily true (for example, if
    your RV child has a large depth). Furthermore, we haven't measured in a
    variety of circumstances that layouts like LinearLayout perform better
    as RecyclerView children so maybe it only held for the one small case we
    used it in.

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -51,7 +51,6 @@
 *StrictMode*kt @mozilla-mobile/Performance
 *ConstraintLayoutPerfDetector* @mozilla-mobile/Performance
 *MozillaRunBlockingCheck.kt @mozilla-mobile/Performance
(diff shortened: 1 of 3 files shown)


MarcLeclair (approved):
This looks good to me. I'm wondering the same as you though, if it's useful to know the RV children.  I do remember that, as a rule of thumb, simpler layouts (especially in RVs) LinearLayout > ConstraintLayout for the first couple of loads until it balanced out.

PR #16520, opened 2020-11-12
Update gradle properties.

pocmo:
The toolchain task is failing on `master`. Here I am trying a couple of things:

* Create a heap dump on "out of memory"
* Use parallel GC (androidx seems to enable that for their repo)
* Enable parallel gradle tasks and kotlin compiler runs
* Force jacoco out of process (which has helped fixing a bunch of problems in AC).


pocmo:
Tested multiple things here. The new PR for landing is #16526.

PR #16480, opened 2020-11-10
Update Android Components version

MickeyMoz:

PR #16524, opened 2020-11-12
Revert "Update Android Components version"

JohanLorenzo:
Reverts mozilla-mobile/fenix#16480

For an unknown reason, the current rebuilds of `toolchain-linux64-android-gradle-dependencies` fail. If I'm not mistaken, reverting this patch enables taskgraph to reuse an existing gradle cache (which is the output of the task `toolchain-linux64-android-gradle-dependencies`). This way, we can run builds and tests while we investigate the root cause. 

[1] https://treeherder.mozilla.org/jobs?repo=fenix&searchStr=gradle-&revision=dbdb68ac72f27b4272759c3cc511506b637d767e&selectedTaskRun=OhrlS1k2SJOTKV-G_h9ULw.0


pocmo (approved):

mstange:
Also, please hold off on updating to a GeckoView build that contains [Bug 1674694](https://bugzilla.mozilla.org/show_bug.cgi?id=1674694), because that fix introduced a [bad crashing bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1676814). The crash starts with the 20201111095506 GeckoView build ID and is not fixed yet.

PR #16528, opened 2020-11-12

MickeyMoz:


pocmo:
I'll rebase this, now that the fix is in that makes the toolchain task more stable.

mstange:
Please hold off until https://bugzilla.mozilla.org/show_bug.cgi?id=1676814 is fixed and makes it into AC, see https://github.com/mozilla-mobile/fenix/pull/16524#issuecomment-726522378 . Otherwise there will be an avalanche of crashes.

ekager:
Got it! Will put a do not land

pocmo:
Looks like a fix landed in mozilla-central on Saturday. AC updated to a version that contains the fix yesterday. So we should be good to go again, I think. Removing the label.

pocmo (approved):

PR #16536, opened 2020-11-12
Closes #16345: Standartizing share icon across all app

### LINKED ISSUE #16345: [Bug] Update share icon in multi-select mode
## Steps to reproduce
1. Go to History or Bookmarks
2. Long press on entry to go into multi-select mode 

### Expected behavior
Share icon looks the same across the app (should look like the one that is in the site menu)

### Actual behavior
Share icon looks different

LeoMaldonado:
### Pull Request checklist

1. click on Show All Checks,

This PR closes #16345. I removed share icon outlined by a filled share icon and put it in all references to share icons.

Screenshot_20201112-165052
Screenshot_20201112-165153
Screenshot_20201112-165258


app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
```diff
@@ -125,15 +125,14 @@ class DefaultToolbarMenu(
...
-            imageResource = R.drawable.mozac_ic_share,
+            imageResource = R.drawable.ic_share_filled,
             contentDescription = context.getString(R.string.browser_menu_share),
             iconTintColorResource = primaryTextColor(),
             listener = {
                 onItemTapped.invoke(ToolbarMenu.Item.Share)
             }
         )
 
-        registerForIsBookmarkedUpdates()
```
ekager, line 136: Why are we removing this?

ekager (changes requested):
Could we just remove the `ic_hollow_share` altogether and use the AC `mozac_ic_share` when needed? The only drawback I can think of is we need to make sure it's tinted correctly where used.

app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
LeoMaldonado: Don't know, putting it back

LeoMaldonado:
> Could we just remove the `ic_hollow_share` altogether and use the AC `mozac_ic_share` when needed? The only drawback I can think of is we need to make sure it's tinted correctly where used.

You're right, but `mozac_ic_share` belongs to a third party lib, and does not follow app theme. I decided this approach because I could use `primaryText` as default color, automatically following any future color changes

ekager:
bors try

ekager (approved):
Okay I'm fine with changing the asset!

ekager:
Merged, thanks for the contribution!

PR #16596, opened 2020-11-16
Integrate new search code from Android Components into Fenix.

### LINKED ISSUE #14455: [Bug] edit search engine - can't change one letter case
## Steps to reproduce
1. Add one search engine. For example, name: "urban dictionary" url: "https://www.urbandictionary.com/define.php?term=%s"
2. Edit the search engine, change name to "Urban dictionary" (U is now uppercase)
### Expected behavior
It works, the search engine is now named Urban dictionary
### Actual behavior
It fails to update name with error:
> Search engine with name "urban dictionary" already exists.
### Device information

* Android device: Pocophone F1
* Fenix version: nightly 20200829

pocmo:
This is ready for review. I would like to land it *after* the release branch was cut though. Now it's a bit too close to let it go to beta immediately. I'm happy to uplift it in the first week though, if it looks good.

Also closes #14455.


pocmo:
I went through the new search test cases on testrail and found two issues that I will debug.

app/src/main/java/org/mozilla/fenix/search/SearchDialogFragment.kt
```diff
@@ -191,6 +188,14 @@ class SearchDialogFragment : AppCompatDialogFragment(), UserInteractionHandler {
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
 
+        consumeFlow(requireComponents.core.store) { flow ->
```
csadilek, line 191: nit: same as above, if we add a `observeSearchChanges` function and pass the store it would be relatively painless to unit tests. That's what I did in `BrowserFragmentTest`, but can also be follow up tickets to improve coverage here.

app/src/main/java/org/mozilla/fenix/search/SearchFragmentStore.kt
```diff
@@ -29,14 +31,21 @@ class SearchFragmentStore(
...
+    }
 
     data class Default(override val searchEngine: SearchEngine) : SearchEngineSource()
     data class Shortcut(override val searchEngine: SearchEngine) : SearchEngineSource()
 }
 
 /**
  * The state for the Search Screen
+ *
+ * TODO: Update kdoc!
```
csadilek, line 47: nit: TODO :)

app/src/main/java/org/mozilla/fenix/search/SearchFragmentStore.kt
```diff
@@ -143,15 +148,26 @@ private fun searchStateReducer(state: SearchFragmentState, action: SearchFragmen
...
-            state.copy(areShortcutsAvailable = action.areShortcutsAvailable)
         is SearchFragmentAction.UpdateQuery ->
             state.copy(query = action.query)
-        is SearchFragmentAction.SelectNewDefaultSearchEngine ->
-            state.copy(searchEngineSource = SearchEngineSource.Default(action.engine))
         is SearchFragmentAction.AllowSearchSuggestionsInPrivateModePrompt ->
             state.copy(showSearchSuggestionsHint = action.show)
         is SearchFragmentAction.SetShowSearchSuggestions ->
             state.copy(showSearchSuggestions = action.show)
+        is SearchFragmentAction.UpdateSearchState -> {
```
csadilek, line 157: Can we cover this in `SearchFragmentStoreTest`?

app/src/main/java/org/mozilla/fenix/settings/search/AddSearchEngineFragment.kt
```diff
@@ -159,22 +167,14 @@ class AddSearchEngineFragment : Fragment(R.layout.fragment_add_search_engine),
                         .getString(R.string.search_add_custom_engine_error_cannot_reach, name)
                 }
                 SearchStringValidator.Result.Success -> {
-                    try {
-                        CustomSearchEngineStore.addSearchEngine(
-                            context = requireContext(),
-                            engineName = name,
-                            searchQuery = searchString
-                        )
-                    } catch (engineNameExists: CustomSearchEngineStore.EngineNameAlreadyExists) {
```
csadilek, line 168: Do we want to allow duplicate names here now or should we still tell the user that the name is used?

csadilek (commented):
Looks great! Made a first pass and did some smoke testing as well.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -338,22 +354,21 @@ class HomeFragment : Fragment() {
...
-            searchIcon.setBounds(0, 0, iconSize, iconSize)
-
-            withContext(Main) {
-                search_engine_icon?.setImageDrawable(searchIcon)
-            }
+        consumeFlow(store) { flow ->
+            flow.map { state -> state.search.selectedOrDefaultSearchEngine }
+                .ifChanged()
+                .collect { searchEngine ->
+                    val context = context ?: return@collect
```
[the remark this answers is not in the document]
pocmo: Correct. I'll replace it with `requireContext`.

app/src/main/java/org/mozilla/fenix/settings/search/AddSearchEngineFragment.kt
pocmo: This is now possible. Previously the name (instead of the ID) was used to identify the default search engine. Now that we generate a UUID for custom search engines, having duplicate names is no problem. I think adding the complexity for this check is not needed anymore: If the user wants to add search engines with the same name then so be it.

csadilek (approved):
Made another pass and tested locally as well! I think this is ready to 🚢 .

Issue #16053, opened 2020-10-20
[Bug] Crash when closing PB tabs from media notification bar java.lang.IllegalStateException: SET SESSION: Current activity: HomeActivity hashcode 143197250 Other activity: HomeActivity hashcode 104749745


abodea:
[comment]: # (Please do your best to search for duplicate issues before filing a new issue so we can keep our issue board clean)
[comment]: # (Every issue should have the exact bug and steps to reproduce described in it. Please do not file feedback list tickets as it is difficult to parse them and address their individual points)
[comment]: # (Read https://github.com/mozilla-mobile/fenix#i-want-to-file-an-issue for more information)

## Steps to reproduce
1. Launch Fenix and open few tabs in PB.
2. Expand the android quick settings menu.
3. From the notification bar close the PB tabs.
### Expected behavior
The PB tabs are closed and the app still works without problems.
### Actual behavior
Fenix crashes.
### Device information

* Android device: **Google Pixel 4 XL (11).**
* Fenix version: **Nightly 10/20.**

[Sentry-Scorro
](https://crash-stats.mozilla.org/report/index/bd4f84d3-ea51-48af-90c8-3bf870201020#tab-details)<details><summary>Log (Caused by: java.lang.IllegalStateException: Display already acquired) </summary>
<p>

#### Caused by: java.lang.IllegalStateException: Display already acquired

```python
9eb2cc26-f7a8-4ae7-a766-58eb71612857
java.lang.IllegalStateException: SET SESSION: Current activity: HomeActivity hashcode 34142228 Other activity: HomeActivity hashcode 137603258
----
java.lang.IllegalStateException: SET SESSION: Current activity: HomeActivity hashcode 34142228 Other activity: HomeActivity hashcode 137603258
at mozilla.components.browser.engine.gecko.GeckoEngineView.render(GeckoEngineView.kt:18)
at mozilla.components.feature.session.engine.EngineViewPresenter.access$onTabToRender(EngineViewPresenter.kt:9)
at mozilla.components.feature.session.engine.EngineViewPresenter$start$1$invokeSuspend$$inlined$collect$1.emit(Collect.kt:2)
at mozilla.components.support.ktx.kotlinx.coroutines.flow.FlowKt$ifAnyChanged$$inlined$filter$1$2.emit(Collect.kt:19)
at mozilla.components.feature.session.engine.EngineViewPresenter$start$1$invokeSuspend$$inlined$map$1$2.emit(Collect.kt:7)
[...]

PR #16604, opened 2020-11-16
#16053: Fix crash when closing private tabs via notification 

csadilek:
This prevents the "Display Already Acquired" crash when closing private tabs via our system notification. We're only launching to home if the app was in private mode. If we're in normal mode there's nothing to do as the outcome is essentially the same.

Also added a fix here for the incorrect tab count when re-opening the app after private tabs have been removed.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/toolbar/TabCounterToolbarButton.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/toolbar/TabCounterToolbarButton.kt
@@ -11,6 +11,8 @@ import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.flow.collect
 import kotlinx.coroutines.flow.map
 import mozilla.components.browser.state.selector.getNormalOrPrivateTabs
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/mozilla/fenix/components/toolbar/TabCounterToolbarButton.kt
```diff
@@ -77,4 +78,8 @@ class TabCounterToolbarButton(
     private fun updateCount(count: Int) {
         reference.get()?.setCountWithAnimation(count)
     }
+
+    private fun isPrivate(store: BrowserStore): Boolean {
```
csadilek, line 82: We're currently working on uplifting this class to A-C, so let's add this + tests there as well: https://github.com/mozilla-mobile/android-components/pull/8872

ekager: We could also make this a `BrowserStore` ext if we end up fixing different versions of this bug

ekager (approved):
🎭

Issue #16353, opened 2020-11-03
[Bug] When in onboarding mode, focusing on address bar should not cause a jump to homescreen


violasong:
Related to https://github.com/mozilla-mobile/fenix/issues/16351

Currently in Nightly, when viewing the onboarding screen and tapping the address bar, the view behind the keyboard switches from onboarding to homescreen. It's then not possible to get back to onboarding. 

Instead, opening the keyboard should not navigate away from onboarding. 

Ideally, when the keyboard is up we would adapt the same behavior as #16351, where you can still interact with the onboarding screen, and scroll on it to dismiss the keyboard.

liuche:
@violasong could you take a look at this issue #15990 as well? I'm wondering if we should allow users to go back to onboarding once they've decided to start browsing. Does clicking on the urlbar indicate an intent to complete/skip the onboarding walkthrough?

eliserichards:
@violasong Another edge case: 
- a user fresh installs Fenix, then goes to their Gmail and clicks on a link (opens in Fenix). At this point the user has _technically_ navigated. Do we dismiss the onboarding? Or are they still going to see it?

violasong:
@eliserichards:
> a user fresh installs Fenix, then goes to their Gmail and clicks on a link (opens in Fenix). At this point the user has technically navigated. Do we dismiss the onboarding? Or are they still going to see it?

This will be affected by the new UX conversation about changing homescreen to a new tab so that it's part of the backstack. Once that happens, I would expect there to be two tabs in this situation - the link from Gmail and the onboarding tab.

For now, since I assume this isn't possible yet, I think we need to just dismiss the onboarding in favor of showing the link from Gmail. @apbitner and @lime124 may be able to confirm.

apbitner:
Yeah ideally we'd open the link in a new tab and leave onboarding alone. But if that's not possible, I'm not sure what other option we have besides dismissing onboarding to open the link.

Issue #16018, opened 2020-10-19
[Bug] The Onboarding tour is not dismissed after opening a page via the search bar


softvision-miralobontiu:
## Prerequisites
Make sure you have a clean profile.

## Steps to reproduce
1. From the Onboarding tour, tap on any link from the sync card, or "See what's new" card, or "Browse privately" card.
2. Tap on the tab counter.
3. Tap on the "+" to open a new tab.
4. In the search bar type some characters and then "go".
5. Repeat steps 2, and 3, then observe the background.

### Expected behavior
The Onboarding tour is dismissed and it is replaced by the home screen.

### Actual behavior
The Onboarding tour is not dismissed no matter how many tabs the user opens following the STR.

### Device information
* Android device: Samsung Galaxy Note 8 (Android 9), Samsung Galaxy Tab S6 (Android 10), HTC 10 (Android 8)
* Fenix version: 10/19 Nightly, Beta 82.0.0-beta.6, RC 82.1.0-RC.2

<img src="https://user-images.githubusercontent.com/40694158/96426104-7743ca80-1205-11eb-936c-c02aefd52029.gif" width="300" height="550" />

softvision-miralobontiu:
I was able to reproduce this issue on:
* Samsung Galaxy Tab A6 (Android 5.1.1)
* Fenix versions: 10/26 Nightly, beta 83.0.0-beta.1, RC 82.1.1

mcarare:
For QA, also see: https://github.com/mozilla-mobile/fenix/pull/16613#issue-522379526.

sflorean:
Verified as fixed on Nightly 11/27 with Samsung Note10(Android 10).

Issue #15990, opened 2020-10-17
[Bug] During Onboarding, user can use tabs tray to browse, and keep Onboarding in backstack


liuche:
### Acceptance Criteria
UX recommendation: Onboarding should be present until the user visits a url (by hitting enter, clicking the suggestion, etc), or opening a new tab.

[comment]: # (Please do your best to search for duplicate issues before filing a new issue so we can keep our issue board clean)
[comment]: # (Every issue should have the exact bug and steps to reproduce described in it. Please do not file feedback list tickets as it is difficult to parse them and address their individual points)
[comment]: # (Read https://github.com/mozilla-mobile/fenix#i-want-to-file-an-issue for more information)

## Steps to reproduce
Start Fenix fresh install
See Onboarding show up
Tap on Tabs Tray tab counter icon (0) (or swipe up on bottom of screen for gesture)

### Expected behavior
Maybe nothing should happen? but this could be a UX question.

- disable opening tabs tray during onboarding?
- Open tabs tray but make sure we do a better job of dismissing onboarding if user start browsing? or going through Settings?

### Actual behavior
Tabs Tray opens "Your open tabs will be shown here"

You can then get into some weird states:
- Onboarding can be in your history back stack (by opening tabs tray, opening a new tab, browsing, then hitting back)
- Putting Onboarding into private mode (by opening tabs tray and opening a private tab)
- Get into Tabs settings (by clicking on 3-dot menu)

You can also long-press the tab icon to show the "Open new Private Tab" quick action, which will put Onboarding into private mode.

### Device information

* Android device: any
* Fenix version: 83 Nightly

mcarare:
This issue https://github.com/mozilla-mobile/fenix/issues/16018 is also relevant for this discussion.

apbitner:
Onboarding was intended to go away whenever the user visited a page whether that be through search, typing in a URL, a bookmark, etc. I agree that even if the user opens a new tab, this should dismiss the onboarding as well and replace it with the standard home screen.

liuche:
Onboarding should be present until the user **visits** a url (by hitting enter, clicking the suggestion, etc).

violasong:
As mentioned in issue #16353, I think this will be affected by the new conversation around changing homescreen to a new tab so that it's part of the backstack. In that case, onboarding can also be treated as a real link that ends up in the backstack. CC: @lime124

mcarare:
[...]

PR #16613, opened 2020-11-17
For #16353: Dismiss onboarding only when user navigates to a web page by typing or choosing suggestions or clipboard

mcarare:
This should also cover requirements in #15990 and #16018.
The onboarding should not be dismissed when opening a web page from another app, or a link from onboarding cards (including redirects to localized SUMO page).

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
@@ -117,6 +117,8 @@ import org.mozilla.fenix.ext.requireComponents
 import org.mozilla.fenix.ext.settings
 import org.mozilla.fenix.home.HomeScreenViewModel
 import org.mozilla.fenix.home.SharedViewModel
+import org.mozilla.fenix.onboarding.FenixOnboarding
+import org.mozilla.fenix.settings.SupportUtils
 import org.mozilla.fenix.theme.ThemeManager
 import org.mozilla.fenix.utils.allowUndo
 import org.mozilla.fenix.wifi.SitePermissionsWifiIntegration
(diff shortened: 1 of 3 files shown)


Amejia481 (changes requested):
Is it possible to add some tests for verifying the fix and avoiding regressions? :)

app/src/main/java/org/mozilla/fenix/browser/MainOnboardingObserver.kt
```diff
@@ -0,0 +1,49 @@
...
+import org.mozilla.fenix.onboarding.FenixOnboarding
+import org.mozilla.fenix.settings.SupportUtils
+
+/**
+ * Dismisses onboarding when user navigates to a page, other than the links from onboarding screen,
+ * nominated in [onboardingLinksList]
+ */
+class MainOnboardingObserver(
+    private val onboarding: FenixOnboarding
+) : Session.Observer {
```
Amejia481, line 18: We are trying to not use `Session.Observer` unless is strictly necessary as we are going to deprecated it soon, lets use the `BrowserStore` instead, we could do it in two ways:

1. Creating an `OnboardingMiddleware` similar to [RecentlyClosedMiddleware](https://github.com/mozilla-mobile/android-components/blob/5451ed708c21750a18c207a6ecae3972aa9aa839/components/feature/recentlyclosed/src/main/java/mozilla/components/feature/recentlyclosed/RecentlyClosedMiddleware.kt#L31) and checking by the `UpdateLoadRequestAction`

2. Observing the `content.loadRequest` via flow you can find an example [here](https://github.com/mozilla-mobile/fenix/commit/d7688c84278b74e4d96688c3c223cbfa14b4885b)

Amejia481 (approved):
Great job 🚢 !

Issue #16615, opened 2020-11-17
[UI tests] Automate the smoke tests suite


sv-ohorvath:
Smoke tests suite: https://testrail.stage.mozaws.net/index.php?/suites/view/3192&group_by=cases:section_id&group_order=asc&group_id=45740

sv-ohorvath:
Work completed, all suitable tests were created or fixed and re-enabled. TestRail is updated wit status and references for each test.

PR #16616, opened 2020-11-17
For #16615: UI smoke test firstRunScreenTest

sv-ohorvath:
The test ran 30 times on Firebase and passed, see: https://github.com/mozilla-mobile/fenix/runs/1412490648


app/src/androidTest/java/org/mozilla/fenix/ui/robots/HomeScreenRobot.kt
```diff
@@ -56,6 +58,12 @@ import org.mozilla.fenix.helpers.withBitmapDrawable
  * Implementation of Robot Pattern for the home screen menu.
  */
 class HomeScreenRobot {
+    val privateSessionMessage =
+        "${appContext.appName} clears your search and browsing history from private tabs when you close them" +
```
AaronMT, line 62: I'm glad we now know how to get the app name!

AaronMT (approved):

PR #16673, opened 2020-11-20
Bug 1678572 - A github-release shouldn't kick off releases if the bot…

JohanLorenzo:
… made it

Full context at: https://bugzilla.mozilla.org/show_bug.cgi?id=1678572#c0

Kudos to @camd for the idea 👍 (I wanted to add you as a reviewer but the Github UI prevents me from doing so)

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -104,7 +104,7 @@ tasks:
                 tasks_for in ["action", "cron"]
                 || (tasks_for == "github-pull-request" && pullRequestAction in ["opened", "reopened", "synchronize"])
                 || (tasks_for == "github-push" && head_branch[:10] != "refs/tags/") && (head_branch != "staging.tmp") && (head_branch != "trying.tmp")
-                || (tasks_for == "github-release" && releaseAction == "published")
+                || (tasks_for == "github-release" && releaseAction == "published" && (ownerEmail != "mozilla-release-automation-bot@users.noreply.github.com") && (ownerEmail != "mozilla-release-automation-bot-staging@users.noreply.github.com"))
               then:
                   $let:
                       level:


.taskcluster.yml
```diff
@@ -104,7 +104,7 @@ tasks:
                 tasks_for in ["action", "cron"]
                 || (tasks_for == "github-pull-request" && pullRequestAction in ["opened", "reopened", "synchronize"])
                 || (tasks_for == "github-push" && head_branch[:10] != "refs/tags/") && (head_branch != "staging.tmp") && (head_branch != "trying.tmp")
-                || (tasks_for == "github-release" && releaseAction == "published")
+                || (tasks_for == "github-release" && releaseAction == "published" && (ownerEmail != "mozilla-release-automation-bot@users.noreply.github.com") && (ownerEmail != "mozilla-release-automation-bot-staging@users.noreply.github.com"))
```
JohanLorenzo, line 107: You might think this code could be simplified into: 

```yaml
(tasks_for == "github-release" && releaseAction == "published" && (event.sender.login != "mozilla-release-automation-bot") && (event.sender.login != "mozilla-release-automation-bot-staging"))
```

To be honest, I wish it could be the case. Sadly, we hit that class of bugs many times. The **full** `if` statement is evaluated before checking the truthiness of each operand. This means, Taskcluster will try to evaluate `event.sender.login` and will raise an error because that object doesn't exist there. Thus, this `.taskcluter.yml` file will be broken for any case that doesn't define `event.sender.login`. This is why you see variables like `releaseAction` defined this way, just above: 

https://github.com/mozilla-mobile/fenix/blob/4e598d16fc8a16edb992d82fc54eb1b8d2837b93/.taskcluster.yml#L98-L101

It's a way to ensure the `if` statement remains "evaluable" under any circumstances. 

Back to `ownerEmail`. It's entirely defined there: 

https://github.com/mozilla-mobile/fenix/blob/4e598d16fc8a16edb992d82fc54eb1b8d2837b93/.taskcluster.yml#L16-L30

I could define a new variable called `ownerLogin`, but this actually increase the complexity of `.taskcluster.yml`. In fact the `$let` statement[1] doesn't enable us to define a variable that depends on another, in the same statement. We have to embed the second statement in the `in` section of the first one. Meaning, if I wanted to factorize `$event.sender.login`, I would have to shift almost every line of `.taskcluster.yml` to get variables under the right scope. I don't think it's worth the effort. I believe this one-liner is much easier and the logic is straightforward to understand.

[1] https://json-e.js.org/#Language/-let-

MihaiTabara (approved):

JohanLorenzo:
I confirm we needed this patch. Shipit just scheduled Fenix 84b2 and this task got created https://firefox-ci-tc.services.mozilla.com/tasks/groups/KvLDY6lmQj2YKZjdXxDVpA. I manually canceled it.

PR #16738, opened 2020-11-25
For #16615: UI smoke test addPredefinedSearchEngineTest

sv-ohorvath:
Ran the test on Firebase 30 times and passed.


mcarare (approved):
LGTM!

PR #16778, opened 2020-11-30
16373 Count the # of inflations done on startup

MarcLeclair:
The perf team has been trying to improve the regression that happen on startup. One idea was to reduce the # of inflations. This PR aims at counting how many we do on start up at the moment and try to not go above that.

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/androidTest/java/org/mozilla/fenix/perf/StartupExcessiveResourceUseTest.kt
+++ b/app/src/androidTest/java/org/mozilla/fenix/perf/StartupExcessiveResourceUseTest.kt
@@ -25,6 +25,7 @@ private const val EXPECTED_RUNBLOCKING_COUNT = 2
 private const val EXPECTED_COMPONENT_INIT_COUNT = 42
 private const val EXPECTED_VIEW_HIERARCHY_DEPTH = 12
 private const val EXPECTED_RECYCLER_VIEW_CONSTRAINT_LAYOUT_CHILDREN = 4
+private const val EXPECTED_NUMBER_OF_INFLATION = 12
 
 private val failureMsgStrictMode = getErrorMessage(
     shortName = "StrictMode suppression",
@@ -54,6 +55,11 @@ private val failureMsgRecyclerViewConstraintLayoutChildren = getErrorMessage(
 ) + "Please note that we're not sure if this is a useful metric to assert: with your feedback, " +
     "we'll find out over time if it is or is not."
 
+private val failureMsgNumberOfInflation = getErrorMessage(
+    shortName = "Number of inflation on start up doesn't match expected count",
+    implications = "The number of inflation can negatively impact start up time. Having more inflations" +
+            "will most likely mean we're adding extra work on the UI thread."
+)
 /**
  * A performance test to limit the number of StrictMode suppressions and number of runBlocking used
  * on startup.
@@ -90,6 +96,8 @@ class StartupExcessiveResourceUseTest {
         val actualViewHierarchyDepth = countAndLogViewHierarchyDepth(rootView, 1)
         val actualRecyclerViewConstraintLayoutChildren = countRecyclerViewConstraintLayoutChildren(rootView, null)
 
+        val actualNumberOfInflations = InflationCounter.inflationCount.get()
+
         assertEquals(failureMsgStrictMode, EXPECTED_SUPPRESSION_COUNT, actualSuppresionCount)
         assertEquals(failureMsgRunBlocking, EXPECTED_RUNBLOCKING_COUNT, actualRunBlocking)
         assertEquals(failureMsgComponentInit, EXPECTED_COMPONENT_INIT_COUNT, actualComponentInitCount)
@@ -99,6 +107,7 @@ class StartupExcessiveResourceUseTest {
             EXPECTED_RECYCLER_VIEW_CONSTRAINT_LAYOUT_CHILDREN,
             actualRecyclerViewConstraintLayoutChildren
         )
+        assertEquals(failureMsgNumberOfInflation, EXPECTED_COMPONENT_INIT_COUNT, actualNumberOfInflations)
     }
 }
 
--- a/app/src/androidTest/java/org/mozilla/fenix/perf/StartupExcessiveResourceUseTest.kt
+++ b/app/src/androidTest/java/org/mozilla/fenix/perf/StartupExcessiveResourceUseTest.kt
@@ -107,7 +107,7 @@ class StartupExcessiveResourceUseTest {
             EXPECTED_RECYCLER_VIEW_CONSTRAINT_LAYOUT_CHILDREN,
             actualRecyclerViewConstraintLayoutChildren
         )
-        assertEquals(failureMsgNumberOfInflation, EXPECTED_COMPONENT_INIT_COUNT, actualNumberOfInflations)
+        assertEquals(failureMsgNumberOfInflation, EXPECTED_NUMBER_OF_INFLATION, actualNumberOfInflations)
     }
 }
 
--- a/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
+++ b/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
@@ -831,7 +831,7 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
     override fun getSystemService(name: String): Any? {
         if (LAYOUT_INFLATER_SERVICE == name) {
             if (inflater == null) {
-                inflater = PerformanceInflater(LayoutInflater.from(baseContext).cloneInContext(this), this)
+                inflater = PerformanceInflater(LayoutInflater.from(baseContext), this)
             }
             return inflater
         }
--- a/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
+++ b/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
@@ -14,6 +14,7 @@ import android.os.SystemClock
 import android.text.format.DateUtils
 import android.util.AttributeSet
 import android.view.KeyEvent
+import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewConfiguration
 import android.view.WindowManager
@@ -89,6 +90,7 @@ import org.mozilla.fenix.library.bookmarks.DesktopFolders
 import org.mozilla.fenix.library.history.HistoryFragmentDirections
 import org.mozilla.fenix.library.recentlyclosed.RecentlyClosedFragmentDirections
 import org.mozilla.fenix.perf.Performance
+import org.mozilla.fenix.perf.PerformanceInflater
 import org.mozilla.fenix.perf.StartupTimeline
 import org.mozilla.fenix.search.SearchDialogFragmentDirections
 import org.mozilla.fenix.session.PrivateNotificationService
@@ -137,6 +139,8 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
     private val webExtensionPopupFeature by lazy {
         WebExtensionPopupFeature(components.core.store, ::openPopup)
     }
+
+    private var inflater : LayoutInflater?  = null
 
     private val navHost by lazy {
         supportFragmentManager.findFragmentById(R.id.container) as NavHostFragment
@@ -822,6 +826,16 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
         base.components.strictMode.resetAfter(StrictMode.allowThreadDiskReads()) {
             super.attachBaseContext(base)
         }
+    }
+
+    override fun getSystemService(name: String): Any? {
+        if(LAYOUT_INFLATER_SERVICE == name){
+            if(inflater == null) {
+                inflater = PerformanceInflater(LayoutInflater.from(baseContext).cloneInContext(this), this)
+            }
+            return inflater
+        }
+        return super.getSystemService(name)
     }
 
     protected open fun createBrowsingModeManager(initialMode: BrowsingMode): BrowsingModeManager {
--- a/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
+++ b/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
@@ -140,7 +140,7 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
         WebExtensionPopupFeature(components.core.store, ::openPopup)
     }
 
-    private var inflater : LayoutInflater?  = null
+    private var inflater: LayoutInflater? = null
 
     private val navHost by lazy {
         supportFragmentManager.findFragmentById(R.id.container) as NavHostFragment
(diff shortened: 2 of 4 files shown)


app/src/test/java/org/mozilla/fenix/perf/PerformanceInflaterTest.kt
```diff
@@ -0,0 +1,82 @@
...
+
+    @Test
+    fun `WHEN we inflate a view, the inflation counter should increase`() {
+        assertEquals(0, InflationCounter.inflationCount.get())
+        perfInflater.inflate(R.layout.fragment_home, null, false)
+        assertEquals(1, InflationCounter.inflationCount.get())
+    }
+
+    @Test
+    fun `WHEN inflating one of our resource file, the inflater should not crash`(){
```
mcomella: This technically tests robolectric's implementation rather than the android framework's – I wonder if we should make this an on-device test rather than a unit test. They're pretty fast when there's no UI to display. Would that also help remove your `<fragment>` hacks?

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -785,6 +789,16 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
         }
     }
 
+    override fun getSystemService(name: String): Any? {
+        if(LAYOUT_INFLATER_SERVICE == name){
+            if(inflater == null) {
+                inflater = PerformanceInflater(LayoutInflater.from(baseContext).cloneInContext(this), this)
```
mcomella, line 795: Why do we need to clone to pass this in? Add a comment

app/src/androidTest/java/org/mozilla/fenix/perf/StartupExcessiveResourceUseTest.kt
```diff
@@ -99,6 +107,7 @@ class StartupExcessiveResourceUseTest {
             EXPECTED_RECYCLER_VIEW_CONSTRAINT_LAYOUT_CHILDREN,
             actualRecyclerViewConstraintLayoutChildren
         )
+        assertEquals(failureMsgNumberOfInflation, EXPECTED_COMPONENT_INIT_COUNT, actualNumberOfInflations)
```
mcomella, line 110: We're using the wrong expected value here (maybe that's why the UI test failed?)

app/src/main/java/org/mozilla/fenix/perf/PerformanceInflater.kt
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    /**
+     * This code was taken from the PhoneLayoutInflater.java (Similarly, AsyncLayoutInflater
+     * implements it the exact same way too). Looking at the `super.OnCreateView(name, attrs)`,
+     * it hardcodes the prefix as "android.view." this means that a xml element such as
+     * ImageButton will crash the app using android.view.ImageButton.
+     */
+    @Throws(ClassNotFoundException::class)
+    override fun onCreateView(name: String?, attrs: AttributeSet?): View? {
```
mcomella, line 52: Should we mention this is only called for short names like `TextView` and not fully qualified names like `androidx.recyclerview...`

mcomella (changes requested):
This looks good to me. The biggest thing I'm concerned is the license: we need to figure that out before landing.

The next biggest things I'm concerned about are:
- Is LayoutInflater as an arg really necessary (and cloning before it's passed in)? It doesn't seem to be used for anything
- the test expected value

But these last few aren't really blocking though. I only need to re-review the license bit.

app/src/main/java/org/mozilla/fenix/perf/PerformanceInflater.kt
```diff
@@ -0,0 +1,68 @@
...
+){
+
+    private val sClassPrefixList = arrayOf(
+        "android.widget.",
+        "android.webkit.",
+        "android.app."
+    )
+
+    override fun cloneInContext(newContext: Context?): LayoutInflater {
+        return PerformanceInflater(inflater, newContext!!)
```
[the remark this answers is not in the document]
MarcLeclair: So this is an interesting one, doing the `this` instead of the `inflater` leads to a multiple of issue because the phone passes us the `PhoneLayoutInflater`. By doing `this` (which the `PhoneLayoutInflater` does), we end up crashing by missing some classes in the `classLoader` (which is what crashes). I'm assuming because the inflater doesn't implement *everything* that needs to be done by an inflater

app/src/main/java/org/mozilla/fenix/perf/PerformanceInflater.kt
```diff
@@ -0,0 +1,68 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
```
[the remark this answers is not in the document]
MarcLeclair: I'll look into that. That's a good point. I didn't think of it since it was the Android Source code.

app/src/test/java/org/mozilla/fenix/perf/PerformanceInflaterTest.kt
MarcLeclair: Good question about the `<fragment>` hack being solve if we use a on-device test. I was wondering if this would work for on device test since I'm manually retrieving all the files from a specific file path. I thought I wouldn't be able to do so because I assumed it kinda acted as an `.apk`

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -148,6 +152,7 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
         )
     }
 
+    lateinit var factory: LayoutInflater.Factory2
```
mcomella, line 155: nit: can we group this variable together with the inflater? They're highly related and grouping them could give cues to future developers.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -263,6 +269,31 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
...
+                        }.asView()
+                        else -> this@HomeActivity.onCreateView(parent, name, context, attrs)
+                    }
+
+            override fun onCreateView(name: String, context: Context, attrs: AttributeSet): View? {
+                return null
+            }
+        }
+    }
     @CallSuper
```
mcomella, line 297: nit: white space in between

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -263,6 +269,31 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
...
     }
 
+    private fun createInflaterFactory() {
+        factory = object : LayoutInflater.Factory2 {
+            override fun onCreateView(parent: View?, name: String, context: Context, attrs: AttributeSet): View? =
+                    when (name) {
+                        EngineView::class.java.name -> components.core.engine.createView(context, attrs).apply {
+                            selectionActionDelegate = DefaultSelectionActionDelegate(
+                                    BrowserStoreSearchAdapter(
+                                            components.core.store,
```
mcomella, line 279: This is an unreasonable amount of indentation: this is 8 levels and I think ~3-4 is the usual comfortable amount. Can we declare this inflater class in a separate file to lower it? Also the indentation changed from the original PR – you have some double-indentations. Can you fix that?

In general, anonymous classes also worry me as they're harder to understand because they have access to many more references.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -785,6 +791,17 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
         }
     }
 
+    override fun getSystemService(name: String): Any? {
+        if (LAYOUT_INFLATER_SERVICE == name) {
+            if (inflater == null) {
+                // We need to clone the inflater the first time to make sure it is associated with the right context.
```
mcomella, line 797: Why can't we just pass in `this` to set it with the right context the first time? I don't think this comment really helps me understand why we need `cloneInContext` or why `baseContext` is needed.

app/src/main/java/org/mozilla/fenix/perf/PerfFragment.kt
```diff
@@ -0,0 +1,20 @@
...
+import android.view.LayoutInflater
+import androidx.fragment.app.Fragment
+import org.mozilla.fenix.HomeActivity
+
+open class PerfFragment : Fragment() {
+    override fun onGetLayoutInflater(savedInstanceState: Bundle?): LayoutInflater {
+        var fragmentInflater = super.onGetLayoutInflater(savedInstanceState)
+        fragmentInflater = PerformanceInflater(fragmentInflater, this.requireContext())
+                .cloneInContext(this.requireContext())
+        fragmentInflater.factory2 = (this.requireActivity() as HomeActivity).factory
```
mcomella, line 17: This could crash if this fragment is used with a different activity – can we use `as?` to avoid that?

mcomella (changes requested):
This patch isn't inflating all of our views correctly, e.g. there is no "Firefox Preview" logo on the onboarding screen:

image

mcomella:
Also, I noticed some of my comments are not addressed – is that intentional? If so, it might be good to leave a comment to say, "I'm not going to do this" so I know you decided not to. If not, how can I help you find all of the comments? :)

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
MarcLeclair: We don't , that was my mistake. Fixed it! Thanks for noticing the redundancy in my code.

app/src/main/java/org/mozilla/fenix/perf/PerformanceInflater.kt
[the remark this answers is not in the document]
MarcLeclair: This was right (again). It used to crash because of something else and I think this was just a symptom when I was investigating. I changed it to this and we don't lose any functionality now.

app/src/main/java/org/mozilla/fenix/perf/PerformanceInflater.kt
```diff
@@ -0,0 +1,68 @@
...
+ * it takes the `inflater` that is given by the base system. This is done in order not to change
+ * the behavior of the app since all we want to do is count the inflations done.
+ *
+ */
+open class PerformanceInflater(val inflater: LayoutInflater, context: Context) : LayoutInflater(
+    inflater,
+    context
+){
+
+    private val sClassPrefixList = arrayOf(
```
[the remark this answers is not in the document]
MarcLeclair: Should be fixed now!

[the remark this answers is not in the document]
MarcLeclair: Should be fixed too!

app/src/main/java/org/mozilla/fenix/perf/PerformanceInflater.kt
```diff
@@ -0,0 +1,75 @@
...
+    @Throws(ClassNotFoundException::class)
+    override fun onCreateView(name: String?, attrs: AttributeSet?): View? {
+        for (prefix in classPrefixList) {
+            try {
+                val view = createView(name, prefix, attrs)
+                if (view != null) {
+                    return view
+                }
+            } catch (e: ClassNotFoundException) {
+            }
```
mcomella, line 66: nit: we should comment inside `catch` that we want to let the super class inflate it if these don't match. `AsyncLayoutinfalter` makes a similar comment

app/src/test/java/org/mozilla/fenix/perf/PerformanceInflaterTest.kt
```diff
@@ -0,0 +1,78 @@
...
+private class MockInflater(
+    inflater: LayoutInflater,
+    context: Context
+) : PerformanceInflater(
+    inflater,
+    context
+) {
+
+    override fun onCreateView(name: String?, attrs: AttributeSet?): View? {
+        if (name!!.contains("fragment")) {
```
mcomella, line 73: Can you add a comment explaining why we do this and why it's not indicative of a bug?

app/src/test/java/org/mozilla/fenix/perf/PerformanceInflaterTest.kt
```diff
@@ -0,0 +1,78 @@
...
+    fun `WHEN we inflate a view,THEN the inflation counter should increase`() {
+        assertEquals(0, InflationCounter.inflationCount.get())
+        perfInflater.inflate(R.layout.fragment_home, null, false)
+        assertEquals(1, InflationCounter.inflationCount.get())
+    }
+
+    @Test
+    fun `WHEN inflating one of our resource file, the inflater should not crash`() {
+        val fileList = File("./src/main/res/layout").listFiles()
+        if (fileList != null) {
```
mcomella, line 48: When would it be null? I assume never – shouldn't we assert the list isn't null and has entries instead of silently passing the test if the list is null?

app/src/test/java/org/mozilla/fenix/perf/PerformanceInflaterTest.kt
```diff
@@ -0,0 +1,78 @@
...
+            for (file in fileList) {
+                val layoutName = file.name.split(".")[0]
+                val layoutId = testContext.resources.getIdentifier(
+                    layoutName,
+                    "layout",
+                    testContext.packageName
+                )
+                if (layoutId != -1 && !layoutNotToTest.contains(layoutName)) {
```
mcomella, line 56: When would layoutId == -1? How do we know any of the layouts are getting passed here? Can we assert that not -1 instead?

app/src/test/java/org/mozilla/fenix/perf/PerformanceInflaterTest.kt
```diff
@@ -0,0 +1,78 @@
...
+import org.mozilla.fenix.R
+import org.mozilla.fenix.helpers.FenixRobolectricTestRunner
+import java.io.File
+
+@RunWith(FenixRobolectricTestRunner::class)
+class PerformanceInflaterTest {
+
+    private lateinit var perfInflater: MockInflater
+
+    private val layoutNotToTest = setOf(
```
mcomella, line 26: nit: layoutsNotToTest

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -785,6 +789,16 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
         }
     }
 
+    override fun getSystemService(name: String): Any? {
```
mcomella, line 792: So in what cases is the inflater used and which isn't it used? e.g. is it used for inflating Fragment views but not Activity views?

I wonder if there's anywhere good we can comment about this that isn't likely to go out of date.

mcomella (approved):
Looks good to me – I ran the PR locally and it seems to create the same view results as before the PR. I think the test can use some clean up to ensure we aren't silently passing it.

PR #16785, opened 2020-11-30
Issue #12731: Log and record breadcrumb when onTrimMemory() gets invoked.

pocmo:
This hopefully can help us debug memory issues and specifically #12731 where we may see the content process getting killed quite often.


csadilek (approved):
🔍

Issue #14443, opened 2020-08-29
Investigate why favicon for frecent sites are slow to fetch


gabrielluong:
## Steps to reproduce
open a new tab
watch the 8 Top Sites slots

### Expected behavior
Top Sites favicons are instantly present

### Actual behavior
Top Sites favicons appear from top left to bottom right over perhaps 0.5 seconds. It looks unpolished.
![HiccupInLoadingFavicons](https://user-images.githubusercontent.com/11428869/98210952-07924700-1f4a-11eb-9dfe-99ca20a0fbd0.gif)
[video](https://drive.google.com/file/d/1JVh-cOmY7y2gLdvFrget23MFv0EBtjf0/view?usp=sharing)

### Device information
* Android device: Galaxy A40
* Fenix version: Today's

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-5912)

violasong:
I was talking to Jaws about this, and he referenced [this issue](https://bugzilla.mozilla.org/show_bug.cgi?id=428378).

gabrielluong:
> I was talking to Jaws about this, and he referenced [this issue](https://bugzilla.mozilla.org/show_bug.cgi?id=428378).

Thanks! This is super helpful to know that this is a common problem.

liuche:
from @mcomella:

    Triage: we think we might be able to do this with better caching (we shouldn't block to make this happen). Moving to inter team list to investigate the possibility.

mcomella:
Triage: suggested action would be to take a profile to determine what's taking so long for the data to arrive. For getting started with profilers, see our docs: https://wiki.mozilla.org/Performance/How_to_get_started_on_Fenix

mcarare:
I have tested this recently on several devices, and I do not see any delay in icon loading anymore.
Can QA also test this (maybe on slow devices?) or @gabrielluong confirm if this is still an issue? Thank you!

Cheap-Skate:
I still have this issue on today's Fenix (201020 17:01), The first favicon - Pocket Top Articles - appears instantly, the other 7 icons might have got a little bit faster but it still takes perhaps 0.25 second for them to appear. 

(it's possible that my 7 Top Sites are slow because they are pinned. Are yours?)

Galaxy A40 (midrange device), Android 10

ebalazs-sv:
[...]

Issue #15499, opened 2020-09-28
[Bug] Top sites flash when removing one with frecent is disabled


jonalmeida:
Created from https://github.com/mozilla-mobile/fenix/issues/14230#issuecomment-696029179

## Steps to reproduce
1. Disable frecent top sites.
2. Add a top site.
3. Remove the top site.

### Expected behaviour
- When removing the top site, the animation should animate only the one top site removed.

### Actual behaviour
- All top sites animate instead.

### Device information

* Android device: Motorola Moto G6 (Android 8), Google Pixel 2 (Android 9), Nexus 5 (Android 6.0.1)
* Fenix version: Nightly 200921 06:05 (Build #2015765171) GV 82.0a1 from 9/21

![](https://user-images.githubusercontent.com/48799384/93756400-bcc4a600-fc0d-11ea-80e6-c5e2451077f4.gif)

lobontiumira:
Verified as fixed on the 92.0a1 Nightly build from 7/21 with Google Pixel (Android 10), and HTC 10 (Android 8).
I've encountered a crash after disabling the "Show most visited sites" and returning to the homescreen (filed #20449).

PR #16810, opened 2020-12-02
For #14443  - Hide topSites until they've loaded

codrut-topliceanu:
Adds a gray cover imageView on top of each topSite icon that is removed after the topSites have loaded.

For https://github.com/mozilla-mobile/fenix/issues/14443

### Pull Request checklist

- [x] **Screenshots**: This PR includes a GIF.

1. click on Show All Checks,

reveal topsites

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/ext/BrowserIcons.kt
+++ b/app/src/main/java/org/mozilla/fenix/ext/BrowserIcons.kt
@@ -4,8 +4,35 @@
 
 package org.mozilla.fenix.ext
 
+import android.graphics.drawable.Drawable
+import android.view.View
 import android.widget.ImageView
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
 import mozilla.components.browser.icons.BrowserIcons
 import mozilla.components.browser.icons.IconRequest
 
-fun BrowserIcons.loadIntoView(view: ImageView, url: String) = loadIntoView(view, IconRequest(url))
+fun BrowserIcons.loadIntoView(
+    view: ImageView,
+    url: String = "",
+    localImage: Drawable? = null,
+    iconCover: ImageView? = null
+): Job? {
+    if (localImage != null) view.setImageDrawable(localImage)
+    GlobalScope.launch {
+        if (url.isNotBlank()) {
+            loadIntoView(view = view, request = IconRequest(url))
+        }
+        delay(DELAY_UNTIL_REVEAL)
+        withContext(Dispatchers.Main) {
+            iconCover?.visibility = View.INVISIBLE
+        }
+    }
+    return null
+}
+
+const val DELAY_UNTIL_REVEAL = 250L
--- a/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/topsites/TopSiteItemViewHolder.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/topsites/TopSiteItemViewHolder.kt
@@ -10,7 +10,11 @@ import android.view.MotionEvent
 import android.view.View
 import android.widget.PopupWindow
 import androidx.appcompat.content.res.AppCompatResources.getDrawable
-import kotlinx.android.synthetic.main.top_site_item.*
+import kotlinx.android.synthetic.main.top_site_item.favicon_cover
+import kotlinx.android.synthetic.main.top_site_item.favicon_image
+import kotlinx.android.synthetic.main.top_site_item.pin_indicator
+import kotlinx.android.synthetic.main.top_site_item.top_site_item
+import kotlinx.android.synthetic.main.top_site_item.top_site_title
 import mozilla.components.browser.menu.BrowserMenuBuilder
 import mozilla.components.browser.menu.item.SimpleBrowserMenuItem
 import mozilla.components.feature.top.sites.TopSite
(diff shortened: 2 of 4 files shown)


app/src/main/java/org/mozilla/fenix/ext/BrowserIcons.kt
```diff
@@ -4,8 +4,35 @@
...
 
-fun BrowserIcons.loadIntoView(view: ImageView, url: String) = loadIntoView(view, IconRequest(url))
+fun BrowserIcons.loadIntoView(
+    view: ImageView,
+    url: String = "",
+    localImage: Drawable? = null,
+    iconCover: ImageView? = null
+): Job? {
+    if (localImage != null) view.setImageDrawable(localImage)
+    GlobalScope.launch {
```
pocmo, line 26: With `GlobalScope` this can live longer than the activity/fragment and keeps a reference to views, so potentially leaks?

CC @csadilek

codrut-topliceanu:
Hey, thanks for taking a look @pocmo . 
I'm not really happy with my solution either and if you have any input please let me know. I initially tried adding a new property to `TopSite` that would indicate if the topSite icon should be shown or not (and after loading one I would send a`HomeFragmentAction` to update the other rows), but that patch got really really large and I still didn't get it to work 100% of the time for some reason, so I just went with this easier but just as inelegant approach.

pocmo:
This was only a drive-by review. I think the actual review should be done by someone more familiar with the top sites code. I'll unassign myself.

> I'm not really happy with my solution either and if you have any input please let me know. 

Maybe it would be an option to use a scope bound to the view (`view.toScope()` from AC?). But I'd delegate this question to someone who has worked on the code here. @gabrielluong @jonalmeida ?

app/src/main/java/org/mozilla/fenix/ext/BrowserIcons.kt
```diff
@@ -4,8 +4,35 @@
...
 import mozilla.components.browser.icons.IconRequest
 
-fun BrowserIcons.loadIntoView(view: ImageView, url: String) = loadIntoView(view, IconRequest(url))
+fun BrowserIcons.loadIntoView(
+    view: ImageView,
+    url: String = "",
+    localImage: Drawable? = null,
+    iconCover: ImageView? = null
+): Job? {
+    if (localImage != null) view.setImageDrawable(localImage)
```
jonalmeida, line 25: If we look at the [internals](https://searchfox.org/mozilla-mobile/source/android-components/components/browser/icons/src/main/java/mozilla/components/browser/icons/BrowserIcons.kt#181) of `loadIntoView`, we already have an API that sets a placeholder until the root image is available. Would that work?

app/src/main/java/org/mozilla/fenix/ext/BrowserIcons.kt
```diff
@@ -4,8 +4,35 @@
...
+    GlobalScope.launch {
+        if (url.isNotBlank()) {
+            loadIntoView(view = view, request = IconRequest(url))
+        }
+        delay(DELAY_UNTIL_REVEAL)
```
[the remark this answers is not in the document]
jonalmeida: It looks to me that we're using this delay to synchronize the setting of image with all the top sites. I don't think an arbitrary delay value would work here because devices and networks vary. Maybe we need to update the top sites adapter to show the top sites icon when all the `loadIntoView` calls return? I recall Petru doing [something similar](https://github.com/mozilla-mobile/android-components/pull/8894) for the tabs tray adapter as well that we could try. What do you think?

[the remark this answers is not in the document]
codrut-topliceanu: Thank you for looking into this @jonalmeida.
> We use this `loadIntoView` in [a few places](https://searchfox.org/mozilla-mobile/search?q=loadIntoView&path=fenix&case=false&regexp=false) other than top sites, so this delay would affect all those cases. Is that what we want?

You are right. Sorry about that :(

 
> I don't think an arbitrary delay value would work here because devices and networks vary. Maybe we need to update the top sites adapter to show the top sites icon when all the loadIntoView calls return?

I've spent quite a bit of time in trying to implement just this, but in the end there were a lot of changes (17 files modified) and it wasn't working great. Plus it was either wait for the first topSite to load and unveil all of them OR wait for all topSites to load and unveil them then, the latter choice could break the whole system if one of the topSite icons just wouldn't load, while the former wouldn't be much of an improvement.

I put this PR together rather quickly and I apologise for its low quality. I have to admit I'm not sure if there is any satisfactory solution to this issue, at least not with my current experience level, so I'm unsure how to proceed. I'll take a look at Petru's PR, but I've already talked to him about this issue.

app/src/main/java/org/mozilla/fenix/ext/BrowserIcons.kt
codrut-topliceanu: I have tried using it, but the refresh of HomeFragment when clicking new tab FAB would make the whole topsites area flicker somehow. I'll look into it some more.

jonalmeida: Maybe you're seeing https://github.com/mozilla-mobile/fenix/issues/15499 which I was struggling to fix as well. Might be time to fix this in that case if it's stopping up for building other features.

app/src/main/java/org/mozilla/fenix/ext/BrowserIcons.kt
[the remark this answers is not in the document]
jonalmeida: I wouldn't want to hold you up on a this if we can't find a better solution, so maybe let's try to localize the solution to just the top sites then?

Not sure if you've tried this already, but instead of using `loadIntoView` how about trying `loadIcon` so you can call `awaitAll` on the icons and notify the adapter when that is complete. I wouldn't spend too much time to get this to work if it ends up being as messy as the other solution since you've already spent time on this. 🙂

codrut-topliceanu:
I'm going to close this PR as I don't think my solution is the one we should go with and I'm having trouble coming up with an alternative. Thank you for your time @jonalmeida ❤️

PR #16812, opened 2020-12-02
For#16591: Split UI smoke tests into smaller tests

sv-ohorvath:
Splits a large smoke UI test (verifyPageMainMenuItemsTest) into smaller pieces for better tracking of flaky parts.


isabelrios (approved):
LGTM! Yeah, I was not totally in favor of having those very long tests... they were written that way to completely follow the manual tests as they are designed. But, I would prefer this approach, tests will be more stable than before and we are still checking the functionality. We would need, if not done yet, to update all the references in Test Rail... Thanks!

sv-ohorvath:
>  We would need, if not done yet, to update all the references in Test Rail... Thanks!

Will do! Thanks!

Issue #16032, opened 2020-10-19
Support installation of Recommended Extensions from AMO


jvillalobos:
### What is the user problem or growth opportunity you want to see solved?

Installation of add-ons from AMO, for those add-ons that are currently supported on Fenix by default. This is currently handled via collections, but AMO has a Recommended Extensions for Fenix list which is what we want to use as the source of truth moving forward.

### How do you know that this problem exists today? Why is this important?

AMO gets significant traffic from mobile browsers and we want it to support basic navigation and installation of the same extensions that are currently supported by default.

### Who will benefit from it?

Fenix users who look for add-ons to install on AMO rather than the Add-ons Manager.

csadilek:
Current plan is to implement this via deep links:

- AMO would have to add (`fenix://`) links including the ID of the recommended add-on to install 
- Fenix would:
  - Verify that request comes from AMO 
  - Verify that add-on is in list of recommended add-ons
  - Trigger installation via Add-ons Manager

brampitoyo:
@csadilek I’ve updated the Miro document to account for this upcoming AMO support.

User flow: https://miro.com/app/board/o9J_kw8Lt8g=/?moveToWidget=3074457351173371580&cot=10

Mockups: https://miro.com/app/board/o9J_kw8Lt8g=/?moveToWidget=3074457349916418330&cot=14

As we’ve discussed a couple of weeks back, user-facing changes will be minor:

1. People will be able to visit and/or search the AMO website in multiple ways:
    - Fenix address bar → type keywords → search engine results page → AMO
    - Fenix address bar → type URL → AMO
    - Fenix Add-ons Manager → “Find more add-ons” button → AMO
    - Fenix Add-ons Manager → “Search addons.mozilla.org” search box → AMO
2. People will be able to install add-ons by tapping the “+ Add to Firefox” button on the AMO website
    - In addition to installing via the Fenix Add-ons Manager – a feature we already support today

This will solve a major user problem: many people find add-ons through search engines and the AMO website, but couldn’t install it there.

It will also future-proof the Fenix Add-ons Manager for the foreseeable future. The Manager will focus on showing a curated list of ≤30 recommended and/or verified add-ons. The rest of our catalogue will be listed and installed on AMO.

csadilek:
[...]

PR #16818, opened 2020-12-02
For #16032: Support installing recommended add-ons from AMO

csadilek:
Brings in functionality to install recommended add-ons from AMO. Note that this will need to be activated on AMO once released.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/AppRequestInterceptor.kt
+++ b/app/src/main/java/org/mozilla/fenix/AppRequestInterceptor.kt
@@ -7,6 +7,7 @@ package org.mozilla.fenix
 import android.content.Context
 import android.net.ConnectivityManager
 import androidx.core.content.getSystemService
+import androidx.navigation.NavController
 import mozilla.components.browser.errorpages.ErrorPages
 import mozilla.components.browser.errorpages.ErrorType
 import mozilla.components.concept.engine.EngineSession
@@ -14,8 +15,18 @@ import mozilla.components.concept.engine.request.RequestInterceptor
 import org.mozilla.fenix.components.metrics.Event
 import org.mozilla.fenix.ext.components
 import org.mozilla.fenix.ext.isOnline
+import java.lang.ref.WeakReference
 
-class AppRequestInterceptor(private val context: Context) : RequestInterceptor {
+class AppRequestInterceptor(
+    private val context: Context
+) : RequestInterceptor {
+
+    private var navController: WeakReference<NavController>? = null
+
+    fun setNavigationController(navController: NavController) {
+        this.navController = WeakReference(navController)
+    }
+
     override fun onLoadRequest(
         engineSession: EngineSession,
         uri: String,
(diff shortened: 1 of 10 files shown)


Amejia481 (approved):
Looks great 🚢 !

app/src/main/java/org/mozilla/fenix/AppRequestInterceptor.kt
```diff
@@ -59,6 +75,42 @@ class AppRequestInterceptor(private val context: Context) : RequestInterceptor {
...
+     * the original request can be skipped, otherwise null to continue loading the page.
+     */
+    private fun interceptAmoRequest(
+        uri: String,
+        isSameDomain: Boolean,
+        hasUserGesture: Boolean
+    ): RequestInterceptor.InterceptionResponse? {
+        // First we execute a quick check to see if this is a request we're interested in i.e. a
+        // request triggered by the user and coming from AMO.
+        if (hasUserGesture && isSameDomain && uri.startsWith(AMO_BASE_URL)) {
```
evilpie, line 92: This seems potentially insecure. `AMO_BASE_URL` doesn't end with a slash.

Amejia481: Could you be a bit more explicit or give an example? :)

evilpie: `AMO_BASE_URL` is `https://addons.mozilla.org`. So when `uri` is for example `https://addons.mozilla.org.evil.com`, `uri.startsWith(AMO_BASE_URL)` would be `true`.

csadilek: Thanks. I will update this in a bit. The actual match against the URL is requiring the `/` though so it shouldn't be insecure as is, but good to change.

Amejia481: We are additionally checking against [AMO_INSTALL_URL_REGEX](https://github.com/mozilla-mobile/fenix/pull/16818/files#diff-38274d9212603acf816a7fa82c02e6820dc25a8ed3a489e50b5e57ac9f6ca205R172)  on line [95](https://github.com/mozilla-mobile/fenix/pull/16818/files#diff-38274d9212603acf816a7fa82c02e6820dc25a8ed3a489e50b5e57ac9f6ca205R95)

evilpie: Ah you are right. Sorry

Amejia481: No, thanks for taking a look ❤️ !

app/src/main/java/org/mozilla/fenix/AppRequestInterceptor.kt
```diff
@@ -116,5 +168,7 @@ class AppRequestInterceptor(private val context: Context) : RequestInterceptor {
     companion object {
         internal const val LOW_AND_MEDIUM_RISK_ERROR_PAGES = "low_and_medium_risk_error_pages.html"
         internal const val HIGH_RISK_ERROR_PAGES = "high_risk_error_pages.html"
+        internal const val AMO_BASE_URL = "https://addons.mozilla.org"
```
willdurand, line 171: Hey, quick question: would it be possible to configure this value with a pref?

csadilek: Hey, Definitely possible. Reason we didn't do it is that we wanted to limit installation to AMO. And also switching to `staging` and `dev` won't "just work", because the add-on IDs need to line up with the IDs from our collection. Is that what you had in mind?

We could introduce one we switch to the recommended API then? Another question would be if you'd want a build config and pref you can change at run-time. If it's the latter, we'd have to limit this to Nightly for now.

PR #16842, opened 2020-12-03
For #16615: Smoke UI test swipeToSwitchTabTest

sv-ohorvath:
Ran on firebase 25 times and passed: https://github.com/mozilla-mobile/fenix/runs/1493320927

### DIFF
--- a/app/src/androidTest/java/org/mozilla/fenix/ui/SmokeTest.kt
+++ b/app/src/androidTest/java/org/mozilla/fenix/ui/SmokeTest.kt
@@ -36,6 +36,15 @@ class SmokeTest {
     private var awesomeBar: ViewVisibilityIdlingResource? = null
     private var searchSuggestionsIdlingResource: RecyclerViewIdlingResource? = null
 
+    // This finds the dialog fragment child of the homeFragment, otherwise the awesomeBar would return null
+    private fun getAwesomebarView(): View? {
+        val homeFragment = activityTestRule.activity.supportFragmentManager.primaryNavigationFragment
+        val searchDialogFragment = homeFragment?.childFragmentManager?.fragments?.first {
+            it.javaClass.simpleName == "SearchDialogFragment"
+        }
+        return searchDialogFragment?.view?.findViewById(R.id.awesome_bar)
+    }
+
     @get:Rule
     val activityTestRule = HomeActivityTestRule()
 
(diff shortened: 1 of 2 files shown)


app/src/androidTest/java/org/mozilla/fenix/ui/robots/BrowserRobot.kt
```diff
@@ -338,6 +338,28 @@ class BrowserRobot {
         assertTrue(pausedStateMessage.waitForExists(waitingTime))
     }
 
+    fun swipeNavBarRight(tabUrl: String) {
+        // failing to swipe on Firebase sometimes, so it tries again
+        try {
+            navURLBar().perform(ViewActions.swipeRight())
+            assertTrue(mDevice.findObject(UiSelector().text(tabUrl)).waitUntilGone(waitingTime))
+        } catch (e: AssertionError) {
+            navURLBar().perform(ViewActions.swipeRight())
```
AaronMT, line 347: Why is the action repeated in the catch?

Same below

app/src/androidTest/java/org/mozilla/fenix/ui/SmokeTest.kt
```diff
@@ -36,6 +36,15 @@ class SmokeTest {
...
 
+    // This finds the dialog fragment child of the homeFragment, otherwise the awesomeBar would return null
+    private fun getAwesomebarView(): View? {
+        val homeFragment = activityTestRule.activity.supportFragmentManager.primaryNavigationFragment
+        val searchDialogFragment = homeFragment?.childFragmentManager?.fragments?.first {
+            it.javaClass.simpleName == "SearchDialogFragment"
+        }
+        return searchDialogFragment?.view?.findViewById(R.id.awesome_bar)
+    }
+
```
AaronMT, line 47: Fragments still relatively new to me, so I'm not sure if there's any possible problems with this – but I don't think there's standing out here that looks off.

AaronMT (approved):
See feedback

app/src/androidTest/java/org/mozilla/fenix/ui/SmokeTest.kt
sv-ohorvath: Actually, that is a part from a different PR that has already landed, here I just moved it. Sorry for the confusion.

app/src/androidTest/java/org/mozilla/fenix/ui/robots/BrowserRobot.kt
sv-ohorvath: I tried to make it repeat the swipe in case the first time didn't work, without failing (twice I think is enough). Or do you think there's a better way to do that, with a loop? You can see the first runs of this test, without this hack that the swipe didn't do anything: https://console.firebase.google.com/u/0/project/moz-fenix/testlab/histories/bh.66b7091e15d53d45/matrices/7088753260638706015/executions/bs.80bdb045aa9a45ce/videos

isabelrios (approved):
LGTM, but I'm concerned about adding new tests that may be flaky to the smoketest... if we are going to use this test suite for beta tests... 
Anyway we could use only a subset of smokest tests for that until we monitor new tests enough to be sure they are 100% reliable

sv-ohorvath:
> LGTM, but I'm concerned about adding new tests that may be flaky to the smoketest... if we are going to use this test suite for beta tests...
> Anyway we could use only a subset of smokest tests for that until we monitor new tests enough to be sure they are 100% reliable

I agree, I was actually thinking to move this out to a different Test class and see if we can use it later as a smoke test.

Issue #10417, opened 2020-05-05
[Bug] Restored tabs from collections show old cached content


staktrace:
## Steps to reproduce
I don't have concrete STR but I see this frequently, where I open Fenix and tap on a link to news.ycombinator.com in one of my collections. This is from a cold start, sometimes even after a device reboot, so it's definitely not Fenix sticking around in the background.

### Expected behavior
Page opens with latest version.

### Actual behavior
Page opens with the same content I was seeing last time I opened the page, typically a day or two previously. My assumption is that the page is fresh but after a reading a few headlines I get suspicious because they all look familiar. I reload the page and see that now I get fresh content.

### Device information
* Android device: Sony Xperia XZ1C
* Fenix version: Nightly 200504 06:01 (Build #21250608)
40.0.20200430224518, 2227e0a76
GV: 77.0a1-20200430082621
AS: 0.58.2

ekager:
Dupe of https://github.com/mozilla-mobile/fenix/issues/5869. Unfortunately still waiting on bugzilla.mozilla.org/show_bug.cgi?id=1630000

PyvesB:
@ekager @eliserichards I'm facing this problem as well. Is this really a duplicate? The two issues you linked seem specific to back button behaviour, whereas this is related to collections, and only became a problem in the past few days. 😉

ekager:
Ah I understand. I will reopen and retitle

ContentPersonality2:
I'm having this issue too. I believe it's specific to collections and I can confirm its happening on mspoweruser.com and bbc.co.uk/news.

JNavas2:
It seems the restoring of old (cached) content is **by design**–see [Comment](https://github.com/mozilla-mobile/fenix/issues/11506#issuecomment-643264112) on #11506
> This is intentional behavior with collections currently. We are restoring the state of the tab with history as well.

I think this unfortunate, because it greatly reduces the usefulness of Collections to me.
I had hoped/assumed Collections would be like Tab Groups.
I'll have to go back to using Bookmark folders. :disappointed:

maroonmoon13:
I agree it's a big "regression" as collections in FF never worked like that. I have been using it since the first preview, and Its a learning curve to get used to this new implementation. But I do feel that for a bunch of stale tabs, a lot of prime real estate on the front page is being wasted. I have removed everything from collections and added everything as top sites. But that's just a bunch of icons which cannot be even moved around or categorized. that's why, "regression"

[...]

PR #16865, opened 2020-12-04
For #10417: Reload tab after restore from collection

eliserichards:
For #10417

- [x] `openToBrowser` then reload session when the tab has been restored so that the page is up to date and the session backstack is maintained
- [x] controller test

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlController.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlController.kt
@@ -16,6 +16,7 @@ import mozilla.components.browser.state.state.selectedOrDefaultSearchEngine
 import mozilla.components.browser.state.store.BrowserStore
 import mozilla.components.concept.engine.Engine
 import mozilla.components.concept.engine.prompt.ShareData
+import mozilla.components.feature.session.SessionUseCases
 import mozilla.components.feature.tab.collections.TabCollection
 import mozilla.components.feature.tab.collections.ext.restore
 import mozilla.components.feature.tabs.TabsUseCases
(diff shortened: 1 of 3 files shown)


app/src/test/java/org/mozilla/fenix/home/DefaultSessionControlControllerTest.kt
```diff
@@ -142,42 +142,29 @@ class DefaultSessionControlControllerTest {
...
+                match<NavDirections> {
+                    it.actionId == R.id.action_global_collectionCreationFragment
+                },
                 null
             )
         }
     }
 
     @Test
-    fun `handleCollectionOpenTabClicked onFailure`() {
```
eliserichards, line 152: We no longer need the failure case because we are not restoring the tab state

ekager:
Hm looking at the last comment with a decision here: https://github.com/mozilla-mobile/fenix/issues/10417#issuecomment-714666960

it looks like we still want to restore the tabs, but we just want to add a refresh of the session/tab right after restore.

eliserichards:
Why would we refresh and then reload? That seems like bad user experience

ekager:
So `openToBrowserAndLoad` will just load the last URL from when the tab was saved into a collection. Restore should also restore the back stack (ie history) of that tab, but could be outdated. FWIU we want to restore the tab (and navigate to it) and refresh so that the tab isn't outdated.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlController.kt
```diff
@@ -212,6 +214,7 @@ class DefaultSessionControlController(
             tab,
             onTabRestored = {
                 activity.openToBrowser(BrowserDirection.FromHome)
+                reloadUrlUseCase.invoke(sessionManager.selectedSession)
```
pocmo, line 217: Can you help me understand why this is needed? It somewhat defeats the purpose of the tab getting restored in its previous state (which is weird at times, but if we do not want that then we should change that feature).

eliserichards: Yes! According to @liuche's [comment](https://github.com/mozilla-mobile/fenix/issues/10417#issuecomment-714666960) on the issue, we want to keep the backstack of the page at the time that it was added to the collection, but have a fresh version of the page (e.g. if it is a news site, we should see the latest news not the news at the time we added to the collection). To keep the backstack, we `restore` the page; then we `reload` to make it up-to-date.

Does that make sense?

ekager (approved):

PR #16891, opened 2020-12-07
Update Android Components to 69.0.20201207103252 (Fixing breaking API changes)

pocmo:
We had some breaking API changes in AC. This PR updates AC and fixes the breaking changes.


mcarare (approved):
LGTM! This just fixes the breaking changes, it does not intend to completely replace the uses of session manager, right?

pocmo:
> LGTM! This just fixes the breaking changes, it does not intend to completely replace the uses of session manager, right?

Correct. This is the long-term goal. But a lot of tiny steps are needed to get there. Progress is tracked on this board:
https://github.com/orgs/mozilla-mobile/projects/31

PR #16908, opened 2020-12-08
16900 async navgraph inflation

MarcLeclair:
This PR aims to put the inflation of the Nav graph.

Timing of current master (WITHOUT async nav graph):  `1.086 seconds` 

Timing with the changes: `1.061 seconds`

Timing if we remove the `job.join()` portion of the changes: `1.052 seconds`

### Pull Request checklist

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -242,15 +244,19 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
 
         components.core.requestInterceptor.setNavigationController(navHost.navController)
 
+        MainScope().launch(Dispatchers.Main) {
```
mcomella, line 247: I think you want:
```kotlin
runBlocking {
    navGraphJob.await()
}
```

I think the code you wrote would launch a co-routine that eventually would run on the main thread that is basically a no-op (it would run, suspend on `await`, resume when `await` is done, and end.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -242,15 +244,19 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
...
     }
 
-    private fun createNavGraph() {
-        val navHostFragment : NavHostFragment  = container as NavHostFragment
-        val inflater = navHostFragment.navController.navInflater
-        val graph = inflater.inflate(R.navigation.nav_graph)
-        navHostFragment.navController.graph = graph
-    }
+    private fun createNavGraph() = MainScope().async(Dispatchers.Default) {
+            val navHostFragment = container as NavHostFragment
```
mcomella, line 255: nit: indentation is wrong. should be 4 spaces in the middle and closing brace lines up with `private fun`, like a normal function

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -175,7 +177,7 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
         setupThemeAndBrowsingMode(getModeFromIntentOrLastKnown(intent))
         setContentView(R.layout.activity_home)
 
-        createNavGraph()
+        val navGraphJob = createNavGraph()
```
mcomella, line 180: Can this break if the inflation takes too long? We can navigate using the `navController` before `onCreate` finishes. For example, see `it.process` (below) -> OpenBrowserIntentProcessor -> activity.openToBrowser which calls `navController.nav` (which I assume needs to have the navGraph available, but maybe not!). You can probably test this by calling `delay` in the async task before it assigns the navGraph.

If so, one thought: could we start async inflation in `FenixApplication`? e.g. get the navInflater (programmatically creating a NavHostFragment if we need to) and starting the inflation early?

mcomella, line 180: Thinking aloud, I'm concerned we're going to lose track of all the async + "barrier" things we're doing... I wonder if we should put a light abstraction layer around this so we don't have to rummage through the code to remember all of them (e.g. put all the async stuff in one file).

Maybe this isn't needed yet but throwing it out there.

mcomella (changes requested):
I'm concerned this could break: what do you think?

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
csadilek: Yes, also I don't this is safe. You will need to make sure no navigation can happen before this runs to completion, and with this (unchanged) various other parts of the system would need to "join" the job.

csadilek (changes requested):

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -175,6 +174,8 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
         setupThemeAndBrowsingMode(getModeFromIntentOrLastKnown(intent))
         setContentView(R.layout.activity_home)
 
+        navGraphJob = createNavGraphAsync()
```
mcomella, line 177: Can we comment why we're doing this and link to the associated bug? This will make it easier to re-evaluate this in the future.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -134,7 +130,10 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
         WebExtensionPopupFeature(components.core.store, ::openPopup)
     }
 
+    private lateinit var navGraphJob : Deferred<Unit>
+
     private val navHost by lazy {
+        loadNavFragment()
```
mcomella, line 136: I'm concerned about some aspects of the current implementation:
- we're not giving `navHost` the maximum amount of time to inflate before we use it. Currently, we're blocking 35ms in a local cold VIEW profile I took https://share.firefox.dev/39QZpuH We should reorder the methods inside `onCreate` (i.e. navHost methods come last) to optimize. Also, can we start this process earlier? e.g. create a  NavHostFragment async in `FenixApplication` and attach the fragment in `HomeActivity` when it's ready. I wonder if that would prevent us from blocking at all?
- it's easy to introduce new regressions or crashes because 1) it's unclear that `navHost` will block if it's not ready yet, 2) you can access `navHost` via the view instead (`container`), and 3) this functionality isn't well encapsulated

I'm thinking we do something like:
```kotlin
/** encapsulates async creation of NavHostFragment. Comment explaining why we did it */
class NavHostFragmentAsyncProvider

fun FenixApplication.onCreate {
    startEarlyAsyncStuff()
}

fun FenixApplication.startEarlyAsyncStuff() {
    startAsyncMegazord()
    startAsyncNavHostFragmentProvider()
}

// HomeActivity.kt...
private lateinit var navHost: NavHostFragment

fun HomeActivity.onCreate() {
    // code not requiring navHost
    navHost = components.navHostFragmentAsyncProvider.getBlocking()
    fragmentManager.beginTransaction().add(... // hopefully not underperformant
    // code requiring navHost
}
```

I like that it encapsulates the async code, features a consistent barrier point (making it easier to follow in profiles, harder to mis-use, and easier to refactor), and maximizes the time navHost. It'll also make for more performant WARM starts since we don't need to reinflate the nav graph.

@MarcLeclair What do you think? I'm concerned that my suggestion is "too perfect" but then again async code is really hard to maintain so... 🤷

mcomella (changes requested):
We have the right approach but I wonder if we can improve the maintainability and performance of the implementation. Let me know what you think!

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
MarcLeclair: I like the suggestion. I'm wondering though, what did you mean by reordering the method?  I see that I tried to give it as much time to complete (I think I put it right before `onCreate` ends). I do see what you're pointing out on the profile so I'll try to change it up and add the encapsulation too.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -253,9 +265,20 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
...
 
+        waitForNavGraphInflation(navHost.navController)
+
         StartupTimeline.onActivityCreateEndHome(this) // DO NOT MOVE ANYTHING BELOW HERE.
     }
 
+    /**
+     * This job inflates the NavGraph and attaches it to our NavController.
+     */
+    private fun createNavGraphAsync() = MainScope().launch(Dispatchers.Default) {
```
mcomella, line 276: Why is this on `MainScope()`? I don't think it adds anything vs. `GlobalScope.launch`. Is it so we can cancel the job? In that case, I think we need to manually add code to cancel `MainScope()`, according to the docs: https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-main-scope.html

app/src/main/java/org/mozilla/fenix/perf/PerfNavController.kt
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * We need to keep a weak reference to our NavController in case our activity
+ * gets re-created and we need to link our NavGraph to the NavController again.
+ * The value of this map is the job associated to the NavController which should link the NavGraph
+ * to our NavController.
+ */
+@VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+val map = WeakHashMap<NavController, Job>()
+
+fun addNavToMap(navController: NavController, job: Job) {
```
mcomella, line 21: This code is hard to follow because implementation details are spread between two files: `HomeActivity` shouldn't handle the responsibility of adding an async nav graph job to a WeakHashMap.

Can we encapsulate this functionality in one file so it's easier to follow? For example,
```kotlin
object AsyncNavGraphProvider {
    fun startAsyncInflation(navController: NavController) { ... }
    fun blockForInflation(navController: NavController) { ... }    
}
```

As a bonus, this will make commenting what we're doing easier because right now, I think the commenting isn't clear enough.

app/src/main/java/org/mozilla/fenix/perf/PerfNavController.kt
```diff
@@ -0,0 +1,37 @@
...
+    map[navController] = job
+}
+
+/**
+ * The job should block the main thread if it isn't completed so that the NavGraph can be loaded
+ * before any navigation is done.
+ */
+fun waitForNavGraphInflation(navController: NavController) {
```
mcomella, line 29: nit: -> `blockForNavGraphInflation`

Let's use the JVM jargon and be explicit we're blocking

app/src/main/java/org/mozilla/fenix/perf/PerfNavController.kt
```diff
@@ -0,0 +1,37 @@
...
+    map.getValue(navController).also {
+        runBlockingIncrement {
+            if (!it.isCompleted) {
```
mcomella, line 32: afaict, this line is unnecessary: `join()` is conceptually just checking for this predicate in a `while` loop.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -144,6 +150,10 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
 
     private var inflater: LayoutInflater? = null
 
+    private lateinit var navGraphJob: Deferred<NavGraph>
```
mcomella, line 153: nit: this variable and the one below it are unused

app/src/test/java/org/mozilla/fenix/perf/PerfNavControllerTest.kt
```diff
@@ -0,0 +1,31 @@
...
+    fun setup() {
+        map.clear()
+    }
+
+    @Test
+    fun `Given a NavController is removed, it should not exist within the map anymore`() {
+        var navControllerOne: NavController? = NavController(testContext)
+        map[navControllerOne] = Job()
+        navControllerOne = null
+        assertNull(map[navControllerOne])
```
mcomella, line 29: This is saying, `null == map[null]`, which (assuming map[null] returns null) is guaranteed to be true based on the implmenetation of WeakHashMap.

Even if it was implemented correctly, I don't think this test is very meaningful – it's testing the functionality of WeakHashMap, not the code we wrote. I wonder, are there tests we should be writing for our code? Async code is hard so I feel like it should be tested. fwiw, if the main code related to this was all in one file (as I suggest above), it might be easier to see that it needs testing.

app/src/main/java/org/mozilla/fenix/ext/NavController.kt
```diff
@@ -10,19 +10,38 @@ import androidx.navigation.NavDirections
...
+    waitForNavGraphInflation(this)
+
     if (id == null || this.currentDestination?.id == id) {
         this.navigate(directions, navOptions)
     } else {
         recordIdException(this.currentDestination?.id, id)
     }
 }
 
+fun NavController.loadNavGraphBeforeNavigate(resId: Int) {
```
mcomella, line 30: Do we have a lint rule to make sure someone doesn't use `navigate` instead?

mcomella, line 30: nit: this is a good descriptive name but when naming something that we're overloading, I prefer to postfix the name because I think it's more likely someone will type in the old method name, notice the new one in auto-complete and use it instead. Here are some (maybe) bad naming ideas:
- `navigateFenix`
- `navigateBlockingForAsyncNavGraph`
- `navigateWithNavGraph`

This doesn't need to change but I think it'd be more intuitive.

app/src/main/java/org/mozilla/fenix/perf/PerfNavController.kt
```diff
@@ -0,0 +1,37 @@
...
+fun addNavToMap(navController: NavController, job: Job) {
+    map[navController] = job
+}
+
+/**
+ * The job should block the main thread if it isn't completed so that the NavGraph can be loaded
+ * before any navigation is done.
+ */
+fun waitForNavGraphInflation(navController: NavController) {
+    map.getValue(navController).also {
```
mcomella, line 30: If the navController is missing (which is unexpected), we'll throw `java.util.NoSuchElementException: Key NavController(navController#32) is missing in the map.`. To someone unfamiliar with this code, it's unintuitive. The underlying problem (we think) is that we didn't actually start the async nav graph inflation. Can we catch this error before `map.getValue()` throwsn and make it more explicit?

app/src/test/java/org/mozilla/fenix/components/toolbar/DefaultBrowserToolbarControllerTest.kt
```diff
@@ -100,6 +104,11 @@ class DefaultBrowserToolbarControllerTest {
     fun setUp() {
         MockKAnnotations.init(this)
 
+        mockkStatic("org.mozilla.fenix.perf.PerfNavControllerKt")
+        every { waitForNavGraphInflation(any()) } returns Unit
```
mcomella, line 108: I'm not sure this is the most correct solution – it's essentially saying, "this method is problematic so we stop calling it" by using mocking, which is a fairly fragile API and is decoupled from the actual implementation of the code.

For example, a more robust solution might inject a testing-only implementation of some class where we don't look in the weak map since we know we're in testing mode and it's not expected that we'd run that.

I don't know if there is a better solution but I wanted to highlight it as something to think about. In general, if "how can I test this code" and "how do I prevent unrelated tests from breaking" is an after thought, it's a smell that there might be more robust ways to design the code, which usually pushes you towards better encapsulation.

mcomella (changes requested):
The implementation looks like it should work but I think there are some major things we need to clean up:
- the code is hard to follow because it's spread around and not thoroughly commented. It's important to make the impl simple for async code
- we're duplicating a lot of code in our tests

Also, I didn't have time to test this myself and actually measure the perf improvement but that's something I'd like to do next time (maybe it's something I should have done before to make sure we're actually making gains here before we invest more time... ah well).

app/src/test/java/org/mozilla/fenix/home/DefaultSessionControlControllerTest.kt
```diff
@@ -106,6 +111,9 @@ class DefaultSessionControlControllerTest {
 
     @Before
     fun setup() {
+
+        mockkObject(NavGraphProvider)
```
mcomella, line 115: I see: we need to duplicate this high level logic ("stub the blocking method") because this isn't a Robolectric test so it isn't handled globally.

However, I think we can do better with the implementation. I'm concerned with:
- duplicating lines of code in many tests that can change in the future
- making mistakes in duplication (e.g. we didn't call `unmockkObject` here)
- devs getting an error they can't interpret when they call `navigateBlockingForAsyncNavGraph` in tests (when the solution is to add these two mysterious lines of code)

I think I have an idea: since it's discouraged to modify a global "runner"/"application"/"superclass test class" in JUnit (like we did with robolectric), let's use JUnit Rules – that will mostly solve problems 1 & 2. And when we throw the exception, we can add a line to it, "If this exception is thrown during testing, please add the `WhateverWeCallItRule`", which solves problem 3, the discoverability issue.

What do you think?

I only skimmed it but this looks like a good article on Rules:  https://ncorti.com/blog/junit-rules

mcomella (commented):
I had an idea on how we could improve the test code: I did an incomplete review so we can get started with it

app/src/main/java/org/mozilla/fenix/perf/NavGraphProvider.kt
```diff
@@ -0,0 +1,55 @@
...
+import kotlinx.coroutines.MainScope
+import kotlinx.coroutines.launch
+import org.mozilla.fenix.R
+import java.util.WeakHashMap
+
+/**
+ * This class asynchronously loads the Navigation xml. It also keeps track of the NavController
+ * to which the inflated NavGraph belongs too.
+ */
+object NavGraphProvider {
```
mcomella, line 22: I wonder, are there any useful tests we can write for this class?

app/src/main/java/org/mozilla/fenix/ext/NavController.kt
```diff
@@ -10,19 +10,38 @@ import androidx.navigation.NavDirections
...
     if (id == null || this.currentDestination?.id == id) {
         this.navigate(directions, navOptions)
     } else {
         recordIdException(this.currentDestination?.id, id)
     }
 }
 
+fun NavController.navigateBlockingForAsyncNavGraph(resId: Int) {
+    NavGraphProvider.blockForNavGraphInflation(this)
+    this.navigate(resId)
```
mcomella, line 32: I think we need the lint rule before we can land this.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -144,7 +146,8 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
 
     private var inflater: LayoutInflater? = null
 
-    private val navHost by lazy {
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    val navHost by lazy {
```
mcomella, line 150: nit: I think this can go back to private

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -186,7 +189,16 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
         components.publicSuffixList.prefetch()
 
         setupThemeAndBrowsingMode(getModeFromIntentOrLastKnown(intent))
-        setContentView(R.layout.activity_home)
+        setContentView(R.layout.activity_home).run {
+            //The NavGraph is being asynchronously inflated for performance gain.
+            //However, it has to be called immediately after the setContentView since it requires
```
mcomella, line 194: Can this, "it needs to be called because of thesve reasons" go into the kdoc for the class? The reasons aren't really appropriate for inline code. WHat we need to convey in inline code is, "This function needs to be called RIGHT HERE and SHOULD NOT BE MOVED. See kdoc for deteails"

mcomella (commented):
Looks functional but I think we can improve the commenting. I'm concerned the clean up code is incorrect.

app/src/test/java/org/mozilla/fenix/NavGraphTestRule.kt
```diff
@@ -0,0 +1,24 @@
...
+package org.mozilla.fenix
+
+import io.mockk.every
+import io.mockk.mockkObject
+import io.mockk.unmockkObject
+import org.junit.rules.TestWatcher
+import org.junit.runner.Description
+import org.mozilla.fenix.perf.NavGraphProvider
+
+class NavGraphTestRule : TestWatcher() {
```
mcomella, line 14: Add kdoc. Can we explain why we need to do this?

app/src/main/java/org/mozilla/fenix/perf/NavGraphProvider.kt
```diff
@@ -0,0 +1,49 @@
...
+    /**
+     * The job should block the main thread if it isn't completed so that the NavGraph can be loaded
+     * before any navigation is done.
+     *
+     * InflateNavGraphAsync must be called before this method.
+     */
+    fun blockForNavGraphInflation(navController: NavController) {
+        val inflationJob = map[navController] ?: throw IllegalStateException("Expected " +
+                "`NavGraphProvider.inflateNavGraphAsync` to be called before this method with the" +
+                " same `navController`. If it was, this may be a bug in the implementation.")
```
mcomella, line 46: Can we tell users that if this occurs in non-robolectric unit tests, they should add the test rule?

app/src/main/java/org/mozilla/fenix/perf/NavGraphProvider.kt
```diff
@@ -0,0 +1,49 @@
...
+import androidx.lifecycle.LifecycleCoroutineScope
+import androidx.navigation.NavController
+import java.util.WeakHashMap
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.launch
+import org.mozilla.fenix.R
+
+/**
+ * This class asynchronously loads the Navigation xml.
```
mcomella, line 16: Can we explain why we asynchronously load the navigation XML? i.e. for performance benefits

It's convey to say why a class exists so if, at some point, other devs can determine if it's no longer necessary and remove it.

app/src/main/java/org/mozilla/fenix/perf/NavGraphProvider.kt
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * This class is defined as an Object since it needs to be called by the NavController extension
+ * function.
+ *
+ * To use this class properly, NavGraphProvider.inflateNavGraphAsync must be called first before
+ * blockForNavGraphInflation using the same navController.
+ */
+object NavGraphProvider {
+
+    private val map = WeakHashMap<NavController, Job>()
```
mcomella, line 26: nit: can we explain why we're using a weak hash map?

app/src/test/java/org/mozilla/fenix/NavGraphTestRule.kt
mcomella, line 14: nit: I feel like this could have a more descriptive name, specifically to convey what it's doing. Maybe `DisableAsyncNavGraphAssertionRule` or `DisableNavGraphAssertionRule`?

app/src/test/java/org/mozilla/fenix/helpers/FenixRobolectricTestApplication.kt
```diff
@@ -37,4 +43,20 @@ class FenixRobolectricTestApplication : FenixApplication() {
...
+    // Beforetest runs before the test class is initialized
+    override fun beforeTest(method: Method?) {}
+
+    // Prepare test runs once the test class  and all its member variables (mock and
+    // non mocks) are initialized. Setting up our mocks here makes more sense since
+    // everything is available to us.
+    // This method runs after application.onCreate
+    override fun prepareTest(test: Any?) {
+        mockkObject(NavGraphProvider)
+        every { NavGraphProvider.blockForNavGraphInflation(any()) } returns Unit
```
mcomella, line 56: Can we share this setUp and tearDown code with the `NavGraphTestRule`?

app/src/test/java/org/mozilla/fenix/crashes/CrashReporterControllerTest.kt
```diff
@@ -16,64 +16,65 @@ import org.junit.Test
 import org.mozilla.fenix.R
 import org.mozilla.fenix.components.Components
 import org.mozilla.fenix.components.metrics.Event
+
 import org.mozilla.fenix.utils.Settings
 
 class CrashReporterControllerTest {
```
mcomella, line 22: This test fails: I think you need to add the NavGraphTestRule.

While we're modifying this file, I think we should avoid the churn of renaming the `navController`: this is already a big change. Can you change it back?

app/src/test/java/org/mozilla/fenix/components/toolbar/DefaultBrowserToolbarControllerTest.kt
```diff
@@ -249,7 +249,7 @@ class DefaultBrowserToolbarControllerTest {
         controller.handleTabCounterItemInteraction(item)
         verify {
             homeViewModel.sessionToDelete = "1"
-            navController.navigate(BrowserFragmentDirections.actionGlobalHome())
+            navController.navigateBlockingForAsyncNavGraph(BrowserFragmentDirections.actionGlobalHome())
```
mcomella, line 252: I renamed `navigateBlockingForAsync` back to `navigate` for the test files and I don't get additional failures: why are we renaming all these instances in the tests? It's already a large change so I'd prefer if we didn't make all these changes unnecessarily.

The commands I ran to change these quickly were (make sure to commit before doing it):
```sh
find app/src/test -type f -exec sed -i "" "s/navigateBlockingForAsyncNavGraph/navigate/g" {} \;
find app/src/test -type f -exec sed -i '' "/import org.mozilla.fenix.ext.navigate/d" {} \;
```

You'll need to uncommit the change to the `*MockMaker` file too

mcomella (commented):
Partial review: overall looks functional but I think we could improve the comments and some naming. Also, it's a large change so I'd prefer if we could avoid some churn in the tests. There are also test and lint failures so it'd be great to address those.

mcomella (commented):
The improvement looks good locally:
- MAIN is a 29ms improvement (median)
- VIEW is a 148ms (!!!) improvement (median)

Measured with that new script (to first frame), 25 iterations.

mcomella:
>     * VIEW is a 148ms (!!!) improvement (median)

Apparently this patch fixes the issue where `HomeFragment` gets called during VIEW (though I suspect it may be a race). Nice!

Here's are some profiles:
- Before the change: https://share.firefox.dev/3qIvESU
- After the change: https://share.firefox.dev/3bHAXy0

mozilla-detekt-rules/src/main/java/org/mozilla/fenix/detektrules/perf/MozillaNavigateCheck.kt
```diff
@@ -0,0 +1,51 @@
...
+
+
+    override fun visitCallExpression(expression: KtCallExpression) {
+        super.visitCallExpression(expression)
+
+        val calledMethod = expression.calleeExpression?.firstChild?.node?.chars
+
+        //We check for the navigate method and we have to ignore our extension function file, since
+        //we call navigate there
+        if (calledMethod == "navigate" && expression.containingFile.name != "NavController.kt") {
```
mcomella, line 42: We should suppress this with `@Suppress` in the file instead because it's more robust: it's tightly coupled to the file so we won't accidentally suppress in a file that happened to be named the same thing and it'll survive file renames.

mcomella:
I've been addressing the nits. To finish the review + fixing nits, I still have to:
- fix MozillaNavigateCheck – it doesn't seem to be working correctly
- try the app and see if I can break it

mcomella:
I opened https://github.com/mozilla-mobile/fenix/pull/18467 to replace this with my nits.

PR #16909, opened 2020-12-08
Closes #16896: Integrate new MediaSession API to nightly or debug builds

### LINKED ISSUE #16896: Integrate new MediaSession API
Media playback and notification issues.

GeckoView have integrated media fixes with a new `MediaSession` API.

### Who will benefit from it?
Users who have issues with Media notification or playback.

rocketsroger:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/Core.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/Core.kt
@@ -39,7 +39,6 @@ import mozilla.components.concept.fetch.Client
 import mozilla.components.feature.customtabs.store.CustomTabsServiceStore
 import mozilla.components.feature.downloads.DownloadMiddleware
 import mozilla.components.feature.logins.exceptions.LoginExceptionStorage
-import mozilla.components.feature.media.middleware.MediaMiddleware
 import mozilla.components.feature.media.middleware.RecordingDevicesMiddleware
 import mozilla.components.feature.pwa.ManifestStorage
 import mozilla.components.feature.pwa.WebAppShortcutManager
@@ -74,7 +73,6 @@ import org.mozilla.fenix.downloads.DownloadService
 import org.mozilla.fenix.ext.components
 import org.mozilla.fenix.ext.settings
 import org.mozilla.fenix.perf.lazyMonitored
-import org.mozilla.fenix.media.MediaService
 import org.mozilla.fenix.search.telemetry.ads.AdsTelemetry
 import org.mozilla.fenix.search.telemetry.incontent.InContentTelemetry
 import org.mozilla.fenix.settings.SupportUtils
@@ -82,6 +80,11 @@ import org.mozilla.fenix.settings.advanced.getSelectedLocale
 import org.mozilla.fenix.utils.Mockable
 import org.mozilla.fenix.utils.getUndoDelay
 import java.util.concurrent.TimeUnit
+import mozilla.components.feature.media.MediaSessionFeature
+import mozilla.components.feature.media.middleware.MediaMiddleware
+import org.mozilla.fenix.FeatureFlags.newMediaSessionApi
+import org.mozilla.fenix.media.MediaService
+import org.mozilla.fenix.media.MediaSessionService
 
 /**
  * Component group for all core browser functionality.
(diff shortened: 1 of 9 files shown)


app/src/main/java/org/mozilla/fenix/components/Core.kt
```diff
@@ -179,10 +181,9 @@ class Core(
      * The [BrowserStore] holds the global [BrowserState].
      */
     val store by lazyMonitored {
-        BrowserStore(
-            middleware = listOf(
+        var middlewareList =
```
pocmo, line 184: nit: This is a `var` but you you do seem to manipulate this reference?

app/src/main/java/org/mozilla/fenix/components/Core.kt
```diff
@@ -179,10 +181,9 @@ class Core(
...
+            listOf(
```
pocmo, line 185: nit: We could consider turning this into a `mutableListOf()` and then change the contents of the list instead of creating two different `BrowserStore`, which is a little bit confusing to read at first.

app/src/main/java/org/mozilla/fenix/tabtray/TabTrayViewHolder.kt
```diff
@@ -94,49 +98,98 @@ class TabTrayViewHolder(
...
+                            context.getString(R.string.mozac_feature_media_notification_action_pause)
+                        setImageDrawable(
+                            AppCompatResources.getDrawable(context, R.drawable.media_state_pause)
+                        )
+                    }
+
+                    else -> {
+                        removeTouchDelegate()
+                        removeAndDisable()
+                    }
```
pocmo, line 128: Oh, ugh, I didn't think of how much work it would be to separate the two implementations in the UI layers.

app/src/main/java/org/mozilla/fenix/tabtray/TabTrayViewHolder.kt
```diff
@@ -94,49 +98,98 @@ class TabTrayViewHolder(
...
-                }
 
-                MediaState.State.PLAYING -> {
-                    showAndEnable()
-                    contentDescription =
-                        context.getString(R.string.mozac_feature_media_notification_action_pause)
-                    setImageDrawable(
-                        AppCompatResources.getDrawable(context, R.drawable.media_state_pause)
-                    )
+        if (Config.channel.isNightlyOrDebug) {
```
pocmo, line 102: It may be easier to introduce a flag in `FeatureFlags` and use that instead:

With that it will be easier to, for example, enable this feature for "beta" without having to change this everywhere. Also it will be easier to find all the code once we can get rid of the flag. :)

pocmo (approved):

app/src/main/java/org/mozilla/fenix/components/Core.kt
rocketsroger: Oops, my original implementation needed var.  But it should be val now.  I'll fix, thanks,

app/src/main/java/org/mozilla/fenix/components/Core.kt
rocketsroger: Sure, I'll use a mutable list.  Thanks,

app/src/main/java/org/mozilla/fenix/tabtray/TabTrayViewHolder.kt
rocketsroger: Yeah :).  But once we're confident with the new MediaSession API changes, I can remove all of this and use a much cleaner implementation.

app/src/main/java/org/mozilla/fenix/tabtray/TabTrayViewHolder.kt
rocketsroger: Sounds good, I'll introduce a feature flag.

app/src/main/java/org/mozilla/fenix/components/Core.kt
rocketsroger: I'll use a mutable list as suggested.

PR #16932, opened 2020-12-09
fix relpro schema

escapewindow:
The description needs to be a string. We accidentally made it a tuple.


jmaher (approved):
oh nice!

JohanLorenzo (approved):
Nice catch!

PR #16951, opened 2020-12-09
Prepare for A-C Session[Manager] observable deprecation

csadilek:
We're now ready to deprecate `SessionManager` and `Session` observables in A-C. There are only four `Session` observers left in Fenix where we need to suppress the warning. There's no more `SessionManager` observer!  🎉 

I've filed and linked to tickets for the remaining observers.

This will be needed once https://github.com/mozilla-mobile/android-components/pull/9175 lands.


pocmo:
Marked the AC PR for landing and will trigger an AC Nightly + Add the version bump here.

pocmo:
Pushed rebased version with latest AC Nightly.

pocmo (approved):

PR #16961, opened 2020-12-10
Set version to 84.1.0 because of bug in shipit

JohanLorenzo:
Original root cause was fixed in https://github.com/mozilla-releng/shipit/pull/550


pocmo (approved):

st3fan (approved):
LGTM

Issue #15181, opened 2020-09-18
Clear downloads log and add it to clear browsing data checklist 


Nirose:
### What is the user problem or growth opportunity you want to see solved?

The download logs cannot be cleared from the Downloads screen. nor it is listed in 'delete browsing data' checklist
 
### How do you know that this problem exists today? Why is this important?
it kind of feels like a basic browser feature. clearing a log of  what was downloaded. 

### Who will benefit from it?
people who want to keep their download logs empty.

nigelh747:
Surely this is a bug as downloads used to be deleted as part of the clear data options

ekager:
This feature request is valid but would be blocked on #352 and we can look at this after that is completed

liuche:
As a first step, we could delete the list of files that Fenix saves when downloading. The copy would have to be clear though, about what is being deleted (e.g. the download log vs the downloads themselves) and we would need to decide whether we want to do one way or the other, or one and then the other.

Amejia481:
> As a first step, we could delete the list of files that Fenix saves when downloading. The copy would have to be clear though, about what is being deleted (e.g. the download log vs the downloads themselves) and we would need to decide whether we want to do one way or the other, or one and then the other.

This can be done using the [DownloadUseCase](https://github.com/mozilla-mobile/android-components/blob/ab29a90c889fa5a357974582e53c48d41e81d9fc/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadsUseCases.kt#L72) class e.g: `usecase.removeAllDownloads()`

mcarare:
This does not look like it's ready to be worked on. The button for clearing downloads already exists in the Downloads list. 
We still need strings if we are to add this option to the "delete browsing data" feature. 
Also, I think the distinction is not clear between deleting a Downloads list entry and deleting the downloaded item itself.

https://github.com/mozilla-mobile/fenix/issues/352 asks for the file to be deleted, but on testing this, the file is not deleted, just the entry in Fenix Download list is removed.

eliserichards:
@betsymi might have some input on the strings here?? 🤔

betsymi:
To clarify, it appears the files aren't actually being deleted. In that case, I wouldn't recommend we use the word `Delete` and use `Clear` or `Remove` instead. We also wouldn't have a destructive delete button either, as that signifies the user is deleting data, which isn't the case. This is something @apbitner and I are working on, new specs coming soon!

[...]

PR #16971, opened 2020-12-10
For #15181: Remove All Downloads From Settings 

kglazko:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/settings/deletebrowsingdata/DeleteAndQuit.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/deletebrowsingdata/DeleteAndQuit.kt
@@ -24,6 +24,7 @@ fun deleteAndQuit(activity: Activity, coroutineScope: CoroutineScope, snackbar: 
         val settings = activity.settings()
         val controller = DefaultDeleteBrowsingDataController(
             activity.components.useCases.tabsUseCases.removeAllTabs,
+            activity.components.useCases.downloadUseCases.removeAllDownloads,
             activity.components.core.historyStorage,
             activity.components.core.permissionStorage,
             activity.components.core.store,
@@ -61,5 +62,6 @@ private suspend fun DeleteBrowsingDataController.deleteType(type: DeleteBrowsing
         DeleteBrowsingDataOnQuitType.PERMISSIONS -> withContext(IO) {
             deleteSitePermissions()
         }
+        DeleteBrowsingDataOnQuitType.DOWNLOADS -> deleteDownloads()
     }
 }
--- a/app/src/main/java/org/mozilla/fenix/settings/deletebrowsingdata/DeleteBrowsingDataFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/deletebrowsingdata/DeleteBrowsingDataFragment.kt
@@ -33,7 +33,7 @@ import org.mozilla.fenix.ext.settings
 import org.mozilla.fenix.ext.showToolbar
 import org.mozilla.fenix.utils.Settings
 
-@SuppressWarnings("TooManyFunctions")
+@SuppressWarnings("TooManyFunctions", "LargeClass")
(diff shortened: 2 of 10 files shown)


app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -527,6 +527,11 @@ class Settings(private val appContext: Context) : PreferencesHolder {
         default = true
     )
 
+    var deleteDownloads by booleanPreference(
+        appContext.getPreferenceKey(R.string.pref_key_delete_downloads_now),
+        default = true
+    )
```
Amejia481, line 533: Let's add tests to `SettingsTest` to verify this new behaviour :)

app/src/main/java/org/mozilla/fenix/settings/deletebrowsingdata/DeleteAndQuit.kt
```diff
@@ -61,5 +62,6 @@ private suspend fun DeleteBrowsingDataController.deleteType(type: DeleteBrowsing
         DeleteBrowsingDataOnQuitType.PERMISSIONS -> withContext(IO) {
             deleteSitePermissions()
         }
+        DeleteBrowsingDataOnQuitType.DOWNLOADS -> deleteDownloads()
```
Amejia481, line 65: Let's add test to `DeleteAndQuitTest` to verify this new behaviour :)

Amejia481 (changes requested):
It's looking good lets add some test to avoid future regressions. 👍🏽 
Just to confirm, do this functionality also have to show in the "Delete Browsing data on quit"  fragment? As I see that we are modifying the `deleteAndQuit` function but I don't it on the ui.

Amejia481:
> It's looking good lets add some test to avoid future regressions. 👍🏽
> Just to confirm, do this functionality also have to show in the "Delete Browsing data on quit" fragment? As I see that we are modifying the `deleteAndQuit` function but I don't it on the ui.

It looks like we have the same entries: 

screenshot-1607639238028

screenshot-1607639292995

app/src/main/java/org/mozilla/fenix/settings/deletebrowsingdata/DeleteBrowsingDataFragment.kt
```diff
@@ -45,6 +45,7 @@ class DeleteBrowsingDataFragment : Fragment(R.layout.fragment_delete_browsing_da
 
         controller = DefaultDeleteBrowsingDataController(
             requireComponents.useCases.tabsUseCases.removeAllTabs,
+            requireComponents.useCases.downloadUseCases.removeAllDownloads,
```
eliserichards, line 48: (This is such a small nit)
Could we make a val for `requireComponents.useCases` and `requireComponents.core` so we can just say `useCases.downloadUseCases.[...]` or `core.store`?

app/src/main/res/values/strings.xml
```diff
@@ -1066,6 +1066,8 @@
     <string name="preferences_delete_browsing_data_cached_files_subtitle">Frees up storage space</string>
     <!-- Title for the site permissions item in Delete browsing data -->
     <string name="preferences_delete_browsing_data_site_permissions">Site permissions</string>
+    <!-- Title for the downloads item in Delete browsing data -->
```
eliserichards, line 1069: Could we change this description a bit? Proposal: Title for the deleting downloads item in Delete browsing data 

Update: oh strings here are all following that pattern.. we should try to change those at some point to make them a bit more descriptive 😄

eliserichards (changes requested):
This looks awesome! Just two small nits and the tests Arturo mentioned, and this is good to go! 🚀

Amejia481 (approved):

eliserichards (approved):
This looks great! You did an awesome job 🎉 🎉 🎉

PR #17015, opened 2020-12-14
Add search current/home region to debug screen.

pocmo:
Saw that we added a screen for debug info. This seems to be a good place for exposing the search current/home region. :)


rocketsroger (approved):
🚢

PR #17033, opened 2020-12-14
No bug- protect nimbus init from StrictMode

jhugman:
As per conversation with @travis79 and @csadilek , this PR ensures that reads at Nimbus init time is not caught by StrictMode.

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/Analytics.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/Analytics.kt
@@ -9,6 +9,7 @@ import android.app.PendingIntent
 import android.content.Context
 import android.content.Intent
 import android.net.Uri
+import android.os.StrictMode
 import mozilla.components.lib.crash.CrashReporter
 import mozilla.components.lib.crash.service.CrashReporterService
 import mozilla.components.lib.crash.service.GleanCrashReporterService
@@ -117,12 +118,15 @@ class Analytics(
                 // from the app unless the user does so from a UI control.
                 // However, the user may have opt-ed out of mako experiments already, so
                 // we should respect that setting here.
-                val enabled = context.settings().isExperimentationEnabled
+                val enabled = context.components.strictMode.resetAfter(StrictMode.allowThreadDiskReads()) {
+                    context.settings().isExperimentationEnabled
+                }
                 if (!enabled) {
                     globalUserParticipation = enabled
                 }
-                context.settings().isExperimentationEnabled = globalUserParticipation
-
+            }
+        }.apply {
+            if (FeatureFlags.nimbusExperiments) {
                 // Nimbus should look after downloading experiment definitions from remote settings
                 // on another thread, and making sure we don't hit the server each time we start.
                 updateExperiments()


app/src/main/java/org/mozilla/fenix/components/Analytics.kt
```diff
@@ -117,12 +118,15 @@ class Analytics(
...
                 // However, the user may have opt-ed out of mako experiments already, so
                 // we should respect that setting here.
-                val enabled = context.settings().isExperimentationEnabled
+                val enabled = context.components.strictMode.resetAfter(StrictMode.allowThreadDiskReads()) {
+                    context.settings().isExperimentationEnabled
+                }
                 if (!enabled) {
                     globalUserParticipation = enabled
                 }
-                context.settings().isExperimentationEnabled = globalUserParticipation
```
jhugman, line 124: Once this is set first time, this will be kept in sync, so we don't need this read here.

app/src/main/java/org/mozilla/fenix/components/Analytics.kt
```diff
@@ -117,12 +118,15 @@ class Analytics(
...
-
+            }
+        }.apply {
+            if (FeatureFlags.nimbusExperiments) {
```
jhugman, line 129: I broke this up, just to keep things clean.

app/src/main/java/org/mozilla/fenix/experiments/Experiments.kt
```diff
@@ -15,7 +15,7 @@ class ExperimentBranch {
     companion object {
         const val TREATMENT = "treatment"
         const val CONTROL = "control"
-        const val A1 = "A1"
-        const val A2 = "A2"
+        const val A1 = "a1"
+        const val A2 = "a2"
```
travis79, line 19: I was able to get them to change this on the server, this is not needed :)

travis79 (changes requested):

travis79 (approved):

app/src/main/java/org/mozilla/fenix/components/Analytics.kt
```diff
@@ -117,12 +118,15 @@ class Analytics(
                 // from the app unless the user does so from a UI control.
                 // However, the user may have opt-ed out of mako experiments already, so
                 // we should respect that setting here.
-                val enabled = context.settings().isExperimentationEnabled
+                val enabled = context.components.strictMode.resetAfter(StrictMode.allowThreadDiskReads()) {
```
csadilek, line 121: I tried reproducing this strict mode violation but can't, even if I execute this at the beginning of `Application.onCreate`. Interesting that the stacktrace we saw doesn't make much sense as we see a call from this lambda directly to `listFiles`, which doesn't exist: https://gist.github.com/travis79/88779741ced2bddf03fe68797c3faa8f

Either way this is OK to land. We're not worried about `context.settings`.

app/src/main/java/org/mozilla/fenix/components/Analytics.kt
csadilek: 👍 Since we now know that the Nimbus constructor is quite expensive (even though it doesn't trigger any IO), I'd like to suggest we make this `nullable` and only create `Nimbus` if we need it to address one of the performance issues. Alternative we could return a "NoopNimbus" instance (cheap to create) which would make the consuming code cleaner.

```kotlin
val experiments: Nimbus? by lazyMonitored {
    val url: String? = BuildConfig.NIMBUS_ENDPOINT
    val serverSettings = if (!url.isNullOrBlank()) {
        NimbusServerSettings(url = Uri.parse(url))
    } else {
        null
    }

    if (FeatureFlags.nimbusExperiments) {
        Nimbus(context, serverSettings).apply {
            val enabled = context.components.strictMode.resetAfter(StrictMode.allowThreadDiskReads()) {
                context.settings().isExperimentationEnabled
            }
            if (!enabled) {
                globalUserParticipation = enabled
            }
            // Nimbus should look after downloading experiment definitions from remote settings
            // on another thread, and making sure we don't hit the server each time we start.
            updateExperiments()
        }
    } else {
        null // or NoopNimbus() :)
    }
}
```

csadilek: This would buy us some time and we can also uplift to beta so we don't take a performance hit there....don't need to decide now.

csadilek (approved):

PR #17039, opened 2020-12-14
Closes #16946: Refactor PwaOnboardingObserver to use browser store

### LINKED ISSUE #16946: Refactor PwaOnboardingObserver to observe session via store
Similar to https://github.com/mozilla-mobile/fenix/issues/16945, `PwaOnboardingObserver` is currently a deprecated Session observer and only tracking `onLoadingStateChanged`.

Since we're in the progress of migrating to browser-state we need to refactor to observe these changes using the browser store instead.

csadilek:
`PwaOnboardingObserver` is meant to show an onboarding fragment on the third visit to an installable page. Turns out this isn't actually working on current master. The page finishes loading before the manifest is updated and therefore we miss to count the visit (`isInstallable` returns false if checked too early).

To reproduce: Open any installable page (e.g. twitter) and let it finish loading completely, open new tab repeat and let finish, repeat again and you should see the onboarding fragment.

This is fixed here and refactored to browser-state, plus unit tests. I kept `PwaOnboardingObserver` to keep this separate from the fragment. Also worth noting that we only need this for the selected tab in the main app.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
@@ -53,6 +53,7 @@ class BrowserFragment : BaseBrowserFragment(), UserInteractionHandler {
 
     private var readerModeAvailable = false
     private var openInAppOnboardingObserver: OpenInAppOnboardingObserver? = null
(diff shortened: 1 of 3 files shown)


pocmo (approved):
Awesome! 👍 

🎉

PR #17043, opened 2020-12-15
Closes #11413: Add a notification of privacy&rights in the first launch in China builds 

### LINKED ISSUE #11413: [Feature:China]Add a notification of privacy&rights in the first launch in China builds
China government requires all the apps to show privacy notification prominently during the first launch and  users can only use the apps normally after they click the consent button; otherwise if they click the disapproval button, the app should quit.

Here is what we do now in Fennec. Basically we use the content of _about:rights_.
图片

It is required by laws and regulations in China.

rxumoz:
Use alert dialog to present a notification.
To prevent users to use fenix without approving, use webview to present the details related. (I'm nor sure if it is appropriate to do so, please correct me if there is a more normative way to deploy the feature.)
Screenshot_20201215_133427_org mozilla firefox
Screenshot_20201215_133440_org mozilla firefox

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
@@ -119,6 +119,7 @@ import org.mozilla.fenix.utils.allowUndo
 import org.mozilla.fenix.whatsnew.WhatsNew
 import java.lang.ref.WeakReference
 import kotlin.math.min
+import org.mozilla.fenix.home.mozonline.showPrivacyPopWindow
 
 @ExperimentalCoroutinesApi
 @Suppress("TooManyFunctions", "LargeClass")
@@ -176,6 +177,10 @@ class HomeFragment : Fragment() {
             if (!onboarding.userHasBeenOnboarded()) {
                 requireComponents.analytics.metrics.track(Event.OpenedAppFirstRun)
             }
+        }
+
+        if (!onboarding.userHasBeenOnboarded() && requireContext().settings().shouldShowPrivacyPopWindow) {
+            showPrivacyPopWindow(requireContext(), requireActivity())
         }
     }
 
--- a/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
@@ -119,6 +119,7 @@ import org.mozilla.fenix.utils.allowUndo
 import org.mozilla.fenix.whatsnew.WhatsNew
 import java.lang.ref.WeakReference
 import kotlin.math.min
+import org.mozilla.fenix.Config
 import org.mozilla.fenix.home.mozonline.showPrivacyPopWindow
 
 @ExperimentalCoroutinesApi
@@ -179,7 +180,9 @@ class HomeFragment : Fragment() {
             }
         }
 
-        if (!onboarding.userHasBeenOnboarded() && requireContext().settings().shouldShowPrivacyPopWindow) {
+        if (!onboarding.userHasBeenOnboarded() &&
+            requireContext().settings().shouldShowPrivacyPopWindow &&
+            Config.channel.isMozillaOnline) {
             showPrivacyPopWindow(requireContext(), requireActivity())
         }
     }
--- a/app/src/main/java/org/mozilla/fenix/home/mozonline/PrivacyContentDisplayActivity.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/mozonline/PrivacyContentDisplayActivity.kt
@@ -1,0 +1,84 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.home.mozonline
+
+import android.app.Activity
+import android.content.Context
+import android.os.Bundle
+import android.util.AttributeSet
+import android.view.View
+import android.widget.ImageButton
+import mozilla.components.concept.engine.EngineSession
+import mozilla.components.concept.engine.EngineView
+import mozilla.components.feature.contextmenu.DefaultSelectionActionDelegate
+import mozilla.components.feature.search.BrowserStoreSearchAdapter
+import mozilla.components.support.ktx.android.content.call
+import mozilla.components.support.ktx.android.content.email
+import mozilla.components.support.ktx.android.content.share
+import org.mozilla.fenix.R
+import org.mozilla.fenix.ext.components
+
+/**
+ * A special activity for displaying the detail content about privacy hyperlinked in alert dialog.
+ */
+
+class PrivacyContentDisplayActivity : Activity(), EngineSession.Observer {
+    private lateinit var engineView: EngineView
+    private lateinit var closeButton: ImageButton
+    private lateinit var engineSession: EngineSession
+    private var url: String? = ""
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setContentView(R.layout.activity_privacy_content_display)
+        val addr = intent.extras
+        if (addr != null) {
+            url = addr.getString("url")
+        }
+
+        engineView = findViewById<View>(R.id.privacyContentEngineView) as EngineView
+        closeButton = findViewById<View>(R.id.privacyContentCloseButton) as ImageButton
+        engineSession = components.core.engine.createSession()
+    }
+
+    override fun onCreateView(
+        parent: View?,
+        name: String,
+        context: Context,
+        attrs: AttributeSet
+    ): View? = when (name) {
+        EngineView::class.java.name -> components.core.engine.createView(context, attrs).apply {
+            selectionActionDelegate = DefaultSelectionActionDelegate(
+                BrowserStoreSearchAdapter(
+                    components.core.store
+                ),
+                resources = context.resources,
+                shareTextClicked = { share(it) },
+                emailTextClicked = { email(it) },
+                callTextClicked = { call(it) }
+            )
+        }.asView()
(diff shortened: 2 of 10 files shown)


app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -177,6 +183,12 @@ class HomeFragment : Fragment() {
                 requireComponents.analytics.metrics.track(Event.OpenedAppFirstRun)
             }
         }
+
+        if (!onboarding.userHasBeenOnboarded() &&
+            requireContext().settings().shouldShowPrivacyPopWindow &&
+            Config.channel.isMozillaOnline) {
+            showPrivacyPopWindow()
```
Amejia481, line 190: Instead of creating a `webview` I think you can take advantage of the whole browser and load the links directly on it by using [openToBrowserAndLoad](https://github.com/mozilla-mobile/fenix/blob/d8f3127e0868c80e0e782ac6fab11462f5fbbc96/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt#L808) as we do to load other web content :)

rxumoz: Thanks! I've tried openToBrowserAndLoad but there are still two problems:
1. Clicking the hyper links make the tab opened behind the alert dialog:
Screenshot_20201216_095416_org mozilla fenix debu
2. If starting a new activity to open the tabs to avoid the alert dialog's shelter, users may change the url in the address bar to use fenix directly, which means they can use fenix without approving the privacy notice.

Amejia481: Fenix supports [custom tabs](https://developer.chrome.com/docs/multidevice/android/customtabs/), talking with @jonalmeida he shared this [example](https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/java/org/mozilla/fenix/settings/account/AuthIntentReceiverActivity.kt#L29-L34) where we used it from Fenix. Worst case scenario if custom tabs are not OK for your use case, you can embed geckoview see an example [here](https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/res/layout/fragment_add_on_internal_settings.xml)

rxumoz: Thanks, I will try it and update this pr.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -1035,6 +1047,42 @@ class HomeFragment : Fragment() {
         view?.sessionControlRecyclerView?.adapter?.notifyDataSetChanged()
     }
 
+    private fun showPrivacyPopWindow() {
```
jonalmeida, line 1050: Let's move this to a separate helper class so we add more things into this home fragment class.

app/src/main/java/org/mozilla/fenix/home/MultiClickableSpan.kt
```diff
@@ -0,0 +1,44 @@
...
+    const val POS3 = 3
+}
+
+object ADDR {
+    const val URL1 = "https://www.mozilla.org/en-US/MPL/"
+    const val URL2 = "https://www.mozilla.org/en-US/foundation/trademarks/policy/"
+    const val URL3 = "https://www.mozilla.org/zh-CN/privacy/firefox/"
+}
+
+class MultiClickableSpan(var pos: Int, var context: Context) :
```
jonalmeida, line 25: This name implies it's a generic clickable span, but it isn't. How about naming it to be more specific? Maybe `PrivacyContentSpan`?

Or we could make this generic so it's re-usable for other places, but that might be out of scope for now.

jonalmeida (commented):
This looks good!

I played around with it and I saw that using Custom Tabs doesn't seem to work as you like; it runs into the same issues of loading the page _below_ the dialog.

How about we move these new classes into an inner package since this is particular for Mozilla Online?

So moving `PrivacyContentDisplayActivity` and `MultiClickableSpan` to `app/src/main/java/org/mozilla/fenix/home/mozonline` perhaps?

app/src/main/java/org/mozilla/fenix/home/mozonline/PrivacyContentDisplayActivity.kt
```diff
@@ -0,0 +1,84 @@
...
+import org.mozilla.fenix.ext.components
+
+/**
+ * A special activity for displaying the detail content about privacy hyperlinked in alert dialog.
+ */
+
+class PrivacyContentDisplayActivity : Activity(), EngineSession.Observer {
+    private var engineView: EngineView? = null
+    private var closeButton: ImageButton? = null
+    private var engineSession: EngineSession? = null
```
Amejia481, line 30: nit: You could use [lateinit var](https://kotlinlang.org/docs/reference/properties.html#late-initialized-properties-and-variables) and avoid the null values :)

Amejia481 (approved):
Looks good, I left a comment for a nit, apart from that 👍🏽

rxumoz:
Thanks for your help @jonalmeida  and @Amejia481

rxumoz:
@Amejia481 @jonalmeida 
Hi hi, I am wondering that do I need to start a ci task or something to make progress in this PR? I'm just a little bit get lost here.
Another thing I want to discuss is that is there any way to keep this PR related strings just Chinese and English? My colleague thinks it seems a waste of time to translate them into so many languages and the users may never see this notice.

Amejia481:
> @Amejia481 @jonalmeida
> Hi hi, I am wondering that do I need to start a ci task or something to make progress in this PR? I'm just a little bit lost here.
> 
> Another thing I want to discuss is that is there any way to keep this PR related strings just Chinese and English? My colleague thinks it seems a waste of time to translate them into so many languages since their users may never see this notice. That is to say, how do I avoid these strings to be exported to l10n? maybe cc @Delphine

This normally happens when an external contributor opens a pr, @st3fan can check if you have the right permissions.

st3fan:
I definitely see @rxumoz in the Fenix team with the same Write permissions as other devs. 

I think the problem is that this PR is originating from a fork. I don't think we have a good solution for that currently.

It may be best to create the PR branch directly inside the fenix project here.

rxumoz:
So maybe I close this PR here and push a new branch to fenix project to start a new PR?

Amejia481:
Yes, that would be the best alternative

rxumoz:
@Amejia481 Okay I'll do it today, thx.
And do you have any idea about the string stuff I mentioned above?

Amejia481:
> @Amejia481 Okay I'll do it today, thx.
> And do you have any idea about the string stuff I mentioned above?

I think @Delphine will be the best person to answer but I noticed that we have [static_strings.xml](https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/res/values/static_strings.xml), I'm not sure if this is the right file to use, but reading the commits on this file could give you an idea

rxumoz:
> > @Amejia481 Okay I'll do it today, thx.
> > And do you have any idea about the string stuff I mentioned above?
> 
> I think @Delphine will be the best person to answer but I noticed that we have [static_strings.xml](https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/res/values/static_strings.xml), I'm not sure if this is the right file to use, but reading the commits on this file could give you an idea

Okay, I'll check it and also wait for @Delphine , thx!

Delphine:
@rxumoz Thanks for flagging! I see from the commit in issue #17043 that the strings have already been translated - correct? Are there any other strings maybe since then?
I agree we shouldn't expose this to all locales. The best would be to add the translation to your own fork/build, because if we expose this like "usual Fenix strings", 90+ other locales would be working on this for nothing

Delphine:
@rxumoz Thanks for flagging! I see from the commit in issue #17043 that the strings have already been translated - correct? Are there any other strings maybe since then?
I agree we shouldn't expose this to all locales. The best would be to add the translation to your own fork/build, because if we expose this like "usual Fenix strings", 90+ other locales would be working on this for nothing

rxumoz:
> @rxumoz Thanks for flagging! I see from the commit in this issue that the strings have already been translated - correct? Are there any other strings maybe since then?
> I agree we shouldn't expose this to all locales. The best would be to add the translation to your own fork/build, because if we expose this like "usual Fenix strings", 90+ other locales would be working on this for nothing

Yes I prepared Chinese and English and in this PR the content is basically copied from _about:rights_ .
Actually our target is to merge all the code we need into Fenix project and then build and release MozillaOnline builds along with global builds, so I'm afraid I could not keep these strings just in MozillaOnline fork. For now I have two thoughts:

1. Add both Chinese and English strings separately into static_strings.xml and mark _translatable=false_ but then we need to change the code in order to use the separate strings in different languages with the same meaning and make them change with language setting, I don't know if it's appropriate.

2. Or we can make a separate xml file (e.g. mozonline_strings.xml) only in _values_  and  _values-zh-rCN_ to record the strings only related Mozilla Online builds, so that the file wouldn't be exported to l10n ( I think only strings.xml would be exported to l 10n). 

With the experiment locally, I personally prefer the second choice.

I don't know if I should cc @st3fan ?

Delphine:
If the second option works that is great, but I admit I don't know what the best option would be at this point 😕 Maybe Stefan can help advise as well.

rxumoz:
@Pike Hi, sorry to bother you but there is a question I want to consult about l10n exporting. In Android projects (or specifically Fenix), is _**strings.xml**_ the only file that would be exported to l10n? which means other xml files with a different name containing strings would **NOT** be exported to l10n? 
For more information maybe you can check the discussions above between Delphine and me in this issue.
Thanks!

pocmo:
@rxumoz The import/export should only care about the files listed in l10n.toml:
https://github.com/mozilla-mobile/fenix/blob/master/l10n.toml#L110-L112

So your approach seems feasible.

rxumoz:
@pocmo Thanks, then I will prepare the PR.

PR #17048, opened 2020-12-15
Change the expected author for automated PRs to MickeyMoz

st3fan:
Change the expected author for automated PRs to MickeyMoz.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -32,7 +32,7 @@ pull_request_rules:
   - name: Release automation
     conditions:
       - base~=releases/.*
-      - author=st3fan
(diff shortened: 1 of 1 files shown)


st3fan:
@pocmo did I do this right?

pocmo:
@st3fan This needs to be the username of the account that creates the PR. If that is opened from a GitHub action then I think we need:
https://github.com/mozilla-mobile/android-components/blob/master/.mergify.yml#L34

st3fan:
@pocmo Yup that makes sense of course. Updated.

pocmo (approved):

PR #17019, opened 2020-12-14
Switch to new feature-tab-collections API.

pocmo:
This is in preparation for an upcoming change in AC:
https://github.com/mozilla-mobile/android-components/pull/9207

As a bonus `DefaultCollectionCreationController` no longer requires `SessionManager`.

PR #17076, opened 2020-12-16

MickeyMoz:


pocmo:
Update will happen in https://github.com/mozilla-mobile/fenix/pull/17019

PR #17082, opened 2020-12-16
Closes #16949: Refactor OpenInAppOnboardingObserver to use browser store

### LINKED ISSUE #16949: Refactor OpenInAppOnboardingObserver to observe session via store
Similar to #16945, `OpenInAppOnboardingObserver` is currently a deprecated Session observer and tracking `onUrlChanged` and `onLoadingStateChanged`

Since we're in the progress of migrating to browser-state we need to refactor to observe these changes using the browser store instead.

csadilek:
Follows the same pattern as https://github.com/mozilla-mobile/fenix/pull/17039 to refactor the observer. Logic is exactly the same as before, just migrated to use browser store, plus some test improvements.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
@@ -51,9 +51,9 @@ import org.mozilla.fenix.trackingprotection.TrackingProtectionOverlay
 class BrowserFragment : BaseBrowserFragment(), UserInteractionHandler {
 
     private val windowFeature = ViewBoundFeatureWrapper<WindowFeature>()
+    private val openInAppOnboardingObserver = ViewBoundFeatureWrapper<OpenInAppOnboardingObserver>()
 
     private var readerModeAvailable = false
-    private var openInAppOnboardingObserver: OpenInAppOnboardingObserver? = null
     private var pwaOnboardingObserver: PwaOnboardingObserver? = null
 
     override fun onCreateView(
@@ -142,6 +142,22 @@ class BrowserFragment : BaseBrowserFragment(), UserInteractionHandler {
                 owner = this,
                 view = view
             )
+
+            if (context.settings().shouldShowOpenInAppCfr) {
+                openInAppOnboardingObserver.set(
+                    feature = OpenInAppOnboardingObserver(
+                        context = context,
+                        store = context.components.core.store,
+                        lifecycleOwner = this,
+                        navController = findNavController(),
+                        settings = context.settings(),
+                        appLinksUseCases = context.components.useCases.appLinksUseCases,
+                        container = browserLayout as ViewGroup
+                    ),
+                    owner = this,
+                    view = view
+                )
+            }
         }
     }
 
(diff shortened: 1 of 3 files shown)


csadilek:
@codrut-topliceanu just a heads-up that this will conflict with https://github.com/mozilla-mobile/fenix/pull/17049. It's easy to fix though as we just switch to a different way to observe the session changes, everything else is the same. Happy to take care of it or help you in case this PR lands first!

app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
```diff
@@ -197,14 +194,12 @@ class BrowserFragment : BaseBrowserFragment(), UserInteractionHandler {
...
-            // TODO Use browser store instead of session observer: https://github.com/mozilla-mobile/fenix/issues/16949
-            getSessionById()?.unregister(openInAppOnboardingObserver!!)
-            openInAppOnboardingObserver = null
-        }
+
+        openInAppOnboardingObserver?.stop()
+        // This observer is initialized in onStart and holds a reference to
+        // fragment's view. Let's set it to null to prevent leaking the view
+        // after the onDestroyView.
+        openInAppOnboardingObserver = null
```
pocmo, line 202: I think this means we need to turn the observer into a `LifecycleFeature` and wrap in our `ViewBound...Thingy` to automatically clear it if the view gets detached? If I remember correctly, we saw the navigation library detach fragments (which destroys the view) and then later attach them again (creating new views) on some navigations and then we leaked the view during that detached state?

app/src/main/java/org/mozilla/fenix/browser/OpenInAppOnboardingObserver.kt
```diff
@@ -7,62 +7,99 @@ package org.mozilla.fenix.browser
...
+                    sessionDomainForDisplayedBanner?.let {
+                        if (tab.content.url.tryGetHostFromUrl() != it) {
+                            infoBanner?.dismiss()
+                        }
+                    }
+                    currentUrl = tab.content.url
+                } else {
+                    // Loading state has changed
+                    maybeShowOpenInAppBanner(tab.content.url, tab.content.loading)
+                }
```
pocmo, line 67: Can there be a state where both values changed (e.g. when the selected tab has changed) and would we care about this?

app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
pocmo: Afaik the previous code had this problem too, so I think we could also decide to handle this in a follow-up issue.

pocmo (approved):

app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
csadilek: Yes, thought about that too, but wanted to keep separate. I will give it a try and file a follow-up if too much.

app/src/main/java/org/mozilla/fenix/browser/OpenInAppOnboardingObserver.kt
csadilek: Yes, but we should be fine here, as we're only displaying the banner after a successful load. So, we want to show the banner after loading = true and loading = false happened. So if we switch to another tab with a different url we don't want to show the banner right away only if it finishes loading (if it does). If we switch to another tab with the exact same url it's fine to display the banner if needed and we haven't already displayed it for the previous tab.

app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
csadilek: OK, it was pretty easy and cleaned things up nicely. Updated.

app/src/main/java/org/mozilla/fenix/browser/OpenInAppOnboardingObserver.kt
csadilek: Tested this again and this actually works better than before. Now when we switch tabs to a page that can not be opened in an external app, we also don't show the banner. Before we'd also show the banner for sites where it didn't make sense if we switched tabs after the banner was displayed.

pocmo (approved):
💯

PR #17098, opened 2020-12-17
For #16615: UI smoke test noCrashWithAddonInstalledTest

sv-ohorvath:
For #16615: UI smoke test noCrashWithAddonInstalledTest
Ran the test 30 times on Firebase and passed.
Added a description to each smoke test.


isabelrios (approved):
LTGM! thanks for your hard work with the Smoketest suite

PR #17120, opened 2020-12-18
Switch to new SessionStorage / session restore API.

pocmo:
For upcoming AC changes: https://github.com/mozilla-mobile/android-components/pull/9250

The AC changes will land in 2021. :)

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
+++ b/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
@@ -41,14 +41,15 @@ import mozilla.components.support.webextensions.WebExtensionSupport
 import org.mozilla.fenix.components.Components
 import org.mozilla.fenix.components.metrics.MetricServiceType
 import org.mozilla.fenix.ext.settings
-import org.mozilla.fenix.perf.StorageStatsMetrics
 import org.mozilla.fenix.perf.StartupTimeline
+import org.mozilla.fenix.perf.StorageStatsMetrics
 import org.mozilla.fenix.perf.runBlockingIncrement
 import org.mozilla.fenix.push.PushFxaIntegration
 import org.mozilla.fenix.push.WebPushEngineIntegration
 import org.mozilla.fenix.session.PerformanceActivityLifecycleCallbacks
 import org.mozilla.fenix.session.VisibilityLifecycleCallback
 import org.mozilla.fenix.utils.BrowsersCache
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -159,6 +163,42 @@ open class FenixApplication : LocaleAwareApplication(), Provider {
...
+        components.useCases.tabsUseCases.restore(sessionStorage)
+
+        // Now that we have restored our previous state (if there's one) let's setup auto saving the state while
+        // the app is used.
+        sessionStorage.autoSave(store)
+            .periodicallyInForeground(interval = 30, unit = TimeUnit.SECONDS)
+            .whenGoingToBackground()
+            .whenSessionsChange()
+
+        // TODO: Move into use case?
```
csadilek, line 179: Yes, I think we should move this. Had left a comment about this on the A-C PR too as I remembered adding this here.

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -159,6 +163,42 @@ open class FenixApplication : LocaleAwareApplication(), Provider {
         components.appStartupTelemetry.onFenixApplicationOnCreate()
     }
 
+    private fun restoreBrowsingSession() = GlobalScope.launch(Dispatchers.Main) {
```
csadilek, line 166: nit: Naming suggestion in A-C PR, but let's keep consistent based on our decision.

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -159,6 +163,42 @@ open class FenixApplication : LocaleAwareApplication(), Provider {
...
+        sessionStorage.autoSave(store)
+            .periodicallyInForeground(interval = 30, unit = TimeUnit.SECONDS)
+            .whenGoingToBackground()
+            .whenSessionsChange()
+
+        // TODO: Move into use case?
+        store.dispatch(RestoreCompleteAction)
+    }
+
+    private fun removeTimedOutTabs() {
```
csadilek, line 183: Let's file a follow-up to cover this with a test. I didn't have one before it seems ...

pocmo:
Updated. And added latest AC Nightly version. This should compile now.

Issue #16829, opened 2020-12-03
[Bug] [a11y] Consider increasing the "Scan" and "Search engine" options


lobontiumira:
## Steps to reproduce
1. Have the Accessibility scanner app installed and running.
2. Tap on "+" in order to open a new tab, then scan the page containing the "Scan" and "Search engine" options.

### Expected behavior
No suggestions are made.

### Actual behavior
The Accessibility scanner suggests to increase the "Scan", and "Search engine" items from 40 dp, to 48 dp or larger.

### Device information
* Android device: Samsung Galaxy Tab S6 (Android 10)
* Fenix version: Beta 84.0.0-beta.3, and 12/2 Nightly
 
![Untitled](https://user-images.githubusercontent.com/40694158/100984574-913a3200-3553-11eb-840e-24a9d23f46dd.png)

Debanshu777:
@Amejia481 I would like to work on this issue

Is it suggested to just increase the size of the View from 40dp to 48dp?
 
<img src="https://user-images.githubusercontent.com/50791485/102749436-07c79400-438a-11eb-80ef-067aed2d8799.jpg" height=300 width= 500/>

Amejia481:
> @Amejia481 I would like to work on this issue
> 
> Is it suggested to just increase the size of the View from 40dp to 48dp?
>  
> <img src="https://user-images.githubusercontent.com/50791485/102749436-07c79400-438a-11eb-80ef-067aed2d8799.jpg" height=300 width= 500/>
> 
>  Hi @Debanshu777 👋,
I'm currently on PTO, feel free to work on the issue, others team members will be there to assist you in case you need any help :)

sflorean:
This is still reproducible on all builds, device: Huawei P40 Lite (Android 11).

eliserichards:
I think we should double check with UX that this is ok with them. Tagging for feedback...

Dunexus:
@eliserichards
I don't know much about current priorities but this may not be relevant if https://github.com/mozilla-mobile/fenix/issues/16659 and https://github.com/mozilla-mobile/fenix/issues/16658 are planned to be implemented soon.

topotropic:
Is it possible to increase the touch area while keeping the design as is? 

I'll let @vesta0 comment on when we plan to work on below changes.

> I don't know much about current priorities but this may not be relevant if #16659 and #16658 are planned to be implemented soon.

vesta0:
[...]

Issue #14150, opened 2020-08-25
Make QR-Code scanner link above keyboard optional in settings


terr72:
### What is the user problem or growth opportunity you want to see solved?

FF 79.0.5 introduced a QR-Code scanner icon above the keyboard, which can't be disabled anymore like it was possible in Gecko-based FFs before.

### How do you know that this problem exists today? Why is this important?

People who don't use this function on a regular basis will touch this icon by mistake.
This is not only hte case for power users, but also for elder people.

### Who will benefit from it?

Everyone who doesn't want to have the QR code scanner link.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-5463)

rc1717:
I also would like to see the qr code button made to be optional. I think it's too cluttered when the keyboard already takes up a lot of space. It would be better to take this an option that can be disabled by people who do not need it.

JonDeereGee:
I would also like to have the option of turning off the QR scanner. Instead of making things easier it gets in my way.

mirko77:
+1, no reason to show it for me, please add an option to hide it

metbic:
Seconded.

trxraptor:
Both QR Scan and Search Engine buttons should be optional.

niederlec:
Yes, it is really annoying because on small or medium screens, when opening a new tab, it's displayed over the favorites and most visited sites. So you MUST first close the keyboard if you want to use favorites.

mirko77:
I agree, showing the keyboard when opening a new tab is an anti pattern, I always need to hide it to tap on a favorite.

pellaeon:
Previously, selecting a search engine only requires one click on the search engine option. Now because of the scan QR code button, it requires two clicks (1. click "Search Engine" 2. click the specific search engine you want)

JonDeereGee:
Why is it so hard to give users a choice? Wait a sec, I DO have a choice - not use Firefox.

cjpcjpcjp:
Any update or movement on this issue? The code scanner is taking up valuable screen real estate. Many don't use it at all and it definitely should be optional.

over1pixel:
what's wrong with your ux designers?? you unilaterally changed the ux in so many parts without giving the user any possibility but accept it. Make the qr core reader bar optional!!!

sinisterseed:
Just ran into this issue earlier browsing for this.

[...]

PR #17162, opened 2020-12-21
Increasing the "Scan" and "Search engine" option items from 40 dp, to…

Debanshu777:
… 48 dp. Solution: #16829
Initial Imge of the layout

Final image of layout

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/res/layout/fragment_search_dialog.xml
+++ b/app/src/main/res/layout/fragment_search_dialog.xml
@@ -173,6 +173,7 @@
     <ToggleButton
         android:id="@+id/qr_scan_button"
         style="@style/search_pill"
+        android:layout_height="@dimen/search_fragment_toggleButton_height"
         android:layout_marginEnd="@dimen/search_fragment_scan_button_margin_end"
         android:textOff="@string/search_scan_button"
         android:textOn="@string/search_scan_button"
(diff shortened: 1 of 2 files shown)


cadeyrn:
It's not my decision, someone from Mozilla's UX people has to decide - but to me it looks like the border radius of the active button should be increased a bit as well with the new height.

Debanshu777:
@cadeyrn okay that's a good point that you mentioned. Let's see what are others view on the changes

apbitner:
@Debanshu777 Thank you for taking a look at this. Couple questions:

1. When you used the accessibility scanner, did you scan the entire app? Reason I ask is if we're changing something like a button size, it should be changed in all instances of the button.
2. While 48dp is the ideal touch target size, it's also acceptable to have an element that is smaller in size, but has enough margin/padding around it to create a 48dp touch target area, do you know if that's the case here?

topotropic:
To pick this up again, is there a way to define the touch target without changing the button size? Changing button sizes has a ripple effect and if it's not possible to define the touch target independent of the visual button size, we have to make sure we address all instances in the app. Thanks!

Mugurell:
There are multiple such issues in the SearchFragment which can be resolved by just increasing the tap area with no visible change to the layout. 
Opened https://github.com/mozilla-mobile/fenix/issues/24461 about that,

@Debanshu777 If you are still interested in helping this, I will close this PR in favor of the work to be done on for #24461.

sharbet:
A quick question: Can the "Scan" button be hidden through preferences/about:config?

"Search engine" I know it can be hidden by selecting only having one engine and removing all others in pref.

For people who do not ever use the "Scan" function, it takes up viewing space in a small screen device.

Screenshot_20220328-222413_Firefox Nightly

cadeyrn:
> A quick question: Can the "Scan" button be hidden through preferences/about:config?

No. An option is already suggested as #14150.

Issue #17143, opened 2020-12-19
[Bug] Crashes initializing Nimbus


GO-jack:
[comment]: # (Please do your best to search for duplicate issues before filing a new issue so we can keep our issue board clean)
[comment]: # (Every issue should have the exact bug and steps to reproduce described in it. Please do not file feedback list tickets as it is difficult to parse them and address their individual points)
[comment]: # (Read https://github.com/mozilla-mobile/fenix#i-want-to-file-an-issue for more information)

## Steps to reproduce

Click "+" button to open new tab.

### Expected behavior

No crash and open new tab.

### Actual behavior

Crash when click "+" button to open new tab.

Report: https://crash-stats.mozilla.org/report/index/0d29e773-b26e-4361-b64b-a07c90201219

### Device information

* Android device: ASUS ZenFone 3 (Android 6.0.1)
* Fenix version: 
![0](https://user-images.githubusercontent.com/68154921/102687140-f4041c80-4227-11eb-8050-fb75fac1eae3.png)

Dojo456:
I could not reproduce this bug at all.

hackel:
I've been getting this for the past few days as well.  Can only start Firefox by clicking as link from another app.  Trying to open it directly or open a new tab immediately crashes.  Unfortunately I have no way to provide any kind of debugging information on Android, but I've been reporting it every time from the crash notification.

csadilek:
Thanks for filing! Could one of you check if this is still a problem in latest Nightly (201221 17:00)? 

It should show Monday 12/21 @ 5:04PM under Settings->About Firefox Nightly.

gitt0r:
It's working again.

Thank you very much!

![Screenshot_2020-12-21-23-37-52-308_org mozilla fenix](https://user-images.githubusercontent.com/65116394/102828531-450d3f80-43e5-11eb-94d0-d84e4a8d25f6.jpg)

csadilek:
OK, this confirms that it was also Nimbus related.

csadilek:
Thanks for verifying!

csadilek:
Filed ticket in Nimbus for remaining crashes that will need to be addressed before enabling again:
https://jira.mozilla.com/browse/SDK-161

Issue #17086, opened 2020-12-16
[Bug] Large regression in MAIN/VIEW start up due to Nimbus network IO


mcomella:
On our FNPRMS testing dashboard:
![image](https://user-images.githubusercontent.com/759372/102416861-77670900-3faf-11eb-90db-e01dbd8dd3a1.png)

We regressed from ~0.5s to ~1.4s. Let's find the root cause and fix it.

### Device information

* Android device: ?
* Fenix version: ?

---

This may be caused by the new strict mode violation https://github.com/mozilla-mobile/fenix/issues/16984 or the other network main thread issues we've seen with experiments. #16901 

Tracking bug for nimbus-sdk team: https://jira.mozilla.com/browse/SDK-157

opusforlife2:
I was about to open a new issue regarding huge lag in startup, but this could be it. In my testing, it seems to be caused by some internet related code that blocks Fenix from loading.

- If internet is off, Fenix opens absolutely fine.
- Same if internet is on and working fast. Fenix opens quickly here as well.
- But if your internet is extremely slow or your device is not yet fully connected (connected to LAN but not to WAN) Fenix seems to wait on a white screen indefinitely until it establishes a successful connection. This sometimes also causes a crash (which, annoyingly, can't be submitted because it only happens when internet is buggy, and once you open Fenix successfully, it can't be reproduced).

Somewhat related to this, the same thing happens with bookmarks loading from the address bar and addons loading in the addons menu. They open fine if internet is on and also when it is off, but take a long time to load if it is very slow. There is some blocking code there as well (probably related to thumbnail/icon generation).

Fenix version: Nightly 201214 17:00 (Build #2015781387)

pocmo:
I see something related with the following steps:

* Launch browser (with tabs to restore)
* Browser screen opened
* Click on tabs tray
* Click on "Add tab" button
* Home screen opens

The first time the home screen launches with a noticeable delay. All further launches of the home screen are faster. I can also confirm that in airplane mode this problem is gone.

pocmo:
I created an emulator with network set to GPRS (to make the problem more visible).

I took profiles from the moment I pressed on "add tab" until the home screen was visible.

This is from master: https://share.firefox.dev/34DS5Rl

The main thread seems to be blocked on `org.mozilla.experiments.nimbus.NimbusClient.getExperimentBranch` for almost 2 seconds in this scenario.

I did completely disable Nimbus locally and took another profile:
[...]

PR #17167, opened 2020-12-21
For #17086 #17143: Temporarily turn off Nimbus

csadilek:
We decided to temporarily disable Nimbus until performance issues and remaining crashes can be resolved. Note that this only solves the more severe performance regression loading the Home screen and affecting users, not the more minor regression introduced by the `Nimbus` constructor. The latter we should fix in January as well.


travis79 (approved):

csadilek:
Hitting intermittent UI failures.

PR #17171, opened 2020-12-21
Closes #17172: Add Firebase Crashlytics support

### LINKED ISSUE #17172: Add Firebase Crashlytics support
Add support and mechanism to allow user to opt-in / opt-out of Crashlytics.

rocketsroger:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -8,7 +8,7 @@ apply plugin: 'kotlin-android-extensions'
 apply plugin: 'jacoco'
 apply plugin: 'androidx.navigation.safeargs.kotlin'
 apply plugin: 'com.google.android.gms.oss-licenses-plugin'
-
+apply plugin: 'com.google.firebase.crashlytics'
 
 import com.android.build.OutputFile
 import groovy.json.JsonOutput
@@ -510,6 +510,7 @@ dependencies {
     implementation Deps.androidx_transition
     implementation Deps.androidx_work_ktx
     implementation Deps.google_material
+    implementation Deps.firebase_crashlytics
 
     implementation Deps.adjust
     implementation Deps.installreferrer // Required by Adjust
--- a/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
+++ b/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
@@ -14,6 +14,7 @@ import androidx.appcompat.app.AppCompatDelegate
 import androidx.core.content.getSystemService
 import androidx.work.Configuration.Builder
 import androidx.work.Configuration.Provider
+import com.google.firebase.crashlytics.FirebaseCrashlytics
 import kotlinx.coroutines.Deferred
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.GlobalScope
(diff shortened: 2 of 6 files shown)


app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -223,6 +224,8 @@ open class FenixApplication : LocaleAwareApplication(), Provider {
         if (settings().isMarketingTelemetryEnabled) {
             components.analytics.metrics.start(MetricServiceType.Marketing)
         }
+
+        FirebaseCrashlytics.getInstance().setCrashlyticsCollectionEnabled(settings().isTelemetryEnabled)
```
rocketsroger, line 228: One issue I have is currently we don't have a preference switch for crash reporting because crashes are manually sent when the user tap on the action in the crash notification.  So I used the telemetry preference.

pocmo:
This is side stepping our whole crash reporting component?

rocketsroger:
> This is side stepping our whole crash reporting component?

Yes, I tried to add Firebase Crashlytics through the our crash reporting component but the Fenix side changes is still needed.  Also there's no way to opt out without Fenix side changes as well.  We can always add lib-crash implementation later with custom info if needed.

pocmo:
Sure, some integration is needed on the Fenix side. But this is now completely side-stepping the crash reporter and using the telemetry pref. As well as enabling this for every build - wasn't this supposed to be a Nightly experiment?

buildSrc/src/main/java/Dependencies.kt
```diff
@@ -50,6 +50,9 @@ object Versions {
     const val google_ads_id_version = "16.0.0"
 
     const val google_play_store_version = "1.8.0"
+    const val google_services_version = "4.3.4"
+    const val firebase_crashlytics = "17.2.2"
```
jonalmeida, line 54: This is may be a problem in the future if we try to upgrade a version here or AC since we know that individual Firebase SDKs need to have compatible versions.

Probably the same comment for `firebase_crashlytics_gradle` as well.

app/build.gradle
```diff
@@ -8,7 +8,7 @@ apply plugin: 'kotlin-android-extensions'
 apply plugin: 'jacoco'
 apply plugin: 'androidx.navigation.safeargs.kotlin'
 apply plugin: 'com.google.android.gms.oss-licenses-plugin'
-
+apply plugin: 'com.google.firebase.crashlytics'
```
jonalmeida, line 11: What does the plugin do?

I ran into this as well when setting up FCM and after inspection, we realized that this plugin was handling the crashlytics initialization which we were better equipped to do ourselves. This may not be the case here, but might be worth checking out if you haven't already. 🙂

buildSrc/src/main/java/Dependencies.kt
rocketsroger: I agree.  I was a bit worry about this as well.  But I couldn't find a way that I can add this only on the A-C side.

rocketsroger:
Closing this PR until we have more clarification on data review and opt-out mechanism.

Issue #17177, opened 2020-12-21
java.lang.NullPointerException: Attempt to invoke virtual method 'android.view.ViewTreeObserver android.view.View.getViewTreeObserver()' on a null object reference


ekager:
https://sentry.prod.mozaws.net/operations/firefox-nightly/issues/9183267/?query=is%3Aunresolved%20getViewTreeObserver

STR:
1. Open tabs in normal browsing
2. Switch to private tab in tabs tray 
3. Click FAB to open private mode home (only on clean build that hasn't seen PBM CFR yet)

PR #17178, opened 2020-12-21
For #17177 - Do not show PBM CFR if entering search on home, make sur…

ekager:
…e fragment attached before showing

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
@@ -85,6 +85,7 @@ import mozilla.components.ui.tabcounter.TabCounterMenu
 import org.mozilla.fenix.BrowserDirection
 import org.mozilla.fenix.HomeActivity
 import org.mozilla.fenix.R
+import org.mozilla.fenix.addons.runIfFragmentIsAttached
 import org.mozilla.fenix.browser.BrowserAnimator.Companion.getToolbarNavOptions
 import org.mozilla.fenix.browser.browsingmode.BrowsingMode
 import org.mozilla.fenix.components.FenixSnackbar
@@ -345,10 +346,6 @@ class HomeFragment : Fragment() {
                 sessionControlView!!.view.layoutManager?.onRestoreInstanceState(parcelable)
             }
             homeViewModel.layoutManagerState = null
-
-            // We have to delay so that the keyboard collapses and the view is resized before the
-            // animation from SearchFragment happens
-            delay(ANIMATION_DELAY)
         }
 
         observeSearchEngineChanges()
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -85,6 +85,7 @@ import mozilla.components.ui.tabcounter.TabCounterMenu
 import org.mozilla.fenix.BrowserDirection
 import org.mozilla.fenix.HomeActivity
 import org.mozilla.fenix.R
+import org.mozilla.fenix.addons.runIfFragmentIsAttached
```
csadilek, line 88: nit: This is in the addons package and I guess everyone started using it now. We should move this out in a follow-up.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -714,10 +712,12 @@ class HomeFragment : Fragment() {
             // We want to show the popup only after privateBrowsingButton is available.
             // Otherwise, we will encounter an activity token error.
             privateBrowsingButton.post {
```
csadilek, line 714: Can you help me make this reproducible? I've tried the STR in https://github.com/mozilla-mobile/fenix/issues/17177#issue-772508613 but no luck.

For me `showCondition` makes sure the popup is not displayed with the STR above: https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/java/org/mozilla/fenix/utils/Settings.kt#L856

`numTimesPrivateModeOpened.value` is only incremented when tapping the private browsing mode button on Home, but not when opening via the FAB. On a clean build, `numTimesPrivateModeOpened.value` is 0 and the popup should never be displayed. Once we hit the button on home and the popup is finally displayed we toggle off `showedPrivateModeContextualFeatureRecommender` and shouldn't show it again either.

There must be something else going on here I am missing? If so let's add a comment to the code here as well :).

ekager: you're right I can't repro with the original STR I wrote... 

I think I can still repro by getting to home screen from the FAB and then toggling the PBM button

ezgif com-video-to-gif(2)

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
ekager: Opened https://github.com/mozilla-mobile/fenix/issues/17303

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
csadilek: OK, got it. Let's add a comment above the `!bundleArgs.getBoolean(FOCUS_ON_ADDRESS_BAR)` check to describe why we're doing it? I think it will help in the future :).

csadilek (approved):

Issue #15464, opened 2020-09-26
[Bug] Address bar completion is stuck after pressing enter


w3Abhishek:
## Steps to reproduce

* Enter URL or search term in the address bar
* Tap Enter in on-screen keyboard

### Expected behavior

Completion section should disappear after clicking enter

### Actual behavior

Sometimes closes and sometimes sticks
Ref: https://youtu.be/OR7516clqj4

### Device information

* Android device: Realme 5
* Fenix version: Nightly 200926 06:03

Shamp0o:
Same behavior here on Xiaomi Redmi Note 5 running LineageOS 17.1. Fenix version: 82.0.0-beta.4

eliserichards:
Same behavior reported here with some additional details: https://www.reddit.com/r/firefox/comments/jiwqal/this_panel_doesnt_go_away_whenever_i_search_by/

giwrgosmant:
Same behavior on POCO F1 running Pixel Experience Plus Android 10. Firefox Stable Version: 84.1.2 (Build #2015783137)
Really annoying, I had to go back on Chromium base browser because of this.😕 I hope for a fix soon 🤞

SPLESHER1:
Same behavior on POCO F1. Firefox Stable Version: 84.1.2 (Build #2015783137)

w3Abhishek:
I hope this solves the issue.

sflorean:
I wasn't able to reproduce the issue on latest Nightly build with Pixel 3 (Android 11). @giwrgosmant @SPLESHER1 could you check this on Nightly? on stable version will still be reproducible until the bug is landed on release.

giwrgosmant:
> 
> 
> I wasn't able to reproduce the issue on latest Nightly build with Pixel 3 (Android 11). @giwrgosmant @SPLESHER1 could you check this on Nightly? on stable version will still be reproducible until the bug is landed on release.

Hello,

I've just tested it and it persists. Tested with Nightly from Play Store and Firefox Beta 85.0.0-beta.6 from Github.
POCO F1 running Android 10 Pixel Experience Plus.

mcarare:
> 
> Hello,
> 
> I've just tested it and it persists. Tested with Nightly from Play Store and Firefox Beta 85.0.0-beta.6 from Github.
> POCO F1 running Android 10 Pixel Experience Plus.

Nightly 210107?

giwrgosmant:
> 
> 
> > Hello,
[...]

PR #17186, opened 2020-12-22
For #15464: Always dismiss search dialog on url committed.

mcarare:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/search/SearchDialogController.kt
+++ b/app/src/main/java/org/mozilla/fenix/search/SearchDialogController.kt
@@ -74,12 +74,12 @@ class SearchDialogController(
                 navController.navigateSafe(R.id.searchDialogFragment, directions)
             }
             "moz://a" -> openSearchOrUrl(SupportUtils.getMozillaPageUrl(SupportUtils.MozillaPage.MANIFESTO))
-            else -> if (url.isNotBlank()) {
-                openSearchOrUrl(url)
-            } else {
-                dismissDialog()
-            }
+            else ->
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/fenix/search/SearchDialogController.kt
```diff
@@ -74,12 +74,12 @@ class SearchDialogController(
...
             }
             "moz://a" -> openSearchOrUrl(SupportUtils.getMozillaPageUrl(SupportUtils.MozillaPage.MANIFESTO))
-            else -> if (url.isNotBlank()) {
-                openSearchOrUrl(url)
-            } else {
-                dismissDialog()
-            }
+            else ->
+                if (url.isNotBlank()) {
+                    openSearchOrUrl(url)
```
ekager, line 79: All of these other methods call `clearToolbarFocus` which I assume then dismiss the dialog somehow? Do we know where that happens and why it doesn't work for certain devices? Or is it a race that can happen on any device?

mcarare: I am not completely sure calling `clearToolbarFocus` would dismiss the dialog on all platforms and devices. It is not mentioned in the documentation for Dialog Fragment as a way to dismiss the dialog. I could not reproduce the bug on multiple devices, but this speculative fix should work in those instances where the dialog is still shown. And for the others, it should not change anything as it checks if the dialog is not already dismissed.

eliserichards (approved):
🙌 🚀

Issue #17073, opened 2020-12-16
Intermittent debug(T) test failure: java.lang.AssertionError: Verification failed: call 5 of 5: class org.mozilla.fenix.components.AnalyticsKt.isSentryEnabled()) should not be called


Archaeopteryx:
Failed 3/5 runs.

@pocmo Please check the recent commits if any of those triggers this.

Log: https://treeherder.mozilla.org/logviewer?job_id=324674867&repo=fenix

```
SUITE: org.mozilla.fenix.shortcut.PwaOnboardingObserverTest
  TEST: GIVEN cfr should not yet be shown WHEN installable page is loaded THEN counter is incremented
  SUCCESS

  TEST: GIVEN cfr should be shown WHEN installable page is loaded THEN we navigate to onboarding fragment
  SUCCESS

  TEST: GIVEN web app is not installable WHEN page with manifest is loaded THEN nothing happens
  FAILURE

java.lang.AssertionError: Verification failed: call 5 of 5: class org.mozilla.fenix.components.AnalyticsKt.isSentryEnabled()) should not be called

Calls:
1) class org.mozilla.fenix.components.AnalyticsKt.isSentryEnabled()


Stack traces:
1)                                                                 io.mockk.impl.InternalPlatform.captureStackTrace              (InternalPlatform.kt:121)           
                                                                     io.mockk.impl.stub.MockKStub.handleInvocation               (MockKStub.kt:247)                  
                                   io.mockk.impl.instantiation.JvmMockFactoryHelper$mockHandler$1.invocation                     (JvmMockFactoryHelper.kt:25)        
                                                            io.mockk.proxy.jvm.advice.Interceptor.call                           (Interceptor.kt:20)                 
                                                         org.mozilla.fenix.components.AnalyticsKt.isSentryEnabled                (Analytics.kt:134)                  
                                                            org.mozilla.fenix.ext.NavControllerKt.recordIdException              (NavController.kt:31)               
                                                            org.mozilla.fenix.ext.NavControllerKt.nav                            (NavController.kt:22)               
                                                            org.mozilla.fenix.ext.NavControllerKt.nav$default                    (NavController.kt:18)               
        org.mozilla.fenix.shortcut.PwaOnboardingObserver$start$1$invokeSuspend$$inlined$collect$1.emit                           (Collect.kt:138)                    
[...]

PR #17191, opened 2020-12-22

MickeyMoz:


csadilek:
Ran into the [intermittent](https://github.com/mozilla-mobile/fenix/issues/17073) which is already fixed on master, but this branch didn't have the fix yet. So... 🤞

PR #17207, opened 2020-12-24
For #17073: Stop observers after test run

csadilek:
Test failed again in https://github.com/mozilla-mobile/fenix/runs/1603080054. I forgot to stop the observer after tests finish so an observer of a previous test can interfere with the test currently running.


jonalmeida (approved):
🎊

PR #17336, opened 2021-01-06
For #16615, #10798, #14526: New download UI smoke test and fix for testDownloadPrompt & testDownloadNotification

sv-ohorvath:
Ran tests on Firebase multiple times to check for flakiness and passed.

### DIFF
--- a/app/src/androidTest/java/org/mozilla/fenix/ui/DownloadTest.kt
+++ b/app/src/androidTest/java/org/mozilla/fenix/ui/DownloadTest.kt
@@ -4,25 +4,21 @@
 
 package org.mozilla.fenix.ui
 
-import android.os.Environment
 import androidx.test.platform.app.InstrumentationRegistry
 import androidx.test.rule.GrantPermissionRule
 import androidx.test.uiautomator.UiDevice
-import kotlinx.coroutines.runBlocking
 import okhttp3.mockwebserver.MockWebServer
 import org.junit.After
 import org.junit.Before
-import org.junit.Ignore
 import org.junit.Rule
 import org.junit.Test
 import org.mozilla.fenix.helpers.AndroidAssetDispatcher
 import org.mozilla.fenix.helpers.HomeActivityTestRule
 import org.mozilla.fenix.helpers.TestAssetHelper
+import org.mozilla.fenix.helpers.TestHelper
 import org.mozilla.fenix.ui.robots.downloadRobot
-import org.mozilla.fenix.ui.robots.homeScreen
 import org.mozilla.fenix.ui.robots.navigationToolbar
 import org.mozilla.fenix.ui.robots.notificationShade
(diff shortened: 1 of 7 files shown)


app/src/androidTest/java/org/mozilla/fenix/ui/DownloadTest.kt
```diff
@@ -56,35 +52,20 @@ class DownloadTest {
...
-    @Suppress("Deprecation")
     @After
     fun tearDown() {
         mockWebServer.shutdown()
 
-        // Clear Download
-        runBlocking {
-            val downloadedFile = File(
-                Environment.getExternalStoragePublicDirectory(Environment.DIRECTORY_DOWNLOADS),
-                "Globe.svg.html"
```
AaronMT, line 68: I remember adding .svg.html because there was a problem saving the svg. It might have been related to the MIME type issues we had with Mockwebserver a while back. I have no idea if it's still needed or not. Does the above approach work with just passing in the filename?

app/src/androidTest/java/org/mozilla/fenix/ui/SmokeTest.kt
```diff
@@ -568,4 +578,38 @@ class SmokeTest {
...
+        }.clickDownload {
+            mDevice.waitForIdle()
+            verifyDownloadNotificationPopup()
+        }
+
+        browserScreen {
+        }.openThreeDotMenu {
+        }.openDownloadsManager {
+            waitForDownloadsListToExist()
+            verifyDownloadedFileName("Globe.svg")
```
AaronMT, line 606: We can probably just create a local variable with the filename so we only need to change it once if we change the filename in the future

AaronMT (approved):
Couple questions but otherwise looks OK to me, thanks @sv-ohorvath

app/src/androidTest/java/org/mozilla/fenix/ui/SmokeTest.kt
sv-ohorvath: sure, good point.

app/src/androidTest/java/org/mozilla/fenix/ui/DownloadTest.kt
sv-ohorvath: It worked so far in my trial runs.  Did it fail to delete it consistently in your case?

AaronMT: It was on my Pixel 2 physical device. Maybe it's not an issue anymore. I trust you that it works now.

sv-ohorvath: Tested and it works well on a physical Pixel 2 as well. Merging.

PR #17375, opened 2021-01-08
Update Android Components version to 71.0.20210108190105

csadilek:
Manual upgrade to address breaking API change in app links feature.


rocketsroger (approved):
🚢

PR #17401, opened 2021-01-11
Update to latest feature-intent and feature-recentlyclosed APIs (and update Android Components).

pocmo:
Fixing the breaking changes that AC introduced.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/library/recentlyclosed/RecentlyClosedController.kt
+++ b/app/src/main/java/org/mozilla/fenix/library/recentlyclosed/RecentlyClosedController.kt
@@ -11,10 +11,10 @@ import androidx.navigation.NavController
 import androidx.navigation.NavOptions
 import mozilla.components.browser.session.SessionManager
 import mozilla.components.browser.state.action.RecentlyClosedAction
-import mozilla.components.browser.state.state.ClosedTab
+import mozilla.components.browser.state.state.recover.RecoverableTab
 import mozilla.components.browser.state.store.BrowserStore
 import mozilla.components.concept.engine.prompt.ShareData
-import mozilla.components.feature.recentlyclosed.ext.restoreTab
+import mozilla.components.feature.tabs.TabsUseCases
 import org.mozilla.fenix.BrowserDirection
 import org.mozilla.fenix.HomeActivity
 import org.mozilla.fenix.R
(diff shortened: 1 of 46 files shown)


pocmo:
One UI test failed with:

> android.database.sqlite.SQLiteConstraintException: UNIQUE constraint failed: recently_closed_tabs.uuid (code 1555 SQLITE_CONSTRAINT_PRIMARYKEY)

I'll investigate and fix.

pocmo:
The correct fix for this is to also do what is described in the following issue:
https://github.com/mozilla-mobile/android-components/issues/8497

I'll implement in AC and update this PR with the new Nightly version once available.

app/src/main/java/org/mozilla/fenix/library/recentlyclosed/RecentlyClosedController.kt
```diff
@@ -64,23 +65,23 @@ class DefaultRecentlyClosedController(
...
-                activity.openToBrowser(
-                    from = BrowserDirection.FromRecentlyClosed
-                )
-            }
+    override fun handleRestore(item: RecoverableTab) {
+        tabsUseCases.restore(item)
+
+        store.dispatch(
+            RecentlyClosedAction.RemoveClosedTabAction(item)
+        )
```
Amejia481, line 81: Not sure if later we could add a use case for this

Amejia481 (approved):
Looks great ✨ , just one comment.

app/src/main/java/org/mozilla/fenix/library/recentlyclosed/RecentlyClosedController.kt
pocmo: Yeah, probably, I guess ideally all action dispatching is hidden behind a nicer use case. I guess we could have a specific restore use case that calls into the restore above and then dispatches this additionally.

bb010g:
> One UI test failed with:
> 
> > android.database.sqlite.SQLiteConstraintException: UNIQUE constraint failed: recently_closed_tabs.uuid (code 1555 SQLITE_CONSTRAINT_PRIMARYKEY)
> 
> I'll investigate and fix.

This exception has been crashing Fenix when tabs from previous launches are (attempted) to be closed.

```
E ExceptionHandler: Uncaught exception handled:
E ExceptionHandler: android.database.sqlite.SQLiteConstraintException: UNIQUE constraint failed: recently_closed_tabs.uuid (code 1555 SQLITE_CONSTRAINT_PRIMARYKEY)
```

PR #16888, opened 2020-12-07
Close #11355: Works with and defaults to China FxA/Sync services

rxumoz:
- Add a switch in China build (_Config.channel.isMozillaOnline == true_) to allow users to switch between global/local sync services.

- Read fennec fxa state to keep the same server setting during update.

- Allow custom server auto login in China build during migration.

I have tried to write some tests about UI changes in the beginning but it seems there is some structure issue, so I dropped them and @bli-mozbj have tested the whole feature manually. If you feel they are necessary, maybe I can think more to overcome, or you can give me some suggestions about the tests.

PR #17409, opened 2021-01-12
Close #11355: Works with and defaults to China FxA/Sync services

### LINKED ISSUE #11355: [Feature:China] Works with and defaults to China FxA/Sync servivces
Works with and defaults to China FxA
with a toggle to use global FxA, simliar to Firefox for iOS

Most Chinese users use China FxA/Sync services

rxumoz:
A copy of #16888
Due to l10n concerns, make it a draft pr temporarily.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/FxaServer.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/FxaServer.kt
@@ -6,6 +6,7 @@ package org.mozilla.fenix.components
 import android.content.Context
 import mozilla.components.service.fxa.ServerConfig
 import mozilla.components.service.fxa.ServerConfig.Server
+import org.mozilla.fenix.Config
 import org.mozilla.fenix.ext.settings
 
 /**
@@ -17,10 +18,18 @@ object FxaServer {
     const val REDIRECT_URL = "urn:ietf:wg:oauth:2.0:oob:oauth-redirect-webchannel"
 
     fun config(context: Context): ServerConfig {
+        // If a server override is configured, use that. Otherwise:
+        // - for all channels other than Mozilla Online, use Server.Release.
+        // - for Mozilla Online channel, if domestic server is allowed, use Server.CHINA; otherwise, use Server.RELEASE
         val serverOverride = context.settings().overrideFxAServer
         val tokenServerOverride = context.settings().overrideSyncTokenServer.ifEmpty { null }
         if (serverOverride.isEmpty()) {
-            return ServerConfig(Server.RELEASE, CLIENT_ID, REDIRECT_URL, tokenServerOverride)
+            val releaseServer = if (Config.channel.isMozillaOnline && context.settings().allowDomesticChinaFxaServer) {
+                Server.CHINA
+            } else {
+                Server.RELEASE
+            }
+            return ServerConfig(releaseServer, CLIENT_ID, REDIRECT_URL, tokenServerOverride)
         }
         return ServerConfig(serverOverride, CLIENT_ID, REDIRECT_URL, tokenServerOverride)
     }
--- a/app/src/main/java/org/mozilla/fenix/settings/SettingsFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/SettingsFragment.kt
@@ -26,6 +26,7 @@ import androidx.navigation.findNavController
 import androidx.navigation.fragment.navArgs
 import androidx.preference.Preference
 import androidx.preference.PreferenceFragmentCompat
+import androidx.preference.SwitchPreference
 import androidx.recyclerview.widget.RecyclerView
 import kotlinx.android.synthetic.main.amo_collection_override_dialog.view.*
 import kotlinx.coroutines.CoroutineScope
@@ -92,7 +93,8 @@ class SettingsFragment : PreferenceFragmentCompat() {
             scope = lifecycleScope,
             accountManager = requireComponents.backgroundServices.accountManager,
             httpClient = requireComponents.core.client,
-            updateFxASyncOverrideMenu = ::updateFxASyncOverrideMenu
+            updateFxASyncOverrideMenu = ::updateFxASyncOverrideMenu,
+            updateFxAAllowDomesticChinaServerMenu = :: updateFxAAllowDomesticChinaServerMenu
         )
 
         // Observe account changes to keep the UI up-to-date.
@@ -447,6 +449,7 @@ class SettingsFragment : PreferenceFragmentCompat() {
         }
 
         setupAmoCollectionOverridePreference(requireContext().settings())
+        setupAllowDomesticChinaFxaServerPreference()
     }
 
     /**
@@ -539,6 +542,22 @@ class SettingsFragment : PreferenceFragmentCompat() {
         }
     }
 
+    private fun updateFxAAllowDomesticChinaServerMenu() {
+        val settings = requireContext().settings()
+        val preferenceAllowDomesticChinaServer =
+            findPreference<SwitchPreference>(getPreferenceKey(R.string.pref_key_allow_domestic_china_fxa_server))
+        // Only enable changes to these prefs when the user isn't connected to an account.
+        val enabled =
+            requireComponents.backgroundServices.accountManager.authenticatedAccount() == null
+        val checked = settings.allowDomesticChinaFxaServer
+        val visible = Config.channel.isMozillaOnline
+        preferenceAllowDomesticChinaServer?.apply {
+            isEnabled = enabled
+            isChecked = checked
+            isVisible = visible
+        }
+    }
+
     private fun updateFxASyncOverrideMenu() {
         val preferenceFxAOverride =
             findPreference<Preference>(getPreferenceKey(R.string.pref_key_override_fxa_server))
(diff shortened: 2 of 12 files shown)


rxumoz:
@grigoryk Hi, can you review this PR when you are free?

app/src/main/java/org/mozilla/fenix/components/FxaServer.kt
```diff
@@ -19,7 +22,27 @@ object FxaServer {
     fun config(context: Context): ServerConfig {
         val serverOverride = context.settings().overrideFxAServer
         val tokenServerOverride = context.settings().overrideSyncTokenServer.ifEmpty { null }
+
```
grigoryk, line 25: There are basically five states here. I'll write out what servers we want to use for them:

1) regular builds - `Server.RELEASE`
2) regular builds where user set a `serverOverride` - `serverOverride`
3) mozilla online builds - `Server.CHINA`
4) mozilla online builds configured with a `serverOverride` - `serverOverride`.
5) mozilla online builds where user asked us to use a global server instead of "local"

In this PR, you're also looking at the old Fennec account state. If I understand correctly, I think what that's trying to achieve is support for state (4) for the migrated mozilla online users, e.g. users of mozilla online that had a custom server configured during a migration.

However, our FxA migration logic specifically detects custom servers and doesn't migrate them. So, these users (a vast, vast minority) won't have their FxA state carried over. They're expected to re-configure their custom servers and login again.

Cases (2) and (4) above are basically the same - we'll just use the override. So we just get:
1) any build configured with a `serverOverride` - use `serverOverride`
2) no `serverOverride`, regular builds - use `Server.RELEASE`
3) no `serverOverride`, mozilla online builds, domestic China servers allowed by user - use `Server.CHINA`
4) no `serverOverride`, mozilla online builds, domestic China servers not allowed by user - use `Server.RELEASE`

So, I think changes to `FxaServer.kt` in this PR can be reduced to the following snippet, which should cover everything:

```
    fun config(context: Context): ServerConfig {
        // If a server override is configured, use that. Otherwise:
        // - for all channels other than Mozilla Online, use Server.Release.
        // - for Mozilla Online channel, if domestic server is allowed, use Server.CHINA; otherwise, use Server.RELEASE
        val serverOverride = context.settings().overrideFxAServer
        val tokenServerOverride = context.settings().overrideSyncTokenServer.ifEmpty { null }
        if (serverOverride.isEmpty()) {
            val releaseServer = if (Config.channel.isMozillaOnline && context.settings().allowChinaDomesticServer) {
                Server.CHINA
            } else {
                Server.RELEASE
            }
            return ServerConfig(releaseServer, CLIENT_ID, REDIRECT_URL, tokenServerOverride)
        }
        return ServerConfig(serverOverride, CLIENT_ID, REDIRECT_URL, tokenServerOverride)
    }
```

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -881,6 +881,32 @@ class Settings(private val appContext: Context) : PreferencesHolder {
         default = false
     )
 
+    var useLocalFxAServer by booleanPreference(
```
grigoryk, line 884: Maybe rename this to something like `allowDomesticChinaFxaServer`? This is a very clear name, to me at least. I'd also add a comment saying that this pref is only used in conjunction with Mozilla Online builds.

Word "local" could mean many different things in a broader context, and could easily be misunderstood for e.g. local development server.

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -881,6 +881,32 @@ class Settings(private val appContext: Context) : PreferencesHolder {
...
+
+    fun switchUseLocalFxAServer() {
+        val key = appContext.getPreferenceKey(R.string.pref_key_use_local_fxa_server)
+        val newValue = !preferences.getBoolean(key, true)
+        preferences.edit()
+            .putBoolean(key, newValue)
+            .apply()
+    }
+
+    var haveReadFxAAccountJson by booleanPreference(
```
grigoryk, line 897: I don't think you need this nor `switchHaveReadFxAAccountJson` below (see my comment in FxaServer.kt)

app/src/main/res/values-zh-rCN/mozonline_strings.xml
```diff
@@ -19,4 +19,12 @@
     <string name="privacy_notice_positive_button">同意并继续</string>
     <!-- Privacy Notice neutral button-->
     <string name="privacy_notice_neutral_button">退出应用</string>
+
+    <!-- FxA server switch-->
+    <!-- Switch Preference for local/global fxa server -->
+    <string name="preferences_use_local_fxa_server">使用本地服务</string>
```
grigoryk, line 25: Rename this pref name, too.

app/src/main/res/values/mozonline_strings.xml
```diff
@@ -19,4 +19,12 @@
     <string name="privacy_notice_positive_button">Agree and Continue</string>
     <!-- Privacy Notice neutral button-->
     <string name="privacy_notice_neutral_button">Exit the App</string>
+
+    <!-- FxA server switch-->
+    <!-- Switch Preference for local/global fxa server -->
+    <string name="preferences_use_local_fxa_server">Use local service</string>
```
grigoryk, line 25: And here.

app/src/main/java/org/mozilla/fenix/settings/SettingsFragment.kt
```diff
@@ -508,6 +511,21 @@ class SettingsFragment : PreferenceFragmentCompat() {
         }
     }
 
+    private fun updateFxAUseLocalMenu() {
+        val settings = requireContext().settings()
+        val preferenceUseLocalFxAServer =
+            findPreference<SwitchPreference>(getPreferenceKey(R.string.pref_key_use_local_fxa_server))
+        val enabled =
```
grigoryk, line 518: Let's carry over comment from below:
`// Only enable changes to these prefs when the user isn't connected to an account.`

app/src/main/java/org/mozilla/fenix/settings/SettingsFragment.kt
```diff
@@ -546,6 +564,31 @@ class SettingsFragment : PreferenceFragmentCompat() {
...
+    private fun setupUseLocalFxaServerPreference() {
+        val useLocalFxAServer = getPreferenceKey(R.string.pref_key_use_local_fxa_server)
+        val preferenceUseLocalFxAServer = findPreference<SwitchPreference>(useLocalFxAServer)
+        val visible = Config.channel.isMozillaOnline
+
+        preferenceUseLocalFxAServer?.apply {
+            isVisible = visible
+        }
+
+        preferenceUseLocalFxAServer?.onPreferenceChangeListener =
```
grigoryk, line 576: You can short-circuit if this isn't a `isMozillaOnline` build. No need to setup `onPreferenceChangeListener` here in that case.

grigoryk (changes requested):
I think this is basically fine, but once we simplify changes to FxaServer.kt. Also, let's rename "local fxa server" to something more specific such as "domestic china server" or similar.

Let me know if I misunderstood your intent behind reading fennec's account state, which I don't think should happen here.

app/src/main/java/org/mozilla/fenix/components/FxaServer.kt
rxumoz: I need to explain a little bit here. 
For Fennec Mozilla Online build users, there are some of them using **Global** server. **Global** is not a default setting in either Fennec or Fenix for Mozilla Online builds, but it may not be thought as the custom server as well. 

So I read fxa.account.json here, trying to keep those Mozilla-Online-build && Global-server users' setting.

I have to say those Mozilla-Online-build && Global-server users should be the minority, but sorry for not having accurate statistics.

With this explanation, we can discuss more about this part. Both ways are acceptable for me.

app/src/migration/java/org/mozilla/fenix/MigratingFenixApplication.kt
```diff
@@ -31,7 +33,7 @@ class MigratingFenixApplication : FenixApplication() {
                 this.components.core.pinnedSiteStorage
             )
             .migrateLogins(this.components.core.lazyPasswordsStorage)
-            .migrateFxa(lazy { this.components.backgroundServices.accountManager })
+            .migrateFxa(lazy { this.components.backgroundServices.accountManager }, fxaExpectChinaServers)
```
grigoryk, line 36: Was this flag present in the migration builds you already shipped? I ask because if it wasn't present, Mozilla-Online users who were not using Global server were logged-out... Hopefully that didn't happen.

app/src/main/java/org/mozilla/fenix/components/FxaServer.kt
grigoryk: So, I wrote a response which github decided to erase... i'll rewrite it.

Thanks for clarifying!

A "custom server" is really just a ui concept. As far as the fxa module is concerned, there's just a `ServerConfig`. In a regular Fenix build, there is only one way to change this ServerConfig - via developer mode, which lets users input custom server URLs. In a MozillaOnline build, this PR also adds a flag to flip the Global server on. But the end result is the same as a user entering a custom server url - we end up modifying ServerConfig.

Regarding `fxa.account.json` - I really, really don't want to add custom migration logic into our main codebase. This is why we have the pretty extensive support/migration package in android-components - to keep everything organized there, so that we can easily delete all of this migration code as soon as we can. And at this point, I don't want to change that package, either. It's really just waiting for the long tail of users to finish migrating, then we delete it as soon as possible.

Most importantly, consider what already happened during the MozillaOnline migration to use Fenix codebase. Users that had Global server configured in Fennec all got logged-out: https://github.com/mozilla-mobile/android-components/blob/44013851e8e9b487e5a62ebd616775eed45d8a7c/components/support/migration/src/main/java/mozilla/components/support/migration/FennecFxaMigration.kt#L282

This PR adds a UI flag to switch-on Global servers, so I'm assuming that these users could only log-into regular Chinese servers. So let's assume they logged-into these Chinese servers. If you now migrate their Fennec preference, you will break these users - FxA module will try to talk to the Global server, while it's logged-into the Chinese server. It won't work, and they'll essentially get logged-out.

So, I don't think you can migrate this flag anymore. It should have been done during the initial migration, but now it's too late.

I would keep this PR simple - remove all of the `fxa.account.json` related code (that is, remove any custom migration logic).

Then, this PR will only do two things:
1) allow users to use Global servers via a UI flag in settings
2) pass along the `fxaExpectChinaServers` flag to migration logic.

See my other question about (2)... I'm assuming you must have had a custom fenix build that already set this flag, right?

app/src/migration/java/org/mozilla/fenix/MigratingFenixApplication.kt
rxumoz: Yes, the builds we shipped have already had this flag present.

grigoryk (approved):

Issue #17199, opened 2020-12-23
[Bug] Tapping on PWA shortcut opens custom tab with last opened tab instead


agi:
## Steps to reproduce

This is only approximate, I see this from time to time (it might be related to the PWA startup crash that was fixed recently)

- Have some tabs opened and a PWA launcher on the home screen (e.g. twitter)
- Leave the browser alone for a little bit
- Launch PWA from home shortcut

### Expected behavior

The PWA website opens

### Actual behavior

A custom tab browser opens, and from what it looks like it's the most recent tab from Fenix, also the title is incorrectly taken from the PWA website.

### Device information

* Android device: Samsung S10e
* Fenix version: Latest from playstore
 
Notice the title says "Compose new Tweet":

<img src="https://user-images.githubusercontent.com/69051186/103031704-358c1480-4513-11eb-816b-360f000f7d3c.jpg" width="400">

fbertsch:
I think I'm seeing a version of this behavior, and it's very reproducible:

1. Open both PWA "app" and Firefox Nightly
2. Navigate to new web page in Firefox Nightly (e.g. `google.com`)
3. Firefox nightly doesn't navigate to new page, but the PWA page does

Expected behavior: PWA page stays the same, Firefox Nightly navigates to new page

Note: This does _not_ happen with searching, just when navigating to a new site. @agi let me know if you'd like me to instead open a new issue.

agi:
@fbertsch Thanks! I think that's the same bug.

sflorean:
Verified as fixed on Nightly 1/18 with Samsung Note10(Android 10).

PR #17429, opened 2021-01-13
For #17199: Use browserStore selectedTabId instead of sessionManager's.

mcarare:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
+++ b/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
@@ -51,6 +51,7 @@ import mozilla.components.feature.contextmenu.DefaultSelectionActionDelegate
 import mozilla.components.feature.privatemode.notification.PrivateNotificationFeature
 import mozilla.components.feature.search.BrowserStoreSearchAdapter
 import mozilla.components.feature.search.ext.legacy
+import mozilla.components.feature.session.SessionUseCases
 import mozilla.components.service.fxa.sync.SyncReason
 import mozilla.components.support.base.feature.UserInteractionHandler
 import mozilla.components.support.ktx.android.arch.lifecycle.addObservers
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -785,7 +786,13 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
         // has removed all of them, or we couldn't load any) we will pass searchTermOrURL to Gecko
         // and let it try to load whatever was entered.
         if ((!forceSearch && searchTermOrURL.isUrl()) || engine == null) {
-            loadUrlUseCase.invoke(searchTermOrURL.toNormalizedUrl(), flags)
+            if (loadUrlUseCase is SessionUseCases.DefaultLoadUrlUseCase) {
```
csadilek, line 789: Thanks for taking this on! Can you describe the fix and maybe add a comment here too? It's not clear to me reading the patch why this fixes the issue.

Since this is an important fix, it would great to cover with a test too in `HomeActivityTest`...could break into a separate method to make it easier to test.

csadilek: @mcarare 😱  So you're seeing the PWA session being "selected" in `SessionManager`, but (correctly) not selected in the `Store`? The custom tab and PWA sessions should never be marked as selected. Can you reproduce this reliably?

If so, we should get to the bottom of this and see how the wrong session gets selected, as this can cause a number of different problems. If we fail and can't find the cause, we can still land the workaround, but we should definitely try.

/cc @pocmo

mcarare: @csadilek Yes, I think the problem lies in `sessionManager` and the way the selected session is updated, not exclusively in relation to PWA or custom tabs. 
Maybe is related to another issue fixed in another of my  PRs: https://github.com/mozilla-mobile/fenix/pull/17412. If you open a tab from collection it is not automatically selected, but it should have been by simply opening a new tab with the respective restored tab session.

mcarare: And yes, I can reproduce this 100% of the time.

mcarare: Tested some more, this only happens with PWA it does not happen with regular not custom tabs.

csadilek: OK, it seems you have reliable STR? Next step is to debug `LegacySessionManager` and see how the PWA tab becomes the selected tab. I think that's the part we need to understand to fix this in all cases. How can that happen?

mcarare: `WebAppIntentProcessor` seems to be the problem. I will open an issue and a PR on AC.

mcarare:
Closing in favor of https://github.com/mozilla-mobile/android-components/pull/9406.

Issue #17377, opened 2021-01-08
[Bug] Focusing a filled address bar shows blank search suggestions view instead of scrim (regression?)


violasong:
### Steps to reproduce
When viewing a website page, and then tapping the address bar, we see a blank white search suggestions view. (In certain cases on google search results pages, we see search suggestions instead.)

### Expected behavior
We should actually show a gray scrim instead of the blank pane. It should be the same experience as when there's nothing in the address bar. As a general rule, if the search suggestions pane is completely blank, we shouldn't show it.

We'll eventually want to add some useful elements here like in other browsers (https://github.com/mozilla-mobile/fenix/issues/9118#issuecomment-724188456), but as an intermediate step, I think we should see a scrim here.

<img width="335" alt="image" src="https://user-images.githubusercontent.com/55994/104076089-c9bfd200-51c9-11eb-8a0e-c3c1033c872e.png">

### Device information

* Android device: Galaxy S20
* Fenix version: Nightly

abodea:
I was able to reproduce this issue on the latest nightly 1/12 with Google Pixel 4 XL (11).

abodea:
Verified as fixed on the latest Nightly 1/25 with Google Pixel 4 XL (11).
Note that the scrim effect is correctly displayed when focusing a filled address bar.

PR #17435, opened 2021-01-13
For #17377 - Fix for empty awesomeBar on searchFragment open

codrut-topliceanu:
The awesomeBar was set to stay hidden for the first consumeFrom(store) of the SearchDialogFragment.kt. However, recent changes send an additional store update when the view is created. To work around it we can take a look at AwesomeBarView.update(), we see that the awesomeBar suggestions get provided only if the query != url. So we can use the same method to keep the awesomeBar hidden until the user changes anything in the url.

### Pull Request checklist

- [x] **Screenshots**: This PR includes screenshots or GIFs

1. click on Show All Checks,

https://user-images.githubusercontent.com/60002907/104474113-5c93ae80-55c6-11eb-8dd3-abd1b44cfb5f.mp4

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/search/SearchDialogFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/search/SearchDialogFragment.kt
@@ -313,22 +313,25 @@ class SearchDialogFragment : AppCompatDialogFragment(), UserInteractionHandler {
         }
 
         consumeFrom(store) {
-            val shouldShowAwesomebar =
-                !firstUpdate &&
-                it.query.isNotBlank() ||
-                it.showSearchShortcuts
-
-            awesome_bar?.visibility = if (shouldShowAwesomebar) View.VISIBLE else View.INVISIBLE
+            /*
+            * firstUpdate is used to make sure we keep the awesomebar hidden on the first run
+            *  of the searchFragmentDialog. We only turn it false after the user has changed the
+            *  query as consumeFrom may run several times on fragment start due to state updates.
+            * */
+            if (it.url != it.query) firstUpdate = false
+            awesome_bar?.visibility = if (shouldShowAwesomebar(it)) View.VISIBLE else View.INVISIBLE
             updateSearchSuggestionsHintVisibility(it)
             updateClipboardSuggestion(it, requireContext().components.clipboardHandler.url)
             updateToolbarContentDescription(it)
             updateSearchShortcutsIcon(it)
             toolbarView.update(it)
             awesomeBarView.update(it)
-            firstUpdate = false
             addVoiceSearchButton(it)
         }
     }
+
+    private fun shouldShowAwesomebar(searchFragmentState: SearchFragmentState) =
+        !firstUpdate && searchFragmentState.query.isNotBlank() || searchFragmentState.showSearchShortcuts
 
     private fun updateAccessibilityTraversalOrder() {
         val searchWrapperId = search_wrapper.id


codrut-topliceanu:
Pushed a fix for the case when the awesomebar would disappear if ever the user would type in the url bar the same url the tab has.

codrut-topliceanu:
I've found two fixes for this issue.
**1st** one is in the code for this PR, and works by hiding the Awesomebar when `SearchDialogFragment` is first opened and it keeps it hidden until the query and url are different. (as mentioned in the commit msg). 
Here it is in action:
Fenix_shouldShowAwesomebar


**2nd** one would be in AC in `BrowserAwesomebar.kt` at the end of the `queryProvidersForSuggestions()` function and works by hiding the awesomebar if the suggestion adapter is empty.
 ```
       job?.invokeOnCompletion {
            this@BrowserAwesomeBar.visibility =
                if (suggestionsAdapter.itemCount == 0) View.INVISIBLE else View.VISIBLE
        }
```
And here is a GIF of it in action:
AC_job_invokeOnCompletion

The 1st solution, in Fenix, is a lot more responsive but not as "elegant" as it uses a rather complex `if` case over there.
The 2nd solution, in AC, is a lot more clear and fixes the root of the problem: no need to show any awesomebar if there isn't anything to show in it. However it is a bit slower as it get's invoked after the job that queries the providers for suggestions gets called. While we do have a time limit on those, it might still seem noticeable to keen eyed users. (If you look carefully at the two GIFs above you can see that the first one displays the awesomebar and then it gets populated with suggestions).

File: app/src/main/java/org/mozilla/fenix/search/SearchDialogFragment.kt [hunk 1]
```diff
@@ -313,23 +313,22 @@ class SearchDialogFragment : AppCompatDialogFragment(), UserInteractionHandler {
...
-            awesome_bar?.visibility = if (shouldShowAwesomebar) View.VISIBLE else View.INVISIBLE
+            awesome_bar?.visibility = if (shouldShowAwesomebar(it)) View.VISIBLE else View.INVISIBLE
             updateSearchSuggestionsHintVisibility(it)
             updateClipboardSuggestion(it, requireContext().components.clipboardHandler.url)
             updateToolbarContentDescription(it)
             updateSearchShortcutsIcon(it)
             toolbarView.update(it)
             awesomeBarView.update(it)
-            firstUpdate = false
+            if (it.url != it.query) firstUpdate = false
```
csadilek, line 323: It looks like if we move this line on top of this block then we can simplify `shouldShowAwesomebar` as the additional url != query check isn't needed.

```kotlin
consumeFrom(store) {
  if (it.url != it.query) firstUpdate = false
  awesome_bar?.visibility = if (shouldShowAwesomebar(it)) View.VISIBLE else View.INVISIBLE
  ...
}
```

```kotlin
private fun shouldShowAwesomebar(searchFragmentState: SearchFragmentState) =
  !firstUpdate && searchFragmentState.query.isNotBlank() || searchFragmentState.showSearchShortcuts

```

csadilek (approved):
I think it's fine to fix this here as it's more responsive, as your wrote, and would need a bit more thought in A-C e.g. is it always the right default to hide the awesomebar if there are no suggestions or do we need to make this configurable?

So +1 to landing this. It regressed a while ago and is already in beta as well, so we should consider uplifting this.

One nit / potential simplification below. It would also be nice to cover this with a unit test, but in this case it can be a follow-up ticket.

app/src/main/java/org/mozilla/fenix/search/SearchDialogFragment.kt
ekager: Could we also add a comment about what `firstUpdate` is and why we use it?

PR #17452, opened 2021-01-14
Closes #17451: remove verbose logs with proguard.

### LINKED ISSUE #17451: Remove verbose logging with proguard
In https://github.com/mozilla-mobile/fenix/issues/14735, we added code to remove debug logging. However, `verbose` logging is still available in the application. The [official docs](https://developer.android.com/reference/android/util/Log) mention verbose should be even less likely to make it in the release builds than `debug`:

> The order in terms of verbosity, from least to most is ERROR, WARN, INFO, DEBUG, VERBOSE. Verbose should never be compiled into an application except during development. Debug logs are compiled in but stripped at runtime. Error, warning and info logs are always kept. 

Note that to allow verbose logs to appear in logcat, you need to run `adb shell setprop log.tag.<insert-logtag-here> VERBOSE` with your specific `logtag`.

mcomella:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/proguard-rules.pro
+++ b/app/proguard-rules.pro
@@ -50,6 +50,7 @@
 ####################################################################################################
 -assumenosideeffects class android.util.Log {
     public static boolean isLoggable(java.lang.String, int);
+    public static int v(...);
     public static int d(...);
 }
 


Amejia481 (approved):

Amejia481:
For the intermittent test https://github.com/mozilla-mobile/fenix/issues/17473

csadilek (commented):
Just a heads-up that we have our own LogSink for A-C logs too which would be inconsistent with this as it only strips debug statements, verbose still goes through: https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/java/org/mozilla/fenix/FenixLogSink.kt#L16

This is needed as we don't use Log.d or Log.v everywhere and proguard can't strip those.

csadilek:
I think if we strip .v here we should also adjust the log sink?

Amejia481 (changes requested):
Lets wait until confirming https://github.com/mozilla-mobile/fenix/pull/17452#issuecomment-760503887

mcomella:
Good find, I didn't know LogSink didn't `Log.e` and friends.

> I think if we strip .v here we should also adjust the log sink?

That makes sense to me. `mozilla...Log.Priority` doesn't actually feature VERBOSE though – I think the best course of action is to add it to ac (even though it's unused) and stop it in `FenixLogSink` to avoid the alternative where we don't change in this PR, VERBOSE is added to ac later, and gets logged because we forget to update it here.

@csadilek Do you agree?

One downside of using `AndroidLogSink` is that any string concatenation in log statement strings are always resolved, unlike direct use of `Log.*` where those statements are entirely stripped by Proguard.

We should also consider stripping `Logger.debug` and (if we add it) `Logger.verbose` – what do you think, @csadilek ?

So to summarize, I propose:
- (current PR) Stripping Log.v with proguard
- Adding VERBOSE priority to AndroidLogSink and updating FenixLogSink to ignore it in non-debug mode
- Adding logger.verbose method to AndroidLogSink
- Stripping logger.debug and logger.verbose with Proguard

csadilek:
> That makes sense to me. mozilla...Log.Priority doesn't actually feature VERBOSE though – I think the best course of action is to add it to ac (even though it's unused) and stop it in FenixLogSink to avoid the alternative where we don't change in this PR, VERBOSE is added to ac later, and gets logged because we forget to update it here. @csadilek Do you agree?

Yes, makes sense! I missed that we didn't even support it yet.

> So to summarize, I propose:
    (current PR) Stripping Log.v with proguard
    Adding VERBOSE priority to AndroidLogSink and updating FenixLogSink to ignore it in non-debug mode

Yes, 👍 

> Adding logger.verbose method to AndroidLogSink
Stripping logger.debug and logger.verbose with Proguard

This part I don't get. Do you want to add these methods so we can use proguard and not rely on runtime filtering? We'd have to change a lot of call sites too and then the API usage would be inconsistent. I'd punt on that to be honest unless you want to do that for all log levels, but even then it might make more sense to file that and address later, separately.

mcomella:
> Do you want to add these methods so we can use proguard and not rely on runtime filtering?

Yes

> We'd have to change a lot of call sites too and then the API usage would be inconsistent.

Why would we have to change call sites? I figured it'd just be a fenix optimization to strip these methods and beyond adding `Logger.verbose` wouldn't change anything on the ac side.

> I'd punt on that to be honest unless you want to do that for all log levels, but even then it might make more sense to file that and address later, separately.

I'm not sure what you mean by "do that for all log levels".

mcomella:
@csadilek ^ Not sure if you got pinged on the edit

mcomella:
Actually, that second bit is a premature optimization – let's skip it. The first part is about preventing an unexpected state to occur: that we're logging verbose messages which can leak PII or create an unexpected performance regression. I'll do that, namely:
- (current PR) Stripping Log.v with proguard
- Adding VERBOSE priority to AndroidLogSink and updating FenixLogSink to ignore it in non-debug mode

csadilek:
>  @csadilek What do you mean by "do that for all log levels".

@mcomella Ah I maybe misread what your plan was re: `AndroidLogSink`. Looks like you just wanted to add these methods and call them conditionally e.g. if log level is verbose don't call `println` but `verbose` instead so you can capture with proguard?

I agree with skipping that part though for now.

mcomella:
I created https://github.com/mozilla-mobile/android-components/pull/9451 to add Log.Priority.VERBOSE. After that lands, I'll make the appropriate updates to this PR.

csadilek (approved):
Meant to approve as we discussed everything here and this can land depending on discussion in https://github.com/mozilla-mobile/android-components/pull/9451.

mcomella:
We can land without the ac PR as per the decision in https://github.com/mozilla-mobile/fenix/issues/17451#issuecomment-764033636

mcomella:
@Amejia481 Can you approve so that this can be landed?

Amejia481 (approved):

PR #17468, opened 2021-01-14
Add experimental WebAuthn support for Nightly only

jonalmeida:
When testing out WebAuthn support with the privileged API,
we need our app to be signed by an allowed signing key.

We're seeing our tests fail with this error when testing locally:

```
  [FidoApiImpl] updateTransaction is called for stop
  [FidoApiImpl] finishSecurityKeyRequestController should not be called when SecurityKeyRequestController is null.
```

Our theory is that if we try this code on our signed APK, we should see
it work.

Needs https://github.com/mozilla-mobile/android-components/pull/8714 to land.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/browser/WebAuthnFeature.kt
+++ b/app/src/main/java/org/mozilla/fenix/browser/WebAuthnFeature.kt
@@ -1,0 +1,61 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.browser
+
+import android.app.Activity
+import android.content.Intent
+import android.content.IntentSender
+import mozilla.components.concept.engine.Engine
+import mozilla.components.concept.engine.activity.ActivityDelegate
+import mozilla.components.support.base.feature.ActivityResultHandler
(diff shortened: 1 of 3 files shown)


jonalmeida:
I forgot, Reference Browser is also on the Play Services allowlist, so we can test it safely there as well. 🙂

app/src/main/java/org/mozilla/fenix/browser/WebAuthnFeature.kt
```diff
@@ -0,0 +1,61 @@
...
+import mozilla.components.concept.engine.Engine
+import mozilla.components.concept.engine.activity.ActivityDelegate
+import mozilla.components.support.base.feature.ActivityResultHandler
+import mozilla.components.support.base.feature.LifecycleAwareFeature
+import mozilla.components.support.base.log.logger.Logger
+
+/**
+ * This implementation of the WebAuthnFeature is only for testing in a nightly signed build.
+ *
+ * ⚠️ IT SHOULD NOT GO TO RELEASE.
```
csadilek, line 19: nit: maybe link to the feature flag here, as the code will go to release, it just won't be enabled :).

csadilek (approved):
👍

app/src/main/java/org/mozilla/fenix/browser/WebAuthnFeature.kt
jonalmeida: Good point, this comment will probably cause confusion otherwise.

PR #17483, opened 2021-01-15
Migrate some "selected tab" reads from SessionManager to BrowserStore.

pocmo:
Those looked like some quick wins.


csadilek (approved):
SessionManager-- 🚢

PR #17371, opened 2021-01-08
String Uplifts for Firefox 85

st3fan:
This PR contains string uplifts that are made according to the instructions in the [Release Checklist](https://github.com/mozilla-mobile/fenix/blob/master/.github/ISSUE_TEMPLATE/release_checklist.md)

That translated to the following commands:

```
git clone git@github.com:mozilla-mobile/fenix.git
cd fenix/
git checkout releases_v85.0.0
git checkout -b st3fan/strings-uplift-v85
./l10n-uplift.py st3fan/strings-uplift-v85 --uplift --verbose
git push --set-upstream origin st3fan/strings-uplift-v85
```

The result of the `l10n-uplift.py` tool was:

```
stefan@Discovery ~/M/fenix (st3fan/strings-uplift-v85)> ./l10n-uplift.py st3fan/strings-uplift-v85 --uplift --verbose

Processing l10n commits for 'st3fan/strings-uplift-v85'...
Since 'st3fan/strings-uplift-v85' split off 'master', there were 13 commit(s) from release+l10n-automation-bot@mozilla.com.

Hashes of those commits on 'master' are: ['7bf0f7595c7d91c531530da8236f1061e9479dd3', '5445e61b6aa3d03bbea390200e5536f493a7e38b', '57d453dc7fe3f3d52b9196162b4a8e38b0b88442', '96b63485cc7d761247c8a319b62d16d29e6b6a0c', '33272d5447491f92249da61f13f3c3543ccc36f1', '16ab8bf1abf651e1410705f3b4236667aa386957', '96be55067dd0e1894d7d634567a9e8a3f2164c41', 'f8f677ce7f5200025cbf3efd6a7383e789f0a993', 'bad65b501782dd1945e03cf42bfadd3c28a2408b', '9c776f1d7fe9d6091e1e8b03e6d509dc5a88214d', 'f95547c27103f778bf578dade8c9a7e8b81d43bb', '9966a85bd5b83d51d2fc6d27b41590fdae057b9f', '1fbf227a61c5776c420591b6ce23faa76bd93cac']

Of those, 0 commit(s) already uplifted.
Hashes of commits already uplifted to 'st3fan/strings-uplift-v85': []

Need to uplift 13 commit(s).
[...]

PR #17494, opened 2021-01-15
String Uplifts for Firefox 85 Jan 15th

st3fan:
String Uplifts for Firefox 85 Jan 15th.

I followed the same steps as from https://github.com/mozilla-mobile/fenix/pull/17371 but the script bailed out with errors (conflicts) for previously uplifted string changes. I wonder if this is because we merged that PR incorrectly and as a result the script could not figure out that those commits were already handled?

For this update I looked at the script dry-run output and then manually ran:

```
git cherry-pick 7fc0ef53794d263881e8c3e106dab5d54ca1b78c -x
git cherry-pick 6a84eed5527f54f466e39d03140f18f6bef286ca -x
git cherry-pick 837eaa32824d6b77012bdb549118adea8a2f1ce9 -x
git cherry-pick 2a19967a326ae5b31d8ea254069fa70844b70c0f -x
```

Which is what the script also does. Let's document better what the exact steps are for next time.

### DIFF
--- a/app/src/main/res/values-ast/strings.xml
+++ b/app/src/main/res/values-ast/strings.xml
@@ -591,6 +591,8 @@
     <string name="tab_tray_multiselect_share_content_description">Compartir les llingüetes abiertes</string>
     <!-- Content description for tabs tray multiselect menu -->
     <string name="tab_tray_multiselect_menu_content_description">Menú de llingüetes abiertes</string>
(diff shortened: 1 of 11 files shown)


Delphine:
(just a heads-up that I don't review these uplifts. Thanks!)

csadilek (approved):
👍 Did a quick comparison too to verify.

PR #17532, opened 2021-01-19
Refactor QuickSettingsDialog to use browser store

csadilek:
Just migrating the QuickSettingsDialog and controller to browser state.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/settings/quicksettings/QuickSettingsController.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/quicksettings/QuickSettingsController.kt
@@ -11,6 +11,8 @@ import androidx.navigation.NavController
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.launch
 import mozilla.components.browser.session.Session
+import mozilla.components.browser.state.selector.findTabOrCustomTab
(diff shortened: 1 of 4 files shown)


Amejia481 (approved):

app/src/main/java/org/mozilla/fenix/settings/quicksettings/QuickSettingsController.kt
```diff
@@ -136,7 +139,8 @@ class DefaultQuickSettingsController(
         val permissions = sitePermissions
 
         sitePermissions = if (permissions == null) {
-            val origin = requireNotNull(session?.url?.toUri()?.host) {
+            val tab = browserStore.state.findTabOrCustomTab(sessionId)
```
Amejia481, line 142: Why not falling back to the selected session and avoid the `requireNotNull` after?

csadilek: Because even with `findTabOrCustomTabOrSelectedTab` you'd still need the null check as the return type is still nullable. And also because we're now only creating the controller with a `sessionId` we know exists.

Amejia481: Got it, thanks for the clarification!

Issue #17309, opened 2021-01-05
Intermittent UI test failure - BookmarksTest.deleteMultipleSelectionTest


sv-ohorvath:
### Firebase Test Run:
Flaky, failed 1/2
https://console.firebase.google.com/project/moz-fenix/testlab/histories/bh.66b7091e15d53d45/matrices/6303712677038501074/executions/bs.313bf4093bd5b936/testcases/1/test-cases
### Stacktrace:
`java.lang.AssertionError
at org.junit.Assert.fail(Assert.java:87)
at org.junit.Assert.assertTrue(Assert.java:42)
at org.junit.Assert.assertNotNull(Assert.java:713)
at org.junit.Assert.assertNotNull(Assert.java:723)
at org.mozilla.fenix.helpers.ext.WaitNotNullKt.waitNotNull(WaitNotNull.kt:20)
at org.mozilla.fenix.ui.robots.ThreeDotMenuMainRobot.clickAddBookmarkButton(ThreeDotMenuMainRobot.kt:84)
at org.mozilla.fenix.ui.robots.BrowserRobot$createBookmark$3.invoke(BrowserRobot.kt:299)
at org.mozilla.fenix.ui.robots.BrowserRobot$createBookmark$3.invoke(BrowserRobot.kt:47)`
### Build: 
Master 1/4/21

From the video it looks like the main menu is opened before the browser manages to completely load the second webpage, so the bookmark button is not in the correct state.

See also: https://github.com/mozilla-mobile/fenix/issues/13447 (for the same error)

sv-ohorvath:
Flaky on 1/20: https://console.firebase.google.com/project/moz-fenix/testlab/histories/bh.66b7091e15d53d45/matrices/7150731363913616220/executions/bs.fa03f857eb4d8a88/testcases/1/test-cases

PR #17546, opened 2021-01-20
For #17309, #13447, #14534 UI tests: changed the way page Url is verified

sv-ohorvath:
Ran BookmarksTest#deleteMultipleSelectionTest & SettingsBasicsTest#toggleShowVisitedSitesAndBookmarks and NavigationToolbarTest#goBackTest on Firebase for 50 times and all passed.

### DIFF
--- a/app/src/androidTest/java/org/mozilla/fenix/ui/SettingsBasicsTest.kt
+++ b/app/src/androidTest/java/org/mozilla/fenix/ui/SettingsBasicsTest.kt
@@ -120,6 +120,7 @@ class SettingsBasicsTest {
 
         navigationToolbar {
         }.enterURLAndEnterToBrowser(page2.url) {
+            verifyUrl(page2.url.toString())
         }.openThreeDotMenu {
             clickAddBookmarkButton()
         }
--- a/app/src/androidTest/java/org/mozilla/fenix/ui/robots/BrowserRobot.kt
+++ b/app/src/androidTest/java/org/mozilla/fenix/ui/robots/BrowserRobot.kt
@@ -57,14 +57,14 @@ class BrowserRobot {
         val mDevice = UiDevice.getInstance(InstrumentationRegistry.getInstrumentation())
         sessionLoadedIdlingResource = SessionLoadedIdlingResource()
 
-        mDevice.waitNotNull(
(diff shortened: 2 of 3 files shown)


app/src/androidTest/java/org/mozilla/fenix/ui/SettingsBasicsTest.kt
```diff
@@ -120,6 +120,7 @@ class SettingsBasicsTest {
 
         navigationToolbar {
         }.enterURLAndEnterToBrowser(page2.url) {
+            verifyUrl(page2.url.toString())
```
sv-ohorvath, line 123: From the test recording it looks like the main menu is opened before the browser manages to completely load the second webpage, so the bookmark button is not in the correct state.
Added this step to make sure the page is on the right URL before proceeding to save it as bookmark

app/src/androidTest/java/org/mozilla/fenix/ui/robots/BrowserRobot.kt
```diff
@@ -297,6 +297,8 @@ class BrowserRobot {
     fun createBookmark(url: Uri) {
         navigationToolbar {
         }.enterURLAndEnterToBrowser(url) {
+            // needs to wait for the right url to load before saving a bookmark
+            verifyUrl(url.toString())
```
sv-ohorvath, line 301: Same reason as above

app/src/androidTest/java/org/mozilla/fenix/ui/robots/BrowserRobot.kt
```diff
@@ -57,14 +57,14 @@ class BrowserRobot {
...
         runWithIdleRes(sessionLoadedIdlingResource) {
-            onView(withId(R.id.mozac_browser_toolbar_url_view))
-                .check(matches(withText(containsString(url.replace("http://", "")))))
+            assertTrue(
+                mDevice.findObject(
+                    UiSelector()
+                        .resourceId("$packageName:id/mozac_browser_toolbar_url_view")
+                        .textContains(url.replace("http://", ""))
+                ).waitForExists(waitingTime)
+            )
```
sv-ohorvath, line 67: Simplified the URL verification and added the waitForExists method which proved more solid so far.

AaronMT (approved):

PR #17552, opened 2021-01-20
Closes #17548: Use browser store exclusively in tab tray

### LINKED ISSUE #17548: Refactor tab tray to use browser store exclusively
This is similar to other recent refactorings. We want to replace all reads from `SessionManager` with the corresponding actions on the browser store. There are a few more leftovers here plus some testing effort so filing separately.

csadilek:
I've also refactored the `TabHistoryDialog` here. Refactoring is straight forward, just replacing `SessionManager` reads with the store.


pocmo (approved):
LGTM.

jonalmeida (approved):
✨ 

Tested it out, LGTM!

PR #17558, opened 2021-01-20
Optimize singleton creation pattern

### LINKED ISSUE #17556: Unnecessary resource usage for Singleton creation
https://github.com/mozilla-mobile/fenix/blob/820bb9d38f1b6fa0e9b9747e85f9f4e353416e60/app/src/geckoRelease/java/org/mozilla/fenix/engine/GeckoProvider.kt#L23

`GeckoProvider.getOrCreateRuntime` is annotated with `@Synchronized` for the purpose of creating a singleton of `GeckoRuntime`. As such, this results in every call to the method utilizing unnecessary resources after the class has been instantiated.

A more optimal singleton pattern which retains thread & null safety while not using synchronous resources after instantiation would be:
```kotlin
object GeckoProvider {
    @Volatile
    private var runtime: GeckoRuntime? = null

    fun getOrCreateRuntime(
        context: Context,
        storage: Lazy<LoginsStorage>,
        trackingProtectionPolicy: TrackingProtectionPolicy
    ): GeckoRuntime {
        return runtime ?: synchronized(this) {
            runtime ?: createRuntime(context, storage, trackingProtectionPolicy)
                .also { runtime = it }
        }
    }

    // createRuntime......
}
```

05nelsonm:
Closes #17556

This PR optimizes the creation/referencing to `GeckoProvider.getOrCreateRuntime` as it removes synchronous execution overhead after `GeckoRuntime` class has been instantiated.

### Pull Request checklist

```
Overkill
``` 
```
See #17556 (as linked above) for thorough explanation.
```
```
Not applicable
```

1. click on Show All Checks,

### DIFF
--- a/app/src/geckoNightly/java/org/mozilla/fenix/engine/GeckoProvider.kt
+++ b/app/src/geckoNightly/java/org/mozilla/fenix/engine/GeckoProvider.kt
@@ -24,17 +24,15 @@ object GeckoProvider {
     const val CN_GET_HASH_URL =
         "https://sb.firefox.com.cn/gethash?client=SAFEBROWSING_ID&appver=%MAJOR_VERSION%&pver=2.2"
 
-    @Synchronized
     fun getOrCreateRuntime(
         context: Context,
         storage: Lazy<LoginsStorage>,
         trackingProtectionPolicy: TrackingProtectionPolicy
     ): GeckoRuntime {
-        if (runtime == null) {
-            runtime = createRuntime(context, storage, trackingProtectionPolicy)
+        return runtime ?: synchronized(this) {
+            runtime ?: createRuntime(context, storage, trackingProtectionPolicy)
+                .also { runtime = it }
         }
-
-        return runtime!!
     }
 
     private fun createRuntime(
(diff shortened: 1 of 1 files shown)


app/src/geckoNightly/java/org/mozilla/fenix/engine/GeckoProvider.kt
```diff
@@ -18,23 +18,22 @@ import org.mozilla.geckoview.ContentBlocking
...
 object GeckoProvider {
+    @Volatile
     private var runtime: GeckoRuntime? = null
     const val CN_UPDATE_URL =
         "https://sb.firefox.com.cn/downloads?client=SAFEBROWSING_ID&appver=%MAJOR_VERSION%&pver=2.2"
     const val CN_GET_HASH_URL =
         "https://sb.firefox.com.cn/gethash?client=SAFEBROWSING_ID&appver=%MAJOR_VERSION%&pver=2.2"
 
-    @Synchronized
     fun getOrCreateRuntime(
```
csadilek, line 28: Since this method is invoked only once during startup there should be no noticeable performance difference between `Synchronized` and the `Volatile` approach here.

There should be no relevant lock contention to cause an impact. It's `Synchronized` because it could be the Gecko fetch client or the engine itself that needs it first. The fetch client could be initialized on a different thread.

In general, we don't want to optimize code before we've demonstrated a performance impact. Did you measure this or notice a difference?

csadilek (changes requested):
Thank you for contributing!

For performance optimizations we would like to measure an improvement before landing optimizations and additional complexity (even if minimal), see my comment below.

mcomella:
Triage: we haven't receive a response on our review comments for the PR so we're going to close it for now. If you decide to continue working on this PR, please reopen it! :)

PR #17564, opened 2021-01-21
Refactor ExternalAppBrowserActivity and ExternalAppBrowserFragment to not use Session(Manager).

pocmo:
As part of this change I modified `initializeUI()` to not pass a `Session` in a circle and instead pass the tab into `initializeUI()`. This is still not the best approach, I think. Initializing UI should not be bound to the tab at all, since that can change dynamic. In the cast where the tab  is fixed `External*` those classes can get the instance themselves and so we wouldn't need that in a generic way in `BaseBrowserFragment`. Overall the interaction between `Base*`, `Browser*` and `External*` is quite complicated to understand and follow - due to the inheritance. Eventually we may want to change that. But one thing at a time, I tried to limit the changes here.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
@@ -184,6 +184,7 @@ abstract class BaseBrowserFragment : Fragment(), UserInteractionHandler,
     protected var webAppToolbarShouldBeVisible = true
 
     private val sharedViewModel: SharedViewModel by activityViewModels()
+    private val homeViewModel: HomeScreenViewModel by activityViewModels()
 
     @VisibleForTesting
     internal val onboarding by lazy { FenixOnboarding(requireContext()) }
@@ -220,7 +221,7 @@ abstract class BaseBrowserFragment : Fragment(), UserInteractionHandler,
     }
 
     final override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
-        browserInitialized = initializeUI(view) != null
+        initializeUI(view)
 
         if (customTabSessionId == null) {
             // We currently only need this observer to navigate to home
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
```diff
@@ -918,6 +921,11 @@ abstract class BaseBrowserFragment : Fragment(), UserInteractionHandler,
             updateThemeForSession(selectedTab)
         }
 
+        val init = browserInitialized
+        val theView = view
+        init.hashCode()
+        theView?.hashCode()
```
csadilek, line 927: ^^ Leftover from debugging :)

app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
```diff
@@ -929,7 +932,13 @@ abstract class BaseBrowserFragment : Fragment(), UserInteractionHandler,
             }
         } else {
             view?.let { view ->
-                browserInitialized = initializeUI(view) != null
+                val tab = getCurrentTab()
```
csadilek, line 935: Since we have this block twice what do you think about adding a private function here in `BaseBrowserFragment`:

```kotlin
 private fun initializeUI(view: View): Boolean {
  val tab = getCurrentTab()
  return if (tab != null) {
    initializeUI(view, tab)
      true
    } else {
      false
    }
  }
```

Then we can call it here and in `onViewCreated` as:

```kotlin
browserInitialized = initializeUI(view)
```

csadilek (approved):
Nice! One suggestion below re: `initializeUI`.

There's also one usage of `SessionManager` left in `ExternalAppBrowserFragment`. We pass it to `CustomTabIntegration` which should also be easy to replace? We can do that as a follow-up too. :)

I think my next step will be to remove `getSessionById` and remove all `SessionManager` references from BrowserFragment. We can also mark the `val` in Core.kt as `@Deprecated` after this.

pocmo:
I changed `initializeUI` the way you described. I looked at `CustomTabIntegration` and decided to not migrate this part yet. It currently manipulates the `CustomTabConfig` synchronously. This is a bit hacky and I think instead I will add the required functionality to AC, to allow the overrides that are needed.

csadilek (approved):

Issue #17044, opened 2020-12-15
[Bug] Search engine is placed at the bottom of the list and set as default when added from the "Add search engine"


ebalazs-sv:
### Precondition
Have Google set as default search engine
## Steps to reproduce
1. Launch Fenix and go to Settings -> Search;
2. Delete "Google" from the Default search engine list;
3. Tap "Add search engine" and add "Google";
4. Observe where "Google" is placed in the list;
### Expected behavior
"Google" is placed at the top of the list, where it was before was deleted. Also it is not set as default.
### Actual behavior
"Google" is placed at the bottom of the list and is set as default.

### Notes
After a restart, "Google" is placed at the top but is still set as default.
This issue is reproducible with other search engines from the list.
Video:
![20201215_130532](https://user-images.githubusercontent.com/48799384/102207206-65656780-3ed6-11eb-8138-2d76dc53c63e.gif)

### Device information

* Android device: Huawei P9 Lite (Android 7), OnePlus 5T (android 9);
* Fenix version: Nightly 201215 05:02 (Build #2015781481) GV 85 from 12/15
Not reproducible on the latest Beta and Release.

ebalazs-sv:
This issue is also reproducible on 85.0.0-beta.3 from 12/21 with Samsung Galaxy Tab S3 (Android 8).

softvision-miralobontiu:
This issue is reproducible on Beta 85.0.0-beta.6 with Sony Xperia Z5 Premium (Android 7.1.1).

mcarare:
Note that this happens with any default search engine that is deleted, not just Google.

mcarare:
As for the engine list order, I am not aware of a rule that says Google should be on top. Currently, they are ordered by type (default from region and additional default - the ones in the additional list), and by the time they were added or readded to the list. If UX does not mention any other order (alphabetical, etc) I do not consider it to be a bug the fact that it is not placed at the top of the list.

abodea:
Removing `eng:qa:needed` until further notice due to the reverts from above.

eliserichards:
We should double-check with the Search team with regards to the order of the search engines when deleted/re-added (for example, if we delete google and re-add it, should it still be at the top?)

mcarare:
@pocmo Do you have more insights on what the desired behavior would be on deleting the default engine? If there is no decision to be made regarding an automatic selection of another engine, maybe we can prompt the user to either:

[...]

PR #17586, opened 2021-01-22
Revert "For #17044: Explicitly set a new default engine when default is deleted."

pocmo:
This reverts commit 633bf384e2b5752f5b2e4ebab7abd568869c6986 / PR #17486.

Using the `selectSearchEngine` UseCase is the wrong approach here: The UseCase is for persisting the search engine user choice of the user. We need to distinguish between a default picked by the search component and a user choice. The default picked by the component can change over time based on constraints. The user choice should always be permanent. If we use the UseCase here then we make this choice permanent even though it was only meant as a fallback.

Since the original issue (#17044) is not critical, I recommend to revert and discuss a new solution on the issue.


csadilek (approved):

pocmo:
Failed test: PwaOnboardingObserverTest (-> Known intermittent: #17473)

mcarare:
@pocmo, I am sorry, but I don't really see the problem with selecting another search engine and making that selection permanent. The selection of the first engine in the list was proposed/accepted by UX. The selection is made only when deleting the previous default engine, so the selection would not change over time. The way the search is done makes having a default search engine compulsory, and I also see that deselecting the deleted engine as default is also something that we must do.

We also use a similar mechanism in other parts of the app, for example, we also select another tab when the selected tab is closed, a tab chosen by us, not by the user.

On testing and implementing this I did not see any downfalls to using this, in other components or features. Sorry if I missed something, but I don't see a better approach to this.

Issue #10686, opened 2020-05-15
[Bug] Top dynamic toolbar should use a custom behavior


Mugurell:
When the dynamic toolbar is set at the bottom we use a [BrowserToolbarBottomBehavior](https://github.com/mozilla-mobile/android-components/blob/f8891561284ad13f05a4ea7900174c08d79a0d08/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/behavior/BrowserToolbarBottomBehavior.kt) from a-c which knows to ignore zoom gestures and when to ignore scroll gestures also.

When the dynamic toolbar is set at the top we are using one-off default flags to control scrolling without accounting for page structure or user gestures.

Like this (toolbar is animated by inner scrolls, zoom gestures, etc) https://github.com/mozilla-mobile/fenix/issues/2278#issuecomment-629077440

Or like this (toolbar animations prevent webpage panning)
![fenixTopToolbar](https://user-images.githubusercontent.com/11428869/82069226-ac3e7500-96db-11ea-976d-6fd995d66aec.gif)

Mugurell:
Asking @sblatz about the best approach here since I see he worked recently here.
I think the best approach would be to use a similar "BrowserToolbar`Top`Behavior" but it seems like there might be more changes needed.

Mugurell:
Touched on possible solutions with Sawyer.
It's probable that we haven't gone with a similar layoutBehavior for the top toolbar because setDynamicToolbarMaxHeight only clips from the bottom, not the top.
So this warrants more investigations about what the best solution would be.

Mugurell:
Investigated what would be needed and left a comment on AC - https://github.com/mozilla-mobile/android-components/issues/7718#issuecomment-703614597

ebalazs-sv:
Verified as fixed on Nightly 2/9 and Beta 86.0.0-beta.3 from 2/8 with Pixel 2 (Android 9).
I will close this issue and remove the qa:needed label.

PR #17588, opened 2021-01-22
For #10686 - Use the AC custom behavior for both the top and bottom toolbars

Mugurell:
Use the AC custom behavior for both the top and bottom toolbars

This comes to unify the experience (with improvements but also specific issues)
for the url toolbar irrespective of it being placed at the bottom or at the top
Going further this will ease development and ensure the best UX for users.

Also needs https://github.com/mozilla-mobile/android-components/pull/9483

With these changes the bottom/top toolbars should function the same. Just mirrored.

[video showing the result](https://drive.google.com/file/d/1xYZZDxepoZ09xCFJScFTVa3Ijfp_rzsn/view?usp=sharing)

### Pull Request checklist

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarView.kt
```diff
@@ -40,6 +33,9 @@ import org.mozilla.fenix.ext.settings
 import org.mozilla.fenix.theme.ThemeManager
 import org.mozilla.fenix.utils.ToolbarPopupWindow
 import java.lang.ref.WeakReference
+import mozilla.components.browser.toolbar.behavior.ToolbarPosition as behaviorToolbarPosition
+import mozilla.components.browser.toolbar.behavior.ToolbarPosition.BOTTOM as bottomToolbarPosition
+import mozilla.components.browser.toolbar.behavior.ToolbarPosition.TOP as topToolbarPosition
```
pocmo, line 38: nit: Wondering about those named imports instead of `ToolbarPosition.*`? Makes it harder to read, at least in the PR, since those look like class members in code.

pocmo (approved):

app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarView.kt
Mugurell: I understand.
Thing is there already exists a aptly named `ToolbarPosition` enum in Fenix so I had to use this import aliases.
Searching now in the project using `import .* as` I see usages such as `..search.SearchEngine as LegacySearchEngine`. maybe I could use something like `MozacToolbarPosition`?

Mugurell: Went with using `MozacToolbarPosition` as import alias.

PR #17622, opened 2021-01-25
For #16615 & #11191: re-enabled ReaderViewTest and added a UI smoke test

sv-ohorvath:
> Just a heads-up that the tests will need some updates as the UI has changed a while ago. Reader mode button is now in the toolbar directly.
https://github.com/mozilla-mobile/fenix/issues/11191#issuecomment-736097268

Refactored and re-enabled the readerView tests, moved one to the SmokeTests class.
Ran on Firebase for 20 times and all passed: [runs](https://github.com/mozilla-mobile/fenix/runs/1749471331)


isabelrios (approved):
LGTM, great work enabling these tests again and adding to Smoketest. Thanks!

Issue #16551, opened 2020-11-13
Gradle: Task caching not working


pocmo:
When running a task like the `test` task a second time, without any code changes, then this task should complete without actually running any tests because Gradle sees their are no changes. This can be observed in AC when running the tests for a component twice. However something seems to break that in Fenix, running the `test` task a second time will run all the tests again.

mawen7:
@pocmo Build cache is not enabled on fenix, see #11621.

#17627 might help a bit, if you want to take a look.

pocmo:
Afaik the build cache is independent from that - for keeping backups of `build/` state even when code changes. But in this case here running the task twice without changing any code should not cause a rerun of everything afaik.

mawen7:
@pocmo You're right it works on AC even with build cache disabled (beside modules using updateExtensionVersion). Sorry for confusion I mixed things up.

Output for `test -i` with static `BUILD_DATE` and #17627:
`Task ':app:testReleaseUnitTest' is not up-to-date because:
  Task.upToDateWhen is false.` (same output for all variants)

Seems like like AGP or Gradle set `upToDateWhen` to false.
Same behavior can be observed on AC if running `testDebugUnitTest` while `testReleaseUnitTest` is up-to-date.

Nevertheless `BUILD_DATE` breaks up-to-date checks for non-debug variants and without #17627 up-to-date checks are broken for every variant.

Maybe `BUILD_DATE` should be replaced with git hash like in AC?

grigoryk:
I can't replicate this anymore. Perhaps this got fixed via https://github.com/mozilla-mobile/fenix/pull/22601

PR #17627, opened 2021-01-25
Incrementalify updateExtensionVersion

mawen7:
Before this PR extension version was updated on every build, causing merge assets task to be run on every build.

With this PR extension version is only updated if a file changed except the target `manifest.json`, resulting in faster incremental builds.

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -733,22 +733,31 @@ if (gradle.hasProperty('localProperties.autoPublish.application-services.dir')) 
 // https://github.com/mozilla-mobile/android-components/issues/7249
 ext.updateExtensionVersion = { task, extDir ->
     configure(task) {
-        from extDir
-        include 'manifest.template.json'
-        rename { 'manifest.json' }
-        into extDir
+        final MANIFEST_FILE = 'manifest.json'
+        final MANIFEST_TEMPLATE = 'manifest.template.json'
 
-        def values = ['version': AndroidComponents.VERSION + "." + new Date().format('MMddHHmmss')]
-        inputs.properties(values)
-        expand(values)
+        inputs.dir project.fileTree(dir: extDir, exclude: MANIFEST_FILE)
+        outputs.file MANIFEST_FILE
+
+        doLast {
+            copy {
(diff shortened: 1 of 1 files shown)


csadilek (changes requested):
Thank you for contributing, @mawen7.

I am leaning towards not landing this here. The two built-in extensions should ideally be uplifted to our Android components (A-C) project: https://github.com/mozilla-mobile/android-components/issues/7249

In A-C we use the same mechanism to generate version numbers and it's important to have this working the same name for all extensions.

How big of a difference do you see with these changes?

The most important part here is that a new version is being generated:
- During development so that every change results in the extension being re-installed
- For all release builds

mawen7:
Thanks for your review @csadilek

> In A-C we use the same mechanism to generate version numbers and it's important to have this working the same name for all extensions.

I noticed this breaks the up-to-date check (caching) for tasks like `test` in modules with extension like `feature-readerview` in AC as well.
Should we fix this in AC first to fix task caching for these modules?
Afterwards it might make more sense to keep fenix in sync while waiting for the uplift as I'm not familiar with the procedure to uplift these and https://github.com/mozilla-mobile/android-components/issues/7249 is still awaiting triage.


> How big of a difference do you see with these changes?

Differences are around these numbers:
No changes: 2x faster 30s->15s
Bigger absolute differences are noticeable if there are code or jni (new AC release) changes:
With Code changes: 4min->~3min
New AC release: 10min->~8min

I think this increases due to larger memory pressure on larger changes.

Maybe more important this change also is one step towards fixing #16551

> The most important part here is that a new version is being generated:
> 
>     * During development so that every change results in the extension being re-installed

I verified `manifest.json` is updated with a new version if any file in the passed `extDir` changes just except mentioned `manifest.json` to avoid recursion.

> 
>     * For all release builds

Since release builds are mainly done on CI non incremental causing the version to be updated always I didn't thought much about this yet.
If you want to ensure version is updated on every AC release we could add back AC version to `inputs.properties` as before but without the date.

PR #17631, opened 2021-01-26
Refactor HomeFragment to use browser store

csadilek:
Mostly mechanical, simply switching to use cases to remove `SessionManager` dependency.


pocmo (approved):
👍  LGTM.

Issue #17418, opened 2021-01-12
Fenix Google Search tile [Telemetry]


lwa-moz:
## Description & Product Manager / Data Scientist User Story
Add telemetry to the existing Google search tile that allows for attribution in a similar fashion to the existing telemetry in other search access points, such as the Awesome Bar.
## What questions will you answer with this data?
Count of users conducting searches from the Google Search tile.
## Acceptance Criteria
- [ ] ENG files a [DS JIRA](https://jira.mozilla.com/projects/DO/issues/DO-228?filter=allopenissues) request outlining their methodology.
- [ ] DS sign off on instrumentation methodology addressing product questions.
- [ ] Event pings can be queried via re:dash
- [ ] Event pings can be queried via amplitude
- [ ] We are sending telemetry events for the actions listed in the requirements
- [ ] We have documented the telemetry
- [ ] We have asked a data steward to [review](https://github.com/mozilla/data-review/blob/master/request.md) the telemetry

codrut-topliceanu:
@lwa-moz Apologies, but could you please provide a bit more info about what you mean when you say "Google search tile"? Do you mean the topSites?

codrut-topliceanu:
Sorry to bother you again @lwa-moz , but I'm not entirely sure what you mean by "attribution in a similar fashion to the existing telemetry in other search access points, such as the Awesome Bar." Below I put up a link to all the Awesomebar related metrics I found and all of them seem to be durations.

Would you mind being more specific with the metrics you'd like to see implemented?

https://github.com/mozilla-mobile/fenix/blob/7c56df41df575aa00cd740540d99d0a16722965f/app/metrics.yaml#L3903

mcarare:
@lwa-moz I am not sure if I am missing some details here, but isn't this the same as [Implement top site search engine attribution ](https://github.com/mozilla-mobile/fenix/issues/17304) ?

gabrielluong:
I think this is calling for telemetry for the "https://www.google.com" top site, but we should wait for further clarifications. You will see in https://github.com/mozilla-mobile/fenix/issues/17304 that we do an exact string match since the default Google top site can be removed and re-added.  

> ## Description & Product Manager / Data Scientist User Story
> Add telemetry to the existing Google search tile that allows for attribution in a similar fashion to the existing telemetry in other search access points, such as the Awesome Bar.
> 

I don't know much about the existing telemetry regarding search access points, but it might be good to take a look what is happening there.

[...]

PR #17637, opened 2021-01-26
For #17418 - Added telemetry for Google Default Top Site

codrut-topliceanu:
### Pull Request checklist

- [x] **Tests**: This PR doesn't include any tests as it is only a minor change
- [x] **Screenshots**: The code in this PR does not include any user facing features. 

1. click on Show All Checks,


codrut-topliceanu:
@lwa-moz I'm not entirely sure what to write in the data review form.

lwa-moz:
@codrut-topliceanu this [PR](https://github.com/mozilla-mobile/fenix/pull/10112) has a good example of a completed data collection review forms. I believe the form inputs in this case are almost identical to those listed in the linked PR.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlController.kt
```diff
@@ -372,6 +372,10 @@ class DefaultSessionControlController(
             TopSite.Type.PINNED -> metrics.track(Event.TopSiteOpenPinned)
         }
 
+        if (url == SupportUtils.GOOGLE_XX_URL) {
+            metrics.track(Event.TopSiteOpenGoogleSearchAttribution)
+        }
```
eliserichards, line 377: Are there tests for the `DefaultSessionControlController` that we can add for this?

eliserichards (changes requested):

eliserichards:

- This probe will be used to better model how the Top Sites location contributes to high-level trends in search. This will help us understand the revenue impact of our Top Sites with regards to Search.

This helps us understand our users' needs and the impact it has on our Search revenue and trends.

- No current alternative methods.

- Currently there are no events to track specifics for the Google top site.

5) **List all proposed measurements and indicate the category of data collection for each measurement, using the Firefox [data](https://wiki.mozilla.org/Firefox/Data_Collection)[collection ](https://wiki.mozilla.org/Firefox/Data_Collection)[categories](https://wiki.mozilla.org/Firefox/Data_Collection) found on the Mozilla wiki.**
- All data is Category 2

- Until 2020-08-01 with the option to renew at that point.

-  All release, beta, and nightly users with telemetry enabled.

- Glean / Amplitude

-  Only on Glean, Amplitude and with mobile teams.

app/metrics.yaml
```diff
@@ -3415,6 +3415,19 @@ top_sites:
     notification_emails:
       - fenix-core@mozilla.com
     expires: "2021-08-01"
+  open_google_search_attribution:
+    type: event
+    description: |
+      A user opened the default google top site
```
gabrielluong, line 3421: We should also revise the description

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlController.kt
gabrielluong: There should be

gabrielluong (changes requested):

gabrielluong:
I have asked megan from DS to take a look in case we are missing anything that is needed for the search telemetries.

st3fan:
I added the `pr:do-not-land` label because I think this is not the correct telemetry to have. I think what we need instead is this one:

https://github.com/mozilla-mobile/fenix/blob/master/app/metrics.yaml#L903-L931

That metric will collect counts for engines and the _source_ to identify how the search was initiated. We probbaly need to introduce a `topsites` source for this.

boek (approved):
# Data Review Form (to be filled by Data Stewards)

Yes, metrics.yaml and metrics.md

Yes, data collection settings

Has expiry

4) Using the **[category system of data types](https://wiki.mozilla.org/Firefox/Data_Collection)** on the Mozilla wiki, what collection type of data do the requested measurements fall under?
Type 2

Default on

6) Does the instrumentation include the addition of **any *new* identifiers** (whether anonymous or otherwise; e.g., username, random IDs, etc.  See the appendix for more details)?
No

7) Is the data collection covered by the existing Firefox privacy notice? **If unsure: escalate to legal if:**
Yes

8) Does there need to be a check-in in the future to determine whether to renew the data?
Fenix team will check-in at expiry

9) Does the data collection use a third-party collection tool?
No

codrut-topliceanu:
Thanks everyone for helping me with this patch. I really appreciate it. ❤️ 

@st3fan I have refactored my patch to use the `search_count` metric as requested. Please let me know if any other changes are needed.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlController.kt
```diff
@@ -376,6 +378,25 @@ class DefaultSessionControlController(
...
+                .store
+                .state
+                .search
+                .searchEngines
+
+            val searchAccessPoint = Event.PerformedSearch.SearchAccessPoint.TOPSITE
+            availableEngines.firstOrNull { engine -> engine.suggestUrl?.contains(baseURL) == true }
+                ?.let { searchEngine ->
+                    searchAccessPoint.let { sap ->
+                        MetricsUtils.createSearchEvent(searchEngine, store, sap)
```
st3fan, line 395: I am not super familiar with this code, but I want to make sure that this is implemented in a way where this event is fired when a search is done and not when the user taps on the tile. A tap may not always lead to a search.

The definitive point where to send this event is when the search is submitted to the search engine.

st3fan (changes requested):

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlController.kt
eliserichards: Yes, I agree. I think our tracking of this should happen in the `SearchDialogController` where we are checking the other `searchAccessPoint`s: https://github.com/mozilla-mobile/fenix/blob/acae32095038ffef7569152aa43c5650bdac72fe/app/src/main/java/org/mozilla/fenix/search/SearchDialogController.kt#L98

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlController.kt
```diff
@@ -376,6 +378,25 @@ class DefaultSessionControlController(
             metrics.track(Event.PocketTopSiteClicked)
         }
 
+        val baseURL = url.tryGetHostFromUrl()
+        if (SupportUtils.GOOGLE_URL.contains(baseURL)) {
```
gabrielluong, line 382: I think we should just do an exact match of `SupportUtils.GOOGLE_URL` for now since that is the only URL we append the search attribution for at the moment, but this will be useful later. One issue I can see if we did move forward with this is that we would be counting visits to something like https://www.google.com/maps, which will not have the search attribution as intended.

gabrielluong (commented):
Can we also add a test for this?

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlController.kt
```diff
@@ -378,6 +380,27 @@ class DefaultSessionControlController(
...
+
+            val searchAccessPoint = Event.PerformedSearch.SearchAccessPoint.TOPSITE
+            val event =
+                availableEngines.firstOrNull { engine -> engine.suggestUrl?.contains(baseURL) == true }
+                    ?.let { searchEngine ->
+                        searchAccessPoint.let { sap ->
+                            MetricsUtils.createSearchEvent(searchEngine, store, sap)
+                        }
+                    }
+            event?.let { activity.metrics.track(it) }
```
gabrielluong, line 401: We should be able to add a test for this in SessionControlControllerTest. There should be something there already for metrics.track(Event.TopSiteOpenGoogle) as well.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlController.kt
```diff
@@ -378,6 +380,27 @@ class DefaultSessionControlController(
...
+                .searchEngines
+
+            val searchAccessPoint = Event.PerformedSearch.SearchAccessPoint.TOPSITE
+            val event =
+                availableEngines.firstOrNull { engine -> engine.suggestUrl?.contains(baseURL) == true }
+                    ?.let { searchEngine ->
+                        searchAccessPoint.let { sap ->
+                            MetricsUtils.createSearchEvent(searchEngine, store, sap)
+                        }
+                    }
```
pocmo, line 400: At the end of the day this code searches the Google `SearchEngine` instance only, with the check above. Should we simplify this?

If we want to keep it generic for supporting more search engines in the future then:
* Should we check `resultUrls` instead of `suggestUrl` since the latter one is the URL we query search suggestions from and doesn't have to be related to the search engine URL. The result URLs may be closer.
* Should we also search in `*.state.search.availableEngines`? Otherwise you will not find a search engine if the user decided to hide it.
* In `appendSearchAttributionToUrlIfNeeded()` we already change the URL to the one with the search attribution, so we already know the context, it seems complicated to then check this here again. 🤔 This would also allow us to directly lookup the search engine (e.g. Google) and not have to search for it here, I assume?

st3fan: Very much agreed on this. I filed https://github.com/mozilla-mobile/fenix/issues/17956 for this work so that we can move this patch into production sooner. I sent out an email to explain my thoughts behind that.

gabrielluong (approved):

st3fan (approved):
Wonderful work - let's get it on Nightly 🤩

codrut-topliceanu:
Thank you everyone who helped with this!! 🤗 You're all awesome!

PR #17652, opened 2021-01-27
[RELENG-169] support staging release promotion projects

escapewindow:
Let's make mozilla-releng/staging-fenix the official staging repository.
By supporting it in the automation, we no longer need to maintain a
staging patchset to test things properly.


jmaher (approved):
just like android-components, I like it

Issue #9188, opened 2020-03-16
Fix intermittent fail on ReaderViewTest.verifyReaderViewAppearanceFontToggle UI test 


rpappalax:
This test passes on API 28 but not on API 25 

```
 androidx.test.espresso.base.DefaultFailureHandler$AssertionFailedWithCauseError
androidx.test.espresso.base.DefaultFailureHandler$AssertionFailedWithCauseError: 'view has effective visibility=VISIBLE' doesn't match the selected view.
Expected: view has effective visibility=VISIBLE
Got: "AppCompatImageView{id=2131297023, res-name=notification_dot, desc=Highlighted, visibility=GONE, width=0, height=0, has-focus=false, has-focusable=false, has-window-focus=true, is-clickable=false, is-enabled=true, is-focused=false, is-focusable=false, is-layout-requested=true, is-selected=false, layout-params=android.widget.FrameLayout$LayoutParams@a1d4c6d, tag=null, root-is-layout-requested=false, has-input-connection=false, x=0.0, y=0.0}" 
```
https://console.firebase.google.com/u/0/project/moz-fenix/testlab/histories/bh.66b7091e15d53d45/matrices/7649116021264846893/executions/bs.86d4d41684492d44/testcases/1/test-cases

AaronMT:
I think `assertReaderViewDetected` needs to wait until the appearance of the blue dot on the 3 dot menu (if it's still part of the feature).

AaronMT:
Reader Mode entry/exit is changed in product so the test would need to be updated. This might not be an issue anymore. Reader mode tests currently blocked by https://github.com/mozilla-mobile/fenix/issues/11191

sv-ohorvath:
#11191 was fixed but this test is flaky now on API 28: https://console.firebase.google.com/project/moz-fenix/testlab/histories/bh.66b7091e15d53d45/matrices/8528873179388097367/executions/bs.c4b507f8aa6ea73/testcases/2/test-cases

sv-ohorvath:
verifyReaderViewAppearanceColorSchemeChange failed 1 out of 2 tries with the same error, on Master 1/27: https://console.firebase.google.com/project/moz-fenix/testlab/histories/bh.66b7091e15d53d45/matrices/5053320726265191428/executions/bs.c2ae71640776911a/testcases/1/test-cases

sv-ohorvath:
[...]

PR #17661, opened 2021-01-27
For #9188: wait for page load in Reader mode

sv-ohorvath:
For #9188: Reader View tests needed to wait for the page to finish loading in reader mode before the Appearance button is visible.


mcarare (approved):
👍

PR #17665, opened 2021-01-27
Remove SessionManager from BrowserFragment/ToolbarGestureHandler

csadilek:
Just switching to using the store in `ToolbarGestureHandler`.


eliserichards (approved):
Awesome! 👍

Issue #17667, opened 2021-01-27
[Bug] Addons metrics for install and enable set to false


eliserichards:
Followup to #16832

## Prerequisites
Have a couple of add-on installed and enabled

## Steps to reproduce
Generate the **`metrics`** ping

### Expected behavior
The following **`boolean`** probes should be properly updated:
```
"metrics": {
        "boolean": {
          ...
          "addons.has_enabled_addons"
          "addons.has_installed_addons"
         ...
        }
```

### Actual behavior
The above mentioned probes remain set to false

## Notes
This metric  should **always reflect the current state of the app**. Currently, we are only evaluating this metric (and others) on startup and this can lead to delayed/incorrect metrics.

PR #17669, opened 2021-01-27
For #17667: update metrics when an addon is installed or enabled

eliserichards:
For #17667 

We are not updating the `addons.has_enabled_addons` and `addons.has_installed_addons` metrics in all of the correct places. This means that the data for these fields could be out of date when the metrics ping is generated. This should resolve the issue.

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/addons/AddonsManagementFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/addons/AddonsManagementFragment.kt
@@ -29,6 +29,7 @@ import mozilla.components.feature.addons.ui.AddonInstallationDialogFragment
 import mozilla.components.feature.addons.ui.AddonsManagerAdapter
 import mozilla.components.feature.addons.ui.PermissionsDialogFragment
 import mozilla.components.feature.addons.ui.translateName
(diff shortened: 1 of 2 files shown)


gabrielluong (commented):
Should we also be removing this from https://github.com/mozilla-mobile/fenix/blob/124c2baf304704b807cbf5934ff2798b87a6845d/app/src/main/java/org/mozilla/fenix/components/metrics/MetricController.kt#L213?

eliserichards:
No, that should stay to get the startup metrics. This covers the cases where the metrics are generated after changes have been made by the user (so we are always showing the most current state of the app, regardless of when the metrics ping is generated).

gabrielluong (commented):
I imagine we should have some tests for this as well that we can add to?

eliserichards:
No, there are not tests for the `InstalledAddonDetailsFragment`. A test exists for the `AddonManagementFragment` which tests the two error messages, but this `positiveButtonClick` private method is not includable (includible?) in that (and we do not typically like to have fragment tests since the logic shouldn't all be put in the fragment 😞 )

gabrielluong (approved):

PR #17679, opened 2021-01-28
For #9188: wait for page content to appear in Reader mode tests

sv-ohorvath:
Follow-up for #17661 which didn't fix the problem entirely.
Ran the test with this change 70 times and all passed: https://github.com/mozilla-mobile/fenix/runs/1783288191


mcarare (approved):
LGTM

Issue #17092, opened 2020-12-17
[Bug]Improperly displayed Tabs Tray, Top Sites, and search engine "More options" overlay


AndiAJ:
## Prerequisites
Have fresh install or a clean profile

## Steps to reproduce
1. Open the Tabs Tray
2. Tap the "More options" overlay
3. Check the position of the overlay on the screen

### Expected behavior
The overlay should emerge from the ⋮ "More options" button

### Actual behavior
The overlay is misplaced, displayed on the upper right corner of the screen 

### Device information

* Android device: 
• OnePlus A3 (Android 6.0.1)
• Xiaomi Mi Pad 2 (Android 5.1)

* Fenix version: 
• Nightly 201217
• 85.0.0-beta.1

### Notes
❗ Reproducible after long tapping any Top Site
❗ Reproducible with both List and Grid view

► Video
![20201217-102507](https://user-images.githubusercontent.com/51314259/102463943-67edcb80-4054-11eb-9f4f-0eabb5dee6bf.gif)

ebalazs-sv:
This issue is reproducible on Beta 85.0.0-beta.6 from 1/8 with Nexus 5 (Android 6.0.1).

rocketsroger:
related https://github.com/mozilla-mobile/android-components/issues/9439

codrut-topliceanu:
@rocketsroger 
Hey, I've spent some time looking into this issue.
The fix I've found is in AC in `BrowserMenu.kt`

```
private fun PopupWindow.showAtAnchorLocation(anchor: View, isCloserToTop: Boolean) {
    val locationOnScreen = IntArray(2)
    anchor.getLocationOnScreen(locationOnScreen)

    if (isCloserToTop) {
        animationStyle = R.style.Mozac_Browser_Menu_Animation_OverflowMenuTop
        showAtLocation(anchor, Gravity.NO_GRAVITY,
        locationOnScreen[0], locationOnScreen[1])
    } else {
        animationStyle = R.style.Mozac_Browser_Menu_Animation_OverflowMenuBottom
        showAtLocation(anchor, Gravity.NO_GRAVITY,
            locationOnScreen[0], locationOnScreen[1])
    }
}
```

rocketsroger:
> @rocketsroger
> Hey, I've spent some time looking into this issue.
> The fix I've found is in AC in `BrowserMenu.kt`

[...]

Issue #15344, opened 2020-09-23
[Bug][Tablet] Main menu is not displayed correctly after rotating the device from landscape to portrait


sflorean:
## Steps to reproduce
1. Open Fenix and rotate the tablet to landscape mode.
2. Tap on the three-dot menu.
3. Rotate the tablet to portrait mode.

### Expected behavior
The menu is correctly displayed after rotation ( bottom-right).

### Actual behavior
The menu is displayed in the right-center of the page.

### Device information
* Android device: Xiaomi Mi Pad 2 (Android 5.1).
* Fenix version: Nightly 9/23, 82.0.0-beta.1.

### Note
- on phones, this is only reproducible for the Add-ons menu.

![Screenshot_2020-09-23-12-39-12](https://user-images.githubusercontent.com/58466444/93998302-0394d600-fd9d-11ea-89f4-95be900367f2.png)

mcarare:
@sflorean is this still reproducible? I've tried to reproduce it on an emulator tablet but the menu shows up in the correct place. Thank you!

sflorean:
This is still reproducible, not exactly the same step but when rotating from portrait to landscape. 
Device: Huawei MediaPad M2 (Android 5.1.1). 
![Screenshot_2020-11-23-17-08-19](https://user-images.githubusercontent.com/58466444/99991491-c01b1000-2dbd-11eb-8839-c8211c676de7.png)

Mugurell:
Should be fixed after the change in AC.

sflorean:
Verified as fixed on Nightly 12/14.

softvision-miralobontiu:
@codrut-topliceanu asked if we could test again if the issue was specific to Huawei, and Xiaomi tables.
We tested on 11/23 build from @5:58 PM, with Samsung Galaxy Tab A6 (Android 5.1.1), and the original issue is reproducing also.

PR #17682, opened 2021-01-28
For #17092: Dismiss toolbar menu when configuration change occurs

rocketsroger:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
@@ -7,6 +7,7 @@ package org.mozilla.fenix.home
 import android.animation.Animator
 import android.content.Context
 import android.content.DialogInterface
+import android.content.res.Configuration
 import android.graphics.drawable.BitmapDrawable
 import android.graphics.drawable.ColorDrawable
 import android.os.Bundle
@@ -166,6 +167,9 @@ class HomeFragment : Fragment() {
 
     private val topSitesFeature = ViewBoundFeatureWrapper<TopSitesFeature>()
 
+    @VisibleForTesting
+    internal var getMenuButton: () -> MenuButton? = { menuButton }
+
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
 
@@ -264,6 +268,12 @@ class HomeFragment : Fragment() {
 
         activity.themeManager.applyStatusBarTheme(activity)
         return view
+    }
+
+    override fun onConfigurationChanged(newConfig: Configuration) {
+        super.onConfigurationChanged(newConfig)
+
+        getMenuButton()?.dismissMenu()
     }
 
     private fun dismissTip(tip: Tip) {
--- a/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
@@ -6,6 +6,7 @@ package org.mozilla.fenix.browser
 
 import android.content.Context
 import android.content.Intent
+import android.content.res.Configuration
 import android.os.Build
 import android.os.Bundle
 import android.view.Gravity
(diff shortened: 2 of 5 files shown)


codrut-topliceanu:
👍 Works great (correctly dismisses the `BrowserMenu` popUp) for "more options" tab tray overflow and 3 dot menu. There are a few more places where `BrowserMenu` is used though (such as collections 3 dot menu and search engine radio button selection). Will this fix need to be implemented for those places too?

Secondly, while this patch fixes the [original issue](https://github.com/mozilla-mobile/fenix/issues/15344) I suppose a second patch is needed to remove the workaround for that issue in order to fix [the current problem](https://github.com/mozilla-mobile/fenix/issues/17092), right? 🤷‍♂️

rocketsroger:
> 👍 Works great (correctly dismisses the `BrowserMenu` popUp) for "more options" tab tray overflow and 3 dot menu. There are a few more places where `BrowserMenu` is used though (such as collections 3 dot menu and search engine radio button selection). Will this fix need to be implemented for those places too?
> 
> Secondly, while this patch fixes the [original issue](https://github.com/mozilla-mobile/fenix/issues/15344) I suppose a second patch is needed to remove the workaround for that issue in order to fix [the current problem](https://github.com/mozilla-mobile/fenix/issues/17092), right? 🤷‍♂️

Tabtray already does dismiss it https://github.com/mozilla-mobile/fenix/blob/e188185b15e48622ed1c804777df6f48e78d365d/app/src/main/java/org/mozilla/fenix/tabtray/TabTrayDialogFragment.kt#L179.  Which is the other 3 dot menu?

rocketsroger:
> > (such as collections 3 dot menu and search engine radio button selection). Will this fix need to be implemented for those places too?

rocketsroger:
> But there may be other places where `BrowserMenu` is used that I haven't tested.
> Wish we could fix the problem at its root somehow. :(

rocketsroger:
Looking into this more I think I'll fix the other two you raised in separate issues.  The reason here is the workaround for Android M and below broke both collection and search 3 dot menu. (Without rotation)  This is bad because just tapping on the overflow 3 dot menu is the normal use case and rotating is the less encountered use case.  

The workaround needs to be removed immediately and I will see if I can uplift it to beta to fix these issues in Android M or below.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -266,6 +272,12 @@ class HomeFragment : Fragment() {
         return view
     }
 
+    override fun onConfigurationChanged(newConfig: Configuration) {
```
Amejia481, line 275: nit: Should we also add a test similar to the one that we have on `BrowserFragmentTest`?

Amejia481 (approved):
✅

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -187,6 +191,8 @@ class HomeFragment : Fragment() {
         val activity = activity as HomeActivity
         val components = requireComponents
 
+        menuButton = view.findViewById<MenuButton>(R.id.menuButton)
```
jonalmeida, line 194: nit: I don't think you need to use findViewById. Synthetics should let you find the view in your `onConfigurationChanged` call itself.

jonalmeida (approved):
LGTM, just one question/nit. 🙂

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
rocketsroger: Agreed.  That is much cleaner.  Thanks I'll update.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
rocketsroger: Sure, not sure why I didn't see that.  I'll add the test.

Amejia481: Sorry my bad I missed the test

Issue #12822, opened 2020-07-22
[Spike] Android 11: Scoped storage


Mugurell:
Based on the current cycle targeting Android 11 is expected to be a requirement in Q3 2021 - https://support.google.com/googleplay/android-developer/answer/113469#targetsdk.
Once we'll target Android 11 we need to support [scoped storage](https://developer.android.com/training/data-storage#scoped-storage)

## Description
Eng needs to investigate 

* our current interactions with Android filesystem 
* what recommended approaches should we use for Android 11 - https://developer.android.com/training/data-storage/use-cases#migrate-legacy-storage
* if there are any legacy storage locations to be migrated.

## Deliverables
* Reach out to Product to go over results of investigation.
* Open new tickets for addressing each part of the needed work

abodea:
Verified as fixed on the latest Nightly with Google Pixel 4 XL (11).
Note that I was able to upload/download files without issues.

Issue #12772, opened 2020-07-21
FNX2-15459 ⁃ [Bug] Cannot open file after download is complete


Mugurell:
## Steps to reproduce
Only reproduceable on Android 10 (possibly on Android 11 also)

* start a download (for any type of file), wait for the download to complete
* try to open the just downloaded file from the dialog / notification

### Expected behavior
Assuming an app that can open that file is installed the file should be opened.

### Actual behavior
The file is tried to be opened by 3rd party apps, opening fails.

### Device information
* Android device: Pixel 3, Android 10
* Fenix version: latest Nightly

softvision-miralobontiu:
Verified as fixed on the latest Nightly build from 7/29, with Google Pixel (Android 10).
I've downloaded a .pdf file, a .avi file, a .mov file, a .jpg file, for which I had the proper apps installed.
I was able to open all the downloaded files from the notification.

Mugurell:
Closing as per https://github.com/mozilla-mobile/fenix/issues/12772#issuecomment-665713574

PR #17701, opened 2021-01-29
For #12822 - Opt-in of scoped storage

Amejia481:
We want to move to opt in to use the scoped storage as we have partial support on [AC](https://github.com/mozilla-mobile/android-components/blob/69790323190b3f23efac3c12af818bc34b48a43b/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt#L804), we want to discover early any issues related to it, as it will block us from fully supporting Android 11.

I did multiple tests in Android 11 and 10:

* Download and open files. 
* Upload files.
* Upload pictures taking from the camera.

I wasn't able to find any issues after removing the flag, to be sure we want to start testing on nightly and monitoring to se if we can progress to other channels.

I wasn't able to reproduce https://github.com/mozilla-mobile/fenix/issues/12772 after applying this patch https://github.com/mozilla-mobile/android-components/pull/9558, the issue that we have right now are related to content type sanitation. 

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -46,7 +46,8 @@ android {
         def deepLinkSchemeValue = "fenix-dev"
         buildConfigField "String", "DEEP_LINK_SCHEME", "\"$deepLinkSchemeValue\""
         manifestPlaceholders = [
-                "deepLinkScheme": deepLinkSchemeValue
+                "deepLinkScheme": deepLinkSchemeValue,
+                "requestLegacyExternalStorage": true
         ]
 
         // Build flag for "Mozilla Online" variants. See `Config.isMozillaOnline`.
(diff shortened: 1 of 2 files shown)


app/build.gradle
```diff
@@ -87,7 +88,7 @@ android {
             buildConfigField "boolean", "USE_RELEASE_VERSIONING", "true"
             def deepLinkSchemeValue = "fenix-nightly"
             buildConfigField "String", "DEEP_LINK_SCHEME", "\"$deepLinkSchemeValue\""
-            manifestPlaceholders = ["deepLinkScheme": deepLinkSchemeValue]
+            manifestPlaceholders = ["deepLinkScheme": deepLinkSchemeValue, "requestLegacyExternalStorage": false]
```
Mugurell, line 91: Should we maybe set 'false' also for debug to allow any new issues related to this be caught in development?

app/build.gradle
```diff
@@ -103,7 +104,8 @@ android {
                     //  - https://issuetracker.google.com/issues/36924841
                     //  - https://issuetracker.google.com/issues/36905922
                     "sharedUserId": "org.mozilla.firefox.sharedID",
-                    "deepLinkScheme": deepLinkSchemeValue
+                    "deepLinkScheme": deepLinkSchemeValue,
+                    "requestLegacyExternalStorage": true
```
Mugurell, line 108: I see `"requestLegacyExternalStorage": true` being already set for `defaultConfig` from which we'll inherit it so maybe this is not needed. Also below.

Mugurell (approved):
Also did some manual testing and saw no issues.
Seeing that we already use`MediaStore` and not accessing external files directly on AndroidQ+ so it looks like we're in good shape for Android 11.

app/build.gradle
Amejia481: I tried to added it to the `debug` version but for some reason the value were not set (It kept reporting `true)`,  not sure if it has do with other variants having  `releaseTemplate >>`, I will give it another try today.

app/build.gradle
Amejia481: I just wanted to be extra careful as the whole array is being created here, avoid possible issues in case it gets deleted from the `defaultConfig`, and as a result leaking this value to stable versions.

app/build.gradle
Amejia481: Addressed!

PR #17733, opened 2021-02-01

MickeyMoz:


Mugurell:
Merge failed because of https://github.com/mozilla-mobile/fenix/issues/17734
Let's try it again in the event that that is an intermittent issue.

csadilek:
Rebased as test is fixed on master now.

PR #17657, opened 2021-01-27
For #17655: Do not pass threshold if showTopFrecentSites is false.

mcarare:
### Pull Request checklist

- [x] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not
- [x] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features. In addition, it includes a screenshot of a successful [accessibility scan](https://play.google.com/store/apps/details?id=com.google.android.apps.accessibility.auditor&hl=en_US) to ensure no new defects are added to the product.

### To download an APK when reviewing a PR:
1. click on Show All Checks,
2. click Details next to "Taskcluster (pull_request)" after it appears and then finishes with a green checkmark,
3. click on the "Fenix - assemble" task, then click "Run Artifacts".
4. the APK links should be on the left side of the screen, named for each CPU architecture

Issue #17655, opened 2021-01-27
[Bug] Most visited top sites shown even when it's turned off


hakkikaancaliskan:
[comment]: # (Please do your best to search for duplicate issues before filing a new issue so we can keep our issue board clean)
[comment]: # (Every issue should have the exact bug and steps to reproduce described in it. Please do not file feedback list tickets as it is difficult to parse them and address their individual points)
[comment]: # (Read https://github.com/mozilla-mobile/fenix#i-want-to-file-an-issue for more information)

## Steps to reproduce
Have most visited top sites turned off
Upgrade from 85-beta9 to 86-beta1

### Expected behavior
It respects to setting and remains off
### Actual behavior
It's shown even when setting turned off
### Device information

* Android device: Xperia XZ1 Android 11
* Fenix version: 86-beta1

ThracianKnight1907:
I also have this issue. It looks like that button does nothing in the current beta

hakkikaancaliskan:
There is no problem on nightly, maybe you guys missed a commit to include?

leadweedy:
This bug is also affecting nightly since today's build: Nightly 210127 17:01 (Build #2015789835)

softvision-miralobontiu:
I was not able to reproduce the issue on the 1/29 Nightly build, but I **am reproducing the issue on the 86.0.0-beta.1** build with Sony Xperia Z5 Premium (Android 7.1.1).
Removing the qa:needed label for now.

kbrosnan:
@amedyne Has this been uplifted to Beta? It would be a significant regression to ship this.

amedyne:
Fix uplifted to Beta.

PR #17812, opened 2021-02-03
For #17655: Do not pass threshold if showTopFrecentSites is false. (#17657)

st3fan:
Uplift cherry picked from commit 8993a0acb28a03ee810fff42d6357ff2929417e5

* For #17655: Do not pass threshold if showTopFrecentSites is false.
* For #17655: Add unit tests for getTopSitesConfig.


gabrielluong (approved):

gabrielluong (commented):
Actually, I think this is suppose to be uplifted to 86.

PR #17834, opened 2021-02-04
Fixes #17738 - Early initialization of Nimbus

### LINKED ISSUE #17738: Nimbus DatabaseNotReady: Nimbus setup is being done too late.
Currently `requireContext().components.analytics.experiments` lazily initializes the `Nimbus` object.

Several jobs are put on I/O threads (from #17684), including the one to open the database and read its initial state.

Until these jobs finish, calling `experiments.withExperiment(experimentSlug)` near start up is going to fail silently.

i.e. this puts all of Nimbus startup off the main thread, but means that nimbus isn't useful for features configured by nimbus near to app-startup.

Optimizing Nimbus further is not the business of this issue: Nimbus needs to be initialized, and is racing the main thread to do it. This issue is about starting the background thread earlier.

Three approaches here in order of complexity:

1. Eagerly initialize the nimbus object.
2. Keep the `by lazyMonitored` to initialize the nimbus object, and call it much earlier, e.g. calling `experiments.initialize()` [in FenixApplication][2].
3. As part of [ContentResolver pattern][1] driven startup effort, not currently seen in Fenix.

Android-Components is not pursuing any systemic componentized startup, so (3) should not be considered.

My favoured solution at this time is (2): it's a one line fix, which give us some degree of control when in the start up sequence nimbus is initialized.

[1]: https://medium.com/androiddevelopers/app-startup-part-1-34f57b65cacd
[2]: https://github.com/mozilla-mobile/fenix/blob/d48bb2a4ccb8dbb3409cca0280f21e7e0f8a756f/app/src/main/java/org/mozilla/fenix/FenixApplication.kt#L309

jhugman:
Fixes #17738.

Nimbus was being initialized, i.e. just-in-time which was racing between initialization and first use in order to avoid blocking the main thread. Initialization is never winning the race.

This PR moves initialization from just before use to just after the earliest time it can be initialized: i.e. just after the megazord has loaded.

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
+++ b/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
@@ -307,6 +307,10 @@ open class FenixApplication : LocaleAwareApplication(), Provider {
             // ... but RustHttpConfig.setClient() and RustLog.enable() can be called later.
             RustHttpConfig.setClient(lazy { components.core.client })
             RustLog.enable(components.analytics.crashReporter)
+            // We want to ensure Nimbus is initialized as early as possible so we can
+            // experiment on features close to startup.
+            // But we need viaduct (the RustHttp client) to be ready before we do.
+            components.analytics.experiments.initialize()
         }
     }
 


app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -307,6 +307,10 @@ open class FenixApplication : LocaleAwareApplication(), Provider {
             // ... but RustHttpConfig.setClient() and RustLog.enable() can be called later.
             RustHttpConfig.setClient(lazy { components.core.client })
             RustLog.enable(components.analytics.crashReporter)
+            // We want to ensure Nimbus is initialized as early as possible so we can
+            // experiment on features close to startup.
+            // But we need viaduct (the RustHttp client) to be ready before we do.
+            components.analytics.experiments.initialize()
```
csadilek, line 313: From a performance perspective this is fine, as all the `initialize` code runs on a separate thread now (even if we didn't launch it on an IO scope here). The default `lazy` thread-safety mode also makes sure our lazy init block in `Analytics` will only run once even in case this code runs concurrently with another part of the app that also accesses `analytics.experiments`.

However, I am concerned about the `Nimbus` instance itself being accessed from multiple threads, in particular about the thread visibility of the mutable class fields. Thread visibility is not guaranteed for non final fields, IIRC, so it's possible that another thread sees `Nimbus.nimbus` and/or `Nimbus.datadir` as null. Some options are making methods on `Nimbus` synchronized or the affected fields `volatile`.

csadilek (changes requested):
I found one thread-safety issue we should discuss, see comment inline.

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
csadilek: Opened a PR in A-C: https://github.com/mozilla-mobile/android-components/pull/9618 (the `datadir` was already removed.)

csadilek: A-C change landed and upgraded here. So this is good now.

csadilek (approved):
See https://github.com/mozilla-mobile/fenix/pull/17834/files#r570492699. Ready to land.

PR #17837, opened 2021-02-04
Remove SessionManager reference from FenixApplication

csadilek:
Just replacing `SessionManager` usage with use cases / store.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
+++ b/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
@@ -20,7 +20,6 @@ import kotlinx.coroutines.GlobalScope
 import kotlinx.coroutines.async
 import kotlinx.coroutines.launch
 import mozilla.appservices.Megazord
-import mozilla.components.browser.session.Session
 import mozilla.components.browser.state.action.SystemAction
 import mozilla.components.browser.state.selector.selectedTab
 import mozilla.components.concept.base.crash.Breadcrumb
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -402,9 +401,19 @@ open class FenixApplication : LocaleAwareApplication(), Provider {
...
+                                selectTab = true,
+                                engineSession = engineSession
+                            )
+                        } else {
+                            components.useCases.tabsUseCases.addTab(
+                                url = url,
+                                selectTab = true,
+                                engineSession = engineSession
+                            )
+                        }
```
Amejia481, line 416: Maybe we could have a use case for this on AC

Amejia481 (approved):
👍🏽

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
csadilek: Thought about that too but it really would only save the if statement. Might still be worth it but since we want to remove those overrides here anyway as soon as the refactor is complete let's land and revisit.

sv-ohorvath:
Please rebase, the blocking test was disabled: https://github.com/mozilla-mobile/fenix/commit/f0b7b3b987202339a5affbe76c5b857405f004a6

pocmo (approved):

pocmo:
Rebased patch and pushed it to this PR.

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
Amejia481: 👍🏽

Issue #17785, opened 2021-02-03
[Bug]Screenshot feature in private browsing mode


Japeskie:
## Steps to reproduce

EDIT: Uploaded video to be more clear.

1. Take a screenshot.
2. Minimize the application.
3. Go back to Fenix.
4. Take a screenshot.

### Expected behavior

You will be able to take a screenshot.

### Actual behavior

Fenix will say "This app doesn't allow taking screenshots".

https://user-images.githubusercontent.com/73644544/106727709-b4ae5580-6646-11eb-959d-702839ab00e4.mp4

### Device information

* Device vendor / model and Android version: Mi 9T Pro / MIUI 12
* Firefox for Android version: Nightly (Build #2015791019)

Dunexus:
You can disable this security/privacy feature in Settings > Private browsing > Allow screenshots in private browsing.

Japeskie:
What I meant is that I should be able to take screenshots using "Private Browsing" mode since I enabled it, but there's a bug when you exit the app and go back it will say "This app doesn't allow taking screenshots" even though the option "Allow screenshots in private browsing" is enabled.

Japeskie:
As of Nightly (Build #2015791947) I think this issue is fixed, I also tried it on different phones (Xiaomi, Samsung, Oppo).

Here's a video of trying to replicate the issue.

https://user-images.githubusercontent.com/73644544/107202116-c91e9380-6a34-11eb-94f4-7ffd6964dfa7.mp4

softvision-miralobontiu:
Verified as fixed on the 2/8 Nightly build with Samsung Galaxy Note 8 (Android 9), and Sony Xperia Z5 Premium (Android 7.1.1).

Issue #11153, opened 2020-06-02
[Private Browsing] Opened apps view should always hide contents of private screen


AmyYLee:
Steps to reproduce:

1. Open a new tab in private browsing
1. Go to open apps view

Expected:
Private mode contents is hidden (for private homescreen and websites)

Actual:
Private mode contents is shown. 

Rational: The idea of private browsing is that we keep your what you're looking at private except when you're actively using Private mode. Showing the contents of private browsing outside of the app (ie. open apps view) violates that.

![](https://user-images.githubusercontent.com/5992746/83549121-ca291980-a4d2-11ea-8593-9f607401892a.png)

-------------------
https://github.com/mozilla-mobile/fenix/issues/11153#issuecomment-662108177 :
> We can add the following string as disclaimer text underneath
> 
> **Allow screenshots in private browsing**
> If allowed, private tabs will also be visible when multiple apps are open

ekager:
It's a setting in PBM settings -> "allow screenshots in private browsing" . The default setting is set to true (screenshots are allowed to be taken in private mode), as described in #2768

AmyYLee:
@ekager - Screenshots can still be allowed to be taken. This issue is specifically for the open apps view where we don't hide private browsing screen (when most other browsers do). Spoke to @vesta0 and she is okay with my recommendation.

ekager:
@AmyYLee I can reopen but I am still not sure what you're talking about. When I turn this setting on, I cannot screenshot any screen of the app, and in the apps view, I see a grey blob over Fenix. Are you talking about changing the default setting for screenshots? Or is this function not working on your build even when you have toggled the setting?

ekager:
From the eng side - we can't differentiate between not allowing screenshots and hiding in the app switcher. It's just one action -> we just tell the system that this is a secure window.

AmyYLee:
Please use Ink 90 #1D1133 and it should cover the entire window including search bar. Please see reference below

![Screen Shot 2020-06-08 at 15 09 55](https://user-images.githubusercontent.com/5992746/84071038-12e14680-a99b-11ea-8b0e-8a8c08f7aceb.png)

mcarare:
@AmyYLee AFAIK there is no way to customize the preview when the `SECURE` flag is set.
As for the search bar part, I am not sure what you are referring to. I do not see any search bar in the previews.

AndiAJ:
[...]

PR #17842, opened 2021-02-04
For #17785 - Use screenshots setting when adding secure flag

ekager:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
+++ b/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
@@ -17,7 +17,7 @@ import android.view.KeyEvent
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewConfiguration
-import android.view.WindowManager
+import android.view.WindowManager.LayoutParams.FLAG_SECURE
 import androidx.annotation.CallSuper
 import androidx.annotation.IdRes
 import androidx.annotation.VisibleForTesting
@@ -278,6 +278,11 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
     override fun onResume() {
         super.onResume()
 
+        // Even if screenshots are allowed, we hide private content in the recents screen in onPause
+        // so onResume we should go back to setting these flags with the user screenshot setting
+        // See https://github.com/mozilla-mobile/fenix/issues/11153
+        updateSecureWindowFlags(settings().lastKnownMode)
+
         // Diagnostic breadcrumb for "Display already aquired" crash:
         // https://github.com/mozilla-mobile/android-components/issues/7960
         breadcrumb(
@@ -341,8 +346,10 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
         settings().shouldReturnToBrowser =
             components.core.store.state.getNormalOrPrivateTabs(private = false).isNotEmpty()
 
+        // Even if screenshots are allowed, we want to hide private content in the recents screen
+        // See https://github.com/mozilla-mobile/fenix/issues/11153
         if (settings().lastKnownMode.isPrivate) {
-            window.addFlags(WindowManager.LayoutParams.FLAG_SECURE)
+            window.addFlags(FLAG_SECURE)
         }
 
         // We will remove this when AC code lands to emit a fact on getTopSites in DefaultTopSitesStorage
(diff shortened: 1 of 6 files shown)


sv-ohorvath:
Please rebase, the blocking test was disabled: https://github.com/mozilla-mobile/fenix/commit/f0b7b3b987202339a5affbe76c5b857405f004a6

app/src/main/java/org/mozilla/fenix/tabtray/TabTrayDialogFragment.kt
```diff
@@ -86,7 +86,11 @@ class TabTrayDialogFragment : AppCompatDialogFragment(), UserInteractionHandler
             else null
 
     private val collectionStorageObserver = object : TabCollectionStorage.Observer {
-        override fun onCollectionCreated(title: String, sessions: List<TabSessionState>, id: Long?) {
+        override fun onCollectionCreated(
```
Mugurell, line 89: nit: this new formatting will skew git history and make it harder when using blame to find why some changes were introduced.

Mugurell (approved):
Tested it and saw it resolving the reported issue.
How, when, where we actually set the `FLAG_SECURE` will probably need a bit more work for which I've created https://github.com/mozilla-mobile/fenix/issues/17854 and https://github.com/mozilla-mobile/fenix/issues/17855

ekager:
@Mugurell you're right! I decided to try to address that here because what we were doing before was not very robust. 

In my mind there are 3 times we should worry about setting this flag and none of them involve the Android lifecycle: 
1. When normal/private mode changes
2. When screenshot setting changes
3. When normal/private tab in tabs tray changes

I've addressed those 3 places now and removed the old lifecycle triggers, can you think of any other times we would care about this?

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -341,10 +341,6 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
         settings().shouldReturnToBrowser =
             components.core.store.state.getNormalOrPrivateTabs(private = false).isNotEmpty()
 
-        if (settings().lastKnownMode.isPrivate) {
```
Mugurell, line 344: I think we still need this because of 
https://github.com/mozilla-mobile/fenix/issues/11153
Even if screenshots are allowed
We should do our best to not display the private content in the recents screen.

Mugurell: If we keep this a counterpart in onResume will probably be needed.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -850,7 +846,18 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
 
     protected open fun createBrowsingModeManager(initialMode: BrowsingMode): BrowsingModeManager {
         return DefaultBrowsingModeManager(initialMode, components.settings) { newMode ->
+            updateSecureWindowFlags(newMode)
```
Mugurell, line 856: 👍

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -407,12 +406,6 @@ class HomeFragment : Fragment() {
             }
         }
 
-        if (browsingModeManager.mode.isPrivate) {
```
Mugurell, line 410: Tested, changing from PWA to private tabs works nicely.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
ekager: Oops totally forgot about this behavior. Good call

app/src/main/java/org/mozilla/fenix/settings/PrivateBrowsingFragment.kt
```diff
@@ -44,7 +46,18 @@ class PrivateBrowsingFragment : PreferenceFragmentCompat() {
         }
 
         requirePreference<SwitchPreference>(R.string.pref_key_allow_screenshots_in_private_mode).apply {
-            onPreferenceChangeListener = SharedPreferenceUpdater()
+            onPreferenceChangeListener = object : SharedPreferenceUpdater() {
+                override fun onPreferenceChange(preference: Preference, newValue: Any?): Boolean {
```
Mugurell, line 50: 👍

Mugurell (approved):
Played a bit with the app, everything seems to work nicely now! 
Other than having to support that edgecase with the recents screen I don't see other issues.

ekager:
Great! I'll merge once everything passes. Thanks for filing those issues - definitely good to nail down the correct behavior while we're already looking at this.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
ekager: Added comments explaining this too to avoid regressions.

Mugurell:
@ekager I saw one more case that will probably need a special treatment to know that everything will be finished

Here, in onResume the SECURE flag is set but then it is not cleared (if not needed) in onPause.
https://github.com/mozilla-mobile/fenix/blob/07748f69b75144e8e247b742f66a756f999cb299/app/src/main/java/org/mozilla/fenix/settings/logins/fragment/SavedLoginsFragment.kt#L60-L63

PR #17864, opened 2021-02-05
Issue #12731: Collect telemetry about the content process of tabs getting killed.

pocmo:
AC issue:
https://github.com/mozilla-mobile/android-components/issues/9366

Required change on the AC side:
https://github.com/mozilla-mobile/android-components/pull/9634

This still needs a data-review. 

r? @csadilek @dblohm7


app/src/main/java/org/mozilla/fenix/components/metrics/EngineTelemetryMiddleware.kt
```diff
@@ -0,0 +1,67 @@
...
+        if (isSelected && ageNanos != null) {
+            EngineMetrics.killForegroundAge.setRawNanos(ageNanos)
+        } else if (ageNanos != null) {
+            EngineMetrics.killBackgroundAge.setRawNanos(ageNanos)
+        }
+    }
+}
+
+@Suppress("MagicNumber")
+private fun EngineState.ageNanos(): Long? {
```
csadilek, line 63: Wait this confused me. This is returning ageMillis, right?

csadilek (commented):
Looks great, one comment below, and should we combine this with the existing `TelemetryMiddleware` or prefer keeping it separate?

app/src/main/java/org/mozilla/fenix/components/metrics/EngineTelemetryMiddleware.kt
csadilek: Oh I see, `elapsedRealtime` actually returns millis, but you want to record nanos. I guess `nowNanos` should be `nowMillis` and `timeStamp` is millis too?

pocmo: Yeah, you are right, I changed this part of the code a couple of times and now the names do not make any sense. :)

pocmo:
> Looks great, one comment below, and should we combine this with the existing `TelemetryMiddleware` or prefer keeping it separate?

Yeah, I saw `TelemetryMiddleware` when I wrote this and thought about it, but it felt like I'd be mixing two independent things and turning `TelemetryMiddleware` into something that's hard to digest. wdyt?

pocmo:

- How often does the content process of foreground/background tabs get killed?
- How "old" was the foreground/background session when it got killed.

See issue: https://github.com/mozilla-mobile/android-components/issues/9366

For debugging purposes. It seems like a large amount of users see unexpected "reloads" of websites they are on. This seems to be caused by content process kills by the Android system. With those probes we want to see how big this problem is across our user base and also see whether mitigation strategies improve the situation.

See:
* https://bugzilla.mozilla.org/show_bug.cgi?id=1682319
* https://github.com/mozilla-mobile/fenix/issues/12731

* We tried manually debugging the situation. However this didn't give us a a good picture on how often/likely those content process kills happen with normal usage on a variety of devices.

* No.

* _Category 1 “Technical data”_
* Maybe it is possible to interpolate a rough estimate about how many (crashed) tabs a user had open at some point (but no information about those tabs)? Would that be _Category 2 “Interaction data”_?

* I picked end of the year (2021-12-31) for now. We want those probes to ride the trains to release and also see the effect of bugfixes in this area in future releases.

* Glean / Amplitude

* Only on Glean, Amplitude and with mobile / GeckoView teams.

csadilek:
> Yeah, I saw TelemetryMiddleware when I wrote this and thought about it, but it felt like I'd be mixing two independent things and turning TelemetryMiddleware into something that's hard to digest. wdyt?

@pocmo Makes sense to keep separate, but we should probably nest them into a `TelemeteryMiddleware.create`, similar to `EngineMiddleware.create`, otherwise it will be confusing that we have two?

app/src/main/java/org/mozilla/fenix/components/metrics/EngineTelemetryMiddleware.kt
csadilek: Ah OK, let's update to `nowMillis` and `timeStampMillis` then? :)

pocmo:
Updated the data collection from.

And updated the PR. Looking at it again your proposal to integrate the changes into `TelemetryMiddleware` seemed fine and that's what it did now.

app/src/test/java/org/mozilla/fenix/TelemetryMiddlewareTest.kt
```diff
@@ -244,4 +268,125 @@ class TelemetryMiddlewareTest {
...
+        store.dispatch(TabListAction.RestoreAction(
+            listOf(
+                createTab("https://www.mozilla.org", id = "foreground"),
+                createTab("https://getpocket.com", id = "background_pocket"),
+                createTab("https://theverge.com", id = "background_verge")
+            ),
+            selectedTabId = "foreground"
+        )).joinBlocking()
+
+        Assert.assertFalse(EngineMetrics.tabKills["foreground"].testHasValue())
```
csadilek, line 283: nit: static import `assertFalse` and `assertTrue`.

csadilek (approved):
👍

boek:
# Data Review Form (to be filled by Data Stewards)

Yes, see metrics.yaml and metrics.md

Yes, data collection settings

Has expiry

4) Using the **[category system of data types](https://wiki.mozilla.org/Firefox/Data_Collection)** on the Mozilla wiki, what collection type of data do the requested measurements fall under?
Type 1

Default on

6) Does the instrumentation include the addition of **any *new* identifiers** (whether anonymous or otherwise; e.g., username, random IDs, etc.  See the appendix for more details)?
No

7) Is the data collection covered by the existing Firefox privacy notice? **If unsure: escalate to legal if:**
Yes

8) Does there need to be a check-in in the future to determine whether to renew the data?
Fenix team will check-in at expiry

9) Does the data collection use a third-party collection tool?
No

dblohm7:
Thank you @pocmo! 🎉

Issue #17855, opened 2021-02-05
[Bug] Private Mode screenshots possible via Tab Tray Overviews (2)


Mugurell:
Similar to https://github.com/mozilla-mobile/fenix/issues/17854 and https://github.com/mozilla-mobile/fenix/issues/12453 one can unknowingly allow the list of opened private tabs screenshotted / recorded.

## Steps to reproduce
- Have Settings -> Private browsing -> Allow screenshots in private mode disabled
- have a few normal tabs and a few private tabs opened
- open one normal tab / got to normal tabs homescreen
- open tabs tray -> choose to create a new private tab
- try taking screenshots of all the tabs shown in tabs tray, incl private ones.

### Expected behavior
A message from Android saying that screenshots are not allowed

### Actual behavior
Screenshots are possible

### To note
When putting the app in background and resuming the FLAG_SECURE gets set (not all the times) so now one cannot take record the screen while in private mode.

Video showing these cases (use the android message / screenshot as indicators of when screenshots are blocked or not):

https://user-images.githubusercontent.com/11428869/107045427-2e248000-67ce-11eb-81dc-17cf98480b44.mov

softvision-miralobontiu:
This is not fixed entirely, I was able to take a screenshot in private browsing when the keyboard is trigerred, as it can be seen in the video below:

https://user-images.githubusercontent.com/40694158/107219007-20166f80-6a19-11eb-83b1-8220f649387c.mp4

Environment:
* 2/8 Nightly build
* Samusng Galaxy Note 8 (Android 9)

softvision-miralobontiu:
Reproducible on Beta 86.0.0-beta.5, and 2/15 Nightly with Sony Xperia Z5 Premium (Android 7.1.1).

softvision-miralobontiu:
Reproducible on RC 86.1.0 with Sony Xperia Z5 Premium (Android 7.1.1).

lobontiumira:
This issue still reproduces on the 3/22 Nightly build with Samsung Galaxy Note 8 (Android 9).
I was able to take a screenshot in private browsing when the keyboard is trigerred.

lobontiumira:
Reproducible on Beta 88.0.0-beta.6 with OnePlus 5T (Android 10).

lobontiumira:
Reproducible on RC 88.1.0 with Xiaomi Mi 8 Lite (Android 9).

lobontiumira:
[...]

PR #17918, opened 2021-02-08
For #17855 - Ensure Secure flags are set on Search Dialog

ekager:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/search/SearchDialogFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/search/SearchDialogFragment.kt
@@ -63,6 +63,7 @@ import org.mozilla.fenix.components.metrics.Event
 import org.mozilla.fenix.components.toolbar.ToolbarPosition
 import org.mozilla.fenix.ext.components
 import org.mozilla.fenix.ext.requireComponents
+import org.mozilla.fenix.ext.secure
 import org.mozilla.fenix.ext.settings
 import org.mozilla.fenix.search.awesomebar.AwesomeBarView
 import org.mozilla.fenix.search.toolbar.ToolbarView
@@ -433,7 +434,7 @@ class SearchDialogFragment : AppCompatDialogFragment(), UserInteractionHandler {
                             dialog.dismiss()
                         }
                         create()
-                    }.show()
+                    }.show().secure(activity)
                 }
             }
         )


Mugurell (approved):
Nice!

Mugurell (changes requested):
Looks like I can still take screenshots when in SearchDialogFragment.
Testing a bit more now.

Mugurell:
@ekager still being able to take screenshots when coming from a private page to the private homescreen and SearchDialogFragment.
Then because the activity does not have `FLAG_SECURE` the `secure()` won't actually do anything.

Because a video is worth >1k words:
(latest master with this patch applied) 

https://user-images.githubusercontent.com/11428869/107403726-24708480-6b0e-11eb-8729-db483bd166bb.mov

Mugurell:
Looked into this a bit.
What happens is that upon pressing the tab tray button to create a new tab (in this case "+ PRIVATE")
- TabTrayController#handleNewTabTapped is called which [will set the secure flag as expected](https://github.com/mozilla-mobile/fenix/blob/f7b7b25213001e8f766a295a65292722a6cc8946/app/src/main/java/org/mozilla/fenix/tabtray/TabTrayController.kt#L109)
- HomeFragment - onResume
- BrowserFragment - onStop
  This will trigger `SecureWindowFeature#stop` which [will purposely clear the secure flag](https://github.com/mozilla-mobile/android-components/blob/b266ac3d16e15cabc969cfe33c9cf965f557e257/components/feature/privatemode/src/main/java/mozilla/components/feature/privatemode/feature/SecureWindowFeature.kt#L55)
  
 So from now on we are not secure.

ekager:
Hm after our changes in https://github.com/mozilla-mobile/fenix/pull/17842 I'm wondering if we can remove `SecureWindowFeature` entirely... Shouldn't the `HomeActivity` be taking care of the secure flag of the window for private/normal tabs now on `mode` change?

Mugurell:
Looking at the ticket that implemented that `SecureWindowFeature` - https://github.com/mozilla-mobile/android-components/issues/3707 (and the then Fenix usecases) it seems indeed that functionality is superseded by the current Fenix code.
That feature would've automatically set/remove the secure flag for opened tabs but Fenix wants to have more app components secured and does this "manually".
I don't think there are any situations in which a private tab could be opened without the browsingModeManager having a chance to set the expected flags so just removing `SecureWindowFeature` from Fenix would probably be better.

And if we do this I think we won't need to secure the qr dialog .. or any other dialog ..?

Issue #17352, opened 2021-01-07
[Bug] Deleted downloads reappear in the Downloads menu


sv-ohorvath:
[comment]: # (Please do your best to search for duplicate issues before filing a new issue so we can keep our issue board clean)
[comment]: # (Every issue should have the exact bug and steps to reproduce described in it. Please do not file feedback list tickets as it is difficult to parse them and address their individual points)
[comment]: # (Read https://github.com/mozilla-mobile/fenix#i-want-to-file-an-issue for more information)

## Steps to reproduce
1. Save a file (you can use file-examples.com)
2. Go to the Downloads menu, delete the file.
3. Go to device storage, to the Downloads folder and delete the file from storage too.
4. Return to the app, go back to the download page, download the file again.
5. Go to the Downloads menu and check the files.
### Expected behavior
The file is shown as downloaded just once.
### Actual behavior
The file appears twice in the Downloads menu.
### Device information

* Android device: Pixel 2 (Android 9)
* Fenix version: Debug 1/7
 
## Scenario 2:
1. Save a file (you can use file-examples.com)
2. Go to the Downloads menu, delete the file. Leave the file in device storage.
3. Return to the app, go back to the download page, download a different file.
4. Go to the Downloads menu and check the files.
### Actual behavior
The deleted (first) file reappears in the Downloads menu.

## Scenario 3:
1. Download a file
2. Go to the Downloads menu, delete the file.
3. Exit the menu before the UNDO snackbar disappears.
4. Go back to the Downloads menu, check the files.
### Actual behavior
The file reappears in the Downloads menu.

softvision-miralobontiu:
Reproducible on Beta 85.0.0-beta.8 with Sony Xperia z5 Premium (Android 7.1.1).

softvision-miralobontiu:
Verified as fixed on the 2/15 Nightly build, and on Beta 86.0.0-beta.5 build with Sony Xperia Z5 Premium (Android 7.1.1).

PR #17930, opened 2021-02-09
For #17352 - Fixes deleted downloads reappearing

codrut-topliceanu:
The `getDeleteDownloadItemsOperation` job would check fragment context not null after the fragment was stopped. Removing that check and only passing the downloadUseCase as a parameter fixes the problem.

For https://github.com/mozilla-mobile/fenix/issues/17352
### Pull Request checklist

- [x] **Tests**: This PR includes no tests as it is only a minor change.
- [x] **Screenshots**: This PR doesn't include any screenshots as there is no immediate visual change

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/library/downloads/DownloadFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/library/downloads/DownloadFragment.kt
@@ -282,7 +282,7 @@ class DownloadFragment : LibraryPageFragment<DownloadItem>(), UserInteractionHan
      * Queues the [getDeleteDownloadItemsOperation] job in [pendingDownloadDeletionJob] in case
      *  the user exits the fragment and we need to quickly execute the queued deletion.
      * And adds the [items] to be deleted to the list of [DownloadFragmentStore.pendingDeletionIds],
-     *  which is used to determine what items to show and what items to hide from the user.
+     * which is used to determine what items to show and what items to hide from the user.
      */
     private fun updatePendingDownloadToDelete(items: Set<DownloadItem>) {
         pendingDownloadDeletionJob = getDeleteDownloadItemsOperation(downloadsUseCases, items)
--- a/app/src/main/java/org/mozilla/fenix/library/downloads/DownloadFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/library/downloads/DownloadFragment.kt
@@ -25,6 +25,7 @@ import mozilla.components.browser.state.state.BrowserState
 import kotlinx.coroutines.launch
 import mozilla.components.browser.state.state.content.DownloadState
 import mozilla.components.feature.downloads.AbstractFetchDownloadService
+import mozilla.components.feature.downloads.DownloadsUseCases
 import mozilla.components.lib.state.ext.consumeFrom
 import mozilla.components.support.base.feature.UserInteractionHandler
 import org.mozilla.fenix.HomeActivity
@@ -50,6 +51,7 @@ class DownloadFragment : LibraryPageFragment<DownloadItem>(), UserInteractionHan
     private lateinit var metrics: MetricController
     private var undoScope: CoroutineScope? = null
     private var pendingDownloadDeletionJob: (suspend () -> Unit)? = null
+    private lateinit var downloadsUseCases: DownloadsUseCases
 
     override fun onCreateView(
         inflater: LayoutInflater,
@@ -59,6 +61,7 @@ class DownloadFragment : LibraryPageFragment<DownloadItem>(), UserInteractionHan
         val view = inflater.inflate(R.layout.fragment_downloads, container, false)
 
         val items = provideDownloads(requireComponents.core.store.state)
+        downloadsUseCases = requireContext().components.useCases.downloadUseCases
 
         downloadStore = StoreProvider.get(this) {
             DownloadFragmentStore(
@@ -85,6 +88,10 @@ class DownloadFragment : LibraryPageFragment<DownloadItem>(), UserInteractionHan
         return view
     }
 
+    /**
+     * Returns a list of available downloads to be displayed to the user.
+     *  Downloads must be COMPLETED and existent on disk.
+     * */
     @VisibleForTesting
     internal fun provideDownloads(state: BrowserState): List<DownloadItem> {
         return state.downloads.values
@@ -128,9 +135,7 @@ class DownloadFragment : LibraryPageFragment<DownloadItem>(), UserInteractionHan
                 setPositiveButton(R.string.delete_browsing_data_prompt_allow) { dialog: DialogInterface, _ ->
                     // Use fragment's lifecycle; the view may be gone by the time dialog is interacted with.
                     lifecycleScope.launch(IO) {
-                        context.let {
-                            it.components.useCases.downloadUseCases.removeAllDownloads()
-                        }
+                            downloadsUseCases.removeAllDownloads()
                         updatePendingDownloadToDelete(downloadStore.state.items.toSet())
                         launch(Dispatchers.Main) {
                             showSnackBar(
@@ -146,6 +151,11 @@ class DownloadFragment : LibraryPageFragment<DownloadItem>(), UserInteractionHan
         }
     }
 
+    /**
+     * Schedules [items] for deletion.
+     * Note: When tapping on a download item's "trash" button
+     *  (itemView.overflow_menu) this [items].size() will be 1.
+     * */
     private fun deleteDownloadItems(items: Set<DownloadItem>) {
         metrics.track(Event.DownloadsItemDeleted)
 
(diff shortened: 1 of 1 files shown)


Mugurell:
This looks good to me and in tests I saw the issue being fixed.
I would've first thought that we could use a more complex implementation with soft deleting the data first and then actually cleaning it up later so that we could stay in the bounds of the coroutine scopes used here and ensure total consistency with the user actions.

Following the previous behavior a little:
- the previous implementation depended heavily (and will still) on `invokePendingDeletion` called in Fragment's onPause / onBack 
- method which will launch a new coroutine to actually delete all pending files (delete operations from `undoScope?.allowUndo` which haven't been invoked yet)
- the new coroutine would be in `getDeleteDownloadItemsOperation` but this process would come to a halt because if the user is pressing back from the screen the Fragment will be destroyed and so [this Fragment context](https://github.com/mozilla-mobile/fenix/blob/574eac463683292260e13311c6ed8348503cbbaa/app/src/main/java/org/mozilla/fenix/library/downloads/DownloadFragment.kt#L253) will be null and we'll not actually get to call `removeDownload`.

Seems like we could just use the fact that `getDeleteDownloadItemsOperation` executes in a new coroutine scope so this won't be cancelled when the `lifecycleScope` is cancelled (when the Fragment is destroyed).
We would just need to leak the Context a bit more or have the `DownloadsUseCases` still available to the coroutine from `getDeleteDownloadItemsOperation`.

In the end this is not exactly bulletproof - users could still delete some downloads by pressing the trashcan icon and then immediately close the process - remove the app from recents (so that the coroutines trying to cleanup here would be cancelled) - but it would work in 99.(9)% of the cases with relatively small changes.

Amejia481 (approved):
LGTM!
Is it possible to add a tests to verify the previous issue?

codrut-topliceanu:
@Amejia481 Awesome. Thanks 🥇 
I'll look into adding some sort of tests.

Amejia481:
👍🏽

app/src/main/java/org/mozilla/fenix/library/downloads/DownloadFragment.kt
```diff
@@ -246,23 +259,33 @@ class DownloadFragment : LibraryPageFragment<DownloadItem>(), UserInteractionHan
...
                 }
                 downloadStore.dispatch(DownloadFragmentAction.ExitDeletionMode)
                 pendingDownloadDeletionJob = null
             }
         }
     }
 
+    /**
+     * Queues the [getDeleteDownloadItemsOperation] job in [pendingDownloadDeletionJob] in case
+     *  the user exits the fragment and we need to quickly execute the queued deletion.
```
gabrielluong, line 283:
```suggestion
     * the user exits the fragment and we need to quickly execute the queued deletion.
```

app/src/main/java/org/mozilla/fenix/library/downloads/DownloadFragment.kt
```diff
@@ -246,23 +259,33 @@ class DownloadFragment : LibraryPageFragment<DownloadItem>(), UserInteractionHan
...
+     * And adds the [items] to be deleted to the list of [DownloadFragmentStore.pendingDeletionIds],
+     *  which is used to determine what items to show and what items to hide from the user.
```
gabrielluong, line 285:
```suggestion
     * which is used to determine what items to show and what items to hide from the user.
```

app/src/main/java/org/mozilla/fenix/library/downloads/DownloadFragment.kt
```diff
@@ -85,6 +88,10 @@ class DownloadFragment : LibraryPageFragment<DownloadItem>(), UserInteractionHan
         return view
     }
 
+    /**
+     * Returns a list of available downloads to be displayed to the user.
+     *  Downloads must be COMPLETED and existent on disk.
```
gabrielluong, line 93:
```suggestion
     * Downloads must be COMPLETED and existent on disk.
```

app/src/main/java/org/mozilla/fenix/library/downloads/DownloadFragment.kt
```diff
@@ -146,6 +151,11 @@ class DownloadFragment : LibraryPageFragment<DownloadItem>(), UserInteractionHan
         }
     }
 
+    /**
+     * Schedules [items] for deletion.
+     * Note: When tapping on a download item's "trash" button
+     *  (itemView.overflow_menu) this [items].size() will be 1.
```
gabrielluong, line 157:
```suggestion
     * (itemView.overflow_menu) this [items].size() will be 1.
```

app/src/main/java/org/mozilla/fenix/library/downloads/DownloadFragment.kt
```diff
@@ -246,23 +259,33 @@ class DownloadFragment : LibraryPageFragment<DownloadItem>(), UserInteractionHan
         metrics.track(Event.DownloadsItemOpened)
     }
 
-    private fun getDeleteDownloadItemsOperation(items: Set<DownloadItem>): (suspend () -> Unit) {
+    /**
+     * Launches the coroutine to delete the provided [items].
+     * */
```
gabrielluong, line 264:
```suggestion
     */
```

app/src/main/java/org/mozilla/fenix/library/downloads/DownloadFragment.kt
```diff
@@ -85,6 +88,10 @@ class DownloadFragment : LibraryPageFragment<DownloadItem>(), UserInteractionHan
         return view
     }
 
+    /**
+     * Returns a list of available downloads to be displayed to the user.
+     *  Downloads must be COMPLETED and existent on disk.
+     * */
```
gabrielluong, line 94:
```suggestion
     */
```

app/src/main/java/org/mozilla/fenix/library/downloads/DownloadFragment.kt
```diff
@@ -146,6 +151,11 @@ class DownloadFragment : LibraryPageFragment<DownloadItem>(), UserInteractionHan
         }
     }
 
+    /**
+     * Schedules [items] for deletion.
+     * Note: When tapping on a download item's "trash" button
+     *  (itemView.overflow_menu) this [items].size() will be 1.
+     * */
```
gabrielluong, line 158:
```suggestion
     */
```

app/src/main/java/org/mozilla/fenix/library/downloads/DownloadFragment.kt
```diff
@@ -210,6 +220,9 @@ class DownloadFragment : LibraryPageFragment<DownloadItem>(), UserInteractionHan
         else -> super.onOptionsItemSelected(item)
     }
 
+    /**
+     * Provides a message to the Undo snackbar.
+     * */
```
gabrielluong, line 225:
```suggestion
     */
```

app/src/main/java/org/mozilla/fenix/library/downloads/DownloadFragment.kt
```diff
@@ -246,23 +259,33 @@ class DownloadFragment : LibraryPageFragment<DownloadItem>(), UserInteractionHan
...
             }
         }
     }
 
+    /**
+     * Queues the [getDeleteDownloadItemsOperation] job in [pendingDownloadDeletionJob] in case
+     *  the user exits the fragment and we need to quickly execute the queued deletion.
+     * And adds the [items] to be deleted to the list of [DownloadFragmentStore.pendingDeletionIds],
+     *  which is used to determine what items to show and what items to hide from the user.
+     * */
```
gabrielluong, line 286:
```suggestion
     */
```

app/src/main/java/org/mozilla/fenix/library/downloads/DownloadFragment.kt
```diff
@@ -273,6 +296,9 @@ class DownloadFragment : LibraryPageFragment<DownloadItem>(), UserInteractionHan
         downloadStore.dispatch(DownloadFragmentAction.UndoPendingDeletionSet(ids))
     }
 
+    /**
+     * Executes pending job(s) when leaving [DownloadFragment].
+     * */
```
gabrielluong, line 301:
```suggestion
     */
```

app/src/main/java/org/mozilla/fenix/library/downloads/DownloadFragment.kt
```diff
@@ -128,9 +135,7 @@ class DownloadFragment : LibraryPageFragment<DownloadItem>(), UserInteractionHan
                 setPositiveButton(R.string.delete_browsing_data_prompt_allow) { dialog: DialogInterface, _ ->
                     // Use fragment's lifecycle; the view may be gone by the time dialog is interacted with.
                     lifecycleScope.launch(IO) {
-                        context.let {
-                            it.components.useCases.downloadUseCases.removeAllDownloads()
-                        }
+                            downloadsUseCases.removeAllDownloads()
```
gabrielluong, line 138: nit: think we over-indented here.

Issue #17819, opened 2021-02-04
Stop depending on JCenter


pocmo:
See Android Components "meta" issue:
https://github.com/mozilla-mobile/android-components/issues/9603

----

- [x] Add Maven Central as repository. It looks like we never did this for Fenix. We also need to setup the dependency caching for that.
- [x] Use Leanplum repository: repo.leanplum.com
- [x] trove4j (dependency of Android gradle plugin)
   - https://youtrack.jetbrains.com/issue/IDEA-261387
   - https://issuetracker.google.com/issues/179291081
- [x] kotlinx-html-jvm, kotlinx-html-common (dependency of detekt)
   - ```https://github.com/Kotlin/kotlinx.html/issues/173```
   - ```https://github.com/detekt/detekt/issues/3461```
- [x] fastlane screengrab (+ falcon)
   - ```https://github.com/fastlane/fastlane/issues/18174```
   - ```https://github.com/jraska/Falcon/issues/52```

PR #17932, opened 2021-02-09
Issue #17819: Remove JCenter from buildscript repositories.

pocmo:
Similar to https://github.com/mozilla-mobile/android-components/pull/9663, but here we do not limit the usage for project repositories yet. First we need to introduce Maven Central as a repository here.


pocmo:
Folded change into #17934.

PR #17934, opened 2021-02-09
Issue #17819: Add Maven Central repository and limit usage of JCenter

pocmo:
This adds Maven Central as a repository and adds it to our dependency caching mechanism. I'm not 100% sure if this is everything that is needed, but let's see what tasklcuster thinks. :)


escapewindow (approved):
I'm not familiar enough to catch anything automation hasn't.
These changes look ok...

Issue #17195, opened 2020-12-23
Intermittent UI test failure - TabbedBrowsingTest.verifyTabTrayNotShowingStateHalfExpanded


sv-ohorvath:
### Firebase Test Run:
https://console.firebase.google.com/project/moz-fenix/testlab/histories/bh.66b7091e15d53d45/matrices/6532404695638087090/executions/bs.2f606de8ea3dcaaf/testcases/3/test-cases
### Stacktrace:
`androidx.test.espresso.NoMatchingViewException: No views in hierarchy found matching: with id: org.mozilla.fenix.debug:id/tab_wrapper`
### Build: 
Master 12/22

sv-ohorvath:
Flaky on 1/26: https://console.firebase.google.com/project/moz-fenix/testlab/histories/bh.66b7091e15d53d45/matrices/5779917191947664972/executions/bs.b8328ff8c334e283/testcases/3/test-cases

Different error: `androidx.test.espresso.NoActivityResumedException: No activities in stage RESUMED. Did you forget to launch the activity. (test.getActivity() or similar)?
at dalvik.system.VMStack.getThreadStackTrace(Native Method)
at java.lang.Thread.getStackTrace(Thread.java:1538)
at androidx.test.espresso.base.DefaultFailureHandler.getUserFriendlyError(DefaultFailureHandler.java:16)
at androidx.test.espresso.base.DefaultFailureHandler.handle(DefaultFailureHandler.java:36)
at androidx.test.espresso.ViewInteraction.waitForAndHandleInteractionResults(ViewInteraction.java:106)
at androidx.test.espresso.ViewInteraction.desugaredPerform(ViewInteraction.java:43)
at androidx.test.espresso.ViewInteraction.perform(ViewInteraction.java:94)
at org.mozilla.fenix.helpers.ViewInteractionKt.click(ViewInteraction.kt:11)
at org.mozilla.fenix.ui.robots.SettingsRobot$Transition.goBack(SettingsRobot.kt:98)
at org.mozilla.fenix.ui.robots.HomeScreenRobot$Transition.dismissOnboarding(HomeScreenRobot.kt:326)
at org.mozilla.fenix.ui.TabbedBrowsingTest.verifyTabTrayNotShowingStateHalfExpanded(TabbedBrowsingTest.kt:260)`

mcarare:
[...]

PR #17961, opened 2021-02-11
For #17195 - Don't try setting an idle resource if the View is already off screen

Mugurell:
Speculative fix based on the Firebase logs which shows that on a second check
of R.id.tab_wrapper it is not found the screen.
It may be because after `advanceToHalfExpandedState` and before getting a
reference to it in `waitForTabTrayBehaviorToIdle` it was already animated off
the screen.
With this in mind I've added a null check for the view reference before trying
to register an idling resource on it's Behavior.

Also added and used a way to click at a specific location in a View, not just
in the default middle in the View.
It was observed from the Firebase videos that a "click" on the topBar actually
selected the private tabs section. This would leave us to believe that the
"click" was caught by that other View which was placed above the x,y middle of
the topBar.

In a previous test the `verifyTabTrayNotShowingStateHalfExpanded` ran 
successfully 50 times with these changes.
https://github.com/mozilla-mobile/fenix/pull/17961/checks?check_run_id=1878656103

### DIFF
--- a/app/src/androidTest/java/org/mozilla/fenix/helpers/ext/ViewInteraction.kt
+++ b/app/src/androidTest/java/org/mozilla/fenix/helpers/ext/ViewInteraction.kt
@@ -4,7 +4,14 @@
 
 package org.mozilla.fenix.helpers
 
+import android.view.InputDevice
+import android.view.MotionEvent
+import androidx.test.espresso.ViewAction
 import androidx.test.espresso.ViewInteraction
+import androidx.test.espresso.action.GeneralClickAction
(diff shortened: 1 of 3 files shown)


mcarare (approved):
LGTM!
I would look into making the click call into an extension function with a parameter.

Mugurell:
@AndiAJ recommended also asking a review from @isabelrios / @AaronMT .

isabelrios (approved):
LGTM, let's go ahead and give this solution a try! Thanks!

PR #18025, opened 2021-02-17
For 17868: Removes listener to avoid leak

CharlesAdam:
## Tests:
Given this PR only adds a line to the onPause method I felt there was no need to add tests.

## Screenshots:
There are no screenshots due to the issue not being a visible feature neither a visible bug, the only feedback we get is from the LeakCanary

## Accessibility
There were no changes in the screen layout. I'm trying to change the language of the scanner because my results were in other language
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/settings/logins/fragment/SavedLoginsFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/logins/fragment/SavedLoginsFragment.kt
@@ -150,6 +150,8 @@ class SavedLoginsFragment : Fragment() {
         toolbarChildContainer.visibility = View.GONE
         (activity as HomeActivity).getSupportActionBarAndInflateIfNecessary().setDisplayShowTitleEnabled(true)
         sortingStrategyMenu.menuController.dismiss()
+        sortLoginsMenuRoot.setOnClickListener(null)
+        setHasOptionsMenu(false)
 
         redirectToReAuth(listOf(R.id.loginDetailFragment), findNavController().currentDestination?.id)
         super.onPause()


CharlesAdam:
I've noticed this same fragment is causing another leak, if you go to android home after entering the Saved Logins and return to the app you should receive this leak:
image

CharlesAdam:
> I've noticed this same fragment is causing another leak, if you go to android home after entering the Saved Logins and return to the app you should receive this leak:

Should this be described in other issue or solved in this PR? It seems this was happening before this changes.

CharlesAdam:
Also I feel removing the listener is not the correct way to do it, but I cannot think of any other solution

mcarare:
@CharlesAdam yes, nullifying the listener is a good approach.

I cannot reproduce the second leak. What Android version are you using?

CharlesAdam:
> @CharlesAdam yes, nullifying the listener is a good approach.
> 
> I cannot reproduce the second leak. What Android version are you using?

I'm using Android 11

mcarare:
@CharlesAdam can you reliably reproduce the second leak? Do you have a STR?

CharlesAdam:
@mcarare yes I can reproduce the leak, what do you mean with STR?

Here is a gif of the leak
[

mcarare:
@CharlesAdam STR means _Steps To Reproduce_. Thank you for the video. As I cannot reproduce the leak myself, I would ask you to try adding `setHasOptionsMenu(false)` in `onPause` and test if it still leaks.

CharlesAdam:
@mcarare Thanks for the tip, that solved it. however, should this be added to this PR or should we create another issue?

mcarare:
You can keep it here, as the original issue was referring to leaks in `SavedLoginsFragment`, not a specific one.

mcarare:
You should squash your commits, and also rebase because a fix for some intermittent test failures was merged recently.
Then you should mark your PR as ready for review.

CharlesAdam:
@mcarare Done

mcarare:
You still have 2 commits. At one point the PR was fine, that is when I opened a PR to run CI for yours.

PR #18056, opened 2021-02-18
Replace Session[Manager] with store in BaseBrowserFragment

csadilek:
Refactors BaseBrowserFragment to fully rely on the browser store instead of SessionManager.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
@@ -318,17 +318,17 @@ abstract class BaseBrowserFragment : Fragment(), UserInteractionHandler, Activit
             }
         )
         val browserToolbarMenuController = DefaultBrowserToolbarMenuController(
+            store = store,
             activity = activity,
             navController = findNavController(),
             metrics = requireComponents.analytics.metrics,
             settings = context.settings(),
             readerModeController = readerMenuController,
-            sessionManager = requireComponents.core.sessionManager,
             sessionFeature = sessionFeature,
             findInPageLauncher = { findInPageIntegration.withFeature { it.launch() } },
             swipeRefresh = swipeRefresh,
             browserAnimator = browserAnimator,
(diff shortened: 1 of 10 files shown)


app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
```diff
@@ -590,14 +591,14 @@ abstract class BaseBrowserFragment : Fragment(), UserInteractionHandler, Activit
...
-                val parentSession = sessionManager.findSessionById(tabId)
+                val parentSession = store.state.findTabOrCustomTab(tabId)
                 val useCase = if (request.isPrivate) {
                     requireComponents.useCases.searchUseCases.newPrivateTabSearch
                 } else {
                     requireComponents.useCases.searchUseCases.newTabSearch
                 }
 
-                if (parentSession?.isCustomTabSession() == true) {
+                if (parentSession is CustomTabSessionState) {
```
csadilek, line 601: This part is strange. Why did we do this? 🤔  How can the parent be a custom tab?

csadilek: Ah I see this is just naming, parent here is just what triggered the search so this should be fine.

app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
```diff
@@ -489,15 +490,15 @@ abstract class BaseBrowserFragment : Fragment(), UserInteractionHandler, Activit
...
+                if (downloadState.sessionId == store.state.selectedTabId) {
                     dynamicDownloadDialog.show()
                     browserToolbarView.expand()
                 }
             }
         }
 
         resumeDownloadDialogState(
-            sessionManager.selectedSession?.id,
+            store.state.selectedTabId,
```
csadilek, line 501: This is actually a bug here which is fixed by: https://github.com/mozilla-mobile/fenix/pull/18057

pocmo (approved):

PR #18082, opened 2021-02-19
Only run the build-contributor-pr workflow on forks

sarentz:
This patch has a better check to see if the PR is running on a pull request that came from a fork. (The current workflow does not trigger at all.)

`if: github.event.pull_request.head.repo.full_name != github.repository`

The `github.repository` is always the original repo, in our case `mozilla-mobile/fenix` while `github.event.pull_request.head.repo` points to the repository where the PR is originating from, which for contributor PRs can never be `mozilla-mobile/*`.

Let's hope this works 🤞


st3fan (approved):
This looks great, let's give it a try!

Issue #18040, opened 2021-02-18
Create smoke test to prevent start up crashes when device doesn't have network access


mcomella:
This is very similar to https://github.com/mozilla-mobile/fenix/issues/18041 The distinction is that **this bug is for no network access;** that bug is for network access but bad results.

We experienced a start up crash when the device was without network access: https://github.com/mozilla-mobile/fenix/issues/16886 afaict, this was fixed in the nimbus SDK https://github.com/mozilla/nimbus-sdk/pull/67/ but there wasn't any generic smoke test put in place to prevent this class of errors (crashes due to having no network access for start up or other UI flows). We should add such a test. This test could be automated or added to the manual QA test plan (then again if it was done manually, it'd still cause our perftests too fail (below) so maybe it's only sufficient if, when we see such crashes, that the perf team is notified directly).

To be explicit, **this is not something the performance team intends to work on** in the short term but **it's important to us** because crashes like these cause our performance tests in mozilla-central to fail, which disrupts our engineers' work stream as we investigate such crashes and possibly missing the cause of regressions as we lose performance test data.

Issue #16886, opened 2020-12-06
[Bug] [org.mozilla.experiments.nimbus.ErrorException$RequestError]: Crash when opening the home screen with no active internet connection


miDeb:
[comment]: # (Please do your best to search for duplicate issues before filing a new issue so we can keep our issue board clean)
[comment]: # (Every issue should have the exact bug and steps to reproduce described in it. Please do not file feedback list tickets as it is difficult to parse them and address their individual points)
[comment]: # (Read https://github.com/mozilla-mobile/fenix#i-want-to-file-an-issue for more information)

https://crash-stats.mozilla.org/signature/?product=Fenix&signature=org.mozilla.experiments.nimbus.ErrorException%24RequestError%3A%20at%20org.mozilla.experiments.nimbus.Error.intoException%28nimbus.kt%3A9%29

## Steps to reproduce
0. Have an open tab.
1. Disable mobile data and wifi.
2. Open the home screen. Happens both when switching from a web page to the home screen and cold starting the app.

### Expected behavior
No crash

### Actual behavior
Crash

### Device information

* Android device: samsung galaxy a50
* Fenix version: nightly

miDeb:
I recently had an issue with my conmection where all my conmections were timing out. When trying to open fenix the home screen would not load and the browser crashed after a few seconds. That makes me think that this request _might_ be happening on the main thread, related to https://github.com/mozilla-mobile/fenix/issues/16852 (but that's really just a hypothesis).

sv-ohorvath:
We were able to reproduce (not with 100% rates) the crash manually and with automated UI tests. I also saw this crash in Firebase UI Tests: https://github.com/mozilla-mobile/fenix/issues/16889

jhugman:
This appears to be fixed by https://github.com/mozilla/nimbus-sdk/pull/67, but is not yet released.

Filing [this Jira ticket][1] to track the bug in Nimbus.

[1]: https://jira.mozilla.com/browse/SDK-140

GabrielGanne:
Hi,

I also have this issue, and I'd like to add that no crash report seems to be submitted (`about:crashes` stays empty).
I have no idea if this is related though.
Hope it helps.

Best regards

9ary:
The crash reports can be found in Settings > About > Crashes.

GabrielGanne:
> 
> 
> The crash reports can be found in Settings > About > Crashes.

Found them. I didn't know it was different on mobile.
Thank you.

[...]

PR #18112, opened 2021-02-22
For #18040: no network access startup tests

sv-ohorvath:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/androidTest/java/org/mozilla/fenix/helpers/TestHelper.kt
+++ b/app/src/androidTest/java/org/mozilla/fenix/helpers/TestHelper.kt
@@ -11,13 +11,16 @@ import android.net.Uri
 import android.os.Build
 import android.os.Environment
 import androidx.preference.PreferenceManager
+import androidx.test.espresso.Espresso
 import androidx.test.espresso.Espresso.onView
+import androidx.test.espresso.IdlingRegistry
 import androidx.test.espresso.action.ViewActions.longClick
 import androidx.test.espresso.assertion.ViewAssertions
 import androidx.test.espresso.matcher.ViewMatchers.withId
 import androidx.test.espresso.matcher.ViewMatchers.withText
 import androidx.test.platform.app.InstrumentationRegistry
 import androidx.test.uiautomator.By
+import androidx.test.uiautomator.UiDevice
 import androidx.test.uiautomator.UiScrollable
 import androidx.test.uiautomator.UiSelector
 import androidx.test.uiautomator.Until
@@ -26,6 +29,7 @@ import org.hamcrest.CoreMatchers
 import org.hamcrest.CoreMatchers.allOf
 import org.mozilla.fenix.R
 import org.mozilla.fenix.helpers.ext.waitNotNull
+import org.mozilla.fenix.helpers.idlingresource.NetworkConnectionIdlingResource
 import org.mozilla.fenix.ui.robots.mDevice
 import java.io.File
 
@@ -114,6 +118,36 @@ object TestHelper {
 
             if (downloadedFile.exists()) {
                 downloadedFile.delete()
+            }
+        }
+    }
+
+    fun setNetworkEnabled(enabled: Boolean) {
+        val mDevice = UiDevice.getInstance(InstrumentationRegistry.getInstrumentation())
+        val networkDisconnectedIdlingResource = NetworkConnectionIdlingResource(false)
+        val networkConnectedIdlingResource = NetworkConnectionIdlingResource(true)
+
+        when (enabled) {
+            true -> {
+                mDevice.executeShellCommand("svc data enable")
+                mDevice.executeShellCommand("svc wifi enable")
+
+                // Wait for network connection to be completely enabled
+                IdlingRegistry.getInstance().register(networkConnectedIdlingResource)
+                Espresso.onIdle {
+                    IdlingRegistry.getInstance().unregister(networkConnectedIdlingResource)
+                }
+            }
+
+            false -> {
+                mDevice.executeShellCommand("svc data disable")
+                mDevice.executeShellCommand("svc wifi disable")
+
+                // Wait for network connection to be completely disabled
+                IdlingRegistry.getInstance().register(networkDisconnectedIdlingResource)
+                Espresso.onIdle {
+                    IdlingRegistry.getInstance().unregister(networkDisconnectedIdlingResource)
+                }
             }
         }
     }
--- a/app/src/androidTest/java/org/mozilla/fenix/ui/NoNetworkAccessStartupTests.kt
+++ b/app/src/androidTest/java/org/mozilla/fenix/ui/NoNetworkAccessStartupTests.kt
@@ -1,0 +1,102 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.ui
+
+import androidx.core.net.toUri
+import org.junit.After
+import org.junit.Rule
+import org.junit.Test
+import org.mozilla.fenix.R
+import org.mozilla.fenix.helpers.HomeActivityTestRule
+import org.mozilla.fenix.helpers.TestHelper.packageName
+import org.mozilla.fenix.helpers.TestHelper.setNetworkEnabled
+import org.mozilla.fenix.helpers.TestHelper.verifyUrl
+import org.mozilla.fenix.ui.robots.browserScreen
+import org.mozilla.fenix.ui.robots.homeScreen
+import org.mozilla.fenix.ui.robots.navigationToolbar
+
+/**
+ * Tests to verify some main UI flows with Network connection off
+ *
+ */
+
+class NoNetworkAccessStartupTests {
+
+    @get:Rule
+    val activityTestRule = HomeActivityTestRule(launchActivity = false)
+
+    @After
+    fun tearDown() {
+        // Restoring network connection
+        setNetworkEnabled(true)
+    }
+
+    @Test
+    // Based on STR from https://github.com/mozilla-mobile/fenix/issues/16886
+    fun noNetworkConnectionStartupTest() {
+        setNetworkEnabled(false)
+
+        activityTestRule.launchActivity(null)
+
+        homeScreen {
+        }.dismissOnboarding()
+        homeScreen {
+            verifyHomeScreen()
+        }
+    }
+
+    @Test
+    // Based on STR from https://github.com/mozilla-mobile/fenix/issues/16886
+    fun networkInterruptedFromBrowserToHomeTest() {
+        val url = "example.com"
+
+        activityTestRule.launchActivity(null)
+
+        navigationToolbar {
+        }.enterURLAndEnterToBrowser(url.toUri()) {}
+
+        setNetworkEnabled(false)
+
+        browserScreen {
+        }.goToHomescreen {
+            verifyHomeScreen()
+        }
+    }
+
+    @Test
+    fun testPageReloadAfterNetworkInterrupted() {
+        val url = "example.com"
+
+        activityTestRule.launchActivity(null)
+
+        navigationToolbar {
+        }.enterURLAndEnterToBrowser(url.toUri()) {}
+
+        setNetworkEnabled(false)
+
+        browserScreen {
+        }.openThreeDotMenu {
+        }.refreshPage {}
+    }
+
+    @Test
+    fun testSignInPageWithNoNetworkConnection() {
+        setNetworkEnabled(false)
+
+        activityTestRule.launchActivity(null)
+
+        homeScreen {
+        }.openThreeDotMenu {
+        }.openSettings {
+        }.openTurnOnSyncMenu {
+            tapOnUseEmailToSignIn()
+            verifyUrl(
+                "firefox.com",
+                "$packageName:id/mozac_browser_toolbar_url_view",
+                R.id.mozac_browser_toolbar_url_view
+            )
+        }
+    }
+}
--- a/app/src/androidTest/java/org/mozilla/fenix/helpers/idlingresource/NetworkConnectionIdlingResource.kt
+++ b/app/src/androidTest/java/org/mozilla/fenix/helpers/idlingresource/NetworkConnectionIdlingResource.kt
@@ -1,0 +1,48 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.helpers.idlingresource
+
+import android.net.ConnectivityManager
+import androidx.core.content.getSystemService
+import androidx.test.espresso.IdlingResource
+import androidx.test.platform.app.InstrumentationRegistry
+import org.mozilla.fenix.ext.isOnline
+
+/**
+ * An IdlingResource implementation that waits until the network connection is online or offline.
+ * The networkConnected parameter sets the expected connection status.
+ * Only after connecting/disconnecting has completed further actions will be performed.
+ */
+
+class NetworkConnectionIdlingResource(private val networkConnected: Boolean) : IdlingResource {
+    private var resourceCallback: IdlingResource.ResourceCallback? = null
+    private val connectionManager =
+        InstrumentationRegistry.getInstrumentation().context.getSystemService<ConnectivityManager>()
+
+    override fun getName(): String {
+        return this::javaClass.name
+    }
+
+    override fun isIdleNow(): Boolean {
+        val idle =
+            if (networkConnected) {
+                isOnline()
+            } else {
+                !isOnline()
+            }
+        if (idle)
+            resourceCallback?.onTransitionToIdle()
+        return idle
+    }
+
+    override fun registerIdleTransitionCallback(callback: IdlingResource.ResourceCallback?) {
+        if (callback != null)
+            resourceCallback = callback
+    }
+
+    private fun isOnline(): Boolean {
+        return connectionManager!!.isOnline()
+    }
+}


sv-ohorvath:
@mcomella if you want to have a look at this, maybe there is something you'd like to add

app/src/androidTest/java/org/mozilla/fenix/helpers/TestHelper.kt
```diff
@@ -117,4 +125,20 @@ object TestHelper {
             }
         }
     }
+
+    fun setNetworkConnection(status: String) {
```
mcomella, line 129: nit: it'd be safer to use a boolean or enum

app/src/androidTest/java/org/mozilla/fenix/helpers/TestHelper.kt
```diff
@@ -117,4 +125,20 @@ object TestHelper {
...
+        val mDevice = UiDevice.getInstance(InstrumentationRegistry.getInstrumentation())
+
+        when (status) {
+            "ON" -> {
+                mDevice.executeShellCommand("svc data enable")
```
mcomella, line 134: Perhaps we should assert that the network is enabled/disabled using the app APIs (`ConnectivityManager` or whatever it is) after running these commands. I'm concerned about a few cases:
- it's also possible to transfer data over ADB, which the firebase data centers may be doing for stability reasons
- we may not be able to execute these commands on certain devices, e.g. firebase may disable shell access for security reasons
- it may take some time for this command to execute so the test isn't actually testing what we want it to test. For example, after enabling wifi, it takes a few seconds before the user actually has network connectivity again.

mcomella (commented):
I'm not familiar with the `svc` APIs but, assuming they'll disable the network as we expect, this looks like the approach I was thinking of – thanks! Also, thanks for looking at the STR for https://github.com/mozilla-mobile/fenix/issues/16886 and writing tests specifically for it.

app/src/androidTest/java/org/mozilla/fenix/helpers/TestHelper.kt
sv-ohorvath: Thanks for pointing these out. 

> Perhaps we should assert that the network is enabled/disabled using the app APIs (`ConnectivityManager` or whatever it is) after running these commands. 
> * it may take some time for this command to execute so the test isn't actually testing what we want it to test. For example, after enabling wifi, it takes a few seconds before the user actually has network connectivity again.

Working on this now.

> * we may not be able to execute these commands on certain devices, e.g. firebase may disable shell access for security reasons

What I've found so far is this response from a Firebase engineer, on[ Firebase community](https://firebase-community.slack.com/archives/C1MTSQ5QT/p1611169571083500): 
" williamhester  1 month ago
We don't support adb shell commands, but your Instrumentation test may run shell commands (we can't really stop you)."

app/src/androidTest/java/org/mozilla/fenix/helpers/TestHelper.kt
sv-ohorvath: done

app/src/androidTest/java/org/mozilla/fenix/helpers/TestHelper.kt
sv-ohorvath: @mcomella Took your suggestions and made some changes: added Idling resources to wait to be online/offline and fixed some nits, if you care to have another look. Thanks!

app/src/androidTest/java/org/mozilla/fenix/helpers/TestHelper.kt
```diff
@@ -117,4 +121,28 @@ object TestHelper {
...
+        val mDevice = UiDevice.getInstance(InstrumentationRegistry.getInstrumentation())
+
+        when (enabled) {
+            true -> {
+                mDevice.executeShellCommand("svc data enable")
+                mDevice.executeShellCommand("svc wifi enable")
+
+                // Stopping any next steps until the network is fully connected
+                // Make sure to unregister the Idling resource after using this method
+                IdlingRegistry.getInstance().register(NetworkConnectedIdlingResource())
```
mcomella, line 135: Why did you decide to use `IdlingResource` as a way to wait for the data to be disabled/enabled? Is it to avoid blocking the test thread and have a mechanism to time out the test if the action never happens?

With the current implementation, I'm concerned that:
- it's very easy to forget to deregister the `IdlingResource` and this state could propagate to other tests
- keeping the `IdlingResource` alive alive longer than it takes to enable/disable the network might have unexpected side effects on the tests (it's _probably_ okay but it seems hard to think through the cases)
- (see my other comment) the `IdlingResource`s are not actually being unregistered

A few alternative implementations I can think of that might help:
- if the IdlingResource deregistered itself once it was idle (though that makes it less re-usable; this might be my least favorite of these implementations)
- implementing the network calls a `Rule` such that the `Rule` registers and deregisters the IdlingResources and give you access to the `setNetworkConnection` APIs
- removing `IdlingResource` and implementing the wait + time out ourselves inside the method to change the network.

The last one seems the simplest to me without unexpected side effects but perhaps it's more error prone? In which case 2 is better but maybe it has unexpected consequences to keep the IdlingResource around? I'm not sure. It seems like a clever and interesting use case for `IdlingResource` but I want to make sure I actually understand why we're using it and the implications as I haven't used it much. :)

app/src/androidTest/java/org/mozilla/fenix/helpers/idlingresource/NetworkConnectedIdlingResource.kt
```diff
@@ -0,0 +1,33 @@
+package org.mozilla.fenix.helpers.idlingresource
```
mcomella, line 1: nit: license

app/src/androidTest/java/org/mozilla/fenix/helpers/idlingresource/NetworkConnectedIdlingResource.kt
```diff
@@ -0,0 +1,33 @@
...
+
+import android.net.ConnectivityManager
+import androidx.core.content.getSystemService
+import androidx.test.espresso.IdlingResource
+import androidx.test.platform.app.InstrumentationRegistry
+import org.mozilla.fenix.ext.isOnline
+
+class NetworkConnectedIdlingResource : IdlingResource {
```
mcomella, line 9: nit: I like all classes to have kdocs

app/src/androidTest/java/org/mozilla/fenix/helpers/idlingresource/NetworkDisconnectedIdlingResource.kt
```diff
@@ -0,0 +1,32 @@
+package org.mozilla.fenix.helpers.idlingresource
+
+import android.net.ConnectivityManager
+import androidx.core.content.getSystemService
+import androidx.test.espresso.IdlingResource
+import androidx.test.platform.app.InstrumentationRegistry
+import org.mozilla.fenix.ext.isOnline
+
+class NetworkDisconnectedIdlingResource : IdlingResource {
```
mcomella, line 9: nit: I like all classes to have kdocs

app/src/androidTest/java/org/mozilla/fenix/helpers/idlingresource/NetworkDisconnectedIdlingResource.kt
```diff
@@ -0,0 +1,32 @@
+package org.mozilla.fenix.helpers.idlingresource
```
mcomella, line 1: nit: license

app/src/androidTest/java/org/mozilla/fenix/helpers/idlingresource/NetworkDisconnectedIdlingResource.kt
mcomella, line 9: If we decide to continue with `IdlingResource`, this change isn't that important because I don't expect the classes to change much but I notice these two classes have a lot of duplication. Is there a way we can share more of their implementation?

app/src/androidTest/java/org/mozilla/fenix/ui/NoNetworkAccessStartupTests.kt
```diff
@@ -0,0 +1,109 @@
...
+ */
+
+class NoNetworkAccessStartupTests {
+    private val activityTestRule = HomeActivityTestRule()
+    private val networkDisconnectedIdlingResource = NetworkDisconnectedIdlingResource()
+
+    @After
+    fun tearDown() {
+        setNetworkConnection(true)
+        IdlingRegistry.getInstance().unregister(NetworkConnectedIdlingResource())
```
mcomella, line 34: The docs don't make it clear if this will actually unregister the resource: it's not clear if we can use a separate instance when registering or unregistering.

app/src/androidTest/java/org/mozilla/fenix/helpers/TestHelper.kt
```diff
@@ -117,4 +121,28 @@ object TestHelper {
             }
         }
     }
+
+    fun setNetworkConnection(enabled: Boolean) {
```
mcomella, line 125: nit: rename to setNetworkEnabled

mcomella (commented):
I want to better understand how we're using `IdlingResource` before we finish the review

app/src/androidTest/java/org/mozilla/fenix/helpers/idlingresource/NetworkConnectedIdlingResource.kt
mcomella: To clarify, all classes except the test classes for classes under test (e.g. no need for kdoc on `NoNetworkAccessStartupTest` unless it has a particular need for it).

The reason I want class kdoc is that sometimes the class name doesn't make the class' purpose obvious to someone using the class. For example, `HomeActivity` seems fairly obvious to fenix developer but to a new dev, it begs many questions: if there is a `HomeActivity`, `HomeFragment`, and `BrowserFragment`, where is the `BrowserActivity`? Is this `Activity` the "main" activity of the app? Are there other significant activities or is this app a single-activity type of app?

app/src/androidTest/java/org/mozilla/fenix/helpers/TestHelper.kt
sv-ohorvath: Yes, we're using Idling Resources in many of our tests to wait just until something is ready, not more or less. If we'd set a strict timeout it could result in flakiness if it takes longer or we'd have tests wait an unnecessary amount of time. I'll take your second suggestion and move both the register & unregister to the test class.

app/src/androidTest/java/org/mozilla/fenix/helpers/idlingresource/NetworkConnectedIdlingResource.kt
sv-ohorvath: done

app/src/androidTest/java/org/mozilla/fenix/helpers/idlingresource/NetworkConnectedIdlingResource.kt
sv-ohorvath: done

app/src/androidTest/java/org/mozilla/fenix/helpers/idlingresource/NetworkDisconnectedIdlingResource.kt
sv-ohorvath: done

sv-ohorvath: I removed this class and modified the implementation to work for both

app/src/androidTest/java/org/mozilla/fenix/ui/NoNetworkAccessStartupTests.kt
sv-ohorvath: done

app/src/androidTest/java/org/mozilla/fenix/helpers/TestHelper.kt
sv-ohorvath: done

sv-ohorvath:
@mcomella Thanks for the thorough review. I've addressed the change requests, please have another look when you can.

app/src/androidTest/java/org/mozilla/fenix/ui/NoNetworkAccessStartupTests.kt
```diff
@@ -0,0 +1,121 @@
...
+    fun tearDown() {
+        IdlingRegistry.getInstance().unregister(networkDisconnectedIdlingResource)
+        // Restoring network connection and waiting to be back online
+        setNetworkEnabled(true)
+        IdlingRegistry.getInstance().register(networkConnectedIdlingResource)
+        Espresso.onIdle {
+            IdlingRegistry.getInstance().unregister(networkConnectedIdlingResource)
+        }
+
+        activityTestRule.finishActivity()
```
mcomella, line 42: nit: I'd think this happens automatically. The `finishActivity` gets called after `base.evaluate`, which I think is where the test content actually runs: https://cs.android.com/androidx/android-test/+/master:runner/rules/java/androidx/test/rule/ActivityTestRule.java;l=549;drc=bf4215f8816fc81db3d105ef2fbc4f748619ab45

So I think this can be removed

app/src/androidTest/java/org/mozilla/fenix/ui/NoNetworkAccessStartupTests.kt
```diff
@@ -0,0 +1,121 @@
...
+import org.mozilla.fenix.ui.robots.homeScreen
+import org.mozilla.fenix.ui.robots.navigationToolbar
+
+/**
+ * Tests to verify some main UI flows with Network connection off
+ *
+ */
+
+class NoNetworkAccessStartupTests {
+    private val activityTestRule = HomeActivityTestRule()
```
mcomella, line 28: Not sure if it's what you want here but I added `HomeActivityTestRule(skipOnboarding = true)` if you want to avoid having to do:
```kotlin
homeScreen {
}.dismissOnboarding()
```

It sets the preference before the app starts so that it thinks it's already shown the onboarding screen.

app/src/androidTest/java/org/mozilla/fenix/ui/NoNetworkAccessStartupTests.kt
```diff
@@ -0,0 +1,121 @@
...
+    @Test
+    // Based on STR from https://github.com/mozilla-mobile/fenix/issues/16886
+    fun noNetworkConnectionStartupTest() {
+        setNetworkEnabled(false)
+        IdlingRegistry.getInstance().register(networkDisconnectedIdlingResource)
+        Espresso.onIdle {
+            IdlingRegistry.getInstance().unregister(networkDisconnectedIdlingResource)
+        }
+
+        activityTestRule.launchActivity(null)
```
mcomella, line 54: I think you need to do `HomeActivityTestRule(launchActivity = false)` to prevent the activity from already being launched by this point

app/src/androidTest/java/org/mozilla/fenix/ui/NoNetworkAccessStartupTests.kt
```diff
@@ -0,0 +1,121 @@
...
+        }
+
+        activityTestRule.finishActivity()
+    }
+
+    @Test
+    // Based on STR from https://github.com/mozilla-mobile/fenix/issues/16886
+    fun noNetworkConnectionStartupTest() {
+        setNetworkEnabled(false)
+        IdlingRegistry.getInstance().register(networkDisconnectedIdlingResource)
```
mcomella, line 49: I'm not sure when we'd want to call `setNetworkEnabled` without waiting for the operation to complete so I feel like this "wait for connected/disconnected" block would better fit inside `setNetworkEnabled` itself. There's also a small duplication of intention (i.e. if `setNetworkEnabled(false)`, you need to correctly call `networkDisconnectedIdlingResource` instead of `*Connected*`) that this would solve.

What do you think?

mcomella (approved):
Looks good to me! I think there are a few nits we can clean up but they're not too major – I think it'd be fine to land even if we decided not to do them.

Thank you!

app/src/androidTest/java/org/mozilla/fenix/ui/NoNetworkAccessStartupTests.kt
mcomella: > I think it'd be fine to land even if we decided not to do them.

Oh actually, I think this one needs to be addressed for this test to work correctly. However, I trust you can implement the correct solution so no need to get a re-review from me before landing unless you want to.

mcomella: Ah, I added code to throw if there's no network on start up and the test runs as expected (i.e. it fails with that change). I think what might be happening is that we're launching the activity twice. So I guess it'd be an optimization to avoid launching the activity twice (if I understand what's happening correctly) but that this test is working correctly.

app/src/androidTest/java/org/mozilla/fenix/ui/NoNetworkAccessStartupTests.kt
sv-ohorvath: I just wanted to do some basic navigation after startup.

app/src/androidTest/java/org/mozilla/fenix/ui/NoNetworkAccessStartupTests.kt
sv-ohorvath: done

app/src/androidTest/java/org/mozilla/fenix/ui/NoNetworkAccessStartupTests.kt
sv-ohorvath: done

app/src/androidTest/java/org/mozilla/fenix/ui/NoNetworkAccessStartupTests.kt
sv-ohorvath: moved all into setNetworkEnabled.

app/src/androidTest/java/org/mozilla/fenix/ui/NoNetworkAccessStartupTests.kt
sv-ohorvath: done

Issue #18102, opened 2021-02-21
Intermittent debug(T) failure - org.mozilla.fenix.FenixApplicationTest - GIVEN onCreate is called THEN the duration is measured


Archaeopteryx:
Log: https://treeherder.mozilla.org/logviewer?job_id=330634809&repo=fenix

```
[task 2021-02-21T00:40:21.225Z] SUITE: org.mozilla.fenix.FenixApplicationTest
[task 2021-02-21T00:40:21.225Z]   TEST: GIVEN onCreate is called THEN the duration is measured
[task 2021-02-21T00:40:21.225Z]     [Robolectric] org.mozilla.fenix.FenixApplicationTest.GIVEN onCreate is called THEN the duration is measured: sdk=28; resources=BINARY
[task 2021-02-21T00:40:21.225Z]     I/MonitoringInstr: Instrumentation started!
[task 2021-02-21T00:40:21.225Z]     I/CursorWindowStats: Created a new Cursor. # Open Cursors=1 (# cursors opened by this proc=1)
[task 2021-02-21T00:40:21.225Z]     D/SQLiteCursor: received count(*) from native_fill_window: 1
[task 2021-02-21T00:40:21.323Z]     I/CursorWindowStats: Created a new Cursor. # Open Cursors=1 (# cursors opened by this proc=1)
[task 2021-02-21T00:40:21.323Z]     D/SQLiteCursor: received count(*) from native_fill_window: 1
[task 2021-02-21T00:40:21.323Z]     I/CursorWindowStats: Created a new Cursor. # Open Cursors=1 (# cursors opened by this proc=1)
[task 2021-02-21T00:40:21.323Z]     D/SQLiteCursor: received count(*) from native_fill_window: 1
[task 2021-02-21T00:40:21.323Z]     I/CursorWindowStats: Created a new Cursor. # Open Cursors=1 (# cursors opened by this proc=1)
[task 2021-02-21T00:40:21.324Z]     D/SQLiteCursor: received count(*) from native_fill_window: 0
[task 2021-02-21T00:40:21.324Z]     I/CursorWindowStats: Created a new Cursor. # Open Cursors=1 (# cursors opened by this proc=1)
[task 2021-02-21T00:40:21.324Z]     D/SQLiteCursor: received count(*) from native_fill_window: 0
[task 2021-02-21T00:40:21.324Z]     I/CursorWindowStats: Created a new Cursor. # Open Cursors=1 (# cursors opened by this proc=1)
[...]

PR #18124, opened 2021-02-23
For #18102: Implement TestLifecycleApplication interface.

mcarare:
@mcomella 

I tried running the test in isolation, a single time and then multiple times. It appears that after the first run, `application.onCreate()` was not called for the following runs. Looking in the [link](https://robolectric.blogspot.com/2013/04/the-test-lifecycle-in-20.html) you provided in the comments I see that the steps are run **if** you implement `TestLifecycleApplication `. 
Maybe the intermittence of the test is determined by the order the test runs. I suspected that another test calls `application.onCreate()` and Robolectric does not call it again for our test. 

With this change of the test, I did not get any failures with multiple runs / random groups of tests ran before. Let's see if it passes on CI too. 

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/test/java/org/mozilla/fenix/FenixApplicationTest.kt
+++ b/app/src/test/java/org/mozilla/fenix/FenixApplicationTest.kt
@@ -13,9 +13,11 @@ import org.junit.Test
 import org.junit.runner.RunWith
 import org.mozilla.fenix.GleanMetrics.PerfStartup
 import org.mozilla.fenix.helpers.FenixRobolectricTestRunner
+import org.robolectric.TestLifecycleApplication
+import java.lang.reflect.Method
 
 @RunWith(FenixRobolectricTestRunner::class)
-class FenixApplicationTest {
+class FenixApplicationTest : TestLifecycleApplication {
 
     @get:Rule val gleanTestRule = GleanTestRule(ApplicationProvider.getApplicationContext())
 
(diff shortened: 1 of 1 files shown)


mcarare:
All tests passed on the first run.

mcomella (changes requested):
It was an interesting idea but it doesn't seem to work for me. As per Sebastian's suggestion, I used AS' "run unit test until failure" functionality to reproduce the failure (see [this Matrix message in #android-tips](https://matrix.to/#/!DFCuKPUwKFvGJvTyvP:mozilla.org/$c3mISTKdMkQLPwhCB0hDelyZYvrk16DQ53SzLUQ_XkQ?via=mozilla.org&via=matrix.org) for how to do that).

@mcarare It seems you behavior changed locally though after your changes – are you able to reproduce the failure with "run until failure"?

By the way, I would think `TestLifecycleApplication` is supposed to be implemented on `FenixRobolectricTestApplication` but I tried that and that didn't solve the problem either.

mcomella:
Sebastian also mentioned they can reproduce by adding another test to this class in https://github.com/mozilla-mobile/fenix/issues/18102#issuecomment-784349105

mcomella:
@mcarare Thanks for thinking about this – I actually don't think addressing this is a priority though so don't feel like you need to spend more time on this. See my reasoning in https://github.com/mozilla-mobile/fenix/issues/18102#issuecomment-784570811 Let me know if you disagree!

mcarare:
@mcomella Weirdly enough, for me locally it sometimes fails after the second run, and sometimes all runs pass.

PR #18125, opened 2021-02-23
Do not run the contributor workflow for MickeyMoz

st3fan:
This patch adds a check to the `build-contributor-pr.yml` workflow to make sure we do not run any actions when the PR was created by the `MickeyMoz` bot. Those are already properly handled by TaskCluster.


pocmo (approved):

escapewindow (approved):
Thanks!

PR #18153, opened 2021-02-24
Remove Session[Manager] references from Fennec intent processors

csadilek:
Switching to `TabsUseCases` to remove Session and SessionManager references from intent processors. This is the last direct usage of `SessionManager` in Fenix.


pocmo (approved):

csadilek:
Blocking this until we can do some manual testing.

pocmo:
I just tested this upgrading from Firefox Beta 68.7b to a build from this branch. Both processors continued to worked correctly. 👍

csadilek:
I did a beta migration as well now and confirmed that web app and bookmark shortcuts still work.

Issue #17804, opened 2021-02-03
Fix Activity.enterToImmersiveMode() deprecation in Android 11


Amejia481:
[Docs](https://youtu.be/acC7SR1EXsI?t=161)

https://github.com/mozilla-mobile/fenix/blob/f4a7b4c066b98e8fd999268d702f4d5d81815021/app/src/main/java/org/mozilla/fenix/ext/Activity.kt#L18

Similar to https://github.com/mozilla-mobile/android-components/issues/8517

Amejia481:
To address this issue we need to target android 11

mcarare:
IMO we should remove this piece of code and replace it with the AC version of the extension, deprecation being handled in https://github.com/mozilla-mobile/android-components/issues/8517.

Amejia481:
Yes, that will be ideal there is some discussion https://github.com/mozilla-mobile/android-components/issues/8517#issuecomment-702131591 about why we have the two different versions

mcarare:
Can QA please check if there are any issues when using fullscreen, especially on devices with Android 11? Ping me if you need more details. Thank you!

abodea:
Verified as fixed on the latest Nightly 3/17 with Google Pixel 4 XL (11) and Samsung Galaxy S7 (7), Samsung Galaxy S10+ (10).
Note that I didn`t encounter any issues regarding full-screen videos

PR #3272, opened 2019-06-07
For #1998 - Removes stable flag in entering fullscreen navigation

ekager:
### Pull Request checklist

- [X] **Quality**: This PR builds and passes detekt/ktlint checks (A pre-push hook is recommended)
- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Changelog**: This PR includes [a changelog entry](https://github.com/mozilla-mobile/fenix/blob/master/CHANGELOG.md) or does not need one
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features

Issue #1998, opened 2019-04-24
FNX-5387 ⁃ [Bug] address bar is partly hidden after using fullscreen mode (Device with physical buttons issue)


cadeyrn:
### Steps to reproduce
1. open https://davidwalsh.name/demo/fullscreen.php
1. press "Launch Fullscreen" button
1. press "Hide Fullscreen" button

### Expected behavior
The browser toolbar is visible.

### Actual behavior
Only the half of the browser toolbar is visible and you can't make it fully visible.
<img width=350 src=https://user-images.githubusercontent.com/1100614/56689528-937fad80-66db-11e9-94a3-3c90ce3e4a8a.png>

### Device information
* Android device: HTC U11 / Android 8.0
* Fenix version: Fenix revision eadce68e520d1bc2c6be853d1481b5f62b620449

ekager:
This will also be closed by https://github.com/mozilla-mobile/fenix/pull/1999

cadeyrn:
@ekager It's *not* fixed by #1999.

ekager:
There hasn't been a new build today yet with #1999 included, did you build master locally?

cadeyrn:
Yes, I tested with a new build from master (after f897c2e295cd1b97d4024c7a9cb45dceb7a2fa89 so #1999 is included).

ekager:
Have you tested this again? I am not able to reproduce on my device on master.

cadeyrn:
Yes. I compiled a new build from master (after e6c34f7045f36c091848c0ecaffbb09d5c95d35e) and it's still an issue.

![20190427_084739](https://user-images.githubusercontent.com/1100614/56845997-90e2ab00-68c9-11e9-95d7-f15546a36cf3.gif)

softvision-miralobontiu:
Hi,

I can reproduce this issue on HTC 10 (Android 8.0), on the latest build 1.0.1920 (#11350639).
I followed the STR from [Comment 0](https://github.com/mozilla-mobile/fenix/issues/1998#issue-436878271):
1. Open https://davidwalsh.name/demo/fullscreen.php
2. Tap on "Launch Fullscreen" from the page loaded at Step 1.
3. Tap on "Hide Fullscreen" from the page loaded at Step 1.
4. Observe the address bar.

Thanks,

Notes: Not reproducible on Google Pixel (Android 9), and Sony Xperia Z5 Premium (Android 7.1.1).

ekager:
[...]

PR #18169, opened 2021-02-25
For #17804: Fix setSystemUiVisibility and flags deprecation.

mcarare:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/ext/Activity.kt
+++ b/app/src/main/java/org/mozilla/fenix/ext/Activity.kt
@@ -5,7 +5,10 @@
 package org.mozilla.fenix.ext
 
 import android.app.Activity
+import android.os.Build
 import android.view.View
+import android.view.WindowInsets
+import android.view.WindowInsetsController
 import android.view.WindowManager
 import mozilla.components.concept.base.crash.Breadcrumb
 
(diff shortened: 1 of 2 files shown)


mcarare:
cc @boek for data review.

jonalmeida (commented):
I'm not sure what the data review is for, but I tested out this patch and found so inconsistencies on Android 11:

When you enter then exit a fullscreen video, the system navigation bar disappears.

Left is this PR, right is original behaviour:
out

On the other hand, I've tested out [this PR](https://github.com/mozilla-mobile/android-components/pull/8577) and when using the AC `enterToImmersiveMode` in place of this code, it seems to work fine. So we could land it and use this PR to remove the Fenix implementation.

If you'd like to land this change, I _think_ we need to take out the `setDecorFitsSystemWindows`. We can land the AC change in a future update. 🙂

mcarare:
@jonalmeida Thanks for testing, I have removed `setDecorFitsSystemWindows(false) `. I think it should be used only after we change the way we set the window adjust/ resize for SDK >=30 and do it properly. I was using it because the documentation suggests it should be set to false in order to not check the deprecated flags anymore. I will revisit this at a later time.

There is no data review requested here, I am not sure what are you referring to. 🤔 

As for the AC extension, this was also my initial thought (https://github.com/mozilla-mobile/fenix/issues/17804#issuecomment-783395001) but when looking into the code I've seen this comment:
>We don't use the equivalent function from Android Components because the stable flag messes
  with the toolbar. See #1998 and #3272.`

jonalmeida:
> There is no data review requested here, I am not sure what are you referring to. 🤔

Maybe this comment was meant for another PR? https://github.com/mozilla-mobile/fenix/pull/18169#issuecomment-788088636




> As for the AC extension, this was also my initial thought ([#17804 (comment)](https://github.com/mozilla-mobile/fenix/issues/17804#issuecomment-783395001)) but when looking into the code I've seen this comment:
> 
> > We don't use the equivalent function from Android Components because the stable flag messes
> > with the toolbar. See #1998 and #3272.`

I agree with your comment on the issue. In my testing those two bugs are no longer seen when we use the AC version now.

Issue #17190, opened 2020-12-22
[Bug] Changing language while notifications are showing will not update notification


ekager:
## Steps to reproduce
1. Enter Private Browsing Mode
2. Open a tab
3. Open device notifications and see PBM notification 
4. Go to settings and change the language 

<img width="322" alt="image" src="https://user-images.githubusercontent.com/43795363/106054898-426fcb00-60b2-11eb-9270-ae699553e216.png">

### Expected behavior
Notification re-creates and reflects new language

### Actual behavior
Notification does not recreate. Language updated only on app restart. 

### Device information

* Android device: Android 11, Pixel 3
* Fenix version: Nightly

eliserichards:
Expected flow (Google Chrome)
1. Open Chrome
2. Open a private browsing session, see the PBM notification in the device notifs
3. Background Chrome
3. Go to device settings, change language
4. Foreground Chrome, see that the language of the notification changes once Chrome is brought back to the foreground

Mugurell:
@eliserichards I'm seeing this ticket in "Ready for engineering"
so with the current backlog running low I am wondering if you are actively working or planning to work on this or it can be picked up by others?

eliserichards:
## For QA

Fenix decides it's language in two ways: either following the device language, or choosing one from Fenix settings. When the user chooses a language in Fenix, we no longer care if the user changes their device language (we persist the language set in the app). 

Steps to test this fix: 
1. Open Fenix
2. Open a private tab(s), see the notification in the system notifications
3. Go to Fenix app settings, select another language
4. Look at the notifications and see the notification language has updated

sflorean:
Verified as fixed on latest Nightly build with Pixel 3 (Android 11).

PR #18179, opened 2021-02-25
For #17190: notifications are updated when locale is changed

eliserichards:
For #17190 
Requires https://github.com/mozilla-mobile/android-components/pull/9779 ✅ 

Allow private notifications to respond & update when a user changes the app language in Settings.

- Define what happens when we receive a locale change from the browser store: when a private notification already exists, `notify` the notification that the locale has changed. This updates the existing notification to the correct language.

STR:
1. Open private tabs, see the notification in the system notifs
2. Go to app settings, select another language
3. Look at the system notifications and see the notification language has updated

## Screenshots

Everything set to English -> Set Fenix language to Bosnian -> Set system language to Spanish
**Fenix notif still set to Bosnian** 👍 
image

Everything set to English -> Set Fenix language to Bosnian
**Notification is updated**
notification_language_updates_3

### Pull Request checklist

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/session/PrivateNotificationService.kt
```diff
@@ -33,9 +37,48 @@ class PrivateNotificationService : AbstractPrivateNotificationService() {
...
+            id = "browsing-session",
+            name = R.string.mozac_feature_privatemode_notification_channel_name,
+            importance = NotificationManagerCompat.IMPORTANCE_LOW
+        )
+
+        // List of Intent actions that will get ignored when they are in the root intent that gets
+        // passed to onTaskRemoved().
+        private val ignoreTaskActions = listOf(
+            "mozilla.components.feature.pwa.VIEW_PWA"
+        )
```
Amejia481, line 81: I don't see where we are using `ignoreTaskActions` ? :)

Amejia481:
We are also going to need update the locale on the `store` on app startup similar to [what we are doing with other states](https://github.com/mozilla-mobile/fenix/blob/607d6382f89d9b10cc9628637dc1d0dab7518ca8/app/src/main/java/org/mozilla/fenix/FenixApplication.kt#L143), as `LocaleManager` is storing the local on the disk but the value on the `store` is not, be have to keep them sync even after app startups.

app/src/main/java/org/mozilla/fenix/session/PrivateNotificationService.kt
eliserichards: THis was a copy/paste from the abstract notification service. It's been removed

Amejia481 (approved):

PR #18216, opened 2021-02-28
Closes #18150: select tabs button not displayed

### LINKED ISSUE #18150: [Bug] "Select tabs" button not displayed
* STR:
open 2 or more websites in normal mode
switch to private browsing mode
open a website
close opened website
Go to normal tabs 
"Select tabs" button not displayed

* expected behaviour
show "Select tabs" button

* actual behaviour
"Select tabs" button not visible

reproducible:
* version: 86.1.1
* version: 87.0.0-beta.1 (Build #2015795065) AC: 73.0.0, 69a800e47 GV: 87.0-20210222154334 AS: 71.0.0
* version: Nightly 210224 (Build #2015795113) AC: 73.0.20210223143117, 05f3e3cc7 GV: 87.0a1-20210222084406 AS: 71.0.0

interesting note (continuation from STR):
go back to pbm
come back to normal bm
"Select tabs" button is visible

excuse cant submit logs.

> android 9, Galaxy J7 Nxt

VitalyVPinchuk:
Fixes #18150

`NotifyItemChanged()` with `payload` has no effect for the first time as the whole view needs to be updated.
Added a method to update private mode state in `SaveToCollectionsButtonAdapter` to address the issue.
Added smoke test but not sure if such a test is needed.

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/androidTest/java/org/mozilla/fenix/ui/SmokeTest.kt
+++ b/app/src/androidTest/java/org/mozilla/fenix/ui/SmokeTest.kt
@@ -1056,6 +1056,33 @@ class SmokeTest {
     }
 
     @Test
+    fun selectTabsButtonVisibilityTest() {
+        homeScreen {
+        }.dismissOnboarding()
+
+        val firstWebPage = TestAssetHelper.getGenericAsset(mockWebServer, 1)
+        val secondWebPage = TestAssetHelper.getGenericAsset(mockWebServer, 2)
+
+        navigationToolbar {
+        }.enterURLAndEnterToBrowser(firstWebPage.url) {
+            mDevice.waitForIdle()
+        }.openTabDrawer {
+        }.openNewTab {
+        }.submitQuery(secondWebPage.url.toString()) {
+            mDevice.waitForIdle()
+        }.openTabDrawer {
+        }.toggleToPrivateTabs {
+        }.openNewTab {
+        }.dismissSearchBar {  }
+
+        homeScreen {  }
+            .openTabDrawer {
+            }.toggleToNormalTabs {
+                verifySelectTabsButton()
+            }
+    }
+
+    @Test
     fun privateTabsTrayWithOpenedTabTest() {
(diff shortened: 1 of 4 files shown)


app/src/androidTest/java/org/mozilla/fenix/ui/SmokeTest.kt
```diff
@@ -1055,6 +1055,33 @@ class SmokeTest {
...
+        }.openTabDrawer {
+        }.openNewTab {
+        }.submitQuery(secondWebPage.url.toString()) {
+            mDevice.waitForIdle()
+        }.openTabDrawer {
+        }.toggleToPrivateTabs {
+        }.openNewTab {
+        }.dismissSearchBar { }
+
+        homeScreen { }
```
Mugurell, line 1078: nit: Maybe we can have the same formatting as the above here?
as in 

```suggestion
        homeScreen {
        }.openTabDrawer {

```

Mugurell (commented):
Thank you for looking into this!
I really like the tests!

I see the issue stemming from the fact that when we decide whether to show or not the button there are two things checked:
- a fixed `isPrivate` value
- a dynamic ~`isModeNormal` value

I'd want to ask you how you'd feel about maybe refactoring that 
`private val isPrivate: Boolean = false` property to something like 
`private val isPrivate: () -> Boolean = { false }`
(that in TabTrayView would be initialized with `{ isPrivateModeSelected }`) ?

There is a lot of technical debt already so I agree there's no clean solution but I'd prefer to not add new APIs (the new setter) and rather force clients to initialize the adapter with `isPrivate` also as a dynamic property.

app/src/androidTest/java/org/mozilla/fenix/ui/SmokeTest.kt
VitalyVPinchuk: Hello @Mugurell 
I really appreciate your feedback and help! I'll fix all the remarks.

Mugurell (approved):
LGTM.
Thank you!

Issue #18223, opened 2021-03-01
"Open in app" stopped working


erikhakansson:
I'm on Firefox Nightly, and since a few days back, I can no longer open links in external apps.

Links just open in the browser, and the old menu item that could manually force opening of the external app is gone.

I have verified that opening in external apps is on in settings, and I've even force killed FF to rule out temporary error.

Even weird is that when some website has an intent link, FF will open the Play Store app page for that app, even if I already have it installed. Similar to this old issue: https://github.com/mozilla-mobile/fenix/issues/13504

hakkikaancaliskan:
Actually duplicate of https://github.com/mozilla-mobile/fenix/issues/18195
Currently firefox nightly targeting Android 11 and that's causing firefox to can't see other installed apps

erikhakansson:
Ah, thank you @hakkikaancaliskan. Sounds reasonable. I'll track that issue instead.

mcarare:
Keeping this issue open to track this specific issue, not the general package visibility issue in https://github.com/mozilla-mobile/fenix/issues/18195.

keegan-lillo:
This is now fixed for me!

Nightly 210304 17:01
Pixel 4 - Android 11 - March 2021 security update.

sflorean:
Verified as fixed on Nightly 5/3 with the following devices:
- Pixel 3 (Android 11)
- LG G7 Fit (Android 8.1)

erikhakansson:
Can confirm. Works fine now!

Issue #18195, opened 2021-02-26
Share dialog and other app lists within Fenix like "Open in App" and context menus are missing almost all apps


KWierso:
Not sure exactly what's going on here. It was working correctly last night, but is having an issue this morning, so maybe a recent Nightly update messed it up?

The issue:
I have Twitter installed as a PWA via Firefox Nightly on my Pixel 3 phone. I see a tweet in that PWA that I want to share into a Discord server with the Discord Android app. I tap the tweet's "share" button, then choose "share tweet via..." from the menu that opens. This pops up Fenix/Android's Share dialog, giving me options to send to various Firefox Sync enabled devices/applications and other Android apps that can receive the shared item.

When things are working correctly, Discord is one of the apps in the Share dialog, and I am able to share the tweet into Discord.

At the moment, none of the other Android apps are being shown in the Share dialog: 
<img src=https://user-images.githubusercontent.com/172215/109328702-9a633400-780e-11eb-9438-73aad89ab2bf.png width = 40%>

I only see "bluetooth" under "recently used", and "nearby share" under "all actions".

KWierso:
The "share" dialog in non-Nightly Fenix works correctly. It does not work for me on any site in Nightly, so steps to reproduce in Nightly are just "visit any site && select the share icon from Fenix's main menu". No PWA or Twitter needed...

LittleAngelwings:
I can't confirm this issue with firefox nightly. 
![Screenshot 2021-02-26 23 40 10](https://user-images.githubusercontent.com/44252450/109362842-26677100-788c-11eb-9204-0794c687c6bd.png)

hakkikaancaliskan:
I can confirm all apps gone with 2/26 nightly.
![Screenshot_20210227-021710_Firefox_Nightly](https://user-images.githubusercontent.com/17825767/109365334-fcb94480-78a1-11eb-8852-0e5a9b82f186.png)

KWierso:
[Random stab in the dark]: I see [this commit](https://github.com/mozilla-mobile/fenix/commit/0e8dc6164d8d11bc5267e31c3e4f8759054f5828#diff-a9a4df7333234d458fdd9fe94a2643cf06d8250c0f5f53d110d72c68cf5fa6b9) from two days ago over in android-components slightly changed ShareViewModel.kt, related to targeting Android 11.

I'm on Android 11, and am now having troubles with the ShareView.

More detailed version information:
Nightly 210226
[...]

PR #18229, opened 2021-03-01
For #18195: Handle Package visibility changes in Android 11

mcarare:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -18,6 +18,10 @@
     <!-- Needed to prompt the user to give permission to install a downloaded apk -->
     <uses-permission-sdk-23 android:name="android.permission.REQUEST_INSTALL_PACKAGES" />
 
+    <!-- Needed to interact with all apps installed on a device -->
+    <uses-permission android:name="android.permission.QUERY_ALL_PACKAGES"
+        tools:ignore="QueryAllPackagesPermission" />
+
     <application
         android:name=".FenixApplication"
         android:allowBackup="false"


pocmo (commented):
The fix looks good. But I wonder if we should go to `QUERY_ALL_PACKAGES` directly. It seems like we will not get around if for some functionality. Even Google lists browsers as an exception.

mcarare:
I thought of that, but I am not sure if and what downside would be to using `QUERY_ALL_PACKAGES`. There are none mentioned and browsers are in the apps listed as proper use cases for that. 

Besides privacy concerns, I am not sure why there is also a warning about using it. I thought we could use specific queries on the Fenix side as there are fewer use cases here, and as our policy is to gahter just info we strictly need.

I am not sure about possible changes in AC. For example, using `QUERY_ALL_PACKAGES` in Fenix manifest does not fix https://github.com/mozilla-mobile/fenix/issues/18223

pocmo:
> Besides privacy concerns, I am not sure why there is also a warning about using it. I thought we could use specific queries on the Fenix side as there are fewer use cases here, and as our policy is to gahter just info we strictly need.

I don't necessarily see a privacy concern since that is not data that gets shared outside of the application.

> I am not sure about possible changes in AC. 

One thing I was thinking about is Android Autofill (password managers seem to be another exception managed by Google). In that situation I need to query the package manager to read the signatures of the application to autofill. And that I cannot really express as an intent query.

Amejia481:
> One thing I was thinking about is Android Autofill (password managers seem to be another exception managed by Google). In that situation I need to query the package manager to read the signatures of the application to autofill. And that I cannot really express as an intent query.

Same requirements for allowing external download managers https://github.com/mozilla-mobile/fenix/issues/18236

mcarare:
> I don't necessarily see a privacy concern since that is not data that gets shared outside of the application.

Personally, I do not see privacy concerns either, I just brought it up because it is mentioned several times in the documentation. So I am not sure how it will impact our Google Play listing for example.

Quoted from [package-visibility](https://developer.android.com/training/basics/intents/package-visibility):
>By making some apps not visible by default, the system helps encourage the principle of least privilege by telling the system which other apps to make visible to your app, and it helps app stores like Google Play assess the privacy and security that your app provides for users.

>To respect user privacy, your app should request the smallest amount of package visibility necessary in order for your app to work.

>In an upcoming policy update, look for Google Play to provide guidelines for apps that need the QUERY_ALL_PACKAGES permission.


IMO using `QUERY_ALL_PACKAGES` is the most direct and simple solution, I just wished the documentation was more clear about it's usage implications.

Amejia481:
Adding  `<uses-permission android:name="android.permission.QUERY_ALL_PACKAGES" />` fixes #18236, I think we should add the permission as we have a valid use case for it.

mcarare:
I have updated the PR with the permission. We should follow-up on the future Google Play policies changes about using this permission and implications to our app listing in the store.

pocmo (approved):
👍 

Maybe this will cause a manual or more thorough review of the app the next time we publish it on Google Play. A couple of times in the past review/updates were delayed. Let's see, maybe this will trigger something too. :)

Issue #12528, opened 2020-07-13
[l10n] Some locales under language list appear as locale codes (vs actual language name)


Delphine:
## Steps to reproduce
Go to Settings  > Languages
Look at list of available languages

### Expected behavior
All locales have their names spelled out correctly in the language list

### Actual behavior
Some locales only appear as locale codes, instead of their full name.
Depending on device/carrier, see for example: Occitan (oc), Triqui (trs), Kaqchikel (cak), Venitian (vec) 

### Device information
(always on latest Fenix Nightly)
I've seen this happen on Pixel 2, Android version 10
Also Sony Z3, Android version 5.0.2
A localizer has sent us the same issue, I don't know what the device info.

The results on locales seem to vary depending on carrier and Android version. For example, on the Pixel 2 I see "Ligurian" (lij) spelled out correctly. On the older devices, I see it only as a locale code, so "lij". Axel mentioned this might be because there are different CLDR versions in different Android versions. https://github.com/unicode-org/cldr/blob/master/common/main/en.xml#L445 exists, but not "cak" or "trs". 
I believe we're using the built-in language names. Neither of them appear as locales in https://github.com/unicode-org/cldr/tree/master/common/main (also not as seed locales).

This has happened in the past with Fennec and Focus Android. I'd file issues for Focus Android so we could specify ourselves the actual locale name. Not sure we can import what we have from Focus Android to help resolve this.

s-ankur:
For me the names are correct on device language, but when you change it all of them change into language codes
![Screenshot_20200714-150403_Firefox_Nightly](https://user-images.githubusercontent.com/33252312/87410128-6e63ad00-c5b5-11ea-96ce-51d3f5ff56b8.png)

liuche:
@person808 while you're in non-standard locales-land from #13066, is this something that could be caused by mismatch between the locale codes? (If so, this could also be a larger problem that we would need to figure out in Pontoon, e.g. where should locale code mismatch resolution happen)

person808:
I'm not sure. I think we get the locale names from Android but I haven't looked at the code for this screen.

liuche:
May need to add a mapping, and that would solve these issues:
https://github.com/boek/focus-android/blob/ba0f3231add6529585dea4ac9b7aed943069fc01/app/src/main/java/org/mozilla/focus/widget/LocaleListPreference.java#L45

mcarare:
[...]

PR #18311, opened 2021-03-03
For #12528: Provide fallback values for locale display name.

mcarare:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/settings/advanced/LocaleViewHolders.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/advanced/LocaleViewHolders.kt
@@ -20,15 +20,230 @@ class LocaleViewHolder(
 ) : BaseLocaleViewHolder(view, selectedLocale) {
 
     override fun bind(locale: Locale) {
+        if (locale.toString().equals("vec", ignoreCase = true)) {
+            locale.toString()
+        }
         // Capitalisation is done using the rules of the appropriate locale (endonym and exonym).
-        locale_title_text.text = locale.getDisplayName(locale).capitalize(locale)
+        locale_title_text.text = getDisplayName(locale)
         // Show the given locale using the device locale for the subtitle.
-        locale_subtitle_text.text = locale.displayName.capitalize(Locale.getDefault())
+        locale_subtitle_text.text = locale.getProperDisplayName()
         locale_selected_icon.isVisible = isCurrentLocaleSelected(locale, isDefault = false)
 
         itemView.setOnClickListener {
             interactor.onLocaleSelected(locale)
         }
+    }
+
+    private fun getDisplayName(locale: Locale): String {
+        val displayName = locale.getDisplayName(locale).capitalize(locale)
+        if (displayName.equals(locale.toString(), ignoreCase = true)) {
+            return LOCALE_TO_DISPLAY_NATIVE_NAME_MAP[locale.toString()] ?: displayName
+        }
+        return displayName
+    }
+
+    @SuppressWarnings("LargeClass")
+    companion object {
+        val LOCALE_TO_DISPLAY_NATIVE_NAME_MAP: Map<String, String> = mapOf(
+            "an" to "Aragonés",
+            "anp" to "अंगिका",
+            "ar" to "العربية",
+            "ast" to "Asturianu",
+            "ay" to "Aimara",
+            "az" to "Azərbaycan dili",
+            "be" to "беларуская мова",
+            "bg" to "български език",
+            "bn" to "বাংলা",
+            "br" to "Brezhoneg",
+            "bs" to "Bosanski jezik",
+            "ca" to "Català",
+            "cak" to "Kaqchikel",
+            "ceb" to "Cebuano",
+            "co" to "Corsu, ",
+            "cs" to "čeština",
+            "cy" to "Cymraeg",
+            "da" to "dansk",
+            "de" to "Deutsch",
+            "dsb" to "dolnoserbski",
+            "el" to "ελληνικά",
+            "eo" to "Esperanto",
+            "es" to "Español",
+            "et" to "Eesti",
+            "eu" to "Euskara",
+            "fa" to "فارسی",
+            "ff" to "Fulfulde",
+            "fi" to "Suomi",
+            "fr" to "Français",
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/fenix/settings/advanced/LocaleViewHolders.kt
```diff
@@ -21,15 +21,226 @@ class LocaleViewHolder(
...
+        val displayName = locale.getDisplayName(locale).capitalize(locale)
+        if (displayName.equals(locale.toString(), ignoreCase = true)) {
+            return LOCALE_TO_DISPLAY_NATIVE_NAME_MAP[locale.toString()] ?: displayName
+        }
+        return displayName
+    }
+
+    @SuppressWarnings("LargeClass")
+    companion object {
+        val LOCALE_TO_DISPLAY_NATIVE_NAME_MAP: Map<String, String> = mapOf(
```
eliserichards, line 44: Hmm I'm not sure that I like having hard-coded names like this in the ViewHolder. I believe there should be a way to handle this using [Locale](https://developer.android.com/reference/java/util/Locale) or [LocaleList](https://developer.android.com/reference/android/os/LocaleList), but I think we should have this logic located in Android Components in the [LocaleManager.kt](https://github.com/mozilla-mobile/android-components/blob/master/components/support/locale/src/main/java/mozilla/components/support/locale/LocaleManager.kt)

eliserichards: What do you think? 🤔

app/src/main/java/org/mozilla/fenix/settings/advanced/LocaleViewHolders.kt
```diff
@@ -21,15 +21,226 @@ class LocaleViewHolder(
...
+            "uk" to "Ukrainian",
+            "ur" to "Urdu",
+            "uz" to "Uzbek",
+            "vec" to "Venitian",
+            "vi" to "Vietnamese"
+        )
+    }
 }
 
 class SystemLocaleViewHolder(
```
gabrielluong, line 246: It seems reasonable to me that we would want smaller classes and files in general for easier readability and maintenance. I would also want to consider the opportunity to extract `SystemLocaleViewHolder` and `BaseLocaleViewHolder` into their own files.

app/src/main/java/org/mozilla/fenix/settings/advanced/LocaleViewHolders.kt
Amejia481: @mcarare does Focus has a similar logic? If so, I think it would be a good idea to have the code on AC to de-duplicate the logic from both source codes, this way we will has a single source of truth. If we can't address it as part of this pr we could add reference ticket for addressing it on follow-up task. Either here or on AC it will be good to add test for covering this new behavior.

app/src/main/java/org/mozilla/fenix/settings/advanced/LocaleViewHolders.kt
mcarare: @gabrielluong I agree, but this is some technical debt existing in many other files and some of them are a matter of style and choice (i.e. having an interface in the same file or different file than classes that implement it). I would gladly handle this tech debt issue in a separate PR.

app/src/main/java/org/mozilla/fenix/settings/advanced/LocaleViewHolders.kt
mcarare: @Amejia481  I am not sure, 🤔 . There already is an extension in AC `Locale.getLocaleTag()` but I am not sure what it is used for. I still do not think Locale Manager should handle logic outside of setting or getting the system locale, meaning logic that describes behavior. IMHO display logic should be left to the client, or at best, handled separately from "management" logic.

mcarare:
@eliserichards, @Amejia481, @gabrielluong While I agree with your comments, they seem outside the scope of this PR, requiring refactoring unrelated to this change. Also, the refactorings you are suggesting need some discussion to set some guide points to be followed in the entire app to ensure consistency. 
As this PR only adds fallback values to existing logic, my proposal is to review only the current fix so we can solve the issue our users have, and handle the technical debt in separate issues that I would gladly work on if needed.

Thank you all for your input!

app/src/main/java/org/mozilla/fenix/settings/advanced/LocaleViewHolders.kt
Amejia481: I was more concern about duplicating logic that other apps could be using as well, if this is the case we could put it in any place on ac as long as apps can share the functionality, but as you mentioned below this could be addressed in separate ticket, just let add some tests to cover the new behavior. :)

Amejia481: Apart from that it's OK for me to land.

mcarare: @Amejia481 I added a couple of tests and confirmed that the results are OS-dependent. I added a comment so we can revise the tests when we can use SDK 30 in unit tests.

For example, if I run the app on a device, the display name for `vec` locale is "Venitian" for SDK 30 and that is the value displayed on the device (It actually might be wrong because the native language name is "Vèneto", but that is a L10n discussion). That would mean that when run on SDK 30 the test would fail. 
The actual result when running the test with SDK 28 is that the display name for `vec` locale is "vec" and the display value uses the fallback value of "Vèneto" and the test passes.

Amejia481 (approved):

app/src/main/java/org/mozilla/fenix/settings/advanced/LocaleViewHolders.kt
Amejia481: 👍🏽

mcarare:
Also tagging @Delphine for a quick look at the 2 lists. Either way, we can easily change or add items based on feedback from users or the L10n team.

Delphine:
@mcarare : I've looked through the file specifically with locale code names matched to native locale, and this makes sense to me. Thanks!

Issue #18132, opened 2021-02-23
Update in-app Icons


amedyne:
Update in-app Icons
(Place holder story for updating In-app Icons.)

RiyaJ2311:
Hello!
I'm Riya, an Outreachy applicant.
I'm interested in this issue and would love to be assigned to it.
@amedyne how can I contribute?

amedyne:
> Hello!
> I'm Riya, an Outreachy applicant.
> I'm interested in this issue and would love to be assigned to it.
> @amedyne how can I contribute?

Hi @RiyaJ2311 ! Thank you for your interest! We are waiting for the icons to be ready, I will let you know once they are. For the mean time, you can start here: https://github.com/mozilla-mobile/shared-docs/blob/master/android/CONTRIBUTING.md

RiyaJ2311:
> Hi @RiyaJ2311 ! Thank you for your interest! We are waiting for the icons to be ready, I will let you know once they are. For the mean time, you can start here: https://github.com/mozilla-mobile/shared-docs/blob/master/android/CONTRIBUTING.md

Thank you for sharing this with me, I went through this and looking forward to contributing to the org.

Amejia481:
Below a set of tasks for this ticket:

- [ ] Create a feature (nightly or debug) flag for new look and feel, an example can be found [here](https://github.com/mozilla-mobile/fenix/commit/6dfd7ef757e6dd42f09e33b976bcd952e68dd3e4).
- [ ] Remove icons from the settings page, this should behind the previously created feature flag. To facilitate the task we could create a temporary [preferences.xml](https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/res/xml/preferences.xml) file without the icons and switch to it when being in a build variant with the feature flag.
- [ ] Update the rest of the app with the new icon set, this should be also behind the same feature flag, it will allow us to update icons as we get them from the designers.

Amejia481:
👋🏽 
@RiyaJ2311 would you like to start working on this issue?, I added some task that can be started now without having the final icons https://github.com/mozilla-mobile/fenix/issues/18132#issuecomment-789966047

Amejia481:
@RiyaJ2311 I opened a [PR](https://github.com/mozilla-mobile/fenix/pull/18342) for addressing some of the tasks, please feel free to take some of the remaining ones if you like :)

Imerion:
[...]

PR #18342, opened 2021-03-05
For issue #18132 remove icons from settings page

Amejia481:
This PR removes the icons from the setting page and put it behind a feature flag only active in debug and nightly. Below a view of the end result.

https://user-images.githubusercontent.com/773158/110135493-969a5900-7d9c-11eb-830d-ebba5c2ad3bd.mp4

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/settings/SettingsFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/SettingsFragment.kt
@@ -40,6 +40,7 @@ import org.mozilla.fenix.BrowserDirection
 import org.mozilla.fenix.Config
 import org.mozilla.fenix.HomeActivity
 import org.mozilla.fenix.R
+import org.mozilla.fenix.FeatureFlags
 import org.mozilla.fenix.components.metrics.Event
 import org.mozilla.fenix.ext.application
 import org.mozilla.fenix.ext.components
@@ -132,7 +133,12 @@ class SettingsFragment : PreferenceFragmentCompat() {
     }
 
     override fun onCreatePreferences(savedInstanceState: Bundle?, rootKey: String?) {
-        setPreferencesFromResource(R.xml.preferences, rootKey)
+        val preferencesId = if (FeatureFlags.newIconSet) {
+            R.xml.preferences_without_icons
+        } else {
+            R.xml.preferences
+        }
+        setPreferencesFromResource(preferencesId, rootKey)
         updateMakeDefaultBrowserPreference()
     }
 
@@ -368,8 +374,10 @@ class SettingsFragment : PreferenceFragmentCompat() {
         val preferenceOpenLinksInExternalApp =
             findPreference<Preference>(getPreferenceKey(R.string.pref_key_open_links_in_external_app))
 
-        preferencePrivateBrowsing.icon.mutate().apply {
-            setTint(requireContext().getColorFromAttr(R.attr.primaryText))
+        if (!FeatureFlags.newIconSet) {
+            preferencePrivateBrowsing.icon.mutate().apply {
+                setTint(requireContext().getColorFromAttr(R.attr.primaryText))
+            }
         }
 
         if (!Config.channel.isReleased) {
--- a/app/src/main/java/org/mozilla/fenix/FeatureFlags.kt
+++ b/app/src/main/java/org/mozilla/fenix/FeatureFlags.kt
@@ -43,4 +43,9 @@ object FeatureFlags {
      * Enables the tabs tray re-write with Synced Tabs.
      */
     val tabsTrayRewrite = Config.channel.isNightlyOrDebug
+
+    /**
+     * Enables the updated icon set look and feel.
+     */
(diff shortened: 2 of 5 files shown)


app/src/main/java/org/mozilla/fenix/settings/SettingsFragment.kt
```diff
@@ -132,7 +133,12 @@ class SettingsFragment : PreferenceFragmentCompat() {
     }
 
     override fun onCreatePreferences(savedInstanceState: Bundle?, rootKey: String?) {
-        setPreferencesFromResource(R.xml.preferences, rootKey)
+        val preferencesId = if (FeatureFlags.newIconSet) {
+            R.xml.preferences_without_icons
```
eliserichards, line 137: Perfect! 👌

app/src/main/java/org/mozilla/fenix/settings/SettingsFragment.kt
```diff
@@ -368,8 +374,10 @@ class SettingsFragment : PreferenceFragmentCompat() {
         val preferenceOpenLinksInExternalApp =
             findPreference<Preference>(getPreferenceKey(R.string.pref_key_open_links_in_external_app))
 
-        preferencePrivateBrowsing.icon.mutate().apply {
-            setTint(requireContext().getColorFromAttr(R.attr.primaryText))
+        if (!FeatureFlags.newIconSet) {
+            preferencePrivateBrowsing.icon.mutate().apply {
+                setTint(requireContext().getColorFromAttr(R.attr.primaryText))
+            }
```
eliserichards, line 380: Why do we need to use `.mutate().apply` instead of just `.apply`?

Amejia481: I'm not really sure, I didn't added initially I just moved the line XD, I will investigated :)

Amejia481: Thanks @cadeyrn  for the clarification :)

NotWoods: Yup, otherwise any changes will affect other instances of the drawable

Amejia481: Thanks 👋🏽

eliserichards (approved):
💯

app/src/main/java/org/mozilla/fenix/settings/SettingsFragment.kt
NotWoods: Miss y'all too!

cadeyrn:
If you remove all the icons from the settings page should the reserved icon space not be removed as well? It looks really odd with no icons but such a large gap. For reference: Neither in Chrome nor in the Samsung browser there is such a large gap.

brampitoyo (approved):
For an initial removal, this looks good!

Our next (and final) step for Settings is to reduce the left-side margin/padding to be equal to the right-side, so that there’s no ‘missing’ empty space between the eliminated icon and the text.

I’m aware that it will cause some layout inconsistencies, especially when a menu item without an icon appears alongside a menu item with radio button on the left. That’s perfectly OK.

We can do this change if it’s easy, or file a follow-up issue if you’d like me to review all Settings page.

Amejia481:
@brampitoyo thanks for taking a look I will try to do it in this PR but if it requires a lot of time I will address it on follow-up

Amejia481:
@brampitoyo  I updated the PR removing the left-side margin/padding. I think we will need to rethink all the different states of the sync item like not logged in, logged in, error etc.
  
https://user-images.githubusercontent.com/773158/110876418-74756f00-82a5-11eb-8dbb-49a2b0a1f7c8.mp4

brampitoyo (approved):
@Amejia481 The removal of icons from Settings root page looks good to me, as it now aligns with desktop and iOS.

Have margins on other Settings sub-pages been tweaked, as well? We can do that in another issue.

Amejia481:
> @Amejia481 The removal of icons from Settings root page looks good to me, as it now aligns with desktop and iOS.
> 
> Have margins on other Settings sub-pages been tweaked, as well? We can do that in another issue.

Thanks for the reviews.
No it hasn't been done, I will address it on another issue :)

Imerion:
I must ask here as well, how come the icons in the options menu are to be removed? They help with finding the right option quickly, look good, makes it easier to keep track of movement when scrolling and gives the menu a professional and polished look. As far as I can see they are also still in the desktop version when going into Settings, and look good there as well. (Though I really with they were colorful and not flat.)

Amejia481:
@brampitoyo could help to clarify your doubts.

MPeti1:
I also don't think it's a good idea to remove the icons. I mean, remove it just because they will soon get removed from the desktop version of Firefox too? That redesign is a huge mistake on it's own, please don't follow them here :(

Issue #18032, opened 2021-02-17
Include Report Site Issue on menu in nightly/debug


eliserichards:
Meta: #17796

UX point of contact: @violasong (Victoria)

Figma designs (bottom right, labeled "new tab menu"): https://www.figma.com/file/NHu4cTmzfYgi3QJz5DIDWD/Fenix-Toolbar-Menus?node-id=0%3A1

## Acceptance criteria
**This work should be behind a feature flag**
* Report site issue item should be available on nightly and debug (this is how it is currently implemented)
* It should be placed right above the divider for "Add to home screen" 
<img width="353" alt="image" src="https://user-images.githubusercontent.com/43795363/109993409-e102d380-7cd1-11eb-864f-60d574492cdc.png">

violasong:
Sorry for the delay on this! 

I think the placement that makes the most sense is at the bottom of the "Find in page" section. I've updated the "full list with all possible conditional items" mockup in "Final design" to reflect this.

Issue #17872, opened 2021-02-05
[Feature] Add three-dot menu icons back


eliserichards:
Meta: #17796 
Followup to #17771 
UX point of contact: @violasong (Victoria)

Figma designs (bottom right, labeled "new tab menu"): https://www.figma.com/file/NHu4cTmzfYgi3QJz5DIDWD/Fenix-Toolbar-Menus?node-id=0%3A1

## Acceptance criteria

#17771 removed some icons on the three-dot menu, but the designs have changed and we are now keeping the icons. This issue is to add those icons back in. 

~Add the ability to remove toolbar menu icons in menu items. This should be reflected in the fenix home menu and the main three-dot menu.~
See files in AC:
*  `BrowserMenuImageText`
* `BrowserMenuHighlightableItem`
* `BrowserMenuImageSwitch`
* etc

**Old design** 
<img width="307" alt="image" src="https://user-images.githubusercontent.com/43795363/107089494-560ee600-67c4-11eb-9b4e-ccd0abf89c6b.png">

sheikh-azharuddin:
Waiting for this to finally switch to some other browser!

ekager:
@sheikh-azharuddin this is not an appropriate comment. This is your final warning. Next step is banning from participation. Please read up on [the CPG](https://www.mozilla.org/en-US/about/governance/policies/participation/) for what behavior will not be tolerated in this repo  and feel free to file specific issues with pain points you are having if you can do so respectfully.

mcarare:
@eliserichards Is this issue still valid?

sheikh-azharuddin:
The bookmark icon on tabs 3 dot menu is not matching with home screen 3 dot menu bookmark icon

Issue #17798, opened 2021-02-03
[Feature] Three-dot menu Sync sign-in item


eliserichards:
Meta: #17796
Followup to #17771 

~Depends on synced tabs in the tabs tray in issue #14969 and #14117~

UX point of contact: @violasong (Victoria)

Figma designs (bottom right, labeled "Final design"): https://www.figma.com/file/NHu4cTmzfYgi3QJz5DIDWD/Fenix-Toolbar-Menus?node-id=0%3A1

## Acceptance criteria
* In the main three-dot menu **AND** the new tab three-dot menu show "Sign in to sync" when a user isn't signed in
* When a user has signed in to sync, they should see their username ("Signed in as _username_")
   * This should be ellipsed, see example in designs

### UX feedback needed:
Open question: what icon is shown next to this item? (need to get updated designs from @violasong)

**Main menu**
<img width="341" alt="image" src="https://user-images.githubusercontent.com/43795363/106791366-7307b580-661a-11eb-8ad9-aa3743d77a67.png"> <img width="349" alt="image" src="https://user-images.githubusercontent.com/43795363/106790813-adbd1e00-6619-11eb-8737-fe05df08e538.png">

**New tab menu**
<img width="354" alt="image" src="https://user-images.githubusercontent.com/43795363/106791502-9cc0dc80-661a-11eb-8236-8c485e2143c3.png">

topotropic:
![Group 1](https://user-images.githubusercontent.com/353960/111535387-55755200-8769-11eb-829c-3d74f3b811fe.jpg)

I believe we have the icons in the code base but let me know in case we need them. Thanks!

kglazko:
Is it ok if I assign myself to this one and work on it @eliserichards ?

gabrielluong:
In case this is helpful, [SyncPreferenceView](https://github.com/mozilla-mobile/fenix/blob/a31ce9bf671bc6c388f3bf4b94d95e55bb803388/app/src/main/java/org/mozilla/fenix/settings/SyncPreferenceView.kt) which is responsible for "Sign in to sync" in "Logins" will get you really far and fast.

eliserichards:
Go for it @kglazko! 🚀

kglazko:
Github isn't automatically linking to the WIP PR for some reason, can see progress here https://github.com/mozilla-mobile/fenix/pull/18609

jonalmeida:
[...]

PR #18372, opened 2021-03-08
For #17872, #18032: Remove disabled color for three dot menu icons and add report site issue

eliserichards:
For #17872 (icons), for #18032 (report site issue)

This makes the three dot menu icons visible again (per design decision). This also adds the "Report site issue" item in nightly/debug.

Confirmed colors in: 
- [x] light mode, normal browsing (see below)
- [x] light mode, private browsing (no screenshots)
- [x] dark mode, normal browsing (see below)
- [x] dark mode, private browsing (no screenshots)

### Screenshots

 

Open in app & reader view appearance: 

Report site issue: (FYI there's an open bug for making the lightbulb icon look better.. I'll link that here)

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
@@ -407,10 +407,10 @@ class DefaultToolbarMenu(
             onItemTapped.invoke(ToolbarMenu.Item.AddonsManager)
         }
 
-        val syncSignIn = BrowserMenuImageText(
-            context.getString(R.string.sync_sign_in),
-            R.drawable.ic_synced_tabs,
-            primaryTextColor()
+        val syncedTabs = BrowserMenuImageText(
+            label = context.getString(R.string.synced_tabs),
+            imageResource = R.drawable.ic_synced_tabs,
+            iconTintColorResource = primaryTextColor()
         ) {
             onItemTapped.invoke(ToolbarMenu.Item.SyncedTabs)
         }
@@ -439,19 +439,6 @@ class DefaultToolbarMenu(
             iconTintColorResource = primaryTextColor()
         ) {
             onItemTapped.invoke(ToolbarMenu.Item.CustomizeReaderView)
-        }
-
-        val openInApp = BrowserMenuHighlightableItem(
-            label = context.getString(R.string.browser_menu_open_app_link),
-            startImageResource = R.drawable.ic_open_in_app,
-            iconTintColorResource = primaryTextColor(),
-            highlight = BrowserMenuHighlight.LowPriority(
-                label = context.getString(R.string.browser_menu_open_app_link),
-                notificationTint = getColor(context, R.color.whats_new_notification_color)
-            ),
-            isHighlighted = { !context.settings().openInAppOpened }
-        ) {
-            onItemTapped.invoke(ToolbarMenu.Item.OpenInApp)
         }
 
         val reportSiteIssuePlaceholder = WebExtensionPlaceholderMenuItem(
@@ -508,12 +495,11 @@ class DefaultToolbarMenu(
                 historyItem,
                 downloadsItem,
                 extensionsItem,
-                syncSignIn,
+                syncedTabs,
                 BrowserMenuDivider(),
                 findInPageItem,
                 desktopSiteItem,
                 customizeReaderView.apply { visible = ::shouldShowReaderViewCustomization },
-                openInApp.apply { visible = ::shouldShowOpenInApp },
                 reportSiteIssuePlaceholder,
                 BrowserMenuDivider(),
                 addToHomeScreenItem.apply { visible = ::canAddToHomescreen },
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
```diff
@@ -370,56 +369,56 @@ class DefaultToolbarMenu(
     private val newCoreMenuItems by lazy {
         val newTabItem = BrowserMenuImageText(
             context.getString(R.string.library_new_tab),
-            R.drawable.ic_bookmark_filled,
```
jonalmeida, line 373: Good catch!

app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
```diff
@@ -186,7 +185,7 @@ class DefaultToolbarMenu(
         appLink(session.content.url).hasExternalApp()
     } ?: false
 
-    private fun shouldShowReaderAppearance(): Boolean = selectedSession?.let {
+    private fun shouldShowReaderViewCustomization(): Boolean = selectedSession?.let {
```
jonalmeida, line 188: nit: this function renaming seems related to the menu refactoring instead of to this PR (i.e. adding 3-dot menu icons). Let's try to keep unrelated changes separate.

app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
```diff
@@ -434,41 +433,59 @@ class DefaultToolbarMenu(
...
+        val customizeReaderView = BrowserMenuImageText(
             label = context.getString(R.string.browser_menu_customize_reader_view),
-            textColorResource = primaryTextColor()
+            imageResource = R.drawable.ic_readermode_appearance,
+            iconTintColorResource = primaryTextColor()
         ) {
             onItemTapped.invoke(ToolbarMenu.Item.CustomizeReaderView)
         }
 
+        val openInApp = BrowserMenuHighlightableItem(
```
jonalmeida, line 444: Similar nit as above: Why are we adding this here? Seems related to a different PR.

app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
```diff
@@ -370,56 +369,56 @@ class DefaultToolbarMenu(
...
             R.drawable.ic_addons_extensions,
-            disabledTextColor()
+            primaryTextColor()
         ) {
             onItemTapped.invoke(ToolbarMenu.Item.AddonsManager)
         }
 
-        val syncedTabsItem = BrowserMenuImageText(
-            context.getString(R.string.library_synced_tabs),
+        val syncSignIn = BrowserMenuImageText(
```
jonalmeida, line 410: Same nit as above: Let's leave this as part of the other PR. This is a half change as well because we need to update the `onItemTapped.invoke(ToolbarMenu.Item.SyncedTabs)` as well.

jonalmeida (changes requested):
Looks good; let's remove some changes that are not part of this PR before landing!

app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
eliserichards: The reader view PR was already merged, and this rename (from "Appearance" to "Cutomization") was missed. I would like to keep it in this PR so I don't have to open a separate one just for a rename

app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
eliserichards: I agree. The name change from synced tabs to sync sign in shouldn't be part of this card. Creating the invocation for ToolbarMenu.Item.SyncSignIn (or whatever it will be called) and changing that name will be part of https://github.com/mozilla-mobile/fenix/issues/17798

app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
eliserichards: It was taken out as a part of the original design, and the updated designs add it back in. I could open a separate PR but that seemed wasteful

jonalmeida: PRs are cheap. 🙂 

It's a good rule of thumb to keep isolated work separately so there is context when someone needs to debug why a change was added to a separate issue. Smaller PRs are also easier for the reviewer.

How about separating this into a different commit so it doesn't fall under "Remove disabled color for icons"?

eliserichards: Sure, I'll tag you ;)

jonalmeida (approved):
LGTM!

PR #18379, opened 2021-03-08
bug 1696457 - restrict signing index

escapewindow:
I'm not 100% sure whether this is tuned perfectly, but this should reduce the number of bogus nightly-simulation indexes to just the nightly graphs, per https://bugzilla.mozilla.org/show_bug.cgi?id=1696457 .


jmaher (approved):
I have had odd results in the past when nested by-* stuff, but if this is working well, it is logical.

gabrielluong (approved):

PR #18397, opened 2021-03-09
For #17804: Use enterToImmersiveMode from support-ktx component

jonalmeida:
This should land after the next AC nightly is out which adds Android 11 support for window insets.

1. click on Show All Checks,


mcarare (approved):
👍

Issue #17800, opened 2021-02-03
[Feature] Request desktop site from new tab menu


eliserichards:
Meta: #17796
Followup to #17770 
UX point of contact: @violasong (Victoria)

Figma designs (bottom right, labeled "Final design"): https://www.figma.com/file/NHu4cTmzfYgi3QJz5DIDWD/Fenix-Toolbar-Menus?node-id=0%3A1

## Acceptance criteria
* User flow: Open a new tab, open the three-dot menu, select "Request desktop site"
* This toggle already exists on the main three-dot menu (when a user is already on a site)
* If a user toggles this to ON, when they enter a URL/search they will automatically be brought to the desktop site

---
<img width="353" alt="image" src="https://user-images.githubusercontent.com/43795363/106794846-fc20eb80-661e-11eb-9b9d-26796d37d9fb.png">

mcarare:
Question: Will the state of the toggle be remembered between multiple new tabs? Or it is used just for the next tab navigation?

violasong:
It should be used just for the navigation within that tab. Additional new tabs should go back to the default of mobile view.

mcarare:
So it will be just to load the desktop version without having to load the mobile one? (The same number of steps as requesting desktop version from a tab).

I thought it was related to a feature request: https://github.com/mozilla-mobile/fenix/issues/7039.

violasong:
I think request #7039 will be important to solve later, but for the purposes of this bug, yes, it's just to load the desktop site without having to load the mobile site. It should work the same way as it currently does, where it doesn't reset back to off for that tab unless the user chooses to toggle it off. Yes, it should work for top sites.

CC: @kbrosnan may be able to help with a more detailed info around this feature, since I believe he gave me the idea originally :).

mcarare:
Waiting for feedback from @kbrosnan

kbrosnan:
Part of what breaks this design is that we have not resolved if the homescreen is a tab or not. Previously it was not but that has become less clear over time. My understanding is resolving the tab status of the homescreen is out of scope for this issue. 

* When the user toggles the request desktop site (RDS) the next load should be as a desktop site
* If the user clicks on links in the page that lead to different domains RDS should keep its state
* If the user loads a URL or preforms a search from the address bar then RDS should keep its state
* Having that work with top sites seems reasonable though if that adds a lot of complexity we could limit this feature to pages loaded via the search/address bar 
[...]

Issue #17770, opened 2021-02-02
[Feature] Reorder New Tab three-dot menu and add navigation


eliserichards:
Meta: #17796
Followup to #17771 

UX point of contact: @violasong (Victoria)

Figma designs (bottom right, labeled "Final design"): https://www.figma.com/file/NHu4cTmzfYgi3QJz5DIDWD/Fenix-Toolbar-Menus?node-id=0%3A1

## Acceptance criteria
This should be the `HomeMenu.kt` file
* **Icons are not being removed. We are waiting on designs to reflect this**
* Reorder the menu items to match designs. From top to bottom:
   * Bookmarks
   * History
   * Downloads
   * Extensions (this is a label change for addons. Followup #17799)
   * Sign in to sync (followup: #17798)
   * Reconnect to Sync
   * Desktop site + toggle (followup: #17800)
   * What's new
   * Help
   * Settings
   * Quit
* Add dividing lines to match designs
* This implementation should add a sticky navigation section to the bottom of the new tab three-dot menu. (Main menu navigation added in #17802)
   * The user should be able to navigate _back_ to the new tab page, then open this three-dot menu to navigate _forwards_ again
   * The back button should be unclickable/grayed out when you are on the new tab page
   * The location of this should match the Top/Bottom orientation of the toolbar setting

---
**Current** (left) vs **new design** (right)
<img width="300" alt="image" src="https://user-images.githubusercontent.com/43795363/106786862-a5aeaf80-6614-11eb-906b-e2d2639a3466.png">. <img width="302" alt="image" src="https://user-images.githubusercontent.com/43795363/106664266-70e61e00-656a-11eb-93cc-36868b3ccd7d.png">

kglazko:
Adding a ticket for UI tests update since this could be a breaking change https://github.com/mozilla-mobile/fenix/issues/17840

kglazko:
@eliserichards @violasong What should we be doing for toggle to desktop in this menu? I only realized while coding it that in the previous PR where this menu is open when a tab is open, we use this `private val selectedSession: TabSessionState? get() = store.state.selectedTab` which makes me wonder what to do when there's no selected tab.

eliserichards:
@kglazko I broke out the implementation of desktop site into another story, if you want to just add the item in this PR: https://github.com/mozilla-mobile/fenix/issues/17800

AndRo-Marian:
Why I need the What's New, Help and Settings to be to bottom?

PR #18427, opened 2021-03-10
For #17770: New tab three-dot menu reorder

eliserichards:
For #17770 

Adds feature flag to hide Request Desktop Site until #17800 is implemented.

## Screenshots
Light mode, dark mode
 

With Quit item

### Pull Request checklist

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/home/HomeMenu.kt
```diff
@@ -211,9 +249,155 @@ class HomeMenu(
...
+            requestDesktopSiteItem,
+            BrowserMenuDivider(),
+            whatsNewItem,
+            helpItem,
+            settingsItem,
+            if (shouldUseBottomToolbar) BrowserMenuDivider() else null,
+            if (shouldUseBottomToolbar) menuToolbar else null
+        ).also { items ->
+            items.getHighlight()?.let { onHighlightPresent(it) }
+        }
```
gabrielluong, line 381: It made sense to me when we separated out to old vs new MenuItems in the `DefaultToolbarMenu` before because the menu items appearance were different, but since this is only reordering and I am assuming we aren't changing the icons. Should we just keep one instance of this initialization function and just have an if statement block for the different ordering to avoid duplicating all the menu item variables.

app/src/main/java/org/mozilla/fenix/home/HomeMenu.kt
```diff
@@ -86,7 +94,35 @@ class HomeMenu(
...
+            primaryImageTintResource = primaryTextColor,
+            isInPrimaryState = { true },
+            secondaryImageTintResource = ThemeManager.resolveAttribute(R.attr.disabled, context),
+            disableInSecondaryState = true,
+            longClickListener = { onItemTapped.invoke(Item.Forward(viewHistory = true)) }
+        ) {
+            onItemTapped.invoke(Item.Forward(viewHistory = false))
+        }
+
+        BrowserMenuItemToolbar(listOf(back, forward))
```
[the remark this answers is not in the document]
eliserichards: It is possible to go _forward_, technically. Open a new tab, navigate to a page, go back (brings you to the new tab page), and then go forwards again. I think the thought here is that if we want the forwards button it makes UX sense to have both forward and backwards, even though backwards will always be disabled

app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
```diff
@@ -66,10 +65,15 @@ class DefaultToolbarMenu(
 
     private var isCurrentUrlBookmarked = false
     private var isBookmarkedJob: Job? = null
-    private val isTopToolbarSelected = shouldReverseItems
+
+    private val shouldUseBottomToolbar = context.settings().shouldUseBottomToolbar
```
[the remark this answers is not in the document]
eliserichards: I would like to keep `shouldUseBottomToolbar` since that's the name of the `setting`

app/src/main/java/org/mozilla/fenix/home/HomeMenu.kt
```diff
@@ -211,9 +249,155 @@ class HomeMenu(
...
+
     init {
+        val menuItems = if (FeatureFlags.toolbarMenuFeature) {
+            newCoreMenuItems()
+        } else {
+            oldCoreMenuItems
+        }
+
+        val menuItemsWithReconnectItem = if (FeatureFlags.toolbarMenuFeature) {
+            newCoreMenuItems() // re-evaluated with reconnect item viewable
```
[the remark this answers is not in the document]
eliserichards: We need to re-evaluate the items here because we need to re-do the visibility of reconnect:

```kotlin
val accountAuthItem =
            if (context.components.backgroundServices.accountManagerAvailableQueue.isReady()
                && context.components.backgroundServices.accountManager.accountNeedsReauth()) {
                reconnectToSyncItem
            } else {
                null
            }
```

The new designs have this `accountAuthItem` in the middle of the list, so we can't just append it to the end like we used to. I'm taking this approach instead of trying to splice the list apart and add the auth item in the middle

app/src/main/java/org/mozilla/fenix/home/HomeMenu.kt
eliserichards: We're reordering **and** adding the navigation and request desktop item. 

I think my approach will be easier when we are going back and deleting the old items: we won't need to decipher any of the logic, just remove the whole block of `oldCoreItems`.

app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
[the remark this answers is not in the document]
eliserichards: Just for consistency across the app

app/src/main/java/org/mozilla/fenix/home/HomeMenu.kt
```diff
@@ -224,17 +410,19 @@ class HomeMenu(
             context.components.backgroundServices.accountManager.register(object : AccountObserver {
                 override fun onAuthenticationProblems() {
                     lifecycleOwner.lifecycleScope.launch(Dispatchers.Main) {
-                        onMenuBuilderChanged(BrowserMenuBuilder(
-                            listOf(reconnectToSyncItem) + coreMenuItems
-                        ))
+                        onMenuBuilderChanged(
```
gabrielluong, line 413: Could we define `menuItemsWithReconnectItem` within this block instead since it's only used here? Alternatively I could just imagine we can inline the if statement block in `BrowserMenuBuilder`.

gabrielluong (approved):

app/src/main/java/org/mozilla/fenix/home/HomeMenu.kt
```diff
@@ -38,25 +43,29 @@ class HomeMenu(
...
-        object AddonsManager : Item()
         object Settings : Item()
-        object SyncedTabs : Item()
-        object History : Item()
-        object Bookmarks : Item()
-        object Downloads : Item()
-        object Quit : Item()
-        object Sync : Item()
+        object Quit : Item() // waiting
+        object ReconnectSync : Item() // waiting
```
gabrielluong, line 59: I think we should just remove the `// waiting` or file sufficient issues to mention that we should wait these comments. At a glance, it's unclear what we are waiting for since I can see `ReconnectSync` is being used, but maybe it's just `Quit`.

gabrielluong (approved):

app/src/main/java/org/mozilla/fenix/home/HomeMenu.kt
eliserichards: Oh shoot, good catch!

eliserichards:
Failing tests: 
* testPageReloadAfterNetworkInterrupted
* mainMenuShareButtonTest
* mainMenuRefreshButtonTest

Issue #17822, opened 2021-02-04
Create a layout for the tabs tray


jonalmeida:
See investigation comment as well: https://github.com/mozilla-mobile/fenix/issues/14969#issuecomment-703350265

We can nest the original normal tabs layout in a ViewPager2 so we don't need to re-write everything at once.

UX designs that may need some updates, but gives us an idea of how we want to structure things: https://share.goabstract.com/c6bd0c2f-8384-44f3-ba59-4946d8cea469?collectionLayerId=ec285dbe-88bf-41cc-a531-49aad233b5ae&mode=design

### Layout 

A basic layout of how the views will be structured. 

This should be straight-forward: a view-pager with individual layout containers in them (mostly RecyclerViews).

0 : Normal tab indicator
P : Private tab indicator
S : Synced Tabs indicator

```
+---+  +---+  +---+           + 
| 0 |  | P |  | S |           +  <---------+ BrowserMenu2
+---+  +---+  +---+           + 

+------------------------------+
|                              |
|                              |
|                              |
|                              |
|                              |
|                              |
|                              |
|                              |
|         ViewPager2           |
|                              |
|                              |
|                              |
|                              |
|                              |
|                              |
|                              |
|                              |
+---------------+--------------+
                ^
                |
+---------------+-------------------------------------------+
+                                                           +

+-----------------+  +-----------------+  +-----------------+
|                 |  |                 |  |                 |
|                 |  |                 |  |                 |
|                 |  |                 |  |                 |
|                 |  |                 |  |                 |
|                 |  |                 |  |                 |
|  Normal tabs    |  |  Private tabs   |  |  Synced tabs    |
|                 |  |                 |  |                 |
|(BrowserTabsList)|  |(BrowserTabsList)|  |(SyncedTabLayout)|
|                 |  |                 |  |                 |
|                 |  |                 |  |                 |
|                 |  |                 |  |                 |
|                 |  |                 |  |                 |
+-----------------+  +-----------------+  +-----------------+

PR #18440, opened 2021-03-11
Issue #17822: Create a tabs tray layout and fragment

jonalmeida:
The tray should look the same with less function while we refactor for now. 🙃 

You can enabled the refactored tray in Secret Settings.

https://user-images.githubusercontent.com/1370580/110819069-d4472800-825b-11eb-869c-b4839aabb8ff.mp4

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/tabstray/TrayPagerAdapter.kt
+++ b/app/src/main/java/org/mozilla/fenix/tabstray/TrayPagerAdapter.kt
@@ -1,0 +1,59 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.tabstray
+
+import android.content.Context
+import android.view.LayoutInflater
+import android.view.ViewGroup
+import androidx.recyclerview.widget.RecyclerView
+import org.mozilla.fenix.tabstray.BrowserTabViewHolder.Companion.LAYOUT_ID_NORMAL_TAB
+import org.mozilla.fenix.tabstray.BrowserTabViewHolder.Companion.LAYOUT_ID_PRIVATE_TAB
+import org.mozilla.fenix.tabtray.FenixTabsAdapter
+
+class TrayPagerAdapter(
+    context: Context,
+    val interactor: TabsTrayInteractor
+) : RecyclerView.Adapter<TrayViewHolder>() {
+
+    private val normalAdapter by lazy { FenixTabsAdapter(context) }
+    private val privateAdapter by lazy { FenixTabsAdapter(context) }
+
+    override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): TrayViewHolder {
+        val itemView = LayoutInflater.from(parent.context).inflate(viewType, parent, false)
+
+        return when (viewType) {
+            LAYOUT_ID_NORMAL_TAB -> BrowserTabViewHolder(itemView, interactor)
+            LAYOUT_ID_PRIVATE_TAB -> BrowserTabViewHolder(itemView, interactor)
+            else -> throw IllegalStateException("Unknown viewType.")
+        }
+    }
+
+    override fun onBindViewHolder(viewHolder: TrayViewHolder, position: Int) {
+        val adapter = when (position) {
+            POSITION_NORMAL_TABS -> normalAdapter
+            POSITION_PRIVATE_TABS -> privateAdapter
+            else -> throw IllegalStateException("View type does not exist.")
+        }
+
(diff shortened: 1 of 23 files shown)


app/src/main/java/org/mozilla/fenix/tabstray/TrayViewHolders.kt
```diff
@@ -0,0 +1,41 @@
...
+
+class BrowserTabViewHolder(
+    containerView: View,
+    interactor: TabsTrayInteractor
+) : TrayViewHolder(containerView) {
+
+    private val trayList: BaseBrowserTrayList = itemView.findViewById(R.id.tray_list_item)
+
+    init {
+        trayList.interactor = interactor
```
jonalmeida, line 29: @Amejia481 this is how we're passing the interactor to the viewholders. Keen if you have other suggestions how to do this. 🙂

jonalmeida:
>   /home/runner/work/fenix/fenix/app/src/main/res/layout/component_tabstray2.xml:26: Error: Flatten the view hierarchy by using one ConstraintLayout, if possible. If the alternative is several nested ViewGroup, it may not help performance and this may be worth suppressing. [MozMultipleConstraintLayouts]
      <androidx.constraintlayout.widget.ConstraintLayout
       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

Will suppress for now and re-assess when we refactor the layout as well.

app/src/main/java/org/mozilla/fenix/tabstray/TrayPagerAdapter.kt
```diff
@@ -0,0 +1,59 @@
...
+
+import android.content.Context
+import android.view.LayoutInflater
+import android.view.ViewGroup
+import androidx.recyclerview.widget.RecyclerView
+import org.mozilla.fenix.tabstray.BrowserTabViewHolder.Companion.LAYOUT_ID_NORMAL_TAB
+import org.mozilla.fenix.tabstray.BrowserTabViewHolder.Companion.LAYOUT_ID_PRIVATE_TAB
+import org.mozilla.fenix.tabtray.FenixTabsAdapter
+
+class TrayPagerAdapter(
```
Amejia481, line 15: Should we be add tests for this? :)

app/src/main/java/org/mozilla/fenix/tabstray/TrayViewHolders.kt
```diff
@@ -0,0 +1,41 @@
...
+
+    abstract fun bind(adapter: RecyclerView.Adapter<out RecyclerView.ViewHolder>)
+}
+
+class BrowserTabViewHolder(
+    containerView: View,
+    interactor: TabsTrayInteractor
+) : TrayViewHolder(containerView) {
+
+    private val trayList: BaseBrowserTrayList = itemView.findViewById(R.id.tray_list_item)
```
Amejia481, line 26: Is OK to do this here? Shouldn't it be provided via the constructor?

Amejia481 (commented):
Awesome work, I left some questions additionally could we increase the test coverage or this is going to be done in separated task.

app/src/main/java/org/mozilla/fenix/tabstray/TrayPagerAdapter.kt
jonalmeida: Yes! Tests are coming in a follow-up PR. Ran into some local dependency issue with Robolectric for now.

app/src/main/java/org/mozilla/fenix/tabstray/TrayViewHolders.kt
jonalmeida: I think it should be fine. We're using the base ViewHolder's `itemView` to retrieve the reference to the inner RecyclerView so that we can use it.

However, we can optimize this a bit more with synthetics instead of the old `findViewById` since the `TrayViewHolder` implements the `LayoutContainer` interface.

app/src/main/java/org/mozilla/fenix/tabstray/TrayPagerAdapter.kt
Amejia481: 👍🏽

app/src/main/java/org/mozilla/fenix/tabstray/TrayViewHolders.kt
Amejia481: 👍🏽

app/src/main/java/org/mozilla/fenix/tabstray/TrayViewHolders.kt
Amejia481: I think it's OK!

Amejia481 (approved):
LGTM!

Issue #15770, opened 2020-10-07
Investigate extractNativeLibs behavior change with new Android gradle plugin 3.6


MarcLeclair:
extractNativeLibs change in behavior: https://developer.android.com/studio/releases/gradle-plugin#3-6-0-behavior

We should 

1) Talk to the GV team  and acresky about the expected behavior for native libs ( compress vs not , etc...) 
2) Investigate depending on answer from #1
3) Set an explicit value so that we understand our reasoning.

MarcLeclair:
Waiting for some answers but by preliminary testing, I found that, using `extractNativeLibs` makes start up 50-100ms slower on the G5.

MarcLeclair:
After talking with the GV team, I was told that it should be fine. From the conversation I had, the following was said:

> 
not GeckoView but it may impact Fenix, although I think we would notice the increase in size if our libraries were suddenly not compressed anymore. Our `libxul.so` is 123MB uncompressed

Overall, I think this change is fine. Although, maybe set it explicitly to false as a follow up with a note as to why it's false.

PR #18442, opened 2021-03-11
Resolves #18381: Explicitly set android:extractNativeLibs to true in ApplicationManifest.xml

### LINKED ISSUE #18381: Explicitly specify extractNativeLibs attribute in ApplicationManifest.xml
The [documentation](https://developer.android.com/studio/releases/gradle-plugin#3-6-0-behavior) for version 3.6.0 of the Android Gradle Plugin states the following:

```
When you build your app, the plugin now sets extractNativeLibs to "false" by default.
```

For Android Gradle Plugin versions prior to 3.6.0, the default value was `true`.

Based on what I see in the [commit](https://github.com/mozilla-mobile/fenix/commit/aeb0205ea5f1e618ad75d5d82badb7baad986fac) that upgraded the plugin from 3.5.0 to 3.6.4, and thus triggering this change, no modifications were made to `AndroidManifest.xml` to account for this revised default value of `extractNativeLibs`.

Could the Fenix team please (re-)evaluate which value you would like `extractNativeLibs` to take, and explicitly set it in the manifest?

dblohm7:
Resolves #18381

The Android Gradle Plugin's default for android:extractNativeLibs changed from
true to false beginning with version 3.6.0. Based on GeckoView's needs, we
should ensure that this attribute is explicitly set to true

### Pull Request checklist

- [ ] **Tests**: No tests, just a manifest attribute flip
- [N/A] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not
- [N/A] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features. In addition, it includes a screenshot of a successful [accessibility scan](https://play.google.com/store/apps/details?id=com.google.android.apps.accessibility.auditor&hl=en_US) to ensure no new defects are added to the product.

1. click on Show All Checks,

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -25,6 +25,7 @@
     <application
         android:name=".FenixApplication"
         android:allowBackup="false"
+        android:extractNativeLibs="true"
         android:icon="@mipmap/ic_launcher"
         android:label="@string/app_name"
         android:roundIcon="@mipmap/ic_launcher_round"


pocmo (approved):
👍 

Will this improve performance of loading those libraries?

pocmo:
Added mcomella as another reviewer since it may affect startup?

https://developer.android.com/guide/topics/manifest/application-element#extractNativeLibs

dblohm7:
> Will this improve performance of loading those libraries?

It should be no change... AFAICT we weren't really ever packaging our APK in such a way that direct loading from the APK was possible (it requires storage of `.so` with no deflation, and also page-aligning the offsets of those stored `.so`). My concern is that I am seeing some suspicious crash reports that suggest that some devices might be trying to load straight from the APK anyway because of the defaults, hence my desire to be explicit about our expectations.

mcomella:
@MarcLeclair You previously investigated this in https://github.com/mozilla-mobile/fenix/issues/15770, right? What was the conclusion there? Is this an okay change?

MarcLeclair:
At first , I saw an increase in ~30ms for start up ( which I found odd since it was a recent change with the gradle plugin and before it was implicitly set to true). So I tested Fenix with 100 runs instead of 25 and found no problems.

MarcLeclair (approved):

mcomella:
This is blocking a top crasher so we should land it and evaluate the perf later.

CI is blocked for some reason: rerunning.

mcomella:
Not sure why CI isn't completing – I'm guessing this needs to be rebased. However, the change seems like it shouldn't break anything so I'm going to force merge.

PR #18456, opened 2021-03-12
Closes #18367 run SupportedAddonsChecker only for users with unsupported extensions

### LINKED ISSUE #18367: Run SupportedAddonsChecker only for users with unsupported extensions
`SupportedAddonsChecker` checks if add-ons that were disabled during migration have become available. This check is only needed for users who have unsupported extensions installed. Let's investigate skipping these checks for everyone else.

Amejia481:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
+++ b/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
@@ -10,6 +10,7 @@ import android.os.Build.VERSION.SDK_INT
 import android.os.StrictMode
 import android.util.Log.INFO
 import androidx.annotation.CallSuper
+import androidx.annotation.VisibleForTesting
 import androidx.appcompat.app.AppCompatDelegate
 import androidx.core.content.getSystemService
 import androidx.work.Configuration.Builder
@@ -23,7 +24,10 @@ import mozilla.appservices.Megazord
 import mozilla.components.browser.state.action.SystemAction
 import mozilla.components.browser.state.selector.selectedTab
 import mozilla.components.concept.base.crash.Breadcrumb
+import mozilla.components.concept.engine.webextension.WebExtension
+import mozilla.components.concept.engine.webextension.isUnsupported
 import mozilla.components.concept.push.PushProcessor
+import mozilla.components.feature.addons.migration.DefaultSupportedAddonsChecker
 import mozilla.components.feature.addons.update.GlobalAddonDependencyProvider
 import mozilla.components.lib.crash.CrashReporter
 import mozilla.components.service.glean.Glean
@@ -438,12 +442,27 @@ open class FenixApplication : LocaleAwareApplication(), Provider {
                 },
                 onExtensionsLoaded = { extensions ->
                     components.addonUpdater.registerForFutureUpdates(extensions)
-                    components.supportedAddonsChecker.registerForChecks()
+                    subscribeForNewAddonsIfNeeded(components.supportedAddonsChecker, extensions)
                 },
                 onUpdatePermissionRequest = components.addonUpdater::onUpdatePermissionRequest
             )
         } catch (e: UnsupportedOperationException) {
             Logger.error("Failed to initialize web extension support", e)
+        }
+    }
+
+    @VisibleForTesting
+    internal fun subscribeForNewAddonsIfNeeded(
+        checker: DefaultSupportedAddonsChecker,
+        installedExtensions: List<WebExtension>
+    ) {
+        val hasUnsupportedAddons = installedExtensions.any { it.isUnsupported() }
+        if (hasUnsupportedAddons) {
+            checker.registerForChecks()
+        } else {
+            // As checks are a persistent subscriptions, we have to make sure
+            // we remove any previous subscriptions.
+            checker.unregisterForChecks()
         }
     }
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -447,6 +449,23 @@ open class FenixApplication : LocaleAwareApplication(), Provider {
...
+        scope.launch {
+            val haveUnsupportedAddons =
+                components.addonManager.getAddons().any { it.isInstalled() && !it.isSupported() }
+            if (haveUnsupportedAddons) {
+                components.supportedAddonsChecker.registerForChecks()
+            } else {
+                // As checks are a persistent subscriptions, we have to me sure
+                // we remove any previous subscriptions.
+                components.supportedAddonsChecker.unregisterForChecks()
+            }
```
Amejia481, line 465: @csadilek before I adding tests I just want to confirm with you that changing the checks from being **sync** to be **async** is not a problem. The change needs to made to be able to query  add-ons from the add-on manager.

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -447,6 +449,23 @@ open class FenixApplication : LocaleAwareApplication(), Provider {
         }
     }
 
+    @VisibleForTesting
+    internal fun subscribeForNewAddonsIfNeeded(
+        scope: CoroutineScope = CoroutineScope(Dispatchers.IO)
+    ) {
+        scope.launch {
+            val haveUnsupportedAddons =
+                components.addonManager.getAddons().any { it.isInstalled() && !it.isSupported() }
```
csadilek, line 458: This condition should be ` { it.isInstalled() && it.isDisabledAsUnsupported() }` because we only want to run if the user as add-ons that were disabled during migration. The supported check would also be false for Nightly overrides.

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
csadilek: The switch to sync/async would not be a problem, but the unconditional call to `getAddons` is. Our motivation for doing this is to reduce the number of network requests. With this logic as is we would still call getAddons on every startup, potentially hitting the network (cache is only 48 hours). So, we need to come up with something better here. Either remember that we already made the decision and not check again for that user or better: Let's do this in `onExtensionsLoaded` above and check against the extension metadata for "disabled as unsupported". This way we won't ever hit the AMO endpoint for this check.

csadilek (changes requested):
We need a slightly different but easier approach here. Comments below :).

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
Amejia481: Ok, then just checking if there are extensions that were disabled during migration will be enough

`val haveUnsupportedAddons = extensions.any { it.isUnsupported() }`

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -447,6 +451,21 @@ open class FenixApplication : LocaleAwareApplication(), Provider {
...
+    ) {
+        val haveUnsupportedAddons = installedExtensions.any { it.isUnsupported() }
+        if (haveUnsupportedAddons) {
+            checker.registerForChecks()
+        } else {
+            // As checks are a persistent subscriptions, we have to make sure
+            // we remove any previous subscriptions.
+            checker.unregisterForChecks()
+        }
+    }
```
Amejia481, line 467: @csadilek I update the PR to just check using  [WebExtension.isUnsupported()](https://github.com/mozilla-mobile/android-components/blob/5aad028635f3b0f28f90360ee814bd0b2660fc32/components/concept/engine/src/main/java/mozilla/components/concept/engine/webextension/WebExtension.kt#L473)

app/src/test/java/org/mozilla/fenix/FenixApplicationTest.kt
```diff
@@ -34,4 +41,32 @@ class FenixApplicationTest {
...
+        every { unSupportedExtension.getMetadata() } returns metadata
+        every { metadata.disabledFlags } returns DisabledFlags.select(DisabledFlags.APP_SUPPORT)
+
+        application.subscribeForNewAddonsIfNeeded(checker, listOf(unSupportedExtension))
+
+        verify { checker.registerForChecks() }
+    }
+
+    @Test
+    fun `GIVEN there are none unsupported addons installed WHEN subscribing for new add-ons checks THEN unregister for checks`() {
```
csadilek, line 60: nit: no unsupported...

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -447,6 +451,21 @@ open class FenixApplication : LocaleAwareApplication(), Provider {
         }
     }
 
+    @VisibleForTesting
+    internal fun subscribeForNewAddonsIfNeeded(
+        checker: DefaultSupportedAddonsChecker,
+        installedExtensions: List<WebExtension>
+    ) {
+        val haveUnsupportedAddons = installedExtensions.any { it.isUnsupported() }
```
csadilek, line 459: nit: hasUnsupportedAddons may be better.

csadilek (approved):
👍 Yes, this looks good now! Wound recommend landing on Monday so we're around.

Amejia481:
> 👍 Yes, this looks good now! Wound recommend landing on Monday so we're around.

Sure, lets land it on Monday, have a nice weekend 👍🏽

PR #18483, opened 2021-03-15
For #14005: fix strict TP tests loading timeout

sv-ohorvath:
Ran tests on Firebase, all passed: https://github.com/mozilla-mobile/fenix/runs/2095917553


isabelrios (approved):
LGTM, thanks for fixing these tests!

PR #18529, opened 2021-03-17
For #17644: Always update add-on's telemetry values before sending metric ping

### LINKED ISSUE #17644: [Telemetry] Web extensions
Follow up to https://github.com/mozilla-mobile/fenix/issues/6174
Related to #16832

## Current implementation
We are currently evaluating metrics on startup and receiving a list of enabled addons and installed addons. These are being used to evaluate these metrics:
* `addons.has_enabled_addon`
* `addons.has_installed_addons`

These metrics do not answer all of the product questions that were originally proposed (see #6174).

## Description & Product Manager / Data Scientist User Story
* Does product still care about the originally proposed questions? Should we re-evaluate these and implement new probes?
* Do we still care about the list of addons for installed & enabled?

## What questions will you answer with this data?
TBD based on above questions.

## Acceptance Criteria
- [ ] ENG files a [DS JIRA](https://jira.mozilla.com/projects/DO/issues/DO-228?filter=allopenissues) request outlining their methodology.
- [ ] DS sign off on instrumentation methodology addressing product questions.
- [ ] Event pings can be queried via re:dash
- [ ] Event pings can be queried via amplitude
- [ ] We are sending telemetry events for the actions listed in the requirements
- [ ] We have documented the telemetry
- [ ] We have asked a data steward to [review](https://github.com/mozilla/data-review/blob/master/request.md) the telemetry

rocketsroger:
Fixes #17644 

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt
@@ -69,6 +69,7 @@ import org.mozilla.fenix.components.toolbar.ToolbarPosition
 import org.mozilla.fenix.ext.components
 import org.mozilla.fenix.ext.settings
 import org.mozilla.fenix.utils.BrowsersCache
+import org.mozilla.fenix.utils.Settings
 
 private class EventWrapper<T : Enum<T>>(
     private val recorder: ((Map<T, String>?) -> Unit),
@@ -859,11 +860,20 @@ class GleanMetricsService(
 
             // setStartupMetrics is not a fast function. It does not need to be done before we can consider
             // ourselves initialized. So, let's do it, well, later.
-            setStartupMetrics()
+            setStartupMetrics(context.settings())
         }
     }
 
-    internal fun setStartupMetrics() {
+    /**
+     * This function is called before the metrics ping is sent.  Part of this function depends on
+     * shared preferences to be updated so the correct value is sent with the metrics ping.
+     *
+     * The reason we're using shared preferences to track some of these values is due to the
+     * limitations of the metrics ping.  Events are only sent in a metrics ping if the user have made
+     * changes between each ping.  However, in some cases we want current values to be sent even if
+     * the user have not changed anything between pings.
+     */
+    internal fun setStartupMetrics(settings: Settings) {
         setPreferenceMetrics()
         with(Metrics) {
             defaultBrowser.set(browsersCache.all(context).isDefaultBrowser)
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -937,6 +937,26 @@ class Settings(private val appContext: Context) : PreferencesHolder {
...
+
+    var enabledAddonsCount by intPreference(
+        appContext.getPreferenceKey(R.string.pref_key_enabled_addons_count),
+        0
+    )
+
+    var enabledAddonsList by stringPreference(
+        appContext.getPreferenceKey(R.string.pref_key_enabled_addons_list),
+        default = ""
+    )
```
gabrielluong, line 958: Let's start adding some kDoc describing the variables we have in `Settings`. Something to describe what they hold in the preference and what they are used for.

app/src/main/res/values/preference_keys.xml
```diff
@@ -217,6 +217,11 @@
 
     <string name="pref_key_open_tabs_count" translatable="false">pref_key_open_tabs_count</string>
 
+    <string name="pref_key_installed_addons_count" translatable="false">pref_key_installed_addons_count</string>
+    <string name="pref_key_installed_addons_list" translatable="false">pref_key_installed_addons_list</string>
+    <string name="pref_key_enabled_addons_count" translatable="false">pref_key_enabled_addons_count</string>
+    <string name="pref_key_enabled_addons_list" translatable="false">pref_key_enabled_addons_list</string>
```
gabrielluong, line 223: It would be nice to be able to group all these telemetry count/list pref key under a particular header. I find myself also adding comments to pref keys if I find them to be helpful.

app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt
```diff
@@ -859,11 +860,11 @@ class GleanMetricsService(
 
             // setStartupMetrics is not a fast function. It does not need to be done before we can consider
             // ourselves initialized. So, let's do it, well, later.
-            setStartupMetrics()
+            setStartupMetrics(context.settings())
         }
     }
 
-    internal fun setStartupMetrics() {
+    internal fun setStartupMetrics(settings: Settings) {
```
gabrielluong, line 867: I am curious if it would've made any difference if we just set `val settings = context.settings()` instead of passing `settings` or is it must always `settings` as a parameter and if there was any particular reason that you found from testing.

gabrielluong (approved):

app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt
rocketsroger: I needed to pass this in for testing.  Otherwise there's not good way to use a mock.

app/src/main/res/values/preference_keys.xml
rocketsroger: Sure, I'll add a comment above this group

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
rocketsroger: Sure, I'll add them.  Thanks,

app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt
gabrielluong, line 867: I think we might also want to add some docs for this function for the timing being until we consider writing telemetry documentations. In particular, I think it would be important to note why the previous approach `Addons.hasInstalledAddons.set` doesn't work and why we need to set values to a settings variable and then set telemetry like `Addons.hasInstalledAddons.set` (metric pings iirc) here.

rocketsroger: Sure, I'll write an explanation on why this function depends on so many shared preferences.  It won't be specific to this particular PR though.  I'll try to be general.

app/src/main/java/org/mozilla/fenix/components/metrics/MetricController.kt
```diff
@@ -213,16 +216,16 @@ internal class ReleaseMetricController(
...
                     Leanplum.setUserAttributes(mapOf("installed_addons" to installedAddons.size))
                 }
             }
 
             metadata?.get("enabled")?.let { enabledAddons ->
                 if (enabledAddons is List<*>) {
-                    Addons.enabledAddons.set(enabledAddons.map { it.toString() })
-                    Addons.hasEnabledAddons.set(enabledAddons.size > 0)
+                    settings.enabledAddonsCount = enabledAddons.size
+                    settings.enabledAddonsList = enabledAddons.joinToString()
```
willdurand, line 228: This change breaks some AMO stats as well as some Redash dashboards because the default value when joining items is `, ` (with a trailing space). The ETL stuff expects a comma-separated list without spaces (we don't trim items after `split`).

rocketsroger: Issue created: https://github.com/mozilla-mobile/fenix/issues/19289

Issue #18533, opened 2021-03-18
[crash] latest nightly crashing whenever opening tab view page


sheikh-azharuddin:
[comment]: # (Please do your best to search for duplicate issues before filing a new issue so we can keep our issue board clean)
[comment]: # (Every issue should have the exact bug and steps to reproduce described in it. Please do not file feedback list tickets as it is difficult to parse them and address their individual points)
[comment]: # (Read https://github.com/mozilla-mobile/fenix#i-want-to-file-an-issue for more information)

## Steps to reproduce
Open any website..click tab button
Firefox is crashing

### Expected behavior
It should show the tabs

### Actual behavior
It's crashing

### Device information

* Device vendor / model and Android version: ? Rog3, A10
* Firefox for Android version: ? (go to Settings -> About Firefox)Latest nightly 17mar

a24294e3-a22f-470b-9cd3-8ecc2bc83578
java.lang.ClassCastException: androidx.recyclerview.widget.LinearLayoutManager cannot be cast to androidx.recyclerview.widget.GridLayoutManager
----
java.lang.ClassCastException: androidx.recyclerview.widget.LinearLayoutManager cannot be cast to androidx.recyclerview.widget.GridLayoutManager
	at org.mozilla.fenix.tabtray.FenixTabsAdapter.onBindViewHolder(FenixTabsAdapter.kt:15)
	at org.mozilla.fenix.tabtray.FenixTabsAdapter.onBindViewHolder(FenixTabsAdapter.kt:5)
	at androidx.recyclerview.widget.RecyclerView$Adapter.bindViewHolder(RecyclerView.java:8)
	at androidx.recyclerview.widget.RecyclerView$Recycler.tryGetViewHolderForPositionByDeadline(RecyclerView.java:150)
	at androidx.recyclerview.widget.LinearLayoutManager$LayoutState.next(LinearLayoutManager.java:9)
	at androidx.recyclerview.widget.LinearLayoutManager.layoutChunk(LinearLayoutManager.java:1)
	at androidx.recyclerview.widget.LinearLayoutManager.fill(LinearLayoutManager.java:12)
	at androidx.recyclerview.widget.LinearLayoutManager.onLayoutChildren(LinearLayoutManager.java:121)
	at androidx.recyclerview.widget.RecyclerView.dispatchLayoutStep2(RecyclerView.java:12)
	at androidx.recyclerview.widget.RecyclerView.dispatchLayout(RecyclerView.java:15)
[...]

PR #18536, opened 2021-03-18
Enable new tabs tray only for debug builds

jonalmeida:
To avoid bug reports like https://github.com/mozilla-mobile/fenix/issues/18533 let's enable this only for debug builds.

1. click on Show All Checks,


mcarare (approved):

jonalmeida:
Restarting build; intermittent network connection resets..

Issue #18538, opened 2021-03-18
Refactor FenixTabsAdapter and TabViewHolder


jonalmeida:
This should shorten the class done for now and add these parts separately:
 - We can remove some of the specific multi-select code in there now - #18443
 - TabViewHolder can be separated into a List/Grid view holder. This will be done in #18535.

jonalmeida:
This work is blocking other issues from moving ahead until we have it landed.

Issue #18443, opened 2021-03-11
Re-add multi-select support on browser tray lists


jonalmeida:
We should rely heavily on the [`recyclerview-selection`](https://developer.android.com/reference/androidx/recyclerview/selection/package-summary) library as much as possible.

A short investigation was done in https://github.com/mozilla-mobile/fenix/issues/14969#issuecomment-705908516 :

> > I'm glad that you were able to get the RecyclerView selection working! And you can trigger it externally via a button?
> 
> Yes, there's an API that let's you define the single-click property. You can see this in the [Sunflower demo app][sunflower] which uses the selection library that we can modify with the patch below in [the adapter][sunflower_adapter]. We would obviously do this dynamically depending on the state we have in the store.

[sunflower]: https://github.com/android/sunflower/
[sunflower_adapter]: https://github.com/android/sunflower/blob/a9b9e64cf4e7d79547a5f3a4150cee04d88bbecc/app/src/main/java/com/google/samples/apps/sunflower/adapters/GardenPlantingAdapter.kt#L92-L96
[cta_banner]: https://github.com/mozilla-mobile/fenix/blob/666781aa62f579308076adcbcb9d69565eee4717/app/src/main/java/org/mozilla/fenix/tabtray/TabTrayView.kt#L260-L281
[0]: https://github.com/mozilla-mobile/fenix/issues/13950#issuecomment-683187827

PR #18541, opened 2021-03-18
Issue #18538: Add BrowserTabsAdapter for tabs tray

jonalmeida:
A base tabs tray so we no longer need to use the old `FenixTabsAdapter`.

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/tabstray/browser/BrowserTabsAdapter.kt
+++ b/app/src/main/java/org/mozilla/fenix/tabstray/browser/BrowserTabsAdapter.kt
@@ -1,0 +1,84 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.tabstray.browser
+
+import android.content.Context
+import android.view.LayoutInflater
+import android.view.ViewGroup
+import androidx.recyclerview.widget.RecyclerView
+import androidx.recyclerview.widget.GridLayoutManager
+import kotlinx.android.synthetic.main.tab_tray_item.view.*
+import mozilla.components.browser.tabstray.TabViewHolder
+import mozilla.components.browser.thumbnails.loader.ThumbnailLoader
+import mozilla.components.concept.tabstray.TabsTray
+import mozilla.components.support.base.observer.Observable
+import mozilla.components.support.base.observer.ObserverRegistry
+import org.mozilla.fenix.R
+import org.mozilla.fenix.ext.components
+import org.mozilla.fenix.ext.settings
+import org.mozilla.fenix.tabtray.TabTrayViewHolder
+
+/**
+ * A [RecyclerView.Adapter] for browser tabs.
+ */
+class BrowserTabsAdapter(
+    private val context: Context,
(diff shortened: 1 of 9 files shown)


app/src/main/java/org/mozilla/fenix/tabstray/browser/BrowserTrayInteractor.kt
```diff
@@ -0,0 +1,62 @@
...
+     */
+    override fun onCloseTab(tab: Tab) {
+        removeUseCases.invoke(tab.id)
+    }
+
+    /**
+     * See [BrowserTrayInteractor.onMultiSelect].
+     */
+    override fun onMultiSelect(enabled: Boolean) {
+        // TODO
```
kglazko, line 60: What will be done here? Could use some more detail in the to do.

app/src/main/java/org/mozilla/fenix/tabstray/browser/BrowserTabsAdapter.kt
```diff
@@ -0,0 +1,87 @@
...
+
+        return TabTrayViewHolder(view, imageLoader)
+    }
+
+    override fun onBindViewHolder(holder: TabViewHolder, position: Int) {
+        super.onBindViewHolder(holder, position)
+
+        holder.tab?.let { tab ->
+            val tabIsPrivate =
+                context.components.core.store.state.findTab(tab.id)?.content?.private == true
```
kglazko, line 68: What's the benefit of doing the '== true' as opposed to just the boolean value of .private?

jonalmeida: Because `findTab` is nullable, the result from `.private` is also nullable. That means we need to check that the value we want to check is not null and then test it i.e. `.private != null && .private`.

This is way is just shorthand for that expression, and also a default lint suggestion from the IDE.

app/src/main/java/org/mozilla/fenix/tabstray/browser/BrowserTrayInteractor.kt
jonalmeida: Sorry, I spoke offline with Codrut who will be working on this in https://github.com/mozilla-mobile/fenix/issues/18443.

This patch was blocking that work, so I just left a to-do comment to appease the linter, but it might end up being an unnecessary interactor method when it's worked on.

app/src/main/java/org/mozilla/fenix/tabstray/browser/BrowserTabsAdapter.kt
jonalmeida: We can get rid of this entire line after we land https://github.com/mozilla-mobile/android-components/pull/9934.

jonalmeida: The AC PR landed, so I can update this line now to use `tab.private` instead.

app/src/main/java/org/mozilla/fenix/tabstray/browser/BrowserTrayInteractor.kt
jonalmeida: Updated the comment to reference the issue number.

rocketsroger (approved):
LGTM

PR #18579, opened 2021-03-22
For #18395: Text selection menu is not dismissed after activating reader mode

### LINKED ISSUE #18395: [Bug]Text selection menu is not dismissed after the page is in reading mode
[comment]: # (Please do your best to search for duplicate issues before filing a new issue so we can keep our issue board clean)
[comment]: # (Every issue should have the exact bug and steps to reproduce described in it. Please do not file feedback list tickets as it is difficult to parse them and address their individual points)
[comment]: # (Read https://github.com/mozilla-mobile/fenix#i-want-to-file-an-issue for more information)

## Steps to reproduce
1. Open an page with a lot of content.
2. Perform a long tap on a word in order to bring up the selection menu.
3. From the URL bar, tap on the reading icon.

### Expected behavior
The text selection menu is dismissed and the page enter in reading mode.

### Actual behavior
The text selection menu is still displayed and no word is selected. 

### Device information
* Device vendor / model and Android version: Xiaomi Redmi Note 8T (Android 9), LG G7 Fit (Android 8.1).
* Firefox for Android version: 86.1.1, 87-beta.4, Nightly 3/9.

VitalyVPinchuk:
Fixes #18395 

Added a variable `actionMode` to track the state of contextual text selection menu and a method `finishActionMode()` to dismiss it before switching to reader mode.

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarController.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarController.kt
@@ -99,6 +99,7 @@ class DefaultBrowserToolbarController(
     }
 
     override fun handleReaderModePressed(enabled: Boolean) {
+        activity.finishActionMode()
         if (enabled) {
             readerModeController.showReaderView()
             metrics.track(Event.ReaderModeOpened)
--- a/app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarController.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarController.kt
@@ -99,7 +99,6 @@ class DefaultBrowserToolbarController(
     }
 
     override fun handleReaderModePressed(enabled: Boolean) {
-        activity.finishActionMode()
         if (enabled) {
             readerModeController.showReaderView()
             metrics.track(Event.ReaderModeOpened)
--- a/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
+++ b/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
@@ -167,6 +167,9 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
 
     private lateinit var navigationToolbar: Toolbar
 
+    // Tracker for contextual menu (Copy|Search|Select all|etc...)
+    private var actionMode: ActionMode? = null
+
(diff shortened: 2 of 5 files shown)


app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarController.kt
```diff
@@ -99,6 +99,7 @@ class DefaultBrowserToolbarController(
     }
 
     override fun handleReaderModePressed(enabled: Boolean) {
+        activity.finishActionMode()
```
Mugurell, line 102: This works but adds to the tight coupling of this Controller to HomeActivity.
(we should probably refactor out the activity reference altogether, but in another ticket)
Ideally the Controller should not know or care whether the activity actually has a `finishActionMode` method.

The easiest way to decouple the Controller from HomeActivity would be to have the Controller receive another callback in it's constructor, something like `onReaderModeChanged: () -> Unit` maybe? (It could also send a Boolean) so that any clients of this Controller could decide whether they are interested or not in this functionality, without the Controller having to know anything about them.

Mugurell (changes requested):
This looks good, thank you!
I had two suggestions on how to improve this a little and I'd also want to ask if you could update [this 2 tests](https://github.com/mozilla-mobile/fenix/blob/975f97df467b82ecbbfcde8ec009adae310e00c4/app/src/test/java/org/mozilla/fenix/components/toolbar/DefaultBrowserToolbarControllerTest.kt#L202-L216) with the new changes.

VitalyVPinchuk:
Thanks for the feedback!
I'll try to fix it.

app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarController.kt
VitalyVPinchuk: Added callback `onReaderModeChanged: () -> Unit = {}` to `DefaultReaderModeController`'s constructor

VitalyVPinchuk:
I've made changes to `DefaultReaderModeControllerTest`, not to `DefaultBrowserToolbarControllerTest` because of `readerModeController`, which is an interface and not an instance of `DefaultReaderModeController` class.
If it's still to be in `DefaultBrowserToolbarControllerTest` as you suggest, I need your advice how to test a class variable through an interface (cast? hint? what else).

Mugurell (approved):
Looks good!
Thank you!

PR #18609, opened 2021-03-23
For #17798: Sync Menu Item

kglazko:
For #17798 

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/androidTest/java/org/mozilla/fenix/ui/robots/SyncedTabsRobot.kt
+++ b/app/src/androidTest/java/org/mozilla/fenix/ui/robots/SyncedTabsRobot.kt
@@ -39,6 +39,8 @@ private fun goBackButton() =
     onView(allOf(withContentDescription("Navigate up")))
 
 private fun assertSyncedTabsMenuHeader() {
-    onView(withText(R.string.synced_tabs))
+    // Replaced with the new string here, the test is assuming we are NOT signed in
+    // Sync tests in SettingsSyncTest are still TO-DO, so I'm not sure that we have a test for signing into Sync
+    onView(withText(R.string.sync_menu_sign_in))
         .check((matches(withEffectiveVisibility(Visibility.VISIBLE))))
 }
--- a/app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
@@ -402,13 +402,21 @@ open class DefaultToolbarMenu(
         id = WebExtensionPlaceholderMenuItem.MAIN_EXTENSIONS_MENU_ID
     )
 
-    val syncedTabs = BrowserMenuImageText(
-        label = context.getString(R.string.synced_tabs),
-        imageResource = R.drawable.ic_synced_tabs,
-        iconTintColorResource = primaryTextColor()
-    ) {
-        onItemTapped.invoke(ToolbarMenu.Item.SyncedTabs)
-    }
+        val accountManager = context.components.backgroundServices.accountManager
+        val account = accountManager.authenticatedAccount()
+        val syncItemTitle = if (account != null && accountManager.accountProfile()?.email != null) {
+            context.getString(R.string.sync_signed_as, accountManager.accountProfile()?.email)
+        } else {
+            context.getString(R.string.sync_menu_sign_in)
+        }
+
+        val syncedTabs = BrowserMenuImageText(
+            syncItemTitle,
+            R.drawable.ic_synced_tabs,
+            primaryTextColor()
+        ) {
+            onItemTapped.invoke(ToolbarMenu.Item.SyncedTabs)
+        }
 
     val findInPageItem = BrowserMenuImageText(
         label = context.getString(R.string.browser_menu_find_in_page),
--- a/app/src/main/java/org/mozilla/fenix/home/HomeMenu.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/HomeMenu.kt
@@ -189,8 +189,16 @@ class HomeMenu(
             onItemTapped.invoke(Item.Settings)
         }
 
+        val accountManager = context.components.backgroundServices.accountManager
+        val account = accountManager.authenticatedAccount()
+        val syncItemTitle = if (account != null && accountManager.accountProfile()?.email != null) {
+            context.getString(R.string.sync_signed_as, accountManager.accountProfile()?.email)
+        } else {
+            context.getString(R.string.sync_menu_sign_in)
+        }
+
         val syncedTabsItem = BrowserMenuImageText(
-            context.getString(R.string.library_synced_tabs),
+            syncItemTitle,
             R.drawable.ic_synced_tabs,
             primaryTextColor
         ) {
@@ -215,13 +223,11 @@ class HomeMenu(
 
         // Only query account manager if it has been initialized.
         // We don't want to cause its initialization just for this check.
-        val accountAuthItem =
-            if (context.components.backgroundServices.accountManagerAvailableQueue.isReady() &&
-                context.components.backgroundServices.accountManager.accountNeedsReauth()) {
-                    reconnectToSyncItem
-            } else {
-                null
-            }
+        val accountAuthItem = if (context.components.backgroundServices.accountManagerAvailableQueue.isReady()) {
+            if (context.components.backgroundServices.accountManager.accountNeedsReauth()) reconnectToSyncItem else null
+        } else {
+            null
+        }
 
         val settings = context.components.settings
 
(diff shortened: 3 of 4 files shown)


kglazko:
Putting this up for feedback @eliserichards @gabrielluong just to see what you think of this approach. I see a bug with this on the emulator where for some reason on the emulator when you install it is never recognized as signed in until you navigate somewhere in the app and come back. Attaching a photo to show how it looks.
Screen Shot 2021-03-23 at 1 21 38 PM

eliserichards:
What are the steps that get to this point? Open the app (fresh build), open menu to settings, sign in...

kglazko:
Yeah, I would have to go to Settings, sign into Sync, and then see whether or not it shows up on the menu. I signed into Sync when it still wasn't working and it kept me signed in so eventually I would see it on a fresh install of the app- but only after navigating somewhere else first, I think the menu loads before Sync does maybe...

eliserichards (commented):
This is a good start! A few pointers: 

* Make sure we're adding this in both the `HomeMenu` (for new tab) and the `DefaultToolbarMenu` (main menu)
* Using the AccountManager is the right way to go. Take a look at the [SyncPreferenceView](https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/java/org/mozilla/fenix/settings/SyncPreferenceView.kt) and how it's registering for account changes from the manager. We want to be listening for any changes and dynamically updating the list when we receive a change
* The [AccountUiView](https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/java/org/mozilla/fenix/settings/account/AccountUiView.kt) might also be helpful
* When we [receive an update](https://github.com/mozilla-mobile/fenix/blob/246c192de377fa180029604846bb3497c4770dd9/app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt#L532) from the `AccountManager`, we want to update the item in our menu. Can we dynamically update _just the item_, or do we need to refresh the whole list? It looks like we're already doing a rebuild of the list here: https://github.com/mozilla-mobile/fenix/blob/246c192de377fa180029604846bb3497c4770dd9/app/src/main/java/org/mozilla/fenix/home/HomeMenu.kt#L243

* > I think the menu loads before Sync does
   * If we are loading too quickly, we want to make sure we wait until all processes are complete before rebuilding the list

kglazko:
Yeah, upon investigating it looks like we are already rebuilding the list on a log out or reuthenticate event. So that's good. I also realized that I couldn't find finished UI tests around sync sign-in, there are some that are skeleton-ed, but we may need to write them for this PR.

kglazko:
>       * If we are loading too quickly, we want to make sure we wait until all processes are complete before rebuilding the list

Do you  have any ideas on how we can do this, @eliserichards ? We probably wouldn't want to hold up the rest of the menu UI from loading if it's waiting on a process.

eliserichards:
For the UI tests, fix up what you can and then @ignore the ones that need more time and Isabel's team will look at them. Make sure you tag the UI test issue in the @ignore string so they know where to look.

WRT loading: you're correct.. the menu should load and shouldn't be waiting on the sync. So it should default to the signed out UI state and then reload everything once we get the auth response

app/src/androidTest/java/org/mozilla/fenix/ui/robots/SyncedTabsRobot.kt
```diff
@@ -39,6 +39,8 @@ private fun goBackButton() =
     onView(allOf(withContentDescription("Navigate up")))
 
 private fun assertSyncedTabsMenuHeader() {
```
eliserichards, line 41: Nit: rename the test something like this so you don't need the inline comment: 
```suggestion
 private fun `GIVEN xxxx WHEN xxxx THEN xxxx`() {
```

app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
```diff
@@ -236,10 +236,18 @@ class DefaultToolbarMenu(
             onItemTapped.invoke(ToolbarMenu.Item.AddToHomeScreen)
         }
 
+        val accountManager = context.components.backgroundServices.accountManager
+        val notSignedIn = accountManager.authenticatedAccount()
```
eliserichards, line 240: Is this an async call? Do we need to ensure that this has finished before evaluating `notSignedIn`?

eliserichards: Also, should this be `signedIn` instead?

app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
```diff
@@ -236,10 +236,18 @@ class DefaultToolbarMenu(
...
+        val syncStringToDisplay = if (notSignedIn != null && accountManager.accountProfile()?.email != null) {
```
eliserichards, line 241: Nit:
```suggestion
        val syncItemTitle
```

app/src/main/res/values/strings.xml
```diff
@@ -953,6 +953,8 @@
     <string name="share_link_all_apps_subheader">All actions</string>
     <!-- Sub-header in the dialog to share a link to an app from the most-recent sorted list -->
     <string name="share_link_recent_apps_subheader">Recently used</string>
+    <!-- An string shown when an account is signed in where %1$s is a placeholder for the email-->
+    <string name="sync_signed_as">Signed in as %1$s</string>
     <!-- An option from the share dialog to sign into sync -->
     <string name="sync_sign_in">Sign in to Sync</string>
```
eliserichards, line 957: Just confirming this is capital S "Sync" not lowercase 😁

eliserichards (changes requested):
Looking good! A few questions and nits. Make sure we're only changing the synced tabs item in `newCoreMenuItems`, not the old menu items.

app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
gabrielluong: I think it probably makes more sense to call it `account` to not confuse this with a boolean value.

app/src/main/res/values/strings.xml
kglazko: Good catch, the design spec shows it as lowercase! Changing it now :)

app/src/androidTest/java/org/mozilla/fenix/ui/robots/SyncedTabsRobot.kt
kglazko: For UI tests, the convention doesn't seem to be 'GIVEN WHEN THEN' but rather the assert style based on a few other robots I looked at. Tagging @AaronMT or @rpappalax to give input

kglazko:
Thank you for your feedback, @eliserichards and @gabrielluong. I incorporated your comments, let me know if there's anything else I can do.

app/src/androidTest/java/org/mozilla/fenix/ui/robots/SyncedTabsRobot.kt
eliserichards: that's totally fine, I didn't realize 😄

eliserichards:
One thing to confirm: when we click on the item, where do we want it to go? right now it's directing to our synced tabs. We can confirm this with Victoria

kglazko:
Filing this as a follow-up after discussing with @eliserichards https://github.com/mozilla-mobile/fenix/issues/18806

eliserichards (approved):
Awesome job! 🎉

app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
```diff
@@ -402,13 +402,21 @@ open class DefaultToolbarMenu(
...
+            context.getString(R.string.sync_menu_sign_in)
+        }
+
+        val syncedTabs = BrowserMenuImageText(
+            syncItemTitle,
+            R.drawable.ic_synced_tabs,
+            primaryTextColor()
+        ) {
+            onItemTapped.invoke(ToolbarMenu.Item.SyncedTabs)
+        }
```
gabrielluong, line 419: I think we got an extra indent here.

app/src/main/java/org/mozilla/fenix/home/HomeMenu.kt
```diff
@@ -215,13 +223,11 @@ class HomeMenu(
...
         // We don't want to cause its initialization just for this check.
-        val accountAuthItem =
-            if (context.components.backgroundServices.accountManagerAvailableQueue.isReady() &&
-                context.components.backgroundServices.accountManager.accountNeedsReauth()) {
-                    reconnectToSyncItem
-            } else {
-                null
-            }
+        val accountAuthItem = if (context.components.backgroundServices.accountManagerAvailableQueue.isReady()) {
+            if (context.components.backgroundServices.accountManager.accountNeedsReauth()) reconnectToSyncItem else null
```
gabrielluong, line 227: Just for awareness I think this extra if statement would be the equivalent as the above using &&. This probably didn't need to be change.

PR #18614, opened 2021-03-24
set tasks_for to action to use release signing in relpro

escapewindow:
I *think* this will fix beta signing.
We may need to uplift this fix to other branches, not sure.

This is the fenix equivalent to https://github.com/mozilla-mobile/android-components/blob/master/taskcluster/ac_taskgraph/release_promotion.py#L154 .


jmaher (approved):
sounds safe

escapewindow:
Doh, I should have waited for the merge to reuse the branch

kbrosnan (approved):

PR #18645, opened 2021-03-26
For #18644 - Temporarily disabling UI tests that fail on Firebase.

Mugurell:
This is a quick solution to resume PR merges.
Investigations and a proper fix is to follow on this same ticket.


AndiAJ:
Hi @Mugurell , please use @ Ignore instead of commenting all the lines.
Thanks! ☺️

mcarare (approved):

AndiAJ (approved):
LGTM !

PR #18719, opened 2021-03-31
For #18265 - [Add card] Integrate the credit card storage and handle adding a new credit card

### LINKED ISSUE #18265: [Add card] Integrate the credit card storage and handle adding a new credit card
This issue is for initializing the autofill credit card storage in Fenix, adding appropriate unit tests for the integration, and as a test we should implement the "Save" button functionality in the "Add card" screen.

Screen Shot 2021-03-02 at 3 41 24 PM

gabrielluong:
Fixes #18265

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/res/menu/credit_card_editor.xml
+++ b/app/src/main/res/menu/credit_card_editor.xml
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<menu xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto">
+    <item
+        android:id="@+id/save_credit_card_button"
+        android:icon="@drawable/mozac_ic_check"
+        android:title="@string/save"
+        app:iconTint="?primaryText"
+        app:showAsAction="ifRoom" />
+</menu>
--- a/app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardEditorFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardEditorFragment.kt
@@ -5,13 +5,22 @@
 package org.mozilla.fenix.settings.creditcards
 
 import android.os.Bundle
+import android.view.Menu
+import android.view.MenuInflater
+import android.view.MenuItem
 import android.view.View
 import android.widget.ArrayAdapter
 import androidx.fragment.app.Fragment
+import androidx.lifecycle.lifecycleScope
 import androidx.navigation.fragment.findNavController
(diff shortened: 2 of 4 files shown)


app/src/main/res/menu/credit_card_editor.xml
```diff
@@ -0,0 +1,21 @@
...
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<menu xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto">
+    <item
+        android:id="@+id/delete_credit_card_button"
+        android:icon="@drawable/ic_delete"
+        android:title="@string/credit_cards_delete_card"
+        android:visible="false"
```
Mugurell, line 11: I see this option not used.
Should it have the same functionality as the "cancel" button from the fragment?
Or if not used this can be removed from the current patch and added when  there's a need for it?

app/src/main/res/menu/credit_card_editor.xml
```diff
@@ -0,0 +1,21 @@
...
+        app:iconTint="?primaryText"
+        app:showAsAction="ifRoom" />
+    <item
+        android:id="@+id/save_credit_card_button"
+        android:icon="@drawable/mozac_ic_check"
+        android:title="@string/save"
+        android:contentDescription="@string/save"
```
Mugurell, line 18: This menu item uses a content description while the above one doesn't.
By default Talkback will read the title anyway so we can go with either approaches but for consistency one approach should be used in all  cases.

Mugurell (commented):
This looks okay but something that caught my eye is the use of an Interactor & Controller that indicates we'd want to have this new feature follow our MVI architecture but here there's no MVI loop, no standalone View used by the Fragment so this two classes seem superfluous to me.

If not following the MVI, the Fragment could contain all the business logic and the methods needed to update the UI or maybe it could have just one other class that it could delegate much of the business logic to but at least the Interactor here seems not needed.

gabrielluong:
> This looks okay but something that caught my eye is the use of an Interactor & Controller that indicates we'd want to have this new feature follow our MVI architecture but here there's no MVI loop, no standalone View used by the Fragment so this two classes seem superfluous to me.
> 
> If not following the MVI, the Fragment could contain all the business logic and the methods needed to update the UI or maybe it could have just one other class that it could delegate much of the business logic to but at least the Interactor here seems not needed.

Thanks for feedback regarding the MVI architecture. On first look, it made sense to me to just follow our current practises. I removed the Interactor for the time being, and kept the Controller to delegate the business logic for time and to ensure it's easily testable. We can revisit the need for having the Interactor as the fragment becomes more complex and whether or not we should delegate more code into a View class.

Mugurell (approved):
Looks good!
Thank you!

Mugurell:
Thank you @gabrielluong for making this changes.
I'd say this looks cleaner. Indeed Interactors were planned to being phased out since they already had very little real responsability - https://github.com/mozilla-mobile/fenix/blob/master/docs/architecture-overview.md#interactor .

Issue #18717, opened 2021-03-30
[Bug] Synced Tabs list cannot be seen when in private browsing mode


jonalmeida:
[comment]: # (Please do your best to search for duplicate issues before filing a new issue so we can keep our issue board clean)
[comment]: # (Every issue should have the exact bug and steps to reproduce described in it. Please do not file feedback list tickets as it is difficult to parse them and address their individual points)
[comment]: # (Read https://github.com/mozilla-mobile/fenix#i-want-to-file-an-issue for more information)
(there was a bug about this before from the first iteration of synced tabs in the tray, but I can't find it any more)

## Steps to reproduce
1. (Optional) Enable synced tabs feature flag to show in tabs tray.
2. Enable Private Browsing mode (the theme should change to purple).
3. Open the tabs tray to the Synced Tabs position.

### Expected behavior
- Be able to see your synced tabs.

### Actual behavior
- The title is invisible and the url is too light.

### Device information

* Device vendor / model and Android version: Android 10, Pixel 3
* Firefox for Android version: ? (go to Settings -> About Firefox) Debug/Nightly 30/03/21

<img width="381" alt="Screen Shot 2021-03-30 at 7 43 37 PM" src="https://user-images.githubusercontent.com/1370580/113070539-ccaadd00-9190-11eb-9b3a-2b0ba0b87a04.png">

PR #18725, opened 2021-03-31
For #18717: Use normal theme color for tab tray synced tabs.

mcarare:
### Pull Request checklist

1. click on Show All Checks,


jonalmeida (approved):
Thanks for this! ✅ 

> I am not sure if we plan to show synced tabs in other parts of the app (like we've done before), but following up on this, if the same XML resource will be used in other screens that have a private theme-specific background, the colors would have to be programmatically adjusted to use color attributes in that screen/ fragment.

I think that's okay because we mean to remove it from the 3-dot menu when we flip this new tabs tray on, and the only place it will show up is in the tray.

> Just as a note, if UX would have designed a private theme-specific background for the tab tray, these adjustments would not have been necessary.

The intention was to have a private theme for the tray, but we didn't have the resources to follow-up on that. Let's assume for now that we're going to keep it this way. 👌

Issue #17904, opened 2021-02-08
[Bug][a11y] Add private shortcut button needs resizing


sflorean:
[comment]: # (Please do your best to search for duplicate issues before filing a new issue so we can keep our issue board clean)
[comment]: # (Every issue should have the exact bug and steps to reproduce described in it. Please do not file feedback list tickets as it is difficult to parse them and address their individual points)
[comment]: # (Read https://github.com/mozilla-mobile/fenix#i-want-to-file-an-issue for more information)
## Prerequisites
Have Google Accessibility Scanner installed.

## Steps to reproduce
1. From the Home screen, tap on the upper right Private Browsing icon and switch to Private Mode. (perform the switch 3 times).
2. Using the Accessibility Scanner, scan the Add Private Browsing Shortcut to your Home screen contextual recommender.

### Expected behavior
The Private Browsing Shortcut recommender is properly scanned and no suggestions are made.

### Actual behavior
![Screenshot_20210204_164533_com google android apps accessibility auditor](https://user-images.githubusercontent.com/58466444/107213900-c8c0d100-6a11-11eb-85ae-f542faf49b6f.jpg)

### Device information
* Device vendor / model and Android version: Huawei P40 Lite (Android 11).
* Firefox for Android version: 86.0.0-beta.2.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-9664)

eliserichards:
https://github.com/mozilla-mobile/fenix/pull/17985

liacds:
can I work on this issue?

liacds:
I can't seem to reproduce this issue in the latest version on Pixel 2, everything works as expected

Apurwa96:
Hi, This is my first time contributing to opensource, let me know if I can work on this.

codesuperstarr:
I am an outreachy applicant. Please, I request to be assigned this issue.

eliserichards:
Hello all! @liacds, are you an Outreachy applicant? If so, since you were the first to comment, you can pick this issue up :)

liacds:
@eliserichards Yes, thank you :)

topotropic:
Thanks for the ping! If this is a problem with the button here, it might be a problem in other places as well. Our component in Figma has a height of 36dp – I remember that Sawyer worked on componentizing buttons at some point last year, so it would be great if we could systematically solve this and not just increase the size of the button here to fix this single case.

[...]

PR #18736, opened 2021-03-31
For #17904: resized add shortcut button

liacds:
Add shortcut button size looks like this now
Screenshot_1617220812
and accessibility scan doesn't produce any suggestions for the button
Screenshot_1617220798

### Pull Request checklist

1. click on Show All Checks,


eliserichards (approved):
Great job! 🚀

gabrielluong:
Let's get a UX review from @topotropic or @brampitoyo since there's screenshots.

topotropic:
Thanks for the ping! If this is a problem with the button here, it might be a problem in other places as well. Our component in Figma has a height of 36dp – I remember that Sawyer worked on componentizing buttons at some point last year, so it would be great if we could systematically solve this and not just increase the size of the button here to fix this single case.

liacds:
> Thanks for the ping! If this is a problem with the button here, it might be a problem in other places as well. Our component in Figma has a height of 36dp – I remember that Sawyer worked on componentizing buttons at some point last year, so it would be great if we could systematically solve this and not just increase the size of the button here to fix this single case.

Hi again! I would like to look into this, but I would need some guidance as to how to tackle this issue.

brampitoyo:
@liacds I am working together with @gabrielluong to resolve this problem, and the approach that I was thinking of taking, is to introduce 6dp “padding” on top and bottom. They’re invisible, but they can still be tapped. By doing this, we’ll increase the tappable area to 48dp, while keeping the button height 36dp.

But I’m not sure whether Android gives us the ability to extend tappable area past the content area itself?

eliserichards (changes requested):
Let's try @topotropic's suggestion and make the tappable area bigger while keeping the same visual "size" of the button

boek:
Closing in favor of #28056

Issue #18142, opened 2021-02-24
Address bar search freezes phone when there's sth super long in the clipboard 


mbalc:
## Steps to reproduce
- copy a very, very long string to the clipboard
    - my specific case was that I had copied a save state from [a browser game under this address](https://jacorb90.github.io/Prestige-Tree/) (using "export to clipboard" option) and then forgot about that soon after I was done with backing it up
- (probably optional) have over 100 tabs open gathered over many months of browser usage
- click address bar to begin a search

### Expected behavior
- keyboard instantly pops up as usual and I can search

### Actual behavior
- the whole phone freezes, it takes roghly 30 seconds for the keyboard to show (and the search suggestions area to fill the whole screen) but the phone remains unresponsive until Android finds Firefox process unresponsive and gives me a prompt with an option to kill the app
- restarting the browser and overwriting the clipboard with any random short string fixes the problem - triggering search then proceeds successfully without any issues

### Device information

* Android device: lg g7 thinq, android 10
* Fenix version: see screenshot below

<details>
  <summary>Click to view my Firefox's 'about' page</summary>
  

![Screenshot_20210224-100055](https://user-images.githubusercontent.com/11033790/108996186-5fda8980-769e-11eb-9559-feb5657cc8ea.png)
  
</details>

mcomella:
Triage: this is a really bad experience even if it might not happen that often so let's look into it, especially because it's a part of the search experience. There's some history around perf issues and really long clipboard strings but we were unable to find the bugs.

mcomella:
The PR provided https://github.com/mozilla-mobile/fenix/pull/18759 is a possible fix but the discussion around it has stalled so we closed the PR and will unassign this as low impact as we're focusing in other areas.

mcomella:
Dupe to https://github.com/mozilla-mobile/fenix/issues/21291, which has a merged PR.

PR #18759, opened 2021-04-01
For #18142: address bar freeze

MarcLeclair:
For #18142

This fixes the jank created when a user has a super long string in their clipboard and tries to click on the URL bar. 

However, I have to point out that by calling `URLStringUtils`, there is one small change in behavior.

- Behavior before the changes:
If a user had the following string in their clipboard `abcdefg mozilla.com` the clipboard would offer to "FIll the link" with 
`mozilla.com`. Another example would be having `mozilla.com https://mozilla.com` would show the option to paste the link `https://mozilla.com`.

- Behavior with this patch:
If a user has the string `abcdefg mozilla.com`, nothing will be shown. However, if the user has  some sort of URL (ie: `c-c.com`, `about-mozilla:mozilla`, etc...) it will offer to paste the link from the clipboard.

The reason for the change is that to fix the problem in `WebURLFinder.kt`, we would need to fix the regex which is used in multiple different place ( one of them being `TabIntentProcessor.kt`). I didn't want to change the behavior of those classes by changing the regex. 
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/utils/ClipboardHandler.kt
+++ b/app/src/main/java/org/mozilla/fenix/utils/ClipboardHandler.kt
@@ -10,7 +10,7 @@ import android.content.Context
 import androidx.annotation.VisibleForTesting
 import androidx.core.content.getSystemService
 import mozilla.components.support.utils.SafeUrl
-import mozilla.components.support.utils.WebURLFinder
+import mozilla.components.support.utils.URLStringUtils
 
 private const val MIME_TYPE_TEXT_PLAIN = "text/plain"
 private const val MIME_TYPE_TEXT_HTML = "text/html"
@@ -38,8 +38,7 @@ class ClipboardHandler(val context: Context) {
     val url: String?
         get() {
             return text?.let {
-                val finder = WebURLFinder(it)
-                finder.bestWebURL()
+               if(URLStringUtils.isURLLike(it)) it else null
             }
         }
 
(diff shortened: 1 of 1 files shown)


mcomella (commented):
To be explicit, I think we should get a reviewer from fenix for this change as they'll be in a better position to validate the correctness of this change. I'm doing a skim for perf issues.

This general approach seems good to me for perf but we should squash commits like, "Fixed indentation" into the original commit (assuming it's not too much trouble).

app/src/main/java/org/mozilla/fenix/utils/ClipboardHandler.kt
```diff
@@ -38,8 +38,7 @@ class ClipboardHandler(val context: Context) {
     val url: String?
         get() {
             return text?.let {
-                val finder = WebURLFinder(it)
-                finder.bestWebURL()
+                if(URLStringUtils.isURLLike(it)) it else null
```
eliserichards, line 41:
```suggestion
                if (URLStringUtils.isURLLike(it)) { it } else { null }
```
should fix lint issue

eliserichards:
I'll check if the use case you mentioned makes a difference, but I think we should be good to go on this once that's confirmed.

eliserichards:
I spoke to Vesta, and we don't want to regress that behavior. 

What is the tradeoff here if we hold off on merging this until a solution can be found to ensure we still have that clipboard behavior?

MarcLeclair:
@eliserichards  the trade off at this moment is the performance / jankiness that occurs when having a long string in the keyboard is can affect users. We can keep the same behavior by duplicating code (i.e: the method that separates the string into multiple part to identify differenty URL isn't the root of the performance problem).  However, anyone with a long string in their clipboard will take longer to get to the search experience. I'd like to add that this doesn't happen on any other browser that I've tried it on (i.e: having a long string in my clipboard doesn't lag)

mcomella:
This is a possible fix but the discussion around it has stalled so we're going to close it. Feel free to reopen and iterate if it's useful to you.

PR #18766, opened 2021-04-02
[SDK-246] Change Nimbus `channel` input to use the same method as Glean

travis79:


app/src/main/java/org/mozilla/fenix/experiments/NimbusSetup.kt
```diff
@@ -47,7 +47,10 @@ fun createNimbus(context: Context, url: String?): NimbusApi =
         // https://github.com/mozilla/probe-scraper/blob/master/repositories.yaml
         val appInfo = NimbusAppInfo(
             appName = "fenix",
-            channel = Config.channel.toString()
+            // Note: Using BuildConfig.BUILD_TYPE is important here so that it matches the value
+            // passed into Glean.  `Config.channel.toString()` turned out to be non-deterministic
```
gabrielluong, line 51:
```suggestion
            // passed into Glean. `Config.channel.toString()` turned out to be non-deterministic
```

gabrielluong (approved):

PR #18782, opened 2021-04-03
Adjusted the "Add-private-shortcut-texts" contrast ratio

codesuperstarr:


gabrielluong:
Since this is dealing with color values that were not identified by UX, I think we should block on a UX review.

app/src/main/res/values/colors.xml
```diff
@@ -85,7 +85,7 @@
     <color name="tab_history_item_selected_background_light_theme">@color/tab_tray_item_selected_background_light_theme</color>
 
     <!-- Dark theme color palette -->
-    <color name="primary_text_dark_theme">#FBFBFE</color>
+    <color name="primary_text_dark_theme">#FCFBFE</color>
```
gabrielluong, line 88: The 2 color variables here are used in a number of places. Are we certain that this is the correct change for all the other instances?

codesuperstarr: @gabrielluong Yes, you are right. I am not certain that these changes would affect other use instances positively.  
However, with your permission, I could set these color variables as is before for other use instances, while creating a new color item in the resource file to adjust the color contrast ratio of this particular touch item that this pull request seeks to address.

gabrielluong: I don't think it would be worth taking more action without more guidance from UX here. I think we can leave it as-is and circle back. If this is for outreachy, I think it is satisfactory to have undergone the process of building, testing and submitting a PR.

gabrielluong:
:construction: **Pull Request Maintenance** :construction: Thank for your contribution! We’re closing this pull request since it has been stale for awhile. If the pull request is still relevant, please feel free to reopen or comment on any open issue if you’re interested in contributing further. Thanks again! :bow:

PR #18785, opened 2021-04-03
Bug 1670902 - Use WebRender platforms for Raptor performance tests

davehunt:
This is an attempt to use the WebRender (-qr) platforms for the Raptor performance tests instead of augmenting the command line and job group/symbols. This is a test-only change.

@gmierz could you help with testing this?

### DIFF
--- a/taskcluster/fenix_taskgraph/transforms/browsertime.py
+++ b/taskcluster/fenix_taskgraph/transforms/browsertime.py
@@ -77,15 +77,17 @@ def build_browsertime_task(config, tasks):
         abi = task["attributes"]["abi"]
         apk = task["attributes"]["apk"]
 
-        test_name = task.pop("test-name")
-
-        task["name"] = "{}-{}-{}".format(task["name"], build_type, abi)
-        task["description"] = "{}-{}".format(build_type, abi)
-
         for key in ("args", "treeherder.platform", "worker-type"):
             resolve_keyed_by(task, key, item_name=task["name"], **{"abi": abi})
 
         task["treeherder"] = inherit_treeherder_from_dep(task, signing)
+
+        test_name = task.pop("test-name")
+        platform = task["treeherder"]["platform"]
+
+        task_name = "{}-{}-{}-{}".format(platform, test_name, build_type, abi)
+        task["name"] = task_name
+        task["description"] = task_name
 
         extra_config = {
             "installer_url": "<signing/{}>".format(apk),
@@ -150,15 +152,13 @@ def enable_webrender(config, tasks):
     for task in tasks:
         if not task.pop("web-render-only", False):
             newtask = copy.deepcopy(task)
+            newtask["name"] = newtask["name"].replace("-qr", "")
+            newtask["description"] = newtask["description"].replace("-qr", "")
+            newtask["treeherder"]["platform"] = newtask["treeherder"][
+                "platform"
+            ].replace("-qr", "")
             yield newtask
         task["run"]["command"].append("--enable-webrender")
-        task["name"] += "-wr"
-        task["description"] += "-wr"
-
-        # Setup group symbol
-        group, sym = task["treeherder"]["symbol"].split("(")
-        task["treeherder"]["symbol"] = "{}-wr({})".format(group, sym[:-1])
-
         yield task
 
 
--- a/taskcluster/ci/browsertime/kind.yml
+++ b/taskcluster/ci/browsertime/kind.yml
@@ -43,8 +43,8 @@ job-defaults:
         tier: 2
         platform:
             by-abi:
-                arm64-v8a: android-hw-p2-8-0-android-aarch64-shippable/opt
-                armeabi-v7a: android-hw-g5-7-0-arm7-api-16-shippable/opt
+                arm64-v8a: android-hw-p2-8-0-android-aarch64-shippable-qr/opt
+                armeabi-v7a: android-hw-g5-7-0-arm7-api-16-shippable-qr/opt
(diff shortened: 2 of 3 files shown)


taskcluster/fenix_taskgraph/transforms/browsertime.py
```diff
@@ -148,17 +148,10 @@ def build_browsertime_task(config, tasks):
...
+        if task.pop("web-render-only", False):
             newtask = copy.deepcopy(task)
             yield newtask
-        task["run"]["command"].append("--enable-webrender")
-        task["name"] += "-wr"
-        task["description"] += "-wr"
-
-        # Setup group symbol
-        group, sym = task["treeherder"]["symbol"].split("(")
-        task["treeherder"]["symbol"] = "{}-wr({})".format(group, sym[:-1])
```
jmaher, line 160: I understand the concept of removing the need for reversing this logic as we will have webrender by default.

1) will we have non webrender at all?
2) where do we add `--enable-webrender` to the command (and other stuff from the removed code)?
3) do we change something in test-sets.yml or equivalent in fenix?
4) do you have a try push or a command to run locally to generate the taskgraph?

davehunt: Thanks for the questions @jmaher as I'm more than a little unfamiliar with this code.

1. The intention is that we'll have WebRender for all tests, and non-WebRender for a subset of high value page load tests.
2. In mozilla-central this is added in a common tests transform [when the platform include "-qr"](https://searchfox.org/mozilla-central/rev/d7e344e956d9da2808ea33e1fe0f963ed10c142d/taskcluster/taskgraph/transforms/tests.py#663-666). I realise this might be entirely independent code from how we transform Fenix tasks. If so, I'll likely need to restore adding the command line option to enable WebRender. The other stuff removed here changes the Treeherder symbol, which should be uneccessary once WebRender and non-WebRender are running on different platforms.
3. Not that I'm aware of.
4. I would love to provide this, but I'm entirely uncertain of how. If you can help that would be appreciated.

gmierz: @davehunt you'll have to add the `--enable-webrender`, the fenix side doesn't do it for you like m-c does.

taskcluster/fenix_taskgraph/transforms/browsertime.py
```diff
@@ -150,15 +152,11 @@ def enable_webrender(config, tasks):
     for task in tasks:
         if not task.pop("web-render-only", False):
             newtask = copy.deepcopy(task)
+            task["name"].replace("-qr", "")
```
gmierz, line 155: This will also modify the original task when I think you only want to modify `newtask`.

davehunt:
@gmierz @jmaher I believe this is ready for review now. There was one failure, which looks like an intermittent crash.

taskcluster/fenix_taskgraph/transforms/browsertime.py
```diff
@@ -77,16 +77,18 @@ def build_browsertime_task(config, tasks):
...
             resolve_keyed_by(task, key, item_name=task["name"], **{"abi": abi})
 
         task["treeherder"] = inherit_treeherder_from_dep(task, signing)
 
+        test_name = task.pop("test-name")
+        platform = task["treeherder"]["platform"]
+
+        task_name = "{}-{}-{}-{}".format(platform, test_name, build_type, abi)
+        task["name"] = task_name
+        task["description"] = task_name
```
gmierz, line 90: Is there a reason why you're changing how the name and description of the task are built? This is quite different from what we currently have.

Here's an example of before and after:

Before:
```
  "metadata": {
    "name": "browsertime-tp6m-allrecipes-nightly-arm64-v8a",
    "owner": "cron@noreply.mozilla.org",
    "source": "https://github.com/mozilla-mobile/fenix/blob/85257dbb0af18efbbce40b4d92ab80b9982dc075/taskcluster/ci/browsertime",
    "description": "nightly-arm64-v8a ([Treeherder push](https://treeherder.mozilla.org/#/jobs?repo=fenix&revision=85257dbb0af18efbbce40b4d92ab80b9982dc075))"
  },
```

After:
```
  "metadata": {
    "name": "browsertime-android-hw-g5-7-0-arm7-api-16-shippable-qr/opt-allrecipes-nightly-armeabi-v7a",
    "owner": "davehunt@users.noreply.github.com",
    "source": "https://github.com/davehunt/fenix/blob/57af16cf8dcf19fd0711558f8f399880c0d5765a/taskcluster/ci/browsertime",
    "description": "android-hw-g5-7-0-arm7-api-16-shippable-qr/opt-allrecipes-nightly-armeabi-v7a ([Treeherder push](https://treeherder.mozilla.org/#/jobs?repo=fenix-pr&revision=57af16cf8dcf19fd0711558f8f399880c0d5765a))"
  },
```

davehunt: I believe we do at least need to include the platform as this indicates the WebRender status and without this we have duplicate names/descriptions. This also brings the task names/descriptions closer to what we have in m-c. For example allrecipes on moto G5 has a task name of 'test-android-hw-g5-7-0-arm7-api-16-shippable/opt-browsertime-tp6m-essential-geckoview-allrecipes-e10s'.

gmierz (approved):
r+ thanks for making these changes @davehunt!

mcomella:
Test failure looks like it's caused by infra issues (unable to fetch deps) so I think this is okay to merge. Merging...

PR #18803, opened 2021-04-05
For #15083 - Add multi select in recently closed tabs

### LINKED ISSUE #15083: Add multi select in Recently closed
## Steps to reproduce
Go to history/recently closed
Try to long press to multi select 

### Expected behavior
Multi select overlay appears
### Actual behavior
Nothing
### Device information

* Android device: Asus Zenfone
* Fenix version: Nightly

ZianeA:
Fixes #15083

https://user-images.githubusercontent.com/14791787/113617821-bd270a80-964e-11eb-9ced-e3203829d71b.mp4

This also fixes a strange behavior in the animation when deleting items. Turns out that this was happening because the Recyclerview's layout_height was set to wrap_content. The bookmarks screen presents the same issue, I'll try to fix it in another PR.

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/library/recentlyclosed/RecentlyClosedAdapter.kt
+++ b/app/src/main/java/org/mozilla/fenix/library/recentlyclosed/RecentlyClosedAdapter.kt
@@ -9,21 +9,35 @@ import android.view.ViewGroup
 import androidx.recyclerview.widget.DiffUtil
 import androidx.recyclerview.widget.ListAdapter
 import mozilla.components.browser.state.state.recover.RecoverableTab
+import org.mozilla.fenix.selection.SelectionHolder
 
 class RecentlyClosedAdapter(
     private val interactor: RecentlyClosedFragmentInteractor
-) : ListAdapter<RecoverableTab, RecentlyClosedItemViewHolder>(DiffCallback) {
+) : ListAdapter<RecoverableTab, RecentlyClosedItemViewHolder>(DiffCallback),
+    SelectionHolder<RecoverableTab> {
+
+    private var selectedTabs: Set<RecoverableTab> = emptySet()
+
     override fun onCreateViewHolder(
         parent: ViewGroup,
         viewType: Int
     ): RecentlyClosedItemViewHolder {
         val view = LayoutInflater.from(parent.context)
             .inflate(RecentlyClosedItemViewHolder.LAYOUT_ID, parent, false)
-        return RecentlyClosedItemViewHolder(view, interactor)
+        return RecentlyClosedItemViewHolder(view, interactor, this)
     }
 
     override fun onBindViewHolder(holder: RecentlyClosedItemViewHolder, position: Int) {
         holder.bind(getItem(position))
+    }
+
+    override val selectedItems: Set<RecoverableTab>
+        get() = selectedTabs
+
+    fun updateData(tabs: List<RecoverableTab>, selectedTabs: Set<RecoverableTab>) {
+        this.selectedTabs = selectedTabs
+        notifyItemRangeChanged(0, tabs.size)
+        submitList(tabs)
     }
 
     private object DiffCallback : DiffUtil.ItemCallback<RecoverableTab>() {
@@ -31,6 +45,6 @@ class RecentlyClosedAdapter(
             oldItem.id == newItem.id
 
         override fun areContentsTheSame(oldItem: RecoverableTab, newItem: RecoverableTab) =
-            oldItem.id == newItem.id
+            oldItem == newItem
     }
 }
--- a/app/src/main/java/org/mozilla/fenix/library/recentlyclosed/RecentlyClosedController.kt
+++ b/app/src/main/java/org/mozilla/fenix/library/recentlyclosed/RecentlyClosedController.kt
@@ -21,18 +21,27 @@ import org.mozilla.fenix.browser.browsingmode.BrowsingMode
 import org.mozilla.fenix.components.FenixSnackbar
 import org.mozilla.fenix.ext.navigateBlockingForAsyncNavGraph
 
+@Suppress("TooManyFunctions")
 interface RecentlyClosedController {
-    fun handleOpen(item: RecoverableTab, mode: BrowsingMode? = null)
-    fun handleDeleteOne(tab: RecoverableTab)
+    fun handleOpen(tab: RecoverableTab, mode: BrowsingMode? = null)
+    fun handleOpen(tabs: Set<RecoverableTab>, mode: BrowsingMode? = null)
+    fun handleDelete(tab: RecoverableTab)
+    fun handleDelete(tabs: Set<RecoverableTab>)
     fun handleCopyUrl(item: RecoverableTab)
-    fun handleShare(item: RecoverableTab)
+    fun handleShare(tab: RecoverableTab)
+    fun handleShare(tabs: Set<RecoverableTab>)
     fun handleNavigateToHistory()
     fun handleRestore(item: RecoverableTab)
+    fun handleSelect(tab: RecoverableTab)
+    fun handleDeselect(tab: RecoverableTab)
+    fun handleBackPressed(): Boolean
 }
 
+@Suppress("TooManyFunctions")
 class DefaultRecentlyClosedController(
     private val navController: NavController,
-    private val store: BrowserStore,
+    private val browserStore: BrowserStore,
+    private val recentlyClosedStore: RecentlyClosedFragmentStore,
     private val tabsUseCases: TabsUseCases,
     private val resources: Resources,
     private val snackbar: FenixSnackbar,
@@ -40,12 +49,30 @@ class DefaultRecentlyClosedController(
     private val activity: HomeActivity,
     private val openToBrowser: (item: RecoverableTab, mode: BrowsingMode?) -> Unit
 ) : RecentlyClosedController {
-    override fun handleOpen(item: RecoverableTab, mode: BrowsingMode?) {
-        openToBrowser(item, mode)
+    override fun handleOpen(tab: RecoverableTab, mode: BrowsingMode?) {
+        openToBrowser(tab, mode)
     }
 
-    override fun handleDeleteOne(tab: RecoverableTab) {
-        store.dispatch(RecentlyClosedAction.RemoveClosedTabAction(tab))
+    override fun handleOpen(tabs: Set<RecoverableTab>, mode: BrowsingMode?) {
+        recentlyClosedStore.dispatch(RecentlyClosedFragmentAction.DeselectAll)
+        tabs.forEach { tab -> handleOpen(tab, mode) }
+    }
+
+    override fun handleSelect(tab: RecoverableTab) {
+        recentlyClosedStore.dispatch(RecentlyClosedFragmentAction.Select(tab))
+    }
+
+    override fun handleDeselect(tab: RecoverableTab) {
+        recentlyClosedStore.dispatch(RecentlyClosedFragmentAction.Deselect(tab))
+    }
+
+    override fun handleDelete(tab: RecoverableTab) {
+        browserStore.dispatch(RecentlyClosedAction.RemoveClosedTabAction(tab))
+    }
+
+    override fun handleDelete(tabs: Set<RecoverableTab>) {
+        recentlyClosedStore.dispatch(RecentlyClosedFragmentAction.DeselectAll)
+        tabs.forEach { tab -> handleDelete(tab) }
     }
 
     override fun handleNavigateToHistory() {
@@ -64,10 +91,13 @@ class DefaultRecentlyClosedController(
         }
     }
 
-    override fun handleShare(item: RecoverableTab) {
+    override fun handleShare(tab: RecoverableTab) = handleShare(setOf(tab))
+
+    override fun handleShare(tabs: Set<RecoverableTab>) {
+        val shareData = tabs.map { ShareData(url = it.url, title = it.title) }
         navController.navigateBlockingForAsyncNavGraph(
             RecentlyClosedFragmentDirections.actionGlobalShareFragment(
-                data = arrayOf(ShareData(url = item.url, title = item.title))
+                data = shareData.toTypedArray()
             )
         )
     }
@@ -75,12 +105,21 @@ class DefaultRecentlyClosedController(
     override fun handleRestore(item: RecoverableTab) {
         tabsUseCases.restore(item)
 
-        store.dispatch(
+        browserStore.dispatch(
             RecentlyClosedAction.RemoveClosedTabAction(item)
         )
 
         activity.openToBrowser(
             from = BrowserDirection.FromRecentlyClosed
         )
+    }
+
+    override fun handleBackPressed(): Boolean {
+        return if (recentlyClosedStore.state.selectedTabs.isNotEmpty()) {
+            recentlyClosedStore.dispatch(RecentlyClosedFragmentAction.DeselectAll)
+            true
+        } else {
+            false
+        }
     }
 }
--- a/app/src/main/java/org/mozilla/fenix/library/recentlyclosed/RecentlyClosedFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/library/recentlyclosed/RecentlyClosedFragment.kt
@@ -7,6 +7,7 @@ package org.mozilla.fenix.library.recentlyclosed
 import android.content.ClipboardManager
 import android.content.Context
 import android.os.Bundle
+import android.text.SpannableString
 import android.view.LayoutInflater
 import android.view.Menu
 import android.view.MenuInflater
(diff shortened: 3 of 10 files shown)


app/src/main/java/org/mozilla/fenix/library/recentlyclosed/RecentlyClosedAdapter.kt
```diff
@@ -31,6 +36,38 @@ class RecentlyClosedAdapter(
             oldItem.id == newItem.id
 
         override fun areContentsTheSame(oldItem: RecoverableTab, newItem: RecoverableTab) =
-            oldItem.id == newItem.id
+            oldItem == newItem
+    }
+
+    fun updateData(tabs: List<RecoverableTab>, selectedTabs: Set<RecoverableTab>) {
```
jonalmeida, line 42: nit: Let's organize the methods so that private bindings are together, overridden methods are together, and private methods are together.

app/src/main/java/org/mozilla/fenix/library/recentlyclosed/RecentlyClosedController.kt
```diff
@@ -22,30 +22,55 @@ import org.mozilla.fenix.components.FenixSnackbar
...
     private val tabsUseCases: TabsUseCases,
     private val resources: Resources,
     private val snackbar: FenixSnackbar,
     private val clipboardManager: ClipboardManager,
     private val activity: HomeActivity,
     private val openToBrowser: (item: RecoverableTab, mode: BrowsingMode?) -> Unit
 ) : RecentlyClosedController {
-    override fun handleOpen(item: RecoverableTab, mode: BrowsingMode?) {
-        openToBrowser(item, mode)
+    override fun handleOpen(tab: RecoverableTab, mode: BrowsingMode?) = openToBrowser(tab, mode)
```
jonalmeida, line 50: Let's put the brackets back onto this method if even though it's a single line method.

app/src/main/java/org/mozilla/fenix/library/recentlyclosed/RecentlyClosedFragment.kt
```diff
@@ -30,33 +32,63 @@ import org.mozilla.fenix.components.FenixSnackbar
...
         super.onResume()
         showToolbar(getString(R.string.library_recently_closed_tabs))
     }
 
     override fun onCreateOptionsMenu(menu: Menu, inflater: MenuInflater) {
-        inflater.inflate(R.menu.library_menu, menu)
+        if (recentlyClosedFragmentStore.state.selectedTabs.isNotEmpty()) {
+            inflater.inflate(R.menu.history_select_multi, menu)
+            menu.findItem(R.id.delete_history_multi_select)?.title =
+                SpannableString(getString(R.string.bookmark_menu_delete_button)).apply {
```
jonalmeida, line 58: We should introduce a separate string for this instead of re-using the bookmarks one.

app/src/main/java/org/mozilla/fenix/library/recentlyclosed/RecentlyClosedFragmentStore.kt
```diff
@@ -41,5 +45,6 @@ private fun recentlyClosedStateReducer(
 ): RecentlyClosedFragmentState {
     return when (action) {
         is RecentlyClosedFragmentAction.Change -> state.copy(items = action.list)
+        is RecentlyClosedFragmentAction.ChangeSelection -> state.copy(selectedTabs = action.tabs)
```
jonalmeida, line 48: Like the other implementations of the selection holder, let's add the `Select/Deselect` logic to our store instead of the controller. Consistency should make it easier to refactor if we decide to update the way this logic works.

Have a look at either the Bookmarks/History/Downloads/TabsTray stores for how multi-select was implemented.

app/src/main/java/org/mozilla/fenix/library/recentlyclosed/RecentlyClosedAdapter.kt
```diff
@@ -9,17 +9,22 @@ import android.view.ViewGroup
 import androidx.recyclerview.widget.DiffUtil
 import androidx.recyclerview.widget.ListAdapter
 import mozilla.components.browser.state.state.recover.RecoverableTab
+import org.mozilla.fenix.library.SelectionHolder
```
jonalmeida, line 12: One thing you will realise, is that we've moved the SelectionHolder and SelectionInteractor to a higher package namespace in-order to re-use it in other components like tabs tray. Let's update these references.

jonalmeida (changes requested):
This looks great, brilliant work! I've left some review comments below how we can align the changes here with the selection implementation to our other library items.

app/src/main/java/org/mozilla/fenix/library/recentlyclosed/RecentlyClosedAdapter.kt
```diff
@@ -31,6 +36,38 @@ class RecentlyClosedAdapter(
...
+            }
+        }
+
+        submitList(tabs)
+    }
+
+    override val selectedItems: Set<RecoverableTab>
+        get() = selectedTabs
+
+    private fun toggledMultiSelectMode(
```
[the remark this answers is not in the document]
ZianeA: I am not sure that inMultiSelectModethe reflects what the function represents. This function is supposed to return true if the selection mode has been switched from on to off or the other way around. In other words, it should return true only if we have just entered or left the selection mode and return false if the selection mode has not changed.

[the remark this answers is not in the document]
jonalmeida: Hmm.. maybe something changed in your local branch that I can't see?

This method only has this state check which is why I recommended the renaming:
```kotlin
currentlySelectedTabs.isEmpty() && selectedTabs.isNotEmpty() || currentlySelectedTabs.isNotEmpty() && selectedTabs.isEmpty()
```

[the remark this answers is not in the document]
ZianeA: Checking if we are in multi-selection mode would look like this:
`selectedTabs.isNotEmpty()`

However, the function is doing more:

```
// This checks if we have entered the multi-selection mode
currentlySelectedTabs.isEmpty() && selectedTabs.isNotEmpty()

// This Checks if we have left the multi-selection mode
currentlySelectedTabs.isNotEmpty() && selectedTabs.isEmpty()
```

The parameters names are not great, maybe I should rename them to this:
currentlySelectedTabs -> currentSelection
selectedTabs -> newSelection

Alternatively, I can greatly simplify the logic by updating all the items (there can only be 10 items in total) whenever the selection changes:

```
fun updateData(tabs: List<RecoverableTab>, selectedTabs: Set<RecoverableTab>) {
        this.selectedTabs = selectedTabs
        notifyItemRangeChanged(0, tabs.size)
        submitList(tabs)
 }
```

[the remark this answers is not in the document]
jonalmeida: Sure, that works too. :)

app/src/main/java/org/mozilla/fenix/library/recentlyclosed/RecentlyClosedFragment.kt
```diff
@@ -30,33 +32,63 @@ import org.mozilla.fenix.components.FenixSnackbar
...
+    private lateinit var recentlyClosedController: RecentlyClosedController
 
     override fun onResume() {
         super.onResume()
         showToolbar(getString(R.string.library_recently_closed_tabs))
     }
 
     override fun onCreateOptionsMenu(menu: Menu, inflater: MenuInflater) {
-        inflater.inflate(R.menu.library_menu, menu)
+        if (recentlyClosedFragmentStore.state.selectedTabs.isNotEmpty()) {
```
jonalmeida, line 55: Sorry, I must have missed this in the previous iteration.

I don't think we need to add a menu here since the parent `LibraryPageFragment` handles the overlay that shows up with the menu built in.

In fact, I think we can entirely remove `onCreateOptionsMenu` and `onOptionsItemSelected` since that pattern is inconsistent with our other "library" views.

app/src/main/java/org/mozilla/fenix/library/recentlyclosed/RecentlyClosedFragmentView.kt
```diff
@@ -57,11 +59,11 @@ interface RecentlyClosedInteractor {
...
 
     /**
-     * Deletes one recently closed tab item.
+     * Called when recently closed tab is selected for deletion.
      *
-     * @param tab the recently closed tab item to delete.
+     * @param tab the recently closed tab to delete.
      */
-    fun onDeleteOne(tab: RecoverableTab)
+    fun onDelete(tab: RecoverableTab)
```
jonalmeida, line 66: Excellent clean up! ✨

app/src/main/res/layout/component_recently_closed.xml
```diff
@@ -18,7 +18,8 @@
     <androidx.recyclerview.widget.RecyclerView
         android:id="@+id/recently_closed_list"
         android:layout_width="match_parent"
-        android:layout_height="wrap_content"
+        android:layout_height="0dp"
+        app:layout_constraintBottom_toBottomOf="parent"
```
jonalmeida, line 22: I was just wondering if was too much to ask to fix this issue since this PR is already in this area. 😄

jonalmeida (commented):
This looks great! I left one comment that I think we can address before we land this.

app/src/main/java/org/mozilla/fenix/library/recentlyclosed/RecentlyClosedFragment.kt
ZianeA: I tried to remove the `onCreateOptionsMenu` and `onOptionsItemSelected` methods, but it didn't work. I have checked `LibraryPageFragment`, from what I can see it doesn't seem to handle the options menu. I also looked at `BookmakrFragment`, `DownloadFragment`, `HistoryFragment` and they all override the two aforementioned methods. 

On the other hand, the `LibraryPageView` does handle updating the toolbar color in selection mode.

app/src/main/res/layout/component_recently_closed.xml
ZianeA: Of course! Are you referring to the weird animation when deleting items in the **bookmark** screen?

app/src/main/java/org/mozilla/fenix/library/recentlyclosed/RecentlyClosedFragment.kt
jonalmeida: You're right, it looks like the 'history' view is the one that is inconsistent; that's a separate issue then. 👌

app/src/main/res/layout/component_recently_closed.xml
jonalmeida: I think this change solves the problem I was seeing with the bottom RecyclerView cutting off the overscroll animation. 🙂 

Thanks!

jonalmeida:
Looks great! If you could squash these commits, I'll run CI on it next. :)

jonalmeida:
In case you have trouble finding the results of the `detekt` failures from CI:

```
981 kotlin files were analyzed.
/home/runner/work/fenix/fenix/app/src/main/java/org/mozilla/fenix/library/recentlyclosed/RecentlyClosedController.kt - 50min debt
	TooManyFunctions - 12/11 - [RecentlyClosedController] at /home/runner/work/fenix/fenix/app/src/main/java/org/mozilla/fenix/library/recentlyclosed/RecentlyClosedController.kt:24:11
	TooManyFunctions - 12/11 - [DefaultRecentlyClosedController] at /home/runner/work/fenix/fenix/app/src/main/java/org/mozilla/fenix/library/recentlyclosed/RecentlyClosedController.kt:39:7
	MozillaNavigateCheck - [handleNavigateToHistory] at /home/runner/work/fenix/fenix/app/src/main/java/org/mozilla/fenix/library/recentlyclosed/RecentlyClosedController.kt:77:23

Overall debt: 50min

MozillaNavigateCheck:
    /home/runner/work/fenix/fenix/app/src/main/java/org/mozilla/fenix/library/recentlyclosed/RecentlyClosedController.kt
    If you named a method `navigate`, please rename it to something a bitmore specific such as `navigateTo...`. However, if you are trying to invoke the Android NavController.navigateplease use `org.mozilla.fenix.ext.NavController.navigateBlockingForAsyncNavGraph`instead. Because using `navigate` directly in the code can lead to the NavGraph not being loadedsince it relies on a blocking call done in `navigateBlockingForAsyncNavGraph`.
```
I think we can suppress the `TooManyFunctions` one since I can't see any quick alternative now that would work for this patch.

app/src/main/res/layout/component_recently_closed.xml
ZianeA: I hadn't even noticed the over scroll problem. I'm glad this solves it. 

However, I was referring to a different problem. It's very subtle, but notice how the last item disappears and reappears when the second-to-last item is removed. 

https://user-images.githubusercontent.com/14791787/119891857-62fd4400-bf31-11eb-8a38-0e26a35e7868.mp4

This is the expected behavior:

https://user-images.githubusercontent.com/14791787/119891754-44974880-bf31-11eb-9004-923d5fcdd968.mp4

jonalmeida: I think this is the cause of `notifyItemRangeChanged`. Using `submitList` should be enough with the `ListAdapter` to know and calculate which items to animate. Nice catch!

jonalmeida:
@ZianeA Just the detekt warnings above that need to be addressed. 👌

app/src/main/res/layout/component_recently_closed.xml
ZianeA: This problem was actually caused by the fact that the Recyclerview height was set to wrap_content. The bookmark screen also has this problem.

jonalmeida: Fair point! Let's remove the extra notify though, it's definitely doing double work there.

From the [`submitList`](https://developer.android.com/reference/androidx/recyclerview/widget/ListAdapter#submitList(java.util.List%3CT%3E)) docs you can see that they do the notifying for you:

> Submits a new list to be diffed, and displayed.
> If a list is already being displayed, a diff will be computed on a background thread, which will dispatch Adapter.notifyItem events on the main thread.

jonalmeida:
@ZianeA 

We're still seeing detekt lint failures. Let me know if you need help with fixing them:

```
 983 kotlin files were analyzed.
/home/runner/work/fenix/fenix/app/src/main/java/org/mozilla/fenix/library/recentlyclosed/RecentlyClosedController.kt - 40min debt
	TooManyFunctions - 12/11 - [RecentlyClosedController] at /home/runner/work/fenix/fenix/app/src/main/java/org/mozilla/fenix/library/recentlyclosed/RecentlyClosedController.kt:24:11
	TooManyFunctions - 12/11 - [DefaultRecentlyClosedController] at /home/runner/work/fenix/fenix/app/src/main/java/org/mozilla/fenix/library/recentlyclosed/RecentlyClosedController.kt:39:7

Overall debt: 40min

```

app/src/main/res/layout/component_recently_closed.xml
ZianeA: Diff only checks the submitted list of `RecoverableTab`s. It does not, however, check the list of selected tabs for changes. Therefore, notifyItemRangeChanged must be called to update the RecyclerView whenever the selection has changed. 

For this to work with diff, the selection must be integrated into the `RecoverableTab` model, for example by adding an `isSelected` flag.

But, as you mentioned, it's still doing double work.

jonalmeida: Hmm yeah, that's a good point. Thank you for clarifying that for me. :)

ZianeA:
@jonalmeida I'm not sure how to fix the detekt lint failures :sweat:

jonalmeida:
No worries! I think you can follow my comment above (https://github.com/mozilla-mobile/fenix/pull/18803#issuecomment-849916887) to just suppress the warning with a patch like this:

```diff
diff --git a/app/src/main/java/org/mozilla/fenix/library/recentlyclosed/RecentlyClosedController.kt b/app/src/main/java/org/mozilla/fenix/library/recentlyclosed/RecentlyClosedController.kt
index 47f1b2935..7cb442567 100644
--- a/app/src/main/java/org/mozilla/fenix/library/recentlyclosed/RecentlyClosedController.kt
+++ b/app/src/main/java/org/mozilla/fenix/library/recentlyclosed/RecentlyClosedController.kt
@@ -21,6 +21,7 @@ import org.mozilla.fenix.browser.browsingmode.BrowsingMode
 import org.mozilla.fenix.components.FenixSnackbar
 import org.mozilla.fenix.ext.navigateBlockingForAsyncNavGraph

+@Suppress("TooManyFunctions")
 interface RecentlyClosedController {
     fun handleOpen(item: RecoverableTab, mode: BrowsingMode? = null)
     fun handleDeleteOne(tab: RecoverableTab)
@@ -30,6 +31,7 @@ interface RecentlyClosedController {
     fun handleRestore(item: RecoverableTab)
 }

+@Suppress("TooManyFunctions")
 class DefaultRecentlyClosedController(
     private val navController: NavController,
     private val store: BrowserStore,

```

ZianeA:
@jonalmeida Thanks for your help! This should work now :crossed_fingers:

jonalmeida (approved):
Brilliant! ✨

PR #18817, opened 2021-04-06
Bug 1609191 - Disable hardware acceleration on devices affected by Adreno shader compilation crash

jamienicol:
On various Adreno devices we have experienced crashes in glLinkProgram
due to a suspected driver bug. There appears to be a race condition
between the gecko renderer thread compiling shaders used to render web
content, and the Android UI render thread compiling shaders used to
render the Fenix UI. Disabling hardware acceleration avoids this crash
as shader compilation will only occur on the gecko renderer thread.

Hardware acceleration must be disabled globally in
AndroidManifest.xml.  We can then opt in to hardware acceleration at
runtime on a per-window basis on suitable devices. We do this by
creating an ActivityLifecycleCallbacks which enables hardware
acceleration for the window whenever an activity is created.

We use a hardcoded list of known-buggy device boards, and only enable
hardware acceleration if the device is not in that list. The list was
chosen by selecting the most common boards that this crash occurs
on (covering >90% of crashes in the past month). Some crashes may
still occur on boards not in this list, however the numbers will be
low and crashes infrequent enough that this should not matter.
Likewise, some devices with these boards may be running driver
versions unaffected by this bug, but it should not matter if hardware
acceleration is disabled anyway.

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/session/HardwareAccelerationLifecycleCallbacks.kt
+++ b/app/src/main/java/org/mozilla/fenix/session/HardwareAccelerationLifecycleCallbacks.kt
@@ -1,0 +1,49 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.session
+
+import android.app.Activity
+import android.app.Application
+import android.os.Build
+import android.os.Bundle
+import android.view.WindowManager.LayoutParams.FLAG_HARDWARE_ACCELERATED
+
+/**
+ * This ActivityLifecycleCallbacks implementation ensures that hardware acceleration is enabled for
+ * a Window when an Activity is created. On some devices we cannot enable UI hardware acceleration
+ * due to graphics driver issues. So we must globally disable hardware acceleration in
+ * AndroidManifest.xml, then enable it where possible on a per-window basis.
+ * See https://bugzilla.mozilla.org/show_bug.cgi?id=1609191
+ */
+@SuppressWarnings("EmptyFunctionBlock")
+class HardwareAccelerationLifecycleCallbacks : Application.ActivityLifecycleCallbacks {
+    val supportsHardwareAcceleration by lazy {
+        // List of boards on which we should keep hardware acceleration disabled.
+        // These were the boards with the highest number of crashes caused by hardware acceleration being enabled.
+        // There will be some affected devices with different boards which therefore may still crash,
+        // however the numbers should be low enough and the crashes infrequent enough for that to be okay.
+        val buggyBoards = listOf("msm8953", "msm8976", "msm8937", "msm8996", "sdm450_mh4x")
+
+        Build.BOARD !in buggyBoards
+    }
+
+    override fun onActivityCreated(activity: Activity, savedInstanceState: Bundle?) {
+        if (supportsHardwareAcceleration) {
+            activity.window.setFlags(FLAG_HARDWARE_ACCELERATED, FLAG_HARDWARE_ACCELERATED)
+        }
+    }
+
+    override fun onActivityStarted(activity: Activity) {}
+
+    override fun onActivityResumed(activity: Activity) {}
+
+    override fun onActivityPaused(activity: Activity) {}
+
+    override fun onActivityStopped(activity: Activity) {}
+
+    override fun onActivitySaveInstanceState(activity: Activity, outState: Bundle) {}
+
+    override fun onActivityDestroyed(activity: Activity) {}
+}
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -32,6 +32,7 @@
         android:supportsRtl="true"
         android:theme="@style/NormalTheme"
         android:usesCleartextTraffic="true"
+        android:hardwareAccelerated="false"
         tools:ignore="UnusedAttribute">
 
         <!--
--- a/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
+++ b/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
@@ -57,6 +57,7 @@ import org.mozilla.fenix.perf.StorageStatsMetrics
 import org.mozilla.fenix.perf.runBlockingIncrement
 import org.mozilla.fenix.push.PushFxaIntegration
 import org.mozilla.fenix.push.WebPushEngineIntegration
+import org.mozilla.fenix.session.HardwareAccelerationLifecycleCallbacks
 import org.mozilla.fenix.session.PerformanceActivityLifecycleCallbacks
 import org.mozilla.fenix.session.VisibilityLifecycleCallback
 import org.mozilla.fenix.utils.BrowsersCache
@@ -77,6 +78,9 @@ open class FenixApplication : LocaleAwareApplication(), Provider {
     open val components by lazy { Components(this) }
 
     var visibilityLifecycleCallback: VisibilityLifecycleCallback? = null
+        private set
+
+    var hardwareAccelerationLifecycleCallbacks: HardwareAccelerationLifecycleCallbacks? = null
         private set
 
     override fun onCreate() {
@@ -183,6 +187,9 @@ open class FenixApplication : LocaleAwareApplication(), Provider {
 
             visibilityLifecycleCallback = VisibilityLifecycleCallback(getSystemService())
             registerActivityLifecycleCallbacks(visibilityLifecycleCallback)
+
+            hardwareAccelerationLifecycleCallbacks = HardwareAccelerationLifecycleCallbacks()
+            registerActivityLifecycleCallbacks(hardwareAccelerationLifecycleCallbacks)
 
             // Storage maintenance disabled, for now, as it was interfering with background migrations.
             // See https://github.com/mozilla-mobile/fenix/issues/7227 for context.


jamienicol:
Not sure if this requires tests, please point me to where to add them if it does. This doesn't affect anything to do with screenshots or accessibility

jamienicol:
I was speaking to @jmathies about this and we feel this may require some time to be tested for side-effects caused by disabling hardware acceleration. (We're unsure about how the fenix release process works, is it like gecko's trains, or if a PR gets merged does it get released soon after?)

Things I have noticed with hardware acceleration disabled:
* The hamburger menu doesn't have a nice drop shadow around it
* When swiping the URL bar to switch tabs the tab content lags slightly behind the drag

To me both of those are preferable to crashing, but some testing to make sure there aren't other issues couldn't hurt, and at the end of the day it's the Fenix team's call.

Amejia481:
> 

> We're unsure about how the fenix release process works, is it like gecko's trains, or if a PR gets merged does it get released soon after?)

It's similar to desktop, if it gets merged now it will get on the next nightly and will ride the trains to Beta -> Release.

eliserichards:
Since there's a guarantee of side effects we definitely don't want to uplift this, I think it should ride the trains 🚂

pocmo (commented):
:+1: (Just talked to Arturo)

Saw that Chromium does something similar:
* https://source.chromium.org/chromium/chromium/src/+/master:chrome/android/java/src/org/chromium/chrome/browser/app/ChromeActivity.java;l=2452-2474;drc=4770965e365681d372956ba49c01a44a2abb9b9b
* https://source.chromium.org/chromium/chromium/src/+/master:base/android/java/src/org/chromium/base/SysUtils.java;l=163-196;drc=db0bc5c31204c9c7ab2f131af6fe2cf193e08876

eliserichards:
Tagging @mcomella for performance

app/src/main/java/org/mozilla/fenix/session/HardwareAccelerationLifecycleCallbacks.kt
```diff
@@ -0,0 +1,49 @@
...
+        // however the numbers should be low enough and the crashes infrequent enough for that to be okay.
+        val buggyBoards = listOf("msm8953", "msm8976", "msm8937", "msm8996", "sdm450_mh4x")
+
+        Build.BOARD !in buggyBoards
+    }
+
+    override fun onActivityCreated(activity: Activity, savedInstanceState: Bundle?) {
+        if (supportsHardwareAcceleration) {
+            activity.window.setFlags(FLAG_HARDWARE_ACCELERATED, FLAG_HARDWARE_ACCELERATED)
+        }
```
Amejia481, line 35: @pocmo discovered an interesting comment on the [Chrome's implementation](https://source.chromium.org/chromium/chromium/src/+/master:chrome/android/java/src/org/chromium/chrome/browser/app/ChromeActivity.java;l=2452-2474;drc=4770965e365681d372956ba49c01a44a2abb9b9b). I think it's useful to add it here for references, and give it a try to see if dialogs have any issues, a way to test dialogs are trying to download a file, opening any page with js prompts like [alerts](https://www.w3schools.com/jsref/met_win_alert.asp)  [confirm](https://www.w3schools.com/jsref/met_win_confirm.asp) and others. 

> 
> 
>             // When HW acceleration is enabled manually for an activity, child windows (e.g.
>             // dialogs) don't inherit HW acceleration state. However, when HW acceleration is
>             // enabled in the manifest, child windows do inherit HW acceleration state. That
>             // looks like a bug, so I filed b/23036374
>             //
>             // In the meanwhile the workaround is to call
>             //   window.setWindowManager(..., hardwareAccelerated=true)
>             // to let the window know that it's HW accelerated. However, since there is no way
>             // to know 'appToken' argument until window's view is attached to the window (!!),
>             // we have to do the workaround in onAttachedToWindow()

Amejia481: I tried to access the bug they mentioned on the on the comment https://issuetracker.google.com/issues/23036374 but it is not publicly available.

mcomella:
> Tagging @mcomella for performance

Thanks – we'll discuss who will review in triage tomorrow.

app/src/main/java/org/mozilla/fenix/session/HardwareAccelerationLifecycleCallbacks.kt
jamienicol: Dialogs from alerts and and downloads look exactly the same to me in all 3 scenarios:
a) No patch - hardware acceleration enabled by default
b) With patch, unaffected device - hardware acceleration enabled programatically
c) With patch, affected device - hardware acceleration disabled

That's not to say the bug doesn't still exist - It could be the case that hardware acceleration is indeed not being enabled for the dialogs in case b). But I wouldn't think it matters much if those are our only uses of dialogs.

I did initially implement this by creating a base Activity class and overriding `onCreate()` there, as I didn't know that `ActivityLifecycleCallbacks` existed. But when I stumbled across it in the docs I thought it was better to use it, as it means we don't have to derive all of Fenix's activities from a single class. If we want to do what chrome does we'll need to derive every activity from a single one (or duplicate some code in every activity) that so we can override `Activity.onAttachedToWindow()`

jonalmeida:
Would this be useful to have in components so that we can share it with Focus too?

app/src/main/java/org/mozilla/fenix/session/HardwareAccelerationLifecycleCallbacks.kt
Amejia481: > Dialogs from alerts and and downloads look exactly the same to me in all 3 scenarios:
> a) No patch - hardware acceleration enabled by default
> b) With patch, unaffected device - hardware acceleration enabled programatically
> c) With patch, affected device - hardware acceleration disabled

I think we can keep the actual implementation and add the comment for future references.

> That's not to say the bug doesn't still exist - It could be the case that hardware acceleration is indeed not being enabled for the dialogs in case b). But I wouldn't think it matters much if those are our only uses of dialogs.

There are other usages of dialogs across Fenix, but the above are the main ones.


> I did initially implement this by creating a base Activity class and overriding onCreate() there, as I didn't know that ActivityLifecycleCallbacks existed. But when I stumbled across it in the docs I thought it was better to use it, as it means we don't have to derive all of Fenix's activities from a single class. If we want to do what chrome does we'll need to derive every activity from a single one (or duplicate some code in every activity) that so we can override Activity.onAttachedToWindow()

This is clean solution, but if it's strictly needed, having a based activity will not be a big issue as Fenix is following [a single activity approach](https://www.youtube.com/watch?v=2k8x8V77CrU) for navigation ([HomeActivity.kt](https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/java/org/mozilla/fenix/HomeActivity.kt)), in most of the app.

Amejia481:
> Would this be useful to have in components so that we can share it with Focus too?

Yes, the idea is to monitor it on nightly and if everything goes well uplift it to AC.

Amejia481 (approved):

MarcLeclair:
I have a phone with that board ( the Moto g5s plus has the msm8953 board). @jamienicol , disabling the HW accelartion globally makes the app janky in its animation / transition. Is it possible to disable it only for GV? or does it have to be enable globally?

jamienicol:
To avoid this crash we _either_ have to disable acceleration for UI or for web content (GV). This patch only disables it for the UI and leaves web content accelerated. While the UI jank it causes is unfortunate, the thought process is that jank in web content would be worse for users. If you want to compare what disabling it for content only would be like then you can set `gfx.webrender.software` to true in about:config in Fenix nightly. (Warning: this may be buggy as it's still in development)

I've been trying to get in touch with Qualcomm to find a better fix but haven't had any luck so far. I have some ideas myself but they may not work and will take some time to prototype.

At the end of the day I guess it's a fenix product call which tradeoff is better (accelerated UI vs web content). It may be that living with the crashes is the best option, given that we believe they only happen once per user after each Firefox update.

MarcLeclair:
Interesting. ( Sorry for all the questions, I'm just trying to get a good understanding of the problem!) 

1) I saw from Bugzilla the bug was created a year ago. Is this crash occurring more often now? If so, do we know what happened?  Also, would it be possible (for now) to use an older implementation 
2) If this is caused by a race condition, can we do  things sequentially? i.e. avoid the overlap and avoid the race condition?

kbrosnan:
1. The crash was found by people testing webrender unofficially. Webrender was  turned on for Adreno  505/506 in bz-1693072.

jamienicol:
@MarcLeclair please ask away! I'm hopeful we can find a better solution!

1. As Kevin says, the numbers spiked when webrender was enabled for Adreno 505 and 506 devices on release. Prior to that it was just enabled for those devices on Nightly (or users who manually enabled it). There were also some Adreno 530 devices running it as we didn't realize it was affected at the time.

2. If we can figure out how to do that, it would be ideal.

Ideally we could do so at quite a low level, eg a mutex around `glLinkProgram()`. Obviously we have full control around the webrender renderer thread, but I'm not sure we have a way to hook in to the UI render thread like this. On Android 11 we _might_ be able to use a [GLES layer](https://developer.android.com/ndk/guides/rootless-debug-gles), but for earlier android versions that requires the app to be _debuggable_, which we don't want.

At a higher level, we might be able to coordinate somewhat, at least enough to make the crash less likely. Webrender by default compiles shaders on-demand when it requires one to render some content. I assume the android UI works the same way. Perhaps delaying initializing webrender at all until the Fenix UI has been fully rendered could work, but there may be a risk if the user opens a new UI element while webrender is compiling shaders.

Alternatively we can make webrender compile all shaders up front. This is a time consuming process (100s of milliseconds if not a few seconds), but only has to be done once after each app upgrade as we cache the result. Perhaps we can prevent Fenix rendering any UI in this time, or maybe start with the Fenix UI unaccelerated, then enable acceleration once we are certain webrender is done compiling shaders.

Part of the problem is that I have been completely unable to "organically" reproduce this crash, so I'm unsure what state UI is in that causes it. eg is it when the user opens a new activity in Fenix while webrender is rendering? I can only reproduce if I enable the debug option to make webrender compile everything on initialization, then the crash occurs just by opening the app.

Another thought I had was that we could create a dedicated process to compile webrender's shaders, and do that all up front in the background. It's not clear what we'd do in this case if the user opened the app and webrender requires one of these shaders before the process has finished (or if loading the cached shader fails and we need to compile a new one).

Lastly, if geckoview had a dedicated GPU process on android that would solve the problem for us. But I imagine this is a lot of work, and I'm not sure whether it's planned for android at all.

Mikaka27:
There is this ticket, but it is quite old: https://bugzilla.mozilla.org/show_bug.cgi?id=1331109

mcomella (changes requested):
Rejecting on half of performance. AIUI, MarcLeclair looked into this with jamienicol and found that disabling Android UI hardware acceleration (on a large number (hundreds?) of device models) was too much of a performance decrease that we should look for an alternative to this change: hence requesting changes on this review.

iirc, the proposed solution was disabling gecko hardware acceleration on first run and then enable it on the second to avoid the shader compilation race.

mcomella:
It looks like we solved this in BZ-1609191 with a different approach – we disabled WebRender (Gecko hardware acceleration) on affected devices rather than disabling Android UI hardware acceleration on those devices (this PR). I think that means this PR can be closed. @jamienicol please reopen if that's not the case!

Issue #18608, opened 2021-03-23
Fenix Toolbar Menu Experiment: (Test branch 1: Menu includes Default Browser as a visible option, Control: Default Browser not an option)


ItsDevyani:
## User Story
- As a user, I want to know how I can easily set up Fenix as my default browser when I click the Toolbar menu.

## Dependencies

   

-  [Toolbar Menu Test design option (done)](https://www.figma.com/file/YvvFxwXKonLkvdF0ZcL63Y/Design-feedback?node-id=0%3A1)
-  Nimbus SDK integration with Fenix - [Implementation guide](https://mozilla.github.io/experimenter-docs/fenix-engineers/)
- DS resources to write measurement plan
- Engineering Time

### Acceptance Criteria    

- Set Default Browser as a visible option on the Toolbar menu.
- We can track default browser activations coming from this source full-funnel (Clicks at each step) including conversion to default.
- Nimbus SDK integration with Toolbar menu to enable testing on it.

### Engineering notes
- [Nimbus engineering docs](https://mozilla.github.io/experimenter-docs/fenix-engineers/)
- Please sync with data science before starting this implementation to confirm the following: 

**Experiment length**: TBD (default: 35 days total, 7 day enrollment period)
**Start date**: TBD
**Experiment size**: TBD
**A/B split**: TBD (default: 50/50)
**Channels**: Release, Nightly
   * > all of the Nimbus experiments should be pushed to nightly first. [...]  Also, they don't need to go to beta afterwards but instead can go to production.

amedyne:
Note: This ticket will be for tracking the Engineering tasks for developing Experiment **Test Branch 1: Menu includes Default Browser** as a visible option and the corresponding telemetry needed for this branch.

Issue #18796, opened 2021-04-05
Nimbus + Fenix Default Browser Experiment


ItsDevyani:
## User Story
- As a user, I want to know how I can easily set up Fenix as my default browser 

## Dependencies
- [Design (Done)](https://www.figma.com/file/YvvFxwXKonLkvdF0ZcL63Y/Design-feedback?node-id=0%3A1)
- [Nimbus Integration](https://mozilla.github.io/experimenter-docs/fenix-engineers/ ) (scheduled in 88)
- Engineering time
- DS time

### Acceptance Criteria
-  Set Default Browser as a visible option on the Toolbar menu, Settings Menu and New Tab banner branches are developed.
- We can track default browser activations coming from these source full-funnel (Clicks at each step) including conversion to default.
- Nimbus SDK integration to enable testing.

###Additional Files
[Nimbus Experiment Brief](https://docs.google.com/document/d/1vyJSPvK5_CoFGEAFZfZhJqYLrnWERpj-ulva28AVst8/edit#heading=h.kzffe2a93ndd)
Related Github Issues:
[Fenix Toolbar Menu Experiment](https://github.com/mozilla-mobile/fenix/issues/18608)
[Fenix Settings Menu Experiment](https://github.com/mozilla-mobile/fenix/issues/18376)
[Fenix New Tab Experiment](https://github.com/mozilla-mobile/fenix/issues/18375)

PR #18844, opened 2021-04-07
For #18608 Add experiment for set default browser MenuItem.

Amejia481:
### Pull Request checklist

1. click on Show All Checks,


Amejia481:
👋🏽  @travis79 would you mind taking a look to this PR, this is my first time working with a Nimbus experiment and I would like to see if I'm on the right track, I'm working on the ticket https://github.com/mozilla-mobile/fenix/issues/18608, thanks in advance.

app/src/main/java/org/mozilla/fenix/experiments/Experiments.kt
```diff
@@ -18,5 +19,6 @@ class ExperimentBranch {
         const val CONTROL = "control"
         const val A1 = "a1"
         const val A2 = "a2"
+        const val BRANCH1 = "branch1"
```
travis79, line 22: You can reuse the `control` and `treatment` if you like.  Those are the default branch names in Experimenter, also.  Unless you have more than 2 branches in your experiment, this would be preferable, or at least a more descriptive branch name maybe?

travis79 (approved):
Aside from the branch naming, this looks good to me!  Just to clarify, and to ensure I reviewed this from the correct perspective, the experiment will return the `BrowserMenuImageText` for the set as default if the user is in the treatment or `BRANCH1` group, or `null` if they are in the control group?

app/src/main/java/org/mozilla/fenix/experiments/Experiments.kt
travis79: Just to add to this, I would like to standardize on things here if possible so reusing branch names `treatment` and `control` for an A/B experiment like this is defintely my preferred option, and shouldn't cause any issues since they are being used with a different feature-id.

Amejia481:
> Aside from the branch naming, this looks good to me! Just to clarify, and to ensure I reviewed this from the correct perspective, the experiment will return the `BrowserMenuImageText` for the set as default if the user is in the treatment or `BRANCH1` group, or `null` if they are in the control group?

Thanks for the swift review, yes that's correct when the user is the experiment we want to show the menu item for setting the default browser (`BrowserMenuImageText`) otherwise we don't have to show the menu item (`null`)

app/src/main/java/org/mozilla/fenix/experiments/Experiments.kt
Amejia481: That make sense but I added a new branch name as it was part of the ticket description see https://github.com/mozilla-mobile/fenix/issues/18608#issuecomment-813439882 maybe @amedyne or @ItsDevyani  could help us to clarify.

amedyne: I would follow what the Nimbus team prefers. For this experiment, there will be 3 branches. Do you have a recommendation for a standard on how to do this @travis79 ?

travis79: > I would follow what the Nimbus team prefers. For this experiment, there will be 3 branches. Do you have a recommendation for a standard on how to do this @travis79 ?

In the case where you have more than a control and a treatment branch, then I would say that having them descriptively named is helpful.  I would defer the question to @tdsmith for further guidance on this to get a DS perspective, and maybe he's fine with `branch1` and `branch2`.

travis79: Also, I only saw 2 branch options, "branch1" and anything else.  Where is the 3rd branch?

amedyne: This evolved from initially 3 different experiments to then being combined to 1 experiment, see #18796 .

tdsmith: Hm; as an analyst I'd generally prefer descriptive names to generic names once we have more than two branches. The branch names currently show up in reporting, though we've already identified that that's not ideal. I realize descriptive names are annoying in the context of having to define new constants for each experiment, which isn't something other platforms have required, so it's sort of a novel question for me as well.

I'd have a weak preference for e.g. `settings_menu, `toolbar_menu`, `newtab_card` but if that's a non-starter then we can ride with branch[123].

gabrielluong: I think adding a prefix to indicate the associated experiment would also be appropriate here. Something like `default_browser_settings_menu` `default_browser_toolbar_menu`. Alternatively, a comment to indicate the constants are branch values for `DEFAULT_BROWSER_TOOLBAR_MENU`.

Amejia481: I updated the name to be `DEFAULT_BROWSER_TOOLBAR_MENU`,  please let me know if that it's OK.

travis79: Works for me, thanks @Amejia481!

Issue #18375, opened 2021-03-08
Fenix New Tab Experiment (Test branch 3: Default Browser Message to a new user on their third session, Control: No Message)


ItsDevyani:
## User Story
- As a user, I want to know how I can easily set up Fenix as my default browser once I've had a chance to use the browser a few times.

## Dependencies
- Test Measurement plan is being requested by the DS team
- Default Browser Dashboard is also being requested by the DS team
- Nimbus SDK integration with Fenix 
-[ UX design work is done](https://www.figma.com/file/YvvFxwXKonLkvdF0ZcL63Y/Design-feedback?node-id=0%3A1)
- Engineering Time

### Acceptance Criteria
- A Message/Notification can be triggered to a user on their third session with Fenix.
- A user can click the notification and be launched into the Default Browser flow.
- Funnel Activity (Clicks on each step) and Conversion to the default browser feature can be measured.
- Nimbus SDK integration with New Tab allowing experimentation on it

### Engineering notes
- [Nimbus engineering docs](https://mozilla.github.io/experimenter-docs/fenix-engineers/)
- Please sync with data science before starting this implementation to confirm the following:

**Experiment length**: TBD (default: 35 days total, 7 day enrollment period)
**Start date**: TBD
**Experiment size**: TBD
**A/B split**: TBD (default: 50/50)
**Channels**: Release, Nightly
   * > all of the Nimbus experiments should be pushed to nightly first. [...]  Also, they don't need to go to beta afterwards but instead can go to production.

ItsDevyani:
Project Update: 
- Nicole Weber has made the designs for this experiment and they are ready to be shared as needed.
- We have has discussions with the Nimbus Team to launch this as a Nimbus Experiment.

amedyne:
Note: This ticket will be for tracking the Engineering tasks for developing Experiment **Test Branch 3: Default Browser Message to a new user on their third session** and the corresponding telemetry needed for this branch.

amedyne:
Tagging @topotropic , too.

ItsDevyani:
Is this influencing the experiment?

_We are comparing three tactics against each other, this is one of those 3 tactics, so showing a banner on the new tab to a new user on their 3rd app open after installing the app is a factor that influences the experiment._

[...]

PR #18846, opened 2021-04-07
For #18608 made set a default browser functionality publicly available.

Amejia481:
For multiple experiments (https://github.com/mozilla-mobile/fenix/issues/18608 ,https://github.com/mozilla-mobile/fenix/issues/18376, https://github.com/mozilla-mobile/fenix/issues/18375)  we need a reusable behavior for setting a default browser, this PR make the functionality reusable. 
  
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/ext/Activity.kt
+++ b/app/src/main/java/org/mozilla/fenix/ext/Activity.kt
@@ -8,6 +8,14 @@ import android.app.Activity
 import android.view.View
 import android.view.WindowManager
 import mozilla.components.concept.base.crash.Breadcrumb
+import android.app.role.RoleManager
+import android.content.Intent
+import android.os.Build
+import android.provider.Settings
+import androidx.core.os.bundleOf
+import org.mozilla.fenix.BrowserDirection
+import org.mozilla.fenix.HomeActivity
+import org.mozilla.fenix.settings.SupportUtils
 
 /**
  * Attempts to call immersive mode using the View to hide the status bar and navigation buttons.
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/fenix/ext/Activity.kt
```diff
@@ -48,3 +56,61 @@ fun Activity.breadcrumb(
         )
     )
 }
+
+/**
+ * Opens the set default browser setting depending on the user OS version:
+ * for devices with Q or lower takes the user to the settings page
+ * and for other allow the user to set the default browser in app.
```
gabrielluong, line 63: I don't think the 3 conditions are quite right from what I am reading in the code. You mean Android Q+ right for the first point?
In the second point, I think that applies to Android N+, but the else condition actually takes ppl to a sumo page. Provided some suggestion but I think we can also change this kDoc to say:
"Opens Android's Manage Default Apps Settings if possible."

Anyways, I want to be crystal that we are navigating out of the app into "Android's Manage Default Apps Settings".

```suggestion
 * - For Android Q+, navigate the user to Android's Manage Default Apps Settings
 * - For anything else, allow the user to set the default browser in app. (I think this is wrong??)
```

app/src/main/java/org/mozilla/fenix/ext/Activity.kt
```diff
@@ -48,3 +56,61 @@ fun Activity.breadcrumb(
...
+                    SupportUtils.SumoTopic.SET_AS_DEFAULT_BROWSER
+                ),
+                newTab = true,
+                from = BrowserDirection.FromSettings
+            )
+        }
+    }
+}
+
+private fun Activity.navigateUserToDefaultAppsSettings() {
```
gabrielluong, line 98:
```suggestion
private fun Activity.navigateToDefaultAppsSettings() {
```

gabrielluong (approved):

app/src/main/java/org/mozilla/fenix/ext/Activity.kt
gabrielluong, line 98: Perhaps also add a kDoc for this. Can never have enough documentation to help people not have to read code.

app/src/main/java/org/mozilla/fenix/ext/Activity.kt
Amejia481: Thanks for catching this, I will go with the straightforward documentation as you suggested "**Opens Android's Manage Default Apps Settings if possible."**

app/src/main/java/org/mozilla/fenix/ext/Activity.kt
```diff
@@ -48,3 +56,59 @@ fun Activity.breadcrumb(
         )
     )
 }
+
+/**
+ *  Opens Android's Manage Default Apps Settings if possible.
```
gabrielluong, line 61:
```suggestion
 * Opens Android's Manage Default Apps Settings if possible.
```

app/src/main/java/org/mozilla/fenix/ext/Activity.kt
```diff
@@ -48,3 +56,59 @@ fun Activity.breadcrumb(
...
+                    SupportUtils.SumoTopic.SET_AS_DEFAULT_BROWSER
+                ),
+                newTab = true,
+                from = BrowserDirection.FromSettings
+            )
+        }
+    }
+}
+
+private fun Activity.navigateToDefaultBrowserAppsSettings() {
```
gabrielluong, line 96: Where's my kDoc?

Amejia481: As it's private helper function I think it wasn't need it, but we can add it too.

Issue #18837, opened 2021-04-07
Show Sync re-auth required badge on 'Settings' button


jonalmeida:
There is one more state of the 'Sign in to sync' menu that we may have overlooked (which I do not see in the designs linked). If an account needs to be re-authenticated we propagate the error to the 'Settings' menu item.

With the 'Sign in' button now in the main menu, the re-authenticated warning previously shown from 'Settings' will be shown on 'Sign in to sync'? This could also be done as a follow-up. 🙂 
   * we should continue to show this on the Settings menu item for now
   * Can followup with Victoria for future moving it to the sync sign in item

<img width="402" alt="Screen Shot 2021-04-02 at 2 56 30 PM" src="https://user-images.githubusercontent.com/1370580/113445440-e08a5500-93c3-11eb-83d9-3781077aaabf.png">

To reproduce this case for testing, you can pair with your device, then go to https://accounts.firefox.com/settings to manage your devices and remove the device from the account console. Re-open Fenix and you should be in the re-auth state.

_Originally posted by @jonalmeida in https://github.com/mozilla-mobile/fenix/issues/17798#issuecomment-812667505_

jonalmeida:
Victoria mentioned [here](https://github.com/mozilla-mobile/fenix/issues/17798#issuecomment-812712088) that this can be done in a follow-up.

violasong:
Yep, moving the badge to "Sign in to sync" can be done as a followup later :)

lobontiumira:
Verified on the 4/9 Nightly build with Samsung Galaxy Note 8 (Android 9), and on Xiaomi mi4i (Android 5.0.2):

![error](https://user-images.githubusercontent.com/40694158/114186587-b39fea00-994f-11eb-998e-a6e653ac25f8.png)

@jonalmeida, @violasong shouldn't there be an error message on the "Signed in as email@email.com" options?

eliserichards:
That is going to be a followup. There isn't UX bandwidth to do that change now.

PR #18849, opened 2021-04-07
For #18837: Settings menu item should be tinted if there is an account issue

eliserichards:
For #18837

> To reproduce this case for testing, you can pair with your device, then go to https://accounts.firefox.com/settings to manage your devices and remove the device from the account console. Re-open Fenix and you should be in the re-auth state.

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
@@ -490,7 +490,9 @@ open class DefaultToolbarMenu(
     val settingsItem = BrowserMenuHighlightableItem(
         label = context.getString(R.string.browser_menu_settings),
         startImageResource = R.drawable.ic_settings,
-        iconTintColorResource = primaryTextColor(),
+        iconTintColorResource = if (hasAccountProblem)
+            ThemeManager.resolveAttribute(R.attr.syncDisconnected, context) else
+            primaryTextColor,
         textColorResource = if (hasAccountProblem)
             ThemeManager.resolveAttribute(R.attr.primaryText, context) else
             primaryTextColor,


app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
```diff
@@ -490,7 +490,9 @@ open class DefaultToolbarMenu(
     val settingsItem = BrowserMenuHighlightableItem(
         label = context.getString(R.string.browser_menu_settings),
         startImageResource = R.drawable.ic_settings,
-        iconTintColorResource = primaryTextColor(),
+        iconTintColorResource = if (hasAccountProblem)
```
eliserichards, line 493: See line 203 (the old settings item) has this same tinting. We missed that in the new item implementation

jonalmeida:
Curious what UX thinks. I mentioned we would move the re-auth icon on the left as well.

gabrielluong (approved):
Maybe we should go with the approach you took in https://github.com/mozilla-mobile/fenix/pull/18850 where we can move the `settingsItem` variable out of both of the old/new menu so that they can be reused in both since they are identical.

kglazko (approved):
Looks good to me after testing. Let's see what UX says and then merge it if all good!

eliserichards:
This issue was to add the pre-existing auth error icons, so I think we can file a new issue for any future UX changes to this design

jonalmeida:
@eliserichards I think there was a misunderstanding - this issue was to move the highlighting to the the 'Signed in as [email]' highlight.

Seems a bit awkward to highlight the 'Settings' that Sync has an issue, right?

eliserichards:
I discussed the change with Victoria and she wants the icon to stay in settings for now (which is the pre-existing functionality). She doesn't have the bandwidth to redesign the sign in button now. That will be post-MR1 work.

Issue #18776, opened 2021-04-02
Verify PWAs still install on new menu 


eliserichards:
Make sure PWAs still install on new menu (I think this was missed in new designs)

Can be tested by:
1. Navigate to paperplanes.world/
2. Open three-dot menu
3. See that "Install" is shown (and works when clicked)

<img width="371" alt="image" src="https://user-images.githubusercontent.com/43795363/113459855-6adeb300-93dc-11eb-89c9-30bcde57b75b.png">

LaurentiuApahideanSV:
Verified as fixed on Firefox Preview Nightly 210413.
### Devices used:
- OnePlus 6T (Android 9)
- Samsung Galaxy S9 (Android 8.0.0)

PR #18831, opened 2021-04-06
No issue: Refactor three dot menu files

eliserichards:
**Default toolbar menu**
* `DefaultToolbarMenu.kt`
* Menu items: `ToolbarMenu.kt` (interface, impl by `DefaultToolbarMenu`)

**Homescreen toolbar menu**
* `HomeToolbarMenu.kt`
* Menu items: `HomeMenu.kt` (interface, impl by `HomeToolbarMenu`)
   
**FxAccount manager**
* `FenixAccountManager.kt` (_new_)
* holds a reference to `context.components.backgroundServices.accountManager`
* handles `observeAccountState`, `getAuthAccountEmail` for menu items
* Put in a new package, `.accounts`, under `org.mozilla.fenix.components` and moved relevant account classes there too

### Pull Request checklist

- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features. In addition, it includes a screenshot of a successful [accessibility scan](https://play.google.com/store/apps/details?id=com.google.android.apps.accessibility.auditor&hl=en_US) to ensure no new defects are added to the product.

### To download an APK when reviewing a PR:
1. click on Show All Checks,
2. click Details next to "Taskcluster (pull_request)" after it appears and then finishes with a green checkmark,
3. click on the "Fenix - assemble" task, then click "Run Artifacts".
4. the APK links should be on the left side of the screen, named for each CPU architecture

PR #18850, opened 2021-04-07
For #18776: PWA install to homescreen is available in main menu

eliserichards:
For #18776 

This was missed in the original designs. See issue for STR

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
@@ -199,6 +199,21 @@ open class DefaultToolbarMenu(
     } ?: false
     // End of predicates //
 
+    val installToHomescreen = BrowserMenuHighlightableItem(
+        label = context.getString(R.string.browser_menu_install_on_homescreen),
+        startImageResource = R.drawable.ic_add_to_homescreen,
+        iconTintColorResource = primaryTextColor,
+        highlight = BrowserMenuHighlight.LowPriority(
+            label = context.getString(R.string.browser_menu_install_on_homescreen),
+            notificationTint = getColor(context, R.color.whats_new_notification_color)
+        ),
+        isHighlighted = {
+            !context.settings().installPwaOpened
+        }
+    ) {
+        onItemTapped.invoke(ToolbarMenu.Item.InstallToHomeScreen)
+    }
+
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
```diff
@@ -199,6 +199,21 @@ open class DefaultToolbarMenu(
...
+            highlight = BrowserMenuHighlight.LowPriority(
+                label = context.getString(R.string.browser_menu_install_on_homescreen),
+                notificationTint = getColor(context, R.color.whats_new_notification_color)
+            ),
+            isHighlighted = {
+                !context.settings().installPwaOpened
+            }
+        ) {
+            onItemTapped.invoke(ToolbarMenu.Item.InstallToHomeScreen)
+        }
```
gabrielluong, line 215: Seems like we got an extra indent here. Might be worth running a reformat.

app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
```diff
@@ -199,6 +199,21 @@ open class DefaultToolbarMenu(
     } ?: false
     // End of predicates //
 
+    val installToHomescreen = BrowserMenuHighlightableItem(
```
gabrielluong, line 202: Just a thought, I am starting to think it would be worthwhile when we refactor all of these to add a comment for each of these items so we can easily decipher what they are without looking too hard at the code. The comment could just be the literal label or some variant for the menu item that we would see in the UI.

For instance: 
```suggestion
    // Install
    val installToHomescreen = BrowserMenuHighlightableItem(
```

gabrielluong (approved):

app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
eliserichards: I have a PR in progress right now to refactor all of this 👍

eliserichards: https://github.com/mozilla-mobile/fenix/pull/18831

Issue #18806, opened 2021-04-05
[Feature] Three-dot menu sync tabs icon should direct to Account Settings when signed in


kglazko:
* Follow-up to https://github.com/mozilla-mobile/fenix/issues/17798 

## Dependencies
- Sync Tabs item shipping (#18609) ✅ 

### Acceptance Criteria
Question: What is the appropriate landing point when signed in & not signed in?
* When a user is signed in to Sync, the menu item should point them to the accounts settings page
* When a user is signed out, it should take them to the Turn On Sync page

eliserichards:
This will correspond to sync tabs moving into the tabs tray, and doesn't have to be done by MR1.

jonalmeida:
> This will correspond to sync tabs moving into the tabs tray, and doesn't have to be done by MR1.

@eliserichards Sync Tabs is moving to the tray as part of MR1.  🙂

violasong:
Awesome! Then, just to clarify, this is what I think the states should be:

- If not signed in: "Sign in to sync," go to page titled "Turn on Sync" (or )
- If signed in: "Signed in as username" + go to account settings

eliserichards:
Thanks so much!! :D

eliserichards:
Waiting on https://github.com/mozilla-mobile/fenix/pull/19035. Can be QA'd once that lands

lobontiumira:
This is not fixed yet. Verified on Beta 89.0.0-beta.1 with Samsung Galaxy Note 8 (Android 9), and Xiaomi Mi 8 Lite (Android 9).
**Note**: cannot test on 4/21 Nightly build due to #19132.

* If **not signed in**: tapping on "Sign in to sync," from the homescreen main menu or on "Synced tabs" from three-dot page menu, redirects the user to Synced tabs, not to "Turn on Sync" area:

https://user-images.githubusercontent.com/40694158/115557683-9e657c80-a2ba-11eb-8341-e0b3ae16481a.mp4

* If **signed in**: tapping on "username@username.com" from the homescreen main menu, the user is not redirected to the account settings, instead it redirects the user to Synced tabs page:

https://user-images.githubusercontent.com/40694158/115568864-5dbf3080-a2c5-11eb-97bb-908444181b0c.mp4

I will remove the qa:needed label.

PR #18859, opened 2021-04-07
For #18806: Show sync tabs or sync account sign in based on feature flag

eliserichards:
For #18806 

**Note**: we still need to confirm the click destination for sync sign in with Victoria

- Shows synced tabs OR sync account sign in status based on the sync tabs redesign feature flag (only ever showing one; once the synced tabs move to the tabs tray then we show sync sign in)
- Sync account sign in should navigate to the account settings page when clicked

### Pull Request checklist

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
```diff
@@ -411,12 +412,23 @@ open class DefaultToolbarMenu(
...
+    } else {
+        BrowserMenuImageText(
+            syncItemTitle,
+            R.drawable.ic_synced_tabs,
+            primaryTextColor()
+        ) {
+            onItemTapped.invoke(ToolbarMenu.Item.SyncAccount)
+        }
+
     }
```
jonalmeida, line 432: If we're trying to coordinate these two features landing at different times and avoiding the time where no 'sync tabs' show up anywhere, then we should show sign in + tabs or just sign in.

In which case, do not use the `tabsTrayRewrite` flag to switch the two, use the flag to show 'Sync tabs' only.

eliserichards: Victoria's designs want synced tabs OR sync sign in. We will never show both. So as soon as the synced tabs move to the tabs tray we will start showing the sync sign in instead :)

jonalmeida: My comment wasn't about altering the final designs, but it looks like the 3-dot menu and synced tabs in the tray will not perfectly in the same weeks, so there will be a time in which nightly users will have the new three-dot menu and no new tabs tray so they won't be able to access their synced tabs. 🙂

eliserichards: Ahh I see what you mean. I have them switched.

jonalmeida: Yeah, that works too!

app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarMenuController.kt
```diff
@@ -224,6 +224,12 @@ class DefaultBrowserToolbarMenuController(
                     BrowserFragmentDirections.actionBrowserFragmentToSyncedTabsFragment()
                 )
             }
+            is ToolbarMenu.Item.SyncAccount -> browserAnimator.captureEngineViewAndDrawStatically {
```
jonalmeida, line 227: Why do we need to call this method? It seems to be used for only some menu items, but I can't decipher the reasoning for the others.

app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarMenuController.kt
```diff
@@ -359,15 +365,12 @@ class DefaultBrowserToolbarMenuController(
     @Suppress("ComplexMethod")
     private fun trackToolbarItemInteraction(item: ToolbarMenu.Item) {
         val eventItem = when (item) {
-            // TODO: These can be removed for https://github.com/mozilla-mobile/fenix/issues/17870
-            // todo === Start ===
             is ToolbarMenu.Item.OpenInFenix -> Event.BrowserMenuItemTapped.Item.OPEN_IN_FENIX
-            is ToolbarMenu.Item.InstallToHomeScreen -> Event.BrowserMenuItemTapped.Item.ADD_TO_HOMESCREEN
+            is ToolbarMenu.Item.InstallPwaToHomeScreen -> Event.BrowserMenuItemTapped.Item.ADD_TO_HOMESCREEN
```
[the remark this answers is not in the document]
gabrielluong: I think InstallPwaToHomeScreen provides more clarity. I do think it's a positive change for better code readability.

app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarMenuController.kt
eliserichards: I wasn't sure the original intention 🤔 do you think I should keep it as a normal `nav`?

app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
```diff
@@ -416,13 +417,26 @@ open class DefaultToolbarMenu(
...
-    val syncedTabs = BrowserMenuImageText(
-        syncItemTitle,
-        R.drawable.ic_synced_tabs,
-        primaryTextColor()
-    ) {
-        onItemTapped.invoke(ToolbarMenu.Item.SyncedTabs)
-    }
+    val syncTabsOrSignInItem =
+        if (tabsTrayRewrite) {
+            // if synced tabs are being shown in tabs tray, show sync sign in here.
```
gabrielluong, line 422: Sorry for this minor nit

```suggestion
            // If synced tabs are being shown in tabs tray, show sync sign in here.
```

eliserichards: 😂

app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarMenuController.kt
jonalmeida: No idea, but let's find out so that we can be sure!

Looking at the docs for the method, it says:
>     /**
>      * Makes the swipeRefresh background a screenshot of the engineView in its current state.
>      * This allows us to "animate" the engineView.
>      */

That sounds rather confusing and unhelpful; we should update this.

Let's look at [the code](https://github.com/mozilla-mobile/fenix/blob/b3f5e557cce658abb20e482e85c755abd544f1b2/app/src/main/java/org/mozilla/fenix/browser/BrowserAnimator.kt#L48-L72) instead.

From what I can read, we're taking the EngineView and retrieving a thumbnail image from it to set as the background for the swipe-to-refresh layout.

So it seems like the idea of it is to show a thumbnail of the page, during an animation, when we navigate away from the browser.

Playing around with this code, I removed the method around settings to see what it looks like without this change compared to 'Addons' which also has the method wrapper:


https://user-images.githubusercontent.com/1370580/114457412-ffe76600-9bab-11eb-8a8f-1a1b50a80695.mp4

What I noticed is that with or without this change we see no change in the animations! My guess is, by also looking at the git history, this animator was used at a time when we did have animation and that bit of code was removed, where as this bit was not.


To answer your question now that we know what the is, let's keep the method wrapper for consistency and file an issue for removing this animator because it does nothing, as far as I can see. 

^ Hope this helps in the future to break down problems. 🙂

jonalmeida (approved):
See my comment below for the longer explanation. I think this looks good to land. 🙂

app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarMenuController.kt
gabrielluong: I would be comfortable moving forward without the usage of the capture here and just do a simple navigation. We can investigate the state of the animation and its existing code to determine its fate in the near future.

jonalmeida: It's probably safer to keep it around. We're hoping to cut over a release soon-ish and it wouldn't be great to introduce a regression that we may not realize exists from just a short investigation.

This seems to have been written as a fix for the "white flashing" in dark mode before GV introduced a fix. Removing all these references together, in this case, is probably fitting so we can delete the entire `BrowserAnimator` class and work on it from scratch when we re-enable animations.

Issue #18507, opened 2021-03-16
Ensure the Credit card screens cannot be screenshot


gabrielluong:
Note that there is an existing PR with feedback and take aways https://github.com/mozilla-mobile/fenix/pull/18890. Please consider using it.

-----

Secure secure flags so that the windows of the Credit cards screen cannot be screenshotted.

clientenq:
Can you explain how you expect this suggestion to solve your issue? Credit card screens or forms are not standardised and are different per merchant.

gabrielluong:
> Can you explain how you expect this suggestion to solve your issue? Credit card screens or forms are not standardised and are different per merchant.

This wouldn't be related to the 'webview'. This would be about securing the preference screens where we plan to provide autofill and management of stored credit card information.

chrishernan:
Hello. Could I work on this issue if possible?

gabrielluong:
> Hello. Could I work on this issue if possible?

Sure, I would suggest pulling in https://github.com/mozilla-mobile/fenix/pull/18808 in order to be fully complete. You need to enable the "Credit Card" feature by going into "About Firefox" in the Settings and tapping the logo 3 times to see the Secret Settings where we add toggles for our feature flags.

chrishernan:
Will do. Thanks for the tip. I spent an hour trying to figure out where the credit card section was haha.

LaurentiuApahideanSV:
Verified as fixed on Firefox Preview Nightly 5/25.
### Devices used:
- OnePlus 6T (Android 9)
- Pixel 3 XL (Android 9)

PR #18890, opened 2021-04-09
For #18507 - Ensure credit card screens cannot be screenshot

chrishernan:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardEditorFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardEditorFragment.kt
@@ -5,7 +5,11 @@
 package org.mozilla.fenix.settings.creditcards
 
 import android.os.Bundle
-import android.view.*
+import android.view.Menu
+import android.view.MenuInflater
+import android.view.View
+import android.view.WindowManager
+import android.view.MenuItem
 import android.widget.ArrayAdapter
 import androidx.fragment.app.Fragment
 import androidx.lifecycle.lifecycleScope
--- a/app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardEditorFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardEditorFragment.kt
@@ -52,7 +52,7 @@ class CreditCardEditorFragment : Fragment(R.layout.fragment_credit_card_editor) 
     }
 
     override fun onResume() {
-        requireActivity().window.setFlags(
+        activity?.window?.setFlags(
             WindowManager.LayoutParams.FLAG_SECURE,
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardEditorFragment.kt
```diff
@@ -50,6 +47,13 @@ class CreditCardEditorFragment : Fragment(R.layout.fragment_credit_card_editor)
         )
     }
 
+    override fun onResume() {
+        requireActivity().window.setFlags(
```
jonalmeida, line 51: nit: we seem to use `addFlags`/`clearFlags` in other places, might be best to follow the same pattern there too.

gabrielluong:
@Mugurell can you take a look at this? I am not too familiar with this area. Thanks!

app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardsSettingFragment.kt
```diff
@@ -65,4 +77,19 @@ class CreditCardsSettingFragment : PreferenceFragmentCompat() {
...
+        removeFlagSecure()
+        super.onStop()
+    }
+
+    /**
+     * Checks var isNextDestinationCreditCardEditor to decide whether to remove or keep
+     * WindowManager.LayoutParams.FLAG_SECURE
+     */
+    private fun removeFlagSecure() {
+        if (!isNextDestinationCreditCardEditor) {
```
Mugurell, line 91: Seeing that `CreditCardEditorFragment` sets `FLAG_SECURE` on it's own (which is probably desired) maybe we don't need this if check and have `onStop` always remove the flag?

app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardEditorFragment.kt
```diff
@@ -50,6 +51,13 @@ class CreditCardEditorFragment : Fragment(R.layout.fragment_credit_card_editor)
         )
     }
 
+    override fun onResume() {
+        activity?.window?.setFlags(
```
Mugurell, line 55: As on `CreditCardSettingFragment` setting `FLAG_SECURE` here might be too late.
Ideally we should set this even before of `super.onStart` and then remove that flag in `onStop`.

Mugurell (changes requested):
Thank you for helping with this.
It generally looks good but given the importance of this feature maybe the code can be a little tighter. 
Also, to prevent against future regressions a few simple unit ensuring `FLAG_SECURE` exists at the expected lifecycle callbacks tests would be nice. We can help with this if needed.

gabrielluong:
Gonna open this up to allow the team to address the feedback.

Issue #15372, opened 2020-09-24
[UX] Optimize the order and messages of onboarding cards


vesta0:
## User Story
- As a NEW user, I want to quickly learn about the most interest or useful features, so I know if this browser is the right one for me and how to customize it to fit my needs. 

### Acceptance Criteria
- Onboarding cards appear in order of most interesting/useful.
- ONLY highlights unique/useful features that are also actionable
- Highlight what's unique about Firefox
- Copy is clear, precise, and impactful
- FxA sign in/sign up is not presented to me as the most important option

### UX Spec
- [Figma](https://www.figma.com/file/HvCGhaudYuBq3yPgwAWOVO/Onboarding?node-id=0%3A1)

betsymi:
Here is the new order on onboarding cards

1. Theme picker
2. Address bar location picker
3. Enhanced Tracking Protection settings
4. Sync Firefox between devices
5. Privacy notice
6. Start browsing call to action 

Note, we are removing: See what's new and Private tabs cards.

Here are the string changes. 


Current String ❌ | New String ✅ | Notes ✍️
-- | -- | --
Welcome to %s | No change | l10n note: Substitute %s for long browser name.
Choose your theme | No change |  
Save some battery and your eyesight by enabling order | Save some battery and your eyesight with dark mode. |  
Get to know %s | REMOVE STRING | No string replacement; Remove
Light theme | No change |  
Dark theme | No change |  
Automatic Adapts to your device settings | No change |  
Take a position | Pick your toolbar placement |  
Try one-handed browsing with the bottom toolbar or move it to the top. | Put the toolbar within easy reach. Keep it on the bottom, or move it to the top. |  
Top |   | Swap order. Bottom should be first, since it is the default.
Bottom |   | Swap order. Bottom should be first, since it is the default.
Automatic privacy | Always On Privacy | Remove toggle
Privacy and security settings block trackers, malware, and companies that follow you. | Firefox automatically stops companies from secretly following you around the web. |  
Standard (default) | No change |  
Blocks fewer trackers. Pages will load normally. | Balanced for privacy and performance. Pages load normally. |  
Strict | No change |  
Blocks more trackers, ads, and popups. Pages load faster, but some functionality might not work. | Blocks more trackers so pages load faster, but some on-page functionally may break. |  
  | Sync Firefox between devices |  
Start syncing bookmarks, passwords, and even more with your Firefox account. | Bring bookmarks, history, and passwords to Firefox on this device. |  
[...]

PR #18906, opened 2021-04-09
For #15372: Correct one more instance of the onboarding string

jonalmeida:
Fixed one more instance where we need to fix the string.

1. click on Show All Checks,


jonalmeida:
Removing Arturo as reviewer while he is PTO. 🙂

gabrielluong (approved):

jonalmeida:
@Delphine confirmed offline that this looks right. Let's land it!

Issue #18931, opened 2021-04-11
Bring Add-to-collection back to the overflow menu


jonalmeida:
Originally, I imagined we could do this as part of #18862, but there is a rather sizeable chunk of code there which I don't fully understand at this time and I'm hesitant to just copy it over without first unravelling it first.

This issue for re-implementing the logic for adding to collections and doing so in a re-usable way that might make it more useful outside of the tabs tray if possible.

jonalmeida:
Reading through, it looks like a lot of the code is just creating two add/create collection dialogs and showing them accordingly.

I think we can move this into the `collections` package and make the entire thing reusable without it needing to live in the tabs tray fragment; none of this code seems related at all.

jonalmeida:
Specifically, the code bits are:
 - `CollectionsAdapter`
 - `TabTrayDialogFragment.showChooseCollectionDialog`
 - `TabTrayDialogFragment.showAddNewCollectionDialog`

The `SaveToCollectionsButtonAdapter` is a lie, it's actually the "Select tabs" button that takes you into multi-select mode. 😄

PR #18932, opened 2021-04-11
Close #18862: Add multi-select banner to tabs tray 

### LINKED ISSUE #18862: Add mutli-select top banner to the tabs tray refactor
I didn't add this part to the refactor since it was already a rather large chunk of code to review.

This part will add the dark blue banner up top when we enter multi-select mode and remove it after.

jonalmeida:
⚠️ This ended up being a more sizeable PR than expected, so please take a look at each commit for easier reviewing.

Some other notes:
- Added BookmarksUseCases because we have other parts of the code where we do the same code repeatedly. For now, I've only made use of it in our tabs tray because refactoring everything else is outside of our scope.
- Added a generic form of `AbstractBinding` [from Focus](https://github.com/mozilla-mobile/focus-android/blob/master/app/src/main/java/org/mozilla/focus/browser/binding/AbstractBinding.kt), that I'd like to upstream to AC [here](https://github.com/mozilla-mobile/android-components/issues/10050).
- Add to collections is not implemented because it's quite a bit of refactoring that needs to be done there as it is. I've filed #18931 as a follow-up.
- There seem to be a lot of layout changes needed specifically for the "handle" of the tray during mode change, so I split up the binding class to handle (get it? 😄 ) just that individually.
- For reference, a lot of the original code is over [here](https://github.com/mozilla-mobile/fenix/blob/db9b3676d3c02c1814cf154a98557a871b235e05/app/src/main/java/org/mozilla/fenix/tabtray/TabTrayView.kt#L474-L618).

Overall, the code logic could do with a large simplification, but this would come at the cost of refactoring the layout files of the tray which we don't have the bandwidth to do.

https://user-images.githubusercontent.com/1370580/114293330-336aa900-9a63-11eb-8fc8-73e03bacbc15.mp4

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayFragment.kt
@@ -32,6 +32,7 @@ import org.mozilla.fenix.ext.requireComponents
 import org.mozilla.fenix.home.HomeScreenViewModel
 import org.mozilla.fenix.tabstray.browser.BrowserTrayInteractor
 import org.mozilla.fenix.tabstray.browser.DefaultBrowserTrayInteractor
+import org.mozilla.fenix.tabstray.ext.showWithTheme
 import org.mozilla.fenix.tabstray.syncedtabs.SyncedTabsInteractor
 
 @Suppress("TooManyFunctions")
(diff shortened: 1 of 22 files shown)


app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayFragment.kt
```diff
@@ -152,6 +166,40 @@ class TabsTrayFragment : AppCompatDialogFragment(), TabsTrayInteractor {
...
+                showOnSelectViews = VisibilityModifier(
+                    collect_multi_select,
+                    share_multi_select,
+                    menu_multi_select,
+                    multiselect_title,
+                    exit_multi_select
+                ),
+                showOnNormalViews = VisibilityModifier(
+                    tab_layout,
+                    tab_tray_overflow
```
jonalmeida, line 187: Ah, just noticed that we need to also hide the fab (`new_tab_button`) when we go into multi-select mode.

rocketsroger (approved):

jonalmeida:
Fixed the broken test and added the FAB hiding.

Issue #137, opened 2019-01-08
FNX-4983 ⁃ Add the app graphic to the homeview


boek:
Parent: #99

PR #18998, opened 2021-04-13
No issue: test one shard per UI test on FTL

AaronMT:
Testing the speed of test execution on FTL by using one shard per UI test.

### DIFF
--- a/taskcluster/ci/ui-test/kind.yml
+++ b/taskcluster/ci/ui-test/kind.yml
@@ -52,7 +52,7 @@ jobs:
         run-on-git-branches: ["^((?!releases[_/]).+)$"]
         run:
             commands:
-                - [automation/taskcluster/androidTest/ui-test.sh, x86, app.apk, android-test.apk, '50']
+                - [automation/taskcluster/androidTest/ui-test.sh, x86, app.apk, android-test.apk, '-1']
         treeherder:
             symbol: debug(ui-test-x86)
     screenshots-x86:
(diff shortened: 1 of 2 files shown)


bhearsum (approved):

AaronMT:
Pretty large log reduction with compact, and promising speed with one shard for test. Re-running a few times.

> [taskcluster 2021-04-15 19:06:25.806Z] Successful task run with exit code: 0 completed in 510.652 seconds

taskcluster/ci/ui-test/kind.yml
```diff
@@ -52,7 +52,7 @@ jobs:
         run-on-git-branches: ["^((?!releases[_/]).+)$"]
         run:
             commands:
-                - [automation/taskcluster/androidTest/ui-test.sh, x86, app.apk, android-test.apk, '50']
+                - [automation/taskcluster/androidTest/ui-test.sh, x86, app.apk, android-test.apk, '-1']
```
rpappalax, line 55: I think this needs a slight bit of housekeeping.  If we're using a config file (the flank file) to contain our variables, why should this one stand out as the single variable we need to pass down?  I can't remember what the logic was for leaving it this way, but we should be invoking this variable from the flank file (and not in both places).  I think you could simply remove it from here and from line #137 and then I believe it would just be picked up from the flank file automagically.

It would probably also be a good idea to make sure the help is up-to-date: https://github.com/mozilla-mobile/fenix/blob/master/automation/taskcluster/androidTest/ui-test.sh#L24-L33

Aside from that nit, LGTM

rpappalax: Once that change is in place, pls land this.  thanks!

Issue #18934, opened 2021-04-11
[Bug] Memory leak with SelectedItemAdapterBinding in tabs tray refactor


jonalmeida:
```
┬───
│ GC Root: System class
│
├─ android.view.inputmethod.InputMethodManager class
│    Leaking: NO (InputMethodManager↓ is not leaking and a class is never leaking)
│    ↓ static InputMethodManager.sInstance
├─ android.view.inputmethod.InputMethodManager instance
│    Leaking: NO (ViewRootImpl↓ is not leaking and InputMethodManager is a singleton)
│    ↓ InputMethodManager.mCurRootView
├─ android.view.ViewRootImpl instance
│    Leaking: NO (PrivateBrowserTrayList↓ is not leaking and ViewRootImpl#mView is not null)
│    ↓ ViewRootImpl.mAttachInfo
├─ android.view.View$AttachInfo instance
│    Leaking: NO (PrivateBrowserTrayList↓ is not leaking)
│    ↓ View$AttachInfo.mScrollContainers
├─ java.util.ArrayList instance
│    Leaking: NO (PrivateBrowserTrayList↓ is not leaking)
│    ↓ ArrayList.elementData
├─ java.lang.Object[] array
│    Leaking: NO (PrivateBrowserTrayList↓ is not leaking)
│    ↓ Object[].[2]
├─ org.mozilla.fenix.tabstray.browser.PrivateBrowserTrayList instance
│    Leaking: NO (View attached)
│    mContext instance of android.view.ContextThemeWrapper, wrapping activity org.mozilla.fenix.HomeActivity with mDestroyed = false
│    View.parent androidx.viewpager2.widget.ViewPager2$RecyclerViewImpl attached as well
│    View#mParent is set
│    View#mAttachInfo is not null (view attached)
│    View.mID = R.id.tray_list_item
│    View.mWindowAttachCount = 1
│    ↓ PrivateBrowserTrayList.tabsTrayStore
│                             ~~~~~~~~~~~~~
├─ org.mozilla.fenix.tabstray.TabsTrayStore instance
│    Leaking: UNKNOWN
│    ↓ TabsTrayStore.subscriptions
│                    ~~~~~~~~~~~~~
├─ java.util.Collections$SetFromMap instance
│    Leaking: UNKNOWN
│    ↓ Collections$SetFromMap.m
│                             ~
├─ java.util.concurrent.ConcurrentHashMap instance
│    Leaking: UNKNOWN
[...]

PR #19014, opened 2021-04-14
Issue #18934: Do not nullify adapter on window detached

jonalmeida:
Previously, to fix a memory leak, we were removing the adapter reference
entirely in order to have the `onDetachedFromRecyclerView` callback
invoked. This causes a side-effect where we can no longer reference the
adapter any more when we re-attach.

The simpler solution is to just invoke the needed callback directly
instead.

What the bug looks like in the tray:

https://user-images.githubusercontent.com/1370580/114648987-b41ce580-9cad-11eb-905b-7a3a5d21a33a.mp4

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/tabstray/browser/BaseBrowserTrayList.kt
+++ b/app/src/main/java/org/mozilla/fenix/tabstray/browser/BaseBrowserTrayList.kt
@@ -6,6 +6,7 @@ package org.mozilla.fenix.tabstray.browser
 
 import android.content.Context
 import android.util.AttributeSet
+import androidx.annotation.VisibleForTesting
 import androidx.recyclerview.widget.RecyclerView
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/fenix/tabstray/browser/BaseBrowserTrayList.kt
```diff
@@ -82,14 +83,15 @@ abstract class BaseBrowserTrayList @JvmOverloads constructor(
         touchHelper.attachToRecyclerView(this)
     }
 
-    override fun onDetachedFromWindow() {
+    @VisibleForTesting
+    public override fun onDetachedFromWindow() {
```
jonalmeida, line 87: Can't make this `internal` because the super has a higher visibility modifier.

codrut-topliceanu (approved):
Works & LGtM 👍

Issue #18867, opened 2021-04-08
[Bug] Three-dot menu refresh leaves lots of empty space


jonalmeida:
## For QA
See UX comments on this issue, we are removing the "Signed in as" string from the sync sign in item.

---

## Steps to reproduce
- Use a nightly build with the three-dot menu refresh (after 08/04/21).
- Sign in to Sync with a long username.
- Open the menu.

### Expected behaviour
- A more compact menu that allows for more web content to be seen.

### Actual behaviour
- There is a lot of empty space in the menu that seems unused.
- See the screenshot below that includes layout bounds for reference.

![output](https://user-images.githubusercontent.com/1370580/113970639-39634e80-9805-11eb-8be3-4688632fb497.png)

### Device information

* Device vendor / model and Android version: Android 10, Pixel 3
* Firefox for Android version: Nightly 08/04/21

jonalmeida:
I've added this to the project board, but I'm not sure what's the preferred way to surface bug reports for this feature to be triaged.

cc: @eliserichards for now.

amedyne:
tagging @violasong and @topotropic for this, too.

violasong:
That looks a lot wider than the screenshot [here](https://github.com/mozilla-mobile/android-components/pull/10039) - I wonder why?

jonalmeida:
> That looks a lot wider than the screenshot [here](https://github.com/mozilla-mobile/android-components/pull/10039) - I wonder why?

The screenshot here doesn't have a Sync account signed in. That might affect it.

violasong:
Ah, it would be ideal to use the narrower width in both situations, but I know deadlines are looming.

In the future, we can also represent the signed in state as `[avatar icon] username@email.com` to reduce the length of that list item. Flagging for @brampitoyo in case this helps with the iOS toolbar menu.

smowtenshi:
> In the future, we can also represent the signed in state as `[avatar icon] username@email.com` to reduce the length of that list item. 

Is this necessary when you can look that in the settings? In my opinion it's a really useless information.

brampitoyo:
@jonalmeida I agree with @violasong in advising to remove the words “Signed in as…”

Instead, we can just show the email address on its own, plus an avatar icon symbol.

I’m also confirming the fact that “Synchronised tabs” will eventually longer be located inside the App Menu? Instead, we’re planning to put it inside the Tab Tray.

eliserichards:
[...]

PR #19035, opened 2021-04-14
For #18867: remove "signed in as" string in three-dot menu

eliserichards:
Also refactored a little bit to clean up the file. More refactoring in progress in other PR.

* Remove "signed in as" string 
* Remove dupe primaryTextColor value
* Remove unnecessary browser nav action and used global action instead (actionBrowserFragmentToSyncedTabsFragment)
* Refactored check for syncTabs vs syncSignIn item

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
@@ -57,7 +57,7 @@ import org.mozilla.fenix.utils.BrowsersCache
  * @param lifecycleOwner View lifecycle owner used to determine when to cancel UI jobs.
  * @param bookmarksStorage Used to check if a page is bookmarked.
  */
-@Suppress("LargeClass", "LongParameterList")
+@Suppress("LargeClass", "LongParameterList", "TooManyFunctions")
 @ExperimentalCoroutinesApi
 open class DefaultToolbarMenu(
     private val context: Context,
@@ -78,7 +78,6 @@ open class DefaultToolbarMenu(
     private val selectedSession: TabSessionState?
         get() = store.state.selectedTab
 
-    private var signedInToFxA = false
     private val accountManager = context.components.backgroundServices.accountManager
 
     override val menuBuilder by lazy {
@@ -530,16 +529,19 @@ open class DefaultToolbarMenu(
         onItemTapped.invoke(ToolbarMenu.Item.SyncedTabs)
     }
 
-    val syncItemTitle =
-        if (accountManager.accountProfile()?.email != null) {
-            signedInToFxA = true
-            accountManager.accountProfile()?.email!!
+    private fun getSyncItemTitle(): String {
+        val authenticatedAccount = accountManager.authenticatedAccount() != null
+        val email = accountManager.accountProfile()?.email
+
+        return if (authenticatedAccount && email != null) {
+            email
         } else {
             context.getString(R.string.sync_menu_sign_in)
         }
+    }
 
     val syncMenuItem = BrowserMenuImageText(
-        syncItemTitle,
+        getSyncItemTitle(),
         R.drawable.ic_synced_tabs,
         primaryTextColor()
     ) {
--- a/app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
@@ -78,8 +78,8 @@ open class DefaultToolbarMenu(
     private val selectedSession: TabSessionState?
         get() = store.state.selectedTab
 
-    private val primaryTextColor =
-        ThemeManager.resolveAttribute(R.attr.primaryText, context)
+    private var signedInToFxA = false
+    private val accountManager = context.components.backgroundServices.accountManager
 
     override val menuBuilder by lazy {
         WebExtensionBrowserMenuBuilder(
@@ -91,7 +91,7 @@ open class DefaultToolbarMenu(
                 },
             endOfMenuAlwaysVisible = shouldUseBottomToolbar,
             store = store,
-            webExtIconTintColorResource = primaryTextColor,
+            webExtIconTintColorResource = primaryTextColor(),
             onAddonsManagerTapped = {
                 onItemTapped.invoke(ToolbarMenu.Item.AddonsManager)
             },
@@ -103,7 +103,7 @@ open class DefaultToolbarMenu(
         val back = BrowserMenuItemToolbar.TwoStateButton(
             primaryImageResource = mozilla.components.ui.icons.R.drawable.mozac_ic_back,
             primaryContentDescription = context.getString(R.string.browser_menu_back),
-            primaryImageTintResource = primaryTextColor,
+            primaryImageTintResource = primaryTextColor(),
             isInPrimaryState = {
                 selectedSession?.content?.canGoBack ?: true
             },
@@ -117,7 +117,7 @@ open class DefaultToolbarMenu(
         val forward = BrowserMenuItemToolbar.TwoStateButton(
             primaryImageResource = mozilla.components.ui.icons.R.drawable.mozac_ic_forward,
             primaryContentDescription = context.getString(R.string.browser_menu_forward),
-            primaryImageTintResource = primaryTextColor,
+            primaryImageTintResource = primaryTextColor(),
             isInPrimaryState = {
                 selectedSession?.content?.canGoForward ?: true
             },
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
```diff
@@ -546,11 +517,35 @@ open class DefaultToolbarMenu(
...
+        context.getString(R.string.synced_tabs),
+        R.drawable.ic_synced_tabs,
+        primaryTextColor()
+    ) {
+        onItemTapped.invoke(ToolbarMenu.Item.SyncedTabs)
+    }
+
+    val syncItemTitle =
+        if (accountManager.accountProfile()?.email != null) {
+            signedInToFxA = true
```
gabrielluong, line 535: I don't see this checked or used anywhere.

app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
```diff
@@ -546,11 +517,35 @@ open class DefaultToolbarMenu(
...
+            accountManager.accountProfile()?.email!!
```
gabrielluong, line 536: Do we need the !! when we check that it's not null in the if statement?

app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
```diff
@@ -78,8 +78,8 @@ open class DefaultToolbarMenu(
     private val selectedSession: TabSessionState?
         get() = store.state.selectedTab
 
-    private val primaryTextColor =
-        ThemeManager.resolveAttribute(R.attr.primaryText, context)
```
gabrielluong, line 82: Sorry.. I should've caught this in my review of https://github.com/mozilla-mobile/fenix/pull/18427.

app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
```diff
@@ -546,11 +517,35 @@ open class DefaultToolbarMenu(
...
+    val syncedTabsItem = BrowserMenuImageText(
+        context.getString(R.string.synced_tabs),
+        R.drawable.ic_synced_tabs,
+        primaryTextColor()
+    ) {
+        onItemTapped.invoke(ToolbarMenu.Item.SyncedTabs)
+    }
+
+    val syncItemTitle =
+        if (accountManager.accountProfile()?.email != null) {
```
gabrielluong, line 534: Are we certain that this guarantees the user is signed into FxA without checking if it is authenticated? I feel like we dig a little into SyncPreferenceView to check.

gabrielluong (changes requested):

app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
eliserichards: Shoot that's a leftover from another PR. Good catch

app/src/main/res/values/strings.xml
```diff
@@ -1725,4 +1723,6 @@
     <string name="onboarding_firefox_account_auto_signin_header_2">You are signed in as %s on another Firefox browser on this phone. Would you like to sign in with this account?</string>
     <!-- Deprecated: Describes the add to homescreen functionality -->
     <string name="add_to_homescreen_description">You can easily add this website to your phone’s Home screen to have instant access and browse faster with an app-like experience.</string>
+    <!-- Deprecated: An string shown when an account is signed in where %1$s is a placeholder for the email-->
+    <string name="sync_signed_as">Signed in as %1$s</string>
```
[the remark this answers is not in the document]
eliserichards: This is our "quarantine" for unused strings. They stay here for a while and then get removed. This _used to_ be the suggestion from Delphine, but maybe she can confirm 😅

app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
eliserichards: I'm gonna take a look at other places in the code and redo this part

app/src/main/res/values/strings.xml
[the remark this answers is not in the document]
Delphine: Oh, we're totally fine with directly removing the strings (unless you think you may need them in the future)! thanks for asking

app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
```diff
@@ -546,11 +516,37 @@ open class DefaultToolbarMenu(
...
+    val syncedTabsItem = BrowserMenuImageText(
+        context.getString(R.string.synced_tabs),
+        R.drawable.ic_synced_tabs,
+        primaryTextColor()
+    ) {
+        onItemTapped.invoke(ToolbarMenu.Item.SyncedTabs)
+    }
+
+    private val syncSignInText = context.getString(R.string.sync_menu_sign_in)
+    private fun getSyncItemTitle(): String {
```
gabrielluong, line 533: I suspect you're gonna use this in https://github.com/mozilla-mobile/fenix/pull/19037. Can we move this to any general Util file that can be used between the 2? Also, a comment for the function would be nice.

app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
```diff
@@ -546,11 +516,37 @@ open class DefaultToolbarMenu(
...
+        val accountExists = accountManager.authenticatedAccount() != null
+        return if (accountExists) {
```
gabrielluong, line 535: We might prefer the removed solution on lines -414 to -419 in order to avoid the ?:. What do you think?

return if (accountManager.authenticatedAccount() != null && accountManager.accountProfile()?.email != null) {
```suggestion
        return if (accountManager.authenticatedAccount() != null && accountManager.accountProfile()?.email != null) {
```

app/src/main/res/values/strings.xml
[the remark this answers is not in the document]
gabrielluong: Let's also remove these deprecated strings either in a followup or here. We will want to check each deprecated strings to make sure they don't reference anything.

gabrielluong (changes requested):

app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
eliserichards: There's a refactoring PR that I have in progress right now, I'm going to do that there

app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
eliserichards: That doesn't work, I need some sort of cast or fallback even if we check for null
image

image

eliserichards: do you have a preference? `as String` would probably be my choice

gabrielluong: Ohh that's wild. I was just looking at this block from before.

```
    val accountManager = context.components.backgroundServices.accountManager
    val account = accountManager.authenticatedAccount()
    val syncItemTitle = if (account != null && accountManager.accountProfile()?.email != null) {
        context.getString(R.string.sync_signed_as, accountManager.accountProfile()?.email)
    } else {
    }
```
I think we should avoid manual casting. There's nothing particularly wrong with your previous solution. The only notable point I would make is that when I see a ?: after an if statement, it usually just means we should overload the if statement to avoid the ?:.

What if we set 
```
val email = accountManager.accountProfile()?.email
```
before the if statement?

eliserichards: It will still need some sort of non-null assertion. Either 

```kotlin
val email = accountManager.accountProfile()?.email!!
``` 

or

```kotlin
val syncItemTitle = if (account != null && accountManager.accountProfile()?.email != null) {   
       email!!
}
```

eliserichards: the old block was working because we had "Signed in as %1" as the string, and this PR is removing that

app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
gabrielluong: Ok, let's make sure we make a note of this to not forget.

eliserichards:
Filed #19059 to remove other deprecated strings :)

app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
eliserichards: It's in the other PR

gabrielluong (approved):

Issue #18819, opened 2021-04-06
Hide the Credit Card preference by default


gabrielluong:
Just noticed that "Credit cards" flashing briefly when navigating to Settings in Nightly. Looks like it's visible by default but being deactivated async. We should set the preference visibility to false by default.

Issue #19051, opened 2021-04-15
Improve the Settings screen


Mugurell:
Currently settings are updated after first being shown on screen.
They may start visible or not and after being displayed on the screen we recheck their status and may decide to add or remove them from the screen. This is a visual bug caused by calling the [update](https://github.com/mozilla-mobile/fenix/blob/a3f4cee388f8d18208ecc11b4ee84900f7e4e659/app/src/main/java/org/mozilla/fenix/settings/SettingsFragment.kt#L177) method in onResume - with the screen already being in the visible state.
This was reported recently also in https://github.com/mozilla-mobile/fenix/issues/18620 and https://github.com/mozilla-mobile/fenix/issues/18819.

There are also recent changes which try to mitigate this issue by loading an entire different PreferenceScreen with just one item changed just to avoid having to add/remove it later, showing again that visual glitch.
I don't think is a maintainable approach because if later another settings needs to be displayed or not we will have an explosion of xmls for all possible permutations and complex nested if checks to decide which settings screen to show.

The solution I'm proposing entails:
- having only one, stable, PreferenceScreen (maybe two at max to support preferences with icons or not)
- all preferences having the `isPreferenceVisible` set to true (default)
- inside `onCreatePreferences` iterate recursively through all `PreferenceGroup`s and `Preference`s. 
This method is called before the settings screen being visible with it's main purpose being to help in creating the right settings options structure.
- call `removePreference` on any `Preference` or even `PreferenceCategory` we don't want.
- [optional] do any other customization for the current iterated to `Preference`.

After the `onCreatePreferences` callback and showing the intended list of settings there is no reason for the shown `Preference`s to be added/removed.
With these changes we'll have a clean, maintainable solution without this sort of issues:

| Removing a setting in `onResume` | Removing the same in `onCreatePreferences` |
| -- | -- |
| <video src="https://user-images.githubusercontent.com/11428869/114865424-e1c37500-9dfa-11eb-99c6-2a7efe6316b7.mp4" /> | <video src="https://user-images.githubusercontent.com/11428869/114865402-dbcd9400-9dfa-11eb-85c5-0f1e63d6ff8c.mp4" />

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-10403)

csadilek:
Moved to bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1809242

[...]

PR #19058, opened 2021-04-15
Issue #18819: Hide Credit Card pref by default (for settings w/o icons)

csadilek:
Just saw the pref is still flashing briefly. We switched to a different preference file in the meantime / since the first fix.


gabrielluong (approved):

Mugurell:
I think this will give the opposite bug, the settings screen will appear without this CC one which will be "added" later and so it and other settings will appear to be moving on the screen.
I proposed here - https://github.com/mozilla-mobile/fenix/issues/19051 a solution to have a static structure of settings.
Otherwise this kind of bugs will keep popping up.

PR #19082, opened 2021-04-16
Closes #18946: Add undo toast for tabstray

### LINKED ISSUE #18946: Implement Undo UI in refactored tabs tray
We have the use cases for this, we just need the tab UI that can observe/react as we did in the old tray.

Consider that:
 - We need to use a similar toast for add to colections/bookmarks.
 - Undo from the last tab needs to close the tray (the current logic, passes it to the home screen to handle.
 - When adding the FAB as an anchor if that would still work when in accessibility mode (which doesn't have that button, see #18816).

rocketsroger:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayFragment.kt
@@ -28,8 +28,11 @@ import kotlinx.android.synthetic.main.tabstray_multiselect_items.*
 import kotlinx.android.synthetic.main.tabstray_multiselect_items.view.*
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.plus
+import mozilla.components.browser.state.selector.findTab
+import mozilla.components.browser.state.selector.getNormalOrPrivateTabs
 import mozilla.components.browser.state.selector.normalTabs
 import mozilla.components.browser.state.selector.privateTabs
+import mozilla.components.browser.state.state.TabSessionState
 import mozilla.components.concept.tabstray.Tab
 import mozilla.components.support.base.feature.ViewBoundFeatureWrapper
 import org.mozilla.fenix.HomeActivity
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/mozilla/fenix/tabstray/ext/TabSessionState.kt
```diff
@@ -13,3 +13,9 @@ fun TabSessionState.filterFromConfig(configuration: Configuration): Boolean {
 
     return content.private == isPrivate
 }
+
+fun TabSessionState.getTrayPosition(): Int =
+    when (content.private) {
+        true -> 1
+        false -> 0
+    }
```
jonalmeida, line 21: nit: If you use `Page` enum values, that might be just one source of truth for these values. Adds a bit more symmetry with the `positionToPage` too. 😄 

```suggestion
    when (content.private) {
        true -> Page.PrivateTabs.ordinal
        false -> Page.NormalTabs.ordinal
    }
```

app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayFragment.kt
```diff
@@ -249,6 +253,8 @@ class TabsTrayFragment : AppCompatDialogFragment(), TabsTrayInteractor {
 
     override fun setCurrentTrayPosition(position: Int, smoothScroll: Boolean) {
         tabsTray.setCurrentItem(position, smoothScroll)
+        tab_layout.getTabAt(position)?.select()
+        tabsTrayStore.dispatch(TabsTrayAction.PageSelected(Page.positionToPage(position)))
```
jonalmeida, line 257: Can we move this as a method of the `TabLayoutMediator`? It would be nicer to have the logic that sync's the two layout types in one place.

jonalmeida (approved):

One small change in API design (depending on how to feel about it).

app/src/main/java/org/mozilla/fenix/tabstray/ext/TabSessionState.kt
rocketsroger: Good idea, I'll update

app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayFragment.kt
rocketsroger: Agreed.  Thanks

Issue #19116, opened 2021-04-19
Refactor main three-dot menu


eliserichards:
Followup to #17796

Clean up the code for the main three-dot menu and make it testable.

jonalmeida:
Consider that we also want to switch to browser-menu2 as well and how that work will fit in. Maybe the refactor can go along with that and code cov can be increased? We did this with tabs tray: re-write + unit test everything.

Issue #19115, opened 2021-04-19
Refactor Homescreen three-dot menu


eliserichards:
Followup to #17796 

Clean up the code for the homescreen three-dot menu and make it testable.

Issue #19114, opened 2021-04-19
[Bug] Three-dot menu shows account settings when not signed in


jonalmeida:
[comment]: # (Please do your best to search for duplicate issues before filing a new issue so we can keep our issue board clean)
[comment]: # (Every issue should have the exact bug and steps to reproduce described in it. Please do not file feedback list tickets as it is difficult to parse them and address their individual points)
[comment]: # (Read https://github.com/mozilla-mobile/fenix#i-want-to-file-an-issue for more information)

## Steps to reproduce
1. Ensure you **don't** have a Sync account.
2. Click on the 'Sign in to sync' button on a browser screen.

### Expected behavior
- You are taken to the screen to sign in to a Sync account.

### Actual behavior
- You are taken to a (disabled) account settings screen that does not function.

### Device information

* Device vendor / model and Android version: Android 10, Pixel 3
* Firefox for Android version: ? (go to Settings -> About Firefox) master as of 8bb7abc3f 
 
<img width="434" alt="Screen Shot 2021-04-19 at 1 49 26 PM" src="https://user-images.githubusercontent.com/1370580/115280552-1181be00-a116-11eb-87be-82e876bb95eb.png">

sflorean:
Verified as fixed on Nightly 4/23 with Pixel 3 (Android 11). Scenarios covered:
- Fresh install
- Disconnect account and perform the steps from the description
- Sign in from 3dot menu, tabs tray -synced tabs, and 3dotmenu from synced tabs
- Normal and private browsing (where applicable).

PR #19037, opened 2021-04-14
For #19005: new tab three-dot menu sync sign in 

eliserichards:
For #19005 

The sync item in the new tab three-dot menu should reflect the tabs tray rewrite. 
* If synced tabs are in the tabs tray, we show our new "sync sign in" item.
* Sync sign in clicks navigate to account settings

  image

### Pull Request checklist

- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features. In addition, it includes a screenshot of a successful [accessibility scan](https://play.google.com/store/apps/details?id=com.google.android.apps.accessibility.auditor&hl=en_US) to ensure no new defects are added to the product.

### To download an APK when reviewing a PR:
1. click on Show All Checks,
2. click Details next to "Taskcluster (pull_request)" after it appears and then finishes with a green checkmark,
3. click on the "Fenix - assemble" task, then click "Run Artifacts".
4. the APK links should be on the left side of the screen, named for each CPU architecture

PR #19118, opened 2021-04-19
For #19114: check state of sync account when navigating from sync sign in menu item

eliserichards:
For #19114

Similar problem in tabs tray: https://github.com/mozilla-mobile/fenix/pull/19119

Right now we need to refactor the menus to listen for FxA changes. This will be a bigger refactor in #19116 and #19115. This PR _works_ with the caveat that once you sign in, you won't see your email address in the menu right away. You have to click on the item and then the menu regenerates and will show the account the next time you open it.

**Tests**
- [x] HomeMenu - covered by UI tests. `HomeFragment` needs refactoring (interactor and controller for handling menu item taps) before it can be testable
- [x] FenixAccountManagerTest
- [x] DefaultBrowserToolbarMenuControllerTest
- [x] SmokeTest

https://user-images.githubusercontent.com/43795363/115291231-1ba5bb80-a11a-11eb-8098-790af53b311c.mp4

https://user-images.githubusercontent.com/43795363/115291318-37a95d00-a11a-11eb-81c7-31c8e266858f.mp4

### Pull Request checklist

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
```diff
@@ -152,7 +160,7 @@ open class DefaultToolbarMenu(
             contentDescription = context.getString(R.string.browser_menu_share),
             iconTintColorResource = primaryTextColor(),
             listener = {
-                onItemTapped.invoke(ToolbarMenu.Item.Share)
+                onItemTapped.invoke(Item.Share)
```
jonalmeida, line 163: nit: Let's avoid changes like these in bugfix PRs.

It makes it easier on the reviewer, and also when some in the future needs to debug and look at git history or perform reverts. It's harder to manipulate changes then.

In the future, can just leave it as `ToolbarMenu.Item.Share` and all the other places.

app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
```diff
@@ -545,11 +555,16 @@ open class DefaultToolbarMenu(
...
     ) {
-        onItemTapped.invoke(ToolbarMenu.Item.SyncAccount)
+        onItemTapped.invoke(Item.SyncAccount(signedInToFxa))
     }
 
     @VisibleForTesting(otherwise = PRIVATE)
     val newCoreMenuItems by lazy {
+
+        // Predicates that are called once, during screen init
+        val shouldShowSaveToCollection = (context.asActivity() as? HomeActivity)
```
jonalmeida, line 565: I'm not sure what this is related to and referencing with the bug for #19114 . Let's remove it if it's not needed or file a separate issue for it and reference it to a new PR?

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -777,53 +779,65 @@ class HomeFragment : Fragment() {
             context,
             onItemTapped = {
                 when (it) {
-                    HomeMenu.Item.Settings -> {
+                    is Item.Settings -> {
```
jonalmeida, line 782: We only need to add the `is` for ~`SyncAccount`~ (see other comment about remove this) and `DesktopMode`, looking at this patch, since those are the only non-`object` classes that need to be type checked, the others are singleton objects; only one will ever exist.

app/src/main/java/org/mozilla/fenix/components/toolbar/ToolbarMenu.kt
```diff
@@ -23,7 +23,7 @@ interface ToolbarMenu {
         object InstallPwaToHomeScreen : Item()
         object AddToHomeScreen : Item()
         object SyncedTabs : Item()
-        object SyncAccount : Item()
+        data class SyncAccount(val signedIn: Boolean) : Item()
```
jonalmeida, line 26: I think we can remove this `signedIn` attribute for now since we aren't using it in this patch.

app/src/main/java/org/mozilla/fenix/home/HomeMenu.kt
```diff
@@ -92,17 +93,35 @@ class HomeMenu(
...
 
+    val syncedTabsItem = BrowserMenuImageText(
+        context.getString(R.string.synced_tabs),
+        R.drawable.ic_synced_tabs,
+        primaryTextColor
+    ) {
+        onItemTapped.invoke(Item.SyncedTabs)
+    }
+
+    private var signedInToFxa = false
```
jonalmeida, line 104: nit: can be removed as well since it's unused.

app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
```diff
@@ -583,17 +598,9 @@ open class DefaultToolbarMenu(
 
     private fun handleBookmarkItemTapped() {
         if (!isCurrentUrlBookmarked) isCurrentUrlBookmarked = true
-        onItemTapped.invoke(ToolbarMenu.Item.Bookmark)
+        onItemTapped.invoke(Item.Bookmark)
     }
 
-    @ColorRes
```
jonalmeida, line 589: Why did we move this code from here to above in this bug fix? How does it relate to this sign-in bug?

app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
```diff
@@ -74,11 +75,18 @@ open class DefaultToolbarMenu(
 
     private val shouldDeleteDataOnQuit = context.settings().shouldDeleteBrowsingDataOnQuit
     private val shouldUseBottomToolbar = context.settings().shouldUseBottomToolbar
+    val accountManager = context.components.backgroundServices.accountManager
```
jonalmeida, line 78: nit: make `private`

jonalmeida (commented):
Works well! Let's just clean this up a bit before we land it.

app/src/main/java/org/mozilla/fenix/home/HomeMenu.kt
eliserichards: It's not unused. See lines 110 and 122

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
eliserichards: I prefer this for consistency

app/src/main/java/org/mozilla/fenix/home/HomeMenu.kt
jonalmeida: Ah, my mistake! 🙂

app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
eliserichards: They are moved to a different place. Code cleanup.

jonalmeida: Let's avoid code cleanup and refactors in bug fixes. See https://github.com/mozilla-mobile/fenix/pull/19118#discussion_r616128770.

app/src/main/java/org/mozilla/fenix/components/toolbar/ToolbarMenu.kt
eliserichards: We are using it to navigate. See `HomeFragment` and `BrowserToolbarMenuController.kt`

eliserichards:
https://github.com/mozilla-mobile/fenix/pull/19037 should land first for UI test fixes.

app/src/main/java/org/mozilla/fenix/home/HomeMenu.kt
```diff
@@ -92,17 +93,35 @@ class HomeMenu(
...
     private fun getSyncItemTitle(): String {
         val authenticatedAccount = accountManager.authenticatedAccount() != null
         val email = accountManager.accountProfile()?.email
 
         return if (authenticatedAccount && email != null) {
+            signedInToFxa = true
```
Amejia481, line 110: Function with side effects is not a good idea,  for `signedInToFxa` to be correct `getSyncItemTitle()` should be executed in an specific order before `signedInToFxa` gets used, that could be fragile as `getSyncItemTitle()` main purpose is not getting the Sync status. I think we could have a function that queries the sync status and can be called from both places L110 and L122.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -777,53 +779,65 @@ class HomeFragment : Fragment() {
...
                         )
                     }
-                    HomeMenu.Item.Bookmarks -> {
+                    is Item.SyncAccount -> {
+                        hideOnboardingIfNeeded()
+                        val directions = if (it.signedIn) {
+                            BrowserFragmentDirections.actionGlobalAccountSettingsFragment()
+                        } else {
+                            BrowserFragmentDirections.actionGlobalTurnOnSync()
+                        }
```
Amejia481, line 803: This looks similar to what we have on` BrowserToolbarMenuController#229`, we could have a helper function.
image

eliserichards: I had helper functions but @jonalmeida didn't want code cleanup in this PR so I removed them

app/src/main/java/org/mozilla/fenix/home/HomeMenu.kt
```diff
@@ -92,17 +93,35 @@ class HomeMenu(
...
             context.getString(R.string.sync_menu_sign_in)
         }
     }
 
+    val syncMenuItem = BrowserMenuImageText(
+        getSyncItemTitle(),
+        R.drawable.ic_synced_tabs,
+        primaryTextColor
+    ) {
+        onItemTapped.invoke(Item.SyncAccount(signedInToFxa))
```
Amejia481, line 122: Having a function that queries the sync status as mentioned on https://github.com/mozilla-mobile/fenix/pull/19118/files#r616145969 will be more accurate to the latest sync status. I'm not sure if this state could change in the background, in case users invalidate their credentials from another device.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
Amejia481: 👍🏽

jonalmeida: Sorry, I'm not sure where the helper function was, my intent was to remove the `FenixAccountManager` and any refactoring code that was not related to the bug fix.

eliserichards:
@jonalmeida I'm going to override you here and add my `FenixAccountManager` back in so I can address Arturo's comments and not have to fix it in a bunch of places.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
Amejia481: Sorry, I just noticed code was duplicated. Agree if the code it's not related to this patch is better to not bring it in now.

app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
Amejia481: Agree let's address this!

app/src/main/java/org/mozilla/fenix/home/HomeMenu.kt
Amejia481: If we check `accountManager.authenticatedAccount() != null` here we won't need `signedInToFxa` and it will be safer as we are checking the latest status :)

app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarMenuController.kt
```diff
@@ -226,10 +226,15 @@ class DefaultBrowserToolbarMenuController(
                 )
             }
             is ToolbarMenu.Item.SyncAccount -> {
+                val directions = if (item.signedIn) {
+                    BrowserFragmentDirections.actionGlobalAccountSettingsFragment()
+                } else {
+                    BrowserFragmentDirections.actionGlobalTurnOnSync()
+                }
```
Amejia481, line 233: Can we use the account manager directly here?   That could safe a lot of work `context.components.backgroundServices.accountManager` and do the check `accountManager.authenticatedAccount() != null` ?

eliserichards: Since the accountManager is tricky, I don't want to add it to the controller as well. The menu is already holding a reference, so we can check this value and pass it here. I don't think it adds any more complexity to pass the `Boolean` `signedIn` to the controller

app/src/main/java/org/mozilla/fenix/home/HomeMenu.kt
```diff
@@ -103,6 +106,14 @@ class HomeMenu(
         }
     }
 
+    val syncMenuItem = BrowserMenuImageText(
+        getSyncItemTitle(),
+        R.drawable.ic_synced_tabs,
+        primaryTextColor
+    ) {
+        onItemTapped.invoke(Item.SyncAccount(accountManager.signedInToFxa()))
```
eliserichards, line 114: Checks auth status

app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
```diff
@@ -545,7 +554,7 @@ open class DefaultToolbarMenu(
         R.drawable.ic_synced_tabs,
         primaryTextColor()
     ) {
-        onItemTapped.invoke(ToolbarMenu.Item.SyncAccount)
+        onItemTapped.invoke(Item.SyncAccount(accountManager.signedInToFxa()))
```
eliserichards, line 557: Check auth status

app/src/androidTest/java/org/mozilla/fenix/ui/robots/SyncSignInRobot.kt
```diff
@@ -44,3 +45,10 @@ private fun assertAccountSettingsMenuHeader() {
     onView(withText(R.string.preferences_account_settings))
         .check((matches(withEffectiveVisibility(Visibility.VISIBLE))))
 }
+
+private fun assertSyncSignInMenuHeader() {
+    // Replaced with the new string here, the test is assuming we are NOT signed in
+    // Sync tests in SettingsSyncTest are still TO-DO, so I'm not sure that we have a test for signing into Sync
```
eliserichards, line 51: Comment is duped from line 43. It's relevant here, too.

app/src/androidTest/java/org/mozilla/fenix/ui/robots/SyncSignInRobot.kt
```diff
@@ -44,3 +45,9 @@ private fun assertAccountSettingsMenuHeader() {
     onView(withText(R.string.preferences_account_settings))
         .check((matches(withEffectiveVisibility(Visibility.VISIBLE))))
 }
+
+private fun assertSyncSignInMenuHeader() {
+    // Sync tests in SettingsSyncTest are still TO-DO, so I'm not sure that we have a test for signing into Sync
```
eliserichards, line 50: Comment is duped from line 43. It's relevant here, too.

app/src/main/java/org/mozilla/fenix/components/accounts/FenixAccountManager.kt
```diff
@@ -0,0 +1,37 @@
...
+import mozilla.components.service.fxa.manager.FxaAccountManager
+import org.mozilla.fenix.ext.components
+
+/**
+ * Component which holds a reference to [FxaAccountManager]. Manages account authentication,
+ * profiles, and profile state observers.
+ */
+open class FenixAccountManager(context: Context) {
+    val accountManager = context.components.backgroundServices.accountManager
+    val authenticatedAccount = accountManager.authenticatedAccount() != null
```
Amejia481, line 17: This should be a function, as it's now it is only going to store the first value when the object of `FenixAccountManager` is created, instead of constantly querying the latest value from `accountManager.authenticatedAccount()`.

app/src/main/java/org/mozilla/fenix/components/accounts/FenixAccountManager.kt
```diff
@@ -0,0 +1,37 @@
...
+
+    /**
+     * Get the email address associated with the authenticated account profile. Returns null if
+     * the account is not authenticated or the email address is null.
+     */
+    fun getAuthAccountEmail(): String? {
+        val email = accountManager.accountProfile()?.email
+        return if (authenticatedAccount && !email.isNullOrEmpty()) email else null
```
Amejia481, line 25: We could simplify this function by just returning directly `accountManager.accountProfile()?.email` as we are not going to have a profile if we are not authenticated and either way we are returning `String?`, this means that the consumer have to do the null-able or empty validation.

app/src/main/java/org/mozilla/fenix/home/HomeMenu.kt
```diff
@@ -93,6 +94,33 @@ class HomeMenu(
...
+
+    private fun getSyncItemTitle(): String {
+        val authenticatedAccount = accountManager.authenticatedAccount
+        val email = accountManager.getAuthAccountEmail()
+
+        return if (authenticatedAccount && email != null) {
+            email
+        } else {
+            context.getString(R.string.sync_menu_sign_in)
+        }
```
Amejia481, line 113: This is an example of the comment above https://github.com/mozilla-mobile/fenix/pull/19118/files#r617093999

Amejia481: We could add `isNullOrEmpty` as part of the email validation.

Amejia481 (changes requested):
It's looking much better now  ✨ ,  I added some comments almost ready to land 🎉 /

app/src/main/java/org/mozilla/fenix/components/accounts/FenixAccountManager.kt
Amejia481: You could use easily solve this by using  `get() =`

app/src/main/java/org/mozilla/fenix/components/accounts/FenixAccountManager.kt
eliserichards: We _could_ have a profile when the account needs reauthentication. I guess in that case it might not matter if we show the email?

Amejia481 (approved):
Awesome work 🥇 !

PR #19127, opened 2021-04-20
Closes #10300: Fix Baidu search telemetry

### LINKED ISSUE #10300: In_content search telemetry for Baidu not tracked
## Steps to reproduce
Go to Baidu and search for anything

### Expected behavior
Sap or organic telemetry is tracked

### Actual behavior
No metric is tracked

### Device information
* Android device: Any
* Fenix version: Latest develop

### Engineering info
The problem is that the Desktop codes for tracking are different than those on mobile. We need to research and understand the differences (example: `word` instead of `wd` for `queryParam`)

rxumoz:
Since there might be two types of urls from Baidu (https://m.baidu.com/s?from=... or https://m.baidu.com/from=*/s?), this patch handles baidu separately in Utils.kt

However, there is still a problem that the codes from Baidu are started with numbers (e.g. 844b/1000969a/...), which would be regarded as invalid characters by Glean.

Need more discussion about this, so leave it a draft.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/search/telemetry/Utils.kt
+++ b/app/src/main/java/org/mozilla/fenix/search/telemetry/Utils.kt
@@ -22,6 +22,20 @@ internal fun getTrackKey(
 
     if (provider.codeParam.isNotEmpty()) {
         code = uri.getQueryParameter(provider.codeParam)
+        if (code.isNullOrEmpty() &&
+            provider.name == "baidu" &&
+            uri.toString().contains("from=")) {
+            code = uri.toString().substringAfter("from=", "")
+                    .substringBefore("/", "")
+        }
+
+        // Glean doesn't allow code starting with a figure
+        if (code != null && code.isNotEmpty()) {
+            val codeStart = code.first()
+            if (codeStart.isDigit()) {
+                code = "_$code"
+            }
+        }
 
         // Try cookies first because Bing has followOnCookies and valid code, but no
         // followOnParams => would tracks organic instead of sap-follow-on
@@ -33,7 +47,7 @@ internal fun getTrackKey(
         }
 
         // For Bing if it didn't have a valid cookie and for all the other search engines
-        if (hasValidCode(code, provider)) {
+        if (hasValidCode(uri.getQueryParameter(provider.codeParam), provider)) {
             val channel = uri.getQueryParameter(CHANNEL_KEY)
             val type = getSapType(provider.followOnParams, paramSet)
             return TrackKeyInfo(provider.name, type, code, channel).createTrackKey()
--- a/app/src/main/assets/extensions/ads/manifest.template.json
+++ b/app/src/main/assets/extensions/ads/manifest.template.json
@@ -13,6 +13,8 @@
       "include_globs": [
         "https://www.google.*/search*",
         "https://www.bing.com/search*",
+        "https://www.baidu.com/*",
+        "https://m.baidu.com/*",
         "https://duckduckgo.com/*"
       ],
       "js": ["ads.js"],
--- a/app/src/main/assets/extensions/cookies/manifest.template.json
+++ b/app/src/main/assets/extensions/cookies/manifest.template.json
@@ -12,8 +12,8 @@
       "matches": ["https://*/*"],
       "include_globs": [
         "https://www.google.*/search*",
-        "https://www.baidu.com/from=844b/s*",
-        "https://www.baidu.com/from=844b/baidu*",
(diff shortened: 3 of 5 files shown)


rxumoz:
Since the problem of Baidu that the codes start with figures, I add prefix to these codes with "_"

app/src/main/java/org/mozilla/fenix/search/telemetry/Utils.kt
```diff
@@ -18,10 +18,30 @@ internal fun getTrackKey(
...
     if (provider.codeParam.isNotEmpty()) {
         code = uri.getQueryParameter(provider.codeParam)
+        if (code.isNullOrEmpty() &&
+            provider.name == "baidu" &&
+            uri.toString().contains("from=")) {
+            code = uri.toString().substringAfter("from=", "")
+                    .substringBefore("/", "")
+        }
+
+        // Telemetry doest not allow code starting with a figure
```
gabrielluong, line 33:
```suggestion
        // Glean doesn't not allow code starting with a figure
```

app/src/main/java/org/mozilla/fenix/search/telemetry/Utils.kt
```diff
@@ -18,10 +18,30 @@ internal fun getTrackKey(
...
+        codeWithPrefix = if (code != null && code.isNotEmpty()) {
+            val codeStart = code.first()
+            val figures = listOf('0', '1', '2', '3', '4', '5', '6', '7', '8', '9')
+            if (codeStart in figures) {
```
gabrielluong, line 37: I think you can do `codeStart.isDigit()` and remove `figures` https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/is-digit.html

app/src/main/java/org/mozilla/fenix/search/telemetry/Utils.kt
```diff
@@ -18,10 +18,30 @@ internal fun getTrackKey(
...
         code = uri.getQueryParameter(provider.codeParam)
+        if (code.isNullOrEmpty() &&
+            provider.name == "baidu" &&
+            uri.toString().contains("from=")) {
+            code = uri.toString().substringAfter("from=", "")
+                    .substringBefore("/", "")
+        }
+
+        // Telemetry doest not allow code starting with a figure
+        codeWithPrefix = if (code != null && code.isNotEmpty()) {
```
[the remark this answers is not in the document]
rxumoz: If we modify the `code` variable and don't add a variable to record the original `code` value, then we need to modify `hasValidCode()` method to trigger `SAP` search metrics correctly.
Maybe I should modify the `code` directly for concisenes and use `originalCode` variable to record?

app/src/main/java/org/mozilla/fenix/search/telemetry/Utils.kt
```diff
@@ -33,7 +48,7 @@ internal fun getTrackKey(
         }
 
         // For Bing if it didn't have a valid cookie and for all the other search engines
-        if (hasValidCode(code, provider)) {
+        if (hasValidCode(originalCode, provider)) {
```
gabrielluong, line 51: Another option is to to call `uri.getQueryParameter(provider.codeParam)` directly here and then we won't need `originalCode`.

app/src/main/java/org/mozilla/fenix/search/telemetry/Utils.kt
```diff
@@ -22,6 +22,21 @@ internal fun getTrackKey(
 
     if (provider.codeParam.isNotEmpty()) {
         code = uri.getQueryParameter(provider.codeParam)
```
gabrielluong, line 24: I can imagine us assigning this to a `codeParam` variable. I wonder how far that could get us without introducing the `originalCode`, but continuing to modify `code`.

app/src/main/java/org/mozilla/fenix/search/telemetry/Utils.kt
rxumoz: I think this could be the easiest way.

app/src/main/java/org/mozilla/fenix/search/telemetry/Utils.kt
rxumoz: I don't quite get. Do you mean adding the "_" after the `codeParam`, or adding other parameters in `SearchProviderModel` apposed with `codeParam`?

gabrielluong (approved):

PR #19133, opened 2021-04-20
For #19125: Move top site pin net to title.

### LINKED ISSUE #19125: New Tab Page stretch goal: move pin icon from the top of shortcut tile, to next to item title
Figma: https://www.figma.com/file/lfgj15zbNkL9Pp4hdZGJgp/Proton-Mobile-UI?node-id=809%3A1

The pin icon is currently position on the top-left corner of the shortcut tile box. We’d like to move it next to the item title.

As for the icon asset, we can use the small redrawn pin icon provided in this artboard: https://www.figma.com/file/UJ5BsuvpQOADF3EPyfvtyv/Icons-WIP?node-id=2326%3A221

We can also wait to replace this asset, until it’s done by @Amejia481.

mcarare:
Fixes #19125

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/res/drawable/ic_new_pin.xml
+++ b/app/src/main/res/drawable/ic_new_pin.xml
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="12dp"
+    android:height="12dp"
+    android:viewportWidth="12"
+    android:viewportHeight="12">
+    <path
+        android:fillColor="@color/top_site_pin_icon_color"
+        android:pathData="M9.05,4.055H7.6L7.156,3.89C7.156,3.89 6.19,2.926 5.612,2.349L6.106,1.855C6.302,1.659 6.302,1.342 6.106,1.147C5.91,0.951 5.593,0.951 5.398,1.147L4.689,1.855L1.147,5.398C0.951,5.594 0.951,5.911 1.147,6.106C1.343,6.302 1.66,6.302 1.855,6.106L2.349,5.612C2.93,6.194 3.904,7.17 3.904,7.17L4.054,7.597L4.055,9.049C4.055,9.425 4.279,9.76 4.626,9.904C4.742,9.952 4.862,9.975 4.981,9.975C5.221,9.975 5.457,9.881 5.634,9.703L7.314,8.023L10.144,10.853C10.34,11.049 10.657,11.049 10.852,10.853C11.048,10.657 11.048,10.34 10.852,10.145L8.022,7.315L9.702,5.635C9.968,5.369 10.046,4.974 9.903,4.627C9.761,4.279 9.425,4.055 9.05,4.055Z" />
+</vector>
--- a/app/src/main/res/values-night/colors.xml
+++ b/app/src/main/res/values-night/colors.xml
@@ -80,7 +80,7 @@
     <color name="top_site_title_text">@color/top_site_title_text_dark_theme</color>
     <color name="top_site_pager_dot">#3A3944</color>
     <color name="top_site_pager_dot_selected">#5B5B66</color>
-    <color name="top_site_pin_icon_background_tint">@color/photonViolet50</color>
+    <color name="top_site_pin_icon_color">@color/foundation_light_theme</color>
     <color name="top_sites_border_color">@color/photonDarkGrey90</color>
 
     <!-- Synced tabs colors-->
--- a/app/src/main/res/values/colors.xml
+++ b/app/src/main/res/values/colors.xml
@@ -277,7 +277,7 @@
     <color name="top_site_title_text">@color/top_site_title_text_light_theme</color>
     <color name="top_site_pager_dot">@color/photonLightGrey30</color>
     <color name="top_site_pager_dot_selected">@color/photonLightGrey50</color>
-    <color name="top_site_pin_icon_background_tint">@color/photonInk20</color>
+    <color name="top_site_pin_icon_color">@color/foundation_dark_theme</color>
     <color name="top_sites_border_color">@color/photonLightGrey30</color>
 
     <!-- Synced tabs colors-->
--- a/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/topsites/TopSiteItemViewHolder.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/topsites/TopSiteItemViewHolder.kt
@@ -64,10 +64,11 @@ class TopSiteItemViewHolder(
     fun bind(topSite: TopSite) {
         top_site_title.text = topSite.title
 
-        pin_indicator.visibility = if (topSite.type == PINNED || topSite.type == DEFAULT) {
-            View.VISIBLE
+        if (topSite.type == PINNED || topSite.type == DEFAULT) {
+            val pinIndicator = getDrawable(itemView.context, R.drawable.ic_new_pin)
+            top_site_title.setCompoundDrawablesWithIntrinsicBounds(pinIndicator, null, null, null)
         } else {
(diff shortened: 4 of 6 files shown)


mcarare:
@brampitoyo There is no pointer in Figma as to the limit of the title width, so I set it to 84 dp (was 64 before we increased favicon from 40 to 60).

mcarare:
Intermittent test failure - org.mozilla.fenix.tabstray.NavigationInteractorTest - onOpenRecentlyClosedClicked calls navigation on DefaultNavigationInteractor .

app/src/main/res/drawable/ic_new_pin.xml
```diff
@@ -0,0 +1,9 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
gabrielluong, line 1: We should add a license for this?

app/src/main/res/values/attrs.xml
```diff
@@ -15,6 +15,7 @@
     <attr name="tabRing" format="reference" />
     <attr name="snackbar" format="reference" />
     <attr name="foundation" format="reference" />
+    <attr name="foundationContrast" format="reference" />
```
gabrielluong, line 18: Can you add a description for this attribute similar to what we did for `menuItemButtonTintColor`? This is to help preserve some knowledge about these color variables since there's no particular person that fully owns/understands all the different color palette attributes anymore. It will also be helpful to add a description now about their use case and where they are used so that it will be helpful when we map these outs in the UI for the UX.

gabrielluong, line 18: I am also wondering if we need this entry in the `attrs.xml` since I usually see another reference in `values/styles.xml`. Technically, this isn't being used at all atm.

app/src/main/res/drawable/ic_new_pin.xml
```diff
@@ -0,0 +1,9 @@
...
+    android:width="10dp"
+    android:height="10dp"
+    android:viewportWidth="10"
+    android:viewportHeight="10">
+    <path
+        android:fillColor="@color/foundation_contrast_normal_theme"
```
gabrielluong, line 7: Rather than adding a new variable that is only used here we can just keep using `top_site_pin_icon_background_tint` although perhaps with an updated name. We should also make to update or remove it.

app/src/main/res/values/colors.xml
```diff
@@ -218,6 +218,7 @@
     <color name="contrast_text_normal_theme">@color/contrast_text_light_theme</color>
     <color name="caption_text_normal_theme">@color/caption_text_light_theme</color>
     <color name="foundation_normal_theme">@color/foundation_light_theme</color>
+    <color name="foundation_contrast_normal_theme">@color/foundation_dark_theme</color>
```
gabrielluong, line 221: Based on some of the points made above, I am wondering if we want to add this variable over just using `top_site_pin_icon_background_tint`. Technically the colours are inverted instead of contrast. I would perhaps reserve adding more of these foundation variables until we figure out a component system where we do map out components colors to appropriate variables.

gabrielluong (changes requested):
Code looks good. Let's figure out the best path for the variable. I am leaning towards just using a specific `top_site_pin_icon_color` color variable since the foundation variable is only ever used once and doesn't quite satisfy the border usage everywhere yet. If we decide to keep the foundation variable, I think using `inverted` vs contrast is more appropriate.

app/src/main/res/values/attrs.xml
mcarare: Yeah, sure! I was adding this until I realised we do not need it for this use case, as we don't have top sites on private mode.

app/src/main/res/values-night/colors.xml
```diff
@@ -80,7 +80,7 @@
     <color name="top_site_title_text">@color/top_site_title_text_dark_theme</color>
     <color name="top_site_pager_dot">#3A3944</color>
     <color name="top_site_pager_dot_selected">#5B5B66</color>
-    <color name="top_site_pin_icon_background_tint">@color/photonViolet50</color>
+    <color name="top_site_pin_icon_background_tint">@color/foundation_light_theme</color>
```
gabrielluong, line 83: I think we should rename this to be 
```suggestion
    <color name="top_site_pin_icon_color">@color/foundation_light_theme</color>
```
since it's no longer used in the `android:backgroundTint` property

app/src/main/res/drawable/ic_new_pin.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="10dp"
+    android:height="10dp"
+    android:viewportWidth="10"
+    android:viewportHeight="10">
```
gabrielluong, line 9: Looking at the figma design spec more closely, I think this should be `13dp` or it might be 12dp with 1dp padding. Perhaps we should check with @brampitoyo here.

Screen Shot 2021-04-21 at 11 40 13 AM

app/src/main/res/drawable/ic_new_pin.xml
```diff
@@ -0,0 +1,13 @@
...
+    <path
+        android:fillColor="@color/top_site_pin_icon_background_tint"
+        android:pathData="M8.05,3.055H6.6L6.156,2.89C6.156,2.89 5.19,1.926 4.612,1.349L5.106,0.855C5.302,0.659 5.302,0.342 5.106,0.147C4.91,-0.049 4.593,-0.049 4.398,0.147L3.689,0.855L0.147,4.398C-0.049,4.594 -0.049,4.911 0.147,5.106C0.343,5.302 0.66,5.302 0.855,5.106L1.349,4.612C1.93,5.194 2.904,6.17 2.904,6.17L3.054,6.597L3.055,8.049C3.055,8.425 3.279,8.76 3.626,8.904C3.742,8.952 3.862,8.975 3.981,8.975C4.221,8.975 4.457,8.881 4.634,8.703L6.314,7.023L9.144,9.853C9.34,10.049 9.657,10.049 9.852,9.853C10.048,9.657 10.048,9.34 9.852,9.145L7.022,6.315L8.702,4.635C8.968,4.369 9.046,3.974 8.903,3.627C8.761,3.279 8.425,3.055 8.05,3.055Z" />
```
gabrielluong, line 12: Looking at the pathData, I think you exported the 10x10 SVG vector. I know you can also export it as 12x12 from my examination. I am not quite sure what is exactly intended. So, perhaps the best thing we can do here is have @brampitoyo export it and we use the provided pathData. This also part of the larger size question above.

gabrielluong (changes requested):

app/src/main/res/drawable/ic_new_pin.xml
mcarare: @brampitoyo Suggested using the proton icon that has 10x10. I am fine with either one.

gabrielluong: Gotcha, I am fine using 10x10 pathData. I am wondering about the final size if it should be 13x13

gabrielluong: I will ping @brampitoyo for the final review

brampitoyo (approved):
Comparing our New Tab Page to desktop’s, this new icon looks good and is visually aligned!

r+

gabrielluong:
@mcarare Just forwarding some of my conversation with @brampitoyo 

Screen Shot 2021-04-22 at 10 45 03 AM

Based on my interpretation of this, we want to go with 12x12.

app/src/main/res/drawable/ic_new_pin.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?><!-- This Source Code Form is subject to the terms of the Mozilla Public
```
gabrielluong, line 1:
```suggestion
<?xml version="1.0" encoding="utf-8"?>
<!-- This Source Code Form is subject to the terms of the Mozilla Public
```

gabrielluong (approved):

PR #19159, opened 2021-04-21
For #18608: Also add default browser item in new menu.

mcarare:
### Pull Request checklist

1. click on Show All Checks,


mcarare:
With the new menu also in Nightly, in order to run the experiment on Nightly the item should also be added to this menu, not only in the old menu that only exists in Beta and Release now.
cc @Amejia481 for confirmation.

Amejia481 (approved):
Thanks for this patch!

app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
```diff
@@ -561,6 +563,8 @@ open class DefaultToolbarMenu(
                 extensionsItem,
                 if (tabsTrayRewrite) syncMenuItem else syncedTabsItem,
                 BrowserMenuDivider(),
+                getSetDefaultBrowserItem(),
+                BrowserMenuDivider(),
```
Amejia481, line 567: If the item is not shown the divider will show either way.

app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
```diff
@@ -362,7 +362,9 @@ open class DefaultToolbarMenu(
             BrowserMenuDivider(),
             reportSiteIssuePlaceholder,
             findInPage,
+            BrowserMenuDivider(),
             getSetDefaultBrowserItem(),
+            BrowserMenuDivider(),
```
Amejia481, line 367: If the item is not shown the divider will show either way.

Amejia481 (changes requested):

app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
Amejia481: As `getSetDefaultBrowserItem()` returns null when the experiment is not active

mcarare: Good catch. I am showing them depending on the value now. Not sure, but I think one of them should be shown either way.

Amejia481 (approved):
🚀 !

app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
Amejia481: 👍🏽

Amejia481:
@Mergifyio backport releases_v89.0.0

Issue #19040, opened 2021-04-14
Turn off all Leanplum integrations on Fenix by May 31,2021


krupa-raj:
Mozilla has decided to not renew our Leanplum contract for 2021-22. The current contract will expire on May 31, 2021. We need to turn off any Leanplum integrations in our products by that date.

This work should include the following:

 - Remove the Leanplum SDK dependency from our builds so that the product does not include it anymore.
 - Remove all Leanplum related code from the product
 - Remove all Leanplum related UI from the product

Currently the _Marketing data_ setting under the _Data collection_ group says

> Shares data about what features you use in Firefox with Leanplum, our mobile marketing vendor.

When we remove Leanplum from the product, this toggle will only enable/disable _Adjust_. That should probably be reflected in the string we use for the _Marketing data_ setting.

@amedyne @rvandermeulen Because of the adjusted 89 schedule, we probably need to ship this change in a 88 update on May 31st.

rvandermeulen:
> Because of the adjusted 89 schedule, we probably need to ship this change in a 88 update on May 31st.

@krupa-raj So I'm clear, do we need to ship this **on** May 31st or **by** May 31st?

pocmo:
I'll look into preparing a PR for making this changes in the 88 release branch.

ItsDevyani:
Just an FYI: We have default browser campaigns live in iOS release and Fenix Nightly and Release. I can stop the Nightly campaign if need be. I was thinking of stopping campaigns on May 21st, but since we are thinking of an 88 dot release I can stop as you chose a date with Rel Eng, Just slack me once you have a decision.

amedyne:
@topotropic FYI. Here's the corresponding GH ticket referred in the email with subject: "Removing Leanplum from Fenix / Needs string change in settings"

pocmo:
We have removed Leanplum from Nightly (string not updated yet). Beta and Release PRs are up, but not ready for landing yet (needs coordination).

pocmo:
Added the "qa needed" label for Nightly. Overall the patch does not seem risky. The "push" integration needed to change in order to remove Leanplum. So one thing we could explicitly test is that web push is still working correctly.

lobontiumira:
Hi @pocmo!

We cannot identify new devices that use the latest Nightly build from 4/28.
We've also tried to send a push notification to a test device that was registered previously and bumped into this error "No push token exist for this device".
[...]

PR #19186, opened 2021-04-22
Issue #19040: Remove Leanplum (Nightly)

pocmo:
This PR removes Leanplum from the codebase in the `master` branch (Nightly, 90.0).

* This patch does not update the settings string containing "Leanplum" yet. It's still to be discussed what the right string has to be.
* @jonalmeida Can you take a look at the push changes in this PR?
* The patch applies to beta and release too. I'll open separate PRs once this has gone through a first review and automation is happy.
* Please review, but do not land yet since this needs a bit of coordination.


app/src/main/java/org/mozilla/fenix/push/FirebasePushService.kt
```diff
@@ -5,64 +5,12 @@
...
 import android.annotation.SuppressLint
-import com.google.firebase.messaging.RemoteMessage
-import com.google.firebase.messaging.FirebaseMessagingService
-import com.leanplum.LeanplumPushFirebaseMessagingService
-import com.leanplum.LeanplumPushService
-import mozilla.components.concept.push.PushService
-import mozilla.components.lib.push.firebase.AbstractFirebasePushService
 import mozilla.components.feature.push.AutoPushFeature
-
-/**
```
jonalmeida, line 16: 🔥

jonalmeida (approved):
> @jonalmeida Can you take a look at the push changes in this PR?

So good to see that hack we wrote for push going away! 😄

app/src/main/java/org/mozilla/fenix/components/BackgroundServices.kt
```diff
@@ -217,13 +217,11 @@ internal class TelemetryAccountObserver(
             metricController.track(it)
         }
 
-        // Used by Leanplum as a context variable.
         settings.fxaSignedIn = true
```
jonalmeida, line 220: Should we remove these settings as well?

pocmo: At first I removed it and then I was wondering whether we would want to keep it for other things. But maybe you are right, no need to keep for now, I'll get rid of it.

pocmo:
There are no active Leanplum campaigns on Nightly anymore. This can land. We can deal with the settings string in a follow-up once a decision has been made.

Issue #12131, opened 2020-06-30
[Telemetry] `string_list` metric type is being misused


Dexterp37:
We just saw that #11118 and #11446: looks like [`string_list`](https://mozilla.github.io/glean/book/user/metrics/string_list.html) is being used to record single values that are bools.

This would make it hard for tools to understand the intent and pushes the burden of parsing data (which is now string data) on the analysis side, increasing both the cost and the likelihood of having bad data.

The [code states the following](https://searchfox.org/mozilla-mobile/rev/6c90ba48ca3bd04c7db3efeab23640e3d90fdd4a/fenix/app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt#742-745):

```
        // We purposefully make all of our preferences the string_list format to make data analysis
        // simpler. While it makes things like booleans a bit more complicated, it means all our
        // preferences can be analyzed with the same dashboard and compared.
```

Setting aside the fact that that this will break GLAM, how does having a list of string having a single value of the wrong type make analysis simpler?

ekager:
cc @sblatz looks like this was a recent addition - do you know why we did this?

Dexterp37:
We had a conversation over Slack with @sblatz and Marissa Gorlick. Let me try to capture what was said, for future reference.

The decision to use `string_list` was made in order to make all "data types the same" to ease future dashboards, in response to Marissa's request for standardization.

While the intent was noble, we all agreed that this was not the right path to pursue, as "standardization" does not necessarily imply "all the is sent as string", which is both a problem for tools and for storage.

The path forward here is:

- boolean values (toggle) can be a `boolean` type
- metric that only record 1 string values, are `string` type
- metrics that record more than 1 string value (e.g. accessibility services) are `string_list` type

While this is a very important change to make, right now we're in code freeze so it can't happen right away.

Any standardization can happen at analysis side, if needed, through UDFs.

I'm happy yo provide support with reviews as needed!

Dexterp37:
cc @mdboom

PR #19252, opened 2021-04-26
Closes #19147: Move startup metrics to right after Glean initialization

### LINKED ISSUE #19147: Investigate into null values in startup metrics
Currently there are large number of startup metrics are reported as null.  Investigate into how startup metrics are reported to understand the cause/solution of the issue.

Example query: https://sql.telemetry.mozilla.org/queries/79505/

rocketsroger:
### Pull Request checklist

1. click on Show All Checks,


gabrielluong:
Is there also a data eng/sci peer that can take a look at this since we are also updating the glean types for some metrics?

rocketsroger:
> Is there also a data eng/sci peer that can take a look at this since we are also updating the glean types for some metrics?

Sure, I'll add reviewer from data science.

rocketsroger:
@Dexterp37 please confirm that this change is what we still want for https://github.com/mozilla-mobile/fenix/issues/12131.  Thanks,

app/metrics.yaml
```diff
@@ -1270,7 +1270,7 @@ metrics:
 
 preferences:
   show_search_suggestions:
-    type: string_list
+    type: boolean
```
badboy, line 1273: Just for your awareness: This will work, but land in a different column.
Old: `metrics.string_list.preferences_show_search_suggestions`
New: `metrics.boolean.preferences_show_search_suggestions`

That means comparing these across versions is a bit more involved, but not impossible.
(There should never be a client that has both set though)

(I'm still waiting for confirmation that what I said is actually true)

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -499,6 +512,167 @@ open class FenixApplication : LocaleAwareApplication(), Provider {
         }
     }
 
+    /**
+     * This function is called right after Glean is initialized.  Part of this function depends on
+     * shared preferences to be updated so the correct value is sent with the metrics ping.
+     *
+     * The reason we're using shared preferences to track some of these values is due to the
+     * limitations of the metrics ping.  Events are only sent in a metrics ping if the user have made
+     * changes between each ping.  However, in some cases we want current values to be sent even if
```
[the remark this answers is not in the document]
badboy: Oh I see now that is just copied from the old code.
Still wonder what this part means.

badboy (changes requested):
Blocking here while I wait for confirmation that just changing the type won't break anything.

This is now code that's executed earlier, is there any way to verify that it has a neglibile impact on startup?

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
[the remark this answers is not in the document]
rocketsroger: Yes, this was copied from the original function.  I will update it so it makes more sense.  Thanks,

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -499,6 +512,167 @@ open class FenixApplication : LocaleAwareApplication(), Provider {
...
+     * the user have not changed anything between pings.
```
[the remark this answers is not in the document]
rocketsroger: Sure, I'll update this since I'll rewrite this.  Not sure why originally this was double spaced.

app/metrics.yaml
rocketsroger: Yes, please confirm that this is OK.  Since this workaround is because currently there are high number of startup metrics that have null values, I would argue that comparing against data we collected before is not very useful.

rocketsroger:
> This is now code that's executed earlier, is there any way to verify that it has a neglibile impact on startup?

Yes, I think this will have some impact on startup performance.  However, this is currently the only workaround suggestion from the Glean team.  Once the Glean investigation is over we will update this if there is a better solution.

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -137,6 +148,8 @@ open class FenixApplication : LocaleAwareApplication(), Provider {
                 false -> "Mozilla"
             }
         )
+
+        setStartupMetrics(components.core.store, settings())
```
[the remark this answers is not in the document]
rocketsroger: Agreed.  I will move this into the `setStartupMetrics`.  Thanks,

gabrielluong:
Will take another look at this thoroughly tomorrow.

app/metrics.yaml
mdboom: Yeah, I agree it sounds like it's worthwhile to make the change.  You still may want to add in the description field "Prior to release X, this metric was of type `string_list`" so that *if* someone wanted to trace back through the (problematic) historical data they'd at least know how...

gabrielluong:
Sorry, didn't get to this today, will review it tomorrow!

app/metrics.yaml
acmiyaguchi: This kind of change is not supported by our tooling, a similar change of a string to a datetime in the past has caused delays on our schema deploys: https://github.com/mozilla/mozilla-schema-generator/issues/118

Type changes is typically frowned upon at a low level because it [breaks schema evolution rules](https://cloud.google.com/bigquery/docs/managing-table-schemas) that allows for low operational overhead during ingestion time. It's possible to support changing types in the schema generator based on the behavior of moving the probe to a new namespace (which avoids a schema incompatible change). However, this will be confusing from a metadata perspective since we lose context on the type change.

I suggest expiring these probes and creating a new probe with the correct type, while clearly documenting the reason for deprecation so that tools like the [glean dictionary](https://dictionary.telemetry.mozilla.org/) can pick this up.

gabrielluong: @rocketsroger Can you ping me when you think you need a final review? Thanks!

rocketsroger: @gabrielluong Sure, this should be the final change but I'll ping you once I have the go ahead from the data science team.  Thanks,

badboy (approved):
Looks good to me now!

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -499,6 +504,175 @@ open class FenixApplication : LocaleAwareApplication(), Provider {
...
+                }
+            }
+        }
+    }
+
+    @Suppress("ComplexMethod")
+    private fun setPreferenceMetrics(
+        settings: Settings
+    ) {
+        // We purposefully make all of our preferences the string_list format to make data analysis
```
gabrielluong, line 609: I imagine we should change this comment if not remove it entirely since they're boolean now. I think this comment shouldn't be applied to the `with` block either since it only applies to line 613-623.

gabrielluong (approved):
Do we need to check with the perf-team?

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
rocketsroger: Good catch.   Thanks I'll update

app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt
```diff
@@ -856,10 +846,7 @@ private val Event.wrapper: EventWrapper<*>?
     }
 
 class GleanMetricsService(
```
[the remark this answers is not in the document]
rocketsroger: Sure, I'll try to describe this service.

rocketsroger:
> Do we need to check with the perf-team?

I'll notify the perf-team.  This will affect the startup performance but unfortunately there's no other solutions.

PR #19266, opened 2021-04-27

MickeyMoz:


pocmo:
Previous failure couldn't find artifacts on `nightly.maven.mozilla.org` even though I was able to download them.

Next run failed with a unit test failure that seems unrelated:
https://firefox-ci-tc.services.mozilla.com/tasks/XIZWBesWTlugo9INXCXOxg/runs/0

Issue #10941, opened 2020-05-27
Use somethig better than 0.0a1 for the Nightly version


kbrosnan:
Can we increment this at the start of every sprint, use something like YearWeekNumber or Gecko version? It would help segment out very old crashes in nightly from crash-stats.mozilla.org

kbrosnan:
Example: This AddonManagerException has been fixed for ages but still shows up as the 3rd most common crash. 

* https://crash-stats.mozilla.org/topcrashers/?product=Fenix&version=0.0a1
* https://crash-stats.mozilla.org/signature/?product=Fenix&version=0.0a1&signature=%5BINFO%5D%20mozilla.components.feature.addons.AddonManagerException%3A%20at%20mozilla.components.feature.addons.AddonManager.getAddons%28AddonManager.kt%3A56%29&_columns=date&_columns=product&_columns=version&_columns=build_id&_columns=platform&_columns=reason&_columns=address&_columns=install_time&_columns=startup_crash&_sort=-date&page=1#reports

eliserichards:
@liuche would you like to take a look at this? Not sure how simple it would be 🤔

kbrosnan:
Did some looking into this and the 0.0a1 comes from the Socorro code. They don't have the ability to parse `Nightly YYMMDD HH:MM` into a format that they can handle. It would be best to just send the Geckoview version. See https://github.com/mozilla-services/socorro/commit/df7e3e17d5a0c9f4827e09904adf11a65582c956 and bz-1624911

pocmo:
Grabbing this issue: Since we have synchronized versions/trains now (GV 90 <-> A-C 90 <-> Fenix 90), we should be able to align our Nightly versioning too (e.g. like desktop use 90.0a1, 91.0a1, ...).

pocmo:
Landed PR and triggered automation.

Looks like the first 90.0a1 made it safely through automation and was uploaded to Google Play. Let's wait for it to get distributed.

pocmo:
I did a quick some test of the build and it seems to work as expected and shows the new version number. A forced crash was uploaded to Socorro as well. On Google Play the version is currently "In Review". Let's see once it reaches users.

pocmo:
Crash Stats:
[...]

PR #19286, opened 2021-04-28
Issue #10941: Align Nightly version with Gecko/A-C version.

pocmo:
Instead of using a timestamp this will synchronize the Nightly version with the version
of Gecko / A-C. So when using 90.0.20210426143115 then the Nightly version will be
90.0a1 - similar to what Firefox desktop uses.


pocmo:
Once this is review and lands, I'll trigger a Nightly build and watch automation for fallout.

@rocketsroger Afaik this should just work with Socorro without any further changes, right? After that I think we can remove the workaround for mapping timestamp versions to `0.0a1` (although those older versions will still report crashes for a while, so maybe let's keep it for now?).

eliserichards (approved):
Awesome! Let's see if this works 😄

PR #19303, opened 2021-04-29
Issue #19040 - Update marketing data setting description

pocmo:
Follow-up from #19186. This adds the new string for the settings description.


pocmo:
Aye, missed a ui test that uses the string too.

jonalmeida (approved):

PR #19332, opened 2021-04-30
Bug 1708631 - Update visual metrics script to the latest.

gmierz:
This patch updates the visual metric script with changes from mozilla-central. It fixes the perma-fails that we are currently seeing in the browsertime tests on the fenix branch.


jmaher (approved):
rubber stamp from releng

PR #19397, opened 2021-05-05
Issue #19147: Move set startup metrics off main thread

### LINKED ISSUE #19147: Investigate into null values in startup metrics

rocketsroger:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
+++ b/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
@@ -141,7 +141,11 @@ open class FenixApplication : LocaleAwareApplication(), Provider {
             buildInfo = GleanBuildInfo.buildInfo
         )
 
-        setStartupMetrics(components.core.store, settings())
+        // We avoid blocking the main thread on startup by setting startup metrics on the background thread.
+        val store = components.core.store
+        GlobalScope.launch(Dispatchers.IO) {
+            setStartupMetrics(store, settings())
+        }
     }
 
     @CallSuper
@@ -622,10 +626,7 @@ open class FenixApplication : LocaleAwareApplication(), Provider {
             privateSearchSuggestions.set(settings.shouldShowSearchSuggestionsInPrivate)
             voiceSearchEnabled.set(settings.shouldShowVoiceSearch)
             openLinksInAppEnabled.set(settings.openLinksInExternalApp)
-
-            val isLoggedIn =
-                components.backgroundServices.accountManager.accountProfile() != null
-            signedInSync.set(isLoggedIn)
+            signedInSync.set(settings.signedInFxaAccount)
 
             val syncedItems = SyncEnginesStorage(applicationContext).getStatus().entries.filter {
                 it.value
--- a/app/src/main/java/org/mozilla/fenix/components/BackgroundServices.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/BackgroundServices.kt
@@ -188,6 +188,7 @@ internal class TelemetryAccountObserver(
     private val metricController: MetricController
 ) : AccountObserver {
     override fun onAuthenticated(account: OAuthAccount, authType: AuthType) {
+        settings.signedInFxaAccount = true
         when (authType) {
             // User signed-in into an existing FxA account.
             AuthType.Signin -> Event.SyncAuthSignIn
@@ -220,5 +221,6 @@ internal class TelemetryAccountObserver(
 
     override fun onLoggedOut() {
         metricController.track(Event.SyncAuthSignOut)
+        settings.signedInFxaAccount = false
     }
 }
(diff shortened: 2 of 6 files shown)


rocketsroger:
As discussed offline, I'll get this merged before performance team can do another round of tests on the build.  Removing @mcomella as assigned reviewer.  Thanks

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -141,7 +141,13 @@ open class FenixApplication : LocaleAwareApplication(), Provider {
             buildInfo = GleanBuildInfo.buildInfo
         )
 
-        setStartupMetrics(components.core.store, settings())
+        // We avoid blocking the main thread on startup by setting startup metrics on the background thread.
+        val isLoggedIn =
+            components.backgroundServices.accountManager.accountProfile() != null
```
jonalmeida, line 146: Moving the account manager to this place would also early initialize it during the `Application` creation which we would want to avoid.

I think using a pref to store the `loggedIn` state when the account manager is done initializing might be a quick fix for this. It's something we're doing for other settings that we want to submit to the ping as well. All future reads of this pref should give you the right value.

An alternative, is if observe the account manager and submit all these metrics when everything is done. I'm not sure how Glean SDK is required to work to know if this will cause faults in our data though, we're already trying to solve this issue right now after all.

jonalmeida:
I discussed this with Roger offline since I'm also working on a similar problem with Glean in Focus as well. One suggestion that might work is if we can set all our values first, before we submit them. Something like the code below:

```kotlin
val waitForMetrics = CoroutineScope(Dispatchers.IO).async {
  Metrics.distributionId.set(value)
  Metrics.defaultBrowser.set(otherValue)
}

MainScope().launch {
  waitForMetrics.await()
  activationPing.checkAndSend()
  installationPing.checkAndSend()
}
```

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
rocketsroger: Good catch, I'll look into solving this issue.

mdboom:
That sounds like a good plan, thanks!

mcomella:
> However, we have not confirmed #19252 does fix the problem.

Since #19252 is in Nightly, I wonder if we can look at the data returned in Nightly to determine if the current fix solves the issue. Then we can land this to-the-background-thread change and see how it changes the result.

rocketsroger:
> > However, we have not confirmed #19252 does fix the problem.
> 
> Since #19252 is in Nightly, I wonder if we can look at the data returned in Nightly to determine if the current fix solves the issue. Then we can land this to-the-background-thread change and see how it changes the result.

Did a quick query limiting results from the last two days.  There are still about 72% of values are null.  However, versus last 28 days of 80% it's an improvement.  Here's the query

rocketsroger:
As discussion offline with @mcomella I'll hold of merging this change until next week.  This gives us more data to confirm that the workaround does work when it's done on the main thread.

badboy:
> > > However, we have not confirmed #19252 does fix the problem.
> > 
> > 
> > Since #19252 is in Nightly, I wonder if we can look at the data returned in Nightly to determine if the current fix solves the issue. Then we can land this to-the-background-thread change and see how it changes the result.
> 
> Did a quick query limiting results from the last two days. There are still about 72% of values are null. However, versus last 28 days of 80% it's an improvement. Here's the query.
> `SELECT metrics.metrics.string.metrics_tab_view_setting AS TYPE, count(*) AS number FROM org_mozilla_fenix.metrics WHERE DATE(submission_timestamp) > DATE_SUB(CURRENT_DATE, INTERVAL 2 DAY) GROUP BY 1`

[If you filter by those client that might have the fix](https://sql.telemetry.mozilla.org/queries/79932/source#198470) (based on the app build & commit date of the fix, not sure if there's a better way to match into what build a certain commit went), then only 7% still sent NULL values for `metrics_tab_view_setting`.
So the fix definitely works as expected.

I agree that an 8% slowdown is not acceptable right now. I'll prioritize the investigation on our end this week.

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -141,7 +141,11 @@ open class FenixApplication : LocaleAwareApplication(), Provider {
             buildInfo = GleanBuildInfo.buildInfo
         )
 
-        setStartupMetrics(components.core.store, settings())
+        // We avoid blocking the main thread on startup by setting startup metrics on the background thread.
+        val store = components.core.store
+        GlobalScope.launch(Dispatchers.Default) {
```
jonalmeida, line 146: nit: This should probably be the `IO` dispatcher since we're reading prefs from disk for this change.

rocketsroger:
Since we can now confirm that the original solution of moving startup metrics to after Glean initialization works as expected.

jonalmeida (approved):
I ran this patch through the profiler and confirmed with a comparison (without the changes) that we get back our gains in pref.

We can verify the Glean bits work after this has landed in that case.

PR #19403, opened 2021-05-06
Complete browser-state migration and remove browser-session dependency.

pocmo:
This is the matching branch for the changes in A-C: https://github.com/mozilla-mobile/android-components/pull/10212

With that Fenix will no longer depend on `browser-session` and will only use `browser-state`.

We want to wait with merging until the beginning of a new Nightly cycle.


csadilek (approved):
🎉 🚀

csadilek:
Network issue, restarting.

Amejia481:
🎉 !

Issue #19054, opened 2021-04-15
Navigate back to the Credit card screen when the last credit card is deleted


gabrielluong:
After deleting the last card the user is brought to an empty "Saved cards" screen (this will still have the "Add credit card" button despite no saved credit cards to list). We should determine whether or not we should go back to the "Saved cards" screen or back to the "Credit cards" screen.

gabrielluong:
@topotropic pinging you on this

topotropic:
Thanks! Good point – let's go back to the "Credit card" screen once people delete their last card.

abodea:
Verified as fixed on the latest Nightly 5/20 with Google Pixel 4 XL (Android 11).
Please note that after deleting the credit card I was redirected without issues to the `Credit cards` section.

Issue #15168, opened 2020-09-17
[Bug] Intermittent failure debug(T) sets up interactor listeners


Archaeopteryx:
Log: https://treeherder.mozilla.org/logviewer.html#/jobs?job_id=315979488&repo=fenix&lineNumber=5319

[task 2020-09-17T17:45:44.912Z]   TEST: sets up interactor listeners
[task 2020-09-17T17:45:44.913Z]     [Robolectric] org.mozilla.fenix.search.toolbar.ToolbarViewTest.sets up interactor listeners: sdk=28; resources=LEGACY
[task 2020-09-17T17:45:44.913Z]     [Robolectric] NOTICE: legacy resources mode is deprecated; see http://robolectric.org/migrating/#migrating-to-40
[task 2020-09-17T17:45:44.913Z]     I/AssetManager: no overlays.list file found
[task 2020-09-17T17:45:44.913Z]     D/StrictMode: onThreadPolicyViolation; policy=65599
[task 2020-09-17T17:45:44.913Z]     D/StrictMode: StrictMode policy violation; ~duration=0 ms: android.os.strictmode.DiskWriteViolation
[task 2020-09-17T17:45:44.913Z]     at android.os.StrictMode$AndroidBlockGuardPolicy.$$robo$$android_os_StrictMode_AndroidBlockGuardPolicy$onWriteToDisk(StrictMode.java:1460)
[task 2020-09-17T17:45:44.913Z]     at android.os.StrictMode$AndroidBlockGuardPolicy.onWriteToDisk(StrictMode.java)
[task 2020-09-17T17:45:44.914Z]     at libcore.io.BlockGuardOs.$$robo$$libcore_io_BlockGuardOs$mkdir(BlockGuardOs.java:196)
[task 2020-09-17T17:45:44.914Z]     at libcore.io.BlockGuardOs.mkdir(BlockGuardOs.java)
[task 2020-09-17T17:45:44.914Z]     at android.system.Os.$$robo$$android_system_Os$mkdir(Os.java:323)
[task 2020-09-17T17:45:44.914Z]     at android.system.Os.mkdir(Os.java)
[task 2020-09-17T17:45:44.914Z]     at android.app.ContextImpl.$$robo$$android_app_ContextImpl$ensurePrivateDirExists(ContextImpl.java:608)
[task 2020-09-17T17:45:44.914Z]     at android.app.ContextImpl.ensurePrivateDirExists(ContextImpl.java)
[...]

PR #19451, opened 2021-05-10
For #19054: Navigate to the Credit card screen when deleting last card.

mcarare:
Fixess #19054

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardsManagementFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardsManagementFragment.kt
@@ -61,6 +61,11 @@ class CreditCardsManagementFragment : Fragment() {
     @ExperimentalCoroutinesApi
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         consumeFrom(creditCardsStore) { state ->
+            if (!state.isLoading && state.creditCards.isEmpty()) {
+                findNavController().popBackStack()
+                return@consumeFrom
+            }
+
             creditCardsView.update(state)
         }
     }


app/src/main/java/org/mozilla/fenix/settings/creditcards/view/CreditCardsManagementView.kt
```diff
@@ -41,6 +41,9 @@ class CreditCardsManagementView(
         credit_cards_list.isVisible = state.creditCards.isNotEmpty()
 
         creditCardsAdapter.submitList(state.creditCards)
+        if (!state.isLoading && state.creditCards.isEmpty()) {
+            interactor.onCreditCardListEmptied()
+        }
```
gabrielluong, line 46: Let's add a new line before these kind of if statement blocks to provide better code separation to make it easier to read.

```suggestion

        if (!state.isLoading && state.creditCards.isEmpty()) {
            interactor.onCreditCardListEmptied()
        }
```

app/src/main/java/org/mozilla/fenix/settings/creditcards/view/CreditCardsManagementView.kt
```diff
@@ -41,6 +41,9 @@ class CreditCardsManagementView(
         credit_cards_list.isVisible = state.creditCards.isNotEmpty()
 
         creditCardsAdapter.submitList(state.creditCards)
+        if (!state.isLoading && state.creditCards.isEmpty()) {
```
gabrielluong, line 44: I am trying to stick with @Mugurell's philosophy around `View` described here https://github.com/mozilla-mobile/fenix/pull/18808#discussion_r608822058. I am wondering if we should just do this check at https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardsManagementFragment.kt#64 to avoid having this logic in the View.

Otherwise, I would recommend moving this to top of `update` to avoid the unnecessary UI updates before popping the nav stack.

@Mugurell also interested in your thoughts here.

gabrielluong: Perhaps another approach would be to check if we have any credit cards after deleting and decide where to navigate in https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/java/org/mozilla/fenix/settings/creditcards/controller/CreditCardEditorController.kt#65.

Mugurell: This is an interesting discussion.
We have Middlewares that sit between Actions and Reducers but no clear way to intercept and act on new States before they get to Views.
Asked Christian about this and he recommended "Handling the state change in consumeFrom is the right place. You can decide right there in the fragment that the view should be closed and call a method on the view directly."

mcarare: @gabrielluong Following the first suggestion I would rather have `CreditCardsManagementFragment` decide to navigate back before updating the view.

Mugurell: Throughout the app there are various ways to handle this situation. It would be great to focus on one particular approach from now on.

mcarare:
Intermittent test failure: https://github.com/mozilla-mobile/fenix/issues/15168.

app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardsManagementFragment.kt
```diff
@@ -61,6 +61,11 @@ class CreditCardsManagementFragment : Fragment() {
     @ExperimentalCoroutinesApi
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         consumeFrom(creditCardsStore) { state ->
+
+            if (state.creditCards.isEmpty()) {
+                findNavController().popBackStack()
+            }
```
gabrielluong, line 67:
```suggestion
            if (state.creditCards.isEmpty()) {
                findNavController().popBackStack()
            }
```

gabrielluong (approved):
@mcarare @Mugurell Before we land this, I am wondering what we thought about the approach where we called `getAllCreditCards` after `deleteCreditCard` and checked if it was empty and decide where we navigate to in the controller https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/java/org/mozilla/fenix/settings/creditcards/controller/CreditCardEditorController.kt#65. Did we prefer calling `popBackStack` to avoid the possibility of going back to the `CreditCardsManagementFragment`?

mcarare:
@gabrielluong From my point of view, I wanted to avoid giving more responsibility to the controller, so the method would have to handle more than just deleting the card. It would also be required to get the updated credit card list and decide based on it where to navigate. In some cases maybe a risk of leaking the view could arise. 

As for the issue you mentioned, avoiding the possibility of going back to the `CreditCardsManagementFragment`, if we navigate from the controller to the `CreditCardsManagementFragment`, we would also have the editor fragment in the back stack and we would navigate to it when pressing back button. But this can be avoided using `popBackStack` with the `creditCardsManagementFragment` id as an argument and `inclusive = true`.

Issue #17537, opened 2021-01-20
Expose credit card & addresses syncing in Account Settings UI


grigoryk:
Issue capturing larger eng work: https://github.com/mozilla-mobile/android-components/issues/9452

## User Story
- As a user, I want to be able to enable and disable credit cards syncing
- As a user, I want to be able to enable and disable addresses syncing

## Dependencies
- https://github.com/mozilla-mobile/android-components/issues/9456

### Acceptance Criteria
- I can do see status of, and enable/disable credit cards syncing under Account Settings
- I can do see status of, and enable/disable addresses syncing under Account Settings

grigoryk:
This will need to be wired up in AccountsSettingsFragment.

gabrielluong:
The credit card sync engine should be available with the latest AC/AS update in https://github.com/mozilla-mobile/fenix/pull/19195. We should wire up the UI, but only display it with the credit card autofill pref enabled.

abodea:
Verified as fixed on the latest Nightly 5/18 with Google Pixel 4 XL (Android 11).

PR #19471, opened 2021-05-11
For #17537: Add preferences for syncing credit cards and addresses

eliserichards:
For #17537 

* Visibility for CC and addresses based on respective Feature Flags
* Device pin needed to enable CC syncing using same logic as Logins (new copy TBD)
* Default values and visibility for CC and address preferences are `false`

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/settings/account/AccountSettingsFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/account/AccountSettingsFragment.kt
@@ -35,6 +35,7 @@ import mozilla.components.service.fxa.sync.SyncReason
 import mozilla.components.service.fxa.sync.SyncStatusObserver
 import mozilla.components.service.fxa.sync.getLastSynced
 import mozilla.components.support.ktx.android.content.getColorFromAttr
+import org.mozilla.fenix.FeatureFlags
 import org.mozilla.fenix.R
 import org.mozilla.fenix.components.FenixSnackbar
 import org.mozilla.fenix.components.StoreProvider
@@ -176,10 +177,17 @@ class AccountSettingsFragment : PreferenceFragmentCompat() {
             SyncEngine.Bookmarks -> R.string.pref_key_sync_bookmarks
             SyncEngine.Passwords -> R.string.pref_key_sync_logins
             SyncEngine.Tabs -> R.string.pref_key_sync_tabs
+            SyncEngine.CreditCards -> R.string.pref_key_sync_credit_cards
+            SyncEngine.Addresses -> R.string.pref_key_sync_address
             else -> throw IllegalStateException("Accessing internal sync engines")
         }
 
-        listOf(SyncEngine.History, SyncEngine.Bookmarks, SyncEngine.Tabs).forEach {
+        listOf(
+            SyncEngine.History,
+            SyncEngine.Bookmarks,
+            SyncEngine.Tabs,
+            SyncEngine.Addresses
+        ).forEach {
             requirePreference<CheckBoxPreference>(it.prefId()).apply {
                 setOnPreferenceChangeListener { _, newValue ->
                     updateSyncEngineState(context, it, newValue as Boolean)
@@ -188,7 +196,7 @@ class AccountSettingsFragment : PreferenceFragmentCompat() {
             }
         }
 
-        // 'Passwords' listener is special, since we also display a pin protection warning.
+        // 'Passwords' and 'Credit card' listeners are special, since we also display a pin protection warning.
         requirePreference<CheckBoxPreference>(SyncEngine.Passwords.prefId()).apply {
             setOnPreferenceChangeListener { _, newValue ->
                 val manager =
@@ -199,7 +207,23 @@ class AccountSettingsFragment : PreferenceFragmentCompat() {
                 ) {
                     updateSyncEngineState(context, SyncEngine.Passwords, newValue as Boolean)
                 } else {
-                    showPinDialogWarning(newValue as Boolean)
+                    showPinDialogWarning(SyncEngine.Passwords, newValue as Boolean)
+                }
+                true
+            }
+        }
+
+        requirePreference<CheckBoxPreference>(SyncEngine.CreditCards.prefId()).apply {
+            setOnPreferenceChangeListener { _, newValue ->
+                val manager =
+                    activity?.getSystemService(Context.KEYGUARD_SERVICE) as KeyguardManager
+                if (manager.isKeyguardSecure ||
+                    newValue == false ||
+                    !context.settings().shouldShowSecurityPinWarningSync
+                ) {
+                    updateSyncEngineState(context, SyncEngine.CreditCards, newValue as Boolean)
+                } else {
+                    showPinDialogWarning(SyncEngine.CreditCards, newValue as Boolean)
                 }
                 true
             }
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/mozilla/fenix/settings/account/AccountSettingsFragment.kt
```diff
@@ -176,10 +177,17 @@ class AccountSettingsFragment : PreferenceFragmentCompat() {
...
+            SyncEngine.Addresses -> R.string.pref_key_sync_address
             else -> throw IllegalStateException("Accessing internal sync engines")
         }
 
-        listOf(SyncEngine.History, SyncEngine.Bookmarks, SyncEngine.Tabs).forEach {
+        listOf(
+            SyncEngine.History,
+            SyncEngine.Bookmarks,
+            SyncEngine.Tabs,
+            SyncEngine.Addresses
```
gabrielluong, line 189: Are wee missing CreditCards here?

app/src/main/java/org/mozilla/fenix/settings/account/AccountSettingsFragment.kt
```diff
@@ -199,7 +207,23 @@ class AccountSettingsFragment : PreferenceFragmentCompat() {
...
+                val manager =
+                    activity?.getSystemService(Context.KEYGUARD_SERVICE) as KeyguardManager
+                if (manager.isKeyguardSecure ||
+                    newValue == false ||
+                    !context.settings().shouldShowSecurityPinWarningSync
+                ) {
+                    updateSyncEngineState(context, SyncEngine.CreditCards, newValue as Boolean)
+                } else {
+                    showPinDialogWarning(SyncEngine.CreditCards, newValue as Boolean)
                 }
```
gabrielluong, line 227: Can we refactor this out to a helper function that can be called by both with the SyncEngine as a param?

app/src/main/res/xml/account_settings_preferences.xml
```diff
@@ -29,6 +29,12 @@
             android:layout="@layout/checkbox_left_preference"
             android:title="@string/preferences_sync_bookmarks" />
 
+        <CheckBoxPreference
+            android:defaultValue="true"
```
gabrielluong, line 33: I think we set the default visibility for both of these to false to prevent it from flashing and disappearing depending on the prefs.

app/src/main/java/org/mozilla/fenix/settings/account/AccountSettingsFragment.kt
eliserichards: Nope, credit cards and passwords are below (line 200 & 216)

app/src/main/java/org/mozilla/fenix/settings/account/AccountSettingsFragment.kt
eliserichards: Yes! good idea

app/src/main/java/org/mozilla/fenix/settings/account/AccountSettingsFragment.kt
```diff
@@ -218,7 +218,27 @@ class AccountSettingsFragment : PreferenceFragmentCompat() {
...
+            || !newValue
+            || !requireContext().settings().shouldShowSecurityPinWarningSync
+        ) {
+            updateSyncEngineState(context, SyncEngine.CreditCards, newValue)
+        } else {
+            showPinDialogWarning(SyncEngine.CreditCards, newValue)
+        }
+    }
+
+    private fun updateSyncEngineState(context: Context, engine: SyncEngine, newState: Boolean) {
```
eliserichards, line 234: moved from above (line 167)

app/src/main/java/org/mozilla/fenix/settings/account/AccountSettingsFragment.kt
gabrielluong: Yes, this makes sense now.

app/src/main/res/xml/account_settings_preferences.xml
```diff
@@ -29,6 +29,12 @@
             android:layout="@layout/checkbox_left_preference"
             android:title="@string/preferences_sync_bookmarks" />
 
+        <CheckBoxPreference
```
gabrielluong, line 32: Still looking for the visibility to be false by default for these to prevent the flash and disappear.

app/src/main/java/org/mozilla/fenix/settings/account/AccountSettingsFragment.kt
```diff
@@ -218,7 +218,27 @@ class AccountSettingsFragment : PreferenceFragmentCompat() {
         )
     }
 
-    private fun showPinDialogWarning(newValue: Boolean) {
+    private fun CheckBoxPreference.updateSyncEngineStateWithPinWarning(newValue: Boolean) {
```
gabrielluong, line 220: Let's add a kDoc for this.

app/src/main/java/org/mozilla/fenix/settings/account/AccountSettingsFragment.kt
gabrielluong: I can imagine people being worried about losing the initial commit history associated with this. To mitigate that concern, I would prefer we also improve the code quality by adding a kDoc with its @param to ensure the loss of history is worthwhile. I would also just check the git blame to see if this function was part of some bigger feature. If it is, I would just note that in the PR or issue somewhere in case, anyone ever wants to refer back to it or weigh whether or not it is worthwhile to make this change.

app/src/main/res/xml/account_settings_preferences.xml
eliserichards: oh I misread your comment. I thought you meant default Check value

kaream671:
17537

app/src/main/java/org/mozilla/fenix/settings/account/AccountSettingsFragment.kt
```diff
@@ -218,7 +217,53 @@ class AccountSettingsFragment : PreferenceFragmentCompat() {
...
+     * @param newValue the value denoting whether or not to sync the specified preference.
+     */
+    private fun CheckBoxPreference.updateSyncEngineStateWithPinWarning(newValue: Boolean) {
+        val manager =
+            activity?.getSystemService(Context.KEYGUARD_SERVICE) as KeyguardManager
+        if (manager.isKeyguardSecure ||
+            !newValue ||
+            !requireContext().settings().shouldShowSecurityPinWarningSync
+        ) {
+            updateSyncEngineState(context, SyncEngine.CreditCards, newValue)
```
gabrielluong, line 235: 🚨 We hardcoded `SyncEngine.CreditCards`
You need to pass the syncEngine as a variable otherwise, we always changing CreditCards.

gabrielluong (approved):
Still need to address one change, but I don't think this needs another review.

app/src/main/java/org/mozilla/fenix/settings/account/AccountSettingsFragment.kt
```diff
@@ -218,7 +217,57 @@ class AccountSettingsFragment : PreferenceFragmentCompat() {
...
     }
 
-    private fun showPinDialogWarning(newValue: Boolean) {
+    /**
+     * Prompts the user if they do not have a password/pin set up to secure their device, and
+     * updates the state of the sync engine with the new checkbox value.
+     *
+     * Currently used for logins and credit cards.
+     *
+     * @param engine the sync engine whose preference has changed.
```
gabrielluong, line 226:
```suggestion
     * @param syncEngine the sync engine whose preference has changed.
```

Maybe we should rename s/engine/syncEngine in `updateSyncEngineState` as well

PR #19477, opened 2021-05-12
For #18287 - Display a credit card autofill prompt

### LINKED ISSUE #18287: [Select credit card] Display a credit card autofill prompt
This issue is to show the "Select credit card" prompt when a credit form is encountered that can be autofilled. This will depend on GV/AC work to wire everything up for this to be displayed. We will not worry about the functionality in the prompt in this issue and it is only focused on wiring things up to display the prompt.

This depends on https://github.com/mozilla-mobile/fenix/issues/18282.

Screen Shot 2021-03-02 at 6 07 24 PM

gabrielluong:
Fixes #18287 

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/Core.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/Core.kt
@@ -120,6 +120,7 @@ class Core(
             defaultSettings,
             GeckoProvider.getOrCreateRuntime(
                 context,
+                lazyAutofillStorage,
                 lazyPasswordsStorage,
                 trackingProtectionPolicyFactory.createTrackingProtectionPolicy()
             )
@@ -155,6 +156,7 @@ class Core(
             context,
             GeckoProvider.getOrCreateRuntime(
                 context,
+                lazyAutofillStorage,
                 lazyPasswordsStorage,
                 trackingProtectionPolicyFactory.createTrackingProtectionPolicy()
             )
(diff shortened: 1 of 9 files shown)


app/src/main/java/org/mozilla/fenix/components/Core.kt
```diff
@@ -120,6 +120,7 @@ class Core(
             defaultSettings,
             GeckoProvider.getOrCreateRuntime(
                 context,
+                lazyAutofillStorage,
```
jonalmeida, line 123: This is not the right way to adding a storage delegate. Unfortunately, we have logins that started doing this with the pattern following.

Maybe we need to file an issue to fix this up to change these APIs before it happens the third or fourth time.

gabrielluong: This isn't adding the storage delegate. We're adding the storage delegate here https://github.com/mozilla-mobile/fenix/blob/8adc69f3fcfb73f83e02c87d62f038cf82ac4aa6/app/src/main/java/org/mozilla/fenix/gecko/GeckoProvider.kt#L93. This is simply about passing the underlying storage so that it can be passed to their respective storage delegate when the GeckoRuntime is created and can be set via `autocompleteStorageDelegate`, which is its proper GV API.

jonalmeida: We're talking about the same thing at a different abstraction level. See the Engine.kt API which is how we should be adding these storage delegates.

I think it's a bit too late in this particular process to clean this up, so let's do it in a follow up.

Amejia481 (approved):

PR #19499, opened 2021-05-13
For #18275: Handle duplicates when adding or updating a credit card.

### LINKED ISSUE #18275: Handle duplicates when adding or updating a credit card
When an user is adding or updating a credit card, we should handle the case where the entered credit card to be saved turns out to be a duplicate. 

This issue is to handle any error messaging to display when the entered credit card to be added turns out to be a duplicate. 

Screen Shot 2021-03-02 at 4 14 41 PM

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-9894)

mcarare:
Fixes #18275

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardEditorFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardEditorFragment.kt
@@ -12,6 +12,8 @@ import android.view.View
 import androidx.lifecycle.lifecycleScope
 import androidx.navigation.fragment.findNavController
 import androidx.navigation.fragment.navArgs
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
 import org.mozilla.fenix.R
 import org.mozilla.fenix.SecureFragment
 import org.mozilla.fenix.ext.components
@@ -26,6 +28,7 @@ import org.mozilla.fenix.settings.creditcards.view.CreditCardEditorView
  */
 class CreditCardEditorFragment : SecureFragment(R.layout.fragment_credit_card_editor) {
 
+    private lateinit var creditCardValidator: CreditCardValidator
     private lateinit var creditCardEditorState: CreditCardEditorState
     private lateinit var creditCardEditorView: CreditCardEditorView
     private val args by navArgs<CreditCardEditorFragmentArgs>()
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardValidator.kt
```diff
@@ -0,0 +1,57 @@
...
+        return false
+    }
+
+    /**
+     * A class that describes a credit card validation result.
+     */
+    sealed class ValidationResult {
+        data class Invalid(@StringRes val value: Int) : ValidationResult()
+        data class Valid(val value: Boolean = true) : ValidationResult()
+    }
```
gabrielluong, line 56: Seeing this I am wondering if we should depend on https://github.com/mozilla-mobile/android-components/pull/9875 which will tell you whether or not a credit can be added or updated.

mcarare: Maybe I will halt this until that is finished and merged. At the moment I do not see any input validation in `CreditCardValidationDelegate`, just info about the possibility of updating or saving a credit card. Is there any flow where new credit cards are saved in other places than `CreditCardEditorFragment`? In a prompt or something?

mcarare: @gabrielluong I am thinking of keeping this PR for just handling the extraction of the validator so we can easily add other validation for date and name if UX/product decides to do this.
For the duplication part, to avoid the need to pass the credit card list to this validator, do you think the check can be handled by the validator you are working on in AC, or should the save/ update call to storage return a response that we can handle in Fenix? Can we enforce the no duplicates rule in the database or do we depend on a-s implementations?

gabrielluong: I think we should ignore the validator I am working on in AC. We should handle duplicates here as well this is actually a P1 priority from product. Looking into the Logins implementation of this, I think they do fetch the list of logins from the DB and have some particular AS functions that we don't have. I don't think we should rely on the list from the state. 

We may need to ask for some AS functions as needed.

I think my AC validator will only used for the prompts, but since I no longer see how my validator fits here, I actually halted work on it since we don't have the save/update prompts which would've utilized it.

app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardEditorFragment.kt
```diff
@@ -59,10 +62,20 @@ class CreditCardEditorFragment : SecureFragment(R.layout.fragment_credit_card_ed
...
+
+        creditCardEditorView = CreditCardEditorView(view, interactor, creditCardValidator)
         creditCardEditorView.bind(creditCardEditorState)
     }
 
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        val storage = requireContext().components.core.autofillStorage
+        lifecycleScope.launch(Dispatchers.IO) {
+            val list = storage.getAllCreditCards()
```
gabrielluong, line 74: I think we want to call this on demand when we do validate the credit card. Otherwise, it's possible that we want to the edge case where someone was just hovering on the CreditCardEditorFragment and meanwhile in sync, there's a new credit card that has been added, but we won't be able to identify that as a duplicate since we cache the list of credits. So, my point here is to avoid caching the lists of credit cards if possible since we only check if it's duplicate when the save button is clicked.

mcarare: While I agree there might be some edge cases, the current implementation helps to get a faster validation response that is useful in all cases, not just in an edge case.

gabrielluong:
@mcarare I am looking to cleanup stale issues on PR. Since we know we won't have anything actionable here for awhile. I am gonna opt to linking this PR to the original issue and closing the PR for the time being.

PR #19505, opened 2021-05-13
For #18264: add biometric prompt to credit card settings

### LINKED ISSUE #18264: [Credit cards] Add the biometric prompt when navigating to "Manage saved cards"
Most of the below is already implemented in https://github.com/gabrielluong/fenix/commit/2bca18187ddbe576b0626784ca572bb76d1f6a2d. I would like for someone smarter than me to see if we can refactor both the Credit card and Saved Logins code into a view similar to SyncPreferenceView working off the commit above. I would imagine we will need to pass in specific strings, but some of these strings are already common between the 2. In particular, I would want someone with deeper understanding of Android to see if we can remove the deprecation and fix the issues around having the Delay. Ideally, we should also add test coverage similar to SyncPreferenceView.

------
We want to add a biometric prompt when pressing the "Manage saved cards" preference to navigate to the "Saved cards" screen. This issue is about adding the biometric prompt feature and then navigating to the "Saved cards".

See
https://github.com/mozilla-mobile/fenix/blob/ccbca2cf78950182278308baabeba5ee0028eb59/app/src/main/java/org/mozilla/fenix/settings/logins/fragment/SavedLoginsAuthFragment.kt#L73
https://github.com/mozilla-mobile/fenix/blob/ccbca2cf78950182278308baabeba5ee0028eb59/app/src/main/java/org/mozilla/fenix/settings/logins/fragment/SavedLoginsAuthFragment.kt#L53.

There is a bug when the biometric prompt is showing, you are able to quickly navigate so we should disable the preferences in the "Credit card" screen while the biometric prompt is showing.

eliserichards:
Fixes #18264 

Followup to use prompt helper in logins auth fragment: https://github.com/mozilla-mobile/fenix/issues/19547

In this PR: 
- Mirror the authentication flow for saved logins
- Create `BiometricPromptPreferenceFragment.kt` to be used with CC, addresses (future), and logins (#19547)
- Management fragment is secure
- Management fragment pops you back to reauth when the fragment is paused
- CC Editor is secure
- Editor pops back to reauth when the fragment is paused

### secure on pause
Whenever the app is paused (backgrounded, etc) after we authenticate, we shouldn't be able to see what's on the screen. This is the same functionality as logins.
image

### authenticate to see cards
When navigating to the list of credit cards from settings, we prompt the user to authenticate (or prompt them to make a device pin/password if they don't have one)
unlock_cc

### reauth on pause - list of cards
If we have authenticated and see the list of credit cards, if the user pauses the app we pop them back to the CC settings so they have to reauth to see the list again.
reauth_cc

### reauth on pause - editing card
If we are in the credit card editor, if the user pauses the app we pop them back to the CC settings so they have to reauth to see the cards again.
reauth_edit_cc

### Pull Request checklist

1. click on Show All Checks,


app/src/main/res/values/strings.xml
```diff
@@ -1555,7 +1555,18 @@
...
+    <string name="credit_cards_warning_dialog_title">Secure your credit cards</string>
+    <!-- Message of warning dialog if users have no device authentication set up -->
+    <string name="credit_cards_warning_dialog_message">Set up a device lock pattern, PIN, or password to protect your saved credit cards from being accessed if someone else has your device.</string>
+    <!-- Positive button to send users to set up a pin of warning dialog if users have no device authentication set up -->
+    <string name="credit_cards_warning_dialog_set_up_now">Set up now</string>
+    <!-- Negative button to ignore warning dialog if users have no device authentication set up -->
+    <string name="credit_cards_warning_dialog_later">Later</string>
+    <!-- Title of PIN verification dialog to direct users to re-enter their device credentials to access their credit cards -->
+    <string name="credit_cards_biometric_prompt_message_pin">Unlock your device</string>
     <!-- Title of the Add search engine screen -->
```
gabrielluong, line 1570: Can you add a new line before since this has nothing to do with `Credit Cards Autofill`

gabrielluong (changes requested):

app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardsManagementFragment.kt
```diff
@@ -72,9 +79,33 @@ class CreditCardsManagementFragment : SecureFragment() {
 
     override fun onResume() {
         super.onResume()
+        activity?.window?.setFlags(
+            WindowManager.LayoutParams.FLAG_SECURE,
+            WindowManager.LayoutParams.FLAG_SECURE
+        )
```
gabrielluong, line 85: Don't think we need this either since it's a `SecureFragment`

app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardsManagementFragment.kt
```diff
@@ -72,9 +79,33 @@ class CreditCardsManagementFragment : SecureFragment() {
...
         showToolbar(getString(R.string.credit_cards_saved_cards))
     }
 
+    /**
+     * If we pause this fragment, we want to pop users back to the settings page to reauthenticate.
+     */
+    override fun onPause() {
```
gabrielluong, line 92: Maybe this should also go into `SecureFragment`

app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardsSettingFragment.kt
```diff
@@ -151,4 +167,88 @@ class CreditCardsSettingFragment : PreferenceFragmentCompat() {
...
+                // https://github.com/mozilla-mobile/fenix/issues/8438
+                delay(SHORT_DELAY_MS)
+                navigateToCreditCardManagementFragment()
+            }
+        }
+    }
+
+    /**
+     * Called when authentication succeeds.
+     */
```
gabrielluong, line 225: I would probably just remove this comment since the others don't have one either, and it's also not fully correct since it's also called on the negative button when we show the pin dialog warning.

app/src/main/java/org/mozilla/fenix/settings/biometric/BiometricPromptPreferenceFragment.kt
```diff
@@ -0,0 +1,123 @@
...
+import androidx.preference.PreferenceFragmentCompat
+import mozilla.components.support.base.feature.ViewBoundFeatureWrapper
+import org.mozilla.fenix.ext.settings
+import org.mozilla.fenix.settings.requirePreference
+
+/**
+ * Helper for creating and implementing the [BiometricPromptFeature]. Currently used
+ * for logins and credit cards.
+ */
+abstract class BiometricPromptPreferenceFragment : PreferenceFragmentCompat() {
```
gabrielluong, line 23: Happy to see this, also really happy to see we're able to reduce code duplications. I just wished we could've also reduced showPinDialogWarning and showPinVerification to just this file as well.

eliserichards:
I agree that those things should be moved into the secure fragment. I think that should be a separate PR, and this is good as-is.

app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardsSettingFragment.kt
eliserichards: good catch. this was copy/paste from logins

app/src/main/java/org/mozilla/fenix/settings/biometric/BiometricPromptPreferenceFragment.kt
eliserichards: I think this can come in a followup as well

app/src/main/java/org/mozilla/fenix/settings/biometric/BiometricPromptPreferenceFragment.kt
```diff
@@ -0,0 +1,123 @@
...
+
+    /**
+     * There is a bug where while the biometric prompt is showing, you were able to quickly navigate
+     * so we are disabling the settings that navigate while authenticating.
+     * https://github.com/mozilla-mobile/fenix/issues/12312
+     *
+     * @param prefList a list of [Preference]s to toggle.
+     * @param enabled whether or not the preferences should be enabled.
+     */
+    fun togglePrefsEnabledWhileAuthenticating(prefList: List<Int>, enabled: Boolean) {
```
gabrielluong, line 52: Probably unnecessary to put preconditions to a function that simply toggles isEnabled for preferences.

```suggestion
    fun togglePrefsEnabled(prefList: List<Int>, enabled: Boolean) {
```

app/src/main/java/org/mozilla/fenix/settings/biometric/BiometricPromptPreferenceFragment.kt
```diff
@@ -0,0 +1,123 @@
...
+    abstract fun navigateOnSuccess()
+
+    /**
+     * Show a warning to set up a pin/password when the device is not secured. This is only used
+     * when BiometricPrompt is unavailable on the device.
+     */
+    abstract fun showPinDialogWarning(context: Context)
+
+    /**
+     * There is a bug where while the biometric prompt is showing, you were able to quickly navigate
```
gabrielluong, line 45: I think this comment should be at the caller which I think is at `onPause` cause it wouldn't make sense reading this for `onAuthFailure`.

I would add a proper description here just describing the functionality of this function.

app/src/main/java/org/mozilla/fenix/settings/biometric/BiometricPromptPreferenceFragment.kt
```diff
@@ -0,0 +1,123 @@
...
+     * @param enabled whether or not the preferences should be enabled.
+     */
+    fun togglePrefsEnabledWhileAuthenticating(prefList: List<Int>, enabled: Boolean) {
+        for (preference in prefList) {
+            requirePreference<Preference>(preference).isEnabled = enabled
+        }
+    }
+
+    /**
+     * Create a prompt to confirm the device's pin/password and start activity based on the result.
```
gabrielluong, line 59:
```suggestion
     * Creates a prompt to confirm the device's pin/password and starts the activity based on the result.
```

app/src/main/java/org/mozilla/fenix/settings/biometric/BiometricPromptPreferenceFragment.kt
```diff
@@ -0,0 +1,123 @@
...
+    @Suppress("Deprecation")
+    abstract fun showPinVerification(manager: KeyguardManager)
+
+    /**
+     * Sets the biometric prompt feature.
+     *
+     * @param view the view that the prompt will be associate with.
+     * @param prefList a list of [Preference]s to toggle.
+     */
+    fun setBiometricPrompt(view: View, prefList: List<Int>) {
```
gabrielluong, line 71: Did we consider overriding `onViewCreated` to avoid having inherited fragment call this manually?

app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardEditorFragment.kt
```diff
@@ -63,8 +65,23 @@ class CreditCardEditorFragment : SecureFragment(R.layout.fragment_credit_card_ed
         creditCardEditorView.bind(creditCardEditorState)
     }
 
+    /**
+     * Close any open dialogs or menus and reauthenticate if the fragment is paused and
+     * the user is not navigating to [CreditCardsManagementFragment].
+     */
+    override fun onPause() {
+        menu.close()
```
gabrielluong, line 73: Wondering about the reasoning behind this? Since there's no context menu or open dialog here, I don't think we are closing anything here.

app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardEditorFragment.kt
```diff
@@ -63,8 +65,23 @@ class CreditCardEditorFragment : SecureFragment(R.layout.fragment_credit_card_ed
...
+
+        redirectToCreditCardReAuth(
```
gabrielluong, line 75: I don't quite understand this. I don't think we need to reauth if we were adding a new credit card. Perhaps if we were editing, but I am really not sure about this experience. I would follow up on this with some product or UX feedback.

app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardsSettingFragment.kt
```diff
@@ -85,32 +113,20 @@ class CreditCardsSettingFragment : PreferenceFragmentCompat() {
...
-                    CreditCardsSettingFragmentDirections.actionCreditCardsSettingFragmentToTurnOnSyncFragment()
-                findNavController().navigateBlockingForAsyncNavGraph(directions)
-            },
-            onReconnectClicked = {
-                val directions =
-                    CreditCardsSettingFragmentDirections.actionGlobalAccountProblemFragment()
-                findNavController().navigateBlockingForAsyncNavGraph(directions)
-            }
+            onSignInToSyncClicked = { navigateToSyncFragment() },
+            onReconnectClicked = { navigateToAccountProblemFragment() }
```
gabrielluong, line 117: I don't think this change is really necessary since the functions are not used more than once. It would be preferable to avoid unrelated changes in such a big PR and file followups for these sort of things. We could also potentially avoid `TooManyFunctions ` by undoing some of this.

app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardsSettingFragment.kt
```diff
@@ -85,32 +113,20 @@ class CreditCardsSettingFragment : PreferenceFragmentCompat() {
...
     }
 
     override fun onPreferenceTreeClick(preference: Preference): Boolean {
         when (preference.key) {
             getPreferenceKey(R.string.pref_key_credit_cards_add_credit_card) -> {
-                val directions =
-                    CreditCardsSettingFragmentDirections
-                        .actionCreditCardsSettingFragmentToCreditCardEditorFragment()
-                findNavController().navigateBlockingForAsyncNavGraph(directions)
+                navigateToCreditCardEditorFragment()
```
gabrielluong, line 126: Same as above. I think inlining is fine unless the navigateX function is reusable and needs to be in a callback.

app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardsSettingFragment.kt
```diff
@@ -151,4 +167,85 @@ class CreditCardsSettingFragment : PreferenceFragmentCompat() {
...
+            }
+
+            setPositiveButton(getString(R.string.credit_cards_warning_dialog_set_up_now)) { it: DialogInterface, _ ->
+                it.dismiss()
+                val intent = Intent(Settings.ACTION_SECURITY_SETTINGS)
+                startActivity(intent)
+            }
+
+            create()
+        }.show().secure(activity)
```
gabrielluong, line 191: I was also expecting some related changes too `incrementShowLoginsSecureWarningCount`. If `showPinDialogWarning` was entirely in `BiometricPromptPreferenceFragment`, I would say we would reuse the same counter, but I still think we should repurpose the logins warning count to be generic that can be shared between the 2 since I don't think we need to have a different counter for doing the same thing.

gabrielluong (changes requested):

app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardsSettingFragment.kt
```diff
@@ -85,32 +113,20 @@ class CreditCardsSettingFragment : PreferenceFragmentCompat() {
...
-                        .actionCreditCardsSettingFragmentToCreditCardEditorFragment()
-                findNavController().navigateBlockingForAsyncNavGraph(directions)
+                navigateToCreditCardEditorFragment()
             }
             getPreferenceKey(R.string.pref_key_credit_cards_manage_saved_cards) -> {
-                val directions =
-                    CreditCardsSettingFragmentDirections
-                        .actionCreditCardsSettingFragmentToCreditCardsManagementFragment()
-                findNavController().navigateBlockingForAsyncNavGraph(directions)
+                verifyCredentialsOrShowSetupWarning(preference.context, creditCardPreferences)
```
grigoryk, line 129: Just a little drive-by comment, but I figured I'll chime in...

There's no reason to require authentication to display a list of cards, or even to open a specific card. Sensitive information (card number) should not be displayed right away. We can't actually display it - it's encrypted! So, that's what I imagined would be guarded by an auth prompt - the action to display a specific credit card number.

To put it in other words - there's a cryptographic key that we need to fetch in order to decrypt or encrypt sensitive information related to a credit card number. The act of obtaining a key, while not technically necessary, should be protected by some authentication mechanism (e.g. a biometrics prompt).

What I think is happening in this PR is more akin to how we're treating logins currently. That's not wrong, of course. However, even for logins we'll be soon switching to a per-field authentication model (vs per-db, as it is currently), which would mean that only username/password fields will be encrypted, while the rest will be in plaintext. That'll allow us to follow the same pattern for both credit cards and logins at the product UX level (e.g. require auth only when accessing sensitive information).

app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardsSettingFragment.kt
eliserichards: What we're doing with that count is unclear to me. If there is a use for it then that should be a separate issue.

app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardsSettingFragment.kt
eliserichards: Thanks for the context! We should probably check in with UX on this because AFAIK this is the user authentication flow that they want (for now)

app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardsSettingFragment.kt
eliserichards: I disagree, I think it makes this much easier to read

app/src/main/java/org/mozilla/fenix/settings/biometric/BiometricPromptFeature.kt
```diff
@@ -28,7 +28,7 @@ import org.mozilla.fenix.settings.biometric.ext.isHardwareAvailable
  * @param onAuthFailure A failure callback if authentication failed.
  */
 class BiometricPromptFeature(
-    private val context: Context,
+    private val context: Context?,
```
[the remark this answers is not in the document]
jonalmeida: Good catch. The consumer (BiometricPromptPreferenceFragment) should use `requireContext()` here.

app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardsSettingFragment.kt
gabrielluong: Incrementing the warning count allows us to determine whether or not we should `shouldShowSecurityPinWarning` or if we have shown the pin warning a number of times and the user decided to ignore it, we just navigate directly. Look for `shouldShowSecurityPinWarning`.

I am a bit cautious about doing this in a separate issue since we have the time now to just address the feedback. There is also no guarantees that we get to working on followups since it's not in the sprint.

app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardsSettingFragment.kt
gabrielluong: I think that thinking is good, but I don't think that is really the issue here. It should be easy enough to read the navigation code as-is since it's only 3 lines and could be 2 if we inlined the `directions`. So, I would be a bit considerate on how we write our commits with consideration on helping streamline the review process.

app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardsSettingFragment.kt
```diff
@@ -4,39 +4,65 @@
...
 
 /**
  * "Credit cards" settings fragment displays a list of settings related to autofilling, adding and
- * syncing credit cards.
+ * syncing credit cards. Authentication for saved credit cards using [BiometricPromptFeature]
+ * or [KeyguardManager] (similar to [SavedLoginsAuthFragment]).
  */
-class CreditCardsSettingFragment : PreferenceFragmentCompat() {
+@SuppressWarnings("TooManyFunctions")
+class CreditCardsSettingFragment : BiometricPromptPreferenceFragment() {
```
mcarare, line 48: Question: Do we really need to secure this fragment? Is there sensitive data displayed here? Fo screenshots we haven't blocked this screen from being screenshotted.

app/src/main/java/org/mozilla/fenix/ext/Fragment.kt
```diff
@@ -72,6 +72,19 @@ fun Fragment.redirectToReAuth(destinations: List<Int>, currentDestination: Int?)
...
 }
 
+/**
+ * Pops the backstack to force users to re-auth if they put the app in the background and return to
+ * it while being inside the Credit Card flow
+ *
+ * Does nothing if the user is currently navigating to any of the [destinations] given as a
+ * parameter.
+ */
+fun Fragment.redirectToCreditCardReAuth(destinations: List<Int>, currentDestination: Int?) {
```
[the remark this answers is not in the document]
eliserichards: That's fair. I wanted to keep them separate since we might have different needs for CC and logins, but I can make this change.

app/src/main/java/org/mozilla/fenix/settings/biometric/BiometricPromptPreferenceFragment.kt
eliserichards: This is a copied comment from logins. Should I update both comments or just this one?

app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardEditorFragment.kt
eliserichards: This is the same experience as logins. It wasn't clear to me that we were securing this differently than logins

app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardEditorFragment.kt
eliserichards: There is an options menu and it has the delete button in it. This menu.close() prevents a visual bug that we found in logins.

app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardsSettingFragment.kt
eliserichards: I will add this. Thanks :)

app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardsSettingFragment.kt
eliserichards: This is following the same flow that we have for logins (`SavedLoginsAuthFragment`) where we ask the user to authenticate before they can see their saved/synced info. We need to talk to UX about whether this flow should be the same or changed (see Grisha's comment above)

eliserichards:
Confirmed with @topotropic that we should be securing CCs the same way we secure logins, pending discussion with Grisha and GL.

app/src/main/java/org/mozilla/fenix/settings/biometric/BiometricPromptPreferenceFragment.kt
gabrielluong: Please provide a response before resolving a conversation if no action was taken. I wouldn't be leaving a comment if there was no merit to it. Best case scenario would be to think about it and decide if changed were warranted or provide some thoughtful feedback to comment. If it's decided we wanted to do this as a followup, leaving the comment unresolve and linking to it back in the followup ticket would also been a better option.

app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardsManagementFragment.kt
gabrielluong: Please provide a response before resolving a conversation if no action was taken. I wouldn't be leaving a comment if there was no merit to it. Best case scenario would be to think about it and decide if changed were warranted or provide some thoughtful feedback to comment. If it's decided we wanted to do this as a followup, leaving the comment unresolve and linking to it back in the followup ticket would also been a better option.

app/src/main/java/org/mozilla/fenix/settings/biometric/BiometricPromptPreferenceFragment.kt
```diff
@@ -0,0 +1,121 @@
...
+            view = view
+        )
+    }
+
+    /**
+     * Use [BiometricPromptFeature] or [KeyguardManager] to confirm device security.
+     *
+     * @param prefList a list of [Preference]s to disable while authentication is happening.
+     */
+    fun verifyCredentialsOrShowSetupWarning(context: Context, prefList: List<Int>) {
```
gabrielluong, line 89: I was testing this and wondering why things weren't working. We never call this when navigating to the saved cards. So, we don't actually show the biometric check. I am aware of what @grigoryk said, but I think for time being we should keep to the UX spec and re-evaluate.

gabrielluong (changes requested):

app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardsSettingFragment.kt
```diff
@@ -158,4 +196,54 @@ class CreditCardsSettingFragment : PreferenceFragmentCompat() {
 
         isCreditCardsListLoaded = true
     }
+
+    /**
+     * Show a warning to set up a pin/password when the device is not secured. This is only used
+     * when BiometricPrompt is unavailable on the device.
+     */
+    override fun showPinDialogWarning(context: Context) {
```
gabrielluong, line 204: Going back to what we said in https://github.com/mozilla-mobile/fenix/pull/19548#discussion_r640037358, we probably don't need a kDoc for these override fun as well

eliserichards: We are specifying functionality that is specific to the credit card setting fragment. It is overriding an abstract function and should be documented.

app/src/main/java/org/mozilla/fenix/settings/biometric/BiometricPromptPreferenceFragment.kt
eliserichards: This was missed in one of the million and a half times we've gone back and forth on this PR. Apologies for that.

app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardEditorFragment.kt
eliserichards: Update: I have confirmed that we are supposed to be doing this the same as logins.

app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardsSettingFragment.kt
gabrielluong: Cool, that's fine.

app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardsManagementFragment.kt
eliserichards: Redirecting could go in the SecureFragment, but since there are also specific things related to logins and credit cards I think this should stay in the respective fragments. Re-writing the secure fragment is a different piece of work than spinning up auth for credit cards.

gabrielluong:
Seeing this in the failure

```
java.lang.IllegalStateException: Fragment CreditCardsSettingFragment{3765b2a8} (1aeee736-e127-4a19-a7a2-770054a54a3b) CreditCardsSettingFragmentTest} does not have a NavController set
	at androidx.navigation.fragment.NavHostFragment.findNavController(NavHostFragment.java:131)
	at androidx.navigation.fragment.FragmentKt.findNavController(Fragment.kt:29)
	at org.mozilla.fenix.settings.creditcards.CreditCardsSettingFragment$updateCardManagementPreference$1.onPreferenceClick(CreditCardsSettingFragment.kt:171)
	at androidx.preference.Preference.performClick(Preference.java:1184)
	at org.mozilla.fenix.settings.creditcards.CreditCardsSettingFragmentTest.GIVEN the list of credit cards is empty, WHEN fragment is displayed THEN the manage credit cards pref is 'Add card'(CreditCardsSettingFragmentTest.kt:99)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.robolectric.RobolectricTestRunner$HelperTestRunner$1.evaluate(RobolectricTestRunner.java:546)
	at org.robolectric.internal.SandboxTestRunner$2.lambda$evaluate$0(SandboxTestRunner.java:252)
	at org.robolectric.internal.bytecode.Sandbox.lambda$runOnMainThread$0(Sandbox.java:89)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
```

gabrielluong:
```
java.lang.IllegalStateException: Can't access the Fragment View's LifecycleOwner when getView() is null i.e., before onCreateView() or after onDestroyView()
	at androidx.fragment.app.Fragment.getViewLifecycleOwner(Fragment.java:328)
	at org.mozilla.fenix.settings.creditcards.CreditCardsSettingFragment.navigateOnSuccess(CreditCardsSettingFragment.kt:68)
	at org.mozilla.fenix.settings.biometric.BiometricPromptPreferenceFragment.verifyCredentialsOrShowSetupWarning(BiometricPromptPreferenceFragment.kt:106)
	at org.mozilla.fenix.settings.creditcards.CreditCardsSettingFragment$updateCardManagementPreference$1.onPreferenceClick(CreditCardsSettingFragment.kt:170)
	at androidx.preference.Preference.performClick(Preference.java:1184)
	at org.mozilla.fenix.settings.creditcards.CreditCardsSettingFragmentTest.GIVEN the list of credit cards is not empty, WHEN fragment is displayed THEN the manage credit cards pref is 'Manage saved cards'(CreditCardsSettingFragmentTest.kt:73)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.robolectric.RobolectricTestRunner$HelperTestRunner$1.evaluate(RobolectricTestRunner.java:546)
	at org.robolectric.internal.SandboxTestRunner$2.lambda$evaluate$0(SandboxTestRunner.java:252)
	at org.robolectric.internal.bytecode.Sandbox.lambda$runOnMainThread$0(Sandbox.java:89)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
```

gabrielluong (approved):

Issue #19684, opened 2021-05-27
Display the card network name besides the card number


gabrielluong:
We should display the credit card network name (Visa, Discover, etc) besides the card number. You will find the list of supported credit card network https://searchfox.org/mozilla-mobile/source/android-components/components/support/utils/src/main/java/mozilla/components/support/utils/CreditCardUtils.kt#40. I think these should be mapped to localized, you will want to map the cardType to their formal credit card network name. We will probably want to ask if (1) the credit card network name should be localized from @Delphine, and (2) make sure we get the correct network names, I looked at desktop and saw MasterCard which I assume is probably incorrect for instance.

<img width="375" alt="Screen Shot 2021-05-27 at 3 54 50 AM" src="https://user-images.githubusercontent.com/1190888/119787613-60eaa500-be9f-11eb-96a0-072b06ad61cb.png">

gabrielluong:
We should probably display nothing if the cardType does not match the list of enum

codrut-topliceanu:
>We will probably want to ask if (1) the credit card network name should be localized from @Delphine, and (2) make sure we get the correct network names, I looked at desktop and saw MasterCard which I assume is probably incorrect for instance.

@Delphine could you provide any input here? Are localized credit card network names a thing?
As far as I know the Network names are an international standard but I'm not really well informed on this.

<img src="https://user-images.githubusercontent.com/60002907/119820770-81941880-befa-11eb-8bc7-e1a762ffc159.png" width="300"/>

Delphine:
Thanks for reaching out! I checked against desktop strings, and these should in face be localizable.
Also, example with MasterCard hebrew website: https://www.mastercard.co.il/he-il.html

gabrielluong:
@codrut-topliceanu I decided to push this to 91 instead. I think we decided to just display the "name on card" instead.

gabrielluong:
Closing this, let's file a new issue for adding "name on card"

PR #19687, opened 2021-05-27
For #19684 - Adds CC Network Name besides number

codrut-topliceanu:
For https://github.com/mozilla-mobile/fenix/issues/19684
### Pull Request checklist

1. click on Show All Checks,

Putting this up as draft while we wait for additional details regarding localization.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/settings/creditcards/view/CreditCardItemViewHolder.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/creditcards/view/CreditCardItemViewHolder.kt
@@ -7,8 +7,18 @@ package org.mozilla.fenix.settings.creditcards.view
 import android.view.View
 import kotlinx.android.synthetic.main.credit_card_list_item.*
 import mozilla.components.concept.storage.CreditCard
+import mozilla.components.support.utils.CreditCardNetworkType.AMEX
+import mozilla.components.support.utils.CreditCardNetworkType.CARTEBANCAIRE
+import mozilla.components.support.utils.CreditCardNetworkType.DINERS
+import mozilla.components.support.utils.CreditCardNetworkType.DISCOVER
+import mozilla.components.support.utils.CreditCardNetworkType.JCB
+import mozilla.components.support.utils.CreditCardNetworkType.MASTERCARD
+import mozilla.components.support.utils.CreditCardNetworkType.MIR
+import mozilla.components.support.utils.CreditCardNetworkType.UNIONPAY
+import mozilla.components.support.utils.CreditCardNetworkType.VISA
 import mozilla.components.support.utils.creditCardIssuerNetwork
 import org.mozilla.fenix.R
+import org.mozilla.fenix.R.string
 import org.mozilla.fenix.settings.creditcards.interactor.CreditCardsManagementInteractor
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/mozilla/fenix/settings/creditcards/view/CreditCardItemViewHolder.kt
```diff
@@ -26,6 +26,9 @@ class CreditCardItemViewHolder(
     fun bind(creditCard: CreditCard) {
         credit_card_logo.setImageResource(creditCard.cardType.creditCardIssuerNetwork().icon)
 
+        credit_card_network.text = creditCard.cardType.creditCardIssuerNetwork().name
+            .capitalize(Locale.getDefault())
```
gabrielluong, line 30: This won't work for "amex", which should either be "AMEX" or "American Express".

app/src/main/java/org/mozilla/fenix/settings/creditcards/view/CreditCardItemViewHolder.kt
```diff
@@ -26,6 +26,9 @@ class CreditCardItemViewHolder(
     fun bind(creditCard: CreditCard) {
         credit_card_logo.setImageResource(creditCard.cardType.creditCardIssuerNetwork().icon)
 
+        credit_card_network.text = creditCard.cardType.creditCardIssuerNetwork().name
```
gabrielluong, line 29: Ideally, `name` would be the correct display name, but it really isn't in this case. We can do this as a followup. I will try to get a list of display names for the network names in the meantime, and we will have to add a mapping similar to https://searchfox.org/mozilla-mobile/source/android-components/components/support/utils/src/main/java/mozilla/components/support/utils/CreditCardUtils.kt#103 in Fenix to land this tomorrow.

gabrielluong: I never gotta a formal list of names from UX or Product. I think we should do a best effort and note that it can be localized https://github.com/mozilla-mobile/fenix/issues/19684#issuecomment-849862981.

gabrielluong:
@codrut-topliceanu I checked with @topotropic and we want to add the name on card instead of network names to align with desktop. Sorry!

codrut-topliceanu:
> @codrut-topliceanu I checked with @topotropic and we want to add the name on card instead of network names to align with desktop. Sorry!

I just realized you already said that in the issue and you were referring to the card holder name, not the `card.name`.
Should I close the PR and issue? Or change the code here, while keeping the issue? What is the recommended procedure here?

gabrielluong:
> > > @codrut-topliceanu I checked with @topotropic and we want to add the name on card instead of network names to align with desktop. Sorry!
> > 
> > 
> > I just realized you already said that in the issue and you were referring to the card holder name, not the `card.name`.
> > Should I close the PR and issue? Or change the code here, while keeping the issue? What is the recommended procedure here?
> 
> I think creating a new PR probably makes sense to preserve this work in case we ever want to come back to it. Having said that, I think it would be best to just file a new issue as well to preserve the history of your changes and the intentions behind. If we decided to come back to the network names, then we can just look up this PR in the future.

Closing this PR as discussed.

PR #19691, opened 2021-05-27
Update gpg pub key

bhearsum:
We just swapped in a new signing key in https://bugzilla.mozilla.org/show_bug.cgi?id=1703397. This updates the checked in pub key to include both the old and new ones. Should fix tasks like https://firefox-ci-tc.services.mozilla.com/tasks/Hi8ZlAnsRPCroxRHgBw7Lg.


bhearsum:
(i don't think that test failure is the fault of my changes)

MihaiTabara (approved):

PR #19698, opened 2021-05-28
Minor account settings cleanup & enable addresses sync for nightly/debug channels

grigoryk:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/BackgroundServices.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/BackgroundServices.kt
@@ -36,6 +36,7 @@ import mozilla.components.service.sync.autofill.AutofillCreditCardsAddressesStor
 import mozilla.components.service.sync.logins.SyncableLoginsStorage
 import mozilla.components.support.utils.RunWhenReadyQueue
 import org.mozilla.fenix.Config
+import org.mozilla.fenix.FeatureFlags
 import org.mozilla.fenix.R
 import org.mozilla.fenix.perf.StrictModeManager
 import org.mozilla.fenix.components.metrics.Event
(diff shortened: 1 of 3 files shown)


grigoryk:
To fully enable addresses sync (so that it's reflected in the UI) we need https://github.com/mozilla-mobile/android-components/pull/10370, but don't need to block on that landing.

app/src/main/res/xml/account_settings_preferences.xml
```diff
@@ -49,14 +48,13 @@
...
         <CheckBoxPreference
-            android:defaultValue="false"
+            android:defaultValue="true"
             android:key="@string/pref_key_sync_tabs"
             android:layout="@layout/checkbox_left_preference"
             android:title="@string/preferences_sync_tabs_2"/>
 
         <CheckBoxPreference
-            android:defaultValue="false"
-            android:visible="false"
```
gabrielluong, line 59: I think we should avoid removing the default visibility hidden here. Otherwise, we would always see the preference showing and being hidden in the release build.

gabrielluong (approved):

app/src/main/res/xml/account_settings_preferences.xml
grigoryk: Ah, good call.

app/src/main/java/org/mozilla/fenix/components/BackgroundServices.kt
```diff
@@ -116,6 +118,9 @@ class BackgroundServices(
             storePair = SyncEngine.CreditCards to creditCardsStorage,
             keyProvider = lazy { creditCardKeyProvider }
         )
+        if (FeatureFlags.addressesFeature) {
```
gabrielluong, line 121: Should we use `context.settings().addressesFeature` since that defaults to `false`?

grigoryk: I wanted to be consistent w/ CC, and there's an equivalent settings flag for it (just the FeatureFlags one). It's all a bit confusing, but the `FeatureFlags.addressesFeature` is at least easy to understand/follow, so I'd just stick with that for now. This gets us `true` for nightly+debug, and `false` elsewhere, which is what we want.

PR #19726, opened 2021-05-31
Revert "Minor account settings cleanup & enable addresses sync for nightly/debug channels"

csadilek:
Let's revert this to help with the A-S team with investigation of https://github.com/mozilla/application-services/issues/4035.

We want to confirm if the spike is triggered by problems with the new addresses sync.


rfk:
Update: we think we found the underlying bug, ref https://github.com/mozilla/application-services/pull/4154; it's potentially an issue for both credit cards and addresses, meaning that disabling just addresses here is not likely to help. I think we should close this PR, and instead the appservices team will prepare a `v77.0.2` point-release with the fix from https://github.com/mozilla/application-services/pull/4154 included.

rfk:
> instead the appservices team will prepare a v77.0.2 point-release with the fix from mozilla/application-services#4154 included.

Ref https://github.com/mozilla-mobile/android-components/pull/10383 for this update.

Issue #19789, opened 2021-06-03
Start on Home


Amejia481:
After being away from the browser for X hours, I like to open the app on HOME so I can start fresh and not be distracted by the last open tab which may no longer be relevant. Related [ticket](https://jira.mozilla.com/browse/FNX-22155).

**Acceptance Criteria:**

- I open the app on HOME after X hours of inactivity (default X tbd later). Inactivity counter starts when app goes to background and we can go with 4 hours as X for iteration 1)
- Bringing custom tab to the foreground does not count as activity
- Going to the background counts as an inactivity
- There is a user-facing setting to change the value of X (predefined values not custom) => this is a P2 but important to call out easily so it is considered in development strategy.

cadeyrn:
@Amejia481 Thanks for working on it!

> There is a user-facing setting to change the value of X (predefined values not custom) => this is a P2 but important to call out easily so it is considered in development strategy.

Could you also consider a value that in fact means: always open on home screen? Well, maybe it still makes sense to open the last open tab if you close Firefox and open it again after a few seconds but a value smaller than one day (like one minute or at least one hour) would be awesome. I almost never want to start with the last open tab.

Amejia481:
Thanks for the feedback @cadeyrn, that sounds `always` like an awesome idea, I will circle back with product to see what they think, maybe we could have `always`, `never` and an spesific set of time.

Amejia481:
QA team please help us to verify this feature is working as expected.

We want to cover the following cases:

1. Users should should only start on home on cold starts when selecting the options **after 4 hours** or **always**, and they have opened tabs.
2. When selecting never on cold starts users should go directly should never see the home screen when they have opened tabs.
3. When a page is open via an external app link, users should always to the browser screen independently of their settings.
4. When users don't have opened tabs should always start on home independently of their settings. 


To test case 1, you can alter your OS data & time, to go further in time, but try to do it on an emulator not on your personal phone, as this could cause other apps to not work.

To access the start on home feature, you have to go to `Settings` -> `Tabs` -> `Start on home`.

![image](https://user-images.githubusercontent.com/773158/121083896-fb0af100-c7ad-11eb-81b8-2d2bb22f6223.png)

Amejia481:
Thanks @lobontiumira!

PR #19791, opened 2021-06-03
For #19789 Start on Home after some amount of time

Amejia481:
image

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
+++ b/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
@@ -6,6 +6,7 @@ package org.mozilla.fenix
 
 import android.content.Context
 import android.content.Intent
+import android.content.Intent.ACTION_MAIN
 import android.content.res.Configuration
 import android.os.Build
 import android.os.Bundle
@@ -116,6 +117,7 @@ import org.mozilla.fenix.tabstray.TabsTrayFragmentDirections
 import org.mozilla.fenix.theme.DefaultThemeManager
 import org.mozilla.fenix.theme.ThemeManager
 import org.mozilla.fenix.utils.BrowsersCache
+import org.mozilla.fenix.utils.Settings
 import java.lang.ref.WeakReference
 
 /**
@@ -217,8 +219,9 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
             it.start()
         }
 
-        if (isActivityColdStarted(intent, savedInstanceState) &&
-                !externalSourceIntentProcessors.any { it.process(intent, navHost.navController, this.intent) }) {
+        if (!shouldStartOnHome() &&
+            shouldNavigateBrowserFragmentOnCouldStart(savedInstanceState)
+        ) {
             navigateToBrowserOnColdStart()
         }
 
@@ -978,6 +981,32 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
                 // Activity was restarted from Recents after it was destroyed by Android while in background
                 // in cases of memory pressure / "Don't keep activities".
                 startingIntent.flags and Intent.FLAG_ACTIVITY_LAUNCHED_FROM_HISTORY == 0
+    }
+
+    /**
+     *  Indicates if the user should be redirected to the [BrowserFragment] or to the [HomeFragment],
+     *  links from an external apps should always opened in the [BrowserFragment].
+     */
+    fun shouldStartOnHome(intent: Intent? = this.intent): Boolean {
+        return components.strictMode.resetAfter(StrictMode.allowThreadDiskReads()) {
+            // We only want to open on home when users tap the app,
+            // we want to ignore other cases when the app gets open by users clicking on links.
+            getSettings().shouldStartOnHome() && intent?.action == ACTION_MAIN
+        }
+    }
+
+    @VisibleForTesting
+    internal fun getSettings(): Settings = settings()
+
+    private fun shouldNavigateBrowserFragmentOnCouldStart(savedInstanceState: Bundle?): Boolean {
+        return isActivityColdStarted(intent, savedInstanceState) &&
+            !externalSourceIntentProcessors.any {
+                it.process(
+                    intent,
+                    navHost.navController,
+                    this.intent
+                )
+            }
     }
 
     companion object {
--- a/app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
@@ -8,6 +8,7 @@ import android.content.Context
 import android.os.StrictMode
 import android.view.View
 import android.view.ViewGroup
+import androidx.annotation.VisibleForTesting
 import androidx.appcompat.content.res.AppCompatResources
 import androidx.lifecycle.Observer
 import androidx.navigation.fragment.findNavController
@@ -183,7 +184,7 @@ class BrowserFragment : BaseBrowserFragment(), UserInteractionHandler {
 
     override fun onStop() {
         super.onStop()
-
+        updateLastBrowseActivity()
         pwaOnboardingObserver?.stop()
     }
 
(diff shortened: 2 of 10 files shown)


app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -120,6 +120,11 @@ class Settings(private val appContext: Context) : PreferencesHolder {
         default = 0L
     )
 
+    var lastBrowserActivity by longPreference(
```
gabrielluong, line 123: Can you add some kDoc for this? I think in general we want to add kDocs for any new Settings since there are so many of them and we want to preserve some knowledge.

Amejia481: Sure 👍🏽

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -219,8 +219,9 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
             it.start()
         }
 
-        if (isActivityColdStarted(intent, savedInstanceState) &&
-                !externalSourceIntentProcessors.any { it.process(intent, navHost.navController, this.intent) }) {
+        if (!settings().shouldStartOnHome() &&
+            isActivityColdStarted(intent, savedInstanceState) &&
+            !externalSourceIntentProcessors.any { it.process(intent, navHost.navController, this.intent) }) {
```
eliserichards, line 224: I wish there was a cleaner way to write this condition 🤔

app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
```diff
@@ -181,9 +181,25 @@ class BrowserFragment : BaseBrowserFragment(), UserInteractionHandler {
         subscribeToTabCollections()
     }
 
+    private fun startOnHomeIfNeeded() {
+        val settings = context?.settings() ?: return
```
eliserichards, line 185: Do we need this validation here? Maybe we could use `requireContext().settings()` instead. What do you think?

eliserichards (changes requested):
This is looking good to me 😄 👍 Do you know if we are going to allow users an option to "never" start on home?

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -120,6 +120,16 @@ class Settings(private val appContext: Context) : PreferencesHolder {
         default = 0L
     )
 
+    /**
+     * Indicates the last when the was interacting with the BrowserFragment,
```
gabrielluong, line 124:
```suggestion
     * Indicates the last time when the was interacting with the BrowserFragment,
```

Amejia481:
I will do another iteration for adding product's feedback, and to put the functionality behind a feature flag. I think the idea it's not to have the final UI ready just to have something basic working and iterate from that.

gabrielluong:
> I will do another iteration for adding product's feedback, and to put the functionality behind a feature flag. I think the idea it's not to have the final UI ready just to have something basic working and iterate from that.

This actually makes sense. If you make this behind a feature flag, you can also go back to ask if we should launch this as a Nimbus experiment. Did we also consider telemetry?

Amejia481:
> > I will do another iteration for adding product's feedback, and to put the functionality behind a feature flag. I think the idea it's not to have the final UI ready just to have something basic working and iterate from that.
> 
> This actually makes sense. If you make this behind a feature flag, you can also go back to ask if we should launch this as a Nimbus experiment. Did we also consider telemetry?

The [user story](https://jira.mozilla.com/browse/FNX-22155) includes telemetry as part of the acceptance criteria, I was looking to land this and make a separate ticket to add the telemetry.

Amejia481:
This is ready for another pass :)

For the moment, as discussed with product we are not going to add a feature flag, as this should be on 91.

Apart from the code review, we want to do a round of manual tests, we want to cover the following cases:

1. Users should should only start on home on cold starts when selecting the options **after 4 hours** or **always**, and they have opened tabs.
2. When selecting never on cold starts users should go directly should never see the home screen when they have opened tabs.
3. When a page is open via an external app link, users should always to the browser screen independently of their settings.
4. When users don't have opened tabs should always start on home independently of their settings. 


To test case 1, you can alter your OS data & time, to go further in time, but try to do it on an emulator not on your personal phone, as this could cause other apps to not work.

app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
Amejia481: Addressed :)

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
Amejia481: What do you think about creating a helper function call `shouldNavigateToBrowserFragmentOnCouldStart()`?

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -351,6 +352,56 @@ class Settings(private val appContext: Context) : PreferencesHolder {
         default = false
     )
 
+    /**
+     * Indicates the last time when the user was interacting with the BrowserFragment,
```
gabrielluong, line 356:
```suggestion
     * Indicates the last time when the user was interacting with the [BrowserFragment],
```

app/src/main/res/values/strings.xml
```diff
@@ -570,6 +570,15 @@
...
+    <!-- Start on Home -->
+    <!-- Title of a preference that allows a user to indicate after a specified amount of time when the app should start on the home screen -->
+    <string name="preferences_start_on_home">Start on home</string>
+    <!-- Option for staring on the home screen after after four hours or inactivity -->
+    <string name="start_on_home_after_four_hours">After four hours</string>
+    <!-- Option for always staring on the home screen -->
+    <string name="start_on_home_always">Always</string>
+    <!-- Option for never staring on the home screen -->
+    <string name="start_on_home_never">Never</string>
     <!-- Summary for tabs preference when auto closing tabs setting is set to manual close-->
```
gabrielluong, line 582: Can you move these back under the "close_tabs_"?

gabrielluong (approved):

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
gabrielluong: It can either be a variable (if you don't think it's gonna be call often in other places, and prefer to keep it within the same code block where it is used) or function call. I think it's probably worthwhile to add a comment as well.

app/src/main/res/values/strings.xml
Amejia481: I put them outside as I'm not quite sure if after UX review they are going to still belong to the tabs settings.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -980,6 +983,17 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
                 startingIntent.flags and Intent.FLAG_ACTIVITY_LAUNCHED_FROM_HISTORY == 0
     }
 
+    fun shouldStartOnHome(intent: Intent? = this.intent): Boolean {
+        return components.strictMode.resetAfter(StrictMode.allowThreadDiskReads()) {
+            // We only want to open on home when users tap the app,
+            // we want to ignore other cases when the app gets open by users clicking on links.
```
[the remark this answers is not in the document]
Amejia481: 👍🏽

app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
```diff
@@ -183,7 +184,7 @@ class BrowserFragment : BaseBrowserFragment(), UserInteractionHandler {
 
     override fun onStop() {
         super.onStop()
-
+        updateLastBrowseActivity()
```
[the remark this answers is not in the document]
Amejia481: We can't add it to the `BaseBrowserFragment` as links opened in custom tabs doesn't count as user activity.

eliserichards (approved):
This is great! 🚀 I just did some manual testing and it looks like it is working perfectly!

Amejia481:
> This is great! 🚀 I just did some manual testing and it looks like it is working perfectly!

Thanks for testing ❤️

Issue #19809, opened 2021-06-04
Change default tabs tray layout to grid view


jonalmeida:
## User Story

As a user, I want to view my tabs in the grid view thumbnail) format so it's easier for me to quickly locate the specific tab I want to get back to.

## Acceptance Criteria
 * My default tabs setting is GRID view.
 * Disable the existing "grid view" CFR
 * When I open the tabs tray for the first time after updating the app, I see a message telling me that the default setting has changed. (To ENG: you can exclude this criteria for now when sizing this ticket, it's not required for iteration 1)

Linked JIRA: [FNX-22614](https://jira.mozilla.com/browse/FNX-22614)

jonalmeida:
I sized this as 1 because this is just a matter of changing the default pref in our settings and removing the CFR code (which exists in [TabsTrayInfoBannerBinding](https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayInfoBannerBinding.kt)).

cadeyrn:
@jonalmeida 

> Locate tabs easier

In my opinion it's MUCH easier to locate the tabs with the current list view. If you want to change the default, it's fine, I can change it back. But I don't agree that it will help to reach this specific goal.

With the grid view:

- … I get larger thumbnails but the bigger size doesn't add any value because I can see exactly the same
- … I can see **less tabs** at once
- … the title of every tab is **much shorter** and uesless in many cases
- … the URL of every tab is **completely missing** 

So how exactly is the grid view supposed to help to locate my tabs easier? For me it's pretty obvious that the grid view makes it harder to find tabs. 🧐

jonalmeida:
Thanks for discussing @cadeyrn ! 

> … I get larger thumbnails but the bigger size doesn't add any value because I can see exactly the same

Based on telemetry and [experimentation data](https://bugzilla.mozilla.org/show_bug.cgi?id=1386755), it showed that more people preferred grid view as the default across all channels (nightly/beta/release). From what I understand, the bigger thumbnail to identify a tab but that's just my personal guess.

> … I can see less tabs at once
> … the title of every tab is much shorter and useless in some cases
> … the URL of every tab is completely missing

This is a good point, we should definitely re-evaluate if we make the grid layout smaller so that we can see more. We had taken some original designs based on Fennec because folks like that UI a lot and allowed us to get the feature out quicker, but we can iterate on this faster now in Fenix ⚡ .

[...]

PR #19810, opened 2021-06-04
 Issue #19809: Change default tab tray view to grid layout 

jonalmeida:
Since I started throwing out code, this was a quick win as well. 😄 

- Removed the unnecessary `listTabView` as well which served no purpose.
- Added the list view key to `pref_key_tab_view_list_do_not_use` because our UI depends on individual prefs but we really don't want to be using two booleans to represent one. 😅 
- Renamed `TabsTrayInfoBannerTest.kt` -> `TabsTrayInfoBannerBindingTest.kt` so Android Studio can understand that this test file is related to `TabsTrayInfoBanner.kt`.

### Pull Request checklist

1. click on Show All Checks,


app/src/main/res/values/preference_keys.xml
```diff
@@ -257,7 +254,8 @@
     <string name="pref_key_show_collections_placeholder_home" translatable="false">pref_key_show_collections_home</string>
 
     <!-- Tabs Settings -->
-    <string name="pref_key_tab_view_list" translatable="false">pref_key_tab_view_list</string>
+    <!-- pref_key_tab_view_list_do_not_use is needed for UI implementation only and should be expected to be useful. -->
+    <string name="pref_key_tab_view_list_do_not_use" translatable="false">pref_key_tab_view_list</string>
```
gabrielluong, line 258:
```suggestion
    <string name="pref_key_tab_view_list_do_not_use" translatable="false">pref_key_tab_view_list_do_not_use</string>
```

app/src/main/java/org/mozilla/fenix/tabstray/browser/BrowserTrayList.kt
```diff
@@ -63,7 +63,7 @@ class BrowserTrayList @JvmOverloads constructor(
         TabsTouchHelper(
             observable = adapter as TabsAdapter,
             onViewHolderTouched = { swipeToDelete.isSwipeable },
-            onViewHolderDraw = { context.components.settings.listTabView }
+            onViewHolderDraw = { context.components.settings.gridTabView.not() }
```
gabrielluong, line 66: How come we prefer `.not()` compare to `!`?

app/src/main/res/values/preference_keys.xml
jonalmeida: We need this value to stay the same so that we don't break existing users. We don't want to reset their prefs if they explicitly set it to list.

app/src/main/java/org/mozilla/fenix/tabstray/browser/BrowserTrayList.kt
jonalmeida: With the long line, it was too easy for myself to not see the `!` at one end. Figured it might make it easier for the next person to see the "gridTabView" and "not" together so it clicks.

gabrielluong: That's fair. This was more of a curiosity thing since I am seeing more `.not()`.

app/src/main/java/org/mozilla/fenix/settings/TabsSettingsFragment.kt
```diff
@@ -40,7 +40,12 @@ class TabsSettingsFragment : PreferenceFragmentCompat() {
     }
 
     private fun setupPreferences() {
-        listRadioButton = requirePreference(R.string.pref_key_tab_view_list)
+        // This should be the only use case for pref_key_tab_view_list_do_not_use.
+        // In the Fenix impl of RadioGroups, we need to always have an individual pref value
+        // for it to work. This is weird for a radio group that should hold a value from that group.
+        // For the tabs tray, we only need a boolean value, so let's rely on only the
+        // pref_key_tab_view_grid and look into using the native RadioGroup in the future.
+        listRadioButton = requirePreference(R.string.pref_key_tab_view_list_do_not_use)
```
gabrielluong, line 48: This one off seems a bit weird no matter how many times I look at it. I am wondering if it would just be better to keep with the existing pattern. I assume `pref_key_tab_view_list` will still flip to the appropriate boolean value it needs and can still be used correctly in `TabsTouchHelper`. This might be better explored in a followup issue to use native RadioGroup (let's file an issue for this if it doesn't already exist).

I can see the argument that we just want to get users into using only one preference for this since this is the new default as a motivation for doing this.

Aside we also just added another radio group in https://github.com/mozilla-mobile/fenix/pull/19791, which is why I feel conflicted about this weirdness - one side it makes total sense to just use one preference, but it's also so icky to see a pref that you should not be using at all.

jonalmeida: I tried looking into it when I put this PR, but it's turtles all the way down: a radio button is meant to be a selected value based on group of options. I didn't want to change the world for this feature PR, but I wanted to remove the concept of two booleans to represent one. I removed all other references to the list pref (which was thankfully just the one now that we deleted the old code), and renamed the string ID so we don't mistakenly use it again.

I agree we need to fix the underlying problem, rather than the individual features. I'll file a bug for this to discuss it there. No need to block #19791 about this either until we have a good solution.

jonalmeida: Filed as https://github.com/mozilla-mobile/fenix/issues/19829.

rocketsroger (approved):
Assuming other comments are resolved.  Looks good to me!

jonalmeida:
Fixed the failing UI test that expected the list view layout as default.

Issue #19191, opened 2021-04-22
[Bug]Tab selection checkmark is not displayed after checking synced tabs


sflorean:
[comment]: # (Please do your best to search for duplicate issues before filing a new issue so we can keep our issue board clean)
[comment]: # (Every issue should have the exact bug and steps to reproduce described in it. Please do not file feedback list tickets as it is difficult to parse them and address their individual points)
[comment]: # (Read https://github.com/mozilla-mobile/fenix#i-want-to-file-an-issue for more information)

## Steps to reproduce
1. Open 2-3 pages.
2. Log in with a Firefox account on 2 devices.
3. Open tabs tray -> synced tabs.
4. While in tabs tray- go to normal tabs.
5. Perform a long tap in order to select the tabs.

### Expected behavior
Tabs are selected and the check is displayed above the links.

### Actual behavior
Tabs are selected but the "check" sign is not displayed. 

### Device information
* Device vendor / model and Android version: LG G7 Fit (Android 8.1), Huawei MediaPad M2 (Android 5.1.1).
* Firefox for Android version: Beta 89-b1, Nightly 4/22
 
![Screenshot_2021-04-22-10-58-50](https://user-images.githubusercontent.com/58466444/115728697-611def00-a38d-11eb-89ba-a3f999f0f45f.png)

jonalmeida:
Thanks! I can reproduce this bug, but it needs some investigation to know what's the cause.

mcarare:
@jonalmeida The issue can be reproduced even without signing in. If you switch from the normal tab to the sync tab and then back the issue reproduces every time. Closing and reopening tab tray restores normal behavior.

jonalmeida:
Added this to the kanban board because it may prove to be an issue later with new MR2 work.

abodea:
Verified as fixed on the latest Nightly 6/9 with Google Pixel 4 XL (11), OnePlus Nord (10).
Note that I was able to select and the checkmark was displayed without issues.

PR #19821, opened 2021-06-04
For #19191 - Fixes missing multi-select checkmarks

codrut-topliceanu:
For https://github.com/mozilla-mobile/fenix/issues/19191
### Pull Request checklist

- [x] **Screenshots**: This PR includes screenshots or GIFs.

1. click on Show All Checks,

https://user-images.githubusercontent.com/60002907/120786092-82e6c600-c536-11eb-96c8-c697d60f58d0.mp4

### Notes:
The problem arises from the fact that when changing the page of the tabs tray to either Sync or Private `onDetachedFromWindow` (`BrowserTrayList.kt`) is called on the Normal adapter, triggering `adapter?.onDetachedFromRecyclerView` ->`selectedItemAdapterBinding.stop()`. 
But when returning to the Normal Tabs Page -> `onAttachedToWindow()` is executed but we would not run `adapter?.onAttachedToRecyclerView` so the `selectedItemAdapterBinding` would not restart.

An alternative solution would be to configure the `tabsTray` pager in `TabsTrayFragment.kt` to `offscreenPageLimit=2` thus keeping all tab page adapters running at the same time and bypassing the `onDetachedFromWindow` call altogether. That would make switching between the pages faster but at the cost of memory.

PS: Thank you @Mugurell for all the help <3

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/tabstray/browser/BrowserTrayList.kt
+++ b/app/src/main/java/org/mozilla/fenix/tabstray/browser/BrowserTrayList.kt
@@ -73,6 +73,8 @@ class BrowserTrayList @JvmOverloads constructor(
         tabsFeature.start()
         swipeToDelete.start()
 
+        adapter?.onAttachedToRecyclerView(this)
+
         touchHelper.attachToRecyclerView(this)
     }
 


jonalmeida (approved):
Excellent find!

Issue #19738, opened 2021-06-01
Set secure flag when on the private tabs page of the tabs tray


jonalmeida:
This is something that the old tabs tray [used to do](https://github.com/mozilla-mobile/fenix/blob/33b27892708d973b8e606c77d99d43cd9c8b12cb/app/src/main/java/org/mozilla/fenix/tabtray/TabTrayDialogFragment.kt#L276-L282) as well.

Should be easy enough with a "SecurePrivateTabBinding" that observes the `TabsTrayState.selectedPage` when it's `Page.PrivateTab` and set the appropriate flags.

You can get some ideas from [`SecureWindowFeature`](https://searchfox.org/mozilla-mobile/source/android-components/components/feature/privatemode/src/main/java/mozilla/components/feature/privatemode/feature/SecureWindowFeature.kt#45).

**⚠️ Note:** We need to also be aware if the `BrowsingModeManager` is also private at the same time before we decide to clear the secure flag.

jonalmeida:
cc: @codrut-topliceanu if you're interested. 🙂

jonalmeida:
I think we need to remove the `drop(1)` from SelectBannerBinding and SelectHandleBinding as well.

jonalmeida:
We may be able to re-enable `createFirstCollectionTest` after this lands.

LaurentiuApahideanSV:
Verified as fixed on Firefox Preview Nightly 6/10. I tested the issues with Tab view set to List and Grid.
### Devices used:
- OnePlus 6T (Android 9)
- Google Pixel 3 XL (Android 9)

PR #19824, opened 2021-06-04
For #19738 - Secure mode enabled in private tabs tray

codrut-topliceanu:
Prevents screenshots while tabs tray with private tabs page is open.

For https://github.com/mozilla-mobile/fenix/issues/19738

### Pull Request checklist

1. click on Show All Checks,

https://user-images.githubusercontent.com/60002907/120822248-e4bb2600-c55e-11eb-91d5-cd446fe9bbf2.mp4

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/tabstray/SecureTabsTrayBinding.kt
+++ b/app/src/main/java/org/mozilla/fenix/tabstray/SecureTabsTrayBinding.kt
@@ -1,0 +1,42 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.tabstray
+
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.collect
+import kotlinx.coroutines.flow.map
+import mozilla.components.lib.state.helpers.AbstractBinding
+import mozilla.components.support.ktx.kotlinx.coroutines.flow.ifAnyChanged
+import org.mozilla.fenix.ext.removeSecure
+import org.mozilla.fenix.ext.secure
+import org.mozilla.fenix.utils.Settings
+
+/**
+ * Sets TabsTrayFragment flags to secure when private tabs list is selected.
+ */
+@OptIn(ExperimentalCoroutinesApi::class)
+class SecureTabsTrayBinding(
+    store: TabsTrayStore,
+    private val settings: Settings,
+    private val tabsTrayFragment: TabsTrayFragment
+) : AbstractBinding<TabsTrayState>(store) {
+
+    override suspend fun onState(flow: Flow<TabsTrayState>) {
+        flow.map { it }
+            .ifAnyChanged { state ->
+                arrayOf(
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayFragment.kt
```diff
@@ -358,6 +360,12 @@ class TabsTrayFragment : AppCompatDialogFragment() {
 
     @VisibleForTesting
     internal fun selectTabPosition(position: Int, smoothScroll: Boolean) {
+        if (position == Page.PrivateTabs.ordinal) {
```
jonalmeida, line 363: This method is for selecting the tab position, so let's make that is all it does. We have other UI "features" that are based on the selected page, but if we put them all here it would be a pretty big method. 😄 

If we use a composable approach with the `AbstractBinding` like the others above, we can isolate this specific feature in one class so we can add/remove it, test it easily and without needing to mock lots of Android framework code, and most importantly for me, it's easier to follow what it's trying to do.

Have a look at the `FloatingActionButtonBinding`, which uses settings and observes the selected page from the store.

app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayFragment.kt
```diff
@@ -358,6 +360,12 @@ class TabsTrayFragment : AppCompatDialogFragment() {
...
+            this.secure()
+        } else if (!requireContext().settings().lastKnownMode.isPrivate) {
```
jonalmeida, line 365: TIL about `lastKnownMode`.

jonalmeida (changes requested):
Nice! Logic looks good, but let's move this into it's own "binding" class like the other features.

app/src/main/java/org/mozilla/fenix/tabstray/SecureTabsTrayBinding.kt
```diff
@@ -0,0 +1,42 @@
...
+import org.mozilla.fenix.utils.Settings
+
+/**
+ * Sets TabsTrayFragment flags to secure when private tabs list is selected.
+ */
+@OptIn(ExperimentalCoroutinesApi::class)
+class SecureTabsTrayBinding(
+    store: TabsTrayStore,
+    private val settings: Settings,
+    private val tabsTrayFragment: TabsTrayFragment
```
jonalmeida, line 24: nit: I think you can get away with using just `Fragment` for type-checking sake.

jonalmeida (commented):
Looks great, let's add some tests and we can land it! 🙂

app/src/main/java/org/mozilla/fenix/tabstray/SecureTabsTrayBinding.kt
```diff
@@ -0,0 +1,43 @@
...
+    private val fragment: Fragment
+) : AbstractBinding<TabsTrayState>(store) {
+
+    override suspend fun onState(flow: Flow<TabsTrayState>) {
+        flow.map { it }
+            .ifAnyChanged { state ->
+                arrayOf(
+                    state.selectedPage
+                )
+            }
```
jonalmeida, line 34: Missed this earlier, but for reference since it's not worth sending back for now. You can write this as:
```suggestion
        flow.map { it.selectedPage }
            .ifChanged()
```
```kotlin
```

jonalmeida (approved):
🙌

PR #19867, opened 2021-06-07
For #19797 - Go to problems page if account exists but not signed in

### LINKED ISSUE #19797: [Bug] The "Sign in to sync" option from the three-dot menu is inactive after changing the password
## Steps to reproduce
1. Sign in with the same account on Desktop, and on mobile.
2. From Desktop, change the password.
3. On mobile trigger sync.
4. After the "Reconnect to sync" option is displayed on the three-dot main menu, open a page.
5. Tap on the three-dot menu from the page. Or tap on the three-dot menu from the homescreen.
6. Tap on "Sign in to sync".

### Expected behavior
The user is redirected to the "Reconnect to Sync" page.

### Actual behavior
Nothing happens. The menu is dismissed.

### Device information
* Devices: Google Pixel (Android 10), Samsung Galaxy Tab A6 (Android 5.1.1), OnePlus A3 (Android 6.0.1)
* Firefox for Android version: Beta 90.0.0-beta.1, RC 89.1.1, and Nightly 91.0a1 from 6/3
* Reproducible in Beta 89.0.0-beta.10 as well

[logcat.txt](https://github.com/mozilla-mobile/fenix/files/6589450/logcat.txt)

https://user-images.githubusercontent.com/40694158/120599297-ffeb4000-c44f-11eb-876b-43f18f1b6115.mp4

* **On OnePlus A3, and Samsung Galaxy Tab A6, the account is still displayed along with the error message**:

Untitled

Mugurell:
Fixes #19797

Previously if an account existed but it wasn't signed in to we would open the fragment to allow the user to sign in to the Firefox account but then [exit early because an account was already existing](https://github.com/mozilla-mobile/fenix/blob/990bfa7e6dd5894d61473a41da6129e6947974e2/app/src/main/java/org/mozilla/fenix/settings/account/TurnOnSyncFragment.kt#L92-L98).

https://user-images.githubusercontent.com/11428869/120995966-942a0f80-c78e-11eb-89fc-ecf6e78baf56.mp4

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/accounts/FenixAccountManager.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/accounts/FenixAccountManager.kt
@@ -5,18 +5,13 @@
 package org.mozilla.fenix.components.accounts
 
 import android.content.Context
-import mozilla.components.service.fxa.manager.FxaAccountManager
 import org.mozilla.fenix.ext.components
 
 /**
- * Component which holds a reference to [FxaAccountManager]. Manages account authentication,
- * profiles, and profile state observers.
+ * Contains helper methods for querying Firefox Account state and its properties.
  */
-open class FenixAccountManager(context: Context) {
-    val accountManager = context.components.backgroundServices.accountManager
-
-    val authenticatedAccount
-        get() = accountManager.authenticatedAccount() != null
+class FenixAccountManager(context: Context) {
+    private val accountManager = context.components.backgroundServices.accountManager
 
     /**
      * Returns the Firefox Account email if authenticated in the app, `null` otherwise.
--- a/app/src/main/java/org/mozilla/fenix/components/accounts/FenixAccountManager.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/accounts/FenixAccountManager.kt
@@ -18,8 +18,15 @@ open class FenixAccountManager(context: Context) {
     val authenticatedAccount
         get() = accountManager.authenticatedAccount() != null
 
-    val accountProfileEmail
-        get() = accountManager.accountProfile()?.email
+    /**
+     * Returns the Firefox Account email if authenticated in the app, `null` otherwise.
+     */
+    val accountProfileEmail: String?
+        get() = if (accountState == AccountState.AUTHENTICATED) {
+            accountManager.accountProfile()?.email
+        } else {
+            null
+        }
 
     /**
      * The current state of the Firefox Account. See [AccountState].
--- a/app/src/main/java/org/mozilla/fenix/components/accounts/FenixAccountManager.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/accounts/FenixAccountManager.kt
@@ -22,6 +22,20 @@ open class FenixAccountManager(context: Context) {
         get() = accountManager.accountProfile()?.email
 
     /**
+     * The current state of the Firefox Account. See [AccountState].
+     */
+    val accountState: AccountState
+        get() = if (accountManager.authenticatedAccount() == null) {
+            AccountState.NO_ACCOUNT
+        } else {
+            if (accountManager.accountNeedsReauth()) {
+                AccountState.NEEDS_REAUTHENTICATION
+            } else {
+                AccountState.AUTHENTICATED
+            }
+        }
+
+    /**
      * Check if the current account is signed in and authenticated.
      */
     fun signedInToFxa(): Boolean {
@@ -30,4 +44,24 @@ open class FenixAccountManager(context: Context) {
 
         return account != null && !needsReauth
     }
+}
+
+/**
+ * General states as an overview of the current Firefox Account.
+ */
+enum class AccountState {
+    /**
+     * There is no known Firefox Account.
+     */
+    NO_ACCOUNT,
+
+    /**
+     * A Firefox Account exists but needs to be re-authenticated.
+     */
+    NEEDS_REAUTHENTICATION,
+
+    /**
+     * A Firefox Account exists and the user is currently signed into it.
+     */
+    AUTHENTICATED,
 }
--- a/app/src/main/java/org/mozilla/fenix/components/accounts/FenixAccountManager.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/accounts/FenixAccountManager.kt
@@ -34,16 +34,6 @@ open class FenixAccountManager(context: Context) {
                 AccountState.AUTHENTICATED
             }
         }
-
-    /**
-     * Check if the current account is signed in and authenticated.
-     */
-    fun signedInToFxa(): Boolean {
-        val account = accountManager.authenticatedAccount()
-        val needsReauth = accountManager.accountNeedsReauth()
-
-        return account != null && !needsReauth
-    }
 }
 
 /**
--- a/app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
@@ -348,7 +348,9 @@ open class DefaultToolbarMenu(
         R.drawable.ic_signed_out,
         primaryTextColor()
     ) {
-        onItemTapped.invoke(ToolbarMenu.Item.SyncAccount(accountManager.signedInToFxa()))
+        onItemTapped.invoke(
+            ToolbarMenu.Item.SyncAccount(accountManager.accountState)
+        )
     }
 
     @VisibleForTesting(otherwise = PRIVATE)
--- a/app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
@@ -332,16 +332,8 @@ open class DefaultToolbarMenu(
         onItemTapped.invoke(ToolbarMenu.Item.Quit)
     }
 
-    private fun getSyncItemTitle(): String {
-        val authenticatedAccount = accountManager.authenticatedAccount
-        val email = accountManager.accountProfileEmail
-
-        return if (authenticatedAccount && !email.isNullOrEmpty()) {
-            email
-        } else {
-            context.getString(R.string.sync_menu_sign_in)
-        }
-    }
+    private fun getSyncItemTitle() =
+        accountManager.accountProfileEmail ?: context.getString(R.string.sync_menu_sign_in)
 
     val syncMenuItem = BrowserMenuImageText(
         getSyncItemTitle(),
--- a/app/src/main/java/org/mozilla/fenix/components/toolbar/ToolbarMenu.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/toolbar/ToolbarMenu.kt
@@ -6,6 +6,7 @@ package org.mozilla.fenix.components.toolbar
 
 import mozilla.components.browser.menu.BrowserMenuBuilder
 import mozilla.components.browser.menu.item.BrowserMenuItemToolbar
+import org.mozilla.fenix.components.accounts.AccountState
 
 interface ToolbarMenu {
     sealed class Item {
@@ -22,7 +23,7 @@ interface ToolbarMenu {
         object AddToTopSites : Item()
         object InstallPwaToHomeScreen : Item()
         object AddToHomeScreen : Item()
-        data class SyncAccount(val signedIn: Boolean) : Item()
+        data class SyncAccount(val accountState: AccountState) : Item()
         object AddonsManager : Item()
         object Quit : Item()
         object OpenInApp : Item()
--- a/app/src/main/java/org/mozilla/fenix/home/HomeMenu.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/HomeMenu.kt
@@ -91,16 +91,8 @@ class HomeMenu(
         }
     }
 
-    private fun getSyncItemTitle(): String {
-        val authenticatedAccount = accountManager.authenticatedAccount
-        val email = accountManager.accountProfileEmail
-
-        return if (authenticatedAccount && !email.isNullOrEmpty()) {
-            email
-        } else {
-            context.getString(R.string.sync_menu_sign_in)
-        }
-    }
+    private fun getSyncItemTitle(): String =
+        accountManager.accountProfileEmail ?: context.getString(R.string.sync_menu_sign_in)
 
     private val syncSignInMenuItem = BrowserMenuImageText(
         getSyncItemTitle(),
--- a/app/src/main/java/org/mozilla/fenix/home/HomeMenu.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/HomeMenu.kt
@@ -24,6 +24,7 @@ import mozilla.components.concept.sync.AuthType
 import mozilla.components.concept.sync.OAuthAccount
 import mozilla.components.support.ktx.android.content.getColorFromAttr
 import org.mozilla.fenix.R
+import org.mozilla.fenix.components.accounts.AccountState
 import org.mozilla.fenix.components.accounts.FenixAccountManager
 import org.mozilla.fenix.experiments.FeatureId
 import org.mozilla.fenix.ext.components
@@ -45,7 +46,7 @@ class HomeMenu(
         object History : Item()
         object Downloads : Item()
         object Extensions : Item()
-        data class SyncAccount(val signedIn: Boolean) : Item()
+        data class SyncAccount(val accountState: AccountState) : Item()
         object WhatsNew : Item()
         object Help : Item()
         object Settings : Item()
@@ -106,7 +107,7 @@ class HomeMenu(
         R.drawable.ic_synced_tabs,
         primaryTextColor
     ) {
-        onItemTapped.invoke(Item.SyncAccount(accountManager.signedInToFxa()))
+        onItemTapped.invoke(Item.SyncAccount(accountManager.accountState))
     }
 
     val desktopItem = BrowserMenuImageSwitch(
--- a/app/src/test/java/org/mozilla/fenix/components/toolbar/DefaultBrowserToolbarMenuControllerTest.kt
+++ b/app/src/test/java/org/mozilla/fenix/components/toolbar/DefaultBrowserToolbarMenuControllerTest.kt
@@ -57,6 +57,7 @@ import org.mozilla.fenix.browser.readermode.ReaderModeController
 import org.mozilla.fenix.collections.SaveCollectionStep
 import org.mozilla.fenix.components.FenixSnackbar
 import org.mozilla.fenix.components.TabCollectionStorage
+import org.mozilla.fenix.components.accounts.AccountState
 import org.mozilla.fenix.components.metrics.Event
 import org.mozilla.fenix.components.metrics.MetricController
 import org.mozilla.fenix.ext.components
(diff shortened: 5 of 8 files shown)


app/src/main/java/org/mozilla/fenix/components/toolbar/ToolbarMenu.kt
```diff
@@ -22,7 +22,7 @@ interface ToolbarMenu {
         object AddToTopSites : Item()
         object InstallPwaToHomeScreen : Item()
         object AddToHomeScreen : Item()
-        data class SyncAccount(val signedIn: Boolean) : Item()
+        data class SyncAccount(val existingAccount: Boolean, val signedIn: Boolean) : Item()
```
gabrielluong, line 25: I think we should go with the same language `isAuthenticatedAccount` since `existingAccount` changes the meaning of it from the original intent. I prefix it with `is` to make it super obvious that it represents a `Boolean` when reading the code.

app/src/test/java/org/mozilla/fenix/components/toolbar/DefaultBrowserToolbarMenuControllerTest.kt
```diff
@@ -562,25 +562,40 @@ class DefaultBrowserToolbarMenuControllerTest {
     }
 
     @Test
-    fun `WHEN sync sign in menu item is pressed AND account is signed out THEN navigate to sync sign in`() = runBlockingTest {
-        val item = ToolbarMenu.Item.SyncAccount(false)
-        val directions = BrowserFragmentDirections.actionGlobalTurnOnSync()
-
+    fun `GIVEN account exists and the user is signed in WHEN sync sign in menu item is pressed THEN navigate to account settings`() = runBlockingTest {
```
gabrielluong, line 565:
```suggestion
    fun `GIVEN account is authenticated and the user is signed in WHEN sync sign in menu item is pressed THEN navigate to account settings`() = runBlockingTest {
```

gabrielluong (approved):

app/src/test/java/org/mozilla/fenix/components/toolbar/DefaultBrowserToolbarMenuControllerTest.kt
Mugurell: The below test would start with
> GIVEN account is authenticated and the user is not signed in ...
This doesn't sound right.

app/src/main/java/org/mozilla/fenix/components/toolbar/ToolbarMenu.kt
Mugurell: I think rather the naming used in fxa should be updated to more closely express the current status of the account.
The documentation for `authenticatedAccount` [says](https://github.com/mozilla-mobile/android-components/blob/bfaf514793dd20f9f12d3f551f9ff3e470f52727/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt#L325-L338)
"..Returned [OAuthAccount] may need to be re-authenticated."

Based on this and my below comment about the logic used in the app and tests `existingAccount` offers more info about the current status while `authenitcatedAccount` but not signed in seems confusing.

Mugurell: Maybe `accountExists` expresses better that that parameter is for a Boolean?

app/src/test/java/org/mozilla/fenix/components/toolbar/DefaultBrowserToolbarMenuControllerTest.kt
gabrielluong:
```suggestion
    fun `GIVEN account is authenticated WHEN sync sign in menu item is pressed THEN navigate to account settings`() = runBlockingTest {
```

app/src/main/java/org/mozilla/fenix/components/toolbar/ToolbarMenu.kt
gabrielluong: For this instance, `authenticatedAccount` is defined in https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/java/org/mozilla/fenix/components/accounts/FenixAccountManager.kt#19. It is slightly odd that we used the same naming as the account manager but don't technically return the `OAuthAccount`.

You're right that the authenticated account might need re-auth, but I am not too particular about that detail since it is in some sort of authenticated state. My main point would be to persist the same naming of instead of changing it midway in the ideal case. That said the naming is open to change - I don't particularly see a problem with simply persisting with the `authenticatedAccount` naming due to the fact that it is still in sort of authenticated state and it is known that consumers should check if it needs to re-auth. 

So, I think it boils down to the use case here, which we basically have tri-state (Authenticated, ReAuth, Not Signed in). Perhaps, we should change `FenixAccountManager` to return something that is super clear about the state instead of us having to basically do the same thing all the time with getting the `authenticatedAccount` and checking `signedIn`. If you look at `FenixAccountManager`, the whole logic is kinda duplicated multiple times. 

Feel free to ping me if this wasn't clear. I was going through the code, thinking, writing and revising as my thinking was kinda all over the place.

Mugurell: Maybe `authenticatedAccountExists` would be a more appropriate parameter name?

Mugurell: After talking with Gabriel I've added a new `accountState` property to FenixAccountManager and will use this simpler way of inferring the directions to which to navigate to.

app/src/test/java/org/mozilla/fenix/components/toolbar/DefaultBrowserToolbarMenuControllerTest.kt
```diff
@@ -562,25 +563,36 @@ class DefaultBrowserToolbarMenuControllerTest {
...
 
-        verify { navController.navigate(directions, null) }
+        verify { navController.navigate(accountProblemDirections, null) }
+    }
+
+    @Test
+    fun `GIVEN account doesn't exist WHEN sync sign in menu item is pressed THEN navigate to sign in`() = runBlockingTest {
+        val turnOnSyncDirections = BrowserFragmentDirections.actionGlobalTurnOnSync()
+        val controller = createController(scope = this, store = browserStore)
+        val signedInUnexistentAccountItem = ToolbarMenu.Item.SyncAccount(AccountState.NO_ACCOUNT)
```
gabrielluong, line 591: I think we might prefer to use `signInToSyncItem` just so it aligns with what we actually see visually in the toolbar menu. Otherwise, I would keep it simple and just keep this as `item`.

I will note that it is a bit odd that we decided to change this variable name when we have been calling it `item` and the other ones.

app/src/test/java/org/mozilla/fenix/components/toolbar/DefaultBrowserToolbarMenuControllerTest.kt
```diff
@@ -562,25 +563,36 @@ class DefaultBrowserToolbarMenuControllerTest {
...
         val controller = createController(scope = this, store = browserStore)
+
         controller.handleToolbarItemInteraction(item)
 
-        verify { navController.navigate(directions, null) }
+        verify { navController.navigate(accountProblemDirections, null) }
+    }
+
+    @Test
+    fun `GIVEN account doesn't exist WHEN sync sign in menu item is pressed THEN navigate to sign in`() = runBlockingTest {
```
gabrielluong, line 588:
```suggestion
    fun `GIVEN account doesn't exist WHEN sign in to sync menu item is pressed THEN navigate to sign in`() = runBlockingTest {
```

This is about aligning it with the same wording as what we see in the toolbar menu.

app/src/test/java/org/mozilla/fenix/components/toolbar/DefaultBrowserToolbarMenuControllerTest.kt
```diff
@@ -562,25 +563,36 @@ class DefaultBrowserToolbarMenuControllerTest {
...
-        verify { navController.navigate(directions, null) }
+        verify { navController.navigate(accountSettingsDirections, null) }
     }
 
     @Test
-    fun `WHEN sync sign in menu item is pressed AND account is signed in THEN navigate to sync sign in`() = runBlockingTest {
-        val item = ToolbarMenu.Item.SyncAccount(true)
-        val directions = BrowserFragmentDirections.actionGlobalAccountSettingsFragment()
-
+    fun `GIVEN account exists and the user is not signed in WHEN sync sign in menu item is pressed THEN navigate to account problem fragment`() = runBlockingTest {
```
gabrielluong, line 577:
```suggestion
    fun `GIVEN account exists and the user is not signed in WHEN sign in to sync menu item is pressed THEN navigate to account problem fragment`() = runBlockingTest {
```

app/src/test/java/org/mozilla/fenix/components/toolbar/DefaultBrowserToolbarMenuControllerTest.kt
```diff
@@ -562,25 +563,36 @@ class DefaultBrowserToolbarMenuControllerTest {
     }
 
     @Test
-    fun `WHEN sync sign in menu item is pressed AND account is signed out THEN navigate to sync sign in`() = runBlockingTest {
-        val item = ToolbarMenu.Item.SyncAccount(false)
-        val directions = BrowserFragmentDirections.actionGlobalTurnOnSync()
-
+    fun `GIVEN account exists and the user is signed in WHEN sync sign in menu item is pressed THEN navigate to account settings`() = runBlockingTest {
```
gabrielluong, line 566:
```suggestion
    fun `GIVEN account exists and the user is signed in WHEN sign in to sync menu item is pressed THEN navigate to account settings`() = runBlockingTest {
```

gabrielluong (approved):

app/src/main/java/org/mozilla/fenix/components/accounts/FenixAccountManager.kt
```diff
@@ -21,6 +21,20 @@ open class FenixAccountManager(context: Context) {
     val accountProfileEmail
```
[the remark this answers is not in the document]
Mugurell: `authenticatedAccount` seem to have valid uses, like [here](https://github.com/mozilla-mobile/fenix/blob/bf3f81d271382eff0780cfdd81f6dea35c13dc7f/app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt#L336)

`signedToFxa` was used in a similar buggy way for the home menu. Switched that to use `accountState`,  the same used in browser menu.
Since that was it's only use I've removed this method.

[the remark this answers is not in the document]
gabrielluong: Maybe we should also refactor `accountProfileEmail` to encapsulate the  `authenticatedAccount` usage to really discourage it from being used erroneously.  For the `accountProfileEmail` getter, I imagine we can call the `authenticatedAccount` check and just return a nullable string for the email.

That way we can simplify `getSyncItemTitle` to 
```
val accountManager.accountProfileEmail()

return if (!email.isNullOrEmpty()) {            
   email        
} else {            
}
```

We probably don't want to ever return the email if the account is not authenticated. I think this will allow us remove the `authenticatedAccount` getter completely or make it private to really discourage it.

```
fun getAccountProfileEmail() : String? {
  return if (accountManager.authenticatedAccount() != null) {
    accountManager.accountProfile()?.email
 } else {
   null
}
```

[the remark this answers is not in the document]
Mugurell: Refactored `accountProfileEmail` to only try and return the email for the account if that is in the authenticated state.
Having this check here, in one place, means consumers will not need to double check the account status.
This also allowed removing the `authenticatedAccount` property from `FenixAccountManager` which now has only two methods, more as helpers for interacting with an account: `accountProfileEmail` and `accountState`.

app/src/main/java/org/mozilla/fenix/components/accounts/FenixAccountManager.kt
```diff
@@ -5,29 +5,55 @@
 package org.mozilla.fenix.components.accounts
 
 import android.content.Context
-import mozilla.components.service.fxa.manager.FxaAccountManager
 import org.mozilla.fenix.ext.components
 
 /**
- * Component which holds a reference to [FxaAccountManager]. Manages account authentication,
- * profiles, and profile state observers.
+ * Contains helper methods for querying FXA account state and it's properties.
```
gabrielluong, line 11:
```suggestion
 * Contains helper methods for querying the FxA account state and its properties.
```

Instead of FXA, we should use FxA which is its proper acronym. Alternatively, I would just spell out "Firefox Account" for outside readers who aren't aware of our acronym for internal systems.

app/src/main/java/org/mozilla/fenix/components/accounts/FenixAccountManager.kt
```diff
@@ -5,29 +5,55 @@
...
  */
-open class FenixAccountManager(context: Context) {
-    val accountManager = context.components.backgroundServices.accountManager
+class FenixAccountManager(context: Context) {
+    private val accountManager = context.components.backgroundServices.accountManager
 
-    val authenticatedAccount
-        get() = accountManager.authenticatedAccount() != null
+    /**
+     * Returns the FXA email if authenticated in the app, `null` otherwise.
```
gabrielluong, line 17:
```suggestion
     * Returns the FxA email if authenticated in the app, `null` otherwise.
```

app/src/main/java/org/mozilla/fenix/components/accounts/FenixAccountManager.kt
```diff
@@ -5,29 +5,55 @@
...
+     */
+    val accountProfileEmail: String?
+        get() = if (accountState == AccountState.AUTHENTICATED) {
+            accountManager.accountProfile()?.email
+        } else {
+            null
+        }
+
+    /**
+     * The current state of the FXA account. See [AccountState].
```
gabrielluong, line 27:
```suggestion
     * The current state of the FxA account. See [AccountState].
```

app/src/main/java/org/mozilla/fenix/components/accounts/FenixAccountManager.kt
```diff
@@ -5,29 +5,55 @@
...
+            } else {
+                AccountState.AUTHENTICATED
+            }
+        }
+}
 
-    val accountProfileEmail
-        get() = accountManager.accountProfile()?.email
+/**
+ * General states as an overview of the current FXA account.
```
gabrielluong, line 42:
```suggestion
 * General states as an overview of the current FxA account.
```

app/src/main/java/org/mozilla/fenix/components/accounts/FenixAccountManager.kt
```diff
@@ -5,29 +5,55 @@
...
+ */
+enum class AccountState {
+    /**
+     * There is no known FXA account.
```
gabrielluong, line 46:
```suggestion
     * There is no known FxA account.
```

app/src/main/java/org/mozilla/fenix/components/accounts/FenixAccountManager.kt
```diff
@@ -5,29 +5,55 @@
...
+     */
+    NO_ACCOUNT,
 
     /**
-     * Check if the current account is signed in and authenticated.
+     * An FXA account exists but needs to be re-authenticated with.
```
gabrielluong, line 51:
```suggestion
     * An FxA account exists but needs to be re-authenticated.
```

app/src/main/java/org/mozilla/fenix/components/accounts/FenixAccountManager.kt
```diff
@@ -5,29 +5,55 @@
...
      */
-    fun signedInToFxa(): Boolean {
-        val account = accountManager.authenticatedAccount()
-        val needsReauth = accountManager.accountNeedsReauth()
+    NEEDS_REAUTHENTICATION,
 
-        return account != null && !needsReauth
-    }
+    /**
+     * An FXA account exists and the user is currently signed in into it.
```
gabrielluong, line 56:
```suggestion
     * An FxA account exists and the user is currently signed into it.
```

app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
```diff
@@ -333,22 +333,17 @@ open class DefaultToolbarMenu(
...
     private fun getSyncItemTitle(): String {
-        val authenticatedAccount = accountManager.authenticatedAccount
-        val email = accountManager.accountProfileEmail
-
-        return if (authenticatedAccount && !email.isNullOrEmpty()) {
-            email
-        } else {
-            context.getString(R.string.sync_menu_sign_in)
-        }
+        return accountManager.accountProfileEmail ?: context.getString(R.string.sync_menu_sign_in)
```
gabrielluong, line 336: Could probably drop the `{ return }` with using `=`

app/src/main/java/org/mozilla/fenix/home/HomeMenu.kt
```diff
@@ -91,22 +92,15 @@ class HomeMenu(
...
     private fun getSyncItemTitle(): String {
-        val authenticatedAccount = accountManager.authenticatedAccount
-        val email = accountManager.accountProfileEmail
-
-        return if (authenticatedAccount && !email.isNullOrEmpty()) {
-            email
-        } else {
-            context.getString(R.string.sync_menu_sign_in)
-        }
+        return accountManager.accountProfileEmail ?: context.getString(R.string.sync_menu_sign_in)
```
gabrielluong, line 95: Same as above, but I will also note the duplication of code between the 2. No idea, if we have a util class that happens to be used by both menus that we can throw this into, if not, don't worry about it for now.

app/src/test/java/org/mozilla/fenix/components/toolbar/DefaultBrowserToolbarMenuControllerTest.kt
```diff
@@ -562,25 +563,36 @@ class DefaultBrowserToolbarMenuControllerTest {
...
 
-        verify { navController.navigate(directions, null) }
+        verify { navController.navigate(accountProblemDirections, null) }
+    }
+
+    @Test
+    fun `GIVEN account doesn't exist WHEN sign in to sync menu item is pressed THEN navigate to sign in`() = runBlockingTest {
+        val turnOnSyncDirections = BrowserFragmentDirections.actionGlobalTurnOnSync()
+        val controller = createController(scope = this, store = browserStore)
+        val signedInUnexistentAccountItem = ToolbarMenu.Item.SyncAccount(AccountState.NO_ACCOUNT)
```
gabrielluong, line 591: I think my last comment about this was to name this as `signInToSyncItem` so it aligns with what we see visually. I don't think we should overthink this and just call it `item` as we have done for all the other tests despite the different account state passed.

 ```suggestion
        val item = ToolbarMenu.Item.SyncAccount(AccountState.NO_ACCOUNT)
```

gabrielluong (approved):

app/src/main/java/org/mozilla/fenix/components/accounts/FenixAccountManager.kt
Mugurell: Thanks, will use "Firefox Account".

Mugurell:
`deleteAllHistory` ui test failed because of
>  androidx.test.espresso.NoMatchingViewException: No views in hierarchy found matching: (with id: org.mozilla.fenix.debug:id/history_empty_view and view has effective visibility=VISIBLE) 

Seems unrelated. Restarting CI.

PR #19897, opened 2021-06-08
No issue: Add full-junit-result in Flank config

AaronMT:
``JUnitReport.xml`` and ``FullJunitReport.xml`` are artifacts with differences. Full will include all tests results, including when tests failed and when passed on retry. This might expose more additional results for capturing as part of my Job/Push collection data on certain tasks. I think Junit report will contain one result per test. If a test fails, and then passes on retry, it will be recorded as a successful execution.


isabelrios (approved):
Sounds good, let's try this to gather more data in case that helps, thanks!

PR #19802, opened 2021-06-03
For #19801 - Run taskgraph generation with Python 3

ahal:
This will involve updating to the latest taskgraph and decision-mobile image.

PR #19954, opened 2021-06-10
Fix unicode issue in release promotion code

bhearsum:
@ahal - this looks like minor fallout from #19802. I caught it while testing another patch on `staging-fenix` :)


escapewindow (approved):

MihaiTabara:
@bhearsum should we cherry-pick this to Android-Components too, given https://github.com/mozilla-mobile/android-components/blob/master/taskcluster/ac_taskgraph/release_promotion.py#L164 ?

escapewindow:
I think we’re still on py2 in a-c.

MihaiTabara:
Ah, good point, we're blocked on cherry-picking https://github.com/mozilla-mobile/fenix/pull/19802 first I suppose. Don't mind me then.

PR #19958, opened 2021-06-10
Closes #19921: Update appcompat and fragment dependencies to 1.3.x

### LINKED ISSUE #19921: Update appcompat and fragment dependencies to 1.3.x.
This will be a prerequisite for using Jetpack Compose eventually.

Follow-ups for deprecation:
https://github.com/mozilla-mobile/fenix/issues/19919
https://github.com/mozilla-mobile/fenix/issues/19920
https://github.com/mozilla-mobile/fenix/issues/19929

rocketsroger:
Fixess #19921

### Pull Request checklist

1. click on Show All Checks,


rocketsroger:
This is the Fenix side change for https://github.com/mozilla-mobile/android-components/pull/10359

pocmo (commented):
Form my side okay. Deferring review to the performance team. I guess we do not really get around the main thread requirement.
I noticed that there's a new navigation library version (patch version) - maybe we are super lucky and they also did some performance improvements? :)

app/src/main/java/org/mozilla/fenix/perf/NavGraphProvider.kt
```diff
@@ -30,12 +31,15 @@ object NavGraphProvider {
     // We want to store member state on the NavController. However, there is no way to do this.
     // Instead, we store state as part of a map: NavController instance -> State. In order to
     // garbage collect our data when the NavController is no longer relevant, we use a WeakHashMap.
-    private val map = WeakHashMap<NavController, Job>()
+    private val map = WeakHashMap<NavController, Deferred<NavGraph>>()
+    private var isGraphSet = false
```
mcomella, line 35: Assuming we need some flag to indicate the graph is set (see below for alternatives), **I don't think this is correct:** we need to have an `isGraphSet` variable for every HomeActivity/`NavController`, i.e. this needs to go in the map. I think the signature of the map should be:
```kotlin
private val map = WeakHashMap<NavController, Pair<Deferred<NavGraph>, Boolean>>()
```

Or alternatively a (data) class that represents `Pair<Deferred<NavGraph>, Boolean`.

I think you should be able to reproduce the bug by starting the app, closing it out by hitting the back button (i.e. closing HomeActivity) and relaunching the app.

app/src/main/java/org/mozilla/fenix/perf/NavGraphProvider.kt
```diff
@@ -54,6 +58,11 @@ object NavGraphProvider {
             "`NavGraphProvider.inflateNavGraphAsync` to be called before this method with the same " +
             "`NavController` instance. If this occurred in a test, you probably need to add the " +
             "DisableNavGraphProviderAssertionRule.")
-        runBlockingIncrement { inflationJob.join() }
+        runBlockingIncrement {
+            if (!isGraphSet) {
```
mcomella, line 62: Rather than creating an additional boolean flag, can we query the `navController` for its state? e.g.
```kotlin
if (navController.graph == null) {
    // set the graph
}
```

Then again, this could be a bad idea: if Google ever changes `navController.graph` for an uninitialized graph, our code could break.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -264,6 +265,7 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
             attachOnHomeActivityOnCreate(lifecycle)
         }
 
+        NavGraphProvider.blockForNavGraphInflation(navHost.navController)
```
mcomella, line 268: (same comment) Why are these new lines necessary? In the initial PR, we added lines like these but they were unnecessary and only ended up slowing down the browser unnecessarily in some cases.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -222,6 +222,7 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
         if (!shouldStartOnHome() &&
             shouldNavigateBrowserFragmentOnCouldStart(savedInstanceState)
         ) {
+            NavGraphProvider.blockForNavGraphInflation(navHost.navController)
```
mcomella, line 225: Why are these new lines necessary? In the initial PR, we added lines like these but they were unnecessary and only ended up slowing down the browser unnecessarily in some cases.

mcomella (changes requested):
I think there is a small bug and I'm not sure about a few lines that were added.

app/src/main/java/org/mozilla/fenix/perf/NavGraphProvider.kt
MarcLeclair: So we tried that one. `NavController.graph == null` is always false since `NavController` checks for a null `graph` and if it is `null` it throws an error, which is why we went with the boolean option.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
rocketsroger: Good catch, I'll remove

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
rocketsroger: I'll remove.  Thanks

app/src/main/java/org/mozilla/fenix/perf/NavGraphProvider.kt
rocketsroger: Good catch.  I'll create a private data class so each `NavController` will have a boolean tracking if the graph is set.   Thanks,

mcomella (approved):
@pocmo Please review this despitem y review.

I requested changes earlier but we removed the changes I was requesting so it's okay for me to land now.

pocmo (approved):

Issue #19197, opened 2021-04-22
[Spike] Evaluate migrating to Data Store


jonalmeida:
## Description

In Fenix, we put things mostly into one SharedPreference file and as we add new features/code we add new prefs. This has a few disadvantages:
 - The file grows larger over time - this affects read/write for all prefs.
 - Organization - we don't have a way to isolate different feature prefs. In AC, each component uses a separate shared pref file that is namespaced. We could try a variation of that.
 - Accessing/type safety - this isn't a big problem for us with our own abstractions; but I'll list it here.

Jetpack also offers a new way to store data that we should consider called [`DataStore`](https://android-developers.googleblog.com/2020/09/prefer-storing-data-with-jetpack.html?linkId=98693079) with two variants.

## Deliverables
- A better way to split our prefs by feature.
- Evaluate if we should use DataStore; does it solve our needs? How we could migrate to it?

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-10494)

jonalmeida:
I noticed we also had two shared pref files that have some settings duplicated between them. Both are updated, but only one is read. 🤔 

I can't remember which ones they were, but I'll load up another debug build again and list them out.

(placeholder here)

jonalmeida:
> Evaluate if we should use DataStore; does it solve our needs? How we could migrate to it?

In the DataStore link above, it does have a helpers to migrate from SharedPreferences:
```kotlin
val dataStore: DataStore<Preferences> = context.createDataStore(
    name = "settings",
    migrations = listOf(SharedPreferencesMigration(context, "settings_preferences"))
)
```

This would be helpful in separating our prefs into separate files for an initial migration too! 

What we have to think about, is how do we keep both in sync as we slowly move to DataStore, before we flip the switch to it. 🤔

jonalmeida:
Closing for now - we're not actively considering the move to Datastore.

PR #19999, opened 2021-06-14
For #19197 - Evaluate migrating to Data Store

gabrielluong:
This is just a **prototype** and is purely in the investigation phase of trying to figure out we can architecture this within Fenix and try to answer some questions about how we can isolate preferences based on features and components. Naming and convention are not final either. I would want to use this investigation to formalize a RFC about how we can possibly utilize DataStore and architecture our preferences. 

As of writing, I think we still want to try to implement more different preferences to explore the many preference use cases within Fenix. 

### Pull Request checklist

1. click on Show All Checks,


jonalmeida:
It would be good to see how you plan on moving all of these settings over to the DataStore: Are you going to convert them all at once? Will this be one big PR? How will it be landed while other in-progress work hapens? How will ensure no regressions/stability?

Have you considered other alternatives like using a Store/State?

Datastore also makes performance claims: In this prototype, did you measure to see if we seeing performance improvements from this type of change? It could be that we are losing something when you zoom out (e.g. context shifting or start-up might have perf losses).

PR #20004, opened 2021-06-14
Update Kotlin to 1.5.10 (and Coroutines to 1.5.0).

pocmo:
Matching A-C upgrade: https://github.com/mozilla-mobile/android-components/pull/10453

Similar to the patch in A-C, I added `@OptIn(DelicateCoroutinesApi::class)` to all existing `GlobalScope` usages. However I think we may want to review them and see if there are better options.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -579,6 +579,10 @@ if (project.hasProperty("coverage")) {
         jacoco.includeNoLocationClasses = true
     }
 
+    jacoco {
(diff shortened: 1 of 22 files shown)


csadilek (approved):
Filed https://github.com/mozilla-mobile/fenix/issues/20380 as a follow up.

👍

pocmo:
Let's go!

Issue #19131, opened 2021-04-20
[Bug] Top sites - can't delete entries from the second page


sflorean:
[comment]: # (Please do your best to search for duplicate issues before filing a new issue so we can keep our issue board clean)
[comment]: # (Every issue should have the exact bug and steps to reproduce described in it. Please do not file feedback list tickets as it is difficult to parse them and address their individual points)
[comment]: # (Read https://github.com/mozilla-mobile/fenix#i-want-to-file-an-issue for more information)

## Steps to reproduce
1. Install Fenix and log in with a Firefox account.
2. Go to Top sites.
3. Swipe to the second page.
4. Perform a long tap on one of the pages.
5. Tap on "Delete from history".

### Expected behavior
The page is deleted.

### Actual behavior
The page is not deleted.

### Device information
* Device vendor / model and Android version: Pixel 3 (Android 11), Samsung Galaxy Note 8 (Android 9),  HTC Nexus 9(Android 7.1.1), Motorola Nexus 6 (Android 7.1.1).
* Firefox for Android version: Nightly 4/20 

https://user-images.githubusercontent.com/58466444/115370339-a4336300-a1d1-11eb-960e-dbee150db187.mp4

sflorean:
The issue is still reproducible on latest Nightly build with Pixel 3 (Android 11).

codrut-topliceanu:
Issue reproducible only when trying to delete the same row/column position multiple times:


https://user-images.githubusercontent.com/60002907/121507070-faac6900-c9ec-11eb-9a67-0c0062522e13.mp4

sflorean:
Verified as fixed on Nightly 7/9 with Samsung Note 10 (Android 11).

Issue #15180, opened 2020-09-18
[Spike] Investigate how to remove the remaining flash in top sites


jonalmeida:
## Title
How to remove the short flash in all top sites when removing one.

## Description
From the discussion in https://github.com/mozilla-mobile/fenix/pull/15126, we should investigate into generating stable IDs that can be used across the two data sources (pinned and frecent), and then use that for diff-ing the layout for `DiffCallback.areItemsTheSame`.

## Deliverables
Description of deliverables, including:
Findings - Are the stable IDs enough to solve the problem?
Next Steps - If yes, fix it! Otherwise, create a new issue to look into reducing the complexity of the top sites nested adapters by writing a custom ViewPager or LayoutManager.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-7070)

jonalmeida:
Moved to bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1811923

Change performed by the [Move to Bugzilla add-on](https://addons.mozilla.org/en-US/firefox/addon/move-to-bugzilla/).

PR #20005, opened 2021-06-14
For #19131 - Fixes top sites refresh and interactability

codrut-topliceanu:
For https://github.com/mozilla-mobile/fenix/issues/19131

### Pull Request checklist

- [x] **Tests**: This PR includes no new tests as it is only a bug fix of previous behavior.
- [x] **Screenshots**: This PR includes screenshots or GIFs.

1. click on Show All Checks,

Video of the fix:

https://user-images.githubusercontent.com/60002907/121899989-99a6cd00-cd2d-11eb-8265-be4f0fce0f9e.mp4

### Notes: 
- The issue was caused by a combination of several factors including the behavior change of using `offscreenPageLimit = 1` in `TopSitePagerViewHolder`(which is needed to keep the height of the topSites accurate)
- There fix reveals a new issue, however: when the items for the topSites list are fewer than 16, the efficient `getChangePayload` isn't triggered in `SessionControlAdapter` and thus all the topSites are refreshed. I will log a separate issue for this and link it here. I think this should be fixed sooner rather than later as it seems very jarring. Below you can see the problem in action:

https://user-images.githubusercontent.com/60002907/121891974-d4583780-cd24-11eb-8d42-629271f9704f.mp4

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/topsites/TopSitesPagerAdapter.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/topsites/TopSitesPagerAdapter.kt
@@ -10,7 +10,7 @@ import androidx.recyclerview.widget.DiffUtil
 import androidx.recyclerview.widget.ListAdapter
 import kotlinx.android.synthetic.main.component_top_sites.view.*
 import mozilla.components.feature.top.sites.TopSite
-import org.mozilla.fenix.home.sessioncontrol.AdapterItem
+import org.mozilla.fenix.home.sessioncontrol.AdapterItem.TopSitePagerPayload
 import org.mozilla.fenix.home.sessioncontrol.TopSiteInteractor
 import org.mozilla.fenix.home.sessioncontrol.viewholders.TopSitePagerViewHolder.Companion.TOP_SITES_PER_PAGE
 import org.mozilla.fenix.home.sessioncontrol.viewholders.TopSiteViewHolder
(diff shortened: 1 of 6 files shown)


codrut-topliceanu:
It seems that there are a few issue on the topic stated in the original comment already: 
https://github.com/mozilla-mobile/fenix/issues/15499
https://github.com/mozilla-mobile/fenix/issues/15180
Additional info here https://github.com/mozilla-mobile/fenix/pull/15126

@jonalmeida Any chance we can get some of those issues in the Ready for Engineering board? 😅

jonalmeida:
🚗 Drive-by comment: I don't have strong opinions on this change (anything is better than what we have right now), but what's a good way to write a test so that we don't regress this behaviour?

gabrielluong:
Sorry, will get to this tomorrow

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/topsites/TopSitesPagerAdapter.kt
```diff
@@ -33,15 +33,26 @@ class TopSitesPagerAdapter(
...
+        }
+    }
+
+    private fun update(
+        payload: TopSitePagerPayload,
+        position: Int,
+        adapter: TopSitesAdapter
+    ) {
+        // Only currently selected page items need to be updated.
+        for (item in payload.changed) {
```
gabrielluong, line 51: Is it possible to destructure item so that we can easily tell that `first` is the page and `second` is the top sites?

gabrielluong (approved):

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/topsites/TopSitesPagerAdapter.kt
codrut-topliceanu: I'll add a comment explaining this in a kdoc for the method. 👍

codrut-topliceanu:
Apologies for not squashing first. 🤦

PR #20013, opened 2021-06-14
Issue FNX-22435: Introduce History metadata

csadilek:
This introduces history metadata behind a feature flag. Only active in debug builds for local development...until at least the next Nightly cycle.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
@@ -206,7 +206,18 @@ class BrowserFragment : BaseBrowserFragment(), UserInteractionHandler {
     override fun onStop() {
         super.onStop()
         updateLastBrowseActivity()
+        if (FeatureFlags.historyMetadataFeature) {
+            updateHistoryMetadata()
+        }
         pwaOnboardingObserver?.stop()
+    }
+
+    private fun updateHistoryMetadata() {
+        getCurrentTab()?.let { tab ->
+            (tab as? TabSessionState)?.historyMetadata?.let {
+                requireComponents.core.historyMetadataService.updateMetadata(it, tab)
+            }
+        }
     }
 
     private fun subscribeToTabCollections() {
--- a/app/src/main/java/org/mozilla/fenix/historymetadata/HistoryMetadataMiddleware.kt
+++ b/app/src/main/java/org/mozilla/fenix/historymetadata/HistoryMetadataMiddleware.kt
@@ -1,0 +1,128 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.historymetadata
+
+import mozilla.components.browser.state.action.BrowserAction
+import mozilla.components.browser.state.action.ContentAction
+import mozilla.components.browser.state.action.HistoryMetadataAction
+import mozilla.components.browser.state.action.MediaSessionAction
+import mozilla.components.browser.state.action.TabListAction
+import mozilla.components.browser.state.selector.findTab
+import mozilla.components.browser.state.selector.normalTabs
+import mozilla.components.browser.state.selector.selectedTab
+import mozilla.components.browser.state.state.BrowserState
+import mozilla.components.browser.state.state.TabSessionState
+import mozilla.components.lib.state.Middleware
+import mozilla.components.lib.state.MiddlewareContext
+import mozilla.components.lib.state.Store
+
+/**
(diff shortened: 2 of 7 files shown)


app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
```diff
@@ -206,9 +206,22 @@ class BrowserFragment : BaseBrowserFragment(), UserInteractionHandler {
...
         updateLastBrowseActivity()
+        if (FeatureFlags.historyMetadataFeature) {
+            updateHistoryMetadata()
+        }
         pwaOnboardingObserver?.stop()
     }
 
+    private fun updateHistoryMetadata() {
+        getCurrentTab()?.let { tab ->
+            if (tab is TabSessionState) {
```
grigoryk, line 217: nit: could be `as?` to avoid extra nesting

app/src/main/java/org/mozilla/fenix/historymetadata/HistoryMetadataMiddleware.kt
```diff
@@ -0,0 +1,110 @@
...
+
+    @Suppress("ComplexMethod", "NestedBlockDepth")
+    override fun invoke(
+        context: MiddlewareContext<BrowserState, BrowserAction>,
+        next: (BrowserAction) -> Unit,
+        action: BrowserAction
+    ) {
+        // Pre process actions
+        when (action) {
+            is TabListAction.AddTabAction -> {
```
grigoryk, line 35: does this need a `private` check? I don't see an action for adding private tabs..

grigoryk: same question for `SelectTabAction` and `RemoveTabAction` - i don't see their private equivalent. If this _is_ ok, can you add a comment explaining why?

grigoryk (approved):

app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
csadilek: 👍 updated.

grigoryk:
🚢

app/src/main/java/org/mozilla/fenix/historymetadata/HistoryMetadataMiddleware.kt
```diff
@@ -0,0 +1,128 @@
...
+            is MediaSessionAction.UpdateMediaMetadataAction -> {
+                context.state.findNormalTab(action.tabId)?.let { tab ->
+                    createHistoryMetadata(context, tab)
+                }
+            }
+        }
+    }
+
+    private fun createHistoryMetadata(context: MiddlewareContext<BrowserState, BrowserAction>, tab: TabSessionState) {
+        val key = historyMetadataService.createMetadata(tab, tab.getParent(context.store))
```
gabrielluong, line 96: Should this be better named `createMetadataKey`?

app/src/main/java/org/mozilla/fenix/historymetadata/HistoryMetadataMiddleware.kt
csadilek: Good catch! Added a bunch new tests and handled in middleware.

app/src/main/java/org/mozilla/fenix/historymetadata/HistoryMetadataMiddleware.kt
csadilek: @gabrielluong Ah no, so this actually stores/creates a metadata (not just the key). It records a document type observation and returns the key. The key is the set of attributes that uniquely identify the metadata of this tab. Naming is not ideal yet but `createMetadata` is the best we could come up with for now.

gabrielluong (approved):

app/src/main/java/org/mozilla/fenix/historymetadata/HistoryMetadataMiddleware.kt
```diff
@@ -0,0 +1,128 @@
...
+}
+
+/**
+ * Finds and returns the normal (non-private) tab with the given id. Returns null if no
+ * matching tab could be found.
+ *
+ * @param tabId The ID of the tab to search for.
+ * @return The [TabSessionState] with the provided [tabId] or null if it could not be found.
+ */
+private fun BrowserState.findNormalTab(tabId: String): TabSessionState? {
```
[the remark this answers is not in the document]
csadilek: Yeah, keen to move them somewhere more generic if they're useful e.g. all the way to A-C even?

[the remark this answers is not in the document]
csadilek: Will land and we can do this as a follow up.

PR #20038, opened 2021-06-15
19886

gabrielluong:
### Pull Request checklist

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
```diff
@@ -239,33 +238,21 @@ class BrowserFragment : BaseBrowserFragment(), UserInteractionHandler {
...
+                    title = tab.content.title,
+                    isSecured = tab.content.securityInfo.secure,
+                    sitePermissions = sitePermissions,
+                    gravity = getAppropriateLayoutGravity(),
+                    certificateName = tab.content.securityInfo.issuer,
+                    permissionHighlights = tab.content.permissionHighlights,
+                    isTrackingProtectionEnabled = isTrackingProtectionEnabled
                 )
-            navController.navigateSafe(R.id.browserFragment, directions)
+            nav(R.id.browserFragment, directions)
```
Amejia481, line 255: Is there any particular reason why we are replacing `navigateSafe()` by `nav()`.

gabrielluong: Oh, I can see why the diff looks weird. I actually deleted `navToTrackingProtectionPanel` and was working in `navToQuickSettingsSheet` which uses `nav`. Otherwise, `navigateSafe` really does the same thing as `nav` if you look into much more deeper. We really should clean this all up.

Amejia481: Thanks for the clarification, maybe we can open a ticket to clear this later

gabrielluong: Filed https://github.com/mozilla-mobile/fenix/issues/20039

app/src/main/java/org/mozilla/fenix/trackingprotection/TrackingProtectionOverlay.kt
```diff
@@ -148,7 +148,7 @@ class TrackingProtectionOverlay(
 
         val toolbar = getToolbar()
         val trackingProtectionIcon: View =
-            toolbar.findViewById(R.id.mozac_browser_toolbar_tracking_protection_indicator)
+            toolbar.findViewById(R.id.mozac_browser_toolbar_security_indicator)
```
Amejia481, line 151: I think we were using this id to show the tracking protection onbording, I'm not sure if we are going to need it any more, if we are going to also remove the TP icon animations? 

image

Amejia481: All the following lines are configuring the dialog.

gabrielluong: Yes, I switched it over from the tracking protection indicator to the security indicator since that was going away. This is another one of those UX questions if we will keep this.

Amejia481 (approved):

app/src/main/java/org/mozilla/fenix/trackingprotection/TrackingProtectionOverlay.kt
gabrielluong: But otherwise, I specify the security indicator here to make sure the window is anchored correctly for the time being even though we don't have a tracking protection icon which the popup mentions.

Amejia481: 👍🏽

PR #20056, opened 2021-06-16
Closes #19937 close tab on select gridView

### LINKED ISSUE #19937: Can accidentally close a tab instead of selecting it
[comment]: # (Please do your best to search for duplicate issues before filing a new issue so we can keep our issue board clean)
[comment]: # (Every issue should have the exact bug and steps to reproduce described in it. Please do not file feedback list tickets as it is difficult to parse them and address their individual points)
[comment]: # (Read https://github.com/mozilla-mobile/fenix#i-want-to-file-an-issue for more information)

## Steps to reproduce
1. Open a website.
2. Open the tabs tray, make sure the tabs layout is grid view.
3. Open the overflow menu, click Select tabs.
4. Tap the area around where the close tab button would be (not necessarily exactly on it).
### Expected behavior
The tab should be selected, no matter where you tap.
### Actual behavior
The tab closes if you accidentally tap it around where the close button would be.
### Device information

* Device vendor / model and Android version: Pixel 2 (Android 9)
* Firefox for Android version: Nightly 6/10

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-10982)

BBBmau:
Thanks for the clarification @Mugurell !

I went back to do some debugging on both the layout file and the viewHolder file. From my experience changing "closeView.isInvisible" to "closeView.isGone" wasn't enough to solve the bug.

`        closeView.isGone =  trayStore.state.mode is TabsTrayState.Mode.Select
`

I did make the change to closeView.isGone and looked deeper into the onBindViewHolder and added the condition to prevent the onClickListener to be active when isGone is true. It looks like it solved the problem! :)

firefox_tabs_gif

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/tabstray/browser/BrowserTabsAdapter.kt
+++ b/app/src/main/java/org/mozilla/fenix/tabstray/browser/BrowserTabsAdapter.kt
@@ -71,7 +71,9 @@ class BrowserTabsAdapter(
 
         holder.tab?.let { tab ->
             holder.itemView.mozac_browser_tabstray_close.setOnClickListener {
-                interactor.close(tab)
+                if(!holder.itemView.mozac_browser_tabstray_close.isGone) {
+                    interactor.close(tab)
+                }
             }
 
             selectionHolder?.let {
--- a/app/src/main/java/org/mozilla/fenix/tabstray/browser/AbstractBrowserTabViewHolder.kt
+++ b/app/src/main/java/org/mozilla/fenix/tabstray/browser/AbstractBrowserTabViewHolder.kt
@@ -102,6 +102,7 @@ abstract class AbstractBrowserTabViewHolder(
         }
     }
 
+    // Function for selecting multiple Tabs at Once.
     fun showTabIsMultiSelectEnabled(isSelected: Boolean) {
         itemView.selected_mask.isVisible = isSelected
         closeView.isGone = trayStore.state.mode is TabsTrayState.Mode.Select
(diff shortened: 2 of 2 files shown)


app/src/main/java/org/mozilla/fenix/tabstray/browser/BrowserTabsAdapter.kt
```diff
@@ -71,7 +71,9 @@ class BrowserTabsAdapter(
 
         holder.tab?.let { tab ->
             holder.itemView.mozac_browser_tabstray_close.setOnClickListener {
-                interactor.close(tab)
+                if(!holder.itemView.mozac_browser_tabstray_close.isGone) {
```
jonalmeida, line 74: Do we need this extra check if the view is already gone?

BBBmau: Yes, when it came to going from "isInvisible" to "isGone" in showTabIsMultiSelectEnabled, I still found that the bug was still there even though it was considered GONE. This extra check was added in order to make sure that the click button was not active when it's set to GONE.

Mugurell: Then I'd say the problem is somewhere else and this issue just got a bit more tricky.
To note:
- there is no issue for the close button of the list item. Same kotlin code that changes visibility, different xml structure.
- just for testing, if we modify `showTabIsMultiSelectEnabled` to be
```
    fun showTabIsMultiSelectEnabled(isSelected: Boolean) {
        itemView.selected_mask.isVisible = isSelected
        closeView.isGone = true
    }
```
the issue doesn't reproduce even for the grid item.
So `View.GONE` should work but needs more investigation as to why it currently doesn't.
And why just for the grid item.
There could be more issues because of that same underlying cause or that same problem could exist in other parts of the app.

BBBmau: good point, I think in my previous tests I did check to see if `closeView.isGone = true` would result in the view disappearing and it did. 

Is their possibly a starting point that we could pin point where the problem could occur? It's interesting how list side is practically the same except for XML. I'll look into it more until then!

PR #20143, opened 2021-06-23
[SDK-279] Adds a way to switch to the "preview" collection in Nimbus

travis79:
This adds a toggle/switch to the debug settings menu that causes Nimbus to use the "preview" collection rather than the "main" collection when it initializes.  Since it is only applied during init, it does require a restart in order to take effect.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -1764,5 +1764,6 @@
 
     <!-- Content description for privacy content close button -->
     <string name="privacy_content_close_button_content_description">Close</string>
+    <string name="mozac_service_nimbus_use_preview_collection">Use Nimbus Preview Collection (requires restart)</string>
 
 </resources>
--- a/app/src/main/java/org/mozilla/fenix/experiments/NimbusSetup.kt
+++ b/app/src/main/java/org/mozilla/fenix/experiments/NimbusSetup.kt
@@ -31,7 +31,11 @@ fun createNimbus(context: Context, url: String?): NimbusApi {
         // Eventually we'll want to use `NimbusDisabled` when we have no NIMBUS_ENDPOINT.
         // but we keep this here to not mix feature flags and how we configure Nimbus.
         val serverSettings = if (!url.isNullOrBlank()) {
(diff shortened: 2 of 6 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -1764,5 +1764,6 @@
 
     <!-- Content description for privacy content close button -->
     <string name="privacy_content_close_button_content_description">Close</string>
+    <string name="mozac_service_nimbus_use_preview_collection">Use Nimbus Preview Collection (requires restart)</string>
```
gabrielluong, line 1767: Can you move this to `static_strings.xml`? We typically keep our secret settings menu strings there since we don't expect that it needs to be localized since it's internal only. Another note is that we typically reserve the "mozac" prefix for strings/xml coming from AC. I think we will want to follow the other preference naming convention here. See "preferences_debug_settings_enable_address_feature" for an example for both the static_strings and preference naming.

gabrielluong (approved):

app/src/main/res/values/strings.xml
travis79: > Can you move this to `static_strings.xml`? We typically keep our secret settings menu strings there since we don't expect that it needs to be localized since it's internal only. Another note is that we typically reserve the "mozac" prefix for strings/xml coming from AC. I think we will want to follow the other preference naming convention here. See "preferences_debug_settings_enable_address_feature" for an example for both the static_strings and preference naming.

Sure thing!  Thanks for the info on the naming conventions.

jonalmeida:
Landing since there doesn't seem to be any reason to not do it.

gabrielluong:
> Landing since there doesn't seem to be any reason to not do it while the PR is still fresh.

This shouldn't have landed since the strings were never moved to `static_strings.xml` and the secret menu strings weren't expected to be localized https://github.com/mozilla-l10n/android-l10n/pull/357.

jonalmeida:
Sorry, I misread the comments thread! 

Let's fix this in https://github.com/mozilla-mobile/fenix/pull/20256 so that we can fix the l10n automation.

PR #20116, opened 2021-06-21
For #19804: Changed `var isDefaultBrowser` to a function

MarcLeclair:
Changes to `isDefaultBrowser` to be a function call instead of a variable. 

with 100 runs before the changes: 
profile(This is an art profile since we can't see the changes with the GV profiler): https://share.firefox.dev/2SIDgdr
```
{'max': 1859.0,
 'mean': 1812.33,
 'median': 1812.0,
 'min': 1780.0,
 'replicate_count': 100,
 'replicates': [1799.0, 1813.0, 1816.0, 1824.0, 1834.0, 1806.0, 1832.0, 1801.0,
                1786.0, 1818.0, 1815.0, 1796.0, 1815.0, 1836.0, 1821.0, 1812.0,
                1810.0, 1804.0, 1790.0, 1799.0, 1859.0, 1814.0, 1822.0, 1827.0,
                1802.0, 1798.0, 1822.0, 1837.0, 1827.0, 1815.0, 1801.0, 1803.0,
                1799.0, 1780.0, 1827.0, 1798.0, 1810.0, 1824.0, 1800.0, 1801.0,
                1811.0, 1805.0, 1782.0, 1810.0, 1825.0, 1822.0, 1803.0, 1812.0,
                1815.0, 1827.0, 1831.0, 1822.0, 1799.0, 1796.0, 1815.0, 1791.0,
                1798.0, 1824.0, 1801.0, 1836.0, 1786.0, 1807.0, 1826.0, 1803.0,
                1794.0, 1821.0, 1803.0, 1801.0, 1806.0, 1830.0, 1827.0, 1802.0,
                1838.0, 1821.0, 1806.0, 1794.0, 1834.0, 1832.0, 1785.0, 1809.0,
                1823.0, 1829.0, 1809.0, 1794.0, 1787.0, 1807.0, 1810.0, 1818.0,
                1827.0, 1833.0, 1840.0, 1808.0, 1819.0, 1814.0, 1816.0, 1822.0,
                1831.0, 1785.0, 1798.0, 1820.0]}
```

and 100 runs with the changes: 
profile: (this also an ART profiles since this happens before the GV profiler can start) https://share.firefox.dev/3d2n8dI
```
{'max': 1836.0,
 'mean': 1792.24,
 'median': 1791.0,
 'min': 1756.0,
 'replicate_count': 100,
 'replicates': [1793.0, 1791.0, 1756.0, 1791.0, 1770.0, 1798.0, 1775.0, 1780.0,
[...]

PR #20147, opened 2021-06-23
Closes #19846: Show Default Browser Notification if browser is not default

### LINKED ISSUE #19846: Show 'Default Browser' Local Notification for New users
Built as a feature of the app, (much like the onboarding flow) that gets scheduled in a user's first session and triggers if they do not set us as default 24 hours after 1st app launch, see an example below.

image

rocketsroger:
https://user-images.githubusercontent.com/6579541/123160851-a9fd2d00-d43c-11eb-8074-56355dfb1545.mp4

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
+++ b/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
@@ -86,6 +86,7 @@ import org.mozilla.fenix.ext.setNavigationIcon
 import org.mozilla.fenix.ext.settings
 import org.mozilla.fenix.home.HomeFragmentDirections
 import org.mozilla.fenix.home.intent.CrashReporterIntentProcessor
+import org.mozilla.fenix.home.intent.DefaultBrowserIntentProcessor
 import org.mozilla.fenix.home.intent.OpenBrowserIntentProcessor
 import org.mozilla.fenix.home.intent.OpenSpecificTabIntentProcessor
 import org.mozilla.fenix.home.intent.SpeechProcessingIntentProcessor
@@ -94,6 +95,7 @@ import org.mozilla.fenix.library.bookmarks.BookmarkFragmentDirections
 import org.mozilla.fenix.library.bookmarks.DesktopFolders
 import org.mozilla.fenix.library.history.HistoryFragmentDirections
 import org.mozilla.fenix.library.recentlyclosed.RecentlyClosedFragmentDirections
+import org.mozilla.fenix.onboarding.DefaultBrowserNotificationWorker
 import org.mozilla.fenix.perf.Performance
 import org.mozilla.fenix.perf.PerformanceInflater
 import org.mozilla.fenix.perf.ProfilerMarkers
@@ -158,7 +160,8 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
             SpeechProcessingIntentProcessor(this, components.core.store, components.analytics.metrics),
             StartSearchIntentProcessor(components.analytics.metrics),
             OpenBrowserIntentProcessor(this, ::getIntentSessionId),
-            OpenSpecificTabIntentProcessor(this)
+            OpenSpecificTabIntentProcessor(this),
+            DefaultBrowserIntentProcessor(this)
         )
     }
 
@@ -337,6 +340,8 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
             }
 
             settings().wasDefaultBrowserOnLastResume = settings().isDefaultBrowser()
+
+            DefaultBrowserNotificationWorker.setDefaultBrowserNotificationIfNeeded(applicationContext)
         }
     }
 
--- a/app/src/main/java/org/mozilla/fenix/home/intent/DefaultBrowserIntentProcessor.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/intent/DefaultBrowserIntentProcessor.kt
@@ -1,0 +1,34 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.home.intent
+
+import android.content.Intent
+import androidx.navigation.NavController
+import org.mozilla.fenix.HomeActivity
+import org.mozilla.fenix.ext.openSetDefaultBrowserOption
+import org.mozilla.fenix.ext.settings
+import org.mozilla.fenix.onboarding.DefaultBrowserNotificationWorker.Companion.isDefaultBrowserNotificationIntent
+
+/**
+ * When the default browser notification is tapped we need to launch [openSetDefaultBrowserOption]
+ *
+ * This should only happens once in a user's lifetime since once the user taps on the default browser
+ * notification, [settings.shouldShowDefaultBrowserNotification] will return false
+ */
+class DefaultBrowserIntentProcessor(
+    private val activity: HomeActivity
+) : HomeIntentProcessor {
+
+    override fun process(intent: Intent, navController: NavController, out: Intent): Boolean {
+        return if (isDefaultBrowserNotificationIntent(intent)) {
+            activity.openSetDefaultBrowserOption()
+            activity.settings().defaultBrowserNotificationDisplayed = true
+
+            true
+        } else {
+            false
+        }
+    }
+}
--- a/app/src/main/java/org/mozilla/fenix/onboarding/DefaultBrowserNotificationWorker.kt
+++ b/app/src/main/java/org/mozilla/fenix/onboarding/DefaultBrowserNotificationWorker.kt
@@ -1,0 +1,126 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.onboarding
+
+import android.app.Notification
+import android.app.NotificationChannel
+import android.app.NotificationManager
+import android.app.PendingIntent
+import android.content.Context
+import android.content.Intent
+import android.os.Build
+import androidx.core.app.NotificationCompat
+import androidx.core.app.NotificationManagerCompat
+import androidx.core.content.ContextCompat
+import androidx.work.ExistingWorkPolicy
+import androidx.work.OneTimeWorkRequest
+import androidx.work.WorkManager
+import androidx.work.Worker
+import androidx.work.WorkerParameters
+import java.util.concurrent.TimeUnit
+import mozilla.components.support.base.ids.SharedIdsHelper
+import org.mozilla.fenix.HomeActivity
+import org.mozilla.fenix.R
+import org.mozilla.fenix.ext.settings
+import org.mozilla.fenix.utils.Settings
+
+class DefaultBrowserNotificationWorker(
+    context: Context,
+    workerParameters: WorkerParameters
+) : Worker(context, workerParameters) {
+
+    override fun doWork(): Result {
+        ensureChannelExists()
+        NotificationManagerCompat.from(applicationContext)
+            .notify(NOTIFICATION_TAG, NOTIFICATION_ID, buildNotification())
+        return Result.success()
+    }
+
+    /**
+     * Build the default browser notification.
+     */
+    private fun buildNotification(): Notification {
+        val channelId = ensureChannelExists()
+        val intent = Intent(applicationContext, HomeActivity::class.java)
+        intent.putExtra(INTENT_DEFAULT_BROWSER_NOTIFICATION, true)
+
+        val pendingIntent = PendingIntent.getActivity(
+            applicationContext,
+            SharedIdsHelper.getNextIdForTag(applicationContext, NOTIFICATION_PENDING_INTENT_TAG),
+            intent,
+            0
+        )
+
+        with(applicationContext) {
+            val appName = getString(R.string.app_name)
+            return NotificationCompat.Builder(this, channelId)
+                .setSmallIcon(R.drawable.ic_status_logo)
+                .setContentTitle(
+                    applicationContext.getString(R.string.notification_default_browser_text, appName))
+                .setBadgeIconType(NotificationCompat.BADGE_ICON_SMALL)
+                .setColor(ContextCompat.getColor(this, R.color.primary_text_light_theme))
+                .setPriority(NotificationCompat.PRIORITY_DEFAULT)
+                .setShowWhen(false)
+                .setContentIntent(pendingIntent)
+                .setAutoCancel(true)
+                .build()
+        }
+    }
+
+    /**
+     * Make sure a notification channel for default browser notification exists.
+     *
+     * Returns the channel id to be used for notifications.
+     */
+    private fun ensureChannelExists(): String {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            val notificationManager: NotificationManager =
+                applicationContext.getSystemService(Context.NOTIFICATION_SERVICE) as NotificationManager
+
+            val channel = NotificationChannel(
+                NOTIFICATION_CHANNEL_ID,
+                applicationContext.getString(R.string.notification_marketing_channel_name),
+                NotificationManager.IMPORTANCE_DEFAULT
+            )
+
+            notificationManager.createNotificationChannel(channel)
+        }
+
+        return NOTIFICATION_CHANNEL_ID
+    }
+
+    companion object {
+        private const val NOTIFICATION_CHANNEL_ID = "org.mozilla.fenix.default.browser.channel"
+        private const val NOTIFICATION_ID = 1
+        private const val NOTIFICATION_PENDING_INTENT_TAG = "org.mozilla.fenix.default.browser"
+        private const val INTENT_DEFAULT_BROWSER_NOTIFICATION = "org.mozilla.fenix.default.browser.intent"
+        private const val NOTIFICATION_TAG = "org.mozilla.fenix.default.browser.tag"
+        private const val NOTIFICATION_WORK_NAME = "org.mozilla.fenix.default.browser.work"
+        private const val NOTIFICATION_DELAY = Settings.ONE_DAY_MS
+
+        fun isDefaultBrowserNotificationIntent(intent: Intent) =
+            intent.extras?.containsKey(INTENT_DEFAULT_BROWSER_NOTIFICATION) ?: false
+
+        fun setDefaultBrowserNotificationIfNeeded(context: Context) {
+            val instanceWorkManager = WorkManager.getInstance(context)
+
+            if (!context.settings().shouldShowDefaultBrowserNotification()) {
+                // cancel notification work if already default browser
+                instanceWorkManager.cancelUniqueWork(NOTIFICATION_WORK_NAME)
+                return
+            }
+
+            val notificationWork = OneTimeWorkRequest.Builder(DefaultBrowserNotificationWorker::class.java)
+                .setInitialDelay(NOTIFICATION_DELAY, TimeUnit.MILLISECONDS)
+                .build()
+
+            instanceWorkManager.beginUniqueWork(
+                NOTIFICATION_WORK_NAME,
+                ExistingWorkPolicy.REPLACE,
+                notificationWork
+            ).enqueue()
+        }
+    }
+}
--- a/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
+++ b/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
@@ -484,6 +484,15 @@ class Settings(private val appContext: Context) : PreferencesHolder {
         return browsers.isDefaultBrowser
     }
 
+    var defaultBrowserNotificationDisplayed by booleanPreference(
+        appContext.getPreferenceKey(R.string.pref_key_should_show_default_browser_notification),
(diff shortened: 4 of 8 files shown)


rocketsroger:
@topotropic can you please confirm if the notification text is correct.  Currently it is "Make <app name> your default browser"

topotropic:
This looks good! Thanks

topotropic (approved):

app/src/main/java/org/mozilla/fenix/onboarding/DefaultBrowserNotificationWorker.kt
```diff
@@ -0,0 +1,103 @@
...
+
+    override fun doWork(): Result {
+        ensureChannelExists()
+        NotificationManagerCompat.from(applicationContext)
+            .notify(NOTIFICATION_TAG, NOTIFICATION_ID, buildNotification())
+        return Result.success()
+    }
+
+    /**
+     * Build the notification with site controls to be displayed while the web app is active.
```
jonalmeida, line 39: What does this have to do with web apps? 🤔

app/src/main/java/org/mozilla/fenix/home/intent/DefaultBrowserIntentProcessor.kt
```diff
@@ -0,0 +1,28 @@
...
+import org.mozilla.fenix.onboarding.DefaultBrowserNotificationWorker.Companion.isDefaultBrowserNotificationIntent
+
+class DefaultBrowserIntentProcessor(
+    private val activity: HomeActivity
+) : HomeIntentProcessor {
+
+    override fun process(intent: Intent, navController: NavController, out: Intent): Boolean {
+        return if (isDefaultBrowserNotificationIntent(intent)) {
+            activity.openSetDefaultBrowserOption()
+            activity.settings().shownDefaultBrowserNotification = true
```
jonalmeida, line 21: hand-wavy nit: I was misreading "shown" as "show" so the logic was a bit confusing at first. We have some examples of "should" but that's the opposite logic that you have here. Maybe a good past tense word could be "defaultBrowserNotificationDisplayed"/"showedDefaultBrowserNotification'?

Again, this is a pretty handy-wavy nit, so by all means, ignore it if you wish.

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -479,6 +479,15 @@ class Settings(private val appContext: Context) : PreferencesHolder {
         return browsers.isDefaultBrowser
     }
 
+    var shownDefaultBrowserNotification by booleanPreference(
+        appContext.getPreferenceKey(R.string.pref_key_should_show_default_browser_notification),
+        default = false
+    )
+
+    fun shouldShowDefaultBrowserNotification(): Boolean {
```
jonalmeida, line 487: nit: Maybe move this to where ever it's only use case is. It doesn't seem like this function needs to be globally accessible from all of settings. It's a strange pattern that seems to exist in this class for some reason. 🤷

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -338,6 +345,8 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
 
             settings().wasDefaultBrowserOnLastResume = settings().isDefaultBrowser()
         }
+
+        setDefaultBrowserNotificationIfNeeded()
```
jonalmeida, line 349: This touches IO via SharedPreferences for every resume. We may want to take some care around adding this.

cc: @mcomella

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -158,7 +164,8 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
             SpeechProcessingIntentProcessor(this, components.core.store, components.analytics.metrics),
             StartSearchIntentProcessor(components.analytics.metrics),
             OpenBrowserIntentProcessor(this, ::getIntentSessionId),
-            OpenSpecificTabIntentProcessor(this)
+            OpenSpecificTabIntentProcessor(this),
+            DefaultBrowserIntentProcessor(this)
```
jonalmeida, line 168: Looking at the implementation of `DefaultBrowserIntentProcessor`, I don't see this touching IO but our intent processors increase, so our VIEW startup performance might be affected.

cc: @mcomella

app/src/main/java/org/mozilla/fenix/onboarding/DefaultBrowserNotificationWorker.kt
```diff
@@ -0,0 +1,103 @@
...
+                .setColor(ContextCompat.getColor(this, R.color.primary_text_light_theme))
+                .setPriority(NotificationCompat.PRIORITY_LOW)
+                .setShowWhen(false)
+                .setContentIntent(pendingIntent)
+                .build()
+        }
+    }
+
+    /**
+     * Make sure a notification channel for the powered by notifications exists.
```
jonalmeida, line 69: Docs from a different place?

jonalmeida: Seems like this class is coming from `PoweredByNotification`. Is there anything around here that would be good to abstract out for re-use?

app/src/main/java/org/mozilla/fenix/onboarding/DefaultBrowserNotificationWorker.kt
```diff
@@ -0,0 +1,103 @@
...
+     * Returns the channel id to be used for notifications.
+     */
+    private fun ensureChannelExists(): String {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            val notificationManager: NotificationManager =
+                applicationContext.getSystemService(NOTIFICATION_SERVICE) as NotificationManager
+
+            val channel = NotificationChannel(
+                NOTIFICATION_CHANNEL_ID,
+                applicationContext.getString(R.string.notification_default_browser_channel_name),
```
jonalmeida, line 80: Is it intentional to have a notification channel _only_ for the default browser notification or is there something more generic if/when we want to have more that fall under that grouping? I can see Brave has a generic "Brave Ads" [group ID](https://github.com/brave/brave-core/blob/1cb5818aa0b70666c6aeea5ea9c06cc4e712171a/android/java/org/chromium/chrome/browser/notifications/channels/BraveChannelDefinitions.java#L19).

jonalmeida (commented):
Left some drive-by comments. 🚗 🙂

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
mcomella: Thanks for catching this. Can we do this on a background thread? afaict, there's nothing here that requires UI thread interaction.

app/src/main/java/org/mozilla/fenix/onboarding/DefaultBrowserNotificationWorker.kt
```diff
@@ -0,0 +1,103 @@
...
+            return NotificationCompat.Builder(this, channelId)
+                .setSmallIcon(R.drawable.ic_status_logo)
+                .setContentTitle(
+                    applicationContext.getString(R.string.notification_default_browser_text, appName))
+                .setBadgeIconType(BADGE_ICON_NONE)
+                .setColor(ContextCompat.getColor(this, R.color.primary_text_light_theme))
+                .setPriority(NotificationCompat.PRIORITY_LOW)
+                .setShowWhen(false)
+                .setContentIntent(pendingIntent)
+                .build()
```
jonalmeida, line 64: Is `PRIORITY_LOW` intentional? That means, by your video as well, that the notification will never "pop-up" when it is delivered but it will hide (minimized) until the user opens up their either notification tray and find it there. That seems like the opposite desire for a notification that we want to catch the user's attention with. We might want to ask UX or marketing about that.

With `BADGE_ICON_NONE` might have similar effects of hiding the notification, but I'm not fully sure of what [the docs](https://developer.android.com/reference/android/app/Notification?hl=ja#BADGE_ICON_NONE) mean.

app/src/main/java/org/mozilla/fenix/onboarding/DefaultBrowserNotificationWorker.kt
rocketsroger: my mistake.  forgot to update the comments.

app/src/main/java/org/mozilla/fenix/home/intent/DefaultBrowserIntentProcessor.kt
rocketsroger: Sure, I'll rename to `defaultBrowserNotificationDisplayed`.  Thanks,

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
rocketsroger: Agreed.  Thanks for catching this.  I'll do this on a background thread.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
rocketsroger: The processing of the intent is fairly quick.  I don't think it will affect the VIEW startup performance unless the intent is from the default browser notification.  It checks if the intents have the `INTENT_DEFAULT_BROWSER_NOTIFICATION` extra.  if not then it returns right away.

app/src/main/java/org/mozilla/fenix/onboarding/DefaultBrowserNotificationWorker.kt
rocketsroger: Sure, I'll update the channel name to `Marketing` to allow for more generic usage in the future.  Thanks,

app/src/main/java/org/mozilla/fenix/onboarding/DefaultBrowserNotificationWorker.kt
rocketsroger: Yes, this might be a good abstraction for the support library.   Currently there's a lot of `ensureChannelExists` function calls in both Fenix and Android Components.  I'll created an issue to track this work.

rocketsroger: https://github.com/mozilla-mobile/android-components/issues/10520

app/src/main/java/org/mozilla/fenix/onboarding/DefaultBrowserNotificationWorker.kt
rocketsroger: This is a good question.  My assumption is that we don't want to make any noise with this notification.  However, this is just my assumption.  @topotropic, do we want this notification more visible to the user?  Thanks,

rocketsroger: I'll update the priority to `PRIORITY_DEFAULT` and `BADGE_ICON_SMALL` so it's more visible.  I'll upload a new recording after.  Thanks,

rocketsroger:
https://user-images.githubusercontent.com/6579541/123292759-5db7f880-d4e1-11eb-95e0-baafe2d0e361.mp4

Updated behaviour.

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
rocketsroger: Sounds good.  I'll move it thanks,

rocketsroger: Hmm, this makes testing more complicated.  I think I'll leave this where it is.  I would like to be able to test this without other dependencies.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
mcomella: In the event that we take the `true` branch of the if statement in that class, I'm also concerned this could impact start up perf:
```kotlin
        return if (isDefaultBrowserNotificationIntent(intent)) {
            activity.openSetDefaultBrowserOption()
            activity.settings().defaultBrowserNotificationDisplayed = true
```

- `openSetDefaultBrowserOption` appears to access a `SystemService`. The `SystemService` classes often rely on IPC which can be a little slow (e.g. this is why we're modifying how `isDefaultBrowser` is called in https://github.com/mozilla-mobile/fenix/pull/20116/)
- We write to settings and any access of shared prefs is has some perf impact

Perhaps you can measure how long this branch takes _on a low-end device_ like the Moto G5? e.g.
```kotlin
val before = SystemClock.elapsedRealtime()
return if (isDefaultBrowserNotificationIntent(intent)) {
            activity.openSetDefaultBrowserOption()
            activity.settings().defaultBrowserNotificationDisplayed = true
            true
...
}.also {
    val end = SystemClock.elapsedRealtime()
    Log.e("lol", "elapsed: ${end - start}")
}
```

If you don't have a low-end device, let me know and I can take the measurement if you send me a build. :)

mcomella: If the timing is 10ms+ on a Moto G5-equivalent device, we probably want to change it. Most code of this size should run in 0-1ms (i.e. 1 is a rounding situation).

rocketsroger: Yes, you're correct that in the event that the intent is coming from the default browser notification, it will have an affect on the startup time.

But this will only happen if the user taps on the notification.  Shouldn't we then perform user's action as soon as we can?

mcomella: rocketsroger on Slack:

> I’m not sure if we should worry about the impact to performance when the intent is coming from the notification.  Since it only happens once in user’s lifetime and only happens if the reason for starting Fenix was because the user tapped on the notification.

Ah, it only happens once? Yeah, probably not worth optimizing then. We should probably add a comment to that effect. Good catch!

rocketsroger: Sure, I'll add a comment  Thanks,

rocketsroger: Added comment in `DefaultBrowserIntentProcessor.kt`

app/src/main/java/org/mozilla/fenix/onboarding/DefaultBrowserNotificationWorker.kt
jonalmeida: > Sure, I'll update the channel name to `Marketing` to allow for more generic usage in the future. Thanks,

Was this string something that the content/UX team verified? This will eventually be translated into various languages and deleting/updating this if we want to use channel grouping for something else will be annoying in the future.

jonalmeida:
Let's make sure the notification configuration meets our general requirements. I'm not sure I've caught everything but some of it seems to be specific to the original "Powered by" notification which was intended to be a persistent notification to keep the browser running and notify the user that it was for PWAs, which is a different use case from this one.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -976,6 +986,26 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
...
+        val instanceWorkManager = WorkManager.getInstance(this)
+
+        if (!settings().shouldShowDefaultBrowserNotification()) {
+                // cancel notification work if already default browser
+                instanceWorkManager.cancelUniqueWork(DefaultBrowserNotificationWorker.NOTIFICATION_WORK_NAME)
+            return
+        }
+
+        val notificationWork = OneTimeWorkRequest.Builder(DefaultBrowserNotificationWorker::class.java)
+            .setInitialDelay(DefaultBrowserNotificationWorker.NOTIFICATION_DELAY, TimeUnit.MILLISECONDS)
```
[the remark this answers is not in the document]
jonalmeida: Ah, I think I can answer my own question, but you can confirm: this wouldn't happen because `defaultBrowserNotificationDisplayed` needs to be set to true and that only happens in the processor.

This was a bit confusing to follow because of the (nested) double negative. 😄

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -488,7 +497,8 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
         )
 
         val intentProcessors =
-            listOf(CrashReporterIntentProcessor()) + externalSourceIntentProcessors
+            listOf(CrashReporterIntentProcessor()) +
+                    externalSourceIntentProcessors
```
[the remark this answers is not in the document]
rocketsroger: I'll fix it.  Thanks,

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -976,6 +986,26 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
         }
     }
 
+    private fun setDefaultBrowserNotificationIfNeeded() {
+        val instanceWorkManager = WorkManager.getInstance(this)
+
+        if (!settings().shouldShowDefaultBrowserNotification()) {
+                // cancel notification work if already default browser
+                instanceWorkManager.cancelUniqueWork(DefaultBrowserNotificationWorker.NOTIFICATION_WORK_NAME)
```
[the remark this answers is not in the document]
rocketsroger: hmm yeah that's weird.  I'll fix as well.  Thanks,

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -976,6 +986,26 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
         }
     }
 
+    private fun setDefaultBrowserNotificationIfNeeded() {
```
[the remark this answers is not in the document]
rocketsroger: Sure, I'll think about his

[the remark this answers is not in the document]
rocketsroger: I'll move it into DefaultBrowserNotificationWorker

app/src/main/java/org/mozilla/fenix/onboarding/DefaultBrowserNotificationWorker.kt
```diff
@@ -0,0 +1,102 @@
...
+     * Build the default browser notification.
+     */
+    private fun buildNotification(): Notification {
+        val channelId = ensureChannelExists()
+        val intent = Intent(applicationContext, HomeActivity::class.java)
+        intent.putExtra(INTENT_DEFAULT_BROWSER_NOTIFICATION, true)
+
+        val pendingIntent = PendingIntent.getActivity(
+            applicationContext,
+            SharedIdsHelper.getNextIdForTag(applicationContext, NOTIFICATION_PENDING_INTENT_TAG),
```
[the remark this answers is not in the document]
rocketsroger: Yeah, `NOTIFICATION_PENDING_INTENT_TAG` is not really needed.  Just want to show what that value is.

app/src/main/java/org/mozilla/fenix/onboarding/DefaultBrowserNotificationWorker.kt
```diff
@@ -0,0 +1,102 @@
...
+        with(applicationContext) {
+            val appName = getString(R.string.app_name)
+            return NotificationCompat.Builder(this, channelId)
+                .setSmallIcon(R.drawable.ic_status_logo)
+                .setContentTitle(
+                    applicationContext.getString(R.string.notification_default_browser_text, appName))
+                .setBadgeIconType(NotificationCompat.BADGE_ICON_SMALL)
+                .setColor(ContextCompat.getColor(this, R.color.primary_text_light_theme))
+                .setPriority(NotificationCompat.PRIORITY_DEFAULT)
+                .setShowWhen(false)
```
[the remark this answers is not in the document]
rocketsroger: I kept it to false since we're not really showing a notification that is time sensitive.  I'm not sure which is better but personally I like the look of the notification without the timestamp.

app/src/main/java/org/mozilla/fenix/onboarding/DefaultBrowserNotificationWorker.kt
```diff
@@ -0,0 +1,102 @@
...
+                .setContentIntent(pendingIntent)
+                .setAutoCancel(true)
```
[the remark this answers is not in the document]
rocketsroger: This is something I added after testing.  Without `setAutoCancel(true)` the user will tap on the notification, set our browser to default and still see the notification is there.  Once the user tap on the notification it should be dismissed.

app/src/main/java/org/mozilla/fenix/onboarding/DefaultBrowserNotificationWorker.kt
rocketsroger: Yes, the original requirement does not have any text specified.  That's why I added @topotropic to review the text and she gave the approval above.

app/src/main/java/org/mozilla/fenix/onboarding/DefaultBrowserNotificationWorker.kt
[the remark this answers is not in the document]
jonalmeida: Sorry, I must have misread this when I reviewed it. Auto-cancel _true_ makes sense now - thanks!

jonalmeida (approved):
Looks good from eng side. ✅

PR #20161, opened 2021-06-24
No issue: Apply workaround for run-ui in contributor workflow

AaronMT:
Due to https://issuetracker.google.com/issues/191799887 the newest Android emulators are not working on `macOS-latest` (10.5) VMs in Github Actions. There is a workaround of using an old emulator. Let's see if this works, otherwise I'll disable this particular job but not the workflow.


isabelrios (approved):
LGTM

PR #20189, opened 2021-06-28
For #15499 - Smoothly updates topSites list on remove

codrut-topliceanu:
For https://github.com/mozilla-mobile/fenix/issues/15499
⚠️ **Requires https://github.com/mozilla-mobile/fenix/pull/20005** even though some of the changes in this PR overwrite those from https://github.com/mozilla-mobile/fenix/pull/20005

### Pull Request checklist

1. click on Show All Checks,

### Implementation details:
- To remove the flash on refresh of the topsites list we have to use submitList, however using this too high up in the hierarchy of our listAdapters within listAdapters will cause the children to refresh at once. The solution to this is to use submitList lower in the hierarchy. Using it in TopSitesPagerAdapter.kt to update the TopSitesAdapter is the way to go. I've also had to use a dummy item for the "removed" Topsite ( with id = -1) so I can manually diff that before using submitList.
- Had to change the `sameAs` function in `SessionControlAdapter.kt:55` to consider two `TopSitePager`s of different sizes the same in order to trigger `contentsSameAs` false, which in turn triggers `getChangePayload` where the magic happens.
- `getChangePayload` now adds a dummy empty "changed" topsite for us to diff further down in the hierarchy. I wasn't able to do without this change.
- `TopSitesPagerAdapter.kt:45 - update()` computes changes for only 1 page at a time, making sure that the indexes of the received changed items match the ones of the current page by using `item.first - (position * TOP_SITES_PER_PAGE)` where item is a `Pair<index,Topsite>`  and `position` is the current page of the TopSites.
- **Known issue:** When deleting topSites and reducing their number from 9 to 8, thus no longer needing a 2nd page, the "page indicator" still stays visible unfortunately. This happens because the code for it lies above in `TopSitePagerViewHolder.lt:61 - bind()` and I'm not sure how to trigger it from `TopSitesPagerAdapter` without a messy callback or adding unneeded observers. But I'll look into it some more.
- I'll keep this PR as a draft until https://github.com/mozilla-mobile/fenix/pull/20005 merges and tests are written.

https://user-images.githubusercontent.com/60002907/123598987-c19e3200-d7fd-11eb-8a59-c36e5f23d143.mp4


jonalmeida:
I'm so excited to see this close to landing! 

codrut-topliceanu:
@jonalmeida If you think this kinda looks okay(if you volunteer as reviewer 😆 ) I'll get started on adding tests. Still unsure on what to do about the page indicator.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/topsites/TopSitesPagerAdapter.kt
```diff
@@ -42,21 +43,43 @@ class TopSitesPagerAdapter(
...
+        currentPageChangedItems.forEach { item ->
+            if (item.second.id == -1L)
+                itemsToRemove.add(item.first - (position * TOP_SITES_PER_PAGE))
+            refreshedItems[item.first - (position * TOP_SITES_PER_PAGE)] = item.second
         }
+
+        // Delete any items marked as such and submit the list, then notify adapter of deletions
+        itemsToRemove.forEach { refreshedItems.removeAt(it) }
+        adapter.submitList(refreshedItems)
+        itemsToRemove.forEach { adapter.notifyItemRemoved(it) }
```
jonalmeida, line 70: That's interesting, these two lines are similar. They both ask the adapter to update the UI. Why do we need to do both?

jonalmeida: What's interesting about this whole method is that we're handling most of the work that DiffUtils should be doing.

app/src/test/java/org/mozilla/fenix/home/sessioncontrol/viewholders/topsites/TopSitesPagerAdapterTest.kt
```diff
@@ -4,44 +4,180 @@
...
+import io.mockk.verify
 import mozilla.components.feature.top.sites.TopSite
+import mozilla.components.feature.top.sites.TopSite.Type.DEFAULT
 import org.junit.Assert.assertEquals
+import org.junit.Before
 import org.junit.Test
-import org.mozilla.fenix.home.sessioncontrol.AdapterItem
+import org.mozilla.fenix.home.sessioncontrol.AdapterItem.TopSitePagerPayload
 
 class TopSitesPagerAdapterTest {
```
jonalmeida, line 20: These are super valuable tests and it would be nice to have one that looks at the other adapter as a blackbox to assert the changes we expect. That would really help us avoid regressions but also nail down or expectations of how the results of the adapter should behave.

Something like:

```
topSitesPagerAdapter.submitList(listOfTopSites)
val result = topSitesPagerAdapter.getPagedTopSites() // List<Int, List<TopSite>>

assert(firstTopSite, result[0].first())
assert(secondTopSite, result[1].first())
// etc..
```

The hope is that, if we change the internals, like that nested RecyclerView structure, we can rely on a test like this to verify we haven't goofed. Would be great to know what you think about it. 🙂

codrut-topliceanu:
Thanks for helping me improve this PR @jonalmeida . The code here has been through several iterations and there's bound to be some issues I've overlooked.

app/src/test/java/org/mozilla/fenix/home/sessioncontrol/viewholders/topsites/TopSitesPagerAdapterTest.kt
codrut-topliceanu: This is a very good idea, I'll try and implement it! 👍

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/topsites/TopSitesPagerAdapter.kt
codrut-topliceanu: > That's interesting, these two lines are similar. They both ask the adapter to update the UI. Why do we need to do both?

I think `notifyItemRemoved` was needed before settling on this latest way of using `submitList`. I've removed it and done several tests, and indeed, it's not needed anymore.

codrut-topliceanu:
To further illustrate the PR's implementation let’s take as an example the following list of TopSites:

0   1   2   3   4   5   6   -index
A   B  C   D   E   F  G   -title

**With current implementation of this PR:**

If we delete C we get a “changed payload” of:
_{(2,D),(3,E), (4,F), (6, removed)}_
then, in `TopSitesPagerAdapter.update()` we create a new list, update its values:
0   1   2   3   4   5   6
A   B  D  E   F   G  removed
and then delete _(6,removed)_,
and submit the new updated list. Additionally this also works if we were to remove more than one topSite at the same time (even though this isn’t actually a thing you can do at the moment).

**Without the dummy “removed” item created in `SessionControlAdapter`:**
 if we delete C we get a “changed payload” of:
_{(2,D),(3,E), (4,F)}_
then, in `TopSitesPagerAdapter.update()` we create a new list, update its values:
0   1   2   3   4   5   6
A   B  D   E   F   G   G
Now we have a problem, we need to delete the correct _“G”_. Of course it’s the last one. But if we were to delete several topSites at once it would be a bit more difficult.

Add to this the fact that we need to handle each page of the topSites separately (`TopSitePagerViewHolder`), but support handling both at the same time due to `offscreenPageLimit = 1`.

Yeah, ideally we’d want a DiffUtil to handle the changes to the data set, but that’s further up in the hierarchy and we need the DiffUtil to work at `TopSitesPagerAdapter`’s level so that we don’t get those flashes.

Now I’m sure that there must be a better way of doing all of this, but it may require we build smth different than a list within a list within a list :)

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlAdapter.kt
```diff
@@ -71,9 +71,12 @@ sealed class AdapterItem(@LayoutRes val viewType: Int) {
             val oldTopSites = (this as? TopSitePager) ?: return null
 
             val changed = mutableSetOf<Pair<Int, TopSite>>()
-            for ((index, item) in newTopSites.topSites.withIndex()) {
-                if (oldTopSites.topSites.getOrNull(index) != item) {
-                    changed.add(Pair(index, item))
+
+            for ((index, item) in oldTopSites.topSites.withIndex()) {
+                val changedItem =
+                    newTopSites.topSites.getOrNull(index) ?: TopSite(-1, "REMOVED", "", 0, FRECENT)
```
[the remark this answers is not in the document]
codrut-topliceanu: I've added a more in-depth explanation in my [comment below.](https://github.com/mozilla-mobile/fenix/pull/20189#issuecomment-877124730)

jonalmeida (approved):
Thank you for the well thought out design! This [comment](https://github.com/mozilla-mobile/fenix/pull/20189#issuecomment-877124730) in particular is SUPER valuable details and I highly recommend in-lining it into the code itself for the next person that is also confused by it. :) 

Let's land this in that case, I don't want to hold up this review any longer.

After working on this bug, you might be interested in re-writing top sites UI in Compose (soon™). :)

PR #20244, opened 2021-07-04
For #1268 Add unit test for PrivateShortcutCreateManager

### LINKED ISSUE #1268: Create unit tests for shortcuts
We did not land unit tests with the shortcuts feature, but we should add some.

Files to test: 
* https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/java/org/mozilla/fenix/components/PrivateShortcutCreateManager.kt
* ...

RotBolt:
### Pull Request checklist

Closes #1268 
PR just contains a unit test.
It does not contain any UI related change or accessibilty change

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/PrivateShortcutCreateManager.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/PrivateShortcutCreateManager.kt
@@ -7,6 +7,7 @@ package org.mozilla.fenix.components
 import android.app.PendingIntent
 import android.content.Context
 import android.content.Intent
+import android.content.IntentSender
 import androidx.core.content.pm.ShortcutInfoCompat
 import androidx.core.content.pm.ShortcutManagerCompat
 import androidx.core.graphics.drawable.IconCompat
@@ -21,7 +22,16 @@ import java.util.UUID
 object PrivateShortcutCreateManager {
 
     fun createPrivateShortcut(context: Context) {
+
         if (!ShortcutManagerCompat.isRequestPinShortcutSupported(context)) return
+        val intent = getHomeScreenActivityIntent(context)
+        val shortcut = getShortcut(context, intent)
+        val intentSender = getIntentSender(context)
+        ShortcutManagerCompat.requestPinShortcut(context, shortcut, intentSender)
+
+    }
+
+    fun getShortcut(context: Context, intent: Intent): ShortcutInfoCompat {
 
         val icon = IconCompat.createWithResource(context, R.mipmap.ic_launcher_private_round)
         val shortcut = ShortcutInfoCompat.Builder(context, UUID.randomUUID().toString())
@@ -38,22 +48,37 @@ object PrivateShortcutCreateManager {
                 )
             )
             .setIcon(icon)
-            .setIntent(Intent(context, HomeActivity::class.java).apply {
-                action = Intent.ACTION_VIEW
-                flags = Intent.FLAG_ACTIVITY_NEW_TASK or Intent.FLAG_ACTIVITY_CLEAR_TASK
-                putExtra(HomeActivity.PRIVATE_BROWSING_MODE, true)
-                putExtra(
-                    HomeActivity.OPEN_TO_SEARCH,
-                    StartSearchIntentProcessor.PRIVATE_BROWSING_PINNED_SHORTCUT
-                )
-            })
+            .setIntent(intent)
             .build()
+
+
+        return shortcut
+    }
+
+    fun getHomeScreenActivityIntent(context: Context): Intent {
+
+        return Intent(context, HomeActivity::class.java)
+            .setAction(Intent.ACTION_VIEW)
+            .setFlags(Intent.FLAG_ACTIVITY_NEW_TASK or Intent.FLAG_ACTIVITY_CLEAR_TASK)
+            .putExtra(HomeActivity.PRIVATE_BROWSING_MODE, true)
+            .putExtra(
+                HomeActivity.OPEN_TO_SEARCH,
+                StartSearchIntentProcessor.PRIVATE_BROWSING_PINNED_SHORTCUT
+            )
+
+    }
+
+    fun getIntentSender(context: Context): IntentSender {
+
         val homeScreenIntent = Intent(Intent.ACTION_MAIN)
             .addCategory(Intent.CATEGORY_HOME)
             .addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
+
         val intentSender = PendingIntent
             .getActivity(context, 0, homeScreenIntent, PendingIntent.FLAG_UPDATE_CURRENT)
             .intentSender
-        ShortcutManagerCompat.requestPinShortcut(context, shortcut, intentSender)
+
+        return intentSender
+
     }
 }
--- a/app/src/test/java/org/mozilla/fenix/components/PrivateShortcutCreateManagerTest.kt
+++ b/app/src/test/java/org/mozilla/fenix/components/PrivateShortcutCreateManagerTest.kt
@@ -1,0 +1,90 @@
+package org.mozilla.fenix.components
+
+import android.content.Context
+import android.content.Intent
+import android.content.IntentSender
+import androidx.core.content.pm.ShortcutInfoCompat
+import androidx.core.content.pm.ShortcutManagerCompat
+import androidx.core.graphics.drawable.IconCompat
+import androidx.test.core.app.ApplicationProvider
+import io.mockk.every
+import io.mockk.mockkStatic
+import io.mockk.spyk
+import io.mockk.verify
+import kotlin.random.Random
+import org.junit.Assert.assertNotNull
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.mozilla.fenix.HomeActivity
+import org.mozilla.fenix.R
+import org.mozilla.fenix.helpers.FenixRobolectricTestRunner
+import org.mozilla.fenix.home.intent.StartSearchIntentProcessor
+
+
+@RunWith(FenixRobolectricTestRunner::class)
+class PrivateShortcutCreateManagerTest {
+
+    @Test
+    fun `GIVEN shortcut pinning is supported WHEN createPrivateShortcut is called THEN create a pinned shortcut`() {
+        val context: Context = ApplicationProvider.getApplicationContext()
+
+        val privateShortcutCreateManager = spyk(PrivateShortcutCreateManager)
+
+        val intent = privateShortcutCreateManager.getHomeScreenActivityIntent(context)
+        `assert HomeActivityIntent correctly built`(intent)
+
+        val shortcut = privateShortcutCreateManager.getShortcut(context, intent)
+        `assert shortcut is correctly built`(shortcut, intent)
+
+        val intentSender: IntentSender = spyk(privateShortcutCreateManager.getIntentSender(context))
+
+        /**
+         *  Robolectric returns null for binder ue to which
+         *  intentSender.asBinder() calls in hashCode() fails with NullPointerException
+         */
+
+        val intentSenderHashCode = Random.nextInt()
+        every { intentSender.hashCode() } returns intentSenderHashCode
+
+        every { privateShortcutCreateManager.getIntentSender(context) } returns intentSender
+        every { privateShortcutCreateManager.getHomeScreenActivityIntent(context) } returns intent
+        every { privateShortcutCreateManager.getShortcut(context, intent) } returns shortcut
+
+        mockkStatic(ShortcutManagerCompat::class)
+
+        privateShortcutCreateManager.createPrivateShortcut(context)
+
+        verify { ShortcutManagerCompat.isRequestPinShortcutSupported(context) }
+        assert(ShortcutManagerCompat.isRequestPinShortcutSupported(context))
+
+        verify { ShortcutManagerCompat.requestPinShortcut(context, shortcut, intentSender) }
+
+    }
+
+    private fun `assert HomeActivityIntent correctly built`(intent: Intent) {
+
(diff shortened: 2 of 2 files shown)


RotBolt:
Please notice the green line for code-coverage

image

RotBolt:
Cc @sblatz and @eliserichards

app/src/test/java/org/mozilla/fenix/components/PrivateShortcutCreateManagerTest.kt
```diff
@@ -0,0 +1,107 @@
...
+
+    private fun getMockedHomeActivityIntent() : Intent {
+        val homeActivityIntent =  mockkClass(Intent::class)
+        every { anyConstructed<Intent>().setAction(Intent.ACTION_VIEW) } returns homeActivityIntent
+        every { homeActivityIntent.setFlags(Intent.FLAG_ACTIVITY_NEW_TASK or Intent.FLAG_ACTIVITY_CLEAR_TASK) } returns homeActivityIntent
+        every { homeActivityIntent.putExtra(HomeActivity.PRIVATE_BROWSING_MODE, true) } returns homeActivityIntent
+        every { homeActivityIntent.putExtra(
+            HomeActivity.OPEN_TO_SEARCH,
+            StartSearchIntentProcessor.PRIVATE_BROWSING_PINNED_SHORTCUT
+        ) } returns homeActivityIntent
```
Mugurell, line 83: This is a good example of why this approach of writing the test is not optimal.
Here we ensure that **if** the code under test calls this methods they will get the intent that we'll match against in the test.
But the test will pass even if the production code does not call these methods (which would be bad and we'd like to protect against such a regression) because in all cases we return from here an empty intent.
Ideally we want to ensure the Intent used in the production code is exactly the one we want and for this we need to check for all intended properties.

Mugurell (changes requested):
Thank you for taking a stab at this!

It's an interesting approach to construct the data in test and then do a 1-1 match with the same from the production code instead of verifying each property and method call.

If it worked I would be okay with it but if (like in the current implementation) you are hardcoding the same data in test and production code and then just match that we are not actually checking the indeded properties and do not protect against regressions.

To abide more to the current way of writing tests on this project (and so ensure an easier maintainance) I'd say multiple tests in which each small part is individually tested (with verify blocks and asserts) would be better.
Also, instead of mocking so many things you could consider relying on `FenixRobolectricTestRunner`.

app/src/test/java/org/mozilla/fenix/components/PrivateShortcutCreateManagerTest.kt
```diff
@@ -0,0 +1,107 @@
...
+        val shortcut = getMockedShortcutInfoCompat(context, homeActivityIntent)
+
+        val intentSender = getMockedIntentSender(context)
+
+        mockShortcutCompatManagerAndMethods(context, shortcut, intentSender)
+
+        privateShortcutCreateManager.createPrivateShortcut(context)
+
+        verify (exactly = 1) { ShortcutManagerCompat.isRequestPinShortcutSupported(context) }
+        verify (exactly = 1) { ShortcutManagerCompat.requestPinShortcut(context,shortcut,intentSender) }
```
Mugurell, line 44: I'd say ideally we'd want to capture the `shortcut` and `intentSender` as constructed from the SUT and then check for all the properties we expect to be there.

RotBolt:
> Also, instead of mocking so many things you could consider relying on `FenixRobolectricTestRunner`.

It was written on `FenixRobolectricTestRunner` that `Use it when really necessary`
But if that's the case I will use `RoboElectric`

app/src/test/java/org/mozilla/fenix/components/PrivateShortcutCreateManagerTest.kt
```diff
@@ -0,0 +1,107 @@
...
+
+    @Test
+    fun `when createPrivateShortcut is called and Pin shortcut is supported then requestPinShortcut should be called`(){
+        val context = mockkClass(Context::class)
+
+        val privateShortcutCreateManager = spyk(PrivateShortcutCreateManager)
+
+        mockkConstructor(Intent::class)
+
+        val homeActivityIntent = getMockedHomeActivityIntent()
```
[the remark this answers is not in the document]
RotBolt: Should I separate `getting home activity intent` in the production code as well and then test with `roboElectric` ?

app/src/test/java/org/mozilla/fenix/components/PrivateShortcutCreateManagerTest.kt
```diff
@@ -0,0 +1,107 @@
...
+
+        val shortcut = getMockedShortcutInfoCompat(context, homeActivityIntent)
```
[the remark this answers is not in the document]
RotBolt: Should I separate `creating shortcut info compat` in the production code as well and then test with `roboElectric`?

app/src/test/java/org/mozilla/fenix/components/PrivateShortcutCreateManagerTest.kt
```diff
@@ -0,0 +1,107 @@
...
+
+        val intentSender = getMockedIntentSender(context)
```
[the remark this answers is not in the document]
RotBolt: Similaly Should I separate `getting intentSender` in the production code as well and then test with `roboElectric`?

app/src/test/java/org/mozilla/fenix/components/PrivateShortcutCreateManagerTest.kt
RotBolt: Ok for this one way I could think of is to separate out is `getting intent sender, homeActivity intent` and `shortcut` into different functions and then `use assertions`
WDYT ?

app/src/test/java/org/mozilla/fenix/components/PrivateShortcutCreateManagerTest.kt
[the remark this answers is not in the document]
Mugurell: You could go with refactoring `PrivateShortcutCreateManager` to extract these producers making them easier to work with, then test each such method individually and then making sure that `createPrivateShortcut`  calls all called all of them and in the correct order.

app/src/test/java/org/mozilla/fenix/components/PrivateShortcutCreateManagerTest.kt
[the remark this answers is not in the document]
Mugurell: Would probably make the code easier to test, reason with and maintain. 👍

app/src/test/java/org/mozilla/fenix/components/PrivateShortcutCreateManagerTest.kt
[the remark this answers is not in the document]
Mugurell: Would probably make the code easier to test, reason with and maintain. 👍

app/src/test/java/org/mozilla/fenix/components/PrivateShortcutCreateManagerTest.kt
Mugurell: If you are breaking `createPrivateShortcut` in multiple methods tested individually here you'd basically test them again.
So maybe you could have some 
> fun `assert IntentSender is correctly built`(actualIntent: PendingIntent) { .. }
etc

methods containing all the assertions for one particular component and then call these both when you're testing each individual method and when testing them all in `requestPinShortcut` to avoid duplicating that checks.

Or you could not refactor `createPrivateShortcut` and then you wouldn't have duplicated checks. Just once, splitted in different assert methods or not.

RotBolt: Let's say If I don't refractor, then How could I capture the internal object creation ? 😅

app/src/test/java/org/mozilla/fenix/components/PrivateShortcutCreateManagerTest.kt
```diff
@@ -0,0 +1,98 @@
...
+         */
+
+        val intentSenderHashCode = Random.nextInt()
+        every { intentSender.hashCode() } returns intentSenderHashCode
+
+        every { privateShortcutCreateManager.getIntentSender(context) } returns intentSender
+        every { privateShortcutCreateManager.getIcon(context) } returns icon
+        every { privateShortcutCreateManager.getHomeScreenActivityIntent(context) } returns intent
+        every { privateShortcutCreateManager.getShortcut(context, icon, intent) } returns shortcut
+
```
RotBolt, line 57: IntentSender object created while running with Robolectric has a problem, it does not have the binder attached
Due to which it fails during verify {}

Also IntentSender does not provide anything to test against. So I decided to spy it and mocked the hashcode()

I am just returning objects created by SUT only (after asserting them that they are built correctly)

I mocked the result for intentSender.hashcode() because running with Robolectric will try to run the actual method , but the IntentSender object created by SUT under Robolectric does not have a binder attached due to which it fails with NullPointerException

this is the error

java.lang.NullPointerException
	at android.content.IntentSender.hashCode(IntentSender.java:295)
	at io.mockk.EqMatcher.hashCode(Matchers.kt)
	at java.util.AbstractList.hashCode(AbstractList.java:541)
	at io.mockk.InvocationMatcher.hashCode(API.kt:3754)
	at io.mockk.RecordedCall.hashCode(API.kt:3800)
	at java.util.HashMap.hash(HashMap.java:339)
	at java.util.HashMap.put(HashMap.java:612)

I think Robolectric returns null for binders
robolectric/robolectric#2579 (comment)

app/src/main/java/org/mozilla/fenix/components/PrivateShortcutCreateManager.kt
```diff
@@ -38,22 +46,35 @@ object PrivateShortcutCreateManager {
...
-                )
-            })
+            .setIntent(intent)
             .build()
+
+
+        return shortcut
+    }
+
+    fun getIcon(context: Context): IconCompat {
```
gabrielluong, line 56: I would probably avoid these one line function since it's not being reused.

app/src/test/java/org/mozilla/fenix/components/PrivateShortcutCreateManagerTest.kt
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    private fun `assert HomeActivityIntent correctly built`(intent: Intent) {
+
+        assert(intent.action == Intent.ACTION_VIEW)
+        assert(intent.flags == Intent.FLAG_ACTIVITY_NEW_TASK or Intent.FLAG_ACTIVITY_CLEAR_TASK)
+
+        val bundle = intent.extras
+
+        assert(bundle != null)
```
gabrielluong, line 76:
```suggestion
        assertNotNull(bundle)
```

app/src/main/java/org/mozilla/fenix/components/PrivateShortcutCreateManager.kt
RotBolt: removed

app/src/test/java/org/mozilla/fenix/components/PrivateShortcutCreateManagerTest.kt
RotBolt: done

Mugurell (changes requested):
Thank you again @RotBolt for taking this on.
I see the refactorization being a lot bigger than really wanted with many issues still to resolve:
- more test cases would be needed like for the the response of `getIntentSender` or usage of`ShortcutManagerCompat.isRequestPinShortcutSupported` 
- issues related to code style caught by ktlint
- the general structure of the proposed tests is quite different from what we have now making it difficult to follow and leading to an increased maintenance cost in the future.

After checking with Gabriel we propose closing this PR.

If you'd like to give it another shot while keeping the production code changes small and following the current structure we use for tests we'd be happy to review another PR.
If you'd like to help on other issues good for project newcomers like the ones from here https://github.com/mozilla-mobile/fenix/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22 your contribution would be greatly appreciated.

RotBolt:
Hello @Mugurell 
Problem with not refractor the production code is that `how to capture the objects created by SUT` ?
Since these are created under the function. I am not sure then how to capture these objects. If you could help a little here then it would be great

Mugurell:
Mockk has special support for capturing - https://mockk.io/#capturing
For testing that ` ShortcutManagerCompat.requestPinShortcut` is called with the right arguments you could use something like:
```
...
val shortcut = slot<ShortcutInfoCompat>()
val intentSender= slot<IntentSender>()

verify { ShortcutManagerCompat.requestPinShortcut(testContext, capture(shortcut), capture(intentSender)) }

assertEquals(
    testContext.getString(R.string.app_name_private_5, context.getString(R.string.app_name)),
    shortcut.captured.shortLabel
)
...
```

Other things to look for in a new PR:
- small, independent tests covering all scenarios
- try to organize each the test code in the 3 big sections: arrange, act, assert. This makes the code easy to follow and reason with. Currently there's an overuse of newlines and a too big test.
- instead of using `assert` which would just fail a test without offering any clue about why try using it's many variations like `assertEquals`, `assertNotNull` etc. As these use an `expected / actual` when this fail we'll know why.
- please use [this](https://github.com/mozilla-mobile/fenix#pre-push-hooks) pre-push hook configuration or at least run manually ktlint and detekt before pushing a commit to ensure the code is clean and follows our code style.

Any of the [current tests](https://github.com/mozilla-mobile/fenix/tree/master/app/src/test/java/org/mozilla/fenix) can serve as an inspiration.

RotBolt:
ok sure 
Thanks for this information will definitely raise another PR for this

Issue #20257, opened 2021-07-05
[Bug] Grid tab view no longer displays as a grid


SeburoSUMO:
[comment]: # (Please do your best to search for duplicate issues before filing a new issue so we can keep our issue board clean)
[comment]: # (Every issue should have the exact bug and steps to reproduce described in it. Please do not file feedback list tickets as it is difficult to parse them and address their individual points)
[comment]: # (Read https://github.com/mozilla-mobile/fenix#i-want-to-file-an-issue for more information)

## Steps to reproduce

1. Make sure that Grid view is selected
2. Open multiple new tabs
3. View tabs in tab drawer

### Expected behavior

Tabs are viewed as a grid

### Actual behavior

Tabs viewed as wide tabs (see image attached to https://support.mozilla.org/en-US/questions/1342405)

### Device information

* Device vendor / model and Android version: Pixel 3 64GB, Android 11
* Firefox for Android version: 89.1.1 (Build #2015812947)

Please see SUMO questions https://support.mozilla.org/en-US/questions/1342405

Nickoriginal:
Hello!
**Firefox 89.1.1**
_Not reproducible_ on:
**Samsung Galaxy A20s** (Android 10)
**Samsung Galaxy J2 Prime** (Android 6.0.1)
<img src = https://user-images.githubusercontent.com/85299944/124547713-393f1300-de35-11eb-96ea-6f9c5f385210.jpg width = 45%>

SoftVision-LorandJanos:
Hi @SeburoSUMO ! Thanks for the report! I was not able to reproduce your issue. I’ve used a device close to yours (Google Pixel 4 with Android 11) with the same FF build (89.1.1 (Build #2015812947)), and the grid is activated as expected. Could you check on the latest Nightly or if you can, clear your data on the same build and check if it’s still reproducible? Thank you!

OlliKn:
Got the same issue on All nightlys since the New gridview was launched.
Honor 6x, lineageos 16

OlliKn:
Screenshot:
![Firefox_Nightly_1](https://user-images.githubusercontent.com/61148677/124558954-6e4e6400-de3b-11eb-8233-ddfba79df76d.png)

codrut-topliceanu:
@OlliKn @SeburoSUMO 
I'm also unable to reproduce this issue, could you record a video of the erroneous behavior please? 🙏 
It would help us a great deal in getting the issue fixed.

[...]

PR #20281, opened 2021-07-07
For #20257 - Sets min no. of grid columns to 2 in TabsTray

codrut-topliceanu:
For https://github.com/mozilla-mobile/fenix/issues/20257
### Pull Request checklist

- [x] **Tests**: This PR includes no tests as it's only a minor change.

1. click on Show All Checks,

image

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/tabstray/ext/Context.kt
+++ b/app/src/main/java/org/mozilla/fenix/tabstray/ext/Context.kt
@@ -15,5 +15,5 @@ internal val Context.numberOfGridColumns: Int
     get() {
         val displayMetrics = resources.displayMetrics
         val screenWidthDp = displayMetrics.widthPixels / displayMetrics.density
-        return (screenWidthDp / MIN_COLUMN_WIDTH_DP).toInt()
+        return (screenWidthDp / MIN_COLUMN_WIDTH_DP).toInt().coerceAtLeast(2)
     }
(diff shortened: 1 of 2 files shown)


Mugurell:
The thumbnails shown in the screenshot are very blurry.
Is that only from setting the minimum number of columns to be two? 😏

codrut-topliceanu:
@Mugurell 😆 No, that's from the low dpi emulator used.

jonalmeida:
Looks like [the old code](https://github.com/mozilla-mobile/fenix/blob/v86.1.1/app/src/main/java/org/mozilla/fenix/tabtray/TabTrayView.kt#L468) defaulted to 2 columns as well, but this was missed during the refactor. Thank you for catching and fixing this!

Could you update/add a test to avoid this regression from happening again?

codrut-topliceanu:
> Looks like [the old code](https://github.com/mozilla-mobile/fenix/blob/v86.1.1/app/src/main/java/org/mozilla/fenix/tabtray/TabTrayView.kt#L468) defaulted to 2 columns as well, but this was missed during the refactor. Thank you for catching and fixing this!

Sooo... no need for UX-feedback in that case right?

I'll get started on those tests.

codrut-topliceanu:
😄 🙏 @jonalmeida

jonalmeida (approved):
✅✅

PR #20395, opened 2021-07-16
Add preference for enabling/disabling Android Autofill from within the app.

pocmo:
This adds a preference to Fenix for enabling/disabling Android Autofill support from within the app. To distinguish the setting from the one for autofilling logins in web content, I changed the string:

autofill_preference


pocmo:
@topotropic Are the new strings okay? Is there someone else I should involve?

topotropic:
Thanks for the ping! @lime124 can help you out here!

pocmo:
Updated strings:

background

jonalmeida (approved):
Looks good code-wise.

jonalmeida:
Not merging in case there is some other review we're waiting for.

PR #20398, opened 2021-07-16

MickeyMoz:


Mugurell:
I see `ui-test-x86-debug` failed because of:
```
Warning: Failed to make bucket for moz-fenix
Cause: Sorry, that name is not available. Please try a different one.
```

Let's try one more time.

Mugurell:
SearchTest#shortcutButtonTest fails [here](https://github.com/mozilla-mobile/fenix/blob/41ba94b9517f9d61dcfc147c910de1ecc189f4fd/app/src/androidTest/java/org/mozilla/fenix/ui/robots/SearchRobot.kt#L225) because of [a recent change](https://github.com/mozilla-mobile/android-components/pull/10599) that prepends the left to right mark for text in [DisplayToolbar].
So the text is now actually 
`\u200Educkduckgo.com/?q=mozilla` instead of the searched for 
`duckduckgo.com/?q=mozilla`

Mugurell:
Created https://github.com/mozilla-mobile/fenix/issues/20420 to temporarily disable that test.
After the patch for it is merged the AC bump should be able to be merged also.

PR #20436, opened 2021-07-20
Bug 1721329 - Fix perma fail on browsertime tests.

bebef1987:
Replace  bbc with dailymail and jianshu with sina


bhearsum (approved):
releng rubberstamp

csadilek (approved):
👍

PR #20456, opened 2021-07-21
No issue: Update Flank to v21.07.1

AaronMT:
Flank changed some of the JSON in `matrix_ids.json` that I am parsing across projects (A-C upgraded already) in a separate project, which means time to update again. Also Flank adds a second `.sha256` release, so we need a little more finesse in downloading the `jar`. This also updates `parse-ui-test.py`


rpappalax (approved):
LGTM

Issue #16330, opened 2020-11-02
Investigate usages of Sentry.capture


ekager:
[comment]: # (Please do your best to search for duplicate issues before filing a new issue so we can keep our issue board clean)
[comment]: # (Every issue should have the exact bug and steps to reproduce described in it. Please do not file feedback list tickets as it is difficult to parse them and address their individual points)
[comment]: # (Read https://github.com/mozilla-mobile/fenix#i-want-to-file-an-issue for more information)

Chatted with @csadilek about sending one off events to Sentry, and we're not positive that our current usages of `Sentry.capture` are actually working, and if they are being sent - they are likely missing all the metadata like ac.version, as.version etc. 

Let's 
1. Test if these are actually being sent (from a quick Sentry search not sure...)
2. Modify or replace them with `crashReporter` uses to work correctly.

csadilek:
See: https://github.com/mozilla-mobile/fenix/pull/20049

Let's remove remaining calls to `Sentry.capture` and replace with our crashReporter and `submitCaughtException`.

abodea:
Synced with @csadilek on slack and he confirmed that this issue doesn't really need manual verification and it was already verified that is working as this was mostly an engineering health issue to make sure we use the right API so that logging to Sentry works.

abodea:
Closing this based on the comment from above as Csadilek already confirmed that this issue doesn't really need manual verification and it was already verified being an engineering health issue.

csadilek:
@abodea Ah sorry for the late response, there are two left-over `Sentry.capture` calls. I will put up a PR to replace them. No need to re-open :).

PR #20458, opened 2021-07-21
Issue #16330: Remove remaining usages of Sentry.capture

csadilek:
There were two remaining call-site of `Sentry.capture`, which didn't work and was never set up.

We want to know how often we generate a new encryption key so I migrated that to our `submitCaughtException`. 

The other call-site, is not worth migrating. An error there would result in a navigation simply not working which should be obvious during application usage i.e, it wouldn't be intermittent or just affect some users, we would see this use case generally not working which we don't to log to Sentry but catch during testing. Decided to log instead. It would also be messy and a bigger refactoring to get access to our crashreporter there. :)


grigoryk (approved):
Great! Love the NavController cleanup :)

Issue #18727, opened 2021-03-31
[Bug][PiP] The media buttons are still displayed when opening a new blank page in PiP


abodea:
## Steps to reproduce
1. Open a YT video on full screen and tap on the recent apps button.
2. The PiP is displayed and the video is running without issues.
3. Long tap on the Fenix icon and select `New tab or New Private tab`.
4. Note that the tab is opened but the media buttons are still displayed and the video can be controlled even tho it is not displayed.
### Expected behavior
Media buttons should not be displayed on the newly opened tab in PiP.
### Actual behavior
The media buttons are still displayed when opening a new blank page in PiP
### Device information

* Device vendor/model and Android version: **Google Pixel 4 XL (Android 11)**
* Firefox for Android version: **Nightly 3/31/21.**

https://user-images.githubusercontent.com/42831109/113163998-5fa74e00-9249-11eb-8c18-f70f45c50ecf.mp4

Mugurell:
The media buttons are not controlled by us but by the framework. They are standard for PIP which should only be used for media.

Since the user is opening a new tab I think the expected result is for the app to exit PIP and open a new tab in the fully shown application, tab that can be interacted with.
Like it would happen when while in PIP users tap on a link to open in Fenix.

Asking for UX's confirmation about the expected result.

topotropic:
Thanks! The new tab opened for me in the tiny PIP window... not useful. I agree with @Mugurell's proposal: exit PIP and open new tab in fully shown application.

czlucius:
May I work on this issue?

czlucius:
I think the activity needs to be expanded/exit PiP mode when the various entry points are launched:
- Notifications
- Links
- App shortcuts

There are much more entry points, but I think every way of launching the apps use `Intent`, which is handled by `IntentRecieverActivity`.

Hence, we can just add `intent.setFlags(Intent.FLAG_ACTIVITY_REORDER_TO_FRONT)` to `IntentReceiverActivity` in the `processIntent()` method. (https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/java/org/mozilla/fenix/IntentReceiverActivity.kt) 

Something like this:
```kotlin
if (isInPictureInPictureMode()) {
     intent.setFlags(Intent.FLAG_ACTIVITY_REORDER_TO_FRONT)
}
```

Mugurell:
I agree this probably needs a solution in `IntentReceiverActivity`.
[...]

PR #20486, opened 2021-07-23
For #18727: Exit PiP when launched externally with Intents

czlucius:
Linked issue: #18727

Adds exit PiP code in `HomeFragment`, exiting PiP on `onCreate`.

To check:
- [x] App doesn't exit PiP when in PiP and screen rotates
- [x] Check that intent in `HomeActivity` isn't null before exiting

### Pull Request checklist

1. click on Show All Checks,

Video of the fixed issue:

https://user-images.githubusercontent.com/58442255/126856970-c57b7fca-a0af-489c-89f8-40a3a634662c.mp4

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
+++ b/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
@@ -7,6 +7,7 @@ package org.mozilla.fenix
 import android.content.Context
 import android.content.Intent
 import android.content.Intent.ACTION_MAIN
+import android.content.Intent.FLAG_ACTIVITY_REORDER_TO_FRONT
 import android.content.res.Configuration
 import android.os.Build
 import android.os.Bundle
@@ -188,6 +189,9 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
             super.onCreate(savedInstanceState)
         }
 
+        // Checks if Activity is currently in PiP mode if launched from external intents, then exits it
+        checkAndExitPiP()
+
         // Diagnostic breadcrumb for "Display already aquired" crash:
         // https://github.com/mozilla-mobile/android-components/issues/7960
         breadcrumb(
@@ -264,6 +268,14 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
         }
 
         StartupTimeline.onActivityCreateEndHome(this) // DO NOT MOVE ANYTHING BELOW HERE.
+    }
+
+    private fun checkAndExitPiP() {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N && isInPictureInPictureMode && intent != null) {
+            // Exit PiP mode
+            moveTaskToBack(false)
+            startActivity(Intent(this, this::class.java).setFlags(FLAG_ACTIVITY_REORDER_TO_FRONT))
+        }
     }
 
     private fun startupTelemetryOnCreateCalled(safeIntent: SafeIntent) {


czlucius:
https://user-images.githubusercontent.com/58442255/126856970-c57b7fca-a0af-489c-89f8-40a3a634662c.mp4

Here's the video of the fixed issue. Tested on Android 11 on Samsung P610 (Tab S6 Lite).
I'll need to investigate into the possible side effects first.

Mugurell:
To me this looks good. We obtain the desired effect and seems the generally accepted solution in place of an official Android api for this.
Curious if @mcomella agrees with the this approach from a performance standpoint.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -187,6 +187,12 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
             super.onCreate(savedInstanceState)
         }
 
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N && isInPictureInPictureMode) {
+            // Exit PiP mode
+            moveTaskToBack(false)
+            startActivity(Intent(this, this::class.java).addFlags(Intent.FLAG_ACTIVITY_NEW_TASK))
```
mcomella, line 193: issue (non-blocking): I'm not sure how results in us exiting PiP mode.

suggestion: can you explain in the comment how this lets us exit PiP mode?

mcomella (commented):
> Curious if @mcomella agrees with the this approach from a performance standpoint.

Thanks for looping me in. From a perf perspective, I was concerned about `HomeActivity.onCreate` getting called twice but that doesn't seem to be the case. Subjectively, I tested on a Moto G5 and it doesn't seem unreasonable. It'd take too long for me to objectively benchmark the change because it doesn't have a clear start and ending point so **I'll say this seems reasonable from a perf perspective.**

---

From a correctness perspective, I'm wary about triggering the android lifecycle unnecessarily. It's difficult to code correctly around the common lifecycle patterns but the Android framework does unexpected things when we're intentionally recreating activities to reset state. For example, we use `activity.recreate()` to switch between normal and private browsing and it calls unexpected code paths such as HomeFragment.onCreateView being called before HomeActivity.onCreate even inflates the layout containing `HomeFragment` (because Android is automatically restoring the fragment stack without us having to inflate and define the state).

That being said, I don't know what `moveTaskToBack(...); startActivity(... this)` does specifically, it may be simpler than the `activity.recreate` case, and it may be the right solution for the problem here. I didn't look at this carefully from the correctness perspective but I just wanted to mention my general preference for finding solutions that don't trigger the lifecycle rather than ones that do. So I'm not blocking the review on this – thanks for the patch!

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
czlucius: https://stackoverflow.com/a/43288507/12204281
The code moves the activity to the back of the activity stack, then restores it to the front, bringing it out of PiP mode.

czlucius:
I seem to have found another solution that is not called in the `HomeActivity`.

We can add this to `processIntent` in `IntentReceiverActivity` so that the resulting intent launched would automatically exit PiP.
```kotlin
        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
            // Exit PiP
            moveTaskToBack(false)
            intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
        }
```

However, in `IntentReceiverActivity`, `isInPictureInPictureMode` returns false even when the previous activity is in PiP. Hence, we cannot check for PiP before executing this action (refer to https://github.com/mozilla-mobile/fenix/issues/18727#issuecomment-885540214), and it may have unintended side effects as we are doing this for every intent launched to Firefox.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
mcomella: I'm not sure I understand why moving the activity to the back of the activity stack and restoring it to the front would bring it out of PiP – it seems like it might be a hack. However, if it seems like this is the only way to do it, I'm fine with that – just link to the Stackoverflow post in a comment so others have context on where this solution came from.

In any case, I don't need to approve or reject the correctness changes: I'll defer to @Mugurell for those. @Mugurell Please finish reviewing – it looks good from my perspective on performance! Sorry if I was unclear about my intent.

Mugurell:
Remembered we were doing something similar on [Fennec also](https://hg.mozilla.org/mozilla-central/rev/add212ee1118#l2.192).
Speaking of which @czlucius can we change to use `FLAG_ACTIVITY_REORDER_TO_FRONT` as the flag for the `startActivity` Intent? Seems more appropriate than `FLAG_ACTIVITY_NEW_TASK`.
Then we can give this a shot in the new 93 train.

Mugurell (dismissed):

czlucius:
Sorry, I forgot to add the import statement before.

czlucius:
I've tested the fix with the new flag, it seems to work fine.

Mugurell:
@Mergifyio refresh

Mugurell:
@czlucius It seems there is an issue with the script we're using for landing PRs and right now this PR is also blocking others from landing so I'll close this just to let the others pass.
Can you please squash your work and rebase on the latest `main` to then try to merge it again?

czlucius:
Ok, I’ll try.

czlucius:
I'll try soon, but lately I've been quite busy, will do soon when I'm free

Mugurell:
> Ok, I’ll try. Do you mean squash my commits into 1 commit then rebase into the main branch?

Yes. Thank you!

czlucius:
Sorry, I have some trouble squashing the commits. Please ignore these commits.

czlucius:
@Mugurell how do I squash and rebase? The 3rd commit, `31fac7925 Change intent flag for PiP` has many (around 300) commits in between them, and if I rebase, I will be rewriting all commits in between.

Here's the interactive rebase preview:

```
pick d2f9c4e9c Exit PiP in HomeActivity after onCreate, addressing #18727
pick 4e10261ed Check if launched via external intent before exiting PiP mode
pick 0e18bd11d Update Android Components version to 91.0.20210613143120.
pick 8bd5ebfab Import l10n.
... more commits in between ...
pick 19139de9e Update Android Components version to 92.0.20210807143113.
pick 642dd517f Import l10n.
pick b8ec03e07 Update Android Components version to 92.0.20210808010301.
pick 31fac7925 Change intent flag for PiP
```

Mugurell:
> @Mugurell how do I squash and rebase? The 3rd commit, `31fac7925 Change intent flag for PiP` has many (around 300) commits in between them, and if I rebase, I will be rewriting all commits in between.

This is because mergify merged the commits from `main` into your branch a couple of times while your branch still had an older parent.
It should still work to squash everything into one big commit which after being rebased on the current `main` (which already contains those ~300 commits) should leave you with just your changes.
Otherwise when the history gets a bit tangled / there are too many conflicts for rebasing a small changeset it might be easier to just start fresh on a new branch from the tip of the branch we want to merge in (in this case the just fetched `main`) and apply our patch there then push with something like `git push <remote> <local branch with new changes name>:fix-issue-18727 -f` (to push in this existing PR and not open another).

If you're going to update this PR it would be nice for the remaining commit to follow the same naming scheme as this PR - "For #\<ticket number\> - short changes description".

We could also help with this but you've put the work to investigate and test the issue and the solution so it would be nice to have this changelist have you as an author.

czlucius:
Ok, I shall start from a new branch and push soon in my free time.
Thanks for the suggestions.

czlucius:
I pushed a new commit already.

czlucius:
Can you help to check if the fix still works after this commit(just to be sure that I did not leave anything out)?
Thanks.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -188,6 +189,13 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
             super.onCreate(savedInstanceState)
         }
 
+        // Checks if in PiP mode, and if launched from external intents, to avoid unintended changes
+         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N && isInPictureInPictureMode && intent != null) {
+             // Exit PiP mode
+             moveTaskToBack(false)
+             startActivity(Intent(this, this::class.java).setFlags(FLAG_ACTIVITY_REORDER_TO_FRONT))
+         }
```
Mugurell, line 197: ` Unexpected indentation (9) (it should be 8) (cannot be auto-corrected)`

```suggestion
        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N && isInPictureInPictureMode && intent != null) {
            // Exit PiP mode
            moveTaskToBack(false)
            startActivity(Intent(this, this::class.java).setFlags(FLAG_ACTIVITY_REORDER_TO_FRONT))
        }
```

Mugurell:
@czlucius `detekt` now fails in CI because of 
```
ComplexMethod - 15/15 - [onCreate] at /home/runner/work/fenix/fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt:180:24
```

It doesn't so locally probably because the patch has an older base and the `onCreate` method has gone through some changes in the meantime which increased the complexity and with this new change it would go over the threshold.

So this needs to be rebased on `main` (the two commits can also be squashed) and then we'll try merging it.

czlucius:
Ok.

czlucius (commented):
I hope the indentation is fine

Mugurell:
@czlucius The indentation seems fine but `detekt` is failing because the `onCreate` method is not too complex - with the if block added by this patch there are too many branches.
The easiest (and probably the best here) solution is to just add the `@Suppress("ComplexMethod")` annotation to `onCreate`.

You can also run detekt and ktlint locally with
`./gradlew detekt` && `./gradlew ktlint` to check everything is ok.

czlucius:
Ok, I will make these changes.
On a side note, would it be ok to extract the code exiting PiP into a function to reduce ˋonCreateˋ's complexity?

Thanks.

Mugurell:
> On a side note, would it be ok to extract the code exiting PiP into a function to reduce ˋonCreateˋ's complexity?

Having that code in onCreate makes it easier to follow but if the new method has an enough of a suggestive name  that would work also 👍.
Maybe something like `checkAndExitPip`?

czlucius:
I've made the relevant changes, if ˋdektektˋ still shows errors, I will replace this with the ˋ@Suppressˋ solution.

Mugurell:
@czlucius Sorry for this being a drag but now it seems the `ktlint` task is failing because of
> /home/runner/work/fenix/fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt:273:1: Needless blank line(s)
/home/runner/work/fenix/fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt:276:1: Unexpected indentation (9) (should be 8)

czlucius:
screenshot-77b9fe3b

`ktlint` seems to be ok with this locally.

Mugurell:
> `ktlint` seems to be ok with this locally.
 
 That warning can be ignored for now.
 But I see in the CI tasks `ktlint` failing because of
 > /home/runner/work/fenix/fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt:273:1: Needless blank line(s)
 
 With this also fixed we can look into merging the patch.

czlucius:
ktlint is ok, but now a test is failing (hence test-debug fails)

Mugurell:
With everything green I think we can merge this.

Mugurell (approved):

czlucius:
Hi @Mugurell
I recently came across this: https://stackoverflow.com/a/19468012/12204281, and we may(I'm not so sure) need to add some other checks for the intent to ensure that the PiP will definitely not exit when onCreate is called, if intent is not null. This change is only applicable if the activity is recreated (onCreate is called again) while in PiP, which I do not think will happen. 
I've checked this, but I cannot find anything that suggests that in PiP, the activity still undergoes lifecycle changes that may accidentally trigger the exit PiP code.

Should I add the check for savedInstanceState just in case?

5hanks

czlucius:
@Mugurell may I ask, if in PiP, the activity still undergoes lifecycle changes? I did not see anything anywhere on the Android Developer website, nor on Stackoverflow that suggests this, but just to be sure.
Thanks.

Mugurell:
> @Mugurell may I ask, if in PiP, the activity still undergoes lifecycle changes? I did not see anything anywhere on the Android Developer website, nor on Stackoverflow that suggests this, but just to be sure.
> Thanks.

Don't think there is any change for the activity while in PiP though?
Nothing modifies about it. Still in foreground but in the `onPause` state Android puts it in for the entire duration of PiP.
From here it would either get `onResume` if returning to the app or `onStop` if PiP and the activity is closed.

Mugurell:
https://github.com/mozilla-mobile/fenix/pull/20831 merged these changes so I'll close the PR.
Thank you @czlucius !

czlucius:
> Thank you @czlucius !

Welcome.

Thank you for merging the fix.

czlucius:
> Don't think there is any change for the activity while in PiP though?
Nothing changes about it. Still in foreground but in the onPause state Android puts it in for the entire duration of PiP.
From here it would either get onResume if returning to the app or onStop if PiP and the activity is closed.

Noted.

PR #20529, opened 2021-07-27
Render debug info screen with Jetpack Compose.

pocmo:
Just a small usage of Jetpack Compose to render the debug info screen (and with that setup all the dependencies etc.)


buildSrc/src/main/java/Dependencies.kt
```diff
@@ -3,7 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 object Versions {
-    const val kotlin = "1.5.20"
+    const val kotlin = "1.5.10"
```
pocmo, line 6: I needed to downgrade from 1.5.20 to 1.5.10 because the Compose compiler plugin was complaining. I assume that with the next release of Compose we can go up again.

mcarare (approved):
LGTM! 🎆

PR #20397, opened 2021-07-16
Add dialog-like theme for AutofillSearchActivity

pocmo:
This adds a theme that makes `AutofillSearchActivity` look like a dialog - similar how the search screen looks in Lockwise.

PR #20548, opened 2021-07-28
Enable Android Autofill in Beta builds

pocmo:
This patch enables Android Autofill support in beta builds and uplifts the patches from #20547, #20397 and #20395.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -167,7 +167,6 @@ android {
         }
         beta {
             java.srcDirs = ['src/migration/java']
-            manifest.srcFile "src/migration/AndroidManifest.xml"
         }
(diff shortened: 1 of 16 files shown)


eliserichards (commented):
Do we care about the order that we receive these events? If we don't care about the order, then these should be aggregated into `metrics` instead of `Event`s

eliserichards:
Also, are the data reviews happening on those individual PRs, or on this one?

pocmo:
Yeah, data & code reviews are happening on the other PRs - This is primarily for uplifting.

Adding "do not land" label: I need to update the telemtry commit (updated it in the other PR before landing)

Issue #19876, opened 2021-06-07
Show a Home screen button to navigate to the home screen


gabrielluong:

lobontiumira:
Verified as fixed on the 6/11 Nightly build with OnePlus 5T (Android 10), Google Pixel (Android 10), and Samsung Galaxy Tab A6 (Android 5.1.1).

lobontiumira:
The Home screen button is not displayed on the page's toolbar, on Beta 91.0.0-beta.3 on any device.
But, the options to "Start on home" are available and working as expected.

amedyne:
@lobontiumira The HomeScreen button should only be displayed in Nightly. Can you clarify what you are referring to on "options to Start on Home"?

lobontiumira:
@amedyne I'm reffering to the options available in Settings - Tabs - Start on home.

amedyne:
@lobontiumira We will have to put the "Start on Home" settings on a nightly only feature flag. @Amejia481 will be doing that.

Amejia481:
> @lobontiumira We will have to put the "Start on Home" settings on a nightly only feature flag. @Amejia481 will be doing that.

This is the [PR](https://github.com/mozilla-mobile/fenix/pull/20555) for only showing the feature on nightly, after it gets merged I'll uplifted it to Beta.

PR #20555, opened 2021-07-28
Put the Start On home setting behind a feature flag.

Amejia481:
Related conversation https://github.com/mozilla-mobile/fenix/issues/19876#issuecomment-888445331

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
+++ b/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
@@ -220,7 +220,7 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
             shouldNavigateBrowserFragmentOnCouldStart(savedInstanceState)
         ) {
             navigateToBrowserOnColdStart()
-        } else {
+        } else if (FeatureFlags.showStartOnHomeSettings) {
             components.analytics.metrics.track(Event.StartOnHomeEnterHomeScreen)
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -220,7 +220,7 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
             shouldNavigateBrowserFragmentOnCouldStart(savedInstanceState)
         ) {
             navigateToBrowserOnColdStart()
-        } else {
+        } else if (FeatureFlags.showStartOnHomeSettings) {
```
eliserichards, line 223: Can we use `shouldStartOnHome()` here instead of this?

Amejia481: Good point, we can but I would like to make explicit that we have to remove this check when removing the feature flag. As internally `shouldStartOnHome()` does a read from disk, that was already performed in the if condition statement above.

eliserichards: That makes sense 👍

eliserichards (approved):
Lookin' good! 🚢

Amejia481:
@Mergifyio backport releases_v91.0.0

PR #20556, opened 2021-07-28
Use stable version of Jetpack Compose and related tooling/dependencies.

pocmo:
🎉


mcarare (approved):
🎆

pocmo:
`WebPushEngineIntegrationTest.delegate calls getSubscription()` - Unrelated, the same mocking lambda problem we have been fixing in other PRs. Will look at this separately.

mcarare:
> `WebPushEngineIntegrationTest.delegate calls getSubscription()` - Unrelated, the same mocking lambda problem we have been fixing in other PRs. Will look at this separately.

just tested, this passes for me locally even for multiple runs

PR #20566, opened 2021-07-29
Fix node toolchain task to use correct extension for artifact.

bhearsum:
This changed at some point in Gecko (I can't find when), and was only recently rebuilt to take effect. Tasks like https://firefox-ci-tc.services.mozilla.com/tasks/XF8w-CnQTA60AMS4HvkqZg, which we depend on, are publishing xz instead of zstd tarballs.

### DIFF
--- a/taskcluster/ci/toolchain/gecko-derived.yml
+++ b/taskcluster/ci/toolchain/gecko-derived.yml
@@ -43,7 +43,7 @@ linux64-minidump-stackwalk:
 
 linux64-node:
     attributes:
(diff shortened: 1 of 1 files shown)


escapewindow (approved):

pocmo (approved):

bhearsum:
Ack! @escapewindow just pointed out that this will break the next time the toolchain task runs on mozilla-central. The right fix here is rerunning the tasks on central. Closing this; will do that instead.

PR #20578, opened 2021-07-29
External source support

grigoryk:
Adds handling of information about external referrer (package, category)
when dealing with external intents.

Companion PR for https://github.com/mozilla-mobile/android-components/pull/10711

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/IntentReceiverActivity.kt
+++ b/app/src/main/java/org/mozilla/fenix/IntentReceiverActivity.kt
@@ -6,10 +6,14 @@ package org.mozilla.fenix
 
 import android.app.Activity
 import android.content.Intent
+import android.content.pm.PackageManager
+import android.os.Build
 import android.os.Bundle
 import android.os.StrictMode
 import androidx.annotation.VisibleForTesting
 import mozilla.components.feature.intent.processing.IntentProcessor
+import mozilla.components.support.utils.EXTRA_ACTIVITY_REFERRER_CATEGORY
+import mozilla.components.support.utils.EXTRA_ACTIVITY_REFERRER_PACKAGE
 import org.mozilla.fenix.HomeActivity.Companion.PRIVATE_BROWSING_MODE
 import org.mozilla.fenix.components.IntentProcessorType
 import org.mozilla.fenix.components.getType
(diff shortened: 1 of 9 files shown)


app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
```diff
@@ -1006,7 +1006,7 @@ abstract class BaseBrowserFragment :
                 .collect {
                     if (!onboarding.userHasBeenOnboarded() &&
                         it.content.loadRequest?.triggeredByRedirect != true &&
-                        it.source !in intentSourcesList &&
+                        it.source is SessionState.Source.External &&
```
csadilek, line 1009: This should be `!is`, i.e., not from an intent.

csadilek (changes requested):
Looks good! Found one typo / bug below.

app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
csadilek: To keep the same behaviour (which we should likely refactor another time): Don't finish onboarding the tab was opened from an external source / intent.

csadilek: One change here would be that after restoring an external tab we would still not finish onboarding. Before this change it would be `Source.Restored` whereas now we restore the external source. Not sure if worth it but we could make it possible to check for both `External` and `Restored`?

app/src/main/java/org/mozilla/fenix/IntentReceiverActivity.kt
```diff
@@ -96,6 +101,25 @@ class IntentReceiverActivity : Activity() {
...
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP_MR1) {
+            return
+        }
+        // NB: referrer can be spoofed by the calling application. Use with caution.
+        referrer?.let {
+            intent.putExtra(EXTRA_ACTIVITY_REFERRER_PACKAGE, it.host)
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+                // Category is supported for API>=26.
+                it.host?.let { host ->
+                    val category = packageManager.getApplicationInfo(host, 0).category
```
csadilek, line 117: nit: Thinking we should handle `NameNotFoundException` here for the `getApplicationInfo` call? Not worth crashing over it. I am sure we'll see this and seems fine to just not add the category in this case.

csadilek (approved):
Looks good, one suggestion below to prevent a potential crash.

Tested quite a bit including migrating tabs.

pocmo (approved):

Issue #20586, opened 2021-07-30
Recommended stories


Mugurell:

Mugurell:
@QA: Please confirm Pocket recommendations appear on the homescreen even after the app is restarted without internet.

sflorean:
Verified as fixed on Nightly 9/30 with:
- Pixel 3 (Android 12)
- Huawei MediaPad M2 (Android 5.1.1)

PR #20588, opened 2021-07-30
For #20586 - Basic implementation.

Mugurell:
A Composable can easily observe a state and recompose for updates without the need for an AC feature with Presenters / Interactors, etc.
Here the HomeFragment is responsable to update the state and the Composable View will immediately display the data.

This is just the basic skeleton for the feature with an example UI that can be built on later.

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
+++ b/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
@@ -259,6 +259,10 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
 
         components.core.requestInterceptor.setNavigationController(navHost.navController)
 
+        if (settings().pocketRecommendations) {
+            components.core.pocketStoriesService.startPeriodicStoriesRefresh()
+        }
+
         StartupTimeline.onActivityCreateEndHome(this) // DO NOT MOVE ANYTHING BELOW HERE.
     }
 
@@ -394,6 +398,7 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
             )
         )
 
+        components.core.pocketStoriesService.stopPeriodicStoriesRefresh()
         privateNotificationObserver?.stop()
     }
 
--- a/app/src/main/java/org/mozilla/fenix/components/Core.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/Core.kt
@@ -59,6 +59,9 @@ import mozilla.components.service.digitalassetlinks.local.StatementApi
 import mozilla.components.service.digitalassetlinks.local.StatementRelationChecker
 import mozilla.components.service.location.LocationService
 import mozilla.components.service.location.MozillaLocationService
+import mozilla.components.service.pocket.Frequency
+import mozilla.components.service.pocket.PocketStoriesConfig
+import mozilla.components.service.pocket.PocketStoriesService
 import mozilla.components.service.sync.autofill.AutofillCreditCardsAddressesStorage
 import mozilla.components.service.sync.logins.SyncableLoginsStorage
 import mozilla.components.support.locale.LocaleManager
@@ -84,6 +87,8 @@ import org.mozilla.fenix.telemetry.TelemetryMiddleware
 import org.mozilla.fenix.utils.Mockable
 import org.mozilla.fenix.utils.getUndoDelay
 import org.mozilla.geckoview.GeckoRuntime
+import java.lang.IllegalStateException
+import java.util.concurrent.TimeUnit
 
 /**
  * Component group for all core browser functionality.
@@ -318,6 +323,14 @@ class Core(
     val thumbnailStorage by lazyMonitored { ThumbnailStorage(context) }
 
     val pinnedSiteStorage by lazyMonitored { PinnedSiteStorage(context) }
+
+    @Suppress("MagicNumber")
+    val pocketStoriesConfig by lazyMonitored {
+        PocketStoriesConfig(
+            BuildConfig.POCKET_TOKEN, client, Frequency(4, TimeUnit.HOURS), 7
+        )
+    }
+    val pocketStoriesService by lazyMonitored { PocketStoriesService(context, pocketStoriesConfig) }
 
     val topSitesStorage by lazyMonitored {
         val defaultTopSites = mutableListOf<Pair<String, String>>()
--- a/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
@@ -47,6 +47,7 @@ import com.google.android.material.snackbar.Snackbar
 import kotlinx.coroutines.Dispatchers.IO
 import kotlinx.coroutines.Dispatchers.Main
 import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.async
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.collect
 import kotlinx.coroutines.flow.map
@@ -68,6 +69,7 @@ import mozilla.components.feature.top.sites.TopSitesConfig
 import mozilla.components.feature.top.sites.TopSitesFeature
 import mozilla.components.lib.state.ext.consumeFlow
 import mozilla.components.lib.state.ext.consumeFrom
+import mozilla.components.service.pocket.stories.PocketStoriesUseCases
 import mozilla.components.support.base.feature.ViewBoundFeatureWrapper
 import mozilla.components.support.ktx.android.content.res.resolveAttribute
 import mozilla.components.support.ktx.kotlinx.coroutines.flow.ifChanged
@@ -239,6 +241,13 @@ class HomeFragment : Fragment() {
             )
         }
 
+        if (requireContext().settings().pocketRecommendations) {
+            lifecycleScope.async(IO) {
+                val stories = PocketStoriesUseCases().GetPocketStories(requireContext()).invoke()
+                homeFragmentStore.dispatch(HomeFragmentAction.PocketArticlesChange(stories))
+            }
+        }
+
         topSitesFeature.set(
             feature = TopSitesFeature(
                 view = DefaultTopSitesView(homeFragmentStore),
@@ -328,6 +337,7 @@ class HomeFragment : Fragment() {
 
         updateLayout(binding.root)
         sessionControlView = SessionControlView(
+            homeFragmentStore,
             binding.sessionControlRecyclerView,
             viewLifecycleOwner,
             sessionControlInteractor,
--- a/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesComposables.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesComposables.kt
@@ -1,0 +1,285 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+@file:OptIn(ExperimentalMaterialApi::class, ExperimentalAnimationApi::class)
+@file:Suppress("MagicNumber")
+
+package org.mozilla.fenix.home.sessioncontrol.viewholders.pocket
+
+import androidx.compose.animation.AnimatedVisibility
+import androidx.compose.animation.ExperimentalAnimationApi
+import androidx.compose.animation.core.animateFloatAsState
+import androidx.compose.foundation.background
+import androidx.compose.foundation.clickable
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.size
+import androidx.compose.foundation.lazy.LazyRow
+import androidx.compose.foundation.lazy.itemsIndexed
+import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.foundation.text.ClickableText
+import androidx.compose.material.Card
+import androidx.compose.material.ContentAlpha
+import androidx.compose.material.ExperimentalMaterialApi
+import androidx.compose.material.Icon
+import androidx.compose.material.IconButton
+import androidx.compose.material.LocalContentAlpha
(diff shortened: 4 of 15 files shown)


jonalmeida:
> Since @jonalmeida you worked on TabsTray and TopSites I'm curious if you think the proposed approach here of initializing the feature with a placeholder view that instead of displaying data will cache it in HomeFragmentStore to be shown later in the real View we want to be used that will be inflated in `SessionControlAdapter` seems reasonable.

Yeah, that sounds quite reasonable. With top sites, it seems like the caching is done in a Fenix wrapper of the top sites storage. We could do this for bookmarks/recent tabs in their respective features too by pushing the state from `HomeFragmentStore` first in `onStart` (if they differ from what's currently there) and then start to observe updates.

With caching it's easy to get this wrong though. I've noticed with recent bookmarks, I can delete one recent ones, and then navigate back to see the same bookmarks showing up there because they are what's currently in the home state (but there's also probably some other bug in there too).

> cache it in HomeFragmentStore to be shown later

The construction of the stores in Fenix is a bit opaque to me. A quick peek at how it works, seems to suggest that the ViewModelStore that contains our home state, is cleared in destroy. If you navigate from home to browser to home, will that give you a new HomeFragmentStore when `StoreProvider.get` is called again? This seems to be an internal implementation of 

That said, looking at AC parts of this PR and the PocketArticlesPlaceholder in this one, I don't follow how this design is caching the state in HomeFragmentStore right now.

Mugurell:
@jonalmeida 
> That said, looking at AC parts of this PR and the PocketArticlesPlaceholder in this one, I don't follow how this design is caching the state in HomeFragmentStore right now.

Maybe not my best choice of words.
The general flow would be that the AC feature will send the list of articles to it's PocketStoriesView for display before Fenix's `SessionControlAdapter` has added the Pocket view to be displayed on homescreen.
So I was planning to use this `PocketArticlesPlaceholder` to update the `HomeFragmentState` with articles instead of displaying them (because we don't have where) so that when `SessionControlAdapter` later adds the proper view to the screen and tries binding the articles list to it it will read them from the `HomeFragmentState`, not from the AC persistence layer. This is what I reffered to as caching - between feature's start/stop the articles are read from this intermediate State, not from the persistance layer which exists in the feature.

jonalmeida:
I think I follow. This sounds like something that would nice to have at the AC component level where the feature or pocket service handles the caching of network responses and knows when to re-fetch.

Related to my comment in https://github.com/mozilla-mobile/android-components/pull/10716#issuecomment-893924063 where I think we only need to expose the final API as the view that wants to accept the data classes. If we can handle the caching internally, that would be better instead of going the route of the [`TabCollectionStorage`](https://github.com/mozilla-mobile/fenix/blob/main/app/src/main/java/org/mozilla/fenix/components/TabCollectionStorage.kt).

Earlier I said, "With top sites, it seems like the caching is done in a Fenix wrapper" - I meant, Collections, in reference to the TabCollectionStorage. 😄

Mugurell:
If all the UI is to sit in Fenix I took this as an opportunity to test how Compose can be leveraged for this, especially in HomeFragment which is a bit too complex.
Seeing a strange hiccup when scrolling to this View, some frames are lost. Seems like "inflating" takes quite a while but haven't profiled why.

Other than that all seems to work nicely but I also saw many areas that needs improvement prior to actually using Compose "in production".
We need our own: 
- Typography
- Shapes
- Themes
- various Composables for most of the now duplicated UI 
  - like an expandable View with a chevron icon (that is used in many places now)
  - Compose does not play sound / trigger haptics when for all `clickable` items. We probably need our own Modifier to ensure the current UX is maintained.
- an easy solution to load images from URLs in Composables

(should this ^^ be a meta?)

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesComposables.kt
```diff
@@ -0,0 +1,283 @@
...
+            AnimatedVisibility(visible = isExpanded) {
+                content()
+            }
+        }
+    }
+}
+
+@Composable
+@Preview
+private fun finalDesign() {
```
Mugurell, line 252: I'd say all Composables should try to preview the end result.
The auto-update is great to get a feel of changes in any child Composables without the need to run the app on a device.
To get past the "non-default parameters" errors when Compose does not know what to render we can use `PreviewParameterProvider` or wrap the Composables in another Composable like here from where we can try to provide values for all parameters.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesComposables.kt
```diff
@@ -0,0 +1,283 @@
...
+            elevation = 6.dp,
+            shape = RoundedCornerShape(4.dp),
+            modifier = Modifier
+                .size(160.dp, 87.dp)
+                .padding(bottom = 8.dp)
+        ) {
+            // Don't yet have a easy way to load URLs in Images.
+            // Default to a solid color to make it easy to appreciate dimensions
+            Box(Modifier.background(Color.Blue))
+            // Image(
```
Mugurell, line 83: Don't think the `ThumbnailLoader` can be easily used now to load images in a Composable.
And using `android.widget.ImageView` here doesn't seem like a good approach.

jonalmeida:
> (should this ^^ be a meta?)

Yes! I recently started [a doc](https://docs.google.com/document/d/1BxGVm2eI7JegtmlWHPPwCbbmBARKClJSYSITn9buYVg/edit#) of questions and problems we can solve with Compose to avoid duplication of UI/styles/theme code when we start working with compose. Feel free to add to it. 🙂

Mugurell:
Regarding the changes from here, I checked with Nicole. This is not the final design so for the time being we need just a simple way to test the functionality while we shouldn't spend too much on UI details.

Mugurell:
I used the approach Sebastian recommended in https://github.com/mozilla-mobile/android-components/pull/10716#issuecomment-900449452.
Seems to work beautifully and be a solid foundation on which we can easily work later!

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -237,6 +239,12 @@ class HomeFragment : Fragment() {
             )
         }
 
+        val applicationContext = requireContext().applicationContext
+        lifecycleScope.launch(Dispatchers.IO) {
```
gabrielluong, line 243: This is more of a curiosity, but what would be the difference here if we used `viewlifecycleScope` instead of `lifecycleScope`?

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -237,6 +239,12 @@ class HomeFragment : Fragment() {
...
+            val stories = PocketStoriesUseCases().GetPocketStories(applicationContext).invoke()
+            homeFragmentStore.dispatch(HomeFragmentAction.PocketArticlesChange(stories))
```
gabrielluong, line 245: I think we also tend to wrap these up into a feature and you can also hide it behind the FeatureFlags.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesComposables.kt
```diff
@@ -0,0 +1,285 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
gabrielluong, line 1: From a file naming convention for Jetpack compose, I am wondering if we do feel the need to add the `Composables` suffix. Looking at the existing compose sample apps written, lead me to conclude that we don't really need to care about such differentiation. In Focus, I opted to just name it `TopSites.kt` to match the top level compose function that we want to be used to set the content. In this case, this would be `PocketStories.kt`.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesViewHolder.kt
```diff
@@ -0,0 +1,51 @@
...
+        }
+    }
+
+    companion object {
+        val LAYOUT_ID = View.generateViewId()
+    }
+}
+
+@Composable
+fun PocketStories(store: HomeFragmentStore) {
```
gabrielluong, line 43: I only care here to see this and feed off the inspiration for the new world.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesComposables.kt
Mugurell: Sebastian added a new Jetpack Compose Image Loader in https://github.com/mozilla-mobile/android-components/pull/10838

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
Mugurell: I see `viewLifecycleScope` as rather a naming idiom we tend to use for parameters referencing `lifecycleScope`.

`lifecycleScope` exists for every `LifecycleOwner` - mostly Fragments and Activities implementations and exists beween the created and destroyed states.
We also use `view.toScope` as maybe an analogue to the above but for plain Views with the scope existing between View's attached / detached states.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
Mugurell: Will do. Thanks!

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesComposables.kt
Mugurell: I agree that a file containing Composables shouln't by default end in `Composable`.
In this case though I went with this naming structure since I saw other files in the `viewholders` package ending in `View`/`Holder` so I wanted to make it a bit more clear about what may be found in this file - just `PocketStories` would not really say much in my opinion.
Can change if required.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesViewHolder.kt
Mugurell: Feedback is welcome!
For now I also wanted to just get a good start for a homescreen composable that can be developed on later.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -260,6 +260,10 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
 
         components.core.requestInterceptor.setNavigationController(navHost.navController)
 
+        if (settings().pocketRecommendations) {
+            components.core.pocketUpdater.startPeriodicStoriesRefresh()
+        }
+
```
pocmo, line 266: So those refreshs, we only want to do them while the HomeActivity is around?

app/src/main/java/org/mozilla/fenix/components/Core.kt
```diff
@@ -319,6 +324,14 @@ class Core(
 
     val pinnedSiteStorage by lazyMonitored { PinnedSiteStorage(context) }
 
+    @Suppress("MagicNumber")
+    val pocketConfig by lazyMonitored {
+        PocketStoriesConfig(
+            BuildConfig.POCKET_TOKEN, client, Frequency(4, TimeUnit.HOURS), 7
+        )
+    }
+    val pocketUpdater by lazyMonitored { PocketStoriesService(context, pocketConfig) }
```
pocmo, line 333: nit: I have been trying to find the "pocketUpdater" implementation in code, until I read this line and noticed it's actually a `PocketStoriesService`. Maybe those should be named similar?

app/src/main/java/org/mozilla/fenix/settings/SecretSettingsFragment.kt
```diff
@@ -76,6 +77,24 @@ class SecretSettingsFragment : PreferenceFragmentCompat() {
...
+            isVisible = Config.channel.isDebug
+            isChecked = context.settings().pocketRecommendations
+            onPreferenceChangeListener = object : SharedPreferenceUpdater() {
+                override fun onPreferenceChange(preference: Preference, newValue: Any?): Boolean {
+                    (newValue as? Boolean)?.let {
+                        if (it) {
+                            context.components.core.pocketUpdater.startPeriodicStoriesRefresh()
+                        } else {
+                            context.components.core.pocketUpdater.stopPeriodicStoriesRefresh()
+                        }
```
pocmo, line 91: Here we start the updater independently from the HomeFragment lifecycle? 🤔

pocmo (dismissed):
👍 

I like this new approach. Just some nits.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesComposables.kt
pocmo: In the `AwesomeBar()` code I started to move most composables to separate files and use packages to organize them.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
pocmo: TL;DR: `viewLifecycleScope` exists during the lifetime of the views (onCreateView/OnDestroyView), while `lifecycleScope` exists during the lifetime of the whole Fragment. This gets only interesting when a Fragment gets detached (views get destroyed) and later reattached (new views get created). I think the rule of thumb is that when you schedule anything that needs views, then use `viewLifecycleScope` to not keep running if the views are no longer actually on the screen.

We use `View.toScope()` (which is similar to `viewLifecycleScope`) in A-C for two reasons:
* In features we have access to `View` objects that get passed in as parameters, but we never see the `Fragment` instance. FWIW we could be running in an Activity without a Fragment.
* We saw a very scary bug where `viewLifecycleScope` actually didn't get cancelled. 😱

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
Mugurell: For the refresh a workmanager would be used independently of app lifecycle.
Since it's important for the data to be .. "fresh" maybe there's no need to download/persist new data if it won't be used.

Mugurell:
Seeing the same issue from https://github.com/mozilla-mobile/fenix/issues/20952 - "INSTALL_FAILED_INSUFFICIENT_STORAGE" in `run-ui`.

mcarare (approved):

Mugurell:
Looks like `run-ui` failing is not blocking the merge.

Issue #20533, opened 2021-07-27
Intermittent test failure - org.mozilla.fenix.library.history.HistoryControllerTest.onCopyItem


jonalmeida:
TC link: https://firefox-ci-tc.services.mozilla.com/tasks/IHY86EjRTY2vphNVgDIHww/runs/0/logs/public/logs/live.log

```
io.mockk.MockKException: Can't instantiate proxy for class kotlin.Function0
	at io.mockk.impl.instantiation.JvmMockFactory.newProxy(JvmMockFactory.kt:64)
	at io.mockk.impl.instantiation.AbstractMockFactory.newProxy$default(AbstractMockFactory.kt:29)
	at io.mockk.impl.instantiation.AbstractMockFactory.mockk(AbstractMockFactory.kt:59)
	at org.mozilla.fenix.library.history.HistoryControllerTest.<init>(HistoryControllerTest.kt:450)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:490)
	at org.junit.runners.BlockJUnit4ClassRunner.createTest(BlockJUnit4ClassRunner.java:217)
	at org.robolectric.RobolectricTestRunner$HelperTestRunner.createTest(RobolectricTestRunner.java:556)
	at org.junit.runners.BlockJUnit4ClassRunner$1.runReflectiveCall(BlockJUnit4ClassRunner.java:266)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.BlockJUnit4ClassRunner.methodBlock(BlockJUnit4ClassRunner.java:263)
	at org.robolectric.internal.SandboxTestRunner$HelperTestRunner.methodBlock(SandboxTestRunner.java:344)
	at org.robolectric.RobolectricTestRunner$HelperTestRunner.methodBlock(RobolectricTestRunner.java:565)
	at org.robolectric.internal.SandboxTestRunner$2.lambda$evaluate$0(SandboxTestRunner.java:274)
[...]

PR #20589, opened 2021-07-30
Issue #20533: Fix intermittent failures of DefaultSessionControlControllerTest

csadilek:
See https://github.com/mozilla-mobile/fenix/pull/20551#issue-698773022.

### DIFF
--- a/app/src/test/java/org/mozilla/fenix/home/DefaultSessionControlControllerTest.kt
+++ b/app/src/test/java/org/mozilla/fenix/home/DefaultSessionControlControllerTest.kt
@@ -34,6 +34,7 @@ import mozilla.components.feature.top.sites.TopSite
 import mozilla.components.support.test.ext.joinBlocking
 import mozilla.components.support.test.rule.MainCoroutineRule
 import org.junit.After
+import org.junit.Assert.assertTrue
 import org.junit.Before
 import org.junit.Ignore
 import org.junit.Rule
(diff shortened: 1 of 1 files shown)


app/src/test/java/org/mozilla/fenix/home/DefaultSessionControlControllerTest.kt
```diff
@@ -739,10 +722,38 @@ class DefaultSessionControlControllerTest {
...
 
-        controller.handleMenuOpened()
+        createController().handleMenuOpened()
 
         verify(exactly = 0) {
             navController.navigateUp()
         }
     }
+
+    private fun createController(
```
gabrielluong, line 732: I am curious about the reasoning  and lessons that can be learned behind adding a `createController` and calling it for each test. Does that help with the intermittent? Just looking at the issue comment, I assume we were just removing the mock for lambdas to help avoid intermittents.

gabrielluong (approved):

app/src/test/java/org/mozilla/fenix/home/DefaultSessionControlControllerTest.kt
csadilek: @gabrielluong yeah, that's all there is to it. There's a problem with `mockk` mocking lambdas that we're working around here for now. This started happening with Java 11 and we didn't want to back that out and all related upgrades that were needed for jetpack compose. We also can't wait until an upstream fix is available because there were too many intermittents to make this feasible. So, we can only work around it for now. We have it on the agenda for our next Android catch up too to decide what to do and how to prevent us from introducing more of those.

pocmo (approved):

PR #20621, opened 2021-08-02
No issue: fix glean metrics tests for expired metrics

eliserichards:
Original probe renewal PR: https://github.com/mozilla-mobile/fenix/pull/20517
Doc with details and Vesta's approval: https://docs.google.com/document/d/1NGlnTa9TPyTnd3ciUPbwujbITjkX8p8vJybXcZrrM2w/edit#
 
We have let these metrics expire (product decision, they may be removed later), and needed to remove them from the tests:
- `preferences.open_links_in_private`
- `preferences.private_search_suggestions`

The renewal for perf probes are being handled by the performance team, so adding an @ignore to the following tests for perf to remove when they are updated:
- StorageStatsMetricsTest.`WHEN reporting THEN the values from the storageStats are accumulated`
- StorageStatsMetricsTest.`WHEN reporting THEN the query duration is measured`

Remove the following probes entirely (cherry-picking from [this PR](https://github.com/mozilla-mobile/fenix/pull/20560) because these will need to land together to fix all of the breaks)
- `tabs_tray.cfr.dismiss`
- `tabs_tray.cfr.go_to_settings`

### Pull Request checklist

- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features. In addition, it includes a screenshot of a successful [accessibility scan](https://play.google.com/store/apps/details?id=com.google.android.apps.accessibility.auditor&hl=en_US) to ensure no new defects are added to the product.

### To download an APK when reviewing a PR:
1. click on Show All Checks,
2. click Details next to "Taskcluster (pull_request)" after it appears and then finishes with a green checkmark,
3. click on the "Fenix - assemble" task, then click "Run Artifacts".
4. the APK links should be on the left side of the screen, named for each CPU architecture

PR #20622, opened 2021-08-02
Fix intermittent test failures in TabsTouchHelperTest

pocmo:
Another one with:

> io.mockk.MockKException: Can't instantiate proxy for class kotlin.Function1

Happened in https://github.com/mozilla-mobile/fenix/pull/20621


pocmo:
Removed the landing label again, let's make sure #20621 lands first. :)

PR #20642, opened 2021-08-03
[SDK-344] Pass in first-run information to Nimbus for targeting

travis79:
Before this PR can build/land, it requires that the following PR in A-C land first and be available to Fenix: https://github.com/mozilla-mobile/android-components/pull/10735


app/src/main/java/org/mozilla/fenix/experiments/NimbusSetup.kt
```diff
@@ -63,7 +63,10 @@ fun createNimbus(context: Context, url: String?): NimbusApi {
             // Note: Using BuildConfig.BUILD_TYPE is important here so that it matches the value
             // passed into Glean. `Config.channel.toString()` turned out to be non-deterministic
             // and would mostly produce the value `Beta` and rarely would produce `beta`.
-            channel = BuildConfig.BUILD_TYPE
+            channel = BuildConfig.BUILD_TYPE,
+            customTargetingAttributes = mapOf(
+                "isFirstRun" to context.settings().isFirstRun.toString()
```
jonalmeida, line 68: @mcomella this looks like it will affect our startup time, but I see above that we touch shared prefs multiple times as part of Nimbus initialization already. Not sure if you're already aware of this since it's required for first run usage with Nimbus.

travis79: This is very good to know, if there is a less impactful way of storing and retrieving this I'd be happy to rework this to be as minimally impactful to startup as possible.

jonalmeida: @travis79 I think I know the answer to this already, but to confirm, the "first run" is needed for visual features that are needed on startup, is that right?

travis79: Right, the idea behind this is the ability to target first run users for experiments.

mcomella: Thanks for the ping – we can benchmark the difference in start up time. I added something to our board for MarcLeclair and I to triage. Is there anything special I need to get this code to run? e.g. add an API key or something?

travis79: @mcomella I don't think you need to do anything special to get this to run, it should run as part of normal init of Nimbus. Regardless of whether it is the actual first-run of the app or not, it should at least read the pref every launch. On the very first run, it would also write to the pref, so there may be some slight additional overhead difference between the very first run, and subsequent runs. Wiping the app data, or performing a fresh install, should allow you to trigger the "first-run" behavior which includes writing to the pref.

travis79: @mcomella @jonalmeida It looks like we are shifting priorities a bit on the first-run targeting, so less pressing to get this resolved soon, but I would appreciate an update on the timeframe for this (or even if there is something I could profile or do myself to help unblock this).
 
Thanks! 
 
cc @jhugman

MarcLeclair (approved):
So I reviewed the changes. From a profile perspective, I saw no big changes.

One thing I did notice is that, from the parent commit before the changes I got this:

```
{'max': 3462.0,
 'mean': 3357.08,
 'median': 3360.0,
 'min': 3230.0,
 'replicate_count': 25,
 'replicates': [3417.0, 3462.0, 3307.0, 3350.0, 3360.0, 3427.0, 3454.0, 3230.0,
                3449.0, 3312.0, 3417.0, 3274.0, 3427.0, 3358.0, 3249.0, 3402.0,
                3441.0, 3339.0, 3242.0, 3360.0, 3311.0, 3287.0, 3370.0, 3392.0,
                3290.0]}
```

After the changes, I got an improvement: 

```
{'max': 2985.0,
 'mean': 2871.44,
 'median': 2880.0,
 'min': 2745.0,
 'replicate_count': 25,
 'replicates': [2935.0, 2818.0, 2797.0, 2873.0, 2938.0, 2884.0, 2823.0, 2960.0,
                2941.0, 2911.0, 2813.0, 2880.0, 2815.0, 2782.0, 2980.0, 2985.0,
                2745.0, 2846.0, 2855.0, 2899.0, 2778.0, 2916.0, 2912.0, 2886.0,
                2814.0]}
```

I'm not sure if it's something else, or the changes, but there's definitely an improvement on the nav start. As for MAIN, everything stayed the same.

FYI, this was taken on my G5+ with Beta variant.

jonalmeida (approved):
Thanks March for perf review! This looks good to me from a code review too. So let's land this on Monday if possible.

PR #20681, opened 2021-08-04
Closes #20679: revert allowaccessmodification proguard disabling.

### LINKED ISSUE #20679: Backout allowaccessmodification proguard change
_Originally filed as https://mozilla-hub.atlassian.net/browse/FXP-1267_

In https://github.com/mozilla-mobile/fenix/pull/10183 we had to disable the `-allowaccessmodification` optimization because it broken funcitonality or crashed the app. This bug was fixed in R8 2.1.28 ([tag](https://r8-review.googlesource.com/c/r8/+/51755)), which it says is standard with AS 4.1 – I believe R8 is bundled with the Android Gradle Plugin so I think this actually means that AS 4.1 will create new projects using the version of the Android Gradle plugin that comes with that version of R8.

We should upgrade to that version of R8 and use the built-in proguard file rather than using our own with the optimization disabled.

—

We can use this version of R8 early than the Android Gradle Plugin would support with this technique https://issuetracker.google.com/issues/157535162#comment4 but this may increase the R8 version by a lot all at once I'm not sure we want to do that.

mcomella:
We disabled the allowaccessmodification proguard option because it broke
functionality or crashed the app (I can't rememeber). As far as we know,
the R8 bug was fixed in the R8 bundled with the Android Gradle Plugin
v4.1. We're now on AGP v7.0.0-rc1 so we should be able to revert this
now.

This commit reverts the following commits:

Revert "Proguard/r8: Do not allow access modification."
This reverts commit 98bf27fdd45b579def3991c472e59b95ea3d3046.

Revert "Dump `proguard-android-optimize.txt` into local configuration for later modification"
This reverts commit 88fe3fbf8238a442bc3fe20f86d597405154f201.

### Pull Request checklist

1. click on Show All Checks,


badboy:
I _think_ running the release-like app was enough to cause a crash in the old configuration back then.
So that would be one way to check if this is now fixed indeed.

I won't get to it today, but can put that on my list for tomorrow.

mcomella:
> I _think_ running the release-like app was enough to cause a crash in the old configuration back then.
> So that would be one way to check if this is now fixed indeed.

I have done this and didn't see any crashes but you're welcome to independently verify. :)

badboy (approved):

badboy:
Nah, then we're good

Issue #20663, opened 2021-08-04
Make inactive tabs section collapsible 


jonalmeida:
Part of the [designs](https://www.figma.com/file/am9EJGg6ow29PXWkc9jcOH/MR2?node-id=708%3A7081), are to make the inactive tab UI collapsible when the clicking the header; the chevron indicates the current UI state.

In the designs, there is no mention of making the collapsible state persist on every new open, so let's assume not for now and we can add that later when we find out.

Linked JIRA: [FNXV2-17122](https://mozilla-hub.atlassian.net/browse/FNXV2-17122)

MaggiWuerze:
Please make the state persistent. It is annoying to always have to scroll past the inactive tabs after closing the browser.

SoftVision-LorandJanos:
Hi @jonalmeida ! Verified as implemented on the latest Nightly (93.0a1 - 2021-08-13T05:08:25.050265).
Devices used:
- Samsung Galaxy Tab S3 (Android 9).
- Huawei P30 (Android 10).
- Xiaomi Redmi 11 Lite (Android 11).

Closing the ticket as verified.

PR #20687, opened 2021-08-05
Issue #20663: Make inactive card collapsible

jonalmeida:
This is starting off as a draft because I'm unsure how to proceed here and looking for comment. Adding the `isExpanded` to prefs seems like an overkill (we don't do this for Collections either) but given that the tabs tray is opened and closed several times, it would unhelpful to not have the expanded state live longer than the tabs tray. So I'm stuck in this predicament of where to place it.

I've tried putting it in the TabsTrayStore, but as mentioned, we'd like to have this live longer than the tabs tray and the store is destroyed when the tray is dismissed.

We do not have an app state store in Fenix, as we do in Focus, but that would be a neat place to keep this too.

My hacky approach, is to use a singleton `object` to store the bool `isExpanded` which gives us the interaction we want.

**Update:** Staying with the hacky approach, will follow-up with #20704 next.

https://user-images.githubusercontent.com/1370580/128439033-78cf1833-8a57-4936-b8ba-c3f3bfb80fd8.mp4

### Pull Request checklist

1. click on Show All Checks,


jonalmeida:
Using emojis in the PR title to get any random reviewer's attention for insight. 😅

Amejia481:
> We do not have an app state store in Fenix, as we do in Focus, but that would be a neat place to keep this too.

As you mentioned, above having the Fenix store to storing this and future state will be ideal, maybe this could be an opportunity to introduce the store.


I'm not sure if the draft, was meant to have some functionality or just a proof of concept, as I'm not able to compile. 

```
[task 2021-08-05T04:08:45.734Z] e: /builds/worker/checkouts/src/app/src/main/java/org/mozilla/fenix/tabstray/browser/NormalBrowserTrayList.kt: (67, 13): Type mismatch: inferred type is BrowserStore but InactiveTabsController was expected
[task 2021-08-05T04:08:45.734Z] e: /builds/worker/checkouts/src/app/src/main/java/org/mozilla/fenix/tabstray/browser/NormalBrowserTrayList.kt: (68, 13): Too many arguments for public constructor DefaultInactiveTabsInteractor(controller: InactiveTabsController) defined in org.mozilla.fenix.tabstray.browser.DefaultInactiveTabsInteractor
[task 2021-08-05T04:08:45.734Z] e: /builds/worker/checkouts/src/app/src/main/java/org/mozilla/fenix/tabstray/browser/NormalBrowserTrayList.kt: (69, 13): Too many arguments for public constructor DefaultInactiveTabsInteractor(controller: InactiveTabsController) defined in org.mozilla.fenix.tabstray.browser.DefaultInactiveTabsInteractor
[task 2021-08-05T04:08:45.734Z] 
```

pocmo:
> As you mentioned, above having the Fenix store to storing this and future state will be ideal, maybe this could be an opportunity to introduce the store.

Agree, that could be a reason. (What I was wondering in https://mozac.org/2021/07/05/whats-next.html). An app store would work really well with Jetpack Compose too.
For an activity-scoped state, you could also consider a `ViewModel` that is attached to the activity, but that's not really a model we have been using so far, so maybe that's not a reason to introduce it.

jonalmeida:
Sounds like creating an AppStore is the right direction. In which case, I'll update to not store any state for now (tabs tray resets on every open), and put up a separate PR to add an AppStore that will holder the state of this.

jonalmeida:
I've filed https://github.com/mozilla-mobile/fenix/issues/20704 to introduce an AppStore and remove the object class here.

I think this is ready to review now. Thanks for feedback! 🙃

rocketsroger (approved):

PR #20720, opened 2021-08-06

MickeyMoz:


csadilek:
Needs a breaking change fixed in: https://github.com/mozilla-mobile/fenix/pull/20722

PR #20729, opened 2021-08-06
Don't rebase GeckoView or Android-Components version bumps PRs before landing

bhearsum:
@csadilek, @rvandermeulen and I talked about this a bit and thought it would probably be OK. @pocmo, what do you think?


.mergify.yml
```diff
@@ -15,8 +15,11 @@ pull_request_rules:
         type: APPROVE
         message: MickeyMoz 💪
       merge:
-        method: rebase
-        strict: smart
+        # These PRs are just bumping versions, and very unlikely to break
+        # anything. It's not worth the delay of rebasing them
+        # (sometimes multiple times) before merging.
+        method: squash
+        strict: false
```
pocmo, line 22: For the bump PRs I guess we can try this. In this case I think we should do the same for the L10N Auto Merge below too.

.mergify.yml
```diff
@@ -15,8 +15,11 @@ pull_request_rules:
         type: APPROVE
         message: MickeyMoz 💪
       merge:
-        method: rebase
-        strict: smart
+        # These PRs are just bumping versions, and very unlikely to break
+        # anything. It's not worth the delay of rebasing them
+        # (sometimes multiple times) before merging.
+        method: squash
```
pocmo, line 21: nit: Wondering why we switch to `squash` here? Couldn't this be `rebase` with `strict: false`? I guess it doesn't matter much for the "single commit" PRs that the bot currently opens.

But apart from that, I just checked the ([docs](https://docs.mergify.io/actions/merge/) and it looks like the merge options are deprecated and have now more fine-grained queue controls (although only for premium plans?).

pocmo (approved):

.mergify.yml
bhearsum: Sure!

.mergify.yml
bhearsum: Yeah, that's a good point. When I wrote this I thought that we couldn't rebase without being up-to-date, but that's silly in retrospect. I'll revert this part.

Issue #17917, opened 2021-02-08
Migrate the deprecated Kotlin Android Extensions compiler plugin


mcarare:
Since Kotlin 1.4.20 `android-kotlin-extensions` is deprecated and it is to be removed ( timeline is September 2021 Kotlin release).

Migration guide: https://developer.android.com/topic/libraries/view-binding/migration.

Note: It seems unlikely a View "injection" library will be a candidate for replacing view binding. (Kotter Knife for example is deprecated.)

--- 

### Special note about handing the View Binding instance in Fragments:
[Google recommends the following idiom](https://developer.android.com/topic/libraries/view-binding#fragments):
```
private var _binding: ResultProfileBinding? = null
// This property is only valid between onCreateView and onDestroyView.
private val binding get() = _binding!!

override fun onCreateView(
    inflater: LayoutInflater,
    container: ViewGroup?,
    savedInstanceState: Bundle?
): View? {
    _binding = ResultProfileBinding.inflate(inflater, container, false)
    val view = binding.root
    return view
}

override fun onDestroyView() {
    super.onDestroyView()
    _binding = null
}
```
because Fragments have a more complex lifecycle than Activities or Android Views.
In fragments that are manually detached or put to backstack (most of ours are) the View will be destroyed but the fragment instance lives on.
Since when coming back to the fragment (or when it is re-attached) onCreateView will be called again (and the binding should be initialized again) Android docs recommend the above idiom to:
- null the View Binding to reduce the memory pressure in between onDestroyView and onCreateView and be good citizens
- have an easy and safe way to access it, not having to check it for null when used while the Fragment is displayed (we control the lifecycle of _binding and know that while the View exists that null assertion is safe).

If the fragment is not put to backstack and does not outlive the View we should not need this idiom as when it will be destroyed it will cleanup all properties by itself.

There is an open issue to simplify the Fragment's lifecycle - https://issuetracker.google.com/issues/124181028.

Mugurell:
For QA: We refactored the entire application in regards to how the app data will be displayed and how user interactions are recorded.
So this will need a more thorough testing to validate the following potential issues do not occur.
- A particular screen / part of is shown as empty
- A click / any interaction doesn't do anything

sflorean:
Tested on latest Nightly builds, and made exploratory on different websites & different devices, and no issues were encountered. Closing as verified.

PR #20757, opened 2021-08-09
For #17917: Use View binding in collections

codrut-topliceanu:
For https://github.com/mozilla-mobile/fenix/issues/17917

### Pull Request checklist

- [x] **Screenshots**: This PR does not include screenshots as no user visible changes have been made.

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/collections/CollectionCreationFragment.kt
```diff
@@ -25,6 +25,9 @@ class CollectionCreationFragment : DialogFragment() {
     private lateinit var collectionCreationStore: CollectionCreationStore
     private lateinit var collectionCreationInteractor: CollectionCreationInteractor
 
+    private var _binding: FragmentCreateCollectionBinding? = null
+    private val binding get() = _binding!!
```
eliserichards, line 29: should this be `!!`?

MozillaNoah: I've personally only seen it as a lateinit var

Mugurell: Fragments need special care since they have a more complex lifecycle.
In fragments that are manually detached or put to backstack (most of ours are) the View will be destroyed but the fragment instance lives on.
Since when coming back to the fragment (or when it is re-attached) onCreateView will be called again (and the binding should be initialized again) Android docs recommend the above idiom to:
- null the View Binding to reduce the memory pressure in between onDestroyView and onCreateView
- have an easy and safe way to access it, not having to check it for null when used while the Fragment is displayed (we control the lifecycle of `_binding` and know that while the View exists that null assertion is safe).

If the fragment is not put to backstack and does not outlive the View we should not need this idiom as when it will be destroyed it will cleanup all properties by itself.

There is an open issue to simplify the Fragment's lifecycle - https://issuetracker.google.com/issues/124181028.

codrut-topliceanu: @eliserichards I also dislike using `!!` 😞

gabrielluong: This is also the recommended approach from https://developer.android.com/topic/libraries/view-binding#fragments

gabrielluong (approved):

eliserichards (approved):
🚢 🚢 🚢

PR #20805, opened 2021-08-12
Fixes for various intermittent test failures that we've seen in the last few days

jonalmeida:
🔥 

Some commits give more details behind each change if context is required.

### Pull Request checklist

- [x] **Tests**: This PR includes thorough tests or an explanation of why it does not.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/exceptions/trackingprotection/ExceptionsFragmentStore.kt
+++ b/app/src/main/java/org/mozilla/fenix/exceptions/trackingprotection/ExceptionsFragmentStore.kt
@@ -6,14 +6,21 @@ package org.mozilla.fenix.exceptions.trackingprotection
 
 import mozilla.components.concept.engine.content.blocking.TrackingProtectionException
 import mozilla.components.lib.state.Action
+import mozilla.components.lib.state.Middleware
 import mozilla.components.lib.state.State
 import mozilla.components.lib.state.Store
 
 /**
  * The [Store] for holding the [ExceptionsFragmentState] and applying [ExceptionsFragmentAction]s.
  */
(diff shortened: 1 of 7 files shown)


Amejia481:
Any learnings? We should not use mocks? XD

jonalmeida:
Still seeing a test failure with `QuickSettingsFragmentReducerTest.WebsitePermissionAction - ChangeAutoplay`:

```
WebsitePermissionAction - ChangeAutoplay

org.mockito.exceptions.misusing.InvalidUseOfMatchersException: 
Misplaced or misused argument matcher detected here:

-> at mozilla.components.support.test.MatchersKt.any(Matchers.kt:17)
-> at mozilla.components.support.test.MatchersKt.any(Matchers.kt:17)
-> at mozilla.components.support.test.MatchersKt.any(Matchers.kt:17)

	at org.mozilla.fenix.settings.quicksettings.QuickSettingsFragmentReducerTest.WebsitePermissionAction - ChangeAutoplay(QuickSettingsFragmentReducerTest.kt:74)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
```

We're not even using `any()`.. any where!

So for now I've removed even the `mock()` calls and there are zero mocks. I suspect though that since we were using `mock()` from AC and we've now updated Mockito to a newer version, this failure won't show up any more either way.

> Any learnings? We should not use mocks? XD

General rule: Mocking everything under the sun is not helpful because then we aren't testing anything. 😄 

We need to change our patterns from verifying the calls to a controller, and instead the state changes from that. Sebastian already mentioned a variant of this, when we have stores that can be used to verify state changes.

It's hard to do this with existing code, but maybe with new code we can find better patterns.

Amejia481 (approved):

PR #20788, opened 2021-08-11
For #17917: Use View binding in tracking protection

codrut-topliceanu:
For https://github.com/mozilla-mobile/fenix/issues/17917

### Pull Request checklist

- [x] **Tests**: This PR includes no tests as it is only a minor change
- [x] **Screenshots**: This PR does not include screenshots as no user visible changes have been made.
- [x] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md).

### To download an APK when reviewing a PR:
1. click on Show All Checks,
2. click Details next to "Taskcluster (pull_request)" after it appears and then finishes with a green checkmark,
3. click on the "Fenix - assemble" task, then click "Run Artifacts".
4. the APK links should be on the left side of the screen, named for each CPU architecture

PR #20813, opened 2021-08-12
For #17917: Use View binding in Login Exceptions

codrut-topliceanu:
For https://github.com/mozilla-mobile/fenix/issues/17917

### Pull Request checklist

- [x] **Screenshots**: This PR does not include screenshots as no user visible changes have been made.

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/settings/logins/fragment/LoginDetailFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/logins/fragment/LoginDetailFragment.kt
@@ -44,7 +44,7 @@ import org.mozilla.fenix.settings.logins.controller.SavedLoginsStorageController
 import org.mozilla.fenix.settings.logins.createInitialLoginsListState
 import org.mozilla.fenix.settings.logins.interactor.LoginDetailInteractor
 import org.mozilla.fenix.settings.logins.togglePasswordReveal
-import org.mozilla.fenix.settings.logins.view.LoginDetailView
+import org.mozilla.fenix.settings.logins.view.LoginDetailsBindingDelegate
 
 /**
  * Displays saved login information for a single website.
@@ -56,7 +56,7 @@ class LoginDetailFragment : Fragment(R.layout.fragment_login_detail) {
     private val args by navArgs<LoginDetailFragmentArgs>()
     private var login: SavedLogin? = null
     private lateinit var savedLoginsStore: LoginsFragmentStore
-    private lateinit var loginDetailView: LoginDetailView
+    private lateinit var loginDetailsBindingDelegate: LoginDetailsBindingDelegate
     private lateinit var interactor: LoginDetailInteractor
     private lateinit var menu: Menu
     private var deleteDialog: AlertDialog? = null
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/mozilla/fenix/settings/logins/fragment/LoginDetailFragment.kt
```diff
@@ -77,7 +77,8 @@ class LoginDetailFragment : Fragment(R.layout.fragment_login_detail) {
             )
         }
         loginDetailView = LoginDetailView(
-            view.findViewById(R.id.loginDetailLayout)
+            view.findViewById(R.id.loginDetailLayout),
+            binding
```
Mugurell, line 81: This seems strange, but again this `LoginDetailView` is not really a View but rather a delegate to handle binding.

Mugurell (dismissed):

app/src/main/java/org/mozilla/fenix/exceptions/ExceptionsView.kt
```diff
@@ -21,21 +20,30 @@ abstract class ExceptionsView<T : Any>(
     protected val interactor: ExceptionsInteractor<T>
 ) : LayoutContainer {
```
Amejia481, line 21: nit: I think we don't need `LayoutContainer` as we are using view binding :)

codrut-topliceanu: I blame all of this on @Mugurell

Mugurell: Right. This is a `android.view.View/ViewGroup`but not a `kotlinx.android.extensions.LayoutContainer`.

Mugurell: We'll have to make sure that when the migration ends there is also no `kotlinx.android.extensions` import anywhere in the project.

Amejia481: Maybe I'm missing something :), but I think the import is from  `kotlinx.android.extensions.LayoutContainer`

image

Mugurell: @Amejia481 Agree. And when the migration ends there should't be any trace of this import in the project. Sorry, I think we were overlapping in comments.
I was saying that this is a `android.view.ViewGroup` (which receives the parent layout in which it inflates and attaches itself) and can be declared as such but is not a `kotlinx.android.extensions.LayoutContainer` so this is a something that should be refactored as part of this migration work.

codrut-topliceanu: Removing `LayoutContainer` breaks `TrackingProtectionExceptionsView`. Unsure what to do next.

Mugurell: Same changes I here, just making the `binding` property from `ExceptionsView` protected to be available in `TrackingProtectionExceptionsView` seems to work for me.

codrut-topliceanu: Then I'll have to wait for https://github.com/mozilla-mobile/fenix/pull/20788/ to merge

codrut-topliceanu:
Added `do-not-land` label to wait for trackingProtection PR to land. (WIP)

Mugurell (approved):
LGTM.

PR #20820, opened 2021-08-12
Ship nightlies to Google Play twice a day

bhearsum:
This was moved to once per day in https://bugzilla.mozilla.org/show_bug.cgi?id=1628413 when Google Play reviews were very slow. That hasn't been a problem for awhile, and the less frequent nightlies are causing delays getting new Android Components bumps shipped out.


jcristau (approved):
r+

jonalmeida:

PR #20836, opened 2021-08-13
Android Autofill: Use AppCompat theme for UnlockActivity.

pocmo:
I hope this is fixing https://github.com/mozilla-mobile/android-components/issues/10792. Unfortunately I was not able to reproduce this problem and it only happened for a very small set of users. However since this never happened for `AutofillConfirmActivity` I am quite confident that this is the right fix.


mcarare (approved):
LGTM!

pocmo:
@mergifyio backport releases_v92.0.0

Issue #19886, opened 2021-06-07
Unify the lock badge with a combined Connection and Tracking Protection menu


gabrielluong:
After adding the Home button to the browser toolbar, we want to explore unifying the lock badge with a combined Connection and Tracking Protection menu to not have the browser toolbar appear overloaded.

Amejia481:
QA team please help us to verify that this feature is working as expected.
We unified the tracking protection dialog with the lock icon, now only the lock icon should be visible.

![image](https://user-images.githubusercontent.com/773158/129793337-9c08ee19-7180-4f08-9a07-bf294419233a.png)


We want to test:

### 1) The connection panel change correctly when visiting  HTTP and HTTPS sites

#### 1.1 With a  HTTP should show -> Insecure connection 
#### 1.2 With a  HTTPS should show -> secure connection 


![image](https://user-images.githubusercontent.com/773158/129793643-5d4dedea-ff48-4fb5-a234-af4b087ca3b5.png)


#### 1.3 The connection sub menu should show the icon, title, URL, connection status  and the certificate info of the site.
![image](https://user-images.githubusercontent.com/773158/129794678-13790156-c346-4368-be89-9b09df6f3877.png)


### 2) Alter permissions on the dialog (allowed/blocked) 

![image](https://user-images.githubusercontent.com/773158/129794977-aa50c28f-f35e-46ad-b3e2-283ef7e3ebb2.png)


### 3) Under the "Details" section you should be able to see trackers information (blocked and allowed)

![image](https://user-images.githubusercontent.com/773158/129795298-85988c8d-0e42-4c8c-9c1d-d463eceb69d8.png)

LaurentiuApahideanSV:
Verified as fixed in Firefox Preview Nightly 2021-08-18. The lock submenus work according to the above screenshots.

Amejia481:
@LaurentiuApahideanSV thanks :)

PR #20874, opened 2021-08-16
For #19886 use block color on disabled permissions

Amejia481:
### Pull Request checklist

1. click on Show All Checks,


app/src/main/res/drawable/ic_autoplay_disabled.xml
```diff
@@ -8,9 +8,9 @@
     android:viewportWidth="24"
     android:viewportHeight="24">
     <path
-        android:fillColor="@color/disabled_text"
+        android:fillColor="?primaryText"
```
MozillaNoah, line 11: Does using a text color reference pose any risk being used for a vector? (e.g. the primary text color is updated by Design, but the disabled icon color stays the same) Or will the two colors always be the same?

gabrielluong (approved):

app/src/main/res/drawable/ic_autoplay_disabled.xml
gabrielluong: We are probably on the latter side of things. Most of our icons appear directly next to text which usually follow the same text colour. Of course, it will generally be great to define variables such as `primaryIconFillColor` and refactor everything, but that is probably beyond the scope here.

PR #20902, opened 2021-08-18
Closes  #20881 : Use armeabi-v7a version code for MozillaOnline builds

### LINKED ISSUE #20881: [Feature: China] Use one version code for MozillaOnline builds of different architectures
According to the requirement of application stores in China, the apks of different architectures (i.e. armeabi-v7a, arm64-v8a, x86, x86_64) with the same application id and the same version name should use the same build id (version code), otherwise we cannot upload multiple apks .

rxumoz:
Due to the requirement of the application stores in China, we need to use a unified version code for different architecture apks.
This patch chooses to use the version code of armeabi-v7a.


app/build.gradle
```diff
@@ -262,16 +262,31 @@ android.applicationVariants.all { variant ->
...
-            // channels that never shipped a Fennec build.
-            def versionCodeOverride = Config.generateFennecVersionCode(abi)
 
-            println("versionCode for $abi = $versionCodeOverride")
+        if (project.hasProperty("mozillaOnline") || gradle.hasProperty("localProperties.mozillaOnline")) {
+            variant.outputs.each { output ->
+                def abi = output.getFilter(OutputFile.ABI)
+                // We use the same version code generator, that we inherited from Fennec, across all channels - even on
+                // channels that never shipped a Fennec build.
+                def versionCodeOverride = Config.generateFennecVersionCode("armeabi-v7a")
```
Amejia481, line 271: It looks like this line is the only one that it is changing between `if` and `else` block. We could have variable and changes it's value to `abi`  or `"armeabi-v7a"` depending on if `project.hasProperty("mozillaOnline") || gradle.hasProperty("localProperties.mozillaOnline")` is `true`

Amejia481 (approved):

rxumoz:
@Mergifyio backport releases_v92.0.0

Issue #20952, opened 2021-08-20
Intermittent UI test failure - InstallException: INSTALL_FAILED_INSUFFICIENT_STORAGE


jonalmeida:
### Firebase Test Run: 

It doesn't seem like these tests are running on Firebase, afaict.

### Stacktrace:

<details>

```
2021-08-20T16:59:36.9569320Z > Task :app:connectedDebugAndroidTest
2021-08-20T16:59:36.9673560Z Unable to install /Users/runner/work/fenix/fenix/app/build/outputs/apk/debug/app-x86_64-debug.apk
2021-08-20T16:59:36.9771760Z com.android.ddmlib.InstallException: INSTALL_FAILED_INSUFFICIENT_STORAGE
2021-08-20T16:59:36.9780100Z 	at com.android.ddmlib.internal.DeviceImpl.installRemotePackage(DeviceImpl.java:1310)
2021-08-20T16:59:36.9782100Z 	at com.android.ddmlib.internal.DeviceImpl.installPackage(DeviceImpl.java:1136)
2021-08-20T16:59:36.9784040Z 	at com.android.ddmlib.internal.DeviceImpl.installPackage(DeviceImpl.java:1112)
2021-08-20T16:59:36.9785880Z 	at com.android.ddmlib.internal.DeviceImpl.installPackage(DeviceImpl.java:1101)
2021-08-20T16:59:36.9788290Z 	at com.android.build.gradle.internal.testing.ConnectedDevice.installPackage(ConnectedDevice.java:127)
2021-08-20T16:59:36.9790910Z 	at com.android.build.gradle.internal.testing.SimpleTestRunnable.run(SimpleTestRunnable.java:135)
2021-08-20T16:59:36.9793350Z 	at com.android.build.gradle.internal.tasks.Workers$ProfileAwareExecutorServiceAdapter$submit$submission$1.run(Workers.kt:165)
2021-08-20T16:59:36.9795370Z 	at java.base/java.util.concurrent.ForkJoinTask$AdaptedRunnableAction.exec(ForkJoinTask.java:1407)
2021-08-20T16:59:36.9796880Z 	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:290)
2021-08-20T16:59:36.9798500Z 	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1020)
[...]

PR #20965, opened 2021-08-23
For #20952 - Increase run-ui emulator storage to 2GB

Mugurell:
1. click on Show All Checks,

### DIFF
--- a/.github/workflows/build-contributor-pr.yml
+++ b/.github/workflows/build-contributor-pr.yml
@@ -124,6 +124,7 @@ jobs:
           target: ${{ matrix.target }}
           arch: x86_64
           profile: pixel_3a
+          sdcard-path-or-size: 3072M
           script:
             "JAVA_HOME=$JAVA_HOME_11_X64 && ./gradlew connectedDebugAndroidTest -Pandroid.testInstrumentationRunnerArguments.class=\
             org.mozilla.fenix.ui.NavigationToolbarTest#visitURLTest"


Mugurell:
Seems like 2GB wasn't enough. Trying with 3GB.

Mugurell:
Same `INSTALL_FAILED_INSUFFICIENT_STORAGE` error wven with 3GB though on a local emulator with just 2GB the ui tests run fine.
Maybe there's a deeper problem.

AaronMT:
Thank you for taking this. 

https://github.com/marketplace/actions/android-emulator-runner

`sdcard-path-or-size`

> Path to the SD card image for this AVD or the size of a new SD card image to create for this AVD, in KB or MB, denoted with K or M. - e.g. path/to/sdcard, or 1000M.

Might just need to denote `M` instead. Here's someones [example](https://github.com/owntracks/android/blob/a12842d2b9fe4d83976b171607ae5e25559d5546/.github/workflows/android.yml#L119) on Github. 

While in this file it's probably helpful to also add `disable-animations: true`

Mugurell:
I see the same "INSTALL_FAILED_INSUFFICIENT_STORAGE" error even though both now and previously at the beginning of logs it says `SD card path or size: 3072M`.
So the value was set. It just didn't help. Don't know where this VM is hosted. Maybe there's a problem there?

AaronMT:
Hello @Mugurell,

I think for now while we investigate let's disable `run-ui` as part of the workflow. The full UI-test suite are still ran normally on Firebase (`ui-test-x86-debug`).

https://github.com/mozilla-mobile/fenix/pull/20981

Issue #20762, opened 2021-08-09
[Bug]: Inactive tabs do not have a grey border 


jonalmeida:
### Steps to reproduce

1. Have an active tab and inactive tab in the tabs tray.
2. Observer the border around both items.

### Expected behaviour

There should be a grey border around the entire inactive section.

### Actual behaviour

Only active tabs have a grey border, the inactive tab section does not.

### Device name

All devices

### Android version

All versions

### Firefox release type

Firefox Nightly

### Firefox version

92.0a1

### Additional information

Observe no grey border around the inactive section in this screenshot:
<img width="527" alt="Screen Shot 2021-08-09 at 12 22 37 PM" src="https://user-images.githubusercontent.com/1370580/128744596-ed24024a-c0fd-4203-9c2c-0dad9eadd5d3.png">

cc: @topotropic

SoftVision-LorandJanos:
Verified as fixed on the latest Nightly (94.0a1 - 2021-10-04T05:10:50.531272).

The border is displayed in dark and light theme.

Devices used:

- Samsung Galaxy Tab S3 (Android 9).
- Huawei P30 (Android 10).
- Xiaomi Redmi 11 Lite (Android 11).

Closing the ticket as verified.

PR #20972, opened 2021-08-23
For #20762 - Adds grey border to inactive tabs

codrut-topliceanu:
For https://github.com/mozilla-mobile/fenix/issues/20762
### Pull Request checklist

- [x] **Tests**: This PR includes no tests as it's only a minor visual change.
- [x] **Screenshots**: This PR includes video of the changes made.

1. click on Show All Checks,

https://user-images.githubusercontent.com/60002907/130449453-a5f8fc80-5344-4ac3-9f47-3fa1720d5a79.mp4

**Notes:** I'm not 100% happy with this approach but the other ones I explored did not pan out to something convenient: 

- 🚫 Move `ConcatAdapter` into another adapter of its own that would contain a cardView "parent".
- 🚫 Add another viewHolder type to `InactiveTabViewHolder` that would be a "parent" card view.
- 🚫 Add create/add stroke to each viewHolder background from InactiveTabs.


app/src/main/res/layout/inactive_header_item.xml
```diff
@@ -1,46 +1,58 @@
...
     android:layout_marginTop="12dp"
     android:layout_marginEnd="16dp"
     android:background="@drawable/card_list_row_background"
-    android:clickable="false"
-    android:clipToPadding="false"
-    android:focusable="true"
-    android:foreground="?android:attr/selectableItemBackground"
-    android:paddingStart="16dp"
-    android:paddingEnd="16dp">
+    android:backgroundTint="@color/photonWhite"
```
jonalmeida, line 13: We probably don't want `photonWhite` but the same thing we use as the divider. Since we aren't adding elevation, we can't follow the collections design entirely.

cc: @topotropic

jonalmeida:
> **Notes:** I'm not 100% happy with this approach but the other ones I explored did not pan out to something convenient:

I agree, it's not super but I can't think of any other way to solve this today. 🤷

app/src/main/res/layout/inactive_header_item.xml
codrut-topliceanu: Thx for taking a look at this ❤️ 
Could you be more specific though? I just tried `neutral_faded_normal_theme` but it seems too faded:
image
image

jonalmeida: @codrut-topliceanu sorry I didn't respond to this sooner.

How about we use `photonLightGrey30`? This seems consistent with what we use today for the grid view tab borders: https://github.com/mozilla-mobile/fenix/blob/main/app/src/main/res/layout/tab_tray_grid_item.xml#L31-L32

jonalmeida (approved):
I know it's not pretty, but thanks for fixing this. 🙂

PR #20990, opened 2021-08-24
No issue: Update Flank to v21.08.1

AaronMT:
Flank https://github.com/Flank/flank/releases/tag/v21.08.1 has the fix for avoiding crash one of their API calls that we hit a couple times.


escapewindow (approved):

AaronMT:
need's a force merge here

PR #21069, opened 2021-08-30
Enable Android Autofill support in all build types.

pocmo:
This patch removes the feature flag and moves the manifest entries back to the main manifest.


st3fan (approved):
Looks good. Excited for this to land.

Issue #21102, opened 2021-09-01
Build with SDK for Android 12 / API 31


pocmo:
See A-C: https://github.com/mozilla-mobile/android-components/issues/10910

The [beta program for Android 12](https://developer.android.com/about/versions/12/overview) is in the "stability phase" now and the APIs are finalized. Now we can start using 31 as `compileSdkVersion` (not `targetSdkVersion` yet) and get blockers out of the way.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-17752)

Amejia481:
We can close this as it was already covered on https://github.com/mozilla-mobile/fenix/commit/099e463b4f8302fd4ab3bdc46b1e4ca277e00702

Amejia481:
Closed by https://github.com/mozilla-mobile/fenix/pull/25494

PR #21103, opened 2021-09-01
Issue #21102: Set compileSdkVersion to 31.

pocmo:
The API 31 SDK is available now and we can start building against it.

Let's see what our automation thinks. [In A-C everything seems to be green so far](https://github.com/mozilla-mobile/android-components/pull/10911).


pocmo:
This needs a forced toolchain task.

pocmo:
Also:

> w: /home/runner/work/fenix/fenix/app/src/main/java/org/mozilla/fenix/ext/String.kt: (84, 41): 'IP_ADDRESS: Pattern!' is deprecated. Deprecated in Java
w: /home/runner/work/fenix/fenix/app/src/main/java/org/mozilla/fenix/search/SearchDialogFragment.kt: (397, 17): 'toggleSoftInput(Int, Int): Unit' is

Issue #21045, opened 2021-08-27
FNXV2-17677


Mugurell:

sflorean:
Verified as fixed on Nightly 9/30 with:
- Pixel 3 (Android 12)
- Huawei MediaPad M2 (Android 5.1.1)

Issue #21043, opened 2021-08-27
FNXV2-17671


Mugurell:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-17718)

Mugurell:
@ QA This just had to use the appropriate endpoint to download stories. If we have results to show then this works as intended.

sflorean:
Verified as fixed on Nightly 9/30 with:
- Pixel 3 (Android 12)
- Huawei MediaPad M2 (Android 5.1.1)

PR #21104, opened 2021-09-01
For #21045 - Categories support

Mugurell:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
@@ -243,7 +243,7 @@ class HomeFragment : Fragment() {
         if (requireContext().settings().pocketRecommendations) {
             lifecycleScope.async(IO) {
                 val stories = components.core.pocketStoriesService.getStories()
-                homeFragmentStore.dispatch(HomeFragmentAction.PocketArticlesChange(stories))
+                homeFragmentStore.dispatch(HomeFragmentAction.PocketStoriesChange(stories))
             }
         }
 
--- a/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
@@ -47,7 +47,6 @@ import com.google.android.material.snackbar.Snackbar
 import kotlinx.coroutines.Dispatchers.IO
 import kotlinx.coroutines.Dispatchers.Main
 import kotlinx.coroutines.ExperimentalCoroutinesApi
-import kotlinx.coroutines.async
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.collect
 import kotlinx.coroutines.flow.map
@@ -111,6 +110,8 @@ import org.mozilla.fenix.home.sessioncontrol.DefaultSessionControlController
 import org.mozilla.fenix.home.sessioncontrol.SessionControlInteractor
 import org.mozilla.fenix.home.sessioncontrol.SessionControlView
 import org.mozilla.fenix.home.sessioncontrol.viewholders.CollectionViewHolder
+import org.mozilla.fenix.home.sessioncontrol.viewholders.pocket.DefaultPocketStoriesController
+import org.mozilla.fenix.home.sessioncontrol.viewholders.pocket.PocketRecommendedStoryCategory
 import org.mozilla.fenix.home.sessioncontrol.viewholders.topsites.DefaultTopSitesView
 import org.mozilla.fenix.onboarding.FenixOnboarding
 import org.mozilla.fenix.settings.SupportUtils
@@ -241,9 +242,12 @@ class HomeFragment : Fragment() {
         }
 
         if (requireContext().settings().pocketRecommendations) {
-            lifecycleScope.async(IO) {
-                val stories = components.core.pocketStoriesService.getStories()
-                homeFragmentStore.dispatch(HomeFragmentAction.PocketStoriesChange(stories))
+            lifecycleScope.launch(IO) {
+                val categories = components.core.pocketStoriesService.getStories()
+                    .groupBy { story -> story.category }
+                    .map { (category, stories) -> PocketRecommendedStoryCategory(category, stories) }
+
+                homeFragmentStore.dispatch(HomeFragmentAction.PocketStoriesCategoriesChange(categories))
             }
         }
 
@@ -327,6 +331,9 @@ class HomeFragment : Fragment() {
             ),
             historyMetadataController = DefaultHistoryMetadataController(
                 navController = findNavController()
+            ),
+            pocketStoriesController = DefaultPocketStoriesController(
+                homeStore = homeFragmentStore
             )
         )
 
--- a/app/src/main/java/org/mozilla/fenix/home/HomeFragmentStore.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/HomeFragmentStore.kt
@@ -63,7 +63,7 @@ data class HomeFragmentState(
     val recentTabs: List<TabSessionState> = emptyList(),
     val recentBookmarks: List<BookmarkNode> = emptyList(),
     val historyMetadata: List<HistoryMetadataGroup> = emptyList(),
-    val pocketArticles: List<PocketRecommendedStory> = emptyList()
+    val pocketStories: List<PocketRecommendedStory> = emptyList()
 ) : State
 
 sealed class HomeFragmentAction : Action {
@@ -89,7 +89,7 @@ sealed class HomeFragmentAction : Action {
     data class RecentTabsChange(val recentTabs: List<TabSessionState>) : HomeFragmentAction()
     data class RecentBookmarksChange(val recentBookmarks: List<BookmarkNode>) : HomeFragmentAction()
     data class HistoryMetadataChange(val historyMetadata: List<HistoryMetadataGroup>) : HomeFragmentAction()
-    data class PocketArticlesChange(val pocketArticles: List<PocketRecommendedStory>) : HomeFragmentAction()
+    data class PocketStoriesChange(val pocketStories: List<PocketRecommendedStory>) : HomeFragmentAction()
     object RemoveCollectionsPlaceholder : HomeFragmentAction()
     object RemoveSetDefaultBrowserCard : HomeFragmentAction()
 }
@@ -132,6 +132,6 @@ private fun homeFragmentStateReducer(
         is HomeFragmentAction.RecentTabsChange -> state.copy(recentTabs = action.recentTabs)
         is HomeFragmentAction.RecentBookmarksChange -> state.copy(recentBookmarks = action.recentBookmarks)
         is HomeFragmentAction.HistoryMetadataChange -> state.copy(historyMetadata = action.historyMetadata)
-        is HomeFragmentAction.PocketArticlesChange -> state.copy(pocketArticles = action.pocketArticles)
+        is HomeFragmentAction.PocketStoriesChange -> state.copy(pocketStories = action.pocketStories)
     }
 }
--- a/app/src/main/java/org/mozilla/fenix/home/HomeFragmentStore.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/HomeFragmentStore.kt
@@ -15,7 +15,10 @@ import mozilla.components.lib.state.State
 import mozilla.components.lib.state.Store
 import mozilla.components.service.pocket.PocketRecommendedStory
 import org.mozilla.fenix.components.tips.Tip
+import org.mozilla.fenix.ext.getFilteredStories
 import org.mozilla.fenix.historymetadata.HistoryMetadataGroup
+import org.mozilla.fenix.home.sessioncontrol.viewholders.pocket.POCKET_STORIES_TO_SHOW_COUNT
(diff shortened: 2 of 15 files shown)


Mugurell:
Based my work on the not yet merge #21043 commit, that's why I still have it in history. Will rebase after that is merged.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlAdapter.kt
```diff
@@ -341,7 +341,13 @@ class SessionControlAdapter(
 
     override fun onViewRecycled(holder: RecyclerView.ViewHolder) {
         when (holder) {
-            is PocketStoriesViewHolder -> holder.composeView.disposeComposition()
```
Mugurell, line 344: I was seeing as much as 300ms needed for `composeView.onAttachedToWindow` everytime the holder was scrolled off an on screen again.
I don't think we need the composable destroyed and keeping it around definitely means a smoother UX.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesComposables.kt
```diff
@@ -117,23 +127,175 @@ fun PocketStories(
...
+            it.measure(constraints)
+        }.toMutableList()
+
+        fun getIndexOfNextPlaceableThatFitsRow(available: List<Placeable>, currentWidth: Int): Int {
+            return available.indexOfFirst {
+                currentWidth + it.width <= constraints.maxWidth
+            }
+        }
+
+        // Populate each row with as many items as possible combining wider with narrower items.
```
Mugurell, line 244: Without this there would be a lot of unused space.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesComposables.kt
```diff
@@ -117,23 +127,175 @@ fun PocketStories(
...
+    ) {
+        Row {
+            Text(
+                text = category.name,
+                modifier = Modifier.alignByBaseline(),
+            )
+            Icon(
+                painter = painterResource(id = R.drawable.mozac_ic_check),
+                contentDescription = "Expand or collapse Pocket recommended stories",
+                modifier = Modifier.alignByBaseline()
```
Mugurell, line 202: The baselines still don't seem to be aligned. Not sure about the final design so I didn't worry much about it. If needed we'll be able to quickly spin up a custom view for this.

app/src/main/java/org/mozilla/fenix/home/HomeFragmentStore.kt
```diff
@@ -144,6 +153,48 @@ private fun homeFragmentStateReducer(
                     }
             )
         }
-        is HomeFragmentAction.PocketArticlesChange -> state.copy(pocketArticles = action.pocketArticles)
+        is HomeFragmentAction.PocketStoriesChange -> state.copy(pocketStories = action.pocketStories)
+        is HomeFragmentAction.PocketStoriesCategoriesChange -> state.copy(
+            pocketStoriesCategories = action.storiesCategories
+        )
+        is HomeFragmentAction.PocketStoriesCategoryClicked -> {
```
Mugurell, line 160: The event is hoisted from the lowest composable to the ViewHolder where I first had this logic and from where I'd just emit an action with updated data which the Reducer would update in Sate.
But a View Holder shouldn't control the functionality, just bind the Views.
So in the spirit of MVI and everything being a flow of data I went with having the reducer ultimately deciding how the data is updated and updating this in State.
Going further without using Interactors and Controllers maybe this is the architecture to use?

Mugurell: Talked with Sebastian about this who said that this is not the responsability of the reducer.
Will move this to the Fragment.

Mugurell: Moved.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -243,7 +246,23 @@ class HomeFragment : Fragment() {
...
+                val categoriesMapping = mutableMapOf<String, MutableList<PocketRecommendedStory>>()
+                stories.forEach { story ->
+                    categoriesMapping[story.category] = categoriesMapping
+                        .getOrPut(story.category) { mutableListOf() }
+                        .also { it.add(story) }
+                }
+                val categories = categoriesMapping.map {
+                    PocketRecommendedStoryCategory(name = it.key, stories = it.value)
+                }
+
```
pocmo, line 259: This took a while to understand. If this is grouping the stories by category in a `PocketRecommendedStoryCategory` instance then the `groupBy` operator may make this shorter. Something like:

```Kotlin
val categories = components.core.pocketStoriesService.getStories()
    .groupBy { story -> story.category }
    .map { (category, stories) -> PocketRecommendedStoryCategory(category, stories) }
```

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -243,7 +246,23 @@ class HomeFragment : Fragment() {
         if (requireContext().settings().pocketRecommendations) {
             lifecycleScope.async(IO) {
```
pocmo, line 247: nit: Unrelated to this PR, but wonder why we are using `async()` here since we do not capture any result. This could be `launch`?

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -1128,6 +1148,58 @@ class HomeFragment : Fragment() {
...
+                    it.copy(
+                        isSelected = !it.isSelected,
+                        lastInteractedWithTimestamp = System.currentTimeMillis()
+                    )
+                }
+                else -> it
+            }
+        }
+
+        homeFragmentStore.dispatch(HomeFragmentAction.PocketStoriesCategoriesChange(updatedCategories))
```
pocmo, line 1192: The actual creation of the new state (the new categories) should preferably happen in the reducer. That way you are guaranteed to always end up with a consistent state. Imagine the processing being slow and the user clicking categories fast: You may generate states here while the new state is not available in the store yet, which will cause inconsistent states. This doesn't not happen in the reducer where the actions will be processed sequentially.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -1149,5 +1221,102 @@ class HomeFragment : Fragment() {
...
+ * Either the stories from the [POCKET_STORIES_DEFAULT_CATEGORY_NAME] either
+ * filtered stories based on the user selected categories.
+ *
+ * @param allCategories the list of all stories Pocket stories categories.
+ * @param selectedCategories the list of the currently user selected categories.
+ * @param neededStoriesCount how many stories are intended to be displayed.
+ * This only impacts filtered results guaranteeing an even spread of stories from each category.
+ */
+@VisibleForTesting
+internal fun getFilteredStories(
```
pocmo, line 1241: It looks like those helpers here are independent from the `HomeFragment` class. Maybe we can move them somewhere else. There's a high risk that the `HomeFragment` will grow a lot. :)

pocmo (commented):
First pass. Haven't read all of it yet. Especially the grid code not yet.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
Mugurell: Definitely!

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
Mugurell: Maybe into a `Controller` 😅? 
Was thinking about an intermediate big interface for all domain logic needed for home composables, something like `HomeComposablesLogic`?(please suggest a better name) with a default implementation that can be delegated to by the HomeFragment? 
This just while migrating current views to composables, though even after fully migrating this screen to Compose, if bringing back all this logic in one place will have that again being immense?

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
Mugurell: That looks a lot better. Thank you!

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesComposables.kt
Mugurell: In the final design this icon won't be needed. So not a priority to work here perfectly.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
Mugurell: Remembered that I tried that first but the reducer seemed a bit big and it seemed to now contain the logic on for how to handle categories being clicked so I then moved that code in composable's parent - this fragment.
https://github.com/mozilla-mobile/fenix/pull/21104#discussion_r700212421

Mugurell:
Sorry for not keeping this up-to-date with main.
Rebase now.
Don't really know where to put that code from HomeFragment though..

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
Mugurell: I think reducers and state updates happen only on the main thread?
If so, instead of me launching coroutines to calculate the updated state and dispatch that I could have all this running on main.
That would avoid reducers becoming too complex while the new state can be calculated a bit closer to the user interaction.

pocmo: I think actions like "SelectCategory" or "DeselectCategory" make totally sense to me - and I don't think it's a problem if the reducer is complex, if the calculation of the new state is complex (as long it's still a function that just transforms input into an output without side effects). If I remember correctly the first approach had something like a "Clicked" action and that felt wrong because that is too close to the UI layer, the fragment handles the click and would dispatch an actual state changing action (not an UI event like "click").

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
pocmo: I don't have a strong opinion on the naming. It just seems like some utility functions (or maybe they could even be extension functions?) that could move to some related utilities file. Just to keep the fragment focused.

pocmo (approved):
Approving, but think the actual state creation should move into the reducer - that's it's job. :)

Mugurell:
- created a new `HomeFragmentState.getFilteredStories()` that could be called from anybody with access to the state, incl the fragment or the reducer
- HomeFragment will now emit `SelectPocketStoriesCategory` and `DeselectPocketStoriesCategory` for the reducer to actually update the selection in BrowserState.

Mugurell:
```
 SUITE: org.mozilla.fenix.share.ShareControllerTest
   TEST: handleShareToApp should start a new sharing activity and close this
     I/glean/Dispatchers: Task queued for execution and delayed until flushed
   FAILURE
 io.mockk.MockKException: Can't instantiate proxy for class kotlin.Function1
```

Restarting CI.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesComposables.kt
```diff
@@ -169,19 +179,174 @@ fun PocketStories(
                 Column(
                     Modifier.padding(end = if (index == halfStoriesIndex) 0.dp else 8.dp)
                 ) {
-                    PocketStory(item, client)
+                    key(item.title) {
+                        PocketStory(item, client)
+                    }
```
pocmo, line 184: Why do we use `key()` here? Is it to avoid recomposition of `PocketStory` unless the title changes? Why the title only though? 🤔

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesController.kt
```diff
@@ -0,0 +1,65 @@
...
+                            false -> category
+                        }
+                    }
+            } else {
+                null
+            }
+        oldestCategoryToDeselect?.let {
+            homeStore.dispatch(HomeFragmentAction.DeselectPocketStoriesCategory(it.name))
+        }
+
```
pocmo, line 61: Still think this should happen in the reducer since you are calculating a new state based on the current state here.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesComposables.kt
Mugurell: The title seemed like a small thing to check and enough to protect against recompositions. The stories are static in nature with no property changing. 
Tested a bit now, don't see any recomposition. Same calls with or without `key`. Will remove it.

Mugurell:
The review request from `mozilla-mobile/performance` was from me suppressing `penaltyDeath.` Updated the PR to remove that change.

PR #21107, opened 2021-09-01
Force rebuid Docker image and print java version in test tasks

pocmo:
🥊


csadilek (approved):

csadilek:
We're doing this to upgrade java 11 and also print its version. This is part of debugging our intermittent problems with mockk tests.

PR #21110, opened 2021-09-01
Do not run unit tests in parallel

csadilek:
This is to investigate the intermittent mockk class generation/loading issues. Since we can not reproduce locally and the failures are intermittent they could be caused by us running unit tests in parallel.


pocmo (approved):

csadilek:
OK this only increased unit test run time from 14 to 19 mins. I will run one more time, but then we can land this.

Issue #20654, opened 2021-08-03
Handle switching between private and normal mode when we show home behind search


eliserichards:
Followup to #20573 

<img width="323" src="https://user-images.githubusercontent.com/43795363/128077998-1932fa94-4a89-486f-ac32-ce5b2a226561.png">

I am on wikipedia, and I tap on the URL and see home in the background. If I tap on the private mode button (mask) at the top of the home screen, what is the expected behavior?

Some options…

1. Open a new private tab -> ~will be opened with the same URL,~ keep the original tab in the normal tabs list
~2. Open a new private tab will be opened with the same URL, close the original tab in the normal tabs list (aka “move” the tab to private mode)~
~3. The button shouldn’t be clickable (I don’t think this is really a good option, just listing it for posterity)~

---

Because we want to preserve private mode security, we don't want to "keep" the URL when switching between private and normal mode. Instead, we are opening a new tab with a fresh session in the selected mode. 

In addition, once we switch to the new tab we start a fresh backstack and the user cannot navigate with the back button to the last tab.

Once the new tab is opened, there is no navigation until the user enters a URL/search and hits GO.

Scenarios: 
1. Start on home (not a tab) -> private mode toggle works normally and doesn't open tabs when switching
2. Start on home -> open search dialog (but do not search, just click on url bar) -> private mode toggle works normally and doesn't open tabs when switching
3. Start on home -> navigate to a URL -> click on URL to see home behind the search dialog -> click private mode button does the following:
   * Open a new tab in the new mode (if you were on normal mode, it opens a new private tab)
   * The original tab still exists
   * URL bar is cleared
4. Start on home -> perform a search (google, DDG, etc) -> click on URL to see search results -> delete search text to see Home -> click private mode button does the following:
   * Open a new tab in the new mode (if you were on normal mode, it opens a new private tab)
   * The original tab with search results still exists
   * URL bar is cleared
5. Start on a URL -> click on URL -> toggle private mode -> click Back button/minimize keyboard should do the following: 
   * Stay on the new tab you've opened
   * Home is visible
   * Private mode button works as normal (in scenario 1)
6. Open a tab -> perform a search (google, DDG, etc) -> click on URL -> toggle private mode -> click Back button/minimize keyboard should do the following: 
   * Stay on the new tab you've opened
   * Home is visible
   * Private mode button works as normal (in scenario 1)

vesta0:
[...]

PR #21115, opened 2021-09-01
For #20654: allow switching between private mode behind search dialog

eliserichards:
For #20654 

This will allow us to toggle private mode using the homescreen mask button while home is visible behind the search dialog.

Two scenarios: (see more in comments below and on main issue)
1. Start on home (not a tab) -> private mode toggle works normally and doesn't open tabs
2. Start on a tab -> click on URL to see home behind the search dialog -> click private mode button does the following: 
   * Open a new tab in the new mode (if you were on normal mode, it opens a new private tab)
   * The original tab still exists
  
### Pull Request checklist

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/home/HomeFragmentStore.kt
```diff
@@ -43,13 +43,14 @@ data class Tab(
  * @property expandedCollections A set containing the ids of the [TabCollection] that are expanded
  *                               in the [HomeFragment].
  * @property mode The state of the [HomeFragment] UI.
- * @property tabs The list of opened [Tab] in the [HomeFragment].
```
eliserichards, line 46: Just updating these docs since I was looking at this store

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlController.kt
```diff
@@ -561,4 +569,35 @@ class DefaultSessionControlController(
...
+            fragmentStore.dispatch(
+                HomeFragmentAction.ModeChange(Mode.fromBrowsingMode(newMode))
+            )
+        }
+
+        // If we are changing the mode using the homescreen mask toggle
+        // from behind the search dialog, open the url in a new tab with the new mode
+        val fromHome = navController.previousBackStackEntry?.destination?.id == R.id.homeFragment
+        val onSearch = navController.currentDestination?.id == R.id.searchDialogFragment
+        val stillOnATab = store.state.selectedTab?.engineState?.engineSession != null
```
eliserichards, line 592: Is there an easier/better way to do these checks?

eliserichards:
@mergifyio rebase

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlController.kt
```diff
@@ -561,4 +569,35 @@ class DefaultSessionControlController(
...
+
+        if (fromHome && onSearch && stillOnATab) {
+            store.state.selectedTab?.content?.url?.let { url ->
```
csadilek, line 595: We shouldn't copy the `content.url` of another tab here esp. if we're switching from private to normal mode. I know this isn't loading the tab yet, but we should also not copy the URL or any other state.

It should be good enough to do this and initialize the new tab with `about:blank`:
```kotlin
if (store.state.selectedTab != null) {
  addTabUseCase.invoke(
    private = isPrivateMode
  )
}
```

csadilek (commented):
👍 this works well! I left one comment below of a change we have to make re: copying URLs and one to clarify the `stillOnTab` check.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlController.kt
eliserichards: Ahh that totally makes sense. I didn't realize we could do this without having the URL 👍

eliserichards:
I think I was confused the first time I read your comments @csadilek.The functionality that we wanted here was to open the same URL in the opposite mode (which involves copying the URL). 

Re: "We shouldn't copy the content.url of another tab here" and "Is stillOnTab a leftover maybe or still covering a case?":
Are you proposing that we open a new, blank tab if the Private Mode button is pressed? The `stillOnTab` case is determining if we should copy the url into the new tab. Otherwise it just opens a blank new tab as per usual

eliserichards:
Other use case: 
Open tab -> toggle -> minimize keyboard -> see about:blank

should this be the last tab we visited?

eliserichards:
Ok this is ready @csadilek. I wrote down all of the scenarios we were testing in the description on https://github.com/mozilla-mobile/fenix/issues/20654.

csadilek (approved):
👍 🚢

PR #21123, opened 2021-09-02
Rollback Flank to v21.08.1

AaronMT:
Flank v21.09.0 introduced a XML tree change to `JUnitReport.xml` which breaks our daily reporting. While I figure out this new schema change, I don't want to break reporting. This was introduced with a [forced Docker update](https://github.com/mozilla-mobile/fenix/pull/21107) which tagged this Docker and pulled in a new version. Pinning an older version here.

https://api.github.com/repos/Flank/flank/releases/48276753 -> Flank v21.08.1


rpappalax (approved):
LGTM

PR #21141, opened 2021-09-03
Issue mozilla-mobile#21140 - Updated recent tab logic to show media tab and second-to-last tab, if the media tab was the last active tab.

### LINKED ISSUE #21140: Always show both last media tab AND last non-media tab
If I am playing media on my phone, I always see the last media tab on my homescreen in the Jump back in section. 

If I have non-media tabs open, I always see the last non-media tab on my homescreen in the Jump back in section. 

(This is a change from always showing the last tab only regardless of type)

MozillaNoah:
Fixes #21140 

### Pull Request checklist

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/ext/BrowserState.kt
```diff
@@ -35,6 +38,15 @@ fun BrowserState.asRecentTabs(): List<TabSessionState> {
 val BrowserState.lastOpenedNormalTab: TabSessionState?
     get() = selectedNormalTab ?: normalTabs.maxByOrNull { it.lastAccess }
 
+/**
+ *  Get the second-to-last accessed normal tab.
+ */
+val BrowserState.secondToLastOpenedNormalTab: TabSessionState?
```
jonalmeida, line 44: Drive-by 🚗 comment: 

(I didn't know if there was a JIRA issue to confirm if the description on github was taken from there, but I'll assume so.)

Based on the description, we need to always show the last media tab in the store which could even be the third, fourth, etc.

Maybe what we need is the sorted list and then go down the list until we find a media tab instead. I see below that we have a `BrowserState.inProgressMediaTab` which should be able to do that.

I might have overlooked something as well. :)

MozillaNoah: Thanks for checking out my PR!

There is a JIRA issue, I just wasn't sure if I needed to link it, or copy/paste it's current description into the GitHub issue.

And you're totally right on the description. The way I have it coded now, it will always show the last media tab AND the last-accessed non-media tab. Prior to my change, if the last-accessed tab was the media tab, it would ONLY show the media tab. My change here is to just look-up the second-to-last tab, for this scenario, so the media tab and the next-recent tab will show.

jonalmeida: > There is a JIRA issue, I just wasn't sure if I needed to link it, or copy/paste it's current description into the GitHub issue.

A link works too for internal use. The description may be updated in JIRA but it's a one-way sync so it wouldn't show up in Github too.

> And you're totally right on the description. The way I have it coded now, it will always show the last media tab AND the last-accessed non-media tab. Prior to my change, if the last-accessed tab was the media tab, it would ONLY show the media tab. My change here is to just look-up the second-to-last tab, for this scenario, so the media tab and the next-recent tab will show.

This makes sense. I think I misunderstood the logic from the naming. 🙂

jonalmeida (approved):

Issue #21138, opened 2021-09-03
Re-enable verifyRunBlockingAndStrictModeSuppresionCount test


mcomella:
It was ignored in https://github.com/mozilla-mobile/fenix/pull/20841

mcomella:
Closed with https://github.com/mozilla-mobile/fenix/pull/21142

Issue #20386, opened 2021-07-15
[Bug]: UI tests unexpectedly not run if CI if outside the `ui` package


mcomella:
_edit: see https://github.com/mozilla-mobile/fenix/issues/20386#issuecomment-893676828 for an updated theory and explanation._

### Steps to reproduce

- Run unit tests locally, specifically `ManufacturersCodeTest`, `LocaleViewHoldersTest`, or `AppRequestInterceptorTest`
- Run UI tests locally, specifically `StartupExcessiveResourceUsageTest`

### Expected behaviour

Tests pass: if CI is running all the tests and _correctly reporting errors_ then either:
- Master should have X next to the commits for the failing tests (they don't)
- Local builds should be passing (they're not)

### Actual behaviour

Some tests fail (particularly the ones named)

### Device name

P2

### Android version

?

### Firefox release type

Firefox Nightly

### Firefox version

?

### Additional information

I'm on commit from up/master 1bf89b6b014957331be61297232acced8efe0890

mcomella:
I looked through the logs for a PR https://github.com/mozilla-mobile/fenix/pull/20176 I have written.

The `test-report` artifact unexpectedly only lists one UI test as having been run: `org.mozilla.fenix.ui.NavigationToolbarTest`. [test-report.zip](https://github.com/mozilla-mobile/fenix/files/6826444/test-report.zip)

The unit tests identified the test failure I left in from my PR as I expected but misattributed the failure of `DefaultRecentBookmarksControllerTest` to `DefaultSessionControlControllerTest` – strange.

mcomella:
I have a theory that the UI tests are only running on the `org.mozilla.fenix.ui` package, when they should be running on all of the subpackages in `org.mozilla.fenix`.

To test this, I created a PR where I moved the start up test that fails locally from the `omf.perf` directly to `omf.ui`. Backing up my theory, the test failed: https://github.com/mozilla-mobile/fenix/pull/20695

mcomella:
This configuration file looks like it might be the one that defines what tests get run: https://searchfox.org/mozilla-mobile/rev/7456e758b207cd608863dba9ff1191383fd2d44f/fenix/automation/taskcluster/androidTest/flank-x86.yml#41

Currently, it's set up as an allowlist of `androidTest` subdirectories:
```yaml
  test-targets:
   - package org.mozilla.fenix.ui
   - package org.mozilla.fenix.glean
```

[...]

PR #21142, opened 2021-09-03
For #21138: re-enable excessive resource use test.

mcomella:
This test wasn't running in CI
https://github.com/mozilla-mobile/fenix/issues/20386 so we changed the
CI configuration so that it will. However, the test was then failing so
this is the revision that unignores the test.

I wonder if componentInit count is useful - it seems like it'd cause
more false positives than not - but I figure we can leave it in and see
how it goes.

### Pull Request checklist

1. click on Show All Checks,


MarcLeclair (approved):
Looks good to me. Hopefully we can keep track of it this time 👍

mcomella:
The test failure was:
```
 java.lang.AssertionError: runBlockingIncrement count does not match expected count.
...
expected:<0> but was:<1> 
```

Unfortunately, I can't reproduce this failure on either of my devices. That implies this test takes a different code path in Firebase UI testing than it does locally. I'm concerned this could also happen on other physical devices which would make this test confusing to debug in the future. That being said, if we really want this to run, perhaps we can specify a range instead, e.g. 0-1 failures is okay. And if we need to increment, we could increment both ends of the range.

mcomella:
@MarcLeclair I pushed a fix for the test failure: please re-review.

MarcLeclair (approved):
👍

mcomella:
It passed so let's land.

Issue #20579, opened 2021-07-29
[Bug]: Tabs unexpectedly re-used in normal browsing mode if "open links in a private tab" is enabled


mcomella:
### Steps to reproduce

- Clear app data
- Launch app
- Enable Settings -> Private Browsing -> Open links in a private tab
- Hit back to return to browser homescreen
- Load a website (e.g. apple.com)
- Add a homescreen shortcut for it in the 3-dot menu. Add the shortcut automatically
- Click the homescreen shortcut

### Expected behaviour

Website is launched in a private tab

### Actual behaviour

Website is opened in the existing normal browsing mode tab.

I have not tested if this would still occur if the tab that created the homescreen shortcut was closed and reopened in normal browsing (i.e. maybe this is only a problem the first time). 

### Device name

P2

### Android version

API 28, API 30

### Firefox release type

Firefox Nightly

### Firefox version

main branch: e2af42a49768fc91423e5e872ca18c222463718b

### Additional information

The code to switch to private browsing seems like it may not be functioning as intended for a few reasons:
- In this process, [`activity.recreate` is called twice by the Theme Manager](https://searchfox.org/mozilla-mobile/rev/0268dc3af225c8843ed5497ebe916abee0be3e51/fenix/app/src/main/java/org/mozilla/fenix/theme/ThemeManager.kt#125) before the activity ever gets recreated
- ^ can only happen because the `currentTheme` gets set to PRIVATE (from the homescreen icon intent) _and then is unexpectedly set back to normal mode_ (from data stored in the app)
- To propagate PB mode, [we mutate the global `HomeActivity.intent`](https://searchfox.org/mozilla-mobile/rev/0268dc3af225c8843ed5497ebe916abee0be3e51/fenix/app/src/main/java/org/mozilla/fenix/theme/ThemeManager.kt#122-123) (which gets unset by the normal browsing event ^), which can have unintended consequences if that intent is re-used, rather than passing an intent explicitly like `finish(); startActivity(HomeActivity, newIntent)`
- On API 28, `HomeActivity.onCreate` gets called twice in response to this interaction. On API 30, it's called once.
- ^ Setting a boolean to ensure `activity.recreate` is only caused once causes the behavior to change such that the homescreen is shown again (i.e. even more incorrect).

I used breakpoints to figure out ^.

[...]

PR #21154, opened 2021-09-05
For #20579: Fix the normal browing tab re-used unexpectedly when open links in private tab enabled 

RotBolt:
### Pull Request checklist

1. click on Show All Checks,

I just removed the code for updating the theme session for the tab in `BaseBrowserFragment`. `updateThemeForSession` is the cause of why the activity was being re-created two times. 
- When we switch to private mode or normal mode. Theme switch is handled by `DefaultThemeManager`.
- But calling update theme session call in BaseBrowserFragment in `onResume`  does force recreate for activity.
- Reason for the bug is when onResume is called, BaseBrowserFragment tries to find the tab with `customtabSessionId`, in case it finds it it will update the theme for it  (a theme, in this case, is Normal or Private only). Now if a particular is already opened in Normal browsing mode, and when the user clicks on the `HomeScreenShortcut` for that particular tab only,  it tries to open the tab in Private at first, but since the tab is already opened in Normal browsing mode `findTabOrCustomTabOrSelectedTab` gets the Tab (or TabSessionState object). Now this tab returned by `findTabOrCustomTabOrSelectedTab` contains the `TabSessionState` whose mode is set as `Normal browsing mode`. Since this `TabSessionState` contains `Normal Browsing Mode` `updateThemeForSession` updates the theme and opens the tab in Normal browsing mode only.

Video for testing after the fix.
- Tab is now not used in normal browsing mode. 
- If tab is not already opened, then open in new private tab
- If tab is already opened, then uses that only

Since video size is large. Here is the viewable link: https://drive.google.com/file/d/1yNTDMYCAxIw4_kVnlKYxxThtJCgUCQtS/view?usp=sharing


RotBolt:
cc @eliserichards

eliserichards (approved):
This is so excellent! You did an amazing job tracking this down. Thank you for the write-up in the PR description and on the bug!

I did some manual testing with the scenarios you mentioned, as well as some common use cases and this looks good! Let's get this merged! 🎉

eliserichards:
@mergifyio rebase

RotBolt:
@eliserichards @jonalmeida I think CI is not working properly. It needs to be merged by someone from Mozilla
Could you please look into this?

eliserichards:
@Mergifyio rebase

csadilek (commented):
@RotBolt Thanks for taking this on. This code was originally introduced to fix https://github.com/mozilla-mobile/fenix/issues/11903 and has gone through a few refactorings.

Let's please test this behaviour is the same before and after your change here so we don't regress.

/cc @eliserichards

csadilek:
And I should add that the change looks good to me and does indeed look like we can remove this cod! Let's just test the custom tab use case above to double-check before landing.

RotBolt:
@csadilek and @eliserichards 
I cannot reproduce the issue of custom tabs. It seems to work fine on devices I tested It is showing the close icon.
Could you please test it ?

csadilek:
Thanks so much for testing @RotBolt. We plan to land this early next week after we cut the beta branch so this change can get some time in Nightly.

csadilek (approved):

PR #21167, opened 2021-09-07
For #21120 - ImageLoader with a solid color placeholder

Mugurell:
Theming ain't perfect. Will be implemented in a followup ticket.

### Pull Request checklist

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesComposables.kt
```diff
@@ -73,20 +81,35 @@ fun PocketStory(
...
-            // )
+            ImageLoader(
+                client = client,
+                // The endpoint allows us to ask for the optimal resolution image.
+                url = story.imageUrl.replace(
+                    "{wh}",
+                    with(LocalDensity.current) {
+                        "${160.dp.toPx().roundToInt()}x${87.dp.toPx().roundToInt()}"
+                    }
+                )
```
pocmo, line 94: You could consider setting `targetSize=...` here too.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesComposables.kt
```diff
@@ -73,20 +81,35 @@ fun PocketStory(
...
+                }
+
+                WithImage { painter ->
+                    Image(
+                        painter,
+                        modifier = Modifier.size(160.dp, 87.dp),
+                        contentDescription = "${story.title} story image"
+                    )
+                }
+            }
```
pocmo, line 108: nit: This whole block could almost be its own `PocketImage` kind of composable?

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesComposables.kt
```diff
@@ -273,3 +291,12 @@ private fun getFakePocketStories(limit: Int = 1): List<PocketRecommendedStory> {
...
 }
+
+private class FakeClient : Client() {
+    override fun fetch(request: Request) = Response(
+        url = request.url,
+        status = 200,
+        body = Response.Body.empty(),
+        headers = MutableHeaders()
+    )
+}
```
pocmo, line 302: I wonder how we can keep track of those fakes, or even provide some standard implementations. For one it would be nice to avoid having multiple of those implementations everywhere, and second, it feels a bit like they pollute the code base, or distract from the actual runtime code.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/Shimmer.kt
```diff
@@ -0,0 +1,77 @@
...
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+
+/**
+ * Simple animation for mimicking a shimmer effect in the parent bounds.
+ */
+@Composable
+@Preview
+@Suppress("MagicNumber")
+fun Shimmer() {
```
pocmo, line 32: Is this coming from UX?

pocmo (approved):

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesComposables.kt
Mugurell: 👍. I like it. 
Was talking with Nicole and Gabriel about sharing the composables with the "Jump back in" items and generally about making the Composables reusable so while these were more to test the basics work for the final design I'll work on breaking off individual Composables that can be reused and recombined for other features.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/Shimmer.kt
Mugurell: The idea yes. The exact specs no. Waiting for @topotropic confirmation based on the result posted on Jira.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesComposables.kt
Mugurell: I don't know of an idiomatic way to achieve this.
Only an ugly way of having a new package - `fakes`, maybe inside `composables` since these would be used in multiple Composables previews?

pocmo: Maybe we can even ship some as AC package. I was recently thinking about something similar for tests, sometimes some good fakes are nicer than mocking manually.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesComposables.kt
Mugurell: It seems that `targetSize` (or `DesiredSize`) works on the premise that width and height are equal?
And while this may have been true for favicons for other images we need to be able to specify the two separately?
In the interim I think we can go with `maxOf(..)`.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/Shimmer.kt
Mugurell: The changes seems bigger than expected. Updated the PR to remove the shimmer animation and default to solid colors as placeholders.

Mugurell:
> TEST: WHEN showUndoSnackbarForTab is called for a normal tab with new tab button not visible THEN an appropriate snackbar is shown
FAILURE
java.lang.InstantiationError: kotlin.jvm.functions.Function1

Restarting CI.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesComposables.kt
Mugurell: Went with using the width which is larger.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesComposables.kt
```diff
@@ -73,20 +87,53 @@ fun PocketStory(
...
+                    )
+                }
+
+                Placeholder {
+                    Box(
+                        Modifier.background(
+                            when (isSystemInDarkTheme()) {
+                                true -> Color(0xFF42414D) // DarkGrey30
+                                false -> PhotonColors.LightGrey30
+                            }
```
pocmo, line 117: nit: We use this twice here, do you expect this to be needed more often? Maybe we want to move this to some extension method or composable function or will this be a property on the theme? 🤔

pocmo (approved):

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesComposables.kt
Mugurell: I think this placeholder / fallback image approach will indeed be the default for future Images so it definitely needs to be standardized.
Maybe we can have a reusable `..Image` composable to which users would pass just the ImageLoader needed arguments and the placeholder / fallbacks are hardcoded.

Issue #21172, opened 2021-09-07
Replace BrowserAwesomeBar (browser-awesomebar) with AwesomeBar composable (compose-awesomebar)


pocmo:
A while ago we integrated the `AwesomeBar()` composable into Focus. Now, after https://github.com/mozilla-mobile/android-components/issues/10922, we are ready to integrate it into Fenix too. I would like to land this change at the beginning of the 94 Nightly cycle.

pocmo:
@AaronMT This is still an issue, right? I think there's a way to get the suggestion count. I'll take a look.

AaronMT:
Andi is using a workaround in https://github.com/mozilla-mobile/fenix/pull/21517

pocmo:
@AaronMT In https://github.com/mozilla-mobile/android-components/pull/11050 I am adding a tag to every suggestion (`mozac.awesomebar.suggestion`). With that I am able to find and count the suggestions:

```Kotlin
    fun ComposeTestRule.countSuggestions(): Int {
        val nodes = onAllNodesWithTag("mozac.awesomebar.suggestion")
        val matchedNodes = nodes.fetchSemanticsNodes(
            atLeastOneRootRequired = false
        )
        return matchedNodes.size
    }
```

And can also wait for them to appear:

```Kotlin
    fun ComposeTestRule.waitForSuggestions(timeoutMillis: Long = 1_000) {
        waitUntil(timeoutMillis) {
            countSuggestions() > 0
        }
    }
```

I guess that should work from an idle resource too, if we pass the `ComposeTestRule` to it.

pocmo:
Closing the issue since the integration PR already landed.

PR #21191, opened 2021-09-08
Issue #21172: Replace BrowserAwesomeBar view with AwesomeBar() composable.

pocmo:
This replaces the `BrowserAwesomeBar` (`browser-awesomebar`) with the new `AwesomeBar()` composable (`compose-awesomebar`). It's feature complete and looking and behaving as before.

To keep the changes minimal I introduced `AwesomeBarWrapper` which wraps the compose code and pretends to still be a `concept-awesomebar` `View` implementation. This way I didn't need to change any of the code interacting with the awesome bar for now.

I need to investigate how to refactor our UI tests. Some of them seem to look into the adapter for assertions and the compose code doesn't have an adapter.


pocmo:
I tried to fix the directly affected UI tests.

However I fail to understand why so many unrelated UI tests seem to fail here:
https://console.firebase.google.com/project/moz-fenix/testlab/histories/bh.66b7091e15d53d45/matrices/8140806191013136917

@rpappalax @sv-ohorvath Do you have any idea? It looks like Espresso is unable to click some of the views... but those areas are unrelated to the change.

pocmo:
Actually there might be still some valid broken cases, that I will try to find. There's such a wall of red that I don't understand why that would be related. :|

AaronMT:
Looking at the Firebase Test Lab (looking at a few Shards) there seems to be a common bump of entering the address bar causing a problem

```Log
at org.mozilla.fenix.ui.robots.NavigationToolbarRobotKt.openEditURLView(NavigationToolbarRobot.kt:263)
at org.mozilla.fenix.ui.robots.NavigationToolbarRobot$Transition.enterURLAndEnterToBrowser(NavigationToolbarRobot.kt:93)
```

https://github.com/mozilla-mobile/fenix/blob/main/app/src/androidTest/java/org/mozilla/fenix/ui/robots/NavigationToolbarRobot.kt#L95

Does this function just need to be tweaked?

pocmo:
Thank you! I'll take a look if/how the new awesomebar somehow affected that area.

mcarare:
> ```
> at org.mozilla.fenix.ui.robots.NavigationToolbarRobotKt.openEditURLView(NavigationToolbarRobot.kt:263)
> at org.mozilla.fenix.ui.robots.NavigationToolbarRobot$Transition.enterURLAndEnterToBrowser(NavigationToolbarRobot.kt:93)
> ```
> 
> https://github.com/mozilla-mobile/fenix/blob/main/app/src/androidTest/java/org/mozilla/fenix/ui/robots/NavigationToolbarRobot.kt#L95

This happens because the views are identified by using the view id.

pocmo:
> This happens because the views are identified by using the view id.

Which is fine for this part of the screen, I think. Note that the `awesomeBar()` helper may be a bit misleading since it's actually targeting the toolbar in edit mode (`R.id.mozac_browser_toolbar_edit_url_view`).

mcarare (approved):
LGTM!

pocmo:
Rebased and added `onScroll = { hideKeyboard() }`.

Issue #21192, opened 2021-09-08
Custom Composable theme


Mugurell:
For a clean and consistent UI built with Composables we need a new app wide theme component.
We currently have the following resources which needs to be integrated into the app.
cc @topotropic 

- Colors 
  - https://www.figma.com/file/DyIIHvRgqt2EXfAVn1gV9c/Fx-Colors?node-id=1%3A3
  - https://www.figma.com/file/TebIgESfWmQkMcEmwNaYZl/Android?node-id=0%3A498
  
- Typography
  - https://www.figma.com/file/TebIgESfWmQkMcEmwNaYZl/Android?node-id=0%3A806
  
- Elevation
  - https://www.figma.com/file/TebIgESfWmQkMcEmwNaYZl/Android?node-id=0%3A457
 
----------
TBD:
- Paddings (incl margins)
- Shapes

Mugurell:
This is superseded by other work lead by gl.

PR #21135, opened 2021-09-03
Issue #21099 - Update items under "Jump back in" section to the latest designs

gabrielluong:
Fixes #21099

### Pull Request checklist

- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features. In addition, it includes a screenshot of a successful [accessibility scan](https://play.google.com/store/apps/details?id=com.google.android.apps.accessibility.auditor&hl=en_US) to ensure no new defects are added to the product.

### To download an APK when reviewing a PR:
1. click on Show All Checks,
2. click Details next to "Taskcluster (pull_request)" after it appears and then finishes with a green checkmark,
3. click on the "Fenix - assemble" task, then click "Run Artifacts".
4. the APK links should be on the left side of the screen, named for each CPU architecture

PR #21193, opened 2021-09-08
For #21192 - New FenixTheme to style Compositions

Mugurell:
Starting with adding the complete list of Firefox colors from which custom
color palettes for light/dark/private mode are built and made available to
child Composable trees.

In an effort to keep the UI consistent these new color palettes do not allow
overriding individual properties though Composables have access to the entire
list of colors and can decide to use another as an exception.

### Pull Request checklist

1. click on Show All Checks,


Mugurell:
Just a draft for now.
Open for feedback while I add the values for typography and elevation.

app/src/main/java/org/mozilla/fenix/theme/FirefoxColors.kt
```diff
@@ -0,0 +1,140 @@
...
+    val Red20 = Color(0xFFFF9AA2)
+    val Red30 = Color(0xFFFF848B)
+    val Red40 = Color(0xFFFF6A75)
+    val Red50 = Color(0xFFFF4F5E)
+    val Red60 = Color(0xFFE22850)
+    val Red70 = Color(0xFFC50042)
+    val Red80 = Color(0xFF810220)
+    val Red90 = Color(0xFF440306)
+
+    // Violet is primarily used for ????? in user interface.
```
Mugurell, line 105: Reminder to update this documentation. Detail currently missing in Figma. 
@topotropic any wording suggestion here?

app/src/main/java/org/mozilla/fenix/theme/FenixColors.kt
```diff
@@ -0,0 +1,257 @@
...
+
+    isPrivate = true,
+    isDark = true
+)
+
+/**
+ * A Material [Colors] implementation which sets all colors to [debugColor] to discourage usage of
+ * [MaterialTheme.colors] in preference to [FenixTheme.colors].
+ */
+internal fun debugColors(
```
pocmo, line 227: We should not break the `MaterialTheme` colors. That's what components can access and use as their defaults - but they can't access this app-specific theme here.

app/src/main/java/org/mozilla/fenix/theme/FirefoxColors.kt
```diff
@@ -0,0 +1,140 @@
...
+    val LightGrey20 = Color(0xFFF0F0F4)
+    val LightGrey30 = Color(0xFFE0E0E6)
+    val LightGrey40 = Color(0xFFCFCFD8)
+    val LightGrey50 = Color(0xFFBFBFC9)
+    val LightGrey60 = Color(0xFFAFAFBA)
+    val LightGrey70 = Color(0xFF9F9FAD)
+    val LightGrey80 = Color(0xFF8F8F9D)
+    val LightGrey90 = Color(0xFF80808E)
+
+    // Dark grey should primarily be used for the Dark theme and secondary buttons.
```
pocmo, line 33: There's already `PhotonColors` provided by `ui-colors`.

app/src/main/java/org/mozilla/fenix/theme/FenixColors.kt
Mugurell: Google has a sample showcasing custom themes - https://github.com/android/compose-samples/tree/main/Jetsnack which I based this work on.
There are significant differences between the small color palette available in `MaterialTheme` and what Fenix uses. In this patch I am following the palette from https://www.figma.com/file/TebIgESfWmQkMcEmwNaYZl/Android?node-id=0%3A498

app/src/main/java/org/mozilla/fenix/theme/FirefoxColors.kt
Mugurell: I can move these to AC.

app/src/main/java/org/mozilla/fenix/theme/FenixColors.kt
Mugurell: Gabriel also saw the `MaterialTheme` being too limited in https://github.com/mozilla-mobile/fenix/pull/21135#discussion_r704045614

Mugurell: I used these `debugColors` just as in the sample to quickly find which components try to use `MaterialTheme` values to know to `migrate` them to our custom theme values.
Maybe we should do the same for typography?

pocmo: I am not opposed to using a custom theme. However A-C and third-party (material) libraries may use `MaterialTheme` colors as default?

Mugurell: That's a great point!
Maybe we should have these themes in AC to sync theme properties and values and ensure the same UX?
Then for third-party libs .. seems like a big caveat in how everything functions..
Maybe we can try to adapt some values from our theme to `MaterialTheme` (similar to how here I am using `Color.Red` for all) and hope that small subset of colors, fonts etc will not look weird?

app/src/main/java/org/mozilla/fenix/theme/FirefoxColors.kt
gabrielluong: This is actually in AC already as PhotonColors.

gabrielluong:
Wanted to provide a quick update that Crystal from UX is working on the design system for mobile at the moment, and the variable names that Bram has provided may end up changing during the work that is happening over the next few weeks.

Issue #21296, opened 2021-09-14
Add profiler markers for dispatchTouchEvent


mcomella:
It'd be helpful to see when the user interacts with the screen in profiles.

mcomella:
PR: https://github.com/mozilla-mobile/fenix/pull/21297

PR #21297, opened 2021-09-14
For #21296: add ProfilerMarkers.addForDispatchTouchEvent.

mcomella:
Here's a sample profile where I scroll down and scroll up in web content, then click a link: https://share.firefox.dev/394hFkd Note that the detail text is hidden due to a bug in the profiler: https://bugzilla.mozilla.org/show_bug.cgi?id=1730622. You can add it to this profile by running the following in the JS console: `fixed = 0; for (thr of profile.threads) { for (data of thr.markers.data) { if (data && data.type === "JavaWithDetails") { data.details = data.name; fixed +=1; } } }; fixed`

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/perf/ProfilerMarkers.kt
+++ b/app/src/main/java/org/mozilla/fenix/perf/ProfilerMarkers.kt
@@ -5,6 +5,7 @@
 package org.mozilla.fenix.perf
 
 import android.app.Activity
+import android.view.MotionEvent
 import android.view.View
 import android.view.ViewTreeObserver
 import androidx.core.view.doOnPreDraw
@@ -26,6 +27,20 @@ object ProfilerMarkers {
     fun homeActivityOnStart(rootContainer: View, profiler: Profiler?) {
         rootContainer.doOnPreDraw {
             profiler?.addMarker("onPreDraw", "expected first frame via HomeActivity.onStart")
+        }
+    }
+
+    fun addForDispatchTouchEvent(profiler: Profiler?, ev: MotionEvent?) {
+        // We only run this if the profiler is active to minimize any possible delay on touch events.
+        if (profiler?.isProfilerActive() == true) {
+            // We only do this subset because 1) other actions like MOVE may be spammy and 2) doing
+            // a generic ev?.action::class.simpleName may be too expensive for dispatchTouchEvent.
+            val detailText = when (ev?.action) {
+                MotionEvent.ACTION_DOWN -> "ACTION_DOWN"
+                MotionEvent.ACTION_UP -> "ACTION_UP"
+                else -> return
+            }
+            profiler.addMarker("dispatchTouchEvent", detailText)
         }
     }
 }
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/fenix/perf/ProfilerMarkers.kt
```diff
@@ -19,4 +20,18 @@ object ProfilerMarkers {
             profiler?.addMarker("onPreDraw", "expected first frame via HomeActivity.onStart")
         }
     }
+
+    fun addForDispatchTouchEvent(profiler: Profiler?, ev: MotionEvent?) {
+        // We only run this if the profiler is active to minimize any possible delay on touch events.
+        if (profiler?.isProfilerActive() == true) {
```
jonalmeida, line 26: Are there places when we want to call `addMarker` without this check first? Could we put this inside the component?

jonalmeida (approved):
Thanks for adding this as well. 🙂

app/src/main/java/org/mozilla/fenix/perf/ProfilerMarkers.kt
mcomella: Under the hood, `addMarker` checks the equivalent of `isProfilerActive` before adding the marker: https://searchfox.org/mozilla-central/rev/e98b9b6a20123cdb1d58ca683de123ce61f1e1a6/mobile/android/geckoview/src/main/java/org/mozilla/gecko/GeckoJavaSampler.java#354-357

`isProfileActive` is best used to avoid doing expensive operations needed to add the marker if the profiler isn't actually on. In this case, I guess I'm just avoiding the `when` statement, which probably isn't expensive so perhaps this wasn't worth it. 😅 I was originally concerned with how many `ACTION_MOVE` events are triggered.

mcomella: @jonalmeida What do you mean put it inside the component? Which component?

mcomella:
Rebased.

mcomella:
If there's a follow-up action needed for https://github.com/mozilla-mobile/fenix/pull/21297#discussion_r711374564, I'll do it in a follow-up PR (to avoid back-and-forth overhead).

app/src/main/java/org/mozilla/fenix/perf/ProfilerMarkers.kt
jonalmeida: > @jonalmeida What do you mean put it inside the component? Which component?

Sorry, I meant in the browser-gecko component where the [profiler is exposed](https://github.com/mozilla-mobile/android-components/blob/main/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/profiler/Profiler.kt) through, but your comment above shows that we no-op out early inside the runtime  already, and your logic to avoid extra operations is reasonable too. 🙂

PR #21321, opened 2021-09-15
Sort history metadata on home and allow limiting results

csadilek:
We should sort metadata groups based on `lastUpdated` time and not show more than 9 groups on home. This will later get more flexible once we add highlights.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/historymetadata/HistoryMetadataFeature.kt
+++ b/app/src/main/java/org/mozilla/fenix/historymetadata/HistoryMetadataFeature.kt
@@ -17,6 +17,8 @@ import org.mozilla.fenix.home.HomeFragmentAction
 import org.mozilla.fenix.home.HomeFragmentStore
 import kotlin.math.max
 
+private const val DEFAULT_MAX_RESULTS = 9
+
 /**
  * View-bound feature that retrieves a list of history metadata and dispatches updates to the
  * [HomeFragmentStore].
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mozilla/fenix/historymetadata/HistoryMetadataFeature.kt
```diff
@@ -17,6 +17,8 @@ import org.mozilla.fenix.home.HomeFragmentAction
 import org.mozilla.fenix.home.HomeFragmentStore
 import kotlin.math.max
 
+private const val DEFAULT_MAX_RESULTS = 9
```
gabrielluong, line 20: I am curious if this should be in the companion object?

gabrielluong: Nevermind, I think it's because we pass this into HistoryMetadataFeature

app/src/main/java/org/mozilla/fenix/historymetadata/HistoryMetadataFeature.kt
```diff
@@ -72,3 +79,6 @@ class HistoryMetadataFeature(
         job?.cancel()
     }
 }
+
+// The last updated time of the group is based on the most recently updated item in the group
+fun HistoryMetadataGroup.lastUpdated(): Long = historyMetadata.maxOf { it.updatedAt }
```
gabrielluong, line 84: Can we move this into HistoryMetadataGroup.kt instead?

gabrielluong (approved):

app/src/main/java/org/mozilla/fenix/historymetadata/HistoryMetadataFeature.kt
csadilek: 👍 yes, moved it over.

Issue #21291, opened 2021-09-14
~1+ s delay when touching the navigation bar on Pixel 5a (with long string in clipboard? WebURLFinder)


mcomella:
_Report via SUMO: https://support.mozilla.org/en-US/questions/1348730_

> Problem occurs immediately upon opening the navigation bar - the keyboard takes a few seconds to appear and input does not appear immediately in the bar. This occurs with both Gboard and SwiftKey. Keyboard display is also laggy, with the highlighted keys taking time to revert to standard appearance. 
> 
> Changing the search engine (tested Bing, DuckDuckGo, and Wikipedia) and disabling all suggestions/alternate input methods does not seem to affect behavior, lag is still present. Autocorrect suggestions appear with no lag and pressing enter/return immediately attempts to load the autofilled page or search, even if the text is not fully visible in the navigation bar.   
> 
> Problem is present using Firefox 91.3.1 and Nightly 93.0a1 on Android 11 on a Pixel 5a. 
> 
> This is a fresh installation of Firefox with no account synced, and I have cleared the cache and storage for the apps after encountering the problem.
> 
> Problem does not occur using Firefox Focus 91.3.0. The only search engines on Focus I tested were Google and DuckDuckGo, but I tested with and without search suggestions. Problem also does not occur with Google Chrome. 
> 
> Happy to help with additional tests and/or videos.

---

Follow-up comment with profiles:

> The navigation bar lag has happened again. 
> 
> I recorded a log in Nightly, with the following actions: 
> 
> 1. Touch address bar. 
> 2. Type in "test." 
> 3. Hold backspace to delete "test." 
> Waiting ~5 seconds in between tasks. 
> 
> https://share.firefox.dev/3k7a2zP
> 
> I rebooted my phone and tried again, this made the lag go away. I repeated the same actions. 
> 
> https://share.firefox.dev/3A6QMIc
> 
> It looks like there are large spikes in the AndroidUI category graphs.
> 
> If it also helps, I took a log without specific timings and typed more things into the search bar. https://share.firefox.dev/3k4ReRK

mcomella:
Triage: we're concerned this might affect ever user of the Pixel 5a. If this device is about to roll out, it's going to affect a lot of people.

csadilek:
[...]

PR #21334, opened 2021-09-16
Issue #21291: SearchDialogFragment: Get URL from clipboard once and not for every state update

pocmo:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/search/SearchDialogFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/search/SearchDialogFragment.kt
@@ -44,6 +44,7 @@ import mozilla.components.concept.storage.HistoryStorage
 import mozilla.components.feature.qr.QrFeature
 import mozilla.components.lib.state.ext.consumeFlow
 import mozilla.components.lib.state.ext.consumeFrom
+import mozilla.components.support.base.coroutines.Dispatchers
 import mozilla.components.support.base.feature.UserInteractionHandler
 import mozilla.components.support.base.feature.ViewBoundFeatureWrapper
 import mozilla.components.support.ktx.android.content.getColorFromAttr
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mozilla/fenix/search/SearchDialogFragment.kt
```diff
@@ -374,6 +376,23 @@ class SearchDialogFragment : AppCompatDialogFragment(), UserInteractionHandler {
...
 
+    override fun onResume() {
+        super.onResume()
+
+        lifecycleScope.launch(Dispatchers.Cached) {
+            // We are delaying the read of the clipboard a bit, because ClipboardManager will return
+            // null if the does app not have input focus and that seems to be happening here on
+            // non-first resumes.
+            @Suppress("MagicNumber")
+            delay(500)
```
pocmo, line 387: This delay is a bit ugly. But on resume after the app was already running I got `null` from `ClipboardManager` (except for the very first resume when the screen is displayed for the first time). Visually the delay is no problem, but I can't be fully sure it is good enough for all devices.

csadilek: @pocmo OK I was able to reproduce the intermittent `null` from `ClipboardManager` but it seems to also be fixable by doing:

```kotlin
view?.post {
  lifecycleScope.launch(Dispatchers.Cached) {
    store.dispatch(
      SearchFragmentAction.UpdateClipboardUrl(
        requireContext().components.clipboardHandler.url
      )
    )
  }
}
```

Wdyt?

csadilek (approved):
👍 I left one alternative suggestion to handle the intermittent `null` from `ClipboardManager` which also seems to work for me.

app/src/main/java/org/mozilla/fenix/search/SearchDialogFragment.kt
pocmo: This will queue the code to run shortly after the resume. Hm, if that works consistently then we may be able to launch the coroutine on the main thread and switch the context. I'll try that too.

pocmo: That didn't work reliable. But the `post` does. I'll update the PR and then land it.

PR #21343, opened 2021-09-16
Updating Fenix to work with the new logins API

bendk:
Switched to always using `Login` instead of the `SavedPassword` alias.

Moved the `SavedLoginsStorageController.kt.syncAndUpdateList` call
inside `add()` and `update()`.  This simplifies the error handling a
bit.

Refactored `findPotentialDuplicates()`:
  - Use getByBaseDomain() now that potentialDupesIgnoringUsername() is
    no longer available.
  - Put the common code in one method
  - Moved the coroutine to fetch the logins inside the main coroutine.
    I think this simplifies the cancellation handling, since the child
    coroutine should be cancelled when its parent is cancelled.

Updated the tests.  I wasn't exactly sure how that mockk code is
supposed to work, please double-check that I'm didn't break them.

The tests are passing, but I haven't done any manual testing because I'm not sure how to install the preview on my emulator yet.

### Pull Request checklist

1. click on Show All Checks,


bendk:
Actually, I realize this one isn't ready yet. There are a couple issues with startup and key management that need to be dealt with first.

PR #21359, opened 2021-09-17
SearchDialogFragment: Start to split consumeFrom() into independent observers that trigger less often

pocmo:
Currently `SearchDialogFragment` uses `consumeFrom()` and calls various methods for every state change. Since the state contains the typed text, it updates quite often and all those methods get called that often too. Some of them are not interested in the typed text and so we call them over and over for no real reason (for example the code that updates the list of search providers in the awesome bar).

In this patch I moved some of those calls out and invoke them from a separate observer using `consumeFlow()`. The awesome bar reacts much faster than Nightly, but that may be mostly because of the change from #21334.


csadilek (approved):
✨

Issue #21294, opened 2021-09-14
Fact.toEvent allocates a Pair for each comparison & is called many times: power use issue


mcomella:
[`Fact.toEvent`](https://searchfox.org/mozilla-mobile/rev/14b6f198ebbe966e5d31eb9ab5fbf6b15fc1ce75/fenix/app/src/main/java/org/mozilla/fenix/components/metrics/MetricController.kt#161) allocates a Pair for each comparison because `to` is an infix function that creates a new Pair. The comparison tree is 35 items long and I subjectively noticed that most events that go through it will hit the worst case, `else`. To test the impact, I did a simple workflow:
-     click wiki homescreen
-     click url bar, nav amazon.com
-     click tabs tray, new tab
-     search "foobar" and click default search result

I added custom profiler markers and took a profile: https://share.firefox.dev/3k6vox3 Individually, these methods are fast enough that the user won't notice the impact. However, in aggregate, it's called 116 times in this simple workflow and takes 79.2ms. **This is a non-trivial amount of work on the CPU that will impact power use** (though we don't have the intuition to say exactly how much): we should optimize this method.

For a specific to fix, the allocations don't seem necessary and, assuming they are what makes this particularly slow, it should be relatively straightforward to address. We could also try to ensure this isn't called so often.

mcomella:
I ran the AS allocation tracker with the following step:
- initial state: search screen with "a" typed
- type "s"

158 Pairs were allocated for a total of 2528 bytes (16 bytes per Pair – 8 byte / 64 bits per reference) on my Pixel 2. That doesn't seem _too_ bad – there are much larger allocations going on.

mcomella:
PR https://github.com/mozilla-mobile/fenix/pull/21368

mcomella:
Simpler PR: https://github.com/mozilla-mobile/fenix/pull/21446

PR #21368, opened 2021-09-18
For #21294: remove excess allocation in Fact.toEvent

mcomella:
There seem like there could be bigger wins in memory but this one is relatively straightforward to implement.

I tried to be careful to do this in step-by-step commits to avoid introducing regressions. If it's too risky, we can not land this (or write more tests but I'm not sure it's worth the time at that point).

Here's the improvement when opening search, typing "a", starting allocation tracking, and typing "p". Before:
Screen Shot 2021-09-17 at 16 41 41

After:
Screen Shot 2021-09-17 at 16 40 06

To be fair, these additional `Pair` allocations could have come from something else running at the time but it seems unlikely to be the root cause of the change.

_edit: I realized there is a change in logic. In the original case, if you had a component that matched but an item that did not, you would match against all 35 values. In the new version, you'll match against the component and only **its** items. I'm pretty sure the `when` statement will protect us against duplicate components so this logic would end up being functionally the same but I just thought I'd mention it. The best way to protect against an error being added from this is to add complete coverage, probably._

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/metrics/MetricController.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/metrics/MetricController.kt
@@ -368,35 +368,39 @@ internal class ReleaseMetricController(
             else -> null
         }
 
-        Component.FEATURE_AUTOFILL to AutofillFacts.Items.AUTOFILL_REQUEST -> {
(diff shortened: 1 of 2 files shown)


mcomella:
I found a better way to formulate this and will reopen a new patch.

PR #21370, opened 2021-09-18
Use UpdateUrlAction to record viewTime observations

grigoryk:
We discovered that in a tab restore scenario we were recording view time
observations that were wrong - we'd record time deltas as-if user was
looking at the page while the browser wasn't running.

This happens because when we record a viewTime observation, we compare
current time with lastAccess time of the tab. In a restore scenario,
that lastAccess time happens to be from when the browser was last
running - which could be days ago.

The simplest solution was to not record a viewTime observation if the
url for a tab didn't change during a load event. To achieve this, we
needed to change which action we were using as a proxy for "navigation
events" - UpdateUrlAction contains the new url, allowing us to compare
against the current tab url.

Alternative solutions would be to keep using loading actions, but
dispatch a lastAccess event before performing a metadata update. This
would have worked, but would result in two lastAccess events being
dispatched for each navigation event instead of just one.

### Pull Request checklist

1. click on Show All Checks,


grigoryk:
Note to go manually test (make sure we update viewTime when navigating away from pages) this before landing.

csadilek (approved):
👍 tested this change and verified it fixes the problem:

Found one more https://github.com/mozilla-mobile/android-components/issues/11004 causing the same inflated view time problem. Will put up a PR in A-C in a bit.

Issue #21275, opened 2021-09-14
FNXV2-17682


Mugurell:
https://mozilla-hub.atlassian.net/browse/FNXV2-17682

Mugurell:
This is a bit harder to test manually since there is no visual indication. Moving it to done.

Issue #21333, opened 2021-09-16
Intermittent debug(T) - org.mozilla.fenix.tabstray.TabsTrayFragmentTest - WHEN showUndoSnackbarForTab is called for a normal tab with new tab button not visible THEN an appropriate snackbar is shown


Archaeopteryx:
### Firebase Test Run:
https://treeherder.mozilla.org/logviewer?job_id=351703154&repo=fenix
### Stacktrace:
```
  TEST: WHEN onConfigurationChanged is called THEN it delegates the tray behavior manager to update the tray
    I/glean/Dispatchers: Task queued for execution and delayed until flushed
  SUCCESS

  TEST: WHEN showUndoSnackbarForTab is called for a normal tab with new tab button not visible THEN an appropriate snackbar is shown
    I/glean/Dispatchers: Task queued for execution and delayed until flushed
  FAILURE

java.lang.InstantiationError: kotlin.jvm.functions.Function1
	at jdk.internal.reflect.GeneratedSerializationConstructorAccessor214.newInstance(Unknown Source)
	at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:490)
	at org.objenesis.instantiator.sun.SunReflectionFactoryInstantiator.newInstance(SunReflectionFactoryInstantiator.java:48)
	at io.mockk.proxy.jvm.ObjenesisInstantiator.instanceViaObjenesis(ObjenesisInstantiator.kt:75)
	at io.mockk.proxy.jvm.ObjenesisInstantiator.instance(ObjenesisInstantiator.kt:42)
	at io.mockk.impl.instantiation.JvmInstantiator$instantiate$2.invoke(JvmInstantiator.kt:16)
	at io.mockk.impl.instantiation.AbstractInstantiator.instantiateViaInstanceFactoryRegistry(AbstractInstantiator.kt:17)
	at io.mockk.impl.instantiation.JvmInstantiator.instantiate(JvmInstantiator.kt:15)
	at io.mockk.impl.recording.states.RecordingState$matcher$signatureValue$1$1.invoke(RecordingState.kt:49)
	at io.mockk.impl.instantiation.JvmAnyValueGenerator$anyValue$2.invoke(JvmAnyValueGenerator.kt:35)
	at io.mockk.impl.instantiation.AnyValueGenerator.anyValue(AnyValueGenerator.kt:34)
[...]

PR #21277, opened 2021-09-14
For #21275 - Sort items by how many times they were actually shown

Mugurell:
Work based off of the not yet merged https://github.com/mozilla-mobile/fenix/pull/21104 which allows for the variations we need to ensure this code works.
Also needs https://github.com/mozilla-mobile/android-components/pull/10978.

Will mark this PR as draft until 21104 is merged.

To see the results: https://mozilla-hub.atlassian.net/browse/FNXV2-17682?focusedCommentId=458185

### Pull Request checklist

- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features. In addition, it includes a screenshot of a successful [accessibility scan](https://play.google.com/store/apps/details?id=com.google.android.apps.accessibility.auditor&hl=en_US) to ensure no new defects are added to the product.

### To download an APK when reviewing a PR:
1. click on Show All Checks,
2. click Details next to "Taskcluster (pull_request)" after it appears and then finishes with a green checkmark,
3. click on the "Fenix - assemble" task, then click "Run Artifacts".
4. the APK links should be on the left side of the screen, named for each CPU architecture

PR #21392, opened 2021-09-20
For #21391 - Updated composables

Mugurell:
This depends on the not yet merged https://github.com/mozilla-mobile/fenix/issues/21045 and https://github.com/mozilla-mobile/fenix/issues/21275.

Will ensure this PR it being kept up-to-date and rebased when necessary.

Draft for now while adding tests for composable but otherwise the code should be in final form.

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/compose/ImagesPlaceholder.kt
+++ b/app/src/main/java/org/mozilla/fenix/compose/ImagesPlaceholder.kt
@@ -1,0 +1,87 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.compose
+
+import androidx.compose.foundation.Image
+import androidx.compose.foundation.isSystemInDarkTheme
+import androidx.compose.foundation.layout.size
+import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.clip
+import androidx.compose.ui.graphics.painter.ColorPainter
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import mozilla.components.support.images.compose.loader.Fallback
+import mozilla.components.support.images.compose.loader.ImageLoaderScope
+import mozilla.components.support.images.compose.loader.Placeholder
+import mozilla.components.ui.colors.PhotonColors
+
+/**
+ * Renders the app default image placeholder while the image is still getting loaded.
+ *
+ * @param modifier [Modifier] allowing to control among others the dimensions and shape of the image.
+ * @param contentDescription Text provided to accessibility services to describe what this image represents.
+ * Defaults to [null] suited for an image used only for decorative purposes and not to be read by
+ * accessibility services.
+ */
+@Composable
+internal fun ImageLoaderScope.WithDefaultPlaceholder(
+    modifier: Modifier,
+    contentDescription: String? = null
+) {
+    Placeholder {
+        DefaultImagePlaceholder(modifier, contentDescription)
+    }
+}
+
+/**
+ * Renders the app default image placeholder if loading the image failed.
+ *
+ * @param modifier [Modifier] allowing to control among others the dimensions and shape of the image.
+ * @param contentDescription Text provided to accessibility services to describe what this image represents.
+ * Defaults to [null] suited for an image used only for decorative purposes and not to be read by
+ * accessibility services.
+ */
+@Composable
+internal fun ImageLoaderScope.WithDefaultFallback(
+    modifier: Modifier,
+    contentDescription: String? = null
+) {
+    Fallback {
+        DefaultImagePlaceholder(modifier, contentDescription)
+    }
+}
+
+/**
+ * Application default image placeholder.
+ *
+ * @param modifier [Modifier] allowing to control among others the dimensions and shape of the image.
+ * @param contentDescription Text provided to accessibility services to describe what this image represents.
+ * Defaults to [null] suited for an image used only for decorative purposes and not to be read by
+ * accessibility services.
+ */
+@Composable
+internal fun DefaultImagePlaceholder(
+    modifier: Modifier,
+    contentDescription: String? = null
+) {
+    val color = when (isSystemInDarkTheme()) {
+        true -> PhotonColors.DarkGrey30
+        false -> PhotonColors.LightGrey30
+    }
+
+    Image(ColorPainter(color), contentDescription, modifier)
+}
+
+@Composable
+@Preview
+private fun DefaultImagePlaceholderPreview() {
+    DefaultImagePlaceholder(
+        Modifier
+            .size(200.dp, 100.dp)
+            .clip(RoundedCornerShape(8.dp))
+    )
+}
--- a/app/src/main/java/org/mozilla/fenix/compose/ListItemTabLarge.kt
+++ b/app/src/main/java/org/mozilla/fenix/compose/ListItemTabLarge.kt
@@ -1,0 +1,160 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.compose
+
+import androidx.compose.foundation.clickable
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.size
+import androidx.compose.foundation.layout.width
+import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.material.Card
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.clip
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import mozilla.components.browser.engine.gecko.fetch.GeckoViewFetchClient
+import mozilla.components.concept.fetch.Client
+import org.mozilla.fenix.theme.FirefoxTheme
+
+/**
+ * Default layout of a large tab shown in a list taking String arguments for title and caption.
+ * Has the following structure:
+ * ```
+ * ---------------------------------------------
+ * | -------------- Title                      |
+ * | |    Image   | wrapped on                 |
+ * | |    from    | three rows if needed       |
+ * | |  imageUrl  |                            |
+ * | -------------- Optional caption           |
+ * ---------------------------------------------
+ * ```
+ *
+ * @param client [Client] instance to be used for downloading the image.
+ * When using [GeckoViewFetchClient] the image will automatically be cached if it has the right headers.
+ * @param imageUrl URL from where the to download a header image of the tab this composable renders.
+ * @param title Title off the tab this composable renders.
+ * @param caption Optional caption text.
+ * @param onClick Optional callback to be invoked when this composable is clicked.
+ */
+@Composable
+fun ListItemTabLarge(
+    client: Client,
+    imageUrl: String,
+    title: String,
+    caption: String? = null,
+    onClick: (() -> Unit)? = null
+) {
+    ListItemTabSurface(client, imageUrl, onClick) {
+        TabTitle(text = title, maxLines = 3)
+
+        if (caption != null) {
+            TabSubtitle(text = caption)
+        }
+    }
+}
+
+/**
+ * Default layout of a large tab shown in a list taking composable arguments for title and caption
+ * allowing as an exception to customize these elements.
+ * Has the following structure:
+ * ```
+ * ---------------------------------------------
+ * | -------------- -------------------------- |
+ * | |            | |         Title          | |
+ * | |    Image   | |       composable       | |
+ * | |    from    | -------------------------- |
+ * | |  imageUrl  | -------------------------- |
+ * | |            | |   Optional composable  | |
+ * | -------------- -------------------------- |
+ * ---------------------------------------------
+ * ```
+ *
+ * @param client [Client] instance to be used for downloading the image.
+ * When using [GeckoViewFetchClient] the image will automatically be cached if it has the right headers.
+ * @param imageUrl URL from where the to download a header image of the tab this composable renders.
+ * @param title Composable rendering the title of the tab this composable represents.
+ * @param subtitle Optional tab caption composable.
+ * @param onClick Optional callback to be invoked when this composable is clicked.
+ */
+@Composable
+fun ListItemTabLarge(
+    client: Client,
+    imageUrl: String,
+    onClick: () -> Unit,
+    title: @Composable () -> Unit,
+    subtitle: @Composable (() -> Unit)? = null
+) {
+    ListItemTabSurface(client, imageUrl, onClick) {
+        title()
+
+        subtitle?.invoke()
+    }
+}
+
+/**
+ * Shared default configuration of a ListItemTabLarge Composable.
+ *
+ * @param client [Client] instance to be used for downloading the image.
+ * When using [GeckoViewFetchClient] the image will automatically be cached if it has the right headers.
+ * @param imageUrl URL from where the to download a header image of the tab this composable renders.
+ * @param onClick Optional callback to be invoked when this composable is clicked.
+ * @param tabDetails [Composable] Displayed to the the end of the image. Allows for variation in the item text style.
+ */
+@Composable
+private fun ListItemTabSurface(
+    client: Client,
+    imageUrl: String,
+    onClick: (() -> Unit)? = null,
+    tabDetails: @Composable () -> Unit
+) {
+    var modifier = Modifier.size(328.dp, 116.dp)
+    if (onClick != null) modifier = modifier.then(Modifier.clickable { onClick() })
+
+    Card(
+        modifier = modifier,
+        shape = RoundedCornerShape(8.dp),
+        backgroundColor = FirefoxTheme.colors.surface,
+        elevation = 6.dp
+    ) {
+        Row(
+            modifier = Modifier.padding(16.dp)
+        ) {
+            val (imageWidth, imageHeight) = 116.dp to 84.dp
+            val imageModifier = Modifier
+                .size(imageWidth, imageHeight)
+                .clip(RoundedCornerShape(8.dp))
+
+            Image(client, imageUrl, imageModifier, false, imageWidth)
+
+            Spacer(Modifier.width(16.dp))
+
+            Column(
+                modifier = Modifier.fillMaxSize(),
+                verticalArrangement = Arrangement.SpaceBetween
+            ) {
+                tabDetails()
+            }
+        }
+    }
+}
+
+@Composable
+@Preview
+private fun ListItemTabLargePreview() {
+    FirefoxTheme {
+        ListItemTabLarge(
+            client = FakeClient(),
+            imageUrl = "",
+            title = "This is a very long title for a tab but needs to be so for this preview",
+            caption = "And this is a caption"
+        ) { }
+    }
+}
--- a/app/src/main/java/org/mozilla/fenix/compose/ListItemTabLargePlaceholder.kt
+++ b/app/src/main/java/org/mozilla/fenix/compose/ListItemTabLargePlaceholder.kt
@@ -1,0 +1,79 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.compose
+
+import androidx.compose.foundation.clickable
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.size
+import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.material.Card
+import androidx.compose.material.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.text.TextStyle
+import androidx.compose.ui.text.style.TextAlign
+import androidx.compose.ui.text.style.TextOverflow
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+import org.mozilla.fenix.theme.FirefoxTheme
+
+/**
+ * Placeholder of a [ListItemTabLarge] with the same dimensions but only a centered text.
+ * Has the following structure:
+ * ```
+ * ---------------------------------------------
+ * |                                           |
+ * |                                           |
+ * | Placeholder text                          |
+ * |                                           |
+ * |                                           |
+ * ---------------------------------------------
+ * ```
+ *
+ * @param text The only [String] that this will display.
+ * @param onClick Optional callback to be invoked when this composable is clicked.
+ */
+@Composable
+fun ListItemTabLargePlaceholder(
+    text: String,
+    onClick: () -> Unit = { }
+) {
+    Card(
+        modifier = Modifier
+            .size(328.dp, 116.dp)
+            .clickable { onClick() },
+        shape = RoundedCornerShape(8.dp),
+        backgroundColor = FirefoxTheme.colors.surface,
+        elevation = 6.dp,
+    ) {
+        Column(
+            modifier = Modifier
+                .fillMaxSize()
+                .padding(16.dp),
+            verticalArrangement = Arrangement.Center
+        ) {
+            Text(
+                text = text,
+                color = FirefoxTheme.colors.textPrimary,
+                textAlign = TextAlign.Center,
+                overflow = TextOverflow.Ellipsis,
+                maxLines = 1,
+                style = TextStyle(fontSize = 20.sp),
+            )
+        }
+    }
+}
+
+@Composable
+@Preview
+private fun ListItemTabLargePlaceholderPreview() {
+    FirefoxTheme {
+        ListItemTabLargePlaceholder(text = "Item placeholder")
+    }
+}
--- a/app/src/main/java/org/mozilla/fenix/compose/SelectableChip.kt
+++ b/app/src/main/java/org/mozilla/fenix/compose/SelectableChip.kt
@@ -1,0 +1,77 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.compose
+
+import androidx.compose.foundation.background
+import androidx.compose.foundation.isSystemInDarkTheme
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.selection.selectable
+import androidx.compose.material.MaterialTheme
+import androidx.compose.material.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.clip
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.text.TextStyle
+import androidx.compose.ui.text.capitalize
+import androidx.compose.ui.text.intl.Locale
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+import mozilla.components.ui.colors.PhotonColors
+import org.mozilla.fenix.theme.FirefoxTheme
+
+/**
+ * Default layout of a selectable chip.
+ *
+ * @param text [String] displayed in this chip. Ideally should only be one word.
+ * @param isSelected Whether this should be shown as selected.
+ * @param onClick Callback for when the user taps this.
+ */
+@Composable
+fun SelectableChip(
+    text: String,
+    isSelected: Boolean,
+    onClick: () -> Unit
+) {
+    val contentColor = when (isSystemInDarkTheme()) {
+        true -> PhotonColors.LightGrey10
+        false -> if (isSelected) PhotonColors.LightGrey10 else PhotonColors.DarkGrey90
+    }
+
+    @Suppress("MagicNumber")
+    val backgroundColor = when (isSystemInDarkTheme()) {
+        true -> if (isSelected) PhotonColors.Violet50 else PhotonColors.DarkGrey50
+        // Custom color codes matching the Figma design.
+        false -> if (isSelected) { Color(0xFF312A65) } else { Color(0x1420123A) }
+    }
+
+    Box(
+        modifier = Modifier
+            .selectable(isSelected) { onClick() }
+            .clip(MaterialTheme.shapes.small)
+            .background(backgroundColor)
+            .padding(16.dp, 10.dp)
+    ) {
+        Text(
+            text = text.capitalize(Locale.current),
+            style = TextStyle(fontSize = 14.sp),
+            color = contentColor
+        )
+    }
+}
+
+@Composable
+@Preview
+private fun SelectableChipPreview() {
+    FirefoxTheme {
+        Box(Modifier.fillMaxSize().background(FirefoxTheme.colors.surface)) {
+            SelectableChip("Chirp", false) { }
+            SelectableChip(text = "Chirp", isSelected = true) { }
+        }
+    }
+}
--- a/app/src/main/java/org/mozilla/fenix/home/HomeFragmentStore.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/HomeFragmentStore.kt
@@ -53,7 +53,7 @@ data class Tab(
  * @property recentTabs The list of recent [RecentTab] in the [HomeFragment].
  * @property recentBookmarks The list of recently saved [BookmarkNode]s to show on the [HomeFragment].
  * @property historyMetadata The list of [HistoryMetadataGroup].
- * @property pocketStories Currently shown [PocketRecommendedStory]ies.
+ * @property pocketStories The list of currently shown [PocketRecommendedStory]s.
  * @property pocketStoriesCategories All [PocketRecommendedStory] categories.
  * Also serves as an in memory cache of all stories mapped by category allowing for quick stories filtering.
  */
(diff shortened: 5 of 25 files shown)


app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesComposables.kt
```diff
@@ -21,160 +15,106 @@ import androidx.compose.foundation.layout.Spacer
...
-    client: Client
+    client: Client,
+    onExternalLinkTapped: (String) -> Unit
 ) {
-    // Items will be shown on two rows. Ceil the divide result to show more items on the top row.
-    val halfStoriesIndex = (stories.size + 1) / 2
+    // Show stories in a 3 by 3 grid
+    val gridLength = 3
 
     LazyRow {
```
Mugurell, line 99: Lazy lists automatically clip children and their shadows on the main axis - https://issuetracker.google.com/issues/186318448
This means here that children's horizontal shadows will be clipped.
Tried using a scrollbable Row. Same result.
The only solution I see involves refactoring the homescreen to remove `the 16dp padding` and let each child handle this on it's own. This way we could ensure we leave enough space for the shadows.
But right now this is the only homescreen that would need this so maybe we can wait a bit more until all homescreen views are migrated to composables and then we can have a cohesive approach.

Mugurell:
Another interesting issue is with Talkback which seems to see [the entire ComposeView](https://github.com/mozilla-mobile/fenix/blob/4e846b3cb6617f1d366b6c8457e29d782f1756e8/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlAdapter.kt#L260) as the one parent for everything so for a better a11y navigation the feature would need more ComposeViews. Tried child Columns, `heading`() semantics, all in vain.

If testing that same composables structure on a new screen, set with `setContent` and so without a parent ComposeView everything is seen individually and a better grouping for a11y is possible.
So I'm thinking that maybe the StoriesViewHolder should be broken off in 3 ViewHolders to force the a11y grouping.

app/src/main/res/drawable/pocket_vector.xml
```diff
@@ -0,0 +1,5 @@
+<vector android:autoMirrored="true" android:height="22dp"
```
gabrielluong, line 1: Do we need a license here?

app/src/test/java/org/mozilla/fenix/home/HomeFragmentTest.kt
```diff
@@ -78,4 +87,222 @@ class HomeFragmentTest {
...
+                it.category != otherStoriesCategory.name && it.category != anotherStoriesCategory.name
+            }
+        )
+    }
+
+    @Test
+    fun `GIVEN two categories are selected WHEN getFilteredStories is called for more than in the categories THEN results topped with ones from the default category are returned`() {
+        val result = getFilteredStories(availableCategories, listOf(otherStoriesCategory, anotherStoriesCategory), 8)
+
+        Assert.assertEquals(8, result.size)
```
gabrielluong, line 157: Can we drop the `Assert.`?

app/src/main/java/org/mozilla/fenix/composables/ListItemTabLarge.kt
```diff
@@ -0,0 +1,154 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.composables
```
gabrielluong, line 5: Can we refactor the naming of this package to be just `compose` instead of `composables` since we already use that naming within AC.

app/src/main/java/org/mozilla/fenix/composables/HeadingText.kt
```diff
@@ -0,0 +1,48 @@
...
+import org.mozilla.fenix.theme.FirefoxTheme
+
+/**
+ * Default layout for the header of a screen section.
+ *
+ * @param text [String] to be styled as header and displayed.
+ * @param modifier [Modifier] to be applied to the [Text].
+ */
+@Composable
+fun HeadingText(
```
gabrielluong, line 26: I think we can just call this `Header`.

app/src/main/java/org/mozilla/fenix/composables/ClickableSubstringText.kt
```diff
@@ -0,0 +1,97 @@
...
+                .firstOrNull()?.let {
+                    onClick()
+                }
+        }
+    )
+}
+
+@Composable
+@Preview
+private fun FinalDesign() {
```
gabrielluong, line 87: Instead of calling these `FinalDesign`, can we go with the XPreview naming convention that is used around Jetpack Compose? In this case, `ClickableSubstringTextPreview`

gabrielluong (changes requested):

app/src/main/java/org/mozilla/fenix/composables/ListItemTabLarge.kt
```diff
@@ -0,0 +1,154 @@
...
+
+        subtitle?.invoke()
+    }
+}
+
+/**
+ * Shared default configuration of a ListItemTabLarge Composable.
+ */
+@Composable
+private fun CommonSurface(
```
gabrielluong, line 106: Maybe `ListItemTabSurface`?

app/src/main/java/org/mozilla/fenix/composables/TabSubtitle.kt
```diff
@@ -0,0 +1,49 @@
...
+@Composable
+fun TabSubtitle(
+    text: String,
+    modifier: Modifier = Modifier
+) {
+    Text(
+        modifier = modifier,
+        maxLines = 1,
+        text = text,
+        style = TextStyle(fontSize = 12.sp,),
```
gabrielluong, line 33:
```suggestion
        style = TextStyle(fontSize = 12.sp),
```

app/src/main/java/org/mozilla/fenix/composables/StaggeredHorizontalGrid.kt
```diff
@@ -0,0 +1,137 @@
...
+ * are placed on the same row as wider ones if the otherwise next item doesn't fit.
+ *
+ * @param modifier [Modifier] to be applied to the layout.
+ * @param horizontalItemsSpacing minimum horizontal space between items. Does not add spacing to layout bounds.
+ * @param verticalItemsSpacing vertical space between items
+ * @param arrangement how the items will be horizontally aligned and spaced.
+ * @param content the children composables to be laid out.
+ */
+@Composable
+fun StaggeredHorizontalGrid(
```
gabrielluong, line 38: I am wondering if you also saw this https://google.github.io/accompanist/flowlayout/? I imagine at some point we don't want to maintain our own flow layout.

app/src/main/java/org/mozilla/fenix/composables/ListItemTabLarge.kt
```diff
@@ -0,0 +1,154 @@
...
+    client: Client,
+    imageUrl: String,
+    onClick: (() -> Unit)? = null,
+    tabDetails: @Composable () -> Unit
+) {
+    var modifier = Modifier.size(328.dp, 116.dp)
```
gabrielluong, line 112: I think we don't even need to call this `ListItemTabLarge` if we add a modifier argument to the parameters and allow the caller to specify the size.

app/src/main/java/org/mozilla/fenix/composables/TabTitle.kt
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Default layout for a tab composable title.
+ *
+ * @param text tab title
+ * @param maxLines maximum number of lines for [text] to span, wrapping if necessary.
+ * If the text exceeds the given number of lines it will be ellipsized.
+ * @param modifier optional [Modifier] to be applied to the layout.
+ */
+@Composable
+fun TabTitle(
```
gabrielluong, line 26: These `TabTitle` and `TabSubtitle` functions might be better off living inside of `ListItemTabLarge` for the time being since the naming suggests it relates to a tab, and it's just a simple wrapper around `Text()`. At some point, you might want to call this `PrimaryText14` or even add a `fontSize` parameter. 

The way they are named now I would just expect people to just write their own `Text()` since a title might not be related to a tab, but a bookmark or history, etc. So, I could see us just grouping these composable functions to where they are currently used instead of separating them up completely.

app/src/main/java/org/mozilla/fenix/composables/SelectableChip.kt
```diff
@@ -0,0 +1,83 @@
...
+
+    val backgroundColor = when (isSystemInDarkTheme()) {
+        true -> if (isSelected) PhotonColors.Violet50 else PhotonColors.DarkGrey50
+        false -> if (isSelected) {
+            PhotonColors.Ink20
+        } else {
+            // Custom color code matching the Figma design.
+            @SuppressWarnings("MagicNumber")
+            val lightBackground = Color(0x14312A65)
+            lightBackground
```
gabrielluong, line 55: Can we just return Color(0x14312A65) here? Do we also have a name for this color in the Figma design?

app/src/main/java/org/mozilla/fenix/composables/ListItemTabLargePlaceholder.kt
```diff
@@ -0,0 +1,79 @@
...
+) {
+    Card(
+        elevation = 6.dp,
+        shape = RoundedCornerShape(8.dp),
+        modifier = Modifier
+            .size(328.dp, 116.dp)
+            .clickable { onClick() },
+        backgroundColor = when (isSystemInDarkTheme()) {
+            true -> PhotonColors.DarkGrey50
+            false -> PhotonColors.White
```
gabrielluong, line 53: This should match FirefoxTheme.color.surface

app/src/main/java/org/mozilla/fenix/home/HomeFragmentStore.kt
```diff
@@ -50,7 +51,9 @@ data class Tab(
  * @property recentTabs The list of recent [TabSessionState] in the [HomeFragment].
  * @property recentBookmarks The list of recently saved [BookmarkNode]s to show on the [HomeFragment].
  * @property historyMetadata The list of [HistoryMetadataGroup].
- * @property pocketArticles The list of [PocketRecommendedStory].
+ * @property pocketStories Currently shown [PocketRecommendedStory]ies.
```
gabrielluong, line 54:
```suggestion
 * @property pocketStories The list of currently shown [PocketRecommendedStory]s.
```

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlAdapter.kt
```diff
@@ -245,11 +247,15 @@ class AdapterItemDiffCallback : DiffUtil.ItemCallback<AdapterItem>() {
...
+@Suppress("LongParameterList")
 class SessionControlAdapter(
     private val store: HomeFragmentStore,
     private val interactor: SessionControlInteractor,
     private val viewLifecycleOwner: LifecycleOwner,
-    private val components: Components
+    private val components: Components,
+    private val onPocketStoriesShown: (List<PocketRecommendedStory>) -> Unit,
+    private val onCategoryClick: (PocketRecommendedStoryCategory) -> Unit,
+    private val onExternalLinkTapped: (String) -> Unit
```
gabrielluong, line 258: Can we add this to an Interactor for Pocket?

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlAdapter.kt
```diff
@@ -259,7 +265,10 @@ class SessionControlAdapter(
             PocketStoriesViewHolder.LAYOUT_ID -> return PocketStoriesViewHolder(
                 ComposeView(parent.context),
                 store,
-                components.core.client
+                components.core.client,
+                onPocketStoriesShown,
+                onCategoryClick,
+                onExternalLinkTapped
```
gabrielluong, line 271: We should just pass an Interactor here similar to `RecentTabViewHolder`

app/src/main/java/org/mozilla/fenix/composables/ListItemTabLarge.kt
Mugurell: Sure. Thanks!

app/src/main/java/org/mozilla/fenix/composables/StaggeredHorizontalGrid.kt
Mugurell: `flowlayout` has two important issues right now for this feature at least:
- no RTL support while `StaggeredHorizontalGrid` correctly handles RTL also.
- lays out all items in the exact order. This can be ok for most cases but it can also lead to situations in which although there is enough space left on one row if the next immediate items does not fit there then that space is left empty. 
So instead of the grid being neatly packed, more closely following the spacing desired the spacing will be almost random.
For this feature the items are rather random while they all should be aligned to start with the same spacing for a clean design, scenario for which `StaggeredHorizontalGrid` suited.

In the future we could use `flowlayout` or tweak this whenever we need withought depending on external functionality/behavior we have little control over.

app/src/main/java/org/mozilla/fenix/composables/SelectableChip.kt
Mugurell: That assignement allows for the `MagicNumber` suppression there instead of suppressing a bit higher up.

Seems like the recent `PhothonColors` update changed the definition of `Ink20` and that color code needs to be hardcoded also (selected background used in light theme).

`14312A65` is `Ink80` but with just 8% alpha so instead of hardcoding that value I can use `PhotonColors.Ink80.copy(alpha = 0.08f)` which is still not great since we're still adding a custom color but makes the relation with the `PhotonColors` clearer.

gabrielluong: I think it is also fine to add `Ink80A8` in `PhotonColors` as well. We already do this in our existing colors.xml

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlAdapter.kt
Mugurell: Talked about this with Sebastian some time ago, where should all the logic stay when using composables.
Things being new we have to try a few approaches to find what works but the general plans would be to try and move away from Interactors/Controllers. For the start. Then continue moving away from even more of the current architecture. More thoughts here - https://docs.google.com/document/d/1BG_aGO_dCP7c1ViY3UG3Rqqi6PCJlRLDfI3LVluWyGM/edit# .

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlAdapter.kt
Mugurell: Same as above. I'd try moving away from Controllers and especially from Interactors which currently don't do much.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlAdapter.kt
gabrielluong: I have read the document and understand where the proposals are coming from. Bigger picture to me would be that everything in SessionControlAdapter is completely remove since everything would be converted to Jetpack Compose in the future.

What I don't particularly want to see is that this becomes an intermediate example that people follow and continue to overload SessionControlAdapter with additional parameters as we can also assume this will go away in the future. As an example, where would we draw the line now if we add more compose components and keep adding onRecentTabClick, onRecentBookmarkClick, onRecentVisitClick into this param list. I think we should keep to our existing lint rule on not running a big parameter list. 

The cleaner solution even though we are sticking with encapsulating common logic into Interactor still holds valid here.

app/src/main/java/org/mozilla/fenix/composables/SelectableChip.kt
Mugurell: I would stay away as much as possible of adding new colors, dimensions, etc on top on what we have to avoid resources blowing up and be hard to keep track of/maintain like it's the situation now.
Google recommends a very small subset of colors etc with a reason and if we're also starting fresh with the goal of a cohesive design and reusability we should strive to stick to our original specs which needs to be stable.
So I'd prefer having something ugly in code that immediately screams custom to easily know that ain't right and needs to be updated.

I see this custom color as rather a sign that it needs a different value from the ones we have in [PhotonColors](https://github.com/mozilla-mobile/android-components/blob/50914d96379f6ef6cfbc13e2965c6e884247e90a/components/ui/colors/src/main/java/mozilla/components/ui/colors/PhotonColors.kt). @topotropic

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlAdapter.kt
Mugurell: @pocmo ?

app/src/main/java/org/mozilla/fenix/composables/ListItemTabLarge.kt
Mugurell: This is `ListItemTabLarge` as a default for this usecase which for a cohesive design should not change.
The image dimensions are hardcoded and so does the size of this to force all users stick to this design. Wherever this is used it will look the same.
For another usecase we'd have `ListiItemTabMedium` / `ListItemTabSmall`, again them having as many as possible hardcodes to enforce the same design all throughout the app.
More of these variants needed -> clear sign that the design becomes bloated and the UX looses it's cohesiveness.

app/src/main/java/org/mozilla/fenix/composables/TabTitle.kt
Mugurell: I have this outside of `ListItemTabLarge` and would expect everyone who needs to show a tab title to use this instead of using their own `Text()` to enforce the same style all throughout the app.
This is how a tab title should look like everywhere in the app.
This colors, fonts, etc.

app/src/main/java/org/mozilla/fenix/composables/HeadingText.kt
Mugurell: Maybe `SectionHeader`?
What I tried in this patch also is to standardize some usecases.
We should have as few Composables as possible, each containing the specs for a particular usecase, reused all throughout the app.
A suggestive name and good description is a hard requirement.
So `SectionHeader` later followed by some other headers make sense to me.

app/src/main/java/org/mozilla/fenix/composables/TabTitle.kt
Mugurell: The flow I think of is that everytime we need to add a composable somewhere in the app we look for what already exists.
If it doesn't yet exist -> app design specs changed. Which means a big change and we should strive to avoid this.

With us just starting on this road, the "app design specs" are bound to change, to be added upon continuously.

But very important: if we want a cohesive design and reusable composables most of the time we'll have to extract a new composable which encomprises a new style/usecase to one that can be reused and avoid implementing it in place.

In this case, since we'd have multiple tab designs which would all need to show a title we can standardize how a tab title should look like and reuse that everywhere. Using a new Text() for this usecase and styling it in place would be a mistake.

Ideally all this reusable composables - their name and specs would come from UX who would have these in Figma and reuse them on new designs, with us doing the same in code. Then for both teams it would be easy to reuse existing designs and rethink the need for others.

Curious about your thoughts also @pocmo @mcarare @topotropic .

mcarare: IMO separating the composable and not tying them to their current usage place enforces reusability. This also makes sure the composable does not contain logic related to the current usage site. It also ensures a design change is reflected everywhere in the app when the same component is used. And it also makes generating previews for components easier. These are just initial thoughts, we can debate any one of them.

app/src/main/java/org/mozilla/fenix/composables/HeadingText.kt
gabrielluong: `SectionHeader` sounds good

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlAdapter.kt
gabrielluong: Another thought is that we still want to be able to write tests for the business logic that might be called by `onPocketStoriesShown`, etc. It is still beneficial to throw the functionality within a controller-type class. In general, I am in favour of dropping one step of indirections caused by the Interactors.

Mugurell: Added an Interactor and Controller in https://github.com/mozilla-mobile/fenix/pull/21104 to be used for all later functionality incl this.

app/src/main/java/org/mozilla/fenix/composables/SelectableChip.kt
Mugurell: Went with the ugly custom color codes.

app/src/main/res/drawable/pocket_vector.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<vector android:autoMirrored="true" android:height="22dp"
```
gabrielluong, line 5: Might want to run a format here. I am not used to seeing multiple android:X properties on the same line. and xmls:android at the bottom.

app/src/main/java/org/mozilla/fenix/home/PocketUpdatesMiddleware.kt
```diff
@@ -0,0 +1,45 @@
...
+class PocketUpdatesMiddleware(
+    private val coroutineScope: CoroutineScope,
+    private val pocketStoriesService: PocketStoriesService
+) : Middleware<HomeFragmentState, HomeFragmentAction> {
+    override fun invoke(
+        context: MiddlewareContext<HomeFragmentState, HomeFragmentAction>,
+        next: (HomeFragmentAction) -> Unit,
+        action: HomeFragmentAction
+    ) {
+
```
gabrielluong, line 26: nit: no new line necessary

Mugurell: That change is not part of this PR. Will update in https://github.com/mozilla-mobile/fenix/pull/21277.

app/src/main/res/drawable/pocket_vector.xml
Mugurell: Done.

gabrielluong:
Will look at this first thing in the morning

Mugurell:
`ui-test-x86-debug` failing apparently intermittent because of `editCustomSearchEngineTest`.
Restarting CI.

Mugurell:
And now an intermittently failing unit - https://github.com/mozilla-mobile/fenix/issues/21333 test failed.
Will look into fixing this.

app/src/main/java/org/mozilla/fenix/compose/ImagesPlaceholder.kt
```diff
@@ -0,0 +1,87 @@
...
+import mozilla.components.support.images.compose.loader.Fallback
+import mozilla.components.support.images.compose.loader.ImageLoaderScope
+import mozilla.components.support.images.compose.loader.Placeholder
+import mozilla.components.ui.colors.PhotonColors
+
+/**
+ * Renders the app default image placeholder while the image is still getting loaded.
+ *
+ * @param modifier [Modifier] allowing to control among others the dimensions and shape of the image.
+ * @param contentDescription text provided to accessibility services to describe what this image represents.
```
gabrielluong, line 26: Just a general feedback that I noticed with all the @param. I would generally just start with capitalizing so it doesn't end up looking a bit awkward when we have multiple sentences for the @param description.


```suggestion
 * @param contentDescription Text provided to accessibility services to describe what this image represents.
```

app/src/main/java/org/mozilla/fenix/compose/ListItemTabLarge.kt
```diff
@@ -0,0 +1,154 @@
...
+) {
+    ListItemTabSurface(client, imageUrl, onClick) {
+        title()
+
+        subtitle?.invoke()
+    }
+}
+
+/**
+ * Shared default configuration of a ListItemTabLarge Composable.
```
gabrielluong, line 103: We could reuse the @param descriptions from above.

app/src/main/java/org/mozilla/fenix/compose/ListItemTabLargePlaceholder.kt
```diff
@@ -0,0 +1,81 @@
...
+        backgroundColor = FirefoxTheme.colors.surface,
+    ) {
+        Column(
+            modifier = Modifier
+                .fillMaxSize()
+                .padding(16.dp),
+            verticalArrangement = Arrangement.Center
+        ) {
+            Text(
+                textAlign = TextAlign.Center,
```
gabrielluong, line 61: I do generally like specifying the param names because it makes it easier in terms of code review to know how the arguments matches up, but I also find myself reordering them to stick with the order that `Text` defines. For instance, `text` would usually appear first follow by `color`, etc.

app/src/main/java/org/mozilla/fenix/compose/ListItemTabLargePlaceholder.kt
```diff
@@ -0,0 +1,81 @@
...
+                text = text,
+                style = TextStyle(fontSize = 20.sp),
+                maxLines = 1,
+                overflow = TextOverflow.Ellipsis,
+                color = when (isSystemInDarkTheme()) {
+                    true -> PhotonColors.LightGrey05
+                    false -> PhotonColors.DarkGrey90
```
gabrielluong, line 68: I feel like we definitely have a FirefoxTheme.colors for this. Probably primaryText?

app/src/main/java/org/mozilla/fenix/compose/ListItemTabLargePlaceholder.kt
```diff
@@ -0,0 +1,81 @@
...
+                },
```
gabrielluong, line 69: I will note that lingering comma here. I don't think we complain about this anymore and is generally encouraged for long running param lists. So, it's just a note in case you prefer this kind of thing removed.

app/src/main/java/org/mozilla/fenix/compose/SelectableChip.kt
```diff
@@ -0,0 +1,78 @@
...
+ * @param isSelected whether this should be shown as selected.
+ * @param onClick callback for when the user taps this.
+ */
+@Composable
+fun SelectableChip(
+    text: String,
+    isSelected: Boolean,
+    onClick: () -> Unit
+) {
+
```
gabrielluong, line 41: nit: new line

```suggestion
```

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesViewHolder.kt
```diff
@@ -76,21 +87,34 @@ fun PocketStories(
...
 
-    ExpandableCard(
-        Modifier
-            .fillMaxWidth()
-            .padding(top = 40.dp)
-    ) {
-        PocketRecommendations {
-            Column {
-                PocketStories(stories ?: emptyList(), client)
+    FirefoxTheme {
```
gabrielluong, line 90: I think we should move this to where `setContent` is. If I think about that future `HomeScreen() {}` composable function that wraps all our home screen feature. We wouldn't necessarily have to specify `FirefoxTheme` at the feature composable level (if that makes sense).

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesComposables.kt
```diff
@@ -2,188 +2,114 @@
...
-            )
+    val imageUrl = story.imageUrl.replace(
+        "{wh}",
+        with(LocalDensity.current) { "${116.dp.toPx().roundToInt()}x${84.dp.toPx().roundToInt()}" }
+    )
+    ListItemTabLarge(
+        client = client,
+        imageUrl = imageUrl,
+        onClick = { onStoryClick(story) },
+        title = { TabTitle(text = story.title, maxLines = 3) },
```
gabrielluong, line 72: Maybe `TabTitle` should go in a new line to follow the same style as `TabSubtitleWithInterdot`. The extra room and spacing might help people to recognize that this is @content more easily.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesComposables.kt
```diff
@@ -2,188 +2,114 @@
...
-                    Spacer(modifier = Modifier.height(24.dp))
+        itemsIndexed(stories.chunked(gridLength)) { rowIndex, columnItems ->
+            Column(Modifier.padding(end = if (rowIndex < gridLength - 1) 8.dp else 0.dp)) {
+                for (index in 0 until gridLength) {
+                    columnItems.getOrNull(index)?.let { story ->
+                        PocketStory(story, client) {
+                            onExternalLinkClicked(story.url)
+                        }
+                    } ?: ListItemTabLargePlaceholder(stringResource(R.string.pocket_stories_placeholder_text)) {
+                        onExternalLinkClicked("http://getpocket.com/explore")
```
gabrielluong, line 107: Wouldn't this cause multiple Explore cards if a column had only 1 item for example? I think you might need a `break` somewhere

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesComposables.kt
```diff
@@ -2,188 +2,114 @@
...
 fun PocketStories(
     @PreviewParameter(PocketStoryProvider::class) stories: List<PocketRecommendedStory>,
-    client: Client
+    client: Client,
+    onExternalLinkClicked: (String) -> Unit
 ) {
-    // Items will be shown on two rows. Ceil the divide result to show more items on the top row.
-    val halfStoriesIndex = (stories.size + 1) / 2
+    // Show stories in a 3 by 3 grid
+    val gridLength = 3
```
gabrielluong, line 96: I don't know if this will useful, but have you seen https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/windowed.html

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesComposables.kt
```diff
@@ -202,270 +128,96 @@ fun PocketStoriesCategories(
...
+            Modifier
+                .fillMaxWidth()
+                .semantics(mergeDescendants = true) { },
+            verticalAlignment = Alignment.CenterVertically
+        ) {
+            Icon(
+                painter = painterResource(id = R.drawable.pocket_vector),
+                contentDescription = null,
+                // Apply the red tint in code. Otherwise the image is black and white.
+                tint = Color(0xFFEF4056)
```
gabrielluong, line 176: Does this work for both light / dark theme?

gabrielluong (approved):

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesComposables.kt
Mugurell: That seems nice, thank you!
There is an update needed, some specs slightly changed regarding what stories to display. I'll try using this in that PR.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesComposables.kt
Mugurell: Theoretically we'd have more than enough items to display since we're also topping filtered items with ones from a `general` category.
This will soon change and the issue raised by you here would have real chances of happening so indeed, the code here needs some improvements to not always try to show a certain max number of items.

app/src/main/java/org/mozilla/fenix/compose/ListItemTabLargePlaceholder.kt
Mugurell: Don't think this was really intended.
If you've brought the discussion I'm all for trailing commas. That way when a new parameter is added the above line is not modified to add a comma and the git history is kept a bit cleaner. Might worth to have a bigger discussion about this style choice.
So in this case if it's already there and lint is okay with it I'd keep it.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesComposables.kt
Mugurell: Tested and it looks okay in both variations.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesViewHolder.kt
Mugurell: Agree. Thanks!

app/src/main/java/org/mozilla/fenix/compose/ListItemTabLargePlaceholder.kt
Mugurell: Indeed. Thanks!
Not a fan of the hardcodes, used them just until a theme is ready. It's nice that we have some specs available at least for the most important usecases.

app/src/main/java/org/mozilla/fenix/compose/ListItemTabLargePlaceholder.kt
Mugurell: Fair point. Thanks!

app/src/main/java/org/mozilla/fenix/compose/ClickableSubstringText.kt
```diff
@@ -0,0 +1,97 @@
...
+            start = clickableStartIndex,
+            end = clickableEndIndex
+        )
+    }
+
+    ClickableText(
+        text = annotatedText,
+        onClick = {
+            annotatedText
+                .getStringAnnotations("link", it, it)
```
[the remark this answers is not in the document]
Mugurell: Indeed. `link` here is just a local tag I use to identify a portion of the text and then check if that is in the range clicked bye the user.

app/src/main/java/org/mozilla/fenix/compose/ClickableSubstringText.kt
```diff
@@ -0,0 +1,97 @@
...
+ * [Text] containing a substring styled as an URL informing when this is clicked.
+ *
+ * @param text full text that will be displayed
+ * @param textColor [Color] of the normal text. The URL substring will have a default URL style applied.
+ * @param clickableStartIndex [text] index at which the URL substring starts.
+ * @param clickableEndIndex [text] index at which the URL substring ends.
+ * @param onClick callback to be invoked only when the URL substring is clicked.
+ */
+@Composable
+fun ClickableSubstringText(
```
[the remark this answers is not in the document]
Mugurell: My train of though was "ClickableSubstring" in a bigger string - "Text"  😅.
We receive "Text", some start and end offsets and inform through a callback when the user taps in between that Text offsets. While indeed formatting that clickable substring as a link.
Will go with "ClickableSubstringLink".

app/src/main/java/org/mozilla/fenix/compose/ImagesPlaceholder.kt
Mugurell: Done.

Mugurell:
Seeing the same `WHEN showUndoSnackbarForTab is called for a normal tab with new tab button not visible THEN an appropriate snackbar is shown` test failing.
So my speculative fix from the second commit didn't work.
Will rebase and drop that then try merging again..

PR #21398, opened 2021-09-20
Make sure to only record view time once for a given tab access

csadilek:
With this change we're making sure that we don't record view time twice for a given tab access i.e., if we've already recorded a view time observation and the tab hasn't been accessed in the time between we won't record it again.

This makes it more tolerant and easier to implement on the consuming side, and also guard against recording the wrong view time if we ever introduce a regression in updating `lastAccess`.

WIP...still needs tests. :)


grigoryk (approved):
:shipit:

Issue #21418, opened 2021-09-21
[Crash] NPE in AddonsManagementFragment


csadilek:
Sentry: https://sentry.prod.mozaws.net/operations/firefox-nightly/issues/13014158/

Add-on installs successfully but the app crashes when updating the UI in the success callback. 

Looks like `binding` is null when the async operation finishes and invokes the callback. @codrut-topliceanu Can you take a quick look? I can't reproduce and the only change that landed recently was: https://github.com/mozilla-mobile/fenix/commit/7070bb5607f5ac9fd0825ab21e7abba675f98bdc.

We have seen cases like this before where the context is still not null, but the view got destroyed when the async operation returns. Then `binding` would be null causing the crash. Let's handle that, and potentially uplift.

```
java.lang.NullPointerException: null
    at org.mozilla.fenix.addons.AddonsManagementFragment$onPositiveButtonClicked$1$installOperation$1.invoke(AddonsManagementFragment.kt:8)
    at mozilla.components.feature.addons.AddonManager$installAddon$4.invoke(AddonManager.kt:9)
    at mozilla.components.browser.engine.gecko.GeckoEngine$installWebExtension$onInstallSuccess$1.invoke(GeckoEngine.kt:16)
    at mozilla.components.browser.engine.gecko.GeckoEngine$$ExternalSyntheticLambda9.onValue
    at org.mozilla.geckoview.GeckoResult.lambda$thenInternal$4(GeckoResult.java:2)
    at org.mozilla.geckoview.GeckoResult.$r8$lambda$AFxsnIxghFV7ZhrvgIHpdLryPVU
    at org.mozilla.geckoview.GeckoResult$$ExternalSyntheticLambda0.run
    at org.mozilla.geckoview.GeckoResult.lambda$dispatchLocked$5(GeckoResult.java:2)
    at org.mozilla.geckoview.GeckoResult.$r8$lambda$PbNUepeldwERMEcQ_iVWEzy-zTE
    at org.mozilla.geckoview.GeckoView$$ExternalSyntheticLambda0.run
    at android.os.Handler.handleCallback(Handler.java:883)
    at android.os.Handler.dispatchMessage(Handler.java:100)
    at android.os.Looper.loop(Looper.java:214)
    at android.app.ActivityThread.main(ActivityThread.java:7356)
    at java.lang.reflect.Method.invoke(Method.java)
[...]

PR #21434, opened 2021-09-22
For #20762 - Fixes NPE crash

codrut-topliceanu:
Keeps the `_binding` from being set to null `onDestroy`. While not ideal it will prevent NPEs from happening on coroutines that live longer than the view.

For https://github.com/mozilla-mobile/fenix/issues/21418
### Pull Request checklist

- [x] **Tests**: This PR includes not tests as it's only a minor change
- [x] **Screenshots**: This PR includes no screenshots or GIFs as the changes aren't visible

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/addons/AddonsManagementFragment.kt
```diff
@@ -90,7 +90,9 @@ class AddonsManagementFragment : Fragment(R.layout.fragment_add_ons_management)
         super.onDestroyView()
         // letting go of the resources to avoid memory leak.
         adapter = null
-        _binding = null
+        // Intentionally leaking the binding since it is used when the async operation returns
+        // It will ultimately be destroyed when the Fragment instance is destroyed.
+        // _binding = null
```
pocmo, line 95: This is not great. Can't we fix the async operation(s) to use a nullable type?

csadilek (commented):
Thanks @codrut-topliceanu for taking this on so quickly! We should also uplift the fix so we don't ship the crash to release, but let's land in Nightly and verify first.

I also think we should check for null in those callbacks or using `binding?.` instead. To prevent a bigger refactor, I think it's also OK to use `_binding` in those callbacks which is already `nullable` with a comment to explain why :).

Mugurell:
Drive by comment:
It seems that the crash stems from 
https://github.com/mozilla-mobile/fenix/blob/7070bb5607f5ac9fd0825ab21e7abba675f98bdc/app/src/main/java/org/mozilla/fenix/addons/AddonsManagementFragment.kt#L84-L85
which tries to update a fragment from the backstack, which in that already gone through `onDestroyView`.
If we are to continue to null the binding in `onDestroyView` and check it for null here then we'd break this functionality.
I'd say we either "leak" the binding past `onDestroyView` either refactor the functionality .. which it isn't clear how it works.

csadilek:
> If we are to continue to null the binding in onDestroyView and check it for null at line 85 then we'd break the current functionality which needs that to not be null.

Hm I don't understand what functionality would break here though. If the view is destroyed we'd simply do nothing if we check for null or use `binding?.` which means we wouldn't hide the `addonProgressOverlay` but since the view was destroyed it wouldn't be visible anymore anyway?

So, it seemed fine to me, but either way we should land a fix so we can uplift next week and get rid of the crash :).

csadilek (approved):
Thanks, let's 🚢  and address the crashes. Curious if you have suggestions how to improve this fragment in general. If you have them, let's file an issue, there's probably a cleaner way we could handle those async ops.

Issue #20893, opened 2021-08-17
Search term groups in history


gabrielluong:

sv-ohorvath:
@gabrielluong doesn't this need QA?

gabrielluong:
> @gabrielluong doesn't this need QA?

Yes, I think so.

LaurentiuApahideanSV:
@gabrielluong Could you provide us with further information or specifications regarding what was changed?

PR #21482, opened 2021-09-23
For #20893: Dedupe urls/site in history groups

csadilek:
For Home we already do that and also correctly count the number of sites in the group. In history we count and display the duplicates which looking at my groups feels really off. One of the motivations here it to make it as easy as possible for users to get back to where they left off, so we shouldn't add a new group entry every time the user opens the site again.

It's also easier to fix in history as we're only sorting by `updatedAt` -> `visitedAt`, and we don't need to aggregate view times.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/history/PagedHistoryProvider.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/history/PagedHistoryProvider.kt
@@ -132,10 +132,14 @@ class DefaultPagedHistoryProvider(
         // item.
         result.addAll(history.filter { item -> historyMetadata.find { it.url == item.url } == null })
 
-        // Filter history metadata items with no view time.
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/fenix/components/history/PagedHistoryProvider.kt
```diff
@@ -132,10 +132,14 @@ class DefaultPagedHistoryProvider(
         // item.
         result.addAll(history.filter { item -> historyMetadata.find { it.url == item.url } == null })
 
-        // Filter history metadata items with no view time.
+        // Filter history metadata items with no view time an dedupe by url.
```
gabrielluong, line 135:
```suggestion
        // Filter history metadata items with no view time and dedupe by url.
```

gabrielluong (approved):

PR #21487, opened 2021-09-23
Fix mapping between history visits and groups to use createdAt

csadilek:
We currently have a 15s buffer to match metadata to its corresponding visit. However, a existing metadata record can be updated more than 15s after it was created e.g. when closing the tab and updating the view time.


app/src/test/java/org/mozilla/fenix/components/history/PagedHistoryProviderTest.kt
```diff
@@ -57,7 +57,7 @@ class PagedHistoryProviderTest {
             key = historyMetadataKey1,
             title = "mozilla",
             createdAt = 5,
-            updatedAt = 5,
+            updatedAt = 10,
```
csadilek, line 60: ^^ Just changing the `updatedAt` values because I want the test to fail if we ever no use `createdAt`. :)

grigoryk (approved):

PR #21527, opened 2021-09-28
Do not capture search terms when user navigates away via app

csadilek:
We currently incorrectly record search terms when navigating from the search results page directly (via the app/toolbar) to any other page. This will make that other page be grouped under those likely unrelated search terms.

With this patch we only capture search terms when the user navigates via web content i.e., follows a link.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/historymetadata/HistoryMetadataMiddleware.kt
+++ b/app/src/main/java/org/mozilla/fenix/historymetadata/HistoryMetadataMiddleware.kt
@@ -6,6 +6,7 @@ package org.mozilla.fenix.historymetadata
 
 import mozilla.components.browser.state.action.BrowserAction
 import mozilla.components.browser.state.action.ContentAction
+import mozilla.components.browser.state.action.EngineAction
 import mozilla.components.browser.state.action.HistoryMetadataAction
 import mozilla.components.browser.state.action.MediaSessionAction
 import mozilla.components.browser.state.action.TabListAction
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/fenix/historymetadata/HistoryMetadataMiddleware.kt
```diff
@@ -29,6 +30,10 @@ class HistoryMetadataMiddleware(
 
     private val logger = Logger("HistoryMetadataMiddleware")
 
+    // Tracks whether a page load is in progress that was triggered directly by the app
+    // e.g. via the toolbar as opposed to via web content.
+    private var directLoadTriggered: Boolean = false
```
csadilek, line 35: I will file a follow-up in A-C to come up with something better in the long run.

grigoryk (approved):
This is starting to get a bit subtle, but I think the general sentiment is right - we don't want to capture searchTerms/referrer for direct navigation events.

I.e., when on search results (page A), when a user directly navigates either to a new search or a link via the awesomebar (page B), we should not record search terms from page A for metadata for page B.

Basically, we should treat direct navigation as a "fresh start". From the point of `UpdateHistoryState` events which trigger a "metadata write", there is no difference between an entirely new "navigation tree" (e.g. as caused by direct navigation) and, say, a click on a link within the page. So, we have to track additional state (e.g. the `directLoadTriggered` flag) and infer. It's a bit a subtle but makes sense from the point of view of the middleware.

This brings up an interesting side topic - a common usage pattern is to revise search terms after making an initial search. With this new behaviour, we're explicitly saying that a direct search "resets" a group being currently populated. It'll be interesting to explore if we can do this conditionally - e.g. if multiple searches are similar enough, perhaps, they should be collapsed? However, if something like that doesn't work perfectly (and it likely won't), it runs a risk of being quite annoying.

app/src/main/java/org/mozilla/fenix/historymetadata/HistoryMetadataMiddleware.kt
csadilek: https://github.com/mozilla-mobile/android-components/issues/11034

csadilek:
Yeah, ideally we should not have to hold additional state as this should already be represented in our `HistoryState` (or other new state): https://github.com/mozilla-mobile/android-components/issues/11034

This would make this a lot less subtle and easier to read.

PR #21545, opened 2021-09-28

MickeyMoz:


Mugurell:
`toolchain-linux64-android-gradle-dependencies` fail because of 
> PocketStoriesComposables.kt: (484, 17): No value passed for parameter 'timesShown'

Will update AC in https://github.com/mozilla-mobile/fenix/pull/21277

Issue #21562, opened 2021-09-29
FNXV2-17905


Mugurell:
Tracking https://mozilla-hub.atlassian.net/browse/FNXV2-17905

Mugurell:
Verified by QA - https://mozilla-hub.atlassian.net/browse/FNXV2-17905?focusedCommentId=465408
Closing.

PR #21571, opened 2021-09-29
For #21562 - Snap to next item when scrolling

Mugurell:
This works by replacing the fling animation with snapping to the next item in
the scroll direction.

Refer to https://mozilla-hub.atlassian.net/browse/FNXV2-17905 for the result.

*based on the not yet merged https://github.com/mozilla-mobile/fenix/pull/21392

### Pull Request checklist

1. click on Show All Checks,


Mugurell:
Draft for now since it isn't working perfectly.
Replacing the fling with a snapping animation seemed nice but for little enough gestures the LazyList will not ask for a fling so our code [from here](https://github.com/mozilla-mobile/fenix/pull/21571/commits/5be0e4fab4cc60cc46d5001020a18fa68b4bc835#diff-4ac3c56e20b9fbcc20a412df280d6b6eec769b95b6103525ce22f5191c43a39bR33) will not execute.

We do have access to [isDraggable](https://developer.android.com/reference/kotlin/androidx/compose/foundation/interaction/InteractionSource#(androidx.compose.foundation.interaction.InteractionSource).collectIsDraggedAsState()) which seem to work nicely in informing us when the user is scrolling but it seems like lazyList doesn't expose the scrollState or scroll callbacks so we cannot know the scroll direction.

Seems like the best solution would be similar to https://github.com/aakarshrestha/compose-pager-snap-helper/blob/main/composePagerSnapHelper/src/main/java/com/aakarshrestha/composepagersnaphelper/ComposePagerSnapHelper.kt - a parent for our list which would allow it to receive the scroll events from which we could infer when and in which direction to snap items.

Mugurell:
I was actually ignoring the case where the LazyList asks for a 0px fling - https://github.com/mozilla-mobile/fenix/pull/21571/commits/5be0e4fab4cc60cc46d5001020a18fa68b4bc835#diff-4ac3c56e20b9fbcc20a412df280d6b6eec769b95b6103525ce22f5191c43a39bR34 

If also snapping for 0px everything works *almost* beautifully.
There is still one edgecase stemming from now snapping to the first visible item when asked for a 0dp fling: The second item might occupy more of the screen so snapping to the less visible first item might seem forced.
To fix this we'd need to know the width of the items (to then see how much of each is showing). Passing this as an argument breaks encapsulation and seems a bit ugly - taking this value from the items or a `BoxWithConstraints`.

I would propose landing this as it is and check for feedback.

codrut-topliceanu (approved):
LGTM 👍

Mugurell:
Confirmed with Nicole that this looks good and can be merged.

Mugurell:
Seems that mergify added a new merge commit and forgot to squash it.
Will do it manually.

Issue #21578, opened 2021-09-29
State of the new Compose Awesomebar sometimes does not reveal tabs and menu buttons


AaronMT:
Build: main / d30583e0e700511bc2396decf6e955db499ad023 (09/29)

Sometimes after dismissing the virtual keyboard, the new Compose Awesomebar will not present itself as the expected state wherein I should be seeing both the tabs and menu buttons.

Sometimes after dismissing the keyboard (toggle back and forth), on AVD, I see:

To reproduce, click the address bar and tap the back button repeat until it gets into this state.
![135333142-b5d219e3-c7cb-4378-a8aa-7331ed4c5b88](https://user-images.githubusercontent.com/102331/135469438-a689c1cf-75aa-4ed9-aa7d-524964176438.png)

Where I should be seeing 

![135333142-b5d219e3-c7cb-4378-a8aa-7331ed4c5b88](https://user-images.githubusercontent.com/102331/135469473-624c77e1-8fef-4811-b169-d30f02ee8bee.png)

This is causing flakiness in UI tests where tests are expecting to see both buttons in tests.

(CC @pocmo)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-18078)

pocmo:
Unfortunately the screenshots are not visible (Is part of GitHub down?), so I am not really sure what's wrong. I'll try to follow the steps you mentioned and try to understand what's wrong.

sv-ohorvath:
@pocmo I was finally able to reproduce this manually as well:
Tap the search bar and tap the back button immediately so that the search view is not shown for too long. You will get this state
![searchBar](https://user-images.githubusercontent.com/4037709/135426977-93eac6d0-1598-4f70-a0a4-088dc4d3dcfd.png)

pocmo:
@sv-ohorvath I am not sure I fully understand. Is the issue that the "scan" and "search engine" actions are still visible? 🤔 I can reproduce this in the release version too, I think.

AaronMT:
Images re-uploaded, weird. But yes, that's correct, essentially sometimes dismissing the VKB will not reveal the menu/tab buttons, but just leave the toolbar in a state where the microphone button/scan and search engine buttons are visible.

pocmo:
> Images re-uploaded, weird.

They still don't show up for me? 😕 

It seems like the same behavior can be triggered in Beta and Release. But it seems like its now more likely to happen in our UI tests? 🤔

PR #21580, opened 2021-09-29
No issue: Fix AssertionError in openThreeDotMenu

AaronMT:
As a workaround for https://github.com/mozilla-mobile/fenix/issues/21578 let us catch the `AssertionError`, do a `goBack` to reveal the full toolbar, and finally `click`.


sv-ohorvath (approved):

sv-ohorvath:
This workaround will only cover this particular scenario in the case of the 3 dot menu, but it might not work for other future issues with the 3 dot button visibility. Let's try it for now and see what comes.

PR #21588, opened 2021-09-30
For #21392 - Update new strings descriptions

Mugurell:
As a followup to https://github.com/mozilla-l10n/android-l10n/pull/390#pullrequestreview-767108902

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -1933,7 +1933,7 @@
     <string name="pocket_stories_placeholder_text">Discover more</string>
     <!-- Title of an app feature. Smaller than a heading.-->
     <string name="pocket_stories_feature_title">Powered by Pocket</string>
-    <!-- Caption which concatenates the text of "pocket_stories_feature_learn_more"  -->
+    <!-- Caption for describing a certain feature. The placeholder is for a clickable text (eg: Learn more) which will load an url in a new tab when clicked.  -->
     <string name="pocket_stories_feature_caption">Part of the Firefox family. %s</string>
     <!-- Clickable text for opening an external link for more information about Pocket. -->
     <string name="pocket_stories_feature_learn_more">Learn more</string>


Delphine (approved):
LGTM thanks!

app/src/main/res/values/strings.xml
```diff
@@ -1920,7 +1920,7 @@
     <string name="pocket_stories_placeholder_text">Discover more</string>
     <!-- Title of an app feature. Smaller than a heading.-->
     <string name="pocket_stories_feature_title">Powered by Pocket</string>
-    <!-- Caption which concatenates the text of "pocket_stories_feature_learn_more"  -->
+    <!-- Caption for describing a certain feature. The placeholder is for a clickable text (eg: Learn more) which will load an url in a new tab when clicked.  -->
```
eliserichards, line 1923: I think we could make this better. "a certain feature" is a bit vague here.

```suggestion
    <!-- Placeholder caption for clickable text (eg: Learn more, which will load an url in a new tab when clicked) in the Pocket feature section of the home page. -->
```

Mugurell: This string is not a placeholder for clickable text. It is a text containing a placeholder which will be clickable.

If we are to combine the two and tie this string to the app feature we would get something like
"Caption for the Pocket recommended stories section of the home screen. Includes a placeholder for clickable text (eg: Learn more) which will load an url in a new tab when clicked"

But I wanted the description to be a bit vague as I don't expect localizers to necessarily be users of the app and know about it's features, where they are used - on the home screen and not sure why this would matter.
Curious if we should offer this much info or if we should avoid it.

@Delphine which of the 3 descriptions (already in PR / from Elise / from this comment) should be used?

eliserichards: That totally makes sense. Let's stick with what you have 👍

eliserichards (approved):

app/src/main/res/values/strings.xml
Delphine: For the record. Localizers are users of the app and are often core fans. They are not payed to do this, they are volunteers. And they need to understand in context what they are translating.

Delphine: Is there no onboarding for localization and working with communtiy?

mathjazz: Thanks for paying attention to string comments, they are very helpful for localizers!

If you look at [this example](https://pontoon.mozilla.org/sl/firefox/all-resources/?string=190587) in Firefox, a comment like "%s will be replaced with pocket_stories_feature_learn_more" would already be enough.

If you also want to explain where the string is used, I'd join it with your proposal in the comment:
"Caption for the Pocket recommended stories section of the home screen. %s will be replaced with pocket_stories_feature_learn_more."

I'd refrain from quoting the string explicitly in the comment (even as an example), because that gets outdated easily.

Issue #21593, opened 2021-09-30
Pocket recommended stories: Persistent topics selection


Mugurell:
Tracking https://mozilla-hub.atlassian.net/browse/FNXV2-17904

When people select a certain topic it should be persisted and not reset by a cold start.

SoftVision-LorandJanos:
Verified as fixed on the latest Nightly (94.0a1 - 2021-10-05T05:10:10.350256).

Devices used:

- Samsung Galaxy Tab S3 (Android 9).
- Huawei P30 (Android 10).
- Xiaomi Redmi 11 Lite (Android 11).

Closing the ticket as verified.

PR #21596, opened 2021-09-30
For #21593 - Persist selected categories status

Mugurell:
Refactored out `isSelected` from `PocketRecommendedStoriesCategory` to a new `PocketRecommendedStoriesSelectedCategory` which will be part of the same `HomeFragmentState` to support updating them independently.

Used Proto Datastore for persisting Pocket stories categories selections as a fast and easy solution with all the ACID requirements that also supports easy migrations if we later need to change the persisted data.

https://user-images.githubusercontent.com/11428869/135449582-6198449e-16d8-4a0a-ba0a-6e47b5df98f7.mp4

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/home/PocketUpdatesMiddleware.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/PocketUpdatesMiddleware.kt
@@ -13,6 +13,7 @@ import mozilla.components.lib.state.Action
 import mozilla.components.lib.state.Middleware
 import mozilla.components.lib.state.MiddlewareContext
 import mozilla.components.lib.state.Store
+import mozilla.components.service.pocket.PocketRecommendedStory
 import mozilla.components.service.pocket.PocketStoriesService
 import org.mozilla.fenix.datastore.SelectedPocketStoriesCategories
 import org.mozilla.fenix.datastore.SelectedPocketStoriesCategories.SelectedPocketStoriesCategory
@@ -60,13 +61,13 @@ class PocketUpdatesMiddleware(
         // Post process actions
         when (action) {
             is HomeFragmentAction.PocketStoriesShown -> {
-                coroutineScope.launch {
-                    pocketStoriesService.updateStoriesTimesShown(
-                        action.storiesShown.map {
-                            it.copy(timesShown = it.timesShown.inc())
-                        }
-                    )
-                }
+                persistStories(
+                    coroutineScope = coroutineScope,
+                    pocketStoriesService = pocketStoriesService,
+                    updatedStories = action.storiesShown.map {
+                        it.copy(timesShown = it.timesShown.inc())
+                    }
+                )
             }
             is HomeFragmentAction.SelectPocketStoriesCategory,
             is HomeFragmentAction.DeselectPocketStoriesCategory -> {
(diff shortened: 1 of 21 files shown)


Mugurell:
Draft for now. Have to add tests.

I'm though not sure that the proposed approach here with using SharedPreferences in which we write a `List<Pair<String, String>>` is really okay. 
It is a simple way to have it all working and also ensure easy cleanup and updates but the code is a bit fragile.
Storing a JSON seems marginally better.
A new synchronized atomic file is which to store a JSON is definitely more complex so not sure if that's worth it or not.

Thoughts?

Mugurell:
Talked with Sebastian about using DataStore for this.

Mugurell:
Another important observation from Sebastian was that reinjecting the selected state whenever the categories change is a bit awkward and it would be better to keep the categories and the selection in different properties in the state, so that updating the categories does not change the “selected” part of the state.

app/src/main/java/org/mozilla/fenix/datastore/Stores.kt
```diff
@@ -0,0 +1,17 @@
...
+package org.mozilla.fenix.datastore
+
+import android.content.Context
+import androidx.datastore.core.DataStore
+import androidx.datastore.dataStore
+
+/**
+ * Application / process unique [DataStore] for IO operations related to Pocket recommended stories selected categories.
+ */
+internal val Context.pocketStoriesSelectedCategoriesStore: DataStore<SelectedPocketStoriesCategories> by dataStore(
```
pocmo, line 14: Nit: Reading the code I noticed that we are now overloading the term "Store", which can be a `lib-state` `Store` or a `DataStore`. Maybe we should make that more obvious and name those things and related parameters `*DataStore` instead of `*Store`? 🤔

app/src/main/java/org/mozilla/fenix/home/PocketUpdatesMiddleware.kt
```diff
@@ -4,25 +4,63 @@
...
         action: HomeFragmentAction
     ) {
+        // Pre process actions
+        when (action) {
+            is HomeFragmentAction.PocketStoriesCategoriesChange -> {
+                // Whenever the stories categories are updated we need to restore their "selected" status.
+
+                coroutineScope.launch {
+                    selectedPocketCategoriesStore.data.collect { persistedSelectedCategories ->
+                        // Replace the original action with one that also
```
pocmo, line 44: "that also" ...? :) (It's also no longer "replacing" I guess)

app/src/main/java/org/mozilla/fenix/home/PocketUpdatesMiddleware.kt
```diff
@@ -4,25 +4,63 @@
...
+                                action.storiesCategories,
+                                persistedSelectedCategories.valuesList.map {
+                                    PocketRecommendedStoriesSelectedCategory(
+                                        name = it.name,
+                                        selectionTimestamp = it.selectionTimestamp
+                                    )
+                                }
+                            )
+                        )
+                    }
```
pocmo, line 56: nit: I tend to move this code that launches coroutines to private functions in the same file and then pass only the data/references they need to them. The advantage is that this code will no longer have access to the `MiddlewareContext` which only makes sense in the thread of the Middleware itself. For example see `EngineDelegateMiddleware`.

pocmo (approved):
LGTM.

app/src/main/java/org/mozilla/fenix/home/PocketUpdatesMiddleware.kt
Mugurell: Thinking of changing to
> // Intercept the original action which would only update categories and
// dispatch a new action which also updates which categories are selected by the user
// from previous locally persisted data.

Issue #21599, opened 2021-09-30
Pocket recommended stories: Add UTM links for opened Pocket recommendations


Mugurell:
- any recommendation link (url in the schema) needs to have ?utm_source=pocket-newtab-android appended.
- other links to pocket URLs within the android client (NOT provided by the API) need to have ?utm_source=ff_android appended.

PR #21601, opened 2021-09-30
For #21599 - Add UTM parameters for Pocket recommendations links

Mugurell:
UTMs confirmed with Priya from Pocket.

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesComposables.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesComposables.kt
@@ -6,6 +6,7 @@
 
 package org.mozilla.fenix.home.sessioncontrol.viewholders.pocket
 
+import android.net.Uri
 import androidx.compose.foundation.background
 import androidx.compose.foundation.isSystemInDarkTheme
 import androidx.compose.foundation.layout.Arrangement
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesComposables.kt
```diff
@@ -102,11 +107,15 @@ fun PocketStories(
             Column(Modifier.padding(end = if (rowIndex < gridLength - 1) 8.dp else 0.dp)) {
                 for (index in 0 until gridLength) {
                     columnItems.getOrNull(index)?.let { story ->
+                        val uri = Uri.parse(story.url)
```
Mugurell, line 110: It could be better to have this as an argument to `onExternalLinkClicked` instead of a separate property, in this only calculating the link when the story is clicked. Everytime the story is clicked. So .. not really sure what would be best.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesComposables.kt
```diff
@@ -116,11 +121,15 @@ fun PocketStories(
                 columnItems.forEachIndexed { rowIndex, story ->
                     if (story == placeholderStory) {
                         ListItemTabLargePlaceholder(stringResource(R.string.pocket_stories_placeholder_text)) {
-                            onExternalLinkClicked("http://getpocket.com/explore")
+                            onExternalLinkClicked("https://getpocket.com/explore?$POCKET_FEATURE_UTM_KEY_VALUE")
                         }
                     } else {
+                        val uri = Uri.parse(story.url)
```
gabrielluong, line 127: One idea here to wrap this as an extension function and add a test for this since it's a bit important that we ensure we are building the right URL. For example, `story.url.(to|as)[PocketStories]UTMLink()`

Mugurell: An extension to String seems excessive. It would just pollute it's namespace with something that is particular to this usecase. We can have the extension private to this file but there would be no guarantee that method is called.
I've created https://github.com/mozilla-mobile/fenix/issues/21680 for testing all composables.

gabrielluong (approved):

PR #21603, opened 2021-09-30
For #21551 - Add delete history metadata in the History view

### LINKED ISSUE #21551: Add delete history metadata in History view

gabrielluong:
Fixes #21551
### Pull Request checklist

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/components/history/PagedHistoryProvider.kt
```diff
@@ -37,6 +37,11 @@ class DefaultPagedHistoryProvider(
 
     private var historyGroups: List<History.Group>? = null
 
+    /**
+     * Map of [History.Metadata] to their corresponding [History.Regular].
+     */
+    private var metadataToVisitMap = mutableMapOf<History.Metadata, History.Regular>()
```
csadilek, line 43: There is a concurrency issue here. The `getHistory` call on the provider is invoked on a background thread, and if multiple threads call it concurrently they may read the wrong data (from a wrong provider or wrong `getHistory` call). We'd have to use a different key (incl. offset/numberOfItems) and use a concurrent map, if we're keeping this approach.

app/src/main/java/org/mozilla/fenix/components/history/PagedHistoryProvider.kt
```diff
@@ -129,9 +142,17 @@ class DefaultPagedHistoryProvider(
         }
         val historyMetadata = historyGroupsInOffset.flatMap { it.items }
 
-        // Add all items that are not in a group filtering out any matches with a history metadata
-        // item.
-        result.addAll(history.filter { item -> historyMetadata.find { it.url == item.url } == null })
+        // Add all history items that are not in a group filtering out any matches with a history
+        // metadata item.
+        for (historyItem in history) {
+            val matchedMetadata = historyMetadata.find { it.url == historyItem.url }
```
csadilek, line 148: This matching logic isn't sufficient yet. You could have `n` visits to the same page, and they are returned in desc order by visit timestamp. So you're basically always saving the oldest visit in the map currently which would then get deleted.

I think what we need is similar to the logic we already have, a mapping based on `visitedAt` and `createdAt` (for metadata). So for every metadata record we delete we will need to look for the corresponding visit and delete it too.

app/src/main/java/org/mozilla/fenix/components/history/PagedHistoryProvider.kt
csadilek: See my comment below though about mapping. I think we should not use a map here, but delete the closest visit by time.

app/src/main/java/org/mozilla/fenix/components/history/PagedHistoryProvider.kt
```diff
@@ -90,6 +91,31 @@ class DefaultPagedHistoryProvider(
...
+                VisitType.DOWNLOAD,
+                VisitType.REDIRECT_TEMPORARY,
+                VisitType.RELOAD,
+                VisitType.EMBED,
+                VisitType.FRAMED_LINK,
+                VisitType.REDIRECT_PERMANENT
+            )
+        )
+        return history.lastOrNull { it.url == historyMetadata.url }.also {
+            historyGroups = null
```
grigoryk, line 115: So, that's technically in the wrong place. it should be right after you call `delete` for the group itself.

I think in practice it'll work out but it's pretty weird having this reset here in the lookup. I.e. this simple lookup function doesn't know anything about group removals, why should it reset the state.

grigoryk (approved):
Just one tweak to where we reset `historyGroups` and I think this is good!

csadilek (approved):
🚢

PR #21633, opened 2021-10-01
Allow opening pocket stories when search dialog is active

csadilek:
Same as for recent bookmarks and top sites. We're gonna want to unify this logic soon, but for now let's make it work.


Mugurell (approved):
Thank you!

Issue #21635, opened 2021-10-01
 Preland strings for the tabs Auto-Close Message


MozillaNoah:
Strings can be found in https://mozilla-hub.atlassian.net/browse/FNXV2-18079

PR #21636, opened 2021-10-01
For #21635 - Preland strings for tab auto close message

MozillaNoah:
For #21635 

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -1897,8 +1897,8 @@
 	<!-- Dialog button text for canceling the rename top site prompt. -->
 	<string name="top_sites_rename_dialog_cancel">Cancel</string>
 
-    <!-- In-activate tabs in the tabs tray -->
-    <!-- Title text displayed in the tabs tray when a tab has been unused for 4 days. -->
+    <!-- Inactive tabs in the tabs tray -->
+    <!-- Title text displayed in the tabs tray when a tab has been unused for 14 days. -->
     <string name="inactive_tabs_title">Inactive tabs</string>
     <!-- Content description for closing all inactive tabs -->
     <string name="inactive_tabs_delete_all">Close all inactive tabs</string>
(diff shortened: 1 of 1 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -1908,6 +1908,13 @@
     <string name="inactive_tabs_30_days">30 days</string>
     <!-- The amount of time until a tab in the "inactive" section of the tabs tray will be closed. See string inactive_tabs_description as well -->
     <string name="inactive_tabs_7_days">1 week</string>
+    <!-- The header of the tab auto-close message -->
```
eliserichards, line 1911: I think we should make these new descriptions a bit more specific (see the "7 days" description)

MozillaNoah: These were added in July, but I'm honestly not sure these are even used. We currently don't allow the user to customize the inactive time period.

app/src/main/res/values/strings.xml
```diff
@@ -1897,8 +1897,8 @@
 	<!-- Dialog button text for canceling the rename top site prompt. -->
 	<string name="top_sites_rename_dialog_cancel">Cancel</string>
 
-    <!-- In-activate tabs in the tabs tray -->
-    <!-- Title text displayed in the tabs tray when a tab has been unused for 4 days. -->
+    <!-- Inactive tabs in the tabs tray -->
+    <!-- Title text displayed in the tabs tray when a tab has been unused for 14 days. -->
```
eliserichards, line 1901: Was this string added recently (last sprint)? If it's older we want to deprecate the old one and add this new one so that "14 days" can be translated correctly. Let me find the wiki..

eliserichards: https://github.com/mozilla-mobile/fenix/wiki/Remove-strings

app/src/main/res/values/strings.xml
eliserichards: Yeah, let's still update these descriptions. That way it's easier for the l10n folks to understand 😄

eliserichards:
I think we'll also need `tools:ignore="UnusedResources"` here for the strings that aren't being used yet

app/src/main/res/values/strings.xml
jonalmeida: I don't think we need to have these re-translated because the context is still the same as before.

eliserichards: I wasn't sure if changing the description would make it repopulate the string in l10n's system

eliserichards (approved):
🚀

PR #21656, opened 2021-10-01
Make search term grouping tolerant to (parent tab) navigation

csadilek:
Make sure search term groups are correct if parent (or child) navigation occurs.


grigoryk (approved):
We paired up on the actual fix, so giving this an r+ after finishing it up, fixing tests, adding a new one and writing comments.

Issue #21582, opened 2021-09-29
[Bug]: Thumbnail images on Jump Back In do not persist after closing/reopening Fenix


amedyne:
### Steps to reproduce

1. Have several tabs open
2. Press on Home button to go back Home
3. Thumbnails for the page are shown on the section under jump back in
4. Close/dismiss Fenix
5. Open the app 

### Expected behaviour

Thumbnail images should still be displayed in the JumpBackIn section

### Actual behaviour

Thumbnails are replaced with Favicons for items under the JumpBackIn section

### Device name

Google Pixel3 XL

### Android version

Android 11

### Firefox release type

Firefox Nightly

### Firefox version

94.0a1

### Additional information

See attached video

https://user-images.githubusercontent.com/68711330/135344851-33346765-dd78-4bcc-a350-18165457a1f0.mp4

.

jonalmeida:
We need to add an `if-else` branch [here](https://github.com/mozilla-mobile/fenix/blob/main/app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabs.kt#L247) to look through the `BrowserThumbnailsStorage` for screenshots of the tab before we make our final fallback to the site favicon.

LaurentiuApahideanSV:
Verified as fixed on Firefox Preview Nigthly 2021-10-06 and 94.0.0-beta.1.

### Devices used:
- Huawei MediaPad M3 (Android 7)
- Oppo Find X3 (Android 11)

PR #21661, opened 2021-10-02
Issue #21582: Use ThumbnailStorage in recent tabs on home

jonalmeida:
Might not be the best way to adding the thumbnail, but I'm open to other suggestions too. 🙂 

The `RecentSearchGroupItem` intentionally doesn't use the `tabId` to keep the group a bit more distinct from the other two items which are always a single tab.

Screen Shot 2021-10-02 at 1 46 47 AM

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabs.kt
```diff
@@ -299,3 +317,30 @@ private fun RecentTabSubtitle(subtitle: String) {
...
+    val size = LocalDensity.current.run { 116.dp.toPx().toInt() }
+    val request = ImageLoadRequest(tabId, size)
+    val storage = components.core.thumbnailStorage
+    val bitmap = rememberBitmap.value
+    LaunchedEffect(tabId) {
+        rememberBitmap.value = storage.loadThumbnail(request).await()?.asImageBitmap()
+    }
+    if (bitmap != null) {
+        val painter = BitmapPainter(bitmap)
+        Image(
```
gabrielluong, line 338: Maybe add a placeholder image. There's some examples around.

app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabs.kt
```diff
@@ -299,3 +317,30 @@ private fun RecentTabSubtitle(subtitle: String) {
...
+    alignment: Alignment
+) {
+    val rememberBitmap = remember(tabId) { mutableStateOf<ImageBitmap?>(null) }
+    val size = LocalDensity.current.run { 116.dp.toPx().toInt() }
+    val request = ImageLoadRequest(tabId, size)
+    val storage = components.core.thumbnailStorage
+    val bitmap = rememberBitmap.value
+    LaunchedEffect(tabId) {
+        rememberBitmap.value = storage.loadThumbnail(request).await()?.asImageBitmap()
+    }
```
gabrielluong, line 335:
```suggestion

    LaunchedEffect(tabId) {
        rememberBitmap.value = storage.loadThumbnail(request).await()?.asImageBitmap()
    }

```

gabrielluong (approved):

app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabs.kt
jonalmeida: I tried looking for an example, but there are none that work for my use case.

Issue #21611, opened 2021-09-30
Jump back in Contextual Hints for Firefox Android 94 Release


Amejia481:
https://mozilla-hub.atlassian.net/browse/FNXV2-17841

Amejia481:
QA team please help us to verify that after one of this sections is visible Jump back in, Recently saved, Recent explorations and Pocket a on boarding dialog is displayed and after the **CRF dialog is displayed.** Thanks in advance.

![image](https://user-images.githubusercontent.com/773158/135862269-a281a86e-4ac1-40b9-bcc9-52de7a4943ab.png)

LaurentiuApahideanSV:
Verified as fixed on Firefox Preview Nightly 2021-10-05.

### Devices used:
- Oppo Find X3 (Android 11)
- OnePlus 6T (Android 9)

PR #21667, opened 2021-10-02
 For #21611: Add jump back in Contextual Hints

Amejia481:
image

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/onboarding/HomeOnboardingDialogFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/onboarding/HomeOnboardingDialogFragment.kt
@@ -4,14 +4,22 @@
 
 package org.mozilla.fenix.onboarding
 
+import android.app.Dialog
+import android.content.Context
+import android.graphics.Color
+import android.graphics.drawable.ColorDrawable
 import android.os.Bundle
+import android.view.Gravity
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
 import androidx.fragment.app.DialogFragment
+import androidx.recyclerview.widget.RecyclerView
 import org.mozilla.fenix.R
 import org.mozilla.fenix.databinding.FragmentOnboardingHomeDialogBinding
+import org.mozilla.fenix.databinding.OnboardingJumpBackInCfrBinding
 import org.mozilla.fenix.ext.settings
+import org.mozilla.fenix.home.recenttabs.view.RecentTabsHeaderViewHolder
 
 /**
  * Dialog displayed once when one or multiples of these sections are shown in the home screen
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mozilla/fenix/onboarding/HomeOnboardingDialogFragment.kt
```diff
@@ -37,7 +45,61 @@ class HomeOnboardingDialogFragment : DialogFragment() {
...
+
+    private fun showJumpCFR() {
+        val jumpBackInView = findJumpBackInView()
+        jumpBackInView?.let {
+            val crfDialog = createJumpCRF(anchor = jumpBackInView)
+            crfDialog.show()
+        }
+    }
+
+    private fun findJumpBackInView(): View? {
```
jonalmeida, line 61: Neat! 😄

app/src/main/java/org/mozilla/fenix/onboarding/HomeOnboardingDialogFragment.kt
```diff
@@ -37,7 +45,61 @@ class HomeOnboardingDialogFragment : DialogFragment() {
...
+            setContentView(popupBinding.root)
+            setCancelable(false)
+            // removing title or setting it as an empty string does not prevent a11y services from assigning one
+            setTitle(" ")
+        }
+        popupBinding.closeInfoBanner.setOnClickListener {
+            popup.dismiss()
+        }
+
+        anchor.getLocationOnScreen(anchorPosition)
```
jonalmeida, line 90: TIL!

It looks like this method defaults to (0, 0) if it can't find a view. Would we need an early-return check in case we can't find this item? Maybe the RecyclerView adapter has the item, but it's not on the screen because it doesn't fit?

I genuinely do not know since I've never used this approach before, but just throwing this out there. 🙂

app/src/main/java/org/mozilla/fenix/onboarding/HomeOnboardingDialogFragment.kt
```diff
@@ -37,7 +45,61 @@ class HomeOnboardingDialogFragment : DialogFragment() {
...
+        popupBinding.root.measure(View.MeasureSpec.UNSPECIFIED, View.MeasureSpec.UNSPECIFIED)
+
+        popup.window?.let {
+            val attr = it.attributes
+            it.setGravity(Gravity.START or Gravity.TOP)
+            attr.x = x
+            attr.y = y - popupBinding.root.measuredHeight
+            it.attributes = attr
+            it.setBackgroundDrawable(ColorDrawable(Color.TRANSPARENT))
+        }
```
jonalmeida, line 102: non-blocking nit: Use `apply` when _applying_ changes to the item.
```suggestion
       popup.window?.apply {
            val attr = attributes
            setGravity(Gravity.START or Gravity.TOP)
            attr.x = x
            attr.y = y - popupBinding.root.measuredHeight
            attributes = attr
            setBackgroundDrawable(ColorDrawable(Color.TRANSPARENT))
        }
```

app/src/main/res/drawable/onboarding_popup_shape.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
```
jonalmeida, line 1: nit: license.

jonalmeida (approved):
One question/consideration, otherwise, this looks good to me in case you want to land this a bit pre-emptively.

app/src/main/java/org/mozilla/fenix/onboarding/HomeOnboardingDialogFragment.kt
Amejia481: Thanks for catching this, I added the early-return check  :)

app/src/main/res/drawable/onboarding_popup_shape.xml
Amejia481: Fixed :)

Issue #21632, opened 2021-10-01
Preland strings for the inactive tabs survey


gabrielluong:
Strings can be found in https://mozilla-hub.atlassian.net/browse/FNXV2-17964

PR #21669, opened 2021-10-03
Issue #21632: Correct inactive tabs survey string

jonalmeida:


st3fan (approved):
As discussed on #l10n-community 🙌🏻

Issue #21642, opened 2021-10-01
Do not display Media tab under Jump Back In 


amedyne:
This change is to disable displaying of additional Media tab under Jump Back In.

SoftVision-LorandJanos:
Verified as fixed on the latest Nightly (95.0a1 - 2021-10-06T05:11:35.245780) and on the latest Beta 94.0.0 beta1.

Devices used:

Samsung Galaxy Tab S3 (Android 9).
Xiaomi Redmi 11 Lite (Android 11).

Closing the ticket as verified.

PR #21670, opened 2021-10-03
Issue #21642: Remove in-progress media tab from homescreen

jonalmeida:
Not removing the other logic because we're going to re-add this after some refinement in the future.

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/ext/BrowserState.kt
+++ b/app/src/main/java/org/mozilla/fenix/ext/BrowserState.kt
@@ -24,15 +24,8 @@ import kotlin.math.max
 fun BrowserState.asRecentTabs(): List<RecentTab> {
     return mutableListOf<RecentTab>().apply {
         val lastOpenedNormalTab = lastOpenedNormalTab
-        val inProgressMediaTab = inProgressMediaTab
 
         lastOpenedNormalTab?.let { add(RecentTab.Tab(it)) }
-
-        if (inProgressMediaTab == lastOpenedNormalTab) {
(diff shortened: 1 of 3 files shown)


gabrielluong (approved):

app/src/main/java/org/mozilla/fenix/ext/BrowserState.kt
```diff
@@ -24,16 +24,9 @@ import kotlin.math.max
 fun BrowserState.asRecentTabs(): List<RecentTab> {
     return mutableListOf<RecentTab>().apply {
         val lastOpenedNormalTab = lastOpenedNormalTab
-        val inProgressMediaTab = inProgressMediaTab
```
gabrielluong, line 27: We could also wrap this behind a feature flag?

jonalmeida: Yeah, I wasn't sure to what extent we should be "removing it". So I opted for this solution so we can land it before the beta cut.

jonalmeida: Another reason I forgot about, is that we get intermittent test failures when we have feature flags here. See https://github.com/mozilla-mobile/fenix/pull/21557.

csadilek: Let's land as is. Thought about this last week and it's a trade-off. We don't want feature flags to be too fine-grained, so arguably this is a change in behaviour of an existing features as opposed to a feature itself. We can revert this to enable again when we want it (have time to add media controls).

csadilek (approved):

Issue #21621, opened 2021-10-01
[Bug]: Pocket stories horizontal shadows are clipped


Mugurell:
### Steps to reproduce

App's homescreen - Pocket recommended stories section

### Expected behaviour

Horizontal shadows should be shown the same as they do for the above recent tab items.

### Actual behaviour

Horizontal shadows are clipped.

<img src="https://user-images.githubusercontent.com/11428869/135585813-a0341e62-0a7a-4052-9d04-5120630af96d.png" width="50%" />

### Android version

All

### Firefox release type

Firefox Nightly

### Firefox version

Latest Nightly

### Additional information

Based on https://issuetracker.google.com/issues/186318448#comment4 clipping main axis shadows is the expected behavior but maybe we can work around this.

Mugurell:
After talking with topotropic about the fact that the "Recently bookmarked" section is not clipped and so it can show shadows but also extend to the edge of the screen when scrolled it might be worth to remove the 16dp padding from the list and have each child declare it's own.
In this case Pocket could declare 0 dp for the composeView and 16 dp for the content needing that which would ensure enough space in which to show the shadow.
Have to test.

Mugurell:
In the interim Nicole proposed adding a 4dp padding to the stories so that they'd have a bit of shadow.

SoftVision-LorandJanos:
Verified as fixed on the latest Nightly (95.0a1 - 2021-10-06T05:11:35.245780) and on the latest Beta 94.0.0 beta1.

Devices used:
- Samsung Galaxy Tab S3 (Android 9).
- Xiaomi Redmi 11 Lite (Android 11).

Closing the ticket as verified.

Issue #6629, opened 2019-11-15
[Product Quality] Private Browsing Mode - Checklist


AmyYLee:
**Feature Level Checklist for Private Browsing Mode**

This is an ongoing list for product quality related bugs for this feature/component. Any non-blocking UX issues should be added here (bugs that impact quality of product but do not hinder functionality or accessibility). Blocking UX bugs should be filed separately and triaged by product. Items in this list that are fixed should be checked off by UX (Engineering please ping UX owner when an item is fixed).

----------------------------------------------------
**Checklist**

- [ ] - 1. Header copy should read “Private tabs”
- [ ] - 2. Increase line spacing (leading) to 20px. Text should be 14px
- [ ] - 3. Snackbar colour should be #312A65 (same as light theme)
- [ ] - 4. Body copy should left align to header “Private session”

![PBM Edits](https://user-images.githubusercontent.com/5992746/69365922-49f04480-0c63-11ea-94d1-5a8d6a40f857.png)

mcarare:
@AmyYLee For 5: The text was changed according to https://github.com/mozilla-mobile/fenix/issues/3604#issuecomment-539679640. Do you want to change it back to "deleted? TY!

AmyYLee:
> @AmyYLee For 5: The text was changed according to [#3604 (comment)](https://github.com/mozilla-mobile/fenix/issues/3604#issuecomment-539679640). Do you want to change it back to "deleted? TY!

Thanks for bringing this to my attention. Please leave it as "tab closed". I'll remove it from the checklist.

abodea:
Verified as fixed on the latest Nightly build from **11/29** with **Google Pixel 3a XL(Android 10).**

PR #21687, opened 2021-10-04
For #21621 - Add invdividual paddings to each home item

Mugurell:
Removed the 16dp padding from the RecyclerView holding all home items and added the appropriate spacing to each individual item.
This all is needed to be able to use [contentPadding](https://developer.android.com/jetpack/compose/lists#content-padding) for the list showing Pocket recommended stories in such allowing the stories (when scrolled) and their shadows to extend on the main axis until screen width.

https://user-images.githubusercontent.com/11428869/135870190-d9054321-e59d-48a2-91db-bc1d2d858554.mp4

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesComposables.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesComposables.kt
@@ -18,7 +18,7 @@ import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.width
 import androidx.compose.foundation.lazy.LazyRow
-import androidx.compose.foundation.lazy.itemsIndexed
+import androidx.compose.foundation.lazy.items
 import androidx.compose.foundation.lazy.rememberLazyListState
 import androidx.compose.material.Icon
 import androidx.compose.material.Text
@@ -33,9 +33,9 @@ import androidx.compose.ui.semantics.semantics
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.tooling.preview.PreviewParameter
 import androidx.compose.ui.tooling.preview.PreviewParameterProvider
+import androidx.compose.ui.unit.Dp
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.sp
-import mozilla.components.concept.fetch.Client
 import mozilla.components.service.pocket.PocketRecommendedStory
 import mozilla.components.ui.colors.PhotonColors
 import org.mozilla.fenix.R
@@ -100,13 +100,15 @@ fun PocketStory(
  * to go to Pocket for more recommendations are added.
  *
  * @param stories The list of [PocketRecommendedStory]ies to be displayed. Expect a list with 8 items.
- * @param client [Client] instance to be used for downloading the story header image.
+ * @param contentPadding Dimension for padding the content after it has been clipped.
+ * This space will be used for shadows and also content rendering when the list is scrolled.
  * @param onExternalLinkClicked Callback for when the user taps an element which contains an
  * external link for where user can go for more recommendations.
  */
 @Composable
 fun PocketStories(
     @PreviewParameter(PocketStoryProvider::class) stories: List<PocketRecommendedStory>,
+    contentPadding: Dp,
     onExternalLinkClicked: (String) -> Unit
 ) {
     // Show stories in at most 3 rows but on any number of columns depending on the data received.
@@ -117,14 +119,14 @@ fun PocketStories(
     val flingBehavior = EagerFlingBehavior(lazyRowState = listState)
 
     LazyRow(
-        contentPadding = PaddingValues(start = 8.dp, end = 8.dp),
+        contentPadding = PaddingValues(horizontal = contentPadding),
         state = listState,
         flingBehavior = flingBehavior,
         horizontalArrangement = Arrangement.spacedBy(8.dp)
     ) {
-        itemsIndexed(storiesToShow) { _, columnItems ->
-            Column {
-                columnItems.forEachIndexed { rowIndex, story ->
+        items(storiesToShow) { columnItems ->
+            Column(verticalArrangement = Arrangement.spacedBy(8.dp)) {
+                columnItems.forEach { story ->
                     if (story == placeholderStory) {
                         ListItemTabLargePlaceholder(stringResource(R.string.pocket_stories_placeholder_text)) {
                             onExternalLinkClicked("http://getpocket.com/explore")
@@ -133,10 +135,6 @@ fun PocketStories(
                         PocketStory(story) {
                             onExternalLinkClicked(story.url)
                         }
-                    }
-
-                    if (rowIndex < maxRowsNo - 1) {
-                        Spacer(modifier = Modifier.height(8.dp))
                     }
                 }
             }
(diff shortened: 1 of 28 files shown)


app/src/main/res/layout/private_browsing_description.xml
```diff
@@ -9,7 +9,8 @@
     android:layout_height="wrap_content"
     android:layout_margin="0.5dp"
```
gabrielluong, line 10: I am wondering if we need to keep this 0.5dp margin that will now only apply to the vertical margins. At the same time, the margin is so small I am wondering if we can double check the old designs to see if this is needed since it's such a small margin. 

This can be a follow-up.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesComposables.kt
```diff
@@ -122,14 +126,14 @@ fun PocketStories(
...
-        contentPadding = PaddingValues(start = 8.dp, end = 8.dp),
+        contentPadding = PaddingValues(horizontal = contentPadding),
         state = listState,
         flingBehavior = flingBehavior,
         horizontalArrangement = Arrangement.spacedBy(8.dp)
     ) {
         itemsIndexed(storiesToShow) { _, columnItems ->
-            Column {
-                columnItems.forEachIndexed { rowIndex, story ->
+            Column(verticalArrangement = Arrangement.spacedBy(8.dp)) {
```
gabrielluong, line 135: Ah yes, this is good stuff when we don't need that Spacer anymore.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesComposables.kt
```diff
@@ -236,19 +244,20 @@ private fun PocketStoriesComposablesPreview() {
...
                     },
-                    emptyList()
-                ) { }
+                    emptyList(),
+                    { }
+                )
                 Spacer(Modifier.height(10.dp))
 
-                PoweredByPocketHeader { }
+                PoweredByPocketHeader({ })
```
gabrielluong, line 260: Could we set `onExternalLinkClicked: (String) -> Unit = {}` so we can avoid the `({ })` here?

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesComposables.kt
```diff
@@ -153,20 +153,25 @@ fun PocketStories(
...
+    onCategoryClick: (PocketRecommendedStoriesCategory) -> Unit,
+    modifier: Modifier = Modifier
 ) {
-    StaggeredHorizontalGrid(
-        horizontalItemsSpacing = 16.dp
-    ) {
-        categories.filter { it.name != POCKET_STORIES_DEFAULT_CATEGORY_NAME }.forEach { category ->
-            SelectableChip(category.name, selections.map { it.name }.contains(category.name)) {
-                onCategoryClick(category)
+    Box(modifier = modifier) {
```
gabrielluong, line 167: This makes me wonder when we should pass `Modifier` as a param. Personally, I would opt to inline the modifier in this instance until `PocketStoriesCategories` needs to be used in multiple places where its layout needs to be modified  for each of the different views. Otherwise, I find it easier to see what the layout parameters are if it was inlined. Passing modifiers also becomes a bit trickier if for example we needed another modifier param for `StaggeredHorizontalGrid`. Usually I am thinking about how we can keep the list of params optimal given its usage.

I had a similar debate with doing this when I was adding `RecentTabImage` https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabs.kt#230 where it did make sense that we want to pass in different sizes for the image.

gabrielluong: > This makes me wonder when we should pass `Modifier` as a param. Personally, I would opt to inline the modifier in this instance until `PocketStoriesCategories` needs to be used in multiple places where its layout needs to be modified for each of the different views. Otherwise, I find it easier to see what the layout parameters are if it was inlined. Passing modifiers also becomes a bit trickier if for example we needed another modifier param for `StaggeredHorizontalGrid`. Usually I am thinking about how we can keep the list of params optimal given its usage.
> 
> I had a similar debate with doing this when I was adding `RecentTabImage` https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabs.kt#230 where it did make sense that we want to pass in different sizes for the image.

I don't think we need to necessary change this. I think this is more food for thought and can help us align with what we feel is a good practise.

app/src/main/res/layout/fragment_home.xml
```diff
@@ -69,7 +69,8 @@
         android:clipChildren="false"
         android:clipToPadding="false"
         android:layout_marginBottom="88dp"
-        android:padding="16dp"
+        android:paddingTop="16dp"
```
gabrielluong, line 72: Is there `paddingVertical` equivalent?

app/src/main/res/layout/no_collections_message.xml
```diff
@@ -7,6 +7,7 @@
     android:id="@+id/no_collections_wrapper"
     android:layout_width="match_parent"
     android:layout_height="wrap_content"
+    android:layout_marginHorizontal="@dimen/home_item_horizontal_margin"
```
gabrielluong, line 10: Is it safe to assume you went through each of the viewholder in SessionControlAdapter and found their respect xml and added the `android:layout_marginHorizontal` for each?

gabrielluong (approved):

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesComposables.kt
Mugurell: Elise saw this improvement. Indeed it's nice!

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesComposables.kt
Mugurell: The click listener is a required parameter which I'd prefer to keep as such.
Although the code here and it's usage is straightforward and we'd know now to actually pass a listener and not rely on the default I feel it's important to keep the composable code exposed to clients as clean and tight as possible while for the private code and here previews we can maybe a bit uglier setup.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesComposables.kt
Mugurell: Thank you!
Indeed I also though about this a little.
What I went with is allowing the ViewHolder to inject the padding to each composable the same as is being done in XML (where we have a top level paddings / margins set)
For Pocket we needed to go one level lower - set the spacing to each layout child instead of the layout parent because we needed to set this on the LazyRow. But for the other two composables ViewHolders I set the spacing in the ComposeView - the same as we'd do in XML for the parent layout.
The idea being that the ViewHolder controls the dimensions of each composable. And then each composable is free to layout how it wants in those dimensions.

app/src/main/res/layout/fragment_home.xml
Mugurell: There is. Thank you!

app/src/main/res/layout/no_collections_message.xml
Mugurell: Yes. First looked in the `SessionControlView` (reason for the commits split) then in `SessionControlAdapter`. Should've covered everything on homescreen.

app/src/main/res/layout/private_browsing_description.xml
Mugurell: I see this implemented in https://github.com/mozilla-mobile/fenix/issues/6629 with no talk about the margin value.
I'd say that after so much time (~2 years) if it wasn't in the specs then in became so in the meantime. 😂.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/pocket/PocketStoriesComposables.kt
```diff
@@ -122,14 +126,14 @@ fun PocketStories(
...
+        contentPadding = PaddingValues(horizontal = contentPadding),
         state = listState,
         flingBehavior = flingBehavior,
         horizontalArrangement = Arrangement.spacedBy(8.dp)
     ) {
         itemsIndexed(storiesToShow) { _, columnItems ->
-            Column {
-                columnItems.forEachIndexed { rowIndex, story ->
+            Column(verticalArrangement = Arrangement.spacedBy(8.dp)) {
+                columnItems.forEachIndexed { _, story ->
```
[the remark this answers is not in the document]
Mugurell: Can also replace `itemsIndexed` with `items` above.

PR #21698, opened 2021-10-04
Move tabs out of search group if direct load occurs

csadilek:
This regressed in our previous fix that made sure child tabs don't mistakenly get moved out of the group if their parent is navigated away, or in case the child tabs are redirected.

However, when a subsequent load occurs in any tab in the group the search terms need to be cleared and the tab removed from the group to prevent false positives.


grigoryk (approved):
Looks good! Paired up on this.

PR #21701, opened 2021-10-04
For #21694 - Fix the layout in the Recent Searches

### LINKED ISSUE #21694: Fix the layout in the Recent Searches
- We are currently only NOT showing the divider line for the last item in a column in Recent Searches. So, this means you will see a divider line even if there's only 1 item in Recent Searches. We should show a divider in between items only.
Screen Shot 2021-10-04 at 4 04 13 PM

- Remove clip(RoundedCornerShape(8.dp) around the Column in Recent Searches. This causes the click gestures to round for the first and last items, which doesn't seem like what we want. https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/java/org/mozilla/fenix/historymetadata/view/RecentlyVisited.kt#8.
This is the green clip for illustration.
Screen Shot 2021-10-04 at 4 09 33 PM

gabrielluong:
Fixes #21694

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/historymetadata/view/RecentlyVisited.kt
+++ b/app/src/main/java/org/mozilla/fenix/historymetadata/view/RecentlyVisited.kt
@@ -85,7 +85,7 @@ fun RecentlyVisited(
                         RecentVisitItem(
                             recentVisit = recentVisit,
                             menuItems = menuItems,
-                            showDividerLine = index < VISITS_PER_COLUMN - 1,
+                            showDividerLine = index < items.size - 1,
                             onRecentVisitClick = onRecentVisitClick
                         )
                     }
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/fenix/historymetadata/view/RecentlyVisited.kt
```diff
@@ -118,8 +117,7 @@ private fun RecentVisitItem(
                 onClick = { onRecentVisitClick(recentVisit) },
                 onLongClick = { menuExpanded = true }
             )
-            .size(268.dp, 56.dp)
-            .background(color = FirefoxTheme.colors.surface),
```
gabrielluong, line 122: This removes a reductant background modifier since it's already defined in the parent Card.

MozillaNoah (approved):

PR #21719, opened 2021-10-05
No issue: Fix closePrivateTabTest intermittent test

AaronMT:


app/src/androidTest/java/org/mozilla/fenix/ui/robots/TabDrawerRobot.kt
```diff
@@ -399,13 +399,12 @@ private fun threeDotMenu() = onView(withId(R.id.tab_tray_overflow))
...
-                .textContains(title)
+        mDevice.waitNotNull(
+            Until.findObjects(By.textContains(title)), waitingTime
+        )
+        mDevice.waitNotNull(
+            Until.findObjects(By.res("$packageName:id/mozac_browser_tabstray_title")), waitingTime
         )
-            .waitForExists(waitingTime)
-
         tab(title).check(matches(isDisplayed()))
```
sv-ohorvath, line 408: The test actually failed here, at this Espresso matcher:
private fun tab(title: String) =
    onView(
        allOf(
            withId(R.id.mozac_browser_tabstray_title),
            withText(title)
        )
    )

Also sometimes fails at:
private fun assertCloseTabsButton(title: String) =
    onView(
        allOf(
            withId(R.id.mozac_browser_tabstray_close),
            withContentDescription("Close tab $title")
        )
    )
        .check(matches(withEffectiveVisibility(ViewMatchers.Visibility.VISIBLE)))

If you want to replace them with the findObjects selector, you can do something like:
mDevice.findObjects(By.res("$packageName:id/mozac_browser_tabstray_title")) - this returns a list of objects matching the resource name.

Issue #21680, opened 2021-10-04
Pocket recommended stories: Tests


Mugurell:
We should ensure proper testing of all composables used for this feature.

sheikh-azharuddin:
I see some commits related to pocket stories...how to enable the feature?

Mugurell:
> I see some commits related to pocket stories...how to enable the feature?
There should already be a "Though-provoking stories" section on Nightly's homescreen.

sheikh-azharuddin:
> > I see some commits related to pocket stories...how to enable the feature?
> 
> There should already be a "Though-provoking stories" section on Nightly's homescreen.

But this is not visible for me..
Rog3 android 11

![Screenshot_20211005-132026468](https://user-images.githubusercontent.com/58127841/135982314-9bd9d381-2945-4647-bfb3-901c4bcee551.jpg)

Mugurell:
> > > I see some commits related to pocket stories...how to enable the feature?
> > 
> > 
> > There should already be a "Though-provoking stories" section on Nightly's homescreen.
> 
> But this is not visible for me.. Rog3 android 11

Only to en-US users, only if enabled in the customization settings.
Please open a new ticket if this conditions are met and and the section does not appear on the homescreen.

sheikh-azharuddin:
> > > > I see some commits related to pocket stories...how to enable the feature?
> > > 
> > > 
> > > There should already be a "Though-provoking stories" section on Nightly's homescreen.
> > 
> > 
> > But this is not visible for me.. Rog3 android 11
> 
> Only to en-US users, only if enabled in the customization settings. Please open a new ticket if this conditions are met and the section does not appear on the homescreen.

After changing the launguage to English US , I got pocket stories
I am from india

PR #21731, opened 2021-10-06
For #21680 - Pocket composables testing using Robolectric

Mugurell:
Important notes:
- to "instantiate" and test composables we need to use an `AndroidComposeTestRule` for an activity in which to add them.
The default activity `createComposeRule` is trying to use - `androidx.activity.ComponentActivity` is not found in the manifest so I went with using an empty Robolectric built instance of `ComponentActivity` in which we'll add the composables we need testing. 
If this approach becomes a pattern we might create our own rule for this
- testing `PocketStories` wasn't possible since just trying to add that with `setContent` would throw a `org.robolectric.android.internal.AndroidTestEnvironment$UnExecutedRunnablesException: Main looper has queued unexecuted runnables.`. Not sure why and how to get past this, needs more investigations.
- finders (onNodeWith..) are cumbersome to use and don't seem to reach in the entire graph, they only go a few levels deep. Haven't found a way to increase this. 
- for composables without a text or contentDescription we'll need to add a `testTag` to easily use them in tests.
Created an accompanying `onFirstChildWithTag()` for then easily finding them through recursion checking only the first child - this being useful for when the composables we need are the content of others serving as decorations.
- assertions seem a bit cumbersome - failed assertions on text values do not provide an immediately straight-forward way to see the difference. Might want to have our own.
- this is more of a poc. We'll probably have to adapt a bit the way we design tests - what scenarios need testing and how to organize the test.

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/compose/ListItemTabLarge.kt
+++ b/app/src/main/java/org/mozilla/fenix/compose/ListItemTabLarge.kt
@@ -18,6 +18,8 @@ import androidx.compose.material.Card
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
+import androidx.compose.ui.platform.testTag
+import androidx.compose.ui.semantics.semantics
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
 import org.mozilla.fenix.theme.FirefoxTheme
@@ -104,7 +106,10 @@ private fun ListItemTabSurface(
     onClick: (() -> Unit)? = null,
     tabDetails: @Composable () -> Unit
 ) {
-    var modifier = Modifier.size(328.dp, 116.dp)
+    var modifier = Modifier
+        .size(328.dp, 116.dp)
+        .testTag("ListItemTabLarge")
+        .semantics { url = imageUrl }
     if (onClick != null) modifier = modifier.then(Modifier.clickable { onClick() })
 
     Card(
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -576,6 +576,7 @@ dependencies {
     testImplementation Deps.mozilla_support_test_libstate
     testImplementation Deps.androidx_junit
     testImplementation Deps.androidx_work_testing
+    testImplementation Deps.androidx_compose_ui_test
     testImplementation (Deps.robolectric) {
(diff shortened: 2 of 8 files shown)


app/src/debug/AndroidManifest.xml
```diff
@@ -19,6 +19,10 @@
     <application
             tools:replace="android:name"
             android:name="org.mozilla.fenix.DebugFenixApplication">
+
```
Mugurell, line 22: This definetly shouldn't be needed but although we're using `Deps.androidx_compose_ui_test_manifest` it seems the instrumentation can't find `ComponentActivity`.
Will have to look more in how the manifests are merged.

app/src/main/java/org/mozilla/fenix/compose/ListItemTabLarge.kt
```diff
@@ -104,7 +105,9 @@ private fun ListItemTabSurface(
     onClick: (() -> Unit)? = null,
     tabDetails: @Composable () -> Unit
 ) {
-    var modifier = Modifier.size(328.dp, 116.dp)
+    var modifier = Modifier
+        .size(328.dp, 116.dp)
+        .testTag("ListItemTabLarge")
```
Mugurell, line 110: Looks like we'll have to add `testTag`s to all custom composables which are more complex than just a text / contentDescription.

Mugurell:
Thinking of also adding a custom semantics property for the URL which might be used in various composables.

app/src/debug/AndroidManifest.xml
```diff
@@ -19,6 +19,10 @@
...
+        <activity
+            android:name="androidx.activity.ComponentActivity"
```
Mugurell, line 24: This definetly shouldn't be needed but although we're using Deps.androidx_compose_ui_test_manifest it seems the instrumentation can't find ComponentActivity.
Will have to look more in how the manifests are merged.

Mugurell:
Curious if the CI will be green.

Mugurell:
@gabrielluong @csadilek Currently the compose code is not tested. Robolectric can help with fasts unit tests without having to run these on a device. Curious on whether this approach sounds good for the future or what will be the testing strategy for compose code.

Mugurell:
Guess this can be closed but serve as a reference point for later.

PR #21805, opened 2021-10-08
Closes #21759: Do not render home fragment when launching to tab

### LINKED ISSUE #21759: 39-127ms COLD VIEW start up regression around 10/1
_This is a breakout from https://github.com/mozilla-mobile/fenix/issues/21749._

We see a regression between 9/29 and 10/1:
- VIEW: 39ms

The regression continues to rise. From 9/29 to 10/5, we see a regression of:
- VIEW: 127ms

VIEW (note the scales are different):
image

This seems like it could be  be two discrete regressions – one between 9/29 & 10/1 and one on 10/5. [Dashboard link](https://earthangel-b40313e5.influxcloud.net/d/s3AT6t7nk/fenix-startup-nightly-via-backfill-py?orgId=1&from=now-30d&to=now). The numbers seem a little noisy so let me know if you think it's worth retaking them. Maybe we should start with a profile though.

### Profiles
VIEW:
- 9/29 VIEW: https://share.firefox.dev/2YoJS2W
- 10/5 VIEW: https://share.firefox.dev/3uUxbcn

One notable visual difference between these builds is the inclusion of Pocket content. There was also an update the synchronized the layouts for the homescreen. These could be the two events I mentioned above. However, this shouldn't affect the VIEW code path: let's find out what's going on.

pocmo:
This patch replaces the `startDestination` of our `navGraph` with an empty fragment instead of the home fragment. In situations where we navigate immediately (to the browser) this avoids creating all the views for the home fragment first. We still navigate to the browser via home though - to build up the back stack.

So when launching to the browser, instead of having to go through `HomeFragment.onCreate()` and `HomeFragment.onCreateView()` we only invoke `HomeFragment.onCreate()`.


mcomella:
I tried running this with Christian but I found a bug:
- Force-stop the app: `adb shell am force-stop org.mozilla.fenix.debug`
- Start the app with a VIEW intent: `adb shell am start -a android.intent.action.VIEW -d 'https://example.com' -n org.mozilla.fenix.debug/org.mozilla.fenix.IntentReceiverActivity`

Expected: opens to website/tab
Actual: opens to homescreen

mcomella:
> I did see that it's possible to set the start destination dynamically: https://developer.android.google.cn/reference/androidx/navigation/NavGraph#setStartDestination(kotlin.Int)
>
>But that's not really what we want anymore, as we always want to navigate to home first (to add it to the back stack), then optionally to the browser fragment. By navigating instead of starting on home we prevented rendering the home view unconditionally. So, what we have here seems better.
>
> @pocmo @mcomella Do you agree?

I'm not familiar with the navigation API so I can't give a confident answer. It sounds like you have an assumption that the `start destination` will always be inflated, whether set in XML or dynamically. If that's true (and it seems reasonable though I'd test it) then I agree with you that what we have is the most performant when using the navigation library.

pocmo:
Tested a rebased version of this patch again. Didn't find any issues/regressions. However in yesterday's discussion we were more in favor of working on `HomeFragment` performance improvements instead (since they are needed anyways), which will improve the situation here too - but avoid the additional complexity from this (potentially fragile?) navigation workaround.

mcomella:
> However in yesterday's discussion we were more in favor of working on `HomeFragment` performance improvements instead (since they are needed anyways), which will improve the situation here too - but avoid the additional complexity from this (potentially fragile?) navigation workaround.

fwiw, 55% of the time spent in `HomeFragment` methods is inflation. When we switch to Compose, that should all go away. Caveat: I only took one profile so the exact value can vary.

mcomella:
I forgot to include my profile from the previous comment: https://share.firefox.dev/3mOcnRy

pocmo:
Over the weekend I was wondering: With the current setup... are we navigating to home in custom tabs / PWA too and we just never see it? If that's the case, and as long as those two activities are intertwined, maybe it is worth to add the startup fragment for that.

mcomella:
I took this profile and confirmed we load the HomeFragment for PWAs: https://share.firefox.dev/3oledJf

pocmo:
@csadilek Let's rediscuss. Not rendering the `HomeFragment` for all "external" tabs may be worth it?

csadilek:
We discussed this again today and are planning to land this after the next beta cut to make sure we get a fully Nightly cycle.

pocmo:
Tested a rebased version of this patch. Looks good. No problems. Let's land in this new Nightly cycle. :)

mcomella:
See the issue https://github.com/mozilla-mobile/fenix/issues/21759#issuecomment-992901952 for the measured improvement from this PR.

Issue #21707, opened 2021-10-05
[Bug]: Tab group, when opening a search tab, Other tabs appears above ordinary tabs


Cheap-Skate:
### Steps to reproduce
1. Search for cats and open a result in a new tab (creating a search group).
1. Search for dogs and open a result in a new tab (creating a search group).
2. Open the tabs tray and delete one tab from either group.

### Expected behaviour

- Title above ordinary tabs is seen above.
- No title above ordinary tabs UNLESS there is a tab group above ordinary tabs

### Actual behaviour

The title above ordinary tabs is not seen when there are normal tabs OR title above ordinary tabs when a single search tab is opened

### Device name

Galaxy A40

### Android version

Android 11

### Firefox release type

Firefox Nightly

### Firefox version

Today's

### Additional information

https://user-images.githubusercontent.com/1370580/136284826-ed2fa82b-28f4-4c73-b003-94ebf651bb5e.mp4

SoftVision-LorandJanos:
Hi @Cheap-Skate ! Thank you for the report.
I could reproduce this issue when I had lots of tabs, I then deleted the group above the "Other tabs" and repeated the steps and couldn't reproduce it anymore.
After that I cleared all the browsing data and started from scratch, accessing a few sites and then made a search, and got to tabs tray. There was no tab title above those tabs.
Could you make a video of the issue you described? Or maybe give me some extra relevant details if you discover what could help reproducing it for me as well. Maybe I needed the longer history that I had the first time I checked? I also had Inactive tabs enabled in the Tabs tray, I closed that as well. So please give me some extra details about this. Thank you!

Verified on the latest Nightly (94.0a1 - 2021-10-05T05:10:10.350256).

Devices used:

- Samsung Galaxy Tab S3 (Android 9).
- Huawei P30 (Android 10).
- Xiaomi Redmi 11 Lite (Android 11).

Cheap-Skate:
I updated to today's Nightly and the problem no longer occurs. So it maybe fixed or may be intermittent? 

Leave this bug open for a few days, I will see if it comes back.

Thanks

SoftVision-LorandJanos:
Well as I said I was able to reproduce it at first, that was after an update to the current version, but it seemed kind of an edge case. If you discover clear steps to get to the issue from scratch, please share. Thanks again!

SoftVision-LorandJanos:
Verified on the latest Nightly 101.0a1 build (2022-04-12).
[...]

PR #21882, opened 2021-10-13
Issue #21707: Correct search grouping logic in extensions

jonalmeida:
At this moment, we have two extension methods that have duplicate
functionality to construct search term groupings. One on [`List<Tab>`](https://github.com/mozilla-mobile/fenix/blob/9439a65e1e09b29f2e1a2ddee797da3241a93b14/app/src/main/java/org/mozilla/fenix/tabstray/browser/TabSorter.kt#L110-L134) and
one on [`List<TabSessionState>`](https://github.com/mozilla-mobile/fenix/blob/9439a65e1e09b29f2e1a2ddee797da3241a93b14/app/src/main/java/org/mozilla/fenix/ext/BrowserState.kt#L78-L101). The former is used for everything
related to tabs piped through the `TabsFeature` and the latter is for
consumers of `BrowserState` directly.

The bug occurs because our implementation of search groupings was
updated only on the former extension, but the `HeaderBinding`, that
observes the BrowserState and updates the title visibility, was using
the latter.

Ideally, we remove this duplication when we no longer have separate data
classes for consumers of `TabsFeature`, but this intermediary fix should
suffice.

**Before fix:**

https://user-images.githubusercontent.com/1370580/137075539-ccf25e06-9869-4cc7-8cc4-d7c83679ee75.mp4 

**After fix:**

https://user-images.githubusercontent.com/1370580/137075563-8fbf648e-6303-4d65-9f0f-024ceba0a653.mp4

1. click on Show All Checks,


jonalmeida:
Hmm.. we don't need to uplift this any more into beta.. so I think I'll cancel the backport PR since I don't know how to tell mergify to stop doing that part. 🙂

jonalmeida:
Argh, I force pushed to the wrong branch..

PR #21885, opened 2021-10-13
Closes #21871 - Eagerly update UI state after search group removal

### LINKED ISSUE #21871: Home screen search group removal tweaks
There are three problems with the group removals on home screen right now:
1) it appears the UI update after removal races with the actual delete operation; it looks like the delete didn't happen (group is still there), but going in and out of home screen clears the group from the screen. the this was called out in the review but we didn't address it then.
2) UI update after tapping "remove" scrolls the home screen to the top; it's quite jarring.
3) after removing a search group, the same group still remains present in the tabs tray and in "jump back in" section.

For (1), we should dispatch a store update alongside async delete operations, to update the UI state right away.
For (2), try to see if we can avoid scrolling the screen.
For (3), we should "ungroup" existing tabs - we can dispatch a metadata key update action with searchTerms/referrer set to `null`. The behaviour this should achieve is the following: removing a search group from the home screen will ungroup related tabs in the tabs tray (and move them to "other tabs"), and remove the search from the "jump back in" section.

Patches for (1) and (3) will need to be uplifted to current beta 94.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-18273)

grigoryk:
Before this patch, this was the behavior - 'remove' button is clicked, we'd ask
the storage to remove metadata (on its IO thread), then navigate to Home
Screen.

This resulted in a race we could end-up on the Home Screen before delete
finishes, so the search groups do not appear to be removed (but,
refreshing the Home Screen again shows that they are removed).

This also resulted in an unnecessary navigation which felt very janky
(screen will "scroll" to the top) and was way more work than necessary.

After this patch, we:
 - dispatch two actions (on browserstore, on homefragmentstore) which
   remove the search groups from any relevant in-memory state; any UI bound to
   this state will be automatically "refreshed"
 - no longer navigate as part of the remove action, so the UI doesn't
   move and removal happens "in-place"

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/historymetadata/controller/HistoryMetadataController.kt
+++ b/app/src/main/java/org/mozilla/fenix/historymetadata/controller/HistoryMetadataController.kt
@@ -9,12 +9,15 @@ import androidx.annotation.VisibleForTesting.PRIVATE
 import androidx.navigation.NavController
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.launch
+import mozilla.components.browser.state.action.HistoryMetadataAction
+import mozilla.components.browser.state.store.BrowserStore
 import mozilla.components.concept.storage.HistoryMetadataStorage
 import org.mozilla.fenix.R
-import org.mozilla.fenix.browser.BrowserFragmentDirections
 import org.mozilla.fenix.historymetadata.HistoryMetadataGroup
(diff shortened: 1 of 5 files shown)


grigoryk:
Landing this in https://github.com/mozilla-mobile/fenix/pull/21895 to combine the toolchain task burden (both need to bump the A-C version).

Issue #21854, opened 2021-10-11
Long delay (487ms+) during scrolling to load Pocket content on the homescreen; also slow on cold start


mcomella:
## Steps to reproduce
- Open enough homescreen sections to have Pocket below the fold (on my Moto G5, I just need to have one tab open for the "Jump back in" section to show)
- Scroll down to reveal Pocket section

### Expected behavior
Scrolling is smooth & section appears immediately

### Actual behavior
Scrolling stops as the main thread is blocked to load the Pocket content and contributes to a feeling of jank. Here's a profile: https://share.firefox.dev/3mKBZxb

According to the profile, the initial `dispatchTouchEvent` runs for 487ms (a very long time) so the delay is likely to last even longer than that. Here's a screen capture:

https://user-images.githubusercontent.com/759372/136838225-d2007a9f-5663-4a50-b3bf-dbd02e05fe1c.mp4

This delay also affects **the cold start experience** where it takes a long time to load the homescreen to visual completeness:

https://user-images.githubusercontent.com/759372/136838599-19d42bab-a1b4-4291-be3a-8b01ad9ddc62.mp4

### Device information

* Android device: Moto G5
* Fenix version: Nightly 10/10

---

To address both issues raised above, I think we need to consider one or both of:
- optimizing the layout to be fast enough to load 1) during the scroll step and 2) without delaying visual completeness on the homescreen for too long
- synchronously load a lightweight, incomplete layout in place of the full Pocket layout (a "skeleton layout") and asynchronously load the remainder of the content into it, never blocking the main thread for too long

CC @Mugurell

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-18266)

Mugurell:
Splitting the current composable from `PocketStoriesViewHolder` into more (we could have even 3 separate composables there: stories, categories and the pocket header) would probably mean some gains (if `composeView.setContent` is not expensive in it's own).
The other thing to look for is ensuring there are no unneeded recompositions.

Another important aspect I think is how much of an effect does this have on the most used devices.
Compose will be a bit slower than xml (at least for now) but on my Pixel3 (which has a more powerful hw than Moto G5) i'm seeing some glitches in debug but no issue on Nightly.

mcomella:
> Another important aspect I think is how much of an effect does this have on the most used devices.

[...]

Issue #8561, opened 2020-02-19
[Bug] Top sites does not appear on screen on cold start if many tabs are open


mcomella:
## Steps to reproduce
- Open many tabs (I have 7 on P2)
- Click app list and swipe Fenix closed
- Open Fenix

### Expected behavior
Top sites are visible

### Actual behavior
Top sites are not visible. If I scroll up, top sites are revealed. It seems they're being hidden by the scroll container for some reason.

### Device information

* Android device: P2
* Fenix version: Nightly 200219

ecsmyth:
Reproduces for me

## Actual behavior
- 5+ open tabs, three rows of top sites are hidden
- 4 open tabs, two rows of top sites are hidden
- 3 open tabs, first row of top sites is hidden
- 1-2 open tabs, all three rows of top sites are visible
- With seven tabs open, three rows of top sites are hidden.

## Device information
- Android device: Samsung Galaxy S10 Int'l
- Fenix version: Nightly 200220

codrut-topliceanu:
**Cause:**

Top sites are added to the recycleView from HomeFragment _after_ the view is created and populated at start, so the scroll position doesn't change, leaving the top sites outside the visible area.

sflorean:
Verified as fixed on Fenix Nightly 200330 with Samsung Galaxy Note10 (Android 10) and Xiaomi Mi Pad 2 (Android 5.1).

PR #22067, opened 2021-10-21
For #21854 - Split the big ComposeView in 3 smaller ones

Mugurell:
This would shorten the time needed to layout all Pocket recommended stories
content in one go, though it may lead to shorten hiccups over a bigger period
of time.

Also focused on keeping recompositions to a minimum, something that having 3
separate composables also helps with.

&nbsp;

Tested on the oldest device I found - https://www.gsmarena.com/allview_p5_emagic-7997.php
| before | after |
| --- | --- |
| <video src="https://user-images.githubusercontent.com/11428869/138274165-8d7228aa-9872-4a14-ac0e-ace1a0f124f2.mov" /> | <video src="https://user-images.githubusercontent.com/11428869/138273128-34cd584c-bac8-4b89-afa2-1cd12f6e90c3.mov" /> |
| [profile](https://drive.google.com/file/d/1g7s7MR7tEcX8XKFP0bK1WEtEsioZQB6s/view?usp=sharing) | [profile](https://drive.google.com/file/d/1TFHuOFn-HqwIAvzL4qopV_SF8SFcDfIu/view?usp=sharing)

&nbsp;

### Pull Request checklist

1. click on Show All Checks,


mcomella:
@Mugurell What's the status of this PR? Are we still interested in landing it?

Mugurell:
> @Mugurell What's the status of this PR? Are we still interested in landing it?

I am interested in landing this but not sure how reviewers feel about it, whether I should make any changes or this can land.

@pocmo , @gabrielluong ?

gabrielluong:
@mcomella Since we talked about this on Friday, I am wondering if we should also do benchmarking to assess the perf improvement and get some concrete data for this PR.

gabrielluong:
To help move this along, I propose 2 actions here to help the entire team learn and assess the perf impact of these changes since this has been a recent theme to learn about perf + compose and develop a perf culture. 

1) I think the point of splitting the big ComposeView into multiple ViewHolders is pretty clear cut as its own PR. I think we should separate that into its own PR and do a benchmark to assess perf % changes prior to landing.
2) I have been most hesitant around the changes to passing State as an argument to the composable functions and was realistically looking for someone to review that bit. My main question around this is "Why?" as framed in https://github.com/mozilla-mobile/fenix/pull/22067#discussion_r734701930. I would suggest making that change its own PR and also doing a separate benchmark for this change so that we can understand the perf impact separate from the multiple ViewHolders change.

This will ultimately give us some numbers to work with and put a number on what kind of changes will yield the best improvements.

jonalmeida:
> Tried to reproduce the issue with top sites being pushed off screen on an emulator with the same screen size and resolution as the Moto G5 thinking it might influence things but for me it doesn't reproduce.
> 
> Based on the video posted (awesome that we can slow it down) it seems like the top sites and pocket sections are shown first then when the "Jump back in" section appears it pushes both the top sites and pocket sections away to have enough room for it to be displayed.
> 
> Remembered #8561 which seems like the same issue. I'd think the issue relates to state updates which don't trigger scrolling to top sites but not sure what the exact conditions for it are.

🚗 Drive-by comment: This sounds similar to the work Roger is doing in [FNXV2-18479](https://mozilla-hub.atlassian.net/browse/FNXV2-18479).

mcomella:
> @mcomella Since we talked about this on Thursday, I am wondering if we should also do benchmarking to assess the perf improvement and get some concrete data for this PR.

CC @MarcLeclair – this improvement relates to the homescreen work we've been talking about with gabrielluong

app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesViewHolder.kt
```diff
@@ -29,39 +29,47 @@ import org.mozilla.fenix.home.HomeFragmentStore
...
+                        modifier = Modifier
+                            .fillMaxWidth()
+                            .padding(horizontal = horizontalPadding.dp)
+                            .wrapContentHeight(align = Alignment.Top)
+                    )
+
+                    Spacer(Modifier.height(17.dp))
+
+                    PocketStories(
+                        store.observeAsComposableState { state -> state.pocketStories },
```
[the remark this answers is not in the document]
pocmo: I went down a similar rabbit hole for the AwesomeBar. I was wondering whether I should pass down the list of `Suggestion`s or `State<List<Suggestion>>`. With similar logs I also saw that observing the state at the lowest level avoided going down the callstack. However what I am not sure about yet is: Does it really matter? A function call is cheap. And compose will still not recompose other child calls whose parameters haven't changed - as I understand it. So it doesn't really seem expensive. Maybe. I haven't continued from there yet. If we continue this all the way down (where possible), wouldn't we end up with UI toolkit replacements like `Text(text: State<String>, ...)`?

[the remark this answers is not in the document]
gabrielluong: I can see that as a potential deep rabbit hole. It would be interesting to benchmark this to see what the results indicate.

Mugurell:
@gabrielluong , @mcomella Seems like there was no update regarding the review of this. Do we still need this? Should I rebase it or close the PR?

gabrielluong:
> @gabrielluong , @mcomella Seems like there was no update regarding the review of this. Do we still need this? Should I rebase it or close the PR?

Took a look at it. I think the changes look sensible. Feel free to rebase it and I will review if this helps us with improving performance.

app/src/main/java/org/mozilla/fenix/home/pocket/PocketCategoriesViewHolder.kt
```diff
@@ -0,0 +1,109 @@
...
+            )
+
+            Spacer(Modifier.height(17.dp))
+
+            PocketStoriesCategories(
+                "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor".split(" ").map {
+                    PocketRecommendedStoriesCategory(it)
+                },
+                emptyList(),
+                { }
```
gabrielluong, line 105: Let's add the named params here since we already do it everywhere else.

app/src/main/java/org/mozilla/fenix/home/pocket/PocketCategoriesViewHolder.kt
```diff
@@ -0,0 +1,109 @@
...
+
+            PocketStoriesCategories(
+                categories = categories ?: emptyList(),
+                selections = categoriesSelections ?: emptyList(),
+                onCategoryClick = interactor::onCategoryClicked,
+                modifier = Modifier
+                    .fillMaxWidth()
+                    .padding(horizontal = horizontalPadding.dp)
+            )
+        }
```
gabrielluong, line 77: Wondering if it's sensible to just refactor this into a public @Composable

app/src/main/java/org/mozilla/fenix/home/pocket/PocketRecommendationsHeaderViewHolder.kt
```diff
@@ -0,0 +1,63 @@
...
+    companion object {
+        val LAYOUT_ID = View.generateViewId()
+    }
+}
+
+@Composable
+@Preview
+fun PocketRecommendationsFooterViewHolderPreview() {
+    FirefoxTheme {
+        PoweredByPocketHeader({ })
```
gabrielluong, line 61:
```suggestion
        PoweredByPocketHeader({})
```

app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesViewHolder.kt
```diff
@@ -52,94 +53,63 @@ class PocketStoriesViewHolder(
...
+                    .wrapContentHeight(align = Alignment.Top)
+            )
+
+            Spacer(Modifier.height(17.dp))
+
+            PocketStories(
+                stories = getFakePocketStories(POCKET_STORIES_TO_SHOW_COUNT),
+                contentPadding = 0.dp,
+                onStoryClicked = { _, _ -> },
+                onDiscoverMoreClicked = { }
```
gabrielluong, line 111:
```suggestion
                onDiscoverMoreClicked = {}
```
This isn't quite a big deal. I was just searching through the AC/Fenix codebase to get a sense of what was more common. I mostly just want us to be cognizant about being consistent about some of these code styles, but ideally have a linter to enforce a style.

app/src/main/java/org/mozilla/fenix/home/pocket/PocketRecommendationsHeaderViewHolder.kt
```diff
@@ -0,0 +1,63 @@
...
+    composeView: ComposeView,
+    viewLifecycleOwner: LifecycleOwner,
+    private val interactor: PocketStoriesInteractor
+) : ComposeViewHolder(composeView, viewLifecycleOwner) {
+
+    @Composable
+    override fun Content() {
+        val horizontalPadding = with(composeView.resources) {
+            getDimensionPixelSize(R.dimen.home_item_horizontal_margin) / displayMetrics.density
+        }
```
gabrielluong, line 38: I just curious about this piece of code. I am wondering what is the difference between doing this vs other places (which might be actually wrong) that only has `val horizontalPadding = composeView.resources.getDimensionPixelSize(R.dimen.home_item_horizontal_margin)`

app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesViewHolder.kt
```diff
@@ -18,32 +17,34 @@ import androidx.compose.ui.Alignment
...
 import org.mozilla.fenix.home.HomeFragmentStore
+import org.mozilla.fenix.theme.FirefoxTheme
 
 internal const val POCKET_STORIES_TO_SHOW_COUNT = 8
-internal const val POCKET_CATEGORIES_SELECTED_AT_A_TIME_COUNT = 8
 
 /**
- * [RecyclerView.ViewHolder] that will display a list of [PocketRecommendedStory]es
- * which is to be provided in the [bind] method.
+ * [RecyclerView.ViewHolder] for displaying the list of [PocketRecommendedStory]s from [store].
```
gabrielluong, line 36:
```suggestion
 * [RecyclerView.ViewHolder] for displaying the list of [PocketRecommendedStory]s from the [HomeFragmentStore].
```

app/src/main/java/org/mozilla/fenix/home/pocket/PocketRecommendationsHeaderViewHolder.kt
Mugurell: I was transforming
> density independent pixel -> pixels -> `compose.ui.unit.Dp`
>  xml: 16                      ->  getDimension: 60    ->  divided by density: 16

because dp is what `Modifier.padding` expected and I forgot compose has a `dimensionResource` that would get us that value without the cumbersome transformation. 

In the other cases I see the old `View.padding` api is used that takes pixels as argument. (in the example above 60).

I'll switch to use `composeView.setPadding(..)` for consistency here also (except for in `PocketStoriesViewHolder` which should occupy the full width for the content to have shadows).

app/src/main/java/org/mozilla/fenix/home/pocket/PocketCategoriesViewHolder.kt
Mugurell: Refactored. Thanks!

app/src/main/java/org/mozilla/fenix/home/pocket/PocketCategoriesViewHolder.kt
```diff
@@ -0,0 +1,111 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+@file:Suppress("MagicNumber")
```
gabrielluong, line 5: Wondering if we need this suppression.

app/src/main/java/org/mozilla/fenix/home/pocket/PocketCategoriesViewHolder.kt
```diff
@@ -0,0 +1,111 @@
...
+) {
+    Column {
+        SectionHeader(
+            text = stringResource(R.string.pocket_stories_categories_header),
+            modifier = Modifier
+                .fillMaxWidth()
+                .wrapContentHeight(align = Alignment.Top)
+        )
+
+        Spacer(Modifier.height(17.dp))
```
gabrielluong, line 87: This is an aside, but we should probably address this. I know from talking with UX @Kate-Galetski that what we see on Figma isn't actually what UX sees when it comes to these padding. I think Figma counts the border or something when we try to examine these spacing between elements. This should probably be 16dp. The rule of thumb is that if you see an odd number then that is probably suspicious and most likely wrong when you look in our codebase.

app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesViewHolder.kt
```diff
@@ -52,94 +54,61 @@ class PocketStoriesViewHolder(
...
+                    .wrapContentHeight(align = Alignment.Top)
+            )
+
+            Spacer(Modifier.height(17.dp))
+
+            PocketStories(
+                stories ?: emptyList(),
+                horizontalPadding,
+                interactor::onStoryClicked,
+                interactor::onDiscoverMoreClicked
```
gabrielluong, line 85: Let's add the named parameters

app/src/main/java/org/mozilla/fenix/home/pocket/PocketCategoriesViewHolder.kt
```diff
@@ -0,0 +1,111 @@
...
+    composeView: ComposeView,
+    viewLifecycleOwner: LifecycleOwner,
+    private val store: HomeFragmentStore,
+    private val interactor: PocketStoriesInteractor
+) : ComposeViewHolder(composeView, viewLifecycleOwner) {
+
+    @Composable
+    override fun Content() {
+        val horizontalPadding =
+            composeView.resources.getDimensionPixelSize(R.dimen.home_item_horizontal_margin)
```
gabrielluong, line 52: You used `dimensionResource ` in PocketStoriesViewHolder. Can we also do that here?

app/src/main/java/org/mozilla/fenix/home/pocket/PocketCategoriesViewHolder.kt
```diff
@@ -0,0 +1,111 @@
...
+        val topPadding = composeView.dpToPx(24f)
```
gabrielluong, line 53: Can we do this using a Spacer instead? I think this would make it easier to parse down the road when we can remove these ViewHolders.

app/src/main/java/org/mozilla/fenix/home/pocket/PocketCategoriesViewHolder.kt
```diff
@@ -0,0 +1,111 @@
...
+
+    companion object {
+        val LAYOUT_ID = View.generateViewId()
+    }
+}
+
+@Composable
+private fun PocketTopics(
+    categories: List<PocketRecommendedStoriesCategory>?,
+    categoriesSelections: List<PocketRecommendedStoriesSelectedCategory>?,
```
gabrielluong, line 76: Can we change the signature of this so that the `categories` and `categoriesSelections` are not nullable, and we can just pass in the `emptyList()` in `Content()`

app/src/main/java/org/mozilla/fenix/home/pocket/PocketRecommendationsHeaderViewHolder.kt
```diff
@@ -0,0 +1,59 @@
...
+    composeView: ComposeView,
+    viewLifecycleOwner: LifecycleOwner,
+    private val interactor: PocketStoriesInteractor
+) : ComposeViewHolder(composeView, viewLifecycleOwner) {
+
+    @Composable
+    override fun Content() {
+        val horizontalPadding =
+            composeView.resources.getDimensionPixelSize(R.dimen.home_item_horizontal_margin)
+        val topPadding = composeView.dpToPx(24f)
```
gabrielluong, line 39: Same prior comments as above.

app/src/main/java/org/mozilla/fenix/home/pocket/PocketCategoriesViewHolder.kt
Mugurell: See https://github.com/mozilla-mobile/fenix/pull/22067#discussion_r788979262.
This value will be used in setting the value for a View and refers to `pixel`s. The same is used in `RecentTabViewHolder` or in `RecentlyVisitedViewHolder`.
In PocketStoriesViewHolder the values is used for setting composables padding and refers to `Dp`s.

app/src/main/java/org/mozilla/fenix/home/pocket/PocketRecommendationsHeaderViewHolder.kt
Mugurell: Added the `Spacer`. Kept the pixels instead of dps.

app/src/main/java/org/mozilla/fenix/home/pocket/PocketCategoriesViewHolder.kt
Mugurell: Created https://github.com/mozilla-mobile/fenix/issues/23551

app/src/main/java/org/mozilla/fenix/home/pocket/PocketCategoriesViewHolder.kt
Mugurell: Thanks. Removed it.

gabrielluong (approved):

Mugurell:
Waiting for the beta cut.

app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesViewHolder.kt
```diff
@@ -52,94 +54,61 @@ class PocketStoriesViewHolder(
...
+        val stories = store
+            .observeAsComposableState { state -> state.pocketStories }.value
 
-    val categoriesSelections = store
-        .observeAsComposableState { state -> state.pocketStoriesCategoriesSelections }.value
+        LaunchedEffect(stories) {
+            // We should report back when a certain story is actually being displayed.
+            // Cannot do it reliably so for now we'll just mass report everything as being displayed.
+            stories?.let {
+                interactor::onStoriesShown
```
jknair0, line 66: @Mugurell  i think there is a problem here! you forgot to invoke this method. instead used a method reference

PR #22087, opened 2021-10-21
Close #21692: Sort tabs after adding search term tabs without group

### LINKED ISSUE #21692: [Bug]: Tab Group - an ungrouped Search tab always stays at the bottom of the tab list
### Steps to reproduce

Switch to Grid view for tabs (probably also true for List view but I haven't checked)
Open a tab and search for something eg "cat"
(Do not open any links from the "cat" search so it is a normal tab not a group)
The "cat" search tab is last in the tab list
-> correct behavior
Open another tab and navigate to a normal site eg mozilla.org
Go to the tabs tray
The "cat" search tab is still last in the tab list
-> incorrect behavior, the last tab should be mozilla.org since it is the most recently opened

### Expected behaviour

The "cat" search tab should behave like any other tab since it is not a group, so more recently opened tabs should come after it in the list

### Actual behaviour

The "cat" search tab remains the last tab in the list even when I open newer tabs

### Device name

Galaxy A40

### Android version

Android 11

### Firefox release type

Firefox Nightly

### Firefox version

Today's

### Device logs

_No response_

### Additional information

_No response_

rocketsroger:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/tabstray/browser/TabSorter.kt
+++ b/app/src/main/java/org/mozilla/fenix/tabstray/browser/TabSorter.kt
@@ -55,7 +55,7 @@ class TabSorter(
         concatAdapter.tabGroupAdapter.submitList(groups)
 
         // Normal tabs.
-        val totalNormalTabs = (normalTabs + remainderTabs)
+        val totalNormalTabs = (normalTabs + remainderTabs).sortedBy { it.lastAccess }
         val selectedTabIndex = totalNormalTabs.findSelectedIndex(selectedTabId)
         concatAdapter.browserAdapter.updateTabs(Tabs(totalNormalTabs, selectedTabIndex))
     }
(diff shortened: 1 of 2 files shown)


jonalmeida:
I'm not sure we should do this. What it implies is that every time you select a (normal) tab, that tab will move to a new place. This requires careful consideration. Have we spoken to product/UX about this?

rocketsroger:
> I'm not sure we should do this. What it implies is that every time you select a (normal) tab, that tab will move to a new place. This requires careful consideration. Have we spoken to product/UX about this?

That's true.  I didn't think about that.  I'll have to think about this more.  Thanks for pointing this out.

jonalmeida:
> Would it be a matter of just sorting by `createdAt,` or would it require something more complex?

`createdAt` may work, however it disrupts the order of parent/child tabs. You can create a child tab next to a parent much later on, but that tab needs to stay adjacent to the parent and not last in the list because it is the newest.

jonalmeida:
To clarify, the parent/child sorted relationship is already done in AC and if we try re-write all of that we're creating more work for us. I suggest we punt on this for now since @csadilek will be working on an RFC to upstream these various (for lack of a better word) groupings so we do not need to re-sort things differently in the app.

PR #22243, opened 2021-11-01
Close #20704: Add AppStore to Components 

### LINKED ISSUE #20704: Introduce an app store to hold state for the lifecycle of the app
In https://github.com/mozilla-mobile/fenix/pull/20687/, we spoke about the value of an AppStore to hold state that needs to live as long as the app (or longer than the fragment).

With the introduction of UI state like the collapsible inactive tabs list, this is a good opportunity to create one and add it in there.

jonalmeida:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/AppStore.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/AppStore.kt
@@ -1,0 +1,22 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.components
+
+import mozilla.components.lib.state.Middleware
+import mozilla.components.lib.state.Store
+import org.mozilla.fenix.components.appstate.AppAction
+import org.mozilla.fenix.components.appstate.AppState
+import org.mozilla.fenix.components.appstate.AppStoreReducer
+
+/**
+ * A [Store] that holds the [AppState] for the app and reduces [AppAction]s
+ * dispatched to the store.
+ *
+ * This store is not persisted to disk and is scoped to the life-cycle of the application.
+ */
+class AppStore(
+    initialState: AppState = AppState(),
+    middlewares: List<Middleware<AppState, AppAction>> = emptyList()
+) : Store<AppState, AppAction>(initialState, AppStoreReducer::reduce, middlewares)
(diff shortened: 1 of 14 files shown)


app/src/main/java/org/mozilla/fenix/components/AppStore.kt
```diff
@@ -0,0 +1,22 @@
...
+import mozilla.components.lib.state.Store
+import org.mozilla.fenix.components.appstate.AppAction
+import org.mozilla.fenix.components.appstate.AppState
+import org.mozilla.fenix.components.appstate.AppStoreReducer
+
+/**
+ * A [Store] that holds the [AppState] for the app and reduces [AppAction]s
+ * dispatched to the store.
+ *
+ * This store is not persisted to disk and is meant to live the lifetime of the application.
```
csadilek, line 17: nit: mean to live the entire lifetime? (or: is scoped to the life-cycle of the application)

app/src/main/java/org/mozilla/fenix/tabstray/browser/InactiveTabViewHolder.kt
```diff
@@ -37,7 +37,7 @@ sealed class InactiveTabViewHolder(itemView: View) : RecyclerView.ViewHolder(ite
 
         init {
             itemView.apply {
-                isActivated = InactiveTabsState.isExpanded
+                isActivated = itemView.context.components.appStore.state.inactiveTabsExpanded
```
csadilek, line 40: The actions are processed asynchronously using `lib-state`. Usually, we'd want all consumers to observer the store (incl. it's initial state). Is this good enough here to simply read directly?

app/src/main/java/org/mozilla/fenix/tabstray/browser/InactiveTabsAdapter.kt
```diff
@@ -92,7 +93,7 @@ class InactiveTabsAdapter(
         }
 
         // If we have items, but we should be in a collapsed state.
-        if (!InactiveTabsState.isExpanded) {
+        if (!context.components.appStore.state.inactiveTabsExpanded) {
```
csadilek, line 96: Same q as above re: observing state.

csadilek (commented):
Good idea to introduce this for app-wide in-memory state that we don't want to persist!

I have one question below re: the store processing actions asynchronously.

csadilek (approved):

app/src/main/java/org/mozilla/fenix/tabstray/browser/InactiveTabViewHolder.kt
jonalmeida: In this case, I think it's fine to just take the state. In our worst case, the tabs tray is not expanded the first time a user re-enters the tabstray if we haven't processed the async action, but it will in the following case.

app/src/main/java/org/mozilla/fenix/tabstray/browser/InactiveTabsAdapter.kt
jonalmeida: See above.

app/src/main/java/org/mozilla/fenix/components/AppStore.kt
jonalmeida: Fixed!

PR #22246, opened 2021-11-01
Introduce test rule to work around intermittent mockk failures

csadilek:
This adds a specific test rule to retry tests in case they run into the mockk bug. With the current default configuration, it will retry twice, which should be good enough, based on the workaround described in the linked ticket (call `unmockAll` and try again).

We can also use this test rule to preload the problematic classes, in case this isn't good enough. See similar workaround in mockk: https://github.com/mockk/mockk/commit/1fe93559c68116bd6fb2ed4936c765e11298a7b2

@jonalmeida @pocmo wdyt? let's try this?

### DIFF
--- a/app/src/test/java/org/mozilla/fenix/browser/FenixSnackbarDelegateTest.kt
+++ b/app/src/test/java/org/mozilla/fenix/browser/FenixSnackbarDelegateTest.kt
@@ -14,17 +14,21 @@ import io.mockk.unmockkObject
 import io.mockk.verify
 import org.junit.After
 import org.junit.Before
-import org.junit.Ignore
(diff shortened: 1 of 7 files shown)


jonalmeida (approved):
This looks great! Let's land after a few retries for safety.

csadilek:
5. run ✅

csadilek:
7. run ✅ 

I'd like to see the retry rule kicking in though. Nothing in the logs of the last two builds.

csadilek:
9. run ✅ . The new rule kicked it on that last build, and the test passed on second attempt.

csadilek:
10. run ✅

PR #22285, opened 2021-11-03
ThreadPenaltyDeathWithIgnoresListener: Ignore violations without Mozilla code in the stack trace.

pocmo:
I still keep running into strict mode crashes when deploying Fenix debug builds (to a Pixel 4a, running Android 12). The violation stack traces do not contain any of our code, so they do not seem to be actionable. So I assume we could just ignore such violations. @mcomella, do you agree?


pocmo:
Just saw that this class has a test. Will try to add a test case.

pocmo:
Looking at the test code for the EdmStorageProvider, I'm actually wondering if we should get rid of the check if we take this patch here, since that trace doesn't include any Mozilla code either - so would be ignored anyways.

mcomella:
> The violation stack traces do not contain any of our code, so they do not seem to be actionable. So I assume we could just ignore such violations. @mcomella, do you agree?

If they're done by the OEM, I agree that they're not actionable. However, it's also possible for these violations-without-our-traces to be caused by our application too. I can think of these cases:
1. A third party dependency is registered into framework callbacks (e.g. ActivityLifecycleCallbacks) and only executes their own code there (e.g. Adjust currently does this and I think Leanplum actually triggered violations back in the day)
2. We call an Android framework method or third party library that schedules a callback on the main thread that only executes its own code and perhaps we could have changed the thread it executes on to avoid the violation (e.g. maybe there's a `DataStore.scheduleDatabaseRead()` method that will post itself to the current thread and execute IO)

In some cases, **StrictMode is the only practical way to catch these** – if it occurs intermittently for us, it may be difficult to catch in a profiler or devs might ignore any hangs as a device doing something in the background. In the worst case, an intermittent issue could be much worse for users than we observe locally.

That being said, **I don't know how frequently we catch these types of violations when they're impactful to users.** Maintaining StrictMode also seems to have non-trivial cost to dev time due to the high false-positive rate caused by violations outside our control from OEMs (and each OS update seems to add more).

So the ideal from a performance perspective is to ignore violations on a case-by-case basis. However, there is an argument to be made that StrictMode false positives are too disruptive to development time and we should blanket ignore violations outside of our package.

@pocmo What do you think? Do you think we're at a point where it makes sense to ignore violations outside of our package? I don't experience the same workflows as fenix devs so I have no context on how disruptive it is.

mcomella (approved):
lgtm, thanks!

Issue #22256, opened 2021-11-02
Intermittent UI test failure - < SmokeTest.editCustomSearchEngineTest >


sv-ohorvath:
### Firebase Test Run:
https://console.firebase.google.com/project/moz-fenix/testlab/histories/bh.66b7091e15d53d45/matrices/9223008020099075063/executions/bs.38fc48eea92dfdb7/testcases/1/test-cases
### Stacktrace:
`java.lang.AssertionError
at org.junit.Assert.fail(Assert.java:87)
at org.junit.Assert.assertTrue(Assert.java:42)
at org.junit.Assert.assertTrue(Assert.java:53)
at org.mozilla.fenix.ui.robots.SettingsSubMenuSearchRobot.saveEditSearchEngine(SettingsSubMenuSearchRobot.kt:140)
at org.mozilla.fenix.ui.SmokeTest$editCustomSearchEngineTest$4.invoke(SmokeTest.kt:571)
at org.mozilla.fenix.ui.SmokeTest$editCustomSearchEngineTest$4.invoke(SmokeTest.kt:563)
at org.mozilla.fenix.ui.robots.SettingsRobot$Transition.openSearchSubMenu(SettingsRobot.kt:136)
at org.mozilla.fenix.ui.SmokeTest.editCustomSearchEngineTest(SmokeTest.kt:563)`
### Build: 
Debug 11/2

Started with https://github.com/mozilla-mobile/fenix/pull/22244

**Blocked** by https://github.com/mozilla-mobile/fenix/issues/23536

csadilek:
Happened again in: https://github.com/mozilla-mobile/fenix/pull/22296

sv-ohorvath:
The problem is with the snackbar - even if it's visible the test doesn't recognize it. Same thing happens in https://github.com/mozilla-mobile/fenix/issues/17484#issuecomment-948539468 or in the context menu tests.

sv-ohorvath:
Started failing 2/2 tries on 11/4:
[logs](https://console.firebase.google.com/project/moz-fenix/testlab/histories/bh.66b7091e15d53d45/matrices/9215873714164574013/executions/bs.e2d03157593b7b2e/test-cases)
[logs](https://console.firebase.google.com/project/moz-fenix/testlab/histories/bh.66b7091e15d53d45/matrices/6249743883900558438/executions/bs.86f5161bc9e66f50/test-cases)

AndiAJ:
Re-enabled in #22339

[...]

PR #22296, opened 2021-11-03
Closes #21659: Add SERPs to history search groups

### LINKED ISSUE #21659: Consider exposing original search results page as part of the search group
Currently, we do capture the search results page on the tabs tray group, but not on the Home view or History view.

In the History view, it makes sense to have the results page appear within the group, as otherwise it appears _outside_ of the group as a single visit, which is a bit odd.

Within the Home view, it's less clear. Some kind of "resume this search" functionality (perhaps, available via the context menu?) could be more appropriate. Or, we could display the results page within the group itself, but with a title clearly indicating what it is.

csadilek:
This adds SERPs to history search groups (in addition to the result pages opened from SERPs). The diff looks worse because it contains cleanup of our comments to make the cases clearer :).


grigoryk (approved):
We paired up on this, worked through the various additional cases, added/fixed tests and did manual testing. It's better than before! And gets us the intended result of having SERP pages included in groups.

csadilek:
This keeps running into https://github.com/mozilla-mobile/fenix/issues/22256. One more try but then we should set that UI test to ignore until we have a fix.

Issue #22147, opened 2021-10-25
MR2 telemetry: Search term groups in history


eliserichards:
## What questions will you answer with this data?

Implement the following telemetry probes: (might be an incomplete list. check with product)
- A user opens a search term group in history
- A user opens a tab from the search term group
- A user closes all tabs in the search term group
- A user closes a single tab in the search term group

## Acceptance Criteria
- [ ] ENG files a [DS JIRA](https://jira.mozilla.com/projects/DO/issues/DO-228?filter=allopenissues) request outlining their methodology.
- [ ] DS sign off on instrumentation methodology addressing product questions.
- [ ] Event pings can be queried via re:dash
- [ ] Event pings can be queried via amplitude
- [ ] We are sending telemetry events for the actions listed in the requirements
- [ ] We have documented the telemetry
- [ ] We have asked a data steward to [review](https://github.com/mozilla/data-review/blob/master/request.md) the telemetry

Mugurell:
Seems like we already have a probe knowing when a search group is opened - https://dictionary.telemetry.mozilla.org/apps/fenix/metrics/history_search_term_group_tapped
Asked Kimmy to confirm the other three.

AndiAJ:
Hi, **verified as fixed** on the latest Nightly build 96.0a1 from 11/19 using a Google Pixel 3a (Android 11)

✔️  A user opens a search term group in history (Previously done and checked in #22172) - Event Ping ac6c92f5-7ef7-4381-939e-3de638ab512c
```
{
          "timestamp": 4645,
          "category": "history",
          "name": "search_term_group_tapped"
        }
```
✔️  A user opens a tab from the search term group - Event Ping 7178208a-5338-4f91-94de-fbee9310ac64
```
{
          "timestamp": 1244,
          "category": "history",
          "name": "search_term_group_open_tab"
        }
```
✔️  A user closes all tabs in the search term group - Event Ping 7178208a-5338-4f91-94de-fbee9310ac64
```
{
          "timestamp": 14669,
          "category": "history",
          "name": "search_term_group_remove_all"
        }
```
✔️  A user closes a single tab in the search term group - Event Ping 7178208a-5338-4f91-94de-fbee9310ac64
```
{
          "timestamp": 13194,
          "category": "history",
          "name": "search_term_group_remove_tab"
        } 
```

[Logcat](https://github.com/mozilla-mobile/fenix/files/7568725/srcgrp.txt)
[...]

Issue #22314, opened 2021-11-04
[Bug]: Tapping a search group item should switch to the previously opened website


AndiAJ:
### Steps to reproduce

1. Search for a term 
2. Open one of the results
3. Exit to the home screen
4. Tap the search group
5. Tap the item
6. Check tab counter
7. Open tabs tray
8. Check open tabs

### Expected behaviour

If the search group item is the same as the tab opened previously it should switch automatically to already existing tab

### Actual behaviour

The search group item is opened in a new tab

### Device name
OnePlus A3 Android 6.0.1

### Firefox version
Firefox 95.0.0 Beta 1
Firefox Nightly 96 from 11/04

### Additional information

► Video

https://user-images.githubusercontent.com/51314259/140317395-9a4e8d2e-9fc0-49f8-bf6b-50f59660cf0b.mp4

SoftVision-LorandJanos:
Verified as fixed on the latest Nightly 98.0a1 (2022-01-26).

Device used for testing:
- Oppo Reno 6 (Android 11).
- Google Pixel 4 (Android 12).

Closing the ticket as fixed.

PR #22385, opened 2021-11-09
For #22147 - Reuse existing tab for search group items

Mugurell:
https://user-images.githubusercontent.com/11428869/140966464-050e6aad-b7eb-4cbc-8951-14a4e8ddc48c.mp4

### Pull Request checklist

1. click on Show All Checks,


gabrielluong:
I am not fully sure about the product requirements/expectations here. So, gonna forward this over to @csadilek to make a call. I think since this PR has nothing to do with telemetry #22147 we should also separate it out into its own issue.

csadilek:
@Mugurell @gabrielluong This doesn't seem right to me 🤔 . Why should opening a grouped item behave differently that a regular history item? It seems like a miscommunication, maybe there's a different ticket?

I can see us wanting to switch to an existing tab with the same url if it already exists using our `SelectOrAddUseCase`, but creating a new tab in one case, and overriding the tab in another doesn't seem right to me? What am I missing?

csadilek:
@Mugurell Yeah so I assume this relates to FNXV2-18563? But there I understand it should be like `SelectOrAddUseCase`, where if there's already a tab with that same URL we just switch to it, but we shouldn't navigate the current tab away.

"If the search group item is the same as the tab opened previously it should switch automatically to already existing tab"...that's what I understood.

Mugurell:
Sorry for the mishap, indeed this patch is for https://github.com/mozilla-mobile/fenix/issues/22314

Mugurell:
> @Mugurell Yeah so I assume this relates to FNXV2-18563? But there I understand it should be like `SelectOrAddUseCase`, where if there's already a tab with that same URL we just switch to it, but we shouldn't navigate the current tab away.
> 
> "If the search group item is the same as the tab opened previously it should switch automatically to already existing tab"...that's what I understood.

Thank you! `SelectOrAddUseCase` seems like the needed API.

rocketsroger (approved):

Issue #22271, opened 2021-11-02
Try to use getPrimaryClipDescription on Android 12 


Amejia481:
On Android 12 there is a new behavior change when accessing the clipboard, the OS shows a toast indicating that the app is is accessing it, more details can be found [here](https://developer.android.com/about/versions/12/behavior-changes-all#clipboard-access-notifications).

When users try to search or open an URL in the toolbar Fenix tries to check on the clipboard to see if there is an URL on the clipboard that the user wants to navigates to, we use these [extension functions](https://github.com/mozilla-mobile/fenix/blob/main/app/src/main/java/org/mozilla/fenix/utils/ClipboardHandler.kt#L28-L29) to try to figure out, but they are triggering [the Clipboard access notifications on Android 12](https://developer.android.com/about/versions/12/behavior-changes-all#clipboard-access-notifications), to avoid that we want to use  [getPrimaryClipDescription()](https://developer.android.com/reference/android/content/ClipboardManager#getPrimaryClipDescription())   

### Steps to show the reproduce [the Clipboard access notifications on Android 12](https://developer.android.com/about/versions/12/behavior-changes-all#clipboard-access-notifications):

1.  Copy a URL to clipboard in some other app (e.g. https://example.org )  
2.  Start Firefox Nightly

dholbert:
One particularly-important use case to consider here:  under some configurations[1] we never show the "Fill link from clipboard" UI, but nonetheless we seem to check the clipboard as if we were going to show the UI.  This ends up meaning that the user sees a "Firefox Nightly pasted from your clipboard" system notification **when there's no clear user-visible indication about why/where Firefox pasted the clipboard contents (or what Firefox found there)**.  This ends up feeling unsettling/concerning from a user perspective.

[1] In particular: on my phone, the "Settings | Search | Show search engines: [enabled]" setting causes the "fill link from clipboard" UI to never appear. (but Firefox still queries my clipboard for apparently no reason)

nathan-castlehow:
Hi,
I'm happy to give this a stab if that will help at all. New to this project so forgive me If I've missed any steps here

I had an initial look into this.
The extension functions listed are already using getPrimaryDescription before then reading the clipboard.

` private fun ClipboardManager.isPrimaryClipPlainText() =
        primaryClipDescription?.hasMimeType(MIME_TYPE_TEXT_PLAIN) ?: false`

   ` private fun ClipboardManager.isPrimaryClipHtmlText() =
[...]

PR #22391, opened 2021-11-09
For #22271 Improve URL accessing from the clipboard for Android 12 and above.

Amejia481:
As accessing the clipboard is sensitive action, on Android 12 and above we changed the UI to not show (access the clipboard) the URL unless the user request to do so.

### Behavior on Android 12 and above

We show the **"Fill link from clipboard"** item (but not the URL item)  when the user clicks it, we fill the toolbar with its content.

Demo

### Behavior pre Android 12 and above

Same as before show the **"Fill link from clipboard"** item and the URL item, when the user clicks it, we navigate to the site.

clipboard_pre_android_12

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/utils/ClipboardHandler.kt
+++ b/app/src/main/java/org/mozilla/fenix/utils/ClipboardHandler.kt
@@ -7,6 +7,8 @@ package org.mozilla.fenix.utils
 import android.content.ClipData
 import android.content.ClipboardManager
 import android.content.Context
+import android.os.Build
+import android.view.textclassifier.TextClassifier
 import androidx.annotation.VisibleForTesting
 import androidx.core.content.getSystemService
 import mozilla.components.support.utils.SafeUrl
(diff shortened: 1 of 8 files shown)


csadilek (commented):
👍 I think this is a good immediate solution without losing any functionality, and we can improve later. I noticed two things though

- After filling in from clipboard the cursor is at the beginning instead of at the end. After pasting text the user expectation is for the cursor to be at the end of the text.
- The keyboard pops up and down when filling in.

app/src/main/java/org/mozilla/fenix/utils/ClipboardHandler.kt
```diff
@@ -64,7 +64,13 @@ class ClipboardHandler(val context: Context) {
...
             val description = clipboard.primaryClipDescription
-            val score = description?.getConfidenceScore(TextClassifier.TYPE_URL) ?: 0F
+            // An IllegalStateException is thrown if the url is too long.
+            val score =
+                try {
+                    description?.getConfidenceScore(TextClassifier.TYPE_URL) ?: 0F
+                } catch (e: IllegalStateException) {
+                    0F
+                }
             score >= 0.7F
```
Amejia481: @csadilek  I added the `try ` as I noticed that `getConfidenceScore` was throwing an `IllegalStateException` when the URL was too long.

Amejia481: > 👍 I think this is a good immediate solution without losing any functionality, and we can improve later. I noticed two things though
> 
>     * After filling in from clipboard the cursor is at the beginning instead of at the end. After pasting text the user expectation is for the cursor to be at the end of the text.
> 
>     * The keyboard pops up and down when filling in.

Added [this](https://github.com/mozilla-mobile/fenix/pull/22391/commits/27c153937b93fa0a059ff73b26adcc5b275925cc#) separated commit to address the listed issues

csadilek (approved):
👍

app/src/main/java/org/mozilla/fenix/utils/ClipboardHandler.kt
csadilek: Tested again, works great now!

Amejia481:
We will wait until early next week to land, after the long weekend pass.

Amejia481:
@Mergifyio backport releases_v95.0.0

PR #22406, opened 2021-11-10
Close #22402: Add top placeholder for home

### LINKED ISSUE #22402: [Bug] Issue with scroll position on homepage after cold start
- Open Firefox Beta from a cold start without any top sites
- Expecting to see the top items on homescreen but instead I land on Pocket articles. See screenshot.

Note: this does not happen when I tap on the hoomescreen button from a tab, only after cold start.

 
index

rocketsroger:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
@@ -377,8 +377,7 @@ class HomeFragment : Fragment() {
             homeFragmentStore,
             binding.sessionControlRecyclerView,
             viewLifecycleOwner,
-            sessionControlInteractor,
-            homeViewModel
+            sessionControlInteractor
         )
 
         updateSessionControlView()
@@ -543,8 +542,6 @@ class HomeFragment : Fragment() {
         if (bundleArgs.getBoolean(FOCUS_ON_ADDRESS_BAR)) {
             navigateToSearch()
         } else if (bundleArgs.getLong(FOCUS_ON_COLLECTION, -1) >= 0) {
-            // No need to scroll to async'd loaded TopSites if we want to scroll to collections.
-            homeViewModel.shouldScrollToTopSites = false
             /* Triggered when the user has added a tab to a collection and has tapped
             * the View action on the [TabsTrayDialogFragment] snackbar.*/
             scrollAndAnimateCollection(bundleArgs.getLong(FOCUS_ON_COLLECTION, -1))
--- a/app/src/main/java/org/mozilla/fenix/home/TopPlaceholderViewHolder.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/TopPlaceholderViewHolder.kt
@@ -1,0 +1,24 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.home
+
+import android.view.View
+import org.mozilla.fenix.R
+import org.mozilla.fenix.utils.view.ViewHolder
+
+/**
+ * View holder for a synchronous, unconditional and invisible placeholder.  This is to anchor home to
+ * the top when home is created.
+ */
+class TopPlaceholderViewHolder(
+    view: View
+) : ViewHolder(view) {
+
+    fun bind() = Unit
+
+    companion object {
+        const val LAYOUT_ID = R.layout.top_placeholder_item
+    }
+}
--- a/app/src/main/res/layout/top_placeholder_item.xml
+++ b/app/src/main/res/layout/top_placeholder_item.xml
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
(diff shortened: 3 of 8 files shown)


app/src/main/res/layout/top_placeholder_item.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
MozillaNoah, line 5: This could optionally just be a <View> with a self-closing "/>" (it could be more performant than a ConstraintLayout too)

`<View xmlns:android="http://schemas.android.com/apk/res/android"
    android:layout_height="0dp"/>
`

rocketsroger: Good idea.  Not sure why I stayed with contraint layout.  I'll try it.  Thanks!

rocketsroger: Works well.  Updated.  Thanks!

app/src/main/java/org/mozilla/fenix/home/TopPlaceholderViewHolder.kt
```diff
@@ -0,0 +1,20 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.home
+
+import android.view.View
+import org.mozilla.fenix.R
+import org.mozilla.fenix.utils.view.ViewHolder
+
+class TopPlaceholderViewHolder(
```
mcomella, line 11: nitpick: I don't understand how this PR addresses the issues from #22402. Could you add a comment explaining what the issue is and how this code addresses it? It doesn't necessarily have to be here, as a class kdoc, but it seemed like an okay place

Sorry to drive by 😅

rocketsroger: > nitpick: I don't understand how this PR addresses the issues from #22402. Could you add a comment explaining what the issue is and how this code addresses it? It doesn't necessarily have to be here, as a class kdoc, but it seemed like an okay place
> 
> Sorry to drive by 😅

Sure, no worries.  Because we don't know what is being inflated first, we end up having an issue where the first inflated element is the one that is shown in the recyclerview.  

Take a look at `SessionControlView` where I removed the check `homeScreenViewModel.shouldScrollToTopSites`.  This code is how we currently scroll to the top.  The part I removed checks if top sites is inflated.  If it is, then it scrolls to the top (index 0).  However, this is assuming that top sites exists.  The bug is when the user removed all top sites then it won't scroll.

After some investigation, the most simple solution (and most performant) is to just synchronously inflate a top marker.  This way it's the first one that is inflated and it is also the top one.  Recyclerview then keep it in view no matter how many other items are added below it.  Let me know if you have any other questions.  I would suggest you try the current build without top sites, set default to home and cold start Fenix.  You'll see the problem this patch fixes.

csadilek: @rocketsroger Your comment here is good: https://github.com/mozilla-mobile/fenix/pull/22406/files#diff-f6968644d19314ed1385ed1bef14d8a3aaa163999143e11b0449b94292e6be56R48 -> Can we add something similar to the `TopPlaceholderViewHolder` to explain why it exists :)

rocketsroger: Sure.  I'll add the comments in `TopPlaceholderViewHoder`.  Thanks

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -543,8 +542,6 @@ class HomeFragment : Fragment() {
         if (bundleArgs.getBoolean(FOCUS_ON_ADDRESS_BAR)) {
             navigateToSearch()
         } else if (bundleArgs.getLong(FOCUS_ON_COLLECTION, -1) >= 0) {
-            // No need to scroll to async'd loaded TopSites if we want to scroll to collections.
-            homeViewModel.shouldScrollToTopSites = false
             /* Triggered when the user has added a tab to a collection and has tapped
             * the View action on the [TabsTrayDialogFragment] snackbar.*/
             scrollAndAnimateCollection(bundleArgs.getLong(FOCUS_ON_COLLECTION, -1))
```
csadilek, line 547: I didn't know we had this scroll to collections functionality. It actually doesn't work :(. Unrelated to the change here though, doesn't seem to work in release for me either. Can you help me verify and then we can file an issue?

- Open tabs tray
- Select a few tabs and add to collection
- Tap on view collection

Expected: Home is opened and collection is visible
Actual: Home is opened and I am somewhere on jump back in or recent searches

If you can reproduce, let's file, as it predates the changes here.

csadilek (approved):
Verified that this fixed the problem and the implementation is simpler than before. I found one follow-up to investigate and file (see below.)

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
rocketsroger: Confirmed that this is reproducible and not related to this change.  https://github.com/mozilla-mobile/fenix/issues/22462 created to track this.

Issue #22155, opened 2021-10-26
MR2 telemetry: inactive tabs


Mugurell:
Tracking https://mozilla-hub.atlassian.net/browse/FNXV2-18425

--- 
We currently have `tabs_tray.has_inactive_tabs`, which is an Event with an extra for the count (`inactive_tabs_count`), but we want to split this out into two separate metrics. `has_inactive_tabs` does not need to be removed yet, that will happen in another issue.

Probes to implement: 

    Metric - boolean: Is the inactive tabs section showing?

    Metric - counter: How many inactive tabs does the user have?
    
-------

After talking with Kimmy we'll only go with one new probe: `inactive_tabs_count` sent in the metrics ping.
Whether the inactive tabs feature is enabled or not is already tracked by https://dictionary.telemetry.mozilla.org/apps/fenix/metrics/preferences_inactive_tabs_enabled

Mugurell:
Removing QA since the current implementation based on [Jon's comment](https://github.com/mozilla-mobile/fenix/pull/22163#discussion_r742482956).

Mugurell:
This should be ready to test.

AndiAJ:
Hi @Mugurell , I've just checked this matter on the latest Nightly 96.0a1 from 11/11 using a Google Pixel 3a (Android 11)

It looks like the inactive tabs counter from the metrics ping remains set to 0.

❌  3 inactive tabs - Metrics Ping 95c8ef82-c3e7-43f6-8844-6edcaa89e8b7
`"metrics.inactive_tabs_count": 0`

![image](https://user-images.githubusercontent.com/51314259/141290333-fb107292-67e9-405f-b4a3-d1a993482a64.png)

[count.txt](https://github.com/mozilla-mobile/fenix/files/7520014/count.txt)
[Glean dashboard](https://debug-ping-preview.firebaseapp.com/pings/AJ1111004)

I'll re move the QA needed label until further notice.

Mugurell:
Thank you Andi!
There was indeed an issue with when we were trying to read the number of inactive tabs. 
We'll now set this when the user opens tabs tray, which should happen more times in the day and so there should definitely be data for the `metrics` ping.

AndiAJ:
Hi, **verified as fixed** on the latest Nightly build 96.0a1 from 11/19 using a Google Pixel 3a (Android 11)

✔️  3 inactive tabs - Metrics ping c70521b1-6833-498c-8e71-db69795e51cc
[...]

PR #22421, opened 2021-11-11
For #22155 - Store inactive tabs count telemetry when user opens tabs tray

Mugurell:
Setting this value in FenixApplication.onCreate was buggy because of a race
with restoring BrowserState.
Setting it here would ensure a better granularity of the events and so to more
accurate reporting.

Checking now the metrics ping shows the expected value - [Glean dashboard](https://debug-ping-preview.firebaseapp.com/pings/petruInactiveTabs7)

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/metrics.yaml
+++ b/app/metrics.yaml
@@ -1255,9 +1255,9 @@ metrics:
     expires: "2022-02-01"
   inactive_tabs_count:
     type: quantity
-    lifetime: application
     description: |
-      How many inactive tabs does the user have.
+      How many inactive tabs does the user have, checked when the user opens
+      the tabs tray.
       Value will be 0 if the feature is disabled.
     send_in_pings:
       - metrics
--- a/app/src/main/java/org/mozilla/fenix/tabstray/browser/TabSorter.kt
+++ b/app/src/main/java/org/mozilla/fenix/tabstray/browser/TabSorter.kt
@@ -57,6 +57,8 @@ class TabSorter(
         if (shouldReportMetrics) {
             shouldReportMetrics = false
 
+            metrics.track(Event.InactiveTabsCountUpdate(inactiveTabs.size))
+
             if (settings.inactiveTabsAreEnabled) {
                 metrics.track(Event.TabsTrayHasInactiveTabs(inactiveTabs.size))
             }
--- a/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
+++ b/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
@@ -80,7 +80,6 @@ import org.mozilla.fenix.components.Core
 import org.mozilla.fenix.components.metrics.Event
 import org.mozilla.fenix.components.metrics.MozillaProductDetector
 import org.mozilla.fenix.components.toolbar.ToolbarPosition
-import org.mozilla.fenix.ext.actualInactiveTabs
 import org.mozilla.fenix.perf.MarkersActivityLifecycleCallbacks
 import org.mozilla.fenix.utils.Settings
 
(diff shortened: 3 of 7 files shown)


app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -363,6 +364,8 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
         settings().shouldReturnToBrowser =
             components.core.store.state.getNormalOrPrivateTabs(private = false).isNotEmpty()
 
+        Metrics.inactiveTabsCount.set(components.core.store.state.actualInactiveTabs(settings()).size.toLong())
```
Mugurell, line 367: Added the probe here looking for the best compromise between 
"not to soon" - FenixApplication#onCreate + "don't impact performance" and 
"not too low so that this will be called more often than needed"
and because I saw the `settings()...` calls just below this which store the right values in SharedPreferences to make them available for the metrics ping. (though about using a similar mechanism to store this value until it's needed for the ping but it seemed superfluous if Glean already does this for probes with `lifetime: Ping`)
Open to suggestions to move this elsewhere.

pocmo: In metrics.yml it just says "How many inactive tabs does the user have." - but that value changes over time. Measuring here vs. measuring it on app start may result in completely different values? 🤔 Unfortunately the spreadsheet also doesn't mention *when* this is getting measured. Does this need to be clarified?

Just wanted to mention, if you need it on app start and want to avoid the race. There's a `restoreComplete` boolean in the state, that you can wait for. I think we use this already to go back to home if after restore there are no tabs anymore (e.g. we purged inactive ones).

Mugurell: Ohh, the `restoreComplete` seems nice, thank you!

I initially added this in FenixApplication#setStartupMetrics since I saw there similar telemetry for the number of open tabs , topSites and thoughout that since we only send the last count of inactive tab once per day to Glean this place is a good fit.
Thinking now about the granularity of inactive tabs updates maybe a more appropriate place would be the tabs tray - from where the event that also contains this count as extra key is logged now. This would cover the scenario in which the user closes some inactive tab, then the app and will only open it again when the `metrics` ping is scheduled for upload thus potentially loosing the update from the previous day (not entirely lost since it could be reported in the next ping).

Will move this to tabs tray, update documentation and check with Kimmy.

app/metrics.yaml
```diff
@@ -1255,9 +1255,9 @@ metrics:
     expires: "2022-02-01"
   inactive_tabs_count:
     type: quantity
-    lifetime: application
     description: |
-      How many inactive tabs does the user have.
+      How many inactive tabs does the user have, checked when the user opens
+      the tabs tray.
```
jonalmeida, line 1260: Thanks for updating the description to reflect the behaviour accurately!

jonalmeida (approved):

jonalmeida:
> Dismissing my review because I didn't read the discussion above first; which means this PR will change.

And it already did.. but my brain and motor functions are working on separate dispatchers which are racing against each other. 🙃

Mugurell:
Kimmy said she's okay with logging the data when the tray is opened.
So I think this is ready for re-review.

app/src/main/java/org/mozilla/fenix/tabstray/browser/TabSorter.kt
```diff
@@ -57,6 +58,8 @@ class TabSorter(
         if (shouldReportMetrics) {
             shouldReportMetrics = false
 
+            Metrics.inactiveTabsCount.set(inactiveTabs.size.toLong())
```
rocketsroger, line 61: I know this is not your fault but is it possible for us to go through `GleanMetricsService` when calling `Metrics`?  I think this is the only case that call directly through `Metrics`.

rocketsroger (approved):
Looks good! 🚢

Issue #22310, opened 2021-11-04
[Bug]:Collections - Tapping on "Deselect all" doesn't change the state for tabs previously selected


sflorean:
### Steps to reproduce

1. Add a collection.
2. Go to homescreen and tap on the collection.
3. Open the 3dotmenu and select "Add tab".
4. Select 2-3 tabs.
5. Tap on "Select All".
6. Tap on "Deselect All".

### Expected behaviour

The tabs are not selected, the checkboxes are clear.

### Actual behaviour

The tabs selected at step 4 are still selected. 

### Device name

Huawei MediaPad M2 , Pixel

### Android version

Android 5.1.1, 12

### Firefox release type

Firefox Beta

### Firefox version

95.0.0-beta.1

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-18559)

Robert-Aron293:
Can I work on this?

sflorean:
Verified as fixed on Nightly 5/24 with Nokia 6 (Android 7.1.1).

PR #22437, opened 2021-11-14
For #22310 - Change tab holder's checkbox state when pressing "Select…

Robert-Aron293:
Change tab holder's checkbox state when pressing "Select All"/"Deselect All".

### Pressed "Select All"
Screenshot_20211114-163817_Firefox Preview
### Pressed "Deselect All"
Screenshot_20211114-163824_Firefox Preview

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/collections/CollectionCreationTabListAdapter.kt
+++ b/app/src/main/java/org/mozilla/fenix/collections/CollectionCreationTabListAdapter.kt
@@ -43,10 +43,8 @@ class CollectionCreationTabListAdapter(
             when (payloads[0]) {
                 is CheckChanged -> {
                     val checkChanged = payloads[0] as CheckChanged
-                    if (checkChanged.shouldBeChecked) {
-                        binding.tabSelectedCheckbox.isChecked = true
-                    } else if (checkChanged.shouldBeUnchecked) {
-                        binding.tabSelectedCheckbox.isChecked = false
+                    if (checkChanged.shouldBeChecked || checkChanged.shouldBeUnchecked) {
+                        holder.changeCheckBoxState(checkChanged.shouldBeChecked)
                     }
                     binding.tabSelectedCheckbox.isGone = checkChanged.shouldHideCheckBox
                 }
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/fenix/collections/CollectionCreationTabListAdapter.kt
```diff
@@ -45,9 +45,9 @@ class CollectionCreationTabListAdapter(
                 is CheckChanged -> {
                     val checkChanged = payloads[0] as CheckChanged
                     if (checkChanged.shouldBeChecked) {
```
Alexandru2909, line 47: Hi, this is just a drive-by comment: 
If you decided to drop true/false and use checkChanged.shouldBeChecked to call holder::changeCheckBoxState, why not also change the if condition?
I was thinking of something like checkChanged.souldBeChecked || checkChanged.shouldBeUnchecked, it is also easier to read.

Robert-Aron293: You are right. Thanks

app/src/main/java/org/mozilla/fenix/collections/CollectionCreationTabListAdapter.kt
```diff
@@ -44,10 +44,8 @@ class CollectionCreationTabListAdapter(
             when (payloads[0]) {
                 is CheckChanged -> {
                     val checkChanged = payloads[0] as CheckChanged
-                    if (checkChanged.shouldBeChecked) {
-                        binding.tabSelectedCheckbox.isChecked = true
-                    } else if (checkChanged.shouldBeUnchecked) {
-                        binding.tabSelectedCheckbox.isChecked = false
+                    if (checkChanged.shouldBeChecked || checkChanged.shouldBeUnchecked) {
+                        holder.changeCheckBoxState(checkChanged.shouldBeChecked)
```
Mugurell, line 48: This check assumes that whenever `shouldBeUnchecked` is true `shouldBeChecked` is false.
Which sounds right as per the naming but the current if/else does not make this assumption and makes me wonder why we had this abstraction in the first place and whether we can just drop the `shouldBeUnchecked` property or it's an important edgecase we need to still acount for.

Alexandru2909: @Robert-Aron293 to minimize the risk of regressions let's keep the previous if/else approach for now and I'll investigate how `CheckChanged` is used

Alexandru2909 (approved):

Alexandru2909:
Changes will be landed in #25315

Alexandru2909:
Closed, as changes were landed in #25315.

PR #22465, opened 2021-11-16
RELENG-489 - Beetmover

hneiva:
Jira ticket: https://mozilla-hub.atlassian.net/browse/RELENG-489

Bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1614763

Adds beetmover functionality to Fenix.
Beetmover uploads binaries (apks in this case) to archive.mozilla.org where we keep all versions of Firefox.

### DIFF
--- a/taskcluster/fenix_taskgraph/transforms/multi_dep.py
+++ b/taskcluster/fenix_taskgraph/transforms/multi_dep.py
@@ -11,6 +11,8 @@ from taskgraph.transforms.base import TransformSequence
 from taskgraph.util.schema import resolve_keyed_by
 from taskgraph.util.treeherder import inherit_treeherder_from_dep, join_symbol
 
+from fenix_taskgraph.util.scriptworker import generate_beetmover_upstream_artifacts
+
 
 transforms = TransformSequence()
 
@@ -24,7 +26,7 @@ def build_name_and_attributes(config, tasks):
         }
         primary_dep = task["primary-dependency"]
         copy_of_attributes = primary_dep.attributes.copy()
-        task.setdefault("attributes", copy_of_attributes)
+        task.setdefault("attributes", {}).update(copy_of_attributes)
         # run_on_tasks_for is set as an attribute later in the pipeline
         task.setdefault("run-on-tasks-for", copy_of_attributes['run_on_tasks_for'])
         task["name"] = _get_dependent_job_name_without_its_kind(primary_dep)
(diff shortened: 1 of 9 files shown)


hneiva:
@st3fan we have a few options for landing this feature and would like your input:
1. Set "dep" credentials in the prod worker, then update credentials once we verify the results are good.
2. Still target the dev worker (hard coded [here](https://github.com/mozilla-mobile/fenix/pull/22465/files#diff-78cef187ec6d54d1013f6c8b16706396955b3c4be208320c0fdf815418cce528R58))
3. Upload to prod and adjust as needed

hneiva:
Jira ticket: https://mozilla-hub.atlassian.net/browse/RELENG-489

ahal (approved):

taskcluster/fenix_taskgraph/transforms/multi_dep.py
```diff
@@ -24,7 +26,7 @@ def build_name_and_attributes(config, tasks):
         }
         primary_dep = task["primary-dependency"]
         copy_of_attributes = primary_dep.attributes.copy()
-        task.setdefault("attributes", copy_of_attributes)
+        task.setdefault("attributes", {}).update(copy_of_attributes)
```
jcristau, line 29: It seems surprising that we'd overwrite any attributes explicitly set on the task, no?

taskcluster/fenix_taskgraph/transforms/multi_dep.py
```diff
@@ -65,19 +67,27 @@ def build_upstream_artifacts(config, tasks):
...
-            ])
-            if paths:
-                worker_definition["upstream-artifacts"].append({
-                    "taskId": {"task-reference": "<{}>".format(dep.kind)},
-                    "taskType": dep.kind,
-                    "paths": paths,
-                })
-
-        task["worker"].update(worker_definition)
+        if "artifact_map" in task["attributes"]:
```
jcristau, line 70: Would it maybe be easier to use a separate transform for the beetmover task, instead of repurposing this one?

jcristau (commented):
A couple of drive-by questions...

taskcluster/fenix_taskgraph/transforms/multi_dep.py
ahal: Yeah that could make sense, though technically `beetmover` isn't the only kind that can use "declarative artifacts".. Maybe the util function should be renamed to `generate_declaritive_upstream_artifacts` or something.

PR #22466, opened 2021-11-16
Closes #22321: Upgrade protobuf version.

### LINKED ISSUE #22321: Cannot find appropriate protobuf compiler when building Fenix on M1 Apple
```
Execution failed for task ':app:generateDebugProto'.
> Could not resolve all files for configuration ':app:protobufToolsLocator_protoc'.
   > Could not find protoc-3.11.4-osx-aarch_64.exe (com.google.protobuf:protoc:3.11.4).
     Searched in the following locations:
         https://repo.maven.apache.org/maven2/com/google/protobuf/protoc/3.11.4/protoc-3.11.4-osx-aarch_64.exePossible solution:
 - Declare repository providing the artifact, see the documentation at https://docs.gradle.org/current/userguide/declaring_repositories.html
```

This is a tracking issue for https://github.com/grpc/grpc-java/issues/7690

Workaround: https://github.com/grpc/grpc-java/issues/7690#issuecomment-747196215

MatthewTighe:
### Pull Request checklist
- [x] **Tests**: This PR does not include tests because it only upgrades a dependency version. It relies on existing regressions to prove compatibility. 
- [x] **Screenshots**: None included because none necessary for this dependency.
- [x] **Accessibility**: No user-facing changes.

1. click on Show All Checks,

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -605,8 +605,14 @@ dependencies {
 }
 
 protobuf {
+    // Mac M1 workaround until we can bump the version. Dependent on A-S.
+    // See https://github.com/mozilla-mobile/fenix/issues/22321
     protoc {
-        artifact = Deps.protobuf_compiler
+        if (osdetector.os == "osx") {
+            artifact = "${Deps.protobuf_compiler}:osx-x86_64"
+        } else {
+            artifact = Deps.protobuf_compiler
+        }
     }
 
     // Generates the java Protobuf-lite code for the Protobufs in this project. See


buildSrc/src/main/java/Dependencies.kt
```diff
@@ -54,7 +54,7 @@ object Versions {
 
     const val google_play_store_version = "1.8.0"
 
-    const val protobuf = "3.11.4" // keep in sync with the version used in AS.
+    const val protobuf = "3.19.1" // keep in sync with the version used in AS.
```
jonalmeida, line 57: Ah, this comment is suspect. I think "AS" stands for App Services (our native components that are used to share code between desktop/iOS/Android) and not Android Studio, so we probably want to have this version in sync: https://github.com/mozilla/application-services/blob/main/build.gradle#L26

@Mugurell can probably confirm if we need to put a PR for app services too.

jonalmeida: I'm guessing this built locally and worked for you so I wonder what happens when these are longer in sync? We might be able to tell when the UI tests run in this build.

MatthewTighe: Oh, that makes way more sense. I'd been deep in a search trying to figure out what version Studio was using. Happy to put up a PR for app services as well if that's the best path forward. Need to get my tooling together for it still, anyway

jonalmeida: > 80 test cases failed, 100 passed, 8 flaky

Sounds like we need to upgrade the a-s version too, which may be more complicated at that layer. We can start by filing an with our intent.

```
java.lang.ClassCastException: java.lang.Class cannot be cast to java.lang.String
	at com.google.protobuf.MessageSchema.newSchemaForRawMessageInfo(MessageSchema.java:507)
	at com.google.protobuf.MessageSchema.newSchema(MessageSchema.java:227)
	at com.google.protobuf.ManifestSchemaFactory.newSchema(ManifestSchemaFactory.java:77)
	at com.google.protobuf.ManifestSchemaFactory.createSchema(ManifestSchemaFactory.java:71)
	at com.google.protobuf.Protobuf.schemaFor(Protobuf.java:90)
	at com.google.protobuf.Protobuf.schemaFor(Protobuf.java:104)
	at com.google.protobuf.GeneratedMessageLite.parsePartialFrom(GeneratedMessageLite.java:1537)
	at com.google.protobuf.GeneratedMessageLite.parseFrom(GeneratedMessageLite.java:1711)
	at com.google.protobuf.GeneratedMessageLite.parseFrom(GeneratedMessageLite.java:1704)
	at mozilla.appservices.places.MsgTypes$BookmarkNode.parseFrom(MsgTypes.java:3615)
	at mozilla.appservices.places.PlacesReaderConnection.getBookmarksTree(PlacesConnection.kt:498)
	at mozilla.components.browser.storage.sync.PlacesBookmarksStorage$getTree$2.invokeSuspend(PlacesBookmarksStorage.kt:38)
	at mozilla.components.browser.storage.sync.PlacesBookmarksStorage$getTree$2.invoke(Unknown Source:8)
	at mozilla.components.browser.storage.sync.PlacesBookmarksStorage$getTree$2.invoke(Unknown Source:4)
	at kotlinx.coroutines.intrinsics.UndispatchedKt.startUndispatchedOrReturn(Undispatched.kt:89)
	at kotlinx.coroutines.BuildersKt__Builders_commonKt.withContext(Builders.common.kt:165)
```

grigoryk: I filed https://github.com/mozilla/application-services/issues/4657.

There's a lot that changed in protobufs since 3.11.4, so it's possible that switchover won't be as easy on the AS side. Good news is that if that does prove to be a problem we can try the workarounds suggested in https://github.com/grpc/grpc-java/issues/7690#issuecomment-747196215 ?

Mugurell: > @Mugurell can probably confirm if we need to put a PR for app services too.

When I added this new dependency in Fenix I saw those strange errors posted in the comment below coming from AS incompatibilities with newer versions so I just used what aplication-services uses to ensure everything will continue to work the same as before.
Since the functionalities are different we could probably try forcing one version in Fenix and one in AS but I don't see any reason to go through these hurdles. And no reason to update now in Fenix if the workaround for M1 works with the current version. If I understand correctly AS plans to drop this dependency soon-ish after which just updating in Fenix should be straight-forward.

MatthewTighe: Looks like the UI tests pass using the workaround! Everything looks good locally as far as I can tell

app/build.gradle
```diff
@@ -605,8 +605,14 @@ dependencies {
 }
 
 protobuf {
+    // Mac M1 workaround until we can bump the version. Dependent on A-S.
+    // See https://github.com/mozilla-mobile/fenix/issues/22321
```
jonalmeida, line 609: Thanks for adding a link to the issue so we can know when to remove this workaround!

jonalmeida (approved):
Great, thanks for getting on this!

jonalmeida:
When you feel comfortable (now is fine), you only need to add the "needs landing" label to the PR.

In this case, the "squashed" option since I see there are three commits that all pertain to the same fix.

Mugurell:
> In this case, the "squashed" option since I see there are three commits that all pertain to the same fix.

Slightly off-topic: the bad thing about using `needs-landing-squashed` is that while this helps with combining changes into one commit and keeping a clean and clear history all of the commit messages are also combined and the result may look weird and not really useful, that's why I prefer squashing locally - to be able to control the final commit title and message, then "push -f", then add the `needs-landing` label.

MatthewTighe:
Looks like this could use a rebase anyway, so I'll follow your advice @Mugurell. Thanks all

jonalmeida:
> 

Yes, this is preferable, but a bigger ask if a contributor might be new to git or is unfamiliar. 

Several commits is fine to land as well if they are clearly distinctive parts that should be separated. (e.g. one commit is a bug fix and another is a bit of formatting that touches some other files but is loosely related). Separate commits are also nice for the review if the change is rather large, and once it's approved, you can squash it locally (or here) and then merge it.

In the end, it's about what makes the git history easier to follow with accessibility. 🙂

PR #22467, opened 2021-11-16
Update lastAccess timestamp of selected tab when launching to home

csadilek:
The `lastAccess` timestamp of a tab is updated whenever the tab is selected by the user directly or as a consequence of adding or removing a tab. A tab selection also occurs on cold start when tabs are restored. However, on warm start, when resuming to the `HomeFragment` no new tab selection occurs and the timestamp of the selected tab is outdated.


grigoryk (approved):
We paired on this, and while this is hacky it's the most logical/simple change we can make right now to improve view time tracking without larger refactoring.

PR #21411, opened 2021-09-21
For #21618: Integrate Nimbus with MR2 Home Page to enable experimentation

Amejia481:
### Pull Request checklist

- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features. In addition, it includes a screenshot of a successful [accessibility scan](https://play.google.com/store/apps/details?id=com.google.android.apps.accessibility.auditor&hl=en_US) to ensure no new defects are added to the product.

### To download an APK when reviewing a PR:
1. click on Show All Checks,
2. click Details next to "Taskcluster (pull_request)" after it appears and then finishes with a green checkmark,
3. click on the "Fenix - assemble" task, then click "Run Artifacts".
4. the APK links should be on the left side of the screen, named for each CPU architecture

PR #22481, opened 2021-11-17
For FNXV2-18419: Nimbus search term groups experiment

eliserichards:
For [FNXV2-18419](https://mozilla-hub.atlassian.net/browse/FNXV2-18419)

Creates a branch to control search term groups on home (jump back in), the tabs tray, and the setting (Settings -> Tabs -> "Search groups" toggle), and records exposure events for each.

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/debug/res/raw/initial_experiments.json
+++ b/app/src/debug/res/raw/initial_experiments.json
@@ -81,6 +81,54 @@
           "proposedEnrollment": 7,
           "userFacingDescription": "Experiment to test the home screen configurations",
           "last_modified": 1621443780172
+      },
+      {
+          "slug": "feature-search-term-groups-android",
+          "appId": "org.mozilla.fenix",
+          "appName": "fenix",
+          "channel": "nightly",
+          "branches": [
+              {
+                  "slug": "enabled",
+                  "ratio": 0,
+                  "feature": {
+                      "enabled": true,
+                      "featureId": "search-term-groups"
+                  }
+              },
+              {
+                  "slug": "disabled",
+                  "ratio": 100,
+                  "feature": {
+                      "enabled": false,
+                      "featureId": "search-term-groups"
+                  }
+              }
+          ],
+          "outcomes": [],
+          "arguments": {},
+          "probeSets": [],
+          "startDate": null,
+          "targeting": "true",
+          "featureIds": [
+              "search-term-groups"
+          ],
+          "application": "org.mozilla.firefox_beta",
+          "bucketConfig": {
+              "count": 0,
+              "start": 0,
+              "total": 10000,
+              "namespace": "nimbus-validation-2",
+              "randomizationUnit": "nimbus_id"
+          },
+          "schemaVersion": "1.5.0",
+          "userFacingName": "Search term groups test",
+          "referenceBranch": "control",
+          "proposedDuration": 14,
+          "isEnrollmentPaused": false,
+          "proposedEnrollment": 7,
+          "userFacingDescription": "Experiment to test the search term groups on tabs tray and homescreen.",
+          "last_modified": 1621443780172
       }
   ]
 }
--- a/app/src/main/java/org/mozilla/fenix/experiments/NimbusFeatures.kt
+++ b/app/src/main/java/org/mozilla/fenix/experiments/NimbusFeatures.kt
@@ -6,7 +6,9 @@ package org.mozilla.fenix.experiments
 
 import android.content.Context
 import org.mozilla.experiments.nimbus.mapKeysAsEnums
+import org.mozilla.fenix.Config
 import org.mozilla.fenix.FeatureFlags
+import org.mozilla.fenix.ReleaseChannel
 import org.mozilla.fenix.ext.components
 import org.mozilla.fenix.ext.getVariables
 
@@ -20,6 +22,10 @@ class NimbusFeatures(private val context: Context) {
 
     val homeScreen: HomeScreenFeatures by lazy {
         HomeScreenFeatures(context)
+    }
+
+    val searchTermGroups: SearchTermGroupFeatures by lazy {
+        SearchTermGroupFeatures(context, false)
     }
 
     /**
@@ -114,6 +120,25 @@ class NimbusFeatures(private val context: Context) {
          */
         fun isTopSitesActive(): Boolean {
             return homeScreenFeatures[HomeScreenSection.topSites] == true
+        }
+    }
+
+    /**
+     * Component that indicates whether search term groups are active. This accompanies
+     * the `FeatureId.SEARCH_TERM_GROUPS` feature.
+     */
+    class SearchTermGroupFeatures(
+        private val context: Context,
+        val default: Boolean = false
+    ) {
+        /**
+         * Indicates if the search term groups feature is active.
+         */
+        fun isActive(): Boolean {
+            val experiments = context.components.analytics.experiments
+            val variables = experiments.getVariables(FeatureId.SEARCH_TERM_GROUPS, false)
+            return variables.getBool("enabled")
+                ?: (default || Config.channel == ReleaseChannel.Nightly)
         }
     }
 }
--- a/app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabViewHolder.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabViewHolder.kt
@@ -9,6 +9,9 @@ import androidx.compose.ui.platform.ComposeView
 import androidx.compose.ui.platform.ViewCompositionStrategy
 import mozilla.components.lib.state.ext.observeAsComposableState
 import org.mozilla.fenix.R
+import org.mozilla.fenix.components.components
+import org.mozilla.fenix.experiments.FeatureId
+import org.mozilla.fenix.ext.recordExposureEvent
 import org.mozilla.fenix.home.HomeFragmentStore
 import org.mozilla.fenix.home.recenttabs.interactor.RecentTabInteractor
 import org.mozilla.fenix.theme.FirefoxTheme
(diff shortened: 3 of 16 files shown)


app/src/main/java/org/mozilla/fenix/experiments/NimbusFeatures.kt
```diff
@@ -116,4 +122,51 @@ class NimbusFeatures(private val context: Context) {
...
+         *
+         * These names should match the names of entries in the JSON.
+         */
+        @Suppress("EnumNaming")
+        private enum class SearchTermGroups(val default: Boolean) {
+            /**
+             * This covers search term groupings in the tabs tray, homescreen, and
+             * the tab settings.
+             */
+            searchTermGroups(true);
```
csadilek, line 141: @eliserichards Drive-by: You will want this to be turned off (set to `false`) by default. Similar to what @Amejia481 did for our 94 branch where we turn on/off home screen sections for the experiment (and default to false): https://github.com/mozilla-mobile/fenix/blob/releases_v94.0.0/app/src/main/java/org/mozilla/fenix/experiments/NimbusFeatures.kt#L54

(In 95, and on main, we default to true because the experiment will be over at that point).

Otherwise all users will see tab groups right away and for those users in control the groups would suddenly disappear when the user is enrolled (this happens async.) Also, all users who opted out of studies would also see tab groups.

So it's better to default to false. Then users in the experiment/treatment group will start seeing groups when they get enrolled, preventing the two problems above.

Amejia481: I agree with @csadilek, as users will have a better experience with the suggested approach.

Amejia481:
@eliserichards I think  we should call `recordExposureEvent` when the feature is shown to users. See an example on the [Fenix home screen experiments](https://github.com/mozilla-mobile/fenix/pull/21411/files#diff-41731714e97083a1b73584da2837ab11c22796c25b892b1e86ce4190ebba2633R374).

jhugman (commented):
A no-nit drive-by 🚲 🚗 🚌 

* I'm not sure if I understand the Fenix code properly; is the `searchTermGroup` a part of the same UI as the `homescreen`? If so, adding another variant to the `HomeScreenSectionId` enum would make the Nimbus parts of this PR much simpler.
* If not, starting at the JSON `value` in the experiment makes for some better modeling of how the feature will be configured.

```json
{
    "enabled": true
}
```

may be all you need, which has considerably less code to deal with than:

```json
{
    "sections-enabled": {
         "searchTermGroup": true
    }
}
```

FWIW: I'm looking forward to the Nimbus tooling getting better to make this easier. :( :)

eliserichards:
Thanks for the drivebys @jhugman @Amejia481 @csadilek! This was very much in-progress while I waited for some answers from product.

eliserichards:
There is an issue with the tab settings to turn off search term groups, which might need to be fixed before we go ahead with this: https://github.com/mozilla-mobile/fenix/issues/22630

app/src/debug/res/raw/initial_experiments.json
```diff
@@ -81,6 +81,54 @@
...
+          "appId": "org.mozilla.fenix",
+          "appName": "fenix",
+          "channel": "nightly",
+          "branches": [
+              {
+                  "slug": "enabled",
+                  "ratio": 0,
+                  "feature": {
+                      "enabled": true,
+                      "featureId": "search-term-groups"
```
eliserichards, line 96: Question: should this featureId (`"search-term-groups"`) match ln 104 and 114?

eliserichards: I'd also love it if someone could look at this whole addition to `initial_experiments.json`, especially lines 108+

eliserichards:
@mergifyio rebase

app/src/debug/res/raw/initial_experiments.json
Amejia481: @jhugman @travis79 could help to clarify :)

app/src/main/java/org/mozilla/fenix/settings/TabsSettingsFragment.kt
```diff
@@ -69,9 +71,15 @@ class TabsSettingsFragment : PreferenceFragmentCompat() {
...
-            it.isChecked = it.context.settings().searchTermTabGroupsAreEnabled
+            it.isVisible = it.context.settings().showSearchGroupsFeature
+            it.isChecked = it.context.settings().showSearchGroupsFeature
             it.onPreferenceChangeListener = SharedPreferenceUpdater()
+
+            if (it.context.settings().showSearchGroupsFeature) {
+                requireContext().components.analytics.experiments.recordExposureEvent(
+                    FeatureId.SEARCH_TERM_GROUPS
+                )
+            }
```
Amejia481, line 82: @jhugman @travis79 I'm not sure if having this section conditionally it's OK?

Amejia481 (approved):
It looks good to me. Just one question to be clarified.

app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabViewHolder.kt
```diff
@@ -37,6 +40,10 @@ class RecentTabViewHolder(
         composeView.setContent {
             val recentTabs = store.observeAsComposableState { state -> state.recentTabs }
 
+            if (!recentTabs.value.isNullOrEmpty()) {
+                components.analytics.experiments.recordExposureEvent(FeatureId.SEARCH_TERM_GROUPS)
+            }
+
```
Amejia481, line 46: Same as https://github.com/mozilla-mobile/fenix/pull/22481#discussion_r763355526

app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayFragment.kt
```diff
@@ -358,6 +360,12 @@ class TabsTrayFragment : AppCompatDialogFragment() {
         setFragmentResultListener(ShareFragment.RESULT_KEY) { _, _ ->
             dismissTabsTray()
         }
+
+        if (requireComponents.settings.showSearchGroupsFeature) {
+            requireContext().components.analytics.experiments.recordExposureEvent(
+                FeatureId.SEARCH_TERM_GROUPS
+            )
+        }
```
Amejia481, line 368: https://github.com/mozilla-mobile/fenix/pull/22481#discussion_r763355526

jhugman (commented):
I see I've made lots of review comments, but not actually submitted my review.

app/src/debug/res/raw/initial_experiments.json
```diff
@@ -81,6 +81,54 @@
...
+          "startDate": null,
+          "targeting": "true",
+          "featureIds": [
+              "search-term-groups"
+          ],
+          "application": "org.mozilla.firefox_beta",
+          "bucketConfig": {
+              "count": 0,
+              "start": 0,
+              "total": 10000,
```
travis79, line 120: This bucket configuration will target no-users, is that intended? If this is just for the purpose of using the Nimbus debugging tools to "opt-in" to this experiment, then it should be fine.

app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabViewHolder.kt
travis79: If the 'recentTabs' is null or empty for the control branch, we still want to record an "exposure" as it indicates the point at which a user _would have_ seen the treatment if there was one.

app/src/main/java/org/mozilla/fenix/settings/TabsSettingsFragment.kt
travis79: James is correct, the "exposure" event is intended to record the point in the application flow where the user encounters the trigger to either display the "placebo" or control experience, or the treatment experience. We want to know when this happens regardless of which branch the user is enrolled in and Nimbus will take care of the rest.

travis79 (commented):
When the open comments are addressed, I think this is looking good!

app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabViewHolder.kt
eliserichards: That totally makes sense. Thanks for the clarification!

eliserichards:
For visibility, something has changed here (i'm assuming some implementation changes on the fenix side) and this no longer turns the feature on/off. I need to investigate to see what changed.

eliserichards:
What's blocking this: 
* How to see if this Nimbus flag is working: Activate the secret settings, then go into the nimbus settings. when you manipulated the on/off setting for this experiment you could see the search term groups being created/removed in the tabs tray. 
* This was working (and the PR approved), but on a last manual test, I realized that something has changed and the settings no longer turned the search term groups on/off.
* **Todo:** figure out what changed (my best estimate is something changed some time in the second half of December) that prevents the nimbus flags from working

app/src/debug/res/raw/initial_experiments.json
eliserichards, line 120: Before this is merged, we want to double check these numbers with Devyani. (`count` can be updated to match `total` if we want all users)

eliserichards: to be clearer: this is a ratio of `start` to `count` out of `total`

e.g. 0-100 out of 10000
`start`: 0
`count`: 100
`total`: 10000

eliserichards:
This will probably need to wait for https://github.com/mozilla-mobile/fenix/pull/23400

eliserichards:
@gabrielluong that's up to Amedyne -- it's still in the backlog and was supposed to be passed off to someone else when I left the team

gabrielluong:
I will close this PR to clean up our pull request backlog since whoever does take over will probably need to open a new PR anyways. I have forwarded the conversation and decision that needs to happen.

PR #22485, opened 2021-11-18
Closes #22484 - Track direct loads per tab; fix search term lookup

### LINKED ISSUE #22484: Incorrect search groups in history
While testing out https://github.com/mozilla-mobile/fenix/issues/22472, noticed that we sometimes miss tagging a paged that was opened via SERP with the correct search terms.

Behaviour is intermittent and timing dependent.

STR:
- Search for "Toronto" (or, any other term)
- Open in a new tab link result link to city of toronto's website
- In the current tab, open a link to Toronto wikipedia page

Expectation:
- in the History view, a Toronto search group should have been formed, with three links in it: SERP, city of Toronto website, Toronto wikipedia page

Reality:
- depending on timing of when pages are opened, sometimes the Wikipedia one falls outside of the group (the "parent" page), sometimes the city of TO (the "child", new tab, page)

grigoryk:
This patch fixes two problems:
1) We were treating "direct tab load" as an event which applies
   uniformally to all tabs, even though it's actually an event which
   happens for a specific tab. This lead to background tabs (pages opened as new tab)
   setting the direct load flag, and then a simultaneously loading
   parent tab would incorrectly interpret that flag for itself.

   The patch switches this tracking from a simple boolean (are we direct
   loading?) to a set of tab IDs that are currently direct loading.

2) In a case when a background tab was loading with a parent who's
   search terms were cleared by a direct load, we were not trying to
   lookup search terms on the background tab's historyMetadata key,
   which exists to capture search terms for this exact scenario.

   The patch adds an additional fallback lookup for that path.

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/historymetadata/HistoryMetadataMiddleware.kt
+++ b/app/src/main/java/org/mozilla/fenix/historymetadata/HistoryMetadataMiddleware.kt
@@ -31,9 +31,9 @@ class HistoryMetadataMiddleware(
 
     private val logger = Logger("HistoryMetadataMiddleware")
 
-    // Tracks whether a page load is in progress that was triggered directly by the app
+    // Tracks whether a load is in progress for a tab/session ID that was triggered directly by the app
     // e.g. via the toolbar as opposed to via web content.
-    private var directLoadTriggered: Boolean = false
+    private var directLoadTriggeredSet = mutableSetOf<String>()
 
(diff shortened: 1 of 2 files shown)


csadilek (approved):
👍 

We paired on this and I did another round of testing now to confirm. Thanks for writing that tricky test! Let's talk about uplifting this one to 95.

PR #22501, opened 2021-11-19
Prevent default browser marketing from waking user

### LINKED ISSUE #20887: Ensure we schedule local push notifications at a normal time
I had all notification categories on silent, I believe.  New categories were apparently created for "marketing" , which sent me a noisy push at 2am to tell me Firefox wasn't default.  Not cool.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-17613)

interfect:
This will fix #20887.

`IMPORTANCE_DEFAULT` is documented to make an audible alert, and without any logic in here to restrict the time of day at which the notification is sent, that's not reasonable behavior. The audible alert has been observed in the wild to wake up users in the middle of the night and annoy them (see #20887).

This sets `IMPORTANCE_LOW` instead, which is as close as you can get to `IMPORTANCE_DEFAULT` without making sound. This should make sure the marketing message is more well-received.

I don't think this sort of configuration setting is really testable in the current test suite. There doesn't seem to be a test for the existing value.

No screenshot is needed because the rendered notification is the responsibility of the OS. Similarly, there's nothing here to scan for accessibility.

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/onboarding/DefaultBrowserNotificationWorker.kt
+++ b/app/src/main/java/org/mozilla/fenix/onboarding/DefaultBrowserNotificationWorker.kt
@@ -90,7 +90,7 @@ class DefaultBrowserNotificationWorker(
             val channel = NotificationChannel(
                 NOTIFICATION_CHANNEL_ID,
                 applicationContext.getString(R.string.notification_marketing_channel_name),
-                NotificationManager.IMPORTANCE_DEFAULT
+                NotificationManager.IMPORTANCE_LOW
             )
 
             notificationManager.createNotificationChannel(channel)


rocketsroger:
This is a requirement question.  @kbrosnan Do we want to lower the notification to `IMPORTANCE_LOW`?

amedyne:
@kbrosnan @athomasmoz Thoughts on this?

app/src/main/java/org/mozilla/fenix/onboarding/DefaultBrowserNotificationWorker.kt
```diff
@@ -90,7 +90,7 @@ class DefaultBrowserNotificationWorker(
             val channel = NotificationChannel(
                 NOTIFICATION_CHANNEL_ID,
                 applicationContext.getString(R.string.notification_marketing_channel_name),
-                NotificationManager.IMPORTANCE_DEFAULT
+                NotificationManager.IMPORTANCE_LOW
             )
```
pocmo, line 94: Random drive-by 🚗: Afaik a `Channel` needs to be recreated for the importance changes to take effect. Otherwise this affects only new users.

interfect: Isn't this channel used just for this one notification? There might not actually be many existing users; only the ones upgrading in the first 24 hours after initial setup.

How does one go about recreating a `Channel`?

interfect:
Sorry, Mergify bot. I don't think that it would be worth my time to rebase this, since it had no conflicts for months and still didn't get merged, since there doesn't seem to be a process that gets engaged to make a decision to accept or reject a PR in a bounded amount of time.

interfect:
@rocketsroger Is this worth the time to rebase? It seems to have degenerated into a lot of people asking for and not getting a lot of other people's opinions.

rocketsroger:
> @rocketsroger Is this worth the time to rebase? It seems to have degenerated into a lot of people asking for and not getting a lot of other people's opinions.

@interfect I'll follow up with product again.  I'll get a clear answer before the end of this week.  Please don't rebase yet.  Thanks.

interfect:
Thanks for digging up an answer @rocketsroger! I'll go ahead and close this then.

But I don't feel like the way this decision was reached quite matches "transparent community-based processes" or promotes "participation, accountability and trust". And not just because it's my PR.

On #20887 and #23726 I have seen users asking for this to be changed, and I haven't seen any users stepping forward to say that they, as users, like the way it works now and appreciated the beeping. Maybe the product team has better information on what users really want in aggregate---I've seen other Github-unpopular product decisions justified by reference to internal metrics that say the people complaining are a negligible fraction---but it sounds like they didn't mention it to you, and in this case that seems unlikely. Glean is not going to be able to measure if the user is annoyed, and nobody is doing 24-hour-long user studies.

So when it is made clear that this is the way that the product team wants it to work, but without a convincing accounting of why despite appearences it is actually best for users for it to be this way, it ends up looking like the product team's design goal is something other than making the most helpful tool for the end user.

Issue #22489, opened 2021-11-18
Investigate removing Fennec -> Fenix migration code


pocmo:
We are contemplating removing the "Fennec to Fenix" migration code from Fenix. We shipped the first Fenix release with migration code over a year ago (~ Summer 2020). Removing the code will reduce the maintenance burden, improve performance (#15581) and reduce the APK size.
We still need to go through the process of looking at data, getting approval etc. - but that doesn't stop us from investigating the engineering work already.

cadeyrn:
> Removing the code will reduce the maintenance burden, improve performance (#15581) and reduce the APK size.

@pocmo: Also the migration of passwords for users with master password can cause a crash: #22279. Removing the migration code would make fixing this crash obsolete.

pocmo:
@cadeyrn Great, thank you for mentioning that!

philipwatmor:
What makes us unique is our solid ability, joined with an essential methodology, scrupulousness and severe working techniques. Our administrations stretch out to the gathering of kitchen furniture and furniture, all things considered, the selection of tones and the plan. We support you in all periods of get together, from key preparation and data design to execution and then some. Since for us, managing your furniture implies how managing you.[https://pimpmykitchen.at](url)

PR #22507, opened 2021-11-19
Issue #22489: Remove "Fennec to Fenix" migration code

pocmo:
Do not land: This still needs to go through the process of looking at data, getting approval etc.

And this needs to be tested thoroughly to not have any unwanted side effects (For example launcher icons disappearing, or sharedUserId not getting set correctly).


app/src/migration/java/org/mozilla/fenix/MigrationPushRenewer.kt
```diff
@@ -1,40 +0,0 @@
...
-        store.flowScoped { flow ->
-            flow.collect { state ->
-                Logger("MigrationPushRenewer").debug("Migration state: ${state.progress}")
-                if (state.progress == MigrationProgress.COMPLETED) {
-                    Logger("MigrationPushRenewer").debug("Renewing registration....")
-
-                    // This should force a recreation of firebase device token, re-registration with
-                    // the autopush service, and subsequent update of the FxA device record with
-                    // new push subscription information.
-                    service?.renewRegistration()
```
pocmo, line 35: @jonalmeida What will be the side effect of not running this for a Fenix users on upgrade? Will they ever recover from this or will they be stuck forever?

jonalmeida: TL;DR: We should be fine for Sync and web content subscriptions may be able to recover on their own.

There are three things here to consider: GCM/Firebase push token, WebPush subscriptions, AutoPush keys

When migrating, away from GCM, Firebase servers knows how to handle the old GCM token and tell FCM clients to request a new token. A new token is then submitted to AutoPush so our servers are aware of the token change. So that covers the two cases.

For existing subscriptions, AutoPush would need to tell us to request new subscriptions or another means. With Sync, if the Sync servers cannot communicate to the device, a future sync from polling would tell the client that push is not working which will trigger the renew process. However if the user doesn't have Sync, then browser clients will not know to renew their subscriptions, but this something that web clients usually should re-request. With mhammond's recent patches for push, this should make the process better along with other [future changes](https://docs.google.com/document/d/11mUtkK6lzSNuzjoYw1K0lDBRGuACbfTIHIXSDibX540/edit#).

mcomella:
@pocmo Is there anything blocking this PR? When do you expect it'd land?

pocmo:
@mcomella It needs sign-off from stakeholders (@st3fan is on the hook for that) and it was decided to not land it right now - to avoid landing a big risk change before the holidays.

grigoryk:
Good time to revisit this!

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -239,13 +238,7 @@ class HomeFragment : Fragment() {
                     topSites = components.core.topSitesStorage.cachedTopSites,
                     tip = components.strictMode.resetAfter(StrictMode.allowThreadDiskReads()) {
                         FenixTipManager(
```
gabrielluong, line 240: Just making a note, but it would be nice to investigate if we need `FenixTipManager` after this removal.

app/src/main/res/values/strings.xml
```diff
@@ -1296,31 +1296,31 @@
 
     <!-- Tips -->
     <!-- text for firefox preview moving tip header "Firefox Preview" and "Firefox Nightly" are intentionally hardcoded -->
-    <string name="tip_firefox_preview_moved_header">Firefox Preview is now Firefox Nightly</string>
+    <string name="tip_firefox_preview_moved_header" moz:removedIn="96" tools:ignore="UnusedResources">Firefox Preview is now Firefox Nightly</string>
```
gabrielluong, line 1299: All these strings need to be bumped to 98

csadilek:
We've got sign-off to review and land this now.

csadilek:
Ah saw that there's a new PR opened: https://github.com/mozilla-mobile/fenix/pull/23651

Closing.

PR #22512, opened 2021-11-19
Bug 1742138 - Remove non-webrender browsertime tests.

gmierz:
This patch disabled non-webrender browsertime tests running on the fenix branch and removes the code that created the tasks.


jmaher (approved):
+1 to removing a transform, thanks for jumping on this.

mcomella:
@gmierz @jmaher I don't think you need a fenix reviewer – the automatic review request doesn't actually block merging. I think you can merge this yourself if you have the ability to add labels (if not, let me know) – you add the `pr:needs-landing` label and the PR will be automatically merged. The standard way to merge PRs in GitHub is disabled which might be confusing.

gmierz:
Ah ok, thanks @mcomella. I thought it was required because it took about a few days for the patch to get merged.

Issue #5298, opened 2019-09-13
[Bug] The same page appears in the History section multiple times


LaurentiuApahideanSV:
## Steps to reproduce
1. Open the same page multiple times;
2. Go to Your Library -> History.
### Expected behavior
The page that was opened multiple times should only show up once in the History section.
### Actual behavior
The same page appears in the History section multiple times.
### Device information

* Android device: Samsung galaxy S8+ (Android 8.0.0)
* Fenix version: 2.0.0-rc.2 9/12
![Screenshot_20190913-140632_Firefox Preview](https://user-images.githubusercontent.com/50983519/64858509-496a7a80-d630-11e9-87b6-792997acc30b.jpg)

nahuhh:
Confirmed on Nightly 190912 20:44 (Build #12552054)
📦: 13.0.0, ccf305cea
🦎: 71.0a1-20190910095613

SubhrajyotiSen:
@sv-ohorvath What would be the approach for this? Show unique items by default or have a toggle for that?

sv-ohorvath:
@vesta0 would be more suitable to answer this.

ekager:
When we redid history with pagination we must have clipped this fix. See
https://github.com/mozilla-mobile/fenix/pull/4122 for previous solution

vesta0:
@ekager can we add it back in?

boek:
@vesta0 no, not easily. We will probably need this filtered from application services.

codrut-topliceanu:
@ekager @boek Hey, do we have a decision on how we should proceed here?

softvision-miralobontiu:
This is reproducible on Beta 5.2.0-beta.1 using Samsung Galaxy Tab S6 (Android 9).

softvision-miralobontiu:
I was able to reproduce this issue on 83.0.0-beta 1 build with Samsung Galaxy Tab S6 (Android 10).

ebalazs-sv:
This issue is reproducible on Beta 88.0.0-beta.1 with Samsung Galaxy Tab S3 (Android 8).

lobontiumira:
Reproducible on Beta 89.0.0-beta.1 with Samsung Galaxy Note 8 (Android 9).

Amejia481:
This should be available in the next nightly @LaurentiuApahideanSV could you please help us to verify? Thanks in advance!

LaurentiuApahideanSV:
[...]

PR #22530, opened 2021-11-22
For issue #5298 the same page appears in the History section multiple times

Amejia481:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/test/java/org/mozilla/fenix/components/history/PagedHistoryProviderTest.kt
+++ b/app/src/test/java/org/mozilla/fenix/components/history/PagedHistoryProviderTest.kt
@@ -452,4 +452,74 @@ class PagedHistoryProviderTest {
         )
         assertEquals(results, actualResults)
     }
+
+    @Test
+    fun `WHEN removeConsecutiveDuplicates is called THEN all consecutive duplicates must be removed`() {
+        val results = listOf(
+            History.Group(
+                id = 1,
+                title = "Group 1",
+                visitedAt = 0,
+                items = emptyList()
+            ),
+            History.Regular(
+                id = 2,
+                title = "No duplicate item",
+                url = "url",
+                visitedAt = 0
+            ),
+            History.Regular(
+                id = 3,
+                title = "Duplicate item 1",
+                url = "url",
+                visitedAt = 0
+            ),
+            History.Regular(
+                id = 4,
+                title = "Duplicate item 2",
+                url = "url",
+                visitedAt = 0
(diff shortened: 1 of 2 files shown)


app/src/test/java/org/mozilla/fenix/components/history/PagedHistoryProviderTest.kt
```diff
@@ -452,4 +452,62 @@ class PagedHistoryProviderTest {
...
+                title = "No duplicate item 2",
+                url = "url",
+                visitedAt = 0
+            ),
+            History.Group(
+                id = 5,
+                title = "Group 5",
+                visitedAt = 0,
+                items = emptyList()
+            ),
```
csadilek, line 488: nit: Let's add another (single) item with the same url at the end to test that we really only filter consecutive duplicates.

app/src/test/java/org/mozilla/fenix/components/history/PagedHistoryProviderTest.kt
```diff
@@ -452,4 +452,62 @@ class PagedHistoryProviderTest {
...
+                title = "No duplicate item",
+                url = "url",
+                visitedAt = 0
+            ),
+            History.Group(
+                id = 5,
+                title = "Group 5",
+                visitedAt = 0,
+                items = emptyList()
+            ),
```
csadilek, line 509: And here ^^ as well to verify it's still in the result after.

csadilek (approved):
👍 

One suggestion below to improve the test. Overall, I think this makes the history view better and we can live with the edge case of still seeing duplicates when there's not enough history to populated a full page yet.

Amejia481:
> 👍
> 
> One suggestion below to improve the test. Overall, I think this makes the history view better and we can live with the edge case of still seeing duplicates when there's not enough history to populated a full page yet.

👍🏽

app/src/test/java/org/mozilla/fenix/components/history/PagedHistoryProviderTest.kt
Amejia481: Added :)!

app/src/test/java/org/mozilla/fenix/components/history/PagedHistoryProviderTest.kt
Amejia481: Added :)!

csadilek (approved):

Issue #22534, opened 2021-11-22
Show history highlights in "Recently visited"


Mugurell:
Help users easily locate relevant content from a previous browsing session by organizing and highlighting content that was meaningful and engaging, and filtering out the noise. 

Tracking https://mozilla-hub.atlassian.net/browse/FNXV2-17824

### Acceptance criteria
- Section has been renamed from “Recent Searches” to “Recently Visited”
- When I open Home, I see a combination of individual tabs that are history highlights and history search term Groups in the “Recently Visited” section. 
- We will show the top 9 scored items. 
- We will show a min of 3 and a max of 6 groups (when available) and a min of 3 and a max of 6 highlights (when available) up to a total of 9 items.
- ~~“Recently Visited” section will include only closed tabs + groups with ALL closed tabs~~

LaurentiuApahideanSV:
I tested the issue on Firefox 96.0.0-beta.1. The history highlights are displayed in the Recently visited section alongside regular websites. I will mark this issue as fixed.
### Devices used:
- Huawei MediaPad M3 (Android 7)
- OnePlus 6T (Android 11)
![Screenshot_20211209-180408](https://user-images.githubusercontent.com/50983519/145432684-9744c87a-4dd3-4a84-80d2-556c67a83fbf.png)

PR #22535, opened 2021-11-22
For #22534 - Show history highlights and groups in "Recently visited"

Mugurell:
For now the metadata groups don't support scoring so as an interim solution we
will show up to 9 items, evenly distributes, first favoring groups sorted by date
then history highlights pre-sorted by default.

Tapping a history highlight will switch to it's already open tab if available
or create a new one in which to load it if needed.

A "Remove" option will also be available for history highlights to remove it from
the screen and also from history.
Currently removing a group / highlight will not query new ones to again show up to
9 items, this will be implemented separately.

The updated feature supports more than history metadata so updating the overall
naming scheme seems needed.
To signal that this is a homescreen feature the entire package is moved to home.

https://user-images.githubusercontent.com/11428869/143200226-0499c429-12bf-4d99-a045-b7dbdf1ac213.mp4

### Pull Request checklist

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/historymetadata/HistoryMetadataFeature.kt
```diff
@@ -4,77 +4,124 @@
...
+        )
+    }
+
+    @VisibleForTesting
+    @Suppress("MagicNumber")
+    internal fun getCombinedHistory(
+        historyHighlights: List<RecentHistoryHighlight>,
+        historyGroups: List<RecentHistoryGroup>,
+    ): List<RecentlyVisitedItem> {
+        val recentGroups = historyGroups.take(3)
```
gabrielluong: Maybe add another const to for the 3 to remove the MagicNumber.

app/src/main/java/org/mozilla/fenix/historymetadata/HistoryMetadataFeature.kt
```diff
@@ -4,77 +4,124 @@
...
+                            title = title,
+                            historyMetadata = data
+                        )
+                    }
+                    .sortedByDescending { it.lastUpdated() }
+                    .take(MAX_RESULTS_OF_EACH_TYPE)
+            }
 
-            homeStore.dispatch(HomeFragmentAction.HistoryMetadataChange(historyMetadata))
+            updateState(historyHighlights.await(), recentHistoryGroups.await())
```
gabrielluong: Curious from a learning point of view, why we decided to make use of async/await?

app/src/main/java/org/mozilla/fenix/home/recentvisits/controller/RecentVisitsController.kt
```diff
@@ -80,19 +88,30 @@ class DefaultHistoryMetadataController(
         )
     }
 
-    override fun handleRemoveRecentHistoryGroup(searchTerm: String) {
+    /**
+     * Removes a [RecentHistoryGroup] with the given title from the homescreen.
+     *
+     * @param groupTitle The title of the [RecentHistoryGroup] to be removed.
+     */
+    override fun handleRemoveRecentHistoryGroup(groupTitle: String) {
```
gabrielluong, line 96: I would probably keep this as `searchTerm` instead of `groupTitle` since all its subsequent callers use `searchTerm`

app/src/main/java/org/mozilla/fenix/home/recentvisits/controller/RecentVisitsController.kt
```diff
@@ -16,60 +16,68 @@ import mozilla.components.feature.tabs.TabsUseCases.SelectOrAddUseCase
...
     private val navController: NavController,
     private val storage: HistoryMetadataStorage,
     private val scope: CoroutineScope,
     private val metrics: MetricController
-) : HistoryMetadataController {
+) : RecentVisitsController {
 
+    /**
+     * Shows the history fragment.
+     */
```
gabrielluong, line 68: I have a history of saying that we don't necessarily need a kDoc for overridden functions since they are defined in the Interface. I don't think I have a strong opinion on it, but I guess it makes it easier to just change the kDoc in only one place versus several when you need to update it.

app/src/main/java/org/mozilla/fenix/home/recentvisits/interactor/RecentVisitsInteractor.kt
```diff
@@ -0,0 +1,40 @@
...
+import org.mozilla.fenix.home.recentvisits.RecentlyVisitedItem.RecentHistoryGroup
+import org.mozilla.fenix.home.recentvisits.RecentlyVisitedItem.RecentHistoryHighlight
+
+/**
+ * All possible user interactions with the "Recent visits" section.
+ */
+interface RecentVisitsInteractor {
+
+    /**
+     * User clicked on the "Show all" button besides the recent visits header.
```
gabrielluong, line 16: The kDoc deviates quite a bit from our established pattern for Interactor kDocs. I would prefer to preserve the existing pattern where the kDoc of an Interaction function should describe what the controller will do (since we usually use @See in the controller function and that extra bit of information is much more helpful when learning and navigating the code otherwise I think the current description as-is does not add much since you can just read the function name as well) and what user interaction triggers the function call.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlInteractor.kt
```diff
@@ -383,21 +383,24 @@ class SessionControlInteractor(
...
 
-    override fun onRemoveRecentHistoryGroup(searchTerm: String) {
-        historyMetadataController.handleRemoveRecentHistoryGroup(searchTerm)
+    override fun onRemoveRecentHistoryGroup(groupTitle: String) {
+        recentVisitsController.handleRemoveRecentHistoryGroup(groupTitle)
     }
 
+    /**
+     *
+     */
```
gabrielluong, line 401: Can probably remove this

app/src/main/java/org/mozilla/fenix/compose/Favicon.kt
```diff
@@ -0,0 +1,73 @@
...
+                    color = when (isSystemInDarkTheme()) {
+                        true -> PhotonColors.DarkGrey30
+                        false -> PhotonColors.LightGrey30
+                    }
+                )
+            )
+        }
+
+        WithIcon { icon ->
+            androidx.compose.foundation.Image(
```
gabrielluong, line 56: Can we import this to drop `androidx.compose.foundation.`?

app/src/main/java/org/mozilla/fenix/home/recentvisits/view/RecentlyVisited.kt
```diff
@@ -191,6 +199,63 @@ private fun RecentVisitItem(
...
 
+/**
+ * A recently visited history item.
+ *
+ * @param recentVisit The [RecentHistoryHighlight] to display.
+ * @param showDividerLine Whether to show a divider line at the bottom.
+ * @param onRecentVisitClick Invoked when the user clicks on a recent visit.
+ */
+@Composable
+private fun RecentlyVisitedHighlight(
```
gabrielluong, line 210:
```suggestion
private fun RecentlyVisitedHistoryHighlight(
```
I don't know if this adds clarity, maybe?

app/src/main/java/org/mozilla/fenix/home/recentvisits/view/RecentlyVisited.kt
```diff
@@ -213,17 +278,33 @@ private fun RecentlyVisitedCaption(count: Int) {
     )
 }
 
+/**
+ * Recent items divider.
```
gabrielluong, line 282:
```suggestion
 * A recent item divider.
```

app/src/main/java/org/mozilla/fenix/home/recentvisits/view/RecentlyVisited.kt
```diff
@@ -213,17 +278,33 @@ private fun RecentlyVisitedCaption(count: Int) {
...
+ *
+ * @param modifier [Modifier] allowing to perfectly place this.
```
gabrielluong, line 284:
```suggestion
 * @param modifier Optional [Modifier] to be applied to the layout.
```
Using one of your previous kDoc description so we hopefully don't keep rewording this differently everywhere.

app/src/main/java/org/mozilla/fenix/home/recentvisits/view/RecentlyVisited.kt
```diff
@@ -191,6 +199,63 @@ private fun RecentVisitItem(
...
+            }
+        }
+    }
+}
+
+/**
+ * The title of a recent visit.
+ *
+ * @param text [String] that will be display. Will be ellipsized if cannot fit on one line.
+ * @param modifier [Modifier] allowing to perfectly place this.
```
gabrielluong, line 242:
```suggestion
 * @param modifier Optional [Modifier] to be applied to the layout.
```
Using one of your previous kDoc description so we hopefully don't keep rewording this differently everywhere.

app/src/main/java/org/mozilla/fenix/home/recentvisits/RecentVisitsFeature.kt
```diff
@@ -0,0 +1,131 @@
...
+    private val historyMetadataStorage: HistoryMetadataStorage,
+    private val historyHighlightsStorage: Lazy<PlacesHistoryStorage>,
+    private val scope: CoroutineScope,
+    private val ioDispatcher: CoroutineDispatcher = Dispatchers.IO,
+) : LifecycleAwareFeature {
+
+    private var job: Job? = null
+
+    override fun start() {
+        job = scope.launch(ioDispatcher) {
```
gabrielluong, line 55: Probably prefer to have @csadilek review this portion of the PR

app/src/main/java/org/mozilla/fenix/historymetadata/HistoryMetadataFeature.kt
Mugurell: Was thinking that this is not really part of the specs but can add it above but just a temporary approach.
Will add it as a top level const.

app/src/main/java/org/mozilla/fenix/historymetadata/HistoryMetadataFeature.kt
Mugurell: This allows for concurrent coroutines for querying highlights and metadata. Otherwise inside the `launch` these would be executed serially.

app/src/main/java/org/mozilla/fenix/home/recentvisits/controller/RecentVisitsController.kt
Mugurell: But the [RecentHistoryGroup] doesn't have a `searchTerm` property and this method is called for that type.
I could rename it's `title` property to `searchTerm` but feel like `title` is more appropriate for where and how it's used. (The other item - `RecentHistoryHighlight` also has a `title` property)

app/src/main/java/org/mozilla/fenix/home/recentvisits/controller/RecentVisitsController.kt
Mugurell: I tried cleaning up some of the comments for the interactor and the controller.
The interface as a general contract should not dictate the behavior. The implementation decides that for handling that action user will be navigated to the history fragment. Other implementations could have different behavior.

Another important refactoring here would've been to remove the references to the interactor from controller's interface. The controller is used by the interactor so it should not know about it.

app/src/main/java/org/mozilla/fenix/home/recentvisits/interactor/RecentVisitsInteractor.kt
Mugurell: The same as the above, I'd say the interface is just a general contract with different implementations then having different behavior.
The controller is passed as an argument to an implementation of this interface (it might not be for all) so the general contract would not know what happens after. 
Loose coupling I'd say it's an advantage in the long term that contributes to the cleanliness of the code now.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlInteractor.kt
Mugurell: Started to add a comment but couldn't come up with something better than in the interface :-)).
Will remove, thank you!

app/src/main/java/org/mozilla/fenix/compose/Favicon.kt
Mugurell: Thank you!
Tried somehow importing that `Loader` from above but didn't try with `Image` also.

app/src/main/java/org/mozilla/fenix/home/recentvisits/view/RecentlyVisited.kt
Mugurell: Inside the same same file I'd say the impact is small but this aligns the naming to the other `RecentlyVisitedHistoryGroup` composable. Thanks!

app/src/main/java/org/mozilla/fenix/home/recentvisits/view/RecentlyVisited.kt
Mugurell: I would't say this is optional.
Though about even not providing a default here to force users pass a Modifier.
Saw there are differences in how this composable should be placed depending on which type of history is shown so while a Modifier could control a lot of different options that comment should hint to the users that it is intended to help place this appropriately.

app/src/main/java/org/mozilla/fenix/home/recentvisits/RecentVisitsFeature.kt
Mugurell: Sure! Thank you for the quick review!

app/src/main/java/org/mozilla/fenix/compose/Favicon.kt
```diff
@@ -0,0 +1,73 @@
...
+ * @param url Website [URL] for which the favicon will be shown.
+ * @param size [Dp] height and width of the image to be loaded.
+ */
+@Composable
+fun Favicon(
+    url: String,
+    size: Dp,
+    isPrivate: Boolean = false
+) {
+
```
gabrielluong, line 38:
```suggestion
```

app/src/main/java/org/mozilla/fenix/home/recentvisits/controller/RecentVisitsController.kt
gabrielluong: I think what you have proposed is good. I think what I would like to see then is that we move these descriptions to replace the @see kDocs we have in the controller interface. I would also like to see the same kind of "Callback" wording that you used in your Pocket Interactor/Controllers.

app/src/main/java/org/mozilla/fenix/home/recentvisits/interactor/RecentVisitsInteractor.kt
gabrielluong: To reiterate https://github.com/mozilla-mobile/fenix/pull/22535#discussion_r754654883, I think we should go back towards the "Callback" phrasing of these descriptions that you used in the Pocket Interactor/Controllers. I think what you have proposed is good, but I would like to ensure we are consistent with how we write these kDocs throughout the codebase so that will serve as something that people will continue to emulate and repeat.

app/src/main/java/org/mozilla/fenix/home/recentvisits/RecentVisitsFeature.kt
```diff
@@ -0,0 +1,131 @@
...
+        historyHighlights: List<RecentHistoryHighlight>,
+        historyGroups: List<RecentHistoryGroup>,
+    ): List<RecentlyVisitedItem> {
+        // Temporary algorithm while scoring is added to metadata history
+        // after which both types will be sortable by their score.
+        val recentGroups = historyGroups.take(3)
+        val recentHighlights = historyHighlights
+            .take((MAX_RESULTS_TOTAL - recentGroups.size).coerceAtMost(MAX_RESULTS_OF_EACH_TYPE))
+
+        return (recentGroups + recentHighlights)
```
csadilek, line 125: Just a first pass. There are two more cases we need to cover here:

- If there are not enough highlights we should "fill up" with groups. So 3 groups is the minimum, but we should still have 9 groups if there are no highlights, or 6 groups if there are 3 highlights.

- If a page is already in a displayed group, we should not also add it separately as a highlight. Otherwise, we can be in a situation where the highlights are (mostly) the same as the items in a group e.g. after extensive research on a topic.

csadilek: Everything else looks good to me, but I will take a closer look and test more!

Mugurell: > If there are not enough highlights we should "fill up" with groups. So 3 groups is the minimum, but we should still have 9 groups if there are no highlights, or 6 groups if there are 3 highlights.

This differs a bit from the specs on Jira based on which I understood there's a hard limit of 6 items of a type.
(The hardcoded max 3 here was just a temporary approach until scoring is available for groups also).
Without that limit of 6 the algorithm then becomes `show a total of 9 groups and highlights distributed evenly or as many as available from each type`.
Since we have to show an uneven number of items of two types one type will always appear more than the other. Will be going with groups. Changing to favor highlights is easy.

One question I have related to groups: should we filter out groups only having only 1 metadata entry similar to [what we do with search groups for "Jump back in"](https://github.com/mozilla-mobile/fenix/blob/116a2f0c2bd9a279476ee0828a9f120e8870980b/app/src/main/java/org/mozilla/fenix/ext/BrowserState.kt#L118) ?

csadilek: Yes, it differs a bit because the edge case wasn't thought about that there could be no highlights at all, which isn't very likely. I mostly pointed it out because we should fill up with groups then. If you want to fill up just to get to 6 that's good enough too. We will likely always have highlights anyway and end up with the 3 most recent groups. (Until we introduce scoring for groups.)

> One question I have related to groups: should we filter out groups only having only 1 metadata entry similar to.
No, for history we decided that groups of 1 item are fine, but it's not a high confidence decision, just a separate debate, plus we'd have to change this in history as well. So, let's keep this independent. :)

csadilek (commented):
@Mugurell another thing here is that we can't land this without some form of removal. I know the longer term plan is being discussed (likely via a block list or similar), but short-term we can do the same as we do for groups and simply show a remove option and delete all metadata for that highlight/url. 

So, for groups we call: https://github.com/mozilla-mobile/android-components/blob/main/components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/PlacesHistoryStorage.kt#L320

And we'd add a similar call which we'd invoke when removing highlights.

csadilek:
@Mugurell opened a PR to add this removal API we should use in the interim for deleting highlights: https://github.com/mozilla-mobile/android-components/pull/11307 (Talked to Nicole as well, we can talk about a better removal/delete approach separately.)

app/src/main/java/org/mozilla/fenix/historymetadata/HistoryMetadataFeature.kt
Mugurell: Moved away from this logic in the latest version of the patch.

app/src/main/java/org/mozilla/fenix/home/recentvisits/RecentVisitsFeature.kt
```diff
@@ -0,0 +1,161 @@
...
+            )
+        )
+    }
+
+    @VisibleForTesting
+    internal fun getCombinedHistory(
+        historyHighlights: List<HistoryHighlight>,
+        historyGroups: List<Pair<String, List<HistoryMetadata>>>
+    ): List<RecentlyVisitedItem> {
+        val dedupedHistoryHighlights = dedupeVisits(historyHighlights, historyGroups)
```
csadilek, line 105: nit: Let's rename this function to `dedupePages` or `dedupeUrl`, because it's not technically visits.

app/src/main/java/org/mozilla/fenix/home/recentvisits/RecentVisitsFeature.kt
```diff
@@ -0,0 +1,161 @@
...
+            .take(MAX_RESULTS_TOTAL)
+
+        val totalItemsCount = recentHighlights.size + recentGroups.size
+
+        return if (totalItemsCount <= MAX_RESULTS_TOTAL) {
+            recentGroups + recentHighlights
+        } else {
+            var groupsCount = 0
+            var highlightCount = 0
+            while ((highlightCount + groupsCount) < MAX_RESULTS_TOTAL) {
```
csadilek, line 130: ⚠️ There's a risk here that this loops infinitely. If both conditions aren't met we never increment any counter.

app/src/main/java/org/mozilla/fenix/home/recentvisits/RecentVisitsFeature.kt
```diff
@@ -0,0 +1,161 @@
...
+                if (recentGroups.getOrNull(groupsCount) != null) {
+                    groupsCount += 1
+                }
+
+                if ((highlightCount + groupsCount) < MAX_RESULTS_TOTAL &&
```
csadilek, line 135: As discussed, let's add highlights first then groups.

app/src/main/java/org/mozilla/fenix/home/recentvisits/RecentVisitsFeature.kt
```diff
@@ -0,0 +1,161 @@
...
+                    title = title,
+                    historyMetadata = data
+                )
+            }
+            .sortedByDescending { it.lastUpdated() }
+            .take(MAX_RESULTS_TOTAL)
+
+        val totalItemsCount = recentHighlights.size + recentGroups.size
+
+        return if (totalItemsCount <= MAX_RESULTS_TOTAL) {
```
csadilek, line 125: We verified that we update the requirements as this is a better solution here! 👍 (but two changes below.)

app/src/main/java/org/mozilla/fenix/home/recentvisits/RecentVisitsFeature.kt
```diff
@@ -0,0 +1,161 @@
...
+                    highlightCount += 1
+                }
+            }
+
+            recentGroups.take(groupsCount) + recentHighlights.take(highlightCount)
+        }
+    }
+
+    @VisibleForTesting
+    internal fun dedupeVisits(
```
csadilek, line 147: nit: Let's add a comment that we do this to not display a highlight separately that is already in a group.

gabrielluong:
Will do my final pass tomorrow.

app/src/main/java/org/mozilla/fenix/home/recentvisits/RecentVisitsFeature.kt
Mugurell: Looked more at this.
I think the first check from the `if` block will guarantee we have enough items for the counters which just count this items of both types to be incremented up to the `MAX_RESULTS_TOTAL`
```
val totalItemsCount = recentHighlights.size + recentGroups.size

return if (totalItemsCount <= MAX_RESULTS_TOTAL) {
    recentGroups + recentHighlights
} else {
    var groupsCount = 0
    var highlightCount = 0
    while ((highlightCount + groupsCount) < MAX_RESULTS_TOTAL) { .. }
```
In the `while` block I'm using `recentHighlights.getOrNull(highlightCount)` to increment the counters starting from position 0 and if I know I have at least `MAX_RESULTS_TOTAL` items available then that `getOrNull` check has to return a real instance at least for one type up to that total we want.

Not sure what the edgecase would be.

csadilek: Ah yes :). I missed that this is in the else block where it's guaranteed that we have at least MAX_RESULTS_TOTAL results.

Mugurell:
After syncing with Christian I've updated the PR with sorting highlights and groups together.
Tested and it works but there is still a  bit of cleanup needed and have to also add unit tests.

app/src/main/java/org/mozilla/fenix/home/recentvisits/RecentVisitsFeature.kt
```diff
@@ -0,0 +1,258 @@
...
+                    HistoryHighlightWeights(MIN_VIEW_TIME_OF_HIGHLIGHT, MIN_FREQUENCY_OF_HIGHLIGHT),
+                    MAX_RESULTS_TOTAL
+                )
+            }
+
+            val allHistoryMetadata = async {
+                historyMetadataStorage.getHistoryMetadataSince(Long.MIN_VALUE)
+            }
+
+            // val dedupedHistoryHighlights = dedupeVisits(historyHighlights.await(), allHistoryMetadata.await())
```
csadilek, line 68: nit: ^^ leftover.

app/src/main/java/org/mozilla/fenix/home/recentvisits/RecentVisitsFeature.kt
```diff
@@ -0,0 +1,258 @@
...
+                    historyMetadata = it.groupItems
+                )
+            } }
+    }
+
+    /**
+     * Filter out highlights that are already part of a history group.
+     */
+    @VisibleForTesting
+    internal fun List<FeatureHistoryHighlight>.filterByMetadata(
```
csadilek, line 227: nit: I think this would be easier to read if it was `removeHighlightsAlreadyInGroups`.

app/src/main/java/org/mozilla/fenix/home/recentvisits/RecentVisitsFeature.kt
```diff
@@ -0,0 +1,258 @@
...
+        }
+    }
+
+    override fun stop() {
+        job?.cancel()
+    }
+}
+
+@VisibleForTesting
+internal sealed class FeatureHistoryItems {
```
csadilek, line 245: nit: The `Feature` prefix confused me a bit reading through this. How about being more feature-specific

`RecentlyVisitedItem`
-> `HistoryHighlight`
->  `HistorySearchGroup`

app/src/main/java/org/mozilla/fenix/home/recentvisits/RecentVisitsFeature.kt
```diff
@@ -0,0 +1,258 @@
...
+
+    /**
+     * Do an in-memory query for the metadata of history highlights to get each item's last accessed date.
+     *
+     * Safe to call:
+     *   - if [metadata] doesn't contain the details for all [highlights].
+     *   The highlights for which the last access tiem is not found will have it set to a default of `0`.
+     *   - if the same [HistoryHighlight.url] appears multiple times in [metadata].
+     *   The latest time of any of these urls was last accessed will be associated to the highlight.
+     *
```
csadilek, line 85: nit: 

If a [highlight] cannot be mapped to a corresponding [metadata] record, its `lastAccessTime` will be set to 0.
If a [highlight] maps to multiple metadata records, its `lastAccessTime` will be set to the most recently updated record.

app/src/main/java/org/mozilla/fenix/home/recentvisits/RecentVisitsFeature.kt
```diff
@@ -0,0 +1,258 @@
...
+
+            val featureHighlights = getFeatureHighlights(historyHighlights.await(), allHistoryMetadata.await())
+            val featureGroups = getFeatureGroups(allHistoryMetadata.await())
+
+            updateState(featureHighlights, featureGroups)
+        }
+    }
+
+    /**
+     * Do an in-memory query for the metadata of history highlights to get each item's last accessed date.
```
csadilek, line 78: nit: Perform an in-memory mapping of a history highlights to metadata records to compute the last access times of highlights.

app/src/main/java/org/mozilla/fenix/home/recentvisits/RecentVisitsFeature.kt
```diff
@@ -0,0 +1,258 @@
...
+     *   - if the same [HistoryHighlight.url] appears multiple times in [metadata].
+     *   The latest time of any of these urls was last accessed will be associated to the highlight.
+     *
+     * @param highlights List<HistoryHighlight> for which to get the last accessed time.
+     * @param metadata List<HistoryMetadata> expected to contain the details for all [highlights].
+     *
+     * @return [Map<HistoryHighlight, Long>] The [highlights] with added time last accessed time.
+     */
+    @VisibleForTesting
+    internal fun getFeatureHighlights(
```
csadilek, line 92: nit: see naming suggestion below. Would prefer just `getHistoryHighlights` and `getHistorySearchGroups`.

csadilek (commented):
Algorithm and feature logic look good to me! I left a few suggestions / nits for naming I would prefer updating before landing.

I think this can land once cleanup is addressed and tests are added, if @gabrielluong is fine with the UI bits!

app/src/main/java/org/mozilla/fenix/home/recentvisits/RecentVisitsFeature.kt
csadilek: Ah I see, you wanted to use different names here to not collide with the types you add to the store. Hm. Makes me wonder could you use the same types for this and add `lastAccessTime` there in RecentlyVisitedItem.kt?

Otherwise, if you want to keep the types separate, maybe:
`RecentlyVisitedItemInternal`
-> `HistoryHighlight`
-> ` HistorySearchGroup`

gabrielluong:
Was told to wait until tomorrow for my review to see the final state of the changes. Otherwise, I was also informed that we should wait to land on Monday so that nothing breaks on the weekend.

app/src/main/java/org/mozilla/fenix/home/recentvisits/RecentVisitsFeature.kt
```diff
@@ -0,0 +1,258 @@
...
+    @VisibleForTesting
+    internal fun getSortedHistory(
+        historyHighlights: List<FeatureHistoryHighlight>,
+        historyGroups: List<FeatureHistoryGroup>
+    ): List<RecentlyVisitedItem> {
+        return (historyHighlights + historyGroups)
+            .sortedByDescending { it.lastAccessedTime }
+            .map { when(it) {
+                is FeatureHistoryHighlight -> RecentHistoryHighlight(
+                    title = it.historyHighlight.title ?: it.historyHighlight.url,
```
csadilek, line 213: On the title here we should check for nullOrEmpty. Did some manual testing and had some empty entries for title. So, let's fall back to the URL with the title is an empty String.

app/src/main/java/org/mozilla/fenix/home/recentvisits/RecentVisitsFeature.kt
Mugurell: It would've been easier to add this property to the `RecentlyVisitedItem` but it wouldn't been used.
It's only useful in this feature, for a particular scenario - helping to keep the code a bit cleaner while data is filtered, computed, etc.
Went with the rename you proposed. Thank you!

app/src/main/java/org/mozilla/fenix/home/recentvisits/RecentVisitsFeature.kt
Mugurell: Previously this was the case and had a test for it.
Added now also.

app/src/main/java/org/mozilla/fenix/home/recentvisits/RecentVisitsFeature.kt
```diff
@@ -0,0 +1,283 @@
...
+                    .take(highlightCount),
+                historyGroups
+                    .sortedByDescending { it.lastAccessedTime }
+                    .take(groupsCount)
+            )
+        }
+    }
+
+    /**
+     * Perform an in-memory mapping of a history highlight to metadata records to compute it's last access time.
```
csadilek, line 141: nit: its.

app/src/main/java/org/mozilla/fenix/home/recentvisits/RecentVisitsFeature.kt
```diff
@@ -0,0 +1,283 @@
...
+                HistoryGroupInternal(
+                    groupName = it.key,
+                    groupItems = it.value
+                )
+            }
+    }
+
+    /**
+     * Maps the internal highlights and search groups to the final objects to be returned.
+     * Items will be sorted by their last accessed date so that the most recent wil be first.
```
csadilek, line 215: nit: will.

csadilek (approved):
This looks good to me now! I also did some more manual testing and everything worked as expected.

@gabrielluong can you take a look as well esp. at the compose bits and refactorings. Unless you spot a blocker I'd say let's land this to get some bake time in Nightly.

gabrielluong (approved):
Looking back at this entire review, I think I would recommend breaking out some of the commits into their own PR next time. For example, the renaming refactor could've been in a separate PR that landed ahead and outside the feature. One of the final checks I was doing (as I am writing this) was to see if the refactoring preserved the git history since there were also additional refactoring on top of the renaming, which made it appear as a completely new file on the GitHub UI. For context, there were some menu item changes in `RecentlyVisited` that I am aware of that needs to be corrected and losing that history would be a bit of an inconvenience.

app/src/main/java/org/mozilla/fenix/home/recentvisits/controller/RecentVisitsController.kt
```diff
@@ -0,0 +1,151 @@
...
+import org.mozilla.fenix.library.history.toHistoryMetadata
+
+/**
+ * All possible updates following user interactions with the "Recent visits" section from the Home screen.
+ */
+interface RecentVisitsController {
+
+    /**
+     * Callback for when the "Show all" link is clicked.
+     */
```
gabrielluong, line 34: I would revert these to going back to @see RecentVisitsInteractor.onHistoryShowAllClicked in another PR since the content are the same. That way we can just focus on changing it once.

app/src/main/java/org/mozilla/fenix/home/recentvisits/RecentVisitsFeature.kt
csadilek, line 215:
```suggestion
     * Items will be sorted by their last accessed date so that the most recent will be first.
```

app/src/main/java/org/mozilla/fenix/home/recentvisits/RecentVisitsFeature.kt
csadilek, line 141:
```suggestion
     * Perform an in-memory mapping of a history highlight to metadata records to compute its last access time.
```

Issue #22495, opened 2021-11-18
Turn off Inactive Tabs Survey


sarah541:
## User Story
[Jira link](https://mozilla-hub.atlassian.net/browse/FNXV2-18528)

SoftVision-LorandJanos:
Verified as fixed on the latest Nightly (2021-12-07).

The survey is no longer activated.

Device used: Xiaomi Mi11 Lite (Android 11).

Closing the ticket as fixed.

PR #22554, opened 2021-11-23
For #22495 - Disable inactive tabs survey

sarah541:
### Pull Request checklist

Once we've collected enough data, we should turn off the inactive tabs survey, so that users aren't spending time filling out a survey that we don't need them to

Issue : https://github.com/mozilla-mobile/fenix/issues/22495
- [ ] **Tests**: This PR does not include tests since a small UI feature was supposed to be removed
- [ ] **Screenshots**: This PR does not include screenshots since nothing new added

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
+++ b/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
@@ -872,7 +872,7 @@ class Settings(private val appContext: Context) : PreferencesHolder {
      */
     var shouldShowInactiveTabsTurnOffSurvey by booleanPreference(
         appContext.getPreferenceKey(R.string.pref_key_should_show_inactive_tabs_turn_off_survey),
-        default = true
+        default = false
     )
 
     fun getSitePermissionsPhoneFeatureAction(


app/src/main/java/org/mozilla/fenix/settings/TabsSettingsFragment.kt
```diff
@@ -40,9 +40,6 @@ class TabsSettingsFragment : PreferenceFragmentCompat() {
     private lateinit var inactiveTabsCategory: PreferenceCategory
     private lateinit var inactiveTabs: SwitchPreference
     private lateinit var searchTermTabGroups: SwitchPreference
-    private val shouldShowInactiveTabsTurnOffSurvey
-        get() = requireContext().settings().isTelemetryEnabled &&
-            requireContext().settings().shouldShowInactiveTabsTurnOffSurvey
```
gabrielluong, line 45: Rather than removing the code completely. I think we're only being asked to turn off the feature for the time being in the Jira ticket. So, we should be able to just set the default value of `shouldShowInactiveTabsTurnOffSurvey` to `false` for the time being.

gabrielluong (changes requested):

app/src/main/java/org/mozilla/fenix/settings/TabsSettingsFragment.kt
```diff
@@ -40,9 +40,15 @@ class TabsSettingsFragment : PreferenceFragmentCompat() {
     private lateinit var inactiveTabsCategory: PreferenceCategory
     private lateinit var inactiveTabs: SwitchPreference
     private lateinit var searchTermTabGroups: SwitchPreference
+    /**
+     * Turning off inactive tabs survey
+
     private val shouldShowInactiveTabsTurnOffSurvey
         get() = requireContext().settings().isTelemetryEnabled &&
-            requireContext().settings().shouldShowInactiveTabsTurnOffSurvey
+                requireContext().settings().shouldShowInactiveTabsTurnOffSurvey
```
gabrielluong, line 48: I was expecting this change in Settings.kt actually. You can set the default value of the SharedPref to false.

app/src/main/java/org/mozilla/fenix/settings/TabsSettingsFragment.kt
```diff
@@ -40,9 +40,15 @@ class TabsSettingsFragment : PreferenceFragmentCompat() {
...
+    **/
+    private val shouldShowInactiveTabsTurnOffSurvey = false
```
gabrielluong, line 50: I was expecting this change in Settings.kt actually. You can set the default value of the SharedPref to false. Otherwise, now we have an unused SharedPrefs for `shouldShowInactiveTabsTurnOffSurvey`

app/src/main/java/org/mozilla/fenix/settings/TabsSettingsFragment.kt
```diff
@@ -71,16 +77,6 @@ class TabsSettingsFragment : PreferenceFragmentCompat() {
...
-            it.onPreferenceChangeListener = SharedPreferenceUpdater()
-        }
-
-        radioManual = requirePreference(R.string.pref_key_close_tabs_manually)
-        radioOneMonth = requirePreference(R.string.pref_key_close_tabs_after_one_month)
-        radioOneWeek = requirePreference(R.string.pref_key_close_tabs_after_one_week)
-        radioOneDay = requirePreference(R.string.pref_key_close_tabs_after_one_day)
-
-        inactiveTabs = requirePreference<SwitchPreference>(R.string.pref_key_inactive_tabs).also {
-            it.isChecked = requireContext().settings().inactiveTabsAreEnabled
```
gabrielluong, line 83: Trying to understand why this was moved. Doesn't this mean the setOnPreferenceChangeListener will affect the searchTermTabGroups preference now?

gabrielluong:
Product wants this turned off for v96 in January, but no sooner. So, we should also consult what is the right time to land this.

gabrielluong (approved):

app/src/main/java/org/mozilla/fenix/settings/TabsSettingsFragment.kt
sarah541: Fixed

PR #22557, opened 2021-11-23
For #22556 - Update to latest taskgraph revision + path changes

### LINKED ISSUE #22556: Update taskgraph to latest revision
Note: Environment variable `VCS_PATH` has changed to `/builds/worker/checkouts/vcs`

hneiva:


try_task_config.json
```diff
@@ -0,0 +1,8 @@
+{
+    "parameters": {
+        "optimize_target_tasks": true,
+        "target_tasks_method": "nightly",
+        "version": "2021.11.23-nightly-stg"
+    },
+    "version": 2
+}
```
sarah-clements, line 8: Whats this used for?

sarah-clements: Looks like you were using it to trigger nightly tasks? Which means you wouldn't want to check that in :) Maybe Aki or Ahal told you to do it this way, but what I've done for Focus is to push my pr to the staging repo (in your case staging-fenix) and trigger the nightly cron task via the Taskcluster UI so you're not actually triggering a production push-apk task for nightly.

MihaiTabara: We may want to adjust https://github.com/mozilla-mobile/shared-docs/blob/main/android/taskcluster_guide.md#how-do-i-run-a-staging-nightlybetarelease-on-prs to document the other way of triggering staging releases (which is via staging repos). Or maybe we document that in taskgraph docs and point the url to that instead?

sarah-clements: oh, didn't even know about that way of testing :) so this file just needs to be removed then. But yeah, mentioning the other way in the docs is a good idea!

MihaiTabara: Taskgraph is a pre-requisite to allow this type of testing so with your recent work there, we're now able to trigger staging releases easier there too, another nice side-effect from that 👍 

Which reminds me we could improve the docs to reflect this:
* for new repos without taskgraph - forking is probably smoother and easier
* for existing taskgraph repos - triggering staging releases in PRs is faster I believe?

MihaiTabara: > oh, didn't even know about that way of testing :) so this file just needs to be removed then. But yeah, mentioning the other way in the docs is a good idea!

+1, we shouldn't land this.

hneiva: It's for running a staging release (I'll have to remove this file before merging)
More info [here](https://github.com/mozilla-mobile/shared-docs/blob/main/android/taskcluster_guide.md#how-do-i-run-a-staging-nightlybetarelease-on-prs)

hneiva: ^ Ignore my reply. GitHub only pushed to browser the first comment from sarah and not the rest of the conversion.

sarah-clements (approved):
lgtm

hneiva:
Ran `nightly` and `release` staging releases, with similar failing results on tests.

MihaiTabara:
@pocmo just to double-check, do we need a formal review here from your end or we should be good with merging this when we're ready?

pocmo:
> @pocmo just to double-check, do we need a formal review here from your end or we should be good with merging this when we're ready?

No, I think you can do those changes on your own. Change looks good... I hope the path change doesn't cause any problems. Good patch for a Friday? 😅

PR #22568, opened 2021-11-24
Closes #22499: Emit search group disband action on history group removal

### LINKED ISSUE #22499: Deleted search groups are repopulated after interacting with their tabs
Not really sure if this is a bug per-se, but the behaviour is potentially surprising from a user's point of view.

STR:
- perform some searches to form a search group. leave the relevant tab(s) open.
- delete search group using the History UI
- navigate in one of the tabs that was part of the search group
- in the History UI, the search group comes back

grigoryk:
This is needed to prevent removed search groups from coming back after
user interacts with the tab. E.g. if we don't fire this action, upon
interaction with the tab after a corresponding search group was removed
we will end up creating metadat with the deleted 'searchTerms' (they'll
be read from the 'historyMetadata' state on the tab).

At this point we probably want to start encapsulating all of this in a
use case - let's do that separately in
https://github.com/mozilla-mobile/android-components/issues/11309, and
just fix the bug for now.

### Pull Request checklist

1. click on Show All Checks,


csadilek (approved):
Tested to confirm this solves https://github.com/mozilla-mobile/fenix/issues/22499. I noticed that we have a race condition here though. When deleting a group and navigating to home quickly it's possible to still see the deleted group in recently visited (until the next time home is opened). We don't provide any guarantee to prevent this right now. Let's discuss separately as part of https://github.com/mozilla-mobile/android-components/issues/11309 though.

Issue #22580, opened 2021-11-24
Track APK size on perfherder


pocmo:
Let's use perfherder to track our APK size(s) in order to see the changes over time and identify regressions. Using perfherder has the advantage that it will be linked to individual commits / pushes, making it easier to identify the cause.

kbrosnan:
https://treeherder.mozilla.org/perfherder/graphs?highlightAlerts=1&highlightChangelogData=1&highlightCommonAlerts=0&series=fenix,4063376,1,2&series=fenix,4063381,1,2

PR #22581, opened 2021-11-24
Issue #22580: Track APK size in perfherder

pocmo:
If I understand perfherder correctly then all that is needed is logging a specific kind of string in a taskcluster build and then perfherder will pick that up, track values and show beautiful graphs. Let's see if that works. This is now logging for *every* build task. I am not sure how perfherder will untangle that - or if at all. Let's just try. Ideally we only track builds on `main` at first. And then maybe Nightly/Beta/Release builds.

### DIFF
--- a/buildSrc/src/main/java/org/mozilla/fenix/gradle/tasks/ApkSizeTask.kt
+++ b/buildSrc/src/main/java/org/mozilla/fenix/gradle/tasks/ApkSizeTask.kt
@@ -1,0 +1,143 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.gradle.tasks
+
+import org.gradle.api.DefaultTask
+import org.gradle.api.tasks.Input
+import org.gradle.api.tasks.TaskAction
+import org.json.JSONArray
+import org.json.JSONException
+import org.json.JSONObject
+import java.io.File
+import java.io.IOException
+import java.nio.file.Files
+import java.nio.file.Paths
+
+/**
+ * Gradle task for determining the size of APKs and logging them in a perfherder compatible format.
+ */
+open class ApkSizeTask : DefaultTask() {
+    /**
+     * Name of the build variant getting built.
+     */
+    @Input
+    var variantName: String? = null
+
+    /**
+     * List of APKs that get build for the build variant.
+     */
+    @Input
+    var apks: List<String>? = null
+
+    @TaskAction
+    fun logApkSize() {
+        val apkSizes = determineApkSizes()
+        if (apkSizes.isEmpty()) {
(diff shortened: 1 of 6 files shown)


pocmo:
Alright.

Good:
* I see the log in taskcluster.

Bad:
* I don't see anything on perfherder yet. Maybe that will change if this is on `main` and actually running for commits there.
* Invoking the task for every `build_gradle_command` doesn't work in all cases. This also gets called for building the `androidTest` APK and `apkSizeAndroidtest` is not a valid task :) - I may want to figure out how to run this only for specific build tasks on `main`.

pocmo:
This looks good now. I _think_ that should work once it's on `main` and running in tasks that are exposed on treeherder.

buildSrc/src/main/java/org/mozilla/fenix/gradle/tasks/ApkSizeTask.kt
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    private fun determineApkSizes(): Map<String, Long> {
+        val basePath = listOf(
+            "${project.projectDir}", "build", "outputs", "apk", variantName
+        ).joinToString(File.separator)
+
+        return requireNotNull(apks).associateWith { apk ->
+            val path = Paths.get("$basePath${File.separator}$apk")
+            Files.size(path)
```
csadilek, line 47: Do we want to fail the build if we can't determine size for whatever reason (e.g. an `IOException`) or should we show the size as `unknown` then?

buildSrc/src/main/java/org/mozilla/fenix/gradle/tasks/ApkSizeTask.kt
```diff
@@ -0,0 +1,126 @@
...
+     */
+    @Input
+    var apks: List<String>? = null
+
+    @TaskAction
+    fun logApkSize() {
+        val apkSizes = determineApkSizes()
+        val json = buildPerfherderJson(apkSizes)
+
+        println("PERFHERDER_DATA: $json")
```
csadilek, line 37: ^^ Did you mean to keep that in for debugging?

csadilek (approved):
👍 Let's try. Two small Qs below.

buildSrc/src/main/java/org/mozilla/fenix/gradle/tasks/ApkSizeTask.kt
pocmo: Ah, good point. We probably never want to fail the build for this. I'll catch this.

buildSrc/src/main/java/org/mozilla/fenix/gradle/tasks/ApkSizeTask.kt
pocmo: If I understand it correctly then this is how perfherder will pick it up. It looks at the taskcluster logs and grabs those values from it.

csadilek: Ah OK! TIL.

PR #22605, opened 2021-11-26
For #22445 - Improve the scrolling behavior for Recently visited

### LINKED ISSUE #22445: [Bug] Recent searches layout issues
The recent searches layout is horizontally scrollable but when scrolled the items have a different layout and the menu is not properly anchored.

Another issue is that the divider is not showing.

https://user-images.githubusercontent.com/11428869/141793072-dec387ca-d5fe-4320-a377-d4bfa8612f10.mp4

cc @gabrielluong @topotropic

Mugurell:
Fixes #22445

One important issue was that the items are smaller that the width of the
screen and so a bit of the next column would be showing also and user could
interact with this only peaking items and a long press would open a dropdown
menu with no place to anchor it.
To mitigate this:
- the items will now snap to the next position when scrolling so that we'll
always show at least one column in it's entirety.
- menus are enabled only if more than half of the item is visible effectively
disabling clicks on that peaking area of neighbour columns.

cc @topotropic 

https://user-images.githubusercontent.com/11428869/143603392-8c701579-dbfe-4bc8-b52e-3e972209daa2.mp4

&nbsp;

#### This PR contains the changes from the yet unmerged https://github.com/mozilla-mobile/fenix/pull/22535 which also refactored this feature. Will make sure to rebase as needed.

&nbsp;

### Pull Request checklist

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/home/recentvisits/view/RecentlyVisited.kt
```diff
@@ -124,6 +134,7 @@ fun RecentlyVisited(
 @Composable
 private fun RecentlyVisitedHistoryGroup(
     recentVisit: RecentHistoryGroup,
+    enabled: Boolean,
```
gabrielluong, line 137: Can we move this before or after `showDividerLine`, just seems nicer to keep the Boolean arguments together?

app/src/main/java/org/mozilla/fenix/historymetadata/HistoryMetadataFeature.kt
```diff
@@ -4,78 +4,278 @@
...
+    /**
+     * Perform an in-memory mapping of a history highlight to metadata records to compute it's last access time.
+     *
+     * - If a `highlight` cannot be mapped to a corresponding `metadata` record, its lastAccessTime will be set to 0.
+     * - If a `highlight` maps to multiple metadata records, its lastAccessTime will be set to the most recently
+     * updated record.
+     *
+     * @param highlights [HistoryHighlight] list for which to get the last accessed time.
+     * @param metadata [HistoryMetadata] list expected to contain the details for all [highlights].
+     *
```
gabrielluong: nitpick: I think we have more kDoc where we don't separate out the @return with a new line. As a suggestion towards consistency with the codebase, I would defer to opt out of using the new line here.

app/src/main/java/org/mozilla/fenix/compose/Favicon.kt
```diff
@@ -0,0 +1,73 @@
...
+ * Load and display the favicon of a particular website.
+ *
+ * @param url Website [URL] for which the favicon will be shown.
+ * @param size [Dp] height and width of the image to be loaded.
+ */
+@Composable
+fun Favicon(
+    url: String,
+    size: Dp,
+    isPrivate: Boolean = false
```
gabrielluong, line 37: Probably could add a @param for isPrivate as well

app/src/main/java/org/mozilla/fenix/historymetadata/view/RecentlyVisited.kt
```diff
@@ -134,63 +150,98 @@ private fun RecentVisitItem(
...
-                                orientation = Orientation.Vertical,
-                                enabled = false
-                            ).fillMaxHeight()
-                    )
-                }
+        RecentlyVisitedMenu(
+            showMenu = isMenuExpanded,
+            menuItems = menuItems,
+            recentVisit = recentVisit,
+            onLeavingMenu = { isMenuExpanded = false }
```
gabrielluong: Could we just stick with the DropdownMenu naming for this and rename this to `onDismissRequest`

app/src/main/java/org/mozilla/fenix/historymetadata/view/RecentlyVisited.kt
```diff
@@ -213,17 +264,79 @@ private fun RecentlyVisitedCaption(count: Int) {
     )
 }
 
+@Composable
+private fun RecentlyVisitedMenu(
```
gabrielluong: nit: Add kDoc

app/src/main/java/org/mozilla/fenix/home/recentvisits/controller/RecentVisitsController.kt
```diff
@@ -16,66 +16,81 @@ import mozilla.components.feature.tabs.TabsUseCases.SelectOrAddUseCase
...
+     */
     override fun handleHistoryShowAllClicked() {
         dismissSearchDialogIfDisplayed()
         navController.navigate(
             HomeFragmentDirections.actionGlobalHistoryFragment()
         )
     }
 
+    /**
+     * Navigates to the history metadata group fragment to display the group.
```
gabrielluong, line 90: Hm.. I was expecting we used this description in the controller interface above instead of having 2 different descriptions between the Interface and the Default implementation when I mentioned replacing the @see, but as is since the Interface still basically uses the description from the Interactor, there wasn't a point in replacing the @see since it is the same kDoc.

This isn't a blocker by any means.

app/src/main/java/org/mozilla/fenix/home/recentvisits/view/RecentlyVisited.kt
Mugurell: Sure!

app/src/main/java/org/mozilla/fenix/historymetadata/HistoryMetadataFeature.kt
Mugurell: We still have many `@return`s with newlines before and I'd say this has a real advantage - helps with legibility, the same as for `@param`s newlines, one will immediately see these 3 sections in a bigger comment block.

app/src/main/java/org/mozilla/fenix/compose/Favicon.kt
Mugurell: Updated in https://github.com/mozilla-mobile/fenix/pull/22535.

app/src/main/java/org/mozilla/fenix/historymetadata/view/RecentlyVisited.kt
Mugurell: Updated in https://github.com/mozilla-mobile/fenix/pull/22535.

app/src/main/java/org/mozilla/fenix/historymetadata/view/RecentlyVisited.kt
Mugurell: Updated in https://github.com/mozilla-mobile/fenix/pull/22535.

app/src/main/java/org/mozilla/fenix/home/recentvisits/controller/RecentVisitsController.kt
Mugurell: To keep the interface abstract it should not dictate behavior.
I see `Navigates to the history metadata group fragment` as an implementation detail.
If this description was set for interface's method declaration then all implementations should behave that way.

app/src/main/java/org/mozilla/fenix/home/recentvisits/view/RecentlyVisited.kt
```diff
@@ -153,13 +164,15 @@ private fun RecentlyVisitedHistoryGroup(
...
+                modifier = Modifier
+                    .padding(top = 7.dp, bottom = 2.dp)
+                    .weight(1f)
             )
 
-            RecentlyVisitedCaption(recentVisit.historyMetadata.size)
+            RecentlyVisitedCaption(recentVisit.historyMetadata.size, Modifier.weight(1f))
 
             if (showDividerLine) {
-                RecentlyVisitedDivider(modifier = Modifier.padding(top = 9.dp))
```
gabrielluong, line 162: Am I correct to assume that the weight fills in the 'remaining' space so that the Row is still 56dp with the divider being at the bottom?

app/src/main/java/org/mozilla/fenix/home/recentvisits/view/RecentlyVisited.kt
```diff
@@ -153,13 +164,15 @@ private fun RecentlyVisitedHistoryGroup(
...
             RecentlyVisitedTitle(
                 text = recentVisit.title,
-                modifier = Modifier.padding(top = 7.dp, bottom = 2.dp)
+                modifier = Modifier
+                    .padding(top = 7.dp, bottom = 2.dp)
+                    .weight(1f)
             )
 
-            RecentlyVisitedCaption(recentVisit.historyMetadata.size)
+            RecentlyVisitedCaption(recentVisit.historyMetadata.size, Modifier.weight(1f))
```
gabrielluong, line 172: Let's add the named parameters here since we do that in most places. It makes it easier to review in GitHub since you don't have the benefits of Android Studio showing you the named parameter hints.

app/src/main/java/org/mozilla/fenix/home/recentvisits/view/RecentlyVisited.kt
```diff
@@ -126,6 +135,7 @@ fun RecentlyVisited(
 private fun RecentlyVisitedHistoryGroup(
     recentVisit: RecentHistoryGroup,
     menuItems: List<RecentVisitMenuItem>,
+    enabled: Boolean,
```
gabrielluong, line 138: I don't have a good name for this, but I think we should rename this to something like `clickableEnabled` to not confuse that there is a disabled ui state.

app/src/main/java/org/mozilla/fenix/home/recentvisits/view/RecentlyVisited.kt
Mugurell: Yes.
Based on [the documentation](https://developer.android.com/reference/kotlin/androidx/compose/foundation/layout/ColumnScope?hl=pt#(androidx.compose.ui.Modifier).weight(kotlin.Float,kotlin.Boolean)) using weights will mean the following succession of events:
- first the unweighted elements are measured - in our case the divider will have the exact height that it needs
- then the remaining space (column height minus unweighted items height) is divided between weighted elements based on their weight - in our case the title and caption having the same weight will each have half of the remaining height
- then the items are laid out based on their measurements - in our case all items have perfect measurements, the title and caption will be placed on screen (with a bit bigger height than needed but the default Top/Start alignment plus the padding set for title gets the placement we want) then the divider as the last item in the column is guaranteed to be shown at the bottom and occupy the exact space it needs.

This approach helps ensuring all items will have the space they need on screen and more importantly by measuring the title and caption after the divider helps ensure the divider will always be shown on screen, at the bottom, while I think previously there were cases in which it would be pushed to the outside of the column.

gabrielluong (approved):

PR #22625, opened 2021-11-29
Prevent flash when navigating to home

csadilek:
We're already using this workaround whenever navigating away from the browser fragment via the `BrowserToolbarMenuController` (e.g. three dot menu to settings), but we didn't need it in the `BrowserToolbarController` until we added the new home button (and home behind search.) 

In my testing this makes the transition as lot smoother, preventing the flash when navigating from browser to home.


gabrielluong (approved):
Is there a way to add even more documentation to `captureEngineViewAndDrawStatically` about when to utilize it? I had looked at this and always wonder if (whether or not we should remove it because it's unclear what it does) and how we should be using it ~~since the existing usage in the toolbar navigates to the browser fragment, whereas your current usage navigates to the home fragment.~~

Edit: totally read the code wrong. It is navigating FROM the browser fragment.

I guess another question is how does this prevent the flash since looking at the function, it's just changing the swipe through animation. Unless you're talking about a totally black background during the transition (which I am imagining), just hard to see on my Pixel.

csadilek:
@gabrielluong 

> Unless you're talking about a totally black background during the transition (which I am imagining), just hard to see on my Pixel.

Yes, that one seems much better for me with this, but I am wondering if this is just a timing behaviour. Let's test this on a few devices first, just wanted to put it up to start the discussion.

Amejia481 (approved):
Tested with @csadilek and we noticed the flash disappeared after applying this patch, below the recordings, we tested on a Moto g5.


https://user-images.githubusercontent.com/773158/143929924-3cc83770-80f1-4a15-baa4-73725f6fed70.mp4


https://user-images.githubusercontent.com/773158/143929939-7be41921-d663-4a92-bfad-41af161680f7.mp4

csadilek:
Alright last commit fixed one of the two failing tests. There's still a problem with `selectSearchEnginesShortcutTest` that needs investigation.

csadilek:
🍏

pocmo (approved):

csadilek:
OK will run tests on more time here. The other intermittent I saw was https://github.com/mozilla-mobile/fenix/issues/22640 which already happens on `main` before this change, so it's unrelated.

csadilek:
OK, 5 successful UI test runs in a row. I am happy here.

csadilek:
We don't have any Nightly time left, so let's land this on Tuesday after we cut the branch.

Amejia481:
👍🏽

Issue #18160, opened 2021-02-24
Show a prompt when trying to leave private browsing with active downloads


Amejia481:
In Firefox desktop when users are in private browsing and try to leave it (closing all private tabs) and have some downloads in progress, a prompt is shown to users indicating that if they leave the downloads will be canceled. 

![image](https://user-images.githubusercontent.com/773158/109063132-cd92af80-76b6-11eb-8b2b-f2b5c69dad88.png)

Amejia481:
On the other hand Chrome does something similar, but instead of showing a prompt when when leaving private browsing it just cancel any pending downloads associated.

mavduevskiy:
The changes could be seen in the latest Nightly build.

SoftVision-LorandJanos:
Verified as fixed in the latest Nightly (2021-12-27).
The prompt is displayed when trying to close the private tabs.
The buttons "Stay in private browsing" and "Cancel 1 download" work as intended.
The prompt has a bit different design than the one presented in the description:
![Screenshot_20220108-092918](https://user-images.githubusercontent.com/48829350/147448134-6f3e32a4-0984-46a1-8674-ee7a0ba89487.png)
Device used: Google Pixel 4 (Android 11).
Closing the ticket as fixed.

PR #22638, opened 2021-11-30
For #18160: Show a prompt when trying to leave private browsing with active downloads

mavduevskiy:
Relates to [...fenix/issue/18160](https://github.com/mozilla-mobile/fenix/issues/18160)

It appears that we have three scenarios when all tabs are being closing by the user.
1) manually closing the tabs individually
2) closing all tabs from drop down menu
3) taping the private notification with action "close all"

For the first two options I have added a simple DialogFragment, that requests user to confirm that she or he is okay with canceling the active downloads. The third scenario is a bit different, it might be triggered while app is in background mode, so I didn't tackle that case.

I also moved analytics from interactor layer to controller, since not every request to close the tab actually results with the tab being closed. So then controller, after actually closing the tab, will fire an analytics event.

I am not sure if I should write UI tests for it.

image

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/tabstray/CancelDownloadFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/tabstray/CancelDownloadFragment.kt
@@ -1,0 +1,104 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.tabstray
+
+import android.app.Dialog
+import android.os.Bundle
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import android.widget.FrameLayout
+import androidx.appcompat.app.AppCompatDialogFragment
+import androidx.fragment.app.setFragmentResult
+import androidx.navigation.fragment.findNavController
+import androidx.navigation.fragment.navArgs
+import com.google.android.material.bottomsheet.BottomSheetBehavior
+import com.google.android.material.bottomsheet.BottomSheetDialog
+import org.mozilla.fenix.R
+import org.mozilla.fenix.databinding.FragmentCancelDownloadBinding
+import org.mozilla.fenix.ext.runIfFragmentIsAttached
+
+class CancelDownloadFragment : AppCompatDialogFragment() {
+
+    private var _binding: FragmentCancelDownloadBinding? = null
+    private val binding get() = _binding!!
+    private val args by navArgs<CancelDownloadFragmentArgs>()
+
+    override fun onCreateDialog(savedInstanceState: Bundle?): Dialog =
+        BottomSheetDialog(requireContext(), this.theme).apply {
+            setOnShowListener {
+                val bottomSheet =
+                    findViewById<View>(com.google.android.material.R.id.design_bottom_sheet) as FrameLayout
+                val behavior = BottomSheetBehavior.from(bottomSheet)
+                behavior.state = BottomSheetBehavior.STATE_EXPANDED
+            }
+        }
+
+    override fun onCreateView(
+        inflater: LayoutInflater,
+        container: ViewGroup?,
+        savedInstanceState: Bundle?
+    ): View {
+        _binding = FragmentCancelDownloadBinding.inflate(inflater, container, false)
+
+        binding.cancelDownloadMessage.text = String.format(
+            binding.root.context.getString(
+                R.string.cancel_active_download_warning_content_description
+            ),
+            args.downloadCount
+        )
+        return binding.root
+    }
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+        binding.cancelDownloadAccept.setOnClickListener {
+            setFragmentResult(
+                RESULT_KEY,
+                Bundle().apply {
+                    putBoolean(ACCEPT_KEY, true)
+                    args.tabId?.let { putString(TAB_KEY, it) }
+                    args.source?.let { putString(SOURCE_KEY, it) }
+                    putBoolean(PRIVATE_KEY, args.private)
+                }
+            )
+            runIfFragmentIsAttached {
+                if (this.isVisible) {
+                    dismiss()
+                }
+                findNavController().popBackStack()
+            }
+        }
+
+        binding.cancelDownloadDeny.setOnClickListener {
+            dismiss()
+        }
+    }
+
+    override fun onDestroyView() {
+        super.onDestroyView()
+        _binding = null
+    }
+
+    companion object {
+        const val RESULT_KEY = "cancelDownloadFragmentResultKey"
+        const val ACCEPT_KEY = "cancelDownloadFragmentAcceptKey"
+        const val TAB_KEY = "cancelDownloadFragmentTabKey"
+        const val SOURCE_KEY = "cancelDownloadFragmentSourceKey"
+        const val PRIVATE_KEY = "cancelDownloadFragmentPrivateKey"
+    }
+}
+
+data class CancelDownloadFragmentArguments(
+    val downloadCount: Int,
+    val trigger: Trigger,
+    val tabId: String? = null,
+    val source: String? = null,
+    val private: Boolean = true
+)
+
+enum class Trigger {
+    SINGLE_TAB, CLOSE_ALL
+}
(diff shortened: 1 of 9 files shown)


app/src/main/java/org/mozilla/fenix/tabstray/CancelDownloadFragment.kt
```diff
@@ -0,0 +1,104 @@
...
+        savedInstanceState: Bundle?
+    ): View {
+        _binding = FragmentCancelDownloadBinding.inflate(inflater, container, false)
+
+        binding.cancelDownloadMessage.text = String.format(
+            binding.root.context.getString(
+                R.string.cancel_active_download_warning_content_description
+            ),
+            args.downloadCount
+        )
```
gabrielluong, line 51: I am curious how come we don't do this in `onViewCreated`?

gabrielluong:
Not to block this in anyway, but since UX and Product might not be aware of this, can we make sure we demo it to them and make sure they are aware and okay with things as-is?

app/src/main/java/org/mozilla/fenix/tabstray/CancelDownloadFragment.kt
```diff
@@ -0,0 +1,104 @@
...
+    companion object {
+        const val RESULT_KEY = "cancelDownloadFragmentResultKey"
+        const val ACCEPT_KEY = "cancelDownloadFragmentAcceptKey"
+        const val TAB_KEY = "cancelDownloadFragmentTabKey"
+        const val SOURCE_KEY = "cancelDownloadFragmentSourceKey"
+        const val PRIVATE_KEY = "cancelDownloadFragmentPrivateKey"
+    }
+}
+
+data class CancelDownloadFragmentArguments(
```
gabrielluong, line 94: I would recommend adding some kDoc for this data class and `Trigger`

gabrielluong: Would also consider renaming this so we don't have the `FragmentArguments` if it makes sense

app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayController.kt
```diff
@@ -94,7 +102,8 @@ class DefaultTabsTrayController(
     private val tabsUseCases: TabsUseCases,
     private val selectTabPosition: (Int, Boolean) -> Unit,
     private val dismissTray: () -> Unit,
-    private val showUndoSnackbarForTab: (Boolean) -> Unit
+    private val showUndoSnackbarForTab: (Boolean) -> Unit,
+    private val showCancelledDownloadWarning: (arg: CancelDownloadFragmentArguments) -> Unit,
```
gabrielluong, line 106: I don't think we necessarily need this `showCancelledDownloadWarning` callback function since we have the `navController` available as a property that we can just use directly in the controlleer.

app/src/main/res/navigation/nav_graph.xml
```diff
@@ -1163,4 +1192,4 @@
                 app:popExitAnim="@anim/slide_out_right" />
         </fragment>
     </navigation>
-</navigation>
+</navigation>
```
gabrielluong, line 1195: nit: please add the newline at eof.

app/src/main/res/values/strings.xml
```diff
@@ -2026,4 +2032,4 @@
     <string name="pocket_stories_feature_caption">Part of the Firefox family. %s</string>
     <!-- Clickable text for opening an external link for more information about Pocket. -->
     <string name="pocket_stories_feature_learn_more">Learn more</string>
-</resources>
+</resources>
```
gabrielluong, line 2035: same as above

app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayFragment.kt
```diff
@@ -370,6 +382,18 @@ class TabsTrayFragment : AppCompatDialogFragment() {
         }
     }
 
+    @VisibleForTesting
+    internal fun showCancelledDownloadWarning(arg: CancelDownloadFragmentArguments) {
```
gabrielluong, line 386: After looking at this some more, it's kinda weird to wrap all the arguments into a data class and then unbundle it here. I would consider just dropping `CancelDownloadFragmentArguments`. I assume you were maybe doing this because callback functions that are being passed from here into the controller.

app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayController.kt
```diff
@@ -134,18 +143,37 @@ class DefaultTabsTrayController(
...
+                            Trigger.SINGLE_TAB,
+                            tabId = tab.id
+                        )
+                    )
+                    return
+                } else {
+                    dismissTabsTrayAndNavigateHome(tabId)
+                }
             }
+            metrics.track(Event.ClosedExistingTab(source ?: "unknown"))
```
gabrielluong, line 176: I am not quite sure what this metrics does, but it made me go - hm, the reviewer might want to take a closer look.

app/src/main/java/org/mozilla/fenix/tabstray/NavigationInteractor.kt
```diff
@@ -60,8 +60,10 @@ interface NavigationInteractor {
 
     /**
      * Called when clicking the close all tabs button.
+     * @param private indicates if tabs are private
+     * @param cancelPrivateDownloadsAccepted user confirmed private downloads cancellation
```
gabrielluong, line 64: `shouldCancelPrivateDownloads` might roll out of the tongue easier?

Apologizes for the run-by comments.

gabrielluong:
```suggestion
     * @param cancelPrivateDownloadsAccepted Whether or not downloads in a private session should be cancelled.
```
For kDocs, I think we usually capitalize and try to not infer too much about what the user did.

app/src/main/java/org/mozilla/fenix/tabstray/NavigationInteractor.kt
```diff
@@ -60,8 +60,10 @@ interface NavigationInteractor {
 
     /**
      * Called when clicking the close all tabs button.
+     * @param private indicates if tabs are private
```
gabrielluong, line 63:
```suggestion
     * @param private Whether or not the session is in private mode.
```

Amejia481 (changes requested):
Awesome job :)
I think this behavior is something that we would like to have in other browsers too, I believe we should try to abstract this functionality on AC for other browsers to benefit from it.

One approach that we could try it's on [DownloadMiddleware](https://github.com/mozilla-mobile/android-components/blob/main/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadMiddleware.kt) observe actions for closing tabs `RemoveAllTabsAction`, `RemoveAllPrivateTabsAction`, ... `RemoveTabsAction` and perform the validation for showing the dialog when we only have one tab and a download it's on going dispatch and action to indicate the we want to show the dialog and stop the initial action from performing by **returning** instead of calling `next(action)`, see an example [here](https://github.com/mozilla-mobile/android-components/blob/0e963caf28a900dae6bc5a408d8cc3d12acb863a/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadMiddleware.kt#L67). 

For showing the dialog, add a new state on `BrowserState` similar to [downloads](https://github.com/mozilla-mobile/android-components/blob/0e963caf28a900dae6bc5a408d8cc3d12acb863a/components/browser/state/src/main/java/mozilla/components/browser/state/state/BrowserState.kt#L44) something like `closeOnGoingDonwloadRequest`  see the download request property on [ContentState](https://github.com/mozilla-mobile/android-components/blob/0e963caf28a900dae6bc5a408d8cc3d12acb863a/components/browser/state/src/main/java/mozilla/components/browser/state/state/ContentState.kt#L76) , then observe changes on this property on same as [we do for showing download dialogs](https://github.com/mozilla-mobile/android-components/blob/main/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadsFeature.kt#L124-L130) .

Let me know what do you think :)

Amejia481:
> Not to block this in anyway, but since UX and Product might not be aware of this, can we make sure we demo it to them and make sure they are aware and okay with things as-is?

Thanks for the head up :)
Product added this ticket to our backlog in [Jira](https://mozilla-hub.atlassian.net/browse/FNXV2-18595). I chatted with @mavduevskiy to move this work to AC and we will ask for UI/UX feedback :)

PR #22647, opened 2021-12-01
Add webrender flag for the  extra option.

gmierz:
This patch adds a no-op --enable-webrender flag to get the `webrender` extra option added to the perfherder data.


jmaher (approved):
thanks!

mcomella:
@gmierz @jmaher Are either of you able to add the `pr:needs-landing` label? If not, let amedyne know and they should be able to add you.

gmierz:
Yes I can add it. Thanks for the prompt @mcomella!

Issue #22516, opened 2021-11-19
Homepage section header polish


MozillaNoah:
https://mozilla-hub.atlassian.net/browse/FNXV2-18100

**Acceptance criteria**
- Confirm "Jump back in", "Recently bookmarked", "Recent searches" and "Thought-provoking stories" are 16sp 
- Confirm "Show all" is 14sp, color is #592ACB in light mode
- Align the "Jump back in", "Recent searches" , and "recent bookmarks" section titles with their respective "show all" buttons 
- ~~AC IN PROGRESS:  potentially update some header-to-content spacing~~ The spacing is acceptable for now

PR #22650, opened 2021-12-01
For #22516 - Fixed alignment of "Show all" buttons on the homescreen

MozillaNoah:
#22516 

This was mainly an issue of trying to align two pieces of text that have different font weight and size. I ultimately decided to turn the "Show All" elements into TextViews to remove any baked-in padding from the Button object, since this was the only possible solution to closely align the bottom line of two differently-styled TextViews (at least in XML). The side effect of this was the tappable area of the "Show all" buttons did decrease slightly, so this should only be a short term solution until these sections are rewritten into Compose. 

### Pull Request checklist

1. click on Show All Checks,

Screenshot_1638380951


MozillaNoah:
CC: @topotropic

gabrielluong (approved):

gabrielluong:
You might want to consider a rebase approach for your workflow, and amending/squashing commits (looking at the content description commit). I am mostly looking at the `Merge branch 'main' into homepage-section-header-polish` commits and wondering what happens if we didn't squash the PRs when we land. 

Also, we probably should update the title of the PR since squash landing takes the title of the PR inside of using the commit messages. Probably another reason to amend/squash commits is so we don't end up with messy commit description which would include that content description fix.

gabrielluong:
> You might want to consider a rebase approach for your workflow, and amending/squashing commits (looking at the content description commit). I am mostly looking at the `Merge branch 'main' into homepage-section-header-polish` commits and wondering what happens if we didn't squash the PRs when we land. I imagine we don't want "Merge" commits landing in the tree.
> 
> Also, we probably should update the title of the PR to include the issue number and align it with your commit message since squash landing takes the title of the PR inside of using the commit messages. Probably another reason to amend/squash commits is so we don't end up with messy commit description which would include that content description fix.

Just adding a do-not-land until my last comment was seen.

MozillaNoah:
> > You might want to consider a rebase approach for your workflow, and amending/squashing commits (looking at the content description commit). I am mostly looking at the `Merge branch 'main' into homepage-section-header-polish` commits and wondering what happens if we didn't squash the PRs when we land. I imagine we don't want "Merge" commits landing in the tree.
> > Also, we probably should update the title of the PR to include the issue number and align it with your commit message since squash landing takes the title of the PR inside of using the commit messages. Probably another reason to amend/squash commits is so we don't end up with messy commit description which would include that content description fix.
> 
> Just adding a do-not-land until my last comment was seen. Oh, we also don't necessarily have to add "Feature" labels in the PRs since that is more for the Issues.

Oh gotcha, that's good to know

MozillaNoah:
> You might want to consider a rebase approach for your workflow, and amending/squashing commits (looking at the content description commit). I am mostly looking at the `Merge branch 'main' into homepage-section-header-polish` commits and wondering what happens if we didn't squash the PRs when we land. I imagine we don't want "Merge" commits landing in the tree.
> 
> Also, we probably should update the title of the PR to include the issue number and align it with your commit message since squash landing takes the title of the PR inside of using the commit messages. Probably another reason to amend/squash commits is so we don't end up with messy commit description which would include that content description fix.

Yeah, totally. This makes sense. I admittedly use the Github desktop client to manage everything since my terminal skills are very very rusty, so I'll poke around that first and reach out for a git refresher.

MozillaNoah:
As an update, I looked into the commit history, and it looks like the squash does in fact omit the "merge from main" commits. Either way, I'll try to keep my commit feeds a bit tidier in the future.
Screen Shot 2021-12-01 at 2 05 10 PM

gabrielluong:
> As an update, I looked into the commit history, and it looks like the landing squash does in fact omit the "merge from main" commits. Either way, I'll try to keep my commit feeds a bit tidier in the future. Screen Shot 2021-12-01 at 2 05 10 PM

Cool, I am removing the do-not-land since I think it's good enough that we have a reference to the issue number in our title now, which will be used when we do a squash land through mergify. I think you can see in the screenshot that the PR title (which does not reference the issue number) was used by mergify when you landed the MR2 telemetry.

I will point that the squashed commits do end up listed in the commit description, which can end up being quite messy if w e were looking at the git blame through Android Studio and the dialog takes up the whole window. So, it does pay off to amend or squash your commits manually before hand. Alternatively, I will also advocate for landing smaller commits with a good reasonable commit message that focuses on a particular change (such as renaming, refactor, formatting). Sometimes, you see a PR with a lot of refactoring, renaming plus the actual feature implementation, which makes it harder to review altogether because the changes does add up. It's also not as nice squashed together if we chose to do a squash landing since you will sometimes find yourself looking at older commits to see how a thing was implement and seeing all the refactoring vs seeing how the thing was actually implemented adds to the noise. All in all, I am basically trying to say that having many small commits that are not squashed is a good practise, but that requires some planning as you code and commit changes.

topotropic:
Looking good! Thanks!

PR #22670, opened 2021-12-02
Closes #22550: Refactor search term tab grouping logic to rely on middleware

### LINKED ISSUE #22550: Refactor search term tab grouping logic to rely on middleware
In https://github.com/mozilla-mobile/android-components/pull/11272 we have added tab partitions/groups to `BrowserState` which will make it easier to represent the state of tab groups, as opposed to inferring those groups on demand.

We will need to add a `middleware` to Fenix that manages the corresponding tab groups.  A draft of what this would look like can be found in the [RFC](https://github.com/mozilla-mobile/android-components/blob/main/docs/rfcs/0008-tab-groups.md#automatic-grouping-with-middlewares).

The `middleware` will have to react to:
- `SetHistoryMetadataKeyAction` to add a tab to the group or remove it in case the search terms got cleared
- `DisbandSearchGroupAction` to remove the entire group
- `RestoreCompleteAction` to update the groups based on all restored tabs after startup
- `RestoreTabAction` to put a tab back in its group e.g. when it was deleted and undo was triggered

csadilek:
This requires https://github.com/mozilla-mobile/android-components/pull/11384.

We will work through #22550 here to refactor search term tab grouping to use the new tab partitions / groups available in A-C.

### Pull Request checklist

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/ext/BrowserState.kt
```diff
@@ -41,7 +42,7 @@ fun BrowserState.asRecentTabs(): List<RecentTab> {
         } else {
             listOf(selectedNormalTab)
                 .plus(normalTabs.sortedByDescending { it.lastAccess })
-                .minus(lastTabGroup?.tabs ?: emptyList())
+                .filterNot { lastTabGroup?.tabIds?.contains(it?.id) ?: true }
```
csadilek, line 45: You want to remove the tabs that are in the last group here. So with `filterNot` you'd need to `?: false`, default to false, no?

app/src/main/java/org/mozilla/fenix/ext/BrowserState.kt
```diff
@@ -80,63 +81,27 @@ val BrowserState.inProgressMediaTab: TabSessionState?
...
-            tabGroup.tabs.count()
+            tabGroup.tabIds.size
         )
     }
 
-/**
- * Returns a pair containing a list of search term groups sorted by last access time, and "remainder" tabs that have
- * search terms but should not be in groups (because the group is of size one).
- */
-fun List<TabSessionState>.toSearchGroup(
```
csadilek, line 106: 🎉 💪 ♻️

app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayStore.kt
```diff
@@ -19,7 +19,7 @@ import org.mozilla.fenix.tabstray.browser.TabGroup
  * @property mode Whether the browser tab list is in multi-select mode or not with the set of
  * currently selected tabs.
  * @property inactiveTabs The list of tabs are considered inactive.
- * @property searchTermGroups The list of tab groups.
+ * @property tabPartitions The map of tab partitions.
```
csadilek, line 22: The docs aren't right, you're not passing a map here, but a single partition? Shouldn't it be a map?

app/src/main/java/org/mozilla/fenix/tabstray/browser/OtherHeaderBinding.kt
```diff
@@ -19,9 +19,13 @@ class OtherHeaderBinding(
     private val showHeader: (Boolean) -> Unit
 ) : AbstractBinding<TabsTrayState>(store) {
     override suspend fun onState(flow: Flow<TabsTrayState>) {
-        flow.ifAnyChanged { arrayOf(it.normalTabs, it.searchTermGroups) }
+        flow.ifChanged { Pair(it.normalTabs, it.tabPartition) }
```
csadilek, line 22: See comment above. I am confused why we have a single partition here. Is it because we're just starting with the search one, and add inactive later?

csadilek: But even then, it would have to have a different name.

app/src/main/java/org/mozilla/fenix/tabstray/browser/OtherHeaderBinding.kt
```diff
@@ -19,9 +19,13 @@ class OtherHeaderBinding(
...
             .collect {
-                if (it.searchTermGroups.isNotEmpty() && it.normalTabs.isNotEmpty()) {
+                if (
+                    it.normalTabs.isNotEmpty() &&
+                    !it.tabPartition?.tabGroups?.filter { tabGroup -> tabGroup.tabIds.isNotEmpty() }
+                        .isNullOrEmpty()
```
csadilek, line 27: nit: ^^ that would be a nice extension function so we can write `tabPartition.isNotEmpty()`. Let's add that.

csadilek (commented):
Looks great, and we're making it simpler overall. I found two things below to improve/discuss.

app/src/main/java/org/mozilla/fenix/ext/BrowserState.kt
rocketsroger: Good catch.  yes should be `?: false` I will update

app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayStore.kt
rocketsroger: Correct, I'll correct the doc.  Thanks

app/src/main/java/org/mozilla/fenix/tabstray/browser/OtherHeaderBinding.kt
rocketsroger: Yes, we're only using one `TabPartition` for search term.  We will update once we add inactive later.  I will rename this to `searchTermPartition`.  Thanks

app/src/main/java/org/mozilla/fenix/tabstray/browser/OtherHeaderBinding.kt
rocketsroger: Sounds good.  I'll add the extension functions. 👍

app/src/main/java/org/mozilla/fenix/ext/TabPartition.kt
```diff
@@ -0,0 +1,26 @@
...
+package org.mozilla.fenix.ext
+
+import mozilla.components.browser.state.state.TabPartition
+
+/**
+ * Check if a [TabPartition] has no tabs
+ *
+ * @return true if the [TabPartition] has no tabs, false otherwise.
+ */
+fun TabPartition?.isEmpty(): Boolean {
```
csadilek, line 14: Let's move these to A-C as well and add to https://github.com/mozilla-mobile/android-components/pull/11384? They are generic and will be useful for any consumer.

csadilek (commented):
This looks great now. Let's do a round of manual testing and land after the break!

app/src/main/java/org/mozilla/fenix/ext/TabPartition.kt
rocketsroger: Done.  Thanks

rocketsroger (approved):

app/src/main/java/org/mozilla/fenix/components/Core.kt
```diff
@@ -210,7 +211,8 @@ class Core(
                 PromptMiddleware(),
                 AdsTelemetryMiddleware(adsTelemetry),
                 LastMediaAccessMiddleware(),
-                HistoryMetadataMiddleware(historyMetadataService)
+                HistoryMetadataMiddleware(historyMetadataService),
+                SearchTermTabGroupMiddleware()
```
MatthewTighe, line 215: Bit of a nitpicky drive-by, but wanted to call out that we could avoid diffs like this in the future by using [trailing commas](https://kotlinlang.org/docs/coding-conventions.html#trailing-commas) everywhere, now that they're available. I've been trying to get in the habit, personally. Just something to keep in mind for the future, no action needed here

rocketsroger: I'm not sure.  Are we doing trailing commas?  I've been making sure to remove them when I see them.  I lean towards no trailing commas for readability.  But I think we should decide if this is something we want to do in the future.

Mugurell: Drive by comment: I'm for standardizing trailing commas to keep the diffs just a bit cleaner and the history easier to follow since in many cases it acts as documentation for the feature.

rocketsroger: @Mugurell Yes, we've got an agreement as a team we'll now add the trailing commas in all future changes.  Thanks!

Issue #21894, opened 2021-10-13
Move Tabs Tray to compose:  Individual tab viewholders: GridViewHolder, ListViewHolder.


amedyne:
From: #21318

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-18293)

MozillaNoah:
@Amejia481, I noticed this issue wasn't flagged and auto closed with your PR the other day. Should this be closed or did it need to be formally QA'd?

Amejia481:
Yes, I think we could close it. As at the moment it's [behind a feature flag that it's only active  on debug builds](https://github.com/mozilla-mobile/fenix/blob/bf522dbd508d9a89476cd8a3293b68cd90340ff3/app/src/main/java/org/mozilla/fenix/FeatureFlags.kt#L104). Ideally we would like to active it on nightly for people to provide feedback, but not sure if we should as we we don't have support for tabs re-ordering in compose yet.

PR #22672, opened 2021-12-02
For #21894: Move Tabs Tray to compose: Individual tab viewholders: ListViewHolder.

Amejia481:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -466,6 +466,7 @@ dependencies {
     implementation Deps.androidx_appcompat
     implementation Deps.androidx_constraintlayout
     implementation Deps.androidx_coordinatorlayout
+    implementation Deps.google_accompanist_drawablepainter
 
     implementation Deps.sentry
 
--- a/app/src/main/java/org/mozilla/fenix/compose/ThumbnailCard.kt
+++ b/app/src/main/java/org/mozilla/fenix/compose/ThumbnailCard.kt
@@ -32,6 +32,7 @@ import mozilla.components.concept.base.images.ImageLoadRequest
 import org.mozilla.fenix.R
 import org.mozilla.fenix.components.components
 import org.mozilla.fenix.theme.FirefoxTheme
+import org.mozilla.fenix.theme.Theme
 
 /**
  * Card which will display a thumbnail. If a thumbnail is not available for [url], the favicon
@@ -40,6 +41,8 @@ import org.mozilla.fenix.theme.FirefoxTheme
  * @param url Url to display thumbnail for.
  * @param key Key used to remember the thumbnail for future compositions.
  * @param modifier [Modifier] used to draw the image content.
+ * @param contentDescription Text used by accessibility services
+ * to describe what this image represents.
  * @param contentScale [ContentScale] used to draw image content.
  * @param alignment [Alignment] used to draw the image content.
  */
@@ -48,6 +51,7 @@ fun ThumbnailCard(
     url: String,
     key: String,
     modifier: Modifier = Modifier,
+    contentDescription: String? = null,
     contentScale: ContentScale = ContentScale.FillWidth,
     alignment: Alignment = Alignment.TopCenter
 ) {
@@ -55,36 +59,42 @@ fun ThumbnailCard(
         modifier = modifier,
         backgroundColor = colorResource(id = R.color.photonGrey20)
     ) {
-        components.core.icons.Loader(url) {
-            Placeholder {
-                Box(
-                    modifier = Modifier.background(color = FirefoxTheme.colors.layer3)
-                )
-            }
-
-            WithIcon { icon ->
-                Box(
-                    modifier = Modifier.size(36.dp),
-                    contentAlignment = Alignment.Center
-                ) {
-                    Image(
-                        painter = icon.painter,
-                        contentDescription = null,
-                        modifier = Modifier
-                            .size(36.dp)
-                            .clip(RoundedCornerShape(8.dp)),
-                        contentScale = ContentScale.Fit
+        if (inComposePreview) {
+            Box(
+                modifier = Modifier.background(color = FirefoxTheme.colors.layer3)
+            )
+        } else {
+            components.core.icons.Loader(url) {
+                Placeholder {
+                    Box(
+                        modifier = Modifier.background(color = FirefoxTheme.colors.layer3)
                     )
                 }
-            }
-        }
 
-        ThumbnailImage(
-            key = key,
-            modifier = modifier,
-            contentScale = contentScale,
-            alignment = alignment
-        )
+                WithIcon { icon ->
+                    Box(
+                        modifier = Modifier.size(36.dp),
+                        contentAlignment = Alignment.Center
+                    ) {
+                        Image(
+                            painter = icon.painter,
+                            contentDescription = contentDescription,
+                            modifier = Modifier
+                                .size(36.dp)
+                                .clip(RoundedCornerShape(8.dp)),
+                            contentScale = ContentScale.Fit
+                        )
+                    }
+                }
+            }
+
+            ThumbnailImage(
+                key = key,
+                modifier = modifier,
+                contentScale = contentScale,
+                alignment = alignment
+            )
+        }
     }
 }
 
@@ -120,11 +130,13 @@ private fun ThumbnailImage(
 @Preview
 @Composable
 private fun ThumbnailCardPreview() {
-    ThumbnailCard(
-        url = "https://mozilla.com",
-        key = "123",
-        modifier = Modifier
-            .size(108.dp, 80.dp)
-            .clip(RoundedCornerShape(8.dp))
-    )
+    FirefoxTheme(theme = Theme.getTheme(isPrivate = false)) {
+        ThumbnailCard(
+            url = "https://mozilla.com",
+            key = "123",
+            modifier = Modifier
+                .size(108.dp, 80.dp)
+                .clip(RoundedCornerShape(8.dp))
+        )
+    }
 }
--- a/app/src/main/java/org/mozilla/fenix/compose/tabstray/MediaImage.kt
+++ b/app/src/main/java/org/mozilla/fenix/compose/tabstray/MediaImage.kt
@@ -1,0 +1,71 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.compose.tabstray
+
+import android.content.res.Configuration
+import androidx.appcompat.content.res.AppCompatResources
+import androidx.compose.foundation.Image
+import androidx.compose.foundation.clickable
+import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.width
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.LocalContext
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import com.google.accompanist.drawablepainter.rememberDrawablePainter
+import mozilla.components.browser.state.state.TabSessionState
+import mozilla.components.browser.state.state.createTab
+import mozilla.components.concept.engine.mediasession.MediaSession.PlaybackState
+import org.mozilla.fenix.R
+import org.mozilla.fenix.theme.FirefoxTheme
+import org.mozilla.fenix.theme.Theme
+
+/**
+ * Controller buttons for the media (play/pause) state for the given [tab].
+ *
+ * @param tab [TabSessionState] which the image should be shown.
+ * @param onMediaIconClicked handles the click event when tab has media session like play/pause.
+ * @param modifier [Modifier] to be applied to the layout.
+ */
+@Composable
+fun MediaImage(
+    tab: TabSessionState,
+    onMediaIconClicked: ((TabSessionState) -> Unit),
+    modifier: Modifier,
+) {
+    val (icon, contentDescription) = when (tab.mediaSessionState?.playbackState) {
+        PlaybackState.PAUSED -> {
+            R.drawable.media_state_play to R.string.mozac_feature_media_notification_action_play
+        }
+        PlaybackState.PLAYING -> {
+            R.drawable.media_state_pause to R.string.mozac_feature_media_notification_action_pause
+        }
+        else -> return
+    }
+    val drawable = AppCompatResources.getDrawable(LocalContext.current, icon)
+    // Follow up ticket https://github.com/mozilla-mobile/fenix/issues/25774
+    Image(
+        painter = rememberDrawablePainter(drawable = drawable),
+        contentDescription = stringResource(contentDescription),
+        modifier = modifier.clickable { onMediaIconClicked(tab) },
+    )
+}
+
+@Composable
+@Preview(uiMode = Configuration.UI_MODE_NIGHT_YES)
+@Preview(uiMode = Configuration.UI_MODE_NIGHT_NO)
+private fun ImagePreview() {
+    FirefoxTheme(theme = Theme.getTheme(isPrivate = false)) {
+        MediaImage(
+            tab = createTab(url = "https://mozilla.com"),
+            onMediaIconClicked = {},
+            modifier = Modifier
+                .height(100.dp)
+                .width(200.dp)
+        )
+    }
+}
--- a/app/src/main/java/org/mozilla/fenix/compose/tabstray/TabListItem.kt
+++ b/app/src/main/java/org/mozilla/fenix/compose/tabstray/TabListItem.kt
@@ -1,0 +1,202 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.compose.tabstray
+
+import android.content.res.Configuration
+import androidx.compose.foundation.ExperimentalFoundationApi
+import androidx.compose.foundation.background
+import androidx.compose.foundation.combinedClickable
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.size
+import androidx.compose.foundation.shape.CircleShape
+import androidx.compose.material.Card
+import androidx.compose.material.Icon
+import androidx.compose.material.IconButton
+import androidx.compose.material.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.res.colorResource
+import androidx.compose.ui.res.painterResource
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+import mozilla.components.browser.state.state.TabSessionState
+import mozilla.components.browser.state.state.createTab
+import org.mozilla.fenix.R
+import org.mozilla.fenix.compose.ThumbnailCard
+import org.mozilla.fenix.ext.toShortUrl
+import org.mozilla.fenix.theme.FirefoxTheme
+import org.mozilla.fenix.theme.Theme
+
+/**
+ * List item used to display a tab that supports clicks,
+ * long clicks, multiselection, and media controls.
+ *
+ * @param tab The given tab to be render as view a list item.
+ * @param isSelected Indicates if the item should be render as selected.
+ * @param multiSelectionEnabled Indicates if the item should be render with multi selection options,
+ * enabled.
+ * @param multiSelectionSelected Indicates if the item should be render as multi selection selected
+ * option.
+ * @param onCloseClick Callback to handle the click event of the close button.
+ * @param onMediaClick Callback to handle when the media item is clicked.
+ * @param onClick Callback to handle when item is clicked.
+ * @param onLongClick Callback to handle when item is long clicked.
+ */
+@OptIn(ExperimentalFoundationApi::class)
+@Composable
+@Suppress("MagicNumber")
+fun TabListItem(
+    tab: TabSessionState,
+    isSelected: Boolean = false,
+    multiSelectionEnabled: Boolean = false,
+    multiSelectionSelected: Boolean = false,
+    onCloseClick: (tab: TabSessionState) -> Unit,
+    onMediaClick: (tab: TabSessionState) -> Unit,
+    onClick: (tab: TabSessionState) -> Unit,
+    onLongClick: (tab: TabSessionState) -> Unit,
+) {
+
+    val contentBackgroundColor = if (isSelected) {
+        FirefoxTheme.colors.layerAccentNonOpaque
+    } else {
+        FirefoxTheme.colors.layer1
+    }
+    Row(
+        modifier = Modifier
+            .fillMaxWidth()
+            .background(contentBackgroundColor)
+            .combinedClickable(
+                onLongClick = { onLongClick(tab) },
+                onClick = { onClick(tab) }
+            )
+            .padding(horizontal = 16.dp, vertical = 8.dp),
+        verticalAlignment = Alignment.CenterVertically
+    ) {
+        Thumbnail(
+            tab = tab,
+            multiSelectionEnabled = multiSelectionEnabled,
+            isSelected = multiSelectionSelected,
+            onMediaIconClicked = { onMediaClick(it) }
+        )
+
+        Column(
+            modifier = Modifier
+                .padding(horizontal = 16.dp)
+                .weight(weight = 1f)
+        ) {
+            Text(
+                text = tab.content.title,
+                fontSize = 16.sp,
+                maxLines = 2,
+                color = FirefoxTheme.colors.textPrimary,
+            )
+
+            Text(
+                text = tab.content.url.toShortUrl(),
+                fontSize = 12.sp,
+                color = FirefoxTheme.colors.textSecondary,
+            )
+        }
+
+        if (!multiSelectionEnabled) {
+            IconButton(
+                onClick = { onCloseClick(tab) },
+                modifier = Modifier.size(size = 24.dp),
+            ) {
+                Icon(
+                    painter = painterResource(id = R.drawable.mozac_ic_close),
+                    contentDescription = stringResource(
+                        id = R.string.close_tab_title,
+                        tab.content.title
+                    ),
+                    tint = FirefoxTheme.colors.iconPrimary
+                )
+            }
+        }
+    }
+}
+
+@Composable
+private fun Thumbnail(
+    tab: TabSessionState,
+    multiSelectionEnabled: Boolean,
+    isSelected: Boolean,
+    onMediaIconClicked: ((TabSessionState) -> Unit)
+) {
+    Box {
+        ThumbnailCard(
+            url = tab.content.url,
+            key = tab.id,
+            modifier = Modifier.size(width = 92.dp, height = 72.dp),
+            contentDescription = stringResource(id = R.string.mozac_browser_tabstray_open_tab),
+        )
+
+        if (isSelected) {
+            Card(
+                modifier = Modifier
+                    .size(size = 40.dp)
+                    .align(alignment = Alignment.Center),
+                shape = CircleShape,
+                backgroundColor = FirefoxTheme.colors.layerAccent,
+            ) {
+                Icon(
+                    painter = painterResource(id = R.drawable.mozac_ic_check),
+                    modifier = Modifier
+                        .matchParentSize()
+                        .padding(all = 8.dp),
+                    contentDescription = null,
+                    tint = colorResource(id = R.color.mozac_ui_icons_fill)
+                )
+            }
+        }
+
+        if (!multiSelectionEnabled) {
+            MediaImage(
+                tab = tab,
+                onMediaIconClicked = onMediaIconClicked,
+                modifier = Modifier.align(Alignment.TopEnd)
+            )
+        }
+    }
+}
+
+@Composable
+@Preview(uiMode = Configuration.UI_MODE_NIGHT_YES)
+@Preview(uiMode = Configuration.UI_MODE_NIGHT_NO)
+private fun TabListItemPreview() {
+    FirefoxTheme(theme = Theme.getTheme(isPrivate = false)) {
+        TabListItem(
+            tab = createTab(url = "www.mozilla.com", title = "Mozilla"),
+            onCloseClick = {},
+            onMediaClick = {},
+            onClick = {},
+            onLongClick = {},
+        )
+    }
+}
+
+@Composable
+@Preview(uiMode = Configuration.UI_MODE_NIGHT_YES)
+@Preview(uiMode = Configuration.UI_MODE_NIGHT_NO)
+private fun SelectedTabListItemPreview() {
+    FirefoxTheme(theme = Theme.getTheme(isPrivate = false)) {
+        TabListItem(
+            tab = createTab(url = "www.mozilla.com", title = "Mozilla"),
+            onCloseClick = {},
+            onMediaClick = {},
+            onClick = {},
+            onLongClick = {},
+            multiSelectionEnabled = true,
+            multiSelectionSelected = true,
+        )
+    }
+}
--- a/app/src/main/java/org/mozilla/fenix/tabstray/browser/compose/ComposeAbstractTabViewHolder.kt
+++ b/app/src/main/java/org/mozilla/fenix/tabstray/browser/compose/ComposeAbstractTabViewHolder.kt
@@ -1,0 +1,52 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.tabstray.browser.compose
+
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.platform.ComposeView
+import androidx.lifecycle.LifecycleOwner
+import androidx.lifecycle.ViewTreeLifecycleOwner
+import androidx.recyclerview.widget.RecyclerView
+import androidx.savedstate.SavedStateRegistryOwner
+import androidx.savedstate.ViewTreeSavedStateRegistryOwner
+import mozilla.components.browser.state.state.TabSessionState
+import mozilla.components.browser.tabstray.SelectableTabViewHolder
+import org.mozilla.fenix.compose.ComposeViewHolder
+import org.mozilla.fenix.theme.FirefoxTheme
+
+/**
+ * [RecyclerView.ViewHolder] used for Jetpack Compose UI content .
+ *
+ * @param composeView [ComposeView] which will be populated with Jetpack Compose UI content.
+ * @param viewLifecycleOwner [LifecycleOwner] life cycle owner for the view.
+ */
+abstract class ComposeAbstractTabViewHolder(
+    private val composeView: ComposeView,
+    private val viewLifecycleOwner: LifecycleOwner
+) : SelectableTabViewHolder(composeView) {
+
+    /**
+     * Composable that contains the content for a specific [ComposeViewHolder] implementation.
+     */
+    @Composable
+    abstract fun Content(tab: TabSessionState)
+
+    /**
+     * Binds the a  composable to the [composeView].
+     */
+    fun bind(tab: TabSessionState) {
+        composeView.setContent {
+            FirefoxTheme {
+                Content(tab)
+            }
+        }
+
+        ViewTreeLifecycleOwner.set(composeView, viewLifecycleOwner)
+        ViewTreeSavedStateRegistryOwner.set(
+            composeView,
+            viewLifecycleOwner as SavedStateRegistryOwner
+        )
+    }
+}
--- a/app/src/main/java/org/mozilla/fenix/tabstray/browser/compose/ComposeListViewHolder.kt
+++ b/app/src/main/java/org/mozilla/fenix/tabstray/browser/compose/ComposeListViewHolder.kt
@@ -1,0 +1,112 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.tabstray.browser.compose
+
+import android.view.View
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.collectAsState
+import androidx.compose.runtime.getValue
+import androidx.compose.ui.platform.ComposeView
+import androidx.lifecycle.LifecycleOwner
+import kotlinx.coroutines.flow.MutableStateFlow
+import mozilla.components.browser.state.state.TabSessionState
+import mozilla.components.browser.tabstray.TabsTray
+import mozilla.components.browser.tabstray.TabsTrayStyling
+import mozilla.components.lib.state.ext.observeAsComposableState
+import org.mozilla.fenix.compose.tabstray.TabListItem
+import org.mozilla.fenix.selection.SelectionHolder
+import org.mozilla.fenix.tabstray.TabsTrayState
+import org.mozilla.fenix.tabstray.TabsTrayStore
+import org.mozilla.fenix.tabstray.browser.BrowserTrayInteractor
+
+/**
+ * A Compose ViewHolder implementation for "tab" items with list layout.
+ *
+ * @param interactor [BrowserTrayInteractor] handling tabs interactions in a tab tray.
+ * @param tabsTrayStore [TabsTrayStore] containing the complete state of tabs tray and methods to update that.
+ * @param selectionHolder [SelectionHolder]<[TabSessionState]> for helping with selecting
+ * any number of displayed [TabSessionState]s.
+ * @param composeItemView that displays a "tab".
+ * @param featureName [String] representing the name of the feature displaying tabs. Used in telemetry reporting.
+ * @param viewLifecycleOwner [LifecycleOwner] to which this Composable will be tied to.
+ */
+class ComposeListViewHolder(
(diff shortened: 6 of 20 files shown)


app/src/main/java/org/mozilla/fenix/compose/TabImage.kt
```diff
@@ -0,0 +1,163 @@
...
+    modifier: Modifier = Modifier,
+    thumbnailDescription: String? = null,
+    contentScale: ContentScale = ContentScale.FillWidth,
+    alignment: Alignment = Alignment.TopCenter
+) {
+    val previewImageUrl = tab.content.previewImageUrl
+    val thumbnail = tab.content.thumbnail
+
+    when {
+        !previewImageUrl.isNullOrEmpty() -> {
```
gabrielluong, line 60: I don't think you can refactor this for the tabs tray because it will always prioritize displaying the hero image, fallback to thumbnail. This image ordering is specific to recent tabs and bookmarks.

app/src/main/java/org/mozilla/fenix/compose/TabUrl.kt
```diff
@@ -0,0 +1,61 @@
...
+@Composable
+fun TabUrl(
+    url: String,
+    publicSuffixList: PublicSuffixList,
+    modifier: Modifier = Modifier,
+    fontSize: TextUnit = 12.sp,
+    color: Color = FirefoxTheme.colors.textSecondary
+) {
+
+    fun getUrl(): String {
```
gabrielluong, line 33: Can we refactor this into support-ktx/utils if it's used in multiple places?

app/src/main/java/org/mozilla/fenix/compose/tabstray/ListTabItem.kt
```diff
@@ -0,0 +1,170 @@
...
+        CloseIcon(
+            tab = tab,
+            visible = !multiSelectionEnabled,
+            delegate = delegate
+        )
+    }
+}
+
+@Composable
+private fun RowScope.Thumbnail(tab: TabSessionState, isSelected:Boolean) {
```
gabrielluong, line 76: This shouldn't be an extension of RowScope

app/src/main/java/org/mozilla/fenix/compose/tabstray/ListTabItem.kt
```diff
@@ -0,0 +1,170 @@
...
+                    contentDescription = null,
+                    tint = colorResource(id = R.color.mozac_ui_icons_fill)
+                )
+            }
+        }
+    }
+}
+
+@Composable
+private fun RowScope.CloseIcon(
```
gabrielluong, line 116: Same as above.

app/src/main/java/org/mozilla/fenix/compose/tabstray/ListTabItem.kt
```diff
@@ -0,0 +1,170 @@
...
+import mozilla.components.browser.state.state.TabSessionState
+import mozilla.components.browser.tabstray.TabsTray
+import org.mozilla.fenix.R
+import org.mozilla.fenix.compose.TabImage
+import org.mozilla.fenix.compose.TabTitle
+import org.mozilla.fenix.compose.TabUrl
+import org.mozilla.fenix.ext.components
+
+@Composable
+fun ListTabItem(
```
gabrielluong, line 44: Maybe rename this `TabListItem`? Doing it that way reminds me more of a BEM (Block - Element - Modifier) CSS naming convention. 

The Block conveys the component, which in this case is the Tab.
The Element is inside of the component, which is the List Item.

app/src/main/java/org/mozilla/fenix/compose/tabstray/ListTabItem.kt
```diff
@@ -0,0 +1,170 @@
...
+    Box(
+        modifier = Modifier.Companion
+            .weight(2f)
+            .padding(start = 16.dp, end = 12.dp, top = 8.dp, bottom = 8.dp)
+    ) {
+        TabImage(
+            tab = tab,
+            modifier = Modifier
+                .size(
+                    dimensionResource(id = R.dimen.tab_tray_list_item_thumbnail_width),
```
gabrielluong, line 86: I can in-vision a world where we simply keep the dimension values inside of the Composable function. In factor, we do a mix of this in our layout xml and Composable function currently. Ideally, when we drop the layout xml for the tab list item, let's also remove the value defined in dimens.xml if we only need the value within the Composable function.

app/src/main/java/org/mozilla/fenix/compose/tabstray/ListTabItem.kt
```diff
@@ -0,0 +1,170 @@
...
+            }
+        }
+    }
+}
+
+@Composable
+private fun RowScope.CloseIcon(
+    tab: TabSessionState,
+    visible: Boolean,
+    delegate: TabsTray.Delegate?
```
gabrielluong, line 119: Rather than doing this, you can define a lambda function such as `onCloseClick`.

gabrielluong:
Apologizes for the drive by review. Let's limit this PR to only the `ListViewHolder` refactoring. I think if we also did the `GridViewHolder` in this PR - it would just get too big with the refactoring and makes it harder to review. There is already enough separation between the two that we don't have to refactor both in one PR. It might even be better to just see how the initial perf is affected in just one view instead of affecting both.

Amejia481:
Thanks for the review really I appreciate it, I like the idea of splitting into two. I'll tag @jonalmeida @rocketsroger  to see what they think.

I almost finished the `GridViewHolder` locally, but still I need to implement media functionality for both view holders.

app/src/main/java/org/mozilla/fenix/compose/TabUrl.kt
Amejia481: This is already an extension function in Fenix :) [see](https://github.com/mozilla-mobile/fenix/blob/e6cae9b8007051df0a72931b3361654c4c018a22/app/src/main/java/org/mozilla/fenix/ext/String.kt#L43) 
I think uplifting it to AC could be a separated ticket as it's outside of the scope of this PR.

Amejia481: I filed https://github.com/mozilla-mobile/fenix/issues/22770 for it

app/src/test/java/org/mozilla/fenix/tabstray/browser/BrowserTabsAdapterTest.kt
```diff
@@ -49,41 +43,41 @@ class BrowserTabsAdapterTest {
...
-        verify { holder.showTabIsMultiSelectEnabled(any(), true) }
-    }
-
-    private val testSelectionHolder = object : SelectionHolder<TabSessionState> {
-        override val selectedItems: Set<TabSessionState>
-            get() = internalState
-
-        val internalState = mutableSetOf<TabSessionState>()
-    }
+    // @Test
```
Amejia481, line 46: Waiting to have everything ✅ before start writing tests

app/src/main/java/org/mozilla/fenix/compose/TabImage.kt
Amejia481: Is there any specific reason about why we couldn't use the hero image on the tabs tray too?

gabrielluong: That's a UX question.

gabrielluong: You would effectively be switching all the thumbnails to be using the hero image with this change assume every site you visit has a hero image. For that reason alone, I don't think we want to pursue this.

Amejia481: I update the approach :)

app/src/main/java/org/mozilla/fenix/compose/tabstray/TabListItem.kt
```diff
@@ -0,0 +1,185 @@
...
+}
+
+@Composable
+private fun CloseIcon(
+    tab: TabSessionState,
+    visible: Boolean,
+    modifier: Modifier,
+    onCloseClicked: ((TabSessionState) -> Unit)?
+) {
+    if (visible) {
```
gabrielluong, line 138: We typically don't want to execute a composable function if it can be avoid. I would move the `if (visible)` check outside of `CloseIcon` and let the caller decide whether or not `CloseIcon` needs to be called.

app/src/main/java/org/mozilla/fenix/compose/tabstray/TabListItem.kt
```diff
@@ -0,0 +1,185 @@
...
+                    .clickable { onCloseClicked?.invoke(tab) }
+            )
+        }
+    }
+}
+
+@Composable
+private fun DescriptionSection(
+    modifier: Modifier,
+    tab: TabSessionState
```
gabrielluong, line 166: We might be able to simplify this by just passing in `title` and `url`. It is probably better not to pass in such big data class like TabSessionState to avoid recomposing when we only care about 2 string states.

app/src/main/java/org/mozilla/fenix/compose/tabstray/TabListItem.kt
```diff
@@ -0,0 +1,185 @@
...
+        }
+    }
+}
+
+@Composable
+private fun CloseIcon(
+    tab: TabSessionState,
+    visible: Boolean,
+    modifier: Modifier,
+    onCloseClicked: ((TabSessionState) -> Unit)?
```
gabrielluong, line 136: I don't think we want onCloseClicked to be nullable. I would prefer if we provided a `{}` default value.

app/src/main/java/org/mozilla/fenix/compose/tabstray/TabListItem.kt
```diff
@@ -0,0 +1,185 @@
...
+
+@Composable
+@Suppress("TooManyFunctions", "LongParameterList", "MagicNumber")
+fun TabListItem(
+    tab: TabSessionState,
+    isSelected: Boolean = false,
+    multiSelectionEnabled: Boolean = false,
+    multiSelectionSelected: Boolean = false,
+    onCloseClicked: ((TabSessionState) -> Unit)? = null,
+    onMediaClicked: ((TabSessionState) -> Unit)? = null,
```
gabrielluong, line 49:
```suggestion
    onCloseClicked: ((TabSessionState) -> Unit) = {},
    onMediaClicked: ((TabSessionState) -> Unit) = {},
```

app/src/main/java/org/mozilla/fenix/compose/TabUrl.kt
```diff
@@ -0,0 +1,61 @@
...
+import org.mozilla.fenix.theme.FirefoxTheme
+
+/**
+ * Default layout for a tab composable caption.
+ *
+ */
+@Composable
+fun TabUrl(
+    url: String,
+    publicSuffixList: PublicSuffixList,
```
gabrielluong, line 27: Is there a way we can refactor this such that we don't have to worry about passing `PublicSuffixList` here?

app/src/main/java/org/mozilla/fenix/compose/tabstray/TabListItem.kt
```diff
@@ -0,0 +1,185 @@
...
+            .height(dimensionResource(id = R.dimen.tab_tray_tab_item_height))
+            .background(colorResource(id = contentBackgroundColor))
+    ) {
+
+        Thumbnail(
+            tab,
+            multiSelectionEnabled,
+            multiSelectionSelected,
+            Modifier.weight(2f),
+            onMediaClicked
```
gabrielluong, line 68: Can we add the named parameters?

app/src/main/java/org/mozilla/fenix/compose/TabUrl.kt
Amejia481: The only way that I could think of it's to use internally via the `context.components.publicSuffixList`, but that will make testing very hard,

Amejia481: Any suggestions?

Amejia481:
@gabrielluong I updated the PR with your suggestions

app/src/main/java/org/mozilla/fenix/compose/TabUrl.kt
jonalmeida: You could make it a lambda that takes and returns a string with a reasonable default - so that it's up the consumer to provide the appropriate function:

```kotlin
@Composable
fun TabUrl(
  url: String,
  shortener: (String) -> String = { it }
  // ...
) { }

fun shortenUrl(): String {
  val publicSuffixList = context.components.publicSuffixList
  return toShortUrl(publicSuffixList).take(MAX_URI_LENGTH)
}

// Using it..
TabUrl(
  "https://veryverylongurl.com",
  ::shortenUrl,
  // ...
) { }
```

Amejia481: 👍🏽

Amejia481:
Added @Mugurell as a reviewer as he has working on migrating many viewholder to compose. It will be great to have his input.

app/src/main/java/org/mozilla/fenix/compose/TabUrl.kt
```diff
@@ -0,0 +1,60 @@
...
+        // Truncate to MAX_URI_LENGTH to prevent the UI from locking up for
+        // extremely large URLs such as data URIs or bookmarklets. The same
+        // is done in the toolbar and awesomebar:
+        // https://github.com/mozilla-mobile/fenix/issues/1824
+        // https://github.com/mozilla-mobile/android-components/issues/6985
+        return url.toShortUrl(publicSuffixList)
+            .take(MAX_URI_LENGTH)
+    }
+
+    TabSubtitle(
```
gabrielluong, line 42: Can you use `SecondaryText()` for this

app/src/main/java/org/mozilla/fenix/compose/TabUrl.kt
```diff
@@ -0,0 +1,60 @@
...
+import mozilla.components.browser.toolbar.MAX_URI_LENGTH
+import mozilla.components.lib.publicsuffixlist.PublicSuffixList
+import org.mozilla.fenix.ext.toShortUrl
+import org.mozilla.fenix.theme.FirefoxTheme
+
+/**
+ * Renders a text widget for an URL.
+ */
+@Composable
+fun TabUrl(
```
gabrielluong, line 24: This might be unique to Tabs Tray so I would move it there.

app/src/main/java/org/mozilla/fenix/compose/TabSubtitle.kt
```diff
@@ -26,18 +28,20 @@ import org.mozilla.fenix.theme.Theme
 @Composable
 fun TabSubtitle(
     text: String,
-    modifier: Modifier = Modifier
+    modifier: Modifier = Modifier,
+    fontSize: TextUnit = 12.sp,
+    color: Color = when (isSystemInDarkTheme()) {
+        true -> FirefoxTheme.colors.textPrimary
+        false -> FirefoxTheme.colors.textSecondary
+    }
```
gabrielluong, line 36: I am actually looking to replace these with `PrimaryText()` and `SecondaryText()`. Can you look to use either one of those in your implementation?

app/src/main/java/org/mozilla/fenix/compose/TabThumbnail.kt
```diff
@@ -0,0 +1,149 @@
...
+            )
+        }
+        else -> {
+            TabIconLoader(modifier, tab, thumbnailDescription, contentScale, alignment)
+        }
+    }
+}
+
+@Composable
+fun TabIconLoader(
```
gabrielluong, line 72: This is duplicating `ThumbnailCard`? Can we look to use that?

app/src/main/java/org/mozilla/fenix/compose/TabTitle.kt
```diff
@@ -28,15 +30,17 @@ import org.mozilla.fenix.theme.Theme
 fun TabTitle(
     text: String,
     maxLines: Int,
-    modifier: Modifier = Modifier
+    modifier: Modifier = Modifier,
```
gabrielluong, line 33: Can you undo the changes to this and look to use `PrimaryText()`

app/src/main/res/drawable/media_state_pause_vector.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
gabrielluong, line 5: Can we just replace the existing drawables with this?

gabrielluong (changes requested):
Left you some thoughts on cleanups.

app/src/main/java/org/mozilla/fenix/compose/tabstray/TabListItem.kt
```diff
@@ -0,0 +1,205 @@
...
+            alignment = Alignment.TopCenter
+        )
+
+        if (isSelected) {
+            Card(
+                modifier = Modifier
+                    .size(40.dp)
+                    .align(Alignment.Center),
+                shape = CircleShape,
+                backgroundColor = colorResource(id = R.color.accent_normal_theme),
```
MozillaNoah, line 135: same about the accent color here

app/src/main/java/org/mozilla/fenix/compose/tabstray/TabListItem.kt
```diff
@@ -0,0 +1,205 @@
...
+        modifier = modifier
+            .fillMaxHeight()
+            .padding(
+                top = 24.dp, bottom = 24.dp
+            ),
+        horizontalAlignment = Alignment.CenterHorizontally
+    ) {
+        Image(
+            alignment = Alignment.Center,
+            colorFilter = ColorFilter.tint(colorResource(id = R.color.fx_mobile_icon_color_primary)),
```
MozillaNoah, line 169:
```suggestion
            colorFilter = ColorFilter.tint(FirefoxTheme.colors.iconPrimary),
```

app/src/main/java/org/mozilla/fenix/compose/tabstray/TabListItem.kt
```diff
@@ -0,0 +1,205 @@
...
+
+@Composable
+private fun DescriptionSection(
+    modifier: Modifier,
+    title: String,
+    url: String
+) {
+    Column(
+        modifier = modifier
+            .padding(top = 16.dp)
```
MozillaNoah, line 190: If it's just a single top padding, I think we prefer to use a `Spacer` instead.

app/src/main/java/org/mozilla/fenix/compose/tabstray/TabListItem.kt
```diff
@@ -0,0 +1,205 @@
...
+    ) {
+        TabTitle(
+            text = title,
+            color = colorResource(id = R.color.fx_mobile_text_color_primary),
+            fontSize = 16.sp,
+            maxLines = 2
+        )
+        TabUrl(
```
MozillaNoah, line 198: nit: missing a newline between the Composables

app/src/main/java/org/mozilla/fenix/compose/tabstray/MediaImage.kt
```diff
@@ -0,0 +1,90 @@
...
+ * @param modifier [Modifier] to be applied to the layout.
+ */
+@Composable
+fun MediaImage(
+    tab: TabSessionState,
+    onMediaIconClicked: ((TabSessionState) -> Unit)?,
+    modifier: Modifier,
+) {
+    val backGroundColor = if (tab.content.private) {
+        R.color.accent_private_theme
```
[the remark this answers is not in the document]
gabrielluong: I have asked Arturo to revisit all the colors since he might have been looking at the legacy code prior to the design token changes.

app/src/main/java/org/mozilla/fenix/compose/TabThumbnail.kt
```diff
@@ -0,0 +1,149 @@
...
+    Card(
+        modifier = modifier,
+        backgroundColor = colorResource(id = R.color.photonGrey20)
+    ) {
+        components.core.icons.Loader(tab.content.url) {
+            Placeholder {
+                Box(
+                    modifier = Modifier.background(
+                        color = when (isSystemInDarkTheme()) {
+                            true -> PhotonColors.DarkGrey30
```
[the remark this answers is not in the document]
gabrielluong: I assume this won't be an issue if we reuse the existing Thumbnail composables.

app/src/main/java/org/mozilla/fenix/compose/TabThumbnail.kt
```diff
@@ -0,0 +1,149 @@
...
+    tab: TabSessionState,
+    thumbnailDescription: String?,
+    contentScale: ContentScale,
+    alignment: Alignment
+) {
+    Card(
+        modifier = modifier,
+        backgroundColor = colorResource(id = R.color.photonGrey20)
+    ) {
+        components.core.icons.Loader(tab.content.url) {
```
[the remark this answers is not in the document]
gabrielluong: I assume this won't be an issue if we reuse the existing Thumbnail composables.

app/src/main/java/org/mozilla/fenix/tabstray/browser/compose/ComposeListViewHolder.kt
```diff
@@ -0,0 +1,117 @@
...
+            multiSelectionEnabled = isSelectModeActive(),
+            multiSelectionSelected = isMultiSelectionSelected.value,
+            onCloseClicked = ::onCloseClicked,
+            onMediaClicked = ::onMediaClicked,
+            onClick = ::onClick,
+            onLongClick = ::onLongClick,
+        )
+    }
+
+    private fun isSelectModeActive() = tabsTrayStore.state.mode is TabsTrayState.Mode.Select
```
Mugurell, line 112: I think this also has to be made observable.
Saw that if I manually tap some tabs to select and then tap them all to deselect at least one will not show the close ("X") button.
We could probably do 
```
        val isSelectModeActive = tabsTrayStore.observeAsComposableState {
            state -> state.mode is TabsTrayState.Mode.Select
        }.value
```

inside the composable if this is the issue.

app/src/main/java/org/mozilla/fenix/tabstray/browser/compose/ComposeAbstractTabViewHolder.kt
```diff
@@ -0,0 +1,53 @@
...
+    /**
+     * Composable that contains the content for a specific [ComposeViewHolder] implementation.
+     */
+    @Composable
+    abstract fun Content(tab: TabSessionState)
+
+    fun bind(tab: TabSessionState) {
+        composeView.setViewCompositionStrategy(
+            ViewCompositionStrategy.DisposeOnViewTreeLifecycleDestroyed
+        )
```
Mugurell, line 40: We probably don't need to set a custom strategy and can leverage the default which would have the composition destroyed when the ViewHolder is recycled.
Otherwise since we're injecting state from outside through `updateSelectedTabIndicator` and the likes this will leak to the newly wrapped composable that would get to use a recycled ViewHolder.
Can be tested by just selecting some items -> deleting them -> tap Undo -> tap one of the same items.

app/build.gradle
```diff
@@ -537,6 +537,7 @@ dependencies {
     debugImplementation Deps.leakcanary
 
     implementation Deps.androidx_compose_ui
+    implementation Deps.androidx_compose_livedata
```
Mugurell, line 540: This can probably be dropped in favor of using the simpler `coroutines.flow.StateFlow` api.

app/src/main/java/org/mozilla/fenix/compose/tabstray/TabListItem.kt
```diff
@@ -0,0 +1,205 @@
...
+    tab: TabSessionState,
+    modifier: Modifier,
+    onCloseClicked: ((TabSessionState) -> Unit) = {}
+) {
+    Column(
+        modifier = modifier
+            .fillMaxHeight()
+            .padding(
+                top = 24.dp, bottom = 24.dp
+            ),
```
Mugurell, line 164:
```suggestion
            .padding(vertical = 24.dp),
```

buildSrc/src/main/java/Dependencies.kt
```diff
@@ -176,6 +176,7 @@ object Deps {
     const val sentry = "io.sentry:sentry-android:${Versions.sentry}"
     const val leakcanary = "com.squareup.leakcanary:leakcanary-android-core:${Versions.leakcanary}"
 
+    const val androidx_compose_livedata = "androidx.compose.runtime:runtime-livedata:${Versions.androidx_compose}"
```
Mugurell, line 179: This can probably be dropped in favor of using the simpler coroutines.flow.StateFlow api.

Mugurell (commented):
Some small edgecases to test and a bit of work to get this up-to-date with the latest compose developments but generally works okay 👍 and it's nice to see more compose in tabs tray.
(I like that a11y which is more often left at the end seems great here)

app/src/main/java/org/mozilla/fenix/tabstray/browser/compose/ComposeListViewHolder.kt
```diff
@@ -0,0 +1,117 @@
...
+        isSelectedTab.value = isSelected
+        bind(tab)
+    }
+
+    override fun updateSelectedTabIndicator(showAsSelected: Boolean) {
+        isSelectedTab.postValue(showAsSelected)
+    }
+
+    override fun showTabIsMultiSelectEnabled(selectedMaskView: View?, isSelected: Boolean) {
+        isMultiSelectionSelected.postValue(isSelected)
```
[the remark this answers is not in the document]
Amejia481: Thanks for all the helpful tips

Amejia481:
Thanks @gabrielluong, @jonalmeida @MozillaNoah, , @Mugurell for all your the helpful tips, I really appreciate it ❤️!!
I rebased and address all the comments based on your feedback!

app/src/main/java/org/mozilla/fenix/compose/tabstray/MediaImage.kt
```diff
@@ -0,0 +1,84 @@
...
+                shape = CircleShape
+            ),
+        backgroundColor = FirefoxTheme.colors.layerAccent
+    ) {
+        Icon(
+            painter = painterResource(id = icon),
+            modifier = Modifier
+                .size(size = 24.dp)
+                .clickable { onMediaIconClicked.invoke(tab) },
+            contentDescription = stringResource(id = contentDescription),
```
MozillaNoah, line 68: nit: no need to use the named parameter with stringResource if it's just the string ID.

app/src/main/java/org/mozilla/fenix/compose/tabstray/MediaImage.kt
```diff
@@ -0,0 +1,84 @@
...
+                color = FirefoxTheme.colors.layer2,
+                shape = CircleShape
+            ),
+        backgroundColor = FirefoxTheme.colors.layerAccent
+    ) {
+        Icon(
+            painter = painterResource(id = icon),
+            modifier = Modifier
+                .size(size = 24.dp)
+                .clickable { onMediaIconClicked.invoke(tab) },
```
MozillaNoah, line 67:
```suggestion
                .clickable { onMediaIconClicked(tab) },
```

app/src/main/java/org/mozilla/fenix/compose/ThumbnailCard.kt
```diff
@@ -48,43 +51,50 @@ fun ThumbnailCard(
...
-                    Image(
-                        painter = icon.painter,
-                        contentDescription = null,
-                        modifier = Modifier
-                            .size(36.dp)
-                            .clip(RoundedCornerShape(8.dp)),
-                        contentScale = ContentScale.Fit
+        if (inComposePreview) {
+            Box(
+                modifier = Modifier.background(color = FirefoxTheme.colors.layer3)
```
MozillaNoah, line 64: I think this `Box` may need the same Modifier treatment treatment as the `Box` in `WithIcon`, so there isn't any size adjustment pop-in. (I can't get this PR to build locally currently because of a Nimbus issue, so can't immediately verify)

app/src/main/java/org/mozilla/fenix/compose/tabstray/MediaImage.kt
```diff
@@ -0,0 +1,84 @@
...
+        PlaybackState.PAUSED -> {
+            R.drawable.media_state_play_vector to R.string.mozac_feature_media_notification_action_play
+        }
+        PlaybackState.PLAYING -> {
+            R.drawable.media_state_pause_vector to R.string.mozac_feature_media_notification_action_pause
+        }
+        else -> return
+    }
+
+    Card(
```
MozillaNoah, line 52: does this need to be `ThumbnailCard`? Is this supposed to be the play/pause + the thumbnail in one Composable?

app/src/main/java/org/mozilla/fenix/compose/tabstray/TabListItem.kt
```diff
@@ -0,0 +1,190 @@
...
+            .combinedClickable(
+                onLongClick = { onLongClick(tab) },
+                onClick = {
+                    onClick(tab)
+                }
+            )
+            .padding(horizontal = 16.dp, vertical = 8.dp),
+        verticalAlignment = Alignment.CenterVertically
+    ) {
+
```
MozillaNoah, line 87: nit: don't need a newline here.

app/src/main/java/org/mozilla/fenix/compose/tabstray/MediaImage.kt
```diff
@@ -0,0 +1,84 @@
...
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import mozilla.components.browser.state.state.TabSessionState
+import mozilla.components.browser.state.state.createTab
+import mozilla.components.concept.engine.mediasession.MediaSession.PlaybackState
+import org.mozilla.fenix.R
+import org.mozilla.fenix.theme.FirefoxTheme
+
+/**
+ * A composable that lays out and draws the image from a given [tab] while showing a default placeholder
```
MozillaNoah, line 29: I think this kdoc needs to be updated, since this Composable just renders the play/pause overlay. Also, I believe we're trying to move away from directly mentioning in the kdocs that it's a Composable.

Amejia481: 👍🏽

app/src/main/java/org/mozilla/fenix/compose/tabstray/TabListItem.kt
Amejia481: 👍🏽

app/src/main/java/org/mozilla/fenix/compose/tabstray/MediaImage.kt
Amejia481: 👍🏽

gabrielluong:
I think we're waiting for an ac-bump, will look at this again when I can build it.

Amejia481:
This is ready for review :)

app/src/main/java/org/mozilla/fenix/compose/tabstray/MediaImage.kt
```diff
@@ -0,0 +1,83 @@
...
+import mozilla.components.browser.state.state.TabSessionState
+import mozilla.components.browser.state.state.createTab
+import mozilla.components.concept.engine.mediasession.MediaSession.PlaybackState
+import org.mozilla.fenix.R
+import org.mozilla.fenix.theme.FirefoxTheme
+
+/**
+ * Draws controller buttons for the media (play/pause) state of the given [tab].
+ *
+ * @param tab the tab which the image should be shown.
```
MozillaNoah, line 31:
```suggestion
 * @param tab [TabSessionState] which the image should be shown.
```

app/src/main/java/org/mozilla/fenix/compose/tabstray/MediaImage.kt
```diff
@@ -0,0 +1,83 @@
...
+                color = FirefoxTheme.colors.layer2,
+                shape = CircleShape
+            ),
+        backgroundColor = FirefoxTheme.colors.layerAccent
+    ) {
+        Icon(
+            painter = painterResource(id = icon),
+            modifier = Modifier
+                .size(size = 24.dp)
+                .clickable { onMediaIconClicked(tab) },
```
MozillaNoah, line 66: I think we're trying to use `IconButton` instead of an `Icon` + `Modifier.clickable`. Alternatively, the Card has a built-in onClick parameter if we want to have the entire card to be clickable. `IconButton` though adds some baked in padding to make the tap area more accessible.

app/src/main/java/org/mozilla/fenix/compose/tabstray/MediaImage.kt
```diff
@@ -0,0 +1,83 @@
...
+        else -> return
+    }
+
+    Card(
+        modifier = modifier
+            .size(size = 24.dp)
+            .clip(shape = CircleShape)
+            .border(
+                width = 2.dp,
+                color = FirefoxTheme.colors.layer2,
```
MozillaNoah, line 57: we have a few border colors under `FirefoxTheme.colors`, should we use one of those instead here?

app/src/main/java/org/mozilla/fenix/compose/tabstray/MediaImage.kt
```diff
@@ -0,0 +1,83 @@
...
+                .size(size = 24.dp)
+                .clickable { onMediaIconClicked(tab) },
+            contentDescription = stringResource(contentDescription),
+            tint = FirefoxTheme.colors.borderPrimary
+        )
+    }
+}
+
+@Composable
+@Preview
```
MozillaNoah, line 74:
```suggestion
@Preview(uiMode = Configuration.UI_MODE_NIGHT_YES)
@Preview(uiMode = Configuration.UI_MODE_NIGHT_NO)
```

app/src/main/java/org/mozilla/fenix/compose/tabstray/TabListItem.kt
```diff
@@ -0,0 +1,183 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.compose.tabstray
```
MozillaNoah, line 5: wondering if this should live in /compose/tabstray or /compose/list

app/src/main/java/org/mozilla/fenix/compose/tabstray/TabListItem.kt
```diff
@@ -0,0 +1,183 @@
...
+            isSelected = multiSelectionSelected,
+            onMediaIconClicked = { onMediaClick(it) }
+        )
+
+        Column(
+            modifier = Modifier
+                .padding(horizontal = 16.dp)
+                .weight(weight = 1f)
+        ) {
+
```
MozillaNoah, line 97: nit: empty line

app/src/main/java/org/mozilla/fenix/compose/tabstray/TabListItem.kt
```diff
@@ -0,0 +1,183 @@
...
+import org.mozilla.fenix.R
+import org.mozilla.fenix.compose.PrimaryText
+import org.mozilla.fenix.compose.SecondaryText
+import org.mozilla.fenix.compose.ThumbnailCard
+import org.mozilla.fenix.ext.toShortUrl
+import org.mozilla.fenix.theme.FirefoxTheme
+import org.mozilla.fenix.theme.Theme
+
+/**
+ * A composable that renders tab list layout.
```
MozillaNoah, line 40:
```suggestion
 * List item used to display a tab that supports clicks, long clicks, multiselection, and media controls.
```

Just something to mirror the other list items like `TextListItem`, where Composable isn't mentioned and its functionality is highlighted.

app/src/main/java/org/mozilla/fenix/compose/tabstray/TabListItem.kt
```diff
@@ -0,0 +1,183 @@
...
+                tab = tab,
+                onMediaIconClicked = onMediaIconClicked,
+                modifier = Modifier.align(Alignment.TopEnd)
+            )
+        }
+    }
+}
+
+@Composable
+@Preview
```
MozillaNoah, line 172:
```suggestion
@Preview(uiMode = Configuration.UI_MODE_NIGHT_YES)
@Preview(uiMode = Configuration.UI_MODE_NIGHT_NO)
```

app/src/main/java/org/mozilla/fenix/tabstray/browser/compose/ComposeListViewHolder.kt
```diff
@@ -0,0 +1,116 @@
...
+ * @param selectionHolder [SelectionHolder]<[TabSessionState]> for helping with selecting
+ * any number of displayed [TabSessionState]s.
+ * @param composeItemView that displays a "tab".
+ * @param featureName [String] representing the name of the feature displaying tabs. Used in telemetry reporting.
+ * @param viewLifecycleOwner [LifecycleOwner] to which this Composable will be tied to.
+ */
+class ComposeListViewHolder(
+    private val interactor: BrowserTrayInteractor,
+    val tabsTrayStore: TabsTrayStore,
+    val selectionHolder: SelectionHolder<TabSessionState>? = null,
```
MozillaNoah, line 38: should these be private?

app/src/main/java/org/mozilla/fenix/compose/tabstray/MediaImage.kt
```diff
@@ -0,0 +1,83 @@
...
+private fun ImagePreview() {
+    MediaImage(
```
MozillaNoah, line 76: I think this needs to be wrapped with `FirefoxTheme(theme = Theme.getTheme(isPrivate = false)) {`

It wasn't rendering for me locally.

app/src/main/java/org/mozilla/fenix/compose/tabstray/TabListItem.kt
Amejia481: I think this is very specific to the tabstray

app/src/main/java/org/mozilla/fenix/compose/tabstray/MediaImage.kt
```diff
@@ -0,0 +1,83 @@
...
+import org.mozilla.fenix.theme.FirefoxTheme
+
+/**
+ * Draws controller buttons for the media (play/pause) state of the given [tab].
+ *
+ * @param tab the tab which the image should be shown.
+ * @param onMediaIconClicked handles the click event when tab has media session like play/pause.
+ * @param modifier [Modifier] to be applied to the layout.
+ */
+@Composable
```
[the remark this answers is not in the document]
Amejia481: I worked 🎉 .
I updated the PR to include it!

app/src/main/java/org/mozilla/fenix/tabstray/browser/compose/ComposeListViewHolder.kt
Amejia481: Good catch!

Amejia481:
@MozillaNoah @gabrielluong I updated the PR with your recommendations!

app/src/main/java/org/mozilla/fenix/compose/tabstray/TabListItem.kt
```diff
@@ -0,0 +1,185 @@
...
+                modifier = Modifier.align(Alignment.TopEnd)
+            )
+        }
+    }
+}
+
+@Composable
+@Preview(uiMode = Configuration.UI_MODE_NIGHT_YES)
+@Preview(uiMode = Configuration.UI_MODE_NIGHT_NO)
+private fun TabListItemPreview() {
```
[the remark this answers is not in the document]
gabrielluong: Can you also add a preview with multiselect enabled

gabrielluong (approved):
r+ assuming all the feedback is addressed or discussed to satisfaction.

app/src/main/java/org/mozilla/fenix/compose/tabstray/TabListItem.kt
```diff
@@ -0,0 +1,185 @@
...
+                shape = CircleShape,
+                backgroundColor = FirefoxTheme.colors.layerAccent,
+            ) {
+                Icon(
+                    painter = painterResource(id = R.drawable.mozac_ic_check),
+                    modifier = Modifier
+                        .matchParentSize()
+                        .padding(all = 8.dp),
+                    contentDescription = null,
+                    tint = colorResource(id = R.color.mozac_ui_icons_fill)
```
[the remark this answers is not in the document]
Amejia481: Yes, that was the goal!

Unfortunately, without adding it, it doesn't pick up the value from `mozac_ic_check`
image

app/src/main/java/org/mozilla/fenix/compose/tabstray/TabListItem.kt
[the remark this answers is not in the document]
Amejia481: Added!

app/src/main/java/org/mozilla/fenix/compose/tabstray/MediaImage.kt
```diff
@@ -0,0 +1,70 @@
...
+        }
+        PlaybackState.PLAYING -> {
+            R.drawable.media_state_pause to R.string.mozac_feature_media_notification_action_pause
+        }
+        else -> return
+    }
+
+    val drawable = AppCompatResources.getDrawable(LocalContext.current, icon)
+
+    Image(
```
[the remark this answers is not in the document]
Amejia481: https://github.com/mozilla-mobile/fenix/issues/25774

app/src/main/java/org/mozilla/fenix/compose/tabstray/TabListItem.kt
```diff
@@ -0,0 +1,185 @@
...
+            url = tab.content.url,
+            key = tab.id,
+            modifier = Modifier.size(width = 92.dp, height = 72.dp),
+            contentDescription = stringResource(id = R.string.mozac_browser_tabstray_open_tab),
+        )
+
+        if (isSelected) {
+            Card(
+                modifier = Modifier
+                    .size(size = 40.dp)
```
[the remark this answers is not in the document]
Amejia481: We already providing it with `backgroundColor = FirefoxTheme.colors.layerAccent,`

Amejia481:
Finally 🎉 !!!!

Issue #17130, opened 2020-12-18
Replace hard-coded strings in ETP Cookies setting


eliserichards:
Currently `settings.blockCookiesSelectionInCustomTrackingProtection` is defined as a string preference: 

```
val blockCookiesSelectionInCustomTrackingProtection by stringPreference(
        appContext.getPreferenceKey(R.string.pref_key_tracking_protection_custom_cookies_select),
        "social"
    )
```

It can be set to the following values: `all`, `social`, `unvisited`, `third-party`, or `none` (see [TrackingProtectionPolicyFactory](https://github.com/mozilla-mobile/fenix/blob/6962c967ee3828a570ce300ba2cd74a5816bfe5a/app/src/main/java/org/mozilla/fenix/components/TrackingProtectionPolicyFactory.kt#L66)). 

### Purpose
Remove hard coded strings when referencing this setting.

mcarare:
Closed in https://github.com/mozilla-mobile/fenix/pull/22748

PR #22748, opened 2021-12-08
For #17130: Replace hardcoded strings with string resources.

mcarare:
### Pull Request checklist

1. click on Show All Checks,


Mugurell (commented):
These changes looks good and will help have a more resilient code but I think some more changes are needed.
`Settings.kt` still contain the hardcoded preferences entry names like for [blockCookiesSelectionInCustomTrackingProtection](https://github.com/mozilla-mobile/fenix/blob/e6cae9b8007051df0a72931b3361654c4c018a22/app/src/main/java/org/mozilla/fenix/utils/Settings.kt#L577-L580) and I'm wondering if we shouldn't also refactor the hardcoded values in `TrackingProtectionPolicyFactoryTest` like in https://github.com/mozilla-mobile/fenix/blob/e6cae9b8007051df0a72931b3361654c4c018a22/app/src/test/java/org/mozilla/fenix/components/TrackingProtectionPolicyFactoryTest.kt#L246-L249

Mugurell:
> I think the tests should still have hardcoded values because it makes the tested case more obvious and also any changes in the string resources would have to be reflected in the tests so it would add extra proofing against non-intended changes. But as I can also find arguments for refactoring those also, I do not have a strong opinion on this.

The feature would work the same irrespective of what the string value is. The value is not part of the API. I see these rather as constants of the feature so reading them as such - everytime from one just one place without caring about the value would make more sense to me.
This also raises an important point: whether we should check the string value in unit tests and I think the general approach is that we use `getString(..)` for resources strings rather than using hardcoded values.

Mugurell (approved):
Thank you!

PR #22773, opened 2021-12-09
Update Android Components version; Fix breaking change in KeyProvider

grigoryk:
`key` was renamed to `getOrGenerateKey` and is now a `suspend`.

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardEditorFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardEditorFragment.kt
@@ -12,6 +12,9 @@ import android.view.View
 import androidx.lifecycle.lifecycleScope
 import androidx.navigation.fragment.findNavController
 import androidx.navigation.fragment.navArgs
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
 import org.mozilla.fenix.R
 import org.mozilla.fenix.SecureFragment
 import org.mozilla.fenix.databinding.FragmentCreditCardEditorBinding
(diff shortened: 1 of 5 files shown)


grigoryk:
Requires https://github.com/mozilla-mobile/android-components/pull/11371

app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardEditorFragment.kt
```diff
@@ -65,10 +69,14 @@ class CreditCardEditorFragment : SecureFragment(R.layout.fragment_credit_card_ed
 
         val binding = FragmentCreditCardEditorBinding.bind(view)
 
-        creditCardEditorState =
-            args.creditCard?.toCreditCardEditorState(storage) ?: getInitialCreditCardEditorState()
-        creditCardEditorView = CreditCardEditorView(binding, interactor)
-        creditCardEditorView.bind(creditCardEditorState)
+        CoroutineScope(Dispatchers.Main).launch {
```
gabrielluong, line 72: Sorry for the dumb question, I am wondering what the difference between `CoroutineScope` vs `lifecycleScope.launch(Dispatchers.Main)`. How do we decide which one to use in these contexts?

gabrielluong (approved):

app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardEditorFragment.kt
grigoryk: Yeah, good call - we do want `lifecycleScope` here. It's already bound to `Dispatchers.Main.immediate`, so just `lifecycleScope.launch { ... }` should be fine.

Since we have a fragment here and the work we may schedule is relevant in context of that fragment being around, tying it to the lifecycle of the fragment makes sense.

PR #22774, opened 2021-12-09
Fix beetmover destination targets

hneiva:
Try run: https://firefox-ci-tc.services.mozilla.com/tasks/OhEuaa8EQzG7ka0o2AJudQ/definition

### DIFF
--- a/taskcluster/fenix_taskgraph/util/scriptworker.py
+++ b/taskcluster/fenix_taskgraph/util/scriptworker.py
@@ -11,6 +11,7 @@ from datetime import datetime
 
 import jsone
 
+from ..release_promotion import read_version_file
 from taskgraph.util.memoize import memoize
 from taskgraph.util.schema import resolve_keyed_by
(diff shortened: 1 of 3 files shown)


taskcluster/fenix_taskgraph/util/scriptworker.py
```diff
@@ -221,15 +225,13 @@ def generate_beetmover_artifact_map(config, job, **kwargs):
...
-                    locale=locale,
-                    path_platform=platform,
-                    version=config.params["version"],
+                    locale=locale
                 )
 
             # This format string should ideally be in the configuration file,
             # but this would mean keeping variable names in sync between code + config.
             destinations = [
-                "{s3_bucket_path}/{dest_path}/{locale_prefix}/{filename}".format(
```
escapewindow, line 232: I was thinking about this; this is probably because we don't do single locale repacks for Android :)

escapewindow (approved):
Thank you! We might want to verify if the Fenix team wants a different version number or path structure when this goes green.

Issue #20513, opened 2021-07-26
[Bug]: Spike in `search.default_engine.submission_url` errors (Invalid Overflow)


badboy:
We're seeing a spike in errors for the `search.default_engine.submission_url` metric since 2021-07-11.
This applies to the stable release.
The error is a "Invalid Overflow", meaning values stored are [longer than 100 characters](https://mozilla.github.io/glean/book/reference/metrics/string.html#limits).
See [this query](https://sql.telemetry.mozilla.org/queries/74333#185794).

Data longer than 100 characters is truncated at 100 characters, so you will still see data, but truncated.
Did anything about the values in use change recently?

kbrosnan:
@eliserichards can you have a look or find someone to delegate to?

badboy:
[Query focused on that metric only](https://sql.telemetry.mozilla.org/queries/81579?p_appid=org_mozilla_firefox&p_metric=%5B%22search.default_engine.submission_url%22%5D#202286)

There's still an increase.
Increase is less visible over the 30 day span, but it's very visible over 90 days.
Given that data is coming in truncated I'm not too worried about the data being useless.
Is this metric part of any dashboard or queries that are looked at regurlarly?

We do have a new [URL metric type](https://mozilla.github.io/glean/book/reference/metrics/url.html) that we might switch this metric too (though it's not yet possible to use that on Android).

eliserichards:
It would be fantastic to switch to the URL metric type once that's available!

mcarare:
Closed in https://github.com/mozilla-mobile/fenix/pull/22787.
Verified in https://github.com/mozilla-mobile/fenix/issues/22879.

Issue #1607, opened 2019-04-12
FNX-52 ⁃ [Telemetry] Handle custom search engines, do not leak user-added engines


liuche:
Owner: Product Manager

This is a follow-up to the telemetry that was added in #1194 and #800. When users gain the ability to add Custom Search Engines, we need to update that telemetry so the new search engines are obfuscated (and sent as "custom" or some other non-user-generated string).

#### Description & Product Manager / Data Scienties User Story
* As a User, I don't want the names or search urls of custom search engines I add to be collected.

#### Hypothesis
* It will be sufficient to know that users are using "custom" engines, and details of the engine don't need to be sent.

#### What questions will you answer with this data?
Whether users are using custom search engines

#### Why does Mozilla need to answer these questions? Are there benefits for users? Do we need this information to address product or business requirements?
Filing pre-emptively to track future work

#### What probes (suggested, if applicable)
* Send "custom" as search engine name, shortcut, or url if user is using that, instead of the name of the "bundled" search engines

### Acceptance Criteria (Added by PM)
* Event pings can be queried via re:dash
* We are sending telemetry events for the actions listed in the requirements
* We have documented the telemetry
* We have asked a data steward to [review](https://github.com/mozilla/data-review/blob/master/request.md) the telemetry
* NOT an AC: Data science to create dashboard or further graphs (this will be a separate issue, this issue is only about hooking up the events described and that they can be queried in re-dash)

liuche:
@vesta0 this isn't an MVP, but there are a couple of search engine telemetry probes that now exist, and I just want to make sure that when custom search engines are added this reminder is here that those probes need to be updated to not send things like "search engine name".

AndiAJ:
Hi, I've just checked this matter on the latest **Nightly Build #20060605 from 01/06** using a Google Pixel 3a (Android 9).

### Shortcuts
❓ Search - Ping 066a5899-af3e-43a5-b10b-c067ce128082
✔️ Baseline - Ping 034d2eaa-87ff-4294-8403-5e4c46d371e6
❓ Suggestion - Ping 9a383cba-b6bf-4359-a64d-7a2edffa73ef
✔️ Baseline - Ping 7ddd5ccb-7cab-4072-b80a-8657fbb35a1c

### Set Custom Search Engine as Default
✔️ Search - Ping 62aa8920-ef77-4d3a-8092-9afd7ac1e049
[...]

PR #22787, opened 2021-12-10
For #20513: Use UrlMetricType for search engine submission url.

mcarare:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/metrics.yaml
+++ b/app/metrics.yaml
@@ -1834,11 +1834,11 @@ search.default_engine:
       - android-probes@mozilla.com
       - erichards@mozilla.com
     expires: never
-  submission_url:
-    type: string
+  search_url:
+    type: url
     lifetime: application
     description: |
-      If the search engine is pre-loaded with Fenix this value will be he base
+      If the search engine is pre-loaded with Fenix this value will be the base
       URL we use to build the search query for the search engine. For example:
       https://mysearchengine.com/?query=%s. If it's a custom search engine
       (defined: https://github.com/mozilla-mobile/fenix/issues/1607) the value
@@ -1847,6 +1847,7 @@ search.default_engine:
       - metrics
     bugs:
       - https://github.com/mozilla-mobile/fenix/issues/800
+      - https://github.com/mozilla-mobile/fenix/issues/20513
     data_reviews:
       - https://github.com/mozilla-mobile/fenix/pull/1606
       - https://github.com/mozilla-mobile/fenix/pull/5216
--- a/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
+++ b/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
@@ -647,7 +647,7 @@ open class FenixApplication : LocaleAwareApplication(), Provider {
                 SearchDefaultEngine.apply {
                     code.set(searchEngine.id)
                     name.set(searchEngine.name)
-                    submissionUrl.set(searchEngine.buildSearchUrl(""))
+                    searchUrl.set(searchEngine.buildSearchUrl(""))
                 }
             }
         }
--- a/app/src/test/java/org/mozilla/fenix/FenixApplicationTest.kt
+++ b/app/src/test/java/org/mozilla/fenix/FenixApplicationTest.kt
@@ -185,6 +185,6 @@ class FenixApplicationTest {
         // not mock most of the objects telemetry is collected from.
         assertFalse(SearchDefaultEngine.code.testHasValue())
         assertFalse(SearchDefaultEngine.name.testHasValue())
-        assertFalse(SearchDefaultEngine.submissionUrl.testHasValue())
+        assertFalse(SearchDefaultEngine.searchUrl.testHasValue())
     }
 }


mcarare:
@badboy , @travis79 

The [documentation](https://mozilla.github.io/glean/book/reference/metrics/url.html#note-on-data_sensitivity-of-url-metrics) for this metric type states that the data sensitivity category for this should be 3 or 4. I guess changing it to 3 (web_activity) in this PR would also need a new data review (and maybe additional approvals).

On the other hand, although it is a URL, this data is related only to the default search engine used, not to a specific website the user has visited. Having this in mind, should we just keep the previous category for data sensitivity? 

Thank you!

app/metrics.yaml
```diff
@@ -1835,7 +1835,7 @@ search.default_engine:
       - erichards@mozilla.com
     expires: never
   submission_url:
-    type: string
+    type: url
```
badboy, line 1838: Unfortunately changing the type means it will require a new name (that's a shortcoming of our current pipeline)

badboy (changes requested):

badboy:
> @badboy , @travis79
> 
> The [documentation](https://mozilla.github.io/glean/book/reference/metrics/url.html#note-on-data_sensitivity-of-url-metrics) for this metric type states that the data sensitivity category for this should be 3 or 4. I guess changing it to 3 (web_activity) in this PR would also need a new data review (and maybe additional approvals).
> 
> On the other hand, although it is a URL, this data is related only to the default search engine used, not to a specific website the user has visited. Having this in mind, should we just keep the previous category for data sensitivity?
> 
> Thank you!

I will leave this to @travis79 as a data steward to answer.

travis79:
The documentation for the UrlMetricType does suggest that most usages of the type would likely be Category 3 or 4 data, since there is the possibility of recording some PII in something such as a query string. In this specific case, since we are only recording the search engine, and that should be limited to a somewhat known or expected set of URLs, this should be okay since it's just changing the underlying type that the data is recorded in, and not the data itself. If there was a risk of collecting Category 3 data from something such as a custom search engine defnintion, then we would already have encountered this while recording it as a StringMetricType. Looking at the data it appears that the query portion is properly being stripped so I don't see where we need to change the data-collection category of this metric, it can remain at Category 2, Technical Data. As for a fresh data-review? Since we are not expanding the collection I feel the original data-review stands and we can review it again when it next comes up for expiration/extension.

mcarare:
Thank you both for your input!

app/metrics.yaml
```diff
@@ -1834,8 +1834,8 @@ search.default_engine:
       - android-probes@mozilla.com
       - erichards@mozilla.com
     expires: never
-  submission_url:
-    type: string
+  search_url:
+    type: url
     lifetime: application
     description: |
       If the search engine is pre-loaded with Fenix this value will be he base
```
badboy, line 1841:
```suggestion
      If the search engine is pre-loaded with Fenix this value will be the base
```

app/metrics.yaml
```diff
@@ -1847,6 +1847,7 @@ search.default_engine:
       - metrics
```
badboy, line 1847: For the description above: What value is used in case of a custom search engine? If it's really `"custom"` then that will result in no value but errors being recorded. The URL metric type does validate its input to be a sort-of valid URL (starts with a scheme like `https`, is ascii-only  plus a few extra characters).

badboy (changes requested):

app/metrics.yaml
mcarare: Yes, I have noticed that in the documentation. According to [this](https://github.com/mozilla-mobile/fenix/issues/1607), we do not want to send the URL of custom engines added by the user. If that is still the case I thought of 3 ways we could handle this :
 - just let it resolve to `invalid_value` and keep this in mind when analyzing data (what the current code does)
 - do not send any metric for custom engines (I am not sure we want this, we would lose related data)
 - send a custom URL agreed upon beforehand. Something like `"https://custom"`
What do you think?

travis79: @mcarare I think that a "magic" value like `https://custom` would work, just be sure and check with the data-science folks or the stakeholders looking at the data to ensure that whatever tools they are using to analyze this is prepared to handle this correctly. My concern being that downstream tools that are consuming this may need updates in order to properly display this.

Standard8: Is it not possible to send a null value? It would seem strange that you wouldn't be allowed to send an empty url.

Standard8:
As I commented in slack, we would like to get to reporting submission urls the same way as desktop, if it is more appropriate to move this out to a separate issue, please do.

Desktop currently records the url if:

* The search engine is one of the ones Firefox provides.
* The hostname is the same domain as Firefox provides.
* The hostname matches one of the well known search domains.

[Code is here](https://searchfox.org/mozilla-central/rev/dc323d0d9a3b722ca8ff0d1b2b752e5bd00dab9b/toolkit/components/search/SearchService.jsm#2351-2419).
[Original bug that added the code](https://bugzilla.mozilla.org/show_bug.cgi?id=1164159).

mcarare:
Upon testing this, it seems that the name and the submission URL are not replaced with "custom" for user-added engines.
Could someone with access to running queries for this data source confirm that we have such values recorded?

Having that said, I would like to merge this with just the metric type change and investigate aligning with desktop and properly updating metric description in another [issue](https://github.com/mozilla-mobile/fenix/issues/22879).

rocketsroger (approved):
With the follow up issue https://github.com/mozilla-mobile/fenix/issues/22879, I think this change is good to go.  @badboy I'll leave the final call up to you.  Thanks

PR #22796, opened 2021-12-10
Change beetmover destination paths to match firefox

hneiva:
See [bug 1614763](https://bugzilla.mozilla.org/show_bug.cgi?id=1614763) for details

Beta staging run: https://firefox-ci-tc.services.mozilla.com/tasks/VB6zFyjzTw-GKYvjBiqVDg/runs/0/logs/public/logs/live_backing.log

Nightly staging run: https://firefox-ci-tc.services.mozilla.com/tasks/cGdpEKeIQQWkbAoXf4w9RA/runs/0/logs/public/logs/live_backing.log

### DIFF
--- a/taskcluster/fenix_taskgraph/manifests/fenix_candidates.yml
+++ b/taskcluster/fenix_taskgraph/manifests/fenix_candidates.yml
@@ -66,22 +66,22 @@ mapping:
         description: "Android package for arm64-v8a"
         pretty_name: fenix-${version}.${locale}.android-arm64-v8a.apk
         destinations:
-            - ${dated_path}-fenix-${version}-android-arm64-v8a
+            - ${folder_prefix}fenix-${version}-android-arm64-v8a
     armeabi-v7a/target.apk:
         <<: *default
         description: "Android package for armeabi-v7a"
         pretty_name: fenix-${version}.${locale}.android-armeabi-v7a.apk
         destinations:
-            - ${dated_path}-fenix-${version}-android-armeabi-v7a
+            - ${folder_prefix}fenix-${version}-android-armeabi-v7a
     x86/target.apk:
         <<: *default
         description: "Android package for x86"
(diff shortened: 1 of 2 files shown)


taskcluster/fenix_taskgraph/manifests/fenix_candidates.yml
```diff
@@ -66,22 +66,22 @@ mapping:
...
         destinations:
-            - ${dated_path}-fenix-${version}-android-armeabi-v7a
+            - ${folder_prefix}fenix-${version}-android-armeabi-v7a
     x86/target.apk:
         <<: *default
         description: "Android package for x86"
         pretty_name: fenix-${version}.${locale}.android-x86.apk
         destinations:
-            - ${dated_path}-fenix-${version}-android-x86
+            - ${folder_prefix}fenix-${version}-android-x86
```
jcristau, line 81: We should probably either make it so `version.txt` is correct on nightly, or change the ${version} subst to use something else, so nightly filenames don't say "beta"?

escapewindow: I'm guessing the fix may involve manually? automatically? bumping `main` to an alpha version, and bumping it to a beta version when we branch to go to beta. If these are off the same branch, then we may need another solution.

jcristau: nightly is off the `main` branch, beta and release are from a `releases_vXX.0` branch, as far as I know.

hneiva: Should I just create an issue for it?

ahal (approved):
Thanks looks good!

I agree that the fact nightly versions contain the string "beta" is a separate issue from this PR (or anything beetmover related even).

Issue #22383, opened 2021-11-09
IllegalStateException: Composed into the View which doesn't propagate ViewTreeLifecycleOwner!


pocmo:
https://sentry.prod.mozaws.net/operations/firefox-nightly/issues/13334589/?referrer=github_plugin

```
IllegalStateException: Composed into the View which doesn't propagate ViewTreeLifecycleOwner!
    at androidx.compose.ui.platform.AndroidComposeView.onAttachedToWindow(AndroidComposeView.android.kt:20)
    at android.view.View.dispatchAttachedToWindow(View.java:20190)
    at android.view.ViewGroup.dispatchAttachedToWindow(ViewGroup.java:3463)
    at android.view.ViewGroup.dispatchAttachedToWindow(ViewGroup.java:3470)
    at android.view.ViewRootImpl.performTraversals(ViewRootImpl.java:2138)
...
(13 additional frame(s) were not displayed)

Composed into the View which doesn't propagate ViewTreeLifecycleOwner!
```

pocmo:
The stack trace doesn't contain any of our code. But it happens enough that it may be worth investigating - and maybe filing against Compose.

It shows up in our list of top crashes in Nightly too.

https://crash-stats.mozilla.org/signature/?product=Fenix&version=96.0a1&signature=java.lang.IllegalStateException%3A%20at%20androidx.compose.ui.platform.AndroidComposeView.onAttachedToWindow%28AndroidComposeView.android.kt%3A20%29&date=%3E%3D2021-11-02T09%3A54%3A00.000Z&date=%3C2021-11-09T09%3A54%3A00.000Z&_columns=date&_columns=product&_columns=version&_columns=build_id&_columns=platform&_columns=reason&_columns=address&_columns=install_time&_columns=startup_crash&_sort=-date&page=1#summary

pocmo:
https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/ui/ui/src/androidMain/kotlin/androidx/compose/ui/platform/AndroidComposeView.android.kt;l=968;drc=6eace1ebfa98add8b00ee1b312204a2efd4140e6

rvandermeulen:
This is showing up in the Fenix 95 topcrash list: https://bugzilla.mozilla.org/show_bug.cgi?id=1747118

mcarare:
This is quite easily reproducible. All you have to do is fling the home screen top to bottom and back a few times.

mcarare:
[...]

Issue #19942, opened 2021-06-10
Device names sticky in sync list


topotropic:
Make the device names sticky in the synced tabs view so that when people scroll they see the device name until another device is listed.

jonalmeida:
Eng notes: We can do this with an `ItemDecoration` that we can extend from our existing `SyncedTabsTitleDecoration`

gabrielluong:
Designs https://www.figma.com/file/hJfV1KGZuOK7OYr29pooL5/Lists?node-id=69%3A6540

PR #22798, opened 2021-12-11
For #21900 - Converted the Tabs Tray Synced Tabs to Compose

### LINKED ISSUE #21900: Move Tabs Tray to compose: Synced Tabs list
Synced Tabs list - SyncedTabsTrayLayout, SyncedTabsAdapter
Note:    We can do this async; doesn't have to wait for the normal/private tab list.

In the SyncedTabsFeature, we do have a `SyncedTabsView.listener` which is tightly coupled to an instance for `SyncedTabsView` that probably needs to be decoupled so interactions can be invoked in a Composable which does not require 'View' classes.

From meta #21318 

https://mozilla-hub.atlassian.net/browse/FNXV2-18288

MozillaNoah:
Fixes #21900 

### Pull Request checklist

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/compose/Text.kt
```diff
@@ -16,21 +18,28 @@ import org.mozilla.fenix.theme.FirefoxTheme
  * @param text The text to be displayed.
  * @param modifier [Modifier] to apply to this layout node.
  * @param fontSize The size of glyphs to use when painting the text.
+ * @param fontFamily The font family to be used when rendering the text. See [TextStyle.fontFamily].
```
gabrielluong, line 21: I am looking at the 26 files changed, and thinking let's rip some of this out into separate PRs and land smaller commits. So, we aren't always reviewing 26 files changes. 😱

The changes here are an easy candidate.

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabsComposables.kt
```diff
@@ -0,0 +1,321 @@
...
+                }
+            )
+        }
+    }
+}
+
+/**
+ * The various types of list items that can be found in a [SyncedTabsList].
+ */
+sealed class SyncedTabsListItem {
```
gabrielluong, line 265: Let's move this into its own file.

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabsComposables.kt
```diff
@@ -0,0 +1,321 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
gabrielluong, line 1: Let's drop the `Composables` suffix in the file name. Looking at the sample Compose apps, this is non-standard naming.

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabsComposables.kt
```diff
@@ -0,0 +1,321 @@
...
+import org.mozilla.fenix.R
+import org.mozilla.fenix.components.components
+import org.mozilla.fenix.compose.PrimaryText
+import org.mozilla.fenix.compose.SecondaryText
+import org.mozilla.fenix.ext.toShortUrl
+import org.mozilla.fenix.theme.FirefoxTheme
+import mozilla.components.browser.storage.sync.Tab as SyncTab
+
+@Composable
+fun SyncedTabsList(syncedTabs: List<SyncedTabsListItem>, listener: SyncedTabsView.Listener) {
```
gabrielluong, line 60: Remember to add kDocs. :)

gabrielluong:
Apologizes for the early drive by.

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabsComposables.kt
```diff
@@ -0,0 +1,321 @@
...
+    /**
+     * A placeholder for a device that has no tabs synced.
+     */
+    data class NoTabs(val deviceId: String) : SyncedTabsListItem()
+
+    /**
+     * A message displayed if an error was encountered.
+     */
+    data class Error(
+        val descriptionResId: Int,
```
gabrielluong, line 286: Rather than storing the descriptionResId, maybe it would be nicer to store the ErrorType and then use the extension functions to derive the stringResId in our composable functions. Just a thought if you are open to entertaining it.

MozillaNoah: This makes sense

app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayFragment.kt
```diff
@@ -58,11 +63,19 @@ import org.mozilla.fenix.tabstray.ext.collectionMessage
...
+import org.mozilla.fenix.tabstray.syncedtabs.toComposeList
+import org.mozilla.fenix.tabstray.syncedtabs.toSyncedTabsListItem
 import org.mozilla.fenix.utils.allowUndo
 import kotlin.math.max
 
 @Suppress("TooManyFunctions", "LargeClass")
-class TabsTrayFragment : AppCompatDialogFragment() {
+class TabsTrayFragment :
+    AppCompatDialogFragment(),
+    SyncedTabsView,
```
jonalmeida, line 76: Let's implement this in an "integration" class.

It's to ask when adding things to a root class: what does Synced Tabs have to do with the TabsTray**Fragment**? 

When we add more things to the fragment, we start creating objects that do too many things and making testing harder.

app/src/main/java/org/mozilla/fenix/tabstray/TrayPagerAdapter.kt
```diff
@@ -85,18 +81,24 @@ class TrayPagerAdapter(
...
+     * Until [TrayPagerAdapter] is replaced with a Compose implementation, some sort of adapter
+     * needs to be passed to [SyncedTabsPageViewHolder] for the abstract class contract.
+     * For more details: https://github.com/mozilla-mobile/fenix/issues/21318
+     */
     override fun onBindViewHolder(viewHolder: AbstractPageViewHolder, position: Int) {
         val adapter = when (position) {
             POSITION_NORMAL_TABS -> normalAdapter
             POSITION_PRIVATE_TABS -> privateAdapter
-            POSITION_SYNCED_TABS -> syncedTabsAdapter
+            POSITION_SYNCED_TABS -> privateAdapter
```
jonalmeida, line 101: We could just take out this line and the throw below, instead of adding side-effects.

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabs.kt
```diff
@@ -0,0 +1,339 @@
...
+)
+
+/**
+ * Helper function to create a [SyncedTabsListItem.Tab] for previewing.
+ */
+private fun generateFakeTab(tabName: String, tabUrl: String): SyncTab = SyncTab(
+    listOf(TabEntry(tabName, tabUrl, null)),
+    0,
+    0L
+)
```
jonalmeida, line 339: Could we put these in the test package itself? Does it need to be in our public packages (even though the have internal/private modifiers)?

jonalmeida:
Screen Shot 2021-12-15 at 11 20 20 AM

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabs.kt
MozillaNoah: Can stuff in the test package be accessed here? Legitimately do not know haha

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabs.kt
```diff
@@ -0,0 +1,339 @@
...
+        letterSpacing = 0.04.sp,
+        maxLines = 1,
+        modifier = Modifier
+            .padding(16.dp, 16.dp, 8.dp, 7.dp)
+            .fillMaxWidth()
+    )
+}
+
+@Composable
+@Preview
```
gabrielluong, line 110: Let's move all the @Preview towards the end of the file.

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabs.kt
jonalmeida: I only noticed the references to the tests initially 🙂  - I don't have strong opinions about adding fake helpers to production code, but it _feels_ like code smell.

app/src/test/java/org/mozilla/fenix/tabstray/FloatingActionButtonBindingTest.kt
```diff
@@ -134,6 +134,6 @@ class FloatingActionButtonBindingTest {
         verify(exactly = 0) { actionButton.hide() }
         verify(exactly = 1) { actionButton.setText(R.string.sync_syncing_in_progress) }
         verify(exactly = 2) { actionButton.setIconResource(R.drawable.ic_fab_sync) }
-        verify(exactly = 2) { actionButton.contentDescription = any() }
+        verify(exactly = 4) { actionButton.contentDescription = any() }
```
jonalmeida, line 137: Why are there 2 new sets of the content description now?

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabs.kt
```diff
@@ -0,0 +1,339 @@
...
+
+/**
+ * Synced tab list item UI
+ *
+ * @param tabTitleText The tab's display text.
+ * @param url The tab's URL.
+ * @param onClick The click handler when this synced tab is clicked.
+ */
+@Composable
+fun SyncedTabsTabItem(tabTitleText: String, url: String, onClick: () -> Unit) {
```
jonalmeida, line 117: titles can be empty strings, in which case we show the full url (until MAX_URI_LENGTH). I don't see that case handled.

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabsListItem.kt
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * A device header for displaying a synced device.
+     */
+    data class Device(val displayName: String, val id: String) : SyncedTabsListItem()
+
+    /**
+     * A tab that was synced.
+     */
+    data class Tab(val tab: mozilla.components.browser.storage.sync.Tab) : SyncedTabsListItem()
```
jonalmeida, line 23: nit: import aliases can make this easier to read.

Example..

```suggestion
import mozilla.components.browser.storage.sync.Tab as SyncTab
    data class Tab(val tab: SyncTab) : SyncedTabsListItem()
```

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabs.kt
```diff
@@ -0,0 +1,339 @@
...
+}
+
+/**
+ * Text header for sections of synced tabs
+ *
+ * @param deviceName The name of the user's device connected that has synced tabs.
+ */
+@Composable
+fun SyncedTabsDeviceItem(deviceName: String) {
+    PrimaryText(
```
jonalmeida, line 97: nit: in the designs, we needed to add a horizontal line above the device name. I think it's fine to leave out if you're planning on picking up https://github.com/mozilla-mobile/fenix/issues/19942 next.

jonalmeida (commented):
Great! Just some nits and questions, but it looks good.

I'll leave the final review for gl though since I only focused on Synced Tabs feature and not the Compose code.

app/src/test/java/org/mozilla/fenix/tabstray/FloatingActionButtonBindingTest.kt
MozillaNoah: Good question. I didn't really investigate this when refactoring, but I'll check if it can just be the one check.

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabs.kt
MozillaNoah: That logic wasn't in the now-deleted SyncedTabsViewHolder. I'm happy to add this in though

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabs.kt
MozillaNoah: Good catch! I noticed that without sticky headers it did become hard to tell one section from the other. My intent was to immediately do the sticky header issue next.

app/src/test/java/org/mozilla/fenix/tabstray/FloatingActionButtonBindingTest.kt
jonalmeida: Please do. It was there at one point, and it might have been removed.

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabs.kt
jonalmeida: Please do. It was there at one point, and it might have been removed.

app/src/test/java/org/mozilla/fenix/tabstray/FloatingActionButtonBindingTest.kt
MozillaNoah: Update: I misunderstood the question originally. The content description wasn't being set for the "Sync" and "Syncing" states of the FAB previously, so with me setting that now in FloatingActionButtonBinding line 73, it now gets called when the TabsTray goes to synced tabs but also when the TabsTrayStore.syncing variable is updated.

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabs.kt
MozillaNoah: Update: it looks like this is the behavior for inactive tabs but it never existed in synced tabs. I'll add it here now.

MozillaNoah:
Most recent push was to fix some of the previews. Fun fact: the Compose previews don't like using stuff in the Components class, so the preview builds error out since Components is dependent on FenixApplication. So the most recent changes were to move the URL parsing to make it text-friendly out of the Composables.

app/src/test/java/org/mozilla/fenix/tabstray/FloatingActionButtonBindingTest.kt
jonalmeida: Ah ha, great catch! Surprising that it isn't 3 though.

app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayFragment.kt
```diff
@@ -60,12 +60,15 @@ import org.mozilla.fenix.tabstray.ext.collectionMessage
...
 import org.mozilla.fenix.tabstray.ext.showWithTheme
+import org.mozilla.fenix.tabstray.syncedtabs.SyncedTabsIntegration
 import org.mozilla.fenix.theme.ThemeManager
 import org.mozilla.fenix.utils.allowUndo
 import kotlin.math.max
 
 @Suppress("TooManyFunctions", "LargeClass")
-class TabsTrayFragment : AppCompatDialogFragment() {
+class TabsTrayFragment :
+    AppCompatDialogFragment() {
```
gabrielluong, line 70: nit: This change seems unnecessary and not something we see a lot of in the codebase.

app/src/main/res/values/strings.xml
```diff
@@ -199,7 +199,7 @@
     <!-- Browser menu toggle that installs a Progressive Web App shortcut to the site on the device home screen. -->
     <string name="browser_menu_install_on_homescreen">Install</string>
     <!-- Menu option on the toolbar that takes you to synced tabs page-->
-    <string name="synced_tabs">Synced tabs</string>
+    <string name="synced_tabs" moz:removedIn="97" tools:ignore="UnusedResources">Synced tabs</string>
```
gabrielluong, line 202: Needs to be bumped to 98

app/src/main/java/org/mozilla/fenix/tabstray/viewholders/SyncedTabsPageViewHolder.kt
```diff
@@ -5,32 +5,40 @@
...
+import org.mozilla.fenix.databinding.ComposeSyncTabsTrayLayoutBinding
 import org.mozilla.fenix.tabstray.TabsTrayStore
+import org.mozilla.fenix.tabstray.syncedtabs.SyncedTabsList
+import org.mozilla.fenix.theme.FirefoxTheme
 
 class SyncedTabsPageViewHolder(
     containerView: View,
-    private val tabsTrayStore: TabsTrayStore
+    private val tabsTrayStore: TabsTrayStore,
+    private val listener: SyncedTabsView.Listener
```
gabrielluong, line 20: Wondering if this should just be `TabClickDelegate` over its interface?

app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayFragment.kt
MozillaNoah: I think this was leftover when TabsTrayFragment had a bunch of other interfaces. I'll clean this up

app/src/main/java/org/mozilla/fenix/tabstray/viewholders/SyncedTabsPageViewHolder.kt
MozillaNoah: I don't recall why i did it this way (possibly a holdover from the original code). I think it makes sense to have it be `TabClickDelegate`.

app/src/main/java/org/mozilla/fenix/sync/ext/ErrorType.kt
```diff
@@ -1,40 +0,0 @@
...
-/**
- * Converts the error type to the appropriate matching string resource for displaying to the user.
- */
-fun ErrorType.toStringRes() = when (this) {
-    ErrorType.MULTIPLE_DEVICES_UNAVAILABLE -> R.string.synced_tabs_connect_another_device
-    ErrorType.SYNC_ENGINE_UNAVAILABLE -> R.string.synced_tabs_enable_tab_syncing
-    ErrorType.SYNC_UNAVAILABLE -> R.string.synced_tabs_sign_in_message
-    ErrorType.SYNC_NEEDS_REAUTHENTICATION -> R.string.synced_tabs_reauth
-    ErrorType.NO_TABS_AVAILABLE -> R.string.synced_tabs_no_tabs
-}
```
gabrielluong, line 22: I am also seeing some changes related to this in https://github.com/mozilla-mobile/fenix/pull/23168. @jonalmeida any particular reasoning we are using the StringRes over the string.

jonalmeida: So we can map the error to the ID for the string until we get to a piece of UI that can get the correct string (context.getString).

This is where it's used: https://searchfox.org/mozilla-mobile/rev/cf261343f08aea1a6a07a018e5942e30c54f7354/fenix/app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabsTrayLayout.kt#95

app/src/test/java/org/mozilla/fenix/tabstray/TabsTrayStoreReducerTest.kt
```diff
@@ -69,4 +70,18 @@ class TabsTrayStoreReducerTest {
 
         assertEquals(expectedState, resultState)
     }
+
+    @Test
+    fun `WHEN UpdateSyncedTabs THEN synced tabs are added`() {
```
gabrielluong, line 75: @jonalmeida Hm.. it's actually interesting to see that we have a Reducer and Store test separately. Usually I see we use the Store test to test the underlying Reducer. Is there any benefit from having them separate like this? I can imagine we end up picking and choosing which actions are dispatched and resulting state in different files.

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabs.kt
```diff
@@ -0,0 +1,346 @@
...
+                    )
+                    Spacer(modifier = Modifier.width(8.dp))
+                    Text(
+                        text = stringResource(id = R.string.synced_tabs_sign_in_button),
+                        color = FirefoxTheme.colors.textInverted,
+                        fontSize = 14.sp,
+                        fontFamily = FontFamily(Font(R.font.metropolis_semibold)),
+                        maxLines = 2,
+                        modifier = Modifier.align(Alignment.CenterVertically)
+                    )
```
gabrielluong, line 211:
```suggestion
                    Icon(
                        painter = painterResource(id = R.drawable.ic_sign_in),
                        contentDescription = null,
                        tint = FirefoxTheme.colors.textInverted,
                    )
                    
                    Spacer(modifier = Modifier.width(8.dp))
                    
                    Text(
                        text = stringResource(id = R.string.synced_tabs_sign_in_button),
                        color = FirefoxTheme.colors.textInverted,
                        fontSize = 14.sp,
                        fontFamily = FontFamily(Font(R.font.metropolis_semibold)),
                        maxLines = 2,
                        modifier = Modifier.align(Alignment.CenterVertically)
                    )
```

One theme here that I persuade you on is to use new lines to separate these compose components and logical blocks of code. This just helps create a visual separation when reading the code to know when a thing ends.

gabrielluong: I would also consider looking at the rest file and adding some new lines where it makes sense to not have everything so squished together.

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabs.kt
```diff
@@ -0,0 +1,346 @@
...
+                        tint = FirefoxTheme.colors.textInverted,
+                    )
+                    Spacer(modifier = Modifier.width(8.dp))
+                    Text(
+                        text = stringResource(id = R.string.synced_tabs_sign_in_button),
+                        color = FirefoxTheme.colors.textInverted,
+                        fontSize = 14.sp,
+                        fontFamily = FontFamily(Font(R.font.metropolis_semibold)),
+                        maxLines = 2,
+                        modifier = Modifier.align(Alignment.CenterVertically)
```
gabrielluong, line 210: Can we reorder the `modifier` to be below `text`? I usually just follow the parameter ordering that is defined in `Text`, but otherwise, I think you went with alphabetical ordering except you prioritized `text` as the first parameter since that's the most important parameter. Only calling this out because I want to make sure we are somewhat consistent and not investing a new ordering of listing out compose parameters that is not reproducible by others.

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabs.kt
MozillaNoah: I don't think I was even cognizant of doing this in the Compose code haha. Thanks for bringing attention to this; I'll definitely do this going forward and fix it here.

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabs.kt
```diff
@@ -0,0 +1,346 @@
...
+        width = 2.dp.toPx()
+        dashLength = 4.dp.toPx()
+    }
+    // As of December 2021, there is not an official way to get a rounded rectangle border with
+    // dashes using Compose. This is a next-best attempt with hard corners.
+    val stroke = Stroke(
+        width = width,
+        pathEffect = PathEffect.dashPathEffect(floatArrayOf(dashLength, dashLength), 0f)
+    )
+    val dashColor = FirefoxTheme.colors.borderDivider
```
gabrielluong, line 163: I would inline this since it's only used once

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabs.kt
```diff
@@ -0,0 +1,346 @@
...
+    Box(
```
gabrielluong, line 164: nit; add a new line before

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabs.kt
```diff
@@ -0,0 +1,346 @@
...
+@Composable
+fun SyncedTabsNoTabsItem() {
+    SecondaryText(
+        text = stringResource(id = R.string.synced_tabs_no_open_tabs),
+        fontSize = 12.sp,
+        maxLines = 1,
+        modifier = Modifier
+            .padding(PaddingValues(16.dp, 7.dp))
+            .fillMaxWidth()
+    )
```
gabrielluong, line 230: nit: same comment about parameter ordering, but I won't call anymore out. You might want to go through your code and make your own call.

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabs.kt
```diff
@@ -0,0 +1,346 @@
...
+
+@Composable
+@Preview
+private fun SyncedTabsNoTabsItemPreview() {
+    FirefoxTheme {
+        Box(Modifier.background(FirefoxTheme.colors.layer2)) {
+            SyncedTabsNoTabsItem()
+        }
+    }
+}
```
gabrielluong, line 290: Maybe just join them all into one Column with some Spacer?

gabrielluong: Mostly thinking about how to make this file smaller. Seeing that all the @Preview are defined very similarly, I think it could make sense to just display it in one view. You can imagine scenarios where you might be trying to see how it looks like with the dark theme, but adding that parameter to each @Preview compare to just one might be annoying.

MozillaNoah: I think this makes sense. I'll condense this and reconsider previews and their brevity going forward.

app/src/test/java/org/mozilla/fenix/sync/ext/ErrorTypeKtTest.kt
```diff
@@ -36,37 +38,22 @@ class ErrorTypeKtTest {
     }
 
     @Test
-    fun `get error item`() {
+    fun `get SyncedTabsListItem_Error item`() {
```
gabrielluong, line 41: The `SyncedTabsListItem_Error` naming threw me off to me while just looking at the `toSyncedTabsListItem` call and I get you meant `SyncedTabsListItem.Error`

I would almost prefer seeing the 'GIVEN WHEN THEN' structure we use: `WHEN toSyncedTabsListItem is called THEN return the correct Error item` It would probably also help that we test that `errorItem.errorType` also returns the correct `SyncedTabsView.ErrorType`

app/src/main/res/layout/compose_sync_tabs_tray_layout.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<androidx.compose.ui.platform.ComposeView
```
gabrielluong, line 5: Wondering if we could do without having this? Mostly, recalling this from the wallpaper setting PR.

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabs.kt
```diff
@@ -0,0 +1,346 @@
...
+import org.mozilla.fenix.R
+import org.mozilla.fenix.compose.PrimaryText
+import org.mozilla.fenix.compose.SecondaryText
+import org.mozilla.fenix.theme.FirefoxTheme
+import mozilla.components.browser.storage.sync.Tab as SyncTab
+
+/**
+ * Top-level list UI for displaying Synced Tabs in the Tabs Tray.
+ *
+ * @param syncedTabs [SyncedTabsListItem] The tab UI items to be displayed.
```
gabrielluong, line 60: I think we should drop `[SyncedTabsListItem]` there's no point repeating the type here since you can find it below and if that was the intent, I think we would also add `List<SyncedTabsListItem>`. This also runs into the consistency issue with the rest of the project since we don't annotate our @param in this style.

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabsIntegration.kt
```diff
@@ -0,0 +1,83 @@
...
+import mozilla.components.feature.syncedtabs.storage.SyncedTabsStorage
+import mozilla.components.feature.syncedtabs.view.SyncedTabsView
+import mozilla.components.service.fxa.manager.FxaAccountManager
+import mozilla.components.support.base.feature.LifecycleAwareFeature
+import mozilla.components.support.base.observer.Observable
+import mozilla.components.support.base.observer.ObserverRegistry
+import org.mozilla.fenix.tabstray.TabsTrayAction
+import org.mozilla.fenix.tabstray.TabsTrayStore
+
+class SyncedTabsIntegration(
```
gabrielluong, line 21: Needs kDoc

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabsIntegration.kt
```diff
@@ -0,0 +1,83 @@
...
+
+    private val syncedTabsFeature by lazy {
+        SyncedTabsFeature(
+            context = context,
+            storage = storage,
+            accountManager = accountManager,
+            view = this,
+            lifecycleOwner = lifecycleOwner,
+            onTabClicked = {
+                // We can ignore this callback here because we're not connecting the adapter
```
gabrielluong, line 40: Might want to revisit this comment since we don't have any adapter anymore.

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabsIntegration.kt
```diff
@@ -0,0 +1,83 @@
...
+                // back to the feature. This works fine in other features, but passing the listener
+                // to other components in this case is annoying.
+            }
+        )
+    }
+
+    override var listener: SyncedTabsView.Listener? = null
+
+    private val syncButtonBinding by lazy {
```
gabrielluong, line 49: Let's move this above `listener` to keep all the lazy initializations together (this is the same ordering coming from `SyncedTabsTrayLayout`).

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabsIntegration.kt
```diff
@@ -0,0 +1,83 @@
...
+        syncedTabsFeature.start()
+        syncButtonBinding.start()
+    }
+
+    override fun stop() {
+        syncedTabsFeature.stop()
+        syncButtonBinding.stop()
+    }
+
+    override fun onError(error: SyncedTabsView.ErrorType) {
```
gabrielluong, line 63: @jonalmeida wanted to double check that it's okay with dropping the `scope.launch {}` seen in https://github.com/mozilla-mobile/fenix/pull/22798/files#diff-1fd4cbb372da638657848eed86633cb5344dceb48f729a2eca27e230bbcdd593L85. It doesn't look like we need it, but wanted to check about the previous context.

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabsIntegration.kt
```diff
@@ -0,0 +1,83 @@
...
+            lifecycleOwner = lifecycleOwner,
+            onTabClicked = {
+                // We can ignore this callback here because we're not connecting the adapter
+                // back to the feature. This works fine in other features, but passing the listener
+                // to other components in this case is annoying.
+            }
+        )
+    }
+
+    override var listener: SyncedTabsView.Listener? = null
```
gabrielluong, line 47: I don't see the `listener` assigned anywhere? Wondering if we need this and  `syncButtonBinding`?

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabsListItem.kt
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * The various types of list items that can be found in a [SyncedTabsList].
+ */
+sealed class SyncedTabsListItem {
+
+    /**
+     * A device header for displaying a synced device.
+     */
+    data class Device(val displayName: String, val id: String) : SyncedTabsListItem()
```
gabrielluong, line 19: General comment, add some @param for these data class

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabsListItem.kt
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * A tab that was synced.
+     */
+    data class Tab(val displayTitle: String, val displayURL: String, val tab: SyncTab) : SyncedTabsListItem()
```
gabrielluong, line 24: This line looks a bit long, might want to run a format code on the file.

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabs.kt
```diff
@@ -0,0 +1,346 @@
...
+ * @param listener [SyncedTabsView.Listener] to apply to this layout node.
+ */
+@Composable
+fun SyncedTabsList(syncedTabs: List<SyncedTabsListItem>, listener: SyncedTabsView.Listener) {
+    val listState = rememberLazyListState()
+    LazyColumn(
```
gabrielluong, line 66: nit: add new line before

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabsIntegration.kt
jonalmeida: Good catch. I don't remember if I tested out the error case, this might end up failing to do anything if we aren't invoking this on the main thread.

The quickest way to test this is to connect a device that has tabs here. Then on desktop, go to the account settings and disconnect your test device. Cold boot Fenix and that should put you in an error state.

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabs.kt
```diff
@@ -0,0 +1,346 @@
...
+ *
+ * @param deviceName The name of the user's device connected that has synced tabs.
+ */
+@Composable
+fun SyncedTabsDeviceItem(deviceName: String) {
+    PrimaryText(
+        text = deviceName,
+        fontSize = 12.sp,
+        fontFamily = FontFamily(Font(R.font.metropolis_semibold)),
+        letterSpacing = 0.04.sp,
```
gabrielluong, line 101: We can revisit this in a followup, but if you have access to the figma designs, can we look into whether or not we need this letter spacing. I just haven't seen any consistency about adding this depending on who is working on the frontend of the feature. I will also check with the design system in the future to see if this is really expected.

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabs.kt
```diff
@@ -0,0 +1,346 @@
...
+}
+
+/**
+ * Error UI to show if there is one of the errors outlined in [SyncedTabsView.ErrorType].
+ *
+ * @param errorType [SyncedTabsView.ErrorType] to be displayed after mapping to a string resource.
+ * @param navController [NavController] to call [NavGraphDirections.actionGlobalTurnOnSync()].
+ */
+@Composable
+fun SyncedTabsErrorItem(errorType: SyncedTabsView.ErrorType, navController: NavController? = null) {
```
gabrielluong, line 150: I don't think `SyncedTabsErrorItem` really needs to know about  `NavController `. I think we should expose a optional lambda function instead for the error button `onClick`. 

In general, I don't think we need to expose controller-type classes into a Composable function since we can just add lambda functions for the event handlers (eg, click, long press), and we should also just pass simple objects that tells the Composable function the necessary state it is trying to display.

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabs.kt
```diff
@@ -0,0 +1,346 @@
...
+    val width: Float
+    val dashLength: Float
+    with(LocalDensity.current) {
+        width = 2.dp.toPx()
+        dashLength = 4.dp.toPx()
+    }
+    // As of December 2021, there is not an official way to get a rounded rectangle border with
+    // dashes using Compose. This is a next-best attempt with hard corners.
+    val stroke = Stroke(
```
gabrielluong, line 159: Any thoughts on using a canvas? https://android--code.blogspot.com/2021/04/jetpack-compose-draw-rounded-rectangle.html

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabs.kt
gabrielluong, line 150: I am noticing the paddings don't quite match the old implementation when I compare to our existing Nightly. I didn't look deep enough to figure out where the issue is.

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabs.kt
```diff
@@ -0,0 +1,346 @@
...
+        Column(
+            modifier = Modifier
+                .padding(16.dp)
+                .fillMaxWidth()
+        ) {
+            PrimaryText(
+                text = stringResource(id = errorType.toStringRes()),
+                fontSize = 14.sp,
+                modifier = Modifier
+                    .padding(0.dp, 4.dp, 0.dp, 0.dp)
```
gabrielluong, line 181:
```suggestion
                    .padding(top = 4.dp)
```
This can be simplify since the default values are 0.dp already.

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabs.kt
```diff
@@ -0,0 +1,346 @@
...
+                    .fillMaxWidth()
+            )
+            if (navController != null) {
+                Spacer(modifier = Modifier.height(8.dp))
+                Row(
+                    modifier = Modifier
+                        .clip(RoundedCornerShape(4.dp))
+                        .background(FirefoxTheme.colors.actionPrimary)
+                        .padding(8.dp, 4.dp)
```
gabrielluong, line 190: I am gonna suggest adding the named parameters to help make it easier to read this on GitHub.

gabrielluong (changes requested):
I think this is a good first pass.

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabs.kt
```diff
@@ -0,0 +1,346 @@
...
+            PrimaryText(
+                text = stringResource(id = errorType.toStringRes()),
+                fontSize = 14.sp,
+                modifier = Modifier
+                    .padding(0.dp, 4.dp, 0.dp, 0.dp)
+                    .fillMaxWidth()
+            )
+            if (navController != null) {
+                Spacer(modifier = Modifier.height(8.dp))
+                Row(
```
gabrielluong, line 186: I would move the Button into its own Composable function. Is it also possible to use a Button instead of a Row for the button?

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabsIntegration.kt
jonalmeida: The listener is part of the component interface that I [mentioned here](https://github.com/mozilla-mobile/fenix/issues/21900). It made sense at one point to have a listener (interactor) tightly coupled with the view at the time, but it no longer makes sense so we should separate those two out.

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabs.kt
MozillaNoah: This one can't be inlined in the `drawRoundRect` call, apparently. I'll take a second pass at this with your comment below regarding a Canvas. 

image

gabrielluong: wild, but makes sense

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabs.kt
MozillaNoah: Big sad
image

I've done the other refactoring to take out the controller from `SyncedTabsView.ErrorType` and `SyncedTabsErrorItem` though.

gabrielluong: I am definitely not expecting us to put the errorSubUI in `SyncedTabsView.ErrorType`.

gabrielluong: In fact, I think I expected you to keep it the same as before.

MozillaNoah: Per offline discussion, I referenced the wrong data class in my above comment. The NavController has been replaced with an optional wrapper data class, SyncedTabsListItem.ErrorButton, which will contain a String for the button text and a lambda for clicks.

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabs.kt
MozillaNoah: Ugh, I was just missing the `cornerRadius` parameter for the my `drawRoundRect` 🤦‍♀️
image

grigoryk:
Does https://github.com/mozilla-mobile/fenix/issues/23199 still happen with this branch? I've quickly tried to verify locally, but don't have time to verify this locally. Adding a log statement in `DefaultController#syncAccount && refreshSyncedTabs` should be enough to easily see that bug in action.

MozillaNoah:
> Does #23199 still happen with this branch? I've quickly tried to verify locally, but didn't have time to rebase this work locally. Adding a log statement in `DefaultController#syncAccount && refreshSyncedTabs` should be enough to easily see that bug in action.

@grigoryk Based on some debugging breakpoints, it looks like those calls are currently being made upon opening the tabs tray.

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabsIntegration.kt
MozillaNoah: I was able to verify one of the error scenarios by turning off WiFi and cold booting the app. 
image

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabsIntegration.kt
MozillaNoah: This listener was some weird black magic during the rewrite. The synced tabs data wasn't actually updating visually until I added the `Observable<SyncedTabsView.Listener> by ObserverRegistry()` line to the `SyncedTabsIntegration` class. I agree that this part of AC could probably be refactored as a larger part of the TabsTray rewrite.

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabsIntegration.kt
gabrielluong: @jonalmeida I am not too familiar with the original design spec. Do we have a Figma that tells us what each of the state should look like?

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabsIntegration.kt
gabrielluong: Probably good to file a followup isssue and drop some deets when it's still clear in your mind.

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabsIntegration.kt
jonalmeida: No. All the errors should look the same as the screenshot above to match these error types: https://github.com/mozilla-mobile/android-components/blob/main/components/feature/syncedtabs/src/main/java/mozilla/components/feature/syncedtabs/view/SyncedTabsView.kt#L61-L88

jonalmeida: You can look at the code to know what it has to look like too.

MozillaNoah: Four of the errors are just text. One is text and a button.

jonalmeida: This bit connects the feature to the UI and tells it when to show the tab list or when to show an error: https://github.com/mozilla-mobile/fenix/blob/main/app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabsTrayLayout.kt

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabsIntegration.kt
MozillaNoah: https://github.com/mozilla-mobile/android-components/issues/11552

Feel free to add any edits to this

app/src/test/java/org/mozilla/fenix/tabstray/TabsTrayStoreReducerTest.kt
jonalmeida: I missed this comment.

What does the store add that needs to be tested? In most of the stores, the store doesn't do anything for us within the structure. The logic is within the state/reducer. Put state in, reduce, test state out.

The store holds an executor and provides single access to the state and to dispatch actions. Nothing worth testing in tabs tray.

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabs.kt
```diff
@@ -0,0 +1,356 @@
...
+import mozilla.components.browser.storage.sync.Tab as SyncTab
+
+/**
+ * Top-level list UI for displaying Synced Tabs in the Tabs Tray.
+ *
+ * @param syncedTabs The tab UI items to be displayed.
+ * @param listener to apply to this layout node.
+ */
+@Composable
+fun SyncedTabsList(syncedTabs: List<SyncedTabsListItem>, listener: SyncedTabsView.Listener) {
```
jonalmeida, line 63:
```suggestion
fun SyncedTabsList(syncedTabs: List<SyncedTabsListItem>, onTabClicked: (Tab) -> Unit) {
```

If you look at how SyncedTabsAdapter is constructed, the TabClickDelegate is used as a weird pass-through to satisfy the API needs for the adapter.

You no longer have this problem in Compose. You can just call the interactor directly!

```kotlin
SyncedTabsList(listOfSyncedTabs) { tab ->
  navigationInteractor.onSyncedTabClicked(tab)
}
```

In the `SyncedTabsTrayLayout` we don't even complete the other end of the call because it's not worth the gymnastics needed to get the listener to work just to have it provide the same data to us in the feature. See this comment [here](https://github.com/mozilla-mobile/fenix/blob/0c34bd93b83ac929ae6b14c3216e95cf7092b79c/app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabsTrayLayout.kt#L53-L57).

The only value the listener now provides is to tell the feature to refresh the list, and you have that covered in `SyncedTabsIntegration` with the [`SyncButtonBinding`](https://github.com/mozilla-mobile/fenix/blob/main/app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncButtonBinding.kt).

What this means, is that it simplifies the refactor in AC because it just needs to be a method on the feature and not by a reference to a listener through a view.

After removing the listener from the feature your `SyncedTabsIntegration` call to refresh would look like this:
```kotlin
private val syncButtonBinding by lazy {
  SyncButtonBinding(store) { syncedTabsFeature.syncNow() }
}
```

app/src/main/java/org/mozilla/fenix/tabstray/TrayPagerAdapter.kt
```diff
@@ -76,30 +70,38 @@ class TrayPagerAdapter(
...
             }
             SyncedTabsPageViewHolder.LAYOUT_ID -> {
                 SyncedTabsPageViewHolder(
-                    itemView,
-                    tabsTrayStore
+                    composeView = ComposeView(parent.context).apply {
+                        layoutParams = ViewGroup.LayoutParams(
+                            ViewGroup.LayoutParams.MATCH_PARENT,
+                            ViewGroup.LayoutParams.MATCH_PARENT
+                        )
```
gabrielluong, line 85: I am wondering if we need this? I haven't seen us needing to do this thus far.

app/src/main/java/org/mozilla/fenix/tabstray/TrayPagerAdapter.kt
```diff
@@ -76,30 +70,38 @@ class TrayPagerAdapter(
...
                 PrivateBrowserPageViewHolder(
-                    itemView,
+                    LayoutInflater.from(parent.context).inflate(viewType, parent, false),
                     tabsTrayStore,
                     browserStore,
                     interactor
                 )
             }
             SyncedTabsPageViewHolder.LAYOUT_ID -> {
                 SyncedTabsPageViewHolder(
```
gabrielluong, line 80: I am wondering if we need to do something similar to `ComposeViewHolder` to avoid https://github.com/mozilla-mobile/fenix/issues/22383

app/src/main/java/org/mozilla/fenix/tabstray/TrayPagerAdapter.kt
```diff
@@ -76,30 +70,38 @@ class TrayPagerAdapter(
...
     override fun onBindViewHolder(viewHolder: AbstractPageViewHolder, position: Int) {
-        val adapter = when (position) {
-            POSITION_NORMAL_TABS -> normalAdapter
-            POSITION_PRIVATE_TABS -> privateAdapter
-            POSITION_SYNCED_TABS -> syncedTabsAdapter
-            else -> throw IllegalStateException("View type does not exist.")
+        when (viewHolder) {
+            is NormalBrowserPageViewHolder -> viewHolder.bind(normalAdapter)
+            is PrivateBrowserPageViewHolder -> viewHolder.bind(privateAdapter)
+            is SyncedTabsPageViewHolder -> viewHolder.bind()
```
gabrielluong, line 103: It might be worthwhile to align more closely with `SessionControlAdapter`. https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlAdapter.kt#361

Also wondering if we need to do something like https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlAdapter.kt#293.

app/src/main/java/org/mozilla/fenix/tabstray/viewholders/SyncedTabsPageViewHolder.kt
```diff
@@ -5,32 +5,46 @@
...
 import org.mozilla.fenix.tabstray.TabsTrayStore
-
+import org.mozilla.fenix.tabstray.syncedtabs.SyncedTabsList
+import org.mozilla.fenix.theme.FirefoxTheme
+
+/**
+ * Temporary ViewHolder to render [SyncedTabsList] until all of the Tabs Tray is written in Compose.
+ *
+ * @param composeView Root ComposeView passed-in from TrayPagerAdapter.
+ * @param tabsTrayStore Store used as a Composable State to listen for changes to TabsTrayState.syncedTabs.
```
gabrielluong, line 20:
```suggestion
 * @param tabsTrayStore Store used as a Composable State to listen for changes to [TabsTrayState.syncedTabs].
```

I am assuming this code is linkable.

app/src/main/java/org/mozilla/fenix/tabstray/viewholders/SyncedTabsPageViewHolder.kt
```diff
@@ -5,32 +5,46 @@
...
+    private val navigationInteractor: NavigationInteractor,
+) : AbstractPageViewHolder(composeView) {
+
+    fun bind() {
+        composeView.setContent {
+            val tabs = tabsTrayStore.observeAsComposableState { state -> state.syncedTabs }.value
+            FirefoxTheme {
+                SyncedTabsList(
+                    syncedTabs = tabs ?: emptyList(),
+                    onTabClicked = { navigationInteractor.onSyncedTabClicked(it) }
```
gabrielluong, line 35:
```suggestion
                    onTabClick = navigationInteractor::onSyncedTabClicked
```
I think we are going with present tense for these callback function naming. Mostly been telling people to use present tense because that's what we have done so far, and what I am basing off as best practice from looking at some Compose internals like DropdownMenu.

Also, just another suggestion for a simpler way to pass in the lambda function.

app/src/main/res/values/strings.xml
```diff
@@ -191,7 +191,7 @@
     <!-- Browser menu toggle that installs a Progressive Web App shortcut to the site on the device home screen. -->
     <string name="browser_menu_install_on_homescreen">Install</string>
     <!-- Menu option on the toolbar that takes you to synced tabs page-->
-    <string name="synced_tabs">Synced tabs</string>
+    <string name="synced_tabs" moz:removedIn="98" tools:ignore="UnusedResources">Synced tabs</string>
```
gabrielluong, line 194: Just wondering about this removal. Are we removing `view_synced_tabs_title` because it is never used?

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabsIntegration.kt
```diff
@@ -0,0 +1,121 @@
...
+    Observable<SyncedTabsView.Listener> by ObserverRegistry() {
+
+    private val syncedTabsFeature by lazy {
+        SyncedTabsFeature(
+            context = context,
+            storage = storage,
+            accountManager = accountManager,
+            view = this,
+            lifecycleOwner = lifecycleOwner,
+            onTabClicked = {
```
gabrielluong, line 51: Maybe file something for AC about cleaning this up.

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabsIntegration.kt
```diff
@@ -0,0 +1,121 @@
...
+                // back to the feature.
+            }
+        )
+    }
+
+    private val syncButtonBinding by lazy {
+        SyncButtonBinding(store) { listener?.onRefresh() }
+    }
+
+    override var listener: SyncedTabsView.Listener? = null
```
gabrielluong, line 62: This is a followup. Same as above about cleaning this up in AC. I would verify with @jonalmeida and make sure there's no other consumers here.

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabsIntegration.kt
```diff
@@ -0,0 +1,121 @@
...
+
+    override fun onError(error: SyncedTabsView.ErrorType) {
+        // We may still be displaying a "loading" spinner, hide it.
+        stopLoading()
+
+        store.dispatch(TabsTrayAction.UpdateSyncedTabs(listOf(error.toSyncedTabsListItem())))
+    }
+
+    /**
+     * Do nothing; the UI is handled with FloatingActionButtonBinding.
```
gabrielluong, line 82:
```suggestion
     * Do nothing; the UI is handled with [FloatingActionButtonBinding].
```

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabsIntegration.kt
gabrielluong: While I am here, is there an actual listener that is attached to this?

app/src/main/java/org/mozilla/fenix/tabstray/TrayPagerAdapter.kt
MozillaNoah: It was crashing without this when the Tabs Tray's recyclerview moved to this ViewHolder

app/src/main/java/org/mozilla/fenix/tabstray/TrayPagerAdapter.kt
MozillaNoah: align with which part? not needing bind() call?

gabrielluong: Ya not needing to call bind(), I don't think we need to do this here. This can be a followup.

app/src/main/res/values/strings.xml
MozillaNoah: Yes. Based on some context clues of git annotations, I am under the impression that Sync previously lived under its own screen + screen title, and the layout + string are leftover from that era after synced tabs were moved to the tabs tray.

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabsIntegration.kt
MozillaNoah: I'd assume it'd be a part of this https://github.com/mozilla-mobile/android-components/issues/11552

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabsIntegration.kt
MozillaNoah: Yes, this is some of the weirdness of the `Observable<SyncedTabsView.Listener> by ObserverRegistry()`. From my understanding it's initialized at runtime but none of that should be necessary, hopefully, after https://github.com/mozilla-mobile/android-components/issues/11552.

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabs.kt
```diff
@@ -0,0 +1,348 @@
...
+            text = deviceName,
+            modifier = Modifier
+                .fillMaxWidth()
+                .padding(start = 16.dp, top = 16.dp, end = 8.dp, bottom = 8.dp),
+            fontSize = 14.sp,
+            fontFamily = FontFamily(Font(R.font.metropolis_semibold)),
+            maxLines = 1
+        )
+
+        Divider(color = FirefoxTheme.colors.borderDefault, thickness = 1.dp)
```
gabrielluong, line 116: Probably can drop `thickness = 1.dp` since that's the default.

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabs.kt
```diff
@@ -0,0 +1,348 @@
...
+        fontSize = 16.sp,
+        maxLines = 1
+    )
+}
+
+@Composable
+@Preview
+private fun SyncedTabsListItemsPreview() {
+    FirefoxTheme(darkTheme = true) {
+        Column(Modifier.background(FirefoxTheme.colors.layer2)) {
```
gabrielluong, line 254: These should be layer1. See https://www.figma.com/file/YcQC7rU1SQJx49kIAXZhcI/Component-and-Color-Usage-Examples?node-id=78%3A11241

app/src/main/java/org/mozilla/fenix/tabstray/TrayPagerAdapter.kt
MozillaNoah: Reading back through this, we also may not even need to file a separate follow-up, since the real follow-up is converting the rest of TabsTray to Compose

MozillaNoah:
Most recent push simply increases the padding between the error text and the error button from 8dp to 12dp.

MozillaNoah:
Push to update from main to fix colors + build issue

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabs.kt
```diff
@@ -0,0 +1,352 @@
...
+@Composable
+fun SyncedTabsTabItem(tabTitleText: String, url: String, onClick: () -> Unit) {
+    Column(
+        modifier = Modifier
+            .clickable(
+                onClickLabel = tabTitleText,
+                onClick = onClick
+            )
+            .padding(horizontal = 16.dp)
+            .defaultMinSize(minHeight = 56.dp),
```
gabrielluong, line 136: General question outside of the review, I am wondering how we decide to uses `defaultMinSize` and  `height(IntrinsicSize.Min)` (in general, nothing in particular about the current code).

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabs.kt
```diff
@@ -0,0 +1,352 @@
...
+            .fillMaxWidth(),
+        fontSize = 16.sp,
+        maxLines = 1
+    )
+}
+
+@Composable
+@Preview
+private fun SyncedTabsListItemsPreview() {
+    FirefoxTheme(darkTheme = true) {
```
gabrielluong, line 257: I think we should specify the darkTheme in the `@Preview` instead.

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabs.kt
```diff
@@ -0,0 +1,352 @@
...
+            }
+        }
+    }
+}
+
+/**
+ * Converts a list of [SyncedDeviceTabs] into a list of [SyncedTabsListItem].
+ */
+fun List<SyncedDeviceTabs>.toComposeList() = asSequence().flatMap { (device, tabs) ->
+    // transform to sticky headers data here https://github.com/mozilla-mobile/fenix/issues/19942
```
gabrielluong, line 313:
```suggestion
    // Transform to sticky headers data here https://github.com/mozilla-mobile/fenix/issues/19942
```

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabs.kt
```diff
@@ -0,0 +1,352 @@
...
+ * Helper function to create a [SyncedTabsListItem.Tab] for previewing.
+ */
+private fun generateFakeTab(tabName: String, tabUrl: String): SyncedTabsListItem.Tab = SyncedTabsListItem.Tab(
+    tabName.ifEmpty { tabUrl },
+    tabUrl,
+    SyncTab(
+        listOf(TabEntry(tabName, tabUrl, null)),
+        0,
+        0L
+    )
```
gabrielluong, line 351: Let's add the named parameters here

gabrielluong (approved):

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabs.kt
MozillaNoah: I used `defaultMinSize` to be a more direct 1:1 to `android:minHeight` in XML (to help ensure with a default text size that there would be enough padding), and the `height(IntrinsicSize.Min)` was used when I was specifically trying out random things to get the dashed border for the error UI to show. The documentation for `Modifier.height()` and `Modifier.defaultMinSize()` aren't the clearest though on when to use either.

app/src/main/java/org/mozilla/fenix/tabstray/syncedtabs/SyncedTabs.kt
MozillaNoah: This was the better solution, since it no longer re-compiles the `@Preview` when just changing from light to dark

PR #22829, opened 2021-12-13
fixes #20833: move home recycler bottom margin into standalone item

### LINKED ISSUE #20833: [Bug]: Little black bar at bottom of home screen
### Steps to reproduce

When scrolling the home screen slowly up and down then sometimes a small black bar seems to appear at the bottom of the screen, hiding some parts of the home screen.

### Expected behaviour

(No black bar)

### Actual behaviour

device-2021-08-13-130657

### Device name

PIxel 4a

### Android version

Android 12

### Firefox release type

Firefox Nightly

### Firefox version

93.0a1 (2021-08-12)

### Device logs

_No response_

### Additional information

_No response_

MatthewTighe:
- No black box over pocket items

image

- `Customize homepage` still visible with required bottom spacing

image

### Pull Request checklist

- [x] **Tests**: N/A
- [x] **Screenshots**: Included
- [x] **Accessibility**: N/A

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/home/BottomSpacerViewHolder.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/BottomSpacerViewHolder.kt
@@ -1,0 +1,28 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.home
+
+import android.view.View
+import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.height
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.ComposeView
+import androidx.compose.ui.unit.dp
+import androidx.recyclerview.widget.RecyclerView
+import org.mozilla.fenix.theme.FirefoxTheme
+
+class BottomSpacerViewHolder(val composeView: ComposeView) : RecyclerView.ViewHolder(composeView) {
+    init {
+        composeView.setContent {
+            FirefoxTheme {
+                Spacer(modifier = Modifier.height(88.dp))
+            }
+        }
+    }
+
+    companion object {
+        val LAYOUT_ID = View.generateViewId()
+    }
+}
--- a/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlAdapter.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlAdapter.kt
@@ -44,6 +44,7 @@ import org.mozilla.fenix.home.sessioncontrol.viewholders.onboarding.OnboardingTh
 import org.mozilla.fenix.home.sessioncontrol.viewholders.onboarding.OnboardingToolbarPositionPickerViewHolder
 import org.mozilla.fenix.home.sessioncontrol.viewholders.onboarding.OnboardingTrackingProtectionViewHolder
 import org.mozilla.fenix.home.tips.ButtonTipViewHolder
+import org.mozilla.fenix.home.BottomSpacerViewHolder
 import org.mozilla.fenix.home.topsites.TopSitePagerViewHolder
 import mozilla.components.feature.tab.collections.Tab as ComponentTab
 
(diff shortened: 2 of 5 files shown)


app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/CustomizeHomeButtonViewHolder.kt
```diff
@@ -48,6 +48,8 @@ class CustomizeHomeButtonViewHolder(
                     CustomizeHomeButton(
                         onButtonClick = { interactor.openCustomizeHomePage() }
                     )
+
+                    Spacer(modifier = Modifier.height(88.dp))
```
gabrielluong, line 52: Just wanted to state that we are not guaranteed to always render this. https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlView.kt#92

gabrielluong:
@MatthewTighe you should add the "#" in the PR title and commit message always you don't get the linking.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/CustomizeHomeButtonViewHolder.kt
gabrielluong: Although looking at the issue, this should be fine, but might be worthwhile to test where the "Customize home" button is not present.

MatthewTighe: I'd considered using a standalone view for just the spacing. That feels the most foolproof to me, but it felt like a bit of an awkward solution. I figured the likelihood of the `CustomizeHomeButtonViewHolder` not being shown was low, given the amount conditions that triggered it.

I was also relying on the Figma designs available for the home page. They show the button as always visible, so I thought this would probably be good enough. Are there conditional layouts in Figma that I'm missing? New to the tool.

Happy to add a standalone view for the spacing. It might make the code more explicit anyway, rather than hiding in whatever happens to be the last view.

MatthewTighe:
> @MatthewTighe can you add the "#" to the issue number in the PR title and commit message otherwise you don't get the linking.

You got me again :joy: Swear I'll figure it out eventually

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/CustomizeHomeButtonViewHolder.kt
gabrielluong: Ya... it's an awkward situation. Without testing this myself, my current guesses is that a standalone view is probably most foolproof as you stated. If I was in your shoes, I would only gain more confidence of the current solution by seeing how it looks without the "Customize home" button, also consider that the various views like "Recent Bookmarks", "Collections", 'Jump back in", "Pocket" and "Recent History" are all customizable and conditional. I am wondering what happens to the bottom spacing for all the different combination of views being visible and not visible. I guess what we are looking for here is just a spacing below the last visible view and the toolbar.

gabrielluong: Alternatively, I would look at refactoring the various views to add a bottom spacer to each view so we aren't adding the spacer at the top of the view only when they are render. See the 68dp spacer in this view as an example.

gabrielluong: Of course, after thinking about all these troubling possibilities. It is possible that simple is best here and adding the standalone view is the way to go, but I leave that to your investigation.

MatthewTighe: The bottom still gets cut off when the `customize homepage` button isn't present, so good call there.

I'll go ahead and add a standalone spacer item for now. It'd be much easier to add bottom spacing to all the views if compose was being used throughout the home screen, but I'm hesitant to add/update custom spacing to each individual item as is.

MatthewTighe: Updated!

gabrielluong (approved):

PR #22843, opened 2021-12-14
Modify strings for Privacy notice for MozillaOnline builds

rxumoz:
Due to an urgent requirement of regulations, we need to change two strings of MozillaOnline builds


app/src/main/res/values-zh-rCN/mozonline_strings.xml
```diff
@@ -18,7 +18,7 @@
     <!-- Privacy Notice positive button-->
     <string name="privacy_notice_positive_button">同意并继续</string>
     <!-- Privacy Notice neutral button-->
-    <string name="privacy_notice_neutral_button">退出应用</string>
+    <string name="privacy_notice_neutral_button">不同意并退出</string>
```
Amejia481, line 21: For translation reasons, we don't update strings, we just create a new string with a sequential number for example `privacy_notice_neutral_button_1` and mark it for future [removal](https://github.com/mozilla-mobile/fenix/wiki/Removing-strings), you could find an example [here](https://github.com/mozilla-mobile/fenix/commit/888bbc770eb1ecd781262fa7c3a5a8b632aeb6fe#diff-5e01f7d37a66e4ca03deefc205d8e7008661cdd0284a05aaba1858e6b7bf9103)

Amejia481 (approved):

app/src/main/res/values-zh-rCN/mozonline_strings.xml
rxumoz: Thanks! I think the strings in mozonline_strings.xml will not be translated into more languages, but I'll update to keep consistency.

rxumoz:
@Mergifyio backport releases_v96.0.0

Issue #22851, opened 2021-12-14
Allow to change the home screen background by tapping on the firefox logo


Amejia481:

sheikh-azharuddin:
Nice implementation 👍
These 2 backgrounds color have issue with the text color ...

![Screenshot_20211220-231215104](https://user-images.githubusercontent.com/58127841/146809690-07f8c015-10a4-4b9f-96b2-fec2f9febdc2.jpg)
![Screenshot_20211220-231234021](https://user-images.githubusercontent.com/58127841/146809698-9e756d7e-5d8d-4c1e-a09b-341168d1deac.jpg)

pocmo:
@Amejia481 Two suggestions:
* I found the commit and this issue by looking at the APK size graphs and saw a bump, which seems to be coming from the new wallpaper PNGs. Did you run them through ImageOptim? It looks like we could optimize them.
* I'd consider adding such state either to the screen or app store (based on whether it's needed locally only or globally) and using the known mechanisms to update/observe state. Every `*Manager` creates another source of truth/state and will make it harder to reason about the app's state.

Amejia481:
👋🏽 thanks for the suggestions.

> Did you run them through ImageOptim?

No I didn't, but I will do :) . These images are the final ones, they are only for gathering feedback about the feature.

> I'd consider adding such state either to the screen or app store (based on whether it's needed locally only or globally) and using the known mechanisms to update/observe state. Every *Manager creates another source of truth/state and will make it harder to reason about the app's state.

Yeah that make sense 👍🏽

Amejia481:
@pocmo I filed https://github.com/mozilla-mobile/fenix/pull/22937 for addressing the images optimization.

delia-pop:
Verified as fixed on Beta 99.0.0-beta.5, Nightly 100.0a1. and RC 98.3.0.

PR #22860, opened 2021-12-14
For #22851 toggle wallpapers when tap on the Firefox log

Amejia481:
### Pull Request checklist

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -158,6 +159,11 @@ class Settings(private val appContext: Context) : PreferencesHolder {
         default = ""
     )
 
+    var currentWallpaper by stringPreference(
+        appContext.getPreferenceKey(R.string.pref_key_current_wallpaper),
+        default = ""
```
MatthewTighe, line 164: Would it make sense for the default to be `Wallpapers.NONE`?

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -178,6 +184,20 @@ class Settings(private val appContext: Context) : PreferencesHolder {
         false
     )
 
+    fun getWallpaper(): Wallpapers {
+        return Wallpapers.valueOf(currentWallpaper)
+    }
```
MatthewTighe, line 189: Looking at the return type here, I think the singular `Wallpaper` would work better for the name of the `enum`.

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -178,6 +184,20 @@ class Settings(private val appContext: Context) : PreferencesHolder {
...
+
+    fun toggleWallpaper(): Wallpapers {
+        val wallpaper = if (currentWallpaper.isEmpty()) {
+            Wallpapers.values().first()
+        } else {
+            Wallpapers.valueOf(currentWallpaper).next()
+        }
+        currentWallpaper = wallpaper.name
+        return wallpaper
+    }
```
MatthewTighe, line 199: Putting this function in this location feels a little awkward to me. It could potentially be combined with the `Wallpaper::next` function in the `Wallpaper` type itself, and receive ` Settings` to perform the update, so that the signature became `Wallpaper.switchToNextWallpaper(settings: Settings)`.

MatthewTighe (commented):
Had a few initial comments. Let me know I'm jumping in too early here, since I'm sure the implementation will evolve as the feature does.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -394,6 +396,11 @@ class HomeFragment : Fragment() {
         requireComponents.core.engine.profiler?.addMarker(
             MarkersFragmentLifecycleCallbacks.MARKER_NAME, profilerStartTime, "HomeFragment.onCreateView",
         )
+
+        if (FeatureFlags.showWallpapers) {
+            val wallpaper = requireContext().settings().getWallpaper()
+            updateWallpaper(wallpaper, requireContext())
+        }
```
gabrielluong, line 403: Can you add a new line before this to space out the if statement block

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -751,6 +758,21 @@ class HomeFragment : Fragment() {
...
+
+        if (FeatureFlags.showWallpapers) {
+            binding.wordmark.setOnClickListener {
+                val wallpaper = context.settings().toggleWallpaper()
+                updateWallpaper(wallpaper, context)
+            }
+        }
+    }
+
+    private fun updateWallpaper(wallpaper: Wallpapers, context: Context) {
```
gabrielluong, line 770: Can't we access the `context` within the function without passing it in as a param?

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
gabrielluong: This is a nice idea for further abstracting the functionality into the same class while also keeping Settings lean.

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
Amejia481: Sure!

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
Amejia481: 👍🏽

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
Amejia481: 👍🏽

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
Amejia481: 👍🏽

Amejia481:
@MatthewTighe  and @gabrielluong I updated the PR with your feedback!

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -751,6 +759,26 @@ class HomeFragment : Fragment() {
         lifecycleScope.launch(IO) {
             requireComponents.reviewPromptController.promptReview(requireActivity())
         }
+
+        if (FeatureFlags.showWallpapers) {
+            binding.wordmark.setOnClickListener {
+                val settings = context.settings()
+                val newWallpaper = settings.getWallpaper().switchToNextWallpaper()
```
gabrielluong, line 766: If I understood @MatthewTighe correctly, I thought the idea was to add a function to set the Settings pref inside of the `Wallpaper` class. That way we aren't putting this logic inside Settings or HomeFragment.

Amejia481: I didn't added it, as we will be adding the settings dependency to the `Wallpaper` enum, and the function will be switching and storing at the same time, for now the only place that we are going to be using this functionality is here.

Amejia481: I guess we could add it, as this is not the permanent solution anyways XD, this is just a placeholder while the functionality evolves

Amejia481: @gabrielluong I updated the PR

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -751,6 +759,23 @@ class HomeFragment : Fragment() {
...
+            }
+        }
+    }
+
+    private fun updateWallpaper(newWallpaper: Wallpaper) {
+        val context = requireContext()
+        if (newWallpaper == Wallpaper.NONE) {
+            binding.homeLayout.setBackgroundColor(context.getColorFromAttr(newWallpaper.drawable))
+        } else {
+            binding.homeLayout.setBackgroundResource(newWallpaper.drawable)
```
MozillaNoah, line 777: when the wallpaper is set here, how does the image get rendered on the device? is it adjusted to the aspect ratio of the device or is it centered within the screen?

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -178,6 +184,17 @@ class Settings(private val appContext: Context) : PreferencesHolder {
...
+    /**
+     * Returns the current selected [Wallpaper].
+     */
+    fun getWallpaper(): Wallpaper {
+        return if (currentWallpaper.isEmpty()) {
+            Wallpaper.NONE
+        } else {
+            Wallpaper.valueOf(currentWallpaper)
+        }
+    }
```
MatthewTighe, line 196: I agree with GL's offline comment that this might be better placed within the `Wallpaper` class itself, since there's already a property exposing the `currentWallpaper`.

MatthewTighe (changes requested):
I'm wondering if we should leave `wallpaper_1.png` and `wallpaper_2.png` out of the remote tree. If we change the feature flag to be completely off it shouldn't matter, but I'm worried that users will run into this unexpectedly before we're ready for them too. It might make the most sense to leave the feature off until we have our telemetry in place.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
MatthewTighe: It should be adjusted to the aspect ratio of the device. We're still waiting on finalized designs to do this testing, but we're going to (hopefully) try this out with both a vector graphic and PNGs. We should have portrait and landscape versions of both. The vector will autoscale, and the PNGs would match based on [the screen density of the device](https://developer.android.com/training/multiscreen/screendensities)

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
gabrielluong: Just reposting my offline comment:

I am thinking what if `Wallpaper` was more of a manager in terms of managing the available wallpaper drawables, knowing what is the current wallpaper, next wallpaper to display, instead of the current enum class that it is.

gabrielluong: It probably means we might want `Wallpaper` to be part of `Components`, but can receive `Settings` as a parameter.

app/src/main/java/org/mozilla/fenix/FeatureFlags.kt
```diff
@@ -105,5 +105,5 @@ object FeatureFlags {
     /**
      * Enables showing the wallpaper functionality.
      */
-    const val showWallpapers = false
+    val showWallpapers = Config.channel.isNightlyOrDebug
```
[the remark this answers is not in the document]
gabrielluong: Alternatively, we can go with `isDebug` if we want to be able to test in local dev builds only.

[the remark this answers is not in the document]
Amejia481: The idea idea of putting it on nightly is to get feedback as soon as possible, as [nightly](https://play.google.com/store/apps/details?id=org.mozilla.fenix) is a developer early preview features version, the bar is not that high and we don't need to have finalize functionality. :)

Amejia481:
👋🏽  @gabrielluong, @MatthewTighe and @MozillaNoah, the PR is updated and ready for review :)

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -751,6 +757,12 @@ class HomeFragment : Fragment() {
         lifecycleScope.launch(IO) {
             requireComponents.reviewPromptController.promptReview(requireActivity())
         }
+
+        if (FeatureFlags.showWallpapers) {
+            binding.wordmark.setOnClickListener {
+                requireComponents.wallpaperManager.switchToNextWallpaper(wallpaperContainer = binding.homeLayout)
```
gabrielluong, line 763: Rather than taking in the container, I think we should just return the next wallpaper from the manager, and let the container do what it needs to do with it.

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -158,6 +159,11 @@ class Settings(private val appContext: Context) : PreferencesHolder {
         default = ""
     )
 
+    var currentWallpaper by stringPreference(
```
gabrielluong, line 162: Can we add a kDoc for this preference

app/src/main/java/org/mozilla/fenix/wallpapers/Wallpaper.kt
```diff
@@ -0,0 +1,27 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.wallpapers
+
+import org.mozilla.fenix.R
+import org.mozilla.fenix.utils.Settings
+
+enum class Wallpaper(val drawable: Int) {
```
gabrielluong, line 10: Some kDocs for the class and methods will also be appreciated

app/src/main/res/values/preference_keys.xml
```diff
@@ -185,6 +185,7 @@
     <string name="pref_key_adjust_network" translatable="false">pref_key_adjust_network</string>
     <string name="pref_key_adjust_adgroup" translatable="false">pref_key_adjust_adgroup</string>
     <string name="pref_key_adjust_creative" translatable="false">pref_key_adjust_creative</string>
+    <string name="pref_key_current_wallpaper" translatable="false">pref_key_current_wallpaper</string>
```
gabrielluong, line 188: Can we add a comment for the pref key?

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
Amejia481: > A comment is a failure to express yourself in code.  If you fail, then write a comment; but try not to fail. <Uncle Bob Martin>
XD
https://twitter.com/unclebobmartin/status/870311898545258497

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
Amejia481: I think this is fine for the moment, as we don't have any other use case for `switchToNextWallpaper`, remember that this is just a prototype no the final version, this will be iterate over multiple times.

Amejia481: It's very likely that we will remove this code

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
Amejia481: When something is explicit we don't need docs :)

Amejia481:
@gabrielluong the PR is updated!

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperManager.kt
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    private fun getCurrentWallpaperFromSettings(): Wallpaper {
+        val currentWallpaper = settings.currentWallpaper
+        return if (currentWallpaper.isEmpty()) {
+            Wallpaper.NONE
+        } else {
+            Wallpaper.valueOf(currentWallpaper)
+        }
+    }
```
MatthewTighe, line 57: The logic in this function could be moved into the getter for the `currentWallpaper` property to keep a consistent style with the setter.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
MatthewTighe: I agree with @gabrielluong here. Generally speaking, I try keep as many Android components out of my classes as possible. It has several benefits such as making classes more unit testable and avoiding retaining references to large and expensive objects like a `Context`.

Even though this code is intended to be iterated on in the near future, I think we should try to ensure that any code we ship to users is production ready. Someone might download this nightly and then not updated for several weeks, for example.

gabrielluong:
I am deferring to @MatthewTighe for the final review. 🙏

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperManager.kt
Amejia481: We are doing this to just read from disk (shared pref)  ones, when initializing `currentWallpaper`, then the current value will be stored in memory [instead of reading from disk every time the property is accessed (get)](https://kotlinlang.org/docs/properties.html#backing-fields)

Amejia481:
@MatthewTighe @MozillaNoah the PR is updated.

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperManager.kt
```diff
@@ -0,0 +1,58 @@
...
+package org.mozilla.fenix.wallpapers
+
+import android.view.View
+import mozilla.components.support.ktx.android.content.getColorFromAttr
+import org.mozilla.fenix.utils.Settings
+
+/**
+ * Provides access to available wallpapers and manages their states.
+ */
+class WallpaperManager(private val settings: Settings) {
```
MozillaNoah, line 14: I really like this wrapper class for the functionality. Just thinking out loud, does this need any test coverage?

MozillaNoah (approved):
did some local testing and this looks good to me. We might want to consider adding a click ripple to the logo now, but that's my sole nit for this haha

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperManager.kt
Amejia481: For the moment this is more like a prototype, as we would like to collect feedback from nightly users while we shape the feature, as requirements at the moment are not fully completed, when we have some more concrete we will cover with tests :)

Amejia481:
> did some local testing and this looks good to me. We might want to consider adding a click ripple to the logo now, but that's my sole nit for this haha

Thanks, I'll take a look for adding the ripple :)

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperManager.kt
MatthewTighe: Gotcha, that makes sense

MatthewTighe (approved):

Issue #22477, opened 2021-11-17
null as title / subject when sharing Youtube videos


celrm:
This happened in my Firefox for Android, both Daylight and Nightly, when I tried to share a Youtube video. I reported at https://bugzilla.mozilla.org/show_bug.cgi?id=1741168 and was told to report here.

### Steps to reproduce:

Open a Youtube video
Click on share button
Click on some apps to share the video (email, telegram, notes)

### Actual results:

As an email, the subject is null and not the video's title.
As a Google Keep note, it's the note's title.
As a telegram message, null is sent above the video.

### Expected results:

I believe the video's title should appear, instead of "null".

mariansima:
I managed to reproduce this issue on Samsung Galaxy A8 (Android 9) with Firefox Nightly 95.0a1 2021-11-18
95.0.0-beta.4 2021-11-15
RC 94.1.2 2021-11-08

VitalyVPinchuk:
Hey @celrm @sflorean !
Google Chrome does the same but leaves empty space instead of 'null' in the title field.
If you click share menu item in the Fenix menu you'll get the expected result (managed by Fenix itself). But if you click share link under the video then it won't work (managed by Gecko engine).

SoftVision-LorandJanos:
Verified as fixed on the latest Nightly 98.0a1 (2022-01-20).
Devices used:
- Samsung Galaxy Tab S3 (Android 9).
- Google Pixel 4 (Android 12).

Closing the ticket as fixed.

PR #22865, opened 2021-12-15
For #22477: Better handle null in share data title.

mcarare:
### Pull Request checklist

1. click on Show All Checks,


topotropic:
> There are no specs from UX related to what the share button in a web page should share as title when none is available, but I left it an empty string, the same as for the menu share button behavior.

I tried to share a YouTube video via the menu and it uses the video title as subject. Would that be possible?

mcarare:
@topotropic We do not receive the title from Gecko, I mentioned this in the original Bugzilla [report](https://bugzilla.mozilla.org/show_bug.cgi?id=1741168). It might be possible to fix it in AC or Fenix, but it might be an overcomplication. AFAIK none of the browsers I tested managed to send the title in this scenario.

mcarare:
@topotropic can we go ahead with using this approach? (displaying empty field instead of "null" string) Thank you!

topotropic:
@mcarare sounds good, let's leave it empty for now. Thanks!

app/src/main/java/org/mozilla/fenix/share/ShareController.kt
```diff
@@ -204,7 +204,9 @@ class DefaultShareController(
     }
 
     @VisibleForTesting
```
gabrielluong, line 206: Can we also add a test for this?

mcarare:
@gabrielluong I added the tests.

app/src/main/java/org/mozilla/fenix/share/ShareController.kt
```diff
@@ -204,7 +204,9 @@ class DefaultShareController(
...
-    internal fun getShareSubject() = shareSubject ?: shareData.map { it.title }.joinToString(", ")
+    internal fun getShareSubject() =
+        shareSubject ?: shareData.filterNot { it.title.isNullOrEmpty() }
+            .joinToString(", ") { it.title.orEmpty() }
```
gabrielluong, line 209:
```suggestion
            .joinToString() { it.title.orEmpty() }
```
Could leave as is or just remove it since ", " is already the default value.

gabrielluong (approved):

gabrielluong (approved):

app/src/main/java/org/mozilla/fenix/share/ShareController.kt
mcarare: For the separator value, I prefer to have the default value explicitly as the parameter for methods where it is not very obvious that there is a parameter and what is the default value for it.
As for the transform function, we cannot simply remove it, because we want to map the ShareData objects to their titles. So at least `{ it.title } `should be used. And as the title is a `CharSequence` an additional method is needed to make it a `String`. I chose `{ it.title.orEmpty() }` but I can also use `{ it.title.toString() }` if you prefer it.

gabrielluong: { it.title.toString() } or { it.title } sounds good to me

PR #22866, opened 2021-12-15

MickeyMoz:


grigoryk:
This will happen in https://github.com/mozilla-mobile/fenix/pull/22866 to account for a breaking change.

PR #22868, opened 2021-12-15
No issue: Try out RetryTestRule in SmokeTest UI test class

AaronMT:


sv-ohorvath (approved):
Looks good to me, a good idea to catch the AssertionError as it is pretty common since replacing most of our methods with UiAutomator. Would we create a different rule for other types of error, say IdlingResources timing out, for example?

AaronMT:
> Looks good to me, a good idea to catch the AssertionError as it is pretty common since replacing most of our methods with UiAutomator. Would we create a different rule for other types of error, say IdlingResources timing out, for example?

I originally tried to catch `Throwable` (https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-throwable/) as a catch-all in this PR but it was too generic for a `detekt` check (https://detekt.github.io/detekt/exceptions.html#toogenericexceptioncaught). You can multi-catch other exceptions but it would require using the same base (https://discuss.kotlinlang.org/t/does-kotlin-have-multi-catch/486/26) for some logic to work. What other exceptions have you seen recently? I'd like to try this out for now and see if this is useful in the Smoketest class. If you'd like to try this Rule in another class too, go ahead.

PR #22872, opened 2021-12-15
Switch beetmover to use production worker and bucket

hneiva:


taskcluster/fenix_taskgraph/transforms/beetmover.py
```diff
@@ -50,14 +50,20 @@ def make_task_description(config, tasks):
         if task.get("locale"):
             attributes["locale"] = task["locale"]
 
+        # Switch between nightly and release scope
+        # Note: if dep bucket is wanted, *temporarily* switch it to "~:dep"
+        if task['build-type']['nightly']:
+            bucket_scope = "project:mobile:fenix:releng:beetmover:bucket:nightly"
+        else:
+            bucket_scope = "project:mobile:fenix:releng:beetmover:bucket:release"
```
escapewindow, line 58: Should we check for beta or release here, and default to dep? Or check level and send `level < 3` to dep?

taskcluster/fenix_taskgraph/transforms/beetmover.py
```diff
@@ -50,14 +50,20 @@ def make_task_description(config, tasks):
...
         task = {
             "label": label,
             "description": description,
             "worker-type": "beetmover",
             "worker": task["worker"],
             "scopes": [
-                # TODO: Adjust scope
-                "project:mobile:fenix:releng:beetmover:bucket:dep",
+                bucket_scope,
                 "project:mobile:fenix:releng:beetmover:action:direct-push-to-bucket",
```
escapewindow, line 67: Possibly let's also use `dep` for `level < 3`

taskcluster/fenix_taskgraph/util/scriptworker.py
```diff
@@ -274,6 +274,7 @@ def generate_beetmover_artifact_map(config, job, **kwargs):
         
         if job["attributes"]["build-type"] == "nightly":
             folder_prefix = upload_date.strftime("%Y/%m/%Y-%m-%d-%H-%M-%S-")
+            version = version.split('-')[0]
```
escapewindow, line 277: Let's add a comment to remove this when version.txt is accurate.

escapewindow (changes requested):

taskcluster/fenix_taskgraph/transforms/beetmover.py
escapewindow: This is a candidate for a `by-level / by-build-type` nested config :)

hneiva: we return straight json here, where would `by-*` keys be calculated?

escapewindow: We could have the `by-` config in the kind.yml under some extra key, and here in the transform we can pop that key, evaluate the keyed_by, and set the scopes accordingly. Does that make sense?

escapewindow: I was thinking more:

[here](https://github.com/mozilla-mobile/fenix/blob/8de4c0b4dbcc185c522903f321030770591c0198/taskcluster/ci/beetmover/kind.yml#L24):
```
job-template:
    scopes:
        by-level:
            '3':
                 by-build-type:
                     'nightly':  # nightly
                     (release|beta):  # release
                      default:  # dep?
            default:  # use dep
```

Thoughts?

taskcluster/ci/beetmover/kind.yml
```diff
@@ -26,3 +26,11 @@ job-template:
...
     treeherder:
         job-symbol: BM
+    bucket-scope:
+        by-level:
+            '3':
+                by-build-type:
+                    'nightly': "project:mobile:fenix:releng:beetmover:bucket:nightly"
+                    (release|beta): "project:mobile:fenix:releng:beetmover:bucket:release"
+                    default: "project:mobile:fenix:releng:beetmover:bucket:dep"
+            default: "project:mobile:fenix:releng:beetmover:bucket:dep"
```
escapewindow, line 36: Perfect. Have you run taskgraph-diff to make sure the scopes look right in each graph?

taskcluster/ci/config.yml
```diff
@@ -55,7 +55,8 @@ workers:
             provisioner: scriptworker-k8s
             implementation: scriptworker-beetmover
             os: scriptworker
-            worker-type: 'mobile-1-beetmover-dev'
+            # Note: if dep bucket is wanted, *temporarily* switch to 'mobile-1-beetmover-dev'
+            worker-type: 'mobile-{level}-beetmover'
```
escapewindow, line 59: We may want to verify our scopes.

escapewindow (approved):

Issue #22508, opened 2021-11-19
[Bug]: The "Remove" button has scrollable text when using holding gesture in "Recent searches" categories


mariansima:
### Steps to reproduce

1. Create some recent search categories on the app homepage
2. Hold a finger down on a recent search group to display the contextual "Remove" button
3. Hold down a finger on the "Remove" button
4. Scroll up and down

### Expected behaviour

The text inside the button is static

### Actual behaviour

The text on the button can be scrolled up and down

https://user-images.githubusercontent.com/89511597/142649932-f702514f-03eb-415c-afcc-183be2543319.mp4

### Device name

Samsung Galaxy A8

### Android version

Android 9

### Firefox release type

Firefox Nightly

### Firefox version

96.0a1 2021-11-19

sflorean:
Verified as fixed on latest Nightly build with Pixel 3 (Android 12).

PR #21298, opened 2021-09-14
For #21287: Long press menu on homescreen

eliserichards:
For #21287

- [x] long press "remove" menu on homescreen "Recently visited" groups
- [x] tests
- [x] updated API

[N/A] long press "remove" for "recently visited" individual items - this functionality isn't being implemented
[N/A] select + delete items from history 3-dot menu -> @gabrielluong taking care of this
[N/A] select + delete All items from history 3-dot -> @gabrielluong taking care of this

### Pull Request checklist

- [x] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [x] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features. In addition, it includes a screenshot of a successful [accessibility scan](https://play.google.com/store/apps/details?id=com.google.android.apps.accessibility.auditor&hl=en_US) to ensure no new defects are added to the product.

### To download an APK when reviewing a PR:
1. click on Show All Checks,
2. click Details next to "Taskcluster (pull_request)" after it appears and then finishes with a green checkmark,
3. click on the "Fenix - assemble" task, then click "Run Artifacts".
4. the APK links should be on the left side of the screen, named for each CPU architecture

PR #22882, opened 2021-12-16
For #22508 - Increase menu height to avoid it become scrollable

Mugurell:
Compose enforces a DropdownMenuItem having at least a 64dp height
(48 dp height + 8dp vertical padding)
so we need to ensure the parent - DropDownMenu has at least 64dp heigh also.

```
internal val DropdownMenuVerticalPadding = 8.dp
private val DropdownMenuItemDefaultMinWidth = 112.dp
private val DropdownMenuItemDefaultMaxWidth = 280.dp
private val DropdownMenuItemDefaultMinHeight = 48.dp
```
https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/material/material/src/commonMain/kotlin/androidx/compose/material/Menu.kt;l=190-193;drc=e62838f496d592c019a3539669a9fbfd33928121

cc @topotropic 

https://user-images.githubusercontent.com/11428869/146376855-6fb7ca64-0eba-4b69-95b4-e580978798b6.mp4

| before | after |
| --- | --- |
|  |  |

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/home/recentvisits/view/RecentlyVisited.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/recentvisits/view/RecentlyVisited.kt
@@ -6,11 +6,8 @@ package org.mozilla.fenix.home.recentvisits.view
 
 import androidx.compose.foundation.ExperimentalFoundationApi
 import androidx.compose.foundation.Image
-import androidx.compose.foundation.ScrollState
 import androidx.compose.foundation.background
 import androidx.compose.foundation.combinedClickable
-import androidx.compose.foundation.gestures.Orientation
-import androidx.compose.foundation.gestures.scrollable
 import androidx.compose.foundation.isSystemInDarkTheme
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
@@ -21,7 +18,6 @@ import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.fillMaxHeight
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
-import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.layout.width
@@ -315,12 +311,6 @@ private fun RecentlyVisitedMenu(
         onDismissRequest = { onDismissRequest() },
         modifier = Modifier
             .background(color = FirefoxTheme.colors.layer2)
-            .height(52.dp)
-            .scrollable(
-                state = ScrollState(0),
-                orientation = Orientation.Vertical,
-                enabled = false
-            )
     ) {
         for (item in menuItems) {
             DropdownMenuItem(
@@ -328,21 +318,14 @@ private fun RecentlyVisitedMenu(
                     onDismissRequest()
                     item.onClick(recentVisit)
                 },
-                modifier = Modifier.fillMaxHeight()
             ) {
                 Text(
                     text = item.title,
                     color = FirefoxTheme.colors.textPrimary,
                     maxLines = 1,
                     modifier = Modifier
-                        .align(Alignment.Top)
-                        .padding(top = 6.dp)
-                        .scrollable(
-                            state = ScrollState(0),
-                            orientation = Orientation.Vertical,
-                            enabled = false
-                        )
                         .fillMaxHeight()
+                        .align(Alignment.CenterVertically)
                 )
             }
         }


app/src/main/java/org/mozilla/fenix/home/recentvisits/view/RecentlyVisited.kt
```diff
@@ -315,7 +315,7 @@ private fun RecentlyVisitedMenu(
         onDismissRequest = { onDismissRequest() },
         modifier = Modifier
             .background(color = FirefoxTheme.colors.layer2)
-            .height(52.dp)
+            .height(64.dp) // minimum height of a menu with one item
```
gabrielluong, line 318: I recall from talking @eliserichards that this 52dp is intentional. I can't find the commit history of this change at the moment, but I think we want to double check this.

But as aside, I am also wondering why our refactor of RecentlyVisited from #22534 did not preserve the commit history for this file. 😦

app/src/main/java/org/mozilla/fenix/home/recentvisits/view/RecentlyVisited.kt
```diff
@@ -315,7 +315,7 @@ private fun RecentlyVisitedMenu(
...
             .scrollable(
```
gabrielluong, line 319: I also wondering if we need this `scrollable` modifier

app/src/main/java/org/mozilla/fenix/home/recentvisits/view/RecentlyVisited.kt
Mugurell: 52 dp might've been intentional but it's according to the compose specs.
64 is the absolute minimum based on how the DropdownMenu works.

app/src/main/java/org/mozilla/fenix/home/recentvisits/view/RecentlyVisited.kt
Mugurell: That seemed to me like a previous attempt to fix this bug. Which didn't really work.

app/src/main/java/org/mozilla/fenix/home/recentvisits/view/RecentlyVisited.kt
Mugurell: The history continues to be available in git, just need to jump through more commits.
Found the ticket/PR in which this was implemented - https://github.com/mozilla-mobile/fenix/pull/21298

app/src/main/java/org/mozilla/fenix/home/recentvisits/view/RecentlyVisited.kt
gabrielluong: Right, this is also my understanding that we added these `scrollable` to attempt to fix the bug. So, I think we should try and revert some of it if it is no longer necessary if the 64.dp will make things work nicely

app/src/main/java/org/mozilla/fenix/home/recentvisits/view/RecentlyVisited.kt
gabrielluong: Ah yes, this is the PR I have attempted to find, but could not actually find it for the life of me.

I am wondering now if we even need to specify a given `height` if it's already the default in the compose implementation.

Mugurell: I'd be fine either way.
Though since this was a bug and an interesting discussion I'd say having this height value and comment here, just for brevity and future reference would be nice.

app/src/main/java/org/mozilla/fenix/home/recentvisits/view/RecentlyVisited.kt
Mugurell: Thank you!
Deleted the scrollable modifier from below but somehow missed this.
Tested with this modifier also removed and it works as expected.

app/src/main/java/org/mozilla/fenix/home/recentvisits/view/RecentlyVisited.kt
gabrielluong:
```suggestion
            .height(64.dp) // The absolute minimum height of a menu with one item #22508
```
Just one suggested change for the comment if we aren't removing the `height`. Keeping this does make me a bit nervous for any Compose version bump in the future since these defaults could change, and we end up with some unexpected behaviour because of the hardcoded height. The alternative would be to add a note in the kDoc.

Mugurell: Good point. Removed the hardcoded height.
Tested the it works as expected.

gabrielluong (approved):

Issue #22736, opened 2021-12-07
[Bug]: Dismissing the keyboard from the toolbar will hide the 3-dot menu


mariansima:
### Steps to reproduce

1. Launch Firefox and tap on the toolbar to bring up the keyboard
2. Tap on the screen to dismiss the keyboard
3. Observe the toolbar

### Expected behaviour

The keyboard is dismissed and the 3-dot menu button is displayed

### Actual behaviour

The keyboard is dismissed but only the search bar is visible

https://user-images.githubusercontent.com/89511597/145047542-43700158-1049-4a3b-836f-92fdbb34c686.mp4

### Device name

Samsung Galaxy A8

### Android version

Android 9

### Firefox release type

Firefox Nightly

### Firefox version

96.0a1 2021-12-07

### Additional information

This issue does not occur when using the back button. Could not reproduce on RC build

SoftVision-LorandJanos:
This issue is reproducible on the latest builds:

- latest Release (95.1.0).
- latest Beta (96.0.0-beta1).
- latest Nightly 97.0a1 (2021-12-10).

Device used: Xiaomi Mi11 Lite (Android 11).

mcarare:
After trying some fixes I came upon some questions that might need answering before trying to tackle this, by better describing the expected app behavior to these user actions.

- Displaying home behind the search fragment means that when you scroll, touch the home area you do not actually leave the search screen/ fragment. So I am not sure having the menu on the toolbar is expected, as you are actually still in the search screen.
- If we choose to navigate to home and as a result also display the menu it should be expected that the first screen interaction would just handle the navigation and keyboard dismissal and we would have the problem described in [this](https://github.com/mozilla-mobile/fenix/issues/20701) issue.

So it is a matter of choice between one behavior or the other and as it is something that can be very subjective I think UX should weigh in on this, having also in mind the context of showing home behind search reasoning.

cc: @topotropic

rocketsroger:
After some investigation, I have the same conclusion as @mcarare.  There's no good way to keep both the scrolling behaviour that we want but to show the 3-dot menu if the user does not press back.  I've put up a PR that will not remove focus from the toolbar so the user will have some hint that they are still in input mode.  Hopefully this way it will give the user enough hint that they need to press back or tap on the toolbar to continue their input.

topotropic:
[...]

Issue #20701, opened 2021-08-05
Dismiss keyboard when scrolling home behind search dialog


eliserichards:
Followup to #20573

This should be the same as iOS. The keyboard should close as you scroll on home, and you should be able to close the keyboard with the back button. 

Single back tap: close keyboard
Back again: go back to your tab

STR: 
1. Open wikipedia
2. Tap on the URL and see Home behind the searchDialogFragment
3. Scroll home
4. The keyboard should close smoothly as you scroll

PR #22883, opened 2021-12-16
For #22736: Dismiss the fragment on touching home screen behind search.

mcarare:
Just dismissing the keyboard leaves the user in an incomplete state, where search cannot be used without another tap on the toolbar, while that still is in edit mode instead of display mode.

This change follows the same behavior as pressing the back button. 

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/search/SearchDialogFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/search/SearchDialogFragment.kt
@@ -17,7 +17,9 @@ import android.os.Build
 import android.os.Bundle
 import android.speech.RecognizerIntent
 import android.text.style.StyleSpan
+import android.view.GestureDetector
 import android.view.LayoutInflater
+import android.view.MotionEvent
 import android.view.View
 import android.view.ViewGroup
 import android.view.ViewStub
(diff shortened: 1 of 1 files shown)


Amejia481 (commented):
The change looks good.
I just would like to make sure the change doesn't regress to #20701 as it looks @rocketsroger 
changed `dismissAllowingStateLoss` on  https://github.com/mozilla-mobile/fenix/commit/3d33a9422cd443a5a0b105a2a822f37b73c0d411

Would you mind confirming ? :)

mcarare:
> 3d33a94

Yes, I tested that. https://github.com/mozilla-mobile/fenix/commit/3d33a9422cd443a5a0b105a2a822f37b73c0d411 is what introduced this issue (#22736)

rocketsroger:
> Yes, I tested that. [3d33a94](https://github.com/mozilla-mobile/fenix/commit/3d33a9422cd443a5a0b105a2a822f37b73c0d411) is what introduced this issue (#22736). Dismissing the dialog fragment also dismisses the keyboard.

Thanks for fixing this.  I'll try it out and review.

mcarare:
Thank you @rocketsroger! It is good to have another pair of eyes on this!

rocketsroger (approved):
After some testing, I think we have to land this first to fix the issue at hand.  However, this fix does re-introduce the problem where the user's first scroll will be ignored and have to re-scroll again.  I will try to figure out another solution for that problem.

🚢 Thanks for fixing this.

rocketsroger:
@mcarare actually after playing with this a bit more, the first scroll being ignore is a bit too high of a cost to fix my regression.  Do you mind trying out other fixes to see if we can get a fix that doesn't have the scroll problem?  Thanks

rocketsroger (changes requested):
I think the first scroll being ignore is not a good user experience.  If possible, can we find a better solution.

mcarare:
@rocketsroger what do you mean by "first scroll is ignored"? I have tested this, and on the very first touch/scroll the fragment is dismissed with this patch.

mcarare:
@rocketsroger I added a way to "dispatch" the scroll to the home screen if one is detected. Please let me know if it covers your scenario. Thank you!

mcarare:
After talking more offline with @rocketsroger it seems we would have to choose between one or the other behavior. After testing multiple solutions it seems that dismissing the fragment interferes with any kind of touch handling or dispatching. Even if a scroll would be started programmatically, it would be stoped when the search fragment is dismissed. Tried with delays, and changing action order, the same issue occurs.

Closing this and adding more details to the issue itself. 

Thanks, @rocketsroger for also testing this.

PR #22894, opened 2021-12-17
Closes #22764: Provide stable, sequential IDs for items returned by PagedHistoryProvider

### LINKED ISSUE #22764: [Bug]: The same search group is displayed multiple times in the History section
### Steps to reproduce

1. Search for a term
2. Open one of the results
3. Open history

### Expected behaviour

One search group should be displayed

### Actual behaviour

The same search group is displayed multiple times

### Device name
OnePlus A3 Android 6.0.1

### Firefox version
Firefox 96.0.0 Beta 1
Firefox Nightly from 12/09

### Additional information

► Video

https://user-images.githubusercontent.com/51314259/145415480-59a08104-cec4-4619-b4f4-cb1a110ff8c2.mp4

grigoryk:
We were converting Long timestamps into Ints (and getting negative
numbers back), and treating that as, basically, a position for the
paging API; the paging API would pass us back the obscure negative
number back as an offset, and we'll mishandle it resulting in an
infinite loop.

This patch removes all of the Long -> Int conversions, and introduces an
explicit 'position' that is calculated once we have a full page of
results completed.

### Pull Request checklist

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/components/history/PagedHistoryProvider.kt
```diff
@@ -94,14 +93,22 @@ class DefaultPagedHistoryProvider(
...
             }
 
-            onComplete(history)
+            val positionedHistory = history.mapIndexed { index, item ->
+                when(item) {
+                    is History.Group -> item.copy(position = offset + index + item.items.size)
+                    is History.Metadata -> item.copy(position = offset + index)
+                    is History.Regular -> item.copy(position = offset + index)
+                }
+            }
```
gabrielluong, line 109: Any way we add a check for the positions in our existing tests?

grigoryk: Oh, yeah, I should have marked this as a WIP. Still not 100% of the approach. I _think_ this is what we want, and I _think_ this really improves a few cases (including the bug), but need to test/think a bit more.

sarah541 (changes requested):

app/src/main/java/org/mozilla/fenix/components/history/PagedHistoryProvider.kt
grigoryk: Found a bug in the logic while re-reading the PR - this needs to be converted to a for-loop. E.g. if first item is a group of 10 items, and second item is a regular history item, current behavior will assign position=2 (or 1..? not sure if it's 0-indexed) to the second item, while it _should_ assign 11.

Not a huge deal, just something to fix before landing & cover w/ tests.

grigoryk:
All right, the tests are now all in place.

The current approach has two limitations:
1) incorrect handling of zero-sized groups, which we won't actually encounter in practice since we don't have zero-size groups,
2) awkward behaviour at the boundaries of pages - the Android `ItemKeyedDataSource` API takes the largest `position` value we provide in our results (i think..) and uses that as the next offset. However, while processing groups, currently we will "collapse" visits if they're present in search groups - if a URL is present in a group, we'll filter it out from the current page of history visits. So, for example, a group of 5 items may remove more than 5 items from history visits. That means our position will be a technically incorrect offset (we're under-counting as far as history visits paging API is concerned). In practice though, this probably doesn't matter and I think this is fine _as is_. Worth thinking about this more though.

mavduevskiy (approved):

grigoryk:
All right, refactored types to remove all of the `!!` bits, much cleaner now. Appears to work fairly well locally, too.

PR #22912, opened 2021-12-20
For #18160: Show a prompt when trying to leave private browsing with …

mavduevskiy:
For #18160: Show a prompt when trying to leave private browsing with active downloads

Relates to [...fenix/18160](https://github.com/mozilla-mobile/fenix/issues/18160) and [android-components/11411](https://github.com/mozilla-mobile/android-components/issues/11411)

It appears that we have three scenarios when all private tabs are being closed by the user.
1) manually closing the tabs individually
2) closing all tabs from drop down menu
3) taping the private notification with action "close all"

First two cases will show the prompt, warning the user that closing the tab/tabs will also cancel the active private downloads. The DownloadCancelDialogFragment was added to the AC project, the displaying logic is implemented inside fenix. The third cases remains untouched, as the app might be in background and have to access to the UI.

What changed: analytics from interactor layer was moved to controller, since not every request to close the tab actually results with the tab being closed. So now the controller, after actually closing the tab, will fire an analytics event.

Added tests.

Screenshots could be found [here](https://github.com/mozilla-mobile/android-components/pull/11412).


app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayController.kt
```diff
@@ -144,18 +159,41 @@ class DefaultTabsTrayController(
...
                 showUndoSnackbarForTab(it.content.private)
             } else {
-                dismissTabsTrayAndNavigateHome(tabId)
+                val privateDownloads = browserStore.state.downloads.filter { map ->
+                    map.value.private && (
+                        map.value.status == DownloadState.Status.INITIATED ||
+                            map.value.status == DownloadState.Status.DOWNLOADING ||
+                            map.value.status == DownloadState.Status.PAUSED
+                        )
+                }
```
Amejia481, line 188: Maybe we could have a extension function to reuse this code (arePrivateOngoingDownloads, it looks the same as what we have here https://github.com/mozilla-mobile/fenix/pull/22912/files#diff-5912831edfde6ad16931cf998b691e99e6e5c81d6e8887b73a0421efc842f272R185-R191

app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayFragment.kt
```diff
@@ -130,18 +133,20 @@ class TabsTrayFragment : AppCompatDialogFragment() {
                 initialState = TabsTrayState(
                     mode = initialMode,
                     focusGroupTabId = args.focusGroupTabId
-                ),
-                middlewares = listOf(
-                    TabsTrayMiddleware(
-                        metrics = requireContext().metrics
-                    )
```
Amejia481, line 137: Is this change intentional? :)

Amejia481 (approved):

Issue #22415, opened 2021-11-11
[Bug][a11y]: Talkback reads "Button" twice for the "Show all" button on homescreen


SoftVision-LorandJanos:
### Steps to reproduce

1. Turn TalkBack (or the available screen reader) on.
2. Have some items on the "Jump back in", "Recent bookmarks" and "Recent searches" sections.
3. Go to homescreen and tap on the available "Show all" buttons.

### Expected behaviour

The "button" is read once.

### Actual behaviour

The "button" is read twice.

https://user-images.githubusercontent.com/48829350/141304272-47445870-f3bc-4dd2-b240-5cfaef76ff84.mp4

### Device name

Xiaomi Mi11 Lite

### Android version

Android 11

### Firefox release type

Firefox Beta

### Firefox version

95.0.0 beta2

mcarare:
Note for QA: After https://github.com/mozilla-mobile/fenix/pull/22650/commits/a71622acaa9b71ef688aa4d65391d73855e9f5d9 the button announcment will not be read.

However, it is considered bad practice to add the state or type in a content description for an element. That is because the role announcement for elements is subject to change and it will provide a bad experience across the apps if we hardcode that role. 

Historically speaking, the button type announcement has been deprecated but kept for legacy support. The relevant announcement for a button or any other clickable element is "double tap to... ". This is the only announcement that is required for a clickable element and we should check for that after https://github.com/mozilla-mobile/fenix/pull/22938 lands.

PR #22938, opened 2021-12-21
For #22415: Remove element type from content descriptions.

mcarare:
Adding the type to content description is not advised according to a11y guidelines.
See https://support.google.com/accessibility/android/answer/7158690?hl=en.

See also more on the [issue](https://github.com/mozilla-mobile/fenix/issues/22415#issuecomment-998920691).
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -55,7 +55,9 @@
     <!-- Title for the button which navigates the user to show all of their saved bookmarks. -->
     <string name="recently_saved_show_all">Show all</string>
     <!-- Content description for the button which navigates the user to show all of their saved bookmarks. -->
-    <string name="recently_saved_show_all_content_description">Show all saved bookmarks button</string>
+    <string name="recently_saved_show_all_content_description_2">Show all saved bookmarks</string>
+    <!-- Content description for the button which navigates the user to show all of their saved bookmarks. -->
+    <string name="recently_saved_show_all_content_description" moz:removedIn="97" tools:ignore="UnusedResources">Show all saved bookmarks button</string>
 
     <!-- About content. The first parameter is the name of the application. (For example: Fenix) -->
     <string name="about_content">%1$s is produced by Mozilla.</string>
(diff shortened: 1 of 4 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -55,7 +55,7 @@
     <!-- Title for the button which navigates the user to show all of their saved bookmarks. -->
     <string name="recently_saved_show_all">Show all</string>
     <!-- Content description for the button which navigates the user to show all of their saved bookmarks. -->
-    <string name="recently_saved_show_all_content_description">Show all saved bookmarks button</string>
+    <string name="recently_saved_show_all_content_description2">Show all saved bookmarks</string>
```
gabrielluong, line 58: We have to keep the old strings and tag it for removal. See https://github.com/mozilla-mobile/fenix/wiki/Removing-strings

Can change the suffix to `_2`?
```suggestion
    <string name="recently_saved_show_all_content_description_2">Show all saved bookmarks</string>
```

gabrielluong (changes requested):

app/src/main/res/values/strings.xml
mcarare: Thank you for the link, I was not aware the procedure has changed. It seems a bit overcomplicated. 
Maybe if the automation part for updating strings that need a translation in Pontoon would be improved we could avoid jumping through these hoops on Fenix / Focus.

app/src/main/res/values/strings.xml
```diff
@@ -55,7 +55,8 @@
     <!-- Title for the button which navigates the user to show all of their saved bookmarks. -->
     <string name="recently_saved_show_all">Show all</string>
     <!-- Content description for the button which navigates the user to show all of their saved bookmarks. -->
-    <string name="recently_saved_show_all_content_description">Show all saved bookmarks button</string>
+    <string name="recently_saved_show_all_content_description_2">Show all saved bookmarks</string>
+    <string name="recently_saved_show_all_content_description" moz:removedIn="97" tools:ignore="UnusedResources">Show all saved bookmarks button</string>
```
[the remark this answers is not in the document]
mcarare: I really do not think that is a real issue, but I will change it.

gabrielluong (approved):

Issue #22939, opened 2021-12-21
[Bug]: Firefox homepage's initial scroll position is random on Android 8.1 device


cpeterso:
### Steps to reproduce

1. Make sure Firefox is configured to open to the Firefox homepage.
2. Cold or warm start Fenix on a device running Android 8.1.

(I can reproduce this bug in Firefox 95, Beta 96, and Nightly 97 on my Moto G5 Plus running Android 8.1.)

### Expected behaviour

The homepage should either always start at the top or its previous scroll position.

### Actual behaviour

* In a cold app start, the homepage is always positioned at "Stories by topic" near the bottom of the page.
* In a warm app start after my recent homepage scroll position was showing the "Recently visited" section or above, the homepage will start at some random position near the second row of Top Sites (with the first row of Top Sites usually off screen).
* In a warm app start after my recent homepage scroll position was showing below the "Recently visited" section (so only showing the "Thought-provoking stories" or "Stories by topic" sections), the homepage usually restores that scroll position correctly.

@rocketsroger can reproduce this bug in an emulator running Android 8.1, but not Android 10. TBD whether this bug also affects Android 5.0-8.0 or Android 9.

This is a bug where different data sources are racing.

Similar to https://github.com/mozilla-mobile/fenix/issues/22402

### Device name

Moto G5 Plus

### Android version

Android 8.1

### Firefox release type

Firefox Nightly

### Firefox version

97

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-18973)

rocketsroger:
There is one case when this can also be reproduce in Android 10+.  When the user first starts using Fenix, after they selected their Theme and Toolbar position, home is not scrolled to the top.  I'll also fix it with this issue.

rocketsroger:
Tried different workarounds and they all causes regression in other areas.  The best solution for this issue is to implement https://github.com/mozilla-mobile/fenix/issues/22755.

Issue #22755, opened 2021-12-08
Improve homescreen rendering time


mcomella:
We've recently added a lot of functionality to the homescreen and that has understandably slowed down the time to render its content (see https://github.com/mozilla-mobile/fenix/issues/21749 for details). Let's **investigate to see if there are optimizations we can make**.

csadilek suggested a general pattern to improve rendering of multiple views populated by async content: we could cache the data in memory, allowing us to fetch it synchronously while removing redundant layout/measure/draw computation.

See https://github.com/mozilla-mobile/fenix/issues/21749#issuecomment-936975793 for some profiles of the homescreen before and after the homescreen content was added.

Here are some sub-tickets for what we can optimize:
- https://github.com/mozilla-mobile/fenix/issues/14443
- https://github.com/mozilla-mobile/fenix/issues/14252

┆Issue is synchronized with this [Jira Epic](https://mozilla-hub.atlassian.net/browse/FNXV2-18864)

csadilek:
Initially we want to solve the following two problems:
- The state needed to render the home screen is computed asynchronously _every time_ the home fragment is opened
- The sections on the screen are rendered whenever the corresponding state is available causing sections to appear non-deterministically (instead of from top to bottom).

We can solve this by storing the required state globally (and cache it) as opposed to computing it from an empty state every time the home fragment is needed. In https://github.com/mozilla-mobile/fenix/issues/20704 we have introduced a new `AppStore` that can hold state for the entire lifecycle of the app, which is exactly what we need here. So, the plan is to:
- Move all state needed from the home screen to the global `AppStore` and use it as a cache
- Render all sections from top to bottom based on the state in the `AppStore`
- Have the UI observe state changes on the `AppStore` so they can update automatically after the initial rendering

This will address the problems, except on cold start to HOME when the `AppStore` is not yet populated. This can be addressed in a follow-up by introduce a persistent cache, although it's likely that this trade-off is worth it i.e., reading from disc on cold start takes time too. Another option is an UI affordance e.g. a loading indicator for the home screen.

gabrielluong:
Our first attempt to validate this approach and pattern will be from migrating Top Sites over to the AppStore in https://github.com/mozilla-mobile/fenix/issues/23306.

Mugurell:
@gabrielluong @mcomella After talking with @csadilek about https://github.com/mozilla-mobile/fenix/issues/21854 he recommended migrating to compose each section separately and then add AppStore incrementally too.

[...]

Issue #22462, opened 2021-11-16
[Bug]: Scroll to collection is not working


rocketsroger:
### Steps to reproduce

1. Open tabs tray
2. Select a few tabs and add to collection
3. Tap on view collection

### Expected behaviour

Home is shown and scrolled to the collection

### Actual behaviour

Home is shown but not scrolled to the collection

### Device name

Google Pixel 3 XL

### Android version

Android 12

### Firefox release type

Firefox Nightly

### Firefox version

96.0a1

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-18647)

LaurentiuApahideanSV:
I can confirm the issue is reproducible on the latest Firefox Preview Nightly 2021-11-23.
### Devices used:
- Oppo Find X3 (Android 11)
- OnePlus 6T (Android 9)

PR #22940, opened 2021-12-21
Issue #22939: Ensure Home always starts from the top

rocketsroger:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlView.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlView.kt
@@ -202,6 +202,8 @@ class SessionControlView(
                 override fun onLayoutCompleted(state: RecyclerView.State?) {
                     super.onLayoutCompleted(state)
 
+                    // scroll to top placeholder
+                    view.scrollToPosition(0)
                     JumpBackInCFRDialog(view).showIfNeeded()
                 }
             }


app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlView.kt
```diff
@@ -202,6 +202,8 @@ class SessionControlView(
                 override fun onLayoutCompleted(state: RecyclerView.State?) {
                     super.onLayoutCompleted(state)
 
+                    // scroll to top placeholder
+                    view.scrollToPosition(0)
```
jonalmeida, line 206: I'm surprised we didn't do this earlier as part of the original fix.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlView.kt
```diff
@@ -202,6 +202,8 @@ class SessionControlView(
                 override fun onLayoutCompleted(state: RecyclerView.State?) {
```
jonalmeida, line 202: Put the scrollToPosition here would mean that _every_ `view.layout` would scroll to the top, which may be undesirable. For example:

1. Scroll down to 'Recently visited'.
2. Click the 'Show all'.
3. Press the back button.

The expected result is that you are in the same scroll position as before, but now you are always taken to the top.

I haven't tried it yet, but maybe what you want is the same thing we do in the tabstray: perform a one-time scroll to position on layout and then remove the listener.

See: [`<VH : RecyclerView.ViewHolder> RecyclerView.Adapter<out VH>.observeFirstInsert`](https://github.com/mozilla-mobile/fenix/blob/main/app/src/main/java/org/mozilla/fenix/tabstray/ext/RecyclerViewAdapter.kt)

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlView.kt
csadilek: I remember we actually tried and talked about this. @rocketsroger do you recall? Was it because it didn't work without the new top placeholder which we added and then thought we don't need to scroll anymore?

csadilek: This was I also when we discovered https://github.com/mozilla-mobile/fenix/issues/22462 which would be hard to fix if we scrolled to the top on layout completion.

rocketsroger: Yes, we didn't add this to the original fix because on Android 10 we didn't see a need for it.  We stay on the top as long as top placeholder was populated right away.

However, on 8.1 it is definitely not the same behaviour.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlView.kt
rocketsroger: I tested it but it didn't scroll up.  It stay at the same scroll position.

jonalmeida: Hmm.. for me on Android 12 (Pixel 3) it jumps to the top in a flashing sort of way. Hmm.. it seems to do that when switching apps too.

Here's a video of what I see with your change (the first app is a debug build, the second is current nightly):

https://user-images.githubusercontent.com/1370580/147001598-aff225d5-c5b3-414f-9c7c-c99b5a5514d2.mp4

rocketsroger: That's weird.  I can't reproduce that.  I'll add the unregister oberver code to be safe though.  Not sure why i can't reproduce it with Android 10.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlView.kt
csadilek: I guess another way of asking. Is the top place holder even needed if we call `scrollToPosition(0)` or would that work  without it? 

Given #22462 and if the workarounds here pile up, it might make sense to wait and address the bigger refactoring to make sure we render home in one pass from top to bottom (and from cache). However, if this makes it much better, it's OK to land in the interim.

rocketsroger: Unfortunately, this workaround shows the reason why we didn't go this route in the first place.  There are multiple cases that this will scroll up in the wrong time.

rocketsroger:
The side effect of this workaround is not acceptable.  I am closing this PR.  Once we implements https://github.com/mozilla-mobile/fenix/issues/22755, this issue will be fixed as well.

PR #22941, opened 2021-12-21
For #22851: Adjust theme based on the selected wallpaper.

Amejia481:
The idea with this PR is adjust the theme based on the selected wallpaper. In a follow up we can explore the option to do that dynamically maybe using https://developer.android.com/training/material/palette-colors
 
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperManager.kt
+++ b/app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperManager.kt
@@ -4,8 +4,11 @@
 
 package org.mozilla.fenix.wallpapers
 
+import android.content.Context
 import android.view.View
+import androidx.appcompat.app.AppCompatDelegate
 import mozilla.components.support.ktx.android.content.getColorFromAttr
+import org.mozilla.fenix.ext.asActivity
 import org.mozilla.fenix.utils.Settings
 
 /**
@@ -30,6 +33,38 @@ class WallpaperManager(private val settings: Settings) {
             wallpaperContainer.setBackgroundResource(newWallpaper.drawable)
         }
         currentWallpaper = newWallpaper
+
+        adjustTheme(wallpaperContainer.context)
+    }
+
+    private fun adjustTheme(context: Context) {
+        val mode = if (currentWallpaper != Wallpaper.NONE) {
+            if (currentWallpaper.isDark) {
+                updateThemePreference(useDarkTheme = true)
+                AppCompatDelegate.MODE_NIGHT_YES
+            } else {
+                updateThemePreference(useLightTheme = true)
+                AppCompatDelegate.MODE_NIGHT_NO
+            }
+        } else {
+            updateThemePreference(followDeviceTheme = true)
+            AppCompatDelegate.MODE_NIGHT_FOLLOW_SYSTEM
+        }
+
+        if (AppCompatDelegate.getDefaultNightMode() != mode) {
+            AppCompatDelegate.setDefaultNightMode(mode)
+            context.asActivity()?.recreate()
+        }
+    }
+
+    private fun updateThemePreference(
+        useDarkTheme: Boolean = false,
+        useLightTheme: Boolean = false,
+        followDeviceTheme: Boolean = false
+    ) {
+        settings.shouldUseDarkTheme = useDarkTheme
+        settings.shouldUseLightTheme = useLightTheme
+        settings.shouldFollowDeviceTheme = followDeviceTheme
     }
 
     /**
--- a/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
+++ b/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
@@ -490,7 +490,7 @@ class Settings(private val appContext: Context) : PreferencesHolder {
         }
     }
 
-    val shouldUseDarkTheme by booleanPreference(
+    var shouldUseDarkTheme by booleanPreference(
         appContext.getPreferenceKey(R.string.pref_key_dark_theme),
         default = false
     )
--- a/app/src/main/java/org/mozilla/fenix/wallpapers/Wallpaper.kt
+++ b/app/src/main/java/org/mozilla/fenix/wallpapers/Wallpaper.kt
@@ -9,8 +9,8 @@ import org.mozilla.fenix.R
 /**
  * A enum that represents the available wallpapers and their states.
  */
-enum class Wallpaper(val drawable: Int) {
-    FIRST(R.drawable.wallpaper_1),
-    SECOND(R.drawable.wallpaper_2),
-    NONE(R.attr.homeBackground);
+enum class Wallpaper(val drawable: Int, val isDark: Boolean) {
+    FIRST(drawable = R.drawable.wallpaper_1, isDark = true),
+    SECOND(drawable = R.drawable.wallpaper_2, isDark = false),
+    NONE(drawable = R.attr.homeBackground, isDark = false);
 }


app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperManager.kt
```diff
@@ -30,6 +33,44 @@ class WallpaperManager(private val settings: Settings) {
...
+        }
+
+        AppCompatDelegate.setDefaultNightMode(mode)
+        context.asActivity()?.recreate()
+    }
+
+    private fun setDarkThemePreference() {
+        settings.shouldUseDarkTheme = true
+        settings.shouldUseLightTheme = false
+        settings.shouldFollowDeviceTheme = false
```
gabrielluong, line 61: Can we not just replace all of this with a single setting that uses a String enum for the tri-state?

gabrielluong: Okay, I am now see this was already something existing.

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperManager.kt
```diff
@@ -30,6 +33,44 @@ class WallpaperManager(private val settings: Settings) {
...
+    private fun setLightThemePreference() {
+        settings.shouldUseLightTheme = true
+        settings.shouldUseDarkTheme = false
+        settings.shouldFollowDeviceTheme = false
+    }
+
+    private fun setFollowDeviceThemePreference() {
+        settings.shouldUseLightTheme = true
+        settings.shouldUseDarkTheme = false
+        settings.shouldFollowDeviceTheme = false
```
gabrielluong, line 73: Probably better to refactor these methods into 1 and just take the shouldUseLightTheme, shouldUseDarkTheme an d shouldFollowDeviceTheme as parameters

gabrielluong: Also, provide everything with a default value to `false` to simplify further.

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperManager.kt
```diff
@@ -30,6 +33,44 @@ class WallpaperManager(private val settings: Settings) {
...
+                AppCompatDelegate.MODE_NIGHT_NO
+            }
+        } else {
+            setFollowDeviceThemePreference()
+            AppCompatDelegate.MODE_NIGHT_FOLLOW_SYSTEM
+        }
+
+        AppCompatDelegate.setDefaultNightMode(mode)
+        context.asActivity()?.recreate()
+    }
```
MatthewTighe, line 56: Couple points:

- This seems like it's adjusting the theme for the entire app, is that correct? I believe what we want is to only adjust the home screen. I think we'll want theming elsewhere in the app to continue following the user preference.
- I still think it's worth investigating whether we can adjust only the text on the home screen. I really liked the demo Kate showed us where the text color changed as wallpapers changed, but other themed elements like surface backgrounds continued to follow the system theme. My thinking was to add a property along the lines of `Wallpaper.textColor`, and then we can propagate the `WallpaperManager` to areas of the home screen that need to investigate the current wallpaper.

Amejia481: > This seems like it's adjusting the theme for the entire app, is that correct? I believe what we want is to only adjust the home screen. I think we'll want theming elsewhere in the app to continue following the user preference.

I couldn't find a way to set dark or light theme to an specific view, what I found it's we could alter the theme to the activity level, but the `HomeActivity` is practically the only activity that we have. To be able to separate theme from the rest of the app we will need to create a new activity only for home and another for the rest of the app. 

> I still think it's worth investigating whether we can adjust only the text on the home screen. I really liked the demo Kate showed us where the text color changed as wallpapers changed, but other themed elements like surface backgrounds continued to follow the system theme. My thinking was to add a property along the lines of Wallpaper.textColor, and then we can propagate the WallpaperManager to areas of the home screen that need to investigate the current wallpaper.

I'll research more about this alternative!

MatthewTighe: I just worry that the user experience will suffer if we alter the entire theme of the app based on the wallpaper color. For example, if a dark mode user might expect things like the settings page and the address bar to remain in dark mode even if they have a light background for the homepage.

Amejia481: > I just worry that the user experience will suffer if we alter the entire theme of the app based on the wallpaper color. For example, if a dark mode user might expect things like the settings page and the address bar to remain in dark mode even if they have a light background for the homepage.

Checking on Safari desktop, it looks like they change the whole screen theme, depending on the background chosen. 
But opening a new tab changes back to the previous theme. 

I guess this could something that could be seen as good or bad, as the theme will be flashing between screen and not being consistent.  For example, in this case I have a dark background in the home screen, then I open a new tab, and I get flash between dark and light, when I switch between screens. I least I prefer to have consistent theme between screen. But this is something that we could ask to UX/UI for input :)

https://user-images.githubusercontent.com/773158/147146323-73ebefb9-ed09-4dff-8818-250b422c2a21.mov

MatthewTighe: I agree that having a consistent theme seems the best, but I'm still of the opinion that it makes sense for the theme to match the user-selected theme rather than dynamically changing it based on the wallpaper. 

That said, I think UI/UX input here is a great idea because I'm certainly no expert 😂

Amejia481: Me neither I just write code XD

Amejia481: > I just worry that the user experience will suffer if we alter the entire theme of the app based on the wallpaper color. For example, if a dark mode user might expect things like the settings page and the address bar to remain in dark mode even if they have a light background for the homepage.
> 
> I agree that it's going to be a tedious amount of text to update and then keep aligned in the future. Maybe there's additional alternatives we can discover that are better compromises


After some testing, I couldn't find a scalable way to inject the new color to text items on the home screen. 
The ones that I took look seen to make things complicated:

1. Have a global/screen state for the current wallpaper and make all the components on the home screen observe it and change it color based on changes. **We will need to change all the components that are showing on the home screen and all new UI components we will need to have this too.**
2. Have theme only for the home screen. **We will need to create a separate activity for the home screen, which could break the actual navigation experience, without taking into account that we are relying on multiple places that we only have a single activity.**

It looks like we could only go **with option 1** or **the option switching the app theme based on the background.** Taking into account the trade offs, **for me it looks like the option of switching the app theme looks more scalable and more natural to the user**. 

What do you think? Do you have other alternatives?

MatthewTighe: One alternative for option 1 is creating a new theme element, along with another set of themes. We could then have the following:

```
LightThemeWithLightWallpaper.textColorOnWallpaper = <dark text>
LightThemeWithDarkWallpaper.textColorOnWallpaper = <light text>
DarkThemeWithLightWallpaper.textColorOnWallpaper = <dark text>
DarkThemeWithDarkWallpaper.textColorOnWallpaper = <light text>
```

However, it sounded @gabrielluong was hesitant to add an additional element to the design tokens, so this alternative might not fly.

In either case, I'm not sure we can avoid updating every text element on the home screen with a new color. I agree that it's not easily scalable, but I haven't come up with anything that gets the desired behavior otherwise.

gabrielluong: You are free to add a new color variable, attributes or styles as needed. I will note that `fx_mobile_` prefix color variables (what I am calling the design tokens) do come from the design system, so I wouldn't necessarily create new design tokens if they aren't coming from the design system team.

Thinking about the problem here, I think we should go with the simplest solution which is to switch the entire app theme based on the chosen wallpaper. I think it's okay to be opinionated here and this helps us remove complexity given the time constraints and how there are no good solutions at the moment. I can imagine we had a red theme selected through colorways and a wallpaper applied, the red theme and wallpaper might not play nice either way.

If I was to think about a solution where we end programatically switching the text color based on the wallpaper for all the text on the home screen, I can come up with some scenarios that are bit iffy:
1. Imagine a shared / generic (Compose) component that might be used in the Home and another fragment (let's say Tab Tray), we will now need to add a special case to check the Wallpaper state when we provide the textColor parameter or alternatively pass the Wallpaper state to the component (less likely I imagine). This adds an extra maintenance burden in terms of always knowing to check the Wallpaper state for any text that is added to the Home screen. Of course, we can accept this technical burden, but you can imagine we would rather not if it can be avoided.
2.  I don't think a single variable `textColorOnWallpaper` will ever fully capture the desired text color for all light or dark (or red, purple, blue, black, yellow) wallpapers. I think we can end up in a scenario where we specify text colors for the choices of wallpapers. (Funny I saw this mentioned as your first comment after I re-read the current discussion as I was making this comment). I can imagine we end up calling the wallpaper manager to tell us the text color to apply if a wallpaper is enabled. This probably loops back to part 1 at this point where our components will need to have an extra parameter.

Ultimately, I don't think WE should making the final decision, but it should be a group decision based on the complexity that we face, time constraints and what UX/Product are happy with based on those constraints.

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperManager.kt
Amejia481: 👍🏽

Amejia481:
I updated the PR following the suggestions :)

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperManager.kt
```diff
@@ -30,6 +33,38 @@ class WallpaperManager(private val settings: Settings) {
...
+            }
+        } else {
+            updateThemePreference(followDeviceTheme = false)
+            AppCompatDelegate.MODE_NIGHT_FOLLOW_SYSTEM
+        }
+
+        if (AppCompatDelegate.getDefaultNightMode() != mode) {
+            AppCompatDelegate.setDefaultNightMode(mode)
+            context.asActivity()?.recreate()
+        }
```
MatthewTighe, line 57: Could this cause the `settings.shouldFollowDeviceTheme` and the `AppCompatDelegate` property to be out of sync? The theme preference would be updated with `followDeviceTheme = false`, but `AppCompatDelegate.setDefaultNightMode` will be called with `MODE_NIGHT_FOLLOW_SYSTEM`

Amejia481: This is just visual improvement to avoid recreating the fragment when it's not needed and avoid flashes, as we are only recreating the fragment/`setDefaultNightMode` when the mode is different of the actual one, I don't think it will get out sync, if it's the same we don't need to update it again and cause a flash.

Amejia481: > The theme preference would be updated with followDeviceTheme = false, but AppCompatDelegate.setDefaultNightMode will be called with MODE_NIGHT_FOLLOW_SYSTEM

In the above example, we are only going to update `setDefaultNightMode`/`asActivity()?.recreate()` if the actual mode (`AppCompatDelegate.getDefaultNightMode()) is different of the actual one, it's the same there is not need for updating it.

Imagine that we have three wallpapers in the following order `[wallpaper 1 is dark,  wallpaper 2 is dark, wallpaper 3 is light]`, if you are on light mode and you tap the logo three times:

1. Wallpaper 1 will be applied and we will call `setDefaultNightMode` to set dark mode.
2. Wallpaper 2 will be applied and we will **NOT call** `setDefaultNightMode` as we are already on dark mode.
3. Wallpaper 3 will be applied and we will call `setDefaultNightMode` to set light mode.

MatthewTighe: What happens when we wrap back around? Is the default wallpaper now light, even if the user originally was following a dark system theme?

Amejia481: OK, I see what do you mean.  I updated  `updateThemePreference` to `updateThemePreference(followDeviceTheme = true)`

MatthewTighe (approved):

PR #22944, opened 2021-12-22
fixes #22943: update androidx.lifecycle to 2.4.0

### LINKED ISSUE #22943: Upgrade androidx.lifecycle version to 2.4.0
The current suggested method of flow collection within UI elements is to use the `repeatOnLifecycle` API. This will automatically cancel the coroutines collecting from a flow during destructive lifecycle events and then relaunch the collection on constructive lifecycle events. This API became available in the above version.

MatthewTighe:
### Pull Request checklist

- [x] **Tests**: Relies on regression
- [x] **Screenshots**: No user-facing changes
- [x] **Accessibility**: No user-facing changes

1. click on Show All Checks,

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -317,7 +317,13 @@
         <provider
             android:name="androidx.startup.InitializationProvider"
             android:authorities="${applicationId}.androidx-startup"
-            tools:node="remove" />
+            android:exported="false"
+            tools:node="merge" >
+            <meta-data
+                android:name="androidx.work.WorkManagerInitializer"
+                android:value="androidx.startup"
+                tools:node="remove" />
+        </provider>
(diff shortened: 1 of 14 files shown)


MatthewTighe:
I think it makes sense to wait this until after the New Year

pocmo (commented):
I recommend to create (and test) a matching upgrade in A-C. Otherwise you'll hard-upgrade here and run the A-C code depending on `lifecycle` in an untested state.

MatthewTighe:
Opened a [PR](https://github.com/mozilla-mobile/android-components/pull/11473) for a matching A-C upgrade.

jonalmeida:
Once a new AC version is released, I'll update this PR with that version bump so we have to land just one PR instead of two (this + AC bump).

PR #22951, opened 2021-12-22

MickeyMoz:


mcarare:
All failed UI tests seem to be related to saved logins.

Amejia481:
We recently put a patch https://github.com/mozilla-mobile/android-components/commit/cc18078227d1e45a9f962987f1eb7cb46649444d for removing idle prompts fragments

mcarare:
@pocmo It seems that with the new change the save login prompt is immediately dismissed after the login occurs and the page navigates to a new page.
I also tried to log in on a real page and the prompt is dismissed after the login occurs, with no time to save the login from the prompt.

pocmo:
Thanks! I'll investigate. That indeed looks like fallout from the patch.

pocmo:
This was fixed in A-C. This PR should be unblocked now. :)

PR #22953, opened 2021-12-22
closes #22831: add wallpaper settings screen 

### LINKED ISSUE #22831: Create a settings page for allowing users to change home screen background

MatthewTighe:
Depends on #22944, which will reduce the scope of the PR to exclude the `androidx.lifecycle` update.

@jonalmeida and I briefly talked that the wallpaper state might better belong in the `HomeFragmentStore`, so this PR could serve as a good forum for discussion about that. Having that state accessible through the store could have additional benefits, such as easier access for text components to update their color like is being working on in #22941. It looks like there was already some discussion about this in #22851, actually.

image

image

### Pull Request checklist

- [x] **Tests**: Adds some tests around the 

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
@@ -34,8 +34,10 @@ import androidx.core.view.isVisible
 import androidx.core.view.updateLayoutParams
 import androidx.fragment.app.Fragment
 import androidx.fragment.app.activityViewModels
+import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.Observer
 import androidx.lifecycle.lifecycleScope
+import androidx.lifecycle.repeatOnLifecycle
 import androidx.navigation.fragment.findNavController
 import androidx.navigation.fragment.navArgs
 import androidx.recyclerview.widget.LinearLayoutManager
@@ -271,9 +273,18 @@ class HomeFragment : Fragment() {
             if (requireContext().settings().showPocketRecommendationsFeature) {
                 val categories = components.core.pocketStoriesService.getStories()
                     .groupBy { story -> story.category }
-                    .map { (category, stories) -> PocketRecommendedStoriesCategory(category, stories) }
+                    .map { (category, stories) ->
+                        PocketRecommendedStoriesCategory(
+                            category,
+                            stories
+                        )
+                    }
 
-                homeFragmentStore.dispatch(HomeFragmentAction.PocketStoriesCategoriesChange(categories))
+                homeFragmentStore.dispatch(
+                    HomeFragmentAction.PocketStoriesCategoriesChange(
+                        categories
+                    )
+                )
             } else {
                 homeFragmentStore.dispatch(HomeFragmentAction.PocketStoriesChange(emptyList()))
             }
@@ -392,14 +403,21 @@ class HomeFragment : Fragment() {
 
         // DO NOT MOVE ANYTHING BELOW THIS addMarker CALL!
         requireComponents.core.engine.profiler?.addMarker(
-            MarkersFragmentLifecycleCallbacks.MARKER_NAME, profilerStartTime, "HomeFragment.onCreateView",
+            MarkersFragmentLifecycleCallbacks.MARKER_NAME,
+            profilerStartTime,
+            "HomeFragment.onCreateView",
         )
 
         if (shouldEnableWallpaper()) {
-            val wallpaperManger = requireComponents.wallpaperManager
-            wallpaperManger.updateWallpaper(binding.homeLayout, wallpaperManger.currentWallpaper)
+            lifecycleScope.launch {
+                repeatOnLifecycle(Lifecycle.State.STARTED) {
+                    val wallpaperManager = requireComponents.wallpaperManager
+                    wallpaperManager.currentWallpaper.collect { wallpaper ->
+                        wallpaperManager.applyToView(binding.homeLayout, wallpaper)
+                    }
+                }
+            }
         }
-
         return binding.root
     }
 
@@ -761,10 +779,7 @@ class HomeFragment : Fragment() {
         if (shouldEnableWallpaper()) {
             binding.wordmark.setOnClickListener {
                 val manager = requireComponents.wallpaperManager
-                manager.updateWallpaper(
-                    wallpaperContainer = binding.homeLayout,
-                    newWallpaper = manager.switchToNextWallpaper()
-                )
+                manager.switchToNextWallpaper()
             }
         }
     }
--- a/app/src/main/java/org/mozilla/fenix/settings/HomeSettingsFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/HomeSettingsFragment.kt
@@ -5,6 +5,7 @@
 package org.mozilla.fenix.settings
 
 import android.os.Bundle
+import androidx.navigation.findNavController
 import androidx.preference.Preference
 import androidx.preference.PreferenceCategory
 import androidx.preference.PreferenceFragmentCompat
@@ -60,6 +61,16 @@ class HomeSettingsFragment : PreferenceFragmentCompat() {
             isVisible = FeatureFlags.historyMetadataUIFeature
             isChecked = context.settings().historyMetadataUIFeature
             onPreferenceChangeListener = CustomizeHomeMetricsUpdater()
+        }
+
+        requirePreference<Preference>(R.string.pref_key_wallpapers).apply {
+            setOnPreferenceClickListener {
+                view?.findNavController()?.navigate(
+                    HomeSettingsFragmentDirections.actionHomeSettingsFragmentToWallpaperSettingsFragment()
+                )
+                true
+            }
+            isVisible = FeatureFlags.showWallpapers
         }
 
         val openingScreenRadioHomepage =
--- a/app/src/main/res/layout/fragment_wallpaper_settings.xml
+++ b/app/src/main/res/layout/fragment_wallpaper_settings.xml
@@ -1,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+    - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    android:orientation="vertical"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent">
+
+    <androidx.compose.ui.platform.ComposeView
+        android:id="@+id/compose_view"
+        android:layout_width="match_parent"
+        android:layout_height="match_parent" />
+
+</LinearLayout>
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -461,6 +461,12 @@
     <!-- Title for the customize home screen section with Pocket. -->
     <string name="customize_toggle_pocket">Pocket</string>
 
+    <!-- Wallpapers -->
+    <!-- Title for preference to open wallpapers -->
+    <string name="customize_wallpapers">Wallpapers</string>
+    <!-- Content description for various wallpapers -->
+    <string name="content_description_wallpaper_name">Wallpaper Item: %1$s</string>
+
     <!-- Add-on Installation from AMO-->
     <!-- Error displayed when user attempts to install an add-on from AMO (addons.mozilla.org) that is not supported -->
     <string name="addon_not_supported_error">Add-on is not supported</string>
--- a/app/src/main/java/org/mozilla/fenix/components/StoreProvider.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/StoreProvider.kt
@@ -38,7 +38,7 @@ class StoreProviderFactory<T : Store<*, *>>(
 ) : ViewModelProvider.Factory {
 
     @Suppress("UNCHECKED_CAST")
-    override fun <VM : ViewModel?> create(modelClass: Class<VM>): VM {
+    override fun <VM : ViewModel> create(modelClass: Class<VM>): VM {
         return StoreProvider(createStore()) as VM
     }
 }
--- a/app/src/main/java/org/mozilla/fenix/components/metrics/BreadcrumbsRecorder.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/metrics/BreadcrumbsRecorder.kt
@@ -5,9 +5,9 @@
 package org.mozilla.fenix.components.metrics
 
 import android.os.Bundle
-import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.DefaultLifecycleObserver
 import androidx.lifecycle.LifecycleObserver
-import androidx.lifecycle.OnLifecycleEvent
+import androidx.lifecycle.LifecycleOwner
 import androidx.navigation.NavController
 import androidx.navigation.NavDestination
 import mozilla.components.concept.base.crash.Breadcrumb
@@ -23,15 +23,13 @@ class BreadcrumbsRecorder(
     private val crashReporter: CrashReporter,
     private val navController: NavController,
     private val getBreadcrumbMessage: (NavDestination) -> String
-) : NavController.OnDestinationChangedListener, LifecycleObserver {
+) : NavController.OnDestinationChangedListener, DefaultLifecycleObserver {
 
-    @OnLifecycleEvent(Lifecycle.Event.ON_CREATE)
-    fun onCreate() {
+    override fun onCreate(owner: LifecycleOwner) {
         navController.addOnDestinationChangedListener(this)
     }
 
-    @OnLifecycleEvent(Lifecycle.Event.ON_DESTROY)
-    fun onDestroy() {
+    override fun onDestroy(owner: LifecycleOwner) {
         navController.removeOnDestinationChangedListener(this)
     }
 
--- a/app/src/main/java/org/mozilla/fenix/customtabs/PoweredByNotification.kt
+++ b/app/src/main/java/org/mozilla/fenix/customtabs/PoweredByNotification.kt
@@ -15,9 +15,8 @@ import androidx.core.app.NotificationCompat.BADGE_ICON_NONE
 import androidx.core.app.NotificationManagerCompat
 import androidx.core.content.ContextCompat
 import androidx.core.content.getSystemService
-import androidx.lifecycle.Lifecycle
-import androidx.lifecycle.LifecycleObserver
-import androidx.lifecycle.OnLifecycleEvent
+import androidx.lifecycle.DefaultLifecycleObserver
+import androidx.lifecycle.LifecycleOwner
 import mozilla.components.browser.state.selector.findCustomTab
 import mozilla.components.browser.state.state.ExternalAppType
 import mozilla.components.browser.state.store.BrowserStore
@@ -32,18 +31,16 @@ class PoweredByNotification(
     private val applicationContext: Context,
     private val store: BrowserStore,
     private val customTabId: String
-) : LifecycleObserver {
+) : DefaultLifecycleObserver {
 
-    @OnLifecycleEvent(Lifecycle.Event.ON_RESUME)
-    fun onResume() {
+    override fun onResume(owner: LifecycleOwner) {
         if (store.state.findCustomTab(customTabId)?.config?.externalAppType === ExternalAppType.TRUSTED_WEB_ACTIVITY) {
             NotificationManagerCompat.from(applicationContext)
                 .notify(applicationContext, NOTIFICATION_TAG, buildNotification())
         }
     }
 
-    @OnLifecycleEvent(Lifecycle.Event.ON_PAUSE)
-    fun onPause() {
+    override fun onPause(owner: LifecycleOwner) {
         NotificationManagerCompat.from(applicationContext)
             .cancel(applicationContext, NOTIFICATION_TAG)
     }
--- a/app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkDeselectNavigationListener.kt
+++ b/app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkDeselectNavigationListener.kt
@@ -5,9 +5,8 @@
 package org.mozilla.fenix.library.bookmarks
 
 import android.os.Bundle
-import androidx.lifecycle.Lifecycle
-import androidx.lifecycle.LifecycleObserver
-import androidx.lifecycle.OnLifecycleEvent
+import androidx.lifecycle.DefaultLifecycleObserver
+import androidx.lifecycle.LifecycleOwner
 import androidx.navigation.NavController
 import androidx.navigation.NavDestination
 import org.mozilla.fenix.R
@@ -16,15 +15,13 @@ class BookmarkDeselectNavigationListener(
     private val navController: NavController,
     private val viewModel: BookmarksSharedViewModel,
     private val bookmarkInteractor: BookmarkViewInteractor
-) : NavController.OnDestinationChangedListener, LifecycleObserver {
+) : NavController.OnDestinationChangedListener, DefaultLifecycleObserver {
 
-    @OnLifecycleEvent(Lifecycle.Event.ON_RESUME)
-    fun onResume() {
+    override fun onResume(owner: LifecycleOwner) {
         navController.addOnDestinationChangedListener(this)
     }
 
-    @OnLifecycleEvent(Lifecycle.Event.ON_PAUSE)
-    fun onDestroy() {
+    override fun onDestroy(owner: LifecycleOwner) {
         navController.removeOnDestinationChangedListener(this)
     }
 
--- a/app/src/main/java/org/mozilla/fenix/settings/OnSharedPreferenceChangeListener.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/OnSharedPreferenceChangeListener.kt
@@ -5,24 +5,19 @@
 package org.mozilla.fenix.settings
 
 import android.content.SharedPreferences
-import androidx.lifecycle.Lifecycle.Event.ON_CREATE
-import androidx.lifecycle.Lifecycle.Event.ON_DESTROY
-import androidx.lifecycle.LifecycleObserver
+import androidx.lifecycle.DefaultLifecycleObserver
 import androidx.lifecycle.LifecycleOwner
-import androidx.lifecycle.OnLifecycleEvent
 
 class OnSharedPreferenceChangeListener(
     private val sharedPreferences: SharedPreferences,
     private val listener: (SharedPreferences, String) -> Unit
-) : SharedPreferences.OnSharedPreferenceChangeListener, LifecycleObserver {
+) : SharedPreferences.OnSharedPreferenceChangeListener, DefaultLifecycleObserver {
 
-    @OnLifecycleEvent(ON_CREATE)
-    fun onCreate() {
+    override fun onCreate(owner: LifecycleOwner) {
         sharedPreferences.registerOnSharedPreferenceChangeListener(this)
     }
 
-    @OnLifecycleEvent(ON_DESTROY)
-    fun onDestroy() {
+    override fun onDestroy(owner: LifecycleOwner) {
         sharedPreferences.unregisterOnSharedPreferenceChangeListener(this)
     }
 
--- a/app/src/main/java/org/mozilla/fenix/settings/about/SecretDebugMenuTrigger.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/about/SecretDebugMenuTrigger.kt
@@ -6,9 +6,8 @@ package org.mozilla.fenix.settings.about
 
 import android.view.View
 import android.widget.Toast
-import androidx.lifecycle.Lifecycle
-import androidx.lifecycle.LifecycleObserver
-import androidx.lifecycle.OnLifecycleEvent
+import androidx.lifecycle.DefaultLifecycleObserver
+import androidx.lifecycle.LifecycleOwner
 import org.mozilla.fenix.R
 import org.mozilla.fenix.utils.Settings
 
@@ -18,7 +17,7 @@ import org.mozilla.fenix.utils.Settings
 class SecretDebugMenuTrigger(
     logoView: View,
     private val settings: Settings
-) : View.OnClickListener, LifecycleObserver {
+) : View.OnClickListener, DefaultLifecycleObserver {
 
     private var secretDebugMenuClicks = 0
     private var lastDebugMenuToast: Toast? = null
@@ -32,8 +31,7 @@ class SecretDebugMenuTrigger(
     /**
      * Reset the [secretDebugMenuClicks] counter.
      */
-    @OnLifecycleEvent(Lifecycle.Event.ON_RESUME)
-    fun clearClickCounter() {
+    override fun onResume(owner: LifecycleOwner) {
         secretDebugMenuClicks = 0
     }
 
--- a/app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
@@ -1,0 +1,162 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.settings.wallpaper
+
+import androidx.compose.foundation.BorderStroke
+import androidx.compose.foundation.ExperimentalFoundationApi
+import androidx.compose.foundation.Image
+import androidx.compose.foundation.background
+import androidx.compose.foundation.border
+import androidx.compose.foundation.clickable
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.aspectRatio
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.lazy.GridCells
+import androidx.compose.foundation.lazy.LazyVerticalGrid
+import androidx.compose.foundation.lazy.items
+import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.material.Surface
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.collectAsState
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.clip
+import androidx.compose.ui.layout.ContentScale
+import androidx.compose.ui.res.painterResource
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import org.mozilla.fenix.R
+import org.mozilla.fenix.theme.FirefoxTheme
+import org.mozilla.fenix.wallpapers.Wallpaper
+import org.mozilla.fenix.wallpapers.WallpaperManager
+
+/**
+ * A screen for various settings related to wallpapers.
+ *
+ * @param wallpaperManager Manager for retrieving and updating current wallpaper.
+ */
+@Composable
+fun WallpaperSettings(
+    wallpaperManager: WallpaperManager
+) {
+    val selectedWallpaper = wallpaperManager.currentWallpaper.collectAsState()
+    Surface(color = FirefoxTheme.colors.layer2) {
+        WallpaperThumbnails(
+            wallpapers = wallpaperManager.wallpapers.toList(),
+            selectedWallpaper = selectedWallpaper.value,
+            onSelectionChanged = { wallpaperManager.updateWallpaperSelection(it) }
+        )
+    }
+}
+
+/**
+ * A grid of selectable wallpaper thumbnails.
+ *
+ * @param wallpapers Wallpapers to add to grid
+ * @param selectedWallpaper The currently selected wallpaper
+ * @param numColumns The number of columns that will occupy the grid.
+ * @param onSelectionChanged Action to take when a new wallpaper is selected.
+ */
+@OptIn(ExperimentalFoundationApi::class)
+@Composable
+private fun WallpaperThumbnails(
+    wallpapers: List<Wallpaper>,
+    selectedWallpaper: Wallpaper,
+    numColumns: Int = 3,
+    onSelectionChanged: (Wallpaper) -> Unit,
+) {
+    LazyVerticalGrid(
+        cells = GridCells.Fixed(numColumns),
+        modifier = Modifier.padding(vertical = 30.dp, horizontal = 20.dp)
+    ) {
+        items(wallpapers) { wallpaper ->
+            WallpaperThumbnailItem(
+                wallpaper = wallpaper,
+                isSelected = selectedWallpaper == wallpaper,
+                onSelectionChanged = onSelectionChanged
+            )
+        }
+    }
+}
+
+/**
+ * A single wallpaper thumbnail.
+ *
+ * @param wallpaper The wallpaper to display.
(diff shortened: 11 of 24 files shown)


MatthewTighe:
In #22851, @pocmo says:

>     * I'd consider adding such state either to the screen or app store (based on whether it's needed locally only or globally) and using the known mechanisms to update/observe state. Every `*Manager` creates another source of truth/state and will make it harder to reason about the app's state.

I'd love to learn more about how this state pattern can best be applied to address the multiple sources of truth problem. What I've seen so far makes it seem like it the stores are still an additional source of truth - the `HomeFragmentStore` for example is initialized based on persisted state but doesn't seem to continue to observe state from various backing providers like the `HistoryMetadataStorage`, for example. This is demonstrated in `RecentVisitsController`, where methods will dispatch actions to update the state and then asynchronously update the backing data.

jonalmeida:
> It looks like there was already some discussion about this in #22851, actually.

I hadn't seen Sebastian's comments before - but +1 there. 🙂

app/src/main/res/layout/fragment_wallpaper_settings.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    android:orientation="vertical"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent">
+
+    <androidx.compose.ui.platform.ComposeView
+        android:id="@+id/compose_view"
```
Amejia481, line 8: nit: I think the background of the `compose_view` should be the same as the wallpaper screen :)

Dark mode
image

Light mode
image

app/src/main/java/org/mozilla/fenix/settings/HomeSettingsFragment.kt
```diff
@@ -62,6 +63,15 @@ class HomeSettingsFragment : PreferenceFragmentCompat() {
...
         }
 
+        requirePreference<Preference>(R.string.pref_key_wallpapers).apply {
+            setOnPreferenceClickListener {
+                view?.findNavController()?.navigate(
+                    HomeSettingsFragmentDirections.actionHomeSettingsFragmentToWallpaperSettingsFragment()
+                )
+                true
+            }
+        }
```
Amejia481, line 73: nit: The setting should be hidden by default and only visible when `FeatureFlags.showWallpapers` is active :) 

https://github.com/mozilla-mobile/fenix/blob/dc576a2ad7bacdd28fa856c06a322916d8dc482f/app/src/main/java/org/mozilla/fenix/settings/SettingsFragment.kt#L458

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -271,9 +273,18 @@ class HomeFragment : Fragment() {
             if (requireContext().settings().showPocketRecommendationsFeature) {
                 val categories = components.core.pocketStoriesService.getStories()
                     .groupBy { story -> story.category }
-                    .map { (category, stories) -> PocketRecommendedStoriesCategory(category, stories) }
+                    .map { (category, stories) ->
+                        PocketRecommendedStoriesCategory(
+                            category,
+                            stories
+                        )
+                    }
```
Amejia481, line 281: nit: For follow up PRs, lets try to format changes related to the PR. It make the diff difficult to read :)

Amejia481 (commented):
Added couple of comments, I thinking after they are addressed, the PR will be ready ✅ 

Additionally, I noticed we are modifying the wallpapers, is there any specific reason? I just asking because there is a opened PR for optimizing the test images https://github.com/mozilla-mobile/fenix/pull/22937

app/src/main/res/layout/fragment_wallpaper_settings.xml
MatthewTighe: Good catch! The light mode one matches the design, which has the wallpapers on a surface rather than directly on the background. I believe this is so the default wallpaper can still be seen, since it's the same color. I'll update the PR such that dark mode reflects that behavior, though

app/src/main/java/org/mozilla/fenix/settings/HomeSettingsFragment.kt
MatthewTighe: Good catch, thanks!

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
MatthewTighe: Ah my bad. I'd seen that I had formatted that file at some point and then forgot to revert the change. I'll leave this for now and be more careful in the future. Thanks for the call out

MatthewTighe:
> Additionally, I noticed we are modifying the wallpapers, is there any specific reason? I just asking because there is a opened PR for optimizing the test images #22937

Just bad git on my part. I had imported the wallpapers when I prototyped this branch, and then when I tried to rebase onto your work I ended up needing to delete and re-import them.

app/src/main/java/org/mozilla/fenix/settings/WallpaperSettingsFragment.kt
```diff
@@ -53,54 +53,111 @@ class WallpaperSettingsFragment : Fragment() {
...
+            )
+        }
+    }
+}
+
+@Composable
+fun WallpaperThumbnailItem(
+    wallpaper: Wallpaper,
+    onSelectionChanged: (Wallpaper) -> Unit,
+    isCurrentlySelected: Boolean
```
gabrielluong: Maybe to be less verbose, I would just rename this to `isSelected`.

gabrielluong:
One workflow consideration is to amend your commits as you're working on the feature, and breaking things up into very isolated commits (for instance, the reformatting of HomeFragment) or squashing the in-between commits so that it is easier to look at the relevant changes (trying to look at everything not related to the lifecycle upgrade)

app/src/main/res/layout/fragment_wallpaper_settings.xml
Amejia481: 👍🏽

app/src/main/java/org/mozilla/fenix/settings/WallpaperSettingsFragment.kt
MatthewTighe: Yeah good call. Updated

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
Amejia481: No problem :)

app/src/main/res/layout/fragment_wallpaper_settings.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
```
gabrielluong, line 1: License please. 📑

Amejia481: 📝

MatthewTighe:
> One workflow consideration is to amend your commits as you're working on the feature, and breaking things up into very isolated commits (for instance, the reformatting of HomeFragment) or squashing the in-between commits so that it is easier to look at the relevant changes (trying to look at everything not related to the lifecycle upgrade)

That's a good idea. I've definitely been thinking more about this, and am still trying to figure out the best way to scope my squashes. I think your examples here are helpful, thanks

app/src/main/java/org/mozilla/fenix/settings/WallpaperSettingsFragment.kt
```diff
@@ -0,0 +1,175 @@
...
+    }
+}
+
+@Composable
+fun WallpaperImageThumbnail(
+    wallpaper: Wallpaper,
+) {
+    val localizedContentDescription = LocalContext.current.resources.getString(
+        R.string.content_description_wallpaper_name, wallpaper.name
+    )
```
gabrielluong, line 158: This look strangely verbose to me since I recall an easier way to get the string resource. Perhaps try `stringResource`

app/src/main/res/layout/fragment_wallpaper_settings.xml
MatthewTighe: Added. Is there a hook I could add to catch these?

app/src/main/java/org/mozilla/fenix/settings/WallpaperSettingsFragment.kt
MatthewTighe: That's the one, thanks. Shortened the unnecessarily verbose name while I was at it

app/src/main/java/org/mozilla/fenix/settings/WallpaperSettingsFragment.kt
```diff
@@ -0,0 +1,175 @@
...
+import org.mozilla.fenix.databinding.FragmentWallpaperSettingsBinding
+import org.mozilla.fenix.ext.requireComponents
+import org.mozilla.fenix.theme.FirefoxTheme
+import org.mozilla.fenix.wallpapers.Wallpaper
+import org.mozilla.fenix.wallpapers.WallpaperManager
+
+class WallpaperSettingsFragment : Fragment() {
+    private val binding by lazy {
+        FragmentWallpaperSettingsBinding.inflate(layoutInflater)
+    }
```
gabrielluong, line 49: I am kinda curious about this pattern for the initializing the `binding`. I know my primary source about how to work with view binding comes with https://developer.android.com/topic/libraries/view-binding#fragments, which I am aware is the convention we currently follow.

app/src/main/res/values/strings.xml
```diff
@@ -454,6 +454,10 @@
     <string name="customize_toggle_recent_searches" moz:removedIn="96" tools:ignore="UnusedResources">Recent searches</string>
     <!-- Title for the customize home screen section with Pocket. -->
     <string name="customize_toggle_pocket">Pocket</string>
+    <!-- Title for preference to open wallpapers -->
+    <string name="preferences_wallpapers">Wallpapers</string>
+    <!-- Content description for various wallpapers -->
+    <string name="content_description_wallpaper_name">Wallpaper Item: %1$s</string>
```
gabrielluong, line 460: There's no formal structure to this, but observing just how our strings are organized - I think we want to flip this around to be `wallpaper_name_content_description`. I am not totally aware of the full context (mostly just reviewing based on style/consistency right now) but I am also wondering if `name` is the right description for the string name given the context.

Also, another aside about how strings are (sometimes) organized, it would be ideal to group related strings under a `<!-- Wallpaper -->` grouping if it makes sense, but otherwise, just prefixing the string based on the feature can be very helpful to finding related strings.

app/src/main/java/org/mozilla/fenix/settings/WallpaperSettingsFragment.kt
MatthewTighe: I thought this was a classic case of Google docs not writing good Kotlin. I actually would've gotten this question wrong on a quiz:
image

Looks like this isn't a feasible pattern, and it makes more sense to follow convention anyway. Updated

app/src/main/res/values/strings.xml
```diff
@@ -454,6 +454,10 @@
     <string name="customize_toggle_recent_searches" moz:removedIn="96" tools:ignore="UnusedResources">Recent searches</string>
     <!-- Title for the customize home screen section with Pocket. -->
     <string name="customize_toggle_pocket">Pocket</string>
+    <!-- Title for preference to open wallpapers -->
+    <string name="preferences_wallpapers">Wallpapers</string>
```
gabrielluong, line 458:
```suggestion
    <string name="customize_wallpapers">Wallpapers</string>
```

I think we uses `preferences_` for the Settings (one screen before) preferences. Just more naming convention gotchas.

app/src/main/res/values/strings.xml
gabrielluong: On another note, I think `content_description_wallpaper_name ` is actually used within the Wallpaper fragment. So, it totally makes sense to create `<!-- Wallpaper -->` grouping instead of keeping it under `Customize home`.

app/src/main/java/org/mozilla/fenix/settings/WallpaperSettingsFragment.kt
```diff
@@ -0,0 +1,175 @@
...
+        }
+        return binding.root
+    }
+}
+
+@Composable
+fun WallpaperSettings(
+    wallpaperManager: WallpaperManager
+) {
+    val currentlySelectedWallpaper = wallpaperManager.currentWallpaper.collectAsState()
```
gabrielluong, line 76: Let's go with `selectedWallpaper` to avoid unnecessary adverbs. 😅

app/src/main/res/values/strings.xml
MatthewTighe: Looks like a pretty mixed bag of `content_description` being a suffix or prefix. I'll defer to your suggestion here and update though. Does `wallpaper_settings_item_content_description` make more sense?

gabrielluong: Ya sounds fine, I just wasn't sure if `name` quite described the true intent given the `Wallpaper Item: %1$s` context that I was working with.

app/src/main/res/values/strings.xml
MatthewTighe: Nit away! I'm actually really grateful for such a detailed review. It'll really help me write code that's consistent with what already exists. Also a good indicator that I should maybe be taking a bit more care in general 😅

app/src/main/res/values/strings.xml
MatthewTighe: Cool. Updating with these suggestions

app/src/main/java/org/mozilla/fenix/settings/WallpaperSettingsFragment.kt
```diff
@@ -0,0 +1,175 @@
...
+    Column {
+        Surface {
```
gabrielluong, line 78: I don't think we need Column and Surface here. A Surface is a Box which is a Layout. A Column is also just a Layout.

MatthewTighe: Artifact from when I'd included more layouts below the Surface, following designs for upcoming feature. I'll remove this for now

app/src/main/java/org/mozilla/fenix/settings/WallpaperSettingsFragment.kt
```diff
@@ -0,0 +1,175 @@
...
+            )
+        }
+    }
+}
+
+@OptIn(ExperimentalFoundationApi::class)
+@Composable
+fun WallpaperThumbnails(
+    wallpapers: List<Wallpaper>,
+    onSelectionChanged: (Wallpaper) -> Unit,
```
gabrielluong, line 92: I have a slight preference for keeping these types of lambdas as the bottom parameter items. This is mostly coming from my React background. 

WallpaperThumbnails(
    wallpapers: List<Wallpaper>,
    onSelectionChanged: (Wallpaper) -> Unit,
 

 
I have a slight preference for keeping these types of lambdas as the bottom parameter items. This is mostly coming from my React background. 

Ultimately, I think about the flow when I am reading the code, it would easier to see the simple parameters (Int, String, etc) first and whereas providing a lambda function disrupt the readability of things.

Here's a visual example:
```
WallpaperThumbnails(
    wallpapers = wallpapers,
    selectedWallpaper = selectedWallpaper,
    numColumns = 4,
    onSelectionChanged = { wallpaper ->
        someFunction(wallpaper)
        anotherCallback(wallpaper)
   }
)
```

app/src/main/java/org/mozilla/fenix/settings/WallpaperSettingsFragment.kt
```diff
@@ -0,0 +1,175 @@
...
+                onSelectionChanged = { wallpaperManager.updateWallpaperSelection(it) },
+                selectedWallpaper = currentlySelectedWallpaper.value
+            )
+        }
+    }
+}
+
+@OptIn(ExperimentalFoundationApi::class)
+@Composable
+fun WallpaperThumbnails(
```
gabrielluong, line 90: I am somewhat shocked that ktlint didn't complain about how there's no kDoc for public functions, but I would definitely add some kDocs for all these public composable functions. Better get use to it early because I will ask for docs.

MatthewTighe: 🤦 you'd think that would be at the top of my brain, as a previous library author. I'll add some docs throughout

app/src/main/java/org/mozilla/fenix/settings/WallpaperSettingsFragment.kt
MatthewTighe: This was actually very intentional. Here's my reasoning:

Jetpack Compose has defined the convention that `@Composable` slots should generally be the final parameter to other composable functions. The reason being that we get the nice visual cues for parent-child relationships in our view tree:
```
@Composable
fun Tree() {
    Surface {
        Button {
            Text()
...
```

If we have trailing lambdas that aren't composables but are visually within a compose scope, that could be also be pretty hostile to the reader:
```
@Composable
fun Tree() {
    Surface {
        WallpaperThumbnails {
            // now we're in the onSelectionChanged lambda
```

I agree that having the lambda types in the middle is pretty gross, but my method allows us to rely on our tools to avoid situations like this completely. I wonder if maybe there's a convention we could come up with that's a compromise, such as defining the lambda outside constructor call as to not interrupt the parameter list:

@Composable
fun Tree() {
    Surface {
        val onWallpaperSelectionChanged: (Wallpaper) -> Unit = {}
        WallpaperThumbnails(
            onSelectionChanged = onWallpaperSelectionChanged
...
```

File: app/src/main/java/org/mozilla/fenix/settings/WallpaperSettingsFragment.kt (hunk 13 above)
Comment (MatthewTighe) [reply app/src/main/java/org/mozilla/fenix/settings/WallpaperSettingsFragment.kt:90]: Unrelated, but I'm also wondering what the best location for composables is. Should they be in a separate file? If so, do they belong under the `settings` package or the new `wallpaper` package?

File: app/src/main/java/org/mozilla/fenix/settings/WallpaperSettingsFragment.kt (hunk 12 above)
Comment (gabrielluong) [reply app/src/main/java/org/mozilla/fenix/settings/WallpaperSettingsFragment.kt:92]: I did revise my comment to add the exception about the @Composable content after submitting it initially. In this case, I was not really thinking about the middle case you presented which is that we use trailing lambdas - which is a pattern I am hoping we don't see often. This is probably my general expectation in terms of how we pass methods into our composable functions https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabViewHolder.kt#41-44.

Comment (MatthewTighe) [reply app/src/main/java/org/mozilla/fenix/settings/WallpaperSettingsFragment.kt:92]: Yeah I would never suggest we intentionally use trailing lambdas the way I demonstrated in that middle case. Just pointing out that by putting functional parameters in the middle of the list we could also never accidentally use them in the middle case way. Compose tends to favor using named parameters anyway, so it should be an easily avoidable hypothetical. 

I'll update these signatures

File: app/src/main/java/org/mozilla/fenix/settings/WallpaperSettingsFragment.kt (hunk 13 above)
Comment (gabrielluong) [reply app/src/main/java/org/mozilla/fenix/settings/WallpaperSettingsFragment.kt:90]: You can always refactor this as a follow up. Looking at our existing organization, I would imagine this would under `settings/wallpaper`. This is based on the idea that secondary(Wallpaper) fragments are packaged under their parent(Setting) fragment. While this isn't very formal by any means, I would try to observe the current patterns/structure and try to adapt. There's nothing wrong with going full maverick and making changes, but I would do that in a separate issue and ensure people are align with the changes we want to make for our Fenix architecture. I think the more important bits is that we want others to be able to adopt the styles.

File: app/src/main/java/org/mozilla/fenix/settings/WallpaperSettingsFragment.kt (hunk 12 above)
Comment (gabrielluong) [reply app/src/main/java/org/mozilla/fenix/settings/WallpaperSettingsFragment.kt:92]: This is a great discussion. Thanks for updating the signatures. Ideally, we try to converge towards a common Fenix/team style for writing composable functions similar message to https://github.com/mozilla-mobile/fenix/pull/22953#discussion_r774230937 - that way it's easier for others to follow and replicate as a standard convention. For the most part, I have been looking at the compose sample apps and even digging into the compose material UI implementation to just try and absorb what is the best practise, but with anything that's a work in progress short of someone writing a linter that enforces an opinionated and accepted style.

File: app/src/main/java/org/mozilla/fenix/settings/WallpaperSettingsFragment.kt (hunk 13 above)
Comment (pocmo) [reply app/src/main/java/org/mozilla/fenix/settings/WallpaperSettingsFragment.kt]: > I am somewhat shocked that ktlint didn't complain about how there's no kDoc for public functions

It's disabled in detekt here:
https://github.com/mozilla-mobile/fenix/blob/main/config/detekt.yml#L43

But enabled in A-C:
https://github.com/mozilla-mobile/android-components/blob/main/config/detekt.yml#L37

🤷‍♂️

**Comment by pocmo** [comment]
> I'd love to learn more about how this state pattern can best be applied to address the multiple sources of truth problem. What I've seen so far makes it seem like it the stores are still an additional source of truth - the `HomeFragmentStore` for example is initialized based on persisted state but doesn't seem to continue to observe state from various backing providers like the `HistoryMetadataStorage`, for example. This is demonstrated in `RecentVisitsController`, where methods will dispatch actions to update the state and then asynchronously update the backing data.

Yeah, those parts of the code are not ideal. But over time a great pattern for this has emerged on the A-C side: frontend/UI code only observers state from a `Store` and dispatches `Action`s on the `Store` to make changes. `"Storage"` classes are never exposed to it.

To actually interact with a storage we have used a `Middleware` on the `Store`. The `Middleware` can see an "init" `Action`, read data from the storage and dispatch an `Action` to initialize the state. The `Middleware` can see `Action`s that change the state and perform the same operation on the storage. The big advantages are: The app/frontend/ui code doesn't need to take care of any of that and just deals with the state. All storage access is nicely encapsulated and can be changed in one place. There's only one storage read needed (in this example) and after that we only do writes if needed.

Some examples in A-C are:
* `ContainerMiddleware`: https://github.com/mozilla-mobile/android-components/blob/main/components/feature/containers/src/main/java/mozilla/components/feature/containers/ContainerMiddleware.kt
* Recently closed tabs: https://github.com/mozilla-mobile/android-components/blob/main/components/feature/recentlyclosed/src/main/java/mozilla/components/feature/recentlyclosed/RecentlyClosedMiddleware.kt

Of course this pattern is only applicable if the data can be reasonably held in memory. You wouldn't want to have all of the user's history in memory at all times (but maybe very recent history for fast display, or "pages" of it).

File: app/src/main/res/values/strings.xml (hunk 8 above)
Comment (pocmo) [reply app/src/main/res/values/strings.xml:462]: Also make sure to explain what `%1$s` gets replaced with in the comment above. The comments will show up in the translation tool (Pontoon) and if this is missing then the L10N team will block the string import until its resolved. :)

**Comment by gabrielluong** [comment]
> > I'd love to learn more about how this state pattern can best be applied to address the multiple sources of truth problem. What I've seen so far makes it seem like it the stores are still an additional source of truth - the `HomeFragmentStore` for example is initialized based on persisted state but doesn't seem to continue to observe state from various backing providers like the `HistoryMetadataStorage`, for example. This is demonstrated in `RecentVisitsController`, where methods will dispatch actions to update the state and then asynchronously update the backing data.
> 
> Yeah, those parts of the code are not ideal. But over time a great pattern for this has emerged on the A-C side: frontend/UI code only observers state from a `Store` and dispatches `Action`s on the `Store` to make changes. `"Storage"` classes are never exposed to it.
> 
> To actually interact with a storage we have used a `Middleware` on the `Store`. The `Middleware` can see an "init" `Action`, read data from the storage and dispatch an `Action` to initialize the state. The `Middleware` can see `Action`s that change the state and perform the same operation on the storage. The big advantages are: The app/frontend/ui code doesn't need to take care of any of that and just deals with the state. All storage access is nicely encapsulated and can be changed in one place. There's only one storage read needed (in this example) and after that we only do writes if needed.
> 
> Some examples in A-C are:
> 
> * `ContainerMiddleware`: https://github.com/mozilla-mobile/android-components/blob/main/components/feature/containers/src/main/java/mozilla/components/feature/containers/ContainerMiddleware.kt
> * Recently closed tabs: https://github.com/mozilla-mobile/android-components/blob/main/components/feature/recentlyclosed/src/main/java/mozilla/components/feature/recentlyclosed/RecentlyClosedMiddleware.kt
> 
> Of course this pattern is only applicable if the data can be reasonably held in memory. You wouldn't want to have all of the user's history in memory at all times (but maybe very recent history for fast display, or "pages" of it).

@pocmo This is really interesting. I would love to convert our top sites to make use of the Middleware and stop exposing the TopSitesStorage. I am curious if this is a sensible candidate.

I am also wondering how our Fenix Stores (such as `HomeFragmentStore`) manages to be able to persist state given that it's also recreated when a fragment runs `onCreateView` and avoids the multiple sources of truth issue with the *Managers. I know we also cache some of the state values in various classes (See `DefaultTopSitesStorage` and [`cachedTopSites`](https://searchfox.org/mozilla-mobile/source/android-components/components/feature/top-sites/src/main/java/mozilla/components/feature/top/sites/DefaultTopSitesStorage.kt#39)) that we just pass into its initialization. Looking at `DefaultTopSitesStorage`, we might also have a multiple sources of truth problem(?), but I am wondering how to identify them and how our current Store solution (vs caching values in a class like *Managers) mitigates it. Although I do understand the Redux/Flux single stores, I am just not sure how to contextualized my understanding of Store coming from JS/Web background with Android which has multiple threads, and lifecycle recreation. Basically, I am wondering how we manage to keep our Stores as a singleton(?) and persistent.

**Comment by pocmo** [comment]
> @pocmo This is really interesting. I would love to convert our top sites to make use of the Middleware and stop exposing the TopSitesStorage. I am curious if this is a sensible candidate.

Yeah, I absolutely think that this is a great candidate. We need to display the home screen all the time - and accessing the disk makes it slower. If I remember correctly that was also one of the things Christian and I talked about.

> I am also wondering how our Fenix Stores (such as HomeFragmentStore) manages to be able to persist state given that it's also recreated when a fragment runs onCreateView and avoids the multiple sources of truth issue with the *Managers. [..] Basically, I am wondering how we manage to keep our Stores as a singleton(?) and persistent.

I see this as three levels:
* BrowserStore: Global, Singleton. Keep the state of the "browser". It's separate because it's independent from the app and shared across our Components that use it.
* "AppStore" (which Jonathan recently added here, and which we have been using in Focus too): Global, Singleton. This is where the app stores global state that is independent from the current screen (or cached globally). For example state that describes what theme the app is using. This state would be independent of a specific screen and needed everywhere. This is the place where I'd consider caching top sites too. They may only be needed for one screen, but we want to keep them in memory at all times.
* "ScreenStore": Scoped to a screen. Keeps the state describing a specific screen. For example if I select items from a list in a screen then this would be kept here.

Middlewares are a great way to add (selective) persistence to this. For BrowserStore we provide persistence in A-C. Currently that code is completely separate. If I'd write this today then I'd write a Middleware for this. The code was written when we were still using `SessionManager` and was simply refactored to work with BrowserStore.

For an "AppStore": As described above middlewares could load initial state and update the state and keep the persistence layer in sync. That's essentially how I would do it for BrowserStore too - if I'd would want to refactor it. The difference being that for BrowserStore we (currently) write/read the state as a whole. While for AppStores we may want separate middlewares by "feature". In a similar way I would have liked to move tab collections to BrowserState, and add a middleware for CRUD operations.

For "ScreenStore": My gut feeling is that a lot of the state is transient and doesn't need to be persisted (-> A screen always initializes in the same state, maybe using state from the other stores). Although the middleware model would work here too. However for this type of store there's a need for handling configuration changes where Android kills an activity or even process. We wouldn't want to lose state in that situation (e.g. a list selection, text in a form). I recently wrote a helper in A-C for creating a locally scoped Store in compose code, that has hooks for save/restore on configuration changes:
https://github.com/mozilla-mobile/android-components/blob/main/components/lib/state/src/main/java/mozilla/components/lib/state/ext/ComposeExtensions.kt#L84

I haven't come to a conclusion yet whether a screen should read from any of those three stores or just its "ScreenStore". (Partial/Mapped) State from the other stores could be "synced" into the "ScreenStore" and I thought we could have helpers that make it easy to write code that does this mapping and syncing. But for simple cases it may be good enough to just read from the store that is the source of the state..

**Comment by MatthewTighe** [comment]
> To actually interact with a storage we have used a Middleware on the Store. The Middleware can see an "init" Action, read data from the storage and dispatch an Action to initialize the state. The Middleware can see Actions that change the state and perform the same operation on the storage. The big advantages are: The app/frontend/ui code doesn't need to take care of any of that and just deals with the state. All storage access is nicely encapsulated and can be changed in one place. There's only one storage read needed (in this example) and after that we only do writes if needed.

> For an "AppStore": As described above middlewares could load initial state and update the state and keep the persistence layer in sync. That's essentially how I would do it for BrowserStore too - if I'd would want to refactor it. The difference being that for BrowserStore we (currently) write/read the state as a whole. While for AppStores we may want separate middlewares by "feature". In a similar way I would have liked to move tab collections to BrowserState, and add a middleware for CRUD operations.

This conversation is great, and coincidentally useful for the work I'm doing to add a blocklist for items on the home screen. The design I've been working on there has been to use a middleware to manage the blocklist and filter state changes to the screen. I was having difficulty coming up with the best way to manipulate the initial state of the page there, however. If the best way to access is initial state is to intercept an "init" Action, I'm wondering if perhaps having an `initialState` constructor parameter for Stores is antithetical to that and we should instead force stores to require an "init" action.

That said, using an `initialState` parameter feels like a more natural flow to me. Would it make sense to update the definition of `Store` such that it propagated that initial state to its middlewares? The current implementation feels like it leads to the following scenarios:

1. Wait for an action to be dispatched to the store, so a middleware can retrieve the previous state of the store. For the use case above, this could lead to situations where the blocklist wasn't filtered on initial page load.
2. Introduce an init action when a middleware and dispatch it on middleware construction to force the store to reveal its initial state. My concern here would be ordering and timing issues.
3. Dispatch an init action from within the store on construction.
4. Construct the initial state and inject it separately into the store and middlewares that require it.

File: app/src/main/java/org/mozilla/fenix/settings/HomeSettingsFragment.kt [hunk 14]
```diff
@@ -62,6 +63,15 @@ class HomeSettingsFragment : PreferenceFragmentCompat() {
...
         }
 
+        requirePreference<Preference>(R.string.pref_key_wallpapers).apply {
+            setOnPreferenceClickListener {
+                view?.findNavController()?.navigate(
+                    HomeSettingsFragmentDirections.actionHomeSettingsFragmentToWallpaperSettingsFragment()
+                )
+                true
+            }
+        }.isVisible = FeatureFlags.showWallpapers
```
gabrielluong, line 73: Couldn't the `isVisible` be inside the `apply { }`?

MatthewTighe: Oh yeah, it definitely should be. Updated

app/src/main/java/org/mozilla/fenix/settings/WallpaperSettingsFragment.kt
gabrielluong: Filed https://github.com/mozilla-mobile/fenix/issues/23046 to investigate the difference between Fenix and AC, and hopefully align to something more common.

gabrielluong:
> > To actually interact with a storage we have used a Middleware on the Store. The Middleware can see an "init" Action, read data from the storage and dispatch an Action to initialize the state. The Middleware can see Actions that change the state and perform the same operation on the storage. The big advantages are: The app/frontend/ui code doesn't need to take care of any of that and just deals with the state. All storage access is nicely encapsulated and can be changed in one place. There's only one storage read needed (in this example) and after that we only do writes if needed.
> 
> > For an "AppStore": As described above middlewares could load initial state and update the state and keep the persistence layer in sync. That's essentially how I would do it for BrowserStore too - if I'd would want to refactor it. The difference being that for BrowserStore we (currently) write/read the state as a whole. While for AppStores we may want separate middlewares by "feature". In a similar way I would have liked to move tab collections to BrowserState, and add a middleware for CRUD operations.
> 
> This conversation is great, and coincidentally useful for the work I'm doing to add a blocklist for items on the home screen. The design I've been working on there has been to use a middleware to manage the blocklist and filter state changes to the screen. I was having difficulty coming up with the best way to manipulate the initial state of the page. If the best way to access is initial state is to intercept an "init" Action, I'm wondering if perhaps having an `initialState` constructor parameter for Stores is antithetical and we should instead force stores to require an "init" action.
> 
> That said, using an `initialState` parameter feels like a more natural flow to me. Would it make sense to update the definition of `Store` such that it propagated that initial state to its middlewares? The current implementation feels like it leads to the following scenarios:
> 
> 1. Wait for an action to be dispatched to the store, so a middleware can retrieve the previous state of the store. For the use case above, this could lead to situations where the blocklist wasn't filtered on initial page load.
> 2. Introduce an init action when a middleware and dispatch it on middleware construction to force the store to reveal its initial state. My concern here would be ordering and timing issues.
> 3. Dispatch an init action from within the store on construction.
> 4. Construct the initial state and inject it separately into the store and middlewares that require it.



> > To actually interact with a storage we have used a Middleware on the Store. The Middleware can see an "init" Action, read data from the storage and dispatch an Action to initialize the state. The Middleware can see Actions that change the state and perform the same operation on the storage. The big advantages are: The app/frontend/ui code doesn't need to take care of any of that and just deals with the state. All storage access is nicely encapsulated and can be changed in one place. There's only one storage read needed (in this example) and after that we only do writes if needed.
> 
> > For an "AppStore": As described above middlewares could load initial state and update the state and keep the persistence layer in sync. That's essentially how I would do it for BrowserStore too - if I'd would want to refactor it. The difference being that for BrowserStore we (currently) write/read the state as a whole. While for AppStores we may want separate middlewares by "feature". In a similar way I would have liked to move tab collections to BrowserState, and add a middleware for CRUD operations.
> 
> This conversation is great, and coincidentally useful for the work I'm doing to add a blocklist for items on the home screen. The design I've been working on there has been to use a middleware to manage the blocklist and filter state changes to the screen. I was having difficulty coming up with the best way to manipulate the initial state of the page. If the best way to access is initial state is to intercept an "init" Action, I'm wondering if perhaps having an `initialState` constructor parameter for Stores is antithetical and we should instead force stores to require an "init" action.
> 
> That said, using an `initialState` parameter feels like a more natural flow to me. Would it make sense to update the definition of `Store` such that it propagated that initial state to its middlewares? The current implementation feels like it leads to the following scenarios:
> 
> 1. Wait for an action to be dispatched to the store, so a middleware can retrieve the previous state of the store. For the use case above, this could lead to situations where the blocklist wasn't filtered on initial page load.
> 2. Introduce an init action when a middleware and dispatch it on middleware construction to force the store to reveal its initial state. My concern here would be ordering and timing issues.
> 3. Dispatch an init action from within the store on construction.
> 4. Construct the initial state and inject it separately into the store and middlewares that require it.

We actually have an `InitAction` https://searchfox.org/mozilla-mobile/source/android-components/components/feature/containers/src/main/java/mozilla/components/feature/containers/ContainerMiddleware.kt#43. I think it depends on what your intended use case during the init. In most cases, you probably want to perform some IO read from storage prior to populating Store, but you don't necessarily to perform a bunch of async storage fetch before initializing the Store.

pocmo:
Note that this `InitAction` is only on `BrowserStore` and would need to be added to other stores if they need it. The main reason for that is that we want actions to be `sealed` and so we can't have a generic init action in `Store`. Not a big deal though, and it makes sense to be more explicit.

https://github.com/mozilla-mobile/android-components/blob/main/components/browser/state/src/main/java/mozilla/components/browser/state/store/BrowserStore.kt#L47

I wouldn't add any "side channels" to make middlewares handle other things. So far all of those "side channels" could be simplified to be a regular middleware, which is just a "simple" function:
https://github.com/mozilla-mobile/android-components/blob/main/components/lib/state/src/main/java/mozilla/components/lib/state/Middleware.kt#L18

This is also in line with how middlewares work in other implementations like redux:
https://redux.js.org/understanding/history-and-design/middleware

In your scenario I am wondering where the initial state is coming from? I assume we wouldn't wait for loading some data to create an initial state. I would expect us to start with an empty state and the middleware to load the data and process it as needed, and then adding it to the state. In that scenario there wouldn't be a case where there's "unprocessed data" being displayed.

MatthewTighe:
@pocmo Here's some abbreviated code from my branch:

The initial state is being loaded from various places, but I still need to filter it before it's received by the store. Otherwise, sensitive items that were removed might be briefly displayed. The disk read is probably a trivial impact, but we could theorize a situation in which the blocklist is synced across clients. We'd want to avoid a situation in which a sensitive bookmark was shown because it was removed from one client but not from another.

```
val initialBlocklist = runBlockingIncrement { /* collect blocklist flow from storage */ }
homeFragmentStore = StoreProvider.get(this) {
    HomeFragmentStore(
        HomeFragmentState(
            collections = components.core.tabCollectionStorage.cachedTabCollections,
            expandedCollections = emptySet(),
            mode = currentMode.getCurrentMode(),
            topSites = components.core.topSitesStorage.cachedTopSites,
            tip = components.strictMode.resetAfter(StrictMode.allowThreadDiskReads()) {
                FenixTipManager(
                    listOf(
                        MasterPasswordTipProvider(
                            requireContext(),
                            ::navToSavedLogins,
                            ::dismissTip
                        )
                    )
                ).getTip()
            },
            recentBookmarks = emptyList(),
            showCollectionPlaceholder = components.settings.showCollectionsPlaceholderOnHome,
            showSetAsDefaultBrowserCard = components.settings.shouldShowSetAsDefaultBrowserCard(),
            // Provide an initial state for recent tabs to prevent re-rendering on the home screen.
            //  This will otherwise cause a visual jump as the section gets rendered from no state
            //  to some state.
            recentTabs = getRecentTabs(components),
            recentHistory = emptyList()
        ).filteredByBlocklist(initialBlocklist),
        listOf(BlocklistMiddleware(), PocketUpdatesMiddleware())
    )
}
```

@gabrielluong I accidentally edited your comment above, instead of quoting it in a reply. Really sorry about that. I re-edited to include at least the stuff I was trying to quote, but lost some of the initial context. I think you were pointing out the `InitAction` in [ContainerMiddleware](https://searchfox.org/mozilla-mobile/source/android-components/components/feature/containers/src/main/java/mozilla/components/feature/containers/ContainerMiddleware.kt#43). Here's my response:

> I think it depends on what your intended use case during the init. In most cases, you probably want to perform some IO read from storage prior to populating Store, but you don't necessarily to perform a bunch of async storage fetch before initializing the Store.

I'm used to doing exactly that - performing async fetches to use a remote as a source of truth and displaying loading states. If we initialize a store using local storage and there are updates to it before remote data is received there can be conflicts that would need to be resolved in an opinionated fashion. Are there places within our code that are examples of that?

> In your scenario I am wondering where the initial state is coming from? I assume we wouldn't wait for loading some data to create an initial state. I would expect us to start with an empty state and the middleware to load the data and process it as needed, and then adding it to the state. In that scenario there wouldn't be a case where there's "unprocessed data" being displayed.

Ultimately, it looks like this use case doesn't match the pattern @pocmo is describing above. The initial state is not empty, so we can't rely on the middleware to alter state changes as they're applied. That kinda circles back to my earlier point though - there aren't constraints on a `Store` that require the initial state to be empty since it receives its initial state during construction.

MatthewTighe:
@gabrielluong @Amejia481 

I took a shot at refactoring this to use `Store`s and `Middleware`s, to try and align it better with the our architectural goals. You can see the diff [here](https://github.com/MatthewTighe/fenix/pull/1/files). Is this something we'd be interested in landing instead of the implementation in this PR?

gabrielluong:
> @gabrielluong @Amejia481
> 
> I took a shot at refactoring this to use `Store`s and `Middleware`s, to try and align it better with the our architectural goals. You can see the diff [here](https://github.com/MatthewTighe/fenix/pull/1/files). Is this something we'd be interested in landing instead of the implementation in this PR?

I think the diff is looking okay. I added some comments to it.

MatthewTighe:
> I think the diff is looking okay. I added some comments to it.

I'll close this PR, then, and open one with the refactor

MatthewTighe:
Re-opening this, as @gabrielluong has suggested merging this and then the refactor separately.

MatthewTighe:
Closing in favor of #23145

PR #22985, opened 2021-12-24
Closes #10737: wrong search view width on language and saved logins s…

### LINKED ISSUE #10737: [Bug] Searching a language/saved password displays a misplaced "x"
## Steps to reproduce
1. Go to Settings - Language and tap on the search icon.
1. Type a few characters and observe the "x" item's position.

### Expected behavior
The "x" is placed at the end of the line.

### Actual behavior
The "x" is in the middle of the line. Even when RTL language is used.

### Device information
* Android device: Samsung Galaxy Tab A6 (Android 5.1.1)
* Fenix version: 5/18 Nightly build
* Note: The same behavior when searching a saved login.

mavduevskiy:
Fixes #10737

A fix for the [...fenix/10737](https://github.com/mozilla-mobile/fenix/issues/10737) issue.

It appears the [widget](https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/widget/SearchView.java#817) we have been using, given no max value, will use the default width for the seach-view. Which happens to be [this](https://chromium.googlesource.com/android_tools/+/HEAD/sdk/extras/android/support/v7/appcompat/res/values/dimens.xml#39) exact value, with no regard to the landscape mode or super wide screen size.

Setting max width to the max possible value forces the search-view to behave as match-parent, respecting the margins.

Also, removed from LocaleSettingsFragment this event – Event.SyncAuthClosed. 
It looks like it was copy-pasted from TurnOnSyncFragment.

Before
image

After
image


app/src/main/java/org/mozilla/fenix/settings/advanced/LocaleSettingsFragment.kt
```diff
@@ -104,12 +103,6 @@ class LocaleSettingsFragment : Fragment() {
...
     override fun onDestroyView() {
         super.onDestroyView()
-
         _binding = null
     }
-
-    override fun onDestroy() {
-        super.onDestroy()
-        requireComponents.analytics.metrics.track(Event.SyncAuthClosed)
-    }
```
gabrielluong, line 114: Any reason we are removing this?

app/src/main/java/org/mozilla/fenix/settings/advanced/LocaleSettingsFragment.kt
```diff
@@ -104,12 +103,6 @@ class LocaleSettingsFragment : Fragment() {
 
     override fun onDestroyView() {
         super.onDestroyView()
-
```
gabrielluong, line 107: nit: let's not remove this new line

app/src/main/java/org/mozilla/fenix/settings/advanced/LocaleSettingsFragment.kt
gabrielluong: Nevermind, should've read PR comment.

gabrielluong (approved):

gabrielluong:
Might also be worthwhile to adjust your commit message to mention your specific changes (instead of what's wrong) - Specify max width for the search view in saved logins and language settings.

gabrielluong:
Also, a couple of workflow notes that others will thank you in the future (if and when we need to look at old historical changess):
- Add the Linked issues (this is the right section below the reviewers). So, that you can easily go to the PR from the issue. You can also easily link the issue by saying "Fixes #10737" in the PR description.
- In the issue, add the Milestone when this is landed.
- When you do land the PR, reopen the issue (not PR) and add "qa-needed" so that QA knows to test it.

Mugurell:
> * When you do land the PR, reopen the issue (not PR) and add "qa-needed" so that QA knows to test it.

To not have the issue closed automatically and then having to reopen it instead of using `Fixes` or `Closes` in the commit title one can use `For #...`.

app/src/main/java/org/mozilla/fenix/settings/advanced/LocaleSettingsFragment.kt
mavduevskiy: Is there any policy around that? There are 33 cases of overriding the `onDestroy` method, and only 10 of those actually have a new line. It tells me two things: there is no consistency and no line version is dominant.

mavduevskiy:
@gabrielluong 
Could you elaborate on "You can also easily link the issue by saying "Fixes #xxxxx" in the PR description." part? What that automatically add the issue from the description to the linked issues side menu?

app/src/main/java/org/mozilla/fenix/settings/advanced/LocaleSettingsFragment.kt
gabrielluong: No strong policy or feelings toward it. I only pointed it out because it was a minor (maybe even unintentional?) change (vs lint forcing the change) compare to what was set out to be changed in the commit. An example of an extreme case would be if you did a bunch of refactoring within the same commit, where the refactoring could've been in its own issue/commit for an easier review and separation of concerns between commits. Objectively speaking, whether or not we had this new line would not have an impact on the PR. At best, it's just a personal nit.

That said if I had to reflect on my own personal preference, I would still chose to use new lines to separate out and group logical blocks (if/when blocks, super() from other pieces of logic, initialization from method calls, etc) to make things easier to read, BUT it's definitely okay to keep your changes as-is after is said.

gabrielluong:
> @gabrielluong Could you elaborate on "You can also easily link the issue by saying "Fixes #xxxxx" in the PR description." part? What that automatically add the issue from the description to the linked issues side menu?

So, I went to your first comment (what I call the PR description) and added "Fixes #xxxx" this is very similar to what you added with 

> A fix for the ...fenix/10737 issue.

However, that won't automatically add the linked issues that you see in the red arrow in the screenshot. 

Screen Shot 2021-12-29 at 3 22 11 PM

app/src/main/java/org/mozilla/fenix/settings/advanced/LocaleSettingsFragment.kt
mavduevskiy: Fair enough, changed back. It sounds reasonable not to do partisan-refactoring.

Issue #22995, opened 2021-12-27
[Bug]: The app crashes when returning to the add-on details menu after changing device language


hokonch:
### Steps to reproduce

1. Tap an add-on that hasn't been added yet in [Add-ons] and show the add-on details
2. Go to Settings app and add a new device language
3. Return to Firefox app

### Expected behaviour

Firefox never crashes.

### Actual behaviour

The app closes unexpectedly.

### Device name

Google Pixel 3a

### Android version

Android 12

### Firefox release type

Firefox

### Firefox version

95.2.0 & 97.0a1

### Device logs

----
ee162a6b-2e23-4342-a0df-2b7502bcf18e
`java.lang.RuntimeException: Unable to start activity ComponentInfo{org.mozilla.fenix/org.mozilla.fenix.HomeActivity}: android.view.InflateException: Binary XML file line #27 in org.mozilla.fenix:layout/activity_home: Binary XML file line #27 in org.mozilla.fenix:layout/activity_home: Error inflating class fragment`
 * Sentry: https://sentry.prod.mozaws.net/operations/fenix/?query=c5118febe8ea404cb967d1e985f0bca0
 * Socorro: https://crash-stats.mozilla.org/report/index/bp-a689f42b-3c0c-4b0c-ac55-5a2820211227
----
~~~
java.lang.RuntimeException: Unable to start activity ComponentInfo{org.mozilla.fenix/org.mozilla.fenix.HomeActivity}: android.view.InflateException: Binary XML file line #27 in org.mozilla.fenix:layout/activity_home: Binary XML file line #27 in org.mozilla.fenix:layout/activity_home: Error inflating class fragment
	at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:3635)
	at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3792)
	at android.app.ActivityThread.handleRelaunchActivityInner(ActivityThread.java:5738)
	at android.app.ActivityThread.handleRelaunchActivity(ActivityThread.java:5630)
	at android.app.servertransaction.ActivityRelaunchItem.execute(ActivityRelaunchItem.java:71)
	at android.app.servertransaction.ActivityTransactionItem.execute(ActivityTransactionItem.java:45)
	at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:135)
	at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:95)
[...]

PR #23001, opened 2021-12-28
For #22995: Address crash when switching languages.

Amejia481:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/addons/AddonDetailsFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/addons/AddonDetailsFragment.kt
@@ -34,13 +34,15 @@ class AddonDetailsFragment : Fragment(R.layout.fragment_add_on_details), AddonDe
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
+        val binding = FragmentAddOnDetailsBinding.bind(view)
+        AddonDetailsBindingDelegate(binding, interactor = this).bind(args.addon)
+    }
 
+    override fun onResume() {
+        super.onResume()
         context?.let {
             showToolbar(title = args.addon.translateName(it))
         }
-
-        val binding = FragmentAddOnDetailsBinding.bind(view)
-        AddonDetailsBindingDelegate(binding, interactor = this).bind(args.addon)
     }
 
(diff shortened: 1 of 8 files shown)


Amejia481:
@Mugurell  to keep consistency I think it's better to just follow the same pattern as we are doing on the rest of the app.

Mugurell:
> @Mugurell to keep consistency I think it's better to just follow the same pattern as we are doing on the rest of the app.

`onResume` seems misused for this kind of things even according to [it's documentation](https://developer.android.com/guide/components/activities/activity-lifecycle#onresume).
If we are updating the toolbar title after it is shown this will cause a bit of a jank.
Having this in `onStart` would be marginally better - less jank
and doing this in `onCreateView` (while making sure that the parent activity has finished it's `onCreate` - so there is a toolbar we can work with) and being able to set it up so that it will be displayed as expected from the first time the screen is shown would be perfect.

That being said, this solves the crash. r+.

Mugurell (approved):

Amejia481:
> > @Mugurell to keep consistency I think it's better to just follow the same pattern as we are doing on the rest of the app.
> 
> `onResume` seems misused for this kind of things even according to [it's documentation](https://developer.android.com/guide/components/activities/activity-lifecycle#onresume). If we are updating the toolbar title after it is shown this will cause a bit of a jank. Having this in `onStart` would be marginally better - less jank and doing this in `onCreateView` (while making sure that the parent activity has finished it's `onCreate` - so there is a toolbar we can work with) and being able to set it up so that it will be displayed as expected from the first time the screen is shown and not updating it again when the app gains focus - because the title should not have changed anyways (thinking about the scenario of switching between apps) would be perfect.
> 
> That being said, this solves the crash. r+.

Thanks for taking a look. Testing locally, I couldn't see any significant impact. For the moment, I will address the issue until we take a more general solution for all the instances.

gabrielluong:
@Mugurell @Amejia481 Let's file a follow up to refactor how and where we call `showToolbar`.

Amejia481:
> @Mugurell @Amejia481 Let's file a follow up to refactor how and where we call `showToolbar`. I would do it myself, but it was on unclear if we prefer to do this in `onStart` vs `onCreateView`.

@gabrielluong I filed https://github.com/mozilla-mobile/fenix/issues/23059 for exploring both options and use which one best fits our needs.

Issue #22722, opened 2021-12-06
In-app crash reporter: React to "crashed" flag instead of Intent


pocmo:
(I was pretty sure I filed this before. But I can't find it. So filing it again)

When a content process crashes then this doesn't crash the whole app. Instead affected tabs are considered crashed and can be restored. For that Fenix implemented an in-app UI that says "this tab crashed" and offers the user to "restore" or "close" the tab. To make this possible Fenix is listening to an `Intent` from the crash reporter ([code](https://github.com/mozilla-mobile/fenix/blob/6a3adb5aeba976a7146eb96a2ef3794fae909304/app/src/main/java/org/mozilla/fenix/home/intent/CrashReporterIntentProcessor.kt#L16)).

This implementation has multiple problems. Primarily it is only able to show the crash reporter if the app is in the foreground listening to this intent. There can be multiple reasons that a tab crashes in the background and the intent does not get processed. If a crash reporter was never shown for a tab then this tab has a crashed `GeckoSession` attached to it. This crashed tab will not be able to render content (and may show content of another tab) and the user cannot interact with it.

Instead Fenix should show the crash reporter for the current tab whenever `tab.engineState.crashed` is true - and ignore the `Intent`. This will make sure that the crash reporter will always overlay a crashed tab, forcing the user to make a decision (restore or close).

Furthermore this will make it possible to build a better UI. Currently, with the `Intent`, Fenix only sees the `Intent` and knows that _a_ tab crashed. But not which one. By reading this flag Fenix can show the crash reporter for the specific, crashed tab - as well as list all crashed tab and/or allow to restore subsets.

kbrosnan:
Added this to the backlog.

Mugurell:
Looked a bit at this and I think more approaches are possible.
Looking at home Chrome handle tab crashes maybe we should have a similar approach, one closer to how we show the error pages allowing the user to freely interact with the app even switch between tabs:

| Chrome handling crashed tabs | Fenix handling error pages | Fenix handling crashed tabs |
| --- | --- | --- |
[...]

PR #23010, opened 2021-12-29
For #22722 - Reacting to the crashed flag

Mugurell:
Whenever the ".crashed" property of the currently displayed
TabSessionState -> EngineState is true we will show an in-app crash reporter
with the usual close tab / restore tab options and also the option to report
all current non-fatal crashes to Mozilla if the setting for sending the crash
reports is enabled in app settings.
This closely mimics the previous crash reporter UI but there might be some
subtle differences stemming from migrating to using a ComposeView.

Whenever the ".crashed" property of the currently displayed
TabSessionState -> EngineState is false we will set the in-app crash reporter
to have a View.GONE visibility effectively removing it from the layout.

The functionality for receiving the non-fatal crashes from the AC CrashReporter
through an Intent is still kept and these crashes will be persisted in memory
until the user closes / restores a tab and so also makes a decision about
sending or not these crashes.

Currently more tabs can crash following just one since more share the same
process and as such there is no way to differentiate between them or link a
certain Crash to a certain tab.
They will all be acted upon at once from any tab the user chooses to close or
restore.

https://user-images.githubusercontent.com/11428869/148430576-9de78567-fb0a-422e-9763-f009e899385e.mp4

### Pull Request checklist

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/crashes/CrashedContent.kt
```diff
@@ -0,0 +1,160 @@
...
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.text.TextStyle
+import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+import org.mozilla.fenix.R
+import org.mozilla.fenix.theme.FirefoxTheme
+
+@Composable
+fun CrashedContent(
```
gabrielluong, line 41: Can we use rename this to `CrashContent`? I am looking at our codebase and just thinking we should probably just avoid using past tense (Crash**ed**), and maybe even adverbs (Recent**ly** and prefer Recent) for naming. My reasoning for this is that we don't use the term Crashed Report (vs Crash Report) so it might be best to be consistent.

app/src/main/java/org/mozilla/fenix/components/appstate/AppState.kt
```diff
@@ -4,14 +4,17 @@
...
  * Value type that represents the state of the tabs tray.
  *
  * @property inactiveTabsExpanded A flag to know if the Inactive Tabs section of the Tabs Tray
  * should be expanded when the tray is opened.
+ * @property nonFatalCrashes List of non-fatal crashes that allow the app to continue being used.
  */
 data class AppState(
-    val inactiveTabsExpanded: Boolean = false
+    val inactiveTabsExpanded: Boolean = false,
+    val nonFatalCrashes: List<Crash> = emptyList()
```
Mugurell, line 19: We receive the content crash(es) through an intent and keep them in memory as pending until the user decides to report them.
First looked into keeping them in BrowserState and completely avoid depending on the crash intent but for that I needed to map a crash to a particular tabId, something not possible atm.

app/src/main/java/org/mozilla/fenix/crashes/CrashContentInteractor.kt
```diff
@@ -0,0 +1,88 @@
...
+import kotlinx.coroutines.launch
+import org.mozilla.fenix.components.AppStore
+import org.mozilla.fenix.components.Components
+import org.mozilla.fenix.components.appstate.AppAction
+import org.mozilla.fenix.utils.Settings
+
+/**
+ * Handles all user interactions with a [CrashContentView].
+ */
+class CrashContentInteractor(
```
Mugurell, line 22: Went with just an Interactor for composables and not delegating to a Controller following the model from https://github.com/mozilla-mobile/focus-android/pull/5985#issuecomment-986848567

gabrielluong, line 22: Can we have a commit within the PR that simply renames `CrashReporterController` to `CrashContentInteractor`? This is mostly for preserving the git history of the file so it doesn't appear as a totally new file. So, only the relevant changes in the file needs to be reviewed instead of the whole file.

app/src/main/java/org/mozilla/fenix/crashes/CrashContentView.kt
```diff
@@ -0,0 +1,201 @@
...
+                onCloseTabClicked = { interactor.handleCloseAndRemove(it) },
+                onRestoreTabClicked = { interactor.handleCloseAndRestore(it) }
+            )
+        }
+    }
+}
+
+@Composable
+@Suppress("LongMethod", "MagicNumber")
+private fun CrashContent(
```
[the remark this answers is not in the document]
gabrielluong: I think we ultimately want to do Compose conversion in isolation instead attaching it with a fix for another related issue.

gabrielluong (dismissed):
Requesting changes for the following reasons:
- Let's move the Compose conversion out of this PR so that we can look at that in isolation. I think there's a lot going on here so we don't want to think about reviewing multiple changes (Compose, AppStore) in one PR.
- Let's try to preserve our git history while refactoring our classes

app/src/main/java/org/mozilla/fenix/crashes/CrashContentInteractor.kt
```diff
@@ -0,0 +1,88 @@
...
+     * and the report crashes setting is enabled.
+     * Also clears the current list of non-fatal crashes irrespective of whether they are reported or not.
+     *
+     * @param reportCrashes A second condition beside crash reporting being enabled in app settings
+     * based on which the current crashes will be reported or immediately disposed off.
+     * @return [Job] allowing to check status / cancel the reporting operation or null if reporting is disabled.
+     */
+    @VisibleForTesting
+    @OptIn(DelicateCoroutinesApi::class) // GlobalScope usage
+    internal fun maybeSubmitPendingNonFatalCrashes(reportCrashes: Boolean): Job? {
```
gabrielluong, line 69: Looking through our codebase, it seems it's better to use the `IfNecessary` suffix for these types of functions over `maybe`. So, perhaps `submitPendingNonFatalCrashesIfNecessary`?

app/src/main/java/org/mozilla/fenix/crashes/CrashContentInteractor.kt
```diff
@@ -0,0 +1,88 @@
...
+        var job: Job? = null
+        if (reportCrashes && settings.isCrashReportingEnabled) {
+            job = GlobalScope.launch(Dispatchers.IO) {
+                val crashes = appStore.state.nonFatalCrashes
+                crashes.forEach {
+                    components.analytics.crashReporter.submitReport(it)
+                    appStore.dispatch(AppAction.RemoveNonFatalCrash(it))
+
+                    Log.i("Crash Reporter", "Report \"${it.uuid}\" submitted.")
```
gabrielluong, line 78: do we need these logs? Can we add an equivalent `Logger('CrashContentInteracator')`?

Mugurell: Removed them.

app/src/main/java/org/mozilla/fenix/crashes/CrashContentInteractor.kt
Mugurell: Will try to accommodate this.

app/src/main/java/org/mozilla/fenix/crashes/CrashContentView.kt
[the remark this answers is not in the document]
Mugurell: My idea was that we'll have to refactor this to compose anyway so we might just do it when we're at it.
Dropped the compose approach, sticked with xml.

app/src/main/java/org/mozilla/fenix/crashes/CrashContentInteractor.kt
Mugurell: NA if using XML.

Mugurell: Looks like we need a separate commit for each renamed file.
Tried to manually merge the diffs but doesn't seem manageable for bigger diffs with multiple usages of the renamings.

app/src/main/java/org/mozilla/fenix/crashes/CrashReporterController.kt
```diff
@@ -4,81 +4,78 @@
...
 class CrashReporterController(
-    private val crash: Crash,
-    private val sessionId: String?,
-    private val navController: NavController,
-    private val components: Components,
-    private val settings: Settings
+    @get:VisibleForTesting internal val sessionId: String,
+    @get:VisibleForTesting internal val components: Components,
+    @get:VisibleForTesting internal val settings: Settings,
+    @get:VisibleForTesting internal val appStore: AppStore
```
gabrielluong, line 22: Wondering how come we had to make all of these `@get:VisibleForTesting internal`

Mugurell: `VisibleForTesting(otherwise = PRIVATE) internal` is the general approach used to be able to access otherwise private functions and properties in tests.
These properties are used in `CrashContentIntegrationTest` to test that the controller is constructed with the proper arguments.

app/src/main/res/layout/fragment_browser.xml
```diff
@@ -60,6 +60,12 @@
                 android:elevation="24dp"
                 android:visibility="gone" />
 
+            <org.mozilla.fenix.crashes.CrashContentView
+                android:id="@+id/crash_reporter_view"
+                android:layout_width="match_parent"
+                android:layout_height="match_parent"
+                android:visibility="gone" />
```
gabrielluong, line 67: I think I am most uncertain about this particular change hence pinging @csadilek for a review about this direction (navigating to a fragment vs displaying the view within fragment_browser) and the AppStore usage. It does make sense to me in a way to keep things cached in the AppStore and a overlaying the crash view, but just looking for more eyes.

Mugurell: To me as a user it would make more sense to have the crash status and reports available even in between app restarts, until I make a decision - whether to send the report or reopen a crash tab. I understand that for now we don't need to persist this though so the `crashed` status is availabe in BrowserState but not persisted so the crashes are also kept in memory. Used Appstore because atm it's not possible to link a crash to a tab to add this info also to BrowserState. Maybe in the future.

Regarding the View that would overlay the browser also in the future I think it would be great to have this as part of the EngineView's API, lifting the need for each GV integrator to implement something similar, maybe it could come as part of the compose refactoring.

app/src/main/java/org/mozilla/fenix/components/appstate/AppState.kt
rocketsroger: Since we're only storing native non-fatal crashes here, should this be `<Crash.UncaughtExceptionCrash>` so it's clear?

app/src/main/java/org/mozilla/fenix/home/intent/CrashReporterIntentProcessor.kt
```diff
@@ -5,27 +5,30 @@
...
         } else {
             false
         }
     }
 
-    private fun openToCrashReporter(intent: Intent, navController: NavController) {
-        val directions = NavGraphDirections.actionGlobalCrashReporter(intent)
-        navController.navigate(directions)
+    @VisibleForTesting
+    fun updateCrashDetails(crash: Crash) {
```
rocketsroger, line 31: Looks like nothing is using this function.  Should this be removed?

app/src/main/java/org/mozilla/fenix/crashes/CrashReporterController.kt
```diff
@@ -4,81 +4,78 @@
...
+                val crashes = appStore.state.nonFatalCrashes
+                crashes.forEach {
+                    components.analytics.crashReporter.submitReport(it)
+                    appStore.dispatch(AppAction.RemoveNonFatalCrash(it))
+                }
             }
-            true
         } else {
-            false
+            appStore.dispatch(AppAction.RemoveAllNonFatalCrashes)
```
rocketsroger, line 76: Does this means if the user untick the box we will be removing all stored non-fatal crashes?  Does this break the possible usecase where if there are multiple non-fatal crashes but the user decides to not submit one but submit other ones?

app/src/main/java/org/mozilla/fenix/crashes/CrashContentIntegration.kt
```diff
@@ -0,0 +1,74 @@
...
+    private val appStore: AppStore,
+    private val toolbar: BrowserToolbar,
+    private val isToolbarPlacedAtTop: Boolean,
+    private val crashReporterView: CrashContentView,
+    private val components: Components,
+    private val settings: Settings,
+    private val sessionId: String?
+) : AbstractBinding<BrowserState>(browserStore) {
+    override suspend fun onState(flow: Flow<BrowserState>) {
+        flow.mapNotNull { state -> state.findTabOrCustomTabOrSelectedTab(sessionId) }
```
rocketsroger, line 48: If `sessionId` is null here, should we continue?

app/src/main/java/org/mozilla/fenix/crashes/CrashReporterController.kt
```diff
@@ -4,81 +4,78 @@
...
     @OptIn(DelicateCoroutinesApi::class) // GlobalScope usage
-    private fun submitReportIfNecessary(sendCrash: Boolean): Job? {
+    internal fun submitPendingNonFatalCrashesIfNecessary(reportCrashes: Boolean): Job? {
         var job: Job? = null
-        val didSubmitReport = if (sendCrash && settings.isCrashReportingEnabled) {
+        if (reportCrashes && settings.isCrashReportingEnabled) {
             job = GlobalScope.launch(Dispatchers.IO) {
-                components.analytics.crashReporter.submitReport(crash)
+                val crashes = appStore.state.nonFatalCrashes
+                crashes.forEach {
```
rocketsroger, line 70: (Related to the comment below)  I see, we're not just submitting the crash related to the `sessionId` but reporting all crashes.  Is this something that we want?

rocketsroger (commented):
Looks good.  Some questions related to the design.  Thanks

app/src/main/java/org/mozilla/fenix/crashes/CrashReporterController.kt
Mugurell: I would very much like to have independent in-app crash reporters - for each tab but it's not possible atm. There is no mapping between a tab and a crash report and a content crash would actually crash multiple tabs - all running in that process.

app/src/main/java/org/mozilla/fenix/crashes/CrashContentIntegration.kt
Mugurell: Yes, I want to continue here based on the following scenario:
- `sessionId` passed to this components is `null` <=> there is no set `customTabSessionId` in BaseBrowserFragment
so the user is in normal browsing => we have to default on checking the `crashed` status of the selected tab.
- `sessionId` passed to this components is not `null` => have to check the state for a particular custom tab - a custom tab cannot be "selected" so just checking for selected would not show this crash reporter for a custom tab.
- when a tab crashes AC will set the `crashed` status for a particular tab - custom or normal.
- user can only see the in-app crash reporter after it selects that tab

So if the user is on a custom tab we'll listen for it's updates otherwise we'll listen for the updates of the currently selected tab after the user selects that tab - and accesses it - and so we'll update (show/hide) the in-app crash reporter only for the custom tab or the currently selected tab.

app/src/main/java/org/mozilla/fenix/components/appstate/AppState.kt
Mugurell: Thanks! This seems like a nice improvement.

app/src/main/java/org/mozilla/fenix/home/intent/CrashReporterIntentProcessor.kt
```diff
@@ -7,25 +7,25 @@ package org.mozilla.fenix.home.intent
...
+ * Process the [Intent] from [CrashReporter] through which the app is informed about
+ * recoverable from native crashes.
  */
-class CrashReporterIntentProcessor : HomeIntentProcessor {
+class CrashReporterIntentProcessor(private val store: AppStore) : HomeIntentProcessor {
 
     override fun process(intent: Intent, navController: NavController, out: Intent): Boolean {
         return if (Crash.isCrashIntent(intent)) {
-            openToCrashReporter(intent, navController)
+            (Crash.fromIntent(intent) as? UncaughtExceptionCrash)?.let {
```
rocketsroger, line 23: Actually should this be `NativeCodeCrash`?  When I made my suggested I was looking at your cod using `UncaughtExceptionCrash`.  But shouldn't this be a non-fatal native crash?

rocketsroger: nit: do we need to log if a fatal/non-native crash arrives here? (should not happen)

Mugurell: Indeed, this was initially ment for `NativeCodeCrash`es and the documentation of the class says so.
Changed accordingly.
Added an error log.

rocketsroger (approved):
Looks good to me! 🚢

app/src/test/java/org/mozilla/fenix/crashes/CrashContentIntegrationTest.kt
```diff
@@ -0,0 +1,106 @@
...
+                tabs = listOf(
+                    createTab("url", id = sessionId)
+                ),
+                selectedTabId = sessionId
+            )
+        )
+    }
+
+    @Test
+    fun `GIVEN a tab WHEN it's content crashes THEN expand the toolbar and show the in-content crash reporter`() {
```
csadilek, line 49: nit: its content process crashes....

app/src/test/java/org/mozilla/fenix/crashes/CrashReporterControllerTest.kt
```diff
@@ -4,83 +4,114 @@
...
+    private val crash: NativeCodeCrash = mockk(relaxed = true)
+    private var appStore = AppStore(
+        AppState(
+            nonFatalCrashes = listOf(crash)
+        )
+    )
+    private var controller = CrashReporterController(sessionId, components, settings, appStore)
+
+    @Test
+    fun `GIVEN reportCrashes true WHEN handleCloseAndRestore is called THEN try submitting non-fatal crashes and recover tabs`() {
```
csadilek, line 35: nitnit :) In general it's better to keep the method names out of the test names. Otherwise all test names will be stale if you anyone ever refactors those methods.

e.g. GIVEN crashes should be reported WHEN user restores tab THEN non-fatal crash report is submitted and tab recovered.

csadilek: (similar for other tests below.)

app/src/main/java/org/mozilla/fenix/home/intent/CrashReporterIntentProcessor.kt
```diff
@@ -5,27 +5,32 @@
...
+class CrashReporterIntentProcessor(private val store: AppStore) : HomeIntentProcessor {
 
     override fun process(intent: Intent, navController: NavController, out: Intent): Boolean {
         return if (Crash.isCrashIntent(intent)) {
-            openToCrashReporter(intent, navController)
+            val crash = Crash.fromIntent(intent)
+            if ((crash as? NativeCodeCrash)?.isFatal == false) {
+                store.dispatch(AppAction.AddNonFatalCrash(crash))
+            } else {
+                // Note that an unexpected crash was sent but avoid leaking potential sensitive information.
```
csadilek, line 28: We had some work here last week remembering/figuring out the details. Would you mind adding some more comments/info here why both fatal and non native crashed don't end up here. Thanks!

csadilek (commented):
Looks good! Just a few nits and a clarification :).

app/src/main/res/layout/fragment_browser.xml
csadilek: Yes, this looks good to me though. In the future, we may be able to report per tab, and having this API as part of the EngineView also makes sense then to me.

csadilek (approved):

app/src/main/java/org/mozilla/fenix/home/intent/CrashReporterIntentProcessor.kt
Mugurell: Added
```
A fatal crash means the app's main process is affected.
An UncaughtExceptionCrash refers to a [Throwable] that would otherwise crash the app
but is intercepted to allow us to gather more info and crash more gracefully.

In both cases the app is left in a bad state so the main process is killed
but not before gathering more info about the crashes to form and persist a crash report and
not before "CrashHandlerService" is started in a separate process to be able to
show a dialog allowing users to send the crash report and maybe restart the app.
```

Mugurell: @csadilek Tried to offer a bit of context about those cases, how they get the app in such a bad state that the main process has to be killed with us then relying on `CrashHandlerService`.
Would adding
> So in these cases where the app cannot be used anymore we don't expect an
Intent for an in-app crash reporter.

help?

Mugurell:
Failing `closeTabCrashedReporterTest` UI test shows a scenario I overlooked: When the crashing tab is the only one opened and the user chooses to close it then we need to navigate to home.
Made the changes to support this and also update the explanation for the log in `CrashReporterIntentProcessor` to explicitly say "We expect other types of crashes to be handled by "CrashHandlerService"".

app/src/main/java/org/mozilla/fenix/home/intent/CrashReporterIntentProcessor.kt
csadilek: Yes, that's good! Thanks. I just wanted us to point out how fatal crashes are handled and why they don't end up here. So goo now!

PR #23068, opened 2022-01-05
For #22851: Add logo animation for wallpapers.

Amejia481:
### Pull Request checklist

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -431,6 +431,16 @@ class Settings(private val appContext: Context) : PreferencesHolder {
         featureFlag = FeatureFlags.inactiveTabs
     )
 
+    /**
+     * Indicates if the Firefox logo on the home screen should be animated,
+     * to show users that they can change the wallpaper by tapping on the Firefox logo.
+     */
+    var shouldAnimeFirefoxLogo by featureFlagPreference(
```
mcarare, line 438: nit: `shouldAnimateFirefoxLogo`?

Amejia481: Thanks for catching this :) 👍🏽

Amejia481: Updated!

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -816,6 +817,26 @@ class HomeFragment : Fragment() {
         // triggered to cause an automatic update on warm start (no tab selection occurs). So we
         // update it manually here.
         requireComponents.useCases.sessionUseCases.updateLastAccess()
+
+        if (shouldEnableWallpaper()) {
+            _binding?.sessionControlRecyclerView?.viewTreeObserver?.addOnGlobalLayoutListener(
```
mcarare, line 822: I think we should still use `binding`, otherwise, we might miss some regressions if this ends up being called at the wrong time in the lifecycle.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -816,6 +817,26 @@ class HomeFragment : Fragment() {
...
+        }
+    }
+
+    // To try to find a good time to show the logo animation, we are waiting until all
+    // the sub-recyclerviews (recentBookmarks, collections, recentTabs,recentVisits
+    // and pocketStories) on the home screen have been layout.
+    private val homeLayoutListenerForLogoAnimation = object : ViewTreeObserver.OnGlobalLayoutListener {
+        override fun onGlobalLayout() {
+            _binding?.let { safeBindings ->
+                requireComponents.wallpaperManager.animateLogoIfNeeded(safeBindings.wordmark)
```
mcarare, line 834: Same as above.

mcarare (approved):
LGTM with the mention about `binding`. I leave it to UX to evaluate the animation. 🚶‍♀️

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
Amejia481: `binding` using `!!` that causes null pointer exceptions in some cases, using  `_binding` we have control over when the `binding` is not available.

https://github.com/mozilla-mobile/fenix/blob/76fb147ed87c32f37b6b92db1a0d0b3541308d86/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt#L139

MatthewTighe: Are there specific cases you hit NPEs with here? Otherwise, I think we can just use `binding`, as that's what's used throughout the app. It should be safe between `onCreateView` and `onDestroyView` according to the [docs](https://developer.android.com/topic/libraries/view-binding#fragments)

MatthewTighe (approved):

Issue #23069, opened 2022-01-05
Add blocklist for items on homescreen. 


MatthewTighe:
https://mozilla-hub.atlassian.net/browse/FNXV2-18720

Users can remove items from home screen (tabs, bookmarks, history item)

If an item is removed from one section of the home screen, it may still show up on the others

When an item is removed just the specific page be removed. Other pages from the same website may still show up

Remove option can be access in the context menu by long-pressing on the item

Context menu matches current design for Top Sites and Recent Searches

String “Remove”

Remove items are blocked from showing up on the homescreen, but not from their original location. For example, removing a tab will mean it no longer shows up on Firefox home, but the tab is still open in the tabs tray

If the user visits the same site again in the future, it’s still blocked from showing on home, but not history

Out of Scope

This ticket does not include the option to unblock a tab/site

This ticket doesn’t include groups functionality and shouldn’t change the current behaviour for Recent Searches

LaurentiuApahideanSV:
I tested the issue on Nightly 2022-02-24. When a site is removed from the Recent Bookmarks section of Recently visited section is is removed from all the Home sections (Recent Bookmarks section, Recently visited and Jump back in) if present, furthermore the site will not appear in those sections anymore. The Top sites and search groups are unaffected when a site is removed.

highroller:
There should be an option to unblock a site.  Is there one?

highroller:
How to unblock a website after I blocked it?

MatthewTighe:
@highroller We don't currently offer a way to do that through the app. Clearing the app's cache or reinstalling it should delete the blocklist, but this will erase all other custom settings as well.

highroller:
> @highroller We don't currently offer a way to do that through the app. Clearing the app's data or reinstalling it should delete the blocklist, but this will erase all other custom settings as well.

Is it possible to do it in desktop?  I really don't want to erase all custom settings. Thanks.

MatthewTighe:
Not currently, the blocklist is local to each device and isn't shared. Adding the functionality as a feature would require UX involvement and need to go through our usual prioritization process. I'd suggest filing a feature request issue if it's something you'd like to see added.

PR #23070, opened 2022-01-05
for #23069: add blocklist middleware for home

MatthewTighe:
Posting this as a draft to try and get some eyes from @pocmo if he has time before the end of the week. This implementation was already WIP when we had a lengthy conversation about middlewares and initial store states on [this PR](https://github.com/mozilla-mobile/fenix/pull/22953). 

To summarize some implementation concerns:
- The initial home screen state should be filtered by the blocklist, which is loaded async.
- The initial home screen state is supplied directly to the store, skipping over middleware.
- Here I'm blocking the initialization of the store until the blocklist is loaded. Probably realistically trivial, but still problematic especially if we were to ever consider a synced blocklist.

Would a better move here to be creating an `InitAction` for the store and supplying the initial state that way? Would that require a dedicated initialization middleware, or supplying the reducer with the components required to fill the state?


app/src/main/java/org/mozilla/fenix/home/recentbookmarks/view/RecentBookmarks.kt
```diff
@@ -4,29 +4,23 @@
...
-import androidx.compose.foundation.layout.Arrangement
-import androidx.compose.foundation.layout.Box
-import androidx.compose.foundation.layout.Column
-import androidx.compose.foundation.layout.PaddingValues
-import androidx.compose.foundation.layout.Spacer
-import androidx.compose.foundation.layout.fillMaxWidth
-import androidx.compose.foundation.layout.height
-import androidx.compose.foundation.layout.size
-import androidx.compose.foundation.layout.width
+import androidx.compose.foundation.*
```
gabrielluong, line 7: We want to use single imports. I am not quite sure if this https://github.com/mozilla-mobile/android-components#style reinforces it.

app/src/main/java/org/mozilla/fenix/home/blocklist/BlocklistMiddleware.kt
```diff
@@ -0,0 +1,185 @@
...
+            }
+        }
+        is HomeFragmentAction.RemoveRecentBookmark -> {
+            action.recentBookmark.url?.let {
+                val updatedBlocklist = blocklist + it
+                provider.addUrlToBlocklist(it)
+                state.filteredByBlocklist(updatedBlocklist).toChangeAction()
+            } ?: action
+        }
+        is HomeFragmentAction.RemoveTopSite -> {
```
gabrielluong, line 98: Reading the jira requirement, it seems like Top Sites should not be filtered here.

app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabs.kt
```diff
@@ -56,63 +40,32 @@ import org.mozilla.fenix.compose.Image
...
 
-/**
- * A list of recent tabs to jump back to.
- *
- * @param recentTabs List of [RecentTab] to display.
- * @param onRecentTabClick Invoked when the user clicks on a recent tab.
- * @param onRecentSearchGroupClicked Invoked when the user clicks on a recent search group.
- */
-@Composable
-fun RecentTabs(
```
gabrielluong, line 67: I am curious why this was moved. I would imagine we want to keep the public function at the top of the file since it's the main function we care about looking at `RecentTabs.kt`. I would otherwise try to avoid too much refactoring to reduce the amount of code that needs to be reviewed in a single commit.

app/src/main/res/values/strings.xml
```diff
@@ -58,6 +58,7 @@
     <string name="recently_saved_show_all_content_description_2">Show all saved bookmarks</string>
     <!-- Content description for the button which navigates the user to show all of their saved bookmarks. -->
     <string name="recently_saved_show_all_content_description" moz:removedIn="97" tools:ignore="UnusedResources">Show all saved bookmarks button</string>
+    <string name="recently_saved_menu_item_remove">Remove</string>
```
gabrielluong, line 61: Needs a doc

app/src/main/java/org/mozilla/fenix/theme/FirefoxTheme.kt
```diff
@@ -15,6 +15,7 @@ import androidx.compose.runtime.remember
 import androidx.compose.runtime.setValue
 import androidx.compose.runtime.staticCompositionLocalOf
 import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.platform.LocalContext
```
gabrielluong, line 18: This looks unused to me, and I would probably skip the new line removal below. My reasoning for this is that we should utilize new lines to separate variable declarations from bigger functional blocks like `ProvideFirefoxColors(colors) {}`.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -230,6 +236,11 @@ class HomeFragment : Fragment() {
             ::dispatchModeChanges
         )
 
+        val blocklistProvider = LocalHomescreenBlocklistProvider(
+            lifecycleScope,
+            requireContext().homescreenBlocklistDataStore
+        )
```
Amejia481, line 242: I think we tend to init dependencies lazily on [Components](https://github.com/mozilla-mobile/fenix/blob/main/app/src/main/java/org/mozilla/fenix/components/Components.kt) class :)

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -256,13 +267,17 @@ class HomeFragment : Fragment() {
                     //  to some state.
                     recentTabs = getRecentTabs(components),
                     recentHistory = emptyList()
-                ),
+                ).filteredByBlocklist(initialBlocklist),
```
MatthewTighe, line 270: @pocmo This is the crux of what I could use some advice on. What's the most idiomatic way to adjust this initial state using this state pattern?

app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabs.kt
MatthewTighe: Not sure how that happened either 🤔 reverted that change

app/src/main/java/org/mozilla/fenix/home/blocklist/BlocklistMiddleware.kt
MatthewTighe: Removed top sites from the implementation

app/src/main/res/values/strings.xml
MatthewTighe: Added

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
Amejia481, line 242: I think we tend to init component object lazily on the [Components](https://github.com/mozilla-mobile/fenix/blob/main/app/src/main/java/org/mozilla/fenix/components/Components.kt)  class.

app/src/main/java/org/mozilla/fenix/home/blocklist/BlocklistMiddleware.kt
```diff
@@ -0,0 +1,170 @@
...
+
+    init {
+        provider.blocklist.map { updatedList ->
+            blocklist = updatedList
+            if (!blocklistLoaded) {
+                filterLatestState()
+            }
+            blocklistLoaded = true
+        }.launchIn(coroutineScope)
+    }
```
Amejia481, line 36: I think we should try to init this lazily when need it the first time we access it. As we be blocking when creating this middleware.

MatthewTighe: - This data should be needed as soon as the object is constructed, in order to immediately filter the initial state of the home screen
- The `launchIn` function is a terminal flow operator that starts collection in a newly launched coroutine, so it shouldn't be blocking.

app/src/main/java/org/mozilla/fenix/home/recentbookmarks/view/RecentBookmarks.kt
MatthewTighe: Caught again by Android Studio defaults. Updated

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
MatthewTighe: Oh good call! I'll update this if the general design of the feature doesn't end up changing

app/src/main/java/org/mozilla/fenix/home/blocklist/BlocklistMiddleware.kt
```diff
@@ -0,0 +1,170 @@
...
+    private var store: Store<HomeFragmentState, HomeFragmentAction>? = null
+
+    init {
+        provider.blocklist.map { updatedList ->
+            blocklist = updatedList
+            if (!blocklistLoaded) {
+                filterLatestState()
+            }
+            blocklistLoaded = true
+        }.launchIn(coroutineScope)
```
gabrielluong, line 35: This might be a style thing, but I think we typically do `scope.launch { }` or some variation

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -230,6 +236,11 @@ class HomeFragment : Fragment() {
             ::dispatchModeChanges
         )
 
+        val blocklistProvider = LocalHomescreenBlocklistProvider(
+            lifecycleScope,
+            requireContext().homescreenBlocklistDataStore
+        )
+        val initialBlocklist = runBlockingIncrement { blocklistProvider.blocklist.first() }
```
gabrielluong, line 243: I think this might've been more complicated because of the introduction of the async preference datastore. I am wondering if this would just be simpler if we added a string preference to `Settings`.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
pocmo: If it is supposed to be a global Singleton. Otherwise, if it's only used locally, then I think it makes sense to only create it in that scope.

app/src/main/java/org/mozilla/fenix/home/blocklist/BlocklistMiddleware.kt
```diff
@@ -0,0 +1,170 @@
...
+    override fun invoke(
+        context: MiddlewareContext<HomeFragmentState, HomeFragmentAction>,
+        next: (HomeFragmentAction) -> Unit,
+        action: HomeFragmentAction
+    ) {
+        if (!blocklistLoaded) {
+            store = context.store
+            return
+        }
+        next(handleUpdate(context.state, action))
```
pocmo, line 51: This pattern looks fragile to me:
* You are dropping actions on the floor (by not calling `next()`). This is problematic since someone wanted those to be dispatched. :)
* Later code uses `store?`. So that code kind of expects you ran into this block to get a store reference and dropped at least one action.
* There seems to be a race since there's no guarantee the loading and the setting of the reference happens in a specific order? They also seem to be read/set from different threads... which may require `@volatile` or `synchronized` to guarantee visibility.
* Why is `filterLatestState()` needed in the first place? Is it an alternative to filtering upfront?

MatthewTighe: - Good call on the dropped actions. I'll fix that
- The `store?` call from `filterLatestState` should only be called once, when the blocklist loads.
- Is this third bullet in reference to `store`? I hadn't considered that the middleware is `invoked()` from a background thread. Thanks for calling that out. 
- I think the following case is still technically possible:
  1. Initial state is filtered by fragment
  2. State is updated by dispatched action before blocklist is loaded within the middleware
  3. If `filterLatestState` were not called when blocklist was loaded, state would remain unfiltered until the next action was intercepted

I think two options for improving this implementation would be:
1. Block the UI thread until the blocklist has been loaded in the middleware, then use the middleware to filter the initial state directly. This has a higher risk of ANRs and slow loads
2. Remove the blocking call from the fragment and completely rely on `filterLatestState` once the blocklist has been loaded. Add more thread safety. The middleware would require the `Store` or initial state on construction, in case it wasn't captured during a dispatch. Higher risk of blocked items being visible and UI jank if those items are displayed and then removed
3. Switch from `DataStore` to `Settings`, hiding the async IO read but hugely simplifying the workflow.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
MatthewTighe: I think you're right, that's easiest for now. I'd be interested in working to develop better patterns around the use of `DataStore`s eventually, but this is seeming like a awkward way to slot this in here.

gabrielluong: Ya, I looked into this briefly in https://github.com/mozilla-mobile/fenix/issues/19197 and https://github.com/mozilla-mobile/fenix/pull/19999. My discussion with @jonalmeida concluded with the need to develop an incremental and long term plan to migrate existing shared preferences (for example, fragment by fragment to ensure app stability), agree on a plan, pattern and troubleshoot various scenarios via an RFC.

gabrielluong: That said there's no problem with using datastore for new features as you see for Pocket, but perhaps maybe something for preferences you can imagine you might just want 1 datastore to hold all preferences.

MatthewTighe: 👍 I'd definitely be interested in being included in future discussions about it

app/src/main/java/org/mozilla/fenix/home/HomeFragmentStore.kt
```diff
@@ -167,7 +169,13 @@ private fun homeFragmentStateReducer(
                 recentHistory = state.recentHistory.filterOut(recentSearchGroup?.searchTerm)
             )
         }
+        is HomeFragmentAction.RemoveRecentTab -> {
+            state.copy(recentTabs = state.recentTabs.filterNot { it == action.recentTab })
```
csadilek, line 173: I think we will need to be able to remove recent tabs (and possibly also recent bookmarks) by their `url` or better even `id`, as opposed to relying on object equality. The reason is that the state could be independently updated (and therefore copied) via an observer e.g.: https://github.com/mozilla-mobile/fenix/blob/main/app/src/main/java/org/mozilla/fenix/home/recenttabs/RecentTabsListFeature.kt#L38

This is very likely to happen for tabs as numerous state changes come in asynchronously while the tab is loading.

Definitely an edge case, but could lead to strange bugs where the user removes a recent tab but the tab they tapped on is out of date relative to the store as the UI hasn't updated yet: https://github.com/mozilla-mobile/fenix/blob/main/app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabViewHolder.kt#L38

So to avoid this ordering problem might be easiest to allow removal by ID / url.

csadilek: This is also a type of bug we had in various places before where we passed the `TabSessionState` around, which wasn't the most up-to-date one anymore, as the object was changed (and therefore copied) in the store. So we switched our use case methods and various interactions over to using the tab.id instead, and look up the up-to-date tab object via the store.

MatthewTighe: Can the `id` ever be equivalent between a `Tab` and a `SearchGroup`?

If this is a common enough comparison, does it make sense to add a method to check equality by id? Something along these lines:

```
sealed class RecentTab
...
    fun hasSameId(other: RecentTab): Boolean =
        this.resolvedId() == other.resolvedId()
    // or
    fun hasSameId(other: RecentTab): Boolean = when {
        this is Tab && other is Tab -> this.state.id == other.state.id
        this is SearchGroup && other is SearchGroup -> this.tabId == other.tabId
        else -> false
    }
                
    
    private fun resolvedId(): String = when(this) {
        is Tab -> state.id
        is SearchGroup -> tabId
    }
}
```

MatthewTighe:
Reworked the implementation to to use `Settings` instead of a `DataStore`, which simplified a lot of the internals.

app/src/main/java/org/mozilla/fenix/home/HomeFragmentStore.kt
csadilek: Not currently as we won't display a tab in jump back in if it's also in the most recent group. That's so we don't display it twice. 

But if I read this ticket correctly, we also don't plan to support blocking of groups/search terms right now. So I assume there won't be a remove option for search groups is jump back in anyway? If we display a remove option for a groups in jump back in, what would it do? If we just remove the `RecentTab.SearchGroup` it would be immediately be visible again if home is re-opened. We could remove the entire tab group, just keep the tabs. We have `DisbandSearchGroupAction` for that, but not sure that's what we want.

csadilek: So, for individual tabs we can block the URL and remove it from the store using its ID/url. For removing groups from "Jump Back In" we need to clarify what we want to happen. Removing a group from recently visited removes the corresponding metadata and disbands the group. Maybe we want to do the same here, so we could just re-use that functionality and existing action: https://github.com/mozilla-mobile/fenix/blob/main/app/src/main/java/org/mozilla/fenix/home/recentvisits/controller/RecentVisitsController.kt#L109

MatthewTighe: It was determined that adding a remove option for search groups in jump back in is out-of-scope for this ticket. I opened #23785 as a follow-up to add that functionality.

For now, I updated tabs to filter based on their ID, and bookmarks to filter by their URL. Thanks again for the detailed feedback!

MatthewTighe:
Apologies for the long delay in updates to this! My attention got shifted pretty hard to all the wallpaper stuff. 

This should be ready for another review cycle, so I'm re-tagging the original reviewers in case anyone would like to take another look.

@csadilek @gabrielluong @Amejia481

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -234,34 +235,37 @@ class HomeFragment : Fragment() {
             ::dispatchModeChanges
         )
 
+        val blocklistMiddleware = BlocklistMiddleware(requireContext().settings())
+        val initialState = HomeFragmentState(
+            collections = components.core.tabCollectionStorage.cachedTabCollections,
+            expandedCollections = emptySet(),
+            mode = currentMode.getCurrentMode(),
+            topSites = components.core.topSitesStorage.cachedTopSites,
```
gabrielluong, line 243: Looks like this is rebased incorrectly since it's missing the `.sort()`. I think it would be better undo this change where we move the `HomeFragmentState` into a variable and just keep it inline.

app/src/main/java/org/mozilla/fenix/home/HomeFragmentStore.kt
```diff
@@ -264,3 +274,13 @@ internal fun List<RecentlyVisitedItem>.filterOut(groupTitle: String?): List<Rece
...
     }
 }
+
+/**
+ * Removes a [RecentTab.Tab] from a list of [RecentTab]. [RecentTab.SearchGroup]s will not be filtered.
+ *
+ * @param tab [RecentTab] to remove from the list
+ */
+@VisibleForTesting
+internal fun List<RecentTab>.filterOutTab(tab: RecentTab): List<RecentTab> = filterNot {
```
gabrielluong, line 284: Can we move this into /ext/RecentTab

app/src/main/java/org/mozilla/fenix/theme/FirefoxTheme.kt
```diff
@@ -26,7 +26,6 @@ fun FirefoxTheme(
     content: @Composable () -> Unit
 ) {
     val colors = if (darkTheme) darkColorPalette else lightColorPalette
-
```
gabrielluong, line 29: nit: let's undo this.

app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabs.kt
```diff
@@ -67,7 +76,8 @@ import org.mozilla.fenix.theme.FirefoxTheme
 fun RecentTabs(
     recentTabs: List<RecentTab>,
     onRecentTabClick: (String) -> Unit = {},
-    onRecentSearchGroupClicked: (String) -> Unit = {}
+    onRecentSearchGroupClicked: (String) -> Unit = {},
+    menuItems: List<RecentTabMenuItem>
```
gabrielluong, line 80: Can we move menuItems below recentTabs? I prefer to keep the callback functions closer to bottom of the parameter list with the exception of `@Composeable content`

app/src/main/java/org/mozilla/fenix/home/HomeFragmentStore.kt
MatthewTighe: Sure! Should I move the similar extension above it into /ext/RecentlyVisited while I'm at it?

app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabs.kt
MatthewTighe: Yep! Missed it on my readthrough, thanks. Updated throughout

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
MatthewTighe: Reverted this to inline, using a `run` block to try and minimize the diff

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -234,9 +235,10 @@ class HomeFragment : Fragment() {
             ::dispatchModeChanges
         )
 
+        val blocklistMiddleware = BlocklistMiddleware(requireContext().settings())
```
gabrielluong, line 238: Can probably inline this as well.

app/src/main/java/org/mozilla/fenix/home/blocklist/BlocklistMiddleware.kt
```diff
@@ -0,0 +1,140 @@
...
+        recentTabs = recentTabs,
+        recentBookmarks = recentBookmarks,
+        recentHistory = recentHistory,
+        mode = mode,
+        collections = collections,
+        tip = tip,
+        showCollectionPlaceholder = showCollectionPlaceholder
+    )
+
+    fun filterState(state: HomeFragmentState): HomeFragmentState =
```
gabrielluong, line 92: This will be required from the linter soon when we upgrade it, but we want to add a kDoc for public functions

app/src/main/java/org/mozilla/fenix/home/blocklist/BlocklistMiddleware.kt
```diff
@@ -0,0 +1,140 @@
...
+import mozilla.components.lib.state.Middleware
+import mozilla.components.lib.state.MiddlewareContext
+import org.mozilla.fenix.home.HomeFragmentAction
+import org.mozilla.fenix.home.HomeFragmentState
+import org.mozilla.fenix.home.recentbookmarks.RecentBookmark
+import org.mozilla.fenix.home.recenttabs.RecentTab
+import org.mozilla.fenix.home.recentvisits.RecentlyVisitedItem
+import org.mozilla.fenix.utils.Settings
+
+class BlocklistMiddleware(
```
gabrielluong, line 16: Add a kDoc for the class

app/src/main/java/org/mozilla/fenix/home/blocklist/BlocklistMiddleware.kt
```diff
@@ -0,0 +1,140 @@
...
+
+        domainsToStrip.forEach { domain ->
+            if (protocolStripped.startsWith("$domain.")) {
+                return protocolStripped.substring(domain.length + 1)
+            }
+        }
+        return protocolStripped
+    }
+
+    @JvmName("filterRecentBookmark")
```
gabrielluong, line 120: Do we need theses `@JvmName`? I am mostly curious because we don't have a history of its usage and I don't know much about their usage.

app/src/main/java/org/mozilla/fenix/home/recentbookmarks/interactor/RecentBookmarksInteractor.kt
```diff
@@ -25,4 +25,10 @@ interface RecentBookmarksInteractor {
      * recently saved bookmarks on the home screen.
      */
     fun onShowAllBookmarksClicked()
+
+    /**
+     * Removes a bookmark from the recent bookmark list. Called when a user clicks the "Remove"
+     * button for recently saved bookmarks on the home screen.
```
gabrielluong, line 31: Should add the @param

app/src/main/java/org/mozilla/fenix/home/recentbookmarks/view/RecentBookmarks.kt
```diff
@@ -49,7 +57,8 @@ import org.mozilla.fenix.theme.FirefoxTheme
 @Composable
 fun RecentBookmarks(
```
gabrielluong, line 58: Update the kDoc

app/src/main/java/org/mozilla/fenix/home/recenttabs/interactor/RecentTabInteractor.kt
```diff
@@ -27,4 +29,10 @@ interface RecentTabInteractor {
      * tabs.
      */
     fun onRecentTabShowAllClicked()
+
+    /**
+     * Removes a bookmark from the recent bookmark list. Called when a user clicks the "Remove"
+     * button for recently saved bookmarks on the home screen.
```
gabrielluong, line 35: Needs @param

app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabMenuItem.kt
```diff
@@ -0,0 +1,12 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.home.recenttabs.view
+
+import org.mozilla.fenix.home.recenttabs.RecentTab
+
+class RecentTabMenuItem(
```
gabrielluong, line 9: Needs kDoc

app/src/main/java/org/mozilla/fenix/home/blocklist/BlocklistMiddleware.kt
MatthewTighe: I assume this is a JVM type erasure thing, where the boxed type in the various `List<*>` extensions defined here isn't known and the JVM needs a way to differentiate between them. 

Alternatively I could define them as non-extension functions with names like `filterRecentBookmarks` and drop the `blocklist` naming convention.

gabrielluong: Oh I see. Seems okay.

app/src/test/java/org/mozilla/fenix/home/BlocklistMiddlewareTest.kt
```diff
@@ -0,0 +1,288 @@
...
+                collections = store.state.collections,
+                tip = store.state.tip,
+                showCollectionPlaceholder = store.state.showCollectionPlaceholder,
+                recentTabs = store.state.recentTabs,
+                recentBookmarks = listOf(updatedBookmark),
+                recentHistory = store.state.recentHistory
+            )
+        ).joinBlocking()
+
+        assertTrue(store.state.recentBookmarks.isEmpty())
```
gabrielluong, line 133:
```suggestion
        every { mockSettings.homescreenBlocklist } returns setOf("https://www.mozilla.org/")

        val store = HomeFragmentStore(
            HomeFragmentState(),
            middlewares = listOf(BlocklistMiddleware(mockSettings))
        )
        val updatedBookmark = RecentBookmark(url = "https://www.mozilla.org/")
        
        store.dispatch(
            HomeFragmentAction.Change(
                topSites = store.state.topSites,
                mode = store.state.mode,
                collections = store.state.collections,
                tip = store.state.tip,
                showCollectionPlaceholder = store.state.showCollectionPlaceholder,
                recentTabs = store.state.recentTabs,
                recentBookmarks = listOf(updatedBookmark),
                recentHistory = store.state.recentHistory
            )
        ).joinBlocking()

        assertTrue(store.state.recentBookmarks.isEmpty())
```

This isn't an exact science by any means, but the formatting pattern that we seem to go with in testss is that we have 'groupings':
VARIABLES INIT
<new line>
WHEN/EVERY declarations
<new line>
SOME ACTION
<new line>
VERIFY/ASSERT

app/src/main/java/org/mozilla/fenix/home/blocklist/BlocklistMiddleware.kt
```diff
@@ -0,0 +1,151 @@
...
+                    )
+                }
+                is HomeFragmentAction.RecentHistoryChange -> {
+                    action.copy(recentHistory = action.recentHistory.filteredByBlocklist(blocklist))
+                }
+                is HomeFragmentAction.RemoveRecentTab -> {
+                    if (action.recentTab is RecentTab.Tab) {
+                        val url = action.recentTab.state.content.url
+                        val updatedBlocklist = blocklist + url
+                        settings.homescreenBlocklist = updatedBlocklist
```
gabrielluong, line 63: I recalled @csadilek saying that desktop encrypts the sites that are blocked by the user. I imagine we should do the same and not store the set as plaintext.

MatthewTighe: It's noted in the [Jira ticket](https://mozilla-hub.atlassian.net/browse/FNXV2-18720) that desktop stores the list as a comma-separated list of hashes, but I didn't think it was a necessarily a requirement. I assumed it was done that way to speed up comparisons, not necessarily for encryption. I figured the size of the list was likely to remain relatively small and I wasn't sure how to solve the hash collision problem, so I went with the straightforward implementation

gabrielluong: Let's check with @csadilek to make sure. I can imagine we might not want to store a list of plaintext user blocked website. Often, following the desktop implementation is an implicit requirement. It's just that not everyone is aware of the desktop details and can often be missed in the product requirements.

app/src/test/java/org/mozilla/fenix/home/BlocklistMiddlewareTest.kt
MatthewTighe: I can move the pure variable declarations up, if you want. The mock setups are required before the initialization of the middleware, otherwise `settings` won't have an answer for `private var blocklist: Set<String> = settings.homescreenBlocklist`.

Is there a document where some of these specific formatting gotchas live? JW if there's something I can point contributors (and myself) to in order to avoid bikeshedding PR cycles since I assume some of these would be hard to create lint rules for

gabrielluong: Only way is probably to look at a couple of existing tests file and just adopt their pattern. Usually you just want to blend in with what the project is already doing. I only called this out because of the inconsistency usage new lines between the variable declaration. The mock setup before the initialization is fine.

gabrielluong: I will look to write something to help formalize this. Apologizes for the bikeshedding.

MatthewTighe: I moved variable declarations to the spot you suggested.

I definitely agree that having stylistic similarity throughout the code base makes it easier to understand any one part. IMO a "good enough" level would be something along the lines of "follows a given-when-then" format is a bigger win than specific formatting details. I just think it's important to try and keep barrier to entry as low as possible for outside contributors. That said, I do understand if we want to hold org members to a higher standard than outside contributors, and I appreciate the guidance in getting up to par.

Mugurell: Quick comment regarding a project-wide style for unit tests.
Splitting the unit test in 3 sections is a common pattern - https://wiki.c2.com/?ArrangeActAssert
Instead of the proposed 4 sections we could have 3 matching exactly `arrange/act/assert` / `given/when/then` with setup including properties intializations and mocking.

app/src/main/java/org/mozilla/fenix/home/HomeFragmentStore.kt
csadilek: OK, perfect!

app/src/main/java/org/mozilla/fenix/home/blocklist/BlocklistMiddleware.kt
```diff
@@ -0,0 +1,156 @@
...
+ * This [Middleware] reacts to item removals from the home screen, adding them to a blocklist.
+ * Additionally, it reacts to state changes in order to filter them by the blocklist.
+ *
+ * @param settings Blocklist is stored here as a string set
+ */
+class BlocklistMiddleware(
+    private val settings: Settings
+) : Middleware<HomeFragmentState, HomeFragmentAction> {
+
+    private var blocklist: Set<String> = settings.homescreenBlocklist
```
csadilek, line 28: This can be a `val` currently because you're not writing to the `blocklist` field. However, this actually points to a problem. The middleware (and therefore this field) is initialized when the home fragment is created pointing to the instance of `settings.homescreenBlocklist` at that time. Below when handling updates, you're not mutating `homescreenBlocklist`, but setting it to a new instance `updatedBlockList` e.g.: https://github.com/mozilla-mobile/fenix/pull/23070/files#diff-4c994d58d9a22e516dcec6ef8d81c5252ccbcdbbec011d32d12b532189b2fcf2R82

So the `blocklist` field of the middelware will be out-of-sync with `settings` after the update. You could make this work by turning this into a `getter` functino so the lookup happens every time, but I recommend simplifying and adding `val blocklist = settings.homescreenBlocklist` to `invoke` directly.

app/src/main/java/org/mozilla/fenix/home/blocklist/BlocklistMiddleware.kt
```diff
@@ -0,0 +1,156 @@
...
+     * Will filter "Change" actions using the blocklist and use "Remove" actions to update
+     * the blocklist.
+     */
+    override fun invoke(
+        context: MiddlewareContext<HomeFragmentState, HomeFragmentAction>,
+        next: (HomeFragmentAction) -> Unit,
+        action: HomeFragmentAction
+    ) {
+        val state = context.state
+        next(
```
csadilek, line 40: nit: I think `next( big block to transform action )` is hard to read. I would add a method or move out:

```kotlin
val filteredAction = applyBlockList(action)
next(filteredAction)
```

or

```kotlin
next(applyBlockList(action))
```

csadilek (changes requested):
This is coming together nicely! I found one problem and a small thing to change.

app/src/main/java/org/mozilla/fenix/home/blocklist/BlocklistMiddleware.kt
```diff
@@ -0,0 +1,156 @@
...
+                        state.filteredByBlocklist(updatedBlocklist).toChangeAction()
+                    } else {
+                        action
+                    }
+                }
+                is HomeFragmentAction.RemoveRecentBookmark -> {
+                    action.recentBookmark.url?.let { url ->
+                        val updatedBlocklist = blocklist + url.stripAndHash()
+                        settings.homescreenBlocklist = updatedBlocklist
+                        state.filteredByBlocklist(updatedBlocklist).toChangeAction()
```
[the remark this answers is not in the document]
MatthewTighe: This is actually the intended behavior! If an item is removed from any section, it should be removed from all of them (with the exception of top sites). You'll see the same behavior replicated across recent history and recent tabs.

 I wrapped this in a more descriptive function and added a comment to try and increase clarity around the behavior.

app/src/main/java/org/mozilla/fenix/home/blocklist/BlocklistMiddleware.kt
MatthewTighe: The implementation that this touched has moved to `BlocklistHandler`, but should avoid this issue now.

app/src/main/java/org/mozilla/fenix/home/blocklist/BlocklistMiddleware.kt
MatthewTighe: This has been cleaned up

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -260,8 +262,9 @@ class HomeFragment : Fragment() {
                     //  to some state.
                     recentTabs = getRecentTabs(components),
                     recentHistory = emptyList()
-                ),
-                listOf(
+                ).run { blocklistMiddleware.filterState(this) },
```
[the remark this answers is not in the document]
MatthewTighe: This took some rather substantial refactoring, since `fitlerState` also relied on the handful of extension functions that were defined to filter the various state properties. I've left it as a separate commit for now, in case there's desire to see the diff between the two versions.

> We don't want other parts of the app to invoke it directly, see also my comment on the middleware as to why. 

Which comment is that? It makes sense to keep the responsibility of the middleware limited in that way to me in any case. This was mostly a side-effect of trying to keep everything initially encapsulated in the middleware class, and the original design clearly wasn't flexible enough for the end goal.

csadilek (approved):
Looks great to me now! Just two nits and let's land on Monday! 

@gabrielluong Can you take another look too?

app/src/main/java/org/mozilla/fenix/home/blocklist/BlocklistMiddleware.kt
```diff
@@ -0,0 +1,99 @@
...
+                state.toActionFilteringAllState(this)
+            }
+            else -> action
+        }
+    }
+
+    // When an item is removed from any part of the state, it should also be removed from any other
+    // relevant parts that contain it.
+    // This is a candidate for refactoring once context receivers lands in Kotlin 1.6.20
+    // https://blog.jetbrains.com/kotlin/2022/02/kotlin-1-6-20-m1-released/#prototype-of-context-receivers-for-kotlin-jvm
```
csadilek, line 85: Interesting! How do you envision this will look like / improve with context receivers?

MatthewTighe: I might just be excited to try them out 😅 

Replacing the `BlocklistHandler` parameter with a `context(BlocklistHandler)` receiver would mean the explicit `this` arguments could be removed from the calls to the function above, since they are already within a `BlocklistHandler` scope

MatthewTighe: Ultimately I just like defining nested scopes as part of a functional type, rather than trying to resolve different scope levels mentally or with explicit `this@<scope>` syntax.

app/src/main/java/org/mozilla/fenix/home/blocklist/BlocklistHandler.kt
```diff
@@ -0,0 +1,66 @@
...
+
+package org.mozilla.fenix.home.blocklist
+
+import androidx.annotation.VisibleForTesting
+import mozilla.components.support.ktx.kotlin.sha1
+import org.mozilla.fenix.home.recentbookmarks.RecentBookmark
+import org.mozilla.fenix.home.recenttabs.RecentTab
+import org.mozilla.fenix.home.recentvisits.RecentlyVisitedItem
+import org.mozilla.fenix.utils.Settings
+
```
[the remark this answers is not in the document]
MatthewTighe: Will do! Wanted to push the implementation before EoD yesterday in case you were able to take a look this morning, but will put all the finishing touches like that together today

app/src/main/java/org/mozilla/fenix/home/blocklist/BlocklistHandler.kt
```diff
@@ -0,0 +1,66 @@
...
+    // This kind of stripping allows us to match "twitter" to "mobile.twitter.com".
+    // Borrowed from DomainMatcher in A-C
+    val domainsToStrip = listOf("www", "mobile", "m")
+
+    domainsToStrip.forEach { domain ->
+        if (stripped.startsWith("$domain.")) {
+            return stripped.substring(domain.length + 1)
+        }
+    }
+    return stripped
```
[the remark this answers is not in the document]
MatthewTighe: I think that's a great idea. This functionality is replicated/split in a few places in Fenix and A-C. See ext/String in Fenix and ext/Uri in A-C, for example.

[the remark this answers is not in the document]
MatthewTighe: [A-C issue](https://github.com/mozilla-mobile/android-components/issues/11743)

app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabs.kt
```diff
@@ -102,17 +114,26 @@ fun RecentTabs(
...
  */
+@OptIn(ExperimentalFoundationApi::class)
 @Suppress("LongParameterList")
 @Composable
 private fun RecentTabItem(
     tab: RecentTab.Tab,
+    menuItems: List<RecentTabMenuItem>,
     onRecentTabClick: (String) -> Unit = {}
 ) {
+
```
gabrielluong, line 125: nit: extra new line

gabrielluong (approved):

app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabs.kt
MatthewTighe: Removed

Issue #22559, opened 2021-11-23
Update leakcanary version to 2.7 or above be able to targe Android 12


Amejia481:
Related https://github.com/square/leakcanary/issues/2076 and https://github.com/mozilla-mobile/fenix/issues/22539
```
/app/build/outputs/apk/debug/app-arm64-v8a-debug.apk
com.android.ddmlib.InstallException: INSTALL_PARSE_FAILED_MANIFEST_MALFORMED: Failed parse during installPackageLI: /data/app/vmdl1056481257.tmp/base.apk (at Binary XML file line #978): leakcanary.internal.activity.LeakLauncherActivity: Targeting S+ (version 31 and above) requires that an explicit value for android:exported be defined when intent filters are present
        at com.android.ddmlib.internal.DeviceImpl.installRemotePackage(DeviceImpl.java:1310)
        at com.android.ddmlib.internal.DeviceImpl.installPackage(DeviceImpl.java:1136)
        at com.android.ddmlib.internal.DeviceImpl.installPackage(DeviceImpl.java:1112)
        at com.android.ddmlib.internal.DeviceImpl.installPackage(DeviceImpl.java:1101)
        at com.android.build.gradle.internal.testing.ConnectedDevice.installPackage(ConnectedDevice.java:127)
        at com.android.build.gradle.internal.tasks.InstallVariantTask.install(InstallVariantTask.java:159)
        at com.android.build.gradle.internal.tasks.InstallVariantTask.lambda$doTaskAction$1(InstallVariantTask.java:96)
        at com.android.builder.testing.api.DeviceProvider.use(DeviceProvider.java:53)

```

mcarare:
Closed in https://github.com/mozilla-mobile/fenix/pull/23082.

PR #23082, opened 2022-01-06
For #22559: Update leak canary version to 2.8.1

mcarare:
### Pull Request checklist

1. click on Show All Checks,


app/src/debug/java/org/mozilla/fenix/DebugFenixApplication.kt
```diff
@@ -8,20 +8,31 @@ import android.os.StrictMode
...
         }
+
         updateLeakCanaryState(isEnabled)
     }
 
     override fun updateLeakCanaryState(isEnabled: Boolean) {
-        AppWatcher.config = AppWatcher.config.copy(enabled = isEnabled)
-        LeakCanary.config = LeakCanary.config.copy(dumpHeap = isEnabled)
+        LeakCanary.showLeakDisplayActivityLauncherIcon(isEnabled)
+        components.strictMode.resetAfter(StrictMode.allowThreadDiskReads()) {
```
mcarare, line 34: see https://github.com/square/leakcanary/issues/2117.
We could watch the issue and remove this reset after that has been solved.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -285,7 +285,7 @@ class HomeFragment : Fragment() {
                 storage = components.core.topSitesStorage,
                 config = ::getTopSitesConfig
             ),
-            owner = viewLifecycleOwner,
+            owner = this,
```
mcarare, line 288: This is for testing purposes only, to generate a memory leak.

app/src/main/res/values/bools.xml
```diff
@@ -2,9 +2,11 @@
...
    - License, v. 2.0. If a copy of the MPL was not distributed with this
    - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
-<resources>
+<resources xmlns:tools="http://schemas.android.com/tools">
     <bool name="theme_is_light">true</bool>
     <bool name="API28">false</bool>
     <bool name="underAPI28">true</bool>
     <bool name="tablet">false</bool>
+    <!-- disable the automatic install of Leak Canary -->
+    <bool name="leak_canary_watcher_auto_install" tools:ignore="UnusedResources">false</bool>
```
pyricau, line 11: You can remove this and depend on leakcanary-android-core instead of leakcanary-android

https://square.github.io/leakcanary/changelog/#androidx-app-startup

app/src/debug/java/org/mozilla/fenix/DebugFenixApplication.kt
```diff
@@ -8,20 +8,31 @@ import android.os.StrictMode
...
+import org.mozilla.fenix.ext.application
 import org.mozilla.fenix.ext.getPreferenceKey
 
 class DebugFenixApplication : FenixApplication() {
 
     override fun setupLeakCanary() {
+        if (!AppWatcher.isInstalled) {
+            AppWatcher.manualInstall(
+                application = application,
+                watchersToInstall = AppWatcher.appDefaultWatchers(application)
```
pyricau, line 20: Since you're just using the default set up, you could rely on kotlin default params instead: `AppWatcher.manualInstall(application)`

One question: why do you need to to check `if (!AppWatcher.isInstalled) {` ?  Are you not disabling auto install on all build flavors? Or is this because `setupLeakCanary()` can be called several times per Application lifetime?

mcarare: @pyricau thanks for taking a look at this. 

The check is just a precautionary measure. The setup method is implemented just in the debug app. 

For the default param, I chose to state them explicitly, in case we later want to filter the watchers we want to use.

app/src/debug/java/org/mozilla/fenix/DebugFenixApplication.kt
mcarare: cc @mcomella

mcomella: Since it's in the debug fenix application, I'm not too concerned. Thanks for the heads up.

Amejia481 (approved):

PR #23098, opened 2022-01-06
For #22851: Load wallpapers dynamically from the assets directory.

Amejia481:
### Pull Request checklist

1. click on Show All Checks,


MatthewTighe (commented):
I'm wondering how we might adapt this to handle images of different resolutions. I believe UX will be providing images in several resolution sets to best accommodate different screen sizes. By using images as drawable resources, the Android system will choose the most appropriate resolution based on the device's pixel density

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperManager.kt
```diff
@@ -85,9 +88,35 @@ class WallpaperManager(private val settings: Settings) {
...
+            }
+        } ?: return
+        wallpapers = (listOf(defaultWallpaper) + loadedWallpapers)
+    }
+
+    companion object {
+        private const val WALLPAPERS_DIRECTORY = "wallpapers"
+        private val wallpaperToIsDark = mapOf("wallpaper_1.png" to true, "wallpaper_2.png" to false)
+        val defaultWallpaper = Wallpaper(name = "homeBackground")
+    }
```
MatthewTighe, line 121: It feels just a bit off to me to have an instantiated class as a static property. Primitive data and constants feel like a better fit. Maybe moving the `DEFAULT_RESOURCE` to the companion here and moving the `defaultWallpaper` inside the `WallpaperManager` would make sense.

Another thing you could consider is creating a list of wallpapers within the wallpaper manager. This might be easier to swap out later as well, once we better define how we're going to load the list from disk. For example:

```
class WallpaperManager {
    private val wallpapers = listOf(Wallpaper1, Wallpaper2, Wallpaper3)
}
```
could be later be easily updated to 
```
class WallpaperManager {
    private val wallpapers = loadWallpaperListFromDisk()
}
```

Could even go so far as to define it as an interface now, making it even easier to refactor later and test now:

```
interface WallpaperLoader {
    fun loadWallpaperList(): List<Wallpaper>
}

class StaticListWallpaperLoader() {
    override fun loadWallpaperList(): List<Wallpaper>
}
```

If you went in a direction like that, it could also make sense to include a `fun loadWallpaperImage` in it to retrieve the images from disk as well.

app/src/main/java/org/mozilla/fenix/wallpapers/Wallpaper.kt
```diff
@@ -4,13 +4,14 @@
...
+ * A class that represents a available wallpaper and its state.
  */
-enum class Wallpaper(val drawable: Int, val isDark: Boolean) {
-    FIRST(drawable = R.drawable.wallpaper_1, isDark = true),
-    SECOND(drawable = R.drawable.wallpaper_2, isDark = false),
-    NONE(drawable = R.attr.homeBackground, isDark = false);
+data class Wallpaper(val name: String, val resource: Drawable? = null, val isDark: Boolean = false) {
+    companion object {
+        const val DEFAULT_RESOURCE = R.attr.homeBackground
+    }
```
MatthewTighe, line 16: I don't think `resource` should be nullable here. It feels like the most essential part of the data, since we can't draw a `Wallpaper` without its resource.

Also, it might make sense to move `DEFAULT_RESOURCE` into the companion of `WallpaperManager` to keep all the static defintions grouped together.

MatthewTighe (commented):
This is looking good! We should sync up soon to talk more about how we want to track the wallpaper list. Also, it'd be great to start seeing some units tests around these changes. Makes me a little nervous to ship untested code 😅

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperManager.kt
```diff
@@ -92,9 +103,41 @@ class WallpaperManager(private val settings: Settings) {
...
+    private fun isLandscape(context: Context): Boolean {
+        return context.resources.configuration.orientation == Configuration.ORIENTATION_LANDSCAPE
+    }
+
+    fun loadWallpapers() {
+        availableWallpapers.clear()
+        availableWallpapers.add(defaultWallpaper)
+        availableWallpapers.addAll(wallpaperStorage.all())
+        currentWallpaper = getCurrentWallpaperFromSettings()
+    }
```
MatthewTighe, line 132: Alternatively to the above comment, you could change the property `availableWallpapers` property to be mutable, since we overwrite the list every time it changes anyway:

```
var availableWallpapers = listOf(defaultWallpaper)
    private set

fun loadWallpapers() {
    availableWallpapers = listOf(defaultWallpaper, *wallpaperStorage.all())
}
```

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperManager.kt
```diff
@@ -92,9 +103,41 @@ class WallpaperManager(private val settings: Settings) {
...
+
+    private fun isLandscape(context: Context): Boolean {
+        return context.resources.configuration.orientation == Configuration.ORIENTATION_LANDSCAPE
+    }
+
+    fun loadWallpapers() {
+        availableWallpapers.clear()
+        availableWallpapers.add(defaultWallpaper)
+        availableWallpapers.addAll(wallpaperStorage.all())
+        currentWallpaper = getCurrentWallpaperFromSettings()
```
MatthewTighe, line 131: I think this `currentWallpaper = getCurrentWallpaperFromSettings()` line can be omitted, as it should be run when the object is instantiated

app/src/main/java/org/mozilla/fenix/components/Components.kt
```diff
@@ -143,8 +144,12 @@ class Components(private val context: Context) {
         AddonManager(core.store, core.engine, addonCollectionProvider, addonUpdater)
     }
 
+    val wallpaperStorage by lazyMonitored {
+        WallpapersAssetsStorage(context)
+    }
```
MatthewTighe, line 149: I think this could be inlined in the `wallpaperManager` creation, as I don't think it will be used by other objects.

app/src/main/java/org/mozilla/fenix/wallpapers/WallpapersAssetsStorage.kt
```diff
@@ -0,0 +1,43 @@
...
+
+    private fun AssetManager.readArray(fileName: String) = JSONArray(
+        open(fileName).bufferedReader().use {
+            it.readText()
+        }
+    )
+
+    companion object {
+        private const val WALLPAPERS_DIRECTORY = "wallpapers"
+    }
```
MatthewTighe, line 42: We could avoid an extra object instantiation by moving this inline, or as a private property on `WallpapersAssetsStorage`.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -214,9 +228,16 @@ private fun WallpaperThumbnailItem(
...
+        val context = LocalContext.current
+        val wallpaperManager = context.components.wallpaperManager
+
         WallpaperSettings(
-            wallpapers = Wallpaper.values().toList(),
-            selectedWallpaper = Wallpaper.NONE,
+            defaultWallpaper = WallpaperManager.defaultWallpaper,
+            loadWallpaperResource ={
+                wallpaperManager.loadWallpaperFromAssets(it, context)
+            },
```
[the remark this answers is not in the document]
Amejia481: Yes it works :)

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperManager.kt
Amejia481: The issue It's not completely straightforward, imagine you have a selected wallpaper that it is one that it is loaded from assets for example `wallpaper_1`, in settings you will have  `settings.currentWallpaper` with the value `wallpaper_1`, when `var currentWallpaper: Wallpaper = getCurrentWallpaperFromSettings()` is initialized we won't be able to find the `wallpaper_1` on `availableWallpapers` and we will fallback to  `defaultWallpaper`, as `availableWallpapers` doesn't have the `wallpaper_1` until we load it from assets.

```kotlin
    val availableWallpapers = mutableListOf(defaultWallpaper)

    var currentWallpaper: Wallpaper = getCurrentWallpaperFromSettings()
        set(value) {
            settings.currentWallpaper = value.name
            field = value
        }

    private fun getCurrentWallpaperFromSettings(): Wallpaper {
        val currentWallpaper = settings.currentWallpaper
        return if (currentWallpaper.isEmpty()) {
            defaultWallpaper
        } else {
            availableWallpapers.find { it.name == currentWallpaper } ?: defaultWallpaper
        }
    }
```

To make sure the `currentWallpaper` is in sync with what we have in `settings.currentWallpaper` we will need to init `currentWallpaper` after we load all the wallpapers from assets.

app/src/main/java/org/mozilla/fenix/components/Components.kt
Amejia481: This could make testing difficult as we won't have control over the `wallpaperStorage` passing it as dependency we won't have to use the real object for testing.

Amejia481:
@MatthewTighe I updated the PR with the suggested changes :)

app/src/main/java/org/mozilla/fenix/components/Components.kt
MatthewTighe: Sorry, I meant inlined in the call to the constructor itself. That way it we still have the ability to inject mocked or faked dependencies during test:

```
val wallpaperManager by lazyMonitored {
        WallpaperManager(settings, WallpapersAssetsStorage(context))
    }
}
```

Amejia481: Thanks for the clarification 👍🏽

Amejia481: Updated!

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperManager.kt
MatthewTighe: Those things being out of sync seems like it could be a cause of concern. I wonder if we should change the `WallpaperManager::loadWallpapers` function to run on initialization, instead of in `FenixApplication::initializeWallpapers`. Then, the two properties should be in sync. 

My concern there would be a perf regression on app startup. I think we would need to revisit this in a followup to make the wallpaper loading asynchronous.

Amejia481: They will be in sync the first time when we load, but if we do subsequent loads like after editing the `wallpapers.json` file it will be out of sync again. In this case, I think there is no way to avoid the performance hit as even moving the load to the initialization we will hit load when we try to access the `wallpaperManager`  normally on the home screen.

Amejia481: At least it could help to reduce the time when users are starting the app without entering the home screen

Amejia481: I updated the PR

app/src/main/assets/wallpapers/wallpapers.json
```diff
@@ -0,0 +1,14 @@
...
+    "landscape": "wallpapers/wallpaper_1.png",
+    "isDark":true
+  },
+  {
+    "name": "wallpaper_2",
+    "portrait": "wallpapers/wallpaper_2.png",
+    "landscape": "wallpapers/wallpaper_2.png",
+    "isDark":false
+  }
+]
```
MatthewTighe, line 14: nit: looks like we're missing a newline here

PR #23113, opened 2022-01-07
closes #22831: add wallpaper settings and refactor to use state pattern

### LINKED ISSUE #22831: Create a settings page for allowing users to change home screen background

MatthewTighe:
Previous implementation can be found in [this closed PR](https://github.com/mozilla-mobile/fenix/pull/22953)

image

### Pull Request checklist

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -492,6 +492,12 @@ class HomeFragment : Fragment() {
             track(Event.HomeScreenViewCount)
         }
 
+        consumeFrom(appStore) {
+            if (shouldEnableWallpaper()) {
+                it.wallpaper.applyToView(binding.homeLayout)
+            }
+        }
+
```
MatthewTighe, line 500: @gabrielluong mentioned that `consumeFrom` should be used over `Store.observe(viewLifecycleOwner)`. It looks like there are some [comments](https://github.com/mozilla-mobile/android-components/blob/379cb84d8874da4e3b7629f370f745ec2d4b1ec0/components/lib/state/src/main/java/mozilla/components/lib/state/ext/Fragment.kt#L38) describing a bug in viewLifecycleOwner that require this.

@pocmo started a discussion about it in [this issue](https://github.com/mozilla-mobile/android-components/issues/3828). It's been quite a while since this issue was opened, and I wonder if it's still applicable. We are recreating the activity to change the theme when the wallpaper changes, and I saw no issues when my implementation used

```
appStore.observe {
    lifecycleScope.launch(Main) {
    ...
    }
}

Comment (gabrielluong) [reply app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt:500]: Now that I am thinking about this I think we should probably look at wrapping this in a Feature such as `TopSitesFeature` or `RecentTabs`. Ideally, we don't have too much logic inside of `HomeFragment`.

File: app/src/main/res/values/strings.xml [hunk 2]
```diff
@@ -461,6 +461,12 @@
     <!-- Title for the customize home screen section with Pocket. -->
     <string name="customize_toggle_pocket">Pocket</string>
 
+    <!-- Wallpapers -->
+    <!-- Title for preference to open wallpapers -->
+    <string name="customize_wallpapers">Wallpapers</string>
+    <!-- Content description for various wallpapers. The first parameter is the name of the wallpaper -->
+    <string name="content_description_wallpaper_name">Wallpaper Item: %1$s</string>
```
gabrielluong, line 468: Can you add an appropriate string prefix for this? Since it's grouped under `<!-- Wallpapers -->`, you will find in our `strings.xml` that we tend to align towards using a prefix to indicate strings that are within the same screen. However, this might mean `customize_wallpapers ` should probably just be under `<!-- Customize Home -->` and used as the title. You will have to look and see what makes sense.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -0,0 +1,163 @@
...
+import org.mozilla.fenix.R
+import org.mozilla.fenix.components.AppStore
+import org.mozilla.fenix.components.appstate.AppAction
+import org.mozilla.fenix.theme.FirefoxTheme
+import org.mozilla.fenix.wallpapers.Wallpaper
+
+/**
+ * A screen for various settings related to wallpapers.
+ *
+ * @param wallpaperManager Manager for retrieving and updating current wallpaper.
```
Amejia481, line 41: nit: Needs to be updated with the `AppStore` :)

app/src/test/java/org/mozilla/fenix/components/appstate/AppStoreReducerTest.kt
```diff
@@ -0,0 +1,30 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.components.appstate
+
+import org.junit.Assert.assertEquals
+import org.junit.Test
+import org.mozilla.fenix.wallpapers.Wallpaper
+
+class AppStoreReducerTest {
```
gabrielluong, line 11: I don't know if we typically have tests against the reducer (vs the Store). Normally, I would expect that we write tests against the AppStore and dispatch the actions and check its resulting state.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -0,0 +1,163 @@
...
+ */
+@Composable
+fun WallpaperSettings(
+    appStore: AppStore
```
gabrielluong, line 45: We do not want to pass in the AppStore into any @Composable function. The reason for this is that it will react to any changes in the AppStore compare to just the wallpaper state. You will want to `observeAsComposableState` prior to calling the Composable content https://github.com/mozilla-mobile/fenix/pull/23113/files#diff-ecfc0b5554895ea8d6552fbd4096ae0b65cb027f3d6bd6e854f9a673ef59c9a8R30. See https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/java/org/mozilla/fenix/home/recentbookmarks/view/RecentBookmarksViewHolder.kt#32

app/src/main/java/org/mozilla/fenix/components/appstate/AppState.kt
```diff
@@ -13,5 +14,6 @@ import mozilla.components.lib.state.State
  * should be expanded when the tray is opened.
  */
 data class AppState(
-    val inactiveTabsExpanded: Boolean = false
+    val inactiveTabsExpanded: Boolean = false,
+    val wallpaper: Wallpaper = Wallpaper.NONE,
```
gabrielluong, line 18: Add a new @property

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -0,0 +1,163 @@
...
+}
+
+/**
+ * A image of a drawable wallpaper.
+ *
+ * @param wallpaper The wallpaper to use.
+ */
+@Composable
+private fun WallpaperImageThumbnail(wallpaper: Wallpaper) {
+    val contentDescription = stringResource(
```
gabrielluong, line 144: I would just inline this since the variable is only used once and consumed immediately. Also, it's fairly easy to tell what it's doing.

app/src/main/res/xml/home_preferences.xml
```diff
@@ -29,6 +29,11 @@
         android:title="@string/customize_toggle_pocket"
         app:isPreferenceVisible="false" />
 
+    <androidx.preference.Preference
+        android:key="@string/pref_key_wallpapers"
+        android:title="@string/customize_wallpapers"
+        app:isPreferenceVisible="true" />
```
gabrielluong, line 35: Should this be visible by default already?

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -0,0 +1,163 @@
...
+            BorderStroke(width = 1.dp, color = FirefoxTheme.colors.borderDefault),
+            thumbnailShape
+        )
+    }
+    val background = if (wallpaper == Wallpaper.NONE) {
+        Modifier.background(color = FirefoxTheme.colors.layer1)
+    } else {
+        Modifier
+    }
+    Box(
```
gabrielluong, line 119: I would add new line before to visually separate out the variable declarations from the actual start of the compose ui.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -0,0 +1,163 @@
...
+@Composable
+fun WallpaperSettings(
+    appStore: AppStore
+) {
+    val selectedWallpaper = appStore.observeAsComposableState { state -> state.wallpaper }.value
+    Surface(color = FirefoxTheme.colors.layer2) {
+        WallpaperThumbnails(
+            wallpapers = Wallpaper.values().toList(),
+            selectedWallpaper = selectedWallpaper ?: Wallpaper.NONE,
+            onSelectionChanged = { appStore.dispatch(AppAction.UpdateWallpaper(it)) }
```
gabrielluong, line 52: You should surface a `onWallpaperSelected` or some variant lambda function for `WallpaperSettings`. Check out some of our existing compose components.

app/src/main/res/xml/home_preferences.xml
gabrielluong: If this is behind a feature flag and only shown for a subset of our users, we tend to default to false to avoid showing the preference and hiding it immediately (causing a flash).

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
MatthewTighe: I'm confused as to the difference, could you explain it in a little more detail? My understanding was that in either case, we're performing (compose) `State` reads in a `Composition`, so we would get smart recompositions down the UI tree. Why is observing the `Store` in one level above preferred?

gabrielluong: I think this was the best resource I found to describe my intentions. https://developer.android.com/jetpack/compose/state#state-hoisting

The composable caller in this cause happens in https://github.com/mozilla-mobile/fenix/pull/23113/files#diff-ecfc0b5554895ea8d6552fbd4096ae0b65cb027f3d6bd6e854f9a673ef59c9a8R30. 

Otherwise, I am seeing that in `WallpaperSettings`, you are already rendering UI, you don't necessarily want to pass and duplicate the AppStore, you may want to make this composable more generic and not necessarily depend on the AppStore because all it cares about is the `wallpaper` state.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
Amejia481: Could we do this in follow up? It will be deal to start testing the settings page in nightly as soon as possible.

gabrielluong: I guess that's fine, but I would maybe recommend landing this after the cut.

app/src/main/java/org/mozilla/fenix/components/appstate/AppAction.kt
```diff
@@ -6,10 +6,14 @@ package org.mozilla.fenix.components.appstate
...
 import org.mozilla.fenix.components.AppStore
+import org.mozilla.fenix.wallpapers.Wallpaper
 
 /**
  * [Action] implementation related to [AppStore].
  */
 sealed class AppAction : Action {
+    object InitAction : AppAction()
     data class UpdateInactiveExpanded(val expanded: Boolean) : AppAction()
+    object SwitchToNextWallpaper : AppAction()
```
jonalmeida, line 17: 🚗 Future consideration: as we add more actions in the future, we should organize these actions to have some categorization of actions similar to BrowserAction (e.g. TabListAction)

jonalmeida: Same goes for the data classes in AppState.

Neither of these need to be actioned on for now. 🙂

MatthewTighe:
Closing this for now in favor of an approach with less refactoring

PR #23122, opened 2022-01-08
Fix lincense in tests

BraisGabin:
I'm creating a PR updating detekt from `1.17.1` to `1.19.0`. Updating it detekt found that there were some test without the license or with one that doesn't match the one defined on the detekt configuration. This PR fixes that (my next PR will update detekt to ensure that this doesn't happen again).

PR #23123, opened 2022-01-08
Update detekt and it's configuration

BraisGabin:
The first commit of this PR is already committed here #23122. I'll wait for it to be merged and then rebase this one but I need those changes to make detekt happy.

This PR contains lots of small commits that just update the current configuration that Fenix have to one that is more similar to the default one of detekt. It seems that the configuration was created in the RC times of detekt. And detekt changed a lot since them so those should be just a cleanup. They shouldn't change anything in how detekt works.

The big ones are:
- e59450decd580b9bec7dab44557fa7a121f35ee7 this one syncs the current configuration with the default configuration of detekt. It only adds things keeping the current configuration as it is.
- fcdac9183638907bce9bf9fdd70e637644a8672d this one is easier. Now that the configuration of fenix is similar to the default one I used these commands to update the configuration: https://github.com/detekt/detekt/discussions/3558#discussioncomment-483027 and then fix the conflicts.

Note: I saw #23112. I'm OK fixing the conflicts that it could generate and creating this same PR to the Android Component Repo if you think it's valuable.

### DIFF
--- a/app/src/androidTest/java/org/mozilla/fenix/customannotations/SmokeTest.kt
+++ b/app/src/androidTest/java/org/mozilla/fenix/customannotations/SmokeTest.kt
@@ -1,3 +1,7 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
 package org.mozilla.fenix.customannotations
 
 /**
--- a/app/src/androidTest/java/org/mozilla/fenix/helpers/Constants.kt
+++ b/app/src/androidTest/java/org/mozilla/fenix/helpers/Constants.kt
@@ -1,3 +1,7 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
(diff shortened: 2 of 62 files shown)


gabrielluong:
@BraisGabin Just thinking about this quickly. In https://github.com/mozilla-mobile/fenix/pull/23112, I attempted to copy over the config from AC https://github.com/mozilla-mobile/android-components/blob/main/config/detekt.yml. Ideally, we would follow the same rules in both project. One of the bigger change I wanted to bring over was linting for UndocumentedPublicClass/Function and the license linting, but I don't expect anyone (except myself) to write all these kDocs.

In general, I am treating the AC config as the "goal" since they have more rules enabled compare to the Fenix config. Rather than making the 2 configs even more differentiated, I want to suggest that we focus on first updating the AC configs and then syncing those changes back to Fenix.

BraisGabin:
Wow, I didn't see the list of issues that #23112 generates. I think that it could have sense to generate a baseline and add all those issues on it and then work to remove them. But at least you ensure that no new issues are added to the code.

In my setup I have a nightly process that automatically updates the baseline. Because CI doesn't allow to add any new issue to the project what that process does is to remove the already fixed issues so they can't appear again.

> Rather than making the 2 configs even more differentiated, I want to suggest that we focus on first updating the AC configs and then syncing those changes back to Fenix.

Ok! I'll do the same for AC.

BraisGabin:
I'm closing this for now. I will open a new pr if the one from AC get merged.

PR #23145, opened 2022-01-11
closes #22831: add wallpaper settings screen

### LINKED ISSUE #22831: Create a settings page for allowing users to change home screen background

MatthewTighe:
Simplified version of #22953 with some refactoring removed, and to have a clean slate for discussion.

image

image


app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperManager.kt
```diff
@@ -40,17 +40,17 @@ class WallpaperManager(private val settings: Settings) {
...
-            // For the default wallpaper, there is not need to adjust the theme,
-            // as we want to allow users decide which theme they want to have.
-            // The default wallpaper adapts to whichever theme the user has.
-            return
+            when {
+                settings.shouldUseLightTheme -> AppCompatDelegate.MODE_NIGHT_NO
+                settings.shouldUseDarkTheme -> AppCompatDelegate.MODE_NIGHT_YES
+                settings.shouldFollowDeviceTheme -> AppCompatDelegate.MODE_NIGHT_FOLLOW_SYSTEM
+                else -> AppCompatDelegate.MODE_NIGHT_UNSPECIFIED
+            }
```
MatthewTighe, line 53: @Amejia481 I noticed that the theme selection preference in the `Customize` settings page was ending up with both "Follow device theme" and either "Light" or "Dark" selected. I assume we only want one of those to ever be selected, and this change also allows us to fall back to the user's original preference. 

The `updateWallpaper` function is called when the `HomeFragment` creates its view anyway, so the theme will always match the wallpaper.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -0,0 +1,152 @@
...
+import androidx.compose.ui.res.painterResource
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import org.mozilla.fenix.R
+import org.mozilla.fenix.theme.FirefoxTheme
+import org.mozilla.fenix.wallpapers.Wallpaper
+
+/**
+ * A grid of selectable wallpaper thumbnails.
```
gabrielluong, line 36: Should the description be the same as `WallpaperThumbnails `?

app/src/main/res/layout/fragment_wallpaper_settings.xml
```diff
@@ -0,0 +1,14 @@
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+    - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
```
gabrielluong, line 3: The formatting looks a bit off. I just reviewed a PR that fixes the formatting so I would prefer we don't regress it so quickly.

gabrielluong (approved):

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -0,0 +1,152 @@
...
+ * @param wallpaper The wallpaper to display.
+ * @param isSelected Whether the wallpaper is currently selected.
+ * @param aspectRatio The ratio of height to width of the thumbnail
+ * @param onSelected Action to take when this wallpaper is selected.
+ */
+@Composable
+private fun WallpaperThumbnailItem(
+    wallpaper: Wallpaper,
+    isSelected: Boolean,
+    aspectRatio: Float = 1.1f,
```
gabrielluong, line 99: I have a question about how we come up with the default `1.1f`?

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -0,0 +1,152 @@
...
+ *
+ * @param wallpapers Wallpapers to add to grid
+ * @param selectedWallpaper The currently selected wallpaper
+ * @param onWallpaperSelected Action to take when a new wallpaper is selected.
+ */
+@Composable
+fun WallpaperSettings(
+    wallpapers: List<Wallpaper>,
+    selectedWallpaper: Wallpaper,
+    onWallpaperSelected: (Wallpaper) -> Unit,
```
gabrielluong, line 46:
```suggestion
    onWallpaperSelect: (Wallpaper) -> Unit,
```
I was looking around our codebase and it seems like we prefer the the present tense for the lambda function naming for Jetpack Compose.

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperManager.kt
Amejia481: Thank for finding the issue.  I'm able to reproduce, it looks like at some point we marked on both settings as `true`? Could you help me to clarify how not persisting the setting will address the issue? Are you not going to get flashes between themes if we don't persist?

Amejia481: If we add `settings.shouldFollowDeviceTheme = false` shold fix the issue

Amejia481: Is there any specific reason why we don't want to persist the theme adjustment ?

Amejia481: This approach address the problem, but without persisting the theme setting we will be switching between themes  when the user starts the app, in some cases the user doesn't go directly to home,  could go directly browser screen when it has an opened tab.

Amejia481: I think there is still some value in keeping in sync the setting and the theme. In some cases, as we  are adjusting the theme but no there setting, users could be in some inconsistent setting states, where the customization page says the selected theme is light but the applied one is the dark theme, because of the wallpaper. 
image

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettingsFragment.kt
```diff
@@ -0,0 +1,59 @@
...
+    private val wallpaperManager by lazy {
+        requireComponents.wallpaperManager
+    }
+
+    override fun onCreateView(
+        inflater: LayoutInflater,
+        container: ViewGroup?,
+        savedInstanceState: Bundle?
+    ): View? {
+        _binding = FragmentWallpaperSettingsBinding.inflate(inflater, container, false)
```
MozillaNoah, line 35: Optionally, you might be able to drop the ViewBinding and the XML file for an inline inflation of a ComposeView

gabrielluong: This might be preferable.

Amejia481 (approved):
I left some comments, I think what we have here is in a good shape 👍🏽

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
MatthewTighe: I just did some math based on the designs from UX. They're pretty clearly not quite squares in the design, and using an aspect ratio allows us to dynamically resize based on screen size.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
MatthewTighe: The present tense looks worse to me but makes more sense logically. I'll update these and then we can push from present tense moving forward for consistency. Though I like `onSelectWallpaper` more than I like `onWallpaperSelect`, what do you think?

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
gabrielluong: I guess what is strange to me is that we just didn't specify the width/height (in dp) that is used in the figma. Since the dp should also just take care of of the different displays.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettingsFragment.kt
MatthewTighe: Good call, thanks

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
gabrielluong: `onSelectWallpaper` is fine. I saw `onDismissRequest` in `DropdownMenu`. So, when in doubt, just look at some Compose internals.

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperManager.kt
MatthewTighe: @Amejia481 That's a good point about the home screen not always being the first view that is loaded, which could cause the theme to be out of sync.

One benefit of this implementation is that the theme is still in sync with what the user has selected. It allows us to return to the theme they had previously selected when they switch back to the default wallpaper. 

We could address your concern and keep this behavior by setting the theme in `FenixApplication` based on whether a wallpaper was selected or not. What do you think of that approach? If we go that way, should we do it in a follow-up PR?

Amejia481: I think we could keep the actual behavior and improve it if need it in a follow up 👍🏽 .

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
MatthewTighe: Nope! I'll update that

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -0,0 +1,152 @@
...
+import org.mozilla.fenix.R
+import org.mozilla.fenix.theme.FirefoxTheme
+import org.mozilla.fenix.wallpapers.Wallpaper
+
+/**
+ * A grid of selectable wallpaper thumbnails.
+ *
+ * @param wallpapers Wallpapers to add to grid
+ * @param selectedWallpaper The currently selected wallpaper
+ * @param onWallpaperSelected Action to take when a new wallpaper is selected.
```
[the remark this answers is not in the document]
MatthewTighe: Periods added

app/src/main/res/layout/fragment_wallpaper_settings.xml
MatthewTighe: Removed this file and switched to inflating a `ComposeView` inline

MatthewTighe:
Included some updates from design with the PR feedback. Using a shadow instead of a border for unselected wallpapers now:

image

Issue #22189, opened 2021-10-27
Replace the Firefox word in all strings with a placeholder


Amejia481:
We would like to replace the word "**Firefox**" in all our strings with a string placeholder (only on `/app/src/main/res/values/strings.xml`). For example:

**This is what we have:** 😞 

````xml
    <string name="synced_tabs_no_tabs">You don’t have any tabs open in Firefox on your other devices.</string>
````

**This is what we need:** 😄 
The string above, we want replace the word "**Firefox**" with a placeholder that we will replace in code:

````xml
    <string name="synced_tabs_no_tabs" moz:removedIn="95" tools:ignore="UnusedResources">You don’t have any tabs open in Firefox on your other devices.</string>
    <string name="synced_tabs_no_tabs_2">You don’t have any tabs open in %1$s on your other devices.</string>
````

Replace each reference of the string in code using the string placeholder `app_name`:

```kotlin

        val appName = resources.getString(R.string.app_name)
        description.text = getString(
            R.string.synced_tabs_no_tabs_2, appName
        )
```

From the code above, you can notice:
1. We don't actually update the string `synced_tabs_no_tabs` we created a new string adding a `_#` (a underscore with number, starting on 2), and added the properties `moz:removedIn="95" tools:ignore="UnusedResources"` to `synced_tabs_no_tabs`. We do this, for limitation described [here](https://github.com/mozilla-mobile/fenix/wiki/Remove-strings).
2. In code we replace the place the placeholder with the string `app_name`, we also have to make sure we replace all reference of the old string (`synced_tabs_no_tabs`) with the new string (`synced_tabs_no_tabs_2`) both in code and XML layouts.

For more context see https://github.com/mozilla-l10n/android-l10n/pull/393#discussion_r728314345

grigoryk:
@Amejia481 I'm assuming this is what you meant, but just to clarify:
I think in case of intentionally hardcoded strings, e.g. something like `Firefox Account` or `Firefox Sync`, we should leave them as-is. That is, we don't have a `Firefox Preview Sync` service.

Amejia481:
@grigoryk thanks for clarifying, yeah, you are right. @S10MC2015 as @grigoryk mentioned we want to keep the ones that are referring to specific Firefox services as they are. We only want to update strings that are related to the app's name.

Alexandru2909:
Hi, I would like to work on this issue!

Amejia481:
[...]

PR #23168, opened 2022-01-12
For #22189 - Replaced the Firefox word in all strings with placeholder

Alexandru2909:
Replaced the 'Firefox' word in strings.xml with a placeholder. The placeholders use app_name instead of 'Firefox'. The strings that have been changed do not include phrases like 'Firefox Sync', 'Firefox Preview is now Firefox Nightly' or other intentionally hardcoded Firefox services.

### Pull Request checklist

1. click on Show All Checks,


app/src/main/res/values/strings.xml
```diff
@@ -83,7 +83,9 @@
 
     <!-- Open in App "contextual feature recommendation" (CFR) -->
     <!-- Text for the info message. 'Firefox' intentionally hardcoded here.-->
-    <string name="open_in_app_cfr_info_message">You can set Firefox to automatically open links in apps.</string>
+    <string name="open_in_app_cfr_info_message" moz:RemovedIn="97" tools:ignore="UnusedResources">You can set Firefox to automatically open links in apps.</string>
```
gabrielluong, line 86:
```suggestion
    <string name="open_in_app_cfr_info_message" moz:RemovedIn="98" tools:ignore="UnusedResources">You can set Firefox to automatically open links in apps.</string>
```
Nightly is actually in 98 now.

app/src/main/res/values/strings.xml
```diff
@@ -83,7 +83,9 @@
...
+    <!-- Text for the info message. Uses app_name placeholder instead of 'Firefox'.-->
```
gabrielluong, line 87: We should also specify what %1$s represents

app/src/main/java/org/mozilla/gecko/search/SearchWidgetProvider.kt
```diff
@@ -176,6 +176,11 @@ class SearchWidgetProvider : AppWidgetProvider() {
                 )?.toBitmap()
             )
         }
+        val appName = context.getString(R.string.app_name)
```
gabrielluong, line 179: Let's add a new line before this to visually separate the if statement block of work from this new block of work to set the content description

app/src/main/res/layout/search_widget_extra_small_v2.xml
```diff
@@ -15,7 +15,7 @@
         android:layout_width="50dp"
         android:layout_height="50dp"
         android:layout_gravity="center"
-        android:contentDescription="@string/search_widget_content_description"
+        android:contentDescription="@string/search_widget_content_description_2"
```
gabrielluong, line 18: Shouldn't we remove the content description since we are setting this in `SearchWidgetProvider`? If we decide to remove this, we should go look for the other search_widget_content_description_2 usage.

gabrielluong (approved):
r+ assuming the feedback is addressed.

app/src/main/res/layout/search_widget_extra_small_v2.xml
Alexandru2909: I would rather keep the .xml content descriptions instead of using tools:ignore.

gabrielluong: Okay. I don't think this was a blocker to me.  I saw you removed the content description in another place where we were setting content description code which is why I brought it up. Otherwise, it was more odd to use the string knowing there is a placeholder in it. Feel free to proceed with shipping.

Issue #13336, opened 2020-08-06
FNX-14517 ⁃ Bookmarks should open in existing tab when navigating to bookmarks UI from a page


yoasif:
### What is the user problem or growth opportunity you want to see solved?

Fennec migration, ease of use, user expectation, competitiveness.

### How do you know that this problem exists today? Why is this important?

Existing behavior in Fennec. People asking for it: https://www.reddit.com/r/firefox/comments/i4kmyz/firefox_android_opening_sites_from_bookmarks/

Previously requested: https://github.com/mozilla-mobile/fenix/issues/10837 https://github.com/mozilla-mobile/fenix/issues/10645

"Don't make me think" and a time saver.

Existing behavior in Chrome (where most new users will be switching from) and Firefox desktop. 

### Who will benefit from it?

People who are directly navigating from existing pages to bookmarks no longer have to clean up their old tabs; anyone who is used to the behavior in other Android browsers.

cadeyrn:
The current behaviour is intentional (there is at least one closed issue about this) and I definitely prefer the current behaviour.

yoasif:
@cadeyrn can you help me with a reference to the previous issue? I want to understand the thinking that went into that.

yoasif:
@cadeyrn the best I could find is https://github.com/mozilla-mobile/fenix/issues/3551 but that doesn't really capture the use case of opening a bookmarks when starting navigating from an existing tab, which is what my issue is about. I think that this is worth discussing.

cadeyrn:
>  This behavior is intended right now.

#10645

yoasif:
I think this merits another look, especially since that use case is not called out in any design documentation or stories that I have access to. The best I can find with that continues to be #3551, and there has not been any _discussion_ since then.

person808:
This appears to be the reason for the current behavior: https://github.com/mozilla-mobile/fenix/issues/1012#issuecomment-479608509

yoasif:
@person808 Thanks for looking into that. Today, Library no longer exists, so I think this is still worth another look. 😄

binocry:
so weird when on bookmark page 3 dot menu you can select "open to new tab" so what the point open bookmark to another tab as default?

benediktwinder:
The current behavior is not good for me. It would be great if i click on a bookmark it opens in the same tab, like other browsers.

hotrodwinston:
[...]

Issue #20012, opened 2021-06-14
Make Home screen with an actual tab


Cheap-Skate:
Has bugged me for ages and the new Home and "Jump back in" buttons, while welcome, really accentuate the issues.

Fenix works differently from any other browser I know, 
1 - tapping the "new tab" button doesn't open a new tab, it just takes me to the Home screen, and
2 - tapping a Top Site always opens it in a new tab

1 is weird to me and 2 is annoying, both seem counter-intuitive

The "Jump back in" button accentuates the issue for me:- now I can open the Tabs tray, hit the New Tab FAB (which doesn't actually open a new tab), then I can Jump back in to my previous tab.  Hey, I thought I'd opened a new tab? Very, very confusing

For consistency with other browers and lack of confusion, perhaps
- the New Tab FAB should always open a new tab, increment the tab counter, and move me to the Home screen. Tapping any shortcut on that Home screen should then open in the newly created tab. Jump back in could be re-labelled as "Cancel new tab"  
- tapping Home should not open a new tab or increment the tab counter but should just move me to the Home screen. Tapping any shortcut on that Home screen will open the site in the tab from which the Home screen was accessed.

I don't think this is a new complaint ;) see eg #19930 and probably some otheres

Thanks & keep up good work

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-11037)

Cheap-Skate:
Fenix still seems to have the paradigm where the "Home" page isn't associated with any particular tab, it's [limbo state] from which new tabs (always) get created.

I think that's wildly unorthodox and unhelpful.

Having three buttons (the New Tab shortcut, the New Tab FAB, and the Home button) with different icons on them is a bit of a con, all they really do is take me to the same [limbo state]. They could all just have the same [limbo state] icon on them.

Once I'm in [limbo state] any tap opens a new tab. So X years into the Fenix project, I *still* can't recycle a tab. I really really want to do what I could in Fennec and my other browsers:- tap Home button or URL, tap a Top Site, watch Top Site open in existing tab.  Fenix's Home button still opens everything in a new freakin' tab. Grrrrr.

 I've never come across a browser with such bizarre behavior. I'm close to giving up on Fenix.

[/rant]

ale82to:
I could not agree more.
in fact the new home button is nothing but a joke. 
[...]

Issue #2871, opened 2019-05-27
Bookmarklets doesn't work from bookmarks


LittleAngelwings:
## Steps to reproduce
Save any site to bookmarks
Edit this bookmark and put in there a Javascript as url,  e.x.
javascript:%20(function%20()%20{%20%20%20%20%20%20window.open('https://invidio.us/watch?v='%20+%20document.location.href.replace(/^.+v=(.{11})(.+)?/g,%20'$1'));%20%20})();
Save this edited bookmark
Use it on YouTube for download a video. 

### Expected behavior
The YT Video will be downloaded

### Actual behavior
The url of the bookmarklet will be searched

### Device information
* Android device: Samsung Galaxy Tab SM-T580 and Huawei P smart 
* Fenix version: 1.0.1921 Nightly and Raptor

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-11738)

bobobo1618:
I'm also able to reproduce this.

Annoyingly, typing `javascript:` URLs into the address bar results in a Google search.

mindeffects:
Fun fact 1: When using the Developer Edition of Firefox, everything works as expected. 
Fun fact2: I spent 2 hours preparing a super luxury preview bookmarklet in FFdev which completly failed when the client wanted to use it in his browser (Firefox Regular Edition). :-D

cadeyrn:
@mindeffects There is no "Developer Edition" of Firefox for Android. Are you talking about Firefox for Desktop? Or do you mean the nightly builds of Firefox for Android when you say "Developer Edition"?

mindeffects:
Oh, sorry. I typed the post on my mobile and mighted have stumbled into the wrong thread... Damn you, small screen! 😉
I had my problem with the desktop version of FF for macOS. Sorry.

mdxe:
I have the same issue and most extensions don't work... Why did they even release this as a final version?

angelsl:
How do I help out with this? Is this blocked on some design issue or does someone just need to write it out?

theabbie:
Is it fixed yet? If this is for security reasons atleast allow people to disable it.

mdxe:
> Is it fixed yet? If this is for security reasons at least allow people to disable it.

Mozilla just fired 1/4 of its employees and stopped caring about their core users (the only ones that haven't switched to Chrome)... They need a new CEO.

theabbie:
[...]

PR #23169, opened 2022-01-12
For #13336: Open bookmarks in current tab

czlucius:
### Pull Request checklist

1. click on Show All Checks,

Hi there
These are my fixes to #13336, to open bookmarks in the current tab. I am not sure whether this change is approved, but as the issue is still open, I am submitting a draft PR.
May I ask the Fenix UX team for their approval on this change, or is the bookmarks opening in a new tab as intended? I can work on it to add other things if needed.
Thank you!

Here are the videos of the proposed change:

https://user-images.githubusercontent.com/58442255/149168581-54cbf4ab-b79b-4532-9fbb-9881cc96c4cc.mp4

Bookmarklets also work:

https://user-images.githubusercontent.com/58442255/149168733-c9d7f058-a252-4337-8000-aa88877e1d5d.mp4

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkController.kt
+++ b/app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkController.kt
@@ -77,7 +77,7 @@ class DefaultBookmarkController(
         val flags = EngineSession.LoadUrlFlags.select(EngineSession.LoadUrlFlags.ALLOW_JAVASCRIPT_URL)
         openInNewTabAndShow(
             item.url!!,
-            true,
+            false,
             BrowserDirection.FromBookmarks,
             activity.browsingModeManager.mode,
             flags
--- a/app/src/test/java/org/mozilla/fenix/library/bookmarks/BookmarkControllerTest.kt
+++ b/app/src/test/java/org/mozilla/fenix/library/bookmarks/BookmarkControllerTest.kt
@@ -110,7 +110,7 @@ class BookmarkControllerTest {
     }
 
     @Test
-    fun `handleBookmarkTapped should load the bookmark in a new tab`() {
+    fun `handleBookmarkTapped should load the bookmark in the current tab`() {
         var invokePendingDeletionInvoked = false
         val flags = EngineSession.LoadUrlFlags.select(EngineSession.LoadUrlFlags.ALLOW_JAVASCRIPT_URL)
 
--- a/app/src/test/java/org/mozilla/fenix/library/bookmarks/BookmarkControllerTest.kt
+++ b/app/src/test/java/org/mozilla/fenix/library/bookmarks/BookmarkControllerTest.kt
@@ -124,7 +124,7 @@ class BookmarkControllerTest {
         verify {
             homeActivity.openToBrowserAndLoad(
                 item.url!!,
-                true,
+                false,
                 BrowserDirection.FromBookmarks,
                 flags = flags
             )
@@ -290,7 +290,7 @@ class BookmarkControllerTest {
         verify {
             homeActivity.openToBrowserAndLoad(
                 item.url!!,
-                true,
+                false,
                 BrowserDirection.FromBookmarks,
                 flags = flags
             )


app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkController.kt
```diff
@@ -77,7 +77,7 @@ class DefaultBookmarkController(
         val flags = EngineSession.LoadUrlFlags.select(EngineSession.LoadUrlFlags.ALLOW_JAVASCRIPT_URL)
         openInNewTabAndShow(
             item.url!!,
-            true,
+            false, // See https://github.com/mozilla-mobile/fenix/issues/13336
```
jonalmeida, line 80: No need to add a reference to the issue here. The commit message contains a reference to it already.

```suggestion
            false,
```

jonalmeida (commented):
The change looks good. You need to fix the tests and comment below. 🙂

czlucius:
@jonalmeida 
I've removed the comment already.
The test  that fails is 
```
TEST: handleBookmarkTapped should open the bookmark
```

According to the test at `BookmarkControllerTest`, it checks if a bookmark is opened in a new tab:
https://github.com/mozilla-mobile/fenix/blob/a49f80d5e8da09968c751b5c3937a5f3dcec71fa/app/src/test/java/org/mozilla/fenix/library/bookmarks/BookmarkControllerTest.kt#L112-L132

However, this issue is for the bookmark to open in the current tab, so the test is failing

jonalmeida:
That's correct! The verify was asserting that we always opened in a new tab, and this the behaviour you've changed, so we just need to fix the test to verify the new behaviour. 🙂

czlucius:
Ok, I've corrected the test.

czlucius:
About the new behaviour of opening in current tab, since Fenix doesn't associate the home screen with a physical tab, I found out that this will cause a problem: If the user opens a new tab(clicks New Tab), then subsequently opens the bookmark, the bookmark would open in the previous tab, which may not be the intended action the user wanted to take. How should we fix this side effect?
Thanks.

czlucius:
I've modified the test, but it's still failing

czlucius:
@jonalmeida 
May I ask why is the test failing even when both values (the one in the test, and the one in the BookmarkController class) are `false`? I've tried this out on my machine, this does not happen when both are `true`. Is there something else that I am missing here?
Thanks.

czlucius:
I've fixed the tests already.



> About the new behaviour of opening in the current tab, since Fenix doesn't associate the home screen with a physical tab, I found out that this will cause a problem: If the user opens a new tab(clicks New Tab), then subsequently opens the bookmark, the bookmark would open in the previous tab, which may not be the intended action the user wanted to take. How should we fix this side effect? (is there a method to find out whether the user is on the home screen page of a new tab? - there could be other instances where the home screen is launched without the intention to open a new tab, e.g. the home button) Thanks.

What do we do about this?


> handleBookmarkTapped should load the bookmark in a new tab

Should I also change the test name to reflect the new behavior?

Thank you.

maverick74:
@jonalmeida @czlucius isn't there a way to, for example, know if the home button is visible?
If we could know, it could be used to either load the bookmark on a new tab or in the present tab...

czlucius:
@jonalmeida what do I do to continue from here?

jonalmeida:
Sorry, I lost track of the message thread!

> I've fixed the tests already.

🎉 

> > About the new behaviour of opening in the current tab, since Fenix doesn't associate the home screen with a physical tab, I found out that this will cause a problem: If the user opens a new tab(clicks New Tab), then subsequently opens the bookmark, the bookmark would open in the previous tab, which may not be the intended action the user wanted to take. How should we fix this side effect? (is there a method to find out whether the user is on the home screen page of a new tab? - there could be other instances where the home screen is launched without the intention to open a new tab, e.g. the home button) Thanks.
> 
> What do we do about this?

Interesting. There is no concept of a "new tab" in Fenix, it's just the home screen. When the bookmark is opened, does it switch to that tab that it loads in? If so, I think that would be acceptable behaviour for now given how Fenix currently works. We can also discuss this further with UX in a separate issue as well.

> > handleBookmarkTapped should load the bookmark in a new tab
> 
> Should I also change the test name to reflect the new behavior?

Yes please! The name of the test should reflect the behaviour of it. If we're changing the behaviour, someone else coming back to this code in the future will be very confused why it says something and does something else. 🙂

czlucius:
Ok 👍

czlucius:
> Interesting. There is no concept of a "new tab" in Fenix, it's just the home screen. When the bookmark is opened, does it switch to that tab that it loads in? If so, I think that would be acceptable behaviour for now given how Fenix currently works. We can also discuss this further with UX in a separate issue as well.

Your comment got me to think about if the user has 0 tabs(starting from a clean state), then this behaviour will attempt to open the bookmark in a nonexistent current tab, which may result in a crash or no-op. I shall go test this behaviour later.

czlucius:
> > Interesting. There is no concept of a "new tab" in Fenix, it's just the home screen. When the bookmark is opened, does it switch to that tab that it loads in? If so, I think that would be acceptable behaviour for now given how Fenix currently works. We can also discuss this further with UX in a separate issue as well.
> 
> Your comment got me to think about if the user has 0 tabs(starting from a clean state), then this behaviour will attempt to open the bookmark in a nonexistent current tab, which may result in a crash or no-op. I shall go test this behaviour later.

Ironically, the code still works well when I open a bookmark when number of tabs is 0. I guess `openToBrowserAndLoad` already has edge case that settled

jonalmeida:
> Your comment got me to think about if the user has 0 tabs(starting from a clean state), then this behaviour will attempt to open the bookmark in a nonexistent current tab, which may result in a crash or no-op. I shall go test this behaviour later.

This is good train of thought to consider and even better that it is already handled by the `openToBrowserAndLoad` method.

jonalmeida:
@czlucius I tried out the patch and it looks good! However there is one more entry point into bookmarks from the home screen's "Recent bookmarks". I wonder if you're interested in also fixing that too? Otherwise, we can file a separate bug for it.

czlucius:
I think the recent bookmarks from the home screen is somewhat related to this issue: https://github.com/mozilla-mobile/fenix/issues/20012

If bookmarks were to open in the current tab, but top sites still open in a new tab, the behaviour will be quite confusing.

Furthermore, the home screen is also shown when the user explicitly clicks "New tab", and opening the recent bookmark in the current tab would not be the intended behaviour in this case.
Perhaps another issue can be filed about this, and I can look into it?

Thanks.

czlucius:
If the fix is good as it is now, you can merge it.

The minor flaw that I described earlier is illustrated in this recording:

https://user-images.githubusercontent.com/58442255/151705088-f332ad9e-c237-4b40-b2df-e9ca5cab4200.mp4

czlucius:
@jonalmeida 
If the fix is good as it is now, you can merge it(and maybe open a new issue for the home screen one - maybe for top sites, bookmarks and also collections?)

The minor flaw that I described earlier is illustrated in this recording:

https://user-images.githubusercontent.com/58442255/151705088-f332ad9e-c237-4b40-b2df-e9ca5cab4200.mp4

Thanks.

Cheap-Skate:
> @jonalmeida If the fix is good as it is now, you can merge it(and maybe open a new issue for the home screen one - maybe for **top sites, bookmarks and also collections?**)

Please open this issue & Thanks for your efforts so far 👍!!

jonalmeida (approved):

czlucius:
https://github.com/mozilla-mobile/fenix/issues/2871 is related as well, and since this is fixed, that issue may be fixed too

czlucius:
@jonalmeida 
The issue that I stated above is being reported by some users:
https://github.com/mozilla-mobile/fenix/issues/23595

Is there a method that returns whether the home screen is opened? We could open in new or current tab based on this value.

Thanks!

Cheap-Skate:
#23659 another one reporting the new behavior as a bug

czlucius:
@jonalmeida this seems to be causing problems, as seen in https://github.com/mozilla-mobile/fenix/issues/23595, https://github.com/mozilla-mobile/fenix/issues/23659, and https://github.com/mozilla-mobile/fenix/issues/23746. I think this should be looked into(before this gets into release).

maverick74:
@czlucius i think you missed bug #20012

czlucius:
No, I was talking about the problems caused by changing the behavior to open bookmarks in the current tab.

maverick74:
Oh! My mistake then! Sorry.
But, i think 20012 should be addressed (after the ones you mentioned) as well to avoid more confusion with this change

Issue #23144, opened 2022-01-11
111% regression in COLD start MAIN time to first frame from 1/5-1/7


mcomella:
<img width="450" alt="image" src="https://user-images.githubusercontent.com/759372/148855499-0587f5c6-ba5a-43e3-8da2-d75757280bfe.png">

We see a recent regression in COLD start MAIN time to first frame:
- 1/5: 1.474s
- 1/6: 2.875s
- 1/7: 3.113s (and it stabilizes)

From **1/5-1/7 is a regression of 111.2% or 1639ms.** Let's investigate to find out the cause.

I took a profile with `Debug` on Jan. 10 upstream/main: https://share.firefox.dev/3HPYaer A profile from the Firefox Profiler would be preferred because of the markers (i.e. shows when first frame is drawn) but that's blocked on crashes: BZ-1749418.

It could be that the time to visual completeness has not regressed and the time to first frame is much closer to it, i.e. the time to fully functional has not regressed. However, this is likely to be a perceived performance issue for users.

mcomella:
I took start up profiles with the Firefox Profiler. Here they are trimmed to the first frame (onPreDraw):
- Nightly 1/5 (1s): https://share.firefox.dev/3zOY8Ra
- Nightly 1/10 (4.5s): https://share.firefox.dev/3nhRzBP

Curiously, they look very similar (i.e. similar proportions to the calls): it's just that in the later Nightly, everything seems to take longer. 🤔 This reminds me of what we saw with https://github.com/mozilla-mobile/fenix/issues/18753. Other possibilities could be increased CPU/disk/memory contention or https://github.com/mozilla-mobile/fenix/issues/17703.

mcomella:
I was able to reproduce this problem locally so worst case we can bisect:
- Nightly 1/5 32bf020b7 (1s): https://share.firefox.dev/3qjhJWE
- Nightly 1/10 7582bce89 (2.3s): https://share.firefox.dev/3Fz5gmd (and I was able to reproduce this behavior again)

The 1/10 local profile looks different from the Nightly one. There is an obviously longer layout call before the first frame is drawn. It looks like we're synchronously handling compose content whereas we were not doing so before:
- before: https://share.firefox.dev/3K92oj6
- after: https://share.firefox.dev/3tk0C95

[...]

PR #23172, opened 2022-01-13
for #23144: only recreate activity when logo is clicked

MatthewTighe:
By only adjusting the theme when the logo is clicked we should be able to avoid perf regressions on start and still update the theme with newly selected wallpapers.


app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperManager.kt
```diff
@@ -55,7 +54,6 @@ class WallpaperManager(private val settings: Settings) {
 
         if (AppCompatDelegate.getDefaultNightMode() != mode) {
             AppCompatDelegate.setDefaultNightMode(mode)
-            context.asActivity()?.recreate()
```
Amejia481, line 58: Here we should only recreating the activity when the theme changed from the previous one, if it didn't change we shouldn't recreate

Issue #12717, opened 2020-07-19
Add a search button to the "Bookmarks" screen


ghost:
Add a search function to the Bookmarks screen. Right now it is _really_ hard to manage bookmarks or find them. Let's say if I want to delete a bookmark then the only way to do so is to search that bookmark manually among hundreds of other bookmarks, this isn't practical! The only other way to do so is to type the bookmarks address in the Fenix's address bar then go to the bookmarked website and then tap the remove bookmark icon by tapping 3-dot menu. This is something that isn't practical if somebody wants to remove multiple bookmarks. And if someone wants to edit or move the bookmarks to different folders then it can only be done by manually finding them manually and moving them one-by-one. This also not practical.
Copying the bookmarked address' link is also something cannot be done easily without a search function.

These are all the reasons that make it really harder to manage, find and organize the bookmarks. And I think a search function is a must. This is already available in Chrome browser for example.

### Chrome browser's bookmarks search function: 
![Screenshot](https://user-images.githubusercontent.com/68282543/87878959-2ab6dc00-c9d7-11ea-9947-333bebf2579e.png)UNITO-UNDERSCORE!20200719-211658!
![Screenshot](https://user-images.githubusercontent.com/68282543/87878963-2f7b9000-c9d7-11ea-93a7-17c7c4e2b2ed.png)UNITO-UNDERSCORE!20200719-211711!

This will add to the history search function asked in #10046 but I can live with history search function not available for a moment as I already have the  suggestions appearing in address bar when I type something but for bookmarks it's just not about searching them but also about managing, editing, copying the bookmarked address' link, etc.

Thank you. :grinning:

Poopooracoocoo:
would be great if it let you show it in its folder if it searches across folders.

Abhishek8394:
I'd like to work on this. @sflorean is the screenshot the desired UX?

eliserichards:
Hi @Abhishek8394! Since this is a feature request let's run it by the UX team to see if they want this or not 👍 If not, I'll find another issue you can work on :)

Abhishek8394:
Just pinging to see if this issue is  still of interest or the review can be expedited. I miss this feature everyday 😿

Mugurell:
@Abhishek8394 I support for searching bookmarks in https://github.com/mozilla-mobile/fenix/pull/23176 with a generic way - `configureSearchViewInMenu` to configure the search menu in any fragments. 
[...]

PR #23176, opened 2022-01-13
For #12717 - Bookmark search

Mugurell:
Allows for an in-memory search of all bookmarks flattened and then the default
folders (like Bookmarks Toolbar, etc) removed.
Search will only start and the current bookmarks list will be updated to
the queried results after the user first enters a valid query (not blank).
Search will end when the SearchView is closed. (Deleting all query characters
will not end the search mode).

https://user-images.githubusercontent.com/11428869/149887801-b1e3c30f-e3b4-41e8-bbb9-ccfb2d709c0a.mp4

cc @topotropic 

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkController.kt
+++ b/app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkController.kt
@@ -7,6 +7,7 @@ package org.mozilla.fenix.library.bookmarks
 import android.content.ClipData
 import android.content.ClipboardManager
 import android.content.res.Resources
+import androidx.annotation.VisibleForTesting
 import androidx.navigation.NavController
 import androidx.navigation.NavDirections
 import kotlinx.coroutines.CoroutineScope
@@ -15,6 +16,7 @@ import mozilla.appservices.places.BookmarkRoot
 import mozilla.components.concept.engine.EngineSession
 import mozilla.components.concept.engine.prompt.ShareData
 import mozilla.components.concept.storage.BookmarkNode
+import mozilla.components.concept.storage.BookmarksStorage
 import mozilla.components.feature.tabs.TabsUseCases
 import mozilla.components.service.fxa.sync.SyncReason
 import org.mozilla.fenix.BrowserDirection
@@ -47,6 +49,8 @@ interface BookmarkController {
     fun handleBookmarkFolderDeletion(nodes: Set<BookmarkNode>)
     fun handleRequestSync()
     fun handleBackPressed()
+    fun handleQuery(previousQuery: String, newQuery: String)
+    fun handleSearchEnded()
 }
 
 @Suppress("TooManyFunctions")
@@ -59,6 +63,7 @@ class DefaultBookmarkController(
     private val sharedViewModel: BookmarksSharedViewModel,
     private val tabsUseCases: TabsUseCases?,
     private val loadBookmarkNode: suspend (String) -> BookmarkNode?,
+    private val bookmarksStorage: BookmarksStorage?,
     private val showSnackbar: (String) -> Unit,
     private val deleteBookmarkNodes: (Set<BookmarkNode>, Event) -> Unit,
     private val deleteBookmarkFolder: (Set<BookmarkNode>) -> Unit,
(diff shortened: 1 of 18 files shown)


Poopooracoocoo:
Please don't forget a menu item on search results to show the bookmark in its folder!

grigoryk:
Nice, @Mugurell ! Started reading the code & playing with this locally. Will try to give some specific feedback tomorrow. Generally it seems to work well!

The main thing that stood out to me right away: I don't think we should switch display to the flattened nodes after toggling the search mode. I.e. in the UI it's really odd to have everything change up on you just because you opened the search bar. I think in the final version of this, the displayed list should not change at all after pressing the search button, and we'll just display filtered items as the user types their query.

app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkController.kt
```diff
@@ -181,6 +186,37 @@ class DefaultBookmarkController(
...
+                ?.flattenAllNodes()
+                // Important to strip all default nodes like the desktop toolbar after all children are flattened.
+                ?.filterNot { node -> BookmarkRoot.values().any { it.id == node.guid } }
+
+            store.dispatch(BookmarkFragmentAction.SearchStarted(allUserNodes ?: emptyList()))
+        }
+    }
+
+    override fun handleSearchEnded() {
+        scope.launch {
```
gabrielluong, line 203: How come this needs a scope.launch?

app/src/main/res/values/strings.xml
```diff
@@ -897,6 +897,8 @@
     <string name="bookmark_delete_multiple_folders_confirmation_dialog">%s will delete the selected items.</string>
     <!-- Snackbar title shown after a folder has been deleted. This first parameter is the name of the deleted folder -->
     <string name="bookmark_delete_folder_snackbar" moz:removedIn="96" tools:ignore="UnusedResources">Deleted %1$s</string>
+    <!-- Placeholder text shown as a hint in the bookmarks search bar before the user enters text -->
+    <string name="bookmark_search_hint">Search bookmarks</string>
```
gabrielluong, line 901: Not to discourage the ongoing work, but I always thought this was still in the UX design phase. I am mostly concern that we are getting ahead of ourselves and not aligning with the Product/UX expectations.

app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkController.kt
Mugurell: Leftover from a previous approach. Will remove. Thanks!

app/src/main/res/values/strings.xml
Mugurell: This was next in our backlog on Jira.
cc @amedyne 
Regarding the string value I went with a similar hint to the one used on other screens
cc @topotropic

Mugurell:
> The main thing that stood out to me right away: I don't think we should switch display to the flattened nodes after toggling the search mode. I.e. in the UI it's really odd to have everything change up on you just because you opened the search bar. I think in the final version of this, the displayed list should not change at all after pressing the search button, and we'll just display filtered items as the user types their query.

Can do exactly as Chrome and only switch to the flattened list once user started typing and keep this list even for when the user deletes the entire query, until the search view is closed.

app/src/main/res/values/strings.xml
gabrielluong: Totally understandable. I only brought this up because I asked in a sprint planning back in December if we had designs for bookmark search, but was pointed out that this is actually in design even though it's in the backlog. I think the issue is the backlog is a shard Eng/UX backlog which leaves room for misinterpretation for what should be worked on. In this case, you'll see a "Need UX" label in the Jira ticket (this was also shown to me when I asked during the meeting).

Mugurell:
> > The main thing that stood out to me right away: I don't think we should switch display to the flattened nodes after toggling the search mode. I.e. in the UI it's really odd to have everything change up on you just because you opened the search bar. I think in the final version of this, the displayed list should not change at all after pressing the search button, and we'll just display filtered items as the user types their query.
> 
> Can do exactly as Chrome and only switch to the flattened list once user started typing and keep this list even for when the user deletes the entire query, until the search view is closed.

Updated the patch to follow this scenario.

PR #23196, opened 2022-01-14
 closes #23220: show snackbar on wallpaper selection

### LINKED ISSUE #23220: Show a snackbar when a new wallpaper is selected on the settings screen.
When a user selects a different wallpaper in the wallpaper settings screen, a snackbar should be displayed. It should have the following properties:

- Message that says: "Wallpaper updated!"
- Action that is label "VIEW" and navigates to the home screen so that the wallpaper can be viewed.

MatthewTighe:
image

image

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
@@ -12,43 +12,122 @@ import androidx.compose.foundation.clickable
 import androidx.compose.foundation.layout.aspectRatio
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.heightIn
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.lazy.GridCells
 import androidx.compose.foundation.lazy.LazyVerticalGrid
 import androidx.compose.foundation.lazy.items
 import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.material.Scaffold
+import androidx.compose.material.Snackbar
+import androidx.compose.material.SnackbarDuration
+import androidx.compose.material.SnackbarHost
 import androidx.compose.material.Surface
+import androidx.compose.material.Text
+import androidx.compose.material.TextButton
+import androidx.compose.material.rememberScaffoldState
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.rememberCoroutineScope
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.layout.ContentScale
 import androidx.compose.ui.res.painterResource
 import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.text.font.Font
+import androidx.compose.ui.text.font.FontFamily
+import androidx.compose.ui.text.style.TextAlign
+import androidx.compose.ui.text.style.TextOverflow
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+import kotlinx.coroutines.launch
 import org.mozilla.fenix.R
 import org.mozilla.fenix.theme.FirefoxTheme
 import org.mozilla.fenix.wallpapers.Wallpaper
+import java.util.Locale
 
 /**
- * The screen for controlling settings around Wallpapers.
+ * The screen for controlling settings around Wallpapers. When a new wallpaper is selected,
+ * a snackbar will be displayed.
  *
  * @param wallpapers Wallpapers to add to grid.
  * @param selectedWallpaper The currently selected wallpaper.
- * @param onSelectWallpaper Action to take when a new wallpaper is selected.
+ * @param onSelectWallpaper Callback for when a new wallpaper is selected.
+ * @param onViewWallpaper Callback for when the view action is clicked from snackbar.
  */
 @Composable
 fun WallpaperSettings(
     wallpapers: List<Wallpaper>,
     selectedWallpaper: Wallpaper,
     onSelectWallpaper: (Wallpaper) -> Unit,
+    onViewWallpaper: () -> Unit,
 ) {
-    Surface(color = FirefoxTheme.colors.layer2) {
-        WallpaperThumbnails(
-            wallpapers = wallpapers,
-            selectedWallpaper = selectedWallpaper,
-            onSelectWallpaper = onSelectWallpaper
-        )
+    val coroutineScope = rememberCoroutineScope()
+    val scaffoldState = rememberScaffoldState()
+
+    Scaffold(
+        backgroundColor = FirefoxTheme.colors.layer1,
+        scaffoldState = scaffoldState,
+        snackbarHost = { hostState ->
+            SnackbarHost(hostState = hostState) {
+                WallpaperSnackbar(onViewWallpaper)
+            }
+        }
+    ) {
+        Surface(color = FirefoxTheme.colors.layer2) {
+            WallpaperThumbnails(
+                wallpapers = wallpapers,
+                selectedWallpaper = selectedWallpaper,
+                onSelectWallpaper = { updatedWallpaper ->
+                    coroutineScope.launch {
+                        scaffoldState.snackbarHostState.showSnackbar(
+                            message = "", // overwritten by WallpaperSnackbar
+                            duration = SnackbarDuration.Short
+                        )
+                    }
+                    onSelectWallpaper(updatedWallpaper)
+                },
+            )
+        }
     }
+}
+
+@Composable
+private fun WallpaperSnackbar(
+    onViewWallpaper: () -> Unit,
+) {
+    Snackbar(
+        modifier = Modifier
+            .padding(8.dp)
+            .heightIn(min = 48.dp),
+        backgroundColor = FirefoxTheme.colors.actionPrimary,
+        content = {
+            Text(
+                modifier = Modifier.padding(start = 16.dp, end = 16.dp, top = 8.dp, bottom = 8.dp),
+                text = stringResource(R.string.wallpaper_updated_snackbar_message),
+                textAlign = TextAlign.Start,
+                color = FirefoxTheme.colors.textInverted,
+                fontFamily = FontFamily(Font(R.font.metropolis_semibold)),
+                fontSize = 18.sp,
+                overflow = TextOverflow.Ellipsis,
+                maxLines = 2
+            )
+        },
+        action = {
+            TextButton(
+                modifier = Modifier.padding(start = 16.dp, end = 16.dp, top = 8.dp, bottom = 8.dp),
+                onClick = onViewWallpaper,
+            ) {
+                Text(
+                    text = stringResource(R.string.wallpaper_updated_snackbar_action).uppercase(
+                        Locale.getDefault()
+                    ),
+                    color = FirefoxTheme.colors.textInverted,
+                    fontFamily = FontFamily(Font(R.font.metropolis_medium)),
+                    fontSize = 14.sp,
+                )
+            }
+        },
+    )
 }
 
 /**
@@ -134,9 +213,20 @@ private fun WallpaperThumbnailItem(
 @Preview
 @Composable
 private fun WallpaperThumbnailsPreview() {
-    WallpaperSettings(
-        wallpapers = Wallpaper.values().toList(),
-        onSelectWallpaper = {},
-        selectedWallpaper = Wallpaper.NONE
-    )
+    FirefoxTheme {
+        WallpaperSettings(
+            wallpapers = Wallpaper.values().toList(),
+            selectedWallpaper = Wallpaper.NONE,
+            onSelectWallpaper = {},
+            onViewWallpaper = {},
+        )
+    }
+}
+
+@Preview
+@Composable
+private fun WallpaperSnackbarPreview() {
+    FirefoxTheme {
+        WallpaperSnackbar(onViewWallpaper = {})
+    }
 }
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -458,6 +458,10 @@
     <!-- Wallpapers -->
     <!-- Content description for various wallpapers. The first parameter is the name of the wallpaper -->
     <string name="wallpapers_item_name_content_description">Wallpaper Item: %1$s</string>
+    <!-- Snackbar message for when wallpaper is selected -->
+    <string name="wallpaper_updated_snackbar_message">Wallpaper updated!</string>
+    <!-- Snackbar label for action to view selected wallpaper -->
+    <string name="wallpaper_updated_snackbar_action">View</string>
 
     <!-- Add-on Installation from AMO-->
     <!-- Error displayed when user attempts to install an add-on from AMO (addons.mozilla.org) that is not supported -->
--- a/app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettingsFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettingsFragment.kt
@@ -15,6 +15,8 @@ import androidx.compose.runtime.setValue
 import androidx.compose.ui.platform.ComposeView
 import androidx.compose.ui.platform.ViewCompositionStrategy
 import androidx.fragment.app.Fragment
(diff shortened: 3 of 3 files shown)


app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -17,37 +17,79 @@ import androidx.compose.foundation.lazy.GridCells
...
-    Surface(color = FirefoxTheme.colors.layer2) {
-        WallpaperThumbnails(
-            wallpapers = wallpapers,
-            selectedWallpaper = selectedWallpaper,
-            onSelectWallpaper = onSelectWallpaper
-        )
+    val coroutineScope = rememberCoroutineScope()
+    val scaffoldState = rememberScaffoldState()
+    val snackbarMessage = stringResource(id = R.string.wallpaper_updated_snackbar_message)
+    val snackbarActionLabel = stringResource(id = R.string.wallpaper_updated_snackbar_action)
```
gabrielluong, line 59: I would just inline these since the variable isn't used more than once.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -17,37 +17,79 @@ import androidx.compose.foundation.lazy.GridCells
...
+    Scaffold(
```
gabrielluong, line 60: nit: Add a new line to visually separate the variable the declaration and the rendered compose component.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -17,37 +17,79 @@ import androidx.compose.foundation.lazy.GridCells
...
+        backgroundColor = FirefoxTheme.colors.layer1,
+        scaffoldState = scaffoldState,
+        snackbarHost = { hostState ->
+            SnackbarHost(hostState = hostState) { data ->
+                Snackbar(
```
gabrielluong, line 65: **Blocking Change:** Looking at the screenshots, this doesn't seem to be styled correctly to our existing Fenix Snackbar (seems to be missing the font family at least). I am blocking on this to avoid shipping a Snackbar that doesn't conform to the existing design spec.

I think I also want to double check the colors used here in a future review.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
MatthewTighe: That was my thought as well, but `stringResource` can't be invoked outside of a `Composable` context, which the `launch`ed scope below is not. I think the alternative would be to capture the `LocalContext` and use it to resolve the strings, but that would lengthen the expression significantly.  We could add an extension function? It looks like there's already a `Context.getStringWithArgSafe` so there would be precedent to add a `Context.getString`

gabrielluong: Makes sense, I saw something similar in another PR. I wouldn't make the suggested changes in this PR. I think it's fine as-is for the purpose of this PR and we can revisit this in a followup if we feel strongly about it.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
MatthewTighe: Updated the font family along with several other style alignments to the existing `fenix_snackbar` layout and [designs](https://www.figma.com/file/TebIgESfWmQkMcEmwNaYZl/Android?node-id=136%3A40).

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
MatthewTighe: This was resolved as a by-product of the font family changes you requested. I think we should consider creating a more generalized `FenixSnackbar` in a follow-up patch.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
MatthewTighe: Added

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
gabrielluong: Yep, I was thinking the same.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
MatthewTighe: image

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -12,45 +12,119 @@ import androidx.compose.foundation.clickable
...
+                    }
+                    onSelectWallpaper(updatedWallpaper)
+                },
+            )
+        }
     }
 }
 
+@Composable
+private fun WallpaperSnackbar(
```
gabrielluong, line 93: Is it possible to write a @preview for this?

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -12,45 +12,119 @@ import androidx.compose.foundation.clickable
...
+    onViewWallpaper: () -> Unit,
+) {
+    Snackbar(
+        modifier = Modifier
+            .padding(8.dp)
+            .heightIn(min = 48.dp),
+        backgroundColor = FirefoxTheme.colors.actionPrimary,
+        content = {
+            Text(
+                modifier = Modifier.padding(start = 16.dp, top = 8.dp, bottom = 8.dp),
```
gabrielluong, line 103: I think we are missing the 16dp end padding. Normally, I would expect it to be balanced with a value for each of start/end.

gabrielluong: Can I also assume we are grabbing the values from https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/res/layout/fenix_snackbar.xml?

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
MatthewTighe: Added

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
MatthewTighe: Yeah, that's the layout I was looking at. Thanks for linking it. I added `end` padding to the text and `start` padding to the action. The intention was to replace the padding defined in the `ConstraintLayout` of the original, but I think this should work too

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -12,45 +12,119 @@ import androidx.compose.foundation.clickable
...
+                WallpaperSnackbar(onViewWallpaper)
+            }
+        }
+    ) {
+        Surface(color = FirefoxTheme.colors.layer2) {
+            WallpaperThumbnails(
+                wallpapers = wallpapers,
+                selectedWallpaper = selectedWallpaper,
+                onSelectWallpaper = { updatedWallpaper ->
+                    coroutineScope.launch {
```
gabrielluong, line 79: is this `coroutineScope.launch {}` neded?

app/src/main/res/values/strings.xml
```diff
@@ -458,6 +458,10 @@
     <!-- Wallpapers -->
     <!-- Content description for various wallpapers. The first parameter is the name of the wallpaper -->
     <string name="wallpapers_item_name_content_description">Wallpaper Item: %1$s</string>
+    <!-- Snackbar message for when wallpaper is selected -->
+    <string name="wallpaper_updated_snackbar_message">Wallpaper updated!</string>
+    <!-- Snackbar label for action to view selected wallpaper -->
+    <string name="wallpaper_updated_snackbar_action">VIEW</string>
```
gabrielluong, line 464: I am wondering if we already have an existing `VIEW` string for the snackbar that can be used here?

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -12,45 +12,119 @@ import androidx.compose.foundation.clickable
...
+        snackbarHost = { hostState ->
+            SnackbarHost(hostState = hostState) {
+                WallpaperSnackbar(onViewWallpaper)
+            }
+        }
+    ) {
+        Surface(color = FirefoxTheme.colors.layer2) {
+            WallpaperThumbnails(
+                wallpapers = wallpapers,
+                selectedWallpaper = selectedWallpaper,
```
gabrielluong, line 77: This should be consider a followup or future consideration, but one way to avoid passing the `sellectedWallpaper` would be to have `isSelected` state in `Wallpaper`.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
gabrielluong: Nvm, I see `showSnackbar ` is a suspend function

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -12,45 +12,119 @@ import androidx.compose.foundation.clickable
...
+            .heightIn(min = 48.dp),
+        backgroundColor = FirefoxTheme.colors.actionPrimary,
+        content = {
+            Text(
+                modifier = Modifier.padding(start = 16.dp, end = 16.dp, top = 8.dp, bottom = 8.dp),
+                text = stringResource(id = R.string.wallpaper_updated_snackbar_message),
+                textAlign = TextAlign.Start,
+                color = FirefoxTheme.colors.textInverted,
+                fontFamily = FontFamily(Font(R.font.metropolis_semibold)),
+                fontSize = 18.sp,
```
gabrielluong, line 108: Can you also add 

```
overflow = TextOverflow.Ellipsis,
maxLines = 2
```

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -12,45 +12,119 @@ import androidx.compose.foundation.clickable
...
+    Snackbar(
+        modifier = Modifier
+            .padding(8.dp)
+            .heightIn(min = 48.dp),
+        backgroundColor = FirefoxTheme.colors.actionPrimary,
+        content = {
+            Text(
+                modifier = Modifier.padding(start = 16.dp, end = 16.dp, top = 8.dp, bottom = 8.dp),
+                text = stringResource(id = R.string.wallpaper_updated_snackbar_message),
+                textAlign = TextAlign.Start,
```
gabrielluong, line 105: I am not quite sure if `textAlign` does anything since the 2 text aren't in the same box.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -12,45 +12,119 @@ import androidx.compose.foundation.clickable
...
+    onViewWallpaper: () -> Unit,
+) {
+    Snackbar(
+        modifier = Modifier
+            .padding(8.dp)
+            .heightIn(min = 48.dp),
+        backgroundColor = FirefoxTheme.colors.actionPrimary,
+        content = {
+            Text(
+                modifier = Modifier.padding(start = 16.dp, end = 16.dp, top = 8.dp, bottom = 8.dp),
```
gabrielluong, line 103: Looking at the @Preview, it seems like the Snackbar has some built-in padding already so we might not actually need this, but let's look at this in a followup.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -12,45 +12,119 @@ import androidx.compose.foundation.clickable
...
+                fontSize = 18.sp,
+            )
+        },
+        action = {
+            TextButton(
+                modifier = Modifier.padding(start = 16.dp, end = 16.dp, top = 8.dp, bottom = 8.dp),
+                onClick = onViewWallpaper,
+            ) {
+                Text(
+                    text = stringResource(id = R.string.wallpaper_updated_snackbar_action),
```
gabrielluong, line 117: Let's drop the named parameter `id`. I know I like to see named parameters, but given that it is a bit obvious what goes into stringResource, and how we didn't do this for the other stringResource (let's be consistent). Named parameters do help when the code isn't obvious without the help an IDE such as `doSomething(true, false, true, 1, 0)`.

gabrielluong (approved):

gabrielluong:
Could you also file an issue in the future, it would good to have a 1:1 github to jira so that things are linked correctly in the event anyone needs to find the PR to something and has additional context through an issue or jira description. Plus, it will help with syncing the completion of work on Jira.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
MatthewTighe: Yeah, `SnackbarHostState::showSnackbar` is a suspending function

app/src/main/res/values/strings.xml
MatthewTighe: We have this one which I could make uppercase at runtime:
```
<!-- Collections -->
...
<!-- Snackbar action to view the collection the user just created or updated -->
    <string name="create_collection_view">View</string>
```

I could move it under `<!-- Snackbar -->` and make the comment more general, but I worry that context could be lost.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
MatthewTighe: The text alignment refers to how it aligns relative to itself, if it's long enough to be a paragraph. Given that we're adding `overflow` and `maxLines` it seems like we should keep it for at least the message. I can remove it from the action label, however.

MatthewTighe:
> Could you also file an issue in the future, it would good to have a 1:1 github to jira so that things are linked correctly in the event anyone needs to find the PR to something and has additional context through an issue or jira description. Plus, it will help with syncing the completion of work on Jira.

Definitely! There was no matching Jira ticket, since this was an iterative design update. Should I create both in the future? Seems a little bookkeep-y to me, but happy to do it if that's the usual workflow

app/src/main/res/values/strings.xml
gabrielluong: Ok, let's not do that and just keep it as-is. However, now I am think the string should be changed to "View" and we add  should add the equivalent of `android:textAllCaps="true"` to the Compose Text().

gabrielluong:
> > Could you also file an issue in the future, it would good to have a 1:1 github to jira so that things are linked correctly in the event anyone needs to find the PR to something and has additional context through an issue or jira description. Plus, it will help with syncing the completion of work on Jira.
> 
> Definitely! There was no matching Jira ticket, since this was an iterative design update. Should I create both in the future? Seems a little bookkeep-y to me, but happy to do it if that's the usual workflow

You only need to file a GitHub issue in general, and just let the data-sync add the Jira equivalent.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
MatthewTighe: I added #23221 to track that work.

gabrielluong: Cool, I was thinking of taking a stab at it if it's good with you.

gabrielluong:
Latest changes look good to me.

PR #23239, opened 2022-01-17
Set a minimum number of sites a search group should contain.

mcarare:
### Pull Request checklist

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/components/history/PagedHistoryProvider.kt
```diff
@@ -133,6 +135,14 @@ class DefaultPagedHistoryProvider(
                                 items = items.map { it.toHistoryDBMetadata() }
                             )
                         }
+                        .filter {
```
csadilek, line 138: To land this we also need to apply this logic to history search groups under "recently visited." Otherwise the logic is inconsistent and we'd have groups in recently visited with a single item, but once the user hits "show all" the groups wouldn't exist in the history view.

csadilek: (Have groups and sizes consistent was important feedback form user testing.)

mcarare: Yes, I have asked about that in the Jira ticket. This is just a draft, while waiting for changes from https://github.com/mozilla-mobile/fenix/pull/23257 to land

rocketsroger (approved):
Looks good.  Just wondering, should we also change `SEARCH_TERM_TAB_GROUPS_MIN_SIZE` in `SearchTermTabGroupMiddleware` to use `SEARCH_GROUP_MINIMUM_SITES`? or do you think it's better if we keep them separated?

mcarare:
@rocketsroger I think we should keep them separate. 
For one, the other `val` is related to tab groups and we might want to keep things separate. Secondly, this here is related to the experiment, and so we should look at it as being subject to change.

rocketsroger:
> @rocketsroger I think we should keep them separate. For one, the other `val` is related to tab groups and we might want to keep things separate. Secondly, this here is related to the experiment, and so we should look at it as being subject to change.

Understood.  Thanks!

PR #23257, opened 2022-01-17
Close #23250: Remove duplicate sites within a time period in history

### LINKED ISSUE #23250: Remove duplicate sites within a time period in history
Description

As a user when I am viewing the history page I see only result for a unique page (URL) for each time period. I want this because it reduces clutter and allows me to identify important sites quickly.
Acceptance Criteria

- When a user has multiple visits for the same site in a time period that only the most recent visit is shown
- This is per full URL
- This page should show on multiple time segments on the history page if the user has visited the page on multiple days.
-  Feature is controlled by a flag for an A/B test

rocketsroger:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/FeatureFlags.kt
+++ b/app/src/main/java/org/mozilla/fenix/FeatureFlags.kt
@@ -96,4 +96,9 @@ object FeatureFlags {
      * Enables the Contile top sites.
      */
     const val contileFeature = false
+
+    /**
+     * Enables history improvement features.
+     */
+    val historyImprovementFeatures = Config.channel.isNightlyOrDebug
 }
--- a/app/src/main/java/org/mozilla/fenix/components/history/PagedHistoryProvider.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/history/PagedHistoryProvider.kt
@@ -12,7 +12,10 @@ import mozilla.components.concept.storage.VisitInfo
 import mozilla.components.concept.storage.VisitType
 import mozilla.components.support.ktx.kotlin.tryGetHostFromUrl
 import org.mozilla.fenix.FeatureFlags
+import org.mozilla.fenix.FeatureFlags.historyImprovementFeatures
 import org.mozilla.fenix.library.history.History
+import org.mozilla.fenix.library.history.HistoryDataSource
+import org.mozilla.fenix.library.history.HistoryItemTimeGroup
 import org.mozilla.fenix.perf.runBlockingIncrement
 import kotlin.math.abs
 
@@ -27,6 +30,9 @@ sealed class HistoryDB {
     abstract val title: String
     abstract val visitedAt: Long
     abstract val selected: Boolean
+    val historyTimeGroup: HistoryItemTimeGroup by lazy {
+        HistoryItemTimeGroup.timeGroupForTimestamp(visitedAt)
+    }
 
     data class Regular(
         override val title: String,
@@ -84,6 +90,9 @@ class DefaultPagedHistoryProvider(
     private val historyStorage: PlacesHistoryStorage,
     private val showHistorySearchGroups: Boolean = FeatureFlags.showHistorySearchGroups,
 ) : PagedHistoryProvider {
+
+    val urlSet = Array<MutableSet<String>>(HistoryItemTimeGroup.values().size) { mutableSetOf() }
+
     /**
      * Types of visits we currently do not display in the History UI.
      */
@@ -113,10 +122,14 @@ class DefaultPagedHistoryProvider(
         numberOfItems: Int,
         onComplete: (List<HistoryDB>) -> Unit,
     ) {
+        if (offset == HistoryDataSource.INITIAL_OFFSET) {
+            urlSet.map { it.clear() }
+        }
+
         // A PagedList DataSource runs on a background thread automatically.
         // If we run this in our own coroutineScope it breaks the PagedList
         runBlockingIncrement {
-            val history: List<HistoryDB>
+            var history: List<HistoryDB>
 
             if (showHistorySearchGroups) {
                 // We need to re-fetch all the history metadata if the offset resets back at 0
@@ -144,6 +157,13 @@ class DefaultPagedHistoryProvider(
                         excludeTypes = excludedVisitTypes
                     )
                     .map { transformVisitInfoToHistoryItem(it) }
+                    .distinctBy { Pair(it.historyTimeGroup, it.url) }
+
+                if (historyImprovementFeatures) {
+                    history = history.distinctBy { Pair(it.historyTimeGroup, it.url) }
+                        .filter { !urlSet[it.historyTimeGroup.ordinal].contains(it.url) }
+                    history.map { urlSet[it.historyTimeGroup.ordinal].add(it.url) }
+                }
             }
 
             onComplete(history)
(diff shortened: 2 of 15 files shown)


app/src/main/java/org/mozilla/fenix/components/history/PagedHistoryProvider.kt
```diff
@@ -144,6 +160,12 @@ class DefaultPagedHistoryProvider(
                         excludeTypes = excludedVisitTypes
                     )
                     .map { transformVisitInfoToHistoryItem(it) }
+                    .distinctBy { Pair(it.historyTimeGroup, it.url) }
```
mavduevskiy, line 163: Won't it filter out the elements with the same `historyTimeGroup` and `url` when `historyImprovementFeatures == false` ?

app/src/main/java/org/mozilla/fenix/library/history/HistoryAdapter.kt
```diff
@@ -83,40 +79,6 @@ class HistoryAdapter(
     }
 
     companion object {
-        private const val zeroDays = 0
```
mavduevskiy, line 86: Why move it away from here? It looks pretty related to `HistoryItemTimeGroup`

app/src/main/java/org/mozilla/fenix/library/history/HistoryDataSource.kt
```diff
@@ -34,10 +34,6 @@ class HistoryDataSource(
     }
 
     override fun loadBefore(params: LoadParams<Int>, callback: LoadCallback<History>) { /* noop */ }
-
-    companion object {
-        private const val INITIAL_OFFSET = 0
```
mavduevskiy, line 39: The same question about moving offset into `ViewModel`. Having a `HistoryDataSource` accessing something from `ViewModel` doesn't look like a very clean data flow solution.

mavduevskiy:
one issue that comes to my mind – deletion of the items. it seems that we basically filter out the same urls during a set period of time, but what happens when the users deletes an item and the list will get reloaded?
I see it that way. We have – let’s say 5 – same items yesterday. We delete the only item we see, and we delete a record from DB, but there are still 4 left. So the next time the list loads – we show the deleted item again. Until we have no items.
If so, possible solution could be deleting not just one item, if the feature is enabled, but also all history items with the same  `historyTimeGroup`

app/src/main/java/org/mozilla/fenix/FeatureFlags.kt
```diff
@@ -91,4 +91,9 @@ object FeatureFlags {
      * Enables showing the wallpaper functionality.
      */
     val showWallpapers = Config.channel.isNightlyOrDebug
+
+    /**
+     * Enables history improvement features.
+     */
+    val historyImprovementFeatures = Config.channel.isNightlyOrDebug
```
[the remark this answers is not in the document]
rocketsroger: Not yet.  We'll have to apply Nimbus to this flag.  It will be a separate change.

app/src/main/java/org/mozilla/fenix/library/history/HistoryViewModel.kt
```diff
@@ -34,6 +36,39 @@ class HistoryViewModel(historyProvider: PagedHistoryProvider) : ViewModel() {
...
+            return Calendar.getInstance().apply {
+                set(Calendar.HOUR_OF_DAY, 0)
+                set(Calendar.MINUTE, 0)
+                set(Calendar.SECOND, 0)
+                set(Calendar.MILLISECOND, 0)
+                add(Calendar.DAY_OF_YEAR, -daysAgo)
+            }.time
+        }
+
+        internal fun timeGroupForTimestamp(timestamp: Long): HistoryItemTimeGroup {
```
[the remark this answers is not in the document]
rocketsroger: Sounds good.  As agreed, I'll move it to it's own class and put the `timeGroupForTimestamp` function in the enum

app/src/main/java/org/mozilla/fenix/components/history/PagedHistoryProvider.kt
rocketsroger: Good catch.  I'll fix.  Thanks

app/src/main/java/org/mozilla/fenix/library/history/HistoryDataSource.kt
rocketsroger: Sounds good I'll move this back.

app/src/main/java/org/mozilla/fenix/library/history/HistoryAdapter.kt
rocketsroger: As agreed, I'll move it to `HistoryItemTimeGroup`

rocketsroger:
> one issue that comes to my mind – deletion of the items. it seems that we basically filter out the same urls during a set period of time, but what happens when the users deletes an item and the list will get reloaded? I see it that way. We have – let’s say 5 – same items yesterday. We delete the only item we see, and we delete a record from DB, but there are still 4 left. So the next time the list loads – we show the deleted item again. Until we have no items. If so, possible solution could be deleting not just one item, if the feature is enabled, but also all history items with the same `historyTimeGroup`

Agreed.  As discussed, we will have to delete the same url between each interval.  Good catch.

app/src/main/java/org/mozilla/fenix/components/history/PagedHistoryProvider.kt
```diff
@@ -27,18 +30,21 @@ sealed class HistoryDB {
     abstract val title: String
     abstract val visitedAt: Long
     abstract val selected: Boolean
+    abstract val historyTimeGroup: HistoryItemTimeGroup
 
     data class Regular(
         override val title: String,
         val url: String,
         override val visitedAt: Long,
+        override val historyTimeGroup: HistoryItemTimeGroup,
```
mavduevskiy, line 39: I was wondering, aren't we always using `visitedAt` value to calculate `HistoryItemTimeGroup`? If so, wouldn't something like that
```
sealed class HistoryDB {
    abstract val title: String
    abstract val visitedAt: Long
    abstract val selected: Boolean
    val historyTimeGroup = HistoryItemTimeGroup.timeGroupForTimestamp(visitedAt)
```
help to remove some code? Because it looks like we are using one parameter to calculate two, and passing the same parameter twice seems like an overkill. Or I am missing something?

mavduevskiy (approved):

app/src/main/java/org/mozilla/fenix/components/history/PagedHistoryProvider.kt
rocketsroger: Yes, I had the same idea during development.  Maybe I was overthinking this.  I was worried that if we add this default behaviour, we end up not copy the the `historyTimeGroup` (if already exists) and do more operations then needed.

rocketsroger: As agreed in a offline discussion, I will add the default helper function.

app/src/main/java/org/mozilla/fenix/components/history/PagedHistoryProvider.kt
```diff
@@ -27,6 +30,9 @@ sealed class HistoryDB {
     abstract val title: String
     abstract val visitedAt: Long
     abstract val selected: Boolean
+    val historyTimeGroup: HistoryItemTimeGroup by lazy {
```
mavduevskiy, line 33: that's a nice solution!

PR #23269, opened 2022-01-18

MickeyMoz:


gabrielluong:
Breaking changes will be fixed in https://github.com/mozilla-mobile/fenix/pull/23271

PR #23311, opened 2022-01-19
For #22851  Add logo animation for wallpapers

Amejia481:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
@@ -17,6 +17,7 @@ import android.view.LayoutInflater
 import android.view.View
 import android.view.View.AccessibilityDelegate
 import android.view.ViewGroup
+import android.view.ViewTreeObserver
 import android.view.accessibility.AccessibilityEvent
 import android.widget.Button
 import android.widget.LinearLayout
@@ -824,6 +825,25 @@ class HomeFragment : Fragment() {
         // triggered to cause an automatic update on warm start (no tab selection occurs). So we
         // update it manually here.
         requireComponents.useCases.sessionUseCases.updateLastAccess()
+        if (shouldAnimateLogoForWallpaper()) {
+            _binding?.sessionControlRecyclerView?.viewTreeObserver?.addOnGlobalLayoutListener(
+                homeLayoutListenerForLogoAnimation
+            )
+        }
+    }
+
+    // To try to find a good time to show the logo animation, we are waiting until all
+    // the sub-recyclerviews (recentBookmarks, collections, recentTabs,recentVisits
+    // and pocketStories) on the home screen have been layout.
+    private val homeLayoutListenerForLogoAnimation = object : ViewTreeObserver.OnGlobalLayoutListener {
+        override fun onGlobalLayout() {
+            _binding?.let { safeBindings ->
+                requireComponents.wallpaperManager.animateLogoIfNeeded(safeBindings.wordmark)
+                safeBindings.sessionControlRecyclerView.viewTreeObserver.removeOnGlobalLayoutListener(
+                    this
+                )
+            }
+        }
     }
 
     override fun onPause() {
@@ -1201,6 +1221,23 @@ class HomeFragment : Fragment() {
         } else {
             emptyList()
         }
+    }
+
+    // We want to show the animation in a time when the user less distracted
+    // The Heuristics are:
+    // 1) The animation hasn't shown before.
+    // 2) The user has onboarded.
+    // 3) It's the third time the user enters the app.
+    // 4) The user is part of the right audience.
+    @Suppress("MagicNumber")
+    private fun shouldAnimateLogoForWallpaper(): Boolean {
+        val localContext = context ?: return false
+        val settings = localContext.settings()
+
+        return shouldEnableWallpaper() && settings.shouldAnimateFirefoxLogo &&
+            onboarding.userHasBeenOnboarded() &&
+            settings.numberOfAppLaunches >= 3 &&
+            FeatureFlags.isThemedWallpapersFeatureEnabled(localContext)
     }
 
     private fun shouldEnableWallpaper() =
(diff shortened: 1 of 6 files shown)


MatthewTighe (approved):

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -1200,6 +1222,17 @@ class HomeFragment : Fragment() {
...
+    // We want to show the animation in a time when the user less distracted
+    // The Heuristics are:
+    // 1) The animation hasn't shown before.
+    // 2) The user has onboarded.
+    // 3) It's the third time the user enters the app.
+    @Suppress("MagicNumber")
+    private fun shouldAnimateLogoForWallpaper() =
+        shouldEnableWallpaper() && context?.settings()?.shouldAnimateFirefoxLogo == true
+            && onboarding.userHasBeenOnboarded()
+            && context?.settings()?.numberOfAppLaunches == 3
```
Amejia481, line 1234: @MatthewTighe just to make you aware, I added some extra logic, after the product input. They want to show the animation when the user is less distracted with other parts of the app, specifically after onboarding has been finished.

MatthewTighe: Thanks for the update! Looks good to me

Amejia481:
👋🏽 
@mcarare @MarcLeclair. I'm having a hard time trying to figure, what it's causing [verifyRunBlockingAndStrictModeSuppresionCount](https://console.firebase.google.com/u/1/project/moz-fenix/testlab/histories/bh.66b7091e15d53d45/matrices/8623599678080407412/executions/bs.8744e3ea1286c83c/test-cases) to failed. If it's not much trouble would you mind giving me a hand :) .

I don't understand how I'm increasing the `suppressionCount`, without using `resetAfter`.
I also tested removing the two `@Suppress("MagicNumber")` locally and I still get the failing test.

```
java.lang.AssertionError: StrictMode suppression count does not match expected count.

    If this PR removed a StrictMode suppression call, great! Please decrease the count.

    Did this PR add or call code that increases the StrictMode suppression count?
    Did you know that suppressing a StrictMode violation can introduce performance regressions?
    Please do your best to implement a solution without adding StrictMode suppression calls.
    Please consult the perf team if you have questions or believe that having this call
    is the optimal solution.


 expected:<20> but was:<21>
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.failNotEquals(Assert.java:835)
	at org.junit.Assert.assertEquals(Assert.java:120)
	at org.mozilla.fenix.perf.StartupExcessiveResourceUseTest.verifyRunBlockingAndStrictModeSuppresionCount(StartupExcessiveResourceUseTest.kt:88)
	at java.lang.reflect.Method.invoke(Native Method)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at androidx.test.rule.ActivityTestRule$ActivityStatement.evaluate(ActivityTestRule.java:531)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:63)
	at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
	at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:293)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:413)
	at org.junit.runners.Suite.runChild(Suite.java:128)
	at org.junit.runners.Suite.runChild(Suite.java:27)
	at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
	at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:293)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:413)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:115)
	at androidx.test.internal.runner.TestExecutor.execute(TestExecutor.java:56)
	at androidx.test.runner.AndroidJUnitRunner.onStart(AndroidJUnitRunner.java:395)
	at android.app.Instrumentation$InstrumentationThread.run(Instrumentation.java:2145)
    
```

Amejia481:
I found the reason of the test regression
It's because [onboarding.userHasBeenOnboarded() ](https://github.com/mozilla-mobile/fenix/pull/23311/files#diff-41731714e97083a1b73584da2837ab11c22796c25b892b1e86ce4190ebba2633R1234). In this case, it's a product requirement that we make sure this animation show after the user has finished onboarding. What we normally do on this cases  @mcarare @MarcLeclair. Should I alter the based line in the test ?

MarcLeclair:
@Amejia481  So, if there's no way around us calling this in the background on startup and since it's only for onboarding, I don't think this is going to be a big issue. However, if we want to keep things consistent with other parts of the code that required removing `StrictMode` to work, we can use [`StrictModeManager.resetAfter()`](https://github.com/mozilla-mobile/fenix/blob/main/app/src/main/java/org/mozilla/fenix/perf/StrictModeManager.kt#L113).  It shouldn't be a big change to the code and should make it pass I think

Amejia481:
> @Amejia481 So, if there's no way around us calling this in the background on startup and since it's only for onboarding, I don't think this is going to be a big issue. However, if we want to keep things consistent with other parts of the code that required removing `StrictMode` to work, we can use [`StrictModeManager.resetAfter()`](https://github.com/mozilla-mobile/fenix/blob/main/app/src/main/java/org/mozilla/fenix/perf/StrictModeManager.kt#L113). It shouldn't be a big change to the code and should make it pass I think

Awesome, thanks I'll give it a try to adding `StrictModeManager.resetAfter()`

PR #23348, opened 2022-01-21
docker: remove tzdata interactive dialog workaround

jcristau:


jcristau:
This also pins the taskcluster python package so we can actually rebuild the base docker image, as the latest taskcluster dropped python2 support.

hneiva (approved):
Curious to know what's missing to upgrade it to py3.
R+

PR #23405, opened 2022-01-25
fixes #23404: add switch to wallpaper settings to disable tap-to-change

### LINKED ISSUE #23404: Add a toggle to the wallpaper settings screen to turn off tap-to-switch
Designs for the wallpaper setting screen include a switch that can be used to enable/disable tapping on the Firefox logo to switch the current wallpaper. Add this to the wallpaper settings screen

MatthewTighe:
image

image

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
@@ -10,6 +10,9 @@ import androidx.compose.foundation.ExperimentalFoundationApi
 import androidx.compose.foundation.Image
 import androidx.compose.foundation.border
 import androidx.compose.foundation.clickable
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.aspectRatio
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
@@ -24,6 +27,8 @@ import androidx.compose.material.Snackbar
 import androidx.compose.material.SnackbarDuration
 import androidx.compose.material.SnackbarHost
 import androidx.compose.material.Surface
+import androidx.compose.material.Switch
+import androidx.compose.material.SwitchDefaults
 import androidx.compose.material.Text
 import androidx.compose.material.TextButton
 import androidx.compose.material.rememberScaffoldState
@@ -57,6 +62,8 @@ import java.util.Locale
  * @param selectedWallpaper The currently selected wallpaper.
  * @param onSelectWallpaper Callback for when a new wallpaper is selected.
  * @param onViewWallpaper Callback for when the view action is clicked from snackbar.
+ * @param tapLogoSwitchChecked Enabled state for switch controlling taps to change wallpaper.
+ * @param onTapLogoSwitchCheckedChange Callback for when state of above switch is updated.
  */
 @Composable
 @Suppress("LongParameterList")
@@ -67,6 +74,8 @@ fun WallpaperSettings(
     selectedWallpaper: Wallpaper,
     onSelectWallpaper: (Wallpaper) -> Unit,
     onViewWallpaper: () -> Unit,
+    tapLogoSwitchChecked: Boolean,
+    onTapLogoSwitchCheckedChange: (Boolean) -> Unit
 ) {
     val coroutineScope = rememberCoroutineScope()
     val scaffoldState = rememberScaffoldState()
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -67,6 +73,8 @@ fun WallpaperSettings(
     selectedWallpaper: Wallpaper,
     onSelectWallpaper: (Wallpaper) -> Unit,
     onViewWallpaper: () -> Unit,
+    tapLogoSwitchChecked: Boolean,
+    tapLogoSwitchOnCheckedChange: (Boolean) -> Unit
```
Amejia481, line 77: nit: lets add docs for the new parameters :)

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettingsFragment.kt
```diff
@@ -49,6 +54,11 @@ class WallpaperSettingsFragment : Fragment() {
                             wallpaperManager.currentWallpaper = selectedWallpaper
                         },
                         onViewWallpaper = { findNavController().navigate(R.id.homeFragment) },
+                        tapLogoSwitchChecked = wallpapersSwitchedByLogo,
+                        tapLogoSwitchOnCheckedChange = {
+                            settings.wallpapersSwitchedByLogoTap = it
+                            wallpapersSwitchedByLogo = it
```
Amejia481, line 60: Nice ✨  this what was causing me some Headaches, as for me the switch was not updating after click 👍🏽

Amejia481: I guess compose it's smart enough to just re-render only the changed bits, right?  As `wallpapersSwitchedByLogo` is parameter of `WallpaperSettings`.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -80,7 +88,7 @@ fun WallpaperSettings(
             }
         }
     ) {
-        Surface(color = FirefoxTheme.colors.layer2) {
+        Column {
```
Amejia481, line 91: ✨ !

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
MatthewTighe: Thanks for the reminder, adding some docs

MatthewTighe:
Added some screen shots which I'd forgetting in the original post

Amejia481:
🚢 !

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -57,6 +62,8 @@ import java.util.Locale
  * @param selectedWallpaper The currently selected wallpaper.
  * @param onSelectWallpaper Callback for when a new wallpaper is selected.
  * @param onViewWallpaper Callback for when the view action is clicked from snackbar.
+ * @param tapLogoSwitchChecked Enabled state for switch controlling taps to change wallpaper.
+ * @param tapLogoSwitchOnCheckedChange Callback for when state of above switch is updated.
```
gabrielluong, line 66:
```suggestion
 * @param onCheckedChange Callback for when state of above switch is updated.
```
Can we use the same `onXXXX` wording for our Compose callbacks here?

MatthewTighe: Good call, updated

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettingsFragment.kt
MatthewTighe: There's a few important bits to make this work. Lemme deconstruct this line a little bit:
`var wallpapersSwitchedByLogo by remember { mutableStateOf(settings.wallpapersSwitchedByLogoTap) }`

`mutableStateOf` changes the underlying type of `wallpapersSwitchedByLogo` to a `State<Boolean>` When `State` variables are read from within a composition, the composition will automatically subscribe to changes to the variable. This means that when `wallpapersSwitchedByLogo` is updated, the composition can recompose the elements of the UI tree that use `wallpapersSwitchedByLogo`. The runtime does this in what they're calling "smart recomposition", which means that only `@Composable` functions that use the variable should get redrawn. 

`by remember { }` is a delegate that does two things: 
- lets us treat the variable as a `Boolean` instead of a `State<Boolean>`
- informs the composition that this value does not need to be initialized again when the composable is called again, which effectively gives the composable memory

Amejia481: Thanks for the clarification 👍🏽

Issue #23378, opened 2022-01-24
[Bug]: ↖️/↙️ buttons' icons in Search Suggestions are not visible in Nightly 


czlucius:
### Steps to reproduce

1. Click on the address bar
2. Type something in the address bar that triggers search suggestions

### Expected behaviour

↖️/↙️ buttons are shown on the right of search suggestions. (I don't know the name of the button)

### Actual behaviour

The buttons are still there, but icons are invisible. 

### Device name

Samsung Galaxy S20 FE 5G 

### Android version

Android 12

### Firefox release type

Firefox Nightly

### Firefox version

98.0a1 (Build #2015859243)

### Additional information

This does not occur in Firefox Stable.
See attached screenshot.
Nightly:
![Screenshot_20220125-005143_Firefox Nightly](https://user-images.githubusercontent.com/58442255/150827628-ff0072ed-6d13-4f5d-beff-62e39997a717.jpg)
(You can see that the buttons are still clickable in the 1st bugzilla entry)
Stable:
![Screenshot_20220125-005001_Firefox](https://user-images.githubusercontent.com/58442255/150827495-fde44924-6e80-408c-bfbf-5f4d46f65fdf.jpg)

kbrosnan:
Similar to https://github.com/mozilla-mobile/fenix/issues/23350 but this is on modern Android

czlucius:
Yes, I saw that issue before posting this one.
But the icons in the other issue are still visible to me.
The issue is also not present in Firefox Stable/Release.

mcarare:
I am inclined to say this is not an issue in AC but in the way we override the `mozac_ui_icons_fill` value. Not sure what are the colors ids for this so I am tagging @gabrielluong for input.

gabrielluong:
Should be fixed with the changes reverted in https://github.com/mozilla-mobile/fenix/issues/23350

PR #23449, opened 2022-01-27
For #23378 - Fix search suggestion buttons not visible

tasosb:
Fixes the problem with the non-visible suggestion button. (regressed by 12347c99999c340e7f7072ab2749cb13c5fd86c9 )
Removed the attribute ?primaryText and added the static color value to to both /colors.xml and night/colors.xml

fx

CC: @gabrielluong who worked on the color changes

### Pull Request checklist
- [x] **Tests**: N/A
- [x] **Screenshots**
- [x] **Accessibility**: N/A

1. click on Show All Checks,

### DIFF
--- a/app/src/main/res/values/colors.xml
+++ b/app/src/main/res/values/colors.xml
@@ -279,10 +279,12 @@
     <color name="toolbar_menu_transparent">@android:color/transparent</color>
 
     <!-- Mozilla Android Component color variable overrides -->
-    <color name="mozac_ui_icons_fill" tools:ignore="UnusedResources">?primaryText</color>
+    <color name="mozac_ui_icons_fill" tools:ignore="UnusedResources">@color/primary_text_light_theme</color>
+
     <!-- Tab Counter colors -->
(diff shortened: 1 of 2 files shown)


app/src/main/res/values/colors.xml
```diff
@@ -279,10 +279,12 @@
     <color name="toolbar_menu_transparent">@android:color/transparent</color>
 
     <!-- Mozilla Android Component color variable overrides -->
-    <color name="mozac_ui_icons_fill" tools:ignore="UnusedResources">?primaryText</color>
+    <color name="mozac_ui_icons_fill" tools:ignore="UnusedResources">@color/primary_text_light_theme</color>
```
gabrielluong, line 282: Will look into this. I don't think this will work for the private theme, which is why `?primaryText` was used.

tasosb: Yes you're right, I didn't think about the private mode. In that case the issue might be in AC where the icon is tinted.

tasosb: OK, I believe I found a (simple) fix in AC, I will make a PR there

gabrielluong: 👀

PR #23525, opened 2022-02-01
Issue #12731: Disable TrimMemoryMiddleware on all channels

csadilek:
This was already disabled in Nightly over two months ago and seems to work much better combined with latest fixes in GV (see links inline for details.)


mcomella (approved):
lgtm as discussed with the GV team. While I'm not too familiar with this code, I trust it because:
- this flag has been disabled in Nightly and no one complained
- the fixes the GV team landed were verified in Nightly by the community, i.e. with this flag off. It makes sense to disable it for all channels too

PR #23538, opened 2022-02-02
docker: run get-secret.py with python3

jcristau:


taskcluster/docker/base/Dockerfile
```diff
@@ -34,19 +34,17 @@ RUN apt-get update -qq \
...
                           tzdata \
                           unzip \
 			  mercurial \
     && apt-get clean
 
-RUN pip install --upgrade pip
-# TODO update to python3 so we can use newer taskcluster
-RUN pip install taskcluster==44.2.2
+RUN pip3 install --upgrade pip
+RUN pip install taskcluster
```
jcristau, line 47: I'm not sure if we want to keep some sort of pinning here, other opinions welcome...

jcristau: `pip3` is from the `python3-pip` deb package, `pip` is pip-installed (I'm also not sure *why* we're updating pip here, but oh well).

escapewindow (approved):

PR #23400, opened 2022-01-25
First use of Nimbus FML plugin

jhugman:
This requires the newly landed Nimbus FML plugin.

Closes #22925 

### Pull Request checklist

- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features. In addition, it includes a screenshot of a successful [accessibility scan](https://play.google.com/store/apps/details?id=com.google.android.apps.accessibility.auditor&hl=en_US) to ensure no new defects are added to the product.

### To download an APK when reviewing a PR:
1. click on Show All Checks,
2. click Details next to "Taskcluster (pull_request)" after it appears and then finishes with a green checkmark,
3. click on the "Fenix - assemble" task, then click "Run Artifacts".
4. the APK links should be on the left side of the screen, named for each CPU architecture

PR #23556, opened 2022-02-03
Fixes #23492: Perf regression of calling isFirefoxDefault from main thread

### LINKED ISSUE #23492: ~100ms regression in cold main time to first frame (1/28; Nimbus FML?)
Caught in our nightly backfill:

image

We go from 1.468s to 1.570s (102ms).

### LINKED ISSUE #23618: Fenix shouldn't show "Set as default browser" button if it's already the default browser
When I started Fenix Nightly today, its homepage had a tile asking me to set it as my default browser ("Set as default browser"). But Fenix Nightly is already my default browser!

@jonalmeida says Fenix should be able to check whether it's already set as the default app.

AC has code to query the default browser, but it might require some legacy permissions: [[https://github.com/mozilla-mobile/android-components/blob/main/components/support/[…]tils/src/main/java/mozilla/components/support/utils/Browsers.kt](https://github.com/mozilla-mobile/android-components/blob/main/components/support/utils/src/main/java/mozilla/components/support/utils/Browsers.kt#L115).](https://github.com/mozilla-mobile/android-components/blob/ec0201897ae2b2d91ec36be41986dff1c3192501/components/support/utils/src/main/java/mozilla/components/support/utils/Browsers.kt#L112-L115)

jhugman:
Fixes #23492
Fixes #23618

Perf regression calling slow isFirefoxDefault from the main thread.

Prior #23400, the default browser message was only displayed when part of an experiment. The check that isFirefoxDefault was only done as a final check before the message is displayed.

#23400 changed the order of this, so the slow check was done each time the toolbar was created.

Fixes #23492; this PR does three things:

- it moves the check for isDefaultBrowser until _after_ the check to see which message surface to display on (as was prior to #23400 )
- it makes the default browser message be able to turn off.
- it changes the feature of showing the default browser message to be off by default (as was prior to #23400 )

Now, the default browser message is displayable when an experiment toggles it on.

The isFirefoxDefault _is_ done when the message is toggled on, but:

1. There are no plans to toggle this on in this release
2. The messaging system is being re-written to avoid these problems.

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
@@ -446,10 +446,13 @@ open class DefaultToolbarMenu(
         }
     }
 
-    private fun getSetDefaultBrowserItem(): BrowserMenuImageText? =
-        if (BrowsersCache.all(context).isFirefoxDefaultBrowser) {
-            null
-        } else if (FxNimbus.features.defaultBrowserMessage.value().messageLocation == MessageSurfaceId.APP_MENU_ITEM) {
+    private fun getSetDefaultBrowserItem(): BrowserMenuImageText? {
+        val browsers = BrowsersCache.all(context)
+        val config = FxNimbus.features.defaultBrowserMessage.value()
+        return if (
+            config.messageLocation == MessageSurfaceId.APP_MENU_ITEM &&
+            !browsers.isFirefoxDefaultBrowser
+        ) {
             BrowserMenuImageText(
                 label = context.getString(R.string.preferences_set_as_default_browser),
                 imageResource = R.mipmap.ic_launcher
@@ -459,4 +462,5 @@ open class DefaultToolbarMenu(
         } else {
             null
         }
+    }
 }
--- a/app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
@@ -447,11 +447,9 @@ open class DefaultToolbarMenu(
     }
 
     private fun getSetDefaultBrowserItem(): BrowserMenuImageText? {
-        val browsers = BrowsersCache.all(context)
-        val config = FxNimbus.features.defaultBrowserMessage.value()
+        val settings = context.components.settings
         return if (
-            config.messageLocation == MessageSurfaceId.APP_MENU_ITEM &&
-            !browsers.isFirefoxDefaultBrowser
+            settings.isDefaultBrowserMessageLocation(MessageSurfaceId.APP_MENU_ITEM)
         ) {
             BrowserMenuImageText(
                 label = context.getString(R.string.preferences_set_as_default_browser),
--- a/app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
@@ -39,10 +39,8 @@ import org.mozilla.fenix.R
 import org.mozilla.fenix.components.accounts.FenixAccountManager
 import org.mozilla.fenix.ext.components
 import org.mozilla.fenix.ext.settings
-import org.mozilla.fenix.nimbus.FxNimbus
 import org.mozilla.fenix.nimbus.MessageSurfaceId
 import org.mozilla.fenix.theme.ThemeManager
-import org.mozilla.fenix.utils.BrowsersCache
 
 /**
  * Builds the toolbar object used with the 3-dot menu in the browser fragment.
--- a/app/src/main/java/org/mozilla/fenix/home/HomeMenu.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/HomeMenu.kt
@@ -175,14 +175,9 @@ class HomeMenu(
 
         // Use nimbus to set the icon and title.
         val nimbusValidation = FxNimbus.features.nimbusValidation.value()
-        val settingsIcon = context.resources.getIdentifier(
-            nimbusValidation.settingsIcon,
-            "drawable",
-            context.packageName
-        )
         val settingsItem = BrowserMenuImageText(
             nimbusValidation.settingsTitle,
-            settingsIcon,
+            R.drawable.mozac_ic_settings,
             primaryTextColor
         ) {
             onItemTapped.invoke(Item.Settings)
--- a/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
+++ b/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
@@ -466,7 +466,7 @@ class Settings(private val appContext: Context) : PreferencesHolder {
      */
     var searchTermTabGroupsAreEnabled by lazyFeatureFlagPreference(
         appContext.getPreferenceKey(R.string.pref_key_search_term_tab_groups),
-        default = { FxNimbus.features.searchTermGroups.value().enabled },
+        default = { FxNimbus.features.searchTermGroups.value(appContext).enabled },
         featureFlag = FeatureFlags.tabGroupFeature
     )
 
--- a/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
+++ b/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
@@ -37,6 +37,7 @@ import org.mozilla.fenix.components.settings.lazyFeatureFlagPreference
 import org.mozilla.fenix.components.toolbar.ToolbarPosition
 import org.mozilla.fenix.ext.components
 import org.mozilla.fenix.ext.getPreferenceKey
+import org.mozilla.fenix.nimbus.DefaultBrowserMessage
 import org.mozilla.fenix.nimbus.FxNimbus
 import org.mozilla.fenix.nimbus.HomeScreenSection
 import org.mozilla.fenix.nimbus.MessageSurfaceId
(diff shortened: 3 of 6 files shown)


app/src/main/java/org/mozilla/fenix/components/toolbar/DefaultToolbarMenu.kt
```diff
@@ -446,10 +446,13 @@ open class DefaultToolbarMenu(
...
-    private fun getSetDefaultBrowserItem(): BrowserMenuImageText? =
-        if (BrowsersCache.all(context).isFirefoxDefaultBrowser) {
-            null
-        } else if (FxNimbus.features.defaultBrowserMessage.value().messageLocation == MessageSurfaceId.APP_MENU_ITEM) {
+    private fun getSetDefaultBrowserItem(): BrowserMenuImageText? {
+        val browsers = BrowsersCache.all(context)
+        val config = FxNimbus.features.defaultBrowserMessage.value()
+        return if (
+            config.messageLocation == MessageSurfaceId.APP_MENU_ITEM &&
+            !browsers.isFirefoxDefaultBrowser
```
Amejia481, line 454: I'm not sure how `BrowsersCache.all()` could impact performace, but  maybe we could make it a bit faster if we  inline the `BrowsersCache.all(context)` call with the condition check, this way we will only call `BrowsersCache.all( )` when it's the right message? . What do you think?

```kotlin
 config.messageLocation == MessageSurfaceId.APP_MENU_ITEM &&
 !BrowsersCache.all(context).isFirefoxDefaultBrowser
```

Amejia481 (approved):
LGTM!

gabrielluong:
Decision Task broke, restarting

mcomella:
I measured the performance difference. Results from original bisection:
- before regression: 1373ms median
- regression: 1501ms

Results from this PR:
- before PR: 1436ms
- after PR: 1402ms

It seems something may have changed after the initial regression causing the overall regression to be 63ms rather than ~100ms. **One notable visual difference is that in the regressing builds there is a new UI element, "Set as default browser".** The home screen races various UI elements against each other (when it should inflate from the top) so that may contribute to the different results. This PR improves the regression by another 34ms so we've still regressed ~29ms from the original state. Ideally we can try to address that (e.g. if we regress 30ms 2 more times, we're back to where we started) but I didn't see anything obviously actionable in the profiles. I'd be curious to see what our Nightly results say about this PR rather than my local measurements.

app/src/main/java/org/mozilla/fenix/home/HomeMenu.kt
```diff
@@ -175,14 +175,9 @@ class HomeMenu(
 
         // Use nimbus to set the icon and title.
         val nimbusValidation = FxNimbus.features.nimbusValidation.value()
-        val settingsIcon = context.resources.getIdentifier(
-            nimbusValidation.settingsIcon,
-            "drawable",
-            context.packageName
-        )
```
jhugman, line 182: Calling `getIdentifier` is entirely avoidable, for the sake of a validation.

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -326,15 +327,25 @@ class Settings(private val appContext: Context) : PreferencesHolder {
...
+
+    fun isDefaultBrowserMessageLocation(surfaceId: MessageSurfaceId): Boolean =
+        defaultBrowserFeature.messageLocation?.let { experimentalSurfaceId ->
+            if (experimentalSurfaceId == surfaceId) {
+                val browsers = BrowsersCache.all(appContext)
+                !browsers.isFirefoxDefaultBrowser
+            } else {
+                false
+            }
+        } ?: false
```
jhugman, line 347: These lines do several things:

 * we put all the checking logic into one place.
 * we now only call into Rust once for the default browser feature.
 * we now only check `isFirefoxDefaultBrowser` after we check the surface id. There are no experiments about this currently, so this never happens now.

jhugman (commented):
I've added more aggressive caching of calling into Rust and checking the browser default.

The FML builds on top of the `getVariables() -> Variables` API and generates straightline kotlin to access those `Variables`.

The rest of the PR #23400 converts existing Fenix code to use the generated code, so this Pr concentrates on that.

@mcomella please could you re-profile? (are there profiles on CI for PRs with the Performance label?)

Amejia481:
> One notable visual difference is that in the regressing builds there is a new UI element, "Set as default browser".

@jhugman this looks similar to what @mcomella described on https://github.com/mozilla-mobile/fenix/pull/23556#issuecomment-1029338172 see https://github.com/mozilla-mobile/fenix/issues/23618#issuecomment-1031970923

Amejia481:
> > One notable visual difference is that in the regressing builds there is a new UI element, "Set as default browser".
> 
> @jhugman this looks similar to what @mcomella described on [#23556 (comment)](https://github.com/mozilla-mobile/fenix/pull/23556#issuecomment-1029338172) see [#23618 (comment)](https://github.com/mozilla-mobile/fenix/issues/23618#issuecomment-1031970923)

I think what is causing the regression is what @mcomella was describing above the **"Set as default browser"** card is showing all the time. Investigating https://github.com/mozilla-mobile/fenix/issues/23618 I found that 
[val isExperimentBranch = feature.messageLocation == MessageSurfaceId.HOMESCREEN_BANNER](https://github.com/mozilla-mobile/fenix/blob/26c7562e62359fa0ec57e8d4f9db1695bb58b45f/app/src/main/java/org/mozilla/fenix/utils/Settings.kt#L331)  it's always true after https://github.com/mozilla-mobile/fenix/commit/82a6f8cae411603ba5abced3b13fc43ca40f335e. See https://github.com/mozilla-mobile/fenix/issues/23618#issuecomment-1032011656 for more context.

mcomella:
> @mcomella please could you re-profile? (are there profiles on CI for PRs with the Performance label?)

I tried to check out and run this PR but it crashes on start up with:
```
02-07 16:39:03.891 19530 19563 E AndroidRuntime: FATAL EXCEPTION: DefaultDispatcher-worker-5
02-07 16:39:03.891 19530 19563 E AndroidRuntime: Process: org.mozilla.fenix, PID: 19530
02-07 16:39:03.891 19530 19563 E AndroidRuntime: org.mozilla.experiments.nimbus.internal.NimbusFeatureException: A Context is needed but not available. Consider passing in a context to the value() method when close to startup
02-07 16:39:03.891 19530 19563 E AndroidRuntime:        at org.mozilla.experiments.nimbus.internal.FeatureHolder.value$default(FeatureHolder.kt:7)
02-07 16:39:03.891 19530 19563 E AndroidRuntime:        at org.mozilla.fenix.utils.Settings$searchTermTabGroupsAreEnabled$2.invoke(Settings.kt:4)
02-07 16:39:03.891 19530 19563 E AndroidRuntime:        at org.mozilla.fenix.components.settings.LazyPreference$property$2.invoke(FeatureFlagPreference.kt:4)
02-07 16:39:03.891 19530 19563 E AndroidRuntime:        at kotlin.SynchronizedLazyImpl.getValue(LazyJVM.kt:5)
02-07 16:39:03.891 19530 19563 E AndroidRuntime:        at org.mozilla.fenix.components.settings.LazyPreference.getValue(FeatureFlagPreference.kt:3)
02-07 16:39:03.891 19530 19563 E AndroidRuntime:        at org.mozilla.fenix.utils.Settings.getSearchTermTabGroupsAreEnabled(Settings.kt:1)
02-07 16:39:03.891 19530 19563 E AndroidRuntime:        at org.mozilla.fenix.FenixApplication$initializeGlean$2.invokeSuspend(FenixApplication.kt:23)
02-07 16:39:03.891 19530 19563 E AndroidRuntime:        at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:3)
02-07 16:39:03.891 19530 19563 E AndroidRuntime:        at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:18)
02-07 16:39:03.891 19530 19563 E AndroidRuntime:        at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:1)
02-07 16:39:03.891 19530 19563 E AndroidRuntime:        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:10)
```

@jhugman btw, did you mean a benchmark or a profile?

> (are there profiles on CI for PRs with the Performance label?)

Unfortunately, no – we haven't had the resources to implement something like that (e.g. our daily graphs from nightly are actually manually run with a script on a device on my desk).

For benchmarks, you can run it locally if you have a low-end phone (or a high-end phone you can reproduce the regression on). See https://wiki.mozilla.org/Performance/Fenix/Performance_reviews#Benchmark_locally. This regression was to the `cold_main_first_frame` benchmark.

jhugman:
Ah. Ok yes, I keep forgetting to say: The default browser message was shown as part of the #23400 , and this PR fixes that.

Fixed this early startup crash, with the suggested fix.

```
 org.mozilla.experiments.nimbus.internal.NimbusFeatureException: A Context is needed but not available. Consider passing in a context to the value() method when close to startup
        …
        at org.mozilla.fenix.utils.Settings$searchTermTabGroupsAreEnabled$2.invoke(Settings.kt:4)
```

jhugman:
> @jhugman btw, did you mean a benchmark or a profile?

I'm not sure, whichever you needed to find that 102ms regression.

Also question from the curious: given the precision of `ms` you're giving, I assumed that the error bars were small, however, knowing you're benchmarking on a local machine, I have no idea now. What sort of variation/error bars do you see in timings?

mcomella:
> > @jhugman btw, did you mean a benchmark or a profile?
> 
> I'm not sure, whichever you needed to find that 102ms regression.

Sure. Results from original bisection:
- before regression: 1373ms median
- regression: 1501ms

Results from this PR, taken with our `cold_main_first_frame` benchmark run on a Moto G5 (i.e. `./perf-tools/measure_start_up.py nightly cold_main_first_frame results.txt`):
- before PR: 1443ms
- after PR: 1406ms

These results are similar to [when I last measured this PR](https://github.com/mozilla-mobile/fenix/pull/23556#issuecomment-1029338172).

> Also question from the curious: given the precision of `ms` you're giving, I assumed that the error bars were small, however, knowing you're benchmarking on a local machine, I have no idea now. What sort of variation/error bars do you see in timings?

I don't know statistics well so I'm not sure how well I can answer this question 😓 (and yes, our results could probably be improved by someone more familiar with statistics). Through trial-and-error we decided to execute 25 iterations of each test and to take the median values of the results to find the duration of a benchmark for a build. When we originally set these values, this had a maximum variation of ~20ms for an A/A test. In the worst case, I think this would let us catch regressions of ~40ms+ (based [on this thinking I did](https://docs.google.com/document/d/1QKHK50VJisT2k5kTfbNwDb5MvKKTi5H9xAKR9s-Hzu8/edit)). However, the average variation is usually smaller and we can look at the results across multiple days (so hundreds of iterations instead of 25) so in practice we've caught smaller regressions. The variation seems larger these days though:  maybe ~30ms. Our [nightly benchmark results dashboards (populated once a week on mondays) is here](https://earthangel-b40313e5.influxcloud.net/d/s3AT6t7nk/fenix-startup-nightly-via-backfill-py?orgId=1&from=now-30d&to=now) if you're curious.

For context, we use local machines right now rather than CI due to limited resourcing: CI would be ideal because it's more accessible but it takes more time to implement and debug when issues occur.

Issue #23563, opened 2022-02-03
Crash IOException: Timed out when downloading wallpapers 


Amejia481:
Sentry https://sentry.prod.mozaws.net/operations/firefox-nightly/issues/17697998/
```
java.io.IOException: Timed out
    at org.mozilla.geckoview.GeckoInputStream.read(GeckoInputStream.java:11)
    at org.mozilla.geckoview.GeckoInputStream.read(GeckoInputStream.java:6)
    at kotlin.io.ByteStreamsKt.copyTo$default(IOStreams.kt:4)
    at org.mozilla.fenix.wallpapers.WallpaperDownloader$downloadWallpaper$2$2$1.invoke(WallpaperDownloader.kt:3)
    at mozilla.components.concept.fetch.Response$Body.useStream(Response.kt:1)
    at org.mozilla.fenix.wallpapers.WallpaperDownloader$downloadWallpaper$2.invokeSuspend(WallpaperDownloader.kt:53)
    at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:3)
    at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:18)
    at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:1)
    at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:10)
```

Amejia481:
Sentry issue: [FIREFOX-NIGHTLY-EYN](https://sentry.prod.mozaws.net/operations/firefox-nightly/issues/17697998/?referrer=github_plugin)

Amejia481:
Addressed on https://github.com/mozilla-mobile/fenix/commit/e290bc1bd6c4a9e3185e8f0bf9ffb433d378620d

PR #23564, opened 2022-02-03
For #23563: Address time out crash when downloading wallpaper

Amejia481:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperDownloader.kt
+++ b/app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperDownloader.kt
@@ -44,14 +44,13 @@ class WallpaperDownloader(
                 method = Request.Method.GET
             )
             Result.runCatching {
-                client.fetch(request)
-            }.onSuccess {
-                if (!it.isSuccess) {
-                    logger.error("Download response failure code: ${it.status}")
+                val response = client.fetch(request)
+                if (!response.isSuccess) {
+                    logger.error("Download response failure code: ${response.status}")
                     return@withContext
                 }
                 File(localFile.path.substringBeforeLast("/")).mkdirs()
-                it.body.useStream { input ->
+                response.body.useStream { input ->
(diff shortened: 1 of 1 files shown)


MozillaNoah (approved):

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperDownloader.kt
```diff
@@ -50,9 +51,17 @@ class WallpaperDownloader(
...
+                    File(localFile.path.substringBeforeLast("/")).mkdirs()
+                    it.body.useStream { input ->
+                        input.copyTo(localFile.outputStream())
+                    }
+                } catch (exception: IOException) {
+                    logger.error(
+                        exception.message ?: "Download failed: no throwable message included.",
+                        exception
+                    )
+                    crashReporter.submitCaughtException(exception)
```
MatthewTighe, line 64: This could be a `map` on the `Result.runCatching` block, so that the `onFailure` block is shared. Then we aren't mixing our exception handling patterns.

Alternatively, the whole thing could be switching to a `try catch`

MatthewTighe (approved):
Thanks for handling this! Just had one small nit

MatthewTighe:
Actually, I'm worried now that some amount of data will still be copied to the output stream. It may be worth deleting the file on failure, to save users space and to make sure we don't get partial bitmaps. Could handle that in a followup as well

Amejia481:
> Actually, I'm worried now that some amount of data will still be copied to the output stream. It may be worth deleting the file on failure, to save users space and to make sure we don't get partial bitmaps. Could handle that in a followup as well

Let's address it in a follow up!

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperDownloader.kt
Amejia481: Updated!

PR #23567, opened 2022-02-04
closes #23565: expire remote wallpapers but let users keep selected

### LINKED ISSUE #23565: Add expiration date for wallpapers
Certain remote wallpapers will have an expiration date tied to them. When wallpapers expire, we want to delete any unused copies of them from user's devices. 

The exception is that if a user has an expired wallpaper selected, they can continue to use it.

MatthewTighe:
1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperManager.kt
+++ b/app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperManager.kt
@@ -7,6 +7,7 @@ package org.mozilla.fenix.wallpapers
 import android.animation.AnimatorSet
 import android.animation.ObjectAnimator
 import android.content.Context
+import android.content.res.Configuration
 import android.graphics.Bitmap
 import android.graphics.BitmapFactory
 import android.graphics.drawable.BitmapDrawable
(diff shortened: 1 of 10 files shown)


Amejia481 (approved):
👏🏽  awesome work !!!

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperManager.kt
```diff
@@ -95,33 +97,43 @@ class WallpaperManager(
         }
     }
 
+    private fun filterExpiredRemoteWallpapers(wallpaper: Wallpaper): Boolean = when (wallpaper) {
+        is Wallpaper.Remote -> {
+            val notExpired = wallpaper.expirationDate?.let { Date().before(it) } ?: true
+            notExpired || wallpaper.name == settings.currentWallpaper
```
jonalmeida, line 103: Drive-by 🚗 : Out of curiosity, if had a remote wallpaper selected at some point, but is no longer selected, would that disappear?

Amejia481: Yes, If the wallpaper is expired it will disappear.

MatthewTighe: That's currently the spec. We'll hopefully add some UX around that to make it a little more user-friendly soon. It's easy to imagine improving the persistence layer eventually to allow people to keep them for longer as well

jonalmeida: Sounds good, thanks that clarifying!

PR #23573, opened 2022-02-04
Close #10046: Add history search

### LINKED ISSUE #10046: Allow user to search within History
### Why/User Benefit/User Problem
As a user, I would like to search from my history (like I can in competing browsers like Chrome). 

### What/Requirements
* Search icon in Library History UI. 
* tapping icon should show input field
* typing in input should show pages from history matching my query

### Acceptance Criteria (how do I know when I’m done?)
When there is an icon in the Library History page that allows me to search previously visited pages.

rocketsroger:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/library/history/HistorySearchDialogFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/library/history/HistorySearchDialogFragment.kt
@@ -1,0 +1,257 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.library.history
+
+import android.annotation.SuppressLint
+import android.app.Dialog
+import android.content.Context
+import android.content.DialogInterface
+import android.os.Build
+import android.os.Bundle
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import android.view.ViewStub
+import android.view.accessibility.AccessibilityEvent
+import android.view.inputmethod.InputMethodManager
+import androidx.appcompat.app.AppCompatDialogFragment
+import androidx.constraintlayout.widget.ConstraintProperties.BOTTOM
+import androidx.constraintlayout.widget.ConstraintProperties.PARENT_ID
+import androidx.constraintlayout.widget.ConstraintProperties.TOP
+import androidx.constraintlayout.widget.ConstraintSet
+import androidx.core.view.isVisible
+import androidx.lifecycle.lifecycleScope
+import kotlinx.coroutines.flow.collect
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.launch
+import mozilla.components.lib.state.ext.consumeFlow
+import mozilla.components.lib.state.ext.consumeFrom
+import mozilla.components.support.base.feature.UserInteractionHandler
+import mozilla.components.support.ktx.android.view.hideKeyboard
+import mozilla.components.support.ktx.kotlinx.coroutines.flow.ifChanged
+import org.mozilla.fenix.BrowserDirection
+import org.mozilla.fenix.HomeActivity
+import org.mozilla.fenix.R
+import org.mozilla.fenix.components.toolbar.ToolbarPosition
+import org.mozilla.fenix.databinding.FragmentHistorySearchDialogBinding
+import org.mozilla.fenix.databinding.SearchSuggestionsHintBinding
+import org.mozilla.fenix.ext.settings
+import org.mozilla.fenix.library.history.awesomebar.AwesomeBarView
+import org.mozilla.fenix.library.history.toolbar.ToolbarView
+import org.mozilla.fenix.settings.SupportUtils
+
+@Suppress("TooManyFunctions")
+class HistorySearchDialogFragment : AppCompatDialogFragment(), UserInteractionHandler {
+    private var _binding: FragmentHistorySearchDialogBinding? = null
+    private val binding get() = _binding!!
+
+    private lateinit var interactor: HistorySearchDialogInteractor
+    private lateinit var store: HistorySearchFragmentStore
+    private lateinit var toolbarView: ToolbarView
+    private lateinit var awesomeBarView: AwesomeBarView
+
+    private var dialogHandledAction = false
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setStyle(STYLE_NO_TITLE, R.style.SearchDialogStyle)
+    }
+
+    override fun onCreateDialog(savedInstanceState: Bundle?): Dialog {
+        return object : Dialog(requireContext(), this.theme) {
+            override fun onBackPressed() {
+                this@HistorySearchDialogFragment.onBackPressed()
+            }
+        }
+    }
+
+    override fun onCreateView(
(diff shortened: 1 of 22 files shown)


app/src/test/java/org/mozilla/fenix/library/history/HistorySearchControllerTest.kt
```diff
@@ -0,0 +1,85 @@
...
+    fun handleTextChangedEmpty() {
+        val text = ""
+
+        createController().handleTextChanged(text)
+
+        verify { store.dispatch(HistorySearchFragmentAction.UpdateQuery(text)) }
+    }
+
+    @Test
+    fun handleUrlTapped() {
```
MozillaNoah, line 59: I've noticed that some tests' names are derivations of the functions they're validating while others are the `WHEN this thing happens THEN this other thing should happen as a result`. What should be the standard practice?

app/src/main/java/org/mozilla/fenix/library/history/HistorySearchFragmentStore.kt
```diff
@@ -0,0 +1,56 @@
...
+ * The state for the History Search Screen
+ *
+ * @property query The current search query string
+ */
+data class HistorySearchFragmentState(
+    val query: String,
+) : State
+
+fun createInitialHistorySearchFragmentState(): HistorySearchFragmentState {
+
```
MozillaNoah, line 31: nit: empty line

app/src/main/java/org/mozilla/fenix/library/history/HistorySearchDialogFragment.kt
```diff
@@ -0,0 +1,261 @@
...
+        val awesomeBar = binding.awesomeBar
+
+        awesomeBarView = AwesomeBarView(
+            activity,
+            interactor,
+            awesomeBar,
+        )
+
+        awesomeBarView.view.setOnEditSuggestionListener(toolbarView.view::setSearchTerms)
+        binding.searchWrapper.background = ColorDrawable(Color.TRANSPARENT)
```
MozillaNoah, line 112: why set this at runtime instead of in the XML?

app/src/main/java/org/mozilla/fenix/library/history/HistorySearchDialogFragment.kt
```diff
@@ -0,0 +1,261 @@
...
+
+        return binding.root
+    }
+
+    @SuppressLint("ClickableViewAccessibility")
```
MozillaNoah, line 117: what's this suppress doing? (haven't used it before)

app/src/main/res/layout/fragment_history_search_dialog.xml
```diff
@@ -0,0 +1,111 @@
...
+        app:browserToolbarInsecureColor="?primaryText"
+        app:browserToolbarMenuColor="?primaryText"
+        app:browserToolbarProgressBarGravity="bottom"
+        app:browserToolbarSecureColor="?primaryText"
+        app:browserToolbarTrackingProtectionAndSecurityIndicatorSeparatorColor="?toolbarDivider"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toTopOf="parent"/>
+
+    <ViewStub
```
MozillaNoah, line 34: I've actually never used a ViewStub before. what's the TLDR on what's going on here?

MozillaNoah:
After you search any query, if you tap on the area below the search results, the search results disappear. Is this intentional?

MozillaNoah:
If you open the search bar, it will remain open if you tap the clock icon.

https://user-images.githubusercontent.com/87384386/152588051-8e17307c-033c-4174-b082-7a68622ce4e8.mp4

MozillaNoah:
Other than the weird behavior quirks, this looks great

app/src/test/java/org/mozilla/fenix/library/history/HistorySearchControllerTest.kt
gabrielluong: I think in general we want to write all new tests using `GIVEN WHEN THEN`, but old habits are tough to break from since it's less verbose.

rocketsroger: Sounds good.  Bad habit.  I'll fix the test function names.  Thanks

app/src/main/java/org/mozilla/fenix/library/history/HistorySearchFragmentStore.kt
rocketsroger: Good catch, thanks.

app/src/main/java/org/mozilla/fenix/library/history/HistorySearchDialogFragment.kt
rocketsroger: Here's a good explanation I found:

"If a View that overrides onTouchEvent or uses an OnTouchListener does not also implement performClick and call it when clicks are detected, the View may not handle accessibility actions properly. Logic handling the click actions should ideally be placed in View#performClick as some accessibility services invoke performClick when a click action should occur."

I need it here is because when the user tap on anything but the toolbar, I want to close the toolbar and dismiss the keyboard.  But this code will take the user action away from the view.

app/src/main/java/org/mozilla/fenix/library/history/HistorySearchDialogFragment.kt
rocketsroger: Actually this is something I forgot to remove. I'll remove it.  Thanks,

app/src/main/res/layout/fragment_history_search_dialog.xml
rocketsroger: From Android developer: 
"A ViewStub is an invisible, zero-sized View that can be used to lazily inflate layout resources at runtime. When a ViewStub is made visible, or when [inflate()](https://developer.android.com/reference/android/view/ViewStub#inflate()) is invoked, the layout resource is inflated."

The reason I use it here is that we show search results if available.  If not it should be invisible.

rocketsroger:
> After you search any query, if you tap on the area below the search results, the search results disappear. Is this intentional?

Yes, this is intended.  Do you think this will confuse the user?  My thought is as long as they tap on anything but the search result, we should assume that they are not opening any URL.  I guess we could make it so we don't do anything.

rocketsroger:
> If you open the search bar and tap on the clock icon, the bar will minimize and remain at the bottom of the screen. Is this the intended behavior?

This is not what I intended, but more how Android decides if a keyboard should be shown.  When user taps on something that is not a input field the keyboard is dismissed.

MozillaNoah:
> > After you search any query, if you tap on the area below the search results, the search results disappear. Is this intentional?
> 
> Yes, this is intended. Do you think this will confuse the user? My thought is as long as they tap on anything but the search result, we should assume that they are not opening any URL. I guess we could make it so we don't do anything.

I think it _could_. Anecdotally, I sometimes tap on the screen in certain workflows so the keyboard is minimized so I have the whole screen real estate to scroll through.

rocketsroger:
> I think it _could_. Anecdotally, I sometimes tap on the screen in certain workflows so the keyboard is minimized so I have the whole screen real estate to scroll through.

Sure, to not block this PR, I've created an issue to track it.  https://github.com/mozilla-mobile/fenix/issues/23580.  Thanks

MozillaNoah (approved):

Issue #23614, opened 2022-02-07
Exception when running taskgraph generation locally with default parameters


ahal:
If you install the `taskcluster-taskgraph` package and run e.g:
```
$ taskgraph full
```
an exception is raised because we don't specify default values for some of Fenix's custom parameters. A workaround for this issue is to run:

```
$ taskgraph full -p taskcluster/test/params/main-repo-push.yml
```
instead. Fixing this will involve bumping Fenix up to the latest version of taskgraph.

PR #23615, opened 2022-02-07
For #23614 - Add default values for Fenix specific parameters

ahal:
Currently an exception will be raised if values for these parameters are not supplied. This will be the case when running with the default set of Parameters (e.g via `taskgraph full` without an explicit parameters file).

This PR adds defaults to these parameters (which also requires updating to the latest taskgraph).

### DIFF
--- a/taskcluster/fenix_taskgraph/parameters.py
+++ b/taskcluster/fenix_taskgraph/parameters.py
@@ -14,13 +14,24 @@ from voluptuous import All, Any, Optional, Range, Required
 
 from .release_promotion import read_version_file
 
+
+def get_defaults(repo_root):
+    return {
+        "pull_request_number": None,
+        "release_type": "",
+        "shipping_phase": None,
+        "next_version": "",
+        "version": "",
+    }
+
+
 extend_parameters_schema({
     Required("pull_request_number"): Any(All(int, Range(min=1)), None),
     Required("release_type"): text_type,
     Optional("shipping_phase"): Any('build', 'ship', None),
     Required("version"): text_type,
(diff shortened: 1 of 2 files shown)


ahal:
Guessing the docker image task failure may be a result of updateing `taskgraph`. Some investigation needed.

ahal:
> Guessing the docker image task failure may be a result of updateing `taskgraph`. Some investigation needed.

Nvm, I was just based on a really old revision that didn't have the `taskcluster` Python 2 fix yet.

taskcluster/fenix_taskgraph/parameters.py
```diff
@@ -14,13 +14,23 @@
...
+        "release_type": "",
+        "next_version": "",
+        "version": "",
+    }
+
+
 extend_parameters_schema({
     Required("pull_request_number"): Any(All(int, Range(min=1)), None),
     Required("release_type"): text_type,
     Optional("shipping_phase"): Any('build', 'ship', None),
```
jcristau, line 30: Does `shipping_phase` not need a default?

ahal: Apparently not? Maybe we use `config.params.get("shipping_phase")` everywhere. I guess it doesn't hurt to add one in anyway.

ahal:
Looks like Fenix is about to cut a new beta branch, so going to hold off landing this for the time being.

ahal:
Latest push updates revision to an even newer taskgraph which includes a fix for an issue we found yesterday due to unicode in commit message.

PR #23621, opened 2022-02-07
fixes #23619: use well-defined name for fetching remote wallpaper assets

### LINKED ISSUE #23619: Do not use simpleName to construct a URI in wallpapers
`simpleName` is not a safe value to use for constructing a URI.

https://github.com/mozilla-mobile/fenix/blob/b7b601eee1580bfb21a32e253e58b7eae016a31e/app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperDownloader.kt#L76-L78

We've had this bug before where the platform underneath can change the output of what we expect `simpleName` to be.

See: https://github.com/mozilla-mobile/fenix/issues/20830

cc: @Amejia481 @MatthewTighe

MatthewTighe:
1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperDownloader.kt
+++ b/app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperDownloader.kt
@@ -73,9 +73,12 @@ class WallpaperDownloader(
     private fun Wallpaper.Remote.toMetadata(context: Context): List<WallpaperMetadata> =
         listOf("landscape", "portrait").flatMap { orientation ->
             listOf("light", "dark").map { theme ->
-                val remoteParent = this::class.simpleName!!.lowercase()
                 val localPath = "wallpapers/$orientation/$theme/$name.png"
-                val remotePath = "${context.resolutionSegment()}/$orientation/$theme/$remoteParent/$name.png"
+                val remotePath = "${context.resolutionSegment()}/" +
+                    "$orientation/" +
+                    "$theme/" +
+                    "$remoteParentDirName/" +
+                    "$name.png"
                 WallpaperMetadata(remotePath, localPath)
             }
         }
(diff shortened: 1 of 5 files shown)


Amejia481 (approved):
LGTM!

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperDownloader.kt
```diff
@@ -73,9 +73,12 @@ class WallpaperDownloader(
     private fun Wallpaper.Remote.toMetadata(context: Context): List<WallpaperMetadata> =
         listOf("landscape", "portrait").flatMap { orientation ->
             listOf("light", "dark").map { theme ->
-                val remoteParent = this::class.simpleName!!.lowercase()
                 val localPath = "wallpapers/$orientation/$theme/$name.png"
-                val remotePath = "${context.resolutionSegment()}/$orientation/$theme/$remoteParent/$name.png"
+                val remotePath = "${context.resolutionSegment()}/" +
```
jonalmeida, line 77: Simple unit tests for this type of code can go a long way to avoid regressions in the future since content is dynamically downloaded. I'll leave that to you to file as a follow-up when time permits.

jonalmeida (approved):
LGTM from a code logic perspective. 🙂

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperDownloader.kt
MatthewTighe: I completely agree! The rate the requirements have changed has made it difficult to write this feature at the level of quality that I would normally expect of myself. Much of this code also clearly depends on platform types, which would mean structuring our dependencies in a more testable fashion. A more robust test suite is something I would like to include as part of [uplifting this feature to A-C](https://github.com/mozilla-mobile/android-components/issues/11679)

PR #23625, opened 2022-02-08
closes #23540: scale wallpaper and translate to bottom of image

### LINKED ISSUE #23540: Investigate ways to avoid sqashing or stretching the image for wallpapers
Some of the images have a different aspect ratio than the device. We should look at ways to avoid this. Letterboxing or cropping were suggested. Out of scope is trying to make sure specific parts of the image are visible on the screen.

MatthewTighe:
1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
@@ -391,14 +391,7 @@ class HomeFragment : Fragment() {
 
         FxNimbus.features.homescreen.recordExposure()
 
-        if (shouldEnableWallpaper()) {
-            val wallpaperManger = requireComponents.wallpaperManager
-            // We only want to update the wallpaper when it's different from the default one
-            // as the default is applied already on xml by default.
-            if (wallpaperManger.currentWallpaper != WallpaperManager.defaultWallpaper) {
-                wallpaperManger.updateWallpaper(binding.homeLayout, wallpaperManger.currentWallpaper)
-            }
-        }
+        displayWallpaperIfEnabled()
 
         // DO NOT MOVE ANYTHING BELOW THIS addMarker CALL!
         requireComponents.core.engine.profiler?.addMarker(
@@ -412,6 +405,7 @@ class HomeFragment : Fragment() {
         super.onConfigurationChanged(newConfig)
 
         getMenuButton()?.dismissMenu()
+        displayWallpaperIfEnabled()
     }
 
     private fun dismissTip(tip: Tip) {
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -1224,6 +1218,17 @@ class HomeFragment : Fragment() {
         }
     }
 
+    private fun displayWallpaper() {
```
rocketsroger, line 1221: nit: maybe name this `displayWallpaperIfEnabled`, or move the check to above before calling this function.

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperManager.kt
```diff
@@ -143,6 +145,39 @@ class WallpaperManager(
         }
     }.getOrNull()
 
+    private fun scaleBitmapToBottom(bitmap: Bitmap, view: ImageView) {
```
rocketsroger, line 148: I'm going to trust you on this calculation here.

rocketsroger (approved):
🚢

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -1242,7 +1247,8 @@ class HomeFragment : Fragment() {
     }
 
     private fun shouldEnableWallpaper() =
-        FeatureFlags.showWallpapers && !(activity as HomeActivity).themeManager.currentTheme.isPrivate
+        FeatureFlags.showWallpapers &&
+            (activity as? HomeActivity)?.themeManager?.currentTheme?.isPrivate?.not() ?: false
```
MatthewTighe, line 1251: This change was required in order to allow the `HomeFragmentTest::WHEN configuration changed menu is dismissed` test to pass, since that method this is now called on configuration change, and the activity cannot be cast to a `HomeActivity`.

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperManager.kt
MatthewTighe: Thanks, but I had some help from [this article](https://medium.com/mobile-app-development-publication/android-image-startcrop-and-endcrop-scaling-18d7297f1e7c). Should I maybe include that in a comment?

rocketsroger: Sure.  There's any example in `StrictModeManager`

PR #23655, opened 2022-02-09
Enable the wallpaper feature

Amejia481:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/FeatureFlags.kt
+++ b/app/src/main/java/org/mozilla/fenix/FeatureFlags.kt
@@ -5,10 +5,8 @@
 package org.mozilla.fenix
 
 import android.content.Context
-import android.os.StrictMode
 import mozilla.components.support.locale.LocaleManager
 import mozilla.components.support.locale.LocaleManager.getSystemDefault
-import org.mozilla.fenix.ext.components
 
 /**
  * A single source for setting feature flags that are mostly based on build type.
@@ -87,7 +85,7 @@ object FeatureFlags {
     /**
      * Enables showing the wallpaper functionality.
      */
-    val showWallpapers = Config.channel.isNightlyOrDebug
+    const val showWallpapers = true
 
     /**
      * Enables the Contile top sites.
@@ -103,11 +101,9 @@ object FeatureFlags {
      * Enables themed wallpapers feature.
      */
     fun isThemedWallpapersFeatureEnabled(context: Context): Boolean {
-        val strictMode = context.components.strictMode
-        return strictMode.resetAfter(StrictMode.allowThreadDiskReads()) {
-            val langTag = LocaleManager.getCurrentLocale(context)
-                ?.toLanguageTag() ?: getSystemDefault().toLanguageTag()
-            listOf("en-US", "es-US").contains(langTag) && Config.channel.isNightlyOrDebug
-        }
+        val langTag = LocaleManager.getCurrentLocale(context)
+            ?.toLanguageTag() ?: getSystemDefault().toLanguageTag()
+        return Config.channel.isNightlyOrDebug &&
+            listOf("en-US", "es-US").contains(langTag)
     }
 }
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mozilla/fenix/FeatureFlags.kt
```diff
@@ -87,7 +87,7 @@ object FeatureFlags {
     /**
      * Enables showing the wallpaper functionality.
      */
-    val showWallpapers = Config.channel.isNightlyOrDebug
```
MatthewTighe, line 90: I think we can remove this flag entirely now, right? By the time this branch hits release, we'll want it available to that channel as well

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperManager.kt
```diff
@@ -239,11 +239,7 @@ class WallpaperManager(
             Wallpaper.Local.Firefox("cerulean", R.drawable.cerulean),
             Wallpaper.Local.Firefox("sunrise", R.drawable.sunrise),
         )
-        private val remoteWallpapers: List<Wallpaper.Remote> = listOf(
-            Wallpaper.Remote.Focus(
-                "focus",
-            ),
-        )
+        private val remoteWallpapers: List<Wallpaper.Remote> = emptyList()
```
MatthewTighe, line 242: I thought we would leave this available in only nightly still, so that we could continue to gather usage data?

MatthewTighe (approved):
Had a couple small questions, but otherwise LGTM

app/src/main/java/org/mozilla/fenix/FeatureFlags.kt
Amejia481: Yeah, we can I just wanted to be extra careful in case we found something that we don't like and it won't go automatically to release. But lets be optimistic 🤞🏽. I'll updated it :)

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperManager.kt
Amejia481: Yes, we will. 
For instance, this PR is only introducing this commit on the branch [mozilla-mobile:releases_v98.0.0](https://github.com/mozilla-mobile/fenix/tree/releases_v98.0.0), that it's what we are using for Beta, and everything that we have in the [main](https://github.com/mozilla-mobile/fenix/tree/main) branch will be what we have on nightly :) 

image

When we uplift patches to other version we don't have to introduce the functionality first on `main` and later target branch (that will depend case by case, that will be the ideal situation) :)

MatthewTighe: Ah, I hadn't noticed the target, thanks for pointing that out. Do we merge release branches back into main at any point, to make sure the two are synced up?

Amejia481: > Ah, I hadn't noticed the target, thanks for pointing that out

No problem :) 

> Do we merge release branches back into main at any point, to make sure the two are synced up?

We don't. As we only uplift only certain bug and functionalities that we know that are safe. When the branch gets stale or a new version arrive ([merge day](https://github.com/mozilla-mobile/fenix/wiki/Creating-a-release-branch)) we just create a new branch that it's based of main.

app/src/main/java/org/mozilla/fenix/FeatureFlags.kt
```diff
@@ -103,11 +101,9 @@ object FeatureFlags {
      * Enables themed wallpapers feature.
      */
     fun isThemedWallpapersFeatureEnabled(context: Context): Boolean {
-        val strictMode = context.components.strictMode
-        return strictMode.resetAfter(StrictMode.allowThreadDiskReads()) {
```
Amejia481, line 107: @MatthewTighe I had to remove strict mode ignore, as it was causing most of the tests to fail [see](https://github.com/mozilla-mobile/fenix/runs/5132271664?check_suite_focus=true), this is not visible in main as are not checking for `isThemedWallpapersFeatureEnabled` on [FenixApplication.kt](https://github.com/mozilla-mobile/fenix/pull/23655/files#diff-9e2b3999e8d0e181af6decd7d25949eef361031a719a578869967a5339b31737), I will continue further exploring tomorrow.

Amejia481: Tested locally, this is a safe approach.

PR #23663, opened 2022-02-09
closes #23662: replace wallpaper thumbnail lazyverticalgrid with manual scroll layout

### LINKED ISSUE #23662: Wallpaper settings screen should be scrollable
The wallpaper settings screen is not currently scrollable. In landscape mode, this can cut off the logo switch and the text describing the switch. Update the settings screen to be scrollable.

MatthewTighe:
1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayMenu.kt
+++ b/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayMenu.kt
@@ -54,14 +54,14 @@ class TabsTrayMenu(
         listOf(
             SimpleBrowserMenuItem(
                 context.getString(R.string.tabs_tray_select_tabs),
-                textColorResource = R.color.primary_text_normal_theme
+                textColorResource = R.color.fx_mobile_text_color_primary
             ) {
                 onItemTapped.invoke(Item.SelectTabs)
             }.apply { visible = shouldShowSelectOrShare },
 
             SimpleBrowserMenuItem(
                 context.getString(R.string.tab_tray_menu_item_share),
-                textColorResource = R.color.primary_text_normal_theme
+                textColorResource = R.color.fx_mobile_text_color_primary
             ) {
                 context.components.analytics.metrics.track(Event.TabsTrayShareAllTabsPressed)
                 onItemTapped.invoke(Item.ShareAllTabs)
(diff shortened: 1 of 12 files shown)


app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -169,18 +176,29 @@ private fun WallpaperThumbnails(
...
-        ) {
-            items(wallpapers) { wallpaper ->
-                WallpaperThumbnailItem(
-                    wallpaper = wallpaper,
-                    defaultWallpaper = defaultWallpaper,
-                    loadWallpaperResource = loadWallpaperResource,
-                    isSelected = selectedWallpaper == wallpaper,
-                    onSelect = onSelectWallpaper
-                )
+        Column(modifier = Modifier.padding(vertical = 30.dp, horizontal = 20.dp)) {
```
MozillaNoah, line 179: Big sad we have to do a grid manually. I'm assuming wrapping `LazyVerticalGrid` in a `Column` didn't work?

MatthewTighe: It's apparently an [error in Compose](https://stackoverflow.com/questions/66908737/what-is-the-equivalent-of-nestedscrollview-recyclerview-or-nested-recyclerv) to nest `Lazy` layouts within other scrollable layouts. 

Compose 1.1 introduces a [span parameter to items](https://developer.android.com/reference/kotlin/androidx/compose/foundation/lazy/LazyGridScope#item(kotlin.Any,kotlin.Function1,kotlin.Any,kotlin.Function1)) in a `LazyGridScope`. I experimented with that, but it would have also been awkward as the trailing item has a different background color and padding considerations. Plus I didn't want to handle the upgrade just for this

gabrielluong: I am curious what is the layout we are trying to get here? Got a link to to figma? It's only an error if we're nesting scroll layouts in the same direction.

Amejia481 (approved):

MatthewTighe:
@Mergifyio backport releases_v98.0.0

Issue #23665, opened 2022-02-10
[Bug] Undo snackbar not working after leaving the screen it originated from


mavduevskiy:
Steps to reproduce: delete an item from download, bookmarks and history screens, the snackbar with an undo action will popup. Go back to the main screen – Snackbar is there, but pressing the Undo button doesn't actually work, the item is gone.

SoftVision-LorandJanos:
Managed to reproduce on a earlier build (Nightly from 2022-02-10).
Verified as fixed on the latest Nightly (2022-04-28).
The Undo button works as expected on the previous screen.
Device used: Google Pixel 4 (Android 12).
Closing the ticket as fixed.

Issue #9539, opened 2020-03-31
[Bug] Can't Use Address Bar while Snackbar is Displayed


theFireProject:
## Steps to reproduce
**Scenarios**
This bugs only occurs for specific snackbars, not for all. From what I've found out, the bug occurs at least for these two:
- "URL copied to clipboard" snackbar
    - go to a website
    - tap and hold the address on the address bar
    - tap 'Copy' to copy the address
- "Deleted «top-level domain»" snackbar
    - go to a website
    - save it as a bookmark
    - edit the saved bookmark
    - delete the bookmark
- try to tap the address to edit it or try to tap another button on the address bar

The bug does not take effect for these snackbars:
- "Bookmark saved! 'EDIT'" snackbar
- "New tab opened 'SWITCH'" snackbar
- "Tab closed 'UNDO'" snackbar
- "Added to top sites!" snackbar

### Expected behavior
After copying the address, the confirmation snackbar is displayed saying "URL copied to clipboard".
It's displayed above the address bar without overlapping it. So the address bar is fully accessible in theory. Tapping the address field is supposed to open the keyboard to edit the URL and tapping other buttons on the address bar should trigger the right action, too.

### Actual behavior
The address bar is not useable while the confirmation snackbar "URL copied to clipboard" is displayed. There is no reaction to tapping anywhere on the address bar.

### Device information

* Android device: Samsung Galaxy S9+ (Android 8.0.0)
* Fenix version: 4.2.0

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-16533)

LaurentiuApahideanSV:
I can confirm that this issue is reproducible using the latest version of Firefox Preview Nightly 3/31 (Build #20910607).

Devices:
OnePlus 6T (Android 9);
Huawei MediaPad M3 lite 10 (Android 7.0)

sblatz:
This is happening in cases where we display the snackbar on `ContentFrameLayout` (which we need to do in order to display it while navigating to another fragment). I'm not sure what options we really have here other than getting rid of those types of snackbars.

SL-Lee:
As of **Nightly 200722 06:08 (Build #22040614)**, I have discovered three more snackbars that not only obstructs the address bar while it is displayed, it also cannot be dismissed by swiping right on it:

[...]

PR #23750, opened 2022-02-15
For #23665 - Dismiss undo snackbar on fragment pause

Alexandru2909:
Changed the behavior of the `Undo` snackbar to be dismissed inside the `onPause` of the history, bookmark and download fragments.

https://user-images.githubusercontent.com/35462038/154079582-69fc4a53-ccde-4900-8073-6aa504862419.mp4

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkFragment.kt
@@ -67,6 +67,7 @@ class BookmarkFragment : LibraryPageFragment<BookmarkNode>(), UserInteractionHan
     private val sharedViewModel: BookmarksSharedViewModel by activityViewModels()
     private val desktopFolders by lazy { DesktopFolders(requireContext(), showMobileRoot = false) }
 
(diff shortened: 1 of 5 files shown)


mavduevskiy:
@Alexandru2909  
This is a solid solution, but removing snackbar as soon as user leaves the screen is a big UX change, which reduces the value of the undo snackbars. We want to give the user a few seconds to undo his action, and there is some code in the app that is supposed to do that. The better way would be to improve what is not working, not changing the feature.

Currently, the issue was that there was no access to the context of the fragment (and consequently, to the stores) after user left the fragment, so the undo actually never worked. Also, there was a mechanism that would trigger removal of the item in a few cases, including leaving the fragment. That seems more as a hack than a good solution. It is not difficult to actually access stores after fragment's context is no more, via the context app, and allow snackbars to actually work.

I have been actually working on the issue for quite some time, but I forgot to assign the issue to myself. I run into a different issue, that snackbar would sometimes get hidden, or not lifted up or [blocking](https://github.com/mozilla-mobile/fenix/issues/9539) the ui beneath it when we go back. 

I am sorry, my bad not assigning it to myself.

PR #23781, opened 2022-02-16
For #23738 - Integrate the Contile top sites updater

### LINKED ISSUE #23738: Integrate the Contile top sites updater

gabrielluong:
Fixes #23738. Depends on https://github.com/mozilla-mobile/android-components/pull/11721

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
+++ b/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
@@ -57,6 +57,7 @@ import mozilla.components.feature.search.BrowserStoreSearchAdapter
 import mozilla.components.service.fxa.sync.SyncReason
 import mozilla.components.support.base.feature.ActivityResultHandler
 import mozilla.components.support.base.feature.UserInteractionHandler
+import mozilla.components.support.base.log.logger.Logger
 import mozilla.components.support.ktx.android.arch.lifecycle.addObservers
 import mozilla.components.support.ktx.android.content.call
 import mozilla.components.support.ktx.android.content.email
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -332,7 +336,18 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
             }
         }
 
-        isFenixTheDefaultBrowser()
+        lifecycleScope.launch(IO) {
+            components.core.contileTopSitesProvider.getTopSites(
+                context = applicationContext,
+                allowCache = true
```
gabrielluong, line 342: @csadilek I am not quite sure if the idea was to call this every time onResume and fetch when the max cache age is hit or if we wanted to expose methods to check if the cache is expired and then do a fetch.

csadilek: @gabrielluong I think this OK, as internally the method doesn't fetch when the cache isn't expired (older than 1 hour). But it does cause an unnecessary disk read from the cache. We're not actually using the result here....so I think I ideally we should have a `contileTopSitesProvider.maybeRefreshTopSites`method that simply does nothing if the cache isn't expired. What do you think? We would still launch on IO but this way we can prevent the unnecessary read if the cache is still fresh.

gabrielluong: Yes, I actually prefer not doing the disk read if possible.

csadilek: OK, agreed, esp. as we have to do this `onResume` which happens a lot. Let's prevent those reads. We should also keep the `cacheLastModified` in a field of `ContileTopSitesProvider` and only read the date from the file when it's null (e.g. after a cold start). And of course, update/reset it after we write to the cache. This will improve this as well. No IO at all `onResume` unless we have to fetch.

Issue #23752, opened 2022-02-15
[Bug]: "Change wallpaper" switch off screen


plwt:
### Steps to reproduce

1. Open Firefox for Android menu
2. Select Customise Homepage
3. Select Wallpapers

### Expected behaviour

"Change wallpaper by tapping Firefox homepage logo" switch should be fully visible.

### Actual behaviour

![Screenshot_20220215-142250](https://user-images.githubusercontent.com/31939113/154082226-7c0a2717-5c94-4d74-b17a-ea3b13bb3520.png)

### Device name

Motorola Moto G30

### Android version

Android 11

### Firefox release type

Firefox Nightly

### Firefox version

99.0a1

cadeyrn:
I can confirm the issue. It's not only not fully visible, the toggle is also not functional at all.

plwt:
Although not fixed in Nightly, it is fixed in Beta (both position and function).

SoftVision-LorandJanos:
Hi @plwt ! Thanks for your report!
I can confirm the issue is reproducible as you described.
Tested on the latest Nightly 99.0a1 (2022-02-16) build.
Device used: Oppo Reno 6 (Android 11).

Amejia481:
A fix landed on `main` it should be in the next nightly update!

PR #23787, opened 2022-02-16
For #23752: Address switch is off the screen in the Wallpaper settings.

Amejia481:
image

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
@@ -68,8 +68,7 @@ import java.util.Locale
  * @param wallpapers Wallpapers to add to grid.
  * @param selectedWallpaper The currently selected wallpaper.
  * @param defaultWallpaper The default wallpaper
- * @param loadWallpaperResource Callback to handle loading a wallpaper bitmap. Only optional in the
- *   default case.
+ * @param loadWallpaperResource Callback to handle loading a wallpaper bitmap. Only optional in the default case.
  * @param onSelectWallpaper Callback for when a new wallpaper is selected.
  * @param onViewWallpaper Callback for when the view action is clicked from snackbar.
  * @param tapLogoSwitchChecked Enabled state for switch controlling taps to change wallpaper.
@@ -164,8 +163,7 @@ private fun WallpaperSnackbar(
  *
  * @param wallpapers Wallpapers to add to grid.
  * @param defaultWallpaper The default wallpaper
- * @param loadWallpaperResource Callback to handle loading a wallpaper bitmap. Only optional in the
- *   default case.
+ * @param loadWallpaperResource Callback to handle loading a wallpaper bitmap. Only optional in the default case.
  * @param selectedWallpaper The currently selected wallpaper.
  * @param numColumns The number of columns that will occupy the grid.
  * @param onSelectWallpaper Action to take when a new wallpaper is selected.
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -269,33 +269,35 @@ private fun WallpaperThumbnailItem(
...
-                    checkedTrackColor = FirefoxTheme.colors.formSurface,
-                    uncheckedTrackColor = FirefoxTheme.colors.formSurface
-                )
+        Text(
+            text = stringResource(R.string.wallpaper_tap_to_change_switch_label_1),
+            color = FirefoxTheme.colors.textPrimary,
+            fontSize = 18.sp,
+            modifier = Modifier
+                .padding(start = 4.dp)
+                .weight(0.8f)
```
Amejia481, line 289: Just to highlighting, this change as the this maybe a bit difficult to read.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -269,33 +269,35 @@ private fun WallpaperThumbnailItem(
...
+        )
+        Switch(
+            modifier = Modifier.weight(0.2f),
```
Amejia481, line 292: Just to highlighting, this change as the this maybe a bit difficult to read.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -269,33 +269,35 @@ private fun WallpaperThumbnailItem(
...
-                    checkedThumbColor = FirefoxTheme.colors.formSelected,
-                    checkedTrackColor = FirefoxTheme.colors.formSurface,
-                    uncheckedTrackColor = FirefoxTheme.colors.formSurface
-                )
+        Text(
+            text = stringResource(R.string.wallpaper_tap_to_change_switch_label_1),
+            color = FirefoxTheme.colors.textPrimary,
+            fontSize = 18.sp,
+            modifier = Modifier
+                .padding(start = 4.dp)
```
gabrielluong, line 288: Should we also add some end padding here to balance it here? We already enforce this in xml. The risk here is that long strings can run to the 'edge' with no spacing between it and the Switch.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -269,33 +269,35 @@ private fun WallpaperThumbnailItem(
...
 @Composable
+@Suppress("MagicNumber")
 private fun WallpaperLogoSwitch(
     checked: Boolean,
     onCheckedChange: (Boolean) -> Unit
 ) {
-    Column(
-        modifier = Modifier.padding(horizontal = 12.dp, vertical = 16.dp),
+    Row(
+        horizontalArrangement = Arrangement.SpaceBetween,
```
gabrielluong, line 278: Can we move this below `modifier` just so we are following the parameter ordering.

gabrielluong (approved):

gabrielluong:
I am not quite sure if the commit/title makes any sense.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
gabrielluong: While I am here, it's interesting that we have padding defined in the parent `Row` and padding here. Without knowing what the designs look like, I almost feel like we can just add this padding to the Row from 12.dp to 16.p

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -269,33 +269,35 @@ private fun WallpaperThumbnailItem(
 }
 
 @Composable
+@Suppress("MagicNumber")
```
gabrielluong, line 272: Is the weights the magic number here?

Amejia481:
> I am not quite sure if the commit message/title makes any sense.

Sorry my bad updating it!

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
Amejia481: Yes

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
MozillaNoah: Alternatively, just the `Text()` could have the weight, and it will always grow to the maximum remaining horizontal space.

Amejia481: 👍🏽

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
gabrielluong: I think you have want to remove the `weight` here. I saw that the Switch not fully aligned to the end in my testing if we kept it.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -269,33 +269,35 @@ private fun WallpaperThumbnailItem(
...
-                )
+        Text(
+            text = stringResource(R.string.wallpaper_tap_to_change_switch_label_1),
+            color = FirefoxTheme.colors.textPrimary,
+            fontSize = 18.sp,
+            modifier = Modifier
+                .padding(horizontal = 4.dp)
+                .weight(0.8f)
+        )
+        Switch(
```
gabrielluong, line 291: Can you add a new line between the 2 components?

gabrielluong:
While we are touching this file, can you also remove the extra indents in the kDoc in the @param on line 72 and 171. We don't typically indent the extra lines for a @param so I rather not encourage this.

gabrielluong (changes requested):
-

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
Amejia481: Addressed!

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
Amejia481: Addressed!

gabrielluong (approved):

Amejia481:
> Gonna block on changes just in case

I addressed the suggestions above, when you have some available time, please take a look :)

PR #23805, opened 2022-02-17
Close #23804: Use version instead of date for telemetry/metrics expiration

### LINKED ISSUE #23804: Use version instead of date for telemetry/metrics expiration
We currently use a fixed date for telemetry expiration which has numerous downsides:

    Pings can expire anytime within a release cycle. We had this happen for search telemetry which dropped off a few days after a dot release. This prompted the change to set business critical metrics to never expire.

    It's technically impossible to achieve reproducible builds using expiration by date. There's always a point in time at which a release/tag can no longer be checked out and shipped. Telemetry tests would fail, or telemetry would simply not be sent in case there's no test coverage. And urgent dot release for a security or stability issue would then require going through completely unrelated telemetry renewal before it can be shipped.

    Telemetry renewals are difficult to carry out as the expiration dates aren't aligned with release dates.

Switching to expiration by version addresses the concerns above. Firefox Desktop already switched to version based expiry and this is supported by Glean. 

See also: 
https://bugzilla.mozilla.org/show_bug.cgi?id=1724760
https://firefox-source-docs.mozilla.org/toolkit/components/glean/user/new_definitions_file.html#how-does-expiry-work

rocketsroger:
This change is created by using a script with the following conversion table
_CONVERSION = {
  '    expires: "2022-05': '    expires: 101',
  '    expires: "2022-06': '    expires: 102',
  '    expires: "2022-07': '    expires: 103',
  '    expires: "2022-08': '    expires: 104',
  '    expires: "2022-09': '    expires: 105',
  '    expires: "2022-10': '    expires: 106',
  '    expires: "2022-11': '    expires: 107',
  '    expires: "2022-12': '    expires: 108',
  '    expires: "2023-01': '    expires: 109',
  '    expires: "2023-02': '    expires: 111',
  '    expires: "2023-03': '    expires: 112',
  '    expires: "2023-04': '    expires: 113',
  '    expires: "2023-05': '    expires: 114',
  '    expires: "2023-06': '    expires: 115',
  '    expires: "2023-07': '    expires: 116',
  '    expires: "2023-08': '    expires: 117',
  '    expires: "2023-09': '    expires: 118',
  '    expires: "2023-10': '    expires: 119',
  '    expires: "2023-11': '    expires: 121',
  '    expires: "2023-12': '    expires: 122',
  '    expires: "2024-02': '    expires: 123',
  '    expires: "2024-03': '    expires: 124',
  '    expires: "2024-04': '    expires: 125',
  '    expires: "2024-05': '    expires: 126',
  '    expires: "2024-06': '    expires: 127',
  '    expires: "2024-07': '    expires: 128',
}

### Pull Request checklist

1. click on Show All Checks,


app/metrics.yaml
```diff
@@ -157,7 +157,7 @@ events:
       - interaction
     notification_emails:
       - android-probes@mozilla.com
-    expires: "2022-12-01"
+    expires: 108
```
csadilek, line 160: Glean docs say "X" (where X is the major version...), and use expires: "88" as an example. Is it fine to skip the quotes?

csadilek (commented):
Thanks for moving this forward! Let's reach out to the Glean team for a review here to double-check.

app/metrics.yaml
rocketsroger: Yes, we don't need to put the quotes around the version number.  I got the example from the Glean team here: https://github.com/mozilla/glean/blob/main/samples/android/app/metrics.yaml

rocketsroger:
> Thanks for moving this forward! Let's reach out to the Glean team for a review here to double-check.

Thanks! I'll add the Glean team to the review.

app/metrics.yaml
csadilek: ok, perfect!

badboy: Woops, guess we didn't update the docs then

Dexterp37:
Removing me from the reviewers list, @badboy is taking care of it. Please make sure he gets to review _before_ merging :-)

app/build.gradle
```diff
@@ -836,3 +836,6 @@ android.applicationVariants.all { variant ->
         dependsOn "package${variant.name.capitalize()}"
     }
 }
+
+// Enable expiration by major version.
+ext.gleanExpireByVersion = 1
```
Dexterp37, line 841: This will expire all your metrics: you need to provide the major version of Fenix here. See the [Glean docs](https://mozilla.github.io/glean/book/language-bindings/android/android-build-configuration-options.html#gleanexpirebyversion)

Dexterp37 (changes requested):

app/build.gradle
rocketsroger: I see, this version tells Glean what is the major version.  So with 1 set then no metrics will expire, correct?

Dexterp37: @rocketsroger correct, but I strongly discourage you from landing it as it is: you should really provide the correct Fenix version in here, within this PR. If you want non-expiring metrics, you can make non-expiring metrics via the metrics.yaml (by requesting the proper data-review).

rocketsroger: @Dexterp37 yes, I won't land this until I update the patch.  Just want to make sure I understand this correctly.  The version number we pass in here is the version number Glean will compare against the metrics `expires:` field correct?  So if we pass in `ext.gleanExpireByVersion = 99` then anything below 100 will expire but metrics that `expires: 100` should still work.  Correct?

app/build.gradle
```diff
@@ -836,3 +836,6 @@ android.applicationVariants.all { variant ->
         dependsOn "package${variant.name.capitalize()}"
     }
 }
+
+// Enable expiration by major version.
+ext.gleanExpireByVersion = Config.androidComponentsMajorVersionName()
```
rocketsroger, line 841: Not sure if this is the best way to get the major version.  It is the simplest way though.

app/build.gradle
Dexterp37: `gleanExpireByVersion = 99` means that Fenix is version 99.x.x so all the metrics bound to expire in versions <= 99 will be expired.

app/build.gradle
csadilek: @rocketsroger I think we can add a method specifically for `Fenix` that will work with both the Nightly and Release versioning schemes. Happy to help with that.

Dexterp37 (approved):
This looks good to me from a data collectio POV. Please make sure a Fenix peer reviews the way you are fetching the major version before merging :-)

csadilek (approved):
Tested with Debug/Nightly and Release builds. All looks good to me.

One TODO we have left is to update the wiki and document our new renewal process, as discussed. On merge day, when we upgrade to the next Nightly version, we want to trigger renewal for metrics expiring in that version + 1.

Dexterp37:
> One TODO we have left is to update the wiki and document our new renewal process, as discussed. On merge day, when we upgrade to the next Nightly version, we want to trigger renewal for metrics expiring in that version + 1.

@csadilek that's interesting, can I hear more about why do you have this process in place? Will you automatically renew all metrics like that or just the subset you'd still need? Why bump just by one version? (did you consider making them never expiring?)

csadilek:
@Dexterp37 We're not bumping by one version, that was ambiguous/incorrect wording on my part after Roger and I had already talked in more detail. We will start **reviewing** (not _renewing_ automatically) all metrics that expire in that version + 1. So when we switch to a new Nightly version we will also look at the metrics that expire in this version and the next to make sure nothing falls through. The rest stays the same. We will still check with product what's still needed and start renewal where needed.

Issue #21743, opened 2021-10-06
Replace deprecated Patterns.IP_ADDRESS 


pocmo:
Follow-up from: https://github.com/mozilla-mobile/fenix/pull/21738

`Patterns.IP_ADDRESS` is deprecated in the Android 12 SDK.

> Please use InetAddresses.isNumericAddress(String) instead. 
> Kept for backward compatibility reasons. It does not match IPv6 addresses.

In our case that may just be okay and we could just suppress indefinitely?

https://github.com/mozilla-mobile/fenix/blob/0b6e67a21b28b1d4fbf7b2ec92d1cd1b35a684f3/app/src/main/java/org/mozilla/fenix/ext/String.kt#L83-L85

mcarare:
Closed in https://github.com/mozilla-mobile/fenix/pull/23859.

Amejia481:
🎉 !

PR #23859, opened 2022-02-21
For #21743: Use InetAddresses isNumericAddress to validate Ipv4 and Ipv6 for SDK>=29.

mcarare:
### Pull Request checklist

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/ext/String.kt
```diff
@@ -71,15 +74,45 @@ fun String.toShortUrl(publicSuffixList: PublicSuffixList): String {
...
 
-// impl via FFTV https://searchfox.org/mozilla-mobile/source/firefox-echo-show/app/src/main/java/org/mozilla/focus/utils/FormattedDomain.java#129
-@Suppress("DEPRECATION")
-fun String.isIpv4(): Boolean = Patterns.IP_ADDRESS.matcher(this).matches()
+/**
+ * Returns true if the string represents a valid Ipv4 IP address.
+ */
+fun String.isIpv4(): Boolean {
+    return try {
+        val hostAddress = Inet4Address.getByName(this).hostAddress
```
csadilek, line 82: It looks like this may now resolve the name and need network access or am I misreading this? I think that's a problem, because this is called via `toShortUrl` from a lot of places in the UI.

csadilek: (So yes, if this hits the network, we can't do this for the current use cases without bigger changes.)

mcarare: AFAIK these methods work even without network access.

csadilek: Yes, but my concerns is about performance of these methods if there's no or a slow network connection. That's what we have to figure out so we don't end up blocking the main thread.

csadilek: Quick research on `lookupAllHostAddr` shows multiple cases where it could block. It's called from `getByName`, but needs more investigation/testing to confirm.

mcarare: I have already tested this on a real device for actions from call sites of the `toShortUrl` method. (deleting a history entry, for example, displaying short URL in other places) I did not notice any visual performance issues.

I Have not analyzed a profile, but using `measureTimeMillis` shows that the first call to the new `isIpv4` method is 5, 6 times faster than the previous implementation. Of course, the next ones are comparable (because the old implementation used a pattern).


```
2022-02-22 23:09:27.687 22005-22005/org.mozilla.fenix.debug E/mcarare check Old: 6
2022-02-22 23:09:27.688 22005-22005/org.mozilla.fenix.debug E/mcarare check New: 1

2022-02-22 23:09:33.461 22005-22005/org.mozilla.fenix.debug E/mcarare check Old: 1
2022-02-22 23:09:33.462 22005-22005/org.mozilla.fenix.debug E/mcarare check New: 1

2022-02-22 23:09:39.636 22005-22005/org.mozilla.fenix.debug E/mcarare check Old: 1
2022-02-22 23:09:39.637 22005-22005/org.mozilla.fenix.debug E/mcarare check New: 1

2022-02-22 23:09:46.952 22005-22005/org.mozilla.fenix.debug E/mcarare check Old: 1
2022-02-22 23:09:46.953 22005-22005/org.mozilla.fenix.debug E/mcarare check New: 1
```

For `isIpv6` with an Ipv4 address:
```
2022-02-22 23:17:48.293 23203-23203/org.mozilla.fenix.debug E/mcarare check Old: 3
2022-02-22 23:17:48.294 23203-23203/org.mozilla.fenix.debug E/mcarare check New: 1
2022-02-22 23:17:53.814 23203-23203/org.mozilla.fenix.debug E/mcarare check Old: 0
2022-02-22 23:17:53.815 23203-23203/org.mozilla.fenix.debug E/mcarare check New: 1
2022-02-22 23:18:02.845 23203-23203/org.mozilla.fenix.debug E/mcarare check Old: 0
2022-02-22 23:18:02.845 23203-23203/org.mozilla.fenix.debug E/mcarare check New: 0
```

For `isIpv6` with an Ipv6 address that also needed `replaceConsecutiveZeros`to validate:

```
2022-02-22 23:19:56.309 23615-23615/org.mozilla.fenix.debug E/mcarare check Old: 2
2022-02-22 23:19:56.310 23615-23615/org.mozilla.fenix.debug E/mcarare check New: 1
2022-02-22 23:20:03.601 23615-23615/org.mozilla.fenix.debug E/mcarare check Old: 0
2022-02-22 23:20:03.603 23615-23615/org.mozilla.fenix.debug E/mcarare check New: 1
2022-02-22 23:20:10.858 23615-23615/org.mozilla.fenix.debug E/mcarare check Old: 1
2022-02-22 23:20:10.859 23615-23615/org.mozilla.fenix.debug E/mcarare check New: 1

````

As a note, there is no difference in these values when running this on a device conected to wi fi or not connected at all.

Comment (csadilek) [reply app/src/main/java/org/mozilla/fenix/ext/String.kt:82]: @mcarare That's good to know, thanks, but it's not clear to me if this was faster because of caching e.g., what would happen in case of a cache miss and a slow network?

/cc @mcomella do you have any insights, or an opinion here. I am concerned about potential network requests on the main thread.

Comment (mcomella) [reply app/src/main/java/org/mozilla/fenix/ext/String.kt:82]: I don't think it's possible for these methods to work without network access or local caching of those requests. In order to map a host name to an IP address, you need to access a DNS server and, since these methods resolve synchronously, they presumably do that on the thread you execute them on.

If this does trigger a main thread network request, on debug builds we should see a StrictMode violation and crash. Can we try that? We should use a hostname that isn't likely to already be cached across the system (e.g. don't use google.com).

Comment (mcarare) [reply app/src/main/java/org/mozilla/fenix/ext/String.kt:82]: @mcomella I have already tested this. These are the findings:

- For numeric addresses, no network calls are made the code just checks for validity.
- For machine names, `UnknownHostException` is thrown. No crash is happening. The logcat does register a NetworkViolation: 

```
2022-02-24 08:38:06.893 13008-13008/org.mozilla.fenix.debug D/StrictMode: StrictMode policy violation; 

~duration=11 ms: android.os.strictmode.NetworkViolation
        at android.os.StrictMode$AndroidBlockGuardPolicy.onNetwork(StrictMode.java:1673)
```

mcarare: This is the API code used: https://android.googlesource.com/platform/libcore/+/master/ojluni/src/main/java/java/net/Inet6AddressImpl.java#113

mcarare: @csadilek We could fall back to keeping the code for versions >=29  (`InetAddresses.isNumericAddress(this)` does not perform a DNS lookup ) and keep the old methods for below. Is that ok with you?

csadilek: @mcarare yes, that sounds good to me! We'd no longer be calling `Inet*Address.getByName()` then so that seems fine.

csadilek (approved):
👍

PR #23908, opened 2022-02-22
Close #23894: Simplify distinct history algorithm

### LINKED ISSUE #23894: [Bug]: History is truncated if `FeatureFlag.historyImprovementFeatures` is turned on
### Steps to reproduce

1. Open Fenix with profile that has multiple (>5) pages of history visits
1. Navigate to three dot menu -> History
2. Scroll down

### Expected behaviour

User can scroll and search through entire history.

### Actual behaviour

Scrolling only works for up to 5 pages (note that searching works fine.)

### Device name

Pixel 3a

### Android version

Android 12

### Firefox release type

Firefox Nightly

### Firefox version

99

### Device logs

_No response_

### Additional information

This only reproduces on Nightly for me, so thinking it's related to the `historyImprovementFeatures` feature flag.

rocketsroger:
With the current algorithm of only displaying history when it is distinct in the current group, sometimes we will (unluckly) get 0 results because all the history we just retrieved are all duplicated in the same group.  When that happens it causes the adapter to think that there is no more history to be displayed.  The simplest solution is to just make sure that for every retrieval we show only the distinct results from each group.  This way we don't have to keep track of the history items that have been shown (which could be in thousands) and will always return something if there is history.

### Pull Request checklist

1. click on Show All Checks,


csadilek (approved):
Tested locally and verified this fixes the problem.

PR #23973, opened 2022-02-24

MickeyMoz:


rvandermeulen:
@Mugurell We're seeing failures related to the AC Kotlin update.

rvandermeulen:
The failures are due to a wider Gradle outage, not specific to this PR.

PR #23992, opened 2022-02-25
No issue: Add telemetry data renewal helper

rocketsroger:
### Pull Request checklist

1. click on Show All Checks,


tools/data_renewal_generate.py
```diff
@@ -0,0 +1,149 @@
+import os
```
gabrielluong, line 1: Should this have a license?

rocketsroger: Yes, I'll add it.  Thanks

gabrielluong:
Should add comments about how to run this script in this file or somewhere else? Either way, I think we should have a comment that directs to wherever the instructions should be.

tools/data_renewal_generate.py
```diff
@@ -0,0 +1,149 @@
...
+            total_count += 1
+
+            # name of the telemtry
+            result = {"#": total_count, "name" : last_key.lstrip('.')}
+            result.update(content)
+
+            # add columns for product to fille out
+            result.update({"keep(Y/N)" : ""})
+            result.update({"new expiry version" : ""})
+            result.update({"reason to extend" : ""})
```
jonalmeida, line 52: Why are these updates instead of added to the `result` binding when it's created a few lines above?

tools/data_renewal_generate.py
jonalmeida, line 1: Specify the env to run with py3.

```suggestion
#!/usr/bin/env python3

import os
```

tools/data_renewal_generate.py
```diff
@@ -0,0 +1,149 @@
...
+    if os.path.exists(new_metrics_filename):
+        print("remove old metrics yaml file")
+        os.remove(new_metrics_filename)
+
+    data_file = open(csv_filename, 'w')
+    csv_writer = csv.writer(data_file)
+    renewal_file = open(renewal_filename, 'w')
+
+    response("", content, current_version, csv_writer, renewal_file)
+    renewal_file.close
```
jonalmeida, line 120: The call won't execute with the parenthesis.

```suggestion
    renewal_file.close()
```

tools/data_renewal_generate.py
```diff
@@ -0,0 +1,149 @@
...
+    print("Total count: " + str(total_count))
+
+    # Go through the metrics.yaml file to mark expired telemetry
+    verify_count = 0
+    f.seek(0, 0)
+    data = f.readlines()
+    with open(new_metrics_filename, 'w') as f2:
+        for line in data:
+            if (line.lstrip(' ').startswith("expires: ") and not(line.lstrip(' ').startswith("expires: never"))):
+               version = int(line.lstrip(' ')[9:])
```
jonalmeida, line 131: Maybe leave a comment where you got this magic number from or calculate it:
```suggestion
               start_pos = len("expires: ")
               version = int(line.lstrip(' ')[start_pos:])
```

tools/data_renewal_generate.py
```diff
@@ -0,0 +1,149 @@
...
+               if (version <= current_version):
+                   verify_count += 1
+                   f2.writelines(line.rstrip('\n') + " /* TODO <" + str(verify_count) + "> require renewal */\n")
+               else:
+                   f2.writelines(line)
+            else:
+                f2.writelines(line)
+        f2.close
```
jonalmeida, line 139: Same as above.

tools/data_renewal_generate.py
jonalmeida: Do you also need to try-catch the `ValueError` if the cast can fail? Are we leaving any existing telemetry to be time-based?

tools/data_renewal_generate.py
```diff
@@ -0,0 +1,149 @@
...
+    verify_count = 0
+    f.seek(0, 0)
+    data = f.readlines()
+    with open(new_metrics_filename, 'w') as f2:
+        for line in data:
+            if (line.lstrip(' ').startswith("expires: ") and not(line.lstrip(' ').startswith("expires: never"))):
+               version = int(line.lstrip(' ')[9:])
+               if (version <= current_version):
+                   verify_count += 1
+                   f2.writelines(line.rstrip('\n') + " /* TODO <" + str(verify_count) + "> require renewal */\n")
```
jonalmeida, line 134: On each probe, we write some like this: 
```yaml
  expires:
    /* TODO <5> require renewal */
```

Is the number a counter and ID?

jonalmeida (commented):
Looks good! I haven't run it locally, but I left a few comments and some questions that were unclear to me.

tools/data_renewal_generate.py
rocketsroger: These are purposely added at the end of result so when the csv file is created, the last 3 columns are these.  These columns are for product to fill in after reviewing.

rocketsroger: I'll add some comments here to explain, thanks

tools/data_renewal_generate.py
rocketsroger: Thanks, good catch.  I'll add it.

tools/data_renewal_generate.py
rocketsroger: I'll fix, thanks

tools/data_renewal_generate.py
rocketsroger: Sure, I'll add the comment explaining why.  No, we don't need to try catch, metrics generation will fail if any telemetry is using date expiry.  We can either use date expiry or version expiry, Glean does not allow mixed.

tools/data_renewal_generate.py
rocketsroger: thanks, I'll fix.

tools/data_renewal_generate.py
rocketsroger: This is a ID that matches what we put in the csv.  This is so when the product review process is over, we can go through each one on the csv and find the corresponding number in `metrics.yaml`.  (Since the last step have to be manually done for now)

tools/data_renewal_generate.py
```diff
@@ -0,0 +1,149 @@
...
+import csv
+import yaml
+import json
+import sys
+
+from yaml.loader import FullLoader
+
+write_header = True
+total_count = 0
+
```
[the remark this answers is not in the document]
rocketsroger: I've decided to not add instructions in docs/README.  Looks like it's mostly about building Fenix.  I'll add instruction in https://github.com/mozilla-mobile/fenix/wiki/Creating-a-release-branch.

rocketsroger:
> Should we add comments about how to run this script in this file or somewhere else? Either way, I think we should have a comment that directs to wherever the instructions should be.

Added instruction in https://github.com/mozilla-mobile/fenix/wiki/Creating-a-release-branch#renew-telemetry

tools/data_renewal_generate.py
jonalmeida: > We can either use date expiry or version expiry, Glean does not allow mixed.

TIL

jonalmeida (approved):

Issue #23337, opened 2022-01-20
Https-only mode error page


Mugurell:
Following https://github.com/mozilla-mobile/fenix/issues/16952 we need to update the default error page to something offering more context and probably a "Continue to HTTP site" button similar to what's been done on Focus.

| Default error page | Focus error page |
| -- | -- | 
| ![image](https://user-images.githubusercontent.com/11428869/150400975-dcd6241c-69ca-40cf-90ae-621749513a5a.png) | ![image](https://user-images.githubusercontent.com/11428869/150400910-b6bf237b-55b2-4f65-891a-a1ded15317aa.png) | 

We could have a different error page offering more context, as suggested in https://github.com/mozilla-mobile/android-components/issues/11306.

Asking UX about the design and text to use. 
cc @topotropic 

-------

Fenix design is available at https://www.figma.com/file/RyIADLmC3B0BIS9jSiAti3/?node-id=1238%3A63076

<img width="399" alt="image" src="https://user-images.githubusercontent.com/11428869/156724139-c3ab9771-6436-497a-b94a-9725553ffb63.png">

Mugurell:
@Amejia481 Regarding the "continue" button which should store an exception from erroring on http I think the exceptions should be handled by GeckoView to allow for a smooth UX that would load the http webpage without any hiccup.
I imagine rather GV/Gecko seeing that the webpage is using http and deciding whether to send an error or continue to load the webpage anyway.
The API Focus uses - `document.reloadWithHttpsOnlyException()` doesn't seem useful on Fenix.
Thoughts?

Amejia481:
> @Amejia481 Regarding the "continue" button which should store an exception from erroring on http I think the exceptions should be handled by GeckoView to allow for a smooth UX that would load the http webpage without any hiccup.
> I imagine rather GV/Gecko seeing that the webpage is using http and deciding whether to send an error or continue to load the webpage anyway.

Yes, agree, I'm not completely sure if this API is already implemented on GV, maybe @agi could help to clarify, maybe the could expose the same API that desktop is using.

> The API Focus uses - document.reloadWithHttpsOnlyException() doesn't seem useful on Fenix.
> Thoughts?

I think we could use it, as we'll only need the API for storing the exemption to run before that, maybe the API is also on JS and we just need to expose reading the exception on GV.

agi:
[...]

Issue #16952, opened 2020-12-09
HTTPS-Only Mode 


Amejia481:
## User Story
- As a user, I want  be able to opt in to use the [HTTPS-Only Mode](https://blog.mozilla.org/security/2020/11/17/firefox-83-introduces-https-only-mode/) 

## Dependencies
- WIP.

### Acceptance Criteria
- When I navigate to any site, the HTTPS version should be always served

Amejia481:
This functionality was already introduced in [desktop 83](https://blog.mozilla.org/security/2020/11/17/firefox-83-introduces-https-only-mode/) and we want to start gathering all the dependencies needed for the feature. 

cc: @vesta0

saschanaz:
Duplicate of #14682 and of #12347

Amejia481:
@saschanaz I think we could close the other ones are they are feature requests, this is an user story issue and it will be used for creating all the UI/UX and engineering tasks :)

kbrosnan:
This needs error page work as well. We need to figure out a way to provide the escape hatch button to http. Need to sort out how the error page will work with PWAs. Need to think about how the escape hatch persistence will work with things like private browsing.

Snowden-rgb:
really needed feature

julian-alarcon:
As bug in Gecko view was already closed, it should be a priority.

cschanaj:
EFF has [announced](https://www.eff.org/deeplinks/2021/04/https-everywhere-now-uses-duckduckgos-smarter-encryption) that HTTPS Everywhere is expected to reach its EOL by the end of 2021. HTTPS-Only Mode on Fenix could be a preferred alternatives to HTTPS Everywhere EASE mode to enforce HTTPS secure connections only and block HTTP connections on insecure mobile networks.  I sincerely hope this feature can be prioritized. 

Related: https://github.com/EFForg/https-everywhere/issues/20048

brampitoyo:
@amedyne For Fenix, I think we can safely implement the same UI as desktop: with identical explanation, radio buttons, and exceptions list. (Let’s consult @betsymi at Content Design first, though!)

Do you know what desktop’s default value is? On Fenix, I feel pretty safe toggling `Enable in private windows only` to ON, but it’s best to follow desktop’s default value and not risk broken sites.

-----

[...]

PR #24095, opened 2022-03-04
For #16952 - Https-only mode support.

Mugurell:
Default will be disabled with users having the possibility to enable this from
a new setting in the `Privacy and security` section.
If enabled then by default this force https for all tabs with the option for
users to switch to forcing https only on private tabs.

An updated error page is to be implemented in https://github.com/mozilla-mobile/fenix/issues/23337

https://user-images.githubusercontent.com/11428869/156723735-0b7813d1-83be-427e-a8e5-1124b414fa12.mp4

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/Core.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/Core.kt
@@ -122,7 +122,8 @@ class Core(
             clearColor = ContextCompat.getColor(
                 context,
                 R.color.fx_mobile_layer_color_1
-            )
+            ),
+            httpsOnlyMode = context.settings().getHttpsOnlyMode()
         )
 
         GeckoEngine(
--- a/app/src/main/java/org/mozilla/fenix/settings/SettingsFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/SettingsFragment.kt
@@ -259,6 +259,9 @@ class SettingsFragment : PreferenceFragmentCompat() {
             resources.getString(R.string.pref_key_private_browsing) -> {
                 SettingsFragmentDirections.actionSettingsFragmentToPrivateBrowsingFragment()
             }
+            resources.getString(R.string.pref_key_https_only_settings) -> {
+                SettingsFragmentDirections.actionSettingsFragmentToHttpsOnlyFragment()
+            }
             resources.getString(R.string.pref_key_accessibility) -> {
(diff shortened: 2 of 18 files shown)


app/src/main/java/org/mozilla/fenix/settings/SettingsFragment.kt
```diff
@@ -602,6 +606,19 @@ class SettingsFragment : PreferenceFragmentCompat() {
         }
     }
 
+    @VisibleForTesting
+    internal fun setupHttpsOnlyPreferences() {
+        val trackingProtectionPreference =
```
mcarare, line 611: nit: naming.

app/src/main/java/org/mozilla/fenix/components/Core.kt
```diff
@@ -474,6 +476,16 @@ class Core(
         }
     }
 
+    fun getHttpsOnlyMode(): HttpsOnlyMode {
```
mcarare, line 479: nit: Just a thought: there already are similar methods in this file that "get" some values for core init, but IMO this would better fit in the Settings file seeing as it returns values that depend only on shared prefs.

mcarare (approved):
LGTM! Nice to see more tests! ⭐ 
Also tested on a device.

app/src/main/java/org/mozilla/fenix/components/Core.kt
Mugurell: Thanks, will move.

Mugurell:
Waiting for a confirmation on whether this should be behind a feature flag or not.

Mugurell:
> Waiting for a confirmation on whether this should be behind a feature flag or not.

Amedyne and Kevin confirmed this can freely ride the trains.

Mugurell:
I see two ui-tests failing:
- `org.mozilla.fenix.ui.ContextMenusTest#privateSelectAndSearchTextTest` because of 
> Test instrumentation process crashed`

Unsure why. Should not be affected by these changes, will try to rerun it.
- ` org.mozilla.fenix.ui.SettingsPrivacyTest. settingsPrivacyItemsTest ` because of
> androidx.test.espresso.AmbiguousViewMatcherException: 'with text: is "Off"' matches multiple views in the hierarchy.

This is a direct result of these changes which probably requires a deeper refactoring for which I've created https://github.com/mozilla-mobile/fenix/issues/24495

Issue #24148, opened 2022-03-08
Remove wrapper for ETP metrics.


mcarare:
Breakout from https://github.com/mozilla-mobile/fenix/issues/21292.

LaurentiuApahideanSV:
I tested the issue on Firefox Nightly 2020-30-03 by testing the following Enhanced tracking protections interactions and the events they generate:

State of the toggle
{
          "timestamp": 17169,
          "category": "events",
          "name": "preference_toggled",
          "extra": {
            "preference_key": "pref_key_tracking_protection",
            "enabled": "false"
          }

{
          "timestamp": 5670,
          "category": "events",
          "name": "preference_toggled",
          "extra": {
            "preference_key": "pref_key_tracking_protection",
            "enabled": "true"
          }
Switch to Strict
{
          "timestamp": 12893,
          "category": "tracking_protection",
          "name": "etp_setting_changed",
          "extra": {
            "etp_setting": "STRICT"
          }
Metrics with Strict
"preferences.enhanced_tracking_protection":"strict"

Switch to Standard
{
          "timestamp": 16085,
          "category": "tracking_protection",
          "name": "etp_setting_changed",
          "extra": {
            "etp_setting": "STANDARD"
          }
Metrics with Standard
"preferences.enhanced_tracking_protection:"standard"

Switch to Custom 
{
          "timestamp": 14102,
          "category": "tracking_protection",
          "name": "etp_setting_changed",
          "extra": {
            "etp_setting": "CUSTOM"
          }
Metrics with Custom
"preferences.enhanced_tracking_protection":"custom"

A user added a tracking protection exception through the TP toggle in the panel.
{
          "timestamp": 15357,
          "category": "tracking_protection",
          "name": "exception_added"
        }

User opening Tracking Protection settings through the panel
{
          "timestamp": 16770,
          "category": "tracking_protection",
          "name": "panel_settings"
        }
User opening Tracking Protection settings through the settings menu
{
          "timestamp": 12122,
          "category": "tracking_protection",
          "name": "etp_settings"
        }

A user pressed into a list of categorized trackers in tracking protection panel.
{
          "timestamp": 11094,
          "category": "tracking_protection",
          "name": "etp_tracker_list"
        },

[Glean Dashboard](https://debug-ping-preview.firebaseapp.com/pings/Laurentiu-3003)

### Device used:
- OnePlus 6T (Android 9)

PR #24149, opened 2022-03-08
For #24148: Remove wrapper from tracking protection metrics.

mcarare:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt
@@ -55,7 +55,6 @@ import org.mozilla.fenix.GleanMetrics.Tabs
 import org.mozilla.fenix.GleanMetrics.TabsTray
 import org.mozilla.fenix.GleanMetrics.ToolbarSettings
 import org.mozilla.fenix.GleanMetrics.TopSites
-import org.mozilla.fenix.GleanMetrics.TrackingProtection
 import org.mozilla.fenix.GleanMetrics.VoiceSearch
 import org.mozilla.fenix.GleanMetrics.Wallpapers
 import org.mozilla.fenix.ext.components
@@ -456,25 +455,7 @@ private val Event.wrapper: EventWrapper<*>?
         is Event.NotificationMediaPause -> EventWrapper<NoExtraKeys>(
             { MediaNotification.pause.record(it) }
         )
-        is Event.TrackingProtectionTrackerList -> EventWrapper<NoExtraKeys>(
-            { TrackingProtection.etpTrackerList.record(it) }
-        )
-        is Event.TrackingProtectionIconPressed -> EventWrapper<NoExtraKeys>(
-            { TrackingProtection.etpShield.record(it) }
-        )
-        is Event.TrackingProtectionSettingsPanel -> EventWrapper<NoExtraKeys>(
-            { TrackingProtection.panelSettings.record(it) }
-        )
-        is Event.TrackingProtectionSettings -> EventWrapper<NoExtraKeys>(
-            { TrackingProtection.etpSettings.record(it) }
-        )
-        is Event.TrackingProtectionException -> EventWrapper<NoExtraKeys>(
-            { TrackingProtection.exceptionAdded.record(it) }
-        )
-        is Event.TrackingProtectionSettingChanged -> EventWrapper(
-            { TrackingProtection.etpSettingChanged.record(it) },
-            { TrackingProtection.etpSettingChangedKeys.valueOf(it) }
-        )
+
         is Event.OpenedLink -> EventWrapper(
             { Events.openedLink.record(it) },
             { Events.openedLinkKeys.valueOf(it) }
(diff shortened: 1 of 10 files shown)


app/src/main/java/org/mozilla/fenix/settings/SettingsFragment.kt
```diff
@@ -248,7 +250,7 @@ class SettingsFragment : PreferenceFragmentCompat() {
                 SettingsFragmentDirections.actionSettingsFragmentToSearchEngineFragment()
             }
             resources.getString(R.string.pref_key_tracking_protection_settings) -> {
-                requireContext().metrics.track(Event.TrackingProtectionSettings)
+                TrackingProtection.etpSettings.record(NoExtras())
```
Mugurell, line 253: Not sure replacing the current telemetry with recording Glean events in place is the right approach.
If we are not using anymore `metrics.track` we are sidestepping the `metrics` configuration

https://github.com/mozilla-mobile/fenix/blob/7ca7190199a9b5ce980b73f4cf3a3c15ac65f3d0/app/src/main/java/org/mozilla/fenix/components/Analytics.kt#L102-L112

 for a [wrapper with multiple responsibilities](https://github.com/mozilla-mobile/fenix/blob/7ca7190199a9b5ce980b73f4cf3a3c15ac65f3d0/app/src/main/java/org/mozilla/fenix/components/metrics/MetricController.kt#L41): 
 - registering multiple analytics services for tracking the same events.
 - checking whether the telemetry should be tracked - users can disable it from app settings.
 - allowing to also easily eanble telemetry depending on build configuration.
 
 --- 
One important piece of Glean functionality that I'd want to also confirm is what happens with a `record` call if Glean is not initialized/doesn't have upload enabled because maybe users disabled the telemetry in app settings. Would Glean just cache the telemetry for this period and then send it when telemetry is enabled again? 
@travis79 Do you know?

mcarare: - We do not send any events directly to Adjust and if we do not plan to in the future IMO always trying to send through both services is inefficient so, on that topic, I do not think we need the controller.

- Glean is already initialized with the value of the `isTelemetryEnabled` so checking it in the controller is redundant. Glean will not send the events if `uploadEnabled`  is false. Maybe @travis79 can clarify what happens with the metrics at that point. 

There might be a problem with changing the setting after Glean is initialized. I do not think we currently reset Glean at that point (@travis79  is calling `setUploadEnabled(false)` recommended  or is it expected that the change is made next time Glean is initialized?) 

- Enabling depending on the build configuration can be as easily configured through `uploadEnabled` parameter on Glean init. 


That being said, I would prefer to try to remove all the wrapping code on the calls. But at the same time, we could keep the controller structure and implement different `track` methods that accept  `EventMetricType<T, E>` as params. 

Maybe @travis79 can clarify some of the issues stated here. (Thanks in advance! )

travis79: If Glean is not initialized, it will cache API calls and play them back once it is initialized (provided telemetry collection is enabled). This has some limits, but nothing we have run into problems with on mobile.

You are correct that trying to wrap recording of metrics in `should we record` conditionals is redundant, the correct way to control this is through the `initialize` function, and through the `setUploadEnabled` API in Glean.
Glean takes care of all of the logic of what the correct thing to do with the metrics is, depending on the state of the preference. When `uploadEnabled` is false, no metrics are recorded or sent. If it is toggled to false during runtime, then Glean will stop recording metrics, delete any metrics it already has recorded, and send a "deletion-request" ping to the server to indicate we should delete all the user's data there also. Calling `setUploadEnabled` during runtime is expected and correct, it should be directly wired to the "Send Usage Data" preference in the app settings.

The Glean team has long advocated for calling the Glean API functions directly at the site of measurement rather than having a wrapper around things to make diagnosing telemetry issues easier and to prevent relying on Glean internals, but we also understood that at the time the wrapper was created we were also sending some duplication of data to Adjust, LeanPlum and Glean and having a Tee was useful.  I defer to the product team to make the best decision here for their purposes, though.

travis79: `Glean.initialized` _must_ be called on startup every time, regardless of the state of the preference in order to ensure that Glean is doing the correct things to follow our data-collection policies such as sending the "deletion-request" ping.

app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt
```diff
@@ -456,25 +455,7 @@ private val Event.wrapper: EventWrapper<*>?
...
-            { TrackingProtection.etpSettings.record(it) }
-        )
-        is Event.TrackingProtectionException -> EventWrapper<NoExtraKeys>(
-            { TrackingProtection.exceptionAdded.record(it) }
-        )
-        is Event.TrackingProtectionSettingChanged -> EventWrapper(
-            { TrackingProtection.etpSettingChanged.record(it) },
-            { TrackingProtection.etpSettingChangedKeys.valueOf(it) }
-        )
+
```
rocketsroger, line 458: nit: remove extra line

rocketsroger (approved):
:shipit:

PR #24202, opened 2022-03-11
For #23337 - Customize the https-only mode error page

Mugurell:
Based on the designs from https://www.figma.com/file/RyIADLmC3B0BIS9jSiAti3/?node-id=1238%3A63076

Needs https://github.com/mozilla-mobile/fenix/pull/24095 and https://github.com/mozilla-mobile/android-components/pull/11842
(The AC patch must also be validated for Focus - https://github.com/mozilla-mobile/focus-android/issues/6640)

There is an open issue for gecko about this new error page - https://bugzilla.mozilla.org/show_bug.cgi?id=1759114

https://user-images.githubusercontent.com/11428869/157865225-42ae6251-c0c6-429b-89ae-1de355d918e1.mp4

@topotropic Because these error pages are driven from AC and there the "Go back" option used on other pages also is actually "Go Back (Recommended)" I used it for this error page also. 
It is possible to have have the text from Figma but this is the default so it's a simpler solution for a more cohesive design. 
Asking if it's okay to keep the default "Go Back (Recommended)" or if I should change to just "Go back" as on Figma.

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -330,6 +330,12 @@
     <string name="preferences_https_only_in_all_tabs">Enable in all tabs</string>
     <!-- Option for the https only setting -->
     <string name="preferences_https_only_in_private_tabs">Enable only in private tabs</string>
+    <!-- Title shown in the error page for when trying to access a http website while https only mode is enabled. -->
+    <string name="errorpage_httpsonly_title">Secure site not available</string>
+    <!-- Message shown in the error page for when trying to access a http website while https only mode is enabled. The message has two paragraphs. This is the first. -->
+    <string name="errorpage_httpsonly_message_title">Most likely, the website simply does not support HTTPs.</string>
+    <!-- Message shown in the error page for when trying to access a http website while https only mode is enabled. The message has two paragraphs. This is the second. -->
+    <string name="errorpage_httpsonly_message_summary">However, it’s also possible that an attacker is involved. If you continue to the website, you should not enter any sensitive info. If you continue, HTTPS-Only mode will be turned off temporarily for the site.</string>
     <!-- Preference for accessibility -->
     <string name="preferences_accessibility">Accessibility</string>
     <!-- Preference to override the Firefox Account server -->
--- a/app/src/main/assets/lowMediumErrorPages.js
+++ b/app/src/main/assets/lowMediumErrorPages.js
@@ -18,11 +18,17 @@ function parseQuery(queryString) {
  * Updates the HTML elements based on the queryMap
  */
 function injectValues(queryMap) {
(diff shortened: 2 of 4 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -312,6 +312,11 @@
     <string name="preferences_screenshots_in_private_mode_disclaimer">If allowed, private tabs will also be visible when multiple apps are open</string>
     <!-- Preference for adding private browsing shortcut -->
     <string name="preferences_add_private_browsing_shortcut">Add private browsing shortcut</string>
+    <string name="errorpage_httpsonly_title">Secure site not available</string>
+    <!-- Message shown in the error page for when trying to access a http website while https only mode is enabled. The message has two paragraphs. This is the first. -->
+    <string name="errorpage_httpsonly_message_title">Most likely, the website simply does not support HTTPs.</string>
+    <!-- Message shown in the error page for when trying to access a http website while https only mode is enabled. The message has two paragraphs. This is the second. -->
+    <string name="errorpage_httpsonly_message_summary">However, it’s also possible that an attacker is involved. If you continue to the website, you should not enter any sensitive info. If you continue, HTTPS-Only mode will be turned off temporarily for the site.</string>
```
gabrielluong, line 319: Can we move these strings to an appropriate section and not inject them in between the `preferences_` strings so that we can maintain keeping related strings to a particular screen together

app/src/main/res/values/strings.xml
```diff
@@ -312,6 +312,11 @@
     <string name="preferences_screenshots_in_private_mode_disclaimer">If allowed, private tabs will also be visible when multiple apps are open</string>
     <!-- Preference for adding private browsing shortcut -->
     <string name="preferences_add_private_browsing_shortcut">Add private browsing shortcut</string>
+    <string name="errorpage_httpsonly_title">Secure site not available</string>
```
gabrielluong, line 315: This needs a localization comment.

app/src/main/res/values/strings.xml
Mugurell: This patch would be added on top of https://github.com/mozilla-mobile/fenix/pull/24095 which adds the strings for these section at lines 321-334 - below the strings for what visually the settings section above - "Private browsing" uses.

mcarare (approved):
LGTM!

Mugurell:
Was thinking of landing this now but maybe it would be better to first land the patch that firsts adds the https-only error page so that this can be cleanly added on top.

Mugurell:
Many checks failed but without any logs or indication about why.
https://github.com/mozilla-mobile/fenix/runs/5740254327?check_suite_focus=true
Trying again.

Issue #24177, opened 2022-03-09
Display the latest synced tab in Jump Back In.


MatthewTighe:
https://mozilla-hub.atlassian.net/browse/FNXV2-19419

Add an additional Jump Back In section that shows the latest tab that was synced from a different device.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-19783)

MatthewTighe:
@SoftVision-LorandJanos My mistake, I misunderstood how the feature flags and secret debug menu were being used by QA in the original work. I've opened #24802 to correct it, and once that's merged the feature should be accessible using the secret debug menu.

MatthewTighe:
Discovered another bug which I filed as #24823, so I'd hold off on QA for now. I'll label this issue again once the feature is a little more stable. Thanks for your patience!

PR #24221, opened 2022-03-11
for #24177: add ui wireframe for synced recent tab

MatthewTighe:
Still waiting on some feedback from design in terms of which color tokens to use for dark mode, but thought I'd move forward in the code review search in the meantime

image

image

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/home/recenttabs/RecentTabsListFeature.kt
```diff
@@ -48,6 +48,13 @@ sealed class RecentTab {
      */
     data class Tab(val state: TabSessionState) : RecentTab()
 
+    data class SyncedTab(
```
gabrielluong, line 51: kDoc required

app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabs.kt
```diff
@@ -69,16 +74,25 @@ import org.mozilla.fenix.theme.FirefoxTheme
...
  * @param onRecentTabClick Invoked when the user clicks on a recent tab.
  * @param onRecentSearchGroupClicked Invoked when the user clicks on a recent search group.
+ * @param onRecentSyncedTabClick Invoked when the user clicks on the recent synced tab.
+ * @param onSyncedTabSeeAllClick Invoked when user clicks on the "See all" button in the synced tab card.
  */
 @Composable
+@Suppress("LongParameterList")
 fun RecentTabs(
     recentTabs: List<RecentTab>,
+    recentSyncedTab: RecentTab.SyncedTab?,
```
gabrielluong, line 89: Can we move this inside of the list of `RecentTab`? That way you can control whether or not to display this SyncedTab by whether or not it is in the list through the `RecentTabsFeature` and we won't need extra parameters `shouldDisplaySyncedTab` or `recentSyncedTab`.

app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabs.kt
```diff
@@ -244,25 +267,121 @@ private fun RecentSearchGroupItem(
...
+ * @param onRecentSyncedTabClick Invoked when item is clicked.
+ * @param onSeeAllClick Invoked when "See all" button is clicked.
+ */
+@Composable
+private fun RecentSyncedTabItem(
+    tab: RecentTab.SyncedTab?,
+    onRecentSyncedTabClick: (RecentTab.SyncedTab) -> Unit,
+    onSeeAllClick: () -> Unit,
+) {
+    val iconPainter = painterResource(id = R.drawable.ic_synced_tabs)
```
gabrielluong, line 283: I would just inline these one liner that are fairly straight forward when it's not being reused.

app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabs.kt
```diff
@@ -244,25 +267,121 @@ private fun RecentSearchGroupItem(
...
+                    backgroundColor = FirefoxTheme.colors.actionSecondary,
+                    contentColor = FirefoxTheme.colors.textPrimary,
+                ),
+                modifier = Modifier
+                    .padding(bottom = 16.dp)
+                    .requiredHeight(36.dp)
+                    .fillMaxWidth()
+            ) {
+                Text(
+                    stringResource(id = R.string.recent_tabs_see_all_synced_tabs_button_text),
```
gabrielluong, line 351: Add new named parameter here just for consistency

app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabs.kt
```diff
@@ -244,25 +267,121 @@ private fun RecentSearchGroupItem(
...
+                    modifier = Modifier.fillMaxHeight()
+                ) {
+                    if (tab == null) {
+                        RecentTabTitlePlaceholder()
+                    } else {
+                        RecentTabTitle(title = tab.title)
+                    }
+
+                    Row {
+                        RecentTabIcon(
```
gabrielluong, line 320: The change to `RecentTabIcon` seems to accommodate the drawable icon, but since we aren't needing to load a browser icon and we are just simply displaying a drawable, I think we should just use an `Image()` here to not overload `RecentTabIcon`

gabrielluong (commented):
Just a first pass to see if we can address some architectural considerations.

app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabs.kt
```diff
@@ -69,16 +74,25 @@ import org.mozilla.fenix.theme.FirefoxTheme
...
+ * @param onRecentSyncedTabClick Invoked when the user clicks on the recent synced tab.
+ * @param onSyncedTabSeeAllClick Invoked when user clicks on the "See all" button in the synced tab card.
  */
 @Composable
+@Suppress("LongParameterList")
 fun RecentTabs(
     recentTabs: List<RecentTab>,
+    recentSyncedTab: RecentTab.SyncedTab?,
     menuItems: List<RecentTabMenuItem>,
+    shouldDisplaySyncedTab: () -> Boolean,
```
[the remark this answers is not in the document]
MatthewTighe: I worry that `show` as a verb here would imply an action, especially given the type of the lambda. Looking at other properties throughout the codebase with the same type shows a preference towards the interrogative case. How does `isSyncedTabDisplayed` sound, to mirror other examples?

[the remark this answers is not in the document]
gabrielluong: `isSyncedTabDisplayed` is fine. We already use `show` a bunch in `RecentVisits`. I would investigate https://github.com/mozilla-mobile/fenix/pull/24221#discussion_r825102591 first since that would just eliminate these extra parameters entirely.

app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabs.kt
MatthewTighe: Can we punt on this suggestion until the implementation? IMO it could end up that synced tabs make more sense to be separated, since they will won't be observed from the `BrowserStore` and have a lifecycle that's outside of local changes.

app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabs.kt
```diff
@@ -244,25 +266,120 @@ private fun RecentSearchGroupItem(
...
+                        }
+                    }
+                }
+            }
+
+            Button(
+                onClick = onSeeAllClick,
+                colors = ButtonDefaults.outlinedButtonColors(
+                    backgroundColor = FirefoxTheme.colors.actionSecondary,
+                    contentColor = FirefoxTheme.colors.textPrimary,
```
gabrielluong, line 341: This should probably be textActionSecondary. We have 1:1 to with the background, icon and text colours for actions.

MatthewTighe: Oh yeah, thanks. I'll update this for now. Still not sure if the background color of the button will change based on UX feedback, since I assume the button and placeholders will want to be the same color and white (in dark mode) seems a little out of place.

app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabs.kt
gabrielluong: So my 2 cents here, I think you can actually go ahead with removing `isSyncedTabDisplayed` because you can assume you won't provide a `recentSyncedTab` if the FeatureFlag is false. Since this also just wireframing, I can also imagine you can append the `RecentTab.SyncedTab` to the end of the `recentTabs` for testing purposes as well. 

I understand we might not be observing from the `BrowserStore`, but we will definitely be observing the `HomeFragmentStore`. Now the future consideration I have, would you be considering adding a new state in `HomeFragmentStore` or would you be appending and updating the list in the reducer when you call a dispatch when you handle a synced tab?

I will admit it is slightly unusual for us to land UI code without the implementation.

MatthewTighe: Okay, that's all good food for thought. I'll remove it for now, and switch the synced tab to be part of the list.

Just for posterity, the other consideration I had in adding `isSyncedTabDisplayed` is that it won't only be dictated by the feature flag. It will also be controlled by whether accounts are logged in and have other devices in their sync constellation in the first place. But agreed that might end up being controlled by a different layer anyway, so I was getting a bit ahead of myself

MatthewTighe:
Think that covers it for first round of feedback, including feedback on colors from UX. @gabrielluong I did want to point out that I made an additional small update to `FirefoxColors`, as I believe it was out-of-sync with the style sheet.

MatthewTighe:
Updated screenshots:

image

image

gabrielluong:
> Think that covers it for first round of feedback, including feedback on colors from UX. @gabrielluong I did want to point out that I made an additional small update to `FirefoxColors`, as I believe it was out-of-sync with the style sheet.

Thanks, can you make a PR to land that update?

app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabs.kt
```diff
@@ -381,16 +511,15 @@ private fun RecentTabMenu(
...
+ * @param icon The icon of the tab. Fallback to loading the icon from the [url] if null
  */
 @Composable
 private fun RecentTabIcon(
     url: String,
     modifier: Modifier = Modifier,
     contentScale: ContentScale = ContentScale.Fit,
     alignment: Alignment = Alignment.Center,
-    icon: Bitmap? = null
+    icon: Bitmap?
```
gabrielluong, line 522: I think we can undo this change.

app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabs.kt
```diff
@@ -244,25 +264,127 @@ private fun RecentSearchGroupItem(
...
+                onClick = onSeeAllClick,
+                colors = ButtonDefaults.outlinedButtonColors(
+                    backgroundColor = if (tab == null) {
+                        FirefoxTheme.colors.layer3
+                    } else {
+                        FirefoxTheme.colors.actionSecondary
+                    }
+                ),
+                modifier = Modifier
+                    .padding(bottom = 16.dp)
```
gabrielluong, line 346: Can this be a Spacer() or be moved into Column?

app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabs.kt
```diff
@@ -244,25 +264,127 @@ private fun RecentSearchGroupItem(
...
+                    }
+
+                    Row {
+                        Image(
+                            painter = painterResource(R.drawable.ic_synced_tabs),
+                            contentDescription = stringResource(
+                                R.string.recent_tabs_synced_device_icon_content_description
+                            ),
+                            modifier = Modifier
+                                .size(18.dp, 18.dp)
```
gabrielluong, line 322:
```suggestion
                            modifier = Modifier.size(18.dp)
```
Can just use the single param version of `size()`

app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabs.kt
```diff
@@ -244,25 +264,127 @@ private fun RecentSearchGroupItem(
     }
 }
 
+/**
+ * Displays a recent synced tab.
```
gabrielluong, line 268:
```suggestion
 * A recent synced tab.
```

app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabs.kt
gabrielluong: Actually I don't think you need this bottom padding since you already have a 16dp padding around the Column.

app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabs.kt
```diff
@@ -244,25 +264,127 @@ private fun RecentSearchGroupItem(
...
+    Card(
+        modifier = Modifier
+            .fillMaxWidth()
+            .requiredHeight(180.dp)
+            .clickable { tab?.let { onRecentSyncedTabClick(tab) } },
+        shape = RoundedCornerShape(8.dp),
+        backgroundColor = FirefoxTheme.colors.layer2,
+        elevation = 6.dp
+    ) {
+        Column(modifier = Modifier.padding(16.dp), verticalArrangement = Arrangement.SpaceBetween) {
```
gabrielluong, line 290: Let's move the named parameters to separate lines

app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabs.kt
```diff
@@ -103,6 +114,22 @@ fun RecentTabs(
...
+                        )
+                    }
+                }
+            }
+            if (FeatureFlags.taskContinuityFeature) {
+                RecentSyncedTabItem(
+                    null,
+                    onRecentSyncedTabClick,
+                    onSyncedTabSeeAllClick,
+                )
```
gabrielluong, line 132: Since this is inside then `when` block, I don't think it was intended for this to remain here.

app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabs.kt
```diff
@@ -72,13 +78,18 @@ import org.mozilla.fenix.theme.FirefoxTheme
...
+ * @param onSyncedTabSeeAllClick Invoked when user clicks on the "See all" button in the synced tab card.
  */
 @Composable
+@Suppress("LongParameterList")
 fun RecentTabs(
     recentTabs: List<RecentTab>,
     menuItems: List<RecentTabMenuItem>,
     onRecentTabClick: (String) -> Unit = {},
-    onRecentSearchGroupClicked: (String) -> Unit = {}
+    onRecentSearchGroupClicked: (String) -> Unit = {},
```
gabrielluong, line 90:
```suggestion
    onRecentSearchGroupClick: (String) -> Unit = {},
```

app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabs.kt
```diff
@@ -72,13 +78,18 @@ import org.mozilla.fenix.theme.FirefoxTheme
...
+    onRecentSyncedTabClick: (RecentTab.SyncedTab) -> Unit = {},
+    onSyncedTabSeeAllClick: () -> Unit = {},
```
gabrielluong, line 92:
```suggestion
    onSyncedTabSeeAllButtonClick: () -> Unit = {},
```
Just wanted to make sure we had noun before the Click.

app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabs.kt
```diff
@@ -244,25 +269,134 @@ private fun RecentSearchGroupItem(
...
+@Composable
+private fun RecentSyncedTabItem(
+    tab: RecentTab.SyncedTab?,
+    onRecentSyncedTabClick: (RecentTab.SyncedTab) -> Unit,
+    onSeeAllButtonClick: () -> Unit,
+) {
+    Card(
+        modifier = Modifier
+            .fillMaxWidth()
+            .requiredHeight(180.dp)
```
gabrielluong, line 289: Did we also want to change this to `height`?

app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabs.kt
```diff
@@ -244,25 +269,134 @@ private fun RecentSearchGroupItem(
...
+                        } else {
+                            Image(
+                                painter = painterResource(R.drawable.ic_synced_tabs),
+                                contentDescription = stringResource(
+                                    R.string.recent_tabs_synced_device_icon_content_description
+                                ),
+                                modifier = Modifier.size(18.dp, 18.dp)
+                            )
+                        }
+                        Spacer(modifier = Modifier.width(8.dp))
```
gabrielluong, line 336: Add a new line above

app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabs.kt
```diff
@@ -381,8 +523,7 @@ private fun RecentTabMenu(
  * @param modifier [Modifier] used to draw the image content.
  * @param contentScale [ContentScale] used to draw image content.
  * @param alignment [Alignment] used to draw the image content.
- * @param icon The icon of the tab. Fallback to loading the icon from the [url] if the [icon]
- * is null.
+ * @param icon The icon of the tab. Fallback to loading the icon from the [url] if null
```
gabrielluong, line 526: Also don't need this change.

gabrielluong (approved):

app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabs.kt
```diff
@@ -244,25 +269,134 @@ private fun RecentSearchGroupItem(
...
+                        } else {
+                            RecentTabSubtitle(subtitle = tab.deviceDisplayName)
+                        }
+                    }
+                }
+            }
+
+            Spacer(modifier = Modifier.height(32.dp))
+
+            Button(
```
gabrielluong, line 349: Looking at the designs. We don't have any elevations. You will want to `elevation = ButtonDefaults.elevation(defaultElevation = 0.dp, pressedElevation = 0.dp),` See https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/CustomizeHomeButtonViewHolder.kt#69

Issue #24224, opened 2022-03-11
[Gleanplum] Send telemetry events related to messaging framework


Amejia481:
We would like to know how user interact with the messages shown.
Acceptance Criteria

- Report when a message is shown.
- Report when the button is clicked.
- Report when the message is dismissed.

Related ticket https://mozilla-hub.atlassian.net/browse/FNXV2-19347

Issue #24223, opened 2022-03-11
 [Gleanplumb] Manages Gleamplum messages


Amejia481:
We would like to create component that will be able to interact with the messaging framework and will able and notify the apps when a new message show be shown.

Acceptance Criteria
- Fetch messages from nimbus
- Keep track of active messages.
- Notify the app when a message show be shown.
- Keep track of updating the status of a message when user interacts with it.

Related ticket https://mozilla-hub.atlassian.net/browse/FNXV2-19346

Issue #24222, opened 2022-03-11
[Gleanplum] Persist user interactions with nimbus messages


Amejia481:
We have to store all the metadata related to an user’s interaction with message like:

- When the button is clicked.
- When a message is shown.
- When the message is dismissed.

Related ticket https://mozilla-hub.atlassian.net/browse/FNXV2-19813

Issue #24220, opened 2022-03-11
Connect Gleanplum messages with the new tab card 


Amejia481:
https://mozilla-hub.atlassian.net/browse/FNXV2-19340

PR #24225, opened 2022-03-11
For #24220 and #24223: Connect Gleanplum messages with the new tab ui card

Amejia481:
To avoid really long tail PR.

I've broken down the GleanPlum (Messaging framework) work in 4 parts, this way they can be reviewed independently :

1. This display a message from GleamPlum on the UI and send UI interactions
2. Provide and manage messages from the nimbus https://github.com/mozilla-mobile/fenix/issues/24223.
3. Store metadata related to user interactions https://github.com/mozilla-mobile/fenix/issues/24222.
4. Send metrics about user's interaction with a message https://github.com/mozilla-mobile/fenix/issues/24224.

This PR only covers the first item, taking a message from GleamPlum, display it on the UI and send user's interaction.

### Pull Request checklist

1. click on Show All Checks,
5. click on the "Fenix - assemble" task, then click "Run Artifacts".
6. the APK links should be on the left side of the screen, named for each CPU architecture


app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/onboarding/MessageCardViewHolder.kt
```diff
@@ -0,0 +1,59 @@
...
+        val binding = NimbusMessageCardBinding.bind(view)
+        val safeMessage = message ?: return
+
+        binding.messageButton.setOnClickListener {
+            interactor.onMessageClicked(safeMessage)
+
+        }
+
+        binding.descriptionText.text = safeMessage.data.text
+        // TODO update with the right info.
```
Amejia481, line 38: @jhugman could you give me a hand? :) I'm still having difficulties bringing the  button from fml, maybe we could pair on it.
`button-label` same as https://github.com/mozilla-mobile/fenix/pull/24048#discussion_r819986757

app/src/main/res/values/strings.xml
```diff
@@ -1737,4 +1737,7 @@
     <string name="pocket_stories_feature_caption">Part of the Firefox family. %s</string>
     <!-- Clickable text for opening an external link for more information about Pocket. -->
     <string name="pocket_stories_feature_learn_more">Learn more</string>
+    <string name="gdpr_has_you_covered_firefox_has_gdpr_covered">gdpr has you covered firefox has gdpr covered</string>
+    <string name="sharing_your_phone_autoclosing_private_tabs_is_for_you">sharing your phone autoclosing private tabs is for you</string>
+    <string name="love_firefox_fill_in_our_survey">love_firefox fill in_our survey</string>
```
Amejia481, line 1742: @jhugman I added these to make the build to compile, not sure how should I handle this.

https://github.com/mozilla-mobile/fenix/pull/24048#discussion_r823238548

jonalmeida:
Drive by: is there anything here we should be putting into components? What happens when we need to add Focus? Are we going to copy paste code then?

Amejia481:
> Drive by: is there anything here we should be putting into components? What happens when we need to add Focus? Are we going to copy paste code then?

Yeah, ideally we would like to only have the UI parts on Fenix or Focus, for more details see https://github.com/mozilla-mobile/fenix/pull/24227#issuecomment-1066772867

app/src/main/java/org/mozilla/fenix/gleanplumb/state/MessagingReducer.kt
```diff
@@ -0,0 +1,32 @@
...
+import org.mozilla.fenix.components.appstate.AppAction
+import org.mozilla.fenix.components.appstate.AppState
+import org.mozilla.fenix.gleanplumb.MessagingState
+
+/**
+ * Reducer for [MessagingState].
+ */
+internal object MessagingReducer {
+    fun reduce(state: AppState, action: AppAction.MessagingAction): AppState = when (action) {
+        is AppAction.MessagingAction.UpdateMessageToShow -> {
```
jonalmeida, line 16: nit: Import `AppAction.MessagingAction`

app/src/main/java/org/mozilla/fenix/gleanplumb/MessagingFeature.kt
```diff
@@ -0,0 +1,33 @@
...
+        flow.map { it }.ifChanged {
+            it.messagingState.messageToShow
+        }.collect { state ->
+            onMessageChange(state.messagingState.messageToShow)
+        }
+    }
+
+    override fun start() {
+        super.start()
+        store.dispatch(AppAction.MessagingAction.Evaluate)
```
jonalmeida, line 31: This is a good place to have it for now. I wonder if this is soon enough for you though? Maybe earlier in the Activity/Fragment lifecycle so by the time we observe, a message will be ready?

A minor optimization you could do here is fire the action _before_ you call super (which starts observing) for the same reason above.

```suggestion
        store.dispatch(AppAction.MessagingAction.Evaluate)
        super.start()
```

app/src/main/java/org/mozilla/fenix/gleanplumb/Message.kt
```diff
@@ -0,0 +1,44 @@
...
+     * A data class that holds metadata that help to identify if a message should shown.
+     *
+     * @param id identifies a message as unique.
+     * @param displayCount Indicates how many times a message is displayed.
+     * @param pressed Indicates if a message has been clicked.
+     * @param dismissed Indicates if a message has been closed.
+     */
+    data class Metadata(
+        val id: String,
+        val displayCount: Int = 0,
```
jonalmeida, line 40: Does this default make sense for Nimbus? I'm honestly not sure myself. Maybe it should be 1?

No default might also be less error-prone if you're also unsure.

app/src/main/java/org/mozilla/fenix/components/appstate/AppAction.kt
```diff
@@ -16,4 +17,14 @@ sealed class AppAction : Action {
     data class AddNonFatalCrash(val crash: NativeCodeCrash) : AppAction()
     data class RemoveNonFatalCrash(val crash: NativeCodeCrash) : AppAction()
     object RemoveAllNonFatalCrashes : AppAction()
+
+    sealed class MessagingAction : AppAction() {
+        object Initialize : MessagingAction()
+        object Evaluate : MessagingAction()
+        data class UpdateMessageToShow(val message: Message?) : MessagingAction()
```
jonalmeida, line 24: We could fire an `UpdateMessageToShow` action without a message?

app/src/main/java/org/mozilla/fenix/gleanplumb/state/MessagingMiddleware.kt
```diff
@@ -0,0 +1,85 @@
...
+
+    private fun removeMessage(
+        context: MiddlewareContext<AppState, AppAction>,
+        message: Message
+    ): List<Message> {
+        return context.state.messagingState.messages - message
+    }
+
+    private fun updateMessage(
+        context: MiddlewareContext<AppState, AppAction>,
```
jonalmeida, line 79: This is fine, but as a lazy person, I would not have typed this all out three times, but instead used a typealias. For example:

```kotlin

typealias AppStoreMiddlewareContext = MiddlewareContext<AppState, AppAction>

class MessagingMiddleware(private val storage: MessagingStorage) : Middleware<AppState, AppAction> {
    override fun invoke(
        context: AppStoreMiddlewareContext,
        next: (AppAction) -> Unit,
        action: AppAction
    ) {
    }
}
```

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -334,6 +346,11 @@ class HomeFragment : Fragment() {
                 settings = components.settings,
                 engine = components.core.engine,
                 metrics = components.analytics.metrics,
+                messageController = DefaultMessageController(
```
jonalmeida, line 349: Tell me a controller needs to be refactored, without telling me a controller needs to be refactored. 😬 

Screen Shot 2022-03-18 at 12 43 59 PM

app/src/main/res/values/strings.xml
```diff
@@ -1763,4 +1763,7 @@
     <string name="pocket_stories_feature_caption">Part of the Firefox family. %s</string>
     <!-- Clickable text for opening an external link for more information about Pocket. -->
     <string name="pocket_stories_feature_learn_more">Learn more</string>
+    <string tools:ignore="UnusedResources" name="gdpr_has_you_covered_firefox_has_gdpr_covered">gdpr has you covered firefox has gdpr covered</string>
```
jonalmeida, line 1766: Are these strings meant to be in this PR? They seem unrelated.

jonalmeida (commented):
Looks good! Mostly nits but this direction seems to be working out so far!

Amejia481:
> Looks good! Mostly nits but this direction seems to be working out so far!

Thanks for the quick, I mostly wanted to see if I capture, what we agreed on, while paired, I'll address the nits and add tests

app/src/main/java/org/mozilla/fenix/gleanplumb/MessagingFeature.kt
Amejia481: Thanks for noticing this. I'll add the suggestion. I'm not completely sure how time consuming could the evaluations be. I guess ideally we would like to make the evaluation in bg thread, and post the results async. cc  @jhugman Do you have an of the performance hit of the evaluations?

app/src/main/java/org/mozilla/fenix/components/appstate/AppAction.kt
Amejia481: Yes, I think it will be useful in cases, where we would like to indicate that a previous message needs to be hidden or to indicate no new messages is available

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -1016,18 +1016,22 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
...
+            )
+        }
+    }
+
     @VisibleForTesting
     internal fun getSettings(): Settings = settings()
 
     private fun shouldNavigateToBrowserOnColdStart(savedInstanceState: Bundle?): Boolean {
         return isActivityColdStarted(intent, savedInstanceState) &&
-            !externalSourceIntentProcessors.any {
```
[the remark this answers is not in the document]
Amejia481: Thanks for caching this, it was removed by mistake.

app/src/main/res/values/strings.xml
Amejia481: They should be address on https://github.com/mozilla-mobile/android-components/pull/11879.
I was adding them to be able to compile locally

app/src/main/java/org/mozilla/fenix/gleanplumb/state/MessagingMiddleware.kt
```diff
@@ -0,0 +1,85 @@
...
+                )
+                val newMessage = oldMessage.copy(
+                    metadata = newMetadata
+                )
+                val newMessages = if (newMetadata.displayCount < oldMessage.data.maxDisplayCount) {
+                    updateMessage(context, oldMessage, newMessage)
+                } else {
+                    removeMessage(context, oldMessage)
+                }
+                context.dispatch(UpdateMessages(newMessages))
```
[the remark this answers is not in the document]
csadilek: Yes, that's right! If you don't need this to happen synchronously `store.dispatch()` will do. Dispatching on the current context is useful in cases where the middleware wants to trigger additional state changes before continuing. Also, important to point out that `context.dispatch` needs to happen on the current thread and will otherwise crash to prevent incorrect usage.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -1016,18 +1016,22 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
         }
     }
 
+    fun processIntent(intent: Intent): Boolean {
```
csadilek, line 1019: nit: looks like this could be `private`.

csadilek: (Ah you refactored and added tests 🎉 , maybe a `VisibleForTesting`) :)

app/src/main/java/org/mozilla/fenix/gleanplumb/MessagingFeature.kt
jonalmeida: The background thread idea is a separate idea aside from this - it's more about, don't you want to start the evaluation as soon as possible (regardless of which thread it'll be on)?

app/src/main/java/org/mozilla/fenix/gleanplumb/Message.kt
Amejia481: For the moment, the messages metadata is only been tracked on the apps to identify when a message should be shown of not, Nimbus only indicates the maximum times that a messages should be shown. We were using a `displayCount` of `0` to signal that the message was not display at that moment, as if we are initializing it with one ,it could indicate that a message is shown right away after creation, a message could be created without being shown, as could be provide by Nimbus, but not shown right away until all conditions are met :) .

app/src/main/java/org/mozilla/fenix/components/appstate/AppAction.kt
jonalmeida: You shouldn't need to respond with "no new messages", but _observe_ that the value `messageToShow` is now available and therefore should be displayed.

If the value is consumed (clicked/closed), the `mesageToShow` will be null and the UI will be told to remove anything on the screen.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
Amejia481: I made it public intentionally to be able to use the same behavior https://github.com/mozilla-mobile/fenix/pull/24225/files#diff-8e4963255d11a17aaae4c9eb198b24188b6dd43122a75a5fdf8b61e8b38f7457R55

app/src/main/java/org/mozilla/fenix/home/HomeFragmentStore.kt
```diff
@@ -59,7 +60,7 @@ data class HomeFragmentState(
     val mode: Mode = Mode.Normal,
     val topSites: List<TopSite> = emptyList(),
     val showCollectionPlaceholder: Boolean = false,
-    val showSetAsDefaultBrowserCard: Boolean = false,
+    val nimbusMessageCard: Message? = null,
```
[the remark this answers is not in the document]
Amejia481: 🎲 😅

app/src/main/java/org/mozilla/fenix/components/Components.kt
```diff
@@ -164,6 +167,20 @@ class Components(private val context: Context) {
         }
     }
 
+    val messageStorage by lazyMonitored {
```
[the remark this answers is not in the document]
Amejia481: Yeap :)

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
Amejia481: 😄

app/src/main/java/org/mozilla/fenix/gleanplumb/MessagingFeature.kt
Amejia481: Yes 👍🏽

app/src/main/java/org/mozilla/fenix/components/appstate/AppAction.kt
Amejia481: Yeah, for this reason is null-able :).
Do you mean that we need a separated action to consume ?

app/src/main/java/org/mozilla/fenix/gleanplumb/state/MessagingMiddleware.kt
```diff
@@ -0,0 +1,85 @@
...
+
+    override fun invoke(
+        context: MiddlewareContext<AppState, AppAction>,
+        next: (AppAction) -> Unit,
+        action: AppAction
+    ) {
+        when (action) {
+            is Evaluate -> {
+                val nextMessage = storage.getNextMessage()
+                context.dispatch(UpdateMessageToShow(nextMessage))
```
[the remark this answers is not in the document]
Amejia481: see https://github.com/mozilla-mobile/fenix/pull/24225#discussion_r830402471
Because we are using the same action to consume `mesageToShow` in case the conditions are not met anymore.

app/src/main/java/org/mozilla/fenix/gleanplumb/MessagingStorage.kt
```diff
@@ -0,0 +1,136 @@
...
+import org.mozilla.experiments.nimbus.GleanPlumbInterface
+import org.mozilla.experiments.nimbus.GleanPlumbMessageHelper
+import org.mozilla.experiments.nimbus.internal.FeatureHolder
+import org.mozilla.experiments.nimbus.internal.NimbusException
+import org.mozilla.fenix.components.AppStore
+import org.mozilla.fenix.components.appstate.AppAction.MessagingAction.UpdateMessages
+import org.mozilla.fenix.nimbus.Messaging
+import org.mozilla.fenix.nimbus.StyleData
+
+class MessagingStorage(
```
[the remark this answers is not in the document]
Amejia481: I was thinking about `NimbusMessagingProvider`
I'll cover it on https://github.com/mozilla-mobile/fenix/pull/24227

Amejia481:
I added tests ✅ , and reduce the scope of this PR, as the description of the PR indicates https://github.com/mozilla-mobile/fenix/pull/24225#issue-1166840692. I took out all the logic related to managing the nimbus messages, I will be covering that on https://github.com/mozilla-mobile/fenix/pull/24227, as the PR is already 31 files 😅  just by covering the core classes and UI interactions, It won't be ideal that PR to be a long tail  and un-reviewable XD

app/src/main/java/org/mozilla/fenix/gleanplumb/MessagingStorage.kt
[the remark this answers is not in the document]
jonalmeida: I'd recommend ending it with "Storage" because that's a naming convention we already use for features that work in similar ways (persistent storage); I think the pattern is now to fallback to "Provider" when the storage is complex and the storage has it's own inner storage layer.

app/src/main/java/org/mozilla/fenix/components/appstate/AppAction.kt
jonalmeida: Yeah, so when `MessageClicked` et al actions are observed, we tell the Nimbus storage to mark the message as consumed but we also need to tell the AppStore as well, so that both state solutions are in sync.

In this current design, we would remove the message from the storage and the "messageToShow" would still be available until our next call to `Evaluate` is made. 

> I took out all the logic related to managing the nimbus messages, I will be covering that on https://github.com/mozilla-mobile/fenix/pull/24227

With the middleware/storage gone from this PR, it's a bit harder to review how these parts are connected to others. Maybe these actions and state management should also be moved to the other PR too?

app/src/main/java/org/mozilla/fenix/components/appstate/AppAction.kt
```diff
@@ -16,4 +18,44 @@ sealed class AppAction : Action {
     data class AddNonFatalCrash(val crash: NativeCodeCrash) : AppAction()
     data class RemoveNonFatalCrash(val crash: NativeCodeCrash) : AppAction()
     object RemoveAllNonFatalCrashes : AppAction()
+
+    /**
+     * [Action]s related to interactions with the Messaging Framework.
+     */
+    sealed class MessagingAction : AppAction() {
```
jonalmeida, line 25: nit: this doesn't need to be a nested class since it already inherits from `AppAction`.

app/src/main/java/org/mozilla/fenix/components/appstate/AppAction.kt
Amejia481: Added this commit for easy to review https://github.com/mozilla-mobile/fenix/pull/24225/commits/86dbb8f875c164c5072d71b16929bb7444ee4ae7. I will committed as part of https://github.com/mozilla-mobile/fenix/pull/24227

Amejia481:
Added this commit https://github.com/mozilla-mobile/fenix/pull/24225/commits/86dbb8f875c164c5072d71b16929bb7444ee4ae7 for easy to review. I will committed as part of https://github.com/mozilla-mobile/fenix/pull/24227

app/src/main/java/org/mozilla/fenix/components/appstate/AppAction.kt
Amejia481: > Yeah, so when MessageClicked et al actions are observed, we tell the Nimbus storage to mark the message as consumed but we also need to tell the AppStore as well, so that both state solutions are in sync.

Sorry for the confusion, these bits are covered in the middleware, we are identifying if the message that was clicked/closed/pressed was the selected one if as we clear the `messageToShow`. 
In this current design, we would remove the message from the storage and the "messageToShow" would still be available until our next call to Evaluate is made.
https://github.com/mozilla-mobile/fenix/pull/24225/commits/86dbb8f875c164c5072d71b16929bb7444ee4ae7#diff-a4f1380abfbf1f76bc6bbd9ef961bd833145c4c75dae3f68dbff419ddd9d5d9cR87-R110

app/src/main/java/org/mozilla/fenix/gleanplumb/MessagingStorage.kt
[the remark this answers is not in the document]
Amejia481: 👍🏽

app/src/main/java/org/mozilla/fenix/components/Components.kt
```diff
@@ -196,6 +200,15 @@ class Components(private val context: Context) {
     val appStartReasonProvider by lazyMonitored { AppStartReasonProvider() }
     val startupActivityLog by lazyMonitored { StartupActivityLog() }
     val startupStateProvider by lazyMonitored { StartupStateProvider(startupActivityLog, appStartReasonProvider) }
+
+    val messagingStorage by lazyMonitored {
```
jonalmeida, line 204: nit: could we move this to the `Analytics` component since we have experiments there too.

app/src/main/java/org/mozilla/fenix/components/Components.kt
```diff
@@ -222,7 +234,8 @@ class Components(private val context: Context) {
                 PocketUpdatesMiddleware(
                     core.pocketStoriesService,
                     context.pocketStoriesSelectedCategoriesDataStore
-                )
+                ),
+                MessagingMiddleware(messagingStorage = messagingStorage)
```
jonalmeida, line 238: There is one thing we haven't considered yet, which is startup performance: we are now initializing and using experiments from the beginning. Even if we made this storage lazy, we _want_ messaging to work at cold boot I assume (right?), so the future of the `KeyPairMessageMetadataStorage` and how the messaging feature inits (off the main thread) might have an effect.

jonalmeida (approved):

app/src/main/java/org/mozilla/fenix/components/Components.kt
Amejia481: 👍🏽

app/src/main/java/org/mozilla/fenix/components/Components.kt
Amejia481: > Even if we made this storage lazy, we want messaging to work at cold boot I assume (right?)

Yeah, ideally. But a guess we have the same restrictions as normal experiments have, I think it's OK that messages will be shown after the second startup run.

Amejia481:
🎉 🎉 🎉 🎉 🎉 🎉

PR #24278, opened 2022-03-15
Closes #24276: updating pagination library, changing history related …

### LINKED ISSUE #24276: Update pagination library
It is a preparation for [...issues/24057](https://github.com/mozilla-mobile/fenix/issues/24057).
We use the 2.1.2 version of pagination library, that was released on March 18, 2020.
New design requirements for history screen (sticky headers, collapsible) require a more fresh version, to be able to work with Composable. 
The idea is to update it upfront and make it sure, that there is a safe option to fallback to, in case we decide to turn off synced history and the new design for history screen.

mavduevskiy:
Closes #24276 

Updating pagination library


mavduevskiy:
> The test failure in https://console.firebase.google.com/project/moz-fenix/testlab/histories/bh.66b7091e15d53d45/matrices/5989625405798366581 looks legitimate. Can you look at these?

Custom  `private val onZeroItemsLoaded: () -> Unit ` inside `PagingSource` caused racing condition, adding addLoadStateListener to `PagingDataAdapter` is the proper way of listening for zero items in the new Pagination lib.

Also, inside `HistoryView.update()` we lost the ability to check if the adapter has an initiated list or just empty: `updateEmptyState(state.pendingDeletionIds.size != historyAdapter.currentList?.size)` was not an option.

Made a nullable `adapterItemCount` to track that.

app/src/main/java/org/mozilla/fenix/components/history/PagedHistoryProvider.kt
```diff
@@ -113,15 +113,11 @@ class DefaultPagedHistoryProvider(
...
         offset: Int,
-        numberOfItems: Int,
-        onComplete: (List<HistoryDB>) -> Unit,
-    ) {
-        // A PagedList DataSource runs on a background thread automatically.
-        // If we run this in our own coroutineScope it breaks the PagedList
-        runBlockingIncrement {
+        numberOfItems: Int
+    ): List<HistoryDB> {
+        return runBlockingIncrement {
```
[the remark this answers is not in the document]
mavduevskiy: fixed, thanks!

app/src/main/java/org/mozilla/fenix/components/history/PagedHistoryProvider.kt
```diff
@@ -77,9 +77,9 @@ interface PagedHistoryProvider {
      *
      * @param offset How much to offset the list by
      * @param numberOfItems How many items to fetch
-     * @param onComplete A callback that returns the list of [HistoryDB]
+     * @return list of [HistoryDB]
      */
-    fun getHistory(offset: Int, numberOfItems: Int, onComplete: (List<HistoryDB>) -> Unit)
+    fun getHistory(offset: Int, numberOfItems: Int): List<HistoryDB>
```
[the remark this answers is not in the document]
mavduevskiy: fixed, thanks!

app/src/main/java/org/mozilla/fenix/library/history/HistoryView.kt
```diff
@@ -31,8 +34,20 @@ class HistoryView(
     var mode: HistoryFragmentState.Mode = HistoryFragmentState.Mode.Normal
         private set
 
-    val historyAdapter = HistoryAdapter(interactor)
+    val historyAdapter = HistoryAdapter(interactor).apply {
+        addLoadStateListener {
+            adapterItemCount = itemCount
+            Log.d("hehehaha", "adapterItemCount = $adapterItemCount")
```
gabrielluong, line 40: 😛

gabrielluong (changes requested):
This doesn't seem to be working. I am not seeing any history items on initial navigation to the History fragment while testing. However, it does appear if I do a pull to refresh.

https://user-images.githubusercontent.com/1190888/159528619-0ee4323d-076a-4011-bb9c-bdecfb44ee9e.mov

app/src/main/java/org/mozilla/fenix/library/history/HistoryView.kt
mavduevskiy: thanks for pointing that out! changing getHistory() method to suspend changed the order of update() calls. For some reasons tests were not affected

app/src/main/java/org/mozilla/fenix/library/history/HistoryDataSource.kt
```diff
@@ -5,38 +5,36 @@
...
 
-    override fun loadInitial(
-        params: LoadInitialParams<Int>,
-        callback: LoadInitialCallback<History>
-    ) {
-        historyProvider.getHistory(INITIAL_OFFSET, params.requestedLoadSize) { history ->
-            callback.onResult(history.positionWithOffset(INITIAL_OFFSET))
+    // params.key is expected to be null for the initial load or a refresh
+    override suspend fun load(params: LoadParams<Int>): LoadResult<Int, History> {
+        val offset = params.key ?: 0
```
gabrielluong, line 23: I am wondering if it should be 1 instead of 0 from looking at the sample code in https://developer.android.com/topic/libraries/architecture/paging/v3-paged-data#pagingsource

app/src/main/java/org/mozilla/fenix/library/history/HistoryView.kt
```diff
@@ -31,8 +33,25 @@ class HistoryView(
...
+            // could work correctly
+            if (itemCount != 0) {
+                adapterItemCount = itemCount
+            }
+            if (it.source.refresh is LoadState.NotLoading &&
+                it.append.endOfPaginationReached &&
+                itemCount < 1
+            ) {
+                adapterItemCount = 0
+                onZeroItemsLoaded.invoke()
```
gabrielluong, line 49: I can kind of see a path where we can remove the `userHasHistory` LiveData (we have been trying to remove LiveData afaik) and instead make use of the `HistoryFragmentStore`. Consequently,  `HistoryFragmentState` also has an unused `items` property, but you can easily dispatch an action to the `HistoryFragmentStore` to tell it whether or not we have any items and update the UI as appropriate. We can file this as a followup with the intent of removing `userHasHistory` LiveData and perhaps cleaning up `HistoryFragmentState`. I don't think we necessarily need to cache the list of history items, but maybe we can just have an equivalent to this `hasHistory` state there so that we can respond to any state changes without doing `model.userHasHistory.observe`.

app/src/main/java/org/mozilla/fenix/settings/sitepermissions/SitePermissionsExceptionsFragment.kt
```diff
@@ -58,27 +60,33 @@ class SitePermissionsExceptionsFragment :
...
+            val permissions = Pager(PagingConfig(MAX_ITEMS_PER_PAGE), null) {
+                dataSourceFactory.asPagingSourceFactory().invoke()
+            }.flow
+
+            permissions.collect {
+                adapter.submitData(it)
             }
         }
+
+        recyclerView.adapter = adapter
```
gabrielluong, line 89: Can we move this just after the adapter was initialized?

app/src/main/java/org/mozilla/fenix/library/history/HistoryViewModel.kt
```diff
@@ -4,34 +4,24 @@
...
-        datasource.value?.invalidate()
-    }
+    var history: Flow<PagingData<History>> = Pager(
+        PagingConfig(PAGE_SIZE),
+        null
+    ) {
+        HistoryDataSource(
+            historyProvider = historyProvider
+        )
+    }.flow
```
gabrielluong, line 24: Do we need `.cachedIn(viewModelScope)` here? https://developer.android.com/topic/libraries/architecture/paging/v3-paged-data#pagingdata-stream

gabrielluong (changes requested):
Sorry for the delay. I think this is a good first pass. Left you some comments to consider.

Can you check if we still need to manually call `binding.historyList.scrollToPosition(0)` in https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/java/org/mozilla/fenix/library/history/HistoryView.kt#49 with the paging changes as well?

mavduevskiy:
@gabrielluong 
Do you have an idea why did we need `binding.historyList.scrollToPosition(0)` in the first place? It seems a bit counterintuitive. When `RefreshListener` is called, adapter position can't be nothing but 0, right?
I have tested the current main, and have noticed no difference with or without this call. The new pagination library works visible better, but still, having or not having this call appears to make zero difference.

app/src/main/java/org/mozilla/fenix/components/history/PagedHistoryProvider.kt
```diff
@@ -77,9 +76,9 @@ interface PagedHistoryProvider {
      *
      * @param offset How much to offset the list by
      * @param numberOfItems How many items to fetch
-     * @param onComplete A callback that returns the list of [HistoryDB]
+     * @return list of [HistoryDB]
```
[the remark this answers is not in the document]
mavduevskiy: I wonder if this change actually makes more sense. The old call didn't name the list a paginated list, and the change I made was not changing the return type, but simple changing `onComplete() `callback to `@return`.
Also, technically, it does not return a paginated list, as there are no pages, but just a list of items with a given offset.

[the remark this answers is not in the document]
gabrielluong: I am okay with ignoring this one. I was going with the description from the interface "An Interface for providing a paginated list of [HistoryDB]."

gabrielluong:
> @gabrielluong Do you have an idea why did we need `binding.historyList.scrollToPosition(0)` in the first place? It seems a bit counterintuitive. When `RefreshListener` is called, adapter position can't be nothing but 0, right? I have tested the current main, and have noticed no difference with or without this call. The new pagination library works visible better, but still, having or not having this call appears to make zero difference.

Right, my intent was to see if we can just remove this call because I was also counting on the fact that it probably made no difference if we had it or not. Looking at git blame, led me to this the original PR https://github.com/mozilla-mobile/fenix/pull/11386. If you are satisfied with your testing and deem this call as unnecessary, I would say go ahead with removing it.

app/src/main/java/org/mozilla/fenix/library/history/HistoryViewModel.kt
gabrielluong: This is necessarily a suggestion to add it, but if it was decided that we don't need to cache the loaded data, we should try to understand what the implication of that is. The docs simply states `The cachedIn() operator makes the data stream shareable and caches the loaded data with the provided CoroutineScope`.

app/src/main/java/org/mozilla/fenix/settings/sitepermissions/SitePermissionsExceptionsFragment.kt
```diff
@@ -58,27 +60,33 @@ class SitePermissionsExceptionsFragment :
         recyclerView = rootView.findViewById(R.id.exceptions)
         recyclerView.layoutManager = LinearLayoutManager(requireContext())
 
+        val adapter = ExceptionsAdapter(this).apply {
+            addLoadStateListener { loadState ->
+                if (loadState.source.refresh is LoadState.NotLoading &&
```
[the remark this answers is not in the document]
mavduevskiy: Oh, it is the same problem we had on History screen: else condition got triggered by first call, where `itemsCount == 0`, but adapter isn't actually empty. Added an additional check, thanks! I couldn't see it on the emulator, but yeah, on a real device I can see the flickering. Thanks for pointing that out!

I kept the same logic it was before: loading == nothing, loaded empty list or something – change the state.
The link you provided, it looks like, follows the same logic: `binding.savedPasswordsEmptyView.isVisible` is false from xml and changes only when loading has stopped and there are actual results inside `state.loginList`. So, we are keeping logic in `SitePermissionsExceptionsFragment` in line with `SavedLoginsListView`.

app/src/main/java/org/mozilla/fenix/library/history/HistoryDataSource.kt
```diff
@@ -5,38 +5,36 @@
...
     private val historyProvider: PagedHistoryProvider
-) : ItemKeyedDataSource<Int, History>() {
+) : PagingSource<Int, History>() {
 
-    // Because the pagination is not based off of the key
-    // we want to start at 1, not 0 to be able to send the correct offset
-    // to the `historyProvider.getHistory` call.
-    override fun getKey(item: History): Int = item.position + 1
+    // having any value but null creates visual glitches in case or swipe to refresh and immediate
+    // scroll down
```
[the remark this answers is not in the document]
mavduevskiy: Rephrased, thanks! Didn't think about capitalizing, make sense though.

app/src/main/java/org/mozilla/fenix/library/history/HistoryDataSource.kt
```diff
@@ -5,38 +5,36 @@
...
+        } else {
+            (offset + historyItems.size) + 1
         }
-    }
-
-    override fun loadBefore(params: LoadParams<Int>, callback: LoadCallback<History>) { /* noop */ }
-
-    companion object {
-        internal const val INITIAL_OFFSET = 0
+        // prevKey is needed in case load would work upwards, so passing null is fine
```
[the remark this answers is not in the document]
mavduevskiy: Added `// Only paging forward.`. My fear is developers without experience could be confused by `prevKey = null`, it might look like scrolling up would not work. I actually got a comment like that already, so would like to bring some context from the doc to our codebase.

app/src/main/java/org/mozilla/fenix/library/history/HistoryDataSource.kt
```diff
@@ -5,38 +5,36 @@
...
+    // scroll down
+    override fun getRefreshKey(state: PagingState<Int, History>): Int? = null
 
-    override fun loadInitial(
-        params: LoadInitialParams<Int>,
-        callback: LoadInitialCallback<History>
-    ) {
-        historyProvider.getHistory(INITIAL_OFFSET, params.requestedLoadSize) { history ->
-            callback.onResult(history.positionWithOffset(INITIAL_OFFSET))
+    // params.key is expected to be null for the initial load or a refresh
```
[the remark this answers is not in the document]
mavduevskiy: Did move and changed text to this. Want to make it clear that null state is expected, and that the same parameter is passed in both reload and initial load.
```
// Get the offset of the last loaded page or default to 0 when it is null, on the initial
// load or a refresh.
```

app/src/main/java/org/mozilla/fenix/library/history/HistoryView.kt
```diff
@@ -61,7 +80,7 @@ class HistoryView(
 
         historyAdapter.updatePendingDeletionIds(state.pendingDeletionIds)
 
-        updateEmptyState(state.pendingDeletionIds.size != historyAdapter.currentList?.size)
+        updateEmptyState(state.pendingDeletionIds.size != adapterItemCount)
```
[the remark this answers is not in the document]
mavduevskiy: Yes, there is. I was also against using it, but it came out of necessity: `historyAdapter.itemCount` is not nullable as `historyAdapter.currentList?.size` was, we can't distinguish an empty state from not yet initialized. 
Using `historyAdapter.itemCount` cases a false call of `updateEmptyState(false)`.

app/src/main/java/org/mozilla/fenix/library/history/HistoryDataSource.kt
mavduevskiy: In the example data source supports pagination: `backend.searchUsers(query, nextPageNumber)`
, but in our case – `historyProvider.getHistory` operates using an offset. 
So the code actually operates with the same concept. That goes like 0/75/100/125 (in case there are no duplicates)
```
return LoadResult.Page(
            data = historyItems,
            prevKey = null, // Only paging forward.
            nextKey = nextOffset
        )
```

app/src/main/java/org/mozilla/fenix/settings/sitepermissions/SitePermissionsExceptionsFragment.kt
mavduevskiy: Yep, put that into  the`apply` block

app/src/main/java/org/mozilla/fenix/library/history/HistoryViewModel.kt
mavduevskiy: It looks like we do not really, it should help survive the orientation state, but it doesn't affect the fenix app because of this:
https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/AndroidManifest.xml#82
Now that I think of it, there is a little use for having a `ViewModel`, there is no needed to support the orientation change. It will work anyway.

app/src/main/java/org/mozilla/fenix/library/history/HistoryView.kt
mavduevskiy: opened an issue
https://github.com/mozilla-mobile/fenix/issues/24434

app/src/main/java/org/mozilla/fenix/library/history/HistoryDataSource.kt
```diff
@@ -5,38 +5,36 @@
...
+    override fun getRefreshKey(state: PagingState<Int, History>): Int? = null
 
-    override fun loadInitial(
-        params: LoadInitialParams<Int>,
-        callback: LoadInitialCallback<History>
-    ) {
-        historyProvider.getHistory(INITIAL_OFFSET, params.requestedLoadSize) { history ->
-            callback.onResult(history.positionWithOffset(INITIAL_OFFSET))
+    override suspend fun load(params: LoadParams<Int>): LoadResult<Int, History> {
+        // Get the offset of the last loaded page or default to 0 when it is null, on the initial
```
gabrielluong, line 22:
```suggestion
        // Get the offset of the last loaded page or default to 0 when it is null on the initial
```

app/src/main/java/org/mozilla/fenix/library/history/HistoryView.kt
```diff
@@ -31,8 +33,25 @@ class HistoryView(
...
-    val historyAdapter = HistoryAdapter(interactor)
+    val historyAdapter = HistoryAdapter(interactor).apply {
+        addLoadStateListener {
+            // first call will always have itemCount == 0, but we want to keep adapterItemCount
+            // as null until we can distinguish an empty list from populated, so updateEmptyState()
+            // could work correctly
+            if (itemCount != 0) {
+                adapterItemCount = itemCount
+            }
+            if (it.source.refresh is LoadState.NotLoading &&
```
gabrielluong, line 44: Can this be an `else if`? Otherwise, can you add a new line before this to separate the 2 if statement blocks so that it would be easier to read large blocks of code separately.

app/src/main/java/org/mozilla/fenix/library/history/HistoryView.kt
```diff
@@ -31,8 +33,25 @@ class HistoryView(
     var mode: HistoryFragmentState.Mode = HistoryFragmentState.Mode.Normal
         private set
 
-    val historyAdapter = HistoryAdapter(interactor)
+    val historyAdapter = HistoryAdapter(interactor).apply {
+        addLoadStateListener {
+            // first call will always have itemCount == 0, but we want to keep adapterItemCount
+            // as null until we can distinguish an empty list from populated, so updateEmptyState()
+            // could work correctly
```
gabrielluong, line 40: Can we fix up this comment by capitalizing and adding a period to the end.

app/src/main/java/org/mozilla/fenix/library/history/HistoryView.kt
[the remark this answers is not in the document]
gabrielluong: Ok, let's hope https://github.com/mozilla-mobile/fenix/issues/24434 can also address this

gabrielluong (commented):
I found 2 issues in my manual testing: 
- Sometimes I won't get the "No history here" when I manually delete all the history items
- I also get some phantom glitches which looks like swipes to refresh when I manually delete all the history items. See the video https://user-images.githubusercontent.com/1190888/160007687-0b0e6366-7a3e-4384-8c32-38e6b49ac40e.mov

app/src/main/java/org/mozilla/fenix/library/history/HistoryDataSource.kt
mavduevskiy: fixed

app/src/main/java/org/mozilla/fenix/library/history/HistoryView.kt
mavduevskiy: fixed

mavduevskiy:
@gabrielluong 
Thanks for pointing out on the glitches. Last comment should fix both of those.
The first was caused by the wrong key for pagination. I calculated it as offset + number of recived items, but groups broke that logic. There could be 10 items in a group, but it would count as a single item. So the offest was lagging behind, and items got duplicated.

The second was triggered by `update()` function inside HistoryView: 
`binding.progressBar.isVisible = state.isDeletingItems`
I am not sure if showing for a brief moment a progressbar, when the item is being deleted with a delay, is actually a helpful feedback, but I decided to keep it, just changed the constraints, so the TabsButton would not jump every time the progress bar was shown.

app/src/main/java/org/mozilla/fenix/settings/sitepermissions/SitePermissionsExceptionsFragment.kt
```diff
@@ -58,25 +60,32 @@ class SitePermissionsExceptionsFragment :
...
-                        }
+        val adapter = ExceptionsAdapter(this).apply {
+            addLoadStateListener { loadState ->
+                if (loadState.source.refresh is LoadState.NotLoading &&
+                    loadState.append.endOfPaginationReached &&
+                    itemCount < 1
+                ) {
+                    showEmptyListMessage()
+                } else {
+                    if (itemCount != 0) {
```
gabrielluong, line 71: Can we convert this to an appropriate else if conditional? Please retest afterwards.

gabrielluong (approved):
Thank you for addressing all the issues I pointed out. Let's land this on Monday to avoid landing since the change is somewhat risky for a Friday.

MozillaNoah:
looks like the UI tests failed, so those might need to be re-run or have a flaky test filed. clearing the landing flag just to keep the queue moving

Issue #23697, opened 2022-02-11
[Bug]: crash with  java.lang.NullPointerException in "org.mozilla.fenix.utils.ToolbarPopupWindow$$ExternalSyntheticLambda1.onClick(Unknown Source:38)"


dholbert:
### Steps to reproduce

Sorry, I unfortunately don't have concrete STR right now, but here's what I've got:

(1) Somehow, last night I ended up with a URL in my clipboard which Android would gladly paste into some apps (e.g. Firefox Nightly and Chrome) but not others (e.g. if I tried to paste into Gmail, I got an empty pasted string; and it'd crash Firefox-for-Android as described below).  The URL happened to be https://mashable.com/video/super-pumped-uber-trailer though I don't think the specific URL is important; rather, there was something special about my clipboard state that made it pastable-for-some-apps-but-not-others.  I unfortunately haven't been able to get back into that state, but I was quite-reliably in that state for a while (most of an hour as shown in my logcat) and I could reproduce this bug 100% reliably while I was in that state.

(2) With the aforementioned URL in my clipboard... Open Firefox or Firefox Beta.

(3) Long-press the URL bar and choose "Paste & Go"

### Expected behaviour

No crash.

### Actual behaviour

Immediate crash (the whole app is taken down).

### Device name

Google Pixel 4a

### Android version

Andoird 12

### Firefox release type

Firefox

### Firefox version

97.1.0

### Device logs

Here's the logcat output from when I crashed a bunch of times in a row when testing this with Firefox Release (currently version 97.1.0, though I was hitting this yesterday so it's possible I was on a slightly-earlier version)

Note that the last crash here is from beta, which I freshly installed from the Play Store last night to see if it was affected (and it was).  It reports itself as being version 97.0.0.beta-6 right now.  (Interesting that release and beta are both currently version 97.something; I assume beta is about to go to v98 but hasn't quite yet).

```
02-10 21:45:33.697 17304 17304 E AndroidRuntime: FATAL EXCEPTION: main
02-10 21:45:33.697 17304 17304 E AndroidRuntime: Process: org.mozilla.firefox, PID: 17304
02-10 21:45:33.697 17304 17304 E AndroidRuntime: java.lang.NullPointerException
02-10 21:45:33.697 17304 17304 E AndroidRuntime: 	at org.mozilla.fenix.utils.ToolbarPopupWindow$$ExternalSyntheticLambda1.onClick(Unknown Source:38)
[...]

PR #24294, opened 2022-03-16
For #23697: Fix up paste and paste-go handling for certain types of clipboard items

grigoryk:
This is a follow-up to https://github.com/mozilla-mobile/fenix/pull/24180 which adds handling for a few clipboard paste/paste-go cases. Specifically, I observed that in certain cases when copying links from, say, Chrome, we'd get MIME type `text/x-moz-url` for the clipboard item, which we weren't recognizing.

Additionally, second commit here allows pasting non-url contents of the clipboard (after stripping them).

Note that there's also a `text/uri-list` MIME type, used for passing around one or more URLs. Currently we don't support it - e.g. we'd need to make assumptions about its associated contents, etc. In my testing I haven't yet seen it used in context of a clipboard copy/paste.

### Pull Request checklist

1. click on Show All Checks,


grigoryk:
Tested this on a Pixel5 running Android 12, and a recent Samsung device.

app/src/main/java/org/mozilla/fenix/utils/ClipboardHandler.kt
```diff
@@ -32,11 +33,12 @@ class ClipboardHandler(val context: Context) {
      */
     var text: String?
```
gabrielluong, line 34: Does it make sense to update the tests in ClipboardHandlerTest?

grigoryk: Yeah, certainly, at the very least because they're failing :)

app/src/main/java/org/mozilla/fenix/utils/ToolbarPopupWindow.kt
```diff
@@ -81,12 +81,14 @@ object ToolbarPopupWindow {
             }
         }
 
-        clipboardUrl?.let { url ->
+        clipboardText?.let { text ->
```
gabrielluong, line 84: It would be nice to add some tests for this in ToolbarPopupWindowTest as well. Can we file a good-first-bug issue for this?

gabrielluong (approved):

Issue #21292, opened 2021-09-14
`MetricController.track` & `Event.wrapper` may use non-trivial power: called frequently


mcomella:
`MetricController.track` calls [the `Event.wrapper` method, which is a long chain of comparisons,](https://searchfox.org/mozilla-mobile/rev/14b6f198ebbe966e5d31eb9ab5fbf6b15fc1ce75/fenix/app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt#101) twice to submit telemetry. To test the impact of these methods, I did a simple workflow:
- click wiki homescreen
- click url bar, nav amazon.com
- click tabs tray, new tab
- search "foobar" and click default search result

I added custom markers and took profiles:
- Event.wrapper: https://share.firefox.dev/3lipdoU
- MetricController.track: https://share.firefox.dev/2Xha5QD (the profiler marker is misnamed GleanMetricsService)

In these profiles of this simple workflow, we see:
- `MetricController.track` is called 19 times for a total duration of 57.3ms
- `Event.wrapper` is called 36 times for a total of 24.3ms (I believe this is a subset of MetricController)
- (caveat: these duration may be inflated because these calls are fast and clock granularity might not be good enough)
- (note: I used `window.filteredMarkers.map((m) => m.end - m.start).reduce((acc, i) => acc + i)` to measure this)

These durations will be unnoticeable to the user, however, they will use battery. wrt `Event.wrapper`, this doesn't need to be a long chain of comparisons – perhaps by calling the Glean API directly and removing `MetricController` – so we can stop the users' device from doing work it doesn't need to. We should investigate `MetricController` for similar wins.

wrt prioritization, there may be larger gains we can do to improve battery life (e.g. https://github.com/mozilla-mobile/fenix/issues/15921) but those are likely harder to implement so it could be worth addressing this issue.

Note: https://github.com/mozilla-mobile/fenix/issues/19967 may also address this issue.

mcarare:
This issue will serve as a meta. I will link here the issues handling the removal of the wrapper in separate components. This will make QA work easier and ensure we do not have regressions by tackling the whole work in smaller pieces.

mcarare:
- [ ] Remove wrapper for Tabs Tray telemetry: https://github.com/mozilla-mobile/fenix/issues/24098

MatthewTighe:
👋 This work looks like great stuff! 

[...]

PR #24321, opened 2022-03-17
For #21292: Enable or disable Glean metrics upload directly.

mcarare:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
+++ b/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
@@ -322,7 +322,7 @@ open class FenixApplication : LocaleAwareApplication(), Provider {
 
     private fun startMetricsIfEnabled() {
         if (settings().isTelemetryEnabled) {
-            components.analytics.metrics.start(MetricServiceType.Data)
+            Glean.setUploadEnabled(true)
         }
 
         if (settings().isMarketingTelemetryEnabled) {


mcarare:
This should only be merged after all `metrics.track` calls have been replaced by direct Glean calls.
That is determined by the fact that the `MetricController` holds a list of "initialized" services and decides based on that list if the metric should be sent or not. (Even though Glean is very much capable of handling it's state on its own).

mcarare:
@rocketsroger `GleanMetricsService` is not used anymore after the wrapper removal. `AdjustMetricsService` was never used. But it was kept for eventual use in the future.

rocketsroger:
> @rocketsroger `GleanMetricsService` is not used anymore after the wrapper removal. `AdjustMetricsService` was never used. But it was kept for eventual use in the future.

But it looks like they are being started https://github.com/mozilla-mobile/fenix/blob/6a798ebfbc0a525d39780e22aae3f9ba29b08035/app/src/main/java/org/mozilla/fenix/components/Analytics.kt#L117

mcarare:
Yes, I kept that to easily allow using Adjust (Marketing metrics) to be used in the future if we want to. But as the track function is not used anymore it is redundant to keep the enable/disable state in `GleanMetricsService` when Glean manages this state internally.

rocketsroger:
> Yes, I kept that to easily allow using Adjust (Marketing metrics) to be used in the future if we want to. But as the track function is not used anymore it is redundant to keep the enable/disable state in `GleanMetricsService` when Glean manages this state internally.

But wouldn't this block will be skipped? https://github.com/mozilla-mobile/fenix/blob/6a798ebfbc0a525d39780e22aae3f9ba29b08035/app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt#L81

mcarare:
Closing this and keeping the refactoring for a later evaluation.

PR #24322, opened 2022-03-17
no issue: update accessibility comment for device icon

MatthewTighe:
1. click on Show All Checks,


MozillaNoah (approved):

Delphine:
Thanks for the update!
The comment still seems a bit unclear to me, would this work?
"Accessibility description for device icon used for recent synced tab"
Let me know if that's not what the comment should say, though ;) Thank you!

MatthewTighe:
@Delphine Yeah, that seems perfect! Updated with that suggestion

Issue #20414, opened 2021-07-18
Crash in [@ android.os.FileUriExposedException: file:/// exposed beyond app through Intent.getData()]


DavidCRicardo:
### Steps to reproduce

Open a pdf file on my firefox desktop
Send tab to device. 
The firefox on the device always crashes with this process.

### Expected behaviour

I can see the tab on my device 

### Actual behaviour

A message box should open and say that you need to restart Firefox on your device

### Android version

Android 7

### Firefox release type

Firefox

### Firefox version

90.1.1

Amejia481:
👋🏽 
Thanks for reporting @DavidCRicardo , if it's a public resource would you mind sharing the link?  :)

kbrosnan:
The specific file/url does not matter.  Any file:/// URI will crash Firefox.

```
android.os.FileUriExposedException: file:///[REDACTED] exposed beyond app through Intent.getData()
	at android.os.StrictMode.onFileUriExposed(StrictMode.java:2141)
	at android.net.Uri.checkFileUriExposed(Uri.java:2391)
	at android.content.Intent.prepareToLeaveProcess(Intent.java:11152)
	at android.content.Intent.prepareToLeaveProcess(Intent.java:11104)
	at android.app.PendingIntent.getActivity(PendingIntent.java:355)
	at android.app.PendingIntent.getActivity(PendingIntent.java:317)
	at org.mozilla.fenix.components.NotificationManager.showReceivedTabs(NotificationManager.kt:6)
	at org.mozilla.fenix.components.BackgroundServices$makeAccountManager$$inlined$also$lambda$1.invoke(BackgroundServices.kt:3)
	at mozilla.components.feature.accounts.push.EventsObserver.onEvents(SendTabFeature.kt:7)
	at mozilla.components.service.fxa.-$$LambdaGroup$ks$ro7rEtSJwcc4uzMXErsGW2kmug8.invoke(com.android.tools.r8.jetbrains.kotlin-style lambda group:8)
	at mozilla.components.support.base.observer.ObserverRegistry.notifyObservers(ObserverRegistry.kt:4)
	at mozilla.components.service.fxa.AccountEventsIntegration.onEvents(AccountStorage.kt:2)
	at mozilla.components.service.fxa.-$$LambdaGroup$ks$ro7rEtSJwcc4uzMXErsGW2kmug8.invoke(com.android.tools.r8.jetbrains.kotlin-style lambda group:3)
[...]

PR #24326, opened 2022-03-18
Don't crash when receiving a bad uri via send tab

skhamis:
fixes https://github.com/mozilla-mobile/fenix/issues/20414 

Ideally this shouldn't have been possible since clients should be not allowing certain URI schemes from being send-capable. With synced tabs we follow this pref: https://searchfox.org/mozilla-central/source/modules/libpref/init/all.js#4372 which I carried over and inputted for android. 

### Pull Request checklist

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/components/NotificationManager.kt
```diff
@@ -129,3 +133,10 @@ class NotificationManager(private val context: Context) {
         return this
     }
 }
+
+internal fun isValidTabSchema(tab: TabData): Boolean {
```
mcarare, line 137: Are we planning to use this just as a workaround until the main cause is fixed? (i.e. not syncing these schemas?)

skhamis: I think this would be a semi-permanent solution as it could be a very long-tail to ensure no version at some point could send a file:// across all browser platforms. Though I do think it's a relatively straightforward solution to just copy what desktop is doing for now but happy to brainstorm a different approach if we feel strongly against this.

skhamis:
Hi @mcarare, how are we feeling about this PR? Anything I need to do to continue moving this one along?

mcarare:
@skhamis The solution presented here is ok. I just feel the fix should not be needed at all if the problem is fixed at its origin.

skhamis:
> @skhamis The solution presented here is ok. I just feel the fix should not be needed at all if the problem is fixed at its origin.

Totally understand the concern @mcarare! While the team will indeed fix underlying issue of having `file://` be an option for send push on Desktop, iOS and Android. This solution should [prevent current users from crashing](https://github.com/mozilla-mobile/fenix/issues/20414) since it'll take quite awhile (most likely months) before those fixes go out to users and we reach a point where users can't send invalid schemes. 

Alternatively, we can indeed just sit/close this PR for now and wait until the send-tab fixes land on the respective repos -- unfortunately, users will still be crashing if they send invalid schemas until then but depending on the need/want to have that I'll leave to the Fenix team!

mhammond:
> @skhamis The solution presented here is ok. I just feel the fix should not be needed at all if the problem is fixed at its origin.

@mcarare further to what @skhamis said, the problem isn't just that the fix is only on Nightly (and thus will not be on release for a couple of months), there's a very long tail of desktop versions - not only is there ESR users, it's not that uncommon to find desktop users 10 versions or so behind. So while this is your call, you should expect that it will be *at least* 6 months before you can be confident that remote Firefoxes crashing Fenix is rare.

mcarare (approved):

PR #24046, opened 2022-03-01
For #23388 - Update scrim to match design system

gabrielluong:
Fixes #23388 

Light Mode:
Old|New
---|---
Screen Shot 2022-03-01 at 11 15 50 AM|Screen Shot 2022-03-01 at 11 15 56 AM
Screen Shot 2022-03-01 at 11 21 53 AM|Screen Shot 2022-03-01 at 11 22 04 AM
Screen Shot 2022-03-01 at 11 24 22 AM|Screen Shot 2022-03-01 at 11 25 28 AM
Screen Shot 2022-03-01 at 11 24 33 AM|Screen Shot 2022-03-01 at 11 26 12 AM

Dark Mode:
Old|New
---|---
Screen Shot 2022-03-01 at 11 16 51 AM|Screen Shot 2022-03-01 at 11 16 27 AM
Screen Shot 2022-03-01 at 11 20 34 AM|Screen Shot 2022-03-01 at 11 20 47 AM
Screen Shot 2022-03-01 at 11 28 42 AM|Screen Shot 2022-03-01 at 11 29 00 AM
Screen Shot 2022-03-01 at 11 28 51 AM|Screen Shot 2022-03-01 at 11 29 07 AM

### Pull Request checklist

- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features. In addition, it includes a screenshot of a successful [accessibility scan](https://play.google.com/store/apps/details?id=com.google.android.apps.accessibility.auditor&hl=en_US) to ensure no new defects are added to the product.

### To download an APK when reviewing a PR:
1. click on Show All Checks,
2. click Details next to "Taskcluster (pull_request)" after it appears and then finishes with a green checkmark,
3. click on the "Fenix - assemble" task, then click "Run Artifacts".
4. the APK links should be on the left side of the screen, named for each CPU architecture

PR #24374, opened 2022-03-22
For #24307 - Implement synced history navigation and feature flag

### LINKED ISSUE #24307: Implement synced history navigation and feature flag
It is a part of [..issues/24057](https://github.com/mozilla-mobile/fenix/issues/...issues/24057).

mavduevskiy:
Closes #24307 
Added a new button to the History screen, which leads to synced fragment. 
`HistoryProvider` is using nullable`isRemote` parameter to filter or pass everything.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/FeatureFlags.kt
+++ b/app/src/main/java/org/mozilla/fenix/FeatureFlags.kt
@@ -88,6 +88,11 @@ object FeatureFlags {
     const val historyImprovementFeatures = true
 
     /**
+     * Separates history into local and synced from other sources.
+     */
+    val showSyncedHistory = Config.channel.isDebug
+
+    /**
      * Enables the Task Continuity enhancements.
      */
     val taskContinuityFeature = Config.channel.isDebug
--- a/app/src/main/java/org/mozilla/fenix/components/history/PagedHistoryProvider.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/history/PagedHistoryProvider.kt
@@ -75,9 +75,10 @@ interface PagedHistoryProvider {
      *
      * @param offset How much to offset the list by
      * @param numberOfItems How many items to fetch
+     * @param isRemote Are items local or synced from other devices
      * @return list of [HistoryDB]
      */
-    suspend fun getHistory(offset: Int, numberOfItems: Int): List<HistoryDB>
+    suspend fun getHistory(offset: Int, numberOfItems: Int, isRemote: Boolean? = null): List<HistoryDB>
 }
 
 /**
@@ -109,11 +110,13 @@ class DefaultPagedHistoryProvider(
         it == VisitType.REDIRECT_PERMANENT || it == VisitType.REDIRECT_TEMPORARY
     }
 
-    @Volatile private var historyGroups: List<HistoryDB.Group>? = null
+    @Volatile
+    private var historyGroups: List<HistoryDB.Group>? = null
 
     override suspend fun getHistory(
         offset: Int,
-        numberOfItems: Int
+        numberOfItems: Int,
+        isRemote: Boolean?
     ): List<HistoryDB> {
         // We need to re-fetch all the history metadata if the offset resets back at 0
         // in the case of a pull to refresh.
@@ -140,7 +143,7 @@ class DefaultPagedHistoryProvider(
                 .toList()
         }
 
-        return getHistoryAndSearchGroups(offset, numberOfItems)
+        return getHistoryAndSearchGroups(offset, numberOfItems, isRemote)
     }
 
     /**
@@ -162,6 +165,7 @@ class DefaultPagedHistoryProvider(
     private suspend fun getHistoryAndSearchGroups(
         offset: Int,
         numberOfItems: Int,
+        isRemote: Boolean?
     ): List<HistoryDB> {
         val result = mutableListOf<HistoryDB>()
         var history: List<HistoryDB.Regular> = historyStorage
@@ -170,6 +174,11 @@ class DefaultPagedHistoryProvider(
                 numberOfItems.toLong(),
                 excludeTypes = excludedVisitTypes
             )
+            .filter { item ->
+                isRemote?.let {
+                    item.isRemote == it
+                } ?: true
+            }
             .map { transformVisitInfoToHistoryItem(it) }
 
         // We'll use this list to filter out redirects from metadata groups below.
--- a/app/src/main/java/org/mozilla/fenix/library/history/HistoryController.kt
+++ b/app/src/main/java/org/mozilla/fenix/library/history/HistoryController.kt
@@ -34,6 +34,10 @@ interface HistoryController {
     fun handleDeleteSome(items: Set<History>)
     fun handleRequestSync()
     fun handleEnterRecentlyClosed()
+    /**
+     * Navigates to [org.mozilla.fenix.library.syncedhistory.SyncedHistoryFragment]
+     */
+    fun handleEnterSyncedHistory()
 }
 
 @Suppress("TooManyFunctions", "LongParameterList")
@@ -98,8 +102,7 @@ class DefaultHistoryController(
     }
 
     override fun handleSearch() {
-        val directions =
-            HistoryFragmentDirections.actionGlobalHistorySearchDialog()
+        val directions = HistoryFragmentDirections.actionGlobalHistorySearchDialog()
         navController.navigateSafe(R.id.historyFragment, directions)
     }
 
@@ -157,5 +160,11 @@ class DefaultHistoryController(
             NavOptions.Builder().setPopUpTo(R.id.recentlyClosedFragment, true).build()
         )
         Events.recentlyClosedTabsOpened.record(NoExtras())
+    }
+
+    override fun handleEnterSyncedHistory() {
+        navController.navigate(
+            HistoryFragmentDirections.actionHistoryFragmentToSyncedHistoryFragment()
+        )
     }
 }
--- a/app/src/main/java/org/mozilla/fenix/library/history/HistoryView.kt
+++ b/app/src/main/java/org/mozilla/fenix/library/history/HistoryView.kt
@@ -12,6 +12,7 @@ import androidx.paging.LoadState
 import androidx.recyclerview.widget.LinearLayoutManager
 import androidx.recyclerview.widget.SimpleItemAnimator
 import mozilla.components.support.base.feature.UserInteractionHandler
+import org.mozilla.fenix.FeatureFlags
 import org.mozilla.fenix.R
 import org.mozilla.fenix.databinding.ComponentHistoryBinding
 import org.mozilla.fenix.ext.components
@@ -114,6 +115,8 @@ class HistoryView(
     private fun updateEmptyState(userHasHistory: Boolean) {
         binding.historyList.isInvisible = !userHasHistory
         binding.historyEmptyView.isVisible = !userHasHistory
+        binding.topSpacer.isVisible = !userHasHistory
+
         with(binding.recentlyClosedNavEmpty) {
             recentlyClosedNav.setOnClickListener {
                 interactor.onRecentlyClosedClicked()
@@ -131,6 +134,14 @@ class HistoryView(
             )
             recentlyClosedNav.isVisible = !userHasHistory
         }
+
+        with(binding.syncedHistoryNavEmpty) {
+            syncedHistoryNav.setOnClickListener {
+                interactor.onSyncedHistoryClicked()
+            }
+            syncedHistoryNav.isVisible = FeatureFlags.showSyncedHistory && !userHasHistory
+        }
+
         if (!userHasHistory) {
             binding.historyEmptyView.announceForAccessibility(context.getString(R.string.history_empty_message))
         }
--- a/app/src/main/java/org/mozilla/fenix/library/history/viewholders/HistoryListItemViewHolder.kt
+++ b/app/src/main/java/org/mozilla/fenix/library/history/viewholders/HistoryListItemViewHolder.kt
@@ -7,6 +7,7 @@ package org.mozilla.fenix.library.history.viewholders
 import android.view.View
 import androidx.core.view.isVisible
 import androidx.recyclerview.widget.RecyclerView
+import org.mozilla.fenix.FeatureFlags
 import org.mozilla.fenix.R
 import org.mozilla.fenix.databinding.HistoryListItemBinding
 import org.mozilla.fenix.ext.components
@@ -31,6 +32,10 @@ class HistoryListItemViewHolder(
     init {
         binding.recentlyClosedNavEmpty.recentlyClosedNav.setOnClickListener {
             historyInteractor.onRecentlyClosedClicked()
+        }
+
+        binding.syncedHistoryNavEmpty.syncedHistoryNav.setOnClickListener {
+            historyInteractor.onSyncedHistoryClicked()
         }
 
         binding.historyLayout.overflowView.apply {
@@ -113,11 +118,15 @@ class HistoryListItemViewHolder(
         }
     }
 
+    @Suppress("NestedBlockDepth")
     private fun toggleTopContent(
         showTopContent: Boolean,
         isNormalMode: Boolean,
     ) {
         binding.recentlyClosedNavEmpty.recentlyClosedNav.isVisible = showTopContent
+        binding.topSpacer.isVisible = showTopContent
+        binding.bottomSpacer.isVisible = showTopContent
+        binding.syncedHistoryNavEmpty.syncedHistoryNav.isVisible = showTopContent && FeatureFlags.showSyncedHistory
 
         if (showTopContent) {
             val numRecentTabs = itemView.context.components.core.store.state.closedTabs.size
@@ -131,6 +140,7 @@ class HistoryListItemViewHolder(
                 ),
                 numRecentTabs
             )
+
             binding.recentlyClosedNavEmpty.recentlyClosedNav.run {
                 if (isNormalMode) {
                     isEnabled = true
@@ -138,6 +148,18 @@ class HistoryListItemViewHolder(
                 } else {
                     isEnabled = false
                     alpha = DISABLED_BUTTON_ALPHA
+                }
+            }
+
+            if (FeatureFlags.showSyncedHistory) {
+                binding.syncedHistoryNavEmpty.syncedHistoryNav.run {
+                    if (isNormalMode) {
+                        isEnabled = true
+                        alpha = 1f
+                    } else {
+                        isEnabled = false
+                        alpha = DISABLED_BUTTON_ALPHA
+                    }
                 }
             }
         }
--- a/app/src/main/res/layout/component_history.xml
+++ b/app/src/main/res/layout/component_history.xml
@@ -20,12 +20,26 @@
         app:layout_constraintStart_toStartOf="parent"
         app:layout_constraintTop_toTopOf="parent" />
 
+    <View
+        android:id="@+id/top_spacer"
+        android:layout_width="match_parent"
+        android:layout_height="8dp"
+        app:layout_constraintTop_toTopOf="parent"
+        android:visibility="gone"/>
+
     <include
         android:id="@+id/recently_closed_nav_empty"
         layout="@layout/recently_closed_nav_item"
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
-        app:layout_constraintTop_toTopOf="parent" />
+        app:layout_constraintTop_toBottomOf="@+id/top_spacer" />
+
+    <include
+        android:id="@+id/synced_history_nav_empty"
+        layout="@layout/synced_history_nav_item"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        app:layout_constraintTop_toBottomOf="@id/recently_closed_nav_empty" />
 
     <TextView
         android:id="@+id/history_empty_view"
@@ -39,14 +53,14 @@
         app:layout_constraintBottom_toBottomOf="parent"
         app:layout_constraintEnd_toEndOf="parent"
         app:layout_constraintStart_toStartOf="parent"
-        app:layout_constraintTop_toBottomOf="@id/recently_closed_nav_empty" />
+        app:layout_constraintTop_toBottomOf="@id/synced_history_nav_empty" />
 
     <androidx.swiperefreshlayout.widget.SwipeRefreshLayout
         android:id="@+id/swipe_refresh"
         android:layout_width="match_parent"
         android:layout_height="0dp"
         app:layout_constraintBottom_toBottomOf="parent"
-        app:layout_constraintTop_toBottomOf="@id/recently_closed_nav_empty">
+        app:layout_constraintTop_toBottomOf="@id/synced_history_nav_empty">
 
         <androidx.recyclerview.widget.RecyclerView
             android:id="@+id/history_list"
--- a/app/src/main/res/layout/fragment_synced_history.xml
+++ b/app/src/main/res/layout/fragment_synced_history.xml
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+
+<FrameLayout
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:id="@+id/historyLayout"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    tools:context="org.mozilla.fenix.library.syncedhistory.SyncedHistoryFragment" />
--- a/app/src/main/res/layout/history_list_item.xml
+++ b/app/src/main/res/layout/history_list_item.xml
@@ -9,9 +9,25 @@
     android:importantForAccessibility="no"
     android:orientation="vertical">
 
+    <View
+        android:id="@+id/top_spacer"
+        android:layout_width="match_parent"
+        android:layout_height="8dp"
+        android:visibility="gone"/>
+
     <include
         android:id="@+id/recently_closed_nav_empty"
         layout="@layout/recently_closed_nav_item" />
+
+    <include
+        android:id="@+id/synced_history_nav_empty"
+        layout="@layout/synced_history_nav_item" />
+
+    <View
+        android:id="@+id/bottom_spacer"
+        android:layout_width="match_parent"
+        android:layout_height="32dp"
+        android:visibility="gone"/>
 
     <TextView
         android:id="@+id/header_title"
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -766,6 +766,10 @@
     <string name="history_older">Older</string>
     <!-- Text shown when no history exists -->
     <string name="history_empty_message">No history here</string>
+    <!-- Text for the button inside History screen that opens Synced History screen. History that is coming from other devices -->
+    <string name="history_synced_from_other_devices">Synced from other devices</string>
+    <!-- The page title for browsing history coming from other devices. -->
+    <string name="history_from_other_devices">From other devices</string>
 
     <!-- Downloads -->
     <!-- Text for the snackbar to confirm that multiple downloads items have been removed -->
--- a/app/src/main/java/org/mozilla/fenix/library/history/HistoryDataSource.kt
+++ b/app/src/main/java/org/mozilla/fenix/library/history/HistoryDataSource.kt
@@ -15,7 +15,8 @@ import org.mozilla.fenix.components.history.PagedHistoryProvider
  * Flow<PagingData>, that provides HistoryAdapter with items to display.
  */
 class HistoryDataSource(
-    private val historyProvider: PagedHistoryProvider
+    private val historyProvider: PagedHistoryProvider,
+    private val isRemote: Boolean? = null,
 ) : PagingSource<Int, History>() {
 
     // The refresh key is set to null so that it will always reload the entire list for any data
@@ -26,7 +27,7 @@ class HistoryDataSource(
         // Get the offset of the last loaded page or default to 0 when it is null on the initial
         // load or a refresh.
         val offset = params.key ?: 0
-        val historyItems = historyProvider.getHistory(offset, params.loadSize).run {
+        val historyItems = historyProvider.getHistory(offset, params.loadSize, isRemote).run {
             positionWithOffset(offset)
         }
         val nextOffset = if (historyItems.isEmpty()) {
--- a/app/src/main/java/org/mozilla/fenix/library/history/HistoryFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/library/history/HistoryFragment.kt
@@ -68,7 +68,8 @@ class HistoryFragment : LibraryPageFragment<History>(), UserInteractionHandler {
         null
     ) {
         HistoryDataSource(
-            historyProvider = historyProvider
+            historyProvider = historyProvider,
+            isRemote = if (FeatureFlags.showSyncedHistory) false else null
         )
     }.flow
 
--- a/app/src/main/java/org/mozilla/fenix/library/history/HistoryInteractor.kt
+++ b/app/src/main/java/org/mozilla/fenix/library/history/HistoryInteractor.kt
@@ -47,6 +47,11 @@ interface HistoryInteractor : SelectionInteractor<History> {
      * Called when the user clicks on recently closed tab button.
      */
     fun onRecentlyClosedClicked()
+
+    /**
+     * Called when the user clicks on synced history button.
+     */
+    fun onSyncedHistoryClicked()
 }
 
 /**
@@ -95,5 +100,9 @@ class DefaultHistoryInteractor(
 
     override fun onRecentlyClosedClicked() {
         historyController.handleEnterRecentlyClosed()
+    }
+
+    override fun onSyncedHistoryClicked() {
+        historyController.handleEnterSyncedHistory()
     }
 }
--- a/app/src/main/java/org/mozilla/fenix/library/syncedhistory/SyncedHistoryFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/library/syncedhistory/SyncedHistoryFragment.kt
@@ -1,0 +1,22 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.library.syncedhistory
+
+import mozilla.components.support.base.feature.UserInteractionHandler
+import org.mozilla.fenix.library.LibraryPageFragment
+import org.mozilla.fenix.library.history.History
+
+/**
+ * A screen displaying history items that were opened on other devices, not local.
+ */
+class SyncedHistoryFragment : LibraryPageFragment<History>(), UserInteractionHandler {
+
+    override fun onBackPressed(): Boolean {
+        return false
+    }
+
+    override val selectedItems: Set<History>
+        get() = setOf()
+}
--- a/app/src/main/res/layout/add_new_collection_dialog.xml
+++ b/app/src/main/res/layout/add_new_collection_dialog.xml
@@ -10,7 +10,7 @@
     android:orientation="vertical">
 
     <View
-        android:id="@+id/top_divider"
+        android:id="@+id/top_spacer"
         android:layout_width="match_parent"
         android:layout_height="1dp"
         android:layout_marginTop="16dp"
@@ -29,7 +29,7 @@
     </ScrollView>
 
     <View
-        android:id="@+id/bottom_divider"
+        android:id="@+id/bottom_spacer"
         android:layout_width="match_parent"
         android:layout_height="1dp"
         android:layout_marginBottom="8dp"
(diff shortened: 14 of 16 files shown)


gabrielluong:
Let's go through https://github.com/mozilla-mobile/fenix/pull/24278 first and have that landed before reviewing this PR.

gabrielluong:
Please rebase this before the review.

app/src/main/java/org/mozilla/fenix/library/history/HistoryFragmentStore.kt
```diff
@@ -105,7 +105,7 @@ fun HistoryMetadata.toHistoryMetadata(position: Int): History.Metadata {
 /**
  * The [Store] for holding the [HistoryFragmentState] and applying [HistoryFragmentAction]s.
  */
-class HistoryFragmentStore(initialState: HistoryFragmentState) :
+open class HistoryFragmentStore(initialState: HistoryFragmentState) :
```
gabrielluong, line 108: This seems unnecessary and I don't expect anyone to override this.

app/src/main/java/org/mozilla/fenix/library/history/HistoryController.kt
```diff
@@ -25,10 +25,14 @@ interface HistoryController {
...
     fun handleEnterRecentlyClosed()
+    /**
+     * Navigates to [org.mozilla.fenix.library.syncedhistory.SyncedHistoryFragment]
+     */
+    fun handleEnterSyncedHistory()
 }
 
 @Suppress("TooManyFunctions")
-class DefaultHistoryController(
+open class DefaultHistoryController(
```
gabrielluong, line 35: This seems unnecessary and I don't expect anyone to override this.

app/src/main/java/org/mozilla/fenix/FeatureFlags.kt
```diff
@@ -89,6 +89,11 @@ object FeatureFlags {
      */
     val historyImprovementFeatures = Config.channel.isNightlyOrDebug
 
+    /**
+     * Separates history into local and synced from other sources.
+     */
+    val syncedHistoryFeature = Config.channel.isNightlyOrDebug
```
gabrielluong, line 95:
```suggestion
    val syncedHistoryFeature = Config.channel.isDebug
```

Probably too early to enable for Nightly.

app/src/main/java/org/mozilla/fenix/components/history/PagedHistoryProvider.kt
```diff
@@ -77,7 +77,7 @@ interface PagedHistoryProvider {
      * @param numberOfItems How many items to fetch
      * @return list of [HistoryDB]
      */
-    suspend fun getHistory(offset: Int, numberOfItems: Int): List<HistoryDB>
+    suspend fun getHistory(offset: Int, numberOfItems: Int, isRemote: Boolean? = null): List<HistoryDB>
```
gabrielluong, line 80: Add the @param for isRemote

app/src/main/java/org/mozilla/fenix/library/history/HistoryAdapter.kt
```diff
@@ -60,7 +62,16 @@ class HistoryAdapter(
             timeGroup = current.historyTimeGroup
         }
 
-        holder.bind(current, timeGroup, position == 0, mode, isPendingDeletion)
+        holder.bind(
+            current,
+            BindingParams(
```
gabrielluong, line 67: Given my above comment, adding a `BindingParams` data class seems like an unnecessary refactoring here especially since we don't need to pass `syncedHistoryVisible` like this. I would prefer if we just kept it the way it is to avoid too many changes that will overwrite the prior history with unwrapping the parameters within `bind`

app/src/main/java/org/mozilla/fenix/library/history/HistoryAdapter.kt
```diff
@@ -16,6 +17,7 @@ import org.mozilla.fenix.library.history.viewholders.HistoryListItemViewHolder
  */
 class HistoryAdapter(
     private val historyInteractor: HistoryInteractor,
+    private val syncedHistoryVisible: Boolean,
```
gabrielluong, line 20: I don't think we need to pass this here. We can call `FeatureFlags.syncedHistoryFeature` from anywhere so we don't need to pass this around like this.

app/src/main/java/org/mozilla/fenix/library/history/HistoryController.kt
```diff
@@ -25,10 +25,14 @@ interface HistoryController {
     fun handleDeleteSome(items: Set<History>)
     fun handleRequestSync()
     fun handleEnterRecentlyClosed()
+    /**
+     * Navigates to [org.mozilla.fenix.library.syncedhistory.SyncedHistoryFragment]
```
gabrielluong, line 29:
```suggestion
     * Navigates to [SyncedHistoryFragment].
```

app/src/main/java/org/mozilla/fenix/library/history/HistoryView.kt
```diff
@@ -131,6 +135,13 @@ class HistoryView(
             )
             recentlyClosedNav.isVisible = !userHasHistory
         }
+        with(binding.syncedHistoryNavEmpty) {
+            val syncedHistoryVisible = FeatureFlags.syncedHistoryFeature && !userHasHistory
+            syncedHistoryNav.setOnClickListener {
+                interactor.onSyncedHistoryClicked()
+            }
+            syncedHistoryNav.isVisible = syncedHistoryVisible
+        }
```
gabrielluong, line 144:
```suggestion

        with(binding.syncedHistoryNavEmpty) {
            val syncedHistoryVisible = FeatureFlags.syncedHistoryFeature && !userHasHistory
            syncedHistoryNav.setOnClickListener {
                interactor.onSyncedHistoryClicked()
            }
            syncedHistoryNav.isVisible = syncedHistoryVisible
        }

```

Add a new line before and after these kinda with/if/when blocks to the code some separation.

app/src/main/java/org/mozilla/fenix/library/history/HistoryView.kt
```diff
@@ -131,6 +135,13 @@ class HistoryView(
             )
             recentlyClosedNav.isVisible = !userHasHistory
         }
+        with(binding.syncedHistoryNavEmpty) {
+            val syncedHistoryVisible = FeatureFlags.syncedHistoryFeature && !userHasHistory
```
gabrielluong, line 139:
```suggestion
            val syncedHistoryVisible = FeatureFlags.syncedHistoryFeature && userHasHistory
```
Shouldn't we want users to have history?

app/src/main/java/org/mozilla/fenix/library/history/viewholders/HistoryListItemViewHolder.kt
```diff
@@ -43,14 +47,12 @@ class HistoryListItemViewHolder(
         }
     }
 
+    @Suppress("UndocumentedPublicFunction")
```
gabrielluong, line 50: I think we can just this without adding the `BindingParams`. There's also no need to change this since you can call the FeatureFlag from here.

app/src/main/java/org/mozilla/fenix/library/history/viewholders/HistoryListItemViewHolder.kt
```diff
@@ -133,6 +141,21 @@ class HistoryListItemViewHolder(
                     alpha = DISABLED_BUTTON_ALPHA
                 }
             }
+            if (syncedHistoryVisible) {
```
gabrielluong, line 144: Just call `FeatureFlag.syncedHistoryFeature` here

app/src/main/java/org/mozilla/fenix/library/history/viewholders/HistoryListItemViewHolder.kt
```diff
@@ -133,6 +141,21 @@ class HistoryListItemViewHolder(
...
+
+    private fun toggleTopContent(isNormalMode: Boolean) {
+        binding.syncedHistoryNavEmpty.syncedHistoryNav.run {
+            if (isNormalMode) {
+                isEnabled = true
+                alpha = 1f
+            } else {
+                isEnabled = false
+                alpha = DISABLED_BUTTON_ALPHA
+            }
```
gabrielluong, line 158: I would just inline this since it's confusing to see this, but also know that the `binding.recentlyClosedNavEmpty.recentlyClosedNav.run {}` block does the same thing but is not in its own function.

app/src/main/res/drawable/ic_synced_history.xml
```diff
@@ -0,0 +1,14 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
gabrielluong, line 1: This icon should already exist in the codebase since it's used in the tabs tray - `ic_synced_tabs`. I would use that for now and we can look to see if we should be updating it as a followup.

app/src/main/res/layout/component_history.xml
```diff
@@ -25,8 +26,17 @@
...
         android:layout_height="wrap_content"
+        android:layout_marginTop="8dp"
         app:layout_constraintTop_toTopOf="parent" />
 
+    <include
+        android:id="@+id/synced_history_nav_empty"
+        layout="@layout/synced_history_nav_item"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:layout_marginTop="8dp"
```
gabrielluong, line 37: There doesn't seem to be any additional spacing between list items. Can you send a screenshot of the figma inspection where you are seeing this 8dp in between the items?
Screen Shot 2022-04-06 at 6 44 02 PM

gabrielluong: If i am understanding this correctly, I think you just want a Spacer view in after the ProgressBar, but please double check with UX if this 8dp spacing from the top app bar is intended.

app/src/main/res/layout/fragment_synced_history.xml
```diff
@@ -0,0 +1,12 @@
...
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+
+<FrameLayout
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:id="@+id/historyLayout"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    tools:context="org.mozilla.fenix.library.syncedhistory.SyncedHistoryFragment" />
```
gabrielluong, line 12: Add a new line at EOF

app/src/main/res/values/strings.xml
```diff
@@ -764,6 +764,10 @@
     <string name="history_older">Older</string>
     <!-- Text shown when no history exists -->
     <string name="history_empty_message">No history here</string>
+    <!-- Option in library for opening Synced History screen. History that is coming from other devices -->
+    <string name="library_synced_history">Synced from other devices</string>
```
gabrielluong, line 768:
```suggestion
    <string name="history_synced_from_other_devices">Synced from other devices</string>
```

app/src/main/res/values/strings.xml
```diff
@@ -764,6 +764,10 @@
...
+    <!--  The page title for browsing history coming from other devices. -->
+    <string name="synced_history">From other devices</string>
```
gabrielluong, line 770:
```suggestion
    <string name="history_from_other_devices">From other devices</string>
```

app/src/main/res/layout/history_list_item.xml
```diff
@@ -11,7 +11,16 @@
 
     <include
         android:id="@+id/recently_closed_nav_empty"
-        layout="@layout/recently_closed_nav_item" />
+        layout="@layout/recently_closed_nav_item"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:visibility="visible"
+        android:layout_marginTop="8dp"
+        android:layout_marginBottom="0dp" />
```
gabrielluong, line 19: I think you only need `marginTop` here, but I am unsure if you need to make any changes here if we just added that 8dp Spacer and that's only if UX confirms that was intentional.

app/src/main/res/layout/component_history.xml
gabrielluong: If this proves too controversial, I would move this 8dp spacer into a followup because it's affecting a lot of your XML and that is gonna also affect the tab history view, which is when you hold on to the back arrow.

gabrielluong (changes requested):

app/src/main/java/org/mozilla/fenix/library/history/HistoryController.kt
mavduevskiy: fixed

app/src/main/java/org/mozilla/fenix/FeatureFlags.kt
mavduevskiy: fixed

app/src/main/java/org/mozilla/fenix/components/history/PagedHistoryProvider.kt
mavduevskiy: fixed

app/src/main/java/org/mozilla/fenix/library/history/HistoryAdapter.kt
gabrielluong: This change is unnecessary because we don't need to pass this `syncedHistoryVisible` around. We can just call the `FeatureFlags` directly where it is needed.

gabrielluong: We also suppress plenty of `@Suppress("LongParameterList")` where it makes sense to. `BindingParams` added unnecessary complexity and refactoring when it was not necessary to do so because we were passing `syncedHistoryVisible` around.

app/src/main/java/org/mozilla/fenix/library/history/HistoryAdapter.kt
mavduevskiy: I would actually prefer passing a parameter rather than accessing a static class within a UI class.
Giving the UI class knowledge about `FeatureFlags` is not the clearest solution.
Also passing parameters, that affect the behavior of a class, inside the constructor makes it easier to understand. We can see the params in charge, where they are used, and can track where they come from.

app/src/main/java/org/mozilla/fenix/library/history/HistoryController.kt
mavduevskiy: It won't work as a link that way, will it?

app/src/main/java/org/mozilla/fenix/library/history/HistoryAdapter.kt
gabrielluong: We don't expect this parameter to be around when the `FeatureFlag` is enabled. When this is enabled, it would be easier to remove your single direct check of the `FeatureFlag` instead of the refactoring out `syncedHistoryVisible`.

app/src/main/java/org/mozilla/fenix/library/history/HistoryView.kt
mavduevskiy: fixed

app/src/main/java/org/mozilla/fenix/library/history/HistoryController.kt
gabrielluong: You can import the full path.

app/src/main/java/org/mozilla/fenix/library/history/HistoryAdapter.kt
gabrielluong: I think simple is best here. I can't pass a solution where we end up deciding to not document our public class/functions after promoting the need for more documentations. The refactoring is also a bit drastic because of the fact that we can call it the FeatureFlag statically and avoid a lot of refactoring altogether. Alternatively, we could've added it as a an extra parameter and suppress LongParameterList. My other concern is that this refactoring also affects the git blame/history of code where it is probably more valuable to preserve.

app/src/main/java/org/mozilla/fenix/library/history/HistoryView.kt
mavduevskiy: It works a bit differently, see the way `recentlyClosedNav.isVisible = !userHasHistory` works just right above ?
There are the elements that are part of the view itself and are displayed when the view is in empty state.
The elements that are inside the list are actually part of a viewholder.

app/src/main/res/layout/component_history.xml
topotropic: Good catch! this depends a little on what we do on bookmarks and downloads; it would be great to add 8dp spacing from top bar to first item but then also have it consistent with the other "Library" screens; if we don't have the spacing there we can ignore it for now and I file a follow up, thanks!

app/src/main/java/org/mozilla/fenix/library/history/viewholders/HistoryListItemViewHolder.kt
mavduevskiy: I followed the lint rule
```
NestedBlockDepth:
    active: true
    threshold: 4
```

app/src/main/java/org/mozilla/fenix/library/history/viewholders/HistoryListItemViewHolder.kt
gabrielluong: See https://github.com/mozilla-mobile/fenix/pull/24374#discussion_r847555086. If you decide to continue this route, I would ask you to drop the `BindingParams` data class in favor of suppressing long parameter list. I would also rename `syncedHistoryVisible` to `showSyncedHistory` to be consistent with the other parameter naming. Also, we expect public class/functions to be documented.

app/src/main/res/drawable/ic_synced_history.xml
mavduevskiy: agreed, changed the existing to `ic_synced`

app/src/main/res/layout/component_history.xml
mavduevskiy: It's not part of the new design, it's to match the existing one: the populated list has extra spacing, but the empty state hasn't. it results into `recentlyClosedTabs` and  `synced_history_nav_item` items jumping when you change between empty and not empty state

app/src/main/res/layout/fragment_synced_history.xml
mavduevskiy: fixed

app/src/main/res/values/strings.xml
mavduevskiy: Followed the pattern from the element next to it – Recently closed tab:
```
<!-- Option in library for Recently Closed Tabs -->
    <string name="library_recently_closed_tabs">Recently closed tabs</string>
```

gabrielluong: Can you move it besides `library_recently_closed_tabs` then? We want this to be sorted by the prefix ideally or section ideally.

app/src/main/res/layout/history_list_item.xml
mavduevskiy: same [case](https://github.com/mozilla-mobile/fenix/pull/24374#discussion_r847610255)

app/src/main/res/values/strings.xml
mavduevskiy: It sits inside the history section already.

I don't know about the concept and it is a bit confusing to me. Will keep the place, but change the name to the suggested one

app/src/main/res/layout/component_history.xml
mavduevskiy: > Good catch! this depends a little on what we do on bookmarks and downloads; it would be great to add 8dp spacing from top bar to first item but then also have it consistent with the other "Library" screens; if we don't have the spacing there we can ignore it for now and I file a follow up, thanks!

We actually have this extra spacing in History screen when there are elements inside the list, but don't have it when the list is empty. I added the spacing to match the populated state of the screen.

app/src/main/java/org/mozilla/fenix/library/history/HistoryAdapter.kt
```diff
@@ -60,7 +60,13 @@ class HistoryAdapter(
...
         }
 
-        holder.bind(current, timeGroup, position == 0, mode, isPendingDeletion)
+        holder.bind(
+            current,
+            timeGroup,
+            position == 0,
+            mode,
+            isPendingDeletion,
+        )
```
gabrielluong, line 69: Let's undo this

app/src/main/res/layout/component_history.xml
```diff
@@ -25,8 +26,17 @@
         layout="@layout/recently_closed_nav_item"
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
+        android:layout_marginTop="8dp"
```
gabrielluong, line 29: Let's move all of these margin changes into a separate issue. I don't want to mix this PR with a UX change and prefer to deal with it in a separate PR and have a proper UX review.

app/src/main/res/layout/component_tabstray2.xml
```diff
@@ -105,7 +105,7 @@
             android:layout_height="match_parent"
             android:contentDescription="@string/tabs_header_synced_tabs_title"
             android:foregroundTint="@color/photonWhite"
-            android:icon="@drawable/ic_synced_tabs" />
+            android:icon="@drawable/ic_synced" />
```
gabrielluong, line 108: Please undo renaming `ic_synced_tabs`. Our icon naming should follow the naming defined by UX Icon design system. This way we can easily link back the figma icon to our icon naming.

app/src/main/java/org/mozilla/fenix/library/history/viewholders/HistoryListItemViewHolder.kt
```diff
@@ -72,7 +78,10 @@ class HistoryListItemViewHolder(
             }
         }
 
-        toggleTopContent(showTopContent, mode === HistoryFragmentState.Mode.Normal)
+        toggleTopContent(
+            showTopContent,
+            mode === HistoryFragmentState.Mode.Normal
+        )
```
gabrielluong, line 84: Undo this change

app/src/main/java/org/mozilla/fenix/library/history/viewholders/HistoryListItemViewHolder.kt
```diff
@@ -43,12 +48,13 @@ class HistoryListItemViewHolder(
...
     }
 
+    @Suppress("UndocumentedPublicFunction")
     fun bind(
         item: History,
         timeGroup: HistoryItemTimeGroup?,
         showTopContent: Boolean,
         mode: HistoryFragmentState.Mode,
-        isPendingDeletion: Boolean = false,
+        isPendingDeletion: Boolean,
```
gabrielluong, line 57: Is this necessary?

app/src/main/java/org/mozilla/fenix/library/history/viewholders/HistoryListItemViewHolder.kt
```diff
@@ -43,12 +48,13 @@ class HistoryListItemViewHolder(
         }
     }
 
+    @Suppress("UndocumentedPublicFunction")
```
gabrielluong, line 51: Probably can undo this @Supress since nothing should've changed here?

app/src/main/java/org/mozilla/fenix/library/history/HistoryView.kt
```diff
@@ -33,7 +34,9 @@ class HistoryView(
     var mode: HistoryFragmentState.Mode = HistoryFragmentState.Mode.Normal
         private set
 
-    val historyAdapter = HistoryAdapter(interactor).apply {
+    val historyAdapter = HistoryAdapter(
+        interactor
+    ).apply {
```
gabrielluong, line 39: Undo this change

gabrielluong (changes requested):

app/src/main/java/org/mozilla/fenix/library/history/HistoryAdapter.kt
mavduevskiy: fixed

app/src/main/res/layout/component_history.xml
mavduevskiy: I think it's better to keep it. We are adding a new button to the screen, the inconsistency is obvious, the change is pretty small and there will be an UX review anyway

app/src/main/res/layout/component_tabstray2.xml
mavduevskiy: how can I find the responding icon? 

the file name [used](https://www.figma.com/file/am9EJGg6ow29PXWkc9jcOH/MR2-(2021)?node-id=1311%3A244184) in the synced screen is `Synchronize tabs`.

app/src/main/java/org/mozilla/fenix/library/history/viewholders/HistoryListItemViewHolder.kt
mavduevskiy: the default value isn't necessary

app/src/main/java/org/mozilla/fenix/library/history/HistoryView.kt
mavduevskiy: undone

app/src/main/java/org/mozilla/fenix/library/history/viewholders/HistoryListItemViewHolder.kt
mavduevskiy: undone

app/src/main/java/org/mozilla/fenix/library/history/HistoryAdapter.kt
mavduevskiy: undone

app/src/main/res/layout/component_history.xml
gabrielluong: The reason I am asking this to move separately is because I don't think the implementation is correct. I think you just want to add a View that has a height of 8dp instead of selectively adding the margin into each of these components. If both `recently_closed_nav_item` and `synced_history_nav_item` are both visible, then you now have a 8dp spacing in between them, which you can see in the Figma should not be the case.

Screen Shot 2022-04-06 at 6 44 02 PM

I also don't see this 8dp spacing in the other "Library" UI (Bookmark or Download) so I think we can also ignore it based on Nicole's comment https://github.com/mozilla-mobile/fenix/pull/24374/files#r847585856.

gabrielluong (changes requested):

app/src/main/java/org/mozilla/fenix/library/history/viewholders/HistoryListItemViewHolder.kt
gabrielluong, line 51: Can you add a comment for this? We shouldn't be suppressing `UndocumentedPublicFunction` since this is easily avoidable.

mavduevskiy: Do you mean removing the unused default value?

gabrielluong: I am assuming removing the unused default value triggered the lint error here? You can either undo it or keep the change but document the public function. We don't currently `@Suppress("UndocumentedPublicFunction")` and I am looking to keep it that way.

mavduevskiy: Yes, it does trigger the lint error. May be if I rebase to the initial commit and revoke that bind method to the original form it might be okay.

A bind method cannot really have a meaningful documentation, can it?

mavduevskiy:
@gabrielluong 
Screen Shot 2022-04-11 at 3 18 02 PM
Screen Shot 2022-04-11 at 3 17 18 PM
On the current main there is extra margin in the not empty state, and there is no margin in empty state. It looks like a glitch, it is an inconsistency and since I have to add a new button below the tab button, I have to adjust the margins.

I will do 8 atop of two buttons and 32 below, according to the design.

gabrielluong:
> On the current main there is extra margin in the not empty state, and there is no margin in empty state. It looks like a glitch, it is an inconsistency and since I have to add a new button below the tab button, I have to adjust the margins.
> 
> I will do 8 atop of two buttons and 32 below, according to the design.

Can you also post a Before/After screenshot? The original problem I have been mentioning in https://github.com/mozilla-mobile/fenix/pull/24374#discussion_r847746083 is the fact that we added an extra 8dp margin in between "Recently closed tabs" and "Synced from other devices".  See the purple margin in between the 2 row items.

Screen Shot 2022-04-11 at 7 25 50 PM

At this point, the only thing holding back this PR is your margin changes, which I don't believe follows the intended design as illustrated above. I also believe aligning with the design is also out of scope for this PR and would prefer if we did this independently so that we can have UX review the Before/After screenshots since these changes would go live immediately for the History view.

mavduevskiy:
Screen Shot 2022-04-11 at 6 01 54 PM
Screen Shot 2022-04-11 at 6 00 59 PM

That's going to be the next fix after Nicole confirms the bottom padding.

mavduevskiy:
Got confirmation from @topotropic that spacing between bottom button and time group title should be 32dp.
Top spacing of 8dp is to keep, there will be a ticket to add padding to download and bookmark screens as well.

mavduevskiy:
@gabrielluong 
Could you find some time to review the changes?

app/src/main/res/layout/component_history.xml
```diff
@@ -14,18 +14,32 @@
...
         android:layout_width="match_parent"
         android:layout_height="8dp"
+        android:layout_marginBottom="8dp"
         android:indeterminate="true"
         android:translationY="-3dp"
         android:visibility="gone"
         app:layout_constraintStart_toStartOf="parent"
         app:layout_constraintTop_toTopOf="parent" />
 
+    <View
```
gabrielluong, line 24: Let's also add `android:visibility="gone"`

app/src/main/res/layout/history_list_item.xml
```diff
@@ -9,10 +9,24 @@
...
         layout="@layout/recently_closed_nav_item" />
 
+    <include
+        android:id="@+id/synced_history_nav_empty"
+        layout="@layout/synced_history_nav_item" />
+
+    <View
+        android:id="@+id/bottom_divider"
+        android:layout_width="match_parent"
+        android:layout_height="32dp" />
```
gabrielluong, line 28: It has always been confusing how we duplicate the content between `history_list_item` and `component_history` and toggle the visibility in different views based on the state when we should just have a clear distinction. Unfortunately, adding this here does have unwanted side effects because `history_list_item` is used in multiple places including Tab History and Recently Closed Tab.

So, this is what I am seeing in Recently closed tabs, which doesn't seem intended and should be addressed.

Screen Shot 2022-04-13 at 4 08 02 PM

gabrielluong (changes requested):
Can you add Before and After screenshots of a History view with items and with no items for a UX review?

app/src/main/java/org/mozilla/fenix/library/history/viewholders/HistoryListItemViewHolder.kt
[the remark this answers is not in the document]
mavduevskiy: I don't see anything wrong with the @Suppress. The default value wasn't ever used, removing it will improve the code base. And the `[example](https://searchfox.org/mozilla-mobile/source/android-components/components/feature/prompts/src/main/java/mozilla/components/feature/prompts/creditcard/CreditCardItemViewHolder.kt#29)` documentation adds zero value in understanding the code. Comments should provide explanation of a higher abstraction layer than the code itself. 
A comment like `Binds the view with the provided [CreditCard].` inside a ViewHolder class on a bind method doesn't add much really.

app/src/main/res/values/strings.xml
```diff
@@ -764,6 +764,10 @@
     <string name="history_older">Older</string>
     <!-- Text shown when no history exists -->
     <string name="history_empty_message">No history here</string>
+    <!-- Text for the button inside History screen that opens Synced History screen. History that is coming from other devices -->
+    <string name="history_synced_from_other_devices">Synced from other devices</string>
+    <!--  The page title for browsing history coming from other devices. -->
```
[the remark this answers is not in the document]
mavduevskiy: Removed the spacing, kept the page title

app/src/main/java/org/mozilla/fenix/library/history/HistoryView.kt
```diff
@@ -131,6 +134,15 @@ class HistoryView(
             )
             recentlyClosedNav.isVisible = !userHasHistory
         }
+
+        with(binding.syncedHistoryNavEmpty) {
+            val syncedHistoryVisible = FeatureFlags.showSyncedHistory && !userHasHistory
```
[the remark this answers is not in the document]
mavduevskiy: fixed

app/src/main/res/layout/component_history.xml
```diff
@@ -14,18 +14,32 @@
         style="@style/Widget.AppCompat.ProgressBar.Horizontal"
         android:layout_width="match_parent"
         android:layout_height="8dp"
+        android:layout_marginBottom="8dp"
```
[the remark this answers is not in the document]
mavduevskiy: you are right, we don't. removed

app/src/main/res/layout/component_history.xml
```diff
@@ -14,18 +14,32 @@
...
         android:indeterminate="true"
         android:translationY="-3dp"
         android:visibility="gone"
         app:layout_constraintStart_toStartOf="parent"
         app:layout_constraintTop_toTopOf="parent" />
 
+    <View
+        android:id="@+id/top_divider"
```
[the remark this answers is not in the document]
mavduevskiy: Fair enough. I was following the naming from [here](https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/res/layout/add_new_collection_dialog.xml#13).

app/src/main/res/layout/history_list_item.xml
```diff
@@ -9,10 +9,24 @@
     android:importantForAccessibility="no"
     android:orientation="vertical">
 
+    <View
```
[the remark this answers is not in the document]
mavduevskiy: fixed

app/src/main/res/layout/history_list_item.xml
```diff
@@ -9,10 +9,24 @@
...
+        android:id="@+id/top_divider"
```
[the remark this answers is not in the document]
mavduevskiy: fixed

app/src/main/res/layout/component_history.xml
mavduevskiy: added

app/src/main/res/layout/history_list_item.xml
mavduevskiy: Nice catch! Fixed.

Yes, it was also confusing to me. But it was already there, so I had to work with it.

app/src/main/java/org/mozilla/fenix/library/history/viewholders/HistoryListItemViewHolder.kt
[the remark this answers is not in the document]
gabrielluong: Sorry, we aren't moving forward with this until we address this. I will even allow you to add this to the detekt-baseline, but since there's no instances of `@Suppress("UndocumentedPublicFunction")` anywhere in our codebase https://searchfox.org/mozilla-mobile/search?q=UndocumentedPublicFunction&path=, I see no reason why we should add this exception here.

gabrielluong (changes requested):

gabrielluong:
> Can you add Before and After screenshots of a History view with items and with no items for a UX review? Here's an example #24046

Please also do this before requesting a review.

mavduevskiy:
Synced history button
Old|New
---|---
photo_2022-04-19 20 50 06|photo_2022-04-19 20 50 02
photo_2022-04-19 20 53 07|photo_2022-04-19 20 53 04
With synced history feature on
photo_2022-04-19 20 56 44|photo_2022-04-19 20 56 39

gabrielluong:
Can you ping @topotropic for a UX review since this new design is gonna be shipping? You might need to update the screenshots without the layout inspection depending on her preference.

mavduevskiy:
@gabrielluong 
I discussed that with Nicole [already](https://github.com/mozilla-mobile/fenix/pull/24374#issuecomment-1097083238), providing her with the screenshots, and got her approval.

app/src/main/java/org/mozilla/fenix/library/history/viewholders/HistoryListItemViewHolder.kt
```diff
@@ -43,12 +48,19 @@ class HistoryListItemViewHolder(
         }
     }
 
+    /**
+     * Displays the data of the given history record.
+     * @param timeGroup used to form headers for different time frames, like today, yesterday, etc.
+     * @param showTopContent enables the Recent tab button.
+     * @param mode switches between editing and regular modes.
+     * @param isPendingDeletion hides the item unless an undo snackbar action is evoked.
+     */
```
gabrielluong, line 57:
```suggestion
    /**
     * Displays the data of the given history record.
     *
     * @param item The [History] item to display.
     * @param timeGroup The [HistoryItemTimeGroup] of the history item to display.
     * @param showTopContent Whether or not to display the "Recently closed tabs" and "Synced from other devices" buttons.
     * @param mode The state for the History Screen.
     * @param isPendingDeletion Whether or not to hide the history item unless an undo snackbar action is evoked.
     */
```

gabrielluong (approved):
Can you make sure https://github.com/mozilla-mobile/fenix/issues/24741 is on track as an immediate follow up so we don't forget about it and ensure we have design consistency between these screens for this upcoming release?

Please also squash your commits since some of them references the wrong issue number.

topotropic (approved):
LGTM, thanks!

PR #24386, opened 2022-03-22

MickeyMoz:


AaronMT:
@rvandermeulen Where can I find the changelog for 20220322143329?

rvandermeulen:
@AaronMT It's a bit involved, unfortunately. You'd want to work your way backwards starting with:
https://nightly.maven.mozilla.org/?prefix=maven2/org/mozilla/components/browser-engine-gecko/

You can use the .pom files to see the revisions the releases were built from and create a changelog range from that. Note that you'll probably end up needing to repeat the same process for the GeckoView revisions that went into the different AC builds.

This would be a great thing for someone to create a tool for, frankly...

PR #24396, opened 2022-03-23
for #24177: sync tabs when home is shown

MatthewTighe:
Still have a few issues to work out, but I thought I'd open a draft PR in case there were any sweeping architectural changes requests. KDocs and tests to come once that stuff is settled.

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/home/recenttabs/RecentTabsListFeature.kt
```diff
@@ -54,13 +54,13 @@ sealed class RecentTab {
      * @param deviceDisplayName The device the tab was viewed on.
      * @param title The title of the tab.
      * @param url The url of the tab.
-     * @param previewImageUrl The url used to retrieve the preview image of the tab.
+     * @param iconUrl The url used to retrieve the icon of the tab.
```
gabrielluong, line 57: Just wanted to make sure that this change is intentional. `previewImageUrl` is a very distinct naming that follows the GV naming and is a preview (hero) image of the page and not an icon (hero image != icon). The preview (hero) image is parsed from the page's metadata (these are `<meta property="og:image">` tags).

app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabViewHolder.kt
```diff
@@ -40,11 +41,15 @@ class RecentTabViewHolder(
     @Composable
     override fun Content() {
         val recentTabs = components.appStore.observeAsComposableState { state -> state.recentTabs }
+        val recentSyncedTabState = components.appStore.observeAsComposableState { state -> state.recentSyncedTabState }
 
         RecentTabs(
             recentTabs = recentTabs.value ?: emptyList(),
+            recentSyncedTabState = recentSyncedTabState.value ?: RecentSyncedTabState.None,
```
gabrielluong, line 48: I am starting to think we should just rip out the Recent Synced Tabs into its own component separate from `RecentTabs` and do the following:

```
Column {
  RecentTabs()
  Spacer()
  RecentSyncedTab()
}
```
This has the benefit of separating out the state and callback parameters of 2 very distinct components. I don't think Layout would be impeded either.

app/src/main/java/org/mozilla/fenix/home/recenttabs/RecentTabsListFeature.kt
MatthewTighe: Thanks for double-checking.  think I was incorrect in the initial implementation by naming it `previewImageUrl`. After some testing, it was looking like it pointed at a favicon and not a hero image. The `iconUrl` naming convention matches the underlying data that is stored as a `TabEntry` in `RemoteTabsStorage` in A-C.

gabrielluong: 👍

app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabViewHolder.kt
MatthewTighe: I separated them into two separate composables. Were you suggesting to also move the synced tab stuff out of `RecentTabController` and into its own handler?

gabrielluong: I will leave that up to you. I am on the fence since there's a `RecentTab.SyncedTab` so maybe I'll draw your attention towards whether or not `SyncedTab` should belong in the sealed class of `RecentTab`. Before I think it made sense assuming we were appending to the `List<RecentTab>` and therefore determining whether or not we would render the component based on it, but if this is gonna be a totally unrelated state, maybe we should also move it out.

app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabs.kt
```diff
@@ -82,13 +83,49 @@ import org.mozilla.fenix.theme.FirefoxTheme
  */
 @Composable
 @Suppress("LongParameterList")
-fun RecentTabs(
+fun JumpBackIn(
```
gabrielluong, line 86: Hm.. my original intention of having the 2 separate components was to dissolve having a @Composable that needed to take the combination of the `RecentSyncedTabState` and `List<RecentTab>` states and their respective callbacks since we can just pass these parameters to the components directly in `Content()` instead of having a wrapper component.

MatthewTighe: Ah, I see. I'll update that. I think it makes more sense to remove the `SyncedTab` from the `RecentTab` sealed class as well, so I'll do some further refactoring.

app/src/main/java/org/mozilla/fenix/home/recenttabs/RecentSyncedTabHomeFeature.kt
```diff
@@ -0,0 +1,124 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.home.recenttabs
```
gabrielluong, line 5: Maybe move all of this into its own /home/recentsyncedtab/ package since we're breaking it out

MatthewTighe: I didn't at first because I wasn't sure where to put shared components between the two packages. I think it makes sense to continue sharing some of the composables in `RecentTabs`, since it seems likely that the design for the two cards will stay largely in sync. I can move the recent synced tab stuff into its own package and just import the recent tabs composables for now, but IMO that shows they are candidates for a more generic UI package in the future.

gabrielluong: Can you make a list of the reusable composable components? The one obvious one I saw in my brief scan is `RecentTabTitle`, which I think we should probably refactor and use our existing reusable `PrimaryText()` component and define a typography. This can be a filed as a followup along with the list of things you see that we can refactor. Alternatively, I would refactor them in a separate issue and land it before this PR.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -163,6 +164,19 @@ class HomeFragment : Fragment() {
...
     }
 
+    private val syncedTabFeature by lazy {
+        RecentSyncedTabHomeFeature(
+            store = requireComponents.appStore,
+            context = requireContext(),
+            addNewTabUseCase = requireComponents.useCases.tabsUseCases.addTab,
+            navController = findNavController(),
+            lifecycleScope = lifecycleScope,
+            storage = requireComponents.backgroundServices.syncedTabsStorage,
```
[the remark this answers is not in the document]
MatthewTighe: I'll update it, but I'm curious as to why we'd like to. `Components.syncedTabsStorage` is `lazyMonitored` in case that hook is ever used again, and immediately initializes the underlying `Lazy<RemoteTabsStorage>`. The `syncedTabFeature` is also lazily instantiated in the `HomeFragment`.

app/src/main/java/org/mozilla/fenix/home/recenttabs/RecentSyncedTabHomeFeature.kt
```diff
@@ -0,0 +1,124 @@
...
+            storage = storage,
+            accountManager = accountManager,
+            lifecycleOwner = lifecycleOwner,
+            onTabClicked = {}
+        )
+    }
+
+    init {
+        lifecycleScope.launchWhenStarted {
+            accountManager.syncNow(SyncReason.User, false, listOf(SyncEngine.Tabs))
```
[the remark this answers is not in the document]
MatthewTighe: Oh good catch, I'll remove that.

app/src/main/java/org/mozilla/fenix/home/recenttabs/interactor/RecentSyncedTabInteractor.kt
```diff
@@ -0,0 +1,13 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.home.recenttabs.interactor
+
+import org.mozilla.fenix.home.recenttabs.RecentSyncedTab
+
+interface RecentSyncedTabInteractor {
```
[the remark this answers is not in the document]
MatthewTighe: Yeah will do! Been holding off on tests and kdocs until the shape of the code is in place.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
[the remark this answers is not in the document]
MatthewTighe: Additionally, it doesn't look like there's a `Lazy<SyncedTabsStorage>`. Would it be best to add one?

MatthewTighe:
@gabrielluong I think this should be ready for another round of review

gabrielluong:
> @gabrielluong I think this should be ready for another round of review

Ok, I might not get to this today. I also get a ping on Slack from the GitHub app if you push the re-request review button.

app/src/main/java/org/mozilla/fenix/home/recentsyncedtabs/view/RecentSyncedTab.kt
```diff
@@ -0,0 +1,227 @@
...
+                    modifier = Modifier.fillMaxHeight()
+                ) {
+                    if (tab == null) {
+                        RecentTabTitlePlaceholder()
+                    } else {
+                        PrimaryText(
+                            text = tab.title,
+                            fontSize = 14.sp,
+                            maxLines = 2,
+                            overflow = TextOverflow.Ellipsis,
```
gabrielluong, line 121: Can we flip this so that we follow the order of the parameters in `PrimaryText()`?

app/src/main/java/org/mozilla/fenix/home/recentsyncedtabs/view/RecentSyncedTab.kt
```diff
@@ -0,0 +1,227 @@
...
+                                    .background(FirefoxTheme.colors.layer3)
+                                    .size(18.dp)
+                            )
+                        } else {
+                            Image(
+                                painter = painterResource(R.drawable.ic_synced_tabs),
+                                contentDescription = stringResource(
+                                    R.string.recent_tabs_synced_device_icon_content_description
+                                ),
+                                modifier = Modifier.size(18.dp, 18.dp)
```
gabrielluong, line 138: IIRC I think we can simplify this to `size(18.dp)` since the width and height are the same.

app/src/test/java/org/mozilla/fenix/home/SessionControlInteractorTest.kt
```diff
@@ -171,6 +175,21 @@ class SessionControlInteractorTest {
         verify { recentTabController.handleRecentTabShowAllClicked() }
     }
 
+    @Test
+    fun `WHEN recent synced tab clicked THEN the tab is handled`() {
```
gabrielluong, line 179:
```suggestion
    fun `WHEN recent synced tab is clicked THEN the tab is handled`() {
```

app/src/test/java/org/mozilla/fenix/home/SessionControlInteractorTest.kt
```diff
@@ -171,6 +175,21 @@ class SessionControlInteractorTest {
...
+        val tab: RecentSyncedTab = mockk()
+        interactor.onRecentSyncedTabClicked(tab)
+
+        verify { recentSyncedTabController.handleRecentSyncedTabClick(tab) }
+    }
+
+    @Test
+    fun `WHEN recent synced tabs show all clicked THEN show all synced tabs handled`() {
```
gabrielluong, line 187:
```suggestion
    fun `WHEN recent synced tabs show all is clicked THEN show all synced tabs is handled`() {
```

app/src/test/java/org/mozilla/fenix/home/recentsyncedtabs/RecentSyncedTabHomeFeatureTest.kt
```diff
@@ -0,0 +1,228 @@
...
+        id = "currentId",
+        displayName = "currentDevice",
+        deviceType = DeviceType.MOBILE,
+        isCurrentDevice = true,
+        lastAccessTime = timeNow,
+        capabilities = listOf(),
+        subscriptionExpired = false,
+        subscription = null
+    )
+    private val deviceAccessedFirst = Device(
```
gabrielluong, line 40: Maybe go with `deviceAccessed1`, `deviceAccessed2`. Up to you

gabrielluong (changes requested):
Looks pretty good so far.

app/src/main/java/org/mozilla/fenix/components/appstate/AppStoreReducer.kt
```diff
@@ -39,6 +39,7 @@ internal object AppStoreReducer {
             topSites = action.topSites,
             recentBookmarks = action.recentBookmarks,
             recentTabs = action.recentTabs,
+            recentSyncedTabState = action.recentSyncedTabState,
```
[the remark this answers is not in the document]
MatthewTighe: I'd been thinking that `AppAction.Change` was intended to expose a way to manipulate the entire `AppState` in a single action, for setting up initial states in tests or otherwise. If that wasn't the intended use-case, then I can remove `recentSyncedTabState` from the action entirely.

app/src/main/java/org/mozilla/fenix/home/recentsyncedtabs/RecentSyncedTabHomeFeature.kt
```diff
@@ -0,0 +1,140 @@
...
+import org.mozilla.fenix.tabstray.Page
+
+/**
+ * HomeFragment delegate to handle layout updates and dispatch actions related to the recent
+ * synced tab.
+ *
+ * @param store Store to dispatch actions to when synced tabs are updated or errors encountered.
+ * @param addNewTabUseCase Use case to open the synced tab when clicked.
+ * @param navController Nav controller to navigate to synced tabs tray.
+ * @param accountManager Account manager used to retrieve synced tab state.
```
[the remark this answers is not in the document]
MatthewTighe: This is a super pedantic stance of mine, but I'd argue from the context of a client of the class, which is calling the constructor and arguably who the kdoc is for, has no knowledge of whether the constructor parameter is a property of the class or not since it's private. 

I'll switch it here, just wanted to drop my two cents for fun

app/src/main/java/org/mozilla/fenix/components/appstate/AppStoreReducer.kt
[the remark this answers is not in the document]
gabrielluong: We used to call `AppAction.Change` in `onStart` to populate the `HomeFragmentStore` with the cached state from various settings and storage. That was one of the reason it took so many params. We don't need to do this anymore since `AppStore` is global. Otherwise, it's current functionality is to change multiple parameters all at once easily in the BlocklistMiddle, but we will want to rethink what parameter needs to be exposed.

app/src/main/java/org/mozilla/fenix/home/recentsyncedtabs/RecentSyncedTabHomeFeature.kt
```diff
@@ -0,0 +1,140 @@
...
+                    iconUrl = tab.iconUrl
+                )
+            } ?: return
+        store.dispatch(
+            AppAction.RecentSyncedTabStateChange(RecentSyncedTabState.Success(syncedTab))
+        )
+    }
+
+    override fun onError(error: SyncedTabsView.ErrorType) {
+        store.dispatch(AppAction.RecentSyncedTabStateChange(RecentSyncedTabState.None))
```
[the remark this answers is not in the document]
MatthewTighe: Good catch, ty

app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabViewHolder.kt
```diff
@@ -40,17 +53,30 @@ class RecentTabViewHolder(
...
+                menuItems = listOf(
+                    RecentTabMenuItem(
+                        title = stringResource(id = R.string.recent_tab_menu_item_remove),
+                        onClick = { tab -> recentTabInteractor.onRemoveRecentTab(tab) }
+                    )
                 )
             )
-        )
+
+            if (FeatureFlags.taskContinuityFeature) {
```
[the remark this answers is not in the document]
MatthewTighe: Ah good call. I had initially thought to try and leave as much logic as possible out of the view holder, but I agree that could lead to to the above.

app/src/main/java/org/mozilla/fenix/home/recentsyncedtabs/RecentSyncedTabHomeFeature.kt
```diff
@@ -0,0 +1,140 @@
...
+
+/**
+ * HomeFragment delegate to handle layout updates and dispatch actions related to the recent
+ * synced tab.
+ *
+ * @param store Store to dispatch actions to when synced tabs are updated or errors encountered.
+ * @param addNewTabUseCase Use case to open the synced tab when clicked.
+ * @param navController Nav controller to navigate to synced tabs tray.
+ * @param accountManager Account manager used to retrieve synced tab state.
+ * @param context Fragment context.
```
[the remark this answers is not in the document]
MatthewTighe: Dug in a bit and wrote a better description. Lemme know if it looks better

app/src/main/java/org/mozilla/fenix/components/appstate/AppStoreReducer.kt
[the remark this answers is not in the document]
MatthewTighe: Makes sense! I don't think it will need to be included for now, then

MatthewTighe:
@gabrielluong thanks as always for the thorough review. Think the only item left from this round is the performance benefits of using `Lazy` wrappers around some properties. It might be best to just go for it for now, since Jonathan is out. I think we want this in nightly as long as possible.

MatthewTighe:
Looking more closely at the concerns about `Lazy` constructor values, I think we can ignore them for now.

- `accountManager` has become a param instead of a property. 
- `accountManager` and the `SyncedTabsStorage` are both only used to construct the underlying `SyncedTabsFeature`. The feature is instantiated lazily, so should not be constructed until a call to `start` triggered by lifecycle components.
- both components are also wrapped in a lazy delegate.

AFAIK, we should have lazy instantiation for every potentially expensive component. IMO, we should move forward with a merge and nightly release to give this as much exposure time as possible and revisit the `Lazy` stuff if there's a need down the line.

app/src/main/java/org/mozilla/fenix/home/recentsyncedtabs/view/RecentSyncedTab.kt
```diff
@@ -0,0 +1,207 @@
...
+ * A recent synced tab card.
+ *
+ * @param recentSyncedTabState The state of the recent synced tab.
+ * @param onRecentSyncedTabClick Invoked when the user clicks on the recent synced tab.
+ * @param onSeeAllSyncedTabsButtonClick Invoked when user clicks on the "See all" button in the synced tab card.
+ */
+@Suppress("LongMethod")
+@Composable
+fun RecentSyncedTab(
+    recentSyncedTabState: RecentSyncedTabState,
```
gabrielluong, line 55: I would go one step further and just make this `RecentSyncedTab?` since we already checked the state prior to this point and avoid processing the state at this point.

app/src/main/java/org/mozilla/fenix/home/recentsyncedtabs/view/RecentSyncedTab.kt
```diff
@@ -0,0 +1,207 @@
...
+/**
+ * A recent synced tab card.
+ *
+ * @param recentSyncedTabState The state of the recent synced tab.
+ * @param onRecentSyncedTabClick Invoked when the user clicks on the recent synced tab.
+ * @param onSeeAllSyncedTabsButtonClick Invoked when user clicks on the "See all" button in the synced tab card.
+ */
+@Suppress("LongMethod")
+@Composable
+fun RecentSyncedTab(
```
gabrielluong, line 54: Can we also add some @Preview for this component

app/src/test/java/org/mozilla/fenix/components/AppStoreTest.kt
```diff
@@ -231,6 +252,7 @@ class AppStoreTest {
             assertEquals(0, appStore.state.recentBookmarks.size)
             assertEquals(0, appStore.state.recentHistory.size)
             assertEquals(Mode.Normal, appStore.state.mode)
+            assertEquals(RecentSyncedTabState.None, appStore.state.recentSyncedTabState)
```
gabrielluong, line 255: Probably don't need this.

app/src/main/java/org/mozilla/fenix/home/recentsyncedtabs/view/RecentSyncedTab.kt
```diff
@@ -0,0 +1,207 @@
...
+                ),
+                modifier = Modifier
+                    .height(36.dp)
+                    .fillMaxWidth()
+            ) {
+                if (tab != null) {
+                    Text(
+                        text = stringResource(R.string.recent_tabs_see_all_synced_tabs_button_text),
+                        textAlign = TextAlign.Center,
+                        color = FirefoxTheme.colors.textActionSecondary
```
gabrielluong, line 161: I have a fixed for the color problems you were seeing in https://github.com/mozilla-mobile/fenix/issues/24653.

gabrielluong (approved):

app/src/main/java/org/mozilla/fenix/home/recentsyncedtabs/view/RecentSyncedTab.kt
MatthewTighe: TY!

PR #24398, opened 2022-03-23
For #23665 - Make the undo snackbar for the library screens work globally

### LINKED ISSUE #23665: [Bug] Undo snackbar not working after leaving the screen it originated from

mavduevskiy:
Closes #23665

Change `.allowUndo` to make `operation` pass context, so the callback would work even is the fragment called `onDestroyView` and it's `context` is null. Change `view` that's passed to `.allowUndo`, so the Undo would be displayed even after leaving the fragment which it was called from.


mavduevskiy:
Made snackbars survive navigation. It used to be a bit inconsistent, some would survive, others wouldn't. Did that by passing the `activity.view`. Passing `Fragment.getView()` actually creates a memory leak.

Made snackbars work after leaving the screen. The problem was having no access to Fragment context inside the `delete` lambda if the user has left the screen. Context was needed to access `components` for metrics or stores. 

Things like `pendingHistoryDeletionJob` worked as a workaround, executing the `delete` block immediately after navigation. Which also made Undo action do nothing, as the removal had already happened.

Removed all `pendingJobs` and passed context into `delete` lambda, so it could work after navigation. Effectively, it is an activity context.

app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkFragment.kt
```diff
@@ -384,12 +366,11 @@ class BookmarkFragment : LibraryPageFragment<BookmarkNode>(), UserInteractionHan
...
                 setPositiveButton(R.string.delete_browsing_data_prompt_allow) { dialog: DialogInterface, _ ->
                     updatePendingBookmarksToDelete(selected)
-                    pendingBookmarkDeletionJob = getDeleteOperation(BookmarkRemoveType.FOLDER)
                     dialog.dismiss()
                     val snackbarMessage = getRemoveBookmarksSnackBarMessage(selected, containsFolders = true)
                     // Use fragment's lifecycle; the view may be gone by the time dialog is interacted with.
-                    lifecycleScope.allowUndo(
-                        requireView(),
+                    MainScope().allowUndo(
+                        requireActivity().getRootView()!!,
```
Amejia481, line 373: Could we get a null pointer exception when calling `getRootView`?

app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkFragment.kt
```diff
@@ -330,8 +311,9 @@ class BookmarkFragment : LibraryPageFragment<BookmarkNode>(), UserInteractionHan
             }
         }
 
-        viewLifecycleOwner.lifecycleScope.allowUndo(
-            requireView(), message,
+        MainScope().allowUndo(
+            requireActivity().getRootView()!!,
```
Amejia481, line 315: Same as https://github.com/mozilla-mobile/fenix/pull/24398/files#r852355860

app/src/main/java/org/mozilla/fenix/library/downloads/DownloadFragment.kt
```diff
@@ -128,15 +125,14 @@ class DownloadFragment : LibraryPageFragment<DownloadItem>(), UserInteractionHan
      */
     private fun deleteDownloadItems(items: Set<DownloadItem>) {
         updatePendingDownloadToDelete(items)
-        undoScope = CoroutineScope(IO)
-        undoScope?.allowUndo(
-            requireView(),
+        MainScope().allowUndo(
+            requireActivity().getRootView()!!,
```
Amejia481, line 129: Same as https://github.com/mozilla-mobile/fenix/pull/24398/files#r852355860

app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkFragment.kt
```diff
@@ -44,13 +46,14 @@ import org.mozilla.fenix.R
...
 import org.mozilla.fenix.ext.components
-import org.mozilla.fenix.ext.minus
-import org.mozilla.fenix.ext.nav
 import org.mozilla.fenix.ext.requireComponents
 import org.mozilla.fenix.ext.setTextColor
+import org.mozilla.fenix.ext.nav
+import org.mozilla.fenix.ext.bookmarkStorage
+import org.mozilla.fenix.ext.minus
 import org.mozilla.fenix.ext.toShortUrl
+import org.mozilla.fenix.ext.getRootView
```
gabrielluong, line 56: There's something weird about these imports. Running the formatter should order them alphabetically. So, I am wondering if you're manually re-ordering things?

app/src/test/java/org/mozilla/fenix/library/bookmarks/BookmarkControllerTest.kt
```diff
@@ -450,39 +398,19 @@ class BookmarkControllerTest {
...
-    fun `WHEN onSearch is called with BookmarkFragment THEN navigate to BookmarkSearchDialogFragment`() {
-        val controller = createController()
-
-        controller.handleSearch()
-        verify {
-            navController.navigate(
-                BookmarkFragmentDirections.actionBookmarkFragmentToBookmarkSearchDialogFragment()
-            )
-        }
-    }
```
gabrielluong, line 477: Why was this test deleted? I can still see we have this function

app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkFragment.kt
```diff
@@ -330,8 +311,9 @@ class BookmarkFragment : LibraryPageFragment<BookmarkNode>(), UserInteractionHan
             }
         }
 
-        viewLifecycleOwner.lifecycleScope.allowUndo(
-            requireView(), message,
+        MainScope().allowUndo(
```
Amejia481, line 314: Just to confirm, the main source of the issue was that the scope was not lasting enough for use to remove?  For this reason we are switching to `MainScope`?

app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkFragment.kt
Amejia481: Just to confirm, here we are switching to make the parent of the snackback to be the root of the `HomeActivity` that is alive even when the `BookmarkFragment` is gone?

Amejia481:
From manual local testing, the PR address the issue.

app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkFragment.kt
mavduevskiy: Nice catch, thanks :)
The default settings for the formatter is to use a wildcard import if there are more then 5 imports. So I imported manually. Changed the rules.

app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkFragment.kt
mavduevskiy: addressed that [here](https://github.com/mozilla-mobile/fenix/pull/24517#discussion_r852309800)

app/src/main/java/org/mozilla/fenix/library/downloads/DownloadFragment.kt
mavduevskiy: Me and Arturo had a call about it, checking the source code and crashes lead us to the conclusion that it's pretty safe. [Reasoning](https://github.com/mozilla-mobile/fenix/pull/24517#discussion_r852309800)

app/src/test/java/org/mozilla/fenix/library/bookmarks/BookmarkControllerTest.kt
mavduevskiy: Brought back, thanks

mavduevskiy:
@Amejia481 
I have checked if `lifecycleScope.allowUndo()` causes a leak, and yep, it does! You could see it on the video.
May be it makes sense to investigate other places we use `allowUndo()` with the `lifecycleScope`.

https://user-images.githubusercontent.com/92760693/164116795-09d4f1db-cd0e-48c7-958a-8c01e9da6ec6.mp4

Amejia481 (approved):
LGTM!

Amejia481:
> @Amejia481 I have checked if `lifecycleScope.allowUndo()` causes a leak, and yep, it does! You could see it on the video. May be it makes sense to investigate other places we use `allowUndo()` with the `lifecycleScope`.
> Untitled.mp4

Thanks for investigating, let's file a ticket to double check the other usages.

gabrielluong:
Please try to have better PR titles that includes the issue number and a description of what it fixes.

PR #24465, opened 2022-03-25
No issue: Disable HomeScreen tests for recent failing

AaronMT:
As per https://console.firebase.google.com/u/1/project/moz-fenix/testlab/histories/bh.66b7091e15d53d45/matrices/6119952028663171131/details?stepId=bs.40bbe80adaf5999e&testCaseId=1

Some recent change on the home-screen has regressed these tests which are merely checking the home screen for elements..


AaronMT:
@sarah541 This is probably a legitimate bug as the home-screen is not visible or loading to completion on API28. Did any prior work land to any recent refactoring from A-C or elsewhere?

Screen Shot 2022-03-25 at 4 10 23 PM

PR #24466, opened 2022-03-25
For #24391 - Add a customizable child Composable to the list header

### LINKED ISSUE #24391: Add a Composable child to ExpandableListHeader to enable secondary actions in the header
Follow-up from #24337. Find a way to add a child Composable so there can be custom secondary action UI in the header.

MozillaNoah:
Closes #24391 

Basically took the existing header and wrapped it inside a `Row` with the new `actions` Composable

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/compose/list/ExpandableListHeader.kt
+++ b/app/src/main/java/org/mozilla/fenix/compose/list/ExpandableListHeader.kt
@@ -36,7 +36,9 @@ import org.mozilla.fenix.theme.FirefoxTheme
  * @param expandActionContentDescription The content description for expanding the section.
  * @param collapseActionContentDescription The content description for collapsing the section.
  * @param onClick Optional lambda for handling header clicks.
+ * @param actions Optional Composable for adding UI to the end of the header.
  */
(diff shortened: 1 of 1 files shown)


gabrielluong (approved):

app/src/main/java/org/mozilla/fenix/compose/list/ExpandableListHeader.kt
```diff
@@ -36,45 +36,54 @@ import org.mozilla.fenix.theme.FirefoxTheme
...
  */
+@Suppress("LongParameterList")
 @Composable
 fun ExpandableListHeader(
     headerText: String,
     expanded: Boolean? = null,
     expandActionContentDescription: String? = null,
     collapseActionContentDescription: String? = null,
     onClick: () -> Unit = {},
+    subContent: @Composable () -> Unit = {},
```
gabrielluong, line 49: Let's rename this to `actions`. I think we should name the parameter accordingly to the expected 'content' that would be added at the end of the header for this component instead of a more generic name like `subContent`, which you would expect to see for something like `Button`.

PR #24473, opened 2022-03-26
For #24222: Persist user interactions with nimbus messages

Amejia481:
### Pull Request checklist

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/gleanplumb/OnDiskMessageMetadataStorage.kt
```diff
@@ -0,0 +1,97 @@
...
+class OnDiskMessageMetadataStorage(
+    private val context: Context
+) : MessageMetadataStorage {
+    private val diskCacheLock = Any()
+
+    @VisibleForTesting
+    internal var metadataMap: MutableMap<String, Message.Metadata> = hashMapOf()
+
+    override suspend fun getMetadata(): Map<String, Message.Metadata> {
+        return if (metadataMap.isEmpty()) {
```
rocketsroger, line 30: nit: I think we can simplify this since both returns metadataMap

rocketsroger (approved):
Looks good to me.  Small nit.

app/src/main/java/org/mozilla/fenix/gleanplumb/OnDiskMessageMetadataStorage.kt
Amejia481: 👍🏽

Amejia481:
> Looks good to me. Small nit. Also, should we create an issue to track implementing clean up of the storage?

Sure

app/src/main/java/org/mozilla/fenix/gleanplumb/OnDiskMessageMetadataStorage.kt
```diff
@@ -0,0 +1,95 @@
...
+import org.json.JSONArray
+import org.json.JSONObject
+import java.io.File
+
+internal const val FILE_NAME = "nimbus_messages_metadata.json"
+
+/**
+ * A storage that persists [Message.Metadata] into disk.
+ */
+class OnDiskMessageMetadataStorage(
```
jonalmeida, line 21: This looks like an incomplete CRUD storage layer - where do we delete the metadata? It looks like we're always appending metadata and not considering user storage.

For comparsion with tab state, that is a core part of the browser, so we should avoid losing tab state data where we can (even then, we delete tab data when the tab is removed). With messaging metadata, the remote server can keep sending our users data and we have no way to control this if the _core browser_ is not able to function properly.

app/src/main/java/org/mozilla/fenix/gleanplumb/OnDiskMessageMetadataStorage.kt
```diff
@@ -0,0 +1,95 @@
...
+    internal fun writeToDisk() {
+        synchronized(diskCacheLock) {
+            val json = metadataMap.values.toList().fold("") { acc, next ->
+                if (acc.isEmpty()) {
+                    next.toJson()
+                } else {
+                    "$acc,${next.toJson()}"
+                }
+            }
+            getFile().writeString { "[$json]" }
```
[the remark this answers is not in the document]
Amejia481: `writeString` is silently handling the catching the exceptions

image

image

app/src/main/java/org/mozilla/fenix/gleanplumb/OnDiskMessageMetadataStorage.kt
Amejia481: I remembered @jhugman and I chatted about that, but we ended agreeing that the amount of data that we are saving could take decades, at the rate that we are sending messages to be significant for users. To implement the purging we will need that Nimbus indicates us when messages is not relevant anymore of us to delete. Happy to opened about and re-explore the possibility :)

app/src/main/java/org/mozilla/fenix/gleanplumb/OnDiskMessageMetadataStorage.kt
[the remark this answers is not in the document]
jonalmeida: Great, thanks! I didn't catch this in the extension function. 🙂

app/src/main/java/org/mozilla/fenix/gleanplumb/OnDiskMessageMetadataStorage.kt
jonalmeida: I think we should handle protecting our users if there is a bug a backend service. We do the inverse to avoid DDOS-ing our services from our users.

Amejia481: That's a valid point 👍🏽 !

Issue #24389, opened 2022-03-22
Newly created collections are not highlighted


Mugurell:
Creating a new collection from tabs tray and pressing "View" in the snackbar should scroll to and focus to the newly created collection while also animating a border around it, something implemented in https://github.com/mozilla-mobile/fenix/issues/2899.

Seems like this functionality has been broken for a while because it was based on a [hardcoded number of homescreen sections above collections](https://github.com/mozilla-mobile/fenix/blob/389cb91267f9113c3679120026d081b4a068f573/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt#L1183-L1187) and there have been many updates to these sections including an option to enabled or disable them which would've affect the functionality.

The old functionality can still be seen if disabling all homescreen sections above collections:

https://user-images.githubusercontent.com/11428869/159529127-cbe07d07-b47c-40e6-b9f0-fa6d0b0ae794.mp4

&nbsp;

Since this isn't working for quite some time maybe the homescreen migration to compose is the right time to assess whether this functionality is needed and should be migrated or not.
@topotropic  @kbrosnan

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-19924)

cpeterso:
Moved to bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1809827

Change performed by the [Move to Bugzilla add-on](https://addons.mozilla.org/en-US/firefox/addon/move-to-bugzilla/).

PR #24501, opened 2022-03-29
For #24333 - Migrate CollectionViewHolder to Compose

### LINKED ISSUE #24333: Migrate CollectionViewHolder to Compose
Breakage of https://github.com/mozilla-mobile/fenix/issues/22755

This refers to:
- [CollectionViewHolder](https://github.com/mozilla-mobile/fenix/blob/f7d317fb4ea76428d4cc9578b76bd9ef6edf8ca0/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/CollectionViewHolder.kt) showing an individual collection. (maybe instead of adding individual composables to the homescreen for each collection we could wrap them all in a list)

The compose versions can be added to the current RecyclerView but before doing so we should ensure there will be no performance regression introduced.
To profile the differences the steps from https://github.com/mozilla-mobile/fenix/issues/21854#issuecomment-1049392266 can be used.

Mugurell:
Fixes #24333

https://user-images.githubusercontent.com/11428869/160660421-c2a0978f-1179-416a-b835-13e69aa57d5e.mp4

Before and after layout bounds:

| Before | After |
| --- | --- |
|  |  |

### Pull Request checklist

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/compose/list/ExpandableListHeader.kt
```diff
@@ -45,16 +45,21 @@ fun ExpandableListHeader(
     expanded: Boolean? = null,
     expandActionContentDescription: String? = null,
     collapseActionContentDescription: String? = null,
-    onClick: () -> Unit = {},
+    onClick: (() -> Unit)? = null,
```
Mugurell, line 48: These changes were needed to be able to handle clicks in the parent and in so show the expected indication.
Otherwise, the indication (and clicks) would look like

https://user-images.githubusercontent.com/11428869/160660920-d6c535b8-6db0-45fd-9a18-4081fc2bc400.mp4

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -534,11 +526,12 @@ class HomeFragment : Fragment() {
 
         if (bundleArgs.getBoolean(FOCUS_ON_ADDRESS_BAR)) {
             navigateToSearch()
-        } else if (bundleArgs.getLong(FOCUS_ON_COLLECTION, -1) >= 0) {
-            /* Triggered when the user has added a tab to a collection and has tapped
-            * the View action on the [TabsTrayDialogFragment] snackbar.*/
-            scrollAndAnimateCollection(bundleArgs.getLong(FOCUS_ON_COLLECTION, -1))
         }
+        // else if (bundleArgs.getLong(FOCUS_ON_COLLECTION, -1) >= 0) {
```
MozillaNoah, line 530: Should this just be deleted?

app/src/main/java/org/mozilla/fenix/home/collections/CollectionMenu.kt
```diff
@@ -0,0 +1,142 @@
...
+    menuItems: List<CollectionMenuItemData>,
+    onDismissRequest: () -> Unit
+) {
+    DisposableEffect(LocalConfiguration.current.orientation) {
+        onDispose { onDismissRequest() }
+    }
+
+    // DropdownMenu uses the medium shape from MaterialTheme.
+    // Override it's corner radius to be the same 8.dp as in mozac_browser_menu_corner_radius
+    MaterialTheme(shapes = MaterialTheme.shapes.copy(medium = RoundedCornerShape(8.dp))) {
```
MozillaNoah, line 53: Should we still be using FireFoxTheme here?

app/src/main/java/org/mozilla/fenix/home/collections/CollectionMenu.kt
```diff
@@ -0,0 +1,142 @@
...
+                .background(color = FirefoxTheme.colors.layer2)
+        ) {
+            for (item in menuItems) {
+                DropdownMenuItem(
+                    onClick = {
+                        onDismissRequest()
+                        item.onClick()
+                    },
+                ) {
+                    Text(
```
MozillaNoah, line 67: does this `Text` need any text size attributes to be set here?

app/src/main/java/org/mozilla/fenix/home/collections/CollectionMenu.kt
```diff
@@ -0,0 +1,142 @@
...
+ * Menu shown for a [TabCollection].
+ *
+ * @see [DropdownMenu]
+ *
+ * @param showMenu Whether this is currently open and visible to the user.
+ * @param menuItems List of options shown.
+ * @param onDismissRequest Called when the user chooses a menu option or requests to dismiss the menu.
+ */
+@Composable
+fun CollectionMenu(
```
MozillaNoah, line 42: can we get a `@Preview` for this?

app/src/main/java/org/mozilla/fenix/home/collections/CollectionViewHolder.kt
```diff
@@ -0,0 +1,194 @@
...
+     * @param expanded Whether to show the collection as expanded or collapsed.
+     */
+    fun bindSession(collection: TabCollection, expanded: Boolean) {
+        this.collection = collection
+        isExpanded = expanded
+    }
+
+    companion object {
+        val LAYOUT_ID = View.generateViewId()
+        const val maxTitleLength = 20
```
MozillaNoah, line 97: is this constant used anywhere?

app/src/main/java/org/mozilla/fenix/home/collections/CollectionViewHolder.kt
```diff
@@ -0,0 +1,194 @@
...
+private val collapsedCollectionShape = RoundedCornerShape(8.dp)
+
+/**
+ * Rectangular shape with only the top corners rounded used to display an expanded collection with other views
+ * placed immediately below this which can be shown immediately next to it, with no visible separation.
+ */
+private val expandedCollectionShape = RoundedCornerShape(topStart = 8.dp, topEnd = 8.dp)
+
+@Composable
+private fun Collection(
```
MozillaNoah, line 113: this file feels pretty large with the Composable living here. could we possibly get it in its own file?

app/src/main/java/org/mozilla/fenix/home/collections/CollectionViewHolder.kt
```diff
@@ -0,0 +1,194 @@
...
+    collection: TabCollection,
+    expanded: Boolean,
+    menuItems: List<CollectionMenuItemData>,
+    interactor: CollectionInteractor,
```
MozillaNoah, line 117: similar to `CollectionMenu`, could you add a `@Preview` for this too?

app/src/main/java/org/mozilla/fenix/home/collections/CollectionViewHolder.kt
```diff
@@ -0,0 +1,194 @@
...
+        ) {
+            Icon(
+                painter = painterResource(R.drawable.ic_tab_collection),
+                contentDescription = null,
+                modifier = Modifier.padding(
+                    start = 16.dp,
+                    end = 8.dp // (24.dp - 16.dp) hardcoded in ExpandableListHeader
+                ),
+                tint = Paint().apply {
+                    color = Color(collection.getIconColor(LocalContext.current))
```
MozillaNoah, line 144: should we maybe get a function in `FirefoxTheme` that handles this? that way we also can just do something like `tint = FirefoxTheme.colors.getCollectionColor()`

app/src/main/java/org/mozilla/fenix/home/collections/CollectionViewHolder.kt
```diff
@@ -0,0 +1,194 @@
...
+                            onClick = {
+                                interactor.onCollectionShareTabsClicked(
+                                    collection
+                                )
+                            }
+                        ) {
+                            Icon(
+                                painter = painterResource(R.drawable.ic_share),
+                                contentDescription = stringResource(R.string.share_button_content_description),
+                                tint = FirefoxTheme.colors.textPrimary
```
MozillaNoah, line 165: I believe this should be `FirefoxTheme.colors.iconPrimary`

app/src/main/java/org/mozilla/fenix/home/collections/CollectionViewHolder.kt
```diff
@@ -0,0 +1,194 @@
...
+                                isMenuShown = !isMenuShown
+                                interactor.onCollectionMenuOpened()
+                            }
+                        ) {
+                            Icon(
+                                painter = painterResource(R.drawable.ic_menu),
+                                contentDescription = stringResource(
+                                    R.string.collection_menu_button_content_description
+                                ),
+                                tint = FirefoxTheme.colors.textPrimary
```
MozillaNoah, line 180: I believe this should be `FirefoxTheme.colors.iconPrimary`

MozillaNoah (commented):
first round of comments. I've tested locally and it seems to be working great so far!

app/src/main/java/org/mozilla/fenix/compose/list/ExpandableListHeader.kt
MozillaNoah: that's a weird interaction. thanks for the fix. this will also make it so any headers without click lambdas won't show the click ripple.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -996,134 +989,6 @@ class HomeFragment : Fragment() {
...
-                }
-                false
-            }
-        return result
-    }
-
-    /**
-     * Will highlight the border of the collection with the specified index.
-     * */
-    private fun animateCollection(indexOfCollection: Int) {
```
gabrielluong, line 1061: Did we discuss with Product or UX about whether or not they cared about keeping this interaction? Just want to know if we need to re-implement this or not.

Mugurell: Removing this is a separate commit for which I've added in the description that a reassessment of the functionality is to happen in https://github.com/mozilla-mobile/fenix/issues/24389.

app/src/main/java/org/mozilla/fenix/home/collections/CollectionViewHolder.kt
Mugurell: Thank you!
`textPrimary` was [the color used in xml](https://github.com/mozilla-mobile/fenix/blob/f924f744348006e2cb01fb172486fc7bfb1b67dc/app/src/main/res/drawable/ic_share.xml#L11).
`FirefoxTheme.colors.iconPrimary` seems to have the same values as `FirefoxTheme.colors.textPrimary` and does fit better in the context, will change to that.

app/src/main/java/org/mozilla/fenix/home/collections/CollectionViewHolder.kt
Mugurell: Thank you, will change to that.

app/src/main/java/org/mozilla/fenix/home/collections/CollectionViewHolder.kt
Mugurell: The color seems to [depend on the particular collection for which it will be used](https://github.com/mozilla-mobile/fenix/blob/f924f744348006e2cb01fb172486fc7bfb1b67dc/app/src/main/java/org/mozilla/fenix/ext/TabCollection.kt#L19).
Case in which the current approach of it being an extension function to the collection seems a bit cleaner by having a smaller scope - if specs change about the collection color we wouldn't need to modify the app theme.

app/src/main/java/org/mozilla/fenix/home/collections/CollectionViewHolder.kt
Mugurell: Will do.

app/src/main/java/org/mozilla/fenix/home/collections/CollectionViewHolder.kt
Mugurell: Sure.

app/src/main/java/org/mozilla/fenix/home/collections/CollectionViewHolder.kt
Mugurell: This was [inherited from the previous ViewHolder](https://github.com/mozilla-mobile/fenix/blob/f924f744348006e2cb01fb172486fc7bfb1b67dc/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/CollectionViewHolder.kt#L109) and seems to be used in [TabCollectionStorage](https://github.com/mozilla-mobile/fenix/blob/f924f744348006e2cb01fb172486fc7bfb1b67dc/app/src/main/java/org/mozilla/fenix/components/TabCollectionStorage.kt#L112) .

app/src/main/java/org/mozilla/fenix/home/collections/CollectionMenu.kt
Mugurell: Don't think so. This menu follows the other used for homescreen composables like the one [for RecentVisits](https://github.com/mozilla-mobile/fenix/blob/f924f744348006e2cb01fb172486fc7bfb1b67dc/app/src/main/java/org/mozilla/fenix/home/recentvisits/view/RecentlyVisited.kt#L328-L335), [for RecentBookmarks](https://github.com/mozilla-mobile/fenix/blob/f924f744348006e2cb01fb172486fc7bfb1b67dc/app/src/main/java/org/mozilla/fenix/home/recentbookmarks/view/RecentBookmarks.kt#L205-L212) or [for RecentTabs](https://github.com/mozilla-mobile/fenix/blob/f924f744348006e2cb01fb172486fc7bfb1b67dc/app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabs.kt#L488-L495).

app/src/main/java/org/mozilla/fenix/home/collections/CollectionMenu.kt
Mugurell: `CollectionMenu` is expected to be in a composition with `FirefoxTheme` applied because it is hardcoded to read `FirefoxTheme.colors`.
`DropDownMenu` is expected to be in a composition with `MaterialTheme` applied because it is implemented as a `Card` which [is hardcoded to use `MaterialTheme.shapes.medium`](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/material/material/src/commonMain/kotlin/androidx/compose/material/Card.kt;l=61;drc=4f2a25b8f8ee02053c57b524a1a6341e1edb9c06) as it's shape for which here we override the actual property.

While not pretty we will have to do something like this in multiple places in our app where we'd use android composables but with our own styling on top and so needing to override the defaults.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
Mugurell: Yes, the functionality that this would call was dropped. Thank you!

app/src/main/java/org/mozilla/fenix/home/collections/CollectionViewHolder.kt
gabrielluong, line 117: We should avoid passing the `interactor` into any @Composable functions. Can you add their equivalent `onXXXX` callback function?

gabrielluong:
While there is no rush, I think we should avoid landing this until after the beta cut next Tuesday given that soft freeze is tomorrow.

app/src/main/java/org/mozilla/fenix/home/collections/CollectionMenu.kt
MozillaNoah: gotcha gotcha. I just get a little anxious seeing key parameters not initialized 😅

app/src/main/java/org/mozilla/fenix/home/collections/CollectionViewHolder.kt
MozillaNoah: gotcha gotcha. part of my concern was this function call not working within a preview, so fingers crossed

app/src/main/java/org/mozilla/fenix/home/collections/CollectionMenu.kt
Mugurell: I know 😅. A reusable `Menu` / `MenuItem` might help with this, with them hardcoding a certain style to also ensure nothing changes in how the menu looks if some defaults change in Compose.

app/src/main/java/org/mozilla/fenix/home/collections/CollectionMenu.kt
MozillaNoah: I've had this issue occur usually because of the `components` object, in case that's what's causing it here. I may poke around later to also take a look.

Mugurell:
- Fixed the collections leakage.
Keeping the composable around to be recycled and binding new data in the viewholder to be observed for recomposition is just a cumbersome flow but it seemed to work in end.

- The shadow issue between the collection and the first containing tab is to be resolved in the migration of that containing tab item. The shadow is actually from this first tab, not from the collection.



- Adding Oana as a reviewer for disabling the collections ui tests.

sv-ohorvath (approved):
Looks ok, we agreed it is better to comment out the code in this case because otherwise there would be too much to start over again.

gabrielluong:
Sorry for the delay, will try to get to this next week.

Mugurell:
Rebasing to rerun flaky ui test.

Mugurell:
Added new screenshots showing the layout bounds based on which when comparing side-by-side I saw that the chevrons `ExpandableListHeader` uses are the reversed of what's currently being used in xml and the reverse of that clicking the header will do. Added a new commit (second) to address this. @MozillaNoah .

gabrielluong:
Should be getting to this tomorrow.

app/src/main/java/org/mozilla/fenix/compose/list/ExpandableListHeader.kt
gabrielluong: Extracted this PR out in https://github.com/mozilla-mobile/fenix/pull/24898

app/src/main/java/org/mozilla/fenix/compose/list/ExpandableListHeader.kt
```diff
@@ -76,7 +76,7 @@ fun ExpandableListHeader(
 
                 Icon(
                     painter = painterResource(
-                        if (expanded) R.drawable.ic_chevron_down else R.drawable.ic_chevron_up
+                        if (expanded) R.drawable.ic_chevron_up else R.drawable.ic_chevron_down
```
gabrielluong, line 79: image

Not sure what to make of this right now since collections actually does the opposite of what the component library is proposing. Will check with @cwzilla tomorrow.

gabrielluong: You're probably right though when I look at https://material.io/components/lists#specs. Again will check with @cwzilla tomorrow. I would generally recommend moving these types of fixes into a separate PR and having them land immediately since the changes are easy to verify and small in nature so very easy to review and just land.

app/src/main/java/org/mozilla/fenix/home/collections/CollectionMenu.kt
```diff
@@ -0,0 +1,198 @@
...
+ * @param onDeleteCollectionTapped Invoked when the user chooses to add tabs to [collection].
+ * @param onDeleteCollectionTapped Invoked when the user chooses to delete [collection].
+ */
+@Composable
+fun getMenuItems(
+    collection: TabCollection,
+    onCollectionOpenTabsTapped: (TabCollection) -> Unit = {},
+    onRenameCollectionTapped: (TabCollection) -> Unit = {},
+    onCollectionAddTabTapped: (TabCollection) -> Unit = {},
+    onDeleteCollectionTapped: (TabCollection) -> Unit = {},
```
gabrielluong, line 117: Since we always expect the consumer to provide a callback for these, I would remove the default parameter to enforce this.

app/src/main/java/org/mozilla/fenix/home/collections/CollectionViewHolder.kt
```diff
@@ -0,0 +1,103 @@
...
+}
+
+/**
+ * Wrapper over a [TabCollection] adding information about whether it should be shown as expanded or collapsed.
+ *
+ * @property collection [TabCollection] to display.
+ * @property isExpanded Whether the collection is expanded to show it's containing tabs or not.
+ */
+@Stable
+private data class CollectionInfo(
```
gabrielluong, line 100: Can this just be `Collection`?

gabrielluong (commented):
This seems pretty good to me. Can you also rebase this since we landed some of your commits since they were good on their own? 

I am still wrestling with the timing of when to land this. Would landing all of your collections change after the beta cut sound good to you since the code freeze is this thursday?

app/src/main/java/org/mozilla/fenix/home/collections/CollectionViewHolder.kt
Mugurell: Trying this rename I get
`Class 'CollectionInfo' will be shadowed by function 'Collection'`
as the Composable showing the collection already has this name.
There are quite a few `collection` related constructs. In this case `CollectionInfo` I think helps express a bit more about what it represents and help in knowing how and where to use it.

gabrielluong: Ok. I was a bit effy about the `Info` suffix since we didn't any data classes that ended in that suffix in Fenix compare to `VisitInfo` in AC.

Mugurell:
> This seems pretty good to me. Can you also rebase this since we landed some of your commits since they were good on their own? My one feedback would be to try to break standalone commits into their PR so we can just land that separately and more quickly instead of keeping the commit inside a big (900+ lines) PR. Also, one of your commits references the wrong issue number.

Rebased now and addressed your suggestions. Thank you!
Regarding breaking this in multiple patches I feel that actually means more work. 
This has sense as a whole and it should be easier and faster to work on, review and test as a whole.

> I am still wrestling with the timing of when to land this. Would landing all of your collections change after the beta cut sound good to you since the code freeze is this thursday?

I'm happy with what I can get. Agree that it would be better to wait a new train to have time to integrate and test all collection migrations and limit bigger issues.

Mugurell:
Rebased to fix conflicts.
@gabrielluong Was still left with two changes related to removing the xml border for xml collections which happened in https://github.com/mozilla-mobile/fenix/pull/25045.
Seems like that didn't contain the removal of `session_border.xml` and the dimens value it would use - `tab_border_width`. Not sure on whether there's a reason to keep them or not.

gabrielluong:
> Rebased to fix conflicts. @gabrielluong Was still left with two changes related to removing the xml border for xml collections which happened in #25045. Seems like that didn't contain the removal of `session_border.xml` and the dimens value it would use - `tab_border_width`. Not sure on whether there's a reason to keep them or not.

I only landed the piece that remove the collection animation in a separate PR since that was good to go. I couldn't remove those xml and dimen since those were still tied to the existing files. I am reviewing this today.

app/src/main/java/org/mozilla/fenix/home/collections/CollectionViewHolder.kt
```diff
@@ -0,0 +1,103 @@
...
+    private var collectionData = CollectionInfo()
+
+    @Composable
+    override fun Content() {
+        val collectionInfo by remember { mutableStateOf(collectionData) }
+
+        collectionInfo.collection?.let { validCollection ->
+            val horizontalPadding =
+                composeView.resources.getDimensionPixelSize(R.dimen.home_item_horizontal_margin)
+            composeView.setPadding(horizontalPadding, 0, horizontalPadding, 0)
```
gabrielluong, line 49: Can we move this out into the `init`? I think we already do this everywhere else.

app/src/main/java/org/mozilla/fenix/home/collections/CollectionViewHolder.kt
```diff
@@ -0,0 +1,103 @@
...
+    fun bindSession(collection: TabCollection, expanded: Boolean) {
+        collectionData = CollectionInfo(
+            collection = collection,
+            isExpanded = expanded
+        )
+    }
+
+    companion object {
+        val LAYOUT_ID = View.generateViewId()
+        const val maxTitleLength = 20
```
gabrielluong, line 89: Can we move this out to where it's being called in `TabCollectionStorage` as a constant? Considering that the days for keeping `CollectionViewHolder` is probably limited we are probably prefer serve to refactor it now. I only bring this refactoring up because the review UI makes this file as appear as entirely new, even though I know we refactored it correctly.

app/src/main/java/org/mozilla/fenix/home/collections/CollectionViewHolder.kt
```diff
@@ -0,0 +1,103 @@
...
+    viewLifecycleOwner: LifecycleOwner,
+    private val interactor: CollectionInteractor,
+) : ComposeViewHolder(composeView, viewLifecycleOwner) {
+    private var collectionData = CollectionInfo()
+
+    @Composable
+    override fun Content() {
+        val collectionInfo by remember { mutableStateOf(collectionData) }
+
+        collectionInfo.collection?.let { validCollection ->
```
gabrielluong, line 46:
```suggestion
        collectionInfo.collection?.let { collection ->
```

gabrielluong (approved):

gabrielluong:
Did we have also have a plan on addressing the line that appears at the bottom when it is expanded?

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlAdapter.kt
```diff
@@ -331,6 +335,11 @@ class SessionControlAdapter(
                 // This View already listens and maps store updates. Avoid creating and binding new Views.
                 // The composition will live until the ViewTreeLifecycleOwner to which it's attached to is destroyed.
             }
+            is CollectionViewHolder -> {
+                // Dispose the underlying composition immediately.
+                // This ViewHolder can be removed / re-added and we need it to show a fresh new composition.
+                holder.composeView.disposeComposition()
```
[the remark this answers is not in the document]
Mugurell: Good question.
Right now all composable viewholders will set a different than default strategy [here](https://github.com/mozilla-mobile/fenix/blob/58a6fafabc0477f6fa8f56adab8d37f849869e52/app/src/main/java/org/mozilla/fenix/compose/ComposeViewHolder.kt#L35-L37).
[DisposeOnViewTreeLifecycleDestroyed](https://developer.android.com/reference/kotlin/androidx/compose/ui/platform/ViewCompositionStrategy.DisposeOnViewTreeLifecycleDestroyed) rather than the default [DisposeOnDetachedFromWindow](https://developer.android.com/reference/kotlin/androidx/compose/ui/platform/ViewCompositionStrategy.DisposeOnDetachedFromWindow).

For them being inside a RecyclerView that may recycle holders the default would mean that when the holders are offscreen they would get recycled and the composition disposed 
to afterwards having to be recreate it when we'd need to show the same homescreen section again.
So setting `DisposeOnViewTreeLifecycleDestroyed` was a performance improvement. Create the holder, the composition and do not dispose it while the user is on the homescreen, only pay the penalty of creating these once.

This worked great for composable viewholders which would observe the store and update themselves.
In the case of the collections viewholders the update flow is a bit more convoluted. They will not observe the store but have the state injected.
In this case I had to go back to the default - `DisposeOnDetachedFromWindow` strategy to avoid them being still being shown on the screen after being deleted to prevent state leakage - the old holders showing stale data - what they previously displayed. This can be seen by commenting that line and deleting the collection / collection items and then undoing that action.
Keeping the default `DisposeOnDetachedFromWindow` in these cases means basically the same flow as for the xml version and ties nicely with the current way of passing data to these holders (highly convoluted, will probably need to be refactored later).
`CollectionViewHolder` could do something similar to [`ComposeViewHolder`](https://github.com/mozilla-mobile/fenix/blob/58a6fafabc0477f6fa8f56adab8d37f849869e52/app/src/main/java/org/mozilla/fenix/compose/ComposeViewHolder.kt#L35-L37) and override the strategy in it's init so we'd not have to do this `disposeComposition()` here but that would mean the holder knows where and how it is used, which it shouldn't, so I kept the default for it but managed the composition inside the adapter which is the actual component responsable for deciding what gets shown.

=> This approach is useful when state is passed from outside to avoid state leakage. 
We only need it because the compose strategy was overriden to be `DisposeOnViewTreeLifecycleDestroyed` and so holders will still remain on the screen until the screen gets destroyed.

Behavior without this:

https://user-images.githubusercontent.com/11428869/167076322-cd42dcd3-bae1-4396-9ffd-ce604e399fc7.mp4

app/src/main/java/org/mozilla/fenix/home/collections/Collection.kt
```diff
@@ -0,0 +1,213 @@
...
+    val isExpanded by remember(collection) { mutableStateOf(expanded) }
+
+    Card(
+        modifier = Modifier
+            .semantics(mergeDescendants = true) {}
+            .clickable { onToggleCollectionExpanded(collection, !isExpanded) }
+            .height(48.dp),
+        shape = if (isExpanded) expandedCollectionShape else collapsedCollectionShape,
+        backgroundColor = FirefoxTheme.colors.layer2,
+        elevation = 5.dp, // This needs to match the elevation of the TabInCollectionViewHolder for the shadow.
```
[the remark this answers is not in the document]
Mugurell: There is a matching comment in the new `TabInCollection` composable - https://github.com/mozilla-mobile/fenix/pull/24682/files#diff-fea09fe41897bb0e80f1fc96ae79d95b4fca055d5a548a5500fcb62a01ede580R118. With how problematic the shadows where I feel like it's an important detail. Will update the comment to rather point to `TabInCollection`. Can remov both if you think they aren't helping.

app/src/main/java/org/mozilla/fenix/home/collections/CollectionMenu.kt
```diff
@@ -0,0 +1,210 @@
...
+    menuItems: List<CollectionMenuItem>,
+    onDismissRequest: () -> Unit,
+) {
+    DisposableEffect(LocalConfiguration.current.orientation) {
+        onDispose { onDismissRequest() }
+    }
+
+    // DropdownMenu uses the medium shape from MaterialTheme.
+    // Override it's corner radius to be the same 8.dp as in mozac_browser_menu_corner_radius
+    MaterialTheme(shapes = MaterialTheme.shapes.copy(medium = RoundedCornerShape(8.dp))) {
```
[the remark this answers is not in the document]
Mugurell: `DropdownMenu` [shows a `Popup` which shows a `DropdownMenuContent`](https://cs.android.com/androidx/platform/tools/dokka-devsite-plugin/+/master:testData/compose/source/androidx/compose/material/AndroidMenu.android.kt;l=95-106;drc=6fed3de7a56143de954d55e508a7449deb9af582).

`DropdownMenuContent` [shows a `Card`](https://cs.android.com/androidx/platform/tools/dokka-devsite-plugin/+/master:testData/compose/source/androidx/compose/material/Menu.kt;l=116;drc=6fed3de7a56143de954d55e508a7449deb9af582).

`Card` [uses by default `MaterialTheme.shapes.medium`](https://cs.android.com/androidx/platform/tools/dokka-devsite-plugin/+/master:testData/compose/source/androidx/compose/material/Card.kt;l=59;drc=6fed3de7a56143de954d55e508a7449deb9af582).

Mugurell:
> Did we have also have a plan on addressing the line that appears at the bottom when it is expanded?

That seemed to actually be from the collection items.
Handled here https://github.com/mozilla-mobile/fenix/pull/24682/files#diff-fea09fe41897bb0e80f1fc96ae79d95b4fca055d5a548a5500fcb62a01ede580R193-R205.

Mugurell:
Thank for your help on this @gabrielluong!
Waiting on one small detail before merging - https://github.com/mozilla-mobile/fenix/pull/24501#discussion_r866523405.

app/src/main/java/org/mozilla/fenix/home/collections/Collection.kt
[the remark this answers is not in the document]
gabrielluong: I would pick between removing it or referencing `home_item_elevation` since I would assume matching the card elevation between the two would be obvious, but maybe the best solution is just have a constant that is used by both. Looking ahead, were we thinking of joining the 2 ViewHolders in some way once we are done migrating both of them into Compose. For example, would it be possible to have this all under a single Card in a followup after we landed `TabInCollection`?

app/src/main/java/org/mozilla/fenix/home/collections/CollectionMenu.kt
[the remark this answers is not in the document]
gabrielluong: Thanks! I think this does warrant a followup issue to investigate the DropdownMenu styles vs BrowserMenu and ensuring we are aligning everywhere. Filed https://github.com/mozilla-mobile/fenix/issues/25149

app/src/main/java/org/mozilla/fenix/home/collections/Collection.kt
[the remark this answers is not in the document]
gabrielluong: I also suspect that we should check with UX whether or not we should keep this at 5dp since all the other cards on Home is using 6dp, but this can be a followup.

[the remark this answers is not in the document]
Mugurell: | For example, would it be possible to have this all under a single Card in a followup after we landed TabInCollection
In the future I'd hope so. Having all the collection details - header + items in a single Composable will probably help resolve many of the current issues with broken layouts / inconsistent state.
But this will require significant work to refactor not just the UI but the entire data flow so it will probably take some time to happen.
In the meantime maybe just removing that comment is best.

Issue #24498, opened 2022-03-29
Remove Event.wrapper for History metrics


Alexandru2909:
Breakout from #21292.

SoftVision-LorandJanos:
Verified as fixed on the latest Nightly 101.0a1 (2022-04-13) by accessing history, opening tabs from history, checking closed tabs, term group, deleting history items, deleting all history.
Some events that were generated:
![image](https://user-images.githubusercontent.com/48829350/163188907-990c3297-f833-4fd5-8ac1-c96bb070c4f8.png)
![image](https://user-images.githubusercontent.com/48829350/163188961-dfc3b5a9-5ffe-44f7-b7ec-dd4e259d26b2.png)
Device used: Samsung Galaxy Tab S3 (Android 9).
Closing the ticket as fixed.

PR #24517, opened 2022-03-30
History group undo

### LINKED ISSUE #24488: [Bug]: Browser crashes
Galaxy A52 (Android 12)
Precondition:
1. The browser is installed and running
2. Any URL has been copied to the buffer in advance.

Steps:
1. Click the browser icon "Mozilla FireFox"
2. Touch and hold the address bar area until the context menu appears.
3. Touch the "paste" item in the context menu.

Expected Result:
Pasted URL is displayed in the address bar

Actual result:
Browser crashes.

A comment:
If the keyboard opens during step 2, close it otherwise the bug will not reproduce.

Screen record:
https://drive.google.com/file/d/1SoQVlM0lfOct1KXl9RbB1BDQaD8GxFup/view?usp=sharing

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-19983)

### LINKED ISSUE #24513: Undo snackbar for history groups
Part of this [jira](https://mozilla-hub.atlassian.net/browse/FNXV2-18111?focusedCommentId=466569) ticket.

Implement undo mechanism on history group screen to reduce accidental removals and make it consistent with the History Screen.

mavduevskiy:
Closes #24513

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/appstate/AppAction.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/appstate/AppAction.kt
@@ -17,6 +17,7 @@ import org.mozilla.fenix.home.recentbookmarks.RecentBookmark
 import org.mozilla.fenix.home.recentsyncedtabs.RecentSyncedTabState
 import org.mozilla.fenix.home.recenttabs.RecentTab
 import org.mozilla.fenix.home.recentvisits.RecentlyVisitedItem
+import org.mozilla.fenix.library.history.PendingDeletionHistory
 import org.mozilla.fenix.gleanplumb.Message
 import org.mozilla.fenix.gleanplumb.MessagingState
 
@@ -57,6 +58,14 @@ sealed class AppAction : Action {
     data class DeselectPocketStoriesCategory(val categoryName: String) : AppAction()
     data class PocketStoriesShown(val storiesShown: List<PocketRecommendedStory>) : AppAction()
     data class PocketStoriesChange(val pocketStories: List<PocketRecommendedStory>) : AppAction()
+    /**
+     * Adds a set of items marked for removal to the app state, to be hidden in the UI.
+     */
+    data class AddPendingDeletionSet(val historyItems: Set<PendingDeletionHistory>) : AppAction()
+    /**
+     * Removes a set of items, previously marked for removal, to be displayed again in the UI.
+     */
+    data class UndoPendingDeletionSet(val historyItems: Set<PendingDeletionHistory>) : AppAction()
     data class PocketStoriesCategoriesChange(val storiesCategories: List<PocketRecommendedStoriesCategory>) :
         AppAction()
     data class PocketStoriesCategoriesSelectionsChange(
--- a/app/src/main/java/org/mozilla/fenix/library/history/HistoryAdapter.kt
+++ b/app/src/main/java/org/mozilla/fenix/library/history/HistoryAdapter.kt
@@ -16,13 +16,21 @@ import org.mozilla.fenix.library.history.viewholders.HistoryListItemViewHolder
  */
 class HistoryAdapter(
     private val historyInteractor: HistoryInteractor,
+    private val onEmptyStateChanged: (Boolean) -> Unit,
 ) : PagingDataAdapter<History, HistoryListItemViewHolder>(historyDiffCallback),
     SelectionHolder<History> {
 
     private var mode: HistoryFragmentState.Mode = HistoryFragmentState.Mode.Normal
-    override val selectedItems get() = mode.selectedItems
-    var pendingDeletionIds = emptySet<Long>()
+    private var pendingDeletionItems = emptySet<PendingDeletionHistory>()
     private val itemsWithHeaders: MutableMap<HistoryItemTimeGroup, Int> = mutableMapOf()
+    // A flag to track the empty state of the list. Items are not being deleted immediately,
+    // but hidden from the UI until the Undo snackbar will execute the delayed operation.
+    // Whether the adapter has actually zero items or all present items are hidden,
+    // the screen should be updated into proper empty/not empty state.
+    private var isEmpty = true
+
+    override val selectedItems
+        get() = mode.selectedItems
 
     override fun getItemViewType(position: Int): Int = HistoryListItemViewHolder.LAYOUT_ID
 
@@ -38,10 +46,45 @@ class HistoryAdapter(
         if (itemCount > 0) notifyItemChanged(0)
     }
 
+    @Suppress("ComplexMethod")
     override fun onBindViewHolder(holder: HistoryListItemViewHolder, position: Int) {
         val current = getItem(position) ?: return
-        val isPendingDeletion = pendingDeletionIds.contains(current.visitedAt)
+        var isPendingDeletion = false
+        var groupPendingDeletionCount = 0
         var timeGroup: HistoryItemTimeGroup? = null
+        if (position == 0) {
+            isEmpty = true
+        }
+
+        if (pendingDeletionItems.isNotEmpty()) {
+            when (current) {
+                is History.Regular -> {
+                    isPendingDeletion = pendingDeletionItems.find {
+                        it is PendingDeletionHistory.Item && it.visitedAt == current.visitedAt
+                    } != null
+                }
+                is History.Group -> {
+                    isPendingDeletion = pendingDeletionItems.find {
+                        it is PendingDeletionHistory.Group && it.visitedAt == current.visitedAt
+                    } != null
+
+                    if (!isPendingDeletion) {
+                        groupPendingDeletionCount = current.items.count { historyMetadata ->
+                            pendingDeletionItems.find {
+                                it is PendingDeletionHistory.MetaData &&
+                                    it.key == historyMetadata.historyMetadataKey &&
+                                    it.visitedAt == historyMetadata.visitedAt
+                            } != null
+                        }.also {
+                            if (it == current.items.size) {
+                                isPendingDeletion = true
+                            }
+                        }
+                    }
+                }
+                else -> {}
+            }
+        }
 
         // Add or remove the header and position to the map depending on it's deletion status
         if (itemsWithHeaders.containsKey(current.historyTimeGroup)) {
@@ -60,11 +103,33 @@ class HistoryAdapter(
             timeGroup = current.historyTimeGroup
         }
 
-        holder.bind(current, timeGroup, position == 0, mode, isPendingDeletion)
+        // If there is a single visible item, it's enough to change the empty state of the view.
+        if (isEmpty && !isPendingDeletion) {
+            isEmpty = false
+            onEmptyStateChanged.invoke(isEmpty)
+        } else if (position + 1 == itemCount) {
+            // If we reached the bottom of the list and there still has been zero visible items,
+            // we can can change the History view state to empty.
+            if (isEmpty) {
+                onEmptyStateChanged.invoke(isEmpty)
+            }
+        }
+
+        holder.bind(
+            current,
+            timeGroup,
+            position == 0,
+            mode,
+            isPendingDeletion,
+            groupPendingDeletionCount
+        )
     }
 
-    fun updatePendingDeletionIds(pendingDeletionIds: Set<Long>) {
-        this.pendingDeletionIds = pendingDeletionIds
+    /**
+     * @param pendingDeletionItems is used to filter out the items that should not be displayed.
+     */
+    fun updatePendingDeletionItems(pendingDeletionItems: Set<PendingDeletionHistory>) {
+        this.pendingDeletionItems = pendingDeletionItems
     }
 
     companion object {
--- a/app/src/main/java/org/mozilla/fenix/library/history/HistoryController.kt
+++ b/app/src/main/java/org/mozilla/fenix/library/history/HistoryController.kt
@@ -4,14 +4,21 @@
 
 package org.mozilla.fenix.library.history
 
+import android.content.Context
 import androidx.navigation.NavController
 import androidx.navigation.NavOptions
 import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
-import mozilla.telemetry.glean.private.NoExtras
+import mozilla.components.browser.state.action.HistoryMetadataAction
+import mozilla.components.service.glean.private.NoExtras
 import org.mozilla.fenix.GleanMetrics.Events
 import org.mozilla.fenix.R
+import org.mozilla.fenix.components.AppStore
+import org.mozilla.fenix.components.appstate.AppAction
+import org.mozilla.fenix.components.history.DefaultPagedHistoryProvider
 import org.mozilla.fenix.components.metrics.MetricController
+import org.mozilla.fenix.ext.components
 import org.mozilla.fenix.ext.navigateSafe
 import org.mozilla.fenix.GleanMetrics.History as GleanHistory
 
@@ -29,15 +36,21 @@ interface HistoryController {
     fun handleEnterRecentlyClosed()
 }
 
-@Suppress("TooManyFunctions")
+@Suppress("TooManyFunctions", "LongParameterList")
 class DefaultHistoryController(
     private val store: HistoryFragmentStore,
+    private val appStore: AppStore,
+    private var historyProvider: DefaultPagedHistoryProvider,
     private val navController: NavController,
     private val scope: CoroutineScope,
     private val openToBrowser: (item: History.Regular) -> Unit,
     private val displayDeleteAll: () -> Unit,
     private val invalidateOptionsMenu: () -> Unit,
-    private val deleteHistoryItems: (Set<History>) -> Unit,
+    private val deleteSnackbar: (
+        items: Set<History>,
+        undo: suspend (Set<History>) -> Unit,
+        delete: (Set<History>) -> suspend (context: Context) -> Unit
+    ) -> Unit,
     private val syncHistory: suspend () -> Unit,
     private val metrics: MetricController
 ) : HistoryController {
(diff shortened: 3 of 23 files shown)


mavduevskiy:
Added delayed removal and an Undo snackbar to history group screen.

Added complexity came from the need to appropriately update the group item count or hide it completely inside the history screen: item deleted inside history group screen -> user backs to history screen -> the group item should consider items pending for removal –> Undo action should updated the group item again.

Adding `pendingDeletionHistoryItems` to `AppState` allows to hide the history items, pending for removal, anywhere in the app. Communication between history group screen and history screen happen through it. Things like recently visited could use the same mechanism.

I was considering removing items from `pendingDeletionHistoryItems` after actual removal, but decided against it. Only pressing Undo button actually affects that set. The thought process was items inside the set should be hidden from the UI no matter if removal has happened.

But removing items from `pendingDeletionHistoryItems` after actual deletion had happened would require tracking if it had been an undo action and in that case we could keep the same data set, but update visibility of the item, or an actual removal, in which case data should be updated to get the list without the deleted item. 
In case of pagination scrolling it might lead to visual glitches when items would be shown again and then hidden, because we already don't have rules to hide the item locally but data set hasn't been updated yet.
Having one source of truth for hiding elements seems more reasonable.

Large data sets with a few dozens deleted items didn't have visual impact on scrolling performance. 

Added `onEmptyStateChanged` to Adapters, that changes the empty state of the history and history group screens. As we manually hide the items, that are about to be removed,  the most reliable way to determine if the list is empty or not, it to actually allow it to be drawn (changes visible/invisible) and if there is a single visible item, change state to notEmpty, and if we have reached the end of the list and there is still zero visible items, we report empty state.

app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkFragment.kt
```diff
@@ -330,8 +311,9 @@ class BookmarkFragment : LibraryPageFragment<BookmarkNode>(), UserInteractionHan
             }
         }
 
-        viewLifecycleOwner.lifecycleScope.allowUndo(
-            requireView(), message,
+        MainScope().allowUndo(
+            requireActivity().getRootView()!!,
```
Amejia481, line 315: I think we should avoid `!!` as it could cause crashes, maybe let? :)

app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkFragment.kt
```diff
@@ -330,8 +311,9 @@ class BookmarkFragment : LibraryPageFragment<BookmarkNode>(), UserInteractionHan
             }
         }
 
-        viewLifecycleOwner.lifecycleScope.allowUndo(
-            requireView(), message,
+        MainScope().allowUndo(
```
Amejia481, line 314: Is there any specific reason why we switched to `viewLifecycleOwner.lifecycleScope` ?

Amejia481: I guess it's for keeping the undo after the fragment is gone?

app/src/main/java/org/mozilla/fenix/library/downloads/DownloadFragment.kt
```diff
@@ -41,9 +42,6 @@ class DownloadFragment : LibraryPageFragment<DownloadItem>(), UserInteractionHan
     private lateinit var downloadStore: DownloadFragmentStore
     private lateinit var downloadView: DownloadView
     private lateinit var downloadInteractor: DownloadInteractor
-    private var undoScope: CoroutineScope? = null
-    private var pendingDownloadDeletionJob: (suspend () -> Unit)? = null
-    private lateinit var downloadsUseCases: DownloadsUseCases
```
Amejia481, line 46: Should this part of another ticket, as it's not part of the history group?

app/src/main/java/org/mozilla/fenix/library/history/HistoryAdapter.kt
```diff
@@ -16,13 +16,17 @@ import org.mozilla.fenix.library.history.viewholders.HistoryListItemViewHolder
...
+    private val onEmptyStateChanged: (Boolean) -> Unit,
 ) : PagingDataAdapter<History, HistoryListItemViewHolder>(historyDiffCallback),
     SelectionHolder<History> {
 
     private var mode: HistoryFragmentState.Mode = HistoryFragmentState.Mode.Normal
-    override val selectedItems get() = mode.selectedItems
-    var pendingDeletionIds = emptySet<Long>()
+    private var pendingDeletionItems = emptySet<PendingDeletionHistory>()
     private val itemsWithHeaders: MutableMap<HistoryItemTimeGroup, Int> = mutableMapOf()
+    private var isEmpty = true
```
Amejia481, line 26: `isEmpty` is a bit generic,  should we add a more descriptive name or add some comments to indicate why we are using it? WDY? :)

app/src/main/java/org/mozilla/fenix/library/history/HistoryAdapter.kt
```diff
@@ -60,11 +99,31 @@ class HistoryAdapter(
...
+            position == 0,
+            mode,
+            isPendingDeletion,
+            groupPendingDeletionCount
+        )
     }
 
-    fun updatePendingDeletionIds(pendingDeletionIds: Set<Long>) {
-        this.pendingDeletionIds = pendingDeletionIds
+    @Suppress("UndocumentedPublicFunction")
```
Amejia481, line 124: Is there any specific reason why we are suppressing ? :)

app/src/main/java/org/mozilla/fenix/components/appstate/AppAction.kt
```diff
@@ -17,12 +17,14 @@ import org.mozilla.fenix.home.recentbookmarks.RecentBookmark
...
 import org.mozilla.fenix.home.recenttabs.RecentTab
 import org.mozilla.fenix.home.recentvisits.RecentlyVisitedItem
+import org.mozilla.fenix.library.history.PendingDeletionHistory
 import org.mozilla.fenix.gleanplumb.Message
 import org.mozilla.fenix.gleanplumb.MessagingState
 
 /**
  * [Action] implementation related to [AppStore].
  */
+@Suppress("UndocumentedPublicClass")
```
rocketsroger, line 27: nit: the convention here is to add new classes to `detekt-baseline.xml`

gabrielluong:
Can we split up the 2 commits for 2 different issues into their own individual PR? Is there any reason we need to have them both in one PR?

app/src/main/java/org/mozilla/fenix/library/history/HistoryController.kt
```diff
@@ -95,7 +108,39 @@ class DefaultHistoryController(
...
+        val pendingDeletionItems = items.map { it.toPendingDeletionHistory() }.toSet()
+        appStore.dispatch(AppAction.UndoPendingDeletionSet(pendingDeletionItems))
+    }
+
+    private fun delete(items: Set<History>): suspend (context: Context) -> Unit {
+        return { context ->
+            CoroutineScope(Dispatchers.IO).launch {
+                store.dispatch(HistoryFragmentAction.EnterDeletionMode)
+                for (item in items) {
+                    GleanHistory.removed.record(NoExtras())
```
Amejia481, line 126: Should this part of `GleanMetricsService` and use the `metrics` to report ?  And following the same patter as on 
 `metrics.track(Event.HistorySearchTermGroupTapped)`

app/src/main/java/org/mozilla/fenix/library/history/HistoryFragmentStore.kt
```diff
@@ -131,7 +136,8 @@ sealed class HistoryFragmentAction : Action {
 data class HistoryFragmentState(
     val items: List<History>,
     val mode: Mode,
-    val pendingDeletionIds: Set<Long>,
+    val pendingDeletionItems: Set<PendingDeletionHistory>,
+    val isEmpty: Boolean,
```
Amejia481, line 140: Could we determine the if is empty from the `items`?

gabrielluong (changes requested):
Adding a request change to have this PR split up. I think the commit with allowing undo to work globally is probably the easier review out of the 2.

app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkFragment.kt
mavduevskiy: Hm, that's a good question! 

I don't think getRootView could actually fail after requireActivity was called successfully. The code goes like this:
`asActivity()?.window?.decorView?.findViewById<View>(android.R.id.content) as? ViewGroup`
So, `window` is there after the activity has been created and never set to null again. `decorView` is not null, and there is always a `android.R.id.content` in the decor view.

We actually have `getRootView()!!` [already](https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/java/org/mozilla/fenix/library/bookmarks/edit/EditBookmarkFragment.kt#220) used with a FenixSnackbar, and I couldn't find any crashes related to that!

Going with let would not execute the removal in case there is no proper view provided by `rootView`. Hm, may be we could provide an alternative, like a fragment view? Like `requireActivity().getRootView() ?: requiteView()`
But in that case having no access to the fragment view will also have the same affect.

I think, relying on `activity.window` on the main thread is a as safe bet as it gets. But I don't have a strong opinion on that, want to here your feedback.

app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkFragment.kt
mavduevskiy: Yes, it's not to hold a reference to `viewLifecycleOwner.lifecycleScope` in case the user leaves the fragment

app/src/main/java/org/mozilla/fenix/library/downloads/DownloadFragment.kt
mavduevskiy: Yes! It is actually part of a [Working undo](https://github.com/mozilla-mobile/fenix/pull/24398) ticket

app/src/main/java/org/mozilla/fenix/library/history/HistoryAdapter.kt
mavduevskiy: The only alternative that comes to my mind is `hasItems`, which I find to be even worse :))
Added explanation.

app/src/main/java/org/mozilla/fenix/library/history/HistoryAdapter.kt
mavduevskiy: It looks like Documentation should add something meaningful, which is not immediately obvious from the code. I can't come up with anything about a
```
fun updatePendingDeletionItems(pendingDeletionItems: Set<PendingDeletionHistory>) {
        this.pendingDeletionItems = pendingDeletionItems
    }
```
really :(

app/src/main/java/org/mozilla/fenix/components/appstate/AppAction.kt
mavduevskiy: I am a bit confused, what should I write about?
```
data class AddPendingDeletionSet(val historyItems: Set<PendingDeletionHistory>) : AppAction()
    data class UndoPendingDeletionSet(val historyItems: Set<PendingDeletionHistory>) : AppAction()
```
The name of the class tells what it does already. `pendingDeletionHistoryItems` are already documented inside the `AppState`.

This new lint rule is a bit too demanding.

app/src/main/java/org/mozilla/fenix/library/history/HistoryController.kt
mavduevskiy: I think it was done during working on `[this](https://github.com/mozilla-mobile/fenix/issues/24498)` issue.
I just brought that code from the fragment into the controller.

app/src/main/java/org/mozilla/fenix/library/history/HistoryFragmentStore.kt
mavduevskiy: Sadly, nope. Items inside the `HistoryFragmentState` are empty. Adapter is being fed data directly, avoiding the state:
```
lifecycleScope.launch {
            history.collect {
                historyView.historyAdapter.submitData(it)
            }
        }
```

mavduevskiy:
@gabrielluong 
It is actually the second PR, linked to [this one](https://github.com/mozilla-mobile/fenix/pull/24398)

app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkFragment.kt
Amejia481: Ok, that make sense. As we don't have a better alternative, let's keep it!

app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkFragment.kt
mavduevskiy: It's actually was causing [a leak](https://github.com/mozilla-mobile/fenix/pull/24398#issuecomment-1103255859)

app/src/main/java/org/mozilla/fenix/library/history/HistoryAdapter.kt
mavduevskiy: added documentation

app/src/main/java/org/mozilla/fenix/components/appstate/AppAction.kt
mavduevskiy: added documentation

app/src/main/java/org/mozilla/fenix/library/history/HistoryAdapter.kt
Amejia481: 👍🏽

app/src/main/java/org/mozilla/fenix/library/history/HistoryFragment.kt
```diff
@@ -165,10 +179,13 @@ class HistoryFragment : LibraryPageFragment<History>(), UserInteractionHandler {
             historyView.update(it)
         }
 
-        userHasHistory.observe(
-            viewLifecycleOwner,
-            historyView::updateEmptyState
-        )
+        requireContext().components.appStore.flowScoped(viewLifecycleOwner) { flow ->
```
Amejia481, line 182: nit: Maybe we could pass the `appStore` to `HistoryFragment` allow easier testing in the future

app/src/main/java/org/mozilla/fenix/library/history/HistoryFragmentStore.kt
```diff
@@ -168,13 +180,9 @@ private fun historyStateReducer(
...
-                pendingDeletionIds = state.pendingDeletionIds + action.itemIds
-            )
-        is HistoryFragmentAction.UndoPendingDeletionSet ->
-            state.copy(
-                pendingDeletionIds = state.pendingDeletionIds - action.itemIds
-            )
+        is HistoryFragmentAction.ChangeEmptyState -> state.copy(isEmpty = action.isEmpty)
+        is HistoryFragmentAction.UpdatePendingDeletionItems -> state.copy(
+            pendingDeletionItems = action.pendingDeletionItems
+        )
```
Amejia481, line 186: Maybe we could tests these section without much effort :)

app/src/main/java/org/mozilla/fenix/library/history/PendingDeletionHistory.kt
```diff
@@ -0,0 +1,58 @@
...
+
+package org.mozilla.fenix.library.history
+
+import mozilla.components.concept.storage.HistoryMetadataKey
+
+/**
+ * Wrapper for the data of the history item that have been marked for removal. Undo snackbar delays
+ * the actual removal, while this class is used to match History items that should be hidden in the
+ * UI.
+ */
```
Amejia481, line 13: 👏🏽

app/src/main/java/org/mozilla/fenix/library/historymetadata/HistoryMetadataGroupFragmentStore.kt
```diff
@@ -91,5 +104,10 @@ private fun historyStateReducer(
         }
         is HistoryMetadataGroupFragmentAction.DeleteAll ->
             state.copy(items = emptyList())
+        is HistoryMetadataGroupFragmentAction.UpdatePendingDeletionItems ->
+            state.copy(pendingDeletionItems = action.pendingDeletionItems)
+        is HistoryMetadataGroupFragmentAction.ChangeEmptyState -> state.copy(
+            isEmpty = action.isEmpty
```
Amejia481, line 110: Maybe we could easily add tests for these sections :)

app/src/main/java/org/mozilla/fenix/library/historymetadata/view/HistoryMetadataGroupAdapter.kt
```diff
@@ -34,16 +38,55 @@ class HistoryMetadataGroupAdapter(
...
+    }
+
     override fun onBindViewHolder(holder: HistoryMetadataGroupItemViewHolder, position: Int) {
-        holder.bind(getItem(position))
+        val current = getItem(position) ?: return
+        if (position == 0) {
+            isEmpty = true
+        }
+
+        val isPendingDeletion = pendingDeletionItems.find {
```
Amejia481, line 51: nit: Maybe any `any` :)

Amejia481 (approved):
LGTM!

app/src/test/java/org/mozilla/fenix/library/history/HistoryFragmentStoreTest.kt
```diff
@@ -72,6 +73,34 @@ class HistoryFragmentStoreTest {
...
     }
 
+    @Test
+    fun changeEmptyState() = runBlocking {
+        val initialState = emptyDefaultState()
+        val store = HistoryFragmentStore(initialState)
+
+        store.dispatch(HistoryFragmentAction.ChangeEmptyState(true)).join()
+        assertNotSame(initialState, store.state)
+        assertEquals(true, store.state.isEmpty)
```
Amejia481, line 83: nit: `assertTrue`

app/src/test/java/org/mozilla/fenix/library/history/HistoryFragmentStoreTest.kt
```diff
@@ -72,6 +73,34 @@ class HistoryFragmentStoreTest {
...
+
+        store.dispatch(HistoryFragmentAction.ChangeEmptyState(false)).join()
+        assertNotSame(initialState, store.state)
+        assertEquals(false, store.state.isEmpty)
```
Amejia481, line 87: nit: `assertFalse`

PR #24527, opened 2022-03-30
For #24519 - Add Private theme to FirefoxTheme

### LINKED ISSUE #24519: Add Private theme to FirefoxTheme
At a time when it wasn't clear what colors will the app use or whether we'll roll our own theme or stick with `MaterialTheme` I started work on a `FenixTheme` which had support for light/dark and private mode also - https://github.com/mozilla-mobile/fenix/pull/21193/files

_Originally posted by @Mugurell in https://github.com/mozilla-mobile/fenix/pull/24505#discussion_r838152723_

gabrielluong:
Fixes #24519 

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/compose/ClickableSubstringLink.kt
+++ b/app/src/main/java/org/mozilla/fenix/compose/ClickableSubstringLink.kt
@@ -16,6 +16,7 @@ import androidx.compose.ui.text.buildAnnotatedString
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.sp
 import org.mozilla.fenix.theme.FirefoxTheme
+import org.mozilla.fenix.theme.Theme
 
 /**
  * [Text] containing a substring styled as an URL informing when this is clicked.
@@ -86,7 +87,7 @@ fun ClickableSubstringLink(
 private fun ClickableSubstringTextPreview() {
     val text = "This text contains a link"
 
-    FirefoxTheme {
+    FirefoxTheme(theme = Theme.getTheme(isPrivate = false)) {
         Box(modifier = Modifier.background(color = FirefoxTheme.colors.layer1)) {
             ClickableSubstringLink(
                 text = text,
--- a/app/src/main/java/org/mozilla/fenix/compose/Favicon.kt
+++ b/app/src/main/java/org/mozilla/fenix/compose/Favicon.kt
@@ -26,6 +26,7 @@ import mozilla.components.browser.icons.compose.WithIcon
 import mozilla.components.ui.colors.PhotonColors
 import org.mozilla.fenix.components.components
 import org.mozilla.fenix.theme.FirefoxTheme
+import org.mozilla.fenix.theme.Theme
 
 /**
  * Load and display the favicon of a particular website.
@@ -85,7 +86,7 @@ private fun FaviconPlaceholder(size: Dp) {
 @Composable
 @Preview(uiMode = Configuration.UI_MODE_NIGHT_YES)
 private fun FaviconPreview() {
-    FirefoxTheme {
+    FirefoxTheme(theme = Theme.getTheme(isPrivate = false)) {
         Box(Modifier.background(FirefoxTheme.colors.layer1)) {
             Favicon(
                 url = "www.mozilla.com",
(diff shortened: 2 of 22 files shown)


app/src/main/java/org/mozilla/fenix/theme/FirefoxTheme.kt
```diff
@@ -17,17 +17,52 @@ import androidx.compose.runtime.remember
...
+    companion object {
+        /**
+         * Returns the current [Theme] that is displayed.
+         *
+         * @param isPrivate Whether or not private browsing mode is enabled.
+         * @return the current [Theme] that is displayed.
+         */
+        @Composable
+        fun getTheme(isPrivate: Boolean = LocalContext.current.settings().lastKnownMode.isPrivate) =
+            if (isPrivate) {
```
gabrielluong, line 41: I wanted to do `!inComposePreview && LocalContext.current.settings().lastKnownMode.isPrivate` for this conditional but hit a compiler error that looked like https://stackoverflow.com/questions/68375629/andand-condition-should-have-an-if-true-then-false.

app/src/main/java/org/mozilla/fenix/compose/Favicon.kt
```diff
@@ -85,7 +86,7 @@ private fun FaviconPlaceholder(size: Dp) {
 @Composable
 @Preview(uiMode = Configuration.UI_MODE_NIGHT_YES)
 private fun FaviconPreview() {
-    FirefoxTheme {
+    FirefoxTheme(theme = Theme.getTheme(isPrivate = false)) {
```
MozillaNoah, line 89: was setting it manually here to get the previews to build?

gabrielluong: Yes because preview doesn't know anything about `LocalContext.current.settings()` which is `components.settings`

app/src/main/java/org/mozilla/fenix/theme/FirefoxTheme.kt
```diff
@@ -42,10 +77,78 @@ object FirefoxTheme {
...
+    borderAccent = PhotonColors.Violet40,
+    borderDisabled = PhotonColors.LightGrey05A40,
+    borderWarning = PhotonColors.Red40
+)
+
 private val darkColorPalette = FirefoxColors(
     layer1 = PhotonColors.DarkGrey60,
     layer2 = PhotonColors.DarkGrey30,
     layer3 = PhotonColors.DarkGrey80,
+    layer4Start = PhotonColors.Purple70,
```
MozillaNoah, line 149: just to confirm, are these new layer colors supposed to be the same for all 3 palettes?

app/src/main/java/org/mozilla/fenix/theme/FirefoxTheme.kt
gabrielluong: Filed https://github.com/mozilla-mobile/fenix/issues/24528 for the future.

app/src/main/java/org/mozilla/fenix/theme/FirefoxTheme.kt
gabrielluong: It's undefined for Light and Dark theme since it's technically private only.

MozillaNoah (approved):

app/src/main/java/org/mozilla/fenix/compose/ClickableSubstringLink.kt
```diff
@@ -86,7 +87,7 @@ fun ClickableSubstringLink(
 private fun ClickableSubstringTextPreview() {
     val text = "This text contains a link"
 
-    FirefoxTheme {
+    FirefoxTheme(theme = Theme.getTheme(isPrivate = false)) {
```
MatthewTighe, line 90: Would it make sense to create a separate preview theme, if only to save some characters when creating previews? Something like:

```
@Composable
fun PreviewTheme(isPrivate: Boolean = false, content: @Composable () -> Unit) {
    FirefoxTheme(theme = Theme.getTheme(isPrivate = isPrivate), content = content)
}
```

gabrielluong: See https://github.com/mozilla-mobile/fenix/pull/24527#discussion_r838957235. In general, I think what you're proposing is right. The ideal solution would be to avoid having the need for either the `PreviewTheme` or `isPrivate` parameter and for `FirefoxTheme` to know it is dealing with a @Preview, but we're stuck with this until Compose 1.1.0. I don't know if it's worth the effort knowing that both of these solution would hopefully be temporary.

PR #24530, opened 2022-03-30
For #24338 - Make a reusable list item Composable

### LINKED ISSUE #24338: Make a re-usable list item in Compose
Design - https://www.figma.com/file/ujdiReypSypu2GhRgzArea/branch/ddGnMfYggXGmcTYfoI0aS8/Mobile-Android-Components?node-id=1802%3A2716
image

The item should have:

- Optional left-aligned icon that could be used for an asset or a favicon
- Primary label
- Optional description
- Optional right-aligned composable
- Support for dark mode

MozillaNoah:
Closes #24338 

V2 draft to make the root Composable private with specific use case Composables

### Pull Request checklist

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/compose/list/ListItem.kt
```diff
@@ -0,0 +1,305 @@
...
+            Spacer(modifier = Modifier.width(16.dp))
+
+            Icon(
+                painter = painterResource(drawable),
+                contentDescription = drawableDescription,
+                modifier = Modifier.size(24.dp),
+                tint = FirefoxTheme.colors.iconPrimary,
+            )
+
+            Spacer(modifier = Modifier.width(16.dp))
```
gabrielluong, line 126: Probably prefer a horizontal padding in the ListItem row

app/src/main/java/org/mozilla/fenix/compose/list/ListItem.kt
```diff
@@ -0,0 +1,305 @@
...
+@Composable
+private fun DrawableListItem(
+    label: String,
+    description: String? = null,
+    onClick: (() -> Unit)? = null,
+    @DrawableRes drawable: Int,
+    drawableDescription: String? = null,
+    @DrawableRes icon: Int? = null,
+    iconDescription: String? = null,
+    onIconCLick: (() -> Unit)? = null,
```
gabrielluong, line 110:
```suggestion
    onIconClick: (() -> Unit)? = null,
```

app/src/main/java/org/mozilla/fenix/compose/list/ListItem.kt
```diff
@@ -0,0 +1,342 @@
...
+ * @param label The primary text in the item.
+ * @param description An optional second line of text.
+ * @param onClick Optional lambda for handling clicks.
+ * @param url The tab's URL.
+ * @param icon Optional DrawableRes used to display a drawable at the end of the list item.
+ * @param iconDescription Content description of the icon.
+ * @param onIconClick Optional lambda for handling clicks on the icon.
+ */
+@Composable
+fun FavIconListItem(
```
gabrielluong, line 89:
```suggestion
fun FaviconListItem(
```
This happens to be a nice example that follows nicely with the Figma component naming. It would be nice if we can align our naming nicely with the Component System. I don't think we need a unique component for Left Favicon + Right Icon for instance, but we can take a more generic naming like Favicon List Item as the core component and know that the Right Icon, Descriptions and such will be optional way to extend a core component.

app/src/main/java/org/mozilla/fenix/compose/list/ListItem.kt
```diff
@@ -0,0 +1,342 @@
...
+                iconDescription = "click me",
+                onIconClick = { println("icon click") },
+            )
+        }
+    }
+}
+
+@Composable
+@Preview(name = "FavIconListItem with an icon and onClicks", uiMode = Configuration.UI_MODE_NIGHT_YES)
+private fun CompleteListItemPreview() {
```
gabrielluong, line 329: Any particular reason these shouldn't just be named after their component `FaviconListItemPreview`? I feel the the @Preview name does provide a helpful description of the intended preview so I am not quite sure if the actual function name needs to be different from the component. I would typically look for the 1:1 component to preview naming if I was trying to see an example of the @Preview

app/src/main/java/org/mozilla/fenix/compose/list/ListItem.kt
```diff
@@ -0,0 +1,342 @@
...
+
+val ICON_SIZE = 24.dp
+
+/**
+ * List item Composable used to display text with an optional second line of text and
+ * an optional interactable icon at the end.
+ *
+ * @param label The primary text in the item.
+ * @param description An optional second line of text.
+ * @param onClick Optional lambda for handling clicks.
```
gabrielluong, line 44:
```suggestion
 * @param onClick Called when the user clicks on the item.
```

I have mostly been emulating the kDoc conventions from the Jetpack Compose components. I am mostly looking for what actions and its subject (ex, the list item) that will invoke the callback. 

Screen Shot 2022-04-14 at 11 24 06 AM
Screen Shot 2022-04-14 at 11 24 14 AM

app/src/main/java/org/mozilla/fenix/compose/list/ListItem.kt
```diff
@@ -0,0 +1,342 @@
...
+
+val LIST_ITEM_HEIGHT = 56.dp
+
+val ICON_SIZE = 24.dp
+
+/**
+ * List item Composable used to display text with an optional second line of text and
+ * an optional interactable icon at the end.
+ *
+ * @param label The primary text in the item.
```
gabrielluong, line 42:
```suggestion
 * @param label The label in the list item.
```
nitpicking a bit on these @param doc since these are our **core** reusable components. I don't think we need to use our internal wording ("primary text") since they are called "Label" in the figma design as well.

app/src/main/java/org/mozilla/fenix/compose/list/ListItem.kt
```diff
@@ -0,0 +1,342 @@
...
+ * @param description An optional second line of text.
```
gabrielluong, line 43:
```suggestion
 * @param description An optional description text below the label.
```

app/src/main/java/org/mozilla/fenix/compose/list/ListItem.kt
```diff
@@ -0,0 +1,342 @@
...
+import org.mozilla.fenix.compose.PrimaryText
+import org.mozilla.fenix.compose.SecondaryText
+import org.mozilla.fenix.theme.FirefoxTheme
+
+val LIST_ITEM_HEIGHT = 56.dp
+
+val ICON_SIZE = 24.dp
+
+/**
+ * List item Composable used to display text with an optional second line of text and
```
gabrielluong, line 39:
```suggestion
 * List item used to display a label with an optional description text and
```
Mostly trying to avoid the internal implementation detail ("Composable") and overall just being very clear semantically.

app/src/main/java/org/mozilla/fenix/compose/list/ListItem.kt
```diff
@@ -0,0 +1,342 @@
...
+ * an optional interactable icon at the end.
+ *
+ * @param label The primary text in the item.
+ * @param description An optional second line of text.
+ * @param onClick Optional lambda for handling clicks.
+ * @param icon Optional DrawableRes used to display a drawable at the end of the list item.
```
gabrielluong, line 45:
```suggestion
 * @param icon Optional [DrawableRes] used to display a drawable at the end of the list item.
```

app/src/main/java/org/mozilla/fenix/compose/list/ListItem.kt
```diff
@@ -0,0 +1,342 @@
...
+        onClick = onClick,
+        beforeListAction = {
+            Spacer(modifier = Modifier.width(16.dp))
+
+            Favicon(
+                url = url,
+                size = ICON_SIZE,
+            )
+
+            Spacer(modifier = Modifier.width(16.dp))
```
gabrielluong, line 110: The before and after `Spacer` shouldn't be necessary since `ListItem` already has a horizontal padding of 16dp. I think we just need a 32dp Spacer here.

Screen Shot 2022-04-14 at 11 40 51 AM

app/src/main/java/org/mozilla/fenix/compose/list/ListItem.kt
```diff
@@ -0,0 +1,342 @@
...
+}
+
+/**
+ * List item Composable used to display text and a [Favicon] with an optional second line of text and
+ * an optional interactable icon at the end.
+ *
+ * @param label The primary text in the item.
+ * @param description An optional second line of text.
+ * @param onClick Optional lambda for handling clicks.
+ * @param url The tab's URL.
```
gabrielluong, line 83: I think the description of this can be better. Can you take a look at the Favicon @param for url?

gabrielluong (changes requested):
Good first pass on this. I didn't look too deeply to see if it fully matched the design. Just thinking about our naming, conventions and how we can make this fit with the Design System.

app/src/main/java/org/mozilla/fenix/compose/list/ListItem.kt
MozillaNoah: The function name is leftover from V1, so that's my bad for not updating it after giving the custom `@Preview` name. I think that's a fair assessment though. For this, I was testing how it looked using a custom preview name to compare apples to apples, but I agree that the function name should align with the component itself.

app/src/main/java/org/mozilla/fenix/compose/list/ListItem.kt
MozillaNoah: I think this makes sense.

app/src/main/java/org/mozilla/fenix/compose/list/ListItem.kt
MozillaNoah: I agree with this. Also didn't think how calling it "primary text" overlaps with the `PrimaryText' Composable

app/src/main/java/org/mozilla/fenix/compose/list/ListItem.kt
```diff
@@ -0,0 +1,342 @@
...
+ */
+@Composable
+fun FavIconListItem(
+    label: String,
+    description: String? = null,
+    onClick: (() -> Unit)? = null,
+    url: String,
+    @DrawableRes icon: Int? = null,
+    iconDescription: String? = null,
+    onIconClick: (() -> Unit)? = null,
```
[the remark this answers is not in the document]
MozillaNoah: I understand where you're coming from. The rigidity here is because the figma for list items has ONLY an Icon at the end currently, but it's not to say that requirement/spec will never change during the lifetime of the component.

app/src/main/java/org/mozilla/fenix/compose/list/ListItem.kt
MozillaNoah: `ListItem` doesn't actually have any padding on the parent `Row` Composable, only on the internal `Column` used for text.

MozillaNoah:
Most recent commit updated the kdocs and previews based on suggestions and removed unnecessary padding from right icons.

app/src/main/java/org/mozilla/fenix/compose/list/ListItem.kt
```diff
@@ -0,0 +1,338 @@
...
+        onClick = onClick,
+        beforeListAction = {
+            Spacer(modifier = Modifier.width(16.dp))
+
+            Favicon(
+                url = url,
+                size = ICON_SIZE,
+            )
+
+            Spacer(modifier = Modifier.width(16.dp))
```
gabrielluong, line 108: This might be a good instance where we should be adding a Modifier parameter to `Favicon`. and prefer to add a `Modifier.padding(horizontal = 16.dp)` over `Spacer`. Also, see https://chris.banes.dev/always-provide-a-modifier/

gabrielluong (approved):

app/src/main/java/org/mozilla/fenix/compose/list/ListItem.kt
```diff
@@ -0,0 +1,338 @@
...
+            false -> Modifier
+        }.then(
+            Modifier
+                .padding(horizontal = 16.dp)
+                .defaultMinSize(minHeight = LIST_ITEM_HEIGHT),
+        ),
+        contentAlignment = Alignment.Center,
+    ) {
+        Icon(
+            painter = painterResource(icon),
```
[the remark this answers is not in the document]
MozillaNoah: I'll take a swing at this

app/src/main/java/org/mozilla/fenix/compose/list/ListItem.kt
MozillaNoah: I did find it odd that `Favicon` wasn't written to have a `Modifier` as a param. I'll make a follow-up story for this.

app/src/main/java/org/mozilla/fenix/compose/list/ListItem.kt
```diff
@@ -0,0 +1,338 @@
...
+        }.then(
+            Modifier.defaultMinSize(minHeight = LIST_ITEM_HEIGHT)
+        ),
+        verticalAlignment = Alignment.CenterVertically,
+    ) {
+        beforeListAction()
+
+        Column(
+            modifier = Modifier
+                .padding(horizontal = 16.dp, vertical = 10.dp)
```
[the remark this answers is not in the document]
gabrielluong: `10.dp` seems a bit arbitrary for such a scenario. I would even prefer to just specify `6.dp` since that was the minimum padding for the label + description scenario.

app/src/main/java/org/mozilla/fenix/compose/list/ListItem.kt
```diff
@@ -0,0 +1,338 @@
...
+            PrimaryText(
+                text = label,
+                fontSize = 16.sp,
+                maxLines = 1,
+            )
+
+            description?.let {
+                SecondaryText(
+                    text = description,
+                    fontSize = 12.sp,
```
gabrielluong, line 212: Just rechecked this based on padding comment since I was wondering if the label + description was actually getting 10.dp padding vertically since it looked a bit big compare to 6.dp. This should be 14.sp

app/src/main/java/org/mozilla/fenix/compose/list/ListItem.kt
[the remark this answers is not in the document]
MozillaNoah: That sounds good to me!

app/src/main/java/org/mozilla/fenix/compose/list/ListItem.kt
MozillaNoah: Thanks for catching this

MozillaNoah:
Most recent push updates the drawable res params to `Painter`s, renames all of their associated params to `endIconPainter`, `endIconDescription`, `endIconOnClick`, `startIconPainter`, etc, updated the description text size to 14, and changed the fallback vertical padding for the text to 6.dp

MozillaNoah:
Created https://github.com/mozilla-mobile/fenix/issues/24873 for the `Favicon` `Modifier`

app/src/main/java/org/mozilla/fenix/compose/list/ListItem.kt
```diff
@@ -0,0 +1,338 @@
...
+/**
+ * List item used to display a label with an optional description text and
+ * an optional, interactable icon at the end.
+ *
+ * @param label The label in the list item.
+ * @param description An optional description text below the label.
+ * @param onClick Called when the user clicks on the item.
+ * @param endIconPainter [Painter] used to display a [ListItemIcon] at the end of the list item.
+ * @param endIconDescription Content description of the end icon.
+ * @param onEndIconClick Called when the user clicks on the end icon.
```
gabrielluong, line 47: I think we should just drop these `end` prefix. I think it's generally understood based on the top level kDoc that the icon goes to the end.

app/src/main/java/org/mozilla/fenix/compose/list/ListItem.kt
```diff
@@ -0,0 +1,338 @@
...
+@Composable
+fun IconListItem(
+    label: String,
+    description: String? = null,
+    onClick: (() -> Unit)? = null,
+    startIconPainter: Painter,
+    startIconDescription: String? = null,
+    endIconPainter: Painter? = null,
+    endIconDescription: String? = null,
+    onEndIconClick: (() -> Unit)? = null,
```
gabrielluong, line 144: We are kinda mixing start/end with before/after here. I think we should just start with before/after.

app/src/main/java/org/mozilla/fenix/compose/list/ListItem.kt
MozillaNoah: Agreed. Updated

app/src/main/java/org/mozilla/fenix/compose/list/ListItem.kt
MozillaNoah: Agreed. Updated

app/src/main/java/org/mozilla/fenix/compose/list/ListItem.kt
```diff
@@ -0,0 +1,338 @@
...
+ * @param onAfterIconClick Called when the user clicks on the icon.
+ */
+@Composable
+fun TextListItem(
+    label: String,
+    description: String? = null,
+    onClick: (() -> Unit)? = null,
+    afterIconPainter: Painter? = null,
+    afterIconDescription: String? = null,
+    onAfterIconClick: (() -> Unit)? = null,
```
gabrielluong, line 56: I am generally in favor of not having any prefix ("after") for TextListItem and FaviconListItem. Keep it simple in these cases where we don't need to decide where the icon is going.

MozillaNoah: I think this makes sense, since the kdoc will indicate where the icon will live in these list items. Updated with those suggestions.

PR #24574, opened 2022-04-04
Fix and re-enable jumpBackInOptionTest and recentBookmarksOptionTest UI tests

AndiAJ:
• Fix and re-enable `jumpBackInOptionTest` and `recentBookmarksOptionTest` ✅  successfully ran 50x on Firebase 
• Ignore and remove `recently_saved_show_all` unused resource that [causes lint to fail](https://github.com/mozilla-mobile/fenix/runs/5817347446?check_suite_focus=true#step:4:1375) (@Mugurell please review and share your thoughts ☺️ )

Just to be safe, I'll add the `pr:do-not-land` label because i'll need to change `moz:removedIn` to 101 in case the PR doesn't get merged until the version bump scheduled for today. 

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -41,7 +41,7 @@
     <!-- Content description for the recently saved bookmarks section on the home screen. -->
     <string moz:removedIn="100" name="recently_saved_bookmarks_content_description" tools:ignore="UnusedResources">Recently saved bookmarks</string>
     <!-- Title for the button which navigates the user to show all of their saved bookmarks. -->
-    <string name="recently_saved_show_all">Show all</string>
+    <string name="recently_saved_show_all" moz:removedIn="101" tools:ignore="UnusedResources">Show all</string>
     <!-- Content description for the button which navigates the user to show all of their saved bookmarks. -->
     <string name="recently_saved_show_all_content_description_2">Show all saved bookmarks</string>
     <!-- Text for the menu button to remove a recently saved bookmark from the user's home screen -->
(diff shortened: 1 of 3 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -41,7 +41,7 @@
     <!-- Content description for the recently saved bookmarks section on the home screen. -->
     <string moz:removedIn="100" name="recently_saved_bookmarks_content_description" tools:ignore="UnusedResources">Recently saved bookmarks</string>
     <!-- Title for the button which navigates the user to show all of their saved bookmarks. -->
-    <string name="recently_saved_show_all">Show all</string>
+    <string moz:removedIn="100" name="recently_saved_show_all" tools:ignore="UnusedResources">Show all</string>
```
Mugurell, line 44: This respects the flow described in https://github.com/mozilla-mobile/fenix/wiki/Removing-strings.
Nice! 🏅

Mugurell: Please make sure the argument for `moz:removedIn`, in this case "100" is in sync with the nightly version at the time of the merge.  If we'll have moved to "101" this will have to be updated to reflect that and in so prevent the string from actually being removed earlier than needed.

sv-ohorvath:
LGTM

app/src/main/res/values/strings.xml
gabrielluong: Can we also reorder this 
```suggestion
    <string name="recently_saved_show_all" moz:removedIn="100" tools:ignore="UnusedResources">Show all</string>
```

AndiAJ: Sure, consider it done!

AndiAJ:
As per @Mugurell 's suggestion changed the removal of the unused resource to 101

PR #24655, opened 2022-04-06
Add Profiler button to profile without ADB

MarcLeclair:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/perf/ProfilerReusableComposable.kt
+++ b/app/src/main/java/org/mozilla/fenix/perf/ProfilerReusableComposable.kt
@@ -1,0 +1,101 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.perf
+
+import androidx.annotation.StringRes
+import androidx.compose.foundation.clickable
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.padding
+
+import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.material.Card
+import androidx.compose.material.CircularProgressIndicator
+import androidx.compose.material.RadioButton
+import androidx.compose.material.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.MutableState
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+
+/**
+ * Dialogue top level card for the profiler
+ */
+@Composable
+fun ProfilerDialogueCard(content: @Composable () -> Unit) {
+    Card(
+        elevation = 8.dp,
+        shape = RoundedCornerShape(12.dp)
+    ) {
+        content()
+    }
+}
+
+/**
+ * Top level radio button for the profiler dialogue
+ */
+@Composable
+fun ProfilerLabeledRadioButton(
+    text: String,
+    subText: String,
+    state: MutableState<String>
+) {
+    Row {
+        RadioButton(
+            selected = state.value == text,
+            onClick = { state.value = text },
+            enabled = true,
+        )
+        Column {
+            Text(
+                text = text,
+                modifier = Modifier
+                    .padding(start = 8.dp)
+                    .clickable {
+                        state.value = text
+                    }
+            )
+            Text(
+                text = subText,
+                fontWeight = FontWeight.ExtraLight,
+                modifier = Modifier
+                    .padding(start = 8.dp)
+                    .clickable { state.value = text }
+            )
+        }
+    }
+}
+
+/**
+ * Profiler Dialogue to display circular spinner when waiting
+ */
+@Composable
+fun WaitForProfilerDialog(
+    @StringRes message: Int
+) {
+    ProfilerDialogueCard {
+        Column(
+            modifier = Modifier.padding(8.dp),
+            verticalArrangement = Arrangement.Center,
+            horizontalAlignment = Alignment.CenterHorizontally
+        ) {
+            Text(
+                text = stringResource(message),
+                fontWeight = FontWeight.Bold,
+                fontSize = 15.sp,
+                modifier = Modifier.padding(8.dp)
+            )
+            Spacer(modifier = Modifier.height(2.dp))
+            CircularProgressIndicator()
+        }
+    }
+}
--- a/app/src/main/java/org/mozilla/fenix/perf/ProfilerStopDialogFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/perf/ProfilerStopDialogFragment.kt
@@ -1,0 +1,174 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.perf
+
+import android.os.Bundle
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import android.widget.Toast
+import androidx.annotation.StringRes
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.width
+import androidx.compose.material.Text
+import androidx.compose.material.TextButton
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.MutableState
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.ComposeView
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+import androidx.compose.ui.window.Dialog
+import androidx.fragment.app.DialogFragment
+import androidx.fragment.app.activityViewModels
+import org.mozilla.fenix.R
+import org.mozilla.fenix.ext.components
+import org.mozilla.fenix.perf.ProfilerUtils.handleProfileSave
+
+/**
+ * Dialogue to stop the Gecko profiler without using ADB.
+ */
+class ProfilerStopDialogFragment : DialogFragment() {
+
+    private val profilerViewModel: ProfilerViewModel by activityViewModels()
+
+    override fun onCreateView(
+        inflater: LayoutInflater,
+        container: ViewGroup?,
+        savedInstanceState: Bundle?
+    ): View {
+        return ComposeView(requireContext()).apply {
+            setContent {
+                StopProfilerCard()
+            }
+        }
+    }
+
+    override fun dismiss() {
+        profilerViewModel.setProfilerState(requireContext().components.core.engine.profiler!!.isProfilerActive())
+        super.dismiss()
+    }
+
+    @Composable
+    private fun StopProfilerCard() {
+
+        val viewStateObserver = remember { mutableStateOf(CardState.UrlWarningState) }
+        Dialog(onDismissRequest = {
+            // In the waiting state, we do not want the users to be able to click away from the dialogue
+            // since the user needs to wait for the profiler data to be ready and we don't want to handle
+            // the process in the background.
+            if (viewStateObserver.value != CardState.WaitForProfilerState) {
+                profilerViewModel.setProfilerState(requireContext().components.core.engine.profiler!!.isProfilerActive())
+                this@ProfilerStopDialogFragment.dismiss()
+            }
+        }) {
+            when (viewStateObserver.value) {
+                CardState.UrlWarningState -> {
+                    UrlWarningCard(viewStateObserver)
+                }
+                CardState.WaitForProfilerState -> {
+                    WaitForProfilerDialog(R.string.profiler_gathering)
+                }
+            }
+        }
+    }
+
+    @Composable
+    private fun UrlWarningCard(
+        viewStateObserver: MutableState<CardState>,
+    ) {
+        ProfilerDialogueCard {
+            Column(modifier = Modifier.padding(8.dp)) {
+                Text(
+                    text = stringResource(R.string.profiler_url_warning),
+                    fontWeight = FontWeight.ExtraBold,
+                    fontSize = 20.sp,
+                    modifier = Modifier.padding(8.dp)
+                )
+                Spacer(modifier = Modifier.height(10.dp))
+                Text(
+                    text = stringResource(R.string.profiler_url_warning_explained),
+                    fontWeight = FontWeight.Medium,
+                    fontSize = 15.sp,
+                    modifier = Modifier.padding(8.dp)
+                )
+                Spacer(modifier = Modifier.height(30.dp))
+                Row(
+                    horizontalArrangement = Arrangement.End,
+                    modifier = Modifier.fillMaxWidth()
+                ) {
+                    TextButton(
+                        onClick = {
+                           displayToastAndDismiss(R.string.profile_stop_dialogue_cancel_save)
+                        }
+                    ) {
+                        Text(stringResource(R.string.profiler_start_cancel))
+                    }
+                    Spacer(modifier = Modifier.width(4.dp))
+
+                    TextButton(
+                        onClick = {
+                            viewStateObserver.value = CardState.WaitForProfilerState
+                            stopProfiler()
+                        }
+                    ) {
+                        Text(stringResource(R.string.profiler_as_url))
+                    }
+                }
+            }
+        }
+    }
+
+    private fun stopProfiler() {
+        requireContext().components.core.engine.profiler!!.stopProfiler(
+            onSuccess = {
+                if (it != null) {
+                    handleProfileSave(
+                        requireContext(),
+                        it,
+                        ::displayToastAndDismiss
+                    )
+                } else {
+                    displayToastAndDismiss(R.string.profiler_no_info)
+                }
+            },
+            onError = {
+                error ->
+                error.message?.let {
+                    displayToastAndDismiss(R.string.profiler_error, " error: $error")
+                }?.let {
+                    displayToastAndDismiss(R.string.profiler_error)
+                }
+            }
+        )
+    }
+
+    private fun displayToastAndDismiss(@StringRes message: Int, extra: String = "") {
+        Toast.makeText(
+            context,
+            resources.getString(message) + extra,
+            Toast.LENGTH_LONG
+        ).show()
+        dismiss()
+    }
+
+    /**
+     * State that represents which card to display within the Stop dialogue.
+     */
+    enum class CardState {
+        UrlWarningState,
+        WaitForProfilerState;
+    }
+}
--- a/app/src/main/java/org/mozilla/fenix/perf/ProfilerStopDialogFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/perf/ProfilerStopDialogFragment.kt
@@ -65,15 +65,20 @@ class ProfilerStopDialogFragment : DialogFragment() {
     private fun StopProfilerCard() {
 
         val viewStateObserver = remember { mutableStateOf(CardState.UrlWarningState) }
-        Dialog(onDismissRequest = {
-            // In the waiting state, we do not want the users to be able to click away from the dialogue
-            // since the user needs to wait for the profiler data to be ready and we don't want to handle
-            // the process in the background.
-            if (viewStateObserver.value != CardState.WaitForProfilerState) {
-                profilerViewModel.setProfilerState(requireContext().components.core.engine.profiler!!.isProfilerActive())
-                this@ProfilerStopDialogFragment.dismiss()
+        Dialog(
+            onDismissRequest = {
+                // In the waiting state, we do not want the users to be able to click away from the dialogue
+                // since the user needs to wait for the profiler data to be ready and we don't want to handle
+                // the process in the background.
+                if (viewStateObserver.value != CardState.WaitForProfilerState) {
+                    profilerViewModel.setProfilerState(
+                        requireContext()
+                            .components.core.engine.profiler!!.isProfilerActive()
+                    )
+                    this@ProfilerStopDialogFragment.dismiss()
+                }
             }
-        }) {
+        ) {
             when (viewStateObserver.value) {
                 CardState.UrlWarningState -> {
                     UrlWarningCard(viewStateObserver)
@@ -111,7 +116,7 @@ class ProfilerStopDialogFragment : DialogFragment() {
                 ) {
                     TextButton(
                         onClick = {
-                           displayToastAndDismiss(R.string.profile_stop_dialogue_cancel_save)
+                            displayToastAndDismiss(R.string.profile_stop_dialogue_cancel_save)
                         }
                     ) {
                         Text(stringResource(R.string.profiler_start_cancel))
--- a/app/src/main/java/org/mozilla/fenix/perf/ProfilerStopDialogFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/perf/ProfilerStopDialogFragment.kt
@@ -71,10 +71,6 @@ class ProfilerStopDialogFragment : DialogFragment() {
                 // since the user needs to wait for the profiler data to be ready and we don't want to handle
                 // the process in the background.
                 if (viewStateObserver.value != CardState.WaitForProfilerState) {
-                    profilerViewModel.setProfilerState(
-                        requireContext()
-                            .components.core.engine.profiler!!.isProfilerActive()
-                    )
                     this@ProfilerStopDialogFragment.dismiss()
                 }
             }
@@ -116,7 +112,8 @@ class ProfilerStopDialogFragment : DialogFragment() {
                 ) {
                     TextButton(
                         onClick = {
-                            displayToastAndDismiss(R.string.profile_stop_dialogue_cancel_save)
+                            requireContext().components.core.engine.profiler?.stopProfiler({}, {})
+                            dismiss()
                         }
                     ) {
                         Text(stringResource(R.string.profiler_start_cancel))
@@ -151,9 +148,9 @@ class ProfilerStopDialogFragment : DialogFragment() {
             },
             onError = {
                 error ->
-                error.message?.let {
+                if (error.message != null) {
                     displayToastAndDismiss(R.string.profiler_error, " error: $error")
-                }?.let {
+                } else {
                     displayToastAndDismiss(R.string.profiler_error)
                 }
             }
--- a/app/src/main/java/org/mozilla/fenix/perf/ProfilerUtils.kt
+++ b/app/src/main/java/org/mozilla/fenix/perf/ProfilerUtils.kt
@@ -14,14 +14,9 @@ import mozilla.components.concept.fetch.Response
 import org.json.JSONObject
 import org.mozilla.fenix.R
 import org.mozilla.fenix.ext.components
-import java.io.BufferedReader
-import java.io.ByteArrayInputStream
 import java.io.File
-import java.io.FileNotFoundException
 import java.io.FileOutputStream
 import java.io.IOException
-import java.io.InputStreamReader
-import java.util.zip.GZIPInputStream
 
 private const val PROFILER_API = "https://api.profiler.firefox.com/compressed-store"
 private const val PROFILER_SERVER_HEADER = "application/vnd.firefox-profiler+json;version=1.0"
@@ -80,7 +75,7 @@ enum class ProfilerSettings(val threads: Array<String>, val features: Array<Stri
 object ProfilerUtils {
 
     private fun saveProfileUrlToClipboardAndToast(profileResult: ByteArray, context: Context): String {
-        //The profile is saved to a temporary file since our fetch API takes a file or a string.
+        // The profile is saved to a temporary file since our fetch API takes a file or a string.
         // Converting the ByteArray to a String would hurt the encoding, which we need to preserve.
         val outputFile = createTemporaryFile(profileResult, context)
         val response = networkCallToProfilerServer(outputFile, context)
--- a/app/src/main/java/org/mozilla/fenix/perf/ProfilerUtils.kt
+++ b/app/src/main/java/org/mozilla/fenix/perf/ProfilerUtils.kt
@@ -1,0 +1,152 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.perf
+
+import android.content.ClipData
+import android.content.ClipboardManager
+import android.content.Context
+import android.util.Base64
+import mozilla.components.concept.fetch.MutableHeaders
+import mozilla.components.concept.fetch.Request
+import mozilla.components.concept.fetch.Response
+import org.json.JSONObject
+import org.mozilla.fenix.R
+import org.mozilla.fenix.ext.components
+import java.io.BufferedReader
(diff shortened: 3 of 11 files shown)


app/src/main/java/org/mozilla/fenix/perf/ProfilerDialogFragment.kt
```diff
@@ -0,0 +1,455 @@
...
+        }
+    }
+
+    override fun dismiss() {
+        super.dismiss()
+        findNavController().previousBackStackEntry?.savedStateHandle?.set("profilerActive",profiler.isProfilerActive())
+
+
+    }
+    @Composable
```
mcomella, line 130: issue: you raised a concern that this PR felt bloated. I think this file might feel bloated because this file has too many purposes: by definition from the Android framework, a Fragment's responsibility is to receive lifecycle events and hook that up to other components (for example, inflating the view, passing it state, and forwarding user input events to the model and back to the views). However, this file and class is including a second responsibliity: defining views.

suggestion: I think if you moved these `@Composable` lines into a separate file, this file would feel much more manageable (e.g. it goes from ~455 lines to ~131 lines).

app/src/main/java/org/mozilla/fenix/perf/ProfilerDialogFragment.kt
```diff
@@ -0,0 +1,455 @@
...
+                        filters, features ->
+                    context.components.core.engine.profiler!!.startProfiler(filters,features)
+                }
+            } else {
+                val profilerErrorDialog = remember { mutableStateOf(false) }
+                var errorMessage: String? = null
+                StopProfilerDialog{
+                    saveType ->
+                    context.components.core.engine.profiler!!.stopProfiler(
+                        onSuccess = {
```
mcomella, line 84: issue: it becomes difficult to "keep it all in my head" when I define a lot of functionality inline, like this callback.

suggestion: I think it'd be easier to conceptualize what's happening here if the lambda was defined in a helper function. Helper functions really help me because, just by giving the operation a name that I can refer back to, it frees me up to think about how that result is used and less about the internal logic.

app/src/main/java/org/mozilla/fenix/perf/ProfilerDialogFragment.kt
```diff
@@ -0,0 +1,455 @@
...
+                            }
+                            Spacer(modifier = Modifier.width(4.dp))
+                            TextButton(onClick = {
+
+                                openDialog.value = false
+                                val settingType = ProfilerSettings.valueOf(profilerSettingsObserver.value)
+                                onPositiveClick(settingType.threads!!, settingType.features!!)
+                                profilerActiveObserver.value = true
+
+                                Thread{
```
mcomella, line 200: thought: we have to make sure this click event is debounced so it doesn't get called twice and cause too many jobs to run.

mcomella (commented):
Some initial thoughts based on your request for feedback – I'll follow-up on Element

app/src/main/java/org/mozilla/fenix/perf/ProfilerUtils.kt
```diff
@@ -0,0 +1,163 @@
...
+    Graphics("Graphics", graphics_threads, graphics_features),
+    Media("Media", media_threads, media_features),
+    Networking("Firefox", networking_threads, networking_features);
+
+    var settingType: String? = settingType
+    var features: Array<String>? = features
+    var threads: Array<String>? = threads
+}
+
+fun saveProfileAsURL(profileResult: ByteArray, context: Context): String {
```
mcomella, line 59: nitpick: these shouldn't be defined at the top-level because it pollutes the global namespace. Can you put these in an `object`?

app/src/main/java/org/mozilla/fenix/perf/ProfilerUtils.kt
```diff
@@ -0,0 +1,163 @@
...
+import java.io.FileOutputStream
+import java.io.InputStreamReader
+import java.util.zip.GZIPInputStream
+
+private val profiler_api = "https://api.profiler.firefox.com/compressed-store"
+private val profiler_header = "application/vnd.firefox-profiler+json;version=1.0"
+private val token = "profileToken"
+private val profileDataURL = "https://profiler.firefox.com/public/"
+
+private val firefox_features = arrayOf("screenshots", "js", "leaf", "stackwalk", "cpu", "java", "processcpu", "ipcmessages", "java")
```
mcomella, line 34: question: will these get out-of-date with the settings that the `about:debugging` UI uses? I wonder if there's a way we can use those values to avoid duplication. Maybe in a follow-up.

app/src/main/java/org/mozilla/fenix/perf/ProfilerUtils.kt
```diff
@@ -0,0 +1,163 @@
...
+import org.mozilla.fenix.R
+import org.mozilla.fenix.ext.components
+import java.io.BufferedReader
+import java.io.ByteArrayInputStream
+import java.io.File
+import java.io.FileOutputStream
+import java.io.InputStreamReader
+import java.util.zip.GZIPInputStream
+
+private val profiler_api = "https://api.profiler.firefox.com/compressed-store"
```
mcomella, line 29: nitpick: -> camel case here and below (ktlint might catch this)

app/src/main/java/org/mozilla/fenix/perf/ProfilerStopDialogFragment.kt
```diff
@@ -0,0 +1,236 @@
...
+                fileOutputStream.write(profileArray)
+                fileOutputStream.close()
+                jsonFile.close()
+                Toast.makeText(context, "The profile was saved locally", Toast.LENGTH_LONG).show()
+                this@ProfilerStopDialogFragment.dismiss()
+            }
+        }
+    }
+
+    enum class CardState(val stateName: String) {
```
mcomella, line 231: todo: stateName is unused

app/src/main/java/org/mozilla/fenix/perf/ProfilerUtils.kt
```diff
@@ -0,0 +1,163 @@
...
+enum class ProfilerSettings(settingType: String, threads: Array<String>, features: Array<String>) {
+    Firefox("Firefox", firefox_threads, firefox_features),
+    Graphics("Graphics", graphics_threads, graphics_features),
+    Media("Media", media_threads, media_features),
+    Networking("Firefox", networking_threads, networking_features);
+
+    var settingType: String? = settingType
+    var features: Array<String>? = features
+    var threads: Array<String>? = threads
+}
```
mcomella, line 57:
```suggestion
enum class ProfilerSettings(val threads: Array<String>, val features: Array<String>) {
    Firefox(firefox_threads, firefox_features),
    Graphics(graphics_threads, graphics_features),
    Media(media_threads, media_features),
    Networking(networking_threads, networking_features);
}
```

app/src/main/java/org/mozilla/fenix/perf/ProfilerUtils.kt
```diff
@@ -0,0 +1,163 @@
...
+private val graphics_features = arrayOf("java", "ipcmessages")
+private val graphics_threads = arrayOf("GeckoMain", "Compositor", "Renderer", "SwComposite", "RenderBackend", "SceneBuilder", "WrWorker", "CanvasWorkers",)
+
+private val media_features = arrayOf("js", "leaf", "stackwalk", "cpu", "audiocallbacktracing", "ipcmessages", "processcpu", "java")
+private val media_threads = arrayOf("cubeb", "audio", "camera", "capture", "Compositor", "GeckoMain", "gmp", "graph", "grph", "InotifyEventThread", "IPDL Background", "media", "ModuleProcessThread", "PacerThread", "RemVidChild", "RenderBackend", "Renderer", "Socket Thread", "SwComposite", "webrtc")
+
+private val networking_features = arrayOf("screenshots", "js", "leaf", "stackwalk", "cpu", "java", "processcpu", "ipcmessages")
+private val networking_threads = arrayOf("Compositor", "DNS Resolver", "DOM Worker", "GeckoMain", "Renderer", "Socket Thread", "StreamTrans", "SwComposite", "TRR Background")
+
+var isProfilerActive: Boolean = false
```
mcomella, line 46: todo (blocking): unused: remove

app/src/main/java/org/mozilla/fenix/perf/ProfilerUtils.kt
```diff
@@ -0,0 +1,163 @@
...
+    Media("Media", media_threads, media_features),
+    Networking("Firefox", networking_threads, networking_features);
+
+    var settingType: String? = settingType
+    var features: Array<String>? = features
+    var threads: Array<String>? = threads
+}
+
+fun saveProfileAsURL(profileResult: ByteArray, context: Context): String {
+    val outputFile = createTemporaryFile(profileResult, context)
```
mcomella, line 60: todo: delete the temp file

app/src/main/java/org/mozilla/fenix/perf/ProfilerStopDialogFragment.kt
```diff
@@ -0,0 +1,236 @@
...
+        profileSaveLauncher = registerForActivityResult(
+            ActivityResultContracts.StartActivityForResult()
+        ) { result ->
+            if (result.resultCode == Activity.RESULT_OK) {
+                val uri = result.data?.data
+                val jsonFile = context?.contentResolver?.openFileDescriptor(uri!!, "w")
+                val fileOutputStream = FileOutputStream(jsonFile!!.fileDescriptor)
+                fileOutputStream.write(profileArray)
+                fileOutputStream.close()
+                jsonFile.close()
```
mcomella, line 224: This should be safer. In particular, it handles closing streams if an exception is thrown (hopefully I got the syntax right):

```suggestion
                context?.contentResolver?.openFileDescriptor(uri!!, "w").use { jsonFile ->
                    FileOutputStream(jsonFile!!.fileDescriptor).use { fileOutputStream ->
                        fileOutputStream.write(profileArray)
                    }
                }
```

app/src/main/java/org/mozilla/fenix/perf/ProfilerReusableComposable.kt
```diff
@@ -0,0 +1,87 @@
...
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.MutableState
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+
+@Composable
+fun DialogueCard(content: @Composable () -> Unit) {
```
mcomella, line 29: nitpick: I think other views might use `DialogueCard` as a name – maybe this should be `ProfilerDialogueCard`? Or maybe all of these functions should be namespaced by being included in an `object` because I think `LabeledRadioButton` has the same issue?

app/src/main/java/org/mozilla/fenix/perf/ProfilerReusableComposable.kt
```diff
@@ -0,0 +1,87 @@
...
+@Composable
+fun LabeledRadioButton(
+    text: String,
+    subText: String,
+    state: MutableState<String>
+) {
+    Row {
+        RadioButton(
+            selected = state.value == text,
+            onClick = { state.value = text },
```
mcomella, line 47: thought (non-blocking): a more robust way to define this would be to have the state represented by an enum. e.g.
```kotlin
fun <T> LabeledRadioButton(
  // ...
  state: MutableState<T>,
  stateWhenSelected: T
) {
  // ...
  RadioButton(
    selected = state.value == stateWhenSelected,
    onClick = { state.value = stateWhenSelected }
  // ...
}

// usage
LabelRadioButton(
  // ...
  threadsAndFeaturesState,
  stateWhenSelected = ProfilerSettings.Firefox
)
```

But we don't need to change this now and can save it for a future iteration of this code.

app/src/main/java/org/mozilla/fenix/perf/ProfilerUtils.kt
```diff
@@ -0,0 +1,163 @@
...
+    val response = networkCallToProfilerServer(outputFile, context)
```
mcomella, line 61: question: do we handle if this network request fails?

app/src/main/java/org/mozilla/fenix/perf/ProfilerUtils.kt
```diff
@@ -0,0 +1,163 @@
...
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            putExtra(DocumentsContract.EXTRA_INITIAL_URI, "profiler.json")
+        }
+    }
+    profileSaveLauncher.launch(intent)
+}
+
+fun createTemporaryFile(profileResult: ByteArray, context: Context): File {
+    val gzipInputStream = GZIPInputStream(ByteArrayInputStream(profileResult))
```
mcomella, line 87: todo: use `.use` on these so they get closed.

app/src/main/java/org/mozilla/fenix/perf/ProfilerUtils.kt
```diff
@@ -0,0 +1,163 @@
...
+
+    val outputDir = context.cacheDir
+    val outputFile = File.createTempFile("tempProfile", ".json", outputDir)
+
+    try {
+        val fileOutputStream =
+            FileOutputStream(outputFile)
+        fileOutputStream.write(profileResult)
+        fileOutputStream.close()
+    } catch (e: Exception) {
```
mcomella, line 105: todo: we should catch a more specific exception

app/src/main/java/org/mozilla/fenix/perf/ProfilerUtils.kt
```diff
@@ -0,0 +1,163 @@
...
+        android.util.Log.e("profiler", e.toString())
+    }
```
mcomella, line 107:
```suggestion
    try {
        FileOutputStream(outputFile).use { fileOutputStream ->
            fileOutputStream.write(profileResult)
        }
    } catch (e: Exception) {
        android.util.Log.e("profiler", e.toString())
    }
```

app/src/main/java/org/mozilla/fenix/perf/ProfilerUtils.kt
```diff
@@ -0,0 +1,163 @@
...
+}
+
+fun saveProfileAsURL(profileResult: ByteArray, context: Context): String {
+    val outputFile = createTemporaryFile(profileResult, context)
+    val response = networkCallToProfilerServer(outputFile, context)
+    val profileToken = decodeProfileToken(response)
+    return profileDataURL + profileToken
+}
+
+fun finishProfileSave(context: Context, url: String) {
```
mcomella, line 66: nitpick: -> `saveProfileUrlToClipboardAndToast`

.experimenter.yaml
```diff
@@ -33,8 +33,8 @@ messaging:
       type: string
```
mcomella, line 33: chore (blocking): can we squash the commits? You can include my later suggestions and don't need to preserve authorship if it's too much trouble.

mcomella (changes requested):
It was a large patch so I wasn't able to look too thoroughly. However, it's a developer feature so I think that's okay.

This looks like it should work correctly. My main concerns are anything with "blocking". In particular, I think a function is getting called in compose when it shouldn't, the strings (probably?) shouldn't be inlined, and there's some duplication of state we can remove.

That being said, I wasn't able to test this for bugs because it didn't compile locally.

app/src/main/java/org/mozilla/fenix/settings/SettingsFragment.kt
```diff
@@ -189,6 +190,18 @@ class SettingsFragment : PreferenceFragmentCompat() {
             scrollToPreference(args.preferenceToScrollTo)
         }
 
+        isProfilerActive = requireContext().components.core.engine.profiler?.isProfilerActive()
+        if (isProfilerActive!!) {
+            findPreference<Preference>(getPreferenceKey(R.string.pref_key_start_profiler))?.title =
```
mcomella, line 195: question: is there a way to refresh this when the dialog closes? I was under the impression `onResume` would take care of that but it doesn't seem to be the case.

We could also simplify this and have a single "Start/stop Profiler` button rather than changing it dynamically.

app/src/main/java/org/mozilla/fenix/perf/ProfilerStopDialogFragment.kt
```diff
@@ -0,0 +1,265 @@
...
+        Dialog(onDismissRequest = { this@ProfilerStopDialogFragment.dismiss() }) {
+            when (viewStateObserver.value) {
+                CardState.StopDialogState -> {
+                    StopCard(viewStateObserver, saveProfileTypeObserver)
+                }
+                CardState.UrlWarningState -> {
+                    UrlWarningCard(viewStateObserver, saveProfileTypeObserver)
+                }
+                CardState.WaitForProfilerState -> {
+                    WaitForProfilerDialog("Gathering the profile")
```
mcomella, line 79: issue (blocking): when the profiler is saving, I can click outside of the waiting dialog to dismiss it. Eventually, we get a callback that access `context` but the context is dead so the app crashes.

suggestion: don't allow the stop profiler waiting dialog to be dismissed (e.g. I also haven't tested the back button's behavior which might have the same effect)

app/src/main/java/org/mozilla/fenix/settings/SettingsFragment.kt
```diff
@@ -189,6 +190,18 @@ class SettingsFragment : PreferenceFragmentCompat() {
             scrollToPreference(args.preferenceToScrollTo)
         }
 
+        isProfilerActive = requireContext().components.core.engine.profiler?.isProfilerActive()
```
mcomella, line 193: issue (blocking): when the profiler is started, sometimes the profiler menu item doesn't appear for me. STR:
1. Open settings, click to open secret debug settings
2. Start profiler
3. Back out to homescreen
4. Open settings

### EXpected
Profiler menu item is there

### Actual
No profiler menu item. I need to click into another settings item for it to appear. I'm not sure why this is happening.

app/src/main/java/org/mozilla/fenix/perf/ProfilerUtils.kt
```diff
@@ -0,0 +1,194 @@
...
+
+    private fun finishProfileSave(context: Context, url: String, onUrlFinish: (String) -> Unit) {
+        val clipboardManager = context.getSystemService(Context.CLIPBOARD_SERVICE) as ClipboardManager
+        val clipData = ClipData.newPlainText("Profiler URL", url)
+        clipboardManager.setPrimaryClip(clipData)
+        onUrlFinish("URL saved to clipboard successfully. $url")
+    }
+
+    private fun saveProfileAsFile(profileSaveLauncher: ActivityResultLauncher<Intent>) {
+        val intent = Intent(Intent.ACTION_CREATE_DOCUMENT).apply {
```
mcomella, line 101: question (blocking): saving a profile to a file in shared storage is also problematic because other apps on the device can read the data. Do you know how much of a privacy concern this is?

app/src/main/java/org/mozilla/fenix/perf/ProfilerUtils.kt
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    private fun finishProfileSave(context: Context, url: String, onUrlFinish: (String) -> Unit) {
+        val clipboardManager = context.getSystemService(Context.CLIPBOARD_SERVICE) as ClipboardManager
+        val clipData = ClipData.newPlainText("Profiler URL", url)
+        clipboardManager.setPrimaryClip(clipData)
+        onUrlFinish("URL saved to clipboard successfully. $url")
+    }
+
+    private fun saveProfileAsFile(profileSaveLauncher: ActivityResultLauncher<Intent>) {
```
mcomella, line 100: issue (blocking): if I stop the profiler, click save as file, then hit the back button before saving, I see the "Gathering the profile" dialog – I would expect for the profiler to continue running or for it to stop but not to see this dialog.

app/src/main/java/org/mozilla/fenix/perf/ProfilerStopDialogFragment.kt
```diff
@@ -0,0 +1,265 @@
...
+                Text(
+                    text = "Saving to URL warning",
+                    fontWeight = FontWeight.ExtraBold,
+                    fontSize = 20.sp,
+                    modifier = Modifier.padding(8.dp)
+                )
+                Spacer(modifier = Modifier.height(10.dp))
+                Text(
+                    text = "Saving to URL could potentially show private information such as " +
+                        "urls and screenshots. If you want to keep these" +
```
mcomella, line 168: todo (blocking): you need a space after "these"

mcomella (changes requested):
I'm concerned that 1) we're not updating the UI state often enough and 2) that the edge and error cases don't seem to be fully handled. Perhaps we can work together to reduce the scope (e.g. having a "Start/Stop Profiler" menu item instead of 2 separate ones) so that there are fewer cases to handle and we can get this landed.

app/src/main/java/org/mozilla/fenix/perf/ProfilerUtils.kt
```diff
@@ -0,0 +1,194 @@
...
+
+    @Throws(IOException::class)
+    private fun networkCallToProfilerServer(outputFile: File, context: Context): Response {
+        val request = Request(
+            url = PROFILER_API,
+            method = Request.Method.POST,
+            headers = MutableHeaders(
+                "Accept" to PROFILER_TOKEN
+            ),
+            body = Request.Body.fromFile(outputFile)
```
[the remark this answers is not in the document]
MarcLeclair: This one is simple, mostly because the api wouldn't take the string itself. I would get bad GZIP headers error

app/src/main/java/org/mozilla/fenix/perf/ProfilerUtils.kt
```diff
@@ -0,0 +1,194 @@
...
+        return context.components.core.client.fetch(request)
+    }
+
+    private fun decodeProfileToken(response: Response): String {
+        val jwtToken = StringBuilder("")
+        response.body.useStream { payload ->
+            payload.bufferedReader().use {
+                val jwt = it.readText()
+                val jwtSplit = jwt.split(".")
+                val decodedBytes = Base64.decode(jwtSplit[1], Base64.DEFAULT)
```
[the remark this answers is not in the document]
MarcLeclair: Yeah we can. This is taken directly from this [payload script](https://github.com/firefox-devtools/profiler-server/blob/master/tools/decode_jwt_payload.py#L47) where they break the array into the 3 part they need with the split (instead of targetting with `[1]`)

app/src/main/java/org/mozilla/fenix/perf/ProfilerUtils.kt
```diff
@@ -0,0 +1,194 @@
...
+    ) {
+        if (saveType == context.resources.getString(R.string.profiler_save_method_url)) {
+            try {
+                val url = saveProfileUrlToClipboardAndToast(profile, context)
+                finishProfileSave(context, url, onUrlFinish)
+            } catch (e: Exception) {
+                when (e) {
+                    is IOException -> onUrlFinish("Something went wrong contacting the Profiler server.")
+                    is FileNotFoundException -> onUrlFinish("Something went wrong writing the profile data to the file")
+                    else -> onUrlFinish("Something went wrong saving the profile")
```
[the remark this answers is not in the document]
MarcLeclair: Good point, I'll just exhaust the list instead from the Gzip and File api. I just wasn't sure about those errors 100% so I felt like a generic message might be more helpful

app/src/main/java/org/mozilla/fenix/settings/SettingsFragment.kt
MarcLeclair: No it doesn't refresh when it closes. the only way would be to use the `navBackStackEntry` as it was done before. I thought it would traverse the tree again, but no. I think we can just add a listener for `lifecycle.ON_RESUME` (which is weird that is being called but not `onResume`)

app/src/main/java/org/mozilla/fenix/perf/ProfilerStopDialogFragment.kt
MarcLeclair: Ah yes, both back button and tapping outside the window would dismiss it. Good point

app/src/main/java/org/mozilla/fenix/perf/ProfilerUtils.kt
MarcLeclair: I'm not aware. I thought of that but there was two issues. 1) If we save locally within the app files, the user can't just access it and send it through gmail / github / whatever other messaging system they use. 2) We could password save it, but then again, I'm not sure if that would be enough.

MarcLeclair:
@mcomella  I added a `viewModel` to be able to update the UI more seamlessly. Lmk if we should just remove that and keep it as `Profiler`. I felt it more indicative of what was happening to update the UI based on the state of the profiler.

app/src/main/res/values/static_strings.xml
```diff
@@ -60,4 +60,44 @@
     <!-- Secret debug info strings -->
     <string name="debug_info_region_home" translatable="false">Home region</string>
     <string name="debug_info_region_current" translatable="false">Current region</string>
+
+    <!-- Profiler settings -->
+    <string name="preferences_start_profiler">Start Profiler</string>
+    <string name="profiler_stop">Stop Profiler</string>
```
mcomella, line 66: It's hard to quickly intuit how these strings are used because of the lack of grouping. I think it'd help to add a newline here.

```suggestion
    <string name="profiler_stop">Stop Profiler</string>
    
```

app/src/main/res/values/static_strings.xml
```diff
@@ -60,4 +60,44 @@
...
+    <string name="profiler_start_dialog_started">Profiler started</string>
+    <string name="profiler_save_method">Would you like to save your profile as:</string>
+    <string name="profiler_save_method_url">URL</string>
+    <string name="profiler_save_method_url_explained">The profile will upload automatically with all the information gathered and provide you a link from profiler.firefox.com</string>
+    <string name="profiler_save_method_file">File</string>
+    <string name="profiler_save_method_file_explained">The file will be saved locally on your android device. To share it, go to profiler.firefox.com on your computer or mobile device</string>
+    <string name="profiler_start_cancel">Cancel</string>
+    <string name="profiler_gathering">Gathering the profile</string>
+    <string name="profiler_waiting_start">Waiting for Profiler to start</string>
+    <string name="profiler_url_warning">Saving to URL warning</string>
```
mcomella, line 86: I think this makes a more natural dialog title

```suggestion
    <string name="profiler_url_warning">Warning: Profile Upload</string>
```

app/src/main/res/values/static_strings.xml
```diff
@@ -60,4 +60,44 @@
...
+    <string name="profiler_url_warning_explained">"Saving to URL could potentially show private information such as " +
+                        "urls and screenshots. If you want to keep these" +
+                        "privates, save your profile as a file instead and head on over to profiler.firefox.com" +
+                        "where you can select what to share and not share."</string>
```
mcomella, line 90: This needs to be updated from the previous version. Also, I think it should be more formal

```suggestion
    <string name="profiler_url_warning_explained">Performance profiles may contain sensitive information such as websites visited and screenshots of them. If you continue, this recent profile will be uploaded publicly. Do you wish to continue?</string>
```

app/src/main/res/values/static_strings.xml
```diff
@@ -60,4 +60,44 @@
...
+    <string name="profiler_file_save">Save as File</string>
+    <string name="profiler_as_url">Get Profile URL</string>
```
mcomella, line 92: To better match my suggestion above.

```suggestion
    <string name="profiler_as_url">Upload profile</string>
```

app/src/main/java/org/mozilla/fenix/perf/ProfilerUtils.kt
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    /**
+     * This will either save the profile locally or send it as a URL to the Firefox profiler server
+     *
+     * @param context Activity context to get access to the profiler API through components.core...
+     * @param profile Data returned from GeckoView as a GZIP ByteArray
+     * @param onUrlFinish function passed in to display a toast with the relevant information once the profile is saved
+     */
+    fun handleProfileSave(
```
mcomella, line 135: thought: it feels kind of scary to me how easy it is to upload sensitive data with the profiler. Maybe we should check with Bas or someone else on the performance team that the UX of this change is acceptable.

mcomella (changes requested):
We're pretty close. I'm requesting to rereview to check after these changes that we're handling the sensitive temporary file safely.

app/src/main/java/org/mozilla/fenix/perf/ProfilerUtils.kt
mcomella: As discussed, we should also consider translating the strings so users outside EN locales understand what is happening.

mcomella:
Note: we pair programmed to clean up most of ^ so we should be pretty much good to land this on the next push (pending discussion with profiler team about UX).

app/src/main/res/values/static_strings.xml
```diff
@@ -60,4 +60,38 @@
...
+
+    <string name="profiler_start_cancel">Cancel</string>
+
+    <string name="profiler_gathering">Gathering the profile</string>
+    <string name="profiler_no_info">No information was gathered</string>
+
+    <string name="profiler_waiting_start">Waiting for Profiler to start</string>
+
+    <string name="profiler_url_warning">⚠️ Warning: Profile Upload</string>
+    <string name="profiler_url_warning_explained">Performance profiles may contain sensitive information such as websites visited and screenshots of them. If you continue, your profile will be uploaded and make it accessible to anyone with the link. Do you wish to continue?</string>
```
fqueze, line 89: Should this be "made accessible" instead of "make it accessible"?

app/src/main/java/org/mozilla/fenix/perf/ProfilerUtils.kt
```diff
@@ -0,0 +1,150 @@
...
+    private fun saveProfileUrlToClipboardAndToast(profileResult: ByteArray, context: Context): String {
+        // The profile is saved to a temporary file since our fetch API takes a file or a string.
+        // Converting the ByteArray to a String would hurt the encoding, which we need to preserve.
+        val outputFile = createTemporaryFile(profileResult, context)
+        try {
+            val response = networkCallToProfilerServer(outputFile, context)
+            val profileToken = decodeProfileToken(response)
+            return PROFILER_DATA_URL + profileToken
+        } finally {
+            outputFile.delete()
```
mcomella, line 86: todo (blocking): we should also add a comment mentioning this file contains sensitive information and it should be handled carefully.

app/src/main/java/org/mozilla/fenix/perf/ProfilerUtils.kt
```diff
@@ -0,0 +1,150 @@
...
+    Media(media_threads, media_features),
+    Networking(networking_threads, networking_features);
+}
+
+/**
+ * Collection of functionality to parse and save the profile returned by GeckoView.
+ */
+object ProfilerUtils {
+
+    private fun saveProfileUrlToClipboardAndToast(profileResult: ByteArray, context: Context): String {
```
mcomella, line 77: todo (blocking): this method no longer toasts so it shouldn't be named to suggest it does.

app/src/main/java/org/mozilla/fenix/perf/ProfilerViewModel.kt
```diff
@@ -0,0 +1,42 @@
...
+import androidx.lifecycle.MutableLiveData
+import androidx.lifecycle.ViewModel
+import org.mozilla.fenix.FenixApplication
+import org.mozilla.fenix.ext.components
+
+/**
+ * [ViewModel] to keep track of the profiler state
+ */
+class ProfilerViewModel(application: Application) : AndroidViewModel(application) {
+    var isProfilerActive: MutableLiveData<Boolean> = MutableLiveData(
```
mcomella, line 19: todo (blocking): this should be val, right? Possibly even private

app/src/main/res/values/static_strings.xml
```diff
@@ -60,4 +60,38 @@
...
+
+    <string name="profiler_waiting_start">Waiting for Profiler to start</string>
+
+    <string name="profiler_url_warning">⚠️ Warning: Profile Upload</string>
+    <string name="profiler_url_warning_explained">Performance profiles may contain sensitive information such as websites visited and screenshots of them. If you continue, your profile will be uploaded and made accessible to anyone with the link. Do you wish to continue?</string>
+    <string name="profiler_as_url">Upload profile</string>
+    <string name="profiler_error">Something went wrong with the profiler</string>
+    <string name="profiler_io_error">Something went wrong contacting the Profiler server.</string>
+
+    <string name="profiler_error_saving">Something went wrong saving the profile</string>
```
mcomella, line 94: chore (blocking): unused

app/src/main/res/values/static_strings.xml
```diff
@@ -60,4 +60,38 @@
...
+    <string name="profiler_uploaded_url_to_clipboard">URL copied to clipboard successfully</string>
+    <string name="profile_stop_dialogue_cancel_save">Profile save operation canceled</string>
```
mcomella, line 96: chore (blocking): unused

mcomella (approved):
lgtm w/ nits. btw, you'll need to rebase to get the CI to pass because it relies on your android-components pull request that landed after this PR was opened.

PR #24682, opened 2022-04-08
Migrate TabInCollectionViewHolder to Compose

Mugurell:
Draft for now. Still a bit of cleanup to do.

### Pull Request checklist

1. click on Show All Checks,


Mugurell:
Rebasing to fix conflicts.

gabrielluong:
Extracted out the WidgetSiteItem PR into https://github.com/mozilla-mobile/fenix/pull/24933

Mugurell:
Rebased on top of latest `main`.

app/src/main/java/org/mozilla/fenix/home/collections/TabInCollection.kt
```diff
@@ -0,0 +1,300 @@
...
+    onClick: () -> Unit,
+    onRemove: (Boolean) -> Unit,
+) {
+    val dismissState = rememberDismissState()
+
+    if (dismissState.isDismissed(StartToEnd) || dismissState.isDismissed(EndToStart)) {
+        onRemove(true)
+    }
+
+    SwipeToDismiss(
```
gabrielluong, line 87: I am curious if you used any particular resources/docs for this. Just want to be looking at the same resources while reviewing this.

Mugurell: There the [official documentation](https://developer.android.com/reference/kotlin/androidx/compose/material/package-summary#SwipeToDismiss(androidx.compose.material.DismissState,androidx.compose.ui.Modifier,kotlin.collections.Set,kotlin.Function1,kotlin.Function1,kotlin.Function1)) based on which I tried to obtain the exact same result as we have now. (The delete icons from the background which appear selectively based on the swipe direction needed special attention)

app/src/main/java/org/mozilla/fenix/home/collections/TabInCollection.kt
```diff
@@ -0,0 +1,300 @@
...
+                tint = FirefoxTheme.colors.iconWarning,
+            )
+        }
+    }
+}
+
+/**
+ * Clips the Composable this applies to such that it cannot draw content / shadows outside it's top bound.
+ */
+private fun Modifier.clipTop() = this.then(
```
MozillaNoah, line 186: do we foresee this being used elsewhere? If so, would it make sense to have this live in the `Modifier` extension file?

app/src/main/java/org/mozilla/fenix/home/collections/TabInCollection.kt
```diff
@@ -0,0 +1,300 @@
...
+                .fillMaxWidth()
+                .defaultMinSize(minHeight = 56.dp),
+            shape = if (isLastInCollection) bottomTabShape else middleTabShape,
+            backgroundColor = FirefoxTheme.colors.layer2,
+            elevation = 5.dp,
+        ) {
+            FaviconListItem(
+                label = tab.title,
+                description = shortenUrl(tab.url),
+                onClick = { onClick() },
```
MozillaNoah, line 122:
```suggestion
                onClick = onClick,
```

app/src/main/java/org/mozilla/fenix/home/collections/TabInCollection.kt
```diff
@@ -0,0 +1,300 @@
...
+                    // Happened once in testing when deleting a tab. Could not reproduce afterwards.
+                    Modifier.clipTop()
+                }
+            }
+        }
+
+        Card(
+            modifier = clippingModifier
+                .fillMaxWidth()
+                .defaultMinSize(minHeight = 56.dp),
```
MozillaNoah, line 114: `defaultMinSize` may not be necessary, since `FaviconListItem` has this already set.

app/src/main/java/org/mozilla/fenix/home/collections/TabInCollection.kt
```diff
@@ -0,0 +1,300 @@
...
+@Preview(uiMode = Configuration.UI_MODE_NIGHT_YES)
+private fun TabInCollectionDarkPreview() {
+    FirefoxTheme(theme = Theme.Dark) {
+        Column {
+            Box(modifier = Modifier.height(56.dp)) {
+                DismissedTabBackground(
+                    dismissDirection = StartToEnd,
+                    isLastInCollection = false,
+                )
+            }
```
MozillaNoah, line 227: nit: missing a couple new line between Compose blocks in the previews

app/src/main/java/org/mozilla/fenix/home/collections/TabInCollection.kt
```diff
@@ -0,0 +1,300 @@
...
+    return if (inComposePreview) {
+        url
+    } else {
+        url.toShortUrl(LocalContext.current.components.publicSuffixList)
+    }
+}
+
+@OptIn(ExperimentalMaterialApi::class)
+@Composable
+@Preview(uiMode = Configuration.UI_MODE_NIGHT_YES)
```
MozillaNoah, line 218:
```suggestion
@Composable
@Preview(uiMode = Configuration.UI_MODE_NIGHT_YES)
@Preview(uiMode = Configuration.UI_MODE_NIGHT_NO)
```

instead of doing two duplicate preview blocks for light/dark, it's possible to do two consecutive preview calls.

MozillaNoah (commented):
initial pass of comments

app/src/main/java/org/mozilla/fenix/home/collections/TabInCollection.kt
gabrielluong: We probably don't want to keep this around longer than we have to. The plan is to join all the collection composables under a single Card. See https://github.com/mozilla-mobile/fenix/pull/24501#discussion_r869410027

Mugurell: Thank you!
There are a number of bugs with the current homescreen collections (also filed recently [some](https://github.com/mozilla-mobile/fenix/issues?q=is%3Aopen+is%3Aissue+author%3AMugurell+label%3AFeature%3ACollections))  I'd say mostly stemming from the cumbersome data flow of this feature. It would need to be migrated to our MVI store and then one composable observing said store would have an easier time of syncing everything.
This patch only ment to do a 1-1 migration from the old View based system to Compose. Same bugs should still exist and work exactly as before 💪.
The above modifier was needed for a new situation in which it seems like two separate composables will show a bit of a shadow in between. May be a bug in Compose but it should also be resolved by having everything in just on ViewHolder. It does seems like a bit of a smell and I'd agree we don't want to keep it around any longer than is expressly needed.

app/src/main/java/org/mozilla/fenix/home/collections/TabInCollection.kt
Mugurell: Was intended on my part. Wanted to group in code what's shown in preview.

app/src/main/java/org/mozilla/fenix/home/collections/TabInCollection.kt
Mugurell: Didn't know that. Awesome tip, thank you!

Mugurell: Tried it now exactly as written here and I see two previews with both using light mode though even after "Build Refresh".
Could be needing a newer Android Studio?
Will remove one version nonetheless.

Mugurell:
Rebased and merged everything in one commit as gabrielluong requested.

app/src/main/java/org/mozilla/fenix/home/collections/TabInCollection.kt
MozillaNoah: Is the theme still being set to light in the `FirefoxTheme` initialization?

Mugurell: 😅 That was it.
Removing it I see errors in the preview rendering:
> java.lang.ClassCastException: class com.android.layoutlib.bridge.android.BridgeContext cannot be cast to class org.mozilla.fenix.FenixApplication (com.android.layoutlib.bridge.android.BridgeContext is in unnamed module of loader com.intellij.ide.plugins.cl.PluginClassLoader @195c4744; org.mozilla.fenix.FenixApplication is in unnamed module of loader org.jetbrains.android.uipreview.ModuleClassLoader @283fdca)   at org.mozilla.fenix.ext.ContextKt.getApplication(Context.kt:29)   at org.mozilla.fenix.ext.ContextKt.getComponents(Context.kt:35)   at org.mozilla.fenix.ext.ContextKt.settings(Context.kt:57)   at org.mozilla.fenix.theme.Theme$Companion.getTheme(FirefoxTheme.kt:40)

Checked the codebase, seems like this cool double preview of the themed composable isn't used anywhere else so maybe we can't use it just yet :-(.

MozillaNoah: It's because of a limitation in `FirefoxTheme` currently, so the initialization needs to be:

`FirefoxTheme(theme = Theme.getTheme(isPrivate = false)) {`

Mugurell: Worked! Thank you!

Mugurell: I think I remember that limitation being because of the compose version used? And I think now that can be avoided?

app/src/main/java/org/mozilla/fenix/home/collections/TabInCollection.kt
```diff
@@ -0,0 +1,262 @@
...
+
+/**
+ * Get a friendlier short url for [url].
+ *
+ * @param url Full url to be shortened.
+ *
+ * @see toShortUrl
+ */
+@Composable
+private fun shortenUrl(url: String): String {
```
gabrielluong, line 206: We landed an extension function for `TabSessionState` that does the same thing in https://github.com/mozilla-mobile/fenix/pull/25242. It's kinda unfortunate, but maybe we should've actually add a `String` extension function instead of `TabSessionState`

app/src/main/java/org/mozilla/fenix/home/collections/TabInCollection.kt
```diff
@@ -0,0 +1,262 @@
...
+ * @param dismissDirection [DismissDirection] of the tab being swiped depending on which this composable
+ * will also indicate the swipe direction by placing a warning icon at the start of the swipe gesture.
+ * If `null` the warning icon will be shown at both ends.
+ * @param isLastInCollection Whether the tab is to be shown between others or as the last one in collection.
+ */
+@OptIn(ExperimentalMaterialApi::class)
+@Composable
+private fun DismissedTabBackground(
+    dismissDirection: DismissDirection?,
+    isLastInCollection: Boolean,
```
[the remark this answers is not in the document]
gabrielluong: I also have similar feelings for `TabInCollection`

app/src/main/java/org/mozilla/fenix/home/collections/TabInCollectionViewHolder.kt
```diff
@@ -0,0 +1,91 @@
...
+
+    @Composable
+    override fun Content() {
+        val tabInfo = remember { mutableStateOf(tabData) }
+
+        tabInfo.value.tab?.let { validTab ->
+            tabInfo.value.collection?.let { validCollection ->
+                val horizontalPadding =
+                    composeView.resources.getDimensionPixelSize(R.dimen.home_item_horizontal_margin)
+                composeView.setPadding(horizontalPadding, 0, horizontalPadding, 0)
```
gabrielluong, line 45: Let's move this into the `init {}` since that's what we do everywhere.

app/src/main/java/org/mozilla/fenix/home/collections/TabInCollectionViewHolder.kt
```diff
@@ -0,0 +1,91 @@
...
+    viewLifecycleOwner: LifecycleOwner,
+    private val interactor: CollectionInteractor,
+) : ComposeViewHolder(composeView, viewLifecycleOwner) {
+    private var tabData = TabInfo()
+
+    @Composable
+    override fun Content() {
+        val tabInfo = remember { mutableStateOf(tabData) }
+
+        tabInfo.value.tab?.let { validTab ->
```
gabrielluong, line 41:
```suggestion
        tabInfo.value.tab?.let { tab ->
```

app/src/main/java/org/mozilla/fenix/home/collections/TabInCollectionViewHolder.kt
```diff
@@ -0,0 +1,91 @@
...
+            tabInfo.value.collection?.let { validCollection ->
```
gabrielluong, line 42:
```suggestion
            tabInfo.value.collection?.let { collection ->
```

app/src/main/java/org/mozilla/fenix/home/collections/TabInCollectionViewHolder.kt
```diff
@@ -0,0 +1,91 @@
...
+                val horizontalPadding =
+                    composeView.resources.getDimensionPixelSize(R.dimen.home_item_horizontal_margin)
+                composeView.setPadding(horizontalPadding, 0, horizontalPadding, 0)
+
+                TabInCollection(
+                    tab = validTab,
+                    isLastInCollection = tabInfo.value.isLastInCollection,
```
gabrielluong, line 49: Maybe we just determine the `shape` here?

app/src/main/java/org/mozilla/fenix/home/collections/TabInCollection.kt
```diff
@@ -0,0 +1,300 @@
...
+ * Display an individual [Tab] as part of a collection.
+ *
+ * @param tab [Tab] to display.
+ * @param isLastInCollection Whether the tab is to be shown between others or as the last one in collection.
+ * @param onClick Invoked when the user click on the tab.
+ * @param onRemove Invoked when the user removes the tab informing also if the tab was swiped to be removed.
+ */
+@OptIn(ExperimentalMaterialApi::class)
+@Composable
+fun TabInCollection(
```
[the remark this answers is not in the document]
Mugurell: Went with `CollectionItem`.

app/src/main/java/org/mozilla/fenix/home/collections/TabInCollection.kt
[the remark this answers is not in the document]
Mugurell: Seems like the decision is about who should control what shape the background looks like for a dismissed tab.

With all the other values related to size, colors and imagery controlled from this composable and no way for it to prevent clients passing a circle for example and breaking it seems like it is better for this composable to be in full control of the shape also and for it to be left to contain all rules on how to adapt a bit of state to show this particular piece of UI.
If others want to pass another shape probably the usecase is different and other properties (sizes, colors, imagery) have to be updated also.

app/src/main/java/org/mozilla/fenix/home/collections/TabInCollection.kt
Mugurell: Not sure I have to do anything about this now.

app/src/main/java/org/mozilla/fenix/home/collections/TabInCollectionViewHolder.kt
Mugurell: As above allowing to pass shapes leads to brittle composables that can easily be broken by clients passing a "wrong" shape.
I feel like this is a good indication that we shouldn't. Only pass indications / state on what needs to be shown and leave this Composable applying it's business rules on adapting that and in this ensuring consistency with the rest of "hardcoded" properties it uses.

Mugurell: If this exact piece of UX (represented by this Composable) will be needed in other places later maybe we can just "promote" it to the `compose` package and change the naming of the `isLastInCollection` parameter but otherwise if there's any need for other shapes then other properties will probably have to be changed also. 

(Still thinking about passing a circle or some kind of an irregular shape. Shape's middle name is "Danger".)

I guess most we could do is allowing to pass a "cornerRadius" value which we can confine in between safe values. If this will be needed later.

Mugurell:
Thank you for helping with this @gabrielluong !
I've rebased and addressed some of the above comments (marked now as resolved).
Open to do more changes if you feel strongly about them.

gabrielluong:
> Thank you for helping with this @gabrielluong ! I've rebased and addressed some of the above comments (marked now as resolved). Open to do more changes if you feel strongly about them.

I think it's fine to land as is

PR #24686, opened 2022-04-09
Sync with Fenix 99.

NF075:
Iceraven is not updated since end of january. Fenix 96
Is it possible to sync with Fenix 99 ?


NF075:
sync firefox 99.0.1

gabrielluong:
I think this was intended

Issue #24708, opened 2022-04-11
Remove Event.wrapper for dark theme selected meetrics.


mcarare:
Breakout from https://github.com/mozilla-mobile/fenix/issues/21292.

PR #24714, opened 2022-04-11
For #24708 - Remove Event.wrapper for DarkThemeSelected telemetry

Alexandru2909:
Removed `Event.wrapper` for `DarkThemeSelected` metric.

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/settings/CustomizationFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/CustomizationFragment.kt
@@ -12,11 +12,10 @@ import androidx.appcompat.app.AppCompatDelegate
 import androidx.preference.PreferenceFragmentCompat
 import androidx.preference.SwitchPreference
 import org.mozilla.fenix.FeatureFlags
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/mozilla/fenix/settings/CustomizationFragment.kt
```diff
@@ -85,9 +84,9 @@ class CustomizationFragment : PreferenceFragmentCompat() {
     private fun bindDarkTheme() {
         radioDarkTheme = requirePreference(R.string.pref_key_dark_theme)
         radioDarkTheme.onClickListener {
-            requireContext().components.analytics.metrics.track(
-                Event.DarkThemeSelected(
-                    Event.DarkThemeSelected.Source.SETTINGS
+            AppTheme.darkThemeSelected.record(
+                AppTheme.DarkThemeSelectedExtra(
+                    "settings"
```
mcarare, line 89: We should try to preserve all the values previously set for extra keys, including the letter case. This would mean here we should register "SETTINGS". This is also the value mentioned in the extra key description in `metrics.yaml` file.

PR #24624, opened 2022-04-05
For #24455 - Migrate NoCollectionsMessageViewHolder to Compose

sarah541:
Fixes #24455 

| Before      | After (Compose) |
| ----------- | ----------- |
Screen Shot 2022-04-26 at 3 36 36 PM|Screen Shot 2022-04-26 at 3 34 18 PM

Measurements for the After (Compose) version:
Screen Shot 2022-04-26 at 2 31 08 PM
The 16px and 14px are meant to be 16sp and 14sp to represent the font size.

### Pull Request checklist

- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features. In addition, it includes a screenshot of a successful [accessibility scan](https://play.google.com/store/apps/details?id=com.google.android.apps.accessibility.auditor&hl=en_US) to ensure no new defects are added to the product.

### To download an APK when reviewing a PR:
1. click on Show All Checks,
2. click Details next to "Taskcluster (pull_request)" after it appears and then finishes with a green checkmark,
3. click on the "Fenix - assemble" task, then click "Run Artifacts".
4. the APK links should be on the left side of the screen, named for each CPU architecture

PR #24735, opened 2022-04-12
 For #21897 - Convert inactive tabs to compose 

### LINKED ISSUE #21897: Move Tabs Tray to compose: Inactive tab list - see InactiveTabsAdapter and it's viewholders.
Inactive tab list - see InactiveTabsAdapter and it's viewholders.

From meta #21318

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-18291)

MozillaNoah:
Closes #21897 

### Pull Request checklist

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/components/appstate/AppState.kt
```diff
@@ -41,6 +41,7 @@ import org.mozilla.fenix.gleanplumb.MessagingState
  */
 data class AppState(
     val inactiveTabsExpanded: Boolean = false,
+    val inactiveTabsAutoCloseDialogShown: Boolean = false,
```
gabrielluong, line 44: I see there is also a `TabTrayStore`. Can you investigate why `inactiveTabsExpanded` was originally in the `HomeFragmentStore` and whether or not both of these **should** go into the `TabTrayStore`? I don't expect to see the refactoring of `inactiveTabsExpanded ` to be done in this PR.

gabrielluong: Ah seems like `inactiveTabsExpanded` was always in `AppState`

MozillaNoah: `inactiveTabsAutoCloseDialogShown` actually isn't used. It's leftover from something else. Sorry for the confusion.

app/src/main/java/org/mozilla/fenix/tabstray/inactivetabs/InactiveTabs.kt
```diff
@@ -0,0 +1,355 @@
...
+ * @param onTabCloseClick Lambda for handling clicks on a specific inactive tab's close button.
+ */
+@Composable
+@Suppress("LongParameterList")
+fun InactiveTabsList(
+    inactiveTabs: List<InactiveTabListItem>,
+    expanded: Boolean,
+    showAutoCloseDialog: Boolean,
+    onHeaderClick: () -> Unit,
+    onDeleteAllClick: () -> Unit,
```
MatthewTighe, line 75: Another potential improvement could be: `onDeleteAllClick: (List<String>) -> Unit`. During manual testing I broke the delete button by the way I hacked in an inactive tab, which could've been avoided if the underlying deletion implementation was more written in a more functional style. Likely out of scope for now though

MatthewTighe (approved):
LGTM! Might wanna let gl get another look

image

yay

app/src/main/java/org/mozilla/fenix/tabstray/browser/InactiveTabsAdapter.kt
```diff
@@ -4,149 +4,69 @@
...
-private typealias Adapter = ListAdapter<InactiveTabsAdapter.Item, InactiveTabViewHolder>
-
-/**
- * The [ListAdapter] for displaying the list of inactive tabs.
+ * The adapter for displaying the section of inactive tabs.
  *
- * @param context [Context] used for various platform interactions or accessing [Components]
- * @param browserTrayInteractor [BrowserTrayInteractor] handling tabs interactions in a tab tray.
- * @param featureName [String] representing the name of the feature displaying tabs. Used in telemetry reporting.
+ * @param viewLifecycleOwner [LifecycleOwner] life cycle owner for the view.
```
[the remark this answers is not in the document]
MozillaNoah: Agreed. I copy/pasted from another file without polishing it.

app/src/main/java/org/mozilla/fenix/tabstray/TrayPagerAdapter.kt
```diff
@@ -42,12 +47,30 @@ class TrayPagerAdapter(
...
     }
+
+    private val inactiveTabsInteractor by lazy {
+        DefaultInactiveTabsInteractor(
+            InactiveTabsController(
+                appStore = context.components.appStore,
+                settings = context.settings(),
+            )
+        )
+    }
```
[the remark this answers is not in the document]
MozillaNoah: Objectively, I agree where you're coming from (this smells code pasta-y haha). I think this just follows the interactor -> controller patter and looks bad here because `DefaultInactiveTabsInteractor` only has the one parameter.

app/src/main/java/org/mozilla/fenix/tabstray/inactivetabs/InactiveTabs.kt
```diff
@@ -0,0 +1,355 @@
...
+ * @param url The tab's URL.
+ * @param onClick Lambda for handling clicks on the inactive tab.
+ * @param onCloseClick Lambda for handling clicks on the inactive tab's close button.
+ */
+@Composable
+fun InactiveTabsItem(
+    tabTitleText: String,
+    url: String,
+    onClick: () -> Unit,
+    onCloseClick: () -> Unit
```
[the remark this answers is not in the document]
MozillaNoah: My bad. I'm trying to shoot for always having trailing commas, so missed it here. Thanks for finding this.

app/src/main/java/org/mozilla/fenix/tabstray/inactivetabs/InactiveTabs.kt
MozillaNoah: Could you elaborate more on how you hacked in an inactive tab?

MatthewTighe: Guess I should've saved the stash 😅 I think I just added a tab to `TabsTrayStore.state.inactiveTabs` in `TabsTrayFragment`. IIRC I had to manipulate `List<TabSessionState>.getInactiveTabs`. That at least got me a tab, but since it wasn't originally from `browserStore.state` the two got out of sync when I tried the delete all button.

I don't think there's any real risk of something like that happening in prod though, I just generally think relying on global state like in `TabsTrayController::handleDeleteAllInactiveTabs` can be a bit of an anti-pattern. Super out-of-scope for this ticket though

gabrielluong:
Will take a look at this in the morning if it's waiting on me.

gabrielluong:
Still catching up on reviews. The plan is to get to through this today, but since it's fairly big, it could slip to tomorrow.

MozillaNoah:
> Still catching up on reviews. The plan is to get to through this today, but since it's fairly big, it could slip to tomorrow.

No worries! I plan on integrating ListItem into this once it lands anyway

app/src/main/java/org/mozilla/fenix/tabstray/inactivetabs/InactiveTabs.kt
```diff
@@ -0,0 +1,317 @@
...
+                contentDescription = stringResource(R.string.inactive_tabs_delete_all),
+                modifier = Modifier.size(20.dp),
+                tint = FirefoxTheme.colors.iconPrimary,
+            )
+        }
+    }
+}
+
+/**
+ * Inactive Tabs list item Composable
```
gabrielluong, line 157:
```suggestion
 * Inactive Tabs list item.
```
Let's drop `Composable`. I think I have been trying to tell you to move away from the internal semantics for your kDocs and go with something more plain that anyone can understand.

gabrielluong (changes requested):
Just did a pass and gonna just request changes to make sure we re-align with the existing xml designs and experience. 

Incomplete list of things I am looking for based on my initial testing:
- The correct ripple effect when clicking on the header
- Double check the font sizes
- The list item should be a WidgetSiteItem

app/src/main/java/org/mozilla/fenix/tabstray/TrayPagerAdapter.kt
```diff
@@ -42,12 +47,30 @@ class TrayPagerAdapter(
      */
     private val normalAdapter by lazy {
         ConcatAdapter(
-            InactiveTabsAdapter(context, browserInteractor, interactor, INACTIVE_TABS_FEATURE_NAME, context.settings()),
+            InactiveTabsAdapter(
+                lifecycleOwner = lifecycleOwner,
+                tabsTrayStore = tabsTrayStore,
+                tabsTrayInteractor = tabsTrayInteractor,
+                inactiveTabsInteractor = inactiveTabsInteractor,
+                inactiveTabsAutoCloseDialogInteractor = inactiveTabsInteractor,
```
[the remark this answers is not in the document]
MozillaNoah: This is an artifact from the original implementation. I'm cool with simplifying during this refactor though.

MozillaNoah:
Most recent push has the suggested changes for simplifying the interactor parameters, removing the `InactiveTabsListItem` wrapper in favor of just using `FaviconListItem`, updating comments, and renaming the delete all button lambda. I also made all of the composables `private` except for `InactiveTabsList`.

MozillaNoah:
Rebased with main

app/src/main/java/org/mozilla/fenix/tabstray/browser/InactiveTabViewHolder.kt
```diff
@@ -5,165 +5,115 @@
...
+                    inactiveTabsInteractor.onCloseClicked()
+                    showAutoClosePrompt = !showAutoClosePrompt
+                },
+                onEnableAutoCloseClick = {
+                    inactiveTabsInteractor.onEnabledAutoCloseClicked()
+                    showAutoClosePrompt = !showAutoClosePrompt
+                    showConfirmationSnackbar()
+                },
+                onTabClick = {
+                    TabsTrayMetrics.openInactiveTab.add()
```
gabrielluong, line 93: For all the metrics here, can we move these into the controller function that is called through onTabSelected and can you add a test in the existing controller function?

app/src/main/java/org/mozilla/fenix/tabstray/inactivetabs/InactiveTabListItem.kt
```diff
@@ -0,0 +1,20 @@
...
+import mozilla.components.browser.state.state.TabSessionState
+
+/**
+ * A data class used to display inactive tabs in the UI
+ *
+ * @param tabTitleText The tab's display text.
+ * @param displayUrl The tab's URL.
+ * @param tab The underlying [TabSessionState] used for any lambda callbacks.
+ */
+data class InactiveTabListItem(
```
gabrielluong, line 16: Don't think this is really necessary since we can get all the information from `TabSessionState` instead of wrapping it inside another class.

app/src/main/java/org/mozilla/fenix/tabstray/inactivetabs/InactiveTabs.kt
```diff
@@ -0,0 +1,294 @@
...
+import org.mozilla.fenix.theme.Theme
+
+private val ROUNDED_CORNER_SHAPE = RoundedCornerShape(8.dp)
+
+/**
+ * Top-level Composable for displaying an expandable section of Inactive Tabs
+ *
+ * @param inactiveTabs List of [InactiveTabListItem] to display.
+ * @param expanded Boolean to determine whether to show the inactive tabs section expanded or collapsed.
+ * @param showAutoCloseDialog Boolean to determine whether to show the auto close inactive tabs dialog.
```
gabrielluong, line 59:
```suggestion
 * @param expanded Whether to show the inactive tabs section expanded or collapsed.
 * @param showAutoCloseDialog Whether to show the auto close inactive tabs dialog.
```

I think this aligns more closely with our current style.

app/src/main/java/org/mozilla/fenix/tabstray/inactivetabs/InactiveTabs.kt
```diff
@@ -0,0 +1,294 @@
...
+            .clip(ROUNDED_CORNER_SHAPE)
+            .background(FirefoxTheme.colors.layer2)
+            .border(
+                width = 2.dp,
+                color = FirefoxTheme.colors.borderPrimary,
+                shape = ROUNDED_CORNER_SHAPE,
+            )
+            .padding(vertical = 8.dp)
+    ) {
+        InactiveTabsHeader(
```
gabrielluong, line 93: The ripple effect I previously mention should still look like:
https://user-images.githubusercontent.com/1190888/168111528-0c235f15-8a26-4411-becc-91a560b7ac42.mov

So, I am expecting the whole "Column" or "Card" to be clickable in this case in order to get the same effect. You might want to look at the Collections composable, but I don't know how helpful that will be since that one is currently rendered view by view.

app/src/main/java/org/mozilla/fenix/tabstray/inactivetabs/InactiveTabs.kt
```diff
@@ -0,0 +1,294 @@
...
+            if (showAutoCloseDialog) {
+                InactiveTabsAutoClosePrompt(
+                    onDismissClick = onAutoCloseDismissClick,
+                    onEnableAutoCloseClick = onEnableAutoCloseClick,
+                )
+            }
+
+            Column {
+                inactiveTabs.forEach { tab ->
+                    FaviconListItem(
```
gabrielluong, line 109: Not currently sure what to make of this yet. The favicon seems a bit small when I compare it to History, but I am not quite sure if this is because it's not loading the browser icon and just using the placeholder that just displays a letter for the URL.  Might want to throw some screenshots for UX to approve of or double check the sizing between a couple of screens.

gabrielluong (changes requested):

MozillaNoah:
Rebased and pushed updates for:

- Removing InactiveTabsListItem and creating two extension functions for generating the display title and URL
- `showInactiveTabsAutoCloseDialog` verbiage update
- kdocs verbiage updates for interactors, stores, booleans, and Composable mentions
- Added a `Card` to wrap the UI since it looks cleaner without all of the `Modifier` params
- Refacted the Box + Icon to an IconButton
- Fixed-up the ripple (just needed to remove some padding that was no longer necessary)
- Refactored the analytics into a controller and added unit tests

app/src/main/java/org/mozilla/fenix/compose/ext/TabSessionState.kt
```diff
@@ -0,0 +1,23 @@
...
+import mozilla.components.browser.toolbar.MAX_URI_LENGTH
+import org.mozilla.fenix.components.components
+import org.mozilla.fenix.compose.inComposePreview
+import org.mozilla.fenix.ext.toShortUrl
+
+/**
+ * Indicates whether the app is currently running in a Composable @Preview.
+ */
+@Composable
+fun TabSessionState.toShortUrl(): String =
```
MozillaNoah, line 18: an extension file already exists for `TabSessionState`, but wasn't sure since this function is only relevant to Compose, if it should live in its own `compose/ext` directory file

app/src/main/java/org/mozilla/fenix/tabstray/browser/InactiveTabViewHolder.kt
```diff
@@ -5,165 +5,115 @@
...
+    override fun Content() {
+        val expanded = components.appStore
+            .observeAsComposableState { state -> state.inactiveTabsExpanded }.value ?: false
+        val inactiveTabs = tabsTrayStore
+            .observeAsComposableState { state -> state.inactiveTabs }.value ?: emptyList()
+        val listItems = inactiveTabs.map {
+            InactiveTabListItem(
+                tabTitleText = it.content.title.ifEmpty { it.content.url.take(MAX_URI_LENGTH) },
+                displayUrl = it.content.url.toShortUrl(components.publicSuffixList).take(MAX_URI_LENGTH),
+                tab = it,
```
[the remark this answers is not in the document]
Amejia481: I filed a PR for adding the extension function https://github.com/mozilla-mobile/fenix/pull/25242

MozillaNoah:
Needed to add a commit to comment out some stale UI test code (most was disabled in #24333) that relied upon an ID that no longer exists post-compose

gabrielluong:
One bug I have found so far is that if I manually make all my tabs inactive, then I don't see the inactive tabs at all. My testing so far has led me to conclude it is caused by your commit set.

MozillaNoah:
> Mostly been manually testing so far before looking at the code. One bug I have found so far is that if I manually make all my tabs inactive, then I don't see the inactive tabs at all. My testing so far has led me to conclude it is caused by your commit set. Also, the border around the card seems might thicker than the original.

@gabrielluong, I had an issue similar to that occur before. I'll take care of it. It's an issue of using compose + the concat adapter implementation. I'll also look into the border thickness.

MozillaNoah:
> Mostly been manually testing so far before looking at the code. One bug I have found so far is that if I manually make all my tabs inactive, then I don't see the inactive tabs at all. My testing so far has led me to conclude it is caused by your commit set. Also, the border around the card seems might thicker than the original.

@gabrielluong just pushed fixes for both. The border was originally 1dp instead of 2dp, so reverted that; good catch. The visibility bug required more of a short term hack until the rest of the TabsTray is converted to compose. See `NormalBrowserPageViewHolder.observeTabsTrayInactiveTabsState` if you're curious. I have a comment documenting it.

MozillaNoah:
Most recent push rebases with main and adds the new TextButton composable to the auto close dialog. I also added light mode to the previews.

app/src/main/java/org/mozilla/fenix/tabstray/inactivetabs/InactiveTabs.kt
```diff
@@ -0,0 +1,280 @@
...
+                    modifier = Modifier.weight(1f),
+                    fontSize = 14.sp,
+                    fontFamily = FontFamily(Font(R.font.metropolis_semibold)),
+                )
+
+                IconButton(
+                    onClick = onDismissClick
+                ) {
+                    Icon(
+                        painter = painterResource(R.drawable.mozac_ic_close),
```
gabrielluong, line 199: We have a `mozac_ic_close_20` that we will use in the MessageCard and CollectionsPlaceholder https://github.com/mozilla-mobile/fenix/pull/24624. See https://www.figma.com/file/ujdiReypSypu2GhRgzArea/branch/SoVHeGS5UR99B7qFX1c8eK/Android-Components?node-id=2124%3A12124

Since you are using `Modifier.size(20.dp)` here, it might make sense to just replace it with this smaller close button and remove the modifier. However, I will note that the ic_close_20 is still very much a different icon in terms of look from ic_close. It does make sense to use ic_close_20 here because we already use it for all the other message looking cards that I mentioned above. I would double check with Crystal.

app/src/main/java/org/mozilla/fenix/tabstray/inactivetabs/InactiveTabs.kt
```diff
@@ -0,0 +1,280 @@
...
+            }
+
+            SecondaryText(
+                text = stringResource(
+                    R.string.tab_tray_inactive_auto_close_body_2,
+                    stringResource(R.string.app_name)
+                ),
+                modifier = Modifier
+                    .fillMaxWidth()
+                    .padding(end = 12.dp),
```
gabrielluong, line 215: I think we can avoid this by making the `Column` have a horizontal padding of 12dp. To get the `IconButton` to be at the right edge, you can specify the size(20.dp) modifier on the IconButton (not the Icon). In my testing with the accessibility scanner, it didn't complain about the tap area.

app/src/main/java/org/mozilla/fenix/tabstray/inactivetabs/InactiveTabs.kt
gabrielluong: In general, I don't ever expect to see a size modifier for an `Icon` since the internal implementation will use the size specified by the drawable. However, I think you did use the size here because of the IconButton's increase tap area, but you should specify the size on the IconButton instead and that won't cause an accessibility issue in my testing. 

To be doubly sure, I would test your entire inactive tabs with the accessibility scanner anyways.

gabrielluong (approved):
r+ assuming all review feedback are addressed or discussed to satisfaction. I will probably monitor the compose changes, but beyond that I don't think I need to do another full review on this.

Side note, I would keep the delete files in the same commit where the compose replaces them. I still find myself undoing the xml deletion to visually compare the 2, which becomes harder to do when they are in separate commits. In general, I would say each individual commit should keep the build in a neutral state meaning that if we were to cherry-pick commits from this PR, we could still land them separately without breaking assuming we land them in order.

app/src/main/java/org/mozilla/fenix/tabstray/browser/InactiveTabViewHolder.kt
```diff
@@ -5,165 +5,95 @@
...
-
-        companion object {
-            const val LAYOUT_ID = R.layout.inactive_tabs_auto_close
+        if (inactiveTabs.isNotEmpty()) {
+            InactiveTabsList(
+                inactiveTabs = inactiveTabs,
+                expanded = expanded,
+                showAutoCloseDialog = showAutoClosePrompt,
+                onHeaderClick = { inactiveTabsInteractor.onHeaderClicked(!expanded) },
+                onDeleteAllButtonClick = tabsTrayInteractor::onDeleteInactiveTabs,
```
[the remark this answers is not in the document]
MozillaNoah: I agree in hindsight that when I implemented the "delete all" functionality last September that it should have just lived within `InactiveTabsInteractor`. Are you wanting this to be refactored now with this PR?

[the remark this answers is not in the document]
gabrielluong: Up to you. The r+ would carryover for this change in this PR since it's so minor that I don't believe it needs a re-review.

app/src/main/java/org/mozilla/fenix/tabstray/inactivetabs/InactiveTabs.kt
```diff
@@ -0,0 +1,280 @@
...
+                        painter = painterResource(R.drawable.mozac_ic_close),
+                        contentDescription =
+                        stringResource(R.string.tab_tray_inactive_auto_close_button_content_description),
+                        modifier = Modifier.size(20.dp),
+                        tint = FirefoxTheme.colors.iconPrimary,
+                    )
+                }
+            }
+
+            SecondaryText(
```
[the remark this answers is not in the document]
MozillaNoah: I intentionally haven't used any updated specs for the inactive tabs-specific content. This has honestly been in review so long that I don't remember why I used `SecondaryText`, but it could have been to mirror collections. I'm happy to switch it to `Primary` though

PR #24772, opened 2022-04-13
For #24303 ⁃ Refactor Onboarding into its own fragment

### LINKED ISSUE #24303: Refactor Onboarding into its own fragment
We would like to refactor the Onboarding first run content into their own fragment so that it will reduce the complexity of the HomeFragment and the number of things to convert to Jetpack Compose for the Homescreen performance work.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-19863)

gabrielluong:
Fixes #24303

🚧🚧🚧🚧 **This is a proof-of-concept - please review strictly for feasibility of whether or not we should continue with this pathway.**

This currently removes all of the onboarding items from the `HomeFragment` into their own `OnboardingFragment`. This allows us to decouple handling of whether or not the user has been onboarded state out of the `HomeFragment` and reduces one mode that needs to be managed in the `HomeFragment`. It also saves us a bit of time with the Compose refactoring of the Home screen with not having to refactor around 8 ViewHolders so this could be 2-4 weeks of time saving, but not a total dealbreaker if we decide to just convert them to Compose if this is not the solution we want to go with. Some future benefit might be that this will be an easier way to navigate back to Onboarding / Customization instead of making this first run only.

The problem I am seeing is that we have to duplicate a lot of code from the `HomeFragment` to implement the function of the Toolbar, Tab Counter, and Home Menu. There will probably be features missing from this current prototype such as whether or not we allow for toggling of the new wallpaper feature by tapping on the Firefox logo. Prior suggestions of base fragment to contain some of these shared code seems unfavourable based on experiences with the `BaseBrowserFragment`.

I would like to evaluate whether or not we should continue down this route or turn back and just convert the Onboarding to Compose for the time being and re-evaluate splitting Onboarding out in the future.

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/onboarding/OnboardingFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/onboarding/OnboardingFragment.kt
@@ -1,0 +1,191 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.onboarding
+
+import android.graphics.drawable.BitmapDrawable
+import android.os.Bundle
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import androidx.core.content.ContextCompat
+import androidx.fragment.app.Fragment
+import androidx.navigation.fragment.findNavController
+import kotlinx.coroutines.flow.collect
+import kotlinx.coroutines.flow.map
+import mozilla.components.browser.state.state.selectedOrDefaultSearchEngine
+import mozilla.components.browser.state.store.BrowserStore
+import mozilla.components.lib.state.ext.consumeFlow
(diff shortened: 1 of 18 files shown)


app/src/main/java/org/mozilla/fenix/onboarding/OnboardingFragment.kt
```diff
@@ -0,0 +1,305 @@
...
+        }
+    }
+
+    override fun onDestroyView() {
+        super.onDestroyView()
+
+        _binding = null
+    }
+
+    private fun observeSearchEngineChanges() {
```
csadilek, line 132: OK, I agree with your point that landing this as-is will cause too much code duplication, even in the short to mid-term. We could follow a similar pattern we use for browser fragment and use an abstract base class here to share the common functionality. Arguably the two versions of the home screen are different modifications based on the state so inheritance wouldn't be bad. Alternatively, we can move this functions to a separate type and make it re-usable that way.

Issue #24758, opened 2022-04-13
Remove Event.wrapper for BrowserSearch telemetry


mcarare:
Breakout from https://github.com/mozilla-mobile/fenix/issues/21292.

SoftVision-LorandJanos:
Verified as fixed on the latest Nightly 101.0a1 (2022-04-27).
Some events that were generated:
![image](https://user-images.githubusercontent.com/48829350/165523165-e4add441-cfe5-41f9-9c34-456c0000b664.png)
![image](https://user-images.githubusercontent.com/48829350/165523368-ef596fe1-1a1d-4117-974a-99c0ee8ebde7.png)
![image](https://user-images.githubusercontent.com/48829350/165523454-a2dc2a22-7975-4827-965c-4b09b52238f8.png)

Device used: Google Pixel 4 (Android 12).
Closing the ticket as fixed.

PR #24791, opened 2022-04-14
For #24758: Remove wrapper from browser search metrics.

mcarare:
### Pull Request checklist

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/components/metrics/MetricController.kt
```diff
@@ -247,6 +248,16 @@ internal class ReleaseMetricController(
             ProgressiveWebApp.installTap.record(NoExtras())
         }
 
+        Component.FEATURE_SEARCH to AdsTelemetry.SERP_ADD_CLICKED -> {
+            BrowserSearch.adClicks[value!!].add()
```
rocketsroger, line 252: I know original code uses `value!!` as well but I think we should change it to `value?.apply`?

mcarare: @rocketsroger I would keep this with the not null assertion because it is an important piece of telemetry we do not want to fail silently.

PR #24802, opened 2022-04-14
for #24177: move recent synced tab to secret debug menu

MatthewTighe:
1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabViewHolder.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabViewHolder.kt
@@ -15,7 +15,6 @@ import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.unit.dp
 import androidx.lifecycle.LifecycleOwner
 import mozilla.components.lib.state.ext.observeAsComposableState
-import org.mozilla.fenix.FeatureFlags
 import org.mozilla.fenix.R
 import org.mozilla.fenix.components.components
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabViewHolder.kt
```diff
@@ -69,7 +68,10 @@ class RecentTabViewHolder(
             )
 
             recentSyncedTabState.value?.let {
-                if (FeatureFlags.taskContinuityFeature && it != RecentSyncedTabState.None) {
+                if (
+                    components.settings.enableTaskContinuityEnhancements &&
```
MozillaNoah, line 72: should we keep the first part of this inline with the `if`?

MozillaNoah (approved):

app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabViewHolder.kt
MatthewTighe: For some reason I'd been under the impression that inlining would cause a lint error, but it apparently does not. I agree it looks better with that suggestion. Updated!

gabrielluong:
> test-debug
> [Unit Test Results](https://firefoxci.taskcluster-artifacts.net/GhC9VYPMQhebySLVo2B-jg/0/public/reports/test/testDebugUnitTest/index.html)
> (404 if compilation failed)

I think this is blocking the queue. Removing the needs-landing

gabrielluong:
Re-running CI.

Issue #24850, opened 2022-04-18
Display an address list for managing existing addresses


gabrielluong:
This adds a "Manage addresses" screen to display a list of addresses with a button to navigate to adding a new address.

<img width="206" alt="Screen Shot 2022-04-18 at 11 04 57 AM" src="https://user-images.githubusercontent.com/1190888/163828366-3b664bfd-0fb8-4876-879c-6fb7a3c10b1e.png">

https://www.figma.com/file/cSKRUEvfJSzNBv2Mp65Pdv/Mobile-CC-Autofill?node-id=1044%3A25739

PR #24863, opened 2022-04-18
For #24849 - Add an address editor screen

### LINKED ISSUE #24849: Add an address editor screen
This adds an "Add address" editor screen for adding a new address.

Screen Shot 2022-04-18 at 11 01 21 AM

gabrielluong:
Fixes #24849 

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/settings/address/AddressEditorFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/address/AddressEditorFragment.kt
@@ -1,0 +1,52 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.settings.address
+
+import android.os.Bundle
+import android.view.View
+import androidx.lifecycle.lifecycleScope
+import androidx.navigation.fragment.findNavController
+import org.mozilla.fenix.R
+import org.mozilla.fenix.SecureFragment
+import org.mozilla.fenix.databinding.FragmentAddressEditorBinding
+import org.mozilla.fenix.ext.components
+import org.mozilla.fenix.ext.showToolbar
+import org.mozilla.fenix.settings.address.controller.DefaultAddressEditorController
+import org.mozilla.fenix.settings.address.interactor.AddressEditorInteractor
+import org.mozilla.fenix.settings.address.interactor.DefaultAddressEditorInteractor
+import org.mozilla.fenix.settings.address.view.AddressEditorView
+
+/**
+ * Displays an address editor for adding and editing an address.
+ */
+class AddressEditorFragment : SecureFragment(R.layout.fragment_address_editor) {
+
+    private lateinit var addressEditorView: AddressEditorView
+    private lateinit var interactor: AddressEditorInteractor
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+
+        val storage = requireContext().components.core.autofillStorage
+
+        interactor = DefaultAddressEditorInteractor(
+            controller = DefaultAddressEditorController(
+                storage = storage,
+                lifecycleScope = lifecycleScope,
+                navController = findNavController()
+            )
+        )
+
+        val binding = FragmentAddressEditorBinding.bind(view)
+
+        addressEditorView = AddressEditorView(binding, interactor)
+        addressEditorView.bind()
+    }
+
+    override fun onResume() {
+        super.onResume()
+        showToolbar(getString(R.string.addresses_add_address))
+    }
+}
--- a/app/src/main/java/org/mozilla/fenix/settings/address/controller/AddressEditorController.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/address/controller/AddressEditorController.kt
@@ -1,0 +1,60 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.settings.address.controller
+
+import androidx.navigation.NavController
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+import mozilla.components.concept.storage.UpdatableAddressFields
+import mozilla.components.service.sync.autofill.AutofillCreditCardsAddressesStorage
+import org.mozilla.fenix.settings.address.AddressEditorFragment
+import org.mozilla.fenix.settings.address.interactor.AddressEditorInteractor
+
+/**
+ * [AddressEditorFragment] controller. An interface that handles the view manipulation of the
+ * credit card editor.
+ */
+interface AddressEditorController {
+
+    /**
+     * @see [AddressEditorInteractor.onCancelButtonClicked]
+     */
+    fun handleCancelButtonClicked()
+
+    /**
+     * @see [AddressEditorInteractor.onSaveAddress]
+     */
+    fun handleSaveAddress(addressFields: UpdatableAddressFields)
+}
+
+/**
+ * The default implementation of [AddressEditorController].
+ *
+ * @param storage An instance of the [AutofillCreditCardsAddressesStorage] for adding and retrieving
+ * addresses.
+ * @param lifecycleScope [CoroutineScope] scope to launch coroutines.
+ * @param navController [NavController] used for navigation.
+ */
+class DefaultAddressEditorController(
+    private val storage: AutofillCreditCardsAddressesStorage,
+    private val lifecycleScope: CoroutineScope,
+    private val navController: NavController,
+) : AddressEditorController {
+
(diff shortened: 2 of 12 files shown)


app/src/main/res/layout/fragment_address_editor.xml
```diff
@@ -0,0 +1,371 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<androidx.core.widget.NestedScrollView
```
gabrielluong, line 5: This mostly reuses the existing [fragment_credit_card_editor.xml](https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/res/layout/fragment_credit_card_editor.xml) to save us a bit of time and unblock work so that more people can work on this in parallel. I will file a follow up to Composify this in the future. I will note that this is not to spec with the proposed Figma design. When we were doing credit cards last year which is what the Address editor is also based on, we also deviated from the Font styles from the Figma spec since this is reusing the same styles that was used in Logins and Credit Cards. I will file a follow up if necessary to realign all our forms (Add search engine, Logins, Credit cards and Address) styles if UX determines that we want to adopt the new style for all of the forms, but otherwise, it seems okay to be consistent with the existing style everywhere for the time being.

app/src/main/java/org/mozilla/fenix/settings/address/view/AddressEditorView.kt
```diff
@@ -0,0 +1,53 @@
...
+            saveAddress()
+        }
+    }
+
+    internal fun saveAddress() {
+        binding.root.hideKeyboard()
+
+        interactor.onSaveAddress(
+            UpdatableAddressFields(
+                givenName = binding.fullNameInput.text.toString(),
```
gabrielluong, line 38: Expecting UX to change the design to split "Full Name" out to separate text input for First, Middle and Last name. So, this is just a placeholder for now.

app/src/main/java/org/mozilla/fenix/settings/address/AddressEditorFragment.kt
```diff
@@ -0,0 +1,60 @@
...
+                navController = findNavController()
+            )
+        )
+
+        val binding = FragmentAddressEditorBinding.bind(view)
+
+        addressEditorView = AddressEditorView(binding, interactor)
+        addressEditorView.bind()
+
+        binding.fullNameInput.apply {
```
MozillaNoah, line 49: any thoughts on having this inside `AddressEditorView`?

app/src/main/java/org/mozilla/fenix/settings/address/view/AddressEditorView.kt
```diff
@@ -0,0 +1,53 @@
...
+ */
+class AddressEditorView(
+    private val binding: FragmentAddressEditorBinding,
+    private val interactor: AddressEditorInteractor
+) {
+
+    /**
+     * Binds the view.
+     */
+    fun bind() {
```
MozillaNoah, line 23: I'm assuming there's a separate story for _editing_ an address and for displaying the current addresses?

app/src/main/java/org/mozilla/fenix/settings/autofill/AutofillSettingFragment.kt
```diff
@@ -173,6 +176,14 @@ class AutofillSettingFragment : BiometricPromptPreferenceFragment() {
...
                 getString(R.string.preferences_addresses_add_address)
         }
+
+        manageAddressesPreference.setOnPreferenceClickListener {
+            navController.navigate(
+                AutofillSettingFragmentDirections
+                    .actionAutofillSettingFragmentToAddressEditorFragment()
+            )
+            super.onPreferenceTreeClick(it)
+        }
```
MozillaNoah, line 186: should this have a conditional for when the user has addresses to take them to a different screen?

app/src/main/java/org/mozilla/fenix/settings/address/view/AddressEditorView.kt
gabrielluong: Ya.

app/src/main/java/org/mozilla/fenix/settings/autofill/AutofillSettingFragment.kt
gabrielluong: Ya, it will happen in https://github.com/mozilla-mobile/fenix/issues/24850

app/src/main/java/org/mozilla/fenix/settings/address/AddressEditorFragment.kt
gabrielluong: Ya, I could make the change

app/src/test/java/org/mozilla/fenix/settings/address/controller/DefaultAddressEditorControllerTest.kt
```diff
@@ -0,0 +1,85 @@
...
+                ioDispatcher = testDispatcher
+            )
+        )
+    }
+
+    @After
+    fun cleanUp() {
+        testCoroutineScope.cleanupTestCoroutines()
+    }
+
```
MatthewTighe, line 51: `MainCoroutineRule` should handle the cleanup of test coroutines automatically

app/src/main/java/org/mozilla/fenix/settings/address/controller/AddressEditorController.kt
```diff
@@ -0,0 +1,63 @@
...
+
+    override fun handleSaveAddress(addressFields: UpdatableAddressFields) {
+        lifecycleScope.launch(ioDispatcher) {
+            storage.addAddress(addressFields)
+
+            lifecycleScope.launch(Dispatchers.Main) {
+                navController.popBackStack()
+            }
+        }
+    }
```
MatthewTighe, line 62: It looks like the storage layer already handles switching to an IO dispatcher, so I don't think it's a necessary property for this class.

IMO it's best in general to handle context switching as close in the stack to the actual operation that requires it. This is evidenced in several popular libraries like Room and Retrofit, which will automatically handle moving to an appropriate dispatcher. It's talked about a bit in [this article about Room](https://medium.com/androiddevelopers/room-coroutines-422b786dc4c5). It allows us to put off the expensive operation of thread switching until the last possible moment, in case it's not actually needed.

MatthewTighe (approved):
Had a few comments but LGTM otherwise

app/src/main/java/org/mozilla/fenix/settings/address/AddressEditorFragment.kt
```diff
@@ -0,0 +1,52 @@
...
+                storage = storage,
+                lifecycleScope = lifecycleScope,
+                navController = findNavController()
+            )
+        )
+
+        val binding = FragmentAddressEditorBinding.bind(view)
+
+        addressEditorView = AddressEditorView(binding, interactor)
+        addressEditorView.bind()
```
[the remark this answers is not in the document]
gabrielluong: We will be using it when we implement the options menu. I don't think it's worth the effort to inline this for now and make it class members immediately afterwards.

app/src/main/java/org/mozilla/fenix/settings/address/controller/AddressEditorController.kt
gabrielluong: @MatthewTighe Can you illustrate what your suggested changes should look like? Are we just dropping the `ioDispatcher` here or is the proposal to change this function to a `suspend` function?

app/src/main/java/org/mozilla/fenix/settings/address/interactor/AddressEditorInteractor.kt
```diff
@@ -0,0 +1,47 @@
...
+) : AddressEditorInteractor {
+
+    override fun onCancelButtonClicked() {
+        controller.handleCancelButtonClicked()
+    }
+
+    override fun onSaveAddress(addressFields: UpdatableAddressFields) {
+        controller.handleSaveAddress(addressFields)
+    }
+}
```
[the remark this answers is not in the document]
gabrielluong: You aren't the first person to propose this. However, I think we would prefer to formalize such an architectural decision via a PRD and do a larger refactor on Fenix as a whole instead of having only Addresses be the one that diverges from the existing architecture.

app/src/main/java/org/mozilla/fenix/settings/address/controller/AddressEditorController.kt
MatthewTighe: I think you can just drop the `ioDispatcher` in this case.

app/src/main/java/org/mozilla/fenix/settings/address/interactor/AddressEditorInteractor.kt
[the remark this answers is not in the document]
MatthewTighe: I added a bullet to the next weekly meeting if we wanna talk about it further then.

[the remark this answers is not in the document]
gabrielluong: I meant RFC instead of PRD. Here's some prior art https://github.com/mozilla-mobile/android-components/blob/main/docs/rfcs/0001-rfc-process.md.

Issue #23876, opened 2022-02-22
Update kotlin-coroutines to 1.6.0


Mugurell:
As a followup to https://github.com/mozilla-mobile/fenix/issues/23871 which will increase the kotlin version to 1.6.10 according to https://kotlinlang.org/docs/releases.html#release-details the recommended version for kotlin.coroutines is 1.6.0.

The 1.6.0 release comes with many changes
- https://blog.jetbrains.com/kotlin/2021/12/introducing-kotlinx-coroutines-1-6-0/
- https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-test/MIGRATION.md

which is why it was left out of https://github.com/mozilla-mobile/fenix/issues/23871.

Mugurell:
Changes landed on both AC and Fenix.
Everything seems stable but we should still observe CI test result in the next days.

PR #24892, opened 2022-04-20
For #23876 - Update kotlinx-coroutines to 1.6.1. Refactor runBlockig in tests

Mugurell:
After the update `TestCoroutineDispatcher`, `TestCoroutineScope` and
`runBlockingTest` are deprecated.
Instead of these new `TestDispatcher`, `TestScope` and `runTest` APIs are used.

Took the oportunity to also replace most of the `runBlocking` calls with the
new `runTest` specially designed for running coroutines in tests API.

Needs https://github.com/mozilla-mobile/android-components/pull/12006
All tests pass locally.

----------

Copying here my reasoning described on AC:
- goal was to refactor out `TestCoroutineScope`, `TestCoroutineDispatcher` and `runBlockingTests` since they are deprecated and would fail the build (if not suppressing the deprecation)
- as recommended I also tried to remove as many `runBlocking` usages as possible but in some cases with more intricate scenarios where interacting with other dispatchers / threads was needed some `runBlocking` calls were kept.
- `runTest` replaces all `runBlocking` and `runBlockingTest` usages.
`TestScope`, `UnconfinedTestDispatcher` and `StandardTestDispatcher` replaces all usages of `TestCoroutineScope`, `TestCoroutineDispatcher` as per the official documentation.
- `runTestOnMain` was used where we'd set the main dispatcher to be able to syncronize runs/code in tests/system under test. Occasionally when the `runTest` replacement would have the test fail I added the `MainCoroutineTest` rule and saw that the test passed with also using `runTestOnMain`.
- there are no modifications in the code being tested and no modifications in the tests logic. This seems like a good guarantee that if all tests pass on CI then the changes are good.

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/test/java/org/mozilla/fenix/exceptions/login/LoginExceptionsInteractorTest.kt
+++ b/app/src/test/java/org/mozilla/fenix/exceptions/login/LoginExceptionsInteractorTest.kt
@@ -6,11 +6,11 @@ package org.mozilla.fenix.exceptions.login
 
 import io.mockk.mockk
 import io.mockk.verify
-import kotlinx.coroutines.test.TestCoroutineScope
-import kotlinx.coroutines.test.runBlockingTest
+import kotlinx.coroutines.test.TestScope
+import kotlinx.coroutines.test.UnconfinedTestDispatcher
+import kotlinx.coroutines.test.runTest
 import mozilla.components.feature.logins.exceptions.LoginException
 import mozilla.components.feature.logins.exceptions.LoginExceptionStorage
-import org.junit.After
 import org.junit.Before
 import org.junit.Test
 
@@ -18,7 +18,7 @@ class LoginExceptionsInteractorTest {
 
     private lateinit var loginExceptionStorage: LoginExceptionStorage
     private lateinit var interactor: LoginExceptionsInteractor
-    private val scope = TestCoroutineScope()
+    private val scope = TestScope(UnconfinedTestDispatcher())
 
     @Before
     fun setup() {
@@ -26,19 +26,14 @@ class LoginExceptionsInteractorTest {
         interactor = DefaultLoginExceptionsInteractor(scope, loginExceptionStorage)
     }
 
-    @After
-    fun cleanUp() {
-        scope.cleanupTestCoroutines()
-    }
-
     @Test
-    fun onDeleteAll() = scope.runBlockingTest {
+    fun onDeleteAll() = scope.runTest {
         interactor.onDeleteAll()
         verify { loginExceptionStorage.deleteAllLoginExceptions() }
     }
 
     @Test
-    fun onDeleteOne() = scope.runBlockingTest {
+    fun onDeleteOne() = scope.runTest {
         val exceptionsItem: LoginException = mockk()
         interactor.onDeleteOne(exceptionsItem)
         verify { loginExceptionStorage.removeLoginException(exceptionsItem) }
--- a/app/src/test/java/org/mozilla/fenix/gleanplumb/OnDiskMessageMetadataStorageTest.kt
+++ b/app/src/test/java/org/mozilla/fenix/gleanplumb/OnDiskMessageMetadataStorageTest.kt
@@ -10,7 +10,7 @@ import io.mockk.coVerify
 import io.mockk.just
 import io.mockk.spyk
 import io.mockk.verify
-import kotlinx.coroutines.test.runBlockingTest
+import kotlinx.coroutines.test.runTest
 import mozilla.components.support.test.robolectric.testContext
 import org.json.JSONArray
 import org.json.JSONObject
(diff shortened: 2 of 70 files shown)


app/src/test/java/org/mozilla/fenix/exceptions/login/LoginExceptionsInteractorTest.kt
```diff
@@ -6,39 +6,34 @@ package org.mozilla.fenix.exceptions.login
...
     }
 
-    @After
-    fun cleanUp() {
-        scope.cleanupTestCoroutines()
-    }
-
     @Test
-    fun onDeleteAll() = scope.runBlockingTest {
+    fun onDeleteAll() = scope.runTest {
```
MatthewTighe, line 30: Just out of curiosity, why are `UnconfinedTestDispatcher` and the extension function version of `runTest` required for this test class?

app/src/test/java/org/mozilla/fenix/gleanplumb/OnDiskMessageMetadataStorageTest.kt
```diff
@@ -43,7 +44,7 @@ class OnDiskMessageMetadataStorageTest {
 
     @Test
     fun `GIVEN metadata is not loaded from disk WHEN calling getMetadata THEN load it`() =
-        runBlockingTest {
+        runTest(UnconfinedTestDispatcher()) {
```
MatthewTighe, line 47: Also curious about the usage of `UnconfinedTestDispatcher` here.

MatthewTighe (commented):
Had a couple questions/comments. LGTM otherwise, I'll come back for an approving review once CI is passing after [12006](https://github.com/mozilla-mobile/android-components/pull/12006).

app/src/test/java/org/mozilla/fenix/exceptions/login/LoginExceptionsInteractorTest.kt
Mugurell: There are two dispatchers to choose from: `UnconfinedTestDispatcher` and `StandardTestDispatcher`.
`UnconfinedTestDispatcher` has the nice advantage (here) that it eagerly enters `launch` and `async` blocks.
We need this since the interactor executes for example
```
ioScope.launch {
    loginExceptionStorage.removeLoginException(item)
}
```
Using the same scope (and dispatcher) in the interactor and for running the test allows to sync the execution.

app/src/test/java/org/mozilla/fenix/gleanplumb/OnDiskMessageMetadataStorageTest.kt
Mugurell: Tested now. Here it doesn't seem to be needed. Doesn't break anything but is not required.
Will remove it in favor of the default `StandardTestDispatcher`.

app/src/test/java/org/mozilla/fenix/settings/deletebrowsingdata/DefaultDeleteBrowsingDataControllerTest.kt
```diff
@@ -57,7 +57,7 @@ class DefaultDeleteBrowsingDataControllerTest {
     }
 
     @Test
-    fun deleteTabs() = runBlockingTest {
+    fun deleteTabs() = runTestOnMain {
```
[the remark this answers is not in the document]
Mugurell: Thanks. This was probably just a quick `use runTestOnMain anywhere MainCoroutineRule is set` though I missed setting the appropriate context.
Just to enforce this automatism I think the best way to reconcile this is to replace the context and keep runTestOnMain.

Mugurell:
Rebased to fix conflicts, replaced a few more `runBlocking` occurences.

app/src/test/java/org/mozilla/fenix/exceptions/login/LoginExceptionsInteractorTest.kt
csadilek: Yes, switching to `UnconfinedTestDispatcher` does make the nested coroutines launch, but they can launch on a different thread and we're not synchronizing the execution i.e., the test code is not waiting for their completion. For many tests this will not be good enough and lead to intermittent failures. I've added and example here: https://github.com/mozilla-mobile/android-components/pull/12006#pullrequestreview-960776443

It's easy to fix, but we should review the tests that launch nested coroutines and make assertion based on them.

Mugurell:
CI is green. Rebasing and trying again.

csadilek (approved):
We've worked through this in A-C and this follows the same pattern: https://github.com/mozilla-mobile/android-components/pull/12006#pullrequestreview-966690533

🚢

csadilek:
💚

Issue #24665, opened 2022-04-07
Recent synced tab placeholders should be displayed sooner. 


MatthewTighe:
#24177 introduced the most recent synced tab onto the homepage. Currently, the card will pop in shortly after the user has authenticated. This can be a pretty jarring visual effect. Update the feature so that the following is true:

- If a user has not ever logged in and will therefore not need to authenticate, never show the card
- If the user will be initiating syncs (to our knowledge), display the placeholder card immediately.

Auth appears to take a noticeable amount of time, so this will probably require a completely localized workaround.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-20101)

PR #24894, opened 2022-04-20
for #24665: Show recent synced placeholders early

MatthewTighe:
for #24665

Separate commits since the first is technically unrelated, in case we want to file it for a different issue. I could also squash these if a single commit makes sense. During investigation I thought to maybe reuse `fxaHasSyncedItems`, but then I noticed that it was actually unused. 

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/home/recentsyncedtabs/RecentSyncedTabFeature.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/recentsyncedtabs/RecentSyncedTabFeature.kt
@@ -85,7 +85,10 @@ class RecentSyncedTabFeature(
     override fun stopLoading() = Unit
 
     override fun onError(error: SyncedTabsView.ErrorType) {
-        store.dispatch(AppAction.RecentSyncedTabStateChange(RecentSyncedTabState.None))
+        val isSyncing = store.state.recentSyncedTabState == RecentSyncedTabState.Loading
+        if ((!isSyncing && error == SyncedTabsView.ErrorType.NO_TABS_AVAILABLE) || error.isFatal()) {
+            store.dispatch(AppAction.RecentSyncedTabStateChange(RecentSyncedTabState.None))
+        }
     }
 
     override fun start() {
@@ -106,6 +109,19 @@ class RecentSyncedTabFeature(
         if (tab == lastSyncedTab) {
             RecentSyncedTabs.latestSyncedTabIsStale.add()
         }
+    }
+
+    // Fatal errors represent any that will force a NONE state for the recent synced tab, such that
+    // it won't be displayed.
+    // SYNC_UNAVAILABLE is sent even though displaySyncedTabs is never called when an account
+    // is not authenticated. NO_TABS_AVAILABLE is only fatal if encountered after a sync is
+    // completed, and is handled separately above.
+    private fun SyncedTabsView.ErrorType.isFatal(): Boolean = when (this) {
+        SyncedTabsView.ErrorType.MULTIPLE_DEVICES_UNAVAILABLE,
+        SyncedTabsView.ErrorType.SYNC_ENGINE_UNAVAILABLE,
+        SyncedTabsView.ErrorType.SYNC_NEEDS_REAUTHENTICATION -> true
+        SyncedTabsView.ErrorType.NO_TABS_AVAILABLE,
+        SyncedTabsView.ErrorType.SYNC_UNAVAILABLE -> false
     }
 }
 
--- a/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
+++ b/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
@@ -1012,6 +1012,16 @@ class Settings(private val appContext: Context) : PreferencesHolder {
         default = true
     )
 
+    var hasFxaAuthenticated by booleanPreference(
+        appContext.getPreferenceKey(R.string.pref_key_fxa_has_authenticated),
+        default = false
+    )
+
+    var lastPlacesStorageMaintenance by longPreference(
+        appContext.getPreferenceKey(R.string.pref_key_last_maintenance),
+        default = 0
+    )
+
     fun addSearchWidgetInstalled(count: Int) {
         val key = appContext.getPreferenceKey(R.string.pref_key_search_widget_installed)
         val newValue = preferences.getInt(key, 0) + count
--- a/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
+++ b/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
@@ -1012,16 +1012,6 @@ class Settings(private val appContext: Context) : PreferencesHolder {
         default = true
     )
 
-    var fxaHasSyncedItems by booleanPreference(
-        appContext.getPreferenceKey(R.string.pref_key_fxa_has_synced_items),
-        default = false
-    )
-
-    var lastPlacesStorageMaintenance by longPreference(
-        appContext.getPreferenceKey(R.string.pref_key_last_maintenance),
-        default = 0
-    )
-
     fun addSearchWidgetInstalled(count: Int) {
         val key = appContext.getPreferenceKey(R.string.pref_key_search_widget_installed)
         val newValue = preferences.getInt(key, 0) + count
--- a/app/src/main/java/org/mozilla/fenix/components/BackgroundServices.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/BackgroundServices.kt
@@ -124,6 +124,8 @@ class BackgroundServices(
         }
     }
 
+    private val accountAuthenticationObserver = AccountAuthenticationObserver(context.settings())
+
     private val telemetryAccountObserver = TelemetryAccountObserver(
         context.settings(),
     )
@@ -163,6 +165,9 @@ class BackgroundServices(
         ),
         crashReporter
     ).also { accountManager ->
+        // Register an authentication account observer to cache status
+        accountManager.register(accountAuthenticationObserver)
+
         // Register a telemetry account observer to keep track of FxA auth metrics.
         accountManager.register(telemetryAccountObserver)
 
@@ -193,6 +198,16 @@ class BackgroundServices(
      */
     private val notificationManager by lazyMonitored {
         NotificationManager(context)
+    }
+}
+
+private class AccountAuthenticationObserver(private val settings: Settings) : AccountObserver {
+    override fun onAuthenticated(account: OAuthAccount, authType: AuthType) {
+        settings.hasFxaAuthenticated = true
+    }
+
+    override fun onLoggedOut() {
+        settings.hasFxaAuthenticated = false
     }
 }
 
--- a/app/src/main/java/org/mozilla/fenix/components/BackgroundServices.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/BackgroundServices.kt
@@ -30,7 +30,6 @@ import mozilla.components.service.fxa.SyncEngine
 import mozilla.components.service.fxa.manager.FxaAccountManager
 import mozilla.components.service.fxa.manager.SCOPE_SESSION
 import mozilla.components.service.fxa.manager.SCOPE_SYNC
-import mozilla.components.service.fxa.manager.SyncEnginesStorage
 import mozilla.components.service.fxa.sync.GlobalSyncableStoreProvider
 import mozilla.components.service.glean.private.NoExtras
 import mozilla.components.service.sync.autofill.AutofillCreditCardsAddressesStorage
@@ -164,11 +163,6 @@ class BackgroundServices(
         ),
         crashReporter
     ).also { accountManager ->
-        // TODO this needs to change once we have a SyncManager
-        context.settings().fxaHasSyncedItems = accountManager.authenticatedAccount()?.let {
-            SyncEnginesStorage(context).getStatus().any { it.value }
-        } ?: false
-
         // Register a telemetry account observer to keep track of FxA auth metrics.
         accountManager.register(telemetryAccountObserver)
 
--- a/app/src/main/java/org/mozilla/fenix/components/Components.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/Components.kt
@@ -37,6 +37,7 @@ import org.mozilla.fenix.ext.sort
 import org.mozilla.fenix.home.PocketUpdatesMiddleware
 import org.mozilla.fenix.home.blocklist.BlocklistHandler
 import org.mozilla.fenix.home.blocklist.BlocklistMiddleware
+import org.mozilla.fenix.home.recentsyncedtabs.RecentSyncedTabState
 import org.mozilla.fenix.perf.AppStartReasonProvider
 import org.mozilla.fenix.perf.StartupActivityLog
 import org.mozilla.fenix.perf.StartupStateProvider
@@ -213,6 +214,11 @@ class Components(private val context: Context) {
                     core.store.state.asRecentTabs()
                 } else {
                     emptyList()
+                },
+                recentSyncedTabState = if (settings.hasFxaAuthenticated) {
+                    RecentSyncedTabState.Loading
+                } else {
+                    RecentSyncedTabState.None
                 },
                 recentHistory = emptyList()
             ).run { filterState(blocklistHandler) },
--- a/app/src/main/res/values/preference_keys.xml
+++ b/app/src/main/res/values/preference_keys.xml
@@ -85,6 +85,7 @@
     <string name="pref_key_sign_out" translatable="false">pref_key_sign_out</string>
     <string name="pref_key_sync_sign_in" translatable="false">pref_key_sync_sign_in</string>
     <string name="pref_key_push_project_id" translatable="false">project_id</string>
+    <string name="pref_key_fxa_has_authenticated" translatable="false">pref_key_fxa_has_authenticated</string>
     <string name="pref_key_search_widget_installed" translatable="false">pref_key_search_widget_installed</string>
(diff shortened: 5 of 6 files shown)


app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -1012,8 +1012,8 @@ class Settings(private val appContext: Context) : PreferencesHolder {
         default = true
     )
 
-    var fxaHasSyncedItems by booleanPreference(
-        appContext.getPreferenceKey(R.string.pref_key_fxa_has_synced_items),
+    var fxaHasAuthenticated by booleanPreference(
```
gabrielluong, line 1015:
```suggestion
    var hasFxaAuthenticated by booleanPreference(
```
Drive-by, maybe?

gabrielluong, line 1015:
```suggestion
    var hasFxaAuthenticated by booleanPreference(
```
Drive-by, maybe?

app/src/main/java/org/mozilla/fenix/home/recentsyncedtabs/RecentSyncedTabFeature.kt
```diff
@@ -107,6 +113,17 @@ class RecentSyncedTabFeature(
             RecentSyncedTabs.latestSyncedTabIsStale.add()
         }
     }
+
+    // SYNC_UNAVAILABLE is sent even though displaySyncedTabs is never called when an account
```
gabrielluong, line 117: Could use a description of what fatal actually means beyond the individual error types

gabrielluong, line 117: Could use a description of what fatal actually means beyond the individual error types

app/src/main/java/org/mozilla/fenix/home/recentsyncedtabs/RecentSyncedTabFeature.kt
```diff
@@ -82,10 +84,14 @@ class RecentSyncedTabFeature(
...
     // UI will either not be displayed if not authenticated (DefaultPresenter.start),
     // or the display state will be tied directly to the success and error cases.
-    override fun stopLoading() = Unit
+    override fun stopLoading() {
+        isSyncing = false
+    }
 
     override fun onError(error: SyncedTabsView.ErrorType) {
-        store.dispatch(AppAction.RecentSyncedTabStateChange(RecentSyncedTabState.None))
+        if ((!isSyncing && error == SyncedTabsView.ErrorType.NO_TABS_AVAILABLE) || error.isFatal()) {
```
gabrielluong, line 92: Instead of having an `isSyncing` variable, could we check `store.state.recentSyncedTabState` to see if it's loading? Not saying this is necessarily the correct thing to do, but in prior reviews I had an inkling that maybe `recentSyncedTabState` should be tri-state with `stopLoading` implemented.

gabrielluong, line 92: Instead of having an `isSyncing` variable, could we check `store.state.recentSyncedTabState` to see if it's loading? Not saying this is necessarily the correct thing to do, but in prior reviews I had an inkling that maybe `recentSyncedTabState` should be tri-state with `stopLoading` implemented.

MatthewTighe: I'll update it to inspect the store's state, but would you mind clarifying about the tri-state? Maybe it's time to revisit that since we've already talked about it at least once. 

Is the idea to introduce an error state? From the consumer's perspective, there isn't (currently) a desire to display error messages, which is why I've been defaulting to `NONE` on error.

gabrielluong: It might be something more like `START`( or `NONE`), `LOADING`, `DONE`. I don't know if this will necessarily be correct, but instead of having `NONE`, we need to have a state that differentiates between whether or not loading was attempted.

MatthewTighe: There currently are three states that I think map pretty well to those: `NONE`, `LOADING`, and `SUCCESS`. Switching `SUCCESS` to a `DONE` state that had either successes or failures could potentially be an improvement, but I'm not sure what it benefits over the current implementation.

Honestly I think the underlying [Controller](https://searchfox.org/mozilla-mobile/source/android-components/components/feature/syncedtabs/src/main/java/mozilla/components/feature/syncedtabs/controller/DefaultController.kt) implementation's usage of `stopLoading` is generally a bit confusing. The only case it's ever really useful IMO is the one where an account isn't authenticated, but `refreshSyncedTabs` shouldn't be called unless an account is authenticated anyway.

app/src/main/java/org/mozilla/fenix/home/recentsyncedtabs/RecentSyncedTabFeature.kt
```diff
@@ -85,7 +85,10 @@ class RecentSyncedTabFeature(
     override fun stopLoading() = Unit
 
     override fun onError(error: SyncedTabsView.ErrorType) {
-        store.dispatch(AppAction.RecentSyncedTabStateChange(RecentSyncedTabState.None))
+        val isSyncing = store.state.recentSyncedTabState == RecentSyncedTabState.Loading
+        if ((!isSyncing && error == SyncedTabsView.ErrorType.NO_TABS_AVAILABLE) || error.isFatal()) {
```
gabrielluong, line 89: Since we do have a RecentSyncedTabFeatureTest, be sure to add or update the tests as necessary.

gabrielluong (approved):

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -1012,8 +1012,8 @@ class Settings(private val appContext: Context) : PreferencesHolder {
         default = true
     )
 
-    var fxaHasSyncedItems by booleanPreference(
-        appContext.getPreferenceKey(R.string.pref_key_fxa_has_synced_items),
+    var hasFxaAuthenticated by booleanPreference(
```
MozillaNoah, line 1015: Using `Settings`/`Preferences` to monitor the authentication state seems a little sus. Were there other approaches you tried that didn't pan out?

MatthewTighe: The underlying issue is that auth is (relatively) slow, to the point where there will be noticeable pop-in of the recent synced tab if we wait for it to finish. To me that demonstrates that we need some kind of cache mechanism so that we can show the placeholders immediately on app startup. If you have any other suggestions, I'm definitely all ears. Is it specifically using `Settings`? Do we have another easy file abstraction that I've so far missed?

One alternative I can think of off the top of my head is to cache it in the `FxaAccountManager` itself, but that would require adding a storage layer there as well and would ultimately be the same type of implementation.

MozillaNoah: (for the purposes of this PR, this can still land)

I was just curious if you looked into having this live in `AppState` instead. That way it could be observed in other places via `appState.observeAsComposableState` in case other areas of the app can easily consume this in Compose, if necessary in the future.

MatthewTighe: I filed #24911 to move it into app state, but the more I think about it the more I wonder whether it's actually the right move. Here's some of my thoughts:

- `Store`s are generally meant to be an observable entity. Any component would need to observe changes to `hasFxaAuthenticated` can also observe changes to auth status directly from an `FxaAccountManager`. 
- `hasFxaAuthenticated` is really meant as a one-time read value on startup
- Even if we move it into the store, it will still need to be provided from settings or some other kind of cache mechanism, so I think we'd just be creating a level of indirection

MozillaNoah: I think those concerns/points are fair

MatthewTighe: I closed it for now, then. Let me know if want to do something different in the future, happy to revisit this again.

jonalmeida:
I would have preferred we didn't land this. Using settings isn't a great idea because we're duplicating _and persisting_ the same state to multiple places.

There also have [other account state holders](https://github.com/mozilla-mobile/fenix/blob/main/app/src/main/java/org/mozilla/fenix/components/accounts/FenixAccountManager.kt), so we've added yet another mechanism to do the same/similar thing.

We should add the account status into the AppStore write out a better abstraction that held the Sync state (e.g. an account data class with an enum and maybe some other attributes) - this was one of the [original applications](https://github.com/mozilla-mobile/fenix/issues/20704#issuecomment-956522471=) for adding the AppStore. It removes the need to know how the account manager works for features that just want to know what is current state at this point in time, or observe it which idioms that we've already created with the Store model.

> `Stores` are generally meant to be an observable entity.

This is true, but we can also "peek" at the current state in the store if it's a safe use-case for it.

> Even if we move it into the store, it will still need to be provided from settings or some other kind of cache mechanism, so I think we'd just be creating a level of indirection

We're currently fetching the account state remotely, so another caching layer would be an in-memory solution which the AppStore would suffice? If you mean a secondary persistent caching, it has risks and benefits there which we can discuss separately.

MatthewTighe:
@jonalmeida Sorry about that, I'll try to be more patient in the future when waiting for reviews. I'm happy to back out that particular part of the change in favor of something else.

> Using settings isn't a great idea because we're duplicating and persisting the same state to multiple places.

Could you point me to where the auth state of an account specifically is persisted? That's the key requirement. The goal of this issue was to avoid pop-in on the home screen on app startup. Since there is a significant delay in account authentication, the only way to do that is with a disk cache.

jonalmeida:
>The goal of this issue was to avoid pop-in on the home screen on app startup. Since there is a significant delay in account authentication, the only way to do that is with a disk cache.

This is the same problem that we encounter on the home screen today with our other sources of data, so we could still use the same solution: the first cold start will be slow, we store it in the AppStore, when we return to the home screen again, we will already have it in memory. Why is the Synced tab a higher priority than the others?

We can also go the persistent cache route, so let's walk backwards to figure out how account manager persists it's state today and see what our options are. Writing out my learnings here as I walk through the code:
1. If we look at the [`FxaAccountManager.kt`](https://github.com/mozilla-mobile/android-components/blob/main/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt) we need to find a reference to some kind of persistence.
2. There is one! `accountOnDisk` looks like what we want, and that leads to the [`StorageWrapper`](https://github.com/mozilla-mobile/android-components/blob/main/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/AccountStorage.kt) where the docs hint that we are in the right direction: "Knows how to read account from disk (or creating a new instance if there's no account), registering necessary watchers."
3. If we keep following along, the `StatePersistenceCallback.persist(String)` is a JSON object as a string.
4. The JSON object is created from an instance of `FirefoxAccount` which calls `toJson` that lives in... [app services](https://github.com/mozilla/application-services/blob/main/components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/PersistedFirefoxAccount.kt)! This is interesting and good, which means Android-land isn't duplicating the account state persistence in a different way from other platforms (iOS/Desktop), so there is a good chance we have what we need for knowing the account is already an authenticated account (🤞 ).
5. So the inner class that holds the data is `FirefoxAccount` which is a generated class from app services' uniffi framework, which is what they use to create language bindings for each platform. So let's look at the `fxa_client.udl` that holds the definition of the class. From there, we can see that [they persist](https://github.com/mozilla/application-services/blob/main/components/fxa-client/src/fxa_client.udl#L910-L933=) the account name, email and avatar in a dictionary.
6. Coming back to what you want, is to know if the account was already authenticated sooner (even though there is a chance it may not always be that way when we check with the servers), I would ask the app services' team what they think is the best way to add the state needed to know if an account was recently authenticated to make start-up faster and accordingly ask for their support (or write some rust!) before proceeding to add it into the Android FxaAccountManager. What they might tell us is that we are fine with the drawbacks, we could write our own method to check for the existence of some other state in the JSON object which we can use to stay that an account was authenticated at some point.

Why go this route? If we start storing loosely maintained state in Fenix, we could start putting clients at risk of incorrectly saying they are (not) authenticated, which over time builds up a painful kind of tech debt. :)

MatthewTighe:
Thanks for the detailed walkthrough! I'm still working on growing the skill of navigating our stack, so I appreciate the breakdown and reminder to think more about what belongs where. I posted some more thoughts below. Next week I'll keep taking a look into it, and figure out the best way to back out the change. This work is currently hidden (even from nightly) behind a flag, so it shouldn't have any real negative impact.

> This is the same problem that we encounter on the home screen today with our other sources of data
I think the biggest difference is just source of the data. AFAIK, most of that data is coming from a local source. The previous implementation of the synced tab was causing the loading placeholder to appear a full few seconds after the rest of the screen had populated.

> Coming back to what you want, is to know if the account was already authenticated sooner (even though there is a chance it may not always be that way when we check with the servers)
We decided that it would be better for the placeholder to disappear in the case that it was shown before an auth failure. The thinking being that pop-in was more visually disruptive than the pop-out.

I'm initially wondering if exposing whether `accountOnDisk` is restored or not would work. I need to look more into whether it would cause harmful side-effects to inspect it on instantiation instead of during the normal processing of the state machine. 

I'm not sure I'm following yet why it would be required to add additional state at the A-S layer. I would've thought that checking the existence of any state there would get us what we were looking for.

MatthewTighe:
> We should add the account status into the AppStore with a better abstraction that holds the Sync state

I'm wondering if we should possible take this a step further. I see additional benefits to introducing the sync state and data to `BrowsterStore` (perhaps even a `SyncStore`) at the A-C level. Any reductions in complexity we would see Fenix-side could be mirrored in existing features and usecases in A-C. In general, having a simpler abstraction around the FxA manager and various sync workers seems useful.

jonalmeida:
> Just to add some detail about expected UX as a result of that: we decided that it would be better for the placeholder to disappear in the case that it was shown before an auth failure. The thinking being that pop-in was more visually disruptive than the pop-out.

How often does the auth failure happen that we should prioritize the visual disruption? With the solution in this PR, you would still run into this case regardless so we are still not covered there; we need to wait for the Sync servers to get back and tell us that the auth token we're trying to use now is no longer valid. Our best effort might just be to live with the edge case of an auth failure from a user that hasn't used Fenix in a while so they see a brief visual error. See below how we can still solve your problem with the help of the A-S team's guidance.

> I'm initially wondering if exposing whether `accountOnDisk` is restored or not would work. I need to look more into whether it would cause harmful side-effects to inspect it on instantiation instead of during the normal processing of the state machine.

This is a good question for the A-S team. They would be able to tell us exactly what is the right attribute to observe in order to consider the account authenticated at some point.

> I'm not sure I'm following yet why it would be required to add additional state at the A-S layer. I would've thought that checking the existence of any state there would get us what we were looking for.

The `accountOnDisk` is a JSON object, so you could have that object exist without ever having an account authenticated within it. Or even a shell JSON object may exist without an account if the FxaAccountManager is only instantiated - from what I can read in the code [here and the related bits](https://github.com/mozilla-mobile/android-components/blob/ac1c4ad38afaaafdf9a7fcd2e03cee04f92fd833/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt#L584-L595=), a piece of code could call `accountManager.account()` to test if an account is created and inadvertently create that shell JSON object on file.
I'm not really sure, so it would be good to ask the A-S folks what is the right thing to do and we can evaluate from there if we should consider the existence of the object on disk or a particular attribute within the object.

> I'm wondering if we should possible take this a step further. I see additional benefits to introducing the sync state and data to `BrowsterStore` (perhaps even a `SyncStore`) at the A-C level. Any reductions in complexity we would see Fenix-side could be mirrored in existing features and usecases in A-C. In general, having a simpler abstraction around the FxA manager and various sync workers seems useful.

Oh yes, 100%. Moving the state to the AC level was [an original ask](https://github.com/mozilla-mobile/android-components/issues/7119#issuecomment-635720254). It doesn't necessarily mean that we need to replace the current API immediately. Rather, provide a nicer API for consumers to migrate towards in addition to the current API, and then we have more freedom to alter the internals of the manager (and re-design the observer pattern it uses or even remove it).

I don't have strong opinions towards whether it should be a separate store within the `BrowserStore`, but I would lean towards a separate store mostly because, if you look at the internals of the account manager, it's basically a state machine and that's really the same thing we get with `lib-state`. So we could eventually replace the custom state machine in the future (historically, the account manager existed before lib-state, which is why it rolled it's own version) - this might be a good question to get @csadilek's opinion on as well before deciding to tackle it.

csadilek:
I am leaning towards a separate store as well. This isn't state needed for core browsing functionality, and from the perspective of Focus, why would this empty state be in the browser state? I think it also aligns better with our module separation and composable design e.g., to only bring in state as needed by the application.

MatthewTighe:
I filed [A-C 12065](https://github.com/mozilla-mobile/android-components/issues/12065) to track the work of adding a sync store. It's work I'm very interested in taking on, but may need to wait until more progress has been made on various items planned for H1. At the latest, I'll plan to push for it to be included in H2.

Additionally, it looks like my understanding of the issue this PR was intended to solve was just completely wrong. I've continued tinkering with it on the A-C and Fenix side, and it appears that auth state will be available as soon as other data from the homepage is. I'll have PRs up today to revert the change introduced here and still achieve desired behavior. Thanks for everyone's patience in engaging in the conversation, as it was very educational. That said, I don't think there will be any need to engage A-S for improvements and any required changes can be handled on our end.

Issue #24929, opened 2022-04-22
Only promotion wallpapers should be locale-restricted.


MatthewTighe:
There were two Firefox wallpapers recently added to the remote wallpaper list. Unlike the promotional offerings, these wallpapers should not be locale-restricted. All downloaded wallpapers are currently hidden behind the same locale flag. 

- Remove the restriction for the Firefox wallpapers. 
- Promotional wallpapers should still be locale-restricted.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-20258)

PR #17580, opened 2021-01-22
For #17304 - Implement top site search engine attribution

gabrielluong:
Fixes #17304 

### Pull Request checklist

- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features. In addition, it includes a screenshot of a successful [accessibility scan](https://play.google.com/store/apps/details?id=com.google.android.apps.accessibility.auditor&hl=en_US) to ensure no new defects are added to the product.

### To download an APK when reviewing a PR:
1. click on Show All Checks,
2. click Details next to "Taskcluster (pull_request)" after it appears and then finishes with a green checkmark,
3. click on the "Fenix - assemble" task, then click "Run Artifacts".
4. the APK links should be on the left side of the screen, named for each CPU architecture

PR #24930, opened 2022-04-22
for #24929: remove locale restriction for remote firefox wallpapers

MatthewTighe:
for #24929

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -1210,7 +1210,7 @@ class HomeFragment : Fragment() {
         return shouldEnableWallpaper() && settings.shouldAnimateFirefoxLogo &&
             onboarding.userHasBeenOnboarded() &&
             settings.numberOfAppLaunches >= 3 &&
-            FeatureFlags.isThemedWallpapersFeatureEnabled(localContext)
+            FeatureFlags.showWallpapers
```
Amejia481, line 1213: I think we could remove it, the check it's already part of `shouldEnableWallpaper` :)

Amejia481 (approved):
LGTM!

Amejia481 (commented):
Looks like we have an issue, in the wallpapers page when we enter there with a local that is not promotional, it looks like we are trying to display all wallpapers available. 

image

MatthewTighe:
This should be fixed in my most recent push, I'd also just noticed it. Thanks for the thorough check!

app/src/main/java/org/mozilla/fenix/wallpapers/Wallpaper.kt
```diff
@@ -55,8 +55,10 @@ sealed class Wallpaper {
                 set(2022, Calendar.APRIL, 30)
                 time
             }
-        ) : Remote() {
+        ) : Remote(), Promotional {
             override val remoteParentDirName: String = "house"
+            override fun isAvailableInLocale(locale: String): Boolean =
+                listOf("en-US", "es-US").contains(locale)
```
Amejia481, line 61: 👏🏽

app/src/main/java/org/mozilla/fenix/wallpapers/Wallpaper.kt
```diff
@@ -49,14 +49,19 @@ sealed class Wallpaper {
...
             override val name: String,
             override val expirationDate: Date? = Calendar.getInstance().run {
                 set(2022, Calendar.APRIL, 30)
                 time
             }
-        ) : Remote() {
+        ) : Remote(), Promotional {
             override val remoteParentDirName: String = "house"
+            override fun isAvailableInLocale(locale: String): Boolean =
+                listOf("en-US", "es-US").contains(locale)
```
gabrielluong, line 64: In case this just happens to be useful information, we can tell a user's region based on their `store.state.search.region`. If we were trying to make the wallpaper available for all of US instead of just based on their locale. See https://github.com/mozilla-mobile/fenix/pull/17580/files#diff-3101d2ecb99584cfe0d93898bc7eee5a629ddf6de9f2d2632f56fb3f04e2c34dR395 for an example.

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperManager.kt
```diff
@@ -96,14 +99,21 @@ class WallpaperManager(
...
+    private fun isWallpaperExpired(wallpaper: Wallpaper): Boolean = when (wallpaper) {
         is Wallpaper.Remote -> {
-            val notExpired = wallpaper.expirationDate?.let { Date().before(it) } ?: true
-            notExpired || wallpaper.name == settings.currentWallpaper
+            val expired = wallpaper.expirationDate?.let { Date().after(it) } ?: false
+            expired && wallpaper.name != settings.currentWallpaper
         }
-        else -> true
+        else -> false
     }
```
Amejia481, line 108: The logic looks good, but I would rather avoid changing it now as this is something that we will very likely uplift directly to release (QA already battled tested the previous version) and we are close to prime time 😅 ,  I know we have a good set of tests, but I would avoid the risk.

MatthewTighe: That makes sense. I'll revert the changes there

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperManager.kt
```diff
@@ -32,11 +32,14 @@ class WallpaperManager(
     private val settings: Settings,
     private val downloader: WallpaperDownloader,
     private val fileManager: WallpaperFileManager,
+    private val currentLocale: String,
     allWallpapers: List<Wallpaper> = availableWallpapers
 ) {
     val logger = Logger("WallpaperManager")
 
-    val wallpapers = allWallpapers.filter(::filterExpiredRemoteWallpapers)
```
Amejia481, line 39: I will just introduce another filter here to reduce the risk of missing something close to prime time 😅

app/src/main/java/org/mozilla/fenix/wallpapers/Wallpaper.kt
Amejia481: That's nice!!!
We could add it as a follow up ticket :)

MatthewTighe: That is actually really good to know. I think when we talked about this initially, we didn't think there was another good way for us to determine locale. That's why we included es-US, since it's another high usage locale in the US.

Amejia481:
🚢 !

Amejia481 (approved):

MatthewTighe:
@mergifyio backport releases_v100.0.0

PR #24962, opened 2022-04-26
For #24935: Dismiss keyboard on add address fragment stop.

### LINKED ISSUE #24935: Hide the keyboard when navigating away from the address editor
Found a small issue. When navigating back to the address list from the "Add address" screen, the keyboard stays open.

_Originally posted by @MozillaNoah in https://github.com/mozilla-mobile/fenix/issues/24928#issuecomment-1106955692_

mcarare:
Fixes #24935

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/settings/address/AddressEditorFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/address/AddressEditorFragment.kt
@@ -8,6 +8,7 @@ import android.os.Bundle
 import android.view.View
 import androidx.lifecycle.lifecycleScope
 import androidx.navigation.fragment.findNavController
+import mozilla.components.support.ktx.android.view.hideKeyboard
 import org.mozilla.fenix.R
 import org.mozilla.fenix.SecureFragment
 import org.mozilla.fenix.databinding.FragmentAddressEditorBinding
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/fenix/settings/address/AddressEditorFragment.kt
```diff
@@ -49,4 +50,9 @@ class AddressEditorFragment : SecureFragment(R.layout.fragment_address_editor) {
         super.onResume()
         showToolbar(getString(R.string.addresses_add_address))
     }
+
+    override fun onStop() {
```
gabrielluong, line 54: I see that we do this in `onPause` in the `CreditCardEditorFragment`. https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/java/org/mozilla/fenix/settings/creditcards/CreditCardEditorFragment.kt#102

Do you have any thoughts on whether this should be done on pause or stop? I would prefer if we align both of them.

gabrielluong: A quick search for `view?.hideKeyboard()` also showed me that we typically call this in `onPause()` if that makes a difference.

mcarare: I know it is used in `onPause` in other places, but I attribute that to the fact that other "cleaning" operations are made at that point and it has been added to the bunch. I do not see any reason to have it in `onPause` having it fire multiple times when it is not necessarily needed (putting the fragment in the background and re-entering it would bring back the keyboard anyway). IMO the optimal place to call this is `onStop`, but I do not have anything against moving it to `onPause`.

gabrielluong (approved):

Issue #12067, opened 2020-06-28
FNX-13085 ⁃ [Bug] Unable to Search Easily


pidus:
## Steps to reproduce
* Set Fenix as default browser (optional).
* Load any webpage with text. 
* Select a word or sentence. 
    1. Expected behavior
Easily get an option to search the selection with default search engine.
    1. Actual behavior
The option is hidden under 3-dot menu and after alternative apps like Edge, Wikipedia. 
    1. Device information
Xiaomi Mi Max 2. MIUI 10
* Android device: 7.1.1
* Fenix version: 77.0.0-beta.2

### Fennec:
![IMG](https://user-images.githubusercontent.com/11857051/85940902-dad08200-b90e-11ea-8c50-9b80ade0ea7f.jpg)UNITO-UNDERSCORE!20200628!UNITO-UNDERSCORE!123814!

### Fenix:
![IMG](https://user-images.githubusercontent.com/11857051/85940940-2125e100-b90f-11ea-81b2-b2bc9bf13bb3.jpg)UNITO-UNDERSCORE!20200628!UNITO-UNDERSCORE!123836!
![IMG](https://user-images.githubusercontent.com/11857051/85940942-26832b80-b90f-11ea-92f8-61c36828e05f.jpg)UNITO-UNDERSCORE!20200628!UNITO-UNDERSCORE!123856!

* Related Reddit post: https://redd.it/hg8d36

brampitoyo:
@pidus Acknowledged. This issue will be solved when we fix #10466.

On https://github.com/mozilla-mobile/fenix/issues/10466#issuecomment-626445627, you’ll see that the ordering of the context menu will prioritise “Search”.

PR #24987, opened 2022-04-27
for #24665: show recent synced tab faster

MatthewTighe:
Some slight improvements in coordination with [A-C #12067](https://github.com/mozilla-mobile/android-components/pull/12066) to reduce errors and get the synced tab onto the home screen faster.

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/home/recentsyncedtabs/RecentSyncedTabFeature.kt
```diff
@@ -54,9 +54,10 @@ class RecentSyncedTabFeature(
     override fun startLoading() {
         syncStartId?.let { RecentSyncedTabs.recentSyncedTabTimeToLoad.cancel(it) }
         syncStartId = RecentSyncedTabs.recentSyncedTabTimeToLoad.start()
-        store.dispatch(
-            AppAction.RecentSyncedTabStateChange(RecentSyncedTabState.Loading)
-        )
+        // No need to show loading state if a tab is already shown
```
jonalmeida, line 57: nit: Comment is a bit unnecessary - it says the negative of the if-condition but doesn't really say anything we can't already read.

app/src/main/java/org/mozilla/fenix/home/recentsyncedtabs/RecentSyncedTabFeature.kt
```diff
@@ -80,13 +81,17 @@ class RecentSyncedTabFeature(
         lastSyncedTab = syncedTab
     }
 
-    // UI will either not be displayed if not authenticated (DefaultPresenter.start),
-    // or the display state will be tied directly to the success and error cases.
-    override fun stopLoading() = Unit
+    // This is called in success cases as well, but the state should only change if there isn't a
+    // tab displayed. The store's state isn't updated in time to rely on it for this condition,
+    // so local state is used instead.
```
jonalmeida, line 86:
```suggestion
    /**
     * This is called in success cases as well, but the state should only change if there isn't a
     * tab displayed. The store's state isn't updated in time to rely on it for this condition,
     * so local state is used instead.
     */
```

nit: could you fix the comment formatting to a doc since you're changing the comment itself. Inner comments use the `//`, on the method should follow the doc style with something like "implementation notes" or "NB" prepended where desired.

app/src/main/java/org/mozilla/fenix/home/recentsyncedtabs/RecentSyncedTabFeature.kt
```diff
@@ -80,13 +81,17 @@ class RecentSyncedTabFeature(
...
+    override fun stopLoading() {
+        if (lastSyncedTab == null) {
```
jonalmeida, line 88: When the feature is detached and then destroyed, the `lastSyncedTab` will always be null. So the `stopLoading`  check would also always be true, right? What are the other cases this check will help with?

MatthewTighe: I'm not sure I'm following. Are you suggesting that `stopLoading` is only called when the feature is destroyed? It is also called at end of every call to `DefaultController::refreshSyncedTabs`. This is mostly meant as a sanity check - normally we'll hit either the success case and `lastSyncedTab` will not be null or a failure case and the error will cause a `NONE` state dispatch.

app/src/main/java/org/mozilla/fenix/home/recentsyncedtabs/RecentSyncedTabFeature.kt
MatthewTighe: Sure thing!

app/src/main/java/org/mozilla/fenix/home/recentsyncedtabs/RecentSyncedTabFeature.kt
jonalmeida: Ack, sorry - I misread this the other day. LGTM.

jonalmeida (approved):

PR #25031, opened 2022-04-29
For #24855: Allow updating and deleting an existing address.

### LINKED ISSUE #24855: Update a saved address information
This issue is to display an existing address in the address editor and allow users to update any address field and save it.

Screen Shot 2022-04-18 at 12 29 11 PM

mcarare:
Fixes #24855

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/settings/address/AddressEditorFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/address/AddressEditorFragment.kt
@@ -8,6 +8,7 @@ import android.os.Bundle
 import android.view.View
 import androidx.lifecycle.lifecycleScope
 import androidx.navigation.fragment.findNavController
+import androidx.navigation.fragment.navArgs
 import mozilla.components.support.ktx.android.view.hideKeyboard
 import org.mozilla.fenix.R
 import org.mozilla.fenix.SecureFragment
@@ -27,6 +28,14 @@ class AddressEditorFragment : SecureFragment(R.layout.fragment_address_editor) {
     private lateinit var addressEditorView: AddressEditorView
     private lateinit var interactor: AddressEditorInteractor
 
+    private val args by navArgs<AddressEditorFragmentArgs>()
+
+    /**
+     * Returns true if an existing address is being edited, and false otherwise.
+     */
+    private val isEditing: Boolean
+        get() = args.address != null
+
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
 
@@ -42,13 +51,17 @@ class AddressEditorFragment : SecureFragment(R.layout.fragment_address_editor) {
 
         val binding = FragmentAddressEditorBinding.bind(view)
 
-        addressEditorView = AddressEditorView(binding, interactor)
+        addressEditorView = AddressEditorView(binding, interactor, args.address)
         addressEditorView.bind()
     }
 
     override fun onResume() {
         super.onResume()
-        showToolbar(getString(R.string.addresses_add_address))
+        if (isEditing) {
+            showToolbar(getString(R.string.addresses_edit_address))
+        } else {
+            showToolbar(getString(R.string.addresses_add_address))
+        }
     }
 
     override fun onStop() {
--- a/app/src/main/java/org/mozilla/fenix/settings/address/AddressEditorFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/address/AddressEditorFragment.kt
@@ -5,6 +5,9 @@
 package org.mozilla.fenix.settings.address
 
 import android.os.Bundle
+import android.view.Menu
+import android.view.MenuInflater
+import android.view.MenuItem
 import android.view.View
 import androidx.lifecycle.lifecycleScope
 import androidx.navigation.fragment.findNavController
@@ -36,6 +39,8 @@ class AddressEditorFragment : SecureFragment(R.layout.fragment_address_editor) {
     private val isEditing: Boolean
         get() = args.address != null
 
+    private lateinit var menu: Menu
+
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
 
@@ -50,9 +55,15 @@ class AddressEditorFragment : SecureFragment(R.layout.fragment_address_editor) {
         )
 
         val binding = FragmentAddressEditorBinding.bind(view)
+        setHasOptionsMenu(true)
 
         addressEditorView = AddressEditorView(binding, interactor, args.address)
         addressEditorView.bind()
+    }
+
+    override fun onPause() {
+        super.onPause()
+        menu.close()
     }
 
     override fun onResume() {
@@ -67,5 +78,26 @@ class AddressEditorFragment : SecureFragment(R.layout.fragment_address_editor) {
     override fun onStop() {
         super.onStop()
         this.view?.hideKeyboard()
+    }
+
+    override fun onCreateOptionsMenu(menu: Menu, inflater: MenuInflater) {
+        inflater.inflate(R.menu.address_editor, menu)
+        this.menu = menu
+
+        menu.findItem(R.id.delete_address_button).isVisible = isEditing
+    }
+
+    override fun onOptionsItemSelected(item: MenuItem): Boolean = when (item.itemId) {
+        R.id.delete_address_button -> {
+            args.address?.let {
+                addressEditorView.showConfirmDeleteAddressDialog(requireContext(), it.guid)
+            }
+            true
+        }
+        R.id.save_address_button -> {
+            addressEditorView.saveAddress()
+            true
+        }
+        else -> false
     }
 }
--- a/app/src/main/java/org/mozilla/fenix/settings/address/controller/AddressEditorController.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/address/controller/AddressEditorController.kt
@@ -28,6 +28,11 @@ interface AddressEditorController {
      * @see [AddressEditorInteractor.onSaveAddress]
      */
     fun handleSaveAddress(addressFields: UpdatableAddressFields)
+
+    /**
+     * @see [AddressEditorInteractor.onUpdateAddress]
+     */
+    fun handleUpdateAddress(guid: String, addressFields: UpdatableAddressFields)
 }
 
 /**
@@ -51,6 +56,16 @@ class DefaultAddressEditorController(
     override fun handleSaveAddress(addressFields: UpdatableAddressFields) {
         lifecycleScope.launch {
             storage.addAddress(addressFields)
+
+            lifecycleScope.launch(Dispatchers.Main) {
+                navController.popBackStack()
+            }
+        }
+    }
+
+    override fun handleUpdateAddress(guid: String, addressFields: UpdatableAddressFields) {
+        lifecycleScope.launch {
+            storage.updateAddress(guid, addressFields)
 
             lifecycleScope.launch(Dispatchers.Main) {
                 navController.popBackStack()
--- a/app/src/main/java/org/mozilla/fenix/settings/address/controller/AddressEditorController.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/address/controller/AddressEditorController.kt
@@ -30,6 +30,11 @@ interface AddressEditorController {
     fun handleSaveAddress(addressFields: UpdatableAddressFields)
 
     /**
+     * @see [AddressEditorInteractor.onDeleteAddress]
+     */
+    fun handleDeleteAddress(guid: String)
+
+    /**
      * @see [AddressEditorInteractor.onUpdateAddress]
      */
     fun handleUpdateAddress(guid: String, addressFields: UpdatableAddressFields)
@@ -56,6 +61,16 @@ class DefaultAddressEditorController(
     override fun handleSaveAddress(addressFields: UpdatableAddressFields) {
         lifecycleScope.launch {
             storage.addAddress(addressFields)
+
+            lifecycleScope.launch(Dispatchers.Main) {
+                navController.popBackStack()
+            }
+        }
+    }
+
+    override fun handleDeleteAddress(guid: String) {
+        lifecycleScope.launch {
+            storage.deleteAddress(guid)
 
             lifecycleScope.launch(Dispatchers.Main) {
                 navController.popBackStack()
--- a/app/src/main/java/org/mozilla/fenix/settings/address/interactor/AddressEditorInteractor.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/address/interactor/AddressEditorInteractor.kt
@@ -25,6 +25,14 @@ interface AddressEditorInteractor {
      * @param addressFields A [UpdatableAddressFields] record to add.
      */
     fun onSaveAddress(addressFields: UpdatableAddressFields)
+
+    /**
+     * Updates the provided address in the autofill storage. Called when a user
+     * taps on the update menu item or "Update" button.
+     *
+     * @param addressFields A [UpdatableAddressFields] record to add.
+     */
+    fun onUpdateAddress(guid: String, addressFields: UpdatableAddressFields)
 }
 
 /**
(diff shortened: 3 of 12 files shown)


app/src/main/java/org/mozilla/fenix/settings/address/view/AddressEditorView.kt
```diff
@@ -36,26 +43,59 @@ class AddressEditorView(
         binding.saveButton.setOnClickListener {
             saveAddress()
         }
+
+        address?.let { address ->
+            binding.saveButton.text =
+                binding.root.context.getString(R.string.addressess_update_address_button)
```
gabrielluong, line 49: Let's not make the save button text change yet. We don't currently do this for "Edit card" so I have noted the inconsistency with UX.

mcarare: Made the change. IMO we should have also changed the button for CC, having it say: "update" or "save changes". It would be more consistent with the editing action. But let's see what UX decides.

gabrielluong: I did mention how it was inconsistent to Kate yesterday. It would be very likely that we want to make the changes everywhere, but the strings aren't final yet. So, that's where we are with this.

app/src/main/java/org/mozilla/fenix/settings/address/AddressEditorFragment.kt
```diff
@@ -26,6 +30,15 @@ class AddressEditorFragment : SecureFragment(R.layout.fragment_address_editor) {
 
     private lateinit var addressEditorView: AddressEditorView
     private lateinit var interactor: AddressEditorInteractor
+    private val args by navArgs<AddressEditorFragmentArgs>()
```
gabrielluong, line 33: Can you add a new line before? I prefer to just keep `lateinit var` and `val` spatially separated for easy parsing

app/src/main/java/org/mozilla/fenix/settings/address/AddressEditorFragment.kt
```diff
@@ -26,6 +30,15 @@ class AddressEditorFragment : SecureFragment(R.layout.fragment_address_editor) {
...
+
+    /**
+     * Returns true if an existing address is being edited, and false otherwise.
+     */
+    private val isEditing: Boolean
+        get() = args.address != null
+
+    private lateinit var menu: Menu
```
gabrielluong, line 41: I don' think we actually needed this `menu` variable. Even in `CreditCardEditorFragment` I didn't think we needed it because it was done so that we can call `menu.close()` in `onPause()`, but the menu never really opens, and I think we just ended up copying code from logins without fully understanding it.

app/src/main/java/org/mozilla/fenix/settings/address/AddressEditorFragment.kt
```diff
@@ -41,18 +54,44 @@ class AddressEditorFragment : SecureFragment(R.layout.fragment_address_editor) {
         )
 
         val binding = FragmentAddressEditorBinding.bind(view)
+        setHasOptionsMenu(true)
 
-        addressEditorView = AddressEditorView(binding, interactor)
+        addressEditorView = AddressEditorView(binding, interactor, args.address)
```
gabrielluong, line 59: Maybe we should pass the address through the `bind()` instead? This seems cleaner to me instead of having a nullable property in `AddressEditorView`. Also, it just follows the current CreditCardEditorView convention.

app/src/main/java/org/mozilla/fenix/settings/address/controller/AddressEditorController.kt
```diff
@@ -57,4 +67,23 @@ class DefaultAddressEditorController(
             }
         }
     }
+
+    override fun handleDeleteAddress(guid: String) {
+        lifecycleScope.launch {
+            storage.deleteAddress(guid)
+            lifecycleScope.launch(Dispatchers.Main) {
```
gabrielluong, line 74: Let's add a new line before this `launch {}` block since you did it below.

app/src/main/java/org/mozilla/fenix/settings/address/view/AddressEditorView.kt
```diff
@@ -36,26 +43,64 @@ class AddressEditorView(
...
         )
+
+        if (address != null) {
+            interactor.onUpdateAddress(address.guid, addressFields)
+        } else {
+            interactor.onSaveAddress(addressFields)
+        }
+    }
+
+    internal fun showConfirmDeleteAddressDialog(context: Context, guid: String) {
```
gabrielluong, line 94: Not quite sure what to think about this. Instinctively, I feel like this should be called through the usual interactor/controller, but I don't know if I have a strong opinion about it.

app/src/main/res/navigation/nav_graph.xml
```diff
@@ -1258,7 +1258,13 @@
         <fragment
             android:id="@+id/addressEditorFragment"
             android:name="org.mozilla.fenix.settings.address.AddressEditorFragment"
-            android:label="@string/addresses_add_address" />
+            android:label="@string/addresses_add_address" >
```
gabrielluong, line 1261:
```suggestion
            android:label="@string/addresses_add_address">
```

app/src/main/res/values/strings.xml
```diff
@@ -1574,6 +1576,16 @@
     <string name="addresses_cancel_button">Cancel</string>
     <!-- The text for the "Delete address" button for deleting an address -->
     <string name="addressess_delete_address_button">Delete address</string>
+    <!-- The title for the "Delete address" confirmation dialog message-->
```
gabrielluong, line 1579:
```suggestion
    <!-- The title for the "Delete address" confirmation dialog -->
```

app/src/main/res/values/strings.xml
```diff
@@ -1574,6 +1576,16 @@
...
+    <string name="addressess_confirm_dialog_message">Are you sure you want to delete this address?</string>
+    <!-- The text for the "Delete address" button in confirmation dialog when deleting an address -->
+    <string name="addressess_confirm_dialog_ok_button">Delete</string>
+    <!-- The text for the "Cancel delete" button in confirmation dialog when deleting an address -->
```
gabrielluong, line 1583: The comments for these strings are a little weird to me because we're saying "Delete address" button since the string for it is "Delete" in the dialog, but I would imagine you're talking about the button in the editor as well, but you also said "Cancel delete" button which is clearly not a thing. 

I would prefer if we stuck to the same language we have used in our previous positive and negative buttons for the dialogs.

app/src/main/java/org/mozilla/fenix/settings/address/interactor/AddressEditorInteractor.kt
```diff
@@ -25,6 +26,22 @@ interface AddressEditorInteractor {
      * @param addressFields A [UpdatableAddressFields] record to add.
      */
     fun onSaveAddress(addressFields: UpdatableAddressFields)
+
+    /**
+     * Deletes the provided address from the autofill storage. Called when a user
+     * taps on the save menu item or "Save" button.
+     *
+     * @param address An [Address] record to delete.
```
gabrielluong, line 34: Wrong @param name and message - this is a `guid`.

app/src/test/java/org/mozilla/fenix/settings/address/controller/DefaultAddressEditorControllerTest.kt
```diff
@@ -73,4 +76,18 @@ class DefaultAddressEditorControllerTest {
             navController.popBackStack()
         }
     }
+
+    @Test
+    fun `GIVEN an existing address record WHEN save address is called THEN update the address record to storage`() = runBlockingTest {
```
gabrielluong, line 81:
```suggestion
    fun `GIVEN an existing address record WHEN save address is called THEN update the existing address record in storage`() = runBlockingTest {
```

gabrielluong (changes requested):

gabrielluong:
Can you squash the strings commit to include it where it's needed? I am mostly thinking about this from a git blame point of view where it's better to see everything related to the dialog in a single commit instead of having the strings in the dialog appearing in a separate commit. 

I also got the sense that we were solving multiple issues in this PR. While I do see we add separate commits for each of the issues, I would also ask that we normally just open new PRs for each of them so that if we ever had to go look up the issues through git blame and into their PR that they are logically separated instead of growing a PR. In this case, I think because everything is in a bigger PR the review comments also blocks you from landing the smaller fixes whenever I think we should try to land smaller PRs as soon as possible because they are easier to review because they are small changes, and avoid making a PR bigger than it needs to be.

app/src/main/java/org/mozilla/fenix/settings/address/AddressEditorFragment.kt
```diff
@@ -26,6 +30,15 @@ class AddressEditorFragment : SecureFragment(R.layout.fragment_address_editor) {
 
     private lateinit var addressEditorView: AddressEditorView
     private lateinit var interactor: AddressEditorInteractor
+    private val args by navArgs<AddressEditorFragmentArgs>()
+
+    /**
+     * Returns true if an existing address is being edited, and false otherwise.
+     */
+    private val isEditing: Boolean
+        get() = args.address != null
```
MatthewTighe, line 39: I think this is out-of-scope for now, as it we're pretty far along in the implementation of this entire feature. Seeing this got me thinking though: why are needing to pass data between simple screens through navigation arguments?

I think we're intending to model the design of this feature after credit cards. However, I believe the intention moving forward is to follow the architecture defined in the `lib-state` module in A-C. 

If we had created a `Store` that was scoped to this feature (using a `StoreProvider` to share it between fragments), it would have allowed us to avoid this kind of thing as well as the kind of delegate-chaining we see in the controller/interactor. The resulting side-effects in those classes could have been moved instead into Middlewares attached to the Store.

No actions needed, just sharing some thoughts

gabrielluong: I believe only `AppState` is the only one that persists in memory. So, creating a new `Store` that is shared between fragments won't actually persist the selected Address between fragments. You would still need to pass the selected Address and inform the newly created `Store` of the state, which gets recreated every time you navigate to the fragment. My current mental model is that a `Store` only lives for as long as the fragment does. However, I can totally be wrong about all of this and I have never looked into this deeply enough to confirm these claims - there are people who can provide an answer if we asked in a team channel, which I think would be a nice thing to do.

app/src/main/java/org/mozilla/fenix/settings/address/AddressEditorFragment.kt
mcarare: The menu opens if the toolbar text is long enough and the device screen is short. I think it will be good if we keep the same pattern everywhere. Also, we could add some tests for it if we keep it.

app/src/main/java/org/mozilla/fenix/settings/address/interactor/AddressEditorInteractor.kt
mcarare: I originally passed the address but forgot to change this when switching to guid.

app/src/main/java/org/mozilla/fenix/settings/address/AddressEditorFragment.kt
mcarare: I prefer to use nav args when data is passed from a single fragment to a "child" fragment ( by child fragment in this context I define a fragment that is accessible only from the one fragment) and the data is not really needed in other fragments /places in the app. We could save all things in a store/state only for the one screen, but that would be overkill IMO.

app/src/main/java/org/mozilla/fenix/settings/address/AddressEditorFragment.kt
mcarare: I would prefer to have the `address` as property because we also use it in other places (the `saveAddress` method) and I would prefer not to pass it around as param from method to method.

app/src/main/java/org/mozilla/fenix/settings/address/view/AddressEditorView.kt
mcarare: I see the pair interactor/controller usage as a bridge between the actions in the interface and the store and sometimes state. Showing a UI piece (a confirmation dialogue in this case) should be the responsibility of the fragment/view IMO. This would also allow for additional testing of the dialog behavior in the future.

app/src/test/java/org/mozilla/fenix/settings/address/controller/DefaultAddressEditorControllerTest.kt
mcarare: Repeating `existing` would be redundant, we already have the adjective in the `GIVEN` part of the test name. 
"to" and "in" are equivalent in this context. You also used "to storage" in the existing `GIVEN a new address record WHEN save address is called THEN save the new address record to storage` test.

mcarare:
> Can you look to squash[...]

I rearranged the code between commits and I moved the strings in the respective commits (Although we have previously pre-landed strings for some features).

I have preferred to have all commits in a single PR because all the work was related and it was based on previous changes. The alternative would have been to start them individually and be blocked by waiting for a review or not wait for a review before starting related work and be forced to constantly rebase work after reviews.

app/src/main/java/org/mozilla/fenix/settings/address/AddressEditorFragment.kt
gabrielluong: Ok, that makes sense to me.

app/src/main/java/org/mozilla/fenix/settings/address/AddressEditorFragment.kt
MatthewTighe: Oh you're right, `StoreProvider.get` is fragment-scoped. My thinking is still being influenced some by my last position, here's some details as to why if you're curious:

The last project I worked on was multi-module, which informed some of the architecture. We typically would have a parent fragment/activity as an entry point to a module, which allowed us to create well-defined entry points as well as share data between child fragments in the module. 

For example, imagine having a `Settings` (or even `AutofillSettings`) module that had a `Store` scoped to the lifecycle of a parent fragment.

We could also theoretically add a `StoreProvider.get` method that used the `HomeActivity` to share stores between fragments, but I'd worry about memory bloat. We'd probably need to think through lifecycle retention pretty deeply

app/src/main/java/org/mozilla/fenix/settings/address/AddressEditorFragment.kt
gabrielluong: That makes sense. Should we add the `menu.close()` handling in `onPause()` in this PR as well? Otherwise, we just have this `menu` that isn't doing anything currently.

app/src/main/java/org/mozilla/fenix/settings/address/view/AddressEditorView.kt
gabrielluong: I generally agree with your stance. Where I would say I differ or don't hold a strong opinion on is the exact scope of what should be handled in the controller - I think handling all `Store` interactions in there makes sense, but otherwise, I would be looking to extract as much functionality out of the Fragment to make functionality easier to tests. Otherwise, I would say the Interactor is called for all user interactions, which is why I am on the fence since this is technically called through a user tapping on the "Delete button", but it's also called through the menu options.

The View in this case does satisfy the fact that we aren't putting the dialog responsibility in the fragment and thus making it easier to test. So, I could've gone either direction.

app/src/test/java/org/mozilla/fenix/settings/address/controller/DefaultAddressEditorControllerTest.kt
```diff
@@ -73,4 +76,18 @@ class DefaultAddressEditorControllerTest {
...
+        val address: Address = mockk()
+        val addressFields: UpdatableAddressFields = mockk()
+        every { address.guid } returns "123"
+
+        controller.handleUpdateAddress(address.guid, addressFields)
+
+        coVerifySequence {
```
gabrielluong, line 88: This is interesting - first time seeing this usage.

gabrielluong (approved):

gabrielluong:
Removing the needs-landing label since it is blocking the AC bump.

Issue #24931, opened 2022-04-22
Display the proper label and description for an Address item


gabrielluong:
> I will file a follow to display the proper label and description. I think we will want to investigate how Desktop formats this.

_Originally posted by @gabrielluong in https://github.com/mozilla-mobile/fenix/pull/24928#discussion_r856524889_

We want to investigate how Desktop formats their Name and Address description and ensure we are following the same format in our address list.

gabrielluong:
Please also link to the Desktop code that does the formatting.

MatthewTighe:
Core of the desktop code can be found [here](https://searchfox.org/mozilla-central/source/toolkit/components/formautofill/FormAutofillUtils.jsm#323)

delia-pop:
Verified as fixed on Nightly 102 from 05/25.

PR #25039, opened 2022-04-29
for #24931: create address label and description based on available data

MatthewTighe:
for #24931

Desktop code is [here](https://searchfox.org/mozilla-central/source/toolkit/components/formautofill/FormAutofillUtils.jsm#323)

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/settings/address/ext/Address.kt
```diff
@@ -0,0 +1,52 @@
...
+ * @property description The subtitled description for an [Address].
+ */
+data class AddressItemText(
+    val label: String,
+    val description: String
+)
+
+private fun Address.getName(): String {
+    val middleName = if (additionalName.isEmpty()) "" else "$additionalName "
+    return if (givenName.isNotEmpty()) "$givenName $middleName$familyName" else familyName
```
gabrielluong, line 33: Just doing a quick scan but are we missing a space between `$middleName$familyName`?

gabrielluong: Oh nvm, I see the space above

app/src/main/java/org/mozilla/fenix/settings/address/ext/Address.kt
```diff
@@ -0,0 +1,52 @@
...
+    streetAddress.joinStreetAddress()
+        .ifEmptyOrPrevious(previous) { addressLevel3 }
+        .ifEmptyOrPrevious(previous) { addressLevel2 }
+        .ifEmptyOrPrevious(previous) { organization }
+        .ifEmptyOrPrevious(previous) { addressLevel1 }
+        .ifEmptyOrPrevious(previous) { country }
+        .ifEmptyOrPrevious(previous) { postalCode }
+        .ifEmptyOrPrevious(previous) { tel }
+        .ifEmptyOrPrevious(previous) { email }
+        .ifEmptyOrPrevious(previous) { "" }
```
MozillaNoah, line 46: would it be possible to get a small kdoc on this? it doesn't come off as immediately obvious what the daisy-chained `ifEmptyOrPrevious` calls is doing (to me anyway).

MozillaNoah, line 46: would it be possible to get a small kdoc on this? it doesn't come off as immediately obvious what the daisy-chained `ifEmptyOrPrevious` calls is doing (to me anyway).

MatthewTighe: I added a comment. Let me know if that makes it more clear 🙂

app/src/main/java/org/mozilla/fenix/settings/address/ext/Address.kt
```diff
@@ -0,0 +1,52 @@
...
+package org.mozilla.fenix.settings.address.ext
+
+import mozilla.components.concept.storage.Address
+
+/**
+ * Generate an [AddressItemText] from a given [Address]. A well-formed [Address] will have
+ * a full name in the label and a street address as a description. If either are not present,
+ * the label and description will be based off lower priority properties.
+ */
+fun Address.toItemText(): AddressItemText {
```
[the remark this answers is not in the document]
MatthewTighe: Happy to move it in as a follow-up if we'd like!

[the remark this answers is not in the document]
MatthewTighe: I separated data class into two extension functions. The one possible issue I see with the difference in implementation is that I think some inherent context is lost. Hopefully the comments will still make it clear, but separating the two pushes the responsibility to the caller of correctly basing the description on the label.

If you like this better, I'll squash the commit and fix the tests up.

app/src/main/java/org/mozilla/fenix/settings/address/ext/Address.kt
```diff
@@ -0,0 +1,44 @@
...
+package org.mozilla.fenix.settings.address.ext
+
+import mozilla.components.concept.storage.Address
+
+/**
+ * Generate a label from a given [Address]. A well-formed [Address] will have
+ * a full name in the label, otherwise it will find the highest priority [Address] property
+ * to use, as based on the desktop code found in: FormAutofillUtils::getAddressLabel
+ */
+fun Address.toLabel(): String = getName().ifEmpty { iterateFields("") }
```
gabrielluong, line 14: Sorry, still nitpicking on the naming here since it is still tying to the UI components. Can we do something like `toFullName` or `getFullName`? Looking for the naming to just be generic with a semantic name for any general purpose.

app/src/main/java/org/mozilla/fenix/settings/address/ext/Address.kt
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Generate a description from a given [Address]. A well-formed address will have a street address
+ * as a description. An address with missing data will use the next available property after [label]
+ * based on the priorities found in the desktop code in: FormAutofillUtils::getAddressLabel
```
gabrielluong, line 19: Can you also add link to the code?

app/src/main/java/org/mozilla/fenix/settings/address/ext/Address.kt
```diff
@@ -0,0 +1,44 @@
...
+ */
+fun Address.toDescription(label: String): String = iterateFields(label)
+
+private fun Address.getName(): String {
+    val middleName = if (additionalName.isEmpty()) "" else "$additionalName "
+    return if (givenName.isNotEmpty()) "$givenName $middleName$familyName" else familyName
```
gabrielluong, line 25: Could we simplify this to something like `[givenName, middleName, familyName].join(" ")`

app/src/test/java/org/mozilla/fenix/settings/address/ext/AddressKtTest.kt
```diff
@@ -0,0 +1,145 @@
...
+            """.trimIndent()
+        )
+
+        val result = addr.toItemText()
+
+        assertEquals("line1 line2 line3", result.description)
+    }
+
+    private fun generateAddress(
+        given: Boolean = true,
```
gabrielluong, line 111: Couldn't we have just specify the default string values for all of these parameters instead of using booleans?

app/src/test/java/org/mozilla/fenix/settings/address/ext/AddressKtTest.kt
```diff
@@ -0,0 +1,145 @@
...
+        family: Boolean = true,
+        org: Boolean = true,
+        streetAddr: Boolean = true,
+        addr3: Boolean = true,
+        addr2: Boolean = true,
+        addr1: Boolean = true,
+        postalCode: Boolean = true,
+        country: Boolean = true,
+        tel: Boolean = true,
+        email: Boolean = true
```
gabrielluong, line 122: If possible, I would just align the parameter names here with the `Address` parameter names

app/src/test/java/org/mozilla/fenix/settings/address/ext/AddressKtTest.kt
```diff
@@ -0,0 +1,145 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.settings.address.ext
+
+import mozilla.components.concept.storage.Address
+import org.junit.Assert.assertEquals
+import org.junit.Test
+
+class AddressKtTest {
```
gabrielluong, line 11:
```suggestion
class AddressTest {
```

app/src/main/java/org/mozilla/fenix/settings/address/ext/Address.kt
```diff
@@ -0,0 +1,44 @@
...
+}
+
+private fun Address.iterateFields(previous: String): String =
+    streetAddress.joinStreetAddress()
```
gabrielluong, line 29: `[streetAddress, addressLevel3, addressLevel2, organization, addressLevel1, ..].map(field -> field.isNotEmpty()).join(", ")` 

I think this will also simplify your current approach

app/src/main/java/org/mozilla/fenix/settings/address/ext/Address.kt
```diff
@@ -0,0 +1,44 @@
...
+ * to use, as based on the desktop code found in: FormAutofillUtils::getAddressLabel
+ */
+fun Address.toLabel(): String = getName().ifEmpty { iterateFields("") }
+
+/**
+ * Generate a description from a given [Address]. A well-formed address will have a street address
+ * as a description. An address with missing data will use the next available property after [label]
+ * based on the priorities found in the desktop code in: FormAutofillUtils::getAddressLabel
+ */
+fun Address.toDescription(label: String): String = iterateFields(label)
```
gabrielluong, line 21: Maybe `toAddressLabel`?

app/src/main/java/org/mozilla/fenix/settings/address/ext/Address.kt
MatthewTighe: It would be theoretically possible with an implementation like that to end up with a label that was `middleName familyName`. That's what I was trying to guard against with this slightly awkward logic, but if we aren't concerned about that case then I'm happy to update it.

gabrielluong: Can you find out what the desktop code is doing for the name label? Your current link only links to the address label generation. We want to emulate as much of the desktop code as possible to account for all the various edge cases when it comes to addresses since we aren't really the expert here.

app/src/main/java/org/mozilla/fenix/settings/address/ext/Address.kt
```diff
@@ -0,0 +1,44 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.settings.address.ext
+
+import mozilla.components.concept.storage.Address
+
+/**
+ * Generate a label from a given [Address]. A well-formed [Address] will have
+ * a full name in the label, otherwise it will find the highest priority [Address] property
+ * to use, as based on the desktop code found in: FormAutofillUtils::getAddressLabel
```
gabrielluong, line 12: `FormAutofillUtils::getAddressLabel` doesn't tell us what they do with the name so, let's find out where that is done and validate our approach

MatthewTighe: Just testing it manually it looks like it does allow for `middleName lastName`, so I'll simplify my implementation to match

app/src/main/java/org/mozilla/fenix/settings/address/ext/Address.kt
```diff
@@ -0,0 +1,44 @@
...
+        .ifEmptyOrPrevious(previous) { organization }
+        .ifEmptyOrPrevious(previous) { addressLevel1 }
+        .ifEmptyOrPrevious(previous) { country }
+        .ifEmptyOrPrevious(previous) { postalCode }
+        .ifEmptyOrPrevious(previous) { tel }
+        .ifEmptyOrPrevious(previous) { email }
+        .ifEmptyOrPrevious(previous) { "" }
+
+private fun String.joinStreetAddress(): String =
+    this.split("\n").joinToString(separator = " ") { it.trim() }
```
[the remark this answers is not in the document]
MatthewTighe: This will trim the split parts. `joinToString` will join the result of the block defined here as `it.trim` with the `separator` parameter.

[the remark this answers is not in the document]
gabrielluong: I know we have a test case for something that is like this already, but I think it also makes sense to just call this function individually since I found a whole suite of test cases for this function alone in https://searchfox.org/mozilla-central/source/browser/extensions/formautofill/test/unit/test_toOneLineAddress.js.

It might make sense to also just use the same naming as desktop.

app/src/test/java/org/mozilla/fenix/settings/address/ext/AddressKtTest.kt
MatthewTighe: Yeah, that's definitely better. Thanks for the suggestion

gabrielluong: Gonna request that we just allow that this function takes in string values for each of the parameter. This type of function can generally be reused in many other tests so it would be best if it was robust.

app/src/main/java/org/mozilla/fenix/settings/address/ext/Address.kt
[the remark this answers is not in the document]
MatthewTighe: Yeah, I can do that. I only structured it that way because I generally try to avoid `@VisibleForTesting`, especially in single module projects.

app/src/main/java/org/mozilla/fenix/settings/address/ext/Address.kt
gabrielluong: I am okay with not implementing the addressField override for now, but I do want us to investigate to see if we should ever be concern about it to have our bases covered. This does require some digging into individual bugzilla issues, code, tests, whatever you can find. I don't believe the original authors are at Mozilla anymore, but it will help if we did our due diligence based on anything we can find. If we did discover we needed this, I would file a separate issue to ensure we have that on file. 

> My implementation also differs in that it returns the 2 parts (label and description) separately since that's what the caller anticipates. Do you think it makes more sense to return it as a single string and allow the caller to split it?

Desktop returns their label all in one line compare to our desired mobile design. So, it's totally okay for us to separate this. Based on my simple test of adding my own address and looking at https://searchfox.org/mozilla-central/source/toolkit/components/formautofill/FormAutofillUtils.jsm#323, I have to imagine they are using the addressField override for different regions since I only see my `<Full Name>, <One line Street Address>` (no country, state, phone, email that you can see in the linked code).

gabrielluong: There's a lot of things to parse through here, but as a way forward we can keep to adding all the fields or we could simplify this to displaying the `<One line Street Address>` and follow up based on the investigation of how desktop is displaying the address label for US/CA.

MatthewTighe: > I have to imagine they are using the addressField override for different regions since I only see my <Full Name>, <One line Street Address> (no country, state, phone, email that you can see in the linked code).

I believe that's because of [this line](https://searchfox.org/mozilla-central/source/toolkit/components/formautofill/FormAutofillUtils.jsm#356). It seems to use the first two, highest priority fields available. You can see that in action if you enter addresses without a name or one line address:

image

This is what I've tried to emulate on the mobile side, where `label` and `description` are the elements on desktop that are comma separated.

> This does require some digging into individual bugzilla issues, code, tests, whatever you can find. I don't believe the original authors are at Mozilla anymore, but it will help if we did our due diligence based on anything we can find.

This feels like it could be a pretty large undertaking. I think we should consider filing a separate spike if it's important enough to do that level of discovery for. Searching for usages of [getAddressLabel](https://searchfox.org/mozilla-central/search?q=getAddressLabel&path=) shows that the `addressFields` parameter defining the override isn't currently used. Would it maybe be easier to just track down a desktop PM and check if it's a feature they intend to continue supporting?

app/src/main/java/org/mozilla/fenix/settings/address/ext/Address.kt
MatthewTighe: I believe the `getAddressLabel` function includes the `name` parameter of the `address` object they're using on desktop, if that's what you're asking. If you're looking for source on where they split and combine the `name` property into `given`, `additional`, and `family`, I've been struggling to track it down. Haven't used JS for about 4 years so I'm shaking some of the rust off 😅 

I can keep looking for it some tomorrow if needed

app/src/main/java/org/mozilla/fenix/settings/address/ext/Address.kt
gabrielluong: > I believe that's because of [this line](https://searchfox.org/mozilla-central/source/toolkit/components/formautofill/FormAutofillUtils.jsm#356). It seems to use the first two, highest priority fields available. You can see that in action if you enter addresses without a name or one line address:
> 

Ok this line does make sense based on what we are seeing. I am kinda satisfied with that wrt to the addressFields override, and it doesn't look like it is used based on my quick search. So, I think we can move on from this topic.

app/src/main/java/org/mozilla/fenix/settings/address/ext/Address.kt
MatthewTighe: I finally found the name combination and splitting [here](https://searchfox.org/mozilla-central/source/toolkit/components/formautofill/FormAutofillNameUtils.jsm#400). It looks like the combination is the a simple `join` on non-empty parts, as is reflected in this PR. Splitting the name seems like it will be important for saving addresses though.

MatthewTighe:
It sounds like we're close to alignment on the field overrides, names, and the aspects of the address which should be included in the label and description, so I think this could be ready for another review pass.

app/src/main/java/org/mozilla/fenix/settings/address/ext/Address.kt
```diff
@@ -0,0 +1,56 @@
...
+import androidx.annotation.VisibleForTesting
+import mozilla.components.concept.storage.Address
+
+/**
+ * Generate a label from a given [Address]. A well-formed [Address] will have
+ * a full name in the label, otherwise it will find the highest priority [Address] property
+ * to use, as based on the desktop code found in FormAutofillUtils::getAddressLabel
+ * https://searchfox.org/mozilla-central/rev/87ecd21d3ca517f8d90e49b32bf042a754ed8f18/toolkit/components/formautofill/FormAutofillUtils.jsm#323
+ */
+fun Address.toAddressLabel(): String = getFullName().ifEmpty { iterateFields("") }
```
gabrielluong, line 16: I feel we might be overthinking this with trying to generate a well formed label and description depending on what fields are available. The UI here is very similar to Chrome's Addresses where it only shows the full name in the label and all other address field appears as the description.

For the time being to get this all landed, I think we should simply call `getFullName` for the label and drop `toAddressLabel`. 

If we did decide to go with your methodology, I think I would like to see something very explicitly stated in a issue and okay'd by UX/Product.

gabrielluong (changes requested):

app/src/main/java/org/mozilla/fenix/settings/address/ext/Address.kt
MatthewTighe: 🤦 I'd been basing my work on what I saw on desktop instead of the current designs. I'll update all this accordingly today. Thanks for your patience in pointing that out

gabrielluong: Okay, the one reasoning why I was kinda strict about looking at seeing what Desktop was doing was to understand if and how they did the address ordering for different regions. It was better that we understood what was done on Desktop and be informed as we did our own mobile implementation. 

Any implicit requirements that we might get from Desktop should be brought up with UX/Product if we felt it needed further consideration.

app/src/main/java/org/mozilla/fenix/settings/address/ext/Address.kt
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Generate a description from a given [Address]. A well-formed address will have a street address
+ * as a description. An address with missing data will use the next available property after [label]
+ * based on the priorities found in the desktop code in FormAutofillUtils::getAddressLabel
+ * https://searchfox.org/mozilla-central/rev/87ecd21d3ca517f8d90e49b32bf042a754ed8f18/toolkit/components/formautofill/FormAutofillUtils.jsm#323
+ *
+ * @param label: The label for an [Address]. See [toAddressLabel].
+ */
+fun Address.toAddressDescription(label: String): String = iterateFields(label)
```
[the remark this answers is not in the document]
MatthewTighe: The only other concern I have about dropping the `previous` handling is planning for when addresses are syncable. I'm happy to ignore it for now, but there's a chance we will be receiving addresses without `name`s for the label in the future. I'm not sure what that would mean from a UX perspective, but I think it's safe to punt on for now.

MatthewTighe:
Needed to add some handling for longer descriptions in our text items, since the combination of fields can really easily reach the max size. This is on a larger, almost tablet-sized device for reference:
image

app/src/main/java/org/mozilla/fenix/compose/list/ListItem.kt
```diff
@@ -51,6 +51,7 @@ fun TextListItem(
     label: String,
     modifier: Modifier = Modifier,
     description: String? = null,
+    maxDescriptionLines: Int = 1,
```
gabrielluong, line 54: Can add a @param for this in `TextListItem ` and `ListItem` below? I would just use a similar description for `maxLines` in `SecondaryText` - `An optional maximum number of lines for the description text to span.`

app/src/main/java/org/mozilla/fenix/settings/address/ext/Address.kt
```diff
@@ -0,0 +1,38 @@
...
+
+import androidx.annotation.VisibleForTesting
+import mozilla.components.concept.storage.Address
+
+/**
+ * Generate a label item text for an [Address]. The combination of names is based on desktop code
+ * found here:
+ * https://searchfox.org/mozilla-central/rev/d989c65584ded72c2de85cb40bede7ac2f176387/toolkit/components/formautofill/FormAutofillNameUtils.jsm#400
+ */
+fun Address.getAddressLabel(): String = listOf(givenName, additionalName, familyName)
```
gabrielluong, line 15:
```suggestion
fun Address.getFullName(): String = listOf(givenName, additionalName, familyName)
```

I am gonna sound like a broken record here because I have said this from the first review, but what I still want for our extension function naming here is to avoid tying the naming to the UI element. Semantically, this is just returning the full name. The reasoning I am very persistent on this naming is because most (if not all) of our extension functions don't really prescribe to UI element labelling, but towards what is actually is.

I am also anticipating we will be using this in the select address prompt, but there's always the possibility that we use these extension functions beyond just for the list item UI. Having a full name field in the data structure is very much also a possibility when you look beyond https://searchfox.org/mozilla-central/source/mobile/android/geckoview/src/main/java/org/mozilla/geckoview/Autocomplete.java#465

app/src/main/java/org/mozilla/fenix/settings/address/ext/Address.kt
```diff
@@ -0,0 +1,38 @@
...
+    .filter { it.isNotEmpty() }
+    .joinToString(" ")
+
+/**
+ * Generate a description item text for an [Address]. The element ordering is based on the
+ * priorities defined by the desktop code found here:
+ * https://searchfox.org/mozilla-central/rev/d989c65584ded72c2de85cb40bede7ac2f176387/toolkit/components/formautofill/FormAutofillUtils.jsm#323
+ */
+fun Address.getAddressDescription(): String = listOf(
```
gabrielluong, line 24:
```suggestion
fun Address.getAddressLabel(): String = listOf(
```
This one is to just follow the desktop naming.

gabrielluong (changes requested):

gabrielluong:
Because of https://searchfox.org/mozilla-central/source/mobile/android/geckoview/src/main/java/org/mozilla/geckoview/Autocomplete.java#465 and since we already calling the builder in https://github.com/mozilla-mobile/android-components/pull/12101, I think we should just move this to AC.

app/src/main/java/org/mozilla/fenix/settings/address/ext/Address.kt
MatthewTighe: Oh that's a good point. I was a little confused because the suggestion will ultimately mean this line in `AddressList`:
`description = address.getAddressLabel()`, but now that you've explained it I see what you're going for. Thanks for your patience on this point, I appreciate the explanation!

app/src/main/java/org/mozilla/fenix/compose/list/ListItem.kt
MatthewTighe: Thanks for the reminder!

app/src/main/java/org/mozilla/fenix/compose/list/ListItem.kt
```diff
@@ -60,6 +62,7 @@ fun TextListItem(
         label = label,
         modifier = modifier,
         description = description,
+        maxDescriptionLines,
```
gabrielluong, line 65:
```suggestion
        maxDescriptionLines = maxDescriptionLines,
```

gabrielluong (approved):

PR #25123, opened 2022-05-05
For #25119 - Add Secondary, Tertiary and Destructive Button composable components

### LINKED ISSUE #25119: Add Secondary, Tertiary and Destructive Button composable components
┆Issue is synchronized with this [Jira Story](https://mozilla-hub.atlassian.net/browse/FNXV2-20382)

gabrielluong:
Fixes #25119

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/compose/button/Button.kt
+++ b/app/src/main/java/org/mozilla/fenix/compose/button/Button.kt
@@ -4,25 +4,23 @@
 
 package org.mozilla.fenix.compose.button
 
+import android.content.res.Configuration
 import androidx.compose.foundation.background
-import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.width
-import androidx.compose.foundation.shape.RoundedCornerShape
-import androidx.compose.material.Button
 import androidx.compose.material.ButtonDefaults
 import androidx.compose.material.Icon
 import androidx.compose.material.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
-import androidx.compose.ui.draw.clip
+import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.painter.Painter
 import androidx.compose.ui.res.painterResource
-import androidx.compose.ui.res.stringResource
-import androidx.compose.ui.text.font.Font
-import androidx.compose.ui.text.font.FontFamily
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.sp
@@ -31,10 +29,59 @@ import org.mozilla.fenix.theme.FirefoxTheme
 import org.mozilla.fenix.theme.Theme
 
 /**
+ * Base component for buttons.
+ *
+ * @param text The button text to be displayed.
+ * @param textColor [Color] to apply to the button text.
+ * @param backgroundColor The background [Color] of the button.
+ * @param icon Optional [Painter] used to display a [Icon] before the button text.
+ * @param tint Tint [Color] to be applied to the icon.
+ * @param onClick Invoked when the user clicks on the button.
+ */
+@Composable
+fun Button(
+    text: String,
+    textColor: Color,
+    backgroundColor: Color,
+    icon: Painter? = null,
+    tint: Color,
+    onClick: () -> Unit,
+) {
+    androidx.compose.material.Button(
+        onClick = onClick,
+        modifier = Modifier
+            .fillMaxWidth()
+            .height(36.dp),
+        elevation = ButtonDefaults.elevation(defaultElevation = 0.dp, pressedElevation = 0.dp),
+        colors = ButtonDefaults.outlinedButtonColors(
+            backgroundColor = backgroundColor,
+        ),
+    ) {
+        icon?.let { painter ->
+            Icon(
+                painter = painter,
+                contentDescription = null,
+                tint = tint,
+            )
+
+            Spacer(modifier = Modifier.width(8.dp))
+        }
+
+        Text(
+            text = text,
+            color = textColor,
+            fontSize = 14.sp,
+            letterSpacing = 0.25.sp,
+            maxLines = 1,
+        )
+    }
+}
+
+/**
  * Primary button.
  *
  * @param text The button text to be displayed.
- * @param icon Optional button icon to be displayed.
+ * @param icon Optional [Painter] used to display an [Icon] before the button text.
  * @param onClick Invoked when the user clicks on the button.
  */
 @Composable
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/mozilla/fenix/compose/button/Button.kt
```diff
@@ -31,40 +31,46 @@ import org.mozilla.fenix.theme.FirefoxTheme
...
+    backgroundColor: Color,
     icon: Painter? = null,
+    iconTint: Color,
     onClick: () -> Unit,
 ) {
-    Button(
+    androidx.compose.material.Button(
         onClick = onClick,
         modifier = Modifier
             .clip(RoundedCornerShape(size = 4.dp))
```
MozillaNoah, line 55: I'm seeing `RoundedCornerShape` in several places now. Long term, we may want to have a top-level Shapes Object which provides a one-stop-shop for things like this.

app/src/main/java/org/mozilla/fenix/compose/button/Button.kt
```diff
@@ -31,40 +31,46 @@ import org.mozilla.fenix.theme.FirefoxTheme
...
             .fillMaxWidth()
             .height(36.dp),
         elevation = ButtonDefaults.elevation(defaultElevation = 0.dp, pressedElevation = 0.dp),
-        colors = ButtonDefaults.outlinedButtonColors(backgroundColor = FirefoxTheme.colors.actionPrimary),
+        colors = ButtonDefaults.outlinedButtonColors(backgroundColor = backgroundColor),
     ) {
         if (icon != null) {
```
MozillaNoah, line 61: since it's just one null check, we could optionally do a `?.let`

app/src/main/java/org/mozilla/fenix/compose/button/Button.kt
```diff
@@ -31,40 +31,46 @@ import org.mozilla.fenix.theme.FirefoxTheme
...
             Spacer(modifier = Modifier.width(8.dp))
         }
 
         Text(
             text = text,
-            color = FirefoxTheme.colors.textOnColorPrimary,
+            color = textColor,
             fontSize = 14.sp,
             fontFamily = FontFamily(Font(R.font.metropolis_semibold)),
             letterSpacing = 0.sp,
```
MozillaNoah, line 76: the specs seem to be using `0.25sp`

app/src/main/java/org/mozilla/fenix/compose/button/Button.kt
```diff
@@ -73,6 +79,98 @@ fun PrimaryButton(
     }
 }
 
+/**
+ * Primary button.
+ *
+ * @param text The button text to be displayed.
+ * @param icon Optional [Painter] used to display a [Icon] before the button text.
```
MozillaNoah, line 86:
```suggestion
 * @param icon Optional [Painter] used to display an [Icon] before the button text.
```

app/src/main/java/org/mozilla/fenix/compose/button/Button.kt
```diff
@@ -86,3 +184,45 @@ private fun PrimaryButtonPreview() {
...
+    }
+}
+
+@Composable
+@Preview
+private fun DestructiveButtonPreview() {
+    FirefoxTheme(theme = Theme.getTheme(isPrivate = false)) {
+        Box(Modifier.background(FirefoxTheme.colors.layer1)) {
+            DestructiveButton(
+                text = stringResource(R.string.recent_tabs_show_all),
```
MozillaNoah, line 222: nit: should this be a destructive string? 😛

app/src/main/java/org/mozilla/fenix/compose/button/Button.kt
```diff
@@ -31,40 +31,46 @@ import org.mozilla.fenix.theme.FirefoxTheme
...
+    iconTint: Color,
     onClick: () -> Unit,
 ) {
-    Button(
+    androidx.compose.material.Button(
         onClick = onClick,
         modifier = Modifier
             .clip(RoundedCornerShape(size = 4.dp))
             .fillMaxWidth()
             .height(36.dp),
```
MozillaNoah, line 57: still sad about the fixed height, but it'll require some more research to find an appropriate solution.

MozillaNoah (commented):
this looks great!

app/src/main/java/org/mozilla/fenix/compose/button/Button.kt
```diff
@@ -86,3 +184,45 @@ private fun PrimaryButtonPreview() {
         }
     }
 }
+
+@Composable
+@Preview
```
MozillaNoah, line 189: nit nit: could the previews be in dark mode? I noticed that the rounded corners were hard to notice on the other buttons otherwise.

app/src/main/java/org/mozilla/fenix/compose/button/Button.kt
```diff
@@ -31,40 +31,46 @@ import org.mozilla.fenix.theme.FirefoxTheme
...
     icon: Painter? = null,
+    iconTint: Color,
     onClick: () -> Unit,
 ) {
-    Button(
+    androidx.compose.material.Button(
         onClick = onClick,
         modifier = Modifier
             .clip(RoundedCornerShape(size = 4.dp))
             .fillMaxWidth()
```
MozillaNoah, line 56: will there ever be a scenario where we don't want a button to fill the max width? Just wondering if this modifier attribute should be one that's passed in via a `Modifier` parameter.

gabrielluong: I do not believe so.

app/src/main/java/org/mozilla/fenix/compose/button/Button.kt
gabrielluong: Agreed, we should have something that feeds into `FirefoxTheme` similar to `MaterialTheme.shapes`. It's not totally clear what they all should be right now or what they should be named, but maybe as the component system is developed it will become clearer.

gabrielluong: It turns out we don't need to clip because of the Button defaults to `RoundedCornerShape(size = 4.dp)` https://cs.android.com/androidx/platform/tools/dokka-devsite-plugin/+/master:testData/compose/source/androidx/compose/material/Button.kt;l=99?q=Button.

app/src/main/java/org/mozilla/fenix/compose/button/Button.kt
```diff
@@ -31,55 +29,213 @@ import org.mozilla.fenix.theme.FirefoxTheme
...
- * @param icon Optional button icon to be displayed.
+ * @param contentColor [Color] to apply to the button text.
+ * @param backgroundColor The background [Color] of the button.
+ * @param icon Optional [Painter] used to display a [Icon] before the button text.
+ * @param tint Tint [Color] to be applied to the icon.
  * @param onClick Invoked when the user clicks on the button.
  */
 @Composable
-fun PrimaryButton(
+fun Button(
```
MozillaNoah, line 42: should the base component be private?

app/src/main/java/org/mozilla/fenix/compose/button/Button.kt
```diff
@@ -31,55 +29,213 @@ import org.mozilla.fenix.theme.FirefoxTheme
...
             )
 
             Spacer(modifier = Modifier.width(8.dp))
         }
 
         Text(
             text = text,
-            color = FirefoxTheme.colors.textOnColorPrimary,
             fontSize = 14.sp,
-            fontFamily = FontFamily(Font(R.font.metropolis_semibold)),
```
MozillaNoah, line 69: why were `color` and `fontFamily` removed?

app/src/main/java/org/mozilla/fenix/compose/button/Button.kt
```diff
@@ -31,55 +29,213 @@ import org.mozilla.fenix.theme.FirefoxTheme
...
+        onClick = onClick,
+    )
+}
+
 @Composable
 @Preview
-private fun PrimaryButtonPreview() {
+private fun ButtonLightThemePreview() {
     FirefoxTheme(theme = Theme.getTheme(isPrivate = false)) {
-        Box(Modifier.background(FirefoxTheme.colors.layer1)) {
```
MozillaNoah, line 80: Alternatively, you can have one preview function but then do two `@Preview` calls:

`@Preview(uiMode = Configuration.UI_MODE_NIGHT_YES)`
`@Preview(uiMode = Configuration.UI_MODE_NIGHT_NO)`

app/src/main/java/org/mozilla/fenix/home/recentsyncedtabs/view/RecentSyncedTab.kt
```diff
@@ -134,30 +131,20 @@ fun RecentSyncedTab(
...
+                    stringResource(R.string.recent_tabs_see_all_synced_tabs_button_text)
+                } else {
+                    ""
+                },
+                contentColor = FirefoxTheme.colors.textActionSecondary,
+                backgroundColor = if (tab == null) {
+                    FirefoxTheme.colors.layer3
+                } else {
+                    FirefoxTheme.colors.actionSecondary
+                },
```
MozillaNoah, line 144: are these null checks just effectively hiding the button?

MozillaNoah:
It looks great though locally! I did notice, however, that the previews in `RecentSyncedTab` aren't building. This could be follow-up though if it's unrelated to the button refactor.

image

app/src/main/java/org/mozilla/fenix/compose/button/Button.kt
gabrielluong: `color` is replaced by the `contentColor`. Removing the `fontFamily` was a flip decision based on the 0.25.sp letterSpacing feedback since the Figma component is technically using 0.25.sp letterSpacing with Roboto whereas our current xml component is using Metropolis Semibold with 0.sp letterSpacing. 

I mostly based my decision here based on the fact that most of the Home screen (most recent was the "See all" link) has already started to convert to using Roboto (system font), and I am looking to make sure we are moving forward with the issue to remove Metropolis.

gabrielluong:
> It looks great though locally! I did notice, however, that the previews in `RecentSyncedTab` aren't building. This could be follow-up though if it's unrelated to the button refactor.

This is because of the `ThumbnailCard` which will eventually be fixed in Arturo's tab list view PR.

app/src/main/java/org/mozilla/fenix/compose/button/Button.kt
gabrielluong: This is currently being used in `RecentSyncedTab` for the Button skeleton, but otherwise TBD.

MozillaNoah (approved):

PR #25206, opened 2022-05-11
DRAFT - Identify the percentage of how much a composable is showing

Mugurell:
Draft for now.
More details about the needed functionality available in https://mozilla-hub.atlassian.net/browse/FNXV2-19878
Possibly affected by https://github.com/mozilla-mobile/fenix/issues/22402

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesComposables.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesComposables.kt
@@ -6,7 +6,10 @@
 
 package org.mozilla.fenix.home.pocket
 
+import android.content.Context
+import android.graphics.Rect
 import android.net.Uri
+import androidx.annotation.FloatRange
 import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
@@ -23,9 +26,19 @@ import androidx.compose.foundation.lazy.rememberLazyListState
 import androidx.compose.material.Icon
 import androidx.compose.material.Text
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.composed
 import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.toAndroidRect
+import androidx.compose.ui.layout.LayoutCoordinates
+import androidx.compose.ui.layout.boundsInWindow
+import androidx.compose.ui.layout.onGloballyPositioned
+import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.platform.LocalDensity
 import androidx.compose.ui.res.painterResource
 import androidx.compose.ui.res.stringResource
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesComposables.kt
```diff
@@ -150,6 +182,52 @@ fun PocketStories(
     }
 }
 
+private fun Modifier.onShown(
```
MozillaNoah, line 185: should this live in the `Modifier` extension function file?

app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesComposables.kt
```diff
@@ -150,6 +182,52 @@ fun PocketStories(
...
+            }
+        }
+    }
+}
+
+private fun LayoutCoordinates.isVisible(
+    context: Context,
+    @FloatRange(from = 0.0, to = 1.0) threshold: Float,
+): Boolean {
+    if (!isAttached) return false
```
MozillaNoah, line 209: will this work if a story Composable is in a Lazy list and is rendered but just off screen? Curious at what point this function returns `true`, basically.

app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesComposables.kt
```diff
@@ -150,6 +182,52 @@ fun PocketStories(
...
+    onVisible: () -> Unit,
+): Modifier {
+    return composed {
+        val context = LocalContext.current
+        var wasEventReported by remember { mutableStateOf(false) }
+
+        onGloballyPositioned { coordinates ->
+            @Suppress("CollapsibleIfStatements")
+            if (!wasEventReported) {
+                if (coordinates.isVisible(context, threshold)) {
```
MozillaNoah, line 196: how come the `@Suppress` here?

app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesComposables.kt
MozillaNoah, line 185: are there any performance concerns with this running a bunch of calculations on the UI thread?

MozillaNoah (commented):
drive-by. this looks pretty interesting though! I'm excited to see what becomes of this, since I know this has always been a tricky problem to solve on Android.

app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesComposables.kt
Mugurell: This is exactly the intended scenario - report a composable as shown/visible only if it is currently in the visible bounds of the screen, otherwise if it offscreen this will return false.

app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesComposables.kt
Mugurell: When trying different approaches separate `if`s were easier to reason about. Will join them together in the final version.

app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesComposables.kt
Mugurell: Pretty much all calculations will be avoided with the help of `if (!wasEventReported) ...` once the composable was visible. Before that, while the screen is scrolled this `onGloballyPositioned` will fire multiple times but I don't expect any visible performance impact as there are just a few, simple calculations.

app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesComposables.kt
Mugurell: An interesting aspect is that since not everything on the screen is part of the same (or parent) composable this will not know if another view is shown on top of it (like it happens with the toolbar shown on homescreen). This will unfortunately mean that that data has to be fed in separately.

Mugurell:
Closing this in favor of https://github.com/mozilla-mobile/fenix/pull/25290 where this was used.
Thank you for the help @MozillaNoah !

PR #25236, opened 2022-05-13
Fixes #24917: add country dropdown to address editor

### LINKED ISSUE #24917: Add a Country dropdown menu list in the Address Editor
Replace the existing country input text field with a dropdown of country values.

Examine how the Desktop Country dropdown is generated (could be this https://searchfox.org/mozilla-central/source/toolkit/components/formautofill/addressmetadata/addressReferences.js but should verify). We're assuming we want to implement a similar Country dropdown instead of having users manually enter their country, but we can also limit this to only US/CA since we are only supporting those countries initially. Ideally, we should see if Android provides something for us to validate addresses or provides a list of countries as well. 

Acceptance Criteria
1. Ensure we are displaying the same exact Country dropdown list as desktop and the expected Country values that AS accepts in their storage (checking for whether or not it's the short code such as US, case sensitivity?)
2. We should autofill the user's country based on their region when they navigate to the address editor.
3. If we have to load a json such as https://searchfox.org/mozilla-central/source/toolkit/components/formautofill/addressmetadata/addressReferences.js which contains country data and various validation metadata, we should probably file a new ticket and possibly do this in `support-utils` in AC. Please confer findings before proceeding. 

No design was available at time of writing.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-20248)

MatthewTighe:
Fixes #24917

The intent with all the `AddressCountry` stuff was to be able to extend it based on future requirements. 

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/settings/address/AddressEditorFragment.kt
```diff
@@ -100,4 +106,12 @@ class AddressEditorFragment : SecureFragment(R.layout.fragment_address_editor) {
         }
         else -> false
     }
+
+    companion object {
+        val defaultCountry = AddressCountry("US", "UNITED STATES")
+        val availableCountries = mapOf(
```
MozillaNoah, line 112: nit: should these be all-caps?

app/src/main/java/org/mozilla/fenix/settings/address/view/AddressEditorState.kt
```diff
@@ -0,0 +1,51 @@
...
+data class AddressEditorState(
+    val country: AddressCountry,
+)
+
+/**
+ * Value type representing properties determined by the country used in an [Address].
+ * This data is meant to mirror the data currently represented on desktop here:
+ * https://searchfox.org/mozilla-central/source/toolkit/components/formautofill/addressmetadata/addressReferences.js
+ */
+data class AddressCountry(
```
MozillaNoah, line 27: will all of the state variable data types be prefaced with `Address`? Might seem silly with `AddressAddressLine2` or something.

app/src/main/res/layout/fragment_address_editor.xml
```diff
@@ -324,6 +324,41 @@
 
         </com.google.android.material.textfield.TextInputLayout>
 
+        <!-- Country -->
+        <LinearLayout
+            android:id="@+id/country_layout"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:layout_marginTop="10dp"
+            android:paddingStart="3dp"
```
MozillaNoah, line 333: 3dp seems pretty specific. is this from a spec or just to mirror the credit card editor layout?

app/src/main/res/layout/fragment_address_editor.xml
```diff
@@ -324,6 +324,41 @@
...
+
+            <androidx.appcompat.widget.AppCompatSpinner
+                android:id="@+id/country_drop_down"
+                android:layout_width="match_parent"
+                android:layout_height="wrap_content" />
+
+            <View
+                android:layout_width="match_parent"
+                android:layout_height="1dp"
+                android:background="?attr/textPrimary" />
```
MozillaNoah, line 358: is this supposed to be an underline? Would border primary be a better choice?

app/src/main/java/org/mozilla/fenix/settings/address/view/AddressEditorState.kt
MatthewTighe: This value type isn't meant to represent a field of an address. If you have suggestions for clarity I'm totally open.

This was intended to represent the various formatting and other related data that we get from the keys in [this file on desktop](https://searchfox.org/mozilla-central/source/toolkit/components/formautofill/addressmetadata/addressReferences.js). There's things in there like available states/provinces, names of fields for the address form, required fields for validity, etc. 

The intent was to be able to expand this class as we brought more of those properties in. The loose plan is to create a static binding to that data that is generated during builds and expose it as an object in GV. This class may even be replaced completely by that object once it's available.

app/src/main/res/layout/fragment_address_editor.xml
MatthewTighe: This is just mirroring what's in the credit card editor. I believe the plan is to migrate it to spec/compose after MVP is hit.

app/src/main/res/layout/fragment_address_editor.xml
MatthewTighe: I just borrowed this directly from the credit card editor. The text is intended to be black in light mode (for the current design), so `formDefault` is what I think we're looking for. I'll update that

app/src/main/java/org/mozilla/fenix/settings/address/AddressEditorFragment.kt
MatthewTighe: I think caps might usually be reserved for `const` vals, which these technically are not. I was following the example we had in `WallpaperManager`. I'm not seeing a ton of consistency in a short search of other `companion object`s. I think the caps casing is also a holdout from a stricter `static` variable convention in Java. I'm not personally a huge fan of it, but I'm happy to switch it if you'd prefer.

app/src/main/res/layout/fragment_address_editor.xml
MatthewTighe: Ah looks like `formDefault` isn't available as an `?attr` so it might be simplest to leave as is until the compose migration.

MozillaNoah: This is a `View` not a `TextView`. I'm fine with leaving it be though with the Compose migration on the roadmap.

app/src/main/java/org/mozilla/fenix/settings/address/view/AddressEditorState.kt
MozillaNoah: Gotcha gotcha. Without that file, I was completely lost about the new data class haha. If this and other data classes are going to be used to read-in data in a similar manner to that file, what if it's something like `ValidAddressCountry`, or maybe just `Country`, or make it a child class of `AddressEditorState` and it could referenced like `AddressEditorState.Country`

app/src/main/java/org/mozilla/fenix/settings/address/view/AddressEditorState.kt
```diff
@@ -0,0 +1,51 @@
...
+import org.mozilla.fenix.settings.address.AddressEditorFragment
+import org.mozilla.fenix.settings.address.AddressEditorFragment.Companion.availableCountries
+import org.mozilla.fenix.settings.address.AddressEditorFragment.Companion.defaultCountry
+
+/**
+ * The state for the [AddressEditorFragment].
+ *
+ * @property country The currently selected [AddressCountry].
+ */
+data class AddressEditorState(
```
gabrielluong, line 18: I believe we can avoid having `AddressEditorState` altogether, and just add all the country values and the default country handling inside of the `AddressEditorView`. We had an `CreditCardEditorState` because I needed a new data type that translated `CreditCard` into something that made sense for the UI - credit card number needed to be decrypted and values needed to be casted, etc. 

I think we can just handle this in the view as we are binding the dropdown values. This also feels cleaner to me instead of having the country values as companion values in the fragment, and having a state data class that only handles the country value. I think we got lucky with `Address` because it didn't need any transformation and we can simply bind and set the text so we just pass it as a parameter into the view.

app/src/main/java/org/mozilla/fenix/settings/address/view/AddressEditorView.kt
```diff
@@ -27,9 +30,11 @@ class AddressEditorView(
 ) {
 
     /**
-     * Binds the view.
+     * Binds the given [AddressEditorState] in the [AddressEditorFragment].
+     *
+     * @param
      */
-    fun bind() {
+    fun bind(state: AddressEditorState) {
```
gabrielluong, line 37: We currently pass the `address` as parameter into `AddressEditorView`. We discussed this briefly in https://github.com/mozilla-mobile/fenix/pull/25031#discussion_r866336823.

gabrielluong, line 37: We currently pass the `address` as parameter into `AddressEditorView`. We discussed this briefly in https://github.com/mozilla-mobile/fenix/pull/25031#discussion_r866336823.

app/src/main/res/layout/fragment_address_editor.xml
```diff
@@ -306,7 +306,7 @@
             app:hintEnabled="false"
             app:layout_constraintEnd_toEndOf="parent"
             app:layout_constraintStart_toEndOf="@+id/state_layout"
-            app:layout_constraintTop_toBottomOf="@+id/zip_title">
+            app:layout_constraintTop_toBottomOf="@+id/zip_title" >
```
gabrielluong, line 309: nit don't think we need this.

gabrielluong, line 309: nit don't think we need this or am i missing something? 🧐

app/src/main/res/navigation/nav_graph.xml
```diff
@@ -1258,12 +1258,12 @@
         <fragment
             android:id="@+id/addressEditorFragment"
             android:name="org.mozilla.fenix.settings.address.AddressEditorFragment"
-            android:label="@string/addresses_add_address">
-        <argument
-            android:name="address"
-            android:defaultValue="@null"
-            app:argType="mozilla.components.concept.storage.Address"
-            app:nullable="true" />
+            android:label="@string/addresses_add_address" >
```
gabrielluong, line 1261:
```suggestion
            android:label="@string/addresses_add_address">
```
nit don't think we need this extra space

gabrielluong, line 1261:
```suggestion
            android:label="@string/addresses_add_address">
```
nit don't think we need this extra space

app/src/test/java/org/mozilla/fenix/library/historymetadata/controller/HistoryMetadataGroupControllerTest.kt
```diff
@@ -188,6 +188,7 @@ class HistoryMetadataGroupControllerTest {
     }
 
     @Test
+    @Ignore("Intermittent test: https://github.com/mozilla-mobile/fenix/issues/25167")
```
gabrielluong, line 191: Since there's some requested changes in this PR, can you land the commit for these intermittent test in a separate PR. Assuming these tests are running intermittently regardless of the work here, it can very well affect other CI runs. So, let's just get this in ASAP and not have it be dependent on this PR.

gabrielluong, line 191: Since there's some requested changes in this PR, can you land the commit for these intermittent test in a separate PR. Assuming these tests are running intermittently regardless of the work here, it can very well affect other CI runs. So, let's just get this in ASAP and not have it be dependent on this PR.

app/src/main/java/org/mozilla/fenix/settings/address/view/AddressEditorView.kt
```diff
@@ -103,4 +110,20 @@ class AddressEditorView(
             create()
         }.show()
     }
+
+    private fun bindCountryDropdown(state: AddressEditorState) {
```
gabrielluong, line 114: Can you ask in the address team slack if we should just follow Desktop and display the full Country name (eg, United States, Canada)? I doubt we really want to display the key. The designs does indicate "USA", but that could just be a placeholder without further clarification.

gabrielluong, line 114: Can you ask in the address team slack if we should just follow Desktop and display the full Country name (eg, United States, Canada)? I doubt we really want to display the key. The designs does indicate "USA", but that could just be a placeholder without further clarification.

gabrielluong (changes requested):

app/src/main/res/layout/fragment_address_editor.xml
MatthewTighe: I'll revert both of these. Was trying to stay consistent with the tags that are closed inline, but I am now seeing that tags that are not closed inline typically don't have the space.

app/src/test/java/org/mozilla/fenix/library/historymetadata/controller/HistoryMetadataGroupControllerTest.kt
MatthewTighe: #25270

I'll add a comment to the issue tracking these as well

app/src/main/java/org/mozilla/fenix/settings/address/view/AddressEditorState.kt
MatthewTighe: Yeah, that makes sense to me. Thanks for the suggestion

app/src/main/java/org/mozilla/fenix/settings/address/view/AddressEditorView.kt
```diff
@@ -103,4 +147,32 @@ class AddressEditorView(
...
+            ?: regionPosition.takeIf { it != -1 }
+            ?: defaultPosition.takeIf { it != -1 }
+            ?: 0
+        binding.countryDropDown.adapter = adapter
+        binding.countryDropDown.setSelection(setPosition)
+    }
+
+    private fun String.toCountryCode() = availableCountries.values.find {
+        it.displayName == this
+    }?.key ?: defaultCountry.key
```
gabrielluong, line 177: `?: defaultCountry.key` looks particularly scary to me since you would hope we would never hit such a case, but knowing that we shouldn't ever hit this, it makes me wonder if we should even have it.

app/src/main/java/org/mozilla/fenix/settings/address/view/AddressEditorView.kt
```diff
@@ -103,4 +147,32 @@ class AddressEditorView(
...
+
+    private fun bindCountryDropdown() {
+        val adapter = ArrayAdapter<String>(
+            binding.root.context,
+            android.R.layout.simple_spinner_dropdown_item,
+        )
+
+        var selectedPosition = -1
+        var regionPosition = -1
+        var defaultPosition = -1
```
gabrielluong, line 159: I believe we can simplify this by having a getter method that returns the `selectedCountryKey` based on the editing `address`, `region` or the fallback to the `defaultCountry`. That way we are just looking for 1 position instead of 3 in this for loop and the long chain to get the right selection.

app/src/main/java/org/mozilla/fenix/settings/address/view/AddressEditorView.kt
```diff
@@ -23,11 +27,49 @@ import org.mozilla.fenix.settings.address.interactor.AddressEditorInteractor
...
+     */
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal data class ValidAddressCountry(
+        val key: String,
+        val name: String,
+        val displayName: String,
+    )
+
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal val defaultCountry = ValidAddressCountry(
```
gabrielluong, line 61: This is just a feeling from reading everything, but I almost feel like it could be suffice to use a constant country code `const val DEFAULT_COUNTRY = "US"` to avoid having an entire `ValidAddressCountry` appear out of place in the map.

app/src/main/java/org/mozilla/fenix/settings/address/view/AddressEditorView.kt
```diff
@@ -23,11 +27,49 @@ import org.mozilla.fenix.settings.address.interactor.AddressEditorInteractor
...
+        "US",
+        "UNITED STATES",
+        "United States"
+    )
+    internal val availableCountries = mapOf(
```
gabrielluong, line 66:
```suggestion
    internal val countries = mapOf(
```
Maybe?

gabrielluong (approved):

app/src/main/java/org/mozilla/fenix/settings/address/view/AddressEditorView.kt
```diff
@@ -23,11 +27,49 @@ import org.mozilla.fenix.settings.address.interactor.AddressEditorInteractor
...
+        "CA" to ValidAddressCountry("CA", "CANADA", "Canada"),
+        "US" to defaultCountry
+    )
+
+    /**
+     * Binds the view in the [AddressEditorFragment], using the current [Address] if available.
```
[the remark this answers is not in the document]
MatthewTighe: What's not true about this kDoc? Should it just include additional details about the region?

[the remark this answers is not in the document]
gabrielluong: My mistake. I read this as the address editor state.

app/src/main/java/org/mozilla/fenix/settings/address/view/AddressEditorView.kt
MatthewTighe: This was mostly to avoid map lookups, which are nullable. For example, this change would mean making substantial changes to tests. Instead of 

```
assertEquals(addressEditorView.defaultCountry.displayName, binding.countryDropDown.selectedItem.toString())
```

it would require something like 

```
assertEquals(addressEditorView.countries[DEFAULT_COUNTRY]!!.displayName, binding.countryDropDown.selectedItem.toString())
```

I'm fine with using the unsafe accessor in tests, but it may mean additional null guards in regular code as well.

app/src/main/java/org/mozilla/fenix/settings/address/view/AddressEditorView.kt
MatthewTighe: Yeah good suggestion, thanks

app/src/main/java/org/mozilla/fenix/settings/address/view/AddressEditorView.kt
```diff
@@ -23,11 +27,49 @@ import org.mozilla.fenix.settings.address.interactor.AddressEditorInteractor
...
+    internal data class ValidAddressCountry(
+        val key: String,
+        val name: String,
+        val displayName: String,
+    )
+
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal val defaultCountry = ValidAddressCountry(
+        "US",
+        "UNITED STATES",
```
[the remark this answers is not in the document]
MatthewTighe: Sounds good

app/src/main/java/org/mozilla/fenix/settings/address/view/AddressEditorView.kt
gabrielluong: On further thought, I am reasoning that this is fine since `find {}` will return something that could be nullable.

app/src/main/java/org/mozilla/fenix/settings/address/view/AddressEditorView.kt
gabrielluong: Makes sense. I am also fine with using the unsafe accessor in tests. I can imagine that when we do have an API that provides a similar map or object that we wouldn't necessarily be able to avoid the map lookup or null guards. 

I am okay with it as is. I will leave the final choice to you.

MatthewTighe: I agree that it makes the initial definition much cleaner, so I went with your suggestion

PR #25266, opened 2022-05-17
Change FxNimbus/Nimbus startup sequence

jhugman:
This is a reup of the #25089 and #25234.

The problem seemed to be two fold: 

1. the first is that the `MessagingAction.Restore` was being called from the `onExperimentsFetched` observer method.
2. the initialization of the Nimbus SDK is now happening earlier, earlier than the registration of the observer.

This caused the message not to be ready for the first display of the home screen, which was caught by the performance test.

The fix is to move the restore action to be dispatched explicitly after the nimbus SDK is ready.

### Pull Request checklist

1. click on Show All Checks,
4. click Details next to "Taskcluster (pull_request)" after it appears and then finishes with a green checkmark,
5. click on the "Fenix - assemble" task, then click "Run Artifacts".
6. the APK links should be on the left side of the screen, named for each CPU architecture


app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -402,7 +404,27 @@ open class FenixApplication : LocaleAwareApplication(), Provider {
...
+
+        fun restore() {
+            components.appStore.dispatch(AppAction.MessagingAction.Restore)
         }
+
+        components.analytics.experiments.register(object : NimbusInterface.Observer {
+            override fun onUpdatesApplied(updated: List<EnrolledExperiment>) {
+                restore()
+            }
+        })
```
Amejia481, line 425: Is it  Ok if we added as part of the existing [onUpdatesApplied]( https://github.com/mozilla-mobile/fenix/pull/25266/files#diff-9e2b3999e8d0e181af6decd7d25949eef361031a719a578869967a5339b31737R772)?

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -402,7 +404,27 @@ open class FenixApplication : LocaleAwareApplication(), Provider {
             // experiment on features close to startup.
             // But we need viaduct (the RustHttp client) to be ready before we do.
             components.analytics.experiments.initialize()
+
+            setupMessaging()
+        }
+    }
+
+    private fun setupMessaging() {
+        if (!FeatureFlags.messagingFeature) {
```
Amejia481, line 413: @jhugman is it OK to remove the `Settings.isExperimentationEnabled` check?

Amejia481 (approved):
Tested manually and it works expected.
I would prefer [if we could reuse the existing event listener](https://github.com/mozilla-mobile/fenix/pull/25266#discussion_r875154059), instead of introducing a new one, I also tested this alternative and it worked as expected too.

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
jhugman: I see you're already checking `isExperimentationEnabled` in the `HomeFragment`, in which case I don't know what the desired behaviour is (if the user flips the studies switch, should they start seeing messages immediately, or on the next startup?).

This is a bit strange, but I added back in.

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
jhugman: I combined the two observers into one.

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
Amejia481: It's a bit confusing. To make sure  users are in a consistent state when they turn on/off experiments, we let them know if they proceed the app will restart. 

image

app/src/test/java/org/mozilla/fenix/settings/SettingsFragmentTest.kt
```diff
@@ -56,8 +54,6 @@ class SettingsFragmentTest {
         mockkObject(Config)
         every { Config.channel } returns ReleaseChannel.Nightly
 
-        FxNimbus.api = NimbusDisabled(testContext)
-
```
jrbenny35, line 60: What does removing this do exactly?

jhugman: The `api` property is no longer public; this is no longer the way we want to initialize the FxNimbus object.

Issue #24299, opened 2022-03-16
Add History to the search engine menu


gabrielluong:
https://mozilla-hub.atlassian.net/browse/FNXV2-19758

SoftVision-LorandJanos:
Hi @gabrielluong!
I have verified the "History" item among the selectable search engines. It looks and functions as expected. However in the ticket attached in the description there is mentioning of "Bookmarks" along with "History" with the same behavior.
I didn't see the Bookmarks part implemented. Is there another ticket for Bookmarks, or maybe will there be? Thank you!

Devices used:
- Google Pixel 4 (Android 12).
- Xiaomi Mi11 Lite (Android 11).
- Samsung Galaxy Tab S3 (Android 9).

Tested on the latest Nightly 102.0a1 (2022-05-25).

PR #25267, opened 2022-05-17
Issue #24299: Add history to the search engine menu

rocketsroger:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/Core.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/Core.kt
@@ -8,14 +8,18 @@ import android.content.Context
 import android.content.res.Configuration
 import android.os.Build
 import android.os.StrictMode
+import androidx.appcompat.content.res.AppCompatResources.getDrawable
 import androidx.core.content.ContextCompat
+import androidx.core.graphics.drawable.toBitmap
 import mozilla.components.browser.engine.gecko.GeckoEngine
 import mozilla.components.browser.engine.gecko.fetch.GeckoViewFetchClient
 import mozilla.components.browser.engine.gecko.permission.GeckoSitePermissionsStorage
 import mozilla.components.browser.icons.BrowserIcons
 import mozilla.components.browser.session.storage.SessionStorage
 import mozilla.components.browser.state.engine.EngineMiddleware
+import mozilla.components.browser.state.search.SearchEngine
 import mozilla.components.browser.state.state.BrowserState
+import mozilla.components.browser.state.state.SearchState
 import mozilla.components.browser.state.store.BrowserStore
 import mozilla.components.browser.storage.sync.PlacesBookmarksStorage
 import mozilla.components.browser.storage.sync.PlacesHistoryStorage
@@ -39,6 +43,7 @@ import mozilla.components.feature.pwa.WebAppShortcutManager
 import mozilla.components.feature.readerview.ReaderViewMiddleware
 import mozilla.components.feature.recentlyclosed.RecentlyClosedMiddleware
 import mozilla.components.feature.recentlyclosed.RecentlyClosedTabsStorage
+import mozilla.components.feature.search.ext.createApplicationSearchEngine
 import mozilla.components.feature.search.middleware.AdsTelemetryMiddleware
 import mozilla.components.feature.search.middleware.SearchMiddleware
 import mozilla.components.feature.search.region.RegionMiddleware
@@ -189,6 +194,17 @@ class Core(
         }
     }
 
+    val applicationSearchEngines: List<SearchEngine> by lazyMonitored {
+        listOf(
+            createApplicationSearchEngine(
+                id = HISTORY_SEARCH_ENGINE_ID,
+                name = context.getString(R.string.library_history),
+                url = "",
+                icon = getDrawable(context, R.drawable.ic_history_search)?.toBitmap()!!,
+            )
+        )
+    }
+
     /**
      * The [BrowserStore] holds the global [BrowserState].
      */
@@ -217,6 +233,15 @@ class Core(
             )
 
         BrowserStore(
+            initialState = BrowserState(
+                search = SearchState(
+                    applicationSearchEngines = if (context.settings().showUnifiedSearchFeature) {
+                        applicationSearchEngines
+                    } else {
+                        emptyList()
+                    },
+                )
+            ),
             middleware = middlewareList + EngineMiddleware.create(
                 engine,
                 // We are disabling automatic suspending of engine sessions under memory pressure.
@@ -477,5 +502,11 @@ class Core(
         private const val RECENTLY_CLOSED_MAX = 10
         const val HISTORY_METADATA_MAX_AGE_IN_MS = 14 * 24 * 60 * 60 * 1000 // 14 days
         private const val CONTILE_MAX_CACHE_AGE = 60L // 60 minutes
+        const val HISTORY_SEARCH_ENGINE_ID = "history_search_engine_id"
+
+        // Maximum number of suggestions returned from the history search engine source.
+        const val METADATA_HISTORY_SUGGESTION_LIMIT = 100
+        // Maximum number of suggestions returned from shortcut search engine.
+        const val METADATA_SHORTCUT_SUGGESTION_LIMIT = 20
     }
 }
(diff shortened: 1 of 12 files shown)


app/src/main/res/drawable/ic_chevron_down_6.xml
```diff
@@ -8,6 +8,6 @@
     android:viewportWidth="6"
     android:viewportHeight="6">
     <path
-        android:fillColor="?attr/textPrimary"
+        android:fillColor="#000000"
```
MozillaNoah, line 11: is it alright that the color will now be black for light/dark?

app/src/main/res/drawable/ic_history_search.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
MozillaNoah, line 5: did I miss this being used somewhere?

app/src/main/res/layout/search_selector.xml
```diff
@@ -14,7 +14,7 @@
         android:layout_height="28dp"
         android:layout_marginTop="14dp"
         android:layout_marginHorizontal="8dp"
-        app:cardBackgroundColor="?attr/layer2"
+        app:cardBackgroundColor="#ffffff"
```
MozillaNoah, line 17: similar to above, just want to make sure this is correct for both light/dark

app/src/main/res/drawable/ic_chevron_down_6.xml
gabrielluong: This looks like an unintended change to me. Should we need to change the icon color, we should only do that through the iconTint in the ImageView

app/src/main/res/layout/search_selector.xml
gabrielluong: Don't think this is intended change either since it goes against the designs.

app/src/main/res/drawable/ic_chevron_down_6.xml
rocketsroger: These colour changes are not mistakes.  Search engines takes a bitmap when created.  At the moment we can't change the colour after we set the icon.  UX approved the solution that we can use this temporarily until a better solution is found.

app/src/main/res/drawable/ic_history_search.xml
rocketsroger: Yes, it's being used when creating the search engine.  I'll comment at the location.

app/src/main/java/org/mozilla/fenix/components/Core.kt
```diff
@@ -189,6 +194,17 @@ class Core(
         }
     }
 
+    val applicationSearchEngines: List<SearchEngine> by lazyMonitored {
+        listOf(
+            createApplicationSearchEngine(
+                id = HISTORY_SEARCH_ENGINE_ID,
+                name = context.getString(R.string.library_history),
+                url = "",
+                icon = getDrawable(context, R.drawable.ic_history_search)?.toBitmap()!!,
```
rocketsroger, line 203: Here is where the new icon is needed.

app/src/main/res/layout/search_selector.xml
rocketsroger: This is intended.  Since we're on this topic, @gabrielluong is there a better white and black colour I should be using here rather than hardcoding?  I need color that stays white / black even with theme changes.

app/src/main/java/org/mozilla/fenix/search/SearchDialogController.kt
```diff
@@ -69,6 +70,11 @@ class SearchDialogController(
 ) : SearchController {
 
     override fun handleUrlCommitted(url: String, fromHomeScreen: Boolean) {
+        // Do not load URL if search engine is application added
+        if (fragmentStore.state.searchEngineSource.searchEngine?.type == SearchEngine.Type.APPLICATION) {
+            return
```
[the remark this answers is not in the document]
rocketsroger: That's one of the requirements.  History should not trigger a URL load.  The user only searches history and either tap on the result or back out.  Maybe we need some kind of "no result" line?

app/src/main/res/drawable/ic_history_search.xml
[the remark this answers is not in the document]
rocketsroger: But the normal ic_history icon changes colour depends on when we use it.  If the user starts on Dark theme, then the icon becomes white and we won't be able to see it in the drop down.

app/src/main/java/org/mozilla/fenix/search/awesomebar/AwesomeBarView.kt
```diff
@@ -355,7 +314,11 @@ class AwesomeBarView(
                     engine,
                     shortcutSearchUseCase,
                     components.core.client,
-                    limit = 3,
+                    limit = if (activity.settings().showUnifiedSearchFeature) {
+                        METADATA_SHORTCUT_SUGGESTION_LIMIT
+                    } else {
+                        METADATA_SUGGESTION_LIMIT
+                    },
```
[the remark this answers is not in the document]
rocketsroger: The current history search results limit in the history page.

app/src/main/res/drawable/ic_chevron_down_6.xml
gabrielluong: Let's talk about this tomorrow to make sure we are synced. At the very least, I don't think we necessarily have to make the UX changes in this PR.

rocketsroger: As agreed for now I will apply the tint to ic_chevron_down_6 and use photonBlack here.  Thanks

app/src/main/java/org/mozilla/fenix/search/SearchDialogController.kt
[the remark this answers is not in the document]
gabrielluong: Just pointing it out since I was just investigating what the code did. I think we should just file a followup issue to have UX investigate what if anything we should do in such a state. No changes needed, but felt it was noteworthy.

app/src/main/java/org/mozilla/fenix/search/SearchDialogController.kt
```diff
@@ -69,6 +70,11 @@ class SearchDialogController(
 ) : SearchController {
 
     override fun handleUrlCommitted(url: String, fromHomeScreen: Boolean) {
+        // Do not load URL if search engine is application added
```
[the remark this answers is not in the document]
gabrielluong: Actually, I am wrong about this suggested change. I think instead of `added` this should be `selected`
```suggestion
        // Do not load URL if application search engine is selected.
```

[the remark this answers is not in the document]
rocketsroger: I'll update.  Thanks

app/src/main/res/drawable/ic_history_search.xml
rocketsroger: As agreed with @gabrielluong issue created as a follow up to see if we can support themes in search engine icons https://github.com/mozilla-mobile/android-components/issues/12187

app/src/main/java/org/mozilla/fenix/search/SearchDialogController.kt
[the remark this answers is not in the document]
rocketsroger: Agreed, issue created to track this https://github.com/mozilla-mobile/fenix/issues/25282

app/src/main/java/org/mozilla/fenix/search/awesomebar/AwesomeBarView.kt
```diff
@@ -213,33 +212,14 @@ class AwesomeBarView(
     fun updateSuggestionProvidersVisibility(
         state: SearchProviderState
     ) {
+        view.removeAllProviders()
```
gabrielluong, line 215: Let's add a new line below this

app/src/main/java/org/mozilla/fenix/search/awesomebar/AwesomeBarView.kt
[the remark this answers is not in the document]
gabrielluong: Since this is coming from the history search results page, could we import those values instead of defining new ones.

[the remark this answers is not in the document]
rocketsroger: Sure, I'll update.  Thanks

Issue #25281, opened 2022-05-18
Pocket sponsored stories rotation and pacing


Mugurell:
Needs https://github.com/mozilla-mobile/android-components/issues/12184

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-20502)

Mugurell:
Since this is hidden in debug I'll set `qa:not-needed` for now.

Issue #25025, opened 2022-04-29
Add Pocket sponsored stories


Mugurell:
Tracking Jira issue - https://mozilla-hub.atlassian.net/browse/FNXV2-19877

Needs https://github.com/mozilla-mobile/android-components/pull/12070

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-20321)

gabrielluong:
@Mugurell should this be closed or set for qa-needed?

Mugurell:
Since this is hidden in debug I'll set `qa:not-needed` for now.

PR #25037, opened 2022-04-29
For #25025 - Integrate Pocket sponsored stories

Mugurell:
~Needs https://github.com/mozilla-mobile/android-components/pull/12070~
~Needs https://github.com/mozilla-mobile/android-components/pull/12183~

~Based on https://github.com/mozilla-mobile/fenix/pull/24892.~

### Pull Request checklist

- [x] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not
- [x] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features. In addition, it includes a screenshot of a successful [accessibility scan](https://play.google.com/store/apps/details?id=com.google.android.apps.accessibility.auditor&hl=en_US) to ensure no new defects are added to the product.

### To download an APK when reviewing a PR:
1. click on Show All Checks,
2. click Details next to "Taskcluster (pull_request)" after it appears and then finishes with a green checkmark,
3. click on the "Fenix - assemble" task, then click "Run Artifacts".
4. the APK links should be on the left side of the screen, named for each CPU architecture

PR #25290, opened 2022-05-18
For #25281: Add support for pacing and rotating sponsored Pocket stories

Mugurell:
~Work based on the unmerged https://github.com/mozilla-mobile/fenix/issues/25025
Will rebase once that is merged.~
Rebased.

Needs https://github.com/mozilla-mobile/android-components/pull/12185

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/ext/AppState.kt
+++ b/app/src/main/java/org/mozilla/fenix/ext/AppState.kt
@@ -8,6 +8,8 @@ import androidx.annotation.VisibleForTesting
 import mozilla.components.service.pocket.PocketStory
 import mozilla.components.service.pocket.PocketStory.PocketRecommendedStory
 import mozilla.components.service.pocket.PocketStory.PocketSponsoredStory
+import mozilla.components.service.pocket.ext.hasFlightImpressionsLimitReached
+import mozilla.components.service.pocket.ext.hasLifetimeImpressionsLimitReached
 import org.mozilla.fenix.components.appstate.AppState
 import org.mozilla.fenix.home.blocklist.BlocklistHandler
 import org.mozilla.fenix.home.pocket.POCKET_STORIES_DEFAULT_CATEGORY_NAME
@@ -15,8 +17,19 @@ import org.mozilla.fenix.home.pocket.PocketRecommendedStoriesCategory
 import org.mozilla.fenix.home.pocket.PocketStory
 import org.mozilla.fenix.home.recenttabs.RecentTab.SearchGroup
 
+/**
+ * Total count of all stories to show irrespective of their type.
+ * This is an optimistic value taking into account that fewer than this stories may actually be available.
+ */
 @VisibleForTesting
 internal const val POCKET_STORIES_TO_SHOW_COUNT = 8
+
+/**
+ * Total count of all sponsored Pocket stories to show.
+ * This is an optimistic value taking into account that fewer than this stories may actually be available.
+ */
+@VisibleForTesting
+internal const val POCKET_SPONSORED_STORIES_TO_SHOW_COUNT = 2
 
 /**
  * Get the list of stories to be displayed based on the user selected categories.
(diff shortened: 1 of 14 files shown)


app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesComposables.kt
```diff
@@ -151,21 +211,96 @@ fun PocketStories(
...
+/**
+ * Add a callback for when this Composable is "shown" on the screen.
+ * This checks whether the composable has at least [threshold] ratio of it's total area drawn inside
+ * the screen bounds.
+ * Does not account for other Views / Windows covering it.
+ */
+private fun Modifier.onShown(
+    @FloatRange(from = 0.0, to = 1.0) threshold: Float,
+    onVisible: () -> Unit,
+): Modifier {
```
Mugurell, line 248: Based on https://github.com/mozilla-mobile/fenix/pull/25206#discussion_r870580603 I could extract this to the `Modifier` file but not sure how to cleanly pass the toolbar height.
Adding two new parameters like `topPaddingToIgnore` / `bottomPaddingToIgnore` seems a bit forced for a scenario common to homescreen but not in other places where other Views may appear on top of a composable and obstruct it.
Ideally this shouldn't be needed but I haven't found a way for the composable to know on it's own about the toolbar appearing above it.

MozillaNoah: One option could be instead to do something like adding a boolean instead to  `Modifier.onShown()` (something like `toolbarPresent` or `calculateWithToolbar`) to indicate if the Toolbar is present or not and then do the calculations as normal, rather than including specific padding values to calculate against.

That being said, we could totally just let this live here until it's needed somewhere else to give it time for a solution to bake.

app/src/main/java/org/mozilla/fenix/ext/AppState.kt
```diff
@@ -17,6 +19,8 @@ import org.mozilla.fenix.home.recenttabs.RecentTab.SearchGroup
 
 @VisibleForTesting
 internal const val POCKET_STORIES_TO_SHOW_COUNT = 8
```
mcarare, line 21: Maybe a comment stating this count will include the sponsored stories would be useful. It is not obvious from the nearest code. (although it can be deduced from `combineRecommendedAndSponsoredStories` method)

app/src/main/java/org/mozilla/fenix/ext/AppState.kt
```diff
@@ -5,31 +5,46 @@
...
+            ?.take(POCKET_STORIES_TO_SHOW_COUNT) ?: emptyList()
+
+        val sponsoredStories = getFilteredSponsoredStories(
+            stories = pocketSponsoredStories,
+            limit = POCKET_SPONSORED_STORIES_TO_SHOW_COUNT,
+        )
+
+        return combineRecommendedAndSponsoredStories(
+            recommendedStories = recommendedStories,
+            sponsoredStories = sponsoredStories.toList()
```
mcarare, line 46: `toList()` seems redundant. `getFilteredSponsoredStories` already returns a `List<PocketSponsoredStory>`

app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesComposables.kt
```diff
@@ -151,21 +211,96 @@ fun PocketStories(
...
-                    } else {
+                    } else if (story is PocketRecommendedStory) {
                         PocketStory(story) {
                             val uri = Uri.parse(story.url)
                                 .buildUpon()
                                 .appendQueryParameter(URI_PARAM_UTM_KEY, POCKET_STORIES_UTM_VALUE)
                                 .build().toString()
                             onStoryClicked(it.copy(url = uri), rowIndex to columnIndex)
                         }
+                    } else if (story is PocketSponsoredStory) {
```
mcarare, line 222: This if-else cascade could be replaced with a when statement.

mcarare (approved):
LGTM with some comments. Will also take another look after the PR it is based on will be merged.

app/src/main/java/org/mozilla/fenix/ext/AppState.kt
Mugurell: Thank you. Will address this in https://github.com/mozilla-mobile/fenix/pull/25037 which extracted this as a top level property.

Mugurell: Seems to make more sense to add this comment here about the value including the count of sponsored stories, while previously just the naming could have been enough. Thank you!

Mugurell:
Tested a bit more the impressions recording and saw two scenarios that should be improved for which I've added a new commit:
- sponsored stories were recorded as shown as part of the impressions recording
for all the stories (both recommended and sponsored). Separated the two.
- sponsored stories were not recorded as shown if the user doesn't scroll.
This could happen if just toggling categories which would result in a new list
of stories to show. Will recheck after the time to settle if the composable is
shown.

Mugurell:
Updated the AC version to the absolute latest.

Issue #25291, opened 2022-05-18
Make a top-level TextButton Composable component


MozillaNoah:
[Specs](https://www.figma.com/file/ujdiReypSypu2GhRgzArea/Android-Components?node-id=1430%3A3433)

Replace the current `TextButton` usages in Wallpaper with the new component.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-20507)

MozillaNoah:
@cwzilla, just to confirm for text-only buttons, should the button label _always_ be capitalized?

cwzilla:
@MozillaNoah I think it's safe to say text-only buttons will always be capitalized and the same component can be used in Snackbars.

gabrielluong:
@MozillaNoah just a friendly reminder to close your issues. 🫡

Issue #23221, opened 2022-01-14
Add a generalized `FenixSnackbar` composable


MatthewTighe:
As of Compose 1.0.5, the platform version of a `Snackbar` is missing parameters that we require in order to easily match the [design spec](https://www.figma.com/file/TebIgESfWmQkMcEmwNaYZl/Android?node-id=1399%3A590) for our snackbars. 

Add a general version that use our designs but still expose customizable messages and actions. 

See also the [XML layout](https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/res/layout/fenix_snackbar.xml) for our snackbar to match previous design.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-19160)

gabrielluong:
Moved to bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1814198

Change performed by the [Move to Bugzilla add-on](https://addons.mozilla.org/en-US/firefox/addon/move-to-bugzilla/).

PR #25377, opened 2022-05-25
 For #25291 - Create a TextButton composable component 

MozillaNoah:
For #25291 

image

### Pull Request checklist

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/compose/button/TextButton.kt
```diff
@@ -0,0 +1,66 @@
...
+    onClick: () -> Unit,
+    modifier: Modifier = Modifier,
+    textColor: Color = FirefoxTheme.colors.textAccent,
+) {
+    androidx.compose.material.TextButton(
+        onClick = onClick,
+        modifier = modifier,
+    ) {
+        Text(
+            text = text.uppercase(),
```
gabrielluong, line 41: Any reason why we dropped the `Locale.getDefault()`? It probably makes sense to consider the locale.

app/src/main/java/org/mozilla/fenix/compose/button/TextButton.kt
```diff
@@ -0,0 +1,66 @@
...
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.text.TextStyle
+import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.sp
+import org.mozilla.fenix.theme.FirefoxTheme
+import org.mozilla.fenix.theme.Theme
+
+/**
+ * Base component for text-only buttons.
```
gabrielluong, line 22: I think we drop the "Base component" part since this is a public and non-extensible component.

app/src/main/java/org/mozilla/fenix/compose/button/TextButton.kt
```diff
@@ -0,0 +1,66 @@
...
+    ) {
+        Text(
+            text = text.uppercase(),
+            style = TextStyle(
+                color = textColor,
+                fontSize = 14.sp,
+                fontWeight = FontWeight.Medium,
+                letterSpacing = 0.25.sp,
+                lineHeight = 14.sp,
+            ),
```
gabrielluong, line 48: One day we probably want to group these `TextStyle` into something like Header, Title, Description, etc if we can find some kind of commonality within the component system, but ultimately needs to be defined within the design system.

app/src/main/java/org/mozilla/fenix/compose/button/TextButton.kt
MozillaNoah: My impression is `uppercase()` is using the default locale when one isn't passed-in.

app/src/main/java/org/mozilla/fenix/compose/button/TextButton.kt
MozillaNoah: Yes!! My current impression is that we'll have some good progress on the components and then we'll start on a `TextStyle` library and eventually a `Shapes` library.

app/src/main/java/org/mozilla/fenix/compose/button/TextButton.kt
MozillaNoah: "[Uppercase] Returns a copy of this string converted to upper case using Unicode mapping rules of the invariant locale."

https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/uppercase.html

gabrielluong: Looking around our codebase, I would feel more comfortable passing the `Locale.getDefault()` from my search for `lowercase`, `capitalize` in both AC/Fenix.

MozillaNoah: I'm fine with doing that to keep it consistent until we decide to re-examine it all

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -143,18 +142,11 @@ private fun WallpaperSnackbar(
...
-                Text(
-                    text = stringResource(R.string.wallpaper_updated_snackbar_action).uppercase(
-                        Locale.getDefault()
-                    ),
-                    color = FirefoxTheme.colors.textOnColorPrimary,
-                    fontFamily = FontFamily(Font(R.font.metropolis_medium)),
-                    fontSize = 14.sp,
-                )
-            }
+                modifier = Modifier.padding(all = 8.dp),
```
gabrielluong, line 147: Any reasoning we are dropping the 16dp start/end?

gabrielluong: I will look to realign this in https://github.com/mozilla-mobile/fenix/issues/23221 with the Figma designs.

MozillaNoah: It looked like too much padding was there (based on eyeballing the symmetry)
image

gabrielluong (approved):

app/src/main/java/org/mozilla/fenix/compose/button/TextButton.kt
gabrielluong: Feel free to file an issue for the `TextStyle` and work on it. It appears ready, but you should double check. https://www.figma.com/file/pEyGeE4KV5ytYHeXMfLcEr/Mobile-Styles?node-id=861%3A54432

Similar to how we added usage comments for the design color tokens. I would also make a note to also include the usage notes that are included in the Figma usage section. I would just copy it verbatim since I think all the documentations related to these component/color usages should be defined in Figma and managed properly from there as the source.

MozillaNoah: https://github.com/mozilla-mobile/fenix/issues/25379

PR #25403, opened 2022-05-26
fixes #25386: switch task continuity flag to nightly and remove secret setting for it

### LINKED ISSUE #25386: Expose task continuity to nightly
Task continuity is in a good enough state that to start getting feedback from the nightly audience on it. 

- Switch the feature flag to be on in nightly and debug
- Stop hiding it behind a secret settings menu

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-20565)

MatthewTighe:
1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/FeatureFlags.kt
+++ b/app/src/main/java/org/mozilla/fenix/FeatureFlags.kt
@@ -87,7 +87,7 @@ object FeatureFlags {
     /**
      * Enables the Task Continuity enhancements.
      */
-    val taskContinuityFeature = Config.channel.isDebug
+    val taskContinuityFeature = Config.channel.isNightlyOrDebug
 
     /**
      * Enables the Unified Search feature.
--- a/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
@@ -276,7 +276,7 @@ class HomeFragment : Fragment() {
                 view = binding.root
             )
 
-            if (FeatureFlags.taskContinuityFeature) {
+            if (requireContext().settings().enableTaskContinuityEnhancements) {
                 recentSyncedTabFeature.set(
                     feature = syncedTabFeature,
                     owner = viewLifecycleOwner,
--- a/app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabViewHolder.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabViewHolder.kt
@@ -15,7 +15,6 @@ import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.unit.dp
 import androidx.lifecycle.LifecycleOwner
 import mozilla.components.lib.state.ext.observeAsComposableState
-import org.mozilla.fenix.FeatureFlags
 import org.mozilla.fenix.R
 import org.mozilla.fenix.components.components
 import org.mozilla.fenix.compose.ComposeViewHolder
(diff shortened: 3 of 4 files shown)


app/src/main/java/org/mozilla/fenix/FeatureFlags.kt
```diff
@@ -87,7 +87,7 @@ object FeatureFlags {
     /**
      * Enables the Task Continuity enhancements.
      */
-    val taskContinuityFeature = Config.channel.isDebug
+    val taskContinuityFeature = Config.channel.isNightlyOrDebug
```
gabrielluong, line 90: I think you realistically only need to make this change. You might want to keep the secret setting so that you can still manually turn this on in Beta for any reason. Other than that keep the preference keys still in place and having a preference in `Settings` also means you can reuse it for Nimbus experimentation. Generally, I remove the secret settings when the feature has shipped, but ultimately if you don't feel the need for the secret setting, then the changes are also good to go. Food for thought depending on what you think the needs are.

gabrielluong (approved):

app/src/main/java/org/mozilla/fenix/FeatureFlags.kt
MatthewTighe: The intent here is to ship the feature to the general nightly audience, not just internal people that know about the secret setting. 

Do you mean manually turn it off in beta/nightly? As I understand it, a false feature flag in a `featureFlagPreference` stops the secret setting from being enabled completely.

Point taken that it could be helpful to be able to control the setting through experimentation though. I'll leave that in and change the default to true so that it's available, and also switch the recent synced tab to use it.

gabrielluong: Right, so changing the FeatureFlag to `isNightlyOrDebug` will get you to shipping to the general nightly audience. However, removing the secret setting does mean you have no way to turn it ON in Beta for internal testing if that is ever necessary. I am not worry about turning the feature OFF since that is still controlled via the FeatureFlag. Otherwise, it is safe to remove the secret setting portion. Whether or not you want to keep the `enableTaskContinuityEnhancements` depends on future needs - Nimbus, User preference in some Settings UI, etc. It probably deserves a rename to detangle it from other task continuity features if they are considered separate.

MatthewTighe: My understanding based on the code doesn't track with that. Is there something I'm missing here? Here's the definition of `featureFlagPreference`:

```
fun featureFlagPreference(key: String, default: Boolean, featureFlag: Boolean) =
    if (featureFlag) {
        booleanPreference(key, default)
    } else {
        DummyProperty()
    }
```

`DummyProperty` only returns false. Wouldn't this mean that if the flag is set to `isNightlyOrDebug` then it could never be turned on in beta anyway?

It seems like we'd need:

```
val taskContinuityFeature = Config.channel.isNightlyOrDebug || Config.channel.isBeta

/* in settings */
var enableTaskContinuityEnhancements by featureFlagPreference(
    key = appContext.getPreferenceKey(R.string.pref_key_enable_task_continuity),
    default = { /* inspect nimbus */},
    featureFlag = FeatureFlags.taskContinuityFeature,
)
```

So I think what we'd want for now is:

```
val taskContinuityFeature = Config.channel.isNightlyOrDebug

/* in settings */
var enableTaskContinuityEnhancements by featureFlagPreference(
    key = appContext.getPreferenceKey(R.string.pref_key_enable_task_continuity),
    default = true,
    featureFlag = FeatureFlags.taskContinuityFeature,
)
```

Am I on the same page now?

gabrielluong: Ah, I guess I am in the wrong here. If we wanted to test this in Beta, we would set the FeatureFlag to true by then.

PR #25419, opened 2022-05-27
 For #25379 - Create a Compose Typography system 

### LINKED ISSUE #25379: Create a typography system and integrate it into FirefoxTheme
[Specs](https://www.figma.com/file/pEyGeE4KV5ytYHeXMfLcEr/Mobile-Styles?node-id=861%3A54432)

Similar to how we added usage comments for the design color tokens, add a note to also include the usage notes that are included in the Figma usage section. For now, just copy it verbatim since all of the documentations related to these component/color usages should be defined in Figma and managed properly from there as the source.

MozillaNoah:
Closes #25379 

image

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/theme/Typography.kt
+++ b/app/src/main/java/org/mozilla/fenix/theme/Typography.kt
@@ -1,0 +1,163 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.theme
+
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.lazy.LazyColumn
+import androidx.compose.foundation.lazy.items
+import androidx.compose.material.MaterialTheme
+import androidx.compose.material.Text
+import androidx.compose.material.Typography
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.text.TextStyle
+import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+
+val FenixTypography = Typography(
+
+//    h1 = , Not currently in design system
+//    h2 = , Not currently in design system
+//    h3 = , Not currently in design system
+//    h4 = , Not currently in design system
+
+    /** Currently not in-use. */
+    h5 = TextStyle(
+        fontSize = 24.sp,
+        fontWeight = FontWeight.W400,
+        letterSpacing = 0.18.sp,
+        lineHeight = 24.sp,
+    ),
+
+    /** Used for headings on Onboarding Modals and App Bar Titles. */
+    h6 = TextStyle(
+        fontSize = 20.sp,
+        fontWeight = FontWeight.W500,
+        letterSpacing = 0.15.sp,
+        lineHeight = 24.sp,
+    ),
+
+    /** Used for Lists. */
+    subtitle1 = TextStyle(
+        fontSize = 16.sp,
+        fontWeight = FontWeight.W400,
+        letterSpacing = 0.15.sp,
+        lineHeight = 24.sp,
+    ),
+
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/mozilla/fenix/theme/Typography.kt
```diff
@@ -0,0 +1,163 @@
...
+    ),
+)
+
+/** Used for headings on Cards, Dialogs, Banners, and Homepage. */
+val Typography.h7: TextStyle
+    @Composable
+    get() {
+        return TextStyle(
+            fontSize = 16.sp,
+            fontWeight = FontWeight.W500,
```
gabrielluong, line 109: Can we change all of the WXXX font weight and use their respective alias such as `FontWeight.Medium`? It would be nicer to align the font weight wording closer to what we can see in the Figma.

app/src/main/java/org/mozilla/fenix/theme/Typography.kt
```diff
@@ -0,0 +1,163 @@
...
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+
+val FenixTypography = Typography(
+
+//    h1 = , Not currently in design system
+//    h2 = , Not currently in design system
+//    h3 = , Not currently in design system
+//    h4 = , Not currently in design system
```
gabrielluong, line 28: Let's just remove these if they aren't in used.

app/src/main/java/org/mozilla/fenix/theme/Typography.kt
```diff
@@ -0,0 +1,163 @@
...
+    overline = TextStyle(
+        fontSize = 10.sp,
+        fontWeight = FontWeight.W400,
+        letterSpacing = 1.5.sp,
+        lineHeight = 16.sp,
+    ),
+)
+
+/** Used for headings on Cards, Dialogs, Banners, and Homepage. */
+val Typography.h7: TextStyle
```
gabrielluong, line 104: Wondering why h7 and h8 are the only ones that appears outside of FenixTypography

app/src/main/java/org/mozilla/fenix/theme/Typography.kt
MozillaNoah: Per an offline convo with @cwzilla, we may be removing the alias names and just using the numeric values going forward.

app/src/main/java/org/mozilla/fenix/theme/Typography.kt
MozillaNoah: The Typography class does not have them as part of their in-built system, so I had to make them extensions

app/src/main/java/org/mozilla/fenix/theme/Typography.kt
MozillaNoah: kk, will do

app/src/main/java/org/mozilla/fenix/theme/Typography.kt
MozillaNoah: I tried just making a custom `Typography` class to bake it all together, but the `Typography`  class is `final`, unfortunately.

app/src/main/java/org/mozilla/fenix/theme/Typography.kt
MozillaNoah: Actually it looks like the style sheet has already been updated with the numbers.

gabrielluong:
Adding a "do-not-land" to avoid landing this before the Beta cut. We should consider how overriding the Material typography theme will affect the material compose components and if the typography for them still matches what we expect them to look like in the design system.

MatthewTighe:
🚗 wondering if it makes sense to expose the typography as part of `FirefoxTheme`, like we do for colors? Seems like it might help reduce overall complexity to be able to access all our custom styling elements in the same way

MozillaNoah:
> 🚗 wondering if it makes sense to expose the typography as part of `FirefoxTheme`, like we do for colors? Seems like it might help reduce overall complexity to be able to access all our custom styling elements in the same way

I'm actually in the middle of working on a V2 PR to do just that! I'll post a link here when it's ready.

MozillaNoah:
V2 PR: https://github.com/mozilla-mobile/fenix/pull/25487

MozillaNoah:
closing this PR in favor of the V2: https://github.com/mozilla-mobile/fenix/pull/25487

Issue #25446, opened 2022-05-30
Ensure for every message shown we have recordExposure


Amejia481:
Chatting with the Data science and the experiments team, they mentioned they were seeing some discrepancy between the nimbus.`recordExposure` vs the `messageShown` events. We want to ensure that every time we have a `messageShown` event, we have a corresponding `recordExposure` event.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-20596)

Amejia481:
Initially I was hoping the Data science and the experiments team could validate, but if this is something that didn't take that much time for QA could helpful too 😃 . Please if you aren't able to simulate, let me know, don't block the beta RC on this as it's a low risk ticket, that will be also validated by the Data science and the experiments team. Thanks in advance!

Instructions.

1. Configure a nimbus message, on the experimenter console, the message should have the `message-under-experiment` property set to the `message-id`. 
2. Publish the message.
3. Build a Firefox version that targets the experimenter server where the message was published.
4. Open a clean version of Firefox.
5. Dismiss the onboarding.
6. Enrolled Firefox to receive the message via the nimbus experiments screen.
7. Close Firefox, and open it again (**then the experiments should be downloaded**).
8. Close Firefox, and open it again (**then the configured messages should show**).
9. Every time it is shown, we should see a report of the glean  [messaging.message_shown](https://dictionary.telemetry.mozilla.org/apps/fenix/metrics/messaging_message_shown) event where the `message_key` is equal to `message-id` and also a [nimbus_events.exposure](https://dictionary.telemetry.mozilla.org/apps/fenix/metrics/nimbus_events_exposure).
10. To force the message to re-show, you could press the middle button to exit the app and re-enter.

PR #25447, opened 2022-05-30
For #25446: Ensure for every message shown we have recordExposure

Amejia481:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/gleanplumb/DefaultMessageController.kt
+++ b/app/src/main/java/org/mozilla/fenix/gleanplumb/DefaultMessageController.kt
@@ -14,7 +14,6 @@ import org.mozilla.fenix.HomeActivity
 import org.mozilla.fenix.components.AppStore
 import org.mozilla.fenix.components.appstate.AppAction.MessagingAction.MessageClicked
 import org.mozilla.fenix.components.appstate.AppAction.MessagingAction.MessageDismissed
-import org.mozilla.fenix.components.appstate.AppAction.MessagingAction.MessageDisplayed
 
 /**
  * Handles default interactions with the ui of GleanPlumb messages.
@@ -42,14 +41,6 @@ class DefaultMessageController(
     override fun onMessageDismissed(message: Message) {
         Messaging.messageDismissed.record(Messaging.MessageDismissedExtra(message.id))
         appStore.dispatch(MessageDismissed(message))
-    }
-
-    override fun onMessageDisplayed(message: Message) {
-        if (message.maxDisplayCount <= message.metadata.displayCount + 1) {
-            Messaging.messageExpired.record(Messaging.MessageExpiredExtra(message.id))
-        }
-        Messaging.messageShown.record(Messaging.MessageShownExtra(message.id))
-        appStore.dispatch(MessageDisplayed(message))
     }
 
     @VisibleForTesting
--- a/app/src/main/java/org/mozilla/fenix/components/appstate/AppAction.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/appstate/AppAction.kt
@@ -137,11 +137,6 @@ sealed class AppAction : Action {
         data class MessageClicked(val message: Message) : MessagingAction()
 
         /**
-         * Indicates the given [message] was shown.
-         */
-        data class MessageDisplayed(val message: Message) : MessagingAction()
-
-        /**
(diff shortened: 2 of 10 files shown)


app/src/main/java/org/mozilla/fenix/gleanplumb/DefaultMessageController.kt
```diff
@@ -48,6 +50,7 @@ class DefaultMessageController(
         if (message.maxDisplayCount <= message.metadata.displayCount + 1) {
             Messaging.messageExpired.record(Messaging.MessageExpiredExtra(message.id))
         }
+        messagingFeature.recordExposure()
```
jhugman, line 53: If there are three messages in the system: 

- message `FML` is hardcoded into the `nimbus.fml.yaml` file.
- message `EXP` is from an experiment
- message `ROL` is from a rollout.

If a `FML` or `ROL` are displayed, the message system  will `recordExposure` here, erroneously.

(If `EXP` doesn't exist, i.e. there are no messages under experiment, this is ok, but when there are, you over record exposure to the treatment even when you're not displaying the treatment).

jhugman: I think you're also recording exposure twice in the case of `EXP`, since you record exposure when the message is selected.

Recording at message selection time is good, because it's a good time to deal with control message.

I guess one good way of doing this would be to add a `recordExposure` boolean to the Message state itself, so that you only call into Nimbus when the message is displayed, but the decision whether to record or not is made at message selection time.

jhugman (changes requested):
Unfortunately, this doesn't quite work as you expect. I've added a comment to suggest a way forward.

app/src/main/java/org/mozilla/fenix/gleanplumb/DefaultMessageController.kt
Amejia481: My mistake, it looks like I didn't committed the change, where I removing the `recordExposure` from where we are selecting the message.

app/src/main/java/org/mozilla/fenix/gleanplumb/NimbusMessagingStorage.kt
```diff
@@ -97,8 +97,6 @@ class NimbusMessagingStorage(
         if (!isMessageUnderExperiment(message, nimbusFeature.messageUnderExperiment)) {
             return message
         }
-        // If the message is under experiment, then we need to record the exposure
-        messagingFeature.recordExposure()
```
Amejia481, line 101: @jhugman I updated the PR to include this change.

app/src/main/java/org/mozilla/fenix/gleanplumb/DefaultMessageController.kt
Amejia481: Could help me to clarify what is a "message ROL is from a rollout"?

Amejia481: > message EXP is from an experiment
Do an message EXP should be shown to the user?

Amejia481:
The issue is we are reporting the message shown event directly on a specific surface, that it's using a [recycle view](https://developer.android.com/guide/topics/ui/layout/recyclerview), as expected it does an efficient use of the created views by it reusing the existing view unless something visual has changed, **as result, we are only sending the shown event a subset of times, that caused a mix match between when the message UI was visible vs when the message shown event** was sent. **On the PR, I'm decoupling the shown event from the surface**, as each surface can have its own way of reusing  and UI components, instead I'm using [messageToShow](https://github.com/mozilla-mobile/fenix/blob/685424de9aa26e655a57659722022486aef43254/app/src/main/java/org/mozilla/fenix/gleanplumb/MessagingState.kt#L15) that is the state that all surfaces will observe for displaying messages **(the source of truth)** , every time we update it, we will send a message shown event, as every time we update this state a surface will displayed the illegible message.

jhugman (approved):
This is excellent sleuthing, and [such a simple fix](https://github.com/mozilla-mobile/fenix/pull/25447/files#diff-a4f1380abfbf1f76bc6bbd9ef961bd833145c4c75dae3f68dbff419ddd9d5d9cR50). Thank you.

Amejia481:
@Mergifyio backport releases_v102.0.0

Issue #24744, opened 2022-04-12
Investigate whether a weak reference is still needed in SearchSelectorToolbarAction


gabrielluong:
This is a follow up from a conversation with @csadilek regarding https://github.com/mozilla-mobile/fenix/pull/24732/files#diff-6878fa8bd134525847765379db3ae0d3d37bdef58007c8dbc6c596c329366d47. 

Notes from @csadilek
> This looks good to me relative to what we have, but we really shouldn't need that weak reference. I can't find the blame, but this looks like a bug workaround for the lifecyclescope thingy.
> 
> If the observer is bound to the viewLifecycleOwner there shouldn't be a case where we collect and update the UI if the view is no longer alive.
> 
> Unless this leaked in a different way I don't understand right now.
> 
> I think we can dig up the blame on the fenix side where this was changed before we upstreamed to A-C.
> 
> It really shouldn't make a differences, unless the Action lives longer than the view  then we would leak.
> 
> (And I am not sure that's true anymore.)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-20144)

Mugurell:
@ QA: This ticket was about refactoring code, please ensure the app functions the same as before in regards to the search engine selector from toolbar (part of unified search).

LaurentiuApahideanSV:
I tested the issue on Firefox Nightly 103.0a1 2022-06-09 by selecting different search engines from the toolbar and didn't encounter any issues.
### Devices used:
- Google Pixel 6 (Android 13)
- OnePlus 6T (Android 9)

PR #25517, opened 2022-06-03
For #24744 - Observe and update selector icon while the View is attached

Mugurell:
This avoids having to pass another LifecycleOwner from outside and will ensure
the View is update only in the bounds of it being attached.

The observe-update code is moved to bind(..) as that seems like a more
idiomatic callback for updating an already constructed View rather than
createView() which should only create and return a View.

UX should be the same as before

https://user-images.githubusercontent.com/11428869/171903109-47dcd7d3-5af5-403f-b6f3-84772f949977.mp4

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/search/toolbar/SearchSelectorToolbarAction.kt
+++ b/app/src/main/java/org/mozilla/fenix/search/toolbar/SearchSelectorToolbarAction.kt
@@ -9,21 +9,23 @@ import android.graphics.Bitmap
 import android.graphics.drawable.BitmapDrawable
 import android.view.View
 import android.view.ViewGroup
-import androidx.lifecycle.LifecycleOwner
-import kotlinx.coroutines.flow.collect
+import androidx.annotation.VisibleForTesting
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.flow.filterNotNull
 import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.launch
 import mozilla.components.browser.state.search.SearchEngine
 import mozilla.components.concept.menu.Orientation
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mozilla/fenix/search/toolbar/SearchSelectorToolbarAction.kt
```diff
@@ -71,19 +58,41 @@ class SearchSelectorToolbarAction(
         }
     }
 
-    override fun bind(view: View) = Unit
+    override fun bind(view: View) {
+        // It may happen that this View is binded multiple times.
+        // Prevent launching new coroutines for every time this is binded and only update the icon once.
+        if (updateIconJob?.isActive != true) {
+            updateIconJob = (view as? SearchSelector)?.toScope()?.launch {
```
Mugurell, line 65: Used this form 
> CoroutineScope.launch { store.flow() .. }

instead of

> View.consumeFrom(Store, LifecycleOwner, (State))

more to just have a flow on which we can easily use the `ifChanged` operator and not having to write it ourselves.

app/src/test/java/org/mozilla/fenix/search/toolbar/SearchSelectorToolbarActionTest.kt
```diff
@@ -65,7 +83,6 @@ class SearchSelectorToolbarActionTest {
             SearchSelectorToolbarAction(
                 store = store,
                 menu = menu,
-                viewLifecycleOwner = lifecycleOwner
```
mcarare, line 68: I think you could also remove the mock from declarations and also the `MockedLifecycleOwner` class.

mcarare (approved):

PR #25543, opened 2022-06-06

MickeyMoz:


jonalmeida:
For now, I think we should suppress the deprecation until we have a fix for https://github.com/mozilla-mobile/android-components/issues/12292 

https://github.com/mozilla-mobile/android-components/issues/12292

PR #25609, opened 2022-06-14
For #22186 - Scroll to collection after creating one

### LINKED ISSUE #22186: [Bug]: "Jump back in" and "Recent bookmarks" options are focused when tapping on "View" after creating a collection
## Prerequisites
"Jump back in" and "Recent bookmarks" options are enabled on homescreen.

### Steps to reproduce

1. Open tabs tray and add a tab to collection 1.
2. Tap on "View" from the snackbar.

### Expected behaviour

The collection 1/new tab added is focused.

### Actual behaviour

"Jump back in" and "Recent bookmarks" options are focused.

### Device name

Pixel 3

### Android version

Android 12

### Firefox release type

Firefox Nightly

### Firefox version

95.0a1, 94 Beta, 94.1.0

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-18486)

sarah541:
Fixes #22186

### Pull Request checklist

1. click on Show All Checks,


gabrielluong:
@Mugurell If you don't mind, are you willing to be the reviewer on this? Thanks.

Mugurell (approved):

Mugurell (approved):
This does not mimic exactly the previous implementation - which would scroll to the newly created collection and highlight that (in maybe a series of already existing collections) but is an improvement and a step in that direction.
We'll have to get back to implementing the same functionality as before after the entire homescreen is migrated to Compose.

Mugurell (changes requested):
Played with this more I saw that scrolling works if adding to an existing collection but not if adding to a new one.

https://user-images.githubusercontent.com/11428869/177530346-de260c6d-c100-4849-badc-b58834e1d99b.mov

Mugurell:
Tested this more and saw that the failures to scroll are actually intermittent, not tied to a particular scenario.
With some logs I saw that getting the `collectionPosition` to scroll to may fail and then because of [this null check](https://github.com/mozilla-mobile/fenix/pull/25609/files#diff-41731714e97083a1b73584da2837ab11c22796c25b892b1e86ce4190ebba2633R570) the code for scrolling is not actually called.

The previous implementation got broken without us knowing because the `collectionPosition` was hardcoded to be a specific index and with all the customization of the homescreen a hardcoded position would work anymore.

I'd think if calculating the `collectionPosition` fails intermittently it's because of a race with between trying to scroll and adding the elements to the RecyclerView adapter so it might require a bigger effort to fix and I'm not sure how close we are to move everything to compose and if working on fixing this for xml has a good ROI.

@gabrielluong Do you know what dependencies still exist to fully migrate the homescreen to compose? Is this still a priority?

sarah541:
> Tested this more and saw that the failures to scroll are actually intermittent, not tied to a particular scenario. With some logs I saw that getting the `collectionPosition` to scroll to may fail and then because of [this null check](https://github.com/mozilla-mobile/fenix/pull/25609/files#diff-41731714e97083a1b73584da2837ab11c22796c25b892b1e86ce4190ebba2633R570) the code for scrolling is not actually called.
> 
> The previous implementation got broken without us knowing because the `collectionPosition` was hardcoded to be a specific index and with all the customization of the homescreen a hardcoded position would work anymore.
> 
> I'd think if calculating the `collectionPosition` fails intermittently it's because of a race with between trying to scroll and adding the elements to the RecyclerView adapter so it might require a bigger effort to fix and I'm not sure how close we are to move everything to compose and if working on fixing this for xml has a good ROI.
> 
> @gabrielluong Do you know what dependencies still exist to fully migrate the homescreen to compose? Is this still a priority?

@Mugurell Will adding a delay of certain millisecond when looking for position solve the issue?

Mugurell:
> @Mugurell Will adding a delay of certain millisecond when looking for position solve the issue?
Sounds fair given that previously [a 100 ms delay was used](https://github.com/mozilla-mobile/fenix/blob/6d6b81bfe21ab280e063f1afe184bc4c342e1f4f/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt#L1188) and this would be just a temporary solution.

Mugurell:
> @Mugurell This happens because every time we hit `view` on snackbar, a new instance of `HomeFragment` is created which shows the `RecyclerView` from the top.

Thank you for looking into this.
Sounds tricky so maybe it can be left for later.

Issue #24815, opened 2022-04-15
Intermittent UI test failure - installAddonTest 


sv-ohorvath:
### Firebase Test Run:
https://console.firebase.google.com/project/moz-fenix/testlab/histories/bh.66b7091e15d53d45/matrices/5922064655051906308/executions/bs.62cb9788baaa1f11
### Stacktrace:
`androidx.test.espresso.IdlingResourceTimeoutException: Wait for [javaClass] to become idle timed out
at dalvik.system.VMStack.getThreadStackTrace(Native Method)
at java.lang.Thread.getStackTrace(Thread.java:1538)
at androidx.test.espresso.base.DefaultFailureHandler.getUserFriendlyError(DefaultFailureHandler.java:16)
at androidx.test.espresso.base.DefaultFailureHandler.handle(DefaultFailureHandler.java:36)
at androidx.test.espresso.ViewInteraction.waitForAndHandleInteractionResults(ViewInteraction.java:106)
at androidx.test.espresso.ViewInteraction.check(ViewInteraction.java:31)
at org.mozilla.fenix.ui.robots.SettingsSubMenuAddonsManagerRobot.assertAddonIsInstalled(SettingsSubMenuAddonsManagerRobot.kt:191)
at org.mozilla.fenix.ui.robots.SettingsSubMenuAddonsManagerRobot.verifyAddonIsInstalled(SettingsSubMenuAddonsManagerRobot.kt:67)
at org.mozilla.fenix.ui.SettingsAddonsTest$installAddonTest$3.invoke(SettingsAddonsTest.kt:95)
at org.mozilla.fenix.ui.SettingsAddonsTest$installAddonTest$3.invoke(SettingsAddonsTest.kt:82)
at org.mozilla.fenix.ui.robots.ThreeDotMenuMainRobot$Transition.openAddonsManagerMenu(ThreeDotMenuMainRobot.kt:344)
at org.mozilla.fenix.ui.SettingsAddonsTest.installAddonTest(SettingsAddonsTest.kt:82)`
### Build: 
4/15 debug

┆Issue is synchronized with this [Jira Story](https://mozilla-hub.atlassian.net/browse/FNXV2-20182)

sv-ohorvath:
The video shows a "Failed to install ublock origin" snackbar, maybe we should catch that.

sv-ohorvath:
[...]

Issue #17416, opened 2021-01-12
Intermittent UI test failure - SmokeTest.noCrashWithAddonInstalledTest


sv-ohorvath:
### Firebase Test Run:
Failed 1 out of 2 tries:
https://console.firebase.google.com/project/moz-fenix/testlab/histories/bh.66b7091e15d53d45/matrices/8052980965356929009/executions/bs.3df2fa55dee69f2a/testcases/2/test-cases
### Stacktrace:
`androidx.test.espresso.IdlingResourceTimeoutException: Wait for [SessionLoadedIdlingResource] to become idle timed out
at dalvik.system.VMStack.getThreadStackTrace(Native Method)
at java.lang.Thread.getStackTrace(Thread.java:1538)
at androidx.test.espresso.base.DefaultFailureHandler.getUserFriendlyError(DefaultFailureHandler.java:16)
at androidx.test.espresso.base.DefaultFailureHandler.handle(DefaultFailureHandler.java:36)
at androidx.test.espresso.ViewInteraction.waitForAndHandleInteractionResults(ViewInteraction.java:106)
at androidx.test.espresso.ViewInteraction.check(ViewInteraction.java:31)
at org.mozilla.fenix.ui.robots.BrowserRobot.verifyUrl(BrowserRobot.kt:65)
at org.mozilla.fenix.ui.SmokeTest$noCrashWithAddonInstalledTest$11.invoke(SmokeTest.kt:613)
at org.mozilla.fenix.ui.SmokeTest$noCrashWithAddonInstalledTest$11.invoke(SmokeTest.kt:39)
at org.mozilla.fenix.ui.robots.ThreeDotMenuMainRobot$Transition.openReportSiteIssue(ThreeDotMenuMainRobot.kt:268)
at org.mozilla.fenix.ui.SmokeTest.noCrashWithAddonInstalledTest(SmokeTest.kt:612)`
### Build: 
Debug 1/12

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-9366)

sv-ohorvath:
Couldn't reproduce or find a cause for this. Since it didn't fain in more than a month, I think we can close it and reopen if needed.

sv-ohorvath:
[...]

PR #25613, opened 2022-06-14
For #24843, #24815, #17416: removed idling resources. Retrying add-on installation if it fails.

sv-ohorvath:
For #24843, #24815, #17416 
- when the RetryTestRule was invoked, some idling resources remained registered, causing crashes because the fragment didn't exist anymore.
- some idling resources timed out when the permission prompt appeared, even if a call to unregister them was made. Don't know what happened there, so I removed and replaced them with the waitForExists method.
- I ran the tests hundreds of times, the last one is https://github.com/mozilla-mobile/fenix/runs/6794296501, the addons still fail to install after restarting the app, but overall the stability of the tests seems to be improved a bit compared to prior runs (because I removed the other possible reasons for failures): https://github.com/mozilla-mobile/fenix/runs/6723804592 or https://github.com/mozilla-mobile/fenix/runs/6726807316

Extra: I've modified a bit the RetryTestRule to unregister any idling resources left before retrying a test, to prevent these kinds of crashes in the future. @AaronMT If you don't agree with that, I can drop that commit.

The conclusion is there's still one flaky test remaining (I could leave it disabled if that's the safest), but the issue of failing to install addons only seems to be related to API 30: https://github.com/mozilla-mobile/fenix/runs/6606662528 and maybe it can't be avoided for now.

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/androidTest/java/org/mozilla/fenix/helpers/RetryTestRule.kt
+++ b/app/src/androidTest/java/org/mozilla/fenix/helpers/RetryTestRule.kt
@@ -11,7 +11,7 @@ import junit.framework.AssertionFailedError
 import org.junit.rules.TestRule
 import org.junit.runner.Description
 import org.junit.runners.model.Statement
-import java.lang.AssertionError
+import org.mozilla.fenix.helpers.IdlingResourceHelper.unregisterAllIdlingResources
 
 class RetryTestRule(private val retryCount: Int = 5) : TestRule {
 
@@ -23,30 +23,37 @@ class RetryTestRule(private val retryCount: Int = 5) : TestRule {
                     base.evaluate()
                     break
                 } catch (t: AssertionError) {
+                    unregisterAllIdlingResources()
                     if (i == retryCount) {
                         throw t
                     }
                 } catch (t: AssertionFailedError) {
+                    unregisterAllIdlingResources()
                     if (i == retryCount) {
                         throw t
                     }
                 } catch (t: UiObjectNotFoundException) {
+                    unregisterAllIdlingResources()
                     if (i == retryCount) {
                         throw t
                     }
                 } catch (t: NoMatchingViewException) {
+                    unregisterAllIdlingResources()
                     if (i == retryCount) {
                         throw t
                     }
                 } catch (t: IdlingResourceTimeoutException) {
+                    unregisterAllIdlingResources()
                     if (i == retryCount) {
                         throw t
                     }
                 } catch (t: RuntimeException) {
+                    unregisterAllIdlingResources()
                     if (i == retryCount) {
                         throw t
                     }
                 } catch (t: NullPointerException) {
+                    unregisterAllIdlingResources()
                     if (i == retryCount) {
                         throw t
                     }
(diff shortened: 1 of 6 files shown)


AndiAJ (approved):

app/src/androidTest/java/org/mozilla/fenix/helpers/RetryTestRule.kt
```diff
@@ -23,30 +23,51 @@ class RetryTestRule(private val retryCount: Int = 5) : TestRule {
...
+                    for (resource in IdlingRegistry.getInstance().resources) {
+                        IdlingRegistry.getInstance().unregister(resource)
+                    }
                     if (i == retryCount) {
                         throw t
                     }
                 } catch (t: NullPointerException) {
+                    for (resource in IdlingRegistry.getInstance().resources) {
+                        IdlingRegistry.getInstance().unregister(resource)
+                    }
```
AaronMT, line 70: For the above unregister calls, I'm not opposed to it but I'm not sure if this a good place. I'd prefer to keep them closer to the register calls in test. Do we know if we could place these preferably in the `tearDown` functions instead? Or write one function that checks for instances as you did above and and unregisters.

app/src/androidTest/java/org/mozilla/fenix/ui/robots/ThreeDotMenuMainRobot.kt
```diff
@@ -121,8 +122,22 @@ class ThreeDotMenuMainRobot {
...
+            )
+        } catch (e: AssertionError) {
+            mDevice.pressBack()
+            browserScreen {
+            }.openThreeDotMenu {
+                openAddonsSubList()
+                assertTrue(
+                    "Addon not listed in the Add-ons menu",
+                    mDevice.findObject(UiSelector().text(addonName)).waitForExists(waitingTime)
+                )
```
AaronMT, line 138: I'd like to get into the habit of using this model for retry rather than reduplicating code. We should add a retryCount global to constants now that I'm thinking about it, maybe set it to 3.  This will be helpful for individual checks.

https://github.com/mozilla-mobile/fenix/blob/main/app/src/androidTest/java/org/mozilla/fenix/helpers/RetryTestRule.kt#L21

e.g,
```Kotlin
for (i in 1..retryCount) {
  try {
   assertTrue()
   break
} catch (e: AssertionError) {
  if (I == retryCount) {
   throw t  
 } else {
   // actions, such as go back, click again
}
}
```
What do you think?

AaronMT (changes requested):

app/src/androidTest/java/org/mozilla/fenix/helpers/RetryTestRule.kt
sv-ohorvath: > I'd prefer to keep them closer to the register calls in test. Do we know if we could place these preferably in the `tearDown` functions instead?

Well, that's the problem: if the test fails before it can reach the unregister calls, it will restart the app and then it crashes. The tearDown methods aren't called until after the 3 retries are done. So we have to make sure to keep unregistering after the registration as before, and also to do this while in the RetryTestRule loop for exceptions like this.

app/src/androidTest/java/org/mozilla/fenix/ui/robots/ThreeDotMenuMainRobot.kt
sv-ohorvath: Sure, I will try that

sv-ohorvath: done.

app/src/androidTest/java/org/mozilla/fenix/helpers/RetryTestRule.kt
sv-ohorvath: move this code to the IdlingResourceHelper.kt class

AaronMT (approved):

Issue #25545, opened 2022-06-06
Task continuity: ~800ms regression on 50% of COLD MAIN first frame start ups (5/28 regression with bimodal behavior)


mcomella:
_see https://github.com/mozilla-mobile/fenix/issues/22144 for bimodal results in COLD VIEW, which is very likely unrelated_

For our COLD MAIN first frame start up test, our replicates are usually scattered evenly (I verified this behavior on the Nighlies 4/29, 5/15, 5/17, 5/18, 5/19, 5/21, and 5/25-5/27). For example, here is the Nightly 5/27 performance test:
![Figure_1](https://user-images.githubusercontent.com/759372/172214246-98b64ebe-f3c8-4c33-995d-5cc27a51b8d0.png)

However, Nightly 5/28 introduces bimodal results:
![Figure_12png](https://user-images.githubusercontent.com/759372/172214679-9555f483-ac39-4544-9742-54cb2c245be8.png)

I verified:
- this bimodal behavior continues on 5/29.
- From 5/30-6/1, we don't see bimodal behavior. Instead, we see a major regression such that all replicates are much longer
- We return to bimodal behavior from 6/2-6/5, where our test results end

**Bimodal results are undesirable** because:
- users are experiencing fast start ups roughly half the time and slower start ups the other half. Why not figure out how to get users the fast start up all the time?
- it can cause false positive regressions making it difficult to identify true regressions (because if we get each of a low result and a high result roughly 50% of the time like in this bug, the performance test result will jump between each regularly since we use the median)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-20662)

mcomella:
One way to debug this would be to take multiple profiles (5?), find one of each behavior (since it occurs roughly 50% of the time, it shouldn't take too long assuming running the profiler doesn't change the behavior), and compare them to see what's different.

Another method would be to install the 5/27 nightly and install the 5/28 nightly and see if anything is visually different on start up to the homescreen (e.g. maybe this regression was introduced by a new view/feature being added to the homescreen).

cpeterso:
I think this is the mozilla-central change log between 5/27 and 5/28:

https://hg.mozilla.org/mozilla-central/pushloghtml?fromchange=914ead848de431dcd0f42c2fda013018704b29b9&tochange=aebf5d6aabfedd693883dc66c07ed2a7b9400e02

MatthewTighe:
[...]

PR #25669, opened 2022-06-17
No issue: create add_secrets to copy local secrets into correct files

MatthewTighe:
During attempts at debugging #25545, we have had some trouble with reproducing the issue regularly. One reason for that might be that builds from different people might include different subsets of features due to the lack of secrets included in the build process. 

It's an error-prone and time-consuming process to copy each secret into the appropriate place in our build scripts, so this attempts to remove that obstacle by making it easier to insert all our secrets during a local build.

I deleted the `.adjust_token` file because it appeared to only contain a placeholder, which was required in local release builds that were lacking the real secret. However, I was concerned that it would get overwritten when this script was used, and since it was already in the source tree it wouldn't be ignored correctly. This could lead to the secret being leaked pretty easily. Instead, the dummy file is created automatically during release builds where it is not present.

Excuse my amateurish python skills 😂 

1. click on Show All Checks,


app/build.gradle
```diff
@@ -318,6 +318,8 @@ android.applicationVariants.all { variant ->
             buildConfigField 'String', 'ADJUST_TOKEN', '"' + token + '"'
             println "(Added from .adjust_token file)"
         } catch (FileNotFoundException ignored) {
+            // Release builds will crash without a dummy file
+            new File("${rootDir}/.adjust_token").text = "--"
```
boek, line 322: If I recall correctly, crashing was intentional to make sure we weren't shipping a build without one?

tools/add_secrets.py
```diff
@@ -0,0 +1,56 @@
...
+except IndexError:
+    print(
+        """
+        Expected 1 argument defined build variant. Probably 'nightly', 'beta', or 'release'.
+        For example: python3 tools/add_secrets.py nightly
+        """
+    )
+    sys.exit(2)
+
+# These mappings are borrowed from taskcluster/transforms/build.py and should be kept in sync.
```
boek, line 34: I wonder if there is an easy way if we can share the mapping between the scripts so we don't have to manually keep them in sync 🤔

app/build.gradle
MatthewTighe: I'll re-examine this PR once the repo migration will happen, but I believe this comment was about the actual build process and not the finished artifact. IIRC I had removed the empty `.adjust_token` file as part of this PR but we still wanted it present and empty during pipeline builds.

tools/add_secrets.py
MatthewTighe: I suppose a shared text file could be parsed between them. I'll consider if that's worth the effort when I come back to this PR

Issue #24740, opened 2022-04-12
In history/bookmark view searching should activate the correct search view


rocketsroger:
Description:
As a user when I tap the search button in the full history or bookmarks view the browser should update my search bar view to search that information.

Acceptance Criteria:
In the full history or bookmarks view when the search button is tapped focus moves to the search bar and the correct type of search is activated

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-20141)

PR #25707, opened 2022-06-21
Issue #24740: Use unified search bar in bookmark search

rocketsroger:
### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/res/navigation/nav_graph.xml
+++ b/app/src/main/res/navigation/nav_graph.xml
@@ -274,12 +274,6 @@
             app:popUpToInclusive="true" />
 
         <action
-            android:id="@+id/action_global_search_dialog"
-            app:destination="@id/searchDialogFragment"
-            app:popUpTo="@id/searchDialogFragment"
-            app:popUpToInclusive="true" />
-
-        <action
             android:id="@+id/action_historyFragment_to_syncedHistoryFragment"
             app:destination="@id/syncedHistoryFragment" />
     </fragment>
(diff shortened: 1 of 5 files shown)


rocketsroger:
https://user-images.githubusercontent.com/6579541/174865777-85e271be-f2a8-4653-a2f9-04f010cc6359.mp4

app/src/main/res/navigation/nav_graph.xml
```diff
@@ -335,6 +335,11 @@
             app:destination="@id/bookmarkSearchDialogFragment"
             app:popUpTo="@id/bookmarkSearchDialogFragment"
             app:popUpToInclusive="true" />
+         <action
+            android:id="@+id/action_global_search_dialog"
```
gabrielluong, line 339: Just noticed this. Why do we need to add an action here if this is already global? https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/res/navigation/nav_graph.xml#37

Otherwise, we should follow the id convention - `action_bookmarkFragment_to_searchDialogFragment`. However, I don't think we need to add this action and just call the global action. We will probably also want to fix this for the history fragment as well.

gabrielluong (approved):

app/src/main/res/navigation/nav_graph.xml
rocketsroger: You're probably right.  I'll check.  Thanks

rocketsroger: Removed.  Good catch, thanks

Issue #25132, opened 2022-05-05
Errors in creating and maintaining UiAutomation connection in Instrumented tests


AaronMT:
Intermittently, an early `getInstrumentation()` call in `beforeActivityLaunched` for our `HomeActivityTestRule.kt`  will throw a `java.lang.RuntimeException` with a `Error while connecting UiAutomation`. This seems to happen more frequent on 1.4.0 and espresso 3.4.0 libraries [2] - just getting a ticket on file here to wrap this call and try again (although I don't know if multiple attempts to get a handle on the running instrumentation will do anything). This is for executing a shell command to set a timeout on long tap. 

Stack:

```Kotlin
java.lang.RuntimeException: Error while connecting UiAutomation@afc456f[id=-1, flags=0]
at android.app.UiAutomation.connect(UiAutomation.java:259)
at android.app.Instrumentation.getUiAutomation(Instrumentation.java:2176)
at androidx.test.uiautomator.UiDevice.getUiAutomation(UiDevice.java:1129)
at androidx.test.uiautomator.QueryController.<init>(QueryController.java:95)
at androidx.test.uiautomator.UiDevice.<init>(UiDevice.java:109)
at androidx.test.uiautomator.UiDevice.getInstance(UiDevice.java:261)
at org.mozilla.fenix.helpers.HomeActivityTestRuleKt.setLongTapTimeout(HomeActivityTestRule.kt:78)
```

[1]  [Firebase Log](https://console.firebase.google.com/u/0/project/moz-fenix/testlab/histories/bh.66b7091e15d53d45/matrices/5262751007738123275/executions/bs.a7816152eb688985/testcases/1/test-cases)
[2] Ticket filed as https://github.com/android/android-test/issues/1305

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-20390)

AndiAJ:
Corresponding intermitent tests:
`customTabShareTextTest` #25099
`privateBrowsingUseAppWhileTabIsCrashedTest` #25076
`verifyExpandedCollectionItemsTest` #25075
`verifyContextOpenImageNewTab` #25054

AndiAJ:
Hey @AaronMT , it's still happening.
`closeAllTabsTest` #25341
`openMainMenuSettingsItemTest` #25342

sv-ohorvath:
New cases started to appear:
`saveLoginFromPromptTest ` #25414
[...]

PR #25719, opened 2022-06-22
No issue: Refactor UIDevice calls in UIAutomator code

AaronMT:
As part of investigating into https://github.com/mozilla-mobile/fenix/issues/25132 I'd like to see if part of the problem is when and where we initialize UiDevice singletons.

Copying from the example in https://developer.android.com/training/testing/other-components/ui-automator, they initialize in `@Before` methods using a `lateinit` variable. There's a likelihood this is a device problem on FTL, but this is a cleanup too. I've defined a reusable `TestHelper` object variable for re-use in robots and refactored the tests to use the style mentioned above if they require use of a `UiDevice` object. It's likely that the tests themselves can also use the `TestHelper` object too.

### DIFF
--- a/app/src/androidTest/java/org/mozilla/fenix/glean/BaselinePingTest.kt
+++ b/app/src/androidTest/java/org/mozilla/fenix/glean/BaselinePingTest.kt
@@ -26,6 +26,7 @@ import okhttp3.mockwebserver.RecordedRequest
 import org.json.JSONObject
 import org.junit.Assert.assertTrue
 import org.junit.Assert.assertFalse
+import org.junit.Before
 import org.junit.BeforeClass
 import org.junit.Rule
 import org.junit.Test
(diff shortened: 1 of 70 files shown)


sv-ohorvath:
@AaronMT This looks good to me, a much-needed cleanup if not else :) I wonder if we don't want to use the lateinit var in the HomeActivityTestRule class as well, since the issue seems to be coming from there mostly.

LE:
 I ran the most affected tests with these changes and I've still found some errors: 
https://console.firebase.google.com/project/moz-fenix/testlab/histories/bh.66b7091e15d53d45/matrices/6175033360166842780/details?stepId=bs.20b670e0c891f095&testCaseId=1

https://console.firebase.google.com/project/moz-fenix/testlab/histories/bh.66b7091e15d53d45/matrices/6532648874751012823/details?stepId=bs.44345cb3d0f4db19&testCaseId=1

more here: https://github.com/mozilla-mobile/fenix/runs/7088955342

AaronMT:
> @AaronMT This looks good to me, a much-needed cleanup if not else :) I wonder if we don't want to use the lateinit var in the HomeActivityTestRule class as well, since the issue seems to be coming from there mostly.
> 
> LE: I ran the most affected tests with your changes to save you time, and I've still found the same and new errors: https://console.firebase.google.com/project/moz-fenix/testlab/histories/bh.66b7091e15d53d45/matrices/6175033360166842780/details?stepId=bs.20b670e0c891f095&testCaseId=1
> 
> https://console.firebase.google.com/project/moz-fenix/testlab/histories/bh.66b7091e15d53d45/matrices/6532648874751012823/details?stepId=bs.44345cb3d0f4db19&testCaseId=1
> 
> more here: https://github.com/mozilla-mobile/fenix/runs/7088955342

Yeah, unrelated to the FTL issue. Investigation with Firebase Community and Google on-going, but this a nice refactor.

PR #25738, opened 2022-06-23
Fixes #25545: wait to draw Pocket until after first frame

### LINKED ISSUE #25545: Task continuity: ~800ms regression on 50% of COLD MAIN first frame start ups (5/28 regression with bimodal behavior)

MatthewTighe:
This PR resolves this perf regression in two ways:

- Pocket content was rendering too early, regularly increasing the time to layout the home screen. This forces pocket content to wait until after the initial render pass
- Additionally, pocket content would often load first. Since the content is focusable, the home screen RecyclerView would automatically scroll the content to the center. This was mostly unrelated, but the perf fix caused this behavior to become consistent.

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/home/pocket/PocketCategoriesViewHolder.kt
```diff
@@ -49,11 +49,17 @@ class PocketCategoriesViewHolder(
...
+            .observeAsComposableState { state -> state.firstFrameDrawn }.value ?: false
+
         val categories = components.appStore
             .observeAsComposableState { state -> state.pocketStoriesCategories }.value
         val categoriesSelections = components.appStore
             .observeAsComposableState { state -> state.pocketStoriesCategoriesSelections }.value
 
+        // Perceived performance can be improved by waiting to draw pocket content until after the
+        // first frame is drawn. See https://github.com/mozilla-mobile/fenix/issues/25545 for details.
+        if (!homeScreenReady) return
```
MozillaNoah, line 62: why not instead just wrap the `Column` with this `if`?

app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesViewHolder.kt
```diff
@@ -52,9 +52,15 @@ class PocketStoriesViewHolder(
...
 
+        val homeScreenReady = components.appStore
+            .observeAsComposableState { state -> state.firstFrameDrawn }.value ?: false
+
         val stories = components.appStore
             .observeAsComposableState { state -> state.pocketStories }.value
 
+        // Perceived performance can be improved by waiting to draw pocket content until after the
+        // first frame is drawn. See https://github.com/mozilla-mobile/fenix/issues/25545 for details.
+        if (!homeScreenReady) return
```
MozillaNoah, line 63: same here

app/src/main/res/layout/top_placeholder_item.xml
```diff
@@ -2,6 +2,10 @@
...
    - License, v. 2.0. If a copy of the MPL was not distributed with this
    - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+
+<!-- RecyclerViews will focus the first element that is focusable. This prevents views below
+the top from being scrolled to unintentionally. -->
 <View xmlns:android="http://schemas.android.com/apk/res/android"
     android:layout_width="match_parent"
-    android:layout_height="0dp"/>
+    android:layout_height="0dp"
+    android:focusable="true" />
```
MozillaNoah, line 11: would this be a short term stop-gap until the Homepage is all Compose, basically?

MatthewTighe: There was a bit of discussion during this process about how best to handle this in general. One suggestion, for example, was to use an invisible skeleton of roughly the same size as the final content. We could then wait until after first frame to populate the skeleton with content.

I don't know off the top of my head if this issue will reappear when the whole screen is compose. If items in large `LazyList`s are focused automatically we may have to do something similar.

app/src/main/java/org/mozilla/fenix/home/pocket/PocketCategoriesViewHolder.kt
MatthewTighe: Less indentation I guess? I've usually just prefer early returns, stylistically. There's additionally reasoning for the choice in the other comment, and I think it would be good to keep the two in sync if we decide to change it.

app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesViewHolder.kt
MatthewTighe: Here, there's additional logic that records telemetry when `stories` has a value. There can definitely be cases where `stories` is usable but we are still waiting for the homescreen to be ready. I don't think we want to record that telemetry until the stories are actually show, based on the code and comments.

Amejia481 (approved):

app/src/main/java/org/mozilla/fenix/home/pocket/PocketCategoriesViewHolder.kt
```diff
@@ -49,11 +49,17 @@ class PocketCategoriesViewHolder(
...
+        val homeScreenReady = components.appStore
+            .observeAsComposableState { state -> state.firstFrameDrawn }.value ?: false
+
         val categories = components.appStore
             .observeAsComposableState { state -> state.pocketStoriesCategories }.value
         val categoriesSelections = components.appStore
             .observeAsComposableState { state -> state.pocketStoriesCategoriesSelections }.value
 
+        // Perceived performance can be improved by waiting to draw pocket content until after the
+        // first frame is drawn. See https://github.com/mozilla-mobile/fenix/issues/25545 for details.
```
mcomella, line 61: non-blocking: I'd prefer if this comment elaborated on why this issue happens and how this solution addresses the issue because:
1. devs won't necessarily go to the issue even if the information would be useful to them
2. if the solution changes they're not likely to update the issue but they'd update the explanation in a comment
3. there is generally a lot of irrelevant context in issues that make digging for the answers to those two questions challenging.

More generally, I prefer making the source code an elaborate source of truth and leaving the issue tracker as a supplement because the source tends to be more accurate & to-the-point.

app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesViewHolder.kt
```diff
@@ -52,9 +52,15 @@ class PocketStoriesViewHolder(
...
         val horizontalPadding = dimensionResource(R.dimen.home_item_horizontal_margin)
 
+        val homeScreenReady = components.appStore
+            .observeAsComposableState { state -> state.firstFrameDrawn }.value ?: false
+
         val stories = components.appStore
             .observeAsComposableState { state -> state.pocketStories }.value
 
+        // Perceived performance can be improved by waiting to draw pocket content until after the
+        // first frame is drawn. See https://github.com/mozilla-mobile/fenix/issues/25545 for details.
```
mcomella, line 62: non-blocking: if you take my suggestion above to elaborate, I'd prefer if these two comments point to a central elaboration comment rather than these two comments be duplicated and elaborate

mcomella (commented):
The fenix team has ownership of keeping the product performant to meet the product's goals so the performance team (or myself) generally does not need to approve changes for performance unless there is something specific you need our expertise for. As such, I'm going to remove my review request but I happened to skim the review and include a drive-by comment too.

PR #25765, opened 2022-06-27
Update compose version and run UI tests on Firebase

AndiAJ:
### Pull Request checklist

1. click on Show All Checks,


buildSrc/src/main/java/Dependencies.kt
```diff
@@ -20,7 +20,7 @@ object Versions {
     const val detekt = "1.19.0"
     const val jna = "5.8.0"
 
-    const val androidx_compose = "1.1.1"
```
gabrielluong, line 23: We don't use alpha versions in the tree.

AndiAJ: I know @gabrielluong ☺️ 

This is only a check, we just want to see if the upcoming release will truly fix some inconsistencies we encounter from time to time in our UI tests.

AndiAJ:
Based on the result, it looks like the next compose release could fix the `Wait for [Compose-Espresso link] to become idle timed out` and `Compose-Espresso link isIdleNow() is returning true, but a message indicating that the resource has transitioned from busy to idle was never sent.` inconsistencies encountered in our UI tests.

PR #25799, opened 2022-06-29
Bug 1776830 - Split `ship` release graph into 2: `promote` and `ship`

JohanLorenzo:
### Pull Request checklist

- [X] **Tests**: Tested as a staging release: 
  * `promote` graph: https://firefox-ci-tc.services.mozilla.com/tasks/groups/ViRQM3VfS8aKB2VA2TlyFg
  * `ship` graph: https://firefox-ci-tc.services.mozilla.com/tasks/groups/HnIHeXaDR4S_IM5pgtVBew (beetmover error is unrelated to these changes) 
- [ ] **Screenshots**: N/A
- [ ] **Accessibility**: N/A


taskcluster/ci/beetmover/kind.yml
```diff
@@ -24,6 +24,7 @@ only-for-build-types:
 job-template:
     attributes:
         artifact_map: taskcluster/fenix_taskgraph/manifests/fenix_candidates.yml
+        shipping_phase: ship
```
jcristau, line 27: shouldn't pushing to candidates be part of promote?

taskcluster/fenix_taskgraph/transforms/multi_dep.py
```diff
@@ -23,7 +23,8 @@ def build_name_and_attributes(config, tasks):
         }
         primary_dep = task["primary-dependency"]
         copy_of_attributes = primary_dep.attributes.copy()
-        task.setdefault("attributes", {}).update(copy_of_attributes)
+        copy_of_attributes.update(task.get("attributes", {}))
+        task["attributes"] = copy_of_attributes
```
jcristau, line 27: I'd rename `copy_of_attributes` to `attributes` since it's no longer just a copy.

taskcluster/ci/beetmover/kind.yml
jcristau, line 27: In gecko the yml has "shipping-phase" with a hyphen, and that gets turned into a "shipping_phase" attribute by the "task" transform.  Can we align here somehow?

JohanLorenzo: I was confused by the name of the `artifact_map` too! As far as I can see, the only beetmover task publishes APKs to the `release` folder and not the `candidates` one. I'm okay renaming the yaml file, if you think that's the right thing to do.

JohanLorenzo: Yeah, I first started using `shipping-phase` but it turns out it's only defined in `gecko_taskgraph` not in the standalone one. It's actually transformed by the `task` transform. @ahal Do you think we should port these bits of the tranform to standalone taskgraph?

[1] https://searchfox.org/mozilla-central/rev/0069a0cdc3cd136ea7411521219df6665f260be7/taskcluster/gecko_taskgraph/transforms/task.py#1958

taskcluster/fenix_taskgraph/transforms/multi_dep.py
JohanLorenzo: Good call, done!

taskcluster/ci/beetmover/kind.yml
jcristau: Seems worth renaming the map for clarity (possibly in another PR).  And then as a followup, I think we should consider actually pushing to candidates in promote and adding a push-to-releases task in the ship phase.

JohanLorenzo: I definitely agree, I was thinking the same 👍

JohanLorenzo: Done as part of this PR.

taskcluster/fenix_taskgraph/target_tasks.py
```diff
@@ -20,12 +20,22 @@ def index_exists(index_path, reason=""):
...
-@_target_task("release")
+@_target_task("promote")
 def target_tasks_default(full_task_graph, parameters, graph_config):
+    return _target_release_promotion(full_task_graph, parameters, shipping_phase="promote")
 
-    # TODO Use shipping-phase
+
+@_target_task("ship")
+def target_tasks_ship(full_task_graph, parameters, graph_config):
+    return _target_release_promotion(full_task_graph, parameters, shipping_phase="ship")
```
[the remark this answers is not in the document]
JohanLorenzo: Good news, we're already doing it with the current implementation! In gecko[1], we only target tasks for a given phase. Then, taskgraph is smart enough to bring back upstream tasks if they were never scheduled before. This mean, tasks in the `ship` phase will bring back tasks from the `push` phase if they weren't scheduled before and if they're required by the `ship` phase. 

That's what makes taskgraph powerful: you just tell it what tasks you want to target and it ensure you have all dependencies scheduled either in the upcoming graph or in a previous one.

[1] For instance, in the dekstop push graph: https://searchfox.org/mozilla-central/rev/4654ce1e24a6af17bc96ab60f1f70c218755142f/taskcluster/gecko_taskgraph/target_tasks.py#644-648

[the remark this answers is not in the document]
JohanLorenzo: Done!

taskcluster/ci/build/kind.yml
```diff
@@ -165,7 +165,7 @@ jobs:
 
     beta:
         attributes:
-            release-type: beta
+            release_type: beta
```
jcristau, line 168: Hmm so I don't think this actually matches gecko?  In gecko `release_type` is a decision task parameter, as far as I can tell, not a task attribute.  So I'm not sure it's worth doing this renaming here for that reason.

JohanLorenzo: You're right. That was some bad wording. What I meant was: in gecko, attributes use underscores instead of dashes. I'll change the commit message to make it less ambiguous.

jcristau (changes requested):
I don't really understand the release-type rename, though I guess I'm ok with it.  The bit in util/scriptworker.py is wrong though afaict.

JohanLorenzo:
All right, I prefer removing my last commit then. I don't want to confuse people. Plus, I realized there is another attribute to take into account `build-type`. I'll do that in a separate PR.

taskcluster/fenix_taskgraph/util/scriptworker.py
```diff
@@ -40,7 +40,7 @@ def generate_beetmover_upstream_artifacts(
         "attributes.artifact_map",
         "artifact map",
         **{
-            "release-type": config.params["release_type"],
+            "release_type": config.params["release_type"],
```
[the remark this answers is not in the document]
JohanLorenzo: Resolved by removing commit.

taskcluster/fenix_taskgraph/util/scriptworker.py
```diff
@@ -149,7 +149,7 @@ def generate_beetmover_artifact_map(config, job, **kwargs):
         "attributes.artifact_map",
         job["label"],
         **{
-            "release-type": config.params["release_type"],
+            "release_type": config.params["release_type"],
```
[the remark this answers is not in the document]
JohanLorenzo: Resolved by removing commit.

jcristau (approved):
Thanks

PR #25871, opened 2022-07-04
Closes #25825 :Disable marketing telemetry until privacy pop window approved in MozillaOnline builds

### LINKED ISSUE #25825: [Feature:China] New regulations in China restrict to get Android_ID by adjust
Recently we Beijing office have received several app stores' feedback about the adjust sdk trying to acquire the android id, which has been restricted by new regulations, saying that **android id can only be acquired after user's approval of privacy notice**, and this action of trying to get android id should also be stated in the privacy notice.

(Users in China need to deal with a pop window of privacy notice before they use any app. For fenix, see this PR: https://github.com/mozilla-mobile/fenix/pull/17043)

According to this, we need to 

1. start AdjustMetricsService after the click of approval of privacy notice pop window from users.

or

2. use `adjustConfig.setPlayStoreKidsAppEnabled(true);` in MozillaOnline builds (`if (Config.channel.isMozillaOnline)`) to prevent adjust to acquire android id according to the document of adjust: https://help.adjust.com/en/article/apps-for-children-android-sdk . This might be the easiest way but a little bit confused in code.

Currently  we build 102.1.1 of fenix locally to disable adjust (we don't have the token so adjust won't work when we build fenix locally) otherwise we cannot release it in many app stores in China. So we want to resolve this asap, thanks!

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-20861)

rxumoz:


app/src/main/java/org/mozilla/fenix/home/mozonline/PrivacyContentDisplayHelper.kt
```diff
@@ -44,15 +47,19 @@ fun showPrivacyPopWindow(context: Context, activity: Activity) {
...
     val builder = AlertDialog.Builder(activity)
         .setPositiveButton(
-            context.getString(R.string.privacy_notice_positive_button),
-            { _, _ ->
-                context.settings().shouldShowPrivacyPopWindow = false
+            context.getString(R.string.privacy_notice_positive_button)
+        ) { _, _ ->
+            context.settings().shouldShowPrivacyPopWindow = false
+            context.settings().preferences.edit {
+                putBoolean(context.getString(R.string.pref_key_marketing_telemetry), true)
```
Mugurell, line 54: For consistency with other preferences usages `isMarketingTelemetryEnabled` should be a `var` and then here we'd have: 
```suggestion
            context.settings().isMarketingTelemetryEnabled = true
```

app/src/main/java/org/mozilla/fenix/home/mozonline/PrivacyContentDisplayHelper.kt
```diff
@@ -44,15 +47,19 @@ fun showPrivacyPopWindow(context: Context, activity: Activity) {
...
             }
-        )
+            if (context.settings().isMarketingTelemetryEnabled) {
```
Mugurell, line 56: If we set `isMarketingTelemetryEnabled = true` above (or even by editing the preference manually) then this `if` check is not needed.

app/src/main/java/org/mozilla/fenix/home/mozonline/PrivacyContentDisplayHelper.kt
```diff
@@ -44,15 +47,19 @@ fun showPrivacyPopWindow(context: Context, activity: Activity) {
...
+                context.components.analytics.metrics.start(MetricServiceType.Marketing)
+            }
+        }
         .setNeutralButton(
-            context.getString(R.string.privacy_notice_neutral_button_2),
-            { _, _ -> exitProcess(0) }
-        )
+            context.getString(R.string.privacy_notice_neutral_button_2)
+        ) { _, _ -> exitProcess(0) }
```
Mugurell, line 62: nit: Change not needed to fix the issue. Cleanups could be part of separate commits to keep the changes small and easy to follow.

Mugurell (approved):
LGTM with some nits.

rxumoz:
@Mergifyio backport releases_v103.0.0

PR #25934, opened 2022-07-07
For #25933 - Ensure sponsored tiles are enable if enrolled in active experiment

### LINKED ISSUE #25933: Sponsored tiles disabled in active experiment
https://mozilla-hub.atlassian.net/browse/EXP-2591

In 101 and prior, the Nimbus implementation in Fenix held the responsibility for caching the experimentation values fetched through Nimbus. These values were only fetched once during the initialization sequence of Fenix.
In 102, the Nimbus startup sequence in Fenix was changed to account for updates to Nimbus’ own internal caching layer and the startup was executed earlier in the application. This caused a problem with the values that are fetched by the application because there was a race condition which led to the application fetching and storing the values from Nimbus before it was fully ready and these values were not updated to the latest values from Nimbus when its startup was completed because they were only fetched once. As a result, it is believed that this caused the experiment to not enable the feature and for the telemetry values to be reported too early with the incorrect values from Nimbus.

Our fix removes a call to an event handler that would report the home screen telemetry metrics that was called before the Nimbus updates have been applied. This should fix the incorrect telemetry metrics reporting from being called too early before Nimbus is ready.
In addition, we removed the underlying Nimbus value caching in the Fenix application and ensure we always fetching the latest value directly from Nimbus

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-20944)

gabrielluong:
Fixes #25933 

### Pull Request checklist

1. click on Show All Checks,

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
+++ b/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
@@ -428,8 +428,6 @@ open class FenixApplication : LocaleAwareApplication(), Provider {
                 onNimbusStartupAndUpdate()
             }
         })
-
-        onNimbusStartupAndUpdate()
     }
 
     private fun onNimbusStartupAndUpdate() {
--- a/app/src/main/java/org/mozilla/fenix/components/settings/FeatureFlagPreference.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/settings/FeatureFlagPreference.kt
@@ -26,15 +26,12 @@ fun featureFlagPreference(key: String, default: Boolean, featureFlag: Boolean) =
 
 private class LazyPreference(val key: String, val default: () -> Boolean) :
     ReadWriteProperty<PreferencesHolder, Boolean> {
-    private val property: ReadWriteProperty<PreferencesHolder, Boolean> by lazy {
-        booleanPreference(key, default())
-    }
 
     override fun getValue(thisRef: PreferencesHolder, property: KProperty<*>) =
-        this.property.getValue(thisRef, property)
+        thisRef.preferences.getBoolean(key, default())
 
     override fun setValue(thisRef: PreferencesHolder, property: KProperty<*>, value: Boolean) =
-        this.property.setValue(thisRef, property, value)
+        thisRef.preferences.edit().putBoolean(key, value).apply()
 }
 
 /**
--- a/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
+++ b/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
@@ -1233,9 +1233,8 @@ class Settings(private val appContext: Context) : PreferencesHolder {
         default = false
     )
 
-    private val homescreenSections: Map<HomeScreenSection, Boolean> by lazy {
+    private val homescreenSections: Map<HomeScreenSection, Boolean> get() =
         FxNimbus.features.homescreen.value().sectionsEnabled
-    }
 
     var historyMetadataUIFeature by lazyFeatureFlagPreference(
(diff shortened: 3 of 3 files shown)


rocketsroger (approved):
🚢 🚢 🔥

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -1233,9 +1233,8 @@ class Settings(private val appContext: Context) : PreferencesHolder {
         default = false
     )
 
-    private val homescreenSections: Map<HomeScreenSection, Boolean> by lazy {
+    private val homescreenSections: Map<HomeScreenSection, Boolean> get() =
```
Amejia481, line 1236: When this was implemented Nimbus didn't have a layer of caching, for this reason we made lazy.

Amejia481: As the value was lazy and we introduced `onNimbusStartupAndUpdate` very earlier when nimbus was not ready, we were caching the first value and never updating it again.

app/src/main/java/org/mozilla/fenix/components/settings/FeatureFlagPreference.kt
```diff
@@ -26,15 +26,12 @@ fun featureFlagPreference(key: String, default: Boolean, featureFlag: Boolean) =
...
-        booleanPreference(key, default())
-    }
 
     override fun getValue(thisRef: PreferencesHolder, property: KProperty<*>) =
-        this.property.getValue(thisRef, property)
+        thisRef.preferences.getBoolean(key, default())
 
     override fun setValue(thisRef: PreferencesHolder, property: KProperty<*>, value: Boolean) =
-        this.property.setValue(thisRef, property, value)
+        thisRef.preferences.edit().putBoolean(key, value).apply()
```
jhugman, line 34: Please could you file issues to rename this class to something more descriptive.

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -428,8 +428,6 @@ open class FenixApplication : LocaleAwareApplication(), Provider {
                 onNimbusStartupAndUpdate()
             }
         })
-
-        onNimbusStartupAndUpdate()
```
jhugman, line 432: Please could you file an issue to rename this (because it's no longer being called at startup).

jhugman (commented):
Some issues to file in the review comments.

gabrielluong:
@mergify backport [releases_v102.0.0, releases_v103.0.0]

Amejia481:
Just for documentation purpose or future debugging, adding GL overview.
 

> In 101 and prior, Fenix held the responsibility for caching the experimentation values fetched through Nimbus. These values were only fetched once during the initialization sequence of Fenix. In 102, the Nimbus startup sequence in Fenix was changed to account for updates to Nimbus’ own internal caching layer and the startup was executed earlier in the application. This caused a problem with the values that are fetched by the application because there was a race condition which led to the application fetching and storing the values from Nimbus before it was fully ready and these values were not updated to the latest values from Nimbus when its startup was completed because they were only fetched once. As a result, this caused the experiment to not enable the feature and for the telemetry values to be reported too early with the incorrect values from Nimbus.Our fix removes an extraneous call to an event handler that would report the home screen telemetry metrics that was called before the Nimbus updates have been applied. This should fix the incorrect telemetry metrics reporting from being called too early before Nimbus is ready. In addition, we removed the underlying Nimbus value caching in the Fenix application and ensure we always fetching the latest value directly from Nimbus.

gabrielluong:
> Are there any tests written for this?

Unfortunately not.

PR #25879, opened 2022-07-04
Synced history sticky collapsable

mavduevskiy:
Draft pull request

### Pull Request checklist

- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features. In addition, it includes a screenshot of a successful [accessibility scan](https://play.google.com/store/apps/details?id=com.google.android.apps.accessibility.auditor&hl=en_US) to ensure no new defects are added to the product.

### To download an APK when reviewing a PR:
1. click on Show All Checks,
2. click Details next to "Taskcluster (pull_request)" after it appears and then finishes with a green checkmark,
3. click on the "Fenix - assemble" task, then click "Run Artifacts".
4. the APK links should be on the left side of the screen, named for each CPU architecture

PR #25941, opened 2022-07-08
Close #25937: update history element design

### LINKED ISSUE #25937: Update history item design
Update history item UI.
Jira ticket: [link](https://mozilla-hub.atlassian.net/browse/FNXV2-19087)
Design: [link](https://www.figma.com/file/SO8Cd5lnmgNqJQY9KpORc9/History%2C-Bookmarks%2C-Downloads?node-id=879%3A41209)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-20946)

mavduevskiy:
Closes #25937:

Updates the UI of history list item and a header margin.
You might notice that the `delete icon` is a bit larger (16dp vs current 12) in the design and has `textSecondary` colour. I will check first if the colour is the correct one and if we don't want to make the cross icon bigger across the whole app.

A part of breaking down [this](https://github.com/mozilla-mobile/fenix/pull/25879) synced history too big of a PR.
Jira ticket: [link](https://mozilla-hub.atlassian.net/browse/FNXV2-19087)
Design: [link](https://www.figma.com/file/SO8Cd5lnmgNqJQY9KpORc9/History%2C-Bookmarks%2C-Downloads?node-id=879%3A41209)

Old|New
---|---
old_no_lb|new_no_lb
old_lb|new_lb


app/src/main/res/layout/library_site_item.xml
```diff
@@ -15,7 +15,7 @@
             android:id="@+id/icon"
             android:layout_width="@dimen/history_favicon_width_height"
             android:layout_height="@dimen/history_favicon_width_height"
```
gabrielluong, line 17: I generally understand we are using 8dp here instead of 16dp because the width/height for the icon is bigger than 24x24. Admittedly, it is notable characteristic of how this xml was written, but I don't have any designs for this checkmark to see if we could just adjust all the icons here to 24x24 and remove the padding in the ImageView.

app/src/main/res/layout/library_site_item.xml
```diff
@@ -77,9 +78,10 @@
             android:layout_height="@dimen/glyph_button_height"
             android:background="?android:attr/selectableItemBackgroundBorderless"
             android:contentDescription="@string/content_description_menu"
-            app:srcCompat="@drawable/ic_menu"
+            app:srcCompat="@drawable/ic_close"
```
gabrielluong, line 81: We might want to update the id here since it's a close button now instead of an overflow menu.

I will also note that layout width/height is also unusual being 40x40 instead of 24x24. Normally, I would think about just changing this to 24x24, but then we might need to call `increaseTapArea` so maybe that was the consideration done here. In Compose, we were able to leave the icon buttons as 24x24 because the underlying internal compose implementation will ensure that the tap area satisfies accessibility requirements.

app/src/main/res/layout/recently_closed_nav_item.xml
```diff
@@ -49,7 +49,8 @@
         android:id="@+id/recently_closed_tabs_description"
         android:layout_width="0dp"
         android:layout_height="wrap_content"
-        android:layout_marginStart="12dp"
+        android:layout_marginStart="24dp"
+        android:layout_marginTop="2dp"
```
gabrielluong, line 53: You will want to update the text size for the header and description here as well and consider dropping the margin top as discussed above.

app/src/main/res/layout/synced_history_nav_item.xml
```diff
@@ -18,7 +18,7 @@
         android:id="@+id/icon"
         android:layout_width="@dimen/history_favicon_width_height"
         android:layout_height="@dimen/history_favicon_width_height"
```
gabrielluong, line 20: It's possible that we can convert both this and recently_closed_nav_item to use 24x24 instead of 40x40 width and height and drop the padding if you want to simplify this. There's no image switching or tap handling for both of these ImageViews so that should simplify matters if anyone ever wonders why the margin start is 8dp instead of 16dp and doesn't realize that you're accounting for the 40x40 size.

topotropic:
@mavduevskiy please use the bigger close icon (16x16) with primary color; for reference https://www.figma.com/file/SO8Cd5lnmgNqJQY9KpORc9/History%2C-Bookmarks%2C-Downloads?node-id=867%3A28219 Thank you!

topotropic:
we can replace the smaller close icon with the bigger one (16x16) across the app

app/src/main/res/layout/library_site_item.xml
mavduevskiy: We could do that by just manually changing the vector size. But to keep the 16 padding with would need to make `ImageSwitcher.width` of 24, which would make the checkmark icon smaller as well, almost twice as small.

gabrielluong: Right, I am not expecting any changes for this at this time.

app/src/main/res/layout/recently_closed_nav_item.xml
mavduevskiy: done

app/src/main/res/layout/synced_history_nav_item.xml
mavduevskiy: done

app/src/main/res/layout/library_site_item.xml
```diff
@@ -58,7 +58,8 @@
             android:id="@+id/url"
```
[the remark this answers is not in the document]
mavduevskiy: done

gabrielluong (approved):

Issue #25955, opened 2022-07-08
Refactor `RecentSyncedTabsFeature` to rely on `SyncStore` instead of implementing `SyncedTabsView`


MatthewTighe:
Now that `SyncStore` is available, `RecentSyncedTabsFeature` can be refactored to move away from the `SyncedTabsView` pattern which should simplify the overall design of the feature.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-20958)

PR #25956, opened 2022-07-08
Fixes #24823: remove recent synced tab when logged out

### LINKED ISSUE #24823: Recent synced tab does not appear after signing in
Steps to reproduce:
- Make sure there isn't a synced tab on home screen
- Sign in to sync

Expected behavior:
The recent synced tab placeholders appear, and then a synced tab is shown.

Actual behavior:
A recent synced tab does not appear until the home screen is reloaded again.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-20187)

MatthewTighe:
Depends on https://github.com/mozilla-mobile/android-components/pull/12449

This adds `SyncStore` observation to the recent synced tab feature, in order to successful determine when the Sync state becomes logged out. I'll have a follow-up PR immediately after this to address https://github.com/mozilla-mobile/fenix/issues/25955. I initially had that issue addressed as part of this patch but wanted to slim it down as it includes quite a bit more refactoring.

1. click on Show All Checks,


app/src/main/java/org/mozilla/fenix/home/recentsyncedtabs/RecentSyncedTabFeature.kt
```diff
@@ -87,22 +97,37 @@ class RecentSyncedTabFeature(
...
-        if (store.state.recentSyncedTabState == RecentSyncedTabState.Loading) {
-            store.dispatch(AppAction.RecentSyncedTabStateChange(RecentSyncedTabState.None))
+        if (appStore.state.recentSyncedTabState == RecentSyncedTabState.Loading) {
+            appStore.dispatch(AppAction.RecentSyncedTabStateChange(RecentSyncedTabState.None))
         }
     }
 
     override fun start() {
         syncedTabsFeature.start()
+        statusScope = syncStore.flowScoped { flow ->
```
MozillaNoah, line 112: `flowScoped` is based off of `MainScoped`, so you might not need to keep the coroutine as a local variable in this instance, since it'll cancel the job in `onDestroy`, if that lifecycle works as a substitute for `stop`.

app/src/main/java/org/mozilla/fenix/components/BackgroundServices.kt
```diff
@@ -183,6 +191,10 @@ class BackgroundServices(
 
         SyncedTabsIntegration(context, accountManager).launch()
 
+        syncStoreSupport = SyncStoreSupport(syncStore, lazyOf(accountManager)).also {
```
MozillaNoah, line 194: can I get a TLDR with what `SyncStoreSupport` does, since it looks like it's on the AC side?

MozillaNoah (commented):
looks good so far! will wait to officially approve it until the AC changes land

app/src/main/java/org/mozilla/fenix/home/recentsyncedtabs/RecentSyncedTabFeature.kt
MatthewTighe: I actually ran into some trouble with these coroutines, and so did a fair bit of investigation that led to this implementation. I'm not sure it's necessarily the best or in style with other examples in the codebase, but I'll break some of it down and if you have suggestions or examples I'm happy to update it.

- Just as a point of clarity: `MainScope` is not tied to lifecycles at all. It just creates a new `CoroutineScope` on the main thread. This could even cause objects to survive longer than anticipated, as I think any running coroutines in the created scope would keep an object from being collected.

- `flowScoped` exposes an optional `owner: LifecycleOwner` parameter which can achieve the behavior you described in your comment. There's additional code in the `Store.flow` extension that it uses to cancel the flow when a lifecycle moves into the `Destroyed` state. 

- _However_, `RecentSyncedTabFeature` is currently set using `ViewBoundFeatureWrapper`, which was in keeping with existing code in the `HomeFragment`. The whole purpose of `ViewBoundFeatureWrapper` is to keep a strong reference to objects across destructive view lifecycle events. This can prevent expensive operations like reloading a bunch of synced tabs. That said, the purpose of `ViewBoundFeatureWrapper` is the thing I understand least here so maybe it's not even necessary in this case.

- Normally in Fragments I've seen it suggested to use `viewLifecycleOwner` for automatic cancellation and re-collection of flows.

All of the above combines to mean that the normal lifecycle owner I'd use here for collecting these flows (viewLifecycleOwner) is destroyed but the `Feature` outlives it. The flow collection will stop as soon as view is destroyed and not started again. This was of manually handling the scope lifecycles prevents all that.

app/src/main/java/org/mozilla/fenix/components/BackgroundServices.kt
MozillaNoah: sounds good, thank you!

app/src/main/java/org/mozilla/fenix/home/recentsyncedtabs/RecentSyncedTabFeature.kt
MozillaNoah: gotcha gotcha. I appreciate the thorough write-up for all of the context! I'm sure you've gone through several iterations of stuff with this overall effort.

Mugurell: Regarding the `ViewBoundFeatureWrapper` it was added in https://github.com/mozilla-mobile/android-components/issues/1943 to fix leaks or even crashes from features that updates various Fragment Views and for that receive them as parameters.

It's a simple convenience to ensure that when the Fragment that creates a certain feature referencing a View is added to the backstack - and so the layout and Views are destroyed - the feature is also destroyed.

As a general rule of thumb if a fragment needs a reference to a View then it should be wrapped in a `ViewBoundFeatureWrapper`.

MatthewTighe: @Mugurell thanks for the clarity around `ViewBoundFeatureWrapper`. `RecentSyncedTabFeature` doesn't need a reference to a view, so it should be safe to stop wrapping it based on your rule of thumb. I'm still not sure I fully understand the case where `ViewBoundFeatureWrapper` prevents leaks, however. Would it be a scenario like this?

```
class MyFragment : Fragment() {
    val feature = MyFeature()

        feature.view = view
    }
}
```

So is `ViewBoundFeatureWrapper` basically equivalent to the following?
```
class MyFragment : Fragment() {
    val feature: MyFeature? = null

        feature = MyFeature(view)
    }

        feature = null
    }
}
```

I've added a separate commit that simplifies the feature's logic in handling the coroutine scopes. To my understanding these updates aren't at risk of creating leaks, but I would be interesting in getting a second opinion there 😂. I'll squash the second commit if it looks safe.

@MozillaNoah as an aside, the A-C patch has landed!

Mugurell: Seeing many features in `HomeFragment` being used inside of the said `ViewBoundFeatureWrapper` though they don't need a View and have injected a reference to the root layout for the required View argument.
While not optimal (based on the above) it should't cause any issues with them being initialized in `onCreateView` and then destroyed in `onDestroyView` (by the wrapper).
Though them being a `LifecycleAwareFeature` that only works while the fragment is in foreground would be enough.

MozillaNoah (approved):
🚢 🚢 🚢

PR #26028, opened 2022-07-14
Fixes #25955: refactor RecentSyncedTabsFeature to use SyncStore

### LINKED ISSUE #25955: Refactor `RecentSyncedTabsFeature` to rely on `SyncStore` instead of implementing `SyncedTabsView`

MatthewTighe:
### Pull Request checklist

### QA

- [ ] **QA Needed**

1. click on Show All Checks,

### GitHub Automation
Fixes #25955


app/src/main/java/org/mozilla/fenix/home/recentsyncedtabs/RecentSyncedTabFeature.kt
```diff
@@ -29,49 +27,70 @@ import org.mozilla.fenix.GleanMetrics.RecentSyncedTabs
...
         if (appStore.state.recentSyncedTabState == RecentSyncedTabState.None) {
             appStore.dispatch(AppAction.RecentSyncedTabStateChange(RecentSyncedTabState.Loading))
         }
     }
 
-    override fun displaySyncedTabs(syncedTabs: List<SyncedDeviceTabs>) {
-        val syncedTab = syncedTabs
+    private suspend fun dispatchSyncedTabs() {
+        val syncedTab = storage.getSyncedDeviceTabs()
             .filterNot { it.device.isCurrentDevice || it.tabs.isEmpty() }
```
MozillaNoah, line 94: you may not need the filter on `it.device.isCurrentDevice`. Based on my work in SyncedTabs in the tabs tray, I believe this happens during `storage.getSyncedDeviceTabs()`

MozillaNoah (approved):
This looks great! Just wondering if the synced tabs in the tabs tray should eventually be moved over to this implementation pattern as well.

MatthewTighe:
> This looks great! Just wondering if the synced tabs in the tabs tray should eventually be moved over to this implementation pattern as well.

I'm probably biased since I worked on the SyncStore, but I found this to be more straightforward than subclassing `SyncedTabView` and using the `SyncedTabFeature` hierarchy. In general I found the old implementation to have a greater number of less-obvious edge cases. So I'd recommend doing it! 

I think long-term we will probably look into doing more substantial refactors of more of the Sync components like `FxaAccountManager` etc with the intent that some of that behavior will be moved into or at least abstracted by the SyncStore.

app/src/main/java/org/mozilla/fenix/home/recentsyncedtabs/RecentSyncedTabFeature.kt
MatthewTighe: I think you're right, I may have been using multiple releases channels of Firefox on the same device and thought they were the same Sync device (since a nightly release and a debug release would show as different devices, for instance). 

That said, I don't think there's anything in the `getSyncedDeviceTabs` name that suggests it would guarantee that to _always_ be the case. I'm happy to change it, but I think in the interest of not being too introspective into the abstraction barrier that I'll otherwise leave it for now. It could also prevent future bugs if the underlying implementation were to change.

MozillaNoah: I believe it's from the callstack of: `SyncedTabsStorage.getSyncedDeviceTabs()` -> `SyncedTabsStorage.syncClients()` -> `ConstellationState.otherDevices`

Either way, not a blocker for this by any means, and I agree that the name/callstack would ideally be more explicit about the filtering used.

Issue #26034, opened 2022-07-15
[meta] Refactor the Wallpaper feature to prep for A-C migration


MatthewTighe:
This issue is intended as a top-level issue tracking work to refactor the Wallpaper feature to host its state in the `AppStore` and initiate the actions it undertakes through the `Middleware` pattern, as well as abstract Fenix-specific implementations out. This has several benefits:

- architectural consistency
- allows clients to focus on state and interactions instead of implementation details
- creates an abstraction barrier

There are several subissues that will need to be completed for this to be considered done:

- [x] https://github.com/mozilla-mobile/fenix/issues/25985
- [x] https://github.com/mozilla-mobile/fenix/issues/26029
- [x] https://github.com/mozilla-mobile/fenix/issues/26052
- [ ] https://github.com/mozilla-mobile/fenix/issues/26030
- [ ] https://github.com/mozilla-mobile/fenix/issues/26033
- [x] https://github.com/mozilla-mobile/fenix/issues/26245

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21015)

MozillaNoah:
You could optionally format this issue like the [meta] issues [here](https://github.com/mozilla-mobile/fenix/issues/24339) and [here](https://github.com/mozilla-mobile/fenix/issues/21318).

MatthewTighe:
> You could optionally format this issue like the [meta] issues [here](https://github.com/mozilla-mobile/fenix/issues/24339) and [here](https://github.com/mozilla-mobile/fenix/issues/21318). That way each individual issue has a built-in status indicator without having to click into them.

Brilliant, thanks for pointing it out! I'll reformat this one if I've got time and definitely remember this for the future

cpeterso:
Moved to bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1809805

Change performed by the [Move to Bugzilla add-on](https://addons.mozilla.org/en-US/firefox/addon/move-to-bugzilla/).

Issue #26030, opened 2022-07-15
Create a WallpaperMiddleware for the AppStore


MatthewTighe:
A WallpaperMiddleware should be added to the AppStore that can initiate side effects. This should take the place of any direct interactions with the `WallpaperManager` and instead react automatically to `Action`s intercepted that update the `WallpaperState` added in https://github.com/mozilla-mobile/fenix/issues/26029.

This may be dependent or cause conflicts with https://github.com/mozilla-mobile/fenix/issues/25985

This should include an action to refresh the current wallpaper bitmap when configuration changes occur.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21014)

cpeterso:
Moved to bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1809806

Change performed by the [Move to Bugzilla add-on](https://addons.mozilla.org/en-US/firefox/addon/move-to-bugzilla/).

PR #26051, opened 2022-07-15
Fixes #26029: Add wallpaper state to AppStore

### LINKED ISSUE #26029: Add Wallpaper state to the `AppStore`
Add a `WallpaperState` data class to the `AppStore`, along with appropriate actions to update the state.

- current wallpaper
- current wallpaper bitmap, if any
- available wallpapers

The wallpaper manager can dispatch updates to the store directly for now.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21013)

MatthewTighe:
This is part of a larger refactor being tracked by https://github.com/mozilla-mobile/fenix/issues/26034.

Note: this will not refresh wallpapers on orientation changes. This was more important for Turning Red wallpapers than it is for the standard Firefox wallpapers, and will be addressed by https://github.com/mozilla-mobile/fenix/issues/26052 and https://github.com/mozilla-mobile/fenix/issues/26030 in follow-ups shortly.

### Pull Request checklist

### QA

- [ ] **QA Needed**

1. click on Show All Checks,

### GitHub Automation
Fixes #26029


app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperState.kt
```diff
@@ -0,0 +1,29 @@
...
+ */
+data class WallpaperState(
+    val currentWallpaper: Wallpaper,
+    val currentWallpaperBitmap: Bitmap?,
+    val availableWallpapers: List<Wallpaper>,
+) {
+    companion object {
+        val default = WallpaperState(
+            currentWallpaper = Wallpaper.Default,
+            currentWallpaperBitmap = null,
```
gabrielluong, line 25: Could we not derive the bitmap from `currentWallpaper`? It seems suspicious to hold into memory the Bitmap twice if we can derive it from the other existing state.

JohanLorenzo:
Hey there! I'm sorry to jump in this PR. I'm just rebasing it because https://github.com/mozilla-mobile/fenix/pull/26073 reached `main`. It renamed the expected check from `pr-complete` to `complete-pr`. This impacts mergify and it may prevent you from merging this PR. That's why the rebase is needed. Let me know if you have any questions.

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperState.kt
csadilek: @MatthewTighe Yes, I agree with the trade-off you're discussing here, and in particular, that we would want to prevent having to read the bitmap from disk every time the home screen is opened. How big are these bitmaps in memory compared to the thumbnails we're already caching?

csadilek: I don't think it's feasible to have wallpapers of 10s of MB in memory or read them from disk every single time. I would think the wallpapers are just a few 100KBs on disk before rendering and we could cache that data?

csadilek: (Also, happy to talk through options offline, of course.)

MatthewTighe: 10s of MBs is a pretty large overestimation now that you mention it. I was able to find the old Turning Red assets and it looks like they're ~3.5 MB. I expect something along those lines to be towards the higher end of what we'd include in the future. I see that 200KB size for the wallpapers on disk as well. Any hypotheses around why they'd look so bloated when inspecting them through the debugger?

I'll look into `IconMemoryCache`. Would your suggestion then be to avoid the Store/Middleware pattern entirely for wallpapers? At a glance, it seems like using a cache like that would require the same hierarchy of utility classes that we've basically got now. That said, I think there's still improvements in terms of abstractions and APIs that could be made in that case.

Mugurell: > 10s of MBs is a pretty large overestimation now that you mention it. I was able to find the old Turning Red assets and it looks like they're ~3.5 MB. I expect something along those lines to be towards the higher end of what we'd include in the future. I see that 200KB size for the wallpapers on disk as well. Any hypotheses around why they'd look so bloated when inspecting them through the debugger?

There's no direct relation between the size on disk and the size of a displayed image.
When displayed we need to feed information about each pixel of the screen so for a regular phone format we'd have:
- a fullscreen image 1080 x 1800px (excluding system insets) means ~2 million pixels
- ARGB.8888 - 4 bytes per pixel (if we need transparency) will mean 8MB in memory.

And then there are of course devices with higher resolutions (and probably faster NANDs also), maybe even double which will mean double the size of the bitmap in memory.

The size in memory of a bitmap can be queried with the `.byteCount` method.

Regarding passing a bitmap through the AppStore even though shallow copies are used and we'd avoid churning through memory maybe just an identifier of the image would help achieve the end goal. 
And then use maybe the `WallpaperManager` as a wrapper over all types of storage where the image may be - internet / memory / disk.
There are some fast NANDs out there and even ignoring recent flagships most devices typically reach a few hundreds MB/s so if we can ensure the images are optimized to have an as low as possible size on disk (<1MB) (there are even AI lossless optimizers and Android also supports [more efficient file formats like WebP](https://developers.google.com/speed/webp/docs/webp_lossless_alpha_study)) then I'd say loading the images from disk should be fine. We could try profiling to be sure and avoid premature optimizations.

Mugurell: `IconMemoryCache` seems like just a LRU cache there to prevent new internet calls to download recent favicons and it has a 25MB limit for all in-memory bitmaps until evicting old icons to make room for the new.
The LRU cache scenario doesn't seem applicable here.
If we decide to keep the wallpapers in memory keeping just one should be enough.

csadilek: > I'll look into IconMemoryCache. Would your suggestion then be to avoid the Store/Middleware pattern entirely for wallpapers? 

@MatthewTighe Ah no, sorry, I just used it as an example for where we already cache bitmaps. It's fine to keep it in AppState as long as we're storing/caching the contents of the file, not the rendered bytes, see also @Mugurell 's explanation above.

> ...then I'd say loading the images from disk should be fine. We could try profiling to be sure and avoid premature optimizations.

@Mugurell We know based on telemetry that reading even relatively small files can be delayed on devices, and since this file can't change and the image is needed every time we open home, I do think we should cache it.

> In place of the bitmap we could cache the image asset - serialize it to a byteArray to keep in memory (<1MB) and then use a variant of [BitmapFactory.decodeByteArray()](https://developer.android.com/reference/android/graphics/BitmapFactory#decodeByteArray(byte%5B%5D,%20int,%20int,%20android.graphics.BitmapFactory.Options)) to decode the bitmap.

👍

MatthewTighe: I'm not sure I see a real benefit to storing this data in the AppStore if a `WallpaperManager` is still required in interactions to load and decode bitmaps, especially in the case of thumbnails when it makes less sense to have them stored in memory for the lifetime of the app.
```
override fun onCreateView() {
    consumeFrom(components.appStore, viewLifecycleOwner) {
        components.wallpaperManager.decodeWallpaper(appStore.currentWallpaperBytes)
    }
}
```

Simpler:
```
override fun onCreateView() {
    components.wallpaperManager.decodeCurrentWallpaper()
}
```

I still think there's improvements to be made to class names and API surfaces, however. Maybe a `WallpaperUseCase` would be more appropriate?

MatthewTighe:
Based on the conversations above and offline, I decided to remove the `currentBitmap` property from the `WallpaperState`. The final design will require consumers of `WallpaperState` to use a `WallpaperUseCase` (or similar) to retrieve and render bitmaps, with any underlying caching mechanism delegated to the UseCase implementation.

This should otherwise be ready for review!

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperManager.kt
```diff
@@ -122,6 +131,8 @@ class WallpaperManager(
             wallpapers.find { it.name == currentWallpaper }
                 ?: fileManager.lookupExpiredWallpaper(currentWallpaper)
                 ?: defaultWallpaper
+        }.also {
+            appStore.dispatch(AppAction.WallpaperAction.UpdateCurrentWallpaper(it))
```
jonalmeida, line 135: Why do we need to do this here when you do it in the setter of `currentWallpaper`?

MatthewTighe: Setters aren't called during initialization, which is the only time this is called. There's another weird instance of it being called in `switchToNextWallpaper`, where `currentWallpaper` isn't actually set. In both cases, `currentWallpaper` is being set during `updateWallpaper`, which is called from client code. These are meant as fallbacks since it felt like bad practice to assume client behavior.

In any case, `updateWallpaper` will be removed in a follow-up PR, and a lot of these oddities will be cleaned up then. Trying to practice breaking patches into small pieces, and you can see some other updates in the pipeline on the [parent issue](https://github.com/mozilla-mobile/fenix/issues/26034)

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperManager.kt
```diff
@@ -40,10 +43,14 @@ class WallpaperManager(
     val wallpapers = allWallpapers
         .filter(::filterExpiredRemoteWallpapers)
         .filter(::filterPromotionalWallpapers)
+        .also {
+            appStore.dispatch(AppAction.WallpaperAction.UpdateAvailableWallpapers(it))
+        }
```
MozillaNoah, line 48: is this intended to be a first time initialization Action?

MozillaNoah (approved):
This looks good to me. This PR has had a good amount of iteration and looks like there's follow-up issues filed for the previous feedback.

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperManager.kt
MatthewTighe: Probably initially, yeah. I picture this either being run automatically by the `WallpaperUseCase` when it is initialized, or as a side-effect of the `WallpaperUseCase` being called from `FenixApplication::onCreate`. I can the imagine the timing of when it's dispatched changing based on longer-term feature discovery, but it's hard to see it happening more than once during the lifetime of the app.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -995,12 +999,20 @@ class HomeFragment : Fragment() {
...
+                    // as the default is applied already on xml by default.
+                    val currentWallpaper = state.wallpaperState.currentWallpaper
+                    if (currentWallpaper != WallpaperManager.defaultWallpaper) {
+                        requireComponents.wallpaperManager.updateWallpaper(
+                            binding.wallpaperImageView,
+                            currentWallpaper
+                        )
+                    }
+                }
+                .launchIn(viewLifecycleOwner.lifecycleScope)
```
gabrielluong, line 1015: This is for my own learning. I am wondering is there any difference between doing `launchIn(viewLifecycleOwner.lifecycleScope)` vs `flowScoped(viewLifecycleOwner.lifecycleScope)`

MatthewTighe: Just as a point of clarity, `flowScoped` accepts a `LifecycleOwner` instead of a `CoroutineScope`. `Store.flow()`, which is used here, is also used within `flowScoped`. `Store.flow()` does some automatic lifecycle tracking and will only `emit` values while the provided lifecycle is not destroyed. It's also important to note that `Store.flow` does _not_ start emitting values again if the `LifecycleOwner` moves back into ` Start`ed state.

`flowScoped` will create a new coroutine scope using the `Main` dispatcher, and then use the that new scope to launch. `launchIn` will instead just launch a coroutine using the provided scope.

Using an injected `CoroutineScope` with `launchIn` is easier to test, IMO. It allows more fine-grained control of the coroutine context because you have a direct reference to the scope, instead of on abstract usages of the test coroutine tooling like `Dispatchers.setMain` etc. I also thought it made it easier to avoid problems with the `LifecycleOwner` auto-subscription stuff mentioned at the top, by basically requiring an active `Lifecycle.lifecycleScope`.

MatthewTighe: It's also first thing in the morning, and I thought this was in reference to the flow collection happening in `RecentSyncedTabFeature` in my other PR. Since this is the `HomeFragment`, most of the above doesn't apply as much, but I was probably following my own example from `RecentSyncedTabFeature`. Normally, I think the convention for collecting Flows in Fragments is to use `Fragment.consumeFrom`, but here I wanted a direct Flow type so I could filter it using `Flow.ifChanged`

PR #26073, opened 2022-07-18
Bug 1779797 - part 2-6: Let mergify automatically merge l10n bumps to main

JohanLorenzo:
Follows https://github.com/mozilla-mobile/fenix/pull/26046 up. More context in https://github.com/mozilla-mobile/fenix/pull/26046#issuecomment-1186888559. Tested in https://github.com/JohanLorenzo/fenix/pull/3 - CI is red there because I took a shortcut by making the new `complete-push` task green without the rest of the graph being done. That said, the dependencies look correct there so I'm not concerned about this part. 

### Pull Request checklist

- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features. In addition, it includes a screenshot of a successful [accessibility scan](https://play.google.com/store/apps/details?id=com.google.android.apps.accessibility.auditor&hl=en_US) to ensure no new defects are added to the product.

### QA

- [ ] **QA Needed**

### To download an APK when reviewing a PR:
1. click on Show All Checks,
2. click Details next to "Taskcluster (pull_request)" after it appears and then finishes with a green checkmark,
3. click on the "Fenix - assemble" task, then click "Run Artifacts".
4. the APK links should be on the left side of the screen, named for each CPU architecture

### GitHub Automation

Used by GitHub Actions.

PR #26046, opened 2022-07-15
Bug 1779797 - Let mergify automatically approve l10n bumps on release…

JohanLorenzo:
… branches

Ports https://github.com/mozilla-mobile/android-components/pull/12509 to fenix.

### Pull Request checklist

- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features. In addition, it includes a screenshot of a successful [accessibility scan](https://play.google.com/store/apps/details?id=com.google.android.apps.accessibility.auditor&hl=en_US) to ensure no new defects are added to the product.

### QA

- [ ] **QA Needed**

### To download an APK when reviewing a PR:
1. click on Show All Checks,
2. click Details next to "Taskcluster (pull_request)" after it appears and then finishes with a green checkmark,
3. click on the "Fenix - assemble" task, then click "Run Artifacts".
4. the APK links should be on the left side of the screen, named for each CPU architecture

### GitHub Automation

Used by GitHub Actions.

PR #26148, opened 2022-07-22
Bug 1780740 - A-C bumps should auto-merge if CI is green on release branches

JohanLorenzo:
This PR is very similar to what https://github.com/mozilla-mobile/fenix/pull/26046 and https://github.com/mozilla-mobile/fenix/pull/26073 did with l10n bumps. However, I couldn't test these changes due to some limitations on both relbot and `automation/taskcluster/update_android_components.sh`. [Relbot hardcodes the repos names](https://github.com/mozilla-mobile/relbot/blob/21bb922118b011253fbdfb61651de373b54c7b32/src/relbot.py#L115-L118) and so does [update_android_components.sh](https://github.com/mozilla-mobile/fenix/blob/b4cf80a86db274b92a6fc97cbb3a656f22c86e60/automation/taskcluster/update_android_components.sh#L9). Although, I could change these hardcodes and deal with credentials, I think we can avoid it this time. Here's my reasoning: 

 * The mergify rules I changed are very similar to the l10n bumps. So there are much fewer unknowns than when I got started with https://github.com/mozilla-mobile/fenix/pull/26046, last week 
 * The logic of the mergify config file is still fresh in my mind.
 * I tested the negative regex on regex101.com
 * I double-checked every value I entered with existing PRs. 
 * I pledge to check the bumps are done correctly after this PR is merged. 

As a reminder, mergeday is monday, so we may avoid an uplift if this PR rides the train before mergeday.


jcristau (approved):
There's a typo in part 3's commit message ("are will be").

LGTM.

JohanLorenzo:
This worked well on both `main` (https://github.com/mozilla-mobile/fenix/pull/26244) and the current release branch (https://github.com/mozilla-mobile/fenix/pull/26231)

PR #26173, opened 2022-07-25
For #26172 - New CFR popup composable

### LINKED ISSUE #26172: CFR composable
Similar to https://github.com/mozilla-mobile/focus-android/issues/6409 Fenix also needs a CFR popup that allows for pinpoint accuracy when anchoring and also simplifies usage for the future.

Advantages to using a composable CFR popup that is drawn programatically:
- easy to enforce a cohesive and consistent design for all CFRs
- easy to anchor the arrow indicator to small targets
- eliminates color issues with the current approach that combines two images with slightly different colors.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21106)

Mugurell:
This will allow for pinpoint accuracy when anchoring and resolve any color
disparities between the popup body and the indicator arrow by having everything
drawn programmatically as one shape.

Because of the async nature of the values for insets and screen rotation
immediately after an orientation change the popup will automatically get
dismissed to prevent any anchoring issues.
While not ideal the effect of this is better than accepting layout issues after
orientation changes and is the same approach used for other of our popups.

As an example of how easy it is to control anchoring programatically:

https://user-images.githubusercontent.com/11428869/153623802-6f09556f-86f5-4aa3-bdec-1f349f8c7493.mp4

The typography to be used is still in discussion - https://mozilla-hub.atlassian.net/browse/FNXV2-20991?focusedCommentId=572965

### Pull Request checklist

### QA

- [ ] **QA Needed**

### GitHub Automation
Fixes #26172

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/compose/cfr/CFRPopup.kt
+++ b/app/src/main/java/org/mozilla/fenix/compose/cfr/CFRPopup.kt
@@ -1,0 +1,123 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.compose.cfr
+
+import android.view.View
+import androidx.annotation.VisibleForTesting
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.dp
+import java.lang.ref.WeakReference
+
+/**
+ * Properties used to customize the behavior of a [CFRPopup].
+ *
+ * @property popupWidth Width of the popup. Defaults to [CFRPopup.DEFAULT_WIDTH].
+ * @property indicatorDirection The direction the indicator arrow is pointing.
+ * @property dismissOnBackPress Whether the popup can be dismissed by pressing the back button.
+ * If true, pressing the back button will also call onDismiss().
+ * @property dismissOnClickOutside Whether the popup can be dismissed by clicking outside the
+ * popup's bounds. If true, clicking outside the popup will call onDismiss().
+ * @property overlapAnchor How the popup's indicator will be shown in relation to the anchor:
+ *   - true - indicator will be shown exactly in the middle horizontally and vertically
+ *   - false - indicator will be shown horizontally in the middle of the anchor but immediately below or above it
+ * @property indicatorArrowStartOffset Maximum distance between the popup start and the indicator arrow.
+ * If there isn't enough space this could automatically be overridden up to 0 such that
+ * the indicator arrow will be pointing to the middle of the anchor.
+ */
+data class CFRPopupProperties(
+    val popupWidth: Dp = CFRPopup.DEFAULT_WIDTH.dp,
+    val indicatorDirection: CFRPopup.IndicatorDirection = CFRPopup.IndicatorDirection.UP,
+    val dismissOnBackPress: Boolean = true,
+    val dismissOnClickOutside: Boolean = true,
+    val overlapAnchor: Boolean = false,
+    val indicatorArrowStartOffset: Dp = CFRPopup.DEFAULT_INDICATOR_START_OFFSET.dp,
+)
+
+/**
+ * CFR - Contextual Feature Recommendation popup.
+ *
+ * @param text [String] shown as the popup content.
+ * @param anchor [View] that will serve as the anchor of the popup and serve as lifecycle owner
+ * for this popup also.
+ * @param properties [CFRPopupProperties] allowing to customize the popup behavior.
+ * @param onDismiss Callback for when the popup is dismissed indicating also if the dismissal
+ * was explicit - by tapping the "X" button or not.
+ * @param action Optional other composable to show just below the popup text.
+ */
+class CFRPopup(
+    private val text: String,
+    private val anchor: View,
+    private val properties: CFRPopupProperties = CFRPopupProperties(),
+    private val onDismiss: (Boolean) -> Unit = {},
+    private val action: @Composable (() -> Unit) = {}
+) {
+    // This is just a facade for the CFRPopupFullScreenLayout composable offering a cleaner API.
+
+    @VisibleForTesting
+    internal var popup: WeakReference<CFRPopupFullScreenLayout>? = null
+
+    /**
+     * Construct and display a styled CFR popup shown at the coordinates of [anchor].
+     * This popup will be dismissed when the user clicks on the "x" button or based on other user actions
+     * with such behavior set in [CFRPopupProperties].
+     */
+    fun show() {
+        anchor.post {
+            CFRPopupFullScreenLayout(text, anchor, properties, onDismiss, action).apply {
+                this.show()
+                popup = WeakReference(this)
+            }
+        }
+    }
+
+    /**
+     * Immediately dismiss this CFR popup.
+     * The [onDismiss] callback won't be fired.
+     */
+    fun dismiss() {
+        popup?.get()?.dismiss()
+    }
+
+    /**
+     * Possible direction for the arrow indicator of a CFR popup.
+     * The direction is expressed in relation with the popup body containing the text.
+     */
+    enum class IndicatorDirection {
+        UP,
+        DOWN
+    }
+
+    companion object {
+        /**
+         * Default width for all CFRs.
+         */
+        internal const val DEFAULT_WIDTH = 335
+
+        /**
+         * Fixed horizontal padding.
+         * Allows the close button to extend with 10dp more to the end and intercept touches to
+         * a bit outside of the popup to ensure it respects a11y recommendations of 48dp size while
+         * also offer a bit more space to the text.
+         */
+        internal const val DEFAULT_HORIZONTAL_PADDING = 10
+
+        /**
+         * How tall the indicator arrow should be.
+         * This will also affect how wide the base of the indicator arrow will be.
+         */
+        internal const val DEFAULT_INDICATOR_HEIGHT = 15
+
+        /**
+         * Maximum distance between the popup start and the indicator.
+         */
+        internal const val DEFAULT_INDICATOR_START_OFFSET = 30
+
+        /**
+         * Corner radius for the popup body.
+         */
+        internal const val DEFAULT_CORNER_RADIUS = 12
+    }
+}
--- a/app/src/main/java/org/mozilla/fenix/compose/cfr/CFRPopupContent.kt
+++ b/app/src/main/java/org/mozilla/fenix/compose/cfr/CFRPopupContent.kt
@@ -1,0 +1,171 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.compose.cfr
+
+import android.content.res.Configuration
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.size
+import androidx.compose.foundation.layout.width
+import androidx.compose.foundation.layout.wrapContentHeight
+import androidx.compose.material.Icon
+import androidx.compose.material.IconButton
+import androidx.compose.material.Surface
+import androidx.compose.material.Text
+import androidx.compose.material.icons.Icons.Filled
+import androidx.compose.material.icons.filled.Close
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.geometry.Offset
+import androidx.compose.ui.graphics.Brush
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.dp
+import org.mozilla.fenix.R
+import org.mozilla.fenix.theme.FirefoxTheme
+
+/**
+ * Complete content of the popup.
+ * [CFRPopupShape] with a gradient background containing [text] and a dismiss ("X") button.
+ *
+ * @param text String message in the popup.
+ * @param indicatorDirection The direction the indicator arrow is pointing to.
+ * @param indicatorArrowStartOffset Maximum distance between the popup start and the indicator arrow.
+ * If there isn't enough space this could automatically be overridden up to 0.
+ * @param onDismiss  Callback for when the popup is dismissed indicating also if the dismissal
+ * was explicit - by tapping the "X" button or not.
+ * @param action Optional other composable to show just below the popup text.
+ */
+@Composable
+@Suppress("LongParameterList", "LongMethod")
+fun CFRPopupContent(
+    text: String,
+    indicatorDirection: CFRPopup.IndicatorDirection,
+    indicatorArrowStartOffset: Dp,
+    onDismiss: (Boolean) -> Unit,
+    popupWidth: Dp = CFRPopup.DEFAULT_WIDTH.dp,
+    action: @Composable (() -> Unit) = {}
+) {
+    val popupShape = CFRPopupShape(
+        indicatorDirection,
+        indicatorArrowStartOffset,
+        CFRPopup.DEFAULT_INDICATOR_HEIGHT.dp,
+        CFRPopup.DEFAULT_CORNER_RADIUS.dp,
+    )
+
+    Box(modifier = Modifier.width(popupWidth + CFRPopup.DEFAULT_HORIZONTAL_PADDING.dp * 2)) {
+        Surface(
+            color = Color.Transparent,
+            // Need to override the default RectangleShape to avoid casting shadows for that shape.
+            shape = popupShape,
+            modifier = Modifier
+                .align(Alignment.Center)
+                .background(
+                    shape = popupShape,
+                    brush = Brush.linearGradient(
+                        colors = listOf(
+                            FirefoxTheme.colors.gradientEnd,
+                            FirefoxTheme.colors.gradientStart
+                        ),
+                        end = Offset(0f, Float.POSITIVE_INFINITY),
+                        start = Offset(Float.POSITIVE_INFINITY, 0f)
+                    )
+                )
+                .wrapContentHeight()
+                .width(popupWidth)
+        ) {
+
+            Column(
+                modifier = Modifier
+                    .padding(
+                        start = 16.dp,
+                        top = 16.dp + if (indicatorDirection == CFRPopup.IndicatorDirection.UP) {
+                            CFRPopup.DEFAULT_INDICATOR_HEIGHT.dp
+                        } else {
+                            0.dp
+                        },
+                        end = 16.dp,
+                        bottom = 16.dp +
+                            if (indicatorDirection == CFRPopup.IndicatorDirection.DOWN) {
+                                CFRPopup.DEFAULT_INDICATOR_HEIGHT.dp
+                            } else {
+                                0.dp
+                            }
+                    )
+            ) {
+                Text(
+                    text = text,
+                    modifier = Modifier.padding(
+                        end = 24.dp, // 8.dp extra padding to the "X" icon
+                    ),
+                    color = FirefoxTheme.colors.textOnColorPrimary,
+                    style = FirefoxTheme.typography.body2
+                )
+
+                action()
+            }
+        }
+
+        IconButton(
+            onClick = { onDismiss(true) },
+            modifier = Modifier
+                .align(Alignment.TopEnd)
+                .padding(
+                    top = if (indicatorDirection == CFRPopup.IndicatorDirection.UP) 14.dp else 0.dp,
+                    end = 6.dp
+                )
+                .size(48.dp)
+        ) {
+            Icon(
+                imageVector = Filled.Close,
+                contentDescription = stringResource(R.string.cfr_dismiss_button_default_content_description),
+                modifier = Modifier
+                    // Following alignment and padding are necessary to visually align the middle
+                    // of the "X" button with the top of the text.
+                    .align(Alignment.TopCenter)
+                    .padding(top = 10.dp)
+                    .size(24.dp),
+                tint = FirefoxTheme.colors.iconOnColor
+            )
+        }
+    }
+}
+
+@Composable
+@Preview(locale = "en", name = "LTR")
+@Preview(locale = "ar", name = "RTL")
+@Preview(uiMode = Configuration.UI_MODE_NIGHT_YES, name = "Dark theme")
+@Preview(uiMode = Configuration.UI_MODE_NIGHT_NO, name = "Light theme")
+private fun CFRPopupAbovePreview() {
+    FirefoxTheme {
+        CFRPopupContent(
+            text = "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt",
+            indicatorDirection = CFRPopup.IndicatorDirection.DOWN,
+            indicatorArrowStartOffset = CFRPopup.DEFAULT_INDICATOR_START_OFFSET.dp,
+            onDismiss = { }
+        )
+    }
+}
+
+@Composable
+@Preview(locale = "en", name = "LTR")
+@Preview(locale = "ar", name = "RTL")
+@Preview(uiMode = Configuration.UI_MODE_NIGHT_YES, name = "Dark theme")
+@Preview(uiMode = Configuration.UI_MODE_NIGHT_NO, name = "Light theme")
+private fun CFRPopupBelowPreview() {
+    FirefoxTheme {
+        CFRPopupContent(
+            text = "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt",
+            indicatorDirection = CFRPopup.IndicatorDirection.UP,
+            indicatorArrowStartOffset = CFRPopup.DEFAULT_INDICATOR_START_OFFSET.dp,
+            onDismiss = { }
+        )
+    }
+}
--- a/app/src/main/java/org/mozilla/fenix/compose/cfr/CFRPopupShape.kt
+++ b/app/src/main/java/org/mozilla/fenix/compose/cfr/CFRPopupShape.kt
@@ -1,0 +1,252 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.compose.cfr
+
+import android.content.res.Configuration
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.width
+import androidx.compose.material.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.geometry.Offset
+import androidx.compose.ui.geometry.Size
+import androidx.compose.ui.graphics.Brush
+import androidx.compose.ui.graphics.Outline
+import androidx.compose.ui.graphics.Path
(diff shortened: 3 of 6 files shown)


app/src/main/java/org/mozilla/fenix/compose/cfr/CFRPopup.kt
```diff
@@ -0,0 +1,610 @@
...
+ * to ensure it respects a11y recomendations of 48dp size while also offer a bit more space to the text.
+ */
+private const val HORIZONTAL_PADDING = 10
+
+/**
+ * Value class allowing to easily reason about what an `Int` represents.
+ * This is compiled to the underlying `Int` type so incurs no performance penalty.
+ */
+@JvmInline
+internal value class Pixels(val value: Int)
```
MozillaNoah, line 78: this could potentially be useful in another component. What if we broke this out into a separate PR where we add `.px` extension functions, similar to the `Dp` value class?

app/src/main/java/org/mozilla/fenix/compose/cfr/CFRPopup.kt
```diff
@@ -0,0 +1,610 @@
...
+ */
+data class CFRPopupProperties(
+    val indicatorDirection: IndicatorDirection = IndicatorDirection.UP,
+    val dismissOnBackPress: Boolean = true,
+    val dismissOnClickOutside: Boolean = true,
+    val overlapAnchor: Boolean = false,
+    val indicatorArrowStartOffset: Dp = 30.dp,
+    val indicatorArrowHeight: Dp = 10.dp,
+    val cornerRadius: Dp = 10.dp,
+    val elevation: Dp = 10.dp,
```
MozillaNoah, line 124: where are the default specs for this obtained from?

app/src/main/java/org/mozilla/fenix/compose/cfr/CFRPopup.kt
```diff
@@ -0,0 +1,610 @@
...
+                            IndicatorDirection.DOWN -> {
+                                when (properties.overlapAnchor) {
+                                    true -> anchorLocation.last() - popupContentSize.height + anchor.height / 2
+                                    else -> anchorLocation.last() - popupContentSize.height
+                                }
+                            }
+                        }
+                    )
+                }
+            },
```
MozillaNoah, line 286: This takes up a considerable size of the instantiation for the `Popup`. Would it make sense to break this into out into a private function so the instantiation is a little cleaner?

app/src/main/java/org/mozilla/fenix/compose/cfr/CFRPopup.kt
```diff
@@ -0,0 +1,610 @@
...
+    }
+}
+
+/**
+ * [AbstractComposeView] that can be added or removed dynamically in the current window to display
+ * a [Composable] based popup anywhere on the screen.
+ */
+@SuppressLint("ViewConstructor") // Intended to be used only in code, don't need a View constructor
+@VisibleForTesting
+internal class CFRPopupFullScreenLayout(
```
MozillaNoah, line 179: could you add the kdocs for the params here?

app/src/main/java/org/mozilla/fenix/compose/cfr/CFRPopup.kt
```diff
@@ -0,0 +1,610 @@
...
+                // For when tapping outside the popup.
+                dismiss()
+                onDismiss(false)
+            }
+        ) {
+            CFRPopupContent(
+                text = text,
+                indicatorDirection = properties.indicatorDirection,
+                indicatorArrowStartOffset = with(LocalDensity.current) {
+                    (indicatorOffset.value).toDp()
```
MozillaNoah, line 302:
```suggestion
                    indicatorOffset.value.toDp()
```

app/src/main/java/org/mozilla/fenix/compose/cfr/CFRPopup.kt
```diff
@@ -0,0 +1,610 @@
...
+                },
+                indicatorArrowHeight = properties.indicatorArrowHeight,
+                cornerRadius = properties.cornerRadius,
+                elevation = properties.elevation,
+                onDismiss = {
+                    // For when tapping the "X" button.
+                    dismiss()
+                    onDismiss(true)
+                },
+                action = action
```
MozillaNoah, line 312:
```suggestion
                action = action,
```

app/src/main/java/org/mozilla/fenix/compose/cfr/CFRPopup.kt
MozillaNoah, line 179: This file is pretty large. Would it make sense to have `CFRPopupFullScreenLayout` and `CFRPopupContent` in their own files?

app/src/main/java/org/mozilla/fenix/compose/cfr/CFRPopup.kt
```diff
@@ -0,0 +1,610 @@
...
+     * @param arrowIndicatorWidth x-distance the arrow indicator occupies.
+     */
+    @Composable
+    private fun computePopupHorizontalBounds(
+        anchorMiddleXCoord: Pixels,
+        arrowIndicatorWidth: Pixels
+    ): PopupHorizontalBounds {
+        val arrowIndicatorHalfWidth = arrowIndicatorWidth.value / 2
+
+        if (LocalConfiguration.current.layoutDirection == View.LAYOUT_DIRECTION_LTR) {
```
MozillaNoah, line 332: nit: the `return` could be inlined with the `if`

app/src/main/java/org/mozilla/fenix/compose/cfr/CFRPopup.kt
```diff
@@ -0,0 +1,610 @@
...
+    private fun getLeftInsets() = ViewCompat.getRootWindowInsets(anchor)
+        ?.getInsets(WindowInsetsCompat.Type.systemBars())?.left
+        ?: 0.coerceAtLeast(0)
+
+    @Px
+    private fun Dp.toPx(): Int {
+        return this.value
+            .dpToPx(anchor.resources.displayMetrics)
+            .roundToInt()
+    }
```
MozillaNoah, line 442: why use this over `fun Dp.roundToPx()` in `Density`?

app/src/main/java/org/mozilla/fenix/compose/cfr/CFRPopup.kt
```diff
@@ -0,0 +1,610 @@
...
+    override fun onViewDetachedFromWindow(v: View?) {
+        onDismiss()
+    }
+}
+
+/**
+ * Complete content of the popup.
+ * [CFRPopupShape] with a gradient background containing [text] and a dismiss ("X") button.
+ *
+ * @param text String offering more context about a particular feature to the user.
```
MozillaNoah, line 463:
```suggestion
 * @param text String message in the popup.
```

app/src/main/java/org/mozilla/fenix/compose/cfr/CFRPopup.kt
```diff
@@ -0,0 +1,610 @@
...
+@Suppress("LongParameterList", "LongMethod")
+private fun CFRPopupContent(
+    text: String,
+    indicatorDirection: IndicatorDirection,
+    indicatorArrowStartOffset: Dp,
+    indicatorArrowHeight: Dp,
+    cornerRadius: Dp,
+    elevation: Dp = 0.dp,
+    onDismiss: () -> Unit,
+    action: @Composable (() -> Unit)? = null
```
MozillaNoah, line 486: where do we foresee this being used?

MozillaNoah, line 486:
```suggestion
    action: @Composable (() -> Unit) = {}
```

app/src/main/java/org/mozilla/fenix/compose/cfr/CFRPopup.kt
```diff
@@ -0,0 +1,610 @@
...
+                    Text(
+                        text = text,
+                        modifier = Modifier.padding(
+                            end = 24.dp, // 8.dp extra padding to the "X" icon
+                        ),
+                        color = FirefoxTheme.colors.textOnColorPrimary,
+                        style = FirefoxTheme.typography.body2
+                    )
+
+                    action?.invoke()
```
MozillaNoah, line 546:
```suggestion
                    action()

File: app/src/main/java/org/mozilla/fenix/compose/cfr/CFRPopup.kt [hunk 12]
```diff
@@ -0,0 +1,610 @@
...
+    text: String,
+    indicatorDirection: IndicatorDirection,
+    indicatorArrowStartOffset: Dp,
+    indicatorArrowHeight: Dp,
+    cornerRadius: Dp,
+    elevation: Dp = 0.dp,
+    onDismiss: () -> Unit,
+    action: @Composable (() -> Unit)? = null
+) {
+    FirefoxTheme {
```
MozillaNoah, line 488: this should be moved to `Content()` in `CFRPopupFullScreenLayout`

app/src/main/java/org/mozilla/fenix/compose/cfr/CFRPopup.kt
```diff
@@ -0,0 +1,610 @@
...
+        Box(modifier = Modifier.width(POPUP_WIDTH_DP.dp + HORIZONTAL_PADDING.dp * 2)) {
+            Surface(
+                color = Color.Transparent,
+                elevation = elevation,
+                // Need to override the default RectangleShape to avoid casting shadows for that shape.
+                shape = CFRPopupShape(
+                    indicatorDirection, indicatorArrowStartOffset, indicatorArrowHeight, cornerRadius
+                ),
```
MozillaNoah, line 496: if this is the same shape used in the modifier, it could be initialized once in a local `val` (and with the named params)

app/src/main/java/org/mozilla/fenix/compose/cfr/CFRPopup.kt
```diff
@@ -0,0 +1,610 @@
...
+    indicatorArrowStartOffset: Dp,
+    indicatorArrowHeight: Dp,
+    cornerRadius: Dp,
+    elevation: Dp = 0.dp,
+    onDismiss: () -> Unit,
+    action: @Composable (() -> Unit)? = null
+) {
+    FirefoxTheme {
+        Box(modifier = Modifier.width(POPUP_WIDTH_DP.dp + HORIZONTAL_PADDING.dp * 2)) {
+            Surface(
```
MozillaNoah, line 490: A `Surface` is a wrapper for a `Box`, so the two containers could be merged here by just having the `.width` Modifier before `.align`

app/src/main/java/org/mozilla/fenix/compose/cfr/CFRPopupShape.kt
```diff
@@ -0,0 +1,252 @@
...
+                text = "This is just a test",
+                color = FirefoxTheme.colors.textOnColorPrimary
+            )
+        }
+    }
+}
+
+@Composable
+@Preview(locale = "en", name = "LTR")
+@Preview(locale = "ar", name = "RTL")
```
MozillaNoah, line 224: nit: do we need additional previews for the Shape for RTL and LTR?

app/src/main/java/org/mozilla/fenix/compose/cfr/CFRPopupShape.kt
```diff
@@ -0,0 +1,252 @@
...
+import androidx.compose.ui.graphics.Shape
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.Density
+import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.LayoutDirection
+import androidx.compose.ui.unit.dp
+import org.mozilla.fenix.theme.FirefoxTheme
+import kotlin.math.roundToInt
+
+private const val INDICATOR_BASE_TO_HEIGHT_RATIO = 1.5f
```
MozillaNoah, line 30: what's a TLDR for this const?

MozillaNoah (commented):
Looking good so far

MozillaNoah:
We may also want a UX green light for this

app/src/main/res/values/strings.xml
```diff
@@ -73,6 +73,9 @@
     <!-- Text for the negative action button -->
     <string name="open_in_app_cfr_negative_button_text">Dismiss</string>
 
+    <!-- Content description for close button used of "contextual feature recommendation" (CFR) popups -->
```
gabrielluong, line 76:
```suggestion
    <!-- Content description for close button used in "contextual feature recommendation" (CFR) popups -->
```

app/src/main/java/org/mozilla/fenix/compose/cfr/CFRPopup.kt
Mugurell: I like the idea.
Fully supporting `Px` would mean adding many of the methods from the [`Dp.kt` file](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/ui/ui-unit/src/commonMain/kotlin/androidx/compose/ui/unit/Dp.kt) maybe to [AC around DisplayMetrics](https://github.com/mozilla-mobile/android-components/blob/71e124a5dbafffb800168ea615fdbc54d7398f50/components/support/ktx/src/main/java/mozilla/components/support/ktx/android/util/DisplayMetrics.kt) ?
Given the bigger scope I'd leave this as a later refactoring.

app/src/main/java/org/mozilla/fenix/compose/cfr/CFRPopup.kt
Mugurell: Thank you for raising this.
I just eyeballed the values to get everything working but did not came back to polish.
Checked now in previous CFRs  ([onboarding_jump_back_in_cfr](https://github.com/mozilla-mobile/fenix/blob/263e4d819be3b67c037917237f55050f3c09b8c9/app/src/main/res/layout/onboarding_jump_back_in_cfr.xml) and [onboarding_inactive_tabs_cfr](https://github.com/mozilla-mobile/fenix/blob/263e4d819be3b67c037917237f55050f3c09b8c9/app/src/main/res/layout/onboarding_inactive_tabs_cfr.xml)) and saw that they use [`15` as the "indicator offset"](https://github.com/mozilla-mobile/fenix/blob/263e4d819be3b67c037917237f55050f3c09b8c9/app/src/main/res/layout/onboarding_jump_back_in_cfr.xml#L55) and [`12` as the cornerRadius](https://github.com/mozilla-mobile/fenix/blob/51715c27b9f84a42b65be14ba3a0bf1cb9864e39/app/src/main/res/drawable/onboarding_popup_shape.xml#L6). Will update accordingly.
The other two properties - `indicatorArrowHeight` and `elevation` remain to be validated by UX maybe when I'll have the TCP CFR PR ready which would use this new CFR.

app/src/main/java/org/mozilla/fenix/compose/cfr/CFRPopup.kt
Mugurell: Also thought about this but kept it here to better encapsulate it as it should not be used by other classes.
But the same can be said about `CFRPopupShape`.
I'll extract it to a new file, thanks!

app/src/main/java/org/mozilla/fenix/compose/cfr/CFRPopup.kt
Mugurell: This allows to just write `dpValue.toPx()` to get what I want
vs having to write
`LocalDensity.current.run { dpValue.roundToPx() }`

Maybe another something to extract for later usages.

app/src/main/java/org/mozilla/fenix/compose/cfr/CFRPopup.kt
Mugurell: Guess that if we have a constants width of the CFR popups the `indicatorArrowHeight` and `cornerRadius` should also be constants. Will update as such. Will update as such.

`onDismiss` will be used for side effects like telemetry
`action` can be another composable for a button / link.

app/src/main/java/org/mozilla/fenix/compose/cfr/CFRPopup.kt
Mugurell: [`align` is scoped to Box](https://cs.android.com/androidx/platform/tools/dokka-devsite-plugin/+/master:testData/compose/source/androidx/compose/foundation/layout/Box.kt;l=219?q=interface%20BoxScope) and I need that to position the `X` button.
Then I use the Surface because it knows to easily handle a shape plus elevation.

app/src/main/java/org/mozilla/fenix/compose/cfr/CFRPopupShape.kt
Mugurell: I do think so. Would sort of us these previews as screenshot tests.
Given that the shapes are drawn programatically and there is different code to draw LRT and RTL these previews do help for a quick validation.

app/src/main/java/org/mozilla/fenix/compose/cfr/CFRPopupShape.kt
Mugurell: A value of `1` would mean the indicator is a equilateral triangle. 
A different value would give an isosceles triangle.
This would be useful in supporting different heights of the indicator with a big enough base. But will probably standardize the look based on UX feedback and remove this.

Mugurell:
Thank you for the (quick) help!
The plan is to have this is a landable form, use it for the TCP CFR and provide an apk for UX to help with the final design validation before merging.

app/src/main/java/org/mozilla/fenix/compose/cfr/CFRPopup.kt
MozillaNoah: Oh dang ok, fair enough. Weird that they didn't trickle down the functionality to `Surface`.

app/src/main/java/org/mozilla/fenix/compose/cfr/CFRPopupContent.kt
```diff
@@ -0,0 +1,176 @@
...
+            modifier = Modifier
+                .align(Alignment.TopEnd)
+                .padding(
+                    top = if (indicatorDirection == IndicatorDirection.UP) 8.dp else 0.dp,
+                    end = 6.dp
+                )
+                .size(48.dp)
+        ) {
+            Icon(
+                Filled.Close,
```
MozillaNoah, line 132: is this missing a named parameter?

MozillaNoah (commented):
There's a weird box being drawn in `CFRPopupContent` that's only showing up in the previews with a downward indicator triangle.

image

Mugurell:
> There's a weird box being drawn in `CFRPopupContent` that's only showing up in the previews with a downward indicator triangle.

Interesting, thanks. Removed setting the elevation in the `Surface` from `CFRPopupContent`. Will default to the elevation the compose Popup uses.

Mugurell:
Refactored to add the static properties declared in `CFRPoup` to a companion object to avoid polluting the general namespace and ensure a bit more context for where those values are used/useful.
Added on Jira - https://mozilla-hub.atlassian.net/browse/FNXV2-21106 more details for Kate to confirm or not.

app/src/main/java/org/mozilla/fenix/compose/cfr/CFRPopupContent.kt
```diff
@@ -0,0 +1,171 @@
...
+                .align(Alignment.TopEnd)
+                .padding(
+                    top = if (indicatorDirection == CFRPopup.IndicatorDirection.UP) 14.dp else 0.dp,
+                    end = 6.dp
+                )
+                .size(48.dp)
+        ) {
+            Icon(
+                imageVector = Filled.Close,
+                contentDescription = stringResource(string.cfr_dismiss_button_default_content_description),
```
MozillaNoah, line 128: Was the `R.string` imported?
```suggestion
                contentDescription = stringResource(R.string.cfr_dismiss_button_default_content_description),
```

MozillaNoah (approved):
This is good to go otherwise.

app/src/main/java/org/mozilla/fenix/compose/cfr/CFRPopupContent.kt
Mugurell: Yes, thank you!
Move some code around and imports got mangled, also added some `R.`s, will check again for all.

app/src/test/java/org/mozilla/fenix/compose/cfr/CFRPopupFullScreenLayoutTest.kt
```diff
@@ -0,0 +1,100 @@
...
+import mozilla.components.support.test.robolectric.testContext
+import org.junit.Assert.assertEquals
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.mozilla.fenix.helpers.FenixRobolectricTestRunner
+
+@RunWith(FenixRobolectricTestRunner::class)
+class CFRPopupFullScreenLayoutTest {
+    @Test
+    fun `WHEN the popup is constructed THEN setup lifecycle owners`() {
```
mavduevskiy, line 29: I have been playing around with the tests, and have noticed that both `ViewTreeSavedStateRegistryOwner.get()` and `ViewTreeLifecycleOwner.get()` always return `null`.

Since they are static, do we have any options but to check if the have been called with proper arguments?
Something like this.

```
    @Test
    fun `WHEN the popup is constructed THEN setup lifecycle owners`() {
        mockkStatic(ViewTreeLifecycleOwner::class)
        mockkStatic(ViewTreeSavedStateRegistryOwner::class)

        val container = View(testContext)
        val popupView = CFRPopupFullScreenLayout("", container, mockk(), mockk()) {}

        val viewTreeLifecycleCaptor = slot<View>()
        val viewTreeSavedStateCaptor = slot<View>()

        verify { ViewTreeLifecycleOwner.set(capture(viewTreeLifecycleCaptor), any()) }
        verify { ViewTreeSavedStateRegistryOwner.set(capture(viewTreeSavedStateCaptor), any()) }

        assertEquals(viewTreeLifecycleCaptor.captured, popupView)
        assertEquals(viewTreeSavedStateCaptor.captured, popupView)
    }
```

app/src/test/java/org/mozilla/fenix/compose/cfr/CFRPopupFullScreenLayoutTest.kt
```diff
@@ -0,0 +1,100 @@
...
+        val context = spyk(testContext)
+        val anchor = View(context)
+        val windowManager = spyk(context.getSystemService(Context.WINDOW_SERVICE) as WindowManager)
+        every { context.getSystemService(Context.WINDOW_SERVICE) } returns windowManager
+        val popupView = CFRPopupFullScreenLayout("", anchor, mockk(), mockk()) {}
+        popupView.show()
+
+        popupView.dismiss()
+
+        assertEquals(null, ViewTreeLifecycleOwner.get(popupView))
```
mavduevskiy, line 55: This check doesn't seem to fulfill its purpose since it is null before `.dismiss()` and `.show()` as well

mavduevskiy (commented):
Looks like a lot of (awesome) work! 
There hasn't been much discussion on tests, so I played around it a bit. I hope I am being helpful.

app/src/test/java/org/mozilla/fenix/compose/cfr/CFRPopupFullScreenLayoutTest.kt
mavduevskiy: If we want to check if `ViewTreeLifecycleOwner.set(this, null)` was called `onDismiss()`, wouldn't mocking static methods and checking if it was called with `null` lifecycle do the trick?

app/src/test/java/org/mozilla/fenix/compose/cfr/CFRPopupFullScreenLayoutTest.kt
Mugurell: Thank you!
Agree that checking for `null` equality is not really useful.
Your proposed approach would work in checking that whatever value the above "..owner"s have are corrently set, even if they are null.
For a somewhat simpler approach and actually checking for a "real" instance though we can set "..owner"s for the anchor and then test that they are properly propagated to our popupView (after which they will be propagated to `androidx.compose.ui.window.PopupLayout`).

app/src/test/java/org/mozilla/fenix/compose/cfr/CFRPopupFullScreenLayoutTest.kt
Mugurell: Thank you!
Same as above we can easily test even with a real initial owner that is actually set to null after the call to `dismiss`.

app/src/main/java/org/mozilla/fenix/compose/cfr/CFRPopup.kt
```diff
@@ -0,0 +1,123 @@
...
+         * a bit outside of the popup to ensure it respects a11y recommendations of 48dp size while
+         * also offer a bit more space to the text.
+         */
+        internal const val DEFAULT_HORIZONTAL_PADDING = 10
+
+        /**
+         * How tall the indicator arrow should be.
+         * This will also affect how wide the base of the indicator arrow will be.
+         */
+        internal const val DEFAULT_INDICATOR_HEIGHT = 15
```
gabrielluong, line 111: Should this be 16dp? https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/res/values/dimens.xml#45

app/src/main/java/org/mozilla/fenix/compose/cfr/CFRPopupContent.kt
```diff
@@ -0,0 +1,171 @@
...
+            )
+        }
+    }
+}
+
+@Composable
+@Preview(locale = "en", name = "LTR")
+@Preview(locale = "ar", name = "RTL")
+@Preview(uiMode = Configuration.UI_MODE_NIGHT_YES, name = "Night mode")
+@Preview(uiMode = Configuration.UI_MODE_NIGHT_NO, name = "Light mode")
```
gabrielluong, line 145:
```suggestion
@Preview(uiMode = Configuration.UI_MODE_NIGHT_YES, name = "Dark theme")
@Preview(uiMode = Configuration.UI_MODE_NIGHT_NO, name = "Light theme")
```
I would prefer if we could go with the current theme terminology within Fenix

gabrielluong (commented):
I find myself looking into this since I pointed out to @sarah541 that it might be worth using this CFR pop composable to show the CFR for the synced tab card in the Jump back in section.

app/src/test/java/org/mozilla/fenix/compose/cfr/CFRPopupFullScreenLayoutTest.kt
```diff
@@ -0,0 +1,112 @@
...
+            ViewTreeSavedStateRegistryOwner.set(this, mockk())
+        }
+        val windowManager = spyk(context.getSystemService(Context.WINDOW_SERVICE) as WindowManager)
+        every { context.getSystemService(Context.WINDOW_SERVICE) } returns windowManager
+        val popupView = CFRPopupFullScreenLayout("", anchor, mockk(), mockk()) {}
+        popupView.show()
+
+        popupView.dismiss()
+
+        assertNull(popupView.findViewTreeLifecycleOwner())
```
mavduevskiy, line 67: Does it make sense to check if it was not `null` before `dismiss` is called? Technically speaking, it could have been `null` all along :)

Mugurell: My reasoning was that we have a test already just above for checking that upon instantiating the popupView the lifecycle owners will not be null.
I can add a check here also to be sure.

app/src/main/java/org/mozilla/fenix/compose/cfr/CFRPopup.kt
Mugurell: The arrow drawable that `cfr_triangle_height == 16dp` applies to - [`ic_cfr_triangle`](https://github.com/mozilla-mobile/fenix/blob/main/app/src/main/res/drawable/ic_cfr_triangle.xml) actually has a lot of padding so the visual height will be maybe half of that value, while `cfr_triangle_width == 16dp` applied to the same drawable would probably result visually in a width of ~15dp.

image

&nbsp;

To me the arrow seemed a bit small in height so I tried a higher default.
Kate just confirmed the designs shown in https://mozilla-hub.atlassian.net/browse/FNXV2-21106 look good so I would say we can go with the value of `15dp` from here and change afterwards if needed.

Mugurell:
Landing this for Sarah to also be able to use it.

PR #26195, opened 2022-07-26
Fixes #26124: recent synced tab will use preview image URLs from history

### LINKED ISSUE #26124: Recent synced tab should retrieve hero image URL from synced history
Currently, a centered favicon is being displayed in the recent synced tab on the homescreen. A better image can be retrieved by looking up the `previewImageUrl` for the Places history entry (which is Sync'd), and using that instead of the `iconUrl` attached to tabs.

See [the bookmark implementation](https://searchfox.org/mozilla-mobile/rev/198e453c187b93d0e2e52776f99fb1c0061ca1aa/fenix/app/src/main/java/org/mozilla/fenix/components/bookmarks/BookmarksUseCase.kt#83) for how this can be done.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21076)

MatthewTighe:
image

### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation
Fixes #26124

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/home/recentsyncedtabs/RecentSyncedTabFeature.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/recentsyncedtabs/RecentSyncedTabFeature.kt
@@ -8,6 +8,7 @@ import android.content.Context
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.flow.launchIn
 import kotlinx.coroutines.flow.onEach
+import mozilla.components.concept.storage.HistoryStorage
 import mozilla.components.concept.sync.DeviceType
 import mozilla.components.feature.syncedtabs.storage.SyncedTabsStorage
 import mozilla.components.lib.state.ext.flow
@@ -19,11 +20,13 @@ import mozilla.components.service.fxa.store.SyncStatus
 import mozilla.components.service.fxa.store.SyncStore
 import mozilla.components.service.fxa.sync.SyncReason
 import mozilla.components.support.base.feature.LifecycleAwareFeature
+import mozilla.components.support.ktx.kotlin.tryGetHostFromUrl
 import mozilla.components.support.ktx.kotlinx.coroutines.flow.ifChanged
 import org.mozilla.fenix.components.AppStore
 import org.mozilla.fenix.components.appstate.AppAction
 import mozilla.telemetry.glean.GleanTimerId
 import org.mozilla.fenix.GleanMetrics.RecentSyncedTabs
+import java.util.concurrent.TimeUnit
 
 /**
  * Delegate to handle layout updates and dispatch actions related to the recent synced tab.
@@ -32,14 +35,17 @@ import org.mozilla.fenix.GleanMetrics.RecentSyncedTabs
  * @property syncStore Store to observe for changes to Sync and account status.
  * @property storage Storage layer for synced tabs.
  * @property accountManager Account manager to initiate Syncs and refresh devices.
+ * @property historyStorage Storage for searching history for preview image URLs matching synced tab.
  * @property coroutineScope The scope to collect Sync state Flow updates in.
  */
+@Suppress("LongParameterList")
 class RecentSyncedTabFeature(
     private val context: Context,
     private val appStore: AppStore,
     private val syncStore: SyncStore,
     private val storage: SyncedTabsStorage,
     private val accountManager: FxaAccountManager,
+    private val historyStorage: HistoryStorage,
     private val coroutineScope: CoroutineScope,
 ) : LifecycleAwareFeature {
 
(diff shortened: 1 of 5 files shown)


gabrielluong:
@MatthewTighe This is more of a follow up, but it might be worthwhile to check with UX (I would ask at least 2 UX Kate and Nicole and maybe even Product since this seems like a medium-size decision, maybe use the eng-ux slack channel) if we should also just change all the "Jump back in" images to use the hero/preview image instead of the thumbnail.

MozillaNoah (approved):
🚢 🚢 🚢 

I agree with @gabrielluong though there may need to be a follow-up for some UI unification.

app/src/main/java/org/mozilla/fenix/home/recentsyncedtabs/RecentSyncedTabFeature.kt
```diff
@@ -129,6 +148,12 @@ class RecentSyncedTabFeature(
     }
 }
 
+/**
+ * The number of days to search history for a preview image URL to display for a synced
+ * tab.
+ */
+private const val DAYS_HISTORY_FOR_PREVIEW_IMAGE = 3L
```
gabrielluong, line 155: Should this just be inside the `companion object` of `RecentSyncedTabFeature`? No strong opinions on this, but it might be something worth doing to make it testable.

app/src/main/java/org/mozilla/fenix/home/recentsyncedtabs/RecentSyncedTabFeature.kt
```diff
@@ -95,12 +100,26 @@ class RecentSyncedTabFeature(
...
+                val currentTime = System.currentTimeMillis()
+                val maxAgeInMs = TimeUnit.DAYS.toMillis(DAYS_HISTORY_FOR_PREVIEW_IMAGE)
+                val history = historyStorage.getDetailedVisits(
+                    start = currentTime - maxAgeInMs,
+                    end = currentTime
+                )
+
+                // Searching history entries for any that share a top level domain and have a
+                // preview image URL available casts a wider net for finding a suitable image.
+                val imageUrl = history.find { entry ->
```
gabrielluong, line 113: Could we just name this `previewImageUrl`?  Generally, I prefer if the variable naming follows the named parameter that we are passing to if there's no obvious naming conflict.

app/src/main/java/org/mozilla/fenix/home/recentsyncedtabs/RecentSyncedTabFeature.kt
gabrielluong, line 155: I am not sure if 3 days might be too short. We have used 10 days in https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/java/org/mozilla/fenix/components/bookmarks/BookmarksUseCase.kt#107.

app/src/main/java/org/mozilla/fenix/home/recentsyncedtabs/RecentSyncedTabFeature.kt
```diff
@@ -155,12 +180,12 @@ sealed class RecentSyncedTabState {
  * @param deviceDisplayName The device the tab was viewed on.
  * @param title The title of the tab.
  * @param url The url of the tab.
- * @param iconUrl The url used to retrieve the icon of the tab.
+ * @param previewImageUrl The url used to retrieve the icon of the tab.
```
gabrielluong, line 183:
```suggestion
 * @param previewImageUrl The url used to retrieve the preview image of the tab.
```

app/src/main/java/org/mozilla/fenix/home/recentsyncedtabs/view/RecentSyncedTab.kt
```diff
@@ -68,13 +70,22 @@ fun RecentSyncedTab(
...
-                        url = tab.url,
-                        key = tab.url.hashCode().toString(),
-                        modifier = Modifier
-                            .size(108.dp, 80.dp)
-                            .clip(RoundedCornerShape(8.dp))
-                    )
+                    val imageModifier = Modifier
+                        .size(108.dp, 80.dp)
+                        .clip(RoundedCornerShape(8.dp))
+                    if (tab.previewImageUrl != null) {
```
gabrielluong, line 76: nit: can you add a new line above this.

gabrielluong (approved):

app/src/main/java/org/mozilla/fenix/home/recentsyncedtabs/RecentSyncedTabFeature.kt
MatthewTighe: I hadn't thought of it from a testability standpoint. A companion object would probably make the most sense in that case. I generally kinda dislike the pattern of instantiating a whole object just for a const val, [this SO post](https://stackoverflow.com/questions/44038721/constants-in-kotlin-whats-a-recommended-way-to-create-them) covers why a bit. A top-level internal val could work if we had modules in Fenix, but I think namespacing it with a companion makes more sense since we don't.

MatthewTighe: I used 3 here to try and keep it a little more performant, we're already struggling with the performance of this feature a little bit. My reasoning being that typically a "recent synced tab" has probably been visited on its original device within the last 3 days

MatthewTighe:
> This is more of a follow up, but it might be worthwhile to check with UX (I would ask at least 2 UX Kate and Nicole and maybe even Product since this seems like a medium-size decision, maybe use the eng-ux slack channel) if we should also just change all the "Jump back in" images to use the hero/preview image instead of the thumbnail. This is more for general consistency since we have both thumbnail and hero image now for "Jump back in" and we already use hero image for Recent Bookmarks and Pocket.

I've filed https://github.com/mozilla-mobile/android-components/issues/11971 and https://github.com/mozilla-mobile/fenix/issues/26230 as followups to add a fallback to a site screenshot, based on feedback from UX.

app/src/main/java/org/mozilla/fenix/home/recentsyncedtabs/RecentSyncedTabFeature.kt
gabrielluong: Hmm.. interesting. I don't know if we really considered the pattern about how we define constants. My initial reaction is simply based on what I have seen in the project, and looking at what we did in the `BookmarksUseCase`. I won't be the best person to weigh on how we should ideally proceed with this, but might always be something to bring up in a discussion if you think it's worthwhile to establish some sort of guidelines.

gabrielluong: Thanks, sounds good to me.

PR #26226, opened 2022-07-28
For #26224 - New TCP CFR popup

### LINKED ISSUE #26224: TCP CFR
https://mozilla-hub.atlassian.net/browse/FNXV2-20991

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21138)

Mugurell:
### Pull Request checklist

- [x] **Screenshots in Jira**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not

### QA

- [x] **QA Needed**

### GitHub Automation
Fixes #26224

### DIFF
--- a/app/metrics.yaml
+++ b/app/metrics.yaml
@@ -2509,7 +2509,6 @@ bookmarks_management:
       - android-probes@mozilla.com
       - erichards@mozilla.com
     expires: never
-
     metadata:
       tags:
         - Bookmarks
@@ -4574,7 +4573,6 @@ tracking_protection:
     notification_emails:
       - android-probes@mozilla.com
     expires: 111
-
     metadata:
       tags:
         - TrackingProtection
@@ -6051,7 +6049,6 @@ addons:
     notification_emails:
       - android-probes@mozilla.com
     expires: 116
-
     metadata:
       tags:
         - WebExtensions
@@ -7012,7 +7009,6 @@ awesomebar:
       - android-probes@mozilla.com
       - erichards@mozilla.com
     expires: never
-
     metadata:
       tags:
         - Search
--- a/app/metrics.yaml
+++ b/app/metrics.yaml
@@ -4578,6 +4578,73 @@ tracking_protection:
     metadata:
       tags:
         - TrackingProtection
+  tcp_cfr_shown:
+    type: event
+    description: |
+      The Total Cookie Protection CFR was shown to the user.
+    bugs:
+      - https://github.com/mozilla-mobile/fenix/issues/26224
+    data_reviews:
+      - https://github.com/mozilla-mobile/fenix/pull/26226#issuecomment-1204007332
+    data_sensitivity:
+      - interaction
+    notification_emails:
+      - android-probes@mozilla.com
+    expires: 119
+    metadata:
+      tags:
+        - TrackingProtection
+  tcp_cfr_implicit_dismissal:
+    type: event
+    description: |
+      The Total Cookie Protection CFR was dismissed by the user by interacting
+      with the outside of the popup.
+    bugs:
+      - https://github.com/mozilla-mobile/fenix/issues/26224
+    data_reviews:
+      - https://github.com/mozilla-mobile/fenix/pull/26226#issuecomment-1204007332
+    data_sensitivity:
+      - interaction
+    notification_emails:
+      - android-probes@mozilla.com
+    expires: 119
+    metadata:
+      tags:
+        - TrackingProtection
+  tcp_cfr_explicit_dismissal:
+    type: event
+    description: |
+      The Total Cookie Protection CFR was dismissed by the user by clicking on
+      the "X" button to close the popup.
+    bugs:
+      - https://github.com/mozilla-mobile/fenix/issues/26224
+    data_reviews:
+      - https://github.com/mozilla-mobile/fenix/pull/26226#issuecomment-1204007332
+    data_sensitivity:
+      - interaction
+    notification_emails:
+      - android-probes@mozilla.com
+    expires: 119
+    metadata:
+      tags:
+        - TrackingProtection
+  tcp_sumo_link_clicked:
+    type: event
+    description: |
+      A user clicked the link in the Total Cookie Protection CFR to open a new
+      SUMO tab detailing the Total Cookie Protection feature.
+    bugs:
+      - https://github.com/mozilla-mobile/fenix/issues/26224
+    data_reviews:
+      - https://github.com/mozilla-mobile/fenix/pull/26226#issuecomment-1204007332
+    data_sensitivity:
+      - interaction
+    notification_emails:
+      - android-probes@mozilla.com
+    expires: 119
+    metadata:
+      tags:
+        - TrackingProtection
 tab:
   media_play:
     type: event
--- a/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
+++ b/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
@@ -574,6 +574,14 @@ class Settings(private val appContext: Context) : PreferencesHolder {
     val enabledTotalCookieProtectionSetting: Boolean
         get() = mr2022Sections[Mr2022Section.TCP_CFR] == true
 
+    /**
+     * Indicates if the total cookie protection CRF should be shown.
+     */
+    var shouldShowTotalCookieProtectionCFR by booleanPreference(
+        appContext.getPreferenceKey(R.string.pref_key_should_show_total_cookie_protection_popup),
+        default = FxNimbus.features.engineSettings.value().totalCookieProtectionEnabled
+    )
+
     val blockCookiesSelectionInCustomTrackingProtection by stringPreference(
         key = appContext.getPreferenceKey(R.string.pref_key_tracking_protection_custom_cookies_select),
         default = if (enabledTotalCookieProtectionSetting) {
--- a/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
+++ b/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
@@ -569,17 +569,17 @@ class Settings(private val appContext: Context) : PreferencesHolder {
     )
 
     val enabledTotalCookieProtection: Boolean
-        get() = FxNimbus.features.engineSettings.value().totalCookieProtectionEnabled
+        get() = mr2022Sections[Mr2022Section.TCP_FEATURE] == true
 
     val enabledTotalCookieProtectionSetting: Boolean
-        get() = mr2022Sections[Mr2022Section.TCP_CFR] == true
+        get() = mr2022Sections[Mr2022Section.TCP_FEATURE] == true
 
     /**
      * Indicates if the total cookie protection CRF should be shown.
      */
     var shouldShowTotalCookieProtectionCFR by booleanPreference(
         appContext.getPreferenceKey(R.string.pref_key_should_show_total_cookie_protection_popup),
-        default = FxNimbus.features.engineSettings.value().totalCookieProtectionEnabled
+        default = mr2022Sections[Mr2022Section.TCP_CFR] == true,
     )
 
     val blockCookiesSelectionInCustomTrackingProtection by stringPreference(
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -76,6 +76,12 @@
     <!-- Content description for close button used in "contextual feature recommendation" (CFR) popups -->
     <string name="cfr_dismiss_button_default_content_description">Dismiss</string>
 
+    <!-- Total cookie protection "contextual feature recommendation" (CFR) -->
+    <!-- Text for the message displayed in the contextual feature recommendation popup promoting the total cookie protection feature. -->
+    <string name="tcp_cfr_message">Our most powerful privacy feature yet isolates cross-site trackers.</string>
+    <!-- Text displayed that links to website containing documentation about the "Total cookie protection" feature. -->
+    <string name="tcp_cfr_learn_more">Learn about Total Cookie Protection</string>
+
     <!-- Text for the info dialog when camera permissions have been denied but user tries to access a camera feature. -->
     <string name="camera_permissions_needed_message">Camera access needed. Go to Android settings, tap permissions, and tap allow.</string>
     <!-- Text for the positive action button to go to Android Settings to grant permissions. -->
--- a/app/src/androidTest/java/org/mozilla/fenix/helpers/FeatureSettingsHelper.kt
+++ b/app/src/androidTest/java/org/mozilla/fenix/helpers/FeatureSettingsHelper.kt
@@ -18,6 +18,7 @@ class FeatureSettingsHelper {
     private var isRecentTabsFeatureEnabled: Boolean = settings.showRecentTabsFeature
     private var isRecentlyVisitedFeatureEnabled: Boolean = settings.historyMetadataUIFeature
     private var isUserKnowsAboutPwasTrue: Boolean = settings.userKnowsAboutPwas
+    private var isTCPCFREnabled: Boolean = settings.shouldShowTotalCookieProtectionCFR
 
     fun setPocketEnabled(enabled: Boolean) {
         settings.showPocketRecommendationsFeature = enabled
@@ -47,6 +48,13 @@ class FeatureSettingsHelper {
         settings.deleteSitePermissions = delete
     }
 
+    /**
+     * Enable or disable showing the TCP CFR when accessing a webpage for the first time.
+     */
+    fun setTCPCFREnabled(shouldShowCFR: Boolean) {
+        settings.shouldShowTotalCookieProtectionCFR = shouldShowCFR
+    }
+
     // Important:
     // Use this after each test if you have modified these feature settings
     // to make sure the app goes back to the default state
@@ -56,5 +64,6 @@ class FeatureSettingsHelper {
         settings.showRecentTabsFeature = isRecentTabsFeatureEnabled
         settings.historyMetadataUIFeature = isRecentlyVisitedFeatureEnabled
         settings.userKnowsAboutPwas = isUserKnowsAboutPwasTrue
+        settings.shouldShowTotalCookieProtectionCFR = isTCPCFREnabled
     }
 }
--- a/app/src/androidTest/java/org/mozilla/fenix/ui/BrowsingErrorPagesTest.kt
+++ b/app/src/androidTest/java/org/mozilla/fenix/ui/BrowsingErrorPagesTest.kt
@@ -39,6 +39,7 @@ class BrowsingErrorPagesTest {
     fun setUp() {
         // disabling the jump-back-in pop-up that interferes with the tests.
         featureSettingsHelper.setJumpBackCFREnabled(false)
+        featureSettingsHelper.setTCPCFREnabled(false)
     }
 
     @After
--- a/app/src/androidTest/java/org/mozilla/fenix/ui/ContextMenusTest.kt
+++ b/app/src/androidTest/java/org/mozilla/fenix/ui/ContextMenusTest.kt
@@ -15,6 +15,7 @@ import org.junit.Test
 import org.mozilla.fenix.customannotations.SmokeTest
 import org.mozilla.fenix.ext.settings
 import org.mozilla.fenix.helpers.AndroidAssetDispatcher
+import org.mozilla.fenix.helpers.FeatureSettingsHelper
 import org.mozilla.fenix.helpers.HomeActivityIntentTestRule
 import org.mozilla.fenix.helpers.RetryTestRule
 import org.mozilla.fenix.helpers.TestAssetHelper
@@ -47,6 +48,8 @@ class ContextMenusTest {
     @JvmField
     val retryTestRule = RetryTestRule(3)
 
+    private val featureSettingsHelper = FeatureSettingsHelper()
+
     @Before
     fun setUp() {
         activityIntentTestRule.activity.applicationContext.settings().shouldShowJumpBackInCFR = false
@@ -55,11 +58,14 @@ class ContextMenusTest {
             dispatcher = AndroidAssetDispatcher()
             start()
         }
+
+        featureSettingsHelper.setTCPCFREnabled(false)
     }
 
     @After
     fun tearDown() {
         mockWebServer.shutdown()
+        featureSettingsHelper.resetAllFeatureFlags()
     }
 
     @SmokeTest
--- a/app/src/androidTest/java/org/mozilla/fenix/ui/CustomTabsTest.kt
+++ b/app/src/androidTest/java/org/mozilla/fenix/ui/CustomTabsTest.kt
@@ -14,6 +14,7 @@ import org.junit.Test
 import org.mozilla.fenix.IntentReceiverActivity
 import org.mozilla.fenix.customannotations.SmokeTest
 import org.mozilla.fenix.helpers.AndroidAssetDispatcher
+import org.mozilla.fenix.helpers.FeatureSettingsHelper
 import org.mozilla.fenix.helpers.HomeActivityIntentTestRule
 import org.mozilla.fenix.helpers.TestAssetHelper
 import org.mozilla.fenix.helpers.TestHelper.createCustomTabIntent
@@ -43,6 +44,8 @@ class CustomTabsTest {
         IntentReceiverActivity::class.java, true, false
     )
 
+    private val featureSettingsHelper = FeatureSettingsHelper()
+
     @Before
     fun setUp() {
         mDevice = UiDevice.getInstance(InstrumentationRegistry.getInstrumentation())
@@ -50,11 +53,14 @@ class CustomTabsTest {
             dispatcher = AndroidAssetDispatcher()
             start()
         }
+
+        featureSettingsHelper.setTCPCFREnabled(false)
     }
 
     @After
     fun tearDown() {
         mockWebServer.shutdown()
+        featureSettingsHelper.resetAllFeatureFlags()
     }
 
     @SmokeTest
--- a/app/src/androidTest/java/org/mozilla/fenix/ui/DownloadFileTypesTest.kt
+++ b/app/src/androidTest/java/org/mozilla/fenix/ui/DownloadFileTypesTest.kt
@@ -52,6 +52,8 @@ class DownloadFileTypesTest(fileName: String) {
     fun setUp() {
         // disabling the jump-back-in pop-up that interferes with the tests.
         featureSettingsHelper.setJumpBackCFREnabled(false)
+        // disable the TCP CFR that appears when loading webpages and interferes with the tests.
+        featureSettingsHelper.setTCPCFREnabled(false)
         // disabling the PWA CFR on 3rd visit
         featureSettingsHelper.disablePwaCFR(true)
     }
--- a/app/src/androidTest/java/org/mozilla/fenix/ui/DownloadTest.kt
+++ b/app/src/androidTest/java/org/mozilla/fenix/ui/DownloadTest.kt
@@ -66,6 +66,8 @@ class DownloadTest {
         mDevice = UiDevice.getInstance(InstrumentationRegistry.getInstrumentation())
         // disabling the jump-back-in pop-up that interferes with the tests.
         featureSettingsHelper.setJumpBackCFREnabled(false)
+        // disable the TCP CFR that appears when loading webpages and interferes with the tests.
+        featureSettingsHelper.setTCPCFREnabled(false)
         // disabling the PWA CFR on 3rd visit
         featureSettingsHelper.disablePwaCFR(true)
         // clear all existing notifications
--- a/app/src/androidTest/java/org/mozilla/fenix/ui/HistoryTest.kt
+++ b/app/src/androidTest/java/org/mozilla/fenix/ui/HistoryTest.kt
@@ -21,6 +21,7 @@ import org.mozilla.fenix.R
 import org.mozilla.fenix.customannotations.SmokeTest
 import org.mozilla.fenix.ext.settings
 import org.mozilla.fenix.helpers.AndroidAssetDispatcher
+import org.mozilla.fenix.helpers.FeatureSettingsHelper
 import org.mozilla.fenix.helpers.HomeActivityTestRule
 import org.mozilla.fenix.helpers.RecyclerViewIdlingResource
 import org.mozilla.fenix.helpers.TestAssetHelper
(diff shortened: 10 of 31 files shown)


app/metrics.yaml
```diff
@@ -4556,6 +4556,73 @@ tracking_protection:
       - android-probes@mozilla.com
     expires: 111
```
gabrielluong, line 4558: just a drive by, don't think we want a new line here?

Mugurell: That line was already there.
I didn't add it but I can remove it though there are other similar declarations.
Can cleanup all in a separate commit. ?

rocketsroger:
> DATA REVIEW REQUEST
> 
>     1. What questions will you answer with this data?
> 
> 
> This data will help us understand how users interact with the Total Cookie Protection Contextual Feature Recommendation Popup.
> 
>     2. Why does Mozilla need to answer these questions? Are there benefits for users?
>        Do we need this information to address product or business requirements?
> 
> 
> To better understand the usefulness of new features.
> 
>     3. What alternative methods did you consider to answer these questions?
>        Why were they not sufficient?
> 
> 
> There are no other.
> 
>     4. Can current instrumentation answer these questions?
> 
> 
> No.
> 
>     5. List all proposed measurements and indicate the category of data collection for each
>        measurement, using the Firefox data collection categories found on the Mozilla wiki.
> 
> 
> Measurement Name 	Measurement Description 	Data Collection Category 	Tracking Bug
> `tracking_protection.tcp_cfr_shown` 	The Total Cookie Protection CFR was shown to the user. 	Category 2 - interaction 	#26224
> `tracking_protection.tcp_cfr_implicit_dismissal` 	The Total Cookie Protection CFR was dismissed by the user by interacting with the outside of the popup. 	Category 2 - interaction 	#26224
> `tracking_protection.tcp_cfr_explicit_dismissal` 	The Total Cookie Protection CFR was dismissed by the user by clicking on the "X" button to close the popup. 	Category 2 - interaction 	#26224
> `tracking_protection.tcp_sumo_link_clicked` 	A user clicked the link in the Total Cookie Protection CFR to open a new SUMO tab detailing the Total Cookie Protection feature. 	Category 2 - interaction 	#26224
> 
>     6. Please provide a link to the documentation for this data collection which
>        describes the ultimate data set in a public, complete, and accurate way.
> 
> 
> This collection is Glean so is documented [in the Glean Dictionary](https://dictionary.telemetry.mozilla.org).
> 
>     7. How long will this data be collected?
> 
> 
> Up until version '119', with the option to renew at that point.
> 
>     8. What populations will you measure?
> 
> 
> All channels, countries, and locales. No filters.
> 
>     9. If this data collection is default on, what is the opt-out mechanism for users?
> 
> 
> These collections are Glean. The opt-out can be found in the product's preferences.
> 
>     10. Please provide a general description of how you will analyze this data.
> 
> 
> Glean and Amplitude. 
> 
>     11. Where do you intend to share the results of your analysis?
> 
> 
> Only on Glean, Amplitude, and with mobile teams. 
> 
>     12. Is there a third-party tool (i.e. not Glean or Telemetry) that you
>         are proposing to use for this data collection?
> 
> 
> No.

# Data Review

Yes, through the metrics.yaml file and the [Glean Dictionary](https://dictionary.telemetry.mozilla.org/apps/fenix )

Yes, through the "Send Usage Data" preference in the application settings

N/A, collection set to end or be renewed by version 119

4) Using the **[category system of data types](https://wiki.mozilla.org/Data_Collection)** on the Mozilla wiki, what collection type of data do the requested measurements fall under?

Category 2, Interaction data

default-on

6) Does the instrumentation include the addition of **any *new* identifiers**?

No

7) Is the data collection covered by the existing Firefox privacy notice? 

Yes

8) Does the data collection use a third-party collection tool? 

No

## Result

data-review+

mavduevskiy:
Could we rebase it, so the first commit which we merged in another PR, wouldn't be present here as well?

app/metrics.yaml
Mugurell: Removed the newlines in a separate commit.

app/src/main/java/org/mozilla/fenix/components/toolbar/CFRPresenter.kt
```diff
@@ -0,0 +1,130 @@
...
+    /**
+     * Start observing [store] for updates which may trigger showing a CFR.
+     */
+    @Suppress("MagicNumber")
+    fun start() {
+        if (settings.shouldShowTotalCookieProtectionCFR) {
+            tcpCfrScope = store.flowScoped { flow ->
+                flow
+                    .mapNotNull { it.findCustomTabOrSelectedTab(sessionId)?.content?.progress }
+                    // The "transformWhile" below ensures that the 100% progress is only collected once.
```
mavduevskiy, line 63: Wouldn't `takeWhile` be more suitable? We want to receive items until it fits out condition (being == 100), and `takeWhile` does  just that. Something like:

```
                    .takeWhile { progress ->
                        val isLoading = progress != 100
                        if (!isLoading) {
                            showTcpCfr()
                        }
                        isLoading
                    }
                    .collect()
```

app/src/main/java/org/mozilla/fenix/components/toolbar/CFRPresenter.kt
```diff
@@ -0,0 +1,130 @@
...
+                },
+                style = FirefoxTheme.typography.body2.copy(
+                    textDecoration = TextDecoration.Underline
+                )
+            )
+        }.apply {
+            settings.shouldShowTotalCookieProtectionCFR = false
+            tcpCfr = this
+            show()
+            TrackingProtection.tcpCfrShown.record(NoExtras())
```
mavduevskiy, line 127: Nit:
May be we could move this function into the body of `fun showTcpCfr()`? `apply` is usually used for setting parameters or at list calls on `this`. I see why one could put an analytics call there, but it looks a bit out of place in there.

app/src/main/res/values/strings.xml
```diff
@@ -76,6 +76,12 @@
     <!-- Content description for close button used in "contextual feature recommendation" (CFR) popups -->
     <string name="cfr_dismiss_button_default_content_description">Dismiss</string>
 
+    <!-- Total cookie protection "contextual feature recommendation" (CFR) -->
+    <!-- Text for the main message. -->
+    <string name="tcp_cfr_message">Our most powerful privacy feature yet isolates cross-site trackers.</string>
+    <!-- Text displayed that links to website containing documentation about the "Total cookie protection" feature. -->
+    <string name="tcp_cfr_learn_more">Learn about Total Cookie Protection</string>
```
mavduevskiy, line 83: Cannot find a design for the feature, but just as a though, may be a link text should be with \<u> tags

app/src/test/java/org/mozilla/fenix/components/toolbar/CFRPresenterTest.kt
```diff
@@ -0,0 +1,324 @@
...
+            context = context,
+            store = mockk(),
+            settings = settings,
+            toolbar = toolbar,
+        )
+
+        presenter.showTcpCfr()
+
+        verify { settings.shouldShowTotalCookieProtectionCFR = false }
+        val cfr = presenter.tcpCfr!!
```
mavduevskiy, line 257: We have so few `!!`, may be there is a way to avoid using it? It looks like we are trying to avoid using it in the project in general!

app/src/test/java/org/mozilla/fenix/components/toolbar/CFRPresenterTest.kt
```diff
@@ -0,0 +1,324 @@
...
+class CFRPresenterTest {
+    @get:Rule
+    val coroutinesTestRule = MainCoroutineRule()
+    @get:Rule
+    val gleanTestRule = GleanTestRule(testContext)
+
+    @Test
+    fun `GIVEN the TCP CFR should be shown for a custom tab WHEN the custom tab is fully loaded THEN the TCP CFR is shown`() {
+        val customTab = createCustomTab(url = "")
+        val browserStore = BrowserStore(
```
mavduevskiy, line 52: Throughout the test we create `browserStore` and `presenter` multiple times. Given it takes a few lines of code, may be it's worth creating a function (with some parameters in case we need to change those like `DefaultBrowserToolbarMenuControllerTest.createController()`?

app/src/test/java/org/mozilla/fenix/components/toolbar/CFRPresenterTest.kt
```diff
@@ -0,0 +1,324 @@
...
+import org.mozilla.fenix.R
+import org.mozilla.fenix.compose.cfr.CFRPopup
+import org.mozilla.fenix.utils.Settings
+import mozilla.telemetry.glean.testing.GleanTestRule
+import org.junit.runner.RunWith
+import org.mozilla.fenix.GleanMetrics.TrackingProtection
+import org.mozilla.fenix.helpers.FenixRobolectricTestRunner
+
+@RunWith(FenixRobolectricTestRunner::class)
+class CFRPresenterTest {
```
mavduevskiy, line 43: Awesome tests🙌

mavduevskiy (changes requested):

app/src/main/res/values/strings.xml
mavduevskiy: Found the design, it is already underlined by `textDecoration = TextDecoration.Underline`. Sorry for raising extra attention.

app/src/main/java/org/mozilla/fenix/components/toolbar/CFRPresenter.kt
```diff
@@ -0,0 +1,130 @@
...
+import org.mozilla.fenix.ext.components
+import org.mozilla.fenix.settings.SupportUtils
+import org.mozilla.fenix.settings.SupportUtils.SumoTopic.TOTAL_COOKIE_PROTECTION
+import org.mozilla.fenix.theme.FirefoxTheme
+import org.mozilla.fenix.utils.Settings
+
+/**
+ * Delegate for handling all the business logic for showing CFRs in the toolbar.
+ *
+ * @param context [Context] used for various Android interactions.
```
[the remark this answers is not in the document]
Mugurell: A `android.content.Context` instance is needed [here](https://github.com/mozilla-mobile/fenix/blob/b45827ea07c63641c11e3e19e214cfe7d63fb55f/app/src/main/java/org/mozilla/fenix/components/toolbar/CFRPresenter.kt#L112).

app/src/main/java/org/mozilla/fenix/components/toolbar/CFRPresenter.kt
```diff
@@ -0,0 +1,130 @@
...
+ * @param store [BrowserStore] which will be observed for tabs updates
+ * @param settings [Settings] allowing to read and write persistent user settings
+ * @param toolbar [BrowserToolbar] that will serve as anchor for the CFRs
+ * @param sessionId [String] optional custom tab id.
+ */
+class CFRPresenter(
+    private val context: Context,
+    private val store: BrowserStore,
```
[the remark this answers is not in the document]
Mugurell: Done.

app/src/main/java/org/mozilla/fenix/components/toolbar/CFRPresenter.kt
```diff
@@ -0,0 +1,130 @@
...
+import org.mozilla.fenix.settings.SupportUtils.SumoTopic.TOTAL_COOKIE_PROTECTION
+import org.mozilla.fenix.theme.FirefoxTheme
+import org.mozilla.fenix.utils.Settings
+
+/**
+ * Delegate for handling all the business logic for showing CFRs in the toolbar.
+ *
+ * @param context [Context] used for various Android interactions.
+ * @param store [BrowserStore] which will be observed for tabs updates
+ * @param settings [Settings] allowing to read and write persistent user settings
```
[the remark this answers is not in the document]
Mugurell: Done.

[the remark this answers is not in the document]
Mugurell: With the other parameter documentation mainly using a feature tense using here `allows` seems a bit forced to me.

app/src/main/java/org/mozilla/fenix/components/toolbar/CFRPresenter.kt
```diff
@@ -0,0 +1,130 @@
...
+ */
+class CFRPresenter(
+    private val context: Context,
+    private val store: BrowserStore,
+    private val settings: Settings,
+    private val toolbar: BrowserToolbar,
+    private val sessionId: String? = null
+) {
+    @VisibleForTesting
+    internal var tcpCfrScope: CoroutineScope? = null
```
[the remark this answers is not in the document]
Mugurell: The scope may not be initialized if there is no need to shown the CFR and observe the store - https://github.com/mozilla-mobile/fenix/blob/b45827ea07c63641c11e3e19e214cfe7d63fb55f/app/src/main/java/org/mozilla/fenix/components/toolbar/CFRPresenter.kt#L59-L60

We could still use a lateinit and then maybe 
```kotlin
if (::tcpCfrScope.isInitialized) {
     tcpCfrScope.cancel()
}
```
or a try-catch in `stop()` but this seems uglier/lower performant.

app/src/main/java/org/mozilla/fenix/components/toolbar/CFRPresenter.kt
```diff
@@ -0,0 +1,130 @@
...
+    @VisibleForTesting
+    internal var tcpCfr: CFRPopup? = null
```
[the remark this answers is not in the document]
Mugurell: Keeping this as a nullable property ease testing with `assertNull` / `assertNotNull`.
The presenter is meant to be used for other CFRs also, naming this property just `popup` would almost assume it will have to be renamed later. Will add a `popup` suffix.

app/src/main/java/org/mozilla/fenix/components/toolbar/CFRPresenter.kt
Mugurell: The above code would indeed work but I don't like executing the work in a filtering operation while collect which is a terminal operator is empty.

app/src/main/java/org/mozilla/fenix/components/toolbar/CFRPresenter.kt
Mugurell: Can use `run` instead of `apply`.
I like how there is one block of code for initializing the popup and then one for any other business logic.

app/src/test/java/org/mozilla/fenix/components/toolbar/CFRPresenterTest.kt
Mugurell: I agree that in production code we should avoid guaranteeing that the value is not null and if the property is nullable is better to use the elvis operator.
But in tests using the not null assertion operator is a test in itself: the value can be null but it shouldn't and I verify this by using `!!` instead of a separate `assertNotNull`.

app/src/test/java/org/mozilla/fenix/components/toolbar/CFRPresenterTest.kt
Mugurell: Looked into this but the arguments are different in most cases so if having to most often pass different parameters the improvements are minimal. Eg: after refactoring

```kotlin
        val presenter = spyk(
            CFRPresenter(
                context = mockk(),
                browserStore = browserStore,
                settings = mockk {
                    every { shouldShowTotalCookieProtectionCFR } returns true
                },
                toolbar = mockk(),
                sessionId = customTab.id
            )
        ) {
            every { showTcpCfr() } just Runs
        }
```

I would get:
```kotlin
        val presenter = spyk(
            createPresenter(
                browserStore = browserStore,
                settings = mockk {
                    every { shouldShowTotalCookieProtectionCFR } returns true
                },
                sessionId = customTab.id
            )
        ) {
            every { showTcpCfr() } just Runs
        }
```

Mugurell:
Adding @Amejia481 as a reviewer for the second commit which makes this CFR controllable through Nimbus - `browserscreen.cfrs-enabled.tcp`.

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -1214,6 +1223,12 @@ class Settings(private val appContext: Context) : PreferencesHolder {
         emptyMap()
     }
 
+    private val browserscreenCFRs: Map<BrowserScreenCfr, Boolean> get() = try {
+        FxNimbus.features.browserscreen.value().cfrsEnabled
+    } catch (e: NimbusFeatureException) {
```
gabrielluong, line 1228: We don't need this try catch for NimbusFeatureException here because the docs are actually out of date. https://github.com/mozilla/application-services/issues/5085

app/src/main/res/values/strings.xml
```diff
@@ -76,6 +76,12 @@
     <!-- Content description for close button used in "contextual feature recommendation" (CFR) popups -->
     <string name="cfr_dismiss_button_default_content_description">Dismiss</string>
 
+    <!-- Total cookie protection "contextual feature recommendation" (CFR) -->
+    <!-- Text for the main message. -->
```
gabrielluong, line 80: Can we be more descriptive here? I feel like the localization team will come back to us if we just left the comment as it is.

app/src/main/java/org/mozilla/fenix/components/toolbar/CFRPresenter.kt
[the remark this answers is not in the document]
gabrielluong: `allows` doesn't sound quite right when you read it out loud. `used` is probably what yall want.

app/src/main/java/org/mozilla/fenix/components/toolbar/CFRPresenter.kt
```diff
@@ -0,0 +1,130 @@
...
+import org.mozilla.fenix.utils.Settings
+
+/**
+ * Delegate for handling all the business logic for showing CFRs in the toolbar.
+ *
+ * @param context [Context] used for various Android interactions.
+ * @param browserStore [BrowserStore] which will be observed for tabs updates
+ * @param settings [Settings] allowing to read and write persistent user settings
+ * @param toolbar [BrowserToolbar] that will serve as anchor for the CFRs
+ * @param sessionId [String] optional custom tab id used to identify the custom tab in which to show a CFR.
```
gabrielluong, line 40:
```suggestion
 * @param sessionId Optional custom tab id [String] used to identify the custom tab in which to show a CFR.
```

app/src/main/java/org/mozilla/fenix/components/toolbar/CFRPresenter.kt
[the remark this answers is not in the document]
mavduevskiy: Didn't think about testing, fair enough!

app/src/main/java/org/mozilla/fenix/components/toolbar/CFRPresenter.kt
mavduevskiy: Fair enough! Do we really need `transformWhile` if we cancel the scope anyway? Wouldn't something like this work? It does pass the tests.
```
                    .filter { it == 100 }
                    .collect{
                        tcpCfrScope?.cancel()
                        showTcpCfr()
                    }
```

app/src/test/java/org/mozilla/fenix/components/toolbar/CFRPresenterTest.kt
mavduevskiy: why we wouldn't use `assertNotNull` though? smth like this
```
presenter.tcpCfrPopup?.let {
            assertEquals("Test", it.text)
            assertEquals(anchor, it.anchor)
            assertEquals(CFRPopup.DEFAULT_WIDTH.dp, it.properties.popupWidth)
            assertEquals(CFRPopup.IndicatorDirection.DOWN, it.properties.indicatorDirection)
            assertTrue(it.properties.dismissOnBackPress)
            assertTrue(it.properties.dismissOnClickOutside)
            assertFalse(it.properties.overlapAnchor)
            assertEquals(CFRPopup.DEFAULT_INDICATOR_START_OFFSET.dp, it.properties.indicatorArrowStartOffset)
        }
       assertNotNull(presenter.tcpCfrPopup)
```

app/src/test/java/org/mozilla/fenix/components/toolbar/CFRPresenterTest.kt
mavduevskiy: It is not the only way to do it, we could also use the default parameters to set the default values to parameters the repeat themselves from test to test. I played around with the tests, and we could reduce the file size by ~60 lines we put the repeated code into constructors, and the rest would like this
```
val customTab = createCustomTab(url = "")
val browserStore = createBrowser(customTab = customTab)
val presenter = createSpyPresenter(browserStore = browserStore, sessionId = customTab.id)
```

mavduevskiy: this is what I went for, and I checked that tests were working
```
    private fun createSpyPresenter(
        context: Context = mockk(),
        anchor: View = mockk(),
        browserStore: BrowserStore = mockk(),
        settings: Settings = mockk { every { shouldShowTotalCookieProtectionCFR } returns true },
        toolbar: BrowserToolbar = mockk(),
        sessionId: String? = null
    ) = spyk(createPresenter(context, anchor, browserStore, settings, toolbar, sessionId)) {
        every { showTcpCfr() } just Runs
    }

    private fun createPresenter(
        context: Context = mockk { every { getString(R.string.tcp_cfr_message) } returns "Test" },
        anchor: View = mockk(),
        browserStore: BrowserStore = mockk(),
        settings: Settings = mockk { every { shouldShowTotalCookieProtectionCFR } returns true },
        toolbar: BrowserToolbar = mockk {
            every { findViewById<View>(R.id.mozac_browser_toolbar_security_indicator) } returns anchor
        },
        sessionId: String? = null
    ) = CFRPresenter(
        context = context,
        browserStore = browserStore,
        settings = settings,
        toolbar = toolbar,
        sessionId = sessionId
    )

    private fun createBrowser(
        normalTab: TabSessionState? = null,
        customTab: CustomTabSessionState? = null,
        selectedTabId: String? = null
    ) = BrowserStore(
        initialState = BrowserState(
            tabs = if (normalTab != null) listOf(normalTab) else listOf(),
            selectedTabId = selectedTabId,
            customTabs = if (customTab != null) listOf(customTab) else listOf()
        )
    )
```

app/src/main/java/org/mozilla/fenix/components/toolbar/CFRPresenter.kt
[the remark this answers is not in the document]
Mugurell: Used "used".

app/src/test/java/org/mozilla/fenix/components/toolbar/CFRPresenterTest.kt
Mugurell: Done.

app/src/test/java/org/mozilla/fenix/components/toolbar/CFRPresenterTest.kt
Mugurell: Done.

app/src/main/java/org/mozilla/fenix/components/toolbar/CFRPresenter.kt
Mugurell: All parameters are documented with the type at front. "Optional custom tab id [String]" seems forced.

app/src/main/res/values/strings.xml
Mugurell: Changed to "Text for the message displayed in the CFR popup promoting the TCP feature".

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
Mugurell: Thank you! Removed the try-catch.

app/src/main/res/values/strings.xml
gabrielluong: Maybe include `"contextual feature recommendation" (CFR)` and `total cookie protection (TCP)`

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -1222,6 +1223,9 @@ class Settings(private val appContext: Context) : PreferencesHolder {
         emptyMap()
     }
 
+    private val browserscreenCFRs: Map<BrowserScreenCfr, Boolean> get() =
```
mavduevskiy, line 1226: nice touch!

mavduevskiy:
@Mugurell 
When you have time, could you take a look at [this](https://github.com/mozilla-mobile/fenix/pull/26226#discussion_r942749549) discussion🙏

app/src/main/java/org/mozilla/fenix/components/toolbar/CFRPresenter.kt
Mugurell: Progress 100 is emitted multiple times and I saw in testing that this would result in multiple CFRs being shown.
Looks like there is time for the flow to emit a new item.

With the following code
```kotlin
tcpCfrScope = browserStore.flowScoped { flow ->
                flow
                    .mapNotNull { it.findCustomTabOrSelectedTab(sessionId)?.content?.progress }
                    // The "transformWhile" below ensures that the 100% progress is only collected once.
                    // .transformWhile { progress ->
                    //     emit(progress)
                    //     progress != 100
                    // }
                    .filter { it == 100 }
                    .collect {
                        tcpCfrScope?.cancel()
                        println("Mugurel: Showing the CFR")
                        showTcpCfr()
                    }
            }
```

I would always see two of those logs printed, eg:
> 2022-08-12 09:20:45.425 org.mozilla.fenix.debug I/System.out: Mugurel: Showing the CFR
2022-08-12 09:20:45.426 org.mozilla.fenix.debug I/System.out: Mugurel: Showing the CFR

app/src/main/res/values/strings.xml
Mugurell: Changed to " Text for the message displayed in the contextual feature recommendation popup promoting the total cookie protection feature. "

app/src/main/java/org/mozilla/fenix/components/toolbar/CFRPresenter.kt
Mugurell: The idea was to get the type in the documentation in a standardized way.
| with types | without types |
| --- | --- |
| image | image |

`@param sessionId [String]` does [already exist in the project](https://searchfox.org/mozilla-mobile/search?q=%40param+sessionId+%5BString%5D&path=&case=false&regexp=false).
`@param Optional custom tab id [String]` or just `tab id [String]` [does not](https://searchfox.org/mozilla-mobile/search?q=tab+id+%5BString%5D&path=&case=false&regexp=false).

Will remove the types.

Mugurell:
> @Mugurell When you have time, could you take a look at [this](https://github.com/mozilla-mobile/fenix/pull/26226#discussion_r942749549) discussion🙏

Responded in that thread.

app/src/main/java/org/mozilla/fenix/components/toolbar/CFRPresenter.kt
mavduevskiy: thanks for clarifying. I though tests should also catch that behavior, not just manual testing, especially this one

`GIVEN the TCP CFR should be shown WHEN the current tab is fully loaded THEN the TCP CFR is only shown once`

mavduevskiy (approved):
Looking good! 

It would be safer if wrong filtering would not only trigger in manual testing, but in the test as well – it looks like we have a test that does exactly that:`GIVEN the TCP CFR should be shown WHEN the current tab is fully loaded THEN the TCP CFR is only shown once`, but logging and manual testing might be enough.

Oh, one more thing. When going through review cycles, it's very convenient if new patches are added as extra commits, so I could just review the small patch that was made. There might be a lot of them in the end, and then rebasing it and collapsing into one would be the way to go (github squashes comments into one string as well, and it becomes unreadable)

Mugurell:
> Oh, one more thing. When going through review cycles, it's very convenient if new patches are added as extra commits, so I could just review the small patch that was made. There might be a lot of them in the end, and then rebasing it and collapsing into one would be the way to go (github squashes comments into one string as well, and it becomes unreadable)

I remember asking @Amejia481 about this some time ago who said that squashed commits are okay. Looking through the commit list I think the general consensus is to avoid separate fixup issues.
Separate fixup commits to me would break the entire commits structure - in this case I made an effort to split the work in multiple commits that revolve around a specific usecase and then an effort to keep the commits clean following review. 
The end result should paint an accurate portrait of what was needed to resolve a particular issue. Later commits that change the logic or update documentation following review would mean the below commits are obsolete - when checking the history on how something was resolved we would only see a small part and would need to continue to investigate to get the entire picture. So avoiding one small effort in review would mean many bigger efforts later.

Amejia481:
👋🏽 
Ideally before merging we would like to have a single commit (**when ever it is possible**). In term of reviews, I think it's OK to push multiple commits to indicate new changes (**as long as we squash before merging**), there are also other ways to avoid that, in my case I rely on the GitHub UI for telling me what have been changed, since the last push you can read more about the feature here https://github.blog/2019-07-01-mark-files-as-viewed/, either way I think it's OK, so far I don't think we have been fixed on any specific approach.

Some examples on this PR:
image


image

Amejia481:
I think we just have to chat with the reviewer to know which method is easier :)
I think it's just a matter of preference!

Amejia481 (approved):

app/src/main/java/org/mozilla/fenix/components/toolbar/CFRPresenter.kt
gabrielluong: Sorry, I didn't mean to say we had to remove the type from the front or completely. I think the other usages were fine because they were used correctly in a sentence, but the description for sessionId wasn't. The point is that we can still reference the type, but just phrase it correctly grammatically.

gabrielluong:
> Looking good!
> 
> It would be safer if wrong filtering would not only trigger in manual testing, but in the test as well – it looks like we have a test that does exactly that:`GIVEN the TCP CFR should be shown WHEN the current tab is fully loaded THEN the TCP CFR is only shown once`, but logging and manual testing might be enough.
> 
> Oh, one more thing. When going through review cycles, it's very convenient if new patches are added as extra commits, so I could just review the small patch that was made. There might be a lot of them in the end, and then rebasing it and collapsing into one would be the way to go (github squashes comments into one string as well, and it becomes unreadable)

@mavduevskiy I would actually suggest not doing fixup commits for review comments. When we move towards landing to mozilla-central, we will actually lose the ability to squash commits since we take more of a stack based approach to patches. See https://moz-conduit.readthedocs.io/en/latest/lando-user.html. In a review queue in phabricator, you will typically see multiple commits named Part 1, 2, 3, etc. If a problem was pointed out in Part 1, it would be expected that Part 1 would be amended since the changes in a commit should address any problems that are pointed instead of creating a Part 4 that might create linting changes for instance. The point would that a reader of Part 1 should see all the necessary changes to reflect the issue it is trying to resolve. 

For that reasoning, I would actually suggest switching towards amending your commits to align with the rest of organization.

Mugurell:
Updated the patch to use the new Nimbus configuration from https://github.com/mozilla-mobile/fenix/pull/26616.

Issue #26245, opened 2022-07-29
Refactor the `WallpaperManager` as a `WallpaperUseCase`.


MatthewTighe:
`WallpaperUseCase` seems to fit well with existing components in the app, as well as communicating more clearly what the intent is. A client of `WallpaperUseCase` will generally be a user-facing feature that needs to use wallpapers. This can be an opportunity to streamline the public API as well. Here are some key parts:

- fetch wallpaper metadata
- download a full-size wallpaper image
- download wallpaper thumbnails
- load the currently set wallpaper (or a wallpaper, more generally. may depend on decisions made around caching the current wallpaper in the usecase)
- load a wallpaper thumbnail

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21149)

delia-pop:
Verified as fixed on the latest Nightly 105.0a1 from 08/15 with Google Pixel 6 (Android 12). The wallpaper feature works as expected, no issues were encountered.

Issue #25985, opened 2022-07-11
Remove ability to toggle wallpapers by tapping Firefox logo


MatthewTighe:
The ability to toggle the selected wallpaper by tapping the Firefox logo introduces additional constraints on the freedom to choose different wallpaper download strategies. It is also a feature that can be activated accidentally pretty easily, leading to unexpected results. It should be removed for now.

Acceptance criteria:
- Logo can no longer be tapped to switch wallpaper
- Logo no longer animates
- Settings data and preference keys used to determine if logo should animate are removed

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-20978)

sunilk9211:
hi @MatthewTighe can i do this?

MatthewTighe:
@sunilk9211 Sure, this seems like a good issue for a community contributor to pick up! I added some more explicit acceptance criteria to the OP to try and help guide you through it. I would start by looking at the following functions:

[WallpaperManager::switchToNextWallpaper](https://searchfox.org/mozilla-mobile/rev/22f697fb68894c09dec5d3c164bcec92b0c2fbe1/fenix/app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperManager.kt#91)
[WallpaperManager::animateLogoIfNeeded](https://searchfox.org/mozilla-mobile/rev/22f697fb68894c09dec5d3c164bcec92b0c2fbe1/fenix/app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperManager.kt#215)

Let me know if I can help with anything else 🙂 Thank you!

sunilk9211:
Thanks @MatthewTighe. 😊
I have one question. Do we want to remove all the code related to this feature?
I can see in code, this feature is also used for some promotion related stuff.

MasterInQuestion:
Is not it optional: "Settings" -> "Homepage" -> "Wallpapers" -> "Change wallpaper by tapping Firefox homepage logo"?

MatthewTighe:
@MasterInQuestion It is optional, but the decision has still been made to remove it based on the reasons quoted in the original issue, in addition to the fact that it had low discoverability.

MasterInQuestion:
Rather than removing the function completely, would not it be less destructive to just override the user config value before the version?

MatthewTighe:
[...]

PR #26249, opened 2022-07-29
Fixes: #26052 remove WallpaperManager::updateWallpaper

### LINKED ISSUE #26052: Remove `WallpaperManager::updateWallpaper`
As part of https://github.com/mozilla-mobile/fenix/issues/26034 and as a follow-up to https://github.com/mozilla-mobile/fenix/pull/26051, we should remove `updateWallpaper` from the `WallpaperManager`. This method had a side-effect on the state of the app that should be handled by dispatching actions to the `AppStore`. 

This will be the first step in switching the `WallpaperManager` to a `WallpaperUseCase`, and consumers of the UseCase should instead load the appropriate bitmap for a wallpaper and handle their rendering logic inline. 

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21029)

MatthewTighe:
This is part of https://github.com/mozilla-mobile/fenix/issues/26034. The main point of this PR was to remove `updateWallpaper`, which included the side-effect of updating the `currentWallpaper`, which should instead now be handled by dispatching the `UpdateCurrentWallpaper`.

The next step will likely be to refactor the `WallpaperManager` into a `WallpaperUseCase`, as filed in https://github.com/mozilla-mobile/fenix/issues/26245. That patch will probably introduce the largest number of changes in this initiative.

### Pull Request checklist

### QA

- [] **QA Needed**

### GitHub Automation
Fixes #26052

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperManager.kt
+++ b/app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperManager.kt
@@ -59,30 +59,6 @@ class WallpaperManager(
     }
 
     /**
-     * Apply the [newWallpaper] into the [wallpaperContainer] and update the [currentWallpaper].
-     */
-    fun updateWallpaper(wallpaperContainer: ImageView, newWallpaper: Wallpaper) {
-        val context = wallpaperContainer.context
-        if (newWallpaper == defaultWallpaper) {
-            wallpaperContainer.visibility = View.GONE
-            logger.info("Wallpaper update to default background")
-        } else {
-            val bitmap = loadSavedWallpaper(context, newWallpaper)
-            if (bitmap == null) {
-                val message = "Could not load wallpaper bitmap. Resetting to default."
-                logger.error(message)
-                currentWallpaper = defaultWallpaper
-                wallpaperContainer.visibility = View.GONE
-                return
-            } else {
-                wallpaperContainer.visibility = View.VISIBLE
-                scaleBitmapToBottom(bitmap, wallpaperContainer)
-            }
-        }
-        currentWallpaper = newWallpaper
-    }
-
-    /**
      * Download all known remote wallpapers.
      */
     suspend fun downloadAllRemoteWallpapers() {
@@ -104,7 +80,7 @@ class WallpaperManager(
         } else {
             values[index]
         }.also {
-            appStore.dispatch(AppAction.WallpaperAction.UpdateCurrentWallpaper(it))
+            currentWallpaper = it
         }
     }
 
@@ -139,10 +115,10 @@ class WallpaperManager(
     /**
      * Load a wallpaper that is saved locally.
      */
-    fun loadSavedWallpaper(context: Context, wallpaper: Wallpaper): Bitmap? =
-        when (wallpaper) {
-            is Wallpaper.Local -> loadWallpaperFromDrawables(context, wallpaper)
-            is Wallpaper.Remote -> loadWallpaperFromDisk(context, wallpaper)
+    fun Wallpaper.load(context: Context): Bitmap? =
+        when (this) {
+            is Wallpaper.Local -> loadWallpaperFromDrawables(context, this)
+            is Wallpaper.Remote -> loadWallpaperFromDisk(context, this)
             else -> null
         }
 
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperManager.kt
```diff
@@ -139,10 +115,10 @@ class WallpaperManager(
     /**
      * Load a wallpaper that is saved locally.
      */
-    fun loadSavedWallpaper(context: Context, wallpaper: Wallpaper): Bitmap? =
-        when (wallpaper) {
-            is Wallpaper.Local -> loadWallpaperFromDrawables(context, wallpaper)
-            is Wallpaper.Remote -> loadWallpaperFromDisk(context, wallpaper)
+    fun Wallpaper.load(context: Context): Bitmap? =
```
MozillaNoah, line 118: nit: Should this extension instead live in a extension function file?

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperManager.kt
```diff
@@ -163,7 +139,14 @@ class WallpaperManager(
...
     }.getOrNull()
 
-    private fun scaleBitmapToBottom(bitmap: Bitmap, view: ImageView) {
+    /**
+     * This will scale the received [Bitmap] to the size of the [view]. It retains the bitmap's
+     * original aspect ratio, but will shrink or enlarge it to fit the viewport. If bitmap does not
+     * correctly fit the aspect ratio of the view, it will be shifted to prioritize the bottom-left
+     * of the bitmap.
+     */
+    fun Bitmap.scaleBitmapToBottomOfView(view: ImageView) {
```
MozillaNoah, line 148: nit: Should this extension instead live in a extension function file?

MozillaNoah (approved):
couple nits. Unrelated follow-up, I also noticed that the `WallpaperThumbnailsPreview` doesn't render, probably because it's using the `.components` extension.

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperManager.kt
MatthewTighe: This one technically could, but structuring it this way was an intentional plan for refactoring the `WallpaperManager` as a `WallpaperUseCase`. It forces a client to have a reference to the `UseCase` to be able to load wallpapers. That way, we can treat the UseCase as an abstraction barrier. Right now, all the loading is done using implicit `Context`-dependent stuff, but I imagine the callstack would look more like this soon:

```
with (wallpaperUseCase) {
    wallpaper.load() -> wallpaperUseCase.fileManager.loadWallpaper()
}
```

Not sure that notation really makes sense lol. But basically we could inject the `context` (or a `WallpaperLoader` or whatever else we wanted to use to actually do disk reading) into the object hierarchy and hide it from call sites of `wallpaper.load()`.

Sidenote: the pattern here might be more clear if we were on Kotlin 1.6.20 and could use [context receivers](https://blog.jetbrains.com/kotlin/2022/02/kotlin-1-6-20-m1-released/#prototype-of-context-receivers-for-kotlin-jvm).

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperManager.kt
MatthewTighe: This one would make more sense to move into an extension function file, IMO, because it isn't necessarily a strictly-wallpaper related function. I just always worry about polluting namespaces without an obvious need. I'd defer to your opinion if you think we should move it

MatthewTighe:
> couple nits. Unrelated follow-up, I also noticed that the WallpaperThumbnailsPreview doesn't render, probably because it's using the .components extension.

Oh thanks, that would definitely explain it! Are there any easy ways around that or would I need to manually instantiate anything I use in previews?

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperManager.kt
MozillaNoah: I don't feel super strongly either way, so if you feel this _will_ be used elsewhere, I'd lean that way. Otherwise, it can just live here for now.

MozillaNoah:
> > couple nits. Unrelated follow-up, I also noticed that the WallpaperThumbnailsPreview doesn't render, probably because it's using the .components extension.
> 
> Oh thanks, that would definitely explain it! Are there any easy ways around that or would I need to manually instantiate anything I use in previews?

You would need to either manually instantiate the real-world use-case or just abstract-out that dependency so either only a lambda or a list<> can be passed-in. If something further upstream needs the context, then we have to do an `if` with `inComposePreview` (see: `Favicon`) to just provide an override for `@Preview`s

MatthewTighe:
Note: the test that was removed in the diff was failing, but that functionality will be removed soon anyway in https://github.com/mozilla-mobile/fenix/issues/25985. I was having trouble getting the UI tests to run locally, and there's quite a lot of follow-up work for this PR so I decided to go with the fastest option. I'll squash before I land

PR #26299, opened 2022-08-03
For #26239 - Avoid displaying multiple download items for the same file

### LINKED ISSUE #26239: [Bug]: Duplicate items in the Downloads list when a deleted file is saved again
### Steps to reproduce

1. Save a file (an image or pdf for example).
2. Go to the Files manager app and delete the file from the device.
3. Check that the file is no longer in the app's Downloads list.
4. Download again the same file.
5. Go to Downloads and check the list again.

### Expected behaviour

The saved file is displayed once, reflecting the app's files in the Files manager.

### Actual behaviour

The saved file is now displayed twice, though there's only one on the device.
Repeating these steps will create more duplicates every time.

### Device name

Pixel 2, Pixel 3

### Android version

Android 9, 11

### Firefox release type

Firefox

### Firefox version

103.1.0, 104.b2

### Device logs

_No response_

### Additional information

_No response_

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21145)

Alexandru2909:
Added `distinctBy` call to `provideDownloads` which filters multiple instances of a `DownloadItem` for the same file.

https://user-images.githubusercontent.com/35462038/182546143-f49dad38-ce1f-4418-95d0-a5354e876e6a.mp4

### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation
Fixes #26239

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/library/downloads/DownloadFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/library/downloads/DownloadFragment.kt
@@ -91,6 +91,7 @@ class DownloadFragment : LibraryPageFragment<DownloadItem>(), UserInteractionHan
     @VisibleForTesting
     internal fun provideDownloads(state: BrowserState): List<DownloadItem> {
         return state.downloads.values
+            .distinctBy { it.fileName }
             .sortedByDescending { it.createdTime } // sort from newest to oldest
             .map {
                 DownloadItem(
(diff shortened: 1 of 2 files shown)


Mugurell (commented):
Wouldn't filtering by just distinct `filename` be enough?
I feel like this is what the issues resolves to: after removing the file on disk it's name is also cleared for reuse so when the same file is downloaded it can reuse the same name and so for the current session we have >=two downloads with the same name, both completed, both existing on disk (because based on the name the filepath is the same).
But otherwise it should not be possible to persist different files with the same names, so just this check should be enough.

Mugurell (approved):
LGTM!
Great that you also added a test for this!

PR #26309, opened 2022-08-03
For #26308 - Add feature flag for MR onboarding

### LINKED ISSUE #26308: Feature flag for MR Onboarding
### Acceptance Criteria

-     Feature flag in place for MR Onboarding that can be controlled through Nimbus
-     Default setting is off

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21192)

sarah541:
### Pull Request checklist

### QA

- [ ] **QA Needed**

### GitHub Automation
Fixes #26308

### DIFF
--- a/.experimenter.yaml
+++ b/.experimenter.yaml
@@ -58,6 +58,14 @@ nimbus-validation:
     settings-title:
       type: string
       description: The title of displayed in the Settings screen and app menu.
+onboarding:
+  description: The onboarding features displayed on homescreen.
+  hasExposure: true
+  exposureDescription: ""
+  variables:
+    sections-enabled:
+      type: json
+      description: This property provides a lookup table of whether or not the given section should be enabled.
 pocket-sponsored-stories:
   description: A feature showing sponsored stories in between the other Pocket recommended stories on homescreen.
   hasExposure: true
--- a/app/src/main/java/org/mozilla/fenix/FeatureFlags.kt
+++ b/app/src/main/java/org/mozilla/fenix/FeatureFlags.kt
@@ -24,6 +24,11 @@ object FeatureFlags {
     const val syncAddressesFeature = false
 
     /**
+     * Enables the on-boarding sync cfr on the home screen.
+     */
+    val showSynCFR = Config.channel.isDebug
+
+    /**
      * Enables the "recent" tabs feature in the home screen.
      */
     const val showRecentTabsFeature = true
--- a/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
+++ b/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
@@ -44,6 +44,7 @@ import org.mozilla.fenix.ext.components
 import org.mozilla.fenix.ext.getPreferenceKey
 import org.mozilla.fenix.nimbus.FxNimbus
 import org.mozilla.fenix.nimbus.HomeScreenSection
+import org.mozilla.fenix.nimbus.OnboardingSection
 import org.mozilla.fenix.settings.PhoneFeature
 import org.mozilla.fenix.settings.deletebrowsingdata.DeleteBrowsingDataOnQuitType
 import org.mozilla.fenix.settings.logins.SavedLoginsSortingStrategyMenu
@@ -1200,6 +1201,9 @@ class Settings(private val appContext: Context) : PreferencesHolder {
         default = false
     )
 
+    private val onboardScreenSection: Map<OnboardingSection, Boolean> get() =
+        FxNimbus.features.onboarding.value().sectionsEnabled
+
     private val homescreenSections: Map<HomeScreenSection, Boolean> get() = try {
         FxNimbus.features.homescreen.value().sectionsEnabled
     } catch (e: NimbusFeatureException) {
@@ -1210,6 +1214,16 @@ class Settings(private val appContext: Context) : PreferencesHolder {
         appContext.getPreferenceKey(R.string.pref_key_history_metadata_feature),
         default = { homescreenSections[HomeScreenSection.RECENT_EXPLORATIONS] == true },
         featureFlag = FeatureFlags.historyMetadataUIFeature || isHistoryMetadataEnabled
+    )
+
+    /**
+     * Indicates if sync on-boarding CFR should be shown
+     * Returns true if the [FeatureFlags.showSynCFR] and [R.string.pref_key_should_show_sync_cfr] are true.
+     */
+    var showSyncCFR by lazyFeatureFlagPreference(
+        appContext.getPreferenceKey(R.string.pref_key_should_show_sync_cfr),
+        featureFlag = FeatureFlags.showSynCFR,
+        default = { onboardScreenSection[OnboardingSection.SYNC_CFR] == true },
     )
 
     /**
--- a/app/src/main/res/values/preference_keys.xml
+++ b/app/src/main/res/values/preference_keys.xml
@@ -236,6 +236,8 @@
     <string name="pref_key_has_inactive_tabs_auto_close_dialog_dismissed" translatable="false">pref_key_has_inactive_tabs_auto_close_dialog_dismissed</string>
     <!-- A value  of `true` means the jump back in onboarding popup has not been shown yet -->
     <string name="pref_key_should_show_jump_back_in_tabs_popup" translatable="false">pref_key_should_show_jump_back_in_tabs_popup</string>
+    <!-- A value  of `true` means the  on-boarding sync CFR have not been shown yet -->
+    <string name="pref_key_should_show_sync_cfr" translatable="false">pref_key_should_show_sync_cfr</string>
 
     <string name="pref_key_debug_settings" translatable="false">pref_key_debug_settings</string>
 
--- a/nimbus.fml.yaml
+++ b/nimbus.fml.yaml
@@ -33,6 +33,23 @@ features:
             "contile-top-sites": false,
           }
         }
+  onboarding:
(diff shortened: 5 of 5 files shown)


app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -887,6 +887,14 @@ class Settings(private val appContext: Context) : PreferencesHolder {
         default = true
     )
 
+    /**
+     * Indicates if the onboarding features should be shown.
+     */
+    var shouldShowOnboardingFeatures by booleanPreference(
```
MozillaNoah, line 893: will this control all of _onboarding_, or just the new enhancements?

sarah541: Currently in scope are Sync CFR and JumpBackIn CFR

.experimenter.yaml
```diff
@@ -15,6 +15,9 @@ homescreen:
   hasExposure: true
   exposureDescription: ""
   variables:
+    onboarding:
```
Mugurell, line 18: Would it make sense to have a new `variable` for each of the homescreen onboarding features? 
Or a collection similar to the below `sections-enabled` which would allow individually controlling each onboarding feature shown on homescreen?

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
Mugurell: Using just one shared preference and nimbus experiment for multiple features may cause issues in the scenario of the CFRs having different preconditions to be shown and based on https://mozilla-hub.atlassian.net/browse/FNXV2-20988 looks like this is the case for these two CFRs.
If only one flag controls multiple CFRs and not all will be shown at the exact same time based on just one flag check then when the first cfr is shown and persists that onboarding was shown (for it) no other onboarding CFRs will be shown afterwards if they check the same flag which now says that onboarding was shown.

Seems like using a Nimbus experiment to control each CFR is part of the requirements of each CFR ticket so maybe it would be best to leave it for each CFR implementation to add it's own feature flag?

Mugurell: Using just one shared preference and nimbus experiment for multiple features may cause issues in the scenario of the CFRs having different preconditions to be shown and based on https://mozilla-hub.atlassian.net/browse/FNXV2-20988 looks like this is the case for these two CFRs.
If only one flag controls multiple CFRs and not all will be shown at the exact same time based on just one flag check then when the first cfr is shown and persists that onboarding was shown (for it) no other onboarding CFRs will be shown afterwards if they check the same flag which now says that onboarding was shown.

Seems like using a Nimbus experiment to control each CFR is part of the requirements of each CFR ticket so maybe it would be best to leave it for each CFR implementation to add it's own feature flag?

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -887,6 +887,14 @@ class Settings(private val appContext: Context) : PreferencesHolder {
...
+        appContext.getPreferenceKey(R.string.pref_key_should_show_onboarding_features),
+        default = FxNimbus.features.homescreen.value().onboarding
```
gabrielluong, line 895: Can you implement this similar to `showRecentTabsFeature`? I don't think we want to call `FxNimbus.features.homescreen.value()` directly since it can throw an exception.

app/src/main/res/values/preference_keys.xml
```diff
@@ -236,6 +236,8 @@
     <string name="pref_key_has_inactive_tabs_auto_close_dialog_dismissed" translatable="false">pref_key_has_inactive_tabs_auto_close_dialog_dismissed</string>
     <!-- A value  of `true` means the jump back in onboarding popup has not been shown yet -->
     <string name="pref_key_should_show_jump_back_in_tabs_popup" translatable="false">pref_key_should_show_jump_back_in_tabs_popup</string>
+    <!-- A value  of `true` means the  onboarding features have not been shown yet -->
```
gabrielluong, line 239:
```suggestion
    <!-- A value  of `true` means the onboarding features have not been shown yet -->
```

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
gabrielluong: I am also expecting that we should have something in `FeatureFlags` so that we can manually toggle the pref on for our local testing.

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
sarah541: I was under the impression they both will be behind the same feature flag and so I will implement another setting different for both CFRs. When sync CFR has been shown, the setting to it will be made false, however no change in the onboarding setting flag

.experimenter.yaml
Amejia481: > Or a collection similar to the below sections-enabled which would allow individually controlling each onboarding feature shown on homescreen?

Agree, this approach will more scalable!

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
jonalmeida: I've been working on Nimbus recently - it's safe to call `FxNimbus.features.homescreen.value()` directly. It doesn't throw an exception any more.

You don't need to pass it a context anymore too (which was part of the cause of the problem then.

nimbus.fml.yaml
```diff
@@ -33,6 +33,25 @@ features:
...
         }
+  onboardingcreen:
+    description: The homescreen that the user goes to when they press home or new tab.
+    variables:
+      sections-enabled:
+        description: "This property provides a lookup table of whether or not the given section should be enabled."
+        type: Map<OnboardingScreenSection, Boolean>
+        default:
+          {
+            "jump-back-in-cfr": true,
```
gabrielluong, line 44: Do we need this "jump-back-in-cfr"? I don't see it used in this PR, but not sure if there's another requirement somewhere.

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -1209,6 +1210,12 @@ class Settings(private val appContext: Context) : PreferencesHolder {
         default = false
     )
 
+    private val onboardScreenSection: Map<OnboardingScreenSection, Boolean> get() = try {
+        FxNimbus.features.onboardingcreen.value().sectionsEnabled
+    } catch (e: NimbusFeatureException) {
```
gabrielluong, line 1215: We can remove this try catch since NimbusFeatureException isn't actually thrown. https://github.com/mozilla/application-services/issues/5085

app/src/main/java/org/mozilla/fenix/FeatureFlags.kt
```diff
@@ -23,6 +23,11 @@ object FeatureFlags {
      */
     const val syncAddressesFeature = false
 
+    /**
+     * Enables the on-boarding sync cfr  in the home screen.
```
gabrielluong, line 27:
```suggestion
     * Enables the on-boarding sync cfr in the home screen.
```

nimbus.fml.yaml
```diff
@@ -33,6 +33,25 @@ features:
             "contile-top-sites": false,
           }
         }
+  onboardingcreen:
```
gabrielluong, line 36:
```suggestion
  onboarding-screen:
```

.experimenter.yaml
```diff
@@ -58,6 +58,14 @@ nimbus-validation:
     settings-title:
       type: string
       description: The title of displayed in the Settings screen and app menu.
+onboardingcreen:
```
gabrielluong, line 61:
```suggestion
onboardingcreen:
```
Similar comment below

gabrielluong, line 61: I am not the most familiar person with the experimenter.yaml and nimbus.fml.yml. So, you'll want to get another review from someone who is more familiar with the two.

app/src/main/java/org/mozilla/fenix/FeatureFlags.kt
```diff
@@ -23,6 +23,11 @@ object FeatureFlags {
      */
     const val syncAddressesFeature = false
 
+    /**
+     * Enables the on-boarding sync cfr on the home screen.
+     */
+    const val showSynCFR = true
```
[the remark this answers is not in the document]
Amejia481: I got a bit confused as the acceptance criteria mentions `Default setting is off`.

nimbus.fml.yaml
```diff
@@ -33,6 +33,23 @@ features:
...
         }
+  onboarding:
+    description: The homescreen that the user goes to when they press home or new tab.
+    variables:
+      sections-enabled:
+        description: "This property provides a lookup table of whether or not the given section should be enabled."
+        type: Map<OnboardingSection, Boolean>
+        default:
+          {
+            "sync-cfr": true
```
Amejia481, line 44: I think this should be `false`, until we enable the feature completely, meanwhile we want to enable it only we get the value from nimbus server.

Amejia481: If we leave it on, it will signal that nimbus is indicating us that functionality should be ON unless we are in nightly. As we are introducing this new functionality the goal is that feature is OFF unless the nimbus server indicate so, that will allow us to turn the feature ON  for some users and see how they interact with it. And later made a decision if we want it ON for user then we can enable this.

Amejia481 (approved):

Issue #26033, opened 2022-07-15
Current wallpaper storage should be abstracted away from `Settings`


MatthewTighe:
There should be an abstraction layer implemented to cover the usage of `Settings` to track the current wallpaper. This is with the goal of eventually porting the bulk of the Wallpaper implementation to A-C, where it can't be guaranteed that the storage layer is the same as it is in Fenix.

Create an interface that can define current wallpaper storage and add an implementation of that interface that uses settings under the hood as is currently done. 

Note that this may require code to handle migrations to a different storage implementation down the line.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21016)

cpeterso:
Moved to bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1809807

Change performed by the [Move to Bugzilla add-on](https://addons.mozilla.org/en-US/firefox/addon/move-to-bugzilla/).

PR #26324, opened 2022-08-05
Fixes #26245: refactor the WallpaperManager as several WallpaperUseCases

### LINKED ISSUE #26245: Refactor the `WallpaperManager` as a `WallpaperUseCase`.

MatthewTighe:
This is intended to refactor the main APIs for interacting with the Wallpaper feature, in order to stabilize them for the rest of the upcoming wallpaper work. This should mostly be a lift from the `WallpaperManager` to the various `UseCases` that were added. Note that there's still improvements to be made to the underlying code, but those should all be achievable down the road now that there's a stable API. Some of those are by https://github.com/mozilla-mobile/fenix/issues/26034 but I think I have a couple more to file. 

There are notably tests missing for the `LoadBitmap` use case. This is not a degradation is test coverage, as this behavior hadn't been tested previously either. My plan is to test those more thoroughly after https://github.com/mozilla-mobile/fenix/issues/26033, which will move the `Bitmap`-specific loading into a `WallpaperStorage` abstraction. 

There's also an awkward bit of code leftover in `WallpaperManager`. That is waiting for the completion of https://github.com/mozilla-mobile/fenix/issues/25985 which a contributor has graciously volunteered for. 

In general, I'll be planning improvements to the underlying code for all of these usecases, some of which will probably happen as a matter of course as we add support for things like downloading thumbnails.

### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation
Fixes #26245


app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -77,7 +76,7 @@ import org.mozilla.fenix.wallpapers.WallpaperManager
 fun WallpaperSettings(
     wallpapers: List<Wallpaper>,
     defaultWallpaper: Wallpaper,
-    loadWallpaperResource: (Wallpaper) -> Bitmap?,
+    loadWallpaperResource: suspend (Wallpaper) -> Bitmap?,
```
MozillaNoah, line 79: didn't even think that lambdas could be `suspends`. neat

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettingsFragment.kt
```diff
@@ -44,18 +48,23 @@ class WallpaperSettingsFragment : Fragment() {
...
                 FirefoxTheme {
-                    var currentWallpaper by remember { mutableStateOf(wallpaperManager.currentWallpaper) }
+                    val wallpapers = appStore.observeAsComposableState { state ->
+                        state.wallpaperState.availableWallpapers
+                    }.value ?: listOf()
+                    var currentWallpaper = appStore.observeAsComposableState { state ->
+                        state.wallpaperState.currentWallpaper
+                    }.value ?: Wallpaper.Default
                     var wallpapersSwitchedByLogo by remember { mutableStateOf(settings.wallpapersSwitchedByLogoTap) }
                     WallpaperSettings(
```
MozillaNoah, line 58: nit: could we add a newline between `wallpapersSwitchedByLogo` and `WallpaperSettings`

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettingsFragment.kt
```diff
@@ -44,18 +48,23 @@ class WallpaperSettingsFragment : Fragment() {
...
-                        defaultWallpaper = WallpaperManager.defaultWallpaper,
+                        wallpapers = wallpapers,
+                        defaultWallpaper = Wallpaper.Default,
                         loadWallpaperResource = { wallpaper ->
-                            with(wallpaperManager) { wallpaper.load(context) }
+                            wallpaperUseCases.loadBitmap(wallpaper)
                         },
                         selectedWallpaper = currentWallpaper,
                         onSelectWallpaper = { selectedWallpaper: Wallpaper ->
                             currentWallpaper = selectedWallpaper
```
MozillaNoah, line 66: the lambdas for `loadWallpaperResource`, `onSelectWallpaper`, and `onTapLogoSwitchCheckedChange` are pretty substantial. Would it make sense to abstract this out into an interactor/controller? It would keep the instantiation clean and allow for the business logic to be testable.

app/src/main/java/org/mozilla/fenix/wallpapers/WallpapersUseCases.kt
```diff
@@ -0,0 +1,246 @@
...
+
+/**
+ * Contains use cases related to the wallpaper feature.
+ *
+ * @param context Used for various file and configuration checks.
+ * @param store Will receive dispatches of metadata updates like the currently selected wallpaper.
+ * @param client Handles downloading wallpapers and their metadata.
+ * @param strictMode Required for determining some device state like current locale and file paths.
+ *
+ * @property initialize Usecase for initializing wallpaper feature. Should be called eagerly.
```
MozillaNoah, line 34: `Should be called eagerly.` This is somewhat ambiguous to me, as someone who has not yet used the Wallpapers API. What are you trying to convey here?

app/src/main/java/org/mozilla/fenix/wallpapers/WallpapersUseCases.kt
```diff
@@ -0,0 +1,246 @@
...
+        private val currentLocale: String,
+        private val possibleWallpapers: List<Wallpaper> = allWallpapers,
+    ) : InitializeWallpapersUseCase {
+
+        /**
+         * Downloads the currently available wallpaper metadata from a remote source.
+         * Updates the [store] with that metadata and with the selected wallpaper found in storage.
+         * Removes any unused promotional or time-limited assets from local storage.
+         */
+        override suspend operator fun invoke() {
```
MozillaNoah, line 90: for reviewing context, what percentage of `invoke` is just re-located code from `WallpaperManager`?

app/src/main/java/org/mozilla/fenix/wallpapers/WallpapersUseCases.kt
```diff
@@ -0,0 +1,246 @@
...
+                )
+                downloadAllRemoteWallpapers(availableWallpapers)
+                store.dispatch(AppAction.WallpaperAction.UpdateAvailableWallpapers(availableWallpapers))
+                store.dispatch(AppAction.WallpaperAction.UpdateCurrentWallpaper(currentWallpaper))
+            }
+        }
+
+        private fun getAvailableWallpapers() = possibleWallpapers
+            .filter(::filterExpiredRemoteWallpapers)
+            .filter(::filterPromotionalWallpapers)
```
MozillaNoah, line 113: Would this potentially be more efficient?
```suggestion
            .filter { filterExpiredRemoteWallpapers(it) && filterPromotionalWallpapers(it) } 
```

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -296,10 +295,8 @@ private fun WallpaperThumbnailsPreview() {
         val wallpaperManager = context.components.wallpaperManager
```
MozillaNoah, line 295: the preview isn't building. let me know if you wanna pair up on this.

MozillaNoah (commented):
initial round of comments before I poke around the actual functionality on a device.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
MatthewTighe: Totally! `suspend` is what `@Composable` is modeled after, I believe so the two are basically interchangeable (in terms of how they affect what they modify)

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettingsFragment.kt
MatthewTighe: `loadWallpaperResource` doesn't seem like it would benefit hugely, as it would just change from `wallpaperUseCases.loadBitmap(wallpaper) to `wallpaperSettingsInteractor.loadBitmap(wallpaper)`.

For the others, they certainly do look out of place in terms of size. It looks like they're just recording metrics though, and could otherwise be a single line as well. Would the metrics be an appropriate thing to just move into the use case, or is an interactor a better place? I'm just a little hesitant to add another layer of abstraction if we can find a way to avoid it.

`onTapLogoSwitchCheckedChange` should also be removed as part of https://github.com/mozilla-mobile/fenix/issues/25985 so we can potentially ignore that for now as well. Come to think of it, should the metric be removed directly from `metrics.yaml` as part of that work or is there a specific process for removing telemetry? Unrelated but would be good to know for later

app/src/main/java/org/mozilla/fenix/wallpapers/WallpapersUseCases.kt
MatthewTighe: I'll update this with some more detail, thanks for calling it out!

app/src/main/java/org/mozilla/fenix/wallpapers/WallpapersUseCases.kt
MatthewTighe: It should all be re-located and slightly re-arranged code from `WallpaperManager`. Outside of making a couple things explicitly `suspend`ing, this PR shouldn't change any functionality.

app/src/main/java/org/mozilla/fenix/wallpapers/WallpapersUseCases.kt
MatthewTighe: Oh yeah, that definitely seems better. IIRC we added them separately which is probably led to that. I don't really like the names much either TBH. Maybe something like `.filter { !it.isExpired() && !it.isUnavailablePromotion() }` would convey the meaning a bit better?

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
MatthewTighe: I'll just need to remember to check them in the future. I ran into problems with my M1 when I got started and got in the habit of not doing so lol. I'll take a look and let you know if I could use some help, thanks 🙂

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettingsFragment.kt
MatthewTighe: Ah, it looks like there was actually some additional logic in `onSelectWallpaper` that can be removed entirely now that the wallpaper data is being observed from the store. I'll update that accordingly and move the metrics in the UseCase for now. If that's not a good place for the metrics to live we can revisit, but it should at least address the concern of high lambda complexity

MozillaNoah: I feel that anything not related to the UI should live within an interactor/controller, to keep the instantiation here clean and keep the UI as dumb as possible. It's entirely possible that, long term, we move away from Fragments and go 100% Compose navigation, and it'll be easier at that time just dropping-in the interactor rather than having to refactor the fragment AND the business logic. Additionally, the use case operations could be abstracted there as well. 

As for metrics, I'd make a separate PR for removing them from `metrics.yaml` and doing whatever other clean up and then tagging-in Roger for the blessing, since he's a data steward. Anecdotally, I didn't have to do anything additional admin for my search term tab groups removal.

app/src/main/java/org/mozilla/fenix/wallpapers/WallpapersUseCases.kt
MozillaNoah: oooo, yeah, making extension or class functions and cleaning up the syntax would make it look crisp.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettingsFragment.kt
MatthewTighe: > and it'll be easier at that time just dropping-in the interactor rather than having to refactor the fragment AND the business logic.

My intention was to have the UseCase take this role. The latest changes hopefully make that a little more clear. I personally would prefer not to wrap the UseCase in an interactor/controller, as I worry that they would act essentially as passthrough objects and I think that can make our codebase more difficult to navigate. I may not be fully understanding your suggestion or our patterns though. My concern is basically that this:

```
loadWallpaperResource = { wallpaperUseCases.loadBitmap(it) }
```

seems more straightforward than this:
```
loadWallpaperResource = { interactor.loadBitmap(it) }

interface WallpaperSettingsInteractor {
    fun loadWallpaperBitmap(wallpaper: Wallpaper)
}

class SessionControlInteractor(val wallpaperSettingsController: WallpaperSettingsController) : WallpaperSettingsInteractor {
    override fun loadWallpaperBitmap(wallpaper: Wallpaper) {
        wallpaperSettingsController.loadWallpaperBitmap(wallpaper)
    }
}

class WallpaperSettingsController(val useCases: WallpaperUseCases) {
    fun loadWallpaperBitmap(wallpaper: Wallpaper) {
        useCases.loadBitmap(it)
    }
}
```

> As for metrics, I'd make a separate PR for removing them from metrics.yaml and doing whatever other clean up and then tagging-in Roger for the blessing, since he's a data steward. Anecdotally, I didn't have to do anything additional admin for my search term tab groups removal.

Great, thanks!

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
MatthewTighe: This should be working now! We are planning to make all wallpapers sourced remotely, so unfortunately there isn't much to show. But it at least renders!

app/src/main/java/org/mozilla/fenix/wallpapers/WallpapersUseCases.kt
MatthewTighe: Updated!

app/src/main/java/org/mozilla/fenix/wallpapers/WallpapersUseCases.kt
MatthewTighe: Added some more detail

MatthewTighe:
Marked most of those as resolved and added a bit of detail. Please unresolve anything that doesn't seem fully addressed! Though I think there could be more discussion to be had around interactors/controllers. Also happy to file follow-ups for anything, this PR is just intended to stabilize the API in hopes we can start parallelizing some of the additional wallpaper work. I think viewing this particularly in the lens of whether the API surface is what we want moving forward would be super helpful 😄

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettingsFragment.kt
MozillaNoah: I think the main thing I was looking for was the abstraction, and the way you wrote the use cases seems to do that well and could be a 2.0, cleaner, more direct version of the interactor/controller pattern. You also have tests written for them, so ✅

MozillaNoah:
Tested this locally, and it's looking good!

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -292,16 +289,12 @@ private fun WallpaperLogoSwitch(
 @Composable
 private fun WallpaperThumbnailsPreview() {
     FirefoxTheme(theme = Theme.getTheme()) {
-        val context = LocalContext.current
-        val wallpaperManager = context.components.wallpaperManager
 
         WallpaperSettings(
```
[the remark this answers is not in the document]
MatthewTighe: I believe the plan is actually to move the local wallpapers to remote, which is why I didn't include them here. This should reduce APK bloat for uninterested users as well as reduce the complexity in needing to handle separate types of wallpapers.

app/src/test/java/org/mozilla/fenix/wallpapers/WallpapersUseCasesTest.kt
```diff
@@ -0,0 +1,219 @@
...
+        coEvery { mockFileManager.lookupExpiredWallpaper(any()) } returns null
+        every { mockSettings.currentWallpaper } returns ""
+        every { mockSettings.currentWallpaper = capture(slot) } just runs
+
+        WallpapersUseCases.DefaultSelectWallpaperUseCase(
+            mockSettings,
+            appStore
+        ).invoke(selectedWallpaper)
+
+        appStore.waitUntilIdle()
```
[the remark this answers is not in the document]
MatthewTighe: Yeah, definitely. If I'd only done the interactor and shoved the metric in it I could've kept this test class fast, my hubris has caught up with me lmao. The glean test rule (and the `AndroidJUnit4` runner) add about 20 seconds to test initialization for me 🙁 

But I agree this is worth testing!

MatthewTighe:
> Tested this locally, and it's looking good!

Thank you!

We'll see if adding the test for the metric somehow solves the (seemingly) unrelated test failure that keeps popping up. Not really sure what's causing that. Have you seen that one happen before? I'll update this comment with the failure if it happens again. I've not been able to reproduce locally and I'm worried that it only occurs when the full suite is run

MatthewTighe:
Yeah, looks like the test failure is still happening for `TEST: WHEN record is called THEN telemetry is recorded with the appropriate label`.

I am not able to reproduce locally when I run it individually, but I'm running the full suite now to see if that triggers it.

app/src/main/java/org/mozilla/fenix/wallpapers/WallpapersUseCases.kt
```diff
@@ -0,0 +1,255 @@
...
+                downloadAllRemoteWallpapers(availableWallpapers)
+                store.dispatch(AppAction.WallpaperAction.UpdateAvailableWallpapers(availableWallpapers))
+                store.dispatch(AppAction.WallpaperAction.UpdateCurrentWallpaper(currentWallpaper))
+            }
+        }
+
+        private fun getAvailableWallpapers() = possibleWallpapers
+            .filter { !it.isExpired() && it.isAvailableInLocale() }
+
+        private suspend fun downloadAllRemoteWallpapers(availableWallpapers: List<Wallpaper>) {
```
MozillaNoah, line 118: why use a parameter + a loop + filter instead of just using `remoteWallpapers` in the companion object?

app/src/main/java/org/mozilla/fenix/wallpapers/WallpapersUseCases.kt
```diff
@@ -0,0 +1,255 @@
...
+         *
+         * @param wallpaper The wallpaper to load a bitmap for.
+         */
+        override suspend operator fun invoke(wallpaper: Wallpaper): Bitmap? = when (wallpaper) {
+            is Wallpaper.Local -> loadWallpaperFromDrawables(context, wallpaper)
+            is Wallpaper.Remote -> loadWallpaperFromDisk(context, wallpaper)
+            else -> null
+        }
+
+        private suspend fun loadWallpaperFromDrawables(
```
MozillaNoah, line 182:
```suggestion
        private suspend fun loadWallpaperFromDrawable(
```

MozillaNoah (approved):
LGTM pending a small discussion comment

app/src/main/java/org/mozilla/fenix/wallpapers/WallpapersUseCases.kt
MatthewTighe: To be able to manipulate the wallpaper list under test, so we can control for things like promotions and expirations. In theory (and as was the case in Turning Red), wallpapers may be locale-restricted and can also separately have their availability expire (this is planned for upcoming seasonal wallpapers). 

Explicitly defining them like they are in the companion object will be going away shortly in any case, as we will be fetching all the wallpapers from metadata stored remotely. The wallpaper type hierarchy may experience some changes as part of that as well.

app/src/test/java/org/mozilla/fenix/perf/StartupTypeTelemetryTest.kt
```diff
@@ -35,7 +35,7 @@ private val validTelemetryLabels = run {
 
 private val activityClass = HomeActivity::class.java
 
-@RunWith(FenixRobolectricTestRunner::class)
+@RunWith(AndroidJUnit4::class)
```
jonalmeida, line 38: @MatthewTighe @MozillaNoah I'm trying to look into https://github.com/mozilla-mobile/fenix/issues/26381 https://github.com/mozilla-mobile/fenix/issues/26380 and the last change that took place to this class happened in this PR. Looking at the diff here, there is no context on why the test runner was changed or why it was part of this wallpaper patch in particular. Could you provide some context here?

MatthewTighe: @jonalmeida Yeah, that is fully my fault. I started hitting the referenced error during CI and in a rush to try and land these wallpapers patches attempted a fix following some advice in [this comment chain](https://github.com/robolectric/robolectric/issues/5356). Clearly, it didn't actually resolve the issue and I recognize that this was a Not Cool Move on my part. I am happy to revert the change if we'd like

PR #26331, opened 2022-08-05
For #26329 Refactor the search widget to use the A-C common behaviour

### LINKED ISSUE #26329: Refactor the search widget to use the A-C common behaviour
This task should contain the implementation of search widget the Fenix part. I moved and refactored  the search widget from Fenix in AC . 
This has appear based on this https://github.com/mozilla-mobile/focus-android/issues/7249#issuecomment-1195533956
This is the task from AC https://github.com/mozilla-mobile/android-components/issues/12565 .
This is the task from Focus https://github.com/mozilla-mobile/focus-android/issues/7249

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21203)

iorgamgabriel:
It should behave like the current version .

### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation
Fixes #26329


Mugurell (approved):
LGTM.
Though nothing stands out I saw a case in which the widget would not show the app icon, don't know if it's a new issue / device issue, to be assessed by QA.

Mugurell:
> LGTM. Though nothing stands out I saw a case in which the widget would not show the app icon, don't know if it's a new issue / device issue, to be assessed by QA.

The issue reproduces on main also for any update of [`search_widget_info.xml`](https://github.com/mozilla-mobile/fenix/blob/0b0e31aae8ae2697b3f449327e85f9b206906787/app/src/main/res/xml/search_widget_info.xml) like this patch needs to do. Looks like this needs more investigations.

iorgamgabriel:
I can reproduce the issue with disappearing icons on Pixel 3XL .
If I install the old version from main , I added the search widget and after that I install this version from this branch both the app icon and the microphone icon will disappear . 
If I keep exactly the layout search_widget_info only the app icon is disappearing .
Until now I didn't find any fix . @Mugurell @jonalmeida

Mugurell:
This then seems like a framework / launcher issue.
If we can't update the configuration without users seeing a broken widget looks like Fenix shouldn't be refactored.
Curious if there are other reports online about this, seems like an important issue. if not - giving how easy it is to reproduce maybe we should at least report it to Google - [issuetracker.google.com](https://issuetracker.google.com)

mcarare:
> > LGTM. Though nothing stands out I saw a case in which the widget would not show the app icon, don't know if it's a new issue / device issue, to be assessed by QA.
> 
> The issue reproduces on main also for any update of [`search_widget_info.xml`](https://github.com/mozilla-mobile/fenix/blob/0b0e31aae8ae2697b3f449327e85f9b206906787/app/src/main/res/xml/search_widget_info.xml) like this patch needs to do. Looks like this needs more investigations.

We could send a one-time `APPWIDGET_UPDATE` as a workaround if we still want to go ahead and change something related to the widget. But maybe that should be a last resort solution.

Mugurell:
> We could send a one-time `APPWIDGET_UPDATE` as a workaround if we still want to go ahead and change something related to the widget. But maybe that should be a last resort solution.

Not sure when and where from if the issue reproduces immediately after updating the app but without the app having a chance to open. ?

mcarare:
For me, the bug reproduced only after opening the updated app.

iorgamgabriel:
For the bug with icons disappearing at update I made a fix (an APPWIDGET_UPDATE in FenixApplication only one time ) , but sometimes I saw that the icons have disappeared random . 
I don't have a scenario how to reproduce this bug and I don't know if this is happening in the old version of the app. 
@jonalmeida what is your opinion regarding this ? Do we merge the patch or not ?

jonalmeida:
@iorgamgabriel I would not land this Fenix if are unsure of what is causing the widget to be removed. For now, the original intention was to share the component in Focus, so we don't need to block that work on this PR. What do you think?

iorgamgabriel:
@jonalmeida It's ok . Thank you for the answer.

PR #26356, opened 2022-08-08
Add "manage account and devices" button to home menu

rxumoz:
Due to the regulations of app stores in China, we should provide a straightforward way for users to delete the accounts.

What I did in this patch is  just to add a button of managing account and devices in the home menu, to tap the button is just open the settings page of FxA in a new tab. 
 
Screenshot_20220812_125845_org mozilla fenix debu
.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/home/HomeMenu.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/HomeMenu.kt
@@ -19,10 +19,12 @@ import mozilla.components.browser.menu.item.BrowserMenuDivider
 import mozilla.components.browser.menu.item.BrowserMenuHighlightableItem
 import mozilla.components.browser.menu.item.BrowserMenuImageSwitch
 import mozilla.components.browser.menu.item.BrowserMenuImageText
+import mozilla.components.browser.menu.item.SimpleBrowserMenuItem
 import mozilla.components.concept.sync.AccountObserver
 import mozilla.components.concept.sync.AuthType
 import mozilla.components.concept.sync.OAuthAccount
 import mozilla.components.support.ktx.android.content.getColorFromAttr
+import org.mozilla.fenix.Config
 import org.mozilla.fenix.R
 import org.mozilla.fenix.components.accounts.AccountState
 import org.mozilla.fenix.components.accounts.FenixAccountManager
@@ -46,6 +48,10 @@ class HomeMenu(
         object Downloads : Item()
         object Extensions : Item()
         data class SyncAccount(val accountState: AccountState) : Item()
+        /**
+         * A button item to open up the settings page of FxA, shown up in mozilla online builds.
+         */
+        object ManageAccountAndDevices : Item()
         object WhatsNew : Item()
         object Help : Item()
         object CustomizeHome : Item()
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/mozilla/fenix/home/HomeMenu.kt
```diff
@@ -201,6 +210,7 @@ class HomeMenu(
             extensionsItem,
             syncSignInMenuItem,
             accountAuthItem,
+            manageAccountAndDevicesItem,
```
Mugurell, line 213: If I understand correctly this new menu option should only be shown in the Chinese version of the app.
To avoid it being shown in all apps (if not needed) we can use it as:

```suggestion
            if (Config.channel.isMozillaOnline) manageAccountAndDevicesItem else null,
```

app/src/main/java/org/mozilla/fenix/home/HomeMenuBuilder.kt
```diff
@@ -114,6 +115,18 @@ class HomeMenuBuilder(
                     }
                 )
             }
+            HomeMenu.Item.ManageAccountAndDevices -> {
+                homeActivity.openToBrowserAndLoad(
+                    searchTermOrURL =
+                    if (Config.channel.isMozillaOnline && context.settings().allowDomesticChinaFxaServer) {
```
Mugurell, line 121: If we use the `Config.channel.isMozillaOnline` check when we build the menu maybe here is not needed again.

detekt-baseline.xml
```diff
@@ -233,6 +233,7 @@
     <ID>UndocumentedPublicClass:HomeMenu.kt$HomeMenu.Item$ReconnectSync : Item</ID>
     <ID>UndocumentedPublicClass:HomeMenu.kt$HomeMenu.Item$Settings : Item</ID>
     <ID>UndocumentedPublicClass:HomeMenu.kt$HomeMenu.Item$SyncAccount : Item</ID>
+    <ID>UndocumentedPublicClass:HomeMenu.kt$HomeMenu.Item$ManageAccountAndDevices : Item</ID>
```
Mugurell, line 236: The baseline file should've just allowed time to add documentation but is currently abused with new ignores instead of adding the proper documentation.
Since the new menu item is new to the app and you know the context for it can you please add a small documentation for it?
I'll also open a ticket for working through the baseline files and clearing them instead of adding new ignores.

app/src/main/java/org/mozilla/fenix/home/HomeMenu.kt
Mugurell, line 213: Just to double check, with the current approach this new menu option would only appear when the user is on the app's homescreen, it would not also appear when browsing (which would need a similar declaration in `DefaultToolbarMenu`). Is this intended?

rxumoz: Yes we plan to make minimal changes, so just add this to home screen menu.

Mugurell (approved):
Looks great!

rxumoz:
@Mergifyio backport releases_v104.0.0

PR #26393, opened 2022-08-09
update isFirstRun to be manually set and set experiments locally on first run

jeddai:
No tests currently exist for the altered functionality. These changes should not affect the user interface.

Relevant JIRA issue: [EXP-2627](https://mozilla-hub.atlassian.net/browse/EXP-2627)

### Pull Request checklist

### QA

- [ ] **QA Needed**

### GitHub Automation

Used by GitHub Actions.


Amejia481:
> Thanks for the patch @jeddai! Could you please provide a bit more context around why those changes are required and linked to nimbus? Thanks in advance!

No need, for further context I got the main idea :)

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -347,18 +347,10 @@ class Settings(private val appContext: Context) : PreferencesHolder {
...
-                )
-                .apply()
-            true
-        } else {
-            false
-        }
+    val isFirstRun: Boolean by booleanPreference(
+        appContext.getPreferenceKey(R.string.pref_key_is_first_run),
+        default = true
+    )
```
Amejia481, line 353: This is just a refactor!

Amejia481: 👍🏽

app/src/main/java/org/mozilla/fenix/experiments/NimbusSetup.kt
```diff
@@ -101,7 +101,8 @@ fun createNimbus(context: Context, url: String?): NimbusApi {
                 globalUserParticipation = enabled
             }
 
-            if (url.isNullOrBlank()) {
+            if (context.settings().isFirstRun || url.isNullOrBlank()) {
```
Amejia481, line 104: Could this cause that we only add theses experiments for new users? As `isFirstRun` is persistent, user that had run the app before won't get the updates.

jeddai: That's the intention from what I understand -- the experiments that are labelled as "first run" are meant to be only for users that are actively onboarding to the application.

Amejia481: Ideally we would like for all users to get startup experiments and all subsequent updates of them.

Amejia481: I think onboard it's the main use case now, but I believe the idea is to be able to experiment on anything that is early on the startup.

Amejia481: Base on the existing PR, if we had mix configured something on those experiments and we would like to push an app update with the correct values, previous users won't get the fixed experiments?

jeddai: > Ideally we would like for all users to get startup experiments and all subsequent updates of them.

I may be misunderstanding what a first run experiment vs a startup experiment is. Can you elaborate on the difference between the two?

jeddai: If it helps at all, this is the epic for the ticket I'm working. https://mozilla-hub.atlassian.net/browse/EXP-2617

Amejia481: > I may be misunderstanding what a first run experiment vs a startup experiment is. Can you elaborate on the difference between the two?

I think you are on the right track with the first run experiments, just that I thought this mechanism could be useful for startup experiments, but chatting with other team members helped me to clarify that it may not be the case. **On the other hand, I think we shouldn't limit this to only new users  (`isFirstRun`), as in some cases we do big app version launches where we add a lot of functionalities and we want to do experimentation for the first time that users open that specific app version. For example, imagine that we are going to launch version 200 and we would like to have a welcome message on their first opening of version 200, with the actual approach we are going to be able to only experiment with new users.**

For some reference https://mozilla-hub.atlassian.net/browse/FNXV2-17820 we wanted to do something similar before but we couldn't. 

image

jeddai: Okay -- that generally makes sense. 

> with the actual approach we are going to be able to only experiment with new users.

From what I understand experiments currently cannot be distinguished between what you're describing as a "first run for this version" vs "first run of the app for this installation", there's only the latter. Is that an incorrect understanding?

Amejia481: Yeah, at the moment we don't distinguish. I don't think there is need for it :)

Amejia481: If everything is clear up, I think we should only have to close this PR :)
And merge https://github.com/mozilla-mobile/fenix/pull/26395

jeddai: For the future I think it would be a great addition, though! This might be oversimplifying things but I think a solution for that could be something along the lines of:
1. Add a new file `initial_version_run_experiments.json`
2. Add a step in the updating Nimbus experiments workflow that mimics the existing step but for the new file. It would need to have different query params on the `experimenter-url` to filter the experiments to only `first_version_run` experiments.
3. Add an additional step in the Nimbus setup to check the current version against a manually set pref for the version, thus allowing the app to know if it's the first run for that version. If it is, then it could apply the experiments in the `initial_version_run_experiments.json` file

jeddai: I am confused, why are you suggesting we close this PR?

Amejia481: > I am confused, why are you suggesting we close this PR?


Because without this PR at the moment, we are adding the `initial_experiments` either way, after we merge https://github.com/mozilla-mobile/fenix/pull/26395, we will have the functionality of bundling experiments with the app.

Amejia481 (approved):

PR #26417, opened 2022-08-10
Close #26198: Add debounce to account manager syncs

### LINKED ISSUE #26198: Every visit to the HomeFragment triggers a full account sync
In https://github.com/mozilla-mobile/fenix/pull/26028, we've added the method `collectAccountUpdates`. This method calls `accountManager.syncNow` and `accountManager.withConstellation { refreshDevices() }` and is called for every `onStart`/`onPause` of the fragment because it's a `LifecycleAwareFeature`.

This is problem because even opening the tabs tray and dismissing it (with the home fragment in the background triggers a sync which leads to multiple GET and POST requests:
Screen Shot 2022-07-26 at 10 05 38 PM

I would have expected that we allow our regular sync polling to update the SyncStore and this feature would observe the state and display the most recent synced tab that we already have. 

cc: @MatthewTighe

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21123)

rocketsroger:
### Pull Request checklist

### QA

- [ ] **QA Needed**

### GitHub Automation
Fixes #26198

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
+++ b/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
@@ -340,8 +340,13 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
             lifecycleScope.launch {
                 // If we're authenticated, kick-off a sync and a device state refresh.
                 components.backgroundServices.accountManager.authenticatedAccount()?.let {
+                    val syncReason = when (isVisuallyComplete) {
+                        true -> SyncReason.User
+                        false -> SyncReason.Startup
+                    }
+
                     components.backgroundServices.accountManager.syncNow(
-                        SyncReason.Startup,
+                        reason = syncReason,
                         debounce = true
                     )
                 }
--- a/app/src/main/java/org/mozilla/fenix/home/recentsyncedtabs/RecentSyncedTabFeature.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/recentsyncedtabs/RecentSyncedTabFeature.kt
@@ -60,7 +60,11 @@ class RecentSyncedTabFeature(
                     // Sync tabs storage will fail to retrieve tabs aren't refreshed, as that action
                     // is what populates the device constellation state
                     accountManager.withConstellation { refreshDevices() }
-                    accountManager.syncNow(SyncReason.User, customEngineSubset = listOf(SyncEngine.Tabs))
+                    accountManager.syncNow(
+                        reason = SyncReason.User,
+                        debounce = true,
+                        customEngineSubset = listOf(SyncEngine.Tabs),
+                    )
                 }
             }.launchIn(coroutineScope)
     }
--- a/app/src/test/java/org/mozilla/fenix/home/recentsyncedtabs/RecentSyncedTabFeatureTest.kt
+++ b/app/src/test/java/org/mozilla/fenix/home/recentsyncedtabs/RecentSyncedTabFeatureTest.kt
@@ -128,7 +128,7 @@ class RecentSyncedTabFeatureTest {
 
         verify { appStore.dispatch(AppAction.RecentSyncedTabStateChange(RecentSyncedTabState.Loading)) }
         coVerify { accountManager.withConstellation { refreshDevices() } }
-        coVerify { accountManager.syncNow(reason = SyncReason.User, debounce = false, customEngineSubset = listOf(SyncEngine.Tabs)) }
+        coVerify { accountManager.syncNow(reason = SyncReason.User, debounce = true, customEngineSubset = listOf(SyncEngine.Tabs)) }
     }
 
     @Test


app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -340,8 +340,13 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
             lifecycleScope.launch {
                 // If we're authenticated, kick-off a sync and a device state refresh.
                 components.backgroundServices.accountManager.authenticatedAccount()?.let {
+                    val syncReason = when (isVisuallyComplete) {
```
rocketsroger, line 343: I've also include this change so we use `SyncReason.Startup` if we haven't visually completed loading.  Otherwise use `SyncReason.User` since it's the user's navigation that's triggering the sync.

MatthewTighe (commented):
I have one outstanding question about the debounce mechanisms, but this otherwise looks good to me! It may be worth updating `DefaultController::syncAccount` for the `SyncedTabFeature` in A-C with the same change. There are a couple places in Fenix that still implement the Feature.

app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkController.kt
```diff
@@ -169,7 +169,10 @@ class DefaultBookmarkController(
     override fun handleRequestSync() {
         scope.launch {
             store.dispatch(BookmarkFragmentAction.StartSync)
-            activity.components.backgroundServices.accountManager.syncNow(SyncReason.User)
+            activity.components.backgroundServices.accountManager.syncNow(
```
MatthewTighe, line 172: I'm wondering if there will be problems when requests are debounced that have different subsets of engine requests. For example this request is a full sync, can it get overwritten by a tabs-only sync from elsewhere?

rocketsroger:
> I have one outstanding question about the debounce mechanisms, but this otherwise looks good to me! It may be worth updating `DefaultController::syncAccount` for the `SyncedTabFeature` in A-C with the same change. There are a couple places in Fenix that still implement the Feature.

That's a good idea.  I'll update that in A-C.  Thanks

app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkController.kt
rocketsroger: I'll look into this.  Thanks for pointing it out.

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
jonalmeida: Nice!

app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkController.kt
```diff
@@ -169,7 +169,10 @@ class DefaultBookmarkController(
...
+                reason = SyncReason.User,
+                debounce = true,
```
jonalmeida, line 174: I'm not sure if this is the right patch to be changing the syncing of bookmarks, but at the very least, if you want to include this patch, let's maybe take out the debounce and use `customEngineSubset = listOf(SyncEngine.Bookmarks)`?

The concern with the home screen is because it is a hot path for users and we are using more resources there as a result whenever they use that flow. We haven't evaluated if bookmarks/history has similar concerns.

app/src/main/java/org/mozilla/fenix/library/history/HistoryFragment.kt
```diff
@@ -370,7 +370,7 @@ class HistoryFragment : LibraryPageFragment<History>(), UserInteractionHandler {
     @Suppress("UnusedPrivateMember")
     private suspend fun syncHistory() {
         val accountManager = requireComponents.backgroundServices.accountManager
-        accountManager.syncNow(SyncReason.User)
+        accountManager.syncNow(reason = SyncReason.User, debounce = true)
```
jonalmeida, line 373: Same comment here as in bookmarks.

app/src/main/java/org/mozilla/fenix/settings/account/AccountSettingsFragment.kt
```diff
@@ -329,7 +332,10 @@ class AccountSettingsFragment : PreferenceFragmentCompat() {
         viewLifecycleOwner.lifecycleScope.launch {
             SyncAccount.syncNow.record(NoExtras())
             // Trigger a sync.
-            requireComponents.backgroundServices.accountManager.syncNow(SyncReason.User)
+            requireComponents.backgroundServices.accountManager.syncNow(
+                reason = SyncReason.User,
+                debounce = true,
```
jonalmeida, line 337: We need a full sync here on the account page - the user is explicitly asking to sync _now_.

app/src/main/java/org/mozilla/fenix/settings/account/AccountSettingsFragment.kt
```diff
@@ -250,7 +250,10 @@ class AccountSettingsFragment : PreferenceFragmentCompat() {
     private fun updateSyncEngineState(engine: SyncEngine, newValue: Boolean) {
         SyncEnginesStorage(requireContext()).setStatus(engine, newValue)
         viewLifecycleOwner.lifecycleScope.launch {
-            requireContext().components.backgroundServices.accountManager.syncNow(SyncReason.EngineChange)
+            requireContext().components.backgroundServices.accountManager.syncNow(
+                reason = SyncReason.EngineChange,
+                debounce = true,
```
jonalmeida, line 255: Same here as above - the user has explicitly changed the sync engines, so we shouldn't debounce here.

app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkController.kt
rocketsroger: I see.  Sure, I'll update and update to use the bookmark engine.

app/src/main/java/org/mozilla/fenix/library/history/HistoryFragment.kt
rocketsroger: Sure, I'll update to use the history engine.

app/src/main/java/org/mozilla/fenix/settings/account/AccountSettingsFragment.kt
rocketsroger: I'll remove.  Thanks

app/src/main/java/org/mozilla/fenix/settings/account/AccountSettingsFragment.kt
rocketsroger: I'll remove.  Thanks,

rocketsroger:
@jonalmeida After some thoughts I've decided to remove history and bookmark sync changes from this PR.  As a separate task I'll look at how we can improve A-C side before updating those sync calls.

jonalmeida (approved):
LGTM!

PR #26479, opened 2022-08-15
For #26423: simplify wallpaper types to single data class

### LINKED ISSUE #26423: Simplify wallpaper types and download metadata about them from the remote source.
We are planning to remove the wallpapers shipped with Firefox to our remote source. Because of that, there will no longer be a need to distinguish between local and remote wallpapers. The only special case will be the default wallpaper, which does not have an asset. 

AC:

- remove explicit references to individual wallpapers
- remove distinction between `Wallpaper.Local` and `Wallpaper.Remote`
- the `Expired` and `Promotional` types can also be removed and will be replaced by metadata
- download metadata as JSON from the remote source and unmarshal it into wallpaper objects directly
- the above may require slight updates to the wallpaper URL, as the metadata will be shared between iOS and Android and the current path references the `android` path specifically. this would require updating the secret in CI as well. it may also be possible to link the metadata into a child directory on the remote host, which would probably be more straightforward

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21266)

MatthewTighe:
This refactors the wallpaper types from a `sealed class` to a `data class`. This should make the type generally more extensible, and opens the door to fetching all required data about a wallpaper instance from our remote source. Moving forward, we won't be keeping metadata like the names of the wallpapers explicitly hard-coded and instead pull it from network. Splitting off this refactor seemed like a more manageable piece for review than doing everything at once, but I will have a follow-up available soon that adds pulling the metadata down from network. The work is mostly done, just need to rebase it over this patch.

### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation
Fixes #26423

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/wallpapers/Wallpaper.kt
+++ b/app/src/main/java/org/mozilla/fenix/wallpapers/Wallpaper.kt
@@ -4,88 +4,43 @@
 
 package org.mozilla.fenix.wallpapers
 
-import androidx.annotation.DrawableRes
-import java.util.Calendar
 import java.util.Date
 
 /**
- * Type hierarchy defining the various wallpapers that are available as home screen backgrounds.
+ * Type that represents wallpapers.
+ *
  * @property name The name of the wallpaper.
+ * @property collectionName The name of the collection the wallpaper belongs to.
+ * @property availableLocales The locales that this wallpaper is restricted to. If null, the wallpaper
+ * is not restricted.
+ * @property startDate The date the wallpaper becomes available in a promotion. If null, it is available
+ * from any date.
+ * @property endDate The date the wallpaper stops being available in a promotion. If null,
+ * the wallpaper will be available to any date.
  */
-sealed class Wallpaper {
-    abstract val name: String
-
-    /**
-     * The default wallpaper. This uses the standard color resource to as a background, instead of
-     * loading a bitmap.
-     */
-    object Default : Wallpaper() {
-        override val name = "default"
-    }
-
-    /**
-     * If a user had previously selected a wallpaper, they are allowed to retain it even if
-     * the wallpaper is otherwise expired. This type exists as a wrapper around that current
-     * wallpaper.
-     */
-    data class Expired(override val name: String) : Wallpaper()
-
-    /**
-     * Wallpapers that are included directly in the shipped APK.
-     *
-     * @property drawableId The drawable bitmap used as the background.
-     */
-    sealed class Local : Wallpaper() {
-        abstract val drawableId: Int
-        data class Firefox(override val name: String, @DrawableRes override val drawableId: Int) : Local()
-    }
-
-    /**
-     * Wallpapers that need to be fetched from a network resource.
-     *
-     * @property expirationDate Optional date at which this wallpaper should no longer be available.
-     */
-    sealed class Remote : Wallpaper() {
-        abstract val expirationDate: Date?
-        abstract val remoteParentDirName: String
-        @Suppress("MagicNumber")
-        /**
-         * A promotional partnered wallpaper.
-         */
-        data class House(
-            override val name: String,
-            override val expirationDate: Date? = Calendar.getInstance().run {
-                set(2022, Calendar.APRIL, 30)
-                time
-            }
-        ) : Remote(), Promotional {
-            override val remoteParentDirName: String = "house"
-            override fun isAvailableInLocale(locale: String): Boolean =
-                listOf("en-US", "es-US").contains(locale)
-        }
-
-        /**
-         * Wallpapers that are original Firefox creations.
-         */
-        data class Firefox(
-            override val name: String,
-        ) : Remote() {
-            override val expirationDate: Date? = null
-            override val remoteParentDirName: String = "firefox"
-        }
-    }
-
-    /**
-     * Designates whether a wallpaper is part of a promotion that is locale-restricted.
-     */
-    interface Promotional {
-        /**
-         * Returns whether the wallpaper is available in [locale] or not.
-         */
-        fun isAvailableInLocale(locale: String): Boolean
-    }
-
+data class Wallpaper(
+    val name: String,
+    val collectionName: String,
+    val availableLocales: List<String>?,
+    val startDate: Date?,
+    val endDate: Date?
+) {
     companion object {
+        const val amethystName = "amethyst"
+        const val ceruleanName = "cerulean"
+        const val sunriseName = "sunrise"
+        const val twilightHillsName = "twilight-hills"
+        const val beachVibeName = "beach-vibe"
+        const val firefoxCollectionName = "firefox"
+        const val defaultName = "default"
+        val Default = Wallpaper(
+            name = defaultName,
+            collectionName = defaultName,
+            availableLocales = null,
+            startDate = null,
+            endDate = null,
+        )
+
         /**
          * Defines the standard path at which a wallpaper resource is kept on disk.
          *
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/mozilla/fenix/wallpapers/WallpapersUseCases.kt
```diff
@@ -96,58 +96,78 @@ class WallpapersUseCases(
...
-        private fun getAvailableWallpapers() = possibleWallpapers
-            .filter { !it.isExpired() && it.isAvailableInLocale() }
+        private fun List<Wallpaper>.getAvailableWallpapers() =
+            this.filter { !it.isExpired() && it.isAvailableInLocale() }
 
         private suspend fun downloadAllRemoteWallpapers(availableWallpapers: List<Wallpaper>) {
-            for (wallpaper in availableWallpapers.filterIsInstance<Wallpaper.Remote>()) {
+            for (wallpaper in availableWallpapers) {
+                if (wallpaper == Wallpaper.Default) continue
                 downloader.downloadWallpaper(wallpaper)
```
MozillaNoah, line 121: personal nit: sometimes the `continue` keyword can be make a logic flow a little ambiguous; I usually prefer to avoid it, if possible. 
```suggestion
                if (wallpaper != Wallpaper.Default) {
                    downloader.downloadWallpaper(wallpaper)
                }
```

app/src/main/java/org/mozilla/fenix/wallpapers/WallpapersUseCases.kt
```diff
@@ -96,58 +96,78 @@ class WallpapersUseCases(
             // This should be cleaned up as improvements are made to the storage, file management,
             // and download utilities.
             withContext(Dispatchers.IO) {
-                val availableWallpapers = getAvailableWallpapers()
+                val availableWallpapers = possibleWallpapers.getAvailableWallpapers()
```
MozillaNoah, line 99: follow-up nit: `getAvailableWallpapers` could be a good candidate for a unit test

app/src/main/java/org/mozilla/fenix/wallpapers/WallpapersUseCases.kt
MozillaNoah: What would hypothetically happen here if `Wallpaper.Default` was passed into `downloader.downloadWallpaper`

app/src/main/java/org/mozilla/fenix/wallpapers/WallpapersUseCases.kt
```diff
@@ -96,58 +96,78 @@ class WallpapersUseCases(
...
+            this.availableLocales?.contains(currentLocale) ?: true
 
         companion object {
-            private val localWallpapers: List<Wallpaper.Local> = listOf(
-                Wallpaper.Local.Firefox("amethyst", R.drawable.amethyst),
-                Wallpaper.Local.Firefox("cerulean", R.drawable.cerulean),
-                Wallpaper.Local.Firefox("sunrise", R.drawable.sunrise),
+            private val localWallpapers: List<Wallpaper> = listOf(
+                Wallpaper(
+                    "amethyst",
```
MozillaNoah, line 136: nit: missed the named parameter for `name` for amethyst and cerulean and beach-vibe

app/src/main/java/org/mozilla/fenix/wallpapers/WallpapersUseCases.kt
```diff
@@ -173,24 +193,30 @@ class WallpapersUseCases(
...
             else -> null
         }
 
         private suspend fun loadWallpaperFromDrawable(
             context: Context,
-            wallpaper: Wallpaper.Local
+            wallpaper: Wallpaper
         ): Bitmap? = Result.runCatching {
+            val drawableRes = when (wallpaper.name) {
+                "amethyst" -> R.drawable.amethyst
```
MozillaNoah, line 207: I noticed this is the third time these wallpaper names are being referenced via direct strings. Would it make sense to make them a const val or a non-translatable string?

MozillaNoah (commented):
initial pass of comments and nits while I wait for it to build locally.

app/src/main/java/org/mozilla/fenix/wallpapers/WallpapersUseCases.kt
gabrielluong: I would generally recommend this as well. Less chances of failure because we misspell a wallpaper name in various usage.

app/src/main/java/org/mozilla/fenix/wallpapers/Wallpaper.kt
```diff
@@ -4,88 +4,36 @@
...
-         */
-        fun isAvailableInLocale(locale: String): Boolean
-    }
-
+data class Wallpaper(
+    val name: String,
+    val collectionName: String,
+    val availableLocales: List<String>?,
+    val startDate: Date?,
+    val endDate: Date?
```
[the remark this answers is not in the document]
MatthewTighe: I had planned to add it along with the PR that actually pulls the metadata down. This PR was intended purely as a refactoring so does not add any new behavior/data

app/src/main/java/org/mozilla/fenix/wallpapers/WallpapersUseCases.kt
MatthewTighe: I can dig your syntax suggestion, will update.

I had thought that the downloader would error if it couldn't find a file (which is wouldn't be able to for the default), but it looks like it just silently fails. I would be okay with reverting that specific line as well, but some of the logic here will probably be reworking anyway once the `possibleWallpapers` are all being fetched remotely.

app/src/main/java/org/mozilla/fenix/wallpapers/WallpapersUseCases.kt
MatthewTighe: I personally prefer to avoid exposing functions individually for unit-testing. My philosophy is more along the lines of "you should unit-test the smallest public API". The functionality of that method is under test in several of the tests in `WallpaperUseCasesTest` already, like 
- `GIVEN wallpapers that expired WHEN invoking initialize use case THEN expired wallpapers are filtered out and cleaned up`
- `GIVEN wallpapers that expired and an expired one is selected WHEN invoking initialize use case THEN selected wallpaper is not filtered out`
- `GIVEN wallpapers that are in promotions outside of locale WHEN invoking initialize use case THEN promotional wallpapers are filtered out`

app/src/main/java/org/mozilla/fenix/wallpapers/WallpapersUseCases.kt
MatthewTighe: ty!

app/src/main/java/org/mozilla/fenix/wallpapers/WallpapersUseCases.kt
MatthewTighe: Oh yeah, that's a great idea. Since I had been planning to remove them immediately afterwards I was being a little more gung-ho, but I think it's still worth doing here.

app/src/main/java/org/mozilla/fenix/wallpapers/WallpapersUseCases.kt
MozillaNoah: I see the logic in this approach. As long as `getAvailableWallpapers` is only used in this one location, I agree it can stay as-is with your aforementioned tests.

MozillaNoah (approved):

Issue #26050, opened 2022-07-15
[Bug]: FxA account menu item doesn't always reflect latest account state


MatthewTighe:
The option in the toolbar three dot menu doesn't always accurately reflect the latest account state. Now that https://github.com/mozilla-mobile/fenix/pull/25956 has introduced the `SyncStore`, the account property should be observable in order to make sure an accurate state is always reflects.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21028)

jonalmeida:
We have another issue that is actively worked on which this is similar to.

Closing as a duplicate of https://github.com/mozilla-mobile/fenix/issues/25486

Issue #25486, opened 2022-06-01
App menu shows "Sign in to sync" menu item until I tap it, even though I'm already signed in to sync


cpeterso:
### Steps to reproduce

1. Log into your FxA Sync in Fenix.
2. Kill Fenix.
3. Launch Fenix.
4. Open the app menu.
5. See "Sign in to sync". I think this is a bug!
6. Tap the "Settings" menu item.
7. See at the top of the Settings screen that you are already signed in.
8. Exit the Settings screen.
9. Open the app menu again.
10. See that the "Sign in to sync" menu item has been replaced with your sync account name. This is the result I expected in step 5 above.

Maybe I am experiencing https://github.com/mozilla-mobile/fenix/issues/24112? But I don't think my account is in a bad authentication state.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-20628)

gabrielluong:
We probably have multiple duplicate issues related to this https://github.com/mozilla-mobile/fenix/issues/19657 and an existing PR that was approved but never landed https://github.com/mozilla-mobile/fenix/pull/21488

Alexandru2909:
While investigating this I noticed that in the current implementation of `HomeMenu`, when the `onAuthenticated` callback is invoked, we update the menu builder, thus [dismissing](https://github.com/mozilla-mobile/android-components/blob/d18be7813ece1c2abd837970fad72cd744893ed3/components/browser/menu/src/main/java/mozilla/components/browser/menu/view/MenuButton.kt#L101) the menu, if it is being displayed at that moment. Updating the menu does not also update the `syncItem`, as it is [stored as a class property](https://github.com/mozilla-mobile/fenix/blob/f1e7b49425bf2a419fb7997c0d29739f13f3626e/app/src/main/java/org/mozilla/fenix/home/HomeMenu.kt#L96-L102). 

This behavior is fixed in the contributor's PR, however the menu dismissing is something we may not want to add.

https://user-images.githubusercontent.com/35462038/183077411-a2cb740d-35f8-4149-8b6c-3ec35ac66ccc.mp4

https://user-images.githubusercontent.com/35462038/183077459-395a1512-0a59-49b3-9868-161b8cf582d7.mp4

Since the authentication state takes some time to be updated and notify the observers about it, the `Sign in to sync` item is displayed for a few seconds before the menu is dismissed.

[...]

PR #26494, opened 2022-08-16
Close #26072: Don't show sync menu item until account manager is available

### LINKED ISSUE #26072: Sign in to sync toolbar menu item, indicates I'm logout when I'm not
### Steps to reproduce

Prerequisites authenticate with your Firefox account. 

1. Kill Nightly.
2. Open Nightly.
3. Tab the 3 dotted menu. 

### Expected behaviour

The menu item should indicate that I'm authenticated 

### Actual behaviour

The menu item indicates I'm not authenticated 

### Device name

_No response_

### Android version

Android 12

### Firefox release type

Firefox Nightly

### Firefox version

104.0a1

### Device logs

_No response_

### Additional information

_No response_

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21040)

rocketsroger:
### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation
Fixes #26072


MozillaNoah:
I can unfortunately still reproduce the bug :(. 

It also looks like this bug is being tracked in: 
- https://github.com/mozilla-mobile/fenix/issues/25486
- https://github.com/mozilla-mobile/fenix/issues/26050

rocketsroger:
> I can unfortunately still reproduce the bug :(
> 
> It also looks like this bug is being tracked with some discussion in:
> 
>     * [App menu shows "Sign in to sync" menu item until I tap it, even though I'm already signed in to sync #25486](https://github.com/mozilla-mobile/fenix/issues/25486)
> 
>     * [[Bug]: FxA account menu item doesn't always reflect latest account state #26050](https://github.com/mozilla-mobile/fenix/issues/26050)

This can only be reproduced when it's a fresh start with no tabs opened.  Signed into account, close all tabs, swipe away Fenix (Force stop)  then start Fenix again.  Once started (should be showing home) tap on menu.

MozillaNoah:
> > I can unfortunately still reproduce the bug :(
> > It also looks like this bug is being tracked with some discussion in:
> > ```
> > * [App menu shows "Sign in to sync" menu item until I tap it, even though I'm already signed in to sync #25486](https://github.com/mozilla-mobile/fenix/issues/25486)
> > 
> > * [[Bug]: FxA account menu item doesn't always reflect latest account state #26050](https://github.com/mozilla-mobile/fenix/issues/26050)
> > ```
> 
> This can only be reproduced when it's a fresh start with no tabs opened. Signed into account, close all tabs, swipe away Fenix (Force stop) then start Fenix again. Once started (should be showing home) tap on menu.

For some clarification, are you meaning that this fix is _just_ for the bug while on Home and not any first app open?

rocketsroger:
> For some clarification, are you meaning that this fix is _just_ for the bug while on Home and not any first app open?

Yes, `HomeMenu` is the one we use in Home.  When opening a session we use the `DefaultToolbarMenu`.

rocketsroger:
Just thought of another use case that can easily produce this issue.  Restart Fenix in private mode will also consistently reproduce this problem since we don't restore session in private mode so we always shows home.

Mugurell:
This seems to just move the issue elsewhere.
If we delay constructing the menu until some later time then users won't be able to interact with it until that later time. And this would affect all users, logged in or not.

https://user-images.githubusercontent.com/11428869/185069343-ce24b59c-fccc-4114-b9ef-653e3f086986.mp4

rocketsroger:
> This seems to just move the issue elsewhere. If we delay constructing the menu until some later time then users won't be able to interact with it until that later time. There is also a visible difference. And this would affect all users, logged in or not.

Hmm, from my testing I never seen that.  Code wise I knew that could happen but account manager was always ready before we finish drawing everything.  We could update the home menu after but it's weird to create the menu twice when we usually just need it to be created when we have updated information.

rocketsroger:
From my testing when the user is not signed in the account manager is ready much faster as well.

rocketsroger:
Here's what I got on my Pixel 3 XL with no account signed in

rocketsroger:
I'll see if I can find a better solution today.

rocketsroger:
@Mugurell After some thoughts, to fix the issue without any change in behaviour, I've just added an extra build after the account is ready.  If the account manager is already ready then we don't do the extra rebuild.  Not sure if this is a good solution.

Mugurell:
What I'm now seeing is that the menu initially opens with "Sign in to sync" and when the account manager is available the menu is automatically closed.

Result on a OnePlus7 with Android 11:

https://user-images.githubusercontent.com/11428869/185310214-20d4909b-3de6-46c5-8017-aac886eef63a.mp4

Given that the menu width is dynamic based on the widest element tinkering with it to switch to showing the email address when available and replacing "Sign in to sync" would probably lead to other issues: either the menu will keep the same width and the email will be ellipsized, either the menu will be invalidated and it would get all of a sudden wider to allow for showing the entire email.

To me this seems like a scenario that should be resolved by UX by maybe introducing another string beside "Sign in to sync" to express that connecting to FXA is currently in progress and the app does not know at that time whether the user is authenticated or not - similar to how in other cases when we have to fetch data we have a way of indicating progress for the meantime.
Saw a discussion about this and the UX being tagged in https://github.com/mozilla-mobile/fenix/issues/25486#issuecomment-1206411322.

rocketsroger:
> What I'm now seeing is that the menu initially opens with "Sign in to sync" and when the account manager is available the menu is automatically closed.

I understand the concern.  But I think this issue applies to users who are signed into their account and tap on the menu right after a fresh start.  I'll create an issue to track removing this when we have a better solution.  But I think this is a small price to pay so the user is frequently confused that they've been signed out.

rocketsroger:
Created the issue to track removing the workaround https://github.com/mozilla-mobile/fenix/issues/26526.  Updated the patch to document it.

gabrielluong:
Can you check if this is related https://github.com/mozilla-mobile/fenix/pull/21488?

rocketsroger:
> Can you check if this is related #21488?

I'll look into it.  Thanks

rocketsroger:
@Mugurell, I took the suggestion from @mcarare to hide the menu item until it's ready.  This seems like the best middle ground solution.  Please let me know if this is acceptable.  Thanks

app/src/main/java/org/mozilla/fenix/home/HomeMenu.kt
```diff
@@ -96,15 +96,26 @@ class HomeMenu(
...
-        R.drawable.ic_synced_tabs,
-        primaryTextColor
-    ) {
-        onItemTapped.invoke(Item.SyncAccount(accountManager.accountState))
+        return when (syncItemTitle) {
+            null -> null
+            else -> {
+                BrowserMenuImageText(
+                    syncItemTitle,
+                    R.drawable.ic_signed_out,
```
rocketsroger, line 112: I noticed that the icon doesn't match the toolbar menu. I've updated the icon to ic_signed_out here but would require confirmation from UX.

rocketsroger:
account manager not ready (left) and account manager ready (right)
Screenshot_20220818_170150

Mugurell:
> @Mugurell, I took the suggestion from @mcarare to hide the menu item until it's ready. This seems like the best middle ground solution. Please let me know if this is acceptable. Thanks

Agree that this seems like the best middle ground. Don't see it causing other issues. 👍

Mugurell (approved):

rocketsroger:
Thanks! I'll land this after the cut.

Issue #26497, opened 2022-08-16
Create updated wallpaper downloader and filemanager


MatthewTighe:
Create an updated wallpaper downloader and filemanager that use the updated directory paths in the remote source, and reflect those same paths locally on the device.

Behind the feature flag from: https://github.com/mozilla-mobile/fenix/issues/26517

- Rename current downloader and file manager as `Legacy`
- Create a new `WallpaperDownloader`
- Create a new `WallpaperFileManager`
- Both should use paths that follow the pattern `<wallpaperName>/<type of image>.png` instead of `<resolution>/<orientation>/<theme>/<wallpaperName.png>`
- Add new default use cases that use them, such that wallpapers can be downloaded and displayed correctly
- Switch which use case is used based on the feature flag

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21313)

PR #26499, opened 2022-08-17
For #26423: add wallpaper metadata fetcher

### LINKED ISSUE #26423: Simplify wallpaper types and download metadata about them from the remote source.

MatthewTighe:
The wallpaper metadata fetcher is currently unused, but will be used after https://github.com/mozilla-mobile/fenix/issues/26497 introduces an updated `InitalizeWallpaperUseCase`. I have tested it locally and confirmed that it can successfully download the correct wallpaper metadata.

### Pull Request checklist

### QA

- [ ] **QA Needed**

### GitHub Automation
Fixes #26423

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperMetadataFetcher.kt
+++ b/app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperMetadataFetcher.kt
@@ -1,0 +1,101 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.wallpapers
+
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
+import mozilla.components.concept.fetch.Client
+import mozilla.components.concept.fetch.Request
+import org.json.JSONArray
+import org.json.JSONObject
+import org.mozilla.fenix.BuildConfig
+import java.text.SimpleDateFormat
+import java.util.Date
+import java.util.Locale
+
+/**
+ * Utility class for downloading wallpaper metadata from the remote server.
+ *
+ * @property client The client that will be used to fetch metadata.
+ */
+class WallpaperMetadataFetcher(
+    private val client: Client
+) {
+    private val metadataUrl = BuildConfig.WALLPAPER_URL.substringBefore("android") +
+        "metadata/v$currentJsonVersion/wallpapers.json"
+
+    /**
+     * Downloads the list of wallpapers from the remote source. Failures will return an empty list.
+     */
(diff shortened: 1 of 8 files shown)


app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperMetadataFetcher.kt
```diff
@@ -0,0 +1,95 @@
...
+        )
+        return getJSONArray("wallpapers").toWallpaperList(collection)
+    }
+
+    private fun JSONArray.getAvailableLocales(): List<String>? =
+        (0 until length()).map { getString(it) }
+
+    private fun JSONObject.getAvailabilityRange(): Pair<Date, Date>? {
+        val formatter = SimpleDateFormat("yyyy-MM-dd", Locale.US)
+        return formatter.parse(getString("start"))!! to formatter.parse(getString("end"))!!
```
Amejia481, line 69: Could we try to avoid using `!!` as we could cause NPE? :)

Amejia481 (approved):
✅ 

Are we going to create some of task to periodicity try to see if there are new updates? :)

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperMetadataFetcher.kt
MatthewTighe: Yeah, I can update it here. Since the top-level function is wrapped in a `Result.runCatching`, throwing NPEs was actually by design in order to quickly cause failures. Looking back at it, I think it's bad to have that behavior implied and spread over the two sites, and runs a risk of causing errors if the `Result.runCatching` was ever removed.

MatthewTighe:
> Are we going to create some of task to periodicity try to see if there are new updates? :)
There will be a `last-updated` property for the metadata for the exact reason that we could do that as a follow up to the MR. I don't think it's necessary for the MR release, however, since we don't have any additional wallpapers planned yet

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperMetadataFetcher.kt
Amejia481: Thanks for the clarification, I missed the `runCatching`

Issue #26399, opened 2022-08-09
Recent synced tabs success states should include a list of tabs.


MatthewTighe:
In order to achieve the behavior required in https://github.com/mozilla-mobile/fenix/issues/26398, the `RecentSyncedTabFeature` should be changed to update the AppStore with a list of recent tabs. This list should be ordered by the [lastUsed](https://searchfox.org/mozilla-mobile/rev/a7d1e928df75a76af53846f8bd90e1ce726dcbae/android-components/components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/RemoteTabsStorage.kt#111) property of the tab. 

Note: there appears to be a bug in the current implementation when choosing the most recent tab. `device.lastAccessTime` does not seem to get regularly updated during long-running sessions. This is an issue, for example, with desktop devices where the browser is not often closed. 

Ordering this list as suggested above should resolve that issue as well.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21246)

Alexandru2909:
@ QA: We should ensure that the after this refactoring the Recent synced tabs feature functions the same.

LaurentiuApahideanSV:
I tested the issue on Fenix Nightly 2022-08-30 and the recent synced tabs feature works as intended.
### Devices used:
- Google Pixel 6 (Android 13)
- Samsung Galaxy S22 Ultra (Android 12)

Issue #26398, opened 2022-08-09
Tab pickup should respect the home screen blocklist.


MatthewTighe:
The tab pickup section of Jump Back In (the recent synced tab) should make use of the home screen blocklist and prevent rendering tabs that have been blocked. This should be done by intercepting `AppAction.RecentSyncedTabStateChange` in `BlocklistMiddleware` and filtering the results based on the blocklist. 

In addition, the tab pickup section should then switch to using the next non-blocked tab from the recent synced tab list.

Dependent on: https://github.com/mozilla-mobile/fenix/issues/26399

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21245)

LaurentiuApahideanSV:
I tested the issue by performing the following steps:
1. On the first device login with a Firefox account;
2. On the first device access any website;
3. From the Jump back in, Recent bookmarks and Recently visited sections long tap the website and select the remove option;
4. On a second device log in the same Firefox account from the first device;
5. Access same website from step 2;
6. Perform a sync on the second device;
7. Perform a sync on the first device;

After performing these steps the blocklist website did not appear in the sync tab tray, but the next non blocklist site.
### Devices used:
- Google Pixel 6 (Android 13
- Samsung Galaxy S22 (Android 12)

PR #26509, opened 2022-08-17
For #26400 - Add long-press option to remove tab pickup from homescreen

### LINKED ISSUE #26400: Add a "Remove" long-press option to tab pickup that adds the tab to the homescreen blocklist
The tab pickup section of "Jump Back In" should have the same removal behavior as the recent tab section. As a user, I should be able to:

1. Long-press the synced tab card to bring up a drop-down menu with a "Remove" option
2. Press the remove button
3. Have the url be added to the blocklist
4. Have every section of the home screen that uses the blocklist remove the item

This is dependent on: https://github.com/mozilla-mobile/fenix/issues/26398, which will add the behavior of updating the tab pickup with the most recent, unblocked tab. This issue is specifically about adding the UI and wiring it up as an action.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21247)

Alexandru2909:
Added option to add recent synced tab to homescreen blocklist by long press.

https://user-images.githubusercontent.com/35462038/185149620-2f1416ae-f337-48a3-b27d-04dbe1420c70.mp4

Depends on the work from #26399 and  #26398

### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation
Fixes #26400

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/home/blocklist/BlocklistHandler.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/blocklist/BlocklistHandler.kt
@@ -7,7 +7,7 @@ package org.mozilla.fenix.home.blocklist
 import androidx.annotation.VisibleForTesting
 import mozilla.components.support.ktx.kotlin.sha1
 import org.mozilla.fenix.home.recentbookmarks.RecentBookmark
-import org.mozilla.fenix.home.recentsyncedtabs.RecentSyncedTab
+import org.mozilla.fenix.home.recentsyncedtabs.RecentSyncedTabState
 import org.mozilla.fenix.home.recenttabs.RecentTab
 import org.mozilla.fenix.home.recentvisits.RecentlyVisitedItem
 import org.mozilla.fenix.utils.Settings
@@ -53,6 +53,28 @@ class BlocklistHandler(private val settings: Settings) {
         }
 
     /**
+     * If the state is set to [RecentSyncedTabState.Success], filter the list of recently synced
+     * tabs by the blocklist. If the filtered list of tabs is empty, change the state to
+     * [RecentSyncedTabState.None]
+     */
+    @JvmName("filterRecentSyncedTabState")
+    fun RecentSyncedTabState.filteredByBlocklist() =
+        if (this is RecentSyncedTabState.Success) {
+            val filteredTabs = settings.homescreenBlocklist.let { blocklist ->
+                this.tabs.filterNot {
+                    blocklistContainsUrl(blocklist, it.url)
+                }
+            }
+            if (filteredTabs.isEmpty()) {
+                RecentSyncedTabState.None
+            } else {
+                RecentSyncedTabState.Success(filteredTabs)
+            }
+        } else {
+            this
+        }
+
+    /**
      * Filter a list of recent history items by the blocklist. Requires this class to be contextually
      * in a scope.
      */
@@ -62,18 +84,6 @@ class BlocklistHandler(private val settings: Settings) {
             filterNot {
                 it is RecentlyVisitedItem.RecentHistoryHighlight &&
                     blocklistContainsUrl(blocklist, it.url)
-            }
-        }
-
-    /**
-     * Filter a list of recently synced tabs by the blocklist. Requires this class to be contextually
-     * in a scope.
-     */
-    @JvmName("filterRecentSyncedTab")
-    fun List<RecentSyncedTab>.filteredByBlocklist(): List<RecentSyncedTab> =
-        settings.homescreenBlocklist.let { blocklist ->
-            filterNot {
-                blocklistContainsUrl(blocklist, it.url)
             }
         }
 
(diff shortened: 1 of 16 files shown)


MozillaNoah:
Small nit: could the commit instead be:
`Add long-press option to remove tab pickup on homescreen`

"Add remove..." sounded confusing haha

MozillaNoah:
Looking at the other two commits, does this branch need to be rebased with `main`?

app/src/main/java/org/mozilla/fenix/home/blocklist/BlocklistMiddleware.kt
```diff
@@ -102,7 +118,18 @@ class BlocklistMiddleware(
...
+                    is RecentSyncedTabState.Success -> {
+                        val filteredTabs = recentSyncedTabState.tabs.filteredByBlocklist()
+                        if (filteredTabs.isEmpty()) {
+                            RecentSyncedTabState.None
+                        } else {
+                            RecentSyncedTabState.Success(filteredTabs)
+                        }
+                    }
+                    else -> recentSyncedTabState
+                }
```
MatthewTighe, line 132: It looks like the logic here could potentially be re-used in the `AppAction.Change` block as well. I think it would be optimal for readability if it were possible to wrap this in an extension function that follows the `filteredByBlocklist` convention. Are there cases that we wouldn't want to switch the `RecentSyncedTabState` to `None` when there were no tabs after filtering?

app/src/test/java/org/mozilla/fenix/home/blocklist/BlocklistMiddlewareTest.kt
```diff
@@ -341,4 +350,55 @@ class BlocklistMiddlewareTest {
...
+                    listOf(blockedTab)
+                )
+            )
+        ).joinBlocking()
+
+        assertEquals(
+            RecentSyncedTabState.None,
+            store.state.recentSyncedTabState
+        )
+    }
```
MatthewTighe, line 403: These look good! I would prefer to also include a test that captures the behavior of switching to the next recent synced tab when one is blocked, as well.

MatthewTighe:
This is definitely close, thanks for including the video. Just had a couple small comments

Alexandru2909:
> Looking at the other two commits, does this branch need to be rebased with `main`?

This is because this is based on the PRs for #26399 and #26398. I updated this PR's description with this.

app/src/main/java/org/mozilla/fenix/home/recentsyncedtabs/RecentSyncedTabFeature.kt
```diff
@@ -218,6 +219,20 @@ sealed class RecentSyncedTabState {
...
+            if (filteredTabs.isEmpty()) {
+                RecentSyncedTabState.None
+            } else {
+                RecentSyncedTabState.Success(filteredTabs)
+            }
+        }
+    } else {
+        this
+    }
+
```
MatthewTighe, line 235: Would it make sense to move the content of this function into the blocklist handler, such that we had `RecentSyncedTabState.filteredByBlocklist` instead of `List<RecentSyncedTab>.filteredByBlocklist`? 

I think that would allow us to have the call sites mirror the rest of the `filteredByBlocklist` calls more closely.

app/src/main/java/org/mozilla/fenix/home/blocklist/BlocklistHandler.kt
```diff
@@ -51,6 +52,32 @@ class BlocklistHandler(private val settings: Settings) {
...
 
+    /**
+     * If the state is set to [RecentSyncedTabState.Success], filter the list of recently synced
+     * tabs by the blocklist. If the filtered list of tabs is empty, change the state to
+     * [RecentSyncedTabState.None]
+     *
+     * @param blocklistHandler The blocklist handler used to filter the recent synced tabs.
+     */
+    @JvmName("filterRecentSyncedTabState")
+    internal fun RecentSyncedTabState.filteredByBlocklist(blocklistHandler: BlocklistHandler) =
```
MatthewTighe, line 63: I don't think this function needs to be marked as `internal` still. If we did ever modularize Fenix, I would expect this function to be an accessible part of the API.

MatthewTighe, line 63: Now that this function has been moved into this class, the `blocklistHandler` parameter can be removed. Defining the extension function within the class requires a class instance to be in scope to call the function. This can be more clearly demonstrated by checking that `this@filteredByBlocklist` and `this@BlocklistHandler` are both in scope in this function. 

That's why all the `filteredByBlocklist` calls are in `with (blocklistHandler)` blocks. I think doing so here will help keep this in convention with the rest of the file.

This is a stylistic thing I've picked up from looking into the source of Jetpack Compose. For example, some `Modifier` extension functions require certain scopes like `RowScope`. [Here is an example](https://developer.android.com/reference/kotlin/androidx/compose/foundation/layout/RowScope#(androidx.compose.ui.Modifier).align(androidx.compose.ui.Alignment.Vertical)) 

Kotlin 1.6.20 introduced a prototype of [context receivers](https://kotlinlang.org/docs/whatsnew1620.html#prototype-of-context-receivers-for-kotlin-jvm) that will allow us to use this pattern outside of the actual class definitions as well. Then this same pattern can be applied to code we don't own like library code.

MatthewTighe (approved):
Just one last round of feedback, thanks for patience on this! Otherwise LGTM, so I'll stamp it now in case you'd like to merge it during your next day. Appreciate you knocking this feature out 😄

app/src/main/java/org/mozilla/fenix/home/blocklist/BlocklistHandler.kt
Alexandru2909: Thanks for the thorough explanation!

Issue #26511, opened 2022-08-17
[Bug]: Wallpaper loading on launch is delayed


delia-pop:
### Steps to reproduce

1. Install the latest Nightly 105.0a1 from 08/17 or Update the app to the latest version and perform a clear data.
2. From Settings > Homepage > Opening screen select "Homepage".
3. From Settings > Homepage > Wallpapers select any wallpaper.
4. Close and Re-open Fenix. 
5. On start-up, observe the wallpaper on the homepage.

### Expected behaviour

The wallpaper loads without visual issues or significant delay, given a good and stable network connection. 

### Actual behaviour

The wallpaper loading on homepage is delayed with 1-3 seconds when launching the app, even with a good and stable network connection.
The issue occurs after every restart, not only the first launch. 

### Device name

Xiaomi 12 Pro

### Android version

Android 12

### Firefox release type

Firefox Nightly

### Firefox version

latest Nightly 105.0a1 from 08/17 

### Additional information

Also reproduced with: Samsung Galaxy A32 (Android 11), Samsung Galaxy S22 Ultra (Android 12), Lenovo Yoga Tab 11 (Android 11).

https://user-images.githubusercontent.com/89388888/185162020-4c023311-7032-4a2c-836b-d7ae3b0ac62a.mp4

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21318)

SoftVision-LorandJanos:
I've tried to verify this issue, and I was blocked by a crash: #26723.
Tested on the latest Nightly 106.0a1 (2022-08-31).
Device used: Oppo Find X5 (Android 12).
Removing the QA:needed label until the crash will be fixed.

delia-pop:
https://github.com/mozilla-mobile/fenix/issues/26723 is verified as fixed on the latest Nightly 106 updated from main locally. The app no longer crashes on startup when a wallpaper is selected.

However, in the locally updated Nightly build, this issue is still reproducible: after launching the app, the wallpaper is still displayed with a short delay. Talked to @Mugurell and we will wait for the official Nightly build containing the fix for the #26723  crash to become available on PlayStore of task index, and re-test this.

Mugurell:
Running on an IO thread and then switching to Main with a simple
> Handler(Looper.getMainLooper()).postAtFrontOfQueue 

[...]

Issue #26216, opened 2022-07-27
Create the wallpaper selection onboarding dialog. 


MatthewTighe:
Create a dialog that appears over the home screen allowing users to preview and test out new wallpapers. It should be hidden behind the feature flag defined in [this JIRA ticket](https://mozilla-hub.atlassian.net/browse/FNXV2-21025). There is more behavior defined in the original ticket defining this work [here](https://mozilla-hub.atlassian.net/browse/FNXV2-20989). Some of the underlying behavior has been broken into separate tickets, listed below. The telemetry will be handled in a follow-up.

AC:
- A wallpaper selection tool will be displayed on the homepage under the following the conditions:
  - Tab Pickup and Jump Back In CFRs are not currently being displayed
  - Wallpaper thumbnails have been downloaded
  - The user has not seen the wallpaper selection tool before

- The wallpaper selection tool includes the following wallpaper options:
  - The first 3 classic wallpapers
  - The first 2 seasonal wallpapers (or additional 2 classic wallpapers if there are no seasonal)
  - A no wallpaper option

- When a user clicks on one of the wallpaper thumbnails in the wallpaper selection tool, it will change their homepage background right away, so that they can see it

   - If the full wallpaper has not been downloaded, then they will see a loading indicator & the thumbnail should go 50% less in its opacity until loading is completed

- If the user clicks on the homepage or clicks the X, then the selection tool will close

- If the user clicks on the button to explore wallpapers, they will be taken to the wallpapers setting in the menu
- The wallpaper selection tool matches the [attached design](https://www.figma.com/file/d7l1iGabQYqnLcnQbHDGzM/Wallpaper-2.0?node-id=5410%3A178252)

Dependent on:

1. https://github.com/mozilla-mobile/fenix/issues/26423
2. https://github.com/mozilla-mobile/fenix/issues/26211
3. https://github.com/mozilla-mobile/fenix/issues/26212
4. https://github.com/mozilla-mobile/fenix/issues/26213
5. https://github.com/mozilla-mobile/fenix/issues/26214
6. https://github.com/mozilla-mobile/fenix/issues/26424

Telemetry:

- https://github.com/mozilla-mobile/fenix/issues/26215

All of this work would likely be better served if it was saved until after the refactor work being tracked here: 

- https://github.com/mozilla-mobile/fenix/issues/26034

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21134)

[...]

PR #26515, opened 2022-08-17
For #26511: load homescreen wallpaper in blocking coroutine

MatthewTighe:
### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation

Used by GitHub Actions.


app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -390,7 +390,8 @@ class HomeFragment : Fragment() {
 
         FxNimbus.features.homescreen.recordExposure()
 
-        displayWallpaperIfEnabled()
+        setWallpaperToCurrent()
```
jonalmeida, line 393: This read like it's a hack for something - we're reading from the store first and setting the wallpaper, before we start to observe the store and set the wallpaper again? What is happening that makes it slow to just have the observer which gives you the initial state?

MatthewTighe: Ah, you're right. The repetition there shouldn't actually be necessary. Thanks!

MatthewTighe: Actually, it looks like it is required to get the response time we'd like. I could only make guesses as to why the observation seems to take longer. I would guess it might relate to the flow being collected in a coroutine that is `launch`ed, which is why I had to switch to this `runBlocking` implementation in the first place.

I believe we'll still want the observation though, as we will want the home screen to react to changes that are made from the upcoming wallpaper selection tool https://github.com/mozilla-mobile/fenix/issues/26216. I don't have a good way to avoid these repeated calls of the top of my head, unless we want to start the observation from wherever we launch the selection tool from instead of from `onCreate`

MatthewTighe: Come to think of it, this may just be related to `viewLifecycleOwner.lifecycleScope`. IIRC it doesn't run coroutines until after `onViewCreated`

MatthewTighe: Changing it to the fragment's direct `lifecycleScope` or even the activity's `lifecycleScope` did not seem to improve the issue. 

I would be okay with removing the observation here for now and punting on finding the right place for it until we do the selection tool. The wallpaper should only ever be updated while the fragment is destroyed, since it can only be currently changed from the wallpaper settings. Previously, we had the additional constraint of it being changed when the logo on the home screen was clicked but we have removed that feature

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -923,30 +925,41 @@ class HomeFragment : Fragment() {
...
+                    showWallpaper(state.wallpaperState.currentWallpaper)
                 }
                 .launchIn(viewLifecycleOwner.lifecycleScope)
         }
     }
 
+    private fun setWallpaperToCurrent() {
+        if (shouldEnableWallpaper()) {
+            val wallpaper = requireComponents.appStore.state.wallpaperState.currentWallpaper
+            runBlockingIncrement {
```
gabrielluong, line 942: Curious why we chose to use `runBlockingIncrement` here instead of launching a coroutine like `lifecycleScope.launch(IO) {}`?

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -923,30 +925,41 @@ class HomeFragment : Fragment() {
...
+                showWallpaper(wallpaper)
+            }
+        }
+    }
+
+    private suspend fun showWallpaper(wallpaper: Wallpaper) = when (wallpaper) {
```
gabrielluong, line 948: Since we are calling `setWallpaperToCurrent()` and `observeWallpaperChanges()`, is it possible for the same wallpaper to be shown twice? If so, can we add a check to not do anything if the wallpaper to be shown is already the currently shown wallpaper?

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
gabrielluong: One way we can proceed here is to comment out the `observeWallpaperChanges` code for now and file a follow investigation. Perhaps @Mugurell or @mcarare will have a cycle to dig deeper and suggest a solution. Other than resolving the discussion here, I think the changes looks good from my perspective.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
MatthewTighe: > Since we are calling setWallpaperToCurrent() and observeWallpaperChanges()

There is some conversation above about whether it might even be correct for now to remove `observeWallpaperChanges`. Doing so could clean this up, but it might add extra work to the wallpaper selection tool. The wallpaper would indeed get shown twice using the current strategy, but there is no visual interruption as far as I can tell. It's possible the platform does something smart in terms of caching `ImageView`s, but I don't know for sure.

> If so, can we add a check to not do anything if the wallpaper to be shown is already the currently shown wallpaper?

I would be open to this strategy. I generally try to avoid introducing mutable state where possible, but this sounds like it might be the simplest solution.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
MatthewTighe: Launching it in a `coroutine` is actually what causes the bug. I believe that is because `lifecycleScope`s do not become active until the Fragment is already passed its `onViewCreated` lifecycle state as described in this comment https://github.com/mozilla-mobile/fenix/pull/26515#discussion_r948499212

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
gabrielluong: Ya, I would want to avoid introducing a mutable state as well. See https://github.com/mozilla-mobile/fenix/pull/26515#discussion_r950405365 if this could be a potential solution with move forward with to unblock this.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
MatthewTighe: I think that's a good idea. I will file a follow up

gabrielluong: Cool, I will pre-attemptively approve this for now since I don't need to re-review commented out changes. My only ask is that you throw it the newly filed issue into the current sprint.

gabrielluong (approved):
I would prefer if we landed this after the Beta cut on Monday, and uplifting this after we get QA verified if there is an a demand to get this fix into 105. Otherwise, ignore the uplift comment if we don't expect this fix to be urgently released in 105.

MatthewTighe:
> I would prefer if we landed this after the Beta cut on Monday, and uplifting this after we get QA verified if there is an a demand to get this fix into 105. Otherwise, ignore the uplift comment if we don't expect this fix to be urgently released in 105.

Cool, that sounds good to me. I think this is worth uplifting since the bug is pretty noticeable and would not like it to go to release, but I will follow up with the relevant parties to see if there is agreement once the QA cycle is finished.

MatthewTighe:
Filed https://github.com/mozilla-mobile/fenix/issues/26555 as a follow up

Mugurell:
Investigating https://github.com/mozilla-mobile/fenix/issues/26555 I see cases in which with this code applied the wallpaper would not get set - https://github.com/mozilla-mobile/fenix/issues/26555#issuecomment-1223785660.

MatthewTighe:
I've rebased this over some other wallpaper work and included some changes to cache metadata properties as described in the comments of https://github.com/mozilla-mobile/fenix/issues/26555#issuecomment-1224296559. Unfortunately, I was not address @Mugurell's concerns. The old version of this patch still exists at a branch on my fork here if anyone needs it https://github.com/MatthewTighe/fenix/tree/old-wallpapers-on-main

MozillaNoah:
@Mugurell @gabrielluong Were there any outstanding changes being requested here or any leftover gotchas, or is this clear for landing to move along https://github.com/mozilla-mobile/fenix/pull/26562?

gabrielluong:
> @Mugurell @gabrielluong Were there any outstanding changes being requested here or any leftover gotchas, or is this clear for landing to move along #26562?

I am gonna defer to @Mugurell for the final decision since he has investigated this more thoroughly than I have. The changes have changed quite a bit since my approval as well. @Mugurell can you also do a quick review and provide your recommendation on how we can best unblock this?

Mugurell:
@MozillaNoah @gabrielluong 
Both the original and the subsequent implementation had the bug about only starting to observe the store in `onStart` - after the screen is already visible so the issue would not be resolved.
This was discussed in https://github.com/mozilla-mobile/fenix/issues/26555.
The latest version put by Matt came to help speed setting the wallpaper for the first time by loading it from settings - and not wait to download it. It removed the race that previously existed with reading directly from the state in `onCreateView` but kept the same observing style that would mean the originally reported bug is still there.

I would propose landing this as part of https://github.com/mozilla-mobile/fenix/pull/26647 in one go, with this commit serving as a good base towards faster dispatching of the initial wallpaper to store and then my second commit coming to observe the store and update the wallpaper before HomeFragment is displayed and so fixing the original bug.
Matt left me a message on Slack that a fix should be uplifted to beta so uplifting both commits in one go would ensure a smooth process.

MozillaNoah:
> I would propose landing this as part of #26647 in one go, with this commit serving as a good base towards faster dispatching of the initial wallpaper to store and then my second commit coming to observe the store and update the wallpaper before HomeFragment is displayed and so fixing the original bug. Matt left me a message on Slack that a fix should be uplifted to beta so uplifting both commits in one go would ensure a smooth process.

@Mugurell Gotcha gotcha. Should we aim to land these on Monday to better align their readiness? Since this PR needs its unit tests fixed, #26647 is still a draft, and I want to be respectful to your time zone.

Mugurell:
> > I would propose landing this as part of #26647 in one go, with this commit serving as a good base towards faster dispatching of the initial wallpaper to store and then my second commit coming to observe the store and update the wallpaper before HomeFragment is displayed and so fixing the original bug. Matt left me a message on Slack that a fix should be uplifted to beta so uplifting both commits in one go would ensure a smooth process.
> 
> @Mugurell Gotcha gotcha. Should we aim to land these on Monday to better align their readiness? Since this PR needs its unit tests fixed, #26647 is still a draft, and I want to be respectful to your time zone.

Thank you!
Updated https://github.com/mozilla-mobile/fenix/pull/26647, tested more, for me the combined changes seems to work great. If it gets approved I'd say that can be landed today so that QA can test on Monday and uplift to beta then. Landing on Monday sounds good also.

MozillaNoah:
> Thank you! Updated #26647, tested more, for me the combined changes seems to work great. If it gets approved I'd say that can be landed today so that QA can test on Monday and uplift to beta then. Landing on Monday sounds good also.

OK awesome! Let's land on Monday, just since this particular PR isn't my own code, so it won't be as easy to troubleshoot if anything bad happens. I feel comfortable having you land this, so you can land #26647 immediately after. Otherwise, we can just coordinate when my day starts Pacific time.

PR #26512, opened 2022-08-17
For #26335: Allow for more inflations in performance tests

Mugurell:
The first patch for this enabled the contile feature by default which based on
the automated tests increased the number of inflations done in HomeActivity.
This patch comes to address that by allowing for up to 6 inflations as seen
reported as the actual count in tests.

### Pull Request checklist

- [x] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [x] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not
- [x] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features. In addition, it includes a screenshot of a successful [accessibility scan](https://play.google.com/store/apps/details?id=com.google.android.apps.accessibility.auditor&hl=en_US) to ensure no new defects are added to the product.

### QA

- [x] **QA Needed**

### To download an APK when reviewing a PR (after all CI tasks finished running):
1. Click on `Checks` at the top of the PR page.
2. Click on the `firefoxci-taskcluster` group on the left to expand all tasks.
3. Click on the `build-debug` task.
4. Click on `View task in Taskcluster` in the new `DETAILS` section.
5. The APK links should be on the right side of the screen, named for each CPU architecture.

### GitHub Automation
Fixes #26335

Issue #26380, opened 2022-08-09
Intermittent Unit Test failure - StartupTypeTelemetryTest.WHEN record is called THEN telemetry is recorded with the appropriate label


Mugurell:
### Test Run:
Test failed in https://github.com/mozilla-mobile/fenix/pull/26378 - https://firefoxci.taskcluster-artifacts.net/OIThuwo1TRSMqXOucAYXqg/0/public/reports/test/testDebugUnitTest/index.html 

### Stacktrace:
> java.lang.AssertionError: expected:<1> but was:<null>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:118)
	at org.junit.Assert.assertEquals(Assert.java:144)
	at org.mozilla.fenix.perf.StartupTypeTelemetryTest.WHEN record is called THEN telemetry is recorded with the appropriate label(StartupTypeTelemetryTest.kt:96)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.rules.TestWatcher$1.evaluate(TestWatcher.java:55)
	at org.junit.rules.RunRules.evaluate(RunRules.java:20)
	at org.robolectric.RobolectricTestRunner$HelperTestRunner$1.evaluate(RobolectricTestRunner.java:591)
	at org.robolectric.internal.SandboxTestRunner$2.lambda$evaluate$0(SandboxTestRunner.java:274)
[...]

PR #26516, opened 2022-08-17
Increase the maximum gradle memory limit

rvandermeulen:
Fenix gradle-dependencies jobs run on `c5d.18xlarge` instances with 144GB of RAM, but we're limiting them to 6GB at the moment. This compares to 32GB allocated for m-c on `c5.4xlarge` instances that have 32GB available to them. Given how frequently we hit OOM crashes and how much more expensive these instances are, we should increase the limit to get the most out of them and hopefully massively alleviate the unreliability of these jobs.

### DIFF
--- a/buildSrc/src/main/java/Dependencies.kt
+++ b/buildSrc/src/main/java/Dependencies.kt
@@ -3,7 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 // If you ever need to force a toolchain rebuild (taskcluster) then edit the following comment.
-// FORCE REBUILD 2021-11-24
+// FORCE REBUILD 2022-08-18
 
 object Versions {
     const val kotlin = "1.6.10"
--- a/gradle.properties
+++ b/gradle.properties
@@ -6,7 +6,7 @@
 # http://www.gradle.org/docs/current/userguide/build_environment.html
 # Specifies the JVM arguments used for the daemon process.
 # The setting is particularly useful for tweaking memory settings.
-org.gradle.jvmargs=-Xmx6144m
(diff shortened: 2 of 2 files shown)


rvandermeulen:
Lots of build crashes. Those run on `m5d.xlarge` instances w/ 16GB of RAM, so I guess that's not shocking.

rvandermeulen:
16GB appears to work well. Shall we give it a shot and see how it does in the wild?

mcarare (approved):

Mugurell:
`test-debug` failed because of https://github.com/mozilla-mobile/fenix/issues/26380
`ui-test-x86-debug` failed because of `verifyRunBlockingAndStrictModeSuppresionCount` failing with [matrix](https://console.firebase.google.com/u/1/project/moz-fenix/testlab/histories/bh.66b7091e15d53d45/matrices/7189899921787493362/executions/bs.b9d314e479738088/testcases/1/test-cases):
>  java.lang.AssertionError: start up inflation count does not match expected count.
This heuristic-based performance test is expected measure the number of occurrences of known
performance anti-patterns and fail when that count changes. Please read the class documentation
for more details about this test and an explanation of what the failed heuristic is expected to
measure. Please consult the performance team if you have questions.
actual: 16
at org.junit.Assert.fail(Assert.java:89)
at org.junit.Assert.assertTrue(Assert.java:42)
at org.mozilla.fenix.perf.StartupExcessiveResourceUseTest.verifyRunBlockingAndStrictModeSuppresionCount(StartupExcessiveResourceUseTest.kt:130) 

Based on the discussions from https://github.com/mozilla-mobile/fenix/pull/26512 I'll look into disabling this test.

PR #26532, opened 2022-08-18
For #26497: Switch wallpaper use cases to V2 versions

MatthewTighe:
For #26497

I could potentially split the 2 commits in this PR into separate PRs if that would be preferred. Landing them together will fully shift the backend functionality into the updates for wallpapers V2, however.

This PR:
- renames the existing `WallpaperDownloader` and `WallpaperFileManager` as `Legacy<name>`
- adds new versions of the above that use updated file paths
- does the same for `WallpaperUseCases` that depended on file paths
- switches between `Legacy` and updated versions based on the `wallpapersV2Enabled` feature flag
- adds tests for the new `WallpaperDownloader` which were missing for the old one

Note: IMO there is still quite a bit of tech debt that could be addressed in each of these various components, but my goal was to keep the lift as small as possible for now and address them later. This should be the last PR required to fully set the stage for new wallpaper work for the MR. This includes:
- thumbnail downloading flow
- wallpaper selection tool
- wallpaper settings updates

All of the above is actively being worked on during the course of this sprint.

There will need to be code to migrate existing wallpapers users' currently selected wallpaper to the new path structure for them to continue using the feature without interruption. This is planned for our next sprint.

I'm having trouble uploading video of it working correctly, but I do have one if anyone is interested in seeing it.

### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation

Used by GitHub Actions.


gabrielluong:
Was just looking through PRs to see what is common between them in terms of all these CI failures. Just wanted to point out you got a genuine test failure in https://github.com/mozilla-mobile/fenix/runs/7904725153?check_suite_focus=true

MatthewTighe:
> Was just looking through PRs to see what is common between them in terms of all these CI failures. Just wanted to point out you got a genuine test failure in https://github.com/mozilla-mobile/fenix/runs/7904725153?check_suite_focus=true

Thanks, must have missed compiled the test packages after renaming that one

app/src/main/java/org/mozilla/fenix/wallpapers/Wallpaper.kt
```diff
@@ -74,7 +74,29 @@ data class Wallpaper(
...
+        fun legacyGetLocalPath(orientation: String, theme: String, name: String): String =
             "wallpapers/$orientation/$theme/$name.png"
+
+        /**
+         * Defines the standard path at which a wallpaper resource is kept on disk.
+         *
+         * @param type The type of image that should be retrieved.
+         * @param name The name of the wallpaper.
+         */
+        fun getLocalPath(type: ImageType, name: String) = "wallpapers/$name/${type.lowercase()}.png"
```
MozillaNoah, line 86: small nit: since name comes first in the path, should it be first in the parameters?

app/src/main/java/org/mozilla/fenix/wallpapers/Wallpaper.kt
```diff
@@ -74,7 +74,29 @@ data class Wallpaper(
...
+    }
+
+    /**
+     * Defines various image asset types that can be downloaded for each wallpaper.
+     */
+    enum class ImageType {
+        Portrait,
+        Landscape,
+        Thumbnail;
```
MozillaNoah, line 95: enum class types are typically all uppercase

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperDownloader.kt
```diff
@@ -48,47 +48,29 @@ class WallpaperDownloader(
...
-                    "${collection.name}/" +
-                    "$name.png"
-                WallpaperMetadata(remotePath, localPath)
-            }
+    private fun Wallpaper.toMetadata(): List<WallpaperMetadata> =
+        listOf(Wallpaper.ImageType.Portrait, Wallpaper.ImageType.Landscape).map { orientation ->
+            val localPath = getLocalPath(orientation, this.name)
+            val remotePath = "${collection.name}/${this.name}/${orientation.lowercase()}.png"
+            WallpaperMetadata(remotePath, localPath)
         }
```
MozillaNoah, line 75: pretty clever way to create the list

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperFileManager.kt
```diff
@@ -36,27 +42,21 @@ class WallpaperFileManager(
     }
 
     private fun getAllLocalWallpaperPaths(name: String): List<String> =
-        listOf("landscape", "portrait").flatMap { orientation ->
-            listOf("light", "dark").map { theme ->
-                Wallpaper.getBaseLocalPath(orientation, theme, name)
-            }
+        Wallpaper.ImageType.values().map { orientation ->
+            getLocalPath(orientation, name)
```
MozillaNoah, line 46: is it only the web path/URL that needs the image type name as lowercase?

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperFileManager.kt
```diff
@@ -36,27 +42,21 @@ class WallpaperFileManager(
...
-
-    private fun cleanChildren(dir: File, wallpapersToKeep: List<String>) {
-        for (file in dir.walkTopDown()) {
-            if (file.isDirectory || file.nameWithoutExtension in wallpapersToKeep) continue
-            file.delete()
+            for (file in wallpapersDirectory.listFiles()?.toList() ?: listOf()) {
+                if (file.isDirectory && !wallpapersToKeep.contains(file.name)) {
+                    file.deleteRecursively()
+                }
+            }
```
MozillaNoah, line 59:
```suggestion
            wallpapersDirectory.listFiles()?.toList()?.forEach { file ->
                if (file.isDirectory && !wallpapersToKeep.contains(file.name)) {
                    file.deleteRecursively()
                }
            }
```

app/src/main/java/org/mozilla/fenix/wallpapers/WallpapersUseCases.kt
```diff
@@ -179,6 +207,51 @@ class WallpapersUseCases(
...
+            store.dispatch(AppAction.WallpaperAction.UpdateCurrentWallpaper(currentWallpaper))
+
+            fileManager.clean(
+                currentWallpaper,
+                possibleWallpapers
+            )
+
+            for (wallpaper in possibleWallpapers) {
+                downloader.downloadWallpaper(wallpaper)
+            }
```
MozillaNoah, line 238:
```suggestion
            possibleWallpapers.forEach { downloader.downloadWallpaper(it) }
```

app/src/main/java/org/mozilla/fenix/wallpapers/WallpapersUseCases.kt
```diff
@@ -14,6 +14,7 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.withContext
 import mozilla.components.concept.fetch.Client
 import mozilla.components.support.locale.LocaleManager
+import org.mozilla.fenix.FeatureFlags
```
MozillaNoah, line 17: nit: would it be possible to have the companion object at the bottom of this file/class? It's a little hard to read currently with the use cause API vals at the top and the companion object in the middle of all of the actual use case code.

MozillaNoah (commented):
First pass before I have time to test locally and read over all your tests. What's the intended lifespan of the legacy handling?

All of the updated code looks good as a whole though.

MatthewTighe:
> First pass before I have time to test locally and read over all your tests. What's the intended lifespan of the legacy handling?

I would expect it to be removed at the same time the feature flag is. Not sure if that will be pre or post MR. Would it make sense to file a follow-up to do it all together?

app/src/main/java/org/mozilla/fenix/wallpapers/Wallpaper.kt
MatthewTighe: I think that's a good idea. Having the string type be first also feels like an improvement to me for some reason lol

app/src/main/java/org/mozilla/fenix/wallpapers/Wallpaper.kt
MatthewTighe: Lemme preface this with saying that I am definitely not willing to die on this hill so am happy to change it if you'd like. First though, I have a rant prepared

`enum class`es were required to be uppercase in Java originally (and maybe still?). Kotlin, however, does not require it. Uppercasing them is a stylistic choice that is largely a holdover from the Java days. Even in our codebase, there are mixed usages of the two styles. For example, `BrowserDirections` is pascal cased but `HomeScreenSection` is upper cased. Unless there's a distinction between the two for a reason I'm unaware of, I just tend to prefer not getting yelled at by my code lol

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperDownloader.kt
MatthewTighe: TY!

MatthewTighe: Hopefully not too clever, in the "this is hard to reason about" way

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperFileManager.kt
MatthewTighe: The image type is also lowercased inline in `getLocalPath`. That function will just be referenced from multiple places, whereas the `remotePath` is only ever used from the downloader.

app/src/main/java/org/mozilla/fenix/wallpapers/WallpapersUseCases.kt
MatthewTighe: The companion object for the `LegacyInitializeUseCase`? It would mean shifting that whole class to the bottom. I would be okay with that, but I would argue against moving the new initialize use case down as well, to help preserve the implied suggestion that `Initialize` should be called first.

MatthewTighe: I could also move the companion to the top-level `WallpaperUseCase`  to achieve that

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperFileManager.kt
MatthewTighe: Oh yeah that's much better, thanks

app/src/main/java/org/mozilla/fenix/wallpapers/Wallpaper.kt
MozillaNoah: lol well it's not a merge blocker, for sure. I didn't realize there was both uppercase and mixed case in our codebase, so maybe someone has some context for that.

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperDownloader.kt
MozillaNoah: I mean, it does take a few more seconds to understand _how_ it's iterating over the image types, but it's still some cool kotlin-fu

app/src/main/java/org/mozilla/fenix/wallpapers/WallpapersUseCases.kt
MozillaNoah: Oh my bad, I didn't see it was just in `LegacyInitializeUseCase`. Curly braces are weird, and it looked like the companion object was a part of the top-level class.

MozillaNoah:
> > First pass before I have time to test locally and read over all your tests. What's the intended lifespan of the legacy handling?
> 
> I would expect it to be removed at the same time the feature flag is. Not sure if that will be pre or post MR. Would it make sense to file a follow-up to do it all together?

Yeah, I think that makes sense!

app/src/main/java/org/mozilla/fenix/wallpapers/WallpapersUseCases.kt
MatthewTighe: Yeah, so the whole thing should be deleted when we delete the Legacy case as well. Part of the intent with all these changes is to remove all that local metadata in the first place

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperDownloader.kt
```diff
@@ -4,42 +4,42 @@
...
  *
- * @param context Required for writing files to local storage.
+ * @param filesDir The top level app-local storage directory.
  * @param client Required for fetching files from network.
+ * @param dispatcher Dispatcher used to execute suspending functions. Default parameter
+ * should be likely be used except for when under test.
  */
 class WallpaperDownloader(
-    private val context: Context,
+    private val filesDir: File,
```
MozillaNoah, line 26: nit: thoughts?
```suggestion
    private val fileDirectory: File,
```

app/src/main/java/org/mozilla/fenix/wallpapers/WallpapersUseCases.kt
```diff
@@ -179,6 +207,49 @@ class WallpapersUseCases(
...
+    internal class DefaultInitializeWallpaperUseCase(
+        private val store: AppStore,
+        private val downloader: WallpaperDownloader,
+        private val fileManager: WallpaperFileManager,
+        private val metadataFetcher: WallpaperMetadataFetcher,
+        private val settings: Settings,
+        private val currentLocale: String,
+    ) : InitializeWallpapersUseCase {
+        override suspend fun invoke() {
+            val currentWallpaperName = withContext(Dispatchers.IO) { settings.currentWallpaper }
```
MozillaNoah, line 220: if `settings.currentWallpaper` is only a name vs the current `Wallpaper` data object, should the name be updated to reflect that?

app/src/test/java/org/mozilla/fenix/wallpapers/LegacyWallpaperFileManagerTest.kt
```diff
@@ -0,0 +1,111 @@
...
+import kotlinx.coroutines.test.runTest
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertTrue
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import org.junit.rules.TemporaryFolder
+import java.io.File
+
+class LegacyWallpaperFileManagerTest {
```
MozillaNoah, line 17: anything to call-out for differences to this batch of tests, or is this just the `legacy` migration?

app/src/test/java/org/mozilla/fenix/wallpapers/WallpaperDownloaderTest.kt
```diff
@@ -0,0 +1,134 @@
...
+    fun `GIVEN that copying the file fails WHEN downloading THEN file is not created`() = runTest {
+        val wallpaper = generateWallpaper()
+        val portraitRequest = Request(
+            url = "$remoteHost/${wallpaper.collection.name}/${wallpaper.name}/portrait.png",
+            method = Request.Method.GET
+        )
+        val landscapeRequest = Request(
+            url = "$remoteHost/${wallpaper.collection.name}/${wallpaper.name}/landscape.png",
+            method = Request.Method.GET
+        )
```
MozillaNoah, line 113: if these requests are the same for all the tests, you could optionally make a `generateXRequest()` helper function like you did with `generateWallpaper()`

app/src/test/java/org/mozilla/fenix/wallpapers/WallpaperDownloaderTest.kt
```diff
@@ -0,0 +1,134 @@
...
+        every { mockPortraitResponse.status } returns 400
+        every { mockLandscapeResponse.status } returns 400
```
MozillaNoah, line 91: appreciate the error handling validation. is it only 400s we have to worry about?

app/src/test/java/org/mozilla/fenix/wallpapers/WallpaperFileManagerTest.kt
```diff
@@ -18,51 +14,85 @@ class WallpaperFileManagerTest {
...
-    fun `GIVEN files exist in all directories WHEN expired wallpaper looked up THEN expired wallpaper returned`() = runTest {
+    fun `GIVEN wallpaper directory exists WHEN looked up THEN wallpaper created with correct name`() = runTest {
         val wallpaperName = "name"
         createAllFiles(wallpaperName)
 
         val result = fileManager.lookupExpiredWallpaper(wallpaperName)
 
         val expected = generateWallpaper(name = wallpaperName)
-        assertEquals(expected, result)
+        Assert.assertEquals(expected, result)
```
MozillaNoah, line 40: accidental explicit `Assert.` here?

MozillaNoah (approved):
LGTM. Just a few questions & minor nits/clean-up. Do we want to land this before or after the cut?

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperDownloader.kt
MatthewTighe: `filesDir` matches the property name on `Context` such that `context.filesDir`. I'd still be okay changing it, but if we did I'd prefer to get even more descriptive, something like `storageRootDirectory` maybe?

app/src/test/java/org/mozilla/fenix/wallpapers/LegacyWallpaperFileManagerTest.kt
MatthewTighe: The legacy files should be straight renames. The `WallpaperDownloader` tests are new, and the new `WallpaperFileManagerTest` has the updated paths but I think that's it.

app/src/test/java/org/mozilla/fenix/wallpapers/WallpaperDownloaderTest.kt
MatthewTighe: 🤦 yeah of course. why did I do that to myself lmao. thanks!

app/src/test/java/org/mozilla/fenix/wallpapers/WallpaperDownloaderTest.kt
MatthewTighe: We could potentially add more robust error-handling down the line but tbh I don't really see the point without telemetry or UX to support it. So AFAIK yes. Our `Response` only buckets statuses into successes and failures as seen [here](https://searchfox.org/mozilla-mobile/rev/e935d8b83b94a8d9996f1fa980409457bdcadb9c/android-components/components/concept/fetch/src/main/java/mozilla/components/concept/fetch/Response.kt#142)

MatthewTighe: The underlying code is just checking `response.isSuccess` for now

app/src/test/java/org/mozilla/fenix/wallpapers/WallpaperFileManagerTest.kt
MatthewTighe: ty

app/src/main/java/org/mozilla/fenix/wallpapers/WallpapersUseCases.kt
MatthewTighe: Sure, that seems like a good change to me.

MatthewTighe:
> LGTM. Just a few questions & minor nits/clean-up. Do we want to land this before or after the cut?

Let's wait until after the cut. No need to ship this to 105.

app/src/main/java/org/mozilla/fenix/wallpapers/Wallpaper.kt
Mugurell: I think the mix we currently have is because Kotlin does not enforce uppercase and allows for either naming scheme [depending on an abstract concept of "usage"](https://kotlinlang.org/docs/coding-conventions.html#property-names):
> For enum constants, it's OK to use either uppercase underscore-separated names ([screaming snake case](https://en.wikipedia.org/wiki/Snake_case)) (enum class Color { RED, GREEN }) or upper camel case names, depending on the usage.

We could look into creating our own coding style guide to eliminate subjectivity, have the code more cohesive and simpler reviews.

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperDownloader.kt
MozillaNoah: I totally forgot that was the name of the variable from `context`. your proposal, `storageRootDirectory`, looks good though!

PR #2489, opened 2019-05-14
For #2490 - Update Dockerfile with UI test deps

rpappalax:
### Pull Request checklist

- [x] **Quality**: This PR builds and passes detekt/ktlint checks (A pre-push hook is recommended)
- [x] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [x] **Changelog**: This PR includes [a changelog entry](https://github.com/mozilla-mobile/fenix/blob/master/CHANGELOG.md) or does not need one
- [x] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features

For running automated UI tests in Fenix, we'll first need to land some dependencies in the Dockerfile (as we did recently w/ A-C: https://github.com/mozilla-mobile/android-components/issues/2796).

NOTE:
The version of Ubuntu 17.10 currently being used by master throws the following error:
```
docker build .
Sending build context to Docker daemon  4.096kB
Step 1/22 : FROM ubuntu:17.10
 ---> e211a66937c6
Step 2/22 : MAINTAINER Sebastian Kaspari "skaspari@mozilla.com"
 ---> Running in a9bcbd965781
Removing intermediate container a9bcbd965781
 ---> 1de1e5752d44
Step 3/22 : RUN apt-get update -qq
 ---> Running in b987913a7c8a
E: The repository 'http://security.ubuntu.com/ubuntu artful-security Release' does not have a Release file.
E: The repository 'http://archive.ubuntu.com/ubuntu artful Release' does not have a Release file.
E: The repository 'http://archive.ubuntu.com/ubuntu artful-updates Release' does not have a Release file.
E: The repository 'http://archive.ubuntu.com/ubuntu artful-backports Release' does not have a Release file.
The command '/bin/sh -c apt-get update -qq' returned a non-zero code: 100
```

I've upgraded to Ubuntu 18.04 which solves that problem, but get a different error:
```
> Task :architecture:processReleaseManifest
> Task :app:preAarch64BetaBuild FAILED
FAILURE: Build failed with an exception.
* Where:
Build file '/opt/fenix/app/build.gradle' line: 1726 actionable tasks: 6 executed

* What went wrong:
Execution failed for task ':app:preAarch64BetaBuild'.
> Release builds require the 'versionName' property, e.g.: './gradlew -PversionName=<...> assembleNightly'
```
[...]

PR #26570, opened 2022-08-21
Use parallel execution for Gradle

rvandermeulen:
This should improve performance and is set to true for most other projects. It's not clear to me why it landed set to false back in #2489, but maybe things have changed in the last 3 years.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -10,13 +10,12 @@ org.gradle.jvmargs=-Xmx8g -Xms2g -XX:MaxMetaspaceSize=6g
 # When configured, Gradle will run in incubating parallel mode.
 # This option should only be used with decoupled projects. More details, visit
 # http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
-# org.gradle.parallel=true
+org.gradle.parallel=true
 # Kotlin code style for this project: "official" or "obsolete":
 kotlin.code.style=official
 android.useAndroidX=true
 android.enableJetifier=false
 android.enableUnitTestBinaryResources=false
-org.gradle.parallel=false
 
 # Enables copying of Jetpack Benchmark results from the device to the build directory.
 android.enableAdditionalTestOutput=true


rvandermeulen:
@rpappalax Do you happen to know why this was set to false back when #2489 landed? Are there any extra tests we should run in particular to make sure this works as expected?

rvandermeulen:
Test runs look good on the CI side of things:
https://firefox-ci-tc.services.mozilla.com/tasks/groups/QHO0m8QQQN2XSJrSdvVXpQ
https://firefox-ci-tc.services.mozilla.com/tasks/groups/fGpLuKL-SVaaP8xbb7fjOg
https://firefox-ci-tc.services.mozilla.com/tasks/groups/dIBJt22gSJ2G3qgyVV4y4A
https://firefox-ci-tc.services.mozilla.com/tasks/groups/OMaHxbgGSWq9N4KUNWCspg
https://firefox-ci-tc.services.mozilla.com/tasks/groups/ZCuUDb-3Rvi51NpEuf_7bw
https://firefox-ci-tc.services.mozilla.com/tasks/groups/WSCylDNnSZaSioDp7mZ86w
https://firefox-ci-tc.services.mozilla.com/tasks/groups/d-zBcVihRhi7PO1_yIdd8Q
https://firefox-ci-tc.services.mozilla.com/tasks/groups/FFjc_Ta9RruYvwFPhyYZqg
https://firefox-ci-tc.services.mozilla.com/tasks/groups/d-NcTTCeR1qjCgK0-gcOHw
https://firefox-ci-tc.services.mozilla.com/tasks/groups/cdU12fisT1CBmQf6l1yYsw

AaronMT:
@rvandermeulen 

> @rpappalax Do you happen to know why this was set to false back when #2489 landed? Are there any extra tests we should run in particular to make sure this works as expected?

We don't have much recollection on this property. This could have been set to back to false to get Docker building at the time. Seems like an easy win to enable parallel execution by re-enabling if this the property is still relevant. Things seem to be working here, so I'm fine with the change. Is anything faster with it on?

`org.gradle.parallel`
> When configured, Gradle will fork up to org.gradle.workers.max JVMs to execute projects in parallel. To learn more about parallel task execution, see [the section on Gradle build performance](https://docs.gradle.org/current/userguide/performance.html#parallel_execution). Default is false.

rvandermeulen:
There's a lot of variability between the runs from what I can see, so it's pretty hard to tell what effect it's had on overall runtime in my Try pushes. Thanks for confirming that it should be safe to use now!

rpappalax:
> There's a lot of variability between the runs from what I can see, so it's pretty hard to tell what effect it's had on overall runtime in my Try pushes. Thanks for confirming that it should be safe to use now!

Thank you both! @rvandermeulen @AaronMT

gradle.properties
```diff
@@ -16,7 +16,7 @@ kotlin.code.style=official
 android.useAndroidX=true
 android.enableJetifier=false
 android.enableUnitTestBinaryResources=false
-org.gradle.parallel=false
+org.gradle.parallel=true
```
gabrielluong, line 19: There's actually a commented out `org.gradle.parallel=true` on line 13. Can we remove the current and uncomment the above since it provides more documentation to what the property does?

gabrielluong (approved):

rvandermeulen:
Hah! I saw that it was commented out for Focus but missed it here 🤦‍♂️

PR #26628, opened 2022-08-24
Bug 1784122 - Add non-fission variant for Fenix perf tests.

gmierz:
This patch adds the non-fission variants for Fenix performance tests. Currently they are all running with Fission enabled.


gmierz:
Tasks selected are expected: https://firefoxci.taskcluster-artifacts.net/UQVo724wTbqWVWveljzLYA/0/public/target-tasks.json

Note how the youtube-playback tests only run with fission disabled which is expected. Only the tp6m tests should be running with fission enabled.

See here for the test runs: https://github.com/mozilla-mobile/fenix/runs/7994028377

gmierz:
I'm not sure why the tests failed, but I think it might have something to do with how I made the PR in a fork. I don't think this patch could cause the signing builds to fail.

escapewindow (approved):
Looks reasonable, and it looks like the tests passed.

gmierz:
Thanks for the review @escapewindow!

PR #26631, opened 2022-08-24
For #26444: Adapt home screen text to wallpaper selection.

### LINKED ISSUE #26444: Home screen text should use values associated with wallpapers
In https://github.com/mozilla-mobile/fenix/pull/26281 we added the ability for the home screen header and link text to react to whether a wallpaper was set. 

After https://github.com/mozilla-mobile/fenix/issues/26423 is resolved, there should be metadata available for each individual wallpaper describing whether light text or dark text should be used to overlay the wallpaper. The implementation should be updated to:

- access the text color property available on the current wallpaper and apply it when the current wallpaper is updated

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21277)

mcarare:
### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation
Fixes #26444

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/compose/SectionHeader.kt
+++ b/app/src/main/java/org/mozilla/fenix/compose/SectionHeader.kt
@@ -7,8 +7,12 @@ package org.mozilla.fenix.compose
 import androidx.compose.material.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.text.style.TextOverflow
 import androidx.compose.ui.tooling.preview.Preview
+import mozilla.components.lib.state.ext.observeAsComposableState
+import org.mozilla.fenix.components.AppStore
+import org.mozilla.fenix.components.components
 import org.mozilla.fenix.theme.FirefoxTheme
 
 /**
@@ -20,12 +24,18 @@ import org.mozilla.fenix.theme.FirefoxTheme
 @Composable
 fun SectionHeader(
     text: String,
-    modifier: Modifier = Modifier
+    modifier: Modifier = Modifier,
+    appStore: AppStore = components.appStore
 ) {
+    val wallpaperState = appStore
+        .observeAsComposableState { state -> state.wallpaperState }.value
+
+    val wallpaperAdaptedTextColor = wallpaperState?.currentWallpaper?.textColor?.let { Color(it) }
+
     Text(
         text = text,
         modifier = modifier,
-        color = FirefoxTheme.colors.textPrimary,
+        color = wallpaperAdaptedTextColor ?: FirefoxTheme.colors.textPrimary,
         overflow = TextOverflow.Ellipsis,
         maxLines = 2,
         style = FirefoxTheme.typography.headline7
@@ -35,5 +45,7 @@ fun SectionHeader(
 @Composable
 @Preview
 private fun HeadingTextPreview() {
-    SectionHeader(text = "Section title")
+    FirefoxTheme {
+        SectionHeader(text = "Section title", appStore = AppStore())
+    }
 }
--- a/app/src/main/java/org/mozilla/fenix/compose/SectionHeader.kt
+++ b/app/src/main/java/org/mozilla/fenix/compose/SectionHeader.kt
@@ -9,7 +9,6 @@ import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.toArgb
-import androidx.compose.ui.text.style.TextOverflow
 import androidx.compose.ui.tooling.preview.Preview
 import mozilla.components.lib.state.ext.observeAsComposableState
 import org.mozilla.fenix.components.AppStore
@@ -36,13 +35,10 @@ fun SectionHeader(
 
     val wallpaperAdaptedTextColor = wallpaperState?.currentWallpaper?.textColor?.let { Color(it) }
 
-    Text(
+    SimpleHeader(
         text = text,
         modifier = modifier,
         color = wallpaperAdaptedTextColor ?: FirefoxTheme.colors.textPrimary,
-        overflow = TextOverflow.Ellipsis,
-        maxLines = 2,
-        style = FirefoxTheme.typography.headline7
     )
 }
 
--- a/app/src/main/java/org/mozilla/fenix/compose/SectionHeader.kt
+++ b/app/src/main/java/org/mozilla/fenix/compose/SectionHeader.kt
@@ -8,12 +8,16 @@ import androidx.compose.material.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.toArgb
 import androidx.compose.ui.text.style.TextOverflow
 import androidx.compose.ui.tooling.preview.Preview
 import mozilla.components.lib.state.ext.observeAsComposableState
 import org.mozilla.fenix.components.AppStore
+import org.mozilla.fenix.components.appstate.AppState
 import org.mozilla.fenix.components.components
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/mozilla/fenix/compose/SectionHeader.kt
```diff
@@ -22,10 +25,15 @@ fun SectionHeader(
     text: String,
     modifier: Modifier = Modifier
 ) {
+    val wallpaperState = components.appStore
+        .observeAsComposableState { state -> state.wallpaperState }.value
+
+    val wallpaperAdaptedTextColor = wallpaperState?.currentWallpaper?.textColor?.let { Color(it) }
```
mcarare, line 31: This assumes the `Long` type of `textColor` param is the corresponding value.

Mugurell (approved):
One interesting aspect is that 
- for `HomeSectionHeaderContent` the wallpaper color would replace `FirefoxTheme.colors.textAccent` and 
- for `SectionHeader` the wallpaper color would replace `FirefoxTheme.colors.textPrimary`.

but this seems to be the expected outcome -> LGTM.

app/src/main/java/org/mozilla/fenix/compose/SectionHeader.kt
gabrielluong, line 31: This could be a followup, but I think we should actually add a `color` parameter to `SectionHeader`, `HomeSectionHeaderContent`, and `PoweredByPocketHeader`. I don't think these components should need to observe the appstore.

app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesComposables.kt
```diff
@@ -403,6 +405,11 @@ fun PoweredByPocketHeader(
     val linkStartIndex = text.indexOf(link)
     val linkEndIndex = linkStartIndex + link.length
 
+    val wallpaperState = components.appStore
+        .observeAsComposableState { state -> state.wallpaperState }.value
+
+    val wallpaperAdaptedTextColor = wallpaperState?.currentWallpaper?.textColor?.let { Color(it) }
```
MozillaNoah, line 411: the call to `components` will break the previews, unfortunately. the text color will need to be passed-in as a parameter and the default could then just be `FirefoxTheme.colors.textPrimary`. 

As @gabrielluong said, this can be follow-up.

gabrielluong: Ah, this is an even more compelling reason that goes with my initial thoughts.

app/src/main/java/org/mozilla/fenix/compose/SectionHeader.kt
mcarare: Agree, for when we would have all the home screen sections (including their list) in Compose. At the moment, hoisting the observation would imply that we need to observe in multiple places (all the different sections that use SectionHeader) and pass the color down. That would mean duplicating code in too many places.

gabrielluong: I think duplicating the code is fine given the fact that we currently lose out on previews of these composable.

app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesComposables.kt
mcarare: I fixed the preview by passing the store with a default value. I do agree that when we finish migrating the whole screen to compose passing the colors directly and observing changes in the topmost composable is a cleaner solution.

gabrielluong: If my count is correct, you would need to do the duplicate code 3-4 times, I think this would still honestly be better than having `appStore` as a parameter to these components.

mcarare: While I might agree with the fact that passing the color is better in some cases, IMO there are more arguments that support this implementation at this stage.

The viewholders do not need the info contained in the state, their UI does not change, with the exception of the section header. Moreover, if we add new sections we will add other code duplications, for components that are temporary in the plan of migrating the screen to full Compose. 

As we migrate the screen to Compose the section headers are still the ones that need to change depending on the state, having the state in the viewholders would add complexity to the refactoring. 

As for the previews, If the section header text is supposed to be wallpaper dependent, I think having the possibility to represent that dependency directly (by adding custom wallpapers as the current wallpaper in the preview param state) is more suggestive than having it change by an explicit color.

mcarare: As for the previews, If the section header text is supposed to be wallpaper dependent, I think having the possibility to represent that dependency directly (by adding custom wallpapers as the current wallpaper in the preview param state) is more suggestive than having it change by an explicit color. 
This is illustrated in the last commit: https://github.com/mozilla-mobile/fenix/pull/26631/commits/1e7eda26b1cd37a7b9e63ca5f564659b4f9a9045.

Having these previews also adds to the argument that the current parent of the section header (the viewholders) does not need to instruct the child on what color to use as they do not need to use that information elsewhere. As a side note, there also is the hypothetical discussion that if we would need 10 values from the state we would pass the whole state, not 10 separate params.

app/src/main/java/org/mozilla/fenix/compose/SectionHeader.kt
```diff
@@ -20,20 +28,53 @@ import org.mozilla.fenix.theme.FirefoxTheme
 @Composable
 fun SectionHeader(
     text: String,
-    modifier: Modifier = Modifier
+    modifier: Modifier = Modifier,
+    appStore: AppStore = components.appStore
```
gabrielluong, line 32: This needs a change to a `color` parameter. Currently, `SectionHeader` is used in `MessageCard` so the changes to impose observing the `appStore` directly in this composable components will affect the reusability of this component.

gabrielluong (changes requested):

app/src/main/java/org/mozilla/fenix/compose/SectionHeader.kt
mcarare: I have extracted the text header from the section header, creating a reusable `header7` composable.
This also uses a more correct naming as dialog, banners, and message cards do not have sections or section headers.

Do you also think we could rename `SimpleHeader` to a name that links it to the Fenix typography? Thank you!

mcarare: I also wonder if, in future refactoring, we should merge the `SectionHeader` with the `HomeSectionHeader` and conditionally handle the visibility of the show-all text.

app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesComposables.kt
MozillaNoah: Personally, I'm in the camp of simple passed-in parameters. If we do one day make it to 100% Compose, there _will_ be Compose code interwoven with interactors/use cases/stores etc, but the ultimate front-end UI `@Compose` functions should be kept as dumb as possible and only get passed parameters that directly drive the experience (lambdas, specific colors, data, etc.). 

An example of our transition to 100% has been our use of ViewHolders. Looking at `InactiveTabViewHolder`, `SyncedTabsPageViewHolder`, and `RecentBookmarksViewHolder`, the state/store is observed in the ViewHolder, and the Composables they contain only receive what is absolutely necessary. This keeps the API super clean and transparent, and streamlines `@Previews` since they don't need to do any substantial object instantiation. When we transition off of ViewHolders, there will still exist an intermediary layer between the interactors/use cases/stores etc and the front-end Consumers (e.g. InactiveTabsList, RecentBookmarks), so it behooves us to do that work now and keep that separation.

mcarare:
Closing this in favor of [insert new PR here] to keep the current state of the code for future reference.

PR #26634, opened 2022-08-24
Set the Gradle max heap size to 7g

rvandermeulen:
Github Actions workers only have 7GB available, so the 8GB limit is causing crashes in that environment.


rvandermeulen:
@AaronMT Confirmed that this is fixing the crashes he's been seeing. We'll see whether gradle-dependencies complains.

gabrielluong (approved):

PR #26653, opened 2022-08-25
Close #26666: Use correct content scale when displaying recent tab thumbnail and icon

### LINKED ISSUE #26591: Places history storage appears to be missing preview URLs
This was discovered during investigation into https://github.com/mozilla-mobile/fenix/issues/26124

The tab pickup feature attempts to find match preview URLs from synced history to a recent synced tab so the image can be displayed on the home screen. When this behavior was initially being worked on, I thought I remembered that `HistoryStorage` had plenty of `previewImageUrl` to match. Now, however, I am not seeing any.

This should be investigated further to determine why there aren't preview image URLs available. I inspected this by placing debug breakpoints at [this line](https://searchfox.org/mozilla-mobile/rev/99f85f445e999edb88c2cca1530aec1fa6a39939/fenix/app/src/main/java/org/mozilla/fenix/home/recentsyncedtabs/RecentSyncedTabFeature.kt#139) and inspecting the `history` variable there.

I would be curious to know whether anything changed at the A-S layer recently that may be causing the lack of preview image URLs. It's also possible that I'm misremembering their existence, but I don't think that's the case.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21359)

### LINKED ISSUE #26666: Recent tab thumbnail and icon not using correct content scale
Currently we're not using the correct content scale in thumbnail and icon.  This could be an issue when the image or icon doesn't have the right ratio or size.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21402)

rocketsroger:
This only fixes the case before the user opens the tab.  Once the tab is opened, we'll retrieve icon will use hero image.
To reproduce the problem.  Open apple.com and set to open on home when start.  Close Fenix and reopen.

### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation
Fixes #26666

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabs.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabs.kt
@@ -126,7 +126,8 @@ private fun RecentTabItem(
                 tab = tab,
                 modifier = Modifier
                     .size(108.dp, 80.dp)
-                    .clip(RoundedCornerShape(8.dp))
+                    .clip(RoundedCornerShape(8.dp)),
+                contentScale = ContentScale.Crop,
             )
 
             Spacer(modifier = Modifier.width(16.dp))
@@ -214,7 +215,8 @@ fun RecentTabImage(
         else -> ThumbnailCard(
             url = tab.state.content.url,
             key = tab.state.id,
-            modifier = modifier
+            modifier = modifier,
+            contentScale = contentScale,
         )
     }
 }
@@ -305,7 +307,7 @@ private fun RecentTabIcon(
                         painter = icon.painter,
                         contentDescription = null,
                         modifier = modifier,
-                        contentScale = ContentScale.Fit
+                        contentScale = contentScale,
                     )
                 }
             }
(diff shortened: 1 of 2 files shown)


rocketsroger:
With fix (Left), without fix (Right)
Screenshot_20220825_162449

gabrielluong (approved):
The commit message doesn't quite match what you're doing here. So, I would update that before landing.

rocketsroger:
> The commit message doesn't quite match what you're doing here. So, I would update that before landing.

Sure, I'll update.  Thanks

app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabs.kt
```diff
@@ -213,8 +214,9 @@ fun RecentTabImage(
         }
         else -> ThumbnailCard(
             url = tab.state.content.url,
-            key = tab.state.id,
-            modifier = modifier
+            key = tab.state.content.url.hashCode().toString(),
```
gabrielluong, line 217: This should actually be using `tab.state.id`

rocketsroger: As discussed, your correct that this key should not be changed.  I'll remove this part and open a new issue for the content scale fix.  Thanks

gabrielluong (approved):

PR #26665, opened 2022-08-26
Close #25919: Move UpdateFirstFrameDrawn to RecyclerView.onLayoutCompleted

### LINKED ISSUE #25919: 533-1279ms regression + bimodal reintroduced for cold main start up (6/28)
We see a regression of 533ms from the 6/27 nightly to the 6/28 nightly:

image

That increases if we look at 6/27 to 6/29 to 1279ms. We also see a regression on the A51 (but it is still a WIP) so we can probably reproduce this on devices that aren't just the Moto G5.

Bimodal behavior also appears starting in 6/28. **The fast case appears to not have regression but the slow case is where we see the regression.**  Here's 6/27 (before):
image

And 6/28:
image

6/29 for completeness:
image

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-20930)

jonalmeida:
The longer investigation that led to this patch can be found 
here: https://github.com/mozilla-mobile/fenix/issues/25919#issuecomment-1226151432

As part of a preventitive measure for home page regression loading, we
fixed the UpdateFirstFrameDrawn call to happen after the first layout in
the main RecyclerView is completed. In addition, we also make pocket
aware of this flag so that it renders itself after the first layout.

This helps prioritize current & future features that are visible first
to render before those that are off screen.

Co-authored-by: Arturo Mejia

### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation
Fixes #25919

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlView.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlView.kt
@@ -12,6 +12,7 @@ import androidx.recyclerview.widget.RecyclerView
 import mozilla.components.feature.tab.collections.TabCollection
 import mozilla.components.feature.top.sites.TopSite
 import mozilla.components.service.pocket.PocketStory
+import org.mozilla.fenix.components.appstate.AppAction
 import org.mozilla.fenix.components.appstate.AppState
 import org.mozilla.fenix.ext.components
 import org.mozilla.fenix.ext.settings
@@ -41,7 +42,8 @@ internal fun normalModeAdapterItems(
     showRecentTab: Boolean,
     showRecentSyncedTab: Boolean,
     recentVisits: List<RecentlyVisitedItem>,
-    pocketStories: List<PocketStory>
+    pocketStories: List<PocketStory>,
+    firstFrameDrawn: Boolean = false,
 ): List<AdapterItem> {
     val items = mutableListOf<AdapterItem>()
     var shouldShowCustomizeHome = false
@@ -86,7 +88,11 @@ internal fun normalModeAdapterItems(
         showCollections(collections, expandedCollections, items)
     }
 
-    if (settings.showPocketRecommendationsFeature && pocketStories.isNotEmpty()) {
+    // When Pocket is enabled and the initial layout of the app is done, then we can add these items
+    // to render to the home screen.
+    // This is only useful while we have a RecyclerView + Compose implementation. We can remove this
+    // when we switch to a Compose-only home screen.
+    if (firstFrameDrawn && settings.showPocketRecommendationsFeature && pocketStories.isNotEmpty()) {
         shouldShowCustomizeHome = true
         items.add(AdapterItem.PocketStoriesItem)
         items.add(AdapterItem.PocketCategoriesItem)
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlView.kt
```diff
@@ -40,7 +41,8 @@ internal fun normalModeAdapterItems(
     showRecentTab: Boolean,
     showRecentSyncedTab: Boolean,
     recentVisits: List<RecentlyVisitedItem>,
-    pocketStories: List<PocketStory>
+    pocketStories: List<PocketStory>,
+    firstFrameDrawn: Boolean = false,
```
MozillaNoah, line 45: outside of the updated test, is there ever a scenario where `firstFrameDrawn` is passed-in as `true`?

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlView.kt
```diff
@@ -85,7 +87,11 @@ internal fun normalModeAdapterItems(
         showCollections(collections, expandedCollections, items)
     }
 
-    if (settings.showPocketRecommendationsFeature && pocketStories.isNotEmpty()) {
+    // When Pocket is enabled and the initial layout of the app is done, then we can add these items
+    // to render to the home screen.
+    // This is only useful while we have a RecyclerView + Compose implementation. We can remove this
+    // when we switch to a Compose-only home screen.
```
MozillaNoah, line 93: love comments like these

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlView.kt
jonalmeida: Yeah, below in the `LinearLayoutManager.onLayoutCompleted` call is when we want this flag to be flipped.

MozillaNoah: I saw that's where the flag is being updated, I just wasn't sure if that flag flip then triggered something in `SessionControlView` or if `SessionControlView` was ever re-initialized with `firstFrameDrawn` as true.

jonalmeida: Ah, I think I understand your comment now. In the `HomeFragment`, we observe the entire `AppState` for changes [here](https://github.com/mozilla-mobile/fenix/blob/f8f8b4f8132366e5b40ef1687d696e4d5a025776/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt#L452-L455) and calls the `sessionControlView.update` method on any state changes which in return re-evaluates the `normalModeAdapterItems` and therefore the `firstFrameDrawn` flag.

jonalmeida:
Added the do not land label for now so we can land this on Monday.

csadilek (approved):
Thanks for walking me through and all your work investigating this tricky issue @jonalmeida @Amejia481 @MatthewTighe.

I think differentiating primary and secondary content on the home screen and not rendering everything at once, especially the parts that are likely below the fold, seems really good to me. We can always make this more flexible later!

PR #26667, opened 2022-08-26
Added secret debug menu preference to override the push server URL ( #24170)

### LINKED ISSUE #24170: Read configuration information from secret settings
The fact that Fenix requires custom builds in order to test non production hosts is unsustainable. The combination of services, run environments, and interactions may have dozens or hundreds of potential configurations. Asking Service Reliability Engineering (SRE) and Quality Assurance (QA) to compile a custom version of fenix for any of these configurations (including staying abreast of all the compilation environment changes) is an unweildy burden. In addition, not being able to specify server impacts Mozilla’s ability to ensure user choice. Fenix should allow for easy modification of these values so that running a test setup requires minimal configuration changes of an existing application. Based on the impact of this on the operational efficiency of other teams, this should be made a priority.

Tracking with [CONSVC Jira ticket](https://mozilla-hub.atlassian.net/browse/CONSVC-1661)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-19779)

bendk:
- Created a new "sync debug" pref screen to that's only enabled when the debug menu is to hold the Fxa, Sync, and Push
  server override prefs.  They were taking a lot of screen space on the  top-level settings menu as individual items
- Added button on that screen to quit FF which is needed to apply the changes.
    - This is definitely not the nicest UI, but hopefully QA can just override the prefs once save them in an emulator and never have to go back to this screen.
    - I do think this is a nicer UI than before, where FF would quit after a change to any of the prefs.  That forces you to restart FF 3 times if you wanted to override all 3 server URLs.

Here's my attempt to describe the UI:
image

I couldn't think of a good way to unit test this, but I tested this with a dev build of Fenix, using the keys from `fenix_firebase.xml` and [a-c#12687](https://github.com/mozilla-mobile/android-components/pull/12687) and sending a tab seemed to work.  I added some debugging in app-services, so you should be able to double check the push server URL by with `adb logcat | grep push` (once a-c picks up a new a-s release).  The log line will look like this:

```
D push    : PushManager server_host: endpoint-autopush.stage.mozaws.net, http_protocol: https
```

I believe the firebase keys built into release builds will work with an overridden push server, but I'm not sure.  Can someone who understands this better than me confirm this?

### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation
Fixes #24170


app/src/main/java/org/mozilla/fenix/components/Push.kt
```diff
@@ -40,7 +43,22 @@ class Push(context: Context, crashReporter: CrashReporter) {
 
         logger.debug("Creating push configuration for autopush.")
         val projectId = context.resources.getString(resId)
-        PushConfig(projectId)
+        val serverOverride = context.settings().overridePushServer
```
jonalmeida, line 46: I'm surprised this line doesn't invoke strict mode. Were you able to test this patch with a first-run install and a synced account setup?

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -1057,6 +1057,11 @@ class Settings(private val appContext: Context) : PreferencesHolder {
         default = ""
     )
 
+    var overridePushServer by stringPreference(
```
jonalmeida, line 1060: Why use a string pref instead of a boolean?
```suggestion
    var overridePushServer by booleanPreference(
```

app/src/main/res/values/strings.xml
```diff
@@ -350,10 +350,18 @@
...
     <!-- Preference for accessibility -->
     <string name="preferences_accessibility">Accessibility</string>
+    <!-- Title of preference for sync debugging (only shown in the when the secret debug menu is enabled) -->
+    <string name="preferences_sync_debug">Sync Debug</string>
     <!-- Preference to override the Firefox Account server -->
     <string name="preferences_override_fxa_server">Custom Firefox Account server</string>
     <!-- Preference to override the Sync token server -->
     <string name="preferences_override_sync_tokenserver">Custom Sync server</string>
+    <!-- Preference to override the Push server -->
+    <string name="preferences_override_push_server">Custom Push server</string>
```
jonalmeida, line 360: Do these strings need to be translated? `Custom Sync Server` are strings that meant for end-users, where as debug ones are usually meant for us (internal contributors).

If they are meant for end-users, our string descriptions should also include the bit about how to access these settings (as you've done above with "only shown in the when the secret debug menu is enabled"). Each string is individually translated by l10n contributors.

If not, we can move these to the [`static_strings.xml`](https://github.com/mozilla-mobile/fenix/blob/main/app/src/main/res/values/static_strings.xml) file.

jonalmeida (commented):
Left a few questions below, but this looks almost ready to land!

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
bendk: It stores the overriden URL.  Maybe I should have called it `overredPushServerUrl`, but I just tried to match the existing `overrideSyncTokenServer` pref.

app/src/main/res/values/strings.xml
bendk: Good catch.  I moved the new strings to `static_strings.xml`.  I considered moving the related Fxa/Token server strings over too, but I noticed the had some translations and were maybe used for the China release, so I just left those alone.

jonalmeida: The Fxa/Token Server ones are translated because they are user facing strings for clients that host custom servers, so I think those make sense.

app/src/main/java/org/mozilla/fenix/utils/Settings.kt
jonalmeida: Ah, that's fair.

jonalmeida (approved):
Excellent, LGTM! 

Will land this tomorrow morning.

PR #26685, opened 2022-08-29
For #26424 - Create wallpaper file migration helper

### LINKED ISSUE #26424: Create wallpaper file migration helper
We plan to simplify the number of assets possible for each wallpaper. Previously, there were combinations of all of portrait/landscape, light/dark, and high/medium/low resolutions for each wallpaper. We will no longer be providing separate assets for themes or resolutions, so there will only be separate assets for different orientation (and thumbnails).

This will mean that currently saved versions of wallpapers will be stored under directory paths that are no longer relevant, like `wallpapers/<orientation>/<theme>/<wallpaperName>.png`. The new paths should follow a convention like `wallpapers/<wallpaperName>/portrait.png` and `wallpapers/<wallpaperName>/landscape.png`.

This will require migrating existing assets to the new path names. This is required because some users may have expired wallpapers that are no longer available, so we cannot just download a fresh asset. It is likely fine to ignore the consideration of updating the themed wallpaper to a non-themed wallpaper. Both the landscape and the portrait versions of an image should be copied.

If there is not a thumbnail available for the wallpaper (as could be the case with Turning Red wallpapers that are no longer available), then the portrait image of the wallpaper should also be copied as `wallpapers/<wallpaperName>/thumbnail.png` to ensure that it can be visible in the wallpaper settings page.

AC:
1. Install a build of the app that is not using the `wallpapersV2Enabled` feature flag, then set a wallpaper.
2. Overwrite it with a build that includes the migration code and has the flag enabled.
3. The selected wallpaper should still be selected.
4. The selected wallpaper can be rotated correctly.
5. The wallpaper is visible in the settings page. For wallpapers that have a thumbnail available remotely, it should use the new remote version.

The `WallpaperDownloader` and `WallpaperFileManager` contain code that would be a good starting point to building context around this issue.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21267)

Alexandru2909:
Migrate legacy wallpapers to the respective V2 paths. If a thumbnail for the legacy wallpaper is available remote, then it should be used, otherwise the portrait wallpapers should be used instead.

### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation
Fixes #26424


app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperDownloader.kt
```diff
@@ -65,6 +65,36 @@ class WallpaperDownloader(
         }
     }
 
+    /**
+     * Attempts to download the thumbnail for a wallpaper.
+     *
+     * @param wallpaper The wallpaper for which we want to download the thumbnail.
+     * @param onFailure onFailure callback invoked when the thumbnail fails to download.
+     */
+    fun downloadThumbnail(wallpaper: Wallpaper, onFailure: (Throwable) -> Unit) {
```
MozillaNoah, line 74: will this conflict with https://github.com/mozilla-mobile/fenix/pull/26562?

Alexandru2909: Yes, I'll rebase and remove this function once the work for #26562 is merged. Since we'll be waiting for the PR to land I'll add the do-not-land label here.

app/src/main/java/org/mozilla/fenix/wallpapers/WallpapersUseCases.kt
```diff
@@ -228,14 +229,14 @@ class WallpapersUseCases(
             // Dispatching this early will make it accessible to the home screen ASAP
             store.dispatch(AppAction.WallpaperAction.UpdateCurrentWallpaper(currentWallpaper))
 
-            fileManager.clean(
-                currentWallpaper,
-                possibleWallpapers
+            val migratedWallpapers = fileManager.migrateAndCleanLegacyWallpapers(
+                wallpaperList,
```
MozillaNoah, line 233: nit: could you add the named parameters here

app/src/main/java/org/mozilla/fenix/wallpapers/WallpapersUseCases.kt
```diff
@@ -228,14 +229,14 @@ class WallpapersUseCases(
...
+                downloader::downloadThumbnail
```
MozillaNoah, line 234: will the migration code be ran every time this use case is invoked?

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperFileManager.kt
```diff
@@ -47,16 +46,99 @@ class WallpaperFileManager(
...
+            landscapeFile.copyTo(
+                File(
+                    storageRootDirectory,
+                    "wallpapers/${name.lowercase()}/landscape.png"
+                )
+            )
         }
+        // Delete the remaining legacy files
+        File(storageRootDirectory, "wallpapers/portrait").deleteRecursively()
+        File(storageRootDirectory, "wallpapers/landscape").deleteRecursively()
```
MozillaNoah, line 142: does there need to be any File exception handling for any of the copy/delete operations?

app/src/main/java/org/mozilla/fenix/wallpapers/WallpapersUseCases.kt
```diff
@@ -248,17 +249,28 @@ class WallpapersUseCases(
...
+                        downloadThumbnail = downloader::downloadThumbnail
+                    )
+                } else {
+                    fileManager.getMigratedWallpapers(
+                        remoteAvailableWallpapers = wallpaperList,
+                        localAvailableWallpapers = possibleWallpapers
+                    )
+                }
+
+            val defaultIncluded = listOf(Wallpaper.Default) + wallpapersWithUpdatedThumbnailState +
```
MatthewTighe, line 272: Including `migratedWallpapers` here shouldn't be necessary, as there is not currently a requirement to display an expired wallpaper in the available wallpaper list in the settings page.

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperFileManager.kt
```diff
@@ -49,17 +50,126 @@ class WallpaperFileManager(
...
+     * @param remoteAvailableWallpapers List of wallpapers available remote.
+     * @param localAvailableWallpapers List of wallpapers available locally.
+     * @return List of migrated wallpapers.
      */
-    suspend fun clean(currentWallpaper: Wallpaper, availableWallpapers: List<Wallpaper>) = withContext(Dispatchers.IO) {
-        scope.launch {
-            val wallpapersToKeep = (listOf(currentWallpaper) + availableWallpapers).map { it.name }
-            wallpapersDirectory.listFiles()?.forEach { file ->
-                if (file.isDirectory && !wallpapersToKeep.contains(file.name)) {
-                    file.deleteRecursively()
```
MatthewTighe, line 59: We should check with product before removing this `clean` method. It is intended to remove promotional wallpaper files once those promotions expire. During the Turning Red promotion, there was quite a bit of focus on making sure users would not be able to access Turning Red assets that they had not intended to keep. I would be concerned that we might have similar requirements from promotional partners in the future.

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperFileManager.kt
```diff
@@ -49,17 +50,126 @@ class WallpaperFileManager(
...
+    suspend fun getMigratedWallpapers(
+        remoteAvailableWallpapers: List<Wallpaper>,
+        localAvailableWallpapers: List<Wallpaper>,
+    ) = withContext(Dispatchers.IO) {
+        File(storageRootDirectory, "wallpapers").listFiles()
+            ?.filter { file ->
```
MatthewTighe, line 64: It might make sense to put the migration code into a specialized `WallpaperLegacyMigration` class, so that the whole thing can be easily deleted once we're satisfied that enough time has passed.

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperFileManager.kt
```diff
@@ -49,17 +50,126 @@ class WallpaperFileManager(
...
+            val landscapeFile = pair.second
+            val name = portraitFile.nameWithoutExtension
+            val directory = "wallpapers/${name.lowercase()}"
+            val remoteWallpaper = availableWallpapers.firstOrNull { it.name == name }
+
+            try {
+                // If there is a thumbnail available for the current wallpaper,
+                // attempt to download it
+                if (remoteWallpaper == null ||
+                    downloadThumbnail(remoteWallpaper) == Wallpaper.ImageFileState.Error
```
MatthewTighe, line 141: I applaud the thoroughness of downloading a correct thumbnail if it exists. In the interest of simplicity, I think it would also be fine to just use the portrait image as the thumbnail. That would also ensure the thumbnail matches the legacy version of the wallpaper.

MatthewTighe (commented):
This is another case of me needing to specify more issues more clearly. The main goal of this issue is to ensure that users don't lose access to their selected wallpaper, or need to re-select it when they update to 106. Because of that, I don't think we need to migrate all existing assets that a user has. We should be able to migrate only their currently selected wallpaper.

app/src/main/java/org/mozilla/fenix/wallpapers/WallpapersUseCases.kt
Alexandru2909: Just to be sure this is the intended behavior: When testing this PR I noticed the migrated wallpaper (while still set) is not displayed in the wallpaper settings. Therefore if the user selects any other wallpaper the migrated one will be cleaned on the next startup. Here is a video of the behavior I described: 

https://user-images.githubusercontent.com/35462038/188904518-938e5480-7d96-407d-a223-7b63335cdb3b.mp4

MatthewTighe: Yes, that behavior is intended (if suboptimal). During wallpapers V1 it was decided that users could keep their selected wallpaper once a promotion ended, but it would no longer appear in settings and if the wallpaper was switched the user would lose access to it. This was mostly a "this is the easiest thing to do for now" decision.

There was some talk about improving that in the future, such that once a user selected a wallpaper it would be available in their "collection" indefinitely. However, such an improvement I think is outside the scope of this issue and could instead be done during further iterations of the feature.

Totally unrelated: I spotted your deftones search - nice taste lol

app/src/main/java/org/mozilla/fenix/wallpapers/WallpapersUseCases.kt
```diff
@@ -242,6 +246,7 @@ class WallpapersUseCases(
             }
             val currentWallpaper = possibleWallpapers.find { it.name == currentWallpaperName }
                 ?: fileManager.lookupExpiredWallpaper(currentWallpaperName)
+                ?: migrationHelper.migrateLegacyWallpaper(currentWallpaperName)
```
MatthewTighe, line 249: Rather than include this here, I think it would make sense to run the migration separately after the current wallpaper is dispatched on line 240. We could also wrap the call in a conditional based on a property in `Settings` so that we only do it once. I believe I saw that included in a previous version of this patch. Something like:

```
Wallpaper.getCurrentWallpaperFromSettings(settings)?.let {
    store.dispatch(AppAction.WallpaperAction.UpdateCurrentWallpaper(it))
}
val currentWallpaperName = withContext(Dispatchers.IO) { settings.currentWallpaperName }
if (settings.requiresWallpaperMigration) {
    migrationHelper.migrateLegacyWallpaper(currentWallpaperName)
}
```

Once the wallpaper is migrated, it should get picked up automatically as part of `possibleWallpapers` or `lookupExpiredWallpaper`. 

My thinking is that this change will make it easier to remove the migration code in the future. Also, it should make it more clear that the migration code is only running once, rather than relying on the conditional in the migration code that looks like `if (!legacyLandscapeFile.exists() || !legacyPortraitFile.exists())`. This will also skip the IO of looking for the legacy files.

MatthewTighe (approved):
One more small comment, but this is otherwise looking good to me! I'll approve now so it can be landed during your next day if you'd like

MatthewTighe:
:shipit:

app/src/main/java/org/mozilla/fenix/wallpapers/LegacyWallpaperMigration.kt
```diff
@@ -0,0 +1,71 @@
...
+                )
+            )
+        } catch (e: IOException) {
+            Logger.error("Failed to migrate legacy wallpaper", e)
+        }
+
+        // Delete the remaining legacy files
+        File(storageRootDirectory, "wallpapers/portrait").deleteRecursively()
+        File(storageRootDirectory, "wallpapers/landscape").deleteRecursively()
+    }
```
MatthewTighe, line 70: There is one more special case that I would like to cover here if you don't mind:

If the wallpaper has been successfully migrated then:

```
if (wallpaperName == "mei" || wallpaperName == "panda") {
    settings.currentWallpaperTextColor = "FFFBFBFE".toLong(radix = 16)
}
```

this will allow us to use an appropriate text color if a user is still using one of the expired Turning Red wallpapers.

MatthewTighe (approved):
One last minute comment based on UX feedback. This is otherwise blocking https://github.com/mozilla-mobile/fenix/pull/26967 so please land it soon if you are feeling comfortable with it

Mugurell:
Can help with getting this landable while Alex is in PTO.
- addressing https://github.com/mozilla-mobile/fenix/pull/26685#discussion_r970088593
- adding trailing commas

Mugurell:
Added the trailing command in the first commit, and support for the migrated Turning Red wallpapers in the second commit.

Issue #26737, opened 2022-08-31
Update typography for the Pocket section


Mugurell:
Following a recent discussion from https://github.com/mozilla-mobile/fenix/pull/26451#pullrequestreview-1073013007 @Kate-Galetski helped standardize the typography used for the Pocket section also.

The new design is available at https://www.figma.com/file/d7l1iGabQYqnLcnQbHDGzM/Wallpaper-2.0-%2B-Onboarding-v106?node-id=5572%3A197382

Typography details:
- [ ] All texts will use the `caption` style
- [x] The main is colored in: `textPrimary`
- [x] The secondary line is colored in: `textSecondary`

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21438)

delia-pop:
Verified as fixed on Nightly 107.0a1 from 09/20 with Google Pixel 6 (Android 13). The Pocket section matches the design.

![1663692939773](https://user-images.githubusercontent.com/89388888/191445363-9aa43fc5-77c0-4fcc-bd36-b480884af3f7.JPEG)

PR #26740, opened 2022-08-31
For #26737: Update the Pocket section typography

Mugurell:
This PR also adds `testTag`s to better support ui testing.

image

### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation

Used by GitHub Actions.


app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesComposables.kt
```diff
@@ -116,10 +116,11 @@ fun PocketStory(
         title = {
             Text(
                 text = story.title,
+                modifier = Modifier.testTag("homescreen.pocket.story.title"),
                 color = FirefoxTheme.colors.textPrimary,
-                fontSize = 14.sp,
                 overflow = TextOverflow.Ellipsis,
                 maxLines = 2,
+                style = FirefoxTheme.typography.caption,
```
gabrielluong, line 123: It seems odd to me that we would use Caption for the story header.

app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesComposables.kt
```diff
@@ -416,17 +423,29 @@ fun PoweredByPocketHeader(
                 Text(
                     text = stringResource(R.string.pocket_stories_feature_title),
                     color = FirefoxTheme.colors.textPrimary,
-                    fontSize = 12.sp,
-                    lineHeight = 16.sp
+                    style = FirefoxTheme.typography.caption
                 )
 
-                ClickableSubstringLink(
```
gabrielluong, line 423: Do we need to keep `ClickableSubstringLink` after this?

Mugurell: Left it in the project although not used now in the idea that it is meant to be reusable and may be useful later.
It makes more sense to use it when there is a string with only one part that is clickable - seeing that in this patch I'm using two strings and hardcoding a space character in between which to me seems uglier and riskier.
Can remove it in the idea that that is problematic for ui tests.

app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesComposables.kt
Mugurell: Thank you.
Discussed with UX who updated the design to use `body2` for titles and `caption` elsewhere.

app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesComposables.kt
```diff
@@ -414,8 +413,7 @@ fun PoweredByPocketHeader(
                 Text(
                     text = stringResource(R.string.pocket_stories_feature_title),
                     color = textColor,
-                    fontSize = 12.sp,
-                    lineHeight = 16.sp
+                    style = FirefoxTheme.typography.caption
```
gabrielluong, line 416:
```suggestion
                    style = FirefoxTheme.typography.caption,
```
I think this is fine, but I am not too sure about it since the designs don't technically specify `caption` text style here, and are a bit of outdate since we can see it using the Inter font. You can proceed with landing it as is or double check with UX that this bit of text should also be `caption`.

gabrielluong (approved):

app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesComposables.kt
Mugurell: Asked for UX feedback.

Mugurell: Crystal confirmed `caption` is right.

PR #26757, opened 2022-09-01
For #26756 - Don't show Jump Back In CFR if the synced tab CFR is shown

### LINKED ISSUE #26756: Don't show Jump Back In CFR if the synced tab CFR is shown
Based on https://mozilla-hub.atlassian.net/browse/FNXV2-20988:
A Jump Back In pop up message will be displayed on the homepage under the following the conditions:
- Should not be shown if the synced tab pickup message is already showing or if the user has already seen the synced tab pickup message

We currently show the Jump back in CFR if the synced tab CFR is shown.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21443)

gabrielluong:
### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation
Fixes #26756

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/onboarding/JumpBackInCFRDialog.kt
+++ b/app/src/main/java/org/mozilla/fenix/onboarding/JumpBackInCFRDialog.kt
@@ -14,6 +14,7 @@ import android.view.View
 import androidx.recyclerview.widget.RecyclerView
 import org.mozilla.fenix.databinding.OnboardingJumpBackInCfrBinding
 import org.mozilla.fenix.ext.settings
+import org.mozilla.fenix.home.recentsyncedtabs.view.RecentSyncedTabViewHolder
 import org.mozilla.fenix.home.recenttabs.view.RecentTabsHeaderViewHolder
 
 /**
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mozilla/fenix/onboarding/JumpBackInCFRDialog.kt
```diff
@@ -50,12 +50,15 @@ class JumpBackInCFRDialog(val recyclerView: RecyclerView) {
 
     private fun createJumpCRF(anchor: View): Dialog? {
         val context: Context = recyclerView.context
-        if (!context.settings().showSyncCFR) {
+
+        if (context.settings().showSyncCFR) {
```
Alexandru2909, line 54: I may have gotten this wrong, but assuming the `showSyncCFR` setting is enabled and no recent synced tab has been found on the first homescreen interaction, wouldn't this make the jump back in CFR not be displayed too (I believe this is the only case when we want to show the jump back in CFR)?

Mugurell: Seems like the code gets a little tangled with all these settings for other CFRs that one CFR should't know or care about. If another CFR is added things might get unmaintainable with complex checks in every CFR & easy to miss details.
I'd like to bring into discussion the cfr manager approach that would be the only one knowing about and controlling all CFRs. It can be the one central point that knows when one is actually displayed to not show others and it can easily control the order in which they are shown.
Can also help with this.

grigoryk: And now yet another CFR is sneaking in! https://github.com/mozilla-mobile/fenix/pull/26758/files#diff-f6968644d19314ed1385ed1bef14d8a3aaa163999143e11b0449b94292e6be56

this is getting officially confusing to read at a glance.

gabrielluong: Discussed with @Mugurell to followup with adding a manager to control all the CFRs that shows up. However, I don't believe we have a bug here since `createJumpCRF` is only called after it finds a view to attach to.

app/src/main/java/org/mozilla/fenix/onboarding/SyncCFRPresenter.kt
```diff
@@ -51,12 +52,15 @@ class SyncCFRPresenter(
...
-            ) {
-            }.apply {
+            ).apply {
                 syncCFR = this
                 show()
-                Onboarding.synCfrShown.record(NoExtras())
             }
+
+            context.settings().showSyncCFR = false
+            context.settings().shouldShowJumpBackInCFR = false
```
sarah541, line 61: Why should `shouldShowJumpBackInCFR` be false here?

sarah541 (approved):

app/src/main/java/org/mozilla/fenix/onboarding/SyncCFRPresenter.kt
gabrielluong: This is part of the acceptance criteria https://github.com/mozilla-mobile/fenix/issues/26756#issue-1358275594
Jump back in “Should not be shown if the synced tab pickup message is already showing or if the user has already seen the synced tab pickup message”

PR #26758, opened 2022-09-01
Closes #26750: add Wallpapers Onboarding dialog

### LINKED ISSUE #26750: Create a simple wallpaper dialog
Breaking down [this](https://github.com/mozilla-mobile/fenix/issues/26216) task, while the different dependencies are being cleared up.

Create a wallpaper selector, that follows these conditions:
-         Tab Pickup and Jump Back In CFRs are not currently being displayed
-         The user has not seen the wallpaper selection tool before

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21441)

mavduevskiy:
Closes #26750

Screenshot_20220831_174940
Screenshot_20220831_174940

### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation
Fixes #26750

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlInteractor.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlInteractor.kt
@@ -169,7 +169,7 @@ interface OnboardingInteractor {
      * Returns true if the call has been passed down to the controller.
      *
      * @param state The wallpaper state.
-     * @return Whether the onboarding dialog is currently shown
+     * @return Whether the onboarding dialog is currently shown.
      */
     fun showWallpapersOnboardingDialog(state: WallpaperState): Boolean
 }
--- a/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlInteractor.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlInteractor.kt
@@ -27,6 +27,7 @@ import org.mozilla.fenix.home.recentvisits.RecentlyVisitedItem.RecentHistoryGrou
 import org.mozilla.fenix.home.recentvisits.RecentlyVisitedItem.RecentHistoryHighlight
 import org.mozilla.fenix.home.recentvisits.controller.RecentVisitsController
 import org.mozilla.fenix.home.recentvisits.interactor.RecentVisitsInteractor
+import org.mozilla.fenix.wallpapers.WallpaperState
 
 /**
  * Interface for tab related actions in the [SessionControlInteractor].
@@ -162,6 +163,15 @@ interface OnboardingInteractor {
      * Opens a custom tab to privacy notice url. Called when a user clicks on the "read our privacy notice" button.
      */
     fun onReadPrivacyNoticeClicked()
+
+    /**
+     * Show Wallpapers onboarding dialog to onboard users about the feature if conditions are met.
+     * Returns true if the call has been passed down to the controller.
+     *
+     * @param state The wallpaper state.
+     * @return Whether the onboarding dialog is currently shown
+     */
+    fun showWallpapersOnboardingDialog(state: WallpaperState): Boolean
 }
 
 interface CustomizeHomeIteractor {
@@ -320,6 +330,10 @@ class SessionControlInteractor(
 
     override fun onReadPrivacyNoticeClicked() {
         controller.handleReadPrivacyNoticeClicked()
+    }
+
+    override fun showWallpapersOnboardingDialog(state: WallpaperState): Boolean {
+        return controller.handleShowWallpapersOnboardingDialog(state)
     }
 
     override fun onToggleCollectionExpanded(collection: TabCollection, expand: Boolean) {
--- a/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlView.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlView.kt
@@ -199,6 +199,9 @@ class SessionControlView(
 
     val view: RecyclerView = containerView as RecyclerView
 
+    // We want to limit feature recommendations to one per HomePage visit.
+    var featureRecommended = false
+
     private val sessionControlAdapter = SessionControlAdapter(
         interactor,
         viewLifecycleOwner,
@@ -206,21 +209,33 @@ class SessionControlView(
     )
 
     init {
+        @Suppress("NestedBlockDepth")
         view.apply {
             adapter = sessionControlAdapter
             layoutManager = object : LinearLayoutManager(containerView.context) {
                 override fun onLayoutCompleted(state: RecyclerView.State?) {
                     super.onLayoutCompleted(state)
 
-                    if (!context.settings().showHomeOnboardingDialog && (
-                        context.settings().showSyncCFR ||
-                            context.settings().shouldShowJumpBackInCFR
-                        )
-                    ) {
-                        HomeCFRPresenter(
-                            context = context,
-                            recyclerView = view,
-                        ).show()
+                    if (!featureRecommended && !context.settings().showHomeOnboardingDialog) {
+                        if (!context.settings().showHomeOnboardingDialog && (
+                                    context.settings().showSyncCFR ||
+                                            context.settings().shouldShowJumpBackInCFR
+                                    )
+                        ) {
+                            featureRecommended = HomeCFRPresenter(
+                                context = context,
+                                recyclerView = view,
+                            ).show()
+                        }
+
+                        if (!context.settings().shouldShowJumpBackInCFR &&
+                            context.settings().showWallpaperOnboarding &&
+                            !featureRecommended
+                        ) {
+                            featureRecommended = interactor.showWallpapersOnboardingDialog(
+                                context.components.appStore.state.wallpaperState
+                            )
+                        }
                     }
 
                     // We want some parts of the home screen UI to be rendered first if they are
--- a/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlView.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlView.kt
@@ -218,9 +218,9 @@ class SessionControlView(
 
                     if (!featureRecommended && !context.settings().showHomeOnboardingDialog) {
                         if (!context.settings().showHomeOnboardingDialog && (
-                                    context.settings().showSyncCFR ||
-                                            context.settings().shouldShowJumpBackInCFR
-                                    )
+                            context.settings().showSyncCFR ||
+                                context.settings().shouldShowJumpBackInCFR
+                            )
                         ) {
                             featureRecommended = HomeCFRPresenter(
                                 context = context,
@@ -233,7 +233,7 @@ class SessionControlView(
                             !featureRecommended
                         ) {
                             featureRecommended = interactor.showWallpapersOnboardingDialog(
-                                context.components.appStore.state.wallpaperState
+                                context.components.appStore.state.wallpaperState,
                             )
                         }
                     }
--- a/app/src/main/java/org/mozilla/fenix/onboarding/HomeCFRPresenter.kt
+++ b/app/src/main/java/org/mozilla/fenix/onboarding/HomeCFRPresenter.kt
@@ -38,13 +38,17 @@ class HomeCFRPresenter(
      * Determine the CFR to be shown on the Home screen and show a CFR for the resultant view
      * if any.
      */
-    fun show() {
-        when (val result = getCFRToShow()) {
-            is Result.SyncedTab -> showSyncedTabCFR(view = result.view)
-            is Result.JumpBackIn -> showJumpBackInCFR(view = result.view)
-            else -> {
-                // no-op
+    fun show(): Boolean {
+        return when (val result = getCFRToShow()) {
+            is Result.SyncedTab -> {
+                showSyncedTabCFR(view = result.view)
+                true
             }
+            is Result.JumpBackIn -> {
+                showJumpBackInCFR(view = result.view)
+                true
+            }
+            else -> false
         }
     }
 
--- a/app/src/main/java/org/mozilla/fenix/onboarding/WallpaperOnboardingDialogFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/onboarding/WallpaperOnboardingDialogFragment.kt
@@ -1,0 +1,97 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.onboarding
+
+import android.annotation.SuppressLint
+import android.content.pm.ActivityInfo
+import android.os.Bundle
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import androidx.compose.material.ExperimentalMaterialApi
+import androidx.compose.runtime.rememberCoroutineScope
+import androidx.compose.ui.platform.ComposeView
+import androidx.compose.ui.platform.ViewCompositionStrategy
+import androidx.navigation.fragment.findNavController
+import com.google.android.material.bottomsheet.BottomSheetDialogFragment
+import kotlinx.coroutines.launch
+import mozilla.components.lib.state.ext.observeAsComposableState
+import org.mozilla.fenix.NavGraphDirections
+import org.mozilla.fenix.R
+import org.mozilla.fenix.ext.requireComponents
+import org.mozilla.fenix.ext.settings
+import org.mozilla.fenix.theme.FirefoxTheme
+import org.mozilla.fenix.wallpapers.Wallpaper
+import org.mozilla.fenix.wallpapers.WallpaperOnboarding
+
+/**
+ * Dialog displaying the wallpapers onboarding.
+ */
+@OptIn(ExperimentalMaterialApi::class)
+class WallpaperOnboardingDialogFragment : BottomSheetDialogFragment() {
+    private val appStore by lazy {
+        requireComponents.appStore
+    }
+
+    private val wallpaperUseCases by lazy {
+        requireComponents.useCases.wallpaperUseCases
+    }
+
+    @SuppressLint("SourceLockedOrientationActivity")
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setStyle(STYLE_NO_TITLE, R.style.WallpaperOnboardingDialogStyle)
+        activity?.requestedOrientation = ActivityInfo.SCREEN_ORIENTATION_PORTRAIT
+    }
+
+    override fun onDestroy() {
+        super.onDestroy()
+        activity?.requestedOrientation = ActivityInfo.SCREEN_ORIENTATION_UNSPECIFIED
+    }
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+        requireContext().settings().showWallpaperOnboarding = false
+    }
+
+    override fun onCreateView(
+        inflater: LayoutInflater,
+        container: ViewGroup?,
+        savedInstanceState: Bundle?
+    ): View = ComposeView(requireContext()).apply {
+        setViewCompositionStrategy(ViewCompositionStrategy.DisposeOnViewTreeLifecycleDestroyed)
+
+        setContent {
+            FirefoxTheme {
+                val wallpapers = appStore.observeAsComposableState { state ->
+                    state.wallpaperState.availableWallpapers.subList(0, THUMBNAILS_COUNT)
+                }.value ?: listOf()
+                val currentWallpaper = appStore.observeAsComposableState { state ->
+                    state.wallpaperState.currentWallpaper
+                }.value ?: Wallpaper.Default
+
+                val coroutineScope = rememberCoroutineScope()
+
+                WallpaperOnboarding(
+                    wallpapers = wallpapers,
+                    currentWallpaper = currentWallpaper,
+                    onCloseClicked = { dismiss() },
+                    onBottomButtonClicked = {
+                        val directions = NavGraphDirections.actionGlobalWallpaperSettingsFragment()
+                        findNavController().navigate(directions)
+                    },
+                    loadWallpaperResource = { wallpaperUseCases.loadThumbnail(it) },
+                    onSelectWallpaper = {
+                        coroutineScope.launch { wallpaperUseCases.selectWallpaper(it) }
+                    }
+                )
+            }
+        }
+    }
+
+    companion object {
+        const val THUMBNAILS_COUNT = 6
+    }
+}
--- a/app/src/main/java/org/mozilla/fenix/onboarding/WallpaperOnboardingDialogFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/onboarding/WallpaperOnboardingDialogFragment.kt
@@ -66,7 +66,7 @@ class WallpaperOnboardingDialogFragment : BottomSheetDialogFragment() {
         setContent {
             FirefoxTheme {
                 val wallpapers = appStore.observeAsComposableState { state ->
-                    state.wallpaperState.availableWallpapers.subList(0, THUMBNAILS_COUNT)
+                    state.wallpaperState.availableWallpapers.take(THUMBNAILS_COUNT)
                 }.value ?: listOf()
                 val currentWallpaper = appStore.observeAsComposableState { state ->
                     state.wallpaperState.currentWallpaper
@@ -78,7 +78,7 @@ class WallpaperOnboardingDialogFragment : BottomSheetDialogFragment() {
                     wallpapers = wallpapers,
                     currentWallpaper = currentWallpaper,
                     onCloseClicked = { dismiss() },
-                    onBottomButtonClicked = {
+                    onExploreMoreButtonClicked = {
                         val directions = NavGraphDirections.actionGlobalWallpaperSettingsFragment()
                         findNavController().navigate(directions)
                     },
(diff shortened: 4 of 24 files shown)


mavduevskiy:
It looks like the dialog fragment breaks a lot (29) of tests. Should I adjust conditions of displaying somehow, or should I fix the tests?

gabrielluong:
> It looks like the dialog fragment breaks a lot (29) of UI tests. Should I adjust conditions of displaying somehow, or should I fix the tests?

It's likely that the wallpaper dialog is showing up in the tests and interrupting their flow. You should ensure that the wallpaper dialog is shown up under the right conditions and you can also add functions to disable the feature/preference by creating a function in https://github.com/mozilla-mobile/fenix/blob/main/app/src/androidTest/java/org/mozilla/fenix/helpers/FeatureSettingsHelper.kt and call it in the affected tests setUp’s like https://github.com/mozilla-mobile/fenix/blob/main/app/src/androidTest/java/org/mozilla/fenix/ui/NavigationToolbarTest.kt#L46

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlView.kt
```diff
@@ -213,16 +217,24 @@ class SessionControlView(
...
-                            JumpBackInCFRDialog(view).showIfNeeded()
+                            featureRecommended = JumpBackInCFRDialog(view).showIfNeeded()
+                        }
+
+                        if (!context.settings().shouldShowJumpBackInCFR &&
+                            context.settings().showWallpaperOnboarding &&
+                            !featureRecommended
+                        ) {
+                            featureRecommended =
+                                interactor.showWallpapersOnboardingDialog(this@SessionControlView.state)
```
MatthewTighe, line 237: Rather than add new state to track whether a piece of onboarding is currently being shown, we could use a `when` statement to simplify this logic
```
when {
    context.settings().showSyncCFR -> SyncCFRPresenter(
        context = context,
        recyclerView = view,
    ).showSyncCFR()
    context.settings().shouldShowJumpBackInCFR -> JumpBackInCFRDialog(view).showIfNeeded()
    context.settings().showWallpaperOnboarding -> interactor.showWallpapersOnboardingDialog(this@SessionControlView.state)
}
```

I think this would also require `showWallpapersOnboardingDialog` to update `Settings.showWallpaperOnboarding`

app/src/main/java/org/mozilla/fenix/onboarding/JumpBackInCFRDialog.kt
```diff
@@ -27,7 +27,7 @@ class JumpBackInCFRDialog(val recyclerView: RecyclerView) {
     /**
      * Try to show the crf dialog if it hasn't been shown before.
      */
-    fun showIfNeeded() {
+    fun showIfNeeded(): Boolean {
```
MatthewTighe, line 30: This change could be reverted if the suggestion to switch to a `when` statement is taken.

app/src/main/java/org/mozilla/fenix/onboarding/SyncCFRPresenter.kt
```diff
@@ -37,7 +37,7 @@ class SyncCFRPresenter(
     /**
      * Find the synced view and if it is available, then show the synced tab CFR.
      */
-    fun showSyncCFR() {
+    fun showSyncCFR(): Boolean {
```
MatthewTighe, line 40: This change could be reverted if the suggestion to switch to a when statement is taken.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlInteractor.kt
```diff
@@ -322,6 +330,21 @@ class SessionControlInteractor(
         controller.handleReadPrivacyNoticeClicked()
     }
 
+    override fun showWallpapersOnboardingDialog(state: AppState?): Boolean {
```
MatthewTighe, line 333: I think `state` could be non-null, especially if the suggestion to switch to a when statement is `SessionControlView` is taken.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlInteractor.kt
```diff
@@ -322,6 +330,21 @@ class SessionControlInteractor(
...
+            wallpapers.filter { wallpaper ->
+                wallpaper.thumbnailFileState == Wallpaper.ImageFileState.Downloaded
+            }.size.let { downloadedCount ->
+                downloadedCount >= WallpaperOnboardingDialogFragment.THUMBNAILS_COUNT &&
+                    downloadedCount == wallpapers.size
+            }.also { isThumbnailsDownloaded ->
+                if (isThumbnailsDownloaded) {
+                    controller.handleShowWallpapersOnboardingDialog()
+                }
+            }
```
MatthewTighe, line 344: I think it would make more sense to move this logic into `SessionControlController`. This suggestion may help with readability as well

```suggestion
if (wallpapers.size >= WallpaperOnboardingDialogFragment.THUMBNAILS_COUNT && wallpapers.all { it.thumbnailFileState == Wallpaper.ImageFileState.Downloaded }) {
   controller.handleShowWallpapersOnboardingDialog()
}
```

MatthewTighe (commented):
First pass: this is looking good so far! I think it's pretty close, just had a few comments to share. Everything seems to work correctly when I try it locally

app/src/main/java/org/mozilla/fenix/onboarding/WallpaperOnboardingDialogFragment.kt
```diff
@@ -0,0 +1,97 @@
...
+    @SuppressLint("SourceLockedOrientationActivity")
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setStyle(STYLE_NO_TITLE, R.style.WallpaperOnboardingDialogStyle)
+        activity?.requestedOrientation = ActivityInfo.SCREEN_ORIENTATION_PORTRAIT
+    }
+
+    override fun onDestroy() {
+        super.onDestroy()
+        activity?.requestedOrientation = ActivityInfo.SCREEN_ORIENTATION_UNSPECIFIED
```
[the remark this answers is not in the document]
mavduevskiy: I copied it from [HomeOnboardingDialogFragment](https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/java/org/mozilla/fenix/onboarding/HomeOnboardingDialogFragment.kt#34), but after some testing I've noticed that [DialogStyleBase](https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/res/values/styles.xml#410) already forces all Dialogs to pauses fullScreen mode. I suppose it is the expected behavior.

app/src/main/java/org/mozilla/fenix/onboarding/WallpaperOnboardingDialogFragment.kt
```diff
@@ -0,0 +1,97 @@
...
+        container: ViewGroup?,
+        savedInstanceState: Bundle?
+    ): View = ComposeView(requireContext()).apply {
+        setViewCompositionStrategy(ViewCompositionStrategy.DisposeOnViewTreeLifecycleDestroyed)
+
+        setContent {
+            FirefoxTheme {
+                val wallpapers = appStore.observeAsComposableState { state ->
+                    state.wallpaperState.availableWallpapers
+                }.value ?: listOf()
```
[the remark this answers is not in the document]
mavduevskiy: I'll limit to `THUMBNAILS_COUNT` for now, will wait till `WallpaperSettingsFragment` would be updated with filtering.

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperOnboarding.kt
```diff
@@ -0,0 +1,135 @@
...
+ */
+
+@Suppress("LongParameterList")
+@ExperimentalMaterialApi
+@Composable
+fun WallpaperOnboarding(
+    wallpapers: List<Wallpaper>,
+    currentWallpaper: Wallpaper,
+    onCloseClicked: () -> Unit,
+    onButtonClicked: () -> Unit,
```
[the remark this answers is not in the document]
mavduevskiy: My thought process was there is just one button and the Yagni principle, but I'll update with `onBottomButtonClicked`

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperOnboarding.kt
```diff
@@ -0,0 +1,135 @@
...
+            Text(
+                text = stringResource(R.string.wallpapers_onboarding_dialog_title_text),
+                color = FirefoxTheme.colors.textPrimary,
+                fontWeight = FontWeight.Bold,
+                fontSize = 16.sp,
+                overflow = TextOverflow.Ellipsis,
+                maxLines = 1,
+            )
+            Spacer(modifier = Modifier.height(4.dp))
+            Text(
```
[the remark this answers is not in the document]
mavduevskiy: fair, changed :)

app/src/main/res/values/styles.xml
```diff
@@ -564,6 +564,10 @@
     <style name="CreateCollectionDialogStyle" parent="DialogStyleBase"/>
     <style name="CreateShortcutDialogStyle" parent="DialogStyleBase"/>
     <style name="HomeOnboardingDialogStyle" parent="DialogStyleBase"/>
+    <style name="WallpaperOnboardingDialogStyle" parent="DialogStyleBase">
+        <item name="android:windowIsFloating">true</item>
+        <item name="android:backgroundDimAmount">0.02</item>
```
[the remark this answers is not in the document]
mavduevskiy: `android:windowIsFloating` should be true for the desired look. `DialogStyleBase` has the same parameter, but it is set to false.

true|false
---|---
tabs_old_no_lb|tabs_new_no_lb

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlView.kt
```diff
@@ -192,13 +192,17 @@ private fun collectionTabItems(collection: TabCollection) =
...
 class SessionControlView(
     containerView: View,
     viewLifecycleOwner: LifecycleOwner,
     private val interactor: SessionControlInteractor,
 ) {
 
     val view: RecyclerView = containerView as RecyclerView
+    // We want to limit feature recommendations to one per HomePage visit.
+    var featureRecommended = false
+    var state: AppState? = null
```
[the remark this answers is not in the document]
mavduevskiy: fair! fixed

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlView.kt
mavduevskiy: I initially took the same approach, but it didn't really work as we want it to. the `onLayoutCompleted` might be called multiple times during one `HomeFragment` lifecycle, and we need to keep the one `CFR` per `HomeFragment` lunch rule. There is a case: jump back CFR is shown, wallpaper isn't, synced isn't. If user signed in and got back to home page, he will see wallpapers first, then the adaper will get a new synced tab, and the synced CFR will get shown.

Also, having a parameter per lifecycle is a pretty clear way to communicate the product intent not to attack the users with prompts too much.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlInteractor.kt
mavduevskiy: good point!

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlView.kt
MatthewTighe: Okay, thanks for walking me through it. I'm cool with keeping it as is in that case. I wonder if maybe `isOnboardingShown` or something similar might be a more clear name, though

app/src/main/java/org/mozilla/fenix/onboarding/WallpaperOnboardingDialogFragment.kt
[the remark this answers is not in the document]
MatthewTighe: I'm fine with waiting to impose the limit along with the filtering, if you'd prefer. Up to you

app/src/main/res/values/styles.xml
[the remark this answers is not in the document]
MatthewTighe: Okay cool, thanks

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlInteractor.kt
mavduevskiy: 👍

mavduevskiy:
After rebase, lint gone berserk, trying to change 931 files. Investigating

mcarare:
@mavduevskiy After https://github.com/mozilla-mobile/fenix/pull/26846 landed, a baseline file for ktlint was added.
After rebasing, the code in your PR either:

 - introduced code that does not follow ktlint rules
 - introduced lines of code in files referenced in basline file, making the references (line, row) useless.
 
In this PR, you can either fix issues raised by ktlint or change the references in ktlint basline file. I would recommend the first option.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlInteractor.kt
```diff
@@ -162,6 +163,13 @@ interface OnboardingInteractor {
      * Opens a custom tab to privacy notice url. Called when a user clicks on the "read our privacy notice" button.
      */
     fun onReadPrivacyNoticeClicked()
+
+    /**
+     * Show Wallpapers onboarding dialog to onboard users about the feature if conditions are met.
+     * Returns true if the call has been passed down to the controller.
+     * @param state The wallpaper state.
```
MatthewTighe, line 170: This comment about returns could be formatted as an `@return` doc string. It may also be helpful to specify what the boolean return means, something like:

`@return Whether the onboarding dialog is currently shown`

MatthewTighe (approved):
Few more small comments, but this is otherwise looking ready to land to me!

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperOnboarding.kt
```diff
@@ -0,0 +1,142 @@
...
+
+@Suppress("LongParameterList")
+@ExperimentalMaterialApi
+@Composable
+fun WallpaperOnboarding(
+    wallpapers: List<Wallpaper>,
+    currentWallpaper: Wallpaper,
+    onCloseClicked: () -> Unit,
+    onBottomButtonClicked: () -> Unit,
+    loadWallpaperResource: suspend (Wallpaper) -> Bitmap?,
```
gabrielluong, line 55: Can we reorder this to be above `onCloseClicked`? We want to group all the event handlers together ideally.

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperOnboarding.kt
```diff
@@ -0,0 +1,142 @@
...
+ */
+
+@Suppress("LongParameterList")
+@ExperimentalMaterialApi
+@Composable
+fun WallpaperOnboarding(
+    wallpapers: List<Wallpaper>,
+    currentWallpaper: Wallpaper,
+    onCloseClicked: () -> Unit,
+    onBottomButtonClicked: () -> Unit,
```
gabrielluong, line 54:
```suggestion
    onExploreMoreButtonClicked: () -> Unit,
```
We could always be more descriptive with our naming here.

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperOnboarding.kt
```diff
@@ -0,0 +1,142 @@
...
+                    .align(Alignment.End)
+            )
+
+            Spacer(modifier = Modifier.height(8.dp))
+
+            Text(
+                text = stringResource(R.string.wallpapers_onboarding_dialog_title_text),
+                color = FirefoxTheme.colors.textPrimary,
+                fontWeight = FontWeight.Bold,
+                fontSize = 16.sp,
```
gabrielluong, line 82: Do we have a matching `FenixTypography` that we can just utilize from the designs? https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/java/org/mozilla/fenix/theme/FenixTypography.kt

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlView.kt
```diff
@@ -191,6 +191,7 @@ private fun collectionTabItems(collection: TabCollection) =
  * @property interactor [SessionControlInteractor] which will have delegated to all user
  * interactions.
  */
+@Suppress("NestedBlockDepth")
```
gabrielluong, line 194: Perhaps move before the init() block

app/src/main/java/org/mozilla/fenix/onboarding/HomeCFRPresenter.kt
```diff
@@ -38,14 +38,16 @@ class HomeCFRPresenter(
...
+    fun show(): Boolean {
+        val result = getCFRToShow()
+        when (result) {
             is Result.SyncedTab -> showSyncedTabCFR(view = result.view)
             is Result.JumpBackIn -> showJumpBackInCFR(view = result.view)
             else -> {
                 // no-op
             }
         }
+        return result != Result.None
```
gabrielluong, line 50:
```suggestion
        return when (val result = getCFRToShow()) {
            is Result.SyncedTab -> { 
               showSyncedTabCFR(view = result.view)
               true
            }
            is Result.JumpBackIn -> {
                showJumpBackInCFR(view = result.view)
                true
            }
            else -> false
        }
```
I would also consider writing it like this so we can get rid of the no-op

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlView.kt
```diff
@@ -212,15 +216,26 @@ class SessionControlView(
...
-                    if (!context.settings().showHomeOnboardingDialog && (
-                        context.settings().showSyncCFR ||
-                            context.settings().shouldShowJumpBackInCFR
-                        )
-                    ) {
-                        HomeCFRPresenter(
-                            context = context,
-                            recyclerView = view,
-                        ).show()
+                    if (!featureRecommended && !context.settings().showHomeOnboardingDialog) {
```
gabrielluong, line 219: Wouldn't `featureRecommended` always be false anyways when this class is created? I think we can avoid this check and simply declare `featureRecommended` inside the if block

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlInteractor.kt
gabrielluong, line 170: nit: Add a new line above to separate the description and @param

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -151,34 +151,38 @@ private fun WallpaperSnackbar(
...
-private fun WallpaperThumbnails(
+fun WallpaperThumbnails(
     wallpapers: List<Wallpaper>,
     defaultWallpaper: Wallpaper,
     loadWallpaperResource: suspend (Wallpaper) -> Bitmap?,
     selectedWallpaper: Wallpaper,
     numColumns: Int = 3,
     onSelectWallpaper: (Wallpaper) -> Unit,
+    verticalPadding: Int = 30,
+    horizontalPadding: Int = 20
```
gabrielluong, line 162: nit: Missing @param

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperOnboarding.kt
gabrielluong: This can be fixed as a followup

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperOnboarding.kt
mavduevskiy: but in that case we create coupling between a string resource and the parameter naming. it might go out of sync too easily

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlView.kt
mavduevskiy: yes, it will be. but `onLayoutCompleted` is being called multiple times during one lifecycle, in cases like adding items, keyboard, orientation change, and it would launch multiple CFR during one homescreen lifecycle if we don't keep track of CFRs being actually displayed during a single fragment lifecycle

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperOnboarding.kt
MozillaNoah: I think this is an assumed risk for the benefit of having clear documentation/APIs. `InactiveTabsList` has similar lamdbas that are directly tied to their use cases & user-facing descriptions (e.g. `onDeleteAllButtonClick`, `onAutoCloseDismissClick`, etc.). Worst case, the lambda's name gets refactored if the button's text is updated and significantly deviates from the original wording.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlView.kt
MatthewTighe: I think in general there is a future opportunity to solidify our work around when different onboarding items are shown. That said, I think this solution would work for now given the details @mavduevskiy has listed and our time constraints.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -147,27 +146,37 @@ private fun WallpaperSnackbar(
...
-private fun WallpaperThumbnails(
+fun WallpaperThumbnails(
     wallpapers: List<Wallpaper>,
     defaultWallpaper: Wallpaper,
     loadWallpaperResource: suspend (Wallpaper) -> Bitmap?,
     selectedWallpaper: Wallpaper,
     numColumns: Int = 3,
     onSelectWallpaper: (Wallpaper) -> Unit,
+    verticalPadding: Int = 30,
+    horizontalPadding: Int = 20,
```
MatthewTighe, line 162: nit: while we're here, we can group the lambdas and the parameters that have defaults

```suggestion
fun WallpaperThumbnails(
    wallpapers: List<Wallpaper>,
    defaultWallpaper: Wallpaper,
    selectedWallpaper: Wallpaper,
    loadWallpaperResource: suspend (Wallpaper) -> Bitmap?,
    onSelectWallpaper: (Wallpaper) -> Unit,
    numColumns: Int = 3,
    verticalPadding: Int = 30,
    horizontalPadding: Int = 20,
```

MatthewTighe (approved):
Couple more small nits, but I think this is otherwise ready to land. Functionality looks good and as I would expect. Thanks for helping out with this!

I would maybe give @gabrielluong a bit to make sure his concerns are addressed before landing.

gabrielluong:
No need to wait for me.

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperOnboarding.kt
mavduevskiy: okay, three versus one, democracy, changing to `onExploreMoreButtonClicked`

mavduevskiy:
@MatthewTighe 
Updated!

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperOnboarding.kt
```diff
@@ -0,0 +1,141 @@
...
+                modifier = Modifier
+                    .align(Alignment.CenterHorizontally)
+                    .fillMaxWidth(),
+                onClick = { onExploreMoreButtonClicked() },
+            ) {
+                Text(
+                    text = stringResource(R.string.wallpapers_onboarding_dialog_explore_more_button_text),
+                    fontWeight = FontWeight.Bold,
+                    color = FirefoxTheme.colors.textAccent,
+                    fontSize = 14.sp,
```
gabrielluong, line 116: This could be `FirefoxTheme.typography.button`. We shouldn't ever have to define `fontWeight` because our typography is well defined. We also typically encourage following the order of the named parameters that you would see defined by the calling function.

```
@Composable
fun Text(
    text: String,
    color: Color = Color.Unspecified,
    fontSize: TextUnit = TextUnit.Unspecified,
    fontStyle: FontStyle? = null,
    fontWeight: FontWeight? = null,
    fontFamily: FontFamily? = null,
    letterSpacing: TextUnit = TextUnit.Unspecified,
    textDecoration: TextDecoration? = null,
    textAlign: TextAlign? = null,
    lineHeight: TextUnit = TextUnit.Unspecified,
    overflow: TextOverflow = TextOverflow.Clip,
    softWrap: Boolean = true,
    maxLines: Int = Int.MAX_VALUE,
    onTextLayout: (TextLayoutResult) -> Unit = {},
    style: TextStyle = LocalTextStyle.current
)
```
So, we would typically expect `style` to be the last named parameter.

csadilek:
@mavduevskiy  @MatthewTighe ⚠️ We have numerous reports of Nightly (home screen) freezing starting Saturday. I will put up a revert to verify if caused by this change, or https://github.com/mozilla-mobile/fenix/pull/26794.

csadilek:
@mavduevskiy @MatthewTighe We can close the revert if it was the other change.

as nit: Please let's squash commits like these before landing "pr review changes" "add format changes" etc.

csadilek:
@mavduevskiy @MatthewTighe  OK, we didn't have the time to verify which PR caused the problem. It's likely it's https://github.com/mozilla-mobile/fenix/pull/26794 instead, and this can just re-land after discussion on Monday. For now, we just want to get a fixed Nightly out.

PR #26794, opened 2022-09-02
For #26511: Parallelize work for setting the wallpaper

### LINKED ISSUE #26511: [Bug]: Wallpaper loading on launch is delayed

Mugurell:
Split loading the bitmap from storage and actually setting it in two operations
with one that can run in parallel with onCreateView for HomeFragment and one
that can be used serially on the main thread to actually set the wallpaper.

This seems like the best compromise to ensure that everytime the homescreen is
shown it will have the wallpaper set but does affect the performance - there is a
delay in showing HomeFrament to account for waiting for the wallpaper to be set.

### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation
Fixes #26511

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
@@ -213,6 +213,7 @@ class HomeFragment : Fragment() {
         if (shouldEnableWallpaper()) {
             wallpapersObserver = WallpapersObserver(
                 appStore = components.appStore,
+                settings = requireContext().settings(),
                 wallpapersUseCases = components.useCases.wallpaperUseCases,
                 wallpaperImageView = binding.wallpaperImageView,
             ).also {
@@ -412,7 +413,6 @@ class HomeFragment : Fragment() {
         getMenuButton()?.dismissMenu()
 
         if (shouldEnableWallpaper()) {
-            // Setting the wallpaper is a potentially expensive operation - can take 100ms.
             // Running this on the Main thread helps to ensure that the just updated configuration
             // will be used when the wallpaper is scaled to match.
             // Otherwise the portrait wallpaper may remain shown on landscape,
@@ -756,6 +756,12 @@ class HomeFragment : Fragment() {
 
         lifecycleScope.launch(IO) {
             requireComponents.reviewPromptController.promptReview(requireActivity())
+        }
+
+        if (shouldEnableWallpaper()) {
+            runBlockingIncrement {
+                wallpapersObserver?.applyCurrentWallpaper()
+            }
         }
     }
 
--- a/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
@@ -115,6 +115,7 @@ import org.mozilla.fenix.tabstray.TabsTrayAccessPoint
 import org.mozilla.fenix.utils.Settings.Companion.TOP_SITES_PROVIDER_MAX_THRESHOLD
 import org.mozilla.fenix.utils.ToolbarPopupWindow
 import org.mozilla.fenix.utils.allowUndo
+import org.mozilla.fenix.wallpapers.WallpaperManager
 import java.lang.ref.WeakReference
 import kotlin.math.min
 
@@ -412,7 +413,8 @@ class HomeFragment : Fragment() {
 
         getMenuButton()?.dismissMenu()
 
-        if (shouldEnableWallpaper()) {
+        val isDefaultTheCurrentWallpaper = WallpaperManager.isDefaultTheCurrentWallpaper(requireContext().settings())
+        if (shouldEnableWallpaper() && !isDefaultTheCurrentWallpaper) {
             // Running this on the Main thread helps to ensure that the just updated configuration
             // will be used when the wallpaper is scaled to match.
             // Otherwise the portrait wallpaper may remain shown on landscape,
@@ -758,7 +760,8 @@ class HomeFragment : Fragment() {
             requireComponents.reviewPromptController.promptReview(requireActivity())
         }
 
-        if (shouldEnableWallpaper()) {
+        val isDefaultTheCurrentWallpaper = WallpaperManager.isDefaultTheCurrentWallpaper(requireContext().settings())
+        if (shouldEnableWallpaper() && !isDefaultTheCurrentWallpaper) {
             runBlockingIncrement {
                 wallpapersObserver?.applyCurrentWallpaper()
             }
--- a/app/src/main/java/org/mozilla/fenix/home/WallpapersObserver.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/WallpapersObserver.kt
@@ -4,21 +4,27 @@
 
 package org.mozilla.fenix.home
 
+import android.graphics.Bitmap
 import android.widget.ImageView
 import androidx.annotation.VisibleForTesting
+import androidx.annotation.WorkerThread
 import androidx.core.view.isVisible
 import androidx.lifecycle.DefaultLifecycleObserver
 import androidx.lifecycle.LifecycleObserver
 import androidx.lifecycle.LifecycleOwner
+import kotlinx.coroutines.CompletableDeferred
+import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.cancel
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
 import mozilla.components.lib.state.Store
 import org.mozilla.fenix.components.AppStore
 import org.mozilla.fenix.components.appstate.AppAction
 import org.mozilla.fenix.components.appstate.AppState
 import org.mozilla.fenix.ext.scaleToBottomOfView
+import org.mozilla.fenix.utils.Settings
 import org.mozilla.fenix.wallpapers.Wallpaper
 import org.mozilla.fenix.wallpapers.WallpapersUseCases
 
@@ -28,18 +34,44 @@ import org.mozilla.fenix.wallpapers.WallpapersUseCases
  * when the [LifecycleOwner] is destroyed.
  *
  * @param appStore Holds the details abut the current wallpaper.
+ * @param settings Used for checking user's option for what wallpaper to use.
  * @param wallpapersUseCases Used for interacting with the wallpaper feature.
  * @param wallpaperImageView Serves as the target when applying wallpapers.
+ * @param backgroundWorkDispatcher Used for scheduling the wallpaper update when the state is updated
+ * with a new wallpaper.
  */
 class WallpapersObserver(
     private val appStore: AppStore,
+    private val settings: Settings,
     private val wallpapersUseCases: WallpapersUseCases,
     private val wallpaperImageView: ImageView,
+    backgroundWorkDispatcher: CoroutineDispatcher = Dispatchers.IO,
 ) : DefaultLifecycleObserver {
     @VisibleForTesting
     internal var observeWallpapersStoreSubscription: Store.Subscription<AppState, AppAction>? = null
+
+    /**
+     * Coroutine scope for updating the wallpapers when an update is observed.
+     * Allows for easy cleanup when the client of this is destroyed.
+     */
     @VisibleForTesting
-    internal var wallpapersScope = CoroutineScope(Dispatchers.Main.immediate)
+    internal var wallpapersScope = CoroutineScope(backgroundWorkDispatcher)
+
+    /**
+     * Setting the wallpaper assumes two steps:
+     * - load - running on IO
+     * - set - running on Main
+     * This property caches the result of [loadWallpaper] to be later used by [applyCurrentWallpaper].
+     */
+    @Volatile
+    @VisibleForTesting
+    internal var currentWallpaperImage: Bitmap? = null
+
+    /**
+     * Listener for when the first observed wallpaper is loaded and available to be set.
+     */
+    @VisibleForTesting
+    internal val isWallpaperLoaded = CompletableDeferred<Unit>()
 
     init {
         observeWallpaperUpdates()
@@ -49,8 +81,20 @@ class WallpapersObserver(
      * Immediately apply the current wallpaper automatically adjusted to support
      * the current configuration - portrait or landscape.
      */
-    fun applyCurrentWallpaper() {
-        showWallpaper()
+    internal suspend fun applyCurrentWallpaper() {
+        isWallpaperLoaded.await()
+
+        withContext(Dispatchers.Main.immediate) {
+            with(currentWallpaperImage) {
+                when (this) {
+                    null -> wallpaperImageView.isVisible = false
+                    else -> {
+                        scaleToBottomOfView(wallpaperImageView)
+                        wallpaperImageView.isVisible = true
+                    }
+                }
+            }
+        }
     }
 
     override fun onDestroy(owner: LifecycleOwner) {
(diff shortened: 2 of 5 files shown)


Mugurell:
This is the best solution I found to help guarantee that when the `HomeFragment` gets shown it already has a wallpapers set.
Details:
- using `appStore.observeManually` to start observing the store even before `onStart` (which signifies that the screen is visible)
- `load` and `set` the wallpaper in two separate operations to allow them running on separate threads.
	- this allows paralelize the work for observing the store and loading the wallpaper image on an IO thread.
	- expose a ` CompletableDeferred ` to allow knowing when the wallpaper is loaded - the right bitmap is available to use.
	- caching the wallpaper bitmap from IO to make it available to a separate call from the Fragment to set the initial wallpaper.
	- Use ` Dispatchers.Main.immediate ` when setting the wallpaper to HomeFragment's View to avoid a context switch if already on Main.
- have `HomeFragment` block in `onStart` until we know that the wallpaper is available and set.

Other approaches previously used:
- peek into the appstore in `onCreateView` and set the current wallpaper (blocking or not). 
	- caveat 1: the correct wallpaper may not be available at that time.
	- caveat 2: both loading the bitmap and setting it to a View would be done on Main to ensure serial operations (can be avoided)
- rely only on `appstore.flow` to observe wallpapers updates. 
	- major caveat: this would start observing the store after the screen is already shown (without a wallpaper).

Tested the app startup with a slightly modified version of `measure_start_up.py` - `measure_startup_until_onResume` available [here](https://drive.google.com/file/d/1nerNZ4zoCFHOynx90sNb-t_k0sTROXJ7/view?usp=sharing) and an appropriate log above `super.onResume` [here](https://github.com/mozilla-mobile/fenix/blob/df9c60b392d111705209d0f781046f5108557ba6/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt#L789-L790)
(because `cold_main_ff` measures `onCreate` and the other probes measure page loads)
on a Pixel 3 with Android 12 and got the following results:

| | Nighly built locally with the code from `main` | Nightly built locally with the code from this patch |
| --- | --- | --- |
| replicate count: | 100 | 100 |
| max: | 1836 | 2328 |
| mean: | 1555 | 1631 | 
| median: | 1564 | 1640 |
| min: | 1361 | 1339 | 
| link |  [results](https://drive.google.com/file/d/1BGJuripSHwGXB-zmjVUaNj6cNpSFyvf-/view?usp=sharing) | [results](https://drive.google.com/file/d/19MDXZ3-mRmuYM0MAMxmhc-s3bxpX7Zhb/view?usp=sharing) |
| graph | resultsFenixNightlyLocalOnResumeNotWaitingForWallpapers | resultsFenixNightlyLocalOnResumeWaitingForWallpapers |

Mugurell:
Regarding the issue I see two possible approaches: 
- accepting the bug: which is evident also when coming back to homescreen from another screen or 
- accepting that we need more time to load and show the wallpaper - time which this patch tries to reduce to a minimum.

Open to any other suggestions regarding the code or directions.
I would propose trying this in Nightly for a few days to get more results about the real world startup times.

csadilek (commented):
Made a first pass, but I am worried about this additional complexity and the performance penalty. Why is this not a problem in Release (104)? We're loading the wallpapers from disk there too and there's no delay, and also no performance impact to COLD_MAIN.

csadilek:
I guess another way to ask. If we have simpler code in release that doesn't have the problem, can we find out why we regressed and fix it?

Mugurell:
I see the added complexity as necessary and with it being extracted in `WallpaperObserver`, with a straightforward logic there and few reasons to change should be easily maintainable.

Mugurell:
Regarding the difference between the code on Release and the code on `main` the main difference is that on Release we are doing everything serially on the Main thread imperatively, without the MVI loop.
So there is a performance penalty Release is paying already with [`onCreateView` calling](https://github.com/mozilla-mobile/fenix/blob/ccc1d6dfb45a1534d62a596b30f7d65742490bd1/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt#L391) and waiting for [displayWallpaperIfEnabled](https://github.com/mozilla-mobile/fenix/blob/ccc1d6dfb45a1534d62a596b30f7d65742490bd1/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt#L996-L1005). From what I see on my slow device we are waiting for ~60ms before continueing execution in `onCreateView` and unblocking the screen to be shown with the wallpaper set.

Testing in parallel with the current implementation it's easy to see where the bulk of the difference comes from: The added MVI loop. Previously we were doing everything imperatively on Main, now there is a lot more code executed, actions queue, context switching, observing the store that was not happening before and in my testing this pushes the wallpaper load to the end of `onCreateView` or even to `onStart` where we must block to load the wallpaper.

It's important to know that both implementation (Release and `main`) will block and in my tests for around the same time - 60ms (the current Nightly does not and this might have skewed the results) so I'm happy with the end result.
In my testing I used some simple logs to observe what's happening and at what intervals, maybe this helps other too: 
- for the Release code with [these logs](https://gist.github.com/Mugurell/5a0cc7662e5a3b679dc01f982fcf4c1f) I see:
> 10:36:58.056: Mugurel: onCreateView entered
10:36:58.132: Mugurel: onCreateView binding inflated
10:36:58.169: Mugurel: onCreateView before displaying wallpaper
10:36:58.169: Mugurel: onCreateView: displayWallpaperIfEnabled: before updateWallpaper
10:36:58.169: Mugurel: updateWallpaper: Firefox(name=amethyst, drawableId=2131230867)
10:36:58.169: Mugurel: updateWallpaper: before loadSavedWallpaper
10:36:58.223: Mugurel: updateWallpaper: after loadSavedWallpaper
10:36:58.223: Mugurel: updateWallpaper: before scaleBitmapToBottom
10:36:58.224: Mugurel: updateWallpaper: after scaleBitmapToBottom
10:36:58.224: Mugurel: onCreateView: displayWallpaperIfEnabled: after updateWallpaper
10:36:58.224: Mugurel: onCreateView after displaying wallpaper
10:36:58.224: Mugurel: onCreateView finished
10:36:58.247: Mugurel: onStart entered
10:36:58.250: Mugurel: onStart finished

- for `main` with [these logs](https://gist.github.com/Mugurell/357b7d7950a3fd49c924cff84db55376) I see:
> 10:39:22.012: Mugurel: onCreateView entered
10:39:22.103: Mugurel: onCreateView binding inflated
10:39:22.149: Mugurel: onCreateView finished
10:39:22.166: Mugurel: AppAction.WallpaperAction.UpdateCurrentWallpaper: amethyst
10:39:22.167: Mugurel: WallpaperObserver: new wallpaper observed: amethyst
10:39:22.168: Mugurel: WallpaperObserver: loadWallpaper entered: amethyst
10:39:22.182: Mugurel: onStart entered
10:39:22.188: Mugurel: onStart: waiting for first wallpaper to be loaded
10:39:22.214: Mugurel: WallpaperObserver: loadWallpaper finished: amethyst
10:39:22.214: Mugurel: onStart: waiting for first wallpaper over. Bitmap was loaded
10:39:22.214: Mugurel: WallpaperObserver: applyCurrentWallpaper entered
10:39:22.215: Mugurel: WallpaperObserver: applyCurrentWallpaper switched to Main
10:39:22.216: Mugurel: WallpaperObserver: applyCurrentWallpaper finished
10:39:22.216: Mugurel: onStart finished

Mugurell:
Will look into testing the current release with the script from https://github.com/mozilla-mobile/fenix/pull/26794#issuecomment-1235675077, cleanup the code and support eagerly setting / avoiding to wait for the default (no wallpaper selected) wallpaper.

Mugurell:
Testing the time to onResume for the current Release code I see very similar times to what we'd get after this patch.
Updated https://github.com/mozilla-mobile/fenix/pull/26794#issuecomment-1235675077 with the results from this build.
(Not sure why to the last the runs gets significantly slower, happened on all builds and for the 104 Nightly the added results are the best from 3 runs).
These graphs and the logs from https://github.com/mozilla-mobile/fenix/pull/26794#issuecomment-1236655145 which show that `onCreateView -> onResume` takes ~200ms on both the code from 104 and the one from after this patch suggest that we'd not introduce any additional delay.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -213,8 +213,9 @@ class HomeFragment : Fragment() {
         if (shouldEnableWallpaper()) {
             wallpapersObserver = WallpapersObserver(
                 appStore = components.appStore,
+                settings = requireContext().settings(),
                 wallpapersUseCases = components.useCases.wallpaperUseCases,
-                wallpaperImageView = binding.wallpaperImageView,
+                wallpaperImageView = binding.wallpaperImageView
```
gabrielluong, line 218: Just a drive-by: Let's keep the trailing comma here. In general, we are encouraging trailing comma and with the ktlint upgrade, we'll also enforce trailing commas.

MatthewTighe:
> Why is this not a problem in Release (104)? We're loading the wallpapers from disk there too and there's no delay, and also no performance impact to COLD_MAIN.

In 104, there was no requirement to observe changes to the current wallpaper. It was possible to directly load the wallpaper when the fragment was being created or during configuration changes. The introduction of the wallpaper selection onboarding tool, https://github.com/mozilla-mobile/fenix/pull/26758, will require us to react to wallpaper selections as they are being made because the tool will be displayed simultaneously with the home screen. Outside of these specific requirements, I thought that having the current wallpaper available in our reactive architecture would be a benefit for all the usual reasons.

It seems to me that the fundamental issue here is with `Store` observations being structured around `onStart` since we cannot wait that long to bind the wallpaper without visual jank. 

I'm wondering if perhaps we're trying too hard to make all these edge cases fit cleanly with our architecture. It might be easiest to do the something like the following:

1. `onCreateView` manually check storage for the currently set wallpaper and bind it to the view. cache the wallpaper in some local property
2. start observing the store for changes to the wallpaper
3. on updates, discard any that have the same name as the already set wallpaper as is being done at https://github.com/mozilla-mobile/fenix/pull/26794/files#diff-3ae7e608c5aecf918749ce14c2069788b747e179265267bc2a34c3523c0f6fbfR122
4. `onConfigurationChanged` do the same as 1.

Alternatively, we may be able to send a callback to the wallpaper selection tool that would bind a newly selected wallpaper to the home screen. We could also avoid the need for observation in that case.

MatthewTighe:
My biggest concern right now is for 105 beta - which this bug is affecting. In the interest of that, I think I would suggest landing this patch now so that it can be uplifted before the first RC. If we would like to make improvements down the line to reduce complexity we always can.

app/src/main/java/org/mozilla/fenix/home/WallpapersObserver.kt
```diff
@@ -28,18 +34,44 @@ import org.mozilla.fenix.wallpapers.WallpapersUseCases
...
+    @Volatile
     @VisibleForTesting
-    internal var wallpapersScope = CoroutineScope(Dispatchers.Main.immediate)
+    internal var currentWallpaperImage: Bitmap? = null
+
+    /**
+     * Listener for when the first observed wallpaper is loaded and available to be set.
+     */
+    @VisibleForTesting
+    internal val isWallpaperLoaded = CompletableDeferred<Unit>()
```
MatthewTighe, line 74: Is there a clean way to combine these two properties, such that we could have `CompletableDeferred<Bitmap?>`?

MatthewTighe (approved):
I'll approve this in light of my comment above: https://github.com/mozilla-mobile/fenix/pull/26794#issuecomment-1238622900

I do generally agree with @csadilek that this feels like a lot of complexity to introduce to solve these specific problems. Having worked through it a bit myself, though, I understand that it may be necessary given the way things are currently set up. Encapsulating this complexity in a class does seem like a good move to me as well.

My only worry is that higher complexity solves tend to introduce a bit more risk as well. Given that this is looking like a last minute uplift to 105, I would be interested in having @csadilek look through my comments here to see if he agrees that it's worth the risk.

We may be able to figure out something a little more straightforward for 106, and I believe the existing population of wallpapers users is relatively low. That said, I would of course prefer to avoid knowingly ship this bug to 105.

MatthewTighe:
@csadilek and I spoke a bit more about this offline this afternoon. We thought it would be good to double-check that this patch has no affect on `COLD_MAIN` perf tests in addition to the `onResume` modified tests you shared above. It would also be good to double-check that the large `2328` max result was an outlier and not indicative of a problem.

We also discussed whether this was worth uplifting in greater detail and came to the conclusion that it probably isn't, given that the userbase of the wallpapers feature is relatively low. Since that removes some time sensitivity, I would also be open to exploring either of the alternatives I listed above. If we decided to pivot that way, I would be happy to pull that work in myself if you were not interested @Mugurell.

Otherwise, we thought it would be a good strategy to go ahead and land this in nightly for now (assuming that there are no perf regressions for COLD_MAIN). If things are looking stable by next week, we can re-assess whether we think it's worth uplifting to 105.

Mugurell:
@MatthewTighe  @csadilek  Testing with the original `measure_start_up.py` script and
> python3 measure_start_up.py -c=100 --product=fenix nightly cold_main_first_frame

I got the following:

| | Nightly built locally from `releases_v104.0.0` | Nighly built locally from `main` | Nightly built locally from this patch | 
| --- | --- | --- | --- |
| replicate count: | 100 | 100 | 100
| max: | 1925 | 1451 | 2080
| mean: | 1749 | 1324 | 1635
| median: | 1715 | 1387 | 1637
| min: | 1397 | 1175 | 1344
| link: | [results](https://drive.google.com/file/d/1fcvURB5ZQsWZrBTiS8i8lMGLoJ7EsV5_/view?usp=sharing) | [results](https://drive.google.com/file/d/1oX5hkFVdem5HXyEnkIXEwt4mtCV4AfCf/view?usp=sharing) | [results](https://drive.google.com/file/d/1VSVCavtWWMjKuFZtfUCSs5dtth9VmXf3/view?usp=sharing) |
| graph | Fenix104ColdMainFirstFrame | Fenix106MainMainFirstFrame | Fenix106PatchColdFirstFrame |

- The current code on main is expected to be a bit faster since we're not waiting anymore for the wallpaper to be set. 
(In a first try on fixing this it worked if using IO but then we saw crashes and after switching to Main (but not blocking) the issue reproduces again)
- Not sure why the numbers show some steps on getting bigger. Saw this in all of my tests.

MatthewTighe:
I think as long as we can demonstrate that the regression shown between main and this patch only occurs for users that have wallpapers set then we can go ahead and land this. We can focus on improving the performance for wallpaper users down the line, but I don't think we can justify that kind of performance hit for users that are not using the feature.

app/src/main/java/org/mozilla/fenix/home/WallpapersObserver.kt
Mugurell: Can't say I'm happy with using all these properties and I did try to keep them to a minimum.

The `CompletableDeferred` is used just to be able to easily `await` having the first wallpaper. Otherwise we would maybe have to use a loop and a condition. This is more idiomatic and does not block the caller thread. 
It is publicly available but "completed" only once without passing other data to callers and allowing the `WallpaperObserver` to encapsulate all other state.

The `currentWallpaperImage` separate property helps decoupling the two operations - load bitmap and set bitmap through the use of this shared state. In my tests this allowed for quicker execution versus just one method with context switching between threads. By having two separate methods for load/set and being able to use a cached bitmap (same that's shown so there's no added memory pressure) we also reduce to a minimum the time needed to set the wallpaper after a configuration changes - we don't have to load again the bitmap.
It is a private property that is of no use to `WallpaperObserver` clients.

MatthewTighe:
Thanks for your due diligence on this one! Let's land this and let it bake in nightly for now, and we can decide whether to include it in a 105 dot release later if it's deemed necessary with product input.

> Hardcoded an early return from trying to setting the wallpaper if users is using the default in 2nd commit.
In my tests this is not needed as there is enough time to observe the store and avoid in the observer spending another 80-100ms to load the bitmap and looks a bit ugly but seems like the maximum we could do to ensure users with the default wallpaper are not affected in any way.

I think we should continue with this direction for now, if only to save ourselves any potential pain in perf regressions later.

csadilek:
@Mugurell @MatthewTighe ⚠️ We have numerous reports of Nightly (home screen) freezing starting Saturday. I will put up a revert to verify if caused by this change.

csadilek:
Revert to test/verify is: https://github.com/mozilla-mobile/fenix/pull/26923

csadilek:
OK we reverted both changes, and verified the issue is gone. @MatthewTighe has ANR logs, it seems more likely that it was this change. Let's 🕵️ and then land again with a fix.

PR #26847, opened 2022-09-06
For #26796 - Fix PocketStoriesComposablesPreview

### LINKED ISSUE #26796: Fix previews in PocketStoriesComposables
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21466)

MozillaNoah:
Added a simple `inComposePreview` check to `Modifier.onShown`, since that modifier isn't necessary when just previewing the UI.

### Pull Request checklist

### QA

- [ ] **QA Needed**

### GitHub Automation
Fixes #26796

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesComposables.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesComposables.kt
@@ -67,6 +67,7 @@ import org.mozilla.fenix.compose.ListItemTabSurface
 import org.mozilla.fenix.compose.SelectableChip
 import org.mozilla.fenix.compose.StaggeredHorizontalGrid
 import org.mozilla.fenix.compose.TabSubtitleWithInterdot
+import org.mozilla.fenix.compose.inComposePreview
 import org.mozilla.fenix.ext.settings
 import org.mozilla.fenix.theme.FirefoxTheme
 import org.mozilla.fenix.theme.Theme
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesComposables.kt
```diff
@@ -275,40 +276,44 @@ private fun Modifier.onShown(
...
-        val toolbarHeight = context.resources.getDimensionPixelSize(R.dimen.browser_toolbar_height)
-        val isToolbarPlacedAtBottom = context.settings().shouldUseBottomToolbar
-        // Get a Rect of the entire screen minus system insets minus the toolbar
-        val screenBounds = Rect()
-            .apply { LocalView.current.getWindowVisibleDisplayFrame(this) }
-            .apply {
-                when (isToolbarPlacedAtBottom) {
-                    true -> bottom -= toolbarHeight
-                    false -> top += toolbarHeight
+        if (inComposePreview) {
```
jonalmeida, line 279: Drive-by comment (not a review): take care to note that if we use more checks like this, we are leading ourselves into a direction where we cannot trust the preview any more because what it shows us will be different from what is actually seen on real devices.

So I'm uncertain if fixing the preview in this way is helping us or if it's better to not have the preview work if we can't make it preview what is seen on a device, so that way we are forced to test it on a device.

MozillaNoah: I can definitely see the concern you're pointing out here. My opinion is that the Compose previews aren't meant to handle _all_ of our very complex UI use cases, such as image loading from storage or accounting for dynamic app-level settings. In those situations, short circuiting some of their logic with a check to `inComposePreview` should suffice. I feel that the previews should more give a window to smaller chunks of a user experience and not the holistic view of the app's more in-depth UI logic. 

That being said, I do think there is a time and place to put effort into coding previews to validate a complex piece of UI logic that is not directly coupled with business logic (such as the `Modifier` here). The preview in this file, however, was not written to validate this, so my use of `inComposePreview` does only fix the particular preview use case of static UI validation. If it makes sense to do so, we can open an issue to refactor this more band-aid solution and add a preview to validate the usage of `Modifier.onShown` with `PocketStories`.

PR #26853, opened 2022-09-07
For #26743: Update URL text styling.

### LINKED ISSUE #26743: Update the Pocket powered by section design
Figma link  / Screenshot of design to be added.

<details><summary>Screenshot</summary>Screen Shot 2022-09-06 at 12 34 57 PM</details>

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21440)

mcarare:
Replace colored text with underlined text as per UX design.

After patch:

### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation
Fixes #26743

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/compose/ClickableSubstringLink.kt
+++ b/app/src/main/java/org/mozilla/fenix/compose/ClickableSubstringLink.kt
@@ -13,6 +13,7 @@ import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.text.SpanStyle
 import androidx.compose.ui.text.buildAnnotatedString
+import androidx.compose.ui.text.style.TextDecoration
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.sp
 import org.mozilla.fenix.theme.FirefoxTheme
@@ -45,7 +46,10 @@ fun ClickableSubstringLink(
         )
 
         addStyle(
-            SpanStyle(color = FirefoxTheme.colors.textAccent),
+            SpanStyle(
+                color = textColor,
+                textDecoration = TextDecoration.Underline
+            ),
             start = clickableStartIndex,
             end = clickableEndIndex
         )


app/src/main/java/org/mozilla/fenix/compose/ClickableSubstringLink.kt
```diff
@@ -45,7 +46,7 @@ fun ClickableSubstringLink(
         )
 
         addStyle(
-            SpanStyle(color = FirefoxTheme.colors.textAccent),
```
MozillaNoah, line 48: if the accent color is being removed, should the `textColor` be applied to the entire `annotatedText` now?

MozillaNoah:
The figma wasn't added to the root issue. Could that be updated?

mcarare:
> The figma wasn't added to the root issue. Could that be updated?

I only have a screenshot from UX. 
cc @Kate-Galetski

app/src/main/java/org/mozilla/fenix/compose/ClickableSubstringLink.kt
mcarare: I was going to apply the color to the whole text directly, but I also had in mind (if this would end up being used in more places) to keep it more flexible. We could add both different coloring and underline, or use an enum to choose between them.

MozillaNoah: Fair enough

MozillaNoah:
I noticed that [this PR to add test tags](https://github.com/mozilla-mobile/fenix/pull/26843) currently has your 1 changed file set to be removed.

Mugurell:
> I noticed that [this PR to add test tags](https://github.com/mozilla-mobile/fenix/pull/26843) currently has your 1 changed file set to be removed.

The change to remove `ClickableSubstringLink` will be dropped from that PR.

MozillaNoah (approved):

app/src/main/java/org/mozilla/fenix/compose/ClickableSubstringLink.kt
```diff
@@ -45,7 +46,10 @@ fun ClickableSubstringLink(
...
+            SpanStyle(
+                color = textColor,
```
gabrielluong, line 50: Should this only be for when we have a wallpaper? All our existing links use textAccent and it is what it is defined for by the design system. So, it would seem strange to me that we were to decide to have a single instance of a link just being the underline outside of private mode.

mcarare: See also my comment [here.](https://github.com/mozilla-mobile/fenix/pull/26853#discussion_r965089966).

For the moment, as we are using it only in one place, I did not want to overcomplicate this composable. But it can be refactored to give the caller the chance to choose between span styles for the linked text, or even receive a `SpanStyle` as an argument if we choose to do so in the future. 

As a side note, taking into consideration the need for UX consistency, for me at least, it would be desirable if all links in the app are represented the same way.

gabrielluong: To be clear, I am talking about shouldn't this still be using text accent when we don't have a wallpaper?

gabrielluong: The current change would mean the link would be using text primary + underline in the default (no wallpaper) scenario, whereas I think we should be using textAccent for the default and textPrimary (textColor) when there is a wallpaper, but this might be a UX question.

mcarare: When talking offline with @Kate-Galetski we did not make a differentiation between having a wallpaper or not, but this as being the new styling of the link in this section. 
Maybe @Kate-Galetski can confirm I understood correctly.

gabrielluong: Ok, let's bring this up in the #firefox-mobile-eng-ux slack. I believe you're going on PTO next week. So, I can follow up on any changes.

Kate-Galetski:
Hi. I am working on the Section Titles color (aka Jump Back in, Recently Saved) and links color (aka Show all, Learn more)when wallpaper is on.
I am proposing to:
Using Light / Text / Primery (Hex Code: 15141A) for dark text.
Using Light / Text /On Color Primary (Hex Code: FBFBFE) for light text

When wallpaper is off (no wallpaper) please let's default to already agreed color styles.

Does it make sense?

app/src/main/java/org/mozilla/fenix/compose/ClickableSubstringLink.kt
gabrielluong: From @Kate-Galetski 
Hi. I am working on the Section Titles color (aka Jump Back in, Recently Saved) and links color (aka Show all, Learn more) when wallpaper is on.
 I am proposing to:
 Using Light / Text / Primary (Hex Code: 15141A) for dark text.
 Using Light / Text /On Color Primary (Hex Code: FBFBFE) for light text
When wallpaper is off (no wallpaper) please let's default to already agreed color styles.

gabrielluong:
Closing this in favor of https://github.com/mozilla-mobile/fenix/pull/26950

Issue #23570, opened 2022-02-04
Increase tap area for 'Show all' links


Alexandru2909:
We want to make 'Show all' buttons from home fragment touch targets larger.
This will improve the user experience for people with poor eyesight and coordination issues. 

![Screenshot 2022-02-04 at 16 53 31](https://user-images.githubusercontent.com/35462038/152550002-3e7c806d-3f04-49b4-88c8-49b6d382eb75.png)

sarah541:
This will be addressed in the Ui refactoring to compose here : https://github.com/mozilla-mobile/fenix/pull/24171

PR #26885, opened 2022-09-08
For #26833 - Make `Show all` button from homescreen header accessible with Talkback

### LINKED ISSUE #26833: [a11y] Can't click on "Show all" from homescreen with Talkback
### Steps to reproduce
- Activate Talkback or other accessibility service
- Navigate on homescreen to a section header with a "Show all" button.

### Expected behavior
- A double tap gesture would trigger the "Show all" action.
- The content description read explicitly states that the gesture contains a button (when appropriately)

### Actual behavior
- A double tab gesture anywhere on the screen does not activate the "Show all" button.
- The content description read explicitly mentions a button only for the "Jump back in section" but not also for the "Recently visited section" 

### App versions affected
All 104 and later. Issue may be older than from just 104.

https://user-images.githubusercontent.com/11428869/188592129-6a8e5f04-b14a-4030-998c-a198517a8bdd.mp4

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21485)

Alexandru2909:
`Show all` button from the homescreen headers should now be accessible in Talkback.

| Before | After |
| --- | --- |
| v1 | v3 |

Ripple effect light:
| Before | After |
| --- | --- |
| https://user-images.githubusercontent.com/35462038/191580656-c5361c17-88d4-47af-adc1-15576ce8a911.mp4 | https://user-images.githubusercontent.com/35462038/191580842-c4c51738-7bce-4373-b0df-3692bcdcdc18.mp4 |

Ripple effect dark:
| Before | After |
| --- | --- |
| https://user-images.githubusercontent.com/35462038/191581050-15144cdb-e49c-477d-9c7c-2fec5be4ae0e.mp4 | https://user-images.githubusercontent.com/35462038/191581101-a3d27949-6e17-4e6f-bd3d-cc99f91600c6.mp4 |

This PR should also fix https://github.com/mozilla-mobile/fenix/issues/23570.

### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation
Fixes #26833

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/compose/button/TextButton.kt
+++ b/app/src/main/java/org/mozilla/fenix/compose/button/TextButton.kt
@@ -7,35 +7,40 @@ package org.mozilla.fenix.compose.button
 import android.content.res.Configuration
 import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.Box
+import androidx.compose.material.ButtonColors
+import androidx.compose.material.ButtonDefaults.textButtonColors
 import androidx.compose.material.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.tooling.preview.Preview
 import org.mozilla.fenix.theme.FirefoxTheme
-import java.util.Locale
 
 /**
  * Text-only button.
  *
- * @param text The button text to be displayed.
- * @param onClick Invoked when the user clicks on the button.
  * @param modifier [Modifier] Used to shape and position the underlying [androidx.compose.material.TextButton].
+ * @param text The button text to be displayed.
+ * @param buttonColors [ButtonColors] for the button.
+ * @param onClick Invoked when the user clicks on the button.
  * @param textColor [Color] to apply to the button text.
  */
 @Composable
+@Suppress("LongParameterList")
 fun TextButton(
-    text: String,
-    onClick: () -> Unit,
     modifier: Modifier = Modifier,
+    text: String,
+    buttonColors: ButtonColors = textButtonColors(),
+    onClick: () -> Unit,
     textColor: Color = FirefoxTheme.colors.textAccent,
 ) {
     androidx.compose.material.TextButton(
         onClick = onClick,
+        colors = buttonColors,
         modifier = modifier,
     ) {
         Text(
-            text = text.uppercase(Locale.getDefault()),
+            text = text,
             color = textColor,
             style = FirefoxTheme.typography.button,
             maxLines = 1,
--- a/app/src/main/java/org/mozilla/fenix/compose/home/HomeSectionHeader.kt
+++ b/app/src/main/java/org/mozilla/fenix/compose/home/HomeSectionHeader.kt
@@ -4,11 +4,11 @@
 
 package org.mozilla.fenix.compose.home
 
+import androidx.compose.foundation.isSystemInDarkTheme
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.fillMaxWidth
-import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.wrapContentHeight
-import androidx.compose.foundation.text.ClickableText
+import androidx.compose.material.ButtonDefaults.textButtonColors
 import androidx.compose.material.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Alignment
@@ -17,15 +17,13 @@ import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.semantics.contentDescription
 import androidx.compose.ui.semantics.semantics
-import androidx.compose.ui.text.AnnotatedString
-import androidx.compose.ui.text.TextStyle
 import androidx.compose.ui.text.style.TextOverflow
 import androidx.compose.ui.tooling.preview.Preview
-import androidx.compose.ui.unit.dp
-import androidx.compose.ui.unit.sp
 import mozilla.components.lib.state.ext.observeAsComposableState
+import mozilla.components.ui.colors.PhotonColors
 import org.mozilla.fenix.R
 import org.mozilla.fenix.components.components
+import org.mozilla.fenix.compose.button.TextButton
 import org.mozilla.fenix.compose.inComposePreview
 import org.mozilla.fenix.theme.FirefoxTheme
 import org.mozilla.fenix.wallpapers.Wallpaper
@@ -104,17 +102,19 @@ private fun HomeSectionHeaderContent(
         )
 
         onShowAllClick?.let {
-            ClickableText(
-                text = AnnotatedString(text = stringResource(id = R.string.recent_tabs_show_all)),
-                modifier = Modifier.padding(start = 16.dp)
-                    .semantics {
-                        contentDescription = description
+            TextButton(
+                text = stringResource(id = R.string.recent_tabs_show_all),
+                buttonColors = textButtonColors(
+                    contentColor = if (isSystemInDarkTheme()) {
+                        PhotonColors.White
+                    } else {
+                        PhotonColors.Black
                     },
-                style = TextStyle(
-                    color = showAllTextColor,
-                    fontSize = 14.sp,
                 ),
                 onClick = { onShowAllClick() },
+                modifier = Modifier
+                    .semantics { contentDescription = description },
+                textColor = showAllTextColor,
             )
         }
     }
--- a/app/src/main/java/org/mozilla/fenix/home/recentbookmarks/view/RecentBookmarksHeaderViewHolder.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/recentbookmarks/view/RecentBookmarksHeaderViewHolder.kt
@@ -48,7 +48,9 @@ class RecentBookmarksHeaderViewHolder(
     @Composable
     override fun Content() {
         Column {
-            Spacer(modifier = Modifier.height(40.dp))
+            // Padding reduced by 12dp to compensate for the header height increase
+            // from https://github.com/mozilla-mobile/fenix/pull/26885
+            Spacer(modifier = Modifier.height(28.dp))
 
             HomeSectionHeader(
                 headerText = stringResource(R.string.recently_saved_title),
@@ -59,7 +61,9 @@ class RecentBookmarksHeaderViewHolder(
                 },
             )
 
-            Spacer(Modifier.height(16.dp))
+            // Padding reduced by 12dp to compensate for the header height increase
+            // from https://github.com/mozilla-mobile/fenix/pull/26885
+            Spacer(Modifier.height(4.dp))
         }
     }
 
--- a/app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabsHeaderViewHolder.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/recenttabs/view/RecentTabsHeaderViewHolder.kt
@@ -47,7 +47,9 @@ class RecentTabsHeaderViewHolder(
     @Composable
     override fun Content() {
         Column {
-            Spacer(modifier = Modifier.height(40.dp))
+            // Padding reduced by 12dp to compensate for the header height increase
+            // from https://github.com/mozilla-mobile/fenix/pull/26885
+            Spacer(modifier = Modifier.height(28.dp))
 
             HomeSectionHeader(
                 headerText = stringResource(R.string.recent_tabs_header),
@@ -58,7 +60,9 @@ class RecentTabsHeaderViewHolder(
                 },
             )
 
-            Spacer(Modifier.height(16.dp))
+            // Padding reduced by 12dp to compensate for the header height increase
+            // from https://github.com/mozilla-mobile/fenix/pull/26885
+            Spacer(Modifier.height(4.dp))
         }
     }
 
--- a/app/src/main/java/org/mozilla/fenix/home/recentvisits/view/RecentVisitsHeaderViewHolder.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/recentvisits/view/RecentVisitsHeaderViewHolder.kt
@@ -40,7 +40,9 @@ class RecentVisitsHeaderViewHolder(
     @Composable
     override fun Content() {
         Column {
-            Spacer(modifier = Modifier.height(40.dp))
+            // Padding reduced by 12dp to compensate for the header height increase
+            // from https://github.com/mozilla-mobile/fenix/pull/26885
+            Spacer(modifier = Modifier.height(28.dp))
 
             HomeSectionHeader(
                 headerText = stringResource(R.string.history_metadata_header_2),
@@ -48,7 +50,9 @@ class RecentVisitsHeaderViewHolder(
                 onShowAllClick = interactor::onHistoryShowAllClicked,
             )
 
-            Spacer(Modifier.height(16.dp))
+            // Padding reduced by 12dp to compensate for the header height increase
+            // from https://github.com/mozilla-mobile/fenix/pull/26885
+            Spacer(Modifier.height(4.dp))
         }
     }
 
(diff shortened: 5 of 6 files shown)


app/src/main/java/org/mozilla/fenix/compose/home/HomeSectionHeader.kt
```diff
@@ -90,27 +90,29 @@ private fun HomeSectionHeaderContent(
...
                 .weight(1f)
-                .wrapContentHeight(align = Alignment.Top)
+                .wrapContentHeight(),
         )
 
         onShowAllClick?.let {
-            ClickableText(
-                text = AnnotatedString(text = stringResource(id = R.string.recent_tabs_show_all)),
-                modifier = Modifier.padding(start = 16.dp)
+            BasicText(
```
gabrielluong, line 104: I am wondering why we are replacing `ClickableText` since it is just a wrapper around `BasicText`

app/src/main/java/org/mozilla/fenix/home/recentbookmarks/view/RecentBookmarksHeaderViewHolder.kt
```diff
@@ -48,7 +48,7 @@ class RecentBookmarksHeaderViewHolder(
     @Composable
     override fun Content() {
         Column {
-            Spacer(modifier = Modifier.height(40.dp))
+            Spacer(modifier = Modifier.height(34.dp))
```
gabrielluong, line 51: Where are the values for these Spacer changes coming from?

app/src/main/java/org/mozilla/fenix/compose/home/HomeSectionHeader.kt
Alexandru2909: I believe this is because the `onClick` implementation for `ClickableText` is made using `Modifier.pointerInput`, whereas `Modifier.clickable` allows talkback to view the modified composable as a button.

I noticed this is also used in [`ClickableSubstringLink`](https://github.com/mozilla-mobile/fenix/blob/aa4a072d73a036b7108740d53a466daadafed397/app/src/main/java/org/mozilla/fenix/compose/ClickableSubstringLink.kt#L81), should we file an issue for that? the links provided by this component are not accessible either.

app/src/main/java/org/mozilla/fenix/home/recentbookmarks/view/RecentBookmarksHeaderViewHolder.kt
Alexandru2909: Because in the changes of this PR the tap area for the `Show all` button is increased, I decreased the space around the header where the `Show all` button is used, to maintain the previous aspect.

app/src/main/java/org/mozilla/fenix/compose/home/HomeSectionHeader.kt
Mugurell: The `ClickableSubstringLink` used for Pocket is merged with the other siblings so there is just one a11y item which does allow double click to open the link.
The same could be done for the `ClickableSubstringLink` used for wallpaper settings but I'm not seeing a link there.

app/src/main/java/org/mozilla/fenix/compose/home/HomeSectionHeader.kt
```diff
@@ -105,18 +108,28 @@ private fun HomeSectionHeaderContent(
...
+                        .semantics {
+                            contentDescription = description
+                        }
+                        .padding(start = 16.dp),
+                    style = TextStyle(
+                        color = showAllTextColor,
+                        fontSize = 14.sp,
+                    ),
+                )
+            }
```
gabrielluong, line 132: This code looks like a `TextButton` to me now. https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/java/org/mozilla/fenix/compose/button/TextButton.kt#28

I don't have an immediate chance to think about this more thoroughly so I will cc @MozillaNoah

app/src/main/java/org/mozilla/fenix/home/recentbookmarks/view/RecentBookmarksHeaderViewHolder.kt
gabrielluong: I haven't had a chance to pull down the code and play around with it. My secondary concern is that these values might be overcompensating since I changed the header to center vertically in https://github.com/mozilla-mobile/fenix/pull/26963/files when we bumped up the Header font size.

app/src/main/java/org/mozilla/fenix/compose/home/HomeSectionHeader.kt
MozillaNoah: I agree that this should just be replaced with a `TextButton`. Depending how the Talkback works, I would either add the semantics to the `TextButton` modifier or add a parameter to our composable to pass it down to the inner `Text()` composable.

MozillaNoah:
Separately, the click area looks a little off currently:
image

Switching the `Box` + `BasicText` to a `TextButton` may fix this for us.

app/src/main/java/org/mozilla/fenix/compose/home/HomeSectionHeader.kt
gabrielluong: @MozillaNoah Thanks for taking a look!

app/src/main/java/org/mozilla/fenix/compose/home/HomeSectionHeader.kt
```diff
@@ -105,18 +108,31 @@ private fun HomeSectionHeaderContent(
         )
 
         onShowAllClick?.let {
-            ClickableText(
-                text = AnnotatedString(text = stringResource(id = R.string.recent_tabs_show_all)),
-                modifier = Modifier.padding(start = 16.dp)
-                    .semantics {
-                        contentDescription = description
+            TextButton(
```
MozillaNoah, line 111: We actually have a `TextButton` composable under `org.mozilla.fenix.compose.button` to use here. Sorry if that wasn't clear in a previous comment.

app/src/main/java/org/mozilla/fenix/compose/home/HomeSectionHeader.kt
```diff
@@ -105,18 +108,31 @@ private fun HomeSectionHeaderContent(
...
+                colors = textButtonColors(
+                    contentColor = if (isSystemInDarkTheme()) {
+                        PhotonColors.White
+                    } else {
+                        PhotonColors.Black
                     },
```
MozillaNoah, line 117: Make sure to use the `showAllTextColor` parameter from `HomeSectionHeader`

app/src/main/java/org/mozilla/fenix/compose/button/TextButton.kt
```diff
@@ -20,23 +24,36 @@ import java.util.Locale
...
  * @param modifier [Modifier] Used to shape and position the underlying [androidx.compose.material.TextButton].
+ * @param contentPaddingValues Padding values for the button's content
  * @param textColor [Color] to apply to the button text.
  */
 @Composable
+@Suppress("LongParameterList")
 fun TextButton(
     text: String,
+    shouldUppercaseText: Boolean = true,
+    buttonColors: ButtonColors = textButtonColors(),
```
MozillaNoah, line 39: why was `buttonColors` added as a parameter? from what I can tell, it's only being used for a text color, which is already covered by the `textColor` parameter

app/src/main/java/org/mozilla/fenix/compose/button/TextButton.kt
```diff
@@ -20,23 +24,36 @@ import java.util.Locale
...
     onClick: () -> Unit,
     modifier: Modifier = Modifier,
```
MozillaNoah, line 41: the parameter ordering should mirror `androidx.compose.material.TextButton`. Modifier is always the first optional parameter with Composables.

app/src/main/java/org/mozilla/fenix/compose/home/HomeSectionHeader.kt
```diff
@@ -105,17 +108,28 @@ private fun HomeSectionHeaderContent(
...
-                        contentDescription = description
+            TextButton(
+                text = stringResource(id = R.string.recent_tabs_show_all),
+                shouldUppercaseText = false,
+                buttonColors = textButtonColors(
+                    contentColor = if (isSystemInDarkTheme()) {
+                        PhotonColors.White
+                    } else {
+                        PhotonColors.Black
                     },
```
MozillaNoah, line 119: this parameter is covered by `textColor = showAllTextColor`. We need to use `showAllTextColor` to adapt the text to a different color based on the selected wallpaper.

app/src/main/java/org/mozilla/fenix/compose/home/HomeSectionHeader.kt
```diff
@@ -105,17 +108,28 @@ private fun HomeSectionHeaderContent(
...
+                modifier = Modifier
+                    .semantics { contentDescription = description },
+                contentPaddingValues = PaddingValues(
+                    start = 16.dp,
+                    end = ButtonDefaults.TextButtonContentPadding.calculateEndPadding(
+                        LocalLayoutDirection.current,
+                    ),
+                    top = ButtonDefaults.TextButtonContentPadding.calculateTopPadding(),
+                    bottom = ButtonDefaults.TextButtonContentPadding.calculateBottomPadding(),
+                ),
```
MozillaNoah, line 131: after switching from `ClickableText` to `TextButton`, the `contentPaddingValues` should no longer be necessary. I've tested this locally, and the clickable area looks fine without this.

app/src/main/java/org/mozilla/fenix/home/recentbookmarks/view/RecentBookmarksHeaderViewHolder.kt
```diff
@@ -59,7 +59,7 @@ class RecentBookmarksHeaderViewHolder(
                 },
             )
 
-            Spacer(Modifier.height(16.dp))
+            Spacer(Modifier.height(4.dp))
```
MozillaNoah, line 62: echoing @gabrielluong's concerns here, the Spacer changes in the viewholders should not be necessary with the change to `TextButton` from `BasicText`

MozillaNoah (commented):
Just did another pass for this PR. The tap area looks great, and it's nice having the touch ripple. I've left a few notes for polish and simplification of some of the parameters.

app/src/main/java/org/mozilla/fenix/compose/button/TextButton.kt
```diff
@@ -20,23 +24,36 @@ import java.util.Locale
...
  * @param onClick Invoked when the user clicks on the button.
  * @param modifier [Modifier] Used to shape and position the underlying [androidx.compose.material.TextButton].
+ * @param contentPaddingValues Padding values for the button's content
  * @param textColor [Color] to apply to the button text.
  */
 @Composable
+@Suppress("LongParameterList")
 fun TextButton(
     text: String,
+    shouldUppercaseText: Boolean = true,
```
gabrielluong, line 38: I don't think `TextButton` should need to know whether or not the text should be uppercase. The caller of `TextButton` should pass the correctly cased text.

app/src/main/java/org/mozilla/fenix/compose/home/HomeSectionHeader.kt
Alexandru2909: This is to used add the ripple effect, I'll update the documentation to reflect this.

app/src/main/java/org/mozilla/fenix/compose/button/TextButton.kt
MozillaNoah: Gotcha gotcha. If you haven't already, could you double check with UX what the ripple color should be? I'm unable to find a token for this with our current color system.

app/src/main/java/org/mozilla/fenix/home/recentbookmarks/view/RecentBookmarksHeaderViewHolder.kt
MozillaNoah: Gotcha gotcha. I'd ask we UX check this then to ensure padding amount is what they're expecting.

Alexandru2909:
Tagged UX to check if the ripple effect and the padding for the header is expected.

app/src/main/java/org/mozilla/fenix/compose/home/HomeSectionHeader.kt
```diff
@@ -105,17 +103,19 @@ private fun HomeSectionHeaderContent(
...
-                modifier = Modifier.padding(start = 16.dp)
-                    .semantics {
-                        contentDescription = description
+            TextButton(
+                text = stringResource(id = R.string.recent_tabs_show_all),
+                buttonColors = textButtonColors(
+                    contentColor = if (isSystemInDarkTheme()) {
+                        PhotonColors.White
+                    } else {
+                        PhotonColors.Black
```
gabrielluong, line 112: Where are these color values coming from?

app/src/main/java/org/mozilla/fenix/home/recentbookmarks/view/RecentBookmarksHeaderViewHolder.kt
gabrielluong: Just want to make sure I am understanding the difference between the screenshots here. If my assumption is correct, the 2 screenshots would be still using the new composable with various spacer changes. However, there is no screenshot of what the current nightly looks like for comparison.

Alexandru2909: I added screenshot for comparison in the PR description, however I didn't provide one to be compared with the spacer changes removed. Here it is:

| Current headers and spacing | Updated header with spacing | Updated header without spacing |
| --- | --- | --- |
| b4 | after_spacing | after_no_spacing |

app/src/main/java/org/mozilla/fenix/compose/home/HomeSectionHeader.kt
Alexandru2909: I added these values to match the ripple effect applied to the button before the compose migration.

app/src/main/java/org/mozilla/fenix/home/recentbookmarks/view/RecentBookmarksHeaderViewHolder.kt
gabrielluong: Hmm.. from I can see is that the "Updated header with spacing" does align better with the current headers and spacing. However, I am not fully sure it if the spacing matched exactly to the current header and spacing as I was visually inspecting and drawing a horizontal line across some of the boundary lines across the 2 different screenshots.

One immediate issue that I want to point out is that we need to document in the Spacer that we are subtracting the 40dp and 16dp spacers by 12dp to account for the padding of this button. Otherwise, it would be too easy for someone to look at the figma designs and say this should be 40dp and revert these changes. I can understand from a coding perspective that it's unfortunate we can't be super explicit about just saying 40dp/16dp for these spacers to exactly match the figma padding specs. However, I understand that all of these buttons have some padding to them that makes them bigger than they seem due to the minimum touch surface and ripple effects.

app/src/main/java/org/mozilla/fenix/compose/home/HomeSectionHeader.kt
gabrielluong: Still just struggling a little bit with understanding this one. Could you point me to the code that we are matching the ripple effect to? For some reason, I generally assume the ripple effect is automatically provided by the built-in material composable components so I am still of the belief that these colors are built-in and provided for free since it's strange to see that we need to provide them. My secondary issue is the fact this wouldn't scale if we had more than the light/dark theme.

Alexandru2909: I believe [this](https://github.com/mozilla-mobile/fenix/blob/e373f507a117679bf58e796834d7a4ebcf805c17/app/src/main/res/layout/recent_bookmarks_header.xml#L31) was the line responsible for adding the ripple effect. For composable buttons this is indeed built-in. In this case, if it wouldn't be specified, the ripple color would be picked up from the color of the text, which is `textAccent` (Violet20). If the ripple color before moving to compose is not the one we want to keep I can revert this change.

PR #26907, opened 2022-09-09
Update hash to be ignored for ktlint update

rvandermeulen:
The original commit was rebased before landing and the hash changed. Update it to the landed one. Also add a reference to the issue number.


gabrielluong (approved):

mcarare:
@rvandermeulen Thanks, I was going to check on that.

PR #26909, opened 2022-09-09
For #26908 - Update wallpapers settings page design

### LINKED ISSUE #26908: Updated Wallpapers Settings Page Design
Wallpaper settings page is updated to new [attached design](https://www.figma.com/file/d7l1iGabQYqnLcnQbHDGzM/Wallpaper-2.0?node-id=5551%3A191299)

    Wallpaper options are split into classic and seasonal

    Up to 6 wallpapers (including null) can be included in classic section

    Up to 6 wallpapers can be included in the seasonal section

    There is a heading for classic wallpapers (use string from Figma)

    There is a heading, description, and learn more link for seasonal wallpapers (use strings from Figma)

        There is also a backup description string that’s landed & localized

        Learn more link points to TBD

Selecting a wallpaper in the settings menu will set your homepage background to that wallpaper. You can exit the settings by clicking the back button

Selecting a wallpaper that has not been downloaded will show a loading indicator while the wallpaper downloads, similar to the wallpaper selection tool:

     

    There is a learn more button for seasonal wallpapers that takes users to a SUMO article (link TBD)

    If the wallpaper thumbnails are not downloaded when a user navigates to the settings page, then only the classic wallpapers will be shown

    Changes are behind the wallpapers v2 feature flag

    Telemetry captured

        Clicks on each wallpaper (indicator of which one)

        How many users have a wallpaper set (existing)

        How many users have each wallpaper set (existing)

        Clicks on link to learn more

Success Metrics

    More users select a wallpaper

    More users have a wallpaper set

    (Learning) Indication of which wallpapers users are most interested in

    (Learning) How many users click to learn more, as an indication of interest in the feature

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21525)

sarah541:
### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation
Fixes #26908

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/settings/wallpaper/Extensions.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/wallpaper/Extensions.kt
@@ -1,0 +1,27 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.settings.wallpaper
+
+import org.mozilla.fenix.wallpapers.Wallpaper
+
+/**
+ * The extension function to group wallpapers according to their name.
+ **/
+fun List<Wallpaper>.groupByDisplayableCollection(): Map<Wallpaper.Collection, List<Wallpaper>> = groupBy {
+    it.collection
+}.filter {
+    it.key.name != "default"
+}.map {
+    val wallpapers = it.value.filter { wallpaper ->
+        wallpaper.thumbnailFileState == Wallpaper.ImageFileState.Downloaded
+    }
+    if (it.key.name == "classic-firefox") {
+        it.key to listOf(Wallpaper.Default) + wallpapers
+    } else {
+        it.key to wallpapers
+    }
+}.toMap().takeIf {
+    it.isNotEmpty()
+} ?: mapOf(Wallpaper.DefaultCollection to listOf(Wallpaper.Default))
--- a/app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
@@ -16,8 +16,10 @@ import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.aspectRatio
+import androidx.compose.foundation.layout.fillMaxHeight
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.rememberScrollState
@@ -25,6 +27,7 @@ import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.foundation.verticalScroll
 import androidx.compose.material.CircularProgressIndicator
 import androidx.compose.material.Surface
+import androidx.compose.material.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.getValue
@@ -37,9 +40,11 @@ import androidx.compose.ui.graphics.asImageBitmap
 import androidx.compose.ui.layout.ContentScale
 import androidx.compose.ui.platform.LocalConfiguration
 import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.text.style.TextDecoration
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
 import org.mozilla.fenix.R
+import org.mozilla.fenix.compose.ClickableSubstringLink
 import org.mozilla.fenix.theme.FirefoxTheme
 import org.mozilla.fenix.theme.Theme
 import org.mozilla.fenix.wallpapers.Wallpaper
@@ -48,34 +53,110 @@ import org.mozilla.fenix.wallpapers.Wallpaper
  * The screen for controlling settings around Wallpapers. When a new wallpaper is selected,
  * a snackbar will be displayed.
  *
- * @param wallpapers Wallpapers to add to grid.
+ * @param wallpaperGroups Wallpapers groups to add to grid.
  * @param selectedWallpaper The currently selected wallpaper.
  * @param defaultWallpaper The default wallpaper.
  * @param loadWallpaperResource Callback to handle loading a wallpaper bitmap. Only optional in the default case.
  * @param onSelectWallpaper Callback for when a new wallpaper is selected.
+ * @param onLearnMoreClick Callback for when the learn more action is clicked from the group description.
  */
 @SuppressLint("UnusedMaterialScaffoldPaddingParameter")
 @Composable
 @Suppress("LongParameterList")
 fun WallpaperSettings(
-    wallpapers: List<Wallpaper>,
+    wallpaperGroups: Map<Wallpaper.Collection, List<Wallpaper>>,
     defaultWallpaper: Wallpaper,
     loadWallpaperResource: suspend (Wallpaper) -> Bitmap?,
     selectedWallpaper: Wallpaper,
     onSelectWallpaper: (Wallpaper) -> Unit,
+    onLearnMoreClick: (String) -> Unit,
 ) {
     Column(
         modifier = Modifier
+            .fillMaxWidth()
+            .fillMaxHeight()
             .verticalScroll(rememberScrollState())
-            .background(color = FirefoxTheme.colors.layer1),
+            .background(color = FirefoxTheme.colors.layer1)
+            .padding(
+                end = 16.dp,
+                start = 16.dp,
+                top = 16.dp,
+            ),
     ) {
-        WallpaperThumbnails(
-            wallpapers = wallpapers,
-            defaultWallpaper = defaultWallpaper,
-            loadWallpaperResource = loadWallpaperResource,
-            selectedWallpaper = selectedWallpaper,
-            onSelectWallpaper = { updatedWallpaper -> onSelectWallpaper(updatedWallpaper) },
+        wallpaperGroups.forEach { (collection, wallpapers) ->
+            if (wallpapers.isNotEmpty()) {
+                WallpaperGroupHeading(
+                    collection = collection,
+                    onLearnMoreClick = onLearnMoreClick,
+                )
+
+                Spacer(modifier = Modifier.height(16.dp))
+
+                WallpaperThumbnails(
+                    wallpapers = wallpapers,
+                    defaultWallpaper = defaultWallpaper,
+                    loadWallpaperResource = loadWallpaperResource,
+                    selectedWallpaper = selectedWallpaper,
+                    onSelectWallpaper = onSelectWallpaper,
+                )
+
+                Spacer(modifier = Modifier.height(32.dp))
+            }
+        }
+    }
+}
+
+@Composable
+private fun WallpaperGroupHeading(
+    collection: Wallpaper.Collection,
+    onLearnMoreClick: (String) -> Unit,
+) {
+    // Since the last new collection of wallpapers was tied directly to an MR release,
+    // it was decided that we should use string resources for these titles
+    // and descriptions so they could be localized.
+    // In the future, we may want to either use the dynamic wallpaper properties with localized fallbacks
+    // or invest in a method of localizing the remote strings themselves.
+    if (collection.name == "classic-firefox") {
+        Text(
+            text = stringResource(R.string.wallpaper_classic_title),
+            color = FirefoxTheme.colors.textSecondary,
+            style = FirefoxTheme.typography.subtitle2,
         )
+    } else {
+        Column {
+            Text(
+                text = stringResource(R.string.wallpaper_limited_edition_title),
+                color = FirefoxTheme.colors.textSecondary,
+                style = FirefoxTheme.typography.subtitle2,
+            )
+
+            Spacer(modifier = Modifier.height(2.dp))
+
+            if (collection.learnMoreUrl.isNullOrEmpty()) {
+                val text = stringResource(R.string.wallpaper_limited_edition_description)
+                Text(
+                    text = text,
+                    color = FirefoxTheme.colors.textSecondary,
+                    style = FirefoxTheme.typography.caption,
+                )
+            } else {
+                val link = stringResource(R.string.wallpaper_learn_more)
+                val text = stringResource(R.string.wallpaper_limited_edition_description_with_learn_more, link)
+                val linkStartIndex = text.indexOf(link)
+                val linkEndIndex = linkStartIndex + link.length
+
+                ClickableSubstringLink(
+                    text = text,
+                    textColor = FirefoxTheme.colors.textSecondary,
+                    linkTextColor = FirefoxTheme.colors.textSecondary,
+                    linkTextDecoration = TextDecoration.Underline,
+                    clickableStartIndex = linkStartIndex,
+                    clickableEndIndex = linkEndIndex,
+                ) {
+                    onLearnMoreClick(collection.learnMoreUrl)
+                }
+            }
+        }
     }
 }
 
@@ -88,8 +169,6 @@ fun WallpaperSettings(
  * @param selectedWallpaper The currently selected wallpaper.
  * @param numColumns The number of columns that will occupy the grid.
  * @param onSelectWallpaper Action to take when a new wallpaper is selected.
- * @param verticalPadding Vertical content padding inside the block.
- * @param horizontalPadding Horizontal content padding inside the block.
  */
 @Composable
 @Suppress("LongParameterList")
@@ -100,39 +179,30 @@ fun WallpaperThumbnails(
     loadWallpaperResource: suspend (Wallpaper) -> Bitmap?,
     onSelectWallpaper: (Wallpaper) -> Unit,
     numColumns: Int = 3,
-    verticalPadding: Int = 30,
-    horizontalPadding: Int = 20,
 ) {
-    Column(
-        modifier = Modifier.padding(
-            vertical = verticalPadding.dp,
-            horizontal = horizontalPadding.dp,
-        ),
-    ) {
-        val numRows = (wallpapers.size + numColumns - 1) / numColumns
-        for (rowIndex in 0 until numRows) {
-            Row {
-                for (columnIndex in 0 until numColumns) {
-                    val itemIndex = rowIndex * numColumns + columnIndex
-                    if (itemIndex < wallpapers.size) {
-                        val wallpaper = wallpapers[itemIndex]
-                        Box(
-                            modifier = Modifier
-                                .weight(1f, fill = true)
-                                .padding(4.dp),
-                        ) {
-                            WallpaperThumbnailItem(
-                                wallpaper = wallpaper,
-                                defaultWallpaper = defaultWallpaper,
-                                loadWallpaperResource = loadWallpaperResource,
-                                isSelected = selectedWallpaper.name == wallpaper.name,
-                                isLoading = wallpaper.assetsFileState == Wallpaper.ImageFileState.Downloading,
-                                onSelect = onSelectWallpaper,
-                            )
-                        }
-                    } else {
-                        Spacer(Modifier.weight(1f))
+    val numRows = (wallpapers.size + numColumns - 1) / numColumns
+    for (rowIndex in 0 until numRows) {
+        Row {
+            for (columnIndex in 0 until numColumns) {
+                val itemIndex = rowIndex * numColumns + columnIndex
+                if (itemIndex < wallpapers.size) {
+                    val wallpaper = wallpapers[itemIndex]
+                    Box(
+                        modifier = Modifier
+                            .weight(1f, fill = true)
+                            .padding(4.dp),
+                    ) {
+                        WallpaperThumbnailItem(
+                            wallpaper = wallpaper,
+                            defaultWallpaper = defaultWallpaper,
+                            loadWallpaperResource = loadWallpaperResource,
+                            isSelected = selectedWallpaper.name == wallpaper.name,
+                            isLoading = wallpaper.assetsFileState == Wallpaper.ImageFileState.Downloading,
+                            onSelect = onSelectWallpaper,
+                        )
                     }
+                } else {
+                    Spacer(Modifier.weight(1f))
                 }
             }
         }
@@ -229,9 +299,10 @@ private fun WallpaperThumbnailsPreview() {
         WallpaperSettings(
             defaultWallpaper = Wallpaper.Default,
             loadWallpaperResource = { null },
-            wallpapers = listOf(Wallpaper.Default),
+            wallpaperGroups = mapOf(Wallpaper.DefaultCollection to listOf(Wallpaper.Default)),
             selectedWallpaper = Wallpaper.Default,
             onSelectWallpaper = {},
+            onLearnMoreClick = {},
         )
     }
 }
--- a/app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettingsFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettingsFragment.kt
@@ -17,7 +17,9 @@ import androidx.navigation.fragment.findNavController
 import kotlinx.coroutines.launch
 import mozilla.components.lib.state.ext.observeAsComposableState
 import mozilla.components.service.glean.private.NoExtras
+import org.mozilla.fenix.BrowserDirection
 import org.mozilla.fenix.GleanMetrics.Wallpapers
+import org.mozilla.fenix.HomeActivity
 import org.mozilla.fenix.R
 import org.mozilla.fenix.components.FenixSnackbar
 import org.mozilla.fenix.ext.requireComponents
@@ -54,7 +56,7 @@ class WallpaperSettingsFragment : Fragment() {
                     val coroutineScope = rememberCoroutineScope()
 
                     WallpaperSettings(
-                        wallpapers = wallpapers,
+                        wallpaperGroups = wallpapers.groupByDisplayableCollection(),
                         defaultWallpaper = Wallpaper.Default,
                         selectedWallpaper = currentWallpaper,
                         loadWallpaperResource = {
@@ -65,6 +67,13 @@ class WallpaperSettingsFragment : Fragment() {
                                 val result = wallpaperUseCases.selectWallpaper(it)
                                 onWallpaperSelected(it, result, this@apply)
                             }
+                        },
+                        onLearnMoreClick = { url ->
+                            (activity as HomeActivity).openToBrowserAndLoad(
+                                searchTermOrURL = url,
+                                newTab = true,
+                                from = BrowserDirection.FromWallpaper,
+                            )
                         },
                     )
                 }
--- a/app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperOnboarding.kt
+++ b/app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperOnboarding.kt
@@ -91,15 +91,17 @@ fun WallpaperOnboarding(
                 style = FirefoxTheme.typography.caption,
             )
 
+            Spacer(modifier = Modifier.height(16.dp))
+
             WallpaperThumbnails(
                 wallpapers = wallpapers,
                 defaultWallpaper = Wallpaper.Default,
                 selectedWallpaper = currentWallpaper,
                 loadWallpaperResource = { loadWallpaperResource(it) },
                 onSelectWallpaper = { onSelectWallpaper(it) },
-                verticalPadding = 16,
-                horizontalPadding = 0,
             )
+
+            Spacer(modifier = Modifier.height(16.dp))
 
             TextButton(
                 modifier = Modifier
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -459,12 +459,21 @@
     <string name="wallpaper_download_error_snackbar_action">Try again</string>
     <!-- Snackbar message for when wallpaper couldn't be selected because of the disk error -->
     <string name="wallpaper_select_error_snackbar_message">Couldn’t change wallpaper</string>
-
+    <!-- Text displayed that links to website containing documentation about the "Limited Edition" wallpapers. -->
+    <string name="wallpaper_learn_more">Learn more</string>
     <!-- Label for switch which toggles the "tap-to-switch" behavior on home screen logo -->
     <string name="wallpaper_tap_to_change_switch_label_1" moz:removedIn="105" tools:ignore="UnusedResources">Change wallpaper by tapping Firefox homepage logo</string>
     <!-- This is the accessibility content description for the wallpapers functionality. Users are
     able to tap on the app logo in the home screen and can switch to different wallpapers by tapping. -->
     <string name="wallpaper_logo_content_description" moz:removedIn="105" tools:ignore="UnusedResources">Firefox logo - change the wallpaper, button</string>
+    <!-- Text for classic wallpapers title. -->
+    <string name="wallpaper_classic_title">Classic Firefox</string>
+    <!-- Text for limited edition wallpapers title. -->
+    <string name="wallpaper_limited_edition_title">Limited Edition</string>
+    <!-- Description text for the limited edition wallpapers with learn more link. The first parameter is the learn more string defined in wallpaper_learn_more-->
+    <string name="wallpaper_limited_edition_description_with_learn_more">The new Independent Voices collection. %s</string>
+    <!-- Description text for the limited edition wallpapers. -->
+    <string name="wallpaper_limited_edition_description">The new Independent Voices collection.</string>
     <!-- Wallpaper onboarding dialog header text. -->
     <string name="wallpapers_onboarding_dialog_title_text">Try a splash of color</string>
     <!-- Wallpaper onboarding dialog body text. -->
--- a/app/src/test/java/org/mozilla/fenix/settings/wallpaper/ExtensionsTest.kt
+++ b/app/src/test/java/org/mozilla/fenix/settings/wallpaper/ExtensionsTest.kt
@@ -1,0 +1,88 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.settings.wallpaper
+
+import org.junit.Assert.assertEquals
+import org.junit.Test
+import org.mozilla.fenix.wallpapers.Wallpaper
+
+class ExtensionsTest {
+    private val classicCollection = getSeasonalCollection("classic-firefox")
+
+    @Test
+    fun `GIVEN wallpapers that include the default WHEN grouped by collection THEN default will be added to classic firefox`() {
+        val seasonalCollection = getSeasonalCollection("finally fall")
+        val classicFirefoxWallpapers = (0..5).map { generateClassicFirefoxWallpaper("firefox$it") }
+        val seasonalWallpapers = (0..5).map { generateSeasonalWallpaperCollection("${seasonalCollection.name}$it", seasonalCollection.name) }
+        val allWallpapers = listOf(Wallpaper.Default) + classicFirefoxWallpapers + seasonalWallpapers
+
+        val result = allWallpapers.groupByDisplayableCollection()
+
+        assertEquals(2, result.size)
+        assertEquals(listOf(Wallpaper.Default) + classicFirefoxWallpapers, result[classicCollection])
+        assertEquals(seasonalWallpapers, result[seasonalCollection])
+    }
+
+    @Test
+    fun `GIVEN no wallpapers but the default WHEN grouped by collection THEN the default will still be present`() {
+        val result = listOf(Wallpaper.Default).groupByDisplayableCollection()
+
+        assertEquals(1, result.size)
+        assertEquals(listOf(Wallpaper.Default), result[Wallpaper.DefaultCollection])
+    }
+
+    @Test
+    fun `GIVEN wallpapers with thumbnails that have not downloaded WHEN grouped by collection THEN wallpapers without thumbnails will not be included`() {
+        val seasonalCollection = getSeasonalCollection("finally fall")
+        val classicFirefoxWallpapers = (0..5).map { generateClassicFirefoxWallpaper("firefox$it") }
+        val downloadedSeasonalWallpapers = (0..5).map { generateSeasonalWallpaperCollection("${seasonalCollection.name}$it", seasonalCollection.name) }
+        val nonDownloadedSeasonalWallpapers = (0..5).map {
+            generateSeasonalWallpaperCollection(
+                "${seasonalCollection.name}$it",
+                seasonalCollection.name,
+                Wallpaper.ImageFileState.Error,
+            )
+        }
+        val allWallpapers = listOf(Wallpaper.Default) + classicFirefoxWallpapers + downloadedSeasonalWallpapers + nonDownloadedSeasonalWallpapers
+
+        val result = allWallpapers.groupByDisplayableCollection()
+
+        assertEquals(2, result.size)
+        assertEquals(listOf(Wallpaper.Default) + classicFirefoxWallpapers, result[classicCollection])
+        assertEquals(downloadedSeasonalWallpapers, result[seasonalCollection])
+    }
+
+    private fun generateClassicFirefoxWallpaper(name: String) = Wallpaper(
+        name = name,
+        textColor = 0L,
+        cardColor = 0L,
+        thumbnailFileState = Wallpaper.ImageFileState.Downloaded,
+        assetsFileState = Wallpaper.ImageFileState.Downloaded,
+        collection = classicCollection,
+    )
+
+    private fun getSeasonalCollection(name: String) = Wallpaper.Collection(
+        name = name,
+        heading = null,
+        description = null,
+        learnMoreUrl = null,
+        availableLocales = null,
+        startDate = null,
+        endDate = null,
+    )
+
+    private fun generateSeasonalWallpaperCollection(
+        wallpaperName: String,
+        collectionName: String,
+        thumbnailState: Wallpaper.ImageFileState = Wallpaper.ImageFileState.Downloaded,
+    ) = Wallpaper(
+        name = wallpaperName,
+        textColor = 0L,
+        cardColor = 0L,
+        thumbnailFileState = thumbnailState,
+        assetsFileState = Wallpaper.ImageFileState.Downloaded,
+        collection = getSeasonalCollection(collectionName),
+    )
+}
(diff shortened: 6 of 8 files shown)


app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -152,33 +256,44 @@ private fun WallpaperSnackbar(
...
-                                isSelected = selectedWallpaper == wallpapers[itemIndex],
-                                onSelect = onSelectWallpaper,
-                            )
+    wallpapers.forEach { (group, wallpapers) ->
+        if (wallpapers.isNotEmpty()) {
+            WallpaperGroups(
+                title = group.title,
+                onLearnMoreClick = onLearnMoreClick,
+            )
+        }
```
MatthewTighe, line 273: I think we should move this bit up a level to `WallpaperSettings`. This will make it easier to re-use `WallpaperThumbnails`, as @mavduevskiy needs to in https://github.com/mozilla-mobile/fenix/pull/26758

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -68,16 +74,40 @@ import org.mozilla.fenix.wallpapers.Wallpaper
...
+            title = it.key,
+        )
+    }.sortedBy { it.title == "default" }
+
+    val maps = HashMap<WallpaperProviderGroup, List<Wallpaper>>()
+    val scaffoldState = rememberScaffoldState()
+    val coroutineScope = rememberCoroutineScope()
+    groups.forEach {
+        maps[it] = it.wallpapers
+    }
```
MatthewTighe, line 110: I think we should move this grouping logic into an extension function located in a new file at `wallpapers/Extensions.kt` that has the signature `List<Wallpaper>.groupByCollection(): Map<Collection, List<Wallpaper>>`. We talked offline a bit, but this is an example suggestion that I think would get the behavior we desire here:

```
fun List<Wallpaper>.groupByCollection(): Map<Wallpaper.Collection, List<Wallpaper>> = groupBy {
    it.collection
}.filter {
    it.key.name != "default"
}.map {
    if (it.key.name == "classic-firefox") {
        it.key to listOf(Wallpaper.Default) + it.value
    } else {
        it.key to it.value
    }
}.toMap()
```

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -107,6 +143,74 @@ fun WallpaperSettings(
...
+ * @property limit The maximum number of wallpapers that will be shown in this group.
+ * @property id A unique ID for this group (uses a generated UUID by default)
+ */
+
+data class WallpaperProviderGroup(
+    var wallpapers: List<Wallpaper>,
+    var title: String? = null,
+    val limit: Int = Integer.MAX_VALUE,
+    val id: String = UUID.randomUUID().toString(),
+)
```
MatthewTighe, line 160: I don't think this data class will be necessary if we decide to use an extension function to group things as I suggested above.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettingsFragment.kt
```diff
@@ -52,6 +55,15 @@ class WallpaperSettingsFragment : Fragment() {
                     var coroutineScope = rememberCoroutineScope()
 
                     WallpaperSettings(
+                        onLearnMoreClick = {
+                            (activity as HomeActivity).openToBrowserAndLoad(
+                                searchTermOrURL = SupportUtils.getGenericSumoURLForTopic(
```
MatthewTighe, line 60: I don't think we will want to convert this into a SUMO URL, as I believe we'll be given links to exact web pages. Even if we end up having a SUMO article for this particular collection, I think in the future it would be possible for this link to lead to a partner site, for example.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -107,6 +143,74 @@ fun WallpaperSettings(
...
+                modifier = Modifier.padding(bottom = 2.dp),
+                color = FirefoxTheme.colors.textSecondary,
+                fontSize = 14.sp,
+            )
+
+            title.let {
+                Row(modifier = Modifier.fillMaxWidth()) {
+                    if (it != null) {
+                        Text(
+                            text = "$it. ",
```
MatthewTighe, line 187: I think we should also introduce an extension function to get a more human friendly version of a collection title. Something like 
```
fun Wallpaper.Collection.getDisplayName(): String = this.name
    .split('-')
    .joinToString(" ") { word -> 
        word.replaceFirstChar { 
            if (it.isLowerCase()) it.titlecase(Locale.getDefault()) else it.toString() 
        } 
    }
```

app/src/main/java/org/mozilla/fenix/settings/SupportUtils.kt
```diff
@@ -54,6 +54,7 @@ object SupportUtils {
         SMARTBLOCK("smartblock-enhanced-tracking-protection"),
         SPONSOR_PRIVACY("sponsor-privacy"),
         HTTPS_ONLY_MODE("https-only-mode-firefox-android"),
+        WALLPAPER_LEARN_MORE("enhanced-tracking-protection-android"),
```
MatthewTighe, line 57: I don't think we will need this enum, given my comment about links below.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -6,52 +6,29 @@ package org.mozilla.fenix.settings.wallpaper
...
-import androidx.compose.foundation.layout.Row
-import androidx.compose.foundation.layout.Spacer
-import androidx.compose.foundation.layout.aspectRatio
-import androidx.compose.foundation.layout.fillMaxSize
-import androidx.compose.foundation.layout.fillMaxWidth
-import androidx.compose.foundation.layout.heightIn
-import androidx.compose.foundation.layout.padding
-import androidx.compose.foundation.rememberScrollState
+import androidx.compose.foundation.*
+import androidx.compose.foundation.layout.*
```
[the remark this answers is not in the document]
sarah541: Thanks

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -68,74 +62,92 @@ import org.mozilla.fenix.wallpapers.Wallpaper
 @Composable
 @Suppress("LongParameterList")
 fun WallpaperSettings(
-    wallpapers: List<Wallpaper>,
+    onLearnMoreClick: (String) -> Unit,
+    wallpaperGroups: Map<Wallpaper.Collection, List<Wallpaper>>,
```
MatthewTighe, line 66: We will want doc strings for these, and I think we should move the `onLearnMoreClicked` down with the rest of the lambda parameters.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -68,74 +62,92 @@ import org.mozilla.fenix.wallpapers.Wallpaper
...
+                        collection,
+                    )
+                    WallpaperThumbnails(
+                        wallpapers = wallpapers,
+                        defaultWallpaper = defaultWallpaper,
+                        loadWallpaperResource = loadWallpaperResource,
+                        selectedWallpaper = selectedWallpaper,
+                        onSelectWallpaper = { updatedWallpaper ->
+                            onSelectWallpaper(updatedWallpaper)
+                        },
```
MatthewTighe, line 94:
```suggestion
                        onSelectWallpaper = onSelectWallpaper
```

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -68,74 +62,92 @@ import org.mozilla.fenix.wallpapers.Wallpaper
...
         }
     }
 }
 
 @Composable
-private fun WallpaperSnackbar(
-    onViewWallpaper: () -> Unit,
+private fun WallpaperGroupHeading(
+    onLearnMoreClick: (String) -> Unit,
+    collection: Wallpaper.Collection,
```
MatthewTighe, line 105: I think we should re-order these so the lambda is at the bottom

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -68,74 +62,92 @@ import org.mozilla.fenix.wallpapers.Wallpaper
...
+                        modifier = Modifier.clickable { onLearnMoreClick(collection.learnMoreUrl) },
+                        style = FirefoxTheme.typography.caption.copy(
+                            textDecoration = TextDecoration.Underline,
+                        ),
+                    )
+                }
+            }
+        } else {
+            Text(
+                text = "Classic Firefox",
```
MatthewTighe, line 145: This should probably be a string resource, since it doesn't need to be dynamic and can be translated.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettingsFragment.kt
```diff
@@ -49,10 +51,17 @@ class WallpaperSettingsFragment : Fragment() {
                         state.wallpaperState.currentWallpaper
                     }.value ?: Wallpaper.Default
 
-                    var coroutineScope = rememberCoroutineScope()
+                    val coroutineScope = rememberCoroutineScope()
 
                     WallpaperSettings(
-                        wallpapers = wallpapers,
+                        onLearnMoreClick = { url->
```
MatthewTighe, line 57: Let's move this lambda definition down with the others here as well

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -68,74 +62,92 @@ import org.mozilla.fenix.wallpapers.Wallpaper
...
-            TextButton(
-                text = stringResource(R.string.wallpaper_updated_snackbar_action),
-                onClick = onViewWallpaper,
-                modifier = Modifier.padding(all = 8.dp),
-                textColor = FirefoxTheme.colors.textOnColorPrimary,
+
+            Row(modifier = Modifier.fillMaxWidth()) {
+                collection.description?.let {
+                    Text(
+                        text = "$it. ",
```
MatthewTighe, line 126: It looks like UX is planning to include punctuation in the descriptions, so we can remove the period from the text parameter here.

MatthewTighe: Actually, come to think of it I believe the intent for this release is to have the strings all included as resources. So we will  need a text resource for the description that is:
`The new Independent Voices collection.` and a text resource for the heading that is `Limited edition`. We can revisit doing this more dynamically in a follow-up to the MR. 

The text resources will be instead of using the properties defined on the `collection`.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/Extensions.kt
```diff
@@ -0,0 +1,35 @@
...
+ * The extension function for formatting wallpaper collection title.
+ **/
+
+fun Wallpaper.Collection.getDisplayName(): String = this.name
+    .split('-')
+    .joinToString(" ") { word ->
+        word.replaceFirstChar {
+            if (it.isLowerCase()) it.titlecase(Locale.getDefault()) else it.toString()
+        }
+    }
```
MatthewTighe, line 35: We can remove this since we will just use a string resource for "Classic Firefox"

MatthewTighe (commented):
First pass, had some comments. This is looking good though, I think we're close! Most of the comments are my fault from when we were pairing anyway 😅

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettingsFragment.kt
```diff
@@ -49,10 +51,17 @@ class WallpaperSettingsFragment : Fragment() {
...
 
                     WallpaperSettings(
-                        wallpapers = wallpapers,
+                        onLearnMoreClick = { url ->
+                            (activity as HomeActivity).openToBrowserAndLoad(
+                                searchTermOrURL = url,
+                                newTab = true,
+                                from = BrowserDirection.FromWallpaper,
+                            )
+                        },
```
MatthewTighe, line 63: nit: can this lambda parameter be moved down with the other lambdas?

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -57,85 +51,100 @@ import org.mozilla.fenix.wallpapers.Wallpaper
...
-        action = {
-            TextButton(
-                text = stringResource(R.string.wallpaper_updated_snackbar_action),
-                onClick = onViewWallpaper,
-                modifier = Modifier.padding(all = 8.dp),
-                textColor = FirefoxTheme.colors.textOnColorPrimary,
+
+            Row(modifier = Modifier.fillMaxWidth()) {
+                Text(
+                    text = stringResource(R.string.seasonal_wallpaper_description),
```
MatthewTighe, line 123: nit: Would you leave a comment about these string resource that reads something like:

```
// Since the last new collection of wallpapers was tied directly to an MR release, it was decided that we should use string resources for these titles and descriptions so they could be localized. In the future, we may want to either use the dynamic wallpaper properties with localized fallbacks or invest in a method of localizing the remote strings themselves. 
```

app/src/main/java/org/mozilla/fenix/settings/wallpaper/Extensions.kt
```diff
@@ -0,0 +1,21 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.wallpapers
+
+/**
+ * The extension function to group wallpapers according to their name.
+ **/
+
```
gabrielluong, line 10: nit remove new line

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -48,34 +53,103 @@ import org.mozilla.fenix.wallpapers.Wallpaper
...
+                            textDecoration = TextDecoration.Underline,
+                        ),
+                    )
+                }
+            }
+        } else {
+            Text(
+                text = stringResource(R.string.classic_wallpaper_title),
+                color = FirefoxTheme.colors.textSecondary,
+                fontSize = 14.sp,
```
gabrielluong, line 150: Is there a typography we can use for this?

app/src/main/res/values/strings.xml
```diff
@@ -459,19 +459,27 @@
...
+    <string name="seasonal_wallpaper_description">The new Independent Voices collection.</string>
     <!-- Wallpaper onboarding dialog header text. -->
     <string name="wallpapers_onboarding_dialog_title_text">Try a splash of color</string>
     <!-- Wallpaper onboarding dialog body text. -->
     <string name="wallpapers_onboarding_dialog_body_text">Choose a wallpaper that speaks to you.</string>
     <!-- Wallpaper onboarding dialog learn more button text. The button navigates to the wallpaper settings screen. -->
     <string name="wallpapers_onboarding_dialog_explore_more_button_text">Explore more wallpapers</string>
 
+    <!-- Label for switch which toggles the "tap-to-switch" behavior on home screen logo -->
+    <string name="wallpaper_tap_to_change_switch_label_1" moz:removedIn="105" tools:ignore="UnusedResources">Change wallpaper by tapping Firefox homepage logo</string>
```
gabrielluong, line 478:
```suggestion
    <string name="wallpaper_tap_to_change_switch_label_1" moz:removedIn="106" tools:ignore="UnusedResources">Change wallpaper by tapping Firefox homepage logo</string>
```

app/src/main/res/values/strings.xml
```diff
@@ -459,19 +459,27 @@
     <string name="wallpaper_download_error_snackbar_action">Try again</string>
     <!-- Snackbar message for when wallpaper couldn't be selected because of the disk error -->
     <string name="wallpaper_select_error_snackbar_message">Couldn’t change wallpaper</string>
-
-    <!-- Label for switch which toggles the "tap-to-switch" behavior on home screen logo -->
-    <string name="wallpaper_tap_to_change_switch_label_1" moz:removedIn="105" tools:ignore="UnusedResources">Change wallpaper by tapping Firefox homepage logo</string>
-    <!-- This is the accessibility content description for the wallpapers functionality. Users are
-    able to tap on the app logo in the home screen and can switch to different wallpapers by tapping. -->
-    <string name="wallpaper_logo_content_description" moz:removedIn="105" tools:ignore="UnusedResources">Firefox logo - change the wallpaper, button</string>
```
gabrielluong, line 467: Let's undo the move for these strings.

app/src/main/res/values/strings.xml
```diff
@@ -459,19 +459,27 @@
...
+    <!-- Text displayed that links to website containing documentation about the "Seasonal wallpapers". -->
+    <string name="wallpaper_learn_more">Learn more</string>
+    <!-- Text for classic wallpapers title. -->
+    <string name="classic_wallpaper_title">Classic Firefox</string>
+    <!-- Text for seasonal wallpapers title. -->
+    <string name="seasonal_wallpaper_title">Limited Edition</string>
+    <!-- Text for seasonal wallpapers description. -->
+    <string name="seasonal_wallpaper_description">The new Independent Voices collection.</string>
```
gabrielluong, line 469: Since all the wallpaper strings start with the prefix `wallpaper`
```suggestion
    <string name="wallpaper_classic_title">Classic Firefox</string>
    <!-- Text for limited edition wallpapers title. -->
    <string name="wallpaper_limited_edition_title">Limited Edition</string>
    <!-- Description text for the limited edition wallpapers. -->
    <string name="wallpaper_limited_edition_description">The new Independent Voices collection.</string>
```

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -48,34 +53,103 @@ import org.mozilla.fenix.wallpapers.Wallpaper
...
-            defaultWallpaper = defaultWallpaper,
-            loadWallpaperResource = loadWallpaperResource,
-            selectedWallpaper = selectedWallpaper,
-            onSelectWallpaper = { updatedWallpaper -> onSelectWallpaper(updatedWallpaper) },
-        )
+        wallpaperGroups.forEach { (collection, wallpapers) ->
+            if (wallpapers.isNotEmpty()) {
+                WallpaperGroupHeading(
+                    collection,
+                    onLearnMoreClick,
```
gabrielluong, line 85: Could you add the named parameters here as well?

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -48,34 +53,103 @@ import org.mozilla.fenix.wallpapers.Wallpaper
...
+                    style = FirefoxTheme.typography.caption,
+                )
+
+                collection.learnMoreUrl?.let {
+                    Text(
+                        text = stringResource(R.string.wallpaper_learn_more),
+                        color = FirefoxTheme.colors.textSecondary,
+                        modifier = Modifier.clickable { onLearnMoreClick(collection.learnMoreUrl) },
+                        style = FirefoxTheme.typography.caption.copy(
+                            textDecoration = TextDecoration.Underline,
```
gabrielluong, line 141: Let's add the `textDecoration` outside the TextStyle since `Text` has a `textDecoration` parameter anyways

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -227,9 +301,10 @@ private fun WallpaperThumbnailItem(
 private fun WallpaperThumbnailsPreview() {
     FirefoxTheme(theme = Theme.getTheme()) {
         WallpaperSettings(
+            onLearnMoreClick = {},
```
gabrielluong, line 304: Could we reorder this to follow the order of the parameters defined by `WallpaperSettings` (eg, move it to the bottom)

app/src/main/res/values/strings.xml
```diff
@@ -459,19 +459,27 @@
     <string name="wallpaper_download_error_snackbar_action">Try again</string>
     <!-- Snackbar message for when wallpaper couldn't be selected because of the disk error -->
     <string name="wallpaper_select_error_snackbar_message">Couldn’t change wallpaper</string>
-
-    <!-- Label for switch which toggles the "tap-to-switch" behavior on home screen logo -->
-    <string name="wallpaper_tap_to_change_switch_label_1" moz:removedIn="105" tools:ignore="UnusedResources">Change wallpaper by tapping Firefox homepage logo</string>
-    <!-- This is the accessibility content description for the wallpapers functionality. Users are
-    able to tap on the app logo in the home screen and can switch to different wallpapers by tapping. -->
-    <string name="wallpaper_logo_content_description" moz:removedIn="105" tools:ignore="UnusedResources">Firefox logo - change the wallpaper, button</string>
+    <!-- Text displayed that links to website containing documentation about the "Seasonal wallpapers". -->
```
gabrielluong, line 462: It's weird to me see "Seasonal wallpapers" when the string for the seasonal wallpaper is "Limited Edition". I would try to be consistent with the terminology here and what is displayed to the user.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/Extensions.kt
```diff
@@ -0,0 +1,21 @@
...
+fun List<Wallpaper>.groupByCollection(): Map<Wallpaper.Collection, List<Wallpaper>> = groupBy {
```
gabrielluong, line 11: Might be nice to add a unit test for this

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -48,34 +53,103 @@ import org.mozilla.fenix.wallpapers.Wallpaper
...
+    ) {
+        // Since the last new collection of wallpapers was tied directly to an MR release,
+        // it was decided that we should use string resources for these titles
+        // and descriptions so they could be localized.
+        // In the future, we may want to either use the dynamic wallpaper properties with localized fallbacks
+        // or invest in a method of localizing the remote strings themselves.
+        if (collection.name != "classic-firefox") {
+            Text(
+                text = stringResource(R.string.seasonal_wallpaper_title),
+                modifier = Modifier.padding(bottom = 2.dp),
```
gabrielluong, line 122: I would use a `Spacer(height = 2.dp)` to make this more explicit

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
gabrielluong: Looks like this can be subtitle2 as well like the above title.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -48,34 +53,103 @@ import org.mozilla.fenix.wallpapers.Wallpaper
...
+@Composable
+private fun WallpaperGroupHeading(
+    collection: Wallpaper.Collection,
+    onLearnMoreClick: (String) -> Unit,
+) {
+    Column(
+        modifier = Modifier.padding(
+            top = 16.dp,
+            start = 18.dp,
+            end = 18.dp,
```
gabrielluong, line 110: We could add this horizontal padding at the top level Column container on line 74. This would simplify things and you won't need to set the horizontal padding in `WallpaperGroupHeading` and `WallpaperThumbnails`

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -48,34 +53,103 @@ import org.mozilla.fenix.wallpapers.Wallpaper
...
+                    onLearnMoreClick,
+                )
+                WallpaperThumbnails(
+                    wallpapers = wallpapers,
+                    defaultWallpaper = defaultWallpaper,
+                    loadWallpaperResource = loadWallpaperResource,
+                    selectedWallpaper = selectedWallpaper,
+                    onSelectWallpaper = onSelectWallpaper,
+                    verticalPadding = 0,
+                    horizontalPadding = 18,
```
gabrielluong, line 94: This would be 0 if you add the horizontal padding at the top level Column. See comment below

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -48,34 +53,103 @@ import org.mozilla.fenix.wallpapers.Wallpaper
...
-            selectedWallpaper = selectedWallpaper,
-            onSelectWallpaper = { updatedWallpaper -> onSelectWallpaper(updatedWallpaper) },
-        )
+        wallpaperGroups.forEach { (collection, wallpapers) ->
+            if (wallpapers.isNotEmpty()) {
+                WallpaperGroupHeading(
+                    collection,
+                    onLearnMoreClick,
+                )
+                WallpaperThumbnails(
```
gabrielluong, line 87: Add a new line before to visually separate the 2 Composable for easier readability.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -48,34 +53,103 @@ import org.mozilla.fenix.wallpapers.Wallpaper
...
+
+                collection.learnMoreUrl?.let {
+                    Text(
+                        text = stringResource(R.string.wallpaper_learn_more),
+                        color = FirefoxTheme.colors.textSecondary,
+                        modifier = Modifier.clickable { onLearnMoreClick(collection.learnMoreUrl) },
+                        style = FirefoxTheme.typography.caption.copy(
+                            textDecoration = TextDecoration.Underline,
+                        ),
+                    )
```
gabrielluong, line 143: You will want to use `ClickableSubstringLink` here actually. I realize the strings for this can't be separated from a localization perspective.

app/src/main/res/values/strings.xml
gabrielluong, line 469:
```suggestion
    <string name="seasonal_wallpaper_description">The new Independent Voices collection. %s</string>
```
This string actually needs to take a parameter, where %s is "Learn more". The reasoning for this is that when it comes to localization, you can't simply join 2 strings together and assume the order because this won't be the cause for RTL languages.

gabrielluong: I imagine you will also need a string without the %s for situations where there is no learn more URL.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -48,34 +53,103 @@ import org.mozilla.fenix.wallpapers.Wallpaper
...
+            end = 18.dp,
+            bottom = 12.dp,
+        ),
+    ) {
+        // Since the last new collection of wallpapers was tied directly to an MR release,
+        // it was decided that we should use string resources for these titles
+        // and descriptions so they could be localized.
+        // In the future, we may want to either use the dynamic wallpaper properties with localized fallbacks
+        // or invest in a method of localizing the remote strings themselves.
+        if (collection.name != "classic-firefox") {
```
gabrielluong, line 119: Maybe flip the order of the blocks in the if statement, it would actually be preferable to do `collection.name == "classic-firefox"` for easier comprehension.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -48,34 +53,103 @@ import org.mozilla.fenix.wallpapers.Wallpaper
...
+                WallpaperThumbnails(
+                    wallpapers = wallpapers,
+                    defaultWallpaper = defaultWallpaper,
+                    loadWallpaperResource = loadWallpaperResource,
+                    selectedWallpaper = selectedWallpaper,
+                    onSelectWallpaper = onSelectWallpaper,
+                    verticalPadding = 0,
+                    horizontalPadding = 18,
+                )
+            }
```
gabrielluong, line 96: I think you need to add a Spacer of 32dp in between the groups

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -48,34 +53,103 @@ import org.mozilla.fenix.wallpapers.Wallpaper
...
+}
+
+@Composable
+private fun WallpaperGroupHeading(
+    collection: Wallpaper.Collection,
+    onLearnMoreClick: (String) -> Unit,
+) {
+    Column(
+        modifier = Modifier.padding(
+            top = 16.dp,
```
gabrielluong, line 108: Move this top padding to the top level Column container on line 74. You will have a separate 32dp Spacer in between the groups.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/Extensions.kt
```diff
@@ -0,0 +1,21 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.wallpapers
```
mavduevskiy, line 5: actually, package is different :) it should be
`package org.mozilla.fenix.settings.wallpaper`

mavduevskiy, line 5: actually, package is different :) it should be
`package org.mozilla.fenix.settings.wallpaper`

app/src/main/java/org/mozilla/fenix/settings/wallpaper/Extensions.kt
mavduevskiy: Just as a side note, wouldn't it make sense to have a Collection object having a list of wallpapers inside? We don't really need that `Map<Wallpaper.Collection, List<Wallpaper>>` map, do we?

app/src/main/java/org/mozilla/fenix/settings/wallpaper/Extensions.kt
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * The extension function to group wallpapers according to their name.
+ **/
+
+fun List<Wallpaper>.groupByCollection(): Map<Wallpaper.Collection, List<Wallpaper>> = groupBy {
+    it.collection
+}.filter {
+    it.key.name != "default"
```
mavduevskiy, line 14: May be we could use `org.mozilla.fenix.wallpapers.Wallpaper.Companion.defaultName` to reduce the number of hard-coded elements? I am trying to come up with the way to replace `classic-firefox`, but fail.

mavduevskiy:
Tested it locally, looks nice! Some things I have noticed. I have empty element on the right side.
image

If I open clear app with no connection (no thumbnails downloaded), there screen is completely empty, even without the default option. I remember previously in such a case we would still have the default, no wallpaper, option.

Also, following that case, after enabling internet the screen would be still empty. It looks like we don't try to download thumbnails again if they are not loaded, I remember having an issue or something about it. I am not sure if it's within the scope of the task, just reaching out @MatthewTighe about what I have noticed

MatthewTighe:
>  After debugging, the wallpaper class in their, but it's thumbnailLodwing state is ERROR. I believe we don't want to try to show items with no downloaded thumbnails, seems like filtering them out would be a good idea.

Agreed, since this logic seems fine to re-use for grouping the wallpapers on the selection tool I think filtering them as part of `groupByCollection` is good enough for now, though I would perhaps rename it to `groupByDisplayableCollection` or something similar.

> If I open clear app with no connection (no thumbnails downloaded), there screen is completely empty, even without the default option. I remember previously in such a case we would still have the default, no wallpaper, option.

Same as above. Given that, an updated version could look like:

```
fun List<Wallpaper>.groupByCollection(): Map<Wallpaper.Collection, List<Wallpaper>> = groupBy {
    it.collection
}.filter {
    it.key.name != "default"
}.map {
    val wallpapers = it.value.filter { wallpaper -> 
        wallpaper.thumbnailFileState == Wallpaper.ImageFileState.Downloaded 
    }
    if (it.key.name == "classic-firefox") {
        it.key to listOf(Wallpaper.Default) + wallpapers
    } else {
        it.key to wallpapers
    }
}.toMap().takeIf { 
    it.isNotEmpty() 
} ?: mapOf(Wallpaper.DefaultCollection to listOf(Wallpaper.Default))
```

> Also, following that case, after enabling internet the screen would be still empty. It looks like we don't try to download thumbnails again if they are not loaded, I remember having an issue or something about it. I am not sure if it's within the scope of the task, just reaching out @MatthewTighe about what I have noticed

This is true and definitely not perfect. It would have been good to implement some kind of retry mechanism. Now, the app will only retry when it is restarted. I think that will be good enough for now given that we don't show the onboarding tool until the thumbnails have downloaded.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/Extensions.kt
gabrielluong: This is still using the wrong package

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -48,33 +56,111 @@ import org.mozilla.fenix.wallpapers.Wallpaper
...
+
+        ClickableText(
+            text = annotatedString,
+            style = FirefoxTheme.typography.caption,
+            onClick = { offset ->
+                annotatedString.getStringAnnotations(tag = "learn", start = offset, end = offset)
+                    .firstOrNull()?.let {
+                        collection.learnMoreUrl?.let { it1 -> onLearnMoreClick(it1) }
+                    }
+            },
```
gabrielluong, line 163: This isn't quite right. Can you look at https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesComposables.kt#398 as an example?

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -48,33 +56,111 @@ import org.mozilla.fenix.wallpapers.Wallpaper
...
+            color = FirefoxTheme.colors.textSecondary,
+            style = FirefoxTheme.typography.subtitle2,
+        )
+    } else {
+        Text(
+            text = stringResource(R.string.wallpaper_limited_edition_title),
+            color = FirefoxTheme.colors.textSecondary,
+            style = FirefoxTheme.typography.subtitle2,
+        )
+        Spacer(modifier = Modifier.height(2.dp))
```
gabrielluong, line 134: nit: add a new line above.

app/src/test/java/org/mozilla/fenix/settings/wallpaper/ExtensionsTest.kt
```diff
@@ -0,0 +1,93 @@
+package org.mozilla.fenix.settings.wallpaper
```
gabrielluong, line 1: nit: needs a license

app/src/test/java/org/mozilla/fenix/settings/wallpaper/ExtensionsTest.kt
```diff
@@ -0,0 +1,93 @@
...
+        val allWallpapers = listOf(Wallpaper.Default) + classicFirefoxWallpapers + downloadedSeasonalWallpapers + nonDownloadedSeasonalWallpapers
+
+        val result = allWallpapers.groupByDisplayableCollection()
+
+        assertEquals(2, result.size)
+        assertEquals(listOf(Wallpaper.Default) + classicFirefoxWallpapers, result[classicCollection])
+        assertEquals(downloadedSeasonalWallpapers, result[seasonalCollection])
+    }
+
+    private val classicCollection = Wallpaper.Collection(
```
gabrielluong, line 52: This can be constructed using `getSeasonalCollection("classic-firefox")`

app/src/test/java/org/mozilla/fenix/settings/wallpaper/ExtensionsTest.kt
```diff
@@ -0,0 +1,93 @@
...
+    )
+
+    private fun getSeasonalCollection(name: String) = Wallpaper.Collection(
+        name = name,
+        null,
+        null,
+        null,
+        null,
+        null,
+        null,
```
gabrielluong, line 78: Can you add all the named parameters? This will futureproof us if we ever have to refactor any of these parameters in `Wallpaper.Collection`

app/src/main/res/values/strings.xml
```diff
@@ -459,12 +459,19 @@
...
     <string name="wallpaper_tap_to_change_switch_label_1" moz:removedIn="105" tools:ignore="UnusedResources">Change wallpaper by tapping Firefox homepage logo</string>
     <!-- This is the accessibility content description for the wallpapers functionality. Users are
     able to tap on the app logo in the home screen and can switch to different wallpapers by tapping. -->
     <string name="wallpaper_logo_content_description" moz:removedIn="105" tools:ignore="UnusedResources">Firefox logo - change the wallpaper, button</string>
+    <!-- Text for classic wallpapers title. -->
+    <string name="wallpaper_classic_title">Classic Firefox</string>
+    <!-- Text for limited edition wallpapers title. -->
+    <string name="wallpaper_limited_edition_title">Limited Edition</string>
+    <!-- Description text for the limited edition wallpapers. -->
+    <string name="wallpaper_limited_edition_description">The new Independent Voices collection.. </string>
```
gabrielluong, line 474:
```suggestion
    <string name="wallpaper_limited_edition_description">The new Independent Voices collection.</string>
```
This still isn't correct. You want a string with "The new Independent Voices collection. %s" to append the "Learn more".

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -48,34 +53,109 @@ import org.mozilla.fenix.wallpapers.Wallpaper
...
+                val linkEndIndex = linkStartIndex + link.length
+                ClickableSubstringLink(
+                    text = text,
+                    textColor = FirefoxTheme.colors.textSecondary,
+                    linkTextColor = FirefoxTheme.colors.textSecondary,
+                    linkTextDecoration = TextDecoration.Underline,
+                    clickableStartIndex = linkStartIndex,
+                    clickableEndIndex = linkEndIndex,
+                ) {
+                    onLearnMoreClick(collection.learnMoreUrl!!)
```
gabrielluong, line 148: Remove the !!

app/src/main/java/org/mozilla/fenix/wallpapers/WallpaperOnboarding.kt
```diff
@@ -91,14 +91,14 @@ fun WallpaperOnboarding(
                 style = FirefoxTheme.typography.caption,
             )
 
+            Spacer(modifier = Modifier.height(16.dp))
```
gabrielluong, line 94: We need to add a Spacer below the `WallpaperThumbnails` as well

app/src/test/java/org/mozilla/fenix/settings/wallpaper/ExtensionsTest.kt
```diff
@@ -0,0 +1,88 @@
...
+        val allWallpapers = listOf(Wallpaper.Default) + classicFirefoxWallpapers + downloadedSeasonalWallpapers + nonDownloadedSeasonalWallpapers
+
+        val result = allWallpapers.groupByDisplayableCollection()
+
+        assertEquals(2, result.size)
+        assertEquals(listOf(Wallpaper.Default) + classicFirefoxWallpapers, result[classicCollection])
+        assertEquals(downloadedSeasonalWallpapers, result[seasonalCollection])
+    }
+
+    private val classicCollection = getSeasonalCollection("classic-firefox")
```
gabrielluong, line 55: Let's move this to the top

app/src/main/res/values/strings.xml
```diff
@@ -459,12 +459,21 @@
...
     <!-- Label for switch which toggles the "tap-to-switch" behavior on home screen logo -->
     <string name="wallpaper_tap_to_change_switch_label_1" moz:removedIn="105" tools:ignore="UnusedResources">Change wallpaper by tapping Firefox homepage logo</string>
     <!-- This is the accessibility content description for the wallpapers functionality. Users are
     able to tap on the app logo in the home screen and can switch to different wallpapers by tapping. -->
     <string name="wallpaper_logo_content_description" moz:removedIn="105" tools:ignore="UnusedResources">Firefox logo - change the wallpaper, button</string>
+    <!-- Text for classic wallpapers title. -->
+    <string name="wallpaper_classic_title">Classic Firefox</string>
+    <!-- Text for limited edition wallpapers title. -->
+    <string name="wallpaper_limited_edition_title">Limited Edition</string>
+    <!-- Description text for the limited edition wallpapers with learn more link. -->
```
gabrielluong, line 473:
```suggestion
    <!-- Description text for the limited edition wallpapers with learn more link. The first parameter is the learn more string defined in wallpaper_learn_more. -->
```

app/src/main/res/values/strings.xml
```diff
@@ -459,12 +459,21 @@
...
+    <string name="wallpaper_limited_edition_description_with_learn_more">The new Independent Voices collection. %s </string>
```
gabrielluong, line 474:
```suggestion
    <string name="wallpaper_limited_edition_description_with_learn_more">The new Independent Voices collection. %s</string>
```

app/src/main/res/values/strings.xml
```diff
@@ -459,12 +459,21 @@
...
+    <!-- Description text for the limited edition wallpapers. -->
+    <string name="wallpaper_limited_edition_description">The new Independent Voices collection. </string>
```
gabrielluong, line 476:
```suggestion
    <string name="wallpaper_limited_edition_description">The new Independent Voices collection.</string>
```

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -48,34 +53,109 @@ import org.mozilla.fenix.wallpapers.Wallpaper
...
+        Column {
+            Text(
+                text = stringResource(R.string.wallpaper_limited_edition_title),
+                color = FirefoxTheme.colors.textSecondary,
+                style = FirefoxTheme.typography.subtitle2,
+            )
+
+            Spacer(modifier = Modifier.height(2.dp))
+
+            if (!collection.learnMoreUrl.isNullOrEmpty()) {
```
gabrielluong, line 135: Let's flip the blocks here

gabrielluong (approved):

app/src/main/java/org/mozilla/fenix/settings/wallpaper/Extensions.kt
sarah541: @mavduevskiy , we can pick this up in a follow up PR

MatthewTighe (approved):
LGTM, thanks!

Issue #26884, opened 2022-09-08
Crash in CFRPopupFullScreenLayout.show @ [CFRPopupFullscreenLayout.kt]


AaronMT:
Seeing this overnight (09/08) in very recent CFR changes on Firebase Test Lab 
@gabrielluong @mcarare  @Mugurell 

https://console.firebase.google.com/u/0/project/moz-fenix/testlab/histories/bh.66b7091e15d53d45/matrices/8268278885395658076/executions/bs.ef9cfb62a48b7d64/issues

```Kotlin
android.view.WindowManager$BadTokenException: Unable to add window -- token null is not valid; is your activity running?
     FATAL EXCEPTION: main
Process: org.mozilla.fenix.debug, PID: 12447
android.view.WindowManager$BadTokenException: Unable to add window -- token null is not valid; is your activity running?
	at android.view.ViewRootImpl.setView(ViewRootImpl.java:1068)
	at android.view.WindowManagerGlobal.addView(WindowManagerGlobal.java:409)
	at android.view.WindowManagerImpl.addView(WindowManagerImpl.java:109)
	at org.mozilla.fenix.compose.cfr.CFRPopupFullScreenLayout.show(CFRPopupFullscreenLayout.kt:121)
	at org.mozilla.fenix.compose.cfr.CFRPopup.show$lambda-1(CFRPopup.kt:76)
	at org.mozilla.fenix.compose.cfr.CFRPopup.$r8$lambda$TXBcbC1ZJIfV_BS4FJxo1uw7w5M(Unknown Source:0)
	at org.mozilla.fenix.compose.cfr.CFRPopup$$ExternalSyntheticLambda0.run(Unknown Source:2)
	at android.os.Handler.handleCallback(Handler.java:938)
	at android.os.Handler.dispatchMessage(Handler.java:99)
	at android.os.Looper.loop(Looper.java:223)
	at android.app.ActivityThread.main(ActivityThread.java:7656)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:592)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:947)
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21516)

AaronMT:
[...]

PR #26962, opened 2022-09-13
For #26884: [UI tests] Reset modified settings just before the Activity finishes

Mugurell:
This should prevent issues with the cleanup being done earlier (even in @After) and affecting the behavior of the Activity which is still running. With this issue fixed I've re-enabled verifyCustomizeHomepageTest.

Because when FeatureSettingsHelper is instantiated it caches the settings values from that instant it's important to only use one instance of it so I've updated all tests that are using HomeActivityTestRule. HomeActivityIntentTestRule may be updated with a similar functionality later.

### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation

Used by GitHub Actions.


Mugurell:
Not sure why but many tests from `BookmarkTest` were failing in `ui-test-x86-debug` (15 / 26).
Seems like the TCP CFR interferes and needs to also be blocked.
There is also a small bug there with the test [disabling forever the jump back in CFR](https://github.com/mozilla-mobile/fenix/blob/c7ebec360227c70ce2ba0243b88321eac39eee89/app/src/androidTest/java/org/mozilla/fenix/ui/BookmarksTest.kt#L65-L66).

AndiAJ (approved):
R+

PR #26972, opened 2022-09-14
Closes #26971: fix wallpaper onboarding display condition

### LINKED ISSUE #26971: bug: wallpaper onboarding is not shown if any of the thumbnails can not be loaded
Changed logic while [refactoring](https://github.com/mozilla-mobile/fenix/pull/26758#discussion_r964257663), want to check if there are enough available wallpapers to show the wallpaper onboarding.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21561)

mavduevskiy:
Closes #26971

Currently one of the thumbnails can not be loaded, and it breaks showing the wallpaper onboarding condition.

### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation
Fixes #26971

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlController.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlController.kt
@@ -56,7 +56,7 @@ import org.mozilla.fenix.gleanplumb.MessageController
 import org.mozilla.fenix.home.HomeFragment
 import org.mozilla.fenix.home.HomeFragmentDirections
 import org.mozilla.fenix.home.Mode
-import org.mozilla.fenix.onboarding.WallpaperOnboardingDialogFragment
+import org.mozilla.fenix.onboarding.WallpaperOnboardingDialogFragment.Companion.THUMBNAILS_SELECTION_COUNT
 import org.mozilla.fenix.settings.SupportUtils
 import org.mozilla.fenix.settings.SupportUtils.SumoTopic.PRIVATE_BROWSING_MYTHS
 import org.mozilla.fenix.utils.Settings
@@ -510,16 +510,19 @@ class DefaultSessionControlController(
     }
 
     override fun handleShowWallpapersOnboardingDialog(state: WallpaperState): Boolean {
-        if (state.availableWallpapers.all { it.thumbnailFileState == Wallpaper.ImageFileState.Downloaded } &&
-            state.availableWallpapers.size >= WallpaperOnboardingDialogFragment.THUMBNAILS_COUNT
-        ) {
-            navController.nav(
-                R.id.homeFragment,
-                HomeFragmentDirections.actionGlobalWallpaperOnboardingDialog(),
-            )
-            return true
+        return state.availableWallpapers.filter { wallpaper ->
+            wallpaper.thumbnailFileState == Wallpaper.ImageFileState.Downloaded
+        }.size.let { downloadedCount ->
+            // We only display the dialog if enough thumbnails have been downloaded for it.
+            downloadedCount >= THUMBNAILS_SELECTION_COUNT
+        }.also { showOnboarding ->
+            if (showOnboarding) {
+                navController.nav(
+                    R.id.homeFragment,
+                    HomeFragmentDirections.actionGlobalWallpaperOnboardingDialog(),
+                )
+            }
         }
-        return false
     }
 
     override fun handleReadPrivacyNoticeClicked() {
--- a/app/src/test/java/org/mozilla/fenix/home/DefaultSessionControlControllerTest.kt
+++ b/app/src/test/java/org/mozilla/fenix/home/DefaultSessionControlControllerTest.kt
@@ -71,8 +71,11 @@ import org.mozilla.fenix.helpers.FenixRobolectricTestRunner
 import org.mozilla.fenix.home.recentbookmarks.RecentBookmark
 import org.mozilla.fenix.home.recenttabs.RecentTab
 import org.mozilla.fenix.home.sessioncontrol.DefaultSessionControlController
+import org.mozilla.fenix.onboarding.WallpaperOnboardingDialogFragment.Companion.THUMBNAILS_SELECTION_COUNT
 import org.mozilla.fenix.settings.SupportUtils
 import org.mozilla.fenix.utils.Settings
+import org.mozilla.fenix.wallpapers.Wallpaper
+import org.mozilla.fenix.wallpapers.WallpaperState
 import mozilla.components.feature.tab.collections.Tab as ComponentTab
 
 @RunWith(FenixRobolectricTestRunner::class) // For gleanTestRule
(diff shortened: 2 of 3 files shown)


MatthewTighe (approved):
LGTM, but I think this would benefit from having a test for it as well

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlController.kt
```diff
@@ -510,16 +510,18 @@ class DefaultSessionControlController(
...
-                R.id.homeFragment,
-                HomeFragmentDirections.actionGlobalWallpaperOnboardingDialog(),
-            )
-            return true
+        return state.availableWallpapers.filter { wallpaper ->
+            wallpaper.thumbnailFileState == Wallpaper.ImageFileState.Downloaded
+        }.size.let { downloadedCount ->
+            downloadedCount >= WallpaperOnboardingDialogFragment.THUMBNAILS_COUNT
+        }.also { isThumbnailsDownloaded ->
+            if (isThumbnailsDownloaded) {
```
gabrielluong, line 518:
```suggestion
            if (downloadedCount >= WallpaperOnboardingDialogFragment.THUMBNAILS_COUNT) {
```
I don't feel too strongly about this, but I think we can simplify this and shave down a couple lines.

gabrielluong: I think we could also use a comment for `THUMBNAILS_COUNT` to indicate that this is minimum threshold to indicate that enough thumbnails have been downloaded.

mcarare: nit: maybe also rename it to make it obvious it is a minimum threshold?

MatthewTighe (commented):
Small nit, but I am otherwise happy with this assuming the other comments are addressed

app/src/test/java/org/mozilla/fenix/home/DefaultSessionControlControllerTest.kt
```diff
@@ -878,6 +881,61 @@ class DefaultSessionControlControllerTest {
         assertNull(TopSites.remove.testGetValue()!!.single().extra)
     }
 
+    @Test
+    fun `GIVEN exactly the required amount of downloaded thumbnails WHEN has no errors THEN handleShowWallpapersOnboardingDialog calls the dialog`() {
```
MatthewTighe, line 885: nit: I think these tests might convey their purpose a bit better if they read like:

`GIVEN exactly the required amount of downloaded thumbnails with no errors WHEN handling wallpaper dialog THEN dialog is shown`

and 

`GIVEN fewer than the required amount of downloaded thumbnails WHEN handling wallpaper dialog THEN the dialog is not shown`

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlController.kt
mavduevskiy: changed the param to `THUMBNAILS_SELECTION_COUNT`. It's not just a trashhold for downloads, it's the amount of items the selector is expected to have. It will take only 6 items if there are more, and we should not try to display the onboarding unless there are less then 6.

mavduevskiy:
@MatthewTighe addressed the naming!

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlController.kt
gabrielluong: I think this is where the comment for the parameter is more important than the proposed name change here. I don't believe the name change added any additional information or clarity for what it represents.

gabrielluong:
@mavduevskiy Could you make sure you are either squashing or amending your commits before landing? This goes back to @csadilek https://github.com/mozilla-mobile/fenix/pull/26758#issuecomment-1242971836 where we don't want commits that are simply addressing review feedback. Since we are an open source project, we expect to maintain a fairly professional commit log that will help our consumers (forks, etc).

In the future, I would recommend doing `git commit --amend --no-edit` when addressing review feedback for your commits. This will futureproof you for the workflow when we move to phabricator and mozilla-central.

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlController.kt
```diff
@@ -510,16 +510,19 @@ class DefaultSessionControlController(
...
-                HomeFragmentDirections.actionGlobalWallpaperOnboardingDialog(),
-            )
-            return true
+        return state.availableWallpapers.filter { wallpaper ->
+            wallpaper.thumbnailFileState == Wallpaper.ImageFileState.Downloaded
+        }.size.let { downloadedCount ->
+            // We only display the dialog if enough thumbnails have been downloaded for it.
+            downloadedCount >= WallpaperOnboardingDialogFragment.THUMBNAILS_SELECTION_COUNT
+        }.also { isThumbnailsDownloaded ->
+            if (isThumbnailsDownloaded) {
```
gabrielluong, line 519:
```suggestion
            if (downloadedCount >= THUMBNAILS_SELECTION_COUNT) {
```
nit: "Less is more" or so the saying goes. Since you added a comment describing the conditional, I don't think there's really a point in naming the conditional to `isThumbnailsDownloaded`. I would suggest shortening the amount of code that is necessary here.

Also for your imports, you could import `import org.mozilla.fenix.onboarding.WallpaperOnboardingDialogFragment.THUMBNAILS_SELECTION_COUNT` to be less verbose.

mavduevskiy: Hey! Thanks for the reminder. I was actually waiting for some feedback from @MatthewTighe or you before landing. And assuming there might be more changes, didn't squash into one, so the trackrecord would be there.

Regarding the shortening, I am not sure I follow you line of though. Are you suggesting something like this? We mess up the returning type this way, it becomes unit, not boolean

```
return state.availableWallpapers.filter { wallpaper ->
            wallpaper.thumbnailFileState == Wallpaper.ImageFileState.Downloaded
        }.size.let { downloadedCount ->
            // We only display the dialog if enough thumbnails have been downloaded for it.
            if (downloadedCount >= WallpaperOnboardingDialogFragment.THUMBNAILS_SELECTION_COUNT) {
                navController.nav(
                    R.id.homeFragment,
                    HomeFragmentDirections.actionGlobalWallpaperOnboardingDialog(),
                )
            }
        }
```

gabrielluong: Yes I was suggesting what you have provided. I do see the boolean now so, I understand that we couldn't shorten this. Apologizes.

mavduevskiy: Oh, okay, gotcha. Changed to `showOnboarding`, as well addressing the import as well.

mavduevskiy:
@MatthewTighe @gabrielluong 
Please let me know if you have any further suggestions before we could land it

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlController.kt
gabrielluong: 👍 Looks good. Thanks!

gabrielluong:
> @MatthewTighe @gabrielluong Please let me know if you have any further suggestions before we could land it

Nothing further from me. I think you're good to go.

PR #26996, opened 2022-09-15
Closes #26995: add sorting of wallpapers on wallpapers onboarding tool

### LINKED ISSUE #26995: wallpapers onboarding filtering
Follow up for [issues/26750](https://github.com/mozilla-mobile/fenix/issues/26750)

Currently only classic wallpapers are shown. We want to have 3 seasonal first, if available, the rest – classic.

[design](https://www.figma.com/file/d7l1iGabQYqnLcnQbHDGzM/Wallpaper-2.0-%2B-Onboarding-v106?node-id=7946%3A223536)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21573)

mavduevskiy:
Filtering of wallpapers for wallpaper onboarding tool.

We are given a list of wallpapers, that consists of classic and seasonal collections. They come from the server, so theoretically the numbers inside each group could be any. Wrote an extension function that has a goal to provide a sorted list, which under ideal conditions would have 3 seasonal wallpapers first and 2 classic after. 

Edge cases I could come up with are: when it's all classic, or all seasonal, or nothing at all, or 1 classic or 1 seasonal, or 5 seasonal and 1 classic, or 1 seasonal and 5 classic.

image

### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation
Fixes #26995

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/onboarding/WallpaperOnboardingDialogFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/onboarding/WallpaperOnboardingDialogFragment.kt
@@ -26,6 +26,7 @@ import org.mozilla.fenix.R
 import org.mozilla.fenix.ext.components
 import org.mozilla.fenix.ext.requireComponents
 import org.mozilla.fenix.ext.settings
+import org.mozilla.fenix.settings.wallpaper.getWallpapersForOnboarding
 import org.mozilla.fenix.theme.FirefoxTheme
 import org.mozilla.fenix.wallpapers.Wallpaper
 import org.mozilla.fenix.wallpapers.WallpaperOnboarding
@@ -82,7 +83,7 @@ class WallpaperOnboardingDialogFragment : BottomSheetDialogFragment() {
         setContent {
             FirefoxTheme {
                 val wallpapers = appStore.observeAsComposableState { state ->
-                    state.wallpaperState.availableWallpapers.take(THUMBNAILS_SELECTION_COUNT)
+                    state.wallpaperState.availableWallpapers.getWallpapersForOnboarding()
                 }.value ?: listOf()
                 val currentWallpaper = appStore.observeAsComposableState { state ->
                     state.wallpaperState.currentWallpaper
@@ -127,6 +128,13 @@ class WallpaperOnboardingDialogFragment : BottomSheetDialogFragment() {
     }
 
     companion object {
+        // The number of wallpaper thumbnails to display.
         const val THUMBNAILS_SELECTION_COUNT = 6
+
+        // The desired amount of seasonal wallpapers inside of the selector.
+        const val SEASONAL_WALLPAPERS_COUNT = 3
+
+        // The desired amount of seasonal wallpapers inside of the selector.
+        const val CLASSIC_WALLPAPERS_COUNT = 2
     }
 }
--- a/app/src/main/java/org/mozilla/fenix/settings/wallpaper/Extensions.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/wallpaper/Extensions.kt
@@ -4,24 +4,65 @@
 
 package org.mozilla.fenix.settings.wallpaper
 
+import org.mozilla.fenix.onboarding.WallpaperOnboardingDialogFragment.Companion.CLASSIC_WALLPAPERS_COUNT
+import org.mozilla.fenix.onboarding.WallpaperOnboardingDialogFragment.Companion.SEASONAL_WALLPAPERS_COUNT
+import org.mozilla.fenix.onboarding.WallpaperOnboardingDialogFragment.Companion.THUMBNAILS_SELECTION_COUNT
 import org.mozilla.fenix.wallpapers.Wallpaper
 
 /**
  * The extension function to group wallpapers according to their name.
  **/
-fun List<Wallpaper>.groupByDisplayableCollection(): Map<Wallpaper.Collection, List<Wallpaper>> = groupBy {
-    it.collection
-}.filter {
-    it.key.name != "default"
-}.map {
-    val wallpapers = it.value.filter { wallpaper ->
-        wallpaper.thumbnailFileState == Wallpaper.ImageFileState.Downloaded
+fun List<Wallpaper>.groupByDisplayableCollection(): Map<Wallpaper.Collection, List<Wallpaper>> =
+    groupBy {
+        it.collection
+    }.filter {
+        it.key.name != "default"
+    }.map {
+        val wallpapers = it.value.filter { wallpaper ->
+            wallpaper.thumbnailFileState == Wallpaper.ImageFileState.Downloaded
+        }
+        if (it.key.name == "classic-firefox") {
+            it.key to listOf(Wallpaper.Default) + wallpapers
+        } else {
+            it.key to wallpapers
+        }
+    }.toMap().takeIf {
+        it.isNotEmpty()
+    } ?: mapOf(Wallpaper.DefaultCollection to listOf(Wallpaper.Default))
+
+/**
+ * Returns a list of wallpapers to display in the wallpaper onboarding.
+ *
+ * The ideal scenario is to return a list of wallpaper in the following order: 1 default, 3 seasonal and
+ * 2 classic wallpapers, but in case where there are less than 3 seasonal wallpapers, the remaining
+ * wallpapers are filled by classic wallpapers. If we have less than 6 wallpapers, return all the available
+ * seasonal and classic wallpapers.
+ */
+fun List<Wallpaper>.getWallpapersForOnboarding(): List<Wallpaper> {
+    val result = mutableListOf(Wallpaper.Default)
+    val classicWallpapers = mutableListOf<Wallpaper>()
+    val seasonalWallpapers = mutableListOf<Wallpaper>()
+
+    for (wallpaper in this) {
+        if (wallpaper == Wallpaper.Default) continue
+
+        if (wallpaper.collection.name == "classic-firefox") {
+            classicWallpapers.add(wallpaper)
+        } else {
+            seasonalWallpapers.add(wallpaper)
+        }
     }
-    if (it.key.name == "classic-firefox") {
-        it.key to listOf(Wallpaper.Default) + wallpapers
+
+    if (seasonalWallpapers.size < SEASONAL_WALLPAPERS_COUNT) {
+        result.addAll(seasonalWallpapers)
+        result.addAll(classicWallpapers.take((THUMBNAILS_SELECTION_COUNT - 1) - seasonalWallpapers.size))
+    } else if (classicWallpapers.size < CLASSIC_WALLPAPERS_COUNT) {
+        result.addAll(seasonalWallpapers.take((THUMBNAILS_SELECTION_COUNT - 1) - classicWallpapers.size))
+        result.addAll(classicWallpapers)
     } else {
-        it.key to wallpapers
+        result.addAll(seasonalWallpapers.take(SEASONAL_WALLPAPERS_COUNT))
+        result.addAll(classicWallpapers.take(CLASSIC_WALLPAPERS_COUNT))
     }
-}.toMap().takeIf {
-    it.isNotEmpty()
-} ?: mapOf(Wallpaper.DefaultCollection to listOf(Wallpaper.Default))
+
+    return result
+}
--- a/app/src/test/java/org/mozilla/fenix/settings/wallpaper/ExtensionsTest.kt
+++ b/app/src/test/java/org/mozilla/fenix/settings/wallpaper/ExtensionsTest.kt
@@ -5,6 +5,7 @@
 package org.mozilla.fenix.settings.wallpaper
 
 import org.junit.Assert.assertEquals
+import org.junit.Assert.assertTrue
 import org.junit.Test
 import org.mozilla.fenix.wallpapers.Wallpaper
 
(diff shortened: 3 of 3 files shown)


app/src/main/java/org/mozilla/fenix/onboarding/WallpaperOnboardingDialogFragment.kt
```diff
@@ -128,5 +129,8 @@ class WallpaperOnboardingDialogFragment : BottomSheetDialogFragment() {
 
     companion object {
         const val THUMBNAILS_COUNT = 6
+
+        // The desirable amount of seasonal wallpapers inside of the selector.
```
gabrielluong, line 133:
```suggestion
        // The desired amount of seasonal wallpapers inside of the wallpaper selector.
```

app/src/main/java/org/mozilla/fenix/settings/wallpaper/Extensions.kt
```diff
@@ -25,3 +27,58 @@ fun List<Wallpaper>.groupByDisplayableCollection(): Map<Wallpaper.Collection, Li
 }.toMap().takeIf {
     it.isNotEmpty()
 } ?: mapOf(Wallpaper.DefaultCollection to listOf(Wallpaper.Default))
+
+/**
+ * The ideal scenario is to have 3 seasonal wallpapers and 2 classic, minding the order.
+ * But in case there are less than 3 or no seasonal wallpapers at all, they are replaced by classic.
+ * Edge cases where there are less then 6 items overall would return less items in the right order.
+ */
+fun List<Wallpaper>.groupForWallpaperOnboarding(): List<Wallpaper> {
```
gabrielluong, line 36: Rather than looping the wallpaper several times, I am wondering what if we first called `groupByDisplayableCollection` which will provide us a mapping of the collection and the list of wallpapers associated to their respective collection. You can then determine if the count of each group matches the desired number of seasonal wallpaper or not, and then take the appropriate number of wallpapers from each group to get the final list of wallpapers to display.

mavduevskiy: That was actually my first approach as well, but I didn't like the implementation. Sadly, I have removed it. 

What I didn't like about it: we can't access the classic wallpapers from the map directly, because the access is done through an object we didn't create. I had to look for it and store it in a separate parameter in a for loop. 

And there might be multiple seasonal groups, so we would have to iterate through them as well. It looked like having collections created more problems than brought benefits. In the context of this particular selector there are just seasonal and classic wallpapers, and we can differentiate them in a list as well.

And lastly, it looks like we can't escape having two loops, because we have two groups of not guaranteed length. One loop might satisfy a single condition (let's say, 3 seasonal) but in case we have just 1 classic for some reason, we would have go back to add another seasonal, that we haven't added yet.

I am open to improving the sorting though. This just looked like the optimal approach given the possibly any input from the backend.

gabrielluong: I don't really know how the wallpapers are structure. So, I am just gonna think out loud here and see if there is any possible approaches here that will make this code simpler. 
1. Is possible to go through the list of wallpaper and construct 2 lists one for seasonal and another for classic wallpapers?
2. After you have the 2 lists, you can check the count of each list and from there take the appropriate number of wallpapers from each.

mavduevskiy: okay, got it! that actually a much easier approach. updated

app/src/main/java/org/mozilla/fenix/settings/wallpaper/Extensions.kt
```diff
@@ -25,3 +28,37 @@ fun List<Wallpaper>.groupByDisplayableCollection(): Map<Wallpaper.Collection, Li
...
+    val result = mutableListOf(Wallpaper.Default)
+
+    val classicWallpapers = mutableListOf<Wallpaper>()
+    val seasonalWallpapers = mutableListOf<Wallpaper>()
+    for (wallpaper in this) {
```
gabrielluong, line 42:
```suggestion
    val result = mutableListOf(Wallpaper.Default)
    val classicWallpapers = mutableListOf<Wallpaper>()
    val seasonalWallpapers = mutableListOf<Wallpaper>()
    
    for (wallpaper in this) {
```
Generally, we want a new line before and after a `for` statement block to separate out and make it easier to read. Otherwise, all our code will just be squished together with no clear logical separation from a reader's perspective.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/Extensions.kt
```diff
@@ -25,3 +28,37 @@ fun List<Wallpaper>.groupByDisplayableCollection(): Map<Wallpaper.Collection, Li
...
+        if (wallpaper.collection.name == "classic-firefox") {
+            classicWallpapers.add(wallpaper)
+        } else {
+            seasonalWallpapers.add(wallpaper)
+        }
+    }
+
+    if (seasonalWallpapers.size < SEASONAL_WALLPAPERS_COUNT) {
+        result.addAll(seasonalWallpapers)
+        result.addAll(classicWallpapers.take((maxSize - 1) - seasonalWallpapers.size))
```
gabrielluong, line 54:
```suggestion
        result.addAll(classicWallpapers.take((maxSize - 1) - seasonalWallpapers.size))
```
My initial reaction was to wonder where `maxSize` was coming from. I think this code would be clearer if we didn't alias `THUMBNAILS_SELECTION_COUNT` as `maxSize`, but I imagine you did this because the name will take up a lot of space, but I would argue that we can just move `.take()` onto the next line.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/Extensions.kt
gabrielluong: Thanks, this does make much more easier to understand and maintain.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/Extensions.kt
```diff
@@ -25,3 +28,37 @@ fun List<Wallpaper>.groupByDisplayableCollection(): Map<Wallpaper.Collection, Li
 }.toMap().takeIf {
     it.isNotEmpty()
 } ?: mapOf(Wallpaper.DefaultCollection to listOf(Wallpaper.Default))
+
+/**
+ * The ideal scenario is to have 3 seasonal wallpapers and 2 classic, minding the order.
+ * But in case there are less than 3 or no seasonal wallpapers at all, they are replaced by classic.
+ * Edge cases where there are less then 6 items overall would return less items in the right order.
+ */
```
gabrielluong, line 36:
```suggestion
/**
 * Returns a list of wallpapers to display in the wallpaper onboarding.
 *
 * The ideal scenario is to return a list of wallpaper in the following order: 1 default, 3 seasonal and 
 * 2 classic wallpapers, but in case where there are less than 3 seasonal wallpapers, the remaining
 * wallpapers are filled by classic wallpapers. If we have less than 6 wallpapers, return all the available 
 * seasonal and classic wallpapers.
 */
```
We do want a line that describes what the function does.  Otherwise, I helped reword your comment to make it flow more naturally from an english speaking perspective.

app/src/main/java/org/mozilla/fenix/onboarding/WallpaperOnboardingDialogFragment.kt
```diff
@@ -128,5 +129,11 @@ class WallpaperOnboardingDialogFragment : BottomSheetDialogFragment() {
 
     companion object {
         const val THUMBNAILS_SELECTION_COUNT = 6
```
gabrielluong, line 131:
```suggestion
        // The number of wallpaper thumbnails to display.
        const val THUMBNAILS_SELECTION_COUNT = 6
```

Let's add a comment for this while we are at it.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/Extensions.kt
```diff
@@ -25,3 +28,37 @@ fun List<Wallpaper>.groupByDisplayableCollection(): Map<Wallpaper.Collection, Li
...
+fun List<Wallpaper>.groupForWallpaperOnboarding(): List<Wallpaper> {
```
gabrielluong, line 37:
```suggestion
fun List<Wallpaper>.getWallpapersForOnboarding(): List<Wallpaper> {
```
We aren't grouping anything like a `Map` so I think we shouldn't use the same naming as above.

app/src/test/java/org/mozilla/fenix/settings/wallpaper/ExtensionsTest.kt
```diff
@@ -54,6 +54,102 @@ class ExtensionsTest {
         assertEquals(downloadedSeasonalWallpapers, result[seasonalCollection])
     }
 
+    @Test
+    fun `GIVEN two collections of enough size WHEN grouped for onboarding THEN result contains 3 seasonal and 2 classic`() {
```
gabrielluong, line 58:
```suggestion
    fun `GIVEN two collections of appropriate size WHEN fetched for onboarding THEN result contains 3 seasonal and 2 classic`() {
```
s/enough/appropriate/g
s/not enough/insufficient/g

app/src/test/java/org/mozilla/fenix/settings/wallpaper/ExtensionsTest.kt
```diff
@@ -54,6 +54,102 @@ class ExtensionsTest {
...
+
+        val result = allWallpapers.groupForWallpaperOnboarding()
+
+        assertEquals(3, result.count { it.collection.name == "finally fall" })
+        assertEquals(2, result.count { it.collection.name == classicCollection.name })
+        assert(result.contains(Wallpaper.Default))
+    }
+
+    @Test
+    fun `GIVEN five collections of not enough size WHEN grouped for onboarding THEN result contains 3 seasonal and 2 classic`() {
```
gabrielluong, line 72:
```suggestion
    fun `GIVEN five collections of insufficient size WHEN fetched for onboarding THEN result contains 3 seasonal and 2 classic`() {
```

app/src/test/java/org/mozilla/fenix/settings/wallpaper/ExtensionsTest.kt
```diff
@@ -54,6 +54,102 @@ class ExtensionsTest {
...
+        val seasonalCollection = getSeasonalCollection("finally fall")
+        val seasonalWallpapers = (0..5).map { generateSeasonalWallpaperCollection("${seasonalCollection.name}$it", seasonalCollection.name) }
+        val classicFirefoxWallpapers = (0..5).map { generateClassicFirefoxWallpaper("firefox$it") }
+        val allWallpapers = listOf(Wallpaper.Default) + classicFirefoxWallpapers + seasonalWallpapers
+
+        val result = allWallpapers.groupForWallpaperOnboarding()
+
+        assertEquals(3, result.count { it.collection.name == "finally fall" })
+        assertEquals(2, result.count { it.collection.name == classicCollection.name })
+        assert(result.contains(Wallpaper.Default))
```
gabrielluong, line 68: Instead of using `assert`, we generally do `assertTrue` or `assertFalse` to be more explicit.

gabrielluong (commented):
I am gonna defer to @MatthewTighe for the final review, but I think it looks good.

app/src/test/java/org/mozilla/fenix/settings/wallpaper/ExtensionsTest.kt
MatthewTighe: IIRC `assert` won't actually trigger test failures in some cases either, as it's a keyword rather than part of the JUnit framework

app/src/main/java/org/mozilla/fenix/settings/wallpaper/Extensions.kt
```diff
@@ -4,24 +4,65 @@
...
+        result.addAll(classicWallpapers)
     } else {
-        it.key to wallpapers
+        result.addAll(seasonalWallpapers.take(SEASONAL_WALLPAPERS_COUNT))
+        result.addAll(classicWallpapers.take(CLASSIC_WALLPAPERS_COUNT))
     }
-}.toMap().takeIf {
-    it.isNotEmpty()
-} ?: mapOf(Wallpaper.DefaultCollection to listOf(Wallpaper.Default))
+
```
MatthewTighe, line 66: This could also potentially be simplified by doing something like the following:

```
val defaultAndSeasonal = listOf(Wallpaper.Default) + seasonalWallpapers.take(SEASONAL_WALLPAPERS_COUNT)

return defaultAndSeasonal + classicWallpapers.take(THUMBNAILS_SELECTION_COUNT - defaultAndSeasonal.size)
```

mavduevskiy:
@gabrielluong 
addressed your comments, thanks!
no worries, I think I sometimes forget to squash before commit during the recent merging rush. I like to keep separate commits when I work on a branch to see the progress, if I squash now, I would erase some of the history about the work progress. I will do that before commiting, you have my word :)

app/src/main/java/org/mozilla/fenix/settings/wallpaper/Extensions.kt
```diff
@@ -4,24 +4,65 @@
...
+            wallpaper.thumbnailFileState == Wallpaper.ImageFileState.Downloaded
+        }
+        if (it.key.name == "classic-firefox") {
+            it.key to listOf(Wallpaper.Default) + wallpapers
+        } else {
+            it.key to wallpapers
+        }
+    }.toMap().takeIf {
+        it.isNotEmpty()
+    } ?: mapOf(Wallpaper.DefaultCollection to listOf(Wallpaper.Default))
```
MatthewTighe, line 31: This looks like just a formatting change. I think in general we want to avoid doing this, but I am fine keeping it here for simplicity and so we don't hit it again in the future

app/src/main/java/org/mozilla/fenix/settings/wallpaper/Extensions.kt
mavduevskiy: that fails the `GIVEN no items in classic collection and some seasonal WHEN grouped for onboarding THEN result contains all seasonal` test! We might face a case where we have to take more seasonal, then just 3, when there is not enough classic. it's an edge case, but given the api, I would like to write something that considers all the possible cases.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/Extensions.kt
mavduevskiy: ops, that was not intentional. but yes, I agree, makes sense to keep it

app/src/test/java/org/mozilla/fenix/settings/wallpaper/ExtensionsTest.kt
gabrielluong: @mavduevskiy Just a bump on this. This is mostly an English wording nit.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/Extensions.kt
MatthewTighe: Ah, good catch. Let's just keep it as is then since it does a good job addressing all these edge cases

app/src/test/java/org/mozilla/fenix/settings/wallpaper/ExtensionsTest.kt
mavduevskiy: I have updated to the suggested version, is there anything else I should do?

gabrielluong: `/g` meant global here so we'll want to do the same for the other tests.

mavduevskiy: oh, got it! what does `/s` stand for?

gabrielluong: Sorry, `s/enough/appropriate/g` was a sed notation which is a linux command for search and transforming strings. You can read this as s/<old>/<new> where <old> will be replaced by <new> `/g` will be global and similar to what you see with regular expressions.

https://linuxhint.com/50_sed_command_examples/#s1

mavduevskiy: thanks! I didn't understand your comment properly, I did get only the suggested part.
updated!

MatthewTighe (approved):
Tested locally, LGTM!

MatthewTighe:
Let's land this before https://github.com/mozilla-mobile/fenix/pull/27018 to make rebasing earlier. Looks like the commits still need to be squashed but otherwise I think this is good to go

mavduevskiy:
@gabrielluong @MatthewTighe 
FYI squashed before landing!

app/src/test/java/org/mozilla/fenix/settings/wallpaper/ExtensionsTest.kt
gabrielluong: @mavduevskiy Sorry looks like the `<old>` and `<new>` were hidden in the markdown. I updated the comment in case it will be helpful.

PR #27078, opened 2022-09-20
Closes #27077: Load first run experiments synchronously.

### LINKED ISSUE #27077: Integrate for loading/applying first run experiments synchronously
Covers the integration changes needed for https://mozilla-hub.atlassian.net/browse/EXP-2697

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21618)

Amejia481:
Depends on https://github.com/mozilla-mobile/android-components/pull/12794 and  https://github.com/mozilla/application-services/pull/5131

### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation

Used by GitHub Actions.


app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -429,7 +429,6 @@ open class FenixApplication : LocaleAwareApplication(), Provider {
             // experiment on features close to startup.
             // But we need viaduct (the RustHttp client) to be ready before we do.
             components.analytics.experiments.apply {
-                initialize()
```
Amejia481, line 432: @jhugman is this change still needed? I don't remember we removed it because debugging.

app/src/main/java/org/mozilla/fenix/experiments/NimbusSetup.kt
```diff
@@ -88,32 +95,26 @@ fun createNimbus(context: Context, url: String?): NimbusApi {
...
-            // `applyPendingExperiments()`) then this has you covered.
-            // This call does its work on the db thread.
-            initialize()
+            val isFirstNimbusRun = if (context.settings().isFirstNimbusRun || url.isNullOrBlank()) {
+                setExperimentsLocally(R.raw.initial_experiments)
+                context.settings().isFirstNimbusRun = false
+                true
+            } else {
+                false
+            }
```
jeddai, line 104: I don't think we need this if-block. The `initialize` method below will set the experiments locally if it is the first nimbus run.

And `context.settings().isFirstNimbusRun = false` could be moved to after the `runBlocking` block below.

Amejia481: but the problem will be that we will `context.settings().isFirstNimbusRun = false` every time, even if it was previously set

jeddai: We can still have the `isFirstNimbusRun` variable if needed for that

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
jonalmeida: No, we do not need this any more because we not have the new `initialize` method that does all the things inside it.

Amejia481: Thanks for the clarification 👍🏽

app/src/main/java/org/mozilla/fenix/experiments/NimbusSetup.kt
Amejia481: Addressed :)

app/src/main/java/org/mozilla/fenix/experiments/NimbusSetup.kt
```diff
@@ -88,32 +95,24 @@ fun createNimbus(context: Context, url: String?): NimbusApi {
...
-            applyPendingExperiments()
-
-            // Now fetch the experiments from the server. These will be available for feature
-            // configuration on the next run of the app. This call launches on the fetch thread.
-            fetchExperiments()
+            if (!enabled) {
+                // This opts out of nimbus experiments. It involves writing to disk, so does its
+                // work on the db thread.
+                globalUserParticipation = enabled
+            }
```
jhugman, line 115: I think we can delete this now (it was there to migrate from mako).

```suggestion
```

app/src/main/java/org/mozilla/fenix/experiments/NimbusSetup.kt
```diff
@@ -88,32 +95,24 @@ fun createNimbus(context: Context, url: String?): NimbusApi {
...
 
-            if (!enabled) {
-                // This opts out of nimbus experiments. It involves writing to disk, so does its
-                // work on the db thread.
-                globalUserParticipation = enabled
+            runBlocking {
+                initialize(
+                    isFirstNimbusRun,
+                    R.raw.initial_experiments,
+                ).joinOrTimeout(TIME_0UT_LOADING_EXPERIMENT_FROM_DISK_MS)
```
jhugman, line 104: Second half:

```suggestion
                initialize(
                    isFirstNimbusRun || url.isNullOrBlank(),
                    R.raw.initial_experiments,
                ).joinOrTimeout(TIME_0UT_LOADING_EXPERIMENT_FROM_DISK_MS)
```

app/src/main/java/org/mozilla/fenix/experiments/NimbusSetup.kt
```diff
@@ -88,32 +95,24 @@ fun createNimbus(context: Context, url: String?): NimbusApi {
             // generated code.
             register(observer)
 
-            // This performs the minimal amount of work required to load branch and enrolment data
-            // into memory. If `getExperimentBranch` is called from another thread between here
-            // and the next nimbus disk write (setting `globalUserParticipation` or
-            // `applyPendingExperiments()`) then this has you covered.
-            // This call does its work on the db thread.
-            initialize()
+            val isFirstNimbusRun = context.settings().isFirstNimbusRun || url.isNullOrBlank()
```
jhugman, line 98: I'm not sure if this make sense to have the url check as part of the first run.

```suggestion
            val isFirstNimbusRun = context.settings().isFirstNimbusRun
```

Semantically, I think it's ok, but it may not always. I've added the url check below.

app/src/main/java/org/mozilla/fenix/experiments/NimbusSetup.kt
```diff
@@ -34,6 +36,11 @@ private val observer = object : NimbusInterface.Observer {
     }
 }
 
+/**
+ * The maximum amount of time the app launch will be blocked to load experiments from disk.
+ */
+private const val TIME_0UT_LOADING_EXPERIMENT_FROM_DISK_MS = 200L
```
jhugman, line 42: Nit: `0` and `O`.

```suggestion
private const val TIME_OUT_LOADING_EXPERIMENT_FROM_DISK_MS = 200L
```

jhugman (approved):
There are some straightforward changes needed to this, but otherwise this is looking really nice, and shouldn't hold up for another review cycle.

app/src/main/java/org/mozilla/fenix/experiments/NimbusSetup.kt
Amejia481: Do we need to pass `enabled` to nimbus to know when users opt-out of experiments on the Fenix ui?

Amejia481: Maybe we should address this in separate ticket, as I'm not sure if this is something we would like to uplift to v106.

app/src/main/java/org/mozilla/fenix/experiments/NimbusSetup.kt
```diff
@@ -88,32 +98,32 @@ fun createNimbus(context: Context, url: String?): NimbusApi {
...
+            // `joinOrTimeout` to limit the blocking until TIME_OUT_LOADING_EXPERIMENT_FROM_DISK_MS.
+            runBlockingIncrement {
+                val job = initialize(
+                    isFirstNimbusRun || url.isNullOrBlank(),
+                    R.raw.initial_experiments,
+                )
+                // We only read from disk when loading first-run experiments. This is the only time
+                // that we should join and block. Otherwise, we don't want to wait.
+                if (isFirstNimbusRun) {
+                    context.settings().isFirstNimbusRun = false
```
jonalmeida, line 117: We moved the pref update here in case there is some way we put the user into a blocked state and we get an ANR. Their second run will then be safe to continue in a non-blocking way.

cc: @jeddai @jhugman

jonalmeida (approved):
Arturo and I did a sanity test of this change locally with an `initial_experiments.json` that includes the MR features.

We tried to trigger the flow with the timeout hitting (by setting the timeout to 1) and the app did not get into an ANR state. On our particular test device, the experiment was loaded fast enough to be used in that situation for onboarding features.

app/src/main/java/org/mozilla/fenix/experiments/NimbusSetup.kt
jonalmeida: Filed https://github.com/mozilla-mobile/fenix/issues/27110.

Amejia481: ty!

Amejia481:
@mergify backport releases_v106.0.0

PR #27210, opened 2022-09-29
Closes #27209: fix WallpapersSettings crash

### LINKED ISSUE #27209: [Bug] IllegalStateException in WallpapersSettings
Addressing the sentry [crash](https://sentry.io/organizations/mozilla/issues/3591939251/?project=6295546&query=is%3Aunresolved+level%3Afatal+firstSeen%3A-2w&statsPeriod=14d)

How to reproduce: trigger the snackbar, leave the screen, press action button.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21687)

mavduevskiy:
Addressing the sentry [crash](https://sentry.io/organizations/mozilla/issues/3591939251/?project=6295546&query=is%3Aunresolved+level%3Afatal+firstSeen%3A-2w&statsPeriod=14d)

After playing around with the fragment, bumped into scenario: trigger the `snackbar`, leave the screen, press action.

Apparently, `snackbars` usually exist alongside with coordinator layout, that would disable them automatically. Switching to Compose stripped away the benifit, hid the `snackbar` manually upon hearing `onStop()`.

I also tested the `snackbar` behavior inside the `WallpaperOnboardingDialogFragment`, but dialog fragment disable `snackbars` automatically when dismissed.

### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation
Fixes #27209


app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettingsFragment.kt
```diff
@@ -139,4 +145,9 @@ class WallpaperSettingsFragment : Fragment() {
         super.onResume()
         showToolbar(getString(R.string.customize_wallpapers))
     }
+
+    override fun onStop() {
+        super.onStop()
+        snackBar?.let { it.dismiss() }
```
MozillaNoah, line 151:
```suggestion
        snackBar?.dismiss()
```

MozillaNoah (approved):

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettingsFragment.kt
```diff
@@ -37,6 +37,8 @@ class WallpaperSettingsFragment : Fragment() {
         requireComponents.useCases.wallpaperUseCases
     }
 
+    private var snackBar: FenixSnackbar? = null
```
jonalmeida, line 40: This looks wrong to me.

> Apparently, `snackbars` usually exist alongside with coordinator layout, that would disable them automatically. Switching to Compose stripped away the benifit, hid the `snackbar` manually upon hearing `onStop()`.

If what you say is true, why not use the fragment view instead of using the `ComposeView` in the [`onWallpaperSelected`](https://github.com/mozilla-mobile/fenix/blob/189b22fb69922d7e52dc89bb82c791e795c2fef5/app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettingsFragment.kt#L98)? This view reference is used only for the snackbar.

jonalmeida (changes requested):

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettingsFragment.kt
mavduevskiy: I followed the approach but had to adjust it. Fragment view in our case was a `ComposeView` (because `onCreateView()` was returning it). Changing in to the Coordinator did the trick!

Thanks for bringing it up!

mavduevskiy: now that I think of it, I should add a comment on why the coordinator layout was used.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettingsFragment.kt
```diff
@@ -128,8 +132,7 @@ class WallpaperSettingsFragment : Fragment() {
                             val retryResult = wallpaperUseCases.selectWallpaper(wallpaper)
                             onWallpaperSelected(wallpaper, retryResult, view)
                         }
-                    }
-                    .show()
```
gabrielluong, line 132: Let's avoid this change since we didn't have to touch this code. I am curious if this was automatically reformatted via your Android Studio since we recently ran a mass ktlint formatting so this seems a bit unexpected to me.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettingsFragment.kt
```diff
@@ -106,8 +111,7 @@ class WallpaperSettingsFragment : Fragment() {
                     .setText(view.context.getString(R.string.wallpaper_updated_snackbar_message))
                     .setAction(requireContext().getString(R.string.wallpaper_updated_snackbar_action)) {
                         findNavController().navigate(R.id.homeFragment)
-                    }
-                    .show()
+                    }.show()
```
gabrielluong, line 114: Same as my other comment.

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettingsFragment.kt
mavduevskiy: it was semi-intentional, reverted

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettingsFragment.kt
mavduevskiy: reverted

app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettingsFragment.kt
MatthewTighe: We could also shift the snackbar creation back to compose. For example, we could do something like this:

```
fun WallpaperSettings(
    wallpaperGroups: Map<Wallpaper.Collection, List<Wallpaper>>,
    defaultWallpaper: Wallpaper,
    loadWallpaperResource: suspend (Wallpaper) -> Bitmap?,
    selectedWallpaper: Wallpaper,
    onLearnMoreClick: (String, String) -> Unit,
) {
    ...
                WallpaperThumbnails(
                    wallpapers = wallpapers,
                    defaultWallpaper = defaultWallpaper,
                    loadWallpaperResource = loadWallpaperResource,
                    selectedWallpaper = selectedWallpaper,
                    onSelectWallpaper = {
                        val result = onSelectWallpaper()
                        showWallpaperSnackbar(result)
                    },
                )

                Spacer(modifier = Modifier.height(32.dp))
            }
        }
    }
}
```

this would keep the snackbar attached to the compose tree and hopefully avoid the problem. It would also help to contain the settings view logic to a single file.

mavduevskiy: Good point. But it would bring back the queuing bug we used to have (also the ui and animation is a bit inconsistent with the FenixSnackbar, but it's not a blocker for me). I would actually consider shifting back to compose, but may be not in the scope of this fix.

here is the differences in the UI between compose we had and FenixSnackbar (and the queuing issue)

[compose.webm](https://user-images.githubusercontent.com/92760693/193153074-2d0b4cf3-c4df-459c-ab02-1a7630f07ce0.webm)
[not-compose-2.webm](https://user-images.githubusercontent.com/92760693/193153040-aab17275-7690-406a-ae1d-127c5a4cd4b7.webm)

MatthewTighe: Okay, in favor of getting a small patch through that fixes the crasher I am cool with landing this. Would you mind filing an issue to track shifting this snackbar back to Compose? It could potentially be a follow-up to https://github.com/mozilla-mobile/fenix/issues/23221.

It turns out the queue bug is pretty simple to deal with. We just need to dismiss the previous one when we dispatch a new one, something like:

```
onSelectWallpaper = { updatedWallpaper ->
    val messageText = when(onSelectWallpaper(updatedWallpaper)) {
        Wallpaper.ImageFileState.Downloaded -> ...
        Wallpaper.ImageFileState.Error -> ...
    }
    coroutineScope.launch {
        scaffoldState.snackbarHostState.currentSnackbarData?.dismiss()
        scaffoldState.snackbarHostState.showSnackbar(
            message = messageText,
            duration = SnackbarDuration.Short,
        )
    }
},
```

MatthewTighe (approved):
Tested locally, looks good to me. Please see https://github.com/mozilla-mobile/fenix/pull/27210#discussion_r984124689 for filing follow ups and don't forget to squash before landing 🙂 

I think we could land this tomorrow (9/30) even though it's a Friday. I don't think this is liable to make the situation worse, and it'll give us some data over the weekend as to whether the issue has been addressed.

mavduevskiy:
@mergify backport releases_v106.0.0

PR #27272, opened 2022-10-05
Update some Gradle dependencies

rvandermeulen:
Similar to https://github.com/mozilla-mobile/android-components/pull/12888. Probably nothing super critical in here, but might as well.

### DIFF
--- a/buildSrc/src/main/java/AndroidComponents.kt
+++ b/buildSrc/src/main/java/AndroidComponents.kt
@@ -3,5 +3,5 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 object AndroidComponents {
-    const val VERSION = "107.0.20221005143122"
+    const val VERSION = "107.0.20221005223456"
 }
--- a/buildSrc/src/main/java/Dependencies.kt
+++ b/buildSrc/src/main/java/Dependencies.kt
@@ -14,11 +14,15 @@ object Versions {
     const val android_gradle_plugin = "7.3.0"
     const val android_lint_api = "30.3.0"
 
-    const val sentry = "5.7.3"
-    const val leakcanary = "2.8.1"
+    const val sentry = "6.4.2"
+    const val leakcanary = "2.9.1"
+
+    // We can't update this past 0.10.4 due to unit test failures:
+    // https://github.com/mozilla-mobile/fenix/issues/27287
     const val osslicenses_plugin = "0.10.4"
+
     const val detekt = "1.19.0"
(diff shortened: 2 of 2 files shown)


rvandermeulen:
For future reference, mockk 1.13.2 had a number of test-debug failures:
```
DefaultBrowserToolbarMenuControllerTest. GIVEN a shortcut page is open WHEN remove from shortcuts is pressed THEN show snackbar
SessionControlInteractorTest. GIVEN a PocketStoriesInteractor WHEN a story is clicked THEN handle it in a PocketStoriesController
SessionControlInteractorTest. GIVEN a PocketStoriesInteractor WHEN a story is shown THEN handle it in a PocketStoriesController
SearchDialogControllerTest. WHEN the search engine is added by the application THEN do not load URL
SearchDialogControllerTest. handleAddonsUrlCommitted
SearchDialogControllerTest. handleBlankUrlCommitted
SearchDialogControllerTest. handleCrashesUrlCommitted
SearchDialogControllerTest. handleMozillaUrlCommitted
SearchDialogControllerTest. handleSearchCommitted
SearchDialogControllerTest. handleSearchTermsTapped
SearchDialogControllerTest. handleUrlCommitted
AboutLibrariesFragmentTest. item click should open license dialog
DefaultInactiveTabsControllerTest. WHEN all inactive tabs are closed THEN perform the deletion and report the telemetry event and show a Snackbar
```

rvandermeulen:
I tried updating oss-licenses-plugin to version 0.10.5, but it breaks `AboutLibrariesFragmentTest` due to changes in how that gets populated in debug builds.
license

Which leads to: `java.lang.IllegalArgumentException: No item found containing text "org.mozilla.geckoview:geckoview"`

Which appears to be from upstream commit https://github.com/google/play-services-plugins/commit/138a15dc9209192c6a9397b341b0ef7ddc346af4

Not sure how we want to handle that eventually, but I guess we can take it to a different issue.

rvandermeulen:
I filed #27287 for the oss-licenses-plugin issue and added a comment to Dependencies.kt referencing it.

jonalmeida:
Gut feeling: These are a lot of misc dependency updates happening at once. I'm a bit worried about random intermittent and runtime crashes that we haven't explored yet.

rvandermeulen:
I'm happy to break this up into more than one patch so we can track regressions more easily. Any thoughts on how you want to break them up @jonalmeida?

rvandermeulen:
Per discussion with Jon on Slack, I'm going to break this into a set of smaller PRs.

Issue #27212, opened 2022-09-29
Update androidx test orchestrator to 1.4.1


AaronMT:
Support for Android API 31

https://developer.android.com/jetpack/androidx/releases/test#orchestrator-1.4.1
https://developer.android.com/jetpack/androidx/releases/test

Shared version `androidx_test_shared_version` is 1.4.0; might split to 1.4.1 or wait for future point release updates. 

https://github.com/mozilla-mobile/fenix/blob/017d45e80df996422a0af0b3cb7d9211824e1d03/buildSrc/src/main/java/Dependencies.kt#L247

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21688)

PR #27304, opened 2022-10-06
Update gradle testing dependencies

rvandermeulen:
Some of the Gradle dependencies we use for testing have updates available and update without issue. Let's update.

### DIFF
--- a/buildSrc/src/main/java/Dependencies.kt
+++ b/buildSrc/src/main/java/Dependencies.kt
@@ -15,7 +15,7 @@ object Versions {
     const val android_lint_api = "30.3.0"
 
     const val sentry = "6.4.2"
-    const val leakcanary = "2.8.1"
+    const val leakcanary = "2.9.1"
     const val osslicenses_plugin = "0.10.4"
     const val detekt = "1.19.0"
     const val jna = "5.8.0"
@@ -51,9 +51,9 @@ object Versions {
     const val junit = "5.5.2"
     const val mockk = "1.12.0"
 
-    const val mockwebserver = "4.9.0"
+    const val mockwebserver = "4.10.0"
     const val uiautomator = "2.2.0"
-    const val robolectric = "4.8.1"
+    const val robolectric = "4.9"
 
     const val google_ads_id_version = "16.0.0"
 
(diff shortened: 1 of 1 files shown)


AaronMT:
While here feel free to update orchestrator :) https://github.com/mozilla-mobile/fenix/issues/27212

rvandermeulen:
> There's also `androidx.test.orchestrator` if you'd like to update in this ticket #27212

Done!

buildSrc/src/main/java/Dependencies.kt
```diff
@@ -51,9 +51,9 @@ object Versions {
     const val junit = "5.5.2"
     const val mockk = "1.12.0"
 
-    const val mockwebserver = "4.9.0"
+    const val mockwebserver = "4.10.0"
```
jonalmeida, line 54: Self-note: In AC, the mockwebserver and okhttp versions both need to be keep in-sync: 

https://github.com/mozilla-mobile/android-components/blob/bc36830dd90bc455cd5ef81721e0dbfab38ff16d/buildSrc/src/main/java/Dependencies.kt#L18-L20

This doesn't matter as much with fenix/focus, because we do not use okhttp directly here and mockwebserver is used for testing Glean and UI tests.

buildSrc/src/main/java/Dependencies.kt
```diff
@@ -235,16 +235,18 @@ object Deps {
     // For the full IDs of these test dependencies, see:
     //   https://developer.android.com/training/testing/set-up-project#android-test-dependencies
     private const val androidx_test_shared_version = "1.4.0" // this appears to be shared with many deps.
+    private const val androidx_test_junit = "1.1.3"
+    private const val androidx_test_orchestrator = "1.4.1"
     const val androidx_test_core = "androidx.test:core:$androidx_test_shared_version"
     private const val androidx_espresso_version = "3.4.0"
```
jonalmeida, line 241: In the next group of bumps, let's group all these const and put them in one place like ones above. Not sure why they are dispersed around the place here.

rvandermeulen: Yes, I'm planning to do a big cleanup & re-org of this file once the current batch of dependency bump PRs lands. I don't want to create more rebase pain in the mean time.

buildSrc/src/main/java/Dependencies.kt
rvandermeulen: Fun story, updating to 4.x on AC is going to need a bit of work as it's not a drop-in replacement unfortunately ☹️.

buildSrc/src/main/java/Dependencies.kt
jonalmeida: Oh fair point. Yeah, this is not worth it then.

PR #27362, opened 2022-10-11
Update Kotlin to 1.7.20

rvandermeulen:
Just an incremental update to 1.7.10. Focus and r-b have both already taken this update without any known side effects.


jonalmeida (approved):
The looks pretty safe - Ryan landed these changes in Focus/RB first and they are only patch version updates. So lgtm.

PR #27363, opened 2022-10-11
Update Adjust to latest release

rvandermeulen:
Our copy of Adjust is 3 years out of date. Pull in the latest 4.32.0 release and the 2.2 release of installreferrer.

### DIFF
--- a/buildSrc/src/main/java/Dependencies.kt
+++ b/buildSrc/src/main/java/Dependencies.kt
@@ -45,8 +45,8 @@ object Versions {
 
     const val mozilla_android_components = AndroidComponents.VERSION
 
-    const val adjust = "4.18.3"
-    const val installreferrer = "1.0"
+    const val adjust = "4.32.0"
+    const val installreferrer = "2.2"
 
     const val junit = "5.5.2"
     const val mockk = "1.12.0"


mcarare (approved):

jonalmeida:
Ah this is a bit trickier - I think we need to do some proper testing. We only find out about bugs on these version bumps when this is in use in release and by then it's too late. I think there are some instructions somewhere on how to do this, so I'll look for those online and run a trial run.

mcarare:
> I see some suspects to consider. For example [this](https://github.com/adjust/android_sdk/releases/tag/v4.31.1):
> 
> > Added permission com.google.android.gms.permission.AD_ID in the SDK's mainfest.

We will also add this explicitely after we target SDK 33. See the [commit](https://github.com/mozilla-mobile/focus-android/pull/7841/commits/b252324b79720c811890407f283d7ec8ef04e533) in Focus (draft PR) for this. Adding this permission should not affect us until we target SDK 33. 
@rvandermeulen I know Focus already has this update. As far as you know, have there been any issues with uploading releases in the Play store?

rvandermeulen:
> @rvandermeulen I know Focus already has this update. As far as you know, have there been any issues with uploading releases in the Play store?

No issues.

jonalmeida (approved):
Went through the changelog for all the adjust versions and it looks good to me.

Issue #13845, opened 2020-08-14
[Bug] Swipe to switch tabs can be triggered in fullscreen mode


hwinnemoe:
## Steps to reproduce
1. Have the device in landscape orientation
2. Start playback of a video, e.g. https://demo.bitmovin.com/public/firefox/av1/
3. Enter fullscreen mode
4. Swipe from the left edge of the device

### Expected behavior
Nothing happens.

### Actual behavior
The swipe to switch tab gesture is triggered. This does not happen when entering fullscreen mode while still being in portrait orientation.

![swipe-tab-landscape](https://user-images.githubusercontent.com/42788925/90279224-185e7f80-de69-11ea-8192-4ef1441e955b.gif)

### Device information

* Android device: Sony Xperia Z2, Android 6.0.1
* Fenix version: Nightly 200814 06:02 (Build #2015757873)
AC: 54.0.20200812130143, 721a690d6
GV: 81.0a1-20200811091731
AS: 61.0.10

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-5043)

person808:
Hm probably a bug with our [hitbox detection](https://github.com/mozilla-mobile/fenix/blob/d9ffac0c890c0d58cd62f00693a45322c030f96d/app/src/main/java/org/mozilla/fenix/browser/ToolbarGestureHandler.kt#L307)... weird that its happening only in those scenarios.

hwinnemoe:
You could just block any swipe action if the device is in fullscreen mode.

I managed to also intermittently trigger the swipe gesture in portrait mode using the following STR:
1. Have the device in portrait orientation
2. Start playback of a video that does not force the device into landscape mode, e.g. the video on https://builders.mozilla.community/
3. Enter fullscreen mode
4. Start swiping from the left or right edge of the screen

![swipe-to-switch-portrait](https://user-images.githubusercontent.com/42788925/90315054-8e6def80-df18-11ea-9831-37f60bbbd07f.gif)

hwinnemoe:
@boek Correct me if I'm wrong but how do these commits relate to this issue?

Djfe:
This always happens to me if I want to move within a video in full screen mode (portrait). If I want to seek forwards/backwards (with the cursor), then fenix switches tabs. Very annoying. This gesture should only be applied if the bar is visible imo.

[...]

PR #27370, opened 2022-10-11
For #27346 - Disable swiping to switch tabs when fullscreen

### LINKED ISSUE #27346: [Bug]: Toolbar swiping activates even in fullscreen videos when toolbar is not shown
### Steps to reproduce

1. Go to **Settings -> Customize** and set **Scroll to hide toolbar** to **disabled** and **Swipe toolbar sideways to switch tabs** to **enabled** (the bug occurs only in this combination of settings)
2. Open **several tabs** (any websites)
3. Open websites with video, e.g. https://www.youtube.com (happens on other sites as well)
4. Rotate your phone to a **landscape** mode (this is an important step)
5. Play some video and switch it into **fullscreen**
6. See that the bottom toolbar is not shown during fullscreen video playback (which is correct).
7. Try to **swipe on the bottom edge** of the screen left and right, as if the toolbar was there. See that you actually **can swipe** there and switch to a different tab.
8. Back in the fullscreen video, try to use the progress bar slider to **move in the video forward and backward**. Depending on the how close is the slider to the bottom edge of the screen (depending on the website and perhaps your phone dimensions), sometimes it can get extremely tricky to move in the video instead of activating swiping on the invisible toolbar.
9. Repeat the whole process, but this time with **Scroll to hide toolbar** set to **enabled**. The problem doesn't appear. It only appears if the combination of both settings mentioned in step 1 is set as described.

### Expected behaviour

Toolbar swiping shouldn't work when the toolbar is not displayed (e.g. in a fullscreen video).

### Actual behaviour

Toolbar swiping works in a fullscreen video in the described situation, which makes seeking in video very hard.

### Device name

Samsung Galaxy A33

### Android version

Android 12

### Firefox release type

Firefox

### Firefox version

105.2.0

### Device logs

_No response_

### Additional information

See a screencast of the problem below. In the second half of the video, you can see how difficult it is for me to seek to a different part of the video. My touches are interpreted as tab swiping.

https://user-images.githubusercontent.com/755451/194754619-fafd13e2-b958-44d5-b276-0f3d9b89bbd1.mp4

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21767)

Mugurell:
https://user-images.githubusercontent.com/11428869/195129687-bc7ec7a9-4535-468d-b485-ba8cf84e5e87.mp4

### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation
Fixes #27346


gabrielluong:
@Mugurell Can you check in with Product and UX to inform them of the issue and the fix that you have proposed here? Just want make sure all of this is expected by all parties before proceeding.

fice-t:
@Mugurell I noticed that I am still able to activate this gesture with the default setup (bottom toolbar + dynamic toolbar) when the toolbar is hidden, which seems like a bug.

Would it make sense to instead check whether or not the toolbar is visible rather than disabling the gesture in fullscreen mode? That should take care of both issues if it's feasible.

Mugurell:
> @Mugurell I noticed that I am still able to activate this gesture (outside of fullscreen mode) with the default setup (bottom toolbar + dynamic toolbar) when the toolbar is hidden, which seems like a bug.

This shouldn't happen. Can you post in the ticket some steps to reproduce / a video?

fice-t:
I believe this should also fix #13845.

MozillaNoah (commented):
This looks good to me. I wasn't able to reproduce this locally with the proposed fix. Per the above comment from @gabrielluong, I'll formally approve this behavior if it has been verified by product/UX.

gabrielluong (approved):
We did raise this with @kbrosnan and got the okay.

Screen Shot 2022-11-16 at 8 11 44 PM

PR #27389, opened 2022-10-12
For #27388 - Add optional background colors to SelectableChip

### LINKED ISSUE #27388: Add optional background color parameters to SelectableChip
Needed for:
- https://github.com/mozilla-mobile/fenix/issues/26520

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21790)

MozillaNoah:
I opted for a nullable, optional parameter because of the use case with wallpapers. The wallpaper card colors are not assumed to be not-null, so rather than have the implementor of `SelectableChip` know about the default colors, I opted for a nullable optional parameter. The other solution would have been to somehow have a named parameter wrapped around an `if`, but I don't think that's possible with Kotlin.

### Pull Request checklist

### QA

- [] **QA Needed**

### GitHub Automation
Fixes #27388

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/compose/SelectableChip.kt
+++ b/app/src/main/java/org/mozilla/fenix/compose/SelectableChip.kt
@@ -18,6 +18,7 @@ import androidx.compose.material.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
+import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.text.TextStyle
 import androidx.compose.ui.text.capitalize
 import androidx.compose.ui.text.intl.Locale
@@ -32,12 +33,16 @@ import org.mozilla.fenix.theme.Theme
  *
  * @param text [String] displayed in this chip. Ideally should only be one word.
  * @param isSelected Whether this should be shown as selected.
+ * @param selectedBackgroundColor Optional background [Color] when the chip is selected.
+ * @param unselectedBackgroundColor Optional background [Color] when the chip is not selected.
  * @param onClick Callback for when the user taps this.
  */
 @Composable
 fun SelectableChip(
     text: String,
     isSelected: Boolean,
+    selectedBackgroundColor: Color? = null,
+    unselectedBackgroundColor: Color? = null,
     onClick: () -> Unit,
 ) {
     Box(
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/fenix/compose/SelectableChip.kt
```diff
@@ -32,12 +33,16 @@ import org.mozilla.fenix.theme.Theme
  *
  * @param text [String] displayed in this chip. Ideally should only be one word.
  * @param isSelected Whether this should be shown as selected.
+ * @param selectedBackgroundColor Optional background color when the chip is selected.
+ * @param unselectedBackgroundColor Optional background color when the chip is not selected.
```
gabrielluong, line 37:
```suggestion
 * @param selectedBackgroundColor Optional background [Color] when the chip is selected.
 * @param unselectedBackgroundColor Optional background [Color] when the chip is not selected.
```
Don't know what you think of this? Might be nice to just reference to the type when it's easy to do so.

app/src/main/java/org/mozilla/fenix/compose/SelectableChip.kt
```diff
@@ -82,17 +88,17 @@ private fun SelectableChipDarkThemePreview() {
 }
 
 @Composable
-@Preview(uiMode = UI_MODE_NIGHT_NO)
-private fun SelectableChipLightThemePreview() {
+@Preview()
```
gabrielluong, line 91: Should this just also have the 2 uiMode?

MozillaNoah: This one overrides the dark/light color tokens, so the second preview isn't necessary here

app/src/main/java/org/mozilla/fenix/compose/SelectableChip.kt
MozillaNoah: I think it's a good enhancement

app/src/main/java/org/mozilla/fenix/compose/SelectableChip.kt
gabrielluong: Okay, I am unfortunately imagining a PR where we go through every @Preview and introduce both uiMode and kinda just standardize our @Preview. It will of course not make sense for certain previews, but we should probably retroactively inspect the @Preview we do have and before we have the ability to see the preview both light and dark mode.

MozillaNoah: Gotcha, yeah, I could see us implementing like a `@LightDarkPreview()` annotation to accomplish this. I'm happy to add in the second `@Preview` here then

gabrielluong (approved):

app/src/main/java/org/mozilla/fenix/compose/SelectableChip.kt
MatthewTighe: +1 to the `LightDarkPreview` idea

PR #27403, opened 2022-10-14
Update Google Play Review dependency to 2.0.0

rvandermeulen:
Changelog:
https://developer.android.com/reference/com/google/android/play/core/release-notes

Additionally, the Google Play Core library was split into multiple separate libraries for each feature now, so change all references to the new Google Play Review component. AFAICT, nothing changed between 1.10.3 and 2.0.0 other than the split.
https://developer.android.com/reference/com/google/android/play/core/release-notes-in_app_reviews

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -590,7 +590,9 @@ dependencies {
 
     implementation Deps.google_ads_id // Required for the Google Advertising ID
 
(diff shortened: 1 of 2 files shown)


jonalmeida (approved):
This is a no-brainer version bump - thanks!

PR #27411, opened 2022-10-14
Update protobuf to 3.21.7

rvandermeulen:
A-S 94.3.2 is bringing in this bump as well. Would be good to have this ready to go once that update makes it into AC.

Changelog:
https://github.com/protocolbuffers/protobuf/releases/tag/v21.7


jonalmeida (approved):
Land when desired (with the AS upgrade).

Issue #21414, opened 2021-09-21
[Bug]: Add logins dialog: save button is enabled when password field is blank


bendk:
### Steps to reproduce

- Open the add logins dialog
- Input a site and username, but not a password
- Click the save button

### Expected behaviour

The save button should be disabled

### Actual behaviour

The save button is enabled.  Clicking it closes the dialog, but no login is added.

### Android version

Android 10

### Firefox release type

Firefox Nightly

### Firefox version

94.0a1

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-17953)

SoftVision-LorandJanos:
Hi @bendk! Thank you for your report.
I can confirm the issue is reproducible as described, with the following detail:
The checkmark is activated even after filling the Site field only (with valid data). So without any username or password.

Firefox versions tested:
- Nightly (94.0a1 - 2021-09-22).

Devices used:

- Samsung Galaxy Tab S3 (Android 9).
- Huawei P30 (Android 10).
- Xiaomi Mi11 Lite (Android 11).

SoftVision-LorandJanos:
This issue is still reproducible on the latest Nightly 101.0a1 (2022-04-14) build.
Device used: Google Pixel 4 (Android 12).
Reopening the ticket.

fice-t:
PR that fixes this: #27430.

SoftVision-LorandJanos:
Verified as fixed on the latest Nightly 109.0a1 (2022-11-17) build.
Device used: Oppo Find X5 (Android 12).
Closing the ticket as fixed.

AdinaPetridean:
Verified as fixed on the Firefox 109.0b1 (2022-12-15) build.
Device used: Poco M4 Pro (Android 12).

Issue #25980, opened 2022-07-11
Upgrade jetpack navigation & co to latest versions


mcarare:
At the time of opening this issue, the current version for navigation is 2.5.0. We aim to upgrade to 2.5.1, a version that should contain suppression of warnings in the generated safe args code. ( see https://issuetracker.google.com/issues/237725966 ).

Note: As navigation 2.5.0 depends on lifecycle 2.5.0, this issue will involve upgrading `androidx.lifecycle `and possibly other dependencies (core, fragment?)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-20975)

PR #27430, opened 2022-10-16
For #21414: Use MenuProvider and MenuHost interface methods

fice-t:
onPrepareOptionsMenu was not being called on invalidateOptionsMenu,
which led to the saveButton being left enabled even with invalid
hostnames, usersnames, and passwords.

### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation

Used by GitHub Actions.


fice-t:
Note: I improved and split off the 2nd commit to a new PR (#27816).

Mugurell (approved):
LGTM, thank you!

Mugurell:
This seems to be a leftover from https://github.com/mozilla-mobile/fenix/issues/25980.
Searching the project for `onPrepareOptionsMenu` / `onCreateOptionsMenu` seems like there is one more use of the now deprecated APIs that we need to fix for which I've opened a new ticket - https://github.com/mozilla-mobile/fenix/issues/27822.

Mugurell:
@fice-t If you would leave a small comment in https://github.com/mozilla-mobile/fenix/issues/21414 , maybe something around you wanted to work on the issue / already having a PR for it we will be able to assign the ticket to you. (Same for the others you've helped with).

fice-t:
>This seems to be a leftover from https://github.com/mozilla-mobile/fenix/issues/25980.
Searching the project for onPrepareOptionsMenu / onCreateOptionsMenu seems like there is one more use of the now deprecated APIs that we need to fix

There however does appear to be more instances of [`invalidateOptionsMenu`](https://searchfox.org/mozilla-mobile/search?q=OptionsMenu&path=&case=true&regexp=false) which may not be correct. CC @mcarare

fice-t:
>@fice-t If you would leave a small comment in https://github.com/mozilla-mobile/fenix/issues/21414 , maybe something around you wanted to work on the issue / already having a PR for it we will be able to assign the ticket to you. (Same for the others you've helped with).

Sure, though does it offer much more than Github showing the linked PR in the issue? Is there some workflow that uses the `assignee` tag instead, or is it more about showing intent before a submitted PR?

Mugurell:
> > This seems to be a leftover from #25980.
> > Searching the project for onPrepareOptionsMenu / onCreateOptionsMenu seems like there is one more use of the now deprecated APIs that we need to fix
> 
> There however does appear to be more instances of [`invalidateOptionsMenu`](https://searchfox.org/mozilla-mobile/search?q=OptionsMenu&path=&case=true&regexp=false) which may not be correct. CC @mcarare

Seems to me that the problematic method - which breaks the intended functionality is `onPrepareOptionsMenu` which [is  deprecated](https://developer.android.com/reference/android/app/Fragment#onPrepareOptionsMenu(android.view.Menu)) and not being called for us anymore.
The same `onPrepareOptionsMenu` from Activity [is not deprecated](https://developer.android.com/reference/android/app/Activity#onPrepareOptionsMenu(android.view.Menu)), neither [invalidateOptionsMenu](https://developer.android.com/reference/android/app/Activity#invalidateOptionsMenu()) and it seems like just renaming `onPrepareOptionsMenu` to `onPrepareMenu` gets everything working as before. Can you confirm?
Searching across the projects there are only [two occurences of `onPrepareOptionsMenu`](https://searchfox.org/mozilla-mobile/search?q=onPrepareOptionsMenu&path=&case=true&regexp=false) that we need to refactor since at the moment they are no-op and a source of bugs.

If `invalidateOptionsMenu` does still trigger a menu update refactoring it to `onPrepareMenu` is not an immediate priority and could be done in a separate `eng-health` ticket.

Mugurell:
> > @fice-t If you would leave a small comment in #21414 , maybe something around you wanted to work on the issue / already having a PR for it we will be able to assign the ticket to you. (Same for the others you've helped with).
> 
> Sure, though does it offer much more than Github showing the linked PR in the issue? Is there some workflow that uses the `assignee` tag instead, or is it more about showing intent before a submitted PR?

More so for visibility of your contributions and also showing to others that that ticket is in progress.

fice-t:
@Mugurell According to the [source](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:fragment/fragment/src/main/java/androidx/fragment/app/FragmentActivity.java;l=756;drc=4b7eab390ed1076f11bf0ecf2458d3e716afe3ac) it appears that `invalidateMenu` just internally calls `invalidateOptionsMenu`, so it seems that using it is just for consistency and thus it's not really an issue.

Mugurell:
> @Mugurell According to the [source](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:activity/activity/src/main/java/androidx/activity/ComponentActivity.java;l=544;drc=4b7eab390ed1076f11bf0ecf2458d3e716afe3ac) it appears that `invalidateMenu` just internally calls `invalidateOptionsMenu`, so it seems that using it is just for consistency and thus it's not really an issue.

Great! Thank you for checking!

PR #27448, opened 2022-10-18
Update jna to version 5.12.1

rvandermeulen:
This has landed everywhere except Fenix at this point without known regressions.

Changelog:
https://github.com/java-native-access/jna/blob/master/CHANGES.md#release-5121


jonalmeida (approved):
🤞

PR #27451, opened 2022-10-18
Bug 1791998 - Disable TCP popup when running performance tests.

gmierz:
This patch adds some code to disable the TCP popup when running performance tests on Fenix.


gmierz:
This patch is ready for review, you can find the perf tests here: https://github.com/mozilla-mobile/fenix/runs/8963834509

Note that the TCP popup is still visible but this is because we aren't passing `--ez performancetest true` as an intent from Raptor. This will be resolved in a patch on mozilla-central.

jonalmeida (approved):
This change looks straight-forward and as far as [I can see](https://searchfox.org/mozilla-mobile/search?path=&q=shouldShowTotalCookieProtectionCFR), does not interact with any other usages of this flag.

gmierz:
Thanks for the review @jonalmeida!

PR #27454, opened 2022-10-18
Bug 1783561 - Better snackbar placement

Mugurell:
This patches will mimic the UX of a modal dialog while using Android Views.
This meant including a scrim that would consume all touches and theming the
navigation bar and status bar.
Avoiding a dialog and a separate window will allow the snackbar to see the
new "dialog" as a sibling in a CoordinatorLayout parent and so be able to
position itself based on the new "dialog".

Previously Android-Components - BrowserToolbarBehavior would be responsible
for positioning the snackbar above the toolbar.
With that responsibility removed we can handle in Fenix positioning the
snackbar depending on the toolbar and many more cases - like positioning it
depending on the download dialogs.

Needs https://github.com/mozilla-mobile/firefox-android/pull/66

Result:

https://user-images.githubusercontent.com/11428869/196496915-852e0254-9c0b-4942-a9c1-8efa6665e33e.mp4

### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation
Fixes #1783561

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesComposables.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesComposables.kt
@@ -279,32 +279,43 @@ fun PocketStories(
         itemsIndexed(storiesToShow) { columnIndex, columnItems ->
             Column(verticalArrangement = Arrangement.spacedBy(8.dp)) {
                 columnItems.forEachIndexed { rowIndex, story ->
-                    if (story == placeholderStory) {
-                        ListItemTabLargePlaceholder(stringResource(R.string.pocket_stories_placeholder_text)) {
-                            onDiscoverMoreClicked("https://getpocket.com/explore?$POCKET_FEATURE_UTM_KEY_VALUE")
-                        }
-                    } else if (story is PocketRecommendedStory) {
-                        PocketStory(
-                            story = story,
-                            backgroundColor = backgroundColor,
-                        ) {
-                            val uri = Uri.parse(story.url)
-                                .buildUpon()
-                                .appendQueryParameter(URI_PARAM_UTM_KEY, POCKET_STORIES_UTM_VALUE)
-                                .build().toString()
-                            onStoryClicked(it.copy(url = uri), rowIndex to columnIndex)
-                        }
-                    } else if (story is PocketSponsoredStory) {
-                        Box(
-                            modifier = Modifier.onShown(0.5f) {
-                                onStoryShown(story, rowIndex to columnIndex)
-                            },
-                        ) {
-                            PocketSponsoredStory(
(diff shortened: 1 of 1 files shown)


Mugurell:
Draft for now. Will add documentation and tests.

app/src/main/res/values/colors.xml
```diff
@@ -337,4 +337,7 @@
 
     <!-- App Spinners colors -->
     <color name="spinner_selected_item">#1415141A</color>
+
+    <!-- Material Design colors -->
+    <color name="material_scrim_color">#52000000</color>
```
gabrielluong, line 342: We should have a design system scrim color

Mugurell: The app already uses a custom scrim color for the share screen but that has a blue tint.
The value I added here is a a black color with 32% opacity which seems to mimic exactly what the scrim would be when using a dialog and is a value recommended in the material design specs - https://m2.material.io/develop/android/theming/motion

app/src/main/res/layout/start_download_dialog_layout.xml
```diff
@@ -0,0 +1,96 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
```
Mugurell, line 2: This file is a full "port" of https://github.com/mozilla-mobile/android-components/blob/main/components/feature/downloads/src/main/res/layout/mozac_downloads_prompt.xml.
This to not need another UX review but at the same time allow to avoid this rather ugly (but necessary there) [code from Fenix](https://github.com/mozilla-mobile/fenix/blob/106b692c410237642defbbb0c203d43f52ea012b/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt#L493-L501) and [this code from AC](https://github.com/mozilla-mobile/android-components/blob/d723a16e11f786f55f809e2891bb1e72aaf1c135/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/SimpleDownloadDialogFragment.kt#L97-L121).
It's simpler (and involves a lot less code) to set the values in xml and this is the difference between this layout and the one from A-C.

Mugurell:
Saw some issues when testing with Talkback.

Mugurell:
> Saw some issues when testing with Talkback.

Fixed the issues in commit #2. 
Needed to 
- prevent other sibling to be selectable by Talkback (the previously used dialog would not have this issue because it would be shown in another window and there would be no other siblings.
- have the title be focused first. By default seems like the close button would be focused first - different from the behavior seen in the current Nightly.


https://user-images.githubusercontent.com/11428869/196899097-07a074a6-66e5-492f-b16d-e2c64092dbda.mp4

PR #27517, opened 2022-10-24
For #27524 Set as default browser toggle get set to true after user elect Fenix as default browser

### LINKED ISSUE #21857: allow pinned Top Sites to be rearranged
Feature request: "pinned" Top sites (there is a pin icon beside it) should be rearrangeable. There could be a few reasons for doing this, but the reason I'm requesting it is because I use my phone with 1 hand and I would like to be able to rearrange the top site tiles so that the ones I use the most are closest to my thumb, and the ones I use the least would require a bit of a thumb stretch to tap. 

Apologies if this is a duplicate

### LINKED ISSUE #27524: [Bug]: Set as default browser doesn't get set to true
### Steps to reproduce

1.Install a debug Firefox Fenix build 108 .
2.Go to settings.
3.Set Fenix as default browser.
4.Toggle remains the same , set to false.

### Expected behaviour

The toggle should be set to true.

### Actual behaviour

Toggle remains the same , set to false.

### Device name

Pixel 3Xl 

### Android version

android 12

### Firefox release type

Firefox

### Firefox version

108

### Device logs

_No response_

### Additional information

https://mozilla-hub.atlassian.net/browse/FNXV2-21857

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21862)

iorgamgabriel:
### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation
Fixes #27524

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/settings/SettingsFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/SettingsFragment.kt
@@ -432,7 +432,8 @@ class SettingsFragment : PreferenceFragmentCompat() {
         val preferenceLeakCanary = findPreference<Preference>(leakKey)
         val preferenceRemoteDebugging = findPreference<Preference>(debuggingKey)
         val preferenceMakeDefaultBrowser =
-            requirePreference<Preference>(R.string.pref_key_make_default_browser)
+            requirePreference<DefaultBrowserPreference>(R.string.pref_key_make_default_browser)
+
         val preferenceOpenLinksInExternalApp =
             findPreference<Preference>(getPreferenceKey(R.string.pref_key_open_links_in_external_app))
         if (!Config.channel.isReleased) {
@@ -451,8 +452,11 @@ class SettingsFragment : PreferenceFragmentCompat() {
             true
         }
 
-        preferenceMakeDefaultBrowser.onPreferenceClickListener =
-            getClickListenerForMakeDefaultBrowser()
+        preferenceMakeDefaultBrowser.apply {
+            updateSwitch()
+            onPreferenceClickListener =
+                getClickListenerForMakeDefaultBrowser()
+        }
 
         preferenceOpenLinksInExternalApp?.onPreferenceChangeListener = SharedPreferenceUpdater()
 


iorgamgabriel:
https://user-images.githubusercontent.com/8118371/197506500-da6de9d3-cb7d-4f45-9b0c-b0087ca6b0d8.mp4

app/src/main/java/org/mozilla/fenix/settings/SettingsFragment.kt
```diff
@@ -431,8 +435,8 @@ class SettingsFragment : PreferenceFragmentCompat() {
         val debuggingKey = getPreferenceKey(R.string.pref_key_remote_debugging)
         val preferenceLeakCanary = findPreference<Preference>(leakKey)
         val preferenceRemoteDebugging = findPreference<Preference>(debuggingKey)
-        val preferenceMakeDefaultBrowser =
-            requirePreference<Preference>(R.string.pref_key_make_default_browser)
+        preferenceMakeDefaultBrowser =
```
Mugurell, line 438:
```suggestion
val preferenceMakeDefaultBrowser =
    requirePreference<DefaultBrowserPreference>(R.string.pref_key_make_default_browser)
```

Don't think we need to cache the preference and update it in other methods.
The proposed flow would have the value updated in `onResume` but that method already calls `update()` which calls this method - `setupPreferences()`. So it looks like here is the place where we'd need to ensure all preferences are correctly configured and displayed.

Seems like below a click listener is set for the default browser preference so that call can be updated to something similar to:
```kotlin
preferenceMakeDefaultBrowser.apply {
    updateSwitch()
    onPreferenceClickListener =
        getClickListenerForMakeDefaultBrowser()
}
```

mcarare:
nit: issue number?

iorgamgabriel:
https://mozilla-hub.atlassian.net/browse/FNXV2-21857

app/src/main/java/org/mozilla/fenix/settings/SettingsFragment.kt
iorgamgabriel: Done

Mugurell (approved):

PR #27525, opened 2022-10-24
 Bug 1796482 - Report save to PDF errors. 

Amejia481:
Requires https://github.com/mozilla-mobile/android-components/pull/12995

### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation

Used by GitHub Actions.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/share/SaveToPDFMiddleware.kt
+++ b/app/src/main/java/org/mozilla/fenix/share/SaveToPDFMiddleware.kt
@@ -1,0 +1,46 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.share
+
+import android.content.Context
+import android.widget.Toast
+import android.widget.Toast.LENGTH_LONG
+import kotlinx.coroutines.DelicateCoroutinesApi
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.launch
+import mozilla.components.browser.state.action.BrowserAction
+import mozilla.components.browser.state.action.EngineAction
+import mozilla.components.browser.state.state.BrowserState
+import mozilla.components.lib.state.Action
+import mozilla.components.lib.state.Middleware
+import mozilla.components.lib.state.MiddlewareContext
+import org.mozilla.fenix.R
+
+/**
+ * [BrowserAction] middleware reacting in response to Save to PDF related [Action]s.
+ * @property context An Application context.
+ */
+class SaveToPDFMiddleware(
+    private val context: Context,
+) : Middleware<BrowserState, BrowserAction> {
+
+    @OptIn(DelicateCoroutinesApi::class)
+    override fun invoke(
+        ctx: MiddlewareContext<BrowserState, BrowserAction>,
+        next: (BrowserAction) -> Unit,
+        action: BrowserAction,
+    ) {
+        if (action is EngineAction.SaveToPdfExceptionAction) {
+            // See https://github.com/mozilla-mobile/fenix/issues/27649 for more details,
+            // why a Toast is used here.
+            GlobalScope.launch(Dispatchers.Main) {
+                Toast.makeText(context, R.string.unable_to_save_to_pdf_error, LENGTH_LONG).show()
+            }
+        } else {
+            next(action)
+        }
+    }
+}
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -1028,6 +1028,8 @@
     <string name="share_button_content_description">Share</string>
     <!-- Text for the Save to PDF feature in the share menu -->
     <string name="share_save_to_pdf">Save as PDF</string>
+    <!-- Text for error message when generating a PDF file Text for error message when generating a PDF file. -->
+    <string name="unable_to_save_to_pdf_error">Unable to generate PDF</string>
     <!-- Sub-header in the dialog to share a link to another sync device -->
     <string name="share_device_subheader">Send to device</string>
     <!-- Sub-header in the dialog to share a link to an app from the full list -->
--- a/app/src/main/java/org/mozilla/fenix/components/Core.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/Core.kt
@@ -92,6 +92,7 @@ import org.mozilla.fenix.perf.StrictModeManager
 import org.mozilla.fenix.perf.lazyMonitored
 import org.mozilla.fenix.settings.SupportUtils
 import org.mozilla.fenix.settings.advanced.getSelectedLocale
+import org.mozilla.fenix.share.SaveToPDFMiddleware
 import org.mozilla.fenix.telemetry.TelemetryMiddleware
 import org.mozilla.fenix.utils.getUndoDelay
 import org.mozilla.geckoview.GeckoRuntime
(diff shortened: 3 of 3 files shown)


app/src/main/java/org/mozilla/fenix/share/SaveToPDFMiddleware.kt
```diff
@@ -0,0 +1,43 @@
...
+class SaveToPDFMiddleware(
+    private val context: Context,
+) : Middleware<BrowserState, BrowserAction> {
+    override fun invoke(
+        ctx: MiddlewareContext<BrowserState, BrowserAction>,
+        next: (BrowserAction) -> Unit,
+        action: BrowserAction,
+    ) {
+        if (action is EngineAction.SaveToPdfErrorAction) {
+            val rootView = context.getRootView() ?: return
```
jonalmeida, line 33: When testing, `context.getRootView()` was always `null`, so this code path was always early returned. This is probably because `context.isUiContext` is always `false` and when I queried for `context.display`, the error message was more useful:
```
java.lang.UnsupportedOperationException: Tried to obtain display from a Context not associated with one. Only visual Contexts (such as Activity or one created with Context#createWindowContext) or ones created with Context#createDisplayContext are associated with displays. Other types of Contexts are typically related to background entities and may return an arbitrary display.
```

app/src/main/res/values/strings.xml
```diff
@@ -1011,6 +1011,8 @@
     <string name="share_button_content_description">Share</string>
     <!-- Text for the Save to PDF feature in the share menu -->
     <string name="share_save_to_pdf">Save as PDF</string>
+    <!-- Text for error message when generating a PDF file. -->
```
jonalmeida, line 1014: nit

```suggestion
    <!-- Text for error message when generating a PDF file fails. -->
```

app/src/main/java/org/mozilla/fenix/share/SaveToPDFMiddleware.kt
jonalmeida: That said, if it did work however, I think we would be potentially opening up a memory leak here: if we using the _application_ Context to attach to a view, then that view is also going to be tied to the BrowserStore. If we do not safely release the `FenixSnackbar`, then I think we are going to hold onto some parts of the snackbar view for the application Context lifetime.

As for a solution to this potential problem, I think there are a few options:

 1. Dispatch some state change to the AppStore that will then be observed from a Context associated with a display to show the `Snackbar` and then consume the value.
 2. Similar to the above, but avoid the indirection and put this state straight into `BrowserStore`. Something a bit like `DownloadState` but on the root of the state object, I think? 
 3. (Not recommended) Use a new pattern that can register/unregister from the middleware so it can observe the action changes from a listener and call `FenixSnackbar` from an appropriate context (like HomeActivity). We're now re-creating how to observe state via odd loop-holes so we should not do this. It was more of a thought experiment. 🙂

app/src/main/java/org/mozilla/fenix/perf/StrictModeManager.kt
```diff
@@ -41,7 +41,7 @@ open class StrictModeManager(
     private val components: Components,
 ) {
 
-    private val isEnabledByBuildConfig = config.channel.isDebug
+    private val isEnabledByBuildConfig = !config.channel.isDebug
```
jonalmeida, line 44: tenor-390591897

jonalmeida (changes requested):
I wasn't able to test out this PR end-to-end (see the review on the AC patch), so I used this patch below to trigger the appropriate action and try and trigger the flow. This might break expected state for the action to be triggered in a way that I haven't considered, so happy to wait for the AC changes before we can re-review this again.

<details>

<summary>Sample patch to dispatch `SaveToPdfErrorAction`.</summary>

```diff
diff --git a/app/src/main/java/org/mozilla/fenix/share/ShareFragment.kt b/app/src/main/java/org/mozilla/fenix/share/ShareFragment.kt
index 0574448ac..4ee9cf768 100644
--- a/app/src/main/java/org/mozilla/fenix/share/ShareFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/share/ShareFragment.kt
@@ -20,6 +20,7 @@ import androidx.lifecycle.lifecycleScope
 import androidx.navigation.fragment.findNavController
 import androidx.navigation.fragment.navArgs
 import mozilla.components.browser.state.action.ContentAction
+import mozilla.components.browser.state.action.EngineAction
 import mozilla.components.browser.state.selector.findTabOrCustomTab
 import mozilla.components.concept.engine.prompt.PromptRequest
 import mozilla.components.feature.accounts.push.SendTabUseCases
@@ -124,7 +125,13 @@ class ShareFragment : AppCompatDialogFragment() {
                 setContent {
                     FirefoxTheme {
                         SaveToPDFItem {
-                            shareInteractor.onSaveToPDF(tabId = args.sessionId)
+                            // I didn't need to remove the line above, but this made
+                            // it easier to ensure the flow was doing what I expected it to.
+                            dismiss()
+                            requireComponents.core.store.dispatch(
+                                EngineAction.SaveToPdfErrorAction(
+                                    args.sessionId!!,
+                                    Exception("AAHHHHHHHHH"),
+                                ),
+                            )
                         }
                     }
                 }
```

</details>

app/src/main/java/org/mozilla/fenix/share/SaveToPDFMiddleware.kt
```diff
@@ -0,0 +1,43 @@
...
+            FenixSnackbar.make(
+                view = rootView,
+                duration = LENGTH_LONG,
+                isDisplayedWithBrowserToolbar = true,
+            ).setText(context.getString(R.string.unable_to_save_to_pdf_error))
+                .show()
+        }
+        next(action)
```
jonalmeida, line 41: Do we also want to report telemetry to know how many times Save to PDF failed @kbrosnan ? 

Value: We have telemetry for `save_to_pdf_tapped` events, so we could better observe the rate of failures as the feature goes out? We can let this expire after 2-3 releases which would be useful platform telemetry to have.

app/src/main/java/org/mozilla/fenix/perf/StrictModeManager.kt
Amejia481: XD

app/src/main/java/org/mozilla/fenix/share/SaveToPDFMiddleware.kt
Amejia481: This will be address in separated ticket

app/src/main/java/org/mozilla/fenix/share/SaveToPDFMiddleware.kt
```diff
@@ -0,0 +1,45 @@
...
+) : Middleware<BrowserState, BrowserAction> {
+
+    @OptIn(DelicateCoroutinesApi::class)
+    override fun invoke(
+        ctx: MiddlewareContext<BrowserState, BrowserAction>,
+        next: (BrowserAction) -> Unit,
+        action: BrowserAction,
+    ) {
+        if (action is EngineAction.SaveToPdfExceptionAction) {
+            // See https://github.com/mozilla-mobile/fenix/issues/27649 for more details.
```
jonalmeida, line 37: nit:
```suggestion
            // See https://github.com/mozilla-mobile/fenix/issues/27649 for more details why a `Toast` is used here.
```

jonalmeida (approved):
Looks good. 👍

app/src/main/java/org/mozilla/fenix/share/SaveToPDFMiddleware.kt
Amejia481: Addressed!

PR #27554, opened 2022-10-26
For #26475 - Remove unnecessary features flags

### LINKED ISSUE #26475: Remove unused features flags
We have some feature flags in the code base that are no longer needed because they have shipped and there are not related bugs that are 
- `showRecentTabsFeature`
- `historyMetadataUIFeature`
- `recentBookmarksFeature`
- `inactiveTabs`
- `tabReorderingFeature` - there is one code reference in there that we can remove as well.
- `showHomeOnboarding`
-  `messagingFeature`

This will also help reduce some dynamic UI that shows up lazily based on flags.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21305)

gabrielluong:
### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation
Fixes #26475


app/src/main/java/org/mozilla/fenix/FeatureFlags.kt
```diff
@@ -31,41 +31,11 @@ object FeatureFlags {
...
-
-    /**
-     * Enables the "recent" tabs feature in the home screen.
-     */
-    const val showRecentTabsFeature = true
-
     /**
      * Enables UI features based on history metadata.
      */
     const val historyMetadataUIFeature = true
```
jonalmeida, line 37: Can we remove `historyMetadataUIFeature` too?

jonalmeida (approved):
Thanks this is much better experience already in Settings UI without the animations of settings jumping in.

app/src/main/java/org/mozilla/fenix/FeatureFlags.kt
gabrielluong: This was intended as a followup. If you look at the conditional in https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/java/org/mozilla/fenix/utils/Settings.kt#1266, it gets kinda weird because we essentially have `true || isHistoryMetadataEnabled`, but losing that `true` from removing `historyMetadataUIFeature` might have some unintended consequences. So, it required digging into `isHistoryMetadataEnabled`, which felt a bit too much in this simple PR.

jonalmeida: `isHistoryMetadataEnabled` is used only for `historyMetadataUIFeature` so it looks like that is fallback code that wasn't removed before which you could remove here too or just turn `featureFlag` to true for `historyMetadataUIFeature` if you'd like.

PR #27444, opened 2022-10-18
Combine Google package name regex into a single one #27336

chetanmuliya:
### Pull Request checklist

- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features. In addition, it includes a screenshot of a successful [accessibility scan](https://play.google.com/store/apps/details?id=com.google.android.apps.accessibility.auditor&hl=en_US) to ensure no new defects are added to the product.

### QA

- [x] **QA Needed**

### To download an APK when reviewing a PR (after all CI tasks finished running):
1. Click on `Checks` at the top of the PR page.
2. Click on the `firefoxci-taskcluster` group on the left to expand all tasks.
3. Click on the `build-debug` task.
4. Click on `View task in Taskcluster` in the new `DETAILS` section.
5. The APK links should be on the right side of the screen, named for each CPU architecture.

### GitHub Automation
Fixes #27336

PR #27557, opened 2022-10-26
CI for #27444

### LINKED ISSUE #27336: Combine Google package name regex into a single one
> When `comGoogleAndroid` and `comGoogleFirebase` was originally added ([commit](https://github.com/mozilla-mobile/fenix/commit/f0464b9e72e0ea4dca94a46f8e0f05a285955c44#diff-75b31066e97e27961c3c9e9668c539a133d73c10398dc76552df3f0119810ac0R236)), I don't think it was possible (or at least easy) to generalize because `comGoogleAndroid` explicitly excluded some `com.google.android.*` deps that hadn't yet migrated from maven central to the google repo, i.e. it was defined as:
> ```kotlin
>     const val comGoogleAndroid = "com\\.google\\.android\\.(?!apps\\.common\\.testing\\.accessibility\\.framework).*"
> ```
> 
> Now that it's defined more generically (i.e. Google is hosting more packages on their repo), it might be possible. In that case, it'd also be preferable to reduce duplication (like the root cause of this issue).

_Originally posted by @mcomella in https://github.com/mozilla-mobile/fenix/issues/27321#issuecomment-1271843462_
      

1. We want to take the `comGoogleFirebase`, `comGoogleTesting`, and `comGoogleAndroid` defined [here](https://github.com/mozilla-mobile/fenix/blob/effc40dc3245f7077830bdb80ae482b61c1ba1f8/buildSrc/src/main/java/Dependencies.kt#L281) to be combined into a single `comGoogle` with the value: `com\\.google\\..*`.
2. We also need to update all the previous references to use just this one.
3. Remove [this comment](https://github.com/mozilla-mobile/fenix/blob/effc40dc3245f7077830bdb80ae482b61c1ba1f8/buildSrc/src/main/java/Dependencies.kt#L284-L289) as well.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21762)

jonalmeida:
Fixes #27336
Fixes #27336
Fixes #27336

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -34,10 +34,8 @@ buildscript {
                 content {
                     // Improve performance: only check google maven for google deps.
                     includeGroupByRegex RepoMatching.androidx
-                    includeGroupByRegex RepoMatching.comGoogleAndroid
-                    includeGroupByRegex RepoMatching.comGoogleFirebase
-                    includeGroupByRegex RepoMatching.comGoogleTesting
                     includeGroupByRegex RepoMatching.comAndroid
+                    includeGroupByRegex RepoMatching.comGoogle
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


gabrielluong:
I believe this is no longer needed since https://github.com/mozilla-mobile/fenix/pull/27444 is merged. Closing.

PR #27602, opened 2022-10-28
For #27401 - refactor SelectableChip into Chip composable

### LINKED ISSUE #27401: Refactor `SelectableChip` to be a top level `Chip` component
The `SelectableChip` Composable is effectively a `Chip` that just handles re-colors based on the `isSelected` boolean. The component would be best served as a generic `Chip` component (similar to `TextButton`) that has a condensed list of parameters and uplifts any stateful logic (such as selected vs unselected) to the actual UI that uses it.

The contract should be something like:

``` @Composable
fun Chip(
    text: String,
    textColor: Color? = null,
    backgroundColor: Color? = null,
    onClick: (() -> Unit)? = null,
) {
```

The other Chip component use cases (such as having an optional icon) could be implemented as well.

The text & default colors should be double checked against their specs in [FIgma](https://www.figma.com/file/ujdiReypSypu2GhRgzArea/Android-Components?node-id=1808%3A7427).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21795)

HarrisonOg:
### Pull Request checklist

### QA

- [ ] **QA Needed**

### GitHub Automation
Fixes #27401


app/src/main/java/org/mozilla/fenix/compose/Chip.kt
```diff
@@ -0,0 +1,97 @@
...
+                if (onClick != null) {
+                    onClick()
+                }
+            }
+            .clip(MaterialTheme.shapes.small)
+            .background(color = backgroundColor)
+            .padding(horizontal = 16.dp, vertical = 10.dp),
+    ) {
+        Text(
+            text = text.capitalize(Locale.current),
```
MozillaNoah, line 59: it might make more sense that we leave the capitalization up to the consumer of this Composable, rather than enforcing it here.

app/src/main/java/org/mozilla/fenix/compose/Chip.kt
```diff
@@ -0,0 +1,97 @@
...
+            style = TextStyle(fontSize = 14.sp),
```
MozillaNoah, line 60: let's take this opportunity to use the corresponding style from FenixTypography.

app/src/main/java/org/mozilla/fenix/compose/Chip.kt
```diff
@@ -0,0 +1,97 @@
...
+    Box(
+        modifier = modifier
+            .selectable(false) {
+                if (onClick != null) {
+                    onClick()
+                }
+            }
+            .clip(MaterialTheme.shapes.small)
+            .background(color = backgroundColor)
+            .padding(horizontal = 16.dp, vertical = 10.dp),
```
MozillaNoah, line 56: Let's instead have the root Composable be the `androidx.compose.material.Chip`, and we'll simply wrap and extend the functionality around that.

app/src/main/java/org/mozilla/fenix/compose/Chip.kt
```diff
@@ -0,0 +1,97 @@
...
+}
+
+/**
+ * An example of a selectable chip.
+ */
+@Composable
+@Preview(uiMode = Configuration.UI_MODE_NIGHT_YES)
+@Preview(uiMode = Configuration.UI_MODE_NIGHT_NO)
+private fun SelectableChipPreview() {
+    val selectedState = remember { mutableStateOf(false) }
```
MozillaNoah, line 73:
```suggestion
    val selectedState by remember { mutableStateOf(false) }
```

app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesComposables.kt
```diff
@@ -399,15 +399,23 @@ fun PocketStoriesCategories(
             verticalItemsSpacing = 16.dp,
         ) {
             categories.filter { it.name != POCKET_STORIES_DEFAULT_CATEGORY_NAME }.forEach { category ->
-                SelectableChip(
+                val selectedState = remember { mutableStateOf(selections.map { it.name }.contains(category.name)) }
```
MozillaNoah, line 402:
```suggestion
                val selected by remember { mutableStateOf(selections.map { it.name }.contains(category.name)) }
```

app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesComposables.kt
```diff
@@ -399,15 +399,23 @@ fun PocketStoriesCategories(
...
+
+                Chip(
                     text = category.name,
-                    isSelected = selections.map { it.name }.contains(category.name),
-                    selectedTextColor = selectedTextColor,
-                    unselectedTextColor = unselectedTextColor,
-                    selectedBackgroundColor = selectedBackgroundColor,
-                    unselectedBackgroundColor = unselectedBackgroundColor,
+                    textColor = if (selectedState.value) {
```
MozillaNoah, line 406:
```suggestion
                    textColor = if (selected) {
```

app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesComposables.kt
```diff
@@ -399,15 +399,23 @@ fun PocketStoriesCategories(
...
+                        selectedTextColor ?: FirefoxTheme.colors.textActionPrimary
+                    } else {
+                        unselectedTextColor ?: FirefoxTheme.colors.textActionPrimary
+                    },
+                    backgroundColor = if (selectedState.value) {
```
MozillaNoah, line 411:
```suggestion
                    backgroundColor = if (selected) {
```

app/src/main/java/org/mozilla/fenix/compose/Chip.kt
MozillaNoah: This [SO post](https://stackoverflow.com/questions/69326413/whats-difference-between-remember-and-by-remember-kotlin-jetpack-compo) has a decent explanation, but essentially `by remember` shortens the syntax and makes it so you don't have to unwrap it with `.value` everywhere.

MozillaNoah (commented):
First round of feedback. Looking good so far!

app/src/main/java/org/mozilla/fenix/compose/Chip.kt
HarrisonOg: ah that makes sense, making the change

app/src/main/java/org/mozilla/fenix/compose/Chip.kt
HarrisonOg: thank you! I didn't know that

app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesComposables.kt
gabrielluong: 👍 on the `selected` naming

app/src/main/java/org/mozilla/fenix/compose/Chip.kt
```diff
@@ -0,0 +1,97 @@
...
+import androidx.compose.ui.text.intl.Locale
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+import org.mozilla.fenix.theme.FirefoxTheme
+
+/**
+ * Default layout for a clickable chip.
+ *
+ * @param text [String] displayed in this chip. Ideally should only be one word.
```
gabrielluong, line 33:
```suggestion
 * @param text [String] displayed in this chip.
```

app/src/main/java/org/mozilla/fenix/compose/Chip.kt
```diff
@@ -0,0 +1,97 @@
...
+ * @param modifider [Modifier] used to be applied to the layout of the chip.
+ * @param textColor optional text [Color] for the chip.
```
gabrielluong, line 35:
```suggestion
 * @param textColor Optional text [Color] for the chip.
```

app/src/main/java/org/mozilla/fenix/compose/Chip.kt
```diff
@@ -0,0 +1,97 @@
...
+ * @param backgroundColor optional background [Color] for the chip.
+ * @param onClick optional Callback for when the user taps this chip.
```
gabrielluong, line 37:
```suggestion
 * @param textColor Optional text [Color] for the chip.
 * @param backgroundColor Optional background [Color] for the chip.
 * @param onClick Optional Callback for when the user taps this chip.
```

Mukuljangir372:
@HarrisonOg Can you please tell me how can I request for reviewer?

app/src/main/java/org/mozilla/fenix/compose/Chip.kt
HarrisonOg: oh I'll fix this too hold on

app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesComposables.kt
```diff
@@ -443,9 +443,11 @@ fun PocketStoriesCategories(
             verticalItemsSpacing = 16.dp,
         ) {
             categories.filter { it.name != POCKET_STORIES_DEFAULT_CATEGORY_NAME }.forEach { category ->
+                var selectedState by remember { mutableStateOf(selections.map { it.name }.contains(category.name)) }
```
gabrielluong, line 446:
```suggestion
                var isSelected by remember { mutableStateOf(selections.map { it.name }.contains(category.name)) }
```

app/src/main/java/org/mozilla/fenix/compose/Chip.kt
```diff
@@ -0,0 +1,78 @@
...
+import org.mozilla.fenix.theme.FirefoxTheme
+
+/**
+ * Default layout for a clickable chip.
+ *
+ * @param text [String] displayed in this chip.
+ * @param modifider [Modifier] used to be applied to the layout of the chip.
+ * @param textColor Optional text [Color] for the chip.
+ * @param backgroundColor Optional background [Color] for the chip.
+// * @param enabled [Boolean] used to determine if chip responds to user input
```
gabrielluong, line 28: nit: Got some cleanup to do before we get back to another review

app/src/main/java/org/mozilla/fenix/compose/SelectableChip.kt
```diff
@@ -4,33 +4,29 @@
 
 package org.mozilla.fenix.compose
 
-import android.content.res.Configuration.UI_MODE_NIGHT_NO
-import android.content.res.Configuration.UI_MODE_NIGHT_YES
+import android.content.res.Configuration
+import android.util.Log
```
gabrielluong, line 8: 👀

app/src/main/java/org/mozilla/fenix/compose/SelectableChip.kt
```diff
@@ -42,40 +38,33 @@ import org.mozilla.fenix.theme.FirefoxTheme
...
-        )
+        selected = !selected
+        onClick()
     }
 }
 
+/**
+ * An example of a [SelectableChip] with the selected value accessible outside the
+ * Chip.
+ *
```
gabrielluong, line 63: 👀

app/src/main/java/org/mozilla/fenix/compose/SelectableChip.kt
```diff
@@ -84,15 +73,21 @@ private fun SelectableChipPreview() {
                 .background(FirefoxTheme.colors.layer1),
             horizontalArrangement = Arrangement.SpaceEvenly,
         ) {
-            SelectableChip(text = "Chirp", isSelected = false) { }
-            SelectableChip(text = "Chirp", isSelected = true) { }
+            var firstSelected by remember { mutableStateOf(false) }
+            var secondSelected by remember { mutableStateOf(true) }
+
+            SelectableChip(text = "ChirpOne", isSelected = firstSelected) {}
+            SelectableChip(text = "ChirpTwo", isSelected = secondSelected) { }
```
gabrielluong, line 80:
```suggestion
            SelectableChip(text = "ChirpTwo", isSelected = secondSelected) {}
```

gabrielluong (commented):
I think this is looking good from my quick glance, but I think we're still in the draft phase even though I got the review ping. There's some code polish that needs to be done before I think a full review in earnest should be done. My recommendation is to look through the full changes yourselves and essentially do a mini review of your own changes. I will naturally point out any inconsistencies, areas where either the naming or docs could be improved or code that could be simplified further.

Also, a note about squashing/amending the typography commit into the main commit. It's okay to do multi-commits when we want to split up complex things, but each commit should still try to resolve an issue instead of a fixup.

app/src/main/java/org/mozilla/fenix/compose/SelectableChip.kt
```diff
@@ -116,3 +111,40 @@ private fun SelectableChipWithCustomColorsPreview() {
...
+    FirefoxTheme {
+        Row(
+            modifier = Modifier
+                .fillMaxWidth()
+                .background(FirefoxTheme.colors.layer1),
+            horizontalArrangement = Arrangement.SpaceEvenly,
+        ) {
+            var isSelected by remember { mutableStateOf(false) }
+
+            androidx.compose.material.FilterChip(
```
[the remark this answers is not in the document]
HarrisonOg: ah I'll remove this, I was experimenting

app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesComposables.kt
gabrielluong: Thinking about this some more - is this mutable state necessary here since we remember the state inside of SelectableChip as well?

HarrisonOg: we don't strictly need this to be mutable if we're not referring to the changed state anywhere, I'll take a further look

app/src/main/java/org/mozilla/fenix/compose/SelectableChip.kt
HarrisonOg: removed log statements

app/src/main/java/org/mozilla/fenix/compose/SelectableChip.kt
```diff
@@ -42,40 +38,33 @@ import org.mozilla.fenix.theme.FirefoxTheme
...
     text: String,
     isSelected: Boolean,
-    selectedTextColor: Color? = null,
-    unselectedTextColor: Color? = null,
-    selectedBackgroundColor: Color? = null,
-    unselectedBackgroundColor: Color? = null,
+    selectedTextColor: Color = FirefoxTheme.colors.textActionPrimary,
+    unselectedTextColor: Color = FirefoxTheme.colors.textActionPrimary,
+    selectedBackgroundColor: Color = FirefoxTheme.colors.actionPrimary,
+    unselectedBackgroundColor: Color = FirefoxTheme.colors.actionTertiary,
```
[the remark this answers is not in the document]
MozillaNoah: Sounds good. I'd suggested doing anything related to `PocketStoriesCategoryColors` as a follow-up to keep this PR just to the Chip.

app/src/main/java/org/mozilla/fenix/compose/Chip.kt
```diff
@@ -0,0 +1,78 @@
...
+        colors = chipColors(
+            backgroundColor = backgroundColor,
+            disabledBackgroundColor = FirefoxTheme.colors.actionTertiary,
+            contentColor = textColor,
+            disabledContentColor = FirefoxTheme.colors.textActionTertiary,
+        )
+    ) {
+        Text(
+            text = text,
+            style = FirefoxTheme.typography.button,
```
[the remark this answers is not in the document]
HarrisonOg: fixed

app/src/main/java/org/mozilla/fenix/compose/Chip.kt
```diff
@@ -0,0 +1,78 @@
...
+) {
+    androidx.compose.material.Chip(
+        onClick = onClick,
+        enabled = enabled,
+        modifier = modifier,
+        colors = chipColors(
+            backgroundColor = backgroundColor,
+            disabledBackgroundColor = FirefoxTheme.colors.actionTertiary,
+            contentColor = textColor,
+            disabledContentColor = FirefoxTheme.colors.textActionTertiary,
```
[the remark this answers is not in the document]
HarrisonOg: they're from this Figma https://www.figma.com/file/ujdiReypSypu2GhRgzArea/Android-Components?node-id=1808%3A7597&t=bHV4Dx8xVSHy6Jv4-0

app/src/main/java/org/mozilla/fenix/compose/Chip.kt
```diff
@@ -0,0 +1,78 @@
...
+ * @param onClick Optional Callback for when the user taps this chip.
+ */
+@OptIn(ExperimentalMaterialApi::class)
+@Composable
+fun Chip(
+    text: String,
+    modifier: Modifier = Modifier,
+    textColor: Color = FirefoxTheme.colors.textActionPrimary,
+    backgroundColor: Color = FirefoxTheme.colors.actionPrimary,
+    enabled: Boolean = false,
```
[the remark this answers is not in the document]
HarrisonOg: ah yes I'll fix this

app/src/main/java/org/mozilla/fenix/compose/SelectableChip.kt
```diff
@@ -42,37 +35,28 @@ import org.mozilla.fenix.theme.FirefoxTheme
...
-                },
-            )
-            .padding(horizontal = 16.dp, vertical = 10.dp),
+    var selected by remember { mutableStateOf(isSelected) }
+
+    Chip(
+        enabled = true,
+        textColor = if (selected) selectedTextColor else unselectedTextColor,
+        backgroundColor = if (selected) selectedBackgroundColor else unselectedBackgroundColor,
+        text = text,
```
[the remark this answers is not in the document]
HarrisonOg: fixed

app/src/main/java/org/mozilla/fenix/compose/Chip.kt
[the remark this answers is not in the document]
MozillaNoah: Am I missing them in the figma? The link takes you to the Square Chip, but even then I only see Chips with/without a background. The Square Chip would be a separate component to implement.

app/src/main/java/org/mozilla/fenix/compose/Chip.kt
```diff
@@ -0,0 +1,78 @@
...
+ * Default layout for a clickable chip.
+ *
+ * @param text [String] displayed in this chip.
+ * @param modifier [Modifier] used to be applied to the layout of the chip.
+ * @param textColor Optional text [Color] for the chip.
+ * @param backgroundColor Optional background [Color] for the chip.
+ * @param enabled [Boolean] used to determine if chip responds to user input
+ * @param onClick Optional Callback for when the user taps this chip.
+ */
+@OptIn(ExperimentalMaterialApi::class)
```
MozillaNoah, line 30: what's the experimental API in this Composable?

app/src/main/java/org/mozilla/fenix/compose/SelectableChip.kt
```diff
@@ -8,71 +8,83 @@ import android.content.res.Configuration.UI_MODE_NIGHT_NO
...
  */
 @Composable
 fun SelectableChip(
     text: String,
     isSelected: Boolean,
-    selectedTextColor: Color? = null,
-    unselectedTextColor: Color? = null,
-    selectedBackgroundColor: Color? = null,
-    unselectedBackgroundColor: Color? = null,
+    colors: SelectableChipColors = SelectableChipColors.buildColors(),
```
MozillaNoah, line 35:
```suggestion
    selectableChipColors: SelectableChipColors = SelectableChipColors.buildColors(),
```

app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesComposables.kt
```diff
@@ -443,13 +444,17 @@ fun PocketStoriesCategories(
...
-                    unselectedTextColor = categoryColors.unselectedTextColor,
-                    selectedBackgroundColor = categoryColors.selectedBackgroundColor,
-                    unselectedBackgroundColor = categoryColors.unselectedBackgroundColor,
+                    isSelected = isSelected,
+                    colors = SelectableChipColors(
+                        selectedTextColor = categoryColors.selectedTextColor,
+                        unselectedTextColor =  categoryColors.unselectedTextColor,
+                        selectedBackgroundColor = categoryColors.selectedBackgroundColor,
+                        unselectedBackgroundColor = categoryColors.unselectedBackgroundColor,
+                    )
```
MozillaNoah, line 457: It might make more sense to break this instantiation outside of the loop and maybe above `Box`. Let's also use `SelectableChipColors.buildColors` in case  more colors are ever added to `SelectableChipColors` and they don't need to be forced downstream here.

app/src/main/java/org/mozilla/fenix/compose/Chip.kt
HarrisonOg: the ```androidx.compose.material.Chip``` is an experimental material api for the moment, might move out of experimental soon.

app/src/main/java/org/mozilla/fenix/compose/Chip.kt
```diff
@@ -0,0 +1,78 @@
...
+import org.mozilla.fenix.theme.FirefoxTheme
+
+/**
+ * Default layout for a clickable chip.
+ *
+ * @param text [String] displayed in this chip.
+ * @param modifier [Modifier] used to be applied to the layout of the chip.
+ * @param textColor Optional text [Color] for the chip.
+ * @param backgroundColor Optional background [Color] for the chip.
+ * @param enabled [Boolean] used to determine if chip responds to user input
```
gabrielluong, line 27: Let's move this one after `modifier` since it's the only non-optional parameter.

gabrielluong, line 27:
```suggestion
 * @param enabled [Boolean] used to determine if chip responds to user input.
```

app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesComposables.kt
```diff
@@ -443,13 +451,12 @@ fun PocketStoriesCategories(
             verticalItemsSpacing = 16.dp,
         ) {
             categories.filter { it.name != POCKET_STORIES_DEFAULT_CATEGORY_NAME }.forEach { category ->
+                val isSelected = selections.map { it.name }.contains(category.name)
```
gabrielluong, line 454: Maybe inline this since it's only ever used once, but this is fine too.

app/src/main/java/org/mozilla/fenix/compose/SelectableChip.kt
```diff
@@ -84,12 +99,15 @@ private fun SelectableChipPreview() {
...
-            SelectableChip(text = "Chirp", isSelected = true) { }
+            SelectableChip(text = "ChirpOne", isSelected = false) {}
+            SelectableChip(text = "ChirpTwo", isSelected = true) {}
         }
     }
 }
 
+/**
+ * An example of a [SelectableChip] with custom colors
+ */
```
gabrielluong, line 110: Let's avoid comments for @Preview instead let's use `name` in the [@Preview annotation](https://developer.android.com/jetpack/compose/tooling/studio#preview) to provide more context in the design view previews.

An example here would be `@Preview(name = "Custom Colors")`.

app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesComposables.kt
```diff
@@ -479,10 +486,10 @@ data class PocketStoriesCategoryColors(
...
         @Composable
         fun buildColors(
-            selectedBackgroundColor: Color = FirefoxTheme.colors.actionPrimary,
-            unselectedBackgroundColor: Color = FirefoxTheme.colors.actionTertiary,
-            selectedTextColor: Color = FirefoxTheme.colors.textActionPrimary,
-            unselectedTextColor: Color = FirefoxTheme.colors.textActionTertiary,
+            selectedBackgroundColor: Color = SelectableChipColors.buildColors().selectedBackgroundColor,
+            unselectedBackgroundColor: Color = SelectableChipColors.buildColors().unselectedBackgroundColor,
+            selectedTextColor: Color = SelectableChipColors.buildColors().selectedTextColor,
+            unselectedTextColor: Color = SelectableChipColors.buildColors().unselectedTextColor,
```
gabrielluong, line 492: Do we need `PocketStoriesCategoryColors` since this is really just a Selectable Chip now?

app/src/main/java/org/mozilla/fenix/compose/Chip.kt
```diff
@@ -0,0 +1,78 @@
...
+ * @param onClick Optional Callback for when the user taps this chip.
+ */
+@OptIn(ExperimentalMaterialApi::class)
+@Composable
+fun Chip(
```
gabrielluong, line 32: Looking to foster discussion and feedback on this comment. 

When it comes to writing generic composables (eg, [Button.kt](https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/java/org/mozilla/fenix/compose/button/Button.kt)) we are looking to expose generic components that have been proposed by the [UX Design Components](https://www.figma.com/file/ujdiReypSypu2GhRgzArea/Android-Components?node-id=1847%3A61820&t=5ruikDRwA1Tr97Aj-0) for public consumption . In this case, I am not sure we would want to expose the base `Chip` as `public` since we should strictly only expose components that we know that fits our component library and matches our design principles. 

To be clear, I don't necessarily want you to add left + right icons that you see in our Chip figma right now. We should only do that when we do have use cases for it, and it should be done outside this refactor either way.

MozillaNoah: I think this makes sense. We initially want two components in this file, essentially.
1. Chip (aka the root chip from which all other Chip Composables are derived.
2. TextChip (the first use case we need from the four currently possible chips designs - text-only, left icon, left + right, square)

gabrielluong: Would it also make sense for everything to also just go into a single file `Chip.kt`?

I don't know if we need to add the left + right icons use case since I don't recall ever seeing a Chip with icons. I could be wrong tho. I am mostly just a bit wary about adding something that we aren't actively using since that energy could be utilized somewhere else.

MozillaNoah: I think it could just live in `Chip.kt`.

As for the other use cases, I think those and square chip are definitely just a "just-in-time" implementation, since we don't see immediate usages of them.

app/src/main/java/org/mozilla/fenix/compose/Chip.kt
HarrisonOg: ah this is an optional parameter with a default value so I will change the description

app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesComposables.kt
HarrisonOg: I'll remove `PocketStoriesCategoryColors`

app/src/main/java/org/mozilla/fenix/compose/Chip.kt
HarrisonOg: ok sounds good to me, I made the changes in two separate commits. One moves `SelectableChip` into `Chip.kt`and the other removes `PocketStoriesCategoryColors`.
If we like the changes, I will do the git squashing and renaming so it becomes two commits: one for adding the `Chip` and `SelectableChip` and the other for removing `PocketStoriesCategoryColors`.

app/src/main/java/org/mozilla/fenix/compose/Chip.kt
```diff
@@ -0,0 +1,202 @@
...
+ * @param onClick Optional Callback for when the user taps this chip.
+ */
+@OptIn(ExperimentalMaterialApi::class)
+@Composable
+private fun Chip(
+    text: String,
+    modifier: Modifier = Modifier,
+    textColor: Color = FirefoxTheme.colors.textActionPrimary,
+    backgroundColor: Color = FirefoxTheme.colors.actionPrimary,
+    enabled: Boolean = true,
```
gabrielluong, line 43: Since we don't actually have a disabled state in the design system, I think it would make sense to remove this use case to make our components more opinionated and aligned with the system.

app/src/main/java/org/mozilla/fenix/compose/Chip.kt
```diff
@@ -0,0 +1,202 @@
...
+    onClick: () -> Unit,
```
gabrielluong, line 44: For the order of our parameters, we are guided by the following:
Mandatory (mandatory stateful parameters should still be ordered higher here compare to styles) -> optional modifier -> optional stateful parameter -> optional styles -> optional Compose lambdas

app/src/main/java/org/mozilla/fenix/compose/Chip.kt
```diff
@@ -0,0 +1,202 @@
...
+        enabled = enabled,
+        modifier = modifier,
+        colors = chipColors(
+            backgroundColor = backgroundColor,
+            contentColor = textColor,
+        ),
+    ) {
+        Text(
+            text = text,
+            style = FirefoxTheme.typography.body2,
```
gabrielluong, line 57: Just a nit to make sure we are following the order of the named parameters - so `style` follows after `color`

app/src/main/java/org/mozilla/fenix/compose/Chip.kt
```diff
@@ -0,0 +1,202 @@
...
+    modifier: Modifier = Modifier,
+    textColor: Color = FirefoxTheme.colors.textActionPrimary,
+    backgroundColor: Color = FirefoxTheme.colors.actionPrimary,
+    enabled: Boolean = true,
+    onClick: () -> Unit,
+) {
+    androidx.compose.material.Chip(
+        onClick = onClick,
+        enabled = enabled,
+        modifier = modifier,
```
gabrielluong, line 49: Let's double check if we are aligned with the designs.

app/src/main/java/org/mozilla/fenix/compose/Chip.kt
HarrisonOg: removed

app/src/main/java/org/mozilla/fenix/compose/Chip.kt
HarrisonOg: reordered

app/src/main/java/org/mozilla/fenix/compose/Chip.kt
HarrisonOg: fixed radius and set fixed height

app/src/main/java/org/mozilla/fenix/compose/Chip.kt
```diff
@@ -0,0 +1,193 @@
...
+    }
+}
+
+/**
+ * Default layout of a selectable chip.
+ *
+ * @param text [String] displayed in this chip.
+ * @param isSelected Whether this should be shown as selected.
+ * @param selectableChipColors The color set defined by [SelectableChipColors] used to style the chip.
+ * @param onClick Callback for when the user taps this.
```
MozillaNoah, line 69:
```suggestion
 * @param onClick Callback for when the user taps this chip.
```

app/src/main/java/org/mozilla/fenix/compose/Chip.kt
```diff
@@ -0,0 +1,196 @@
...
+    }
+}
+
+/**
+ * Wrapper for the color parameters of [SelectableChip].
+ *
+ * @param selectedTextColor Text [Color] when the chip is selected.
+ * @param unselectedTextColor Text [Color] when the chip is not selected.
+ * @param selectedBackgroundColor Background [Color] when the chip is selected.
+ * @param unselectedBackgroundColor Background [Color] when the chip is not selected.
```
MozillaNoah, line 100:
```suggestion
 * @param selectedBackgroundColor Background [Color] when the chip is selected.
 * @param unselectedBackgroundColor Background [Color] when the chip is not selected.
 * @param selectedTextColor Text [Color] when the chip is selected.
 * @param unselectedTextColor Text [Color] when the chip is not selected.
```
The param ordering for text/background are swapped.

app/src/main/java/org/mozilla/fenix/compose/Chip.kt
```diff
@@ -0,0 +1,196 @@
...
+            unselectedBackgroundColor = unselectedBackgroundColor,
+            selectedTextColor = selectedTextColor,
+            unselectedTextColor = unselectedTextColor,
+        )
+    }
+}
+
+@Composable
+@Preview(uiMode = Configuration.UI_MODE_NIGHT_YES)
+@Preview(uiMode = Configuration.UI_MODE_NIGHT_NO)
```
MozillaNoah, line 130: make sure to replace this with `LightDarkPreview` after that PR lands

app/src/main/java/org/mozilla/fenix/compose/Chip.kt
```diff
@@ -0,0 +1,196 @@
...
+ * @param modifier [Modifier] used to be applied to the layout of the chip.
+ * @param textColor Optional text [Color] for the chip.
+ * @param backgroundColor Optional background [Color] for the chip.
+ * @param onClick Callback for when the user taps this chip.
+ */
+@OptIn(ExperimentalMaterialApi::class)
+@Composable
+private fun Chip(
+    text: String,
+    modifier: Modifier = Modifier.clip(RoundedCornerShape(25.dp)),
```
MozillaNoah, line 42:
```suggestion
    modifier: Modifier = Modifier,
```
This should stay as `Modifier` (best practice but also a lint warning) and the `Shape` should be passed in via the `shape` parameter of the root `androidx.compose.material.Chip`

MozillaNoah (commented):
A few nits but this looks good otherwise

app/src/main/java/org/mozilla/fenix/compose/Chip.kt
HarrisonOg: fixed!

app/src/main/java/org/mozilla/fenix/compose/Chip.kt
HarrisonOg: fixed!

app/src/main/java/org/mozilla/fenix/compose/Chip.kt
HarrisonOg: I'll fix the commits after I include this change after the PR for `LightDarkPreview` lands

app/src/main/java/org/mozilla/fenix/compose/Chip.kt
```diff
@@ -0,0 +1,196 @@
...
+private fun Chip(
+    text: String,
+    modifier: Modifier = Modifier,
+    textColor: Color = FirefoxTheme.colors.textActionPrimary,
+    backgroundColor: Color = FirefoxTheme.colors.actionPrimary,
+    onClick: () -> Unit,
+) {
+    androidx.compose.material.Chip(
+        onClick = onClick,
+        modifier = modifier.clip(RoundedCornerShape(25.dp)),
```
MozillaNoah, line 49: `androidx.compose.material.Chip` provides a parameter, `chip` which takes a `Shape` as a parameter. Let's pass the `RoundedCornerShape` there.

MozillaNoah:
Looks like there's a couple failures in the CI. The lint should be easy but the the UI test failure look legit. If the fix is easy, go ahead and resolve it here. Otherwise, wrap it in an `@ignore` with a comment that has a link to an opened issue to track the disabled test so the automation team can be notified. See https://github.com/mozilla-mobile/fenix/issues/27714 for a quick example.

It looks like the `LightDarkPreview` still needs to be incorporated, and the Modifier.clip needs to be refactored out in favor of using androidx.compose.material.Chip's `shape` parameter before I can rubberstamp this.

HarrisonOg:
@MozillaNoah I tried to fix the UI test but I'm marking it as `@Ignore` for the time being and adding the new ticket for the test [Issue](https://github.com/mozilla-mobile/fenix/issues/28472).
I fixed the shape call too.

app/src/androidTest/java/org/mozilla/fenix/ui/HomeScreenTest.kt
```diff
@@ -299,6 +299,7 @@ class HomeScreenTest {
         }
     }
 
+    @Ignore("failing, see https://github.com/mozilla-mobile/fenix/issues/28472")
```
MozillaNoah, line 302:
```suggestion
    @Ignore("failing after a design refactor, see https://github.com/mozilla-mobile/fenix/issues/28472")
```

There's a nonzero chance we need to add a `testtag` parameter via `Modifier.semantics` to this as a follow-up. (See: `RecentlyVisited.RecentlyVisitedHistoryHighlight`)

app/src/main/java/org/mozilla/fenix/compose/Chip.kt
```diff
@@ -0,0 +1,197 @@
...
+    text: String,
+    modifier: Modifier = Modifier,
+    textColor: Color = FirefoxTheme.colors.textActionPrimary,
+    backgroundColor: Color = FirefoxTheme.colors.actionPrimary,
+    onClick: () -> Unit,
+) {
+    androidx.compose.material.Chip(
+        onClick = onClick,
+        modifier = modifier,
+        shape = MaterialTheme.shapes.small.copy(CornerSize(25.dp)),
```
MozillaNoah, line 51:
```suggestion
        shape = RoundedCornerShape(25.dp),
```

We're currently straying away from using the `MaterialTheme` library of shapes/colors/styles, so a simple `RoundedCornerShape` is sufficient here for now until we do a close examination of how to handle shapes in Fenix.

MozillaNoah (changes requested):
2 quick nits. I unfortunately also discovered that the original component used to style the Pocket categories was for iOS and neither the Chip nor square Chip in the design system match up. I'm going to start a conversation with UX to align on this. This has to be blocked until that's resolved 😞

app/src/main/java/org/mozilla/fenix/compose/Chip.kt
HarrisonOg: fixed

app/src/androidTest/java/org/mozilla/fenix/ui/HomeScreenTest.kt
HarrisonOg: added additional context

app/src/main/java/org/mozilla/fenix/compose/Chip.kt
```diff
@@ -0,0 +1,197 @@
...
+    text: String,
+    modifier: Modifier = Modifier,
+    textColor: Color = FirefoxTheme.colors.textActionPrimary,
+    backgroundColor: Color = FirefoxTheme.colors.actionPrimary,
+    onClick: () -> Unit,
+) {
+    androidx.compose.material.Chip(
+        onClick = onClick,
+        modifier = modifier,
+        shape = RoundedCornerShape(4.dp),
```
MozillaNoah, line 51: Rather than set the underlying default `Chip` to the Square Chip corner radius, let's instead add a boolean `isSquare` parameter (defaulting to false). Since we only have the two spec variations of a standard chip and a square chip, I'd lean towards this approach vs having a radius in the API. This also has the effect of keeping the both radius specs contained within the root `Chip`. Add the appropriate kdoc and preview and this should be good to go!

HarrisonOg: @MozillaNoah for the kdoc for the `Chip` param `isSquare`: "Optional [Boolean] to determine if chip is square chip or rounded chip."

MozillaNoah: That's a good pass. What if we just mention that it's shaping the chip's corners, rather than referencing the two specs we have in the design system currently?

`Optional [Boolean] to control whether the Chip's corners are square or rounded.`

HarrisonOg: fixed!

app/src/main/java/org/mozilla/fenix/compose/Chip.kt
```diff
@@ -0,0 +1,207 @@
...
+    textColor: Color = FirefoxTheme.colors.textActionPrimary,
+    onClick: () -> Unit,
+) {
+    androidx.compose.material.Chip(
+        onClick = onClick,
+        modifier = modifier,
+        shape = if (isSquare) RoundedCornerShape(4.dp) else RoundedCornerShape(25.dp),
+        colors = chipColors(
+            backgroundColor = backgroundColor,
+            contentColor = textColor,
```
MozillaNoah, line 55: why is `contentColor` being set here?

app/src/main/java/org/mozilla/fenix/compose/Chip.kt
```diff
@@ -0,0 +1,207 @@
...
+            SelectableChip(
+                text = "Yellow",
+                isSelected = false,
+                selectableChipColors = SelectableChipColors(
+                    selectedBackgroundColor = Color.Yellow,
+                    unselectedBackgroundColor = Color.DarkGray,
+                    selectedTextColor = Color.Black,
+                    unselectedTextColor = Color.Gray,
+                ),
+            ) {}
```
MozillaNoah, line 194: nit: add a new line between the selectable chips in these two previews

app/src/main/java/org/mozilla/fenix/compose/Chip.kt
HarrisonOg: oh good catch, removed!

app/src/main/java/org/mozilla/fenix/compose/Chip.kt
```diff
@@ -0,0 +1,207 @@
...
+ * @param onClick Callback for when the user taps this chip.
+ */
+@OptIn(ExperimentalMaterialApi::class)
+@Composable
+private fun Chip(
+    text: String,
+    modifier: Modifier = Modifier,
+    backgroundColor: Color = FirefoxTheme.colors.actionPrimary,
+    isSquare: Boolean = false,
+    textColor: Color = FirefoxTheme.colors.textActionPrimary,
```
MozillaNoah, line 46:
```suggestion
    isSquare: Boolean = false,
    backgroundColor: Color = FirefoxTheme.colors.actionPrimary,
    textColor: Color = FirefoxTheme.colors.textActionPrimary,
```

re order the params/kdocs so the colors are bundled together

MozillaNoah (approved):
One last nit, but this looks good to me! 🎉

Issue #27632, opened 2022-10-31
Selected branch in the Nimbus Experiments settings does not have a check mark


cpeterso:
### Steps to reproduce

1. Enable Fenix's debug mode:
a. Open the Fenix settings
b. Open "About Firefox".
c. Tap the Firefox logo four times.
2. Open the "Nimbus experiments" settings menu.
3. Select any experiment.
4. Tap one of the experiment's branch names, such as `control` or `treatment-a`.

### Expected behaviour

A check mark should be next to the selected branch name. Here is a screenshot of the expected behavior on @Mugurell's phone:

![image](https://user-images.githubusercontent.com/629193/199083424-c705b0f7-52ba-4805-8324-69420a16665a.png)

### Actual behaviour

There is no check mark next to the selected branch name. Here is a screenshot of a Nimbus experiment's branches. No matter how many times I tap or close and re-open the settings, there is no check mark.

![image](https://user-images.githubusercontent.com/629193/199083830-ebc9db5f-fb8b-4dce-8749-e6da7daef907.png)

I can reproduce this with any Nimbus experiment in all the Fenix release channels: Release 106, Beta 107, and Nightly 108.

### Device name

Samsung Galaxy A32

### Android version

Android 12

### Firefox release type

Firefox

### Firefox version

106

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21903)

cpeterso:
Moved to bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1813430

Change performed by the [Move to Bugzilla add-on](https://addons.mozilla.org/en-US/firefox/addon/move-to-bugzilla/).

Issue #19534, opened 2021-05-17
Check mark barely visible on Nimbus Experiments branch selection (when on Light Theme)


amedyne:
[comment]: # (Please do your best to search for duplicate issues before filing a new issue so we can keep our issue board clean)
[comment]: # (Every issue should have the exact bug and steps to reproduce described in it. Please do not file feedback list tickets as it is difficult to parse them and address their individual points)
[comment]: # (Read https://github.com/mozilla-mobile/fenix#i-want-to-file-an-issue for more information)

## Steps to reproduce
- Customize and set "Theme" to "Light" (go to Settings-Customize-Theme)
- Enable debug Menu
- Go to Nimbus Experiments
- Select an experiment

### Expected behavior
When a user selects a branch, check mark is visible/clear

### Actual behavior
- User can barely see the check mark for selected experiment branch. Note that in Dark Theme the check mark is visible
![Screenshot_20210513-105256](https://user-images.githubusercontent.com/68711330/118509362-f5b20d80-b6fd-11eb-9426-c589759ce45b.png)

- 
### Device information

* Device vendor / model and Android version: Pixel 3XL
* Firefox for Android version: ? (go to Settings -> About Firefox) Firefox Nightly  
90.0a1 (Build #2015810763)
AC: 90.0.20210516143141, be02514d7
GV: 90.0a1-20210516091748
AS: 76.0.0

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-10714)

SoftVision-LorandJanos:
This issue is still reproducible on the latest Nightly 96.0a1 (2021-11-18).
Device used: Xiaomi Mi11 Lite (Android 11).

SoftVision-LorandJanos:
This issue is still reproducible with the latest Nightly 102.0a1 (2022-05-18) build.
Device used: Google Pixel 4 (Android 12).

SoftVision-LorandJanos:
This issue is still reproducible on the latest Nightly 110.0a1 (2023-01-17).
Device used: Oppo Find X5 (Android 12).
Reopening the ticket.

boek:
Moved to bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1814138

Change performed by the [Move to Bugzilla add-on](https://addons.mozilla.org/en-US/firefox/addon/move-to-bugzilla/).

PR #27640, opened 2022-11-01
Converting Nimbus Branches and Nimbus Fragments to Compose.

vrrashkov:
Based on this issue #[19534](https://github.com/mozilla-mobile/fenix/issues/19534) and I believe this might also be the same one #[27632](https://github.com/mozilla-mobile/fenix/issues/27632). It was design related but instead of fixing only the color I migrated the code to compose for Nimbus Branches and Nimbus Experiments.

Experiments:
image
image

Branches:
image
image

The logic is kept the same, there are no changes to the BranchesStore/BranchesController. That's why there are no new tests and the old ones are working.

This mozilla.components.service.nimbus this module is no longer used. The new elements for the UI are all inside the BranchesFragment and ExperimentsFragment.  I was not sure if I should put them in the mozilla.components.service.nimbus module, so I checked other places and saw that "WallpaperSettings" for example is using compose element directly in the Fragment and not in a separate module so I did the same.


gabrielluong:
Hey @vrrashkov, this looks like a very good first pass at converting the Nimbus view into Compose. Thank you for the contribution.

I have various review feedback that ranges from formatting, naming, etc that would help align the code closer to what we would expect from each other in the team. I am wondering if you would prefer to receive such feedback and address them. I know this could be cumbersome depending on your other commitments. So, I can either offer the feedback if you prefer otherwise I could also help make the adjustments to get your work through the finish line. 

Thanks!

vrrashkov:
Yes sounds good. I do plan to do more contributions in the future, so I am open to feedback.

PR #27647, opened 2022-11-01
Closes #25809: tapping on search engine triggers engine selection

### LINKED ISSUE #25809: Tapping on search engine icon should not enter input mode
When user tap on the search engine icon in the toolbar, we should allow search engine selection menu popup without entering input mode. (No virtual keyboard should show)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-20847)

mavduevskiy:
Adding search selector to the toolbar in `Display mode`. Flow is well described [here](https://www.figma.com/file/epu0Sj6mCPk7K282sNpHgI/Site?node-id=3851%3A77540#296011078).

Old|New
---|---
unified_on_old|unified_on_new
unified_off_old|unified_off_new

There is also a spacing issue that is particularly visible on devices that have arrow down keyboard button.
https://user-images.githubusercontent.com/92760693/199746717-05e0c70e-ce85-4056-830f-e8342fb3e0e6.mp4

I could make it consistent by adding more spacing to the display mode , but not sure if I should. An easy fix would look like this. 
https://user-images.githubusercontent.com/92760693/199754394-c09b423e-b36e-4efe-812d-065be20850d0.mp4

Will take a look if it's possible to fix the padding in the edit mode (so that an icon without selector wouldn't take extra space), as a follow up.

### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation
Fixes #25809


app/src/main/java/org/mozilla/fenix/search/SearchFragmentStore.kt
```diff
@@ -117,11 +117,13 @@ data class SearchFragmentState(
 /**
  * Creates the initial state for the search fragment.
  */
+@Suppress("LongParameterList")
 fun createInitialSearchFragmentState(
     activity: HomeActivity,
     components: Components,
     tabId: String?,
     pastedText: String?,
+    selectedFromMenuEngine: SearchEngine?,
```
gabrielluong, line 126:
```suggestion
    searchEngine: SearchEngine?,
```
Could we also reorder this to be below `searchAccessPoint` to align with the param ordering you did above in `SearchDialogFragment`

app/src/main/java/org/mozilla/fenix/search/toolbar/SearchSelectorMenu.kt
```diff
@@ -65,3 +66,16 @@ class SearchSelectorMenu(
         )
     }
 }
+
+/**
+ * Interface for search selector action.
+ */
+interface SearchSelectorInteractor {
```
gabrielluong, line 73: Let's move this to its own file

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -330,6 +347,28 @@ class HomeFragment : Fragment() {
...
+                }
+
+                UnifiedSearch.searchMenuTapped.record(mozilla.telemetry.glean.private.NoExtras())
+                searchSelectorMenu.menuController.show(anchor = it, orientation = orientation, forceOrientation = true)
+            }
+
+            setBackgroundResource(
+                context.theme.resolveAttribute(android.R.attr.selectableItemBackgroundBorderless),
+            )
+        }
```
gabrielluong, line 370: I have an interest to see more code refactored out of the `HomeFragment` and into their corresponding classes so that we could encapsulate all the logic around the search selector into the appropriate classes. Would this be possible here? Would it make sense add a public update function of some sort in `SearchSelectorMenu` or something else that can be called here?

gabrielluong: Actually, this might fit more with the `Builder` classes we did for the `HomeFragment`. See `TabCounterBuilder` and `HomeMenuBuilder`.

gabrielluong: We don't necessarily have to do this in this PR. This is more food for thought if we were able to encapsulate the logic here better so that our enormous `HomeFragment` won't hold so much of the logic related to various functions.

app/src/main/java/org/mozilla/fenix/search/toolbar/SearchSelectorMenu.kt
mavduevskiy: Could you explain a bit more? I see a lot of classes (that relate to the UI, like views and delegates) that take interactor as a parameter declaring the interface inside the file. Well, even this one was chipped away from `ToolbarInteractor`, which is the same file with `ToolbarView`.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
mavduevskiy: Yes, I also think that it has to be encapsulated. Actually, it is – inside `SearchSelectorToolbarAction`, which is a `Toolbar.Action` and can't be used on a "fake" toolbar we have on the homescreen. Creating a more real toolbar for the `HomeFragment` could be a direction to go? I have little context why it was implemented that way.

app/src/main/java/org/mozilla/fenix/search/toolbar/SearchSelectorMenu.kt
gabrielluong: This is more about moving the Interactor interface into its own file and keeping them out of the views. We have found that there will be times where the Interactor interface could be reused in a variety of places so it would make more sense to keep them to their own file from the get go.

I would imagine you are referring to Bookmark and History views for what you're seeing, but those are also some of our oldest views that we don't actively polish compare to the Home screen. Other than that I would say we would rather keep files smaller for better development ergonomics if there's a combination of classes, functions and interfaces that are all in one place. I could also imagine the views are more volatile to changes and deletion compare to its own separate file that holds the interactor interface.

Ultimately, this is just about developer ergonomics. It's easier to work with smaller files and you aren't at risk of refactoring in the future.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
gabrielluong: When I wrote `SearchSelectorToolbarAction`, it was for encapsulating the `Toolbar.Action` functionality that would be added to the edit toolbar actions. This is indeed a direction I am trying to get at, but I am getting the sense now that we should probably be a bit less concern about the ideas I presented in this conversation for the sake of this PR and we can revisit this more in a future cleanup of `HomeFragment`. We'll want to think about what makes sense to group up and where (should it be be done based on the component in the `binding`, etc). 

We can resolve this conversation.

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -602,20 +649,47 @@ class HomeFragment : Fragment() {
...
-                    if (searchEngine != null) {
-                        val iconSize =
-                            requireContext().resources.getDimensionPixelSize(R.dimen.preference_icon_drawable_size)
-                        val searchIcon =
-                            BitmapDrawable(requireContext().resources, searchEngine.icon)
-                        searchIcon.setBounds(0, 0, iconSize, iconSize)
-                        binding.searchEngineIcon.setImageDrawable(searchIcon)
-                    } else {
-                        binding.searchEngineIcon.setImageDrawable(null)
+                    searchEngine?.let {
```
gabrielluong, line 673: I think we should stick to using a simple if statement check for `searchEngine != null` here instead of the `let` follow by the `run`

mavduevskiy:
At the same time, [setBounds](https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt#615) method does not have any effect on the `BitmapDrawable` size. It keeps the Bitmap the initial size, and `BitmapDrawable` takes its parameters. 

We don't notice it with the icon in the `HomeFragment` because of the default scaling done by `ImageView`. But it has a visible effect on the search selector, as it uses specified scaling: `android:scaleType="center"`. It creates extra padding when is fitting a bigger icon into a smaller frame. 

The [extension](https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/java/org/mozilla/fenix/search/toolbar/SearchSelectorToolbarAction.kt#90) actually works, because it creates a smaller Bitmap. But that disables the [test](https://searchfox.org/mozilla-mobile/source/fenix/app/src/androidTest/java/org/mozilla/fenix/ui/SettingsSearchTest.kt#74) that verifies bitmap in the `ImageView` matches the on of the search engine. 

I am removing not working `setBounds` code from the `HomeFragment` and scaling from the `search_selector`. It helps to keep the test and looks safe.

As a side note, why would  `search_selector` has a special scaling if it receives already scaled  by the extension `BitmapDrawable`? I might be missing something.

app/src/main/java/org/mozilla/fenix/search/toolbar/SearchSelectorMenu.kt
```diff
@@ -26,12 +26,13 @@ typealias MozSearchEngine = SearchEngine
...
 class SearchSelectorMenu(
     private val context: Context,
-    private val interactor: ToolbarInteractor,
+    private val interactor: SearchSelectorInteractor,
 ) {
 
     /**
      * Items that will appear in the search selector menu.
      */
+
```
gabrielluong, line 35: nit no new line necessary

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -330,6 +346,28 @@ class HomeFragment : Fragment() {
...
+
+        binding.searchSelector.apply {
+            setOnClickListener {
+                val orientation = if (context.settings().shouldUseBottomToolbar) {
+                    Orientation.UP
+                } else {
+                    Orientation.DOWN
+                }
+
+                UnifiedSearch.searchMenuTapped.record(mozilla.telemetry.glean.private.NoExtras())
```
MozillaNoah, line 362:
```suggestion
                UnifiedSearch.searchMenuTapped.record(NoExtras())
```

Can we clean up the full import here?

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -330,6 +346,28 @@ class HomeFragment : Fragment() {
...
+                searchSelectorMenu.menuController.show(anchor = it, orientation = orientation, forceOrientation = true)
+            }
+
+            setBackgroundResource(
+                context.theme.resolveAttribute(android.R.attr.selectableItemBackgroundBorderless),
+            )
```
MozillaNoah, line 368: should this just live in the XML?

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlController.kt
```diff
@@ -660,4 +668,26 @@ class DefaultSessionControlController(
...
+
+    override fun handleMenuItemTapped(item: SearchSelectorMenu.Item) {
+        when (item) {
+            SearchSelectorMenu.Item.SearchSettings -> {
+                navController.nav(
+                    R.id.homeFragment,
+                    HomeFragmentDirections.actionGlobalSearchEngineFragment(),
+                )
+            }
+            is SearchSelectorMenu.Item.SearchEngine -> {
```
MozillaNoah, line 680:
```suggestion
            SearchSelectorMenu.Item.SearchEngine -> {
```

MozillaNoah (commented):
🚗 driving by with a couple comments

app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlController.kt
mavduevskiy: It does look odd, but `SearchEngine` is a data class, so won't work without `is`

mavduevskiy:
@MozillaNoah @gabrielluong 
Thanks for taking your time to review! updated

app/src/main/java/org/mozilla/fenix/search/toolbar/SearchSelector.kt
```diff
@@ -21,9 +22,21 @@ internal class SearchSelector @JvmOverloads constructor(
 ) : RelativeLayout(context, attrs, defStyle) {
 
     private val binding = SearchSelectorBinding.inflate(LayoutInflater.from(context), this)
+    private var marginTop: Int = 0
```
gabrielluong, line 25: I am curious what kind of issues we ran into that required us to programmatically set the margin in the class. 
Could we also set the margin in the XML in `fragment_home` to avoid adding the code here?

app/src/main/res/layout/fragment_home.xml
```diff
@@ -140,16 +140,23 @@
                 android:layout_height="24dp"
                 android:layout_gravity="start|center_vertical"
                 android:layout_marginStart="8dp"
+                android:layout_marginEnd="12dp"
                 android:clickable="false"
                 android:focusable="false"
-                android:importantForAccessibility="no" />
+                android:importantForAccessibility="no"/>
+
+            <org.mozilla.fenix.search.toolbar.SearchSelector
```
gabrielluong, line 148: Could we set the marginTop here?

app/src/main/res/layout/fragment_home.xml
```diff
@@ -140,16 +140,23 @@
                 android:layout_height="24dp"
                 android:layout_gravity="start|center_vertical"
                 android:layout_marginStart="8dp"
+                android:layout_marginEnd="12dp"
                 android:clickable="false"
                 android:focusable="false"
-                android:importantForAccessibility="no" />
+                android:importantForAccessibility="no"/>
```
gabrielluong, line 146: Could we undo the spacing change here?

app/src/main/res/layout/search_selector.xml
```diff
@@ -30,7 +29,6 @@
                 android:id="@+id/icon"
                 android:layout_width="24dp"
                 android:layout_height="24dp"
-                android:scaleType="center"
```
gabrielluong, line 33: Trying to understand this change. Could you explain what kind of image scale issues you were seeing here and what was the mitigation?

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -602,20 +648,34 @@ class HomeFragment : Fragment() {
...
             flow.map { state -> state.search.selectedOrDefaultSearchEngine }
                 .ifChanged()
                 .collect { searchEngine ->
-                    if (searchEngine != null) {
-                        val iconSize =
-                            requireContext().resources.getDimensionPixelSize(R.dimen.preference_icon_drawable_size)
-                        val searchIcon =
-                            BitmapDrawable(requireContext().resources, searchEngine.icon)
-                        searchIcon.setBounds(0, 0, iconSize, iconSize)
-                        binding.searchEngineIcon.setImageDrawable(searchIcon)
```
gabrielluong, line 616: Trying to understand if this change was intentional. I see we aren't setting the bounds anymore, but I am wondering what is the reasoning behind the change? I see we do call this type `setBounds(0, 0, iconSize, iconSize)` in many places in our codebase https://searchfox.org/mozilla-mobile/search?q=setBounds&path=, but I don't have any context on its significance or usage. Since we are removing it here, I would like to learn a bit more what types of problem we were encountering and solving.

app/src/main/java/org/mozilla/fenix/search/toolbar/SearchSelector.kt
mavduevskiy: `SearchSelector` has a hardcoded top margin, that made it placing in alongside the [searchIcon](https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/res/layout/fragment_home.xml#138) difficult.

app/src/main/res/layout/fragment_home.xml
mavduevskiy: It won't affect the top margin of the material card inside the `[search_selector](https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/res/layout/search_selector.xml#15)`

app/src/main/res/layout/search_selector.xml
mavduevskiy: I tried to explain it [here](https://github.com/mozilla-mobile/fenix/pull/27647#issuecomment-1301656042).

In short, this scaling type would create extra margin for icons bigger then the view size. It can be seen with duck duck go icon, if one revers the change. Default scaling already works correctly for [searchIcon](https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/res/layout/fragment_home.xml#138), so I took that approach

app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
mavduevskiy: yes, I was talking about it [here](https://github.com/mozilla-mobile/fenix/pull/27647#issuecomment-1301656042).

short version, this particular code doesn't do anything, since we create a `BitmapDrawable` from a `Bitmap`. Changing it's bounds afterwards doesn't affect the Bitmap itself, so it's height and width will be used.
You can see that by changing `preference_icon_drawable_size` to 2 or 200 – it won't have any effect.

app/src/main/java/org/mozilla/fenix/search/toolbar/SearchSelector.kt
gabrielluong: This was a random lunch thought, but I haven't really looked into the code again with consideration of your comment yet. 

I am thinking in an ideal situation all of us would probably prefer to see all the layout properties defined in the XML layout. If another engineer was working on the search selector layout, they might be confused by where the top margin is coming from in their testing because they don't see the margin top 14dp defined in the XML. They would have to dig much deeper into the code in order to figure out that the margin is set manually in the code and that could be both surprising and maybe somewhat frustrating when working with this code. 

So, I am wondering if we could do some more exploring to see if we could define this margin in the XML. A naive idea I have would be to try `marginTop=-14dp` and see if that works? I don't necessarily have the answers right now, but I will take some time to play around with your PR to see if there is a reasonable solution when I have a chance. 

I am also happy to accept this as a followup investigation and it is possible that the only solution is to set the margin programatically. I just want to make sure we put some more thought into this since this will only improve how people work with this code.

app/src/main/res/layout/search_selector.xml
mavduevskiy: here is the [patch](https://github.com/mozilla-mobile/fenix/pull/27726) with reverted scaling code. with unified search is on, duckduck, amazon, google icons will show the problem! one could also play around with `preference_icon_drawable_size` parameter to see if it has any effect.

gabrielluong:
I haven't gotten around to the review because I am still looking to see if you had any response to 2 questions:
(1) Could we somehow set the margin setting in the XML instead of in the code? This is mostly to avoid surprises in terms of developer ergonomics when one is looking at the search selector layout properties in the XML and only learns that the margin top is set programmatic much later. https://github.com/mozilla-mobile/fenix/pull/27647#discussion_r1014374484
(2) I haven't had a chance to test and verify that `setBound` does nothing in relation to https://github.com/mozilla-mobile/fenix/pull/27647#discussion_r1013584431. I am mostly just suspicious of this change because this is the only line of code that will ship live because it is not behind a feature flag. It appears to be used commonly elsewhere as I have pointed out and it was added by pocmo by in the day so I have to imagine it was done for some purpose even though I do not yet understand why. So, my main ask here is to test this line of code a bit more and ensure we're still getting the expected results even in older Android versions. This means testing in Android 6 as stated in https://github.com/mozilla-mobile/fenix/pull/27726#issuecomment-1305120483.

mavduevskiy:
@gabrielluong 
I created two follow ups, not to block this PR. 
1. [Investigate](https://github.com/mozilla-mobile/fenix/issues/27763) the top margin setting through xml.
2. [Investigate](https://github.com/mozilla-mobile/fenix/issues/27764) (apparently) not working scaling code in the HomeFragment.
 
Btw, even when approved, I don't think we should land it now. Next week merging into Beta looks like a safer option.

gabrielluong:
> @gabrielluong I created two follow ups, not to block this PR. Reverted `setBounds` code, refactored.
> 
> 1. [Investigate](https://github.com/mozilla-mobile/fenix/issues/27763) the top margin setting through xml.
> 2. [Investigate](https://github.com/mozilla-mobile/fenix/issues/27764) (apparently) not working scaling code in the HomeFragment.
> 
> Btw, even when approved, I don't think we should land it now. Next week merging into Beta looks like a safer option.

Okay, I see that we reverted the `setBounds` code and will investigate it as a followup and that we will land this next week. I don't have concerns here. I will leave it to Roger to test and evaluate. Thank you for following up on the changes and investigation.

mavduevskiy:
@gabrielluong 
Thank you for being so diligent. Looking back, the idea of removing code that looks like it is not doing anything as an act of random refactoring that is not really contributing to the task, does not look like a good one. It might not do anything on my device/OS/manufacturer, but it doesn't mean it won't break anything if we remove it. 

If I had good understanding why it was written, why we don't need it now and if it's safe to remove – that's another story. But we need more investigation, hence the follow up task. Lesson learnt👍

rocketsroger (approved):
Looks good to me, but let's land this after the cut.  Thanks

PR #27650, opened 2022-11-01
For #26320: Reorganize Nimbus Startup

### LINKED ISSUE #26320: Nimbus initialization should be triggered and complete before first page load starts
This is a follow-up and summary of discussions between @jonalmeida, @jhugman and myself when working through https://github.com/mozilla-mobile/fenix/pull/26228. 

We want to improve and simplify Nimbus startup and application initialization ordering.

**Problem**
Nimbus initialization is currently started after engine initialization, and carried out asynchronously. This is by design to limit its impact on startup performance. However, the consequence is a race condition when an experiment and feature configuration is targeted at the engine, e.g., a tracking protection or site isolation setting that's controlled by a Nimbus feature. This means that the first page load after cold start may be carried out with the default settings (the feature's default value), as opposed to the values based on the experiment branch, if Nimbus initialization has not completed. Since cold starts are very common on mobile, and users tend to have short sessions opening only a single link before switching back to another application, this can (more significantly) impact experiment results.

Another problem with late Nimbus initialization is that it's hard to reason about the state of the application after startup. To compensate for the fact that Nimbus initialization may complete at a later point in time, we currently update the corresponding features when Nimbus is ready. We use `onUpdatesApplied` for this, which will eventually become a large subset of seemingly unrelated settings and feature updates, and ultimately a maintenance burden. While this could be addressed with separate observers, we think the following proposal could solve both problems.

**Proposal**
To prevent this entire set of problems, we propose changing initialization ordering so that Nimbus initializes and completes<sup>1</sup> right after `Glean`, but before everything else including the `engine`. As a side note, engine initialization is currently accidentally triggered by [initializeGlean](https://github.com/mozilla-mobile/fenix/blob/c356982a4decc7a6a2cf92f3b67579c9cac3bf33/app/src/main/java/org/mozilla/fenix/FenixApplication.kt#L160) when the store is accessed for startup metrics. This should be fixed as well as part of this, and rely on the explicit [warmup](https://github.com/mozilla-mobile/fenix/blob/c356982a4decc7a6a2cf92f3b67579c9cac3bf33/app/src/main/java/org/mozilla/fenix/FenixApplication.kt#L188) call again.

<sup>1</sup> This does not include waiting for the latest network fetch, just the local initialization and reading from storage.

**Performance Impact**
Delaying engine startup until Nimbus is complete will have a performance impact we need to measure and land telemetry for. We need to understand the startup cost of Nimbus. This is also required to understand the severity of the current race condition, i.e., how many users are affected from the first page load happening before Nimbus is ready. (Even if that impact is low, it's still worth fixing this to gain simplicity and deterministic initialization ordering.)

I think we should accept **some** cost to startup time for gaining simplicity and stability with this change.

**Known issues**
We built a quick prototype that confirmed this can work, but noticed that there's a risk for a cyclical dependency and infinite loop, if Nimbus uses the engine during initialization. This will happen because of its use of our `concept.fetch` [client](https://github.com/mozilla-mobile/fenix/blob/c356982a4decc7a6a2cf92f3b67579c9cac3bf33/app/src/main/java/org/mozilla/fenix/FenixApplication.kt#L409), or any accidental reference of the `engine` in [createNimbus](https://github.com/mozilla-mobile/fenix/blob/c356982a4decc7a6a2cf92f3b67579c9cac3bf33/app/src/main/java/org/mozilla/fenix/experiments/NimbusSetup.kt#L38) such as in [strictMode.resetAfter](https://github.com/mozilla-mobile/fenix/blob/c356982a4decc7a6a2cf92f3b67579c9cac3bf33/app/src/main/java/org/mozilla/fenix/experiments/NimbusSetup.kt#L66). It's important we investigate a way to guard against this, e.g., by making sure Nimbus doesn't use the `engine` at all.

We will use this ticket to prototype and likely turn it into a meta, as work will be required in both Nimbus and Fenix/A-C to complete this. This should also serve as a guideline and solution for Focus.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21200)

jhugman:
Fixes [EXP-2891](https://mozilla-hub.atlassian.net/browse/EXP-2891)
Fixes #26320.

### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation
Fixes #26320

### DIFF
--- a/app/src/androidTest/java/org/mozilla/fenix/perf/StartupExcessiveResourceUseTest.kt
+++ b/app/src/androidTest/java/org/mozilla/fenix/perf/StartupExcessiveResourceUseTest.kt
@@ -33,7 +33,7 @@ import org.mozilla.fenix.helpers.HomeActivityTestRule
  *
  * Say no to main thread IO! 🙅
  */
-private const val EXPECTED_SUPPRESSION_COUNT = 19
+private const val EXPECTED_SUPPRESSION_COUNT = 18
 
 /**
  * The number of times we call the `runBlocking` coroutine method on the main thread during this
--- a/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
+++ b/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
@@ -50,7 +50,6 @@ import mozilla.components.service.glean.net.ConceptFetchHttpUploader
 import mozilla.components.support.base.facts.register
 import mozilla.components.support.base.log.Log
 import mozilla.components.support.base.log.logger.Logger
-import mozilla.components.support.base.observer.Observable
 import mozilla.components.support.ktx.android.content.isMainProcess
 import mozilla.components.support.ktx.android.content.runOnlyInMainProcess
 import mozilla.components.support.locale.LocaleAwareApplication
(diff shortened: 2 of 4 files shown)


jonalmeida (approved):
Christian and I tested this again to ensure the ordering is consistent regardless of whether there is an engine experiment or not.

PR #27660, opened 2022-11-02
Converting Nimbus Branches and Nimbus Fragments to Compose

vrrashkov:
Based on this issue #[19534](https://github.com/mozilla-mobile/fenix/issues/19534) and I believe this might also be the same one #[27632](https://github.com/mozilla-mobile/fenix/issues/27632). It was design related but instead of fixing only the color I migrated the code to compose for Nimbus Branches and Nimbus Experiments.

Experiments:
image
image

Branches:
image
image

The logic is kept the same, there are no changes to the BranchesStore/BranchesController. That's why there are no new tests and the old ones are working.

This mozilla.components.service.nimbus this module is no longer used. The new elements for the UI are all inside the BranchesFragment and ExperimentsFragment.  I was not sure if I should put them in the mozilla.components.service.nimbus module, so I checked other places and saw that "WallpaperSettings" for example is using compose element directly in the Fragment and not in a separate module so I did the same.

@gabrielluong Sorry had an issue with my repo so I closed the [previous PR](https://github.com/mozilla-mobile/fenix/pull/27640) and created a clean one.


app/src/main/java/org/mozilla/fenix/nimbus/NimbusBranchesFragment.kt
```diff
@@ -30,65 +52,145 @@ import org.mozilla.fenix.nimbus.view.NimbusBranchesView
 class NimbusBranchesFragment : Fragment() {
 
     private lateinit var nimbusBranchesStore: NimbusBranchesStore
-    private lateinit var nimbusBranchesView: NimbusBranchesView
-    private lateinit var controller: NimbusBranchesController
 
     private val args by navArgs<NimbusBranchesFragmentArgs>()
 
+    override fun onResume() {
```
gabrielluong, line 58: Could you undo the move of `onResume` and `loadExperimentBranches`? Since both these functions didn't actually change and were only moved above `onCreateView`, we would prefer to minimize the amount of changes in the PR.

app/src/main/java/org/mozilla/fenix/nimbus/NimbusBranchesFragment.kt
```diff
@@ -30,65 +52,145 @@ import org.mozilla.fenix.nimbus.view.NimbusBranchesView
...
-                val branches = experiments.getExperimentBranches(args.experimentId) ?: emptyList()
-                val selectedBranch = experiments.getExperimentBranch(args.experimentId) ?: ""
+    /**
+     * The list for the Nimbus Branches,
+     *
+     * @param branchesList List of branches
+     * @param onSelectedBranch Callback for when item is selected.
+     */
+    @Composable
+    fun NimbusBranches(
```
gabrielluong, line 132: Let's refactor the Composables to a new file called `NimbusBranches`

app/src/main/java/org/mozilla/fenix/nimbus/NimbusExperimentsFragment.kt
```diff
@@ -5,81 +5,119 @@
...
 @Suppress("TooGenericExceptionCaught")
-class NimbusExperimentsFragment : Fragment(R.layout.mozac_service_nimbus_experiments) {
-
-    private var adapter: NimbusExperimentAdapter? = null
-
-    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
-        super.onViewCreated(view, savedInstanceState)
-        bindRecyclerView(view)
-    }
+class NimbusExperimentsFragment : Fragment() {
```
gabrielluong, line 34: Let's do 2 separate PRs - 1 for the Nimbus Branches and 1 for the Nimbus Experiments. I would prefer if we focused on converting Experiments first. Please don't do both at the same time for easier reviews and let's take a more iterative approach here. I do not believe in future iteration we would need to have 2 different fragments because we should be able just add a `NavHost` and navigate between the 2 different screens within the same fragment. See https://developer.android.com/jetpack/compose/navigation

app/src/main/java/org/mozilla/fenix/nimbus/NimbusExperimentsFragment.kt
```diff
@@ -5,81 +5,119 @@
...
+    }
+
+    /**
+     * The list for the Nimbus Experiments,
+     *
+     * @param availableExperiments List of available items
+     * @param onSelectedExperiment Callback for when item is selected.
+     */
+    @Composable
+    fun NimbusExperiments(
```
gabrielluong, line 75: Let's move the Composables to a separate file called `NimbusExperiments`

app/src/main/java/org/mozilla/fenix/nimbus/NimbusExperimentsFragment.kt
```diff
@@ -5,81 +5,119 @@
...
+                )
+            }
+        }
+    }
+
+    /**
+     * The list for the Nimbus Experiments,
+     *
+     * @param availableExperiments List of available items
+     * @param onSelectedExperiment Callback for when item is selected.
```
gabrielluong, line 72: I will have further suggestions for improving the kDoc for your Composable. However, since I am already asking for a lot of changes from the get go, I would prefer if you could take some time to look at our existing @Composable and try to emulate the same style of how we write our kDoc.

app/src/main/java/org/mozilla/fenix/nimbus/NimbusExperimentsFragment.kt
```diff
@@ -5,81 +5,119 @@
...
+            }
+        }
+    }
+    /**
+        * Nimbus experiment item
+        *
+        * @param item current experiment
+    */
+    @Composable
+    fun NimbusExperimentItem(
```
gabrielluong, line 94: We have a `ListItem.kt` that has a variety of list item components designed by our Design System team. I think it makes sense to reuse this component for the experiment items.

app/src/main/java/org/mozilla/fenix/nimbus/NimbusExperimentsFragment.kt
```diff
@@ -5,81 +5,119 @@
 package org.mozilla.fenix.nimbus
 
 import android.os.Bundle
+import android.view.LayoutInflater
 import android.view.View
-import android.widget.TextView
-import androidx.core.view.isVisible
+import android.view.ViewGroup
+import androidx.compose.foundation.clickable
+import androidx.compose.foundation.layout.*
```
gabrielluong, line 12: We don't use wildcard imports. See https://github.com/mozilla-mobile/firefox-android/tree/main/android-components#style for configuring your Android Studio

gabrielluong (changes requested):
@vrrashkov Thanks again for the contribution. I went through a first pass at reviewing this. Some of the changes request and feedback will seem cumbersome at times, but I will try to explain the reasoning around them. My main ask as our first iteration is to try to take a more iterative approach here where we only convert the `NimbusExperimentsFragment` and make use of the existing reusable components. 

After that is landed, I feel like we can explore removing the `NimbusBranchesFragment` in favour of doing everything in a single Fragment. I am not sure if that is possible, but I can definitely see it being worthwhile to simplify our navgraph and fragments. We don't necessarily have to go with this approach and keep with your existing approach as well, but let's just convert Experiments first, and then you can let me know what your preferences are.

gabrielluong:
@vrrashkov I am gonna close this PR since we're carrying over the work on the new PRs.

PR #27689, opened 2022-11-03
Close #7227: Re-enable storage maintenance call.

### LINKED ISSUE #7227: Re-enable storage maintenance call
In https://github.com/mozilla-mobile/fenix/issues/2765 we added periodic storage maintenance calls that run once a day (or so), during startup.

In https://github.com/mozilla-mobile/fenix/issues/6937 we saw that this interferes with our fennec migration code, which also runs on startup. So, maintenance was disabled.

In https://github.com/mozilla-mobile/fenix/pull/6938 I proposed a solution for this problem, but it doesn't apply now that we're running migration code in a background service.

This issue is to re-enable maintenance call in a way that doesn't interfere with other parts of the browser that may need to acquire higher levels of SQL locks (reserved+).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-14853)

kycn:
Re-enable storage maintenance call by introducing WorkManager worker on A-C side and consuming it from Fenix. The work request is periodic and the repeat interval is 24h. It requires the device to be idle and not to have low battery. This feature is available only for Nightly for now.

Please take a look at [this PR](https://github.com/mozilla-mobile/firefox-android/pull/59) for the related A-C changes.

_**Note:** WorkManager does not ensure that the Worker will run exactly at the same time after the period delay passes, so the periodic waiting times may change while testing because of the work request constraints and OS operations. In addition, one of the work request's constraints is setRequiresDeviceIdle(true) and it is hard to meet immediately. Tried forcing the idle mode but could not trigger the worker right away. (My guess is the retry time intervals that are bound to BackoffPolicy which is, by default, EXPONENTIAL. Though, it is not quite clear from the docs whether OS retries the work request to run right after the constraints are met or it aligns with the BackoffPolicy.) I suspect that if the user is using the device actively, meeting the same 24h execution window for the work request is going be very hard._ 

### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation
Fixes #7227

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/FeatureFlags.kt
+++ b/app/src/main/java/org/mozilla/fenix/FeatureFlags.kt
@@ -121,4 +121,9 @@ object FeatureFlags {
      * Enables the save to PDF feature.
      */
     const val saveToPDF = true
+
+    /**
+     * Enables storage maintenance feature
+     * */
+    val storageMaintenanceFeature = Config.channel.isNightlyOrDebug
 }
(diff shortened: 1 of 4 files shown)


jonalmeida (approved):
Excellent! :shipit:

PR #27730, opened 2022-11-06
Closes #27290 - Refactor DefaultLoadBitmapUseCase to receive orientation as argument

### LINKED ISSUE #27290: Refactor `DefaultLoadBitmapUseCase` to receive orientation as argument
The `WallpaperUseCases.DefaultLoadBitmapUseCase` currently accepts a `Context` as an argument. It then determines the current device orientation inline in order to load the correct (landscape/portrait) bitmap for a wallpaper. 

By removing the `Context` parameter and replacing it with the orientation, the function should be more testable and maintainable. It is also the only `Context`-dependent part of the updated Use Cases and is a copied holdover from the legacy versions, so it will mean we should be able to completely remove Context from `WallpaperUseCases` after https://github.com/mozilla-mobile/fenix/issues/26968

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21728)

PrashantSaroj:
### Pull Request checklist

 Tests can be taken as separate task
 UI is not affected

### QA

- [ ] **QA Needed**
- [ ] No

### GitHub Automation
Fixes #27290

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/wallpapers/WallpapersUseCases.kt
+++ b/app/src/main/java/org/mozilla/fenix/wallpapers/WallpapersUseCases.kt
@@ -76,7 +76,10 @@ class WallpapersUseCases(
     }
     val loadBitmap: LoadBitmapUseCase by lazy {
         if (FeatureFlags.wallpaperV2Enabled) {
-            DefaultLoadBitmapUseCase(context)
+            DefaultLoadBitmapUseCase(
+                filesDir = context.filesDir,
+                getOrientation = { context.resources.configuration.orientation },
+            )
         } else {
             LegacyLoadBitmapUseCase(context)
         }
@@ -372,17 +375,19 @@ class WallpapersUseCases(
     }
 
     @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
-    internal class DefaultLoadBitmapUseCase(private val context: Context) : LoadBitmapUseCase {
+    internal class DefaultLoadBitmapUseCase(
+        private val filesDir: File,
+        private val getOrientation: () -> Int,
+    ) : LoadBitmapUseCase {
         override suspend fun invoke(wallpaper: Wallpaper): Bitmap? =
-            loadWallpaperFromDisk(context, wallpaper)
+            loadWallpaperFromDisk(wallpaper)
 
         private suspend fun loadWallpaperFromDisk(
-            context: Context,
             wallpaper: Wallpaper,
         ): Bitmap? = Result.runCatching {
-            val path = wallpaper.getLocalPathFromContext(context)
+            val path = wallpaper.getLocalPathFromContext()
             withContext(Dispatchers.IO) {
-                val file = File(context.filesDir, path)
+                val file = File(filesDir, path)
                 BitmapFactory.decodeStream(file.inputStream())
             }
         }.getOrNull()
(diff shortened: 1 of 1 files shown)


gabrielluong:
@PrashantSaroj Thanks for the contribution, could you include a description to your commit message? Something like "Closes #27290 - Refactor DefaultLoadBitmapUseCase to receive orientation as argument"

app/src/main/java/org/mozilla/fenix/wallpapers/WallpapersUseCases.kt
```diff
@@ -372,17 +372,19 @@ class WallpapersUseCases(
     }
 
     @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
-    internal class DefaultLoadBitmapUseCase(private val context: Context) : LoadBitmapUseCase {
+    internal class DefaultLoadBitmapUseCase(
+        private val filesDir: File,
+        private val orientation: Int,
```
jonalmeida, line 377: These are lazy singletons. So doesn't that mean the first orientation that is queried will be used for that application lifecycle?

PrashantSaroj: @jonalmeida I was under the impression that this object will be recreated on configuration changes. I checked and it is indeed singleton. Does this mean, we would not be able to decouple context and this class?

jonalmeida: I think you can decouple it from the class constructor but will probably need to keep it in the method parameters to query at the time of use for the orientation.

PrashantSaroj: If we pass it in the method, then we will need to change LegacyBitmapUseCase.invoke() as well (apply wallpaper method of HomeFragment assumes invoke() of both Legacy and Default UseCase to follow the same signature  ). Should I go ahead and make the change?

MatthewTighe: Rather than update the `invoke` function, I would prefer if we changed the `orientation` parameter sent to `DefaultLoadBitmapUseCase` to be a functional parameter. That way, the `Context` can be re-queried for the current orientation every time `invoke` is run. So something like:

```
    DefaultLoadBitmapUseCase(context.filesDir, { context.resources.configuration.orientation })

    internal class DefaultLoadBitmapUseCase(
        private val filesDir: File,
        private val getOrientation: () -> Int,
    )
```

PrashantSaroj: @MatthewTighe I have updated the PR with your suggestion.

MatthewTighe (approved):
This looks good to me, thanks for your contribution! According to our [process for merging contributor patches](https://github.com/mozilla-mobile/fenix/wiki/Guide-to-merging-contributor-PRs), I have opened https://github.com/mozilla-mobile/fenix/pull/27937 in order to land your change 🙂

gabrielluong:
@PrashantSaroj Thank you for the contribution! I am closing this PR since it was already landed via https://github.com/mozilla-mobile/fenix/pull/27937.

PR #26902, opened 2022-09-09
mergify: allow complete-push check as an alternative to complete-pr i…

jcristau:
…n queue conditions

PRs created by github actions don't get a pull-request taskcluster graph scheduled (because that is only done for PRs created by collaborators). However, if the PR is created in the main fenix repo, typically because it comes from one of our internal github workflows, then we get a push graph, which is sufficient to run the required checks.

This change lets mergify consider a green `complete-push` task as sufficient to queue a PR (in addition to all the other conditions, e.g. wrt review), which will allow PRs created by the
update-nimbus-experiments workflow to be queued by mergify instead of having to be merged by a repo admin.

### Pull Request checklist

- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features. In addition, it includes a screenshot of a successful [accessibility scan](https://play.google.com/store/apps/details?id=com.google.android.apps.accessibility.auditor&hl=en_US) to ensure no new defects are added to the product.

### QA

- [ ] **QA Needed**

### To download an APK when reviewing a PR (after all CI tasks finished running):
1. Click on `Checks` at the top of the PR page.
2. Click on the `firefoxci-taskcluster` group on the left to expand all tasks.
3. Click on the `build-debug` task.
4. Click on `View task in Taskcluster` in the new `DETAILS` section.
5. The APK links should be on the right side of the screen, named for each CPU architecture.

### GitHub Automation

Used by GitHub Actions.

PR #27774, opened 2022-11-09
Update mergify queue conditions to account for update-nimbus-experime…

jcristau:
…nts PRs

The update-nimbus-experiments workflow creates PRs that don't schedule a `complete-pr` task (because the github-actions bot is untrusted), but in this case we have a `complete-push` task which is good enough.

### Pull Request checklist

### QA

- [ ] **QA Needed**

### GitHub Automation

Used by GitHub Actions.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -3,18 +3,7 @@ queue_rules:
     conditions:
       - or:
         - status-success=complete-pr
-        - and:
-          # For more context, see "Auto Merge" rules down below
-          - status-success=complete-push
-          - or:
-            - and:
-              - base~=^releases[_/].*
-              - or:
-                - head~=^automation/sync-strings-\d+
-                - head~=^relbot/fenix-\d+
-            - and:
-              - base=main
-              - head~=^relbot/AC-Nightly-.+
+        - status-success=complete-push
 pull_request_rules:
   - name: Resolve conflict
     conditions:


jcristau:
AIUI effectively the queue_rules need to be a subset of the pull_request_rules otherwise (if a PR matches pull_request_rules but not queue_rules) we end up with PRs that are queued but can never actually merge, and block the queue for everything else.  This change fixes a bug introduced in #26902, and which bit us with #27467 yesterday.

JohanLorenzo (approved):
These changes make sense to me. The reason why https://github.com/mozilla-mobile/fenix/pull/27467 blocked the rest of the queue is because it's not an automated PR we automatically approve.

It's still created by `github-actions[bot]` (that's why there's no `complete-pr` status) but mergify doesn't have any rules about it. That's the reason why it didn't check enough boxes[1]:

image

I wonder whether we should just allowlist this workflow too. I'm actually unsure. On the one hand, we must know what github-action create which automated PR - thus we should list it in `.mergify.yml`. On the other hand, anytime a Github action is created/modified, people have to remember to update `.mergify.yml`. That's not a great idea.

Moreover, if `complete-push` is green, that's already a strong assessment that: 
 * the code lives in on the main repo (and not a fork)
 * it comes from a trusted party

Therefore, I agree with the changes, let's just make this rule simpler and accept commits that are already in the queue and got their `complete-push` job green.

[1] https://github.com/mozilla-mobile/fenix/runs/9369391857

PR #27793, opened 2022-11-10
For#27753 - Update search engines immediately after changing language

### LINKED ISSUE #27753: [Bug]: Search engines are not updated immediately after changing language
### Steps to reproduce

1. Change language from system language or app settings
2. Check the available search engines

### Expected behaviour

The list of available search engines is automatically updated for the newly selected locale

### Actual behaviour

An app restart is needed for the list of search engines to be updated.
The issues happens with or without unified search enabled.

### Firefox version

Seems like an old issue, happens in the current release also.

### Additional information

https://user-images.githubusercontent.com/11428869/200523702-ea08d843-a454-49e1-91a9-4246197008b7.mp4

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21950)
┆Link To Issue: https://mozilla-hub.atlassian.net/browse/FNXV2-21950

DreVla:
Needs https://github.com/mozilla-mobile/firefox-android/pull/135

When changing language inside the application:

https://user-images.githubusercontent.com/32488956/201075839-de6249e4-01ef-44f7-bce2-796ba8abc7c3.mp4

When changing language from device settings:

https://user-images.githubusercontent.com/32488956/201076616-7b6b7d93-9b85-4e08-a227-80cf0d81ad9a.mp4

### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation
Fixes #27753

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
+++ b/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
@@ -42,6 +42,7 @@ import kotlinx.coroutines.delay
 import kotlinx.coroutines.launch
 import mozilla.appservices.places.BookmarkRoot
 import mozilla.components.browser.state.action.ContentAction
+import mozilla.components.browser.state.action.SearchAction
 import mozilla.components.browser.state.search.SearchEngine
 import mozilla.components.browser.state.selector.getNormalOrPrivateTabs
 import mozilla.components.browser.state.state.SessionState
@@ -384,6 +385,11 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
             DefaultBrowserNotificationWorker.setDefaultBrowserNotificationIfNeeded(applicationContext)
             ReEngagementNotificationWorker.setReEngagementNotificationIfNeeded(applicationContext)
         }
+
+        // This was done in order to refresh search engines when app is running in background
+        // and the user changes the system language
+        // More details here: https://github.com/mozilla-mobile/fenix/pull/27793#discussion_r1029892536
+        components.core.store.dispatch(SearchAction.RefreshSearchEnginesAction)
     }
 
     override fun onStart() {
--- a/app/src/androidTest/java/org/mozilla/fenix/ui/SettingsSearchTest.kt
+++ b/app/src/androidTest/java/org/mozilla/fenix/ui/SettingsSearchTest.kt
@@ -420,11 +420,7 @@ class SettingsSearchTest {
                 "Bing",
                 "Amazon.com",
                 "DuckDuckGo",
-                "eBay",
-                /* Disabled Arabic Wikipedia verification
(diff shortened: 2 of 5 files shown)


app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -378,6 +379,8 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
 
             DefaultBrowserNotificationWorker.setDefaultBrowserNotificationIfNeeded(applicationContext)
         }
+
+        components.core.store.dispatch(InitAction)
```
Mugurell, line 383: While this `InitAction` will work to reconfigure the search engines it will also reset the current state.
Looks like we'll need a new action to only re-initialize the search engines list.

DreVla: A new action `RefreshSearchEnginesAction` was added to the `BrowserAction` to allow for refreshing search engines when app locale is changed. There are more details here: https://github.com/mozilla-mobile/firefox-android/pull/135

app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -378,6 +379,8 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
 
             DefaultBrowserNotificationWorker.setDefaultBrowserNotificationIfNeeded(applicationContext)
         }
+
+        components.core.store.dispatch(SearchAction.RefreshSearchEnginesAction)
```
MatthewTighe, line 383: It seems like the update behavior is covered between the `InitAction` during app startup or the dispatch from `DefaultLocaleSettingsController`. Do we also need this dispatch, or can we rely on the other two?

DreVla: This call covers the case when the app is running in the background and the user changes the language from system settings then comes back in the app. Removing this call will lead to the following behavior:

https://user-images.githubusercontent.com/32488956/203505119-032e6e66-11a4-4841-bfb3-c7b358094f21.mp4

Nevertheless you are right that it is not clear enough why the dispatch is there, therefore I will add a comment before, linking to this thread. Thanks for pointing it out!

MatthewTighe:
I've verified this patch locally, but the test failure looks like it could be real. I'm assuming it's because whatever language is switched to in the test doesn't include eBay in the search listings. Happy to approve once the test is fixed.

In general, I (and I think several others) will get notifications if you use the "re-request review" button in the reviewers section of the PR. Please use it when a patch is ready for another pass 🙂

DreVla:
I've squashed the commit which modified tests with the one containing the tested code. Still kept two commits separated since they handle difference scenarios.

PR #27813, opened 2022-11-12
For #27812: Use upstream android.R.string.{cancel, ok} and merge common string resources

### LINKED ISSUE #27812: Use upstream android.R.string.{cancel, ok} and merge common string resources
I've noticed that there are some strings that are duplicated several times in `strings.xml` where they don't have to be, including `Cancel`, `Ok`, `Delete`, `Edit`, `Close`, `Remove`, etc. Translation efforts would likely improve with merging the common uses together (I have noticed some languages miss translations of some instances of e.g. `Cancel`), and there is minor reductions in apk size associated with merging these resources (~`20 kB` with just changing all `Cancel` instances to `android.R.string.cancel`).

A PR will be linked shortly that tackles what I saw were the obvious cases.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21976)

fice-t:
There are some more resources that can be included here (such as `add`, `undo` ...), but I figured I should stop and throw this out for review before proceeding further.

### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation
Fixes #27812

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -10,6 +10,7 @@
     <!-- Labels common to multiple components -->
     <string name="common_label_delete">Delete</string>
     <string name="common_label_edit">Edit</string>
+    <string name="common_label_save">Save</string>
 
     <!-- Home Fragment -->
     <!-- Content description (not visible, for screen readers etc.): "Three dot" menu button. -->
@@ -844,8 +845,6 @@
     <string name="bookmark_menu_open_all_in_tabs_button">Open all in new tabs</string>
     <!-- Bookmark overflow menu open all in private tabs button -->
     <string name="bookmark_menu_open_all_in_private_tabs_button">Open all in private tabs</string>
-    <!--Bookmark overflow menu save button -->
-    <string name="bookmark_menu_save_button">Save</string>
     <!-- Bookmark multi select title in app bar
      The first parameter is the number of bookmarks selected -->
     <string name="bookmarks_multi_select_title">%1$d selected</string>
@@ -988,8 +987,6 @@
     <string name="create_collection_tab_saved">Tab saved!</string>
     <!-- Content description (not visible, for screen readers etc.): button to close the collection creator -->
     <string name="create_collection_close">Close</string>
-    <!-- Button to save currently selected tabs in the "select tabs" step of the collection creator-->
-    <string name="create_collection_save">Save</string>
     <!-- Snackbar action to view the collection the user just created or updated -->
     <string name="create_collection_view">View</string>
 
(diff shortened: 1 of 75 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -7,6 +7,15 @@
...
     <string name="app_name_private_4">%s (Private)</string>
 
+    <!-- Labels common to multiple components -->
+    <string name="common_label_close">Close</string>
+    <string name="common_label_delete">Delete</string>
+    <string name="common_label_edit">Edit</string>
+    <string name="common_label_off">Off</string>
+    <string name="common_label_on">On</string>
+    <string name="common_label_remove">Remove</string>
+    <string name="common_label_save">Save</string>
```
Mugurell, line 17: These labels would probably need individual comments to help localizers understand the context of where they are used.
For example this is how each string (in this case `recently_saved_menu_item_remove`) would appear for localizers - https://pontoon.mozilla.org/ro/firefox-for-android/mozilla-mobile/fenix/app/src/main/res/values/strings.xml/?search=recently_saved_menu_item_remove&string=242751

Mugurell:
Reusing strings may cause issues when one usecase needs a different value but this may be handled later on a case by case basis - we will then add a new string to use for that specific usecase.
Not sure if leveraging the Android strings for "ok" and "cancel" is the best approach for us though, I don't know in what languages each string is translated while if using the strings from the app we would have more control.
Probably @Delphine can review and suggest the best approach here.

gabrielluong:
I tried to do this at some point. I don't believe this was okay with the l10n team.

gabrielluong:
Found some of my PRs where I added generic strings https://github.com/mozilla-mobile/fenix/pull/18730 and then the subsequent removal in https://github.com/mozilla-mobile/fenix/pull/18769. I will admit we're lacking a bit of documented context here. I believe this was discussed internally with the l10n team where it was pointed out that we can't have these generic strings because for localization we need to document their specific usages because that could have translation implications because of how the string is used in their specific context. This is why we cannot simply have generic strings (eg, Ok, Delete) that could simply be used everywhere. I will inquire with @Delphine to reconfirm this, but it's very likely we won't be able to use the upstream (Ok, Cancel) strings either and would probably just prefer what we already have.

fice-t:
> I believe this was discussed internally with the l10n team where it was pointed out that we can't have these generic strings because for localization we need to document their specific usages because that could have translation implications because of how the string is used in their specific context.

That makes sense generally, but I'm not sure how it applies to (at least most of) the simple action strings here. I did some manual cross-checking and didn't find any differences outside of some languages missing some variants. Is there a simple automatic way to diff compare translations for resources over all languages?

>I will inquire with @Delphine to reconfirm this, but it's very likely we won't be able to use the upstream (Ok, Cancel) strings either and would probably just prefer what we already have.

How about a hybrid approach? It appears possible to create string aliases in XML files using e.g.:

```xml
<string name="common_label_cancel">@android:string/cancel</string>
```
for the upstream resources and:

```xml
<string name="recently_saved_menu_item_remove">@string/common_label_remove</string>
```
for local resources.

That way there exists a common default that languages can optionally override if useful. I'm not sure if Android can optimize these aliases out, but it seems like a useful way to minimize translation efforts and reduce missing translations for similar strings.

gabrielluong:
I would imagine we wouldn't want to do any aliasing as well. These strings are imported to other repos where it is then uploaded to a backend db where translators could then see any new strings that are added and requires translation. Since we don't necessarily own the process to localization, I don't believe we are in the position to prescribe any changes here.

Delphine:
[This is the documentation ](https://mozilla-l10n.github.io/documentation/localization/dev_best_practices.html#dont-reuse-strings-in-different-contexts)explaining why we avoid re-using strings in most cases.
It's very tricky to know which locales use which terms in certain areas. What works for most doesn't work for all.

gabrielluong:
@fice-t Thank you once again for the contribution. I am closing the PR since we won't be able move forward with this - sorry about that.

Issue #26855, opened 2022-09-07
Upcoming changes to our Android repositories


moz-mobile-admin:
📢 Hi everyone,

Today we're announcing two upcoming changes to our Android repositories and development processes that will make getting involved, building our products and shipping improvements to Firefox users simpler, easier, and a lot faster.

In the coming months, we will be consolidating the [Fenix](https://github.com/mozilla-mobile/fenix), [Focus](https://github.com/mozilla-mobile/focus-android/) and [Android Components](https://github.com/mozilla-mobile/android-components) projects into a single new [Firefox-Android](https://github.com/mozilla-mobile/firefox-android) repository. As part of that change, we also plan to move issue tracking to Bugzilla, migrating open GitHub tickets there in the process. 

Having our Android code under one roof has some obvious benefits as far as testing, CI and shipping are concerned - and we're expecting some big wins there! - but it also lowers barriers to community involvement and improves development speed. Whenever you can turn a complex process into one command in one repo everyone benefits, from partners and first-time contributors to veteran developers and release engineers alike. 

Moving issue tracking to Bugzilla offers many similar benefits, including tighter coordination with Mozilla's internal teams and outside partners, as well as robust community safety tools to name a few.

Fenix, Focus and Android Components will stay on the same release cadence, we'll continue to maintain and publish components and all functionality as usual, and our artifact, package and application IDs will also stay the same.

We wanted to announce this well ahead of time: while the schedule isn't firm yet, we intend to make this change in the next small number of months. As we prepare for this change, we want to thank you all for your help and support; your work has been critical to getting Android Components, Fenix, and Focus this far. We're grateful for your contributions, and believe these changes will make it easier to continue the amazing work that has brought these projects so far. We will reach out individually on all open Pull Requests to discuss next steps, and we'll also take care of migrating open tickets, when the time comes.

⚠️ Please file any new tickets in Bugzilla using the [Fenix product](https://bugzilla.mozilla.org/enter_bug.cgi?product=Fenix&bug_type=defect), and select the corresponding component. If the component is not in the list, select "General". 

If you’d like to get in touch, you can find us on [Matrix](https://wiki.mozilla.org/Matrix#Connect_to_Matrix) in the following channels:
[#fenix:mozilla.org](https://chat.mozilla.org/#/room/#fenix:mozilla.org)
[#focus-android:mozilla.org](https://chat.mozilla.org/#/room/#focus-android:mozilla.org)
[...]

PR #27816, opened 2022-11-12
For #27815: Properly enforce http(s) protocol for new login hostname

### LINKED ISSUE #27815: Make non-http(s) protocols invalid when adding new login
Non-http(s) protocols are apparently not supported by Fenix at the moment (going by the `@string/add_login_hostname_invalid_text_3` hint), but the `AddLoginFragment` currently does not enforce this hint. A PR will be linked shortly that fixes this by upgrading the hint to a proper error.

At the same time, desktop Firefox auto-adds `https://` when adding a new login when no protocol is present, so I have implemented that for feature parity and a better user experience.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21977)

fice-t:
This also auto-adds the https protocol if no protocol is present for feature parity with desktop Firefox.

### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation
Fixes #27815

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/settings/logins/fragment/AddLoginFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/logins/fragment/AddLoginFragment.kt
@@ -199,6 +199,13 @@ class AddLoginFragment : Fragment(R.layout.fragment_add_login), MenuProvider {
             },
         )
 
+        binding.hostnameText.setOnFocusChangeListener { _, hasFocus ->
+            val hostnameText = binding.hostnameText.editableText
+            if (!(hasFocus || hostnameContainsProtocol || hostnameText.isEmpty())) {
+                hostnameText.insert(0, "https://")
+            }
+        }
+
         binding.usernameText.addTextChangedListener(
             object : TextWatcher {
                 override fun afterTextChanged(u: Editable?) {
--- a/app/src/main/java/org/mozilla/fenix/settings/logins/fragment/AddLoginFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/settings/logins/fragment/AddLoginFragment.kt
@@ -52,6 +52,7 @@ class AddLoginFragment : Fragment(R.layout.fragment_add_login), MenuProvider {
     private var validPassword = false
     private var validUsername = true
     private var validHostname = false
+    private var hostnameContainsProtocol = false
     private var usernameChanged = false
 
     private var _binding: FragmentAddLoginBinding? = null
@@ -155,6 +156,7 @@ class AddLoginFragment : Fragment(R.layout.fragment_add_login), MenuProvider {
             object : TextWatcher {
                 override fun afterTextChanged(h: Editable?) {
                     val hostnameText = h.toString()
+                    hostnameContainsProtocol = "://" in hostnameText
 
                     when {
                         hostnameText.isEmpty() -> {
@@ -163,6 +165,10 @@ class AddLoginFragment : Fragment(R.layout.fragment_add_login), MenuProvider {
                         }
                         !Patterns.WEB_URL.matcher(hostnameText).matches() -> {
                             setHostnameError()
+                            binding.clearHostnameTextButton.isEnabled = true
+                        }
+                        hostnameContainsProtocol && !hostnameText.startsWith("http") -> {
+                            setHostnameError(invalidProtocol = true)
                             binding.clearHostnameTextButton.isEnabled = true
                         }
                         else -> {
@@ -317,16 +323,28 @@ class AddLoginFragment : Fragment(R.layout.fragment_add_login), MenuProvider {
         }
     }
 
-    private fun setHostnameError() {
+    private fun setHostnameError(invalidProtocol: Boolean = false) {
         binding.inputLayoutHostname.let { layout ->
             validHostname = false
-            layout.error = context?.getString(R.string.add_login_hostname_invalid_text_2)
+            if (invalidProtocol) {
+                layout.error = context?.getString(R.string.add_login_hostname_invalid_text_3)
+            } else {
+                layout.error = context?.getString(R.string.add_login_hostname_invalid_text_2)
+            }
             layout.setErrorIconDrawable(R.drawable.mozac_ic_warning_with_bottom_padding)
             layout.setErrorIconTintList(
                 ColorStateList.valueOf(
                     ContextCompat.getColor(requireContext(), R.color.fx_mobile_text_color_warning),
                 ),
             )
+        }
+    }
+
+    private fun hostnameEnsureProtocol(hostname: String): String {
+        return if (hostnameContainsProtocol) {
+            hostname
+        } else {
+            "https://$hostname"
         }
     }
 
@@ -362,7 +380,7 @@ class AddLoginFragment : Fragment(R.layout.fragment_add_login), MenuProvider {
         R.id.save_login_button -> {
             view?.hideKeyboard()
             interactor.onAddLogin(
-                binding.hostnameText.text.toString(),
+                hostnameEnsureProtocol(binding.hostnameText.text.toString()),
                 binding.usernameText.text.toString(),
                 binding.passwordText.text.toString(),
             )
--- a/app/src/main/res/layout/fragment_add_login.xml
+++ b/app/src/main/res/layout/fragment_add_login.xml
@@ -42,8 +42,6 @@
         app:layout_constraintEnd_toEndOf="@id/hostnameHeaderText"
         app:layout_constraintStart_toStartOf="@id/hostnameHeaderText"
         app:layout_constraintTop_toBottomOf="@id/hostnameHeaderText"
-        app:helperTextEnabled="true"
-        app:helperText="@string/add_login_hostname_invalid_text_3"
         app:hintEnabled="false" >
 
         <com.google.android.material.textfield.TextInputEditText


app/src/main/java/org/mozilla/fenix/settings/logins/fragment/AddLoginFragment.kt
```diff
@@ -165,6 +167,10 @@ class AddLoginFragment : Fragment(R.layout.fragment_add_login), MenuProvider {
                             setHostnameError()
                             binding.clearHostnameTextButton.isEnabled = true
                         }
+                        hostnameContainsProtocol && !hostnameText.startsWith("http") -> {
```
Mugurell, line 170: Because above this code there is a check for `http|https|rtsp|ftp` it seems like this we would show the `add_login_hostname_invalid_text_3` error just for the `rtsp|ftp` protocols.
This might be okay - we would still be showing an error for other protocols but it shows that we have duplicated code for checking the protocol so the new case introduced here should be merged with the already existing `!Patterns.WEB_URL.matcher(hostnameText).matches()` check. Maybe we can replace the existing one?

app/src/main/res/layout/fragment_add_login.xml
```diff
@@ -42,8 +42,6 @@
         app:layout_constraintEnd_toEndOf="@id/hostnameHeaderText"
         app:layout_constraintStart_toStartOf="@id/hostnameHeaderText"
         app:layout_constraintTop_toBottomOf="@id/hostnameHeaderText"
-        app:helperTextEnabled="true"
-        app:helperText="@string/add_login_hostname_invalid_text_3"
```
Mugurell, line 46: This seems like an important hint regarding what URLs are accepted. Why was it removed?

app/src/main/java/org/mozilla/fenix/settings/logins/fragment/AddLoginFragment.kt
```diff
@@ -330,6 +340,14 @@ class AddLoginFragment : Fragment(R.layout.fragment_add_login), MenuProvider {
         }
     }
 
+    private fun hostnameEnsureProtocol(hostname: String): String {
```
Mugurell, line 343: Could we use a focus listener to set the `https` protocol immediately after the user finishes editing the site url - similar to how it functions on Desktop?
Just to have the protocol clear for the user.
I imagine a case in which the user might omit the protocol, we would add "https" without immediately showing it but the website should user "http". If we show this immediately the user will have the chance to update "https" to "http" immediately.

app/src/main/res/layout/fragment_add_login.xml
fice-t: I moved it from being a hint to being an error, partially because previously the backend silently discarded otherwise-valid new logins that did not start with `http(s)`, and partially because showing the hint to the user when no hostname has been entered might make the user think that they still have to manually add the protocol themselves.

app/src/main/java/org/mozilla/fenix/settings/logins/fragment/AddLoginFragment.kt
fice-t: It could, though I think having a specific error message for non-https protocols is more user-friendly. Are you suggesting to combine the `!hostnameText.startsWith("http")` check with the previous check?

In any case, `hostnameContainsProtocol` is useful for the auto-`https` code, so the end result would be quite similar (`!hostnameText.startsWith("http")` would be converted to a `val` to be both checked and passed to `setHostnameError`).

app/src/main/java/org/mozilla/fenix/settings/logins/fragment/AddLoginFragment.kt
fice-t: I tried that, but there's an issue with Android here in that tapping on the save button immediately after editing the hostname results in the focus listener being called _after_ the save button action is performed -- in this case if the string does not already contain a valid protocol then the backend silently drops the entire login!

I agree though that giving the user a chance to edit is useful. 2 options here:

1) Force the user to tap save again after auto-adding `https` -- I think this inconsistency would be annoying for users.
2) Use both a focus listener and the current PR's `hostnameEnsureProtocol` method -- this would give users that enter/edit the hostname first a chance to see the `https` addition as well as ensure that no invalid hostnames get silently dropped.

Something else to note is that desktop Firefox displays the login detail page right after adding it while Fenix just pops you back into the saved logins list. I think the auto-addition would be smoother if the user would be able to instantly see the new login after creating it. That is, if they have `example.com` and tap the save button, they would be presented with the `LoginDetailFragment` that includes `Site: https://example.com`, which they can then change if they wish.

app/src/main/java/org/mozilla/fenix/settings/logins/fragment/AddLoginFragment.kt
Mugurell: > the end result would be quite similar (!hostnameText.startsWith("http") would be converted to a val to be both checked and passed to setHostnameError). 

Sounds good.
Otherwise it seems to me that the above check for `!Patterns.WEB_URL.matcher(hostnameText).matches()` would catch most of the other protocols and instead of the intent here to show the `Web address must contain "https://" or "http://"` error we would show `Valid hostname required`.

app/src/main/res/layout/fragment_add_login.xml
Mugurell: Can we still keep it the initial hint also?
It's a design that users are accustomed to and helps to more gently inform about the URL restrictions than through an immediate error.
(Also opened https://github.com/mozilla-mobile/fenix/issues/27839 to not be so quick on showing errors)

app/src/main/java/org/mozilla/fenix/settings/logins/fragment/AddLoginFragment.kt
Mugurell: > Something else to note is that desktop Firefox displays the login detail page right after adding it while Fenix just pops you back into the saved logins list. I think the auto-addition would be smoother if the user would be able to instantly see the new login after creating it. That is, if they have example.com and tap the save button, they would be presented with the LoginDetailFragment that includes Site: https://example.com, which they can then change if they wish. 

Seems like something that may worth a new ticket and discussion with UX.

fice-t: > Could we extract the host verification in a separate method which would be executed also when the save button is tapped such that we'd execute the save action only if the host verification passes?

Yes, though with the current solution of focus listener + `hostnameEnsureProtocol` the save button should only be enabled when the login is either valid by itself, or valid with the addition of `https://` (which `hostnameEnsureProtocol` takes care of).

fice-t:
I added a new commit that implements the focus listener.

app/src/main/res/layout/fragment_add_login.xml
fice-t: The problem I have with leaving the hint is that it sort of undermines the auto-addition of `https://`, at least with the current string.

What could be a good middle-ground is a new text hint that states something like "hostname must start with http(s)://. https:// will be added if no protocol is entered manually".

app/src/main/java/org/mozilla/fenix/settings/logins/fragment/AddLoginFragment.kt
fice-t: > Otherwise it seems to me that the above check for !Patterns.WEB_URL.matcher(hostnameText).matches() would catch most of the other protocols and instead of the intent here to show the Web address must contain "https://" or "http://" error we would show Valid hostname required.

I'm a bit confused here, in what cases would it show `Valid hostname required` where you say it should show `Web address must contain "https://" or "http://"`? My intent is to only show the `http(s)` error string if `hostnameText` is a valid non-http(s) URL, so an invalid `hostnameText` URL should indeed show `Valid hostname required` before the `http(s)` error string.

It turns out that merging the two branches with no logic change results in messier code IMO, so I think if no logic change is required then it would be better to leave this part alone.

app/src/main/res/layout/fragment_add_login.xml
Mugurell: Adding a new string requires UX approval.
The proposed one would modify the UI a bit being quite long.

app/src/main/java/org/mozilla/fenix/settings/logins/fragment/AddLoginFragment.kt
Mugurell: There are two checks for protocols. [One previously existing](https://github.com/mozilla-mobile/fenix/blob/48288f0198394b4e57d14549682f4d22a04ea30e/app/src/main/java/org/mozilla/fenix/settings/logins/fragment/AddLoginFragment.kt#L164-L167) and the new one you are adding just below.
This would mean that a valid url using a different protocol than what `Patterns.WEB_URL` matches would show `Valid hostname required` and only urls with protocols matching `Patterns.WEB_URL` but not using `https` / `https` would show `Web address must contain "https://" or "http://"`.

https://user-images.githubusercontent.com/11428869/202380006-de6ecfe2-4f43-4131-b72e-b4bc90e1e8b8.mov


To me it looks like this checks should be combined (by removing the current `Patterns.WEB_URL` match):
- if there is a protocol we should match it with your new `http(s)` check or show the `Web address must contain "https://" or "http://"` error.
- if there is no protocol and the url is invalid then show the `Valid hostname required` error.

app/src/main/java/org/mozilla/fenix/settings/logins/fragment/AddLoginFragment.kt
Mugurell: Thank you! Tested the version with the new focus listener and it really improves the entire experience.
Asked UX to confirm the scenario of automatically adding the `https` protocol when the save button is pressed (mentioned that it's the same behavior on desktop).

Mugurell:
Thank you! This does looks nice 👍.
One issue I'm seeing though in the latest version of the patch is that the save button can be tapped with a missing username but cannot be after the username is deleted.
Issue does not reproduce for missing password but available username:

https://user-images.githubusercontent.com/11428869/202386652-27c7bfb0-304a-44f0-bf8d-56e3d8e78385.mov

Mugurell:
@fice-t Wondering if you plan to continue the work or we can take over to fix the issue from https://github.com/mozilla-mobile/fenix/pull/27816#issuecomment-1318220662 and land this?

gabrielluong:
Hey there, thank you for the contribution! We have migrated the original issue over to Bugzilla https://bugzilla.mozilla.org/show_bug.cgi?id=1815632. Could you please update your commit message and PR title to reference the new Bugzilla issue using the following format: "Bug xxxxxx - \<short description\>" prior to landing. Thanks!

csadilek:
Hey @fice-t, Thank you so much for your contribution. Unfortunately, we are in the process of moving this repository to [firefox-android](https://github.com/mozilla-mobile/firefox-android), and won't be able review or land this PR until after  the migration.

It looks there's follow-up needed, see https://github.com/mozilla-mobile/fenix/pull/27816#issuecomment-1408532139, but please feel free to reopen a new PR against the new repo once migration is complete. You will have to clone the new repository and reapply your changes. Thank you again, and sorry for the inconvenience. 

For more information, please see https://github.com/mozilla-mobile/fenix/issues/26855.

PR #27843, opened 2022-11-15
For #26826 - Add test tags allowing to differentiate Pocket stories

### LINKED ISSUE #26826: Support UI testing composables by using testTag
After discussing with the QA team it was apparent that testing is currently hindered by the inability / inefficiency of identifying composables on the screen.
`testTag`s are the approach recommended by the Compose team to help with that - https://developer.android.com/jetpack/compose/testing and we are already using them in A-C - https://searchfox.org/mozilla-mobile/search?q=testTag&path=&case=true&regexp=false

I propose starting to add (even retroactively for the scenarios raised by QA) `testTag`s to the composables used on Fenix with the following naming scheme:
```kotlin
<feature>.<composable>(.<optional other sub composables>)
```

Compose v1.2 will also allow better interoperability with UIAutomator by allowing the `testTag`s to be used in already common calls in the form of
> findObject(By.res(<`testTag`>)).

An update to this newer version is prepared in https://github.com/mozilla-mobile/fenix/pull/26763.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21482)

Mugurell:
### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation
Fixes #26826

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesComposables.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesComposables.kt
@@ -279,32 +279,43 @@ fun PocketStories(
         itemsIndexed(storiesToShow) { columnIndex, columnItems ->
             Column(verticalArrangement = Arrangement.spacedBy(8.dp)) {
                 columnItems.forEachIndexed { rowIndex, story ->
-                    if (story == placeholderStory) {
-                        ListItemTabLargePlaceholder(stringResource(R.string.pocket_stories_placeholder_text)) {
-                            onDiscoverMoreClicked("https://getpocket.com/explore?$POCKET_FEATURE_UTM_KEY_VALUE")
-                        }
-                    } else if (story is PocketRecommendedStory) {
-                        PocketStory(
-                            story = story,
-                            backgroundColor = backgroundColor,
-                        ) {
-                            val uri = Uri.parse(story.url)
-                                .buildUpon()
-                                .appendQueryParameter(URI_PARAM_UTM_KEY, POCKET_STORIES_UTM_VALUE)
-                                .build().toString()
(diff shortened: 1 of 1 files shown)


AndiAJ (approved):
LGTM! 👍🏽

app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesComposables.kt
```diff
@@ -279,32 +279,43 @@ fun PocketStories(
...
-                            },
-                        ) {
-                            PocketSponsoredStory(
+                    Box(
+                        modifier = Modifier.semantics {
+                            testTagsAsResourceId = true
+                            testTag = when (story) {
+                                placeholderStory -> "pocket.discover.more.story"
+                                is PocketRecommendedStory -> "pocket.recommended.story"
+                                else -> "pocket.sponsored.story"
```
gabrielluong, line 288: Maybe we should refactor all of these testTags string to constants so that they can be reused in the UI tests and much more durable if we ever refactor the tag naming.. This could be a followup.

gabrielluong (approved):

app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesComposables.kt
Mugurell: Agree.
Another thing about these changes is that it shows a pattern of wrapping existing composables in a `Box` to add the test tags. 
I'm not a fan of this.
Was wondering if we should consider adding a new `testTag` parameter to all reusable composables to ease setting it without new wrappers.

gabrielluong: Let's prototype a small minimal change to see how that looks and feels in terms of developer ergonomics.

Issue #27575, opened 2022-10-27
Refactor usages of `toLong(radix = 16)` to an extension function


MozillaNoah:
Per comment https://github.com/mozilla-mobile/fenix/pull/27556#discussion_r1006259265:

![image](https://user-images.githubusercontent.com/87384386/198367759-f4ee70b5-cd37-48e0-a7e5-50018698469a.png)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21877)

chetanmuliya:
private fun String.toHexColorLong(): Long = 16 like this ? can you assign me this issue?

MozillaNoah:
> private fun String.toHexColorLong(): Long = 16 like this ? can you assign me this issue?

Hi @chetanmuliya! We're specifically looking to create an extension function for Strings to shorthand `String.toLong(radix = 16)` to something like `String.toHexColorLong()`. From there, in a second commit, we'd refactor all usages of `String.toLong(radix = 16)` with the new extension function.

chetanmuliya:
I had created a extension function -> **private fun String.toHexColorLong(): Long = 16**

and usage -> **TURNING_RED_WALLPAPER_TEXT_COLOR.toHexColorLong()**

Is that what we wanted?

MozillaNoah:
> I had created a extension function -> **private fun String.toHexColorLong(): Long = 16**
> 
> and usage -> **TURNING_RED_WALLPAPER_TEXT_COLOR.toHexColorLong()**
> 
> Is that what we wanted?

The above `toHexColorLong` simply returns the number 16. We instead want `toHexColorLong` to convert the _entire_ String to a number with a data type of Long. The number `16` is used as the `radix` parameter in `.toLong()` to indicate the base of the number we want to convert to. By default, this is **base 10**, but since color values are represented using hexidecimal (**base 16**), such as `#FDE9C3` and `#532906`, we need to normally do `someString.toLong(radix = 16)`. The issue created here is to create an extension function to do:

`someString.toHexColorLong()` instead of `someString.toLong(radix = 16)`

We have colors hardcoded as Strings (see `LegacyWallpaperMigration.TURNING_RED_WALLPAPER_TEXT_COLOR` for an example) that are written has hexidecimal (base 16) that have to be converted to a Long. The `toHexColorLong()` extension function acts as a more human-readable way to represent the conversion from a hex color String to a Long.

Does that make a little more sense?

[...]

PR #27847, opened 2022-11-15
For #27575: Add String.toHexColor extension

kl3jvi:
Added an extension function on the Colors.kt file, to replace toLong(radix=16) function with toHexColor()


kl3jvi:
@jonalmeida @MozillaNoah Can you please review this pr if this is better?

jonalmeida:
Thanks @kl3jvi , your title is correct, but the commit is "added hex color to" instead of using the same thing as the title. Would you like some help correcting this? 🙂

kl3jvi:
@jonalmeida Sure

jonalmeida (approved):
Fixed and it looks good. Let's land it. 🙂

PR #27851, opened 2022-11-15
For #27457 - Refactor Pocket category colors

### LINKED ISSUE #27457: Refactor the Pocket category colors into a singular data class
After the introduction of wallpaper-adjusted colors, `PocketCategoriesViewHolder` initializes four colors to cover the selection state. For parameter simplification and down-the-road refactoring quality of life, this would benefit from having the colors wrapped in a new data class, such as `PocketCategoryColors`. This way, a single parameter can be passed around rather than a growing list of colors as the business logic evolves.

The default/fallback colors should be be inlined with the data class and the fallbacks when the passed-in colors are null.

Any nullability handling should be handled with a secondary constructor.

```
data class PocketCategoryColors(
    val selectedBackgroundColor: Color,
    val unselectedBackgroundColor: Color,
    val selectedTextColor: Color,
    val unselectedTextColor: Color,
)
```
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21821)

MozillaNoah:
### Pull Request checklist

### QA

- [ ] **QA Needed**

### GitHub Automation
Fixes #27457

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/home/pocket/PocketCategoriesViewHolder.kt
+++ b/app/src/main/java/org/mozilla/fenix/home/pocket/PocketCategoriesViewHolder.kt
@@ -4,6 +4,7 @@
 
 package org.mozilla.fenix.home.pocket
 
+import android.content.res.Configuration
 import android.view.View
 import androidx.compose.foundation.isSystemInDarkTheme
 import androidx.compose.foundation.layout.Column
@@ -12,7 +13,6 @@ import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
-import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.platform.ComposeView
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.tooling.preview.Preview
@@ -25,7 +25,6 @@ import org.mozilla.fenix.components.components
 import org.mozilla.fenix.compose.ComposeViewHolder
 import org.mozilla.fenix.compose.home.HomeSectionHeader
 import org.mozilla.fenix.theme.FirefoxTheme
-import org.mozilla.fenix.theme.Theme
 import org.mozilla.fenix.wallpapers.WallpaperState
 
 internal const val POCKET_CATEGORIES_SELECTED_AT_A_TIME_COUNT = 8
@@ -61,10 +60,8 @@ class PocketCategoriesViewHolder(
         val wallpaperState = components.appStore
             .observeAsComposableState { state -> state.wallpaperState }.value ?: WallpaperState.default
 
-        var selectedBackgroundColor: Color? = null
-        var unselectedBackgroundColor: Color? = null
-        var selectedTextColor: Color? = null
-        var unselectedTextColor: Color? = null
+        var (selectedBackgroundColor, unselectedBackgroundColor, selectedTextColor, unselectedTextColor) =
+            PocketStoriesCategoryColors.buildColors()
         wallpaperState.composeRunIfWallpaperCardColorsAreAvailable { cardColorLight, cardColorDark ->
             if (isSystemInDarkTheme()) {
                 selectedBackgroundColor = cardColorDark
@@ -79,16 +76,20 @@ class PocketCategoriesViewHolder(
             }
         }
 
+        val categoryColors = PocketStoriesCategoryColors(
+            selectedTextColor = selectedTextColor,
+            unselectedTextColor = unselectedTextColor,
+            selectedBackgroundColor = selectedBackgroundColor,
+            unselectedBackgroundColor = unselectedBackgroundColor,
+        )
+
         // See the detailed comment in PocketStoriesViewHolder for reasoning behind this change.
         if (!homeScreenReady) return
         Column {
             Spacer(Modifier.height(24.dp))
 
             PocketTopics(
-                selectedBackgroundColor = selectedBackgroundColor,
-                unselectedBackgroundColor = unselectedBackgroundColor,
-                selectedTextColor = selectedTextColor,
-                unselectedTextColor = unselectedTextColor,
+                categoryColors = categoryColors,
                 categories = categories ?: emptyList(),
                 categoriesSelections = categoriesSelections ?: emptyList(),
                 onCategoryClick = interactor::onCategoryClicked,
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesComposables.kt
```diff
@@ -454,6 +448,41 @@ fun PocketStoriesCategories(
...
+        /**
+         * Builder function used to construct an instance of [PocketStoriesCategoryColors].
+         */
+        @Composable
+        fun buildColors(
+            selectedTextColor: Color = FirefoxTheme.colors.actionPrimary,
+            unselectedTextColor: Color = FirefoxTheme.colors.actionTertiary,
+            selectedBackgroundColor: Color = FirefoxTheme.colors.textActionPrimary,
+            unselectedBackgroundColor: Color = FirefoxTheme.colors.textActionTertiary,
+        ): PocketStoriesCategoryColors =
```
gabrielluong, line 476:
```suggestion
        ) =
```
I think this might be unnecessary.

app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesComposables.kt
```diff
@@ -408,10 +408,7 @@ private fun Rect.getIntersectPercentage(realSize: IntSize, other: Rect): Float {
  *
  * @param categories The categories needed to be displayed.
  * @param selections List of categories currently selected.
- * @param selectedTextColor Text [Color] when the category is selected.
- * @param unselectedTextColor Text [Color] when the category is not selected.
- * @param selectedBackgroundColor Background [Color] when the category is selected.
- * @param unselectedBackgroundColor Background [Color] when the category is not selected.
+ * @param categoryColors The color set defined by [PocketStoriesCategoryColors] used to style Pocket categories.
  * @param onCategoryClick Callback for when the user taps a category.
  * @param modifier [Modifier] to be applied to the layout.
```
gabrielluong, line 413: Let's fix the order of the @param while we are at.

app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesComposables.kt
```diff
@@ -421,12 +418,9 @@ private fun Rect.getIntersectPercentage(realSize: IntSize, other: Rect): Float {
 fun PocketStoriesCategories(
     categories: List<PocketRecommendedStoriesCategory>,
     selections: List<PocketRecommendedStoriesSelectedCategory>,
-    selectedTextColor: Color? = null,
-    unselectedTextColor: Color? = null,
-    selectedBackgroundColor: Color? = null,
-    unselectedBackgroundColor: Color? = null,
-    onCategoryClick: (PocketRecommendedStoriesCategory) -> Unit,
     modifier: Modifier = Modifier,
```
gabrielluong, line 421: I am curious does Android Studio complain about the `modifier` position here? I believe I have seen it recommend that `modifier` should be the second optional parameter if the composable function accepts a modifier parameter. If it does complain, let's fix it here as well.

app/src/main/java/org/mozilla/fenix/home/pocket/PocketCategoriesViewHolder.kt
```diff
@@ -121,21 +119,18 @@ private fun PocketTopics(
...
     }
 }
 
 @Composable
-@Preview
+@Preview(uiMode = Configuration.UI_MODE_NIGHT_YES)
+@Preview(uiMode = Configuration.UI_MODE_NIGHT_NO)
 private fun PocketCategoriesViewHolderPreview() {
-    FirefoxTheme(theme = Theme.getTheme()) {
+    FirefoxTheme {
```
gabrielluong, line 133: I didn't verify this, but assuming we already have a default value for `theme`. Let's file a followup to clean all of this up (eg, removing all the `(theme = Theme.getTheme())`) if there's still any leftover in the codebase.

app/src/main/java/org/mozilla/fenix/home/pocket/PocketCategoriesViewHolder.kt
```diff
@@ -61,34 +60,36 @@ class PocketCategoriesViewHolder(
...
+        var (selectedTextColor, unselectedTextColor, selectedBackgroundColor, unselectedBackgroundColor) =
+            PocketStoriesCategoryColors.buildColors()
         wallpaperState.composeRunIfWallpaperCardColorsAreAvailable { cardColorLight, cardColorDark ->
             if (isSystemInDarkTheme()) {
-                selectedBackgroundColor = cardColorDark
-                unselectedBackgroundColor = cardColorLight
                 selectedTextColor = FirefoxTheme.colors.textActionPrimary
                 unselectedTextColor = FirefoxTheme.colors.textActionSecondary
+                selectedBackgroundColor = cardColorDark
+                unselectedBackgroundColor = cardColorLight
```
gabrielluong, line 70: I am looking at this and thinking for `PocketStoriesCategoryColors` and `buildColors`, we should just switch up the parameter orders so that the background color are before the text color. Just so we can avoid this reordering these assignments although we probably didn't have to reorder these assignments regardless.

No strong feelings here - you could opt to take no actions as well. It's mostly odd to me that we reordered these assignments to follow the parameter ordering. The main consideration here would be that we're changing git blame history for these particular lines due to the refactoring, but it's not a big deal here since we are very active in this area of the codebase and it's not particularly scary code that we haven't worked with in awhile.

app/src/main/java/org/mozilla/fenix/home/pocket/PocketCategoriesViewHolder.kt
```diff
@@ -103,10 +104,7 @@ class PocketCategoriesViewHolder(
 
 @Composable
 private fun PocketTopics(
-    selectedTextColor: Color? = null,
-    unselectedTextColor: Color? = null,
-    selectedBackgroundColor: Color? = null,
-    unselectedBackgroundColor: Color? = null,
+    categoryColors: PocketStoriesCategoryColors = PocketStoriesCategoryColors.buildColors(),
```
gabrielluong, line 107: Let's reorder the parameters here. We should ensure `categories` and `categoriesSelections` come before the colors since those are the main parameters for this composable compare to the optional parameter of `categoryColors`.

gabrielluong (approved):

app/src/main/java/org/mozilla/fenix/home/pocket/PocketStoriesComposables.kt
MozillaNoah: It complains if it isn't the first optional parameter

gabrielluong: Oh okay, then everything is fine then.

app/src/main/java/org/mozilla/fenix/home/pocket/PocketCategoriesViewHolder.kt
MozillaNoah: We have a default, and yeah, a refactor a few months back added all of the `theme = Theme.getTheme()` calls.

MozillaNoah: Opened: 
- https://github.com/mozilla-mobile/fenix/issues/27854

app/src/main/java/org/mozilla/fenix/home/pocket/PocketCategoriesViewHolder.kt
MozillaNoah: I think that's fair

PR #27884, opened 2022-11-17
No issue: RTL nav bar UI tests

sv-ohorvath:
I wrote a test that needs an RTL language set on the device, and for that, I need to allow Config changes permissions for the app in any variant. 
If GrantPermission is not declared, we can't change the device language for some tests, like following the system default locale.

I found that Kate did this in the past here https://github.com/mozilla-mobile/fenix/pull/3562/files - but she has set that only for the debug version.

I would need to change that to work on other variants as well if we run our tests in Nightly/Beta/Release. I'm not sure that's ok to do in the main manifest file, so I need someone's informed advice & approval.

### Pull Request checklist

### QA

- [ ] **QA Needed**

### GitHub Automation

Used by GitHub Actions.

### DIFF
--- a/app/src/androidTest/java/org/mozilla/fenix/helpers/TestHelper.kt
+++ b/app/src/androidTest/java/org/mozilla/fenix/helpers/TestHelper.kt
@@ -6,6 +6,7 @@
 
 package org.mozilla.fenix.helpers
 
+import android.Manifest
 import android.app.ActivityManager
 import android.app.PendingIntent
 import android.content.ActivityNotFoundException
@@ -42,6 +43,7 @@ import androidx.test.espresso.matcher.ViewMatchers.withParent
 import androidx.test.espresso.matcher.ViewMatchers.withText
 import androidx.test.platform.app.InstrumentationRegistry
 import androidx.test.rule.ActivityTestRule
+import androidx.test.runner.permission.PermissionRequester
 import androidx.test.uiautomator.By
 import androidx.test.uiautomator.UiDevice
 import androidx.test.uiautomator.UiObject
@@ -57,6 +59,7 @@ import org.hamcrest.Matcher
 import org.junit.Assert
 import org.junit.Assert.assertFalse
 import org.junit.Assert.assertTrue
+import org.mozilla.fenix.Config
 import org.mozilla.fenix.HomeActivity
 import org.mozilla.fenix.R
 import org.mozilla.fenix.customtabs.ExternalAppBrowserActivity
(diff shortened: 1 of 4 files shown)


AaronMT:
I'm not sure I understand the need to adjust the Android manifest in addition to using `GrantPermissionRule`.  There might be other consequences of adjusting the permissions here. Would it be better to annotate or wrap/mark the test for debug variant only? Maybe a new custom annotation or other approach.

sv-ohorvath:
> I'm not sure I understand the need to adjust the Android manifest in addition to using `GrantPermissionRule`. There might be other consequences of adjusting the permissions here. Would it be better to annotate or wrap/mark the test for debug variant only? Maybe a new custom annotation or other approach.

If that permission is not declared in the manifest, GrantPermission won't work on Nightly. This test will fail: https://console.firebase.google.com/project/moz-fenix/testlab/histories/bh.80a1a071cfb2770e/matrices/5737493424594287496/executions/bs.56d2c108143718a4/testcases/2/errors

@AaronMT @Mugurell who else do you think I should get to review this to be sure?

Mugurell:
> @AaronMT @Mugurell who else do you think I should get to review this to be sure?

@csadilek @jonalmeida @gabrielluong

app/src/main/AndroidManifest.xml
```diff
@@ -29,6 +29,10 @@
     <uses-permission android:name="android.permission.QUERY_ALL_PACKAGES"
         tools:ignore="QueryAllPackagesPermission" />
 
+    <!-- Allows changing locales -->
+    <uses-permission android:name="android.permission.CHANGE_CONFIGURATION"
+        tools:ignore="ProtectedPermissions" />
```
jonalmeida, line 34: The warning that we're suppressing here has an important one: "Permission is only granted to system apps"

We definitely can't submit fenix to the Play store if we add this to any of our release variants.

Since we run our UI tests on debug builds IIRC. Can we add this permission to the app/src/debug/AndroidManifest.xml so that it's limited to only that variant as mentioned in the comments above (assuming we have no options and need to add this permission).

jonalmeida (dismissed):
Apologises, I missed this notification - I left a comment below that's a similar solution to the ones above.

Landing this patch as is will result us in having our app denied by the Play Store.

app/src/main/AndroidManifest.xml
sv-ohorvath: Sure @jonalmeida! Thank you, that's what I wanted to know.

sv-ohorvath:
I’ve made the corrections and this is now ready for another review. Thank you, everyone!

app/src/androidTest/java/org/mozilla/fenix/helpers/TestHelper.kt
```diff
@@ -389,20 +392,31 @@ object TestHelper {
 
     /**
      * Changes the default language of the entire device, not just the app.
+     * Only works on the Debug version, not allowed to change system settings in other production builds.
```
AaronMT, line 395: I'd say, "Runs on Debug variant as we don't want to adjust Release permission manifests"

app/src/androidTest/java/org/mozilla/fenix/helpers/TestHelper.kt
```diff
@@ -389,20 +392,31 @@ object TestHelper {
...
      * Runs the test in its testBlock.
-     * Cleans up and sets the default locale after it's are done.
+     * Cleans up and sets the default locale after it's done.
      */
     fun runWithSystemLocaleChanged(locale: Locale, testRule: ActivityTestRule<HomeActivity>, testBlock: () -> Unit) {
-        val defaultLocale = Locale.getDefault()
+        if (BuildConfig.DEBUG) {
```
AaronMT, line 400: I looked into better handling build variants. Ideally we should have set-up our structure to place variant tests inside the variant folder (`src/variant`), as within `src/androidTest/`root they are shared across variants and that's probably not what we want these days.

https://developer.android.com/studio/test/advanced-test-setup#create-instrumented-test-for-build-variant

I'm not sure if there's a way to filter annotate the test with this variant check so it its not executed by the runner 

I looked at https://developer.android.com/reference/androidx/test/filters/package-summary and don't see anything. I guess this is suffice?

AaronMT (approved):
Looks OK to me pending variant check

app/src/androidTest/java/org/mozilla/fenix/helpers/TestHelper.kt
jonalmeida: I think you want `Config.isDebug`.

`BuildConfig.DEBUG` is whether the app is "[debuggable](https://developer.android.com/guide/topics/manifest/application-element#debug)" and not the debug _variant_. It's confusing unfortunately.

app/src/androidTest/java/org/mozilla/fenix/helpers/TestHelper.kt
sv-ohorvath: done, thanks

app/src/androidTest/java/org/mozilla/fenix/helpers/TestHelper.kt
sv-ohorvath: done. Thanks!

PR #27952, opened 2022-11-22
For #27950: add first week days of use growth data

### LINKED ISSUE #27950: Add first week days of use growth data
See the epic the Jira ticket is linked to for more details

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-22036)

MatthewTighe:
### Pull Request checklist

### QA

- [ ] **QA Needed**

### GitHub Automation
Fixes #27950

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/components/metrics/Event.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/metrics/Event.kt
@@ -31,5 +31,10 @@ sealed class Event {
          * Event recording the first time a URI is loaded in Firefox in a 24 hour period.
          */
         object FirstUriLoadForDay : GrowthData("ja86ek")
+
+        /**
+         * Event recording the first time Firefox is used 3 days in a row in the first week of install.
+         */
+        object FirstWeekSeriesActivity : GrowthData("20ay7u")
     }
 }
--- a/app/src/main/java/org/mozilla/fenix/components/metrics/MetricsStorage.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/metrics/MetricsStorage.kt
@@ -12,6 +12,9 @@ import mozilla.components.support.utils.ext.getPackageInfoCompat
 import org.mozilla.fenix.ext.settings
 import org.mozilla.fenix.nimbus.FxNimbus
 import org.mozilla.fenix.utils.Settings
+import java.text.SimpleDateFormat
+import java.util.Calendar
+import java.util.Locale
 
 /**
  * Interface defining functions around persisted local state for certain metrics.
@@ -33,13 +36,20 @@ internal class DefaultMetricsStorage(
     private val settings: Settings,
     private val checkDefaultBrowser: () -> Boolean,
     private val shouldSendGenerally: () -> Boolean = { shouldSendGenerally(context) },
+    private val getInstalledTime: () -> Long = { getInstalledTime(context) },
     private val dispatcher: CoroutineDispatcher = Dispatchers.IO,
 ) : MetricsStorage {
+
+    private val dateFormatter = SimpleDateFormat("yyyy-MM-dd", Locale.US)
+
     /**
      * Checks local state to see whether the [event] should be sent.
      */
     override suspend fun shouldTrack(event: Event): Boolean =
         withContext(dispatcher) {
+            // The side-effect of storing days of use needs to happen during the first two days after
+            // install, which would normally be skipped by shouldSendGenerally.
+            updateDaysOfUse()
             shouldSendGenerally() && when (event) {
                 Event.GrowthData.SetAsDefault -> {
                     !settings.setAsDefaultGrowthSent && checkDefaultBrowser()
@@ -49,6 +59,9 @@ internal class DefaultMetricsStorage(
                 }
                 Event.GrowthData.FirstUriLoadForDay -> {
                     settings.uriLoadGrowthLastSent.hasBeenMoreThanDaySince()
+                }
+                Event.GrowthData.FirstWeekSeriesActivity -> {
+                    shouldTrackFirstWeekActivity()
                 }
             }
         }
@@ -64,21 +77,80 @@ internal class DefaultMetricsStorage(
             Event.GrowthData.FirstUriLoadForDay -> {
                 settings.uriLoadGrowthLastSent = System.currentTimeMillis()
             }
+            Event.GrowthData.FirstWeekSeriesActivity -> {
+                settings.firstWeekSeriesGrowthSent = true
+            }
         }
     }
 
+    private fun updateDaysOfUse() {
+        val daysOfUse = settings.firstWeekDaysOfUseGrowthData
+        val currentDate = Calendar.getInstance(Locale.US)
+        val currentDateString = dateFormatter.format(currentDate.time)
+        if (currentDate.timeInMillis.withinFirstWeek() && daysOfUse.none { it == currentDateString }) {
+            settings.firstWeekDaysOfUseGrowthData = daysOfUse + currentDateString
+        }
+    }
+
+    private fun shouldTrackFirstWeekActivity(): Boolean = Result.runCatching {
+        if (!System.currentTimeMillis().withinFirstWeek() || settings.firstWeekSeriesGrowthSent) {
+            return false
+        }
+
+        val daysOfUse = settings.firstWeekDaysOfUseGrowthData.map {
+            dateFormatter.parse(it)
+        }.sorted()
+
+        // This loop will check whether the existing list of days of use, combined with the
+        // current date, contains any periods of 3 days of use in a row.
+        for (idx in daysOfUse.indices) {
+            if (idx + 1 > daysOfUse.lastIndex || idx + 2 > daysOfUse.lastIndex) {
+                continue
+            }
+
+            val referenceDate = daysOfUse[idx]!!.time.toCalendar()
+            val secondDateEntry = daysOfUse[idx + 1]!!.time.toCalendar()
+            val thirdDateEntry = daysOfUse[idx + 2]!!.time.toCalendar()
+            val oneDayAfterReference = referenceDate.createNextDay()
+            val twoDaysAfterReference = oneDayAfterReference.createNextDay()
+
+            if (oneDayAfterReference == secondDateEntry && thirdDateEntry == twoDaysAfterReference) {
+                return true
+            }
+        }
+        return false
+    }.getOrDefault(false)
+
     private fun Long.hasBeenMoreThanDaySince(): Boolean =
         System.currentTimeMillis() - this > dayMillis
+
+    private fun Long.toCalendar(): Calendar = Calendar.getInstance(Locale.US).also { calendar ->
+        calendar.timeInMillis = this
+    }
+
+    private fun Long.withinFirstWeek() = this < getInstalledTime() + fullWeekMillis
+
+    private fun Calendar.createNextDay() = (this.clone() as Calendar).also { calendar ->
+        calendar.add(Calendar.DAY_OF_MONTH, 1)
+    }
 
     companion object {
         private const val dayMillis: Long = 1000 * 60 * 60 * 24
         private const val windowStartMillis: Long = dayMillis * 2
         private const val windowEndMillis: Long = dayMillis * 28
 
+        // Note this is 8 so that recording of FirstWeekSeriesActivity happens throughout the length
+        // of the 7th day after install
+        private const val fullWeekMillis: Long = dayMillis * 8
+
+        /**
+         * Determines whether events should be tracked based on some general criteria:
+         * - user has installed as a result of a campaign
+         * - user is within 2-28 days of install
+         * - tracking is still enabled through Nimbus
+         */
         fun shouldSendGenerally(context: Context): Boolean {
-            val installedTime = context.packageManager
-                .getPackageInfoCompat(context.packageName, 0)
-                .firstInstallTime
+            val installedTime = getInstalledTime(context)
             val timeDifference = System.currentTimeMillis() - installedTime
             val withinWindow = timeDifference in windowStartMillis..windowEndMillis
 
@@ -86,5 +158,9 @@ internal class DefaultMetricsStorage(
                 FxNimbus.features.growthData.value().enabled &&
                 withinWindow
         }
+
+        fun getInstalledTime(context: Context): Long = context.packageManager
+            .getPackageInfoCompat(context.packageName, 0)
+            .firstInstallTime
     }
 }
--- a/app/src/main/java/org/mozilla/fenix/components/metrics/MetricsMiddleware.kt
+++ b/app/src/main/java/org/mozilla/fenix/components/metrics/MetricsMiddleware.kt
@@ -24,6 +24,7 @@ class MetricsMiddleware(
         is AppAction.ResumedMetricsAction -> {
             metrics.track(Event.GrowthData.SetAsDefault)
             metrics.track(Event.GrowthData.FirstAppOpenForDay)
+            metrics.track(Event.GrowthData.FirstWeekSeriesActivity)
         }
         else -> Unit
     }
--- a/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
+++ b/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
@@ -1444,4 +1444,14 @@ class Settings(private val appContext: Context) : PreferencesHolder {
         key = appContext.getPreferenceKey(R.string.pref_key_growth_uri_load_last_sent),
         default = 0,
     )
+
+    var firstWeekSeriesGrowthSent by booleanPreference(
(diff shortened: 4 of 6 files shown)


MozillaNoah:
I noticed that the Github issue and Jira issue are both blank. If there is a root Jira issue this was spun off of, could that be linked to the auto-generated one as a "fulfilled by"? Just want to make sure the paper trail is all synced up.

MatthewTighe:
> I noticed that the Github issue and Jira issue are both blank. If there is a root Jira issue this was spun off of, could that be linked to the auto-generated one as a "fulfilled by"? Just want to make sure the paper trail is all synced up.

The autogenerated Jira is now linked to the epic that has details. I'll mention as much in the Github issue as well

app/src/main/java/org/mozilla/fenix/components/metrics/MetricsStorage.kt
```diff
@@ -64,27 +77,80 @@ internal class DefaultMetricsStorage(
...
+
+        // This loop will check whether the existing list of days of use, combined with the
+        // current date, contains any periods of 3 days of use in a row. It's mostly this ugly
+        // because of the inherent mutability of the Calendar type, but hopefully we can rip
+        // this code out soon™ anyway
+        for (idx in daysOfUse.indices) {
+            if (idx + 1 > daysOfUse.lastIndex || idx + 2 > daysOfUse.lastIndex) {
+                continue
+            }
+            val day1 = daysOfUse[idx]!!.time.toCalendar()
```
MozillaNoah, line 112: are the `!!` necessary?

app/src/main/java/org/mozilla/fenix/components/metrics/MetricsStorage.kt
```diff
@@ -64,27 +77,80 @@ internal class DefaultMetricsStorage(
...
+            dateFormatter.parse(it)
+        }.sorted()
+
+        // This loop will check whether the existing list of days of use, combined with the
+        // current date, contains any periods of 3 days of use in a row. It's mostly this ugly
+        // because of the inherent mutability of the Calendar type, but hopefully we can rip
+        // this code out soon™ anyway
+        for (idx in daysOfUse.indices) {
+            if (idx + 1 > daysOfUse.lastIndex || idx + 2 > daysOfUse.lastIndex) {
+                continue
```
MozillaNoah, line 110: thoughts on using the `continue` vs wrapping the entire block inside the `if` instead?

app/src/main/java/org/mozilla/fenix/components/metrics/MetricsStorage.kt
```diff
@@ -64,27 +77,80 @@ internal class DefaultMetricsStorage(
...
+
+    @Suppress("MagicNumber")
+    private fun Long.withinFirstWeek() = this < getInstalledTime() + dayMillis * 8
+
     companion object {
         private const val dayMillis: Long = 1000 * 60 * 60 * 24
         private const val windowStartMillis: Long = dayMillis * 2
         private const val windowEndMillis: Long = dayMillis * 28
 
         fun shouldSendGenerally(context: Context): Boolean {
```
MozillaNoah, line 142: not necessary for the PR, but could this function later get some documentation? Its purpose/result seems ambiguous.

app/src/main/java/org/mozilla/fenix/components/metrics/MetricsStorage.kt
```diff
@@ -64,27 +77,80 @@ internal class DefaultMetricsStorage(
...
+            }
         }
     }
 
+    private fun updateDaysOfUse() {
+        val daysOfUse = settings.firstWeekDaysOfUseGrowthData
+        val currentDate = Calendar.getInstance()
+        val asString = dateFormatter.format(currentDate.time)
+        if (currentDate.timeInMillis.withinFirstWeek() && daysOfUse.none { it == asString }) {
+            settings.firstWeekDaysOfUseGrowthData = daysOfUse.plus(asString)
```
MozillaNoah, line 91:
```suggestion
            settings.firstWeekDaysOfUseGrowthData = daysOfUse + asString
```

app/src/main/java/org/mozilla/fenix/components/metrics/MetricsStorage.kt
```diff
@@ -64,27 +77,80 @@ internal class DefaultMetricsStorage(
...
+            Event.GrowthData.FirstWeekSeriesActivity -> {
+                settings.firstWeekSeriesGrowthSent = true
+            }
         }
     }
 
+    private fun updateDaysOfUse() {
+        val daysOfUse = settings.firstWeekDaysOfUseGrowthData
+        val currentDate = Calendar.getInstance()
+        val asString = dateFormatter.format(currentDate.time)
```
MozillaNoah, line 89:
```suggestion
        val currentDateString = dateFormatter.format(currentDate.time)
```

app/src/main/java/org/mozilla/fenix/components/metrics/MetricsStorage.kt
```diff
@@ -33,13 +36,20 @@ internal class DefaultMetricsStorage(
     private val settings: Settings,
     private val checkDefaultBrowser: () -> Boolean,
     private val shouldSendGenerally: () -> Boolean = { shouldSendGenerally(context) },
+    private val getInstalledTime: () -> Long = { getInstalledTime(context) },
```
MozillaNoah, line 39: these are neat lambdas. Are the default lambdas overridden anywhere to necessitate their parameterization?

app/src/main/java/org/mozilla/fenix/components/metrics/Event.kt
```diff
@@ -31,5 +31,10 @@ sealed class Event {
          * Event recording the first time a URI is loaded in Firefox in a 24 hour period.
          */
         object FirstUriLoadForDay : GrowthData("ja86ek")
+
+        /**
+         * Event recording the first time Firefox is used 3 days in a row in the first week of install.
+         */
+        object FirstWeekSeriesActivity : GrowthData("20ay7u")
```
MozillaNoah, line 38: are the strings codes from the Adjust SDK?

app/src/main/java/org/mozilla/fenix/components/metrics/MetricsStorage.kt
```diff
@@ -64,27 +77,80 @@ internal class DefaultMetricsStorage(
...
+    }
+
+    private fun shouldTrackFirstWeekActivity(): Boolean = Result.runCatching {
+        if (!System.currentTimeMillis().withinFirstWeek() || settings.firstWeekSeriesGrowthSent) {
+            return false
+        }
+
+        val daysOfUse = settings.firstWeekDaysOfUseGrowthData.map {
+            dateFormatter.parse(it)
+        }.sorted()
```
MozillaNoah, line 102: would we be able to save ourselves some work if we always save this data as sorted?

app/src/main/java/org/mozilla/fenix/components/metrics/MetricsStorage.kt
```diff
@@ -64,27 +77,80 @@ internal class DefaultMetricsStorage(
...
+                continue
+            }
+            val day1 = daysOfUse[idx]!!.time.toCalendar()
+            val day2 = daysOfUse[idx + 1]!!.time.toCalendar()
+            val day3 = daysOfUse[idx + 2]!!.time.toCalendar()
+            var areSequential = true
+            day1.add(Calendar.DAY_OF_MONTH, 1)
+            areSequential = areSequential && day1 == day2
+            day1.add(Calendar.DAY_OF_MONTH, 1)
+            areSequential = areSequential && day1 == day3
```
MozillaNoah, line 119:
```suggestion
            val referenceDate = daysOfUse[idx]!!.time.toCalendar()
            val oneDayAfterReference = referenceDate.also { it.add(Calendar.DAY_OF_MONTH, 1) }
            val twoDaysAfterReference = referenceDate.also { it.add(Calendar.DAY_OF_MONTH, 2) }
            val secondDateEntry = daysOfUse[idx + 1]!!.time.toCalendar()
            val thirdDateEntry = daysOfUse[idx + 2]!!.time.toCalendar()
            var areSequential = true
            areSequential = areSequential && oneDayAfterReference == secondDateEntry
            areSequential = areSequential && twoDaysAfterReference == thirdDateEntry
```

Thoughts on rewording some things to make it a bit clearer what's being compared?

MozillaNoah (commented):
first pass. Mainly have questions for my edification and for some potential clean-up. The PR looks sound though, given the constraints of this project. There also isn't a super clean way to do a sequential date check.

app/src/main/java/org/mozilla/fenix/components/metrics/MetricsStorage.kt
MatthewTighe: They are overridden when under test

app/src/main/java/org/mozilla/fenix/components/metrics/Event.kt
MatthewTighe: Yes. These codes are generated per event/project combo which means they need to be different during testing. If we continue down this path we should invest in a dynamic system for inserting them at build time or something similar

app/src/main/java/org/mozilla/fenix/components/metrics/MetricsStorage.kt
MatthewTighe: Since SharedPreferences are saved as `Set`s I'm not sure we would have strong guarantees that the data would remain sorted. By sorting before use we can guarantee the dates are in the order we expect

app/src/main/java/org/mozilla/fenix/components/metrics/MetricsStorage.kt
MatthewTighe: According to the compiler 😅 `dateFormatter.parse(it)` returns the platform type `Date!`

MatthewTighe: This is also why this function is wrapped in a `runCatching` block, though

MatthewTighe: I thought about leaving a comment to that effect. Would that be helpful?

app/src/main/java/org/mozilla/fenix/components/metrics/MetricsStorage.kt
MatthewTighe: I can change it if you'd prefer. I just try to avoid indentation where possible in case of awkward line breaks

app/src/main/java/org/mozilla/fenix/components/metrics/MetricsStorage.kt
MatthewTighe: I do like the name changes, I'll work on updating to something similar. The real issue here is that `Calendar::add` mutates the underlying object instead of returning a new one with the updated date, so things like `referenceDate.also { it.add(Calendar.DAY_OF_MONTH, 1) }` don't actually work. 

I worked around that with a `copy` extension in the test file but thought it might be bloat-y in the actual code. Do you think it's worth including here as well?

MatthewTighe:
Made a few updates based on feedback

app/src/main/java/org/mozilla/fenix/components/metrics/MetricsStorage.kt
MatthewTighe: I've included an extension function to clean up the creation of a new Calendar object for the subsequent day

app/src/main/java/org/mozilla/fenix/components/metrics/MetricsStorage.kt
```diff
@@ -64,27 +77,87 @@ internal class DefaultMetricsStorage(
...
             }
+            Event.GrowthData.FirstWeekSeriesActivity -> {
+                settings.firstWeekSeriesGrowthSent = true
+            }
         }
     }
 
+    private fun updateDaysOfUse() {
+        val daysOfUse = settings.firstWeekDaysOfUseGrowthData
+        val currentDate = Calendar.getInstance()
```
jonalmeida, line 88: Should these Calendar instances also be using the same locale as the time format (Locale.US)?

```suggestion
        val currentDate = Calendar.getInstance(Locale.US)
```

jonalmeida (approved):
Let a few comments to consider. 👍

app/src/main/java/org/mozilla/fenix/components/metrics/MetricsStorage.kt
```diff
@@ -64,27 +77,80 @@ internal class DefaultMetricsStorage(
...
+
     private fun Long.hasBeenMoreThanDaySince(): Boolean =
         System.currentTimeMillis() - this > dayMillis
 
+    private fun Long.toCalendar(): Calendar = Calendar.getInstance().also {
+        it.timeInMillis = this
+    }
+
+    @Suppress("MagicNumber")
+    private fun Long.withinFirstWeek() = this < getInstalledTime() + dayMillis * 8
```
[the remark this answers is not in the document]
MatthewTighe: This is actually because we still want to record the event through the entirety of day 7, so a comment seems helpful 😄

app/src/main/java/org/mozilla/fenix/components/metrics/MetricsStorage.kt
```diff
@@ -64,27 +77,80 @@ internal class DefaultMetricsStorage(
...
+            day1.add(Calendar.DAY_OF_MONTH, 1)
+            areSequential = areSequential && day1 == day2
+            day1.add(Calendar.DAY_OF_MONTH, 1)
+            areSequential = areSequential && day1 == day3
+            if (areSequential) {
+                return true
+            }
+        }
+        return false
+    }.getOrDefault(false)
```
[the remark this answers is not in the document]
MatthewTighe: The `runCatching` was intended to ease concerns about the `!!` unwrapping of the `Date!` returned by `formatter.parse()`. 

If you inspect that definition you will find that the platform type is not actually nullable (just missing the `@NonNull` annotation) but it can throw when a date can't be parsed. Since we fully control the behavior of packaging these to disk, I think that reporting unexpected behavior here is probably unnecessary

app/src/main/java/org/mozilla/fenix/components/metrics/MetricsStorage.kt
MozillaNoah: I admittedly didn't notice the `runCatching` block 😅 

Since the block is wrapped, the `!!` unwrapping is unnecessary, but I could see the argument for keeping them so in the future we know that that operation is volatile. I don't feel super strongly though, especially if this has _actually_ has a short shelf life.

app/src/main/java/org/mozilla/fenix/components/metrics/MetricsStorage.kt
MozillaNoah: The hesitation to use `continue` is mostly a personal coding preference of mine. It would have been more intuitive if it had been possible to do an iOS equivalent of `guard` with a variable and just return/continue if a variable couldn't be initialized, but I tried and couldn't get it to work. 

I think both are readable; I just think `continue` can make _some_ blocks of code less intuitive. I think it's fine here.

app/src/main/java/org/mozilla/fenix/components/metrics/MetricsStorage.kt
MozillaNoah: That's my bad. I misinterpreted the Android Studio tooling when trying to cross-check its usages. Carry on

app/src/main/java/org/mozilla/fenix/components/metrics/MetricsStorage.kt
MozillaNoah: this is non-blocking, but I did find a handy function we could use if we wanted to relocate the sorting, `sortedSetOf(...)`

https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/sorted-set-of.html

I admittedly haven't spent a ton of time thinking if one location is optimal over the other, or if it matters.

app/src/main/java/org/mozilla/fenix/components/metrics/MetricsStorage.kt
MozillaNoah: This looks much better!
image

MozillaNoah (approved):

app/src/main/java/org/mozilla/fenix/components/metrics/MetricsStorage.kt
MatthewTighe: Oh this function definitely looks useful for the future! In the interest of avoiding bike-shedding an optimization of this size too much I think I will leave it for now. The size of the list will never be larger than a few elements.

MatthewTighe:
@mergifyio backport releases_v108.0.0

PR #27993, opened 2022-11-28
Bug 1802817 - Update Fenix to Gradle 7.6

rvandermeulen:


MatthewTighe:
Tested this with several gradle workflows and everything seems to be in working order except for autoPublish. Looking around briefly that is likely just a side-effect of the repo changes though.

Is there anything specific, process-wise, for reviewing build dependency updates like this? If not, I'm happy to approve.

gabrielluong:
> Tested this with several gradle workflows and everything seems to be in working order except for autoPublish. Looking around briefly that is likely just a side-effect of the repo changes though.
> 
> Is there anything specific, process-wise, for reviewing build dependency updates like this? If not, I'm happy to approve.

Just for my understanding does this gradle change break the local autopublish workflow for working with AC? We haven't made any changes to Fenix yet as part of the monorepo project. If the gradle update does break the local autopublish workflow for working with AC, I believe that would be consideration to not land this.

There is no specific process for reviewing build dependency that I am aware of. However, due diligence to ensure the gradle changes will work as expected is always appreciated.

MatthewTighe:
> Just for my understanding does this gradle change break the local autopublish workflow for working with AC? We haven't made any changes to Fenix yet as part of the monorepo project. If the gradle update does break the local autopublish workflow for working with AC, I believe that would be consideration to not land this.

I think it may be broken, but I don't think it's a result of this Gradle update. The `publish_to_maven_local_if_modified.py` script wants to run from repo root, and I think that may be getting confused now that A-C is a subdirectory of firefox-android. I wasn't able to fully debug it yet though, so I can't say for sure

gabrielluong:
> > Just for my understanding does this gradle change break the local autopublish workflow for working with AC? We haven't made any changes to Fenix yet as part of the monorepo project. If the gradle update does break the local autopublish workflow for working with AC, I believe that would be consideration to not land this.
> 
> I think it may be broken, but I don't think it's a result of this Gradle update. The `publish_to_maven_local_if_modified.py` script wants to run from repo root, and I think that may be getting confused now that A-C is a subdirectory of firefox-android. I wasn't able to fully debug it yet though, so I can't say for sure

Gotcha, no worries from me if the autopublish workflow is unrelated to the Gradle update. We won't need to rely on the autpublish workflow once we migrate Fenix to the monorepo.

kycn:
Auto-publish was buggy at first when we migrated the repo, but it should have already been fixed. At least, after that fix, I did not have any problems while using autopublish with the new path of android-components package in my local.properties file.

csadilek:
Yes, auto-publish works fine on current main. Just tested again. You may need to adjust your path if it's not currently working, see: https://github.com/mozilla-mobile/fenix/pull/27665

Also, even after Fenix moves to the Monorepo, auto publish is still needed for Application Services. So, please don't land any change that will break it.

MatthewTighe:
Turns out I'd been missing an environment variable since my computer wipe 😅 autoPublish and other common Gradle tasks work for me.

MatthewTighe (approved):

rvandermeulen:
FWIW, I'm a bit hesitant to land this update (and the Focus bump) until we better understand the issues with the AC bump related to the refactored Dependencies plugin. My assumption is that Fenix and Focus will also end up be susceptible to the same issue when they migrate to the monorepo and I don't want to add more complication into that project needlessly.

csadilek:
@rvandermeulen We have to close all PRs for the Monorepo migration, and since we won't land this before Tues., I am closing this one too. Let's reopen and discuss the upgrade on the Monorepo! Thanks.

mcarare:
This will be completed in https://github.com/mozilla-mobile/firefox-android/pull/732, along with Focus and AC upgrades and it will include AGP upgrade.

PR #28024, opened 2022-11-29
Bug 1800700 - Add default search suggest header

sarah541:
| | |
|----|----|
|Screenshot_20221129-141015|Screenshot_20221129-141038|

### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation

Used by GitHub Actions.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/search/awesomebar/AwesomeBarView.kt
+++ b/app/src/main/java/org/mozilla/fenix/search/awesomebar/AwesomeBarView.kt
@@ -10,6 +10,7 @@ import androidx.core.graphics.BlendModeCompat.SRC_IN
 import androidx.core.graphics.drawable.toBitmap
 import mozilla.components.browser.state.search.SearchEngine
 import mozilla.components.browser.state.state.searchEngines
+import mozilla.components.browser.state.state.selectedOrDefaultSearchEngine
 import mozilla.components.concept.awesomebar.AwesomeBar
 import mozilla.components.concept.engine.EngineSession
 import mozilla.components.feature.awesomebar.provider.BookmarksStorageSuggestionProvider
@@ -176,6 +177,7 @@ class AwesomeBarView(
                     BrowsingMode.Normal -> false
                     BrowsingMode.Private -> true
                 },
+                suggestionsHeader = getSearchEngineSuggestionsHeader(),
             )
 
         defaultSearchActionProvider =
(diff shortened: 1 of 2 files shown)


boek (approved):
Looks good! Do we also need to show the header when selecting a search engine shortcut?

sarah541:
> Looks good! Do we also need to show the header when selecting a search engine shortcut?

Hey Jeff. Thanks for  the approval! No, we do not show header when selecting shortcut since this is only for default search engine!. Also this PR will be merged once the PR on the AC gets merged : https://github.com/mozilla-mobile/firefox-android/pull/234. Still having a few rounds of review there and I'll add the approach we will be taking in that PR

app/src/main/java/org/mozilla/fenix/search/awesomebar/AwesomeBarView.kt
```diff
@@ -207,6 +210,25 @@ class AwesomeBarView(
         searchSuggestionProviderMap = HashMap()
     }
 
+    private fun getSearchEngineSuggestionsHeader(): String? {
+        val searchState = activity.components.core.store.state.search
+        var searchEngine = searchState.selectedOrDefaultSearchEngine?.name
```
kycn, line 215: nit - maybe injecting components as param to this method and using one variable would be enough as well :)

kycn (approved):
LGTM! 🎉

gabrielluong (dismissed):
Just wanted to note that we have discussed on Slack yesterday about adding a `enum` in Fenix to have a single source of truth for indicating the provider priority order. This was also independently pointed out by @kycn so I definitely appreciate that this seems to be a common agreed upon pattern. See https://github.com/mozilla-mobile/firefox-android/pull/234/files#r1036865644.

To expand further on the AC side https://github.com/mozilla-mobile/firefox-android/pull/234, the suggestion priorities were hardcoded, which is something we want to avoid because that would dictate all consumer behaviour and goes against the generic component philosophy of AC. Currently, we have a bunch of classes that implements `AwesomeBar.SuggestionProvider` across multiple files that have a specified priority from 1-6. Because everything is spread across multiple file, it’s not very easy to see from a high level what the order of everything is, and that would be a bit sad for developer ergonomics. For example, (1) what happens if someone unfamiliar with the code needs to add a new implementation of `AwesomeBar.SuggestionProvider` and doesn’t understand that priority needs to follow a certain order and doesn’t know we already have 6 other `AwesomeBar.SuggestionProvider` and then randomly puts the `priority = 1` in their implementation or (2) what if in some situations that we don’t show a particular `AwesomeBar.SuggestionProvider` and in those situation we need a different priority order of things because we decided we didn’t want to show the bookmark suggestions. Furthermore, removing a single `AwesomeBar.SuggestionProvider` implementation will mean we need to potentially reorder all of these default values (eg, what happens if somebody removes the class with `priority = 1`, I mean it should be fine, but what if product said the order of everything also had to change along with the removal).

PR #28063, opened 2022-12-04
Bug 1799996 - Speculative fix checks CFRPopup.anchor is attached to window

jonalmeida:
We're seeing crashes in the Fenix copy of the `CFRPopup` that are happening when our activity has already been destroyed, and then we are trying to add a view to it with the WindowManager.

Our guess from looking at the traces are that the
`BrowserToolbarCFRPresenter` is receiving the `collect call on the main thread, but further down the event loop. As well as the Runnable in `CFRPopup#show` that is posting to the end event loop - both of these calls are putting us in a state where the activity we want has already been destroyed before we get to the moment where we want to use it.

This is a crash that is difficult to synthesize or write a test for. As a result, our fix is rather speculative but still straight-forward and safe to uplift, should it work.

There are three remaining follow-ups to do when fix is confirmed:
1. There is a CFRPopup that was upstreamed which is based on the Fenix implementation. We should upstream this patch to that component as well.
2. We should remove this copy of the CFRPopup and use the upstreamed version to avoid confusion between the two copies.
3. With this change, we are now gracefully failing when our users get into this state, however our telemtry and onboarding flows are unaware that the CFR was never shown. We need to to correct telemetry for `TrackingProtection.tcpCfrShown` to stop incorrectly reporting false positives and reset our `shouldShowTotalCookieProtectionCFR` pref to false so that we can try again at the appropriate time.

### Pull Request checklist

### QA

- [ ] **QA Needed**

### GitHub Automation

Used by GitHub Actions.


mavduevskiy (approved):
looking safe to 🚢

app/src/main/java/org/mozilla/fenix/compose/cfr/CFRPopup.kt
```diff
@@ -72,6 +72,19 @@ class CFRPopup(
...
+            // When we're in this Runnable, the 'show' method might have been called right before
+            // the activity is no longer attached to the WindowManager. When we get to calling
+            // the CFRPopupFullscreenLayout#show method below, we are now trying to attach the View
+            // with the WindowManager that has an unusable Activity.
+            //
+            // To protect against this, within this same Runnable, we check if the anchor view is
+            // safe to use before continuing.
+            //
+            // See: https://bugzilla.mozilla.org/show_bug.cgi?id=1799996
+            if (anchor.context == null && !anchor.isAttachedToWindow) {
```
Mugurell, line 84: The context would probably always be valid here.
Maybe we can use an `or` check instead?

Testing with manually detaching the anchor and a small delay I'm getting another exception - `java.lang.IllegalStateException: ViewTreeLifecycleOwner not found from org.mozilla.fenix.compose.cfr.CFRPopupFullscreenLayout`.

We could also just check [this token](https://github.com/mozilla-mobile/fenix/blob/8ed3bf704f8d8b9b2e3b9142c994a08c2ad33cce/app/src/main/java/org/mozilla/fenix/compose/cfr/CFRPopupFullscreenLayout.kt#L354) for null and abort if wanting to focus on the `BadTokenException` but checking if the anchor is attached earlier in code seems better - better visibility, avoid extra calls until the abort.

Can the `show` method be updated to return a Boolean so that callers will know when a call to show a CFR failed and try it again later / only persist in shared prefs that the CFR was shown only if this method returned `true`?

Mugurell (commented):
Would've hoped that running everything on `Main` would be enough but maybe there is a tricky timing issue somewhere.

app/src/main/java/org/mozilla/fenix/compose/cfr/CFRPopup.kt
jonalmeida: > Testing with manually detaching the anchor and a small delay I'm getting another exception - java.lang.IllegalStateException: ViewTreeLifecycleOwner not found from org.mozilla.fenix.compose.cfr.CFRPopupFullscreenLayout.

Yeah, it was quite hard to test this manually. With some suggestions from Christian, we tried:
- Adding a delay before adding the view (so that we could leave the fragment).
- Removing the anchor, delaying, adding the view.
- Launching another Activity in the Runnable before adding the view.

In all those cases we weren't able to reproduce the same stack trace. 🥲 

> The context would probably always be valid here.
> Maybe we can use a `logical OR - ||` check instead?

That was my suspicion as well about the context. I figured the `isAttachedToWindow` is enough, but alone it could have made an assumption that the WindowManager's context is still accessible. This is speculative, so I was trying to be overcautious. 😄 The logical OR sounds better for that reason.

jonalmeida:
@Mergifyio backport releases_v108

We've got approval to uplift this bugfix to 108 - it can go out with any future version of 108 (no need to respin).

jonalmeida:
@Mergifyio backport releases_v108.0.0

PR #18869, opened 2021-04-08
For #7924: Added an undo snackbar when a pinned site is removed

priyanka0906:
fixes: #7924
### Pull Request checklist

- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features. In addition, it includes a screenshot of a successful [accessibility scan](https://play.google.com/store/apps/details?id=com.google.android.apps.accessibility.auditor&hl=en_US) to ensure no new defects are added to the product.

### To download an APK when reviewing a PR:
1. click on Show All Checks,
2. click Details next to "Taskcluster (pull_request)" after it appears and then finishes with a green checkmark,
3. click on the "Fenix - assemble" task, then click "Run Artifacts".
4. the APK links should be on the left side of the screen, named for each CPU architecture

WhatsApp Image 2021-04-08 at 11 03 09 PM

Fixes #7924
Fixes #7924
Fixes #7924
Fixes #7924

PR #28079, opened 2022-12-05
CI For #18869

boek:


gabrielluong (approved):
It might be difficult to land the original PR at times. I am carrying over the approval in case it will help ease the process.

PR #28100, opened 2022-12-06
Bug 1804107 - Fixed configuration change crash when BrowserFragment UI not initialized

### LINKED ISSUE #27540: [Bug]: search bar becomes very small after fold from large to small screen
### Steps to reproduce

1. Obtain foldable device (occurs 100% of the time on fold 4)
2. Open Firefox in unfolded mode and open a website (only happens if app is started in unfolded mode)
3. Close phone
4. Search bar is now the wrong size

### Expected behaviour

The search bar should resize to the proper size as if app is opened in folded mode 

### Actual behaviour

The searchbar is very small, with the lock icon detailing website security taking up most of the space. Upon clicking on visible part of searchbar (not very easy), original size and function is restored and now works properly when folded

### Device name

Samsung Galaxy Z Fold 4

### Android version

Android 12L/13

### Firefox release type

Firefox

### Firefox version

106.1.0

### Device logs

_No response_

### Additional information

_No response_

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21867)

zmckenney:
To fix [Bugzilla 1804107](https://bugzilla.mozilla.org/show_bug.cgi?id=1804107) for when the BrowserFragment configuration change occurs before the UI has been drawn and the toolbar has been initialized (still null). 

### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation
Fixes #27540

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
@@ -33,7 +33,6 @@ import com.google.android.material.snackbar.Snackbar
 import kotlinx.coroutines.Dispatchers.IO
 import kotlinx.coroutines.Dispatchers.Main
 import kotlinx.coroutines.Job
-import kotlinx.coroutines.flow.collect
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.mapNotNull
 import kotlinx.coroutines.launch
@@ -1400,6 +1399,11 @@ abstract class BaseBrowserFragment :
         binding.swipeRefresh.isEnabled = shouldPullToRefreshBeEnabled(inFullScreen)
     }
 
+    @CallSuper
+    internal open fun onUpdateToolbarForConfigurationChange(toolbar: BrowserToolbarView) {
+        toolbar.dismissMenu()
+    }
+
     /*
      * Dereference these views when the fragment view is destroyed to prevent memory leaks
      */
@@ -1472,7 +1476,9 @@ abstract class BaseBrowserFragment :
     override fun onConfigurationChanged(newConfig: Configuration) {
         super.onConfigurationChanged(newConfig)
 
-        _browserToolbarView?.dismissMenu()
+        _browserToolbarView?.let {
+            onUpdateToolbarForConfigurationChange(it)
+        }
     }
 
     // This method is called in response to native web extension messages from
--- a/app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
@@ -5,7 +5,6 @@
 package org.mozilla.fenix.browser
 
 import android.content.Context
-import android.content.res.Configuration
 import android.os.StrictMode
 import android.view.View
 import android.view.ViewGroup
@@ -32,6 +31,7 @@ import org.mozilla.fenix.GleanMetrics.ReaderMode
 import org.mozilla.fenix.R
 import org.mozilla.fenix.components.FenixSnackbar
 import org.mozilla.fenix.components.TabCollectionStorage
+import org.mozilla.fenix.components.toolbar.BrowserToolbarView
 import org.mozilla.fenix.components.toolbar.ToolbarMenu
 import org.mozilla.fenix.ext.components
 import org.mozilla.fenix.ext.nav
@@ -90,7 +90,7 @@ class BrowserFragment : BaseBrowserFragment(), UserInteractionHandler {
 
         browserToolbarView.view.addNavigationAction(homeAction)
 
-        setScreenSize(isTablet = resources.getBoolean(R.bool.tablet))
+        updateToolbarActions(isTablet = resources.getBoolean(R.bool.tablet))
 
         val readerModeAction =
             BrowserToolbar.ToggleButton(
@@ -170,7 +170,14 @@ class BrowserFragment : BaseBrowserFragment(), UserInteractionHandler {
         }
     }
 
-    private fun setScreenSize(isTablet: Boolean) {
+    override fun onUpdateToolbarForConfigurationChange(toolbar: BrowserToolbarView) {
+        super.onUpdateToolbarForConfigurationChange(toolbar)
+
+        updateToolbarActions(isTablet = resources.getBoolean(R.bool.tablet))
+    }
+
+    @VisibleForTesting
+    internal fun updateToolbarActions(isTablet: Boolean) {
         if (isTablet == this.isTablet) return
 
         if (isTablet) {
@@ -280,6 +287,8 @@ class BrowserFragment : BaseBrowserFragment(), UserInteractionHandler {
         refreshAction?.let {
             browserToolbarView.view.addNavigationAction(it)
         }
+
+        browserToolbarView.view.invalidateActions()
     }
 
     private fun removeTabletActions() {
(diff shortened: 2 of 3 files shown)


zmckenney:
I will land this squashed if approved to prevent multiple commits from main merges.

app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
```diff
@@ -1472,7 +1476,9 @@ abstract class BaseBrowserFragment :
     override fun onConfigurationChanged(newConfig: Configuration) {
         super.onConfigurationChanged(newConfig)
 
-        _browserToolbarView?.dismissMenu()
+        _browserToolbarView?.run {
+            onUpdateToolbarForConfigurationChange(this)
+        }
```
jonalmeida, line 1481: nit: simplify. :)

```suggestion
        _browserToolbarView?.onUpdateToolbarForConfigurationChange(this)
```

app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
```diff
@@ -280,6 +287,8 @@ class BrowserFragment : BaseBrowserFragment(), UserInteractionHandler {
         refreshAction?.let {
             browserToolbarView.view.addNavigationAction(it)
         }
+
+        browserToolbarView.view.invalidateActions()
```
jonalmeida, line 291: Wouldn't this call still throw a nullptr for the same reason?

app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
```diff
@@ -170,7 +170,14 @@ class BrowserFragment : BaseBrowserFragment(), UserInteractionHandler {
...
 
-    private fun setScreenSize(isTablet: Boolean) {
+    override fun onUpdateToolbarForConfigurationChange(toolbar: BrowserToolbarView) {
+        super.onUpdateToolbarForConfigurationChange(toolbar)
+
+        setToolbarActionsForSize(isTablet = resources.getBoolean(R.bool.tablet))
+    }
+
+    @VisibleForTesting
+    internal fun setToolbarActionsForSize(isTablet: Boolean) {
```
gabrielluong, line 180:
```suggestion
    internal fun updateToolbarActions(isTablet: Boolean) {
```
I am wondering if this would be a better function name since we aren't actually setting any value within the function.

jonalmeida (commented):
Drive-by, leaving this up to gl who reviewed the initial PR.

gabrielluong (approved):
We spoke over Slack, but let's try to squash the commits and update the PR title to reflect the intended Bugzilla issue that is being fixed here. https://bugzilla.mozilla.org/show_bug.cgi?id=1804107

I will admit we're in a bit of a weird state since we haven't fully migrated Fenix from GH issues to Bugzilla so to ensure we are getting QA testing for this crash. Let's make sure we reopen #27540 and have "eng:qa-needed" so that we can verify that this crash is no longer happening. Please add a note in #27540 pointing to https://bugzilla.mozilla.org/show_bug.cgi?id=1804107 with some STR the crash so QA will have more directions on how to test and verify this. Thank you!

app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
zmckenney: agreed, will update to `updateToolbarActions`

app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
zmckenney: this function (`addTabletActions`) is only called when the `onUpdateToolbarForConfigurationChange` has been called which is null checked in the BaseBrowserFragment (since it is a private backing property).

app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
zmckenney: actually this is a good callout and bigger than a nit. Your suggestion wouldn't work because this really should change to `?.let` since it was just a null check and shadowing will be fine.

zmckenney: > Thanks for this comment. I was looking at this and in my head, I was thinking using `run` is a bit peculiar because I was thinking "huh this is like the same as `let`" and at the same time, I am all for simplicity.
`
Somehow I missed your comment which came first. You were correct, I think @jonalmeida was saying this should be called on the `_browserToolbarView` since he saw a `run` but this needs to be a `let` as you were thinking.

zmckenney: it does resolve the null pointer issue! It works fine and would be safe to merge as-is but `run` is more commonly used to return the changed object from the calling block (as in a `run` to change fields in an object and return that object). Because it was assumed that `_browserToolbarView?.onUpdateToolbarForConfigurationChange(this)` would work it tells me at a glance people will believe this is being called on itself when in reality it is not. This is pseudo-incorrect code (works fine but not convention) so I will change it to what it should be :)

gabrielluong: Hmm.. reading https://kotlinlang.org/docs/scope-functions.html#run states that it invokes with `let` so I think using `run` is fine, but `let` would probably make this clearer to me.

gabrielluong: @jonalmeida your original suggestion wouldn't work because `onUpdateToolbarForConfigurationChange` is a function in `BaseBrowserFragment`

zmckenney: @gabrielluong you're correct and this is completely safe as you mentioned. The difference is just in convention and what is returned (which in this case is not used). If we use `let` here then the result of the lambda is returned but if we use `run` like I did we return the resulting object which _could_ be useful in the future which is why I initially chose it but for readability `?.let` is maybe better.

jonalmeida: Oh yes, sorry - was fooled too by this call. Semantics aside between run/let, maybe we need a comment here since two of us have been caught by this call.

Architectural context: We usually try to avoid inheritance-like structures or build scaffolding (see [AC components architecture docs](https://mozac.org/contributing/architecture) and `UserInteractionHandler` for smaller code examples) around it so you we don't end up writing fixes that involve inheritance because it can be confusing. There are times when you can't fight it because Android/Java is built that way, and this might be one of those cases. Where I don't have good suggestions to void in a crash-fix PR. :)

gabrielluong: Let's file a followup to investigate whether or not we can avoid the inheritance-like structure.  The change is simple enough right now for the crash fix. We can look at the follow investigation afterwards.

zmckenney: Agreed but if @gabrielluong agrees I want to change this to `?.let` so it's clear, readable, and doesn't require a comment to understand.

About inheritence structures I completely agree and I actually have a separate branch done without it that we can debate whether it is better or not in the future. I think it still mostly makes sense for this and considering future changes but look forward to hearing other thoughts!

gabrielluong: I think we can go with`?.let` without the extra comment that should be clear enough (hopefully).

Issue #28111, opened 2022-12-07
[Bug]: Allows users to choose as defaults only general or custom search engines


DreVla:
### Steps to reproduce

1. Go to Settings
2. Go to Search
3. Inspect default search engine options

### Expected behaviour

Only general or custom search engines should be available to be set as default search engine.
If users have a `topic specific` search engine set as default then the default should be automatically migrated to the first available `general` search engine. Only if no other general search engines are available then the user will remain with the current `topic-specific` engine set as default.

### Actual behaviour

Any search engine can be selected as default search engine, including topic specific engines (amazon, ebay, etc...).

<img width="466" alt="Screenshot 2022-12-07 at 16 22 06" src="https://user-images.githubusercontent.com/32488956/206203714-c965be75-b8ab-4597-a509-4587403bb604.png">

### Android version

Android 10

### Firefox release type

Firefox Nightly

### Firefox version

109.0a1

### Additional information

Because topic specific search engines - like amazon / ebay / wikipedia will only show results from those specific websites - a restricted scope they are not suited for general searches - all around the web.

We need to limit the user’s ability to set such engines as defaults.
Since these search engines come bundled with the app we have control over them and can manually split each list of search engines for a specific locale to general or topic specific (naming not necessarily following this scheme).

This will not apply to user added search engines. All custom search engines will be treated as general search engines.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-22112)

cadeyrn:
> We need to limit the user’s ability to set such engines as defaults.

@DreVla Why do you see this need? I would argue that an user who sets a topic specific search engine as default search engine, for example Amazon, is well aware that they will only find products on Amazon. And if they use the browser very frequently for the search for products (or tweets or articles on Wikipedia or whatever), what is wrong with allowing the user to set that search engine as the default search engine? I don't get the point of this change. 🤔

Mugurell:
This limitation will be part of the new unified search UX which will allow users to easily switch from the toolbar to another search engine - like Amazon or to another application source - like open tabs. 
There will be important differences between the UX for a general search engine like DuckDuckGo and the UX for a search engine like eBay.
Also the UX reasoning is that quick searches should show results from all over the internet - hence only search engines like DuckDuckGo, Google, Bing, can be set as default.
Will double check with UX and product.

SoftVision-LorandJanos:
[...]

PR #28112, opened 2022-12-07
For #28111: Filter out topic specific search engines from default list

DreVla:
Since topic specific search engines, like amazon, ebay, etc... will only show results from those specific websites, they will be filtered out of the default search engines list.

The implementation for now is a hardcoded, since we do not have a clear separation between "general" search engines and "topic specific" search engines. When that patch lands, this PR will require modifications.

Screenshot 2022-12-07 at 16 33 33

### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation

Used by GitHub Actions.


app/src/main/java/org/mozilla/fenix/settings/search/RadioSearchEngineListPreference.kt
```diff
@@ -76,14 +76,19 @@ class RadioSearchEngineListPreference @JvmOverloads constructor(
...
         state.searchEngines.filter { engine ->
             engine.type != SearchEngine.Type.APPLICATION
         }.forEach { engine ->
+            val isGeneralOrCustomSearchEngine = engine.isGeneral || engine.type == SearchEngine.Type.CUSTOM
             val searchEngineView = makeButtonFromSearchEngine(
                 engine = engine,
                 layoutInflater = layoutInflater,
                 res = context.resources,
-                allowDeletion = state.searchEngines.size > 1,
+                allowDeletion = !(isGeneralOrCustomSearchEngine && isLastGeneralOrCustomSearchEngine),
```
DreVla, line 91: This is needed in order to not allow the last general/custom search engine to be removed, therefore leading to not having one to set as default search engine. Open to feedback on how else this could be handled.

DreVla:
There will still be an issue rooted in AC with this ticket. Upon deleting search engines that are set as default, the first remaining search engine in the list will be set as default. This is a problem when the first search engine will be a topic specific one (see video). 

https://user-images.githubusercontent.com/32488956/206997058-56c9d991-b1dc-4cd9-9244-ff955d61f684.mp4

The fix I am looking into is instead of setting the first search engine in the list, find the next search engine that is general or custom and if it is the last one, do not allow the deletion of it.

DreVla:
@Verdi I am tagging you here to ask for feedback on handling how the topic specific search engines are displayed in the list. My first approach was to remove them completely, but this proved to be wrong, since the user would not be able to delete topic specific search engines at all from the list. The current approach is to gray out the topic specific search engines and the radio button, still showing them on the list but disabling the option of selecting them as default search engine. Let me know what approach you think is better suited or if you have any other in mind. Thank you!

Verdi:
> Let me know what approach you think is better suited or if you have any other in mind. Thank you!

Without getting into a whole reorganization of the search settings this will work fine. Thanks.

Mugurell:
> The fix I am looking into is instead of setting the first search engine in the list, find the next search engine that is general or custom and if it is the last one, do not allow the deletion of it.

That should work 👍.
There is another edgecase to consider: not allowing users to delete the last general search engine even if more topic specific ones would be available.

app/src/main/java/org/mozilla/fenix/settings/search/RadioSearchEngineListPreference.kt
```diff
@@ -155,6 +166,10 @@ class RadioSearchEngineListPreference @JvmOverloads constructor(
         context: Context,
         engine: SearchEngine,
     ) {
+        val nextSearchEngine = context.components.core.store.state.search.searchEngines.first {
+            it.id != engine.id && (it.isGeneral || it.type == SearchEngine.Type.CUSTOM)
+        }
+        context.components.useCases.searchUseCases.selectSearchEngine(nextSearchEngine)
```
DreVla, line 172: Thanks to @Mugurell this fixes the behavior of selecting a topic specific search engine instead of the next general/custom engine.

https://user-images.githubusercontent.com/32488956/207871980-c8ccd126-fe34-41fb-a0d5-271ee8e7e342.mp4

Mugurell:
Seems like `ui-test-arm-debug` is failing because of [deleteAllSearchEnginesTest](https://console.firebase.google.com/u/1/project/moz-fenix/testlab/histories/bh.66b7091e15d53d45/matrices/6252985103684773410/details?stepId=bs.13494eee019bc92&tabId=test-cases&testCaseId=3) probably when the last general search engines is tried to be deleted, something which is expected after this patch.
You could update the test yourself or disable it with `@Ignore` and mention there the new ticket for QA to investigate how to re-enable it.

Mugurell:
Thank you!
This looks good but there is one important detail we have might missed?: What happens if users already have "topic-specific" search engines set as default?

DreVla:
In the current implementation, in case a user has a topic specific search engine set as default before this patch, when updated, that search engine will remain his default.

cc @kbrosnan & @Verdi is this something we would want to change or should we keep it this way?

Mugurell:
> In the current implementation, in case a user has a topic specific search engine set as default before this patch, when updated, that search engine will remain his default.
> 
> cc @kbrosnan & @Verdi is this something we would want to change or should we keep it this way?

Synced now with @Verdi about the current status of unified search and regarding this they recommended automatically selecting the default search engine for that region or the first available. Which I think for us would mean selecting the first "general" search engine available.

app/src/androidTest/java/org/mozilla/fenix/ui/SettingsSearchTest.kt
```diff
@@ -348,6 +348,7 @@ class SettingsSearchTest {
         }
     }
 
+    @Ignore("Test failure caused by: https://github.com/mozilla-mobile/fenix/issues/28111")
```
Mugurell, line 351: We should open a new ticket with details about which tests were disabled and why for QA to know about this and then link that new ticket in the `@Ignore` annotation.

app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -560,6 +564,22 @@ open class FenixApplication : LocaleAwareApplication(), Provider {
...
+     */
+    @Suppress("NestedBlockDepth")
+    private fun migrateTopicSpecificSearchEngines() {
+        components.core.store.state.search.selectedOrDefaultSearchEngine.let { currentSearchEngine ->
+            if (currentSearchEngine?.isGeneral == false) {
+                components.core.store.state.search.searchEngines.firstOrNull() { nextSearchEngine ->
+                    nextSearchEngine.isGeneral
+                }.let {
+                    if (it != null) components.useCases.searchUseCases.selectSearchEngine(it)
+                }
```
Mugurell, line 578: The `let` scope function is used most of the time in combination with the safe call operator - `?.` to execute a lambda only on non-null objects:

```suggestion
                }?.let {
                   components.useCases.searchUseCases.selectSearchEngine(it)
                }
```

app/src/main/java/org/mozilla/fenix/settings/search/RadioSearchEngineListPreference.kt
```diff
@@ -76,14 +76,20 @@ class RadioSearchEngineListPreference @JvmOverloads constructor(
             ViewGroup.LayoutParams.WRAP_CONTENT,
         )
 
+        val isLastGeneralOrCustomSearchEngine = state.searchEngines.filter {
+            it.isGeneral ||
+                it.type == SearchEngine.Type.CUSTOM
```
Mugurell, line 81: Here and below: The `isGeneral` check should already account for the `CUSTOM` type of a search engine - https://github.com/mozilla-mobile/firefox-android/blob/f99a255860a9faa4a9845f206a6a7b4694871bc6/android-components/components/feature/search/src/main/java/mozilla/components/feature/search/storage/SearchEngineReader.kt#L74-L78

app/src/main/java/org/mozilla/fenix/settings/search/RadioSearchEngineListPreference.kt
```diff
@@ -99,12 +103,18 @@ class RadioSearchEngineListPreference @JvmOverloads constructor(
         isSelected: Boolean,
     ): View {
         val isCustomSearchEngine = engine.type == SearchEngine.Type.CUSTOM
+        val isTopicSpecific = !(engine.isGeneral || isCustomSearchEngine)
```
Mugurell, line 106: The `isGeneral` check should already account for the CUSTOM type of a search engine - https://github.com/mozilla-mobile/firefox-android/blob/f99a255860a9faa4a9845f206a6a7b4694871bc6/android-components/components/feature/search/src/main/java/mozilla/components/feature/search/storage/SearchEngineReader.kt#L74-L78.

I understand the `isGeneral` property may not easily transmit this. We could have a new extension property for a search engine - `isTopicSpecific` that would just be `!isGeneral` if this will make the functionality easier to reason about.

app/src/main/java/org/mozilla/fenix/settings/search/RadioSearchEngineListPreference.kt
```diff
@@ -76,14 +76,18 @@ class RadioSearchEngineListPreference @JvmOverloads constructor(
...
         state.searchEngines.filter { engine ->
             engine.type != SearchEngine.Type.APPLICATION
         }.forEach { engine ->
             val searchEngineView = makeButtonFromSearchEngine(
                 engine = engine,
                 layoutInflater = layoutInflater,
                 res = context.resources,
-                allowDeletion = state.searchEngines.size > 1,
+                allowDeletion = !(engine.isGeneral && isLastGeneralOrCustomSearchEngine) &&
+                    state.searchEngines.count { it.type != SearchEngine.Type.APPLICATION } > 1,
```
Mugurell, line 90: Just a question about why we need to have an `APPLICATION` search engine available to allow deletion since `Bookmarks` / `Tabs` / `History` search engines can't be set as default?

DreVla: This is to fix a bug I caught during the implementation. The user is able to delete all the search engines from the settings list, leading to an empty list.

https://user-images.githubusercontent.com/32488956/209302478-ec9c00d4-89f6-4a76-a52f-6b002931e91a.mp4

Since this is not connected to the ticket in question here, I will create a new one and add a PR for it. Thank you for pointing it out!

app/src/main/java/org/mozilla/fenix/settings/search/AddSearchEngineFragment.kt
```diff
@@ -182,6 +182,7 @@ class AddSearchEngineFragment :
                         name,
                         searchString.toSearchUrl(),
                         requireComponents.core.icons.loadIcon(IconRequest(searchString)).await().bitmap,
+                        isGeneral = true,
```
DreVla, line 185: As suggested here: https://github.com/mozilla-mobile/firefox-android/pull/273/files#r1045456951

DreVla:
Currently waiting on this for a decision regarding adding telemetry

DreVla:
During 11th January Kanban meeting a decision has been made to migrate users that have topic specific search engine as default to first available regional general or custom search engine. This has been already implemented here: https://github.com/mozilla-mobile/fenix/pull/28112#discussion_r1055210204

DreVla:
There is no need for new telemetry for now. Regarding migration, it will remain and we will monitor user feedback.
This PR is now ready for review!

Mugurell (approved):
LGTM. Thanks!

Issue #28093, opened 2022-12-06
[Bug]: [Experiment] [Viewpoint Survey (2022 Dec, Android)] The Viewpoint survey message is not displayed on the Homepage of the Firefox Release 107.2.0 and 106.1.0 versions


SoftVision-CarmenFat:
### Steps to reproduce

**[Prerequisites]:**
- Have the device’s language set to any of the following: 'de', 'en', 'es', 'fr', 'ja', 'pl', 'pt', 'ru', 'zh'.
- Have the Firefox 107 Release build installed on a device.

**[Steps to reproduce]:**
1. Open the 107.2.0 Release build from the prerequisites.
2. Navigate to the Firefox "Settings" menu.
3. Scroll down to the bottom of the list and tap the "About Firefox" option.
4. Tap the "Firefox Browser" logo 5 times and go back to the "Settings" menu.
5. Scroll down to the bottom of the list and tap the "Secret Settings" option.
6. Enable the “Use Nimbus Preview Collection” option and restart the browser.
7. Navigate again to the Firefox "Settings" menu -> "About Firefox" option.
8. Tap the "Firefox Browser" logo 5 times and go back to the "Settings" menu.
9. Scroll down to the bottom of the list and tap the "Nimbus Experiments" option.
10. Tap the "Viewpoint Survey (2022 Dec, Android)" experiment.
11. Tap the “treatment” branch and restart the browser.
12. Observe the Homepage.

### Expected behaviour

- A message that asks the user to complete a survey is displayed on the Homepage. 

### Actual behaviour

- The Survey message is not displayed on the Homepage. 

### Device name

Samsung S9, Oneplus 5

### Android version

Android 10

### Firefox release type

Firefox

### Firefox version

107.2.0, 106.1.0 

### Additional information

- The issue is partially reproducible on Firefox Release 103, but in this case, the message has the wrong label displayed on the message.
![Screenshot_20221206-145129](https://user-images.githubusercontent.com/26000308/205949286-69a08a9a-62cc-47d5-87be-797e2c62203c.jpg)

- The issue is NOT reproducible with the Firefox Release 105 version on Android devices.
- The issue is NOT reproducible using Firefox 108 RC on an iOS device.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-22105)

Mugurell:
Thank you for the report!

Based on the versions in which this was tested it seems like 
[...]

PR #28147, opened 2022-12-09
For #28093 - Try to show Nimbus messages after the messages are loaded

Mugurell:
Currently the `Evaluate` action to try showing Nimbus messages is dispatched from `MessagingFeature` when the app's `onStart` method is called - and the application is visible.
The `Restore` action which would load the available messages is dispatched based on a callback that although is set when the app is opened it was seen in manual tests that it could fire after app's `onStart` and as such the loaded messages would not be tried to be shown.

Adding a new call to evaluate messages after we know they are loaded will ensure they will be shown as soon as possible.

The race condition between loading messages and trying to show them is resolved in version 109 of the app (current app version) by switching from a callback system for knowing when the Nimbus configuration is done to having all needed operations run on `Main` in application's `onCreate` making the data available to when later `MessagingFeature` tries to show the next available message.

Video showing the survey appearing on a local build of 108 with this patch applied:

https://user-images.githubusercontent.com/11428869/206699060-9372b8e6-ad66-4a38-87e9-3e6673cd05ad.mp4

### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation

Used by GitHub Actions.

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/gleanplumb/state/MessagingMiddleware.kt
+++ b/app/src/main/java/org/mozilla/fenix/gleanplumb/state/MessagingMiddleware.kt
@@ -40,6 +40,12 @@ class MessagingMiddleware(
                 coroutineScope.launch {
                     val messages = messagingStorage.getMessages()
                     context.store.dispatch(UpdateMessages(messages))
+                    // It may be that the first `Evaluate` from `MessagingFeature` was dispatched
+                    // before the messages were loaded based on the `Restore` action and so
+                    // it won't lead to messages being shown because none were loaded.
+                    // Calling `Evaluate` again after the messages are loaded based on the `Restore` action
+                    // will ensure the newly made available messages will be shown.
+                    context.store.dispatch(Evaluate)
                 }
             }
 


app/src/main/java/org/mozilla/fenix/gleanplumb/state/MessagingMiddleware.kt
```diff
@@ -40,6 +40,7 @@ class MessagingMiddleware(
                 coroutineScope.launch {
                     val messages = messagingStorage.getMessages()
                     context.store.dispatch(UpdateMessages(messages))
+                    context.store.dispatch(Evaluate)
```
Amejia481, line 43: I'm a bit confused if this will reliably sole the issue,  as the `Evaluate` action [will trigger an UI update](https://github.com/mozilla-mobile/fenix/blob/49ccb63dd2232464e9e1a3d96d4f47f670b25c7e/app/src/main/java/org/mozilla/fenix/gleanplumb/state/MessagingMiddleware.kt#L46-L54), but here we are initializing the app when we haven't loaded the  `HomeFragment` where we are showing the message.

Amejia481 (approved):
As mentioned on this [comment](https://github.com/mozilla-mobile/fenix/pull/28147/files#r1044502982) , I'm not sure if this will solved the issue reliably, but you already verify that this improve the situation I won't block it.

app/src/main/java/org/mozilla/fenix/gleanplumb/state/MessagingMiddleware.kt
Mugurell: The race was between 
- [this callback from `FenixApplication`](https://github.com/mozilla-mobile/fenix/blob/49ccb63dd2232464e9e1a3d96d4f47f670b25c7e/app/src/main/java/org/mozilla/fenix/FenixApplication.kt#L456) which [would dispatch MessagingAction.Restore](https://github.com/mozilla-mobile/fenix/blob/49ccb63dd2232464e9e1a3d96d4f47f670b25c7e/app/src/main/java/org/mozilla/fenix/FenixApplication.kt#L472) and 
- `MessagingFeature` which would dispatch [MessagingAction.Evaluate](https://github.com/mozilla-mobile/fenix/blob/49ccb63dd2232464e9e1a3d96d4f47f670b25c7e/app/src/main/java/org/mozilla/fenix/gleanplumb/MessagingFeature.kt#L19)

With some logs I saw yesterday together with James that `MessagingAction.Evaluate` would be dispatched first by `MessagingFeature` before the Nimbus callback would dispatch `MessagingAction.Restore`.
And as such we would try to show messages before they are available.

The changes from this patch will help ensure that `MessagingAction.Evaluate` is always called after `MessagingAction.Restore` and `MessagingAction.Update`, after the messages are available.

Mugurell: This changes will help with the scenario in which `HomeFragment` was loaded but `MessagingAction.Restore` would not yet be dispatched.
With this patch we are dispatching `MessagingAction.Evaluate` a second time to complete the first one which didn't show anything because messages were not yet available when `HomeFragment` is shown.

jhugman: I'm less familiar with the unidirectional data flow, and where the Middleware sits in the stack, so am happy to defer to @Amejia481 's good judgement here.

jhugman (commented):
I will defer to @Amejia481 judgement here.

Either landing this PR, or uplifting 
https://github.com/mozilla-mobile/fenix/pull/27650 (and even https://github.com/mozilla-mobile/fenix/pull/27927) would be reasonable approaches to fixing this bug.

app/src/main/java/org/mozilla/fenix/gleanplumb/state/MessagingMiddleware.kt
Amejia481: The only drawback is as we are dispatching twice the `MessagingAction.Restore`, in some cases we would be reduce the times that users will see a message, ideally we would improve that by observing the `messages` in `MessagingFeature`, and dispatching when they are updated.

Amejia481: For now, I think this address the main issue, we could do a follow up ticket to the solution improve it!

Mugurell: This patch is only needed for 108.
109 (current Nightly soon to become Beta) and forward switched from depending on that Nimbus callback and so won't need any change.

Mugurell: Thank you!
Created https://github.com/mozilla-mobile/fenix/issues/28156 to evaluate and fix in later versions also.

Mugurell: Added a comment in code to offer a bit more context for why `Evaluate` has to be called after `Restore`.

Amejia481: Perfect, feel free to land when you are ready!

Mugurell:
The decision was to base the behavior on https://github.com/mozilla-mobile/fenix/pull/27650 which will be present in the next release - v109.
Closing this PR as it is not needed anymore.

PR #28195, opened 2022-12-14
Bug 1807279 - create new cookie banner reduction UI smoke tests

### LINKED ISSUE #26144: [UI test automation] Improve coverage of the UI smoke test suite
### Summary

Improve coverage of the [UI smoke test suite](https://testrail.stage.mozaws.net/index.php?/suites/view/3192&group_by=cases:section_id&group_order=asc&group_id=45740):
• Create new UI tests
• Add missing steps

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21090)

AndiAJ:
Bug 1807279 - create new cookie banner reduction UI smoke tests ✅ both tests successfully passed 100x on Firebase.

#### Summary:
► How this feature works ❔
 
This feature works only on certain websites, **hard coded** in the [cookie banner rule list](https://github.com/mozilla/cookie-banner-rules-list/blob/main/cookie-banner-rules-list.json).
After you navigate to one of the listed websites, Gecko will be able to identify the corresponding cookie banner and buttons for that particular website and programatically click for the user either "Reject all" or "Accept all" (depending on the cookie banner)

► Reasons for picking `voetbal24.be` to be used in our tests:

1. The tests will run on Firebase and this website displays the cookie banner even if you access the website from the US (due to the fact that there isn't any cookie law in the US, most of the websites will not display the banner)

2. The banner gets displayed faster, compared to other websites

3. Ease of querying and verifying the cookie banner compared to other websites:
Resource id of the [cookie banner](https://github.com/mozilla/cookie-banner-rules-list/blob/main/cookie-banner-rules-list.json#L641) from the rule list -> [cookie banner](https://github.com/mozilla-mobile/fenix/pull/28195/files#diff-a14b9f04389140fd0dcc9fb550bd69103bc83d54075c0ca0360ea9b95c696f5aR1247) variable in our test.

► Note 
There's a [known inconsistency](https://github.com/mozilla-mobile/fenix/issues/28197), because of which I had to add a [retry loop and refresh the page](https://github.com/mozilla-mobile/fenix/pull/28195/files#diff-a14b9f04389140fd0dcc9fb550bd69103bc83d54075c0ca0360ea9b95c696f5aR901) to verify that the cookie banner isn't displayed (sometimes the banner is dismissed after revisiting the page, other times you need to refresh the page)

### Pull Request checklist

### QA

- [ ] **QA Needed**

5. Click on `View task in Taskcluster` in the new `DETAILS` section.
6. The APK links should be on the right side of the screen, named for each CPU architecture.

### GitHub Automation
Fixes #26144


sv-ohorvath (approved):

sv-ohorvath:
Looks good, my only concern is if we should continue with this if a real website is the only way to check the feature. I'll leave it up to @AaronMT

AndiAJ:
The only way to use `mockwebserver` would be to navigate to exactly the same url as listed in the rule list and respond with our local asset. (not sure if it's feasible ?)

Another worth mentioning aspect is that the cookie banner and its elements should have exactly the same ids as the pages from the rule list.

app/src/androidTest/java/org/mozilla/fenix/ui/SettingsPrivacyTest.kt
```diff
@@ -705,6 +705,90 @@ class SettingsPrivacyTest {
             }
         }
     }
+
+    @SmokeTest
+    @Test
+    fun verifyCookieBannerReductionTest() {
+        val webSite = "voetbal24.be"
```
Amejia481, line 712: Could checking with a real site make the test flaky? As we don't have control over the site, could we have a local site instead?

AaronMT:
From reading https://github.com/mozilla/cookie-banner-rules-list/blob/main/README.md#test-rules, it might be possible to add a JSON rule for what would be localhost (maybe?) and then load up any generic page asset. But again this would require manipulating Gecko prefs

AndiAJ:
Thanks, for the review and suggestions, will see what I can do.

app/src/androidTest/java/org/mozilla/fenix/ui/SettingsPrivacyTest.kt
Amejia481: Please don't feel the need to block on this comment, maybe this is something that could be improve in the feature.

AaronMT:
> Unfortunately, can't change the `cookiebanners.listService.testRules` pref, meaning that I can't use a custom cookie banner rule for localhost.
> 
> Should I leave the test as it is ? or mark it as unsuitable ?
> 
> @AaronMT , let me know what you think.

Let's ask @Trikolon here for advice. Hi Paul, we're looking to create an instrumented UI test for Fenix that verifies the new cookie banner. We could use a live website but prefer to if possible to use localhost through our mock web server. We found the readme here [1] that mentions using test-rules in a Gecko pref. @AndiAJ tested that above and it doesn't seem to work. I don't know the entirety of @AndiAJ's issue but he did mention that the pref was blank after setting it. Do you know if this is an issue or if we can use localhost in a test-rule in a debug builds of which our tests are ran?

[1] https://github.com/mozilla/cookie-banner-rules-list/blob/main/README.md#test-rules

Trikolon:
The mechanism itself in Gecko already has automated test coverage. You can find our tests here: https://searchfox.org/mozilla-central/source/toolkit/components/cookiebanners/test
If you want tests in addition to that you could serve a local test page like we do in the mochitests. Here is an example click test we have: https://searchfox.org/mozilla-central/rev/57527d50ef5d3df412caa5d99536f0709399be6f/toolkit/components/cookiebanners/test/browser/browser_bannerClicking.js#64 It injects rules via the cookie banner service exposed to chrome code via `Services.cookieBanners`. You can also use the test pref to inject rules. I'm not aware of any cases where the pref would be reset. We don't have code that can do this. When a test rule set via `cookiebanners.listService.testRules` fails to parse it should generate an error in the browser toolbox.

AndiAJ:
Thanks @Trikolon !

@Mugurell was kind enough to help me out to try and update the pref locally, but still, somethings not working properly.
Added the [patch](https://github.com/mozilla-mobile/fenix/pull/28195/commits/03807391f31c5fd2f85ef61f715d3fd3cf6c4fdf) I've used to test if it works locally.

Perhaps @Amejia481 could also weigh in on this and share his thoughts ? ☺️

Amejia481:
> Thanks @Trikolon !
> 
> @Mugurell was kind enough to help me out to try and update the pref locally, but still, somethings not working properly. Added the [patch](https://github.com/mozilla-mobile/fenix/pull/28195/commits/03807391f31c5fd2f85ef61f715d3fd3cf6c4fdf) I've used to test if it works locally.
> 
> Perhaps @Amejia481 could also weigh in on this and share his thoughts ? ☺️

I think we don't have an API that is exposed to the app's level that allows to add new test rules, I think at the moment is OK to use a live site due to the limitations.

app/src/androidTest/java/org/mozilla/fenix/ui/SettingsPrivacyTest.kt
```diff
@@ -705,6 +705,90 @@ class SettingsPrivacyTest {
...
+    //         clickCookieBannerReductionToggle()
+    //         verifyCheckedCookieBannerReductionToggle(false)
+    //         exitMenu()
+    //     }
+    //     browserScreen {
+    //     }.openThreeDotMenu {
+    //     }.refreshPage {
+    //         verifyCookieBannerExists(exists = false)
+    //     }
+    // }
```
AaronMT, line 791: Need some context in comment on commenting these out. Would it better to disable these instead?

app/src/main/java/org/mozilla/fenix/gecko/GeckoProvider.kt
```diff
@@ -58,7 +61,26 @@ object GeckoProvider {
...
+                    }
+
+                    println("Andi: content of $it afterwards:")
+                    println("-------------------------")
+                    println(File(it).bufferedReader().use { it.readText() })
+                    println("-------------------------")
+
+                    println("Andi: Config sent to GV: ${context.filesDir.path + File.separatorChar + GECKO_DEBUG_CONFIG_FILE_NAME}")
+                },
+            ).build()
```
AaronMT, line 83: @Amejia481 if you don't mind reviewing this bit, thanks

app/src/androidTest/java/org/mozilla/fenix/ui/SettingsPrivacyTest.kt
AndiAJ: You can ignore this Aaron, I've commented out these steps for myself while trying to update the gecko pref.

app/src/main/java/org/mozilla/fenix/gecko/GeckoProvider.kt
AndiAJ: This doesn't work @AaronMT , no needs for review.

AndiAJ:
Based on what, @Amejia481 replied, it looks like we can't update gecko prefs.

@AaronMT , how do you want us to proceed ?

Mark the test as unsuitable ?
Use a real website - The test worked properly using a real website in [my initial commit](https://github.com/mozilla-mobile/fenix/pull/28195/commits/bf14a79d4fc0f9306ce0c25cfddbdcba08bba9a4)

AaronMT (approved):

PR #28202, opened 2022-12-14
For #28201 - Make a LightDarkPreview annotation

### LINKED ISSUE #28201: Make a “@LightDarkPreview()” annotation that makes two previews from one annotation
Currently we use two annotations in Compose Preview functions in order to display previews with both dark mode on and off.

This task is for creating a multi-annotation class that wraps both of these annotations into a single annotation, helping with readability and any future extension of the multi-annotation so it will include other rendering options.

Tasks:
- [ ] Create multi-annotation for both dark mode on and off
- [ ] Replace all uses of the two annotations with the single multi-annotation

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-22148)

HarrisonOg:
### Pull Request checklist

### QA

- [ ] **QA Needed**

### GitHub Automation
Fixes #28201

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/compose/Divider.kt
+++ b/app/src/main/java/org/mozilla/fenix/compose/Divider.kt
@@ -4,7 +4,6 @@
 
 package org.mozilla.fenix.compose
 
-import android.content.res.Configuration
 import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
@@ -16,8 +15,8 @@ import androidx.compose.foundation.layout.width
 import androidx.compose.material.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
-import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
+import org.mozilla.fenix.compose.annotation.LightDarkPreview
 import org.mozilla.fenix.theme.FirefoxTheme
 
 /**
@@ -39,8 +38,7 @@ fun Divider(
  * An example of a vertical divider.
  */
 @Composable
-@Preview(uiMode = Configuration.UI_MODE_NIGHT_YES)
-@Preview(uiMode = Configuration.UI_MODE_NIGHT_NO)
+@LightDarkPreview
 private fun VerticalDividerPreview() {
     FirefoxTheme {
         Box(
@@ -74,8 +72,7 @@ private fun VerticalDividerPreview() {
  * An example of divider usage in a list menu.
  */
 @Composable
-@Preview(uiMode = Configuration.UI_MODE_NIGHT_YES)
-@Preview(uiMode = Configuration.UI_MODE_NIGHT_NO)
+@LightDarkPreview
 private fun HorizontalDividerPreview() {
     FirefoxTheme {
         Box(
--- a/app/src/main/java/org/mozilla/fenix/compose/ListItemTabLarge.kt
+++ b/app/src/main/java/org/mozilla/fenix/compose/ListItemTabLarge.kt
@@ -4,7 +4,6 @@
 
 package org.mozilla.fenix.compose
 
-import android.content.res.Configuration
 import androidx.compose.foundation.clickable
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Column
@@ -22,9 +21,9 @@ import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.text.style.TextOverflow
-import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.sp
+import org.mozilla.fenix.compose.annotation.LightDarkPreview
 import org.mozilla.fenix.theme.FirefoxTheme
 
 /**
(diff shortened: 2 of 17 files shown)


app/src/main/java/org/mozilla/fenix/compose/LightDarkPreview.kt
```diff
@@ -0,0 +1,8 @@
+package org.mozilla.fenix.compose
+
+import android.content.res.Configuration
+import androidx.compose.ui.tooling.preview.Preview
+
+@Preview(uiMode = Configuration.UI_MODE_NIGHT_YES)
+@Preview(uiMode = Configuration.UI_MODE_NIGHT_NO)
+annotation class LightDarkPreview
```
MozillaNoah, line 8: let's move this initial change into its own commit, so we'll ultimately land two commits with this PR

app/src/main/java/org/mozilla/fenix/compose/LightDarkPreview.kt
```diff
@@ -0,0 +1,8 @@
+package org.mozilla.fenix.compose
```
MozillaNoah, line 1: this file needs the license

MozillaNoah, line 1: to prevent the root `.../compose/` directory from becoming too bloated, let's move this to `.../compose/annotation`. We'll add any more similar annotation classes there in the future.

MozillaNoah (commented):
This looks great! I've left some comments for a few minor things before we land this. Just make sure to run detekt/lint on your branches before opening PRs or pushing new commits to help save some review cycles and get the CI all-green.

gabrielluong:
Noice 👌

app/src/main/java/org/mozilla/fenix/compose/LightDarkPreview.kt
MatthewTighe, line 8: 🚗 I'm not super familiar with annotation classes, but I assume `LightDarkPreview` doesn't expose the parameters available to `@Preview` by some transitive magic or anything. Should we consider adding those same parameters here? Is there a way to pass them to both of the super classes? 

I'm just concerned that we could lose some flexibility in the our preview options if this becomes a common usage.

app/src/main/java/org/mozilla/fenix/compose/LightDarkPreview.kt
HarrisonOg: sounds good, will make the new directory for annotation

HarrisonOg: ah I knew this file was looking a little empty. fixing

app/src/main/java/org/mozilla/fenix/compose/LightDarkPreview.kt
HarrisonOg: this is a good point, I'll look into the options for `@Preview`

gabrielluong:
Just a reminder that we'll want to do a squash landing. It would also be helpful to start incorporating an amend/squash workflow since we want to avoid fixup-type commits landing in the tree (eg, "[fix import ordering](https://github.com/mozilla-mobile/fenix/pull/28202/commits/4254f344a7739a4651a733abe43b9de854f9c43f)").

app/src/main/java/org/mozilla/fenix/compose/LightDarkPreview.kt
HarrisonOg: hi @MatthewTighe 

after some further investigation, I have discovered that we mainly use only three fields for the following scenarios:
1) locale : this is used for Right to Left and Left to Right display previews. 
2) uiMode: this is used for Night Mode on and off 
3) name: used for labeling  

This new multi-annotation preview class would mainly be used to replace option 2 as it is the most common, used just with the night mode on and off.

  I also discovered that I cannot pass the parameters up to the multi-annotations like in this example: 
```
@Preview(name = name, uiMode = Configuration.UI_MODE_NIGHT_YES)
@Preview(name = name, uiMode = Configuration.UI_MODE_NIGHT_NO)
annotation class LightDarkPreview(
    val name: String = "",
    val group: String = "",
    @IntRange(from = 1) val apiLevel: Int = -1,
    // TODO(mount): Make this Dp when they are inline classes
    val widthDp: Int = -1,
    // TODO(mount): Make this Dp when they are inline classes
    val heightDp: Int = -1,
    val locale: String = "",
    @FloatRange(from = 0.01) val fontScale: Float = 1f,
    val showSystemUi: Boolean = false,
    val showBackground: Boolean = false,
    val backgroundColor: Long = 0,
    @UiMode val uiMode: Int = 0,
    @Device val device: String = Devices.DEFAULT
)
```
 my proposal is that when more annotation fields are needed, use the original @Preview() annotation with the fields filled in while keeping this multi-annotation preview class for just the Night Mode on and off scenario.

MozillaNoah (approved):
LGTM. Just make sure to edit the message of the second commit to add the issue number

app/src/main/java/org/mozilla/fenix/compose/LightDarkPreview.kt
MatthewTighe: That sounds good to me! Shame the parameters can't be shared. Seems like it would be super helpful to have the name propagate, for instance.

Issue #9413, opened 2020-03-25
[Bug] (regression) changing language requires an app restart now


cadeyrn:
### Steps to reproduce

1. change the language within Fenix

### Expected behavior

The change applies immediately, as before https://github.com/mozilla-mobile/fenix/commit/5202c0a2f925b772ec8dca9e652b2e9593a620bf.

### Actual behavior

You have to restart Fenix to see the change of the language. It's a regression from the appcompat update in https://github.com/mozilla-mobile/fenix/commit/5202c0a2f925b772ec8dca9e652b2e9593a620bf.

But good news is that the update fixed #7559 - an app start is also required in this case, but it's much better than a feature that is not working at all.

/cc @eliserichards @ekager 

### Device information

* Android device: OnePlus 7T Pro McLaren Edition, OxygenOS 10.0.7 (Android 10)
* Fenix version: master branch revision 40ba43311e1782ffe6602e90d70c41a7b104998e

Diana-Rus:
Hi, I confirm issue on FIrefox Preview Nightly 200326 (Build #20860605) GV: 76.0a1-20200324093140 with:
- Google Pixel 3 XL (Android 9)
![20200326-160131](https://user-images.githubusercontent.com/49859863/77656182-2a07bf00-6f7c-11ea-9978-1f0023540e92.gif)

- Samsung Galaxy S9 (Android 8)
- Sony Xperia Z5 (Android 7)

liuche:
This should be fixed by https://github.com/mozilla-mobile/android-components/issues/6517, which landed earlier, so it will probably be in the next AC nightly version that gets pulled into fenix.

cadeyrn:
This is working for me now.

softvision-miralobontiu:
Hi all,

This is not entirely fixed yet. 
- In Fenix, I've set the language to Arabic/Urdu (or any RTL language), the language changed as expected in the app.
- Change back to follow the device's language, the app remained stuck between languages, meaning the language is changed, but the app is stuck in the RTL position.
- If Fenix is in the background and reopened, nothing changes. 
- If Fenix is closed and reopened, it is working as expected again. 

**Device information:**
* Android devices: HTC 10 (Android 8), Samsung Galaxy S9 (Android 8), and Motorola G6 (Android 8),
* Fenix version: Nightly build from 4/8
[...]

PR #28343, opened 2022-12-29
Closes #28342: change the ui direction after a locale change

### LINKED ISSUE #28342: [BUG] change RTL/LTR languages doesn't update the UI until new launch
Issue that was mentioned [here](https://github.com/mozilla-mobile/fenix/issues/21306#issuecomment-1101778924).

`Switching from an RTL to an LTR locale (or vice versa) doesn't update the UI direction without restarting Fenix first`

### Steps to reproduce

Go to language screen and change the language to a language with the opposite direction from the one you are currently using.

### Expected behaviour

The layout of the app will change to accommodate the direction of the set language.
### Actual behaviour

The direction of the layout keeps the same until relaunching the app.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-22207)

mavduevskiy:
https://user-images.githubusercontent.com/92760693/209997180-e66ed269-88b8-46ae-8ce8-caf6540cb30d.mp4

### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation
Fixes #28342


app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -229,6 +231,10 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
 
         components.publicSuffixList.prefetch()
 
+        // Changing a language on the Language screen restarts the activity, but the activity keeps
+        // the old layout direction. We have to update the direction manually.
+        window.decorView.layoutDirection = TextUtils.getLayoutDirectionFromLocale(Locale.getDefault())
```
MozillaNoah, line 236: Drive by: This seems like a last resort location to do an initialization step from a setting. Would it be possible to include this in the locale settings flow when updating the language? `DefaultLocaleSettingsController` might be a better alternative.

mavduevskiy: That was my first though as well, but this approach has an issue – invalidating the layout and recreating the activity happens not synchronously and looks like a glitch.

https://user-images.githubusercontent.com/92760693/211407358-b70ea490-f7b1-4cff-86e7-bfb4c22c5382.mp4

What concerned did you have in mind here? Applying proper orientation during `onCreate` looks okay to me. `setLayoutDirection` has also a safety check to invalidate the layout only if it is needed.

MozillaNoah: Where is the orientation being set in HomeActivity?

My primary concern is that this was even needed at all and might be a heavy hammer, but I did notice that `HomeActivity` extends from `LocaleAwareAppCompatActivity` which has this same line in its own `onCreate`. Based on the adjacent comment in `HomeActivity`, the `super.onCreate` only gets called after some disk read violations, but it was obviously needed.

mavduevskiy: The use-case I am working on is `DefaultLocaleSettingsController` changing the Locale to something that has a different text orientation and restarts the activity. Changing the locale should also change the layout orientation, but it doesn't and the activity is recreated with the old orientation. To avoid glitching, enforcing changed orientation looks like a good idea to me. What do you think?

MozillaNoah: Judging by the conversation in https://github.com/mozilla-mobile/fenix/issues/9413, I think I'm OK with this approach, as it's a natural side effect of us having to take care of all of the configuration changes. Maybe it won't be necessary in a 100% Compose world 🤞

MozillaNoah (approved):

mavduevskiy:
Thank you, Noah! @MozillaNoah

PR #28348, opened 2022-12-30
Draft/ POC - Improve layouts for smaller screens

### LINKED ISSUE #26929: New onboarding has scrolled off content
Issue initially seen by Mira on Focus - https://github.com/mozilla-mobile/focus-android/issues/7500#issuecomment-1237045914 which also reproduces on Fenix:

The onboarding is set to be vertically scrollable and this may be an issue on devices with a low screen resolution / density (especially for the first screen which is shown with the action button initially off screen):

[FenixOnboardingCutOff.webm](https://user-images.githubusercontent.com/11428869/189587634-54e8f0b0-1f2c-4cca-b288-883c86312b43.webm)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-21537)

rahulsainani:
### What
– Introduce `FirefoxDimens` with a grid system to use a consistent spacing across the app. It also allows to create multiple variants of dimension sets which can be used for different screen sizes. This PR introduces two dimension sets:

- `smallDimensions` – for devices lower than [sw360dp ](https://developer.android.com/guide/topics/large-screens/support-different-screen-sizes#alternative_layout_resources) like Nexus One. There are very small number of devices that fall in this category.
- `sw360Dimensions` - For devices larger than sw360dp like Pixels. 
The same strategy can be used to add dimension sets for tablets like sw600dp and sw720dp. 

This PR initially intended to solve #26929 and adding dimensions set is a fundamental step so I have marked this as draft to get feedback on that. However, feel free to comment/ask questions on the usage of dimensions as well. 

### Extras - can be moved to another PR 
– Move the state to ViewModel and add tests. Add [turbine](https://github.com/cashapp/turbine) to test `SharedFlow`s.
– Make OnboardingPage reusable which takes `OnboardingPageUiState` and click listeners. The idea was to allow another `OnboardingPage` wit only updating the state mapping in the `ViewModel` and handling the clicks.
– Use [BoxWithConstraints](https://developer.android.com/jetpack/compose/layouts/basics#responsive-layouts) to make the image take 40% of the screen size. Previously it would take the space based on the size defined in the resource.
– Vertical scrolling is still enabled as some locales have longer text and Text doesn't auto size text as of now. Ideally, most devices should not need to scroll.

### Preview Screenshots for width 320dp and height 540dp

| Before | After | 
|------|-------|
|Screenshot 2022-12-30 at 17 19 00|Screenshot 2022-12-30 at 17 19 12

### Annotation for checking multiple size previews
```
@Preview(name = "small", widthDp = 240, heightDp = 400)
@Preview(name = "small", widthDp = 240, heightDp = 400, locale = "de")
@Preview(name = "medium", widthDp = 320, heightDp = 540)
@Preview(name = "default")
@Preview(name = "pixel", device = Devices.PIXEL)
@Preview(name = "nexus 5", device = Devices.NEXUS_5)
@Preview(name = "tablet", device = Devices.TABLET)
annotation class MultiDevicePreview

```

### Notes
– [Window Size Classes](https://developer.android.com/guide/topics/large-screens/support-different-screen-sizes#alternative_layout_resources) is another complimentary strategy to build responsive layouts for a large percentage of device sizes in different configurations.
– This got bigger as some fundamental additions were needed 😅 

### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation
Fixes #26929

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingViewModel.kt
+++ b/app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingViewModel.kt
@@ -1,0 +1,158 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.onboarding.view
+
+import androidx.annotation.DrawableRes
+import androidx.annotation.StringRes
+import androidx.lifecycle.ViewModel
+import androidx.lifecycle.viewModelScope
+import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.SharedFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.launch
+import mozilla.telemetry.glean.private.NoExtras
+import org.mozilla.fenix.GleanMetrics.Onboarding
+import org.mozilla.fenix.R
+
+/**
+ * ViewModel holding the screen state for [OnboardingScreen].
+ */
+class OnboardingViewModel : ViewModel() {
+
+    private val _state = MutableStateFlow<OnboardingScreenState>(OnboardingScreenState.Initial)
+    internal val state: StateFlow<OnboardingScreenState> = _state
+    private val _navigationEvent = MutableSharedFlow<OnboardingNavigationEvent>()
+    internal val navigationEvent: SharedFlow<OnboardingNavigationEvent> = _navigationEvent
+
+    /**
+     * This should be called onLaunch of [OnboardingScreen] to update to the Content State.
+     */
+    fun onLaunch(isUserSignedIn: Boolean) {
+        if (_state.value is OnboardingScreenState.Initial) {
+            _state.value = OnboardingScreenState.Content(
+                onboardingState = OnboardingState.Welcome,
+                isUserSignedIn = isUserSignedIn,
+            )
+        }
+    }
+
+    /**
+     * Click handler for primary button.
+     */
+    fun onPrimaryButtonClick() {
+        viewModelScope.launch {
+            stateIsContent {
+                when (it.onboardingState) {
+                    OnboardingState.Welcome -> {
+                        Onboarding.welcomeGetStartedClicked.record(NoExtras())
+                        if (it.isUserSignedIn) {
+                            _navigationEvent.emit(OnboardingNavigationEvent.DISMISS)
+                        } else {
+                            _state.value = it.copy(onboardingState = OnboardingState.SyncSignIn)
+                        }
+                    }
+                    OnboardingState.SyncSignIn -> {
+                        Onboarding.syncSignInClicked.record(NoExtras())
+                        _navigationEvent.emit(OnboardingNavigationEvent.SIGN_IN)
+                    }
+                }
+            }
+        }
+    }
+
+    /**
+     * Click handler for secondary button.
+     */
+    fun onSecondaryButtonClick() {
+        viewModelScope.launch {
+            stateIsContent {
+                when (it.onboardingState) {
+                    OnboardingState.Welcome -> {
+                        // nothing as welcome doesn't have a secondary button
+                    }
+                    OnboardingState.SyncSignIn -> {
+                        Onboarding.syncSkipClicked.record(NoExtras())
+                        _navigationEvent.emit(OnboardingNavigationEvent.DISMISS)
+                    }
+                }
+            }
+        }
+    }
+
+    private suspend fun stateIsContent(block: suspend (OnboardingScreenState.Content) -> Unit) {
+        if (_state.value is OnboardingScreenState.Content) {
+            block(_state.value as OnboardingScreenState.Content)
+        }
+    }
+}
+
+/**
+ * Enum that represents the onboarding page that is displayed.
+ */
+enum class OnboardingState {
+    Welcome,
+    SyncSignIn,
+}
+
+/**
+ * State of the Onboarding Screen.
+ */
+sealed class OnboardingScreenState {
+    /**
+     * The Initial state of the screen when the content isn't loaded.
+     * Can be used to show a progress bar if required.
+     */
+    object Initial : OnboardingScreenState()
+
+    /**
+     * The Content state of the screen.
+     * @param onboardingState the page that is selected.
+     * @param isUserSignedIn boolean that signifies if the user has signed in.
+     * pageUiState is a derived property that contains the page data based on [onboardingState].
+     */
+    data class Content(
+        val onboardingState: OnboardingState,
+        val isUserSignedIn: Boolean,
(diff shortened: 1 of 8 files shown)


buildSrc/src/main/java/Dependencies.kt
```diff
@@ -72,6 +73,7 @@ object FenixDependencies {
     const val kotlin_reflect = "org.jetbrains.kotlin:kotlin-reflect:${FenixVersions.kotlin}"
     const val kotlin_coroutines = "org.jetbrains.kotlinx:kotlinx-coroutines-core:${FenixVersions.coroutines}"
     const val kotlin_coroutines_test = "org.jetbrains.kotlinx:kotlinx-coroutines-test:${FenixVersions.coroutines}"
+    const val cashapp_turbine = "app.cash.turbine:turbine:${FenixVersions.cashapp_turbine}"
```
gabrielluong, line 76: I think this is gonna need a bit more scrutiny because adding a new dependency isn't a simple matter.

app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingViewModel.kt
```diff
@@ -0,0 +1,158 @@
...
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.launch
+import mozilla.telemetry.glean.private.NoExtras
+import org.mozilla.fenix.GleanMetrics.Onboarding
+import org.mozilla.fenix.R
+
+/**
+ * ViewModel holding the screen state for [OnboardingScreen].
+ */
+class OnboardingViewModel : ViewModel() {
```
gabrielluong, line 23: We have moved away from using `ViewModel` in favour of [Store](https://searchfox.org/mozilla-mobile/source/firefox-android/android-components/components/lib/state/src/main/java/mozilla/components/lib/state/Store.kt), which follows more closely with [Redux](https://redux.js.org/). 

Some examples:
https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayStore.kt
https://searchfox.org/mozilla-mobile/search?q=Store&path=&case=false&regexp=false

app/src/main/java/org/mozilla/fenix/theme/FirefoxDimens.kt
```diff
@@ -0,0 +1,90 @@
...
+
+@file:Suppress("ConstructorParameterNaming")
+
+package org.mozilla.fenix.theme
+
+import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.dp
+
+/**
+ * A grid system. Smaller components can align to a 'sub' grid [grid_0_25] or [grid_0_5].
```
gabrielluong, line 13: I am wondering where this grid system is coming from? I am a bit hesitant to ship a grid system that isn't guided by our UX design system.

mavduevskiy:
While having system of that sort looks tempting, introducing it in order to fix image scaling on one screen looks like two separate (and very different in scope) tasks.

My question was about the solution itself, how should the solution to the [problem](https://github.com/mozilla-mobile/fenix/issues/26929) described in the issue look like? 

We don't want to have any offscreen content, right? Would scaling down the image to fit the space between the button and the top of the screen be the solution? I tested the patch in the split mode, and it had some offscren content still.

Also it scales down the background as well, which is different from the overall behavior of the app in such extreme conditions.

75%|50%|25%
---|---|---
||

buildSrc/src/main/java/Dependencies.kt
rahulsainani: Gotcha! Thanks for flagging, i will ask about this 👍

app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingViewModel.kt
rahulsainani: Good call! this logic in the VM can be moved to the Store and reducer can handle the actions. Yes, the intention was to move the state and logic outside the composables and even some of the navigation events that depend on the state. I'm going to explore this.

app/src/main/java/org/mozilla/fenix/theme/FirefoxDimens.kt
rahulsainani: Hmm the thinking was that we are hard coding these dimensions on the screens, and it might help to centralize them as a grid system. 

We can look into aligning them with the design system and involve UX in it. Although that could be a step ahead, we could look into how to use the already hardcoded dimensions as a common variable of the theme as a first improvement. wdyt?

rahulsainani:
@mavduevskiy 
1. Yeah one solution could be to only use ratios and `BoxWithConstraints` and leave out the rest of the changes. Adding dimensions helped it further. I see that these can be separate patches, they went well together and I thought it would be good to get some feedback on this. I will mark this as a draft / POC for now but would be great to keep getting feedback on this.
2. Scaling the image seemed good as it fits the design better instead of the size from the resource itself, and the 40% came from the image height to screen height ratio in the mock-up. Let me know your thoughts? it could have some off screen content in extreme situations (like scenario 3 ) and that's why we kept the scrolling functionality. Could you please confirm that this scrolls in scenario 2 and 3, as that's the intended behavior.
3. Could you please describe what does scaling the background means here? I'm not sure I understand. 😅

app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingViewModel.kt
Mugurell: The big advantage of a `ViewModel` is that it survives configuration changes. Which in the current architecture and with the app handling screen rotations (with the UI not being destroyed) doesn't warrant it's a bit more complex machinery.
With the reducers not having access to other dependencies and them being `pure` functions the domain logic would usually stay in `Controller`s (`Interactor`s are also being phased out with them having little to no separate responsability) - https://github.com/mozilla-mobile/fenix/blob/main/docs/architecture-overview.md.

mavduevskiy:
@rahulsainani 

1.2. Let's step back from the implementation talk, I am still trying to understand the solution we are going for. The [reported bug](https://github.com/mozilla-mobile/fenix/issues/26929) was about scrolled off content. As far as I understand, we want to have the action button on the screen always. Are we going for creating a layout that will display the action button regardless of the display size?
3. The thing I was talking about is this. Scaling the window back to 50 and 25 percent creates side spacing. On the 25% the DialogFragment background looks like a card, positioned vertically, not taking the available horizontal space. But! It is not the result of the patch, it is the current implementation. I just though that if we are trying to fix the layout of the onboarding screen for smaller devices, we could make the UI adjustable for split modes as well.

boek:
Moved the issue to bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1814190

csadilek:
@rahulsainani We will have to close this draft in preparation of the Monorepo migration. Please re-open on the new repo once ready. You will have to reapply your patch, because of the updated history. Thank you!

rahulsainani:
Thanks for the heads up @csadilek ! Sounds good!

PR #28350, opened 2022-12-30
Bug #1807716: Fix for launch crash on Huawei and other devices

zmckenney:
Fixes launch crash for Bugzilla [1807716](https://bugzilla.mozilla.org/show_bug.cgi?id=1807716). 

The pushlog doesn't show any changes in Fenix or GV which would have caused this crash and it is only reproducible when using GV artifacts after linking it to Fenix. There was likely a change in Glean or somewhere else that is causing the crash but one way we can fix the launch crash which is also technically correct is to initialize the runtime earlier instead of doing it in Glean init. 

This also has the added benefit of the crash reporter catching the current crash. 

1807716 will still need to be investigated for root cause.

### Pull Request checklist

### QA

- [ ] **QA Needed**

### GitHub Automation
Fixes #1807716

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
+++ b/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
@@ -203,6 +203,11 @@ open class FenixApplication : LocaleAwareApplication(), Provider {
         ProfilerMarkerFactProcessor.create { components.core.engine.profiler }.register()
 
         run {
+            // Make sure the engine is initialized and ready to use.
+            components.strictMode.resetAfter(StrictMode.allowThreadDiskReads()) {
+                components.core.engine.warmUp()
+            }
+
             // We need to always initialize Glean and do it early here.
             initializeGlean()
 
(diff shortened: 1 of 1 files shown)


csadilek (approved):
We worked on this for the investigation of https://bugzilla.mozilla.org/show_bug.cgi?id=1807716.

I think this change is good to make either way, because the "warmup" call, which has the purpose of initializing the Gecko runtime and engine, is currently unnecessary. In practice, we currently initialize the Gecko runtime as a side-effect of initializing Glean and its `ConceptFetchHttpUploader`, which needs a `components.core.client`. That client is a `GeckoFetchClient`, which triggers runtime initialization. 

With this change, we make initialization of engine explicit again, and it turns out, it also fixes the startup crash we're investigating in bug 1807716. However, we don't understand why Glean initializing the `components.core.client` is now causing a segmentation fault all of a sudden. This still needs to be investigated, but landing this should allow the affected clients to start up again.

csadilek:
Also /cc @jonalmeida. We worked together on recent startup refactorings, and also noticed and didn't like that Gecko was initialized via Glean/init, as opposed to by the `warmup` call we both expected. So this addresses that.

PR #28080, opened 2022-12-05
Close #19919: Migrate away from startActivityForResult

rocketsroger:
### Pull Request checklist

- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features. In addition, it includes a screenshot of a successful [accessibility scan](https://play.google.com/store/apps/details?id=com.google.android.apps.accessibility.auditor&hl=en_US) to ensure no new defects are added to the product.

### QA

- [x] **QA Needed**

### To download an APK when reviewing a PR (after all CI tasks finished running):
1. Click on `Checks` at the top of the PR page.
2. Click on the `firefoxci-taskcluster` group on the left to expand all tasks.
3. Click on the `build-debug` task.
4. Click on `View task in Taskcluster` in the new `DETAILS` section.
5. The APK links should be on the right side of the screen, named for each CPU architecture.

### GitHub Automation
Fixes #19919

PR #28351, opened 2022-12-30
Close #19919: Migrate away from startActivityForResult

### LINKED ISSUE #19919: Migrate away from Fragment.startActivityForResult()
Fragment.startActivityForResult(). This API is deprecated in androidx.fragment:fragment 1.3+.  Migrate away from it.
https://developer.android.com/training/basics/intents/result

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-10967)

rocketsroger:
Reopening for https://github.com/mozilla-mobile/fenix/pull/28080.  (couldn't reopen)

### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation
Fixes #19919

### DIFF
--- a/app/src/main/java/org/mozilla/fenix/ext/Fragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/ext/Fragment.kt
@@ -4,7 +4,12 @@
 
 package org.mozilla.fenix.ext
 
+import android.app.Activity
+import android.content.Intent
 import android.view.WindowManager
+import androidx.activity.result.ActivityResult
+import androidx.activity.result.ActivityResultLauncher
+import androidx.activity.result.contract.ActivityResultContracts
 import androidx.annotation.IdRes
 import androidx.annotation.StringRes
 import androidx.appcompat.app.AppCompatActivity
@@ -127,4 +132,20 @@ fun Fragment.removeSecure() {
     this.activity?.window?.clearFlags(
         WindowManager.LayoutParams.FLAG_SECURE,
     )
+}
+
+/**
+ * Register a request to start an activity for result.
+ */
+fun Fragment.registerForActivityResult(
+    onFailure: (result: ActivityResult) -> Unit = {},
+    onSuccess: (result: ActivityResult) -> Unit,
+): ActivityResultLauncher<Intent> {
+    return registerForActivityResult(ActivityResultContracts.StartActivityForResult()) { result ->
+        if (result.resultCode == Activity.RESULT_OK) {
+            onSuccess(result)
+        } else {
+            onFailure(result)
+        }
+    }
 }
--- a/app/src/main/java/org/mozilla/fenix/search/SearchDialogFragment.kt
+++ b/app/src/main/java/org/mozilla/fenix/search/SearchDialogFragment.kt
@@ -6,7 +6,6 @@ package org.mozilla.fenix.search
 
 import android.Manifest
 import android.annotation.SuppressLint
-import android.app.Activity
 import android.app.Dialog
 import android.content.Context
 import android.content.DialogInterface
@@ -27,6 +26,7 @@ import android.view.ViewStub
 import android.view.WindowManager
 import android.view.accessibility.AccessibilityEvent
 import android.view.inputmethod.InputMethodManager
+import androidx.activity.result.ActivityResultLauncher
 import androidx.annotation.VisibleForTesting
 import androidx.appcompat.app.AlertDialog
 import androidx.appcompat.app.AppCompatDialogFragment
@@ -87,6 +87,7 @@ import org.mozilla.fenix.databinding.SearchSuggestionsHintBinding
 import org.mozilla.fenix.ext.components
 import org.mozilla.fenix.ext.getRectWithScreenLocation
 import org.mozilla.fenix.ext.increaseTapArea
+import org.mozilla.fenix.ext.registerForActivityResult
 import org.mozilla.fenix.ext.requireComponents
 import org.mozilla.fenix.ext.settings
 import org.mozilla.fenix.search.awesomebar.AwesomeBarView
@@ -96,7 +97,6 @@ import org.mozilla.fenix.search.toolbar.SearchSelectorMenu
 import org.mozilla.fenix.search.toolbar.SearchSelectorToolbarAction
 import org.mozilla.fenix.search.toolbar.ToolbarView
 import org.mozilla.fenix.settings.SupportUtils
-import org.mozilla.fenix.widget.VoiceSearchActivity
 
 typealias SearchDialogFragmentStore = SearchFragmentStore
 
@@ -110,6 +110,7 @@ class SearchDialogFragment : AppCompatDialogFragment(), UserInteractionHandler {
     private lateinit var toolbarView: ToolbarView
     private lateinit var inlineAutocompleteEditText: InlineAutocompleteEditText
     private lateinit var awesomeBarView: AwesomeBarView
+    private lateinit var startForResult: ActivityResultLauncher<Intent>
 
     private val searchSelectorMenu by lazy {
         SearchSelectorMenu(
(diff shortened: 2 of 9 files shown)


rocketsroger:
When this lands, I will ask QA to verify that WebAuthn is still working. We can only do this on Mozilla signed builds (nightly).

app/src/main/java/org/mozilla/fenix/ext/Fragment.kt
```diff
@@ -128,3 +133,14 @@ fun Fragment.removeSecure() {
         WindowManager.LayoutParams.FLAG_SECURE,
     )
 }
+
+/**
+ * Register a request to start an activity for result.
+ */
+fun Fragment.registerForActivityResult(onSuccess: (result: ActivityResult) -> Unit): ActivityResultLauncher<Intent> {
+    return registerForActivityResult(ActivityResultContracts.StartActivityForResult()) { result ->
+        if (result.resultCode == Activity.RESULT_OK) {
```
mavduevskiy, line 142: It could actually cause bugs, in case the receiver expects and handles any other results but the success or is waiting for any result. Maybe adding a second callback for cancelled result for edge cases (like `BaseVoiceSearchActivity`)?

app/src/main/java/org/mozilla/fenix/search/SearchDialogFragment.kt
```diff
@@ -155,6 +156,15 @@ class SearchDialogFragment : AppCompatDialogFragment(), UserInteractionHandler {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setStyle(STYLE_NO_TITLE, R.style.SearchDialogStyle)
+
+        startForResult = registerForActivityResult { result ->
+            val intent = result.data
+            intent?.getStringArrayListExtra(RecognizerIntent.EXTRA_RESULTS)?.first()?.also {
```
mavduevskiy, line 162: Looks like we could omit the parameter:
`result.data?.getStringArrayListExtra(RecognizerIntent.EXTRA_RESULTS)?.first()?.also {`

Less code, happier us!

app/src/main/java/org/mozilla/fenix/settings/biometric/BiometricPromptPreferenceFragment.kt
```diff
@@ -108,7 +107,10 @@ abstract class BiometricPromptPreferenceFragment : PreferenceFragmentCompat() {
         }
     }
 
-    override fun onActivityResult(requestCode: Int, resultCode: Int, data: Intent?) {
+    /**
+     * Use [BiometricPromptFeature] or [KeyguardManager] to handle pin request results.
+     */
+    fun pinRequestResult(requestCode: Int, resultCode: Int) {
```
mavduevskiy, line 113: We could remove this function: it duplicates check for the `RESULT_OK` and there is not need for the request constant in the new API.

app/src/main/java/org/mozilla/fenix/widget/VoiceSearchActivity.kt
```diff
@@ -37,6 +41,23 @@ class VoiceSearchActivity : AppCompatActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
 
+        startForResult = registerForActivityResult(ActivityResultContracts.StartActivityForResult()) {
```
mavduevskiy, line 44: I guess we should use the extension here as well. 

Here the receiver expects the callback to be called even without `RESULT_OK`. But the extension function consumes `results` that are not `RESULT_OK` without any feedback, which might be unexpected. Maybe adding a second callback for cancelled state for edge cases?

app/src/test/java/org/mozilla/fenix/widget/VoiceSearchActivityTest.kt
```diff
@@ -141,50 +134,6 @@ class VoiceSearchActivityTest {
         assertNull(shadow.peekNextStartedActivityForResult())
     }
 
-    @Test
-    fun `handle speech result`() {
```
mavduevskiy, line 145: Couldn't we adapt the tests to the new API? Just removing them seems a bit risky.

mavduevskiy (commented):
The long awaited refactoring! Great!

Since we are adding an extension function `Fragment.registerForActivityResult` and doing refactoring, wouldn't it be the right time to refactor direct calls to the API by the extension? (`BookmarkSearchDialogFragment`, `HistorySearchDialogFragment`,`VoiceSearchActivity`).

And some comments.

app/src/main/java/org/mozilla/fenix/ext/Fragment.kt
rocketsroger: Sure, I'll add the onFailure function and set it to empty by default.  Currently from testing, I don't see any need to handle results when failed.

app/src/main/java/org/mozilla/fenix/search/SearchDialogFragment.kt
rocketsroger: Sure, I'll remove `val intent`

app/src/main/java/org/mozilla/fenix/settings/biometric/BiometricPromptPreferenceFragment.kt
rocketsroger: This is currently being called in `AutofillSettingFragment.kt`.  Do you mean to move it?

app/src/main/java/org/mozilla/fenix/widget/VoiceSearchActivity.kt
rocketsroger: I think we should just use this here since currently there is only one activity doing this.  Once there are more than one activity, moving it to an extension function makes sense then.

app/src/test/java/org/mozilla/fenix/widget/VoiceSearchActivityTest.kt
rocketsroger: I spent most of my time on trying to get these two tests working.  However, with the new implementation, `shadow.receiveResult()` doesn't work anymore.  I couldn't find another way to perform the tests here and had to remove them.

rocketsroger:
Looking at the release calendar.  I think we'll wait till next cut before landing this change.

app/src/test/java/org/mozilla/fenix/widget/VoiceSearchActivityTest.kt
mavduevskiy: Yeah, I couldn't find any way to make `ShadowActivity` work with the new API as well. 

Maybe we could do without it? It looks like the tests are checking the code that handles the [activity result](https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/java/org/mozilla/fenix/widget/VoiceSearchActivity.kt#89) – if the new activity is called and the current is closed. 

We can't make the system pass the call through the new API, but we still could test the code that should work after the system call. We could move the code that used to be in `onActivityResult` into a separate, internal and visible for testing function, and call it from on `registerForActivityResult` and from the tests as well!

That way we could keep the tests, just changing one line – `shadow.receiveResult` to something like `activity.handleActivityResult` :)

app/src/main/java/org/mozilla/fenix/settings/biometric/BiometricPromptPreferenceFragment.kt
mavduevskiy: I am not sure, actually. What confuses me is, `AutofillSettingFragment` inherits `BiometricPromptPreferenceFragment`, and the child used to call`startActivityForResult` and the parent would receive the result. But the new API for `activityResult` doesn't use `requestCode` and is designed to handle the result in a lambda. We also handle `resultCode == RESULT_OK` inside the `registerForActivityResult` extension.
 
 Hm. If we want to keep the inheritance, how about we define `var startForResult: ActivityResultLauncher<Intent>` inside the parent, and children would just call `startForResult.launch(intent)`? Right now there is no other children, but it looks like the initial implementation of the `BiometricPromptPreferenceFragment` was aiming to handle `activityResult` in the parent, so we could keep the design.

mavduevskiy (commented):
Had a call today, writing down thoughts. `AutofillSettings` is more of a nit, but if it's possible to keep the tests, I think it's better to keep them!

app/src/test/java/org/mozilla/fenix/widget/VoiceSearchActivityTest.kt
rocketsroger: That's a good idea.  I'll update.  Thanks

app/src/main/java/org/mozilla/fenix/settings/biometric/BiometricPromptPreferenceFragment.kt
rocketsroger: I think I see what you mean.  Make sense.  I'll update.  Thanks

mavduevskiy (approved):
looks good to me!
🚢

PR #28367, opened 2023-01-03
Bug 1807727 - Dismiss SearchDialogFragment on editing cancelled

Alexandru2909:
In the current implementation, if we quickly go back from the search dialog fragment, the event may not reach the `onBackPressed` callback from `SearchDialogFragment` while still [invoking](https://github.com/mozilla-mobile/firefox-android/blob/36a1e5b4bf12c76a1071d48c28283742114fa11b/android-components/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/edit/EditToolbar.kt#L124) `onEditingCanceled`.

By dismissing the dialog in the `onEditingCanceled` callback, we avoid having the toolbar go into display mode when going back quickly.

| Before | After |
| --- | --- |
| https://user-images.githubusercontent.com/35462038/210335411-0dbb2bf0-d1bc-4b50-a190-408aaa7108c3.mp4 | https://user-images.githubusercontent.com/35462038/210335507-a4001a92-6d4b-4abe-87e0-8b11b52196a8.mp4 |

### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation

Used by GitHub Actions.


app/src/main/java/org/mozilla/fenix/search/SearchDialogFragment.kt
```diff
@@ -636,24 +637,28 @@ class SearchDialogFragment : AppCompatDialogFragment(), UserInteractionHandler {
...
-                }
-                view?.hideKeyboard()
-                dismissAllowingStateLoss()
+                goBack()
                 true
             }
         }
     }
 
+    private fun goBack() {
```
t-p-white, line 646: I like that this has been extracted to a function 👍 
Just looking at the comment

> // See mirror of this logic in BrowserToolbarController#handleToolbarClick.

Is this still relevant, as that logic block doesn't appear to mirror this necessarily?

app/src/main/java/org/mozilla/fenix/search/SearchDialogFragment.kt
```diff
@@ -636,24 +637,28 @@ class SearchDialogFragment : AppCompatDialogFragment(), UserInteractionHandler {
...
+        // so we don't need to navigate "back to" browser fragment.
+        // See mirror of this logic in BrowserToolbarController#handleToolbarClick.
+        if (store.state.searchTerms.isBlank()) {
+            val args by navArgs<SearchDialogFragmentArgs>()
+            args.sessionId?.let {
+                findNavController().navigate(
+                    SearchDialogFragmentDirections.actionGlobalBrowser(null),
+                )
+            }
+        }
```
t-p-white, line 657: nit: If the above comment does not apply to the entire block I think it's useful to have an empty line to make that clear - so here would a new line before 658 - or if it does it could be a fun description instead

app/src/main/java/org/mozilla/fenix/search/SearchDialogFragment.kt
t-p-white, line 646: would it be clearer if this was also called dismissDialogAndGoBack() ?

Alexandru2909: I believe this refers to the case when interacting with the toolbar from a tab without search terms (which does not have search terms set). In `handleToolbarClick` we navigate home before showing the search dialog for the home fragment to be visible in the decor. In `onBackPressed` from `SearchDialogFragment` we want to return to the browser fragment when dismissing the search dialog, instead of the home fragment. It seems to me that the comment is still relevant, having this case in mind. Should we maybe update it to be more descriptive?

t-p-white: I see, gotcha, thanks for explaining 👍 I misinterpreted this as literally mirroring the logic. Happy to keep comment as is, unless you have an update in mind?

t-p-white (approved):
lgtm 👍 approved as is however if you have an updated comment in mind then I can re-review

app/src/main/java/org/mozilla/fenix/search/SearchDialogFragment.kt
Alexandru2909: I'm ok with it as is too, thanks!

Issue #28330, opened 2022-12-28
Refactor Onboarding to use NavHost


gabrielluong:
In https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/java/org/mozilla/fenix/onboarding/view/Onboarding.kt, we are maintaining 2 states to display various content when we should really be assuming each of the views as a separate screen. We can achieve this by using [NavHost](https://developer.android.com/jetpack/compose/navigation#create-navhost).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-22205)

boek:
Moved to bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1812458

Change performed by the [Move to Bugzilla add-on](https://addons.mozilla.org/en-US/firefox/addon/move-to-bugzilla/).

PR #28376, opened 2023-01-03
Bug 1808358 - Refactor Onboarding Page to make it reusable

rahulsainani:
### What
– Create `OnboardingPage` that takes in `OnboardingPageUiState` to display the content. 
– Create `OnboardingContent` composable that selects the click actions and content based on the state and passes on the right content to OnboardingPage.

### Pull Request checklist

### QA

- [x] **QA Needed**

### GitHub Automation

Used by GitHub Actions.


app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingPage.kt
```diff
@@ -0,0 +1,127 @@
...
+private fun PreviewOnboardingPage() {
+    FirefoxTheme {
+        OnboardingPage(
+            onboardingPageUiState = OnboardingPageUiState(
+                image = R.drawable.ic_onboarding_sync,
+                title = R.string.onboarding_home_sync_title_3,
+                description = R.string.onboarding_home_sync_description,
+                primaryButtonText = R.string.onboarding_home_sign_in_button,
+                secondaryButtonText = R.string.onboarding_home_skip_button,
+                recordImpressionEvent = { },
```
MozillaNoah, line 118:
```suggestion
                recordImpressionEvent = {},
```

app/src/main/java/org/mozilla/fenix/onboarding/view/UpgradeOnboardingScreen.kt
```diff
@@ -0,0 +1,215 @@
...
+import org.mozilla.fenix.R
+import org.mozilla.fenix.theme.FirefoxTheme
+
+/**
+ * Enum that represents the onboarding page that is displayed.
+ */
+private enum class UpgradeOnboardingState {
+    Welcome,
+    SyncSignIn,
+}
```
MozillaNoah, line 50: Any thoughts on combining `OnboardingPageUiState` and `UpgradeOnboardingState` or making `UpgradeOnboardingState` a child class? It could simplify the `OnboardingPage(...)` initialization with the giant `when`.

app/src/main/java/org/mozilla/fenix/onboarding/view/UpgradeOnboardingScreen.kt
```diff
@@ -0,0 +1,215 @@
...
+                }
+            }
+        },
+        onSecondaryButtonClick = {
+            when (onboardingState) {
+                UpgradeOnboardingState.Welcome -> {
+                    // nothing as welcome doesn't have a secondary button
+                }
+                UpgradeOnboardingState.SyncSignIn -> {
+                    Onboarding.syncSkipClicked.record(NoExtras())
```
MozillaNoah, line 98: do the metrics pings compile with the Compose previews?

app/src/main/java/org/mozilla/fenix/onboarding/view/UpgradeOnboardingScreen.kt
```diff
@@ -0,0 +1,215 @@
...
+ * @param isUserSignedIn Whether or not the user is signed into their Firefox Sync account.
+ * @param onDismiss Invoked when the user clicks on the close or "Skip" button.
+ * @param onSignInButtonClick Invoked when the user clicks on the "Sign In" button
+ */
+@Composable
+fun UpgradeOnboardingScreen(
+    isUserSignedIn: Boolean,
+    onDismiss: () -> Unit,
+    onSignInButtonClick: () -> Unit,
+) {
```
MozillaNoah, line 64: Curious what you think, @gabrielluong, but this seems like a Composable that's serving UI for 2+ screens because we don't have Compose-based navigation. The modularity is appreciated and the API seems straightforward. It's just not 100% clear to me if all onboarding screens are/will/should conform to this template.

MozillaNoah (commented):
quick pass as I'm multitasking

app/src/main/java/org/mozilla/fenix/onboarding/view/UpgradeOnboardingScreen.kt
gabrielluong: My understanding was that this work was motivated by https://www.figma.com/file/d7l1iGabQYqnLcnQbHDGzM/Wallpaper-2.0-%2B-Onboarding-v106?node-id=8924%3A220946&t=ihw1RVyD5zhJ7Gze-4 which is to add an onboarding screen for push notifications. Therefore, keeping `OnboardingPageUiState` might still make sense.

gabrielluong: I have also briefly discussed with @rahulsainani about the pros and cons with the status quo with managing this `enum` state vs adopting a `NavHost` to get rid of the `when`. At the end of the day, both approaches would be  fine. I have filed https://github.com/mozilla-mobile/fenix/issues/28330 last week with the intent to move this to use navigation to remove the `when` that is going on.

Rahul have said "there's not a lot of navigation happening and we'd have to clear the back stack anyway as we don't allow users to go back on Welcome screen. It's fine either way imo". I am mostly okay with keeping the status quo since the navigation is minimal and not having to handle the back stack. We can keep https://github.com/mozilla-mobile/fenix/issues/28330 around for future exploration if we felt strongly about it.

app/src/main/java/org/mozilla/fenix/onboarding/view/UpgradeOnboardingScreen.kt
```diff
@@ -0,0 +1,215 @@
...
+        modifier = Modifier
+            .size(6.dp)
+            .clip(CircleShape)
+            .background(color),
+    )
+}
+
+@Composable
+@Preview
+private fun PreviewUpgradeOnboardingScreen() {
```
gabrielluong, line 207:
```suggestion
private fun UpgradeOnboardingScreenPreview() {
```
This is just to stick with the Compose Preview naming convention.

app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingPage.kt
```diff
@@ -0,0 +1,127 @@
...
+) {
+    Column(
+        modifier = modifier.semantics(mergeDescendants = true) {},
+        horizontalAlignment = Alignment.CenterHorizontally,
+        verticalArrangement = Arrangement.SpaceEvenly,
+    ) {
+        // Since the vertical arrangement is SpaceEvenly, the text and buttons are wrapped
+        // their own containing Composables so the OuterColumn only has 3 children with
+        // space in between and around them
+        // Item 1
```
gabrielluong, line 54: I think we could remove comments like these. I would hope it's generally understood when someone sees `verticalArrangement = Arrangement.SpaceEvenly` rather than have to explain what a property of `verticalArrangement` is doing.

app/src/main/java/org/mozilla/fenix/onboarding/view/UpgradeOnboardingScreen.kt
gabrielluong: This is fair point.

app/src/main/java/org/mozilla/fenix/onboarding/view/UpgradeOnboardingScreen.kt
MozillaNoah: Gotcha gotcha. I'm definitely open to there being a navigation refactor down the road, and I'm glad yall had that talk to figure out what makes sense. I'm open to moving forward as is, with the caution if onboarding, for some reason, gets interest in becoming more complex.

app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingPage.kt
```diff
@@ -0,0 +1,127 @@
...
+                SecondaryButton(
+                    text = stringResource(id = onboardingPageUiState.secondaryButtonText),
+                    onClick = onSecondaryButtonClick,
+                )
+            }
+        }
+    }
+
+    LaunchedEffect(onboardingPageUiState) {
+        // record impression event for onboarding page
```
gabrielluong, line 102: Let's remove the comment here. I think we are repeating the code here with more or less the same exact words.

app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingPageUiState.kt
```diff
@@ -0,0 +1,20 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.onboarding.view
+
+import androidx.annotation.DrawableRes
+import androidx.annotation.StringRes
+
+/**
+ * Model containing data for the [OnboardingPage].
```
gabrielluong, line 11: Let's add the @param description.

app/src/main/java/org/mozilla/fenix/onboarding/view/UpgradeOnboardingScreen.kt
gabrielluong: I briefly eluded to this in https://github.com/mozilla-mobile/fenix/pull/28376/files#r1060997395, but I think the intent of this is to be used only in HomeOnboardingDialogFragment.kt.

app/src/main/java/org/mozilla/fenix/onboarding/view/UpgradeOnboardingScreen.kt
gabrielluong: I don't expect this to work nor do I think we need this.

app/src/main/java/org/mozilla/fenix/onboarding/view/UpgradeOnboardingScreen.kt
MatthewTighe: @MozillaNoah were you suggesting something like this?

```
enum class OnboardingPageUiState(
    @DrawableRes val image: Int,
    @StringRes val title: Int,
    @StringRes val description: Int,
    @StringRes val primaryButtonText: Int,
    @StringRes val secondaryButtonText: Int? = null,
    val recordImpressionEvent: () -> Unit,
) {
    Welcome(
        image = ...
    ),
    SyncSignIn(
        image = ...
    )
}
```

If so, then I'd +1 it, as I think it well encapsulates the static differences between the pages well. I do think we will need to move the metrics callbacks into a a `Controller` class and passed to the Composables per gl's comment below in order to get previews rendering.

app/src/main/java/org/mozilla/fenix/onboarding/view/UpgradeOnboardingScreen.kt
MatthewTighe: I am interested in continuing the discussion of simplifying the Controller/Interactor pattern, so I added it as a topic to our next experience team sync.

gabrielluong: @MatthewTighe Sounds good to me. I think my ask would always be to start up a RFC - it doesn't have to be long or anything, but we just want to do some due diligence and provide a forum to address any concerns from the wider team.

app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingPageUiState.kt
```diff
@@ -0,0 +1,20 @@
...
+ */
+data class OnboardingPageUiState(
```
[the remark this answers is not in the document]
gabrielluong: +1 on `onRecordImpression` convention. I do like the idea with following a callback naming convention that has been adopted in the wider project.

app/src/main/java/org/mozilla/fenix/onboarding/view/UpgradeOnboardingScreen.kt
gabrielluong: @MatthewTighe This is also an interesting idea.

rahulsainani: The idea was that `OnboardingPage` is in a different component and doesn't need to who/what uses it, similar to a `Button`, the callers can keep the objects and pass on the data, as there might be more features using the OnboardingPage. Also makes it possible to move `OnboardingPage` to a different module, let me know your thoughts

app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingPageUiState.kt
rahulsainani: Think the params are quite self descriptive :) Similar to the comment here --> https://github.com/mozilla-mobile/fenix/pull/28376#discussion_r1061006455

app/src/main/java/org/mozilla/fenix/onboarding/view/UpgradeOnboardingScreen.kt
rahulsainani: Hmm yeah, the idea was to start the function with Preview so it doesn't pollute autocomplete, let me know what you think? 
If there's a convention, can i read it somewhere?

app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingPageUiState.kt
rahulsainani: I ended up adding the param docs as it might be used by other callers 👍

app/src/main/java/org/mozilla/fenix/onboarding/view/UpgradeOnboardingScreen.kt
rahulsainani: Btw the preview should work, I think maybe because metrics are now only called in side effects (click listeners and effect handlers), and they are not in the composition.

app/src/main/java/org/mozilla/fenix/onboarding/view/UpgradeOnboardingScreen.kt
rahulsainani: Yeah the previews work, I think maybe because the metrics ping is in the side effects

app/src/main/java/org/mozilla/fenix/onboarding/view/UpgradeOnboardingScreen.kt
gabrielluong: We have mostly adopted `Preview` being the suffix everywhere because that's how it is in the Compose Preview docs https://developer.android.com/jetpack/compose/tooling/studio#preview

app/src/main/java/org/mozilla/fenix/onboarding/view/UpgradeOnboardingScreen.kt
MatthewTighe: That is a good point. If we consider this with the possibility of modularizing Fenix in the future I think our options are:

1. Treat onboarding as a complete feature vertical, in which case my suggestion above would still work. Other features that required onboarding could come add their static properties in the onboarding module itself.
2. Define onboarding pages per feature module and pass them to the onboarding module, as it sounds like you're suggesting here. If we went this route, another convention you could consider is to define a top-level `val` for each onboarding page. For example, in `/feature/onboarding/Onboarding.kt` we could have `val welcomeOnboarding = OnboardingPageUiState(...)`.

I don't necessarily have a strong opinion here, but I would lean towards the 1st option still. My impression when onboarding is discussed here is that we will be treating it as a distinct feature in terms of additions, experimentation, etc. I think my example in 2 also illustrates what could become a slightly awkward module structure - it would make sense to have something like `/feature/sync/Onboarding` or `/feature/notifications/Onboarding`, but then those instances are split from any number of pages that could end up defined in `/feature/onboarding/Onboarding`.

It also really doesn't matter for now, I don't think, as this is all hypothetical work. Just wanted to share some of my thoughts. Plans for the future never survive first contact anyway, so I'd stick with whatever feels most comfortable to you 🙂

app/src/main/java/org/mozilla/fenix/onboarding/view/UpgradeOnboardingScreen.kt
MozillaNoah: Because the preview functions are private, they won't pollute the global namespace.

app/src/main/java/org/mozilla/fenix/onboarding/view/UpgradeOnboardingScreen.kt
MozillaNoah: @MatthewTighe, yes exactly that!

rahulsainani: @MatthewTighe thanks for the explanation and the thought process. I think that's a bit outside the scope of this PR. Also, Welcome and Sync Sign In Onboarding uses Indicators (which are not touched) and the upcoming notification pre permission dialog doesn't use it, that's why the smallest possible component (OnboardingPage) was extracted so we can reuse that part. I'm open to commonalising more parts of the UI but I would say we should do that going forward as things get more clear based on the usages. wdyt?

app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingPage.kt
```diff
@@ -0,0 +1,120 @@
...
+        ) {
+            PrimaryButton(
+                text = stringResource(id = onboardingPageState.primaryButtonText),
+                onClick = onPrimaryButtonClick,
+            )
+
+            if (onboardingPageState.secondaryButtonText != null) {
+                Spacer(modifier = Modifier.height(8.dp))
+                SecondaryButton(
+                    text = stringResource(id = onboardingPageState.secondaryButtonText),
```
MatthewTighe, line 88: I think we can leave the named `id` parameter out of these `stringResource` calls

app/src/main/java/org/mozilla/fenix/onboarding/view/UpgradeOnboardingScreen.kt
MatthewTighe, line 64: As gl mentioned in [this comment](https://github.com/mozilla-mobile/fenix/pull/28376/files#r1060997395), the pattern I usually see is to host top-level state and events in the parent fragment. This will make it easier to move these metrics calls into a controller as well. An example of that could be the [WallpaperSettingsFragment](https://searchfox.org/mozilla-mobile/rev/825ba5537cdc9901b5a62bd4e8f9016da29e7c9d/fenix/app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettingsFragment.kt#61).

Basically, I think we could remove the `UpgradeOnboardingScreen` entirely here, and call directly to `OnboardingContent` though it would likely make sense to change the name as well.

MatthewTighe (approved):
This is looking pretty good to me! I just had a couple more small things, including some discussion about moving some of these click handlers into a controller eventually. I think that could be done as a follow-up in order to get this moving again, but it would be good to file an issue for it. 

Since the time gap is so large, I'm going to approve for now to unblock you with the assumption that my few small concerns are addressed. That said, it might be worth running it by gl one more time if there is any overlap for the two of you.

app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingPage.kt
gabrielluong: I have discussed this with @MozillaNoah a long time ago, but I think it's fine to just leave these `id` in since they are autofilled when you autocomplete `stringResource` and I rather not have to force developers to delete it.

app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingPageState.kt
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Model containing data for the [OnboardingPage].
+ *
+ * @param image [DrawableRes] displayed on the page.
+ * @param title [StringRes] of the page.
+ * @param description [StringRes] of the page.
+ * @param primaryButtonText callback for primary button click.
+ * @param secondaryButtonText callback for secondary button click.
+ * @param onRecordImpressionEvent callback for recording impression event.
```
gabrielluong, line 18:
```suggestion
 * @param title [StringRes] of the onboarding headline text.
 * @param description [StringRes] of the onboarding body text.
 * @param primaryButtonText [StringRes] of the primary button text.
 * @param secondaryButtonText [StringRes] of the secondary button text.
 * @param onRecordImpressionEvent callback for recording impression event.
```

app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingPage.kt
```diff
@@ -0,0 +1,120 @@
...
+fun OnboardingPage(
+    onboardingPageState: OnboardingPageState,
+    onPrimaryButtonClick: () -> Unit,
+    onSecondaryButtonClick: () -> Unit,
+    modifier: Modifier = Modifier,
+) {
+    Column(
+        modifier = modifier.semantics(mergeDescendants = true) {},
+        horizontalAlignment = Alignment.CenterHorizontally,
+        verticalArrangement = Arrangement.SpaceEvenly,
```
gabrielluong, line 49: **This is a blocking change**: This is unfortunately gonna change the layout from the designs in https://www.figma.com/file/d7l1iGabQYqnLcnQbHDGzM/Wallpaper-2.0-%2B-Onboarding-v106?node-id=6058%3A209483&t=bMa3YpVzO7xeRq9s-0. In particular, there's too much space between the Buttons and the Indicator when we would ideally want the Buttons to be at the bottom. We should ideally not change the layout from the desired specs in this refactoring.

Before (left) vs After (right):
Screen Shot 2023-01-05 at 10 10 02 PM

gabrielluong (changes requested):

app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingPage.kt
```diff
@@ -0,0 +1,120 @@
...
+    ) {
+        Image(
+            painter = painterResource(id = onboardingPageState.image),
```
gabrielluong, line 52: I believe the solution to our image scale issue can be solved via `contentScale = ContentScale.FillBounds`. https://developer.android.com/reference/kotlin/androidx/compose/ui/layout/ContentScale.Companion#FillBounds()

We currently use this in https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/res/layout/fragment_turn_on_sync.xml#38.

app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingPage.kt
```diff
@@ -0,0 +1,120 @@
...
+ */
+@Composable
+fun OnboardingPage(
+    onboardingPageState: OnboardingPageState,
+    onPrimaryButtonClick: () -> Unit,
+    onSecondaryButtonClick: () -> Unit,
+    modifier: Modifier = Modifier,
+) {
+    Column(
+        modifier = modifier.semantics(mergeDescendants = true) {},
```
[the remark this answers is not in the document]
rahulsainani: Good catch! removed it! The idea before the refactoring was to merge image and title with one parent so this might have made sense then, not anymore! 
Yeah this modifier merges the composables in the semantics tree, way better explanation than I can provide --> https://www.youtube.com/watch?v=iV0h_6G3CYk

app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingPage.kt
rahulsainani: Umm I think this specific preview size works best for the left option but a different preview size might work better for the option on the right. At this moment, none of these are ideal, but as we discussed in the [dimensions + box with constraints PR](https://github.com/mozilla-mobile/fenix/pull/28348) and slack, that we could use grids and ratios to solve the problem of supporting different screen sizes. Adding a hard coded dimension here to make the button move down would look good in our previews but won't on different screen sizes. If you have a concrete solution, I'm open.

app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingPage.kt
rahulsainani: I'm afraid that doesn't work and it's not related to this change, let's move this thread to the relevant issue

app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingPageState.kt
```diff
@@ -0,0 +1,27 @@
...
+ */
+data class OnboardingPageState(
+    @DrawableRes val image: Int,
+    @StringRes val title: Int,
+    @StringRes val description: Int,
+    @StringRes val primaryButtonText: Int,
+    @StringRes val secondaryButtonText: Int? = null,
```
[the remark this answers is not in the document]
rahulsainani: We could've, it's just another approach, an advantage of using `Int`s is that these can be file level variables but if they're composables they can only be accessed from a composable scope.

app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingPage.kt
gabrielluong: One approach would be to pursue the original `SpaceBetween` approach with the 3 containers being:
- Dismiss button
- Image and text
- Buttons and Indicator
I do understand this change will likely break some of the foundations how you have chosen to refactor each section.

The spacing between the Button and Indicator does fall between 24dp and 32dp depending on whether or not the Secondary Button is shown. Since 24 and 32 are actual spacing values that we can find in the [Spacing Spec](https://www.figma.com/file/pEyGeE4KV5ytYHeXMfLcEr/Mobile-Styles?node-id=1017%3A56444&t=MxgC2NnW8WhXmN0q-0), we should treat the spacing here as intentional. Using `SpaceEvenly` would not satisfy this requirement because the buttons would be anchored to the Column with image and text, and disconnected from the indicators. There's also 16dp spacer between the button and indicator in [OnboardingContent](https://github.com/mozilla-mobile/fenix/pull/28376/files#diff-cee11c3fc1e078c563b8bc30d76866b9f01ead3fe2f7e1deebf70e271744cf6dR161) so I would imagine the spacing between the button and indicator would never follow the spec even if we had a grid system. 

While I do understand what you're saying, we can't simply just deviate from the original design spec for this reasoning alone since we have already gotten prior UX review for this. 

My ask here is that we shouldn't change the layout from the status quo in a refactoring. If we want to address the layout and scaling issues, let's address that separately or fix it before this refactoring so we are better guided on how to best refactor this. The underlying problem that I still want to focus on is only the spacing between the indicators and buttons.

rahulsainani: Hmm, updated it, could you take a look again?

gabrielluong: I discussed what I was still seeing with @rocketsroger and @MozillaNoah. There are still fundamental issues with how things are laid out that are breaking the design specs with the current approach. I would still suggest laying things out based on how I have described it and can be seen in the original implementation:

Column [Space Between]:
  - Dismiss button
  - Image and text
  - Buttons and Indicator

In your current implementation, we can't use `SpaceBetween` between the image and text because the spacing between them will now be dynamic. The image and text should be in the same container with a 32dp spacer between them to match the designs:
Screen Shot 2023-01-09 at 5 38 23 PM

The spacing between the button and indicators are also dependent on how many buttons are shown:
| One Button | Two Buttons |
|---|---|
|Screen Shot 2023-01-09 at 5 41 46 PM|Screen Shot 2023-01-09 at 5 41 51 PM|

The `Spacer` as the first item in this `SpaceBetween` Column is also a bit of a bandaid solution that is masking other problems, which I don't really recommend since the intent of the refactoring was to make all of this better. We have already drastically changed how items are wrapped in their various column containers and as a result the layout has also changed. We should proceed somewhat cautiously to avoid further review churn. I am aware this is high priority work, but my understanding from @rocketsroger was that there is still a path forward that does not depend on this refactoring.

gabrielluong (changes requested):
See https://github.com/mozilla-mobile/fenix/pull/28376#discussion_r1065184222

rahulsainani:
Closing this as #28630 and further PRs will address commonalising the `OnboardingPage` component.

